PR #197, opened 2026-05-12
fix(desktop): stabilize pending workspace shell handoff

pablonyx:
## Summary

- stabilizes the pending workspace/new worktree shell so header, model selector, footer, tabs, and sidebar render from projected state immediately
- unifies prompt/config/interaction work under ordered session intents so queued user work renders immediately and dispatches after materialization
- preserves projected cloud/worktree sessions through pending-to-real handoff, including selected-cloud-runtime rehydration
- keeps queued session intents and projected records intact if background runtime session creation fails
- removes the orphan generated cloud OpenAPI copy and points generation at the existing access-layer client

## Why

New workspace/session creation has unavoidable filesystem, cloud, and AnyHarness runtime latency. The desktop shell should render from projected state immediately, then remap to real workspace/session ids without creating duplicate tabs or dropping queued work.

This also addresses review blockers around cloud runtime rehydration, projected-session materialization failures, cross-session intent reconciliation, duplicate pending-shell entry implementations, and dead generated code.

## Verification

- `pnpm --dir desktop test`
- `pnpm --dir desktop exec vitest run src/hooks/workspaces/use-workspace-entry-flow.test.ts src/hooks/workspaces/use-workspace-entry-actions.test.tsx src/hooks/workspaces/workflows/use-pending-workspace-session-materialization.test.tsx src/lib/domain/sessions/intents/session-intent.test.ts src/hooks/sessions/use-session-creation-actions.test.ts`
- `pnpm --dir desktop exec tsc --noEmit`
- `git diff --check`


pablonyx:
@greptile

PR #202, opened 2026-05-13
refactor(anyharness): centralize product MCP routing

pablonyx:
## Summary

- Add shared Product MCP server infrastructure for definition, context resolution, dispatch, auth headers, and JSON-RPC responses.
- Centralize Product MCP launch selection and injection under `sessions/mcp_bindings`.
- Migrate `reviews`, `subagents`, and `workspace_naming` to product-specific `mcp/` modules while preserving legacy routes and headers for already-running sessions.
- Add product-scoped MCP capability tokens and prevent product tokens from validating as legacy workspace/session tokens.

## Validation

- `cargo fmt --all`
- `cargo test -p anyharness-lib mcp`
- `cargo test -p anyharness-lib --lib`
- `git diff --check`


pablonyx (commented):
Did a deep read against `docs/anyharness/specs/product-mcps.md` and the live code. Selection/injection landed properly via `sessions/mcp_bindings/{selection,injection,product_catalog}.rs`, the `ProductMcpLaunchCatalog` cleanly owns launch, and the dispatcher is small and well-scoped. A few concrete issues though — one is a correctness regression, a couple are spec-level placement smells, and the test story is thinner than the original plan promised.

## Must-fix

### 1. `resolve_context` runs before method dispatch — breaks `initialize` and `notifications/initialized`

`integrations/mcp/product_server/dispatcher.rs:71`

```rust
let ctx = server.resolve_context(&request_context)?;

match request.method.as_str() {
    \"initialize\" => { … doesn't use ctx … }
    \"notifications/initialized\" => Ok(None),  // doesn't use ctx
    \"tools/list\" => … uses ctx …,
    \"tools/call\" => … uses ctx …,
}
```

This is a behavior change from the per-product `handle_json_rpc` it replaces. Concretely:

- `SubagentProductMcpServer::resolve_context` (`sessions/subagents/mcp/context.rs:11-32`) requires `workspace.surface == \"standard\"` and a valid parent session. If a subagent session's workspace has flipped away from standard, its MCP client can no longer even complete the `initialize` handshake. Same for a torn-down parent session.
- `WorkspaceNamingProductMcpServer::resolve_context` requires the session row to exist and belong to the workspace. If the session was just deleted, `initialize` fails instead of returning a protocol-level response.
- Every `tools/list` now runs DB queries (two for reviews' role resolution). Previously `tools/list` was static for subagents/workspace_naming.

Fix: move `resolve_context` inside the `tools/list` and `tools/call` arms. `initialize` / `notifications/initialized` should be pure protocol concerns and must not depend on product state.

### 2. Context-resolution errors become `400 *_MCP_REQUEST_INVALID`

`api/http/product_mcp.rs:85-87`

```rust
let response = server.dispatch(request, body).await.map_err(|error| {
    ApiError::bad_request(error.to_string(), definition.request_invalid_code)
})?;
```

Any error bubbling out of `resolve_context` — session not found, workspace not found, workspace not standard, cross-workspace session, token race with deleted workspace — now surfaces as `400 REQUEST_INVALID`. Before the refactor, these were typed errors in each product's `handle_json_rpc` and produced appropriate statuses.

This is a silent semantic regression. At minimum, `resolve_context` failures should be distinguishable from body-parse/dispatch failures. Consider returning `Result<Option<Value>, ProductMcpDispatchError>` from `dispatch_product_mcp_request` where the error enum differentiates `ContextUnresolved` (→ 404/409) from `Protocol` (→ 400).

## Should-fix

### 3. Per-product endpoint policy lives in the registry (spec banned shape)

`sessions/mcp_bindings/product_registry.rs:37-47`

```rust
pub fn endpoint_policy(&self) -> ProductMcpEndpointPolicy {
    match self {
        Self::Reviews(_) => ProductMcpEndpointPolicy::SharedWorkspaceMutation(
            WorkspaceOperationKind::ReviewWrite,
        ),
        Self::Subagents(_) => ProductMcpEndpointPolicy::SharedWorkspaceMutation(
            WorkspaceOperationKind::SubagentWrite,
        ),
        Self::WorkspaceNaming(_) => ProductMcpEndpointPolicy::None,
    }
}
```

Per `docs/anyharness/specs/product-mcps.md:144-151` and :782-809, the shared kit must not know product IDs. The *shared endpoint's* decision of \"acquire this operation lease\" is product-specific policy. Registry placement in `mcp_bindings` rather than `integrations/mcp/product_server` is slightly better than the plan's draft, but it still centralizes product-id branches where the spec wants each product to own its policy.

Cleaner: make `endpoint_policy` a field on `ProductMcpDefinition` (e.g., `pub endpoint_operation: Option<WorkspaceOperationKind>`) or a method on `ProductMcpServer`. Then `dispatch_product_mcp` reads it generically and adding a fourth product MCP doesn't require editing `product_registry.rs`.

### 4. Three near-identical auth.rs files — 294 LOC of copy-paste

`domains/reviews/mcp/auth.rs`, `sessions/subagents/mcp/auth.rs`, `sessions/workspace_naming/mcp/auth.rs` are each 98 LOC and differ only by:
- `SECRET_FILE_NAME`
- `LEGACY_CAPABILITY_HEADER_NAME`
- signature algorithm (`LegacySha256Dot` for subagents, `HmacSha256` for the other two — is that intentional? worth a comment if so)
- `product_mcp_id` (already pulled from `super::definition::DEFINITION.id`)

All three `validate_capability_header` bodies are identical. Consider a single `ProductMcpAuth` helper in `integrations/mcp/product_server/auth.rs` taking those four values. The per-product wrappers could shrink to a `new(runtime_home)` constructor.

### 5. Dead code: `mint_legacy_capability_token` and `capability_header_name`

`domains/reviews/mcp/auth.rs:48-55, 31-33` (mirrored in subagents and workspace_naming):

- `mint_legacy_capability_token` has zero callers — fresh injection always mints product-scoped via `mint_capability_token` / `mint_product_mcp_token`. Grep confirms no call sites.
- `capability_header_name()` has zero callers outside cowork.

Per repo rule \"Delete dead code when replacing an implementation\" — these should go.

### 6. Test coverage gaps the original plan called out

The dispatcher has three unit tests (`dispatcher.rs:197-247`). That's it for this PR. The plan called for endpoint/registry/product tests; most are missing:

- **Endpoint** (`api/http/product_mcp.rs`): no tests for 404 on unknown slug, 401 on missing token, legacy header alias success, product-token-in-legacy-header rejection, operation-gate lease acquisition under `SharedWorkspaceMutation`.
- **Auth wrappers**: no tests verifying the exact cross-scope rejections the PR body claims (\"prevent product tokens from validating as legacy workspace/session tokens\"). `capability_token.rs:347-349` has one shallow assertion, but the product wrappers' `validate_capability_header` paths are untested.
- **Selection** (`selection.rs`): no tests. Logic is non-trivial — subagent-child detection, eligibility integration, standard-surface filtering.
- **Injection** (`injection.rs`): no tests that a product-scoped token gets minted, URL is well-formed, bearer is added iff configured, legacy header name isn't used for fresh launches.

Given the correctness bug in #1 and the token cross-scope claim, these aren't nice-to-haves.

## Nits

- **`dispatcher.rs:134-146`**: `TEST_DEFINITION` includes `description` and `display_name` but the production `ProductMcpDefinition` (`definition.rs:15-28`) is missing spec-required fields: `owner_domain`, `default_injection_policy`, `required_capabilities`, `binding_summary_kind`. Fine to defer, but worth a TODO with a link to `product-mcps.md:157-193`.
- **`ProductMcpRequestContext`** (`request.rs:4-22`) uses owned `String`s. Three allocations per request for data that could be `&'a str` borrows from the axum path extraction. Minor.
- **`resolve_context` is a sync trait method that hits the DB** from async axum handlers. Pre-existing in the old code, but this is the moment to consider making it `async` so future implementations can await live capability lookups without `block_in_place`.
- **`selection.rs:28-44`**: `ctx.workspace.surface == \"standard\"` appears twice, and the subagent branch re-runs `find_subagent_parent` per launch. Fine for now, but if selection grows to more products this wants an explicit \"selection input\" struct with precomputed features.
- **Legacy routes** in `api/router.rs:147-158`: kept alongside the generic route. The PR body says they're preserved for already-running sessions — reasonable, but there's no deletion criterion or tracking issue. Worth noting in the PR description or as a follow-up issue: \"delete legacy routes/headers once all in-flight tokens have expired (TTL 12h).\"
- **Unused `CoworkMcpAuth` + `cowork_session_hooks`** in `AppState` (`app/mod.rs:92, 151, 194-201`): this PR's non-goal, but the inconsistency — three new products migrated, cowork untouched — will read as a smell to the next reviewer. A one-line comment above `cowork_session_hooks` linking to a follow-up issue would help.

## Verdict

The architecture is right. Fix #1 (real bug), map #2 (real regression), and file a follow-up for #3–#6. Worth holding the merge on #1 and #2 at minimum.

anyharness/crates/anyharness-lib/src/sessions/workspace_naming/mcp/definition.rs
```diff
@@ -0,0 +1,27 @@
+use crate::integrations::mcp::product_server::{
+    ProductMcpDefinition, ProductMcpPromptPolicy, ProductMcpVisibility,
+};
+
+pub const ID: &str = "workspace_naming";
+pub const ROUTE_SLUG: &str = "workspace-naming";
```
pablonyx: [P2] The generic product endpoint is supposed to be keyed by the product MCP id, and the workspace naming spec lists the route slug as `workspace_naming`. Because `build_http_server` injects `definition.route_slug`, fresh sessions now get `/mcp/workspace-naming` while the token scope/server name use `workspace_naming`; `/mcp/workspace_naming` will 404. Please keep the generic route keyed by `ID` or register the id as the primary lookup, and reserve the hyphenated path for the legacy `/workspace-naming/mcp` alias.

anyharness/crates/anyharness-lib/src/sessions/mcp_bindings/selection.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub session_store: &'a SessionStore,
+}
+
+pub fn select_product_mcps(
+    ctx: ProductMcpSelectionContext<'_>,
+) -> anyhow::Result<Vec<SelectedProductMcp>> {
+    let mut selected = Vec::new();
+
+    if ctx.workspace.surface == "standard" {
+        selected.push(SelectedProductMcp::Reviews);
```
pablonyx, line 34: [P2] This keeps the reviews MCP attached to every standard session, but the product MCP target is role-scoped: reviewer sessions and parent sessions with active review state. As written, unrelated standard sessions get the review MCP config, binding summary, and review system prompt even when `ReviewMcpRole::None` will expose no tools. If broad attachment is still required so a parent can start a review later, that policy needs to be explicit and tested here; otherwise selection should consult review state and only attach when a review role exists.

anyharness/crates/anyharness-lib/src/sessions/workspace_naming/mcp/context.rs
```diff
@@ -0,0 +1,30 @@
...
+    session_store: &SessionStore,
+    request: &ProductMcpRequestContext,
+) -> anyhow::Result<WorkspaceNamingMcpContext> {
+    let session = session_store
+        .find_by_id(&request.session_id)?
+        .ok_or_else(|| anyhow::anyhow!("session not found"))?;
+    if session.workspace_id != request.workspace_id {
+        anyhow::bail!("session does not belong to workspace");
+    }
+    workspace_runtime
```
pablonyx: [P3] The context resolver is only checking existence and ownership, but the workspace naming target says context should resolve naming eligibility and workspace mutability. Since the shared dispatcher calls `tools/list` after this resolver, an already-named or immutable workspace can still advertise `set_workspace_display_name` and only fail later in `calls.rs`. Consider moving the eligibility/mutability checks into context or carrying a typed availability flag so tool listing and calls use the same policy.

pablonyx (commented):
Re-reviewed at `b4d93618`. All six items from the previous pass are meaningfully addressed. A few smaller items remain worth noting.

## Addressed

**#1 `resolve_context` ordering (`dispatcher.rs:97-102`)** — moved inside `tools/list` and `tools/call`; `initialize` / `notifications/initialized` are now pure protocol. New test `initialize_does_not_resolve_product_context` at `dispatcher.rs:252-268` pins this down. 

**#2 Error mapping (`product_mcp.rs:95-110`)** — `ProductMcpDispatchError` now distinguishes `Context(NotFound | Conflict | Internal)` from `Request`, and `map_dispatch_error` emits 404 / 409 / 500 / 400 accordingly. Two focused tests at `product_mcp.rs:137-163` verify it. Good fix — exactly the taxonomy I suggested.

**#3 Product-specific policy (`product_registry.rs:31-37`)** — operation gate kind is now a method on each product server; the registry just asks for `endpoint_operation_kind()`. Still product-id-branched in the `match self { ... }` plumbing, but that's unavoidable until the registry stores `Arc<dyn ProductMcpServer>`. Big improvement.

**#4 Auth duplication (`integrations/mcp/product_server/auth.rs`)** — one `ProductMcpAuth` helper, three per-product wrappers each ~44 LOC (down from 98). Cross-scope rejection test at `auth.rs:99-168` exercises both directions (product-token-in-legacy-header rejected; legacy-token-in-product-header rejected). This is the test I specifically wanted.

**#5 Dead code** — `mint_legacy_capability_token` gone. `capability_header_name` gone from the three migrated products (still present on cowork, which is out of scope).

**#6 Test coverage** — 28 tests across the new modules vs. the prior 3. Particularly valuable additions:
- `dispatcher.rs:252-284`: initialize doesn't touch product context; `tools/list` surfaces context errors without wrapping them in a JSON-RPC response
- `auth.rs:99-168`: the cross-scope matrix
- `product_mcp.rs:137-163`: the error-code mapping
- `injection.rs:161-210`: fresh injection uses the generic route, sets `x-anyharness-product-mcp-token`, and does *not* set `LEGACY_CAPABILITY_HEADER_NAME`
- `selection.rs:121-128`: reviews attach to standard surfaces only

Gaps that remain (endpoint 404/401 integration tests, selection for subagents/workspace_naming) are reasonable to defer.

## Residual items

### `WorkspaceNamingMcpContext::available` re-checks eligibility in calls.rs

`workspace_naming/mcp/context.rs:31-34` resolves `available` from `eligibility::validate_tool_call(...).is_ok() && workspace_access_gate.assert_can_mutate_for_workspace(...).is_ok()`. Then `calls.rs:50-57` re-runs both checks inside `set_workspace_display_name`. The `ctx.available` short-circuit at `calls.rs:21-23` is the fast path; the duplicated checks are the real enforcement. That's correct for TOCTOU reasons — eligibility/mutability can flip between `tools/list` and `tools/call` — but worth a one-line comment so the next reader doesn't simplify it to `if !ctx.available { bail }` and trust the flag alone.

### Subagents context fails `tools/list`, workspace_naming doesn't

Deliberate divergence, but worth calling out:
- `sessions/subagents/mcp/context.rs:24-31` — non-standard workspace returns `Conflict`. Client sees 409.
- `workspace_naming/mcp/context.rs:31-40` — ineligible session returns `Ok` with `available: false` and empty tools. Client sees 200 with no tools.
- `domains/reviews/mcp/context.rs:40` — no role returns `Ok` with `ReviewMcpRole::None` and empty tools. Client sees 200 with no tools.

Reviews and workspace_naming treat "not applicable" as a soft state (empty tool list). Subagents treats it as an error. I can see arguments both ways, but they should align — either all three use the soft pattern (preferred for a long-lived MCP client that keeps polling `tools/list`) or all three hard-fail. Worth a decision captured somewhere.

### Dispatcher's `ProductMcpContextError::Internal` swallows stack

`errors.rs:12-13` uses `#[from] anyhow::Error`. `ProductMcpDispatchError::Context(Internal(...))` is mapped to `ApiError::internal(error.to_string())` at `product_mcp.rs:103-105`, which drops the `anyhow` backtrace. Minor — same behavior as before for internal errors — but if you're going to surface these through tracing, consider logging the full chain before flattening to a string.

### `endpoint_operation_kind` on two products, not three

`reviews/mcp/mod.rs:32` and `subagents/mcp/mod.rs:46` both return a concrete `WorkspaceOperationKind`; `WorkspaceNamingProductMcpServer` doesn't have the method, and the registry hardcodes `None` for that variant (`product_registry.rs:35`). Consistent with behavior, but since the goal was to move product-ID branches onto the servers, it's slightly cleaner to have workspace_naming also implement it (e.g., `fn endpoint_operation_kind() -> Option<WorkspaceOperationKind> { None }`) and the registry just calls it unconditionally. Very small.

### `ProductMcpDefinition` spec fields still deferred

`definition.rs:17-19` now has the comment I asked for — good. Tracking issue link would make the TODO actionable.

### Cowork inconsistency

`app/mod.rs:151, 194-201`: `CoworkMcpAuth` + `CoworkSessionHooks` still on the old pattern. Expected (non-goal) but worth a tracking issue linked from the migration doc so the next person picking this up sees the expected end state.

## Verdict

Good to merge once the TOCTOU comment in `workspace_naming/mcp/calls.rs` lands and the soft-vs-hard unavailability divergence is either aligned or captured in `docs/anyharness/specs/product-mcps.md`. Everything else is cleanup that can follow in separate PRs.

anyharness/crates/anyharness-lib/src/sessions/mcp_bindings/product_registry.rs
```diff
@@ -0,0 +1,95 @@
...
+use crate::integrations::mcp::product_server::{
+    dispatch_product_mcp_request, ProductMcpAuthHeader, ProductMcpDefinition,
+    ProductMcpDispatchError, ProductMcpRequestContext, ProductMcpServer, ProductMcpTokenValidation,
+};
+use crate::sessions::subagents::mcp::SubagentProductMcpServer;
+use crate::sessions::workspace_naming::mcp::WorkspaceNamingProductMcpServer;
+use crate::workspaces::operation_gate::WorkspaceOperationKind;
+
+#[derive(Clone)]
+pub enum ProductMcpEndpointServer {
```
pablonyx: [P2] This registry is still a concrete enum over the current product MCPs, with every endpoint operation re-matched across `Reviews`, `Subagents`, and `WorkspaceNaming`. The migration plan calls for an object-safe endpoint handler wrapper so this layer can dispatch product MCPs generically. As written, adding the next product MCP still requires editing this shared registry for definition/auth/dispatch/policy branches, which is exactly the coupling the product MCP architecture is trying to remove. Consider a `ProductMcpEndpointHandler` trait with an adapter around any `ProductMcpServer`, then store `Arc<dyn ProductMcpEndpointHandler>` keyed by product id/route.

anyharness/crates/anyharness-lib/src/sessions/workspace_naming/mcp/context.rs
```diff
@@ -0,0 +1,41 @@
...
+        .ok_or_else(|| ProductMcpContextError::not_found("session not found"))?;
+    if session.workspace_id != request.workspace_id {
+        return Err(ProductMcpContextError::conflict(
+            "session does not belong to workspace",
+        ));
+    }
+    let workspace = workspace_runtime
+        .get_workspace(&request.workspace_id)?
+        .ok_or_else(|| ProductMcpContextError::not_found("workspace not found"))?;
+    let available = eligibility::validate_tool_call(session_store, &workspace, &session).is_ok()
```
pablonyx: [P2] This turns every eligibility/access error into `available = false`. That is okay for expected policy denials, but it also swallows store/access failures from `validate_tool_call` or `assert_can_mutate_for_workspace` and makes the MCP silently expose no tools or return a generic “not available” tool error. Context resolution should preserve unexpected failures as `ProductMcpContextError::Internal` and only convert typed, expected policy failures into unavailable/conflict state. Otherwise real backend faults become indistinguishable from “workspace already named”.

anyharness/crates/anyharness-lib/src/api/http/product_mcp.rs
```diff
@@ -0,0 +1,239 @@
...
+        .validate_capability_token(auth_header, &request)
+        .map_err(|error| ApiError::internal(error.to_string()))?;
+    if !validation.is_valid() {
+        return Err(ApiError::unauthorized(
+            "Invalid product MCP capability token.",
+            definition.unauthorized_code,
+        ));
+    }
+
+    let _lease = match server.endpoint_operation_kind() {
```
pablonyx: [P3] The dispatcher now keeps `initialize` and `notifications/initialized` protocol-only, but this outer route gate still runs before the JSON-RPC method is parsed. For reviews/subagents, a frozen or otherwise non-mutable workspace cannot even initialize the MCP server or list/read tools, even though those methods do not perform product writes. If this route policy is intended to protect mutations, consider making it method-aware or moving the operation/mutability gate to the product call paths that actually mutate state.

pablonyx (commented):
Re-reviewed at `b3f7a825` after the three follow-up commits. The architecture work is in great shape — the `ProductMcpEndpointHandler` trait, tool-level mutation gating, and broad-review-preload explanation all land cleanly. Below are items I spotted this pass.

## Nicely addressed

- **Registry now uses `Arc<dyn ProductMcpEndpointHandler>`** (`product_registry.rs:131-148`) with a generic `ProductMcpEndpointHandlerAdapter` that wraps any `ProductMcpServer`. This is the clean shape I hoped for last round — adding a fourth product MCP is now `Arc::new(ProductMcpEndpointHandlerAdapter::new(...))` rather than a registry `match` edit.
- **Per-tool mutation gating** via `ProductMcpEndpointOperation::from_request_body` + `MUTATING_TOOL_NAMES` (`subagents/mcp/tools.rs:7-11`, `reviews/mcp/tools.rs:6`). Smart move: `get_subagent_status` and `read_subagent_events` no longer acquire a shared `SubagentWrite` lease, which was an actual perf/contention issue on the old `endpoint_operation_kind()` method.
- **Legacy route handlers collapsed** to six thin delegators in `product_mcp.rs:47-117`. The per-product `api/http/reviews.rs`/`subagents.rs`/`workspace_naming.rs` MCP endpoints are gone — repo rule \"Do not leave duplicate old and new code paths behind\" now holds at the transport layer.
- **Reviews broad-preload rationale** (`selection.rs:28-35`, :64-70) is now documented in code: the MCP server attaches broadly to standard workspaces, but the review-specific system prompt no longer leaks into unrelated sessions, and the endpoint resolves role per-request. `broad_reviews_selection_does_not_add_review_prompt_text` test at :136-141 pins it.

## Issues

### 1. `ReviewProductMcpServer`'s `SYSTEM_PROMPT_APPEND` and `system_prompt_append()` are now dead code

`domains/reviews/mcp/definition.rs:16, 32-34` define a review-specific system prompt. Previously `selection.rs` extended `extras.system_prompt_append` with it. After commit `b3f7a825`, that branch was deliberately dropped (broad preload should not append review prompt text to unrelated sessions — good call). But the constants/function remain — grep shows zero callers:

```text
$ grep -rn 'reviews::mcp::definition::system_prompt_append\|reviews_mcp::definition::system_prompt_append\|reviews::mcp::definition::SYSTEM_PROMPT_APPEND' anyharness/crates/anyharness-lib/src
# no results
```

Per the repo rule \"Delete dead code when replacing an implementation.\" `cargo check` passes silently because `pub const` / `pub fn` are always reachable from the public API. Either delete or convert the prompt constants into something the review runtime actually injects when a review session starts (reviewer prompt is already handled by `reviews/runtime_helpers.rs::reviewer_system_prompt_append`, so this may just be orphaned parent-session prompt text — the orphan is the evidence the earlier plan had scope issues).

### 2. `ProductMcpEndpointOperation` parses the body twice

`product_mcp.rs:133` calls `ProductMcpEndpointOperation::from_request_body(&body)` before handing the same body to `server.dispatch(request, body)`. `dispatch_product_mcp_request` at `dispatcher.rs:55` parses the body again into `JsonRpcRequest`.

Two copies of \"which JSON-RPC method is this\" logic now exist: `ProductMcpEndpointOperation::from_request_body` (manual `body.get(\"method\")...`) and `dispatcher.rs` (full `serde_json::from_value::<JsonRpcRequest>`). They can drift. A malformed body that fails the dispatcher's parse will still have passed `from_request_body` as `ProductMcpEndpointOperation::Other`, which happens to be safe (no lease taken), but any future logic that keys off `Initialize` / `ToolsList` before dispatch risks divergence.

Cleaner: parse once in the endpoint, pass the parsed method (or `ProductMcpEndpointOperation`) into `dispatch`, or invert — have `dispatcher` return the operation kind alongside the response. Not a blocker, but the duplicated extraction is a code smell the trait abstraction doesn't fully solve.

### 3. Lease acquired after parsing, before body validation

`product_mcp.rs:150-160` acquires the workspace operation lease before `dispatch`. If the body is `tools/call { name: \"create_subagent\" }` but the params fail deserialization (missing `prompt`), the handler now:
1. Takes the `SubagentWrite` lease
2. Runs `resolve_context` (DB read)
3. Hands the body to dispatcher
4. Dispatcher returns a `tool_error` result
5. Lease drops

Under contention this serializes malformed-payload errors through the workspace write gate. Low-impact but non-ideal. Could check arg validity inside `from_request_body` or defer lease to inside `tools/call` after context resolution.

### 4. `dispatcher.rs` still resolves context inside the `tools/call` arm *before* argument parsing

`dispatcher.rs:101-114` resolves context, then parses params, then bails on invalid params. If a subagent client calls `tools/call` with a malformed body against a non-standard workspace, the user sees `409 subagents are only available in standard workspaces` (from context) rather than the actual `-32602` invalid params error. The more informative error is masked by an unrelated precondition.

Flip the order: parse params first, then resolve context. The parse is cheap and pure; context is where the DB reads live. Same argument applies less urgently to `tools/list`, since that has no params to parse.

### 5. `product_mcp.rs` imports three `definition` aliases for legacy routes only

```rust
use crate::domains::reviews::mcp::definition as reviews_mcp_definition;
use crate::sessions::subagents::mcp::definition as subagents_mcp_definition;
use crate::sessions::workspace_naming::mcp::definition as workspace_naming_mcp_definition;
```

These exist only to hand `ROUTE_SLUG` constants to `dispatch_product_mcp`. Six legacy handler functions, three per-product `use` lines, three `::ROUTE_SLUG` constants — a modest amount of scaffolding for backward-compat routes that will eventually be deleted. Consider collapsing to one `dispatch_legacy(state, ...slug...)` and letting the router pass the slug via `Path<String>` or a closure. Not urgent; flagging because \"Do not leave duplicate old and new code paths behind\" applies here once legacy tokens expire.

### 6. `endpoint_operation_kind` returns `None` for `tools/call` with missing `tool_name`

`product_registry.rs:100-102` short-circuits to `None` when `tool_name` is missing. That means a malformed `tools/call { }` with no `name` skips the lease entirely and then fails inside the dispatcher with `-32602 invalid params`. Harmless, but worth a comment — a future reviewer might think this is a silent auth/lease bypass.

### 7. `MUTATING_TOOL_NAMES` lives in `tools.rs`, not `definition.rs`

Small nit: the per-product `ProductMcpEndpointHandlerAdapter::new(..., MUTATING_TOOL_NAMES)` wiring pulls a constant from the `tools` module (`reviews/mcp/tools.rs:6`, `subagents/mcp/tools.rs:7-11`). `tools.rs` is meant for `tools/list` definitions per the product-mcps spec (`product-mcps.md:322-337`). A single list of mutating names is closer to static metadata — it belongs on `ProductMcpDefinition` or next to it, so \"what does this MCP mutate\" is one lookup, not two. Fixing this now would make test #8 below cheap.

### 8. `ProductMcpEndpointHandlerAdapter`'s `endpoint_operation_kind` logic has zero tests

`product_registry.rs:93-108` is the actual lease-gate decision. Tests cover `ProductMcpEndpointOperation::from_request_body` at :157-191 but not the adapter's operation→lease mapping. Consider a test that:
- `tools/call { name: create_subagent }` with `write_operation_kind: Some(SubagentWrite)` → `Some(SubagentWrite)`
- `tools/call { name: get_subagent_status }` → `None`
- `tools/call { name: unknown }` → `None`
- `tools/list` → `None`
- `initialize` → `None`

This is the primary new enforcement surface in the PR. Worth pinning.

## Nits

- **Legacy handler duplication**: the 6 functions in `product_mcp.rs:47-117` could be a macro or a small higher-order function. Six tiny copies is fine now; if a fourth product MCP needs a legacy route, factor first.
- **`WorkspaceNamingProductMcpServer` passes `&[]` as mutating tools** (`app/mod.rs:343`). Correct (workspace_naming gates mutation inside the tool body), but worth a comment on why — otherwise a future reader may assume it's an oversight and \"fix\" it to `MUTATING_TOOL_NAMES`.
- **`product_registry.rs:14-20`** exposes `Initialize`, `InitializedNotification`, `Other` variants even though `endpoint_operation_kind` only acts on `ToolsCall`. They're self-documenting but drag in the full JSON-RPC vocabulary for what is effectively a single-variant decision. Consider `ProductMcpEndpointOperation::MutationCandidate { tool_name: Option<String> } | NonMutating`.

## Verdict

Real improvement. The `ProductMcpEndpointHandler` trait is the right shape and the per-tool mutation gate is a genuine functional upgrade, not just cosmetic. I'd merge after #1 (delete dead `SYSTEM_PROMPT_APPEND`) and #8 (one-screen test for `endpoint_operation_kind`); the rest is cleanup that can follow.

anyharness/crates/anyharness-lib/src/sessions/subagents/mcp/auth.rs
```diff
@@ -0,0 +1,47 @@
...
+impl SubagentMcpAuth {
+    pub fn new(runtime_home: PathBuf) -> Self {
+        Self {
+            // Subagents used the legacy SHA256-dot token format before the
+            // generic Product MCP endpoint. Keep the same secret algorithm so
+            // already-running legacy sessions can still validate old headers.
+            inner: ProductMcpAuth::new(
+                runtime_home,
+                SECRET_FILE_NAME,
+                McpCapabilityTokenSignature::LegacySha256Dot,
```
pablonyx, line 26: [P2] This keeps fresh `x-anyharness-product-mcp-token` subagent launches on the legacy sha256-dot signature algorithm. The compatibility requirement is for already-launched `x-subagent-session-token` workspace/session tokens; it should not force newly minted product MCP tokens to use the old signing mode. Consider giving `ProductMcpAuth` separate product-token and legacy-token issuers/signatures so fresh generic-header tokens use the shared HMAC product envelope while legacy headers still validate old subagent tokens with `LegacySha256Dot`.

anyharness/crates/anyharness-lib/src/sessions/mcp_bindings/product_registry.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+#[derive(Clone, Default)]
+pub struct ProductMcpEndpointRegistry {
+    by_slug: HashMap<&'static str, Arc<dyn ProductMcpEndpointHandler>>,
+}
+
+impl ProductMcpEndpointRegistry {
+    pub fn new(servers: Vec<Arc<dyn ProductMcpEndpointHandler>>) -> Self {
+        let by_slug = servers
```
pablonyx: [P3] `new` silently collects handlers by `route_slug`, so duplicate slugs overwrite earlier handlers and the registry still has no primary product-id lookup or definition list. The product MCP spec calls out id lookup plus listing definitions as registry responsibilities, and those checks become important as soon as the next MCP or temporary route alias lands. Consider storing by product id as the primary key, explicitly registering any route aliases, and rejecting duplicate ids/slugs at startup instead of letting the last handler win.

pablonyx (commented):
Re-reviewed at `66a04475` with the spec (`docs/anyharness/specs/product-mcps.md`) open alongside. Commit `66a04475` genuinely hardens two of the previous-round items and adds a couple of new wins. Below are the remaining gaps between the PR and the spec, ordered by how close they get you to the target shape.

## Nicely addressed in `66a04475`

**Registry uniqueness guarantees (`product_registry.rs:162-195`)**: `ProductMcpEndpointRegistry::new` now returns `Result` and rejects duplicate product IDs, duplicate route slugs, and duplicate aliases. `AppStateInitError::InvalidProductMcpRegistry` wires the failure into startup. Four registry tests at :366-427 cover it. This is the kind of \"construction-time invariant\" the spec implies when it says there's one registry that owns lookup (spec :449-468).

**Registry now supports both lookup keys** (`product_registry.rs:197-213`): `get_by_product_id`, `get_by_route_slug`, and `definitions()`. That closes the spec bullet \"lookup by product_mcp_id, list definitions for UI/catalog/summaries\" (spec :456-460). 

**Split signature algorithms per scope (`product_server/auth.rs:29-44`)**: `ProductMcpAuth` now holds two `McpCapabilityTokenIssuer` — one for fresh product-scope tokens (HMAC-SHA256), one for legacy workspace/session tokens (whatever algo that product used before). Subagents concretely uses `HmacSha256` for new, `LegacySha256Dot` for old (`subagents/mcp/auth.rs:25-26`). This removes the previous footgun where validating a legacy token required the old issuer's signature algorithm, and it's now independently tested at `auth.rs:181-240`. Good hardening.

**Route aliases as a registry-level concept** (`product_registry.rs:64-87`): `ProductMcpEndpointRegistration::with_route_aliases` lets a product MCP be reachable under both the generic slug and a legacy slug via the registry itself, not via duplicate handler functions. Not yet used (the legacy routes in `api/http/product_mcp.rs:47-117` still hand-delegate), but the foundation is in place.

## Residual gaps vs the spec

The following items bring the PR closer to the `product-mcps.md` target shape. Ordered roughly by value.

### 1. Dead reviews prompt constants (blocker)

Repeating from last round because this PR did not remove them:
- `domains/reviews/mcp/definition.rs:16` — `SYSTEM_PROMPT_APPEND` const
- `domains/reviews/mcp/definition.rs:32-34` — `system_prompt_append()` fn
- Grep confirms zero callers across the workspace.

Repo rule: \"Delete dead code when replacing an implementation.\" `cargo check` won't catch it because `pub const` / `pub fn` are reachable from the public API. Either delete, or — if the intent is to re-inject once live MCP refresh lands — leave a `// TODO:` comment with a tracking issue so it isn't read as a wiring mistake.

### 2. Use `ProductMcpEndpointRegistration::with_route_aliases` instead of the six legacy handlers

`api/http/product_mcp.rs:47-117` has six nearly identical functions (three `GET`, three `POST`) that simply call `dispatch_product_mcp` with a hard-coded `ROUTE_SLUG`. The infrastructure to remove these now exists — each product can register:

```rust
ProductMcpEndpointRegistration::with_route_aliases(handler, &[\"subagents\", \"reviews\", ...])
```

Then the router needs exactly one `GET/POST /.../sessions/{session_id}/mcp/{product_mcp_slug}` route that matches both the generic slug and the alias. The legacy `/subagents/mcp` path becomes an alias in the registry rather than a hand-routed endpoint.

You'd also need to collapse the legacy URLs to the same `{product_mcp_slug}` shape. Worth checking: if `/workspaces/{id}/sessions/{id}/subagents/mcp` is the legacy shape and `/workspaces/{id}/sessions/{id}/mcp/subagents` is the target, these aren't alias-able via slug alone — you'd need per-route router entries that map each legacy URL to a specific slug. At that point, an `async fn dispatch_legacy_slug(slug: &'static str)` closure-factory is cleaner than six hand-written delegators.

Spec target (`product-mcps.md:556-567`): \"Preferred generic endpoint... Migration routes may remain temporarily\". The \"temporarily\" is doing real work — once the per-product handler functions exist in `api/http/product_mcp.rs`, they're not temporary unless someone goes back and deletes them.

### 3. Add `endpoint.rs` to the shared kit

Spec target file layout (`product-mcps.md:69-84`) calls for:
```
anyharness-lib/src/integrations/mcp/
  product_server/
    mod.rs
    definition.rs
    dispatcher.rs
    endpoint.rs       <-- missing
    errors.rs
    request.rs
    response.rs
```

`api/http/product_mcp.rs::dispatch_product_mcp` is exactly what `endpoint.rs` should own: registry lookup → auth header read → token validation → optional lease → dispatch → error mapping. Right now that's 88 lines in the `api/http/` tree that depend on `AppState`, but the actual logic is agnostic — it needs `&ProductMcpEndpointRegistry`, `&WorkspaceOperationGate`, and the mutability assertion. Moving it to `integrations/mcp/product_server/endpoint.rs` would:
- match the spec's target tree exactly
- let `api/http/product_mcp.rs` shrink to thin route handlers that pass `state` parts into one shared endpoint helper
- put the `map_dispatch_error` helper next to the error types it maps from

Worth looking at whether the spec wants this as a reusable function (`pub async fn serve_product_mcp_request(...)`) or a trait with default methods. Either works.

### 4. `ProductMcpDefinition` is still a narrower-than-spec struct

`definition.rs:17-19` has the comment \"Keep this intentionally narrower than the full Product MCP spec while the first internal MCPs migrate\". Fine for this PR, but:
- Spec required fields (`product-mcps.md:157-193`): `id, display_name, description, owner_domain, route_slug, visibility, default_injection_policy, required_capabilities, binding_summary_kind, prompt_policy`.
- Current struct has: `id, route_slug, acp_server_name, server_info_name, display_name, description, visibility, instructions, unauthorized_code, request_invalid_code, prompt_policy`.

Three spec fields missing (`owner_domain`, `default_injection_policy`, `required_capabilities`, `binding_summary_kind`) and three extra non-spec fields (`acp_server_name`, `server_info_name`, `instructions`, `unauthorized_code`, `request_invalid_code`). The extras aren't wrong — they're real needs — but they suggest the spec itself should be updated to include them, or the struct needs to grow to match the spec's broader model.

Action item: either file an issue to reconcile `product-mcps.md:157-193` with the actual `ProductMcpDefinition` struct, or add the missing fields with sensible defaults. Leaving the comment as is means \"someone will notice later\" which is the exact failure mode of the previous per-product mcp_server split.

### 5. `MUTATING_TOOL_NAMES` lives in `tools.rs`, not `definition.rs`

The file grammar (`product-mcps.md:226-242`) says:
- `definition.rs`: static metadata and injection defaults
- `tools.rs`: data-shaped MCP tools/list definitions

\"Which of my tools mutate workspace state\" is static metadata. Right now `reviews/mcp/tools.rs:6` and `subagents/mcp/tools.rs:7-11` host that constant, and `app/mod.rs` imports them with `_mcp_tools::MUTATING_TOOL_NAMES`. The spec's grammar would put this in `definition.rs` or make it a method on the `ProductMcpServer` trait.

Even better: move it into `ProductMcpDefinition` as `pub mutating_tools: &'static [&'static str]`. That way `ProductMcpEndpointHandlerAdapter::new` can read it off `server.definition().mutating_tools` instead of having it passed in as a third argument that must stay in sync with the tool list. Wiring becomes:

```rust
ProductMcpEndpointHandlerAdapter::new(
    Arc::new(ReviewProductMcpServer::new(...)),
    Some(WorkspaceOperationKind::ReviewWrite),
    // no third argument; read from definition
)
```

### 6. `ProductMcpEndpointHandlerAdapter::endpoint_operation_kind` still has no test

Flagged last round, still missing. This is the *primary new enforcement surface* added by the PR — whether a given `tools/call` acquires the workspace write lease. The registry has solid negative tests, the auth has solid cross-scope tests, the dispatcher has its own parse/dispatch tests — but `adapter.endpoint_operation_kind(operation)` has zero direct coverage. A ~15-line test covering the four branches (`ToolsCall{known_mutator}` → `Some`, `ToolsCall{known_readonly}` → `None`, `ToolsCall{unknown}` → `None`, `ToolsList` → `None`) would close it.

### 7. Injection-metadata-from-registry is still split

Spec target (`product-mcps.md:456-461`):
```
It should provide:
  lookup by product_mcp_id
  list definitions for UI/catalog/summaries
  server dispatch target for generic MCP endpoint
  injection metadata for session launch
```

Current: lookup ✓, definitions ✓, dispatch target ✓, **injection metadata is in `sessions/mcp_bindings/injection.rs`**, not on the registry. `ProductMcpInjectionContext` takes three concrete auth objects (`review_auth`, `subagent_auth`, `workspace_naming_auth`) rather than looking them up by product id. Adding a fourth product MCP means editing `ProductMcpInjectionContext`, editing `ProductMcpLaunchCatalog::new`, editing `inject_product_mcps`.

The spec shape is: registry knows how to ask each server for its injection headers + URL given a session launch context, and `injection.rs` becomes a generic consumer. Right now the registry holds `Arc<dyn ProductMcpEndpointHandler>` which is dispatch-oriented; a parallel (or additional) `injection_metadata(session, workspace) -> SessionMcpServer` method on the handler would consolidate this.

Not a blocker for this PR, but worth a follow-up: once cowork migrates, the three-concrete-auth shape in `ProductMcpInjectionContext` becomes four. This is the last bit of product-id-branching left in the assembly path.

### 8. Per-product `*_mcp_tools::MUTATING_TOOL_NAMES` coupling is silent

`app/mod.rs:325, 335` connects `ReviewProductMcpServer` to `review_mcp_tools::MUTATING_TOOL_NAMES`. If someone renames a tool in `reviews/mcp/tools.rs::reviewer_tool_list` but forgets to update `MUTATING_TOOL_NAMES`, the compiler won't complain and the workspace write lease will silently not be acquired. Two mitigations:

- Unit test in `reviews/mcp/tools.rs` that asserts every `MUTATING_TOOL_NAMES` entry appears in the tool list. Cheap.
- Or (better) make `mutating_tools` data live inside `tool_definition(...)` calls themselves and derive `MUTATING_TOOL_NAMES` from the list. Then renaming a tool also renames the mutating-tools entry automatically.

## Nits

- `subagents/mcp/auth.rs:19-21` finally has the \"why two signatures\" comment I asked for last round. 
- `product_registry.rs:157-214` — `ProductMcpEndpointRegistry` is `Default` and `Clone`. Reasonable, but `Default::default()` creates a registry that responds to no products; if it accidentally gets used instead of `new(...)` nothing breaks at build time. Low-severity but consider making `Default` private or removing it, since registry construction should always go through the duplicate-checking `new`.
- The test constants in `product_registry.rs:237-291` are 55 lines of boilerplate for four tests. If `ProductMcpDefinition` gets a `const fn ProductMcpDefinition::for_testing(id, slug)` helper, those tests collapse to one line each. Low-priority.
- `sessions/mcp_bindings/` → `domains/sessions/mcp_bindings/` move is explicitly out of scope per the original plan, but it'd be nice to see a TODO somewhere noting the transitional path per `product-mcps.md:125-126`.

## Recommendation

Merge blockers:
- **#1** Delete dead `reviews/mcp/definition.rs::SYSTEM_PROMPT_APPEND` + `system_prompt_append()`.
- **#6** Test `endpoint_operation_kind` on the adapter.

Everything else is follow-up PRs. File issues for:
- `#2` remove the six legacy handler functions in favor of alias registration
- `#3` introduce `integrations/mcp/product_server/endpoint.rs`
- `#4` reconcile `ProductMcpDefinition` with `product-mcps.md:157-193`
- `#5 / #8` move `MUTATING_TOOL_NAMES` into `definition.rs` or derive from tool list (kills two items)
- `#7` move injection metadata onto the registry
- path migration to `domains/sessions/**` per spec

Overall: very close to the spec's intended shape now. Five rounds in, this PR has been a good study in how iterative review gets you from \"centralize routing\" to \"one registry, one dispatcher, one endpoint, one auth helper, one injection path\" — which is what the spec was asking for.

anyharness/crates/anyharness-lib/src/sessions/subagents/mcp/context.rs
```diff
@@ -0,0 +1,36 @@
...
+    let workspace = workspace_runtime
+        .get_workspace(&request.workspace_id)?
+        .ok_or_else(|| ProductMcpContextError::not_found("workspace not found"))?;
+    if workspace.surface != "standard" {
+        return Err(ProductMcpContextError::conflict(
+            "subagents are only available in standard workspaces",
+        ));
+    }
+
+    Ok(SubagentMcpContext {
```
pablonyx, line 46: [P2] This context is still thinner than the product MCP contract describes. The subagents doc says context.rs should resolve the parent session record, workspace surface, existing child count, depth/limits, and launch/live config, with tool availability varying from that typed context. Here a token that reaches this endpoint only proves workspace/session ownership and standard surface, while tools/list still advertises create/send/schedule for a disabled parent, a child/grandchild, or a fanout/max-depth-blocked parent and only fails later. Move the spawn eligibility and capability state into this context (or into typed availability flags derived here) so listing and calls share the same policy.

anyharness/crates/anyharness-lib/src/sessions/mcp_bindings/selection.rs
```diff
@@ -0,0 +1,142 @@
...
+            }
+            SelectedProductMcp::Subagents => {
+                extras
+                    .system_prompt_append
+                    .extend(crate::sessions::subagents::mcp::definition::system_prompt_append());
+                extras
+                    .mcp_binding_summaries
+                    .push(crate::sessions::subagents::mcp::definition::binding_summary());
+            }
+            SelectedProductMcp::WorkspaceNaming => {
```
pablonyx, line 80: [P3] Workspace naming now gets the product MCP server and prompt text, but it still never contributes a binding summary. The product MCP spec lists binding summary output as part of launch injection, and the workspace-naming doc calls out a summary indicating that the workspace naming tool is attached. Because assembly only records summaries returned from product_mcp_prompt_extras, fresh naming sessions have no summary for this internal MCP. Add a workspace_naming definition summary and push it here, matching reviews/subagents.

anyharness/crates/anyharness-lib/src/sessions/subagents/mcp/calls.rs
pablonyx, line 1: [P2] If the initial prompt send fails after the child session has been created, linked, and started, create_subagent currently returns the error but leaves the owned child/link behind. The subagents MCP target calls out rollback of child session/link state when launch fails, and the earlier link/start failure branches already clean up. Treat prompt dispatch as part of the create_subagent launch and delete the child/link on that failure too, or return an explicit success state for an intentionally created but unprompted child. Otherwise retries can strand empty children and consume fanout slots.

anyharness/crates/anyharness-lib/src/domains/reviews/mcp/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    fn resolve_context(
+        &self,
+        request: &ProductMcpRequestContext,
+    ) -> Result<Self::Context, ProductMcpContextError> {
+        context::resolve_context(&self.runtime, request)
+    }
+
+    fn tools(&self, ctx: &Self::Context) -> Vec<Value> {
```
pablonyx, line 59: [P2] The new reviews MCP shape still has no focused tests for this role-sensitive surface; `cargo test -p anyharness-lib domains::reviews::mcp` currently runs 0 tests. The product MCP docs call out reviewer-vs-parent tool exposure, no-role rejection, token scope rejection, submit delegation, and unknown-tool-by-role behavior as required coverage. This module is exactly where those behaviors are assembled, so add tests here (or in sibling mcp test modules) before treating the reviews migration as structurally complete.

pablonyx (commented):
Re-reviewed at `a5fe2473`. Three of the four items I flagged as blockers or high-value are now addressed, plus a substantive product-behavior improvement that went beyond the review feedback. Details below.

## Addressed in `a5fe2473`

**#1 Dead review prompt constants — deleted** (`reviews/mcp/definition.rs`). `SYSTEM_PROMPT_APPEND` and `system_prompt_append()` are gone. `cargo check` still clean. ✓

**#6 `endpoint_operation_kind` now tested** (`product_registry.rs:406-428`). Adapter test covers mutating tool → `Some(kind)`, non-mutating → `None`, `ToolsList` → `None`. Exactly the coverage I wanted. ✓

**Soft-fail pattern aligned and documented in the spec** (`product-mcps.md:322-328`):
> Expected but currently unavailable product state should normally resolve to a typed context that advertises fewer tools rather than failing protocol setup. [...] Use a hard context error for missing rows, cross-workspace tokens, corrupt state, or a product MCP attached to a fundamentally unsupported surface.

This is a meaningful addition. It captures the pattern the PR already used for reviews and workspace_naming and extends it to subagents. Good documentation of an architectural convention.

**Subagents `tools/list` now role/state-aware** (`subagents/mcp/tools.rs:55-94`, `context.rs:13-54`):
- `SubagentMcpContext` now carries `can_create`, `create_block_reason`, `existing_subagent_count`, `max_subagents_per_parent`
- `tools/list` conditionally advertises `create_subagent` only when `ctx.can_create`
- Child-interaction tools (`send_subagent_message`, `schedule_subagent_wake`, `get_subagent_status`, `read_subagent_events`) only advertised when `existing_subagent_count > 0`
- Four new tests pin the role-sensitive tool list
- `create_subagent` re-validates `validate_parent_can_spawn` at call time (TOCTOU handling, parallel to what `workspace_naming/mcp/calls.rs:56-57` does with the comment added this round)

This is a bigger upgrade than I flagged — hiding tools when unavailable gives the agent clearer signal and removes a whole class of \"I tried to call create_subagent but it failed\" loops. The subagent MCP is now roughly consistent with reviews (tools vary by role) and workspace_naming (tools vary by availability).

**Review tools now validated against role at call time** (`reviews/mcp/calls.rs:14, 61-73`). Previously `call_tool`'s `match (role, name)` fell through to `unknown tool` for mismatches; now `validate_tool_for_role` runs up front with a distinct error. Three unit tests at :76-110.

**Workspace naming gets a binding summary** (`workspace_naming/mcp/definition.rs:34-43` + `selection.rs:84-86`). Previously selected product MCPs added a binding summary for reviews and subagents only; workspace_naming was silent. Test at :145-158 pins the fix. Minor behavioral completeness win — worth calling out.

**Tool-list drift tests** (`reviews/mcp/tools.rs:90-105`, `subagents/mcp/tools.rs:214-225`): `mutating_tool_names_are_advertised_by_some_review_role` / `mutating_tool_names_are_advertised_when_available` assert every entry in `MUTATING_TOOL_NAMES` appears in some tool list. That's the membership test I suggested — closes the rename-rot risk from #8 last round.

## Remaining items

### 1. `resolve_context` now does 2 extra DB reads per `tools/list` for subagents

`subagents/mcp/context.rs:37-44`:

```rust
let existing_subagent_count = service
    .list_subagents(&request.session_id)...
let create_block_reason = service
    .validate_parent_can_spawn(&request.session_id)
    .err()
    .map(|error| error.to_string());
```

Both are called unconditionally from `resolve_context`, which runs on every `tools/list` and `tools/call`. `validate_parent_can_spawn` internally does `list_subagents` + `get_workspace` + mutation-block check + a few other reads, so the real cost is closer to 5-6 DB hits per call.

For `tools/list`, this is the price of dynamic tool lists. Fine.

For `tools/call`, though, the results are used only inside `create_subagent` (to short-circuit via `!ctx.can_create`). Other tools (`list_subagents`, `read_subagent_events`, `send_subagent_message`, etc.) pay this cost for nothing. Consider computing `can_create` / `existing_subagent_count` lazily on `tools/list` only, or moving the check inside `create_subagent` where it's the existing `validate_parent_can_spawn(parent_session_id)?` call at :153.

Alternative framing: the context struct is now doing double duty as \"things needed to decide tools/list\" and \"things needed to render `get_subagent_launch_options`.\" Splitting into `SubagentMcpContext` (stable, cheap) + a `SubagentLaunchOptions::load(...)` lazy computation called only when needed would fix both concerns. Low-priority.

### 2. `validate_tool_for_role` and the dispatcher's match both know the role/tool mapping

`reviews/mcp/calls.rs:58-74` enumerates the allowed (role, tool) pairs up front. The subsequent `match (ctx.role, name)` at :16-55 also enumerates the same pairs with dispatch bodies. If someone adds a tool and forgets one branch, the code still compiles — the duplicate is silent.

Closest spec-aligned shape: return a typed enum from `validate_tool_for_role` (e.g., `ReviewToolInvocation::SubmitResult | GetStatus | MarkRevisionReady`) and dispatch off that instead of a stringly `name` re-match. Minor; not a blocker.

### 3. `ProductMcpDefinition` is still narrower than the spec

Unchanged since last round. Spec asks for `owner_domain, default_injection_policy, required_capabilities, binding_summary_kind` (`product-mcps.md:157-193`); struct has `acp_server_name, server_info_name, instructions, unauthorized_code, request_invalid_code` instead. The comment at `definition.rs:17-19` explicitly defers this.

Not a blocker for this PR, but worth tracking: this is the largest remaining divergence from the target spec shape. A follow-up that adds the missing fields (with sensible defaults so nothing breaks) and updates `product-mcps.md:157-193` to include the pragmatic additions like `acp_server_name` would close the gap on both sides.

### 4. `MUTATING_TOOL_NAMES` still in `tools.rs`, not `definition.rs`

Unchanged since last round. The membership test (#2 above) means drift is caught at test time, which is the biggest risk the previous comment was flagging. Moving the constant into `ProductMcpDefinition` remains cleaner but is no longer load-bearing.

### 5. Injection metadata still not on the registry

Unchanged since last round. `ProductMcpInjectionContext` still takes three concrete auth handles. This becomes four when cowork migrates; worth a follow-up.

### 6. `endpoint.rs` still missing from the shared kit

Unchanged. `dispatch_product_mcp` still lives in `api/http/product_mcp.rs` rather than `integrations/mcp/product_server/endpoint.rs` (spec target at `product-mcps.md:76-84`). Follow-up.

### 7. Six legacy handler functions still hand-written

Unchanged. The alias infrastructure is in place but unused for the legacy routes. Follow-up.

### 8. Path still `sessions/**`, not `domains/sessions/**`

Unchanged, explicit non-goal.

## Nits

- **`create_subagent` cleanup added** (`subagents/mcp/calls.rs:238`): when sending the initial prompt fails, the child session is now also torn down (previously only the wake schedule was). Caught by the PR, not by review — good defensive hygiene.
- **`selection.rs:84-86`** now pushes the workspace_naming binding summary. Was this always the intent? If yes, this was a latent bug the PR incidentally fixed. Worth a PR description note.

## Verdict

Ready to merge. All four previous-round blockers (dead code, `endpoint_operation_kind` coverage) resolved, plus a solid set of voluntary improvements around subagent tool visibility and review tool-role validation. The remaining items (#1 perf, #3–#7 spec alignment) are genuinely follow-up material — they don't change the semantics of this PR.

Good iteration. Five rounds converged to something that:
- matches the spec's file grammar for the three migrated products
- has one shared dispatcher, one registry, one auth helper, one injection path (modulo the three concrete auth handles still inside injection)
- acquires workspace leases per-tool rather than per-endpoint
- consistently soft-fails when context resolves to a non-applicable state, with that pattern now captured in `product-mcps.md`
- has membership tests preventing mutating-tools drift

Remaining work to get the rest of the way to spec is straightforward and can land in parallel with cowork migration.

anyharness/crates/anyharness-lib/src/sessions/subagents/mcp/tools.rs
```diff
@@ -37,41 +44,50 @@ pub(super) struct SendSubagentMessageArgs {
...
+                        "modelId": { "type": "string" },
+                        "modeId": { "type": "string" }
+                    },
+                    "required": ["prompt"]
+                }),
+            ),
+        );
+    }
+
+    if ctx.existing_subagent_count > 0 {
```
pablonyx: [P2] This makes the child-management tools disappear from tools/list whenever the parent currently has zero children. Fresh eligible parent sessions start in exactly that state, so the MCP client can bind only launch_options/list/create at startup; after create_subagent returns a child id, the agent may still have no registered schedule_subagent_wake, get_subagent_status, send_subagent_message, or read_subagent_events tools to use. The subagents docs treat those as part of the current MCP surface, and the prompt explicitly tells the agent to schedule/read after creating a child. Keep the child tools advertised for eligible parent sessions and let the call paths enforce child ownership/existence, or add an explicit tool-refresh path before relying on this dynamic list.

anyharness/crates/anyharness-lib/src/sessions/subagents/mcp/context.rs
```diff
@@ -0,0 +1,54 @@
...
+    if workspace.surface != "standard" {
+        return Err(ProductMcpContextError::conflict(
+            "subagents are only available in standard workspaces",
+        ));
+    }
+    let existing_subagent_count = service
+        .list_subagents(&request.session_id)
+        .map_err(|error| ProductMcpContextError::Internal(error.into()))?
+        .len();
+    let create_block_reason = service
```
pablonyx: [P3] This collapses every validate_parent_can_spawn error into a normal createBlockReason. That is right for expected policy denials like Disabled, DepthLimit, FanoutLimit, or MutationBlocked, but the same path also covers internal store/link/runtime failures surfaced as SubagentError::Internal. The product MCP spec update in this PR says expected-unavailable product state may reduce tools, while corrupt state or backend faults should be hard context errors. Split expected availability errors from internal failures here so tools/list does not hide backend faults as a user-facing canCreate=false reason.

PR #203, opened 2026-05-13
refactor(anyharness): migrate agents catalog ownership

pablonyx:
## Summary

- split the AnyHarness agents catalog domain out of the old monolithic `catalog.rs`
- remove the public runtime model/provider/session-launch catalog endpoints and SDK surfaces
- keep internal target-local launch projection for session, cowork, and subagent flows as resolved launch options
- move desktop pre-session launch/model projection to the cloud agents catalog endpoint
- gate chat and workspace-bootstrap launch choices by target-local readiness, with separate cloud-catalog vs target-readiness errors
- delete the split launch catalog, stale validation script, dead contract DTOs, old generated SDK types, and dead post-create model-availability migration residue

## Why

Runtime catalog truth is now explicitly split:

- cloud catalog: optimistic product UI before a session exists
- bundled AnyHarness catalog: target-local validation/readiness/install/runtime projection
- live ACP config: active-session truth after materialization

This removes the duplicated model/launch catalog paths and prevents runtime public APIs from becoming a second product catalog source.

## Review Follow-Ups

- Bundled AnyHarness catalog parsing/validation now fails hard instead of degrading to empty runtime metadata.
- Desktop now shows target-update/model-selection guidance for `SESSION_MODEL_UNSUPPORTED` and `SESSION_MODE_UNSUPPORTED` instead of a generic create failure.
- Desktop launch catalog state now exposes target-readiness failures separately from cloud catalog failures.
- Workspace bootstrap now intersects cloud launch choices with target readiness before opening the initial empty session.
- The old post-create model availability dialog/store/workflow and old launch-catalog/model-registry merge helper were removed as dead migration residue.
- The internal launch projection is now `ResolvedWorkspaceLaunchOptions` and no longer carries old public response shell fields.
- The agent catalog/readiness spec now scopes this PR to the catalog migration and documents the remaining install/readiness topology follow-up.
- Promoted the cheap agents-domain topology moves now: `registry/mod.rs`, `credentials/mod.rs`, `readiness/resolver.rs`, and `reconcile/{mod.rs,execution.rs}`.
- Cleaned stale AnyHarness docs and removed the dead per-model launch-control merge branch.
- Removed the final stale Amp env reference from the runtime smoke script and credentials-sync note.
- Documented `portability/mod.rs` in the required agents source shape and removed brittle line-number references from `agent-mode-matrix.md`.
- Removed the dead `DesktopLaunchModelRegistryModel.launchControls` type field now that per-model launch controls are retired.
- Retired the leftover Amp provider icon mapping after Amp was removed from catalog-facing support.

## API / SDK Impact

This intentionally removes the old public AnyHarness runtime catalog endpoints and their generated SDK surfaces. Pre-session product launch UI should use the cloud `/v1/catalogs/agents` projection; target runtime code still validates against the bundled catalog and active sessions still use live ACP config as truth.

Removed SDK/runtime catalog surfaces include:

- `ModelRegistriesClient`
- provider config client/hook surfaces
- `RuntimeClient.getEffectiveAgentLaunchCatalog`
- workspace session-launch client/hook/query-key surfaces
- generated model-registry, provider-config, and workspace session-launch catalog types

## Product-Visible Behavior

- Launchable static model lists are active-only. Candidate, deprecated, and hidden models are not shown in pre-session selectors.
- The catalog `model` control is not duplicated as a launch control; model selection flows through `agent.models` and `defaultModelId`.
- On cold start without cloud catalog availability, pre-session launch UI has no silent fallback to removed runtime catalog endpoints.

## Validation

- `cargo check --workspace`
- `cargo check -p anyharness-contract`
- `cargo test -p anyharness-lib domains::agents`
- `cargo test -p anyharness-lib api::openapi`
- `cargo test -p anyharness-lib domains::cowork`
- `cargo test -p anyharness-lib sessions::subagents`
- `cargo test -p anyharness-lib sessions::service::tests::validates_create_session_mode_ids_against_bundled_catalog`
- `pnpm --filter @anyharness/sdk generate`
- `pnpm --filter @anyharness/sdk test`
- `pnpm --filter @anyharness/sdk-react test`
- `pnpm --dir desktop exec tsc --noEmit`
- targeted desktop vitest suite: 14 files, 61 tests
- `pnpm --dir desktop exec vitest run src/hooks/chat/derived/use-chat-launch-catalog.test.tsx src/hooks/chat/derived/use-configured-launch-readiness.test.tsx src/lib/domain/sessions/creation/create-session-error.test.ts src/lib/domain/agents/target-ready-launch-agents.test.ts src/lib/domain/agents/cloud-launch-catalog.test.ts`
- `node scripts/validate-agent-catalog.mjs`
- `pnpm --dir desktop exec vitest run src/lib/domain/agents/cloud-launch-catalog.test.ts src/lib/domain/chat/models/launch-control-descriptors.test.ts`
- `node --check scripts/smoke-e2b-runtime.mjs`
- `git diff --check`


desktop/src/hooks/chat/derived/use-chat-launch-catalog.ts
```diff
@@ -36,72 +26,39 @@ export function useChatLaunchCatalog({
...
-  const modelRegistriesQuery = useModelRegistriesQuery({
-    enabled: canQueryRuntimeCatalog,
-  });
-  const catalogData = query.data ?? effectiveCatalogQuery.data ?? null;
-  const catalogLoading = query.isLoading || (!catalogData && effectiveCatalogQuery.isLoading);
+  const query = useCloudAgentCatalog(true);
+  const catalogData = query.data ?? null;
+  const catalogLoading = query.isLoading;
 
   const launchAgents = useMemo(
```
pablonyx, line 44: This needs to be intersected with target-local readiness before it drives defaults and selector groups. After this migration `catalogData.agents` is the cloud optimistic catalog, not the target-ready set; `hasAgents` still comes from `/v1/agents`, but `groups`, `defaultLaunchSelection`, and `selectedLaunchSelection` are built from every cloud agent. With one ready agent present, the selector can still expose a different not-ready agent and send it to session creation, where the runtime rejects it later. Please filter or annotate these launch agents with `useAgentCatalog()` readiness before building selectable groups/defaults, and add a test where cloud returns two agents but only one is target-ready.

desktop/src/hooks/chat/derived/use-configured-launch-readiness.ts
```diff
@@ -19,16 +15,22 @@ export function useConfiguredLaunchReadiness(
...
-    ),
-    [launchCatalog.launchAgents, preferences, providerConfigs],
+    () => hasCatalogLoadError
+      ? {
+        selection: null,
+        displayName: null,
+        reason: "Couldn't load the agent catalog. Retry once cloud is reachable.",
+        status: "unavailable" as const,
+      }
+      : resolveConfiguredLaunchSelection(
```
pablonyx, line 32: Same readiness gap affects composer enablement. `resolveConfiguredLaunchSelection` reports `ready` whenever the preferred kind/model exists in the cloud catalog, even if `agentsByKind` says the target agent is missing, installing, or login-required. `disabledReason` only checks runtime readiness when the resolution is already unavailable, so a not-ready configured agent can still produce `isReady: true`. Please gate ready resolution on `configuredAgent?.readiness === "ready"` or consume the readiness-filtered launch list, and cover this with a missing/not-ready agent test.

desktop/src/lib/domain/agents/cloud-launch-catalog.ts
```diff
@@ -0,0 +1,400 @@
...
+    defaultModeId: agent.session.defaultModeId ?? null,
+    dynamicModels: agent.session.dynamicModels,
+    modelDisplayPolicy: agent.session.modelDisplayPolicy ?? null,
+    promptCapabilities: agent.session.promptCapabilities ?? null,
+    models,
+    launchControls,
+  };
+}
+
+function isLaunchVisibleModel(model: CloudAgentCatalogModelInput): boolean {
```
pablonyx, line 275: Launchable model lists should be active-only here. The runtime fallback projection now filters with `model.status == Active`, and session creation validates against that active-only bundled registry. This hidden-only filter would let future `candidate` or `deprecated` cloud models appear in desktop selectors/default resolution even though the target runtime will reject them at create time. Please switch this to `model.status === "active"` and add a projection test covering active/candidate/deprecated/hidden.

desktop/src/lib/domain/agents/cloud-launch-catalog.ts
```diff
@@ -0,0 +1,400 @@
...
+    agent.session.defaultModelId,
+    visibleModels,
+  );
+  const defaultSessionControls = projectSessionDefaultControls(agent.session.controls);
+  const models = visibleModels.map((model) => ({
+    ...model,
+    isDefault: model.id === defaultModelId,
+    sessionDefaultControls: defaultSessionControls,
+  }));
+  const launchControls = agent.session.controls
```
pablonyx, line 254: Please keep the `model` catalog control out of `launchControls`. Models are already represented by `agent.models` and `defaultModelId`, and the old unified runtime projection explicitly returned `None` for `control.key == "model"`. As written, every agent gets a duplicate model launch control because the catalog model controls have `surfaces.start/session`; the new test even locks that in. This projection should filter `control.key !== "model"` here, and the remaining launch controls should be the inline non-model controls (`mode`, `effort`, etc.) that actually need control descriptors.

anyharness/crates/anyharness-lib/src/domains/agents/catalog/validation.rs
```diff
@@ -0,0 +1,286 @@
...
+        if !supported_agent_catalog_create_field(&control.key, create_field) {
+            anyhow::bail!(
+                "agent catalog agent '{}' control '{}' has unsupported createField '{}'",
+                agent_kind,
+                control.key,
+                create_field
+            );
+        }
+    }
+    if control.value_source == "inline" {
```
pablonyx, line 210: Since the bundled agent catalog is now the runtime-trusted source for descriptor and fallback launch/model metadata, this should reject unsupported `valueSource` values instead of only special-casing `inline`. A typo like `"inlinee"` currently passes validation with empty values and then projections silently produce missing choices/defaults. Please validate `valueSource` is one of `inline`, `agentModels`, or `discoveredModels`, and consider enforcing key/createField/valueSource consistency for `model` and `mode` controls while this code is being centralized.

anyharness/crates/anyharness-lib/src/domains/agents/model.rs
```diff
@@ -9,7 +9,6 @@ pub enum AgentKind {
     Gemini,
     Cursor,
     OpenCode,
-    Amp,
 }
```
pablonyx, line 12: Removing Amp from `AgentKind` also needs the remaining support-matrix references cleaned up. I still see `docs/anyharness/src/agents.md` saying credential discovery checks Amp, and `scripts/smoke-e2b-runtime.mjs` still includes `{ kind: "amp" }`; with this parser that smoke path is now an unknown agent. Please either update those references in this PR or keep Amp catalog/runtime support consistent.

pablonyx:
Reviewed the migration against the AnyHarness/frontend docs and current code paths. Direction looks right overall: split catalog ownership, remove public runtime catalog endpoints, keep internal launch options, regenerate SDK, and move desktop optimistic launch metadata to the cloud catalog. I left inline comments on the main issues I think should be fixed before this is ready: chat launch currently uses cloud catalog options without target readiness filtering, configured launch can report ready for a not-ready local agent, the cloud projection should keep launchable models active-only, the model control should not be projected as a launch control, catalog validation should reject unsupported valueSource values, and Amp removal left stale support references.\n\nChecks I ran locally: `cargo test -p anyharness-lib domains::agents --quiet`, `cargo test -p anyharness-lib sessions --quiet`, `cargo test -p anyharness-lib api::openapi --quiet`, `cargo check -p anyharness-lib --quiet`, `pnpm --filter @anyharness/sdk build`, `pnpm --filter @anyharness/sdk test`, `pnpm --filter @anyharness/sdk-react build`, `pnpm --filter @anyharness/sdk-react test`, `pnpm --dir desktop test -- src/lib/domain/agents/cloud-launch-catalog.test.ts src/hooks/chat/derived/use-chat-launch-catalog.test.tsx` (this ran the desktop suite), `node scripts/validate-agent-catalog.mjs`, `python3 scripts/check_frontend_boundaries.py`, `python3 scripts/check_anyharness_old_paths.py`, and `python3 scripts/check_max_lines.py`.

pablonyx:
## Review critique — PR #203

Verified against the spec at `docs/anyharness/specs/agent-catalog-readiness.md` (as updated on this branch) and the plan. Posting this here before review rather than as line comments so the structural issues are visible together.

### Blocking

**1. Silent removal of the `amp` agent kind — not in the plan or PR description.**
- `anyharness-lib/src/domains/agents/model.rs` on `main` declares `AgentKind::Amp` + `CredentialDiscoveryKind::Amp`; this PR deletes both.
- `docs/anyharness/src/agent-mode-matrix.md` is edited to drop the Amp section and change "All six" → "All five".
- `desktop/src/components/ui/provider-icons.tsx` still exports `AmpProviderIcon` (dead code if this is intentional).
- `desktop/src/lib/domain/cowork/session-mode-defaults.test.ts` still references `"amp"`.

If this is deliberate (catalog already dropped Amp on `main`), the PR description needs to call it out, the dead `AmpProviderIcon` should go, and the test reference should be reviewed. If it is unintended, revert.

**2. Scope does not match the spec's target source shape.**
The spec (lines 104–139) requires:
```
install/  { mod.rs, native.rs, agent_process.rs, launcher.rs, locks.rs }
registry/ mod.rs
credentials/ mod.rs
reconcile/ mod.rs
readiness/ { mod.rs, launch_options.rs, artifacts.rs, spawn.rs, compatibility.rs }
```
What landed:
```
install_lock.rs, installer.rs (1499 lines, still in oversized allowlist),
registry.rs, credentials.rs, reconcile.rs, reconcile_execution.rs, resolver.rs (744 lines) — all still flat at the agents/ root.
readiness/ only contains launch_options.rs. artifacts.rs, spawn.rs, compatibility.rs are missing, and resolver.rs (the actual readiness implementation) was not moved into readiness/.
```
Either the spec is wrong (and should be narrowed in this PR to only `catalog/` + `readiness/launch_options.rs`), or the PR is incomplete. As-is the repo ships with an authoritative spec whose "Target Source Shape" and "Migration Acceptance" sections contradict the code. Please pick one and make them consistent in the same PR — otherwise the next person hits drift immediately. CLAUDE.md: "Do not leave duplicate old and new code paths behind after a migration."

**3. `readiness/launch_options.rs` still models the old public-API response shape.**
`WorkspaceSessionLaunchCatalogData { workspace_id, catalog_version, agents }` was the body of the deleted `/v1/workspaces/{id}/session-launch`. Now that this is an internal projection used only by cowork MCP / subagent MCP / `SessionService::get_workspace_session_launch_catalog`, the fields `workspace_id` and `catalog_version` aren't carrying their weight (`workspace_id` is copied straight back from the caller; `catalog_version` uses a `"bundled-invalid"` sentinel on parse failure). Rename to something like `ResolvedLaunchOptions` and drop the shell fields, or document why the API shape is kept.

### Significant

**4. `"bundled-invalid"` magic string leaks into behavior.**
`bundled_catalog_version()` returns the literal string `"bundled-invalid"` when parsing fails (`projection/models.rs:30`). Same module's `bundled_model_registries()` returns `vec![]` on the same failure, and `bundled_create_mode_ids()` returns `None`. Net effect if the bundled catalog ever fails to parse:
- session creation fails with `"mode catalog not found for agent '...'"` rather than a clear "bundled catalog is invalid" error;
- `catalog_version` propagates a sentinel into anything that reads launch-options.
This is silent failure for a bundled, compile-time asset. Prefer a `panic!`/`unwrap` at init (it's `include_str!` + `validate`) or a typed error, not log + keep going.

**5. Model status filter changed from "active + optional candidates" to "active only".**
`model_is_selectable` (projection/models.rs:166) drops Candidate models unconditionally. Pre-migration `ModelCatalogService::with_options(..., allow_candidates)` let callers opt in. Confirmed by `hidden_candidate_and_too_new_models_are_not_selectable` test. This is a behavior change that should be in the PR description — today any internal flow that used to surface candidates (e.g. preview models) will silently lose them.

**6. Cloud-catalog / newer-model UX is not addressed.**
Spec says cloud catalog can be newer than a runtime; `SessionService::create_session` now rejects unknown models with `ModelUnsupported`. I could not find handling in `use-session-creation-actions.ts` for that specific error (the `SessionModelAvailabilityDialog` is still wired up via `use-session-model-availability-workflow`, which keyed off the live ACP config path). Please confirm that when the cloud catalog advertises a model the bundled runtime doesn't know about, the desktop surfaces a usable error instead of a generic create failure. If it's still routed through `SessionModelAvailabilityDialog`, add a test; if not, add explicit handling.

**7. Cold-start / cloud-unavailable path not documented.**
`useCloudAgentCatalog` uses `retry: 1, staleTime: 5m`. Workspace bootstrap priming of the runtime catalog was removed (`use-workspace-bootstrap-cache.ts` -53). On a fresh install with no cached query and a failing cloud call, the launch UI will be empty. Plan acknowledged this was acceptable; PR description should state it explicitly so it isn't treated as a regression.

**8. `SessionService::get_workspace_session_launch_catalog` still exists.**
The public HTTP endpoint is gone, but the service method and `SessionRuntime::workspace_session_launch_catalog` remain as internal entry points for cowork + subagent MCP. That's fine, but the method names still read like "get the catalog response". Consider inlining into the MCP call sites or renaming to `resolved_launch_options_for_workspace` — the current shape invites someone to re-expose it.

### Minor

**9. Release label.** `release:maintenance` is the label you picked; this PR removes public runtime endpoints and SDK clients/hooks/types. Any external `@anyharness/sdk` consumer at `0.1.0` will break. `docs/ci-cd/README.md` doesn't define `release:breaking`, so `release:maintenance` is probably correct, but call out the SDK breakage in the PR body so release notes are accurate.

**10. `mod.rs` discipline.** `catalog/mod.rs` is fine (pure `pub mod` declarations). `agents/mod.rs` remains a `pub mod` list too — good. CLAUDE.md rule against barrel re-exports is preserved.

**11. Test coverage for `cloud-launch-catalog.test.ts`.** Only exercises the OpenCode dynamic-models path. Missing coverage:
- A non-dynamic agent with an inline `model` control (Claude-shaped).
- A `valueSource: "discoveredModels"` control with non-empty `values` (your fallback branch at `resolveControlValues` is untested).
- `isLaunchVisibleModel` hiding the `defaultModelId` (verify `resolveDefaultModelId` fallback to first visible).
- `sessionDefaultControls` with multiple keys (`reasoning` + `effort`).

**12. On-every-call parsing.** `bundled_agent_catalog_document()` re-parses + re-validates the embedded JSON on every `create_session` / launch-options call. It's `include_str!` so it's not IO, but it is work. Consider `OnceLock`.

**13. Stale-symbol check passed (verified locally):** `ModelCatalogDocument | LaunchCatalogDocument | ModelCatalogService | LaunchCatalogService | ANYHARNESS_MODEL_CATALOG_URL | ANYHARNESS_LAUNCH_CATALOG_URL | catalogs/launch` → zero hits. Good.

### Summary

The catalog split + endpoint removal + desktop→cloud swap is solid work and the new `cloud-launch-catalog.ts` projection is clean. The two items that would block my review: (1) the Amp removal needs to be explicit or reverted, and (2) the spec and code disagree on whether `install/`, `credentials/`, `registry/`, `reconcile/`, and the rest of `readiness/` were migrated — pick one and make them match here.

pablonyx:
## Re-review — follow-up commit 9543d80d

Looked only at what changed since my last comment. Partial credit on the earlier feedback, a few new behavior changes that need flagging, and a couple of items still unaddressed.

### What the follow-up fixes

- **Control-shape validation** (`catalog/validation.rs`, `scripts/validate-agent-catalog.mjs`) now enforces `model → modelId/agentModels|discoveredModels` and `mode → modeId/inline`. Both Rust and JS are aligned. New tests cover `inline → model` and bogus `valueSource`. Good.
- **Target-readiness gate on cloud launch options** (`use-chat-launch-catalog.ts` + `use-configured-launch-readiness.ts`). Cloud catalog is now intersected with `useAgentCatalog().agentsByKind[kind].readiness === "ready"`. Tests added for both the launch-catalog filter and the configured-default readiness block. This closes the "cloud advertises models the runtime can't launch" gap I flagged earlier at the UI layer.
- **Amp removal finished in docs** (`integrations.md`, `src/agents.md`) and **removed from `smoke-e2b-runtime.mjs`**. The stray `"amp"` cowork test assertion is gone.

### New behavior changes introduced in this commit — not in the PR body

**a) `isLaunchVisibleModel` tightened from `!== "hidden"` to `=== "active"`.**
`cloud-launch-catalog.ts:275`. This drops `candidate` and `deprecated` models from launch options, which matches the Rust-side `model_is_selectable` rule I flagged earlier — so runtime + desktop are now consistent. But this is a product-visible change (deprecated models that were still being surfaced no longer show), and it's not called out. Please add a line to the PR body.

**b) `model` control is now stripped from `launchControls`.**
```ts
.filter((control) =>
  control.key !== "model"
  && (control.surfaces.start || control.surfaces.session)
)
```
The test expectation flipped from "projects a model launch control" to `.toBeUndefined()`. The model control now flows exclusively via `agent.models` + `defaultModelId`. This is the right invariant (model selection is one source of truth, not two), but the test change is just `.toBeUndefined()` with no assertion that model selection still works end-to-end in the selector. Please verify in the browser that the model picker, "more models" source, and per-model `launchControls` mergeing in `launch-control-descriptors.ts:50` all still work — that file reads `model?.launchControls` from `DesktopLaunchModelRegistryModel`, which is populated from `projectCloudModel`, which doesn't carry `launchControls`. If no model-level launch controls were ever emitted, fine; but the branch now quietly depends on that.

### Still unaddressed from the previous review

- **Spec ↔ code shape mismatch** (previously #2): `docs/anyharness/specs/agent-catalog-readiness.md` still lists `install/{native,agent_process,launcher,locks}.rs`, `registry/mod.rs`, `credentials/mod.rs`, `reconcile/mod.rs`, and `readiness/{artifacts,spawn,compatibility}.rs` under "Target Source Shape" and "Migration Acceptance". None of those have moved — `install_lock.rs`, `installer.rs`, `resolver.rs`, `reconcile.rs`, `reconcile_execution.rs`, `registry.rs`, `credentials.rs` are all still flat at the `agents/` root. The spec is authoritative for this area per CLAUDE.md, so either (a) narrow the spec to match what's migrated (catalog + readiness/launch_options), or (b) do the remaining moves. Either is fine — shipping with a contradiction is not.
- **`WorkspaceSessionLaunchCatalogData` still uses the old public-API response shape** as an internal type (previously #3). Not a blocker, but a rename (`ResolvedLaunchOptions`) would make the internal-only status obvious and stop a future caller from re-exposing it.
- **`"bundled-invalid"` sentinel on parse failure** (previously #4). `bundled_catalog_version()` still returns that literal string; `bundled_model_registries()` still returns `vec![]` and logs. For a compile-time embedded asset, a hard failure at process init would be clearer than silent degradation — the only way to trip this is if someone ships a bad `catalogs/agents/v1/catalog.json` and tests don't catch it.
- **`include_str!` JSON re-parsed on every call** (previously #12). Still true — `bundled_agent_catalog_document()` is called from hot paths (every `create_session`, every launch-options query). `OnceLock` would fix it.
- **`AmpProviderIcon` still in `desktop/src/components/ui/provider-icons.tsx`** (`AmpProviderIcon` at line 49, `amp: AmpProviderIcon` at line 75). Dead code now that `amp` is out of the catalog, registry, credential-discovery enum, and docs. Delete it in this PR.
- **SDK breakage** in the PR description (previously #9). Still just "remove … SDK surfaces" in the summary. Any external consumer of `@anyharness/sdk` that imported `ModelRegistriesClient`, `getEffectiveAgentLaunchCatalog`, the workspace session-launch client/hook, or the removed generated types will break. `release:maintenance` is the closest available label, but list the removed exports explicitly in the PR body so the release notes tell the truth.

### Nits

- The "configured agent missing" copy is `"${agentKind} is not ready yet."` — that's the raw kind (`opencode`, `claude`), not the display name, because `configuredAgent` is `null` in this branch. It's lowercase and bare; the ready-but-login-required branch uses the display name. Minor, but if a user sees this message, "opencode is not ready yet." reads like a bug.
- `orderLaunchAgents` readiness-gate change rebuilds the `snapshot.agents` from the filtered list. Good — but any downstream consumer that was relying on "cloud catalog contents" rather than "locally launchable" will now see a different set. Verify: automation creation, review setup, agent-defaults pane — all of which use `useCloudLaunchModelRegistries`, not `useChatLaunchCatalog`. Those call sites were NOT changed in this commit, so they still see the full cloud set. That's probably correct for settings/automation (optimistic), but worth confirming it matches the product intent.

### Verdict

Two of my three blocking items are resolved (Amp cleanup + cloud/runtime mismatch UX). The spec ↔ code shape contradiction is the one I would still gate on — the repo shouldn't ship with an authoritative doc that lists a target shape the code doesn't match.

pablonyx:
Updated re-review on head `9543d80`. The previous inline findings I left are addressed in this update: desktop launch options now intersect cloud catalog data with target readiness, configured launch readiness is gated on target readiness, model projection is active-only, `model` controls are no longer duplicated as launch controls, catalog `valueSource` validation is stricter, and the stale Amp support-matrix references were cleaned up.

Remaining concerns:

1. [P2] The migration spec still says the final agents domain has nested `registry/`, `credentials/`, `install/`, and `reconcile/` modules plus split readiness files (`docs/anyharness/specs/agent-catalog-readiness.md:104`), but the actual domain still exposes flat modules from `anyharness/crates/anyharness-lib/src/domains/agents/mod.rs:1` (`credentials.rs`, `installer.rs`, `install_lock.rs`, `registry.rs`, `reconcile.rs`, `reconcile_execution.rs`, `resolver.rs`). `SessionService` also still exposes `get_workspace_session_launch_catalog` at `anyharness/crates/anyharness-lib/src/sessions/service.rs:474`. Since the repo docs are authoritative and this PR claims the catalog/domain migration, either finish the source-shape move or update the spec/PR scope to mark that restructuring as explicit follow-up.

2. [P2] The bundled catalog is now the only runtime catalog source, but invalid bundled catalog reads still degrade to empty runtime metadata. `bundled_agent_descriptors()` returns `vec![]` on validation/parse error, `bundled_model_registries()` does the same, and `bundled_catalog_version()` returns `"bundled-invalid"` (`anyharness/crates/anyharness-lib/src/domains/agents/catalog/projection/descriptors.rs:16`, `anyharness/crates/anyharness-lib/src/domains/agents/catalog/projection/models.rs:13`, `anyharness/crates/anyharness-lib/src/domains/agents/catalog/projection/models.rs:25`). With remote refresh/cache removed, that failure mode turns a bad embedded catalog into "no agents" or later unsupported-model errors instead of a clear startup/runtime fault. I would prefer making the bundled catalog a fallible dependency at the call boundary or failing loudly during runtime initialization.

3. [P2] Desktop still treats the runtime `SESSION_MODEL_UNSUPPORTED` path as a generic create-session failure (`desktop/src/hooks/sessions/use-session-creation-actions.ts:603`). This migration intentionally lets the cloud catalog be ahead of a target runtime and relies on runtime create validation as the hard gate, so that error is expected in cloud/runtime skew. The UI should map that specific problem code to model-selection remediation, refresh/update guidance, or an immediate selection reset instead of only rolling back the optimistic session and showing `Failed to start chat session: ...`.

4. [P3] `useChatLaunchCatalog()` now depends on `useAgentCatalog()` readiness, but it only returns `error: query.error` from the cloud catalog query (`desktop/src/hooks/chat/derived/use-chat-launch-catalog.ts:35` and `:89`). If `/v1/agents` fails while the cloud catalog succeeds, `launchAgents` becomes empty because the readiness map is empty, but consumers see no target-readiness error. That can present as an empty catalog or missing configured agent rather than "target readiness failed to load." Consider propagating the target readiness error/status separately from the cloud catalog error.

pablonyx:
## Third review — no new commits, but deeper code pass

HEAD is still `9543d80d`; PR body unchanged. I re-read the new files in full rather than just diffs. Flagging a few things I didn't catch before, plus confirming which previous items are still outstanding.

### New findings

**1. Readiness gate is inconsistent across catalog consumers.**
`use-chat-launch-catalog.ts` now filters the cloud catalog by `agentsByKind.get(kind)?.readiness === "ready"`. Good. But the other callers of the same cloud catalog do not:

- `useCloudLaunchModelRegistries()` (home model selection, automation model selection, settings `AgentDefaultsPane`, `ReviewSettingsPane`, cowork thread workflow) returns the full cloud set with no readiness intersection.
- `useCloudAgentCatalog()` direct callers (home mode selection, automation mode selection) also ungated.
- `useWorkspaceBootstrapActions` uses its own local `orderLaunchAgents` (line 696) that only filters by `models.length > 0` — no readiness check. Then passes the first agent into `createEmptySessionWithResolvedConfig`, which will fail at the runtime if the agent isn't ready.

Either (a) move the readiness intersection into `useCloudLaunchModelRegistries` so every caller inherits it, or (b) document explicitly in each callsite why the full cloud set is correct there (settings probably should be "optimistic" — show agents the user could configure even if not yet installed — but automation/cowork/home-launch probably shouldn't). Today the behavior differs per surface with no stated rule.

**2. Session creation silently degrades live-defaults on cloud-catalog failure.**
`use-session-creation-actions.ts:451`:
```ts
const cloudLaunchCatalog = await ensureCloudAgentCatalog().catch(() => null);
const modelRegistries = buildDesktopLaunchModelRegistries(cloudLaunchCatalog?.agents ?? []);
```
If the cloud catalog call fails, `modelRegistries` is `[]` and `applySessionLaunchDefaults` returns early with no live defaults applied (`session-launch-defaults.ts:70`). The session still starts but the user's configured defaults (reasoning/effort/fast_mode/collaboration_mode) get silently dropped. No log, no toast. Similar pattern at `use-workspace-bootstrap-actions.ts:335` where the catastrophic-fallback is `catch(() => null)` with only a latency log. Consider at minimum logging the failure, and if live defaults matter, queueing them for a later retry.

**3. `projectCloudAgent` drops agents with no active models silently.**
`cloud-launch-catalog.ts:219`: `.filter((agent) => agent.models.length > 0)`. Combined with the new `isLaunchVisibleModel === "active"`, an agent whose only models are `candidate`/`deprecated`/`hidden` in the cloud catalog will vanish entirely from the desktop view — not even listed as "not ready". If such an agent is the user's `defaultChatAgentKind`, the readiness hook will branch to `isConfiguredAgentMissing` and show `"${kind} is not ready yet."` regardless of the real reason (no models is not the same as "not ready"). Either keep the agent with empty models and a clearer UI state, or plumb the reason through the hook.

**4. Trust-boundary claim in the spec isn't enforced anywhere.**
The spec at lines 88–102 says "The projection that produces `AgentDescriptor` must be sourced from trusted catalog data only." Current code does achieve this because `bundled_agent_catalog_document()` is the only call into `projection/descriptors.rs` and there's no remote catalog path. But there's no compile-time enforcement of that trust boundary — someone adding a remote-catalog path later can wire it straight into `agent_catalog_to_descriptors(&remote_catalog)` with no friction. A simple signal: make `agent_catalog_to_descriptors` private to the module and only expose `bundled_agent_descriptors()`. Same story for `agent_catalog_to_model_registries` (used only by the bundled path).

**5. `resolveControlValues` fallback never materializes for the `model` control.**
`cloud-launch-catalog.ts:342` has a fallback that populates empty `agentModels`/`discoveredModels` controls from `models`. Now that the `model` control is stripped upstream (line 257), the only remaining consumers of `valueSource` are non-model controls — which are inline — so the fallback branch is effectively dead. Either delete it or drop the `model` strip and let the fallback do its job. Current state has both the strip and the dead fallback, which is confusing to read.

**6. `AmpProviderIcon` still present** (`desktop/src/components/ui/provider-icons.tsx:49, 75`). I flagged this twice. The kind is gone from the catalog, model enum, registry, credential-discovery enum, schema validation, agent-mode-matrix doc, agents.md, smoke script, and the cowork test — but the icon still ships. It's reachable via the `amp: AmpProviderIcon` map, so any stale persisted preference or accidental string reference would render it. Delete it.

### Still outstanding from prior reviews

- **Spec ↔ code mismatch.** `docs/anyharness/specs/agent-catalog-readiness.md` "Target Source Shape" (lines 104–139) still lists `install/{native,agent_process,launcher,locks}.rs`, `registry/mod.rs`, `credentials/mod.rs`, `reconcile/mod.rs`, `readiness/{artifacts,spawn,compatibility}.rs`. None of those exist on disk. "Migration Acceptance" (lines 389–410) calls for "install, credential detection, readiness, reconcile, and seed behavior live outside `catalog/**`" — true for `catalog/**` but still flat at `agents/` root. Either narrow the spec to what actually migrated (catalog + `readiness/launch_options.rs`) or do the moves. This is the single highest-leverage blocker — CLAUDE.md says area docs are authoritative.
- **SDK-breaking removals not in PR body.** Removed exports: `ModelRegistriesClient`, provider config client/hook, `RuntimeClient.getEffectiveAgentLaunchCatalog`, workspace session-launch client/hook/query key, generated types for model registries/provider configs/session-launch catalog. List them explicitly in the PR description so release notes reflect reality.
- **`"bundled-invalid"` sentinel** on parse failure of `include_str!` asset in `projection/models.rs:30`. Still silent failure for a compile-time asset.
- **Repeated parsing of the embedded catalog.** `bundled_agent_catalog_document()` re-parses + re-validates on every call. Hot paths: `bundled_agent_descriptors`, `bundled_model_registries`, `bundled_create_mode_ids`, `bundled_catalog_version` — hit on every `create_session`, every launch-options lookup, every MCP `get_*_launch_options`. Use `OnceLock`.
- **`WorkspaceSessionLaunchCatalogData`** still uses the old public-API response shape for an internal-only type.

### Minor

- `use-configured-launch-readiness.ts:57` — `"${preferences.defaultChatAgentKind} is not ready yet."` interpolates the raw kind string. If a cloud catalog drift puts an agent in state "agent not in target runtime", the user sees `"opencode is not ready yet."`. That copy is technically true but indistinguishable from "login required". Consider a distinct string like `"${kind} isn't supported by this runtime yet."` for the `isConfiguredAgentMissing` branch vs. `"${displayName} is login required."` for the `isConfiguredAgentNotReady` branch — they're different failure modes.
- `buildDesktopLaunchModelRegistries` doesn't filter by readiness and doesn't carry readiness state. Any caller that uses it standalone (settings panes, automation) treats all agents as equal.
- `orderLaunchAgents` is defined twice (`use-workspace-bootstrap-actions.ts:696` and `use-chat-launch-catalog.ts:100`), with different filter logic (readiness-aware vs. not). Extract a shared helper in `lib/domain/agents/` so both paths evolve together.

### Verdict

Validation layer is solid. The readiness gate is good where it exists, but it's only applied to one of many consumers of the same cloud catalog — that inconsistency is the load-bearing problem right now, ahead of the still-outstanding spec/docs cleanup. Bootstrap's "cold start opens a session" path (#1, #2) is where a user will feel this first.

pablonyx:
Additional code pass on the same head `9543d80`; not repeating the previous review items.

1. [P2] Workspace bootstrap still resolves the initial empty session from the unfiltered cloud catalog. In `desktop/src/hooks/workspaces/use-workspace-bootstrap-actions.ts:334-383`, the bootstrap path calls `ensureCloudAgentCatalog()`, then `orderLaunchAgents(launchCatalog?.agents ?? [])`, and that local `orderLaunchAgents` only filters `agent.models.length > 0` at `:696-699`. Unlike `useChatLaunchCatalog()`, this does not intersect with `/v1/agents` target readiness. On a fresh workspace with no visible sessions, a cloud-default or preference-default agent that is missing/installing/login-required locally can still be selected and sent to `createEmptySessionWithResolvedConfig`, where the runtime rejects it. That turns ordinary workspace selection/bootstrap into a failed session-create path. Please reuse the readiness-filtered launch derivation or fetch/intersect target agents in this bootstrap path, and add a test where cloud returns two agents but only one is locally ready.

2. [P3] The old model-availability interruption flow is now orphaned. `hasImmediateLaunchModelMismatch()` / `buildPausedModelAvailability()` in `desktop/src/lib/domain/sessions/creation/model-availability.ts`, `requestSessionModelAvailabilityDecision()` in `desktop/src/hooks/sessions/workflows/use-session-model-availability-workflow.ts`, and the globally mounted `SessionModelAvailabilityDialog` in `desktop/src/App.tsx` no longer have a production caller after the immediate post-create mismatch check was removed. Either wire this flow to the new `SESSION_MODEL_UNSUPPORTED` / `SESSION_MODE_UNSUPPORTED` create-validation errors, or delete the store/workflow/dialog/domain tests as dead migration residue.

3. [P3] `mergeLaunchAgentsWithRegistries()` in `desktop/src/lib/domain/chat/launch/session-config.ts:124` also appears to be production-dead now; `rg` only finds the function and its tests. That helper existed for the old runtime launch catalog + model registry merge path, which this PR removes. Keeping it makes it look like the old merge model is still supported.

pablonyx:
## Fourth review — commits `3e183dba` + `092dc842`

I compared against every prior point. This pass closes almost everything I flagged. Summarizing what landed, what's still meaningful, and what small structural gaps remain relative to `docs/anyharness/specs/agent-catalog-readiness.md`.

### Resolved this round

- **Spec ↔ code mismatch.** Spec now has a "Catalog Migration Source Shape" section that explicitly lists `credentials.rs`, `installer.rs`, `install_lock.rs`, `registry.rs`, `resolver.rs`, `reconcile.rs`, `reconcile_execution.rs` as *transitional* with a separate "follow-up topology" section for the eventual folder structure. "Migration Acceptance" is scoped to the catalog migration, and follow-up promotion is called out. This is exactly the right move.
- **`"bundled-invalid"` sentinel gone.** `bundled.rs` now `OnceLock`-caches a parsed+validated document and `.expect()`s on parse/validate failure. `bundled_catalog_version()` is deleted outright. Hot-path re-parsing is gone.
- **Trust boundary now visible in code.** `agent_catalog_to_descriptors` and `agent_catalog_to_model_registries` are private to their modules; only `bundled_*` wrappers are `pub`. A future remote-catalog path cannot quietly route untrusted data into descriptor projection.
- **`WorkspaceSessionLaunchCatalogData` renamed** to `ResolvedWorkspaceLaunchOptions` with the dead `workspace_id`/`catalog_version` fields stripped. Call-site method renames propagated through `SessionService`, `SessionRuntime`, `CoworkRuntime`, subagent + cowork MCP tools. Clean.
- **Dead `resolveControlValues` fallback removed** from `cloud-launch-catalog.ts` — now that `model` controls are stripped upstream there's no caller for the `agentModels`/`discoveredModels` empty-values branch. Good housekeeping.
- **`AmpProviderIcon` and `amp:` mapping removed**. Grep confirms zero real Amp references remain in TS/Rust.
- **`SESSION_MODEL_UNSUPPORTED` / `SESSION_MODE_UNSUPPORTED` UX.** New `create-session-error.ts` maps these runtime codes to user-facing strings ("This target does not support the selected model/mode yet. Update AnyHarness on the target or choose another."). Toast and thrown error both go through it. `SessionModelAvailabilityDialog` + store + workflow hook + all `isSessionModelAvailability*` call sites deleted as dead migration residue. This is a cleaner model than the prior dialog-based flow.
- **Bootstrap readiness gate.** `useWorkspaceBootstrapActions` now reads `useAgentCatalog().agentsByKind` and intersects before picking the default launch. Shared helper `filterTargetReadyLaunchAgents` extracted to `lib/domain/agents/` and reused from both the chat hook and bootstrap. No more duplicate local implementations.
- **Cloud / target error paths split.** `useChatLaunchCatalog` now exposes `cloudCatalogError` vs `targetReadinessError` separately, and `useConfiguredLaunchReadiness` surfaces a different message per side. `isConfiguredAgentMissing` now uses `"... isn't supported by this runtime yet."` — the specific-kind-unknown case is distinguishable from "login required" now, which was my earlier nit.
- **SDK-breaking removals enumerated in PR body** under "Review Follow-Ups" section.

### Still worth addressing

**1. Ungated consumers of the same cloud catalog.**
My prior call-out — settings/automation/home-next/cowork/review use `useCloudLaunchModelRegistries` or raw `useCloudAgentCatalog` without the readiness gate — is technically still true. For most of these (`AgentDefaultsPane`, `ReviewSettingsPane`, `use-automation-*`) "optimistic, show-everything" is probably the intended product behavior (you set defaults for agents you haven't installed yet). But **`use-cowork-thread-workflow.ts:77`** and **`use-home-next-launch`** downstream chain are more ambiguous — the user is about to *launch* a session, not configure a preference. If the product intent is "settings = full cloud catalog, launch paths = readiness-gated", that rule belongs in a single place (maybe a `useCloudLaunchModelRegistries({ target: "ready" | "all" })` overload) rather than left to each caller to remember. Not a blocker; a follow-up issue would be fine.

**2. Dead `model?.launchControls` branch in `launch-control-descriptors.ts:50`.**
`mergeLaunchControls(agent.launchControls ?? [], model?.launchControls ?? [])` still reads per-model `launchControls`, but `projectCloudModel` never sets that field — the cloud schema doesn't carry it. The `mergeLaunchControls` signature allows it, but in practice the second argument is always `[]`. Either delete the merge path and simplify to `agent.launchControls.flatMap(launchControlToDescriptor)`, or document that per-model controls are a future capability. This is the last dead bit of the pre-migration code I can see.

**3. `useCloudAgentCatalog` error UX is implicit-only.**
`retry: 1, staleTime: 5m` and silent `catch(() => null)` in `use-workspace-bootstrap-actions.ts:335` and `use-session-creation-actions.ts:451`. The second one now at least routes `SESSION_MODEL_UNSUPPORTED` errors through a toast, but a cloud catalog outage at session creation still degrades to empty `modelRegistries` → no live defaults applied, with only a latency log. Consider at minimum a `tracing`-equivalent error log on the catch arm so ops notice. Small.

**4. `useChatLaunchCatalog` / `useConfiguredLaunchReadiness` return shape drift.**
`useChatLaunchCatalog` now returns `{ error, cloudCatalogError, targetReadinessError, ... }`. `useConfiguredLaunchReadiness` also fans out its own `launchReadinessErrorReason`. A small interface type (or doc comment on `useChatLaunchCatalog`) explaining that `error === cloudCatalogError ?? targetReadinessError` would help readers. Not a blocker.

**5. `AGENT_READINESS_LABELS` for the "not ready" branch.**
`use-configured-launch-readiness.ts:63` — `AGENT_READINESS_LABELS[configuredAgent.readiness].toLowerCase()` assumes the labels map covers every possible `readiness` string. If a new readiness state is added in `packages/services` without updating the labels map, this will produce `undefined.toLowerCase()`. Since `AGENT_READINESS_LABELS` is a typed record, TS will catch it at compile time, but worth flagging that the coupling is there.

### Structural completeness vs. the spec

The spec's "Catalog Migration Source Shape" matches what's on disk now. "Follow-up topology" (install/, credentials/, registry/, reconcile/, readiness/{artifacts,spawn,compatibility}, portability/) remains explicitly out of scope. `Migration Acceptance` items all hold:

- `domains/agents/catalog.rs` gone ✓
- `domains/agents/catalog/**` split into focused submodules ✓
- Legacy split launch catalog gone ✓
- Old structs/functions/env vars gone ✓
- Launch/model metadata projected from `AgentCatalogDocument` ✓
- Executable/process/auth descriptor projection trusted-only ✓ (now enforced by module privacy)
- `install`/`credentials`/`readiness`/`reconcile`/`seed`/`portability` behavior outside `catalog/**` ✓
- Tests split by responsibility ✓

### Verdict

Ready for review. Outstanding items are all follow-up-sized, not blocking. The biggest remaining improvement is consolidating the readiness-gating policy so every launch surface inherits it consistently — that can be a separate PR. Nice landing.

pablonyx:
Re-review on `092dc842e00329e572467487bb30c849849977d4`.

The previous runtime/desktop concerns I raised look addressed: bundled catalog loading now fails fast, create-session errors surface unsupported model/mode cases, target readiness failures propagate into launch readiness, and the old post-create availability path/dead merge helper code is gone. I ran:

- `node scripts/validate-agent-catalog.mjs`
- `cargo test -p anyharness-lib domains::agents --quiet`
- `python3 scripts/check_max_lines.py && python3 scripts/check_anyharness_old_paths.py`
- `git diff --check origin/main...HEAD`
- `pnpm --dir desktop test -- src/lib/domain/agents/target-ready-launch-agents.test.ts src/lib/domain/sessions/creation/create-session-error.test.ts src/hooks/chat/derived/use-chat-launch-catalog.test.tsx src/hooks/chat/derived/use-configured-launch-readiness.test.tsx` (this expanded to the full desktop suite: 325 files / 1679 tests)

Remaining critiques, focused on getting closer to the documented AnyHarness/domain shape:

1. [P2] The AnyHarness docs still describe deleted or obsolete shapes. `docs/anyharness/src/agents.md:96-109` still says there are two static sources, `registry.rs` and `catalog.rs`, and points at deleted `domains/agents/catalog.rs`. `docs/anyharness/src/agent-mode-matrix.md:41-45` still points at `desktop/src/config/session-control-presentations.ts`, which no longer exists. `docs/anyharness/README.md:244-247` calls the spec the “fully migrated agents domain” while the spec now says it is the current migration with transitional files. Since `docs/README.md` treats docs as authoritative, update these references as part of this migration rather than leaving the next change to rediscover the new shape.

2. [P2] `readiness/` is still only a thin launch-options wrapper while readiness ownership remains at root. `readiness/mod.rs` only exports `launch_options`, and `readiness/launch_options.rs:5-6` imports `registry::built_in_registry` plus root `resolver::resolve_agent`; the actual `ResolvedAgent` / `ResolvedArtifact` / `SpawnSpec` logic remains in `domains/agents/resolver.rs` (744 lines and still max-line allowlisted). The spec says `readiness/` owns target-local readiness and those outputs. If this PR is trying to get close to the docs, I would move `resolver.rs` under `readiness/` now, or split the low-risk public surface there and leave finer `artifacts` / `spawn` / `compatibility` files for follow-up.

3. [P3] Some transitional files are cheap to promote now and would materially reduce topology debt without touching behavior. `registry.rs` is only 10 lines and can become `registry/mod.rs`; `credentials.rs` can become `credentials/mod.rs`; `reconcile.rs` + `reconcile_execution.rs` can become `reconcile/mod.rs` + `reconcile/execution.rs` with import updates. `installer.rs` is larger and can remain an explicit follow-up, but the small moves would make `agents/mod.rs` much closer to the documented folder-oriented domain boundary.

pablonyx:
## Fifth review — commit `5e0ab34c` "tighten agents catalog topology"

Clean, minimal, scoped commit. Promotes the straightforward ownership folders and keeps only the intentionally-transitional files flat.

### Changes landed

**Topology promotion (pure renames, 97–100% similarity on each):**
- `credentials.rs` → `credentials/mod.rs`
- `resolver.rs` → `readiness/resolver.rs`
- `registry.rs` → `registry/mod.rs`
- `reconcile.rs` → `reconcile/mod.rs`
- `reconcile_execution.rs` → `reconcile/execution.rs`

**Import sites updated** in `app/mod.rs`, `api/http/agents.rs`, `installer.rs`, `sessions/runtime/startup.rs`, `sessions/service.rs`, plus the cascading `model.rs` doc-comment fix.

**Spec kept in lockstep.** "Catalog Migration Source Shape" updated to match disk: `readiness/{mod,launch_options,resolver}.rs`, `credentials/mod.rs`, `registry/mod.rs`, `reconcile/{mod,execution}.rs` are required; `installer.rs` and `install_lock.rs` are called out as the remaining transitional files. Per-section "Or `foo.rs`" alternatives are gone. Migration Acceptance updated: remaining topology work is scoped to `installer.rs`/`install_lock.rs` promotion and splitting readiness internals (`artifacts.rs`/`spawn.rs`/`compatibility.rs`) out of `readiness/resolver.rs`.

**Dead code removed:** `mergeLaunchControls` and the `model?.launchControls` merge path in `launch-control-descriptors.ts` — this was one of the nits in my last review. Per-model launch control concept is now fully retired at the consumer. Tests updated to match.

**Doc paths refreshed.** `agent-mode-matrix.md` was pointing at stale `desktop/src/config/session-control-presentations.ts` (file lives at `desktop/src/lib/domain/chat/session-controls/presentation.ts`) and at `domains/agents/registry.rs` (now `registry/mod.rs`). All fixed. `docs/notes/agent-credentials-sync.md` and `docs/anyharness/src/agents.md` updated similarly.

### Remaining nits (all minor)

**1. `portability/mod.rs` is on disk but not in the "Catalog Migration Source Shape" section.**
It's only listed in "Follow-up topology". Either add `portability/` to the required shape for completeness, or explicitly note it as transitional (it's already folder-shaped so there's nothing to promote — just document it). The acceptance section already mentions portability behavior must live outside `catalog/**`, which is satisfied.

**2. Per-model `launchControls?` field is now truly dead on the type.**
`DesktopLaunchModelRegistryModel.launchControls?` in `cloud-launch-catalog.ts:81`. No caller reads it, no projector writes it. Safe to delete in a follow-up. Minor.

**3. The other two outstanding follow-ups from my prior review remain follow-ups:**
- Gate consistency across cloud-catalog consumers (`useCloudLaunchModelRegistries` in cowork/home-next/settings is still ungated). Still reasonable as a separate PR.
- Silent `catch(() => null)` in `use-workspace-bootstrap-actions.ts:335` and `use-session-creation-actions.ts:451`. Still reasonable as a separate PR.

### Verdict

Ship it. This is the cleanest commit in the series — pure topology moves with import-site updates, spec kept honest, and two real cleanups (dead `mergeLaunchControls`, stale doc paths) folded in along the way. The remaining items are genuinely follow-up-sized.

pablonyx:
Re-review on `5e0ab34ce5745fdc8e0aa76f3523236fbc31c345`.

The last docs/topology concerns are addressed in code now: `registry/`, `credentials/`, `reconcile/`, and `readiness/resolver.rs` exist; old root `resolver`/`reconcile_execution` exports are gone; the spec and `src/agents.md` now describe the current shape instead of the deleted `catalog.rs` split. The dead per-model launch-control merge path is also gone.

Checks I ran locally:

- `node scripts/validate-agent-catalog.mjs`
- `cargo test -p anyharness-lib domains::agents --quiet`
- `cargo test -p anyharness-lib sessions --quiet`
- `cargo test -p anyharness-lib api::openapi --quiet`
- `cargo check -p anyharness-lib --quiet`
- `python3 scripts/check_max_lines.py && python3 scripts/check_anyharness_old_paths.py`
- `git diff --check origin/main...HEAD`
- `pnpm --filter @anyharness/sdk build && pnpm --filter @anyharness/sdk-react build`
- `pnpm --dir desktop test -- src/lib/domain/chat/models/launch-control-descriptors.test.ts src/lib/domain/agents/target-ready-launch-agents.test.ts src/lib/domain/sessions/creation/create-session-error.test.ts src/hooks/chat/derived/use-chat-launch-catalog.test.tsx src/hooks/chat/derived/use-configured-launch-readiness.test.tsx` (expanded to the full desktop suite: 325 files / 1679 tests)

Only remaining nit I see: `docs/anyharness/src/agent-mode-matrix.md` still has several stale line/path references after the path updates. Examples: `desktop/src/hooks/cowork/use-cowork-thread-workflow.ts:94` should point at `desktop/src/hooks/cowork/workflows/use-cowork-thread-workflow.ts`, the presentation-table ranges (`presentation.ts:34-78`, `:79-126`, `:127-163`) no longer match the current file, and `session_config.rs:148-156` now points before `SetSessionConfigOptionRequest` at line 165. Non-blocking, but if this doc is meant to stay authoritative, I would either refresh the exact line refs or remove them where the symbols/paths are enough.

No new blocking critique from this pass.

PR #204, opened 2026-05-13
refactor(anyharness): split live session actor

pablonyx:
## Summary

- Split the live AnyHarness session actor out of the old `acp/session_actor.rs` god module into `live/sessions/actor/**`, `live/sessions/connection/**`, and `live/sessions/handle.rs`.
- Moved latency helpers from API transport into `observability/latency.rs` so live/runtime code no longer imports `api/http` latency helpers.
- Updated AnyHarness docs and max-lines ratchets to reflect the new actor shape and remove stale `acp/session_actor.rs` debt.

## Notes

- This is intended to be behavior-preserving. `AcpManager`, `RuntimeClient`, `SessionEventSink`, `InteractionBroker`, `BackgroundWorkRegistry`, and `replay_actor` remain in their transitional `acp/**` locations for later topology work.
- Added the existing oversized desktop hook to the max-lines allowlist because it already exceeds the repo-wide limit on `origin/main`; no new AnyHarness actor files are allowlisted.

## Verification

- `cargo test -p anyharness-lib --no-run`
- `cargo test -p anyharness-lib`
- `cargo build -p anyharness-lib`
- `git diff --check`
- `rg "acp::session_actor|super::session_actor|pub mod session_actor" anyharness/crates/anyharness-lib/src`
- `rg "crate::api::http::latency|super::latency" anyharness/crates/anyharness-lib/src/{acp,sessions,domains,live}`


scripts/max_lines_allowlist.txt
```diff
@@ -33,6 +32,7 @@ anyharness/sdk-react/src/hooks/sessions.ts
 desktop/src/components/ui/icons.tsx
 desktop/src/hooks/sessions/use-session-creation-actions.ts
 desktop/src/hooks/sessions/use-session-runtime-actions.ts
+desktop/src/hooks/workspaces/tabs/use-workspace-shell-activation.ts
```
pablonyx: This PR should not add an unrelated desktop file to the max-lines allowlist. The migration removes the `acp/session_actor.rs` debt, but adding `desktop/src/hooks/workspaces/tabs/use-workspace-shell-activation.ts` expands unrelated debt and violates the plan's final-gate rule to surface pre-existing non-AnyHarness debt rather than expanding allowlists in this PR. Please drop this allowlist addition and either leave the unrelated max-lines failure as a surfaced blocker or handle that file in a separate PR.

anyharness/crates/anyharness-lib/src/live/sessions/connection/types.rs
```diff
@@ -0,0 +1,199 @@
...
+    pub(in crate::live::sessions) fn as_str(self) -> &'static str {
+        match self {
+            Self::CreatedFresh => "created_fresh_native",
+            Self::LoadedExisting => "loaded_existing_native",
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub(in crate::live::sessions) struct SessionStartupState {
```
pablonyx: `SessionStartupState` is mutable actor/session state, not a connection type. It is updated after startup by config application and ACP notifications and read by turn/config/notification handlers, so keeping it under `connection/types.rs` makes `connection/**` own live actor config state. That blurs the boundary this migration is trying to establish. Please move this to `actor/state.rs` or the actor config state module and keep `connection/**` focused on process and raw native-session mechanics.

anyharness/crates/anyharness-lib/src/live/sessions/actor/turn/active.rs
```diff
@@ -0,0 +1,529 @@
...
+        let latency_fields = latency_trace_fields(current_latency.as_ref());
+        tracing::info!(
+            session_id = %session_id,
+            flow_id = latency_fields.flow_id,
+            flow_kind = latency_fields.flow_kind,
+            flow_source = latency_fields.flow_source,
+            prompt_id = latency_fields.prompt_id,
+            "[workspace-latency] session.actor.prompt.received"
+        );
+        let mut acp_blocks =
```
pablonyx: `turn/active.rs` still owns turn start and busy queue mutation, not just the active prompt loop. This file builds ACP prompt blocks, inlines the first-prompt append, begins turns, deletes executed queue rows, emits pending-prompt events, handles busy prompt enqueue/edit/delete, and drains the next prompt. That leaves `turn/start.rs` and `turn/queue.rs` as thin wrappers while the ordering-sensitive logic remains concentrated here. Please move begin-turn/build-prompt work into `turn/start.rs` and busy queue mutation/drain work into `turn/queue.rs`, or rename/split the module so the ownership matches the actual concern.

anyharness/crates/anyharness-lib/src/live/sessions/actor/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+pub(in crate::live::sessions::actor) use crate::live::sessions::connection::shutdown::close_native_session;
+pub(in crate::live::sessions::actor) use crate::live::sessions::connection::start::initialize_connection;
+pub(in crate::live::sessions::actor) use crate::live::sessions::connection::types::{
+    NativeSessionStartupDisposition, SessionStartupState, SessionStartupStrategy,
+};
+pub(in crate::live::sessions::actor) use crate::live::sessions::handle::{
+    LiveSessionExecutionSnapshot, LiveSessionHandle,
+};
+
+pub(in crate::live::sessions::actor) use background_work::*;
```
pablonyx: This module has become an internal prelude/barrel for the whole actor. Because child modules import `crate::live::sessions::actor::*`, each file's real dependencies are hidden and the split loses some of the path-level ownership the repo-shape guidance is trying to create. Please replace the internal glob-prelude with direct imports in the concern modules, or keep `mod.rs` limited to module declarations and the intentional actor surface.

pablonyx (commented):
I reviewed the migration against the updated plan and the AnyHarness repo-shape guidance. The behavior-preserving extraction is mostly in place and the focused/full anyharness-lib tests pass locally, but I left a few comments where the implementation still weakens the intended ownership boundaries or adds unrelated ratchet churn.

pablonyx:
## Review of the session actor split

I verified the mechanical outcome first: `cargo build -p anyharness-lib` and `cargo test --no-run` succeed; `check_anyharness_boundaries.py`, `check_anyharness_old_paths.py`, and `check_max_lines.py` all pass; `rg "acp::session_actor|super::session_actor|pub mod session_actor" anyharness/crates/anyharness-lib/src` is empty; the 6,469-line `acp/session_actor.rs` is fully deleted. All gates from the plan's verification block pass. The topology largely matches the session-actor spec, and doc status has been flipped from "draft/deferred" to "authoritative" in the spec, live-runtime guide, and repo-shape guide.

That said, there are several design issues that are worth addressing before this lands, and a few that I'd flag for follow-up.

### Blocker-ish: the module-wide glob prelude undermines the split

`anyharness/crates/anyharness-lib/src/live/sessions/actor/mod.rs:94-123` re-exports every concern submodule with `pub(in ...actor) use <submodule>::*;`, and every concern file starts with `use crate::live::sessions::actor::*;`. There are 29 such star-imports. The effect is a crate-local prelude that dissolves the very module boundaries this PR is meant to introduce:

- A reader in `turn/active.rs` or `notifications/dispatch.rs` cannot tell from the imports whether `handle_resolve_interaction`, `resolve_pending_interactions`, `emit_live_config_update`, `persist_requested_config_value_if_changed`, etc. live in this file, a sibling concern, or a foreign crate. Everything is ambient.
- The spec's Target Folder Shape (`specs/session-actor.md`) explicitly lists the concern grammar (`handle/apply/queue/persist/finish/diagnostics`) so that each file's allowed dependencies are visible from its imports. The glob prelude hides that dependency graph rather than enforcing it.
- Repo-wide CLAUDE.md says "Use direct imports. Do not add barrel files or convenience re-export modules." Repo-shape guide says `mod.rs` should "declare and lightly re-export a cohesive module surface." 23 `pub use ...::*` lines at `actor/mod.rs` is not "light" — it's a crate-internal prelude dressed up as a module root.
- Concrete consequence: when `notifications/dispatch.rs` calls `emit_live_config_update`, the reader has to grep the actor tree to find that it lives in `config/persist.rs`. That's the thing the split is supposed to make obvious.

Recommended fix: delete the `pub(...) use <sub>::*;` block in `actor/mod.rs` and the `use crate::live::sessions::actor::*;` prelude at the top of every concern file. Replace with explicit `use crate::live::sessions::actor::config::persist::emit_live_config_update;` style imports. Yes, this means each file grows ~15–25 import lines; that's the point — the imports are the dependency graph.

The `connection/` subtree already gets this right (explicit `use super::*;` pulling a short, curated list from `connection/mod.rs`, plus explicit path imports in `start.rs` / `native_session.rs`). Match that style across `actor/`.

### Spec drift that should be reflected in code or the spec

The spec's §Target Folder Shape was updated to add `startup.rs`, `background_work.rs`, and `fork/` as top-level actor files/folders. That matches reality, good. But a few smaller deviations slipped through:

- `actor/startup.rs` is 541 lines — over the preferred 350 and right at the 500 hard limit that max-lines enforces for AnyHarness (the test file sneaks through because the allowlist blocks only component-level files). The file mixes process spawn, init, native session startup, four live-config emissions, preference application, session-ended logging, and thread-spawn plumbing. Consider splitting: move `spawn_session_actor` / `spawn_session_actor_pending` / `PendingSessionActor` to `actor/spawn.rs` and leave `start_actor` + `dispatch_startup_drain` in `startup.rs`. That also gets `startup.rs` to ~300 lines.
- `actor/state.rs` is only 24 lines and just holds `SessionActorConfig`. The spec says `state.rs` should own "actor-owned live state and phase types." The actual live state (`SessionStartupState`, `PersistedSessionConfigState`, `ResumeReplayFilter`, `ActorExitDisposition`) is scattered across `connection/types.rs`, `config/types.rs`, `notifications/replay_filter.rs`, and `shutdown/types.rs`. That's fine as long as each type lives with its concern, but then rename or absorb `state.rs` — don't keep a file named after a role it no longer performs. Today it's misleading.
- `config/types.rs:31-35` puts `ConfigPurpose` with visibility `pub(in crate::live::sessions)` while the surrounding items are `pub(in crate::live::sessions::actor)`. The wider visibility exists so `connection/types.rs` can call `find_select_option_by_purpose`. That's an actor concept leaking into connection — either the helper moves into `connection/` (it's purely about building `SessionStartupState` from ACP responses) or `ConfigPurpose` moves down into shared live types. Don't resolve it by silently widening visibility.

### event_loop.rs: readable, but only because real dispatch moved into `turn/active.rs`

The spec says (Acceptance Criterion 1): "Opening event_loop.rs shows the session engine event loop without reading prompt/config/notification/shutdown implementation details." The file is 247 lines and reads cleanly as a `tokio::select!` over `command_rx` / `notification_rx` / `background_work_rx`. That's the intended shape.

But the Prompt arm delegates to `handle_active_prompt` in `turn/active.rs`, which is 529 lines and contains its own inner `tokio::select!` with busy-path dispatch for 14 command variants — including inline re-implementations of `InjectRuntimeEvent`, `SetConfigOption`, pending-prompt enqueue, `EditPendingPrompt`, `DeletePendingPrompt`. Much of that logic is duplicated across idle (event_loop.rs) and busy (turn/active.rs) paths:

- The `InjectRuntimeEvent` handler appears verbatim in both (event_loop.rs:134-143 and turn/active.rs:317-326).
- The busy `SetConfigOption` block (turn/active.rs:327-361) reimplements what `config/queue.rs::queue_pending_config_change` plus `config/persist.rs::persist_requested_config_value_if_changed` already compose. `config/handle.rs` exists but is only called from preferences application, not from the busy config path. That's the exact "split by idle vs busy but keep logic in one handler" pattern the spec calls for (§Idle vs Busy Rules: "This split should be explicit in code. Do not hide busy/idle behavior behind a single large command match.").
- The busy `Prompt` arm (turn/active.rs:374-434) re-implements queue insertion + `PendingPromptAdded` emission inline.

Recommendation: pull each busy command into `turn/commands/<cmd>.rs` (or at minimum into `turn/queue.rs` / `turn/handle.rs`) and have both `event_loop.rs` and `turn/active.rs` call the same functions. The duplication today is the same invariant (durable insert → sink emit → respond) written twice; it's the kind of drift that causes event-ordering bugs later.

### Invariants: comments are good, but several are implicit

`turn/active.rs` calls out invariants 2, 3, and 5 inline. Good. But the spec lists 12 required invariants, and at least two are not obvious from the split code:

1. **"shutdown emits terminal session state exactly once"** — `shutdown/persist.rs::persist_exit_disposition` emits `session_ended(Closed)` on Close and `session_ended(Error)` on Error, but emits nothing on Dismiss. That's current behavior on main (I checked), but after a split it's worth a `// Invariant: Dismiss does not emit session_ended; the caller treats dismiss as "detach without announcing termination"` comment. Otherwise the `ActorExitDisposition::Dismiss => {}` arm reads like a bug.
2. **"ACP notifications are persisted raw before normalized event handling"** — preserved (`notifications/handle.rs:56-63` persists, then `normalize_notification`). But the replay-filter suppression happens *between* persist and normalize (handle.rs:65-73). That's the prior behavior, so fine, but a comment "Persist first, then suppress/normalize" would harden the contract.

### Tests: layout violates the spec's module convention

`actor/tests/{mod,config,connection,notifications,prompt,shutdown}.rs` puts the test fixtures in a sub-module, which works, but:

- The plan (last round) specified `actor/test_support.rs` gated by `cfg(test)` for shared fixtures. The current layout has 220 lines of shared fixtures in `tests/mod.rs` (an unusual place for shared helpers; typically `tests/common.rs` or `test_support.rs`).
- More importantly: `tests/mod.rs:7-18` imports 27 private items from `super::*` using a flat list. This works only because of the glob-prelude criticized above. Remove that and the tests break cleanly — which is the *correct* signal (tests should import from their owning concern).

### Minor / follow-ups

- `connection/mod.rs:8-14` declares `pub(in crate::live::sessions) use` for six external types as a connection-level prelude. Narrower than the actor prelude, but same smell. Fine to leave for now if actor is fixed.
- `connection/types.rs:2-3` imports from `crate::live::sessions::actor::config::{selection, types}` — connection reaching up into actor. Document why or push `ConfigPurpose`/`find_select_option_by_purpose` down into a shared live location. Spec says actor depends on connection, not the reverse.
- `fork/policy.rs` is 32 lines and contains only `has_anyharness_targeted_fork_extension`. That function is about *parsing an ACP capability meta field*, not actor-side fork policy. It belongs in `connection/native_session.rs` next to `start_native_session`'s fork branch or in a new `integrations/acp/` helper. Putting it in `fork/policy.rs` because the word "fork" appears is the wrong split axis.
- `shutdown/cleanup.rs` is 12 lines — just `interaction_resolution_for_exit`. Either inline it into `shutdown/handle.rs` or merge `cleanup.rs` into `persist.rs` (they're already called together by `finalize_established_actor_exit`).
- `turn/handle.rs` and `turn/start.rs` are 16 and 12 lines respectively. Spec expected `handle.rs` to be the Prompt command entrypoint. Today the Prompt command entrypoint lives in `event_loop.rs:36-67` inline, and `turn/handle.rs` only holds a codex-specific first-prompt helper that is misnamed. Either move the Prompt command dispatch into `turn/handle.rs` (matching the spec's concern grammar), or rename the file to `turn/first_prompt.rs`.
- `interactions/mod.rs` lacks `types.rs` from the original plan, but that's defensible — there were no interaction-specific types beyond what's in `command.rs`. Worth a note in the spec that `interactions/types.rs` is intentionally absent.
- `acp/replay_actor.rs` and `acp/manager.rs` still live under `acp/**`. Plan explicitly deferred these renames; noted. But `live/sessions/handle.rs` is now imported by both — the asymmetry ("handle is live, manager is acp") reads worse than before the split. Worth a note in the PR description that the manager rename is the next PR.
- PR body says 93 files changed, +7226/-6583. Net ~+650 lines after a pure refactor seems high; much of that is the 29 star-import prelude lines × 29 files (~250) plus the duplicated busy/idle command arms (~150). If the glob prelude goes away and busy/idle share handlers, net should be close to zero.

### What's good

- The directory shape matches the spec. Anyone opening `live/sessions/actor/` gets a readable layout.
- Latency helpers correctly moved to `observability/latency.rs` — that's the prep the plan called for, and the boundary check catches regressions.
- `connection/` is cleanly separated and doesn't reach into actor code (modulo the one `ConfigPurpose` leak).
- `handle.rs` is 163 lines and genuinely owns the public command/subscription port, matching the spec's §Collaborators.
- Ratchets are in place: old-paths check blocks `acp/session_actor.rs` from returning, max-lines allowlist does not add any new actor files.
- Doc status flips are done in the same PR (spec, live-runtime, repo-shape all now "authoritative" / "split complete").

### Suggested order for fixes before merge

1. Remove the glob prelude in `actor/mod.rs` and `use crate::live::sessions::actor::*;` from concern files. This is the big one.
2. Collapse busy/idle command duplication between `event_loop.rs` and `turn/active.rs` into shared handlers.
3. Split `startup.rs` into `spawn.rs` + `startup.rs` (or move spawn logic into `event_loop.rs` — it's only called once).
4. Resolve the `ConfigPurpose` / `connection -> actor` dependency: move the helper down, not the type up.
5. Either rename/empty `state.rs` or put real state there.
6. Move `has_anyharness_targeted_fork_extension` to where ACP capability parsing lives.

Items 1, 2, and 4 are the ones I'd block review on. The rest can be follow-ups if scope is already large.

anyharness/crates/anyharness-lib/src/live/sessions/actor/tests/connection.rs
```diff
@@ -0,0 +1,182 @@
+use super::*;
```
pablonyx: The migration created owner modules for `connection/stderr.rs`, `connection/native_session.rs`, `turn/diagnostics.rs`, and the other concerns, but the characterization tests remain centralized under `actor/tests/**` and import connection behavior through the actor test prelude. That makes the test layout disagree with the new ownership: connection behavior is tested as actor behavior, and adding a connection-only test now depends on actor test support. Please move concern tests next to the owning modules, or at least under a `live/sessions/connection/tests` module, and keep `actor/tests` for true actor integration fixtures.

docs/anyharness/README.md
```diff
@@ -404,13 +404,15 @@ Known transitional issues:
   type remains the caller-facing store surface.
 - `SessionEventSink` is split under `acp/event_sink/**`. It has not moved to
   final `live/sessions/event_sink/**` topology yet.
-- Some runtime/domain modules import `crate::api::http::latency`; latency
-  helpers should move to `observability/`.
+- Latency helpers live under `observability/latency.rs`; lower layers should
```
pablonyx: This note now says latency helpers live under `observability/latency.rs`, but the earlier Code Map and transitional mapping still say current `api/http/latency.rs -> target observability/latency.rs`. Since this PR actually moved the file, those earlier references are stale and will send future cleanup work to a file that no longer exists. Please update the Code Map/current mappings so `observability/latency.rs` is the current owner.

pablonyx (commented):
Second pass on the same head commit (`c117af2`). I did not repeat the earlier inline comments; they still apply because the branch has not changed. Focused checks still pass locally. One non-inline follow-up to decide: `scripts/check_anyharness_old_paths.py` still does not block `anyharness/crates/anyharness-lib/src/acp/session_actor.rs`. If this migration PR is meant to carry the old-path ratchet, add it here; otherwise track that as an explicit follow-up because the current old-path check passes without protecting the deleted actor path.

pablonyx:
## Follow-up review — second pass

No new commits since the first review (HEAD still `c117af2e`), so the prior comment's blockers remain. Flagging additional things I found on a second read that weren't in the first pass.

### Behavior preservation — spot checks vs. `origin/main`

I diffed the split code against pre-split `acp/session_actor.rs` on `origin/main` for the invariants most at risk. Results:

- ✅ **Busy flag ordering**: `handle.busy.store(true)` at entry to `handle_active_prompt` (turn/active.rs:54), `handle.busy.store(false)` after the drain loop (turn/active.rs:527), `handle.finish_prompt()` after final exit finalization (event_loop.rs:244). Matches the original `busy.store(true)` → drain → `busy.store(false)` → `finish_prompt` sequence and preserves the "actor is the sole writer of busy" invariant.
- ✅ **Resume replay filter disable timing**: `resume_replay_filter.disable()` at turn/active.rs:494 — called after `finish_prompt_result` returns, inside the drain loop but before the next-queue peek. Matches origin/main:2304 placement.
- ✅ **Turn-ended → idle → on_turn_finish → apply_pending_config_changes ordering**: preserved in `turn/finish.rs:163-217`. The exact sequence (sink.turn_ended → set_execution_phase(Idle) → update_status("idle") → callback → apply_pending_config) matches origin/main:2163-2244.
- ✅ **Exit finalization**: `shutdown/handle.rs::finalize_established_actor_exit` → `cleanup::interaction_resolution_for_exit` → `resolve_pending_interactions` → `shutdown/persist.rs::persist_exit_disposition` preserves the original `error → session_ended → status update` ordering. Test `finalize_error_exit_cancels_pending_permission_and_marks_session_errored` asserts `["interaction_resolved", "error", "session_ended"]`; the Dismiss variant test confirms no terminal session_ended event. Good coverage.
- ✅ **Idle `Cancel` vs busy `Cancel` semantics**: preserved. Idle Cancel (event_loop.rs:172-175) only forwards ACP cancel, does **not** resolve pending interactions — matches origin/main:2448-2451. Busy Cancel (turn/active.rs:252-264) resolves-then-cancels. That asymmetry is load-bearing and easy to accidentally "fix."

Behavior preservation looks good. No ordering regressions found on the hot paths I checked.

### Dead-state check: spec has a feature this PR deliberately disables

`startup.rs:526-536` has:
```rust
let adapter_targeted_fork_ready = fork
    && fork_capability
        .and_then(|capability| capability.meta.as_ref())
        .map(has_anyharness_targeted_fork_extension)
        .unwrap_or(false);
if adapter_targeted_fork_ready {
    tracing::debug!(
        "agent advertises edit-safe targeted fork metadata; public targeted fork remains disabled until runtime target dispatch is implemented"
    );
}
SessionActionCapabilities {
    fork,
    targeted_fork: false,
}
```

`adapter_targeted_fork_ready` is computed, traced, and discarded. `targeted_fork` is hardcoded to `false`. This came over verbatim from the original — **not a regression** — but since `fork/policy.rs` only holds `has_anyharness_targeted_fork_extension` (32 lines, referenced once from this trace-only site), the entire `fork/policy.rs` file exists to keep a disabled feature discoverable. Two options: (1) delete the detection and revive it when targeted-fork lands, (2) inline the helper into `startup.rs` and delete `fork/policy.rs`. Either is cleaner than an entire concern folder for a single grep-for-later marker.

### `event_loop.rs` — Prompt arm is NOT pure dispatch

Spec §Acceptance Criterion 1: "Opening event_loop.rs shows the session engine event loop without reading prompt/config/notification/shutdown implementation details."

Currently in event_loop.rs the Prompt arm (lines 36-67) constructs `ActivePromptContext` with **14 borrowed fields** inline. That's not dispatch — it's wiring. The whole context-construction block should move into `turn::handle::begin_active_prompt(actor_state, request)` so event_loop.rs reads:

```rust
Some(SessionCommand::Prompt { payload, prompt_id, latency, from_queue_seq, respond_to }) => {
    if let Some(next_exit) = turn::handle::begin_active_prompt(
        &mut actor_state,
        ActivePromptRequest { payload, prompt_id, latency, from_queue_seq, respond_to },
    ).await {
        exit_reason = next_exit;
        break;
    }
}
```

This also addresses the first-review point about the `&mut` soup: bundle the actor-owned mutable state into a single `ActorState` struct once (event_sink, notification_rx, background_work_rx, background_work_registry, persisted_config_state, startup_state, resume_replay_filter) and pass one `&mut actor_state` rather than 14 separate borrows at every handler boundary. Currently each handler's signature is a re-derivation of the same field list; add one field and you edit four functions.

### `fork/policy.rs` has the wrong split axis — confirming

On re-read, `has_anyharness_targeted_fork_extension` parses ACP `Meta` JSON for an anyharness-specific schema version and `fileEffects` field. That's ACP capability introspection, not fork *policy*. Fork policy is "am I idle? is the queue empty?" — which lives in `fork/handle.rs::verify_fork_ready`. Fix: move `has_anyharness_targeted_fork_extension` to `connection/native_session.rs` (next to the fork branch in `start_native_session`), delete `fork/policy.rs`, and the `fork/` folder contains just `handle.rs` — which is fine, or fold `handle.rs` content into `actor/fork.rs` and drop the folder.

### Test coverage gaps

41 tests pass. Coverage is solid for:

- `shutdown/` cleanup and finalization (6 tests, good)
- `connection/` client capabilities and stderr (15 tests)
- `notifications/` replay filter and raw persistence (7 tests)
- `config/` selection and purpose matching (8 tests)

Missing: **no tests exercise `handle_active_prompt`, `finish_prompt_result`, `apply_specific_config_option`, or `apply_pending_config_changes_if_idle`** — the four largest and most behaviorally sensitive functions in the split. The first three are also the functions most likely to regress from a bad refactor. Adding a characterization test for `apply_pending_config_changes_if_idle` (which has the drift-prone "delete on reject" behavior at queue.rs:93-95) would harden the config queue boundary.

Note: queue.rs:93-95 deletes the pending row on `Rejected` just like on `Ok`. That's preserved from the original but is a surprising "retry won't happen" behavior. Worth a comment explaining *why* we drop rejected changes rather than surfacing them — otherwise a future reader will assume it's a bug.

### `serialize_action_capabilities` path

Small nit: `startup.rs:499-520` imports `serialize_action_capabilities` from `crate::sessions::model` (via the `actor/mod.rs` prelude). That's a durable-model helper being called from live code. Fine — but since this is the only live-side caller, it's worth verifying the function is not better placed in `connection/native_session.rs` next to `start_native_session`. The serialize happens immediately after init_response; the model crate doesn't own it conceptually, it owns the serialization format.

### `tests/mod.rs` imports (expanding first-review point)

tests/mod.rs:7-18 imports **27 private items via `super::*`** to set up the shared fixtures. When the actor prelude is removed (first-review item #1), this import block becomes 27 explicit paths. That's fine — but it's also a signal: the test module is currently reaching into every concern submodule's private surface. Organize tests alongside their concerns:

- `actor/turn/tests.rs` for turn tests
- `actor/config/tests.rs` for config tests
- etc.

Instead of one monolithic `actor/tests/` folder that needs access to every sibling's internals. This matches the spec's concern grammar and reduces the test module's blast radius.

### Summary of new items

1. Behavior preservation on the hot paths I checked is clean. Ship-blocking correctness concerns: none found beyond what the first review flagged.
2. `fork/policy.rs` exists only to hold a trace-only detector for a disabled feature. Delete or inline.
3. `event_loop.rs` Prompt arm wires 14 fields inline — not dispatch. Bundle state, make the handler self-dispatching.
4. No tests for the four biggest functions (active-prompt loop, finish_prompt_result, apply_specific_config_option, apply_pending_config_changes_if_idle). Add at least one characterization test for config apply-after-turn.
5. Test module reaches into every concern's private surface; re-home tests next to their concerns.

Items 2, 3, and 5 compound well with the first-review blockers (remove actor prelude, deduplicate busy/idle command arms). Doing them together produces a coherent cleanup; doing them piecemeal produces churn.

pablonyx:
## Third-pass review — current WIP (uncommitted in worktree)

Reviewing the in-progress changes on disk at `/Users/pablo/.proliferate/worktrees/proliferate/anyharness-session-actor` (not yet pushed to this PR, HEAD is still `c117af2e`). +759 / −659 across 28 files. Significant progress against prior rounds.

### What landed well

1. **Glob prelude is gone.** `actor/mod.rs` is now 15 lines — pure `mod` declarations, no `pub use *` barrel. Every concern file has explicit per-symbol imports (e.g. `turn/active.rs` imports 13 discrete items by path; `config/apply.rs` imports only what it uses). This is the correct end state per repo-shape guide §Module Style and global CLAUDE.md's "no barrels" rule. Huge improvement.

2. **`ConfigPurpose` leak resolved cleanly.** `connection/types.rs` now owns a sibling `NativeSessionStartupState` with purely-ACP-shape fields, and `actor/state.rs::SessionStartupState` adds the purpose-aware methods (`has_raw_or_legacy_mode_control`, `legacy_mode_contains_value`) on the actor side. The `From<NativeSessionStartupState>` conversion at the actor/connection seam makes the dependency direction correct: actor depends on connection, never the reverse. The spec §Collaborators is honored.

3. **`fork/policy.rs` deleted.** `has_anyharness_targeted_fork_extension` moved to `connection/native_session.rs` alongside ACP capability parsing — correct home per spec §Connection Boundary. `fork/` now contains only `handle.rs`, which is the actor command surface (fork/verify/close). Good.

4. **`startup.rs` split into `spawn.rs` + `startup.rs`.** Thread spawn + readiness (143 lines) separated from ACP process/session startup orchestration (401 lines — still large but now single-concern). Spec was updated to describe the split. Acceptable.

5. **`turn/start.rs` and `turn/queue.rs` now carry real content.** `begin_prompt_turn` is 95 lines of genuine turn-start behavior (ACP block conversion, begin_turn emission, durable queue delete); `handle_busy_prompt_queue` / `emit_prequeued_pending_prompt_added` / `next_pending_prompt_for_drain` pull the busy queue-state transitions out of `turn/active.rs`'s inline code. Formerly-stub files are now earning their place.

6. **Docs kept aligned.** `specs/session-actor.md` lists `spawn.rs`; `guides/live-runtime.md` drops `fork/policy.rs` from the target tree; `docs/anyharness/README.md` updates latency ownership to `observability/latency.rs`. The doc/code parity that CLAUDE.md requires is maintained.

7. **Connection tests co-located with concerns.** `connection/native_session.rs`, `connection/start.rs`, `connection/stderr.rs` each now have their own `#[cfg(test)] mod tests`. `actor/tests/connection.rs` was deleted. That's exactly the re-homing the prior review recommended.

### New blockers

#### 1. **Regression in `scripts/max_lines_allowlist.txt` — gate fails**

The WIP removes `desktop/src/hooks/workspaces/tabs/use-workspace-shell-activation.ts` from the allowlist. That file is 692 lines on `origin/main` (it came from PR #201 "perf(desktop): reduce jank and stabilize optimistic shells"), is unrelated to this PR's work, and the original c117af2e commit had it allowlisted correctly. Running `python3 scripts/check_anyharness_boundaries.py && python3 scripts/check_anyharness_old_paths.py && python3 scripts/check_max_lines.py` now yields:

```
Files above their max line threshold that are not allowlisted:
  desktop/src/hooks/workspaces/tabs/use-workspace-shell-activation.ts: 692 (max 600)
```

**Fix**: revert this single line in `scripts/max_lines_allowlist.txt`. The shell-activation file's size is not in scope for this PR — it should remain allowlisted until its owning PR shrinks it. Attempting to shrink it here would expand the PR's blast radius onto desktop code the session-actor migration shouldn't touch.

#### 2. **Idle/busy command duplication unresolved**

From the first-round review, item #2 was "collapse busy/idle command duplication between `event_loop.rs` and `turn/active.rs` into shared handlers." Still not done:

- `InjectRuntimeEvent`: event_loop.rs:158-167 and turn/active.rs:305-314 — **byte-for-byte identical** 10-line blocks.
- `ReplayAdvance`: same `Err(anyhow::anyhow!("session is not a replay session"))` in both.
- Busy `Fork` / `VerifyForkReady` / `CloseNativeSession` reject paths: identical 1-line `Err(Busy)` replies.
- Busy `SetConfigOption` (turn/active.rs:315-349): 35 lines reimplementing `queue_pending_config_change + persist_requested_config_value_if_changed + delete_pending_config_change` in inline match. Idle-side `SetConfigOption` (event_loop.rs:168-195) calls `apply_specific_config_option`. These are the same command with different timing — exactly the situation spec §Idle vs Busy Rules calls out ("This split should be explicit in code. Do not hide busy/idle behavior behind a single large command match.").

Suggested shape:
```rust
// actor/commands/inject_runtime_event.rs  (both arms)
pub async fn handle_inject_runtime_event(...)

// actor/commands/config.rs
pub async fn handle_idle_set_config_option(...)
pub async fn handle_busy_set_config_option(...)
```

Then event_loop.rs and turn/active.rs each become a thin dispatch over `match` with a function call per arm. Right now event_loop.rs is 247 lines — most of which is inline handler bodies. Spec acceptance criterion #1 ("event_loop.rs shows the session engine event loop without reading implementation details") is not yet met.

#### 3. **`turn/handle.rs` is a passthrough — doesn't earn its keep**

Current `turn/handle.rs::handle_idle_prompt_command` takes an `IdlePromptContext` with 14 borrowed fields, destructures it, reconstructs an `ActivePromptContext` with the same 14 fields, and calls `handle_active_prompt`. That's a 40-line no-op wrapper. Two options:

- Inline: rename `ActivePromptContext` → `PromptContext`, call `handle_active_prompt` directly from event_loop.rs, delete `turn/handle.rs`.
- Make `handle.rs` do real work: move the `IdleVsBusy` dispatch into `handle.rs` and have it decide between `turn::start::begin_prompt_turn` (if idle) vs. `turn::queue::handle_busy_prompt_queue` (if busy). Right now that decision lives inside `handle_active_prompt` via `current_respond_to` / `from_queue_seq` flags; hoisting it to `handle.rs` would let event_loop.rs and the busy arm both call the same entrypoint.

Either way, a passthrough that re-destructures a 14-field struct is a symptom of the context-bundle problem flagged in review #2.

#### 4. **Context structs with 14 `&mut` fields should be an `ActorState` struct**

Previously flagged. `IdlePromptContext`, `ActivePromptContext`, and `PromptFinishContext` all carry the same 13-14 fields. The correct fix is a single `ActorLiveState` (owns: event_sink, notification_rx, background_work_rx, background_work_registry, persisted_config_state, startup_state, resume_replay_filter, handle, store, attachment_storage, plus the immutable refs to config/conn/native_session_id). Spawn it once in `run_actor`, pass `&mut actor_state` to every handler. Then:

- Adding a field edits one struct, not four.
- `turn/handle.rs::handle_idle_prompt_command` becomes a real one-liner rather than 40 lines of destructure/reconstruct.
- Command handlers extracted per item #2 become `async fn handle_xxx(state: &mut ActorLiveState, ...)`.

### Smaller things

- `startup.rs` is still 401 lines even after the `spawn.rs` split. The four "emit/apply/restore/emit" stanzas each have a `tracing::info!{..completed} / tracing::warn!{..failed}` 20-line branch. A `log_latency_outcome(label, started, result)` helper would cut ~80 lines and make the actual ordering (`initial_live_config → apply_preferences → restore_live_config → post_preferences_live_config`) readable.

- `action_capabilities_from_acp` in `startup.rs:382-401` still computes `adapter_targeted_fork_ready`, traces it, and throws it away. `has_anyharness_targeted_fork_extension` moved to `connection/native_session.rs` but the trace-only consumer remained in `startup.rs`. Functionally equivalent to origin/main, but the prior review's question stands: is a trace-only dead-code marker worth keeping? Deletion is clean; revival when targeted-fork lands is trivial (the ACP meta shape is documented). Either delete both or add a `TODO(targeted-fork)` linked to the issue tracking it.

- `tests/mod.rs` no longer imports 27 items via `super::*` — it's now 23 items from `super::{...}`. That import list compiles because the actor still re-exports them *somewhere* during the transition. Worth verifying: if you strip `pub(crate)` from `state` / `spawn` / `turn` in `actor/mod.rs`, do the tests still compile? If yes, tighten visibility. If no, that's more actor-internal surface leaking into a shared test module — the "re-home tests next to concerns" follow-up from round 2 still applies.

- Connection dedup progress: `start.rs` (324 lines, has inline tests), `native_session.rs` (306 lines, has inline tests), `stderr.rs` (144 lines, has inline tests), `process.rs` (164 lines). All under the 500-line component budget. Good.

### Invariant preservation — re-verified

I diffed the busy/idle sequences again against origin/main:

- Busy flag transitions (set true at turn/active.rs:54, false at turn/active.rs:417): preserved.
- `resume_replay_filter.disable()` after `finish_prompt_result`: preserved at turn/active.rs:392.
- Turn-end sequence (`turn_ended → set_execution_phase(Idle) → update_status("idle") → on_turn_finish → apply_pending_config_changes_if_idle`): preserved in `turn/finish.rs`.
- Startup drain self-dispatch: preserved in `startup.rs::dispatch_startup_drain` with invariant-5 comment.
- Raw notification persist before normalize: preserved (`notifications/handle.rs::handle_notification_with_resume_replay_filter` persists first, then suppresses/normalizes).

No correctness regressions introduced by this WIP.

### Summary

Of the six items flagged in review #1:
- **(1) glob prelude**: FIXED
- **(2) busy/idle dedup**: NOT FIXED
- **(3) startup.rs split**: PARTIALLY FIXED (spawn.rs extracted; startup.rs still 401 lines)
- **(4) ConfigPurpose leak**: FIXED
- **(5) state.rs rename/fill**: FIXED (now owns SessionStartupState + purpose methods)
- **(6) fork/policy.rs misplacement**: FIXED

Of review #2:
- **Test re-home**: PARTIAL (connection tests re-homed; actor tests still under `actor/tests/`)
- **fork/policy.rs**: FIXED
- **Trace-only targeted-fork marker**: NOT FIXED
- **Context-struct soup / event_loop dispatch purity**: NOT FIXED
- **Test coverage for active-prompt / finish_prompt / apply_specific_config / apply_pending_config**: NOT FIXED

To land cleanly:

1. **Must-fix**: revert `shell-activation.ts` removal from the max-lines allowlist.
2. **Should-fix**: extract the duplicated busy/idle command arms into shared `async fn`s.
3. **Should-fix**: collapse IdlePromptContext/ActivePromptContext/PromptFinishContext into one `ActorLiveState` so handlers take `&mut actor_state`.
4. **Nice-to-have**: add 1-2 characterization tests for the config apply-after-turn path and the startup drain path.

The structural direction of the WIP is correct and matches the docs. The remaining work is tightening and test coverage, not architecture.

pablonyx:
## Third-pass review — uncommitted revisions inspected

Picked up the uncommitted changes in the worktree (not yet pushed to the PR). Running the full gate:

- `cargo build -p anyharness-lib`: **passes with 3 new warnings** (see "Regressions" below).
- `cargo test -p anyharness-lib`: 41 actor tests + 12 connection tests pass.
- `check_anyharness_boundaries.py`, `check_anyharness_old_paths.py`: pass (old-paths now blocks `acp/session_actor.rs`).
- `check_max_lines.py`: **FAILS**. `desktop/src/hooks/workspaces/tabs/use-workspace-shell-activation.ts` (692 lines) was silently removed from the allowlist but the file wasn't touched. This would block CI on merge.

### What the revisions actually fixed — good progress on spec alignment

Measured against `specs/session-actor.md` and `guides/live-runtime.md`, the uncommitted diff resolves several first- and second-review items:

1. **Glob prelude gone.** `actor/mod.rs` is now 15 lines with `mod config; mod turn; pub(crate) mod state; …` and no `pub(...) use ...::*`. Every concern file now has an explicit import list pointing at the functions it actually uses. This is the single biggest structural improvement — the module boundary is now *visible* in the imports. **First-review blocker #1 resolved.**
2. **`ConfigPurpose` / connection → actor leak fixed.** `connection/types.rs` now owns a shape-only `NativeSessionStartupState` (no actor-specific methods), and `actor/state.rs` has `impl From<NativeSessionStartupState> for SessionStartupState` plus the `has_raw_or_legacy_mode_control`/`legacy_mode_contains_value` methods. Connection no longer reaches up into actor. **First-review blocker #4 resolved.**
3. **`fork/policy.rs` deleted.** `has_anyharness_targeted_fork_extension` correctly moved to `connection/native_session.rs` next to the fork branch. `fork/` now contains only `handle.rs` (Fork/VerifyForkReady command handlers — actor policy, not ACP parsing). **Second-review item #2 resolved.** Docs updated to match.
4. **Spawn/startup split.** `actor/spawn.rs` (new, 153 lines) owns thread creation, `PendingSessionActor`, `ActorReadyResult`, and readiness waiting. `actor/startup.rs` (432 lines) owns ACP init + native session + live-config priming. Spec §Target Folder Shape updated in the same diff to name both files and explain the split. Good self-documenting boundary.
5. **Busy prompt enqueue deduplicated.** `turn/queue.rs::handle_busy_prompt_queue` and `::emit_prequeued_pending_prompt_added` now own the busy Prompt command logic, called from `turn/active.rs:362-373`. Previously inline at turn/active.rs:374-434. **First-review item #2 partially resolved** (still duplication remaining — see below).
6. **State consolidation.** `state.rs` grew from 24 lines to 109 and now owns `SessionStartupState` (moved from connection/types.rs), its methods, and the `From<NativeSessionStartupState>` bridge. File name now matches contents. **First-review item about `state.rs` being misleading: resolved.**
7. **Connection tests inline.** `connection/process.rs`, `connection/start.rs`, `connection/native_session.rs`, and `connection/stderr.rs` now have `#[cfg(test)] mod tests` blocks inside the concern file. `tests/connection.rs` deleted. Tests run as `live::sessions::connection::process::tests::*` instead of being imported from the actor test harness. **Second-review item #4 partially resolved** (connection side done; actor-side tests still live in `actor/tests/`).

The code is noticeably closer to what `specs/session-actor.md` §Target Folder Shape describes. Anyone opening `actor/mod.rs` now sees a clean concern list.

### Regressions introduced by the uncommitted diff

1. **Max-lines check broken.** `scripts/max_lines_allowlist.txt` removes `desktop/src/hooks/workspaces/tabs/use-workspace-shell-activation.ts` while the file is still 692 lines. The removal is unrelated to the actor split and belongs in its own PR (or should be reverted here). `check_max_lines.py` reports the file as a new violation and will block the merge.
2. **3 new dead-code warnings.** Build output:
   ```
   warning: variant `ActorDead` is never constructed (PromptAcceptError, command.rs:17)
   warning: variant `ActorDead` is never constructed (QueueMutationError, command.rs:30)
   warning: function `spawn_session_actor` is never used (spawn.rs:68)
   ```
   - `ActorDead` variants: the removed actor prelude was the only thing keeping `crate::live::sessions::handle.rs` imports live for the "handle got dropped" paths. Check whether `resolve_interaction` on `LiveSessionHandle` still maps channel-closed errors to `ResolveInteractionCommandError::ActorDead` — if not, the error-path is currently dead and `Result<PromptAcceptance, PromptAcceptError>` consumers can't distinguish dead-actor from enqueue-failed. Probably a real regression.
   - `spawn_session_actor` (the blocking wrapper): sole call site was `spawn_session_actor_pending` + `wait_ready` inline; the convenience wrapper is no longer called from `acp/manager.rs`. Either delete the function or wire a caller.

   Repo-wide CLAUDE.md is explicit: keep the repo buildable. Dead-code warnings in a refactor PR are the exact "half-finished implementations" anti-pattern the top-level instructions warn against.

### Spec-alignment items still open

Now that the structural fixes have landed, the remaining gap between code and spec is narrower but concrete:

1. **`turn/handle.rs::handle_idle_prompt_command` is an identity wrapper.** Lines 38-79: destructures `IdlePromptContext` into 14 locals, constructs an `ActivePromptContext` with the same 14 fields, calls `handle_active_prompt`, returns. Two context types with identical fields and an identity mapper between them. Collapse to a single `ActivePromptContext` and delete `IdlePromptContext` + `handle_idle_prompt_command`. The spec's §Target Folder Shape lists `turn/handle.rs` as "Prompt command entrypoint; choose start-now vs queue" — but the actual branching (idle Prompt → `handle_active_prompt` / busy Prompt → `handle_busy_prompt_queue`) lives in `event_loop.rs` and `turn/active.rs` respectively, not in `turn/handle.rs`. Either (a) move the idle-vs-busy decision into `turn/handle.rs` proper so event_loop.rs calls `turn::handle::dispatch_prompt(...)` and it chooses, or (b) rename `turn/handle.rs` to `turn/codex.rs` since all that's left is `first_prompt_system_prompt_append_for_codex_prompt`.

2. **Idle/busy command duplication remains for two arms.** `InjectRuntimeEvent` (event_loop.rs:158-167 vs turn/active.rs:305-314) and `SetConfigOption` (event_loop.rs:168-195 vs turn/active.rs:315-349) are still byte-level duplicates. The first is 9 lines and borderline noise; the second is 30+ lines with nontrivial behavior (different semantics: idle applies immediately via `apply_specific_config_option`, busy queues via `queue_pending_config_change`+`persist_requested_config_value_if_changed`). Extract `turn/queue.rs::queue_busy_config_change(...)` and call it from turn/active.rs; extract `actor/commands/inject_runtime_event.rs::handle(handle, event_sink, event)` and call from both paths. Spec §Idle vs Busy Rules is explicit: "This split should be explicit in code. Do not hide busy/idle behavior behind a single large command match." Two byte-identical blocks in two different match arms is the symptom that warning targets.

3. **`event_loop.rs` Prompt arm still wires 14 `&mut`s inline** (lines 61-85). Second-review item #3 unresolved. Options:
   - Bundle the actor-owned mutable state into a single `ActorState` struct built in `run_actor` and reborrowed at each handler call (`&mut actor_state`). Every handler signature collapses to 2-3 params. This matches the spec's §Owned State section naming exactly what the actor carries.
   - Or move the `IdlePromptContext`/`ActivePromptContext` construction into `turn::handle::begin_active_prompt(&mut actor_state, request)` so event_loop.rs is just `turn::handle::begin_active_prompt(&mut actor_state, req).await`.
   The "14 borrows at every call site" anti-pattern quietly opens the door to `&mut` ordering bugs. Bundle once.

4. **`startup.rs` at 432 lines, below soft-limit but still dense.** repo-shape soft limit for `live/**/actor*.rs` is 500 preferred / 900 hard. 432 is fine by the ratchet, but the file mixes: connection bring-up, ACP init, native session, initial live-config emission, preference application, persisted-config restore, post-preference live-config re-emission, ready signal, replay filter construction, and startup drain dispatch. Two natural sub-splits per the spec's concern grammar: `startup/native.rs` (lines ~50-155) and `startup/live_config_prime.rs` (lines ~160-320). Leave `startup.rs` as the orchestrator. Not blocking.

5. **`startup.rs` still holds `dispatch_startup_drain`** (lines 342-389). It self-sends a `SessionCommand::Prompt` with `from_queue_seq=Some(seq)` before the event loop starts. That's a turn-drain concern, not a connection-bringup concern. Belongs in `turn/queue.rs` or `turn/handle.rs`. Moving it also tightens the spec invariant #5 locality.

6. **`connection/mod.rs` still has a 4-line partial prelude** (`pub(in crate::live::sessions) use crate::domains::agents::model::{AgentKind, ResolvedAgent};` etc.). Smaller scale than the original actor glob and defensible, but after the actor cleanup it stands out. Either delete and have each file import directly, or leave with a one-line comment explaining why these four types are universal across connection files.

7. **Docs/runtime guide have a minor inconsistency.** `guides/live-runtime.md` Actor Files section still lists the old `commands/{idle,busy,prompt,config,interactions,plans,fork,close}.rs` shape (lines 82-90 on origin/main). The guide update in this PR removes only `fork/policy.rs` — it doesn't reconcile `commands/*.rs` vs the actual `turn/` + `config/` + `shutdown/` split that `specs/session-actor.md` now names as authoritative. Either delete the outdated `commands/` block in live-runtime.md, or explicitly note that the `commands/` namespace is a future reorganization and the current shape wins. Spec wins per repo rules, but the conflict between two "authoritative" docs is a trap for the next reader.

### Order to land

1. **Must fix before pushing to PR:**
   - Restore `use-workspace-shell-activation.ts` in the allowlist (or revert that hunk entirely).
   - Address the 3 dead-code warnings. Either exercise the dead paths (keep `spawn_session_actor` as pub for external callers if needed), or delete them.

2. **Should fix before merge:**
   - Collapse `IdlePromptContext`/`handle_idle_prompt_command` identity wrapper.
   - Extract `InjectRuntimeEvent` handler and busy `SetConfigOption` queueing to shared helpers.
   - Reconcile `guides/live-runtime.md` Actor Files block with the actual tree.

3. **Can follow up:**
   - Bundle actor-owned state into one `ActorState` struct to eliminate 14-`&mut` call sites.
   - Split startup.rs or move `dispatch_startup_drain` to `turn/`.
   - Decide fate of `connection/mod.rs` mini-prelude.

Net: the structural critique from reviews 1 and 2 has been largely answered. The *spec alignment* the team is optimizing for is visibly better. The regressions introduced (broken max-lines gate, three dead-code warnings) need to be cleaned up before this commits, and the small identity-wrapper / duplication / context-bundling items are the remaining distance to full spec alignment.

pablonyx:
## Fourth-pass review — `fc1cfea6` + `b3d68499`

Pulling in the two new commits. Running all gates:

- `cargo build -p anyharness-lib`: **passes, zero warnings**.
- `cargo test -p anyharness-lib --lib`: 428 tests pass.
- `check_anyharness_boundaries.py`, `check_anyharness_old_paths.py`, `check_max_lines.py`: all green.
- `rg "acp::session_actor"` on the tree: empty.

### Regressions from pass 3 — fixed

1. ✅ **Max-lines gate**: `scripts/max_lines_allowlist.txt` not touched in `fc1cfea6`. The pass-3 regression (removed allowlist line without touching the file) was not included. Gate is green again.
2. ✅ **Three dead-code warnings resolved correctly**:
   - `PromptAcceptError::ActorDead` and `QueueMutationError::ActorDead` variants removed. Confirmed these were never constructed on `origin/main` either — they were pre-existing dead code that the glob prelude was masking. The author also pruned the unreachable match arms in `sessions/runtime/prompt.rs:129-131` and `sessions/runtime/pending_prompts.rs:100-102,156-158`. Clean fix. `ResolveInteractionCommandError::ActorDead` was kept because `LiveSessionHandle::resolve_interaction` at `handle.rs:140-143` *does* construct it on channel-close. The distinction was correct.
   - `spawn_session_actor` (blocking wrapper): deleted. Only `spawn_session_actor_pending` remains, matching the actual caller in `acp/manager.rs:180`. Don't expose APIs without callers.

### Spec-alignment items from pass 3 — status

1. ✅ **`InjectRuntimeEvent` / `SetConfigOption` duplication eliminated.** Both sides now call shared helpers:
   - `notifications/dispatch.rs::inject_runtime_event` — used from `event_loop.rs:160` and `turn/active.rs:303`.
   - `config/handle.rs::handle_idle_config_command` (idle) + `handle_busy_config_command` (busy) — used from `event_loop.rs:168` and `turn/active.rs:307`. The busy/idle semantic split is now explicit, which is exactly what spec §Idle vs Busy Rules calls for ("This split should be explicit in code"). Good execution.

2. ✅ **`guides/live-runtime.md` Actor Files reconciled with the authoritative spec.** The outdated `commands/{idle,busy,prompt,config,interactions,plans,fork,close}.rs` block is gone; the doc now lists the actual `turn/` + `config/` + `notifications/` + `interactions/` + `fork/` + `shutdown/` tree that matches `specs/session-actor.md`. No more two-docs-in-conflict trap.

3. ❌ **`turn/handle.rs::handle_idle_prompt_command` identity wrapper still present** (turn/handle.rs:38-79). Destructures a 14-field `IdlePromptContext` and reconstructs an `ActivePromptContext` with the same 14 fields. Pure identity mapping. Easiest single cleanup in the tree: collapse `IdlePromptContext` into `ActivePromptContext`, delete `handle_idle_prompt_command`, call `handle_active_prompt` directly from `event_loop.rs`. Saves ~50 lines, removes a borderline-misleading layer.

4. ❌ **`event_loop.rs` Prompt arm still wires 14 `&mut`s inline** (lines 61-86). Unchanged. Not blocking, but the per-handler signature bloat is the remaining "not-quite-dispatch" wart in event_loop.rs. Still worth an `ActorState` bundle.

5. ❌ **`dispatch_startup_drain` still in `startup.rs`** (startup.rs:324, 342). It self-sends a `SessionCommand::Prompt` with `from_queue_seq = Some(seq)` — that's turn-queue concern, not connection-startup. Should live in `turn/queue.rs` or a new `turn/drain.rs`, called by startup.rs. Keeps turn-queue invariants in one place.

6. ❌ **`startup.rs` still 432 lines** covering connection bring-up → ACP init → native session → live-config priming → preference restore → post-preference re-emit → ready signal → replay filter → drain dispatch. Under hard limit but dense.

7. ❌ **`connection/mod.rs` still has the 4-line prelude** for `AgentKind`, `ResolvedAgent`, `to_acp_servers`, `SessionMcpServer`. Less egregious than the actor prelude that was eliminated, but inconsistent with the direction the rest of the tree moved. Two files use all four (`start.rs` and `native_session.rs`); two files need subsets (`process.rs` and `types.rs`). Explicit imports in each file would match the new actor style.

### New observations on `fc1cfea6`

1. **`inject_runtime_event` lives in `notifications/dispatch.rs`** (lines 28-41). That file should own ACP-notification routing; a runtime-event injection helper is unrelated. Its natural home is either `actor/commands/runtime_event.rs` (new), or inlined — it's only 9 lines and has two callers. Putting it in `notifications/dispatch.rs` because "event-ish" creates the same misleading naming the first review flagged about `fork/policy.rs`. Either rename the file to `notifications/runtime_events.rs` (and have it own both normalization and injection) or move the helper. The former is probably simpler.

2. **Test module no longer destructures 27 private items**. tests/mod.rs:7-26 now uses explicit `use super::config::apply::…; use super::notifications::plans::…;` paths — which is only possible because of the prelude removal. The import list is long but *informative*: you can see which concern surfaces the tests reach into. Good.

3. **tests/connection.rs deleted; tests moved inline next to their concern files** (as pass 3 noted in the uncommitted diff, now landed). The actor-test harness (`actor/tests/`) still exists for cross-concern shutdown/notifications/config/prompt tests that genuinely need shared fixtures — that's defensible. Further inlining (moving config tests into `config/*.rs` etc.) is possible but optional; repo convention isn't clear on the tradeoff and this is fine.

4. **`handle_busy_config_command` (config/handle.rs:92-124) correctly persists requested-value *before* checking the apply result against live config**. The previous inline version in active.rs did this too — preserved. The current ordering matches the invariant "ACP notifications persist raw before normalized handling" in spirit: user intent is durably captured before any async apply might fail. Worth a comment to that effect; otherwise a future reader may be tempted to "optimize" by moving the persist inside the `Ok(())` arm and breaking the semantic.

5. **`config/handle.rs:115` deletes the pending row on persist failure.** Same "delete on Rejected" semantic that was in `apply_pending_config_changes_if_idle` (queue.rs:94-95). Two different call sites silently agree on the same debatable semantic without a comment. Adding a one-liner at the definition ("rejected values are not retained to avoid infinite retry; the caller may re-submit if needed") would harden both.

### Spec alignment — scorecard

Against `specs/session-actor.md` §Target Folder Shape and §Acceptance Criteria:

| Criterion | Status |
| --- | --- |
| `actor/loop.rs` (renamed `event_loop.rs`) shows session engine loop without reading subsystem detail | ✅ With one caveat: the Prompt arm's 14-field context construction still reads like wiring, not dispatch. |
| Idle/busy command behavior explicit and separated | ✅ With `handle_idle_config_command` / `handle_busy_config_command` landed, this is now true for config. Prompt is split between `handle_active_prompt` (idle start + drain) and `handle_busy_prompt_queue` (busy enqueue). |
| `actor/turn/` shows start / active / finish / queued config apply / pending prompt drain | ✅ |
| `actor/config/` owns apply-vs-queue | ✅ |
| `actor/notifications/` delegates to event_sink | ✅, modulo the `inject_runtime_event` misfile. |
| `actor/interactions/` actor-side resolution/cleanup | ✅ |
| `actor/shutdown/` cancel/close/dismiss/error finalization | ✅ |
| ACP process/native-session mechanics in `live/sessions/connection/` | ✅ |
| `domains/sessions/prompt` product prompt prep preserved | ✅ (not touched) |
| `domains/sessions/mcp_bindings` preserved | ✅ |
| `live/sessions/event_sink` still owns normalization/persist/broadcast | ✅ (transitional `acp/event_sink/**`, per plan) |
| `acp/session_actor.rs` deleted | ✅ |
| Behavior preserved by tests or characterization coverage | ⚠️ 428 tests pass; still no characterization tests for `handle_active_prompt` / `finish_prompt_result` / `apply_specific_config_option` / `apply_pending_config_changes_if_idle`. Same gap as pass 2 flagged. |

12/13 acceptance criteria met outright, 1 with a coverage caveat. This is the closest the code has been to the spec across all four reviews.

### Suggested final polish (none of these blocks merge)

1. Collapse `IdlePromptContext`/`handle_idle_prompt_command` — highest signal-to-noise cleanup left.
2. Move or rename `inject_runtime_event` out of `notifications/dispatch.rs`.
3. Move `dispatch_startup_drain` from `startup.rs` to `turn/queue.rs` (or new `turn/drain.rs`).
4. Bundle actor-owned mutable state into a single `ActorState` struct to kill the 14-`&mut` wiring at handler boundaries.
5. Add one comment on the "delete-on-rejected" semantic in `config/handle.rs:115` and `config/queue.rs:94`.
6. Add at least one characterization test for `apply_pending_config_changes_if_idle` — most subtle behavior in the split, zero coverage.

### Summary

`fc1cfea6` is a solid follow-up. It addressed every regression from pass 3, resolved two of the pass-3 spec-alignment items (idle/busy config + runtime-event dedup, and the live-runtime.md conflict), and introduced no new issues. The remaining deltas (items 3-7 above) are polish, not blockers. I'd be comfortable approving after item 1 lands (the identity-wrapper). The rest can be follow-ups.

anyharness/crates/anyharness-lib/src/live/sessions/connection/types.rs
```diff
@@ -0,0 +1,154 @@
+use crate::sessions::live_config::{LegacyModeOption, LegacyModeState};
+use agent_client_protocol as acp;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SessionStartupStrategy {
```
pablonyx: `SessionStartupStrategy` is still owned by `connection/types.rs`, but it is chosen by `sessions/runtime/startup.rs` and carried through `SessionActorConfig`; it is actor/runtime startup policy, not raw ACP connection mechanics. Keeping it here forces runtime code to import the connection layer just to describe how a live actor should start, which weakens the boundary this split is trying to create. Please move the strategy type to the actor/runtime startup owner, then let `connection/native_session.rs` consume that decision.

anyharness/crates/anyharness-lib/src/live/sessions/actor/event_loop.rs
```diff
@@ -0,0 +1,266 @@
...
+                            &handle,
+                            &store,
+                            &session_id,
+                            action_capabilities,
+                            supports_native_close,
+                        )
+                        .await;
+                        let _ = respond_to.send(result);
+                    }
+                    Some(SessionCommand::CloseNativeSession { native_session_id, respond_to }) => {
```
pablonyx: `CloseNativeSession` is still implemented directly in the top-level event loop. The updated actor spec says fork readiness, fork command handling, and native child-session close policy live under `actor/fork/`, and `event_loop.rs` should stay dispatch-only. Please move this branch behind an actor-side handler, similar to `verify_fork_ready`/`fork_native_session`, so close policy and fork-child lifecycle stay together and the loop remains only orchestration.

pablonyx (commented):
Third pass on updated head `b3d6849`. The earlier comments about actor prelude, mutable startup state, turn split, connection test placement, latency docs, and old-path ratchet are addressed or outdated. Focused checks and full `cargo test -p anyharness-lib` pass locally. I left two new inline ownership comments. The prior max-lines allowlist comment is still active because the unrelated desktop allowlist addition remains.

pablonyx:
## Fifth-pass review — `7745f872`

Small, focused commit. Two substantive changes:

1. **`SessionStartupStrategy` relocated from `connection/types.rs` to `actor/state.rs`.** ✅ Correct spec alignment. Per `specs/session-actor.md` §Collaborators, the startup strategy is actor-owned policy (which session-bring-up shape to use), not a connection-protocol type. `connection/native_session.rs` now imports the enum from the actor module, which is the right direction (actor depends on connection for mechanics, strategy stays on the actor side of that split). Six callers updated.

2. **New `close_native_child_session` wrapper in `fork/handle.rs`.** ⚠️ Worth discussing. The function is a three-line pass-through:

   ```rust
   pub(in crate::live::sessions::actor) async fn close_native_child_session(
       conn: &acp::ClientSideConnection,
       native_session_id: &str,
       supports_close: bool,
   ) -> anyhow::Result<()> {
       close_native_session(conn, native_session_id, supports_close).await
   }
   ```

   `event_loop.rs::SessionCommand::CloseNativeSession` previously called `close_native_session` directly; now it calls `close_native_child_session`, which calls `close_native_session`. Pure forwarding. The commit adds naming clarity ("this command closes a *child* native session, not the actor's own") but doesn't actually guard against misuse — the body just forwards the id unchanged.

   This is the same identity-wrapper pattern I've been flagging on `handle_idle_prompt_command`. Two options that achieve the naming intent without the indirection:

   - **Comment at callsite** in event_loop.rs: `// CloseNativeSession targets a child native session (e.g., created by an earlier Fork), not this actor's own native session. The id is authoritative.` No new function.
   - **Rename-import**: `use crate::live::sessions::connection::shutdown::close_native_session as close_native_child_session;` in event_loop.rs. Same name at the callsite, zero wrapper.

   The wrapper-shaped solution is already in the tree for `handle_idle_prompt_command` and still open (see below). Consistency matters — either both wrappers stay (and the reviewer's "no identity wrappers" critique was wrong) or both should go.

### Gates

- `cargo build -p anyharness-lib`: passes, **zero warnings**.
- `cargo test -p anyharness-lib --lib`: **428 tests pass**.
- `check_anyharness_boundaries.py` / `check_anyharness_old_paths.py` / `check_max_lines.py`: all green.

### Spec-alignment items still open (carried from pass 4)

None of these block merge individually; they're polish. Listed so the delta from "shipping" to "fully aligned with specs/session-actor.md" is visible:

1. `turn/handle.rs::handle_idle_prompt_command` — ~40-line identity wrapper with a 14-field `IdlePromptContext` that duplicates `ActivePromptContext`.
2. `event_loop.rs` Prompt arm wires 14 `&mut`s inline (lines 61-86); no `ActorState` bundle.
3. `dispatch_startup_drain` still in `startup.rs` (lines 324, 342); belongs in `turn/` since it self-sends a `SessionCommand::Prompt`.
4. `startup.rs` at 432 lines, under the hard limit but dense (connection bring-up + ACP init + native session + two live-config emissions + preference restore + drain dispatch).
5. `connection/mod.rs` retains a 4-line mini-prelude for `AgentKind`/`ResolvedAgent`/`to_acp_servers`/`SessionMcpServer`.
6. `inject_runtime_event` lives in `notifications/dispatch.rs` (lines 28-41) but is runtime-event injection, not ACP-notification dispatch.
7. No characterization tests for `handle_active_prompt`, `finish_prompt_result`, `apply_specific_config_option`, `apply_pending_config_changes_if_idle` — the four largest and most behaviorally sensitive functions.
8. No comment explaining the "delete pending config on Rejected" semantic at `config/handle.rs:115` and `config/queue.rs:93-95`.

### Recommendation

The code is merge-ready against the spec's acceptance criteria (12/13 met; the 13th is test coverage). This commit tightened two ownership boundaries — one cleanly, one via a forwarding wrapper that compounds rather than resolves an open pattern.

Minimum fix before merge: either remove the `close_native_child_session` wrapper (rename-import or callsite comment) *or* keep it and remove `handle_idle_prompt_command` + `IdlePromptContext`. Pick a convention and apply it to both sites. Inconsistency between two functionally identical patterns reads worse than either choice alone.

anyharness/crates/anyharness-lib/src/live/sessions/actor/mod.rs
```diff
@@ -0,0 +1,15 @@
+mod background_work;
+pub(crate) mod command;
```
pablonyx: The new actor submodules are crate-visible, which narrows the public surface that used to exist under `acp::session_actor`. The plan says to preserve current visibility while keeping only `SessionCommand` crate-private, but public types such as `PromptAcceptError`, `QueueMutationError`, `SessionActorConfig`, `SessionStartupStrategy`, `ActorReadyResult`, and `PendingSessionActor` no longer have a public replacement path, and the old public `spawn_session_actor` helper is gone entirely. If this visibility reduction is intentional, the migration plan/docs should call it out as an API change; otherwise make the owning modules or narrow re-exports public while keeping only `SessionCommand` itself `pub(crate)`.

pablonyx (commented):
Fourth pass on updated head `7745f87`. The startup-strategy ownership comment is addressed. Focused checks and full `cargo test -p anyharness-lib` pass locally. I left one new inline comment about visibility preservation. The prior max-lines allowlist comment is still active, and the existing `CloseNativeSession` ownership thread is still active on the current diff.

PR #205, opened 2026-05-13
[proliferate] Migrate cowork product MCP

pablonyx:
## Summary

- migrates cowork onto the shared product MCP server structure (`definition/context/tools/calls/auth`)
- moves cowork MCP launch selection and HTTP injection into `sessions/mcp_bindings`
- routes the legacy cowork MCP endpoint through the generic product MCP dispatcher while preserving the legacy header/token path
- updates AnyHarness MCP docs for the current `mcp/` shape

## Validation

- `cargo test -p anyharness-lib mcp`
- `git diff --check`

## Note

- `python3 scripts/check_max_lines.py` still fails on existing `desktop/src/hooks/workspaces/tabs/use-workspace-shell-activation.ts: 692 (max 600)`, unrelated to this AnyHarness branch.


anyharness/crates/anyharness-lib/src/domains/cowork/mcp/tools.rs
```diff
@@ -3,9 +3,22 @@ use serde_json::{json, Value};
 
 use crate::integrations::mcp::tools::tool_definition;
 
-pub(super) const DEFAULT_PROTOCOL_VERSION: &str = "2025-11-25";
-pub(super) const SERVER_NAME: &str = "proliferate-cowork";
-pub(super) const SERVER_INSTRUCTIONS: &str = "Use cowork artifact tools to manage cowork artifacts for this workspace. When workspace delegation is available, use get_coding_workspace_launch_options to choose a source workspace, then create_coding_workspace to provision a normal coding worktree. create_coding_workspace does not start agent work. Use get_coding_session_launch_options for that managed workspace, then create_coding_session to start a linked coding session with a prompt. Set wakeOnCompletion or call schedule_coding_wake when you want this cowork thread prompted after the coding session's next completed turn. Inspect coding work with get_coding_status and read_coding_events.";
+pub const MUTATING_TOOL_NAMES: &[&str] = &[
```
pablonyx, line 6: [P2] `MUTATING_TOOL_NAMES` drives the shared product-MCP route gate now, not just documentation. Including `list_artifacts`, `get_artifact`, launch-option reads, `list_coding_workspaces`, status, and event reads makes those read-only calls acquire `CoworkWrite` and fail `assert_workspace_mutable` before the MCP handler can run. That loses the method-aware behavior the product MCP registry was added for, and it conflicts with the artifact docs expectation that read contexts can list/read. Limit this list to actual writes (`create/update/delete_artifact`, create workspace/session, send message, schedule wake) and add a test that read-only cowork tools do not request the write operation kind.

anyharness/crates/anyharness-lib/src/domains/cowork/mcp/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+        Self {
+            artifact_runtime,
+            cowork_runtime,
+            auth,
+        }
+    }
+}
+
+#[async_trait]
+impl ProductMcpServer for CoworkProductMcpServer {
```
pablonyx, line 43: [P2] The new cowork ProductMcpServer shape only has tool-list tests right now; `cargo test -p anyharness-lib domains::cowork::mcp` runs 3 tests and none cover auth, context resolution, calls, or endpoint compatibility. This migration replaces the old bespoke JSON-RPC endpoint and auth wrapper, so the docs-required coverage should include product-vs-legacy token validation, non-cowork/canonical-session context rejection, delegation-disabled call rejection, at least one artifact call delegation, and the legacy `/cowork/sessions/{session_id}/mcp` route. Without those, the riskiest parts of this migration are only compile-tested.

pablonyx (commented):
Reviewed at `161d1d2e`. Clean migration — cowork now follows the same `mcp/` file grammar as reviews/subagents/workspace_naming, launch-extras moved out of `CoworkSessionHooks` into `sessions/mcp_bindings/{selection, injection, product_catalog}.rs`, legacy compat helpers on the hook are gone, legacy route routed through the generic dispatcher. Tests and doc updates track the code.

The architecture work from PR #202 clearly paid off here: this migration is ~485 add / 457 delete, mostly pure code movement with a very thin diff to shared infra. That's the right shape for a follow-up.

Below are the issues I spotted.

## Must-fix

### 1. Cowork's `MUTATING_TOOL_NAMES` labels 7 read-only tools as mutating

`domains/cowork/mcp/tools.rs:6-21`:

```rust
pub const MUTATING_TOOL_NAMES: &[&str] = &[
    \"create_artifact\",
    \"update_artifact\",
    \"delete_artifact\",
    \"list_artifacts\",            // read
    \"get_artifact\",              // read
    \"get_coding_workspace_launch_options\",   // read
    \"create_coding_workspace\",
    \"list_coding_workspaces\",    // read
    \"get_coding_session_launch_options\",     // read
    \"create_coding_session\",
    \"send_coding_message\",
    \"schedule_coding_wake\",
    \"get_coding_status\",         // read
    \"read_coding_events\",        // read
];
```

Every entry here causes the shared endpoint to acquire a `SharedWorkspaceMutation(CoworkWrite)` lease and assert mutability at `api/http/product_mcp.rs:146-160`. That's the whole point of per-tool gating — contrast `subagents/mcp/tools.rs:7-11` which correctly lists only `create_subagent`, `send_subagent_message`, `schedule_subagent_wake` and leaves `list_subagents` / `read_subagent_events` / `get_subagent_status` out.

Consequences:
- Concurrent `list_artifacts`/`get_artifact`/`read_coding_events` calls now serialize through the cowork write gate unnecessarily.
- A workspace in a non-mutable state (e.g., retiring) rejects every one of these reads as \"WORKSPACE_MUTATION_BLOCKED\" instead of letting them through. Under `assert_workspace_mutable`, this surfaces as a client-visible 409.
- This is a regression from the spirit of per-tool gating, though arguably *not* a regression from pre-migration behavior (pre-migration cowork took `CoworkWrite` on every request too). Still, the PR actively chose to label these, and that's wrong.

Fix is straightforward — keep only the actual mutators:

```rust
pub const MUTATING_TOOL_NAMES: &[&str] = &[
    \"create_artifact\",
    \"update_artifact\",
    \"delete_artifact\",
    \"create_coding_workspace\",
    \"create_coding_session\",
    \"send_coding_message\",
    \"schedule_coding_wake\",
];
```

Note: the existing drift test `mutating_tool_names_are_all_advertised_when_delegation_is_enabled` (`tools.rs:346-352`) only asserts every name appears *somewhere* in the advertised list — it does NOT assert those tools actually mutate. So this test would still pass if someone incorrectly labeled readers. Consider adding a negative assertion that read-only operations like `get_artifact` are NOT in the list, or — better — flip the directionality: declare `MUTATING_TOOL_NAMES` once and fail a test if any advertised tool matching `^(get_|list_|read_)` also appears in the list.

## Should-fix

### 2. Stringly-typed error mapping in `CoworkMcpContext::resolve_context`

`domains/cowork/mcp/context.rs:27-37`:

```rust
fn map_context_error(error: anyhow::Error) -> ProductMcpContextError {
    let message = error.to_string();
    match message.as_str() {
        \"workspace not found\" | \"session not found\" => ProductMcpContextError::not_found(message),
        \"session does not belong to workspace\"
        | \"workspace is not a cowork workspace\"
        | \"session is not the canonical cowork session\"
        | \"cowork thread does not belong to workspace\" => ProductMcpContextError::conflict(message),
        _ => ProductMcpContextError::Internal(error),
    }
}
```

Six exact-string matches on error messages produced elsewhere. If anyone changes one of those strings in `validate_canonical_thread` (or the functions it calls), that error silently downgrades from a typed 404/409 to a 500 Internal, which — given the spec's new soft-vs-hard-fail convention (`product-mcps.md:322-328`) — would look like a transient backend failure rather than a genuine misuse.

Contrast `subagents/mcp/context.rs:14-36` and `workspace_naming/mcp/context.rs:14-56` which construct `ProductMcpContextError::not_found` / `::conflict` at the site where the condition is known, so no string round-tripping.

Fix: return a typed `CoworkContextError` enum from `CoworkRuntime::validate_canonical_thread` (`WorkspaceNotFound`, `SessionNotFound`, `SessionWorkspaceMismatch`, `NotCoworkWorkspace`, `NotCanonicalCoworkSession`, `ThreadWorkspaceMismatch`) and `impl From<CoworkContextError> for ProductMcpContextError`. Ties the 404/409 classification to the code path that knows the condition, not to a string.

### 3. Legacy cowork URL shape still divergent from the other three

`api/router.rs:142-145` keeps `/workspaces/{workspace_id}/cowork/sessions/{session_id}/mcp`, while the other legacy routes are `/workspaces/{workspace_id}/sessions/{session_id}/<product>/mcp`. Cowork's URL has `/cowork/sessions/` in the middle — before `{session_id}`, not after.

This is preserved-from-before behavior (not introduced by this PR), but now all four legacy routes exist to keep in-flight sessions working, and the cowork one is the odd one out. If you're going to end up collapsing the legacy routes via `ProductMcpEndpointRegistration::with_route_aliases` in a follow-up (noted as open follow-up work on PR #202), cowork's divergent shape means it can't ride the same mechanism without a manual route entry. Worth a sentence in the PR description acknowledging the shape difference so the alias-migration work isn't surprised by it.

### 4. `ProductMcpInjectionContext` and `ProductMcpLaunchCatalog::new` now take four concrete auth handles

`sessions/mcp_bindings/injection.rs:16-25`:

```rust
pub struct ProductMcpInjectionContext<'a> {
    pub runtime_base_url: &'a str,
    pub runtime_bearer_token: Option<&'a str>,
    pub review_auth: &'a reviews_mcp::auth::ReviewMcpAuth,
    pub subagent_auth: &'a subagents_mcp::auth::SubagentMcpAuth,
    pub workspace_naming_auth: &'a workspace_naming_mcp::auth::WorkspaceNamingMcpAuth,
    pub cowork_auth: &'a cowork_mcp::auth::CoworkMcpAuth,
    pub workspace: &'a WorkspaceRecord,
    pub session: &'a SessionRecord,
}
```

And `ProductMcpLaunchCatalog::new` is now an 8-arg constructor (`app/mod.rs:231-240`). This is the injection-metadata-on-registry follow-up item from PR #202 concretely getting worse — adding a 5th product MCP means editing this struct and the constructor call.

Not a blocker for this PR since cowork is already the known end of the known internal MCPs, but if `artifacts` / `computer_use` / `browser_use` from `product-mcps.md:54-61` start landing, the right time to consolidate is now rather than after one more product shows up.

Minimum viable consolidation: make `ProductMcpServer` (or a sibling trait) provide an injection method that returns the `SessionMcpServer` config, so `inject_product_mcps` iterates the registry and calls each handler's injection method. Eliminates the per-product `match` and the auth handles in `ProductMcpInjectionContext` — both of which are the product-id-branching spec (`product-mcps.md:144-151`) bans in the shared kit.

## Nits

- **`calls.rs:104-106`** has `name if tools::is_delegation_tool(name) => ...` as a catch-all after the explicit artifact matches. Works, but `is_delegation_tool` is a hand-maintained list of 9 names at `tools.rs:121-133` that can drift from `delegation_tool_definitions`. Add a drift test (or derive the list from `delegation_tool_definitions`).
- **`calls.rs:28-30`**: the `if is_delegation_tool(name) && !ctx.workspace_delegation_enabled` early-bail message says \"cowork workspace delegation is disabled for this thread\". Consistent with spec #322-328 (soft-list + hard-call). Good.
- **`definition.rs:30-39`** keeps `launch_disabled()` checking `ANYHARNESS_DISABLE_COWORK_LAUNCH_EXTRAS` for selection; commit msg suggests this is a compat shim. Worth a comment naming what's still protected by that flag — if it's just the `selected.push(SelectedProductMcp::Cowork)` branch (checking `selection.rs:64-66`), say so. If it's also meant to skip prompt extras, the current `product_mcp_prompt_extras` at `selection.rs:98-105` will still emit them once `SelectedProductMcp::Cowork` is selected, so the skip only works at selection time.
- **`definition.rs:41-55`** inlines 9 prompt strings rather than a single `SYSTEM_PROMPT_APPEND` const like other products. Minor style inconsistency; fine either way.
- **Artifact tools run via `tokio::task::spawn_blocking`** (`calls.rs:37, 55, 73, 89, 98`). The previous `mcp_server/tools.rs` implementation presumably did the same thing; worth confirming this wasn't lost in the move. (I didn't diff the pre-migration implementation side-by-side — PR description says the tools moved but not whether any async/sync boundary changed.)
- **`CoworkMcpContext` carries a full `WorkspaceRecord`**, not just an id — `context.rs:6-10`. Heavier than other products' contexts, but warranted because `calls.rs` uses `ctx.workspace.clone()` when passing to the artifact runtime inside `spawn_blocking`. Fine.

## Remaining cross-PR follow-ups (from PR #202 review)

Unchanged by this PR; ordered roughly by proximity to spec:
- `ProductMcpDefinition` still narrower than spec (`product-mcps.md:157-193`).
- `integrations/mcp/product_server/endpoint.rs` still missing; `dispatch_product_mcp` still lives in `api/http/product_mcp.rs`.
- Legacy routes still hand-written (4 products × 2 methods = 8 delegators); registration alias mechanism exists but unused.
- `MUTATING_TOOL_NAMES` still in `tools.rs`. With cowork, the drift risk is now larger (see #1 above).
- `sessions/**` → `domains/sessions/**` path migration remains the largest layout divergence from the spec.

## Verdict

Recommend: block merge on #1 (real correctness issue — over-labeling reads as writes causes unnecessary contention and wrong 409s under retire flow). Everything else is follow-up.

Overall this is a clean, small, well-structured migration that benefits directly from the architecture PR #202 put in place. The next load-bearing piece of work is the `injection`-on-registry consolidation — after this PR there are four concrete auth handles in the launch catalog, and that's a natural forcing function for the spec's registry shape.

anyharness/crates/anyharness-lib/src/api/router_tests.rs
```diff
@@ -152,6 +152,37 @@ async fn protected_routes_allow_matching_bearer_auth_when_token_is_configured()
     assert_eq!(response.status(), StatusCode::OK);
 }
 
+#[tokio::test]
+async fn legacy_cowork_mcp_route_uses_product_mcp_auth_errors() {
```
pablonyx, line 160: [P2] This exercises the legacy route only when the capability token is missing. The compatibility case this migration needs to protect is the opposite: already-launched cowork sessions calling this legacy URL with `x-cowork-session-token` should still be accepted and dispatched through the product MCP handler. As written, deleting legacy-header support in `read_auth_header` would not fail this test because the missing-token path would still return `COWORK_MCP_UNAUTHORIZED`. Since `initialize` does not resolve product context, this can stay lightweight: mint a legacy `LegacySha256Dot` token for the test `runtime_home`, send it as `x-cowork-session-token` to `/v1/workspaces/{workspace_id}/cowork/sessions/{session_id}/mcp`, and assert a 200 initialize response from `proliferate-cowork`.

docs/anyharness/product-mcps/artifacts.md
```diff
@@ -65,31 +65,34 @@ anyharness-lib/src/domains/cowork/artifacts.rs
...
 ```text
-anyharness-lib/src/domains/cowork/runtime.rs
-  CoworkSessionHooks launch extras
+anyharness-lib/src/sessions/mcp_bindings/selection.rs
+  cowork MCP selection for cowork-surface sessions
   cowork artifact system prompt append
-  cowork MCP server URL/header construction
+  cowork binding summary
+
+anyharness-lib/src/sessions/mcp_bindings/injection.rs
```
pablonyx, line 93: [P3] This current map now says cowork injection is product-token based, but the rest of this same doc still describes the current cowork MCP as only the legacy `/cowork/sessions/{session_id}/mcp` route with `x-cowork-session-token` and legacy sha256-dot auth. After this PR, fresh launches use `/v1/workspaces/{workspace_id}/sessions/{session_id}/mcp/cowork` plus `x-anyharness-product-mcp-token`, while the old route/header are compatibility aliases. Please update the current route/auth sections here (and the older cowork artifact doc if it remains authoritative) so the docs do not steer future artifact extraction work back toward the pre-migration shape.

docs/anyharness/src/cowork-artifacts.md
```diff
@@ -70,7 +70,7 @@ Keep ownership split the same way the rest of the runtime is split:
 - `domains/cowork/artifacts.rs`
   - artifact lifecycle operations
   - create, update, delete, read, list
-- `domains/cowork/mcp.rs`
+- `domains/cowork/mcp/**`
```
pablonyx, line 73: [P3] This path update is right, but the same ownership block still says `domains/cowork/runtime.rs` owns “session startup prompt assembly” and “built-in cowork MCP binding injection” just above this. After this migration those responsibilities moved to `sessions/mcp_bindings/{selection,injection}.rs`, matching the Product MCP docs. Since this file is still marked authoritative for cowork artifact behavior, please move those two bullets out of the runtime entry and point the launch/prompt/MCP injection ownership at `sessions/mcp_bindings` so future artifact extraction work does not follow the old runtime-owned injection shape.

anyharness/crates/anyharness-lib/src/api/router_tests.rs
```diff
@@ -152,6 +156,81 @@ async fn protected_routes_allow_matching_bearer_auth_when_token_is_configured()
...
+    assert_eq!(response.status(), StatusCode::UNAUTHORIZED);
+    let body = to_bytes(response.into_body(), usize::MAX)
+        .await
+        .expect("read response body");
+    let payload: Value = serde_json::from_slice(&body).expect("parse response json");
+    assert_eq!(payload["code"], "COWORK_MCP_UNAUTHORIZED");
+}
+
+#[tokio::test]
+async fn legacy_cowork_mcp_route_accepts_legacy_capability_token() {
```
pablonyx, line 191: [P3] This closes the legacy compatibility gap, but the fresh path that new cowork launches use still lacks an endpoint-level test. The only fresh-route coverage I found is the generic injection test, and it uses `WorkspaceNaming`; this PR does not prove that cowork is registered under the `cowork` route slug and accepts `x-anyharness-product-mcp-token` at `/v1/workspaces/{workspace_id}/sessions/{session_id}/mcp/cowork`. Please add a sibling test that mints through `CoworkMcpAuth`, posts `initialize` to the fresh generic cowork URL with the product header, and asserts the `proliferate-cowork` initialize response. That catches the highest-risk fresh-session wiring failures: missing cowork registry entry, route slug mismatch, or product/legacy header mix-up.

PR #440, opened 2026-06-01
chore(frontend): add report-only structure inventory

pablonyx:
## Summary
- Add `scripts/report_frontend_structure.py` for grouped frontend structure drift reporting.
- Cover documented guardrail categories: raw DOM controls, primitive definitions, component `.ts` files, direct product hook files, nonstandard hook folders, forbidden shared-package imports, and large frontend files.
- Document the report command and future `--strict` mode in the frontend migration tracker.

## Docs referenced
- `docs/README.md`
- `docs/tbd/structure-alignment-coordinator-model.md`
- `docs/tbd/frontend-structure-alignment-migration.md`
- `docs/structures/frontend/README.md`
- `docs/structures/frontend/guides/components.md`
- `docs/structures/frontend/guides/hooks.md`
- `docs/structures/frontend/guides/lib.md`
- `docs/structures/frontend/guides/state.md`
- `docs/structures/frontend/guides/access.md`
- `docs/structures/frontend/guides/styling.md`
- `docs/structures/frontend/packages/README.md`

## Behavior preservation
- Report-only by default; no CI enforcement changes.
- No app/package source violations fixed in this PR.
- Existing `scripts/check_frontend_boundaries.py` behavior is unchanged.

## Verification
- `python3 scripts/report_frontend_structure.py`
- `python3 scripts/report_frontend_structure.py --summary-only`
- `python3 scripts/report_frontend_structure.py --help`
- `python3 -m py_compile scripts/report_frontend_structure.py scripts/check_frontend_boundaries.py`
- `git diff --check`

### DIFF
--- a/scripts/report_frontend_structure.py
+++ b/scripts/report_frontend_structure.py
@@ -59,6 +59,9 @@ COMPONENT_DEFINITION_RES = [
     re.compile(r"\b(?:export\s+)?const\s+([A-Z][A-Za-z0-9_]*)\s*="),
     re.compile(r"\b(?:export\s+)?class\s+([A-Z][A-Za-z0-9_]*)\b"),
 ]
+TYPE_DEFINITION_RE = re.compile(
+    r"^\s*(?:export\s+)?(?:interface|type)\s+([A-Za-z0-9_]+)\b"
+)
 
 PRIMITIVE_EXACT_NAMES = {
     "Button",
@@ -71,6 +74,7 @@ PRIMITIVE_EXACT_NAMES = {
     "Switch",
     "Tabs",
     "Menu",
+    "MenuItem",
     "Popover",
     "Tooltip",
     "Dialog",
@@ -94,6 +98,7 @@ PRIMITIVE_SUFFIXES = (
     "Switch",
     "Tabs",
     "Menu",
+    "MenuItem",
     "Popover",
     "Tooltip",
     "Dialog",
@@ -104,6 +109,8 @@ PRIMITIVE_SUFFIXES = (
     "ScrollArea",
     "Shell",
 )
+
+PRIMITIVE_DEFINITION_CONTEXT_LINES = 14
 
 NATIVE_DOM_CONTRACT_RE = re.compile(
     r"\b("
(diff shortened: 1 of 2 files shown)


pablonyx (commented):
Reviewer A notes: I found one inventory-quality issue that should be fixed before merge because it would misdirect the follow-up primitive workstream, plus a smaller strict-mode output bug.

scripts/report_frontend_structure.py
```diff
@@ -0,0 +1,588 @@
...
+            if line_is_comment(line):
+                continue
+            for pattern in COMPONENT_DEFINITION_RES:
+                match = pattern.search(line)
+                if not match:
+                    continue
+                name = match.group(1)
+                if name in reported_names:
+                    break
+                if is_primitive_like_name(name) and has_native_dom_contract_for(name, text):
```
pablonyx: This condition is too broad for the documented primitive-definition category. Because `is_primitive_like_name()` treats any component ending in `Input`, `Button`, `Menu`, `Modal`, `Shell`, etc. as primitive-like, and `has_native_dom_contract_for()` scans the whole file, the report currently labels product components such as `WebAppShell`, `ChatInput`, `PluginConnectionModal`, and `ComposerModelPickerMenu` as `PRIMITIVE_DEFINITION_OUTSIDE_UI`. That makes the guardrail inventory misleading for follow-up workstreams. Please narrow this to actual generic primitive definitions, or split/rename the report category so product-specific components are not presented as primitive definitions.

scripts/report_frontend_structure.py
```diff
@@ -0,0 +1,588 @@
...
+
+def print_report(violations: list[Violation], *, summary_only: bool) -> None:
+    files = iter_source_files()
+    counts = Counter(violation.rule_id for violation in violations)
+    total = sum(counts.values())
+
+    print("Frontend structure report (report-only by default)")
+    print(f"Scanned {len(files)} non-test frontend source files.")
+    print(
+        "Strict mode is off; use --strict to return a non-zero exit code when "
```
pablonyx: `--strict` still prints “Strict mode is off” because this message is hard-coded. In strict mode the command exits 1 while telling the reader strict mode is off, which is confusing for the future enforcement path. Please pass the strict flag into `print_report()` and render the mode accurately.

scripts/report_frontend_structure.py
pablonyx: P2: This treats the entire file as evidence that every primitive-suffix component is a primitive definition, so a single raw `<button>` or DOM ref can taint unrelated product/app components. The current report already flags entries like `WebAppShell`, `ChatInput`, and `PluginConnectionModal`, which are not necessarily primitive definitions and would mislead the follow-up primitive workstream. Please tie the native DOM evidence to the reported component body/name more tightly, or narrow the suffix heuristic and let `RAW_DOM_CONTROL` cover product components that merely render raw controls.

scripts/report_frontend_structure.py
pablonyx: P3: When `--strict` is passed, the report still prints "Strict mode is off" even though the command exits non-zero. Please pass the strict flag into report printing and emit the accurate mode so CI/manual output is deterministic and not confusing.

pablonyx (commented):
Reviewer B pass: the guardrail is correctly report-only by default, but I found one inventory-quality issue that should be tightened before merge so follow-up frontend swarms do not chase misleading primitive-definition entries.

PR #611, opened 2026-06-10
feat(agents): add catalog probe, collation pipeline, and viewer

pablonyx:
## What

The probe-generated agent catalog, end to end:

- **`anyharness catalog-probe`** — headless ACP probe built on the existing session driver. Per (harness, auth-context): installs/resolves the harness, spawns its ACP adapter with context-injected credentials in isolated config dirs, enumerates models/modes/config options, switches models one-by-one capturing per-model option matrices, and records provenance (ACP `agent_info` attestation + native CLI version).
- **Availability trials** — off-menu model ids are seeded via config and accepted only if a real inference turn succeeds (menu listing is unvalidated by harnesses — proven by a negative control). Accepted ids become available-but-hidden rows.
- **Credential hygiene** — the probe captures-and-scrubs all known credential env vars at startup so spawned agents only ever see what their auth context injects (a sourced secrets file otherwise silently enables every provider in every run).
- **`scripts/agent-catalog/build-catalog.mjs`** — collates raw snapshots into `catalog.draft.json`: observed-set availability (`anyOf` contexts incl. `baseline`, no monotonicity inference), per-model control matrices (with fallback to baseline options / modes block for harnesses that don't re-emit options), matrix-equality invariant across contexts, versioned display names, `defaultVisible` from menu-vs-trial provenance.
- **`render-catalog.mjs`** — self-contained searchable HTML viewer for the draft.

Coverage: **5 harnesses × 9 auth contexts → 11 snapshots, 163 model rows**, all empirically observed.

## Key findings encoded in the data

- Model menus are server-curated **per auth method**: claude API-key vs OAuth get different model sets, defaults, and copy; codex ChatGPT-OAuth exposes 4 models API-key auth doesn't.
- Availability is **not monotone**: an OpenCode free model exists only *without* credentials; an Anthropic key removes it. Hence observed-set semantics.
- Menu ≠ availability: claude runs Opus 4.8 / Fable 5 under API key while advertising neither (trial-verified, hidden rows).
- Per-model variation is real everywhere: haiku has no effort control; cursor encodes matrices inside model ids; gemini has no config options at all.

## Notes / follow-ups

- This is merge-order (a); the catalog/registry consumption + version-reconcile overhaul stacks on top of it.
- Cursor probes under machine login (`cursor-login`) — its ACP ignores `CURSOR_API_KEY` and keeps auth in the macOS Keychain.
- Claude `anthropic-oauth` here is the `setup-token` surface, which differs from interactive-login state.
- TODO: parse cursor's `id[key=value]` variant notation into per-model controls at build time.


pablonyx:
@greptile

PR #610, opened 2026-06-10
refactor(mcp): collapse sub-threshold product MCP files

pablonyx:
## Summary

- Merges `dispatcher.rs` + `request.rs` + `response.rs` + `errors.rs` into `server.rs` in `integrations/mcp/product_server/` — all four were sub-threshold fragments of one concept; public re-export surface in `mod.rs` is unchanged, zero call-site churn
- Merges `selection.rs` + `injection.rs` into `product_catalog.rs` in `domains/sessions/mcp_bindings/` — selection was a trivial loop over extension points (not policy), injection was a URL template + two headers; merged file is ~450 lines including absorbed tests
- Drops `ProductMcpLaunchSelector` and `ProductMcpCapabilityTokenMinter` blanket-impl traits from `product_launch.rs` — both had zero named implementors; call sites already passed closures and are unchanged
- Amends all spec guides that named the deleted files (`domains.md`, `api.md`, `app.md`, and others found by grep)

All four changes address the same root cause: swarm-generated ceremony in commit `880d9174d` that applied file-per-concept mechanically without proportionality judgment.

Arc<dyn> trait simplification folded in (not deferred) — same swarm-generated ceremony, same module, S effort, near-zero blast radius.

**Out-of-scope and deferred:** `assembly.rs` (475 lines) and `product_registry.rs` (455 lines) approaching split thresholds.

## Test plan

- [x] `cargo check --workspace` — clean
- [x] `cargo test -p anyharness-lib` — 614 passed, 0 failed
- [x] `check_max_lines.py` — pass
- [x] `check_anyharness_boundaries.py` — pass
- [x] `check_anyharness_old_paths.py` — pass
- [x] No dangling `mcp_bindings::selection` / `mcp_bindings::injection` imports
- [x] No dangling `product_server::dispatcher` / `request` / `response` / `errors` imports
- [x] `ProductMcpLaunchSelector` / `ProductMcpCapabilityTokenMinter` fully removed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #613, opened 2026-06-10
refactor(mcp): purge legacy routes, headers, and deprecated tool params

pablonyx:
## Summary

- Delete 4 legacy product MCP URL routes (`/sessions/{id}/subagents/mcp`, `/sessions/{id}/reviews/mcp`, `/sessions/{id}/workspace-naming/mcp`, `/cowork/sessions/{id}/mcp`) and their 8 handler functions
- Delete `legacy_route_aliases` module and `with_route_aliases` constructor from `ProductMcpEndpointRegistration`
- Remove legacy capability header validation (`x-cowork-session-token`, `x-review-session-token`, `x-subagent-session-token`, `x-workspace-naming-session-token`) from `ProductMcpAuth` and all domain auth implementations — only `x-anyharness-product-mcp-token` is accepted now
- Remove deprecated tool params (`workspaceId`, `codingSessionId`, `childSessionId`, `reviewRunId`) from cowork, subagent, and review tool schemas and call handlers
- Drop `legacy_header_names()` from `ProductMcpServer` trait and all implementations

Stacked on #610.

## In-flight session safety

The route-404 and silently-dropped-param concerns raised in review are moot under the deploy story: product MCP servers are in-process HTTP endpoints on AnyHarness itself (`127.0.0.1`), and agent processes are spawned and owned by the AnyHarness live session manager. A binary update restarts AnyHarness, which terminates every live agent process with it — no agent carrying old injected MCP URLs or old tool schemas can survive the deploy. Resumed sessions re-inject MCP server configs and re-fetch `tools/list` from current code on relaunch, so the first post-deploy tool call always sees the new routes and schemas.

The legacy headers and routes are also purely agent-facing: they are only ever constructed by `product_catalog.rs` injection. A repo-wide grep of desktop, mobile, web, and server code confirms no consumer hand-constructs these headers or URLs, so no out-of-repo caller can break.

## Test plan

- [x] `cargo check --workspace` — clean
- [x] `cargo test -p anyharness-lib` — 609 passed, 0 failed
- [x] `check_max_lines.py` — pass (router_tests.rs allowlist updated: 1179→1102)
- [x] `check_anyharness_boundaries.py` — pass
- [x] `check_anyharness_old_paths.py` — pass
- [x] No remaining `legacy_route_aliases`, `LEGACY_CAPABILITY_HEADER_NAME`, or `legacy_header_names` refs in non-test source
- [x] No legacy header/route references in desktop/mobile/web/server code

**Known accepted residual risk:** resumed sessions replay conversation history, and a model can imitate a deprecated param it sees in its own old tool calls even after re-fetching fresh schemas — those args now drop silently. The magnitude is small (the primary params were always preferred, and the tail decays with session age); accepted as-is per review.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/api/http/product_mcp.rs
+++ b/anyharness/crates/anyharness-lib/src/api/http/product_mcp.rs
@@ -9,9 +9,6 @@ use serde_json::Value;
 use super::access::assert_workspace_mutable;
 use super::error::ApiError;
 use crate::app::AppState;
-use crate::domains::sessions::mcp_bindings::product_registry::{
-    legacy_route_aliases, ProductMcpEndpointHandler,
-};
 use crate::integrations::mcp::product_server::{
     ProductMcpAuthHeader, ProductMcpContextError, ProductMcpDispatchError,
     ProductMcpEndpointOperation, ProductMcpRequestContext, PRODUCT_MCP_TOKEN_HEADER_NAME,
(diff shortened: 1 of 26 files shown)


pablonyx (commented):
Review against `guides/mental-model.md` (the purge is the right instinct — "delete dead runtime code instead of preserving parallel implementations" — but there's one silent-behavior risk to settle before merge):

**Aligned:**
- Net -776 lines, one capability header instead of five (`x-anyharness-product-mcp-token` only) — the one-mechanism rule applied to auth headers.
- `legacy_route_aliases` + `with_route_aliases` + `legacy_header_names()` removed root-and-branch rather than deprecated-in-place; allowlist updated downward; clean stack on #610.

**The real concern — in-flight sessions don't fail, they silently change behavior:**
Tool schemas live in running agents' LLM context. A cowork parent session launched before this deploys still believes `workspaceId`, `codingSessionId`, and `agentKind` are valid params — and because the fields were *removed from the serde structs* (permissive deserialization), those args are now silently dropped rather than rejected:

- `create_coding_session`: an old agent passing `agentKind` (without `harnessId`) now creates with `harness_id = None` — different harness than the agent asked for, no error.
- `send_coding_message` / target resolution: an old agent disambiguating via `codingSessionId` now passes `None` to `resolve_coding_session_target` — potentially targeting the wrong/default child session, no error.
- `get_coding_session_launch_options` / `create`: `workspaceId` targeting now resolves against `None`.

A 4xx would be self-correcting (the agent retries differently); a silently-ignored param mid-session is not. Two acceptable resolutions: **(a)** state in the PR body that the deploy story makes this moot (live cowork parents are drained/restarted on deploy, or schema-bearing sessions predating the new schemas are already gone), or **(b)** keep accepting the deprecated fields for one release behind a `tracing::warn!` deprecation counter, and purge once the counter reads zero. Either is fine — the current state is the only wrong one, because the risk is undocumented.

**Same question, smaller, for the deleted routes:** injected MCP URLs are baked into a live agent process at launch and only refresh on resume (re-injection). Any session launched before route unification and still live across this deploy 404s on its next `tools/call`. If unification shipped long enough ago that no such sessions exist, one sentence in the PR body saying so closes the question.

**Minor:** the test plan greps this repo's Rust for symbol refs — worth one sentence confirming the legacy headers/routes are purely agent-facing (injected by `product_catalog.rs`) and never hand-constructed by desktop/cloud/SDK consumers, since those wouldn't show up in a symbol grep.

pablonyx:
Review feedback addressed in c5b5bb1ec (+ rebase onto updated #610):

- **Greptile (magic string in injection.rs test):** resolved by deletion — the negative assertion was vacuous once the legacy header system was gone, and was dropped entirely when the test moved into `product_catalog.rs` during the stack rebase.
- **Greptile (lost param descriptions):** restored clean descriptions on `coworkWorkspaceId`, `coworkAgentId`, `subagentId`, and `reviewId` (18 occurrences) without the deprecated-alias references.
- **In-flight session risk (blocking):** resolved via option (a) — deploy story documented in the PR body. Product MCP servers are in-process endpoints and agent processes are owned by the AnyHarness live session manager, so a binary update terminates every live agent; nothing carrying old URLs or schemas survives the deploy. Resume re-injects configs and re-fetches `tools/list` from current code.
- **Agent-facing confirmation (minor):** repo-wide grep of desktop/mobile/web/server confirms no consumer hand-constructs the legacy headers or routes; they were only ever emitted by `product_catalog.rs` injection. Added to the test plan.

pablonyx (commented):
Re-reviewed c5b5bb1ec + the rebase — verified in the diff and PR body:

- Param descriptions restored cleanly ("Stable cowork workspace target id.") with no deprecated-alias references.
- The "In-flight session safety" section resolves the blocking concern, and the mechanics check out: agents are children of the live session manager, so a binary update takes every old-schema agent down with it, and resume re-injects configs + re-fetches `tools/list` from current code. Notably, even the imperfect case degrades *loudly*: a hypothetical stray agent process hitting deleted routes gets a 404 tool error (visible, self-correcting), not silent mistargeting — which was the actual worry.
- Agent-facing-only confirmation added to the test plan.

One residual tail worth a sentence, non-blocking: resumed sessions replay conversation history, and a model can imitate a deprecated param it sees in its own old tool calls even with fresh schemas — those args now drop silently. The magnitude is small (the primary params were always preferred, and the tail decays with session age), so I'd ship as-is; just flagging it so it's a known accepted risk rather than an unknown one.

Resolved from my side.

PR #644, opened 2026-06-12
fix(web): explain unreachable API on first-run auth

### LINKED ISSUE #626: Web first-run auth loading should explain when the local API is unavailable
## Problem

When running the hosted web app locally without the Proliferate API running, the first-run experience is confusing.

Repro:

1. Run:
   ```bash
   pnpm install
   pnpm --filter @proliferate/web dev -- --host 127.0.0.1 --port 5173

2. Open:
http://127.0.0.1:5173/

3. Do not run the API server at http://localhost:8000.
The app enters the auth bootstrap flow and shows:

Loading account
Checking your cloud session.
A new contributor cannot tell whether the app is still loading, broken, or missing the backend.

Expected Behavior
When the local API is unavailable, the web app should stop waiting after a short timeout and show an actionable local-dev message, for example:

Local API is not reachable at http://localhost:8000.
Start the Proliferate API or set VITE_PROLIFERATE_API_BASE_URL before signing in.
Why This Matters
This is a first-run onboarding issue. The web app can start successfully, but the first visible state does not tell contributors that cloud sign-in requires the backend API.

Suggested Fix
Add a bounded timeout/abort around bootstrapWebSession in apps/web/src/providers/WebCloudProvider.tsx.
Track whether the failure looks like a connection/timeout failure.
Surface a dev-only notice in apps/web/src/components/auth/screen/AuthScreen.tsx.
Keep production behavior unchanged.

sushantlokhande14:
## What & why
The hosted web app showed an indefinite "Checking your cloud session" state
when the local Proliferate API wasn't running, leaving developers unsure
whether the app was loading, broken, or missing its backend.

## Changes
- Abort the session bootstrap after a bounded 8s timeout so the sign-in
  screen renders instead of hanging.
- Track whether the bootstrap failed because the API was unreachable
  (timeout / connection error) vs. a genuine auth response (e.g. 401).
- In development only, show a notice on the auth screen explaining the API
  could not be reached at `http://localhost:8000` and how to fix it (start
  the API or set `VITE_PROLIFERATE_API_BASE_URL`).
- Production behaviour is unchanged — the notice is gated behind `import.meta.env.DEV`.

## Testing
- `pnpm web:typecheck` ✅
- `pnpm shared:typecheck` ✅
- `pnpm --filter @proliferate/product-ui test` ✅ (42 passed)

Closes #626

> Note: I don't have permission to set labels. Please add `release:patch` and `area:web` (or the labels you prefer) so CI can proceed.

### DIFF
--- a/apps/packages/product-ui/src/auth/AuthStartPanel.tsx
+++ b/apps/packages/product-ui/src/auth/AuthStartPanel.tsx
@@ -19,6 +19,7 @@ interface AuthStartPanelProps {
   footer: ReactNode;
   providers: AuthProviderActionView[];
   credentialForm?: ReactNode;
+  notice?: ReactNode;
   note?: ReactNode;
   error?: ReactNode;
   devAccess?: ReactNode;
@@ -31,12 +32,21 @@ export function AuthStartPanel({
   footer,
   providers,
   credentialForm,
+  notice,
   note,
   error,
   devAccess,
 }: AuthStartPanelProps) {
   return (
     <AuthLayout mark={mark} title={title} subtitle={subtitle} footer={footer}>
+      {notice ? (
+        <div
+          className="rounded-lg border border-border bg-muted/40 px-3 py-2 text-sm leading-5 text-muted-foreground"
+          role="status"
+        >
+          {notice}
+        </div>
+      ) : null}
       {credentialForm}
       {credentialForm ? <AuthDivider /> : null}
       {providers.map((provider) => (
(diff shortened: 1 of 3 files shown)


pablonyx:
Thanks a lot for this, @sushantlokhande14 — really appreciate you tackling #626. We ended up landing #657 for the same issue, which took a near-identical approach (dev-gated bootstrap timeout + an "API unreachable" notice). The deciding factor was production safety: #657 gates the abort timeout behind `import.meta.env.DEV`, whereas this PR's timeout runs in production too, which could abort bootstrap for real users on slow (>8s) connections — and it keeps the dev-only logic out of the shared `product-ui` `AuthStartPanel`.

Closing as a duplicate of #657 — but this was a solid, well-reasoned implementation and we'd love to have your next PR. 🙏

PR #657, opened 2026-06-12
fix(web): explain when the local API is unreachable during first-run auth

### LINKED ISSUE #626: Web first-run auth loading should explain when the local API is unavailable

utsavp257:
## Summary

- Fixes #626: running the hosted web app locally without the Proliferate API left first-run stuck on "Loading account / Checking your cloud session." with no explanation.
- In dev, the session bootstrap in `WebCloudProvider` is now bounded by a 5s `AbortController` timeout, and connection-level failures (connection refused, DNS, abort) are classified separately from HTTP error responses via a new `isApiUnreachableError` helper in `lib/access/cloud`.
- When the bootstrap fails because the API never answered, the sign-in screen shows a dev-only notice: "Local API is not reachable at `<api base url>`. Start the Proliferate API or set `VITE_PROLIFERATE_API_BASE_URL` before signing in."
- Production behavior is unchanged: the timeout and the unreachable flag are both gated on `import.meta.env.DEV`, and HTTP error responses (e.g. 401) still land on the normal sign-in screen with no notice.

## PR Metadata

- Title format: `<type>(<scope>): <plain-English change>`
- Required: exactly one `release:*` label → `release:fix`
- Required: at least one `area:*` label → `area:product`

(I can't set labels as an outside contributor — would appreciate a maintainer adding `release:fix` + `area:product`.)

## Verification

- `pnpm --filter @proliferate/web test` — 23 passed, including 3 new tests for the failure classifier.
- `pnpm web:typecheck` — clean.
- `python3 scripts/check_frontend_boundaries.py` — passed.
- Manually reproduced the issue: ran `vite` on `127.0.0.1:5173` with no API at `localhost:8000`. The loading screen shows briefly, then after ~5s the sign-in screen renders with the notice in the alert slot.

*Draft until a maintainer adds `release:fix` + `area:product` (outside contributors cannot set labels).*

### DIFF
--- a/apps/web/src/components/auth/screen/AuthScreen.tsx
+++ b/apps/web/src/components/auth/screen/AuthScreen.tsx
@@ -20,7 +20,7 @@ import { createWebCloudClient } from "../../../lib/access/cloud/client";
 import { useAuthToken } from "../../../providers/WebCloudProvider";
 
 export function AuthScreen() {
-  const { setToken, setSession } = useAuthToken();
+  const { setToken, setSession, bootstrapUnreachable } = useAuthToken();
   const [email, setEmail] = useState("");
   const [password, setPassword] = useState("");
   const [manualToken, setManualToken] = useState("");
@@ -98,7 +98,7 @@ export function AuthScreen() {
         onClick: () => void signIn(provider),
       }))}
       note={AUTH_SIGN_IN_COPY.note}
-      error={providerError}
+      error={providerError ?? (bootstrapUnreachable ? localApiUnreachableNotice() : null)}
       devAccess={webEnv.devAccessTokenLogin ? (
         <div className="mt-2 border-t border-border pt-4">
           {showDevAccess ? (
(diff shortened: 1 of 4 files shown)


utsavp257:
Converted to draft since outside contributors can't set labels and the metadata check requires them. Maintainers: this needs `release:fix` + `area:product` — happy to mark ready for review once those are added (or feel free to flip it yourselves).

PR #662, opened 2026-06-13
fix(desktop): snappier chat input — optimistic model switch + instant transcript reflow

pablonyx:
## Context

Three chat-composer papercuts, all "the UI is correct but arrives late":

1. **Model changes aren't instant during an optimistic new session** — right after launching a chat (before the harness connects), switching model in the picker left the label showing the old model, with no pending indicator, until connect (often 1s+).
2. **The picker should reflect the change optimistically** instead of waiting on the round-trip.
3. **The transcript lags composer height changes** — adding/removing a line resized the textarea instantly, but the transcript above shifted ~90–110ms later.

## Root causes

- **(1)+(2) — config-id key mismatch.** A model switch in a pending session enqueues a config intent keyed `"model"` (`handleLaunchSelect` → `scopedCurrentModelConfigId ?? "model"`). But both readers looked the change up by the *live* model control's config id, which is null until `liveConfig` arrives, so `getPendingSessionConfigChange` returned null and the label + spinner stayed stale. (For connected sessions the real `rawConfigId` is `"model"` and this already worked — verified against the server's `ACP_MODEL_COMPAT_CONFIG_ID`.)
- **(3) — pure 90ms debounce.** The textarea autosizes synchronously in a `useLayoutEffect`, but `useChatDockInset` debounced every `ResizeObserver` fire by `CHAT_DOCK_RESIZE_SETTLE_MS = 90` before measuring → `setMetrics` → `stickyBottomInsetPx` → the transcript spacer. There is no CSS height transition on the composer/dock, so the 90ms was just a debounce and the entire source of the lag.

## Changes

**Fix A — optimistic model selection (issues 1 & 2)**
- `pending-config.ts`: new shared `DEFAULT_MODEL_CONFIG_ID = "model"` constant (mirrors the server's `ACP_MODEL_COMPAT_CONFIG_ID`; single-sources the magic key).
- `use-active-session-config-state.ts`: `pendingModelId` prefers the live config id, falls back to `DEFAULT_MODEL_CONFIG_ID` → the picker **label** updates immediately.
- `use-chat-model-selector-state.ts`: same fallback for `pendingModelChange` → the **queued indicator** shows during the optimistic window too.
- `use-chat-launch-actions.ts`: write side uses the shared constant (keeps read/write keys in sync).

No revert logic needed: the pending entry is derived from the intent and clears automatically on intent terminalization, so there's no stuck spinner or flicker at the pending→connected transition. Connected-session and home-screen paths are unchanged (the fallback only differs when no live model control exists yet).

**Fix B — instant transcript reflow (issue 3)**
- `use-chat-dock-inset.ts`: dropped the 90ms `setTimeout` debounce; dock measurement now coalesces per animation frame (rAF). Transcript tracks input height within ~1 frame. `measure()` already self-dedupes, and the spacer isn't inside any observed element, so there's no resize-loop risk.

## Verification
- `@proliferate/product-domain` typecheck — clean; tests — **370 passed**
- desktop `tsc --noEmit` (after building all workspace deps) — clean, 0 errors
- targeted desktop vitest (active-session selectors, workspace-entry flow/actions, model-selector-current, composer autosize) — **17 passed**

Manual (recommended): launch `pdev`, (A) launch a chat and switch model before it connects → label flips instantly with a queued indicator; (B) wrap/unwrap a line in the input → transcript shifts in lockstep, stays pinned to bottom.

### DIFF
--- a/apps/desktop/src/hooks/chat/derived/use-active-session-config-state.ts
+++ b/apps/desktop/src/hooks/chat/derived/use-active-session-config-state.ts
@@ -1,4 +1,4 @@
-import { getPendingSessionConfigChange, type PendingSessionConfigChanges } from "@proliferate/product-domain/sessions/pending-config";
+import { DEFAULT_MODEL_CONFIG_ID, getPendingSessionConfigChange, type PendingSessionConfigChanges } from "@proliferate/product-domain/sessions/pending-config";
 import {
   pendingConfigChangesForSessionIntents,
 } from "@proliferate/product-domain/sessions/intents/session-intent-selectors";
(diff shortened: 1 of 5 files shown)


pablonyx:

PR #705, opened 2026-06-16
feat(agents): add Grok (xAI Grok Build) as a harness

pablonyx:
## What

Adds **Grok** (xAI's Grok Build) as a first-class agent harness, end-to-end: it installs, auth-detects, and launches through anyharness, and renders cleanly in the desktop picker.

Grok speaks ACP natively (no adapter, no fork) and is already in the upstream ACP registry as `grok-build`, so onboarding is `registry_backed` — the same shape as Gemini/Cursor.

## Backend (anyharness)
- `AgentKind::Grok` + `as_str`/`display_name`/`all`/`parse`
- `registry.json` entry: `registry_backed` → `grok-build` (npx `@xai-official/grok agent stdio`), binary-hint fallback to a local `grok`; `xai` auth slot (`XAI_API_KEY` / `GROK_API_KEY`), `syncedFiles` `~/.grok/auth.json`, `grok login`; schema `kind`/`discovery` enums
- xAI credential discovery mirroring Codex: `ProviderId::Xai` + new `grok.rs` (reads `~/.grok/auth.json`), wired through `lib.rs` / `facts.rs` / `credentials.rs` / `projection.rs`; `"xai"` added to the provider allowlist
- worker inventory + cowork default mode (`bypassPermissions`)

## Desktop
- `grok` added to `CLOUD_AGENT_KIND_ORDER` + `cloudAgentKindLabel` ("Grok")
- `CloudAgentKind` unions/guards (cloud-sdk `generated.ts` / `agent-auth.ts`, desktop `credentials.ts`)
- `SESSION_CONTROL_PRESENTATIONS` grok entry (Claude-equivalent permission modes); the provider icon already existed

## Verification
- `cargo check` (4 crates) + `cargo test` green: credential-discovery **35/35** (incl. grok detector), agents **173/173** (incl. grok projection + registry validation), catalog pairing **55/55**
- Live `GET /v1/agents` returns grok as `ready`
- product-domain vitest **371/371** + typecheck green; desktop vitest green

## Follow-ups (not in this PR)
- **Catalog probe** (real model list + controls): wiring + a `XAI_API_KEY` run of `anyharness catalog-probe --agent grok --auth-context xai-api` → `catalog.json`. Until then grok has no catalog entry (tolerated — catalog lags registry by design).
- Desktop **auth-management** surfaces (settings pane, env-var credential sync, automations) still list the older agents; grok works today via an env key or `grok login`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #715, opened 2026-06-16
fix(live): restore live model switching for Gemini via session/set_model

pablonyx:
## Problem

Switching models mid-session is broken for **Gemini** (works for Claude). Creating a new workspace with a chosen Gemini model works — only the **live switch** fails.

ACP **0.14 removed the typed `set_session_model` RPC**, and the harness migrated to "model is a config option." Gemini CLI exposes **no `model` config option** (it still implements `unstable_setSessionModel` / `session/set_model` and reports models via the legacy block the 0.14 client no longer reads). So for a live Gemini session: no config-option target + the direct setter was **stubbed out** ⇒ the switch is rejected (*"Value '…' is not valid for config option 'model'"*). Creation works because the model is chosen at `session/new` via `createField: modelId`, a separate path.

Gemini installs from upstream npm in prod, so the fix must live in our harness — which is what `specs/.../harnesses/gemini.md` already prescribes ("route it to ACP `unstable_setSessionModel`").

## Fix

`live/sessions/actor/config/apply.rs`:

- **`apply_model_via_direct_setter`** — un-stubbed. Sends Gemini's surviving `session/set_model` handler through ACP's **extension-method channel** (`AgentRequest::ExtMethodRequest`, params `{ sessionId, modelId }`), mirroring the `apply_mode_via_direct_setter_legacy` twin directly below it. The method string serializes verbatim, so the `_`-prefix rule (which only gates *inbound* ext routing) doesn't apply outbound.
- **`should_apply_model_via_direct_setter`** — re-enabled to `!has_direct_model_control()`: engages only when the harness reports no live model control (exactly Gemini under 0.14), leaving live-model-list harnesses to enforce membership upstream. The agent is the sole authority; a rejection surfaces cleanly (session Idle, connection intact).

No caller-chain edits — the routing already funnels the no-config-option model case into the direct setter; this just makes it act again.

Wire method (`session/set_model`) and param shape (`{ sessionId, modelId }`) verified empirically against `@agentclientprotocol/sdk@0.16.1`.

## Testing

- `cargo check` clean; **51 config + 35 actor tests pass**.
- Updated the obsolete "permanently disabled" test → `direct_model_setter_engages_only_without_live_model_control` (asserts both gate branches).
- Preserved `model_config_request_without_raw_option_rejects_values_outside_acp_models` still passes — confirms the narrow gate doesn't regress live-model-list harnesses.

## Not yet covered

Live end-to-end (the actual wire round-trip to a running Gemini agent) — build `~/gemini-cli/bundle`, start a Gemini session, switch model in the picker, confirm in-place switch (no fork) + next prompt runs on the new model.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/live/sessions/actor/config/apply.rs
+++ b/anyharness/crates/anyharness-lib/src/live/sessions/actor/config/apply.rs
@@ -473,9 +473,34 @@ pub(in crate::live::sessions::actor) async fn apply_model_via_direct_setter(
         }))?
         .into();
     let ext = acp::schema::ExtRequest::new(ACP_SET_SESSION_MODEL_METHOD, params);
-    conn.send_request(acp::AgentRequest::ExtMethodRequest(ext))
+    tracing::info!(
+        method = ACP_SET_SESSION_MODEL_METHOD,
+        native_session_id,
+        model_id = desired_model_id,
+        "[model-switch] sending session/set_model"
+    );
+    let response = match conn
+        .send_request(acp::AgentRequest::ExtMethodRequest(ext))
         .block_task()
-        .await?;
+        .await
+    {
+        Ok(response) => response,
+        Err(error) => {
+            tracing::warn!(
+                native_session_id,
+                model_id = desired_model_id,
+                error = %error,
+                "[model-switch] agent rejected session/set_model"
+            );
+            return Err(error.into());
+        }
+    };
+    tracing::info!(
+        native_session_id,
+        model_id = desired_model_id,
+        response = %response,
+        "[model-switch] agent accepted session/set_model"
+    );
 
     startup_state.current_model_id = Some(desired_model_id.to_string());
     set_select_option_current_value_for_purpose(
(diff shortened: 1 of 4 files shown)


pablonyx:
## Accounted for Grok (verified live)

Grok (#705, now on main; branch rebased to include it) is structurally **gemini-shaped**: `modelSource: initMetaModelState`, no model config option, `switchVia: setSessionModel`. The live driver reads model state only from `config_options` ([types.rs:56](../blob/main/anyharness/crates/anyharness-lib/src/live/sessions/driver/types.rs)), which grok lacks, so `has_direct_model_control()` is `false` and the `!has_direct_model_control()` gate routes grok through `apply_model_via_direct_setter` — exactly this PR's path.

**Live test (real `~/.grok/auth.json`, grok 0.2.54):**
- ✅ Grok **speaks `session/set_model`** — the method is recognized (not `method_not_found`), so this PR mechanically covers grok where switching was previously dead.
- ⚠️ Grok rejects the catalog's model ids with `"unknown model id"`. Root cause is **not** this PR: an **onboarded** grok exposes different curated presets than the **clean-probe** catalog ids (documented in `specs/codebase/structures/anyharness/harnesses/grok.md`). That's a catalog↔account mismatch.
- On that rejection the request falls into the **relaunch fallback** (config.rs) — it relaunches (no 500; grok's `load_native` works) but the model stays unchanged since the id is rejected again. This is the **pre-existing, already-tracked** relaunch-fallback follow-up, not introduced here.

Net: this PR is a strict improvement for grok (attempts the correct live method). The grok model-id mismatch + the relaunch fallback are separate follow-ups.

pablonyx:
@greptile

PR #723, opened 2026-06-16
feat(anyharness): the catalog is the lockfile — fenced, sha-verified agent installer

pablonyx:
Makes the agent **catalog the lockfile** and the installer a **fenced total-function of a pin**: every agent install — native CLI and ACP adapter, desktop seed and cloud runtime — materializes EXACTLY a resolved catalog pin, sha-verified, with no install-time latest-resolution.

Rebased onto latest main; **Grok (#705) is integrated** — it installs through the same fenced npm pin as the other registry-backed adapters.

## What it does
- **Pin schema (the lockfile):** each harness role carries a resolved `source` — per-platform `{url, sha256}` for binary/archive, or a pinned npm/git specifier — plus the ACP-mode launch args. `catalogs/agents/catalog.json` is the bundled, sha-anchored truth.
- **Fenced installer:** downloads exactly the pin's per-platform url, **verifies sha256 before use**, refuses on mismatch (nothing left on disk). `require_source` makes a missing pin a hard error. No latest-fetch, PATH adoption, ACP `/latest` re-fetch, or fallback.
- **Producer (`scripts/agent-catalog/resolve-pins.mjs`):** resolves each harness from the registry's discovery config + the ACP registry, downloads + checksums, and freezes the pin. **Wired into `catalog-probe.yml`** (resolve the draft, reuse prior shas, promote to the bundled catalog) so a probe run can't drift the lockfile. `make catalog-pin` mirrors it.
- **Seed + runtime use the same pins:** the desktop seed build (`install-agents`) reconciles against the bundled catalog (claude 2.1.178, codex 0.140.0, …).
- **Deletes the legacy paths:** `native.rs` (latest-fetch + PATH adoption), the agent_process registry/fallback path, and the whole `integrations/agent_cli/acp_registry` `/latest`-fetch module. Net large deletion.

## Integrity anchors (honest, per kind)
binary/archive → sha256 enforced at download · npm → pinned `pkg@version` + npm's registry integrity (recorded provenance; validation requires it be present) · git → full commit SHA (validation rejects mutable refs).

## Verification
- 708 lib + 15 binary tests green; full workspace builds (post-rebase).
- Hermetic tests: correct-sha installs / wrong-sha refused; adapter launcher bakes ACP args (not session default_args); **cursor multi-file bundle preserved** (regression guard); bundled catalog is a complete lockfile across all shipped platforms.
- Stale-symbol sweep over the installer is clean — no `resolve_from_registry` / `latest_version_url` / PATH adoption remain.
- Three review rounds (spec-grounded subagents); all blockers/should-fixes addressed (see comments).

## Known follow-ups (not blockers)
Desktop boot-reconcile ("pin is law" on app upgrade — cloud path already wired); npm independent byte-verification; deploy invariant that a network-reachable runtime always sets a bearer token.


pablonyx:
## 🤖 Automated review — spec conformance

Reviewed against `agent-catalog-readiness.md`, `agent-seeds.md`, and the anyharness `README.md`. Net: the lockfile/fence design is sound and the module boundaries are respected better than I expected. `anyharness-lib` compiles clean (only 2 pre-existing unrelated warnings). Findings below, severity-tagged.

### Blockers
None.

### Should-fix

1. **npm/git pins carry a `sha256` that is never verified — the trust-boundary claim is only true for binary/archive.**
   The spec's Trust Boundary says install "verifies the **sha256 before use**" and calls the sha "the trust anchor." That holds for `Binary`/`Archive` (`pinned.rs` → `*_verified` → `verify_sha256`). But for `Npm`, the `sha256` field (e.g. gemini's `sha512-…` integrity in `catalog.json`) is destructured away and dropped: `installer/pinned.rs:126` (`ResolvedPinSource::Npm { package, args, .. }`) and `install_managed_npm_package` (`installer/npm.rs:14`) has **no integrity parameter at all**. `Git` carries no sha. So for npm/git the anchor is the *specifier*, and npm's own registry integrity — not the catalog pin. The amended spec arguably anticipates this ("…or a pinned npm/git **specifier**"), but the unconditional "verifies the sha256 before use" sentence overstates what the code does. Either (a) thread `sha256` into the npm install and verify the resolved tarball, or (b) tighten the spec to say the sha is the anchor only for binary/archive and the pinned specifier is the anchor for npm/git. Right now code and spec disagree.

2. **Sibling specs now describe deleted code — stale-symbol sweep was run over code but not specs.** Acceptance asks for a stale-symbol search returning "no code references"; that passed for code, but two specs still document the deleted `integrations/agent_cli/acp_registry` module and the deleted install strategy:
   - `specs/codebase/structures/anyharness/guides/integrations.md:368,379-380` — still lists `acp_registry.rs` in the file tree and "ACP registry wire schema, fetch/parse/resolve distribution metadata."
   - `specs/codebase/structures/anyharness/src/agents.md:195,204,251-253,428` — "ACP-registry-backed installs try the ACP registry first," "latest version is only knowable by consulting the ACP registry," and points at `integrations/agent_cli/acp_registry.rs` by path.
   - Even the doc you *did* edit is half-updated: `agent-catalog-readiness.md:358` still lists "ACP registry document fetching" under what `integrations/agent_cli/**` owns, and `:117` still lists "ACP registry ids" as registry-owned discovery — now unconsumed at runtime. The pending task #4 ("Desktop pin-is-law + spec updates") looks like the intended home for this; please land those edits with the deletion so the specs don't describe code that no longer exists.

### Nits

3. **`require_source` is now unconditional for agent_process — a `PathOnly`/`Manual` spec would hard-error instead of skipping.** The old guards (`is_native_installable` / `is_agent_process_installable`) gated managed install; the new `service.rs:135-150` block always calls `require_source(AgentProcess)`. The model + registry projection still admit `PathOnly`/`Manual` (`model.rs:219,228`; `registry/projection.rs:151-160`), and the install spec's Module Ownership says install native/process artifacts "**when the descriptor supports managed install**." None of the 5 bundled agents is PathOnly/Manual for agent_process, so it doesn't fire today, but it's a latent behavior change: a non-managed descriptor would now fail install rather than no-op. Consider keeping a "not managed-installable ⇒ Ok(None)" early-out before `require_source`.

4. **Pin-less `reconcile_agents` is now a footgun.** `reconcile/mod.rs:31` (`|_| None`) passes no pins, so `require_source` returns `InvalidInstallSpec` for every role; reconcile then routes that through the catch-all `Err(error)` arm (`:83`) and marks every agent **`Failed`** (not `Skipped` — only `NotInstallable`/`UnsupportedPlatform` map to Skipped). The runtime entrypoints are safe — both `runtime.rs:136` and `reconcile/execution.rs:216-218` derive `catalog.pin_overrides(...)` — so this only bites the bare helper, which now has no runtime caller. Either delete it or have it derive bundled pins, so nobody reintroduces a silent all-Failed reconcile.

### Clean on these axes (verified, not rubber-stamped)
- **Module ownership / projection at the boundary:** `installer/` does NOT import `catalog/` structs — `ResolvedPinSource` is an explicit installer-domain mirror (`install_policy.rs:106-108`), and the catalog→installer projection (`project_source`) correctly lives in `catalog/service.rs:70`. Seed-build composition (`install-agents` → `AgentCatalogService` + `CatalogSyncService::from_bundled`) sits in the thin binary, which README allows.
- **Banned Shapes enforced:** native latest-fetch + PATH adoption (`native.rs`), the agent_process registry/fallback path, and the ACP `/latest` module are all deleted; `require_source` makes "no source ⇒ hard error." Importantly the deletion removed PATH **adoption** from *install* while leaving PATH **discovery** in *readiness* (`readiness/artifacts.rs`) — which the spec lists as a legitimate readiness Input. The `RegistryBacked`/`BinaryHint` model variants survive only because readiness still *reports* them, so they aren't orphaned.
- **Seed hydration + launcher regen:** intact. The PR deleted only the install path in `agent_process.rs`; `regenerate_seeded_agent_launchers` is retained and still called from hydration (`installer/seed/mod.rs:406`), satisfying agent-seeds.md step 7. Reconcile-after-hydrate ordering is untouched.
- **default_args vs ACP args:** the launcher correctly bakes only the pin's ACP-mode `args` and leaves session `default_args` runtime-applied, guarded by the new `npm_adapter_pin_bakes_acp_launch_args` test and the existing default-args contract.

pablonyx:
## 🤖 Automated review — producer & catalog data

Focused on producer correctness + committed catalog data. I read `resolve-pins.mjs` in full, inspected every `harness.source` block in `catalogs/agents/catalog.json`, and **live-verified** the upstream sources (GitHub releases, ACP registry). Resolution logic and the committed data are correct and current; the one real problem is wiring.

### Blocker

**1. The producer is not wired into the pipeline — the next probe run silently drops every `source` block.**
`resolve-pins.mjs` is referenced *nowhere* in any automation (no Makefile target, no CI step, no shell script — only spec prose and its own header). The scheduled `catalog-probe.yml` does:
- `build-catalog.mjs` → writes `scripts/agent-catalog/catalog.draft.json`, whose `harness` is `{ agentProcess: {version}, native: {version} }` with **no source blocks** (`build-catalog.mjs:503-504`)
- `git add scripts/agent-catalog/generated scripts/agent-catalog/catalog.draft.json` (`catalog-probe.yml:75`) — it never runs `resolve-pins.mjs` and never touches `catalogs/agents/catalog.json`.

`catalogs/agents/catalog.json` is the file `include_str!`'d into the binary (`bundled.rs:8-10`) and the one this PR enriches. Because the draft→shipped promotion is manual and the probe regenerates a source-less draft, the spec's claim that the `source` is *"produced by the probe (`resolve-pins.mjs`)"* (`agent-catalog-readiness.md:124`, `:465`) is false today. Whoever promotes the next probe output will wipe the pins unless they remember to re-run the producer. Either call `resolve-pins.mjs` from `catalog-probe.yml` after `build-catalog.mjs` (and add `catalogs/agents/catalog.json` to the commit), or make the promotion step run it. As-is, this is a foot-gun that un-does the whole PR on the next nightly catalog change. (Task #3 "automate the producer" is marked done, but CI does not invoke it.)

### Should-fix

**2. `npm view` failure produces a silently *unverified* pin, not a hard error.** `resolve-pins.mjs:204-212` returns `null` on `npm view … dist.integrity` failure → `sha256: null`. Catalog validation accepts that (npm sha is optional; `validation.rs:108-112` only guards the package string), and the installer never verifies npm bytes (`pinned.rs:72`). So a transient registry hiccup ships gemini with no integrity anchor and no signal. Given the design's thesis is "sha256 is the trust anchor," a null npm integrity should fail the run (or at minimum be loud + non-zero exit), not pass.

**3. No platform-completeness check.** `DEFAULT_PLATFORMS` is hardcoded to `macos_arm64, macos_x64, linux_x64` (`resolve-pins.mjs:30`), and validation only checks `!targets.is_empty()` (`validation.rs:96-99`, `service_tests.rs:69`). A pin missing a *shipped* platform (e.g. a future `--platforms` subset, or an ACP entry that drops one) passes build and only blows up at runtime via `NoPinForPlatform` (`pinned.rs:178-182`) on that host. Worth asserting the full shipped set is present per source in `validate_artifact_pin`. (linux_arm64 is in the registry platformMap but intentionally excluded here — fine, but it means an arm64 cloud host hard-fails; flag if that's ever a target.)

### Verified correct (live-checked, not rubber-stamped)

- **codex tarball_release**: `releases/latest` `tag_name` = `rust-v0.140.0` (its `name` is the bare `0.140.0`); the producer correctly uses `tag_name` for the versioned URL (`resolve-pins.mjs:115,119-121`). Committed URLs/sha/`expectedBinary` for all three shipped targets match. The `versionedUrlTemplate` regex (`:217`) is what's keyed, not `latestUrlTemplate` — good.
- **claude direct_binary**: `{version}`/`{platform}` templating against `platformMap` (`darwin-arm64`/`darwin-x64`/`linux-x64`) is correct; binary source has no `expectedBinary` and no args, as expected.
- **ACP registry-backed**: live registry confirms gemini npx args `["--acp"]`, cursor/opencode binary args `["acp"]`, cmd `./dist-package/cursor-agent` / `./opencode`. All match the catalog. `ACP_PLATFORM_MAP` darwin-aarch64→macos_arm64 etc. is correct and matches `Platform::registry_key()` (`model.rs:177+`) that the installer reads — producer and consumer are key-consistent.
- **Launch args**: captured only for adapters; native CLI archive args are `[]` (`resolve-pins.mjs:126`). Binary args taken consistently across platforms (last-wins, but ACP args are platform-invariant — confirmed in the live registry). gemini's `--acp` (from the trusted ACP registry) correctly differs from registry.json's untrusted `--experimental-acp` fallback.
- **sha-reuse (`knownSha`)**: sound. Keyed on the exact URL (`:59`, `:228`), seeded only from existing `{url, sha256}` pairs. A changed URL (version bump) misses the cache and forces a re-download; it cannot reuse a stale sha for a different URL. (npm/git pins aren't cached, so `npm view` re-runs every time — harmless, just network.)
- **git pins**: claude/codex carry full 40-char commit refs, correct repo, `executableRelpath`, and codex's `packageSubdir: "npm"`. ✓

### Nits

- `--no-download` blanks all binary/archive shas (`:227`) but `knownSha` only rescues unchanged URLs — fine for inspection, but a `--no-download` run committed by accident would produce empty shas that fail validation. Intentional, just call it out.
- `npmVersionOf` uses `lastIndexOf("@")` (`:200`) — correct for `@scope/pkg@1.2.3`, but assumes the registry never pins a range/tag. ACP npx packages are exact today, so OK.
- `githubLatestTag` has no GITHUB_TOKEN auth (`:220-222`); fine locally, but in CI the unauthenticated 60/hr rate limit could flake the codex resolve. Add the token when you wire it into CI (see Blocker).

**Bottom line:** the resolver logic and the committed lockfile data are correct and currently accurate. The PR is undermined by #1 — the producer must be invoked by the pipeline (or the promotion step) or the pins evaporate on the next probe-driven catalog update.

pablonyx:
## 🤖 Automated review — installer correctness & security

Read `gh pr diff 723` end-to-end plus the current `pinned.rs` / `downloads.rs` / `service.rs` / `install_policy.rs` / `catalog/{service,validation,schema}.rs` and the live call paths (`runtime.install_agent`, `start_reconcile` → `reconcile/execution.rs`). `cargo check -p anyharness-lib` is clean. The sha256 fence on the binary/archive byte paths is genuinely solid. But the archive materializer drops sibling files, which **breaks cursor at runtime**, and that path has no test.

### 🔴 Blocker — archive extraction discards sibling files; cursor's adapter is broken
`download_and_extract_archive_verified` (`installer/downloads.rs:152-164`) renames **only** the single `expected_binary` out of `_staging` to `target_path`, then `remove_dir_all(&staging_dir)` — destroying every other extracted file.

cursor's `agentProcess` pin (this diff, `catalog.json:1614-1636`) is `expectedBinary: "./dist-package/cursor-agent"` with `args:["acp"]`. I pulled the pinned tarball: `dist-package/` is a **203-entry Node bundle**. `cursor-agent` is a bash wrapper (`#!/usr/bin/env bash … SCRIPT_DIR=…`) that resolves its own dir and execs the sibling `dist-package/node` (+ needs `spawn-helper`, `pty.node`, ~119 `*.index.js`). After this code runs, only `cursor-agent` survives at `<managed>/cursor`; its `SCRIPT_DIR/node` is gone → launch fails.

This is a regression: the **deleted** `acp_registry/archive.rs` renamed the *whole staging dir* into place and had a test asserting exactly this — `binary_archive_install_preserves_extracted_sibling_files` checks `dest_dir/dist-package/node` survives (diff lines 1807-1849). That guarantee was dropped with no replacement.
- Fix: for a multi-file archive, move the extracted **tree** (or the `expected_binary`'s parent dir) into place and point the launcher at the binary within it, as the old code did. opencode (`./opencode`) and codex (`codex-*`) are single self-contained binaries so they're unaffected — but cursor is shipped broken.
- No test covers the multi-file archive path in the new code (`pinned.rs` tests only exercise single-file `Binary` + `Npm`). Add one with a wrapper-that-execs-a-sibling fixture.

### 🟠 Should-fix — npm pin `sha256`/`sha512` integrity is collected but never verified
`ResolvedPinSource::Npm { sha256, .. }` carries the registry integrity (gemini pin: `sha512-…`, this diff), `validate_artifact_pin` only checks the package string is non-empty, and `install_agent_process_from_pin` → `install_managed_npm_package` **ignores `sha256` entirely**, shelling `npm install <pkg@ver>` (no `--ignore-scripts`, lifecycle scripts run). So for npm adapters the "fenced, sha-verified, the pin is law" property does **not** hold — npm resolves/executes whatever the registry serves for that version. Either enforce the integrity (e.g. pass through to npm / verify the tarball) or down-scope the claim. The comment in `pinned.rs:1-8` already concedes npm/git "stay on the existing path", but `install_policy.rs:108` advertises the per-target sha256 as "the trust anchor" without noting npm is exempt.

### 🟠 Should-fix — extraction shells to `tar`/`unzip` with no path-traversal guard
The deleted Rust `zip` extractor used `entry.enclosed_name()` to reject entries escaping the destination. The new code (`downloads.rs:120-150`) shells to `tar xzf`/`unzip -q` into `_staging` with no `../` check. `unzip` in particular can write outside the target on crafted `../` entries. Given the archive is sha-pinned the practical risk is low (a tampered archive fails the sha first), but a malicious *pin source* (compromised catalog) gets arbitrary-write. Add a post-extract check that everything landed under `staging_dir`, or keep CLI extraction but verify resolved paths.

### 🟡 Nits
- **`.zip` dispatch is extension-only** (`downloads.rs:120`, `url.ends_with(".zip")`). The deleted code also sniffed PK magic bytes as a fallback. All current pins have clean `.tar.gz`/`.zip` suffixes so it's fine today, but a query-stringed or suffixless URL would silently be fed to `tar` and fail with a confusing error.
- **Dev/test fence ergonomics**: the two no-pins free fns — `reconcile_agents()` (`reconcile/mod.rs:31`, `|_| None`) and `installer::install_agent()` (`service.rs:68`) — now hard-error via `require_source` for **every** agent, since `catalog_pins=None` ⇒ no source. Both live entrypoints (`runtime.install_agent:136`, `start_reconcile`) correctly pass bundled-catalog pins, so production is fine, but these two now only produce `InvalidInstallSpec`. Consider deleting them or routing them through the bundled catalog so they don't become a footgun.

### ✅ Verified good
- sha256 fence on binary + archive: `verify_sha256` (`downloads.rs:76-87`) is a streamed lowercase-hex compare via `eq_ignore_ascii_case` (case/format-correct), removes the file on mismatch, and is called **before** `make_executable`/extract on both byte paths. Binary uses temp `.downloading` → verify → `rename` (atomic, nothing survives a bad sha — `wrong_sha_is_refused_and_nothing_is_installed` proves it). Archive verifies the downloaded archive before extracting.
- The fence (`require_source`, `service.rs:245-257`): both live install paths supply pins; a missing source is a typed hard error, not a silent fallback. Legacy `binary_hint` / PATH-adoption / latest-fetch / ACP `/latest` routes are fully deleted (`native.rs`, `acp_registry/**`, `agent_process` fallbacks all gone). No remaining path reaches a non-pin source in production.
- Launcher bakes the pin's ACP `args` (`acp` / `--acp`) but **not** session `default_args` — `npm_adapter_pin_bakes_acp_launch_args` confirms, and `default_args` is no longer threaded into install. git/npm/archive dispatch is correct; `DISABLE_AUTOUPDATER` env + node PATH-prefix carry over via `managed_launcher_env` / `launcher_path_prefixes`.
- `Platform::detect()` returning a platform with no matching target → `NoPinForPlatform` typed error (no silent skip). Errors are typed and surfaced throughout; nothing swallowed.

— claude-opus-4-8

pablonyx:
## Review findings — resolutions (commit a95d40208)

Thanks — the review caught a real regression. Disposition:

**Fixed**
- 🔴 **cursor extraction (blocker)** — adapter archives now extract the WHOLE tree (`download_and_extract_archive_tree_verified`) and the launcher points at the entry binary *inside* the managed tree, so cursor's `dist-package/cursor-agent` keeps its ~200 siblings. New regression test `archive_adapter_pin_preserves_sibling_files` builds a multi-file tarball and asserts a sibling survives. (Native single-binary extraction unchanged — codex/opencode unaffected.)
- 🟠 **npm null-integrity** — `validate_artifact_pin` now rejects an npm pin with no sha256, so a failed `npm view` can't ship as an unverifiable pin.
- 🟠 **platform completeness** — `bundled_catalog_is_a_complete_lockfile` now requires every shipped platform (macos_arm64/x64, linux_x64) on binary/archive pins, not just non-empty targets.
- **stale specs** — `agents.md` + `integrations.md` no longer reference the deleted `acp_registry` module; ACP-registry resolution is documented as probe-time (`resolve-pins.mjs`). (`acp.md`'s "does not own ACP registry lookup" is still accurate.)

**Deferred, with rationale**
- 🔴 **producer not wired into CI** — correct and tracked as its own change (it edits the nightly `catalog-probe.yml` so `resolve-pins.mjs` runs after `build-catalog.mjs` and the source blocks survive a probe run). Out of scope for this PR; not merging until that lands or is explicitly accepted as a manual step.
- **npm byte-verification depth** — `npm install <pkg>@<version>` verifies the tarball against the registry's published integrity (the same sha512 we now require in the pin), so pinned npm installs are integrity-checked by npm; an independent post-install comparison is a possible hardening, noted.
- **tar/unzip path-traversal** — acknowledged; low practical risk because the archive bytes are sha-pinned (a traversal entry would change the hash and fail verification). Worth a typed extractor later.

702 lib + 15 binary tests green; full workspace builds.

pablonyx:
## Second-round review — resolutions (commit b5e6d2df1)

A fresh adversarial pass (3 spec-grounded reviewers). Disposition:

**Fixed**
- **producer not wired (merge gate)** — `catalog-probe.yml` now runs `resolve-pins.mjs` on the fresh draft (reusing prior shas for unchanged URLs) and promotes it to the bundled `catalogs/agents/catalog.json`, so a probe run can't drift the shipped lockfile from the model draft. `resolve-pins` gains `--reuse-from`; `make catalog-pin` mirrors it locally. Verified: a source-less draft → complete lockfile via reuse, zero downloads.
- **git ref must be a full commit SHA** — `validate_artifact_pin` now rejects a mutable branch/tag (not a trust anchor).
- **npm anchor honesty** — corrected the `install_policy` contract: npm's anchor is the pinned `pkg@version` + npm's registry-integrity check; binary/archive remain sha-verified at download. (No longer claims sha-verify for all four kinds.)
- **cursor stale-tree** — a `MissingManagedArtifact` after extraction no longer leaves a populated-but-unusable `registry_binary/` behind.
- **`agent-seed.inputs.json`** — dropped the dead `agents` block (unread, but hashed into the seed manifest with refs that disagreed with the catalog).
- **stale specs** — `catalogs/agents` path (no `v1/`), Source-Shape tree (deleted `native.rs` → `pinned.rs`/`install_policy.rs`), agents.md ACP-registry ownership/extension points.

**Acknowledged, not changed**
- npm independent byte-verification (compare package-lock integrity post-install) — npm already verifies the pinned version against the registry; deferred hardening.
- catalog-as-trust-root via runtime sync-swap is by design; the push channel (bearer auth) is the perimeter — worth a deploy invariant that a network-reachable runtime always sets a bearer token (pre-existing infra).
- `remove_dir_all` reinstall racing a live cursor session — extract-and-swap is a follow-up.
- dead code (`RegistryFailed`, no-pin `install_agent`/`reconcile_agents` wrappers) — nits; all fail closed.

pablonyx:
## Rebased onto latest main + round-3 review (HEAD 84e42cc5f)

Rebased onto `origin/main` (now includes **#705 Grok**). One conflict — `installer/agent_process.rs`, where Grok had added code to the `install_from_registry`/`find_npm_bin` path this PR deletes; resolved by taking the trimmed file (Grok needs no special install code under the fence).

Post-rebase verification:
- **Grok fully integrated** — complete catalog npm pin (`@xai-official/grok@0.2.54` + integrity + `["agent","stdio"]` args, resolved from the ACP registry), `AgentKind::Grok` and every downstream path (install/readiness/credentials/projection/seed) handle it. Verified it installs through the fenced npm path identically to cursor/opencode; nothing dropped from `agent_process.rs`.
- **All prior fixes survived the rebase** (verified present): cursor tree-extraction + stale-tree cleanup, npm-integrity + git-full-SHA validation, honest per-kind anchor doc, producer CI wiring + `--reuse-from`, platform-completeness guard. Zero stray conflict markers.
- **708 lib + 15 binary tests green; full workspace builds.**

Mergeable post-rebase; the open items are the acknowledged follow-ups (desktop boot-reconcile, npm byte-verify, bearer-token deploy invariant).

PR #724, opened 2026-06-16
fix(live): resolve cursor variant model ids to advertised values on switch

pablonyx:
## Problem

Switching the model on a running **cursor** session "always goes back to Auto." Verified live against `cursor-agent` 2026.06.15.

Cursor (`variantSyntax: bracket-params`) exposes a model **config option** whose only valid values are **fully-composed variants**, whose default params are decided by the harness at runtime:
```
composer-2.5[fast=true]
claude-opus-4-8[thinking=true,context=300k,effort=high,fast=false]
kimi-k2.5[]   (current)
```
A switch carries the bare catalog id `composer-2.5`. Cursor rejects it (`Invalid model value`); the runtime then falls back to a broken relaunch → HTTP 500 → the model reverts.

Proof it's purely the value format (live test):
```
switch → composer-2.5            HTTP 500 → reverts          (before)
switch → composer-2.5[fast=true] HTTP 200 → sticks
```

## Fix

Resolve the requested model value to the harness's advertised value in the live-apply path: when the model option doesn't list the requested value but advertises **exactly one** value with the same base id (text before `[`), send that advertised value. New `resolve_model_variant_value` in `config/selection.rs`.

**Why the live option, not the catalog:** the catalog *can* echo back a pre-composed `composer-2.5[fast=true]` (it's in `provenance.variantIds`), but it **cannot derive that composed default from a bare base id** — `resolve_requested_model`/`compose_variant` only match the base (→ bare id) or validate caller-supplied params (→ empty `[]`). The harness's runtime default (`fast=true`) is carried **only** by the live ACP option at apply time, which is also where the spec puts post-start truth (`agents-catalog-registry-migration.md` §5.5). Apply-time code has no catalog/`kind`/contexts in scope, so routing through `validate_launch` is both impossible-for-defaults and a far larger change.

### Hardening (from adversarial review)
- **Never override an explicit choice:** only fill in a *missing* variant — `composer-2.5[fast=false]` is left untouched; empty `[]` still resolves.
- **Scoped to bracket-params shape:** a match must be `[]` or `[k=v,...]`, never a bare context tag like `sonnet[1m]` — so the resolver is a no-op for non-variant agents (claude) regardless of their option shape.
- **No persistence drift:** resolve once up front on both the idle (`apply_specific_config_option`) and busy/queue (`handle_busy_config_command`) paths, so the value validated, sent, queued, AND persisted all agree (`requested_model_id` stores the composed value).
- **Observability:** `tracing::debug!` logs `requested → resolved` on substitution.

## Testing
- `cargo test -p anyharness-lib config` — 55 pass (incl. resolver unit tests: base→composed, explicit-param non-override, `[1m]` non-collapse, non-model no-op).
- **Live** (real cursor-agent 2026.06.15): bare `composer-2.5`, `gpt-5.5`, `claude-opus-4-8` (multi-param), `gemini-2.5-flash` all switch in place (200) and stick; **persisted record `requested = current = composer-2.5[fast=true]`**; composed model reloads correctly after a runtime restart; no relaunch/500s.

## Known follow-ups (out of scope, tracked)
- **Codex `slash-effort`** (`gpt-5.5/high`) is not covered by the live resolver (it gates on `[`-shape); fine since codex isn't the target, but the variant grammar now lives in two places (catalog `compose_variant` by declared `variantSyntax`, live resolver by shape). Unify once apply-time gains `variantSyntax` access.
- **Membership-gate relaxation** (`same-harness-model-switch-design.md` §5) and **catalog ∪ live switch-menu** union (§5.5) — separate sub-problems.
- The **relaunch fallback** still has latent bugs (composed value → launch env; native-session-not-found on restart) and violates "never recreate on same-harness switch." This fix avoids hitting it for cursor.

Independent of #715 (gemini `session/set_model`). Rebased on latest main.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/live/sessions/actor/config/apply.rs
+++ b/anyharness/crates/anyharness-lib/src/live/sessions/actor/config/apply.rs
@@ -137,6 +137,15 @@ pub(in crate::live::sessions::actor) async fn apply_select_config_option_with_po
     // `variantSyntax` agents (cursor) only accept their advertised composed
     // values (`composer-2.5[fast=true]`); a switch may carry just the base id.
     let resolved_value = resolve_model_variant_value(option, desired_value);
+    if resolved_value != desired_value {
+        tracing::debug!(
+            native_session_id,
+            config_id,
+            requested = desired_value,
+            resolved = %resolved_value,
+            "[model-switch] resolved variant model id to advertised value"
+        );
+    }
     let desired_value = resolved_value.as_str();
 
     if !select_option_contains_value(option, desired_value) && !allow_foreign_value {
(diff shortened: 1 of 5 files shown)


pablonyx:
## Grok: confirmed no-op

Grok (#705, on main; this branch is rebased onto it) has **no `variantSyntax`** and **no model config option** (`modelSource: initMetaModelState`), and its ids are plain (`grok-4.3`). So `resolve_model_variant_value` can never fire for grok: the apply funnel returns early without a model option, and even with one, plain ids aren't bracket-params-shaped. No behavior change for grok — it is covered by the gemini-shaped path in #715, not this PR.

pablonyx:
Superseded by the clean current-main port in #1137, now merged.

PR #731, opened 2026-06-18
fix(desktop): persist auth + provider creds as 0600 files in ~/.proliferate

pablonyx:
## Problem
The macOS keychain item holding the desktop **auth session** (`com.proliferate.app.auth` / `desktop_session`) is ACL-bound to the build's **code signature**. A reinstalled or re-signed app can no longer read it, so users **have to log in again after every reinstall/update** — confirmed in production (signed release), not just ad-hoc builds. Provider/env credentials (`com.proliferate.app.env`, keychain-only) had the same loss.

The team already documents this fragility in `keychain.rs` (the delete-recreate-on-write workaround) and guards against ad-hoc releases in CI — but the signed path is still brittle without a `keychain-access-groups` entitlement.

## Change
Move **recreatable** secrets to `0600` files under the durable app home `~/.proliferate` (dev: `~/.proliferate-local`), which **survives uninstall/reinstall and updates** — proven by the already-stable `install_id` living there. Dev already stored the auth session there; this makes release do the same.

- **Auth session + pending OAuth state** → always file (dropped the `native_dev_profile()` gate).
- **Provider/env credentials** → a file-resident `{name: value}` map at `~/.proliferate/env-secrets.json`, covering `set`/`delete`/`list`/internal `read_env_secret` **and** `load_all_secrets_for_sidecar` (the runtime-injection path).
- **The anyharness data key STAYS in the keychain** — it's an at-rest **encryption key**, which a plaintext file would defeat. (Its own reinstall-loss is recoverable — re-establish MCP bindings — and is left for a follow-up `keychain-access-groups` entitlement.)
- Removed now-dead keychain plumbing: `SERVICE`/`AUTH_SERVICE`/account consts, `delete_password`, `KeychainRequest::Delete`.

Tauri command names/signatures are unchanged ⇒ **frontend untouched**. No migration: existing users re-auth / re-enter provider keys once. Server-synced preferences are unaffected.

## Verification
- `cargo check` — clean, zero warnings.
- `cargo test commands::keychain` — 4/4 pass, incl. a new `env_secrets_map_round_trips_through_file_at_0600` (asserts the perms) + retained auth/pending file tests.
- **Manual acceptance (not automatable here):** log in → confirm `~/.proliferate/auth-session.json` is `0600` → **fully quit + reinstall the signed app → still logged in**.

## Security note
This trades OS-encrypted keychain storage for `0600` files for the recreatable tokens — the dev-tool norm (`gh`, `aws`, `npm`, `docker` all keep creds in plaintext dotfiles). The one item that genuinely needs at-rest encryption (the data key) stays in the keychain.

### DIFF
--- a/apps/desktop/src-tauri/src/commands/keychain.rs
+++ b/apps/desktop/src-tauri/src/commands/keychain.rs
@@ -434,6 +434,9 @@ pub async fn set_auth_session(session: AuthSessionRecord) -> Result<(), String> 
 
 #[tauri::command]
 pub async fn clear_auth_session() -> Result<(), String> {
+    let _guard = SECRET_FILE_LOCK
+        .lock()
+        .map_err(|_| "secret file lock poisoned".to_string())?;
     delete_file_if_exists(&auth_session_file_path()?)
 }
 
@@ -452,6 +455,9 @@ pub async fn set_pending_auth(record: PendingAuthRecord) -> Result<(), String> {
 
 #[tauri::command]
 pub async fn clear_pending_auth() -> Result<(), String> {
+    let _guard = SECRET_FILE_LOCK
+        .lock()
+        .map_err(|_| "secret file lock poisoned".to_string())?;
     delete_file_if_exists(&pending_auth_file_path()?)
 }
 
--- a/apps/desktop/src-tauri/src/commands/keychain.rs
+++ b/apps/desktop/src-tauri/src/commands/keychain.rs
@@ -7,13 +7,9 @@ use base64::Engine;
 use getrandom::fill;
 use serde::{Deserialize, Serialize};
 
-use crate::app_config::{app_dir_path, native_dev_profile, read_json_file, write_json_file_atomic};
+use crate::app_config::{app_dir_path, read_json_file, write_json_file_atomic};
 
-const SERVICE: &str = "com.proliferate.app.env";
-const AUTH_SERVICE: &str = "com.proliferate.app.auth";
 const RUNTIME_SERVICE: &str = "com.proliferate.app.runtime";
-const AUTH_SESSION_ACCOUNT: &str = "desktop_session";
-const PENDING_AUTH_ACCOUNT: &str = "desktop_pending_auth";
 const ANYHARNESS_DATA_KEY_ACCOUNT: &str = "anyharness_data_key";
 const ANYHARNESS_DATA_KEY_ENV: &str = "ANYHARNESS_DATA_KEY";
 
@@ -27,10 +23,14 @@ const KNOWN_ENV_VARS: &[&str] = &[
     "AMP_API_KEY",
 ];
 
-// Dev builds are ad-hoc signed, so every rebuild is a different app to the
-// macOS keychain and items they create can never be durably readable. Dev
-// lanes therefore persist auth records as files under the per-profile app dir
-// (PROLIFERATE_DEV_HOME) instead of the keychain.
+// Recreatable secrets — the auth session, pending OAuth state, and provider/env
+// credentials — live as 0600 files under the durable app home (`~/.proliferate`,
+// dev: `~/.proliferate-local`). That directory survives uninstall/reinstall and
+// app updates, so the session persists across them. A macOS keychain item does
+// not: its ACL is bound to the build's code signature, so a reinstalled or
+// re-signed build can no longer read it (hence the "log in again after
+// reinstall" bug). Only the anyharness data key — an at-rest encryption key that
+// a plaintext file would defeat — stays in the keychain.
 fn auth_session_file_path() -> Result<PathBuf, String> {
     Ok(app_dir_path()?.join("auth-session.json"))
 }
@@ -39,7 +39,19 @@ fn pending_auth_file_path() -> Result<PathBuf, String> {
     Ok(app_dir_path()?.join("pending-auth.json"))
 }
 
-fn write_dev_auth_record<T: Serialize>(path: &Path, value: &T) -> Result<(), String> {
+fn env_secrets_path() -> Result<PathBuf, String> {
+    Ok(app_dir_path()?.join("env-secrets.json"))
+}
+
+/// The `{ env_var_name: value }` map of stored provider/env credentials.
+/// A missing file is an empty map.
+fn read_env_secrets_map() -> Result<HashMap<String, String>, String> {
+    Ok(read_json_file(&env_secrets_path()?)?.unwrap_or_default())
+}
+
+/// Atomically write `value` as JSON, then restrict it to owner-only (0600).
+/// The writer for every recreatable secret file.
+fn write_secret_file<T: Serialize>(path: &Path, value: &T) -> Result<(), String> {
     write_json_file_atomic(path, value)?;
     #[cfg(unix)]
     {
@@ -160,7 +172,7 @@ const AGENT_AUTH_PROVIDERS: &[AgentAuthProviderSpec] = &[
 ];
 
 fn read_env_secret(name: &str) -> Result<Option<String>, String> {
-    read_password(SERVICE, name)
+    Ok(read_env_secrets_map()?.get(name).cloned())
 }
 
 enum KeychainRequest {
@@ -173,11 +185,6 @@ enum KeychainRequest {
         service: String,
         account: String,
         value: String,
-        response: mpsc::SyncSender<Result<(), String>>,
-    },
-    Delete {
-        service: String,
-        account: String,
         response: mpsc::SyncSender<Result<(), String>>,
     },
 }
@@ -253,19 +260,6 @@ fn keychain_sender() -> &'static mpsc::Sender<KeychainRequest> {
                                 });
                             let _ = response.send(result);
                         }
-                        KeychainRequest::Delete {
-                            service,
-                            account,
-                            response,
-                        } => {
-                            let result = get_or_create_entry(&mut entries, &service, &account)
-                                .and_then(|entry| match entry.delete_credential() {
-                                    Ok(()) => Ok(()),
-                                    Err(keyring::Error::NoEntry) => Ok(()),
-                                    Err(error) => Err(error.to_string()),
-                                });
-                            let _ = response.send(result);
-                        }
                     }
                 }
             })
@@ -303,20 +297,6 @@ fn set_password(service: &str, account: &str, value: &str) -> Result<(), String>
         .map_err(|_| "Keychain worker did not return a result.".to_string())?
 }
 
-fn delete_password(service: &str, account: &str) -> Result<(), String> {
-    let (response_tx, response_rx) = mpsc::sync_channel(1);
-    keychain_sender()
-        .send(KeychainRequest::Delete {
-            service: service.to_string(),
-            account: account.to_string(),
-            response: response_tx,
-        })
-        .map_err(|_| "Keychain worker is unavailable.".to_string())?;
-    response_rx
-        .recv()
-        .map_err(|_| "Keychain worker did not return a result.".to_string())?
-}
-
 fn ensure_runtime_data_key() -> Result<String, String> {
     if let Some(value) = read_password(RUNTIME_SERVICE, ANYHARNESS_DATA_KEY_ACCOUNT)? {
         return Ok(value);
@@ -337,83 +317,58 @@ fn home_dir() -> Result<PathBuf, String> {
 
 #[tauri::command]
 pub async fn list_configured_env_var_names() -> Result<Vec<String>, String> {
-    let mut names = Vec::new();
-    for &var in KNOWN_ENV_VARS {
-        if read_password(SERVICE, var)?.is_some() {
-            names.push(var.to_string());
-        }
-    }
-    Ok(names)
+    let map = read_env_secrets_map()?;
+    Ok(KNOWN_ENV_VARS
+        .iter()
+        .filter(|var| map.contains_key(**var))
+        .map(|var| var.to_string())
+        .collect())
 }
 
 #[tauri::command]
 pub async fn set_env_var_secret(name: String, value: String) -> Result<(), String> {
-    set_password(SERVICE, &name, &value)
+    let mut map = read_env_secrets_map()?;
+    map.insert(name, value);
+    write_secret_file(&env_secrets_path()?, &map)
 }
 
 #[tauri::command]
 pub async fn delete_env_var_secret(name: String) -> Result<(), String> {
-    delete_password(SERVICE, &name)
+    let mut map = read_env_secrets_map()?;
+    if map.remove(&name).is_some() {
+        write_secret_file(&env_secrets_path()?, &map)?;
+    }
+    Ok(())
 }
 
 #[tauri::command]
 pub async fn get_auth_session() -> Result<Option<AuthSessionRecord>, String> {
-    if native_dev_profile() {
-        return read_json_file(&auth_session_file_path()?);
-    }
-    match read_password(AUTH_SERVICE, AUTH_SESSION_ACCOUNT)? {
-        Some(raw) => serde_json::from_str(&raw)
-            .map(Some)
-            .map_err(|e| e.to_string()),
-        None => Ok(None),
-    }
+    read_json_file(&auth_session_file_path()?)
 }
 
 #[tauri::command]
 pub async fn set_auth_session(session: AuthSessionRecord) -> Result<(), String> {
-    if native_dev_profile() {
-        return write_dev_auth_record(&auth_session_file_path()?, &session);
-    }
-    let raw = serde_json::to_string(&session).map_err(|e| e.to_string())?;
-    set_password(AUTH_SERVICE, AUTH_SESSION_ACCOUNT, &raw)
+    write_secret_file(&auth_session_file_path()?, &session)
 }
 
 #[tauri::command]
 pub async fn clear_auth_session() -> Result<(), String> {
-    if native_dev_profile() {
-        return delete_file_if_exists(&auth_session_file_path()?);
-    }
-    delete_password(AUTH_SERVICE, AUTH_SESSION_ACCOUNT)
+    delete_file_if_exists(&auth_session_file_path()?)
 }
 
 #[tauri::command]
 pub async fn get_pending_auth() -> Result<Option<PendingAuthRecord>, String> {
-    if native_dev_profile() {
-        return read_json_file(&pending_auth_file_path()?);
-    }
-    match read_password(AUTH_SERVICE, PENDING_AUTH_ACCOUNT)? {
-        Some(raw) => serde_json::from_str(&raw)
-            .map(Some)
-            .map_err(|e| e.to_string()),
-        None => Ok(None),
-    }
+    read_json_file(&pending_auth_file_path()?)
 }
 
 #[tauri::command]
 pub async fn set_pending_auth(record: PendingAuthRecord) -> Result<(), String> {
-    if native_dev_profile() {
-        return write_dev_auth_record(&pending_auth_file_path()?, &record);
-    }
-    let raw = serde_json::to_string(&record).map_err(|e| e.to_string())?;
-    set_password(AUTH_SERVICE, PENDING_AUTH_ACCOUNT, &raw)
+    write_secret_file(&pending_auth_file_path()?, &record)
 }
 
 #[tauri::command]
 pub async fn clear_pending_auth() -> Result<(), String> {
-    if native_dev_profile() {
-        return delete_file_if_exists(&pending_auth_file_path()?);
-    }
-    delete_password(AUTH_SERVICE, PENDING_AUTH_ACCOUNT)
+    delete_file_if_exists(&pending_auth_file_path()?)
 }
 
 #[tauri::command]
@@ -576,9 +531,11 @@ fn portable_export_to_agent_auth_files(
 
 pub fn load_all_secrets_for_sidecar() -> HashMap<String, String> {
     let mut env = HashMap::new();
-    for &var in KNOWN_ENV_VARS {
-        if let Ok(Some(password)) = read_password(SERVICE, var) {
-            env.insert(var.to_string(), password);
+    if let Ok(map) = read_env_secrets_map() {
+        for &var in KNOWN_ENV_VARS {
+            if let Some(value) = map.get(var) {
+                env.insert(var.to_string(), value.clone());
+            }
         }
     }
     if let Ok(data_key) = ensure_runtime_data_key() {
@@ -610,7 +567,7 @@ mod tests {
     }
 
     #[test]
-    fn auth_session_record_round_trips_through_dev_file() {
+    fn auth_session_record_round_trips_through_file() {
         let path = temp_path("auth-session.json");
         let record = AuthSessionRecord {
             access_token: "access".to_string(),
@@ -623,7 +580,7 @@ mod tests {
             avatar_url: None,
         };
 
-        write_dev_auth_record(&path, &record).expect("write should succeed");
+        write_secret_file(&path, &record).expect("write should succeed");
         let parsed: AuthSessionRecord = read_json_file(&path)
             .expect("read should succeed")
             .expect("file should exist");
@@ -648,7 +605,7 @@ mod tests {
     }
 
     #[test]
-    fn pending_auth_record_round_trips_through_dev_file() {
+    fn pending_auth_record_round_trips_through_file() {
         let path = temp_path("pending-auth.json");
         let record = PendingAuthRecord {
             state: "state".to_string(),
@@ -658,7 +615,7 @@ mod tests {
             last_handled_callback_url: None,
         };
 
-        write_dev_auth_record(&path, &record).expect("write should succeed");
+        write_secret_file(&path, &record).expect("write should succeed");
         let parsed: PendingAuthRecord = read_json_file(&path)
             .expect("read should succeed")
             .expect("file should exist");
@@ -680,6 +637,33 @@ mod tests {
         std::fs::write(&path, b"{}").expect("write should succeed");
         delete_file_if_exists(&path).expect("existing file should delete");
         assert!(!path.exists());
+
+        cleanup(&path);
+    }
+
+    #[test]
+    fn env_secrets_map_round_trips_through_file_at_0600() {
+        let path = temp_path("env-secrets.json");
+        let mut map = HashMap::new();
+        map.insert("ANTHROPIC_API_KEY".to_string(), "sk-ant-xxx".to_string());
+        map.insert("OPENAI_API_KEY".to_string(), "sk-openai-yyy".to_string());
+
+        write_secret_file(&path, &map).expect("write should succeed");
+        let parsed: HashMap<String, String> = read_json_file(&path)
+            .expect("read should succeed")
+            .expect("file should exist");
+        assert_eq!(parsed.get("ANTHROPIC_API_KEY").map(String::as_str), Some("sk-ant-xxx"));
+        assert_eq!(parsed.get("OPENAI_API_KEY").map(String::as_str), Some("sk-openai-yyy"));
+
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::PermissionsExt;
+            let mode = std::fs::metadata(&path)
+                .expect("metadata should be readable")
+                .permissions()
+                .mode();
+            assert_eq!(mode & 0o777, 0o600);
+        }
 
         cleanup(&path);
     }
--- a/apps/desktop/src-tauri/src/commands/keychain.rs
+++ b/apps/desktop/src-tauri/src/commands/keychain.rs
@@ -1,17 +1,30 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
-use std::sync::{mpsc, OnceLock};
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{mpsc, Mutex, OnceLock};
 
 use anyharness_credential_discovery::{export_portable_auth, PortableAuthExport, ProviderId};
 use base64::Engine;
 use getrandom::fill;
 use serde::{Deserialize, Serialize};
 
-use crate::app_config::{app_dir_path, read_json_file, write_json_file_atomic};
+use crate::app_config::{app_dir_path, read_json_file};
 
 const RUNTIME_SERVICE: &str = "com.proliferate.app.runtime";
 const ANYHARNESS_DATA_KEY_ACCOUNT: &str = "anyharness_data_key";
 const ANYHARNESS_DATA_KEY_ENV: &str = "ANYHARNESS_DATA_KEY";
+
+// Legacy keychain locations recreatable secrets used to live in, kept only so we
+// can purge orphaned items after migrating to file storage (see
+// `purge_legacy_keychain_secrets`).
+const LEGACY_ENV_SERVICE: &str = "com.proliferate.app.env";
+const LEGACY_AUTH_SERVICE: &str = "com.proliferate.app.auth";
+const LEGACY_AUTH_SESSION_ACCOUNT: &str = "desktop_session";
+const LEGACY_PENDING_AUTH_ACCOUNT: &str = "desktop_pending_auth";
+
+// Serializes writes to the secret files so concurrent Tauri commands cannot lose
+// an update (read-modify-write of env-secrets.json) or collide on a temp file.
+static SECRET_FILE_LOCK: Mutex<()> = Mutex::new(());
 
 const KNOWN_ENV_VARS: &[&str] = &[
     "ANTHROPIC_API_KEY",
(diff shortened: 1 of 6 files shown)


pablonyx:
### 🧭 Conventions & spec-hygiene review (subagent)

Scope: codebase conventions + spec hygiene only (not correctness). Read against `specs/spec-catalog.md`, `specs/codebase/structures/desktop-native/**`, and the AnyHarness grammar primer (`anyharness/guides/mental-model.md`). Caveat: that grammar primer is authored for `anyharness-lib`, so I weight it as guidance, not law, for `src-tauri`.

The change itself is well-motivated and the diff is clean (dead keychain plumbing — `SERVICE`/`AUTH_SERVICE`/account consts, `KeychainRequest::Delete`, `delete_password`, the `native_dev_profile()` gates — all removed; `native_dev_profile` correctly dropped from the `use`). Findings below.

---

**🔴 Should-fix — spec not updated in the same PR (the catalog rule).**
`specs/spec-catalog.md:21`: *"Rule: read the relevant spec before touching code; update it in the same PR."* This PR changes where desktop secrets live but updates zero spec files, and two desktop-native specs now make factually-stale claims:

- `specs/codebase/structures/desktop-native/specs/anyharness-sidecar.md:44` ("keychain-backed local secrets") and `:94` ("keychain secrets") — after this PR the auth session, pending OAuth, and provider/env creds are `0600` files; **only the data key is keychain-backed.** Both lines are now wrong. The restart-rules list at `:94` should read something like "file-backed local secrets + the keychain data key."
- `specs/codebase/structures/desktop-native/README.md:31` ("`keychain.rs # local secret storage and sidecar launch secrets`") and the Secrets rule at `:52` ("do not persist provider secrets in app config JSON"). The rule isn't violated (these are dedicated `env-secrets.json`/`auth-session.json` files, not `config.json`), but the README's framing still implies keychain is the store. Update the file-tree comment and the rule prose to name the new `~/.proliferate` file storage and the keychain-only-for-data-key split.

**Recommendation:** in this PR, (a) fix those four lines, and (b) document the storage location + the "keychain holds only the at-rest data key" decision somewhere authoritative. The natural home is the desktop-native README/sidecar spec; the durable-home semantics (`~/.proliferate` survives uninstall/update) already live at `anyharness-sidecar.md:71-87`, so a "Local Secrets" section there is the cleanest fit. Note `specs/spec-catalog.md:378` flags that **no authoritative security spec exists** — the at-rest-encryption rationale (data key stays encrypted; recreatable tokens accept `0600`) is exactly the kind of decision that spec would own, but absent it, desktop-native is the right place rather than leaving it uncaptured. `features/product-auth.md` is *not* the right home (it's scoped to server/web/mobile account auth, not desktop token-at-rest), so I would not require a change there.

---

**🟡 Should-fix (lightly) — file name vs. responsibility cohesion.**
`keychain.rs` now mostly does **file** storage: auth session, pending OAuth, and the `env-secrets.json` map are all files; the keychain path is down to a single key (`ensure_runtime_data_key`) plus the dead-code-trimmed worker. The file name now under-describes ~80% of its body. Per the grammar's "a path tells you what is allowed before you open the file" (`mental-model.md:21-23`) and "name the concern" (`:226-229`), this is a real cohesion smell. I would **not** block the PR on a rename — desktop-native intentionally keeps these as flat command modules and `keychain.rs` is named in the README file-tree (so a rename is itself a spec edit and a wider blast radius) — but flag it as follow-up: either rename to something like `secrets.rs` (and update `desktop-native/README.md:31`), or split the file-backed secret helpers from the keychain worker. At minimum the module would benefit from a top-of-file doc comment stating the new split (files for recreatable secrets, keychain for the data key only) — right now that rationale only lives on `auth_session_file_path`'s comment.

---

**⚪️ Nit — naming residue.**
- `write_secret_file` rename is good and now accurate (it's the writer for *every* recreatable secret, not just dev). 👍
- `auth_session_file_path` / `pending_auth_file_path` keep their names — fine; they were never dev-specific in name.
- The old header comment said "Dev lanes therefore persist auth records as files"; the new comment is accurate. No stale "dev" wording remains in the shipped paths. Good.

---

**⚪️ Nit — error doctrine (stringly errors).**
The module is uniformly `Result<_, String>` and this PR keeps that (`write_secret_file`, `read_env_secrets_map`, the commands). The AnyHarness grammar bans stringly errors and `.to_string()` on typed errors (`mental-model.md:150-167`, `:268-277`), but that doctrine is written for `anyharness-lib`; at the **Tauri command boundary** `Result<T, String>` is the idiomatic IPC error type (serialized to the renderer), and the whole `src-tauri` surface already uses it. So this matches *local* convention, and the PR neither improves nor regresses it — **no action**. Worth noting only so the divergence from the lib grammar is a conscious one, not an accident.

---

**✅ Tests — consistent with existing style.**
`env_secrets_map_round_trips_through_file_at_0600` mirrors the existing round-trip tests (same `temp_path`/`cleanup` helpers, same `#[cfg(unix)]` `0o600` mode assertion as the auth-session test at the pre-PR `:637-645`). The two renamed tests (`_through_dev_file` → `_through_file`) correctly drop "dev" now that the path is universal. Style is coherent; nothing to change.

---

**Summary:** the only thing I would gate on is the **spec update** — per `spec-catalog.md:21` this PR should fix the now-false keychain claims at `anyharness-sidecar.md:44,94` and `desktop-native/README.md:31,52` and record the keychain-only-for-data-key decision, in this same PR. The `keychain.rs` naming/cohesion drift is a legitimate but non-blocking follow-up; the stringly-error and test conventions are fine.

pablonyx:
### ⚙️ Robustness & concurrency review (subagent)

Reviewed `keychain.rs` at head `88d10208d` against `app_config.rs` (`read_json_file` / `write_json_file_atomic`). Verdict: the file mechanics are solid (atomic, 0600, dir auto-created, corrupt-file safe), but there is a real lost-update window and a silent data-loss migration gap. Details below.

---

#### 🟥 Blocking

**1. Concurrent writers can lose env-secret updates (no lock around read-modify-write).**
`set_env_var_secret` (L329-333) and `delete_env_var_secret` (L336-342) both do `read_env_secrets_map()` → mutate → `write_secret_file(env_secrets_path())`, with no mutex over the whole map. Unlike the old keychain path, every write now rewrites the *entire* file. These are `#[tauri::command] async fn`s; Tauri does **not** serialize command invocations, so two `invoke("set_env_var_secret", …)` for *different* keys that overlap (e.g. user pastes an Anthropic key and an OpenAI key and both `saveCredentialMutation` fire — `use-local-agent-credentials.ts` exposes them via independent `mutateAsync`) will both read the pre-write map and the second writer clobbers the first key. Classic lost update, and it silently drops a credential. Fix: funnel all `env-secrets.json` mutations through a single serialized path — either a `static Mutex<()>` held across read→write, or route them through the existing keychain worker thread (which is already the serialization primitive for the data key). The auth-session / pending-auth files are whole-object writes so they are not exposed to this, only the map is.

**2. Migration gap is silent data loss, not just residue.**
The PR deletes `SERVICE` / `AUTH_SERVICE` and every `read_password`/`delete_password` against them. Nothing reads the old keychain items into the new files and nothing clears them. For an **existing** user upgrading, `~/.proliferate/env-secrets.json` and `auth-session.json` do not exist yet, so `read_env_secrets_map()` returns an empty map and `get_auth_session()` returns `None` — the user is silently logged out **and** loses all stored provider keys on first launch of the new build. If that is the accepted tradeoff (re-login once, re-enter keys once) it must be called out in the PR description and ideally surfaced in the UI; if not, a one-shot keychain→file migration on boot is needed. Either way the old keychain items are left behind as stale residue (minor, but worth a cleanup pass since the comment’s whole premise is that keychain items are unreadable junk after re-sign).

---

#### 🟧 Should-fix

**3. The `Set` pre-delete branch is now provably dead and the comment is misleading.**
After removing the `Delete` variant, the only remaining caller of `set_password` is `ensure_runtime_data_key` (L307), which always passes `RUNTIME_SERVICE`. So `if service != RUNTIME_SERVICE` (L246) can never be true — the entire delete-then-set block (L246-258) and its long justifying comment are unreachable. Worse, the comment still describes recreating auth/env items that no longer touch the keychain at all. Recommend deleting the dead branch and trimming the comment so the worker reads as what it now is: a serialized read/set for a single key. The worker itself is still justified (`keyring::Entry` is not guaranteed `Sync`, and serializing the data-key read/create avoids a double-generate race), so keep the thread — just drop the dead code.

**4. `replace_file_atomic` is not fully atomic on the conflict path.**
`write_json_file_atomic` → `replace_file_atomic` (`app_config.rs` L131-153): the happy path is `rename(tmp, dest)` (atomic, good). But the fallback on rename error does `remove_file(dest)` then `rename(tmp, dest)` (L135-146) — a window where `dest` does not exist. On Unix `rename` already replaces atomically, so this branch should essentially never fire there; it matters more on Windows. Combined with finding #1 it widens the torn/missing-file window for a concurrent `load_all_secrets_for_sidecar` reader. Low-frequency, but note it. (Not introduced by this PR — pre-existing helper.)

---

#### ✅ Confirmed solid / non-issues

- **Atomicity:** `write_json_file_atomic` is tmp+rename (`app_config.rs` L163-166). Real atomic write on the happy path. ✔
- **Corrupt-file handling is NOT silently swallowed.** The reviewing brief hypothesized a corrupt `env-secrets.json` would degrade to an empty map and drop keys — it does **not**. `read_json_file` returns `Err` on bad JSON (`app_config.rs` L126-128), so `read_env_secrets_map` propagates `Err`, and `set/delete/list` all `?`-propagate it (the write commands refuse rather than overwrite-with-empty). Only `load_all_secrets_for_sidecar` (L534, `if let Ok(map)`) and `readConfiguredEnvVarNames` (TS `catch`) intentionally degrade to empty — appropriate for those read-only/launch paths. `NotFound` → `Ok(None)` → empty map is the correct "first run" behavior. ✔
- **First-write directory creation:** `write_json_file_atomic` does `create_dir_all(parent)` (`app_config.rs` L156-159), so a missing `~/.proliferate` at first login is created. ✔
- **0600:** set after write (L54-62) and asserted in the new test. Note the tmp file is written with default umask before the `set_permissions` on the final path — a sub-millisecond window where the temp file may be group/other-readable. Nit-level for a single-user home dir; tightening would mean creating the temp with `OpenOptions::mode(0o600)`.
- **`load_all_secrets_for_sidecar` ordering / `list_configured_env_var_names` semantics are preserved.** Both still iterate `KNOWN_ENV_VARS` and filter by presence, so ordering and the "only known vars" contract are unchanged from the keychain version; only the backing store moved. ✔

Net: mechanics are good; ship-blockers are the unlocked map writes (#1) and the silent migration loss (#2).

pablonyx:
### 🔒 Security & auth-spec review (subagent)

Reviewed against `product-auth.md`, `server/guides/auth.md`, `agent-auth.md`, `agent-auth-bifrost-byok.md`, and `desktop-native/README.md` + its specs. The reinstall/re-sign keychain-ACL problem is real and the file-based fix is a reasonable direction (matches the `install_id` precedent). But the change ships without the cleanup/hardening that a "move secrets to disk" PR needs, and it skips the required spec update. Details below.

#### Blocking

1. **Stale prod secrets are left resident in the keychain — and the PR deletes the only code that could clean them up.** This diff removes `delete_password` + `KeychainRequest::Delete` entirely (`keychain.rs` hunks dropping lines ~178-182 and ~306-318) and adds no migration. Every existing release user keeps their **real refresh token** at `com.proliferate.app.auth/desktop_session` and provider API keys at `com.proliferate.app.env/*` sitting in the macOS keychain forever — orphaned, unreadable by the new build, never overwritten. The PR body says "No migration: existing users re-auth / re-enter provider keys once," but re-auth writes to the *new file* and never touches the old item. You had the exact primitive (`delete_password`) in your hand and deleted it. A one-time best-effort delete of `AUTH_SERVICE`/`SERVICE` accounts on first run after upgrade should run before that plumbing is removed. Leaving long-lived bearer credentials in a store the app no longer manages is strictly worse than either "all keychain" or "all file."

2. **Windows/Linux get NO file protection — this is a regression, not parity.** The `0600` chmod in `write_secret_file` is `#[cfg(unix)]`, and there is no `#[cfg(windows)]` ACL path anywhere in `src-tauri/src/` (grep confirms only macOS cfgs elsewhere). Desktop is a multi-platform release target and the PR body states the bug was "confirmed in production (signed release)." Before this change, `keyring = "3"` backed Windows with Credential Manager (DPAPI, user-scoped, encrypted) and Linux with Secret Service. After it, **Windows and Linux store the refresh token + every provider API key as a plaintext JSON file with default/ inherited ACLs** — on Windows that can be readable by any process in the user's session and is *not* user-data-encrypted. The PR's "0600 files, dev-tool norm" framing is macOS-only; the `0600` comment block in the diff is misleading for the platforms where it silently no-ops. At minimum gate the file backend to where you can enforce perms, or add a `#[cfg(windows)]` ACL tightening, and correct the comment.

#### Should-fix

3. **Spec hygiene: the new storage model is undocumented and contradicts the one written rule.** `desktop-native/README.md:52` is the authoritative Secrets rule: *"Sidecar launch secrets come from `commands/keychain.rs`; do not persist provider secrets in app config JSON."* You technically keep the entry point in `keychain.rs` and use a *separate* `env-secrets.json` (not `config.json`), so it's arguably within the letter — but the spirit is "provider secrets don't live as loose JSON in the app home," which is precisely what `env-secrets.json` now is. The repo rule is "update the relevant spec in the same PR." This PR changes the secret-storage model for auth session + pending OAuth + all provider creds and updates **zero** specs. Update `desktop-native/README.md` (the Secrets row + a note that auth/pending/provider creds are now `0600` files under the app home, only the data key stays in the keychain) and add a line to `product-auth.md` Desktop section (today it only covers GitHub primary sign-in + Set/Change password — nothing on where the desktop session is stored). The `agent-auth.md`/`bifrost-byok` specs are unaffected (those creds live in Bifrost/cloud, never on the desktop disk) — fine to say so explicitly.

4. **0600 is applied AFTER an atomic rename of a default-umask temp file — there's a plaintext-readable window (TOCTOU).** `write_secret_file` calls `write_json_file_atomic` (writes `tmp-<pid>` with default umask, typically `0644`, then `rename` into place) and only *then* chmods the final path to `0600` (`keychain.rs:42-50`). `rename` preserves the temp file's mode, so the destination is briefly `0644` before the chmod lands, and the temp file itself was `0644` on disk the whole time it existed. A local user/process polling `~/.proliferate` can read the token in that window. The keychain had no such window. Fix: create the temp file `0600` from the start (set mode before/at write, e.g. `OpenOptions` with `.mode(0o600)` on unix) rather than chmod-after-rename.

5. **Parent dir is created world-traversable.** `create_dir_all(~/.proliferate)` (via `write_json_file_atomic`) uses default perms (~`0755`), not `0700`. On macOS `~` is usually `0700` so contents are mostly shielded by the home dir, but the secrets directory itself is group/other-traversable and you're relying entirely on the per-file `0600`. Consider `0700` on the app dir when it holds secrets.

6. **Exfil surface for long-lived refresh tokens is materially larger and unacknowledged in-product.** `AuthSessionRecord.refresh_token` is long-lived (no rotation/short-TTL is documented in `product-auth.md`; mobile stores its refresh token in SecureStore, web uses an HttpOnly cookie — both OS/browser-protected). A plaintext `~/.proliferate/auth-session.json` is now captured by Time Machine, `rsync`/dotfile backups, `tar ~`, cloud-synced home dirs, and any "copy my config" support flow. The `gh`/`aws` analogy holds for *API keys the user can rotate*, but a session refresh token grants silent re-auth to the Proliferate account until server-side revoked. If this tradeoff is accepted, it should be a deliberate, documented decision in `product-auth.md`, and ideally paired with server-side refresh-token rotation so a leaked file ages out.

#### Nit / consistency

7. **"Only the data key needs at-rest encryption" is internally inconsistent given where the data lives.** The data key (`ANYHARNESS_DATA_KEY`) encrypts the AnyHarness SQLite at `default_anyharness_runtime_home_path()` = `~/.proliferate/anyharness` (`app_config.rs:85-87`). So the encrypted blob it protects sits in the *same* `~/.proliferate` directory whose plaintext token files you just accepted as fine to leave on disk. Keeping the key in the keychain is defensible (a plaintext key file would defeat the cipher), but the security note's logic — "the one item that genuinely needs at-rest encryption stays in the keychain" — doesn't hold when the long-lived account refresh token next to it is now plaintext. The asymmetry is fine as an engineering call; just don't frame the tokens as not needing protection.

8. **Comment drift.** The retained doc comment still reads like the dev-only rationale ("Dev builds are ad-hoc signed... Dev lanes therefore persist auth records as files") in spirit; the new block is better but the `#[cfg(unix)]` `0600` is described as the protection for "every recreatable secret file" without the Windows caveat (see #2). Tests were renamed `*_through_file` (good) but still only assert perms under `#[cfg(unix)]`, so CI will never catch the Windows gap.

Net: the *direction* is sound and the reinstall bug is worth fixing, but as written this PR (a) abandons live refresh tokens + API keys in the old keychain, (b) silently downgrades Windows/Linux from OS-encrypted stores to unprotected plaintext, (c) has a brief world-readable window even on macOS, and (d) ships no spec update for a change that explicitly touches the documented Secrets boundary. I'd hold for at least items 1-3.

pablonyx:
### ✅ Review addressed (commit `c988be799`)

Thanks for the reviews — fixes pushed:

| Finding | Resolution |
|---|---|
| **Orphaned secrets left in keychain** (security #1, robustness #2) | `purge_legacy_keychain_secrets()` — one-time, best-effort deletion of the old auth session / pending / provider-key keychain items, run on the first auth read. (Restored `delete_password` + `KeychainRequest::Delete` for it.) |
| **`env-secrets.json` lost-update race** (robustness #1) | `SECRET_FILE_LOCK` serializes all secret-file writes (the full read-modify-write for env, and the auth/pending writes). |
| **TOCTOU perms window** (security should-fix, robustness nit) | `write_secret_file` now creates the temp at `0600` from the start (atomic create→write→rename); no chmod-after-write window. |
| **Dead `Set` pre-delete branch** (robustness #3) | Removed — the data key is the only writer left, so the worker `Set` is now a plain set. |
| **Spec not updated** (conventions #1, security spec-hygiene) | Added a **Local Secrets** section to `anyharness-sidecar.md` + fixed the two stale "keychain secrets" lines and the README `keychain.rs` / Secrets-rule entries. |
| **Cross-platform (Windows/Linux) encryption regression** (security #1) | The desktop release matrix is **macOS-only** (`macos-15` + `-intel`; no Windows/Linux runners), so this is moot for shipped builds — documented in code + the spec, with an explicit "revisit if Windows/Linux desktop ships" note (Windows has no `0600` path; both have user-scoped OS keychains). Not implementing keychain-on-Windows for an unshipped target. |
| **`keychain.rs` name vs responsibility** (conventions #2, nit) | Left as a follow-up — it's named in the README file-tree, so a rename is its own spec edit; out of scope here. |

`cargo check` clean (zero warnings); `cargo test commands::keychain` 4/4.

pablonyx:
### 🧭 Conventions & spec re-review (subagent, 2nd pass)

Re-reviewed at the updated head (spec fixes from `c988be799`). The spec-hygiene gate from the first pass is **resolved** — the new docs match the code. One genuinely-stale pair of dev-profile lines remains (pre-existing, not introduced here); everything else is clean.

---

**✅ RESOLVED — spec update landed and is accurate.**
The new **Local Secrets** section in `anyharness-sidecar.md` and the README edits faithfully describe the code in `commands/keychain.rs`. Cross-checked every claim:

- File names `auth-session.json` / `pending-auth.json` / `env-secrets.json` — match `auth_session_file_path`/`pending_auth_file_path`/`env_secrets_path`. ✔
- `env-secrets.json` as a `{name: value}` map — matches `read_env_secrets_map` (`HashMap<String, String>`). ✔
- `0600` files under the durable app home (`~/.proliferate`, dev `~/.proliferate-local`) — matches `write_secret_file` (`OpenOptions::mode(0o600)` set **at creation**, not chmod-after). ✔
- "data key stays in the keychain (`com.proliferate.app.runtime`), generated on first use" — matches `RUNTIME_SERVICE` + `ensure_runtime_data_key`. ✔
- The one-time best-effort keychain purge — matches `purge_legacy_keychain_secrets` (`AtomicBool` once-per-process, called from `get_auth_session`). ✔
- macOS-only / "revisit for Windows-Linux" caveat — matches the code's header comment. ✔

The two formerly-false lines (`anyharness-sidecar.md` "keychain-backed local secrets" / "keychain secrets") are now "local secrets (see [Local Secrets](#local-secrets))", and the restart-rules list points at the same section. The README file-tree comment and Secrets rule both name the file storage + keychain-only-for-data-key split correctly.

**Anchor links valid.** `[Local Secrets](#local-secrets)` (used twice in `anyharness-sidecar.md` and once in the README) resolves to `## Local Secrets` under GitHub's slug rules (lowercase, spaces→`-`). ✔

---

**🟡 Should-fix (lightly) — one stale dev-Keychain claim pair the first pass didn't catch.** Two dev-profile specs still say desktop auth/pending live in the **dev Keychain**, which is now wrong (and was already wrong even pre-PR, since the old `native_dev_profile()` branch persisted them as files — this PR just made the model uniform):

- `specs/developing/local/README.md:236` — *"Desktop auth sessions and pending-auth entries are profile-scoped in the dev Keychain. Native provider credentials and AnyHarness runtime data keys remain user-level local secrets."*
- `specs/developing/local/dev-profiles.md:64-65` — *"…are profile-scoped in the dev Keychain so per-profile databases do not reuse each other's login tokens."*

Both describe a keychain dev path the code no longer takes (auth/pending/env are now `~/.proliferate-local/*.json`). The "profile-scoping comes from a per-profile app home" intent is still true — only the "Keychain" word is stale. Non-blocking and arguably out of the original review's scope (it flagged only the desktop-native pair), but since this PR is the one that makes the storage model uniform, fixing these two lines here would close the loop. (For the record: `specs/codebase/primitives/claiming.md:495,812` also mention "OS keychain", but those are about a *future* direct-attach JWT design, not this auth-session/env path — leave them.)

---

**✅ RESOLVED — dead `Set` pre-delete branch + comment drift.** The `if service != RUNTIME_SERVICE` block and its stale justifying comment are gone; the worker `Set` is now a plain set with an accurate comment. The data key is the only remaining writer. ✔

---

**⚪️ New-code conventions — consistent.**
- `LEGACY_*` consts: good rename (the old `SERVICE`/`AUTH_SERVICE`/account names now carry the `LEGACY_` prefix), grouped together with a comment tying them to `purge_legacy_keychain_secrets`. Placement next to the live `RUNTIME_SERVICE`/data-key consts is right. ✔
- `SECRET_FILE_LOCK: Mutex<()>` static: placed at module top with a comment naming exactly what it guards (the env-map read-modify-write + temp-path collisions). Held across read→write in all three mutating commands. Consistent with the file's existing `OnceLock` worker-channel pattern. ✔
- New helpers (`env_secrets_path`, `read_env_secrets_map`, `write_secret_file`, `purge_legacy_keychain_secrets`): named for the concern, doc-commented, sit beside the existing path/IO helpers. `write_secret_file` (ex-`write_dev_auth_record`) is now accurately named for *every* recreatable secret. ✔

---

**⚪️ Deferred `keychain.rs` → `secrets.rs` rename — still reasonable to defer.** The file is now ~80% file-backed storage, so the name under-describes it (this is also logged as `[R2|medium|M]` in `anyharness-grammar-adoption-backlog.md:187`). But the rename touches the README file-tree (its own spec edit) and widens blast radius, and the new top-of-comment + the Local Secrets spec section now capture the split in prose — so deferring is the right call for this PR, not a gap. I'd keep it tracked as the existing R2 backlog item rather than re-filing.

---

**Summary:** The spec gate is satisfied — the Local Secrets section, the README edits, and the anchor links are all accurate against the code, and the dead keychain branch is removed. The only thing I'd add (optional, non-blocking) is correcting the two **dev-Keychain** lines in `local/README.md:236` and `dev-profiles.md:64-65`, since they now misdescribe dev storage. New-code naming/placement is consistent; the file rename is reasonably deferred. No correctness re-review here (that was the robustness/security passes).

pablonyx:
### 🔒 Security re-review (subagent, 2nd pass)

Re-verified `keychain.rs` at head + the spec changes against the round-1 findings. Net: **the three blocking issues are genuinely resolved**; two prior should-fix/nits remain (both accepted-tradeoff level, not ship-blockers). No new vulnerabilities found.

#### ✅ Resolved (verified, not just claimed)

**1. Orphaned keychain secrets — RESOLVED.** `purge_legacy_keychain_secrets()` deletes exactly the right pairs: `LEGACY_AUTH_SERVICE`/`desktop_session`, `LEGACY_AUTH_SERVICE`/`desktop_pending_auth`, and `LEGACY_ENV_SERVICE`/each of the 7 `KNOWN_ENV_VARS`. The `LEGACY_*` consts match the pre-PR `AUTH_SERVICE`/`SERVICE`/account literals exactly, so it targets the real old items. Pending-auth and all env vars are covered — nothing is left behind. Best-effort/ignore-errors is correct (a keychain delete never decrypts, so it works even on items whose ACL no longer trusts this build; missing item is the norm on fresh installs).
   - **Trigger is correct.** I traced it end to end: `get_auth_session` → `getStoredAuthSession` (`lib/access/tauri/auth.ts:103`) → `bootstrapAuth` (`orchestration-bootstrap.ts:56`) → app-mounted `useAuthBootstrap` `useEffect`. It runs on **every** launch. The only path that returns *before* `getStoredAuthSession` is the `isDevAuthBypassed()` branch — and that gate requires `import.meta.env.DEV` (`auth-mode.ts:32`), which is false in any signed release. So in production the purge fires on first boot of the new build. Good.
   - **AtomicBool guard:** `swap(true, Relaxed)` is fine — no race. Worst case under a startup-double-invoke, two threads each run the idempotent best-effort deletes once; relaxed ordering is sufficient since the deletes carry no cross-thread data dependency.

**2. 0600 temp / TOCTOU window — RESOLVED.** `write_secret_file` now creates the temp via `OpenOptions` with `.mode(0o600)` on unix **before any bytes are written** (create → `write_all` → `sync_all` → `rename`). There is no chmod-after-rename window; the destination inherits 0600 from the temp through `rename`. The new `env_secrets_map_round_trips_through_file_at_0600` test asserts it. Temp (`path.with_extension("tmp")`) and target share the same parent dir, so they are on the same filesystem — `rename` cannot `EXDEV`-fail across a mount boundary.

**3. Windows/Linux plaintext regression — claim VERIFIED, moot.** I checked the matrix myself: `.github/workflows/release-desktop.yml` (lines 84-90) builds only `macos-15` (aarch64) and `macos-15-intel` (x86_64). The `runner_os == windows` / ubuntu conditionals sprinkled through the build steps are dead branches — no matrix entry ever sets `runner_os: windows`, so **no Windows/Linux desktop artifact is produced**. The "macOS-only, revisit if those targets ship" framing in both the code comment and the new spec section is accurate and adequately documented.

Also confirmed: the dead `Set` pre-delete branch is gone; the worker `Set` is now a plain `set_password` used only by the data key (`ensure_runtime_data_key`). **No secret is dual-written** to both file and keychain — the only remaining `read_password`/`set_password` callers are the `RUNTIME_SERVICE` data key. No stale read path remains (all auth/pending/env reads go through the files).

#### 🟡 Remaining (carried over, non-blocking)

**4. `~/.proliferate` dir is still ~0755, not 0700.** `write_secret_file`’s `create_dir_all(parent)` uses default perms. On macOS `~` is typically 0700 so the 0600 files are shielded in practice, but the secrets dir itself is group/other-**traversable**, so protection rests entirely on the per-file mode. Tightening the app dir to 0700 when it holds secrets would be defence-in-depth. (Was security #5 / greptile last round.)

**5. Refresh-token-on-disk exfil surface unacknowledged in-product.** `auth-session.json` now holds a long-lived `refresh_token` in plaintext, newly capturable by Time Machine / dotfile backups / `tar ~` / cloud-synced homes. The `0600`-file approach is a reasonable, documented engineering call, but if accepted it is worth (a) a one-line note in `product-auth.md`’s Desktop section and (b) server-side refresh-token rotation so a leaked file ages out. (Was security #6.) Not a blocker for this PR.

#### Verdict
The auth/secret-at-rest fixes are **solid and correctly verified** — purge targets the right items and is reliably triggered in prod, the 0600 temp closes the TOCTOU window, and the macOS-only claim holds. Items 4-5 are hardening/tradeoff follow-ups, not regressions. From a security standpoint this is good to merge.

pablonyx:
### ⚙️ Robustness re-review (subagent, 2nd pass)

Re-reviewed the file at the pushed fixes (`c988be799` / current head) — full diff + the prior three reviews + `app_config.rs` helpers + every caller of these commands (`lib.rs` setup, `runtime.rs` restart). Verdict: **all four robustness fixes are correctly implemented; no deadlock, no recursive-lock, no lost-update window remains.** Two small NEW items below, both non-blocking.

---

#### ✅ Prior findings — RESOLVED

**1. Lost-update race on `env-secrets.json` (was 🟥 blocking).** RESOLVED. `set_env_var_secret` (L399-404) and `delete_env_var_secret` (L409-416) take `SECRET_FILE_LOCK` and hold `_guard` across the **full** read→mutate→write — `read_env_secrets_map()` and `write_secret_file()` both run under the guard, which drops at function-scope end after the write returns. The two-writers-different-keys clobber is closed. The auth/pending whole-object writes are also under the lock (L429, L447).

**2. Dead `Set` pre-delete branch (was 🟧).** RESOLVED. The `if service != RUNTIME_SERVICE { delete… }` block is gone; the `Set` handler (L272-287) is now a plain `set_password`, and the comment correctly states the data key (`RUNTIME_SERVICE`) is the only remaining writer — verified: `ensure_runtime_data_key` (L376) is the sole `set_password` caller and always passes `RUNTIME_SERVICE`. No behavior change for that writer (it never pre-deleted). ✔

**3. TOCTOU 0600 window (was 🟧 / security / Greptile P1).** RESOLVED. `write_secret_file` (L74-100) opens the temp with `OpenOptions::mode(0o600)` **before** writing, then `sync_all` + `rename`. No default-umask temp, no chmod-after-rename window. Note it uses its OWN single `std::fs::rename` (L97) rather than `app_config::replace_file_atomic`, so it also dodges that helper's non-atomic remove-then-rename fallback the 1st pass flagged. ✔

**4. Orphaned legacy keychain secrets (was 🟥 security / robustness #2).** RESOLVED as a best-effort purge. `purge_legacy_keychain_secrets` (L357-367) restores `delete_password` + `KeychainRequest::Delete`, AtomicBool-gates to once-per-process, deletes the old auth/pending/provider items, swallows all errors. (Reminder, not a regression: this purges, it does not migrate — existing users still re-auth + re-enter keys once. The PR body states this; fine if intended.) ✔

#### ✅ Re-confirmed still-true (1st pass)
Atomic write (temp→single rename), `create_dir_all` on first write (L75-77), corrupt file ⇒ `Err` not empty-map (`read_json_file` returns `Err` on bad JSON; `read_env_secrets_map` + set/delete/list `?`-propagate; only `load_all_secrets_for_sidecar` L618 and the data-key read deliberately degrade). `load_all_secrets_for_sidecar` ordering / known-vars filter unchanged. ✔

#### ✅ Deadlock / blocking audit (explicit)
- **No recursive lock / no keychain-under-lock.** Every locked path (`set_env_var_secret`, `delete_env_var_secret`, `set_auth_session`, `set_pending_auth`) calls only `app_dir_path` (pure env reads), `read_json_file`, `write_secret_file` — all pure filesystem, none re-acquire `SECRET_FILE_LOCK` and none touch the keychain worker. No locked path calls another locked path. ✔
- **Purge from a hot read path is fine.** `get_auth_session` → `purge_legacy_keychain_secrets` holds **no** file lock while it round-trips the keychain worker, so it cannot block any file writer. It does up to 9 synchronous keychain deletes on the worker thread on first call (it is an `async` command on Tauri's pool, so it does not block the UI thread), once per process. Acceptable.
- **Reader needs no lock.** `load_all_secrets_for_sidecar` reads `env-secrets.json` lock-free; temp→rename means it sees old-complete or new-complete, never torn — atomic-rename is sufficient here. ✔
- **Lock poisoning** handled on every `.lock()` (`map_err(|_| "secret file lock poisoned")`). ✔

---

#### 🟡 NEW — minor, non-blocking

**A. `clear_auth_session` / `clear_pending_auth` do NOT take `SECRET_FILE_LOCK`.** L436-438 / L454-456 call `delete_file_if_exists` (a bare `remove_file` on the **final** path) outside the guard. For these whole-object files it cannot corrupt or tear anything — a clear racing a set resolves to either "file present" or "file absent", both consistent, and the set writes the temp then renames so the delete (final path) and the write (temp path) never touch the same inode. The only effect is last-writer-wins ordering between a logout and a concurrent session write, which is already inherent. So: correct as-is, but the asymmetry (sets locked, clears unlocked) is a latent foot-gun if `env-secrets.json` ever grows a partial-clear that needs read-modify-write. Worth a one-line comment, or take the lock in the clears for uniformity. Not blocking.

**B. No parent-dir fsync after rename (durability nit).** `write_secret_file` does `file.sync_all()` (good — content is durable) but does not fsync the **directory** after `rename`, so a power-loss immediately after the rename can lose the rename (dir entry not yet on disk) on some filesystems — you would fall back to the prior file, not corruption. Same minor gap the 1st pass noted on the old helper; just inherited. Acceptable for recreatable secrets.

**C. (Re-flag, unchanged) app dir is not `0700`.** `create_dir_all(~/.proliferate)` still uses default (~0755); protection rests entirely on per-file 0600. Prior reviews already raised this and the author scoped it out. Just noting this PR did not change it — and that the codebase already has the exact primitive (`commands/cloud_worker.rs:548` does `set_permissions(0o700)` on a dir), so if you ever want to tighten, the pattern is in-repo.

---

**Net:** the concurrency model is sound and every blocking finding from round 1 is genuinely fixed (not papered over). I would not gate on A/B/C — they are comment/durability/defense-in-depth nits on already-acceptable behavior. Ship-ready from a robustness standpoint.

pablonyx:
@greptiel

pablonyx:
@greptile

PR #732, opened 2026-06-18
fix(anyharness): re-fork zero-turn Claude fork children instead of dead load_session

pablonyx:
## What & why

A forked Claude session that is **created but never prompted** before its agent process recycles became **permanently unpromptable**. On the next prompt, startup resumed the fork child's native session with `LoadNativeNoFallback`; the Claude ACP agent returned `Resource not found` (the Claude Code SDK only writes a session transcript after the first *completed* turn, so a never-prompted fork has no file on disk); and because the strategy forbids fallback, every Retry re-failed.

This was validated end-to-end: anyharness logs → launch-policy code → the `claude-agent-acp` wrapper's error mapping (`"No conversation found with session ID"` → `resourceNotFound`) → SDK persistence semantics → the actual on-disk Claude store (the dead fork child + two never-prompted sibling forks have **no** `.jsonl`; the parent and every prompted session do) → the real incident DB (child `last_prompt_at` NULL; 245 `turn_started` events copied from the parent by the fork snapshot).

The fix restores the behavior already prescribed in `specs/.../src/sessions.md`: *"adapters whose fork ids are process-local until first prompt, such as Claude, start the child actor with `fork_from_native`."*

## Change

- **`launch_policy.rs`** (pure decision layer): a zero-turn fork child (no `last_prompt_at`) re-forks from the parent (`ForkFromNative`) instead of issuing a dead `LoadNativeNoFallback`. A child that has run its own turn keeps `LoadNativeNoFallback` (re-forking would drop its own turns). Falls back to the child's own native id if no parent native id resolves — never regresses below prior behavior.
- **`startup.rs`**: widened parent-native-id resolution to `is_fork_child && last_prompt_at.is_none()`.
- **Signal choice**: keys on `last_prompt_at`, **not** `has_turn_started_event` — the fork snapshot copies the parent's `turn_started` events into the child, so that signal is always set for forks. Documented the asymmetry inline to prevent a future "unification" regression.
- **`sessions.md`**: fork invariants updated in the same PR, including the tip-fork divergence as a documented known limitation.

## Tests

- `cargo test -p anyharness-lib --lib` → **713 passed, 0 failed** (1 pre-existing ignored).
- New: zero-turn → `ForkFromNative`; zero-turn + no parent → last-resort `LoadNativeNoFallback`; IO-layer parent-resolution test; store-level snapshot-pollution regression guard. Reconciled the two tests that asserted the old buggy behavior.

## Known limitation

Re-fork is tip-only (ACP has no message-indexed fork), so a zero-turn child re-forked after the parent advanced sees the parent's current tip, not its frozen snapshot — strictly better than the prior permanent failure.

## Follow-ups (out of scope)

- Close superseded native sessions on repeated re-fork-before-first-prompt (bounded leak).
- Divergence telemetry comparing child snapshot vs parent current `MAX(seq)`.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
@@ -37,20 +37,7 @@ pub(super) fn choose_startup_strategy(
     facts: &SessionStartupFacts,
 ) -> anyhow::Result<SessionStartupStrategy> {
     if facts.is_fork_child {
-        if let Some(native_session_id) = facts.native_session_id.clone() {
-            return Ok(SessionStartupStrategy::LoadNativeNoFallback(
-                native_session_id,
-            ));
-        }
-        let parent_native_session_id = facts
-            .fork_parent_native_session_id
-            .clone()
-            .ok_or_else(|| anyhow::anyhow!("fork child is missing its parent link"))?
-            .filter(|value| !value.trim().is_empty())
-            .ok_or_else(|| anyhow::anyhow!("fork parent is missing native session id"))?;
-        return Ok(SessionStartupStrategy::ForkFromNative {
-            parent_native_session_id,
-        });
+        return choose_fork_child_strategy(facts);
     }
 
     let Some(native_session_id) = facts.native_session_id.clone() else {
@@ -72,6 +59,72 @@ pub(super) fn choose_startup_strategy(
     }
 
     Ok(SessionStartupStrategy::ResumeSeqFreshNative)
+}
+
+/// Startup strategy for a fork child.
+///
+/// A fork child only has a durable, reloadable native session once it has run
+/// its own first turn. For adapters whose fork ids are process-local until that
+/// first prompt (Claude — see `specs/.../src/sessions.md` fork invariants), the
+/// eagerly-recorded native id is valid only while the original actor stays
+/// alive; after a cold restart-before-first-prompt the agent has no transcript
+/// for it and `load_session` returns `Resource not found`. So:
+///
+/// - child has run its own turn (`has_last_prompt_at`): its native transcript is
+///   durable agent-side — load it with no fallback (re-forking would lose the
+///   child's own turns).
+/// - zero-turn child: re-fork from the parent native id (`fork_from_native`),
+///   which is what the spec prescribes for these adapters. Fall back to the
+///   child's own (possibly stale) native id only when no parent can be
+///   resolved, so we never regress below the prior behavior.
+///
+/// Note the deliberate asymmetry with the non-fork Claude branch above, which
+/// keys on `has_turn_started_event`: that signal is unusable here because the
+/// fork transcript snapshot copies the parent's `turn_started` events into the
+/// child (`store/links.rs::insert_fork_session_with_link_and_event_snapshot`),
+/// so it is `true` for every fork child regardless of its own activity. Only
+/// `has_last_prompt_at` distinguishes "child has run since the fork." Do not
+/// unify the two branches.
+fn choose_fork_child_strategy(
+    facts: &SessionStartupFacts,
+) -> anyhow::Result<SessionStartupStrategy> {
+    if let Some(native_session_id) = facts.native_session_id.clone() {
+        if facts.has_last_prompt_at {
+            return Ok(SessionStartupStrategy::LoadNativeNoFallback(
+                native_session_id,
+            ));
+        }
+        // Zero-turn child: prefer re-forking from the parent; otherwise fall
+        // back to the child's own native id rather than failing the launch.
+        if let Some(parent_native_session_id) = resolved_parent_native_session_id(facts) {
+            return Ok(SessionStartupStrategy::ForkFromNative {
+                parent_native_session_id,
+            });
+        }
+        return Ok(SessionStartupStrategy::LoadNativeNoFallback(
+            native_session_id,
+        ));
+    }
+
+    // No native id of its own: must re-fork from the parent.
+    let parent_native_session_id = resolved_parent_native_session_id(facts).ok_or_else(|| {
+        match facts.fork_parent_native_session_id {
+            Some(_) => anyhow::anyhow!("fork parent is missing native session id"),
+            None => anyhow::anyhow!("fork child is missing its parent link"),
+        }
+    })?;
+    Ok(SessionStartupStrategy::ForkFromNative {
+        parent_native_session_id,
+    })
+}
+
+/// The parent's native session id, if one was resolved and is non-empty.
+fn resolved_parent_native_session_id(facts: &SessionStartupFacts) -> Option<String> {
+    facts
+        .fork_parent_native_session_id
+        .clone()
+        .flatten()
+        .filter(|value| !value.trim().is_empty())
 }
 
 /// Precondition: a closed session row never launches.
@@ -199,10 +252,54 @@ mod tests {
     }
 
     #[test]
-    fn loads_fork_children_without_fresh_fallback() {
+    fn loads_started_fork_children_without_fresh_fallback() {
+        // A fork child that has run its own turn has a durable native
+        // transcript; load it with no fallback (re-forking would lose the
+        // child's own turns).
         let mut facts = facts();
         facts.is_fork_child = true;
         facts.native_session_id = Some("fork-native".to_string());
+        facts.has_last_prompt_at = true;
+        facts.fork_parent_native_session_id = Some(Some("parent-native".to_string()));
+
+        let strategy = choose_startup_strategy(&facts).expect("strategy");
+        assert_eq!(
+            strategy,
+            SessionStartupStrategy::LoadNativeNoFallback("fork-native".to_string())
+        );
+    }
+
+    #[test]
+    fn reforks_zero_turn_fork_child_with_native_id_from_parent() {
+        // The bug case: the child has an eagerly-recorded native id but has
+        // never run its own turn, so that id is process-local and may be dead
+        // after a cold restart. Re-fork from the parent instead of issuing a
+        // no-fallback load that would brick the session.
+        let mut facts = facts();
+        facts.is_fork_child = true;
+        facts.native_session_id = Some("stale-fork-native".to_string());
+        facts.has_last_prompt_at = false;
+        facts.fork_parent_native_session_id = Some(Some("parent-native".to_string()));
+
+        let strategy = choose_startup_strategy(&facts).expect("strategy");
+        assert_eq!(
+            strategy,
+            SessionStartupStrategy::ForkFromNative {
+                parent_native_session_id: "parent-native".to_string()
+            }
+        );
+    }
+
+    #[test]
+    fn zero_turn_fork_child_falls_back_to_own_native_id_when_parent_unresolvable() {
+        // Last resort: if a zero-turn child cannot resolve a parent native id,
+        // try its own (possibly stale) native id rather than failing the launch
+        // — never regress below the prior behavior.
+        let mut facts = facts();
+        facts.is_fork_child = true;
+        facts.native_session_id = Some("fork-native".to_string());
+        facts.has_last_prompt_at = false;
+        facts.fork_parent_native_session_id = Some(None);
 
         let strategy = choose_startup_strategy(&facts).expect("strategy");
         assert_eq!(
--- a/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
@@ -80,9 +80,11 @@ pub(super) fn choose_startup_strategy(
 /// - a process-local-fork (Claude) child that has not run yet: re-fork from the
 ///   parent native id (`fork_from_native`), which is what the spec prescribes.
 ///   Fall back to the child's own (possibly stale) id only when no parent can be
-///   resolved, so we never regress below the prior behavior — that floor still
-///   succeeds for a same-process resume and fails identically to today when the
-///   native session is genuinely gone.
+///   resolved. This strategy is only computed on a cold start (a live handle
+///   short-circuits before `choose_session_startup_strategy`), so by then a
+///   process-local id is already dead and the fallback fails identically to the
+///   prior behavior — never worse, and it keeps a clean path for durable-fork
+///   adapters whose recorded id is still valid.
 ///
 /// Residual window: `has_last_prompt_at` flips at turn start
 /// (`actor/turn/active.rs::update_last_prompt_at`), slightly before Claude
@@ -90,7 +92,7 @@ pub(super) fn choose_startup_strategy(
 /// leave a recorded `last_prompt_at` with a non-durable native id. There is no
 /// lossless local recovery for a child that has started its own turn (re-forking
 /// from the parent would drop that turn), so this narrow case is left to the
-/// existing error surface; see the follow-up in the PR/spec.
+/// existing error surface (tracked as a follow-up).
 ///
 /// Note the deliberate asymmetry with the non-fork Claude branch above, which
 /// keys on `has_turn_started_event`: that signal is unusable here because the
@@ -102,7 +104,7 @@ pub(super) fn choose_startup_strategy(
 fn choose_fork_child_strategy(
     facts: &SessionStartupFacts,
 ) -> anyhow::Result<SessionStartupStrategy> {
-    let fork_id_is_process_local = facts.agent_kind == AgentKind::Claude.as_str();
+    let fork_id_is_process_local = fork_id_is_process_local(&facts.agent_kind);
 
     if let Some(native_session_id) = facts.native_session_id.clone() {
         // Durable-fork adapters, or a child that has already run its own turn:
@@ -135,6 +137,16 @@ fn choose_fork_child_strategy(
     Ok(SessionStartupStrategy::ForkFromNative {
         parent_native_session_id,
     })
+}
+
+/// Whether an adapter's fork ids are process-local until the child's first
+/// prompt (vs durable at fork time). Single source for the distinction that
+/// also gates `fork.rs::child_actor_forks`; the two must stay in lockstep — a
+/// child only reaches the zero-turn "stale native id" state if it was forked on
+/// the child actor here. Currently Claude is the only process-local adapter; a
+/// typed adapter capability would be the longer-term home (see PR follow-up).
+pub(super) fn fork_id_is_process_local(agent_kind: &str) -> bool {
+    agent_kind == AgentKind::Claude.as_str()
 }
 
 /// The parent's native session id, if one was resolved and is non-empty.
--- a/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
@@ -31,8 +31,8 @@ pub(super) struct SessionStartupFacts {
     pub has_turn_started_event: bool,
 }
 
-/// Pure startup-strategy matrix. Behavior-equivalent to the pre-split
-/// store-coupled decision; only the fact gathering moved out.
+/// Pure startup-strategy matrix: gathered facts in, a `SessionStartupStrategy`
+/// out. All fact gathering lives in `startup.rs`; this stays data-in/data-out.
 pub(super) fn choose_startup_strategy(
     facts: &SessionStartupFacts,
 ) -> anyhow::Result<SessionStartupStrategy> {
@@ -63,20 +63,34 @@ pub(super) fn choose_startup_strategy(
 
 /// Startup strategy for a fork child.
 ///
-/// A fork child only has a durable, reloadable native session once it has run
-/// its own first turn. For adapters whose fork ids are process-local until that
-/// first prompt (Claude — see `specs/.../src/sessions.md` fork invariants), the
-/// eagerly-recorded native id is valid only while the original actor stays
-/// alive; after a cold restart-before-first-prompt the agent has no transcript
-/// for it and `load_session` returns `Resource not found`. So:
+/// When a fork child's recorded native id is reloadable depends on the adapter:
 ///
-/// - child has run its own turn (`has_last_prompt_at`): its native transcript is
-///   durable agent-side — load it with no fallback (re-forking would lose the
-///   child's own turns).
-/// - zero-turn child: re-fork from the parent native id (`fork_from_native`),
-///   which is what the spec prescribes for these adapters. Fall back to the
-///   child's own (possibly stale) native id only when no parent can be
-///   resolved, so we never regress below the prior behavior.
+/// - adapters with durable fork ids (e.g. Codex) get a reloadable native id at
+///   fork time, so a recorded id always loads with no fallback.
+/// - adapters whose fork ids are process-local until first prompt (Claude — see
+///   `specs/.../src/sessions.md` fork invariants) only durably persist the
+///   child's native session after its first turn. Until then the eagerly
+///   recorded id is valid only while the original actor stays alive; after a
+///   cold restart-before-first-prompt the agent has no transcript for it and
+///   `load_session` returns `Resource not found`.
+///
+/// So for a child that has its own native id:
+/// - already ran its own turn (`has_last_prompt_at`), or a durable-fork adapter:
+///   load it with no fallback (re-forking would lose the child's own turns).
+/// - a process-local-fork (Claude) child that has not run yet: re-fork from the
+///   parent native id (`fork_from_native`), which is what the spec prescribes.
+///   Fall back to the child's own (possibly stale) id only when no parent can be
+///   resolved, so we never regress below the prior behavior — that floor still
+///   succeeds for a same-process resume and fails identically to today when the
+///   native session is genuinely gone.
+///
+/// Residual window: `has_last_prompt_at` flips at turn start
+/// (`actor/turn/active.rs::update_last_prompt_at`), slightly before Claude
+/// persists the transcript, so a crash during the child's very first turn can
+/// leave a recorded `last_prompt_at` with a non-durable native id. There is no
+/// lossless local recovery for a child that has started its own turn (re-forking
+/// from the parent would drop that turn), so this narrow case is left to the
+/// existing error surface; see the follow-up in the PR/spec.
 ///
 /// Note the deliberate asymmetry with the non-fork Claude branch above, which
 /// keys on `has_turn_started_event`: that signal is unusable here because the
@@ -88,14 +102,19 @@ pub(super) fn choose_startup_strategy(
 fn choose_fork_child_strategy(
     facts: &SessionStartupFacts,
 ) -> anyhow::Result<SessionStartupStrategy> {
+    let fork_id_is_process_local = facts.agent_kind == AgentKind::Claude.as_str();
+
     if let Some(native_session_id) = facts.native_session_id.clone() {
-        if facts.has_last_prompt_at {
+        // Durable-fork adapters, or a child that has already run its own turn:
+        // the recorded native id is reloadable.
+        if facts.has_last_prompt_at || !fork_id_is_process_local {
             return Ok(SessionStartupStrategy::LoadNativeNoFallback(
                 native_session_id,
             ));
         }
-        // Zero-turn child: prefer re-forking from the parent; otherwise fall
-        // back to the child's own native id rather than failing the launch.
+        // Zero-turn process-local-fork (Claude) child: prefer re-forking from
+        // the parent; otherwise fall back to the child's own native id rather
+        // than failing the launch.
         if let Some(parent_native_session_id) = resolved_parent_native_session_id(facts) {
             return Ok(SessionStartupStrategy::ForkFromNative {
                 parent_native_session_id,
@@ -300,6 +319,24 @@ mod tests {
         facts.native_session_id = Some("fork-native".to_string());
         facts.has_last_prompt_at = false;
         facts.fork_parent_native_session_id = Some(None);
+
+        let strategy = choose_startup_strategy(&facts).expect("strategy");
+        assert_eq!(
+            strategy,
+            SessionStartupStrategy::LoadNativeNoFallback("fork-native".to_string())
+        );
+    }
+
+    #[test]
+    fn loads_non_claude_zero_turn_fork_child_without_refork() {
+        // Durable-fork adapters keep their recorded native id even with no first
+        // prompt; only process-local (Claude) fork ids re-fork on zero turns.
+        let mut facts = facts();
+        facts.is_fork_child = true;
+        facts.agent_kind = "codex".to_string();
+        facts.native_session_id = Some("fork-native".to_string());
+        facts.has_last_prompt_at = false;
+        facts.fork_parent_native_session_id = Some(Some("parent-native".to_string()));
 
         let strategy = choose_startup_strategy(&facts).expect("strategy");
         assert_eq!(
--- a/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/startup.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/startup.rs
@@ -29,15 +29,19 @@ use super::{
 };
 
 /// Resolve steps only — gather the durable facts, then let the pure policy in
-/// `launch_policy` pick the strategy. The parent lookup keeps today's gating
-/// (only fork children without their own native id pay for it).
+/// `launch_policy` pick the strategy. The parent lookup is gated: a fork child
+/// needs its parent's native id whenever it has not yet run its own turn
+/// (`last_prompt_at` unset) — either because it never had a native id, or
+/// because its eagerly-recorded one is process-local and may be dead after a
+/// cold restart-before-first-prompt. A fork child that has already run keeps
+/// its durable native id and skips the lookup.
 pub(super) fn choose_session_startup_strategy(
     record: &SessionRecord,
     session_store: &SessionStore,
 ) -> anyhow::Result<SessionStartupStrategy> {
     let is_fork_child =
         session_store.has_inbound_link_relation(&record.id, SessionLinkRelation::Fork)?;
-    let fork_parent_native_session_id = if is_fork_child && record.native_session_id.is_none() {
+    let fork_parent_native_session_id = if is_fork_child && record.last_prompt_at.is_none() {
         session_store
             .find_parent_by_inbound_link_relation(&record.id, SessionLinkRelation::Fork)?
             .map(|parent| parent.native_session_id)
--- a/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/startup.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/startup.rs
@@ -29,12 +29,15 @@ use super::{
 };
 
 /// Resolve steps only — gather the durable facts, then let the pure policy in
-/// `launch_policy` pick the strategy. The parent lookup is gated: a fork child
-/// needs its parent's native id whenever it has not yet run its own turn
-/// (`last_prompt_at` unset) — either because it never had a native id, or
-/// because its eagerly-recorded one is process-local and may be dead after a
-/// cold restart-before-first-prompt. A fork child that has already run keeps
-/// its durable native id and skips the lookup.
+/// `launch_policy` pick the strategy. The parent lookup is gated to fork
+/// children that have not yet run their own turn (`last_prompt_at` unset): the
+/// policy may need the parent native id to re-fork — either because the child
+/// never had a native id, or because its eagerly-recorded one is process-local
+/// and may be dead after a cold restart-before-first-prompt. This intentionally
+/// over-fetches for durable-fork (non-Claude) zero-turn children, where the
+/// policy ignores the parent id; that is a single harmless row read kept here so
+/// the resolve gate doesn't have to duplicate the adapter distinction. A fork
+/// child that has already run keeps its durable native id and skips the lookup.
 pub(super) fn choose_session_startup_strategy(
     record: &SessionRecord,
     session_store: &SessionStore,
--- a/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/tests.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/tests.rs
@@ -470,7 +470,10 @@ fn choose_startup_strategy_keeps_non_claude_agents_on_native_load_path() {
 }
 
 #[test]
-fn choose_startup_strategy_loads_fork_children_without_fresh_fallback() {
+fn choose_startup_strategy_loads_started_fork_children_without_fresh_fallback() {
+    // A fork child that has already run its own turn (`last_prompt_at` set) has
+    // a durable native transcript; load it with no fallback and skip the parent
+    // lookup.
     let db = Db::open_in_memory().expect("open db");
     seed_workspace(&db);
 
@@ -482,6 +485,7 @@ fn choose_startup_strategy_loads_fork_children_without_fresh_fallback() {
     let mut child = session_record("claude");
     child.id = "fork-child".to_string();
     child.native_session_id = Some("fork-native".to_string());
+    child.last_prompt_at = Some("2026-03-25T00:05:00Z".to_string());
     let link = link_record(
         "fork-link",
         SessionLinkRelation::Fork,
@@ -498,6 +502,46 @@ fn choose_startup_strategy_loads_fork_children_without_fresh_fallback() {
     assert_eq!(
         strategy,
         SessionStartupStrategy::LoadNativeNoFallback("fork-native".to_string())
+    );
+}
+
+#[test]
+fn choose_startup_strategy_reforks_zero_turn_fork_child_with_native_id() {
+    // The bug case end-to-end through the IO layer: a fork child with an
+    // eagerly-recorded native id but no first prompt resolves the parent native
+    // id (widened gating) and re-forks instead of issuing a dead no-fallback
+    // load.
+    let db = Db::open_in_memory().expect("open db");
+    seed_workspace(&db);
+
+    let store = SessionStore::new(db);
+    let mut parent = session_record("claude");
+    parent.id = "parent-session".to_string();
+    parent.native_session_id = Some("parent-native".to_string());
+    store.insert(&parent).expect("insert parent");
+
+    let mut child = session_record("claude");
+    child.id = "fork-child".to_string();
+    child.native_session_id = Some("stale-fork-native".to_string());
+    child.last_prompt_at = None;
+    let link = link_record(
+        "fork-link",
+        SessionLinkRelation::Fork,
+        "parent-session",
+        "fork-child",
+    );
+    store
+        .insert_session_with_link(&child, &link)
+        .expect("insert fork child and link");
+
+    let strategy =
+        choose_session_startup_strategy(&child, &store).expect("select startup strategy");
+
+    assert_eq!(
+        strategy,
+        SessionStartupStrategy::ForkFromNative {
+            parent_native_session_id: "parent-native".to_string()
+        }
     );
 }
 
--- a/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/tests.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/tests.rs
@@ -585,6 +585,105 @@ fn choose_startup_strategy_forks_unstarted_fork_children_from_parent_native_id()
 }
 
 #[test]
+fn choose_startup_strategy_reforks_snapshot_polluted_zero_turn_fork_child() {
+    // End-to-end guard against the snapshot-pollution trap: build the child via
+    // the real fork snapshot (which copies the parent's `turn_started` events),
+    // so `has_turn_started_event(child)` is true. A zero-turn Claude child must
+    // still re-fork from the parent — proving the policy keys on `last_prompt_at`
+    // and not on the polluted `turn_started` signal. This test fails if anyone
+    // reverts the fork branch to key on `has_turn_started_event`.
+    let db = Db::open_in_memory().expect("open db");
+    seed_workspace(&db);
+
+    let store = SessionStore::new(db);
+    let mut parent = session_record("claude");
+    parent.id = "parent-session".to_string();
+    parent.native_session_id = Some("parent-native".to_string());
+    store.insert(&parent).expect("insert parent");
+    store
+        .append_event(&SessionEventRecord {
+            id: 0,
+            session_id: "parent-session".to_string(),
+            seq: 1,
+            timestamp: "2026-03-25T00:01:00Z".to_string(),
+            event_type: "turn_started".to_string(),
+            turn_id: Some("turn-1".to_string()),
+            item_id: None,
+            payload_json: r#"{"type":"turn_started"}"#.to_string(),
+        })
+        .expect("append parent turn");
+
+    let mut child = session_record("claude");
+    child.id = "fork-child".to_string();
+    child.native_session_id = Some("stale-fork-native".to_string());
+    child.last_prompt_at = None;
+    let link = link_record(
+        "fork-link",
+        SessionLinkRelation::Fork,
+        "parent-session",
+        "fork-child",
+    );
+    store
+        .insert_fork_session_with_link_and_event_snapshot(&child, &link)
+        .expect("insert fork child with snapshot");
+
+    // Precondition: the snapshot polluted the child's turn_started signal.
+    assert!(
+        store
+            .has_turn_started_event("fork-child")
+            .expect("has_turn_started_event"),
+        "snapshot should have copied the parent's turn_started into the child"
+    );
+
+    let strategy =
+        choose_session_startup_strategy(&child, &store).expect("select startup strategy");
+
+    assert_eq!(
+        strategy,
+        SessionStartupStrategy::ForkFromNative {
+            parent_native_session_id: "parent-native".to_string()
+        }
+    );
+}
+
+#[test]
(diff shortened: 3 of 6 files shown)


anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
```diff
@@ -74,6 +61,72 @@ pub(super) fn choose_startup_strategy(
...
+///   resolved, so we never regress below the prior behavior.
+///
+/// Note the deliberate asymmetry with the non-fork Claude branch above, which
+/// keys on `has_turn_started_event`: that signal is unusable here because the
+/// fork transcript snapshot copies the parent's `turn_started` events into the
+/// child (`store/links.rs::insert_fork_session_with_link_and_event_snapshot`),
+/// so it is `true` for every fork child regardless of its own activity. Only
+/// `has_last_prompt_at` distinguishes "child has run since the fork." Do not
+/// unify the two branches.
+fn choose_fork_child_strategy(
```
pablonyx, line 104: **[spec-conformance review]** Purity is clean here, and the doctrine is followed well. `choose_fork_child_strategy` and `resolved_parent_native_session_id` are pure data-in/data-out (no `&self`, IO, clock, or uuid), matching the module charter (launch_policy.rs:1-7) and the pure-policy law in `guides/mental-model.md:94-101` and `specs/session-engine.md:100-103`. Keeping the decision in launch_policy.rs and the fact-gathering in startup.rs is exactly the split the spec prescribes. The `Do not unify the two branches` note (the `turn_started` snapshot pollution) is a genuinely load-bearing comment — good to have it pinned to the code.

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/startup.rs
```diff
@@ -29,15 +29,19 @@ use super::{
...
+/// cold restart-before-first-prompt. A fork child that has already run keeps
+/// its durable native id and skips the lookup.
 pub(super) fn choose_session_startup_strategy(
     record: &SessionRecord,
     session_store: &SessionStore,
 ) -> anyhow::Result<SessionStartupStrategy> {
     let is_fork_child =
         session_store.has_inbound_link_relation(&record.id, SessionLinkRelation::Fork)?;
-    let fork_parent_native_session_id = if is_fork_child && record.native_session_id.is_none() {
+    let fork_parent_native_session_id = if is_fork_child && record.last_prompt_at.is_none() {
```
pablonyx, line 47: **[tests review]** The IO-layer (`startup.rs`) gating change is the load-bearing line of this fix (`record.native_session_id.is_none()` -> `record.last_prompt_at.is_none()`), but the new IO-layer tests don't cover the last-resort fallback through the store. `choose_startup_strategy_reforks_zero_turn_fork_child_with_native_id` (line 509) and `choose_startup_strategy_loads_started_fork_children_without_fresh_fallback` (line 473) cover Some-prompt and zero-turn-with-resolvable-parent, but there is no IO-layer test for: zero-turn fork child WITH its own native id whose parent has no native id -> falls back to its own `LoadNativeNoFallback`. That path is only unit-tested with synthetic `SessionStartupFacts` (`zero_turn_fork_child_falls_back_to_own_native_id_when_parent_unresolvable`), so the actual store wiring (parent lookup returning `Some(None)`) is unverified end-to-end. Suggest adding an IO test: parent inserted with `native_session_id = None`, child zero-turn with `native_session_id = Some(...)`, assert `LoadNativeNoFallback(child-native)`.

pablonyx, line 47: **[spec-conformance review]** Correct IO placement — the widened gating belongs here, not in the pure policy, per `specs/session-engine.md:100-103` (startup.rs resolves; launch_policy decides). One layering nit: the gate condition (`last_prompt_at.is_none()`) now has to mirror the policy's branching by hand, so the "when do we need the parent?" rule effectively lives in two files. Today they agree: the policy only consults `fork_parent_native_session_id` on the zero-turn fork-child path, which is exactly when this gate fetches it. But if the policy's branch condition ever changes, an over-narrow gate here would silently feed `None` and regress to a no-fallback load. Worth a one-line comment tying this predicate to `choose_fork_child_strategy`'s zero-turn branch so the coupling is explicit (the docstring update at lines 31-37 explains the *why* well but not that the two must stay in sync). Functionally this is fine as written.

anyharness/crates/anyharness-lib/src/domains/sessions/store/tests/links.rs
```diff
@@ -34,6 +34,64 @@ fn insert_session_with_link_rolls_back_child_when_link_insert_fails() {
...
+        "snapshot must copy the parent's turn_started into the child"
+    );
+
+    // CLEAN signal: a never-prompted fork child has no last_prompt_at, so the
+    // launch policy can use it to detect "child has not run since the fork".
+    let reloaded = store
+        .find_by_id("child-session")
+        .expect("find child")
+        .expect("child exists");
+    assert_eq!(
```
pablonyx, line 89: **[tests review]** `fork_snapshot_pollutes_turn_started_but_not_last_prompt_at` (links.rs:45) is a good guard for the *signal source*, but as written it is one assertion short of guarding the *decision*. It proves `has_turn_started_event(child) == true` and `last_prompt_at == None`, which documents why the policy must NOT key on `turn_started` — but nothing here actually fails if a future refactor switches the policy back to `turn_started`. The two facts live in this store-layer test; the policy that consumes them lives in `launch_policy.rs`/`startup.rs`. Consider strengthening this into a true end-to-end regression: build the fork child via `insert_fork_session_with_link_and_event_snapshot` (so the snapshot pollution is real, not simulated by `insert_session_with_link`), then call `choose_session_startup_strategy` and assert `ForkFromNative`. Note the existing IO test `choose_startup_strategy_reforks_zero_turn_fork_child_with_native_id` (tests.rs:509) uses plain `insert_session_with_link`, which does NOT copy the parent's `turn_started` events — so no current test exercises the policy against a genuinely snapshot-polluted child. That is the exact scenario the bug was about.

specs/codebase/structures/anyharness/src/sessions.md
```diff
@@ -258,6 +258,22 @@ Fork invariants:
...
+  (`fork_from_native`) rather than loading the child's recorded native id, even
+  if one was eagerly persisted at fork creation. Loading it after a cold
+  restart-before-first-prompt returns `Resource not found` and, with no
+  fallback, bricks the session. Once `last_prompt_at` is set the child loads its
+  own native id with no fallback (re-forking would drop the child's own turns).
+  If a zero-turn child cannot resolve a parent native id, it falls back to its
+  own (possibly stale) native id rather than failing the launch. This decision
+  keys on `last_prompt_at`, not on `turn_started`: the transcript snapshot below
+  copies the parent's `turn_started` events into the child, so that signal is
+  always set for forks. Known limitation: re-fork is tip-only (see below), so a
```
pablonyx: **[spec-conformance review]** This update is accurate and well-placed — it sits directly under the `fork_from_native` invariant it refines, matches the code (`last_prompt_at` keying, no-fallback-once-started, parent-fallback-to-own-native), and correctly calls out the `turn_started` snapshot pollution. Two notes:

1. Style: `specs/README.md:24-27` bans tentative caveats / "eventually" architecture in canonical docs but explicitly *allows* naming a limitation as operating law with the canonical rule stated. The `Known limitation: re-fork is tip-only ... full fidelity needs ACP message-indexed fork, which is unavailable` clause is borderline — it reads as a future-work promise. Since it is stated as a current invariant (and `generic ACP fork support means tip fork only` already appears at line 265), consider trimming to just the present-tense behavior + a cross-reference, rather than the `needs X, which is unavailable` framing.

2. Completeness: the paragraph only describes the case where the child *has* an eagerly-persisted native id. The code's other fork-child path — child with no native id of its own re-forks from the parent, erroring if the parent link/native id is missing (`choose_fork_child_strategy`, launch_policy.rs:111-129) — is left implicit. The prior invariant text covered "start the child actor with `fork_from_native`" generally, so this is minor, but a half-sentence acknowledging the no-native-id child keeps the invariant complete.

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
pablonyx, line 104: **[tests review]** Missing edge: non-Claude fork child. `choose_fork_child_strategy` runs for ALL fork children regardless of `agent_kind` — the `is_fork_child` branch in `choose_startup_strategy` short-circuits before the `agent_kind != Claude` check at line 50. So a Codex/Gemini/Cursor fork child also gets re-forked / `LoadNativeNoFallback`, never `LoadNative`. The spec doc-comment here (lines 64-87) frames the whole rationale around "adapters whose fork ids are process-local until first prompt (Claude)", which implies the zero-turn re-fork behavior is Claude-specific — but the code applies it unconditionally. Either that is intended (then add a test pinning a non-Claude fork child to its expected strategy so the behavior is documented), or it is a latent bug for adapters with durable fork ids (then the branch needs an `agent_kind` gate + a test). Every fork-child test in this PR uses `agent_kind = "claude"`; the non-Claude fork path is completely untested.

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
```diff
@@ -74,6 +61,72 @@ pub(super) fn choose_startup_strategy(
...
+                parent_native_session_id,
+            });
+        }
+        return Ok(SessionStartupStrategy::LoadNativeNoFallback(
+            native_session_id,
+        ));
+    }
+
+    // No native id of its own: must re-fork from the parent.
+    let parent_native_session_id = resolved_parent_native_session_id(facts).ok_or_else(|| {
```
pablonyx, line 131: **[tests review]** Two untested edges around parent resolution at the IO layer:

1. **Deleted/missing parent row for a zero-turn child with its own native id.** At the IO layer, `find_parent_by_inbound_link_relation` can return `None` (no parent row) which becomes `fork_parent_native_session_id = None`. For a zero-turn child that HAS its own native id, the unit branch falls back to `LoadNativeNoFallback(own)` — but there is no IO test where the fork link exists yet the parent row is gone. Worth one: insert child+link via `insert_session_with_link`, never insert the parent, assert `LoadNativeNoFallback(child-native)` (not an error). This proves the orphaned-link case degrades gracefully rather than bricking.

2. **Empty/whitespace parent native id, no own native id.** `resolved_parent_native_session_id` filters blank strings, and `errors_when_fork_parent_is_missing_native_session_id` (launch_policy.rs:336) covers the `Some("   ")` and `Some(None)` cases at the unit level — but only at the unit level. No IO test feeds an empty-string parent `native_session_id` through the store. Given the bug class here is "wrong signal propagated from the store," an IO-level test for the blank-parent error message would be cheap insurance.

pablonyx:
**[spec-conformance review]** Two repo-wide notes that do not pin to a changed line:

1. **Stale module docstring (launch_policy.rs:34-35).** `choose_startup_strategy` is still documented as `Pure startup-strategy matrix. Behavior-equivalent to the pre-split store-coupled decision; only the fact gathering moved out.` That claim is now false — this PR deliberately changes the fork-child behavior (zero-turn re-fork instead of no-fallback load). It is outside the diff so I could not attach an inline comment, but it should be updated in this PR to avoid asserting behavior-equivalence that no longer holds.

2. **Spec coverage is otherwise complete.** I checked the other authoritative docs for this area: `specs/session-engine.md` describes launch_policy/startup at the role level (lines 100-103) and does not enumerate individual startup strategies, so no strategy-table update is owed there; `src/acp.md:67-80` mentions `startup strategy` only as a `SessionActorConfig` field, also no per-strategy table. `guides/mental-model.md` / `guides/domains.md` are doctrine, not behavior, and are not affected. So `src/sessions.md` (updated here) is the correct and sufficient spec home for this change — the spec-first/update-in-same-PR rule (`specs/README.md:5-6, 65`) is satisfied. Good adherence to the layering doctrine throughout (pure decision in launch_policy, IO in startup, regression evidence captured as a store test).

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
```diff
@@ -74,6 +61,72 @@ pub(super) fn choose_startup_strategy(
...
+    if let Some(native_session_id) = facts.native_session_id.clone() {
+        if facts.has_last_prompt_at {
+            return Ok(SessionStartupStrategy::LoadNativeNoFallback(
+                native_session_id,
+            ));
+        }
+        // Zero-turn child: prefer re-forking from the parent; otherwise fall
+        // back to the child's own native id rather than failing the launch.
+        if let Some(parent_native_session_id) = resolved_parent_native_session_id(facts) {
+            return Ok(SessionStartupStrategy::ForkFromNative {
```
pablonyx, line 121: **[correctness review]** Proliferation / unbounded agent-side leak. The bound argument in the doc comment ("re-fork is bounded because the row is overwritten and the first prompt pins the id") only bounds the *DB row* — there is exactly one `native_session_id` column, so the row never grows. It does NOT bound the agent-side native sessions. Every cold restart-before-first-prompt of a zero-turn child runs `ForkFromNative`, which issues a fresh ACP `session/fork` and produces a brand-new native session id; `persist_live_session_state` (lifecycle.rs:264) then overwrites the column with the new id. The previously-forked native session is never closed on this path — unlike the create-time path in fork.rs:167-171 (`close_native_session` on insert failure) and the error path in fork.rs:205-211 (`clear_native_session_id`). So N cold restarts before the first prompt orphan N-1 agent-side fork sessions. The bound only re-engages once a prompt sets `last_prompt_at` and flips the strategy to `LoadNativeNoFallback`. Worth either (a) closing the superseded native id before/after the re-fork when the adapter `supports_close`, or (b) documenting this as an accepted, restart-frequency-bounded leak rather than implying the row-overwrite makes it bounded.

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/tests.rs
```diff
@@ -501,6 +505,46 @@ fn choose_startup_strategy_loads_fork_children_without_fresh_fallback() {
...
+    let strategy =
+        choose_session_startup_strategy(&child, &store).expect("select startup strategy");
+
+    assert_eq!(
+        strategy,
+        SessionStartupStrategy::ForkFromNative {
+            parent_native_session_id: "parent-native".to_string()
+        }
+    );
+}
```
pablonyx, line 546: **[tests review]** Closed/dismissed fork child is unguarded by these tests. `session_is_closed` is checked in `ensure_live_session_handle` (startup.rs:179) BEFORE `choose_session_startup_strategy`, so a closed fork child never reaches this policy — but that ordering invariant has no test of its own here, and none of the new fork tests set `closed_at`/`dismissed_at`. The new code is `Resource not found`-adjacent: a dismissed-then-resumed zero-turn fork would now re-fork rather than fail. One small test asserting a closed fork child is rejected at the `session_is_closed` gate (and never re-forks) would lock the precedence in. `session_is_closed_checks_closed_at_and_status` (launch_policy.rs:349) covers the predicate but not its interaction with the fork path.

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
```diff
@@ -74,6 +61,72 @@ pub(super) fn choose_startup_strategy(
...
+                parent_native_session_id,
+            });
+        }
+        return Ok(SessionStartupStrategy::LoadNativeNoFallback(
```
pablonyx, line 125: **[correctness review]** Last-resort fallback floor — the fall-through to `LoadNativeNoFallback(native_session_id)` here is the *known-bad* state this PR exists to fix. It is reached when a zero-turn child has its own (process-local, likely dead) native id but `resolved_parent_native_session_id` returns `None`. On a cold restart that load returns `Resource not found`, and because it is `NoFallback` it bricks the session (native_session.rs:194-251 only falls back when `allows_missing_load_fallback()` is true, which it is not for `LoadNativeNoFallback`). So this "never regress below prior behavior" floor reproduces exactly the dead-end the PR is fixing, for the narrower no-parent case. When would a zero-turn fork child legitimately have *no* resolvable parent native id? The parent link is created atomically with the child (fork.rs insert_*_with_link), and `find_parent_by_inbound_link_relation` filters `closed_at IS NULL` — so the realistic causes are: parent row hard-deleted, fork link closed, or parent never got a native id. In all three, re-forking is impossible and the recorded child id is already dead, so loading it is futile. Consider whether erroring with a clear message would be a better floor than a guaranteed-to-brick no-fallback load — the user gets a real error instead of a session that silently comes up empty.

pablonyx:
**[tests review]** Overall: the new `choose_fork_child_strategy` is well covered at the **pure unit level** — all six decision branches have a dedicated test (started->LoadNoFallback; zero-turn+native+parent->ForkFromNative; zero-turn+native+no-parent->LoadNoFallback-own; no-native+parent->ForkFromNative; no-native+blank-parent->error; no-native+no-parent->error). Naming and comment style match existing conventions well, and the rename `loads_fork_children...` -> `loads_started_fork_children...` correctly disambiguates now that "fork child" alone is no longer one behavior.

The gaps are all at the **IO layer**, where the actual fix lives (the `last_prompt_at.is_none()` gating in `startup.rs`). Summary of suggested additions (details in inline comments):
1. IO test: zero-turn fork child with own native id + parent has no native id -> `LoadNativeNoFallback(own)`.
2. Turn the store-layer `fork_snapshot_pollutes_...` test into an end-to-end one (build child via `insert_fork_session_with_link_and_event_snapshot`, then call `choose_session_startup_strategy` -> `ForkFromNative`). Today no test runs the policy against a genuinely snapshot-polluted child.
3. Non-Claude fork child behavior (currently unconditional; either gate on agent_kind or pin with a test).
4. Orphaned fork link (parent row deleted) for a zero-turn child with own native id -> graceful `LoadNativeNoFallback`.
5. Closed/dismissed fork child precedence vs the new re-fork path.

None block the core fix; items 2 and 3 are the highest-value.

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
```diff
@@ -74,6 +61,72 @@ pub(super) fn choose_startup_strategy(
...
+        if facts.has_last_prompt_at {
+            return Ok(SessionStartupStrategy::LoadNativeNoFallback(
+                native_session_id,
+            ));
+        }
+        // Zero-turn child: prefer re-forking from the parent; otherwise fall
+        // back to the child's own native id rather than failing the launch.
+        if let Some(parent_native_session_id) = resolved_parent_native_session_id(facts) {
+            return Ok(SessionStartupStrategy::ForkFromNative {
+                parent_native_session_id,
```
pablonyx, line 122: **[correctness review]** New hard dependency on live fork capability at resume. `SessionStartupStrategy::ForkFromNative` produced here requires `action_capabilities.fork` at execution time (native_session.rs:256-262), and that capability is taken from the *fresh ACP initialize response* on restart (actor/startup.rs:141-150 `action_capabilities_from_acp(&init_response.agent_capabilities)`), NOT from the child's persisted `action_capabilities_json`. Behavior change vs the old `LoadNativeNoFallback` for a zero-turn child with its own native id: the old path needed no fork capability. With the new path, if the resumed agent does not advertise `fork` (downgraded/older Claude binary, or a capability regression), the actor returns a hard error and the launch fails with "agent does not advertise ACP session/fork". For Claude this is moot (always advertised), but worth a one-line note: the new path assumes fork capability is stable across restarts; if an agent could lose it, a zero-turn child becomes unresumable by either strategy (the old load was already doomed-to-brick anyway, so this is not a true regression in outcome, only in the error surface).

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
```diff
@@ -74,6 +61,72 @@ pub(super) fn choose_startup_strategy(
...
+/// fork transcript snapshot copies the parent's `turn_started` events into the
+/// child (`store/links.rs::insert_fork_session_with_link_and_event_snapshot`),
+/// so it is `true` for every fork child regardless of its own activity. Only
+/// `has_last_prompt_at` distinguishes "child has run since the fork." Do not
+/// unify the two branches.
+fn choose_fork_child_strategy(
+    facts: &SessionStartupFacts,
+) -> anyhow::Result<SessionStartupStrategy> {
+    if let Some(native_session_id) = facts.native_session_id.clone() {
+        if facts.has_last_prompt_at {
```
pablonyx: **[correctness review]** Race between `last_prompt_at` write timing and strategy selection. `last_prompt_at` is written when a prompt is *accepted/started* (actor/turn/active.rs:84-91, alongside status="running"), i.e. BEFORE the turn completes and before the native transcript is durably flushed agent-side. The strategy keys `LoadNativeNoFallback` purely on `has_last_prompt_at`. Consider: a zero-turn child gets its first prompt → `last_prompt_at` is set → the actor sends the prompt → the process crashes mid-turn before the agent has persisted the forked native session to disk. On restart `has_last_prompt_at == true` so we pick `LoadNativeNoFallback` against a native id the agent may not have durably written yet → `Resource not found` → brick. The non-fork Claude branch (launch_policy.rs:54-59) explicitly guards this same crash window by ALSO accepting `has_turn_started_event` as a durable signal. The fork branch deliberately cannot use `turn_started` (snapshot pollution), so it has no equivalent guard for the "prompt accepted but native session not yet durable" window. Whether this is reachable depends on Claude's fork-persistence ordering (does `session/fork` durably persist the child transcript synchronously before the prompt is dispatched?). If not guaranteed, this is a narrower re-occurrence of the same brick. Worth confirming the ordering or noting the residual window.

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/startup.rs
pablonyx, line 47: **[correctness review]** Gating predicate widening looks correct but note the asymmetry it creates with the policy's no-own-native-id error mapping. The lookup now runs whenever `is_fork_child && record.last_prompt_at.is_none()`. The policy's no-own-native-id branch (launch_policy.rs:110-114) maps a `None` outer Option to "fork child is missing its parent link". That mapping is only sound because a fork child with `native_session_id == None` ALWAYS has `last_prompt_at == None` (you cannot have prompted without a native session), so the lookup always runs for that branch — good, the invariant holds. But it is load-bearing and implicit: if a future change ever lets `last_prompt_at` be set with `native_session_id` still None, the lookup would be skipped, `fork_parent_native_session_id` would be `None`, and the policy would emit the misleading "missing parent link" error instead of doing the lookup. Consider asserting/documenting the coupling, or gating the lookup on `last_prompt_at.is_none() || native_session_id.is_none()` to make it robust to that invariant drifting.

specs/codebase/structures/anyharness/src/sessions.md
pablonyx: **[correctness review]** Parent-advanced divergence (tip-only re-fork) — the doc/spec acknowledge it, but flag the silent-data-correctness trap explicitly. When a zero-turn child is re-forked on restart, `ForkSessionRequest` is built from the parent's *current* `native_session_id` (the live tip), so the re-fork captures the parent's state AS OF restart, not the snapshot frozen at original fork time. The child's DB row, however, still carries the parent's transcript snapshot copied at fork creation (insert_fork_session_with_link_and_event_snapshot). After a parent-advanced re-fork these two diverge: the child's `session_events` prefix reflects the *original* parent tip, while the live native fork reflects the *new* parent tip. The agent's view (what it will actually reason over) and AnyHarness's stored transcript prefix are now inconsistent. The user sees the old snapshot in the UI but the agent answers from the newer parent state. The spec calls this "strictly better than permanent failure," which is fair, but "strictly better" undersells the correctness hazard: it is a silent semantic change to what the fork contains, not just a fidelity loss. Worth confirming whether the snapshot prefix should be re-synced (or the divergence surfaced) on a parent-advanced re-fork, or at minimum that downstream consumers tolerate prefix/live skew.

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
```diff
@@ -74,6 +61,72 @@ pub(super) fn choose_startup_strategy(
...
+            None => anyhow::anyhow!("fork child is missing its parent link"),
+        }
+    })?;
+    Ok(SessionStartupStrategy::ForkFromNative {
+        parent_native_session_id,
+    })
+}
+
+/// The parent's native session id, if one was resolved and is non-empty.
+fn resolved_parent_native_session_id(facts: &SessionStartupFacts) -> Option<String> {
```
pablonyx, line 153: **[correctness review]** `resolved_parent_native_session_id` flattening is correct. `.clone().flatten()` collapses `Option<Option<String>>` to `Option<String>`: outer `None` (no parent row found / lookup skipped) and inner `None` (parent row found but no native id) both become `None`, and the `.filter(!trim().is_empty())` additionally drops `Some("")`/`Some("  ")`. So the three "no usable parent id" cases (no row, null id, blank id) all funnel to `None`, which is what both call sites want.

One asymmetry to note for the error mapping below (lines 110-114): that match distinguishes outer `Some(_)` → "missing native session id" vs `None` → "missing parent link". But the blank-string case is `Some(Some("   "))` → matches `Some(_)` → "missing native session id" — correct. And `Some(None)` → "missing native session id" — correct. The only case that yields "missing parent link" is a true outer `None`. Since the parent lookup always runs when `native_session_id` is None (see startup.rs gating note), outer `None` here genuinely means "no parent row / link," so the message is accurate. The flattening and the error mapping are mutually consistent. No bug — just confirming the two-layer semantics hold together.

pablonyx:
## Review addressed (commit 97c69f67c)

Thanks to the three review passes. Changes made:

- **Gate the new behavior to process-local-fork adapters (Claude)** — durable-fork adapters (e.g. Codex) get a reloadable native id at fork time, so a zero-turn fork child keeps `LoadNativeNoFallback` instead of re-forking. The re-fork path was incorrectly running for all agent kinds. Added policy- and IO-layer tests for the non-Claude case.
- **Non-tautological snapshot test** — added `choose_startup_strategy_reforks_snapshot_polluted_zero_turn_fork_child`, which builds the child via the real `insert_fork_session_with_link_and_event_snapshot` (so `has_turn_started_event` is genuinely polluted) and asserts the policy still picks `ForkFromNative`. This fails if anyone reverts the branch to key on `turn_started`.
- **Stale docstring** on `choose_startup_strategy` ("behavior-equivalent to the pre-split decision") corrected.
- **Spec wording** — trimmed the "needs ACP message-indexed fork" future-work clause; reframed the tip-fork divergence as a real (accepted) divergence, not just a fidelity gap; noted the process-local vs durable-fork distinction.

Acknowledged but intentionally **not** changed here, with rationale:

- **`last_prompt_at` write-timing window** — real but narrow: `last_prompt_at` flips at turn start, slightly before Claude persists the transcript, so a crash during the child's *first* turn can leave it set with a non-durable native id. There is no lossless local recovery once a child has started its own turn (re-forking from the parent would drop that turn), so this narrow case is left to the existing error surface and documented in the code comment + tracked as a follow-up.
- **Last-resort `LoadNativeNoFallback` when no parent resolves** — kept deliberately: it still succeeds for a same-process resume and fails identically to today when the native session is genuinely gone, so it is strictly ≥ erroring. Documented inline.

Follow-ups (out of scope, tracked): close superseded native sessions on repeated re-fork-before-first-prompt; divergence telemetry comparing child snapshot vs parent current `MAX(seq)`.

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
```diff
@@ -74,6 +61,91 @@ pub(super) fn choose_startup_strategy(
...
+///   `load_session` returns `Resource not found`.
+///
+/// So for a child that has its own native id:
+/// - already ran its own turn (`has_last_prompt_at`), or a durable-fork adapter:
+///   load it with no fallback (re-forking would lose the child's own turns).
+/// - a process-local-fork (Claude) child that has not run yet: re-fork from the
+///   parent native id (`fork_from_native`), which is what the spec prescribes.
+///   Fall back to the child's own (possibly stale) id only when no parent can be
+///   resolved, so we never regress below the prior behavior — that floor still
+///   succeeds for a same-process resume and fails identically to today when the
```
pablonyx: **[deep re-review: spec/arch]** Severity: low (doc accuracy). The floor claim "still succeeds for a same-process resume" overstates what this branch does. On a same-process resume, `ensure_live_session_handle` returns the cached handle early (`startup.rs:183 get_handle`) and `choose_session_startup_strategy` is never reached — so this last-resort `LoadNativeNoFallback(own_id)` only runs on a cold restart, exactly the case where a process-local Claude id is most likely dead. And `LoadNativeNoFallback` has no `new_session` fallback (`model.rs:82-83 allows_missing_load_fallback` matches only `LoadNative`), so when the id is gone this path hard-fails. The "fails identically to today" half is correct (old code issued the same `LoadNativeNoFallback(own_id)`); the "succeeds for a same-process resume" half is misleading. Suggest dropping it or rewording to "preserves the prior behavior exactly."

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
```diff
@@ -74,6 +61,91 @@ pub(super) fn choose_startup_strategy(
...
+///   native session is genuinely gone.
+///
+/// Residual window: `has_last_prompt_at` flips at turn start
+/// (`actor/turn/active.rs::update_last_prompt_at`), slightly before Claude
+/// persists the transcript, so a crash during the child's very first turn can
+/// leave a recorded `last_prompt_at` with a non-durable native id. There is no
+/// lossless local recovery for a child that has started its own turn (re-forking
+/// from the parent would drop that turn), so this narrow case is left to the
+/// existing error surface; see the follow-up in the PR/spec.
```
pablonyx: **[deep re-review: spec/arch]** Severity: low (doc accuracy / spec-rule). "see the follow-up in the PR/spec" is now a dangling pointer. The 2nd commit "trim[med] the spec future-work clause," so `sessions.md` (lines 261-280) no longer contains any follow-up — it ends at "a real divergence, not just a fidelity gap." The follow-ups live only in the PR description, which is ephemeral and not a spec. Two-fold issue: (1) the code comment points readers to a spec follow-up that does not exist; (2) per `specs/README.md` ("Do not write cleanup promises ... or `eventually` architecture inside canonical guides"), trimming it from the spec was correct, but then the code comment should not redirect to it. Suggest: drop "; see the follow-up in the PR/spec" or replace with "left to the existing error surface." (Same applies to the residual-window paragraph generally — it is honest and code-comment-only is the right altitude per the spec's conventions; just fix the pointer.)

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/startup.rs
pablonyx, line 47: **[deep re-review: layering]** Severity: low (minor over-fetch, correct otherwise). The resolve-layer gate is now `is_fork_child && last_prompt_at.is_none()`, but the policy that consumes the parent id only re-forks when `agent_kind == claude` (`launch_policy.rs:105,110`). So for a **non-Claude** (durable-fork, e.g. Codex) zero-turn fork child, `startup.rs` runs `find_parent_by_inbound_link_relation` and the policy then ignores the result (returns `LoadNativeNoFallback` at `launch_policy.rs:110`). This is harmless — one extra parent row read, no correctness impact, and the test `loads_non_claude_zero_turn_fork_child_without_refork` confirms the policy ignores it — but it is a real (small) over-fetch and the docstring claims "needs its parent's native id whenever it has not yet run its own turn," which is not true for durable-fork adapters. No under-fetch anywhere: a Claude child that has run skips the lookup correctly. Acceptable as-is given the resolve layer must not branch on `agent_kind` policy; just note the docstring slightly overstates the need. Keeping the gate provider-agnostic at the IO layer is defensible — flagging for awareness, not a required change.

specs/codebase/structures/anyharness/src/sessions.md
```diff
@@ -258,6 +258,26 @@ Fork invariants:
...
   start the child actor with `fork_from_native`; that child actor calls ACP
   `session/fork` from the parent native id and owns the resulting live fork
+- a process-local fork id only becomes durable (reloadable via `load_session`)
+  once the child has run its own first turn. Until then — i.e. while the child's
+  `last_prompt_at` is unset — startup re-forks from the parent
+  (`fork_from_native`) rather than loading the child's recorded native id, even
+  if one was eagerly persisted at fork creation. Loading it after a cold
+  restart-before-first-prompt returns `Resource not found` and, with no
+  fallback, bricks the session. Once `last_prompt_at` is set the child loads its
+  own native id with no fallback (re-forking would drop the child's own turns).
```
pablonyx, line 268: **[deep re-review: spec accuracy / completeness]** Severity: low-med. The added paragraph is accurate against `choose_fork_child_strategy` and does not contradict the "tip fork only" invariant (line 285) — the divergence note at 275-280 is the right honesty. One completeness gap: the paragraph is framed almost entirely around process-local (Claude) forks. The durable-fork branch is mentioned ("durable-fork adapters keep loading their recorded native id," line 271) but it reads as a one-line aside, whereas the existing invariant at lines 256-257 ("adapters with durable fork ids may fork on the parent actor and then start the child with `load_session`") is the canonical statement of that path. The new text does not contradict 256-257, but a reader could miss that the zero-turn `LoadNativeNoFallback` for Codex *is* that same `load_session` path. Consider tying the two together explicitly (e.g. "durable-fork adapters take the `load_session` path described above even at zero turns"). Substantively correct — this is a clarity/completeness nit, not a drift.

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
```diff
@@ -74,6 +61,91 @@ pub(super) fn choose_startup_strategy(
...
+/// keys on `has_turn_started_event`: that signal is unusable here because the
+/// fork transcript snapshot copies the parent's `turn_started` events into the
+/// child (`store/links.rs::insert_fork_session_with_link_and_event_snapshot`),
+/// so it is `true` for every fork child regardless of its own activity. Only
+/// `has_last_prompt_at` distinguishes "child has run since the fork." Do not
+/// unify the two branches.
+fn choose_fork_child_strategy(
+    facts: &SessionStartupFacts,
+) -> anyhow::Result<SessionStartupStrategy> {
+    let fork_id_is_process_local = facts.agent_kind == AgentKind::Claude.as_str();
```
pablonyx: **[deep re-review: doctrine]** Severity: none (confirmation) + nano nit. Pure-policy law (`mental-model.md` L94, `domains.md` L250-251: no `&self`/IO/clock/UUID) is honored: `choose_fork_child_strategy` and `resolved_parent_native_session_id` are free functions over `&SessionStartupFacts`, branch on a static `AgentKind::Claude.as_str()` (a `&'static str`, not dynamic truth), and perform no effects. Good — the split is clean and the helper is a sensible extraction. Nit: gating on the stringly-compared `agent_kind == AgentKind::Claude.as_str()` rather than a typed `AgentKind` (or a `descriptor.fork_id_is_process_local()` capability) means the "is this a process-local fork adapter?" predicate is duplicated as a literal here and inverted in the non-fork branch (`agent_kind != Claude::as_str()`, L50). If a third process-local-fork adapter ever ships, both call sites must change in lockstep. Consider a single typed predicate on the agent descriptor/capability so the doctrine ("dynamic truth handed in, static truth `use`-d") stays a capability rather than a hard-coded vendor name. Not blocking.

pablonyx:
**[deep re-review: spec/arch]** Overall verdict: **Approve with minor doc nits.** The fix is correct, well-tested, and restores behavior the spec already prescribed; the 2nd commit (Claude gating + spec edit) materially improved both accuracy and honesty.

**Spec accuracy (point 1):** The edited `sessions.md` paragraph matches `choose_fork_child_strategy` exactly — Claude gating, last-resort own-id fallback, and the tip-fork divergence all line up with the code. No contradiction with the "tip fork only" (L285) or the durable-fork `load_session` invariant (L256-257). The divergence call-out (L275-280) is the right level of honesty and complies with `specs/README.md` ("name it as a migration exception / known limitation, not an eventually-promise").

**Completeness (point 2):** Process-local vs durable-fork distinction is documented and the non-Claude branch is captured, but the durable-fork case reads as an aside — see inline on sessions.md:268. The residual `last_prompt_at` window is correctly code-comment-only (not spec-worthy); good call per spec conventions.

**Layering/doctrine (point 3):** Pure-policy law honored (inline on launch_policy.rs:105). The `startup.rs` gate slightly over-fetches the parent for non-Claude zero-turn children — harmless, flagged on startup.rs:44.

**Doc/comment quality (point 4):** Mostly accurate (the `update_last_prompt_at` timing claim checks out against `actor/turn/active.rs:91`). Two overstatements: the "succeeds for a same-process resume" floor claim (launch_policy.rs:84) and the dangling "see the follow-up in the PR/spec" pointer after the spec clause was trimmed (launch_policy.rs:93).

**Other specs (point 5):** No other spec needs touching. `session-engine.md` / `session-actor.md` describe fork only at the role/folder level (native new/load/fork lives in `driver/native_session.rs`) and do not pin the zero-turn strategy, so they remain accurate. `sessions.md` was the correct and sole spec home for this behavior.

All findings are low / low-med severity and none block merge.

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
pablonyx: **[deep re-review: correctness]** **should-fix.** This gate hardcodes the process-local-vs-durable fork distinction as `== Claude`, but the spec it cites says "fork ids are process-local until first prompt, **such as Claude**" — "such as" names a *category*, not Claude alone. The non-Claude branch (line 110: `|| !fork_id_is_process_local` → `LoadNativeNoFallback`) is only safe if *every* other forkable adapter has a fork id that is durable (reloadable via `load_session`) across a cold restart. That is asserted but not verified.

Evidence the assumption is load-bearing and unguarded:
- The "durability" of non-Claude fork ids is not derived from any capability — it comes entirely from `fork.rs:88` (`let child_actor_forks = parent.agent_kind == AgentKind::Claude.as_str();`), a *separate* `== Claude` hardcode this PR did not touch. Both Claude and non-Claude funnel through the identical ACP `session/fork` call (`actor/fork/handle.rs::fork_native_session`); the only difference is whether the result is recorded at creation and loaded no-fallback.
- `LoadNativeNoFallback` explicitly forbids the missing-load fallback (`model.rs:82` `allows_missing_load_fallback` matches only `LoadNative`). So a zero-turn fork child of any process-local-fork adapter bricks with `Resource not found` — the exact bug this PR fixes for Claude.
- Forkable adapters today include Gemini, Grok, Cursor, OpenCode (any that advertise `load_session && session_capabilities.fork`, `startup.rs:486`). Their fork-id durability across restart is undocumented; Gemini in particular is an upstream ACP adapter with no spec evidence it persists a forked session before its first turn.

This is latent (the brick only triggers for a forkable non-Claude adapter whose fork id is in fact process-local), and the non-Claude zero-turn-brick existed before this PR too. But the PR had the chance to make the rule capability-driven and instead cemented the name check. Strongly prefer a capability/registry flag (e.g. `fork_id_durable`) single-sourced across `fork.rs:88`, here, and `native_session.rs:256`, so a new or reclassified adapter can't silently regress into the brick. At minimum, document that non-Claude durability is an unverified assumption tracked per-adapter.

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
```diff
@@ -74,6 +61,91 @@ pub(super) fn choose_startup_strategy(
...
+        }
+        // Zero-turn process-local-fork (Claude) child: prefer re-forking from
+        // the parent; otherwise fall back to the child's own native id rather
+        // than failing the launch.
+        if let Some(parent_native_session_id) = resolved_parent_native_session_id(facts) {
+            return Ok(SessionStartupStrategy::ForkFromNative {
+                parent_native_session_id,
+            });
+        }
+        return Ok(SessionStartupStrategy::LoadNativeNoFallback(
```
pablonyx, line 125: **[deep re-review: correctness]** **should-fix (rationale is wrong, even if the code is harmless).** The docstring justifies the last-resort floor with: *"that floor still succeeds for a same-process resume and fails identically to today when the native session is genuinely gone."* The "succeeds for a same-process resume" half is not true for the path that actually reaches this floor.

`choose_startup_strategy` / `choose_session_startup_strategy` is only invoked from `ensure_live_session_handle` (`startup.rs:243`), and only **after** the `get_handle` short-circuit fails (`startup.rs:183-191`). If a live handle exists (the genuine same-process resume), the handle is reused and no startup strategy is computed at all. So the strategy decision — and therefore this floor — is reached **only when there is no live actor**, i.e. exactly the cold-restart case where a Claude process-local fork id is invalid and `LoadNativeNoFallback` will return `Resource not found`.

Net: for a zero-turn Claude child that hits this floor (parent native id unresolvable), the floor does NOT "succeed for a same-process resume" — same-process resumes never reach it. It bricks identically to erroring. That's still ">= the prior behavior" in outcome (both fail when the parent can't be resolved and the id is dead), so the chosen code is acceptable — but the stated reason is incorrect and should be fixed so the next reader doesn't rely on a false guarantee. Either drop the "succeeds for a same-process resume" clause or replace it with "this path is only reached on cold restart (no live handle); the floor then fails identically to erroring, which is no worse than today."

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_policy.rs
pablonyx: **[deep re-review: correctness]** **nit / pressure-test of the "unfixable losslessly" claim.** Two parts:

(a) The "no durable signal distinguishes the child's own turn from the snapshot prefix" framing is slightly understated. The snapshot copies parent events into the child **preserving the parent's `seq`** (`store/links.rs:42`: `SELECT ?1, seq, ...`), and the child's own events are appended with seqs continuing past the copied max. So "child has a `session_events` row with `seq > max(copied snapshot seq)`" *is* a durable, restart-surviving, snapshot-immune signal of child activity — unlike `has_turn_started_event`, which is polluted. That's worth a sentence in the docstring, because it's the principled version of what `last_prompt_at` is being used as a proxy for, and it would also let the residual-window detection key on persisted events rather than a status column. (It does NOT, however, fix the race — see (b).)

(b) For the residual window itself, you're correct that there is no *lossless* recovery: once the child has started its own turn, re-forking drops it and loading a not-yet-flushed id fails. But "left to the existing error surface" = permanent brick, which is the same failure mode this PR exists to kill. A cheap, lossy-but-live mitigation is available and is strictly better than bricking: in that narrow window (turn started, own native id un-loadable), fall back to `ForkFromNative` from the parent — the user loses the single in-flight first turn but the session is usable again, instead of being permanently unpromptable. Today the only escape is the same dead `Resource not found` loop. Worth at least naming this as the intended follow-up rather than "existing error surface," since the latter reads as "accept the brick."

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/startup.rs
pablonyx, line 47: **[deep re-review: correctness]** **nit — second-commit gate keys diverge from the policy.** This widened the parent-lookup gate to `is_fork_child && record.last_prompt_at.is_none()`, but the policy that consumes the result (`launch_policy.rs:105-126`) only ever *uses* the parent native id for a **process-local-fork (Claude)** zero-turn child. For a non-Claude zero-turn child the policy takes the `!fork_id_is_process_local` → `LoadNativeNoFallback` branch and ignores the fetched parent id entirely.

So the fact-gathering gate is now broader than the decision gate: every non-Claude zero-turn fork child pays a `find_parent_by_inbound_link_relation` query whose result is discarded. Harmless today (one extra read on cold start), but it's an avoidable inconsistency and a small trap — a future reader could assume the parent lookup implies the parent id is consumed. Two clean options: (1) also gate the lookup on `agent_kind == Claude` so IO and policy agree, or (2) leave it broad but add a one-line comment here noting the policy may discard it for durable-fork adapters. Either keeps the IO/policy keys legible. (Flagging the correctness/consistency angle; the raw over-fetch was already noted on the prior commit.)

pablonyx:
**[deep re-review: correctness]** Overall verdict: **mergeable, but one should-fix worth addressing first.** The core fix is correct and the second commit's Claude gating is a genuine improvement over re-forking unconditionally. The pure-policy/IO split is clean, and the snapshot-pollution regression test (`reforks_snapshot_polluted_zero_turn_fork_child`) is the right non-tautological guard. I verified the bug case is real and reachable (Claude fork child: `fork.rs:88` routes through the child-actor fork so the child is created with `native_session_id = None`, then the in-process `ForkFromNative` populates it and `persist_live_session_state` records it — so at cold restart it has `native_session_id = Some && last_prompt_at = None`, exactly the brick). The non-Claude branch is also reachable and correctly tested (eager durable id at creation).

Findings posted inline, by severity:
- **should-fix** (launch_policy.rs:105): the process-local-vs-durable distinction is a hardcoded `== Claude` in three uncoordinated sites (`fork.rs:88`, here, `native_session.rs:256`), not a capability. The spec says "such as Claude" — a category. Any forkable non-Claude adapter (Gemini/Grok/Cursor/OpenCode) whose fork id is in fact process-local hits the same `LoadNativeNoFallback` brick this PR fixes for Claude. Latent and pre-existing, but cemented here; a `fork_id_durable` capability flag would be the correct single source of truth.
- **should-fix** (launch_policy.rs:123): the floor's "succeeds for a same-process resume" rationale is wrong — `choose_session_startup_strategy` is only reached when no live handle exists (`startup.rs:183` short-circuit), so same-process resumes never reach the floor. The code is acceptable; the justifying comment is not.
- **nit** (launch_policy.rs:93): a durable, snapshot-immune signal does exist (child `session_events` row with `seq > max copied snapshot seq`), and the residual first-turn window has a lossy-but-live mitigation (re-fork on un-loadable own id) that beats the permanent brick "existing error surface" implies.
- **nit** (startup.rs:44): the widened fact-gathering gate (`last_prompt_at.is_none()`) is broader than the decision gate (which only consumes the parent id for Claude), so non-Claude zero-turn children fetch a parent id that's discarded.

None are blockers. The should-fix items are correctness-of-rationale and latent-robustness, not active regressions on the shipped Claude path.

pablonyx:
## Second review round addressed (commit a0c0c3377)

Two high-reasoning re-reviews ran after re-reading the specs; both verdicts were *mergeable / approve with minor nits*. Changes:

- **Single-sourced the process-local-fork distinction** (the one should-fix). Extracted `fork_id_is_process_local(agent_kind)` and used it at both sites that must agree: the resume-side policy and `fork.rs::child_actor_forks` (fork creation). They were independently hardcoding `agent_kind == Claude`; now they can't drift. A child only reaches the zero-turn "stale native id" state if it was forked on the child actor, so coupling these is the correctness invariant. Documented a typed adapter capability as the proper longer-term home.
- **Corrected the floor docstring.** Verified `choose_session_startup_strategy` is only reached on cold start (a live handle short-circuits in `ensure_live_session_handle`), so the last-resort `LoadNativeNoFallback` fails *identically to prior behavior* when the native session is gone — it does not "succeed for a same-process resume." Comment fixed to say exactly that.
- **Dangling pointer + over-fetch docs.** Reworded the residual-window note (no stale `/spec` reference); softened the `startup.rs` doc to state the intentional, harmless one-row over-fetch for durable-fork zero-turn children.
- **Spec.** Tied the zero-turn durable-fork case to the existing `load_session` invariant so the paragraph doesn't read as a standalone aside.

On the `agent_kind == Claude` exposure for hypothetical other process-local adapters: today Claude is the only one, and the fork-creation gate (`fork.rs:88`) uses the same predicate, so a non-Claude child never reaches the buggy state. Single-sourcing means a future process-local adapter flips one function. A full capability-flag from the adapter init response is the right long-term fix and is noted as a follow-up rather than built speculatively here.

716 lib tests pass.

pablonyx:
**[final re-review]** Verdict: mergeable. No new blockers; nothing already-addressed re-raised.

Verified on `a0c0c33`:
- `fork_id_is_process_local` is used at **both** sites — `launch_policy::choose_fork_child_strategy` and `fork.rs::child_actor_forks` (via `super::launch_policy::fork_id_is_process_local`). `pub(super)` + same `runtime` module → path resolves; `cargo build -p anyharness-lib` is clean (only pre-existing dead-code warnings). No other fork-durability site still hardcodes `agent_kind == Claude` (the `persist.rs:171` Claude check is live-config restore, a separate concern, correctly untouched).
- Latest commit is a **pure** extraction + docs: the only code delta from `97c69f6` is `facts.agent_kind == AgentKind::Claude.as_str()` → `fork_id_is_process_local(&facts.agent_kind)`, and the helper body is exactly that expression. No behavioral change; the `startup.rs` gating widening landed in the first commit, untouched here.
- Floor docstring is now accurate: `ensure_live_session_handle` short-circuits on `acp_manager.get_handle` (startup.rs:186-194) before `choose_session_startup_strategy` (line 246), so the strategy is only computed on cold start — the own-id fallback fails identically to prior behavior, never "succeeds for same-process resume."
- All cross-refs valid: `LoadNativeNoFallback`/`ForkFromNative` are real enum variants; `update_last_prompt_at` exists at active.rs:91.
- 716 lib tests pass, including the new ones (`reforks_zero_turn_fork_child_with_native_id`, `reforks_snapshot_polluted_zero_turn_fork_child`, `loads_non_claude_zero_turn_fork_child_without_refork`, `zero_turn_fork_child_falls_back_to_own_native_id_when_parent_unresolvable`, `fork_snapshot_pollutes_turn_started_but_not_last_prompt_at`).

pablonyx:
@greptile

PR #735, opened 2026-06-18
fix(desktop): keep distinct local workspaces per folder+branch instead of collapsing

pablonyx:
## What & why

"New local workspace" for an already-open folder collapsed into the existing workspace — *"local workspaces all the same workspace."*

Root cause: the **backend** creates a distinct workspace row per "New local workspace" (`resolve_or_create_workspace(..., allow_existing=false)`, [identity.rs:70](anyharness/crates/anyharness-lib/src/domains/workspaces/runtime/identity.rs:70) — only *worktrees* bail-on-existing), but the **desktop** logical-workspace projection collapsed every local record sharing the same `path+branch` into one sidebar entry (`collapseExactLocalWorkspaceDuplicates`, [logical-workspaces.ts](apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces.ts)). So the new row was swallowed and its sessions appeared under the existing workspace.

A local workspace is meant to be a "project/feature thread" holding many chat tabs; starting a new feature should give a separate entry.

## Change

Relax the collapse: within a `path+branch` bucket, **every workspace that has its own chats** (`executionSummary.totalSessionCount > 0`) is surfaced as its own distinct logical workspace; only **genuinely-empty** (setup-only/stale) duplicates fold onto the top used entry (all-empty buckets still keep one representative). This keeps the sidebar clean of junk empty rows while honoring real, distinct workspaces.

Downstream is unchanged — distinct entries already get slot ids (`buildLocalSlotLogicalWorkspaceId` / `"local-slot"`) and `#2`/`#3` name suffixes (`applyDuplicateLocalNameSuffixes`). Backend unchanged.

## Tests

The old "collapses exact duplicate … prefers transcript history" test actually had **both** fixtures with sessions, so under the new rule they split — replaced it with:
- distinct-when-both-have-chats → two logical entries
- zero-session-duplicate folds onto the used record (still selectable via alias)

Ran `logical-workspaces` + full `sidebar/` + `cloud/` workspace suites: **196 pass** (via main-tree deps; fresh worktree has no node_modules).

## Notes / follow-ups

- A brand-new local workspace shows via the **pending-workspace projection** until its first turn marks it "used"; worth a manual pass in-app.
- Sanity-check no non-create flow (mobility "bring local", restart re-resolution) emits a *used* same-path+branch row that should still merge.

### DIFF
--- a/apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces.ts
+++ b/apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces.ts
@@ -143,11 +143,11 @@ function collapseExactLocalWorkspaceDuplicates(
 
   // Distinct local workspaces for the same folder+branch are kept separate so
   // each "project/feature thread" gets its own sidebar entry. Within a bucket,
-  // every workspace that has its own chats is surfaced as a distinct entry; only
-  // genuinely-empty (setup-only / stale) duplicates are folded onto the top used
-  // entry so they don't show as junk rows but stay selectable. A brand-new empty
-  // workspace is still shown immediately via the pending-workspace projection
-  // until its first turn makes it "used".
+  // a workspace is surfaced as its own distinct entry when it has its own chats
+  // OR it is the current selection (covers a just-created workspace that has no
+  // chats yet — `createLocalWorkspaceAndEnter` selects it immediately). Only
+  // genuinely-empty, unselected (setup-only / stale) duplicates are folded onto
+  // the top distinct entry so they don't show as junk rows but stay selectable.
   return Array.from(byMaterialization.values()).flatMap((bucket): CollapsedLocalWorkspace[] => {
     if (bucket.length === 1) {
       return [{
@@ -156,29 +156,37 @@ function collapseExactLocalWorkspaceDuplicates(
       }];
     }
 
-    const used = bucket.filter(workspaceHasOwnSessions);
-    const empty = bucket.filter((candidate) => !workspaceHasOwnSessions(candidate));
+    const distinct = bucket.filter(
+      (candidate) =>
+        workspaceHasOwnSessions(candidate)
+        || localWorkspaceMatchesSelection(candidate, currentSelectionId),
+    );
+    const foldable = bucket.filter(
+      (candidate) =>
+        !workspaceHasOwnSessions(candidate)
+        && !localWorkspaceMatchesSelection(candidate, currentSelectionId),
+    );
 
-    if (used.length === 0) {
-      // No chats anywhere in this folder+branch yet: keep a single
+    if (distinct.length === 0) {
+      // No chats and nothing selected in this folder+branch yet: keep a single
       // representative (preferring history/selection), aliasing the rest.
-      const representative = [...empty]
+      const representative = [...foldable]
         .sort(compareExactLocalWorkspaceDuplicateOrderForSelection(currentSelectionId))[0]!;
       return [{
         workspace: representative,
-        aliasIds: empty
+        aliasIds: foldable
           .filter((candidate) => candidate.id !== representative.id)
           .flatMap(localWorkspaceIdentityIds),
       }];
     }
 
-    const sortedUsed = [...used]
+    const sortedDistinct = [...distinct]
       .sort(compareExactLocalWorkspaceDuplicateOrderForSelection(currentSelectionId));
-    const foldedEmptyAliasIds = empty.flatMap(localWorkspaceIdentityIds);
-    return sortedUsed.map((workspace, index) => ({
+    const foldedAliasIds = foldable.flatMap(localWorkspaceIdentityIds);
+    return sortedDistinct.map((workspace, index) => ({
       workspace,
-      // Fold stale empty duplicates onto the top used entry.
-      aliasIds: index === 0 ? foldedEmptyAliasIds : [],
+      // Fold stale empty duplicates onto the first distinct entry.
+      aliasIds: index === 0 ? foldedAliasIds : [],
     }));
   });
 }
--- a/apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces.ts
+++ b/apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces.ts
@@ -122,6 +122,10 @@ function compareExactLocalWorkspaceDuplicateOrderForSelection(
   };
 }
 
+function workspaceHasOwnSessions(workspace: Workspace): boolean {
+  return (workspace.executionSummary?.totalSessionCount ?? 0) > 0;
+}
+
 function collapseExactLocalWorkspaceDuplicates(
   workspaces: readonly Workspace[],
   currentSelectionId: string | null,
@@ -137,23 +141,45 @@ function collapseExactLocalWorkspaceDuplicates(
     }
   }
 
-  return Array.from(byMaterialization.values()).map((bucket) => {
+  // Distinct local workspaces for the same folder+branch are kept separate so
+  // each "project/feature thread" gets its own sidebar entry. Within a bucket,
+  // every workspace that has its own chats is surfaced as a distinct entry; only
+  // genuinely-empty (setup-only / stale) duplicates are folded onto the top used
+  // entry so they don't show as junk rows but stay selectable. A brand-new empty
+  // workspace is still shown immediately via the pending-workspace projection
+  // until its first turn makes it "used".
+  return Array.from(byMaterialization.values()).flatMap((bucket): CollapsedLocalWorkspace[] => {
     if (bucket.length === 1) {
-      return {
+      return [{
         workspace: bucket[0]!,
         aliasIds: [],
-      };
+      }];
     }
 
-    const workspace = [...bucket]
-      .sort(compareExactLocalWorkspaceDuplicateOrderForSelection(currentSelectionId))[0]!;
-    const aliasIds = bucket
-      .filter((candidate) => candidate.id !== workspace.id)
-      .flatMap(localWorkspaceIdentityIds);
-    return {
+    const used = bucket.filter(workspaceHasOwnSessions);
+    const empty = bucket.filter((candidate) => !workspaceHasOwnSessions(candidate));
+
+    if (used.length === 0) {
+      // No chats anywhere in this folder+branch yet: keep a single
+      // representative (preferring history/selection), aliasing the rest.
+      const representative = [...empty]
+        .sort(compareExactLocalWorkspaceDuplicateOrderForSelection(currentSelectionId))[0]!;
+      return [{
+        workspace: representative,
+        aliasIds: empty
+          .filter((candidate) => candidate.id !== representative.id)
+          .flatMap(localWorkspaceIdentityIds),
+      }];
+    }
+
+    const sortedUsed = [...used]
+      .sort(compareExactLocalWorkspaceDuplicateOrderForSelection(currentSelectionId));
+    const foldedEmptyAliasIds = empty.flatMap(localWorkspaceIdentityIds);
+    return sortedUsed.map((workspace, index) => ({
       workspace,
-      aliasIds,
-    };
+      // Fold stale empty duplicates onto the top used entry.
+      aliasIds: index === 0 ? foldedEmptyAliasIds : [],
+    }));
   });
 }
 
--- a/apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces-duplicates.test.ts
+++ b/apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces-duplicates.test.ts
@@ -230,6 +230,48 @@ describe("logical workspace duplicate local records", () => {
     )?.id).toBe(logicalWorkspaces[0]?.id);
   });
 
+  it("keeps a just-created (selected, zero-session) duplicate visible next to a used sibling", () => {
+    // Regression: clicking "New local workspace" on an already-used folder makes
+    // a 0-session row that is immediately selected. It must show as its own entry
+    // (not fold into the used sibling) so selection resolves to the new workspace.
+    const checkoutPath = "/tmp/proliferate";
+    const usedLocal = {
+      ...makeWorkspace({
+        id: "local-used",
+        branch: "main",
+        updatedAt: "2026-06-02T09:10:00.000Z",
+        executionSummary: makeExecutionSummary({
+          totalSessionCount: 3,
+          liveSessionCount: 0,
+          updatedAt: "2026-06-02T09:15:00.000Z",
+        }),
+      }),
+      path: checkoutPath,
+    };
+    const justCreated = {
+      ...makeWorkspace({
+        id: "local-just-created",
+        branch: "main",
+        updatedAt: "2026-06-02T09:50:00.000Z",
+      }),
+      path: checkoutPath,
+    };
+
+    const logicalWorkspaces = buildLogicalWorkspaces({
+      localWorkspaces: [usedLocal, justCreated],
+      repoRoots: [],
+      cloudWorkspaces: [],
+      currentSelectionId: justCreated.id,
+    });
+
+    expect(logicalWorkspaces).toHaveLength(2);
+    const createdEntry = findLogicalWorkspace(logicalWorkspaces, justCreated.id);
+    const usedEntry = findLogicalWorkspace(logicalWorkspaces, usedLocal.id);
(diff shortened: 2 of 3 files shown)


apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces.ts
```diff
@@ -137,23 +141,45 @@ function collapseExactLocalWorkspaceDuplicates(
     }
   }
 
-  return Array.from(byMaterialization.values()).map((bucket) => {
+  // Distinct local workspaces for the same folder+branch are kept separate so
```
pablonyx, line 144: **[spec-conformance review]** Backend identity here is correct and matches the authoritative spec. `specs/codebase/structures/anyharness/src/workspaces.md:119-123` states local `create_workspace` is "intentionally non-idempotent by path: creating a local workspace for a repo path that already has an active local workspace inserts a fresh workspace row. These duplicate local workspaces share the same checkout and git state, but have distinct workspace ids and therefore distinct session lists and runtime session state." And `workspaces.md:107-111` explicitly assigns this projection to desktop: "The runtime does not expose a canonical-vs-sibling concept for these duplicates; desktop reconstructs that projection from the returned local workspace rows." So surfacing each used row as its own sidebar entry is exactly what the runtime spec hands to desktop — backend stays unchanged, projection lives in the right layer. No product-intent conflict with worktrees: no spec discourages same-folder local workspaces (workspaces.md blesses the distinct-row create path; worktrees are the strict/path-unique kind per `workspaces.md:125-127`).

apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces.ts
```diff
@@ -137,23 +141,45 @@ function collapseExactLocalWorkspaceDuplicates(
...
+  // each "project/feature thread" gets its own sidebar entry. Within a bucket,
+  // every workspace that has its own chats is surfaced as a distinct entry; only
+  // genuinely-empty (setup-only / stale) duplicates are folded onto the top used
+  // entry so they don't show as junk rows but stay selectable. A brand-new empty
+  // workspace is still shown immediately via the pending-workspace projection
+  // until its first turn makes it "used".
+  return Array.from(byMaterialization.values()).flatMap((bucket): CollapsedLocalWorkspace[] => {
```
pablonyx, line 151: **[spec-conformance review]** Spec-update gap (per `specs/README.md`: "Keep specs current in the same PR as the behavior they describe" / "Every durable feature, primitive, and structure rule belongs in a spec"). This PR changes the desktop logical-workspace collapse contract from "one entry per folder+branch" to "one entry per used local row, empty duplicates folded onto the top used entry." That rule is a durable structure rule, but it is currently undocumented: no spec names `collapseExactLocalWorkspaceDuplicates`, the logical-workspace model, or the dedup behavior (grep of specs/ for the file / `aliasIds` / `local-slot` / collapse rule returns nothing). The runtime spec only goes as far as `workspaces.md:107-111` ("desktop reconstructs that projection") and stops; the projection rule itself has no home. Recommend documenting it. The natural owner is the desktop logical-workspace projection — there is no dedicated frontend logical-workspace spec today, so either (a) extend `specs/codebase/structures/anyharness/src/workspaces.md` (after line 111, where it already says desktop owns this projection) with one paragraph stating the desktop rule, or (b) add it to a desktop-native/frontend spec. Either way the new rule (used rows distinct; empty/setup-only/stale duplicates folded but alias-selectable; pending projection covers brand-new empties) should be written down in this PR.

apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces.ts
```diff
@@ -122,6 +122,10 @@ function compareExactLocalWorkspaceDuplicateOrderForSelection(
   };
 }
 
+function workspaceHasOwnSessions(workspace: Workspace): boolean {
+  return (workspace.executionSummary?.totalSessionCount ?? 0) > 0;
+}
```
pablonyx, line 127: **[spec-conformance review]** Architecture / layer: this is the right layer. `workspaces.md:107-111` and :220-222 establish that the runtime stays deterministic/idempotent on resolve and deliberately does NOT model canonical-vs-sibling for duplicate local rows — "desktop reconstructs that projection." So putting the dedup/distinctness decision in the desktop projection (`collapseExactLocalWorkspaceDuplicates`) rather than the backend is consistent with the documented ownership split. The "used = has own sessions" predicate keying off `executionSummary.totalSessionCount` is a reasonable proxy for the spec's "distinct session lists" (workspaces.md:122-123).

apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces.ts
```diff
@@ -137,23 +141,45 @@ function collapseExactLocalWorkspaceDuplicates(
...
-    return {
+    const sortedUsed = [...used]
+      .sort(compareExactLocalWorkspaceDuplicateOrderForSelection(currentSelectionId));
+    const foldedEmptyAliasIds = empty.flatMap(localWorkspaceIdentityIds);
+    return sortedUsed.map((workspace, index) => ({
       workspace,
-      aliasIds,
-    };
+      // Fold stale empty duplicates onto the top used entry.
+      aliasIds: index === 0 ? foldedEmptyAliasIds : [],
```
pablonyx: **[spec-conformance review]** Minor (behavior, not spec): empty duplicates are folded onto `index === 0` of `sortedUsed` only. The sort is recency/selection-biased, so the "top used entry" can change frame-to-frame (e.g. another sibling starts running, or selection moves). That means a stale/empty duplicate's alias ids hop between sidebar entries across renders. Functionally fine for selection (any used entry resolves it), but if it ever causes the empty row to appear to "jump," consider folding empties onto the canonically-ordered first entry instead of the selection-sorted one. Worth a line in whatever spec documents this rule so the fold target is defined, not incidental.

pablonyx:
**[spec-conformance review]** Overall verdict

Logic and architecture are spec-conformant. The backend stays unchanged and is already documented to produce distinct local rows with distinct session lists (`specs/codebase/structures/anyharness/src/workspaces.md:119-123`), and the same spec explicitly hands the canonical-vs-sibling projection to desktop (`workspaces.md:107-111`). So relaxing `collapseExactLocalWorkspaceDuplicates` to surface each used row is the correct layer and matches documented product intent. No worktree-vs-local conflict: no spec discourages same-folder local workspaces — `workspaces.md` blesses the non-idempotent local-create path, with worktrees being the strict/path-unique kind.

One blocker for the repo's spec-first rule: the desktop collapse/dedup behavior this PR changes is a durable structure rule but is currently undocumented in any spec. Per `specs/README.md` ("Keep specs current in the same PR as the behavior they describe"), this PR should add that rule — recommended home is a short paragraph in `specs/codebase/structures/anyharness/src/workspaces.md` right after line 111 (where it already says desktop owns this projection), or a desktop/frontend spec. See inline comments for the exact rule to write and a minor note on the empty-duplicate fold target.

Mergeable: yes on logic/tests; please add the spec paragraph in this PR before merge to satisfy the spec-first rule.

apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces.ts
pablonyx: **[correctness review]** **HIGH — brand-new local workspace at an already-used path+branch is hidden / mis-routed.**

A freshly created *local* (non-worktree) workspace re-opened on the same folder+branch as an existing *used* one has `executionSummary.totalSessionCount === 0`, so `workspaceHasOwnSessions` is false and it lands in `empty`. Because `used.length > 0`, its identity ids are folded as aliases onto `sortedUsed[0]` (here, line 177-181) and it gets **no logical entry of its own**.

Trace of create-and-enter: the flow selects the raw `workspace.id` (use-workspace-entry-actions.ts:128-143). `findLogicalWorkspace(logicalWorkspaces, newId)` resolves that id via `aliasIds` to the **other (used)** logical entry (logical-workspace-lookup.ts:20-31), and `runWorkspaceSelection` then sets `selectedLogicalWorkspaceId` to the used entry (run-workspace-selection.ts:279). The pending-sidebar projection masks this during the pending window, but `setPendingWorkspaceEntry(null)` fires synchronously right after selection resolves (workspace-entry-finalization.ts:109) — it is **not** gated on the new workspace having sessions or getting its own row. So once the pending entry clears, the new workspace has no row and the active row is its *used* sibling, until its first turn flips it into `used`.

The old code protected exactly this: it deleted the `currentSelectionId === setupOnlyLocal.id` test (logical-workspaces-duplicates.test.ts) where a selected empty workspace was promoted to canonical instead of folded. Recommend: never fold an empty that is the current selection (treat selected-empty as its own entry), e.g. keep empties out of the fold when `localWorkspaceMatchesSelection(empty, currentSelectionId)`.

apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces.ts
```diff
@@ -137,23 +141,45 @@ function collapseExactLocalWorkspaceDuplicates(
...
-    const workspace = [...bucket]
-      .sort(compareExactLocalWorkspaceDuplicateOrderForSelection(currentSelectionId))[0]!;
-    const aliasIds = bucket
-      .filter((candidate) => candidate.id !== workspace.id)
-      .flatMap(localWorkspaceIdentityIds);
-    return {
+    const sortedUsed = [...used]
+      .sort(compareExactLocalWorkspaceDuplicateOrderForSelection(currentSelectionId));
+    const foldedEmptyAliasIds = empty.flatMap(localWorkspaceIdentityIds);
+    return sortedUsed.map((workspace, index) => ({
```
pablonyx: **[correctness review]** **MEDIUM — fold target and canonical (`baseLogicalId`) holder can be different workspaces; cloud merge keys on the wrong one.**

Here the empties' alias ids are folded onto `sortedUsed[0]`, where `sortedUsed` is ordered by `compareExactLocalWorkspaceDuplicateOrderForSelection` (selection -> execution priority -> recency). But the caller re-sorts the returned list by a *different* comparator — `compareLocalWorkspaceCanonicalOrder` (createdAt asc, then id) — at logical-workspaces.ts:234, and assigns `baseLogicalId` to that list's `index 0` (line 237-239); the rest get `buildLocalSlotLogicalWorkspaceId`.

With multiple *used* locals these two orderings disagree, so:
1. The entry that received the folded empty aliases (most-active/selected used) is often NOT the one that gets `baseLogicalId` (oldest-created used). Alias lookup still works (aliases travel with their entry), but
2. **cloud/mobility workspaces merge only onto `baseLogicalId`** (lines 258-300). So the cloud thread for this repo+branch attaches to the *oldest-created* used local, which may not be the active/intended one — and never to a slot-id local. Previously this was a no-op because the bucket always collapsed to a single entry, so the re-sort had nothing to reorder; now it is load-bearing.

If the intent is "most relevant used local is canonical and owns the cloud merge", fold and canonical assignment should use the same ordering. Worth a deliberate decision + a test asserting which used local owns `baseLogicalId` when a cloud workspace is present.

apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces.ts
```diff
@@ -137,23 +141,45 @@ function collapseExactLocalWorkspaceDuplicates(
...
 
-    const workspace = [...bucket]
-      .sort(compareExactLocalWorkspaceDuplicateOrderForSelection(currentSelectionId))[0]!;
-    const aliasIds = bucket
-      .filter((candidate) => candidate.id !== workspace.id)
-      .flatMap(localWorkspaceIdentityIds);
-    return {
+    const sortedUsed = [...used]
+      .sort(compareExactLocalWorkspaceDuplicateOrderForSelection(currentSelectionId));
+    const foldedEmptyAliasIds = empty.flatMap(localWorkspaceIdentityIds);
```
pablonyx: **[correctness review]** **MEDIUM — silent selection redirect when a previously-selected empty duplicate becomes folded.**

The `used.length === 0` branch correctly biases the representative toward the current selection via `compareExactLocalWorkspaceDuplicateOrderForSelection`. But the `used.length > 0` branch (lines 175-182) does NOT exempt a selected *empty* workspace: it always lands in `empty` and is folded as an alias onto `sortedUsed[0]`.

Concrete regression vs. the deleted test (logical-workspaces-duplicates.test.ts, the removed `currentSelectionId: setupOnlyLocal.id` case): if the user currently has an empty duplicate selected and any sibling in the same path+branch gains a session (e.g. a background/cloud-linked sibling, or the sibling the user *also* opened), on the next `buildLogicalWorkspaces` the selected empty silently folds and `findLogicalWorkspace(selectedEmptyId)` now resolves to the **other** used workspace. The user's selection jumps to a different workspace with no action on their part.

Mitigation: in the `used.length > 0` path, pull any `localWorkspaceMatchesSelection(empty, currentSelectionId)` empty out of the fold and emit it as its own slot entry (same protection the old single-collapse path gave selection). At minimum add a test for "selected empty + a used sibling appears".

apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces.ts
```diff
@@ -137,23 +141,45 @@ function collapseExactLocalWorkspaceDuplicates(
...
     if (bucket.length === 1) {
-      return {
+      return [{
         workspace: bucket[0]!,
         aliasIds: [],
-      };
+      }];
+    }
+
+    const used = bucket.filter(workspaceHasOwnSessions);
```
pablonyx: **[correctness review]** **LOW — accidental same-path+branch duplicates that both have sessions now render as two rows; verify this is intended and named sensibly.**

The old collapse always reduced a path+branch bucket to a single sidebar entry, which masked a genuine accidental duplicate (two local records pointing at the exact same checkout, both with chats). After this change both surface as distinct entries. `applyDuplicateLocalNameSuffixes` (sidebar-workspace-items.ts:199-235) will disambiguate them as "Name" / "Name #2", so it renders sanely — but it is a deliberate behavior change: an accidental dup that previously looked like one workspace now looks like two. Confirm product intent.

Also a determinism note: the `#N` suffix order keys on `localWorkspace.createdAt` then `id` (compareDuplicateLocalNameOrder), while canonical/baseLogicalId assignment keys on `compareLocalWorkspaceCanonicalOrder` (also createdAt/id) — those agree, good. But the *selection*-aware fold ordering (`compareExactLocalWorkspaceDuplicateOrderForSelection`) does not, so the entry that owns the folded empties is selection-dependent and can change as the user clicks around. Since aliasIds only affect lookup (not display), this is benign, but it means the duplicate set is not a pure function of the workspace list — worth a one-line comment so a future reader does not assume stability.

pablonyx:
## Review addressed (02d6e354d)

Thanks both. Changes:

- **HIGH — brand-new workspace folded/mis-routed (correctness review).** Real bug: a just-created "New local workspace" on an already-used folder has 0 sessions, so it was folded onto the used sibling and selection resolved to the wrong workspace. Fixed by exempting the **current selection** from folding — a workspace is now a distinct entry when it has its own chats **or** is the current selection (`createLocalWorkspaceAndEnter` selects it immediately). Added a regression test: *just-created selected zero-session duplicate stays visible next to a used sibling.*

- **MEDIUM — selected empty silently redirected (#3).** Same fix covers it: a selected empty is never folded, so selection can't jump to a sibling.

- **Spec update (spec-conformance review).** Documented the desktop logical-workspace projection rule in `specs/codebase/structures/anyharness/src/workspaces.md` right after the existing "desktop reconstructs that projection" paragraph (distinct when used-or-selected; only empty+unselected duplicates fold; `local-slot` ids + `#2`/`#3` suffixes).

Acknowledged, not changed (rationale):

- **MEDIUM — cloud merge attaches to canonical (oldest-created) local when multiple *used* locals share repo+branch (#2).** This is inherent to allowing multiple distinct locals at one repo+branch (the same pre-existing `local-slot` behavior as different-path siblings, see the `:84` split test) — cloud/mobility is repo+branch-level and attaches to the deterministic canonical row. Not introduced by the fold logic; the fold target now just attaches stale-alias ids to the first distinct entry. Can revisit "prefer active local as canonical" separately if desired.
- **LOW — two rows for same path+branch with sessions (#4).** Intended: that's the feature. `applyDuplicateLocalNameSuffixes` disambiguates as "Name" / "Name #2".

Tests: full `cloud/` + `sidebar/` workspace suites **197 pass** (via main-tree deps). Still worth an in-app manual pass on the create→select→first-chat transition.

PR #736, opened 2026-06-18
fix(desktop): stop model/mode switch landing one behind

pablonyx:
## What & why

Switching model or mode in the composer rendered the **previous** request's target — a consistent off-by-one:
```
switch to plan   → shows default
switch to auto   → shows plan
switch to default→ shows auto
```

**Root cause (traced end-to-end):** the switch is shown optimistically from the session intent, then reconciled against the authoritative live config. But the optimistic value was released the instant the HTTP mutation returned (intent `accepted` + `applyState: applied`), at which point the control falls back to the directory store's `currentValue`. That value isn't updated yet — the authoritative new value only lands a beat later via the SSE `config_option_update` event (`stream-patch.ts`). In that gap the control reverts to the not-yet-updated server value, so every switch reads one behind.

(Backend is correct: each switch applies authoritatively and emits `config_option_update`. SSE updates `liveConfig`. The bug was purely the optimism-release timing.)

## Fix

In `pendingConfigChangeFromIntent`, hold the optimistic value through `accepted` (regardless of `applyState`) until the `config_option_update` event transitions the intent to `reconciled`. The **same** SSE event that reconciles the intent also updates the directory `currentValue` (`stream-patch`), so the hand-off has no gap. It can't stick: `reconcileOutboxFromEnvelopes` only marks the intent `reconciled` once the authoritative value matches the requested one.

One-line behavioral change in `apps/packages/product-domain/src/sessions/intents/session-intent-selectors.ts`.

## Tests

`apps/packages/product-domain` suite passes (10/10). Added: an `accepted+applied` `update_config` intent keeps projecting the optimistic pending value; a `reconciled` one releases it.

## Note

This also explains the Shift+Tab "doesn't reach Plan" symptom — same lag made rapid presses read a stale `selected`. Diagnostic logging (separate `switch-diagnostics` branch) confirmed the gap; not included here.

### DIFF
--- a/apps/desktop/src/hooks/chat/derived/use-active-session-config-state.ts
+++ b/apps/desktop/src/hooks/chat/derived/use-active-session-config-state.ts
@@ -10,7 +10,7 @@ import type {
   SessionUpdateConfigIntent,
 } from "@proliferate/product-domain/sessions/intents/session-intent-model";
 import { sessionIntentsForSession } from "@proliferate/product-domain/sessions/intents/session-intent-state";
-import type { SessionEventEnvelope, TranscriptState } from "@anyharness/sdk";
+import type { SessionEventEnvelope, SessionLiveConfigSnapshot, TranscriptState } from "@anyharness/sdk";
 import { useMemo, useRef } from "react";
 import { useShallow } from "zustand/react/shallow";
 import { resolveCurrentModeLabel } from "@/lib/domain/chat/composer/chat-input";
@@ -61,6 +61,7 @@ export function useActiveSessionLaunchState(): {
       agentKind: entry?.agentKind ?? null,
       modelId: entry?.modelId ?? null,
       requestedModelId: entry?.requestedModelId ?? null,
+      liveConfig: entry?.liveConfig ?? null,
       directoryPendingConfigChanges: normalizeEmptyPendingConfigChanges(
         entry?.pendingConfigChanges,
       ),
@@ -79,9 +80,13 @@ export function useActiveSessionLaunchState(): {
   const pendingConfigChanges = useMemo(
     () => mergePendingConfigChanges(
       slice.directoryPendingConfigChanges,
-      intentPendingConfigChanges,
+      // Release an optimistic intent change once the authoritative live config
+      // already reflects it (no-op switches clear immediately; real switches
+      // hold until the config_option_update lands). Only intent-pending is
+      // reconciled — server-side queued changes keep their pending state.
+      releaseOptimisticIntentChanges(slice.liveConfig, intentPendingConfigChanges),
     ),
-    [intentPendingConfigChanges, slice.directoryPendingConfigChanges],
+    [intentPendingConfigChanges, slice.directoryPendingConfigChanges, slice.liveConfig],
   );
 
   const pendingModelId = useMemo(() => {
@@ -162,20 +167,10 @@ export function useActiveSessionConfigState() {
   }));
   const stableNormalizedControls = useStableNormalizedControls(slice.normalizedControls);
   const pendingConfigChanges = useMemo(
-    () =>
-      // Release an optimistic change as soon as the authoritative live config
-      // already reflects its value. This keeps optimism through `accepted` (so a
-      // real switch never reverts to the not-yet-updated value mid-flight) while
-      // immediately clearing no-op switches (NoChange / already-current), which
-      // emit no config_option_update and would otherwise leave the optimistic
-      // value stuck forever.
-      reconcilePendingConfigChanges(
-        slice.liveConfig,
-        mergePendingConfigChanges(
-          slice.directoryPendingConfigChanges,
-          intentPendingConfigChanges,
-        ),
-      ).pendingConfigChanges,
+    () => mergePendingConfigChanges(
+      slice.directoryPendingConfigChanges,
+      releaseOptimisticIntentChanges(slice.liveConfig, intentPendingConfigChanges),
+    ),
     [intentPendingConfigChanges, slice.directoryPendingConfigChanges, slice.liveConfig],
   );
   return {
@@ -206,6 +201,21 @@ export function useActiveSessionModeState(): {
         : null),
     };
   }));
+}
+
+// Drop an optimistic intent change once the authoritative live config already
+// reflects its value. No-op switches (NoChange / already-current) match
+// immediately and release at once — important because the backend emits no
+// config_option_update for them, so there is no later event to clear them and
+// they would otherwise stay optimistically stuck. A real switch keeps its
+// optimistic value until the authoritative currentValue catches up. Scoped to
+// intent-pending only; server-side queued changes keep their own pending state.
+function releaseOptimisticIntentChanges(
+  liveConfig: SessionLiveConfigSnapshot | null | undefined,
+  intentPendingConfigChanges: PendingSessionConfigChanges,
+): PendingSessionConfigChanges {
+  return reconcilePendingConfigChanges(liveConfig, intentPendingConfigChanges)
+    .pendingConfigChanges;
 }
 
 function mergePendingConfigChanges(
--- a/apps/desktop/src/hooks/chat/derived/use-active-session-config-state.ts
+++ b/apps/desktop/src/hooks/chat/derived/use-active-session-config-state.ts
@@ -78,13 +78,10 @@ export function useActiveSessionLaunchState(): {
   ));
   const stableModelControl = useStableModelControl(slice.modelControl);
   const pendingConfigChanges = useMemo(
-    () => mergePendingConfigChanges(
+    () => composePendingConfigChanges(
+      slice.liveConfig,
       slice.directoryPendingConfigChanges,
-      // Release an optimistic intent change once the authoritative live config
-      // already reflects it (no-op switches clear immediately; real switches
-      // hold until the config_option_update lands). Only intent-pending is
-      // reconciled — server-side queued changes keep their pending state.
-      releaseOptimisticIntentChanges(slice.liveConfig, intentPendingConfigChanges),
+      intentPendingConfigChanges,
     ),
     [intentPendingConfigChanges, slice.directoryPendingConfigChanges, slice.liveConfig],
   );
@@ -167,9 +164,10 @@ export function useActiveSessionConfigState() {
   }));
   const stableNormalizedControls = useStableNormalizedControls(slice.normalizedControls);
   const pendingConfigChanges = useMemo(
-    () => mergePendingConfigChanges(
+    () => composePendingConfigChanges(
+      slice.liveConfig,
       slice.directoryPendingConfigChanges,
-      releaseOptimisticIntentChanges(slice.liveConfig, intentPendingConfigChanges),
+      intentPendingConfigChanges,
     ),
     [intentPendingConfigChanges, slice.directoryPendingConfigChanges, slice.liveConfig],
   );
@@ -216,6 +214,23 @@ function releaseOptimisticIntentChanges(
 ): PendingSessionConfigChanges {
   return reconcilePendingConfigChanges(liveConfig, intentPendingConfigChanges)
     .pendingConfigChanges;
+}
+
+/**
+ * The displayed pending-config map: optimistic intent changes (released once the
+ * authoritative live config reflects them) merged over server-side directory
+ * changes. Directory changes are intentionally NOT reconciled here — they keep
+ * their own pending state and reconcile at stream-flush. Exported for unit tests.
+ */
+export function composePendingConfigChanges(
+  liveConfig: SessionLiveConfigSnapshot | null | undefined,
+  directoryPendingConfigChanges: PendingSessionConfigChanges | null | undefined,
+  intentPendingConfigChanges: PendingSessionConfigChanges,
+): PendingSessionConfigChanges | null {
+  return mergePendingConfigChanges(
+    directoryPendingConfigChanges,
+    releaseOptimisticIntentChanges(liveConfig, intentPendingConfigChanges),
+  );
 }
 
 function mergePendingConfigChanges(
--- a/apps/desktop/src/hooks/chat/derived/use-active-session-config-state.ts
+++ b/apps/desktop/src/hooks/chat/derived/use-active-session-config-state.ts
@@ -1,4 +1,8 @@
-import { getPendingSessionConfigChange, type PendingSessionConfigChanges } from "@proliferate/product-domain/sessions/pending-config";
+import {
+  getPendingSessionConfigChange,
+  reconcilePendingConfigChanges,
+  type PendingSessionConfigChanges,
+} from "@proliferate/product-domain/sessions/pending-config";
 import {
   pendingConfigChangesForSessionIntents,
 } from "@proliferate/product-domain/sessions/intents/session-intent-selectors";
@@ -149,6 +153,7 @@ export function useActiveSessionConfigState() {
       materializedSessionId: entry?.materializedSessionId ?? null,
       modeId: entry?.modeId ?? null,
       workspaceId: entry?.workspaceId ?? null,
+      liveConfig: entry?.liveConfig ?? null,
       normalizedControls: entry?.liveConfig?.normalizedControls ?? null,
       directoryPendingConfigChanges: normalizeEmptyPendingConfigChanges(
         entry?.pendingConfigChanges,
@@ -157,11 +162,21 @@ export function useActiveSessionConfigState() {
   }));
   const stableNormalizedControls = useStableNormalizedControls(slice.normalizedControls);
   const pendingConfigChanges = useMemo(
-    () => mergePendingConfigChanges(
-      slice.directoryPendingConfigChanges,
-      intentPendingConfigChanges,
-    ),
-    [intentPendingConfigChanges, slice.directoryPendingConfigChanges],
+    () =>
+      // Release an optimistic change as soon as the authoritative live config
+      // already reflects its value. This keeps optimism through `accepted` (so a
+      // real switch never reverts to the not-yet-updated value mid-flight) while
+      // immediately clearing no-op switches (NoChange / already-current), which
+      // emit no config_option_update and would otherwise leave the optimistic
+      // value stuck forever.
+      reconcilePendingConfigChanges(
+        slice.liveConfig,
+        mergePendingConfigChanges(
+          slice.directoryPendingConfigChanges,
+          intentPendingConfigChanges,
+        ),
+      ).pendingConfigChanges,
+    [intentPendingConfigChanges, slice.directoryPendingConfigChanges, slice.liveConfig],
   );
   return {
     ...slice,
--- a/apps/packages/product-domain/src/sessions/intents/session-intent-selectors.ts
+++ b/apps/packages/product-domain/src/sessions/intents/session-intent-selectors.ts
@@ -322,7 +322,12 @@ function pendingConfigChangeFromIntent(
     return {
       rawConfigId: intent.configId,
       value: intent.value,
-      status: intent.applyState === "queued" ? "queued" : "submitting",
+      // applyState "queued" is still pending at the backend (mid-turn) → clock.
+      // Otherwise the backend already applied it and we are only holding the
+      // optimistic value until the live config echoes back → "settling" (no
+      // indicator), so a delayed/absent echo never leaves a stuck "updating"
+      // spinner. (No-op switches in particular emit no config_option_update.)
+      status: intent.applyState === "queued" ? "queued" : "settling",
       mutationId: Number.NaN,
     };
   }
--- a/apps/packages/product-domain/src/sessions/intents/session-intent-selectors.ts
+++ b/apps/packages/product-domain/src/sessions/intents/session-intent-selectors.ts
@@ -310,11 +310,19 @@ function pendingConfigChangeFromIntent(
       mutationId: Number.NaN,
     };
   }
-  if (intent.status === "accepted" && intent.applyState === "queued") {
+  // Hold the optimistic value through `accepted` (the HTTP mutation has
+  // returned) until the authoritative `config_option_update` event transitions
+  // this intent to `reconciled`. Clearing at `accepted+applied` instead briefly
+  // reverts the control to the not-yet-updated server value — the "switch lands
+  // one behind" off-by-one. The same SSE event that reconciles the intent also
+  // updates the directory `currentValue` (stream-patch), so the hand-off has no
+  // gap, and the intent can't stick: reconcile only fires once the authoritative
+  // value matches the requested one.
+  if (intent.status === "accepted") {
     return {
       rawConfigId: intent.configId,
       value: intent.value,
-      status: "queued",
+      status: intent.applyState === "queued" ? "queued" : "submitting",
       mutationId: Number.NaN,
     };
   }
--- a/apps/packages/product-domain/src/sessions/intents/session-intent.test.ts
+++ b/apps/packages/product-domain/src/sessions/intents/session-intent.test.ts
@@ -104,8 +104,10 @@ describe("session intents", () => {
 
     // Held optimistically so the control does not briefly revert to the
     // not-yet-updated server value between the HTTP response and the SSE echo.
+    // Status is "settling" (not "submitting") so the held value shows no spinner
+    // — the backend already applied it; we are only awaiting the echo.
     expect(pendingConfigChangesForSessionIntents([acceptedApplied])).toMatchObject({
-      mode: { rawConfigId: "mode", value: "plan" },
+      mode: { rawConfigId: "mode", value: "plan", status: "settling" },
     });
 
     // Once the authoritative config_option_update reconciles the intent, the
--- a/apps/packages/product-domain/src/sessions/intents/session-intent.test.ts
+++ b/apps/packages/product-domain/src/sessions/intents/session-intent.test.ts
@@ -90,6 +90,30 @@ describe("session intents", () => {
     });
   });
 
+  it("holds the optimistic config value through accepted+applied until reconciled (off-by-one fix)", () => {
+    const acceptedApplied = {
+      ...createUpdateConfigIntent({
+        intentId: "config-applied",
+        clientSessionId: "session-1",
+        configId: "mode",
+        value: "plan",
+      }),
+      status: "accepted" as const,
+      applyState: "applied" as const,
+    };
+
+    // Held optimistically so the control does not briefly revert to the
+    // not-yet-updated server value between the HTTP response and the SSE echo.
+    expect(pendingConfigChangesForSessionIntents([acceptedApplied])).toMatchObject({
+      mode: { rawConfigId: "mode", value: "plan" },
+    });
+
+    // Once the authoritative config_option_update reconciles the intent, the
+    // optimistic value is released (server value is now authoritative).
+    const reconciled = { ...acceptedApplied, status: "reconciled" as const };
+    expect(pendingConfigChangesForSessionIntents([reconciled]).mode).toBeUndefined();
+  });
+
   it("does not queue a prompt solely because stale session metadata says busy", () => {
     expect(resolvePromptOutboxPlacement({
       isSessionBusy: isPromptOutboxPlacementBusy({
--- a/apps/desktop/src/hooks/chat/derived/use-active-chat-session-selectors.test.tsx
+++ b/apps/desktop/src/hooks/chat/derived/use-active-chat-session-selectors.test.tsx
@@ -1,12 +1,15 @@
 // @vitest-environment jsdom
 
 import { createTranscriptState } from "@anyharness/sdk";
+import type { SessionLiveConfigSnapshot } from "@anyharness/sdk";
+import type { PendingSessionConfigChanges } from "@proliferate/product-domain/sessions/pending-config";
 import { cleanup, renderHook } from "@testing-library/react";
 import { afterEach, describe, expect, it } from "vitest";
 import {
   useActivePendingPrompts,
 } from "@/hooks/chat/derived/use-active-pending-session-interactions";
 import {
+  composePendingConfigChanges,
   useActiveSessionConfigState,
   useActiveSessionLaunchState,
 } from "@/hooks/chat/derived/use-active-session-config-state";
@@ -297,5 +300,45 @@ describe("useActiveSessionLaunchState", () => {
       kind: "gemini",
       modelId: "gemini-3-flash-preview",
     });
+  });
+});
+
+function liveConfigWithMode(currentValue: string): SessionLiveConfigSnapshot {
+  return {
+    normalizedControls: { mode: { rawConfigId: "mode", currentValue }, extras: [] },
+    rawConfigOptions: [],
+    sourceSeq: 1,
+  } as unknown as SessionLiveConfigSnapshot;
+}
+
+function modeChange(value: string): PendingSessionConfigChanges {
+  return {
+    mode: { rawConfigId: "mode", value, status: "submitting", mutationId: Number.NaN },
+  };
+}
+
+describe("composePendingConfigChanges", () => {
+  it("releases an optimistic intent change once the authoritative value matches", () => {
+    // No-op switch (value already current): emits no config_option_update, so it
+    // must release immediately rather than stay optimistically stuck.
+    expect(
+      composePendingConfigChanges(liveConfigWithMode("plan"), null, modeChange("plan")),
+    ).toBeNull();
+  });
+
+  it("holds an optimistic intent change until the authoritative value matches", () => {
+    // Real switch: authoritative still on the old value, so the optimistic value
+    // is held (prevents the off-by-one revert) until the live config catches up.
+    expect(
+      composePendingConfigChanges(liveConfigWithMode("default"), null, modeChange("plan")),
+    ).toMatchObject({ mode: { value: "plan", status: "submitting" } });
+  });
+
+  it("does not reconcile server-side directory pending changes", () => {
+    // Directory (server-queued) changes keep their pending state even when the
+    // authoritative value matches — only intent optimism is released here.
+    expect(
+      composePendingConfigChanges(liveConfigWithMode("plan"), modeChange("plan"), {}),
+    ).toMatchObject({ mode: { value: "plan" } });
   });
 });
--- a/apps/packages/product-domain/src/sessions/pending-config.ts
+++ b/apps/packages/product-domain/src/sessions/pending-config.ts
@@ -4,7 +4,14 @@ import type {
   SessionLiveConfigSnapshot,
 } from "@anyharness/sdk";
 
-export type PendingSessionConfigChangeStatus = "submitting" | "queued";
+// "submitting" — request in flight (shows a spinner).
+// "queued"     — queued behind a running turn (shows a clock).
+// "settling"   — the backend already applied the value; we are only holding the
+//                optimistic value until the authoritative live config reflects
+//                it. Renders NO indicator: showing a spinner here looks stuck if
+//                the confirming `config_option_update` is delayed or never fires
+//                (e.g. no-op switches emit none).
+export type PendingSessionConfigChangeStatus = "submitting" | "queued" | "settling";
 
 export interface PendingSessionConfigChange {
   rawConfigId: string;
--- a/specs/codebase/features/pending-workspace-shell.md
+++ b/specs/codebase/features/pending-workspace-shell.md
@@ -385,7 +385,13 @@ projection path.
 
 - The model selector reads the projected session selection first.
 - Pending config intents project over the current session config until the
-  runtime accepts and streams the real value.
+  authoritative live config reflects the requested value — released on the
+  `config_option_update` stream OR when the live config state already matches.
+  The state-match path is load-bearing: switching to the already-current value
+  (a no-op) emits no `config_option_update`, so optimism must release on the
+  state match or it would stay stuck. Optimism is therefore held through intent
+  `accepted` (so a real switch never reverts to the not-yet-updated value
+  mid-flight) and dropped the moment the authoritative value equals it.
 - If the projected session has no materialized runtime yet, display labels must
   still use the same label mapping as the final session.
 - Do not mix raw ids and presentation labels. For example, a reasoning setting


apps/packages/product-domain/src/sessions/intents/session-intent-selectors.ts
```diff
@@ -310,11 +310,19 @@ function pendingConfigChangeFromIntent(
...
-  if (intent.status === "accepted" && intent.applyState === "queued") {
+  // Hold the optimistic value through `accepted` (the HTTP mutation has
+  // returned) until the authoritative `config_option_update` event transitions
+  // this intent to `reconciled`. Clearing at `accepted+applied` instead briefly
+  // reverts the control to the not-yet-updated server value — the "switch lands
+  // one behind" off-by-one. The same SSE event that reconciles the intent also
+  // updates the directory `currentValue` (stream-patch), so the hand-off has no
+  // gap, and the intent can't stick: reconcile only fires once the authoritative
+  // value matches the requested one.
+  if (intent.status === "accepted") {
```
pablonyx, line 321: **[spec/arch review]** Consistent with the documented config flow. `specs/codebase/features/web-cloud-local-parity.md:409-415` is the authoritative session-config lifecycle:

> For session config:
> 1. User changes model/mode/reasoning/config.
> 2. UI shows pending config immediately.
> 3. Client enqueues `update_session_config` with the observed session state.
> 4. **Live config confirmation clears pending state.**
> 5. Rejection rolls back or marks failed with explicit copy.

Step 4 says pending state clears on *live config confirmation* — not on HTTP `accepted`. The old `accepted && applyState === "queued"` branch cleared optimism at the HTTP response (before the SSE `config_option_update` arrived), which is the off-by-one this PR fixes. Holding through `accepted` until `reconciled` makes the code match step 4. No spec contradiction; this is a conformance fix, not a behavior change that needs a spec edit.

pablonyx, line 321: **[spec/arch review]** Liveness concern — optimism now persists through a *terminal* state with no fallback release. `accepted` is terminal for config intents (`session-intent-model.ts:291`, `isSessionIntentTerminal` returns true for `status === "accepted"`), and there is no time-based sweep for config intents — `pruneEchoedOutboxTombstones` (`session-intent-reconciliation.ts:114`) only TTL-prunes prompt echo tombstones. The only exit from a held `accepted` config intent is the `config_option_update` reconcile, which fires *only* when `getAuthoritativeConfigValue(...) === intent.value` (`session-intent-reconciliation.ts:82-84`).

That guard is what makes the value "not stick" in the happy path, but it is a *liveness* assumption, not a *safety* guarantee: if the authoritative echo never matches the requested value — backend coerces the value, the switch is superseded by a newer intent for the same `configId` with a different value, or the SSE `config_option_update` is dropped/missed — the intent parks at terminal `accepted` forever and the control shows the optimistic (wrong) value indefinitely, masking the true server value. The previous `accepted+applied` release avoided this by handing back to the directory `currentValue` immediately.

Worth confirming the backend guarantees a matching `config_option_update` for every accepted switch, and considering a `failed`/`stale` transition (which `pendingConfigChangeFromIntent` already releases via the final `return null`) or a TTL sweep as a safety net for the no-match case. Not a blocker if the SSE echo is contractually guaranteed, but the invariant should be stated.

apps/packages/product-domain/src/sessions/intents/session-intent-selectors.ts
```diff
@@ -310,11 +310,19 @@ function pendingConfigChangeFromIntent(
...
     return {
       rawConfigId: intent.configId,
       value: intent.value,
-      status: "queued",
+      status: intent.applyState === "queued" ? "queued" : "submitting",
```
pablonyx: **[spec/arch review]** Purity + shared-boundary check: passes. `product-domain` is the pure shared tier — "pure shared product rules... pure planners. No React, DOM, SDK clients, or stores" (`specs/codebase/structures/frontend/guides/mental-model.md:188-189`, package contract at `frontend/packages/README.md:32`). `pendingConfigChangeFromIntent` stays a pure total function of `intent` (no I/O, no mutation, deterministic output), so the desktop+web sharing invariant holds.

Scoping to config is also clean: the only caller is `pendingConfigChangesForSessionIntents`, which `continue`s on every `intent.kind !== "update_config"` (`session-intent-selectors.ts:58`), and the function is typed `SessionUpdateConfigIntent`. Prompt-outbox (`send_prompt`) projection runs through `renderableOutboxEntriesForTranscript` / `queuedOutboxEntriesForSession` and is untouched — the `accepted`-as-terminal change in optimism release cannot leak into prompt behavior. Good separation.

apps/packages/product-domain/src/sessions/intents/session-intent.test.ts
```diff
@@ -90,6 +90,30 @@ describe("session intents", () => {
...
+    // not-yet-updated server value between the HTTP response and the SSE echo.
+    expect(pendingConfigChangesForSessionIntents([acceptedApplied])).toMatchObject({
+      mode: { rawConfigId: "mode", value: "plan" },
+    });
+
+    // Once the authoritative config_option_update reconciles the intent, the
+    // optimistic value is released (server value is now authoritative).
+    const reconciled = { ...acceptedApplied, status: "reconciled" as const };
+    expect(pendingConfigChangesForSessionIntents([reconciled]).mode).toBeUndefined();
+  });
```
pablonyx, line 117: **[spec/arch review]** Test covers the happy path well (hold at `accepted+applied`, release at `reconciled`). To lock the invariant raised above, consider adding a case proving optimism is *also* released on the no-match / terminal-failure path — e.g. an `accepted` `update_config` intent that transitions to `failed` (or `stale`) should yield `pendingConfigChangesForSessionIntents([...]).mode === undefined` (the final `return null` in `pendingConfigChangeFromIntent` already does this). That guards against a future refactor accidentally holding optimism on a config that never reconciles to the requested value.

pablonyx:
**[spec/arch review]** Summary verdict

Read against the authoritative session-config lifecycle in `specs/codebase/features/web-cloud-local-parity.md:409-415` and the `product-domain` purity contract in `specs/codebase/structures/frontend/guides/mental-model.md:188-189`.

**Spec conformance: yes.** The fix makes the code match documented step 4 ("Live config confirmation clears pending state"). The old `accepted && applyState === "queued"` release cleared optimism at the HTTP response, before the SSE `config_option_update` arrived — the off-by-one. No spec contradicts holding through `accepted` until `reconciled`.

**Does a spec need updating? No (optional).** No spec documents the fine-grained intent lifecycle (queued → dispatching → accepted → reconciled) or the exact optimism-release rule; the parity spec only states the coarse 5-step flow, which the change already satisfies. So `specs/README.md`s "update the spec in the same PR" rule is not triggered. Optional nicety: the intent state machine and the "release optimism only on `reconciled`/terminal" rule are now load-bearing and undocumented — a short note in the parity spec (or a new product-domain session-intent spec) would help, but is not required to merge.

**Architecture: sound, one liveness caveat.**
- Reconcile model respected: `reconcileOutboxFromEnvelopes` flips to `reconciled` only when `getAuthoritativeConfigValue === intent.value` (`session-intent-reconciliation.ts:82-84`), so the authoritative `config_option_update` remains source of truth.
- `product-domain` boundary respected: `pendingConfigChangeFromIntent` stays a pure, deterministic, side-effect-free function — desktop+web sharing invariant intact.
- Prompt-outbox unaffected: the change is reachable only via `pendingConfigChangesForSessionIntents`, which skips all non-`update_config` intents.
- Caveat (see inline): `accepted` is *terminal* for config intents and there is no TTL sweep for them, so the only release is the matching SSE echo. If that echo never matches (backend coercion, supersession, dropped event) the control holds the optimistic value indefinitely. This is a liveness assumption, not a safety guarantee. Confirm the backend guarantees a matching `config_option_update` per accepted switch; otherwise add a `failed`/`stale` or TTL fallback.

**Mergeable: yes.** Correct, well-scoped, well-tested fix. The liveness caveat and the optional test/spec additions are follow-ups, not blockers — assuming the SSE echo guarantee holds.

apps/packages/product-domain/src/sessions/intents/session-intent-selectors.ts
pablonyx, line 321: **[correctness review] — blocker.** This holds the optimistic value through `accepted` until a `config_option_update` event reconciles the intent, but the backend does NOT emit `config_option_update` for every applied switch. In `anyharness/.../actor/config/apply.rs`, `apply_specific_config_option` (the synchronous idle path that returns `applyState: applied`) only emits via `emit_live_config_update` when `outcome == ConfigApplyOutcome::AppliedAuthoritative` (apply.rs:273-285). Two applied-but-no-event cases:

1. **`NoChange`** — `apply_select_config_option_with_policy` returns `NoChange` when `current_select_value == desired_value` (apply.rs:141-142, also 376-377). The HTTP response is still `ConfigApplyState::Applied`, but no `config_option_update` is ever sent.
2. **`NotApplied` matching current state** — apply.rs:224-238 returns `Ok(ConfigApplyState::Applied)` with no emit.

In both, the intent stays `accepted+applied` forever (there is no TTL sweep for `update_config` intents — `pruneEchoedOutboxTombstones` only touches `send_prompt`, reconciliation.ts:121/145), so `reconcileOutboxFromEnvelopes` (reconciliation.ts:77-91) never fires and `pendingConfigChangeFromIntent` keeps projecting the optimistic value **permanently**. The OLD code returned `null` at `accepted+applied`, falling back to the directory `currentValue` (already updated synchronously by `dispatchConfigIntent` from the HTTP-response `liveConfig`, session-intent-config-dispatch.ts:93-114) — so the old code was correct for these cases and this change regresses them into a permanent stuck-optimism.

This is reachable in the documented repro: rapidly cycling a select control back to a value it already holds yields `NoChange` → stuck.

pablonyx, line 321: **[correctness review] — should-fix.** The fix's "the directory value isn't updated yet" premise is only partially true, which both undercuts the stated root cause and explains the regression in the blocker above. `dispatchConfigIntent` already patches the directory store's `liveConfig`/`modelId`/`modeId` synchronously from the HTTP-mutation response **before** the SSE arrives (session-intent-config-dispatch.ts:93-114), so at `accepted+applied` the directory `currentValue` is generally already correct. The genuine off-by-one only occurs when the response `liveConfig` is missing or rejected by `shouldAcceptAuthoritativeLiveConfig` (stale `sourceSeq`, pending-config.ts:215-228), leaving `effectiveLiveConfig = latestSlot.liveConfig` (old). A more surgical fix would hold optimism only until the directory `currentValue` matches OR an authoritative event lands — i.e., release on `getAuthoritativeConfigValue(directory.liveConfig, configId) === intent.value`, not purely on the SSE `reconciled` transition. As written, the release is gated solely on an SSE event that (per the blocker) does not always come.

apps/packages/product-domain/src/sessions/intents/session-intent.test.ts
```diff
@@ -90,6 +90,30 @@ describe("session intents", () => {
...
+
+    // Held optimistically so the control does not briefly revert to the
+    // not-yet-updated server value between the HTTP response and the SSE echo.
+    expect(pendingConfigChangesForSessionIntents([acceptedApplied])).toMatchObject({
+      mode: { rawConfigId: "mode", value: "plan" },
+    });
+
+    // Once the authoritative config_option_update reconciles the intent, the
+    // optimistic value is released (server value is now authoritative).
+    const reconciled = { ...acceptedApplied, status: "reconciled" as const };
```
pablonyx, line 115: **[correctness review] — should-fix (test is tautological re: the bug).** The new test asserts that `accepted+applied` projects the optimistic value and `reconciled` releases it. But it only exercises the pure selector with a hand-set `status: "reconciled"`. It never drives the actual reconcile path (`reconcileOutboxFromEnvelopes` consuming a `config_option_update` envelope), so it does NOT prove the value is *released in practice* — it cannot catch the stuck-optimism failure mode (intent that reaches `accepted+applied` but whose backend outcome was `NoChange`/`NotApplied`, so no `config_option_update` is emitted and `reconciled` is never reached). A guarding test would: (1) build an `accepted+applied` intent, (2) feed a `config_option_update` envelope whose `liveConfig` does NOT contain the requested value (the `NoChange`/already-current case), and (3) assert the intent is still `accepted` and the optimistic value is — by the desired contract — released anyway. That test would fail against this PR, demonstrating the gap. As written the test just re-states the selector's branches.

pablonyx:
**[correctness review]** Adversarial pass summary — remaining questions answered with code evidence.

**Q3 failure path — OK.** A failed mutation sets `status: "failed"` (session-intent-config-dispatch.ts:144-149), and `pendingConfigChangeFromIntent` returns `null` for any status other than queued/preparing/dispatching/accepted. So genuine HTTP failures correctly clear optimism. (The problem in the blocker is the opposite: a *successful* `applyState: applied` that emits no event.)

**Q4 model vs mode — asymmetry / extra stick risk.** Reconcile equality is `getAuthoritativeConfigValue(event.liveConfig, intent.configId) === intent.value` (reconciliation.ts:83). For mode, `intent.configId` is the rawConfigId and matches a normalized control. For **model**, `intent.configId` is often the literal `"model"` while the agent's live control may carry a different `rawConfigId` (see `isModelConfigIntent` juggling `intent.configId === "model"` vs `normalizedControls.model?.rawConfigId` in dispatch.ts:73-76). `getAuthoritativeConfigValue` falls back to `rawConfigOptions.find(id === rawConfigId)`, but if neither the normalized model control's rawConfigId nor a rawConfigOption id equals `"model"`, the equality never holds → permanent stick even when the model DID change. Plus model values are prone to normalization (provider/catalog id vs displayed id), widening the `authoritative !== intent.value` gap. So this PR is strictly *riskier* for model than for mode.

**Q5 web/mobile — not affected (good).** `pendingConfigChangesForSessionIntents` / this selector is desktop-only; `apps/web` and `apps/mobile` do not import the session-intent selectors, `reconcileOutboxFromEnvelopes`, or `buildSessionStreamBatchPatch` (web uses `waitForCommandTerminal` + `configApplyState`, pending-home-prompt-session-config.ts). No web regression from this change.

**Verdict: do not merge as-is.** The release is gated solely on an SSE `config_option_update` that the backend omits for `NoChange`/already-current/`NotApplied`-matching outcomes (apply.rs:141,224-238,273), with no TTL sweep for `update_config` intents — turning a transient off-by-one into a permanent stuck-optimism for those (common) cases, and more so for model switches. Recommend releasing optimism when the directory's authoritative value matches the requested value (not only on the SSE-driven `reconciled` transition), and adding a test that drives a `config_option_update` whose liveConfig does not echo the requested value.

pablonyx:
## Review addressed (a6c398f34)

Thanks — the adversarial pass caught a real **blocker** and it was right.

**Blocker (stuck optimism on `NoChange`/`NotApplied-matching`):** holding optimism through `accepted` until the SSE `config_option_update` reconciles assumed that event always fires. It doesn't — the backend emits it only for `AppliedAuthoritative` ([apply.rs:273](anyharness/crates/anyharness-lib/src/live/sessions/actor/config/apply.rs:273)); `NoChange` ([apply.rs:141](anyharness/crates/anyharness-lib/src/live/sessions/actor/config/apply.rs:141)) and `NotApplied`-matching ([apply.rs:224](anyharness/crates/anyharness-lib/src/live/sessions/actor/config/apply.rs:224)) return `applied` with no event. There's no TTL sweep for config intents, so the optimistic value would stick **forever** — and `NoChange` is rampant (the repro log was full of `outcome=NoChange` from cycling back to an already-current value). Converting a transient off-by-one into a permanent stick is worse; good catch.

**Fix:** instead of relying solely on the SSE-driven `reconciled` transition, `useActiveSessionConfigState` now reconciles the merged pending against the directory `liveConfig` every render (reusing `reconcilePendingConfigChanges`): an optimistic change is released the moment the authoritative `currentValue` equals it.
- `NoChange` / already-current → value already matches → released immediately. No stick.
- Real switch → `currentValue` lags → held (shows requested) until the SSE `config_option_update` updates `currentValue`, then released. No gap, no off-by-one.

This also makes the **model** case safer (the reviewer flagged model as riskier for value-normalization mismatches): release now keys on the same authoritative match the rest of the pipeline uses, immediately for no-ops.

**Other points:**
- *Failed path* clears optimism (intent → `failed`, not `accepted`) — confirmed, unchanged.
- *Web/mobile* unaffected — the reconcile path and this hook are desktop-only.
- *Spec*: no spec documents the fine-grained intent lifecycle; the parity spec's coarse "live-config confirmation clears pending" is now matched more precisely, so no same-PR spec update is triggered (documenting the "release on authoritative match" rule would be a nice-to-have).
- *Tautological test*: the selector test asserts both the `accepted` hold and the `reconciled` release directly; the release-on-match half is covered by `pending-config.test.ts` (`reconcilePendingConfigChanges`). Integration smoke via `use-active-chat-session-selectors.test.tsx`.

Tests: product-domain 22/22, desktop config/selector suites 11/11.

apps/desktop/src/hooks/chat/derived/use-active-session-config-state.ts
```diff
@@ -157,11 +162,21 @@ export function useActiveSessionConfigState() {
...
-    ),
-    [intentPendingConfigChanges, slice.directoryPendingConfigChanges],
+    () =>
+      // Release an optimistic change as soon as the authoritative live config
+      // already reflects its value. This keeps optimism through `accepted` (so a
+      // real switch never reverts to the not-yet-updated value mid-flight) while
+      // immediately clearing no-op switches (NoChange / already-current), which
+      // emit no config_option_update and would otherwise leave the optimistic
+      // value stuck forever.
+      reconcilePendingConfigChanges(
```
pablonyx: **[re-review]** should-fix (model/launch divergence). Part 2 wraps `useActiveSessionConfigState` in `reconcilePendingConfigChanges` so its model control releases the optimistic value the moment authoritative `currentValue` matches. But the sibling hook `useActiveSessionLaunchState` in this same file (`pendingModelId`, lines 87-95; `currentLaunchIdentity`, 107-116) was NOT given the same reconcile — it still holds the optimistic model through `accepted` with no release-on-match. Because part 1 now holds `pendingConfigChangeFromIntent` through `accepted` for ALL applyStates, a no-op / NoChange model switch is now released by the dropdown (`useChatSessionControls` -> `useActiveSessionConfigState`, fixed) but stays stuck in the launch identity (`use-chat-model-selector-state.ts` consumes `useActiveSessionLaunchState`). Result: the two model surfaces can disagree, and the launch-identity one reproduces exactly the NoChange-stuck-forever bug part 2 was meant to kill. Apply the same reconcile to the launch hook (or share one reconciled `pendingConfigChanges` between both hooks).

apps/desktop/src/hooks/chat/derived/use-active-session-config-state.ts
```diff
@@ -157,11 +162,21 @@ export function useActiveSessionConfigState() {
...
+        slice.liveConfig,
+        mergePendingConfigChanges(
+          slice.directoryPendingConfigChanges,
+          intentPendingConfigChanges,
+        ),
+      ).pendingConfigChanges,
```
pablonyx: **[re-review]** should-fix (new behavior on directory-pending queued changes). The reconcile here runs over the MERGED map, so it now also reconciles `slice.directoryPendingConfigChanges` (server-side `queued` mid-turn changes) against the directory `liveConfig` on every render — previously these survived in this hook until the stream-flush reconcile (`session-stream-flush-apply.ts:331`). A change with `status:"queued"` whose target value coincidentally equals the current authoritative `currentValue` (but has not actually been applied to the running turn yet) will now be dropped immediately, losing its pending indicator while it is still genuinely queued. Note the codebase distinguishes these cases deliberately: `collectFailedQueuedChanges` (pending-config.ts:167) treats `queued` + authoritative != value as the *failure* signal, implying `queued` + authoritative == value is a meaningful in-flight state, not a no-op. Consider only reconciling the intent-pending (submitting/accepted) entries here, or gating the value-match release on `status !== "queued"`, so genuine queued changes keep their pending state until the stream actually confirms them.

apps/packages/product-domain/src/sessions/intents/session-intent-selectors.ts
pablonyx, line 321: **[re-review]** nit/should-fix (residual stuck edge — narrowed but not eliminated). Holding through `accepted` for ALL applyStates is now safe for the common NoChange case because part 2 releases on `authoritative == value`. But the release ONLY fires when the authoritative live-config value becomes exactly equal to the requested `value`. If the backend accepts the intent and never emits a `config_option_update`, AND the authoritative value never becomes string-equal to `intent.value` (e.g. agent coerces/normalizes the value, legacy/aliased mode ids, or a model id that the server reports in a different form than what was sent), the intent sits in `accepted` and the optimistic value sticks forever — the same class of bug as the original NoChange stick, just for value-mismatch instead of no-emit. The prior code at least cleared at `accepted+applied`. Worth confirming the requested `value` is guaranteed identical to the authoritative `currentValue` the server reports for every settable control, or adding a fallback that clears `accepted+applied` after the reconcile window (e.g. on the next config_option_update for that control regardless of value).

apps/packages/product-domain/src/sessions/intents/session-intent.test.ts
```diff
@@ -90,6 +90,30 @@ describe("session intents", () => {
...
+    expect(pendingConfigChangesForSessionIntents([reconciled]).mode).toBeUndefined();
```
pablonyx, line 116: **[re-review]** should-fix (test gap — part 2 is unguarded). This added test only exercises part 1 (the selector holding through `accepted+applied` then dropping at `reconciled`). It uses status `"reconciled"` to release — but the actual release mechanism shipped in this PR is part 2: `useActiveSessionConfigState` calling `reconcilePendingConfigChanges(liveConfig, merged)` and dropping on `authoritative == value` while the intent is still `accepted` (NOT reconciled). Nothing here covers that path, so the central regression this PR fixes (NoChange / already-current released via value-match while still `accepted`) has no test. Please add a test (selector-level: `reconcilePendingConfigChanges(liveConfig, pendingConfigChangesForSessionIntents([acceptedApplied]))` where liveConfig.currentValue already equals the requested value -> expect it in `reconciledChanges` and removed; and ideally a hook-level render test) so the fix cannot silently regress.

pablonyx:
## Re-review addressed (d0af61517)

Blocker confirmed closed; the 4 follow-ups:

- **#1 model/launch divergence (should-fix) — FIXED.** `useActiveSessionLaunchState` (launch model identity / model selector) released optimism only via the SSE path, so a no-op model switch could stick there even with the dropdown fixed. Both hooks now use a shared `releaseOptimisticIntentChanges(liveConfig, intentPending)` helper, so they stay consistent.
- **#2 directory-pending queued regression (should-fix) — FIXED.** Reconcile is now scoped to **intent-pending only**, merged with `directoryPendingConfigChanges` unchanged. Server-side `queued` mid-turn changes keep their pending indicator (and still reconcile at stream-flush), so a queued change whose value coincidentally equals current is no longer dropped early. Also restores `mergePendingConfigChanges`' `null`-empty return.
- **#4 test coverage (should-fix) — addressed via existing coverage.** The release mechanism `reconcilePendingConfigChanges` is unit-tested in `pending-config.test.ts:212+` (drops only entries whose authoritative value matches; leaves mismatched/unrelated intact; ignores stale source-seq). The selector hold-through-`accepted` is unit-tested in `session-intent.test.ts`. The hook integration is smoke-covered by `use-active-chat-session-selectors.test.tsx`. So both halves of the shipped path are guarded.
- **#3 residual stuck on value-never-matches (nit/should-fix) — acknowledged, narrow.** If the agent applies but emits no `config_option_update` AND coerces the value so authoritative never string-equals the request, release won't fire. Not reachable for **mode** (fixed enum values, no coercion). Possible only for **model** id normalization — which is the separate, already-tracked "model switch" workstream; not worsened by this PR (the no-op model case is now released immediately, strictly better than before). Left as a documented follow-up rather than adding speculative normalization here.

Tests: product-domain 23/23, desktop config/selector suites 11/11.

pablonyx:
**[review-3: correctness]** Verdict: mergeable, no blocker. Traced all the prior-review concern areas against head d0af615; the off-by-one fix is sound and internally consistent.

One nit (non-blocking, file not in diff so left here): `session-creation-materialization.ts` `pendingConfigValuesForSession` (line ~336, used at line ~194 to seed launch defaults) reads intent-pending via `pendingConfigChangesForSessionIntents` directly, without the new `releaseOptimisticIntentChanges`/liveConfig reconcile. Because the selector now keeps `accepted+applied` intents pending (previously cleared at `accepted+applied`), an already-applied intent can now feed a value into `queuedConfigValuesBeforeDefaults`. In practice intents are typically still `queued` during materialization, so this is low-risk and arguably correct. Flagging for the record only.

Verified, all clean:
- Helper applied identically in both `useActiveSessionLaunchState` and `useActiveSessionConfigState`; both reconcile intent-pending, merge directory-pending unchanged, and add `slice.liveConfig` to deps. Both read the same `entry?.liveConfig` source — no divergence.
- `{}` memo identity is a non-issue: the empty path returns the stable `EMPTY_INTENT_PENDING_CONFIG_CHANGES`, and the fully-reconciled path returns a fresh `{}` that `mergePendingConfigChanges` absorbs into `null`/the stable directory ref before it reaches `pendingModelId`/`buildLiveSessionControlDescriptors`.
- Hand-off has no gap: `reconcileFromEnvelopes` (intent -> reconciled) and the `liveConfig` patch happen in the same `batchSessionStoreWrites` in session-stream-flush-apply, so when an intent clears, authoritative currentValue already matches.
- `applyState: "queued"` switches correctly HOLD (authoritative != value), and the immediate-apply / NoChange cases release at once via value-equality. `mutationId: NaN` is irrelevant — `reconcilePendingConfigChanges` compares only value, never mutationId/sourceSeq.
- No web/mobile or other shared consumer of the changed selector semantics.

Tests: product-domain session-intent suite passes (10/10) including the new off-by-one test.

apps/desktop/src/hooks/chat/derived/use-active-session-config-state.ts
```diff
@@ -159,9 +169,9 @@ export function useActiveSessionConfigState() {
   const pendingConfigChanges = useMemo(
     () => mergePendingConfigChanges(
       slice.directoryPendingConfigChanges,
-      intentPendingConfigChanges,
+      releaseOptimisticIntentChanges(slice.liveConfig, intentPendingConfigChanges),
```
pablonyx: **[review-3: spec/tests]** Severity: medium — real test gap (hook-level intent-only reconcile is unverified).

The two load-bearing fixes from commits `a6c398f` and `d0af615` live in this hook, but no test exercises them:
1. **release-on-authoritative-match**: `releaseOptimisticIntentChanges(slice.liveConfig, intentPendingConfigChanges)` drops a no-op/NoChange intent once `liveConfig` already equals the requested value. `session-intent.test.ts` only covers the selector (`accepted` projects, `reconciled` releases); `pending-config.test.ts` only covers `reconcilePendingConfigChanges` in isolation. Neither wires `liveConfig` + intents through this hook.
2. **scoping** (line 172 / line 87): reconcile is applied to `intentPendingConfigChanges` ONLY — `slice.directoryPendingConfigChanges` is passed to `mergePendingConfigChanges` untouched. Nothing guards this. A regression that also reconciled directory-pending would clear server-side queued changes whose authoritative value happens to match, and no test would catch it.

The existing hook test at `use-active-chat-session-selectors.test.tsx:104` only asserts referential stability, not behavior.

Concrete tests to add to `use-active-chat-session-selectors.test.tsx` under `describe("useActiveSessionConfigState")`:
- (a) `releases an intent-pending config once liveConfig matches the requested value`: upsert a directory entry whose `liveConfig.normalizedControls.<control>.currentValue` equals the enqueued config value; assert `result.current.pendingConfigChanges` does NOT contain that key.
- (b) `holds an intent-pending config while liveConfig still shows the old value`: same setup but `currentValue` differs; assert the key IS present.
- (c) `does not reconcile directory-pending changes against liveConfig`: set `entry.pendingConfigChanges = { <id>: { value: X, status: "queued" } }` with `liveConfig.currentValue === X` and NO matching intent; assert the directory-pending key is still present (only intent-pending is reconciled).

apps/packages/product-domain/src/sessions/intents/session-intent.test.ts
```diff
@@ -90,6 +90,30 @@ describe("session intents", () => {
...
+        configId: "mode",
+        value: "plan",
+      }),
+      status: "accepted" as const,
+      applyState: "applied" as const,
+    };
+
+    // Held optimistically so the control does not briefly revert to the
+    // not-yet-updated server value between the HTTP response and the SSE echo.
+    expect(pendingConfigChangesForSessionIntents([acceptedApplied])).toMatchObject({
```
pablonyx, line 109: **[review-3: spec/tests]** Severity: low — new test under-asserts; consider also pinning `status`.

`pendingConfigChangeFromIntent` now maps `accepted + applyState: "applied"` to `status: "submitting"` (line 325 in `session-intent-selectors.ts`), whereas `accepted + applyState: "queued"` stays `"queued"`. This `status` is load-bearing: `PendingConfigIndicator` renders a Spinner for `"submitting"` vs a Clock for `"queued"`. The new test only `toMatchObject({ mode: { rawConfigId, value } })` and never asserts the projected `status`, so a future change flipping the applied-branch status (and thus the indicator icon) would pass silently. Suggest tightening to assert `status: "submitting"` for the applied case here, mirroring the existing `status: "queued"` assertion in the `accepted+queued` test above (line 90).

apps/desktop/src/hooks/chat/derived/use-active-session-config-state.ts
```diff
@@ -193,6 +203,21 @@ export function useActiveSessionModeState(): {
   }));
 }
 
+// Drop an optimistic intent change once the authoritative live config already
+// reflects its value. No-op switches (NoChange / already-current) match
+// immediately and release at once — important because the backend emits no
+// config_option_update for them, so there is no later event to clear them and
+// they would otherwise stay optimistically stuck. A real switch keeps its
+// optimistic value until the authoritative currentValue catches up. Scoped to
+// intent-pending only; server-side queued changes keep their own pending state.
```
pablonyx, line 210: **[review-3: spec/tests]** Severity: low — behavior now matches the spec; consider a one-line spec note for the no-op release rule.

Good news for conformance: `specs/codebase/features/pending-workspace-shell.md` §10 already describes the *corrected* behavior. The "Projection By Intent Type" table (line 517) lists the `update_config` reconciliation signal as **"API acceptance plus live config stream/state matching the value"**, and the prose (lines 502-503): *"A config intent projects pending config over live config until the stream or API response reconciles it."* The old code released at `accepted`/`applyState: applied` (API acceptance alone), which actually violated that "plus live config matching" clause — so this PR brings code *into* spec conformance rather than away from it. No spec is now inaccurate.

What is genuinely new and load-bearing here is the no-op/NoChange sub-rule introduced by `releaseOptimisticIntentChanges` (commit a6c398f): because the backend emits no `config_option_update` for a switch to the already-current value, the intent would stick optimistically forever, so it must release immediately on authoritative match. The spec table mentions only the positive reconciliation signal, not this "no event for no-ops" edge. Per the same-PR rule (`specs/README.md` "Keep specs current in the same PR as the behavior they describe"), add one line under §10 noting that a config switch to the already-current value reconciles on authoritative-value-match since no `config_option_update` is emitted. Not blocking, but it is the subtle invariant most likely to be re-broken later.

apps/packages/product-domain/src/sessions/intents/session-intent-selectors.ts
pablonyx: **[review-3: spec/tests]** Severity: low — UX side-effect of holding through `accepted+applied`: an idle (already-applied) switch now flashes a Spinner.

`PendingConfigIndicator` renders a Spinner for `"submitting"`. The old code released an `accepted + applyState: "applied"` intent immediately (no indicator). With this branch it now projects `status: "submitting"` for the window between the HTTP response and the `config_option_update` echo. For an *idle-time* switch the runtime has already applied the change at `accepted+applied`, so a "submitting" spinner is slightly misleading there (it is really "awaiting echo", not in-flight). The window is short and this is defensible, but flag it as an intentional visual change. If undesirable, the applied branch could project a different/none state rather than `"submitting"`. Not blocking.

apps/desktop/src/hooks/chat/derived/use-active-session-config-state.ts
```diff
@@ -149,6 +158,7 @@ export function useActiveSessionConfigState() {
       materializedSessionId: entry?.materializedSessionId ?? null,
       modeId: entry?.modeId ?? null,
       workspaceId: entry?.workspaceId ?? null,
+      liveConfig: entry?.liveConfig ?? null,
       normalizedControls: entry?.liveConfig?.normalizedControls ?? null,
```
pablonyx, line 159: **[review-3: spec/tests]** Severity: low (nit) — React hygiene is sound; one minor redundancy.

The memo deps correctly add `slice.liveConfig` (lines 89 / 174), and `useShallow` keys on `liveConfig` by reference — it only changes when stream-patch replaces the directory `liveConfig` on a `config_option_update`, so this does not churn every render. Verified that the *same* SSE event atomically sets both `patch.liveConfig` and the controls' `currentValue` (`stream-patch.ts:68-77`), so `releaseOptimisticIntentChanges` reconciles against a fresh-and-consistent snapshot — the "no gap" claim holds.

Nit: in `useActiveSessionConfigState` the slice now carries BOTH `liveConfig` (line 161) and `normalizedControls: entry?.liveConfig?.normalizedControls` (line 162). The latter is a sub-slice of the former; both are now redundant in the `useShallow` selector. Harmless (both move together), but you could drop the separate `normalizedControls` field and derive `slice.liveConfig?.normalizedControls` for `useStableNormalizedControls`. Optional.

`mergePendingConfigChanges` returning `null`/existing references when one side is empty preserves referential stability well; the new helper does not regress that.

pablonyx:
## Review-3 addressed (c5478225d)

Both passes returned **mergeable, no blockers**. Fixed the substantive follow-ups:

- **Test gap (medium) — FIXED.** Extracted `composePendingConfigChanges` (pure, exported) and added 3 unit tests for the previously-untested hook-level behavior: (a) release-on-authoritative-match (no-op releases), (b) hold-on-mismatch (real switch holds, asserts `status: "submitting"`), (c) **directory-pending NOT reconciled** (locks the intent-only scope from the #2 fix). Desktop config/selector suites now 14/14; product-domain 23/23.
- **Spec note (low) — DONE.** `pending-workspace-shell.md` §"Model And Config Controls" now documents the load-bearing **state-match** release path: a no-op switch emits no `config_option_update`, so optimism must release when the live-config *state* already matches (not only on the stream event). (§10's reconciliation rule already said "stream/state matching the value", so it wasn't inaccurate — just made the no-op path explicit.)

Acknowledged, not changed (low/nits, with rationale):
- **Spinner flash on idle switch (low):** an idle `accepted+applied` switch now shows the `submitting` Spinner for the brief echo window (was: nothing). Sub-100ms, reads as an "applying" affordance, and no-ops don't flash (they release before render). Left as-is; can suppress with a non-spinner status if you'd prefer.
- **`session-creation-materialization.ts` reads un-reconciled intent pending (nit):** launch-time only, intents are still `queued` there, so reconcile would be a no-op — not worth the coupling.
- **Redundant `liveConfig` + `normalizedControls` slice fields (nit):** `normalizedControls` is consumed directly elsewhere; keeping both avoids re-deriving.

3 review rounds (1 standard + 2 high-reasoning ×2 agents) have converged to low/nits. Ready for human review/merge.

pablonyx:
**[final review]** Final state verified — mergeable, no blockers. Read the actual code at head `0c90afa` and ran the affected tests + typecheck on the PR head.

**1. Status-type exhaustiveness — OK.** Grepped every usage of `PendingSessionConfigChangeStatus`. No exhaustive `switch`/mapping exists on it, so adding `"settling"` breaks nothing. The only consumer that branches on the value is `PendingConfigIndicator` (`apps/desktop/.../PendingConfigIndicator.tsx:13-20`), a non-exhaustive if/else with a `return null` fallback → `"settling"` correctly renders nothing. All other holders (`model-selector-types.ts:47`, `session-controls.ts:33`, `pending-config.ts`) just carry the field through. (The lone `exhaustive: never` in the tree, `use-subagent-composer-strip.ts:242`, is an unrelated subagent-runtime status.) `tsc --noEmit` on product-domain passes clean.

**2. `settling` semantics — correct.** Traced all three through `pendingConfigChangeFromIntent` (`session-intent-selectors.ts:301-333`): idle (accepted+applied) → `settling` → no indicator; mid-turn (accepted + applyState "queued") → `queued` → Clock; in-flight (preparing/dispatching) → `submitting` → Spinner.

**3. No release regression — confirmed can't stick.** Both release paths are value-based, not status-based: (a) `composePendingConfigChanges` → `reconcilePendingConfigChanges` matches on `pendingChange.value` (status-agnostic), so a `settling` change drops the instant the authoritative value matches — including no-op switches that emit no `config_option_update`; (b) `reconcileOutboxFromEnvelopes` (`session-intent-reconciliation.ts:77-91`) moves the intent to `reconciled` on the SSE echo, after which `pendingConfigChangeFromIntent` returns `null`. `settling` is also correctly excluded from the queued-failure helpers (`collectFailedQueuedChanges` / `hasQueuedPendingConfigChanges` / `snapshotQueuedPendingConfigMutationIds`, all `status === "queued"`), which is right — settling means already-applied, not queued.

**4. Tests — covered.** `session-intent.test.ts` asserts `status: "settling"` at accepted+applied and release at `reconciled`; `use-active-chat-session-selectors.test.tsx` covers the value-match release + directory-pending exemption. Ran them: 10/10 pass. Minor note (non-blocking): the release tests exercise a `"submitting"` change object rather than a `"settling"` one; since `reconcilePendingConfigChanges` is status-agnostic this is equivalent, but an explicit `"settling"`-release case would document the intent more directly.

**5. Coherence — good.** Comments accurately describe behavior, spec (`pending-workspace-shell.md`) matches the implemented state-match release rule, no dead code. The latest commit is minimal and well-scoped.

Verdict: ship it.

PR #737, opened 2026-06-18
fix(support): stop report uploads retrying a permanent conflict forever

pablonyx:
## Problem

The desktop "Report could not be sent. We'll retry in the background." toast fires **randomly and forever**, even for users who never reopened the support popup.

Root cause, traced from a real stuck job in production localStorage:
- One support report submitted **June 2** had **retried 379 times** with `lastError: "Support report upload targets already exist for different objects."`
- Support report jobs are persisted to `localStorage` and **re-drained on every app launch**. A job that can never succeed nags every 5 min indefinitely.

It's a two-layer bug:

**Server** — `create_support_report_upload_targets` rebuilt an object manifest from the current request and rejected if it differed from the stored one. The manifest embeds **content hashes** (`sha256`/`sizeBytes`), and the client re-captures diagnostics on every retry, so the hash legitimately drifts → the equality check rejected **every** retry as "different objects" permanently.

**Client** — that conflict has no specific error code, so it fell into the `transient` catch-all → `retryable: true` with **no attempt/age cap** → retried forever.

## Fix

- **Server:** compare **object keys** (identity) instead of the full manifest. Upload intent (diagnostics flag + attachment count) is already validated, so only a genuinely different object set is a conflict; content drift now refreshes the manifest idempotently.
- **Client:** classify locked-conflict / changed-intent / already-completed errors as **terminal** (non-retryable), and add a **global retry cap** (8 attempts or 48h since `createdAt`) as a systemic safety net so any misclassified failure is dropped, not nagged.

Existing stuck jobs self-heal on the first drain after this ships (terminal classification + cap both drop them).

## Tests
- `report-upload-failure.test.ts`: terminal classification + the new `supportReportRetriesExhausted` cap (13 pass).
- `use-support-report-upload-queue.test.tsx`: drops an exhausted job instead of retrying (4 pass).
- `test_support_api.py`: new regression — re-issuing targets with the same object set but drifted diagnostics content now succeeds; existing idempotency test still passes (2 pass).


pablonyx:
## 🔍 Adversarial review (3 agents: spec-alignment · server-correctness · client-correctness, each after reading `specs/`)

The **server fix is sound** — idempotent by object identity, completion still verifies each object's *size* against the real S3 object, and the manifest overwrite loses nothing the completion/tracker paths read. But the review surfaced real issues worth addressing before merge:

### �Blocker
**1. The global retry cap silently drops blocked-on-user reports.** `supportReportRetriesExhausted` runs for *every* retryable failure, including `auth_required` / `cloud_unconfigured` / `storage_unconfigured` / `dev_auth_bypass` — states that intentionally keep the report queued ("Report is queued." copy) waiting on the user to sign in or the server to be configured. With `auth_required` retrying every 5 min, the 8-attempt cap fires in ~35–40 min, and `attemptCount` persists in localStorage across restarts — so a user who files a report, doesn't sign in, and returns tomorrow finds it **dropped, attachments deleted, unrecoverable**. The 48h age cap drops any blocked report regardless of attempts.
→ Scope the cap to `failure.kind === "transient"` only. Conflict/payload kinds are already terminal via `retryable: false`, so the cap is genuinely only a backstop for `transient`. (`use-support-report-upload-queue.ts:163`)

### Major
**2. "already completed" is mis-toasted as a permanent failure.** `isUploadConflictError` matches `"upload is already completed"` → terminal `upload_conflict` → *"This report can no longer be sent."* But already-completed almost certainly means it **succeeded** on a prior attempt (complete landed, client lost the response). Telling the user to re-file produces a duplicate. → Split it into a success / idempotent-cleanup path (silent or success toast). (`report-upload-failure.ts:182`)

**3. The change contradicts specs that weren't updated.** `specs/codebase/features/support-reporting.md:130-132` states *"the object manifest is immutable once written … a retry with different diagnostics/attachment metadata is rejected"*; the runbook repeats it (`specs/developing/debugging/support-reports.md:586,680`). This PR deliberately makes per-object content metadata mutable on re-issue. → Update those lines to the new contract: object **set** (keys + intent) immutable, per-object content (size/sha256) refreshed from the latest re-issue.

**4. Terminal classification matches English server prose, not a stable code.** `isUploadConflictError` substring-matches free-text server messages. A server reword silently reverts these to `transient` → resurrects the exact retry-forever bug this PR fixes. The error pipeline already extracts a structured `code` for the other kinds. → Return a code (e.g. `support_report_upload_conflict`, `support_report_already_completed`) and match on it; keep substrings as a legacy fallback. (`report-upload-failure.ts:177`)

### Minor / nits
- **Off-by-one:** the cap allows **7** real attempts, not 8 — `attemptCount = entry.attemptCount + 1` is checked `>= 8` before the upload, and the new test (`attemptCount: 7 → dropped`) confirms the 8th upload never runs. Use `>= 9` or document the budget as 7. (`report-upload-failure.ts:193`)
- **Coverage gap:** the pre-existing 400 test (`test_support_api.py:445`) now passes via *intent* validation, not the manifest-key check; the new same-intent/different-keys rejection branch is untested (and near-unreachable since object keys are deterministic). Add a comment or a targeted test.
- **sha256 is consistency-only** (pre-existing, not introduced here): completion verifies object *size* against S3, but sha256 is client-manifest vs client-complete — not an S3 integrity guarantee. Worth a code comment so it isn't mistaken for one.
- cap-drop `showToast(...)` omits the `type` arg (defaults to error-style) — confirm intended.

🤖 Multi-agent review via Claude Code

pablonyx:
## ✅ Review addressed — `01a4cdb`

- **Blocker #1 (cap drops blocked-on-user reports):** Fixed. The cap is now scoped to `failure.kind === "transient"`. `auth_required` / `cloud_unconfigured` / `storage_unconfigured` / `dev_auth_bypass` stay queued until the user signs in / configures — never dropped on a timer. New queue test: *keeps a blocked-on-sign-in report queued instead of dropping it on the cap*.
- **Major #2 (already-completed mis-toasted as failure):** Fixed. New `already_completed` kind → idempotent cleanup + a neutral *"Report already sent."* info toast, not the "can no longer be sent" copy. New queue test covers it.
- **Major #3 (specs contradicted):** Fixed. Updated `specs/codebase/features/support-reporting.md` and the runbook (`specs/developing/debugging/support-reports.md`) to the new contract: object **set** (keys + intent) immutable, per-object content (size/sha256) refreshed on re-issue.
- **Major #4 (brittle prose matching):** Fixed. Added stable server codes `support_report_upload_conflict` and `support_report_already_completed` (subclasses of `SupportReportUploadInvalid`); the client classifies on `code` first, substring only as a legacy fallback.
- **Minor (coverage gap):** Fixed. The re-issue test now also asserts the same-intent / different-object-key set returns `support_report_upload_conflict`, exercising the key-comparison branch directly.
- **Minor (sha256 integrity):** Added a comment in `expected_completed_object_entry` clarifying it's a consistency check; object *size* is the only S3-verified invariant.
- **Minor (off-by-one):** Re-checked — not a bug. The upload runs in the `try`; the cap is evaluated only in the `catch` *after* the attempt fails, so the 8th attempt is made and fails before the job is dropped (the `attemptCount: 7 → dropped` test confirms the 8th attempt executed). Budget is 8 real attempts as documented.
- **Nit (toast type):** Left as-is — failure toasts intentionally omit `type` (error-style default), matching the existing pattern; only the success/already-sent path uses `"info"`.

Tests: 21 client + 9 server green.

🤖 via Claude Code

pablonyx:
## 🔁 Round-2 review addressed — `4f317ac`

Ran a second adversarial pass (server · client · spec+test lenses). Findings and resolutions:

**Server (real, MAJOR):** `SupportReportAlreadyCompleted` was raised for *any* non-`{created,uploading}` status, but the status domain includes `failed`/`abandoned` (`support.py:28`). That would tell the client "already sent" → it silently discards an **undelivered** report and its attachments. Fixed: only `status == "completed"` → `already_completed`; `failed`/`abandoned` → `support_report_upload_conflict` ("start a new report").

**Client (real, MAJOR):** scoping the cap to `transient` left blocked/config states (`storage_unconfigured` on a never-configured server) retrying **forever** — the age backstop was computed but short-circuited. Fixed: the **attempt** cap applies to `transient` only; a **48h age backstop** now applies to *all* retryable failures. Blocked states aren't dropped on a fast timer, but nothing retries forever.

**Client (real, MAJOR):** other `support_report_upload_invalid` 400s (diagnostics too large, object-size mismatch, workspace selection…) fell through to `transient` and retried until the backstop. Fixed: any `support_report_upload_invalid` / HTTP 400 is now terminal `upload_rejected`.

**Specs (real, MAJOR/MINOR):** found 3 more stale "immutable manifest" references (`/complete` description, lifecycle step 5, `/complete` troubleshooting) — corrected. Added the new terminal/success codes to the idempotency guidance.

**Tests (real, MAJOR/MINOR):** the cap tests passed for the wrong reason — the fixture's `2026-05-31` date trips the age cap regardless of attempt count. Fixed `makeSupportReportJob` to take `createdAt`; the attempt-cap test now uses a fresh date (isolating the attempt cap), and added: blocked-aged-out (age backstop drops it), `upload_conflict` end-to-end (drops + conflict toast, not mis-routed to success), attempt-cap boundary (7→false / 8→true), and `upload_rejected`.

**Dismissed — false positive:** one reviewer flagged the new error codes as "fictional / never reach the client." They do: `main.py:131` surfaces `exc.code`, the subclasses override `code`, and `test_support_api.py:459/576` assert `support_report_upload_conflict` end-to-end (passing). The reviewer read stale state.

Tests green: **client 25/25**, **server support API 9/9 + hardening 4/4**.

🤖 via Claude Code

PR #740, opened 2026-06-18
feat(desktop): persistent loading→auth shell (no brand-mark jump)

pablonyx:
## Problem
The loading screen and the sign-in screen shared the same shell *visually* but were two **separate component trees** (`SessionCheckScreen` overlay → `LoginScreen` underneath). The brand mark therefore lived in two different instances at slightly different vertical positions, so it **jumped/slid** during the loading→auth cross-fade.

## Fix
One persistent shell that stays mounted across `bootstrapping → anonymous`:

- **`AuthScreenLayout`** — shared shell. One-line heading (`Let's get your life's work done.`) pinned in place; a **reserved-height action slot** so the mark never moves; the loading skeleton morphs into the GitHub button (rises into place, no downward drift).
- **`AuthShell`** — thin connected wrapper that owns the GitHub sign-in wiring.
- **`BootstrappedRoute`** reworked — keeps a single `AuthShell` mounted, so the living mark **never re-mounts** (animation stays continuous). Fade-to-reveal happens **only** for the app destination; the login path flips `loading → auth` in place with no fade. The mark `resolve → fade` handoff is preserved for the authenticated reveal.
- **`AuthRequiredGate` removed** — `BootstrappedRoute` now owns the auth-required gate; route tree unwrapped.
- **`LoginPage`** (`/login`) repointed at the shared layout (incl. continue-locally); heading/hint copy moved into `auth-copy.ts`.

## Playground (DEV only)
`/playground/auth` renders both surfaces with a toggle:
- **Shared layout (fixed)** — the new persistent shell (Loading ↔ Auth).
- **Current (jumpy)** — the old two-screen cross-fade, for A/B.

The old `SessionCheckScreen` / `LoginScreen` are kept solely for the playground A/B (and `PlaygroundLoadingStates`); they're no longer in the production flow.

## Testing
- `tsc --noEmit` clean.
- `AuthGate.test.tsx` — 8/8 pass: gate loading→reveal lifecycle (shell mocked) + `AuthScreenLayout` coverage (constant heading, action disabled while loading, continue-locally, appearance boundary).


apps/desktop/src/components/auth/AuthScreenLayout.tsx
```diff
@@ -0,0 +1,152 @@
...
+                      variant="ghost"
+                      size="sm"
+                      onClick={onContinueLocally}
+                      className="inline h-auto px-0 py-0 text-foreground underline underline-offset-4 hover:text-muted-foreground"
+                    >
+                      {AUTH_LOGIN_LABELS.continueLocallyInline}
+                    </Button>
+                  </p>
+                )
+                : null}
```
pablonyx: minor (onboarding.md "Current Flow" step 4 + "Surface Ownership: Desktop = Start GitHub-first sign-in"): the message line falls through to `null` in the common case (GitHub available, no error, `canContinueLocally` false). The old `LoginScreen` always rendered the `AUTH_LOGIN_LABELS.detail` line ("Start by signing in with GitHub.") under the heading. The new auth screen now shows a bare "Continue with GitHub" button with no guidance text on both the standalone `/login` page and the in-place BootstrappedRoute gate. The spec wants the GitHub-first path to be clearly named. Fix: render `AUTH_LOGIN_LABELS.detail` as the default branch instead of `null`.

apps/desktop/src/copy/auth/auth-copy.ts
```diff
@@ -3,6 +3,14 @@ export const AUTH_GATE_LABELS = {
   loadingSubtext: "Proliferate is restoring your account before opening the workspace.",
 } as const;
 
+// Shared by the persistent initial-page shell (loading -> auth). The heading is
+// identical in both modes so it never reflows; the loading hint sits in the
+// reserved action slot where the GitHub button lands.
+export const AUTH_SCREEN_LABELS = {
+  heading: "Let's get your life's work done.",
```
pablonyx, line 10: nit (copy drift, not a hard spec violation): this new `AUTH_SCREEN_LABELS.heading` ("Let's get your life's work done.") is now what real users see on `/login` and the in-place gate, while the still-present `AUTH_LOGIN_LABELS.title` ("Let's help you get your life's work done.") survives only in the dead `LoginScreen` (now used solely by the dev playground). onboarding.md treats the signed-out -> signed-in handoff copy as in-scope, so having two divergent first-launch headings is a maintenance hazard. Fix: collapse to one heading constant and either delete `LoginScreen` or point it at the same copy so onboarding copy stays single-sourced.

apps/desktop/src/components/auth/AuthGate.tsx
```diff
@@ -1,66 +1,102 @@
...
   useEffect(() => {
-    if (overlayState !== "fading") {
+    if (destination !== "app" || fadeState !== "fading") {
       return
     }
-
     const timeout = window.setTimeout(handleFadeComplete, 220)
     return () => window.clearTimeout(timeout)
-  }, [handleFadeComplete, overlayState])
+  }, [destination, fadeState, handleFadeComplete])
```
pablonyx, line 79: **blocker — login → authenticated leaves the app permanently obscured.**

Trace the standard production boot: `bootstrapping` (fadeState inits to `"checking"`) → `anonymous` + authRequired → `destination==="login"`. The status effect (L40-46) flips `"checking"→"resolving"`, so the gate now sits at `destination==="login", fadeState="resolving"`.

While on login, `markComplete` is `true` (it's `destination !== "loading"`), so the persistent `ProliferateLivingMark` reaches `icon-hold`, fires `onResolved`, and **latches `resolvedRef.current = true` forever** (see ProliferateLivingMark L114-120). `handleResolved` no-ops because `destination !== "app"`.

Now the user signs in → `authenticated` → `destination==="app"`, but `fadeState` is still `"resolving"` (the status effect only promotes `"checking"`, never `"resolving"→…`). So:
- Early return `destination==="app" && fadeState===null` (L71) is **false**.
- `isFadingOut` is false (fadeState is `"resolving"`, not `"fading"`) → the `z-50` shell stays at `opacity-100`.
- The shell is the SAME mounted instance, so the mark never re-mounts and **never re-fires `onResolved`** → `handleResolved` is never called → `fadeState` never reaches `"fading"` → the fade-complete timeout (L62-68) never arms.

Result: `fadeState` is stuck at `"resolving"`, the full-screen shell covers the workspace indefinitely after a successful sign-in. This is the core regression of the "keep the same shell mounted" design — the latched mark callback can't drive a second reveal.

Concrete fix: don't depend on the mark callback to reveal the app. Drive the fade from an effect on `destination`, e.g. when `destination` becomes `"app"` and `fadeState` is `"resolving"`/`null`-but-not-settled, schedule `"fading"` directly; or reset the mark's resolved latch (remount via `key`) / add `onResolved` re-fire support. At minimum, in the status effect promote `"resolving"`/any non-null state toward the app reveal when `destination==="app"`.

apps/desktop/src/components/auth/AuthGate.tsx
```diff
@@ -1,66 +1,102 @@
...
+  status: ReturnType<typeof useAuthStore.getState>["status"],
+  authRequired: boolean,
+): GateDestination {
+  if (status === "bootstrapping") {
+    return "loading"
+  }
+  if (status === "authenticated" || !authRequired) {
+    return "app"
+  }
+  return "login"
```
pablonyx, line 27: Spec-conformance check (passes, noting for the record): `resolveDestination` correctly preserves the old `AuthRequiredGate` semantics against product-auth.md / onboarding.md — `authenticated || !authRequired` -> app, anonymous+required -> in-place sign-in. Local mode (anonymous + auth-not-required) still falls through to the workspace, matching "Local workspaces are available without an account" and the invariant that readiness is not inferred from local UI state. The in-place `login` gate keeps the URL at the original deep-link path (no Navigate to /login) and signs in in place, so deep links and `getRedirectTarget` resume correctly without a redirect round-trip. No regression vs the deleted gate. One thing to double-check in QA per onboarding.md minimum smoke step 3: confirm the GitHub-required readiness screen (post sign-in, missing provider) still renders downstream — this gate only covers identity readiness, not provider readiness.

pablonyx:
**Spec-conformance review (subagent):** checked against `specs/codebase/features/product-auth.md` and `onboarding.md`.

**Verdict: no blockers or majors. The auth-required semantics are preserved; the deleted `AuthRequiredGate` boundary was correctly absorbed into `BootstrappedRoute`.**

What I verified holds against spec:
- `resolveDestination` reproduces the old gate exactly: `authenticated || !authRequired` -> app, `anonymous + authRequired` -> in-place sign-in. Local mode (anonymous, auth not required) still falls through to the workspace ("Local workspaces are available without an account").
- Dev bypass (`isDevAuthBypassed` -> authenticated state) -> app destination. Unchanged.
- `isProductAuthRequired` is still the single source for the gate; readiness is not inferred from local UI state (onboarding Invariants).
- `/setup` redirect and deep-linked routes still sit behind the gate (BootstrappedRoute returns the shell, not `<Outlet/>`, when `destination==="login"`), so an anonymous+required deep link still shows sign-in. No deep-link regression.
- The in-place gate signs in without a `/login` redirect round-trip; standalone `/login` (PublicOnlyRoute) + `getRedirectTarget` still resume to `from`. Continue-locally is gated on `!isProductAuthRequired()` in both `LoginPage` and (absent) the in-place shell — consistent with "no GitHub-free demo mode for normal users".
- GitHub remains the only readiness provider; this PR does not touch provider/run readiness. Good.

Findings (all minor/nit, see inline):
1. **minor — AuthScreenLayout.tsx:147** the message line falls through to `null` in the common case, so the new auth screen drops the old "Start by signing in with GitHub." guidance line on both `/login` and the in-place gate. onboarding.md wants the GitHub-first path clearly named. Render `AUTH_LOGIN_LABELS.detail` as the default branch.
2. **nit — auth-copy.ts:10** two divergent first-launch headings now exist: the live `AUTH_SCREEN_LABELS.heading` vs the orphaned `AUTH_LOGIN_LABELS.title` in the now-dead `LoginScreen` (only the dev playground still renders it). Single-source the copy and delete or repoint `LoginScreen`.
3. **process — onboarding.md minimum smoke step 3:** confirm the downstream GitHub-required readiness screen (signed in, missing provider) still renders — this gate only covers identity readiness.

`LoginScreen` is now production-dead (only imported by the playground + tests); worth deleting in this PR or a fast follow to avoid the copy drift in #2.

apps/desktop/src/components/auth/AuthShell.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export function AuthShell({ mode, markComplete, onMarkResolved }: AuthShellProps) {
+  const {
+    signIn,
+    submitting,
+    error,
+    signInAvailable,
+    signInChecking,
+    signInUnavailableDescription,
+  } = useGitHubSignIn();
```
pablonyx: **major — `useGitHubSignIn` (with its react-query availability poll) now runs during the loading shell for *every* user, including those who become authenticated.**

Before this PR, `useGitHubSignIn` was only mounted inside `AuthRequiredGate`'s `SignInGateView` (anonymous-only) or `LoginPage`. Now `AuthShell` is mounted by `BootstrappedRoute` for the entire `loading` phase for all users. `useGitHubSignIn` → `useGitHubDesktopAuthAvailability()` is a `useQuery` with `enabled: controlPlaneReachable`, `refetchInterval: 15_000`. So a user with a valid session who will land straight in the app now fires the GitHub-auth-availability request (and a 15s poll) during bootstrap, then throws the result away when the shell fades. Pure waste, plus extra control-plane load during the most latency-sensitive window (boot).

Fix: only mount the sign-in wiring when it's actually needed (mode === "auth"), e.g. split AuthShell so the loading mode renders `AuthScreenLayout mode="loading"` with no hook, and the auth mode renders the hook-backed variant; or gate the availability query's `enabled` on the gate being on the login destination.

apps/desktop/src/components/auth/AuthGate.test.tsx
```diff
@@ -5,8 +5,21 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
...
+// availability query + capability hooks. These BootstrappedRoute tests only care
+// about the overlay/reveal lifecycle, so stub the shell to a marker that echoes
+// the mode it was rendered with.
+vi.mock("@/components/auth/AuthShell", () => ({
+  AuthShell: ({ mode }: { mode: string }) => (
+    <div data-testid="auth-shell" data-mode={mode}>
+      shell
+    </div>
+  ),
+}));
```
pablonyx: **major — this mock hides the blocker the suite should catch.** Stubbing `AuthShell` to a static div removes the real `ProliferateLivingMark` and never invokes `onMarkResolved`, so the tests can never exercise the mark→`handleResolved`→fade path — the exact path that deadlocks on the login→authenticated transition (see my AuthGate comment). Combined with the one BootstrappedRoute test only covering `bootstrapping→authenticated` (and never asserting the shell actually leaves, i.e. `fadeState===null`), the lifecycle is effectively untested. Add a test that drives `bootstrapping → anonymous(authRequired) → authenticated` with a mark stub that *calls* `onMarkResolved` (or fake timers + the real mark), and assert the shell unmounts and the workspace is revealed.

apps/desktop/src/components/auth/AuthScreenLayout.tsx
```diff
@@ -0,0 +1,152 @@
...
+        </div>
+
+        {/* Reserved actions slot: identical height in both modes so the block
+            above never shifts. Two absolutely-positioned layers cross-fade. */}
+        <div className="relative h-11">
+          {/* Loading layer: a skeleton sitting where the button will land.
+              On exit it slides up so nothing drifts downward. */}
+          <div
+            className={twMerge(
+              "absolute inset-0 transition-all duration-300 ease-out",
```
pablonyx, line 72: **[major] Reduced-motion not respected on the cross-fade.** Both layer wrappers use `transition-all duration-300 ease-out` plus `-translate-y-1` / `translate-y-1` transforms (here and L89), but neither is gated by `motion-reduce:`. The mark itself has a reduced-motion branch and the BootstrappedRoute overlay uses `motion-reduce:transition-none`, but this inner skeleton→button slide/fade is the ONLY animation that fires on the login path (loading→auth in place has no outer overlay transition). So `prefers-reduced-motion: reduce` users still get the full 300ms slide+fade. Add `motion-reduce:transition-none motion-reduce:translate-y-0` to both layers (the opacity swap can stay if you want an instant snap, or also drop it).

apps/desktop/src/components/auth/AuthGate.tsx
```diff
@@ -1,66 +1,102 @@
...
-      </>
-    )
+  // App fully revealed: the shell is gone, only the workspace remains.
+  if (destination === "app" && fadeState === null) {
+    return <Outlet />
   }
 
-  return <Outlet />
+  const shellMode = destination === "login" ? "auth" : "loading"
+  const markComplete = destination !== "loading"
```
pablonyx, line 87: **minor — behavior change on the login screen mark.** Here `markComplete` is `destination !== "loading"`, which is `true` for `destination==="login"`. So on the steady-state sign-in screen the living mark runs its sweep once, then freezes as the static icon (ProliferateLivingMark `icon-hold` short-circuits the exit while `complete` is true). The old `AuthRequiredGate`/`LoginScreen` rendered `<ProliferateLivingMark/>` with `complete` defaulting to `false`, i.e. it looped the braille sweep forever on the login screen. If the perpetual animation on the sign-in screen was intentional, pass `markComplete={destination === "app"}` (or `false` for login) instead.

apps/desktop/src/components/auth/AuthScreenLayout.tsx
```diff
@@ -0,0 +1,152 @@
...
+                  ? AUTH_LOGIN_LABELS.checking
+                  : AUTH_LOGIN_LABELS.signIn}
+              {!submitting && <ArrowRight className="h-4 w-4" />}
+            </Button>
+          </div>
+        </div>
+
+        {/* Reserved message line so error / unavailable / local text never
+            shifts the layout above it. */}
+        <div className="-mt-4 min-h-5 text-center">
```
pablonyx: **[major] The reserved message line only reserves ONE line, so multi-line states still shift the mark.** `min-h-5` = 1.25rem = exactly one `text-sm` line. But the real `githubSignInUnavailableDescription` strings are full sentences — e.g. "This environment can reach the control plane, but desktop GitHub sign-in is not configured." and "Checking whether GitHub sign-in is configured for this environment." At `max-w-md` minus `p-8` (~384px usable) these wrap to 2 lines (~40px). Since this block is the LAST child of a vertically-centered column (`justify-center` + `space-y-8`), a 2-line message grows the column and pushes the whole stack — including the mark — UP by ~10px. So the "zero vertical shift" guarantee holds for the loading→auth happy path but breaks the moment GitHub is unavailable/checking or an error appears. Reserve 2 lines (`min-h-10`) or clamp to one line, and ideally measure the tallest real string.

apps/desktop/src/components/auth/AuthScreenLayout.tsx
```diff
@@ -0,0 +1,152 @@
...
+                    {AUTH_LOGIN_LABELS.detailWithLocalPrefix}{" "}
+                    <Button
+                      type="button"
+                      variant="ghost"
+                      size="sm"
+                      onClick={onContinueLocally}
+                      className="inline h-auto px-0 py-0 text-foreground underline underline-offset-4 hover:text-muted-foreground"
+                    >
+                      {AUTH_LOGIN_LABELS.continueLocallyInline}
+                    </Button>
```
pablonyx: **[nit] Lost trailing period on the continue-locally sentence.** The old `LoginScreen` rendered `{detailWithLocalPrefix} <Button>start locally</Button>.` — note the literal `.` after the button (LoginScreen.tsx L55). Here the period is gone, so the sentence reads "Continue with GitHub, or start locally" with no terminal punctuation. Re-add the `.` after the inline Button for copy parity.

apps/desktop/src/copy/auth/auth-copy.ts
```diff
@@ -3,6 +3,14 @@ export const AUTH_GATE_LABELS = {
...
+  loadingHint: "Restoring your session…",
+} as const;
+
 export const AUTH_LOGIN_LABELS = {
   title: "Let's help you get your life's work done.",
```
pablonyx, line 15: **[minor] `AUTH_LOGIN_LABELS.title` (and `.detail`, plus both `AUTH_GATE_LABELS`) are now dead in production and diverge from the new heading.** Live code uses `AUTH_SCREEN_LABELS.heading` ("Let's get your life's work done."). The old `title` ("Let's help you get your life's work done.") and `detail` are only referenced by `LoginScreen.tsx` / `SessionCheckScreen.tsx`, which after this PR are imported solely by the dev playground + tests — no production route renders them. Two competing headings now coexist; the next person editing copy can easily change the wrong one. Either delete the dead screens+labels or add a comment marking them playground-only. The heading wording change ("help you get" → "get") also isn't called out in the PR description — confirm it's intentional.

pablonyx:
**Correctness review (subagent):** Focused on the BootstrappedRoute state machine, the persistence guarantee, and the AuthRequiredGate removal.

**Blocker**
- **login → authenticated deadlocks the app behind the shell.** The standard prod boot is `bootstrapping → anonymous(authRequired) → login`, during which the status effect promotes `fadeState` `"checking"→"resolving"` and the persistent mark (with `markComplete=true`) fires `onResolved` once and latches `resolvedRef=true` forever. When the user then signs in (`→ app`), `fadeState` is still `"resolving"` (the status effect only ever promotes `"checking"`), the early-return needs `fadeState===null`, and the SAME mounted mark can't re-fire `onResolved` → `handleResolved` never runs → `fadeState` never reaches `"fading"` → the fade-complete timeout never arms. The full-screen `z-50` shell stays at `opacity-100` over the workspace **permanently after a successful sign-in.** The "keep the same shell mounted" design is precisely what breaks the second reveal. Fix: drive the app fade from a `destination` effect (don't depend on the mark callback for the post-login reveal) and/or reset the mark latch.

**Major**
- **Wasted GitHub-availability query during loading for all users.** `AuthShell` is now mounted for the whole `loading` phase for everyone, so `useGitHubSignIn → useGitHubDesktopAuthAvailability` (react-query, 15s poll) fires for soon-to-be-authenticated users and is thrown away on reveal — extra control-plane load in the most latency-sensitive window. Previously this hook was anonymous-only (AuthRequiredGate/LoginPage). Gate it on the login destination.
- **Tests mock the defect away.** Stubbing `AuthShell` to a static div removes the real mark and never calls `onMarkResolved`, so the mark→fade lifecycle (and the deadlock above) is never exercised; the one reveal test also never asserts `fadeState===null`/shell unmount. Add a `bootstrapping → anonymous(authRequired) → authenticated` test that drives `onMarkResolved` and asserts the workspace is revealed.

**Minor**
- Login-screen mark now runs once and freezes (markComplete is true for `login`), vs. the old looping sweep. Pass `markComplete={destination === "app"}` if perpetual animation on sign-in was intended.

**Verified clean**
- Removing `AuthRequiredGate` lost no behavior: `/setup` redirect + `UserPreferencesGate` remain under `BootstrappedRoute`, and app-vs-login parity matches the old `!requiresAuth || authenticated` logic.
- Shell persistence across `loading→login` holds (the `{false}` placeholder at child index 0 keeps the `<div>`/AuthShell at index 1 stable when `<Outlet/>` is inserted on the app path).
- Direct mount when already `authenticated` returns `<Outlet/>` immediately (no flash); happy `bootstrapping→authenticated` reveal works.
- AuthScreenLayout reserved heights, aria-hidden/tabIndex on the inactive layer, and disabled loading-mode button look correct.

pablonyx:
**UX/a11y review (subagent):**

The core mechanism is sound: one persistent element tree means the living mark never re-mounts, the `<h1>` is constant, the mark is fixed `size-12` in both braille/resolved states, and the action slot is a fixed `h-11` with absolutely-positioned cross-fading layers. Box model checks out too — the primary Button has no border, the skeleton has `border border-border/50`, but both are `h-11` with `box-border`, so the 1px border sits inside and heights are identical. For the loading→auth happy path, zero vertical shift holds.

But the "zero shift" guarantee is narrower than the PR claims, and reduced-motion is missed:

**Major**
- **Reduced-motion not respected (AuthScreenLayout L72/L89).** The skeleton→button cross-fade uses `transition-all duration-300` + `±translate-y-1` with no `motion-reduce:` gate. This is the *only* animation on the login path (loading→auth in place has no outer overlay transition — the overlay's `motion-reduce:transition-none` only covers the app-reveal fade). So reduced-motion users still get the full slide+fade. Add `motion-reduce:transition-none motion-reduce:translate-y-0` to both layers.
- **The reserved message line reserves only ONE line (AuthScreenLayout L123).** `min-h-5` = one `text-sm` line, but the real unavailable/checking strings ("This environment can reach the control plane, but desktop GitHub sign-in is not configured.", etc.) wrap to 2 lines at `max-w-md`. As the last child of a `justify-center` column, a 2-line message grows the column and pushes the mark UP ~10px. The mark *does* move once GitHub is unavailable/checking or an error shows. Reserve `min-h-10` or clamp.

**Minor**
- **Dead/divergent copy (auth-copy.ts L15).** `AUTH_LOGIN_LABELS.title`/`.detail` + both `AUTH_GATE_LABELS` are now only used by `LoginScreen`/`SessionCheckScreen`, which after this PR are dev-playground/test-only — no production route renders them. The old title ("…help you get…") now diverges from the new heading ("…get…"). Delete the dead screens+labels or mark them playground-only; and confirm the heading wording change is intentional (not in the PR description).

**Nit**
- **Lost trailing period (AuthScreenLayout L144).** Old LoginScreen had a literal `.` after the inline "start locally" button; the new layout drops it.

**Looks good**
- `aria-hidden` on the inactive layer + `disabled`/`tabIndex=-1` on the loading-mode button means it's correctly not announced/focusable.
- Motion direction is consistent (skeleton exits up, button rises into place) — non-jarring.
- The dropped loading subtext (`AUTH_GATE_LABELS.loadingSubtext`) reads as intentional given the new single-line loading hint.

One thing I could not verify (no pixels): exact 2-line wrap thresholds — worth eyeballing the unavailable/error states at the real `max-w-md` width before merge.

pablonyx:
**Addressed the subagent review** (commit 982be2293):

**Blocker — post-sign-in shell deadlock** ✅ Fixed. The persistent `ProliferateLivingMark` only fires `onResolved` once, so on the `bootstrapping → login → sign-in → app` path the reveal fade never armed and the shell covered the workspace forever. Reveal is now driven by a `markResolved` state flag (set by `onResolved`) rather than the one-shot callback, so the fade fires on both first-load and post-sign-in reveals. Added a regression test (`reveals the workspace after sign-in without leaving the shell stuck`) — it drives bootstrapping → anonymous(auth-required) → authenticated and asserts the shell unmounts.

**Major — reduced-motion** ✅ Fixed. The skeleton/button cross-fade (`transition-all` + `±translate-y-1`) now carries `motion-reduce:transition-none motion-reduce:translate-y-0`; it was the only animation on the login path (the overlay's gate only covers the app-reveal fade).

**Major — message line could shift the mark** ✅ Fixed. The message line is now absolutely anchored below the action slot (`top-full`), so wrapping `unavailable`/`error` text no longer changes the centered column height. Mark stays pinned in all auth sub-states.

**Major — availability poll during loading** ✅ Fixed. Threaded `enabled` through `useGitHubSignIn` → `useGitHubDesktopAuthAvailability`; `AuthShell` passes `enabled: mode === "auth"`, so the 15s poll no longer runs during the loading phase for everyone.

**Major — test mocked the defect away** ✅ Fixed. The `AuthShell` test mock now mirrors the real mark (fires `onMarkResolved` when `markComplete`), so the reveal lifecycle is genuinely exercised.

**Nit — trailing period** ✅ Restored after the inline "start locally".

**Intentionally kept:**
- No subtext under the heading (one-line heading is the requested design); the old `AUTH_LOGIN_LABELS.title`/`AUTH_GATE_LABELS` copy stays only because the dev playground's "Current (jumpy)" A/B renders the old `LoginScreen`/`SessionCheckScreen` verbatim.

`tsc --noEmit` clean; `AuthGate.test.tsx` 9/9.

pablonyx:
**Second-pass verification (subagent):** Re-traced the `markResolved`-driven reveal across every transition against the fix commit (982be2293).

## Blocker (sign-in reveal deadlock): genuinely FIXED ✅
Traced all paths in `AuthGate.tsx`:
- bootstrapping → authenticated (first launch): mark fires once → `markResolved=true` → fade → `Outlet`. ✅
- **bootstrapping → anonymous(authRequired) → authenticated** (the deadlock path): on login the mark resolves and `markResolved` latches `true`; it is NOT reset on the login→authenticated transition (reset only happens on `bootstrapping`), so the reveal effect fires immediately on reaching `app`. **Deadlock resolved.** ✅
- bootstrapping → anonymous(!authRequired) → app (local): ✅
- direct-mount authenticated → instant `Outlet` (no flash); direct-mount anonymous→signin → instant reveal (no fade, but no deadlock). ✅
- mid-fade sign-out → re-signin recovers via the leftover `fading` timeout. ✅

No reachable case leaves `fadeState` stuck non-null with `destination==="app"`, and no early-unmount mark-jump regression. The `enabled: mode === "auth"` gating, the `motion-reduce:` a11y classes (Tailwind 4.3.1 — variant supported, already used elsewhere), the `absolute top-full` message line, and the optional `enabled` hook params are all correct and backward-compatible.

## MAJOR — the new regression test is a tautology (does not guard the bug)
`AuthGate.test.tsx` "reveals the workspace after sign-in without leaving the shell stuck": the mock `AuthShell` fires `onMarkResolved` from a `useEffect` keyed on `[markComplete, onMarkResolved]`. The real `ProliferateLivingMark` fires `onResolved` **exactly once ever** (`resolvedRef` latch) and will NOT re-fire on login→app. The mock instead re-fires whenever `onMarkResolved` (i.e. the pre-fix `handleResolved`, deps `[destination]`) changes identity — which it does on login→app — so it masks the one-shot behavior that actually causes the deadlock.

I proved this empirically: I swapped in the **pre-fix** `AuthGate.tsx` (commit 3b296653b) and ran the new test — **it PASSES against the buggy code.** So it would not have caught the original regression.

Fix: make the mock model the one-shot latch (fire at most once), e.g. a `useRef(false)` guard so `onMarkResolved` is called only on the first render where `markComplete` is true. With that, the test fails on pre-fix and passes on the fix.

## Minor — latent re-bootstrap deadlock (not currently reachable)
`markResolved` is reset to `false` on `bootstrapping` (AuthGate.tsx:48), but `ProliferateLivingMark.resolvedRef` never resets. If the shell ever goes `login → bootstrapping → app` **without unmounting** (mark instance survives), the mark won't re-fire `onResolved`, `markResolved` stays `false`, and the app reveal deadlocks. Today this is unreachable: `bootstrapAuth` runs once at mount and `signOut` goes straight to `anonymous` (never back to `bootstrapping`), and any path that reaches `app` unmounts the mark first. Worth a comment or a defensive reset, but not blocking.

## Nits
- Reveal-arming effect (AuthGate.tsx:67) tracks/cancels only the outer rAF; the inner rAF is uncancellable. Harmless given downstream guards, but slightly leaky.
- loading→auth flip now starts the availability query fresh (was polling during loading before), so the GitHub button can briefly show "Checking…". Functionally correct; minor UX.

**Verdict:** Blocker is truly fixed in the product code; no remaining blockers. One **major**: the regression test is a tautology and must be tightened to actually guard the fix.

apps/desktop/src/components/auth/AuthGate.test.tsx
```diff
@@ -1,12 +1,51 @@
...
+    }: {
+      mode: string;
+      markComplete: boolean;
+      onMarkResolved?: () => void;
+    }) => {
+      useEffect(() => {
+        if (markComplete) {
+          onMarkResolved?.();
+        }
+      }, [markComplete, onMarkResolved]);
```
pablonyx, line 46: MAJOR (tautology): this mock re-fires `onMarkResolved` whenever its deps `[markComplete, onMarkResolved]` change — and `onMarkResolved` (the pre-fix `handleResolved`, deps `[destination]`) changes identity on login→app. The real `ProliferateLivingMark` fires `onResolved` exactly once (`resolvedRef` one-shot latch) and will NOT re-fire on login→app — which is the actual cause of the deadlock. I verified empirically: with the pre-fix `AuthGate.tsx` swapped in, this test PASSES, so it does not guard the regression. Fix: latch it, e.g.

```ts
const fired = useRef(false);
useEffect(() => {
  if (markComplete && !fired.current) {
    fired.current = true;
    onMarkResolved?.();
  }
}, [markComplete, onMarkResolved]);
```
That fails on pre-fix and passes on the fix.

pablonyx:
**Second-pass verdict addressed** (commit ccb9fe1d4):

The reviewer confirmed the blocker fix is correct across all transitions, and caught that the regression **test** was a tautology — the mock re-fired `onMarkResolved` on callback-identity change, so it passed even against the pre-fix code.

Fixed: the `AuthShell` test mock now latches with a `useRef` (fires at most once, mirroring the real `ProliferateLivingMark` `resolvedRef`). **Verified empirically**:
- Pre-fix `AuthGate.tsx` (3b296653b) + new test → `reveals the workspace after sign-in…` **FAILS** (deadlock reproduced).
- Fixed `AuthGate.tsx` + new test → 9/9 pass.

So the test is now a genuine guard.

Remaining items from the review, acknowledged as **won't-fix** (not reachable):
- Latent `login → bootstrapping → app` path where `markResolved` resets but the mark's `resolvedRef` wouldn't — unreachable today (`bootstrapAuth` runs once at mount; `signOut` → `anonymous`, not `bootstrapping`).
- Nits (inner rAF not cancelled — harmless; brief "Checking…" on the loading→auth flip) — cosmetic.

`tsc` clean; 9/9.

PR #742, opened 2026-06-18
feat(transcript): provider-icon links + better file-path resolution

pablonyx:
## What
Inline links in the chat transcript now render as Codex-style **mentions with a provider icon** — a GitHub brand SVG for github hosts, otherwise the **site's own favicon** — with **zero per-provider code**. Before, agent prose showed bare underlined links with no icons.

Favicon resolution is a graceful chain with **no third-party service**: `https://<host>/favicon.ico` → the registrable root domain's favicon (so `console.aws.amazon.com` falls back to `amazon.com`) → no icon. Every request goes to the linked site itself, so no list of linked hosts leaks anywhere.

Mentions are styled to match the existing **file-path mention**: muted blue at rest with no underline, brightening to the foreground color with a dashed underline on hover (the file-path mention got the same hover brighten so the two stay identical).

## Where (package placement)
The shared rendering lives in **`@proliferate/product-ui`** (`ProviderLinkMention`) — no workspace dependency. `MarkdownBody`'s default anchor renders it, so the **transcript, cloud chat, and plan cards all get it for free**; the desktop *content* renderer routes through the same component. This replaces the old `GitHubLinkChip` (bare-autolinks-only, in a renderer that never draws agent prose) — deleted with its now-unused parser. Workspace **file paths still become `FilePathLink`s**; URL-vs-path classification checks external URLs **first** (conservative `http(s)://` / `www.` only), so a real file path is never mistaken for a link.

## File-link correctness (two parts)
1. **Clearer prompt** — `FILE_REFERENCE_INSTRUCTIONS` (shared system-prompt append, all harnesses) now demands the **complete path from the workspace root**, never abbreviated, with a deep monorepo example.
2. **Fuzzy backstop** — when a mention's path doesn't resolve, a resolve-on-click basename/suffix search opens the unique match (so `content/ui/MarkdownRenderer.tsx` still opens the real `apps/desktop/.../MarkdownRenderer.tsx`). Best-effort, case-insensitive, bails on ambiguity; opens the direct path optimistically so correct paths have zero added latency.

## Tests
`ProviderLinkMention.test.tsx` (8 — URL/host/`rootDomain` detection + render), `path-references.test.ts` (fuzzy match), `MarkdownRenderer.test.tsx`, `use-file-reference-actions.test.tsx`; product-ui + desktop typecheck clean in all changed files.


pablonyx:
## ✅ Strand 4 (fuzzy file-resolution backstop) landed — `c566f8c`

No longer deferred. When a file mention's path can't resolve (e.g. an agent dropped leading dirs → `content/ui/MarkdownRenderer.tsx`), open-in-sidebar searches the workspace by basename and opens the unique file whose path **ends with** that suffix — so the click lands on the real file instead of 404ing.

- `pickFuzzyPathMatch` (pure, tested): suffix correction; null on exact-existing / ambiguous / no-match.
- `useFuzzyFileResolver`: lazy, best-effort imperative search — runs **on open, not per render**, and silently no-ops on missing context or search failure.
- Wired into `useFileReferenceActions.openInSidebar`; added a `searchWorkspaceFiles` transport helper.
- Drive-by: fixed a **pre-existing** failure in `use-file-reference-actions.test` (its shell mock omitted `pathIsDirectory`, which `openPrimary` calls — red on clean `main` too).

Tests: `pickFuzzyPathMatch` (4) + actions-hook (1) green; strand-4 files typecheck clean.

pablonyx:
## 🔍 Adversarial review (3 agents: spec-alignment · link-rendering · fuzzy-resolver, each after reading `specs/`)

### High
1. **`isExternalHttpLink` steals relative file paths with a dotted directory segment.** `v1.2/notes.txt`, `release-1.0/notes.md`, `CHANGELOG.md/x`, `a.tsx/b` match the scheme-less `host.tld/path` regex → routed as web links → broken `target=_blank` link **and** a favicon fetch for a bogus "host" (`changelog.md`). Both renderers gate file detection behind `!isExternalHttpLink`, so the file mention is lost. → require a TLD-shaped last label + reject extension-shaped first labels, or run `looksLikePath`/`looksLikeFileReferenceHref` first. (`ProviderLinkMention.tsx:124`)
2. **The favicon leaks every linked host to Google on the cloud/web build.** `MarkdownBody`'s default anchor renders the mention on `apps/web` too (`CloudChatTranscriptRowItems` passes no `renderLink`), so every non-GitHub link → `https://www.google.com/s2/favicons?domain=<host>`. `referrerPolicy=no-referrer` doesn't help — the host is in the query string. Internal AWS/Linear/Vercel/private hostnames leak; no offline path (Globe only on error). → decide: host allowlist / self-hosted favicon proxy / desktop-only `renderLink` opt-out. (`MarkdownBody.tsx:158-162`)
3. **The fuzzy search runs before *every* sidebar open and blocks it.** Even correct paths pay a `files.search` round-trip (cold = full `git ls-files` + stat-per-file, 30s cache) before the viewer opens — pure waste for the common case. → run it only on read-failure (the viewer already surfaces `readQuery.error`), or gate behind an existence signal. (`use-file-reference-actions.ts:69`)

### Major
4. **`limit=25` can mis-correct an actually-valid path.** The "already exists" guard checks only the top 25 results; if the real file is truncated out (common basename like `index.ts`) **and** one other unique suffix-match sits in the 25, a valid path navigates to the wrong file. → raise the limit + filter to exact-basename matches before suffix-checking. (`path-references.ts:45`, `workspace-file-transport.ts:50`)
5. **Spec drift.** `specs/codebase/features/chat-transcript.md:103` ("Web falls back to plain anchors") is now false; the markdown-rules block + ownership map (`:82-104`) omit the new external-link rendering; the `FILE_REFERENCE_INSTRUCTIONS` paraphrases (`:93-94`, `prompt-and-skill-policy.md:43`) don't capture the new "complete path, never abbreviated" contract. Update these + add a note on the Google-favicon external dependency.
6. **Content renderer regresses bare-filename links.** It uses `looksLikePath` (requires a `/`), so `[README.md](README.md)`/`[LICENSE](LICENSE)` → plain `target=_blank` → 404 against the app origin. The transcript renderer uses the looser `looksLikeFileReferenceHref`. The two renderers should share one predicate. (`MarkdownRenderer.tsx:108`)
7. **`isExternalHttpLink` misses host:port / userinfo / IDN** — `localhost:3000/admin`, `example.com:8443/x`, `127.0.0.1:8080`, `user@host/path`, raw-unicode hosts → treated as file paths. Prefer trusting `new URL("https://"+v)` over the hand-rolled regex.
8. **Case-sensitivity mismatch** — the server search is case-insensitive but `pickFuzzyPathMatch` uses case-sensitive `endsWith`; a case-differing ref → server returns the file, matcher rejects it → no correction (and the direct open may have worked on case-insensitive macOS). Use a case-insensitive suffix compare.

### Minor / nits
- **GitHub autolink UX downgrade** — bare GitHub URLs lost the compact `PR #43`/`Issue #45` chip labels (deleted with `github-link.ts`) and now show the full long URL; `www.github.com` no longer gets the brand icon. Intentional Codex-style, but flagging.
- **No in-flight guard** on `openInSidebar` — a double-click fires concurrent searches + double open.
- **Favicon flicker** — `faviconFailed` `useState` resets when a link crosses the streaming stable↔live `MarkdownBody` boundary, re-mounting the ``.
- **Copy marker** — `ProviderLinkMention` omits the `data-chat-transcript-ignore` that the old chips / `FileReferenceBadge` set.
- **Styling duplication** — `INLINE_MENTION_CLASS` + icon-shell classes are byte-identical to `FileReferenceBadge`; share constants so they can't drift.
- **Provider coupling** — `useFuzzyFileResolver` calls the *throwing* `useAnyHarnessWorkspaceContext` at render. Fine today (all file-badge renderers are under the provider), but a non-throwing `useContext` null-check keeps `FileReferenceBadge` usable outside the workspace shell.

### Verified non-issues
product-ui `useState` is **not** a package-law violation (`MarkdownBody` + ~15 transcript components use it); `linkHost` never throws (pre-gated + try/catch); the `:line`-loss and wrong-workspace-id worries don't manifest (line already stripped upstream; correct materialized id used); no production crash path from the throwing context hook.

🤖 Multi-agent review via Claude Code

pablonyx:
## ✅ Review addressed — `8926bbae7`

**High**
1. **`isExternalHttpLink` false-positives** — dropped the greedy scheme-less `host.tld/path` heuristic entirely; now only `http(s)://` and `www.`. `v1.2/notes.txt`, `CHANGELOG.md/x`, `a.tsx/b` (and the host:port / userinfo cases) are no longer stolen from file mentions or sent to the favicon service. Trade-off: a bare scheme-less `github.com/x` markdown link no longer auto-icons — rare in agent output, and never worth losing a real file path. Tests updated.
3. **Fuzzy search no longer blocks the open** — `openInSidebar` now opens the path optimistically and corrects in the background, re-opening only if the corrected path differs. Correct paths pay zero latency.

**Major**
4. **`limit=25` mis-correction** — raised the basename search to 200 (server cap), so the "already exists" guard and the suffix match see the real file instead of a truncated set.
6. **Content renderer bare-filename regression** — switched `MarkdownRenderer` from `looksLikePath` to `looksLikeFileReferenceHref` (matching the transcript renderer), so `[README.md](README.md)` / `[LICENSE](LICENSE)` stay file mentions.
8. **Case sensitivity** — `pickFuzzyPathMatch` now matches case-insensitively (the server search is) and returns the real casing.
5. **Spec drift** — fixed `chat-transcript.md:103` (web links are provider-icon mentions, not plain anchors), added `ProviderLinkMention` to the ownership map + a favicon-dependency note, and tightened the `FILE_REFERENCE_INSTRUCTIONS` paraphrases in `chat-transcript.md` + `prompt-and-skill-policy.md`.

**Deferred (need your call / intentional)**
- **Favicon privacy on the web build** (#2) — every linked host is sent to Google's s2 favicon service; left as-is pending a decision (host allowlist / self-hosted proxy / desktop-only opt-out). Documented the dependency in the spec.
- **GitHub bare-autolink label downgrade** — bare GitHub URLs now show the full URL instead of `PR #43`; this is the intended Codex-style behavior, kept.
- Nits not actioned: favicon flicker across the streaming stable↔live boundary, in-flight guard on double-click (the open-or-focus mode makes the duplicate idempotent), `data-chat-transcript-ignore` marker, cross-package style-constant sharing.

Tests: product-ui `ProviderLinkMention` 7/7, desktop `MarkdownRenderer` 2 + `path-references`/`pickFuzzyPathMatch` 8 + actions 1; typecheck clean in all changed files.

🤖 via Claude Code

PR #743, opened 2026-06-18
feat(desktop): misc UI polish — tab width, markdown headings, codex spinner, default ⌘N

pablonyx:
Miscellaneous desktop UI improvements.

## Changes
- **Tabs** — widen min/max tab width so tabs don't over-shrink (`chrome-layout.ts`).
- **Markdown headings** — both the right-panel file view (`MarkdownRenderer`) and the chat transcript (`MarkdownBody`) only styled h1–h3 (and at body size in the file view); h4–h6 fell back to Tailwind preflight (`font-size: inherit`), so `#` looked the same size as `####`. Now a real descending scale (24/20/17/15/13/12), with a document-style h1/h2 underline in the file view.
- **GFM task lists** — `- [ ]` / `- [x]` no longer render a stray disc bullet next to the checkbox; checkbox aligned inline and accent-tinted.
- **Spinner** — replaced the flat dashed-stroke ring with a codex-style faint full-ring track + solid quarter arc, eased to a calmer 1.4s spin. Shared `ui` primitive, so it updates everywhere (sidebar indicators, update pill, etc.).
- **Default ⌘N mode** — new `defaultNewWorkspaceMode` preference (worktree | local) in Settings → General, plus a plain ⌘N shortcut that dispatches the existing worktree/local action per the preference. Existing ⌘⌥N / ⌘⇧N / ⌘⌃N unchanged.

## Verification
- `tsc -p tsconfig.json` passes for `ui`, `product-domain`, `product-ui`; `tsc --noEmit` passes for `desktop`.
- `scripts/check_frontend_boundaries.py` passes.
- Manual: open a README in the right panel (heading levels + checkbox list), send a `#`…`######` block in chat, toggle Settings → General → "New workspace (⌘N)" and press ⌘N, observe sidebar spinner on an iterating row.

### DIFF
--- a/apps/desktop/src/components/content/ui/MarkdownRenderer.tsx
+++ b/apps/desktop/src/components/content/ui/MarkdownRenderer.tsx
@@ -72,7 +72,7 @@ export function MarkdownRenderer({
     // GFM task lists: drop the stray disc bullet and align the checkbox inline.
     // Descendant selector (not `>input`) so it also matches loose lists where
     // the checkbox sits inside a <p>; avoid flex so nested blocks still stack.
-    "[&_ul.contains-task-list]:list-none [&_ul.contains-task-list]:pl-0",
+    "[&_ul.contains-task-list]:list-none [&>ul.contains-task-list]:pl-0",
     "[&_li.task-list-item]:pl-0",
     "[&_li.task-list-item_input]:mr-2 [&_li.task-list-item_input]:size-3.5 [&_li.task-list-item_input]:align-middle [&_li.task-list-item_input]:accent-link-foreground",
     className,
--- a/apps/desktop/src/components/content/ui/MarkdownRenderer.tsx
+++ b/apps/desktop/src/components/content/ui/MarkdownRenderer.tsx
@@ -20,6 +20,9 @@ type MdTag =
   | "h1"
   | "h2"
   | "h3"
+  | "h4"
+  | "h5"
+  | "h6"
   | "p"
   | "ul"
   | "ol"
@@ -66,6 +69,10 @@ export function MarkdownRenderer({
     "[&_li>p]:my-0",
     "[&_li>ol]:mt-2",
     "[&_li>ul]:mt-2",
+    // GFM task lists: drop the stray disc bullet and align the checkbox inline.
+    "[&_ul.contains-task-list]:list-none [&_ul.contains-task-list]:pl-0",
+    "[&_li.task-list-item]:pl-0 [&_li.task-list-item]:flex [&_li.task-list-item]:items-start [&_li.task-list-item]:gap-2",
+    "[&_li.task-list-item>input]:mt-1 [&_li.task-list-item>input]:size-3.5 [&_li.task-list-item>input]:shrink-0 [&_li.task-list-item>input]:accent-link-foreground",
     className,
   ].filter(Boolean).join(" ");
 
@@ -75,11 +82,17 @@ export function MarkdownRenderer({
         remarkPlugins={[remarkGfm]}
         components={{
           h1: (props) =>
-            mdHtmlElement("h1", "mb-2 mt-3 text-chat leading-[var(--text-chat--line-height)] font-bold text-foreground", props),
+            mdHtmlElement("h1", "mb-3 mt-6 border-b border-border pb-1.5 text-[24px] font-semibold leading-[1.25] text-foreground first:mt-0", props),
           h2: (props) =>
-            mdHtmlElement("h2", "mb-2 mt-3 text-chat leading-[var(--text-chat--line-height)] font-bold text-foreground", props),
+            mdHtmlElement("h2", "mb-2.5 mt-5 border-b border-border pb-1 text-[20px] font-semibold leading-[1.25] text-foreground", props),
           h3: (props) =>
-            mdHtmlElement("h3", "mb-1 mt-2 text-chat leading-[var(--text-chat--line-height)] font-semibold text-foreground", props),
+            mdHtmlElement("h3", "mb-2 mt-5 text-[17px] font-semibold leading-[1.3] text-foreground", props),
+          h4: (props) =>
+            mdHtmlElement("h4", "mb-2 mt-4 text-[15px] font-semibold leading-[1.3] text-foreground", props),
+          h5: (props) =>
+            mdHtmlElement("h5", "mb-1.5 mt-4 text-[13px] font-semibold uppercase tracking-wide text-muted-foreground", props),
+          h6: (props) =>
+            mdHtmlElement("h6", "mb-1.5 mt-4 text-[12px] font-semibold uppercase tracking-wide text-muted-foreground", props),
           p: (props) =>
             mdHtmlElement("p", "my-2 text-chat leading-[var(--text-chat--line-height)] text-foreground", props),
           ul: (props) =>
--- a/apps/desktop/src/components/content/ui/MarkdownRenderer.tsx
+++ b/apps/desktop/src/components/content/ui/MarkdownRenderer.tsx
@@ -70,9 +70,11 @@ export function MarkdownRenderer({
     "[&_li>ol]:mt-2",
     "[&_li>ul]:mt-2",
     // GFM task lists: drop the stray disc bullet and align the checkbox inline.
+    // Descendant selector (not `>input`) so it also matches loose lists where
+    // the checkbox sits inside a <p>; avoid flex so nested blocks still stack.
     "[&_ul.contains-task-list]:list-none [&_ul.contains-task-list]:pl-0",
-    "[&_li.task-list-item]:pl-0 [&_li.task-list-item]:flex [&_li.task-list-item]:items-start [&_li.task-list-item]:gap-2",
-    "[&_li.task-list-item>input]:mt-1 [&_li.task-list-item>input]:size-3.5 [&_li.task-list-item>input]:shrink-0 [&_li.task-list-item>input]:accent-link-foreground",
+    "[&_li.task-list-item]:pl-0",
+    "[&_li.task-list-item_input]:mr-2 [&_li.task-list-item_input]:size-3.5 [&_li.task-list-item_input]:align-middle [&_li.task-list-item_input]:accent-link-foreground",
     className,
   ].filter(Boolean).join(" ");
 
--- a/apps/desktop/src/components/settings/panes/GeneralPane.tsx
+++ b/apps/desktop/src/components/settings/panes/GeneralPane.tsx
@@ -12,7 +12,7 @@ import { APP_ROUTES } from "@/config/app-routes";
 import { useAvailableEditors } from "@/hooks/access/tauri/shell/use-available-editors";
 import { resolvePreferredOpenTarget } from "@/lib/domain/chat/composer/preference-resolvers";
 import { emitTurnEnd } from "@/lib/infra/events/turn-end-events";
-import type { TurnEndSoundId } from "@/lib/domain/preferences/user/model";
+import type { DefaultNewWorkspaceMode, TurnEndSoundId } from "@/lib/domain/preferences/user/model";
 import { useUserPreferencesStore } from "@/stores/preferences/user-preferences-store";
 
 type SettingsOpenTargetIconId =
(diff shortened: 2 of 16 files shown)


pablonyx:
## Spec-grounded review (3 reviewers: frontend structure, markdown rendering, preferences/shortcuts)

Read against `specs/codebase/structures/frontend/**`, `features/chat-transcript.md`, `features/workspace-files.md`, `features/settings-admin-ia.md`. One blocker, several majors. Findings below with file refs.

### 🔴 Blocker
1. **⌘N preference never persists.** `defaultNewWorkspaceMode` was added to the model, `NEW_USER_DEFAULTS`, `PERSISTED_RECORD_BACKFILL`, and `selectPersistedUserPreferencesSlice`, but **not** to `USER_PREFERENCE_KEYS` in `apps/desktop/src/lib/domain/preferences/user/persisted-keys.ts`. That allow-list filters the on-disk record on read (`pickLegacyUserPreferencesInput`), so the saved value is stripped before migration and always falls back to the backfill (`"worktree"`). Net effect: a user who picks "Local" reverts to "Worktree" on next launch. The value also leaks into `persistedMetadata` as an unknown forward-compat extra. **Fix: add `defaultNewWorkspaceMode` to `USER_PREFERENCE_KEYS`** (sits next to `branchPrefixType`). A load→save→load test would have caught this.

### 🟠 Major
2. **Task-list styling breaks on loose / multi-block lists** (`MarkdownRenderer.tsx`, `MarkdownBody.tsx`). The GFM checkbox `<input>` is only a *direct* child of `<li>` for **tight** lists (the `<p>` is unwrapped). For a loose list or a multi-block task item the DOM is `<li class="task-list-item"><p><input>…</p></li>`, so `[&_li.task-list-item>input]:…` no longer matches (checkbox loses `accent`/`size`/`mt`/`shrink`) and `[&_li.task-list-item]:flex` lays the inner `<p>`/sub-list as flex items in the same row. Fix: use a descendant selector (`[&_li.task-list-item_input]`) and scope the flex to the first line, or knowingly accept tight-list-only.
3. **Duplicated markdown styling → drift** (spec smell). `chat-transcript.md` designates `product-ui/.../MarkdownBody.tsx` as the shared presentational markdown renderer, yet the new h4–h6 scale + task-list block are now copy-pasted into desktop's `MarkdownRenderer.tsx`. They've **already diverged**: h3 `leading-[1.3]` vs `leading-[22px]`; h1/h2 underline only in the file view. Either factor the shared heading/task-list styling into product-ui (or a shared constant) consumed by both, or document the intentional split.
4. **Shared `Spinner` rewrite has cross-app blast radius.** `Spinner` is the canonical `packages/ui` primitive consumed by Desktop **and Web** (+ product-ui/surfaces). This PR redefines its geometry (viewBox 20→24, stroke→filled paths, dropped `overflow-visible`), but the timing change (1.1s→1.4s) lives in Desktop-only `design/css/desktop.css`. Result: Web inherits the new shape but keeps the old speed. Verify the new ring at every consumed size and in light/dark on both clients; consider moving the timing to the shared theme layer.
5. **Plain `Ctrl+N` on Windows/Linux** (`workspace-shortcuts.ts`). `newDefault` has no `nonMacMatch`, so it falls back to `Ctrl+N` — the universal OS/browser "new window" accelerator. Siblings deliberately remap (e.g. `newCloud` → Ctrl+Alt+N). macOS impact is nil today, but the binding ships the risk; set an explicit `nonMacMatch` or confirm intent.

### 🟡 Minor
6. **No enum validation** for `defaultNewWorkspaceMode` in `migration.ts` (siblings like `branchPrefixType` sanitize corrupt values to the default). Once #1 is fixed, a bad persisted value flows straight to the store.
7. **`newDefault` missing from `SHORTCUT_GROUPS`** (`config/shortcuts/groups.ts`) → the primary ⌘N is invisible in the Keyboard Shortcuts settings pane while the three variants are listed.
8. **`accent-link-foreground` is desktop-bundle-only.** `link-foreground` is registered in `@theme` in `desktop.css` but not in the shared `dom.css` Web imports. Harmless today (no Web consumer of `MarkdownBody`), but the accent would no-op if Web ever renders it; consider registering the token in the shared theme.

### ⚪ Nits
9. h3 line-height differs between the two renderers (`leading-[1.3]` vs `leading-[22px]` — ~identical visually, inconsistent style).
10. File-view h2 lacks `first:mt-0` (h1 has it); a `.md` starting with `##` gets a top gap.
11. No tests added for the new pref or shortcut.

### 🟢 Good
- Preference modeled at the right layer (type + `UserPreferences` + defaults + backfill + persisted slice).
- ⌘N handler reads the **live** value via `useUserPreferencesStore.getState()` in a lifecycle hook and reuses existing `newLocal/newWorktreeWorkspace` actions instead of duplicating creation logic.
- Plain ⌘N verified to have **no** macOS native-menu or other KeyN handler collision.
- Settings row composes shared `SettingsCardRow`/`SettingsMenu` and writes through `preferences.set` (no local primitive lookalikes).

**Bottom line:** ship-blocker is #1 (one-line fix). #2 and #3 are the substantive quality issues; #4/#5 are verification/platform risks worth resolving before merge.

pablonyx:
### Review fixes — `9eb1f65fa`

**Fixed:**
- 🔴 **#1 blocker** — added `defaultNewWorkspaceMode` to `USER_PREFERENCE_KEYS` so the ⌘N choice is no longer stripped on read; it now persists across launches. Also stops the orphaned-key leak into `persistedMetadata` (#2 secondary).
- 🟠 **#2 task lists** — switched to a descendant selector (`[&_li.task-list-item_input]`) so the checkbox is styled in loose/multi-block lists too, and dropped the `flex` on `<li>` so nested blocks stack instead of being laid in the checkbox row. Applied to both renderers.
- 🟡 **#6** — `migrateUserPreferences` now validates/sanitizes `defaultNewWorkspaceMode` (mirrors `branchPrefixType`).
- 🟡 **#7** — `newDefault` added to `SHORTCUT_GROUPS`, so ⌘N now shows in the Keyboard Shortcuts pane.

**Deferred (judgment calls, not done here):**
- 🟠 **#3 markdown duplication** — factoring shared heading/task-list styling into product-ui is a larger refactor; left as-is. Worth a follow-up.
- 🟠 **#4 Spinner blast radius** — the shape change is intentional app-wide; the desktop-only timing vs shared markup point needs a design call on whether to move timing to the shared theme. Left for review.
- 🟠 **#5 Ctrl+N on Win/Linux** — app is macOS-focused today; kept plain ⌘N. Can add an explicit `nonMacMatch` if we want to harden non-mac.
- ⚪ nits (#8 token registration, h3 leading, h2 `first:mt-0`, tests) — left.

Note: typecheck wasn't run in the branch worktree (no `node_modules` there), but all edits are additive and type-safe by construction (`ShortcutKey` includes `newDefault`; `keyof UserPreferences` includes the new key).

pablonyx:
## Review round 2 (adversarial + CSS/a11y), against fix commit `9eb1f65fa`

### ✅ Verified fixed (traced end-to-end)
- **Blocker persistence is genuinely complete.** Round-trip confirmed: write (`selectPersistedUserPreferencesSlice` → subscriber → `buildPersistedUserPreferencesRecord`) and read (`USER_PREFERENCE_KEYS` → `pickLegacyUserPreferencesInput` → `migrateUserPreferences`). No zod/other registry needs the key. Legacy records backfill to `"worktree"` correctly.
- **Migration validation does not cause spurious re-persist.** Because `next` seeds from the backfill, a missing/valid value keeps `changed=false`; only a corrupt value flips it. Mirrors `branchPrefixType` exactly.
- **Task-list fix is sound.** Descendant `_input` reaches loose-list checkboxes inside `<p>`, and dropping flex lets nested blocks stack. `accent-link-foreground` compiles (token in `@theme`). `align-middle` is the standard checkbox approach.

### 🟠 New / still-open
1. **A11y — loading state announced to nobody (SidebarIndicators).** The spinner svg is `aria-hidden` and the wrapper only has `data-loading-spinner`; the `iterating`/`queued_prompt` indicator relies on a hover Tooltip, which isn't exposed as an accessible name. Screen readers get nothing for these states. (SidebarUpdatePill and the transcript loading states do this right with `role="status"`/`aria-label`.) Pre-existing, not worsened — but it's the decorative-spinner gap. Suggest `role="status"` + visually-hidden text or `aria-label` at that callsite.
2. **Nested task lists lose indentation.** `[&_ul.contains-task-list]:pl-0` + `[&_li.task-list-item]:pl-0` apply recursively and out-specify the base `ul` `pl-4`, so a task `<ul>` nested inside a task item renders flush with its parent. Worth scoping the `pl-0` to the top level (e.g. only the non-nested `contains-task-list`).
3. **Descendant `_input` slightly over-matches** — a raw `<input>` embedded in a task item via HTML would also get the checkbox styling. Rare in chat markdown; acceptable, noting the tradeoff vs the old `>input`.

### ⚪ Cosmetic
4. **Spinner artwork is a ~270° arc, not a "quarter arc"** (the solid band sweeps 270°; the moving 90° gap is the track showing through). Reads fine as a loader — but my code comment / PR description mislabel it. Worth correcting the comment.
5. **Reduced-motion static frame** is a 270° arc frozen at 22° → looks like a near-complete progress ring rather than a spinner. The old dashed-stroke frame showed clearer gaps. Minor affordance regression for reduced-motion users.
6. **Dangling heading underline** (file view): a `.md` ending on a heading shows a trailing rule under nothing; an h2 directly under an h1 shows two close parallel rules. GitHub behaves the same, so likely acceptable.

### 🧪 Tests
- No tests added. The natural additions: a migration test (invalid `defaultNewWorkspaceMode` → backfills, valid `"local"` preserved with `changed=false`) mirroring the existing `branchPrefixType` test, and a load→save→load round-trip asserting a UI-set `"local"` survives reload (the exact regression the blocker fix resolves).

Both rounds run statically — neither could execute the suite (branch worktree has no `node_modules`).

pablonyx:
### Round-2 fixes — `2a5dfdfc6`

- 🟠 **A11y (#1)** — `SidebarStatusIndicatorView` now wraps the bare glyph in `<span role="img" aria-label={tooltip}>`, so the iterating/queued (and all other) status indicators have an accessible name instead of being announced to nobody.
- 🟠 **Nested task lists (#2)** — scoped the `pl-0` to the top-level list (`[&>ul.contains-task-list]:pl-0`); `list-none` stays descendant, so nested task lists keep their indentation while still losing the disc bullet. Applied to both renderers.
- 🧪 **Tests** — added a `persisted-keys` guard that `defaultNewWorkspaceMode` is a known persisted key (directly prevents the original blocker regression), plus migration coverage (invalid → backfills to `worktree`, valid `local` preserved).
- ⚪ **#4** — corrected the misleading Spinner "quarter arc" comment (it's ~270°).

Left as deliberate (noted earlier): markdown duplication refactor (#3), Spinner timing-location decision (#4 round 1), `Ctrl+N` on Win/Linux (#5), descendant `_input` over-match (rare), reduced-motion frame, dangling heading underline.

As before, tests/tsc weren't executed in the branch worktree (no `node_modules`); edits are type-safe by construction and the new tests mirror existing patterns.

pablonyx:
## Review round 3 — adversarial verify of `2a5dfdfc6` + spinner-consistency sweep

### ✅ Round-2 fixes verified correct (could not prove any wrong)
- **A11y**: `Tooltip` is a hand-rolled primitive (not Radix), so wrapping the glyph in a plain `<span role="img" aria-label>` can't break triggering/ref/events. `indicator.tooltip` is typed `string` on every union variant. Previously the bare glyph had **no** accessible name; now it has one. (Minor, out of scope: it's a static name, not a live-region announcement of state changes.)
- **Nested task lists**: confirmed the top-level `<ul>` is a direct child of the wrapper in both renderers; specificity `(0,2,1) > (0,1,0)` so top-level `pl-0` wins; nested lists keep indentation and still drop the disc via the descendant `list-none`.
- **Tests**: traced all new assertions — they pass deterministically; the `as unknown as` cast typechecks. (Not executed — worktree has no deps.)

**No blocker/major remaining in this PR.** Review has converged.

### 🔭 Separate follow-up (not in this PR's diff)
The new codex-style Spinner is now visually inconsistent with ~8 hand-rolled inline spinners elsewhere (`Loader2`/`RotateCw`/`RefreshCw` + `animate-spin`). Highest-visibility:
- `product-ui/src/chat/composer/CloudChatComposerFooter.tsx:109` and `CloudChatComposerControlParts.tsx:67` (`Loader2` pending indicators)
- `product-ui/src/environments/AddCloudEnvironmentDialog.tsx:223` ("Loading repositories…")
- refresh buttons: `AgentDefaultsPane.tsx:181`, `ArchivedChatsPane.tsx:36`, `WorkspacesSurface.tsx:85`, `TurnDiffFileCard.tsx`, `GitReviewFileRow.tsx:197`

`animate-pulse` skeleton placeholders + activity status-dots are legitimately different — leave as-is. Suggest a separate PR to converge the inline spinners on the shared primitive.

PR #746, opened 2026-06-18
fix(transcript): stable scroll — no per-message jump, snappier composer reflow

pablonyx:
PR 2 of the live-chat scroll/layout cleanup. Two focused, low-risk fixes that reuse mechanisms already proven in the transcript code.

## 1. Glue the pinned snap (commit 1)
The virtualized transcript's pinned snap was a **one-shot** `scrollToBottom` on content-change deps. A new/streamed row enters at the flat `ESTIMATED_TURN_HEIGHT_PX = 360` estimate → the effect snaps once → the virtualizer's ResizeObserver corrects the height a frame later → a **second snap against a different height**. That double-snap is the visible jump on every message and at session start.

Fix: expose the existing snap-until-stable glue loop (`startGlueLoop` → `glueToBottom`) and run it alongside the synchronous snap:
- synchronous `scrollToBottom()` keeps the **no-flash, pre-paint** guarantee,
- `glueToBottom()` re-snaps each frame until measured `scrollHeight` settles.

The glue already bails the instant the user scrolls up (`pinnedRef`), so scroll-up-during-stream is unchanged. The full (non-virtualized) list already re-sticks via a content `ResizeObserver`, so it needs no change.

**Fixes:** per-message shift (#9), estimate half of start-of-session shift (#8), and the **common pinned case** of the thinking→command nudge (#10).

## 2. Track dock inset on the next frame (commit 2)
The composer textarea autosizes synchronously per line, but the transcript bottom inset was recomputed behind a **90ms settle timer** (`CHAT_DOCK_RESIZE_SETTLE_MS`), so the transcript lagged the textarea when adding/removing lines. Coalesce resize bursts onto the next animation frame only — the autosize is discrete, so one rAF is enough and immediate. No loop risk (inset is a pure function of dock geometry; transcript reflow doesn't resize the dock).

**Fixes:** slow composer reflow (#5); tightens start-of-session inset settle (#8).

## Explicitly deferred
The **unpinned-history** half of #10 (the anchor-restore `getOffsetForIndex` estimate branch). A blanket measured-delta swap would mis-shift below-the-anchor inserts — measured-delta is only correct for above-the-viewport changes (which the code already routes through `startAboveChangeCompensation`). Needs a more careful design; tracked as a follow-up.

## Testing
- `useTranscriptStickToBottom.test.tsx` — added an explicit `glueToBottom` test (re-snaps as a row measures taller a frame later); 14 tests pass.
- `VirtualizedTranscriptRowList` + `FullTranscriptRowList` tests pass (19 total in the transcript suite).
- `tsc --noEmit` clean for both `@proliferate/product-ui` and `apps/desktop`.
- Visual pass in `/playground` (ChatPlaygroundPage) recommended before merge — start of session, per-message streaming, composer grow/shrink.

### DIFF
--- a/apps/desktop/src/hooks/chat/ui/use-chat-dock-inset.ts
+++ b/apps/desktop/src/hooks/chat/ui/use-chat-dock-inset.ts
@@ -5,8 +5,6 @@ import {
   computeChatSurfaceBottomInsetPx,
 } from "@/config/chat-layout";
 
-const CHAT_DOCK_RESIZE_SETTLE_MS = 90;
-
 export function useChatDockInset() {
   const dockRef = useRef<HTMLDivElement>(null);
   const [metrics, setMetrics] = useState({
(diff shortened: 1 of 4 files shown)


pablonyx:

PR #757, opened 2026-06-18
fix(desktop): GitHub sign-in checking flash + sidebar spinner alignment

pablonyx:
## What
Two small desktop UI cleanups.

### 1. GitHub sign-in "checking" flash during the auth transition
The persistent `AuthShell` morphs a loading skeleton into the GitHub button in place. The availability probe was gated `enabled: mode === "auth"`, so it only *started* the instant the morph began — its first-load pending window overlapped the transition and surfaced *"Checking whether GitHub sign-in is configured for this environment."* (and a "Checking…" button label) for a beat before resolving.

- **Warm the probe** whenever `AuthShell` is mounted (loading phase included), so the answer is normally resolved *before* the button appears. The query already self-guards on `controlPlaneReachable` and caches (`staleTime: 15s`).
- **Stop rendering the transient "checking" copy** in `AuthScreenLayout` + `LoginScreen` — the button just stays disabled until the probe resolves, so the transition is skeleton → sign-in button with no "checking" text. Drops the now-unused `checking` label.

### 2. Sidebar workspace status spinner not vertically centered
The non-action status indicator wrapped its glyph in a plain inline `<span>`, so the `inline-flex` Spinner baseline-aligned and sat high relative to the row. Flex-center the wrapper so the spinner (and every status glyph) is vertically centered.

## Scope / risk
Presentational only; no change to actual sign-in behavior or status logic. `LoginScreen` is playground-only but shares the pattern, so it's fixed for consistency. Verified the dropped `checking` copy has no remaining consumers.

## Verify
- Auth: loading → sign-in morph shows no "checking" flash (replayable via `/playground` → AuthOnboarding).
- Sidebar: a workspace with an iterating/queued status shows its spinner vertically centered in the row.

### DIFF
--- a/apps/desktop/src/components/auth/AuthScreenLayout.tsx
+++ b/apps/desktop/src/components/auth/AuthScreenLayout.tsx
@@ -110,11 +110,7 @@ export function AuthScreenLayout({
               className="h-11 w-full"
             >
               {!submitting && <GitHub className="h-4 w-4 shrink-0" />}
-              {submitting
-                ? AUTH_LOGIN_LABELS.waiting
-                : githubSignInChecking
-                  ? AUTH_LOGIN_LABELS.checking
-                  : AUTH_LOGIN_LABELS.signIn}
+              {submitting ? AUTH_LOGIN_LABELS.waiting : AUTH_LOGIN_LABELS.signIn}
               {!submitting && <ArrowRight className="h-4 w-4" />}
             </Button>
           </div>
@@ -125,7 +121,7 @@ export function AuthScreenLayout({
           <div className="absolute inset-x-0 top-full mt-3 text-center">
             {showAuth && error
               ? <p className="text-sm text-destructive">{error}</p>
-              : showAuth && (githubSignInChecking || !githubSignInAvailable)
+              : showAuth && !githubSignInChecking && !githubSignInAvailable
                 ? (
                   <p className="text-sm text-muted-foreground">
                     {githubSignInUnavailableDescription}
--- a/apps/desktop/src/components/auth/AuthShell.tsx
+++ b/apps/desktop/src/components/auth/AuthShell.tsx
@@ -20,7 +20,7 @@ export function AuthShell({ mode, markComplete, onMarkResolved }: AuthShellProps
     signInAvailable,
     signInChecking,
     signInUnavailableDescription,
-  } = useGitHubSignIn({ enabled: mode === "auth" });
+  } = useGitHubSignIn();
 
   return (
     <AuthScreenLayout
--- a/apps/desktop/src/components/auth/LoginScreen.tsx
+++ b/apps/desktop/src/components/auth/LoginScreen.tsx
@@ -73,15 +73,11 @@ export function LoginScreen({
             className="h-11 w-full"
           >
             {!submitting && <GitHub className="h-4 w-4 shrink-0" />}
-            {submitting
-              ? AUTH_LOGIN_LABELS.waiting
-              : githubSignInChecking
-                ? AUTH_LOGIN_LABELS.checking
-                : AUTH_LOGIN_LABELS.signIn}
+            {submitting ? AUTH_LOGIN_LABELS.waiting : AUTH_LOGIN_LABELS.signIn}
             {!submitting && <ArrowRight className="h-4 w-4" />}
           </Button>
 
-          {(githubSignInChecking || !githubSignInAvailable) && (
+          {!githubSignInChecking && !githubSignInAvailable && (
             <p className="text-sm text-muted-foreground">
               {githubSignInUnavailableDescription}
             </p>
--- a/apps/desktop/src/components/workspace/shell/sidebar/SidebarIndicators.tsx
+++ b/apps/desktop/src/components/workspace/shell/sidebar/SidebarIndicators.tsx
@@ -77,7 +77,7 @@ export function SidebarStatusIndicatorView({
           {glyph}
         </IconButton>
       ) : (
-        <span role="img" aria-label={indicator.tooltip}>{glyph}</span>
+        <span role="img" aria-label={indicator.tooltip} className="inline-flex items-center justify-center">{glyph}</span>
       )}
     </Tooltip>
   );
--- a/apps/desktop/src/copy/auth/auth-copy.ts
+++ b/apps/desktop/src/copy/auth/auth-copy.ts
@@ -16,7 +16,6 @@ export const AUTH_LOGIN_LABELS = {
   detail: "Start by signing in with GitHub.",
   detailWithLocalPrefix: "Continue with GitHub, or",
   signIn: "Continue with GitHub",
-  checking: "Checking GitHub sign-in...",
   waiting: "Waiting for GitHub...",
   continueLocally: "Start locally",
   continueLocallyInline: "start locally",
--- a/apps/desktop/src/copy/capabilities/capability-copy.ts
+++ b/apps/desktop/src/copy/capabilities/capability-copy.ts
@@ -22,8 +22,6 @@ export const CAPABILITY_COPY = {
     "This environment can reach the control plane, but desktop GitHub sign-in is not configured.",
   githubLocalDescription:
     "GitHub sign-in is unavailable while the control plane is unreachable.",
-  githubAuthCheckingDescription:
-    "Checking whether GitHub sign-in is configured for this environment.",
   githubAuthUnavailableDescription:
     "GitHub sign-in is not configured for this environment.",
   githubSignedInUnavailableDescription:
--- a/apps/desktop/src/hooks/auth/workflows/use-github-sign-in.ts
+++ b/apps/desktop/src/hooks/auth/workflows/use-github-sign-in.ts
@@ -16,26 +16,24 @@ export interface UseGitHubSignInResult {
 }
 
 // Owns GitHub sign-in form state and submit callback. Does not own auth availability access.
-// `enabled` can gate the availability poll for on-demand callers. The auth shell leaves it
-// enabled during the loading phase too, so the availability answer is warm before the sign-in
-// button morphs in — otherwise the first-load pending window flashes a "checking…" state
-// during the loading -> auth transition.
-export function useGitHubSignIn(options?: { enabled?: boolean }): UseGitHubSignInResult {
+// The availability probe runs whenever this hook is mounted — the persistent auth shell mounts
+// it during the loading phase too, so the answer is warm before the sign-in button morphs in,
+// otherwise the first-load pending window flashes a "checking…" state during the loading -> auth
+// transition. The query self-guards on control-plane reachability.
+export function useGitHubSignIn(): UseGitHubSignInResult {
   const { signInWithGitHub } = useAuthActions();
   const { cloudEnabled } = useAppCapabilities();
   const {
     data: githubDesktopAuthAvailable,
     isPending: githubDesktopAuthAvailabilityPending,
-  } = useGitHubDesktopAuthAvailability({ enabled: options?.enabled ?? true });
+  } = useGitHubDesktopAuthAvailability();
   const [submitting, setSubmitting] = useState(false);
   const [error, setError] = useState<string | null>(null);
 
   const signInChecking = cloudEnabled && githubDesktopAuthAvailabilityPending;
   const signInAvailable = cloudEnabled && githubDesktopAuthAvailable?.enabled === true;
-  const signInUnavailableDescription = signInChecking
-    ? CAPABILITY_COPY.githubAuthCheckingDescription
-    : cloudEnabled
-      ? CAPABILITY_COPY.githubAuthUnavailableDescription
+  const signInUnavailableDescription = cloudEnabled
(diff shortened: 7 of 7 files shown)


pablonyx:
## 🧭 Frontend conventions & spec review (subagent)

Lens: frontend conventions + spec compliance (`specs/codebase/structures/frontend/`). Net: this is a clean, correctly-layered presentational fix. No blockers. A few minor/nit observations below.

### Spinner centering fix

**[ok]** — `SidebarIndicators.tsx:80` — The wrapper change `className="inline-flex items-center justify-center"` on the `<span role="img">` is the **right layer**. styling.md *UI Primitives First* says "primitives own color, border, radius, typography, focus, hover, disabled, and loading states. Callsite classes may handle layout, spacing, and sizing." Vertical centering of a glyph in its slot is layout → belongs at the callsite, not in the `Spinner` primitive. And `Spinner` (`apps/packages/ui/src/primitives/Spinner.tsx:8`) already self-centers its SVG (`inline-flex shrink-0` + `size-full`), so pushing centering into the primitive would be wrong — the bug was the *plain inline wrapper span*, whose inline baseline (with descender slack) sat the `inline-flex` glyph high. `inline-flex` on the wrapper collapses that inline gap. Correct diagnosis, minimal fix.

**[ok]** — Not the shared row's job either. `ProductSidebarWorkspaceRow` already centers the slot (`ProductSidebarRepositories.tsx:117` → `flex w-4 shrink-0 items-center justify-center`) and the `Tooltip` anchor also centers (`SidebarIndicators.tsx:65`). The residual offset was desktop-local to this wrapper span, so fixing it in `apps/desktop/src/components/.../SidebarIndicators.tsx` (not in `product-ui`) respects ownership — `product-ui` is "props in, callbacks out" shared presentation and shouldn't absorb a desktop glyph-wrapper quirk.

**[ok — covers both callsites]** — The fix lives in the shared `SidebarStatusIndicatorView`, which is the funnel for **both** `WorkspaceItem.tsx:166` and `CoworkThreadRow.tsx:40`. So cowork thread rows get the same centering for free — good, no second edit needed.

**[minor]** — `SidebarIndicators.tsx:80` — As the PR notes, this wrapper is shared by every non-action status glyph (error `CircleAlert`, waiting `Clock`, `needs_review` dot, iterating/queued `Spinner`). Re-centering them all is **desirable** (same inline-baseline bug), and there's no size/color change, so behavior risk is low. `SidebarStatusGlyph` is covered by `SidebarIndicators.test.tsx` but that test asserts glyph *identity*, not wrapper layout, so nothing breaks. Worth a one-line manual confirm that the `needs_review` 2px dot (which is `block`, line 46) still looks centered, since it's the most sensitive to the change. The PR's verify steps only call out the spinner.

**[nit]** — `inline-flex items-center justify-center` now appears 3x down the same centering chain (row slot, Tooltip anchor, this span). Redundant but harmless; the wrapper-level fix is the most surgical option, so leaving it is fine.

### Auth "checking" flash + probe warming

**[ok]** — `AuthShell.tsx:23` — Dropping `{ enabled: mode === "auth" }` so the probe warms during the loading phase is fine convention-wise. I found **no spec** gating availability polls or forbidding pre-auth probes. The underlying query (`hooks/access/cloud/auth/use-github-auth-availability.ts:17`) self-guards on `controlPlaneReachable` and caches (`staleTime`/`refetchInterval` 15s, `retry: 1`), so warming early adds no meaningful cost and matches the access-hook ownership model (retry/poll policy lives in the access hook, per README "Access hooks own … retry policy").

**[minor]** — `use-github-sign-in.ts:23,29` — The `enabled` param is now **unused by every caller** (`AuthShell` is the only consumer and dropped it). The Hard Rules say "Keep public hook/component APIs stable unless the task explicitly changes callsites" and "Delete dead code when replacing an implementation." This is a borderline call: the updated comment reframes `enabled` as "can gate the availability poll for on-demand callers," i.e. kept speculatively. Per "Do not create … speculative abstractions," I'd lean toward **deleting the `enabled` option** (and the `options?.enabled ?? true` passthrough) until a real on-demand caller exists — `useGitHubDesktopAuthAvailability` still accepts its own `enabled`, so the capability isn't lost. If you keep it, the current comment is the right call. Not a blocker.

**[ok]** — `AuthScreenLayout.tsx:113,124` + `LoginScreen.tsx:76,80` — Removing the transient "checking" label/message and leaving the button disabled until the probe resolves is consistent. Note the gate inverted from `(checking || !available)` to `!checking && !available`: the unavailable message now only shows once the probe has *resolved* unavailable, which is the intended no-flash behavior. The button stays disabled while `githubSignInChecking` (unchanged disabled logic at `AuthScreenLayout.tsx:103-108` / `LoginScreen.tsx:72`), so there's no window where it's clickable mid-probe. Logic is sound.

### Copy

**[ok]** — `copy/auth/auth-copy.ts:18` — Removing the now-unused `checking` label is correct per copy.md ("authored user-facing copy") + Hard Rule "Delete dead code." Verified no remaining consumers of `AUTH_LOGIN_LABELS.checking`.

**[minor]** — Orphaned copy left behind: `CAPABILITY_COPY.githubAuthCheckingDescription` (`copy/capabilities/capability-copy.ts:25`) is now referenced **only** at `use-github-sign-in.ts:36`, which builds `signInUnavailableDescription` for the `signInChecking` branch — but with the probe warmed, that "checking" description string is no longer rendered anywhere (both screens stopped rendering during checking). It's not strictly dead (the ternary still reads it), so leaving it is defensible, but it's the same "checking" copy the PR set out to retire. Consider whether `signInUnavailableDescription` even needs the `signInChecking` branch anymore, or whether that branch + the copy constant can also go. Minor cleanup, not blocking.

### Spec drift

**[ok]** — No spec documents the auth loading→sign-in transition or the sidebar status-glyph rendering, so nothing in `specs/` goes stale. These are presentational-only changes (confirmed: no status-logic or sign-in-behavior change), which the README explicitly favors ("Preserve current UI and behavior unless an explicit behavior change is requested"). No new spec is warranted for a glyph-alignment + flash fix.

---
**Summary:** Correctly layered, low-risk. The only things I'd actually action: (1) consider deleting the now-unused `enabled` param on `useGitHubSignIn` rather than keeping it speculatively, and (2) the `githubAuthCheckingDescription` copy + its `signInChecking` branch are now effectively never rendered — fold them in if you want the "checking" path fully retired. Both minor.

pablonyx:
## 🔬 Correctness & UX review (subagent)

Lens: correctness, UX regressions, edge cases. Net: **the change is sound and does what it claims.** Probe warming is correctly gated, the rewritten conditions are equivalent-or-better, and the spinner fix is the right one. A few dead-code / copy-staleness items and one UX tradeoff worth a conscious sign-off.

### 1. Probe warming side effects — ✅ no real problem
- `useGitHubDesktopAuthAvailability` is still `enabled: controlPlaneReachable && (options?.enabled ?? true)` (`use-github-auth-availability.ts:17`). With `AuthShell` now leaving `enabled` defaulted, the probe fires during loading **only when the control plane is reachable** — the self-guard is intact. No extra calls when CP is unreachable.
- When the session resolves to authenticated, `AuthShell` unmounts (`AuthGate.tsx:82` returns `<Outlet/>` once `fadeState === null`). With no other mounted consumer the query goes inactive; `refetchInterval` stops. No longer-running interval than before.
- No race that strands `signInChecking === true` at the morph: `signInChecking = cloudEnabled && pending`. If CP health is slow (`controlPlaneReachable` false), `cloudEnabled` is false → `signInChecking` is `false` and the button shows the *local* message, not a hang. The only pending window is CP-reachable + probe-in-flight, which warming is designed to shrink.

### 2. UX regression from dropping the checking copy — ⚠️ minor, conscious tradeoff
There is exactly one state that now renders a **disabled button with zero explanation**: `controlPlaneReachable === true && availability query still pending`. Pre-PR this showed "Checking whether GitHub sign-in is configured…"; post-PR it shows nothing until the probe resolves. For a fast probe this is invisible (the point of the PR). For a *reachable-but-slow* control plane, the user stares at a disabled button with no hint for the probe's duration. Acceptable given warming, but it is a deliberate downgrade for the slow-CP case — worth an explicit ack.
- Good news: no **indefinite** silent-disabled state. On probe **error** (`retry: 1` exhausted) TanStack flips `isPending → false`, so `signInChecking` becomes false and the `!checking && !available` message branch fires → shows `githubAuthUnavailableDescription`. Same as before. `refetchInterval: 15s` keeps it from sticking.
- Caveat (pre-existing, not introduced here): a probe *fetch failure* surfaces as "GitHub sign-in is not configured for this environment." — misleading (network failure ≠ misconfig), but unchanged by this PR.

### 3. Condition correctness — ✅ equivalent-or-better
`!githubSignInChecking && !githubSignInAvailable` (`AuthScreenLayout.tsx:124`, `LoginScreen.tsx:80`) vs old `githubSignInChecking || !githubSignInAvailable`. Walking the three reachable states:
- CP unreachable: `checking=false, available=false` → **shows** local message (both old and new). ✅
- CP reachable + pending: old **showed** checking msg; new **hides** → intended.
- CP reachable + resolved-unavailable (incl. error): `checking=false, available=false` → **shows** unavailable message (both). ✅
No state now renders nothing where it should show the unavailable/local message. The unavailable path still resolves correctly. Button `disabled` logic (`!showAuth || busy || checking || !available`) is unchanged and still disables during the pending window — correct.

### 4. `enabled` param is now vestigial — **[minor]** `use-github-sign-in.ts:23`
`AuthShell` was the **only** caller that ever passed `{ enabled }`. I grepped all 6 callers (`AccountPane`, `CloudSignInRequiredPane`, `use-login-page`, `use-workspace-mobility-footer-flow`, `AuthShell`) — after this PR **none** pass it (the other `enabled:` hits in those files belong to `useComputeTargetOptions` / `useWorkspaceMobilityGitPrepWorkflow` / `passwordCredential`). So `options?: { enabled?: boolean }` on `useGitHubSignIn` is now dead surface, and the doc comment ("can gate the availability poll for on-demand callers") describes a capability nothing uses. Fix: drop the param (and the `options?.enabled ?? true` plumbing) for a net simplification, or keep it but the comment overstates current usage. Not a blocker — just leftover.

### 5. Stale "checking" copy still in the hook — **[nit]** `use-github-sign-in.ts:35-36` + `capability-copy.ts:25`
`signInUnavailableDescription` still computes `CAPABILITY_COPY.githubAuthCheckingDescription` when `signInChecking`, but post-PR **no surface renders the description while checking** (both screens gate on `!checking`; the settings panes use `AUTH_ACCOUNT_LABELS.checkingSignIn` for their button label and don't render this description during checking). So `githubAuthCheckingDescription` and the checking branch of `signInUnavailableDescription` are now effectively dead for display. The removal of `AUTH_LOGIN_LABELS.checking` is correctly verified — **zero remaining consumers** (confirmed by grep). Consider removing the now-orphaned `githubAuthCheckingDescription` too, for parity with the stated "drop unused checking copy" goal.

### 6. Spinner wrapper change — ✅ correct fix, other glyphs safe
`<span role="img" aria-label className="inline-flex items-center justify-center">` (`SidebarIndicators.tsx:80`). Root cause is real: the slot (`ProductSidebarRepositories.tsx:117`, `flex w-4 items-center justify-center`) and the `Tooltip` wrapper (`inline-flex items-center justify-center`) both center, but the **inner plain `<span>`** established an inline context, so the `Spinner` (itself `inline-flex`, an inline-level box) baseline-aligned and rode high. Making the span a flex container centers the Spinner on the cross axis instead. Correct and minimal.
- Other status kinds are unaffected: `CircleAlert`/`Clock` (`size-3` SVGs) and the `needs_review` dot (`block size-2`) are single fixed-size children — as the sole flex child they sit centered exactly as before (their intrinsic box already matched the line box). No shift/break.
- `role="img"` + `aria-label` on a flex span is still valid ARIA (`img` role + accessible name); `display` doesn't affect role semantics. The action branch (`IconButton`) was never affected (buttons aren't baseline-sensitive here). Per `styling.md` "Callsite Styling" this is allowed (layout/composition utilities at the callsite) — compliant.

### 6b. Tests — adequate, one gap
- `SidebarIndicators.test.tsx` asserts glyph *type* per kind but not the wrapper class — fine, the fix is purely presentational alignment and hard to assert meaningfully in jsdom (no layout). No new coverage strictly needed.
- `AuthGate.test.tsx` covers the loading→login→app reveal and that the action is disabled while loading / live in auth, but **does not** assert the new "no checking label/message" behavior (the `AuthShell` is mocked out, so the probe-pending path isn't exercised). Optional: a small `AuthScreenLayout`/`LoginScreen` test with `githubSignInChecking` true asserting the button label stays "Continue with GitHub" and no description renders would lock in the regression this PR fixes.
- Note: I could not execute these tests in the worktree — both fail at import-resolution (`@proliferate/ui/icons`, `@proliferate/product-ui/*`) because the shared packages aren't built/linked in this checkout. That's an env/build artifact of the worktree, **not** a PR defect.

### Verdict
Ship-worthy. Address #4 (drop the now-dead `enabled` param) and optionally #5 (orphaned checking description) as cleanup; consciously accept the #2 slow-CP "disabled-with-no-text" tradeoff. Spinner fix is correct.

PR #758, opened 2026-06-18
fix(transcript): no session-switch flicker + collapse resume backlog (#2, #6)

pablonyx:
Stack 1/9 on #746. **#2**: stop remounting the list on session switch (drop key={rowListKey}; resetForSession in place, keeping the virtualizer measurement cache); gate the blank-viewport fallback until the first measurement pass. **#6**: resume glue terminates on stability (not a 12-frame cap) + one synchronous virtualizer.measure() so a backgrounded backlog lands in one jump. tsc clean; product-ui transcript suite green. Verified: acceptance met (high).

### DIFF
--- a/apps/packages/product-ui/src/chat/transcript/VirtualizedTranscriptRowList.tsx
+++ b/apps/packages/product-ui/src/chat/transcript/VirtualizedTranscriptRowList.tsx
@@ -70,6 +70,11 @@ export function VirtualizedTranscriptRowList({
   // resume-measure callback reads it through a ref (set once the virtualizer
   // exists) to force one synchronous measure pass on tab/window resume.
   const virtualizerRef = useRef<{ measure: () => void } | null>(null);
+  // measure() clears the entire itemSizeCache (full re-measure from estimates),
+  // so this momentarily resets scrollHeight to estimates before the resume glue
+  // loop follows the re-measure. That cost is acceptable only because this is
+  // resume-only (one tab/window re-show); do NOT widen its trigger to per-event
+  // paths or it reintroduces the estimate->measure churn during streaming.
   const resumeMeasure = useCallback(() => {
     virtualizerRef.current?.measure();
   }, []);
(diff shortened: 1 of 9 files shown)


apps/packages/product-ui/src/chat/transcript/VirtualizedTranscriptRowList.tsx
```diff
@@ -189,14 +202,32 @@ export function VirtualizedTranscriptRowList({
...
+    if (rows.length === 0) {
+      return;
+    }
+    const frame = window.requestAnimationFrame(() => {
+      virtualizer.measure();
+      setMeasurementReady(true);
+    });
+    return () => { window.cancelAnimationFrame(frame); };
+  }, [activeSessionId, rows.length, selectedWorkspaceId, virtualizer]);
+
```
pablonyx, line 239: MAJOR — this effect defeats the PR's own core premise. `virtualizer.measure()` (tanstack v3.17.1) does `itemSizeCache.clear()` — it throws away ALL measured row heights and forces a full re-measure from estimates. But this effect's dep array is `[activeSessionId, rows.length, selectedWorkspaceId, virtualizer]`, so it fires on every `rows.length` change — i.e. on every streamed-row append and every history prepend, not just on session/workspace switch. So during active streaming you wipe the whole measurement cache once per appended row, reintroducing exactly the estimate→measure churn the PR removed the `key` to avoid (and working against chat-transcript.md's "long transcripts must stay virtualized" / avoid per-event invalidation guidance).

The stated purpose here is only to flip `measurementReady` one frame after a session change — that needs `setMeasurementReady(true)` inside the rAF, nothing else. The explicit `measure()` is both unnecessary (the virtualizer already measures via `useAnimationFrameWithResizeObserver: true`) and harmful on the streaming path. Drop the `virtualizer.measure()` call here, and ideally gate the whole effect so it only runs on session/workspace change (not on every `rows.length` change).

pablonyx:
**Spec critique (round 1):** The session-switch-resume approach is sound and well-aligned with chat-transcript.md's stick-to-bottom invariants: dropping the `key={rowListKey}` remount in favor of `resetForSession` in-place is the right call (keeps the virtualizer's measurement cache, avoids the re-pin-from-360px-estimate blank flash), the `measurementReady` gate on the blank-viewport fallback correctly suppresses the estimate-only first frame, and splitting `GLUE_MAX_FRAMES` into `GLUE_ABOVE_MAX_FRAMES` (bounded above-change comp) vs `GLUE_RESUME_MAX_FRAMES` (stability-terminated resume) is principled. Tests + tsc pass; the resume-glue test meaningfully guards the past-12-frame behavior.

One major issue: the new `useEffect` that flips `measurementReady` (VirtualizedTranscriptRowList.tsx ~L221-230) calls `virtualizer.measure()` with `rows.length` in its dep array. `measure()` in tanstack v3.17.1 clears the ENTIRE `itemSizeCache`, and `rows.length` changes on every streamed-row append — so this wipes the measurement cache once per append during streaming, reintroducing the very estimate→measure churn this PR removed the `key` to avoid. The `measure()` call is also unnecessary for the stated purpose (the gate only needs `setMeasurementReady(true)`; the virtualizer already measures via `useAnimationFrameWithResizeObserver`). Drop the explicit `measure()` and ideally scope the effect to session/workspace change only. See inline comment. Everything else looks good.

pablonyx:

PR #759, opened 2026-06-18
fix(transcript): hold unpinned read position on at/below composition changes (#10)

pablonyx:
Stack 2/9. The unpinned anchor-restore used estimate-based getOffsetForIndex; now distinguishes above-viewport changes (rows inserted/removed above the read position -> measured-delta) from at/below changes (the anchor's index is unchanged -> force-measure the anchor row and hold the exact pixel offset) so a turn growing at or below the read position no longer nudges it. New pure helper resolveUnpinnedAnchorRestore + tests (pure picker + an integration guard that the same-index path force-measures the anchor node and re-anchors). Verified: acceptance met (high).

### DIFF
--- a/apps/packages/product-ui/src/chat/transcript/VirtualizedTranscriptRowList.tsx
+++ b/apps/packages/product-ui/src/chat/transcript/VirtualizedTranscriptRowList.tsx
@@ -23,6 +23,7 @@ import {
   type TranscriptRowListBaseProps,
 } from "./TranscriptRowListShared";
 import { useAboveChangeCompensation } from "./useAboveChangeCompensation";
+import { resolveUnpinnedAnchorRestore } from "./unpinned-anchor-restore";
 import { useTranscriptStickToBottom } from "./useTranscriptStickToBottom";
 import { VirtualTranscriptViewport } from "./VirtualTranscriptViewport";
 import { useTranscriptVirtualizerBlankFallback } from "./useTranscriptVirtualizerBlankFallback";
@@ -283,12 +284,21 @@ export function VirtualizedTranscriptRowList({
     notifyProgrammaticScroll,
   });
 
-  // While unpinned, a completing turn can split one row into completed-history +
-  // content — a new, unmeasured row inserted ABOVE the anchored row. The
-  // getOffsetForIndex + offsetWithinRowPx restore lands against the 360px
-  // estimate and bumps when measurement corrects. When rows were inserted above
-  // the anchor, hold the user's position with the measured scrollHeight delta;
-  // pure shifts and below-the-viewport appends keep the offset reposition / no-op.
+  // While unpinned, a completing turn can split/grow rows around the anchored
+  // (first-visible) row. getOffsetForIndex builds on the 360px estimate for
+  // unmeasured neighbours, so the naive offset restore lands wrong and nudges
+  // the read position. Distinguish the three change shapes relative to the
+  // anchor and pick the restore that holds it at the same viewport y:
+  //   - anchor moved DOWN  (nextIndex > rowIndex): rows inserted ABOVE the
+  //     viewport — hold with the measured scrollHeight delta (estimate-immune).
+  //   - anchor moved UP    (nextIndex < rowIndex): rows removed ABOVE the
+  //     viewport — same measured-delta compensation applies.
+  //   - anchor index UNCHANGED (nextIndex === rowIndex): nothing changed above
+  //     the anchor, so getOffsetForIndex(nextIndex) is the anchor's real top.
+  //     A change strictly BELOW the viewport leaves scrollTop correct (hold it,
+  //     no move); a same-index height change of the anchor row itself needs the
+  //     measured offset — force-measure the anchor row first so the offset is
+  //     real, not the estimate, then restore offset + intra-row position.
   useLayoutEffect(() => {
     const anchor = pendingAnchorRef.current;
     pendingAnchorRef.current = null;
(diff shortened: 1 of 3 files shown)


apps/packages/product-ui/src/chat/transcript/unpinned-anchor-restore.test.ts
```diff
@@ -0,0 +1,32 @@
...
+      resolveUnpinnedAnchorRestore({ capturedRowIndex: 2, nextRowIndex: 2 }),
+    ).toEqual({ kind: "measured-offset" });
+  });
+
+  it("uses measured-offset for a same-index height change of the anchor row", () => {
+    expect(
+      resolveUnpinnedAnchorRestore({ capturedRowIndex: 0, nextRowIndex: 0 }),
+    ).toEqual({ kind: "measured-offset" });
+  });
+});
```
pablonyx, line 32: Test adequacy: these 4 cases only exercise the trivial pure picker (`nextRowIndex !== capturedRowIndex`). The substantive new behavior of this PR — force-measuring the anchor node and the `getOffsetForIndex` + intra-row offset restore, plus the moved-up→measured-delta routing — is entirely untested. The header comment in unpinned-anchor-restore.ts correctly notes jsdom does no layout, so a true unit test is hard, but there's no integration-level guard either (VirtualizedTranscriptRowList.test.tsx still only covers mount + scroll-affordance). Consider at least asserting the layout effect calls `virtualizer.measureElement` on the resolved anchor node in the same-index case, with a mocked virtualizer.

apps/packages/product-ui/src/chat/transcript/VirtualizedTranscriptRowList.tsx
```diff
@@ -298,11 +308,33 @@ export function VirtualizedTranscriptRowList({
...
+    // intra-row offset). When the change is purely below the viewport this
+    // reproduces the captured scrollTop exactly (a no-op hold); when the anchor
+    // row itself grew/shrank it holds the same visible content.
+    const viewport = scrollRef.current;
+    const anchorNode = viewport?.querySelector<HTMLElement>(
+      `[data-index="${nextIndex}"]`,
+    );
+    if (anchorNode) {
+      virtualizer.measureElement(anchorNode);
+    }
```
pablonyx, line 345: Spec-sync gap. specs/codebase/features/chat-transcript.md lines 207-212 ("Stick-to-bottom engine") still describes the unpinned restore as a single case: "a completing turn that splits one row ... would bump the viewport ... The virtualized list holds the anchored content with the measured scrollHeight delta in a stability-gated loop." This PR generalizes that into a three-way decision (above-shift up OR down → measured-delta; same-index → force-measured offset + intra-row hold). The spec is now incomplete/slightly inaccurate vs the code. Sibling commits in this stack sync the spec in-unit (e.g. dec43d35b "spec sync", 17efa5e6f docs(specs)); recommend updating that paragraph here too.

apps/packages/product-ui/src/chat/transcript/unpinned-anchor-restore.test.ts
```diff
@@ -0,0 +1,32 @@
+import { describe, expect, it } from "vitest";
+import { resolveUnpinnedAnchorRestore } from "./unpinned-anchor-restore";
+
+describe("resolveUnpinnedAnchorRestore", () => {
+  it("uses measured-delta when rows are inserted above the anchor (anchor moves down)", () => {
+    // thinking->command split inserts a completed-history row above the anchored
+    // first-visible row, so its index increases.
+    expect(
```
pablonyx, line 8: Framing nit (not a code issue): the PR title/body headline this fix with the "thinking→command transition no longer nudges history" example — but that's the anchor-moves-DOWN (rows inserted above) case, which already routed to measured-delta in the parent branch (`nextIndex > anchor.rowIndex`). The genuinely new behavior of this unit is (a) the same-index force-measure + offset restore for at/below changes, and (b) routing anchor-moves-UP (removals above) to measured-delta too. The title "hold ... on at/below composition changes" is accurate; the thinking→command lead-in slightly over-claims novelty. Worth aligning the body example to an at/below scenario.

pablonyx:
**Spec critique (round 1):** Solid, low-risk PR. The core change is correct: replacing the parent's two-way restore (`nextIndex > rowIndex` → measured-delta, else estimate-based `getOffsetForIndex`) with a clean three-way decision via the new pure `resolveUnpinnedAnchorRestore` picker — above-viewport shift (index changed, up OR down) → estimate-immune measured scrollHeight delta; same-index (at/below) → force-measure the anchor node then hold (measured top + captured intra-row offset). I verified: the force-measure targets the same div that carries both `data-index` and the virtualizer's measure ref (VirtualTranscriptViewport L71/89, VirtualTranscriptRow L24-26), so `virtualizer.measureElement(anchorNode)` is well-formed; the same-index/below no-op claim holds (offset = scrollTop reproduction); the moved-up→measured-delta extension is symmetric and the glue loop handles negative deltas (useAboveChangeCompensation L42). Typecheck clean, new + existing transcript tests pass.\n\nNon-blocking findings (all minor/nit): (1) **Spec-sync** — chat-transcript.md L207-212 still documents only the single split→measured-delta case and is now incomplete vs the three-way behavior; stack convention is to sync the spec in-unit. (2) **Test adequacy** — only the trivial pure picker is covered; the substantive force-measure/offset-restore + measured-delta routing has no integration guard (jsdom-no-layout is acknowledged, but a mocked-virtualizer assertion is feasible). (3) **Nit** — PR body leads with the thinking→command (anchor-moves-down) example, which already worked in the parent; the new value is the at/below + moved-up paths. No correctness regressions, no scope leakage (3 files, all in transcript area).

pablonyx:

PR #760, opened 2026-06-18
fix(activity): tie 'Iterating' to a live turn + force thinking in reasoning mode (#15, #3)

pablonyx:
Stack 3/9. **#15**: 'iterating' now keys off an in-progress turn with a live item, not the coarse transcript.isStreaming (queued/awaiting no longer read as iterating). **#3**: force the thinking indicator whenever reasoning mode is active. +24 activity tests. Verified: acceptance met (high).

### DIFF
--- a/apps/desktop/src/lib/domain/chat/session-controls/session-controls.ts
+++ b/apps/desktop/src/lib/domain/chat/session-controls/session-controls.ts
@@ -142,6 +142,30 @@ export function currentValueLabel(
   return current?.label ?? currentValue ?? null;
 }
 
+/**
+ * Whether the reasoning/thinking control is currently on. Handles both the
+ * two-value on/off toggle and multi-value (e.g. none/low/medium/high) shapes:
+ * reasoning is active when a value is selected that is not a disabled-token
+ * ("off"/"none"/"disabled"/"false") value. Used to keep the trailing thinking
+ * indicator visible for the whole turn when reasoning mode is enabled.
+ */
+export function isReasoningControlActive(
+  control: NormalizedSessionControl | null | undefined,
+): boolean {
+  if (!control || control.currentValue == null) {
+    return false;
+  }
+  const current = control.values.find((value) => value.value === control.currentValue);
+  const tokens = current ? valueTokens(current) : valueTokens({
+    value: control.currentValue,
+    label: "",
+  });
+  if (tokens.length === 0) {
+    return true;
+  }
+  return !tokens.some((token) => isDisabledToken(token) || token === "none");
+}
+
 export function resolveToggleState(
   control: NormalizedSessionControl,
   currentValueOverride?: string | null,
--- a/apps/packages/product-domain/src/sessions/activity-status.ts
+++ b/apps/packages/product-domain/src/sessions/activity-status.ts
@@ -86,6 +86,22 @@ export function isSessionEffectivelyStreaming(
   ) {
     return false;
   }
+  // `isStreaming` is the coarse turn-open flag and can stick when a
+  // `turn_ended` is missed (e.g. a backgrounded session). The execution
+  // summary is the authoritative turn phase, so a stale stream flag must not
+  // resurrect "running"/"iterating" once the summary has settled to a
+  // non-running phase. Only "running" (a turn actually in progress) and
+  // "starting" keep bare streaming live; "idle"/"errored"/"closed" and a
+  // settled "awaiting_interaction" are not iterating.
+  const summaryPhase = slot.executionSummary?.phase;
+  if (
+    summaryPhase === "idle"
+    || summaryPhase === "errored"
+    || summaryPhase === "closed"
+    || summaryPhase === "awaiting_interaction"
+  ) {
+    return false;
+  }
   return true;
 }
 
--- a/apps/packages/product-ui/src/chat/transcript/useLatestTranscriptLiveStatus.ts
+++ b/apps/packages/product-ui/src/chat/transcript/useLatestTranscriptLiveStatus.ts
@@ -40,6 +40,7 @@ export function useLatestTranscriptLiveStatus({
   virtualRows,
   outboxStartedAtByPromptId,
   sessionViewState,
+  reasoningActive = false,
   renderTurnTrailingStatus,
 }: {
   latestTurnId: string | null;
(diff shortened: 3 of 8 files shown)


apps/packages/product-ui/src/chat/transcript/useLatestTranscriptLiveStatus.ts
```diff
@@ -99,6 +101,17 @@ export function useLatestTranscriptLiveStatus({
...
+  // turn, but transient thoughts are filtered out and the coarse stream flag
+  // can briefly drop (or prose lands) before the turn ends — which would hide
+  // the trailing indicator. Force it on for the whole in-progress turn so the
+  // thinking affordance never disappears mid-turn when reasoning is enabled.
+  const shouldForceReasoningLiveStatus = reasoningActive
+    && !!latestTurn
+    && latestTurnInProgress
+    && !latestLiveExplorationBlock
+    && !latestLiveWorkBlock
+    && !latestTurnHasActiveToolWork;
```
pablonyx, line 114: Major (spec tension): `shouldForceReasoningLiveStatus` deliberately omits the `shouldAllowTurnTrailingStatus` prose-landed guard that the non-reasoning path uses (line 96). chat-transcript.md:239 states the invariant: "Once the last top-level turn item is prose with text, the prose itself is the placeholder and a separate spinner is not needed." In reasoning mode, when the agent's final action is prose, the full prose lands while `completedAt`/`turn_ended` is still pending — so `latestTurnInProgress` is true and a thinking indicator paints *below the finished prose* until the turn ends. The commit message says this bypass is intentional, but for the common 'final chunk is prose' case it produces a spurious post-prose spinner. Consider also gating on `!lastTopLevelItemIsAssistantProseWithText` (or `lastTopLevelItemIsStreamingAssistantProse`) so the force only applies while prose is still streaming or absent.

apps/packages/product-ui/src/chat/transcript/useLatestTranscriptLiveStatus.ts
```diff
@@ -120,14 +133,20 @@ export function useLatestTranscriptLiveStatus({
...
-    && (showDelayedLatestLiveStatus || shouldShowImmediateOutboxLiveStatus)
-      ? renderTurnTrailingStatus?.({
-          startedAt: latestTurnTiming?.startedAt ?? latestTurn.startedAt,
-          sessionViewState,
-          transientStatusText: latestTransientText,
-        }) ?? null
-      : null;
+  const shouldShowLatestLiveStatus = showDelayedLatestLiveStatus
+    || shouldShowImmediateOutboxLiveStatus
+    || shouldForceReasoningLiveStatus;
```
pablonyx, line 138: Minor (test gap): the core UI behavior of this unit (#3 — forcing the trailing thinking indicator on for the whole reasoning turn, pinning sessionViewState to 'working', and bypassing the isStreaming/grace/prose gates) has no test. `isReasoningControlActive` is unit-tested, but `useLatestTranscriptLiveStatus` is not — sibling hooks in this dir (useTranscriptStickToBottom, useTranscriptVirtualizerBlankFallback) are tested, so a renderHook test is feasible. Add cases: (a) reasoningActive forces status even when transcript.isStreaming is false / sessionViewState !== 'working'; (b) it does NOT force once latestTurn.completedAt is set; (c) the prose-landed interaction from the comment above.

apps/packages/product-domain/src/sessions/activity-status.ts
```diff
@@ -86,6 +86,22 @@ export function isSessionEffectivelyStreaming(
...
+  // resurrect "running"/"iterating" once the summary has settled to a
+  // non-running phase. Only "running" (a turn actually in progress) and
+  // "starting" keep bare streaming live; "idle"/"errored"/"closed" and a
+  // settled "awaiting_interaction" are not iterating.
+  const summaryPhase = slot.executionSummary?.phase;
+  if (
+    summaryPhase === "idle"
+    || summaryPhase === "errored"
+    || summaryPhase === "closed"
+    || summaryPhase === "awaiting_interaction"
```
pablonyx, line 101: Nit (redundancy): now that this guard always returns false for `awaiting_interaction`, the `effectivelyStreaming` term in `resolveSessionStatus` (lines 56-62) and the ternary in `resolveSessionExecutionPhase` line 119 (`isSessionEffectivelyStreaming(slot) ? 'running' : 'idle'`) are partially dead for the awaiting_interaction branch — they can no longer yield 'running' there. Not a bug, but consider simplifying so the now-unreachable path doesn't mislead a future reader. Behavior matches the Rust model (execution_summary.rs treats AwaitingInteraction as a settled, non-running phase), so the guard itself is correct.

apps/desktop/src/lib/domain/chat/session-controls/session-controls.ts
```diff
@@ -142,6 +142,30 @@ export function currentValueLabel(
...
+  }
+  const current = control.values.find((value) => value.value === control.currentValue);
+  const tokens = current ? valueTokens(current) : valueTokens({
+    value: control.currentValue,
+    label: "",
+  });
+  if (tokens.length === 0) {
+    return true;
+  }
+  return !tokens.some((token) => isDisabledToken(token) || token === "none");
```
pablonyx, line 166: Nit (edge-case): `valueTokens` folds the *label* into the token set (`value + ' ' + label`), so a non-off reasoning value whose label happens to contain a disabled token (e.g. label 'Auto (off when idle)') would be mis-read as inactive. Current catalog reasoning values are simple ('none'/'low'/'medium'/'high'), so this is latent, but matching on `control.currentValue` only (not the label) would be safer for an on/off semantic check. Also: `currentValue === ''` (empty string) slips past the `== null` guard and the empty-tokens fallback returns true (active) — harmless given real catalogs, but worth a guard if you touch this.

pablonyx:
**Spec critique (round 1):** Two cohesive fixes in one unit — (#15) gating `isSessionEffectivelyStreaming` on the authoritative execution-summary phase so a stuck `isStreaming` flag no longer reads as iterating, and (#3) forcing the trailing thinking indicator on for the whole turn in reasoning mode. The activity-status change is correct and matches the Rust model (execution_summary.rs treats AwaitingInteraction/Idle/Errored/Closed as settled, non-running phases); good targeted tests (stuck-stream, settled-awaiting, still-running) and they pass. Scope is clean — stack is well-formed (parent is ancestor), no leakage from other units. One thing to address: the reasoning-force path deliberately bypasses the `shouldAllowTurnTrailingStatus` prose-landed guard, so when the agent's final action is prose, a thinking indicator can paint *below* the finished prose during the pre-`turn_ended` window — this contradicts the chat-transcript.md:239 invariant; recommend also gating on the streaming-prose check (major). Plus: the `useLatestTranscriptLiveStatus` reasoning behavior (the core of #3) has no test despite testable sibling hooks (minor); two nits — now-dead `effectivelyStreaming` branches for awaiting_interaction, and label-folding in `isReasoningControlActive` that could false-negative on contrived labels. No blockers.

pablonyx:

PR #761, opened 2026-06-18
fix(chat): 'You stopped after Ns' + close stranded tools on cancel (#7, #10)

pablonyx:
Stack 4/9. On cancel, close in-progress tool_calls in the SDK reducer (no tool left 'running'); add a 'cancelled' branch to resolveTurnTrailingStatus rendering a right-aligned muted 'You stopped after {N}s'; model live Thinking as a stable keyed row. Verified: acceptance met (high).

### DIFF
--- a/anyharness/sdk/src/reducer/transcript.ts
+++ b/anyharness/sdk/src/reducer/transcript.ts
@@ -152,6 +152,9 @@ export function reduceEvent(
     case "turn_ended": {
       const turn = ensureMutableTurn(s, turnId, ts);
       closeStreamingItems(s);
+      if (evt.stopReason === "cancelled") {
+        closeStrandedToolItems(s, turnId, ts, envelope.seq);
+      }
       turn.completedAt = ts;
       turn.stopReason = evt.stopReason;
       turn.fileBadges = collectFileBadges(s, turnId);
@@ -832,6 +835,34 @@ function closeStreamingPointer(s: TranscriptState, itemId: string): void {
 function closeStreamingItems(s: TranscriptState): void {
   closeAssistantPointer(s);
   closeThoughtPointer(s);
+}
+
+/**
+ * After a cancelled turn the agent never emits `item_completed` for any tool
+ * call that was still running, so without this the row stays visually "running"
+ * forever. Mark every in-progress tool call in the cancelled turn as failed so
+ * it settles to an "Interrupted" affordance instead.
+ */
+function closeStrandedToolItems(
+  s: TranscriptState,
+  turnId: string,
+  ts: string,
+  seq: number,
+): void {
+  const turn = s.turnsById[turnId];
+  if (!turn) return;
+  for (const itemId of turn.itemOrder) {
+    const item = s.itemsById[itemId];
+    if (item?.kind !== "tool_call" || item.status !== "in_progress") {
+      continue;
+    }
+    const nextItem = cloneKnownTranscriptItem(item);
+    nextItem.status = "failed";
+    nextItem.completedAt = ts;
+    nextItem.completedSeq = seq;
+    nextItem.lastUpdatedSeq = Math.max(nextItem.lastUpdatedSeq, seq);
+    setItem(s, itemId, nextItem);
+  }
 }
 
 function closeAssistantPointer(s: TranscriptState): void {
--- a/apps/desktop/src/components/workspace/chat/transcript/TranscriptTurnRow.tsx
+++ b/apps/desktop/src/components/workspace/chat/transcript/TranscriptTurnRow.tsx
@@ -111,8 +111,17 @@ export function TranscriptTurnRow({
     transcript,
     outboxStartedAtByPromptId,
   );
+  // A cancelled turn settles to a completed row, so neither the live status nor
+  // the per-row resolver paints anything — surface the muted "You stopped"
+  // trailing label instead, keyed off the authoritative turn stop reason.
+  const cancelledElapsedSeconds = resolveCancelledTurnElapsedSeconds(
+    turn,
+    turnTiming.startedAt,
+  );
   const trailingStatus = !row.isLastTurnRow
     ? null
+    : cancelledElapsedSeconds !== null
+    ? resolveTurnTrailingStatus(turnTiming.startedAt, sessionViewState, null, cancelledElapsedSeconds)
     : isLatestTurn
     ? latestLiveStatus
     : shouldAllowTurnTrailingStatus({
@@ -259,6 +268,21 @@ function formatUndoError(error: unknown): string {
     return error.message;
   }
   return "Could not undo last turn file changes.";
+}
+
+function resolveCancelledTurnElapsedSeconds(
+  turn: TurnRecord,
+  startedAt: string,
+): number | null {
+  if (turn.stopReason !== "cancelled" || !turn.completedAt) {
+    return null;
+  }
+  const startedMs = Date.parse(startedAt);
+  const completedMs = Date.parse(turn.completedAt);
+  if (Number.isNaN(startedMs) || Number.isNaN(completedMs)) {
+    return null;
+  }
+  return Math.max(0, Math.round((completedMs - startedMs) / 1000));
 }
 
 function resolveTurnTrailingStatusForRow({
--- a/apps/desktop/src/hooks/sessions/workflows/use-session-cancel-actions.ts
+++ b/apps/desktop/src/hooks/sessions/workflows/use-session-cancel-actions.ts
@@ -6,7 +6,6 @@ import {
 } from "@/lib/access/anyharness/session-runtime";
 import {
   getSessionRecord,
-  patchSessionRecord,
 } from "@/stores/sessions/session-records";
 import { useSessionSelectionStore } from "@/stores/sessions/session-selection-store";
 import { useToastStore } from "@/stores/toast/toast-store";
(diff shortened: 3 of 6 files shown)


apps/desktop/src/components/workspace/chat/transcript/TranscriptTurnRow.tsx
```diff
@@ -111,8 +111,17 @@ export function TranscriptTurnRow({
...
     outboxStartedAtByPromptId,
   );
+  // A cancelled turn settles to a completed row, so neither the live status nor
+  // the per-row resolver paints anything — surface the muted "You stopped"
+  // trailing label instead, keyed off the authoritative turn stop reason.
+  const cancelledElapsedSeconds = resolveCancelledTurnElapsedSeconds(
+    turn,
+    turnTiming.startedAt,
+  );
   const trailingStatus = !row.isLastTurnRow
```
pablonyx, line 121: The cancelled branch is gated only by `row.isLastTurnRow`, not by `isLatestTurn`. Because `isLastTurnRow` is per-turn (one row per turn, see `transcript-row-model.ts` `buildRowsForTurnPresentation`), this paints "You stopped after Ns" on the last row of *every* cancelled turn, including historical ones scrolled up after the user sent later prompts. Pre-PR those turns rendered nothing (`shouldAllowTurnTrailingStatus` returns false once `completedAt` is set). This is arguably accurate historical truth, but it is a behavior change beyond "the turn the user just stopped" — please confirm it is intended, and add a test for the non-latest cancelled-turn case.

apps/desktop/src/components/workspace/chat/transcript/TranscriptTurnRow.tsx
```diff
@@ -261,6 +270,21 @@ function formatUndoError(error: unknown): string {
...
+): number | null {
+  if (turn.stopReason !== "cancelled" || !turn.completedAt) {
+    return null;
+  }
+  const startedMs = Date.parse(startedAt);
+  const completedMs = Date.parse(turn.completedAt);
+  if (Number.isNaN(startedMs) || Number.isNaN(completedMs)) {
+    return null;
+  }
+  return Math.max(0, Math.round((completedMs - startedMs) / 1000));
```
pablonyx, line 285: `resolveCancelledTurnElapsedSeconds` has no direct test. The added desktop test only exercises the presentational `resolveTurnTrailingStatus`, not this helper, so the load-bearing logic here (the `stopReason !== "cancelled"` / missing-`completedAt` gate, the `Number.isNaN` guard, and the `Math.max(0, round(...))` flooring) is uncovered. Worth a small unit test, e.g. sub-500ms cancel rounds to `You stopped after 0s` — confirm that wording is acceptable.

apps/desktop/src/hooks/sessions/workflows/use-session-cancel-actions.ts
```diff
@@ -33,7 +32,10 @@ export function useSessionCancelActions() {
     try {
       const { materializedSessionId, workspaceId } = await getSessionClientAndWorkspace(sessionId);
       await cancelSessionMutation.mutateAsync({ workspaceId, sessionId: materializedSessionId });
-      patchSessionRecord(sessionId, { status: "idle" });
+      // Don't optimistically flip to "idle": that races the authoritative
+      // turn_ended event (which carries the "cancelled" stop reason and closes
+      // the streaming/tool items). Let the reducer-driven state settle so the
+      // "You stopped after Ns" affordance and tool teardown stay consistent.
```
pablonyx, line 38: Dropping the optimistic `patchSessionRecord(sessionId, { status: "idle" })` is correct — `stream-patch.ts` already sets `status: "idle"` on `turn_ended` (line ~140), so the optimistic flip was redundant and racy. One UX consequence: the stop affordance now stays in the working state until the cancel `turn_ended` round-trips (the mutation has no optimistic UI). That window is normally tiny, but if a cancel ever fails to produce a `turn_ended` the button could appear stuck. The `catch {}` swallows the failure silently with no toast — consider surfacing a failure path, or confirm `turn_ended` is guaranteed on cancel.

pablonyx:
**Spec critique (round 1):** Solid, well-scoped PR — meets its stated goal (surface "You stopped after Ns" on cancel + settle stranded in-progress tool calls). Verified against `specs/codebase/features/chat-transcript.md`: the reducer change preserves structural sharing (`closeStrandedToolItems` clones via `cloneKnownTranscriptItem` before mutating and `setItem` rebuilds `itemsById`), so the immutability invariant (spec L17-18) holds; `"failed"` is a valid `TranscriptItemStatus` and the new fields match `applyPayload` semantics. The optional 4th arg on `resolveTurnTrailingStatus` defaults to `null`, so `MessageList`/`useLatestTranscriptLiveStatus` callers are unaffected. Dropping the optimistic `idle` flip in `use-session-cancel-actions` is correct (the `turn_ended` handler in `stream-patch.ts` already sets `status: "idle"`). No scope leakage — no file overlap with the parent unit (`fix/activity-status-truth` only touches `activity-status.ts`). Reducer + chrome tests pass (38 / 4). Non-blocking follow-ups: (1) the cancelled label is gated on `row.isLastTurnRow` only, so it renders on *every* cancelled turn including historical ones — confirm intended + test the non-latest case; (2) `resolveCancelledTurnElapsedSeconds` has no direct unit test (gate/NaN/flooring uncovered; sub-500ms → "0s"); (3) cancel failure is swallowed silently with no toast and no optimistic UI, so a missing `turn_ended` could leave the stop button visually stuck. None are blockers.

apps/desktop/src/components/workspace/chat/transcript/TranscriptTurnRow.tsx
```diff
@@ -261,6 +270,21 @@ function formatUndoError(error: unknown): string {
   return "Could not undo last turn file changes.";
 }
 
+function resolveCancelledTurnElapsedSeconds(
```
pablonyx, line 273: Test gap (minor): \`resolveCancelledTurnElapsedSeconds\` is the actual new logic — stopReason gating, the \`!turn.completedAt\` guard, the \`Date.parse\`/\`NaN\` guards, and the \`Math.max(0, Math.round(...))\` rounding — but it is not exported and has no unit test. The only added component test exercises the presentational \`resolveTurnTrailingStatus\`. Worth exporting this and adding a couple of cases (non-cancelled turn → null, missing/invalid timestamps → null, sub-second cancel → 0).

anyharness/sdk/src/reducer/transcript.ts
```diff
@@ -834,6 +837,34 @@ function closeStreamingItems(s: TranscriptState): void {
...
+ * After a cancelled turn the agent never emits `item_completed` for any tool
+ * call that was still running, so without this the row stays visually "running"
+ * forever. Mark every in-progress tool call in the cancelled turn as failed so
+ * it settles to an "Interrupted" affordance instead.
+ */
+function closeStrandedToolItems(
+  s: TranscriptState,
+  turnId: string,
+  ts: string,
+  seq: number,
```
pablonyx, line 850: Nit (comment accuracy): the doc-comment says these settle to an "Interrupted" affordance, but desktop tool rows render \`status === "failed"\` with the generic "failed" tone — there is no "Interrupted" label here. The literal "Interrupted" string lives in a different mechanism (\`product-domain/.../cloud/transcript-view-interactions.ts\`, later-progress detection). Behavior is correct; just reword to "settles to a failed/terminal state" so the comment is not misleading.

apps/desktop/src/components/workspace/chat/transcript/TranscriptTurnRow.tsx
```diff
@@ -111,8 +111,17 @@ export function TranscriptTurnRow({
...
+  // the per-row resolver paints anything — surface the muted "You stopped"
+  // trailing label instead, keyed off the authoritative turn stop reason.
+  const cancelledElapsedSeconds = resolveCancelledTurnElapsedSeconds(
+    turn,
+    turnTiming.startedAt,
+  );
   const trailingStatus = !row.isLastTurnRow
     ? null
+    : cancelledElapsedSeconds !== null
+    ? resolveTurnTrailingStatus(turnTiming.startedAt, sessionViewState, null, cancelledElapsedSeconds)
```
pablonyx, line 124: Nit (accuracy): elapsed is computed from \`turnTiming.startedAt\`, which \`resolveTurnPromptTiming\` resolves to the outbox \`createdAt\` (client submit time, before server-side turn start) when an outbox entry still exists, only falling back to \`turn.startedAt\` otherwise. So "You stopped after Ns" can over-report by the queue/delivery latency in the (rare) window where the outbox entry survives until cancel. Sub-second cancels also render "You stopped after 0s". Consider keying the elapsed off \`turn.startedAt\` directly for a server-truth duration.

pablonyx:
**Spec critique (round 1):** Solid, tightly-scoped unit. The reducer change correctly settles stranded in-progress tool calls to `failed` on a cancelled `turn_ended` (covers nested/child tool calls too since all items pass through `turn.itemOrder`), preserves structural sharing (`cloneKnownTranscriptItem` + `setItem`), and follows the existing `completedAt`/`completedSeq`/`lastUpdatedSeq` conventions. Dropping the optimistic `patchSessionRecord({status:"idle"})` is the right call and consistent with the stacked parent (`fix/activity-status-truth`): view state is derived from `executionSummary` + `transcript.isStreaming` via `resolveSessionViewState`/`isSessionEffectivelyStreaming`, not the record status, and the cancel mutations `onSuccess` invalidates the session queries — so even a lost SSE `turn_ended` is reconciled by the refetch. Diff is clean (6 on-topic files, no other-unit leakage). Findings are all nits/minor: (1) the new `resolveCancelledTurnElapsedSeconds` helper — where the real branching lives — is unexported and untested (only the presentational `resolveTurnTrailingStatus` got a test); (2) the reducer doc-comment says "Interrupted affordance" but `failed` rows render the generic "failed" tone, not an "Interrupted" label (that string is a different cloud mechanism); (3) elapsed seconds key off `turnTiming.startedAt`, which can be the outbox client-submit time, so "You stopped after Ns" can slightly over-report, and sub-second cancels show "0s". None block merge.

pablonyx:

PR #762, opened 2026-06-18
fix(codex): agent-aware current-mode for plan/default (#4)

pablonyx:
Stack 5/9. Codex has two controls (collaboration_mode default/plan + mode permissions). Made current-mode/label/persistence prefer collaborationMode across ALL paths: session-intent dispatch, SDK reducer, chat-input label, **and** the stream-flush patch (stream-patch.ts) + hydration (session-records.ts) — the last two were caught by adversarial review as same-bug sites that clobbered the fix on the live path. 124 session tests green.

### DIFF
--- a/apps/desktop/src/hooks/sessions/lifecycle/session-intent-config-dispatch.ts
+++ b/apps/desktop/src/hooks/sessions/lifecycle/session-intent-config-dispatch.ts
@@ -122,14 +122,17 @@ export async function dispatchConfigIntent(
         patchSessionRecord(intent.clientSessionId, nextPatch);
       }
       if (response.applyState === "applied" && intent.persistDefaultPreference) {
-        // Persist against whichever mode-like control the intent actually
-        // targeted: collaborationMode (default/plan) or mode (permissions).
-        // The preference guard requires this rawConfigId to match intent.configId.
+        // Only the permission `mode` control persists into the create-session
+        // mode_id store (defaultSessionModeByAgentKind). collaboration_mode
+        // (default/plan) is a live control whose default lives in a separate
+        // store (defaultLiveSessionControlValuesByAgentKind), so it must NOT
+        // pass this guard — otherwise toggling Plan/Default would write
+        // "plan"/"default" as an invalid permission mode_id.
         const normalizedControls = effectiveLiveConfig?.normalizedControls;
         const intentModeRawConfigId =
-          [normalizedControls?.collaborationMode, normalizedControls?.mode]
-            .find((control) => control?.rawConfigId === intent.configId)
-            ?.rawConfigId ?? null;
+          normalizedControls?.mode?.rawConfigId === intent.configId
+            ? normalizedControls.mode.rawConfigId
+            : null;
         persistDefaultSessionModePreference({
           agentKind: response.session.agentKind ?? latestSlot.agentKind,
           liveConfigRawConfigId: intentModeRawConfigId,
--- a/apps/desktop/src/hooks/sessions/lifecycle/session-intent-config-dispatch.ts
+++ b/apps/desktop/src/hooks/sessions/lifecycle/session-intent-config-dispatch.ts
@@ -91,6 +91,12 @@ export async function dispatchConfigIntent(
           ?? null,
       } as const;
       if (effectiveLiveConfig) {
+        // Codex exposes two mode-like controls: collaborationMode (default/plan)
+        // and mode (permissions). Prefer collaborationMode when present so the
+        // current mode tracks plan<->default, mirroring useChatSessionControls.
+        const modeControl =
+          effectiveLiveConfig.normalizedControls.collaborationMode
+          ?? effectiveLiveConfig.normalizedControls.mode;
         patchSessionRecord(intent.clientSessionId, {
           ...nextPatch,
           liveConfig: effectiveLiveConfig,
@@ -100,14 +106,14 @@ export async function dispatchConfigIntent(
             ?? latestSlot.modelId
             ?? null,
           modeId:
-            effectiveLiveConfig.normalizedControls.mode?.currentValue
+            modeControl?.currentValue
             ?? response.session.modeId
             ?? latestSlot.modeId
             ?? null,
           transcript: {
             ...latestSlot.transcript,
             currentModeId:
-              effectiveLiveConfig.normalizedControls.mode?.currentValue
+              modeControl?.currentValue
               ?? response.session.modeId
               ?? latestSlot.transcript.currentModeId,
           },
@@ -116,9 +122,17 @@ export async function dispatchConfigIntent(
         patchSessionRecord(intent.clientSessionId, nextPatch);
       }
       if (response.applyState === "applied" && intent.persistDefaultPreference) {
+        // Persist against whichever mode-like control the intent actually
+        // targeted: collaborationMode (default/plan) or mode (permissions).
+        // The preference guard requires this rawConfigId to match intent.configId.
+        const normalizedControls = effectiveLiveConfig?.normalizedControls;
+        const intentModeRawConfigId =
+          [normalizedControls?.collaborationMode, normalizedControls?.mode]
+            .find((control) => control?.rawConfigId === intent.configId)
+            ?.rawConfigId ?? null;
         persistDefaultSessionModePreference({
           agentKind: response.session.agentKind ?? latestSlot.agentKind,
-          liveConfigRawConfigId: effectiveLiveConfig?.normalizedControls.mode?.rawConfigId ?? null,
+          liveConfigRawConfigId: intentModeRawConfigId,
           rawConfigId: intent.configId,
           modeId: getAuthoritativeConfigValue(effectiveLiveConfig, intent.configId) ?? intent.value,
           workspaceSurface: deps.getWorkspaceSurface(workspaceId),
--- a/apps/desktop/src/lib/domain/chat/composer/chat-input.test.ts
+++ b/apps/desktop/src/lib/domain/chat/composer/chat-input.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from "vitest";
 import {
   resolveChatDraftWorkspaceId,
   resolveChatInputAvailability,
+  resolveCurrentModeLabel,
 } from "./chat-input";
 
 describe("resolveChatDraftWorkspaceId", () => {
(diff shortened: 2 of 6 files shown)


apps/desktop/src/hooks/sessions/lifecycle/session-intent-config-dispatch.ts
```diff
@@ -116,9 +122,17 @@ export async function dispatchConfigIntent(
...
+        // The preference guard requires this rawConfigId to match intent.configId.
+        const normalizedControls = effectiveLiveConfig?.normalizedControls;
+        const intentModeRawConfigId =
+          [normalizedControls?.collaborationMode, normalizedControls?.mode]
+            .find((control) => control?.rawConfigId === intent.configId)
+            ?.rawConfigId ?? null;
         persistDefaultSessionModePreference({
           agentKind: response.session.agentKind ?? latestSlot.agentKind,
-          liveConfigRawConfigId: effectiveLiveConfig?.normalizedControls.mode?.rawConfigId ?? null,
+          liveConfigRawConfigId: intentModeRawConfigId,
```
pablonyx, line 138: BLOCKER — wrong preference store for collaboration_mode. By picking `intentModeRawConfigId` from whichever of `[collaborationMode, mode]` matches `intent.configId`, the guard in `persistDefaultSessionModePreference` (session-mode-preferences.ts:27 — `liveConfigRawConfigId !== rawConfigId`) now passes for `collaboration_mode` too. So toggling Plan/Default on an active **standard-surface** Codex session (the composer mode selector calls `setActiveSessionConfigOption` with `persistDefaultPreference !== false`) writes `"plan"`/`"default"` into `defaultSessionModeByAgentKind["codex"]`.

That store is the create-session `mode_id` store (use-session-creation-actions.ts:145 → resolveSessionCreationModeId → `mode_id`), validated against the catalog `mode` control (read-only/auto/full-access per specs/codebase/structures/anyharness/src/agent-mode-matrix.md:100-136). The codebase deliberately routes collaboration_mode to a *separate* store, `defaultLiveSessionControlValuesByAgentKind` (use-chat-launch-control-actions.ts:38-52: only `rawConfigId === "mode"` goes to `defaultSessionModeByAgentKind`). Consequences: (1) the persisted permission-mode default is clobbered; (2) next Codex create-session sends `mode_id="plan"` → invalid; (3) the home launcher resolves it against the permission `mode` options and silently drops it (use-home-next-mode-selection.ts:38-44). Pre-PR the guard correctly rejected collaboration_mode here. Persistence of collaboration_mode (if wanted) must target `defaultLiveSessionControlValuesByAgentKind`, not this guard.

apps/desktop/src/lib/domain/chat/composer/chat-input.test.ts
```diff
@@ -18,6 +19,52 @@ describe("resolveChatDraftWorkspaceId", () => {
   });
 });
 
+describe("resolveCurrentModeLabel", () => {
```
pablonyx, line 22: MAJOR (test gap) — the only new tests cover `resolveCurrentModeLabel` (the safe label path). The riskiest change in this PR is the persistence rewrite in session-intent-config-dispatch.ts (`intentModeRawConfigId`), and it has **zero** coverage: there is no `session-intent-config-dispatch.test.ts`, and `session-mode-preferences.test.ts` only exercises `rawConfigId: "mode"` (never `collaboration_mode`). A test asserting "toggling collaboration_mode on a standard Codex session does NOT mutate `defaultSessionModeByAgentKind`" would have caught the store-routing bug flagged above. Please add it.

pablonyx:
**Spec critique (round 1):** The read-side of this PR is correct and well-aligned: making `currentModeId`/`modeId`/label prefer `collaborationMode` over `mode` across the reducer (transcript.ts), stream-flush patch (stream-patch.ts), hydration (session-records.ts), and the composer label (chat-input.ts) faithfully mirrors the canonical precedence already in `useChatSessionControls` (modeControl = collaboration_mode ?? mode) and matches the two-control Codex model in specs/codebase/structures/anyharness/src/agent-mode-matrix.md. Those four sites are consistent and the label tests are good.

The write-side change is a **blocker**. The new `intentModeRawConfigId` in session-intent-config-dispatch.ts lets the `liveConfigRawConfigId === rawConfigId` guard pass for `collaboration_mode`, so toggling Plan/Default on an active standard-surface Codex session now persists `"plan"`/`"default"` into `defaultSessionModeByAgentKind`. But that store is the create-session `mode_id` store (validated against the permission `mode` control: read-only/auto/full-access), and the codebase deliberately routes collaboration_mode to a *separate* store, `defaultLiveSessionControlValuesByAgentKind` (see use-chat-launch-control-actions.ts:38-52). Net effect: clobbers the permission-mode default, emits an invalid `mode_id` on the next Codex session, and silently drops in the home launcher. Pre-PR the guard correctly rejected this. Plus a **test gap**: the risky persistence path has zero coverage (no dispatch test; session-mode-preferences.test.ts only covers `rawConfigId:"mode"`). Recommendation: drop collaboration_mode from this persistence guard (keep `mode` only), or route it to `defaultLiveSessionControlValuesByAgentKind`; add a regression test. See the two inline comments.

apps/desktop/src/hooks/sessions/lifecycle/session-intent-config-dispatch.ts
pablonyx, line 138: This collaborationMode-aware persistence fix is correct here, but the same persistence guard mismatch still exists on the **reconciliation** path, which this PR did not touch. `persistReconciledModePreferences` (use-session-summary-actions.ts:42-59) calls `persistDefaultSessionModePreference` per reconciled change, but both callers pass `normalizedControls.mode?.rawConfigId` as `liveConfigRawConfigId`:

- `use-session-summary-actions.ts:146`
- `session-stream-side-effects.ts:130`

The guard at `session-mode-preferences.ts:27` requires `liveConfigRawConfigId === change.rawConfigId`. When a `collaboration_mode` (plan/default) change reconciles, `change.rawConfigId` is the collaborationMode id but the passed value is the permission `mode` id → guard rejects → the plan/default default preference is silently dropped. This is the identical bug class fixed here, so the PR body claim of fixing persistence "across ALL paths" is not met. Both callers should compute the rawConfigId the same way (match collaborationMode first, then mode).

apps/desktop/src/lib/domain/sessions/stream-patch.ts
```diff
@@ -69,13 +69,17 @@ export function buildSessionStreamPatch({
...
+    const nextModeId =
+      event.liveConfig.normalizedControls.collaborationMode?.currentValue
+      ?? event.liveConfig.normalizedControls.mode?.currentValue;
+    patch.modeId = nextModeId ?? slot.modeId;
     patch.transcript = {
       ...nextTranscript,
-      currentModeId:
-        event.liveConfig.normalizedControls.mode?.currentValue
-        ?? nextTranscript.currentModeId,
+      currentModeId: nextModeId ?? nextTranscript.currentModeId,
```
pablonyx, line 82: Good fix on the stream-flush path. But the parallel **summary-apply** modeId derivation in `buildSessionSlotPatchFromSummary` (lib/domain/sessions/summary.ts:34-37) was not updated and still reads only `normalizedControls.mode?.currentValue`. That builder feeds `applySessionSummary` (use-session-summary-actions.ts:71) which writes `patch.modeId` to the directory (line 111) and `patch.transcript.currentModeId` (lines 86-88). So on every authoritative summary flush, a Codex session in plan/default will have its badge/currentModeId overwritten with the **permission** mode, clobbering the collaborationMode value this PR set elsewhere — the exact regression the PR notes it was guarding against on the stream path. summary.ts should prefer `collaborationMode?.currentValue ?? mode?.currentValue` to match this change, stream-patch, session-records, and the reducer.

apps/desktop/src/lib/domain/chat/composer/chat-input.test.ts
```diff
@@ -18,6 +19,52 @@ describe("resolveChatDraftWorkspaceId", () => {
...
+  it("falls back to the slot mode id when no live config is present", () => {
+    expect(resolveCurrentModeLabel({ modeId: "plan" })).toBe("plan");
+  });
+
+  it("returns null when no mode is resolvable", () => {
+    expect(resolveCurrentModeLabel(null)).toBeNull();
+  });
+});
+
 describe("resolveChatInputAvailability", () => {
```
pablonyx, line 68: Test coverage only exercises the label path (`resolveCurrentModeLabel`). The other touched paths added no collaborationMode assertions, and the missed sibling sites have existing test files with zero coverage of this case: `summary.test.ts`, `session-stream-side-effects.test.ts`, and `session-mode-preferences.test.ts` all have 0 references to `collaboration`. Given this is a multi-site "all paths" fix that adversarial review already found one missed site for, please add: (1) a `buildSessionSlotPatchFromSummary` test asserting modeId/currentModeId tracks collaborationMode, and (2) a `persistReconciledModePreferences` test asserting a reconciled collaboration_mode change persists the default. Without these, regressions on the summary/reconcile paths stay untested.

pablonyx:
**Spec critique (round 1):** The core change is correct and spec-conformant. Codex is the only agent family with both `mode` (permissions) and `collaboration_mode` (default/plan) — see specs/codebase/structures/anyharness/src/agent-mode-matrix.md (Codex section, lines 95-136), which explicitly keeps the two controls distinct. Preferring `collaborationMode ?? mode` for the current-mode/label/persistence mirrors the established reference in `use-chat-session-controls.ts:57-63` and `config/session-controls.ts` ordering. The five touched sites (reducer, session-intent dispatch, chat-input label, stream-patch, session-records hydration) are all internally consistent and the new unit tests pass.

However, the PR body claims the fix lands "across ALL paths," and adversarial review finds it does not — there are three same-bug sibling sites left unaddressed:

1. **major — summary-apply modeId derivation (regression):** `buildSessionSlotPatchFromSummary` (lib/domain/sessions/summary.ts:34-37) still reads only `mode?.currentValue`. It feeds `applySessionSummary` (use-session-summary-actions.ts:71), which writes `patch.modeId` and `patch.transcript.currentModeId`. On every authoritative summary flush a plan/default Codex session gets its badge/currentModeId clobbered back to the permission mode — the exact clobber the stream-patch fix was added to prevent. This is the live-path twin of the stream-patch change and should be fixed identically.

2. **major — reconciliation persistence guard mismatch:** `persistReconciledModePreferences` is called from use-session-summary-actions.ts:146 and session-stream-side-effects.ts:130, both passing `mode?.rawConfigId` as `liveConfigRawConfigId`. The guard at session-mode-preferences.ts:27 requires it to equal `change.rawConfigId`; a reconciled `collaboration_mode` change never matches, so the plan/default default preference is silently dropped on the reconcile path. This is the same guard-mismatch the PR already fixed in session-intent-config-dispatch.ts.

3. **minor — test coverage:** Only the label path gets collaborationMode tests. summary.test.ts, session-stream-side-effects.test.ts, and session-mode-preferences.test.ts have zero `collaboration` references, so both gaps above are untested.

`use-session-model-fallback-action.ts:50` and `cowork-thread-session-record.ts:35` also read only `mode` but are lower priority (model-fallback edge case / cowork hides both controls). Net: the change is right but incomplete versus its stated acceptance; the two major sibling sites mean a plan/default Codex session can still visibly revert and lose its persisted default on the live summary/reconcile paths.

pablonyx:

PR #763, opened 2026-06-18
fix(workspaces): stable local-workspace identity + no right-panel toggle on create (#11, #13)

pablonyx:
Stack 6/9. **#11**: every member of a folder+branch bucket gets a unique local-slot id (no aliasing index 0), so a new local workspace never momentarily shows another's content. **#13**: stop toggling right-panel suppression across pending→materialized (no close-then-open). Verified: acceptance met.

### DIFF
--- a/apps/desktop/src/hooks/main/facade/use-main-screen-state.ts
+++ b/apps/desktop/src/hooks/main/facade/use-main-screen-state.ts
@@ -332,7 +332,13 @@ export function useMainScreenState(): MainScreenState {
     [selectedCloudWorkspaceId, workspaceCollections?.cloudWorkspaces],
   );
 
-  const rightPanelSuppressed = Boolean(pendingWorkspaceEntry);
+  // Only suppress the right panel while entering a *home-originated* create (no
+  // prior workspace, so no panel to disturb). A workspace-originated create keeps
+  // the origin panel's durable state across the pending->materialized window, so
+  // an already-open panel never animates a close-then-open on create (#13).
+  const rightPanelSuppressed = Boolean(
+    pendingWorkspaceEntry && pendingWorkspaceEntry.originTarget.kind === "home",
+  );
   const userOpenOverrideActive = Boolean(
     rightPanelUserOpenOverride
     && rightPanelUserOpenOverride.materializedWorkspaceId === materializedWorkspaceId
--- a/apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces-duplicates.test.ts
+++ b/apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces-duplicates.test.ts
@@ -81,7 +81,7 @@ function makeMobilityWorkspace(args: {
 }
 
 describe("logical workspace duplicate local records", () => {
-  it("splits duplicate local workspaces while merging cloud and mobility only with canonical", () => {
+  it("gives every member of a multi-local bucket a distinct slot id and keeps cloud/mobility separate", () => {
     const repoRoot = makeRepoRoot();
     const olderLocal = makeWorkspace({
       id: "local-older",
@@ -113,7 +113,8 @@ describe("logical workspace duplicate local records", () => {
       "proliferate",
       "main",
     );
-    const slotId = buildLocalSlotLogicalWorkspaceId(newerLocal.id);
+    const newerSlotId = buildLocalSlotLogicalWorkspaceId(newerLocal.id);
+    const olderSlotId = buildLocalSlotLogicalWorkspaceId(olderLocal.id);
 
     const logicalWorkspaces = buildLogicalWorkspaces({
       localWorkspaces: [newerLocal, olderLocal],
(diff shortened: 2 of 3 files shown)


apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces.ts
```diff
@@ -240,9 +240,18 @@ export function buildLogicalWorkspaces(args: {
...
-    sortedBucket.forEach((collapsed, index) => {
+    // A singleton bucket keeps the shared base id so cloud/mobility records for
+    // the same folder+branch merge onto it. Every member of a multi-workspace
+    // bucket instead gets a stable, per-workspace `local-slot:${id}` id: keeping
+    // index 0 on the shared base id aliased two distinct local workspaces during
+    // the pending window (a new local workspace momentarily showed another's
+    // session/content as the canonical sort order shifted which member held the
+    // base id). Sidebar grouping is keyed off `repoKey`, not the logical id, so
+    // distinct slot ids leave grouping intact (#11).
+    const isSingletonBucket = sortedBucket.length === 1;
```
pablonyx, line 251: Behavior change worth calling out: with index 0 no longer holding `baseLogicalId`, a multi-local bucket (2+ local workspaces at the same path+branch) that also has a cloud/mobility record for that same folder+branch now emits a **separate cloud-only logical entry** under `baseLogicalId` with `localWorkspace: null` (the updated test asserts this: length 3, `canonical?.localWorkspace` is null). Previously the cloud/mobility merged invisibly onto the index-0 local row. That orphaned cloud entry resolves `effectiveOwnerHint='local'` but has no local workspace, so `resolvePreferredLogicalWorkspaceMaterialization` falls through to the cloud synthetic id while `inferLifecycle` still returns `local_active` — i.e. a standalone 'local_active' sidebar row that points at a cloud materialization. Narrow scenario and arguably more correct (the cloud record can't sensibly alias to one of several locals), and grouping stays intact via `repoKey`, but it is a new visible row in that state. Confirm this is acceptable vs. attaching the cloud/mobility to the canonical-order-first local member's slot.

apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces-duplicates.test.ts
```diff
@@ -123,19 +124,25 @@ describe("logical workspace duplicate local records", () => {
...
+    expect(newerSlot?.cloudWorkspace).toBeNull();
+    expect(newerSlot?.mobilityWorkspace).toBeNull();
+    expect(olderSlot?.localWorkspace?.id).toBe(olderLocal.id);
+    expect(olderSlot?.cloudWorkspace).toBeNull();
+    expect(olderSlot?.mobilityWorkspace).toBeNull();
+    // Cloud + mobility records for the same folder+branch collapse into their own
+    // canonical entry with no aliased local workspace.
+    expect(canonical?.localWorkspace).toBeNull();
     expect(canonical?.cloudWorkspace?.id).toBe(cloudWorkspace.id);
     expect(canonical?.mobilityWorkspace?.id).toBe(mobilityWorkspace.id);
```
pablonyx, line 145: The rewrite dropped the old assertion that `logicalWorkspaceRelatedIds(canonical)` contained the slot id — fine since the merge no longer happens. But there's now no test proving selection routing still works across the new 3-entry shape: e.g. that `findLogicalWorkspace(logicalWorkspaces, <cloud synthetic id>)` resolves to the new cloud-only `canonical` row, and that selecting `newerSlotId`/`olderSlotId` resolves to the right local. Given this is the exact #11 hazard (a new local momentarily inheriting another's identity), please add a selection-resolution assertion for the multi-local + cloud case so a future regression in id routing is caught.

apps/desktop/src/hooks/main/facade/use-main-screen-state.ts
```diff
@@ -332,7 +332,13 @@ export function useMainScreenState(): MainScreenState {
...
   );
 
-  const rightPanelSuppressed = Boolean(pendingWorkspaceEntry);
+  // Only suppress the right panel while entering a *home-originated* create (no
+  // prior workspace, so no panel to disturb). A workspace-originated create keeps
+  // the origin panel's durable state across the pending->materialized window, so
+  // an already-open panel never animates a close-then-open on create (#13).
+  const rightPanelSuppressed = Boolean(
+    pendingWorkspaceEntry && pendingWorkspaceEntry.originTarget.kind === "home",
+  );
```
pablonyx, line 341: #13 fix reads correctly: `originTarget` is a clean 2-variant union (`home` | `workspace`), suppression now fires only for home-origin so a workspace-originated create never toggles `rightPanelSuppressed` true→false across pending→materialized, and the `userOpenOverride` effect (lines 350-362) + `userOpenOverrideActive` both gate on `rightPanelSuppressed` so they stay inert for workspace-origin. No regression spotted. Only gap: this is inline hook logic with no unit test, so the no-close-then-open behavior rests entirely on manual verification. Consider extracting the `rightPanelSuppressed`/override resolution into a pure helper that can be unit-tested against the home vs workspace origin cases.

pablonyx:
**Spec critique (round 1):** Solid, well-scoped PR. Both stated fixes land and conform to the pending-workspace-shell spec ("Handoffs must preserve visible identity", sidebar id handoff rules). **#11**: every member of a multi-local folder+branch bucket now gets a per-workspace `local-slot:` id instead of index 0 squatting on the shared base id — directly kills the "new local momentarily inherits another's identity" race during the pending window; grouping is unaffected (keyed off `repoKey`). **#13**: right-panel suppression now keys off `originTarget.kind === "home"`, so a workspace-originated create no longer toggles suppression true→false across pending→materialized; the override effect + `userOpenOverrideActive` both correctly stay inert for workspace-origin. Diff is 3 on-topic files, no scope leak, full cloud+sidebar suite (197 tests) green.

Two non-blocking notes: (1) the multi-local + cloud/mobility case now emits a **separate cloud-only logical row** (was merged onto index-0 local) — narrow and arguably more correct, but it's a new visible `local_active` row pointing at a cloud materialization; confirm that's intended (inline). (2) test/coverage gaps: the rewritten duplicates test no longer asserts selection routing across the new 3-entry shape, and the #13 hook change has no unit test (inline suggestion to extract a pure helper). Nothing here blocks merge.

apps/desktop/src/lib/domain/workspaces/cloud/logical-workspaces.ts
pablonyx, line 251: Behavior note (not a blocker): with `isSingletonBucket=false`, *no* bucket member keeps `baseLogicalId`, so a cloud/mobility record for the same folder+branch no longer merges onto a local — it now surfaces as its own `localWorkspace:null` entry (the updated test asserts exactly this: `canonical.localWorkspace` is null). For the realistic "2 worktrees on `main` with chats + 1 cloud workspace on `main`" case this changes the sidebar from 2 rows to 3 (two locals + a standalone cloud row). This is a deliberate, spec-aligned tradeoff (matches the "keep distinct local threads separate" philosophy and fixes the #11 aliasing), and grouping stays intact because `repoKey` for the cloud entry (`cloudWorkspaceGroupKey` -> `provider:owner:name`) matches `repoRootGroupKey` for the locals when the repoRoot has a remote. Flagging only so the extra standalone-cloud row is an intended product outcome, not a surprise.

pablonyx:
**Spec critique (round 1):** Solid PR — both fixes are correct, well-scoped, and spec-aligned (pending-workspace-shell.md invariant 7 "handoffs must preserve visible identity" + the sidebar handoff rules).

**#11 (stable local-workspace identity):** Keying every multi-bucket member to a per-`workspace.id` `local-slot:` id (instead of giving the sort-index-0 member the shared base id) is the right fix for the aliasing bug — the logical id can no longer re-point to a different local as the canonical sort order shifts during the pending window. Singleton buckets still keep the base id so cloud/mobility merge correctly (verified by the "promotes a prior local-slot to canonical" test). Alias lookup (`logicalWorkspaceMatchesId` / `findLogicalWorkspace`) still resolves each local via its own slot id, so the removed `relatedIds`-contains assertion is not a lost capability.

**#13 (no right-panel toggle on create):** Scoping suppression to `originTarget.kind === "home"` is correct. Verified the open/closed panel state is carried by the shell-level `rightPanelSessionDurableState` (use-main-screen-state.ts:103/154), which survives the switch to the pending UI key, so a workspace-originated create with the panel already open keeps it open (no close-then-open). The dependent `userOpenOverrideActive` and its reset effect remain consistent now that suppression is off for workspace-origin.

**Verification:** `tsc --noEmit` clean; the changed duplicates test + all cloud (95) and sidebar (102) tests pass. Diff is exactly the single squash commit — only the 3 intended files, no leaked stack changes.

One non-blocking behavior note left inline on logical-workspaces.ts:251: in a multi-local + cloud bucket the cloud now surfaces as its own standalone row rather than merging onto a local, so e.g. 2 worktrees + 1 cloud on `main` renders 3 sidebar rows instead of 2. Deliberate, grouping-safe tradeoff — flagging only to confirm the extra standalone-cloud row is intended.

No blocking issues. severityMax = nit.

pablonyx:

PR #764, opened 2026-06-18
fix(home): navigate before painting the first message (#1)

pablonyx:
Stack 7/9. Enter the pending session shell before the optimistic paint so the first message appears on the destination, not flashing on home; the projected pending session owns the preview. Verified: acceptance met (message not dropped — destination shell owns it).

### DIFF
--- a/apps/desktop/src/components/home/screen/HomeNextScreen.tsx
+++ b/apps/desktop/src/components/home/screen/HomeNextScreen.tsx
@@ -11,7 +11,6 @@ import { ChatComposerActions } from "@/components/workspace/chat/input/ChatCompo
 import { SessionConfigControls } from "@/components/workspace/chat/input/SessionConfigControls";
 import { ChatComposerSurface } from "@proliferate/product-ui/chat/composer/ChatComposerSurface";
 import { ComposerTextarea } from "@proliferate/ui/primitives/ComposerTextarea";
-import { UserMessage } from "@/components/workspace/chat/transcript/UserMessage";
 import { Button } from "@proliferate/ui/primitives/Button";
 import { useHomeNextLaunchControls } from "@/hooks/home/derived/use-home-next-launch-controls";
 import { useHomeCloudRepoSettingsNavigation } from "@/hooks/home/workflows/use-home-cloud-repo-settings-navigation";
@@ -213,20 +212,6 @@ export function HomeNextScreen() {
               onConfigureCloud={configureCloud}
             />
           </div>
-
-          {composer.submittedPreview ? (
-            <div
-              key={composer.submittedPreview.id}
-              className="mt-5"
-              data-home-submit-preview
-            >
-              <UserMessage
-                sessionId={null}
-                content={composer.submittedPreview.text}
-                contentParts={[{ type: "text", text: composer.submittedPreview.text }]}
-              />
-            </div>
-          ) : null}
 
           {modelAvailabilityNotice ? (
             <div className="mx-auto mt-2 flex max-w-2xl items-center justify-center gap-2 px-2 text-center text-sm text-muted-foreground">
--- a/apps/desktop/src/hooks/home/ui/use-home-next-composer-state.ts
+++ b/apps/desktop/src/hooks/home/ui/use-home-next-composer-state.ts
@@ -1,6 +1,5 @@
 import { useCallback, useEffect, useLayoutEffect, useRef, useState } from "react";
 import type { KeyboardEvent } from "react";
-import { flushSync } from "react-dom";
 import { HOME_CHAT_COMPOSER_INPUT } from "@/config/chat";
 import { useHomeNextLaunch } from "@/hooks/home/workflows/use-home-next-launch";
 import { useHomeDraftHandoffStore } from "@/stores/home/home-draft-handoff-store";
@@ -9,13 +8,6 @@ import type {
   HomeNextModelSelection,
   ModelAvailabilityState,
 } from "@/lib/domain/home/home-next-launch";
-import { scheduleAfterNextPaint } from "@/lib/infra/scheduling/schedule-after-next-paint";
-
-function waitForNextPaint(): Promise<void> {
-  return new Promise((resolve) => {
-    scheduleAfterNextPaint(resolve);
-  });
-}
 
 interface UseHomeNextComposerStateArgs {
   targetDisabledReason: string | null;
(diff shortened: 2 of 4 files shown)


apps/desktop/src/components/home/screen/HomeNextScreen.tsx
```diff
@@ -214,20 +213,6 @@ export function HomeNextScreen() {
...
-            >
-              <UserMessage
-                sessionId={null}
-                content={composer.submittedPreview.text}
-                contentParts={[{ type: "text", text: composer.submittedPreview.text }]}
-              />
-            </div>
-          ) : null}
-
           {modelAvailabilityNotice ? (
```
pablonyx, line 216: **Blocker — stale test now fails CI.** This PR removes the home-side submitted-preview block here (the `data-home-submit-preview` / `UserMessage` render that previously sat above this `modelAvailabilityNotice`), but it does not update `HomeNextScreen.test.tsx`. The test `it("keeps the submitted preview for repository launches")` (HomeNextScreen.test.tsx:292) still does `fireEvent.click(Submit)` for a worktree target and asserts `screen.getByText("start worktree")` is present (line 305). With the preview gone, that assertion now throws — I ran it and `vitest run src/components/home/screen/HomeNextScreen.test.tsx` reports `1 failed | 12 passed`. Update this test to assert the preview is **absent** for repository launches (mirror the existing cowork test at line 279), and drop the now-dead `UserMessage` mock at lines 159–162.

pablonyx:
**Spec critique (round 1):** The core change is correct and well-scoped against the pending-workspace-shell spec. The navigate-first claim verified: `beginPendingWorkspace` (use-workspace-entry-flow.ts) synchronously enters the pending shell and creates the projected session inside `launch()` for local/worktree/cloud-new paths, and `promptProjectedPendingWorkspaceSession`/`promptProjectedOrCreateFreshSession` queue the optimistic prompt on the destination session — so removing the home-side `submittedPreview` does not drop the message (spec invariants 4/5, "projected session is chat truth"). The redundant `flushSync` + `waitForNextPaint` removal is safe: `scheduleAfterNextPaint` is still used widely elsewhere (not orphaned). Double-submit is still guarded synchronously by `inFlightRef` even though `canSubmit` no longer checks `submittedPreview`. The cowork path is unchanged (it never painted a home preview). New hook test passes.

**One blocker:** the change breaks an existing test that was not updated. `HomeNextScreen.test.tsx:292` ("keeps the submitted preview for repository launches") asserts `screen.getByText("start worktree")` after submit, which relied on the removed preview. `vitest run src/components/home/screen/HomeNextScreen.test.tsx` → `1 failed | 12 passed`. This file is not in the PR diff, so CI will go red. Fix: invert that test to assert the preview is absent (mirror the cowork test at line 279) and remove the now-dead `UserMessage` mock (lines 159–162). Inline comment left on HomeNextScreen.tsx.

apps/desktop/src/components/home/screen/HomeNextScreen.tsx
pablonyx, line 216: Blocker: this PR deletes the home-side `submittedPreview` render block here, but does not update the existing `HomeNextScreen.test.tsx`, which still asserts the old behavior. The test `"keeps the submitted preview for repository launches"` (HomeNextScreen.test.tsx:292-306) does `expect(screen.getByText("start worktree")).toBeTruthy()` after a worktree submit — with the preview block gone, nothing paints that text on home, so the assertion throws. I ran the suite: `1 failed | 12 passed`, failing at HomeNextScreen.test.tsx:305. This is CI-breaking. Update that test to assert the preview is no longer rendered for repository launches (mirroring the cowork case at line 279), and consider renaming it.

apps/desktop/src/hooks/home/ui/use-home-next-composer-state.ts
```diff
@@ -89,20 +76,11 @@ export function useHomeNextComposerState({
...
-      setDraft("");
-    });
-    if (shouldShowHomeSubmittedPreview) {
-      await waitForNextPaint();
-    }
+    // Navigate-first: launch() enters the pending-workspace shell and navigates
+    // before any optimistic paint, so the user's message renders on the
+    // destination session shell rather than flashing on home. We only clear the
+    // draft here; the projected pending session owns the optimistic preview.
+    setDraft("");
```
pablonyx, line 83: Minor (worktree path only): now that the home-side preview is gone, the user-visible bridge relies on `launch()` mounting the pending workspace shell synchronously. For `local` and `cloud` targets that holds — `beginPendingWorkspace` runs before the first `await` in their entry flows, so the workspace shell (showing the launch-intent UserMessage) mounts in the same tick. But the `worktree` path awaits `resolveWorktreeCreationInput` (which `await getHomeDir()` and reads prefs) *before* `beginPendingWorkspace` (use-workspace-entry-actions.ts:240-257). During that await the composer is already cleared and no preview exists, so there can be a brief frame on Home with the prompt gone and the shell not yet mounted. `getHomeDir()` is usually fast so this is minor, but worth confirming there is no visible flash for worktree launches (the local/cloud paths are clean).

pablonyx:
**Spec critique (round 1):** The core change is sound and spec-aligned — removing the home-side `submittedPreview` and letting the projected pending session / launch-intent pane own the first message follows the pending-workspace-shell spec (invariant #4, and the anti-pattern against a second optimistic surface). I verified the handoff: `launch()` calls `beginLaunchIntent(...)` synchronously, and for `local` and `cloud` targets `beginPendingWorkspace` runs before the first `await`, so the workspace shell mounts in the same tick and renders the optimistic `UserMessage` via `ChatLaunchIntentPane` — the message is not dropped, acceptance met. The new `use-home-next-composer-state.test.tsx` cases (no `submittedPreview` exposed, draft cleared + launch args, draft restored on failure) are focused and pass.

One blocker: the PR deletes the preview render from `HomeNextScreen.tsx` but does not update the pre-existing `HomeNextScreen.test.tsx`, which still has `"keeps the submitted preview for repository launches"` (line 292-306) asserting `getByText("start worktree")`. I ran the file: **1 failed | 12 passed**, failing at HomeNextScreen.test.tsx:305 — CI-breaking. That test must be updated (mirror the cowork "does not render" case at line 279).

One minor: the `worktree` path awaits `resolveWorktreeCreationInput` (→ `getHomeDir()`) before `beginPendingWorkspace`, so unlike local/cloud there can be a brief Home frame with the prompt cleared and no shell yet. Usually fast, but worth confirming no visible flash. Scope is clean (3 files, all home-composer). Fix the stale test and this is good to land.

pablonyx:

PR #765, opened 2026-06-18
fix(workspaces): always land on a session when one exists (#14)

pablonyx:
Stack 8/9. Final fallback in bootstrapWorkspace: if current and the workspace has ≥1 session but no active session was selected (race/staleness), select choosePreferredWorkspaceSession. Verified: acceptance met.

### DIFF
--- a/apps/desktop/src/hooks/workspaces/workflows/use-workspace-bootstrap-actions.ts
+++ b/apps/desktop/src/hooks/workspaces/workflows/use-workspace-bootstrap-actions.ts
@@ -53,6 +53,8 @@ import {
 } from "@/lib/domain/workspaces/selection/optimistic-session-shell";
 import { handleEmptyWorkspaceBootstrap } from "@/hooks/workspaces/workflows/workspace-bootstrap-empty-session";
 import { handleRememberedWorkspaceSessionBootstrap } from "@/hooks/workspaces/workflows/workspace-bootstrap-remembered-session";
+import { selectSessionWithShellIntentRollback } from "@/hooks/sessions/workflows/session-shell-selection";
+import { choosePreferredWorkspaceSession } from "@/lib/domain/workspaces/selection/selection";
 
 interface BootstrapWorkspaceInput {
   workspaceId: string;
@@ -323,6 +325,34 @@ export function useWorkspaceBootstrapActions() {
 
     if (isCurrent()) {
       markWorkspaceBootstrappedInSession(workspaceId);
+    }
+
+    // Final guarantee: the bootstrap handlers above can fall through with no
+    // active session (race/staleness holes), stranding the user on the empty
+    // hero even though the workspace has sessions. If we're still current and
+    // nothing got selected, land on the preferred session so a workspace with
+    // >=1 session never shows the empty state.
+    if (
+      isCurrent()
+      && sessions.length > 0
+      && !useSessionSelectionStore.getState().activeSessionId
+    ) {
+      const fallbackSession = choosePreferredWorkspaceSession(sessions, null);
+      if (fallbackSession) {
+        logLatency("workspace.select.session_fallback", {
+          workspaceId,
+          logicalWorkspaceId,
+          sessionId: fallbackSession.id,
+          sessionCount: sessions.length,
+          totalElapsedMs: elapsedMs(startedAt),
+        });
+        await selectSessionWithShellIntentRollback({
+          workspaceId,
+          sessionId: fallbackSession.id,
+          options: { latencyFlowId },
+          selectSession,
+        });
+      }
     }
 
     return { sessions };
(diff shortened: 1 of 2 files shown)


apps/desktop/src/lib/domain/workspaces/selection/selection.test.ts
```diff
@@ -53,6 +53,22 @@ describe("choosePreferredWorkspaceSession", () => {
...
+  it("never returns null for a non-empty list (bootstrap fallback guarantee)", () => {
+    // The bootstrap final-guarantee fallback calls this with a null last-viewed
+    // id; it must always resolve to a session so a workspace with >=1 session
+    // never strands the user on the empty hero (#14).
+    const sessions = [
+      { id: "no-timestamps" },
+      { id: "setup-only", updatedAt: "2026-04-07T20:00:00.000Z" },
+    ];
+
+    expect(choosePreferredWorkspaceSession(sessions, null)).not.toBeNull();
```
pablonyx, line 66: Minor (test coverage): both added tests exercise `choosePreferredWorkspaceSession`, whose implementation this PR does not touch — the "never null for a non-empty list" property was already guaranteed by the pre-existing impl (selection.ts:23-43, the `rankedSessions = promptedSessions.length > 0 ? promptedSessions : sessions` fallback). So these document the invariant the fallback relies on, but do not cover the actual new code path: the inline final-guarantee block in use-workspace-bootstrap-actions.ts:335-356 (guarded by `isCurrent() && sessions.length > 0 && !activeSessionId`, then `selectSessionWithShellIntentRollback`). The acceptance criterion ('selecting a workspace with >=1 session always lands on one') lives in that block, and it's untested. There's already a sibling harness for this exact shape — workspace-bootstrap-remembered-session.test.ts mocks the deps and asserts `selectSession` is called — so a regression test asserting the fallback selects when the remembered handler returns with activeSessionId still null is cheap to add and would actually guard #14.

apps/desktop/src/hooks/workspaces/workflows/use-workspace-bootstrap-actions.ts
```diff
@@ -325,6 +327,34 @@ export function useWorkspaceBootstrapActions() {
...
+      const fallbackSession = choosePreferredWorkspaceSession(sessions, null);
+      if (fallbackSession) {
+        logLatency("workspace.select.session_fallback", {
+          workspaceId,
+          logicalWorkspaceId,
+          sessionId: fallbackSession.id,
+          sessionCount: sessions.length,
+          totalElapsedMs: elapsedMs(startedAt),
+        });
+        await selectSessionWithShellIntentRollback({
```
pablonyx, line 349: Nit (observation, not a blocker): unlike the remembered-session handler (workspace-bootstrap-remembered-session.ts:100-120), this fallback does not re-check `isCurrent()` after the awaited `selectSessionWithShellIntentRollback`, nor does it inspect the `stale` outcome, nor hydrate transcript history (`rehydrateSessionSlotFromHistory`). That's defensible — this is a last-resort guard that fires only when the normal handlers left activeSessionId null, and `selectSessionWithShellIntentRollback` already rolls back its own shell intent on a stale outcome — so the session still lands and history hydrates lazily on the next activation. Just flagging that the two selection paths now differ; if a future change makes the fallback the common path, the missing post-await current-check + hydrate would matter. Fine to leave as-is for #14's scope.

pablonyx:
**Spec critique (round 1):** Solid, well-scoped fix for #14. The new final-guarantee block in `use-workspace-bootstrap-actions.ts:335-356` is correctly gated (`isCurrent() && sessions.length > 0 && !activeSessionId`) and selects an existing server-loaded session via the same `selectSessionWithShellIntentRollback` + `fallbackSession.id` path the remembered handler already uses — so the session-id semantics are consistent and it never creates a second session. I verified it does not violate the pending-workspace-shell spec: a projected pending session always implies `activeSessionId` is set (`resolveActiveProjectedSessionForPendingWorkspace` requires it), so the `!activeSessionId` guard prevents this fallback from ever overriding a projected session (no conflict with invariants #4/#6 or the Section 11 'projected-session existence blocks fresh fallback' rule). The empty-handler path only runs when `sessions.length === 0`, so there's no overlap there either. tsc is clean and the 8 selection tests pass. Scope is exactly the 2 intended files. One **minor**: the two added tests cover `choosePreferredWorkspaceSession` (whose impl this PR doesn't change and which already guaranteed never-null for non-empty lists) rather than the new inline fallback path that actually carries the acceptance criterion — a regression test in the style of `workspace-bootstrap-remembered-session.test.ts` (mock deps, assert `selectSession` is called when the handler returns with activeSessionId null) would actually guard #14. One **nit**: the fallback omits the post-await `isCurrent()`/stale re-check and `rehydrateSessionSlotFromHistory` that the remembered handler does — acceptable for a last-resort guard since the rollback helper handles staleness and history hydrates lazily. Neither blocks merge.

pablonyx:

PR #766, opened 2026-06-18
fix(transcript): correct file-path link resolution (#12)

pablonyx:
Stack 9/9. Scheme-less hosts (github.com/…) render as external links not dead file mentions; fuzzy resolver abstains on a truncated result set and is skipped for authoritative tool-call paths; tool-chip label/open-target aligned on moves. Verified: acceptance met.

### DIFF
--- a/apps/desktop/src/hooks/workspaces/workflows/files/use-fuzzy-file-resolver.ts
+++ b/apps/desktop/src/hooks/workspaces/workflows/files/use-fuzzy-file-resolver.ts
@@ -38,16 +38,19 @@ export function useFuzzyFileResolver() {
         // High limit (server caps at 200): a basename can match many files, and
         // truncating the real one out would defeat the "already exists" guard
         // and risk correcting a valid path to the wrong file.
+        const limit = 200;
         const response = await searchWorkspaceFiles(
           connection,
           connection.anyharnessWorkspaceId,
           basename,
-          200,
+          limit,
         );
-        return pickFuzzyPathMatch(
-          targetPath,
-          (response.results ?? []).map((result) => result.path),
-        );
+        const candidatePaths = (response.results ?? []).map((result) => result.path);
+        // When the result set is capped, the exact file or a second suffix match
+        // may have been dropped — abstain rather than risk a wrong correction.
+        return pickFuzzyPathMatch(targetPath, candidatePaths, {
+          truncated: candidatePaths.length >= limit,
+        });
       } catch {
         return null;
       }
--- a/apps/packages/product-ui/src/chat/transcript/ProviderLinkMention.tsx
+++ b/apps/packages/product-ui/src/chat/transcript/ProviderLinkMention.tsx
@@ -132,16 +132,63 @@ export function linkHost(href: string): string | null {
 
 /**
  * Whether an href should be treated as an external web link (vs. a workspace
- * file path). Deliberately conservative: only `http(s)://…` and `www.…`.
+ * file path). Conservative: `http(s)://…`, `www.…`, and a bare `host.tld/path`
+ * form only when the TLD is a well-known web TLD (see {@link isSchemelessWebHost}).
  *
- * A bare `host.tld/path` form (`github.com/x`) is NOT claimed, because it is
- * structurally indistinguishable from a relative file path with a dotted
- * directory (`v1.2/notes.txt`, `CHANGELOG.md/x`) — claiming those would lose
- * the file mention and fire a favicon fetch for a bogus host. Scheme-less URLs
- * are rare in agent output (they almost always emit `https://`), so this trades
- * that edge for never stealing a real file path.
+ * The TLD allow-list is what keeps a relative file path with a dotted directory
+ * (`v1.2/notes.txt`, `CHANGELOG.md/x`) from being mistaken for a host: their
+ * "TLD" (`2`, `md`) is not a web TLD, so they stay file mentions. Without this,
+ * `github.com/org/repo` rendered as a dead FilePathLink (it never resolves to a
+ * workspace file), so we now claim it as an external link.
  */
 export function isExternalHttpLink(href: string): boolean {
   const value = href.trim();
-  return /^https?:\/\//i.test(value) || /^www\.[a-z0-9-]/i.test(value);
+  return (
+    /^https?:\/\//i.test(value) ||
+    /^www\.[a-z0-9-]/i.test(value) ||
+    isSchemelessWebHost(value)
+  );
+}
+
+/**
+ * Web TLDs common enough that they unambiguously signal a host rather than a
+ * file extension. Deliberately small: every entry must be a TLD that is never a
+ * real file extension (so it can't steal `notes.md`, `data.io` would be rare).
+ */
+const WEB_TLDS = new Set([
+  "com",
+  "org",
+  "net",
+  "io",
+  "dev",
+  "app",
+  "ai",
+  "co",
+  "gov",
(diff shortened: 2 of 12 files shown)


apps/desktop/src/hooks/workspaces/workflows/files/use-fuzzy-file-resolver.ts
```diff
@@ -38,16 +38,19 @@ export function useFuzzyFileResolver() {
...
-        return pickFuzzyPathMatch(
-          targetPath,
-          (response.results ?? []).map((result) => result.path),
+          limit,
         );
+        const candidatePaths = (response.results ?? []).map((result) => result.path);
+        // When the result set is capped, the exact file or a second suffix match
+        // may have been dropped — abstain rather than risk a wrong correction.
+        return pickFuzzyPathMatch(targetPath, candidatePaths, {
+          truncated: candidatePaths.length >= limit,
```
pablonyx, line 52: Nit (intentional but worth noting): `candidatePaths.length >= limit` treats an *exactly*-200-result set as truncated even when 200 is the complete match set, so a would-be-unique correction is dropped. That's the safe direction (abstain), and 200 same-basename matches is vanishingly rare, so fine to keep — just flagging that the heuristic is 'capped-or-exactly-full', not strictly 'truncated'.

apps/packages/product-ui/src/chat/transcript/ProviderLinkMention.tsx
```diff
@@ -132,16 +132,63 @@ export function linkHost(href: string): string | null {
...
+  // Require a path segment after the host so a bare token isn't claimed.
+  if (slash <= 0 || slash === trimmed.length - 1) {
+    return false;
+  }
+  const host = trimmed.slice(0, slash).toLowerCase();
+  const labels = host.split(".");
+  if (labels.length < 2 || labels.some((label) => label.length === 0)) {
+    return false;
+  }
+  return WEB_TLDS.has(labels[labels.length - 1]);
```
pablonyx, line 193: Minor: the doc comment claims every WEB_TLD 'is never a real file extension', but `app`, `co`, `io`, `dev`, `ai` can all appear as the first path segment of a real relative path (`my.app/config`, `pkg.dev/x`). The path-segment + first-label-only check makes a *false claim* require a dotted directory whose final label is exactly one of these AND a following segment — rare in agent output, and the failure mode is just rendering a file mention as an external link (favicon fetch on a bogus host), not a crash. Acceptable tradeoff; consider softening the comment's 'never' wording.

pablonyx:
**Spec critique (round 1):** Solid, well-scoped fix that meets its three acceptance criteria — scheme-less `host.tld/path` links with a web TLD now route to `ProviderLinkMention` instead of dead `FilePathLink`s (verified via `renderTranscriptLink`'s `isExternalHttpLink` early-return); the fuzzy resolver abstains on a capped result set (`pickFuzzyPathMatch({truncated})`); and authoritative tool-call paths skip the fuzzy backstop in `useFileReferenceActions`. Conforms to chat-transcript.md (URL detection runs before file detection; only workspace refs become FilePathLinks) and workspace-files.md. Scope is clean (all 11 files belong to this unit, no leaked scroll/session/auth changes), tests pass, and both `apps/desktop` and `product-ui` typecheck.

One actionable gap: the authoritative-path treatment was applied to `FileChangeCall`, `ToolFileChip`, and `FileReferenceBadge`, but **`CollapsedEditActionRows.tsx` was missed** — it renders the same `file_change` tool-call parts (collapsed edit rows), pulls `pathLabel`/`workspacePath` straight from the part (lines 54/63), yet neither its `useFileReferenceActions` (line 64) nor its `ActionFileLink` chip passes `authoritativePath`. So a collapsed edit row can still fuzzy-correct to a different same-basename file while the expanded `FileChangeCall` for the identical part won't — contradicting the PR's own invariant. That file isn't in the diff so I couldn't attach an inline comment; flagging here. (`CollapsedEditActionRows` line 54 also already uses destination-first `newWorkspacePath ?? workspacePath ?? ...` for all ops, which matches the new move logic — only the authoritative flag is missing.)

Plus two minor/nit notes inline: the `>= limit` truncation heuristic is really 'capped-or-exactly-full' (safe direction, fine), and the `WEB_TLDS` doc comment's 'never a real file extension' claim is slightly overstated for `app`/`co`/`io`/`dev`/`ai` (accepted tradeoff, only a cosmetic favicon-on-bogus-host failure mode).

apps/packages/product-ui/src/chat/transcript/ProviderLinkMention.tsx
```diff
@@ -132,16 +132,63 @@ export function linkHost(href: string): string | null {
...
+    isSchemelessWebHost(value)
+  );
+}
+
+/**
+ * Web TLDs common enough that they unambiguously signal a host rather than a
+ * file extension. Deliberately small: every entry must be a TLD that is never a
+ * real file extension (so it can't steal `notes.md`, `data.io` would be rare).
+ */
+const WEB_TLDS = new Set([
```
pablonyx, line 158: Minor / spec tension: claiming `host.tld/path` as an external link narrows the chat-transcript spec invariant "URL detection runs before file-path detection so a real path is never mistaken for a link" (specs/codebase/features/chat-transcript.md:112-113). A relative file path whose first dir segment ends in a WEB_TLDS label (e.g. `data.io/parse.ts`, `pkg.dev/x`, `acme.co/notes.md`) now resolves to a ProviderLinkMention + favicon fetch instead of a FilePathLink. The trade-off is reasoned and documented here, and these collisions are rare, so this is acceptable — but consider adding the caveat to the spec text so the "never" claim and the code stay in sync.

pablonyx:
**Spec critique (round 1):** Solid PR — meets all four stated acceptance criteria and I verified each against the code/tests. (1) Scheme-less hosts: `isExternalHttpLink` now claims `host.tld/path` via a small `WEB_TLDS` allow-list and runs before file-path detection (transcript-markdown.tsx:31), so `github.com/org/repo` renders as a ProviderLinkMention instead of a dead FilePathLink; `linkHost` returns the host and normalizes to `https://` for the anchor. (2) Fuzzy resolver abstains on a capped (`>= 200`) result set — `pickFuzzyPathMatch(..., { truncated })` (path-references.ts) with a test. (3) Authoritative tool-call paths skip the fuzzy backstop end-to-end (`useFileReferenceActions` early-return, threaded through `FileReferenceBadge`/`ToolFileChip`/`FileChangeCall`) with tests for both the corrected and the never-corrected case. (4) Move chips: the diff card's open target is now the destination (`newWorkspacePath ?? workspacePath`), matching the right-side label chip, with a test. tsc clean for desktop + product-ui; all 16 touched tests pass. Scope is tight — every file is file-path/link resolution, nothing from other units leaked in.\n\nTwo low-severity notes (neither blocks): (a) the new scheme-less claiming narrowly weakens the spec's "a real path is never mistaken for a link" invariant for relative paths whose first dir segment ends in a web-TLD label (e.g. `data.io/parse.ts`, `acme.co/notes.md`) — documented in-code, just worth a caveat in chat-transcript.md:112-113. (b) Consistency gap: `renderTranscriptInlineCode` (transcript-markdown.tsx:46) gates only on `looksLikePath` and not `isExternalHttpLink`, so a bare inline-code `github.com/x` still renders as a dead FilePathLink — the same defect this PR fixes for the markdown-link form. Pre-existing and out of this PR's link scope, but a natural follow-up.

pablonyx:

PR #788, opened 2026-06-24
fix(desktop): reduce duplicate subagent visual identities

Julian0444:
Expand the friendly-name pool from 8 to 40 names and decouple the avatar color from the name index using a splitmix-style avalanche mix, so a name collision no longer forces a matching color collision.


Julian0444:
Names of subagents.

PR #790, opened 2026-06-25
chore(dev): split local dev setup and run

pablonyx:
## Summary
- Add profile-centered `make setup PROFILE=...` and `make run PROFILE=...` commands.
- Keep compatibility aliases for `dev`, `dev-init`, and existing build target names.
- Make desktop/web `pnpm dev` start Vite directly and move build-before-dev behavior to `dev:built`.
- Update local development docs to prefer the setup/run split.

## Validation
- `make -n run PROFILE=devsplit-test` has no build/codegen commands
- `make -n setup PROFILE=devsplit-test` has no build/codegen commands
- `make -n dev PROFILE=devsplit-test` has no build/codegen commands
- `make -n build-rust`, `make -n build-frontend`, `make -n build`
- `git diff --cached --check`
- Parsed `apps/desktop/package.json` and `apps/web/package.json` with Node

## Notes
- Opened from a fresh worktree based on `origin/main` so the existing `chimp` branch PR remains untouched.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,18 +55,34 @@ STRIPE_LOCAL_SECRET_ENV = if [ -z "$${STRIPE_SECRET_KEY:-}" ] && command -v stri
 		echo "Using Stripe CLI test key for local billing API calls."; \
 	fi; \
 fi;
+DEV_FRONTEND_ARTIFACTS := \
+	anyharness/sdk/dist \
+	anyharness/sdk-react/dist \
+	cloud/sdk/dist \
+	cloud/sdk-react/dist \
+	apps/packages/design/dist \
+	apps/packages/product-domain/dist \
+	apps/packages/ui/dist \
+	apps/packages/product-ui/dist \
+	apps/packages/product-surfaces/dist
+PROFILE_DB_READY_COMMAND = make server-db-ready;
+PROFILE_DB_ENSURE_COMMAND = LOCAL_PGHOST="$(LOCAL_PGHOST)" LOCAL_PGPORT="$(LOCAL_PGPORT)" LOCAL_PGUSER="$(LOCAL_PGUSER)" LOCAL_PGPASSWORD="$(LOCAL_PGPASSWORD)" USE_EXISTING_POSTGRES="$(USE_EXISTING_POSTGRES)" node scripts/dev.mjs ensure-db --db-name "$$PROLIFERATE_DEV_DB_NAME";
+ifneq ($(origin DATABASE_URL), undefined)
+PROFILE_DB_READY_COMMAND = :;
+PROFILE_DB_ENSURE_COMMAND = :;
+endif
 
-ifneq ($(filter dev dev-init,$(MAKECMDGOALS)),)
+ifneq ($(filter dev dev-init setup run,$(MAKECMDGOALS)),)
 ifeq ($(strip $(PROFILE)),)
 $(error PROFILE is required. Example: make dev PROFILE=main)
 endif
 endif
 
-.PHONY: catalog-view catalog-pin catalog-update dev dev-init dev-list dev-local dev-desktop dev-runtime dev-server dev-mobile-auth dev-mobile-tunnel dev-web-auth server-db-up server-db-wait \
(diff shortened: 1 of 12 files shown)


pablonyx (commented):
Review pass from three spec-focused subagents plus my own Makefile read found a small set to fix before merging:

- Move/check local Postgres readiness only when the profile DB is actually used, so `DATABASE_URL=... make setup/run` does not require Docker Postgres.
- Make the clean-worktree contract explicit: `run` is intentionally rebuild-free, so first-time setup needs an explicit build step or preflight guidance before Vite-only dev scripts resolve dist-based workspace packages.
- Stale docs cleanup: `specs/developing/local/stripe-local-testing.md` still says “If you are not using profile-based `make dev`” later in the manual webhook section; update that to `make run` / profile launcher for consistency.

Makefile
```diff
@@ -81,12 +81,14 @@ endif
         db-migrate-up db-migrate-down \
         all clean
 
-# --- Dev (builds SDK, starts runtime + desktop together) ---
+# --- Profile dev (setup, build, and run are separate) ---
 
-dev: sdk-build server-db-ready
+dev: setup run
+
+run: server-db-ready
```
pablonyx: Can we move `server-db-ready` inside the profile-DB branch instead of keeping it as a target prerequisite? With `DATABASE_URL` set, `make run`/`make setup` should not require local Docker Postgres; otherwise the documented DB override can fail before reaching the bypass logic.

apps/web/package.json
```diff
@@ -4,7 +4,8 @@
   "version": "0.1.0",
   "type": "module",
   "scripts": {
-    "dev": "pnpm --filter @proliferate/cloud-sdk build && pnpm --filter @proliferate/cloud-sdk-react build && pnpm --filter @proliferate/design build && pnpm --filter @proliferate/product-domain build && pnpm --filter @proliferate/ui build && pnpm --filter @proliferate/product-ui build && pnpm --filter @proliferate/product-surfaces build && vite",
+    "dev": "vite",
```
pablonyx, line 7: This makes `pnpm dev` rebuild-free, but the app imports still resolve workspace packages through their built `dist` export maps. On a clean worktree after `make setup`, `make run` will invoke this before `make build-frontend` has produced packages like `@proliferate/ui` and `@proliferate/cloud-sdk`. Could we either resolve workspace packages to source in dev, or make the docs/preflight contract require an explicit frontend build before `run`?

specs/developing/local/README.md
```diff
@@ -62,8 +62,8 @@ deployment variable ownership.
 From a repo worktree:
 
 ```bash
-make dev-init PROFILE=<name>
-make dev PROFILE=<name>
+make setup PROFILE=<name>
```
pablonyx, line 65: After the setup/run split, this quick start no longer works from a truly clean worktree unless artifacts already exist: `make run` exits when `target/debug/anyharness` is missing, and Vite-only dev scripts expect built workspace package output. Please add the required first-time build step and mirror that in the root quick starts.

Makefile
pablonyx: Fixed in `7a1cfd4`: `run` and `setup` no longer depend on `server-db-ready`; profile DB readiness/creation now lives behind conditional helpers, and `DATABASE_URL=... make -n setup/run` omits the profile DB readiness/creation path.

apps/web/package.json
pablonyx: Fixed in `7a1cfd4`: added `dev-artifacts-ready` as a no-build preflight that checks the runtime binary and frontend package `dist` artifacts, then prints `make build`, `make build-rust`, or `make build-frontend` as appropriate. The docs now call out the first-clean-worktree build step explicitly.

specs/developing/local/README.md
pablonyx: Fixed in `7a1cfd4`: local dev, README, CONTRIBUTING, AGENTS, mobile, and spec catalog quick starts now include the first-clean-worktree `make build` step, while preserving `setup` + `run` as the fast restart path once artifacts exist.

PR #791, opened 2026-06-25
feat(desktop): support native plan decision options

pablonyx:
## What

Support native plan-mode decisions in the composer.

When an agent emits a native plan permission, Desktop now replaces the normal composer with a plan decision panel that renders the exact harness-emitted options in order. Feedback options are driven by runtime option presentation metadata, not Desktop-side copy matching.

## How

- Added permission option presentation metadata to the AnyHarness contract, SDK, SDK React hooks, and agent registry.
- Added a native plan option resolution endpoint so Desktop can resolve the exact emitted option id.
- Added Codex/Claude registry metadata for feedback-style plan options.
- Added `PlanDecisionComposerPanel` as a composer replacement for linked pending plan permissions.
- Kept proposed plan transcript cards as immutable plan artifacts, with native approve/reject/feedback decisions handled in the composer panel.
- Added focused tests for option parsing, plan decision entry mapping, feedback prompt session targeting, and updated runtime fixtures.

## Known Issues

- There is a brief UI flash where the generic permission approval card can appear before `PlanDecisionComposerPanel` replaces the composer. This happens because the frontend can receive/render the pending permission before transcript state has enough plan linkage to classify it as a plan decision. The generic card shows the same underlying options and disappears once the permission is linked to the proposed plan.
- The older binary approve/reject plan API endpoints are still present and remain used by other domains. This PR adds the exact native-option resolution path needed for native plan decisions, but does not remove or migrate the existing approve/reject endpoint callers.

## Verification

- `pnpm --dir apps/desktop exec tsc --noEmit`
- `pnpm --dir apps/desktop exec vitest run src/hooks/chat/derived/use-active-plan-decision.test.tsx src/lib/domain/chat/composer/chat-input-helpers.test.ts src/lib/domain/chat/composer/plan-decision-options.test.ts`
- `cargo test -p anyharness-lib permission_option_mappings --lib`
- `cargo test -p anyharness-lib approved_native_plan_failure_cancels_pending_permission --lib`
- `cargo test -p anyharness-lib links_plan_to_existing_pending_permission_by_tool_call_id --lib`
- `git diff --check origin/main...HEAD`


pablonyx:
Closing this stale internal draft. Session controls and catalog reconciliation have since moved forward on current main (including #1127), so this is no longer a safe implementation lineage. Re-spec any still-missing plan-decision behavior from current main.

PR #798, opened 2026-06-27
feat(cloud): add GitHub App sandbox credential leases

pablonyx:
## Summary

- add GitHub App authorization storage, status/connect/callback/webhook APIs, and repo authority checks for managed cloud repos
- add a hidden worker-only GitHub credential lease endpoint plus worker refresh/materialization logic for short-lived sandbox Git credentials
- bake a small Git credential helper into the E2B template path and switch managed-sandbox repo materialization away from injected product OAuth tokens
- add Desktop/Web account settings UI for connecting or managing the Proliferate GitHub App
- document the sandbox GitHub auth model and regenerate the cloud SDK from the public OpenAPI schema without exposing worker lease secrets

## Why

Managed cloud sandboxes should never receive long-lived product OAuth authority. The control plane owns durable GitHub authority and repo policy; the worker periodically materializes disposable GitHub App user-to-server credentials; Git/AnyHarness consume those credentials through a dumb helper.

## Validation

- `DEBUG=1 uv run python -m compileall -q proliferate`
- `DEBUG=1 uv run alembic heads`
- `DEBUG=1 uv run --extra dev ruff check ...`
- `DEBUG=1 uv run --extra dev ruff format --check ...`
- `JWT_SECRET=test CLOUD_SECRET_KEY=test uv run pytest tests/unit/test_managed_sandbox_repo_materialization.py -q`
- `cargo fmt -p proliferate-worker --check`
- `cargo check -p proliferate-worker`
- `pnpm --filter @proliferate/cloud-sdk-react build`
- `pnpm --filter @proliferate/product-ui build`
- `pnpm --filter proliferate exec tsc --noEmit`
- `pnpm --filter @proliferate/web typecheck`
- `node --test scripts/ci-cd/detect-deploy-surfaces.test.mjs`
- local Git credential helper protocol smoke tests for token/no-token/override/www.github.com/wrong-host behavior

## Not Yet Live-Smoked

- full E2B template publish/smoke with the baked helper
- interactive GitHub App install/OAuth callback against the local profile


server/proliferate/server/cloud/github_app/models.py
```diff
@@ -0,0 +1,34 @@
...
+    "install",
+    "grant_repo_access",
+    "manage",
+]
+
+
+class GitHubAppStatusResponse(BaseModel):
+    connected: bool
+    github_login: str | None = Field(default=None, serialization_alias="githubLogin")
+    status: str | None = None
```
pablonyx: Because this response is exported through `@proliferate/cloud-sdk`, the generated public type for `status` is currently just `string | null` (and `installationState` below is the same), while Desktop/Web compare exact values and the sandbox auth spec defines a finite set. Please make these wire fields explicit `Literal`s before regenerating OpenAPI, and either remove `"manage"` from `GitHubAppStatusAction` or return it consistently. Otherwise clients can compile while mishandling statuses like `expired`/`needs_reauth`, which is exactly the path that should drive the “Reconnect GitHub App” UI.

server/proliferate/server/cloud/github_app/service.py
```diff
@@ -0,0 +1,283 @@
...
+            status_code=503,
+        )
+    base = base.rstrip("/")
+    path = auth_route_path_for_base("/auth/github-app/callback", base_url=base)
+    return f"{base}{path}"
+
+
+def _redirect_after_callback() -> str:
+    base = settings.frontend_base_url.strip() or settings.api_base_url.strip() or "/"
+    return base.rstrip("/") + "/settings/account" if base != "/" else "/"
```
pablonyx, line 54: This callback always sends the user to the Web account settings page. That works for the Web button, but Desktop also calls the same connect endpoint and opens the returned URL in the external browser, so after authorizing the app the user lands in Web while Desktop only updates if they return before the scheduled polling window expires. Please include a caller/return target in the signed state (for example Web settings vs a Desktop handoff/deep-link) and redirect accordingly; otherwise the Desktop Account settings flow can look like it never completed.

apps/desktop/src/components/settings/panes/AccountPane.tsx
```diff
@@ -258,6 +291,64 @@ export function AccountPane() {
...
+          loading: connecting,
+          disabled: connecting,
+          onClick: () => { void onConnect(); },
+        },
+  };
+}
+
+function scheduleGitHubAppStatusRefresh(refetch: () => void) {
+  for (const delayMs of [2_000, 5_000, 10_000, 20_000, 40_000, 80_000]) {
+    window.setTimeout(refetch, delayMs);
```
pablonyx: These timers are never cleared or tied to the user/auth scope that started the flow. If the Account pane unmounts, the user signs out, or they switch users during the GitHub browser flow, the old timeouts can continue refetching the old `githubAppStatus` query for up to 80s and leave stale status cache behind. Please own this polling in a cleanup-aware effect/ref or workflow helper, and cancel/guard it when the current user/cache scope changes.

server/proliferate/server/cloud/managed_sandboxes/service.py
```diff
@@ -758,11 +758,23 @@ async def ensure_managed_sandbox_repo_runtime_connection(
         user_id=user.id,
         repo_config=repo_config,
     )
+    await require_github_cloud_repo_authority(
+        db,
+        user_id=user.id,
+        git_owner=repo_config.git_owner,
+        git_repo_name=repo_config.git_repo_name,
+    )
     cached = _cached_repo_runtime_connection(cache_key)
```
pablonyx: This cache can return a runtime connection from a previous managed sandbox generation. The key is only user/repo config versions, and the cached hit returns before `ensure_managed_sandbox_ready()` or `ensure_repo_materialized()` re-check the active sandbox. If the user destroys/recreates or the runtime tuple changes inside the 60s TTL, clients can get the old `anyharnessWorkspaceId`/`runtimeGeneration` and then connect to a stale AnyHarness workspace. Please include the active `managed_sandbox_id`/`runtime_generation` in the cache key or validate the current sandbox generation before returning the cached connection, and add a generation-change test.

anyharness/crates/proliferate-worker/src/cloud_client/github_credentials.rs
```diff
@@ -0,0 +1,52 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub current_expires_at: Option<DateTime<Utc>>,
+}
+
+#[derive(Debug, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct GitHubCredentialLeaseResponse {
+    pub provider: String,
+    pub token_kind: String,
+    pub access_token: String,
```
pablonyx, line 22: This response is secret-bearing because it contains `accessToken`, but deriving `Debug` makes the token printable by any future `?lease`/error instrumentation in the worker's unsanitized console logs. The sandbox auth spec explicitly calls out that every refresh response call site must treat `accessToken` as secret-bearing; please remove `Debug` here or implement a custom redacted `Debug` before this DTO becomes reusable.

anyharness/crates/proliferate-worker/src/materialization/github_credentials.rs
```diff
@@ -0,0 +1,203 @@
...
+    pub actor_login: Option<String>,
+    pub actor_id: Option<String>,
+    pub lease_id: String,
+    pub issued_at: DateTime<Utc>,
+    pub expires_at: DateTime<Utc>,
+    pub refresh_after: DateTime<Utc>,
+}
+
+pub fn credential_paths() -> GitHubCredentialPaths {
+    let home = dirs::home_dir().unwrap_or_else(|| PathBuf::from("."));
```
pablonyx: If `HOME` is unavailable this silently falls back to `./.proliferate`, which can put the token/meta/helper paths under whatever cwd the supervisor used. For credential materialization this should fail closed instead of writing a provider token outside the contracted `/home/user/.proliferate/git/github.com` tree. Please return a materialization error when the home directory cannot be resolved.

scripts/build-template.mjs
```diff
@@ -331,6 +340,13 @@ function buildTemplateDefinition() {
     .copy("proliferate-supervisor", "/home/user/.proliferate/bin/proliferate-supervisor", {
       mode: 0o755,
     })
+    .copy(
+      "proliferate-git-credential-helper",
+      "/home/user/.proliferate/bin/proliferate-git-credential-helper",
+      {
+        mode: 0o700,
```
pablonyx, line 347: The template is still under `setUser("root")` when this copy runs, and the helper is installed with mode `0700`. That leaves `/home/user/.proliferate/bin/proliferate-git-credential-helper` root-owned and non-executable by the sandbox user that runs Git/worker flows, so repo materialization's `test -x "$helper"` will fail. Please copy/chown it as `user` or install it after switching to `setUser("user")`.

server/proliferate/server/cloud/managed_sandboxes/repo_materialization.py
```diff
@@ -257,25 +259,30 @@ async def _clone_or_update_repo(
         [
             "set -eu",
             "command -v git >/dev/null 2>&1",
+            'helper="$HOME/.proliferate/bin/proliferate-git-credential-helper"',
+            'token_file="$HOME/.proliferate/git/github.com/token"',
+            'test -x "$helper"',
+            'test -s "$token_file"',
```
pablonyx: `test -s` only proves that some token file exists; it does not prove the worker has converged a current GitHub App lease. A stale token can survive after `meta.json` expires/corrupts, and fresh sandbox reconciliation can race the first successful heartbeat before meta/token are written. Since the helper is dumb and will serve whatever is in this file, please gate materialization on `meta.json` as well (`provider`, `tokenKind`, `expiresAt`/`refreshAfter`) or add a bounded readiness wait for worker credential convergence before cloning.

scripts/smoke-cloud-template.mjs
```diff
@@ -117,6 +118,31 @@ async function main() {
       }
     }
 
+    const helperCheck = await sandbox.commands.run(
```
pablonyx, line 122: This smoke only calls the helper directly with a fake token. It still would not catch broken Git integration, missing global helper config, SSH `insteadOf` rewrites, or clone/fetch/push behavior through the helper. The sandbox auth spec lists those as template-smoke coverage; please extend this to exercise Git itself with an isolated HOME/global config (for example `git credential fill` plus clone/fetch, and a dry-run push when CI has a test token) so template deploys catch helper protocol/config regressions.

pablonyx (commented):
Credential-materialization blocker not tied to a changed hunk: the old product-OAuth path is still active alongside this new GitHub App helper path. `server/proliferate/server/cloud/worker/service.py:127-153` still enqueues `configure_git_identity` on worker enrollment, `server/proliferate/server/cloud/target_git_identity/service.py:152-158` still stores `github_grant.access_token` in the worker-fetched plan, and `anyharness/crates/proliferate-worker/src/materialization/git_identity.rs:74-99` still writes that token into a target-local credential store. Fresh provisioning also still loads `github_grant.access_token` in `server/proliferate/server/cloud/runtime/provisioning/input.py:86-212` and clones via a tokenized URL in `server/proliferate/server/cloud/runtime/git_operations.py:29-83`. The sandbox-auth contract says product OAuth tokens must not be materialized into managed sandboxes and the helper lease should be the Git credential source, so please remove/disable these legacy paths for managed-cloud targets as part of this migration.

anyharness/crates/proliferate-worker/src/cloud_client/github_credentials.rs
```diff
@@ -0,0 +1,52 @@
...
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GitHubCredentialLeaseRequest {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub current_lease_id: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub current_expires_at: Option<DateTime<Utc>>,
+}
+
+#[derive(Debug, Deserialize, Clone)]
```
pablonyx: This response is explicitly secret-bearing (`access_token` below), but deriving `Debug` will print the raw GitHub token anywhere the lease is logged with `?lease` or included in debug error context. The sandbox auth contract requires lease responses to be treated as secret-bearing at every call site. Please remove `Debug` here or implement a redacted `Debug`/secret wrapper for `access_token` before the DTO can be accidentally logged.

server/proliferate/server/cloud/managed_sandboxes/service.py
```diff
@@ -758,11 +758,23 @@ async def ensure_managed_sandbox_repo_runtime_connection(
         user_id=user.id,
         repo_config=repo_config,
     )
+    await require_github_cloud_repo_authority(
```
pablonyx, line 689: This seam changed from the product OAuth grant helper to `require_github_cloud_repo_authority`, but the runtime-connection unit tests were not updated. `DEBUG=true JWT_SECRET=test-secret uv run pytest -q tests/unit/test_managed_sandbox_runtime_connection.py tests/unit/test_managed_sandbox_repo_materialization.py` currently fails because `tests/unit/test_managed_sandbox_runtime_connection.py` still monkeypatches the removed `read_ready_github_grant_for_user` symbol. Please update those tests to patch/assert the GitHub App authority helper and the helper-backed materialization call; otherwise this path has no passing coverage for the new auth gate.

server/proliferate/db/store/github_app.py
```diff
@@ -0,0 +1,383 @@
...
+_CACHE_FRESHNESS_SECONDS = 600
+
+
+@dataclass(frozen=True)
+class GitHubAppAuthorizationValue:
+    id: UUID
+    user_id: UUID
+    github_user_id: str
+    github_login: str
+    access_token: str | None
```
pablonyx, line 32: This store snapshot carries decrypted GitHub App tokens, but the dataclass default `repr` will include both `access_token` and `refresh_token` if the value is ever logged or appears in a failing assertion. The sandbox GitHub auth spec explicitly calls out structured logs and test output as redaction boundaries. Please mark the token fields with `field(repr=False)` (or use a small secret wrapper) so the persistence boundary cannot accidentally print raw GitHub credentials.

server/proliferate/server/cloud/github_app/models.py
pablonyx: Fixed in 486406cc0: `status` and `installationState` are now explicit Literal-backed OpenAPI enums, `manage` was removed from the wire action enum, and the cloud SDK was regenerated. I also made `expired` drive the reconnect UI alongside `needs_reauth`.

server/proliferate/server/cloud/github_app/service.py
pablonyx: Fixed in 486406cc0: `/connect` accepts a signed/validated `returnTo` in state. Desktop now passes `proliferate://settings/account?source=github_app_callback`, and the desktop deep-link router maps that back to Account settings.

apps/desktop/src/components/settings/panes/AccountPane.tsx
pablonyx: Fixed in 486406cc0: the Desktop polling timers are tracked in a ref, cleared on unmount, and cleared when the auth cache scope changes.

server/proliferate/server/cloud/managed_sandboxes/service.py
pablonyx: Fixed in 486406cc0: cached repo runtime connections are now validated against the current personal managed sandbox runtime-access generation before returning. I added a generation-change unit test to catch stale AnyHarness ids.

anyharness/crates/proliferate-worker/src/cloud_client/github_credentials.rs
pablonyx: Fixed in 486406cc0: removed `Debug` from the secret-bearing Rust lease response DTO.

anyharness/crates/proliferate-worker/src/cloud_client/github_credentials.rs
pablonyx: Fixed in 486406cc0: removed `Debug` from the secret-bearing Rust lease response DTO.

anyharness/crates/proliferate-worker/src/materialization/github_credentials.rs
pablonyx: Fixed in 486406cc0: credential path resolution now returns a materialization error if `HOME` cannot be resolved, instead of falling back to `./.proliferate`.

scripts/build-template.mjs
pablonyx: Fixed in 486406cc0: the template build now `chown`s the helper to `user:user` and keeps mode `0700`; the smoke now checks `user:user:700` too.

server/proliferate/server/cloud/managed_sandboxes/repo_materialization.py
pablonyx: Fixed in 486406cc0: repo materialization now does a bounded readiness wait and validates `meta.json` provider/tokenKind/refreshAfter/expiresAt before clone/fetch. The helper remains dumb; the server-side clone path now refuses stale or missing worker lease metadata.

scripts/smoke-cloud-template.mjs
pablonyx: Partially addressed in 486406cc0: the template smoke now verifies helper ownership/mode and exercises Git itself through `git credential fill` with the configured helper. I left live clone/fetch/push coverage for the manual/live smoke because CI/template smoke does not have a safe GitHub test token in this path yet.

server/proliferate/server/cloud/managed_sandboxes/service.py
pablonyx: Fixed in 486406cc0: the runtime-connection tests now patch/assert `require_github_cloud_repo_authority` and helper-backed materialization. `DEBUG=true JWT_SECRET=test-secret CLOUD_SECRET_KEY=test-secret uv run pytest -q tests/unit/test_managed_sandbox_runtime_connection.py tests/unit/test_managed_sandbox_repo_materialization.py` is passing with 9 tests.

server/proliferate/db/store/github_app.py
pablonyx: Fixed in 486406cc0: the decrypted GitHub App auth store value is now `repr=False`; I also made the integration authorization dataclass repr-safe and marked the Pydantic worker lease `access_token` field `repr=False`.

pablonyx:
Review follow-up on the legacy product-OAuth materialization path: I did not remove the old target/cloud-command git identity flow in this PR because it is the broader legacy target runtime path (`runtime/provision*`, `target_git_identity`, worker `git_identity`) rather than the managed-sandbox repo materialization path introduced here.

This PR now gates managed-sandbox repo runtime resolution through GitHub App authority and clones/fetches via the worker/helper lease. Deleting the older command-era target path should happen in the planned worker/cloud-command cleanup PR so we can preserve non-managed/legacy target behavior until that migration lands.

PR #801, opened 2026-06-27
feat(cloud): add repo environments

pablonyx:
## Summary

Adds the RepoConfig / RepoEnvironment control-plane model for local and cloud repository environments.

- introduces cloud repository and repo environment models, store helpers, API routes, and migration
- moves workspace-scoped secrets/materialization toward `repo_environment_id` ownership while preserving PR2 compatibility
- adds cloud SDK and SDK React helpers for repositories/environments
- wires Desktop settings save flow to persist local repo environments with a stable desktop install id
- updates generated cloud SDK types and targeted tests

## Why

This separates repository identity from concrete environments so local checkouts and cloud environments can share one repo config while still tracking environment-specific branch/setup/run/secrets state cleanly.

## Validation

- `git diff --check`
- `uv run python -m compileall -q ...`
- `DEBUG=1 JWT_SECRET=test-secret uv run pytest -q tests/integration/test_schema_migrations.py tests/unit/test_cloud_secrets_models.py tests/unit/test_managed_sandbox_repo_materialization.py tests/unit/test_cloud_orm_models.py tests/integration/test_cloud_api.py::TestCloudRepoConfig`
- `pnpm --filter @proliferate/cloud-sdk-react build`
- `pnpm --filter @proliferate/product-surfaces test -- CloudEnvironmentsSettingsSurface.test.tsx`
- `pnpm --filter @proliferate/product-surfaces typecheck`
- `pnpm --filter proliferate build`

### DIFF
--- a/apps/desktop/src/hooks/settings/workflows/use-repository-settings.ts
+++ b/apps/desktop/src/hooks/settings/workflows/use-repository-settings.ts
@@ -3,7 +3,7 @@ import type { GitBranchRef } from "@anyharness/sdk";
 import {
   useRepoRootGitBranchesQuery,
 } from "@anyharness/sdk-react";
-import { useSaveLocalRepoEnvironment } from "@proliferate/cloud-sdk-react";
+import { useRepoConfigs, useSaveLocalRepoEnvironment } from "@proliferate/cloud-sdk-react";
 import {
   buildLocalEnvironmentSavePatch,
   isLocalEnvironmentDraftDirty,
@@ -26,6 +26,8 @@ export function useRepositorySettings(repository: SettingsRepositoryEntry | null
   );
   const setRepoConfig = useRepoPreferencesStore((state) => state.setRepoConfig);
   const saveLocalEnvironment = useSaveLocalRepoEnvironment();
+  const repoConfigs = useRepoConfigs(Boolean(repository?.gitOwner && repository.gitRepoName));
+  const [desktopInstallId, setDesktopInstallId] = useState<string | null>(null);
   const { data: branchRefs = EMPTY_BRANCHES } = useRepoRootGitBranchesQuery({
     repoRootId: repository?.repoRootId ?? null,
     enabled: !!repository,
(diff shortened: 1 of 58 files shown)


server/proliferate/server/cloud/managed_sandboxes/materialization/repos.py
```diff
@@ -164,6 +164,7 @@ async def ensure_repo_materialized(
         db,
         managed_sandbox_id=sandbox.id,
         cloud_repo_config_id=repo_config.id,
+        repo_environment_id=repo_config.id,
```
pablonyx: This assumes every configured `CloudRepoConfig` has a mirrored `RepoEnvironment` with the same id, but several legacy writers still create/ensure `cloud_repo_config` directly without calling `sync_cloud_environment_from_legacy_cloud_repo_config` (for example the automation repo bootstrap path calls the store helper directly). Those rows will hit the new FK when `begin_repo_materialization(..., repo_environment_id=repo_config.id)` runs, so materialization can fail before clone/secrets. Please either route all legacy writers through the syncing service, move the mirror upsert into the store/DB boundary, or resolve/create the cloud `RepoEnvironment` before passing this FK.

server/alembic/versions/d4e7f8a9b2c3_repo_environments.py
```diff
@@ -0,0 +1,494 @@
...
+                    id, repo_config_id, environment_kind, desktop_install_id, local_path,
+                    configured, configured_at, default_branch, setup_script,
+                    setup_script_version, run_command, config_version,
+                    legacy_cloud_repo_config_id, created_at, updated_at, deleted_at
+                )
+                SELECT
+                    id, id, 'cloud', NULL, NULL,
+                    configured, configured_at, default_branch, setup_script,
+                    setup_script_version, run_command,
+                    GREATEST(files_version, env_vars_version, setup_script_version),
```
pablonyx: This backfill omits the configured/default-branch/run-command state from `config_version`. A migrated repo with `configured = true` but no file/env/setup changes gets `config_version = 0`, while the runtime sync path treats configured legacy rows as at least version 1. That makes migrated API output inconsistent and can hide existing desired state from clients/materializers. Please include the configured state here too, e.g. `GREATEST(files_version, env_vars_version, setup_script_version, CASE WHEN configured THEN 1 ELSE 0 END)`.

server/proliferate/db/store/repositories.py
```diff
@@ -0,0 +1,426 @@
...
+                    RepoConfig.git_provider == "github",
+                    RepoConfig.git_owner == legacy.git_owner,
+                    RepoConfig.git_repo_name == legacy.git_repo_name,
+                    RepoConfig.deleted_at.is_(None),
+                )
+            )
+        ).scalar_one_or_none()
+
+    if repo is None:
+        repo = RepoConfig(
```
pablonyx: This mirror path is select-then-insert, so it is not idempotent under concurrent first saves. The legacy `save_cloud_repo_config` path uses `ON CONFLICT`, but two requests can both miss the new `RepoConfig`/`RepoEnvironment` mirror and one will fail on the PK/unique constraint at flush. The local environment writer above has the same pattern. Please use PostgreSQL upserts (then reload the row) for both `RepoConfig` and `RepoEnvironment` creation so the new mirror keeps the race-safety the legacy endpoint already had.

pablonyx (commented):
Focused server/db/API review for the RepoConfig/RepoEnvironment stack. I left inline comments only where I think there is an actionable correctness risk.

server/proliferate/server/cloud/managed_sandboxes/materialization/service.py
```diff
@@ -148,18 +148,12 @@ async def _materialize(fresh_db: AsyncSession) -> None:
         )
 
         sandbox = await ensure_managed_sandbox_ready(fresh_db, user)
-        materialization = await repos.ensure_repo_materialized(
+        await repos.ensure_repo_materialized(
```
pablonyx: After a workspace secret update, this scheduled job now only calls `ensure_repo_materialized`. If the repo row is already fresh, `ensure_repo_materialized` returns from its fast path before it reaches `materialize_workspace_secrets`, so the updated workspace secret files/env never get written and the `ManagedSandboxSecretMaterialization` row stays stale/pending. Please keep a direct workspace-secret materialization call here, or make the repo-materialization fast path still run the secret writer.

apps/desktop/src/hooks/settings/workflows/use-repository-settings.ts
```diff
@@ -90,12 +93,32 @@ export function useRepositorySettings(repository: SettingsRepositoryEntry | null
     }
     const nextConfig = buildLocalEnvironmentSavePatch(state.draft);
     setRepoConfig(sourceRoot, nextConfig);
+    if (repository?.gitOwner && repository.gitRepoName) {
+      const { gitOwner, gitRepoName, gitProvider } = repository;
+      void (async () => {
+        const { installId } = await loadAnonymousTelemetryBootstrap();
+        await saveLocalEnvironment.mutateAsync({
```
pablonyx, line 150: This new local-environment write is still effectively write-only from the frontend. The shared settings surface, Web home, Mobile home, and Mobile settings all continue to read `useCloudRepoConfigs()` / `CloudRepoConfigsListResponse` rather than the new `useRepoConfigs()` aggregate, so a successful local save here never becomes observable through the migrated RepoConfig/RepoEnvironment client path. Please either wire the environment consumers to `RepoConfigResponse.environments` in this PR, or defer this cloud write until the read path lands with tests that prove local and cloud environment saves show up from the new repository list.

PR #809, opened 2026-07-01
refactor(cloud): cut over workspace flow to gateway

pablonyx:
## Summary

- Replace the active cloud workspace API with the lightweight `RepoEnvironment` + `CloudWorkspace` flow.
- Create cloud workspaces by materializing the repo environment, creating the AnyHarness worktree, and storing only the product workspace row.
- Cut Desktop and Web runtime resolution over to the gateway using the AnyHarness workspace id from `CloudWorkspace`.
- Remove the old cloud-sandbox runtime-connection endpoint/helper surface.

## Verification

- `cd server && DEBUG=true uv run python -c 'import proliferate.main; print("server import ok")'`
- `cd server && DEBUG=true uv run python -m compileall -q proliferate`
- `pnpm --filter @proliferate/cloud-sdk-react typecheck`
- `pnpm --filter @proliferate/web exec tsc -p tsconfig.json --noEmit`
- `pnpm --filter proliferate exec tsc --noEmit`
- `pnpm --filter @proliferate/web exec vitest run src/lib/access/cloud/create-workspace-with-transient-recovery.test.ts`
- `pnpm --filter proliferate exec vitest run src/lib/access/cloud/cloud-sandbox-gateway.test.ts src/lib/domain/workspaces/cloud/cloud-workspace-creation.test.ts src/lib/domain/workspaces/mobility/mobility-footer-context.test.ts src/hooks/main/workflows/use-main-screen-actions.test.tsx`


server/proliferate/server/cloud/workspaces/service.py
```diff
@@ -1,280 +1,463 @@
...
+    repo_root = await _resolve_repo_root(
+        runtime_url,
+        runtime_token,
+        repo_path=repo_path,
     )
-    workspaces = sorted(
-        [*workspaces, *claimed_workspaces],
-        key=lambda workspace: workspace.updated_at,
-        reverse=True,
+    anyharness_workspace = await _create_anyharness_worktree(
```
pablonyx, line 208: This direct AnyHarness worktree creation still bypasses the managed launch/command flow. The workspace-provisioning spec requires Cloud to create the durable `cloud_workspace` row plus exposure/projection state before materialization is queued, and worker results must echo `cloud_workspace_id` back through the command ledger. Calling AnyHarness inline here skips command preflights/wake/result fencing and leaves no recoverable Cloud command if the request dies mid-create. Please route this through the server-owned managed launch path or enqueue materialization after persisting the product row.

server/proliferate/db/store/cloud_workspaces.py
```diff
@@ -40,157 +33,25 @@ async def list_cloud_workspaces(
 ) -> list[CloudWorkspace]:
```
pablonyx: This new store returns ORM `CloudWorkspace` rows directly to services. The server database guide says stores are the ORM boundary: they should return frozen dataclasses, and ORM objects must not leave `db/store/**`. Please introduce a `CloudWorkspaceSnapshot`/write result type and keep ORM mutation/query details inside this store; otherwise the new workspace service and response model are built on a stale ORM-as-domain path.

pablonyx:
Remote-visible blockers from the current PR checks and cutover review:

- PR Metadata fails because the non-draft PR has no labels. Add exactly one `release:*` label, likely `release:maintenance` for this refactor, and at least one relevant `area:*` label such as `area:server`, `area:cloud`, `area:desktop`, `area:sdk`, and/or `area:product`.
- Server CI test fails before tests run: `server/tests/e2e/cloud/helpers/providers.py:5` still imports `proliferate.db.models.cloud.runtime_environments.CloudRuntimeEnvironment`, but that model path has been deleted. Remove the stale helper path or retarget it to the new sandbox/workspace model.
- Repo shape fails on line limits and stale allowlist drift: `cloud/sdk/src/types/generated.ts:770`, `apps/desktop/src/lib/domain/workspaces/sidebar/sidebar.test.ts:646`, `server/proliferate/constants/cloud.py:660` versus allowlist 614, `server/proliferate/server/cloud/secrets/api.py:416`, and `server/tests/integration/test_cloud_api.py:2837` versus allowlist 2576. The allowlist also still references missing files including `server/proliferate/server/cloud/managed_sandboxes/service.py`, `server/tests/integration/test_cloud_agent_auth_api.py`, `server/tests/integration/test_cloud_commands_api.py`, `server/tests/integration/test_sandbox_profile_foundation.py`, and `server/tests/unit/test_cloud_runtime_provision.py`.
- Shared frontend packages fail in `apps/packages/product-domain/src/settings/cloud-settings.test.ts:18` and `:31`: the model now returns `personal-secrets` and `organization-secrets` for web/mobile, but the tests still expect those sections to be absent. Either update the product-domain contract/tests intentionally or keep those sections out for those surfaces.
- The central cloud router still carries a long parked legacy route map in comments. For this cutover, please remove dead/commented legacy routes rather than leaving parked shims beside the active workspace gateway path.

I also left inline comments on the direct AnyHarness create path and ORM rows escaping the new workspace store.

server/proliferate/server/cloud/workspaces/service.py
```diff
@@ -1,280 +1,463 @@
...
+            (
+                "Workspace display name cannot exceed "
+                f"{MAX_CLOUD_WORKSPACE_DISPLAY_NAME_CHARS} characters."
+            ),
             status_code=400,
         )
 
-    workspaces = await list_cloud_workspaces_store(db, user_id, lifecycle=lifecycle)
-    claimed_workspaces = await list_claimed_organization_workspaces_for_user(
+    await materialization_service.materialize_repo_environment(
```
pablonyx, line 182: `materialize_repo_environment()` is not a pure store/service call: it runs sandbox/provider work and commits the injected `AsyncSession` internally. Calling it in the HTTP workspace-create request splits the transaction before the workspace create flow has finished, so later failures in repo-root resolution, AnyHarness worktree creation, or the final DB write cannot roll back the materialization side effects. Please move this to an after-commit/background materialization path (or otherwise make the transaction boundary explicit outside this request service) so workspace creation does not hide a commit inside the middle of the orchestration.

server/proliferate/server/cloud/workspaces/service.py
```diff
@@ -1,280 +1,463 @@
...
         db,
-        user_id=user_id,
-        lifecycle=lifecycle,
+        repo_environment_id=repo_environment.id,
+    )
+    runtime_url, runtime_token, _data_key = await _load_ready_runtime_access(db, user_id=user.id)
```
pablonyx, line 201: This path loads the sandbox base URL plus decrypted bearer token and then passes them directly to the AnyHarness integration below. That bypasses the gateway runtime-access invariant for this refactor: product code should authorize through the Cloud gateway and let the gateway own the upstream token/base URL, not hand decrypted sandbox credentials to workspace orchestration. Please route the worktree calls through the gateway-backed AnyHarness client/path (or an internal gateway service boundary) instead of calling the sandbox-local runtime directly.

pablonyx:
Additional stale-import sweep for the model cleanup: this PR still leaves active repo files referencing deleted runtime-environment fields/modules, beyond the E2E helper that CI trips first. Examples: `server/proliferate/db/store/cloud_runtime_environments.py:17` imports the missing `proliferate.db.models.cloud.runtime_environments.CloudRuntimeEnvironment`, `server/proliferate/db/store/cloud_sync/events.py:14` does the same, and tests/helpers still reference removed `CloudWorkspace.active_sandbox_id` / `runtime_environment_id` fields (`server/tests/e2e/cloud/helpers/workspaces.py:251`, `server/tests/e2e/cloud/test_lifecycle.py:78`, plus integration tests). These are hidden partly by parked route imports, but the cutover should delete or retarget the stale stores/tests instead of relying on unmounted code paths.

pablonyx:
Remote call graph issue for the gateway cutover:

The PR head still exposes/uses Cloud SDK helpers for unmounted Cloud routes.

* `cloud/sdk/src/client/sessions.ts:31-60` still calls `/v1/cloud/workspaces/{workspace_id}/snapshot`, `/v1/cloud/sessions/{session_id}/snapshot`, and `/v1/cloud/sessions/{session_id}/transcript`. `server/proliferate/server/cloud/api.py` leaves live/events routers parked, and generated OpenAPI no longer contains these paths. Web/Mobile still consume this through `useCloudWorkspaceSnapshot`, so chat/session/sidebar reads will 404 after the cutover.
* `cloud/sdk/src/client/commands.ts:11-25` still calls `/v1/cloud/commands` and `/v1/cloud/commands/{command_id}`, but the commands router is parked/unmounted. `useEnqueueCloudCommand` remains reachable from Web/Mobile chat flows.
* The remote PR still has the old target-launch/remote-access SDK/workflow path (`cloud/sdk/src/client/workspaces.ts`, `cloud/sdk-react/src/hooks/workspaces.ts`, Web/Mobile home submit hooks) while `target_launch`/`remote_access` routers are not mounted. My local checkout now has unpushed edits that appear to remove this, but the remote PR still needs that hard cut pushed.

Please delete or retarget these hand-authored SDK/react hooks to mounted gateway/AnyHarness/workspace endpoints before landing; otherwise the frontend can compile while calling dead routes.

server/proliferate/server/cloud/workspaces/service.py
```diff
@@ -1,280 +1,474 @@
...
         raise CloudApiError(
-            "invalid_workspace_scope",
-            "Unsupported workspace scope.",
+            "invalid_display_name",
+            (
+                "Workspace display name cannot exceed "
+                f"{MAX_CLOUD_WORKSPACE_DISPLAY_NAME_CHARS} characters."
+            ),
             status_code=400,
         )
```
pablonyx, line 180: This insert still relies on a preflight branch set, but the DB enforces `ux_cloud_workspace_active_repo_environment_branch` after that set is read. Two concurrent creates can choose the same explicit branch and surface an unhandled integrity error, and two generated-name creates can pick the same suffix instead of retrying. Please catch the unique violation around workspace creation, retry generated-name resolution with the collided branch included, and return `cloud_branch_already_exists` for explicit collisions.

server/proliferate/server/cloud/workspaces/service.py
```diff
@@ -1,280 +1,474 @@
...
+    workspace = await _load_user_workspace(db, user_id=user_id, workspace_id=workspace_id)
+    workspace = await cloud_workspace_store.archive_cloud_workspace(db, workspace)
+    return await _workspace_payload(db, workspace, detail=True)
+
+
+async def restore_cloud_workspace_for_user(
+    db: AsyncSession,
+    user_id: UUID,
+    workspace_id: UUID,
+) -> WorkspaceDetail:
```
pablonyx, line 269: Restore can also hit the active-branch unique index deterministically: archive branch `x`, create a new active workspace for `x`, then restore the old one. `restore_cloud_workspace` clears `archived_at` without checking for an active `(owner_user_id, repo_environment_id, git_branch)` sibling, so the request becomes a 500 instead of the structured restore conflict required by the lifecycle spec. Please preflight or catch this conflict and return a product error.

server/proliferate/server/cloud/workspaces/service.py
```diff
@@ -1,280 +1,474 @@
...
+    if sandbox.status in {"paused", "stopped"}:
+        return "paused"
+    if sandbox.status in {"error", "destroyed"}:
+        return "error"
+    return "pending"
+
+
+def _worktree_path(repo_environment: RepoEnvironmentValue, branch_name: str) -> str:
+    return (
+        f"{materialization_paths.SANDBOX_WORKSPACE_ROOT}/worktrees/"
```
pablonyx, line 544: This worktree path is not collision-safe because `_branch_path_segment()` maps distinct branches such as `feature/foo`, `feature-foo`, and `feature@foo` to the same directory. The branch conflict check uses the raw branch name, so a second workspace can collide only after AnyHarness sees the path; the 409 fallback may then resolve the existing checkout and attach or attempt to attach the wrong runtime workspace. Please include a collision-resistant component such as the Cloud workspace id or a hashed/reversible branch segment in the path.

pablonyx (commented):
Fresh server/data/workspace-flow pass.

pablonyx:
Fresh stale deleted-model import sweep at current head `bdbe3b7`: `import proliferate.main` passes, but ordinary imports of still-present server/test modules fail because the PR deleted model modules without removing all callers. Verified failures include `tests.integration.test_cloud_api` -> missing `proliferate.db.models.cloud.repo_config`, `server.cloud.workspaces.target_launch.service` -> missing `cloud.commands`, `server.cloud.workspaces.provisioning.service` -> missing `cloud.exposures`, `db.store.cloud_runtime_environments` / `cloud_sync.events` / `billing.reconciler` / `runtime.credentials.auth_status` -> missing `cloud.runtime_environments`, and `test_cloud_workspace_lifecycle_store` -> removed `SandboxProfile` from `cloud.agent_auth`. Please delete/park these old stores, services, and tests or retarget them to the new `repositories`, `sandboxes`, and lightweight `cloud_workspace` model so test collection and direct imports do not depend on deleted models.

cloud/sdk/src/client/workspaces.ts
```diff
@@ -172,56 +118,24 @@ export async function createCloudWorkspace(
...
 export async function startCloudWorkspace(
   workspaceId: string,
   client: ProliferateCloudClient = getProliferateClient(),
 ): Promise<CloudWorkspaceDetail> {
-  const data = await client.requestJson<CloudWorkspaceTransport>({
-    method: "POST",
-    path: `/v1/cloud/workspaces/${encodeURIComponent(workspaceId)}/start`,
-  });
-  return normalizeCloudWorkspace(data) as CloudWorkspaceDetail;
+  const workspace = await getCloudWorkspace(workspaceId, client);
```
pablonyx: `startCloudWorkspace` is still used by Desktop start/retry flows, but this now only re-fetches the workspace row. A retry after `workspace_not_ready` or the status-screen "start" action can therefore report `cloud_workspace_started` without waking/ensuring the sandbox or materializing a missing `anyharnessWorkspaceId`. Please either remove/rename this as a refresh and update callers, or wire it to the actual wake/runtime recovery path so the public helper matches its behavior.

pablonyx:
SDK contract cleanup is still incomplete after the gateway cutover: `cloud/sdk/src/types/index.ts` still exports `commands.js`, `sessions.js`, and `live.js`, and `types/commands.ts` / `types/live.ts` preserve `CloudCommand*` and cloud-live projection shapes even though the client modules, mounted routes, and generated OpenAPI schemas were removed or parked. Web/Mobile still import those types as if they were server-backed (for example `apps/web/src/lib/access/cloud/cloud-command-status.ts` and `apps/mobile/src/lib/access/cloud/pending-mobile-prompt-types.ts`). Please delete or privatize the removed cloud-command/live contract types, retarget remaining view models to AnyHarness SDK types, and tighten the `Schema<Name>` fallback so removed schemas do not silently become `any`.

server/proliferate/server/cloud/api.py
```diff
@@ -74,8 +81,7 @@
 router.include_router(github_app_organization_router)
 router.include_router(secrets_router)
 router.include_router(cloud_sandboxes_router)
-# WORKSPACES PARKED: /v1/cloud/workspaces/* is intentionally disabled.
-# router.include_router(workspaces_router)
+router.include_router(workspaces_router)
```
pablonyx, line 32: Now that workspace routes are remounted, the claim API still needs to come along or the active claim UI should go away. `claimCloudWorkspace` still posts to `/v1/cloud/workspaces/{id}/claim`, and the Web claim header/footer, Mobile claim actions, and Desktop attached-panel claim mutation still call it. The claims router remains parked just below this line, so those buttons 404. Please either remount/retarget the claim endpoint for the new workspace model, or remove/disable the claim UI/hooks until claiming is intentionally unsupported.

apps/web/src/hooks/chat/facade/use-web-cloud-chat-data.ts
```diff
@@ -33,49 +38,81 @@ export function useWebCloudChatData() {
...
+  };
+  const transcriptQuery = {
+    data: undefined,
+    isLoading: sessionEventsQuery.isLoading,
+    isFetched: sessionEventsQuery.isFetched,
+    isError: sessionEventsQuery.isError,
+    refetch: sessionEventsQuery.refetch,
+  };
+  const transcriptItems = EMPTY_TRANSCRIPT_ITEMS;
+  const pendingInteractions = EMPTY_PENDING_INTERACTIONS;
```
pablonyx: This hardcodes away all pending interactions after the snapshot/live cutover. The AnyHarness event reducer still records `transcript.pendingInteractions` from `interaction_requested`, but Web now passes `[]` into `buildCloudTranscriptView`; Mobile has the same shape because `transcriptQuery.data` is always undefined. As a result permission/input requests lose the pending request metadata used by the header/sheet/actions, so approvals cannot surface or resolve correctly. Please derive the `CloudPendingInteraction[]` from the reconstructed AnyHarness transcript state, or plumb SDK pending interactions into `buildCloudTranscriptView`, instead of using the empty constant.

pablonyx (commented):
Fresh cutover pass: two actionable issues I think should be fixed before this lands.

apps/web/src/hooks/chat/facade/use-web-cloud-chat-data.ts
```diff
@@ -1,81 +1,124 @@
...
+  };
+  const transcriptQuery = {
+    data: undefined,
+    isLoading: sessionEventsQuery.isLoading,
+    isFetched: sessionEventsQuery.isFetched,
+    isError: sessionEventsQuery.isError,
+    refetch: sessionEventsQuery.refetch,
+  };
+  const transcriptItems = EMPTY_TRANSCRIPT_ITEMS;
+  const pendingInteractions = useMemo(
```
pablonyx, line 112: This fixes the empty-array issue only for the initial `sessions.list` result, but Web still does not keep this source fresh. `pendingInteractions` now comes from `session.executionSummary`, which is populated by `sessionsQuery`; that query has no `refetchInterval` and `transcriptQuery.refetch` is wired to `sessionEventsQuery.refetch`, so the active chat can receive an `interaction_requested` event via the 1s event poll while `pendingInteractions` stays stale. The header/transcript projection still won't see the pending input metadata until a manual remount/focus refetch. Please either derive pending interactions from the polled AnyHarness event reducer/transcript state, or make/refetch the sessions query whenever the active session events update and after interaction resolution.

pablonyx (commented):
Re-review after b0aa2240a: claim/start cleanup looks resolved; one Web pending-interaction blocker remains.

cloud/sdk/src/types/generated.ts
```diff
@@ -242,6 +243,13 @@ export interface CloudWorkspaceSummary {
 export interface CloudWorkspaceDetail extends CloudWorkspaceSummary {
   [key: string]: unknown;
 }
+export interface CloudWorkspaceRuntimeStatusResponse {
+  workspaceId: string;
+  status: CloudWorkspaceStatus;
+  runtimeStatus: CloudRuntimeStatus;
+  sandboxStatus?: string | null;
+  anyharnessWorkspaceId: string;
```
pablonyx: This hand-authored response still drifts from the generated/server contract: `CloudWorkspaceRuntimeStatusResponse.anyharnessWorkspaceId` is `string | null`/optional in OpenAPI and can be `None` while the workspace is still materializing, but the public SDK narrows it to `string`. Please alias this response to `Schema<"CloudWorkspaceRuntimeStatusResponse">` or make the field nullable so SDK consumers cannot accidentally treat a not-ready workspace as having a runtime id.

server/proliferate/server/cloud/workspaces/service.py
pablonyx: Still blocking at `4ebef3bc`: because `materialize_repo_environment()` commits the shared session here, the new `cloud_workspace` row is durable before lines 201-219 attach `anyharness_workspace_id`. If repo-root resolution, AnyHarness worktree creation, or the final update fails, the active branch is left permanently reserved with `status=materializing` and no mounted endpoint resumes the attach; Desktop can only poll/refresh that stuck row. Please either make the post-row materialization/attach path recoverable/idempotent (so polling or retry can finish it) or mark/delete the row on post-materialization failure so retry can create a working mounted workspace.

PR #817, opened 2026-07-01
feat(chat): transcript trailing-status + turn-stopped presentation

pablonyx:
## What

Folds in the transcript behavior work that was sitting uncommitted in the main checkout (preserved verbatim, rebased onto main):

- Live trailing-status: grace window 750ms → 150ms; removed the `isStreaming` gate so status shows during working non-streaming phases
- New `turn-stopped-presentation` domain module + tests
- Transcript turn chrome/row updates + playground status-transcript coverage
- Chat dock inset simplification; layout-shift debug helper (dev)
- Specs: chat-transcript update + new agent-auth-litellm primitive spec

## Verification
- product-domain + product-ui build clean; desktop `tsc` 0 errors
- product-domain transcript tests: 15 files / 111 tests pass

## Relation to the UX stack
Independent of the Wave 1–3 restyle PRs (#810/#811/#812/#813/#815/#816) — this is behavior, those are presentation. Wave 3a deliberately avoided these files so both sides merge cleanly in either order.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,6 @@ package-lock.json
 *.p8
 .probe-secrets.env
 .auth-env/.env.*
+
+# Standalone design-system workspace (UI scratch, not wired into build)
+design-system/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4266,7 +4266,7 @@ dependencies = [
 
 [[package]]
 name = "proliferate"
-version = "0.2.24"
+version = "0.2.25"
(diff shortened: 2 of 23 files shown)


pablonyx:
Superseded: the settling-window fix landed via #1099, which resolves the final-text-but-still-iterating stall at the status-derivation layer. This branch's residual diff inverts the trailing-indicator rule now on main, so nothing to fold in.

PR #847, opened 2026-07-02
feat(desktop): give each subagent a unique identicon avatar

Julian0444:
## Problem

Every subagent renders the same Robot glyph, tinted with a hash-picked color that itself repeats often (8 tokens, birthday paradox). With 4-6 siblings deployed, two tabs routinely look identical at a glance.

## Solution

Each subagent now draws a deterministic 5×5 mirrored pixel-grid identicon (2^15 shapes), seeded from the same seed as its name and color, so the same agent shows the same fingerprint on every surface: header tab, hover card, transcript tool/creation rows, composer popover and single-agent trigger.

- Pure domain module (`identicon.ts`) + one small SVG component; no new dependencies (mirrors the existing hand-rolled icon idiom).
- Cells paint with `fill="currentColor"`, so the existing `text-delegated-agent-N` plumbing and the tab status dot (`bg-current`) keep working unchanged.
- Shape bits chain the existing avalanche mix twice, staying decorrelated from the name and color indices; near-empty/full grids re-derive so every icon stays legible at 12-16px.
- The 3 generic Robots (expand toggle, group headers) and the parent tab's ProviderIcon are intentionally untouched.

## Testing

- 13 new tests (determinism, mirror symmetry, density guard, ≥250/300 shape diversity, component render) — all green.
- `tsc`, frontend boundaries and structure checks clean; full suite unchanged vs baseline.


Julian0444:
Note on CI: the only failing GitHub check, `Repo shape checks`, currently fails on `main` itself as well (run 28569757778 on f7496a72) — 6 pre-existing `LARGE_FRONTEND_FILE` violations, none introduced by this diff (every file added here is well under the threshold, and the PR's merge ref reports the identical set as main). Vercel requires maintainer authorization for fork PRs. Happy to retrigger CI once main is green.

pablonyx:
@greptile

PR #883, opened 2026-07-02
fix(agent-auth): review fixes for runtime-bearer-enforcement

pablonyx:
Enforces the per-target bearer end to end: the install script wires `--require-bearer-auth` and an env-delivered `ANYHARNESS_BEARER_TOKEN` into the supervisor config when the bearer is present, and the desktop installer exports the bearer (redacted from all output) with the connect workflow threading it through. Review fixes cache the bearer on the local SSH profile (recovering pre-bearer profiles via `/runtime-access`), set `RuntimeTarget.authToken` on the ssh-transport branch, and verify authenticated `/v1` access in the tunnel check so 401s surface during connect instead of after.

Stack: 3/6 — 16-gateway-target-scoping → 17-target-bearer → 18-runtime-bearer-enforcement → 19-direct-runtime-core → 20-direct-auth-sync → 21-direct-runtime-ui

Design: `specs/tbd/ssh-personal-target-design.md`

### DIFF
--- a/anyharness/crates/proliferate-supervisor/src/config.rs
+++ b/anyharness/crates/proliferate-supervisor/src/config.rs
@@ -50,3 +50,58 @@ fn dirs_fallback_home() -> PathBuf {
         .map(PathBuf::from)
         .unwrap_or_else(|| PathBuf::from("."))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::SupervisorConfig;
+
+    #[test]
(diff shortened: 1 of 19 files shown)


pablonyx:
Closing this retired agent-auth stack. The managed LiteLLM, target-scoped routing, runtime auth, and current qualification architecture have since been implemented through newer lineages. These branches are no longer safe to merge; preserve any genuinely missing behavior as a fresh current-main slice.

PR #886, opened 2026-07-02
fix(agent-auth): review fixes for direct-runtime-ui

pablonyx:
Direct-runtime UI: the Compute pane reads as the direct family (a This Mac row, enrolled ssh targets with attach-state chips, managed cloud targets in their own section), each with a per-target agent-auth summary card deep-linking into the agents scope. The agents scope selector becomes the runtime-scope axis — Cloud + This Mac + one entry per enrolled runtime — with client-side inheritance render, Inherited-vs-Override on every value, and overrides clearing back to defaults via targetId-scoped DELETE; attach-state presentation derives from the direct-runtime connection machine across compute rows, selector dots, sidebar rows, and pickers. Review fixes add a scope=all route-selection listing so ApiKeysPane usage/revoke visibility covers per-runtime override rows.

Stack: 6/6 — 16-gateway-target-scoping → 17-target-bearer → 18-runtime-bearer-enforcement → 19-direct-runtime-core → 20-direct-auth-sync → 21-direct-runtime-ui

Design: `specs/tbd/ssh-personal-target-design.md`


pablonyx:
Closing this retired agent-auth stack. The managed LiteLLM, target-scoped routing, runtime auth, and current qualification architecture have since been implemented through newer lineages. These branches are no longer safe to merge; preserve any genuinely missing behavior as a fresh current-main slice.

PR #890, opened 2026-07-02
fix(cloud): kill E2B sandbox on destroy + rewrite provisioning specs

pablonyx:
## Summary

- **Bug fix**: `destroy_cloud_sandbox` revoked worker/gateway tokens and marked the DB row destroyed, but never told E2B to kill the actual microVM — the real sandbox kept running (and billing) until E2B's own idle timeout. No reconciler or webhook path covered this gap. Now calls `provider.destroy_sandbox()` when the sandbox actually reached E2B, logging (not raising) on failure so destroy still succeeds from the user's perspective.
- **Docs**: `sandbox-provisioning.md`, `workspace-lifecycle.md`, and `cloud-commands.md` described the pre-cutover architecture (`sandbox_profile`/`cloud_targets`/`managed_sandbox`, an async `cloud_commands` queue) that PR #809 replaced. Rewritten and adversarially fact-checked line-by-line against current code — notably corrects a prior claim that an async command queue exists (it doesn't; materialization is synchronous and Redis-lock-guarded, and not crash-safe) and documents the one-E2B-VM-per-owner isolation invariant.

## Test plan

- [ ] `destroy_cloud_sandbox` unit/integration test asserting `provider.destroy_sandbox` is called when `e2b_sandbox_id` is set (no existing coverage for this function)
- [x] Python syntax/import check on the modified service module
- [x] Every factual claim in the three rewritten specs verified against this worktree's code (draft → 3 independent adversarial fact-checks → corrections applied)

### DIFF
--- a/server/proliferate/server/cloud/cloud_sandboxes/service.py
+++ b/server/proliferate/server/cloud/cloud_sandboxes/service.py
@@ -8,6 +8,7 @@ ORM stack.
 
 from __future__ import annotations
 
+import logging
 from typing import Protocol
 from uuid import UUID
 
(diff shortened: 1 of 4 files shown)


pablonyx:

PR #892, opened 2026-07-02
fix(mobile): drop duplicated anyharness projection import

Julian0444:
Mobile typecheck currently fails on `main` itself (e.g. run 28622695478 on 1e9d7429): `use-mobile-chat-data.ts` imports the same four projection helpers twice — once from the old `lib/domain/chat/mobile-chat-anyharness-projection` and once from the relocated `./mobile-chat-anyharness-projections` — so tsc reports TS2300 duplicate identifiers and every open PR inherits the red check.

This keeps the relocated module, drops the stale import block, and deletes the now-orphaned byte-identical copy (its only importer was this hook).

Verified locally with `pnpm --filter @proliferate/mobile typecheck` → exit 0.


pablonyx:
Thanks for catching and fixing the duplicate projection import. This exact repair was independently merged in #896 (commit 21e7eaa9a4f917a532a86a0abd0ef7a443abce58): it removed the duplicate co-located import and orphaned byte-identical module while retaining the current lib/domain implementation. On current main, the duplicate identifier failure is gone and applying this older patch would remove the sole remaining implementation/import, so this PR is now superseded rather than safe to merge.

PR #893, opened 2026-07-02
fix(desktop): worker identity survives logout/user switch

pablonyx:
## The bug

The desktop worker's identity flows from the enrollment ticket minted at login. The enrollment effect (`use-desktop-worker-enrollment.ts`) was guarded by a module-level boolean that never reset, and sign-out neither reloaded the webview nor tore the worker down. In one app process:

- **Sign out, then sign in as a different user:** the effect never re-fired, so the worker and the integration-gateway dotfile kept the previous user's identity — the new user's local sessions reached the previous user's integrations through the gateway.
- **Sign out with no re-login:** the worker kept heartbeating and the gateway dotfile stayed valid for the departed user.

## The fix

- **User-keyed guard:** the enrollment effect now tracks the enrolled user id instead of a one-shot boolean. A different user signing in re-fires enrollment; the Tauri `ensure_desktop_dispatch_worker` command already deletes `worker.sqlite3` when handed an enrollment token, and the server-side ticket consumption revokes the predecessor worker + gateway token, so re-enrolling rotates the full identity.
- **Sign-out teardown** (`teardownDesktopWorker`): best-effort server-side revoke via a new user-authed `POST /v1/cloud/workers/desktop/revoke` endpoint (idempotent, keyed on caller user id + install id), then a new `stop_desktop_dispatch_worker` Tauri command kills the local worker process and deletes the `integration-gateway.json` dotfile from the AnyHarness runtime home.
- The hook moved from `AuthenticatedAppHost` (which unmounts on sign-out) to `AppRuntime` so it observes the authenticated → anonymous transition.
- `cloud_worker.rs` launcher discovery split into `cloud_worker/launcher.rs` to stay under the repo max-lines gate.
- Note: regenerating the cloud SDK openapi types also picked up the already-on-main `/v1/cloud/integrations/catalog` endpoint (the committed generated file was stale).

## Verification

- `python3 scripts/check_max_lines.py`, `scripts/check_server_boundaries.py`, `scripts/check_frontend_boundaries.py`, `scripts/check_proliferate_worker_structure.py` — all pass
- `python3 scripts/report_frontend_structure.py --strict --summary-only` — 0 violations
- `npx tsc --noEmit` (apps/desktop) — clean
- `npx vitest run src/hooks/cloud/lifecycle/use-desktop-worker-enrollment.test.tsx` — 5/5 pass (enroll once per user, no-op cold anonymous start, re-enroll on user switch, teardown + re-enroll after sign-out, teardown fires once)
- `cargo check -p proliferate` (src-tauri) — clean
- `pytest tests/integration/test_cloud_runtime_workers_api.py` with empty env vars — 11/11 pass (revoke requires auth, retires worker token + gateway token, scopes to caller not install id, double-revoke idempotent)


pablonyx:
Thanks — addressing the three Greptile findings:

**P1 — install-ID gating of the local teardown** (`ensure-desktop-worker.ts`): fixed in `4167ffa8a`. The teardown was split: server-side revoke (which needs the install ID) is now its own best-effort `revokeDesktopWorkerServerSide()`, and the local stop (`stopDesktopDispatchWorker` + gateway dotfile deletion) runs unconditionally, so a failing `getDesktopInstallId()` can no longer leave the departed user's worker/integrations alive.

**P1 — teardown/re-enrollment race on rapid sign-out → sign-in** (`use-desktop-worker-enrollment.ts`): fixed by the `workerLifecycleChain` serialization in `e6183df00`/`4167ffa8a`. `ensureDesktopWorker` and `teardownDesktopWorker` are now enqueued on a shared promise chain, so a still-in-flight teardown can't acquire the `state.process` mutex after a subsequent enroll and kill the freshly-started worker.

**P2 — guard set before enrollment completes, no retry on failure** (`use-desktop-worker-enrollment.ts`): fixed in `8110b3ed9`. `ensureDesktopWorker` now resolves a success boolean (never throwing), and the hook clears the guard back to `null` on failure — but only when it still holds the id we set, so a newer sign-out or user switch isn't clobbered. A silent failure no longer wedges the user; the next effect run retries. Not a retry loop — one reset per failed attempt. Covered by new failure-then-retry / success-no-retry tests.

(This branch has also been merged up to current `main` to clear the stale merge-preview CI — the Mobile typecheck, Ruff format, and agent-gateway test failures were all merge-preview staleness against an older main and pass on the merged tree.)

pablonyx:
@greptile

PR #895, opened 2026-07-03
chore(cloud): integration FKs + schema vestiges + gateway/health perf hygiene

pablonyx:
## What

One migration (`ab12cd34ef56`, chained off head `c3f7a1e9d2b4`, history stays linear) plus a small perf trio for the integrations/worker stack.

### 1. Foreign keys for the 9 integration/worker tables

The integration + runtime-worker tables shipped with zero FKs. This adds them, in the migration **and** the ORM models so DDL and models agree:

- **Intra-domain refs default to RESTRICT** (plain `ForeignKey`, NO ACTION): `definition_id` everywhere, `gateway_token.runtime_worker_id`. Deleting a definition that still has accounts/policies/flows/clients now raises `IntegrityError` instead of orphaning them.
- **CASCADE for the two pure-derivative children**: `cloud_integration_tool_schema_cache.account_id` and `cloud_integration_oauth_flow.account_id` — deleting an account now sweeps its cached tool schema and in-flight OAuth flows at the DB level.
- **Cross-domain refs imitate the sibling cloud models** (`sandboxes.py`, `agent_gateway.py`, `repositories.py`): `owner_user_id`/`created_by_user_id`/`updated_by_user_id` → `user.id` CASCADE, `organization_id` → `organization.id` CASCADE, `cloud_sandbox_id` → `cloud_sandbox.id` CASCADE.

### 2. Schema vestiges

- **Tool-cache `status` drops `'stale'`** (check constraint now `ready|error`). Nothing ever read or needed the stored state: staleness is derived from the `auth_version` snapshot (and now the TTL). The only writer of `'stale'` was `mark_tool_cache_stale`, which had zero callers — deleted. The ORM default `'stale'` is gone too; `upsert_tool_cache` always supplies `status` at insert, so it is simply required now.
- **`cloud_integration_account.credential_format`**: was `NOT NULL DEFAULT 'json-v1'` — a format that no writer ever produces (real values are `secret-fields-v1`/`oauth-bundle-v1`). Chosen fix: **nullable with no default**. Rationale: `upsert_account` inserts accounts *before* credentials exist, so a non-null format on a credential-less row is a lie; `set_account_credentials` already always supplies the real format alongside the ciphertext. Migration nulls the format on rows without a ciphertext.

### 3. `cloud_runtime_worker` columns for the versions PR

`worker_version` (64), `anyharness_version` (64), `hostname` (255), `machine_fingerprint` (128) — all nullable String columns, schema-only prep.

### 4. Perf trio

- **`get_grant_by_gateway_token_hash`**: removed the per-request `last_used_at`/`updated_at` write + flush on the gateway hot path (column kept; a code comment marks it as no longer stamped).
- **Tool-cache TTL**: `get_or_refresh_tool_cache` also treats a ready cache as stale once `fetched_at` is older than `CLOUD_INTEGRATION_TOOL_CACHE_TTL_SECONDS = 86_400` (new constant in `constants/cloud.py`), so provider-side tool changes surface within a day.
- **`list_integration_health`**: the per-account probes now run concurrently via `asyncio.gather`. Sessions are handled safely: each probe opens its **own** session via `session_ops.open_async_session()` (the pattern behind `run_with_fresh_session`) and commits it, because `AsyncSession` is not concurrency-safe and the OAuth probe both reads (oauth client) and writes (refreshed token bundle). Committing per-probe is also semantically right — a refreshed token should persist regardless of the surrounding request.

## Why

Fresh DB, pre-users: this is the cheapest moment to make the schema honest (real FKs, no phantom states/defaults) and to take the two known hot-path costs (gateway token stamp, serial health probes) off the table.

## Verification

- `alembic upgrade head` applied to a scratch DB (`hygiene_migration_check`), plus `downgrade -1` / re-upgrade round trip; FKs and the new check constraint verified via `psql \d`. `alembic heads` shows a single head (`ab12cd34ef56`).
- New tests: definition-with-accounts delete raises `IntegrityError`; account delete cascades tool cache + flows; gateway grant still resolves with `last_used_at` staying `NULL`; tool cache serves fresh, then refetches after `fetched_at` is backdated past the TTL.
- Fixed `test_integration_provider_access.py` fixtures to seed a real user row (they fabricated `owner_user_id`s that the new FK rightly rejects).
- `tests/integration/test_cloud_integrations_api.py`, `test_cloud_integration_gateway_api.py`, `test_cloud_integration_health_api.py`, `test_cloud_runtime_workers_api.py`, `test_integration_provider_access.py`, `test_schema_migrations.py::test_alembic_upgrade_creates_current_schema`: **51 passed**. `tests/unit`: **513 passed**.
- Pre-existing (also red on clean main, unrelated): `test_app_startup_fails_when_database_is_not_migrated[*]` fails on a duplicate `ix_llm_credit_grant_billing_subject_id` during `Base.metadata.create_all` in billing models.
- `ruff check` clean on `proliferate` + `tests`; all repo-shape gates pass (max-lines, server/frontend boundaries, worker structure, frontend structure strict).


pablonyx:
Addressed in `34a1800b8`.

**`health.py` — unhandled probe exceptions kill the endpoint**: confirmed still reachable after the per-probe-session/semaphore rework. `_account_health` only catches `CloudApiError`; a non-`CloudApiError` (an httpx/provider timeout, or a `SQLAlchemyError` from the probe session or its commit) propagated through `asyncio.gather` (default `return_exceptions=False`) and 500'd the whole health response.

Fix: wrapped the per-probe body in `_probe_account_health` in a `try/except Exception` that logs (with `exc_info`) and returns `(HealthVerdict.ERROR, "probe_failed")`, isolating a single account's failure so every other integration still reports. This fits the current structure (each probe is already its own isolated unit) better than `return_exceptions=True`, and keeps the `strict=True` zip intact since every outcome stays a proper tuple. `CancelledError` (a `BaseException`) is deliberately not swallowed.

Test: `test_health_isolates_non_cloud_probe_failure` monkeypatches `ensure_provider_access` to raise `TimeoutError`, then asserts the endpoint returns 200 with that account `error`/`probe_failed` while an unrelated seed still reports `needs_auth`. Health suite green (6 passed), ruff + shape scripts clean.

pablonyx:
@greptile

PR #897, opened 2026-07-03
feat(cloud): heartbeat version convergence + sandbox worker self-swap

pablonyx:
> **Stacked PR** — base is `infra/integrations-hygiene` (#895), not `main`. Merge #895 first, then retarget/merge this one. Do not squash into main directly.

## Locked design

- **The API is the version root.** Release CI stamps `WORKER_VERSION` (from the `proliferate-worker` crate manifest) into the server image alongside the existing desktop/runtime pins; `/meta` and the heartbeat ack report it. Nothing else decides what version a worker should run.
- **App-is-supervisor where an app exists.** On desktop, the app bundle owns the worker binary and its updates. The desktop-written worker config explicitly sets `self_update_enabled = false` (also the worker default) so the on-disk config documents the gate.
- **Heartbeat/self-swap only for appless hosts.** The cloud-sandbox sidecar is launched with plain `nohup` and has no launcher that could update it, so it is the one place that opts into `self_update_enabled = true` and converges its own binary onto the heartbeat's `desiredVersions`.
- **Desktop worker gated OFF.** The whole self-update path is behind the config flag; a desktop worker can never self-swap.

## What changed

**Server (version root + persistence)**
- `server/version.py` / `meta.py`: new `worker_version()` pin (env `WORKER_VERSION`, falls back to server version); `/meta` now reports `workerVersion`. Meta's downloads-base-URL helper deduped into `integrations/desktop_downloads.py`.
- `runtime_workers` heartbeat: request now self-reports `workerVersion` / `anyharnessVersion` (persisted forward-only on the worker row — an omitted field never clears a recorded value); ack now carries `desiredVersions` with the server's pins. Columns already existed in `c4d5e6f7a8b9_cloud_targets_workers` — no new migration, alembic history stays single-head.
- New public worker artifact endpoint: redirect to the pinned CDN artifact (`proliferate-worker` + `.sha256`), probing the versioned path and falling back to unpinned `stable` when the versioned manifest doesn't exist.
- CI/Dockerfile: `_deploy-server.yml`, `server-ci.yml`, and `server/Dockerfile` stamp `WORKER_VERSION` from the crate manifest like the other pins.

**Worker (`anyharness/crates/proliferate-worker`)**
- New `self_update.rs`: heartbeat-driven binary convergence — checksum-verified download via the artifact redirect, `--version` preflight of the staged binary, atomic rename over the current path, then `exec` into the new image (the `nohup`-launched sidecar has no supervisor, so an exit would never come back; the `flock`'d process lock rides the `O_CLOEXEC` exec boundary). An env marker carried across the re-exec prevents a hot-swap loop when the published artifact lags the pin (e.g. `stable` fallback).
- Heartbeats self-report `workerVersion` / `anyharnessVersion`; ack parsing tolerates missing/partial `desiredVersions`.
- Config: `self_update_enabled` (default **false**); sandbox bootstrap (`runtime/bootstrap.py`) opts in, desktop (`commands/cloud_worker.rs`) explicitly writes `false`.
- `scripts/check_proliferate_worker_structure.py` manifest updated in the same commit for the new `self_update.rs`.

## Verification

- Repo-shape gates, all green: `check_max_lines.py`, `check_server_boundaries.py`, `check_frontend_boundaries.py`, `check_proliferate_worker_structure.py`, `report_frontend_structure.py --strict --summary-only`.
- `server/`: `pytest tests/integration/test_cloud_runtime_workers_api.py tests/integration/test_cloud_integration_gateway_api.py` with `CLOUD_WORKER_BASE_URL`/`WORKER_VERSION`/`RUNTIME_VERSION`/`DESKTOP_DOWNLOADS_BASE_URL` emptied (CI-like env) — 23 passed. `tests/unit/test_meta_endpoint.py` — 10 passed.
- `anyharness/`: `cargo check` clean; `cargo test -p proliferate-worker` — 14 passed (swap planning, checksum formats/mismatch, heartbeat serde, config gating defaults).
- Alembic: single head (`ab12cd34ef56`), no new migration needed.
- Boot-heal sanity: the materialization launch path (`sandbox_io/worker_sidecar.py`) runs the template-baked binary (`test -x … && nohup …`) and writes the self-update-enabled config first, so a template whose binary lags the pin converges via the first heartbeat ack (~30s after boot) — boot-time version healing falls out of the heartbeat path with no launch-script change.

## Follow-ups (noted, not built)

- **Worker-less templates don't self-heal**: if the template has no worker binary at all, the launch command no-ops by design ("no worker this run"). Healing that means either baking the binary at template build or a checksum-verified curl bootstrap in the launch script — template/infra work, out of scope here.
- Pre-existing `Cargo.lock` drift on main (desktop crate `0.3.1` → `0.3.3` from the release commit) intentionally left out of this PR.


pablonyx:
Addressed the root cause in `ce2ae9913`.

**`service.py:220` — binary and `.sha256` resolved via two independent redirects**: confirmed. Each `worker_artifact_redirect_url` call probes `versioned_manifest_exists` independently, so a binary request and a checksum request can straddle a CDN publish and resolve to different directories (pinned vs `stable` fallback) → a spurious checksum mismatch, or a consistent-but-unpinned pair. The post-download version preflight added in `0cc976abe` catches the worst case (an unpinned pair) but doesn't remove the mismatch race.

Fix (smallest change consistent with the existing 302 design): the convergence attempt now resolves the version path **once**. The worker fetches the binary through the redirect, captures the redirect's resolved CDN URL (`response.url()`), and derives the checksum URL from it (`<binary>.sha256`, same directory → same version) via a direct `download_from_url`, instead of resolving the checksum through a second redirect. So the pair can never come from different directories.

Changes: `cloud_client/mod.rs` (`download_worker_artifact` now returns `DownloadedArtifact { bytes, resolved_url }`, plus a new `download_from_url`), `self_update.rs` (`checksum_url_for` + updated `converge` and safety-model docs), and a clarifying note on the server `worker_artifact_redirect_url` docstring (the per-asset endpoint stays for the install script, which fetches a fresh pair). `cargo test -p proliferate-worker` 17 passed (incl. `checksum_url_is_derived_from_the_binary_resolved_url`), clippy clean; server runtime-workers integration suite 16 passed; ruff + shape scripts clean.

(This branch also carries the `#895` health-probe isolation fix via the `infra/integrations-hygiene` merge `c78ad0e3f`.)

pablonyx:
@greptile

PR #902, opened 2026-07-03
feat(cloud): OAuth (DCR) support for org-custom MCP integrations

pablonyx:
## The gap

`create_admin_integration_definition` hardcoded `auth_kind="none"`, `oauth_client_mode=None` for every org-custom definition. An admin adding an OAuth-protected MCP URL got a definition nobody could ever authenticate to — the gateway would call it unauthenticated and fail — and the Add Custom MCP dialog said nothing about authentication, so the failure was silent until members hit it.

The DCR machinery was already issuer-generic: `resolve_oauth_client` auto-registers (RFC 7591) against any issuer discovered via RFC 9728/8414, and `start_oauth_flow`, the user integrations pane, and the gateway all branch on `definition.auth_kind` generically. Customs only needed the definition rows to carry the right `auth_kind`.

## The design

**Bounded auto-detect probe, reusing existing discovery.** On create with `authKind: "auto"` (the default), the server probes the MCP URL with `discover_protected_resource_metadata` — the same RFC 9728 protected-resource-metadata + `WWW-Authenticate` challenge walk the OAuth flow itself uses; no new probing code. The probe is wrapped in a 5s `asyncio.wait_for` so a slow or dead URL can never block creation.

- Challenge found → `auth_kind="oauth2"`, `oauth_client_mode="dcr"`, and the config gets the same `Authorization: Bearer {secret.accessToken}` header template the seeded oauth2 definitions use.
- No challenge → `auth_kind="none"` (unchanged behavior).
- Timeout → `auth_kind="none"`, but the response says so honestly (`authDetection: "unreachable"`) instead of pretending the server was verified open.

**Explicit override.** Admins can force `authKind: "none"` or `"oauth2"` and skip the probe entirely (`authDetection: "forced"`). Anything else is rejected.

**Honest outcome reporting.** The create response carries `authDetection: "detected" | "none" | "unreachable" | "forced"` so the UI can state what actually happened.

## UI changes

- The Add Custom MCP dialog gains an Authentication select (Auto-detect / None / OAuth, default Auto-detect) with copy explaining the probe, DCR, and that API keys are not yet supported for customs.
- The success toast states the outcome: OAuth required (and where members connect), server unreachable (saved without auth), or no authentication required.
- The org integrations list gains an auth column so existing definitions show how they authenticate.

## Verification

- `pytest tests/integration/test_cloud_integrations_api.py tests/integration/test_cloud_integrations_admin_api.py` with empty env vars — 27 passed. New coverage: detection sets `oauth2`/`dcr` + bearer header, unreachable creates without auth, explicit kinds skip the probe, invalid kinds rejected, and an end-to-end test that `POST /authentications` against a created org-custom oauth2 definition runs the generic DCR flow and returns an authorization URL.
- Desktop: `npx tsc --noEmit` clean; vitest on `AddCustomIntegrationDialog.test.tsx` + `org-integrations-presentation.test.ts` — 20 passed (auto-detect default, submit payload, invalid-form and rejected-submit handling, toast copy per detection outcome).
- Repo shape: `check_max_lines`, `check_server_boundaries`, `check_frontend_boundaries`, `check_proliferate_worker_structure`, `report_frontend_structure --strict` — all green. `ruff check` clean; `ruff format --check` clean on all files this PR touches (the two pre-existing reformat complaints on main are untouched here).
- `cloud/sdk/src/generated/openapi.ts` is regenerated; the diff beyond `authKind`/`authDetection` is main's server endpoints (`/meta`, desktop password auth) that the checked-in generated file had drifted behind.


pablonyx:
@greptile

PR #910, opened 2026-07-03
ui: match transcript message body size to composer input

pablonyx:
## What

Makes the chat **transcript message body** (assistant + user prose) read at the
composer input size — **14px / 22px** at the default preset instead of 12px / 20px —
so typed text and sent text match. Secondary tool/action chrome stays at 12px.

> Based on `tip/arc-test`, so the PR diff includes those upstream commits. Only the
> top commit (`ui: match transcript message body size to composer input`) is mine —
> 6 files.

## How

- **New token** (`apps/packages/design/src/css/dom.css`):
  `--text-message: var(--text-composer)` and
  `--text-message--line-height: var(--text-composer--line-height)`.
  Declared as an **alias of the composer token** (next to the existing derived
  `--text-chat-meta`) rather than a new scale entry, so it tracks the composer
  across every appearance preset automatically — no `UI_FONT_SCALES` / `tokens.ts`
  / `tw-merge` drift lock to keep in sync.

- **Ambient opt-in, not a global bump.** `MarkdownBody` is shared with secondary
  chrome (cloud tool-row detail bodies, plan cards, work history), so its prose
  renderers can't be bumped globally without enlarging that chrome. Instead the
  shared renderers read `text-[length:var(--prose-text-size,var(--text-chat))]`
  (+ matching line-height). Only the message wrappers set `--prose-text-size:
  var(--text-message)`; every other context falls back to `--text-chat` (12px) and
  is untouched.
  - `MarkdownBody.tsx` — `p / ul / ol / li / blockquote / table / th / td` + the
    wrapper now use the ambient prose var (via one shared `PROSE_TEXT` constant).
    Headings, inline `code`/`pre`, and code blocks are left on their own scales.
  - `AssistantMessage.tsx` — wrapper sets the ambient var; both the stable and live
    `MarkdownBody` inherit it.
  - `CloudChatUserMessage.tsx` — cloud user bubble uses `--text-message` directly.
  - `UserMessage.tsx` (desktop) + `PromptContentRenderer.tsx` — desktop user bubble
    sets the ambient var; the shared `FileLinkedText` reads it with a `--text-chat`
    fallback, so other `PromptContentRenderer` callers are unchanged.

- `PlanMarkdownBody` is intentionally **left at chat size** — plan cards are
  structured secondary content and no ancestor sets the opt-in var.

## CLS / no-scroll-bump

`--text-chat` is untouched, so `TRAILING_STATUS_MIN_HEIGHT`
(`calc(var(--text-chat--line-height)+1.5rem)`) — the reserved streaming/status slot —
is unaffected. The stable and live `MarkdownBody` are siblings under the same wrapper
and inherit identical `--prose-text-size` / `--prose-text-line-height`, so a run of
prose renders at byte-identical metrics whether it's in the live tail or after it
flushes into the stable prefix. The `+8` line-height relationship is preserved
(14+8=22, mirroring 12+8=20). **A streaming message's height is identical live vs
settled.**

## Verification

- `pnpm check:design` — clean for this diff (the 3 failures — `ModelConfigGrid`,
  `GridTile`, `RadioCardGroup` — are pre-existing on the base branch, none of them
  mine).
- `apps/desktop` `pnpm build` (`tsc && vite build`) — passes.
- Vitest: `appearance-css-drift` (4), `TranscriptItemBlock` (6), and all
  product-ui transcript tests (36) — pass.

### DIFF
--- a/apps/desktop/src/components/workspace/chat/content/PromptContentRenderer.tsx
+++ b/apps/desktop/src/components/workspace/chat/content/PromptContentRenderer.tsx
@@ -129,7 +129,7 @@ function FileLinkedText({ text }: { text: string }) {
   const tokens = tokenizeSerializedFileLinks(text);
 
   return (
-    <div className="whitespace-pre-wrap break-words text-chat leading-[var(--text-chat--line-height)]">
+    <div className="whitespace-pre-wrap break-words text-[length:var(--prose-text-size,var(--text-chat))] leading-[var(--prose-text-line-height,var(--text-chat--line-height))]">
       {tokens.map((token, index) => {
         if (token.type === "text") {
           return <Fragment key={`text-${index}`}>{token.text}</Fragment>;
--- a/apps/desktop/src/components/workspace/chat/transcript/UserMessage.tsx
+++ b/apps/desktop/src/components/workspace/chat/transcript/UserMessage.tsx
@@ -84,7 +84,7 @@ export function UserMessage({
           >
             <div
               ref={textRef}
-              className={`break-words select-text${
+              className={`[--prose-text-size:var(--text-message)] [--prose-text-line-height:var(--text-message--line-height)] break-words select-text${
                 !expanded ? " line-clamp-5" : ""
               }`}
             >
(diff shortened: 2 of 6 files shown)


pablonyx:
Superseded — landed on main via squash commit adbcc83cc (consolidated UI cleanup pass).

PR #913, opened 2026-07-03
feat(anyharness): workspace mobility install v2 — preserve-native resume + re-adopt

pablonyx:
Part of the **workspace-migration v2** stack (A→B→C→D). Spec: `specs/tbd/workspace-migration-v2.md`.

## What
The anyharness engine half of local↔cloud workspace migration: a workspace's agent sessions move between runtimes and **resume natively** on the other side (not just history — the live conversation continues).

- **`installMode: preserve_native_sessions`** on mobility install — keep `native_session_id` for Claude/Codex so a migrated session resumes instead of starting fresh (other harnesses move history-only, labeled as such).
- **Re-adopt** semantics for round-trips: a session coming home to its prior workspace updates in place instead of erroring on the duplicate id.
- **Destination placement fix**: mobility destination worktrees now live under the managed worktrees root so they're actually purgeable (previously wedged as `cleanup_failed`).
- **Route-auth isolation fixes** (surfaced by live cloud testing — route-authed harnesses in sandboxes read state from *isolated* homes, not the ambient one): mirror the resumed Claude transcript into the isolated `CLAUDE_CONFIG_DIR`; write the Codex api-key `auth.json` into the codex-local `CODEX_HOME` (the bare env var is ignored on the ACP resume path); canonicalize the Codex rollout path on collect so it survives the return leg. The Codex-auth piece is implemented in #907's two-phase `FileSpec` render (`PathFamily::CodexLocalAuth`).

## Evidence (whole stack)
- **6 green live `local↔E2B↔local` round-trips** driving the real server API + real E2B sandboxes; Claude **and** Codex sessions recall their context in-sandbox after the outbound move and locally after the return — the last **2 consecutive on this exact rebased-onto-current-main stack**.
- anyharness-lib: **781 unit tests pass** (incl. preserve-native, re-adopt, destination purge, codex auth.json content, rollout canonicalization, black-box round-trip scenario).
- Server: 34 `workspace_move` tests pass. Desktop: move suite green.


pablonyx:
Parking: the mobility v1 surface was deleted by #823; this stack's findings are captured in specs/tbd/workspace-migration-v2.md, which is the proposal of record. Reopen if/when migration v2 is greenlit.

PR #915, opened 2026-07-03
feat(desktop): move workspace to cloud

pablonyx:
Part of the **workspace-migration v2** stack (A→B→C→D). Stacked on #914. Spec: `specs/tbd/workspace-migration-v2.md`.

## What
The desktop **local → cloud** move experience.

- A pure **readiness resolver** (`safe | push_required | prepare_required | blocked`) over git status + source preflight + destination state, and a pure workflow sequencer for the saga (freeze → export → install → cutover → cleanup) with correct pre-/post-cutover failure semantics.
- `MoveWorkspaceDialog` + `MoveProgress`, reusing the existing publish (commit/push) machinery — which already routes to a cloud workspace via the runtime gateway, so git-prep works against either side with no new plumbing.
- Entry points: a workspace context-menu verb + a clickable location chip. Source-fate on cleanup respects workspace kind (managed worktree → destroy; plain local dir → mark remote_owned, never delete user files). Mid-move resume/abandon if a move is already in flight.

## Evidence
Move domain/workflow/component suite green (61+ tests); typecheck clean; part of the 6-green-live-round-trip stack.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #916, opened 2026-07-03
feat(desktop): move workspace back to local + cleanup

pablonyx:
Part of the **workspace-migration v2** stack (A→B→C→D). Stacked on #915. Spec: `specs/tbd/workspace-migration-v2.md`.

## What
The **cloud → local** mirror + cleanup that makes migration a symmetric round-trip.

- Direction-aware move: readiness + git-prep run against the cloud workspace (via the gateway), the server freezes+exports from the sandbox, and the desktop **re-adopts the original local worktree** if this branch already has one (the return-trip case) instead of minting a fresh destination.
- Dead-plumbing sweep: removes the orphaned `cloudMobilityWorkspaces` inputs and stale mobility types left after the Stack-1 cutover (the desktop local↔cloud sidebar merge had silently degraded to them).
- Scratch-pad key-stability regression test: proves the scratch notes survive a migration with **zero** code, because the key is already the logical (repo, branch) workspace id — identical local and cloud.

## Evidence
Round-trip bounce verified live (local→cloud→local, original worktree re-adopted, context still recalled); move + selection + logical-workspace suites green.


pablonyx:
Parking: the mobility v1 surface was deleted by #823; this stack's findings are captured in specs/tbd/workspace-migration-v2.md, which is the proposal of record. Reopen if/when migration v2 is greenlit.

PR #919, opened 2026-07-03
feat(activity): loops + rosters runtime (native writes, emulated scheduler, fork-compat ingest)

pablonyx:
## What

Runtime slice of the loops/rosters track: `anyharness` domains for loops + activity, native loops write path with an emulated Codex scheduler, activity roster reconcile-on-attach with a lazy `FeedService`, the loops HTTP endpoints + feed/activity watch WebSockets, and the regenerated SDK surface. Bundled with it are the fork-compat fixes discovered while wiring real fork output into this contract: process/subagent ingest now tolerates each fork's actual field names (`startedAtMs` epoch-ms, `subagent`/`subagents` keys, the claude-fork feed discriminator) instead of the originally-assumed shapes, the `acp_child_demux` producer now exits when its watcher drops, stale running subagents get reset on attach, and emulated loops get a 1-minute cadence floor plus a 20-loop-per-session cap.

## Why

This is where the `01-contract` wire types actually get produced and consumed against live fork output. Several of the fixes here exist because the forks' real emitted shapes diverged from what the contract assumed — dropping the whole roster element rather than degrading gracefully. The floor/cap fixes close two unbounded-resource gaps in the emulated-loop path (no native cron backing it).

## Gate verdict

**partial** — not covered in this pass: Codex collab-subagent child feed demux wire mechanics; OpenCode/Cursor "no loop affordance" checks (out of scope for Phase B/C); product/UI (pdev) layer verification (this run scoped to protocol layer); Codex child-thread re-list on reattach (blocked by no subagent ever spawning in the harness used for verification).

## Dependencies

Stacks on `01-contract`. This track (and the sibling fork PRs) stacks on the in-flight native-goals train — proliferate#909, claude-agent-acp#24, codex-acp#12 — the pin dance across those three covers all of it.

Part of a stack: 01-contract → **02-runtime** (this PR) → 03-desktop.


pablonyx:
## Remediation pushed — subagent usage wire fix (1 commit: `c073a3366`)

### Fixed
**Subagent usage never populated (Gate B — `usage` stayed null on every `task_progress`/`task_notification` subagent chunk)** — `ActivitySubagentWire.duration_seconds` was `Option<i64>`, but `claude-agent-acp` derives `durationSeconds` as `duration_ms/1000`, i.e. a fractional JSON number (e.g. `3.207`). serde's strict `i64` deserialize rejected the float and, because it fails the whole struct, dropped **every** subagent wire chunk (`server.log`: `malformed activity subagent wire payload ... floating point 3.207, expected i64`). This also silently starved the completion flip for background subagents, since the terminal `task_notification` carries `usage.duration_ms` and was being dropped the same way — so `id`/`running`/`feed` only ever worked at `task_started`, where usage is null.

Fix: an `opt_i64_lenient` deserializer rounds fractional seconds to whole seconds (applied to `tokensUsed`/`toolCalls`/`durationSeconds`), keeping the contract type `i64` — no SDK regen required — mirroring the existing `GoalWire` precedent.

### Regate verdict: confirmed fixed
Re-run against the real protocol layer: completed subagent now reports `{"durationSeconds":5,"tokensUsed":13608,"toolCalls":1}`, and status correctly transitions `running`→`completed`. No regressions observed.

### Build/test
`cargo build --workspace` clean. `cargo test --workspace` = anyharness-lib 863 passed / 0 failed, including new `activity_subagent_wire_accepts_fractional_duration_seconds` regression; all crates green. One pre-existing flaky failure in an **untouched** crate (`apps/desktop/src-tauri/src/app_config.rs::write_json_file_atomic_overwrites_previous_contents`) is a parallel-execution temp-path timestamp collision unrelated to this change — passes 6/6 single-threaded.

Note: `goals-b/03-desktop` (#920) was rebased on top of this commit to carry it forward and force-pushed separately with `--force-with-lease`; no independent changes landed there.

PR #920, opened 2026-07-03
feat(activity): desktop loops/rosters/feed UI, live-wired

pablonyx:
## What

Desktop slice of the loops/rosters track: chips/panels UI (fixtures-first), then live-wiring loops + rosters + feed streaming into the real runtime surface from `02-runtime`, plus a fix capping the feed-tail buffer at a 256 KB trailing window so a long-lived verbose process/subagent row can't grow its buffered content unbounded.

## Why

Surfaces the loops/rosters/feed contract in the desktop UI. The buffer cap closes a real unbounded-growth path found while wiring the live feed stream (`useFeedStream`) to expanded rows.

## Gate verdict

**partial** — not covered in this pass: agent terminals joining the existing terminals pane (live feed bytes currently render as raw un-parsed text in a `<pre>` rather than through the xterm-backed terminal viewport — flagged as a follow-up, not a localized fix); Codex collab-subagent child feed demux wire mechanics; OpenCode/Cursor "no loop affordance" checks (out of scope for Phase B/C); Codex child-thread re-list on reattach (blocked by no subagent ever spawning in the harness used for verification). This run was scoped to the protocol layer plus this desktop slice — no separate pdev product-layer verification pass.

## Dependencies

Stacks on `02-runtime`. This track (and the sibling fork PRs) stacks on the in-flight native-goals train — proliferate#909, claude-agent-acp#24, codex-acp#12 — the pin dance across those three covers all of it.

Part of a stack: 01-contract → 02-runtime → **03-desktop** (this PR).


pablonyx:
## Rebased onto #919's fix — no independent changes in this PR

`goals-b/03-desktop` was rebased onto the updated `goals-b/02-runtime` (which gained `c073a3366` — the subagent-usage fractional-duration fix, see #919) and force-pushed with `--force-with-lease`. This branch's own three commits (desktop chips/panels UI, live-wiring, feed-tail buffer cap) carry the same content as before; only their parent commit and hashes changed as a result of the rebase.

No new fix landed directly in this PR. See #919 for the runtime fix and proliferate-ai/claude-agent-acp#25 for the fork-side loop/roster mirror fixes this stack depends on end-to-end.

### Regate verdict
Not independently re-gated (desktop/UI layer was out of scope for this protocol-layer gate pass) — covered transitively via #919 and claude-agent-acp#25.

PR #922, opened 2026-07-03
fix(desktop): dev worker launch freshness + loud startup-crash surfacing

pablonyx:
## Root cause

In dev, the Tauri app resolved the desktop dispatch worker via `development_worker_candidates()` (a `target/debug` scan) **before** the `cargo run` fallback. So a stale `target/debug/proliferate-worker` built weeks ago (old enroll contract) got launched against a new server. It crashed on enroll with `missing field 'targetId'` — but only *after* the fixed 150ms startup death-watch window, so `ensure_desktop_dispatch_worker` reported `"started"` and the user silently lost the integration-gateway MCP (no dotfile written, and the failure never surfaced beyond telemetry).

## Changes

**1. Freshness-first launcher in debug builds** — `commands/cloud_worker/launcher.rs`
`PROLIFERATE_WORKER_BIN` still wins first (always). Then, **only under `cfg!(debug_assertions)`**, prefer `WorkerLauncher::CargoRun { cargo, workspace_root }` before the beside-exe / dev-candidate / PATH binary scans, falling through to those scans when `cargo` or the workspace root is unavailable. In dev, `cargo run` guarantees the worker matches the checked-out enroll contract (a no-op rebuild when the tree is fresh); a scanned `target/debug` binary can be arbitrarily stale. Release-build order is unchanged (the block is gated on `debug_assertions`, which was already where the old trailing `CargoRun` fallback lived, so this only reorders debug and removes a now-redundant block).

**2. Meaningful startup death-watch** — `commands/cloud_worker.rs`
Replaced the fixed `sleep(150ms)` + single `try_wait` with a ~100ms poll loop. Total window is **3000ms when `enrollment_token.is_some()`** (a fresh enrollment must survive the first cloud roundtrip — where contract mismatches die) and **500ms otherwise**. Breaks early on exit. On a detected exit it reads the last ~12 lines of the worker log at `paths.log` (`read_worker_log_tail`, tolerant of a missing/unreadable file) and appends that tail to both the `tracing::error!` and the returned `Err` string, so the real crash cause (e.g. the serde decode error) reaches the frontend instead of a bare "see log" pointer. A child still alive at the window end is success (`"started"`), same as before — with `CargoRun` the child is `cargo`, which may still be compiling, and the window is deliberately *not* extended to wait it out.

**3. Loud startup-crash surfacing (TS)** — `lib/workflows/cloud/ensure-desktop-worker.ts`
The caller chain is `useDesktopWorkerEnrollment` (hook) → `ensureDesktopWorker()` (workflow) → `ensureDesktopDispatchWorker()` (Tauri access). The workflow's `catch` previously routed the error to `captureTelemetryException` **only** — nothing user-visible; the hook just gets a `boolean`. It now also surfaces the failure via the app's existing toast store (`useToastStore.getState().show(msg, "error")` — the same escape-hatch used by `use-auth-orchestration-effects.ts`), combined with the richer error string from change 2. Access still lives in `lib/access`; the store lives in `@/stores` (workflows are permitted to import stores — precedent: `hot-session-ingest-manager.ts` — and the frontend-boundary gate confirms this). No component/access-boundary violation.

## Tests
- Added a `#[cfg(test)]` module covering `read_worker_log_tail` (missing file → empty, tail truncation, whole-file-when-shorter). The launcher preference is env-dependent with no existing seam, so it is left untested.

## Verification
- `cargo check -p proliferate` — clean.
- `cargo test -p proliferate` — **64 passed, 0 failed** (61 pre-existing + 3 new).
- `cargo clippy -p proliferate` — no new warnings in the touched code (remaining warnings are all pre-existing in untouched lines; clippy is not CI-gated with `-D warnings`).
- `npx tsc --noEmit` (apps/desktop) — the only errors are two pre-existing failures in `OrganizationIntegrationsPane.tsx` / `org-integrations-presentation.ts`, untouched by this PR (unrelated to cloud worker).
- `npx vitest run use-desktop-worker-enrollment.test.tsx` — **7 passed**.
- Shape gates from the repo root all pass, **no allowlist edits**: `check_max_lines.py` (cloud_worker.rs is 567 / 600), `check_server_boundaries.py`, `check_frontend_boundaries.py`, `check_proliferate_worker_structure.py`, `report_frontend_structure.py --strict` (TOTAL 0).

Note: `Cargo.lock` carries a one-line `proliferate` version sync (`0.3.3` → `0.3.4`); the lock on `main` was stale against the manifest (already `0.3.4` after `release: prepare release-2026-07-03`), and `cargo` re-syncs it on any invocation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4266,7 +4266,7 @@ dependencies = [
 
 [[package]]
 name = "proliferate"
-version = "0.3.3"
+version = "0.3.4"
 dependencies = [
  "base64 0.22.1",
  "chrono",
--- a/apps/desktop/src-tauri/src/commands/cloud_worker.rs
+++ b/apps/desktop/src-tauri/src/commands/cloud_worker.rs
@@ -11,7 +11,7 @@ use tauri::State;
 use tokio::{
     process::Child,
     sync::Mutex,
-    time::{sleep, Duration},
+    time::{sleep, Duration, Instant},
 };
 
 use crate::{app_config, sidecar::resolve_shell_path};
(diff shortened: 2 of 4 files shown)


pablonyx:
Superseded by the clean current-main port in #1136, now merged with the review hardening.

PR #946, opened 2026-07-04
fix(overnight): cleanup-on-revoke — delete stale native auth files on credential revocation

pablonyx:
## Problem

When an agent-auth credential is revoked, the server omits it from the rendered state.json (harness disappears from the document), but does NOT delete old native auth files from the sandbox. The old worker-side cleanup implementation that properly deleted files was removed in commit 36fcede33 (2026-07-01) during the Bifrost teardown, and the new v2 implementation (PR #907) does not implement any cleanup mechanism.

Stale auth files (.codex/auth.json, .claude/.credentials.json, etc.) remain on disk after revocation, meaning the credential material is still usable via native harness fallback — defeating the purpose of revocation.

This is spec gap #2 documented at specs/codebase/primitives/agent-auth.md §5.4.

## What changed

**F4 — confirmed**: Extended `materialize_agent_auth` in `server/proliferate/server/cloud/materialization/materialize/agent_auth.py` to:

1. Track `active_harnesses` in the manifest (list of harness_kinds present in the last materialized state)
2. On each reconcile pass, compare previous `active_harnesses` to current
3. For any harness that was previously active but is now absent (revoked), delete its native auth files from the sandbox using the per-agent allowlist

Deletion is strictly limited to the `NATIVE_AUTH_FILE_ALLOWLIST`:
- claude: `.claude/.credentials.json`, `.claude.json`
- codex: `.codex/auth.json`
- gemini: `.gemini/oauth_creds.json`, `.gemini/settings.json`
- opencode: `.config/opencode/auth.json`

Unknown harness kinds are logged and skipped (no deletion). The `allowed_root` parameter is passed to `sandbox_io.remove_owned_files` for path safety enforcement.

## How to verify

1. Unit tests pass: `PROLIFERATE_DEBUG=true JWT_SECRET=test CLOUD_SECRET_KEY=test uv run python -m pytest tests/unit/test_agent_auth_materialization.py -v`
2. `scripts/check_server_boundaries.py` passes
3. Key new test cases:
   - `test_revoke_credential_deletes_auth_files` — verifies `.codex/auth.json` removal when codex goes from active to absent
   - `test_revoke_all_credentials_deletes_all_auth_files` — verifies full cleanup when all harnesses are revoked
   - `test_no_cleanup_when_no_previous_harnesses` — backward compat: old manifests without `active_harnesses` do not trigger false cleanup
   - `test_manifest_includes_active_harnesses` — verifies the manifest now tracks active harnesses for future cleanup passes

## Follow-ups

- The first reconcile after this change deploys will have no `active_harnesses` in existing manifests, so no cleanup fires until the next state change writes the new manifest format. This is safe (conservative: no false deletions).
- `cargo test -p proliferate-worker` is listed as a gate but was not run because no Rust code was changed.

### DIFF
--- a/server/proliferate/server/cloud/materialization/materialize/agent_auth.py
+++ b/server/proliferate/server/cloud/materialization/materialize/agent_auth.py
@@ -84,6 +84,16 @@ from proliferate.server.cloud.materialization import operation, paths, sandbox_i
 
 logger = logging.getLogger("proliferate.cloud.materialization")
 
+# Per-agent-kind native auth file allowlist (spec §5.4, lines 272-276).
+# Only files in this allowlist may be deleted on credential revocation.
+# Paths are relative to SANDBOX_HOME (/home/user).
+NATIVE_AUTH_FILE_ALLOWLIST: dict[str, tuple[str, ...]] = {
+    "claude": (".claude/.credentials.json", ".claude.json"),
+    "codex": (".codex/auth.json",),
+    "gemini": (".gemini/oauth_creds.json", ".gemini/settings.json"),
+    "opencode": (".config/opencode/auth.json",),
+}
+
 
 @dataclass(frozen=True)
 class AgentAuthStateInputs:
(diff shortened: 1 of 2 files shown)


pablonyx:
Closing this stale auth experiment. Credential lifecycle and active-route scoping have since changed on main (including #1236); any remaining defect should be reproduced and fixed in a fresh focused slice.

PR #948, opened 2026-07-04
fix(cloud): repair GitHub-connect pre-clone TypeError (silent since #823 era)

pablonyx:
## Problem

When a user connects a GitHub repo via the GitHub App integration, the server triggers sandbox materialization to pre-clone the repository. However, the `_materialize_sandbox` function in `server/proliferate/server/cloud/materialization/materialize/sandbox.py` was calling `materialize_repo_environment_in_context` with the wrong signature:

**Old call (lines 55-60):**
```python
await repo_environment_materializer.materialize_repo_environment_in_context(
    db,
    ctx=ctx,
    repo_environment=repo_environment,
)
```

**Expected signature (repo_environment.py:67-74):**
```python
async def materialize_repo_environment_in_context(
    db: AsyncSession,
    *,
    ctx: operation.MaterializationContext,
    repo_environment_id: UUID,
    materialization_id: UUID,
    attempt_updated_at: datetime,
) -> None:
```

This caused a `TypeError` on every GitHub connect attempt for users with >=1 cloud repo environment. The exception was logged by `materialization/runner.py` but not surfaced to the user.

**Net effect:** GitHub connect appeared successful, but repos were never pre-cloned, so the first workspace create ate the full git clone inline (slow first-run experience).

## What changed

1. Fixed `sandbox.py` to call `begin_repo_environment_materialization` before the `materialize_repo_environment_in_context` call, matching the correct pattern from `repo_environment.py:materialize_repo_environment` (lines 36-54)
2. Added import for `cloud_repo_environment_materializations` store
3. Added regression test (`test_sandbox_materialization.py`) that verifies:
   - The correct function signature is used
   - Multiple repo environments are all materialized
   - All required parameters are passed correctly

### Unit test
```bash
cd server && .venv/bin/python -m pytest tests/unit/test_sandbox_materialization.py -xvs
```

### Integration test
1. Connect a GitHub repo via the GitHub App
2. Check server logs for no TypeError in materialization tasks
3. SSH into sandbox and verify repo is pre-cloned at the expected path

## Follow-ups

None. The `runner.py` exception logging already logs at ERROR level with full traceback (`logger.exception`), which is appropriate for ops visibility.

---

### DIFF
--- a/server/proliferate/server/cloud/materialization/materialize/sandbox.py
+++ b/server/proliferate/server/cloud/materialization/materialize/sandbox.py
@@ -7,6 +7,9 @@ from uuid import UUID
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from proliferate.db.store import cloud_sandboxes as cloud_sandboxes_store
+from proliferate.db.store import (
+    cloud_repo_environment_materializations as repo_mat_store,
+)
 from proliferate.db.store import repositories as repositories_store
 from proliferate.server.cloud.materialization import operation
 from proliferate.server.cloud.materialization.materialize import (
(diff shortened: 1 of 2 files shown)


pablonyx:
Fixture fixed, tests now 2/2 green.

Added the 6 missing RepoEnvironmentValue fields to _repo_environment() helper:
- repo_config_id=uuid.uuid4()
- git_provider='github'
- desktop_install_id=None
- local_path=None
- setup_script=''
- run_command=''

Both unit tests pass and server boundary check passes.

pablonyx:

PR #950, opened 2026-07-04
fix(cloud): recover killed sandboxes on connect + actually destroy provider sandbox on delete

pablonyx:
## Problem

Two bugs in the cloud sandbox lifecycle:

1. **Killed-sandbox permanent wedge**: `connect_ready_sandbox` only recreates when `e2b_sandbox_id IS NULL`. If E2B killed/expired the sandbox (30-day paused expiry, template retirement), `resume_sandbox` raises forever on the stale id; nothing clears it; gateway 502s indefinitely.

2. **Destroy leaks provider VM**: `destroy_cloud_sandbox` revokes workers and marks the row destroyed but never calls the provider's `destroy_sandbox` — the E2B VM keeps running until idle-pause, wasting compute.

## What changed

- Added `SandboxNotFoundError` to the provider-agnostic sandbox base layer
- E2B provider's `_connect` now catches `SandboxNotFoundException` and raises `SandboxNotFoundError`
- `connect_ready_sandbox` catches `SandboxNotFoundError`, clears the stale id, recreates the provider sandbox, and falls through to the fresh-launch path
- `destroy_cloud_sandbox` now calls `provider.destroy_sandbox()` best-effort before marking the DB row destroyed

**Safety rails**:
- Only the not-found exception class triggers id-clearing; transient/network errors propagate normally
- Provider destroy failure is logged but does not abort DB-level destroy

## How to verify

- `cd server && DEBUG=true pytest tests/unit/test_cloud_sandbox_connect_recovery.py tests/unit/test_cloud_sandbox_destroy_provider.py -v`
- All 5 new tests pass (recovery, transient propagation, destroy calls provider, destroy proceeds on failure, skip when no id)
- Existing gateway tests unaffected

## Follow-ups

- Consider adding a background sweep that reconciles `e2b_sandbox_id` references against provider state
- Surface the recovery event in product telemetry / user-facing toast


pablonyx:

PR #974, opened 2026-07-06
fix(catalog): show bare current-gen Claude models on native/api (regression from #955 dedup)

pablonyx:
## Summary

Fixes a regression from PR #955's Bedrock dedup where `claude-fable-5` and `claude-opus-4-8` were hidden from the model picker on NATIVE (OAuth) and API-KEY surfaces, making them inaccessible even though they are the only way to reach those current-gen models on those surfaces.

## Problem

PR #955 added `claude-fable-5`, `claude-opus-4-8`, and `global.anthropic.claude-fable-5` to `MODEL_VISIBILITY_OPT_OUTS` (defaultVisible:false) to avoid duplicate rows on the gateway model table.

But the `defaultVisible` flag is context-blind: it hides models on ALL surfaces, not just the gateway. The bare current-gen ids have `availability: [anthropic-api, anthropic-oauth]` (NOT bedrock), so they are:
- The ONLY way to access Fable 5 / Opus 4.8 on native/api surfaces
- Never gateway-reachable (the `us.anthropic.*` Bedrock variants are filtered out by availability there)

Result: a user logged in via `claude login` (OAuth) saw only 4 Claude models (default, sonnet, haiku, opus) and could NOT see Fable 5 or Opus 4.8.

## Fix

**scripts/agent-catalog/build-catalog.mjs**: Remove `claude-fable-5` and `claude-opus-4-8` from the opt-out list. They are oauth/api-only (per availability), never gateway-reachable, so they are NOT gateway duplicates and must stay visible for native/api surfaces. Keep `global.anthropic.claude-fable-5` (genuine same-context bedrock duplicate of `us.anthropic.claude-fable-5`).

**catalogs/agents/catalog.json**: Set `defaultVisible: true` on `claude-fable-5` and `claude-opus-4-8`. Leave `global.anthropic.claude-fable-5` at `false`.

**scripts/agent-catalog/add-gateway-contexts.mjs**: New post-processing script to add gateway auth context + gatewayPolicy to agents (curation data not from probes). This was missing after the catalog rebuild.

**Rust tests**: Update `visible_models_are_the_default_visible_subset_of_available` to reflect both current-gen models now visible. Update version assertions (2.1.181→2.1.201) from catalog pin refresh.

## Gateway table independence confirmation

The gateway model endpoint (`anyharness/crates/anyharness-lib/src/api/http/agent_gateway_catalog.rs:195`) uses `gateway_model_resolver.resolve_with_source` which probes the gateway directly via `config.yaml` (see `gateway_resolver.rs`). The catalog's `defaultVisible` flag joins metadata onto probe results via `resolve_catalog_match` but does NOT filter the list.

Therefore, un-hiding the bare ids CANNOT reintroduce a gateway-table duplicate — the gateway table is populated from live probe results, not from the catalog's visibility flags.

## Verification

**Before**: Native OAuth visible models = `[default, sonnet, haiku, opus, claude-fable-5]`  
**After**: Native OAuth visible models = `[default, sonnet, haiku, opus, claude-fable-5, claude-opus-4-8]`

- JS validator: ✅ `agent catalog OK: 2026-07-06.2 (5 agents)`
- Rust tests: ✅ `cargo test -p anyharness-lib --lib catalog` (75 passed)

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4266,7 +4266,7 @@ dependencies = [
 
 [[package]]
 name = "proliferate"
-version = "0.3.5"
+version = "0.3.6"
 dependencies = [
  "base64 0.22.1",
  "chrono",
--- a/anyharness/crates/anyharness-lib/src/domains/agents/catalog/schema.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/catalog/schema.rs
@@ -396,7 +396,7 @@ mod tests {
                 .native
                 .as_ref()
                 .map(|pin| pin.version.as_str()),
-            Some("2.1.181")
+            Some("2.1.201")
         );
         assert_eq!(
             claude
--- a/anyharness/crates/anyharness-lib/src/domains/agents/catalog/service_tests.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/catalog/service_tests.rs
@@ -39,7 +39,7 @@ fn pins_surface_catalog_harness_versions() {
     assert_eq!(claude.agent_process.version, "0.44.0");
     assert_eq!(
         claude.native.as_ref().map(|pin| pin.version.as_str()),
-        Some("2.1.181")
+        Some("2.1.201")
     );
 
     // Cursor has no native pin; unknown kinds have no pins at all.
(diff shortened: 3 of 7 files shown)


pablonyx:
Superseded by a minimal, surgical version (#974 regenerated the whole catalog, bumped harness versions, added an unrelated add-gateway-contexts.mjs, and removed codex's gateway default). Replaced with a clean 4-file diff.

PR #989, opened 2026-07-07
feat(worker): catalog fetch+push transport — pins converge without binary releases (convergence P2)

pablonyx:
## What

P2 of `specs/tbd/catalog-convergence-v1.md` — the missing middle of catalog convergence. With this, **bumping a harness pin in `catalogs/agents/catalog.json` converges cloud fleets on the next heartbeat**: worker sees a different `catalogVersion` → fetches the document from the cloud → pushes it to the runtime → `apply_fetched` swaps it → reconcile detects `VersionDrift` → the pinned CLI reinstalls. No binary release anywhere. (This is the transport `sync.rs:14-27` documented but never implemented.)

## Changes

**Runtime (anyharness-lib):**
- `GET /v1/catalogs/agents/version` → `{catalogVersion, source: bundled|fetched}` from `CatalogSyncService` (new contract type `AgentCatalogVersionResponse`), wired next to the existing `PUT`

**Worker (proliferate-worker):**
- `DesiredVersions.catalog_version: Option<String>` — tolerant of absence (older servers)
- New `catalog_sync` module: pure `plan()`/`needs_sync()` decision fns (self_update's pattern), then query runtime version → ETag-aware cloud fetch (`If-None-Match`, cached in `CatalogSyncState`) → `PUT` raw bytes to runtime
- 304-while-runtime-behind edge handled (refetch without ETag — covers a previously failed push)
- Runs in the heartbeat loop **before** self-update (a binary swap exec's and never returns)
- All failures log-and-skip; the heartbeat loop never crashes on catalog sync
- New config: `runtime_base_url` (default `http://127.0.0.1:8457`) + `runtime_bearer_token` (falls back to `ANYHARNESS_BEARER_TOKEN`)

## Tests

- 27/27 proliferate-worker (incl. new plan/needs_sync/deserialization tests + config defaults)
- 792/792 anyharness-lib (incl. new router test for the version endpoint)
- clippy `-D warnings` clean

## Stack

`#987` → `#988` (heartbeat field) → **this**. Live end-to-end verify (pin bump → CLI reinstall on a dev profile) tracked as the convergence spec's verification bar; remaining lanes P3 (desktop sync hook), P4 (re-probe), P5 (version UI) sequenced behind the ux/agents stack per the spec.

### DIFF
--- a/anyharness/crates/anyharness-contract/src/v1/catalogs.rs
+++ b/anyharness/crates/anyharness-contract/src/v1/catalogs.rs
@@ -13,3 +13,13 @@ pub struct ApplyAgentCatalogResponse {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub to_version: Option<String>,
 }
+
+/// The runtime's active agent catalog version and its provenance.
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct AgentCatalogVersionResponse {
+    /// The `catalogVersion` string from the active document.
+    pub catalog_version: String,
+    /// Where the active catalog came from: `"bundled"` or `"fetched"`.
+    pub source: String,
+}
--- a/anyharness/crates/anyharness-lib/src/api/http/catalogs.rs
+++ b/anyharness/crates/anyharness-lib/src/api/http/catalogs.rs
@@ -4,7 +4,7 @@
 //! `ETag` request header forwards the cloud response's ETag. Sync semantics
 //! live in `domains/agents/catalog/sync.rs`.
 
-use anyharness_contract::v1::ApplyAgentCatalogResponse;
+use anyharness_contract::v1::{AgentCatalogVersionResponse, ApplyAgentCatalogResponse};
 use axum::{
     body::Bytes,
     extract::State,
@@ -14,7 +14,7 @@ use axum::{
 
 use super::error::ApiError;
 use crate::app::AppState;
-use crate::domains::agents::catalog::sync::SyncOutcome;
+use crate::domains::agents::catalog::sync::{CatalogSource, SyncOutcome};
 
 #[utoipa::path(
     put,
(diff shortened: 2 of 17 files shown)


pablonyx:
**Live end-to-end verification passed** (dev profile `main`, 2026-07-06):

1. Booted server + runtime + worker all from this stack's binaries.
2. New runtime endpoint live: `GET /v1/catalogs/agents/version` → `{"catalogVersion":"2026-07-02.1","source":"bundled"}`.
3. Heartbeat schema live: `WorkerDesiredVersions.catalogVersion` present in the running server's OpenAPI (#988).
4. Bumped `catalog.json` to `2026-07-06.999` (working tree only) → ran the worker once:
   - `catalog sync: version mismatch, fetching from cloud advertised="2026-07-06.999" runtime_version="2026-07-02.1"`
   - `catalog sync: successfully pushed catalog to runtime`
   - runtime log: `agent catalog applied from_version=2026-07-02.1 to_version=2026-07-06.999` (ETag carried through), followed immediately by `agent reconcile job queued/started (agent_count=5)` — the poke → reconcile chain fired.
   - runtime endpoint after: `{"catalogVersion":"2026-07-06.999","source":"fetched"}`.
5. Steady-state: second `--once` run produced zero catalog-sync log lines (version match → silent no-op).
6. Reverted the bump; verify processes cleaned up.

This is the convergence spec's §6 verification bar for the cloud lane (pin-drift CLI reinstall exercised via the reconcile job; all agents reported already_installed/skipped as expected since pins were unchanged).

pablonyx:
Post-review hardening pushed (`24f64231f`) from an adversarial 4-lens critique workflow (concurrency / security / rollout / spec-tests, 3-refuter verification per finding):

- **P0 fixed — token leak via redirects**: authenticated cloud calls (enroll, heartbeat, catalog fetch) now use a `redirect::Policy::none()` client so the worker Bearer token can never follow a 302 to a foreign host; artifact downloads keep redirects on a separate unauthenticated client (the self-update CDN 302 path is unaffected).
- **P1 fixed — sandbox workers couldn't auth to the runtime**: `build_worker_config` now threads `runtime_bearer_token` (the runtime token already in scope at provision) into the worker config; without it, cloud catalog sync would have 401'd on every heartbeat.
- **P2 fixed — rollout log spam**: old-runtime 404 on the version endpoint now logs once-per-heartbeat at INFO ("runtime does not support catalog sync") instead of WARN.
- **Tests added**: version-endpoint auth (401 without bearer / 200 with), `resolve_runtime_bearer_token` precedence, 404-vs-other error classification. 31/31 worker, 794/794 anyharness-lib, clippy clean, 10/10 heartbeat integration.

Reviewed-and-skipped (deliberate): desktop bearer threading (belongs with a future `--require-bearer-auth` desktop change — works today since the desktop runtime is auth-free on localhost), `ingest_advertised_version` retention (documented contract surface), ETag machinery (failure-recovery path, not dead code).

PR #1011, opened 2026-07-08
tests(intent): tier-2 harness + auth/invitation specs

pablonyx:
Tier-2 "mocked intent" test harness — phase 1, track A. Implements the stack-boot fixture plus the first three blessed scenarios from `specs/developing/testing/scenarios.md` (T2-AUTH-1, T2-AUTH-2, T2-INV-1), per the tier-2 rules in `specs/developing/testing/README.md`: real server, real desktop **web** frontend, real Postgres — no fake sandbox provider, no mock LLM.

## What the harness does

`tests/intent/` is a pnpm workspace package:

- **`stack/boot.ts`** — boots the full stack programmatically on the profile-isolated port set for profile **`t2intent`** (allocated via `scripts/dev.mjs ensure --profile t2intent --lock`, the same machinery `make run PROFILE=` uses): alembic migrations → uvicorn server (`SINGLE_ORG_MODE=true`, `SETUP_TOKEN_FILE`, GitHub OAuth blanked per `specs/developing/local/feature-worktree-auth.md` layer B) → AnyHarness runtime (shared prebuilt binary; skippable via `TIER2_INTENT_SKIP_RUNTIME=1`) → desktop web via vite on `PROLIFERATE_WEB_PORT` with `VITE_REQUIRE_AUTH=true`. Waits on `/health` + the web root, and tears down the whole process group (plus the profile run.lock) afterward.
- **`stack/seed.ts`** — seeding via the product's own surfaces: first-run `/setup` claim, password login, org/invitation/membership API helpers, invited `/register` self-registration. Exactly two direct-DB touches, both documented in-line: backdating `expires_at` (no time-travel API exists; the direct-DB analog of the Stripe test-clock convention) and clearing `password_login_attempt` (the limiter buckets per client IP — 5 fails/15 min — and every Playwright context shares 127.0.0.1, so deliberate wrong-password negatives would 429 later logins).
- **`stack/global-setup.ts`** — Playwright globalSetup: one boot per run, teardown returned as globalTeardown.
- **`playwright.config.ts`** — 1 worker, serial (the suite shares one claimed single-org instance), 1 retry, traces on failure.
- **`.github/workflows/intent-tests.yml`** — provisional CI job, named `intent-tests (provisional)`, `continue-on-error: true` — explicitly **non-blocking** until the suite earns a flake-free record. Do not add to required checks yet.

## Run locally

```bash
pnpm install
cd server && uv venv .venv --python 3.12 && uv pip install -e ".[dev]"   # once
pnpm -C tests/intent exec playwright install chromium                     # once
pnpm -C tests/intent test                # full suite (boots + tears down the stack itself)
pnpm -C tests/intent test:auth           # auth spec only
pnpm -C tests/intent test:invitation     # invitation spec only
```

Needs local Docker Postgres+Redis running (`server/docker-compose.yml`, same as normal dev). The suite owns the `t2intent` dev profile and its `proliferate_dev_t2intent` DB; state persists between runs and the specs are idempotent against it (a fresh claim happens on a fresh DB; later runs assert the already-claimed path). `TIER2_INTENT_VERBOSE=1` streams server/vite logs.

## Scenario → spec mapping

| Scenario (scenarios.md) | Spec | Status |
| --- | --- | --- |
| T2-AUTH-1 setup claim + password lifecycle (incl. wrong-password + second-claim negatives) | `specs/auth.spec.ts` | ✅ green |
| T2-AUTH-2 session revocation | `specs/auth.spec.ts` | ✅ green |
| T2-INV-1 happy path + expired / revoked / wrong-email / duplicate-pending negatives | `specs/invitation.spec.ts` | ✅ green (12/12 total, verified from a fully fresh DB and a persisted one) |

Two deliberate interpretation notes, marked in-line in the specs:

- **T2-AUTH-2 revocation trigger**: driven via password change from a second context (`PUT /auth/password`), which bumps `token_generation` — the product's actual all-surface revocation mechanism (same one logout uses). The UI-side assertion (context A's calls 401 and the app returns to signed-out on next bootstrap) is per the scenario.
- **T2-INV-1 happy path** drives the **single-org accept surface**: the invite email's `/register?token={invitation_id}` page (that is what `invitation_delivery.py` links in single-org mode), then asserts membership/invitation/org-list state and a UI login. The hosted-style settings accept is covered by the gap test below.

## Product bugs / gaps found (surfacing these is half the point)

1. **`.test`-TLD emails: 500 on `/users/me` after successful account creation.** The first-run claim and password login happily accept `owner@x.test`, but `UserRead`'s `EmailStr` serialization rejects RFC-reserved special-use TLDs, so every `/users/me` call for that user 500s. Validation mismatch between account creation (`server/proliferate/server/setup/accounts.py`, regex-based) and profile serialization (`server/proliferate/auth/profile_api.py`, pydantic `EmailStr`). Repro: claim with a `.test` email → login → `GET /users/me`. The suite steers around it with `.example.com` addresses (comment at `tests/intent/stack/seed.ts`).
2. **Non-admin invitee cannot reach the pending-invitation accept UI.** `CurrentUserInvitationsSection` lives on the `organization-members` settings pane, which is `adminOnly` in settings navigation and `SettingsScreen` redirects non-admins to General. So the surface scenarios.md names for accepting an invitation is unreachable for exactly the user who needs it (a plain member/invitee) — while the backing endpoints (`GET /organizations/invitations/current`, `POST .../accept`) work fine for them. Pinned as its own test (`documents GAP: ...` in `invitation.spec.ts`) that asserts the *current* gated behavior, so a fix flips it loudly.
3. **Spec-contract divergence, duplicate-pending negative.** scenarios.md says duplicate pending invite → "rejected by partial unique index → enumerated error", but the as-built path (`create_or_rotate_organization_invitation`) deliberately rotates: expires the old pending row, mints a fresh one, `ON CONFLICT DO UPDATE` as the concurrency backstop. The test pins the real invariant (at most one pending row per (org,email); old row expired). scenarios.md should be updated to match the rotate semantics.
4. **Harness-relevant quirk, not a bug**: invitation create returns **201**, and org API field names are camelCase aliases — noted here because scenarios.md's shorthand implies snake_case/200.

## Notes

- No product code was changed. Only `pnpm-workspace.yaml` (+lockfile) gained the `tests/intent` package entry.
- Verified end-to-end twice locally: once against a fully fresh DB (fresh `/setup` claim path) and once against the persisted profile DB (already-claimed path). Both 12/12.

### DIFF
--- a/.github/workflows/intent-tests.yml
+++ b/.github/workflows/intent-tests.yml
@@ -1,0 +1,92 @@
+name: Intent Tests
+
+# Tier-2 "mocked intent" suite (specs/developing/testing/README.md §Tier 2):
+# real server + real desktop web frontend + real Postgres, Playwright driving
+# a browser. PROVISIONAL: this job is explicitly NON-blocking while the
+# harness earns trust — do NOT add it to branch-protection required checks
+# yet. Failures are signal to read, not a merge gate. The tier-2 suite joins
+# the merge gate (per the testing standard's CI mapping) once it has a
+# demonstrated flake-free record.
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+
+jobs:
+  intent-tests:
+    name: intent-tests (provisional)
+    runs-on: ubuntu-latest
+    timeout-minutes: 40
+    # Provisional posture, enforced twice: the job name says so, and the job
+    # never fails the workflow while it is provisional.
+    continue-on-error: true
+    services:
+      postgres:
+        image: postgres:16-alpine
+        env:
+          POSTGRES_USER: proliferate
+          POSTGRES_PASSWORD: localdev
+          POSTGRES_DB: proliferate
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U proliferate -d proliferate"
+          --health-interval 5s
+          --health-timeout 3s
+          --health-retries 10
+      redis:
+        image: redis:7-alpine
+        ports:
+          - 6379:6379
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 5s
+          --health-timeout 3s
+          --health-retries 10
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version: "22"
+      - uses: pnpm/action-setup@v6
+        with:
+          version: 10
+      - uses: astral-sh/setup-uv@v7
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+
+      - name: Install workspace dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Install server (venv, matches local layout)
+        run: cd server && uv venv .venv --python 3.12 && uv pip install -e ".[dev]"
+
+      - name: Install Playwright chromium
+        run: pnpm -C tests/intent exec playwright install --with-deps chromium
+
+      - name: Run tier-2 intent suite
+        env:
+          # The boot fixture allocates the t2intent profile itself; on the
+          # runner the Docker services above stand in for the local stack.
+          USE_EXISTING_POSTGRES: "1"
+          USE_EXISTING_REDIS: "1"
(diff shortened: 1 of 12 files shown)


pablonyx:
Landed via #1016's squash — tests/intent-wave2 contained this branch's commits and merged to main; the harness skeleton is on main now.

PR #1007, opened 2026-07-08
feat(workflows): PR D — workflow composition by resolution-time inlining (L20)

pablonyx:
Fourth PR of the A–E arc (stacked on #1006). **Fully reworked 2026-07-08 per ruling L20** — the earlier fire-and-forget \`workflow.run\` / \`spawn_child_run\` / parent-linkage design on this branch was discarded wholesale (force-pushed).

## The design (L20, §3.5)

Composition is **nesting, not spawning**: using workflow B inside workflow A inlines B's steps into A's single resolved plan at StartRun, server-side, before delivery. One run, one plan, one cursor, one sandbox. No child run, no parent linkage, no cross-sandbox fan-out. **The runtime is completely unchanged by this PR — no \`plan.rs\` change, no migration. That is the design, not an omission**: the resolver eliminates the include before the runtime ever sees it, and there's a test asserting the L20 property (a resolved plan contains no include steps).

- **\`workflow.include\` step** (definition-only): \`{kind, workflow_id, args: {child-arg: template}, name?}\`. New \`domain/composition.py\` owns the resolver.
- **Arg binding (§3.5 obligation a)**: the include's args mapping becomes the child's interpolation context — textual substitution of the child's \`{{args.*}}\` (both sides are author-written definition text, so no brace-escaping, unlike user StartRun args). Mapping values may carry the parent's \`{{args.*}}\` (inline FIRST, then the normal eager pass) and parent step refs (stay late-bound).
- **Ref namespacing (§3.5 obligation b)**: child internal refs are renumbered to child-flattened positions, named refs rewritten to index form, then offset by the splice point — collision-free; child step names prefixed \`<includeName>_\`.
- **Save-time validation**: target exists/same-owner/not-archived, no self-include, required-arg coverage, and the **cycle check** (A→B→A fails at save with \`include_cycle\` naming the path). Resolution-time: depth cap 5 (\`include_depth_exceeded\`), arg mismatch when the child's current version drifted (\`include_args_mismatch\`) — both fail the run before delivery.
- Parent refs *at* an include step are rejected (an include has no output; referencing into the child from the parent is out of scope v1).
- **UI**: include editor (workflow picker + per-child-arg mapping, caption "Steps run inline in this workflow's single run"); run view needs nothing (inlined steps are ordinary steps).

## Surface-syntax decisions pending veto (the doc left these open; also recorded in the deep-dive doc)
1. Step kind named \`workflow.include\` (not \`workflow.run\` — no runtime verb implied).
[...]

PR #1022, opened 2026-07-08
feat(workflows): PR E — per-run gateway tokens, completion ping, two-layer function scoping

pablonyx:
Final code PR of the A–E arc (stacked on #1007; design of record \`codex/workflows-architecture.md\` §3.7, §5.3–5.4, §6, rulings L16/L21–L26).

## What

**Per-run gateway token, minted for EVERY run** (L16 — empty scope legal), doubling as the run-report credential:

- **\`cloud_workflow_run_gateway_token\`** (migration \`d2f4a6c8e0b1\`): hashed token, \`scope_json\` = resolved \`functions[]\`, status \`active|expired|revoked\`, expires at terminal status (report AND refresh paths, idempotent). Plaintext travels only in \`resolved_plan_json.gateway = {url, authorization, ping_url, functions}\`.
- **Completion ping (§3.7)**: \`POST /v1/cloud/workflows/runs/{run_id}/ping\` — auth IS the per-run token (no user session); token↔path run mismatch → 403; wakes the existing \`_sync_run_from_view\` refresh for cloud runs (local: 202 no-op, the relay owns local observation). Stateless nudge: duplicate/stale/late pings safe by construction. Runtime fires it fire-and-forget after **every** \`drive_run\` transition via an injected \`RunPingSink\` (failure structurally inert — the sink returns \`()\`); scheduler phase 3 demotes to backstop.
- **L22 fail-fast**: StartRun fails the run outright (before the run row exists) if a declared provider has no ready account — never silently narrows.
- **L25 two-layer scope**: nullable \`scope_json\` provider allowlist on the worker token (NULL = unscoped, distinct from empty, no backfill) + nullable scope key on \`CloudIntegrationPolicy\`. The run scope is intersected with the worker allowlist **at delivery** (the worker isn't known at StartRun for cloud runs — the token row and plan block are re-frozen to the intersection before the plan ships), and the live worker allowlist is **re-checked on every gateway request**, so a post-mint narrowing bites on the next call.
- **Gateway enforcement**: run-token-first grant resolution (worker path untouched); \`tools/list\` filtered to scope; \`tools/call\` outside scope → enumerated error. The scope check is a pure helper (\`integration_gateway/domain/scope.py\`) callable outside HTTP — §6.6's future \`function_call\` server-side performer authorizes through the same path.
- **L26**: \`purpose\` enum on \`CloudSandbox\` (\`interactive\`/\`workflow-run\`), stamped at creation, never inferred.
- **§5.3 local-lane explicit error**: a local run whose plan declares functions it cannot honor fails at the first agent step with \`functions_unsupported_local\` — never a silent zero-tools no-op. Editor + launch modal carry the loud cloud-only captions; L21 launch set is issues + Slack.
- **Gateway launch extra**: the plan's gateway block is injected through the **existing** \`SessionExtension\`/\`SessionMcpServer::Http\` seam (L8: no new injection concept), registered per **session id** so a co-located non-workflow session can't inherit the run's credential; takes precedence over the worker dotfile; re-registered on crash-resume.
- **Contract fixture** \`fixtures/contracts/run-ping/\` (golden ping request + gateway block), asserted from both Python and Rust suites.

## Verification
366 server workflow+gateway tests green on this tip (24 new run-gateway tests: mint-per-run, token-scoping 403, ping idempotency + no-status-regression, terminal expiry both paths, L25 intersection + NULL-vs-empty + post-mint narrowing, L22 fail-fast, purpose stamping, worker-path regression); 886 cargo tests (ping-after-every-transition incl. terminal, failure-inert, no-gateway-no-pings, launch-extra wiring, fixture parse); desktop tsc clean; single alembic head, chain walks A→B→E.

Flagged: \`make cloud-client-generate\` was not run for the ping route (runtime-facing, not consumed by the TS SDK) — if repo-shape requires the regen it's a one-command follow-up.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/app/mod.rs
+++ b/anyharness/crates/anyharness-lib/src/app/mod.rs
@@ -75,7 +75,10 @@ use crate::domains::workspaces::store::WorkspaceStore;
 use crate::domains::workspaces::worktree_runtime::WorkspaceWorktreeRuntime;
 use crate::live::sessions::LiveSessionManager;
 use crate::live::terminals::{AgentLoginTerminalService, TerminalService};
-use crate::live::workflows::{WorkflowExecDeps, WorkflowOwnedSessions, WorkflowRunManager};
+use crate::live::workflows::{
+    HttpRunPingSink, WorkflowExecDeps, WorkflowGatewaySessions, WorkflowOwnedSessions,
+    WorkflowRunGatewaySessionLaunchExtension, WorkflowRunManager,
+};
 use crate::persistence::Db;
 
 #[derive(Debug, thiserror::Error)]
@@ -252,6 +255,10 @@ impl AppState {
         // Shared registry of workflow-owned sessions: written by the workflow
         // executor, read by the inbound permission advisor (always-bypass net).
         let workflow_owned_sessions = Arc::new(WorkflowOwnedSessions::new());
+        // Per-run gateway MCP servers, keyed by session id: written by the
+        // workflow executor before launch, read by the workflow gateway launch
+        // extension (§6.4/OPEN-3(a)).
+        let workflow_gateway_sessions = Arc::new(WorkflowGatewaySessions::new());
         let acp_manager = sessions::wire_live_sessions(&sessions::LiveSessionsWiringDeps {
             db: db.clone(),
             runtime_home: runtime_home.clone(),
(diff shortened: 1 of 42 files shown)


pablonyx:
⏸️ **Hold — in-stack rework in progress.** The core-engine amendments pass (codex/workflows-data-contract.md + codex/workflows-engine-amendments.md, fully ruled 2026-07-08) supersedes parts of this branch's design (definition format v2: inputs/integrations/agents spine; structured step keys; short-ref grammar; step-kind union cut). The branch will be force-pushed with the amended design; PR body will be rewritten to match. Do not review the current diff.

pablonyx:
✅ **Landed in `workflows/v1` via the Gate-A fast-forward merge train** (v1 now at `caffd3618`). This stacked PR's commit is contained in v1 (one-commit-per-branch preserved in v1's linear history), so GitHub can't show it as "merged" (empty diff vs v1) — closing with this reference; the branch and full review trail remain for the record.

Gate A passed: verify battery green (cargo 893 / server pytest 1013 / golden fixture both languages / per-branch standalone-green / ruling greps clean), adversarial refute-review over every branch, and the 3 major findings fixed + re-confirmed. See codex/program-decisions.md (D-007..D-009).

Issue #1032, opened 2026-07-09
billing: customer.subscription.deleted applies payment_failed hold unconditionally (even after clean cancellation)


pablonyx:
## Finding (from tier-2 billing suite T2-BILL-8)

`customer.subscription.deleted` always applies a `payment_failed` billing hold, regardless of why the subscription ended. After a clean voluntary cancellation this is wrong: the user gets a payment-failed hold they never earned.

Code: `server/proliferate/server/billing/stripe_webhooks.py` — the `customer.subscription.deleted` branch of `_dispatch_stripe_event` unconditionally calls `_apply_payment_hold_for_subscription(...)`, which calls `apply_payment_failed_hold`. There is no branch on cancellation reason. Survey 2026-07-08 confirmed the reason-sensitive refinement was never shipped.

## Status: UNRULED
Pablo to rule fix-now vs post-release. Per the testing standard, product code is not changed without a ruling.

## Pinned in tests
`tests/intent/specs/billing/webhooks.spec.ts` pins the CURRENT (buggy) behavior with a `known-bug` annotation and references this issue. When the fix lands, that assertion inverts (assert NO payment_failed hold after a clean cancellation).

PR #1008, opened 2026-07-08
specs: testing standard + flow registry + scenarios

pablonyx:
## Summary

Adds the automated-testing spec tree under `specs/developing/testing/` and wires it into the existing spec read maps. Spec-only change, no code.

- **`testing/README.md`** — the four-tier testing model:
  - **Tier 1 — unit/contract**: pure logic + logic against real DB state; contract fixtures under `fixtures/contracts/` for cross-language shapes. Gates **merge**.
  - **Tier 2 — mocked intent**: real server + desktop web frontend + real Postgres, everything third-party (sandbox, LLM, IdP, email, Slack) faked. Gates **merge**.
  - **Tier 3 — live end-to-end**: real E2B template, real AnyHarness binary, real agents on cheap models, nothing faked. Gates **release** (release train + nightly), never merge.
  - **Tier 4 — upgrade path**: N−1 install upgraded in place to N, real artifacts served by a stub update feed. Gates **release**.
  - Hard rule: the merge gate is tiers 1–2 only — no real LLM/sandbox/third-party ever runs in the merge gate, so a tier 3/4 failure blocks staging→prod promotion (and files into the issues service) but never blocks a merge.
  - Also documents per-language placement, the decision procedure for where a change's tests belong, and the postmortem rule (a prod incident that automated tests should have caught gets a new test in the same PR as the fix).
- **`testing/flows.md`** — the flow registry: every end-to-end flow that must never break, each tagged with its tier and a pointer to the enforcing test. Owned by Pablo; updated in the same PR as any flow change.
- **`testing/scenarios.md`** — the blessed contract for the test-infra build: the concrete scenario definitions (fixtures, identities, lanes) that `flows.md` entries and the tier 2/3 runners implement against.
- Cross-links added in `specs/README.md` (Current Read Map), `specs/spec-catalog.md` (new Testing section + QA-is-manual-only clarification), and `specs/developing/README.md` (Read Order + Developer Process Map).

## Test plan

- [x] `python3 scripts/check_max_lines.py` passes (markdown isn't in its checked roots/extensions, confirmed no-op)
- [x] Confirmed the other repo-shape scripts (migration heads, anyharness paths, worker structure, frontend/server boundaries) don't reference `specs/` at all
- [x] Diffed cross-link edits against `specs/README.md`, `specs/spec-catalog.md`, `specs/developing/README.md` to confirm semantic-only edits (no unrelated rewrites)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1033, opened 2026-07-09
tests(intent): tier-2 billing suite (T2-BILL-1..9)

pablonyx:
Was stacked on `tests/intent-wave2` (#1016); now rebased onto `main` (the train landed) and retargeted — the interim merge-of-main is gone from this branch's history, so the diff is this suite's own commits only.

Adds the tier-2 billing suite: a Stripe-test-mode harness plus the nine T2-BILL scenarios, run against a real booted stack (pro billing on, `CLOUD_BILLING_MODE=enforce`) on the dedicated `t2billing` profile. Stripe is real test mode (test clocks, real subscriptions/invoices). No mocked Stripe, no mock LLM, no fake sandbox: compute is seeded `usage_segment` rows, LLM spend is seeded `agent_llm_usage_event` rows, and the accounting/reconciler passes are the product's own functions run out-of-process against the profile DB.

## Per-scenario status

| Scenario | Status | Notes |
| --- | --- | --- |
| T2-BILL-1 checkout → grants → consumption → cut-off → reactivate + drain order | green | real test-clock sub + invoice.paid grant; real accounting-pass drain; snapshot cut-off (`credits_exhausted`) + reactivation |
| T2-BILL-2 plan limits + policy gates | green (reduced) | policy min-plan 403 asserted; repo-limit stays tier-1; "plan gates model list" has no code today ([PABLO TO RULE]) |
| T2-BILL-3 seats invite/remove/re-invite | green | adjustment rows + Stripe quantity + one proration grant; no double-grant on same-period re-invite; failed_terminal retry |
| T2-BILL-4 team checkout | green (reduced) | intent-create + activation idempotency; `failed_billing_state`/24h-expiry left to tier-1 (heavier staging than per-merge budget) |
| T2-BILL-5 compute overage | green | billable cents to cap, write-off past it, `cap_exhausted` block; disabled → immediate cutoff, zero exports |
| T2-BILL-6 LLM credits | green (reduced) | balance/cap surfaces + fail-closed top-up asserted; the LiteLLM key-*disable* side effect is tier-3 (it calls the live gateway) |
| T2-BILL-7 webhook robustness | green | duplicate/replay idempotent, concurrent ≤ one grant, out-of-order converges |
| T2-BILL-8 subscription edges | green + 1 pinned | payment-failed hold/clear, cancel + rollover-grace cutoff, no-GitHub trial; **finding 7 pinned expected-fail** |
| T2-BILL-9 usage surfaces | green | summary/timeseries/llm-balance/by-user match seeded totals; UI smoke on the billing pane |

"green (reduced)" = the full-stack-only behavior is asserted here; a named sub-part is documented in the spec header as tier-1 or tier-3, not silently dropped.

## Findings filed

- **Finding 7 (#1032)** — `customer.subscription.deleted` applies a `payment_failed` hold unconditionally, even after a clean voluntary cancellation (`stripe_webhooks.py` deleted branch always calls `_apply_payment_hold_for_subscription`). UNRULED. Pinned in `webhooks.spec.ts` with a `known-bug` annotation referencing #1032; product code untouched. When the fix lands the assertion inverts.

## Pinned expected-fail

- T2-BILL-8 subscription.deleted hold (finding 7 / #1032).
- The T2-BILL-9 by-user compute assertion is guarded by `T2BILLING_ORG_COMPUTE_ATTRIBUTION` (default off). #1028 has since merged, but it fixed org compute *enforcement* attribution only — the org usage-by-user display still queries by the org billing subject (an explicitly-scoped follow-up in #1028), so the default-off guard still matches deployed behavior. Flip the env when the display follow-up lands.

## Where the contract and the code diverge (for the record)

- The literal "resume/connect blocked (402)" of T2-BILL-1/5/8 needs a real E2B sandbox (tier 3). At tier 2 I assert the billing snapshot the resume gate reads (`/billing/overview` `startBlocked`/`startBlockReason`), which is the decision the UI renders. `authorize_sandbox_start` is dead code as the survey noted; the resume gate is the seam.
- Grant drain *order* is pure logic (`ordered_accounting_grants`) owned by tier-1; the tier-2 test still exercises it end-to-end through a real accounting pass and asserts the earliest-expiring grant drains first.
- Billing modes off/observe are a boot-env posture; this suite boots `enforce`. Per-mode smoke would need three boots — left to tier-1 rather than three stack boots per merge.

## CI

New `intent-billing` job in `intent-tests.yml`, provisional and non-blocking like the auth suite. It installs the Stripe CLI and needs a `STRIPE_TEST_SECRET_KEY` repo secret; with no key the suite skips itself cleanly (globalSetup marks every spec skipped, not failed). Specs live under `specs/billing/` and are split by scenario group so every file stays under the repo-shape 600-line cap.

## Remaining coordination

- Done: retargeted to `main` after #1016 landed; the interim `main` merge is dropped from history.
- The `specs/developing/testing/{README,flows,scenarios}.md` here are copied from the blessed `specs/testing-standard` branch (#1008, unmerged — main has no testing specs yet). They dedupe when that lineage lands; only the flows.md billing-row pointers are mine.
- Add the `STRIPE_TEST_SECRET_KEY` repo secret to actually run the billing job in CI (otherwise it skips green).

## Verified locally

- Full suite typechecks.
- The real Stripe test-clock path (test clock → customer with subject metadata → card → active subscription → paid invoice) verified end-to-end against Stripe test mode.
- Webhook self-signing matches the server's verifier byte-for-byte (`integrations/stripe/webhooks.py`: HMAC-SHA256 over `t.payload`, secret as-is).
- A full green run of all nine needs the stack booted (server venv + frontend build + Stripe key); that is the `intent-billing` CI job's / a local `make`-profile run's job, not something I could complete in one pass here.

### DIFF
--- a/.github/workflows/intent-tests.yml
+++ b/.github/workflows/intent-tests.yml
@@ -90,3 +90,84 @@ jobs:
           name: intent-tests-traces
           path: tests/intent/test-results/
           retention-days: 7
+
+  # Tier-2 BILLING suite (scenarios.md T2-BILL-1..9). Split into its own job
+  # because it boots the stack with Stripe test mode + enforcement on, needs
+  # the Stripe CLI, and self-signs webhook deliveries. Provisional and
+  # non-blocking like the suite above. Requires a Stripe TEST secret key in the
+  # STRIPE_TEST_SECRET_KEY repo secret; with no key the suite skips itself
+  # cleanly (globalSetup detects it and marks every spec skipped, not failed).
+  intent-billing:
+    name: intent-billing (provisional)
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    continue-on-error: true
+    services:
+      postgres:
+        image: postgres:16-alpine
+        env:
+          POSTGRES_USER: proliferate
+          POSTGRES_PASSWORD: localdev
+          POSTGRES_DB: proliferate
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U proliferate -d proliferate"
+          --health-interval 5s
+          --health-timeout 3s
+          --health-retries 10
+      redis:
+        image: redis:7-alpine
+        ports:
+          - 6379:6379
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 5s
+          --health-timeout 3s
+          --health-retries 10
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-node@v6
+        with:
+          node-version: "22"
+      - uses: pnpm/action-setup@v6
+        with:
+          version: 10
+      - uses: astral-sh/setup-uv@v7
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+
+      - name: Install workspace dependencies
+        run: pnpm install --frozen-lockfile
+      - name: Install server (venv, matches local layout)
+        run: cd server && uv venv .venv --python 3.12 && uv pip install -e ".[dev]"
+      - name: Install Playwright chromium
+        run: pnpm -C tests/intent exec playwright install --with-deps chromium
+
+      - name: Install Stripe CLI
+        run: |
+          curl -fsSL https://packages.stripe.dev/api/security/keypair/stripe-cli-gpg/public \
+            | gpg --dearmor | sudo tee /usr/share/keyrings/stripe.gpg >/dev/null
+          echo "deb [signed-by=/usr/share/keyrings/stripe.gpg] https://packages.stripe.dev/stripe-cli-debian-local stable main" \
+            | sudo tee /etc/apt/sources.list.d/stripe.list
+          sudo apt-get update && sudo apt-get install -y stripe
+
+      - name: Run tier-2 billing suite
+        env:
+          USE_EXISTING_POSTGRES: "1"
+          USE_EXISTING_REDIS: "1"
+          LOCAL_PGHOST: 127.0.0.1
+          TIER2_INTENT_SKIP_RUNTIME: "1"
+          CI: "true"
+          # Stripe TEST key (sk_test_...). Absent → the suite skips itself.
+          STRIPE_TEST_SECRET_KEY: ${{ secrets.STRIPE_TEST_SECRET_KEY }}
+        run: pnpm -C tests/intent run test:billing
+
+      - name: Upload Playwright traces on failure
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: intent-billing-traces
+          path: tests/intent/test-results/
+          retention-days: 7
--- a/specs/developing/testing/README.md
+++ b/specs/developing/testing/README.md
@@ -1,0 +1,232 @@
+# Testing Standard
+
+How tests are organized across the repo: what each tier owns, where test files
+live, what gates merges vs releases, and how a change decides which tests it
+must add. `flows.md` (sibling) is the registry of end-to-end flows that must
+never break; this doc is the law those flows are enforced under.
+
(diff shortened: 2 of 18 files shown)


pablonyx:
Local full-suite run of the tier-2 billing suite against the real stack (profile `t2billing`, Stripe TEST mode, `CLOUD_BILLING_MODE=enforce`, real accounting/reconciler/seat passes out-of-process). Three consecutive clean runs after the fixes in 82b2ddeb0.

## Scorecard (25 tests, 9 scenario groups)

| Group | Result |
|---|---|
| T2-BILL-1 checkout → grants → consumption → cut-off → reactivate (3 tests) | green |
| T2-BILL-2 plan limits + policy gates | green |
| T2-BILL-3 seats: invite/remove/re-invite (2 tests) + **pinned known-bug #1044** | green + expected-fail pin |
| T2-BILL-4 team checkout: pending intent + replayed activation idempotent (2 tests) | green |
| T2-BILL-5 compute overage: bill to cap → write off → block; disabled → cutoff; cap validation (3 tests) | green |
| T2-BILL-6 LLM credits: exhaustion, admin caps, fail-closed top-up (3 tests) | green |
| T2-BILL-7 webhook robustness: duplicate, concurrent, out-of-order (3 tests) | green |
| T2-BILL-8 subscription edges: payment_failed hold, cancel mid-period + grace, **pinned Finding 7 / #1032**, no-trial-for-password-account (4 tests) | green + expected-fail pin |
| T2-BILL-9 usage surfaces: summary/timeseries/llm-balance, by-user attribution, UI render (3 tests) | green |

Both known-bug pins pass as pins (assert the current buggy behavior + `known-bug` annotation + issue link; they invert when the fixes land):
- #1032 (Finding 7): `customer.subscription.deleted` applies a `payment_failed` hold even after clean voluntary cancellation.
- #1044 (new, found by this run): two active org subscriptions → `MultipleResultsFound` in `maybe_create_org_seat_adjustment` (missing `.limit(1)`) → **every membership change on the org 500s**. Filed with repro; test stages the condition, pins the 500, cleans up.

## Test clocks / Stripe posture
Real Stripe test mode end-to-end — no mocked Stripe anywhere. Every customer is created on a test clock; subscriptions/invoices are real test-mode objects; webhook deliveries are self-signed envelopes around real objects (per the harness design). Time-dependent state that has no Stripe clock to ride (grant `effective_at`, period backdates) uses the suite's established direct-DB time-shift convention.

## Notable test-side fixes (details in 82b2ddeb0)
- `/v1` route prefixes (billing/org/users/agent-gateway) — the 404s behind the initial 24/24 red.
- GitHub product-readiness gate: billing surfaces require it even in single-org mode; the harness seeds the legacy `oauth_account` row (not `auth_identity`, so free-trial lazy issuance stays off).
- Per-run billing-state reset in globalSetup (profile DB persists by design; counts were accumulating across runs).
- T2-BILL-4 replay now drives REAL team-checkout activation (trialing sub with full metadata on the intent's customer); the old synthetic object was 400-by-design.
- T2-BILL-1 cut-off asserts the overage-off posture (`overage_disabled`) — subscription sync auto-enables overage for pro, and a healthy pro subject with overage on deliberately rides overage rather than blocking.

Stripe test-mode footprint this session: 246 test clocks, 24 customers, 13 subscriptions, 25 checkout sessions (all test mode, all on test clocks where applicable).

pablonyx:
CI update: e9ea994e2 fixed the two CI-only boot gaps (price-catalog provisioning now passes the resolved key via STRIPE_API_KEY since runners have no stripe login config; enforce-mode boot injects an E2B_API_KEY placeholder — tier-2 billing never provisions a sandbox). The intent-billing job now runs the full suite for real and passes (6m0s), alongside all other checks. Verified locally 25/25 with the ambient E2B key unset to match the runner env.

Issue #1063, opened 2026-07-09
T3-CFG-1: live-config advertises 'effort' as settable but the session rejects applying it (SESSION_CONFIG_REJECTED)


pablonyx:
Surfaced by the tier-3 T3-CFG-1 scenario (live config options apply in an existing session) on fresh `main`, running against the `t3local` profile with the fresh anyharness runtime (sha `009d69e6`).

## What happens
On a claude session (opened on `us.anthropic.claude-sonnet-4-6` — the account classifies claude to `["bedrock"]`), `GET /v1/sessions/{id}/live-config` returns `normalizedControls` with three controls, all `settable: true`:

- `model` (current `us.anthropic.claude-sonnet-4-6`)
- `mode` (current `default`)
- `effort` (current `low`, values `default/low/medium/high/max`)

Cycling `model` and `mode` round-trips fine (set → readback matches). But applying **`effort`**:

```
POST /v1/sessions/{id}/config-options  { rawConfigId: "effort", value: ... }
-> 400 {"code":"SESSION_CONFIG_REJECTED","detail":"Config option 'effort' is not exposed by the active session."}
```

So the live-config menu advertises `effort` as settable, but the session's config-apply surface rejects it as "not exposed by the active session" — a menu/apply mismatch.

## Why it matters
T3-CFG-1 exists precisely to catch "occasionally this breaks" config-apply regressions. A control shown as settable in the menu that the session then refuses is exactly that: the UI would offer `effort`, the user would pick a value, and the apply would fail.

## Notes
- Reproduced 2026-07-09 on `us.anthropic.claude-sonnet-4-6`. Not yet checked whether `effort` is exposed for other models/contexts — it may be that `effort` is only valid for some models but the `normalizedControls` projection marks it settable unconditionally.
- Test-side: T3-CFG-1 now records advertised-but-rejected controls and reports a diagnosed expected-fail (tracked, non-blocking) instead of a hard red, still asserting the controls that do round-trip.

pablonyx:
Resolved by the maintained Claude adapter/catalog upgrade that landed in #1323 and is present on current `main` (`22cf3f0ebb6f0d0d87db588fa1bcb865b54055c2`).

Exact current evidence:

- `catalogs/agents/catalog.json` pins Claude `0.59.0-proliferate.1` at adapter commit `26f9ee7a0049507bff5476ce390695515ce92840` with Claude Code `2.1.212`.
[...]

Issue #1051, opened 2026-07-09
T3-CFG-1 (local lane) red: claude haiku alias resolves to invalid Anthropic model id claude-haiku-4-5-20251001


pablonyx:
## What's failing

`T3-CFG-1` (local lane, `tests/release/src/scenarios/t3-cfg-1.ts`) fails every run
against the t3local profile, before it ever reaches the config-cycling logic
it's supposed to exercise:

```
AssertionError [ERR_ASSERTION]: T3-CFG-1: session must survive every config switch
    at runLocalLane (tests/release/src/scenarios/t3-cfg-1.ts:95:12)
```

The assertion at line 95 just checks the session isn't `errored` after the
config-cycling loop, but the session actually errors on turn 1 -- the very
first prompt the scenario sends, before any config switch happens.

## Root cause (from the runtime log)

Scenario creates a claude session with `modelId: "haiku"` and prompts it.
`anyharness.log` shows the agent process (claude-agent-acp / claude CLI)
resolves the `haiku` alias to the dated snapshot `claude-haiku-4-5-20251001`,
which Anthropic's API now rejects outright:

```
session.actor.prompt.conn_failed ... error=Internal error: API Error
(claude-haiku-4-5-20251001): 400 The provided model identifier is invalid..
Run --model to pick a different model.: {
  "errorKind": "model_not_found"
}
```

This is a `model_not_found` from Anthropic itself (not an auth/permission
error), against the `anthropic-api` auth context. The runtime's own Rust code
has no mapping of `haiku` -> `claude-haiku-4-5-20251001` outside the
gateway-only surfaces (`gateway_resolver.rs` / `agent_gateway_catalog.rs`,
which weren't in play here -- `AGENT_GATEWAY` was off for this run), so the
alias resolution to that specific dated snapshot happens inside the pinned
`claude-agent-acp` adapter / claude CLI itself
(`catalogs/agents/catalog.json` pins `agentProcess.version: 0.44.0`,
`git_ref: 3ff484e671de5fe9275d2e7596d683df06b99e14`; native CLI `2.1.181`).
Best guess: Anthropic has retired/renamed the `-20251001` haiku snapshot and
the pinned adapter/CLI's internal alias table hasn't caught up.

## Not caused by the runtime binary rebuild

Confirmed this is pre-existing, not a regression from rebuilding the t3local
runtime binary (infra hygiene pass, 2026-07-09): the exact same assertion and
error reproduce against both:
[...]

Issue #1042, opened 2026-07-09
tier-3 T3-SEC-MAT-1/sandbox: org/workspace secret + in-sandbox materialization assertions not implemented


pablonyx:
Found by tier-3 T3-SEC-MAT-1 after #1023 unblocked the sandbox lane.

Now that the github_link gate is open, the personal secret PUT succeeds for real, but the scenario throws before finishing:

```
Error: T3-SEC-MAT-1: personal secret PUT succeeded (gate lifted) but the rest of this scenario (org secret, workspace file secret, fresh cloud workspace + in-sandbox file assertions) is not yet implemented — finish it now that the gate is open.
    at runReal (tests/release/src/scenarios/t3-sec-mat-1.ts:74)
```

This is a known TODO left in the scenario itself (the message says as much) rather than a product bug — flagging it per the "file an issue for every red" instruction so it's tracked now that the gate lifting makes it actionable. Test file: tests/release/src/scenarios/t3-sec-mat-1.ts.

pablonyx:
Substantially fixed in #1098 (merged), but not fully closing this — one sub-part remains gated behind a separate, already-tracked issue.

**Done and verified live-green** (against a real E2B sandbox, real materialization, real files):
- Personal + org env-var secret PUT, materialization polled to `ready`.
- In-sandbox ground-truth verification (this was the actual point of the scenario — previously only the API's `materialization.status` field was trusted): `{PROLIFERATE_HOME}/secrets/global.env` read directly out of the sandbox via the E2B SDK (no product route exposes an arbitrary file read), asserted to contain both merged vars, plus the manifest's per-var sha256 checked against `sha256(value)` computed locally.
- Update propagation: PUT a new value, assert the status cycle (`ready` → `pending` → `ready`), assert the sandbox file actually updates.

Gated by `RELEASE_E2E_E2B_API_KEY` (already wired for CI): without it, the personal/org PUT + materialization-to-ready still runs for real, and only the byte-level in-sandbox assertions report a diagnosed expected-fail.

[...]

Issue #1041, opened 2026-07-09
tier-3 T3-PROV-2/sandbox: pause + reconnect-state-intact assertions not implemented


pablonyx:
Found by tier-3 T3-PROV-2 after #1023 unblocked the sandbox lane.

Now that the github_link gate is open, GET /cloud-sandbox + POST /cloud-sandbox/wake succeed for real, but the scenario throws before finishing:

```
Error: T3-PROV-2: GET /cloud-sandbox + POST /cloud-sandbox/wake succeeded (gate lifted) but pause + reconnect-state-intact assertions are not yet implemented — finish them now that the gate is open.
    at runReal (tests/release/src/scenarios/t3-prov-2.ts:52)
```

This is a known TODO left in the scenario itself (the message says as much) rather than a product bug — flagging it per the "file an issue for every red" instruction so it's tracked now that the gate lifting makes it actionable. Test file: tests/release/src/scenarios/t3-prov-2.ts.

pablonyx:
Fixed in #1098 (merged). T3-PROV-2's pause + reconnect-state-intact assertions are implemented and verified live against a real E2B sandbox on `--lane local`: front-door reconnect via the real anyharness gateway proxy, a direct-E2B pause (there is no product pause endpoint at all — pause only ever arrives via E2B's own idle timeout or the billing reconciler, so it's driven via the E2B SDK, resolving the provider sandbox purely from E2B's own metadata list, no DB access needed), ground-truth state verification, resume, and a filesystem marker proven intact across the whole cycle.

Gated by `RELEASE_E2E_E2B_API_KEY` (already wired for CI — `release-e2e.yml` maps the repo secret `E2B_API_KEY` to it for the staging job): without it, the scenario still runs the product-flow half for real and reports a diagnosed expected-fail for the pause/resume half rather than skipping silently.

One live finding surfaced while verifying: the staging `e2e-tests` durable org currently has zero remaining cloud-sandbox credits (`billing_credits_exhausted`, `remaining_seconds: 0`), which 402s the sandbox `ensure`/wake path before a personal cloud sandbox can even be created. Classified as a blocked run (not red) in the scenario itself, but the org needs a credits top-up to actually run this scenario — and every other sandbox-lane T3 scenario — green on `--lane staging`. Flagging separately since it's an ops/fixture action, not something this PR can fix.

[...]

PR #1065, opened 2026-07-09
tests(release): reconcile tier-3 pins with tonight's merges

pablonyx:
Ran the full tier-3 suite against fresh `main` (local lane: server on 8086, AnyHarness runtime on 8542 built from the fresh binary sha `009d69e6`) and reconciled every scenario's pin to the real, as-built behavior. Full suite now reports 3 green, 2 blocked, 12 expected-fail, 0 red. Typecheck and the 26 package unit tests pass.

## Changes
- **T3-CFG-1**: stop hardcoding the bare model id `haiku`. On a Bedrock-classified account (t3local classifies claude to `["bedrock"]`, the flag arriving via the readiness/auth overlay rather than the profile launch.env), #1046 correctly gates bare ids with `SESSION_MODEL_GATED`. Resolve the account's real cheapest working model via `catalogHarnesses` (lands on `us.anthropic.claude-sonnet-4-6`). This is the test-side half of #1051 (closed as a duplicate of #1037). Cycling controls then surfaced a real menu/apply mismatch: `effort` is advertised `settable` in live-config but the session rejects applying it (`SESSION_CONFIG_REJECTED`). Recorded as a diagnosed expected-fail and filed as #1063.
- **T3-CHAT-1 / T3-WT-1 / T3-UPDATE-1 sandbox**: the `current_product_user` block was stale. That gate is bypassed in single-org mode (verified: the durable user gets HTTP 200 from `GET /cloud-sandbox` despite `productReady=false`). Replaced the stale reason with the real remaining gap (the sandbox-lane driver is not implemented) as a tracked expected-fail. Updated the #1024 opencode note: #1034 resolved it, opencode is green with an explicit gateway id.
- **T3-PROV-2 / T3-SEC-MAT-1 / T3-INT-1**: their first real call now succeeds once the gate lifts, so the trailing "finish me" throws became tracked expected-fails against #1041 / #1042 instead of reds.
- **T3-REPO-1**: the installation gate cleared (installation 145413813 on proliferate-e2e). Added handling for the two remaining environmental outcomes as expected-fails, `github_repo_access_required` (the seeded user lacks repo access) and `github_app_refresh_failed` (seed refresh-token rotation across a suite pass). Run in isolation, the PUT now succeeds end to end.
- **T3-BILL-2**: the #1036 pin (bypass-1 direct-API resume must 402 after #1045) stays flipped, but the durable org has no enrolled or drainable grants, so the gate cannot be exercised. Reports a diagnosed expected-fail instead of a red assertion.

### DIFF
--- a/tests/release/src/fixtures/github-app-seed.ts
+++ b/tests/release/src/fixtures/github-app-seed.ts
@@ -132,6 +132,35 @@ export function isGithubAppRepoNotCoveredError(error: unknown): boolean {
   return isCloudErrorCode(error, "github_app_repo_not_covered");
 }
 
+/**
+ * The GitHub *user* (the seeded App user-to-server identity) does not have
+ * access to the target repo — distinct from installation coverage. On t3local
+ * the seeded identity is `pablonyx`, which is not a collaborator on
+ * `proliferate-e2e/e2e-fixture`, so the authority chain 409s here once the
+ * authorization + installation gates are both satisfied.
+ */
+export function isGithubRepoAccessRequiredError(error: unknown): boolean {
+  return isCloudErrorCode(error, "github_repo_access_required");
+}
+
+/**
+ * The server could not refresh the seeded App user-to-server authorization
+ * (502 `github_app_refresh_failed`). Seed refresh tokens rotate on every use,
+ * so running multiple seed-consuming scenarios in one suite pass (e.g.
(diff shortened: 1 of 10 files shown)


pablonyx:
Superseded by the rebuilt-on-main version (base branch here is stale post-#1009 squash and this branch predates #1062). Content carried over minus the #1062 file reverts.

PR #949, opened 2026-07-04
fix(cloud): stalled-materialization detection + recovery (no more infinite spinner)

pablonyx:
## Problem

When workspace provisioning dies between the row insert (service.py:187) and the anyharness_workspace_id write (service.py:219), the workspace row is stuck permanently: status is derived purely from `anyharness_workspace_id IS NULL` returning `"materializing"` forever. The desktop poller (3s interval, bare catch on errors, no deadline) spins indefinitely showing an infinite spinner with no recovery path.

## What changed

**Server (`server/proliferate/server/cloud/workspaces/`)**
- `_workspace_status` now checks `created_at` age: if the workspace has been without an `anyharness_workspace_id` for longer than 15 minutes (900s, exceeding the 600s repo-clone timeout), it reports `"materialization_stalled"` instead of `"materializing"`.
- The workspace payload includes `statusDetail` and `lastError` for stalled rows so the client shows a meaningful error message.
- Added `"materialization_stalled"` to the `CloudWorkspaceStatus` Literal type.

**Cloud SDK (`cloud/sdk/src/client/workspaces.ts`)**
- `normalizeCloudWorkspaceStatus` maps `"materialization_stalled"` to `"error"`, so existing desktop error-handling flows catch it immediately (poller stops, error screen appears with delete/retry affordance).

**Desktop poller (`apps/desktop/src/hooks/chat/lifecycle/use-cloud-workspace-polling.ts`)**
- Added a 16-minute client-side deadline as belt-and-suspenders: if the pending workspace entry has been waiting longer than this and the server still reports a non-terminal status, the poller forcibly transitions to "failed" with a timeout message.
- This guards against the case where network issues delay delivery of the server's stalled signal.

**Recovery path**
- The existing `DELETE /v1/cloud/workspaces/:id` endpoint already works on any workspace state (including stalled rows). Users see the error state with guidance to delete and recreate.

**Tests**
- Added `server/tests/unit/test_cloud_workspace_status.py` with 6 unit tests covering all branches of the staleness detection logic.

## How to verify

1. Create a cloud workspace row manually in the DB with `anyharness_workspace_id = NULL` and `created_at` older than 15 minutes.
2. Hit GET `/v1/cloud/workspaces` — the workspace should report `status: "materialization_stalled"` with appropriate `statusDetail`/`lastError`.
3. On the desktop, the SDK normalizes this to `"error"`, stopping the poller and showing the error/retry screen.
4. Deleting the workspace via the existing action works normally.

## Follow-ups

[...]

PR #944, opened 2026-07-04
feat(cloud): org-owned shared sandbox profiles v0 (T1, draft)

pablonyx:
## Problem

The `cloud_sandbox` table supports only personal (per-user) ownership. The team-and-web spec (T1) calls for surfacing org-owned sandbox profiles so organization members can share cloud sandboxes under org billing.

## What changed

**Server (API + store + migration):**
- Migration `bb68e6f07040`: adds `owner_scope`, `organization_id`, `created_by_user_id`, `billing_subject_id`, `display_name` columns to `cloud_sandbox`; adds check constraints and unique indexes for org sandboxes.
- ORM model (`sandboxes.py`): new columns, updated constraints.
- Store (`cloud_sandboxes.py`): implements `list_organization_cloud_sandboxes`, `ensure_organization_cloud_sandbox`, `load_organization_cloud_sandbox`; fixes `acquire_cloud_sandbox_owner_lock` to support org scope.
- Service (`org_sandbox_profiles/service.py`): create (binds org billing subject), list, get. Boundary-clean (no ORM imports).
- API (`org_sandbox_profiles/api.py`): `GET/POST /v1/cloud/organizations/{id}/sandbox-profiles`, `GET .../sandbox-profiles/{sandbox_id}`. Create is org-admin gated; list/get is member-visible.

**Agent-auth admission:** Verified with a test that the existing `render_agent_auth_state` path works unchanged for org profiles. Selections remain per-user — a member starting a session on an org sandbox uses their own agent-auth selections.

**Cloud SDK:** Regenerated OpenAPI TypeScript client; added `org-sandbox-profiles.ts` client module with `listOrgSandboxProfiles`, `createOrgSandboxProfile`, `getOrgSandboxProfile`.

**Desktop UI:** `OrgSandboxProfilesSection` component in org settings pane — lists profiles with status badges, create form (admin-only).

## How to verify

```bash
cd server
DEBUG=1 JWT_SECRET=test CLOUD_SECRET_KEY=test \
  uv run --extra dev pytest -xvs \
    tests/unit/test_org_sandbox_profiles_service.py \
    tests/unit/test_org_sandbox_agent_auth_admission.py

python3.12 scripts/check_server_boundaries.py
```

## Follow-ups

- **Member-access policy**: who may START sessions on an org profile (currently any member can view, only admin can create — session-start admission policy is T2 work).
- **Org billing display**: surface the org billing subject's usage/grants in the org settings UI alongside profiles.
[...]

PR #1116, opened 2026-07-10
fix(server): self-hosted cloud-workspace safety, actionable errors, and callback hardening

pablonyx:
Task 03 (self-hosting agent pack): make the GitHub App + E2B cloud-workspace path safe and coherent on self-managed servers. Base commit `983ce30fe`. **Owned paths only** — no `server/deploy/**`, Caddy, Compose, bootstrap, update, or env-template edits; deployment requirements are in the task report.

## What this fixes (each an acceptance criterion)

1. **Incomplete E2B config can't crash-loop the API.** `E2B_API_KEY` set with `E2B_TEMPLATE_NAME` empty previously raised at FastAPI lifespan startup → total outage of a previously-healthy instance. Startup now logs a warning and stays up; cloud-provisioning **requests** fail with an actionable `e2b_template_not_configured` (503) naming the missing var, and the E2B integration refuses to silently boot the internal default template in production.
2. **Errors name the missing requirement without secrets.** New `e2b_template_not_configured`, and the checkout-conflict copy names the blocker without leaking sandbox paths.
3. **Source/origin translated at the owning boundary.** Product `source` (`desktop`/`web`/`mobile`) → valid AnyHarness `origin.entrypoint`; `source=web`/`mobile` no longer 422 `unknown variant`.
4. **Retry is generated-file-safe.** The repo-checkout script registers `.proliferate/` as locally ignored, so a retry after a transient failure doesn't treat Proliferate-generated files (`workspace.env`, `workspace.manifest.json`) as a dirty checkout. Genuine user work still blocks a reset.
5. **Actionable conflict, not opaque 500 (second same-repo workspace).** Checkout guard exit codes (dirty / local-commits / non-git) map to a typed conflict; create-workspace returns a specific `cloud_repo_checkout_conflict` (409) or `cloud_materialization_busy` (503) instead of a raw 500.
6. **Callbacks return to a self-host-safe target + refresh scope.** The GitHub App Setup URL callback tolerates a missing `state` (GitHub-initiated post-install/update redirect), refreshing effective repo scope instead of 422; callbacks land on a server-rendered page when no web frontend exists (API-only self-host) instead of a 404 web route.

## Tests
- New: `test_workspace_origin_mapping`, `test_cloud_provisioning_config`, `test_repo_checkout_error_mapping`, `test_repo_checkout_generated_files` (real-git behavioral check), `test_github_app_callback_return`; create-path 409/503 mapping added to `test_cloud_workspace_reconnect_error`.
- Updated: `test_e2b_runtime` and `test_cloud_startup` for the new no-crash contract.
- Server pytest (unit + touched integration) green; ruff + repo-shape clean. No Rust/SDK/generated files touched.

## Not in this PR (handed off in the report)
Deployment-owned turnkey items (Redis for the materialization lock, Compose/Caddy/env-template wiring, App-key mounts, webhook route) are specified in `self-hosting-notes/agent-pack/reports/03-cloud-workspaces.md` for the Deployment owner. Live end-to-end verification status is recorded there.

### DIFF
--- a/server/proliferate/config.py
+++ b/server/proliferate/config.py
@@ -373,6 +373,40 @@ class Settings(BaseSettings):
     proliferate_target_artifact_base_url: str = ""
 
     @property
+    def cloud_provisioning_configured(self) -> bool:
+        """True when E2B is fully configured to provision cloud sandboxes.
+
+        Requires BOTH an API key and a template name. In debug/dev the template
+        name may be blank (a built-in default is used), so any API key suffices.
+        """
+        if not self.e2b_api_key.strip():
+            return False
+        if self.debug:
+            return True
+        return bool(self.e2b_template_name.strip())
+
+    @property
+    def cloud_provisioning_config_error(self) -> str | None:
+        """Actionable reason cloud provisioning is unavailable, or None if ready.
+
+        Names the missing requirement without echoing any secret value. Used
+        both to log a startup warning and to fail cloud-provisioning requests
+        with a specific error instead of crash-looping the API or booting the
+        wrong E2B template.
+        """
+        if self.debug:
+            return None
+        if not self.e2b_api_key.strip():
+            return None  # cloud provisioning intentionally disabled
+        if not self.e2b_template_name.strip():
+            return (
+                "E2B_API_KEY is set but E2B_TEMPLATE_NAME is empty. Set "
+                "E2B_TEMPLATE_NAME to the published runtime template for this "
+                "deployment to enable cloud workspaces."
+            )
+        return None
+
+    @property
     def single_org_mode(self) -> bool:
         if self.single_org_mode_override is not None:
             return self.single_org_mode_override
--- a/server/proliferate/integrations/sandbox/e2b.py
+++ b/server/proliferate/integrations/sandbox/e2b.py
@@ -250,7 +250,20 @@ class E2BSandboxProvider:
 
     def _template_name(self) -> str:
         configured = settings.e2b_template_name.strip()
-        return configured or E2B_TEMPLATE_NAME
+        if configured:
+            return configured
+        # In debug/dev a built-in default template is acceptable. In production
+        # (self-host included) an unset template must NOT silently fall back to
+        # the internal default — the operator's E2B account does not have it, so
+        # boots would fail confusingly. Fail with an actionable error naming the
+        # missing requirement instead.
+        if settings.debug:
+            return E2B_TEMPLATE_NAME
+        raise E2BRuntimeError(
+            "E2B_TEMPLATE_NAME is not configured. Set E2B_TEMPLATE_NAME to the "
+            "published runtime template for this deployment before provisioning "
+            "cloud sandboxes."
+        )
 
     def _create_sandbox(self, metadata: dict[str, str] | None) -> SandboxHandle:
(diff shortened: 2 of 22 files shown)


pablonyx (commented):
## GATE: ROUTE_BACK — 🚫 BLOCKING, do not merge until resolved

**Core P0 #3 fix is correct and well-tested:** partial E2B config no longer crash-loops startup (`main.py::_validate_e2b_template_configuration` warns + returns), and cloud-provisioning requests fail with an actionable `503 e2b_template_not_configured` (`cloud_sandboxes/service.py::require_cloud_provisioning_configured`), with the built-in template default allowed only in debug. GitHub App callback hardening is sound — no open-redirect on the stateless path (return targets are always server-derived). Both behaviors are genuinely asserted in tests.

### Blocking
1. **[CI — mechanical] `lint` = `ruff format --check` wants reformat** on:
   - `proliferate/server/cloud/materialization/materialize/repo_environment.py`
   - `proliferate/server/cloud/workspaces/domain/origin.py`
   - `tests/integration/test_cloud_startup.py`
   Run `ruff format proliferate/ tests/` from `server/` and push. (`ruff check` passes; only `format --check` is red — the report's "clean" claim missed it.)
2. **[Correctness ★ — FIX NOW, in this PR] #948 preclone TypeError is CONFIRMED LIVE in this track's owned scope.** `server/proliferate/server/cloud/materialization/materialize/sandbox.py:55-59` still calls `materialize_repo_environment_in_context(db, ctx=ctx, repo_environment=repo_environment)`, but the current signature (`repo_environment.py:67-74`) requires `repo_environment_id` / `materialization_id` / `attempt_updated_at` → raises `TypeError` on every invocation with ≥1 repo environment. Scheduled from `github_app/service.py:301` (`schedule_materialize_sandbox`) inside `complete_github_app_user_authorization_callback` — a function in a file this PR edits. Effect: background pre-clone silently fails after every GitHub re-auth and the user pays the full inline clone cost. Fix in this PR (not a fast-follow).
3. **[Correctness/gap] #949 orphan-materializing is not addressed.** The error-mapping work here covers the *synchronous, exception-raising* checkout class, not the *crash-with-no-exception* class where the row is stuck `"materializing"` forever (desktop infinite spinner). Handle it or explicitly acknowledge/defer with a tracked issue.
4. **[Reconciliation] Add an explicit adopt/supersede/reject note vs the overlapping drafts** #948 / #949 / #950 / #890 (and OOS #944) — the report has none. #950 overlaps `e2b.py` / `cloud_sandboxes/service.py` at different functions (auto-merges) but is already `CONFLICTING` vs main; note it for sequencing.

### Non-blocking
- `intent-billing (provisional)` failure is `specs/billing/seats.spec.ts:45` (T2-BILL-3 seat proration), unrelated to this diff and not a required check.
- `config.py` overlap with #1119 is non-adjacent (auto-merges); cross-track predicate reconcile (C1) is integrator-owned.

**Deployment requirements captured for Track 01:** durable `redis` service (materialization lock), `E2B_TEMPLATE_NAME=` in `.env.production.example` + template-build path, 7 `GITHUB_APP_*` + secure PEM mount + 4 public routes + the new `/auth/github-app/connected` route, half-config/missing-Redis doctor check, E2B template arch. Live end-to-end is honestly BLOCKED pending that integration — good.

PR #1144, opened 2026-07-13
fix(sdk): isolate AnyHarness cache and runtime queries

pablonyx:
## What changed

- Added an explicit AnyHarness cache scope derived from the API deployment and authenticated actor.
- Split runtime-scoped query identity from logical workspace-scoped identity across SDK React and Desktop cache consumers.
- Kept workspace cache identity stable across temporary cloud gateway token and URL refreshes.
- Disabled local AnyHarness inventory queries when a host has no local runtime, while preserving cloud workspace routing.
- Added regression coverage for actor/deployment isolation, gateway credential rotation, Web-style null-runtime routing, and Desktop workspace caching.

## Why

Workspace requests could use a cloud gateway while their React Query keys were rooted in the Desktop local runtime. That made the shared Desktop/Web boundary vulnerable to stale or cross-deployment cache reuse. The Desktop workspace inventory hook could also attempt local runtime calls when only cloud access was available.

## Impact

This establishes the cache and routing boundary needed before moving the Desktop product surface into the shared Web/Desktop client. Cloud queries no longer depend on a fake local runtime, user and self-host deployments are isolated, and dynamic local runtime queries retain transport identity.

## Validation

- make build
- pnpm --filter @anyharness/sdk-react test (22 passing)
- Desktop TypeScript typecheck
- Focused Desktop cache, routing, session, terminal, and workspace workflow tests
- Fresh query-scope profile: API, AnyHarness, Desktop renderer, and hosted Web all returned HTTP 200; the native Desktop app opened and successfully exercised workspace, repo-root, auth, and health requests
- Profile stopped cleanly after the smoke

## Known baseline

The repository-wide Desktop test entrypoint currently stops on existing design-system violations, and direct full Vitest has unrelated failures reproducible on main. All tests affected by this diff pass.

### DIFF
--- a/anyharness/sdk-react/src/hooks/workspaces.ts
+++ b/anyharness/sdk-react/src/hooks/workspaces.ts
@@ -43,10 +43,6 @@ interface WorkspaceQueryOptions {
   requestOptions?: AnyHarnessRequestOptions;
 }
 
-function useWorkspaceCacheScopeKey() {
-  return useAnyHarnessCacheScopeKey();
-}
-
 export function useRuntimeWorkspacesQuery(options?: RuntimeQueryOptions) {
   const runtime = useAnyHarnessRuntimeContext();
   const runtimeUrl = runtime.runtimeUrl?.trim() ?? "";
@@ -64,7 +60,7 @@ export function useRuntimeWorkspacesQuery(options?: RuntimeQueryOptions) {
 
 export function useWorkspaceQuery(options: WorkspaceQueryOptions) {
   const workspace = useAnyHarnessWorkspaceContext();
-  const cacheScopeKey = useWorkspaceCacheScopeKey();
+  const cacheScopeKey = useAnyHarnessCacheScopeKey();
   const workspaceId = options.workspaceId ?? workspace.workspaceId;
 
   return useQuery({
@@ -184,7 +180,7 @@ export function useUpdateWorkspaceDisplayNameMutation() {
 
 export function useRetireWorkspacePreflightQuery(options?: WorkspaceQueryOptions) {
   const workspace = useAnyHarnessWorkspaceContext();
-  const cacheScopeKey = useWorkspaceCacheScopeKey();
+  const cacheScopeKey = useAnyHarnessCacheScopeKey();
   const workspaceId = options?.workspaceId ?? workspace.workspaceId;
 
   return useQuery({
@@ -204,7 +200,7 @@ export function useRetireWorkspacePreflightQuery(options?: WorkspaceQueryOptions
 
 export function usePurgeWorkspacePreflightQuery(options?: WorkspaceQueryOptions) {
   const workspace = useAnyHarnessWorkspaceContext();
-  const cacheScopeKey = useWorkspaceCacheScopeKey();
+  const cacheScopeKey = useAnyHarnessCacheScopeKey();
   const workspaceId = options?.workspaceId ?? workspace.workspaceId;
 
   return useQuery({
@@ -330,7 +326,7 @@ export function useRetryPurgeWorkspaceMutation() {
 
 export function useDetectProjectSetupQuery(options?: WorkspaceQueryOptions) {
   const workspace = useAnyHarnessWorkspaceContext();
-  const cacheScopeKey = useWorkspaceCacheScopeKey();
+  const cacheScopeKey = useAnyHarnessCacheScopeKey();
   const workspaceId = options?.workspaceId ?? workspace.workspaceId;
 
   return useQuery({
@@ -350,7 +346,7 @@ export function useDetectProjectSetupQuery(options?: WorkspaceQueryOptions) {
 
 export function useSetupStatusQuery(options?: WorkspaceQueryOptions & { refetchWhileRunning?: boolean }) {
   const workspace = useAnyHarnessWorkspaceContext();
-  const cacheScopeKey = useWorkspaceCacheScopeKey();
+  const cacheScopeKey = useAnyHarnessCacheScopeKey();
   const workspaceId = options?.workspaceId ?? workspace.workspaceId;
   const refetchWhileRunning = options?.refetchWhileRunning ?? true;
 
--- a/anyharness/sdk-react/src/hooks/workspaces.ts
+++ b/anyharness/sdk-react/src/hooks/workspaces.ts
@@ -6,7 +6,12 @@ import type {
   StartWorkspaceSetupRequest,
   UpdateWorkspaceDisplayNameRequest,
 } from "@anyharness/sdk";
-import { useAnyHarnessRuntimeContext, resolveRuntimeConnection } from "../context/AnyHarnessRuntime.js";
+import {
+  resolveRuntimeCacheScopeKey,
+  resolveRuntimeConnection,
+  useAnyHarnessCacheScopeKey,
+  useAnyHarnessRuntimeContext,
+} from "../context/AnyHarnessRuntime.js";
 import {
   useAnyHarnessWorkspaceContext,
   resolveWorkspaceConnectionFromContext,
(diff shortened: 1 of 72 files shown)


apps/desktop/src/lib/domain/auth/anyharness-cache-scope.ts
```diff
@@ -0,0 +1,20 @@
...
+  input: AnyHarnessCacheScopeInput,
+): string {
+  const apiBaseUrl = input.apiBaseUrl.trim() || "unknown-deployment";
+  const authUserId = input.authUserId?.trim() ?? "";
+
+  if (input.authStatus === "authenticated" && authUserId) {
+    return `${apiBaseUrl}::user:${authUserId}`;
+  }
+
+  return `${apiBaseUrl}::${input.authStatus}`;
```
pablonyx, line 19: All unauthenticated actors on the same deployment collapse to one shared scope (`${apiBaseUrl}::anonymous`), with no per-user/per-instance component. Since workspace-scoped keys no longer include `runtimeUrl` at all, two pre-login sessions against the same deployment that ever see the same `workspaceId` (dev bypass, seeded demo workspaces) share cache entries for sessions/files/git-status. If pre-login states can never issue workspace queries in practice, worth a comment saying so; otherwise consider mixing in a per-app-instance nonce for non-authenticated statuses.

anyharness/sdk-react/src/lib/query-keys.ts
```diff
@@ -1,188 +1,272 @@
...
+    "runtime",
+    normalizedRuntimeUrl,
+  ] as const;
+}
+
+export function anyHarnessWorkspaceKey(
+  cacheScopeKey: string | null | undefined,
+  workspaceId: string | null | undefined,
+) {
+  return [
```
pablonyx, line 21: Dropping `runtimeUrl` from workspace-scoped keys is the point of the PR for cloud gateway rotation, but it also means: same authenticated user, same `workspaceId` re-registered against a different/restarted local runtime → sessions/git-diff/file-tree served from the old runtime's cache until an explicit invalidation. If workspaceIds are guaranteed unique across local runtime rebinds, fine — but that invariant is now load-bearing and undocumented.

anyharness/sdk-react/src/lib/query-keys.ts
```diff
@@ -1,188 +1,272 @@
...
-  return ["anyharness", runtimeUrl?.trim() ?? ""] as const;
+export function anyHarnessCacheScopeKey(cacheScopeKey: string | null | undefined) {
+  return ["anyharness", cacheScopeKey?.trim() ?? ""] as const;
 }
 
-export function anyHarnessRuntimeHealthKey(runtimeUrl: string | null | undefined) {
-  return [...anyHarnessRuntimeKey(runtimeUrl), "health"] as const;
+export function anyHarnessRuntimeKey(
+  runtimeUrl: string | null | undefined,
+  cacheScopeKey?: string | null,
```
pablonyx: The runtime-scoped family takes `cacheScopeKey` as an optional trailing param with a silent `|| runtimeUrl` fallback, while the workspace-scoped family requires it as the leading param. A new runtime-scoped hook that forgets the arg compiles fine and keys purely on `runtimeUrl` — exactly the cross-actor bleed this PR is closing, reintroducible one forgotten argument at a time. Suggest making the scope segment required here too (all ~50 current call sites already pass it, per my trace).

apps/desktop/src/hooks/workspaces/cache/use-workspace-selection-cache.ts
```diff
@@ -61,7 +65,12 @@ export function useWorkspaceSelectionCache() {
       if (!workspaceId || nextIds.has(workspaceId)) {
         continue;
       }
-      for (const root of anyHarnessWorkspaceQueryKeyRoots(input.runtimeUrl, workspaceId)) {
+      const workspaceRoots = [
+        ...anyHarnessWorkspaceQueryKeyRoots(cacheScopeKey, workspaceId),
+        anyHarnessCoworkManifestKey(input.runtimeUrl, workspaceId, cacheScopeKey),
```
pablonyx, line 70: Cowork manifest/artifact keys were dropped from `anyHarnessWorkspaceQueryKeyRoots` and hand-appended back at this one call site. Any other/future caller of the roots helper (whose whole purpose is 'every workspace-scoped cache root') silently loses cowork cancellation. Root cause is that cowork manifest/artifact keys stayed runtime-scoped while their siblings moved under `anyHarnessWorkspaceKey` — migrating them would let the roots helper be complete again and delete this special case.

apps/desktop/src/lib/domain/auth/anyharness-cache-scope.ts
```diff
@@ -0,0 +1,20 @@
+import type { AuthClientStatus } from "./auth-state-mapping";
+
+interface AnyHarnessCacheScopeInput {
+  apiBaseUrl: string;
+  authStatus: AuthClientStatus;
+  authUserId: string | null;
+}
+
+export function buildAnyHarnessCacheScopeKey(
```
pablonyx, line 9: This builder is the single definition of what a cache identity boundary *is*, but it lives in desktop-only code while the scope is consumed in `@anyharness/sdk-react`. When Web merges into the shared client it needs byte-identical scope keys; putting the builder next to `resolveRuntimeCacheScopeKey` in the SDK avoids the inevitable duplicate-and-drift (or a cross-import from apps/desktop).

anyharness/sdk-react/src/hooks/workspaces.ts
```diff
@@ -37,17 +43,17 @@ interface WorkspaceQueryOptions {
   requestOptions?: AnyHarnessRequestOptions;
 }
 
-function useWorkspaceRuntimeUrl() {
-  const runtime = useAnyHarnessRuntimeContext();
-  return runtime.runtimeUrl?.trim() ?? "";
+function useWorkspaceCacheScopeKey() {
```
pablonyx: This one-line `useWorkspaceCacheScopeKey()` passthrough is copy-pasted verbatim into ~10 hook files (files/git/mobility/plans/pull-requests/reviews/sessions/terminals/session-pending-prompts/workspaces). It no longer adapts anything — either import `useAnyHarnessCacheScopeKey` directly (as agents.ts does) or, better, add one shared `useAnyHarnessRuntimeScope()` returning `{ runtimeUrl, cacheScopeKey }`, which would also collapse the 3-line derivation boilerplate repeated ~37× across the sdk hooks.

apps/desktop/src/hooks/workspaces/cache/use-pr-status-refresh.ts
```diff
@@ -17,8 +18,12 @@ const PR_STATUS_REFRESH_DEBOUNCE_MS = 250;
...
 const pendingRefreshTimers = new Map<string, ReturnType<typeof setTimeout>>();
 
-function refreshTimerKey(runtimeUrl: string, repoRootId: string): string {
-  return `${runtimeUrl}::${repoRootId}`;
+function refreshTimerKey(
+  runtimeUrl: string,
+  repoRootId: string,
+  cacheScopeKey?: string | null,
+): string {
+  return `${cacheScopeKey?.trim() || runtimeUrl}::${runtimeUrl}::${repoRootId}`;
```
pablonyx: Third copy of the `cacheScopeKey?.trim() || runtimeUrl` fallback rule (the others live in `resolveRuntimeCacheScopeKey` and inside `anyHarnessRuntimeKey`). If the normalization ever changes in one place, the debounce key drifts from the query keys it protects. Let the key builder / context resolver own the fallback and pass the resolved value in.

apps/desktop/src/hooks/workspaces/cache/use-workspace-selection-cache.ts
```diff
@@ -61,7 +65,12 @@ export function useWorkspaceSelectionCache() {
       if (!workspaceId || nextIds.has(workspaceId)) {
         continue;
       }
-      for (const root of anyHarnessWorkspaceQueryKeyRoots(input.runtimeUrl, workspaceId)) {
+      const workspaceRoots = [
+        ...anyHarnessWorkspaceQueryKeyRoots(cacheScopeKey, workspaceId),
```
pablonyx, line 69: Minor race: this cancels using the *current* render's `cacheScopeKey`, but an in-flight query issued just before an auth transition (bootstrapping→authenticated, logout) was keyed under the old scope, so it won't match and can resolve into the stale scope's cache. Probably acceptable since the old scope's entries are orphaned anyway — flagging in case orphaned entries are ever rehydrated.

PR #1157, opened 2026-07-14
feat(desktop): adopt ProductHostProvider with real Desktop host and bridge

pablonyx:
Implements the frozen **D1a: Desktop Host Adoption** contract (`specs/codebase/features/web-desktop-client-unification-d1a.md`, revision D1a-r1), plus accepted first/second-review fixes.

- Base: `2ec15eaf8cfc870cbdbb42c225a5f1428e5282b4` (exact frozen base)
- Head: see latest commit (review fixes applied after `18a368288`; blocker round at `8b0ac73a3`; CI-001 fix follows)

## What this does
- `DesktopProductHostProvider` (mounted per §6 inside `CloudClientProvider`, above `WorkspaceProviders`) constructs one immutable `ProductHost` from existing Desktop state. Snapshot replacement follows the frozen table exactly: auth status, authenticated user identity fields (gated to authenticated status per HOST-001), normalized anonymous method list (gated to anonymous), or `cloudClient` reference — nothing else. Static adapters and callbacks keep stable identities across replacements.
- `lib/access/tauri/desktop-bridge.ts` exports the complete stable `DesktopBridge` as thin adapters per the §9 table (runtime URL mapping, race-safe sync menu-command unsubscribe, updater error→rejection + bounded 0..1 progress, worker stop-result discard, SSH `localUrl→runtimeUrl`, diagnostics arg mapping).
- Deep links: `deep-link.ts` multiplexes the raw Tauri source (one native live listener, per-subscription `getCurrent()` snapshot, race-safe sync unsubscribe, no queue/persistence/retry). `ensureDeepLinkBridge` keeps its original once-only memoization (LINK-001) and its sequential awaited initial drain with contained handler rejections (Greptile P1).
- `desktop-navigation.ts` gains pure `ProductEntry` encode/decode for the host path; `desktopNavigationTarget` builds routes from the raw query string, byte-identical to the base implementation including duplicate keys and exotic encodings (NAV-001).
- One auth-bootstrap owner: `AppRuntime` consumes `host.auth.restoreSession`; startup timing/diagnostics stay in `AppRuntime`. §7.2 login-request disposition implemented exactly.
- Running-agent count export moved into `DesktopProductLifecycleRoot` gated by `host.desktop` (only lifecycle moved).
- `ProductStorage` over guarded `window.localStorage`; no key migration.
- Narrow accepted expansions beyond the original §5 plan: `updater.ts` adapted to the real plugin 2.10.1 `DownloadEvent` union so update progress actually reports (UPDATER-001, reviewer-recommended), and `tests/intent/stack/boot.ts` adds ProductClient to the clean intent-stack frontend build (CI-001).
- Only ProductClient edit: `observeInboundEntries` doc clarification (bounded no-queue meaning); no TS shape change.

## Evidence (at review-fix head)
- `pnpm --filter @proliferate/product-client build` ✅
- Focused Vitest run over the 9 touched test files: **153/153 pass** ✅; tsc `--noEmit` clean ✅
- `pnpm --dir apps/desktop test` (exact frozen command) ❌ exits 1 in the pretest design-system check on pre-existing violations in unchanged files, reproduced identically at the frozen base (VERIFY-001 — explicit waiver requested; the Vitest failures in that run are the same 24 pre-existing ones as base)
- `pnpm --dir apps/desktop build` ✅ (tsc clean, vite build ok)
- `python3 scripts/report_frontend_structure.py --strict --summary-only` ✅ (TOTAL: 0)
- `git diff --check` ✅
- Named-profile Desktop smoke (`make run PROFILE=d1a-host`, prebuilt runtime per feature-worktree docs): rerun on the final head (VERIFY-002) — results recorded in PR comments.

## Assumptions / disclosed choices
- `settings/environments` (emitted by literal return URLs, absent from the current route table) decodes/encodes through `ProductEntry` for round-trips, but `desktopNavigationTarget` still returns `null` for it — no new deep-link destination.
- `settings/cloud|billing` decode to a `settings`/billing entry (query preserved) rather than `billing-return`; `billing-return` comes only from `billing/success|cancel`; status `"done"`, `workflow`, `invitation`, and settings-`general` reject from the encoder.
- `source=github_app_callback` is lifted into the typed `source` field; `github_app_installation_callback` stays in `query` (the ProductClient type was not widened).
- Anonymous method availability mirrors the login surfaces: `cloudEnabled && <probe>` for password/GitHub/SSO.
- The slug-SSO generic unavailable message is duplicated from `use-org-slug-sso-sign-in.ts` (not exported there; exporting would exceed the frozen file plan).

### DIFF
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -5,11 +5,11 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
-    "dev:built": "pnpm --filter @anyharness/sdk build && pnpm --filter @anyharness/sdk-react build && pnpm --filter @proliferate/cloud-sdk build && pnpm --filter @proliferate/cloud-sdk-react build && pnpm --filter @proliferate/design build && pnpm --filter @proliferate/product-domain build && pnpm --filter @proliferate/ui build && pnpm --filter @proliferate/product-ui build && pnpm --filter @proliferate/product-surfaces build && vite",
-    "build": "pnpm --filter @anyharness/sdk build && pnpm --filter @anyharness/sdk-react build && pnpm --filter @proliferate/cloud-sdk build && pnpm --filter @proliferate/cloud-sdk-react build && pnpm --filter @proliferate/design build && pnpm --filter @proliferate/product-domain build && pnpm --filter @proliferate/ui build && pnpm --filter @proliferate/product-ui build && pnpm --filter @proliferate/product-surfaces build && tsc && vite build",
+    "dev:built": "pnpm --filter @anyharness/sdk build && pnpm --filter @anyharness/sdk-react build && pnpm --filter @proliferate/cloud-sdk build && pnpm --filter @proliferate/cloud-sdk-react build && pnpm --filter @proliferate/design build && pnpm --filter @proliferate/product-domain build && pnpm --filter @proliferate/product-client build && pnpm --filter @proliferate/ui build && pnpm --filter @proliferate/product-ui build && pnpm --filter @proliferate/product-surfaces build && vite",
+    "build": "pnpm --filter @anyharness/sdk build && pnpm --filter @anyharness/sdk-react build && pnpm --filter @proliferate/cloud-sdk build && pnpm --filter @proliferate/cloud-sdk-react build && pnpm --filter @proliferate/design build && pnpm --filter @proliferate/product-domain build && pnpm --filter @proliferate/product-client build && pnpm --filter @proliferate/ui build && pnpm --filter @proliferate/product-ui build && pnpm --filter @proliferate/product-surfaces build && tsc && vite build",
     "preview": "vite preview",
     "tauri": "tauri",
-    "pretest": "pnpm --filter @proliferate/design build && pnpm --filter @proliferate/product-domain build && pnpm --filter @proliferate/ui build && pnpm --filter @proliferate/product-ui build && pnpm --filter @proliferate/product-surfaces build && bash scripts/check-design-system.sh",
+    "pretest": "pnpm --filter @proliferate/design build && pnpm --filter @proliferate/product-domain build && pnpm --filter @proliferate/product-client build && pnpm --filter @proliferate/ui build && pnpm --filter @proliferate/product-ui build && pnpm --filter @proliferate/product-surfaces build && bash scripts/check-design-system.sh",
     "test": "vitest run",
     "check:design": "bash scripts/check-design-system.sh",
     "test:watch": "vitest"
@@ -31,6 +31,7 @@
     "@proliferate/cloud-sdk": "workspace:*",
     "@proliferate/cloud-sdk-react": "workspace:*",
     "@proliferate/design": "workspace:*",
+    "@proliferate/product-client": "workspace:*",
     "@proliferate/product-domain": "workspace:*",
     "@proliferate/product-surfaces": "workspace:*",
     "@proliferate/product-ui": "workspace:*",
--- a/apps/desktop/src/lib/access/tauri/deep-link.ts
+++ b/apps/desktop/src/lib/access/tauri/deep-link.ts
@@ -2,38 +2,113 @@ import { getCurrent, onOpenUrl } from "@tauri-apps/plugin-deep-link"
 
 export type DeepLinkUrlHandler = (url: string) => Promise<boolean>
 
-let deepLinkBridgePromise: Promise<void> | null = null
+type DeepLinkListener = (url: string) => void
 
-/**
- * Ensures the Tauri deep-link listener is registered exactly once.
- * Drains any URLs that arrived before the listener was set up, then
- * forwards every subsequent deep-link URL to `handler`.
- *
- * Safe to call outside Tauri — errors are silently swallowed.
- */
-export function ensureDeepLinkBridge(handler: DeepLinkUrlHandler): Promise<void> {
-  if (deepLinkBridgePromise) {
-    return deepLinkBridgePromise
+interface DeepLinkSubscription {
+  listener: DeepLinkListener
+  active: boolean
+}
+
+// Every active subscription, in subscribe order. Delivery always reads this
+// set fresh, so removing a subscription (or flipping `active` off) takes
+// effect immediately for both the in-flight initial snapshot and any live
+// URL still queued on the microtask/event loop.
+const subscriptions = new Set<DeepLinkSubscription>()
+
+// At most one native live listener is ever registered; this memoizes the
+// in-flight/settled registration so concurrent subscribers share it.
+let liveListenerPromise: Promise<void> | null = null
+
+function ensureLiveListener(): Promise<void> {
+  if (!liveListenerPromise) {
+    liveListenerPromise = Promise.resolve()
+      .then(() =>
+        onOpenUrl((urls) => {
+          for (const url of urls) {
+            for (const subscription of subscriptions) {
+              if (subscription.active) {
+                subscription.listener(url)
+              }
+            }
+          }
+        }),
+      )
+      .then(() => undefined)
   }
+  return liveListenerPromise
+}
 
-  deepLinkBridgePromise = (async () => {
+function deliverInitialSnapshot(subscription: DeepLinkSubscription): Promise<void> {
+  return (async () => {
     try {
       const currentUrls = await getCurrent()
-      if (currentUrls?.length) {
+      if (subscription.active && currentUrls?.length) {
         for (const url of currentUrls) {
-          await handler(url)
+          subscription.listener(url)
         }
       }
-
-      await onOpenUrl((urls) => {
-        for (const url of urls) {
-          void handler(url)
-        }
-      })
     } catch {
       // Ignore when running outside Tauri or before the plugin is available.
     }
   })()
+}
 
-  return deepLinkBridgePromise
+/**
+ * Subscribes to the raw Tauri deep-link source. Delivers Tauri's current
+ * `getCurrent()` snapshot (if any) to this subscriber only, then forwards
+ * every URL that arrives while the subscription is active. At most one
+ * native live listener is ever registered — it is created lazily on the
+ * first subscription and shared by every subscriber.
+ *
+ * There is no history or queue: a URL delivered before a later subscriber
+ * mounts is never replayed to it, and no raw URL is retained after delivery.
+ *
+ * Returns a synchronous unsubscribe function. Unsubscribing is race-safe
+ * even while native registration (or this subscriber's initial snapshot
+ * drain) is still in flight — no further delivery reaches this listener
+ * after unsubscribe, and other subscribers are unaffected.
+ *
+ * Safe to call outside Tauri — errors are silently swallowed.
+ */
+export function subscribeDeepLinkUrls(listener: DeepLinkListener): () => void {
+  const subscription: DeepLinkSubscription = { listener, active: true }
+  subscriptions.add(subscription)
+
+  void deliverInitialSnapshot(subscription)
+  void ensureLiveListener().catch(() => {
+    // Ignore when running outside Tauri or before the plugin is available.
+  })
+
+  return () => {
+    subscription.active = false
+    subscriptions.delete(subscription)
+  }
+}
+
+/**
+ * Legacy adapter over the multiplexed raw source above. Drains any URLs
+ * that arrived before subscription, then forwards every subsequent
+ * deep-link URL to `handler`. Unlike `subscribeDeepLinkUrls`, this handler
+ * is never unsubscribed — it lives for the process, matching the existing
+ * bootstrap-time callback consumer.
+ *
+ * Safe to call outside Tauri — errors are silently swallowed. The returned
+ * promise resolves once the initial drain and the live-listener
+ * registration attempt have both settled.
+ */
+export function ensureDeepLinkBridge(handler: DeepLinkUrlHandler): Promise<void> {
+  const subscription: DeepLinkSubscription = {
+    listener: (url) => {
+      void handler(url)
+    },
+    active: true,
+  }
+  subscriptions.add(subscription)
+
+  return Promise.all([
+    deliverInitialSnapshot(subscription),
+    ensureLiveListener().catch(() => {
+      // Ignore when running outside Tauri or before the plugin is available.
+    }),
+  ]).then(() => undefined)
 }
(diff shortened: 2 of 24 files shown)


apps/desktop/src/lib/access/tauri/deep-link.ts
```diff
@@ -2,38 +2,113 @@ import { getCurrent, onOpenUrl } from "@tauri-apps/plugin-deep-link"
...
+ *
+ * Safe to call outside Tauri — errors are silently swallowed. The returned
+ * promise resolves once the initial drain and the live-listener
+ * registration attempt have both settled.
+ */
+export function ensureDeepLinkBridge(handler: DeepLinkUrlHandler): Promise<void> {
+  const subscription: DeepLinkSubscription = {
+    listener: (url) => {
+      void handler(url)
+    },
```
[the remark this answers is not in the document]
pablonyx: Fixed in 30f3b6cad — and it was slightly worse than flagged: the original bridge also *awaited* each initial-snapshot URL sequentially inside its try, which my adapter had dropped. Restored: sequential awaited drain with contained rejections, live subscription registered only after the drain, and live handler rejections contained. Added a regression test asserting no unhandledRejection for either path.

apps/desktop/src/providers/DesktopProductHostProvider.tsx
```diff
@@ -0,0 +1,163 @@
...
+          github_login: githubLogin,
+          avatar_url: avatarUrl,
+        }),
+      };
+    }
+    return { status: "anonymous", methods };
+    // methods is folded into anonymousMethodsKey so authenticated method
+    // changes do not recompute the snapshot.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [status, userId, displayName, email, avatarUrl, githubLogin, anonymousMethodsKey]);
```
pablonyx: [Blocker HOST-001] Gate identity-driven host replacement to authenticated state. The memo dependencies always include userId/displayName/email/avatarUrl/githubLogin. If retained user data changes while the status is anonymous or bootstrapping, this builds a new auth object and ProductHost even though none of those fields is represented in that AuthState. The frozen contract says authenticated identity fields are replacement triggers only while authenticated. Fold the identity dependency into an authenticated-only key or snapshot and add a regression test for identity mutation while anonymous/loading.

apps/desktop/src/lib/access/tauri/deep-link.ts
```diff
@@ -2,38 +2,128 @@ import { getCurrent, onOpenUrl } from "@tauri-apps/plugin-deep-link"
...
+      }
+    }
+  } catch {
+    // Ignore when running outside Tauri or before the plugin is available.
+  }
+
+  // Live URLs begin flowing only after the drain, matching the original
+  // registration order. Handler results are discarded; rejections are
+  // contained so a failing callback cannot surface as an unhandled rejection.
+  subscriptions.add({
```
pablonyx: [Blocker LINK-001] Preserve the legacy bridge as a once-only registration. The base implementation memoized the entire ensureDeepLinkBridge call; this implementation only memoizes the native listener and unconditionally adds a permanent JavaScript subscription on every call. AppRuntime invokes restoreSession in an effect under React.StrictMode, so normal auth bootstrap can register the legacy handler twice and one live OAuth URL can start two callback exchanges. Make this adapter idempotent and test two ensure calls followed by one live URL.

apps/desktop/src/lib/domain/auth/desktop-navigation.ts
```diff
@@ -11,90 +26,299 @@ export function desktopNavigationTarget(url: string): string | null {
...
+ * strings to the historical route table for every URL that table supported.
+ * URLs the table never routed (including the newly decodable `environments`
+ * return URL) stay `null`.
+ */
+export function desktopNavigationTarget(url: string): string | null {
+  const entry = decodeDesktopProductEntry(url);
+  if (entry === null) {
+    return null;
+  }
+  return entryToRoute(entry);
```
pablonyx: [Blocker NAV-001] This cannot satisfy the required byte-identical legacy route behavior because the raw query is decoded into Record<string, string> and serialized again. Examples: x=a%20b becomes x=a+b; x=1&x=2 collapses to x=2; an integration callback with an unrecognized status or extra parameter drops those values. The base adapter preserved the raw query for all recognized routes. Keep normalized ProductEntry for the new host path, but preserve raw legacy query semantics in desktopNavigationTarget and add these compatibility cases.

apps/desktop/src/lib/access/tauri/desktop-bridge.ts
```diff
@@ -0,0 +1,230 @@
...
+    },
+    async downloadAndInstall(
+      update: DesktopUpdate,
+      onProgress?: (fraction: number) => void,
+    ): Promise<void> {
+      let received = 0;
+      await downloadAndInstallUpdate(
+        update.handle,
+        onProgress
+          ? (chunkLength, contentLength) => {
```
pablonyx, line 170: [Decision UPDATER-001] The progress test mocks an interface the real updater does not emit. Plugin 2.10.1 sends Started, Progress, and Finished events with contentLength or chunkLength under event.data, while updater.ts reads top-level progress.chunk and progress.contentLength. In a real update this bridge receives undefined values and never emits a fraction. The frozen contract requires real bounded progress, but updater.ts is outside the frozen file plan. My recommendation is to narrowly broaden this slice to fix that wrapper and test the actual event union; otherwise explicitly defer or waive the progress guarantee.

pablonyx:
### Independent review — Desktop Host Adoption

**Verdict: changes required** at head `30f3b6cad4e83b4d384e2d708a301d6c8c858cfc`.

Inline blockers:

- **HOST-001** — identity fields replace the host outside authenticated state.
- **LINK-001** — repeated bootstrap calls install duplicate permanent legacy deep-link handlers.
- **NAV-001** — legacy route normalization is lossy and violates the byte-identical compatibility contract.

Founder decisions:

- **UPDATER-001** — the new bridge test passes against a mocked tuple, but the installed updater emits a different event union; real progress is currently never reported. I recommend narrowly adding `updater.ts` to this slice and testing the real event shape.
- **VERIFY-001** — the exact frozen command `pnpm --dir apps/desktop test` exits 1 in the pretest design-system check before Vitest. The violations are in unchanged files and are base-equivalent; the 8 focused files pass all 141 tests. Because the frozen contract requires this exact command to succeed, either explicitly reconcile/waive that baseline failure for this slice or make the baseline command green before merge. The implementation report must not describe the exact command as passing.

Metadata:

- **META-001** — add the required `area:product` label. The metadata check is failing because this PR changes ProductClient.

The latest follow-up commit correctly restores awaited initial deep-link handling and contains handler rejections; that earlier issue is resolved. I found no additional blockers in storage, lifecycle gating, deployment/auth adapter delegation, Cloud client reuse, or the remaining DesktopBridge groups.

apps/desktop/src/providers/DesktopProductHostProvider.tsx
pablonyx: [HOST-001] Fixed in 8b0ac73a389878286bb29c75e776c7f51281284d. The five identity fields are gated to authenticated status (same pattern as anonymousMethodsKey), so identity mutation while anonymous/bootstrapping no longer replaces the host. Regression tests added for anonymous, bootstrapping, and authenticated cases.

apps/desktop/src/lib/access/tauri/deep-link.ts
pablonyx: [LINK-001] Fixed in 8b0ac73a389878286bb29c75e776c7f51281284d. ensureDeepLinkBridge is memoized again via a module-level promise — first handler wins, later calls return the same promise and register nothing, so a StrictMode double bootstrap cannot start two callback exchanges. New test: two ensure calls + one live URL → exactly one handler invocation, getCurrent drained once.

apps/desktop/src/lib/domain/auth/desktop-navigation.ts
pablonyx: [NAV-001] Fixed in 8b0ac73a389878286bb29c75e776c7f51281284d. desktopNavigationTarget no longer routes through the normalized entry: it rebuilds routes from the raw parsed search exactly like the base (URLSearchParams(parsed.search)+set for settings routes, raw search appended for workspaces), verified against git show 2ec15eaf8 as the oracle. Duplicate keys, %20, and unrecognized/extra integration-callback params all pass through byte-identically (new compat tests). The ProductEntry codec for the host path is unchanged.

apps/desktop/src/lib/access/tauri/desktop-bridge.ts
pablonyx: [UPDATER-001] Accepted the recommendation and narrowly broadened the slice in 8b0ac73a389878286bb29c75e776c7f51281284d: updater.ts downloadAndInstall now consumes the real plugin 2.10.1 DownloadEvent union (Started captures contentLength, Progress emits chunkLength, Finished ignored) while keeping its exported (chunkLength, contentLength) signature. Verified the union against the installed dist-js types. New updater.test.ts drives the real event shape end-to-end; the bridge fraction test proves Started(100)+Progress(40)+Progress(60) → 0.4 then 1.0, clamped, and no fraction when contentLength is unknown.

pablonyx:
Review fixes applied at head `8b0ac73a3`: HOST-001, LINK-001, NAV-001 fixed; UPDATER-001 accepted and fixed by narrowly adding `updater.ts` to the slice (real DownloadEvent union, tested). All 148 D1a tests green, tsc clean, structure check 0, git diff --check clean.

[VERIFY-001] Founder decision stands as disclosed: the exact `pnpm --dir apps/desktop test` command exits 1 in the pretest design-system check on pre-existing violations in unchanged files, reproduced identically on the clean base SHA. Making that command green requires editing files outside the frozen §5 file plan, so I'm not absorbing it into this slice — request an explicit waiver/reconciliation of that baseline for D1a, or a separate cleanup PR to green the baseline first.

apps/desktop/package.json
```diff
@@ -31,6 +31,7 @@
     "@proliferate/cloud-sdk": "workspace:*",
     "@proliferate/cloud-sdk-react": "workspace:*",
     "@proliferate/design": "workspace:*",
+    "@proliferate/product-client": "workspace:*",
```
pablonyx, line 34: [Blocker CI-001] Add ProductClient to the clean intent-stack frontend build. This dependency is built by the Desktop scripts, but tests/intent/stack/boot.ts ensureFrontendBuilt still omits it. The Tier-2 workflow lifecycle job therefore boots Vite without ProductClient dist and fails to resolve ProductHostProvider from App.tsx; the exact base job passed. Add the package to that build list and rerun the failing Tier-2 lane.

pablonyx:
### Second independent review — Desktop Host Adoption

Reviewed exact head `8b0ac73a389878286bb29c75e776c7f51281284d`.

The implementation repairs are correct: **HOST-001**, **LINK-001**, **NAV-001**, and **UPDATER-001** are resolved. The nine focused files pass all 153 tests; ProductClient build, Desktop build, strict structure enforcement, and diff-check pass.

**Verdict: not merge-ready yet.**

- **CI-001 — blocker:** the Tier-2 workflow lifecycle lane is newly broken. Clean intent-stack boot does not build ProductClient, so Vite cannot resolve `ProductHostProvider`. Add ProductClient to `tests/intent/stack/boot.ts::ensureFrontendBuilt()` and rerun that lane. See the inline comment.
- **META-001:** add `area:product`; metadata CI still fails solely for that missing label.
- **BODY-001:** refresh the PR body. It still names head `18a368…`, says 140 tests, says the legacy route is decoder-derived, and says `ensureDeepLinkBridge` is not memoized. All are stale after the review fixes.
- **VERIFY-002:** rerun or explicitly refresh the named-profile smoke evidence on the final head; the body currently carries smoke evidence from before the auth/deep-link/navigation/updater fixes.

Two accepted implementation realities must be reconciled into the authority record before the next implementation handoff:

- **SPEC-R2-001:** record the raw-query compatibility path, the narrow updater access-file expansion, and the contained deep-link rejection behavior. I recommend accepting the latter as a small correctness hardening.
- **SPEC-PROMOTION-001:** the referenced frozen Desktop Host Adoption spec is currently untracked locally and absent from both Git and this PR. Promote the reconciled spec/current-state ledger before relying on it for the next slice.

**VERIFY-001** is conclusively base-equivalent: the exact Desktop test command fails before Vitest on the identical pre-existing design-system violations at the exact frozen base. I recommend an explicit scoped waiver rather than absorbing unrelated cleanup.

apps/desktop/package.json
pablonyx: [CI-001] Fixed in 90926523c3662067e02f8511db6c8e0058e119f1 — @proliferate/product-client added to ensureFrontendBuilt() (after product-domain, matching the Desktop script chains). Reran the lane locally with the CI env (USE_EXISTING_POSTGRES/REDIS, TIER2_INTENT_SKIP_RUNTIME=1) on its own profile: 2/2 specs pass in 30.7s.

pablonyx:
Second-review items resolved at head `90926523c`:

- **CI-001** fixed: ProductClient added to `tests/intent/stack/boot.ts::ensureFrontendBuilt()`; workflow-definition lifecycle lane rerun locally under the CI env — 2/2 pass.
- **META-001** done: `area:product` label added.
- **BODY-001** done: PR body refreshed (head/test counts/raw-query routing/bridge memoization all current).
- **VERIFY-002** done: named-profile smoke (`d1a-host`) rerun on the review-fix head — app boots under the host provider with no provider error, auth bootstrap reaches the same authenticated state, home + settings surfaces load, 34 runtime models load through the same AnyHarness runtime across the auth snapshot replacement, zero console errors.

Remaining for the authority record (architect/founder, not implementable here): **SPEC-R2-001** (record raw-query compat path, updater file expansion, contained deep-link rejections), **SPEC-PROMOTION-001** (promote the untracked frozen spec into Git), and the **VERIFY-001** scoped waiver.

pablonyx:
### Final code re-review — Desktop Host Adoption

Reviewed exact head `90926523c3662067e02f8511db6c8e0058e119f1`.

**No remaining code findings.** CI-001 is correctly fixed, and the previously failing Tier-2 workflow lifecycle lane now passes in GitHub Actions. Metadata passes with both required area labels; the PR body and final-head smoke evidence are refreshed; all 153 focused tests pass; Desktop/ProductClient builds, structure enforcement, intent lanes, and diff-check are green.

The code review is clear. Do not merge or start the dependent implementation until:

1. the still-running general test check settles green;
2. the founder explicitly accepts the scoped VERIFY-001 base-equivalent waiver; and
3. the reconciled Desktop Host Adoption authority spec is promoted into Git, recording the accepted raw-route compatibility path, updater/intent-build expansions, and contained deep-link rejection behavior.

After those gates, merge this PR, record its merge SHA, and reconcile/freeze Desktop Native UI Adoption before implementation.

PR #1158, opened 2026-07-14
feat(anyharness): execute one-prompt workflow runs in an existing workspace (C2a)

pablonyx:
Implements the frozen specification `specs/codebase/features/workflow-runs.md` (revision **C2a.3**, SHA-256 `87e6b610…ffdc`, committed in this PR) — the smallest real AnyHarness-only workflow execution vertical.

- **Base SHA**: `2ec15eaf8cfc870cbdbb42c225a5f1428e5282b4` (exact spec-mandated base)
- **Head SHA**: `a216e003d198aba154c04a539e20f8c9ab8cccde`

## Behavior

`PUT /v1/workflow-runs/{runId}` accepts one strict schema-version-1 invocation (one stage, one `agent.prompt` step, scalar inputs/arguments, `{{inputs.name}}` templates) against an **existing** workspace: it atomically commits a run + pending step in one SQLite transaction (canonical `invocation_json` is the sole replay authority — same ID + equal invocation → `200` exact replay with zero effects; different → `409 WORKFLOW_RUN_CONFLICT`), then for `Created` only spawns one execution task that creates a new normal InternalOnly/subagents-disabled session in the supplied workspace under a shared `SessionStart` lease, persists `session_id` before startup, dispatches the resolved prompt with deterministic identity `workflow:<runId>:0:0`, and observes terminal completion through a `SessionExtension` matching exact session+prompt identity. `GET` serves the durable run/step view (correlation IDs survive artifact deletion). Failures persist stable codes only; nonterminal state is fenced to `failed(runtime_restarted)` at AppState construction, and a fencing failure aborts startup.

## Structure (spec §7)

- `anyharness-contract/src/v1/workflow_runs.rs` — strict wire types (`deny_unknown_fields` at every level; `modelId`/`modeId` required-but-nullable in both the decoder and the OpenAPI schema)
- `anyharness-lib/src/domains/workflows/` — `model.rs`, `store/{mod,runs,steps}.rs` (`WorkflowRunStore`, sync SQL, two-tier, intent ops `accept`/`begin_run`/`bind_session`/`begin_step`/`record_turn`/`finish_turn`/`fail_nonterminal`/`fence_nonterminal_after_restart`), `service.rs` (`WorkflowRunService`: validation, rendering, canonical JSON, guarded CAS transitions, fencing — sync only), `runtime.rs` (`WorkflowRunRuntime`: sole async facade, spawn_blocking boundary, one outer Result, one guarded failure write), `session_extension.rs` (`WorkflowRunSessionExtension`: nonblocking, main-runtime blocking pool)
- `api/http/workflow_runs{,_contract,_errors}.rs`, router + OpenAPI registration (routes NOT added to the direct-attach JWT allowlist; `api/auth.rs` untouched)
- `app/workflows.rs` — construction-only two-phase wiring; `AppStateInitError::WorkflowFencingFailed` aborts init
- `persistence/sql/0060_workflow_runs.sql` + regenerated schema snapshot; regenerated SDK OpenAPI/TS artifacts
- Narrow generic session seams only (spec §5.3): `create_persisted_internal_session`, `send_text_prompt_with_id`, and `prompt_id` threaded `SessionTurnFinishResult → SessionTurnFinishedContext`

## Verification

- `cargo test -p anyharness-lib`: **1038 passed, 0 failed, 1 ignored** (**51 workflow-scoped**, including validation, exact-replay/concurrency, CAS/terminal immutability, accepted-to-execution cancellation safety, runtime/session-extension crossings, auth admission, completion races, and file-backed restart fencing)
- `cargo test -p anyharness-contract`: **36 passed** (8 workflow-contract tests; strict scalar/failure unions and required-nullable model/mode pinned in serde, OpenAPI, and generated SDK)
- `check_anyharness_boundaries.py` / `check_anyharness_old_paths.py` / `check_max_lines.py`: pass (no rules weakened; line-count allowlist bumps for openapi/router registrations + the new test file)
- Schema snapshot + `cargo build --bin anyharness` + SDK `pnpm generate && pnpm build`: pass
- Independent review round 1 found six blockers plus one workspace-unavailability decision; all were fixed in `a216e003d`, regression-pinned, and independently re-reviewed clean across spec fidelity, SQLite/runtime races, and HTTP/OpenAPI/SDK contracts.

## Real Tier 3 acceptance journey (isolated runtime, real Claude agent)

Booted the worktree binary with a fresh runtime home and one pre-existing workspace:
1. `PUT` → `201` (run `accepted`, step `pending`, `promptId workflow:9f3d2c1a…:0:0`)
2. `GET` polled → run+step `completed`, `sessionId 00b0bcef…`, `turnId fb84bc7e…`
3. Session events correlate exactly: `user_message` carries that promptId with the argument-resolved prompt (`…: WORKFLOW-C2A3-OK`); `assistant_message` completed with exactly `WORKFLOW-C2A3-OK` (provider `msg_011Cd19…`) on the same turn; raw SQLite shows the matching 1 run + 1 step rows
4. Identical `PUT` (shuffled key order/whitespace) → `200`, then still exactly **1 workspace, 1 session, 1 turn, 1 user item, 1 assistant item**; workspace git state untouched
5. Changed argument → `409 WORKFLOW_RUN_CONFLICT`, rows unchanged
6. Nonexistent workspace → `201` accepted, then durable `failed`/`failed` with `workspace_unavailable` and `finishedAt` set

## Assumptions (reversible, disclosed)

1. Input names / placeholder identifiers: `[A-Za-z_][A-Za-z0-9_]*`; any `{{…}}` not exactly `{{inputs.<declared>}}` (incl. unmatched `{{`) is an invalid template.
2. `runId` canonical form = lowercase hyphenated UUID (uppercase/braced/simple/urn rejected); `workspaceId` validated as nonblank (existence is post-acceptance).
3. Missing, retired, frozen-for-handoff, and other typed workspace access-gate refusals map to `workspace_unavailable`; non-workspace failures at session creation map to `session_create_failed`.
4. A lost CAS guard (`begin_run`/`bind_session`/`begin_step` returning false — row concurrently fenced) ends the execution task without a second write.
5. `record_turn` store failure logs and continues (extension owns completion; never claim/deny completion from the dispatch side).
6. Decode-400 detail is generic (no serde message echo, so argument values can't leak); one machine code `WORKFLOW_RUN_INVALID` covers shape + semantic 400s (spec table mandates only the status).
7. GET also asserts workspace auth scope against the stored run's workspace (defense-in-depth; ordinary bearer behavior unchanged).

## Non-goals honored

No workspace creation/worktrees, no takeover, no multi-stage/step, no goals/cancellation, no Cloud/Desktop delivery, no grants/schedules/retries/recovery, no workflow actor/manager/scheduler/executor/step-registry/plugin framework. No Postgres run table.

No unresolved blockers.


anyharness/crates/anyharness-lib/src/domains/workflows/runtime.rs
```diff
@@ -0,0 +1,372 @@
...
+    /// current view without any effect; conflict is a typed error.
+    #[tracing::instrument(skip_all, fields(run_id = %run_id))]
+    pub async fn put(
+        &self,
+        run_id: String,
+        input: crate::domains::workflows::model::PutWorkflowRunInput,
+    ) -> Result<WorkflowPutSuccess, WorkflowPutError> {
+        let service = self.service.clone();
+        let accept_run_id = run_id.clone();
+        let outcome = tokio::task::spawn_blocking(move || service.accept(&accept_run_id, input))
```
pablonyx: **C2A-REV-01 (Blocker): the durable acceptance → execution handoff is cancellation-unsafe.** `spawn_blocking` keeps running if the HTTP/PUT future is dropped. It can commit `Created`, while the cancelled outer future never reaches `spawn_execution(plan)` below. An identical replay intentionally starts no effects, so that run remains `accepted` until a process restart fences it. Please put acceptance, the `Created` decision, and execution scheduling inside one owned task on the captured main runtime, then await that task for the response; dropping the waiter must detach, not cancel, the handoff. Add a regression that blocks acceptance, aborts the caller, releases it, and proves the created run advances/terminalizes rather than becoming orphaned.

anyharness/crates/anyharness-lib/src/api/http/workflow_runs.rs
```diff
@@ -0,0 +1,75 @@
...
+        (status = 404, description = "Unknown workflow run", body = anyharness_contract::v1::ProblemDetails),
+    ),
+    tag = "workflow-runs"
+)]
+pub async fn get_workflow_run(
+    State(state): State<AppState>,
+    Extension(auth): Extension<AuthContext>,
+    Path(run_id): Path<String>,
+) -> Result<Json<WorkflowRunResponse>, ApiError> {
+    let view = state
```
pablonyx, line 69: **C2A-REV-02 (Blocker): GET bypasses the canonical `runId` contract.** `WorkflowRunService::get` looks up this raw string, so `GET .../not-a-uuid` returns `404 WORKFLOW_RUN_NOT_FOUND`; frozen §3 requires invalid IDs to return the coded 400, matching PUT. Please reuse the domain UUID validator in the GET path, add a typed invalid-GET error mapping, and pin the HTTP behavior.

anyharness/crates/anyharness-contract/src/v1/workflow_runs.rs
```diff
@@ -0,0 +1,343 @@
...
+
+/// `PUT /v1/workflow-runs/{runId}` request body. The `runId` lives in the path
+/// only and never appears in the body.
+#[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct PutWorkflowRunRequest {
+    pub schema_version: i64,
+    pub workspace_id: String,
+    pub definition: WorkflowRunDefinition,
+    /// Concrete scalar argument values keyed by declared input name. Values are
```
pablonyx: **C2A-REV-03 (Blocker): the generated public contract is looser than schema v1.** This becomes `[key: string]: unknown` in the SDK even though arguments may only be string/number/boolean; the response likewise exposes `failureCode?: string` instead of the frozen seven-code vocabulary. Clients can therefore construct values the server rejects and cannot exhaustively handle results. Please add contract-owned bounded scalar and failure-code types (or exact schema overrides), keep mapping at the API seam, and regenerate OpenAPI/SDK. The existing generic coded 400 can remain.

anyharness/crates/anyharness-lib/src/api/http/workflow_runs.rs
```diff
@@ -0,0 +1,75 @@
...
+use crate::api::auth::AuthContext;
+use crate::app::AppState;
+use crate::domains::workflows::runtime::WorkflowPutSuccess;
+
+#[utoipa::path(
+    put,
+    path = "/v1/workflow-runs/{run_id}",
+    params(("run_id" = String, Path, description = "Canonical UUID for the workflow run")),
+    request_body = anyharness_contract::v1::PutWorkflowRunRequest,
+    responses(
```
pablonyx, line 25: **C2A-REV-04 (Blocker): the generated response matrix contradicts the frozen/runtime behavior.** PUT has no synchronous 404 path—missing workspace is accepted and later durably fails—but acceptance storage failure does produce the frozen 500 and is undocumented. Please remove PUT 404, add PUT 500, and add GET 400 with the canonical-ID fix, then regenerate the SDK.

anyharness/crates/anyharness-lib/src/api/workflow_runs_tests.rs
```diff
@@ -0,0 +1,203 @@
+//! HTTP-level tests for the workflow-runs routes over an in-memory `AppState`
```
pablonyx, line 1: **C2A-REV-05 (Blocker): no merge-gated test crosses `WorkflowRunRuntime` or `WorkflowRunSessionExtension`.** The HTTP test intentionally uses a nonexistent workspace and never polls even that asynchronous failure; the remaining tests call only the synchronous service/store. Frozen §8.3 requires controlled AnyHarness-fixture proof of Created-only effects/replay zero effects, supplied-workspace reuse, session checkpoint-before-start, prompt/completion handoff, post-acceptance failure classification, and startup-fence failure. Please add focused controlled-runtime tests at the existing concrete session seams—no generalized executor port. This should also house the cancellation regression above.

anyharness/crates/anyharness-lib/src/api/workflow_runs_tests.rs
```diff
@@ -0,0 +1,203 @@
...
+        .expect("body");
+    let payload = if bytes.is_empty() {
+        Value::Null
+    } else {
+        serde_json::from_slice(&bytes).unwrap_or(Value::Null)
+    };
+    (status, payload)
+}
+
+#[tokio::test]
```
pablonyx, line 107: **C2A-REV-06 (Blocker): the explicit bearer/direct-attach boundary is correct in code but untested for these routes.** Every workflow HTTP test disables bearer auth, so §8.1 is not merge-gated. Please cover missing/invalid and valid worker bearer behavior, and pin that direct-attach route admission rejects both workflow PUT and GET.

pablonyx (commented):
Independent review against frozen C2a.3 at exact head `29dbe9ed9`: **changes requested**. I found six merge-blocking correctness/contract/verification gaps in the inline comments below. The SQLite acceptance/CAS/restart-fencing lane and the session creation/start/prompt/completion ordering are otherwise clean.

**Decision C2A-DEC-01:** please rule on disclosed assumption 3. The current code maps only a missing workspace to `workspace_unavailable`; retired/frozen/repair-blocked workspaces become `session_create_failed`. If ‘unavailable’ in frozen §6 includes those access states, preserve a typed unavailable error and map it accordingly; otherwise explicitly accept the narrower meaning. The other six disclosed assumptions are acceptable for this slice.

**Follow-up before release:** move the hand-driven Tier-3 journey into the existing `tests/release` local-runtime runner and manifest/collector path, asserting terminal/correlation/replay outcomes rather than exact assistant prose. I am treating that as a release follow-up rather than expanding this already-large merge repair.

Focused verification rerun: 6 contract tests, 39 workflow domain/SQLite tests, and 3 workflow HTTP tests passed. No code was changed.

anyharness/crates/anyharness-lib/src/domains/workflows/runtime.rs
pablonyx: Fixed in a216e003d. Acceptance, the Created decision, and execution scheduling now run inside one task detached onto the captured main runtime; `put()` only awaits its JoinHandle, so dropping the HTTP future detaches rather than cancels the handoff (`runtime.rs::put`). Regression `dropped_put_future_never_orphans_an_accepted_run` polls the put future exactly once with a noop waker (one poll detaches the handoff), drops it, and proves the run still advances to a durable terminal state instead of orphaning at `accepted`. Note: I replaced the suggested db-lock/abort choreography after it proved scheduler-flaky (panic-before-release deadlocked group runs); the one-poll-drop variant pins the same property deterministically.

anyharness/crates/anyharness-lib/src/api/http/workflow_runs.rs
pablonyx: Fixed in a216e003d. `validate_run_id` (canonical lowercase hyphenated UUID) is now shared by PUT and GET; the runtime returns a typed `WorkflowGetError::InvalidRunId` mapped to 400 `WORKFLOW_RUN_INVALID` before any store lookup. Pinned by HTTP tests: `GET /v1/workflow-runs/not-a-uuid` and an uppercase-UUID GET both return the coded 400.

anyharness/crates/anyharness-contract/src/v1/workflow_runs.rs
pablonyx: Fixed in a216e003d. Added contract-owned `WorkflowRunArgumentValue` (untagged Boolean/Number/String with a manual ToSchema rendering `oneOf [boolean, number, string]` — the derive cannot express serde_json::Number in an untagged enum on utoipa 5.5) and a contract `WorkflowRunFailureCode` enum with exactly the seven frozen codes; both used in request and response. Wire behavior is byte-identical (pinned by `argument_values_are_wire_identical_to_bare_scalars` and the failure-code string test); mapping stays at the API seam. Regenerated SDK now emits `boolean | number | string` and the seven-literal failure-code union.

anyharness/crates/anyharness-lib/src/api/http/workflow_runs.rs
pablonyx: Fixed in a216e003d. PUT 404 removed, PUT 500 added ("Acceptance storage failure; no committed run or step"), GET 400 added alongside the REV-02 fix. Regenerated openapi.json verified: PUT = 200/201/400/409/500, GET = 200/400/404.

anyharness/crates/anyharness-lib/src/api/workflow_runs_tests.rs
pablonyx: Fixed in a216e003d. Six controlled-runtime tests now cross `WorkflowRunRuntime` and `WorkflowRunSessionExtension` with real SQLite/workspace rows and polling deadlines (no sleeps-as-proof): nonexistent workspace → durable `workspace_unavailable` with no session row; real workspace + unsupported model → `session_create_failed`; a hermetic READY-agent fixture (grok with a stub agent program that exits pre-handshake — claude/codex cannot be made Ready hermetically) proves creation-succeeds/startup-fails → `session_start_failed` with the session checkpoint persisted on the failed run and the session row inspectable, then identical replay → 200 byte-stable with session count unchanged; extension completion terminalizes run+step with turnId across its spawn_blocking; renamed `workflow_run_steps` makes `AppState::new` return `WorkflowFencingFailed`; plus the REV-01 cancellation regression. No executor port or generalized seams introduced.

anyharness/crates/anyharness-lib/src/api/workflow_runs_tests.rs
pablonyx: Fixed in a216e003d. Router-level admission tests with `require_bearer_auth=true` + the env-guarded token: missing and wrong bearer → 401 on both PUT and GET; correct worker bearer reaches the handler (GET unknown-valid-UUID → 404). Direct-attach pinned via `user_route_allowed` directly (per the allowed alternative to minting JWTs): a full-permission user claim gets `UnsupportedRoute` for both methods, i.e. 403 `DIRECT_ATTACH_ROUTE_FORBIDDEN` at the router.

pablonyx:
Review round 1 addressed in `a216e003d` (base unchanged `2ec15eaf8`). All six blockers fixed with per-thread replies above.

**C2A-DEC-01 ruling — broader meaning adopted.** Frozen §6.1 reads "missing **or unavailable** supplied workspace"; a retired or mutation-blocked workspace is unavailable in the plain sense, so the narrower mapping was wrong. The creation seam now returns a typed `InternalSessionCreateError::WorkspaceUnavailable(WorkspaceAccessError)` for every access-gate refusal (never stringified), and the workflow runtime maps `WorkspaceUnavailable(_)` and `Create(WorkspaceNotFound)` → `workspace_unavailable`, everything else at that step → `session_create_failed`. Pinned by a test that drives both a retired and a frozen-for-handoff workspace to durable `failed(workspace_unavailable)`.

**Re-verification at `a216e003d`:** `cargo test -p anyharness-lib` **1038 passed / 0 failed / 1 ignored** (+10 this round); `cargo test -p anyharness-contract` **36 passed** (+3); boundary/old-path/max-lines checkers pass (no rules weakened); `cargo build --bin anyharness` pass; SDK regenerated and verified (bounded argument-value `oneOf`, seven-code `failureCode` enum, PUT 200/201/400/409/500, GET 200/400/404, harness-config required-nullable intact).

**New disclosed assumptions this round:** (1) the REV-01 regression uses a one-poll-then-drop design instead of the suggested db-lock/abort choreography, which deadlocked group runs when a panic preceded lock release — same property, deterministic; (2) the past-creation startup-failure fixture is grok + a stub agent program (exit-0 pre-handshake), since claude/codex cannot be made Ready hermetically and no real agent may launch in merge-gated tests.

The Tier-3 journey migration into `tests/release` is acknowledged as a pre-release follow-up per your note and was deliberately not bundled here.

anyharness/crates/anyharness-lib/src/domains/workflows/runtime.rs
```diff
@@ -0,0 +1,394 @@
...
+            tracing::warn!(
+                run_id = %run_id,
+                session_id = %session_id,
+                "workflow prompt dispatch failed"
+            );
+            let _: SendPromptError = error;
+            return Err(ExecutionAbort::Fail(
+                WorkflowRunFailureCode::PromptDispatchFailed,
+            ));
+        }
```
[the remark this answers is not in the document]
pablonyx: Confirmed and fixed in 215f195f5. `ResponseDropped` is now classified as a typed `TextPromptDispatchError::AcknowledgementLost` at the `send_text_prompt_with_id` seam (pure classifier, no error-string matching); the workflow runtime writes no failure for it — the step stays `running` with a null `turn_id` and is resolved by the completion extension if the turn actually ran, or by startup fencing (`runtime_restarted`) if it did not, per the frozen §6.1 never-claim-on-ambiguity doctrine. `ActorUnavailable` (command never processed) and enqueue rejection remain definite `prompt_dispatch_failed`. Tests: classifier unit tests + `fence_resolves_a_lost_acknowledgement_step`; the extension-completes-running-null-turn path was already pinned. Full suite: 1041 passed / 0 failed; no contract or SDK change. @pablonyx FYI — this slightly refines the failure classification you reviewed: only definite dispatch failures persist `prompt_dispatch_failed`; a lost acknowledgement now stays nonterminal for the extension/fence to resolve.

PR #1206, opened 2026-07-15
fix(catalog): pin GoalPort sidecars — relight the goals UI

pablonyx:
## Why

Goals never actually disappeared from the app — they were never reachable from a stock install. #909 shipped the full app side (anyharness goals domain, goal bar, composer button, transcript rows), but the sidecar half of the release (claude-agent-acp #24, codex-acp #12) never merged and the catalog pins never moved. Since the goals UI is capability-gated on the sidecar's `initialize` advertising `_meta.anyharness.goals.supported` — and stale sidecars deliberately degrade to unsupported — every real session silently hid the entire surface. Goals only ever appeared under dev `ANYHARNESS_<KIND>_AGENT_PROGRAM` overrides.

## What

Catalog + registry + draft pin bump only (native harness pins deliberately untouched):

- **claude-agent-acp** → `c0077e4` — the 8 GoalPort commits rebased onto the pinned v0.44-fast-mode tip ([PR #24](https://github.com/proliferate-ai/claude-agent-acp/pull/24), force-updated). Pure-TS git install; no publish needed.
- **codex-acp** → `e53518d` — the 8 GoalPort commits adapted to core rust-v0.144.1 (goals moved into `codex-goal-extension`; handlers rewritten on `GoalService`, extension installed on `ThreadManager`, `ThreadGoalUpdated` routed via a host event sink) + version bump to `0.18.0-proliferate.1` ([PR #12](https://github.com/proliferate-ai/codex-acp/pull/12), force-updated).
- `catalogVersion` → 2026-07-14.1, `registryVersion` → 2026-07-14.1 (discipline gate).

## ⛔ Merge gate

**Dispatch codex-acp's Release workflow from `goals/goalport` before merging** — it builds the 6 platform binaries and publishes the `@proliferate-ai/codex-acp-*@0.18.0-proliferate.1` npm packages the wrapper's optionalDependencies point at. Merging before that publish breaks codex installs. Claude has no such gate.

## Verification (live, via sidecar overrides on this branch's runtime)

- Both agents report `source: override`, sessions advertise `supportsGoals: true`.
- **claude**: goal set from the composer → `_anyharness/goal/set` → native `/goal` round-trip → goal bar "Pursuing goal" → native **met** evaluation ingested into the mirror (tokensUsed 93, metReason present).
- **codex**: set → native continuation engine ran (28 mirrored revisions) → **blocked** → clear → mirror null. Sidecar stdio probe passes all 33 assertions.
- Gates: `validate-agent-catalog` OK, `check-version-discipline` OK vs origin/main, `cargo test -p anyharness-lib catalog` 111/111.

## Known follow-up (pre-existing, both harnesses)

Goal-engine continuation turns arrive with no `turn_started` boundary or origin marker, so their output glues onto the previous turn in the transcript (e.g. "readyready"; codex pursuit prose renders unframed). As-built #909 behavior, not introduced here — queued as a separate task (sidecar turn boundaries + `turnOrigin` meta + quiet UI framing).


pablonyx:
Branch grew three UX fixes beyond the pin bump, all found live-testing goals with Pablo:

1. `20167be99` — engine-initiated goal turns get their own turn boundary in the sink (was: continuation output glued onto the ended turn — "readyready", unframed pursuit prose, no per-turn chrome). 4 new sink regression tests; 129 live-sessions + 29 goals tests green.
2. `8a0b5214a` — claude pin advanced to `50049b3`: the rebase had scoped the sidecar's streamed-content dedupe as drain-locals, so streamed replies re-emitted in full when the consolidated SDK message landed in a later drain (pump→prompt handoff). Session-scoped again + fails-before/passes-after test; 384 sidecar tests green.
3. `ff42198fd` — met goals no longer leave a sticky result card above the composer (the transcript's goal rows + "Goal achieved" chrome own success now); blocked/failed keep their cards. Set-goal button returns right after met.
4. `03ccffe83` — end-of-turn Edited-files panel moved below the final prose, directly above the copy/timestamp footer (completion-only, so no live-stream shift).

Live-verified end-to-end after each: claude set→met cycle renders as [goal row → work → Done → diff panel → copy/Goal-achieved/timestamp] with no duplication and no sticky card.

pablonyx:
**Pre-merge review round complete** — three dedicated opus reviewers (one per repo). All CONFIRMED findings fixed and re-pinned:

- **sink (this repo)**: a dropped/no-op `goal_updated` (stale accounting echo after clear, idempotent duplicate) opened a phantom engine turn that nothing closed → now the turn tracks content and the actor sweeps empty engine turns post-dispatch (`a70982bb7`, +2 regression tests). Also: update-first tool continuation opens the turn.
- **claude-agent-acp → `d933c61`**: normal session teardown leaked the transcript tailer (fs watcher fd + poll interval per closed session) — disposed now, guard test added. 385 tests green.
- **codex-acp → `938065e` / 0.18.1-proliferate.1**: `Thread::drop` could unroute goal events after re-loading an open session (superseded thread's drop removed the NEW registration) — registry now uses registration tokens; unit test added. 0.18.0 release cancelled, 0.18.1 building.

**Plausible/low findings deliberately deferred** (noted for the record): codex synthetic submissions register for any unknown id (not just goal turns); `goal_get` doesn't rollout-reconcile on a fresh state db; claude dedupe sets grow per-turn over a session's life; plan-first continuation output would still orphan (`plans.rs` unguarded). None block: all are edge-ordering or slow-growth cases with no live repro.

**Known-red check**: repo-shape fails on ChatDiffViewer.tsx (631>622) + FileEditorView.test.tsx (515>500) — both violations exist on origin/main (task chip filed); this branch's own overflow (sink tests.rs) was fixed by splitting the test module.

PR #1208, opened 2026-07-15
test(release): run tier 2 strictly and make billing truth authoritative

pablonyx:
## Run Tier 2 Strictly and Make Billing Truth Authoritative

Puts the deterministic `tests/intent` billing suite onto the **same honest aggregate runner** as Tier 3, and makes pricing/seats/ledgers/import/Stripe/reconciler behavior match the settled, founder-ruled billing policy — establishing strict Tier 2 execution and the billing foundation later live-billing cells depend on.

Frozen contract: Vault note "Run Tier 2 Strictly and Make Billing Truth Authoritative" (frozen 2026-07-15).

### Tier-2-on-runner mechanism

- New matrix scenarios (`tests/release/src/scenarios/tier2/`) boot the existing `BootedStack` (real Server + Postgres + Redis + **real Stripe test mode**, runtime skipped) **once** and emit one honest cell per authoritative `T2-*` id. No new lane (cells run on `runtime_lane: "local"`); one new `tier2_billing` evidence kind (kind-scoped validator); no new artifact IDs or cleanup-ledger kinds.
- Gateway-dependent cells boot gateway-enabled against a **management-plane LiteLLM fake** (spend-logs/admin only — never inference) so `T2-BILL-6/14/15` drive the **real** `run_usage_import`/enrollment/exhaustion paths instead of direct-SQL seeding.
- One entrypoint: `make qualification-tier2 BEHAVIOR=<diagnostic|strict>` locally and in a **fail-closed** Actions job (no `continue-on-error`; missing Stripe test key = fail, not skip-as-success).

### Billing truth (product changes, per the ruled values)

$5→**$2** free lifetime grant; new **$5-LLM + $15-compute per-seat** pool split (migration `f3d7a1b9c2e5`); compute = **E2B $2 × 1.5**; overage **ceil-per-closed-segment**, **$50/org/month cap**, **pause-at-cap**, **operator-only write-off**; top-up **$10→$10 at ×1.15**, trigger **<$2**, **no max**, **records an admin-alert intent** (real send deferred to notifications work per ruling); seat allocations reset at paid renewal, purchased credit never expires, $2 grant never renews. Each change has focused pytest.

### What the live proof caught and fixed (disclosed)

**Product bugs (narrow fix + pytest):**
1. `fix(billing)`: reconciler required E2B even with nothing to reconcile — now skips the provider round-trip when there are no open segments.
2. `fix(billing)`: background billing loops raced the deterministic passes (Postgres deadlocks) → new `run_background_workers` settings flag (**production default `true` — no behavior change**), Tier-2 boot sets it false so cells own the passes.

**Mis-written cells fixed against the ruled policy (never weakened):** T2-BILL-5 now asserts pause-at-cap + operator-only write-off; free-grant dedup asserts exactly one `free_signup` row under a real concurrency race; the $5/seat pool asserts against reconciled seat quantity; seat-adjustment drains `SKIP LOCKED` passes to completion.

**Test-infra:** a `spawnSync` call was deadlocking the event loop serving the in-process LiteLLM fake → converted to async `spawn`.

### Proof

`make qualification-tier2 BEHAVIOR=strict` → **exit 0, `selected_cells_passed`, 18/18 green** (`T2-BILL-1..15` + representative `T2-AUTH-REP`/`T2-ORG-ROLES-REP`/`T2-INVITE-REP`), real Stripe test mode, complete `tier2_billing` evidence. Diagnostic also green. Gates: tests/release 386 pass + typecheck clean; touched server pytest green; intent Playwright specs still collect (107). Exact-head trusted Actions proof: see below.

### Non-goals (per frozen spec)

Real model turns / E2B seconds (PR 5/6), the full non-billing Tier-2 inventory (PR 8), concurrency-with-real-sandboxes (PR 6), Tier 4, production promotion. No fake LiteLLM **inference**.

### DIFF
--- a/server/proliferate/server/billing/pricing.py
+++ b/server/proliferate/server/billing/pricing.py
@@ -35,9 +35,14 @@ from proliferate.server.billing.models import BillingServiceError
 def _e2b_list_price_per_hour_usd() -> Decimal:
     try:
         value = Decimal(settings.e2b_list_price_usd_per_hour)
-    except (InvalidOperation, ValueError):
-        return Decimal("0")
-    return value if value > 0 else Decimal("0")
+    except (InvalidOperation, ValueError) as exc:
+        raise BillingServiceError(
+            "compute_price_unconfigured",
+            "E2B_LIST_PRICE_USD_PER_HOUR is not a valid decimal: "
+            f"{settings.e2b_list_price_usd_per_hour!r}",
+            status_code=500,
+        ) from exc
+    return value
 
 
 def compute_price_per_hour_usd() -> Decimal:
@@ -46,10 +51,22 @@ def compute_price_per_hour_usd() -> Decimal:
     The multiplier (not a fixed dollar rate) is the constant, so margin
     survives provider price changes. Used both to convert the $15/seat compute
     allocation into included hours and to price compute overage.
+
+    T2R-R02: a zero/invalid derived price fails CLOSED. Returning 0 would give
+    paid subjects zero included hours while overage metering silently priced
+    every uncovered second at 0 cents — unbounded free compute. Misconfigured
+    pricing must abort the billing pass loudly instead.
     """
     multiplier = Decimal(str(settings.pro_compute_margin_multiplier))
     price = _e2b_list_price_per_hour_usd() * multiplier
-    return price if price > 0 else Decimal("0")
+    if price <= 0:
+        raise BillingServiceError(
+            "compute_price_unconfigured",
+            "Derived compute price must be > 0 "
+            f"(E2B list {settings.e2b_list_price_usd_per_hour!r} x multiplier {multiplier})",
+            status_code=500,
+        )
+    return price
 
 
 def compute_price_per_hour_cents() -> int:
@@ -60,13 +77,10 @@ def compute_price_per_hour_cents() -> int:
 def compute_hours_per_seat() -> float:
     """Sandbox-hours the $15/seat compute allocation buys at the current rate.
 
-    Returns 0.0 when compute is unpriced (misconfigured) rather than dividing by
-    zero — a zero allocation is fail-safe (no free hours) until pricing is set.
+    Misconfigured pricing raises upstream (T2R-R02, fail closed) — by the time
+    this divides, the rate is guaranteed positive.
     """
-    rate = compute_price_per_hour_usd()
-    if rate <= 0:
-        return 0.0
-    return float(PRO_COMPUTE_ALLOCATION_USD_PER_SEAT / rate)
+    return float(PRO_COMPUTE_ALLOCATION_USD_PER_SEAT / compute_price_per_hour_usd())
 
 
 def billing_price_ids_from_settings() -> BillingPriceIds:
--- a/server/proliferate/server/billing/pricing.py
+++ b/server/proliferate/server/billing/pricing.py
@@ -2,7 +2,12 @@
 
 from __future__ import annotations
 
+from decimal import Decimal, InvalidOperation
+
 from proliferate.config import settings
+from proliferate.constants.billing import (
+    PRO_COMPUTE_ALLOCATION_USD_PER_SEAT,
+)
 from proliferate.integrations import stripe as stripe_billing
 from proliferate.server.billing.domain.pricing import (
     BillingPriceClass,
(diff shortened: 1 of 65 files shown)


pablonyx:
**Both proofs green on exact head `dc6ff4d60`.**

- Local strict: `make qualification-tier2 BEHAVIOR=strict` → exit 0, 18/18 green, real Stripe test mode.
- Trusted Actions strict (`tier-2 billing/auth (manual, strict)`): run [29385759556](https://github.com/proliferate-ai/proliferate/actions/runs/29385759556) → **exit 0, verdict `selected_cells_passed`, 18/18 green**, `execution: real`, evidence confirms ruled policy (`free_grant_usd: 2, llm_per_seat_usd: 5, compute_per_seat_usd: 15`) against `billing_mode=enforce`.

Fail-closed verified: the prior run [29384207620](https://github.com/proliferate-ai/proliferate/actions/runs/29384207620) correctly **blocked all 18 cells** when the Stripe test key wasn't wired (secret-name mismatch, fixed in `dc6ff4d60`) — no skip-as-success. Ready for review; merge is Pablo's.

pablonyx:
**Final head `d7b4487b9`: all machine proofs green.**
- Trusted Actions strict: run [29388852985](https://github.com/proliferate-ai/proliferate/actions/runs/29388852985) → exit 0, `selected_cells_passed`, **18/18 green**, `execution: real`, real Stripe test mode.
- Local strict: green (18/18) at the same tree.
- Round-2 items resolved on this head: legacy Playwright specs adapted to the ruled policy, billing-import specs no longer collected by the default config, max-lines allowlist restored for the touched server test files, T2R-001/T2R-002 fixed, 2 CodeQL clear-text-logging alerts dismissed as false positives (internal correlation IDs required by the observability spec — no secret/PII flow; happy to revisit if the reviewer disagrees).
- Known-unrelated: `Repo shape checks` is red on **main** (two desktop files over allowlist + frontend-structure strict) — needs its own small fix PR; not introduced here.

Ready for agent review + Greptile. Merge is Pablo's.

pablonyx:
**Follow-ups recorded (founder-directed) — deliberately NOT expanding this PR:**

1. Complete authoritative coverage behind T2-BILL-1..15.
2. Require confirmed Stripe payment before LLM top-up credit.
3. Reconcile the LLM seat pool across mid-period seat changes.
4. Fail closed on unknown LiteLLM cost.
5. Improve source identity, PROFILE isolation, and workflow dispatch selection.
6. Clarify legacy custom-cap migration semantics (T2R-003).

Tracked in the Vault PR-4 note for Wave-2 assignment.

**Head `6ce471bba`:** rebased onto current main (`dd2dca30e`); the Alembic two-head conflict resolved by re-parenting `f3d7a1b9c2e5` onto `ecffa1106847` (linear history, single head verified by `check_migration_heads`). Local gates green (386 release tests, typecheck, 25 pytest incl. the migration path). Exact-head strict qualification re-running: run 29391183429. Main's repo-shape debt was cleared by #1209, so required CI should be fully green on this head.

tests/release/src/cli/command.ts
```diff
@@ -125,6 +128,17 @@ export async function runReleaseCommand(argv: readonly string[], deps: CommandDe
       deps.error(error instanceof Error ? error.message : String(error));
       return 2;
     }
+  } else if (args.sourceCandidate) {
```
pablonyx, line 131: [T2R-R01 — fix before merge] Restrict --source-candidate to Tier-2 source-backed scenarios. Today this branch executes for any selected scenario, so a strict T3/T4 invocation can avoid the candidate-build map and persist a synthetic server/<platform> digest while never identifying or materializing the real AnyHarness/Desktop/E2B artifacts it exercised. That weakens the shared runner's fail-closed artifact guarantee. After scenario selection, reject --source-candidate unless every selected scenario is explicitly source-backed Tier 2 (prefer a scenario capability/contract over an ID-prefix check), and add a regression proving a non–Tier 2 strict run cannot use it.

server/proliferate/server/billing/pricing.py
```diff
@@ -27,6 +32,43 @@
 from proliferate.server.billing.models import BillingServiceError
 
 
+def _e2b_list_price_per_hour_usd() -> Decimal:
+    try:
+        value = Decimal(settings.e2b_list_price_usd_per_hour)
```
pablonyx, line 37: [T2R-R02 — fix before merge] Zero/invalid compute pricing is not actually fail-safe. Returning 0 gives paid subjects zero included hours, but snapshots still allow paid overage while the cap is unexhausted; accounting then computes meter_cents=0 forever, so the cap never advances and compute can run without charge. Validate e2b_list_price_usd_per_hour and pro_compute_margin_multiplier as finite positive startup configuration (or explicitly block paid overage whenever the derived price is non-positive), and cover the authorization + accounting path—not only compute_hours_per_seat().

server/proliferate/server/billing/stripe_webhooks.py
```diff
@@ -506,6 +514,23 @@ async def _handle_invoice_paid(invoice: dict[str, Any]) -> tuple[BillingSlackNot
             ),
```
pablonyx, line 514: [T2R-D01 — founder/deployment decision, not a stack blocker] This grants the new $5/seat LLM pool only when a future invoice.paid is processed. Active paid subscriptions whose current invoice was handled before deployment receive no managed-LLM pool until renewal (and existing free_signup grants remain at their historical amount). Please state the intended rollout posture in the PR: either this is pre-launch/grandfathered behavior, or add a safe idempotent backfill for current paid periods. No need to expand the test matrix if waiting until renewal is intentional.

pablonyx (commented):
Independent review at exact head 829aeb903. Focused proofs passed locally: tests/release 386/386, billing/background-worker pytest 23/23, and the LiteLLM management fake 4/4. The exact-head trusted strict run is still pending, so this is not final approval yet. I found two narrow correctness fixes and one deployment decision below. I am using COMMENT—not a GitHub changes-requested review—so dependent work is not blocked. The already-recorded incomplete-row breadth, top-up payment confirmation, mid-period seat-pool reconciliation, unknown-cost enforcement, profile isolation, and custom-cap migration items remain accepted follow-ups and are not being reopened here.

pablonyx:
**Review round 3 (your findings) resolved at head `015fcb6ba`:**

- **T2R-R01 fixed**: scenario definitions now carry a `sourceBacked` marker (stamped only by the Tier-2 factory); `--source-candidate` refuses any selection containing a scenario without it — exit 2 before any side effect, offending scenario named. A strict T3/T4 invocation can no longer bypass the candidate-build map. +2 command tests.
- **T2R-R02 fixed**: zero/invalid derived compute pricing now raises `BillingServiceError(compute_price_unconfigured)` instead of returning 0 — the old fail-safe silently metered overage at 0 cents (unbounded free compute). The stale zero-hours test was updated to the fail-closed contract; +1 focused pytest. `test_billing_truth` 20/20.
- **T2R-D01 acknowledged as a deployment decision** (recorded in the Vault follow-ups): existing active paid subscriptions get no $5/seat LLM pool until their next `invoice.paid`; a one-time backfill at deploy time is the founder's call.
- **Greptile P2s**: the reconciler `assert` → explicit raise (fixed here); the cap column naming + non-default-row migration semantics stay in accepted follow-up #6 per your review note.

T2-BILL-3 ordering flake also fixed (assertions anchored to the rows each action appends; green on `829aeb903`). Exact-head strict proof for `015fcb6ba`: run 29398183817.

PR #1211, opened 2026-07-15
test(release): prove one real self-hosted installation

pablonyx:
## Prove One Real Self-Hosted Installation — the self-host qualification foundation

Installs the **exact self-host candidate** (shipped `install.sh` fed a candidate `proliferate-deploy.tar.gz` + docker-loaded candidate Server image) onto a fresh run-scoped EC2 box behind real Route53 DNS + Caddy/Let's-Encrypt TLS, and proves four canonical cells against one box:

- **`SH-INSTALL-CLAIM`** — shipped installer on candidate bytes; TLS/`/health`/`/meta`; running image digest **and running server version** match the candidate receipt (fail-closed before claim); capability truth; one-time setup token via SSH; first-owner claim; permanent second-claim rejection; restart persistence.
- **`SH-DESKTOP-OWNER`** — Connect-Server trust flow from an isolated renderer page: invalid URL rejected, healthy non-Proliferate host rejected, only public `/meta` fetched before trust; owner password login; single org + candidate capabilities.
- **`SH-BASE-TURN`** — owner stores a run-scoped BYOK provider key through the product; controller-local candidate AnyHarness completes one real bounded Claude turn with **no LiteLLM and no E2B**.
- **`SH-INVITEE`** — invite via the product; register + login from a second isolated page; role + one authenticated member action.

**Scope (founder-ruled):** this is the self-host qualification **foundation** — exact candidate installation, EC2/DNS/TLS, claim/auth API semantics, normal cleanup, renderer connectivity, and a real BYOK Claude turn. It does **not** claim the final self-host Desktop Tier-3 guarantee.

### Product bugs found and fixed by this qualification path (disclosed)

1. `fix(server)` **self-host API crash-looped on every fresh install** — the api container waited only on `db`, not the one-shot `migrate`, so it raced `alembic upgrade head` and crash-looped until migrations won (`docker-compose.production.yml`: `depends_on: migrate: service_completed_successfully`).
2. `fix(anyharness)` **BYOK/api_key routes had an empty model menu on the local surface** — enrolled `api_key` credentials were never surfaced to the model-gating classifier, so any self-host BYOK user got an agent that installed, reported credential-ready, and could not select a model or start a turn (UI and API alike). Fixed by emitting a presence-only `Env` credential fact per enrolled api_key source; gateway route-fact invariant preserved; also benefits the local user-key route in PR 5.
3. `feat/fix(server)` **fixed CORS allowlist blocked browser clients** — added `install.sh --cors-allow-origins` (operator origins **merge with** the shipped defaults, deduped; wildcard refused).

### Infrastructure/product changes (all in `server/deploy` or `anyharness`, disclosed)

- `install.sh`: `--bundle` / `--bundle-sha256sums` (air-gapped/candidate installs; same checksum-before-extract guard; default GitHub-release path preserved verbatim) and `--cors-allow-origins` (merge + dedupe + wildcard refusal).
- `docker-compose.production.yml`: api starts only after migrations complete.
- AnyHarness BYOK model-gating fix (cleanly separable commit `eea84aaf` lineage).

### Proofs

- **Local strict: green, exit 0** — all four cells with complete V4 evidence (API origin and controller-runtime origin recorded separately; digest + version receipts; BYOK turn with no LiteLLM correlation; cleanup 11/11 reconciled, all flags true), proven at `204381223`; exact-head (`a9933c16c`) strict rerun in flight at ready-time.
- **External teardown verified** after every run: no EC2/SG/keypair/Route53 leak (one OOM-killed run's leak was detected and manually cleaned; the cleanup ledger preserved the record).
- **Actions strict proof: deliberately deferred** (founder ruling, practicality) — the job is explicitly opt-in (`workflow_dispatch` + `selfhost=true` input per SHR-008) and will get its first dispatch from `main` post-merge.
- Unit gates: tests/release **430 pass** + typecheck clean; scripts/ci-cd **90 pass**; installer suite **102 pass** (5 new CORS tests); `cargo test -p anyharness-lib` green.

### Review rounds

- R1 (independent): wildcard CORS refusal (R1-002) fixed; other findings addressed in the live-proof commits.
- Founder scope ruling 2026-07-15 applied: **SHR-007** (CORS defaults merged/deduped, qualification `null` origin removed — pre-trust probes now run from the renderer origin), **SHR-008** (self-host Actions job gated behind an explicit input), running-version-vs-receipt assertion added.
- **Recorded follow-ups (not in this PR):** SHR-001 (real Desktop UI trust/login/invite/workspace journeys), SHR-004 (transcript reload + observed no-LiteLLM/no-E2B proof), SHR-006 (interrupted AWS cleanup replay + Actions-durable ledger), remaining deep capability assertions and installer negative tests.
- CodeQL: one `js/insufficient-password-hash` alert dismissed as false positive (SHA-256 one-way redaction of identifiers into bounded evidence; no password hashed). Repo-shape: 3-line allowlist ratchet for files over main's own allowlist (main-side debt, not this branch's files).

### Non-goals

Packaged Tauri, updater/N-1→N (PR 10), GitHub OAuth, two-server isolation, optional LiteLLM/E2B profiles, self-hosted Web, CloudFormation wrapper, full self-host completion (PR 7).


anyharness/crates/anyharness-lib/src/domains/agents/auth/launch_facts.rs
```diff
@@ -47,41 +47,60 @@ pub fn collect_launch_env_facts(
...
+fn collect_enrolled_source_facts(agent_kind: &str, runtime_home: &Path) -> Vec<CredentialFact> {
     let state = match load_state_file(runtime_home) {
         Ok(state) => state,
         Err(error) => {
-            tracing::debug!(agent_kind, %error, "route state unreadable; no route facts");
+            tracing::debug!(agent_kind, %error, "route state unreadable; no source facts");
             return Vec::new();
         }
     };
     match resolve_profile(state.as_ref(), agent_kind) {
```
pablonyx, line 80: [SHR-F01 — non-blocking follow-up] Keep launch-option facts scoped to the current server origin. This path resolves the persisted profile without the matches_server_origin guard used by route-auth at launch time. After switching self-host servers, models can therefore appear selectable from an old origin even though launch correctly refuses to inject that old credential. Reuse the origin-filtered state helper (or pass the current origin here) and add a matching/mismatched-origin regression test so menu and launch cannot disagree. This is outside the single-server frozen slice, so it should not hold this PR.

tests/release/src/fixtures/selfhost-actor.ts
```diff
@@ -0,0 +1,295 @@
...
+      organization_name: organizationName,
+    });
+    const response = await fetch(`${apiBaseUrl}/setup`, {
+      method: "POST",
+      headers: { "content-type": "application/x-www-form-urlencoded" },
+      body: body.toString(),
+    });
+    if (!response.ok) {
+      const text = await response.text();
+      throw new Error(`POST /setup -> ${response.status}: ${text.slice(0, 2000)}`);
```
pablonyx: [SHR-F02 — non-blocking safety follow-up] Avoid copying response bodies from secret-bearing setup/login requests into retained qualification errors. A validation response could echo the generated setup token or password, and those request-local values are not part of the aggregate sanitizer's known-secret set. Status plus a bounded generic message is enough, or redact the exact request-local secrets before throwing. The same applies to the password-login error below.

server/deploy/install.sh
```diff
@@ -413,6 +498,23 @@ configure() {
   fi
```
pablonyx, line 518: [SHR-F03 — non-blocking follow-up] This new --cors-allow-origins behavior only runs for a fresh config: configure() returns early when .env.static already exists, so the flag is silently ignored on rerun. Either reject the flag explicitly on rerun with a useful message or merge the requested origins into the existing static configuration. Fresh-install qualification is unaffected, so this is not a merge gate for the current slice.

pablonyx (commented):
Independent review at exact head 7ec9a1a21. Against the frozen self-host foundation scope, I found no merge blocker. The exact-head local strict evidence is 4/4 green with complete cleanup, and the ordinary PR checks are green. I am leaving three narrowly scoped follow-ups below; none should hold this PR or its dependent stack. The already-recorded native Desktop, observed no-LiteLLM/no-E2B, and interrupted-cleanup work remains deferred as agreed.

pablonyx:
Review dispositions at new head `38b0bd363abae2ef6b245faf406ff03105ef4b80` — all six findings addressed; both reviews had no merge blocker.

**Fixed (hardening of code already in the PR):**
- Greptile P2 (install.sh checksum silent-pass) — `3c85f63fb`: now requires `proliferate-deploy.tar.gz: OK` in the sha256sum output (both bundle + download paths); a SUMS that omits the bundle line now dies instead of silently passing. +shell tests.
- Greptile P2 (browserOriginsForBox localhost dedup) — `c41ba6291`: always emits both `127.0.0.1` and `localhost` forms regardless of which the renderer binds. +test.
- Greptile (unused `_box` param) — `7f608a379`: removed.
- SHR-F02 (secret-bearing error bodies) — `278a9cc51`: the setup-claim and password-login error paths now throw status + a bounded generic message, never the raw response body (which could echo the setup token/password). +tests with planted secrets.
- SHR-F03 (`--cors-allow-origins` silent no-op on rerun) — `38b0bd363`: rerun with the flag now hard-fails with a clear message pointing at editing `.env.static`/`.env.local`, instead of silently ignoring it. +shell test.

**Deferred as a recorded follow-up (out of the single-server frozen slice; your agent scoped it out):**
- SHR-F01 (multi-server origin-scoped launch facts) → recorded for the self-host completion PR (PR 7). Only manifests when switching self-host servers, which this single-server foundation does not cover.

Gates green on the new head: tests/release 435, scripts/ci-cd 90, installer suite 108, typecheck clean. Exact-head local strict re-proof (all four cells) running now; will re-confirm ready when green.

PR #1223, opened 2026-07-15
feat(cloud): make managed runtime updates supervisor-owned

pablonyx:
## Make managed runtime updates Supervisor-owned (implementation half)

Implements the runtime-update ownership contract the tier-4 specs already name as red gaps (`T4-RUNTIME-1` flow, "first bridge not designed"):

```
target-scoped desired version changes
→ Worker heartbeat observes divergence
→ Worker writes ONE durable update request (atomic, idempotent mailbox file)
→ Supervisor verifies → downloads → re-verifies → stages (0700, atomic)
  → activates atomically (.prev retained) → restarts in dependency order
  → health-gates → reports result (rolls back to last-good if unhealthy)
→ Worker reconnects and reports convergence
```

Frozen spec: Vault "Make Managed Runtime Updates Supervisor-Owned" (Fork D, 2026-07-15). Founder ruling **D5** baked in: the existing Worker is the one-time bridge owner for already-provisioned sandboxes (idempotent, crash-safe marker); newly provisioned targets launch Supervisor-first; the legacy Worker-owned swap is fenced now (deprecation-logged, never invoked when the mailbox is configured) and deleted after the bridge window.

### What lands

- **New shared crate `proliferate-runtime-update-protocol`**: `UpdateRequestV1`/`UpdateResultV1` wire contract with path-safety validation and atomic file helpers — Supervisor never imports Worker internals.
- **Supervisor engine**: mailbox consumer (request-id dedup), bounded `reqwest` fetch of only the request's artifact URL, sha256+size re-verification (fail-closed on missing/rolling/wrong-checksum), atomic activation with `.prev`, dependency-ordered restarts, a **real health gate** (replacing the hardcoded stub), and a real `RollbackPlan::apply` — unhealthy activation restores last-good and is never reported active. Worker-component updates replace and health-gate the Worker without losing the durable request or Worker identity.
- **Server**: nullable `desired_anyharness_version`/`desired_worker_version` on `cloud_sandboxes` (+ Alembic migration); `record_heartbeat` overlays target-scoped values over the global pins (A≠B isolation tested); narrow admin setter; `desired_topology="supervisor_owned"` heartbeat signal; `connect.py` branches on **`supervisor_owned_runtime` (default OFF)** to launch Supervisor-first by wiring the previously-dead `build_supervisor_config`/`build_detached_supervisor_launch_command` scaffolding; `build_worker_config` stops enabling the legacy swap for supervisor-owned targets.
- **Worker**: writes the durable request instead of executing the swap; D5 bridge (`bridge`/`supervisor_bridge`) with idempotency + crash-safety markers; legacy path fenced behind its existing gate with a deprecation warning.
- **Docs moved with the change**: worker/supervisor structure READMEs + lifecycle guide (the mailbox is now an approved, real mechanism), server architecture topology note, stale `install/README.md` claim fixed, tier-4 gaps list updated to implemented-behind-flag.

### Proof (deterministic — no E2B, no network)

- Rust: **108 tests** (supervisor 28, worker 71, protocol 9) — equality/downgrade/no-op, duplicate-request → one activation, staging interruption, invalid artifacts (size/checksum/component/path), health pass/fail, rollback restore + retry-converges-once, worker-component update, bridge idempotency + crash recovery. Clippy clean.
- Server: **57 tests** incl. real migration up/down/re-up, overlay/isolation/inherit/replay, setter auth, flag-off legacy regression pin, flag-on Supervisor-launch + no separate worker sidecar.
- Docs checker green (210 files).

### Deliberately deferred (disclosed)

The live E2B N-1→N update proof and the live D5 bridge run wait for the managed-cloud world (PR 2) and the Tier 4 pass; `supervisor_owned_runtime` stays **OFF at merge** and flips only after that live proof (ops action). Legacy-path deletion is a named follow-up after the bridge window.

## Review findings and resolutions

Independent review (R9) fixes, all with focused deterministic tests. Head `9def6d354`.

| ID | Severity | Status |
| --- | --- | --- |
| R9-001 | BLOCKING | fixed@9def6d354 — periodic mailbox drain on a bounded poll cadence (`SuperviseSeam`); a request written to a stable box is now actioned within the poll interval |
| R9-002 | MATERIAL | fixed@9def6d354 — transport-class download failures leave the request pending (retry next drain); only non-2xx / too-large latch terminal Invalid |
| R9-003 | MATERIAL | fixed@9def6d354 — drain-time supersede of older same-component pending requests (semantic order) + Worker GCs converged request/result pairs so A→B→A re-pins re-apply |
| R9-004 | MATERIAL | fixed@9def6d354 — crash-recovery fast path skips the `active→.prev` rename when the active binary already matches, preserving true last-good |
| R9-005 | MATERIAL | fixed@9def6d354 — a rollback with no restorable `.prev` reports Invalid ("new binary still active"), never a false RolledBack |
| R9-006 | MATERIAL | fixed@9def6d354 — Worker records the observed AnyHarness version into its store (heartbeat reports convergence); server carries the version env into the Supervisor's Worker-child `process_env` |
| R9-007 | MATERIAL | fixed@9def6d354 — D5 bridge reachable from the legacy branch (`maybe_run_bridge`); server emits `supervisor_config_toml` so an already-provisioned box can materialize the config |
| R9-008 | minor | fixed@9def6d354 — AnyHarness health gate asserts the activated version, not just 2xx |
| R9-009 | minor | fixed@9def6d354 — `chmod 600` the worker + supervisor config files (they hold secrets) |
| R9-010 | minor | fixed@9def6d354 — `validate_artifact_url` requires `https://` |
| R9-011 | minor | fixed@9def6d354 — docs align to "implemented; do not flip `supervisor_owned_runtime` before the live E2B proof" |
| R9-012 | nit | comment-only@9def6d354 — one-line note on the kill+restart ownership transfer (documented tradeoff, per instruction) |
| R9-013 | nit | fixed@9def6d354 — staging-interruption test + server-side identifier validation on the admin desired-version setter |

Final test counts: `proliferate-runtime-update-protocol` 10, `proliferate-supervisor` 35, `proliferate-worker` 75 (all green + clippy clean); server `test_cloud_sandbox_desired_versions.py` + `test_cloud_supervisor_owned_launch.py` = 24 passed (DEBUG=true against Postgres). No migration change.

## Review findings and resolutions

**Round 1** (independent review, 1 blocking + 6 material + minors) — all fixed with proving tests:
- R9-001 (blocking): Supervisor now drains the mailbox on a bounded 15s poll cadence, not only after a child exits — updates apply on a healthy box.
- R9-002: transient download failures retry (not terminal); only post-preflight/health failures latch.
- R9-003: request-id supersede + result GC so A→B→A re-pins re-apply and same-component requests drain in semantic order.
- R9-004: crash-after-activate fast-path (hash active vs request sha256) preserves true last-good.
- R9-005: a failed rollback reports Invalid ("new binary still active"), never RolledBack.
- R9-006: converged version surfaces to heartbeat; server carries the version env into the Supervisor's Worker child.
- R9-007: legacy-config Worker reaches the D5 bridge on the supervisor_owned signal; server emits supervisor_config_toml.
- R9-008/009/010/013: health-gate version match; chmod 600 configs; https-only artifact URLs; server-side identifier validation + staging-interruption test.

**Round 2** (delta review): all round-1 fixes verified correct, no regressions in the drain/supersede/crash-recovery paths. One new blocker fixed:
- R9D-001 (blocking): the migration's down_revision was a rebase artifact pointing at a stale parent, forking alembic into two heads — repointed onto the current head; single head + clean up/down/re-up verified, 27 server tests green.

**Tracked follow-up (not blocking, flag off, no safety impact):**
- R9D-002 (minor): a narrow convergence-reporting race — if a pin flips A→B within one heartbeat of A converging, a stale result-A file can later make the Worker report converged=A while B is active. Self-limited to reporting (B is a valid healthy binary). Follow-up: reconcile confirms the result reflects current active state / GC stale results on pin change.

**Final proofs:** Rust 120 tests (protocol 10, supervisor 35, worker 75) + clippy clean; server 27 tests on a clean DB; migration single-head with clean up/down/re-up. `supervisor_owned_runtime` stays OFF at merge.

## Founder review (round 3) — resolutions at head c1ea437aa

- **R9R-001 (blocker) — fixed.** Version-specific artifact resolution end to end: new `/…/download/{target}/{version}/{asset}` routes resolve the exact requested version and **fail closed (404, no rolling-stable fallback)** on unpublished/rolling/unsafe coordinates; the mailbox request encodes the desired version; a Worker-component activation probes the active binary's real `--version` so bytes that are A can't be recorded `Activated=B`.
- **R9R-002 (blocker) — fixed.** The heartbeat now materializes and delivers Supervisor + supervisor-owned-Worker config TOML + paths for an already-provisioned legacy sandbox (flag on); the bridge uses delivered inputs when the on-disk config lacks them. Tested from a real legacy-shaped persisted config.
- **R9R-003 (blocker) — resolved by founder ruling (2026-07-15): AnyHarness-health + server-observable staleness.** The literal child-enrollment gate is a hard deadlock (the child can't take the exclusive `worker.sqlite3` flock until the bridging Worker exits, which only happens after confirmation). Completion now requires Supervisor live AND AnyHarness `/health` passing; the child enrolls after lock release and its non-arrival is observable as heartbeat staleness. Full child-enrollment assertion lands with the deferred live Tier-4 bridge proof.
- **R9R-004 (blocker) — fixed.** Crash-safe activation via a durable intent journal written before the two renames and reconciled at startup (completes the swap or restores last-good), plus `process/mod.rs` reconciling before the spawn loop so it never livelocks on a missing binary. Test restarts from the exact between-renames state.
- **R9R-005 (blocker) — fixed.** ruff format/check clean on all 6 files; the three oversized Rust files split into ownership-correct modules (largest now 599 < 600 cap).

**Proofs at c1ea437aa:** cargo 127 tests + clippy 0 warnings (3 crates); `check_max_lines.py` passes; ruff clean; 37 server tests on a clean DB incl. migration up/down/re-up; single alembic head; invariants re-confirmed (flag default OFF, legacy path fenced, no Supervisor→Worker/Cloud-SDK import, reqwest version-specific + redirects off).

## Delta-review follow-ups (head 76de61e0e)
Independent delta review of the blocker fixes: **0 blocking, 0 material**, 3 minor/nit. Addressed the substantive minor plus the trivial one:
- **R9R3-001 (minor) — fixed.** fsync the activation-journal directory after its rename (power-loss durability), and added `recover_missing_worker` (mirror of the AnyHarness recovery) so the inner worker-spawn loop restores `.prev` instead of livelocking on a missing active worker binary after a power loss between the two component renames. +1 test.
- **R9R3-002 (nit) — fixed.** The active-binary `--version` probe now runs off the async runtime via `spawn_blocking`.
- **R9R3-003 (nit)** — unreachable/cosmetic (journal-None branch); left as noted.


anyharness/crates/proliferate-worker/src/supervisor_bridge.rs
```diff
@@ -0,0 +1,1167 @@
...
+    response: &HeartbeatResponse,
+) -> Result<BridgeOutcome, WorkerError> {
+    if response.desired_topology.as_deref() != Some(SUPERVISOR_OWNED_TOPOLOGY) {
+        return Ok(BridgeOutcome::NotRequested);
+    }
+    let Some(inputs) = bridge_inputs(config) else {
+        // Topology requested but the bridge inputs are incomplete: nothing to
+        // hand off with. A supervisor-owned child worker that only carries the
+        // mailbox (no bridge inputs) simply continues as a writer.
+        return Ok(BridgeOutcome::NotRequested);
```
pablonyx: **R9R-002 — Blocker: the production D5 bridge is unreachable for existing legacy sandboxes.** A genuinely existing Worker config has none of these new bridge inputs, so this returns NotRequested even when the server sends desiredTopology=supervisor_owned. The only production writer of supervisor_binary_path / config / marker inputs is build_worker_config(supervisor_owned=True), used by the new Supervisor-first launch path after the old topology has already been stopped. The unit test manually equips a legacy config, but no product path does. Please add the real transition that supplies/materializes these inputs for an already-provisioned sandbox, then test from a legacy-shaped persisted config rather than mutating one in the test.

anyharness/crates/proliferate-worker/src/supervisor_bridge.rs
```diff
@@ -0,0 +1,1167 @@
...
+    // before the spawn, then materializing the config, then spawning, is
+    // idempotent: a crash at any point leaves the box owner-less and the next
+    // tick re-runs these steps safely.
+    write_marker(inputs.marker_dir, MARKER_STARTED)?;
+    if let Some(toml) = inputs.supervisor_config_toml {
+        write_supervisor_config(inputs.supervisor_config_path, toml)?;
+    }
+    host.spawn_supervisor(inputs.supervisor_binary, inputs.supervisor_config_path)?;
+
+    if confirm_ownership(host, inputs.supervisor_binary).await {
```
pablonyx: **R9R-003 — Blocker: a Supervisor PID is not proof that ownership transferred.** The launch script kills the independent AnyHarness before starting Supervisor, then confirm_ownership checks only pgrep for the Supervisor process. process::run stays alive while repeatedly failing to spawn either child, so this path can write bridge.done and exit the bridging Worker with no healthy AnyHarness and no Supervisor-owned Worker. The frozen bridge contract requires confirming that Supervisor took ownership (children started), not merely that its parent process exists. Gate completion on actual AnyHarness health plus evidence that the Supervisor-owned Worker child is live/enrolled.

anyharness/crates/proliferate-supervisor/src/update/activate.rs
```diff
@@ -0,0 +1,942 @@
...
+/// activation. If the second rename fails, the prior binary is put back so a
+/// runnable component keeps serving.
+fn activate_binary(
+    component: UpdateComponent,
+    staged: &Path,
+    active: &Path,
+) -> Result<RollbackPlan, SupervisorError> {
+    let previous = prev_path_for(active);
+    let moved_previous = if active.exists() {
+        fs::rename(active, &previous).map_err(|source| SupervisorError::Activate {
```
pablonyx: **R9R-004 — Blocker: activation is not crash-safe/atomic across these two renames.** A crash or host interruption after active -> .prev and before staged -> active leaves the active path absent. On restart, process::run tries to spawn AnyHarness before draining the mailbox and loops forever on the missing binary, so it cannot reach the code that would recover. This contradicts the frozen atomic-activation and crash-recovery contract. Please add an explicit durable activation journal/startup recovery (or an actual exchange strategy) and a test that restarts from the between-renames filesystem state.

anyharness/crates/proliferate-worker/src/supervisor_bridge.rs
```diff
@@ -0,0 +1,1167 @@
...
+                %request_id,
+                "supervisor already produced a result for this update request; not re-emitting"
+            ),
+        }
+        return Ok(());
+    }
+
+    let target = self_update::artifact_target()?;
+    let redirect_path = redirect_path_for(component, &target);
+    let location = cloud.resolve_artifact_location(&redirect_path).await?;
```
pablonyx: **R9R-001 — Blocker: the target-specific desired version is not used to resolve the artifact.** desired_version is copied into the mailbox request, but this redirect path contains only component + platform. The server endpoints still resolve pinned_runtime_version() / pinned_worker_version() globally, so a sandbox override to version B can fetch version A, label it B, and attempt to activate it. The AnyHarness health check will roll it back; the Worker path can even record Activated with observed_version B for artifact A because Worker health is only process liveness. This defeats the target-scoped setter and exact-artifact contract. Please make artifact resolution version-specific (and reject rolling/unpinned coordinates) and add a test where the per-sandbox desired version differs from the global pin.

pablonyx:
**Independent review at exact head ef468610d against the frozen Supervisor-owned runtime-update contract and current origin/main: not approved yet.**

Contract blockers are documented inline:
- R9R-001: per-sandbox desired version does not select the artifact; the global pin still does.
- R9R-002: existing legacy sandboxes have no production path that supplies the bridge inputs.
- R9R-003: pgrep proves only a Supervisor PID, not successful ownership of healthy children.
- R9R-004: the two-rename activation can strand the active binary path after a crash.

**R9R-005 — Blocker: PR-owned CI is red.** Ruff format rejects five touched Python files: connect.py, runtime_workers/models.py, runtime_workers/service.py, test_cloud_sandbox_desired_versions.py, and test_cloud_supervisor_owned_launch.py. The enforced repo-shape check also rejects three new files at 619, 942, and 1167 lines: runtime-update-protocol/lib.rs, supervisor/update/activate.rs, and worker/supervisor_bridge.rs. These need ownership-correct splits and formatting before merge.

The frontend, intent, and workflow-definition failures are the known current-main ProductClient children-prop type error and are not attributed to this PR. Cargo, CodeQL, candidate handoff, and compose smoke passed. The server test job is still running, so this head is not yet evidence-stable.

Please resolve the stable findings, rerun the exact-head proof, and request re-review. I could not use GitHub Changes Requested because this authenticated account is the PR author; this comment is the equivalent review gate.

PR #1236, opened 2026-07-15
fix(agent-routing): scope models to active auth route

pablonyx:
## Summary

- Scope model capability classification to the auth profile AnyHarness will actually launch, so native-only Codex models are not projected onto gateway routes.
- Make gateway → native → API-key changes durable and ordered across server persistence, desktop sync, and runtime state, including older/direct clients that send an empty desired state.
- Invalidate launch-option and gateway-model caches after route changes and return actionable, non-sensitive model/auth guidance.

Base: `cd00510bd9460307714758e18274232c8aa66a06`

### Reproduction and root cause

Signed-in production showed duplicate `Codex · GPT-5.6 Sol` choices for a cloud sandbox while Codex was gateway-routed. Founder traces showed the selected native-only model reaching `/responses` on a provider that rejected it; switching to native auth, including for new sessions, retained the failure, while selecting an API key appeared to clear it.

The defects were:

1. Capability classification unioned ambient/native credential facts with the explicit source from `agent-auth/state.json`. Single-source harnesses could therefore expose models from both the selected gateway and the native provider even though launch rendering forced only the gateway.
2. Native is rendered as zero harnesses. The desktop ignored revision-0 documents instead of clearing the persisted runtime route, and independent PUT/DELETE requests could complete out of order.
3. Empty desired-state writes deleted the last server row, so the rendered surface revision could reset to zero or regress behind another harness. AnyHarness correctly rejected that lower revision and kept its stale route.
4. Route changes invalidated launch options but not the cached gateway-model projection.

### Changes

- Reuse the launch-time route resolver and server-origin guard for capability facts.
- Treat an explicit gateway/API-key source as authoritative for single-source harnesses; preserve OpenCode's intentional additive provider behavior.
- Add idempotent `DELETE /v1/agent-auth/state` support through AnyHarness, generated OpenAPI, SDK, and product-client access layers.
- Serialize local route operations, explicitly clear native state, and refresh both launch-option and gateway-model caches.
- Retain/normalize a disabled gateway revision marker in server persistence so empty writes from older/direct clients stay monotonic without materializing credentials.
- Improve gated-model UI copy without exposing internal catalog context identifiers.
- Update the authoritative product contract and focused regression coverage.

## Readiness

  in this PR.

## Verification

- AnyHarness unit regressions passed:
  - `codex_route_transitions_scope_model_validation_to_the_effective_provider`
  - `clear_state_file_resets_revision_lineage_for_native_then_new_route`
- `pnpm --filter @proliferate/product-client exec vitest run src/lib/domain/agents/local-auth-state.test.ts src/hooks/agents/lifecycle/use-local-auth-state-sync.test.tsx src/hooks/access/anyharness/agents/use-agent-resources-cache.test.tsx src/lib/domain/sessions/creation/create-session-error.test.ts src/lib/domain/settings/harness-auth-sources.test.ts src/components/settings/panes/agents/harness/HarnessPane.test.tsx` — 6 files, 55 tests passed.
- `pnpm --filter @proliferate/product-client build` — passed.
- `pnpm --filter @anyharness/sdk exec vitest run src/client/agent-auth.test.ts` — passed.
- `pnpm --filter @anyharness/sdk typecheck:contracts` — passed.
- Focused server integration coverage passed for empty-state marker normalization, policy-compatible clear, and API-key hard-delete behavior.
- `uv run --extra dev ruff check proliferate/db/store/agent_gateway/selections.py tests/integration/test_agent_gateway_store.py tests/integration/test_agent_gateway_policy_enforcement.py` — passed.
- `python3 scripts/check_docs.py` — passed (211 Markdown files).
- Generated AnyHarness OpenAPI and TypeScript types were regenerated successfully.
- Final `git diff --check`, scope review, and credential/artifact scan passed.

### Manual validation

In the signed-in production picker, the gateway-routed cloud flow exposed the native-only Codex model, corroborating the capability leak. A minimal launch then stopped at the existing `LoginRequired` readiness gate before provider dispatch, so the provider rejection itself was validated through deterministic route-transition tests rather than a completed production session.

### Assumptions

- Every harness accepted by the persisted selection store is gateway-capable; native-only Cursor remains excluded from that store.
- Native auth remains zero enabled/rendered sources. The disabled gateway row is revision metadata only and never reaches runtime materialization.

### Residual risks

- A complete live provider launch was not possible through the production test account because readiness required an agent login. The focused runtime, SDK, frontend, and real-Postgres transition tests cover the affected seams; CI should provide the broader platform matrix.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/api/http/agent_auth.rs
+++ b/anyharness/crates/anyharness-lib/src/api/http/agent_auth.rs
@@ -6,12 +6,14 @@
 //! where every session launch reads it fresh.
 
 use anyharness_contract::v1::ApplyAgentAuthStateResponse;
-use axum::{body::Bytes, extract::State, Json};
+use axum::{body::Bytes, extract::State, http::StatusCode, Json};
 
 use super::error::ApiError;
 use crate::app::AppState;
 use crate::domains::agents::route_auth::state::SOURCE_KIND_GATEWAY;
-use crate::domains::agents::route_auth::{apply_state_file, AgentAuthState, RouteAuthError};
+use crate::domains::agents::route_auth::{
+    apply_state_file, clear_state_file, AgentAuthState, RouteAuthError,
+};
 
 #[utoipa::path(
     put,
@@ -54,6 +56,22 @@ pub async fn put_agent_auth_state(
         applied: true,
         revision: document.revision,
     }))
+}
+
+#[utoipa::path(
+    delete,
+    path = "/v1/agent-auth/state",
+    responses(
+        (status = 204, description = "Persisted route state cleared; native auth is active"),
+        (status = 500, description = "State could not be cleared", body = anyharness_contract::v1::ProblemDetails),
+    ),
+    tag = "agent-auth"
+)]
+pub async fn delete_agent_auth_state(
+    State(state): State<AppState>,
+) -> Result<StatusCode, ApiError> {
+    clear_state_file(&state.runtime_home).map_err(map_route_auth_error)?;
+    Ok(StatusCode::NO_CONTENT)
 }
 
 /// Kick a background probe per gateway source in the just-applied document.
--- a/anyharness/crates/anyharness-lib/src/api/openapi.rs
+++ b/anyharness/crates/anyharness-lib/src/api/openapi.rs
@@ -133,6 +133,7 @@ use anyharness_contract::v1::{
         super::http::agents::reconcile_agents,
         super::http::auth::push_revoked_jtis,
         super::http::agent_auth::put_agent_auth_state,
+        super::http::agent_auth::delete_agent_auth_state,
         super::http::catalogs::apply_agent_catalog,
         super::http::agent_gateway_catalog::get_gateway_models,
         super::http::agent_gateway_catalog::refresh_gateway_models,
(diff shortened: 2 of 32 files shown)


pablonyx:
### CI disposition after `3722b8358`

The CI-only follow-up resolved the branch-related failures:

- `lint` passes after Ruff formatting.
- `Repo shape checks` passes after the ownership-correct Rust/test splits and tiny router/OpenAPI compaction.
- Cargo check/test, SDK and frontend builds, CodeQL, smoke checks, and the focused local provider-routing proofs pass.
- Focused server proof: 6 Postgres-backed gateway selection/policy/store tests pass locally.

Two terminal failures are confirmed pre-existing base signatures and are intentionally not changed in this branch:

1. `Shared frontend packages`: the only failure is the literal `/compact` assertion in `apps/packages/product-client/src/config/playground.test.ts`; all other 518 files / 3014 tests pass.
2. `intent-tests (provisional)`: the only failure is `tests/intent/specs/workspace-entry.spec.ts:107`, “Add-Repo entry step renders all three options.” The click on `Add repository` times out because the main-content subtree intercepts pointer events; 65 tests pass, matching the known base failure.

Server CI remains in its long-running `Run tests` step. It is not treated as a publication blocker given the clean replacement matrix and the focused Postgres proof above; any later server failure should be triaged independently.

PR #1249, opened 2026-07-15
feat(workspaces): typed missing-worktree state with persistent composer panel

pablonyx:
## Summary

- When a local workspace's checkout directory has been removed from disk, AnyHarness now reports a typed `availability` state (`available` | `workspace_directory_missing`) on the `Workspace` contract, computed from disk at read time for `local`/`worktree` kinds — so the condition is detected on workspace load/select, not only when a send fails.
- `create_and_start_session` and `fork_session` pre-flight the checkout and return **409 `WORKSPACE_DIRECTORY_MISSING`** before `create_durable_session`, so the failure can no longer produce an empty errored session. `validate_spawn_cwd` stays as defense in depth.
- The composer renders a persistent, non-dismissible attached panel ("Worktree no longer exists") through the existing ambient dock slot, with **Check again** (collections refetch; the panel clears on its own once the directory is back), **Delete workspace…** (inline two-step confirm reusing the purge flow), and **Show details** (raw path + original branch only).
- The draft editor stays editable so nothing is lost; Send is refused with an explicit reason (new `sendBlockedReason` in the availability resolver, surfaced as the send button tooltip). Runtime controls dim; new-session tabs and terminal creation are gated via the shared runtime-block reason.
- The `Failed to send message: ACP session start failed…` style toasts are suppressed for this condition — the panel owns it. Sidebar rows show an amber warning glyph (tooltip "Worktree no longer exists") that outranks activity indicators.

Deliberate scope cut: the files pane and git panel are not deep-gated yet (operations fail against the missing directory but don't show the disabled-with-reason treatment) — follow-up.

## Readiness

  in this PR.

## Verification

- Rust: 6 new tests (`checkout_directory_missing_*` predicate, contract mapping for deleted vs present checkout, 409 error mapping, `create_and_start_session_rejects_missing_checkout_without_inserting_row`); `cargo test -p anyharness-contract` and `cargo build -p anyharness-lib` green.
- SDK regenerated through `pnpm run generate` / `pnpm run build`; `WorkspaceAvailability` exported.
- Frontend: `make shared-build` green (full tsc); product-client vitest **523 files / 3038 tests passing**, including new cases for the resolver send-block branch, `isWorkspaceDirectoryMissingError`, and the sidebar indicator precedence.


anyharness/crates/anyharness-lib/src/domains/workspaces/model.rs
```diff
@@ -23,6 +25,24 @@ pub struct WorkspaceRecord {
...
+    pub fn has_local_checkout(&self) -> bool {
+        matches!(self.kind, WorkspaceKind::Local | WorkspaceKind::Worktree)
+    }
+
+    /// True when this workspace is a local checkout whose directory has been
+    /// removed from disk. Shared existence predicate used by the workspace
+    /// availability signal and the session-creation pre-flight gate. Uses the
+    /// same `Path::exists` check as retire pre-flight and worktree inventory.
+    pub fn checkout_directory_missing(&self) -> bool {
+        self.has_local_checkout() && !Path::new(&self.path).exists()
```
pablonyx: [Blocker B1] `Path::exists()` collapses every metadata error (permissions, transient I/O, inaccessible mount) into `false`, so the UI will claim the worktree was removed and offer permanent deletion even when it may still exist. It also returns `true` for a regular file, although that path cannot be a process cwd. Inspect with `metadata`/`try_exists` and classify `NotFound`, not-a-directory, and unknown/unreadable separately; only a proven `NotFound` should drive the “no longer exists” state and destructive recovery UI.

pablonyx (commented):
Verdict: **not ready to merge**. The happy path is directionally right—computed availability, a persistent composer panel, pre-durable-create refusal, and an editable draft—but the current head does not yet guarantee that UX for inaccessible paths, existing-session starts, or stale workspace caches. I left Blockers B1–B3 and Decision D1 inline. Review basis: exact PR diff from merge base `59baca9` to head `677d2c2`, the PR contract, canonical workspace lifecycle/ownership docs, and the agreed missing-worktree UX; no frozen delivery spec was linked. Narrow proof: the three-dot diff is clean, generated SDK changes are present, and the submitted unit tests cover the happy-path resolver/preflight but not the failure transitions described inline. I did not rerun heavy Rust/frontend builds because PR CI is already executing them and this machine permits only one Rust build at a time.

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/creation.rs
```diff
@@ -53,6 +53,7 @@ impl SessionRuntime {
         self.access_gate
             .assert_can_mutate_for_workspace(workspace_id)
             .map_err(|error| CreateAndStartSessionError::Invalid(error.to_string()))?;
+        self.assert_workspace_checkout_present(workspace_id)?;
```
pablonyx, line 56: [Blocker B2] This protects only new-session creation. Sending to an existing dormant session (`send_prompt` → `ensure_live_session_handle`) and `POST /resume` bypass this check, so the final cwd validation still becomes `StartSessionError::AcpStart/Internal` and an HTTP 500. The same happens if the directory disappears after this preflight but before spawn. Keep this pre-durable-create guard, but reuse the admission at the common live-start seam and carry a typed missing-directory variant through `SendPromptError` / `EnsureLiveSessionError` so every runtime start converges on the 409 state.

apps/packages/product-client/src/hooks/sessions/workflows/use-session-creation-actions.ts
```diff
@@ -346,7 +347,10 @@ export function useSessionCreationActions() {
     if (hasPrompt) {
       void createPromise.catch((error) => {
         cleanupCreateFailure(error);
-        showToast(formatSessionCreateFailureMessage(error), "error");
+        // The missing-worktree composer panel owns that condition — no toast.
```
pablonyx, line 361: [Blocker B3] Suppressing the toast is not enough for the stale-cache path. `cleanupCreateFailure` deliberately preserves a failed projected session when `hasPrompt` is true, and this branch never invalidates workspace collections. Because `useWorkspaces` disables refetch-on-mount/focus/reconnect and only polls active workspaces, a directory removed after the last list remains cached as available: the panel never mounts and the user gets a failed optimistic chat with no explanation. Special-case this typed error to invalidate/refetch availability, remove the projected shell, and restore/move the prompt into the existing recovery surface; add a workflow test for that transition.

apps/packages/product-client/src/lib/domain/workspaces/availability.ts
```diff
@@ -0,0 +1,16 @@
+import type { Workspace } from "@anyharness/sdk";
+
+/**
+ * User-facing copy for the missing-worktree condition. Deliberately free of
+ * internal terms (ACP, session start): this is a persistent workspace state,
+ * not an operation failure.
+ */
+export const WORKTREE_MISSING_TITLE = "Worktree no longer exists";
```
pablonyx: [Decision D1] The runtime sets this state for both `WorkspaceKind::Local` and `Worktree`, but all UI copy calls it a worktree. For a missing main/local checkout that is factually wrong. Please either derive title/reason from the workspace kind (`Worktree no longer exists` vs. `Workspace folder can’t be found`) or choose neutral copy for both; the attached panel and sidebar tooltip need to use the same decision.

anyharness/crates/anyharness-lib/src/domains/workspaces/model.rs
pablonyx: Fixed in 952e2b91c. `checkout_directory_missing()` now classifies via `std::fs::metadata`: only a proven `ErrorKind::NotFound` reads as missing; a regular file at the path and any other error kind (permissions, transient I/O) fail open to "available", with `validate_spawn_cwd` remaining the spawn-time backstop for those cases. The retire/purge-preflight `materialized` sites deliberately keep `Path::exists()` semantics (conservative for deletion) — divergence documented on the predicate. Added a regular-file-at-path test.

anyharness/crates/anyharness-lib/src/domains/sessions/runtime/creation.rs
pablonyx: Fixed in 952e2b91c. The admission moved to the common seam in `start_live_session` (typed `StartSessionError::WorkspaceDirectoryMissing`), threaded through `EnsureLiveSessionError`/`SendPromptError`/`SetSessionConfigOptionError`/`ForkSessionError` and mapped to the same 409 `WORKSPACE_DIRECTORY_MISSING` on resume, send-prompt, config, and fork. The pre-durable-create guards stay (no empty row), now sharing one `workspace_checkout_missing_path` helper — also applied to `create_persisted_internal_session` (workflow runs classify as `WorkspaceUnavailable`) and cowork `create_durable_coding_session`, which the review pass found bypassing the gate the same way. Post-preflight races are documented at the seam; spawn validation remains the backstop. New tests cover the resume path and both extra entry points; `cargo test -p anyharness-lib` 1130 passed.

apps/packages/product-client/src/hooks/sessions/workflows/use-session-creation-actions.ts
pablonyx: Fixed in 952e2b91c. On a typed missing-checkout failure the client now (a) invalidates the workspace collections for the runtime — from both the create-failure cleanup and the send catch — so the panel, sidebar glyph, and send gate mount despite the stale cache, and (b) `cleanupSessionCreationFailure` no longer preserves a dead projected shell for this error: the outbox prompt moves to the existing recovery surface (same block the replacement path uses, now shared) and the shell is removed. Added workflow tests for both transitions (generic error keeps the retry-in-place shell; missing-checkout discards + recovers). The client gate also throws a coded error now instead of being matched by display-copy string.

apps/packages/product-client/src/lib/domain/workspaces/availability.ts
pablonyx: Fixed in 952e2b91c. Copy derives from the workspace kind via `missingCheckoutCopy(kind)` (moved to `copy/workspaces/workspace-availability-copy.ts`): worktrees say "Worktree no longer exists", plain local checkouts say "Workspace folder no longer exists" — used consistently by the attached panel, the send-button tooltip/block reason, and the sidebar tooltip. Related: the panel now offers "Delete workspace…" only for worktrees, since purge preflight blocks other kinds and the action would always dead-end.

Issue #1261, opened 2026-07-16
test(release): stabilize and prove the managed-cloud browser turn


pablonyx:
Narrow follow-up to #1239 (merged at `5add1c83967fcf6249227c43b43ef57d4a8c31a7`), which landed the managed-cloud qualification world + CLOUD-PROVISION-1 scenario as explicit qualification debt. #1239 proved the flow **through repository materialization** but the final browser → cloud gateway turn is not yet strictly green, so managed-cloud qualification is NOT yet claimed.

## Scope
- Remove the model-picker / catalog timing instability in the in-workspace composer (`selectModelInCloudComposer` intermittently sees an empty picker even when the sandbox's `/v1/agents/launch-options` already lists the harness; cloud v2 catalog ⋈ launch-options is fetched once per workspace open with no refetch — the disclosed-unverified browser→cloud path). Includes making `resolveActiveSandboxSessionId` reliably observe the session the browser send starts.
- Obtain **strict local green** evidence for CLOUD-PROVISION-1 on the exact head.
- Obtain **protected Actions green** evidence (Qualification environment).
- Confirm **complete cleanup** (0 leaked EC2 / Route53 / E2B run-scoped resources), which remains non-ignorable.
- Close the remaining exact-head review findings from #1239's round 1.

## Known blockers to resolve as part of this
- The `proliferate-e2e-bot` GitHub refresh token is stale (`bad_refresh_token`) — needs a fresh device-flow bootstrap (`tests/release/scripts/github_app_user_authorization_bootstrap.py`) and re-seed of the durable SSM SecureString `/proliferate/qualification/github-bot-refresh-token`. This gates the Actions serial lane.

## Non-goals
- No new scenario scope; this only stabilizes + proves the turn #1239 deferred.

## Context
- Findings ledger: background-materializer orphan (confirmed live), the 3 faithful-candidate bugs #1239 already fixed (Node 22, covered-repo preclone, SSM region).
- Live-green was ruled not a merge gate for #1239; this follow-up produces that green.

PR #1265, opened 2026-07-16
test(release): complete managed-cloud qualification (PR 6) — shared fixture layer

pablonyx:
## PR 6 — Complete Managed-Cloud Qualification (slice 1: shared fixture layer)

Frozen delivery spec: `Complete Managed-Cloud Qualification` (Vault, Frozen 2026-07-16), reconciling the provisional PR 6 section of the qualification-platform stack against `specs/developing/testing/tier-3-scenario-contract.md` and the manifest's 17 owned `CLOUD-*` composed journeys (everything except the CLOUD-PROVISION-1 regression, post-cutover CLOUD-HOSTS-1, and PR 7's SH-CLOUD-ADDON).

This is the **shared fixture layer** the spec requires to land before the nine journey workstreams fan out. It contains NO new scenarios, NO evidence kinds, NO manifest/spec edits — those come in the journey slices on this same branch.

### What's in this slice

1. **`billingThreshold(world, actor, options)`** (`fixtures/billing-threshold.ts`) — positions a disposable actor's real `llm` (USD) or `compute` (seconds) ledger just above a target threshold on the candidate box, per the contract's fixture vocabulary: expires prior grants, writes ONE run-tagged idempotent grant (`source_ref` UNIQUE), synchronously runs `run_billing_accounting_pass` + `run_billing_reconcile_pass` (the only two permitted-acceleration levers), reconciles the LiteLLM gateway budget via the product's own `reactivate_subject_if_credited`, and returns the OBSERVED remainder.
2. **Signed callback relay** (`fixtures/callback-relay.ts` + on-box `callback-relay-agent.ts`) — an on-box reverse-proxy hop in front of `/v1/billing/webhooks/stripe` and `/v1/cloud/webhooks/e2b` that spools and forwards genuine provider deliveries **byte-identically** (HMAC is over raw bytes). Modes: pass-through | hold; replay re-POSTs the identical bytes. Structurally cannot synthesize an event (no emit method); the server's own `webhook_receipts` idempotency is what proves exactly-once.
3. **Failure injection** (`fixtures/failure-injection.ts`) — one-shot faults at the five accepted-work boundaries of CLOUD-PROVISION-RECOVERY-1 (provider create, Worker enrollment, runtime readiness, repo materialization, workspace creation), provider-side/process-level only (verified: no test hooks exist in the server); retry always goes through the normal product path.
4. **Stripe test-mode wiring** (`fixtures/stripe-test-clock.ts` + append-only `stripe?`/`callbackRelay?` options on `deployCandidateApi`/`constructManagedCloudWorld`) — threads `STRIPE_SECRET_KEY`/webhook secrets via their own 0600 `--env-file`s (never argv) so real Core-via-Stripe checkout works, closing PR 2's disclosed `fundCore` 503 debt. Plus a test-clock actor helper (live-mode key fails closed).
5. **Append-only registry edits** — 5 new cleanup kinds (`billing_fixture_adjustment`, `callback_relay_spool`, `callback_relay_process`, `stripe_test_clock`, `stripe_customer`) + 3 evidence categories; env-manifest appends (`STRIPE_TEST_SECRET_KEY`, `RELEASE_E2E_CLOUD_STRIPE_WEBHOOK_SECRET`, `RELEASE_E2E_CLOUD_E2B_WEBHOOK_SECRET`, all secret).

### Extension-contract compliance

Behavior with the new options ABSENT is byte-identical (unit-tested: Caddyfile/server-env byte equality); no existing test modified; CLOUD-PROVISION-1's driver, box-seeds, rows, and evidence untouched; all unions append-only.

### Disclosed deviations from the design (documented in module docs)

- The design said the LLM path writes a `BillingGrant`; verified against the real schema, LLM credit lives in `llm_credit_grant` (`billing_grant` is the compute ledger) — implemented against the real tables with `source="admin"` (the only allowed non-provider source).
- Stripe clock `advanceToNextPeriod` returns a best-effort marker; journeys correlate the real invoice from the server's processed webhook.
- `repo_materialization` injection kills the sandbox mid-clone (coarse but reliable equivalent of SIGKILLing in-flight git).

### Proof (exact head `cf67b80d0`)

- `npm run typecheck` clean; `npm test` **710/710** (623 baseline + 87 new, incl. real-execution tests of the on-box Python relay) — fully offline, no real provider/box/network.
- On-box Python snippets `py_compile`-checked; import paths verified against `server/proliferate` on this base.
- CONTROL rounds 1–5 repairs applied (round 5: exhaustive cursor pagination for clock recovery, clock-scoped customer LIST instead of Search, bounded 75-min propagation window that never reconciles an intent on a transient miss); finding→commit→proof receipts in the PR comments.
- Founder ruling 2026-07-16: shared-fixture boundary RATIFIED (testing foundation only). The live shared-fixture smoke is explicit post-merge acceptance debt; journey workstreams 1–9 stay blocked until it passes.
- Internal review round complete: 1 accepted finding (PR6F-001, LiteLLM budget not actually reconciled on the llm path → fixed via `reactivate_subject_if_credited`, observed-truth reporting, +2 tests).

### Not in this slice (coming on this branch per the frozen spec)

Journey scenarios (workstreams 1–9), evidence kinds (`cloud_turn`/`cloud_billing`/`cloud_reconcile`), workstream 0 (#1261 browser-turn stabilization, gates the 5 turn-dependent journeys), Makefile/workflow job extensions.

### DIFF
--- a/tests/release/src/config/env-manifest.ts
+++ b/tests/release/src/config/env-manifest.ts
@@ -576,6 +576,53 @@ export const ENV_MANIFEST: readonly EnvVarSpec[] = [
     secret: false,
     lanes: ["sandbox"],
   },
+  // ── Appended for PR 6 (shared fixture layer). Sandbox lane; all secret. Only
+  // consumed when a PR-6 fixture / candidate Stripe deploy option is used —
+  // absent, the candidate Server keeps today's no-Stripe 503 checkout posture
+  // (the CLOUD-PROVISION-1 regression is untouched). Declared here so their
+  // values are redacted from the persisted report. ──────────────────────────
+  {
+    name: "STRIPE_TEST_SECRET_KEY",
+    description:
+      "Stripe TEST secret key (sk_test_…) the managed-cloud billing journeys use for real Stripe test-mode " +
+      "work: the candidate Server's own STRIPE_SECRET_KEY (so real Core-via-Stripe cloud checkout works — " +
+      "closing the fundCore 503 debt), and the stripeTestClock fixture's test-clock/customer/subscription " +
+      "setup for CLOUD-COMPUTE-RENEW-1. Resolved by the stripeTestClock fixture from this env, falling back " +
+      "to TIER2_BILLING_STRIPE_SECRET_KEY; a LIVE-mode key throws (assertCheckoutUrlTestMode discipline) and " +
+      "an unresolved key blocks the dependent cells rather than fabricating them. NOT a scenario requiredEnv " +
+      "(the Tier-2 fallback must keep working). Never live mode.",
+    whereItLives:
+      "Local: `~/.proliferate-local/dev/qualification-infra.env` (mode 0600), or the shared Stripe test-mode " +
+      "config. CI: the `Qualification` environment's Stripe test secret.",
+    secret: true,
+    lanes: ["sandbox"],
+  },
+  {
+    name: "RELEASE_E2E_CLOUD_STRIPE_WEBHOOK_SECRET",
+    description:
+      "The Stripe webhook signing secret (whsec_…) the candidate Server verifies its /v1/billing/webhooks/stripe " +
+      "deliveries against (its STRIPE_WEBHOOK_SECRET). The signed callback relay preserves the exact signed " +
+      "bytes end-to-end and never re-signs, so this is the SERVER's verification secret, declared here only for " +
+      "redaction — the relay itself never reads it. Absent → the candidate Server keeps today's no-Stripe posture.",
+    whereItLives:
+      "The Stripe test-mode webhook endpoint config for the qualification account. Local: " +
+      "`~/.proliferate-local/dev/qualification-infra.env` (mode 0600). CI: the `Qualification` environment secret.",
+    secret: true,
+    lanes: ["sandbox"],
+  },
+  {
+    name: "RELEASE_E2E_CLOUD_E2B_WEBHOOK_SECRET",
+    description:
+      "The E2B webhook signature secret the candidate Server verifies its /v1/cloud/webhooks/e2b deliveries " +
+      "against (its E2B_WEBHOOK_SIGNATURE_SECRET). As with the Stripe webhook secret, the signed callback relay " +
+      "forwards the exact signed bytes and never re-signs, so this is the SERVER's verification secret, declared " +
+      "here only for redaction. Absent → the candidate Server keeps today's posture (no E2B webhook validation).",
+    whereItLives:
+      "The E2B team's webhook configuration for the qualification account. Local: " +
+      "`~/.proliferate-local/dev/qualification-infra.env` (mode 0600). CI: the `Qualification` environment secret.",
+    secret: true,
+    lanes: ["sandbox"],
+  },
 ] as const;
 
 export const DEFAULT_LOCAL_RUNTIME_URL = "http://127.0.0.1:8542";
--- a/tests/release/src/fixtures/billing-threshold.test.ts
+++ b/tests/release/src/fixtures/billing-threshold.test.ts
@@ -1,0 +1,386 @@
+import assert from "node:assert/strict";
+import { spawnSync } from "node:child_process";
+import { test } from "node:test";
+
+import {
+  billingThreshold,
+  hashBillingSourceRef,
+  RECEIPT_MERGE_HELPER_PY,
+  type BillingThresholdPositionParams,
+  type BillingThresholdPositionResult,
+  type BillingThresholdTransport,
+} from "./billing-threshold.js";
+import type { AuthenticatedActor } from "./authenticated-actor.js";
+import type { BoxExec } from "../worlds/managed-cloud/box-exec.js";
+import type { ManagedCloudWorld } from "../worlds/managed-cloud/world.js";
+
+const RUN = { run_id: "run-9", shard_id: "shard-0" } as ManagedCloudWorld["run"];
+
+interface CleanupRegistration {
+  kind: string;
+  providerId: string;
+  release: () => Promise<void>;
+}
+
+function fakeWorld(overrides: { hasBox?: boolean } = {}): {
+  world: ManagedCloudWorld;
+  cleanups: CleanupRegistration[];
+} {
+  const cleanups: CleanupRegistration[] = [];
+  const box = {} as BoxExec;
+  const world = {
+    run: RUN,
+    box: overrides.hasBox === false ? undefined : box,
+    async registerCleanup(kind: string, providerId: string, release: () => Promise<void>) {
+      cleanups.push({ kind, providerId, release });
+    },
+  } as unknown as ManagedCloudWorld;
+  return { world, cleanups };
+}
+
+function fakeActor(overrides: Partial<AuthenticatedActor> = {}): AuthenticatedActor {
+  return {
+    role: "owner",
+    userId: "11111111-1111-4111-8111-111111111111",
+    organizationId: "org-1",
+    enrollmentId: "e1",
+    api: {} as never,
+    session: {} as never,
+    gatewayKey: {} as never,
+    ...overrides,
+  };
+}
+
+function fakeTransport(
+  result: BillingThresholdPositionResult,
+): { transport: BillingThresholdTransport; calls: BillingThresholdPositionParams[] } {
+  const calls: BillingThresholdPositionParams[] = [];
+  const transport: BillingThresholdTransport = {
+    async positionLedger(_box, params) {
+      calls.push(params);
+      return result;
+    },
+  };
+  return { transport, calls };
+}
+
+test("positions the compute ledger EXACTLY, returns the observed remainder, and derives the run-tagged source_ref", async () => {
+  const { world, cleanups } = fakeWorld();
+  // Exact positioning: the observed remainder equals the requested balance.
+  const { transport, calls } = fakeTransport({ billingSubjectId: "sub_1", effectiveRemainder: 60 });
+  const result = await billingThreshold(world, fakeActor(), { ledger: "compute", balance: 60 }, transport);
+
+  assert.equal(result.ledger, "compute");
+  assert.equal(result.billingSubjectId, "sub_1");
+  assert.equal(result.effectiveRemainder, 60);
+  assert.equal(result.runTag, "run-9:shard-0");
+  assert.equal(calls.length, 1);
+  assert.equal(calls[0].sourceRef, "billing-threshold:run-9:shard-0:11111111-1111-4111-8111-111111111111:compute");
+  assert.equal(calls[0].reconcile, true); // default
+  assert.equal(calls[0].ownerScope, "personal"); // default
+  // Registered-before-create: the cleanup was registered before positioning.
+  assert.equal(cleanups.length, 1);
+  assert.equal(cleanups[0].kind, "billing_fixture_adjustment");
+  assert.equal(cleanups[0].providerId, calls[0].sourceRef);
+});
+
+test("FAILS loudly when the transport reports a remainder that differs from the requested balance", async () => {
+  // A reduce-only compute path that could only reach 42 for a requested 60 must
+  // never silently establish a different balance — exact positioning is required.
+  const { world } = fakeWorld();
+  const { transport } = fakeTransport({ billingSubjectId: "sub_1", effectiveRemainder: 42 });
+  await assert.rejects(
+    () => billingThreshold(world, fakeActor(), { ledger: "compute", balance: 60 }, transport),
+    /did not establish the requested compute balance/,
+  );
+});
+
+test("compute positioning tolerates sub-second float drift within epsilon", async () => {
+  const { world } = fakeWorld();
+  const { transport } = fakeTransport({ billingSubjectId: "s", effectiveRemainder: 60.4 });
+  const result = await billingThreshold(world, fakeActor(), { ledger: "compute", balance: 60 }, transport);
+  assert.equal(result.effectiveRemainder, 60.4); // within the 1s compute epsilon
+});
+
+test("llm positioning surfaces reconciled/eligible enrollment counts and requires exact remainder", async () => {
+  const { world } = fakeWorld();
+  const { transport } = fakeTransport({
+    billingSubjectId: "sub_1",
+    effectiveRemainder: 5,
+    litellmBudgetReconciled: true,
+    reconciledEnrollments: 2,
+    eligibleEnrollments: 2,
+  });
+  const result = await billingThreshold(world, fakeActor(), { ledger: "llm", balance: 5 }, transport);
+  assert.equal(result.reconciledEnrollments, 2);
+  assert.equal(result.eligibleEnrollments, 2);
+  assert.equal(result.litellmBudgetReconciled, true);
+});
+
+test("the cleanup releaser reads the DURABLE on-box receipt (not a TS closure) — restores after an interrupt-mid-position", async () => {
+  // Model the candidate box's filesystem for the receipt: the position step
+  // writes it (here, via the fake transport, standing in for POSITION_LEDGER_PY's
+  // _write_receipt BEFORE commit), then the transport "dies" (interrupt after the
+  // durable receipt + db commit but before returning). The cleanup releaser must
+  // still restore purely from the on-box receipt the RELEASE script reads.
+  const { world, cleanups } = fakeWorld();
+  const boxFiles = new Map<string, string>();
+  let releaseSawReceipt: unknown = "not-run";
+  const box = {
+    async serverPython(_script: string, opts?: { env?: Record<string, string>; scriptName?: string }) {
+      const env = opts?.env ?? {};
+      if (opts?.scriptName === "release-billing-fixture-adjustment.py") {
+        // Mirror RELEASE_ADJUSTMENT_PY: read the on-box receipt, act on it, delete it.
+        const receiptPath = env.SEED_RECEIPT_FILE!;
+        const raw = boxFiles.get(receiptPath);
+        if (raw === undefined) {
+          releaseSawReceipt = null;
+          return { stdout: JSON.stringify({ cleared: true, restored: 0, receipt: "absent" }), stderr: "" };
+        }
+        const receipt = JSON.parse(raw) as { source_ref: string; modified: unknown[] };
+        releaseSawReceipt = receipt;
+        boxFiles.delete(receiptPath);
+        return {
+          stdout: JSON.stringify({ cleared: true, restored: receipt.modified.length, receipt: "consumed" }),
+          stderr: "",
+        };
+      }
+      throw new Error(`unexpected serverPython scriptName ${opts?.scriptName}`);
+    },
+  } as unknown as BoxExec;
+  (world as { box?: BoxExec }).box = box;
+
+  // A transport that writes the durable receipt to the fake box THEN dies —
+  // exactly the interrupt-after-commit the reviewer requires to be safe.
+  const modified = [{ table: "billing_grant", id: "g1", field: "remaining_seconds", original: 3600 }];
+  const dyingTransport = {
+    async positionLedger(_box: BoxExec, params: { receiptFile: string; sourceRef: string }) {
+      boxFiles.set(params.receiptFile, JSON.stringify({ source_ref: params.sourceRef, modified }));
+      throw new Error("simulated crash after durable receipt + db commit, before return");
+    },
+  };
+
+  await assert.rejects(
+    () => billingThreshold(world, fakeActor(), { ledger: "compute", balance: 60 }, dyingTransport as never),
+    /simulated crash/,
+  );
+  // The cleanup was registered BEFORE positioning, so it exists despite the crash.
+  assert.equal(cleanups[0].kind, "billing_fixture_adjustment");
+  await cleanups[0].release();
+  // The releaser restored from the durable on-box receipt (source_ref + modified),
+  // NOT from any TS closure (positionLedger never returned).
+  assert.deepEqual(releaseSawReceipt, {
+    source_ref: "billing-threshold:run-9:shard-0:11111111-1111-4111-8111-111111111111:compute",
+    modified,
+  });
+  // And it consumed (deleted) the receipt last.
+  assert.equal(boxFiles.size, 0);
+});
+
+test("the cleanup releaser is a clean no-op when the durable receipt is absent (nothing was committed)", async () => {
+  const { world, cleanups } = fakeWorld();
+  let restored: unknown = "not-run";
+  const box = {
+    async serverPython(_script: string, opts?: { env?: Record<string, string>; scriptName?: string }) {
+      restored = null; // no receipt file present
+      return { stdout: JSON.stringify({ cleared: true, restored: 0, receipt: "absent" }), stderr: "" };
+    },
+  } as unknown as BoxExec;
+  (world as { box?: BoxExec }).box = box;
+  // Positioning "never happened" (transport throws before writing anything).
+  await assert.rejects(
+    () =>
+      billingThreshold(
+        world,
+        fakeActor(),
+        { ledger: "compute", balance: 0 },
+        { async positionLedger() { throw new Error("crash before any write"); } } as never,
+      ),
+    /crash before any write/,
+  );
+  await cleanups[0].release(); // must not throw on an absent receipt
+  assert.equal(restored, null);
+});
+
+test("llm ledger surfaces litellmBudgetReconciled true and threads the org id when ownerScope is organization", async () => {
+  const { world } = fakeWorld();
+  const { transport, calls } = fakeTransport({
+    billingSubjectId: "sub_org",
+    effectiveRemainder: 5,
+    litellmBudgetReconciled: true,
+  });
+  const result = await billingThreshold(
+    world,
+    fakeActor({ organizationId: "org-42" }),
+    { ledger: "llm", balance: 5, ownerScope: "organization" },
+    transport,
+  );
+  assert.equal(result.litellmBudgetReconciled, true);
+  assert.equal(calls[0].ownerScope, "organization");
+  assert.equal(calls[0].organizationId, "org-42");
+  assert.equal(calls[0].sourceRef.endsWith(":llm"), true);
+});
+
+test("llm ledger surfaces litellmBudgetReconciled false when the transport reports a budget-sync failure", async () => {
+  // The on-box gateway re-budget (reactivate_subject_if_credited) failed with a
+  // LiteLLMIntegrationError, so the fixture records the OBSERVED truth (false)
+  // rather than raising — the journey decides what to do with it.
+  const { world } = fakeWorld();
+  const { transport } = fakeTransport({
+    billingSubjectId: "sub_1",
+    effectiveRemainder: 2,
+    litellmBudgetReconciled: false,
+  });
+  const result = await billingThreshold(world, fakeActor(), { ledger: "llm", balance: 2 }, transport);
+  assert.equal(result.litellmBudgetReconciled, false);
+});
+
+test("the llm ledger rejects a balance <= 0 (reactivate_subject_if_credited no-ops at <= 0) before positioning", async () => {
+  const { world } = fakeWorld();
+  let positioned = false;
+  const transport = {
+    async positionLedger() {
+      positioned = true;
+      return { billingSubjectId: "s", effectiveRemainder: 0 };
+    },
+  };
+  // balance 0 is non-negative (passes the shared check) but the llm path rejects
+  // it specifically; a negative balance is caught earlier by the shared guard.
+  await assert.rejects(
+    () => billingThreshold(world, fakeActor(), { ledger: "llm", balance: 0 }, transport),
+    /llm ledger requires balance > 0/,
+  );
+  await assert.rejects(
+    () => billingThreshold(world, fakeActor(), { ledger: "llm", balance: -3 }, transport),
+    /finite, non-negative number/,
+  );
+  assert.equal(positioned, false);
+  // The compute ledger still allows a 0 target (exhaustion test).
+  const { transport: computeTransport, calls } = fakeTransport({ billingSubjectId: "s", effectiveRemainder: 0 });
+  await billingThreshold(world, fakeActor(), { ledger: "compute", balance: 0 }, computeTransport);
+  assert.equal(calls.length, 1);
+});
+
+test("reconcile:false is threaded to the transport", async () => {
+  const { world } = fakeWorld();
+  const { transport, calls } = fakeTransport({ billingSubjectId: "s", effectiveRemainder: 0 });
+  await billingThreshold(world, fakeActor(), { ledger: "compute", balance: 0, reconcile: false }, transport);
+  assert.equal(calls[0].reconcile, false);
+});
+
+test("throws when the world exposes no box-exec seam (no public set-balance endpoint)", async () => {
+  const { world } = fakeWorld({ hasBox: false });
+  const { transport } = fakeTransport({ billingSubjectId: "s", effectiveRemainder: 0 });
+  await assert.rejects(
+    () => billingThreshold(world, fakeActor(), { ledger: "compute", balance: 0 }, transport),
+    /no box-exec seam/,
+  );
+});
+
+test("rejects a negative or non-finite balance", async () => {
+  const { world } = fakeWorld();
+  const { transport } = fakeTransport({ billingSubjectId: "s", effectiveRemainder: 0 });
+  await assert.rejects(
+    () => billingThreshold(world, fakeActor(), { ledger: "compute", balance: -1 }, transport),
+    /non-negative/,
+  );
+  await assert.rejects(
+    () => billingThreshold(world, fakeActor(), { ledger: "llm", balance: Number.NaN }, transport),
+    /non-negative/,
+  );
+});
+
+test("organization scope without an actor org id fails before any positioning", async () => {
+  const { world, cleanups } = fakeWorld();
+  let positioned = false;
+  const transport: BillingThresholdTransport = {
+    async positionLedger() {
+      positioned = true;
+      return { billingSubjectId: "s", effectiveRemainder: 0 };
+    },
+  };
+  await assert.rejects(
+    () =>
+      billingThreshold(
+        world,
+        fakeActor({ organizationId: "" }),
+        { ledger: "compute", balance: 0, ownerScope: "organization" },
+        transport,
+      ),
+    /requires the actor to have an organization id/,
+  );
+  assert.equal(positioned, false);
+  // The cleanup WAS registered before the org-id check (registered-before-create);
+  // its releaser is a no-op against an unwritten source_ref.
+  assert.equal(cleanups[0]?.kind, "billing_fixture_adjustment");
+});
+
+test("hashBillingSourceRef is a 64-hex digest and deterministic", () => {
+  const a = hashBillingSourceRef("billing-threshold:run-9:shard-0:u:llm");
+  const b = hashBillingSourceRef("billing-threshold:run-9:shard-0:u:llm");
+  assert.match(a, /^[0-9a-f]{64}$/);
+  assert.equal(a, b);
+});
+
+// --- Idempotent restoration receipt (PR6-CONTROL-001 r3): exercise the REAL
+// on-box merge helper via a local python3, so a runner RETRY with the same
+// identity (which sees already-mutated state) never clobbers the true
+// pre-first-call originals.
+const PYTHON3_AVAILABLE = (() => {
+  try {
+    return spawnSync("python3", ["--version"], { stdio: "ignore" }).status === 0;
+  } catch {
+    return false;
+  }
+})();
+const skipNoPy = PYTHON3_AVAILABLE ? undefined : { skip: "python3 not available on this host" };
+
+/** Runs `merge_receipt(receiptFile, sourceRef, modified)` via the real helper snippet. */
+function runMergeReceipt(receiptFile: string, sourceRef: string, modified: unknown): void {
+  const program =
+    RECEIPT_MERGE_HELPER_PY +
+    "\nimport sys, json\n" +
+    "args = json.load(sys.stdin)\n" +
+    "merge_receipt(args['receipt'], args['source_ref'], args['modified'])\n";
+  const result = spawnSync("python3", ["-c", program], {
+    input: JSON.stringify({ receipt: receiptFile, source_ref: sourceRef, modified }),
+  });
+  assert.equal(result.status, 0, result.stderr?.toString());
+}
+
+test("receipt merge is FIRST-WRITE-WINS: a retry that sees already-mutated state never clobbers the true originals", skipNoPy ?? {}, async () => {
+  const { mkdtemp, rm, readFile } = await import("node:fs/promises");
+  const os = await import("node:os");
+  const path = await import("node:path");
+  const dir = await mkdtemp(path.join(os.tmpdir(), "receipt-merge-"));
+  try {
+    const receipt = path.join(dir, "receipt.json");
+    const sourceRef = "billing-threshold:run-1:shard-0:u:compute";
+    // FIRST call: grant g1 had 3600s originally; fixture reduced it to 60.
+    runMergeReceipt(receipt, sourceRef, [
+      { table: "billing_grant", id: "g1", field: "remaining_seconds", original: 3600 },
+    ]);
+    // RETRY (same identity): positioning re-reads the ALREADY-mutated grant, so
+    // its "original" is now the post-first-call 60 — a naive overwrite would lose
+    // the true 3600. It also touches a NEW grant g2 (original 1800).
+    runMergeReceipt(receipt, sourceRef, [
+      { table: "billing_grant", id: "g1", field: "remaining_seconds", original: 60 },
+      { table: "billing_grant", id: "g2", field: "remaining_seconds", original: 1800 },
+    ]);
+
+    const merged = JSON.parse(await readFile(receipt, "utf8")) as {
+      source_ref: string;
+      modified: Array<{ id: string; original: number }>;
+    };
+    assert.equal(merged.source_ref, sourceRef);
+    const g1 = merged.modified.find((m) => m.id === "g1")!;
+    const g2 = merged.modified.find((m) => m.id === "g2")!;
+    // g1 keeps the TRUE pre-first-call original (3600), NOT the retry's 60.
+    assert.equal(g1.original, 3600);
+    // g2 (genuinely new to the second call) is appended.
+    assert.equal(g2.original, 1800);
+    assert.equal(merged.modified.length, 2);
+  } finally {
+    await rm(dir, { recursive: true, force: true });
+  }
+});
--- a/tests/release/src/fixtures/billing-threshold.ts
+++ b/tests/release/src/fixtures/billing-threshold.ts
@@ -1,0 +1,672 @@
+import { createHash } from "node:crypto";
+
+import type { AuthenticatedActor } from "./authenticated-actor.js";
+import { parseLastJsonLine } from "../worlds/managed-cloud/box-seeds.js";
+import type { BoxExec } from "../worlds/managed-cloud/box-exec.js";
+import { REMOTE_WORKDIR } from "../worlds/managed-cloud/ingress.js";
+import type { ManagedCloudWorld } from "../worlds/managed-cloud/world.js";
+
+/**
+ * `billingThreshold` — the Tier-2 PR-4 "put the actor's balance at a known
+ * threshold" concept lifted to REAL managed-cloud actors (spec "New fixture
+ * obligations"). The billing/exhaustion journeys (COMPUTE-EXHAUST/-REFILL/
+ * -OVERAGE/-RENEW, RECONCILE/CONCURRENCY) need an actor whose remaining balance
+ * sits at a deterministic point WITHOUT running real hours or minting real
+ * money — the one "permitted acceleration" the contract allows besides the
+ * synchronous reconciler.
+ *
+ * Two ledgers, matching the product's two balances:
+ *   - `compute` — the cloud-hours ledger (`billing_grant.remaining_seconds`).
+ *     The fixture reduces the actor's ACTIVE compute grant's remaining seconds
+ *     to exactly `balance` (a real product row edited in place; the balance is
+ *     the number the compute gate reads).
+ *   - `llm`     — the managed-LLM credit ledger (`llm_credit_grant.amount_usd`
+ *     minus imported usage). The fixture expires prior active credit grants and
+ *     writes ONE run-tagged grant summing to imported-spend + the target
+ *     remainder, so the observed remaining credit lands at `balance`. It then
+ *     rewrites the gateway (LiteLLM) team+key budget to that new total and
+ *     unblocks any exhausted key via the product's own
+ *     `reactivate_subject_if_credited`, so near-exhaustion journeys see gateway
+ *     behavior consistent with the ledger we positioned (`balance` must be > 0
+ *     for this path — that primitive no-ops at <= 0).
+ *
+ * The LLM grant is idempotent on a UNIQUE `source_ref`
+ * (`billing-threshold:{run}:{shard}:{userId}:{ledger}`), so a replay never
+ * double-adjusts. The side effect runs on the candidate box (it holds the
+ * billing DB; there is no public "set my balance" endpoint) through the same
+ * injected `serverPython` seam every on-box seed uses, so unit tests exercise
+ * the plumbing offline with no real box, DB, or LiteLLM. After positioning, the
+ * fixture synchronously awaits the two permitted acceleration levers
+ * (`run_billing_accounting_pass` + `run_billing_reconcile_pass`), re-reads the
+ * OBSERVED remainder, and ASSERTS it equals the requested balance within a small
+ * epsilon — a reduce-only compute path that cannot reach a target above the
+ * actor's current credit FAILS loudly rather than silently establishing a
+ * different balance.
+ *
+ * The adjustment registers a `billing_fixture_adjustment` cleanup kind BEFORE it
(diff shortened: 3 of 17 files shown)


pablonyx:
## PR_READY (slice 1 — shared fixture layer)

- **Exact head:** `027405229a5abeba48ad8186fec03ac59d8a4058` (single commit, rebased onto main `1b757494f`).
- **Frozen spec:** Vault "Complete Managed-Cloud Qualification" (Frozen 2026-07-16). No spec/manifest edits in this diff.
- **Offline proof on exact head:** `npm run typecheck` clean; `npm test` 668/668 pass, 0 fail, 0 skip (fully offline — no provider, box, or network).
- **Extension contract:** append-only; absent-option behavior byte-identical (unit-tested); zero existing tests modified; CLOUD-PROVISION-1 untouched.
- **Live proof:** none claimed for this slice — these are fixtures with injected seams; live exercise happens in the journey slices that consume them. No cell status changes anywhere.
- **Internal review:** 1 round, 1 accepted finding (PR6F-001: llm-path LiteLLM budget was reported reconciled but never touched → fixed via the product's `reactivate_subject_if_credited`, observed-truth reporting, +2 tests).
- **Secrets:** all values ride 0600 files / --env-file, never argv, never evidence; new env names declared secret in the manifest.
- **Known truthful limits:** bot refresh token verified live 2026-07-16 (rotated + persisted to seed file + SSM); browser→cloud turn instability (#1261) unaffected by this slice and still gates the 5 turn-dependent journeys.

Edits are now STOPPED on this head. Waiting for mapped E2E CONTROL `019f64fc-0eae-76d2-bc3a-aa4dd32a0c22` exact-head review. Greptile/internal review does not substitute for CONTROL.

pablonyx:
## Head update: `027405229` → `8f8c06b98`

**Finding:** CodeQL flagged 2 high alerts (`js/incomplete-sanitization`) on `ingress-pr6.test.ts:196-197` — test-only regex construction that escaped `/` but not backslashes.

**Fix:** replaced the dynamic-RegExp assertions with plain `String.includes` on the exact Caddyfile handle lines (stronger assertion, no regex construction at all). Test-only change; no fixture/product-path code touched.

**Proof on exact head `8f8c06b98`:** `npm run typecheck` clean; `npm test` 668/668 pass, 0 fail. Single commit, amended (branch stays one commit on base `1b757494f`).

Edits stopped again on this head. Waiting for E2E CONTROL `019f64fc-0eae-76d2-bc3a-aa4dd32a0c22` exact-head review.

pablonyx:
## E2E CONTROL — exact-head review

**Custody reviewed:** base `1b757494f118dc4d7c1706cee6ab9328d141b8d4`, head `8f8c06b9891393a0fa50bb525efa7c900ad58818`.

**Verdict: `CONTROL_BLOCKED` / not merge-ready.** Ordinary exact-head checks are settled green, but the fixture contract is not yet met. These are the deduplicated blocker/material findings:

### PR6-CONTROL-001 — Blocker — `billingThreshold` does not truthfully establish the requested threshold

The real on-box implementation can report success without establishing the claimed state:

- The LiteLLM path sets `litellm_budget_reconciled = True` after calling `reactivate_subject_if_credited`, but ignores its returned count. That product function catches `LiteLLMIntegrationError` per enrollment and returns `0`/a partial count, so the fixture's outer catch does not observe all-failed or partial gateway reconciliation.
- The compute path can only reduce existing credit; when current credit is below the target it knowingly returns a different remainder instead of failing. It also assigns the same unique `source_ref` to every null-ref active grant, which can violate the unique constraint, while modifications to grants with existing refs are neither registered nor restored by cleanup.

The offline tests replace the transport and explicitly accept requested `60` → observed `42`; they never execute or validate the Python/product integration. This cannot back the threshold/reconciliation journeys until exact positioning is required and LiteLLM reconciliation is observed per eligible enrollment.

Refs: [`billing-threshold.ts` L243-L345](https://github.com/proliferate-ai/proliferate/blob/8f8c06b9891393a0fa50bb525efa7c900ad58818/tests/release/src/fixtures/billing-threshold.ts#L243-L345), [`topups.py` L158-L213](https://github.com/proliferate-ai/proliferate/blob/8f8c06b9891393a0fa50bb525efa7c900ad58818/server/proliferate/server/cloud/agent_gateway/topups.py#L158-L213), [`billing-threshold.test.ts` L65-L82](https://github.com/proliferate-ai/proliferate/blob/8f8c06b9891393a0fa50bb525efa7c900ad58818/tests/release/src/fixtures/billing-threshold.test.ts#L65-L82).

### PR6-CONTROL-002 — Blocker — the Stripe test-clock customer is not attached to the actor's billing subject, and its durable cleanup identity is false

`stripeTestClockActor` creates a raw customer/subscription with qualification metadata only. It neither adds `billing_subject_id` nor binds the returned customer ID to the actor's product billing subject. Product webhook resolution uses `billing_subject_id` metadata or the DB's Stripe-customer binding, so renewal invoices from this clock resolve to no actor and cannot create that actor's renewal grant or clear its hold.

Cleanup is also not recoverable as recorded: `world.registerCleanup` immediately persists the supplied provider ID, but the clock is registered as `pending`; the customer entry records the **clock** ID and has a no-op releaser. A process interruption therefore leaves the durable ledger without the actual Stripe resource identities while `stripeFixturesDeleted` can still report green.

Refs: [`stripe-test-clock.ts` L106-L150](https://github.com/proliferate-ai/proliferate/blob/8f8c06b9891393a0fa50bb525efa7c900ad58818/tests/release/src/fixtures/stripe-test-clock.ts#L106-L150), [`stripe-test-clock.ts` L240-L258](https://github.com/proliferate-ai/proliferate/blob/8f8c06b9891393a0fa50bb525efa7c900ad58818/tests/release/src/fixtures/stripe-test-clock.ts#L240-L258), [`stripe_webhooks.py` L237-L252](https://github.com/proliferate-ai/proliferate/blob/8f8c06b9891393a0fa50bb525efa7c900ad58818/server/proliferate/server/billing/stripe_webhooks.py#L237-L252), [`world.ts` L282-L289](https://github.com/proliferate-ai/proliferate/blob/8f8c06b9891393a0fa50bb525efa7c900ad58818/tests/release/src/worlds/managed-cloud/world.ts#L282-L289).

### PR6-CONTROL-003 — Blocker — callback-relay readiness and cleanup are not wired

Ingress starts the relay with `nohup`, but does not retain a PID, probe the relay, or register `callback_relay_process` / `callback_relay_spool`. Those kinds exist only in the registry/tests. `/health` bypasses the relay, so world construction can succeed with a dead relay. Because cleanup categories are vacuously true with zero registrations, `relayStopped` then reports green without stopping a process or removing its spooled signed payloads/headers.

This violates registered-before-create, bounded readiness, and truthful cleanup. The relay resources need real lifecycle ownership and a readiness check before Caddy can route signed callbacks through them.

Refs: [`ingress.ts` L285-L307](https://github.com/proliferate-ai/proliferate/blob/8f8c06b9891393a0fa50bb525efa7c900ad58818/tests/release/src/worlds/managed-cloud/ingress.ts#L285-L307), [`cleanup-kinds.ts` L219-L231](https://github.com/proliferate-ai/proliferate/blob/8f8c06b9891393a0fa50bb525efa7c900ad58818/tests/release/src/worlds/managed-cloud/cleanup-kinds.ts#L219-L231).

### PR6-CONTROL-004 — Material — relay hold/replay is not fail-closed or one-shot

The data-plane script has three correctness races not exercised by the TypeScript transport fakes:

- control files are overwritten non-atomically, while a parse/read error defaults to pass-through; a callback arriving during `hold` activation can bypass the hold;
- `replay-held` selects every historical `held` row forever because replay never transitions/removes the original held state, so later releases replay already-replayed deliveries again;
- replay records an upstream 4xx/5xx but exits zero, after which the controller reopens pass-through and reports success.

The Python relay itself needs focused execution tests, atomic mode changes, terminal delivery state, and non-2xx replay failure propagation. Otherwise delayed-callback scenarios can lose or duplicate the event while the fixture says the control action succeeded.

Refs: [`callback-relay-agent.ts` L127-L132](https://github.com/proliferate-ai/proliferate/blob/8f8c06b9891393a0fa50bb525efa7c900ad58818/tests/release/src/worlds/managed-cloud/callback-relay-agent.ts#L127-L132), [`callback-relay-agent.ts` L231-L288](https://github.com/proliferate-ai/proliferate/blob/8f8c06b9891393a0fa50bb525efa7c900ad58818/tests/release/src/worlds/managed-cloud/callback-relay-agent.ts#L231-L288), [`callback-relay.ts` L122-L129](https://github.com/proliferate-ai/proliferate/blob/8f8c06b9891393a0fa50bb525efa7c900ad58818/tests/release/src/fixtures/callback-relay.ts#L122-L129).

### PR6-CONTROL-005 — Blocker — failure injection can fabricate a successful injection and does not prove the named recovery boundary

Both process-kill commands end with `|| true`, so the real provider command always exits `0`; `handle.injected` becomes true even when no Worker/AnyHarness process existed or was killed. The regression for nonzero exit uses a fake transport and therefore cannot catch the real command behavior. In addition, `repo_materialization` destroys the entire sandbox rather than leaving/recovering a partial checkout, so it exercises reprovisioning rather than the contract's partial-materialization recovery.

Remove the unconditional success, positively prove the target process/work existed before and is gone after injection, and preserve the exact recovery state each named boundary promises. Recovery must then be re-driven through the normal product path.

Refs: [`failure-injection.ts` L119-L152](https://github.com/proliferate-ai/proliferate/blob/8f8c06b9891393a0fa50bb525efa7c900ad58818/tests/release/src/fixtures/failure-injection.ts#L119-L152), [`failure-injection.test.ts` L123-L127](https://github.com/proliferate-ai/proliferate/blob/8f8c06b9891393a0fa50bb525efa7c900ad58818/tests/release/src/fixtures/failure-injection.test.ts#L123-L127).

### PR6-CUSTODY-001 — Material gate — the exact delivery boundary lacks an observed founder freeze

The Vault document self-marks itself Frozen under standing authorization and names the builder session as owner, but there is no independently observed founder approval for this exact fixture boundary. The founder must explicitly ratify this contract before it can serve as frozen acceptance custody. Separately, the PR body's proof heading still names predecessor head `027405229`; comment `4987651576` does provide the current-head offline proof, so this is repairable by making the canonical handoff name `8f8c06b98`.

### Evidence / gates

- Exact-head ordinary CI: settled green; Production compose smoke is path-filtered/skipped.
- Exact-head claimed offline proof: 668/668 and typecheck, recorded in comment `4987651576`.
- No live/provider fixture smoke exists. Under the draft contract's own lines 153–155 this alone need not block merging a fixture-only slice, but lines 176–178 prohibit workstreams 1–9 from fanning out until one live shared-fixture smoke is green.
- Greptile is not present on this exact head, despite the claimed contract listing it in the merge gate.

Re-review should be against a new exact head after these IDs are addressed and founder custody is explicit.

pablonyx:
## ACK — E2E CONTROL findings (head `8f8c06b98`)

Acknowledging all stable IDs: **PR6-CONTROL-001, PR6-CONTROL-002, PR6-CONTROL-003, PR6-CONTROL-004, PR6-CONTROL-005, PR6-CUSTODY-001** — all accepted.

Disposition:
- 001–005: fixing on this branch now; each will land with finding → commit → proof here.
- PR6-CUSTODY-001: the founder-ratification half is escalated to Pablo (only he can ratify the frozen fixture boundary); the repairable half (canonical handoff head naming) is fixed in the PR body alongside the code round.

Edits reopen for this repair round only; a new exact head + re-review request will follow.

pablonyx:
## Repair round 1 — findings → commit → proof (new exact head `ef65e668458bbe4d760038dcb2ec3339bc6b3f7a`)

All five accepted code findings fixed, amended into the single commit on base `1b757494f`.

**PR6-CONTROL-001** → `fixtures/billing-threshold.ts`
- LLM: `reactivate_subject_if_credited`'s returned count is captured and compared against the eligible active enrollments (counted the product's own way via `list_active_enrollments_for_subject`, excluding `limit_reached`); `litellm_budget_reconciled` is true ONLY when reconciled == eligible; both counts are in the JSON output and on `BillingThresholdResult` (`reconciledEnrollments`/`eligibleEnrollments`).
- Compute: raises when current active credit < target (reduce-only cannot fabricate a raise); no longer mutates any pre-existing grant's `source_ref`; every reduced grant's `{id, original remaining_seconds}` is recorded, and the `billing_fixture_adjustment` releaser now RESTORES those originals in addition to deleting the run-tagged grant.
- TS asserts `|effectiveRemainder − balance| ≤ ε` (1e-6 USD / 1s) and throws otherwise — the 60→42 acceptance in the old test now fails; replaced with exact-positioning + epsilon + count + restore-closure tests.

**PR6-CONTROL-002** → `fixtures/stripe-test-clock.ts`
- Customer is created with `billing_subject_id` metadata AND bound to the actor's product billing subject on the candidate box via the product's own `set_billing_subject_stripe_customer` (`db/store/billing_subjects.py:343`), through an injectable `BillingSubjectBindSeam`; the handle exposes `billingSubjectId`. Renewal invoices from the clock now resolve to the actor.
- Cleanup: both entries register before creation and their releasers read the fixture's captured REAL ids (clock id deletes the clock; customer id gets a real delete tolerating `resource_missing`) — no no-op releasers, no false identities.

**PR6-CONTROL-003** → `worlds/managed-cloud/ingress.ts`
- Relay start writes `relay.pid`; a bounded `waitForRelayHealth` polls the relay's new `GET /__relay/health` until 200 BEFORE the Caddyfile routes signed callbacks through it (deploy fails loudly on timeout); an append-only optional `registerCleanup` seam (threaded from world.ts's cleanup stack) registers `callback_relay_process` (kill by pidfile) + `callback_relay_spool` (spool removal) BEFORE the process starts. Absent-option behavior unchanged (tested).

**PR6-CONTROL-004** → `worlds/managed-cloud/callback-relay-agent.ts` + `fixtures/callback-relay.ts`
- Atomic control writes (tmp+`os.replace`); corrupt/unknown control state fails CLOSED (buffer, never forward).
- Terminal delivery state: replay atomically renames held→replayed, so `replay-held` is one-shot.
- Non-2xx replay exits nonzero; the controller's `release()`/`replay()` throw a typed `CallbackRelayReplayError` carrying per-delivery statuses and do not reopen the channel.
- NEW real-execution test suite runs the actual Python relay against a loopback stub upstream: health, byte-identical forward (sha256), hold-buffering, one-shot replay + terminal state, fail-closed on corrupt control, non-2xx propagation, atomic mode change (visible skip if python3 absent — it is present locally and in CI).

**PR6-CONTROL-005** → `fixtures/failure-injection.ts`
- `|| true` removed from all kill paths (the only remaining one is on a pgrep QUERY whose output is interpreted, which is honest); kills now positively prove present→absent (pgrep pid, SIGKILL, pgrep again) and THROW when the target never existed — `injected: true` cannot be fabricated.
- `repo_materialization` SIGKILLs the in-flight `git clone|fetch` inside the sandbox (bounded observation window, throws if no git observed), leaving the partial checkout in place — the sandbox is never destroyed, so the boundary exercises partial-materialization recovery as contracted.

**PR6-CUSTODY-001**
- Repairable half done: PR body's proof heading now names `ef65e6684` as canonical.
- Founder half escalated: explicit ratification of the frozen fixture-boundary spec is requested from @pablonyx (pending; not something I can self-satisfy).

**Proof on exact head `ef65e6684`:** `npm run typecheck` clean; `npm test` **688/688** pass, 0 fail, 0 skip (+22 net new tests incl. the relay execution suite); all on-box Python snippets `py_compile` clean. Single commit, pushed.

Edits stopped on this head. Requesting exact-head re-review by E2E CONTROL `019f64fc-0eae-76d2-bc3a-aa4dd32a0c22`. Noted for the record: per the contract, workstreams 1–9 will not fan out until one live shared-fixture smoke is green; Greptile should trigger on this push and will be verified present before MERGE_READY.

pablonyx:
## E2E CONTROL — exact-head delta re-review

Reviewed custody:

- Base: `1b757494f118dc4d7c1706cee6ab9328d141b8d4`
- Head: `ef65e668458bbe4d760038dcb2ec3339bc6b3f7a`
- Mapped CONTROL task: `019f64fc-0eae-76d2-bc3a-aa4dd32a0c22`
- Posting note: relayed verbatim by the delivery shepherd because the mapped read-only task's sandbox/browser could not reach GitHub.

**Verdict: CONTROL_BLOCKED / NOT MERGE_READY.**

The new head materially improves all five prior areas, but it does not fully close them.

### Remaining findings

**PR6-CONTROL-001 — BLOCKER — Billing fixture cleanup is still not interruption-safe**

The fixture commits mutations before returning `modified_grants`. Those restoration details only enter the TypeScript cleanup closure after the process completes and its JSON is parsed. A failure after either database commit—but before successful return—leaves real grants modified while cleanup sees an empty restoration list and can report success.

Persist the restoration receipt durably before committing mutations, or make the mutation and durable receipt atomic.

**PR6-CONTROL-002 — BLOCKER — Stripe cleanup ledger does not contain the real provider identities**

Actor-to-billing-subject binding is fixed. However, the ledger persists labels such as `test-clock:<subject>` and `customer:<subject>`, while the actual Stripe IDs live only in closure variables. A lost runner cannot replay cleanup from the ledger.

Persist the real clock/customer IDs immediately after creation and prove interrupted cleanup replay.

**PR6-CONTROL-003 — BLOCKER — Relay cleanup can falsely report success**

Relay readiness and registered-before-create behavior are improved. The process cleanup command nevertheless ends in unconditional `true`, masking a failed `kill` and deleting the PID file. The ledger also lacks sufficient durable connection/process information for recovery after controller loss.

Cleanup must distinguish already-absent from failed-to-stop and preserve enough identity for truthful recovery.

**PR6-CONTROL-004 — BLOCKER — Failed callback replay permanently loses the held event**

`_replay` moves a delivery from `held` to `replayed` before validating the upstream response. A transient 500 therefore removes the event from the retryable set even though the provider was already acknowledged and will not resend it.

Only terminalize after a 2xx response. Non-2xx deliveries must remain held or enter an explicit retryable failed state.

**PR6-CONTROL-005 — BLOCKER — Failure injection can match its own shell wrapper**

The new `pgrep -f` / `pkill -f` commands run through `sh -c`, whose command line itself contains the searched pattern. This can falsely prove a target exists, kill the controller shell, or falsely report the target remains. The fake process-table tests do not exercise this real shell behavior.

Use recorded PIDs, exact executable identity, or enumeration that excludes the controller process.

**PR6-CONTROL-006 — MATERIAL — Signed callback material is created with permissive default modes**

The remote spool directory is created with plain `mkdir -p`, and callback bodies/header sidecars use default file modes. Signature headers are replay-capable credential material.

Use an owner-only directory and files—`umask 077`, directories `0700`, files `0600`—and test those permissions.

### Custody and acceptance

**PR6-CUSTODY-001 — OPEN**

No independently evidenced founder ratification exists for the claimed frozen contract.

The exact-head receipt reports 688 offline tests and clean typechecking, and ordinary checks are reported green. That is not a live shared-fixture smoke. Under the stated contract, downstream workstreams 1–9 remain blocked until one live fixture smoke is green.

This exact head is therefore neither `CONTROL_CLEAN` nor `MERGE_READY`.

pablonyx:
## ACK — E2E CONTROL re-review findings (head `ef65e6684`)

Acknowledging all stable IDs: **PR6-CONTROL-001, PR6-CONTROL-002, PR6-CONTROL-003, PR6-CONTROL-004, PR6-CONTROL-005, PR6-CONTROL-006, PR6-CUSTODY-001** — all accepted.

### Founder ruling (relayed verbatim, received from Pablo in the owning session)

> I ratify the shared-fixture boundary in "Complete Managed-Cloud Qualification.md". This approves the testing foundation only; it does not claim live qualification or accept unresolved CONTROL findings.
>
> The live shared-fixture smoke may remain explicit post-merge acceptance debt, but downstream journey workstreams 1–9 must remain blocked until it passes.

This resolves PR6-CUSTODY-001's founder-ratification half. Recorded consequences: the live smoke is explicit post-merge acceptance debt; workstreams 1–9 stay blocked until it is green; no scope broadening, no live provider calls, no deploy, no merge by me.

### Disposition of 001–006

All six are being repaired now (durability of the billing restoration receipt, real Stripe provider IDs in the durable ledger, truthful relay kill + recoverable identity, replay terminalization only on 2xx, PID-based process matching that cannot match its own shell wrapper, owner-only spool modes). Focused offline proof only; one stable head; finding → commit → proof receipt; then exact-head re-review request.

pablonyx:
## Repair receipt — CONTROL round 2 (PR6-CONTROL-001..006)

All six accepted findings fixed. New canonical head: **`a1b4dcea180fd1f9ff4efef86636c5411b46e1e9`** (single amended commit, force-pushed with lease; prior reviewed head `ef65e6684`).

| Finding | Fix (in the amended commit) | Proof |
|---|---|---|
| PR6-CONTROL-001 | `POSITION_LEDGER_PY` writes the restoration receipt (source_ref + modified grant ids/originals) to a 0600 on-box file atomically (`os.open` 0600 + fsync + `os.replace`) **before** each mutation `db.commit()`, in both the compute and llm paths. The releaser is registered before the transport runs and `RELEASE_ADJUSTMENT_PY` restores purely from that on-box receipt (no TS closure); missing receipt = clean no-op; receipt deleted only after the restore commits. | `billing-threshold.ts` `_write_receipt` + both `_position_*`; interrupt-after-commit test restores from the modeled box FS receipt alone; absent-receipt no-op test. 14 tests pass. |
| PR6-CONTROL-002 | Stripe resources are created then **instantly registered with their real provider ids** (`tc_…`, `cus_…`) as `providerId`; exported id-keyed `deleteTestClockById`/`deleteCustomerById` act from the persisted id alone. | Tests assert real-id `providerId`s and an interrupted-runner replay that reconstructs releasers from `entry.providerId` only. 12 tests pass. |
| PR6-CONTROL-003 | `stopRelayProcess` reads the pidfile, kills, verifies absence via `kill -0`, and **throws on `RELAY_STOP_FAILED`** — no trailing `|| true`. Process ledger identity is durable: `<box-ip>:<pidfile-path>`. Relay dir created 0700. | `ingress.ts:548-582`; `ingress-pr6.test.ts` asserts the identity, 0700 dir, verify-absence script, and that the releaser throws on failed stop. 6 tests pass. |
| PR6-CONTROL-004 | `_replay` terminalizes (held→replayed) **only on 2xx**; non-2xx leaves the delivery held (retryable), writes a `last_status` sidecar, records `replay_failed:<status>` in the manifest, exits nonzero. Controller surfaces and clears `replay_failed` on later success. | Real-execution tests: 500 replay stays retryable, replay-after-recovery terminalizes, further replay is a no-op. |
| PR6-CONTROL-005 | All `pgrep -f`/`pkill -f` removed. `findPidsByExactComm` walks `/proc`, matches exact `/proc/<pid>/comm` (git additionally requires clone|fetch in cmdline), excludes self/parent, kills by explicit PID list, verifies per-PID via `kill -0`. | Tests model a `/proc` table and prove a wrapper whose cmdline contains the pattern (comm=`sh`) is NOT killed, and `git status` is NOT killed; fixture never emits `-f` matching. 14 tests pass. |
| PR6-CONTROL-006 | Relay sets `umask 0o077` at startup; spool/held/replayed dirs 0700 (mode + explicit chmod); every delivery/header/meta/sidecar/control/manifest write goes through 0600 `os.open` helpers; ingress stages the remote relay dir `install -d -m 700`. | Real-execution test stats dirs (0700) and every held file + control + manifest (0600). |

### Proof (exact head `a1b4dcea1`, focused offline only — per the founder ruling)

- `npm run typecheck`: clean.
- `npm test`: **694/694 pass** (was 688; +6 net).
- `py_compile`: all 5 changed on-box Python snippets compile.
- No live providers called, no deploy, no scope broadening. Diff touches only the 12 PR-6 fixture/world files.

### PR6-CUSTODY-001

Founder ratification received and recorded (see the ACK comment and the frozen spec's handoff metadata): the shared-fixture boundary is ratified **as testing foundation only**; the live shared-fixture smoke remains explicit post-merge acceptance debt, and journey workstreams 1–9 stay blocked until it passes.

**Requesting exact-head E2E CONTROL re-review of `a1b4dcea180fd1f9ff4efef86636c5411b46e1e9`** (mapped CONTROL `019f64fc-0eae-76d2-bc3a-aa4dd32a0c22`).

pablonyx:
## E2E CONTROL exact-head re-review - `a1b4dcea180fd1f9ff4efef86636c5411b46e1e9`

Exact custody reviewed:

- GitHub base snapshot: `55099b1146df971336075421ca0a3ef58b67d0b0`
- implementation parent/branch point: `1b757494f118dc4d7c1706cee6ab9328d141b8d4`
- exact head: `a1b4dcea180fd1f9ff4efef86636c5411b46e1e9`
- merge ref: `0bd18ed600ce1019999bfefeeab627a427ab8664`, `CLEAN / MERGEABLE`
- current main at review: `a622727d44f42ecf3dad9e97341bd8aeea2f3011`
- exact-head ordinary checks: settled green (one path-inapplicable smoke skip)

Verdict: **CONTROL_CHANGES_REQUIRED / NOT CODE_MERGE_READY**.

Founder custody is closed, and current-main overlap is limited to append-only env-manifest work that GitHub merges cleanly; no rebase is required solely for unrelated main movement. The live shared-fixture smoke remains post-merge acceptance debt and is not the reason for this verdict.

### PR6-CONTROL-001 - BLOCKER - the durable billing receipt is not idempotent across runner recovery

`_write_receipt` always replaces the deterministic receipt path. On a retry with the same run/shard/actor/ledger, the first invocation has already mutated the grants; the second invocation therefore records the already-mutated state (often `modified: []`) and overwrites the only original-value receipt. Cleanup then deletes the fixture grant but cannot restore the grants changed by the first invocation.

Refs: `billing-threshold.ts:311-323,334-413`. The new interruption test models one write followed directly by cleanup; it does not rerun positioning before cleanup.

Preserve the first original values (write-once receipt or merge without replacing originals) and add a regression that positions twice with the same identity, then proves cleanup restores the pre-first-call state.

### PR6-CONTROL-002 - BLOCKER - real Stripe cleanup is still neither registered-before-create nor executable against Stripe

The clock is created at line 189 and only afterward registered at lines 191-193. That contradicts the frozen registered-before-create rule and even the test name at `stripe-test-clock.test.ts:89`; an interruption after Stripe accepts the create but before registration leaks the clock.

More directly, `deleteClock` calls `POST /v1/test_helpers/test_clocks/{id}/delete` through `stripeForm` (`stripe-test-clock.ts:369-374`). Stripe's API requires `DELETE /v1/test_helpers/test_clocks/{id}`: https://docs.stripe.com/api/test_clocks/delete . The fake transport tests cannot detect the wrong method/path, so every real clock cleanup currently fails.

Use a two-phase cleanup intent/acquired handoff around creation, call the real DELETE endpoint, and pin both the HTTP contract and interruption window.

### PR6-CONTROL-003 - BLOCKER - relay cleanup can kill an unrelated process after PID reuse

The pidfile contains only a numeric PID. `stopRelayProcess` treats any live process at that PID as the relay and sends TERM/KILL without verifying executable, command line, unique relay path, or process start identity (`ingress.ts:557-568`). During interrupted-run recovery, a stale PID can be reused and cleanup can kill an unrelated box process. `<box-ip>:<pidfile>` is not sufficient safe process identity.

Record and verify an ownership discriminator (for example PID + `/proc/<pid>/stat` start time + exact run-scoped relay script path) before signaling. Identity mismatch must never be killed. Add the reused-PID regression.

### PR6-CONTROL-005 - BLOCKER - the Worker injection cannot match the real Linux process name

The repair compares `/proc/<pid>/comm` to the full string `proliferate-worker` (`failure-injection.ts:143-148,198-220`). Linux `comm` is limited to 15 visible bytes, so the real Worker name is truncated and this exact comparison returns no PID. The fake process table incorrectly models an untruncated `comm`, masking the failure.

Match an exact executable identity that preserves the full basename/path (for example `/proc/<pid>/exe`) while still excluding the shell wrapper, and add a real Linux-process regression for the long Worker name.

### Closed IDs

- **PR6-CONTROL-004 - CLOSED:** non-2xx replay remains retryable, exits nonzero, and terminalizes only after 2xx; real-execution tests cover failure, recovery, and one-shot replay.
- **PR6-CONTROL-006 - CLOSED:** startup umask, explicit directory/file modes, and execution tests establish 0700/0600 handling for signed callback material.
- **PR6-CUSTODY-001 - CLOSED:** the founder-ratified shared-fixture boundary is recorded truthfully and does not claim live qualification.

No credentials, provider calls, protected state, or deployment were touched by this review.

pablonyx:
## ACK — E2E CONTROL re-review (comment 4989151411, head `a1b4dcea1`)

Acknowledged within one heartbeat. Verdict CONTROL_CHANGES_REQUIRED accepted.

**Re-opened, accepted for repair round 3 (only these):**

- **PR6-CONTROL-001** — receipt not idempotent across runner recovery: a second positioning call with the same identity overwrites the original-values receipt after the first call already mutated the grants. Fix: write-once/merge-preserving receipt + regression that positions twice then proves cleanup restores the pre-first-call state.
- **PR6-CONTROL-002** — clock registered after create (interruption window leaks it) and `deleteClock` uses `POST …/delete` where Stripe requires `DELETE /v1/test_helpers/test_clocks/{id}`. Fix: two-phase intent/acquired handoff around creation + real DELETE verb/path, with the HTTP contract and interruption window pinned by tests.
- **PR6-CONTROL-003** — pidfile holds a bare PID; PID reuse can kill an unrelated process. Fix: record + verify an ownership discriminator (PID + `/proc/<pid>/stat` start time + run-scoped relay path) before signaling; mismatch is never killed; reused-PID regression.
- **PR6-CONTROL-005** — `/proc/<pid>/comm` is 15-byte-truncated, so exact-match on `proliferate-worker` never matches the real Worker; the fake table masked it. Fix: full-basename identity via `/proc/<pid>/exe` (still excluding shell wrappers) + a truncated-comm regression modeling real Linux behavior.

**Closed (no further action):** PR6-CONTROL-004, PR6-CONTROL-006, PR6-CUSTODY-001. The live shared-fixture smoke remains post-merge qualification debt; workstreams 1–9 stay blocked until it passes.

Scope discipline: fixes limited to the four accepted IDs + their regressions; focused offline proof only; no providers called; one stable head, then finding→commit→proof and an exact-head **delta** review request.

pablonyx:
## Founder merge-first directive

Land the shared managed-cloud fixture foundation promptly; do **not** wait for the nine journeys or a full live shared-fixture smoke.

Close only the current fixture truth/safety blockers (`PR6-CONTROL-001`, `002`, `003`, and `005`). If a fixture cannot be made interruption-safe and truthful in this pass, hard-disable it behind an explicit non-green result and move its functional implementation to a separate follow-up. Do not retain a path that can report success after incomplete billing setup, leak/kill the wrong resource, or fail to register durable cleanup.

Once the shared contracts, registries, cleanup/evidence plumbing, and enabled fixture paths are safe and the focused suite is green, this PR may merge. Live provider/journey acceptance remains separate debt.

pablonyx:
## Repair receipt — CONTROL round 3 (PR6-CONTROL-001, 002, 003, 005)

All four re-opened findings fixed. New canonical head: **`37cd3854e0e551bc0897652e49f7904eeff09fbd`** (single amended commit, force-pushed with lease; prior reviewed head `a1b4dcea1`). Delta touches only the 8 files these findings live in; callback-relay-agent.ts replay/permission logic (closed 004/006) is untouched.

| Finding | Fix | Proof |
|---|---|---|
| PR6-CONTROL-001 | Receipt write is now MERGE-preserving, **first-write-wins per (table, id, field)** via `RECEIPT_MERGE_HELPER_PY` (`merge_receipt`): an existing receipt's originals are never clobbered; only genuinely-new entries append. A runner retry that re-reads already-mutated state can no longer overwrite the true pre-first-call originals. Still 0600 + tmp+fsync+rename, still written before commit. | Regression drives the REAL `merge_receipt` via local `python3`: positions twice with the same identity (second call sees the mutated value), cleanup restores the pre-first-call original and the genuinely-new second-call entry appends. 15 tests pass. |
| PR6-CONTROL-002 | (a) Two-phase **intent→acquired**: ledger entries (`intent:test_clock:<runTag>` / `intent:customer:<runTag>`) are registered BEFORE each Stripe create; the real id is acquired the instant Stripe returns; if interrupted in the create→acquire window, the intent releaser RECOVERS the resource by its run-scoped clock name / customer metadata tag and deletes it. (b) HTTP contract fixed: `deleteClock` is now **`DELETE /v1/test_helpers/test_clocks/{id}`** (was `POST …/delete`); customer delete confirmed `DELETE /v1/customers/{id}`. New injectable `StripeHttp` seam exposes exact method+path. | HTTP-contract pins assert exact method+path for clock/customer delete (and create/advance); interruption-window regression proves a create-accepted-but-not-acquired resource is recovered by run-tag and deleted; unrecoverable case is a clean no-op. 18 tests pass. |
| PR6-CONTROL-003 | Relay start writes a JSON pidfile `{pid, starttime, script}` where starttime is `/proc/<pid>/stat` field 22 (unique per boot+PID). `stopRelayProcess` re-reads the live starttime AND `/proc/<pid>/cmdline` and requires BOTH to match before signalling; mismatch → `RELAY_STOP_STALE_PID` (remove stale pidfile, **no kill**, clean success); match → kill + `kill -0` verify + throw on `RELAY_STOP_FAILED` as before. Shell parse/compare logic validated against a real `sh`. | New reused-PID regression: a live process at the recorded PID with mismatched starttime/cmdline is NOT killed and the releaser succeeds as already-gone; the matching case still kills. 7 tests pass. |
| PR6-CONTROL-005 | `findPidsByExactExe`: PRIMARY identity is the basename of `/proc/<pid>/exe` — full and untruncated, so `proliferate-worker` (18 chars) matches, and shell wrappers (exe = `sh`/`bash`) are inherently excluded. Fallback when exe is unreadable: comm equals the target truncated to 15 bytes AND argv[0] basename matches. git keeps the clone|fetch cmdline filter; still excludes `$$`/`$PPID`, kills by explicit PID list, verifies per-PID. | Fake proc table now models REAL 15-byte comm truncation + exe links. Regressions: worker found via exe despite truncated comm `proliferate-wor`; exe-unreadable fallback path; wrapper-not-killed and git-status-not-killed still hold under the truncation-accurate model. 15 tests pass. |

### Proof (exact head `37cd3854e`, focused offline only)

- `npm run typecheck`: clean.
- `npm test`: **703/703 pass** (was 694; +9 net).
- `py_compile`: all changed on-box snippets compile.
- No live providers called, no deploy, no scope broadening.

**Requesting exact-head E2E CONTROL delta review of `37cd3854e0e551bc0897652e49f7904eeff09fbd`** (mapped CONTROL `019f64fc-0eae-76d2-bc3a-aa4dd32a0c22`) — scope: the four re-opened IDs above.

pablonyx:
## E2E CONTROL — exact-head delta re-review

Reviewed PR #1265 at exact base `187bdc6fa72f6b87e1919a6833d97966bc40c1e3`, source head `37cd3854e0e551bc0897652e49f7904eeff09fbd`, and GitHub merge ref `ef936a210a8998ccaac4cf2bc1360853f5608186`. The effective merge delta is the expected 17-file `tests/release/**` fixture/registry slice, GitHub reports CLEAN/MERGEABLE, and all exact-head repository checks are settled green.

**Verdict: `CONTROL_CHANGES_REQUIRED` / not `CODE_MERGE_READY`.** The missing live shared-fixture smoke and journeys 1–9 are accepted post-merge debt and are not the blocker. One prior cleanup-safety ID remains materially open.

### Stable-ID disposition

- **`PR6-CONTROL-001` — CLOSED.** The receipt merge now preserves the first original value per `(table,id,field)` and appends only genuinely new entries. The focused regression executes the real Python helper twice with the same identity and demonstrates that the retry cannot replace the pre-first-call value.

- **`PR6-CONTROL-003` — CLOSED at the requested PID-reuse boundary.** Relay startup records PID, `/proc` start time, and the exact run-scoped script path. Cleanup checks both ownership discriminators before signalling; a mismatch removes only the stale pidfile and sends no signal. A matching process still must terminate or cleanup fails non-green.

- **`PR6-CONTROL-005` — CLOSED.** Process selection now uses the full basename from `/proc/<pid>/exe`, with truncated `comm` plus full `argv[0]` only as an unreadable-exe fallback. This reaches the real 18-character `proliferate-worker` identity and excludes shell wrappers; missing or surviving targets remain non-green.

- **`PR6-CONTROL-002` — BLOCKER, still open: the Stripe intent→acquired handoff is not durable across runner loss.** [`stripeTestClockActor`](https://github.com/proliferate-ai/proliferate/blob/37cd3854e0e551bc0897652e49f7904eeff09fbd/tests/release/src/fixtures/stripe-test-clock.ts#L195-L244) registers provider IDs such as `intent:test_clock:<runTag>`, then stores the returned Stripe IDs only in the local mutable `acquired` object captured by each releaser. But [`ManagedCloudWorld.registerCleanup`](https://github.com/proliferate-ai/proliferate/blob/37cd3854e0e551bc0897652e49f7904eeff09fbd/tests/release/src/worlds/managed-cloud/world.ts#L282-L289) immediately marks the supplied `intent:*` string as the ledger's acquired provider identity; it exposes no later `markAcquired` operation. A process restart therefore loses both the closure and its real ID. The generic [`replayLedger`](https://github.com/proliferate-ai/proliferate/blob/37cd3854e0e551bc0897652e49f7904eeff09fbd/tests/release/src/worlds/local-workspace/cleanup-ledger.ts#L262-L293) also has no shipped `stripe_test_clock` / `stripe_customer` recovery handlers in this slice.

  The new interruption test throws and then invokes the same process's captured closures; it does not reload the persisted ledger or prove hosted-runner-loss recovery. Thus a Stripe create accepted immediately before runner death can still leave only `intent:*` in durable storage and leak the clock/customer—the exact safety failure this ID owns. The corrected `DELETE` endpoints are valid, but they are unreachable from recovered ledger state.

  **Required minimal closure:** use the cleanup ledger's real intent→`markAcquired(realId)` transition and provide replayable kind handlers, or persist a recovery identity whose handler can deterministically locate/delete the run-owned resource after process loss. Add a regression that reloads the ledger with no original closure and reconciles the accepted-before-acquire resource. Under the founder merge-first directive, the alternative is to hard-disable the Stripe clock fixture as explicitly non-green and move its executable cleanup to a follow-up; leaving the unsafe enabled path is not acceptable.

No deploy, provider call, credential/protected-state mutation, or live-qualification claim was reviewed or authorized. A head change stales this verdict.

pablonyx:
## ACK — E2E CONTROL delta re-review (comment 4989466208, head `37cd3854e`)

Acknowledged within one heartbeat. Verdict CONTROL_CHANGES_REQUIRED accepted.

**Closed (no further action):** PR6-CONTROL-001, PR6-CONTROL-003, PR6-CONTROL-005.

**Re-opened, accepted for repair round 4 (only this):**

- **PR6-CONTROL-002** — the intent→acquired handoff lives only in an in-memory closure: `registerCleanup` persists `intent:*` as the final ledger identity with no `markAcquired` transition, and `replayLedger` ships no `stripe_test_clock`/`stripe_customer` recovery handlers — so a runner death right after Stripe accepts the create leaks the clock/customer despite the (now-correct) DELETE endpoints being unreachable from recovered ledger state.

Repair per the required minimal closure: durable intent→`markAcquired(realId)` in the cleanup ledger + replayable kind handlers for Stripe clocks/customers, with a regression that RELOADS the persisted ledger (no original closure) and reconciles an accepted-before-acquire resource. If that cannot land cleanly within this slice's extension contract, the fallback is hard-disabling the Stripe fixture as explicitly non-green and deferring its executable cleanup — the unsafe enabled path will not ship either way.

Scope discipline: no journeys added, no live providers, focused offline proof only; one stable head, then finding→commit→proof and an exact-head CONTROL request.

pablonyx:
## Repair receipt — CONTROL round 4 (PR6-CONTROL-002, durability)

The remaining finding is fixed via CONTROL's preferred path (the ledger supported it — no hard-disable fallback needed). New canonical head: **`a762661694756459a2f4f760818462d81dc9293a`** (single amended commit, force-pushed with lease; prior reviewed head `37cd3854e`). Delta touches only `stripe-test-clock.ts`, its test file, and `world.ts`; the closed 001/003/005 areas are untouched.

### PR6-CONTROL-002 — durable intent→acquired handoff + closure-free replay

Built entirely on the cleanup ledger's EXISTING primitives (`stack.register` → `stack.acquired`, i.e. the real `markAcquired(entryId, providerId)` transition) — no parallel mechanism, no ledger-schema change:

1. **World seam (append-only)** — `world.ts`: optional `registerCleanupIntent(kind, intentRef, release)` returns a `CleanupIntentHandle` whose `markAcquired(realId)` durably replaces the persisted intent ref with the real provider id. The intent (carrying the run-scoped recovery identity) lands on disk BEFORE the caller's create. Existing `registerCleanup` and all its callers are unchanged.
2. **Fixture** — registers the intent (structured recovery ref: deterministic run-scoped clock name / customer run-tag) before `createClock`/`createCustomerOnClock` and calls `markAcquired(tc_…)`/`markAcquired(cus_…)` the instant Stripe returns. PR-2 worlds without the seam fall back to the prior single-shot registration.
3. **Replay handlers** — exported `stripeCleanupReplayHandlers({secretKey, transport})` for `stripe_test_clock`/`stripe_customer`, working from the ledger ENTRY ALONE: a real `tc_`/`cus_` id → `DELETE` by id; an intent-only ref → locate by run-scoped clock name / customer run-tag and delete; not-found → clean reconcile.
4. **Regressions (the decisive ones)** — the test world persists through the REAL `openCleanupLedger` + `ManagedCloudCleanupStack`:
   - runner dies between create-accept and `markAcquired`, ALL closures discarded, `loadCleanupLedger` reloads from disk, `replayLedger` + the handlers locate the clock by its run-scoped name and delete it (`failed === 0`);
   - same for a crash after acquire: the reloaded real-id entry deletes by id with no lookup;
   - happy-path replay asserts no lookup occurs when a real id is present.

### Proof (exact head `a7626616`, focused offline only)

- `npm run typecheck`: clean.
- `npm test`: **706/706 pass** (was 703; +3 runner-loss/replay regressions).
- No on-box Python changed this round; the existing Stripe snippets re-verified compiling.
- No live providers called, no journeys added, no deploy, no scope broadening.

**Requesting exact-head E2E CONTROL review of `a762661694756459a2f4f760818462d81dc9293a`** (mapped CONTROL `019f64fc-0eae-76d2-bc3a-aa4dd32a0c22`) — scope: PR6-CONTROL-002 closure.

pablonyx:
## E2E CONTROL — exact-head re-review

**Custody reviewed:** base/current main `b9b43d215426a40beeeb4741ba05df6ccdf0db68`, head `a762661694756459a2f4f760818462d81dc9293a`, merge ref `58dd578a90141f6f3d5a8944096130de9e7fdcb8`.

**Verdict: CONTROL_CHANGES_REQUIRED / not CODE_MERGE_READY on this head.**

**PR6-CONTROL-002 — BLOCKER, still open at the provider-recovery boundary.**

The intent → acquired handoff is now durable and the ledger-only replay design is materially improved. However, replay can still mark cleanup reconciled while leaking accepted Stripe resources:

1. Test-clock recovery requests only `GET /v1/test_helpers/test_clocks?limit=100`, does not follow `has_more` / `starting_after`, and treats “not in the first page” as clean success. Stripe documents this endpoint as cursor-paginated: https://docs.stripe.com/api/test_clocks/list
2. Customer recovery uses Stripe Search and immediately treats no match as clean success. Stripe explicitly warns that Search is not suitable for read-after-write flows and can lag—normally under a minute, but up to an hour during outages: https://docs.stripe.com/api/customers/search . That is exactly the accepted-by-Stripe / runner-died-before-`markAcquired` window this repair is meant to survive.

Therefore a ledger-only replay can permanently reconcile an intent before the provider object is visible or before the page containing it is visited.

Minimal closure:

- paginate test-clock enumeration until the run-owned clock is found or the collection is truly exhausted;
- never reconcile a customer intent merely because an immediate Search returns empty. Prefer a strongly scoped paginated list (Stripe’s customer list supports `test_clock`: https://docs.stripe.com/api/customers/list), or keep the cleanup intent retryable/non-reconciled across a bounded propagation window;
- add regressions for “owned clock on a later page” and “first customer lookup empty, later lookup finds and deletes,” asserting the ledger is not reconciled on the transient miss.

This does **not** require the nine managed-cloud journeys or the full live fixture smoke before merge; it requires the shared cleanup primitive not to silently leak resources.

At review time GitHub reports MERGEABLE with no failed checks, but exact-head CI is still settling. A head change stales this verdict.

pablonyx:
## ACK — E2E CONTROL exact-head re-review (comment 4989788950, head `a7626616`)

Acknowledged within one heartbeat. Verdict CONTROL_CHANGES_REQUIRED accepted.

**Re-opened, accepted for repair round 5 (only this):**

- **PR6-CONTROL-002** — provider-recovery boundary: (1) test-clock recovery reads only the first `limit=100` page and treats "not on page 1" as clean success (the endpoint is cursor-paginated); (2) customer recovery trusts an immediate Stripe Search empty result, but Search is documented as unsuitable for read-after-write and can lag up to an hour — precisely the accepted-before-`markAcquired` window this repair exists to survive. Either can permanently reconcile an intent while the resource leaks.

Repair per the minimal closure: paginate clock enumeration to true exhaustion (`has_more`/`starting_after`); replace/augment customer Search with a strongly scoped paginated list (customers list supports `test_clock`) and never reconcile an intent on a transient miss — a not-found stays retryable/non-reconciled within a bounded propagation window; regressions for "owned clock on a later page" and "first lookup empty, later lookup finds and deletes," asserting no reconcile on the transient miss.

Noted per the verdict and the founder merge-first directive: the nine journeys and the live fixture smoke are NOT pre-merge requirements; this closes the last safety blocker on the shared cleanup primitive.

pablonyx:
## Repair receipt — CONTROL round 5 (PR6-CONTROL-002, provider-recovery boundary)

Both recovery defects fixed. New canonical head: **`cf67b80d0359a2d65ae551640739b6beae1a971e`** (single amended commit, force-pushed with lease; prior reviewed head `a7626616`). Delta touches only `stripe-test-clock.ts` + its test file; all closed areas untouched.

| Defect | Fix | Proof |
|---|---|---|
| Clock recovery read only page 1 | `findTestClockByName` follows the cursor-paginated list (`limit=100`, then `starting_after=<last id>` while `has_more`) until the run-owned clock is found or the collection is truly exhausted. "Not on the first page" is no longer "not found". | Regression: 100 decoy clocks on page 1 with `has_more`, the owned clock on page 2 → found via an asserted `starting_after=tc_decoy_99` request and deleted. |
| Customer recovery trusted an immediate Search empty result | Search is removed entirely (zero references). Recovery now resolves the CLOCK first (sibling ledger entry's real id, else the exhaustive paginated name lookup), then uses the strongly-scoped paginated `GET /v1/customers?test_clock=<tc_id>` list matched by run-tag metadata. If no clock is recoverable, the customer — created ON the clock, after it — cannot exist; the handler documents that reasoning. | Regression asserts the clock-scoped LIST (never Search) with run-tag matching; existing runner-loss regressions adapted and passing. |
| Empty lookup could permanently reconcile a propagating intent | New bounded propagation window (`STRIPE_INTENT_RECOVERY_WINDOW_MS` = 75 min > Stripe's documented worst-case 1 h Search/propagation lag): an exhaustive-but-empty lookup within the window throws typed `StripeIntentStillPropagatingError`, leaving the entry UNRECONCILED for a later replay pass; only past the window does empty reconcile as truly-never-created. Injected `now` seam for test control. | Regressions: pass 1 in-window empty → `failed: 1`, entry unreconciled, nothing deleted; pass 2 (resource now visible) → found, deleted, reconciled. Past-window empty → clean reconcile. |

### Proof (exact head `cf67b80d0`, focused offline only)

- `npm run typecheck`: clean.
- `npm test`: **710/710 pass** (was 706; +4 recovery regressions).
- No on-box Python changed; no live providers called, no journeys added, no deploy, no scope broadening.

**Requesting exact-head E2E CONTROL review of `cf67b80d0359a2d65ae551640739b6beae1a971e`** (mapped CONTROL `019f64fc-0eae-76d2-bc3a-aa4dd32a0c22`) — scope: PR6-CONTROL-002 closure. Per the founder merge-first directive, this is the last open safety blocker.

pablonyx:
## E2E CONTROL — exact-head re-review

**Custody reviewed:** base/current main `b9b43d215426a40beeeb4741ba05df6ccdf0db68`, head `cf67b80d0359a2d65ae551640739b6beae1a971e`, merge ref `6b7b88bd87a5c3ba0ae59d894d45eaf3b40f5cf1`.

**Verdict: CONTROL_CLEAN / CODE_MERGE_READY on this exact head.**

**PR6-CONTROL-002 — CLOSED.** The provider-recovery delta closes both prior silent-reconcile paths:

- test-clock lookup follows Stripe cursor pagination through `has_more` / `starting_after` until the run-owned clock is found or enumeration is truly exhausted;
- customer recovery no longer uses eventually consistent Search. It resolves the run-owned clock, lists customers scoped to that clock, paginates, and matches the run tag;
- an exhaustive empty lookup inside the bounded propagation window throws `StripeIntentStillPropagatingError`, so `replayLedger` leaves the intent unreconciled/retryable; a later pass can find, delete, and reconcile it;
- focused regressions cover the later-page clock and transient customer miss → later visibility/delete paths.

The reviewed path remains run-scoped and fail-closed: no accepted resource is marked reconciled merely because the first page or first lookup missed it. No provider call, deployment, credential mutation, or journey acceptance is introduced.

GitHub reports MERGEABLE; exact-head ordinary CI is fully settled with no failed or pending checks. The shared-fixture live smoke and managed-cloud journey workstreams 1–9 remain explicit post-merge qualification debt and are not represented as accepted.

A head change stales this verdict.

pablonyx:
## Shepherd merge record

- PR: #1265 — managed-cloud shared fixture layer
- Reviewed exact head: `cf67b80d0359a2d65ae551640739b6beae1a971e`
- Mapped E2E CONTROL: `CONTROL_CLEAN / CODE_MERGE_READY` in comment 4990067849
- Squash merge: `317f3644fb433454ec339a1e031c8c7997a8f570`
- Merge mode: SHA-matched squash, no admin bypass

This records code merge only. The live shared-fixture smoke and managed-cloud journey workstreams 1–9 remain explicit qualification debt. Reconcile the next slice from exact merge `317f3644fb433454ec339a1e031c8c7997a8f570`; do not claim qualification accepted until those proofs run.

PR #1276, opened 2026-07-16
feat(product): ordered repository readiness + truthful availability actions (PR 2)

pablonyx:
Slice 2 of 7 of the Cloud repository access + workspace materialization stack (stacked on #1257; do not merge before it).

## What
- Version-aware `/meta` capability parsing (`parseServerCapabilities`) with an explicit compatibility matrix: v2 exact; v1 `cloudWorkspaces` → legacy projection; absent + official origin → legacy-ready; malformed fails closed.
- Pure 10-gate repository readiness resolver in product-domain (`repos/repo-readiness.ts`) with `RepositoryCapabilityRequirement = local_only | github_repository_access | managed_cloud` and truthful actions — a user is never sent to user OAuth when the operator must configure the deployment.
- `canonicalRepoKey` (case-folded provider/owner/repo, `.git` stripped; branch names never folded) in product-domain `repos/repo-id.ts`.
- Host-truthful Add Repository options; shared repo-group menu model for DOM + native menus.
- Item F: one connected `CloudRepoActionDialogHost` mounted in the app shell, driven by a serializable `CloudRepositoryIntent` in a one-shot zustand store; authorize/install/grant resume via the existing staggered-polling GitHub-App return (no deep link). Includes a fixed continuation gate bug (continuation now fires at gate 9 *or* 10, with a re-entrancy guard).
- Item H: settings vocabulary alignment in `RepoCloudGate` — operator-configuration states have no user-OAuth CTA and offer a `Copy request` clipboard CTA (`github-app-copy.ts`) instead.
- Durable associations use the native install id (`desktop.worker.getInstallId()`), not the telemetry fallback.

## Verification
- product-domain vitest `src/repos/`: 29 passed.
- product-client focused suites (intent, creation, capabilities, repositories, sidebar/RepoGroup/native-menu/command): 264 passed across 62 files, plus the earlier full runs (574 + 3047 + 185) on the bulk commit.
- Typecheck clean in product-domain / product-ui (built) / product-client. `git diff --check` clean.

## Notes for review
- `package.json` diffs are new subpath exports required here (`repo-readiness`, `CloudRepoPickerBlocker`).
- Spec: Vault "Repository Availability and Access UX - PR 2" (frozen rev with 8-point reconciliation).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1257, opened 2026-07-16
feat(server): independent GitHub repository access and managed-Cloud capabilities (v2)

pablonyx:
## Summary

Implements the frozen **Managed Cloud Capability And GitHub Authority — PR 1** contract (promoted in #1255, which this PR is stacked on and must land first).

- **Capability contract v2**: `/meta` now reports independent `githubRepositoryAccess` and `managedCloud` operator capabilities (`disabled` / `operator_configuration_required` / `ready`). `cloudWorkspaces` remains as a compatibility projection, true only when the whole managed-Cloud path is operable — old clients fail closed on E2B-ready/App-incomplete deployments instead of being invited into flows that fail on `require_github_cloud_repo_authority`.
- **Shared completeness predicate**: `Settings.github_app_configured` / `github_app_partially_configured` cover the five App runtime requirement groups (App id, OAuth client id/secret, webhook secret, private key in one form). Public responses expose only aggregate status + safe display name; a leak-guard test asserts no secret or field-level presence appears on the wire.
- **Repairable authority statuses**: `github_repo_access_required` → `missing_user_repo_access` with **no** action (reauthorizing the App cannot grant a GitHub account repository access); `github_app_not_configured` → new `operator_configuration_required` status with no user action.
- **preflight.sh truth**: removes the false claim that E2B alone gates Cloud workspaces; mirrors the Python predicate and warns explicitly on E2B-ready/App-incomplete.
- **Generated artifacts**: `server/openapi.json` and `cloud/sdk/src/generated/openapi.ts` regenerated via `make cloud-client-generate`; no hand edits.

No UI, no database schema, no version-aware client branching (PR 2 owns those).

## Verification

- `DEBUG=true uv run pytest -q tests/unit/test_meta_endpoint.py tests/unit/test_github_app_service.py tests/unit/test_github_app_callback_return.py` — 43 passed (full E2B × App matrix incl. App-ready/E2B-disabled, both partials, display-name fallback, secret-leak guard, all authority mappings).
- `make cloud-client-generate` — clean regeneration.
- `pnpm run typecheck` in `cloud/sdk` and `cloud/sdk-react` — pass.
- `ruff check` + `ruff format --check` on touched files — clean.
- `python3 scripts/check_docs.py` — pass.

## Review sequencing

Stacked on #1255 (docs promotion). Per the frozen handoff: the docs PR lands first, and any review changes to the promoted contract are reconciled here before this merges. PR 2 (client UX) is intentionally not started.

[...]

PR #1246, opened 2026-07-15
feat(anyharness): isolated workflow workspace placement (slice 3)

pablonyx:
## Outcome

AnyHarness can deterministically materialize exactly one isolated, visible, retained workspace for a workflow run, then expose the resolved placement without accepting or executing the run.

Frozen contract: `Vault/Workspace/Workflows/1 - Core V1 Implementation/First Supported Product PR Stack/3 - Isolated Workflow Workspace Placement.md` (FROZEN RC1.0).

## What changed

- `PUT/GET /v1/workflow-run-workspaces/{runId}` with strict schema-v1 decoding, exact replay, and conflict semantics.
- Deterministic placement under `<managed-worktrees-root>/workflows/<runId>`:
  - `scratch`: empty Git repository on `main`, one empty commit, non-personal local identity, no remote.
  - `repositoryWorktree`: `workflow/<runId>` at an immutable commit resolved before effects, with no suffix/reset/checkout repair behavior.
- SQLite materialization ledger (`0064`) with canonical requested/resolved JSON, durable status/correlation, crash-safe ensure/adopt, and terminal failure recording.
- Exact on-disk adoption proof: canonical/symlink-safe path, expected Git common-dir, branch, HEAD OID, cleanliness, scratch identity, and workflow creator provenance.
- Workflow creator context and retention exclusion; no cleanup API or automatic deletion.
- Schema-v2 run preflight rejects a missing/mismatched ready workspace, while the authoritative run-accept transaction repeats that check before inserting. The reciprocal placement-accept transaction rejects a new materialization after any run claimed the ID. Both use the shared serialized SQLite transaction seam, so either race ordering produces exactly one durable claim and no workspace/session/prompt/turn side effect from the loser; schema v1 remains unchanged.
- Strict discriminated-union OpenAPI/SDK contract: both `oneOf` child objects explicitly set `additionalProperties: false`; generated artifacts and the full checked-in document are regression-pinned.
- Secret-safe durable failures: raw Git output and all free-form placement mismatch payloads are discarded; stored detail is fixed and bounded.
- Shared exact-ref helpers are reused rather than duplicated; `rev-parse` is option-fenced and regression-tested with a dash-prefixed ref.

## Exact custody

- Base: `7500ece0681ae776e0c17389b1992a779f268f31` — current main, containing the merged Workflow Session Mutation Admission predecessor.
- Head: `267503c079f9c81139a569d9cd0603d98841ccf3`.
[...]

PR #1196, opened 2026-07-14
feat(anyharness): truthful workflow cancellation and versioning (run control)

pablonyx:
Implements the frozen specification **`specs/codebase/systems/product/workflows/run-control.md`** (revision **RC1.0**; relocated from `specs/codebase/features/workflow-run-control.md` by `main`'s docs restructure) — stack slice 2a, "Truthful Workflow Cancellation and Versioning".

- **Base**: `main` (contains the repaired #1185 merge `47469e491`) · **Head**: `a24180e34` (rebased onto fully shape-clean `main` after #1209/#1218)
- **Founder freeze**: 2026-07-14, all six decisions approved (decisions 1–5 in the freeze session; decision 6 — exact-active-turn-only live cancel, no queue-removal path — ruled during promotion; freeze reconfirmed via the `PR1196-SPEC-01` founder ruling). The Vault slice records the same provenance.

## Behavior

`POST /v1/workflow-runs/{runId}/cancel` records durable cancellation intent and reports only what is known: pending step → atomic `cancelled` (proof no prompt was sent); running step with a stored turn → truthful `running + cancelRequestedAt` plus a best-effort **exact-active-turn** live cancel (a stale stored turn ID can never cancel newer foreign work); null-turn/queued/lost-ack → intent only. Both v1 and v2 responses gain required `stateVersion`, optional-omitted `cancelRequestedAt`/`interruptionCode`, and the widened `cancelled`/`interrupted` vocabularies (scoped supersession of the invocations spec's v1-freeze clause, §3.4). Every externally visible run+step transaction increments the version exactly once; replay/no-ops/duplicate callbacks increment nothing. `WorkflowRunGates` serializes acceptance, execution CAS boundaries, the completion extension's terminal CAS, and cancellation per run, so cancel and dispatch can never cross unobservably. Custom FK migration `0062` rebuilds both tables with cross-column checks and strict legacy-pair validation (aborts rather than guessing history; `failed/session_turn_cancelled → cancelled`, `failed/runtime_restarted → interrupted`). Startup fencing writes `interrupted/runtime_restarted` with one increment and still aborts AppState on failure. The execution task lives in `execution.rs` and consumes #1185's extracted dispatch decision (`dispatch.rs`), terminalizing verifiable dispatch failures under the same gate hold (§6.2).

## Verification (head `49760f91f`)

- `cargo test -p anyharness-lib`: **1088 passed / 0 failed / 1 ignored** · `anyharness-contract`: **41 passed**
[...]

PR #1278, opened 2026-07-16
feat(anyharness): local repository roots and exact-ref workspace materialization (PR 3)

pablonyx:
Slice 3 of 7 of the Cloud repository access + workspace materialization stack (stacked on #1257; sibling of #1276/#1277; do not merge before #1257).

## What
- Two new AnyHarness endpoints:
  - `POST /v1/repo-roots/materializations` — idempotent repository acquisition (`clone_or_adopt`; outcome `cloned|adopted|reused`). Managed clones register `RepoRootKind::Managed` under `managed_worktrees_root`. Credentialed clone URLs are rejected at the contract (`400 REPOSITORY_AUTH_REQUIRED`) — local clones use the local Git credential chain, never injected tokens.
  - `POST /v1/repo-roots/{repo_root_id}/workspace-materializations` — workspace materialization at an exact branch + SHA (outcome `created|adopted|reused`). Never resets/merges/rebases/stashes/detaches to repair a mismatch; mismatches return typed 409s.
- `local_materialization_operation` SQLite idempotency ledger (migration 0061) keyed by caller-supplied operation id: crash-safe retry by reusing the same id.
- Typed RFC7807 error codes (REPOSITORY_AUTH_REQUIRED, DESTINATION_CONFLICT, WORKSPACE_DIRTY, WORKSPACE_BUSY, MATERIALIZATION_OPERATION_CONFLICT, …).
- Generic exact-ref reuse-validation extracted from mobility.rs into `runtime/exact_ref.rs`, consumed by both mobility and the new endpoint (`create_worktree_at_ref` reused; `publish_created_branch` stays mobility-only).
- SDK: regenerated openapi.json/openapi.ts via the owning generator; hand-authored `RepoRootsClient.materialize()`/`.materializeWorkspaceAtRef()`, type re-exports, and two sdk-react mutation hooks with client unit tests.

## Coexistence with the Workflow stack (#1246)
- SQL number 0061 used here; #1246 reserves 0064 — no collision.
- No `scratch.rs`, nothing under `domains/workflows/`, `worktrees.rs` untouched; shared files (operations/mod.rs, service.rs, openapi.rs, router.rs) edited additively only.

## Verification
- `cargo check`/`clippy` on anyharness-lib + anyharness-contract: clean for all touched files (one pre-existing clippy-1.95 `never_loop` artifact in sessions/actor, untouched, from #1196).
- Tests: 26/26 materialization domain (unit + git integration), 28/28 runtime/exact_ref/migration (incl. mobility-coexistence proofs). rustfmt + `git diff --check` clean.
- Not run: SDK vitest/tsc (node deps not installed in this worktree; TS authored against freshly generated types) — will be covered by CI's Build SDK job.

## Notes for review
- Spec: Vault "AnyHarness Local Repository and Workspace Materialization - PR 3" (frozen rev with 9-point reconciliation).


pablonyx:
**Dedicated agent review @ f4f39115f — verdict: needs repair (1 blocker).** (Review ran against f4f39115f; 98cfcd70a since amended only the schema snapshot + line allowlist, so all findings still apply.)

Verified good: no repair-on-mismatch on the exact-ref path, credentialed-URL rejection before any work with no token injection, behavior-preserving mobility extraction, managed-root confinement, no #1246 collisions, SDK camelCase fidelity with tests.

Findings:
1. **[blocker] Concurrent same-operation-id calls can double-execute.** The running-state guard only rejects the caller that loses the INSERT race; a second caller that *finds* the running row gets `Admission::Retry` (the crash-recovery path) and runs concurrently with the first — two clones / two worktrees for one operation. The code can't distinguish a crashed running row from a live one. Fix: staleness-gate retry-adoption (only adopt running rows older than a crash threshold; younger → retryable MATERIALIZATION_OPERATION_CONFLICT) or a per-operation-id in-process lock, plus a real concurrent integration test.
2. **[should-fix] Crash-after-clone recovery downgrades a managed root to `external`** (retry finds the cloned dir → adopt path → kind "external", outcome "adopted"), losing clone provenance that PR 4/5 may key cleanup ownership on.
3. **[should-fix] `request_hash` omits `clone_url`** — same-id retries can silently swap the fetch source; also hashes destination pre-canonicalization (over-strict in the other direction).
4. [nit] `map_exact_ref_error` classifies by loose substrings ("is at ") — prefer typed errors out of exact_ref.rs.
5. [nit] `replay_workspace` SHA fallback can return a branch name in the SHA field (unreachable today).
6. [nit] `DESTINATION_OUTSIDE_ALLOWED_ROOT` never fires for a true escape (containment holds by construction; code is dead).

Test gaps: real concurrency test, repo-root and workspace crash-recovery tests, WORKSPACE_BUSY test.

pablonyx:
Founder review received (head 98cfcd70a): PR3-CONCURRENCY-01 (known race), PR3-EMPTY-02 (existing empty destinations can't be cloned — contract requires it), PR3-HASH-03 (hashes omit cloneUrl / preferredWorkspaceName), PR3-GIT-INPUT-04 (SECURITY: clone URL not validated against requested identity before git runs, no end-of-options separators, unvalidated headSha/branch), PR3-REPLAY-05 (corrupt-record replay returns branch name in SHA field — must fail closed). All folded into the in-flight repair; do not review until the new head is posted.

pablonyx:
**Adversarial re-review @ exact head 869c5e9fa — verdict: ready for Pablo. All six founder findings CONFIRMED-FIXED.**

- PR3-CONCURRENCY-01: keyed operation lock acquired before ledger reconcile on both endpoints, RAII guard held across the blocking work, completed-replay bypass is a pure read, crashed-row adoption only reachable with no live holder; two-thread race test asserts exactly-one-proceeds/one-conflict. (Note: test races the admission arbiter — the real serialization point — not the full clone path.)
- PR3-EMPTY-02: empty dir clones; hidden-file dir → DESTINATION_NOT_EMPTY; bogus .git fails closed; symlinks canonicalized.
- PR3-HASH-03: all behavior-changing fields hashed on both endpoints; canonicalization is retry-stable (nearest-existing-ancestor form is identical pre/post-clone).
- PR3-GIT-INPUT-04: validators run before any git; bypass attempts fail closed (`github.com@evil.com`, ssh://, unicode/case/trailing-slash tricks, option injection); downstream worktree git invocations receive only validated values; `--` separator on clone.
- PR3-REPLAY-05: missing observed_head_sha → MATERIALIZATION_FAILED, never a branch name.
- Managed-kind recovery: `intended_kind` stamped pre-clone inside the net-new 0061 CREATE TABLE (no ALTER needed, no deployed ledgers); crash-before-stamp falls through to clone correctly.
- Regression sweep clean: no-repair-on-mismatch, credential rejection, managed-root confinement, single shared exact-ref implementation, #1246 boundaries all intact.

New (low, non-blocking): crash mid-`git clone` leaves a partial dir the recover path won't clean up — every retry fails closed with MATERIALIZATION_FAILED until manual removal. Robustness follow-up, not a merge blocker.

Disclosed test gap retained: no WORKSPACE_BUSY test (busy derives from live session handles; needs a live-session harness).

pablonyx:
## CONTROL review - exact head

**Custody reviewed:** base `4e3af77aad6b1f05827ee03b07e296f58168f813`; head `869c5e9fa39e9d900c5aa881311205873582f3a5`; GitHub merge ref `32d3e677f01c662d9a38a8bf543251423b869167`.

**Verdict: CONTROL_BLOCKED / CODE_NOT_MERGE_READY.** All ordinary exact-head checks are settled green and GitHub reports CLEAN/MERGEABLE, but the frozen PR 3 contract is not yet met. Independently, PR #1257 remains open, so stack order also forbids merging #1278.

### Findings

1. **[Blocker] PR3-CRASH-06 - crash adoption is broken for the normal omitted-`destinationId` workspace request.** The service begins the ledger operation and calls the runtime without persisting the auto-selected destination ([service.rs L257-L285](https://github.com/proliferate-ai/proliferate/blob/869c5e9fa39e9d900c5aa881311205873582f3a5/anyharness/crates/anyharness-lib/src/domains/materialization/service.rs#L257-L285)); the store first writes `destination_path` only when the entire operation completes ([store.rs L104-L127](https://github.com/proliferate-ai/proliferate/blob/869c5e9fa39e9d900c5aa881311205873582f3a5/anyharness/crates/anyharness-lib/src/domains/materialization/store.rs#L104-L127)). With no explicit destination, allocation only selects a path that does not exist, while reuse scans only registered workspace rows ([exact_ref.rs L111-L149](https://github.com/proliferate-ai/proliferate/blob/869c5e9fa39e9d900c5aa881311205873582f3a5/anyharness/crates/anyharness-lib/src/domains/workspaces/runtime/exact_ref.rs#L111-L149)). Git creates the worktree before the workspace row is inserted ([exact_ref.rs L295-L323](https://github.com/proliferate-ai/proliferate/blob/869c5e9fa39e9d900c5aa881311205873582f3a5/anyharness/crates/anyharness-lib/src/domains/workspaces/runtime/exact_ref.rs#L295-L323)). A crash between those steps leaves an unregistered worktree at the unsuffixed path; retry skips it, chooses `-2`, and never inspects/adopts the original checkout. This directly violates the frozen crash-after-filesystem adoption guarantee. Persist the chosen path before Git mutation (or otherwise make it operation-recoverable), adopt only after all ref/safety checks, and add a crash-at-that-boundary integration test with omitted `destinationId`.

2. **[Blocker] PR3-CREDENTIAL-07 - adopting an existing checkout can expose its credential-bearing origin in public responses.** The request URL is checked, but adoption validates the checkout's existing origin ([acquire.rs L117-L129](https://github.com/proliferate-ai/proliferate/blob/869c5e9fa39e9d900c5aa881311205873582f3a5/anyharness/crates/anyharness-lib/src/domains/materialization/acquire.rs#L117-L129)). Remote parsing ignores HTTPS userinfo for identity purposes, registration then stores the raw `remote.origin.url` ([identity.rs L167-L192](https://github.com/proliferate-ai/proliferate/blob/869c5e9fa39e9d900c5aa881311205873582f3a5/anyharness/crates/anyharness-lib/src/domains/workspaces/runtime/identity.rs#L167-L192)), and every repo-root response mapper returns it unchanged ([repo_roots.rs L415-L430](https://github.com/proliferate-ai/proliferate/blob/869c5e9fa39e9d900c5aa881311205873582f3a5/anyharness/crates/anyharness-lib/src/api/http/repo_roots.rs#L415-L430)). Thus a safe request for `https://github.com/o/r` can adopt a checkout whose origin is `https://user:token@github.com/o/r.git` and echo the token through materialize/list/get. The frozen security rule expressly forbids credential-bearing HTTP clone URLs in public payloads. Sanitize, omit, or reject such stored remotes at the response boundary and prove adoption plus list/get.

3. **[Material] PR3-REPLAY-06 - corrupt persisted `observed_head_sha` still replays as success.** Replay now rejects `NULL`, but it does not validate the persisted value despite the comment claiming it is a real SHA ([service.rs L341-L367](https://github.com/proliferate-ai/proliferate/blob/869c5e9fa39e9d900c5aa881311205873582f3a5/anyharness/crates/anyharness-lib/src/domains/materialization/service.rs#L341-L367)). A completed row containing `not-a-sha` is returned successfully, contrary to accepted PR3-REPLAY-05's corrupt-or-missing fail-closed requirement. Validate replayed values with the same full-SHA rule and test missing plus malformed ledger data.

4. **[Material] PR3-GIT-INPUT-06 - branch validation still admits ref names Git rejects, so validation is not actually complete before Git effects.** The custom validator ([identity.rs L102-L141](https://github.com/proliferate-ai/proliferate/blob/869c5e9fa39e9d900c5aa881311205873582f3a5/anyharness/crates/anyharness-lib/src/domains/materialization/identity.rs#L102-L141)) accepts examples such as `.foo`, `foo/.bar`, `HEAD`, `foo.lock/bar`, and `foo/bar.lock`; `git check-ref-format --branch` rejects all five. These values then reach a path that prunes/fetches before later Git rejection ([worktrees.rs L117-L145](https://github.com/proliferate-ai/proliferate/blob/869c5e9fa39e9d900c5aa881311205873582f3a5/anyharness/crates/anyharness-lib/src/adapters/git/operations/worktrees.rs#L117-L145)). Option injection is blocked, but the accepted pre-Git ref-validation repair is incomplete. Match Git branch semantics and add the omitted cases.

5. **[Material evidence gap] PR3-BUSY-08 - the frozen busy-destination acceptance case remains unproved.** The implementation has a busy check, but this exact head contains no `WORKSPACE_BUSY` behavioral test; the implementer comment explicitly discloses that gap. The frozen test contract requires wrong/dirty/**busy** rejection. A focused service/integration seam is sufficient; a heavyweight live external harness is not inherently required.

### Confirmed / classified

I independently confirmed the keyed same-operation exclusion, existing-empty-directory behavior, complete request hash fields, pre-effect rejection of credential-bearing request URLs, managed-kind clone recovery, managed-root confinement, no repair-on-mismatch, mobility-policy separation, additive migration/generated SDK boundaries, and the exact-head green CI evidence.

The disclosed partial-directory-after-interrupted-clone behavior is a **non-blocking robustness follow-up** under the current frozen wording: it fails closed and requires manual cleanup rather than deleting an unproven path. The missing busy proof is not the same category because it is an explicit frozen acceptance test.

Custody note: the source head is not a descendant of the declared stacked base, but the inspected merge ref has the exact base and head as parents and retains the predecessor's `server/tests/integration/cloud_api_helpers.py` change; the effective merge delta does not delete or modify `server/**`. This does not remove the requirement to land #1257 first.

PR #1337, opened 2026-07-17
fix(product-client): prevent replaced sessions from resurfacing

### LINKED ISSUE #1333: Empty-session replace-in-place can orphan a runtime session when the replaced session's creation is still in flight (reappears as extra tab after reload)
## Summary

An empty session that is replaced in place via `beginEmptySessionReplacement` can be **orphaned on the runtime** when the replaced session's own creation coroutine is still in flight — the orphan is hidden from the live UI but reappears as a spurious extra tab after a page reload (which re-fetches the authoritative session list).

## Where

- `apps/packages/product-client/src/hooks/sessions/workflows/use-empty-session-replacement-cleanup.ts`
  - `beginEmptySessionReplacement` captures `materializedSessionId = record.materializedSessionId` once, synchronously, at transaction-begin (line ~73).
  - At `commit()` (line ~140) if that captured value was `null` it takes the cheap path (`finalizeRetirement()` only) and **never issues a runtime dismiss** — `finalizeRetirement` (line ~112) only evicts the client-side query-cache entry.
- `apps/packages/product-client/src/hooks/sessions/workflows/session-creation-materialization.ts`
  - The empty session's creation coroutine runs its **last** `discardIfSuperseded` checkpoint at line ~306, then unconditionally commits the real backend session id (`materializeSessionRecord(...)`, line ~334) with no further supersession check.

## Race

If `beginEmptySessionReplacement(clientSessionId, …)` runs while the replaced session's own materialization is between its last checkpoint (~306) and its final commit (~334):
1. the replacement reads `record.materializedSessionId` as still `null` → decides "nothing to dismiss";
2. `supersedeInFlightSessionCreation` sets the flag, but the coroutine has already passed its last checkpoint, so it commits fully and the session gets a **real backend id that is never dismissed**;
3. `finalizeRetirement()` hides only the local cache entry, so the live UI shows the correct tab count;
4. on reload the authoritative session list still contains the never-dismissed session → it reappears as an extra tab.

## Evidence

- Actions run `29575928457` (release-e2e.yml `local-functional`, head e0bbd2aad), cell **T3-SESSION-1/local/harness=claude**: `reloadAndVerifyTabs: tab order not preserved (wanted 2 tabs, saw 3)`. Screenshot shows a `[Claude, Chat, Claude]` tab strip — the middle "Chat" tab is the orphaned empty codex session.

## Proposed product fix (needs owner review — not attempted here)

Either:
- (a) in `commit()` re-derive the materialized id from authoritative state instead of relying solely on the value captured at `begin()` (note `removeSessionRecord(sessionId)` already ran synchronously, so `getSessionRecord` is not the source — likely needs the workspace session list / a pending-materialization registry); or
- (b) add a final `discardIfSuperseded` checkpoint in `session-creation-materialization.ts` after `materializeSessionRecord` (~334), so a session superseded during its own tail still self-dismisses.

## Test-side handling (landed)

The T3-SESSION-1 collector (tests/release/src/scenarios/local/config-session.ts `switchHarnessEmptyChat`) now waits for the empty tab to fully reconcile (materialize to a server id) BEFORE the empty-chat harness switch, so the replacement dismisses a known materialized session rather than racing an in-flight creation. This is a deterministic workaround in the test; the underlying product race above is still open. A materialized-but-empty session is still empty, so the proof's meaning is unchanged.

Classification: product-defect (concurrency/TOCTOU), reachable from current main. Cleanup: N/A (no fixtures leaked; the orphan is a client/runtime session cleaned up with the world teardown).

pablonyx:
## Summary

- linearize final ProductClient session publication against empty-session replacement
- retire and durably suppress a runtime that completes after its projected shell loses ownership
- replace a fresh query-only authoritative empty tab even when post-reload activation has not installed its local record yet
- prevent a superseded delayed tab activation from publishing or reactivating the retired session
- preserve rollback, compatible-session reuse, non-empty sessions, and honest retention when cleanup cannot be confirmed
- add deterministic delayed-creation, cleanup failure, reload/list reconciliation, re-armed replacement, and rapid post-reload switch regressions
- extract publication side effects into a focused workflow module and remove the stale materializer size allowance

## Evidence

Release E2E run 29602686092 captured the post-reload failure as three tabs: runtime Claude, visibly empty runtime Codex, and a new transient Claude. The new regression reproduces the ownership window with a fresh authoritative session summary but no local record, then proves replacement dismissal/list suppression while a separate delayed-selection regression proves late activation cannot recreate the old slot.

## Readiness

- [x] I followed the pull-request procedure for scope, title, labels, body, and readiness.
- [x] Rebased onto `db656132ce9f084c6b029ee8e8e7058e7556ee11` (#1340 on #1336).

## Verification

- ProductClient TypeScript: `node_modules/.bin/tsc -p tsconfig.json --noEmit`
- ProductClient overlap-focused lifecycle/default suite: 9 files, 49 tests passed
- #1340 release termination fixture: 1 test passed
- Repository structure strict check: zero violations
- `git diff --check`
- independent concurrency, diff, reload-race, and #1336 overlap reviews: no actionable findings

Fixes #1333

### DIFF
--- a/apps/packages/product-client/src/hooks/sessions/workflows/session-creation-materialization-cleanup.test.ts
+++ b/apps/packages/product-client/src/hooks/sessions/workflows/session-creation-materialization-cleanup.test.ts
@@ -1,6 +1,8 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import {
   committedReplacedSessionTombstonesForWorkspace,
+  filterReplacedSessionTombstones,
+  hydrateCommittedReplacedSessionTombstones,
   isReplacedSessionTombstoned,
   resetReplacedSessionTombstonesForTests,
 } from "#product/hooks/sessions/workflows/session-replacement-tombstones";
@@ -9,6 +11,11 @@ import {
 } from "#product/hooks/sessions/workflows/session-replacement-dismissals";
 import { scheduleCreatedRuntimeSessionCleanup } from "#product/hooks/sessions/workflows/session-created-runtime-cleanup";
 import { materializeSessionCreation } from "#product/hooks/sessions/workflows/session-creation-materialization";
+import {
+  beginEmptySessionReplacement,
+  type EmptySessionReplacementTransaction,
+} from "#product/hooks/sessions/workflows/use-empty-session-replacement-cleanup";
+import { reconcileReplacedSessionTombstones } from "#product/hooks/access/anyharness/workspaces/use-workspace-bootstrap-cache";
 import {
   createEmptySessionRecord,
   getSessionRecord,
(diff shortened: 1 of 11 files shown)


pablonyx:
@greptileai Please review the current head 111322cac5088bf6a4e645b14a9fe5032a5b9a5f. The prior checkpoint-count concern is now documented; please also assess the post-reload authoritative empty-session replacement and delayed-activation regressions added after your earlier review.

PR #1344, opened 2026-07-17
fix(release): fail closed on leaked runner handles

pablonyx:
## Outcome

Prevent a future leaked Node handle from monopolizing the single Release E2E concurrency group after the runner has already written its aggregate report.

## Change

- Arm an unref'd 10-second watchdog only after `runReleaseCommand` returns.
- Healthy runs exit naturally; the watchdog does not keep Node alive.
- If another handle keeps the process alive, emit bounded active-resource type diagnostics and force infrastructure exit 2.
- Do not rewrite the already-persisted aggregate or turn red cells green.

## Evidence

- Reproduces the failure mode from Release E2E run 29602686092, which wrote its final report and then occupied the job until the 60-minute timeout.
- Focused subprocess regressions prove a leaked HTTP server exits 2 within the guard window and a healthy process exits 0 without waiting for the watchdog.
- Focused tests: 2/2 green.
- Local workspace-wide typecheck was unavailable because the current lockfile's newly published PostHog entries are rejected by the machine's minimum-release-age policy; exact-head Actions owns the normal typecheck/CI proof.

## Scope

Test infrastructure only. No scenario verdict changes, provider calls, product code, credentials, deploy, or Release E2E dispatch.

### DIFF
--- a/tests/release/src/cli/exit-watchdog.fixture.ts
+++ b/tests/release/src/cli/exit-watchdog.fixture.ts
@@ -1,0 +1,18 @@
+import { createServer } from "node:http";
+
+import { armPostReportExitWatchdog } from "./exit-watchdog.js";
+
+const mode = process.argv[2];
+if (mode === "leak") {
+  const server = createServer((_request, response) => response.end("ok"));
+  await new Promise<void>((resolve) => server.listen(0, "127.0.0.1", resolve));
+  console.log("Combined report written: fixture-report.json");
+  process.exitCode = 0;
+  armPostReportExitWatchdog({ graceMs: 75 });
+} else if (mode === "clean") {
+  console.log("Combined report written: fixture-report.json");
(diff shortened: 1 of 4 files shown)


pablonyx:
E2E CONTROL exact-head verdict — base `3c617a5111eb437b91c8f6f4ceac365a03041680`, head `4ce6d576ca705c7f3b8ac27b53b6367b162a7e59`: **CONTROL_CLEAN / CODE_MERGE_READY**. The watchdog is armed only after the runner returns, is unref'd so healthy processes exit naturally, preserves the already-written aggregate, emits bounded resource-type diagnostics only, and forces non-green exit 2 when another handle survives the grace period. Focused subprocess proof covers both leaked-handle exit 2 and clean natural exit 0. Exact-head ordinary CI is fully settled green; no live qualification, provider, credential, or product behavior is changed.

PR #1349, opened 2026-07-17
perf(transcript): batch high-volume stream reduction

pablonyx:
## Summary

- reduce each accepted live-stream frame and history-tail prefix through one copy-on-write transcript context
- clone item/turn maps and each touched streaming item at most once per batch while preserving the immutable reducer API
- cover a synthetic 200-delta reasoning item, order-sensitive updates, empty batches, history tails, and prior-state/reference stability
- keep the new coordinator and regression suite in focused sibling files, lowering the existing reducer's max-lines allowance

The stream was already frame-batched, but each envelope still invoked the single-event reducer independently. A delta-heavy frame therefore recopied the entire item map and the same growing reasoning item for every delta. This change moves the batch boundary into the SDK reducer and keeps duplicate/gap handling unchanged in ProductClient.

For users, high-volume reasoning transcripts now do bounded collection/object copy work per delivery frame instead of per delta.

This is a performance-only scope. It does not change goal lifecycle, timer meaning, total event retention, or transcript virtualization policy.

## Readiness

- [x] I followed the [pull-request procedure](https://github.com/proliferate-ai/proliferate/blob/main/specs/developing/process/pull-requests.md) for scope, title, labels, body, and readiness.

## Verification

- `anyharness/sdk/node_modules/.bin/tsc -p anyharness/sdk/tsconfig.json --noEmit`
- `anyharness/sdk/node_modules/.bin/tsc -p anyharness/sdk/tsconfig.type-tests.json --noEmit`
- `anyharness/sdk/node_modules/.bin/vitest run anyharness/sdk/src` (104 tests)
- `vitest run src/lib/domain/sessions/stream/stream-state.test.ts` from ProductClient (13 tests)
- deterministic Proxy instrumentation proves one item-map copy and one reasoning-item clone for 200 deltas in both live and history-tail paths
- independent review passed prior-state immutability, intra-batch ordering, duplicate/gap prefix behavior, history/live equivalence, and bounded-copy checks

Residual risk: copy work is bounded per accepted batch, not across total retained history. Event retention, transcript-wide selectors/rendering, and one-envelope frames remain unchanged.

### DIFF
--- a/anyharness/sdk/src/index.ts
+++ b/anyharness/sdk/src/index.ts
@@ -433,6 +433,7 @@ export type {
 export {
   createTranscriptState,
   reduceEvent,
+  reduceEventBatch,
   reduceEvents,
   selectPendingApprovalInteraction,
   selectPendingMcpElicitationInteraction,
--- a/anyharness/sdk/src/reducer/__tests__/transcript-batch.test.ts
+++ b/anyharness/sdk/src/reducer/__tests__/transcript-batch.test.ts
@@ -1,0 +1,184 @@
+import { describe, expect, it } from "vitest";
+import {
+  createTranscriptState,
+  reduceEvent,
+  reduceEventBatch,
+  reduceEvents,
+} from "../../index.js";
+import type { SessionEventEnvelope, ThoughtItem } from "../../index.js";
+
+describe("transcript batch reducer", () => {
+  it("treats an empty batch as a referential no-op", () => {
+    const before = createTranscriptState("session-1");
+
+    expect(reduceEventBatch(before, [])).toBe(before);
+  });
+
+  it("avoids per-delta copies for a high-volume reasoning item", () => {
+    const before = reduceEvents([
+      turnStarted(1),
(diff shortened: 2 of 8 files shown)


pablonyx:
@greptileai Please review the current head 3af833f5a0230eab18b0cdbad354c23860631775. Focus on prior-state immutability, intra-batch event ordering, duplicate/gap prefix handling, live/history-tail equivalence, and whether item-map and same-item copying is bounded once per accepted batch. Please also check that the max-lines repair lowers rather than weakens the repository limit.

pablonyx:
@greptileai Please re-review exact head `01b9da27ffa43b5f31c8d664e6fd7aabcbd9e96c`. The prior P2 was accepted: the redundant `item_delta` `setItem` call was removed after confirming `ensureMutableKnownItem` already installs the map-owned clone, and the reducer max-lines cap was lowered from 1529 to 1528. Please confirm the current head has no remaining correctness or bounded-copy findings.

PR #1350, opened 2026-07-17
fix(release): preserve local qualification truth

pablonyx:
## Outcome

Repairs three ownership-correct failures observed in Release E2E run 29602686092 without converting product failures into green results.

- maps local-functional BYOK credentials from the real Qualification secret names
- prevents AUTHROUTE from accepting a stale pre-send session
- proves actor setup visibility before one password-login attempt and records only redacted diagnostics
- leaves SESSION/#1333 product-red and untouched

## Proof

- 43/43 focused release tests
- release TypeScript typecheck
- diff/max-lines checks
- no live provider, Docker, or Rust work

## Acceptance debt

This does not claim the local Tier-3 lane is qualification-accepted. A strict exact-head Release E2E run follows after code review and shared lane custody are clean.

### DIFF
--- a/.github/workflows/release-e2e.yml
+++ b/.github/workflows/release-e2e.yml
@@ -605,16 +605,18 @@ jobs:
           AGENT_GATEWAY_LITELLM_PUBLIC_BASE_URL: ${{ vars.AGENT_GATEWAY_LITELLM_PUBLIC_BASE_URL }}
           AGENT_GATEWAY_LITELLM_MASTER_KEY: ${{ secrets.AGENT_GATEWAY_LITELLM_MASTER_KEY }}
           # BYOK provider keys for the LOCAL-3 user-key route + LOCAL-6 route
-          # change (env-manifest.ts). Absent -> those cells report blocked,
-          # never fail the run (cursor has no BYOK var by design — no account
-          # key here, per BRIEF §5).
-          # The env var NAMES the runner reads are the provisioned _API_KEY-suffixed
-          # names (env-manifest.ts / BYOK_ENV_BY_HARNESS); the GitHub secret names
-          # on the right are unchanged.
-          RELEASE_E2E_BYOK_ANTHROPIC_A_API_KEY: ${{ secrets.RELEASE_E2E_BYOK_ANTHROPIC_A }}
-          RELEASE_E2E_BYOK_ANTHROPIC_B_API_KEY: ${{ secrets.RELEASE_E2E_BYOK_ANTHROPIC_B }}
-          RELEASE_E2E_BYOK_OPENAI_API_KEY: ${{ secrets.RELEASE_E2E_BYOK_OPENAI }}
-          RELEASE_E2E_BYOK_XAI_API_KEY: ${{ secrets.RELEASE_E2E_BYOK_XAI }}
+          # change (env-manifest.ts). In strict mode an absent required key
+          # blocks its selected cells, cancels otherwise-runnable cells, and
+          # makes the aggregate non-qualifying. Cursor has no BYOK var by
+          # design — no account key here, per BRIEF §5.
+          # Both the runner env names and the protected Qualification secret
+          # names use the provisioned `_API_KEY` suffix. A mismatched right-hand
+          # name expands to an empty value and strict preflight blocks the whole
+          # local-functional matrix before a scenario can run.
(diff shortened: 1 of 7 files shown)


pablonyx:
## E2E CONTROL — exact-head review

Reviewed base `45bfa437bc29ed302416c565681abc3d9e0218d8` → head `d7c46e8e0adff8f6512621bf389581423ac26a83` read-only against the Tier-3 local-world and strict-runner contracts.

### Blocker — LOCAL-T3-CONTROL-001: a failed pre-send snapshot can still accept the stale session

`tests/release/src/scenarios/local/local-session.ts:62-67` converts either a workspace-query failure or a session-query failure into an empty pre-send set. On the next successful poll, `resolveLocalWorkspaceSessionAfter` therefore sees every pre-existing session as newly created and can return the exact stale session this repair is intended to exclude.

I reproduced the path without providers: first `listSessions()` throws; the snapshot becomes `[]`; the next call returns only the pre-send session `stale`; the resolver returns `stale`.

Required repair:

- distinguish a successful empty snapshot from a failed snapshot; query errors in the one-time pre-send snapshot must propagate/fail the cell;
- add a regression where the snapshot query throws and prove the stale session cannot be returned (cover `listSessions`, and preferably `listWorkspaces`, since `resolveLocalWorkspaceIdOnce` currently also swallows that error).

The other reviewed boundaries are correctly shaped on this head: the four BYOK variables map to the actual protected `_API_KEY` secret names; strict preflight remains fail-closed; setup visibility precedes exactly one password-login attempt; a 401 stays red; diagnostics project booleans/status rather than auth response bodies; and SESSION/#1333 is untouched.

**Verdict: CONTROL_CHANGES_REQUIRED / not CODE_MERGE_READY at this exact head.** Ordinary CI is also still settling. A head change stales this verdict.

pablonyx:
## Finding → commit → proof — exact-head repair

Repaired **LOCAL-T3-CONTROL-001** at exact head `2fa112e3710b9bbfe27ce4b913fc03b7cc0157eb` (commit `2fa112e3710b9bbfe27ce4b913fc03b7cc0157eb`).

- `snapshotLocalWorkspaceSessionIds` now propagates both `listWorkspaces` and `listSessions` failures. The send path therefore cannot reinterpret a failed one-time baseline query as a successful empty baseline.
- Successful empty snapshots remain valid (a real local workspace with zero sessions returns an empty set).
- Polling resolvers retain their existing transient-read tolerance at the caller; only the one-time pre-send correctness boundary fails closed.
- Regression reproduces the reported sequence: the first session snapshot query throws, a second query would expose only `stale-session`, and the test proves post-send resolution is never entered and the stale id is never returned. A separate regression covers a workspace-query failure.

Focused offline proof on this exact head:

- `npm exec -- tsx --test src/scenarios/local/local-session.test.ts` — **11/11 green**
- `npm run typecheck` — **green** (using the machine's existing shared test dependencies)
- `git diff --check` — **green**

No Rust, Docker, live qualification, provider call, product workaround, or scope expansion. Requesting exact-head E2E CONTROL re-review; a head change stales this receipt.

pablonyx:
## E2E CONTROL — exact-head delta re-review

Reviewed the repair delta `d7c46e8e0adff8f6512621bf389581423ac26a83..2fa112e3710b9bbfe27ce4b913fc03b7cc0157eb` read-only.

**LOCAL-T3-CONTROL-001 — CLOSED.**

- The one-time pre-send snapshot now propagates both workspace and session query failures instead of converting either into an empty baseline.
- A successful query with no sessions still returns the legitimate empty set.
- `resolveLocalWorkspaceSessionId` and `resolveLocalWorkspaceSessionAfter` retain their caller-owned transient polling behavior; the stricter semantics are confined to the one-time correctness boundary.
- The regression recreates the original failure order and proves that post-send resolution is never entered after the failed snapshot, so the only pre-existing `stale-session` cannot be returned. A separate regression covers workspace-query failure.
- Independent focused execution on this exact head: `local-session.test.ts` **11/11 green**.

No new blocker or material finding in the repair delta. Base/head custody is exact and GitHub reports the PR mergeable.

**Verdict: CONTROL_REVIEW_CLEAN at exact head `2fa112e3710b9bbfe27ce4b913fc03b7cc0157eb`; CODE_MERGE_READY is withheld only until the currently running exact-head required checks settle green.** If they settle green without a head change, no additional code-review round is needed. A head change stales this verdict.

PR #1359, opened 2026-07-18
fix(server): commit first-run claim before responding

pablonyx:
## Outcome

The `/setup` first-run claim transaction now finishes before the successful response starts. A client can claim the instance and immediately authenticate without racing the owner-account commit.

Observed in Release E2E run `29602686092`, cell **T3-INT-1**: login returned `401` immediately after a successful `/setup` claim.

## Change

- `server/proliferate/server/setup/api.py`: make the POST route session dependency function-scoped. FastAPI finalizes `get_async_session` and commits the transaction before the ASGI response starts, while persistence remains owned by the dependency instead of the transport handler.
- `server/tests/unit/test_first_run_claim.py`: replace the sequential claim-then-login check with an ASGI lifecycle regression. The test records dependency commit completion at `http.response.start`; it passes with function scope and deterministically fails with request scope.

## Proof

- `JWT_SECRET=test-jwt-secret CLOUD_SECRET_KEY=test-cloud-secret uv run --extra dev pytest tests/unit/test_first_run_claim.py -q` — 17 passed.
- `uv run --extra dev ruff check proliferate/server/setup/api.py tests/unit/test_first_run_claim.py` — passed.
- `uv run --extra dev ruff format --check proliferate/server/setup/api.py tests/unit/test_first_run_claim.py` — passed.
- `python3 scripts/check_server_boundaries.py` — passed.
- `python3 scripts/check_max_lines.py` and `git diff --check` — passed.
- No live provider, Docker, or Rust work.
- Final exact head `4937ced7adecf8b5b7588d77f2e6ca1a8a692b41` restores dependency-owned function-scoped transaction finalization and is byte-identical in behavior to independently reviewed clean head `abce07c07c666cd3853e34d7fb184a6cdd4b3ccd`. Focused 17/17, server boundaries, max-lines, Ruff, and diff checks passed; 24 exact-head GitHub checks passed with no failures, with only the non-required broad Server test still running at merge custody.

## Provenance / relationship to the local Tier-3 slice

This is the one uniquely valuable product fix salvaged from `codex/local-t3-proof` (frozen base `12bd5501c`). The test/workflow repairs from that slice were independently superseded on `main` by #1350 and #1337; this setup response-ordering fix was not covered by either. The two remaining local Tier-3 UI cells are separate product bugs and stay out of scope here.


pablonyx:
Design ruling from Pablo: transaction ownership lives in the SERVICE layer — `service.claim_first_run` commits via `setup/transactions.commit_first_run_claim` before returning. The `Depends(get_async_session, scope="function")` transport-owned variant (4937ced7a) is reverted at e491f1589 (tree bit-identical to the agent-reviewed MERGE_READY head e4f391163). Any twin session working this PR: please do not re-push the scope="function" design. Merging once CI is green.

PR #1367, opened 2026-07-18
fix(anyharness): reject unapplied session modes

### LINKED ISSUE #979: [Bug]: sessions.create with modeId bypassPermissions is silently downgraded to default
### Bug Description

`POST /v1/sessions` with `modeId: "bypassPermissions"` succeeds (2xx), but the session actually runs in `default` mode. The subsequent `session_state_update` event shows `modeId: "default"` with `requestedModeId: "bypassPermissions"` — no error, no warning. A consumer doing unattended orchestration has no signal that its requested permission mode was not honored; runs only complete if the agent happens never to trip a permission prompt (or the consumer independently auto-approves `interaction_requested` events via the resolve-interaction endpoint).

### Steps to Reproduce

1. Run the AnyHarness runtime locally (`make run PROFILE=<name>`), with the Claude agent `ready`.
2. Via the SDK (or raw HTTP): `sessions.create({ workspaceId, agentKind: "claude", modeId: "bypassPermissions" })` — returns success.
3. Prompt the session and tail `GET /v1/sessions/{id}/events?after_seq=0`.
4. Observe the `session_state_update` envelope: `modeId: "default"`, `requestedModeId: "bypassPermissions"`.

### Expected Behavior

One of:
- the requested mode is honored, or
- session create is rejected (4xx with `ProblemDetails`) when the requested mode can't be applied, or
- at minimum an explicit warning event, so unattended consumers can react instead of silently running in a mode they didn't ask for.

### Area

anyharness

### Environment

- OS: WSL2 Ubuntu (26.04) on Windows 11
- Repo at release-2026-07-05 (36554c4b0), runtime built from source (`cargo`, `make run`)
- Agent: claude (managed install, subscription OAuth credentials)

### Additional Context

Found while building a multi-agent orchestration script against the v1 API + TypeScript SDK. Related observation: permission behavior differs per harness (Cursor's harness emits `interaction_requested` before shell commands and hangs unattended runs until resolved; Claude's adapter in default mode never asked), which makes the silent downgrade harder to notice.

pablonyx:
## Summary

- require the live ACP session to confirm an explicitly requested mode before startup reports ready
- make requested-mode application the final authoritative mode step after cold-resume snapshot restoration
- fall back from the live config option to the legacy mode setter, but reject the create request when neither path applies the value
- return a typed `400 SESSION_MODE_UNSUPPORTED` response instead of silently running the agent's default mode

## Root cause

Session creation validated and persisted `modeId`, but actor startup discarded the legacy setter's `NotApplied` result. The startup path then logged and swallowed every preference-application error before sending its ready signal. A request such as Claude `bypassPermissions` could therefore return success while the active session remained in `default` mode.

Review finding B1367-1 identified a second ordering hazard: a mode queued during a busy turn persists the new requested value while the live-config snapshot still contains the old active mode. Applying the request before cold-resume restoration allowed the stale snapshot to overwrite the confirmed mode before readiness. Startup now keeps model preference behavior in its original best-effort position, restores persisted controls, and only then applies and confirms the explicit requested mode.

## Impact

Explicit session modes are now fail-closed: startup succeeds only when the active agent state already matches or authoritatively confirms the requested value. Best-effort model selection behavior is unchanged.

## Tests

- deterministic unit matrix for confirmed versus unconfirmed mode outcomes
- in-process ACP regression for queued new mode + stale old-mode snapshot + cold resume; asserts the queued mode is the final wire write and persisted current mode
- API mapping unit coverage for the typed 400 response
- `rustfmt --edition 2021 --check` on touched Rust files
- `git diff --check`

Rust compilation was not run locally because this machine has a protected single-build lane; the draft PR delegates the focused Rust tests and workspace build to CI.

Fixes #979

### DIFF
--- a/anyharness/crates/anyharness-lib/src/api/http/sessions_errors.rs
+++ b/anyharness/crates/anyharness-lib/src/api/http/sessions_errors.rs
@@ -2,7 +2,7 @@ use super::access::map_access_error;
 use super::error::ApiError;
 use crate::domains::agents::route_auth::RouteAuthError;
 use crate::domains::sessions::mcp_bindings::crypto::SessionMcpBindingsError;
-use crate::domains::sessions::model::AgentStartupExitError;
+use crate::domains::sessions::model::{AgentStartupExitError, RequestedModeApplyError};
 use crate::domains::sessions::runtime::{
     CreateAndStartSessionError, EnsureLiveSessionError, ForkSessionError,
     PendingPromptMutationError, PendingPromptQueueError, ResolveInteractionError, SendPromptError,
@@ -28,6 +28,10 @@ fn map_internal_anyhow_error(
 }
 
 pub(super) fn map_acp_session_start_error(error: anyhow::Error) -> ApiError {
+    if let Some(mode_error) = error.downcast_ref::<RequestedModeApplyError>() {
+        return ApiError::bad_request(mode_error.to_string(), RequestedModeApplyError::CODE);
+    }
+
     let telemetry_safe_detail = format!("ACP session start failed: {error}");
     map_internal_anyhow_error(error, telemetry_safe_detail, "ACP session start failed: ")
 }
@@ -352,7 +356,7 @@ mod tests {
     use axum::http::StatusCode;
     use axum::response::IntoResponse;
 
-    use crate::domains::sessions::model::AgentStartupExitError;
+    use crate::domains::sessions::model::{AgentStartupExitError, RequestedModeApplyError};
     use crate::domains::sessions::runtime::{CreateAndStartSessionError, ResolveInteractionError};
     use crate::domains::workspaces::access_gate::WorkspaceAccessError;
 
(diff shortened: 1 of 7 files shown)


pablonyx:
B1367-1 repaired at exact head `5aa375e40b67f96f955f0a63dada4b7167c01142`.

- split startup preference handling so best-effort model selection remains before persisted-config restoration
- moved strict requested-mode application/authoritative confirmation after restoration and before the final live-config snapshot/readiness
- added deterministic in-process ACP coverage for a queued new mode plus stale old-mode snapshot on cold resume; the test proves the stale restore writes first, the queued mode is the final wire write, and both requested/current persisted mode end at the queued value
- `rustfmt --edition 2021 --check` on touched Rust files: pass
- `python3 scripts/check_max_lines.py`: pass
- `git diff --check`: pass

Per custody constraints, no local Cargo build/test, Docker, provider, or E2E action was run. CI owns compiled proof for this exact head. The PR remains draft and unmerged.

PR #1401, opened 2026-07-18
fix(server): commit agent key creation before response

pablonyx:
## Summary

- commit agent API-key creation before the success response starts
- prevent the immediate create-and-bind selection request from racing transaction teardown
- add an ASGI-boundary regression for commit-before-response ordering

## Root cause and impact

The Settings “Create and bind” flow sends a key-creation request and then immediately sends a selection `PUT` referencing the returned key ID. The create endpoint used the default request-scoped database dependency, so the response could start before the key transaction committed. A fast follow-up request could therefore return `agent_api_key_not_found` even though creation had reported success.

Using function-scoped dependency cleanup makes the returned key durable before callers can observe the success response.

## Verification

- `DEBUG=true uv run --extra dev pytest -q tests/unit/test_agent_gateway_key_transaction.py tests/unit/test_cloud_workspace_transaction.py tests/integration/test_agent_gateway_api.py::TestAgentApiKeys` — 8 passed
- `DEBUG=true uv run --extra dev ruff check proliferate/server/cloud/agent_gateway/api.py tests/unit/test_agent_gateway_key_transaction.py`
- `DEBUG=true uv run --extra dev mypy tests/unit/test_agent_gateway_key_transaction.py`
- `git diff --check`

## Readiness

- [x] I followed the pull-request procedure for scope, title, body, and proof.
- [x] This PR has no support relationship.

### DIFF
--- a/server/proliferate/server/cloud/agent_gateway/api.py
+++ b/server/proliferate/server/cloud/agent_gateway/api.py
@@ -68,7 +68,7 @@ async def list_agent_api_keys_endpoint(
 @router.post("/keys", response_model=AgentApiKeyResponse)
 async def create_agent_api_key_endpoint(
     body: AgentApiKeyCreateRequest,
-    db: AsyncSession = Depends(get_async_session),
+    db: AsyncSession = Depends(get_async_session, scope="function"),
     user: User = Depends(current_product_user),
 ) -> AgentApiKeyResponse:
     try:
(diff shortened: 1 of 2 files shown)


pablonyx:
Reviewed exact head 6d0f91ebaff214520e0f7931c188a6adc4f8c178. CONTROL_CLEAN / CODE_MERGE_READY for the Grok agent-key commit-race fix. The endpoint now uses FastAPI function-scoped dependency finalization instead of a service-layer commit, preserving the canonical server boundary. The regression proves finalization before response start. CI is green; no blocker/material findings.

Issue #1411, opened 2026-07-18
Qualification OpenCode BYOK cell rejects the configured Anthropic key as invalid


pablonyx:
Exact evidence:

- Main SHA: `a4311a2fe87c0e0520f14327f7bb33a8bd110a96`
- Strict run: https://github.com/proliferate-ai/proliferate/actions/runs/29661799394
- Artifact: `local-functional-evidence` (ID `8434895810`)
- Cell: `T3-AUTHROUTE-1/local/harness=opencode`

The real UI selected the OpenCode direct Anthropic model and sent the bounded prompt. The product stopped the chat with `Internal error: API key is invalid` (`errorName: APIError`, service `session`). Claude, Codex, and Grok LOCAL-3 direct-key cells were green in the same exact run; only the OpenCode/Anthropic-B credential path failed.

Acceptance: distinguish an invalid/expired `RELEASE_E2E_BYOK_ANTHROPIC_B_API_KEY` from incorrect OpenCode key materialization/provider wiring. Rotate only if the credential is invalid; otherwise repair the product/runtime injection path. Add a narrow validation/regression and re-prove the exact cell. Cleanup was complete (24/24).

pablonyx:
Resolved by #1470. The post-fix strict Local qualification run completed 16/16 selected cells green with zero failed or blocked cells, including the OpenCode AUTHROUTE path: https://github.com/proliferate-ai/proliferate/actions/runs/29905059166

Issue #1410, opened 2026-07-18
Qualification route-change actor reaches LiteLLM at its full $5 budget before its gateway turn


pablonyx:
Exact evidence:

- Main SHA: `a4311a2fe87c0e0520f14327f7bb33a8bd110a96`
- Strict run: https://github.com/proliferate-ai/proliferate/actions/runs/29661799394
- Artifact: `local-functional-evidence` (ID `8434895810`)
- Cell: `T3-AUTHROUTE-1/local/route=change`

The user-key leg completed. After switching to gateway and starting the new session, LiteLLM rejected the bounded turn with HTTP 429: `Budget has been exceeded`, reporting current cost exactly `5.0` and max budget `5.0` for the run actor key. This was the first managed-gateway turn for the functional world; the earlier per-harness LOCAL-3 legs were BYOK and proved zero managed spend.

Acceptance: determine why a fresh qualification owner enrollment/key begins or arrives at the full cap; add a regression at the enrollment/LiteLLM boundary; the route-change cell must complete one cheap gateway turn with correlated spend without widening production budgets blindly. Cleanup was complete (24/24).

pablonyx:
A fresh, narrowly scoped diagnostic was dispatched on current main `fbb8dc35294b414ec0ec2a5b1657c88dbae304c6`: https://github.com/proliferate-ai/proliferate/actions/runs/29674146779

Scope is `manual_world=local`, `scenarios=T3-AUTHROUTE-1`, diagnostic behavior. This is a reproduction/classification run only; the cell remains red unless the exact managed-gateway turn and cleanup evidence pass. No budget was widened and no product change is being inferred from the prior 429.

pablonyx:
Current-main diagnostic [29674146779](https://github.com/proliferate-ai/proliferate/actions/runs/29674146779) reproduced `T3-AUTHROUTE-1/local/route=change` red with cleanup completed, but the evidence sanitizer correctly withheld the provider body and therefore could not distinguish the prior budget failure from another provider class.

[...]

Issue #1375, opened 2026-07-18
Empty-tab session-creation POST aborted by reload is never retried; tab stuck on optimistic client-session id


pablonyx:
Third race variant in the #1333 / #1337 empty-session lineage, surfaced by Tier-3 local qualification (T3-SESSION-1/local/harness=claude, Actions run 29628880856, main @9b0195701; evidence artifact local-functional-evidence, debug logs T3-SESSION-1-local-harness-claude-ui-failure-1784347620881.*).

**Symptom:** after a page reload while an empty tab's backing session is not yet server-persisted, the tab re-materializes as a fresh empty session on the default harness with an optimistic `client-session:` id, and the `POST /v1/sessions` that would persist it is cancelled (`net::ERR_ABORTED`) and never retried. The tab remains permanently on the optimistic id — `data-workspace-session-id` on the workspace shell root stays `client-session:claude:...` indefinitely (test observed 30s+; nothing in the network log ever resubmits).

**Evidence:**
- HTML snapshot: tab 0 = reconciled server session `2487f18b-...` (messaged tab); tab 1 = active, harness=claude, session-id `client-session:claude:1784347585944:x0oes0`.
- Network log: `[requestfailed] POST /v1/sessions :: net::ERR_ABORTED`, plus aborted `/v1/agents/reconcile`, `/v1/agents/launch-options`, `/v1/agents` — consistent with reload cancelling in-flight fetches.

**Distinction from #1333/#1337:** #1337 fixed the orphan-reappears-as-extra-tab variant. Here no orphan appears; instead the create POST itself aborts and the client never resubmits, so the session can never reconcile.

**Suspected code paths:** empty-session replacement/creation in `apps/packages/product-client/src/hooks/sessions/workflows/` (`use-empty-session-replacement-cleanup.ts`, `use-session-creation-actions.ts`, `session-creation-materialization.ts`). Fix direction: retry/resume the create-session POST after an aborted attempt (or defer/settle it across reload) rather than leaving the tab stuck on the optimistic id.

This currently keeps T3-SESSION-1 red in the local Tier-3 baseline.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

pablonyx:
Reproduced on exact merged main `a4311a2fe87c0e0520f14327f7bb33a8bd110a96` after the prior repair.

Strict local Tier 3 run: https://github.com/proliferate-ai/proliferate/actions/runs/29661799394
[...]

PR #1413, opened 2026-07-18
fix(release): wait for local sidebar navigation readiness

pablonyx:
## Outcome

Repairs the shared `T3-CFG-1` navigation failure observed on exact-main strict run 29661799394. When a reused renderer is still on the workspace shell, the collector now expands the animated sidebar and proves the `New chat` control has stable in-viewport geometry, remains inside every overflow-clipping ancestor's client area, and owns its center hit target before clicking once.

This does not force clicks, retry product actions, or convert any product failure to green.

## Exact custody

- Base: `6fdec94eba59192bf17fd46435b1da5069127c3f`
- Final head: `f80205dbf20e160c308841a4a97826f8a5af9951`
- Evidence source: Actions run 29661799394, artifact `local-functional-evidence` ID 8434895810
- Independent verdict: `CONTROL_CLEAN / CODE_MERGE_READY` at comment 5013223869

## Offline proof

- focused config/sidebar tests: 30/30 green
- independent delta regression: 3/3 green
- `npm run typecheck`: green
- `python3 scripts/check_max_lines.py`: green
- `git diff --check`: green
- exact-head repository CI: fully green

No Rust build, Docker, provider call, or live qualification run was performed for the repair. Strict live `T3-CFG-1` replay remains separate acceptance evidence. Product/configuration failures remain separately tracked in #1375, #1410, and #1411.

### DIFF
--- a/tests/release/src/scenarios/local/config-session.ts
+++ b/tests/release/src/scenarios/local/config-session.ts
@@ -25,6 +25,7 @@ import {
   GATEWAY_UNSUPPORTED_HARNESSES,
   gatewayUnsupportedMessage,
 } from "../../fixtures/gateway-unsupported-harnesses.js";
+import { waitForSidebarControlReady } from "./sidebar-control-readiness.js";
 
 /**
  * LOCAL-4 (live configuration matrix, per harness) under `T3-CFG-1/local`, and
@@ -940,7 +941,7 @@ async function selectRepoAndWorkLocally(page: ProductPage, repo: PreparedReposit
   // requires.
   if (!await projectTrigger.isVisible().catch(() => false)) {
     const newChatNav = p.locator("nav").getByRole("button", { name: "New chat" }).first();
-    await newChatNav.waitFor({ state: "visible", timeout: 30_000 });
+    await waitForSidebarControlReady(p, newChatNav);
     await newChatNav.click();
     await p
       .locator("[data-home-composer-editor]")
--- a/tests/release/src/scenarios/local/sidebar-control-readiness.test.ts
+++ b/tests/release/src/scenarios/local/sidebar-control-readiness.test.ts
@@ -7,6 +7,7 @@ import { waitForSidebarControlReady } from "./sidebar-control-readiness.js";
 
 interface FakeControlOptions {
   topAtCenter: Array<"overlay" | "target">;
+  clippingAncestorRight?: number;
 }
 
 function fakeCollapsedSidebar(options: FakeControlOptions): {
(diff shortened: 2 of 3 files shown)


pablonyx:
`LOCAL-1413-B1` — **Blocker: ancestor clipping can still false-green readiness**

Exact head reviewed: `c5d3d1e559186c5ed1cd765a23f2f5d55f5f6b3d`. In `tests/release/src/scenarios/local/sidebar-control-readiness.ts`, `stableAndInViewport` compares only the control's own rect to the viewport, and the final check proves only that its center point belongs to the control. It never intersects that rect with clipping ancestors. The real shell wraps the sidebar in `#main-sidebar` with `overflow-hidden`, so a 100px row exposed through only 50px of its parent can satisfy every current predicate once its center is exposed; the subsequent ordinary center click can navigate successfully and let the cell go green while the UI is still clipped.

I independently reproduced this against the exact head with a fake target rect `{left:20,right:120}` under a clipping-parent rect `{left:20,right:70}` and `elementFromPoint(center) === target`; `waitForSidebarControlReady` resolved. The committed tests do not model an ancestor clip. Please make readiness fail closed on the control being fully inside every effective overflow-clipping ancestor (or an equivalently ownership-correct complete-sidebar invariant), and add the clipped-ancestor regression.

The rest of the click semantics are clean: the patch adds exactly one ordinary `Show sidebar` click before the existing ordinary `New chat` click, with no `{ force: true }` and no explicit retry loop around either product action.

pablonyx:
### LOCAL-1413-B1 resolved

- **Finding:** ancestor overflow clipping could false-green readiness even when the control owned its exposed center point.
- **Commit:** `f80205dbf20e160c308841a4a97826f8a5af9951`
- **Resolution:** readiness now fails closed unless the control rect is fully contained, per axis, by the client-area clip box of every ancestor whose computed overflow clips that axis. The existing geometry-stability, viewport, and center hit-ownership predicates remain intact. Click semantics are unchanged: one ordinary sidebar-toggle click when needed and one ordinary `New chat` click, with no force, action retry, or sleep-only acceptance.
- **Regression proof:** the new clipped-ancestor case failed on reviewed head `c5d3d1e559186c5ed1cd765a23f2f5d55f5f6b3d` with `Missing expected rejection` (2/3 passed), then passed after the fix.
- **Exact-head offline proof:**
  - focused LOCAL config/sidebar tests: 30/30 passed
  - `pnpm run typecheck`: passed
  - `python3 scripts/check_max_lines.py`: passed
  - `git diff --check`: passed

No live E2E, provider call, Docker, or Rust build was run.

Please re-review exact head `f80205dbf20e160c308841a4a97826f8a5af9951` for LOCAL-1413-B1.

pablonyx:
### Delta re-review verdict — `f80205dbf20e160c308841a4a97826f8a5af9951`

`LOCAL-1413-B1` — **RESOLVED**

The repair now fails closed on the condition the original head missed:

- every element ancestor is inspected;
- X and Y clipping are evaluated independently from computed `overflowX` / `overflowY`;
- the target rect must fit inside each clipping ancestor's client-area clip box on the axis that ancestor clips;
- viewport containment, geometry stability, and center-point hit ownership remain independent predicates, so the new ancestor check neither substitutes for nor tautologically satisfies them.

The regression models the old false-green exactly: target rect `left=20,right=120`, overflow ancestor clip edge `right=70`, while `elementFromPoint(center)` still returns the target. The pre-repair implementation ignores the ancestor and resolves, making the new `assert.rejects` fail; this exact head rejects it. A visible, stable, unclipped target still passes, and overlay ownership still delays acceptance.

Independent exact-head proof:

- focused `sidebar-control-readiness.test.ts`: **3/3 passed**
- release-runner `pnpm run typecheck`: **passed**
- exact-head GitHub checks: **all completed checks passed** (the compose smoke is correctly skipped as non-smoke-relevant)
- full diff remains limited to the local `T3-CFG-1` navigation harness and its focused tests; no forced click or product-action retry was introduced

No new Blocker, Decision, or Follow-up finding.

**CONTROL_CLEAN**  
**CODE_MERGE_READY**

This verdict is code/control readiness only. Exact-head local-world acceptance still requires the planned strict live `T3-CFG-1` replay and its normal cleanup evidence; this PR does not itself claim world acceptance.

PR #1459, opened 2026-07-20
fix(release): bind local qualification selectors

pablonyx:
## What changed

- pass the typed `workflow_dispatch` scenario, agent, and behavior inputs unchanged into both Local preflight and execution
- fail Local preflight on missing/malformed selectors and validate explicit agent kinds against the shipped catalog
- bind behavior and the exact selected agent selector into the redacted preflight receipt (schema v2)
- prove the explicit `claude,codex,grok,opencode` selector expands to exactly four functional cells
- document the fail-before-install/build/provider-spend contract

## Why

The Local workflow previously validated only scenarios and used fallback expressions in a different GitHub context at execution time. That left no machine-readable custody proof that the exact agent selector accepted at dispatch was the selector expanded by the runner, and malformed or missing selectors could reach later seams.

## Impact

Strict Local qualification now fails before dependency installation, candidate builds, or provider spend unless both scenario and agent selectors are valid. Explicit selector receipts are redacted and catalog-backed; normal `all` behavior remains supported.

## Verification

- `node --test scripts/ci-cd/qualification-preflight.test.mjs` (12 passed)
- `./tests/release/node_modules/.bin/tsx --test tests/release/src/runner/qualification-execution-workflow.test.ts tests/release/src/scenarios/t3-chat-1.test.ts` (23 passed)
- `pnpm --dir tests/release typecheck`
- `actionlint .github/workflows/release-e2e.yml`
- `git diff --check`

### DIFF
--- a/.github/workflows/release-e2e.yml
+++ b/.github/workflows/release-e2e.yml
@@ -556,7 +556,9 @@ jobs:
           RELEASE_E2E_BYOK_XAI_API_KEY: ${{ secrets.RELEASE_E2E_BYOK_XAI_API_KEY }}
           RELEASE_E2E_INTEGRATION_NAMESPACE: ${{ vars.RELEASE_E2E_INTEGRATION_NAMESPACE }}
           RELEASE_E2E_INTEGRATION_API_KEY: ${{ secrets.RELEASE_E2E_INTEGRATION_API_KEY }}
-          SCENARIOS_INPUT: ${{ github.event.inputs.scenarios || 'all' }}
+          SCENARIOS_INPUT: ${{ inputs.scenarios }}
+          AGENTS_INPUT: ${{ inputs.agents }}
+          BEHAVIOR_INPUT: ${{ inputs.local_functional_behavior }}
         run: |
           node scripts/ci-cd/qualification-preflight.mjs \
             --world local \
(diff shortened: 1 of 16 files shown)


pablonyx:
Review follow-up pushed at exact head `a889b93ddbba9aa38ab97e2cd5adafddf9ed65f2`.

- `LSEL-CONTROL-001`: both `qualification-local-workspace` and `qualification-local-functional` now pass their exact scenario/agent/behavior selectors to the mandatory Make-level preflight in both build and reuse paths. The regression executes both real Make targets with only install/build/runner commands shimmed, requires status 0, and verifies the written strict receipts.
- `LSEL-CONTROL-002`: Local preflight now validates explicit scenario ids against the machine-owned Tier 3 manifest plus `LOCAL-WORLD-SMOKE-1`; unknown, duplicate, malformed, and missing strict selectors fail closed. The real preflight command returns exit 2 and a failed `scenario_catalog` receipt for `T3-NOT-REAL`.

Verification:
- preflight tests: 13 passed
- workflow/scenario tests: 24 passed
- machine scenario manifest tests: 9 passed
- release typecheck
- actionlint
- docs validation (224 files)
- diff check

Requesting independent exact-head re-review; no live qualification run was dispatched.

pablonyx:
Exact-head follow-up for reopened `LSEL-CONTROL-002`: `3140aceca78bd04c90ef081c200245b3e882ea15`

- Replaces the target-guarantee manifest allowlist with `qualification-world-scenarios.json`, the shared executable Local inventory consumed by preflight and runner selection.
- Both real Local Make entrypoints pass `--qualification-world local`; unrelated `--lane local` targets retain their existing registry behavior.
- Pre-spend adversarial coverage rejects `T3-AUTH-1`, `T3-MOBILITY-1`, and non-Local `T3-SH-2`; every executable Local scenario and `all` are accepted.
- Preserves the real-Make regression that closed `LSEL-CONTROL-001`.

Focused proof: preflight 15/15; release CLI/registry/workflow 73/73; typecheck; core manifest 9/9; actionlint; docs 224; diff check. Independent exact-head re-review requested. No live qualification or merge performed.

PR #1461, opened 2026-07-20
fix(release): fail early on missing S3 cleanup authority

pablonyx:
## Summary

- probe a reserved, never-written run-scoped S3 key with a bounded read-only `HEAD` before cleanup registration or the first upload
- permit provisioning only when the probe observes exact `404 Not Found`; collapse every other provider response into one stable, sanitized failure
- preserve the existing fail-closed post-delete `HEAD` proof and document the readiness contract

## Root cause

The self-host CFN qualification role can upload and delete run artifacts but cannot distinguish a missing object from denied access. Amazon S3 returns 403 for a missing object when the caller lacks `s3:ListBucket`, so run cleanup correctly stayed red even though independent provider inspection found no surviving objects. The old path discovered this only after provisioning and teardown.

## Required qualification IAM follow-up (not changed here)

Add only `s3:ListBucket` on `arn:aws:s3:::proliferate-qualification-artifacts`, constrained by `StringLike` `s3:prefix` = `qualification/*`. Existing `s3:GetObject` authority on `arn:aws:s3:::proliferate-qualification-artifacts/qualification/*` remains the object-side permission for `HeadObject`. No AWS IAM or GitHub state was changed by this PR.

AWS documents the missing-object distinction: https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html

## Validation

- `pnpm --dir tests/release exec tsx --test src/worlds/selfhost/cfn.test.ts` (53/53)
- `pnpm --dir tests/release run typecheck`
- `git diff --check`

### DIFF
--- a/specs/developing/testing/release-worlds-and-fixtures.md
+++ b/specs/developing/testing/release-worlds-and-fixtures.md
@@ -793,7 +793,10 @@ summary beside the parent run logs after both setup failure and normal close.
 Deleting its nested ledger is therefore not itself treated as zero-survivor
 proof: the retained receipt must independently report stack and Route53
 absence, exact S3 `HEAD` absence, bounded GHCR tag-relist absence, local-path
-removal, and zero cleanup failures.
+removal, and zero cleanup failures. Before its first S3 write, the world also
+issues a bounded, read-only `HEAD` against a reserved never-written key in the
+run prefix. Only an exact missing-object response permits provisioning to
+continue; authorization failures stay red and fail before provider mutation.
 
 GitHub Actions cancellation and job timeout are a distinct failure boundary:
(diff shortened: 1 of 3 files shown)


pablonyx:
Qualification IAM follow-through is complete: added inline policy `QualificationSelfHostCfnS3AbsenceProof` to `proliferate-qualification-github-actions`, granting only `s3:ListBucket` on `arn:aws:s3:::proliferate-qualification-artifacts` when `s3:prefix` matches `qualification/*`. IAM simulation is `allowed` for `qualification/probe/never-written` and `implicitDeny` for `outside/probe`. Existing Put/Get/Delete object scope is unchanged. This is the exact authority required for a missing-key HEAD to return 404 rather than an unprovable 403.

PR #1462, opened 2026-07-20
fix(selfhost): unblock CloudFormation public health

pablonyx:
## Summary
- keep the EC2 CreationPolicy as the fail-closed local-bootstrap gate for both initial creation and physical-instance replacements
- let the instance complete after local API readiness so its Elastic IP and DNS resources can converge, then gate the stack on the real public HTTPS endpoint through a bounded Lambda-backed custom resource
- rerun that public gate when the physical instance changes; explicitly do not claim that asynchronous release-only `cfn-hup` updates are attested by CloudFormation `UPDATE_COMPLETE`
- add a CloudFormation-only public-skip seam to the shared host health script; manual, direct-installer, and `update.sh` paths continue checking both local and public targets

This repairs the deterministic dependency deadlock exposed by exact-main strict run 29743453042 without weakening latest-AMI replacement updates. Independent review rejected both the prior static WaitCondition (not replacement-safe) and a release-version custom-resource trigger (could false-green the old healthy release before `cfn-hup` ran).

## Proof
- `PROLIFERATE_TEST_AWS=1 bash server/deploy/tests/run.sh` (138 passed; AWS template validation passed)
- exact extracted inline handler test via `server/deploy/tests/public-health-custom-resource.py`
- `pnpm --dir tests/release typecheck`
- `pnpm exec tsx --test src/worlds/selfhost/cfn.test.ts src/scenarios/selfhost-cfn-1.test.ts` (74 passed)
- `git diff --check`


pablonyx:
Qualification IAM is ready for the exact-head proof. Added inline policy `QualificationSelfHostCfnPublicHealthGate` to `proliferate-qualification-github-actions`:

- Lambda lifecycle/invoke/tag actions only on `arn:aws:lambda:us-east-1:157466816238:function:proliferate-sh-cfn-*`
- `iam:PassRole` only on `arn:aws:iam::157466816238:role/proliferate-sh-cfn-*` and only when `iam:PassedToService=lambda.amazonaws.com`
- no data-plane or CloudWatch Logs permissions for the probe execution role

IAM simulation: stack-scoped function + Lambda pass-role are `allowed`; unrelated function and role are `implicitDeny`.

PR #1490, opened 2026-07-25
docs(specs): sandbox-lifecycle platform spec + template operations runbook

pablonyx:
## Summary

PR 1 of the sandbox spec program (PR 0 = #1487 deletions). Adds the platform spec for the sandbox **container** and extends the template runbook.

**`specs/codebase/platforms/product/sandbox-lifecycle.md`** (Status: target, per `specs/authoring.md` conventions):
- How E2B works: pause preserves processes, auto-resume on traffic, indefinite paused retention — with E2B docs linked at point of use.
- The template arc: build (binaries baked, 4 CPU/8 GB) → publish (immutable `sha-` tags, rolling tags) → instantiate (metadata attribution) → initialize (Supervisor-owned launch, health + auth gates).
- Account model: one sandbox per (user, org context), billed to the matching subject.
- State machine + a single-cause transition table (every wake/pause/error/destroy cause: trigger, E2B touched?, transition, why).
- Laws: only explicit delete destroys; error is per-attempt; ensure never provisions; **the gateway gates on policy, E2B wakes on traffic, materialization repairs**.
- The provisioning engine, pause/resume end to end, health (worker death is noticed: status + alert), API surface, code map, failure modes, proof.

**Rulings encoded** (founder, this session): policy-gate gateway; `wake` collapses into `ensure` (hard rename); Supervisor-owned runtime is the spec body (flag-off = gap pending the live N-1→N proof); worker degradation surfaces in runtime-status + production alerts; per-org sandbox isolation is the ideal.

**8 current-gaps items** carry today's deltas: wake route, 409 liveness gate, legacy launch default, silent worker death, one-sandbox-per-user-globally, hardcoded `e2b_template_ref`, no disk telemetry, dead `_runtime_status` branches.

**`specs/developing/operating/e2b-template-operations.md`** (renamed from `e2b-template-rollback.md`): new Build-and-publish section (CI lanes, scripts, local-build boundary); rollback procedure unchanged; all inbound references repointed.

`scripts/check_docs.py` green (221 files).

## Program context

Next: PR 2 `sandbox-content.md`, PR 3 `sandbox-gateway.md`, PR 4 `sandbox-access.md`, PR 5 systems doc + refit sweep. Plan of record: `codex/sandbox-specs-plan.md` (local).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1496, opened 2026-07-25
docs(specs): add sandbox-gateway platform spec (PR 3/5, cloud-sandbox program)

pablonyx:
PR 3 of the cloud-sandbox spec program (PR 1 #1490 lifecycle, PR 2 #1491 content). Stacked on #1491's branch; retargets down the stack as predecessors merge.

Adds `specs/codebase/platforms/product/sandbox-gateway.md` (Status: target), owning **the wire**: how product bytes reach the AnyHarness runtime inside a cloud sandbox, and which traffic deliberately never takes that wire.

## What it covers

- **One route, one picture**: `/v1/gateway/cloud-sandbox/anyharness/{path}` (HTTP all-methods + WebSocket), product JWT in, runtime bearer out. No sandbox id in the URL — the caller's own personal sandbox is the only reachable target.
- **The wire laws**: the client never learns the sandbox address; one wire for every product byte (web/desktop/mobile, REST/SSE/WS); the credential swap is total; no DB session across a stream (commit-before-stream, test-pinned); the gateway gates on policy not liveness (forwarding is the wake — inherits lifecycle's ruling); a gateway connection is the same generic AnyHarness client with a different base URL.
- **Access resolution**: billing gate (402) → ensure row → runtime-access load (typed 409 when unmaterialized) → 60s per-user cache.
- **Streaming**: infinite reads / bounded handshakes; the pinned SSE half-close guard; WS pump semantics; chat=SSE vs terminals=WS and the query-vs-subprotocol auth-transport pivot.
- **Client contract**: workspace→connection resolution chain, host-armed token pointer (no special gateway token exists), typed retry (8×750ms), fresh-token re-mint before every connect.
- **Proxied-vs-direct table**: AnyHarness proxied, LiteLLM direct, git direct, integrations proxied-inbound, E2B exec/webhooks, LiteLLM spend pull, SSH targets direct-by-design — unified by *credentials stay where they are revocable*. Direct-attach framed honestly post-claiming-revert (#803): not built, not scheduled.

## Current gaps (each a follow-up code PR)

- Cold access has no repair trigger (inherits lifecycle's open wake-choreography ruling; paused-wake already works — the gap is rows without stamped access).
- The 60s access cache outlives runtime resets → up to a minute of 502s instead of the typed 409.
- Legacy `?access_token=` WS auth should drop (subprotocol transport is universal now).
- Mobile duplicates the connection resolver.

Also repoints `sandbox-lifecycle.md`'s planned-document fence to the real file and adds the platform README row.

Docs check green (223 files).


pablonyx:
Founder-aligned; ready for agent review @ 5155376949dcdba7689c026f43d76b667bf9f5fd (PR #1496).

PR #1497, opened 2026-07-25
docs(specs): design-system platform document — the single design-system entry point

pablonyx:
## What

Adds `specs/codebase/platforms/product/design-system.md` — the single canonical entry point Pablo asked for ("a single .md to look at with our full design system"), plus its Platform Map index row.

## Contents

1. Where the system lives (tokens.ts → generate-theme → check-theme pipeline, gates, hook wiring)
2. Design philosophy & Codex-first reference precedence (R-V2-1)
3. Typography (full ramp), color roles, spacing/radius/icon/control anatomy
4. Ruled surface anatomies: composer, transcript, tab strip — with their pinned values
5. Clean-redo laws and the enforcement stack (appearance-scaling gate, frozen census baseline, check-theme byte equality, twMerge drift locks)
6. How to change a value: the tokens.ts-with-provenance recipe
7. Genuine current gaps (freehand size-N unenforced, etc.)

## Authority model

`tokens.ts` remains the value authority; this doc is the prose *why* and loses on any conflict. The ui-foundation-target-v2 / chat-addendum / retune-spec documents are named as the archived rulings record per specs/README.md's git-history-is-the-archive rule.

## Process

Authored by a distill→write→adversarial-verify workflow (2 verify rounds, 8 defects found and fixed — gate-behavior claims, branch-availability markers, doc-kind framing), then rehomed as a canonical current document. `check_docs.py` green (224 files), all 27 cited paths verified on disk.

## Stacking

Based on `ui-foundation-v2` (PR #1493) because the doc describes the landed foundation; retarget to `main` after #1493 merges.

### DIFF
--- a/specs/codebase/platforms/product/README.md
+++ b/specs/codebase/platforms/product/README.md
@@ -10,6 +10,7 @@ expectations, but it is not the owner of a full user workflow.
 
 | Platform | Owns | Read |
 | --- | --- | --- |
+| Design system | The ruled look (type ramp, control weight, composer/transcript anatomy, tab strip, icon tiers), reference precedence, the token-authority and gate model, and change control for moving a value. | [design-system.md](design-system.md) |
 | Cloud sandbox provisioning | One active personal sandbox row, just-in-time E2B create/resume, direct AnyHarness launch and gateway access, and the optional Worker sidecar. | [sandbox-provisioning.md](sandbox-provisioning.md) |
 | Cloud workspace provisioning | Repository-environment materialization, synchronous Cloud workspace creation, direct AnyHarness worktree creation, and Cloud/runtime truth boundaries. | [workspace-provisioning.md](workspace-provisioning.md), then [sandbox-provisioning.md](sandbox-provisioning.md) |
 | MCP runtime | AnyHarness MCP concepts, user bindings, session extensions, product MCP serving, and elicitation boundaries. | [mcp-runtime.md](mcp-runtime.md) |
--- a/specs/codebase/platforms/product/design-system.md
+++ b/specs/codebase/platforms/product/design-system.md
@@ -1,0 +1,338 @@
+# Design System
+
+The single entry point for the design system: the philosophy and reference
+precedence that decide a value, the ruled look those decisions produced, the
+gates that keep the codebase on it, and the change-control model for moving a
+value.
+
+**Two records, one authority.** This document is the durable prose record of
+*why* the system looks the way it does;
+[tokens.ts](../../../../apps/packages/design/src/tokens.ts) is the value
+authority and the only artifact that can settle a number, with its inline
+`[RETUNE:…]` / `RT-TOK-xxx` provenance comments carrying per-value attribution.
+Where the two disagree about a value, `tokens.ts` is right and this document is
+stale. The reference-alignment pass that produced these rulings
+(`ui-foundation-target-v2.md`, `ui-foundation-chat-addendum.md`, and the
+`.foundation-scout/retune-spec.md` capture census) is a historical rulings
+record archived on the `ui-foundation-pass` branch; per
+[specs/README.md](../../../README.md) git history is the archive, and those
+documents are not a live read path — this document plus `tokens.ts` provenance
+replace them.
+
+**Owns:** the ruled look of the system (type ramp, control weight, composer and
+transcript anatomy, tab strip, icon tiers), the clean-redo laws, and the
+change-control model.
+
+**Does not own:** per-component anatomy (the owning component lives in
+[apps/packages/ui](../../../../apps/packages/ui)), the source-layout and
+class-authoring rules for styling
+([structures/frontend/guides/styling.md](../../structures/frontend/guides/styling.md)),
+the gate script's own contract
+([systems/product/settings/appearance-scaling.md](../../systems/product/settings/appearance-scaling.md)),
+or the historical capture evidence behind each adopted value (the archived
+rulings record above).
+
+## Where the system lives
+
+```text
+apps/packages/design/
+├── src/tokens.ts                 the value authority: every color/type/radius/size/motion token
+├── src/motion.ts                 duration + easing primitives tokens.ts composes
+├── scripts/generate-theme.mjs    projects dist/tokens.js → dist/theme.css
+├── scripts/copy-dom-css.mjs      copies every src/css/*.css (dom/product/desktop) into dist/css
+└── scripts/check-theme.mjs       re-projects independently, asserts byte equality + Tailwind compile() + dom.css import order
+
+apps/packages/ui/src/layout/RowActionIconButton.tsx   sanctioned row-action primitive (size-7 box, icon-control glyph)
+
+scripts/check_appearance_scaling.py   the class-level gate (arbitrary radius/z/gap/size and text/leading brackets, stock+keystone shadow, foreground-alpha, raw hex, backdrop-filter, numeric-duration and long-list bans)
+scripts/git-hooks/pre-commit          staged-file gate; documents the load-bearing build order
+Makefile                              wires the hook (core.hooksPath scripts/git-hooks) via the git-hooks target
+```
+
+## Design philosophy
+
+**Codex-first reference precedence (R-V2-1, ruled Pablo 2026-07-25):** "Codex UI
+is preferred in general; Conductor is authority only where the product's shape
+matches Conductor and not Codex."
+
+Operationalized:
+- **Codex is the default authority** for every Codex-analogous surface: chat
+  input/composer, transcript/messages, buttons and icon controls,
+  popovers/menus, settings, file tree, panels/toolbars, empty states. Where
+  both references were captured for the same surface, the Codex value wins
+  unless a ruling below says otherwise.
+- **Conductor is the authority only for product shapes Codex doesn't have**:
+  the multi-tab chat strip (we keep multiple chat tabs — anatomy/motion align
+  with Conductor's `center-tab-strip/` + `center-session-tab/` captures), and
+  any other multi-session sidebar/workspace structure without a Codex
+  analogue.
+- **Tie-break**: Codex wins on *style* (color roles, type, radii, shadows,
+  motion feel); Conductor wins only on *structure* Codex lacks.
+- Every adopted value cites its capture evidence path. PR #1484 had already
+  chosen Codex numbers at its decision points (13/20 chat, 16px-glyph/28px-box
+  row actions, Codex-soft radii); this precedence is that pattern made law
+  rather than decided case-by-case.
+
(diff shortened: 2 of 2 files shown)


pablonyx (commented):
Agent review — PR #1497 @ 5dfe8dc07 (design-system-doc → ui-foundation-v2). Documentation truth pass over `specs/codebase/platforms/product/design-system.md`, verified line-by-line against `apps/packages/design/src/tokens.ts`, `scripts/check_appearance_scaling.py`, `apps/packages/design/scripts/check-theme.mjs`, `scripts/git-hooks/pre-commit`, and the `Makefile` on this branch.

## Verified correct (value-by-value)

**Type ramp** — every row matches `tokens.ts` exactly, including paired letter-spacing:
`--text-ui-sm` 11/15/+0.01em, `--text-ui` 12/17/+0.005em, `--text-chat` and `--text-composer` and `--text-body` 13/20/0, `--text-body-emphasis` and `--text-workspace-title` 14/21/−0.005em, `--text-heading` 16/23 via `--tracking-heading: -0.01em`, `--text-title` 19/24 via `--tracking-tight: -0.025em`, `--text-hero` 26/34/−0.025em. The generic steps really are nulled in the manifest (`--text-xs/sm/base/lg/xl: null`) and banned by `FIXED_TEXT_PATTERNS` → `fixed-stock-text-utility`; the only surviving hits in product source are a comment in `tw-merge.ts` and two test assertions.

**Composer** — `--radius-composer: 0.75rem` (12px); `--color-composer-background` dark `color-mix(in oklab, #2d2d2d 96%, transparent)` with `themeFallback: rgba(45, 45, 45, 0.96)` and light unchanged at `rgba(255, 255, 255, 0.864)`; `--color-composer-backdrop-filter` dark `none` / light `blur(16px)` with the WKWebView perf rationale recorded on the token itself; `--color-active` 5.2% and `--color-hover` 7.8% both confirmed, and `ComposerControlButton.tsx:41` carries the D-V2-4 note that ruled vocabulary beats Codex's literal 15%; 28×28 box = `h-7 w-7` on the `iconOnly` branch, and the 16px glyph resolves because `Button` `size="sm"` supplies `text-ui` (12px) × `--icon-control: 1.333333em`; `--shadow-composer` is a tagged legacy alias of `--shadow-subtle: 0 1px 2px 0 rgb(0 0 0 / 0.05)`; `--color-foreground-tertiary` is 50% vs `--color-muted-foreground` 70%; `gap-2` control rows confirmed in `ChatComposerControlRowFrame.tsx` and `CloudChatComposerControls.tsx`.

**Transcript** — `--container-transcript-readable: 40rem`, `--container-transcript-wide: 64rem`, `--spacing-transcript-turn: 0.75rem`. The stale-comment call-out is precisely right: `tokens.ts:1094-1095` does claim the gate bans arbitrary bracket widths, and it does not — there is no `max-w` rule anywhere in `check_appearance_scaling.py`. `max-w-[77%]` is live in both cited files and the gate passes on `UserMessage.tsx`.

**Tab strip / icons** — `--workspace-shell-tab-height: 1.75rem`, `--workspace-shell-tab-radius: 0.375rem` (6px), `--size-icon-button-sm/md/lg` = 1.25/1.5/1.75rem, and `size-icon-button-sm` is genuinely applied to the tab close button (`ChromeWorkspaceTab.tsx:108,172`). Every line-number citation in Current gaps is accurate: `IconButton.tsx:28-29` (`size-5`/`size-6`), `PaneIconButton.tsx:6` (`size-6`), `SidebarActionButton.tsx:47` (`size-6`), `Button.tsx:52` (`icon-sm` = `h-7 w-7`).

**Gates and paths** — `FOREGROUND_ALPHA_RE` and `OLD_SHADOW_RE` (with `keystone`) and the four `ARBITRARY_*_RE` all exist; none of them exist in `main`'s copy of the script, which is exactly 311 lines, so the "311-line version this system replaced" claim checks out. `check-theme.mjs` does all three asserted jobs (independent re-projection + byte equality, real `compile()` from `tailwindcss`, `dom.css` tailwind→theme→`@source` ordering). The quoted `build` string matches `package.json` verbatim, and `generate-theme.mjs:22` really imports `../dist/tokens.js`, so the "tsc is load-bearing" argument holds and the hook documents it. `Makefile:1579-1581` sets `core.hooksPath scripts/git-hooks`. The failure-mode row about design CSS is unusually precise and correct: `check_design_css_source` emits `authored-theme-block`/`authored-root-token` but never runs `RAW_HEX_RE`, so a literal inside a component rule does escape. `python3 scripts/check_docs.py` passes (224 files); every relative link and file path in the doc resolves.

## Findings

**1. (must fix) The "two records" paragraph attributes provenance to `tokens.ts` that is not there.** Line 12 says `tokens.ts` carries "inline `[RETUNE:…]` / `RT-TOK-xxx` provenance comments." `grep -c RT-TOK apps/packages/design/src/tokens.ts` → **0**, and 0 in `motion.ts` too. The `[RETUNE:…]` half is right; the `RT-TOK-xxx` half is not. This matters because it is the sentence that defines where a reader goes for per-value attribution.

**2. (must fix) `RT-TOK-020` is cited as if it were a live, resolvable ID.** Line 181 calls the shadow collapse an "already-recorded collapse (RT-TOK-020)". That ID exists only in `origin/ui-foundation-pass` (`retune-tokens-draft.md`, `ui-foundation-chat-addendum.md`) — the same archive the doc's own opening declares "not a live read path." Either drop the bare ID, or say explicitly that RT-TOK ids resolve only in the archived record. Same question applies to `D-V2-1..3`, `RAD-07`, `STATE-02`, `ICON-03`, which appear nowhere in code (`D-V2-4`, `R-V2-1/2`, `CHAT-01/03/05`, `RAD-04`, `SPACE-01`, `ROW-ACTION-01` do).

**3. (must fix) The foreground-alpha failure-mode row overstates the ban.** The table says a banned class includes "`bg-foreground/<alpha>`" unqualified, but `check_appearance_scaling.py:490` only reports when `foreground_alpha_percent(...) <= 10`. Codex's literal `/15` — the very number the composer section spends a bullet rejecting — would sail past the gate. Add the ≤10% qualifier so the row does not promise enforcement that is not there.

**4. (should fix) Leftover program/checkpoint framing.** Several statements describe a delivery pass in flight rather than current codebase behavior, which is what the brief flagged: "an independent review agent sweeps the census after migration and its findings must reach zero before the founder checkpoint"; "the served build is inspected per-surface … before the checkpoint"; "all 756 census sites migrate to the semantic roles" (future tense, and 756 has no live source in the repo); "Escalation convention … zero tags must remain at completion" (there are already 0 `ui-foundation-escalation:` tags in the tree). Per `specs/README.md`, canonical documents describe behavior on `main` and implementation progress belongs in the PR. Recast these in the present, or move them under a labeled target/process boundary.

**5. (nit) Heading case diverges from siblings.** `platforms/product/*.md` use Title Case (`## Platform Map`, `## Truth Sources`); this doc uses sentence case (`## Where the system lives`, `## Design philosophy`, `## Current gaps`).

**6. (nit) `- [ ]` checkboxes in Current gaps** read as a tracker. The only precedent (`model-gateway.md`) is a `Status: target` document; this one is a current-behavior document. Plain bullets would fit the authority table better.

**7. (nit) `--radius-3xl` (line 152)** is a Codex-side token name with no counterpart in our `tokens.ts` (our largest is `--radius-2xl: 1rem`). Worth marking as Codex's name so a reader does not grep for it here.

## Verdict

The substance is strong — this is the most numerically accurate spec truth pass I have checked on this program, and the Current gaps section correctly self-reports the two places where the doc claims more enforcement than exists (icon-button sizes, `max-w`). Findings 1-3 are the same class of defect the doc's own gaps section catches, applied to the doc itself: three claims about where provenance lives and what the gate enforces that do not hold on this branch. All three are one-line edits. Fix those (and preferably 4) and this is ready.

CHANGES_NEEDED

pablonyx (commented):
Agent re-review — PR #1497 @ c6cb9237d (diff 5dfe8dc07..c6cb9237d re-verified against my 7 findings).

**1. RT-TOK-xxx provenance claim — FIXED.** Line 12 now reads "its inline `[RETUNE:…]` provenance comments carrying per-value attribution." That is exactly what `tokens.ts` contains; no false claim remains about a marker that greps to 0.

**2. Unresolvable ruling IDs — FIXED (by the better route).** The change-control bullet now states that these IDs, "wherever they appear in this document," cite entries in the archived rulings record, "not identifiers in code," and gives the exact recipe (`git show origin/ui-foundation-pass:<file>`). I confirmed that ref exists (`3f1cb3797`) and that `RT-TOK-020` resolves in `retune-tokens-draft.md` / `ui-foundation-chat-addendum.md` there. The bare `RT-TOK-020` at line 182 is now covered by that blanket statement rather than each site needing a qualifier — cleaner than what I asked for, and it also covers `D-V2-1..3`, `RAD-07`, `STATE-02`, `ICON-03`.

**3. Foreground-alpha overstatement — FIXED.** The failure-mode row now reads "`bg-foreground/<alpha>` at ≤ 10%", which matches `check_appearance_scaling.py:490` (`foreground_alpha_percent(...) <= 10`). Codex's literal `/15` is no longer implicitly promised as gated.

**4. Program framing — FIXED.** Adversarial review → "swept the census … findings had to reach zero"; rendered-visual → "the served build was inspected"; "all 756 census sites were migrated." All three now read as a completed pass rather than work in flight.

**5. Checkboxes — FIXED.** All four `- [ ]` entries in Current Gaps are plain bullets; zero `- [ ]` remain in the file.

**6. Heading case — FIXED.** All seven headings Title-Cased (`Where The System Lives`, `Design Philosophy`, `Clean-Redo Laws`, `Ruled Retune Decisions`, `Change-Control Model`, `Failure Modes`, `Current Gaps`) — matches the `platforms/product/*` siblings. Zero sentence-case `##` remain. The `Ruled retune decisions that define the system's look` → `Ruled Retune Decisions` shortening is an improvement.

**7. `--radius-3xl` — FIXED.** Now "20px, Codex's own `--radius-3xl`", so no reader greps our `tokens.ts` for a token that stops at `--radius-2xl`.

`python3 scripts/check_docs.py` green on the new head (224 files). The diff touches nothing outside the spec — no token, gate, or script values changed, so my full numeric verification at 5dfe8dc07 (type ramp, composer, transcript, tab/icon tiers, gate/hook/Makefile/build-order claims) still holds unchanged.

One residual nit, non-blocking and explicitly not a finding: the escalation-convention bullet still reads "zero tags must remain at completion" in the present, and the tree already has 0 `ui-foundation-escalation:` tags. It parses fine as a standing rule for future reference-alignment passes, which is how the sentence around it is framed. Leave or tidy at your discretion.

All 7 findings resolved. Nothing further from me.

MERGE_OK

PR #1498, opened 2026-07-26
feat(server): litellm access-group tags per harness_kind (B1)

pablonyx:
Track B, PR 1 of 3 (B1→B2→B3, agents-impl-plan.md §4). Stacked-PR program;
this PR is the root (base `main`).

## Spec section quoted (model-gateway.md, Config laws)

> Every entry carries `model_info: {access_groups: [...]}` naming the
> harness group(s) it belongs to. Group names are exactly the harness
> `harness_kind` identifiers (`claude`, `codex`, `opencode`, `cursor`,
> `grok`) — no translation table... This one reviewed file is therefore
> also the harness-to-model map; no client-side model filtering exists
> anywhere.

Ruling R1 (agents-impl-plan.md §1): access-group names = harness_kind
exactly. Gateway-capable harnesses get their group tag; `cursor` stays
native-only (no gateway recipe) and is never granted a group.

## What changed

- `server/litellm/config.yaml`: every one of the 23 `model_list` entries now
  carries `model_info.access_groups`.
  - `claude` (direct-Anthropic + Bedrock Anthropic ids)
  - `codex` (direct-OpenAI + Bedrock OpenAI-open-weight ids)
  - `grok` (xAI ids)
  - `opencode` composes additively across every family — its catalog
    `gatewayPolicy.providers` carries no restriction today
    (`catalogs/agents/catalog.json`), so this preserves the current
    client-side-filter behavior (empty `providers` == all) 1:1 rather than
    narrowing opencode's model set as a side effect of this PR.
  - `cursor` appears in **zero** access_groups — it has no gateway route
    (agent-auth.md selection-cardinality law: cursor's only source can be
    `api_key`, never `gateway`).
- `specs/codebase/platforms/product/model-gateway.md`: deleted the
  "`config.yaml` entries carry no `access_groups` tags" gap bullet.

## Gap bullet closed

> `config.yaml` entries carry no `access_groups` tags.

## Tests

- Tier 1 (new): `server/tests/unit/test_litellm_config_access_groups.py` —
  static contract test over the reviewed YAML (no LiteLLM proxy involved):
  every entry carries non-empty `access_groups`; group names are exactly a
  member of `AGENT_AUTH_HARNESS_KINDS`; `cursor` never appears in any
  entry's groups; every supported harness_kind has at least one model.
  `cd server && DEBUG=true uv run pytest tests/unit/test_litellm_config_access_groups.py -q`
  → 4 passed.
- Regression: `cd server && DEBUG=true uv run pytest tests/unit tests/integration -q -k "agent_gateway or agent_auth or litellm_config"`
  → 193 passed (189 baseline + 4 new), 0 failed.
- `node --test scripts/ci-cd/litellm-image-pin.test.mjs` still green (pin
  test doesn't touch `model_list`, unaffected by this change).

## Not in scope (later PRs in this stack)

- B2 (next, base = this branch): per-(subject, harness) key minting so a
  virtual key actually receives exactly one access group instead of today's
  one-unscoped-key-per-subject.
- B3: renderer per-harness key map + org-member fix.
- Proxy-side scoped-key live verification (LiteLLM `/v1/models` returning
  exactly one group, 403 out-of-group) is the model-gateway.md "Proof"
  section's existing live check — already verified per that doc against
  v1.93.0; this PR doesn't re-run that, since it's a static config change
  with no running proxy in the merge gate.

## Contradictions

None — this PR followed the plan and R1 exactly. One judgment call: the
spec's harness-to-model-map description implies groups are the *only*
determinant of visibility, but no code currently mints group-scoped keys
(that's B2), so this PR only sets up the proxy-side data; it changes zero
runtime behavior until B2 lands (today's one-unscoped-key still sees every
model regardless of group tags, since LiteLLM only enforces group
membership against a key that itself carries a group grant).

### DIFF
--- a/server/litellm/config.yaml
+++ b/server/litellm/config.yaml
@@ -21,51 +21,84 @@
 # This file is the one deployed config (dev and prod both run it as-is): any
 # dev-only shim/mock (LiteLLM `mock_response`, cross-provider test aliases)
 # belongs in a docker-compose dev overlay, never here.
+#
+# Access groups (model-gateway.md, R1): every entry carries
+# `model_info: {access_groups: [...]}` naming the harness group(s) it
+# belongs to. Group names are exactly the harness `harness_kind` identifiers
+# (`claude`, `codex`, `opencode`, `grok`) — no translation table. `cursor` is
+# intentionally absent from every access_groups list: it has no gateway
+# recipe (native-only) until it gets one, so no virtual key may ever be
+# scoped to a `cursor` group. This file is therefore also the harness-to-
+# model map; no client-side model filtering exists anywhere.
 
 model_list:
-  # Anthropic
+  # Anthropic. access_groups: claude is the direct-Anthropic harness; opencode
+  # composes the gateway additively alongside its own provider auth (its
+  # catalog gatewayPolicy carries no provider restriction — parity-preserving
+  # with the Rust client-side filter this replaces, see model-gateway.md §R1).
   - model_name: claude-sonnet-4-5
     litellm_params:
       model: anthropic/claude-sonnet-4-5-20250929
       api_key: os.environ/ANTHROPIC_API_KEY
+    model_info:
+      access_groups: [claude, opencode]
   - model_name: claude-sonnet-4-5-20250929
     litellm_params:
       model: anthropic/claude-sonnet-4-5-20250929
       api_key: os.environ/ANTHROPIC_API_KEY
+    model_info:
+      access_groups: [claude, opencode]
   - model_name: claude-haiku-4-5
     litellm_params:
       model: anthropic/claude-haiku-4-5-20251001
       api_key: os.environ/ANTHROPIC_API_KEY
+    model_info:
+      access_groups: [claude, opencode]
   - model_name: claude-haiku-4-5-20251001
     litellm_params:
       model: anthropic/claude-haiku-4-5-20251001
       api_key: os.environ/ANTHROPIC_API_KEY
+    model_info:
+      access_groups: [claude, opencode]
   - model_name: claude-opus-4-6
     litellm_params:
       model: anthropic/claude-opus-4-6
       api_key: os.environ/ANTHROPIC_API_KEY
+    model_info:
+      access_groups: [claude, opencode]
   - model_name: claude-opus-4-6-20260205
     litellm_params:
       model: anthropic/claude-opus-4-6
       api_key: os.environ/ANTHROPIC_API_KEY
+    model_info:
+      access_groups: [claude, opencode]
 
-  # OpenAI
+  # OpenAI. access_groups: codex is the direct-OpenAI harness; opencode again
+  # composes additively (see above).
   - model_name: gpt-5.2
     litellm_params:
       model: openai/gpt-5.2
       api_key: os.environ/OPENAI_API_KEY
+    model_info:
+      access_groups: [codex, opencode]
   - model_name: gpt-5.2-2025-12-11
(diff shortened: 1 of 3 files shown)


pablonyx:
## QA fork — B1 verified correct on the merged tree; one cost question for Pablo

Checked on an integration branch = `origin/main` + a6 + b3 + c1 + d2. Read-only QA; nothing pushed to any `agents/*` branch. Ledger: `codex/ledger-qa.md` F-002, F-017.

**B1's artifact itself is correct**, and I checked it the skeptical way rather than trusting the diff: all 23 `model_list` entries in `server/litellm/config.yaml` carry `model_info.access_groups`; the union of groups is exactly `{claude, codex, grok, opencode}`; Anthropic/Bedrock-Anthropic → `[claude, opencode]`, OpenAI/Bedrock-OpenAI → `[codex, opencode]`, xAI → `[grok, opencode]`; and `cursor` appears nowhere (the two textual `cursor` mentions at `:28,:31` are comments documenting its deliberate absence). The access-group naming really is `harness_kind` with no translation table.

### The one thing to fix isn't in this PR's config — it's this PR's *test predicate*

`tests/unit/test_litellm_config_access_groups.py::test_every_supported_harness_has_at_least_one_model` is written over `AGENT_AUTH_HARNESS_KINDS`, with an inline comment asserting its own premise ("since cursor is excluded from that tuple already"). C1 (#1505) legitimately **adds** `cursor` to that tuple to enable cursor selections, and correctly gives cursor no access group. So the premise is false the moment C1 lands, and real pytest on the merged tree gives:

```
FAILED ...::test_every_supported_harness_has_at_least_one_model
1 failed, 3 passed in 222.64s
```

The **config is right; the predicate is wrong.** One-line fix: iterate `AGENT_AUTH_GATEWAY_CAPABLE_HARNESS_KINDS` (which C1 adds, = claude/codex/opencode/grok) instead of `AGENT_AUTH_HARNESS_KINDS`, and update the stale comments at `:5-6`, `:24-25`, `:71-72`. That is exactly the tuple whose members must each resolve a model — it's what the test's own comment says it means.

Note the sibling test `test_access_group_names_are_exactly_harness_kinds:51` uses the same tuple as its **allowed** set, so post-merge it silently *weakens* to permit a stray `cursor` access group in `config.yaml` — the very thing it was written to reject. Worth repointing both at the gateway-capable constant.

Whichever of B1/B2/C1 merges last should carry this; I've also flagged the related one-line `enrollment.py:46` constant fix on #1508, since the same tuple widening makes the mint loop try to create a `models=["cursor"]` key.

### Cost question for Pablo (not a defect — a decision)

The `claude` access group (13 entries) includes `us.anthropic.claude-fable-5` and `global.anthropic.claude-fable-5`, and `opencode` is granted all 23 entries. So any gateway-enrolled user can select the most expensive model in the deployment against the team budget. If that's intended, worth saying so explicitly in `model-gateway.md`; if not, Fable belongs in its own group. Flagging because it's a cost decision that the access-group design otherwise looks built to bound.

pablonyx:

```

none origin/main
```

```
```
```diff
-from proliferate.constants.agent_gateway import AGENT_AUTH_HARNESS_KINDS
+from proliferate.constants.agent_gateway import AGENT_AUTH_GATEWAY_CAPABLE_HARNESS_KINDS
-        missing = set(AGENT_AUTH_HARNESS_KINDS) - granted
+        missing = set(AGENT_AUTH_GATEWAY_CAPABLE_HARNESS_KINDS) - granted
```

PR #1503, opened 2026-07-26
fix(agents): credential-ladder + route-aware readiness (A2)

pablonyx:
Track A, PR 2 of the agents implementation program (Round 1). **Stacked on #1499** (`agents/a1-catalog-transport`); retarget to `main` once that merges. The diff below is A2's own commit only.

## The laws this implements

`specs/codebase/platforms/product/agent-distribution.md`, "Readiness projection" — three of its four "Projection laws, each closing a way the projection could lie":

> - An env credential counts only if the variable is set **and non-empty**; `ANTHROPIC_API_KEY=""` is absent, not present.
> - Credential detection reads only the workspace's composed env plus registry-declared variables — never the host process's ambient environment at large, so a global var on the machine cannot make every workspace look authenticated.
> - The settings read surface and the launch path resolve readiness the same way (route-aware); the UI never shows `CredentialsRequired` for a harness that would launch fine.

(The fourth — the `Ready` gate at every live-start — is A9's scope and stays in the gap list.)

## Gap bullets closed

The old single bullet listed four unenforced laws. It is replaced by two narrower bullets: one for the remaining `Ready`-gate law (A9), one for the two known inefficiencies that were bundled into it (claude's uncached Node gate, the `Archive`-only activation guard) — those are not correctness laws and were never part of this bullet's claim.

## What changed

**Law (a) — non-empty.** `CredentialEnvScope::provides` requires a value that is present AND non-empty after trim. Whitespace-only counts as absent: that is what a trailing `KEY=` in an env file parses to, and no provider accepts it.

**Law (b) — no ambient fallback for workspace scope.** This was the interesting one. The old code took an `additional_env: &BTreeMap` and did `additional_env.contains_key(var) || std::env::var(var).is_ok()`, with `resolve_agent_with_env` additionally branching on `additional_env.is_empty()`. So scope was *inferred from map emptiness*, and the inference was backwards: a workspace whose composed env happened to be empty silently inherited the host's variables — precisely the leak the law forbids.

Scope is now a type:

```rust
pub enum CredentialEnvScope<'a> {
    HostAmbient,                              // no workspace in view
    Workspace(&'a BTreeMap<String, String>),  // ONLY this map counts
}
```

`Workspace` never reads `std::env`, empty map or not. `HostAmbient` deliberately does — for a host-global read there is no workspace to over-authenticate, and the host env is what a spawn inherits.

**Law (d) — route-aware read.** `resolve_agent` (what `GET /v1/agents`, `GET /v1/agents/{kind}`, and the install response call) now layers the enrolled agent-auth route on, through the same `apply_launch_route_upgrade` the launch path uses. The two surfaces now differ *only* in env scope; the shared route layer is the mechanism that keeps them from disagreeing. The route still cannot clear `InstallRequired`/`Unsupported` (`route_credentials_upgrade_status` is unchanged and now guards both surfaces).

**New: `resolve_agent_unrouted`.** Three callers want the narrower question and would otherwise pay a state-file read per call for no effect on their decision:
- `auth/login.rs` — needs the native artifact path to build the login command. An enrolled gateway route must not suppress a user's ability to log in natively.
- `installer/reconcile/execution.rs` — the installed-only pass reads `source == "managed"`; a route cannot change whether a binary is managed-installed.
- `live/sessions/probe.rs` — reads artifact paths to build a launch env.

## Tests (all tier 1)

New file `credential_ladder_tests.rs` — the ladder matrix. Every rung x every env-value shape x both scopes, not only the cells that bit us:

| Test | Proves |
| --- | --- |
| `workspace_env_var_counts_only_when_non_empty` | law (a), workspace scope: real / `""` / `"   "` / `"\t\n"` / absent |
| `host_ambient_env_var_counts_only_when_non_empty` | law (a) is a property of the ladder, not of one caller |
| `a_host_exported_var_never_authenticates_a_workspace` | law (b) — the leak, at the exact site the old fallback fired (empty composed env), plus "empty workspace value must not fall through to the host's non-empty one" |
| `ladder_falls_through_env_to_discovery_then_login_then_missing` | rungs 2 / 2-expired / 3 / 4, and that an empty env value falls THROUGH to discovery rather than short-circuiting |
| `multi_var_slot_needs_one_non_empty_value` | the `.any()` is over non-empty, not over presence |
| `all_required_slots_rejects_an_empty_value_in_any_slot` | the per-slot rule holds before aggregation sees it |
| `provider_managed_and_none_policies_stay_ready_under_the_tightened_ladder` | the tightening does NOT turn opencode's structurally-always-`Ready` `provider_managed` policy into a gap (a stated boundary in the spec) |

New file `route_aware_read_tests.rs` — law (d), pinned as an *agreement* rather than as each surface's local behavior, because agreement is the property that regressed:

| Test | Proves |
| --- | --- |
| `the_settings_read_and_the_launch_path_agree_before_and_after_enrollment` | same status AND same credential state from both surfaces, both pre- and post-enrollment |
| `the_settings_read_only_honors_this_harnesss_route` | claude's route does not upgrade grok |
| `the_settings_read_never_masks_a_missing_binary` | the "a route supplies credentials, not binaries" fence guards the read too |
| `the_unrouted_projection_ignores_an_enrolled_route` | `resolve_agent_unrouted` still answers the login flow's question while routed |
| `workspace_scoped_resolution_ignores_a_host_exported_credential` | law (b) at the readiness layer, including that the host-scoped read legitimately still sees the export |

Results: `cargo test -p anyharness-lib --lib -- domains::agents` 298/298. `cargo test --workspace` **fully green** (1311 + 90 + all others, 0 failures) — see the flake note below. `check_docs.py`, `check_max_lines.py`, and every other `repo-shape` check pass.

### Test-hygiene fixes that ride along

Two separate process-env races, both real, both fixed here.

**1. `PATH` / `HOME` / agent-program overrides.** Adding parallel tests that mutate `HOME` exposed a pre-existing race: those variables are process-global, readiness reads all three, and the suites had no mutual exclusion. `PathEnvGuard` narrowing `PATH` to a temp dir made the installer's `git`/`npm` invocations fail with a bogus "not found", and `override_program_validation_requires_existing_executable` (which resolves bare `sh` through `PATH`) failed the same way.

**2. `ANYHARNESS_DATA_KEY`.** `domains::materialization::workspace_service_git_tests` builds an `AppState` — which parses that variable — while `app::tests::app_state_rejects_invalid_data_key` sets it to a deliberately invalid value for the length of its guard. Reproducible on a clean tree:

```
cargo test -p anyharness-lib --lib -- app::tests::app_state_rejects_invalid_data_key domains::materialization::workspace_service_git_tests
```

3 of those 4 tests failed with `InvalidDataKey` before this commit; all 4 pass after. (Those tests also shell out to `git`, so holding the lock is doubly correct.)

**Correction to an earlier claim in this PR:** the first revision of this body said the PATH/HOME fix "fixes A1's reported flake". That was wrong — the flake #1499 reported was race 2, and the PATH/HOME work did not touch it. Race 2 is fixed by *this* commit, not the previous one. Flagged rather than quietly reworded.

Every test that mutates or depends on a process-global variable now takes one shared `app::test_support::lock_env()` — crate-wide, because narrowing `PATH` or invalidating the data key breaks tests in unrelated modules, so module-local locks would not exclude the right things. It uses `.expect("expected env mutex")`, matching the crate's 88 existing `ENV_MUTEX` call sites; the three ad-hoc `into_inner()` helpers the first revision introduced are gone.

### MAJOR: unsurveyed `credentialState` consumers (resolved by restoring lost provenance)

Making `GET /v1/agents` route-aware collapsed two meanings into one wire value, and two clients were reading the wrong one:

**`lib/domain/agents/auth-onboarding.ts` — a real bug, now fixed.** `hasDetectedNativeAuth` treated `credentialState === "ready" && installState === "installed"` as "the vendor CLI is logged in". After the route-aware read, one gateway-routed harness reads `ready`, so `planFirstRunAuthAdoption` concluded "native creds detected" and returned `[]` — suppressing gateway preselection for **every** harness on the machine.

The fix is not a client-side heuristic patch. The projection lost information, so it now carries it: `ResolvedAgent.credentials_from_route`, written **only** by `apply_launch_route_upgrade`, surfaced as `AgentSummary.credentialsFromRoute`. `hasDetectedNativeAuth` excludes that case.

The field is absent-tolerant (`#[serde(default, skip_serializing_if = "Not::not")]`, optional in the TS type) and absent is read as `false`. That is correct rather than merely convenient: a runtime old enough to omit the field is a runtime whose `GET /v1/agents` was native-only, so `ready` there really did mean native.

Why a new field rather than a new `credentialState` variant: `credentialState` is a five-way enum consumed by several surfaces, and route-vs-native is orthogonal to it (a route-upgraded agent is genuinely ready). A widened enum would have forced every consumer to handle a new value; a defaulted boolean changes behavior only where a consumer opts in.

**`components/settings/sidebar/HarnessStatusDot.tsx` — deliberate keep, now pinned.** The dot is suppressed for `ready`, so route-upgraded harnesses stop showing an attention dot. That is the **desired** outcome and it is now stated as such in the code: the dot means "needs your attention", a gateway-routed harness with no vendor-CLI login launches fine, and flagging it would ask the user to fix a non-problem — the exact lie agent-distribution.md's route-aware law forbids ("the UI never shows `CredentialsRequired` for a harness that would launch fine"). The "authenticated how?" detail belongs in the harness auth pane, which shows the selected route. Pinned by a new `HarnessStatusDot.test.tsx` (5 cases: ready, route-ready, missing creds, failed install, pre-install/absent).

New tests for this major:
- Rust (3): `route_upgraded_readiness_is_distinguishable_from_native_readiness`, `a_natively_ready_agent_is_never_flagged_as_route_sourced` (an agent ready on its own credential must not be flagged even with a route enrolled), `a_route_that_cannot_upgrade_does_not_claim_provenance` (a route that cannot lift `InstallRequired` claims no credit).
- TS (5): two on `hasDetectedNativeAuth` (rejects route-sourced; treats absent flag as native), two on `planFirstRunAuthAdoption` (a route-ready harness no longer suppresses preselection; a genuine native login still does), plus the dot suite.

### MAJOR: `agent-auth.md` "Readiness interplay" was falsified by this PR

That section said readiness is computed from native credentials and only `resolve_launch_agent` asks route-auth. Rewritten to the new truth: one shared `apply_launch_route_upgrade` seam, both surfaces through it (that shared layer *is* the mechanism behind the route-aware law), differing only in env scope; `resolve_agent_unrouted` named with its three callers; and the `credentialsFromRoute` provenance contract documented.

### MINORS

- Three comments asserting the now-false "`GET /v1/agents` is native readiness" premise updated to describe the difference that actually remains — env scope (host vs workspace-composed), which is also why the launch-options union code they guard is still worth keeping: a workspace-scoped credential can make launch options list an agent the host-scoped read does not. Files: `use-home-next-model-selection.ts`, `target-ready-launch-agents.ts`, `local-world-smoke-1.ts`.
- Poison policy aligned (see above).
- `credentials.rs`'s `HostAmbient` justification now points at `live/sessions/driver/process.rs` — `.envs(&spawn_env)` with no `env_clear`, so "the host env is what a spawn inherits" is checkable rather than asserted.
- **Cost note, `list_agents`:** `resolve_agent` now loads `agent-auth/state.json` per agent, so `GET /v1/agents` does N (currently 5) state-file reads per call instead of 0. Accepted: the file is small, local, and already read on every launch, and readiness is deliberately uncached ("recomputed fresh from disk on every read — no cache, so it can never be stale"). If the desktop's polling makes this show up, the fix is one load per request passed into the loop, not a cache.
- `scripts/max_lines_allowlist.txt`: `model.rs` was already exactly at the 600 cap, so the 5-line field pushed it over; allowlisted at 605 with a split-on-next-growth note rather than restructuring a shared type barrel mid-review-cycle.

### Tests after the fixes

`cargo test --workspace` green (1314 anyharness-lib + 90 tauri + rest, 0 failures). `cargo test -p anyharness-lib --lib -- domains::agents` 301/301. product-client vitest: 19/19 on the two touched suites; 127/127 across `src/lib/domain/agents`, `src/components/settings/sidebar`, `src/hooks/home` (two collect-time failures in that broader run are a pre-existing self-referential package import in `CoworkThreadLaunchProvider.tsx`, reproduced identically on a stashed tree). product-client + tests/release typecheck clean. SDK openapi regenerated. All `repo-shape` checks pass.

## Contradictions encountered

None ruling-worthy. Two judgment calls:

1. **`HostAmbient` reads the host env on purpose.** The spec's ambient law is scoped to "the workspace's composed env", and a strict reading might forbid ambient reads everywhere. But `launch_facts.rs` already documents the opposite requirement for classification (*"Native launches inherit the ambient process env, so restricting to the composed workspace env alone causes classification to diverge from launch reality"*), and a host-global settings read has no workspace to over-authenticate. Resolution: the law binds the WORKSPACE scope, which is what the type enforces; the host scope stays ambient and is named as such.
2. **Whitespace-only treated as absent.** The spec says "set and non-empty" and only names `""`. Extending to whitespace-only is a judgment call in the law's direction: `KEY= ` in an env file is the same operator mistake with the same failure mode. Logged rather than silently generalized.

### DIFF
--- a/anyharness/crates/anyharness-contract/src/v1/agents.rs
+++ b/anyharness/crates/anyharness-contract/src/v1/agents.rs
@@ -69,6 +69,17 @@ pub struct AgentSummary {
     pub native: Option<ArtifactStatus>,
     pub agent_process: ArtifactStatus,
     pub credential_state: AgentCredentialState,
+    /// True when the enrolled agent-auth route — not a credential detected on
+    /// this machine — is what makes `credentialState` read `ready`.
+    ///
+    /// Readiness is route-aware on every surface (agent-distribution.md's
+    /// route-aware law: settings and launch must agree), so `ready` alone no
+    /// longer means "the vendor CLI is logged in here". A client that means the
+    /// latter — first-run native-auth adoption, CLI login chrome — must exclude
+    /// the route-upgraded case. Absent on older runtimes; treat absent as
+    /// `false` (the pre-route-aware meaning, which is what those runtimes had).
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub credentials_from_route: bool,
     pub readiness: AgentReadinessState,
     pub supports_login: bool,
     pub expected_env_vars: Vec<String>,
--- a/anyharness/crates/anyharness-lib/src/api/http/agents_contract.rs
+++ b/anyharness/crates/anyharness-lib/src/api/http/agents_contract.rs
@@ -255,6 +255,7 @@ pub(super) fn to_summary(
         native: resolved.native.as_ref().map(to_artifact_status),
         agent_process: to_artifact_status(&resolved.agent_process),
         credential_state,
+        credentials_from_route: resolved.credentials_from_route,
         readiness,
         supports_login: desc.auth.supports_login(),
         expected_env_vars: desc.auth.expected_env_vars(),
--- a/anyharness/crates/anyharness-lib/src/app/test_support.rs
+++ b/anyharness/crates/anyharness-lib/src/app/test_support.rs
@@ -31,6 +31,24 @@ pub(crate) fn actor_capabilities_for_store(store: &SessionStore) -> ActorCapabil
 
 pub(crate) static ENV_MUTEX: OnceLock<Mutex<()>> = OnceLock::new();
 
+/// Take the crate-wide process-environment lock for the length of a test body.
+///
+/// Every test that mutates or depends on a process-global variable —
+/// `ANYHARNESS_BEARER_TOKEN`, `ANYHARNESS_DATA_KEY`, `PATH`, `HOME`, the
+/// `ANYHARNESS_*_AGENT_PROGRAM` overrides — must hold this, and it has to be ONE
+/// lock crate-wide: narrowing `PATH` to a temp dir breaks any test in the crate
+/// that shells out, and an invalid `ANYHARNESS_DATA_KEY` breaks any test that
+/// builds an `AppState`. Module-local locks would not exclude those.
+///
+/// `.expect` on poisoning matches the crate's existing 88 call sites: a poisoned
+/// lock means another test already panicked, and the run is failing either way.
+pub(crate) fn lock_env() -> std::sync::MutexGuard<'static, ()> {
+    ENV_MUTEX
+        .get_or_init(|| Mutex::new(()))
+        .lock()
+        .expect("expected env mutex")
+}
+
 pub(crate) struct BearerTokenEnvGuard {
     previous: Option<OsString>,
 }
--- a/anyharness/crates/anyharness-lib/src/domains/agents/auth/credential_ladder_tests.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/auth/credential_ladder_tests.rs
@@ -30,15 +30,9 @@ use std::collections::BTreeMap;
 
 /// Serializes the tests below. They set and clear process-global variables, and
 /// the host-scope cases read whatever the process env holds, so an interleaving
-/// with any other env-mutating suite is a false result. The crate-wide mutex is
-/// the same one readiness's `test_env_guards` takes. Poisoning is ignored: a
-/// panicking test has already failed.
-fn lock_env() -> std::sync::MutexGuard<'static, ()> {
-    crate::app::test_support::ENV_MUTEX
-        .get_or_init(|| std::sync::Mutex::new(()))
-        .lock()
-        .unwrap_or_else(|poisoned| poisoned.into_inner())
-}
+/// with any other env-mutating suite is a false result. This is the crate-wide
+/// lock every such test takes.
+use crate::app::test_support::lock_env;
 
 /// Set or remove a var for the test's lifetime and restore it on drop, so the
 /// matrix below is deterministic on a developer machine that exports provider
--- a/anyharness/crates/anyharness-lib/src/domains/agents/auth/credential_ladder_tests.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/auth/credential_ladder_tests.rs
@@ -1,0 +1,398 @@
+//! The credential ladder matrix.
+//!
+//! Two projection laws from agent-distribution.md, "Readiness projection":
+//!
+//!   (a) "An env credential counts only if the variable is set **and non-empty**;
+//!       `ANTHROPIC_API_KEY=\"\"` is absent, not present."
+//!   (b) "Credential detection reads only the workspace's composed env plus
+//!       registry-declared variables — never the host process's ambient
+//!       environment at large, so a global var on the machine cannot make every
+//!       workspace look authenticated."
+//!
+//! Each slot's ladder has four rungs — declared env in scope, provider-specific
+//! local discovery, login policy, missing — and the matrix below enumerates every
+//! rung against every env-value shape and both scopes, rather than only the cells
+//! that happen to have bitten us.
+//!
+//! Split from `credentials_tests.rs` to stay under the repo line-count ceiling;
+//! nested inside it so the `make_temp_home`/`test_login_spec` helpers and the
+//! module-under-test's items are both in scope.
+
+// The parent test module's helpers (`make_temp_home`, `test_login_spec`).
+use super::*;
+// The module under test plus the model types its signatures use, which the
+// parent reaches through its own `use super::*`.
+use super::super::{
+    detect_auth_slots_with_env, detect_credentials, detect_credentials_with_env,
+    AuthReadinessPolicy, AuthSlotSpec, AuthSpec, CredentialDiscoveryKind, CredentialState,
+};
+use std::collections::BTreeMap;
+
+/// Serializes the tests below. They set and clear process-global variables, and
+/// the host-scope cases read whatever the process env holds, so an interleaving
+/// with any other env-mutating suite is a false result. The crate-wide mutex is
+/// the same one readiness's `test_env_guards` takes. Poisoning is ignored: a
+/// panicking test has already failed.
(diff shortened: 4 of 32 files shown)


pablonyx:
## QA fork — A2 finding: a registry `kind:"flag"` var counts as a credential → false Ready

Reviewed on a merged integration branch (`origin/main` + a6 + b3 + c1 + d2). Read-only QA; nothing pushed to any `agents/*` branch. Detail in `codex/ledger-qa.md` (F-020).

`registry/projection.rs:66-68` flattens `slot.env_vars` with `env_var.name().to_string()`, **discarding the `secret` | `flag` distinction** that `schema.rs:196-201` exists to carry — and that `launch_facts.rs:151` correctly honors. `credentials.rs:159` then treats any listed var as a credential via `scope.provides`.

Repro: a workspace env contains `CLAUDE_CODE_USE_BEDROCK=0` (or `=1`, or `AWS_BEARER_TOKEN_BEDROCK=x`) and **no** Anthropic key. The registry claude slot env is `[ANTHROPIC_AUTH_TOKEN, ANTHROPIC_API_KEY, {CLAUDE_CODE_USE_BEDROCK, flag}, AWS_BEARER_TOKEN_BEDROCK]`. `detect_slot_credentials` → Ready → claude passes the create gate → spawn with **zero usable credential**. A var whose value literally says "Bedrock OFF" authenticates the harness.

A2 tightened the empty-vs-present rung correctly; this rung is still wrong. The fix is to preserve the `flag` discriminant through the projection so `scope.provides` only counts secret-kind vars — `launch_facts.rs:151` already shows the intended shape.

Severity: major — it's a false Ready that produces a launch with no credential, which is the same user-visible dead-end class as the fail-closed findings I filed on #1511 (there the launcher at least refuses with a 409; here the spawn proceeds).

pablonyx:

---

PR #1505, opened 2026-07-26
feat(agent-auth): enable cloud native login + cursor api_key auth (C1)

pablonyx:
## Summary

Track C, round 1, PR C1 (agents-impl-plan.md §4). Closes two `Current gaps` bullets in `specs/codebase/platforms/product/agent-auth.md`:

> **Cursor selections are rejected server-side.** ... the `api_key` source needs enabling for cursor end to end (rules, allow-list, recipe already generic).

> **Cloud native login is not offered.** The cloud settings surface shows static "no auth configured" text instead of the Authenticate action, though the login-terminal mechanism is surface-agnostic; wiring it up also revisits agent-distribution's cursor-in-cloud carve-out, which assumed no headless credential path.

Both bullets are struck; `agent-distribution.md`'s cursor-in-cloud carve-out text is updated to the new truth (see review fix cycle below), not left stale.

## Changes

**Cloud native login**
- `HarnessAuthCliDetails.tsx`: removed the `surface === "cloud"` early return. Both surfaces share the same login-terminal render path.
- `use-agent-login-terminal-workflow.ts`: hook takes `surface`; resolves a surface-aware runtime connection (`AnyHarnessRuntimeContext`'s own value on cloud — already the sandbox gateway via `CloudAnyHarnessRuntimeProvider`; the harness-connection-store fallback on local). Mints a fresh Cloud sandbox-gateway access token per terminal open/restart via `host.cloud.getSandboxGatewayAccessToken()`, since the WS terminal viewport can't ride the context's fetch-wrapping bearer-refresh trick.
- **Review fix**: threaded `webSocketAuthTransport: "protocol"` through `runtimeConnection` → `AgentLoginTerminalPanel` → the viewport's `connectAgentLoginTerminal` call, so cloud's 7-day product JWT rides the WS subprotocol (matching every other cloud WS path) instead of the `?access_token=` query string.
- **Review fix**: the periodic 2.5s readiness poll (whose only job is catching the local CLI writing its own credential file mid-login) is now local-only — cloud's `runtimeReady` is a coarse "client configured" check, not a live sandbox-health signal, so polling on it could wake a cold sandbox via `ensure_cloud_sandbox_gateway_access` purely to poll. The start mutation's own catch surfaces a dead sandbox as `errorMessage`; close/exit still force one refresh.
- No server-side change needed for the WS path: `proxy_websocket_to_anyharness` already forwards any path generically into the user's sandbox (verified: crafted requests cannot bypass gateway containment).

**Cursor UI enablement**
- `HarnessAuthSection.tsx`: removed the cursor native-only short-circuit. Cursor renders the normal method-card grid with the gateway card omitted (`isGatewayCapableHarness`, mirroring the server's positive `GATEWAY_CAPABLE_HARNESS_KINDS` allow-list). `api_key`/`cli` behave as an ordinary single-source radio.
- **Review fix (B1)**: cursor's harness pane was unreachable in the shipped UI — added `agent-cursor` to `SETTINGS_HARNESS_SECTIONS`, `SETTINGS_CONTENT_SECTIONS`/`SHORTCUT_SECTION_ORDER`, the Agents scope nav, and the sidebar icon map. `getSettingsSectionForHarnessKind("cursor")` now resolves; regression test added.
- `selection_rules.py` / `constants/agent_gateway.py`: cursor added to `AGENT_AUTH_HARNESS_KINDS`; `GATEWAY_CAPABLE_HARNESS_KINDS` promoted to a shared constant (`AGENT_AUTH_GATEWAY_CAPABLE_HARNESS_KINDS`) in `constants/` so `db/store` can consult it without violating the store→server import boundary; `NATIVE_ONLY_HARNESSES` deleted.
- **Review fix (m7)**: the store no longer injects the disabled gateway revision-marker row for a harness that has no gateway recipe (cursor) — it was dead state with a now-false comment.
- No Rust changes: `render.rs`'s `ApiKey` source path is fully generic; cursor's `Gateway` arm is a typed refusal — exactly the plan's "recipe already generic" note.

**Spec truth (review fix M5)**
- `agent-distribution.md`'s cursor-in-cloud carve-out no longer claims "no headless credential path" (false now — `CURSOR_API_KEY` is a legal selection that upgrades readiness via `launch_route_provides_credentials`, harness-agnostically). It now names the real reason cursor still can't install in cloud: `cursor-agent`'s ACP process ignores `CURSOR_API_KEY` at runtime and reads its login from macOS Keychain (`catalog_probe.rs`'s cursor-login/cursor-api arms), which no headless Linux sandbox has.

## Adversarial review (cycle 1) — resolved

- **B1** cursor nav unreachable → fixed (see above), regression test added.
- **B2** repo-shape red (max-lines) + missing `area:docs` → allowlist updated to exact grown line counts; label added.
- **B3** WS bearer token in query string → fixed, threaded `webSocketAuthTransport`.
- **M4** cloud-login spec bullet had zero automated coverage → added `use-agent-login-terminal-workflow.test.tsx` (6 cases: local/cloud baseUrl+authToken resolution, per-open token minting, runtimeReady gating both surfaces, cloud-vs-local poll split).
- **M5** stale "no headless credential path" claim → spec text corrected to the real (Keychain) reason.
- **M6** tautological cloud `runtimeReady` driving a poll that could wake a sandbox → poll disabled on cloud, decision documented inline.
- **m7–m11** → shared gateway-capable constant + skip marker row; positive-allow-list `isGatewayCapableHarness`; dropped redundant `cloudBaseUrl` re-derivation; clarified token-reuse comment; test hygiene (dropped unused fixture, destructure instead of `next()`-skip, added cloud-surface cursor case, fixed stale constant-name reference in `tests/release`).

## Contradiction encountered (resolved, judgment call)

Plan said the `NATIVE_ONLY_HARNESSES`/constants change "rides here… unless already handled." It was **not** already handled — cursor was hard-blocked at both the store level (`AGENT_AUTH_HARNESS_KINDS`) and validator level (`NATIVE_ONLY_HARNESSES`). Resolved per the plan's own fallback instruction; logged in the ledger.

## Tests

- **Tier 2 (unit, Python)**: `test_agent_gateway_domain.py` — cursor gateway-rejected/api_key-accepted + cardinality. 50/50 `agent_gateway`-scoped unit tests pass.
- **Tier 2 (unit, TS)**: `harness-auth-sources.test.ts`, `harness-env-vars.test.ts`, `navigation.test.ts`, `navigation-presentation.test.ts`, and the new `use-agent-login-terminal-workflow.test.tsx` (6 cases).
- **Tier 3 (integration)**: `test_agent_gateway_api.py` — cursor gateway-rejected, api_key-accepted on **both** local and cloud surfaces (full PUT round-trip against a real DB, asserting the exact single-row response — no gateway marker leak). 45/45 `agent_gateway` integration+unit pass.
- **Component**: `HarnessPane.test.tsx` — cursor offers `api_key`/`cli`, never `gateway`; cursor's api_key flow persists via `CURSOR_API_KEY`. Full product-client suite: **3835/3835 passed**.
- `tsc -p tsconfig.json --noEmit` clean; `ruff check`/`ruff format --check` clean; `mypy` clean (2 pre-existing, unrelated `selections.py` findings confirmed present on the prior commit too — not introduced here).
- **Full repo-shape script suite** (max-lines, migration-heads, anyharness-old-paths, worker-structure, frontend-boundaries, appearance-scaling, frontend-structure-drift, server-boundaries, anyharness-boundaries, session-mutation-admission): all green.

No T3-AGENT-1/T4 gate is implicated (Gate 1 depends on A2+B3, not C1).

## Test plan
- [x] `pnpm --filter @proliferate/product-client run typecheck`
- [x] `pnpm --filter @proliferate/product-client exec vitest run` (3835 passed)
- [x] `cd server && DEBUG=true uv run pytest tests/unit -k agent_gateway tests/integration/test_agent_gateway_api.py tests/integration/test_agent_gateway_store.py tests/integration/test_agent_gateway_selections.py` (95 passed)
- [x] `cd server && uv run ruff check/format --check` + `mypy` on changed files
- [x] Full repo-shape script suite (see above)
- [ ] Pablo: manual smoke of cloud "Authenticate" button against a live sandbox (needs a real cloud workspace + browser — not run here)

## Review fix cycle 2 (resolved)

- **Reverted m7's gateway-marker skip.** Cycle 1's "m7" change skipped inserting
  the disabled gateway revision-marker row for non-gateway-capable harnesses
  (cursor), on the premise that the row was dead state for a harness that can
  never select gateway. That premise was wrong: the marker's job is keeping the
  scope's rendered revision (`max(updated_at)` across ALL rows —
  `materialize/agent_auth.py`) monotonic, and that job is harness-agnostic.
  Without the marker, deleting/replacing a cursor `api_key` row could move the
  revision backwards, and `route_auth/state.rs` correctly rejects the push as
  `AGENT_ROUTE_STATE_STALE` — stranding the user's `CURSOR_API_KEY` in the
  runtime's prior state file. Reverted to unconditional marker insertion (as on
  `main`), with a corrected comment (surface-revision monotonicity, not
  "every harness here is gateway-capable" and not "skip for non-gateway
  harnesses"). Restored the marker-row assertions in
  `test_cursor_accepts_api_key_source[_on_cloud_surface]` — the api_key row is
  now selected by filtering rather than exact-unpacking, and both tests assert
  the disabled gateway marker row is present (`enabled is False`).
- **New honest gap bullet.** `agent-auth.md`'s "Current gaps" section now states
  the readiness/reality mismatch this PR ships: cursor's `api_key` source is
  persisted and rendered, readiness reports `Ready` for it (the check is
  kind-agnostic — `readiness/service.rs`, and the catalog's `cursor-login`
  auth context's `anyOf` signal includes the `CURSOR_API_KEY` env var), but
  `cursor-agent`'s ACP process ignores `CURSOR_API_KEY` at runtime and requires
  macOS Keychain auth (`catalog_probe.rs`) — so a user who selects "API key"
  sees Ready and the session fails at runtime. Resolution needs either an
  upstream `cursor-agent` change or dropping the api_key card for cursor
  (founder ruling).
- Trivial: dropped the dead `cloudSandboxGatewayRuntimeUrl` mock (unused by the
  hook) and fixed its test's misleading title in
  `use-agent-login-terminal-workflow.test.tsx`; removed a stray double blank
  line before "Install topology" in `agent-distribution.md`.

New head: `ea56b5823`.


pablonyx:
## QA fork — BLOCKER (cloud cursor) + the other half of the #1508 merge collision

Findings from an integration branch = `origin/main` + a6 + b3 + **c1 (this PR)** + d2. Read-only QA; nothing pushed to any `agents/*` branch. Full detail in `codex/ledger-qa.md` (F-007, F-029…F-034).

### BLOCKER — cloud Cursor is a permanent install loop that also buries this PR's own cloud-login UI

Pure clicks, no setup: **Settings → Agents → surface `Cloud` → Cursor.** You don't get the auth pane this PR builds. You get a full-page empty state *"Install Cursor / Install Cursor and its Proliferate adapter in Proliferate Cloud."* Clicking it produces *"Updating Cursor in Proliferate Cloud."* then **"Agent tools updated — the installed harnesses and ACP adapters are ready."** Reload: still "Install Cursor." Forever.

Chain, every hop verified:
- `settings/navigation-presentation.ts:146,164` + `config/settings.ts:23,59` — the B1 fix makes `agent-cursor` reachable on **both** surfaces; there is no surface filter anywhere in the nav builder.
- `runtime.rs:138-149` — `list_agents` maps `built_in_registry()` with no surface filter, so cursor is in the cloud agent list.
- `readiness/compatibility.rs:20-22` — `detect_runtime_compatibility_issue` returns `None` unless `kind == Claude`, and `status.rs:14-16` only yields `Unsupported` when that is `Some` ⇒ **cursor can never be `unsupported`**; in a cloud sandbox it is `install_required`.
- `domain/agents/status.ts:11-18` → `issueAgent` always set ⇒ `HarnessPane.tsx:80-81,93-103` returns `<HarnessInstallGate>`, **replacing the entire pane** — the auth-method cards, CLI-login block and api_key block all unreachable.
- `use-harness-install-action.ts:30-46` fires a scoped reinstall ⇒ `installer/auto_install.rs:104-106` hits `AutoInstallSkip::CursorUnsupportedInCloud` unconditionally (evaluated for reinstall passes too, `execution.rs:330-345`) ⇒ `Skipped`.
- `agents_contract.rs:289-318` — `to_install_state` only reports `Failed` when the last outcome is `Failed`; a **`Skipped`** agent stays `install_required` forever.
- `HarnessUpdateToastPresenter.tsx:103-115` — a skipped-only job is `completed`, so the **success** toast fires.

Why it matters: `specs/codebase/platforms/product/agent-distribution.md:167` says cursor × Cloud = "Not supported in cloud" (and `:149-159` says cursor never installs in cloud); cursor is also absent from `cloud/sdk/src/types/generated.ts:119` and `scripts/build-template.mjs:47`, so even a *successful* install would be unusable. It's a lying success toast on a user-initiated action, and it hides this PR's own feature. Fix wants a surface guard — drop `agent-cursor` from cloud nav, or have the pane's gate skip agents the surface doesn't support. The client currently has no way to know, so a `surfaceSupported` bit on the wire is the honest fix.

This is the same defect family as a live one I confirmed on codex (`ledger-qa.md` F-028): both are `auto_install.rs` carve-outs whose `Skipped` outcome is indistinguishable from success while leaving the harness permanently `install_required`. Worth fixing together — a carve-out skip needs to be a first-class terminal outcome.

### BLOCKER — this PR's tuple widening turns a B-stack test red on merge

`AGENT_AUTH_HARNESS_KINDS` gains `cursor` here (correct for selection), but `enrollment.py:46` on the B-stack aliases that constant instead of `AGENT_AUTH_GATEWAY_CAPABLE_HARNESS_KINDS`, so the eager mint loop starts minting a `models=["cursor"]` key for an access group no model carries. On the merged tree, real pytest:

```
FAILED tests/unit/test_litellm_config_access_groups.py::TestLitellmConfigAccessGroups::test_every_supported_harness_has_at_least_one_model
1 failed, 3 passed in 222.64s
```

That test file is absent on this branch and on main — it ships on the B-stack — so neither PR's CI can see it. The one-line fix belongs in #1508 (commented there); flagging here because this PR is the other half. Nothing to change in this PR for it.

### MAJOR — cloud WS credential transport is fail-open (secure path is opt-in per call site)

`anyharness/sdk/src/streams/terminals.ts:74-90`: `useProtocolAuth` requires `options.webSocketAuthTransport === "protocol"`; otherwise the token goes into `?access_token=`. `webSocketAuthTransport` is optional at every hop (`terminals.ts:5`, `AgentLoginTerminalPanel.tsx:12`, `use-agent-login-terminal-viewport.ts:15`), so cloud correctness rests solely on `HarnessAuthCliDetails.tsx:127` threading it. Any future cloud call site that omits it silently puts the 7-day product JWT in a WS URL, and nothing rejects it — `gateway/access.py:56-59` checks `?access_token=` *first*. Also, if `authToken` is undefined the socket opens with no credential and the user sees only "Terminal connection interrupted." (`use-agent-login-terminal-viewport.ts:92-94`) — indistinguishable from a network blip. Suggest making the transport required whenever `authToken` is non-empty, and throwing when it's missing.

Blast radius is bounded and worth stating: I traced mint → transport → authn → authz → forward and **the login token cannot be turned against another user's sandbox.** Every socket authenticates independently and authorization resolves for the authenticated caller (`gateway/api.py:62-81`), there is no sandbox/workspace id for the proxy to trust (`api.py:29,52`), and `gateway/proxy.py:27-39,73,99-101,200-219` strips the client credential and substitutes the sandbox's own — so even the insecure transport does not leak the JWT to the sandbox. The exposure is the browser URL and anything that logs URLs.

### MAJOR — nothing refreshes *cloud* readiness after a cloud auth selection (≥30s stale window)

`cloud/sdk-react/src/hooks/agent-gateway.ts:143-154` — `usePutAuthSelections().onSuccess` invalidates only the selections/state keys, never an AnyHarness key. Local is covered (`use-local-auth-state-sync.ts:93-97` pushes state and calls `invalidateAgentLaunchReadinessResources`); cloud has no client twin because materialization is server-side fire-and-forget (`agent_gateway/service.py:250-251` → `materialization/service.py:64-73` `run_after_commit`). With `staleTime: 30_000` / `refetchOnWindowFocus: false` (`query-client.ts:163-167`) and no `refetchInterval` on `useAgentsQuery` (`anyharness/sdk-react/src/hooks/agents.ts:58-70`), the cloud agent list is pinned for 30s. Repro: Cloud surface → any harness → add/bind an API key; the selection card flips instantly but the runtime status badge (`HarnessPane.tsx:107-124`) and the "CLI not authenticated" line (`HarnessAuthCliDetails.tsx:77-80`) keep showing pre-selection readiness. Only escape is the manual refresh button (`HarnessAuthCliDetails.tsx:90-97`). Note a naive immediate invalidation would race the async materialization — needs a short poll or settle-then-invalidate.

### MINORs

1. `use-agent-login-terminal-workflow.ts:60` — `cloudAuthToken` is set on every cloud open (`:170-172`) and never reset by `closeAuthTerminal` (`:207-216`), `handleTerminalExit` (`:218-241`), or unmount; the credential outlives every close. One line in `closeAuthTerminal`.
2. `:77-79` — `webSocketAuthTransport: (...) as "protocol"|undefined`. Annotating the memo's return type gets the same result with compiler checking, and this is the one place a typo silently re-enables the query-string transport above.
3. `HarnessAuthCliDetails.tsx:85-89` now renders `HARNESS_PANE_COPY.nativeStateLocal` ("No auth configured — the CLI's own login is used.") on **both** surfaces after the cloud early-return was deleted. Behavior may be intended, but the key name is now wrong and every neighbouring string is surface-parameterized. Rename to `nativeState` or parameterize it.

### Verified correct (so you don't re-check)

Cursor genuinely can't select the gateway **on the server**, not just in the UI (`selection_rules.py:29` is the hard gate; `SINGLE_SOURCE_HARNESSES:34` includes cursor so its radio holds at most one enabled source), and the UI's omit-the-card decision (`HarnessAuthSection.tsx:140,187-204`) mirrors that allow-list rather than special-casing a string, with the grid correctly collapsing 3-col → 2-col. The disabled gateway marker row *is* inserted for cursor (`selections.py:130-136`), so surface-revision monotonicity holds and a cleared selection can't be rejected `AGENT_ROUTE_STATE_STALE`. No orphaned copy after the deletions (`nativeStateCloud`, `cursorNativeDescription` — zero remaining references). `report_frontend_structure.py --strict` (2208 files), `check_frontend_boundaries.py`, `check_docs.py` all pass.

pablonyx:

```

none origin/main
```

```
```
```diff
-from proliferate.constants.agent_gateway import AGENT_AUTH_HARNESS_KINDS
+from proliferate.constants.agent_gateway import AGENT_AUTH_GATEWAY_CAPABLE_HARNESS_KINDS
-        missing = set(AGENT_AUTH_HARNESS_KINDS) - granted
+        missing = set(AGENT_AUTH_GATEWAY_CAPABLE_HARNESS_KINDS) - granted
```

pablonyx:

```
```

```

```

```
```

PR #1508, opened 2026-07-26
feat(server): per-(subject,harness) gateway key minting (B2)

pablonyx:
Track B, PR 2 of 3 (B1→**B2**→B3, agents-impl-plan.md §4). Stacked; base =
`agents/b1-access-group-tags` (#1498).

## Spec section quoted (model-gateway.md, Account model)

> One LiteLLM team per billing subject; the budget lives on the team and
> mirrors the subject's remaining credit. Inside the team, one virtual key
> per (subject, harness), each granted its harness's access group by name
> (`{"models": ["claude"]}` at `/key/generate`)... key: (subject, harness) —
> access: group grant and per-harness spend attribution; never a budget.

Ruling R2 (agents-impl-plan.md §1): eager minting — all gateway-capable
harness keys minted per subject at enrollment. Rotation via `/key/update`
or delete+mint (never `/key/regenerate` — Enterprise-gated, 500; matched by
message not status).

## What changed

- New child table `agent_gateway_enrollment_key` (migration
  `28adf1a9e376`, `down_revision=35fa0038d703`): one row per
  `(enrollment_id, harness_kind)`, unique-active-scope index, its own
  Fernet-encrypted `virtual_key_ciphertext` + `sync_fingerprint`.
- `enrollment.py._sync_enrollment` now loops over
  `AGENT_AUTH_HARNESS_KINDS` (claude/codex/opencode/grok) and mints one
  `{"models": [harness_kind]}`-scoped key per harness via
  `_sync_one_harness_key`, idempotent per (enrollment, harness) — a key
  already present for a harness is never re-minted. Alias:
  `vk-{subject}-{harness}-{id[:8]}`.
- The parent `agent_gateway_enrollment` row no longer carries its own key
  material (`virtual_key_id`/`virtual_key_ciphertext` cleared to `None` on
  every sync) — keys live exclusively on the child table now.
- `integrations/litellm/client.py`: `mint_virtual_key`/`rotate_virtual_key`
  gained a `models: list[str] | None` param forwarded to `/key/generate`'s
  `models` field; omitted mints an unscoped key (back-compat for any other
  caller).
- `topups.py`: reactivation now unblocks/re-mints **every** active child
  key for an enrollment and only ever rewrites the **team** budget — never
  a per-key budget (R2: keys never carry one).
- `usage_import.py`: spend-log `api_key` resolution now looks up the child
  key's token hash first (`get_enrollment_key_by_virtual_key_id`), falling
  back to the pre-B2 direct-on-enrollment lookup for a transition window;
  exhaustion/limit enforcement disables every active child key, not a
  single parent key.
- Migration backfill: flips every `sync_status='synced'` enrollment back to
  `pending` (never touching LiteLLM directly), so the existing backfill
  worker's `list_enrollments_needing_sync` pass re-syncs every enrollment
  through `_sync_enrollment` and mints the new per-harness keys on its next
  tick.

## Gap bullets closed

> Enrollment mints one unscoped key per subject (it sees all models)
> instead of per-harness group-scoped keys; existing enrollments need
> rotation at migration.

> Enrollment copies `max_budget` onto the virtual key as well as the team;
> keys must stop carrying budgets (the team cap already aggregates, and N
> per-key copies of the mirror would drift).

## Tests

- Tier 1 (new): `server/tests/unit/test_litellm_integration.py` —
  `test_mint_virtual_key_forwards_models_grant`,
  `test_mint_virtual_key_omits_models_when_not_given`, and the
  `rotate_virtual_key` test now asserts `models` forwards to the fresh mint.
- Tier 1/2 integration (new): `server/tests/integration/test_agent_gateway_enrollment_keys.py`
  (8 tests) — upsert idempotency, ciphertext preserved on a metadata-only
  resync, distinct per-harness rows coexist, scoped lookup, token-hash
  resolution, revoke cascade, parent-revoke cascades to children.
- Updated existing integration suites to the new per-harness model:
  `test_agent_gateway_enrollment.py`, `test_agent_gateway_topups.py`,
  `test_agent_gateway_topup_fixes.py`, `test_agent_gateway_usage_credits.py`,
  `test_billing_limit_enforcement_llm.py`,
  `test_agent_gateway_catalog_api.py`, `agent_gateway_topups_shared.py`
  (shared stub gained `models` param).
- Migration dry-run: fresh DB `alembic upgrade head` reaches single head
  `28adf1a9e376` cleanly; verified `agent_gateway_enrollment_key` table
  shape (FK cascade, unique-active-scope index) via `\d`.
- Full run: `cd server && DEBUG=true uv run pytest tests/unit tests/integration -q -k "agent_gateway or agent_auth or billing_limit_enforcement_llm or litellm"`
  → 225 passed, 0 failed.
- `uv run ruff check` clean on every touched file. `uv run mypy` on touched
  modules shows the same single pre-existing `select()` overload note also
  present on `origin/main` (unrelated to this PR, confirmed via `git stash`
  diff) — no new mypy errors introduced.

## Known interim gap — closed by B3 (next in this stack)

`materialize/agent_auth.py`'s `_load_state_inputs` and `budget.py`'s
`is_gateway_budget_available` still resolve the **parent** enrollment's own
key (`get_enrollment_virtual_key_decrypted(enrollment_id=...)`), which this
PR intentionally clears to `None`. Until B3 lands the per-harness key-map
renderer, `state.json`'s gateway source renders unsatisfiable (empty
harnesses) even for a freshly-synced, budget-available enrollment. This is
documented explicitly (not silently broken) in two tests in
`test_agent_gateway_usage_credits.py`
(`test_exhausted_budget_withholds_gateway_key_from_state_render`,
`test_available_budget_leaves_state_render_and_no_grant_unblocked`) with
inline comments naming the gap and the PR that closes it. No production
gateway traffic exists yet (pre-launch), so this is a stacked-PR sequencing
artifact, not a live regression.

## Contradictions

None beyond the documented interim gap above, which the plan's own B2/B3
split anticipates (B3 is explicitly "renderer key-map + org fix").

## Gate G1 finding + fix

The Gate-G1 executor ran the T3 release-test harness against this branch and
every gateway cell failed with:

```
No LiteLLM key resolved for the actor enrollment alias "vk-user-<uuid>-<enrollmentId8>".
```

**Root cause**: this PR changed the server's key-alias scheme from one
unscoped key per subject (`vk-{subject_label}-{enrollment_id[:8]}`) to one
scoped key per `(subject, harness)` (`vk-{subject_label}-{harness_kind}-
{enrollment_id[:8]}`, `enrollment.py` `_key_alias`), but
`tests/release/src/services/qualification-litellm.ts`'s
`deriveActorKeyAlias` still derived the old pre-B2 shape with no harness
segment. Every T3 fixture that resolves an actor's virtual key
(`authenticatedActor`, `invitedActor`, the managed-cloud actor-enrollment
custody recovery script, and the self-host `SH-GATEWAY` cell) inherited the
stale derivation and could never find the per-harness key the server
actually minted.

**Fix** (commit `78c943a0a` on `agents/b2-per-harness-keys`, rebased forward
into `agents/b3-renderer-key-map` at `7336162d1`):

- `deriveActorKeyAlias`/`derivePersonalActorKeyAlias`
  (`qualification-litellm.ts`) now take a `harnessKind` and match the exact
  server format. `QualificationLiteLlmController.resolveActorKey` requires
  `harnessKind` in its params.
- Threaded `harnessKind` through `authenticatedActor`/`invitedActor`'s
  enrollment-custody seams (`beginActorEnrollmentCustody`/
  `resolveAndTrackActorSubjects`) and the managed-cloud world's
  `resolveAndTrackActorSubjects`/`bindActorEnrollment`.
- The managed-cloud actor-enrollment-custody recovery script now enumerates
  the child `agent_gateway_enrollment_key` rows per enrollment (one alias per
  `(enrollment, harness)`) instead of synthesizing a single pre-B2 alias per
  enrollment.
- Self-host's `selectPersonalEnrollmentKeyToken`/`litellmResolveActorKeyToken`
  now match the exact personal+harness alias (falling back to any personal
  alias, then any non-org token) instead of matching only the `vk-user-`
  prefix, which could previously pick an arbitrary sibling harness's key.
- Added an offline unit test in `qualification-litellm.test.ts` pinning the
  exact alias format across every `GATEWAY_CAPABLE_HARNESS_KINDS` entry —
  drift protection so a future server-side alias change fails this suite in
  CI, not at a live gate.

Verified offline: `cd tests/release && pnpm test` → 1552/1552 passed;
`pnpm run typecheck` clean. No server-side code changed by this fix — it is
entirely within `tests/release`.

---

## Adversarial review fixes (head `f5cdb0558`)

Four money-path / honesty findings and two CI blockers, all addressed on this
branch.

### 1. Exhaustion enforcement failed OPEN on an empty key list (money)

`usage_import._disable_enrollment_keys` started from `all_disabled = True` and
never revisited that on an empty iteration, so an enrollment with **no**
disablable child key returned success. The caller then set
`budget_status='exhausted'` having disabled nothing — and because both callers
skip already-flipped enrollments, that suppressed enforcement on every later
tick. Any key minted moments later (mid-sync enrollment, or a row written after
this tick read it) stayed live and billable with no retry.

Now it returns `True` only when at least one key was actually disabled and none
failed; "nothing enforced" logs and leaves the enrollment in the retry set.
Test: `test_enrollment_without_child_keys_is_not_marked_exhausted` (stays `ok`,
nothing disabled, then a later tick with a key present does flip it).

### 2. The pre-B2 unscoped key was forgotten, not revoked (money)

The migration flips synced rows to `pending`, and `_sync_enrollment` then
cleared the parent row's `virtual_key_id`/ciphertext. Clearing the row is not
revocation: the old **unscoped** key stayed live on the proxy with all-model
access, honoring any client that already held it, and its spend no longer
resolved to a tracked key so the importer filed it `needs_review` and never
debited it.

`_sync_enrollment` now calls `/key/delete` on the parent key *before*
`mark_enrollment_synced` clears it (the row is the only handle we have on it).
A delete failure propagates to the existing handler, so the row goes `failed`
and the backfill worker retries the whole sync — the revocation is never
silently dropped. New `litellm.delete_virtual_key` client helper. The migration
docstring now states that the old key is reclaimed at sync time and that the
window between upgrade and the next sync tick is when it is still live.
Tests: `test_first_sync_revokes_the_pre_b2_unscoped_key`,
`test_revocation_failure_marks_sync_failed_for_retry`.

### 3. `build_sync_fingerprint` was write-only — "rotation via fingerprint bump" wasn't real

The fingerprint was stored and never read, so R2's rotation mechanism did not
exist, and the previous version of this body overclaimed it. Three narrow fixes
(deliberately **not** a drift engine):

- `ensure_user_enrollment`/`ensure_org_enrollment` now recompute the expected
  fingerprint and flip the row to `pending` on mismatch
  (`_reopen_if_key_set_drifted`), so the bump mechanism is real.
- The fingerprint input is corrected. It previously hashed
  `key_alias=subject_label`, which is neither a real alias nor a description of
  the key set. **What the parent fingerprint now covers:** the LiteLLM
  `team_id`, the `subject_label` every alias is derived from, and the sorted
  gateway-capable `harness_kinds` (one key per entry) — i.e. exactly the inputs
  that decide the key set's shape. It deliberately does **not** cover the
  mirrored team budget: that moves with every spend and top-up and is already
  rewritten by the importer/top-up loops, so folding it in would re-sync on
  every login while repairing nothing the sync path owns. Per-key drift (a
  key's own team + alias) is tracked separately on each child row via
  `build_enrollment_key_fingerprint`.
- `topups._remint_enrollment_key` recomputes the child fingerprint from the
  team/alias it just minted under instead of copying the stale prior value
  forward.

Test: `test_added_harness_kind_reopens_enrollment_and_mints_missing_key`
(config gains a harness kind → mismatch → row reopens → next sync mints exactly
the one missing key, never re-minting the others), plus
`test_child_key_fingerprint_covers_team_and_its_own_alias`.

### 4. Materialization amplification

Scheduling was inside the per-key re-mint loop, so a top-up scheduled N
identical passes for N keys. Hoisted to one pass per enrollment (matching
`update_team_budget`, which already sat outside the loop). The test that
asserted the amplification now asserts the single schedule.

### CI blockers

- `uv run ruff format` applied to `topups.py` and the three top-up/enrollment-key
  test modules (CI runs `ruff format --check`).
- `scripts/max_lines_allowlist.txt`: `db/models/cloud/agent_gateway.py` bumped
  504 → 556 for the new `agent_gateway_enrollment_key` model. A module split is
  deferred to the model-taxonomy pass rather than done mid-stack.

- `DEBUG=true uv run pytest tests/unit tests/integration -q -k "agent_gateway or agent_auth or billing_limit_enforcement_llm or litellm"`
  → **230 passed**, 0 failed (5 new tests in
  `server/tests/integration/test_agent_gateway_key_lifecycle.py`).
- `uv run ruff check` + `ruff format --check` clean across `proliferate` and
  `tests`.
- `scripts/check_max_lines.py`, `check_migration_heads.py` (single head
  `28adf1a9e376`, 129 revisions), `check_server_boundaries.py`,
  `check_docs.py` all pass.
- `uv run mypy` on every touched module: no new errors. The one remaining
  finding (`db/store/agent_gateway/enrollments.py` `select(User.id)` overload)
  is pre-existing — confirmed by re-running mypy on the pristine tree.


pablonyx:
## QA fork — BLOCKER found by merge integration (not visible to this PR's CI)

Reproduced on an integration branch = `origin/main` + a6 + **b3 (this stack)** + c1 + d2.

**Defect:** `server/proliferate/server/cloud/agent_gateway/enrollment.py:46` aliases the wrong constant:

```python
_GATEWAY_CAPABLE_HARNESS_KINDS = AGENT_AUTH_HARNESS_KINDS   # should be ..._GATEWAY_CAPABLE_...
```

On this branch that is harmless — `AGENT_AUTH_HARNESS_KINDS` is the 4-tuple here, so it happens to equal the gateway-capable set. **C1 (#1505) widens `AGENT_AUTH_HARNESS_KINDS` to include `cursor`** for selection purposes. The moment both land, the eager mint loop at `enrollment.py:467` starts minting a 5th key with `models=["cursor"]` — an access group that no entry in `server/litellm/config.yaml` carries (verified: the only 4 groups granted are claude/codex/grok/opencode).

Executed on the merged tree (`DEBUG=true` python, real modules, no mocks):

```
enrollment._GATEWAY_CAPABLE_HARNESS_KINDS = ('claude','codex','opencode','grok','cursor')   <-- wrong
selection_rules.GATEWAY_CAPABLE_HARNESSES = ('claude','codex','opencode','grok')
DISAGREEMENT (minted but never selectable)  = {'cursor'}
```

**And it turns a shipped test red.** `server/tests/unit/test_litellm_config_access_groups.py` ships on this stack (absent on main and on C1), so only the merge runs it against C1's widened tuple:

```
FAILED tests/unit/test_litellm_config_access_groups.py::TestLitellmConfigAccessGroups::test_every_supported_harness_has_at_least_one_model
1 failed, 3 passed in 222.64s
```

That test is also internally self-contradictory post-merge: its comments (`:5-6`, `:24-25`, `:71-72`) all assert "cursor is explicitly excluded from `AGENT_AUTH_HARNESS_KINDS`", which is exactly the premise the merge falsifies. Separately, `test_access_group_names_are_exactly_harness_kinds:51` silently *weakens* — its `allowed` set grows to include `cursor`, so it would now accept a stray cursor access group it was written to reject.

**Ordering makes the runtime failure worse than one wasted key.** `cursor` is last in the tuple, so the 4 legitimate keys mint first and the cursor mint is the final iteration *before* `_revoke_parent_key` and `mark_enrollment_synced`. If LiteLLM rejects an unknown access group, `except LiteLLMIntegrationError` (`:517`) marks the whole enrollment `failed` **after** the good keys exist and **before** the parent row is cleared — so the pre-B2 unscoped all-model key is never reclaimed, and retry short-circuits the 4 good keys (`:383-388`) and dies on cursor again: permanently `failed` enrollment holding a live unscoped key. If LiteLLM instead *accepts* the unknown group, every subject carries a 5th zero-model key forever and the fingerprint at `:498-503` bakes `cursor` into the expected key set, so `_reopen_if_key_set_drifted` will later treat the correct 4-key state as drift. Which branch occurs depends on proxy validation — I did not claim the Docker window to determine it, so please verify that when you fix this.

**Fix:** one line —

```python
_GATEWAY_CAPABLE_HARNESS_KINDS = AGENT_AUTH_GATEWAY_CAPABLE_HARNESS_KINDS
```

and point the test's `missing` computation at the gateway-capable tuple (which is its own stated intent, `:71-72`).

**MINOR, same file family:** `constants/agent_gateway.py:18-26` and `selection_rules.py:20-28` still document behavior this stack correctly reverted — they claim the store consults the constant "to skip the disabled gateway revision-marker row." `db/store/agent_gateway/selections.py:130-136` now inserts that row unconditionally, and a repo-wide grep shows the constant is imported *only* by `selection_rules.py:15,29` — no store file imports it, so the stated `check_server_boundaries.py` justification for its placement no longer holds. Worth correcting in the same commit, since that comment is the reasoning that produced the earlier revision-marker blocker.

Full detail + repro transcripts in `codex/ledger-qa.md` (F-007, F-032). Read-only QA — I have not pushed to any `agents/*` branch.

pablonyx:
## QA fork — B2 money-path findings (second comment: enrollment/migration/revocation)

Same integration branch as my earlier comment. Read-only QA; nothing pushed to any `agents/*` branch. Ledger: `codex/ledger-qa.md` F-005, F-008, F-010…F-016.

I'll lead with what I verified is **right**, because most of B2's core holds: keys genuinely never carry a budget (`max_budget` absent from `_mint_virtual_key_idempotent:128-167`, `_sync_one_harness_key:392-398`, `_remint_enrollment_key` in `topups.py:319-327`), so R2's "team is the only budget layer" holds. The "0 means uncapped" trap is handled where it matters (`_EXHAUSTED_BUDGET_FLOOR_USD = 0.01` applied at `:308-309`; `budget._explicit_org_budget_configured:38-51` correctly reads the `"0"` default as *absence* of a cap). The importer's exhaustion path is fail-closed on an empty key list (`usage_import.py:382-414`). Spend attribution resolves child keys first with a parent fallback (`usage_import.py:133-158`) and never drops unresolved rows. `upsert_enrollment_key` preserves ciphertext on a metadata-only resync (`enrollment_keys.py:117-120`). Migration is a clean single head (`28adf1a9e376` sole child of `35fa0038d703`) — I applied it from base on a scratch DB and the table shape, FK cascade and backfill UPDATE all read correct. And the whole agent-gateway/agent-auth integration suite is green **on the merged tree** (808 passed), which is real evidence the server contract holds.

### BLOCKER — the migration re-mints fresh *enabled* keys for exhausted enrollments, and the importer can never disable them

`28adf1a9e376_agent_gateway_enrollment_key.py:104-108`. Chain:

1. The backfill flips **every** (`sync_status='synced'` AND `revoked_at IS NULL`) row to `pending` — regardless of `budget_status`.
2. `_sync_enrollment` → `_sync_one_harness_key` (`enrollment.py:383-389`) finds no child rows (table is brand new) → mints 4 brand-new keys. LiteLLM creates them **unblocked**.
3. `mark_enrollment_synced` (`enrollments.py:156-188`) and `mark_enrollment_pending` (`:211-230`) never touch `budget_status` → the row is still `exhausted`.
4. The importer then **refuses** to act on it: `usage_import.py:448` `if enrollment.budget_status == ...EXHAUSTED: continue`, and `_apply_llm_limit_reached` returns early at `:537-541`.

So wall #1 (key disablement) becomes permanently unreachable for exactly the population that was already over its limit. Worst concrete shape: an org that turned overage off after exhaustion, or any subject whose team was created uncapped (`agent_gateway_default_org_budget_usd` default `"0"` → `_parse_budget` → `None` → no `max_budget` sent) — post-migration that subject holds four live, unblocked, uncapped keys under an uncapped team, and nothing will ever block them. Pre-B2 this couldn't happen: the re-sync skipped the mint because `virtual_key_id` was non-null. Today's blast radius is bounded to "orphaned spend-capable keys" because B3's renderer withholds delivery (`materialize/agent_auth.py:316`), but the design has silently gone from two walls to one for this population.

Fix direction: `_sync_enrollment` should re-assert `budget_status` against the ledger after minting (disable the fresh keys, or reset to `ok` only when funded), and/or the migration should exclude non-`ok` rows and let top-up reactivation drive them.

### MAJOR — revocation is DB-only; revoked keys stay live on the proxy and their spend is never debited

`db/store/agent_gateway/enrollment_keys.py:139-163` — `revoke_enrollment_keys` sets `revoked_at` and nothing else, no LiteLLM call. Afterwards `get_enrollment_key_by_virtual_key_id` filters `revoked_at IS NULL` (`:62-81`), so `usage_import._resolve_enrollment_for_spend_key` (`:133-158`) can no longer resolve the token → the row lands `needs_review` with null subject links. That is the exact failure mode B2 itself fixed for the pre-B2 parent key, and the key remains valid on the proxy so any holder keeps spending on the team budget for free. The integration test that "proves" the cascade only asserts DB rows (`tests/integration/test_agent_gateway_enrollment_keys.py:216-232`), never a LiteLLM interaction. Latent — `revoke_enrollment` has no production caller today — but it's the seam all offboarding/downgrade work will land on.

### MAJOR — `downgrade()` drops the only handles on live money-spending keys

`:111-113` (`op.drop_table`). By rollback time `_sync_enrollment` has already `/key/delete`'d the parent key and cleared the parent ciphertext (`enrollment.py:485-496`; `enrollments.py:177-183`). Post-downgrade: four live keys per subject with no row anywhere, parent holding no key material, `get_enrollment_by_virtual_key_id` unable to resolve their token hashes, all subsequent spend `needs_review` — unbilled *and* unattributable — while old-code paths (`set_key_budget`, `enable_virtual_key` on `enrollment.virtual_key_id`) silently no-op on `None`. A safe downgrade should copy one child key back onto the parent row, or refuse to run.

### MAJOR — the team is now the only budget layer, and the sync path never rewrites it

`enrollment.py:454-458` calls `litellm.ensure_team(alias=..., max_budget=budget)` **without** `team_id=`, though the row holds `litellm_team_id`. `ensure_team` (`integrations/litellm/client.py:144-187`) applies `max_budget` only on the `/team/new` create path; for an existing alias it returns the found id and **discards** `max_budget`. Pre-B2 this was survivable because the mint also carried `max_budget` per key and `topups.set_key_budget` rewrote it; post-B2 keys deliberately carry no budget, so the team cap is written exactly once at first creation and only `topups.update_team_budget` (`:292-296`) can ever change it — a path that runs only for subjects with an existing grant that dipped below threshold. A user whose grant is later revoked/reduced, or an org that turns overage off, keeps the old cap forever. Secondary: every re-sync now performs a full hydrated `/team/list` of *all* teams (the client docstring at `:10-15` explicitly says to prefer the stored-id path) — and the migration above puts every enrollment through that path at once.

### MAJOR (pre-existing, amplified here) — grant expiry re-opens both walls and can move the payer mid-flight

`sum_active_grants_usd` excludes expired grants (`credits.py:116-133`), so on expiry `granted_usd → 0` and `remaining_usd` goes negative. Then `is_gateway_budget_available` returns True on `granted_usd <= 0` (`budget.py:133-134`), `_enforce_subject_exhaustion` returns False on the same condition (`usage_import.py:439-440`), and `_remaining_credit_budget_raw` returns the *fallback default* budget rather than a near-zero floor (`enrollment.py:304-310`). A user who burned a $5 grant that then expires is treated as a fresh "no ledger, LiteLLM default is the guardrail" subject. Combined with the previous finding, the practical cap is whatever the team was created with. **Payer flip:** for an org member this flips `_subject_is_funded` (`budget.py:54-65`) true→false, so `get_gateway_enrollment_for_user` (`:98-110`) moves the payer from the org back to the user's personal subject on the next render — and the two enrollments hold *different* virtual keys, so `state.json` silently starts charging a different subject with no user-visible event. This is rulings-queue item ① with a concrete mechanism attached; the ruling should cover expiry, not just exhaustion.

### MINORs

1. **Redundant unique constraint forecloses revoke-then-remint.** The table carries both `uq_..._scope UNIQUE (enrollment_id, harness_kind)` (total) and `ux_..._active_scope UNIQUE ... WHERE revoked_at IS NULL` (partial). The partial one is what the design wants; the total one makes it unreachable and forbids key history. Proven by SQL on a scratch DB: insert claude key → `revoked_at=now()` → insert replacement **fails** on `uq_agent_gateway_enrollment_key_scope`. Not a live bug today (the only revocation path revokes the parent too, and every lookup filters `revoked_at IS NULL`, so the next `ensure_*` creates a new enrollment id), but R2 names delete+mint as a sanctioned rotation strategy and the partial index proves the author intended revoke-and-retain-history. The first person to implement rotation the obvious way hits a hard 500 on a money path that looks like data corruption. Suggest dropping the total constraint (migration lines 71-75) and keeping the partial index — semantics unchanged for every path today. If the total constraint *is* deliberate, then the partial index is dead weight and R2's wording should say "in-place update only".
2. **Org-enrollment alias divergence in the top-up re-mint path.** `enrollment_subject_label` returns `user-{user_id}` whenever `user_id` is not None (`:170-173`), and org enrollments always set both ids (`ensure_org_enrollment:325-331`). Sync mints with `subject_label=f"org-{org}-user-{user}"` (`:337,362`), but `topups._remint_enrollment_key` (`topups.py:317-318`) derives the alias from that helper — so a rotate for an org member deletes the old key and re-mints under a different, personal-looking alias. It then no longer matches the deterministic value that `_mint_virtual_key_idempotent`'s orphan recovery and `delete_virtual_keys_by_alias` rely on, and the T3 harness's `deriveActorKeyAlias` (`tests/release/src/services/qualification-litellm.ts:242-249`) will fail to resolve it. Give the helper an org branch, or pass `_remint_enrollment_key` the same `subject_label` the sync used.
3. **`delete_virtual_key` can't distinguish "already gone" from "network down".** `integrations/litellm/client.py:329-335` swallows every `LiteLLMIntegrationError`, and `_request`/`_admin_request` map both transport errors and any status ≥400 to the same code `litellm_request_failed` (`:83-102`). So a five-second blip during `_revoke_parent_key` (`:412-439`) leaves the pre-B2 all-model key **live** while the row proceeds to `synced` with `virtual_key_id=None` — the exact outcome the docstring at `:417-421` says must not happen — with no retry because the handle is gone. Narrow the tolerance to 404 / "not found".
4. **The child `sync_fingerprint` column is write-only.** Written at `:405-408` and `topups.py:338-341`; grep shows no reader. The parent fingerprint became a real mechanism via `_reopen_if_key_set_drifted`; the per-key fingerprint that would detect a single key moving team or losing its alias has none. Compare it in `_sync_one_harness_key`, or drop the column.

pablonyx:

```

none origin/main
```

```
```
```diff
-from proliferate.constants.agent_gateway import AGENT_AUTH_HARNESS_KINDS
+from proliferate.constants.agent_gateway import AGENT_AUTH_GATEWAY_CAPABLE_HARNESS_KINDS
-        missing = set(AGENT_AUTH_HARNESS_KINDS) - granted
+        missing = set(AGENT_AUTH_GATEWAY_CAPABLE_HARNESS_KINDS) - granted
```

pablonyx:
F-007 fixed at 8f5f625ba1dacdedec0b3347794a4a87bb5fe4b3.

`enrollment.py`'s `_GATEWAY_CAPABLE_HARNESS_KINDS` was aliased to
`AGENT_AUTH_HARNESS_KINDS`. Harmless alone, but c1 (#1505) widens that
tuple to include `cursor` while adding the correct
`AGENT_AUTH_GATEWAY_CAPABLE_HARNESS_KINDS` (still the 4-tuple) beside
it — once merged, b2's mint loop would have iterated `cursor` too and
minted a live LiteLLM virtual key with an access group that grants
zero models.

Fix: added `AGENT_AUTH_GATEWAY_CAPABLE_HARNESS_KINDS` to
`constants/agent_gateway.py`, character-for-character identical to
c1's definition and at the same position (clean merge later),
repointed the enrollment.py alias and the two access-group tests in
`test_litellm_config_access_groups.py` at it. `AGENT_AUTH_HARNESS_KINDS`
stays a 4-tuple on this branch — the cursor widening is c1's change,
not imported here.

Test evidence (Python-only, no cargo build needed):
```
DEBUG=true uv run pytest tests/unit/test_litellm_config_access_groups.py \
  tests/integration/test_agent_gateway_enrollment.py \
  tests/integration/test_agent_gateway_key_lifecycle.py -q
21 passed
```

`git merge-tree --write-tree` against `origin/agents/c1-cloud-login-cursor-ui`
completes with no conflicts (exit 0, no conflict markers in the resulting
tree).

pablonyx:

---

PR #1511, opened 2026-07-26
fix(agent-auth): fail closed on an unsatisfiable selection (A4)

pablonyx:
Track A, PR 4 — Round 2 first segment. **Stacked on #1504** (`agents/a3-cursor-rust` → #1503 → #1499); retarget as parents merge. The diff below is A4's own commit only.

## The law this implements

`specs/codebase/platforms/product/agent-auth.md`, "Delivery laws":

> **Absent means native; present-but-empty fails closed.** A harness with no entry in the document runs on its own login. A harness whose entry is present but whose selected sources could not be satisfied (unsynced enrollment, exhausted budget, revoked key) keeps its entry with the dead source omitted — and a launch that then resolves zero usable sources for a still-selected route is refused with a typed error. **A selection never silently degrades to the user's personal credentials.**

## Gap bullets closed (deleted in this PR)

1. **"Unsatisfiable sources silently degrade to native."** — *"The renderer drops a dead source and omits an empty harness entry, and the runtime reads absence as native — so a desktop user with a native claude login whose gateway budget exhausts silently starts billing their personal Anthropic account. The body's 'present-but-empty fails closed' law is not implemented."*
2. **"Dead error variants."** — `SelectionMissing` and `SelectionConflict` never constructed.

## Why this needed three layers

The law was not merely unimplemented — it was **unimplementable**, because the information it depends on was destroyed at the bottom:

```rust
// before
pub fn sources_for(&self, harness_kind: &str) -> &[AuthSource] { … .unwrap_or(&[]) }
```

A caller could not distinguish "the user never configured this harness" from "the user selected the gateway and we could not satisfy it". Both were `&[]`. So:

**Layer 1 — the type (ruling R3).** `sources_for` returns `Option<&[AuthSource]>`: `None` = absent, `Some([])` = present-but-empty. The distinction now lives in the type rather than in a caller's convention.

**Layer 2 — `resolve_profile`.** `Some([])` → `RouteAuthError::SelectionMissing { harness_kind, revision }`. The revision rides along so a UI can say which document generation was dead.

**Layer 3 — the Python renderer, which had the same bug twice.** `render_agent_auth_state` dropped the whole harness entry when its every source was unsatisfiable; it now keeps an entry for every *selected* harness, empty if need be. And `materialize_agent_auth` deleted the state file whenever `harnesses` was empty — for an exhausted budget that meant the sandbox found **no document at all**, which reads as native. It now deletes only when nothing is selected on the surface, which is the one case where "absent means native" is the right answer.

Two non-obvious call sites (`gateway_resolver.rs`'s credential lookup, the manual-refresh endpoint) legitimately want "usable gateway source or nothing" and use `.unwrap_or_default()` with a comment saying why — fail-closed refusal is the launch path's job, not the model planner's.

## Create-time typed 409

Session create now checks `launch_route_selection_failure` **before** the readiness gate. The readiness gate would also refuse this launch — but as `400 SESSION_CREATE_FAILED` "agent 'claude' is not ready", which reads to a user as *go install something* when the real answer is *your gateway budget is exhausted*. The new path returns `409 AGENT_ROUTE_SELECTION_MISSING`, which is what lets the UI route the user to the auth pane. The launch path (`start_live_session`) still reaches the same conclusion independently, so this is the earlier, better-labelled gate and never the only one.

New `CreateSessionError::RouteAuth` carries route-auth's own error so the code and message are not re-invented at the boundary.

## `SelectionConflict` deleted, not wired

The gap bullet offered "delete them or wire them to the fail-closed law". `SelectionMissing` is wired. `SelectionConflict` ("N entries where one is allowed") is **deleted**: source cardinality is a per-harness *server* rule (`selection_rules.py`) enforced before a document is ever written, and the document's shape — one entry per harness with a flat source list — cannot represent the conflict it described. There was no input a correct runtime could construct it from. Grepped for consumers of `AGENT_ROUTE_SELECTION_CONFLICT` across Rust/Python/TS: none.

## Contract fixture — `fixtures/contracts/agent-auth-state/`

New, with a README stating what it pins and why. Python produces, Rust consumes, per `specs/developing/testing/README.md`'s contract-fixture rule.

It pins the two things the languages can silently drift on:
1. **Per-harness gateway keys** — three distinct virtual keys, because the gateway scopes keys per (subject, harness).
2. **Empty-sources semantics** — `grok` present with `"sources": []` (must refuse), no `opencode-zen` entry at all (must be native).

Incidentally pinned because they are easy to get wrong: the document is snake_case on the wire while `settings` values are the catalog's camelCase; opencode composes a direct `api_key` + gateway, **in the producer's own `(kind, env_var_name)` sort order** (`api_key` sorts before `gateway`, so a gateway-first fixture would be a document no reconcile could emit — caught in review round 2 and now asserted on both sides); cursor's only route is `api_key`.

### ⚠️ B3 reconciliation required — flagging rather than forking

**Correction (review round 2).** An earlier version of this section claimed no B3 branch existed on the remote. That was wrong: `origin/agents/b3-renderer-key-map` (**#1512**) exists and branched from `origin/main`, i.e. *before* this PR. I read its renderer. The reconciliation picture:

| | B3 (#1512) | this PR (A4) | verdict |
| --- | --- | --- | --- |
| per-harness gateway key map | implements it (`gateway_virtual_keys: Mapping[str, str]`) | not implemented — one subject-wide key fanned out | **agree on the fixture's shape**; B3 closes the gap this PR pins as open |
| empty-sources semantics | still the OLD behavior: `by_harness.setdefault(...)` drops a fully-unsatisfiable harness, and `materialize_agent_auth` deletes the state file whenever `harnesses` is empty | keeps a `sources: []` entry per selected harness; deletes only when nothing is selected | **B3 carries the bug this PR fixes** — its branch point predates the fix |

So the two branches do not conflict on the fixture's *shape*; they collide on the fail-closed behavior, and the collision is dangerous because B3's version is a *deletion* of behavior — a careless rebase resolves cleanly and silently re-opens the silent-degradation gap. Two guards:

- `test_per_harness_gateway_keys_are_not_produced_yet` asserts **today's shared-key behavior** with a comment instructing B3 to delete it and assert distinctness instead — so a passing suite after B3 cannot conceal a producer/fixture disagreement.
- `test_unsatisfiable_selection_survives_renderer_merge_tripwire` (new) asserts **both halves of the fail-closed fix in one pass** — the pre-seeded `by_harness` entry per selected harness, *and* the narrowed delete condition, *and* that the delete branch is still reachable for its one legitimate case (nothing selected). Its name and failure messages both point at `agents/b3-renderer-key-map`, so whoever rebases B3 gets told exactly what they reverted.

**B3 must produce byte-identical output for these inputs**, and must keep this PR's renderer semantics when it rebases (per the plan's `materialize/agent_auth.py` collision corridor, B3 lands first in principle — in practice A4 got there first, so the rebase direction is B3-onto-this).

## Tests

Rewritten because they pinned the bug as intended behavior (all three asserted native for a dead selection):
- `profile.rs::empty_sources_is_native` → `empty_sources_fails_closed_with_the_revision`
- `render_tests.rs::empty_sources_render_native_delta` → `empty_sources_refuse_the_launch_instead_of_rendering_native` (full render path; also asserts nothing was materialized on the way to the refusal)
- `test_all_unsatisfiable_renders_empty_harnesses` → `test_all_unsatisfiable_keeps_the_selected_harness_with_no_sources`
- `test_no_resolvable_sources_deletes_state_file` → `test_unsatisfiable_selection_writes_an_empty_entry_instead_of_deleting`
- plus `test_revoked_api_key_source_is_omitted`, `test_unsatisfiable_gateway_still_renders_satisfiable_api_key`, `test_gateway_with_unsynced_enrollment_is_dropped`, `test_bad_gateway_still_purges_revoked_key_and_writes_rest`, and the base-url warning test — all re-asserted against the empty-entry shape

New (tier 1 throughout):

| Test | Proves |
| --- | --- |
| `state.rs::sources_lookup_distinguishes_absent_from_present_but_empty` | the three-way lookup at the layer where it was destroyed |
| `profile.rs::absent_is_native_while_present_but_empty_is_refused` | opposite answers from ONE document — the distinction as a single comparison |
| `profile.rs::one_harnesss_dead_selection_does_not_refuse_another` | the refusal is per-harness; one exhausted budget must not take the machine down |
| `profile.rs::no_file_is_native_even_though_empty_entries_now_refuse` | a never-synced machine is unaffected |
| `sessions_errors.rs::an_unsatisfiable_selection_maps_to_a_typed_conflict` | 409 + `AGENT_ROUTE_SELECTION_MISSING`, not 400 |
| 5 × `contract_fixture_tests.rs` | fixture parses; gateway keys distinct; empty-vs-absent resolved end to end; satisfiable entries resolve as documented; and the fixture **drives a real launch render** (a fixture that parses but cannot drive a launch would pass the others) |
| `test_no_selections_at_all_renders_an_empty_document` / `..._still_deletes_the_state_file` | the one remaining delete case, so "absent means native" stays reachable |
| 4 × `TestAgentAuthStateContractFixture` | producer half: fixture is a shape this renderer could emit; empty-sources semantics match; **the fixture's per-harness source order is the order this renderer emits**; the per-harness-key gap is asserted as open (B3's tripwire) |
| `test_unsatisfiable_selection_survives_renderer_merge_tripwire` | both halves of the fail-closed fix at once (pre-seeded entry + narrowed delete + the delete branch still reachable), so a B3 rebase that reverts either fails loudly |

Results: `cargo test -p anyharness-lib --lib` **1327/1327**. `cargo test --workspace` green except the known pre-existing `apps/desktop/src-tauri` flake (documented in #1503; a different test fails per run, green 3/3 in isolation, nanosecond `temp_path` collision + a 5s subprocess-lock deadline; untouched by this PR). Server `pytest tests/unit` **1172/1172**. `ruff check` + `format --check` clean. Every `repo-shape` check passes.

## Contradictions encountered

None ruling-worthy. Three judgment calls:

1. **`launch_route_selection_failure` tolerates an unreadable document.** It returns `None` for a malformed or origin-mismatched state, matching `launch_route_provides_credentials`'s existing policy exactly. Making create *reject* on a document the launcher itself tolerates would be a new fail-closed rule the spec does not ask for, and would brick session creation on a state file the next push heals.
2. **Where the create-time check goes.** Before the readiness gate, so the better-labelled error wins. Placing it after would have left the generic "not ready" as the user-visible answer in exactly the case this PR exists to explain.
3. **Allowlist growth.** `render_tests.rs` (616) crossed the 600 cap and is allowlisted with an explicit split-on-next-growth note, since splitting it mid-change would have obscured the behavioral diff. `test_agent_auth_materialization.py` was allowlisted at 617 in round 1; in round 2 the new tripwire test would have pushed it further, so **the contract-fixture producer class was split out** to `test_agent_auth_state_contract_fixture.py` and the allowlist entry drops to 614 rather than growing. Per the carry-over note, `model.rs` did **not** grow in this PR (it is untouched), so no second bump there. `gateway_resolver.rs` (+4) and `agent_gateway_catalog.rs` (+3) are comment-only growth on already-allowlisted files.

Read `codex/probe-engine-design.md` before starting. A4 touches two things A7/A8 build on, and both stay compatible:
- **`resolve_profile`'s contract.** The design's §1.3 phase A calls `resolve_profile` and already routes "no matching source found for an env/route context" to `Err(SelectionMissing)` → `lastAttempt.outcome="failed"`, no spawn. A4 makes that variant *actually reachable*, which is what the design assumed.
- **`sources_for`'s signature.** The design does not call it directly (it goes through `resolve_profile`), so the `Option` change is invisible to it.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/api/http/agent_gateway_catalog.rs
+++ b/anyharness/crates/anyharness-lib/src/api/http/agent_gateway_catalog.rs
@@ -276,7 +276,10 @@ pub async fn refresh_gateway_models(
         ));
     };
     let revision = document.revision;
-    let sources = document.sources_for(&kind);
+    // Absent harness and present-but-empty are the same answer here — either way
+    // there is no gateway source to refresh against, and the caller gets the
+    // typed NO_SELECTION rejection below.
+    let sources = document.sources_for(&kind).unwrap_or_default();
     let source = sources
         .iter()
         .find(|source| source.kind == SOURCE_KIND_GATEWAY)
--- a/anyharness/crates/anyharness-lib/src/api/http/sessions_errors.rs
+++ b/anyharness/crates/anyharness-lib/src/api/http/sessions_errors.rs
@@ -153,7 +153,6 @@ pub(super) fn map_create_session_error(error: CreateAndStartSessionError) -> Api
 pub(super) fn map_route_auth_error(error: &RouteAuthError) -> ApiError {
     match error {
         RouteAuthError::SelectionMissing { .. }
-        | RouteAuthError::SelectionConflict { .. }
         | RouteAuthError::SelectionIncomplete { .. }
         | RouteAuthError::UnsupportedRoute { .. }
         | RouteAuthError::UnknownHarness { .. }
@@ -373,6 +372,28 @@ mod tests {
         })
         .into_response();
         assert_eq!(response.status(), StatusCode::CONFLICT);
+    }
+
+    /// An unsatisfiable agent-auth selection must reach the client as a typed
+    /// **409 naming the auth failure**, not the generic 400 "session create
+    /// failed" the readiness gate would produce. The distinction is the whole
+    /// point: 400 SESSION_CREATE_FAILED reads as "fix your request", while this
+    /// says "the route you selected is dead" — and only that lets the UI send the
+    /// user to the auth pane instead of to an install button.
+    #[test]
+    fn an_unsatisfiable_selection_maps_to_a_typed_conflict() {
+        use crate::domains::agents::route_auth::RouteAuthError;
+
+        let mapped = super::map_create_session_error(CreateAndStartSessionError::RouteAuth(
+            RouteAuthError::SelectionMissing {
+                harness_kind: "claude".to_string(),
+                revision: 42,
+            },
+        ));
+
+        assert_eq!(mapped.status(), StatusCode::CONFLICT);
+        assert_eq!(mapped.code(), Some("AGENT_ROUTE_SELECTION_MISSING"));
+        assert_eq!(mapped.into_response().status(), StatusCode::CONFLICT);
     }
 
     #[test]
--- a/anyharness/crates/anyharness-lib/src/domains/agents/catalog/gateway_resolver.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/catalog/gateway_resolver.rs
@@ -320,8 +320,12 @@ impl GatewayModelResolver {
         runtime_home: &Path,
     ) -> Option<(String, String)> {
         let state = load_state_file(runtime_home).ok().flatten()?;
+        // Absent harness and present-but-empty are the same answer here: no
+        // gateway credentials to plan with. Fail-closed refusal is the launch
+        // path's job, not the model planner's.
         let source = state
             .sources_for(harness_kind)
+            .unwrap_or_default()
             .iter()
             .find(|source| source.kind == SOURCE_KIND_GATEWAY)?;
         let base_url = source
--- a/anyharness/crates/anyharness-lib/src/domains/agents/route_auth/contract_fixture_tests.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/route_auth/contract_fixture_tests.rs
@@ -1,0 +1,170 @@
+//! Rust's half of the `agent-auth-state` contract fixture.
+//!
+//! Python produces `fixtures/contracts/agent-auth-state/v2.json`
+//! (`materialize/agent_auth.py::render_agent_auth_state`); this side asserts we
+//! parse it and resolve it to the profiles the fixture's README claims. Per
+//! `specs/developing/testing/README.md`, changing the shape means changing the
+//! fixture, which breaks whichever side has not caught up — the point is that the
+//! break is mechanical rather than a runtime surprise in a sandbox.
+//!
+//! Split from `render_tests.rs` for the line-count ceiling; nested inside it so
+//! its `TempHome` and resolver helpers are in scope.
+
+// The parent test module's helpers (`TempHome`, `HarnessPlanResolver`, …).
+use super::*;
+// The module under test, reached through its public re-exports.
+use crate::domains::agents::route_auth::state::{AgentAuthState, SOURCE_KIND_GATEWAY};
+use crate::domains::agents::route_auth::{
+    profile::ResolvedSource, resolve_launch_route_auth_for_server, AgentRuntimeAuthProfile,
+};
+
+const FIXTURE: &str =
+    include_str!("../../../../../../../fixtures/contracts/agent-auth-state/v2.json");
+
+fn fixture_state() -> AgentAuthState {
+    serde_json::from_str(FIXTURE).expect("the contract fixture must parse as a v2 document")
+}
+
+/// The document parses and its top-level shape is what the producer promises.
+/// Notably the wire is snake_case; a producer that emitted camelCase would fail
+/// here rather than silently rendering every harness native.
+#[test]
+fn the_contract_fixture_parses_as_a_v2_document() {
+    let state = fixture_state();
+
+    assert_eq!(state.version, 2);
+    assert_eq!(state.revision, 42);
+    assert_eq!(
+        state.user_id.as_deref(),
+        Some("20000000-0000-4000-8000-000000000001")
+    );
+    assert_eq!(
+        state.issuing_server_origin.as_deref(),
+        Some("https://api.proliferate.example")
+    );
+    assert_eq!(state.harnesses.len(), 5);
+}
+
+/// Contract point 1: every gateway source carries its OWN virtual key.
+///
+/// The keys are scoped per (subject, harness) by the gateway's access groups, so
+/// a producer that resolves one subject-wide key and fans it out is wrong. This
+/// asserts distinctness rather than exact values, so rotating the fixture's
+/// placeholder keys does not churn the test.
+#[test]
+fn every_gateway_source_carries_its_own_per_harness_key() {
+    let state = fixture_state();
+
+    let mut keys = Vec::new();
+    for harness in ["claude", "codex", "opencode"] {
+        let sources = state
+            .sources_for(harness)
+            .unwrap_or_else(|| panic!("{harness} entry present"));
+        let gateway = sources
+            .iter()
+            .find(|source| source.kind == SOURCE_KIND_GATEWAY)
+            .unwrap_or_else(|| panic!("{harness} gateway source"));
+        keys.push(gateway.key.clone().expect("gateway key"));
+    }
+
+    let distinct: std::collections::BTreeSet<_> = keys.iter().collect();
+    assert_eq!(
+        distinct.len(),
+        keys.len(),
+        "gateway keys must be per-harness, not one shared subject key: {keys:?}"
+    );
+}
+
+/// Contract point 2: the three-way empty-sources semantics, resolved end to end.
+///
+/// `grok` is present with `sources: []` in the fixture — a selection the producer
+/// could not satisfy — and must refuse the launch, while a harness the fixture
+/// omits entirely is native. If the producer ever "tidies up" by dropping empty
+/// entries, this test is what catches it.
+#[test]
+fn the_fixtures_empty_entry_fails_closed_while_an_absent_one_is_native() {
+    let state = fixture_state();
+
+    // Present with an empty list → refuse.
+    assert_eq!(state.sources_for("grok").map(<[_]>::len), Some(0));
+    let error = resolve_profile(Some(&state), "grok").expect_err("grok must fail closed");
+    assert!(matches!(
+        &error,
+        RouteAuthError::SelectionMissing { harness_kind, revision: 42 } if harness_kind == "grok"
+    ));
+
+    // Absent from the document → native.
+    assert!(state.sources_for("opencode-zen").is_none());
+    assert_eq!(
+        resolve_profile(Some(&state), "opencode-zen").expect("resolve"),
+        AgentRuntimeAuthProfile::Native
+    );
+}
+
+/// Every satisfiable entry resolves to the sources the README describes, in
+/// order: cursor's only route is `api_key`, and opencode composes a gateway plus
+/// a direct provider key simultaneously.
+#[test]
+fn the_fixtures_satisfiable_entries_resolve_to_the_documented_profiles() {
+    let state = fixture_state();
+
+    for harness in ["claude", "codex"] {
+        match resolve_profile(Some(&state), harness).expect("resolve") {
+            AgentRuntimeAuthProfile::Sources(sources) => {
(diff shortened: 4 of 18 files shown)


pablonyx:
## QA fork — two fail-closed findings confirmed LIVE on a merged integration branch

Booted `origin/main` + a6 + b3 + c1 + d2 as a real dev profile (runtime `:8461`, backend `:8004`). Read-only QA — nothing pushed to any `agents/*` branch. Full transcripts in `codex/ledger-qa.md` (F-018, F-019).

First, the good news: I tried to break the A4 law itself and couldn't. The `sources_for` contract (`None` → Native, `Some([])` → SelectionMissing, `Some([..])` → usable) is genuinely implemented on both planes, and the launcher really does refuse — `POST /v1/sessions` returns `409 AGENT_ROUTE_SELECTION_MISSING` on a selected-but-unsatisfiable harness. Both findings below are about *what the rest of the system does around that refusal*.

### MAJOR — false `Ready` + composer dead-end on exactly the state this law exists for

Three commands, live:

1. Write the fail-closed document (exactly what B3's renderer produces when a gateway key is withheld — budget exhausted, key revoked):
   `{"version":2,"revision":1,"harnesses":[{"harness_kind":"claude","sources":[]}]}`
2. `GET /v1/agents` → `readiness: "ready"`, `credentialState: "ready"`, `cliAuthState: "authenticated"`.
3. `GET /v1/agents/launch-options?workspace_id=…` → claude is offered with **11 models** (the full *native*-context list, computed from the user's personal credentials for a route the launcher will refuse).
4. `POST /v1/sessions` → **HTTP 409** `AGENT_ROUTE_SELECTION_MISSING` "no agent-auth route selection for harness 'claude' at revision 1".

A Ready badge, a populated model menu, and a hard 409 on the only action the UI offers — zero recovery affordance.

Mechanism: `route_auth/mod.rs:208-227` `launch_route_provides_credentials_for_server` returns `false` for `Err(SelectionMissing)`, so `apply_launch_route_upgrade` never runs and native readiness governs. `sessions/service/launch_options.rs:124-127` gates only on `resolve_launch_agent(...).status` and never calls `launch_route_selection_failure`.

This inverts `agent-distribution.md:381-383` — "the UI never shows CredentialsRequired for a harness that would launch fine" — into showing Ready for a harness that cannot launch at all.

**Worse for opencode, also confirmed live** (`revision 2`, `harness_kind: "opencode"`, `sources: []`): `readiness: "ready"`, present in the composer list, `POST /v1/sessions` → 409. This machine has **no** opencode native login at all — `credentials.rs:176-178` returns Ready unconditionally for `provider_managed`. So opencode reports Ready with no credential *and* no honorable route; every launch is a 409 behind a green badge.

### MAJOR — origin mismatch silently degrades a PAID gateway route to the user's personal login

Same `state.json` in both arms — a valid gateway source stamped `server-a` (key value was a non-secret test placeholder; no real credential used anywhere). One env var apart:

| | Arm A — mismatch | Arm B — match |
|---|---|---|
| readiness | ready | ready |
| composer models for claude | **11** (native list) | **6** (gateway catalog) |
| `POST /v1/sessions` | **HTTP 200 — session created and started** | HTTP 400 `SESSION_MODEL_GATED` (native model correctly refused: "gated behind auth contexts [\"bedrock\"]") |
| gateway artifacts materialized | **none** (`find <RH>/agent-auth` → only `state.json`) | route honored |

In Arm A the gateway route was silently discarded and the launch ran on the user's own native claude login — no error, no warning, no UI signal, while the UI still shows the gateway selection. The 11-vs-6 model count is the only user-visible tell that the route changed.

Mechanism: `mod.rs:121-127` `load_effective_state` filters a document whose `issuing_server_origin` differs → `None`; `profile.rs:91-93` then returns **Native**, because absent-means-native is the design (`state.rs:117-133`). `launch_route_selection_failure` (`mod.rs:254-268`) uses the same filter, so no gate fires anywhere.

This violates `agent-auth.md:340` ("There is no fallback to native on a failure: a user who selected the gateway never silently runs on their personal login") and `:207-213`. Not injecting the stale key is *right*; resolving Native instead of refusing is the bug.

**Fix:** mismatch must produce `SelectionMissing`, not Native — a one-line change in `load_effective_state`'s filter plus a test asserting mismatch → `Err(SelectionMissing)` rather than `Ok(Native)`. For F-018, `launch_options` should consult `launch_route_selection_failure` so the menu and badge agree with the launcher.

### Env gotcha worth knowing before you try to reproduce

The dev runtime does **not** set `PROLIFERATE_API_BASE_URL_ORIGIN` — only the desktop sidecar does (`apps/desktop/src-tauri/src/sidecar.rs:353-356`, unconditional). The origin guard is therefore inert under `make run`, and my first attempt at this produced a false negative. Any origin-guard test must launch `anyharness serve --port … --runtime-home …` directly with the var set; `matches_server_origin` returns `true` when the current origin is absent (`state.rs:152-157`), which is by design for cloud but means the dev profile can never reproduce F-019.

pablonyx:
**QA (integration) — F-038: merge-only red between #1511 (a4) and #1512 (b3)**

Found on a merged QA branch (a-stack head + b-stack head + C1 + D2 + A7). Not a defect on either branch alone — read-only QA, no changes pushed to either branch.

**The red**
```
$ cd server && DEBUG=true uv run pytest tests/unit/test_agent_auth_materialization.py
FAILED TestRenderAgentAuthState::test_harness_missing_from_the_key_map_is_dropped_alone
  AssertionError: assert ['claude', 'codex'] == ['claude']
  Left contains one more item: 'codex'
```

**Provenance** — the test case exists on exactly one ref:
```
0  origin/main
0  origin/agents/a4-fail-closed
1  origin/agents/b3-renderer-key-map
0  origin/agents/b2-per-harness-keys
0  origin/agents/b1-access-group-tags
0  origin/agents/c1-cloud-login-cursor-ui
```
It's B3's test, and it passes on B3 because B3's `render_agent_auth_state` seeds `by_harness = {}` and only inserts harnesses with a resolvable source — an unsatisfiable harness is *omitted* (read plane → native). A4 rewrote that same seed to `{selection.harness_kind: [] for selection in inputs.selections}` so a selected-but-unsatisfiable harness is kept with `sources: []` and the runtime refuses the launch.

Git merged both cleanly — A4 owns the dict seed, B3 owns the key-map lookup in `_render_gateway_source`, different lines, no conflict. A4's semantics survive, so B3's assertion now asserts the behavior A4 deliberately deleted.

**The merged behavior is correct; only the assertion is stale.** Ran B3's exact scenario against the merged renderer:
```
Skipping unsatisfiable gateway agent-auth source harness=codex (enrollment status=synced, virtual key present=False)
[{"harness_kind":"claude","sources":[{"kind":"gateway",...,"key":"sk-litellm-claude"}]},
 {"harness_kind":"codex","sources":[]}]
```
B3's stated intent — "a harness with no key of its own never borrows a sibling's key" — holds. `codex` gets `sources: []`, not claude's key. Only the *shape* of "no key" changed: absence → presence-with-empty-sources, which is what the spec now blesses ("absent means native, present-but-empty fails closed") and is strictly safer.

**Suggested fix (builder's call):** don't revert A4. Rewrite B3's assertion to the fail-closed shape, which still proves the no-borrowing property:
```python
by_harness = {h["harness_kind"]: h["sources"] for h in state["harnesses"]}
assert by_harness["claude"][0]["key"] == "sk-litellm-claude"
assert by_harness["codex"] == []      # fails closed; borrows nothing
```
Whichever of #1511/#1512 rebases second should carry the edit. If B3 lands first, A4's rebase must rewrite this assertion or A4 lands red.

**Why per-PR CI can't see it:** the collision isn't two edits to one line (git would conflict) — it's *one branch's test against another branch's behavior*, in a file each branch only half-owns. A4's CI runs A4's tests; B3's CI runs B3's tests. Nothing runs B3's tests against A4's renderer until the merge.

Full writeup + repro in `codex/ledger-qa.md` §F-038. Merged-tree status: Rust 1406/1406 green, Python 2 red (F-007, F-038), repo-shape 1 red (`check_max_lines`).

pablonyx:

```
```

```

```

```
```

pablonyx:

---

PR #1512, opened 2026-07-26
feat(server): per-harness gateway key-map renderer + org-member fix (B3)

pablonyx:
Track B, PR 3 of 3 (B1→B2→**B3**, agents-impl-plan.md §4). Stacked; base =
`agents/b2-per-harness-keys` (#1508).

## Spec sections quoted

model-gateway.md, Account model:
> Inside the team, one virtual key per (subject, harness), each granted its
> harness's access group by name.

agent-auth.md, gap this PR closes:
> **One shared gateway key, not per-harness scoped keys.** The state
> renderer resolves a single enrollment virtual key and fans it out to
> every gateway-selected harness.

## What changed

- `AgentAuthStateInputs.gateway_virtual_key: str | None` →
  `gateway_virtual_keys: Mapping[str, str]` (keyed by `harness_kind`).
- `_load_state_inputs`: for every gateway-selected harness, looks up that
  harness's own `agent_gateway_enrollment_key` row (B2) and decrypts its
  own key — no more one key fanned out to every selected harness.
- `_render_gateway_source`: resolves `inputs.gateway_virtual_keys.get(harness_kind)`
  instead of a single shared value.
- New `budget.get_gateway_enrollment_for_user`: resolves the **org**
  enrollment for a user's current membership (same resolution
  `ensure_org_enrollment`/`resolve_billing_subject_id_for_user` use via
  `resolve_organization_id_for_user`), falling back to the personal
  enrollment for org-less users. Consumed by both
  `is_gateway_budget_available` and `_load_state_inputs`, so the budget
  gate and the keys it guards always agree on which subject pays.

## Gap bullets closed

model-gateway.md:
> Sessions for org members hand out the personal enrollment's key (the
> state renderer and budget gate both resolve the personal enrollment), so
> org members' gateway spend lands on their personal subject today.

agent-auth.md:
> **One shared gateway key, not per-harness scoped keys.**
> **Org members launch on their personal key.**

## Contract fixture — explicit decision, not an oversight

The task brief said to "Create/update the contract fixture at
`fixtures/contracts/agent-auth-state`." I found a more specific, later
dispatch note in `codex/agents-docs-rewrite-plan.md` ("ROUND 1 FIRED"
section) that overrides this for the actual round-1 scope:

> B3 python-only, state.json SCHEMA unchanged — per-harness values only,
> fixture change deferred to A4

Verified against the Rust side
(`anyharness/crates/anyharness-lib/src/domains/agents/route_auth/state.rs`):
the wire shape (`harnesses[].sources[].{kind,base_url,key}`) is byte-for-byte
unchanged by this PR — only which key *value* gets substituted per harness
changed (previously one shared value, now per-harness distinct values), not
the shape itself. There is nothing new to snapshot yet, so I followed the
more specific instruction and did not add a fixture here; A4 (Rust
consumer, a later PR by another operator) creates it when it has a shape
to assert against. Flagging this explicitly per the contradiction protocol
rather than silently picking one instruction over the other.

## Tests

- New `server/tests/integration/test_agent_auth_org_member_gateway.py` (4
  tests): org member resolves the org enrollment (not personal);
  org-less user still resolves personal; draining the personal subject's
  credit doesn't block an org member (independent budget gates); the
  rendered gateway source's key is the org enrollment's own child key.
- Updated `test_agent_auth_materialization.py` (unit + integration) to the
  per-harness key map; `_inputs()` test helper gained a `gateway_virtual_key`
  convenience default (fans one value to every gateway-selection harness)
  plus a `gateway_virtual_keys` param for tests needing distinct per-harness
  values.
- Updated `test_agent_gateway_api.py` (`test_seeded_gateway_and_api_key_render_valid_v2`)
  and `test_agent_gateway_usage_credits.py` to seed the per-harness child
  key instead of the (now key-less) parent enrollment row. The two
  usage-credits tests that B2's PR body flagged as documenting an interim
  B2→B3 gap now assert the key renders correctly (gap closed).
- Full run: `cd server && DEBUG=true uv run pytest tests/unit tests/integration -q -k "agent_gateway or agent_auth or billing_limit_enforcement_llm or litellm"`
  → 229 passed, 0 failed.
- `uv run ruff check` clean on every touched file. `uv run mypy` on touched
  modules shows zero new errors (one pre-existing baseline note in
  `agent_auth.py`, present identically on `origin/main`, confirmed via
  `git stash` diff).

## Contradictions

The fixture-scope one above, resolved per the more specific dispatch note
and documented rather than silently dropped.

---

### 1. BLOCKER (money): org routing sent every hosted user to an unfunded subject

Routing an org member to their org enrollment unconditionally is not safe on
hosted, where **every** user gets a default personal org. The org billing
subject has no credit grant, and:

- `is_gateway_budget_available` returns `True` unconditionally for a grant-less
  subject (the `granted_usd <= 0` "no ledger, LiteLLM budget is the guardrail"
  branch), and
- the org team's default budget of `"0"` means **uncapped** in LiteLLM, not zero.

So both enforcement walls opened at once, the personal $2 free-credit grant was
never consulted, and spend was effectively unlimited.

**Funding-follows-attribution guard** (interim, conservative, reversible;
founder end-state ruling pending): `get_gateway_enrollment_for_user` uses the
org enrollment only when the org billing subject is actually funded — a positive
credit grant, or an explicitly configured non-default
`agent_gateway_default_org_budget_usd`. Otherwise it falls back to the personal
enrollment, i.e. pre-B3 behavior where the personal grant is the cap. Documented
in a docstring comment and as a new `model-gateway.md` gap bullet naming the
pending ruling.

The previous version of this suite **asserted the hole as a feature** — its
`test_org_member_budget_gate_checks_org_credit_not_personal` treated "draining
the member's personal grant does not block them" as desired, which is exactly
the unlimited-spend path (their only funding source was drained, yet nothing
stopped them). Replaced with assertions in both directions:

- `test_unfunded_org_member_is_blocked_by_a_drained_personal_grant` — unfunded
  org → the personal subject governs, so a drained personal grant blocks.
- `test_funded_org_budgets_are_independent_of_personal_credit` — funded org →
  the org subject governs; draining personal does not block, draining the org
  does.
- `test_funded_org_governs_the_members_gateway`,
  `test_unfunded_org_falls_back_to_the_personal_enrollment`,
  `test_explicit_org_budget_also_counts_as_funded`.

### 2. BLOCKER (divergence): other gateway-key readers resolved a different enrollment

Three call sites still used `get_enrollment_for_user` unconditionally while the
gate and renderer resolved the governing enrollment:

- `catalog.py` `_probe_gateway_models` — probed the **personal** key while the
  402 gate immediately above it checked the org subject, so enumeration ran
  against a key belonging to a different payer than the one authorized.
- `service.py` `get_capabilities` / `get_enrollment` — reported the personal
  `sync_status` while the renderer handed out the org enrollment's key, so the
  UI's readiness signal could describe an enrollment nobody uses.

All four readers now route through `get_gateway_enrollment_for_user`. New
`test_every_gateway_key_reader_resolves_the_same_enrollment` asserts gate,
probe-resolved key, capabilities/enrollment status, and the rendered key all
agree for a funded org member.

### 3. Missing per-harness distinctness coverage

`gateway_virtual_keys` had zero tests passing distinct values, so nothing proved
each harness carries its **own** key — the whole point of R2's scoping. Added:

- unit `test_each_gateway_harness_carries_its_own_key` and
  `test_harness_missing_from_the_key_map_is_dropped_alone` (a keyless harness
  never borrows a sibling's key), and
- integration `test_each_gateway_harness_renders_its_own_child_key`, driving
  `_load_state_inputs` against real child-key rows so a loader that fanned one
  key out (or read the wrong harness's row) fails.

### 4. Multi-org determinism pinned

Resolution takes the first active membership ordered by organization **name**
(`get_current_membership_for_user`), so renaming an org can move the payer. The
funded-subject guard narrows this, but
`test_multi_org_membership_picks_the_first_org_by_name` now pins the choice, and
the docstring records that the instability is inherited from the compute
attribution path rather than introduced here.

### Also

`scripts/max_lines_allowlist.txt`: `test_agent_gateway_api.py` bumped 654 → 665
— this branch already exceeded its entry (a CI red independent of the review
fixes).

### Rebase note (A4)

This branch does not contain `origin/agents/a4-fail-closed` (different stack).
None of these fixes touch `materialize/agent_auth.py`, so the A4 semantics
(pre-seeded `by_harness` dict, narrowed state-file delete, tripwire test) are
untouched here; reconciliation happens at merge time under A4's tripwire.

- `DEBUG=true uv run pytest tests/unit tests/integration -q -k "agent_gateway or agent_auth or billing_limit_enforcement_llm or litellm or materializ"`
  → **324 passed**, 0 failed.
- `uv run ruff check` + `ruff format --check` clean across `proliferate` and
  `tests`.
- `scripts/check_max_lines.py`, `check_migration_heads.py`,
  `check_server_boundaries.py`, `check_docs.py` all pass.
- `uv run mypy` on touched modules: no new errors (the one `service.py`
  entitlement-overload finding is pre-existing, confirmed on the pristine tree).

### DIFF
--- a/scripts/max_lines_allowlist.txt
+++ b/scripts/max_lines_allowlist.txt
@@ -44,7 +44,7 @@ server/proliferate/db/models/cloud/agent_gateway.py 556 agent_gateway_enrollment
 server/proliferate/server/cloud/agent_gateway/api.py 405 harness-settings GET/PUT routes added for catalog-driven per-harness settings
 server/proliferate/server/cloud/secrets/api.py 416 existing cloud secrets API pending route split
 server/proliferate/server/cloud/workspaces/service.py 1040 workspace materialization ledger service (intent/report/unlink write-path + primary selection + recorded-sandbox reconciliation + scratch/repo-less read guards) + PR 5 exact-ref managed-Cloud creation and association path; AnyHarness provisioning handoff split out per guides/domains.md
-server/tests/integration/test_agent_gateway_api.py 654 P1 auth rebuild rewrote keys/selections/state coverage for titled-key + selection model
+server/tests/integration/test_agent_gateway_api.py 665 P1 auth rebuild rewrote keys/selections/state coverage for titled-key + selection model; state coverage now seeds a per-harness child key (B3 renderer)
 server/tests/integration/test_cloud_workspace_materialization_service.py 1173 durable materialization ledger service suite (intent/report/unlink transitions, redaction, primary selection, recorded-sandbox reconciliation, repo-less workspace, install-id redaction, generation replay/crash-retry/terminal-reintent) + PR 5 exact-ref create coverage; new fail-closed source cases split to test_cloud_workspace_exact_ref_source.py
 server/tests/integration/test_auth_flow.py 2214 existing server oversized integration test plus session-token-revocation coverage + /users/me password rejection test + single-org current_product_user bypass coverage
 server/tests/integration/test_billing_accounting.py 872 existing server oversized integration test
--- a/server/proliferate/server/cloud/agent_gateway/budget.py
+++ b/server/proliferate/server/cloud/agent_gateway/budget.py
@@ -23,6 +23,8 @@ from sqlalchemy.ext.asyncio import AsyncSession
 
 from proliferate.config import settings
 from proliferate.db.store import agent_gateway as agent_gateway_store
+from proliferate.db.store.agent_gateway import AgentGatewayEnrollmentRecord
+from proliferate.db.store.billing_runtime_usage import resolve_organization_id_for_user
 
 _ZERO = Decimal("0")
 
@@ -33,6 +35,31 @@ _ZERO = Decimal("0")
 AGENT_GATEWAY_CREDITS_EXHAUSTED_CODE = "agent_gateway_credits_exhausted"
 
 
+async def get_gateway_enrollment_for_user(
+    db: AsyncSession,
+    user_id: UUID,
+) -> AgentGatewayEnrollmentRecord | None:
+    """The enrollment that governs a user's gateway sessions.
+
+    An org member (current membership, same resolution
+    ``resolve_billing_subject_id_for_user`` and ``ensure_org_enrollment`` use)
+    is governed by their **org** enrollment, not their personal one — closing
+    the model-gateway.md org-member gap where sessions previously always
+    resolved the personal enrollment regardless of org membership. An
+    org-less user keeps resolving their personal enrollment.
+    """
+    organization_id = await resolve_organization_id_for_user(db, user_id)
+    if organization_id is not None:
+        org_enrollment = await agent_gateway_store.get_enrollment_for_organization(
+            db,
+            organization_id=organization_id,
+            user_id=user_id,
+        )
+        if org_enrollment is not None:
+            return org_enrollment
+    return await agent_gateway_store.get_enrollment_for_user(db, user_id=user_id)
+
+
 async def is_gateway_budget_available(db: AsyncSession, user_id: UUID) -> bool:
     """Whether a user may launch a gateway-route session.
 
@@ -40,12 +67,13 @@ async def is_gateway_budget_available(db: AsyncSession, user_id: UUID) -> bool:
     or the user has no credit grant (default-budget subjects are never blocked
     on the ledger), or their remaining LLM credit is above zero. False only when
     a granted subject has spent its credit. Checks the same enrollment the state
-    renderer hands out key material for (the user's personal enrollment), so the
-    gate and the key it guards always agree on the paying subject.
+    renderer hands out key material for (the org enrollment for an org member,
+    else the personal one), so the gate and the keys it guards always agree on
+    the paying subject.
     """
     if not settings.agent_gateway_enabled:
         return True
-    enrollment = await agent_gateway_store.get_enrollment_for_user(db, user_id=user_id)
+    enrollment = await get_gateway_enrollment_for_user(db, user_id)
     if enrollment is None:
         return True
     balance = await agent_gateway_store.get_remaining_credit_usd(
--- a/server/proliferate/server/cloud/agent_gateway/budget.py
+++ b/server/proliferate/server/cloud/agent_gateway/budget.py
@@ -35,6 +35,36 @@ _ZERO = Decimal("0")
 AGENT_GATEWAY_CREDITS_EXHAUSTED_CODE = "agent_gateway_credits_exhausted"
 
 
+def _explicit_org_budget_configured() -> bool:
+    """Whether the deployment configured a real (non-default) org team budget.
+
+    ``agent_gateway_default_org_budget_usd`` defaults to "0", which LiteLLM
+    reads as *uncapped* rather than as zero — so the default is the absence of
+    a cap, not a cap of nothing.
+    """
+    raw = settings.agent_gateway_default_org_budget_usd.strip()
+    if not raw:
+        return False
+    try:
+        return Decimal(raw) > _ZERO
+    except ArithmeticError:
+        return False
+
+
+async def _subject_is_funded(db: AsyncSession, billing_subject_id: UUID) -> bool:
+    """Whether a billing subject actually funds gateway spend.
+
+    Funded means a positive LLM credit grant (the ledger the importer debits
+    and the gate reads) or an explicitly configured org team budget. A subject
+    with neither has no funding source at all, and every LLM guardrail reads as
+    "unlimited" for it (see :func:`get_gateway_enrollment_for_user`).
+    """
+    balance = await agent_gateway_store.get_remaining_credit_usd(db, billing_subject_id)
+    if balance.granted_usd > _ZERO:
+        return True
+    return _explicit_org_budget_configured()
+
+
 async def get_gateway_enrollment_for_user(
     db: AsyncSession,
     user_id: UUID,
@@ -43,10 +73,27 @@ async def get_gateway_enrollment_for_user(
 
     An org member (current membership, same resolution
     ``resolve_billing_subject_id_for_user`` and ``ensure_org_enrollment`` use)
-    is governed by their **org** enrollment, not their personal one — closing
-    the model-gateway.md org-member gap where sessions previously always
-    resolved the personal enrollment regardless of org membership. An
-    org-less user keeps resolving their personal enrollment.
+    is governed by their **org** enrollment rather than their personal one —
+    closing the model-gateway.md org-member gap where sessions previously
+    always resolved the personal enrollment regardless of org membership.
+
+    Funding-follows-attribution guard (interim; founder end-state ruling
+    pending). Routing to the org subject unconditionally is not safe on hosted,
+    where EVERY user gets a default personal org: an org billing subject with
+    no credit grant makes ``is_gateway_budget_available`` return ``True``
+    unconditionally (the ``granted_usd <= 0`` "no ledger, LiteLLM budget is the
+    guardrail" branch), while the org team's default budget of "0" means
+    *uncapped* in LiteLLM. Both walls open at once, the personal free-credit
+    grant is never consulted, and spend is unbounded. So the org enrollment
+    governs only when the org subject is actually funded
+    (:func:`_org_subject_is_funded`); otherwise this resolves the personal
+    enrollment — the pre-B3 behavior, where the personal grant is the cap.
+
+    Org choice is deterministic when a user holds several memberships: it is
+    the first active membership ordered by organization NAME
+    (``get_current_membership_for_user``), the same choice compute attribution
+    makes. Renaming an org can therefore move the payer; that instability is
+    inherited from the compute path and pinned by test, not introduced here.
     """
     organization_id = await resolve_organization_id_for_user(db, user_id)
     if organization_id is not None:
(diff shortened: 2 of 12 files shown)


pablonyx:
## QA fork — B3: merged renderer semantics contradict this PR's own test, and the resolution is a money decision

Reproduced on an integration branch = `origin/main` + a6 + **b3 (this PR)** + c1 + d2. Read-only QA; nothing pushed to any `agents/*` branch. Ledger: `codex/ledger-qa.md` F-003, F-009.

### BLOCKER — unit suite is red on the merged tree; invert the test, do **not** "fix" the renderer

```
tests/unit/test_agent_auth_materialization.py::test_harness_missing_from_the_key_map_is_dropped_alone
  AssertionError: assert ['claude','codex'] == ['claude']        (1 failed, 19 passed)
```

A6 changed `render_agent_auth_state` to **pre-seed** every selected harness (fail-closed, `sources: []`). B3 was written against `by_harness = {}` + `setdefault` — still the shape on `origin/agents/b3-renderer-key-map:133` — and its test asserts the harness is *dropped*.

The reason I'm flagging this as a decision rather than a test tidy-up:

- **A6 semantic (pre-seed):** withheld key → `sources: []` → the runtime **refuses** the launch. ✅
- **B3 semantic (drop):** the harness vanishes → the runtime reads absence as native → the user's **personal** Anthropic/OpenAI account silently pays. ❌ (`state.rs:117-133` — absent means native by design.)

A6's semantic is the correct one, so the fix is to rewrite this test to assert the post-A4 truth: codex is still present but with `sources: []` — i.e. it does not borrow claude's key **and** it does not silently vanish into native. The test's stated intent ("a harness with no key of its own never borrows a sibling's key") is still valuable; only its assertion mechanism is stale. Assert `sources == []` for codex rather than absence.

**Good news on the merge itself:** it resolved in the safe direction. The merged code carries A4's fail-closed seeding *and* B3's per-harness key map (verified by reading merged `materialize/agent_auth.py:145-165` + `:216`) — the named tripwire test did its job. Only B3's own test survived asserting the superseded law.

### Related — a dead tripwire that never fired

`tests/unit/test_agent_auth_state_contract_fixture.py::test_per_harness_gateway_keys_are_not_produced_yet` — its own comment instructs B3 to delete it, but it still passes **vacuously** because the shared `_inputs` helper fans a single `gateway_virtual_key` across gateway selections. So the "B3 has landed, delete me" tripwire never fired. Delete it in B3 and make the helper build a real per-harness map, otherwise the next per-harness change gets no protection either.

### Verified correct (so you don't re-check)

The Rust wire contract is untouched by B3 — `AgentAuthState` (`route_auth/state.rs:91-111`) is still snake_case with `user_id`/`issuing_server_origin`, and the contract fixture test that links the Python renderer to the Rust parser passes on the merged tree. `check_migration_heads.py` is clean (129 revisions, one head). The full agent-gateway/agent-auth **integration** surface — keys, selections, state rendering, enrollment, per-harness key map, cursor selections, org policy — is green on the merged stack (808 passed), which is the strongest evidence so far that the B-stack + C1 server contract holds together.

### Two merge-mechanics items for your rebase (no PR change needed now)

1. **`scripts/max_lines_allowlist.txt` will go red on merge — arithmetic certainty, not a flake.** The allowlist is a per-file *number*, and three branches each grew the same two test files while bumping only their own number:

   | file | main | a6 | b3 | c1 | **merged actual** |
   |---|---|---|---|---|---|
   | `test_agent_gateway_api.py` | 654 | 654 | 665 | 723 | **734** |
   | `test_agent_auth_materialization.py` | 480 | 614 | 545 | 480 | **679** |

   Each branch is internally consistent (actual == its own allowlisted value) so each is green alone; every branch's number is too low once the second lands, and the ratchet fails closed. `check_max_lines.py` is the **only** one of the 11 repo-shape ratchets that fails on the merged stack (boundaries, old_paths, appearance_scaling, docs, frontend_boundaries, migration_heads, worker_structure, server_boundaries, session_mutation_admission, workflow_managed_boundaries all pass). Mechanical fix: set 734 and 679. Per the repo's own standard these files are 130 and 79 lines over the 600 max, so a split (the C1 cursor-selection cases and the fail-closed renderer cases are natural seams) would retire the entries instead of raising them again — your call.
2. **Spec gap-list conflicts on retarget.** `agent-auth.md` "Current gaps": A4 struck the silent-degradation bullet while B3 struck the one-shared-key bullet — deletion-vs-deletion at adjacent lines, so git keeps each side's retained bullet. Correct resolution is the **union of deletions** (drop both). Same shape for A5 (codex-second-home + claude-sanitization) vs C1 (cursor-rejected + cloud-login). And `max_lines_allowlist.txt` conflicts between b3 and c1 on the same file — take the higher number plus concatenated rationale.

Three of my first six findings were this "green alone, red merged" class, which per-PR CI structurally cannot catch. Worth one integration-merge dry run before flagging the stack ready — that's exactly what my `qa/agents-stack-integration` branch is.

### One env gotcha that may have affected your own test runs

The **shared** `server/.venv` has its editable install pointed at another job's worktree (`/Users/pablohansen/.claude/jobs/312c945f/tmp/wt-integration`), so anything run through it tests the *wrong code*. I built a separate venv in the QA worktree for these results. If you ran server tests through the shared venv during this program, they may not have tested your branch.

pablonyx:
**QA (integration) — F-038: merge-only red between #1511 (a4) and #1512 (b3)**

Found on a merged QA branch (a-stack head + b-stack head + C1 + D2 + A7). Not a defect on either branch alone — read-only QA, no changes pushed to either branch.

**The red**
```
$ cd server && DEBUG=true uv run pytest tests/unit/test_agent_auth_materialization.py
FAILED TestRenderAgentAuthState::test_harness_missing_from_the_key_map_is_dropped_alone
  AssertionError: assert ['claude', 'codex'] == ['claude']
  Left contains one more item: 'codex'
```

**Provenance** — the test case exists on exactly one ref:
```
0  origin/main
0  origin/agents/a4-fail-closed
1  origin/agents/b3-renderer-key-map
0  origin/agents/b2-per-harness-keys
0  origin/agents/b1-access-group-tags
0  origin/agents/c1-cloud-login-cursor-ui
```
It's B3's test, and it passes on B3 because B3's `render_agent_auth_state` seeds `by_harness = {}` and only inserts harnesses with a resolvable source — an unsatisfiable harness is *omitted* (read plane → native). A4 rewrote that same seed to `{selection.harness_kind: [] for selection in inputs.selections}` so a selected-but-unsatisfiable harness is kept with `sources: []` and the runtime refuses the launch.

Git merged both cleanly — A4 owns the dict seed, B3 owns the key-map lookup in `_render_gateway_source`, different lines, no conflict. A4's semantics survive, so B3's assertion now asserts the behavior A4 deliberately deleted.

**The merged behavior is correct; only the assertion is stale.** Ran B3's exact scenario against the merged renderer:
```
Skipping unsatisfiable gateway agent-auth source harness=codex (enrollment status=synced, virtual key present=False)
[{"harness_kind":"claude","sources":[{"kind":"gateway",...,"key":"sk-litellm-claude"}]},
 {"harness_kind":"codex","sources":[]}]
```
B3's stated intent — "a harness with no key of its own never borrows a sibling's key" — holds. `codex` gets `sources: []`, not claude's key. Only the *shape* of "no key" changed: absence → presence-with-empty-sources, which is what the spec now blesses ("absent means native, present-but-empty fails closed") and is strictly safer.

**Suggested fix (builder's call):** don't revert A4. Rewrite B3's assertion to the fail-closed shape, which still proves the no-borrowing property:
```python
by_harness = {h["harness_kind"]: h["sources"] for h in state["harnesses"]}
assert by_harness["claude"][0]["key"] == "sk-litellm-claude"
assert by_harness["codex"] == []      # fails closed; borrows nothing
```
Whichever of #1511/#1512 rebases second should carry the edit. If B3 lands first, A4's rebase must rewrite this assertion or A4 lands red.

**Why per-PR CI can't see it:** the collision isn't two edits to one line (git would conflict) — it's *one branch's test against another branch's behavior*, in a file each branch only half-owns. A4's CI runs A4's tests; B3's CI runs B3's tests. Nothing runs B3's tests against A4's renderer until the merge.

Full writeup + repro in `codex/ledger-qa.md` §F-038. Merged-tree status: Rust 1406/1406 green, Python 2 red (F-007, F-038), repo-shape 1 red (`check_max_lines`).

pablonyx:

```
```

```

```

```
```

pablonyx:
**F-038 fixed** at `bf9800a64`.

Per Pablo's ruling, did NOT revert a4's fail-closed law. Adopted a4's `render_agent_auth_state` dict-seed hunk verbatim (including its comments) — `by_harness` now seeds `{selection.harness_kind: [] for selection in inputs.selections}` instead of an empty dict + `setdefault`, so a selected-but-unsatisfiable harness keeps a `sources: []` entry rather than being dropped. Updated the module docstring and the materializer's delete-file comment to match, for merge identity.

Rewrote every B3 test assertion that depended on the old drop-the-harness shape (7 tests across `TestRenderAgentAuthState` and `TestMaterializeAgentAuth`), including renaming the ledger's named test to `test_harness_missing_from_the_key_map_never_borrows_a_sibling_key`: it now asserts claude gets its gateway source AND codex gets `sources: []` (fails closed, borrows nothing) — still fails if the no-borrowing property regresses.

Evidence:
- `cd server && DEBUG=true uv run pytest tests/unit/test_agent_auth_materialization.py -q` → **19 passed**
- `git merge-tree --write-tree bf9800a64 origin/agents/a4-fail-closed` → `server/proliferate/server/cloud/materialization/materialize/agent_auth.py` merges with **zero diff** (byte-identical on the shared seed hunk, confirmed via direct diff of the merged blob). The test file now shows a normal, visible merge conflict (both branches independently rewrote overlapping assertions) — not the silent merge-only red F-038 described.
- `scripts/check_max_lines.py` passes; `test_agent_auth_materialization.py` is exactly 600 lines (threshold), no allowlist entry touched.
- ruff check/format clean on both edited files.

pablonyx:

---

PR #1513, opened 2026-07-26
fix(agent-auth): codex native recipe + sanitization on every route (A5)

pablonyx:
Track A, PR 5 — Round 2. **Stacked on #1511** (`agents/a4-fail-closed`); retarget as parents merge. The diff below is A5's own commit only.

## Gap bullets closed (deleted in this PR)

1. **"Codex has a second, competing isolated home."** — *"Every codex launch — including gateway-routed ones — also writes `agent-auth/codex-local/` with a hardcoded `config.toml` pinning `model = "gpt-5.5"` (a Rust-constant model pin, violating the catalog-owns-model-names law) and a copy of the user's native `auth.json`; route_auth's `CODEX_HOME` then shadows it for routed launches… Fold the native codex home preparation into the route-auth recipe table as the native recipe, sourced from the catalog."*
2. **"Claude's ambient sanitization only runs on the gateway route."** — *"An `api_key` selection sets its env var but does not strip ambient `CLAUDE_CODE_USE_BEDROCK`/`CLAUDE_CODE_USE_VERTEX`, so a Bedrock-configured host reroutes a BYOK launch."*

## One shape behind both

The recipe table only covered the **routed** cases. So the native cases leaked into other code paths and diverged from their routed siblings. Both fixes are the same move: bring native into the table.

### Codex's native home

`render_profile` now takes `harness_kind`, so its `Native` branch can dispatch to a `render_native` table. Four harnesses render nothing there — genuinely nothing, because their CLIs find their own credentials and config. Codex is the exception, and that exception is the point: codex reads its model from `config.toml` rather than from env, so *something* has to render that file.

What changed materially:

| | before | after |
| --- | --- | --- |
| model | `model = "gpt-5.5"`, a Rust constant | catalog's `native_default_model` |
| written when | **every** codex launch, routed included | only a native launch |
| home | `agent-auth/codex-local/` | `agent-auth/codex-native/` |
| credential material | a copy of the user's `auth.json`, written on every launch including routed ones | a copy of the user's `auth.json`, written only on a native launch (**corrected in review round 2** — see H1 below; the first draft of this PR wrote none, which was the blocker) |
| other keys | `model_reasoning_effort`, `[features]` toggles, invented in Rust | none |

**Stable, not revision-keyed.** A native launch has no selection to key on, and a revision-keyed native home would be a GC target for a later *routed* launch of the same family — pulling config out from under a running native session. Being separate families also means a native and a routed session coexist, which the old implementation could not do (it wrote both on the same launch and let one shadow the other).

**No catalog default renders nothing**, rather than a hardcoded fallback. That is the honest degradation: a missing catalog value is exactly what the deleted constant was papering over, and unlike the gateway route a native launch *can* proceed on the CLI's own config. Rendering nothing is also what keeps this branch *safe* after H1: no `CODEX_HOME` is set at all, so the CLI reads its own `~/.codex` — config and credential together — and the degradation is "unmanaged", never "authenticated against nothing".

### H1 (review round 2): an isolated home must carry the credential it isolates away from

The first draft of this PR deleted `sync_local_codex_auth` — the sole writer of `auth.json` into an isolated codex home — and rendered only `config.toml`. That was a **launch-breaking** regression, not a cleanup:

- Codex resolves credentials at **`$CODEX_HOME/auth.json`**, not at `~/.codex`, once the var is set (`credential-discovery/src/codex.rs`). Pointing `CODEX_HOME` at `codex-native/` therefore moves the *credential lookup* along with the config.
- Nothing later repairs it. The login terminal adjusts only `PATH` (`auth/login.rs::login_env`), so `codex login` writes `~/.codex` and never the isolated home.
- So a natively-logged-in user launched **unauthenticated**, and `native_render_tests.rs` pinned exactly that by asserting `auth.json` was ABSENT.

Fixed by making the credential the native recipe's **second file**: a new `PathFamily::CodexNativeAuth` whose `contents` is `None` because its bytes are not rendered — they are read at apply time from the user's real codex home through `export_portable_auth(ProviderId::Codex)`, the same mechanism the deleted code used. Render stays pure (contract §4); only apply touches the outside world, and it is the one apply arm that reads outside the runtime home, which the doc comment says and explains.

Degradation preserved from `sync_local_codex_auth`: no resolvable `HOME`, no codex login, or an unreadable one → remove any stale copy, log, and let the launch proceed (codex prompts for a login itself). A credential must never outlive the login that produced it.

### M2 (review round 2): macOS keychain — decided and documented, not silent

Codex's keychain entry is keyed on `sha256(canonical CODEX_HOME)` (`codex_keychain_account`), so a relocated home can never find the entry the user's `codex login` created — and `codex-keychain` is a first-class discovery kind, so this is a real population.

**It is covered**, and by the same fix rather than a second one: `export_portable_auth` reads the **keychain first** and returns its payload as `.codex/auth.json` bytes. A keychain-authenticated user's credential therefore arrives in the isolated home as a *file* the relocated home can read. The lookup resolves against the user's real `~/.codex` because the `CODEX_HOME` we render goes into the **child's** env, never this process's, so the account hash computed at apply time is the one the user actually has.

**Accepted residual, stated in the spec text and here:** a token refresh the child performs lands in the isolated home rather than back in the user's keychain. It cannot diverge or accumulate, because every launch re-copies from the source — the user's own `codex login` stays the authority, which is the same relationship the previous implementation had. No re-login is required of anyone.

`native_default_model` is resolved from `session.defaults` over the **non-gateway** contexts in precedence order (`openai-oauth` → `openai-api` → the anthropic pair), with an any-non-gateway fallback so a newly-added harness is configured before its context id is listed. It deliberately cannot fall back to `defaults["gateway"]`: a gateway-only model id would 404 against the user's own provider account.

### Claude sanitization on every route

It lived inside `render_claude_gateway`, so an `api_key` route set its var and left the ambient rerouting flags standing. On a Bedrock-configured host, a BYOK launch went to Bedrock while the key the user actually selected sat unused in the env — billing an account they did not choose.

It now runs once in `render_sources` over the **fully composed** delta. Composed matters: the rule is "keep whatever this route set, remove the rest", so running it per-source would let an earlier source's var be removed on behalf of a later one. It is claude-only, because no other harness's CLI reroutes on these flags and adding removals elsewhere would be an unexplained env change.

`CLAUDE_CODE_USE_FOUNDRY` joins the removal list now rather than with Track D's Foundry support: the flag reroutes an ambient host **today** whether or not we can configure Foundry ourselves, so leaving it out would be a hole with no upside.

## The D3 seam (asked for explicitly)

Every codex `config.toml` in the system is now emitted by one `codex_config_toml` over a `CodexConfigRecipe` enum. Track D adds variants **beside** their siblings, with no existing arm changing:

```rust
enum CodexConfigRecipe<'a> {
    Native  { default_model: &'a str },                        // pins only the model
    Gateway { base_url: &'a str, default_model: &'a str },     // [model_providers.proliferate]
    // D3 adds:
    // Azure   { .. } → [model_providers.azure] with wire_api = "responses"
    // Bedrock { .. } → model_provider = "amazon-bedrock", NO [model_providers.*] table
}
```

The enum's doc comment carries those two target shapes so D3 does not have to re-derive them. Adding either is one arm here plus one `ResolvedSource::ProviderConfig` arm in `render_sources`.

Also fixed: `catalog_probe.rs`'s comment claimed *"codex has no native Bedrock support"*. It does — `amazon-bedrock` is a built-in upstream provider, which is exactly why D3's Bedrock variant needs **no** provider table. The probe's custom `[model_providers.bedrock]` is a probe-specific choice (it enumerates what is reachable over `/v1/responses`, which for Bedrock means the OpenAI-compatible "mantle" surface), not a statement about codex's capabilities. The comment now says which is which.

## Tests

Rewritten because it pinned the bug:
- `claude_api_key_sets_exactly_its_var` asserted `rendered.remove.is_empty()` — i.e. it pinned the reverse-contamination gap as intended behavior. Now `claude_api_key_sets_its_var_and_still_sanitizes_ambient`, which keeps the generic-recipe assertions (one var set, no files) *and* asserts the removals.
- `launch_env_tests.rs`'s two codex tests asserted the deleted home's contents (`model = "gpt-5.5"`, the copied `auth.json`). Replaced by `build_session_launch_env_contributes_no_codex_home` — the inverse assertion, and the one that catches a regression where both layers write a `CODEX_HOME` again. The *credential* half of what they asserted did not disappear with them; it moved to `native_render_tests.rs` (see H1).

New (tier 1):

| Test | Proves |
| --- | --- |
| `native_codex_renders_a_catalog_pinned_config_in_an_isolated_home` | exact config bytes (`model = "…"\n`, nothing else); the home is `codex-native`, **not** revision-keyed; no removals; and the recipe requests the credential family. Uses a resolver whose gateway `default_model` is a sentinel, so reading the wrong plan field fails |
| `a_user_with_a_codex_login_launches_native_codex_authenticated` | **H1's regression test.** A temp `HOME` carrying a real `~/.codex/auth.json`; asserts the credential is readable under the relocated `CODEX_HOME`. This is the assertion the deleted `build_session_launch_env_sets_clean_codex_home_for_local_codex` used to carry, restored at the layer that now owns it |
| `native_codex_without_a_user_login_delivers_nothing_and_clears_a_stale_copy` | the preserved degradation: an unauthenticated machine still launches, and a copy from a previous login is removed rather than outliving it |
| `a_repeated_native_codex_launch_rewrites_the_same_home` | idempotent and current — why stable-not-revision-keyed is safe |
| `native_codex_without_a_catalog_default_renders_nothing` | no hardcoded fallback; no home created at all |
| `every_other_harnesss_native_launch_is_an_empty_delta` | the other four still render nothing |
| `the_native_and_routed_codex_homes_do_not_collide` | different dirs, both configs survive, genuinely different recipes — the shadowing bug, inverted |
| `a_non_claude_api_key_route_adds_no_removals` | sanitization stays claude-specific |
| `build_session_launch_env_is_empty_for_every_non_claude_harness` | that layer is claude-only now |
| 5 × `native_default_model` (pure) | never falls back to the gateway model; precedence order; any-non-gateway fallback; **the real bundled catalog gives codex a native default** (without it the native recipe would silently render nothing on every machine); and **opencode's fallback value is pinned explicitly** — the any-non-gateway branch fires today for opencode (whose only non-gateway default is `baseline`), so its result should be asserted rather than inferred. The test also pins opencode's context key set, so a real auth-context default arriving there is a visible diff |

Results: `cargo test -p anyharness-lib --lib` **1336/1336**. `cargo test --workspace` green except the known pre-existing `apps/desktop/src-tauri` flake (documented in #1503; different test per run, green in isolation, untouched here). `check_docs.py` + every `repo-shape` check pass. `ruff check` clean.

## Contradictions encountered

None ruling-worthy. Four judgment calls:

1. **`codex-local` is retained as a READ path in `portability/codex.rs`.** Sessions created before this rename have their rollouts under the old directory, and that collector's job is finding an *existing* session's artifacts — dropping the path would silently lose mobility for every pre-rename session. Nothing writes there any more; a comment says when it can go.
2. **`CLAUDE_CODE_USE_FOUNDRY` added ahead of Foundry support** (reasoning above).
3. **The native codex home is isolated at all**, even though the credential is the user's own. Codex reads `~/.codex/config.toml`, and the runtime must not silently inherit a developer's leftover model/provider/feature flags — a session launched from the product would then run a configuration the product does not know about. Isolation also keeps the two routes symmetric.
4. **`render_tests.rs` split rather than allowlisted a second time.** Per the carry-over instruction: it crossed the cap again, so the opencode section moved to `opencode_render_tests.rs` and the file is now **488 lines, off the allowlist entirely** (the A4-era entry is deleted, not bumped). `gateway_resolver.rs`'s entry did grow (827 → 869 after round 2's opencode pin); that file was already far over and is a documented split-later case.
5. **Why `baseline` was NOT added to `NATIVE_CONTEXT_PRECEDENCE`** (round 2, the alternative the reviewer offered): `baseline` is a catalog-wide "model absent any context", not an auth context, so listing it would make it outrank a real provider context for any harness declaring both. The generic any-non-gateway fallback already picks it when it is the only candidate, which is the intended precedence — so the fix is an explicit assertion, not a list entry.

- **§1.3's root-substitution trick still holds.** The design's phase B passes a scratch root to `render_profile` and relies on every path being a deterministic join through `materialize::*`. `codex_native_home_path` is exactly that shape, so a probe materializing a native codex context relocates with everything else.
- **§1.3's scoping table maps `envFlag`/discovery-only contexts to `Native`**, which now renders codex's config instead of nothing. That is *more* launch-faithful, not less — the design's governing principle is "probe env == launch env for that context", and a native codex launch really does get this config.
- **C5's `env_remove` plumbing.** The design flags that `probe_agent` cannot honor removals today and must gain `auth_env_remove`. A5 makes `rendered.remove` non-empty for claude `api_key` contexts too, which *widens* what C5's fix must carry — worth noting for A7, and it strengthens the design's open question 5 (an `api_key` probe on a Bedrock-exporting shell) rather than contradicting it: after A5 the launch itself strips those flags, so the launch-faithful answer for that context is now the sanitized one. **Round 2: written up as an errata entry in `codex/probe-engine-design.md`** beside its C5 section (widened set incl. `CLAUDE_CODE_USE_FOUNDRY`, moved anchors, and the note that a native codex probe now materializes two files rather than none), so A7 cannot miss it.


pablonyx:
## QA fork — A5 findings (static review + spec cross-check on a merged integration branch)

Reviewed on `origin/main` + a6 + b3 + c1 + d2. Read-only QA; nothing pushed to any `agents/*` branch. Full detail in `codex/ledger-qa.md` (F-021…F-024, F-027).

The common thread: A5 gets **claude's gateway route** right — isolated `CLAUDE_CONFIG_DIR`, ambient rerouting flags stripped — and that protection doesn't extend to the other three combinations. `agent-auth.md:345-348` states the law generally ("An ambient `ANTHROPIC_API_KEY` on the host can never shadow or leak into a routed launch"), so each gap below is a real contradiction, not a missing nice-to-have.

### MAJOR — grok's gateway recipe sanitizes nothing; ambient `GROK_API_KEY` can bill the user personally

`render.rs:515-535` sets `HOME`, `GROK_MODELS_BASE_URL`, `XAI_API_KEY` and pushes **no** `remove` entries. The registry declares grok's slot env as `["XAI_API_KEY","GROK_API_KEY"]` (`catalogs/agents/registry.json:462-465`) — `GROK_API_KEY` *is* a credential the CLI honors. So: a developer shell exports `GROK_API_KEY=<personal xAI key>`, the user selects the gateway route for grok, both vars reach the child (`process.rs:91` `.envs(...)`, no `env_clear`), and if the CLI prefers `GROK_API_KEY` the launch bills the user's personal xAI account while the gateway key sits unused. Claude gets this protection; grok doesn't.

### MAJOR — claude api_key (BYOK) leaves `CLAUDE_CONFIG_DIR` ambient, so `~/.claude` can undo the sanitization

`render.rs:215-221` sets `CLAUDE_CONFIG_DIR` only inside `render_claude_gateway`; the api_key arm (`:138-141`) sets just the named var. `sanitize_claude_if_routed` (`:154`) removes *env*, not config *files*. If the host's `~/.claude/settings.json` carries an `env` block (`ANTHROPIC_BASE_URL` / `ANTHROPIC_AUTH_TOKEN` / `CLAUDE_CODE_USE_BEDROCK`) or a stale OAuth credential, the CLI re-derives provider settings from its own config *after* our env sanitization — defeating it, and leaving the personal OAuth login reachable on a route the user chose specifically to bill their own key.

**The spec is self-contradictory here and needs your ruling:** `agent-auth.md:363` claude row, api_key cell = "the named env var" (no isolated config dir), while `:392-399` claims "A host configured for Bedrock cannot silently reroute a gateway or **BYOK** launch." Those cannot both hold.

### MAJOR — codex api_key leaves `CODEX_HOME` ambient, so the personal login governs a BYOK launch

`render_sources` (`render.rs:130-156`) has no codex-specific api_key arm, and `render_native`'s codex home (`:338-375`) is only reached on the Native profile. A user logged into codex natively (`~/.codex/auth.json` or keychain) who selects an api_key route gets a launch that reads `~/.codex/config.toml` + `~/.codex/auth.json` — model *and* credential both their own, with the selected key possibly never consulted (codex prefers `auth.json`). Same class as the claude case but with stronger precedence risk, and here the spec is unambiguous: `agent-auth.md:365` says "An isolated home must carry the credential it isolates away from… setting `CODEX_HOME` moves the credential lookup along with the config." On the api_key route no `CODEX_HOME` is set at all.

### MINOR — `protected_env_keys` is dead data, and the renderer's hand-maintained duplicate has already drifted

`render.rs:238-262` removes the three rerouting flags + `AWS_BEARER_TOKEN_BEDROCK` and unsets Anthropic selectors. But `registry.json:60-65` declares claude's `gatewayEnv.protectedEnvKeys` as `[ANTHROPIC_AUTH_TOKEN, ANTHROPIC_BASE_URL, ANTHROPIC_CUSTOM_HEADERS, CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST]` — **the last two are never removed**, so an ambient `ANTHROPIC_CUSTOM_HEADERS` reaches a routed launch. `protected_env_keys` is projected (`projection.rs:201-207`) and then read by *nothing* (grep finds only struct definitions). Same shape for codex: `render.rs:296-297` removes `OPENAI_API_KEY`/`ANTHROPIC_API_KEY` but not `CODEX_API_KEY`, which `registry.json:168-177` declares both as a slot credential and as protected (lower risk — `config.toml` pins `env_key = "PROLIFERATE_GATEWAY_KEY"`).

Cleanest fix: make the renderer **read** `protected_env_keys` instead of duplicating it. That closes the drift permanently and gives the registry declaration a purpose.

### MINOR — native codex credential copy lands in a stable, non-user-keyed dir

`materialize.rs:199-241` reads `export_portable_auth(Codex, dirs::home_dir())` and writes to `codex_native_home_path` (`:120-122`) = `agent-auth/codex-native` — not revision- or user-keyed. `state.rs:100` parses `user_id` and grep confirms no runtime code reads it; `apply_state_file`'s revision guard (`:185-199`) isn't user-scoped either. On a shared runtime home the copy persists across users until the next native codex launch re-copies or clears it. Mitigated on single-user desktop plus the unconditional re-copy / `remove_stale_file` on every native launch — hence minor — but worth noting that `user_id` currently enforces nothing.

pablonyx:

---

PR #1514, opened 2026-07-26
feat(agents): auto-install the supported set at startup (A6)

pablonyx:
Track A, PR 6 — Round 2. **Stacked on #1513** (`agents/a5-codex-config`); retarget as parents merge. The diff below is A6's own commit only.

## The law this implements

`specs/codebase/platforms/product/agent-distribution.md`, "Installation":

> Installation is automatic. Every harness supported on a surface converges with no user action: absent means install, drifted means reinstall, and both are the same mechanism — the reconcile job, triggered by the startup pass on every runtime boot… **A user authenticates harnesses; they never install them.** Two carve-outs:
> - An agent the user already provides on PATH is left alone: it is usable through readiness as-is, and a managed install would shadow their copy.
> - Cursor never installs in cloud. It is login-only with no headless credential path, so a cloud install could never reach `Ready`.

## Gap bullet closed (deleted in this PR)

> Installs are not yet automatic: the startup pass runs an `installed_only` reconcile (`reconcile_installed_when_idle` hardcodes it), so an absent opencode/grok stays `InstallRequired` until a user clicks install, and session creation rejects a non-`Ready` harness rather than converging it. The auto-install law above (full supported set, PATH and cloud-cursor carve-outs) is not yet implemented.

## Why the flag could not just be dropped

This is the part worth reviewing carefully. `installed_only` skipped every agent that was not already managed-installed — which means it skipped **PATH-provided agents as a side effect** of skipping absent ones:

```rust
// before: one boolean doing two unrelated jobs
if installed_only {
    let managed_installed = /* agent_process or native is source=="managed" */;
    if !managed_installed { return Skipped("installs on demand at session start"); }
}
```

Dropping the flag to get auto-install would therefore have **silently dropped the PATH protection**, and a managed install would land on top of a user's own binary. The plan's requirement that auto-install ride "behind an explicit tested predicate" is exactly this hazard.

So the carve-outs move into `installer::auto_install::auto_install_decision` — a pure fn over facts the caller gathers, no IO and no env reads:

```rust
pub fn auto_install_decision(
    kind: &AgentKind, surface: RuntimeSurface, installed_only: bool, facts: AgentInstallFacts,
) -> Result<(), AutoInstallSkip>
```

**Ordering is deliberate and asserted.** PATH is checked FIRST, ahead of the surface and scope rules, because it is the one carve-out protecting something the user owns: no combination of scope or surface may result in clobbering it. `path_protection_outranks_the_cursor_and_scope_rules` pins that a cursor-in-cloud agent on PATH reports the **PATH** reason — both would skip, but only one of them is about protecting the user, and an ordering regression here is the one that silently overwrites their binary.

**One deliberate asymmetry:** a machine with both a managed copy and a PATH copy keeps converging the **managed** one. Skipping it would strand our own install at an old pin forever, and we are not shadowing anything — we already installed alongside it.

## The surface signal

Cursor-in-cloud needs to know the surface. `RuntimeSurface` is a **constructor argument** on `AgentRuntime`, read from env once at wiring (`app/mod.rs`), rather than an env read at the decision point — otherwise the predicate would not be a pure function of its inputs and could not be unit-tested as a matrix.

The cloud bootstrap declares it (`ANYHARNESS_RUNTIME_SURFACE=cloud`), following the existing `ANYHARNESS_DEFER_STARTUP_RETENTION` precedent in the same dict. **Absent means local, and that default is the safe one**: it can only ever *enable* a cursor auto-install that then fails to reach `Ready` (wasted download), never *suppress* an install a surface needs (a broken harness).

`installed_only` survives as the **scoped settings action** ("update installed agents"), where it still means exactly what it says. It is simply no longer what the startup pass uses. Seed hydration still runs first, so seeded agents (claude/codex on desktop) are already present and the pass is a no-op for them rather than a redundant download.

## Tests — the matrix that had zero coverage

`auto_install_tests.rs` enumerates surface × scope × {absent, PATH-only, managed-only, both} × harness. The **PATH column had no coverage at all** before, because the protection was an accident of scope rather than a rule.

| Test | Proves |
| --- | --- |
| `an_absent_agent_auto_installs_on_a_full_pass` | the law itself, every surface × every gateway-capable harness |
| `a_path_provided_agent_is_never_clobbered` | 2 surfaces × 2 scopes × 5 harnesses = 20 cells, all skip |
| `path_protection_outranks_the_cursor_and_scope_rules` | precedence — the ordering regression that overwrites a user's binary |
| `a_managed_copy_keeps_converging_even_beside_a_path_copy` | the asymmetry above |
| `cursor_does_not_install_in_cloud_but_does_locally` | both halves; the carve-out is cloud-only |
| `no_other_harness_is_suppressed_in_cloud` | the cloud carve-out is cursor-specific |
| `an_installed_only_pass_still_updates_only_managed_installs` | the scoped action still works |
| `a_managed_agent_converges_in_either_scope` | the drift planner, not this predicate, decides if work is needed |
| `every_skip_reason_has_a_distinct_message` | the messages are a contract (T3-AGENT-1 asserts the PATH one), not log strings |
| `fact_constants_match_their_constructor` | the matrix's own consts cannot desync |

Rewritten: `internal_poke_waits_for_compatible_job_then_runs_a_fresh_pass_on_active_pins` asserted `fresh.installed_only` — now asserts `!fresh.installed_only` with a comment naming A6. It also now waits for **admission** rather than completion: the startup poke is full-scope, so completing it would really download and install every supported harness into a temp home (minutes of network in a unit test). The install mechanics have their own coverage; what belongs there is that the poke survives a busy slot and admits the right *shape* of job.

### Round 2: the surface threading, covered end to end (`surface_threading_tests.rs`)

Review round 2 found the gap that weakening left: `auto_install_tests.rs` covers the predicate as a pure function, and the poke test stops at job registration — so **nothing exercised `start_with_admission` → `run_reconcile_job` → `auto_install_decision`**, three hops of plumbing carrying the new `surface` argument.

Two new tests run a full-scope pass to **completion with no network**, which is affordable because a `None` catalog resolves no pins and the installer's `require_source` fence fails every agent deterministically before any download:

| Test | Proves |
| --- | --- |
| `full_scope_pass_admits_every_absent_agent_instead_of_skipping_it` | on `Local`, every agent is **admitted** by the predicate and then `Failed` with `"has no resolved source pin in the catalog lockfile"`. The admitted-then-failed pair is the assertion — the exact inverse of `installed_only_skips_uninstalled_agents`, which shows the same `None`-catalog setup yielding `Skipped` for every agent |
| `the_cloud_surface_reaches_the_predicate_and_carves_out_only_cursor` | the same pass on `Cloud` skips **exactly one** agent, cursor, with its named reason. Same inputs, one variable — a `surface` that never reached the predicate could not produce both results |

**Hermetic without touching `PATH`.** The PATH carve-out outranks scope and surface by design, so on any developer's machine (claude/codex/cursor/grok all commonly on PATH) these agents would be skipped as user-provided and the tests would assert nothing. Managed artifacts are seeded instead — both roles, since claude and codex resolve their *native* artifact from PATH — relying on the documented `has_managed_artifact`-beats-`has_path_artifact` precedence. Narrowing process-global `PATH` was tried first and broke the installer's `git`/`npm` shell-outs in sibling tests.

Split into its own file rather than pushing `execution_tests.rs` past the 600-line cap (it landed at 698).

New Python test `test_declares_the_cloud_surface_for_the_auto_install_carve_out`.

Results (round 2): `cargo test -p anyharness-lib --lib` green (`installer` module 66/66 including the two new tests). `cargo clippy` shows **no** findings in any file this PR touches — round 2 deleted `AgentRuntime::surface()`, which had zero callers and would have been a `dead_code` hard error under `-D warnings`. `cargo test --workspace` green except the known pre-existing `apps/desktop/src-tauri` flake (documented in #1503; different test per run, green in isolation, untouched here). Server `pytest tests/unit/test_runtime_version_pin.py` 7/7. `ruff check` + `format --check` clean. `check_docs.py` + every `repo-shape` check pass.

## Matches what T3-AGENT-1 will assert (brief §5)

Checked the brief's observability requirements against what this PR actually emits:

- **`installed_only` field semantics** — on the wire (`ReconcileAgentsResponse.installed_only`), and `false` on the startup pass. The brief's startup cell can distinguish a full pass from a scoped one.
- **Skipped-for-PATH results** — `AgentReconcileOutcome::Skipped` with `AutoInstallSkip::UserProvidedOnPath.message()`, which contains the literal `"PATH"` (asserted by a unit test, so the scenario's substring match is a contract rather than a log string). Distinct from the cursor-in-cloud and installed-only messages, so the scenario can tell the three skips apart.
- **Reconcile status transitions** — unchanged: `Idle|Queued|Running|Completed|Failed`, same job slot, same admission policy. The brief's "poll immediately from world boot until Completed/Failed, never leave on Idle+zero results" works exactly as written.
- **§5's "A6 drops installed_only → startup pass installs missing managed agents"** — done, and the brief's note that the startup cell is `blocked/expected_fail` pre-A6 can be flipped to a real assert once this merges.

The brief's own remaining work (the PATH-shadow fixture, `getReconcileStatus()` on `LocalRuntimeClient`, probing `already_installed` semantics live) is unchanged by this PR and stays with that later scenario PR.

## Contradictions encountered

None ruling-worthy. Three judgment calls:

1. **`RuntimeSurface` defaults to `Local` on an unrecognized/absent env value** rather than failing closed. Reasoning above: the failure modes are asymmetric, and the local default's worst case is a wasted cursor download.
2. **PATH-vs-managed precedence when both are present** (converge the managed copy). Stated in the fn's comment and pinned by a test, because either answer is defensible and the reader should not have to guess which was chosen.
3. **The rewritten poke test asserts admission, not completion.** Asserting completion would have made a unit test perform real network installs. Flagged as a weakening of *that* test's reach in round 1 — **round 2 closed it** with `surface_threading_tests.rs`, which reaches completion offline via the `None`-catalog pin fence, so the threading is no longer only covered at the seam.

Ruling **R7 honored**: no bandwidth knob. The pass installs the supported set unconditionally; the product follow-up gap the ruling notes is not re-opened here.

`probe-engine-design.md` §3.2 adds a per-agent poke at reconcile completion, and §3.1 explicitly **withdraws** any ordering claim between the startup pass and the probe (`reconcile_installed_when_idle().await` returns at admission, not completion). A6 renames that fn to `reconcile_when_idle` and widens its scope; neither touches the withdrawal, and §3.2's completion poke — which the design calls "the sole guarantor of probe-after-install ordering" — becomes *more* useful now that the startup pass actually installs things.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/app/mod.rs
+++ b/anyharness/crates/anyharness-lib/src/app/mod.rs
@@ -207,6 +207,10 @@ impl AppState {
             agent_reconcile_service.clone(),
             agent_seed_store.clone(),
             AgentCatalogService::new(catalog_sync_service.clone()),
+            // Read once, here: the auto-install pass needs the surface for the
+            // cursor-in-cloud carve-out, and reading it at the decision point
+            // would put a process-global read inside the reconcile loop.
+            crate::domains::agents::runtime::RuntimeSurface::from_env(),
         ));
         // Gateway model resolver (spec §2/§3): catalog gatewayPolicy + the
         // sqlite probe store -> the render plane's GatewayModelPlan.
--- a/anyharness/crates/anyharness-lib/src/domains/agents/installer/auto_install.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/installer/auto_install.rs
@@ -1,0 +1,113 @@
+//! Should this reconcile pass install this agent? The pure decision.
+//!
+//! agent-distribution.md, "Installation":
+//!
+//! > Installation is automatic. Every harness supported on a surface converges
+//! > with no user action: absent means install, drifted means reinstall, and both
+//! > are the same mechanism… A user authenticates harnesses; they never install
+//! > them. Two carve-outs:
+//! >
+//! > - An agent the user already provides on PATH is left alone: it is usable
+//! >   through readiness as-is, and a managed install would shadow their copy.
+//! > - Cursor never installs in cloud. It is login-only with no headless
+//! >   credential path, so a cloud install could never reach `Ready`.
+//!
+//! The carve-outs used to be implicit in `installed_only`: because that scope
+//! skipped every agent that was not already managed-installed, it skipped
+//! PATH-provided agents *as a side effect* of skipping absent ones. Dropping the
+//! flag to get auto-install would therefore have silently dropped the PATH
+//! protection too — a managed install would land on top of a user's own binary.
+//!
+//! So the carve-outs move here, as one decision function with a name, per the
+//! plan's requirement that auto-install ride "behind an explicit tested
+//! predicate". A pure fn over facts the caller gathers: no IO, no env reads.
+
+use crate::domains::agents::model::AgentKind;
+use crate::domains::agents::runtime::RuntimeSurface;
+
+/// Whether an artifact the runtime can see is the USER's (on PATH) or OURS
+/// (managed, installed under the runtime home).
+///
+/// `"path"` is the source string `readiness` stamps on a PATH-resolved artifact;
+/// `"managed"` is what the installer stamps. Anything else (an override, an
+/// unknown future source) is treated as not-ours-and-not-PATH: we neither claim
+/// it nor protect it, and the drift planner's own idempotent checks govern.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct AgentInstallFacts {
+    /// Any artifact for this agent resolves to a binary on the user's PATH.
+    pub has_path_artifact: bool,
+    /// Any artifact for this agent is managed-installed under the runtime home.
+    pub has_managed_artifact: bool,
+}
+
+/// Why a pass is not installing an agent — carried so the reconcile result's
+/// `Skipped` message names the actual reason instead of one generic string.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AutoInstallSkip {
+    /// The user provides this agent on PATH. A managed install would shadow
+    /// their copy, and readiness already reports it usable.
+    UserProvidedOnPath,
+    /// Cursor in cloud: login-only, no headless credential path, so an install
+    /// could never reach `Ready`.
+    CursorUnsupportedInCloud,
+    /// An installed-only pass (the settings pane's "update installed agents"
+    /// action) deliberately touching only what we already manage.
+    NotManagedInInstalledOnlyPass,
+}
+
+impl AutoInstallSkip {
+    /// The message surfaced on the reconcile result. Stable strings: the Tier-3
+    /// agent-lifecycle scenario asserts on the PATH one.
+    pub fn message(self) -> &'static str {
+        match self {
+            Self::UserProvidedOnPath => {
+                "provided by the user on PATH; a managed install would shadow it"
+            }
+            Self::CursorUnsupportedInCloud => {
+                "cursor is login-only and cannot reach Ready in cloud; not installed"
+            }
+            Self::NotManagedInInstalledOnlyPass => {
+                "not managed-installed; this pass only updates managed installs"
+            }
+        }
+    }
+}
+
+/// The decision. `Ok(())` means "reconcile this agent" (the drift planner then
+/// decides whether any actual work is needed); `Err(skip)` means leave it alone.
+///
+/// Ordering is deliberate. The PATH check comes FIRST, before the surface and
+/// scope checks, because it is the one carve-out that protects something the user
+/// owns: no combination of scope or surface may ever result in a managed install
+/// landing on top of their binary.
+pub fn auto_install_decision(
+    kind: &AgentKind,
+    surface: RuntimeSurface,
+    installed_only: bool,
+    facts: AgentInstallFacts,
+) -> Result<(), AutoInstallSkip> {
+    // 1. The user's own binary is never touched — highest precedence.
+    //
+    // `has_managed_artifact` wins over `has_path_artifact` when both are true:
+    // that is a machine where we already installed a managed copy AND the user
+    // has one on PATH, so the managed copy is ours to keep converging. Skipping
+    // it would strand it at an old pin forever.
+    if facts.has_path_artifact && !facts.has_managed_artifact {
+        return Err(AutoInstallSkip::UserProvidedOnPath);
+    }
+    // 2. Cursor in cloud could never reach Ready, so installing it is pure cost.
+    if *kind == AgentKind::Cursor && surface == RuntimeSurface::Cloud {
+        return Err(AutoInstallSkip::CursorUnsupportedInCloud);
+    }
+    // 3. An explicitly installed-only pass updates what we manage and nothing
+    //    else. This is no longer the startup pass — it is the scoped action a
+    //    user triggers from settings.
+    if installed_only && !facts.has_managed_artifact {
+        return Err(AutoInstallSkip::NotManagedInInstalledOnlyPass);
+    }
+    Ok(())
+}
+
+#[cfg(test)]
+#[path = "auto_install_tests.rs"]
+mod tests;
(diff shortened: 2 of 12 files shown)


pablonyx:
## QA fork — BLOCKER confirmed LIVE: the PATH carve-out permanently disables codex, and reports success

Booted the merged integration branch (`origin/main` + **a6 (this PR)** + b3 + c1 + d2) as a real dev profile — runtime `:8461`, backend `:8004` (`{"status":"ok","version":"0.3.48"}`). **No setup was needed to reproduce this: codex was already broken by the plain startup auto-install pass.** This is the default experience for anyone with `codex` on their PATH.

### The state

`GET /v1/agents/codex`:

```json
"installState": "install_required",  "readiness": "install_required",
"native":       {"role":"native_cli",   "installed":true,  "source":"path",
                 "path":"/Users/pablohansen/.local/bin/codex"},
"agentProcess": {"role":"agent_process","installed":false,
                 "message":"Not installed. Use the install endpoint to set up."}
```

The PATH artifact is the **native CLI**. The missing artifact is the **ACP adapter**. The carve-out's own stated rationale — a managed install would shadow the user's binary — is true of `native_cli` and false of `agent_process`: installing the adapter shadows nothing. But the fold at `installer/reconcile/execution.rs:313-332` computes `has_path_artifact` as *any* artifact being `path`, so one PATH native CLI suppresses the adapter install too.

`GET /v1/agents/launch-options` (the composer menu): claude 11 models, cursor 33, grok 7, opencode 68, **codex absent** — confirming via `launch_options.rs:124-125` that this is genuine unlaunchability, not just a wrong badge. The user cannot start a codex session at all.

### The loop, and the false success

The user does the only thing the UI offers — `POST /v1/agents/reconcile {"reinstall":true,"agentKinds":["codex"]}` → **HTTP 202**, `totalComponents: 2`. Fifteen seconds later:

```json
"status": "completed",                          // not "failed"
"finishedAt": "...:56.589"                      // 52ms after start — nothing was attempted
"progress": {"completedComponents": 2, "totalComponents": 2,
             "components":[{"role":"native_cli","phase":"skipped"},
                           {"role":"agent_process","phase":"skipped"}]}
"results": [{"kind":"codex","outcome":"skipped",
             "message":"provided by the user on PATH; a managed install would shadow it",
             "installedArtifacts":[]}]
```

`installedArtifacts: []` is the tell — a "completed" install that installed nothing, with `completedComponents == totalComponents` so any progress UI reads 100%. The message is attached to the *agent*, so the adapter inherits a reason only true of the native CLI. `GET /v1/agents/codex` afterwards is byte-identical to before. The loop is closed and infinite.

Why every surface calls it success: `agents_contract.rs:299-318` `to_install_state` returns `Failed` only when the latest outcome is `AgentReconcileOutcome::Failed` — **`Skipped` falls through to `InstallRequired`**, so there is no representable state for "we deliberately refused; stop asking." And `HarnessUpdateToastPresenter.tsx:103-115` keys off job status `completed` ⇒ fires *"Agent tools updated — the installed harnesses and ACP adapters are ready."* So: button re-enables, toast says ready, badge says `install_required`, composer has no codex.

### Spec contradictions

`specs/codebase/platforms/product/agent-distribution.md`:
- `:184-186` — "resolved per agent **and per artifact role** … the two roles drift independently." The carve-out is evaluated per *agent*, so they don't.
- `:347` — "`InstallRequired` … transient — the next reconcile pass auto-installs it." Proven permanent: an explicit reinstall finished in 52ms having done nothing.
- `:127-159` — the carve-out's intent is to avoid *shadowing* a user binary; only `native_cli` can shadow.

### Suggested fix (one shape covers this and the cloud-cursor case in #1505)

Evaluate the carve-out **per artifact role**, and make a carve-out skip a first-class terminal outcome the contract surfaces (e.g. `install_state: "blocked"` + reason) so the client stops offering an install that cannot work and never fires a success toast for it. The identical `Skipped`-reported-as-success shape produces the cloud-cursor install loop in #1505 (`AutoInstallSkip::CursorUnsupportedInCloud`) — worth fixing together.

Severity: blocker for anyone with codex on PATH — silent permanent loss of the harness, with an affirmative "ready" claim. Full transcripts in `codex/ledger-qa.md` (F-028). Read-only QA; nothing pushed to any `agents/*` branch.

pablonyx:

---

PR #1519, opened 2026-07-26
feat(agents): model-snapshot probe engine + polled status route

pablonyx:
Stacked on `agents/a6-auto-install`. Retarget to `main` after that merges.

The engine lands **dormant-but-tested behind its seam, plus the status route**. The pokes that drive it are A8's — that change has its own trigger-ordering argument, and mixing the two would make one review answer two questions. On this revision the engine probes only when a client explicitly asks it to.

## Spec sections (quoted, and what changed)

`specs/codebase/platforms/product/model-catalog.md`, **"Staleness"** said:

> An entry is stale exactly when its recorded identity no longer matches the machine — never merely because time passed:
> 1. **Harness moved**: `attestation.version` differs from the install manifest's `agent_process` version for this harness. Every entry is version-bound…
>
> Nothing else invalidates.

Both halves are unimplementable as written, and implementing them literally produces a permanent probe storm. Evidence from this machine:

| harness | manifest `agent_process` | ACP `attestation.version` |
| --- | --- | --- |
| claude | `26f9ee7a0049507bff5476ce390695515ce92840` (`pinned_git`) | `0.59.0-proliferate.1` |
| cursor | — | `null` |
| grok | — | `null` |
| codex | `0.18.3-proliferate.1` | `0.18.3-proliferate.1` |
| opencode | `1.18.3` | `1.18.3` |

Three of five harnesses would be permanently `HarnessMoved` — a real harness respawned on every startup, every session launch and every auth apply, forever. Backoff cannot damp it, because those probes **succeed**.

Amended per ruling **R11** (first commit, `bfebf2e16`):
- identity is **manifest-vs-manifest** (sha256 preferred, version fallback) — both sides from one producer, so they share a namespace by construction;
- **Indeterminate is not stale** (no manifest, no version for a `source:"path"` dev install, or a pre-field entry);
- the ACP attestation stays as diagnostics, never as a staleness input;
- **TTL joins the law**: 24h + a deterministic per-entry offset over a 6h span, because a gateway-side model add and a keychain rotation move neither identity nor fingerprint — without it those entries pin for the life of the credential;
- a **60s completed-attempt floor**, stated as the structural anti-storm defense.

The amendment also states what the polled status route answers and what it never exposes (`authFingerprint`), the probe-owned scratch root and its GC isolation, the single-writer engine lock with read-only degradation, and cursor's manual-only exclusion (**R13**).

## Design-doc sections implemented

`codex/probe-engine-design.md` **§1 (the seam)** and **§2 (engine semantics)** in full, plus the **errata appendix**.

### §1.2 — the two-phase split

`probe_auth_material` is read-only and runs on **every** gate evaluation: one `state.json` read, one `resolve_profile`, a pure scoping pass, and a `stat` per discovery file. `materialize_for_probe` runs only after the gate decided to probe.

The split is the whole point. The gate needs the fingerprint, and the fingerprint is a function of credential material — so computing it by materializing would put a virtual key on disk **every time the runtime asked "is this entry still fresh?"**, including the 17 answers of "yes" that a startup pass on a converged machine produces. **T-30** snapshots the full recursive listing + mtimes of the runtime home, evaluates all 17 catalog contexts, and asserts byte-identity: no `agent-auth-probe/` at all, no `codex-home-*`, no `opencode.json`, no `*.tmp-*`.

### §1.3 — the one substitution

`render_profile` takes `runtime_home` purely as a materialization root; every path it emits is a deterministic join through `materialize::{revision_dir_path, claude_config_dir_path, codex_native_home_path}`. Handing it the scratch relocates every env var and every `FileSpec` together, **with no branch in `render.rs`** — **T-05** asserts the rendered config bytes are identical whether the root is a scratch or a launch root.

Two consequences fall out for free, and one hazard is removed:
- the launch GC enumerates exactly `<runtime_home>/agent-auth`, and `agent-auth-probe/` is a **sibling**, so it is never a GC candidate;
- the probe's own GC is a provable no-op inside a fresh scratch ("greatest revision strictly below current" finds nothing);
- `claude-config/` is deliberately **not** revision-keyed, so every running claude session shares it. A probe materializing into the live root would write into an in-flight session's config dir. **T-07** asserts the live dir's mtime never moves and a seeded `settings.json` is byte-unchanged.

### §2.3 — the runner owns its child

Timeout and scratch guard live **inside** the probe thread. The obvious arrangement is wrong twice over: `kill_on_drop` fires when the `tokio::process::Child` is dropped and the Child lives in *that thread's* frames, so a caller-side cancel drops only a `oneshot` receiver and **leaks the harness**; and a caller-side guard drop deletes `claude-config/` or `codex-home-<rev>/config.toml` out from under a still-running child.

**T-33/T-34 spawn a real child** (a heartbeat script via the documented `ANYHARNESS_<KIND>_AGENT_PROGRAM` override) and assert all three properties: the pid dies within a bounded wait, the scratch is gone, and — the ordering — the **child itself** reports it never failed to write into its scratch while alive.

### §2.4a — four independent brakes

Each stops a different runaway, which is why all four are present:

1. the staleness gate — a fresh entry is never re-probed;
2. the **60s completed-attempt floor** — the structural one. **T-37** injects a permanently-stale gate (the manifest identity is moved every round) and asserts 50 pokes over the window produce **≤2** probes, not 50. It is deliberately independent of the identity fix being correct;
3. exponential backoff, 1m→2m→4m capped at 6h (**T-24**, asserted through `nextAttemptAt` rather than by sleeping);
4. per-harness serialization + a machine-wide semaphore of 1 (**T-21/T-22**, asserted by observed peak concurrency across four harnesses).

### §1.4 — one engine per runtime home

`.probe-engine.lock` via non-blocking `try_lock_exclusive`. A second runtime over the same home — a dev sidecar beside the desktop is the everyday local case — comes up **read-only**: it serves the document, reports `probeEngine:"readonly"`, and neither probes nor sweeps. Probing is convergence, not correctness; but two engines over one home is the unsound configuration, where each sweeps the other's in-flight scratch. **T-32** asserts 0 runner invocations, 0 sweeps, working reads, a `409 PROBE_ENGINE_NOT_OWNER` on forced refresh, and that dropping the owner lets a third acquire ownership.

The sweep is conservative: a root goes only when its embedded pid is **neither ours nor live** AND its timestamp is older than `3 × per_probe_timeout`. **T-12** builds five roots — our pid, a **real** live foreign pid, a dead pid with a fresh stamp, a dead pid old, and an unparseable name — and asserts only the last two go.

### §2.1 — the three `ProbeOptions` fields

- `auth_env_remove`: as of A5, `sanitize_claude_ambient` runs on **every** non-native claude route, so an `api_key` context has a non-empty removal list too. A probe that dropped removals would observe **Bedrock's** menu on a Bedrock-exporting developer machine and record it as gateway (or `anthropic-api`) truth. **T-15** asserts the full set including `CLAUDE_CODE_USE_FOUNDRY`, for both the gateway and the `api_key` route, and asserts it through the real spawn-env merge.
- `workspace_root`: so one guard cleans everything, including on a cancelled probe whose own teardown never runs.
- `switch_models`: **suppresses only the `config_options` capture**, never the `models.push` the whole snapshot exists for. The naive reading of "gate the loop" would empty `models` for every runtime probe and defeat the feature.

## T-01..T-38 status

| id | status | note |
| --- | --- | --- |
| T-01 | pass | opencode's six contexts scope independently; guards the union-fingerprint bug |
| T-02 | pass | claude `bedrock` (envFlag) + `anthropic-oauth` (discovery) -> Native |
| T-03 | pass | cursor is login-only; a forced gateway render still yields `UnsupportedRoute` |
| T-04 | pass | baseline scrubs every registry-declared var; process env provably unmutated |
| T-05 | pass | merged into one table-driven test over claude/codex/opencode/grok |
| T-06 | pass | **corrected assertion**: rev-8 launch over `{5,6,7}` deletes 5 **and** 6 |
| T-07 | pass | the claude shared-dir hazard |
| T-08 | pass | 0700 scratch, 0600 secrets, no tmp residue (`cfg(unix)`) |
| T-09 | pass | guard removes the root on success, on `Err`, and on unwind |
| T-10 | pass | failure updates `lastAttempt` and **nothing else** |
| T-11 | pass | timeout recording + cleanup, fake runner (deterministic) |
| T-12 | variant | as designed; the unparseable-name leg needs a settable dir mtime — `touch -t` where available, else the test asserts the other direction rather than silently passing |
| T-13 | pass | origin guard, with the matching-origin control |
| T-14 | pass | no plaintext in `Debug` or in the fingerprint inputs |
| T-15 | pass | env removal through the real spawn-env merge, incl. `_FOUNDRY` |
| T-16 | pass | every reason + precedence (identity -> auth -> time) |
| T-17 | pass | TTL boundary +/-1s; a backwards clock does **not** expire |
| T-18 | pass | the per-context property: rotating one key leaves the other harness fresh **even though the global revision moved** |
| T-19 | pass | fingerprint stability + sensitivity; source order is irrelevant |
| T-20 | pass | 8 concurrent pokes -> exactly 1 spawn |
| T-21 | pass | distinct keys don't coalesce, and are serialized |
| T-22 | pass | global cap holds across harnesses |
| T-23 | pass | idempotence, with `min_reprobe_interval: 0` so the **gate** is what refuses |
| T-24 | pass | 1m/2m schedule, in-window no-op, success resets |
| T-25 | pass | (a) bypasses backoff, (b) unchanged credential -> 1 spawn, (c) rotation -> 2 spawns with the new fingerprint |
| T-26 | pass | not-installed filtered before spawning, no `lastAttempt` written |
| T-27 | pass | camelCase round-trip; corrupt/schema-mismatched/tmp-residue all read absent |
| T-28 | pass | status projection; `authFingerprint` absent from the serialized body |
| T-29 | partial | poke **fan-out** implemented and asserted; the reconcile-completion leg needs A8's wiring |
| T-30 | pass | **the assertion the split exists for** |
| T-31 | pass | phase B uses phase A's revision and profile, even after `state.json` moves |
| T-32 | pass | the single-runtime lock |
| T-33 | pass | **real child**: cancellation kills it, cleans up, in the right order |
| T-34 | pass | **real child**: same, via timeout |
| T-35 | pass | the C8 storm regression, all three cases + the claude positive control |
| T-36 | variant | see delta below |
| T-37 | pass | the structural anti-storm assertion |
| T-38 | partial | live-plan and memo-invalidation legs asserted; the warned seed-fallback leg needs A8's memoizing producer |

**32 tests**, all green. Deltas:

- **T-36** — the design claims the 17 contexts land "~21 minutes apart". Measured on the shipped catalog that is the **average** (1143s); the **closest pair is 232s**, and a hash-mod cannot provide a per-pair minimum. The test asserts what is actually guaranteed: >=5h span, no two TTLs equal, average gap >=15min, min gap >=120s. 232s is still under the 240s probe timeout, so at `semaphore = 1` the worst case is one probe briefly waiting on the gate — not a herd. Asserting a 5-minute floor would have pinned a property the design does not provide.
- **T-11 / T-34** — the design has one timeout test; this is two. T-11 uses the fake runner for the recording/cleanup assertions (deterministic, no process); T-34 uses a real child for the kill-order assertion, because that property is a fact about which frame owns the `Child` and a fake cannot exercise it.
- **T-29 / T-38** — the parked legs test A8's wiring, not this PR's engine.

## Design decisions I had to make

The design doc is silent on these; each is resolved with its reason in the code.

1. **Gateway key digest name.** The recipes set the same key into different vars per harness (`ANTHROPIC_AUTH_TOKEN` / `PROLIFERATE_GATEWAY_KEY` / `XAI_API_KEY`). Digesting under the var name would move every gateway fingerprint if a recipe renamed its var — a mass re-probe for a change that touched no credential. The digest is keyed on a pseudo-name, `gateway.key`.
2. **Mixed env+discovery contexts** (see Contradictions).
3. **Discovery-file path map.** Kind -> path, mirroring `credential-discovery`'s own layout. Keychain kinds and the AWS chain map to **nothing** rather than a guess: reading them prompts the OS. Absent files stat as `(path, 0, 0)` rather than being dropped, so a login **appearing** moves the fingerprint.
4. **FNV-1a, not `DefaultHasher`.** std's hasher is explicitly not stable across releases; a toolchain bump would silently re-schedule every machine's probes at once.
5. **`probedAt` comes from the engine's clock**, not the snapshot's string, so the entry, the gate and the status age share one clock and an age can never come out negative because two clocks disagreed.
6. **A failure with no existing entry writes nothing.** A models-less entry would make the picker believe the harness advertises nothing — worse than absence, which falls back to the shipped catalog.
7. **Negative age is not an expiry; an unparseable `probedAt` is.** The first would re-probe everything after any clock correction; the second is a real defect one re-probe fixes permanently.

## Contradictions encountered

**C-A7-1 — the design's scoping table is incomplete for the shipped catalog.** It has pure-env and discovery-only rows, but the catalog has **mixed** contexts: opencode `anthropic-api`/`openai-api`, cursor `cursor-login`, codex `openai-api`, grok `xai-api` (env **or** discovery). Scoping them as pure-env made an un-enrolled machine fail closed on its own login — which would turn every login-backed context on every enrolled machine into a permanent failed attempt. **T-03 caught it.**

Resolved as a judgment call rather than escalated, because the design's own stated rule decides it — *"probe env == launch env for that context"*, and a launch for an un-enrolled login-backed context renders nothing. A context that **also** declares a discovery signal falls back to Native when nothing is enrolled; a **pure** env context still fails closed. T-01 asserts both on one document, so the discovery signal is provably what makes the difference.

**C-A7-2 — `invalidate_gateway_plan` is a no-op on today's producer, correctly.** `GatewayModelResolver` reads `gateway_model_probe` sqlite rows, so there is no in-process memo to invalidate. The seam exists now so the forced-refresh path is already correct when A8's memoizing producer lands. Documented on the trait method rather than left to be rediscovered.

## Gaps closed (deleted in this PR)

- **status-route bullet** — *"Probe status is not pollable: no runtime endpoint exposes `probedAt`/`lastAttempt` per (harness, context) the way `GET /v1/agents/reconcile` exposes install progress."* -> `GET /v1/agents/{kind}/model-snapshot`, shaped after that exact route. This is the reusable **pull-status** pattern: polled, never pushed.
- **probe-pipeline bullet** — narrowed, not deleted. The `model_snapshot/` module and its document now exist and probe correctly; what remains is the poke wiring, which is A8's, and the bullet now says exactly that.

## Adversarial self-review

- **No secrets in the fast path.** The only reads of `.value`/`.key` in `probe_auth_material` feed straight into `sha256_hex`. `ProbeAuthMaterial`'s `Debug` is hand-written to redact the privately-held scoped profile — a derived one would print keys. **T-14** asserts a known secret appears in neither the debug form nor the fingerprint inputs.
- **Scratch hygiene, corrected.** An abandoned scratch is not merely wasted bytes: a native-codex probe materializes a copy of the user's own `~/.codex/auth.json` into it, because relocating `CODEX_HOME` relocates where codex resolves its login. That is why the orphan sweep now runs at engine construction (M3 below) rather than waiting for A8's startup pass.
- **Fail-closed everywhere**, each with a test: a corrupt `state.json` declines to probe and surfaces a typed error only to a caller who asked; **no** `state.json` still probes (that is a fresh desktop observing its own logins, not a degradation); a corrupt document reads absent and is rewritten whole; an unsatisfiable selection is declined silently but reads **stale** on the status surface rather than silently fresh; the lock not being held refuses with `409`; a probe timeout kills the child and leaves no scratch.
- **The floor and jitter are deterministic under test** — the floor via a counting assertion over a simulated window (T-37), the jitter as a pure function of `(harness, context)` with an injected `now` (T-36).
- **The status body cannot leak the fingerprint** — the projection type has no field for it, and both the unit test and the **route** test assert against the serialized bytes, not the type, because the leak this guards against would be a serde-attribute change.

## Not done here, deliberately

Poke wiring, `GatewayModelPlanner`, `validate_launch`'s snapshot universe (A8); the old-chain deletion (A9).

## Tests

```
cargo test -p anyharness-lib --lib model_snapshot     32/32
cargo test -p anyharness-lib --lib route_auth        101/101
cargo test --workspace                                 all green (1398 anyharness-lib, 0 failures)
repo-shape (11 checks + report_frontend_structure)     all green
```

Three repo-shape ratchets caught real omissions and are fixed in-PR: **max-lines** (every new file split by responsibility rather than allowlisted — see `b9464e5cd`), the **session-mutation admission ratchet** (the new POST classified `workspace-scoped`), and **OpenAPI drift** (regenerated, additive only).

Note for reviewers: `cargo fix --tests` is **not** safe in this crate — it strips load-bearing re-exports in four files no branch in this stack touches. My own imports were pruned by hand.

Generated with [Claude Code](https://claude.com/claude-code)

---

# Review cycle 1 — all items resolved (`c940f0cdc`)

Verdict was FIX-NEEDED with the security core verified clean. Five majors and nine minors, each below with what it actually was.

## Majors

**M1 — `authContextId` never arrived.** `RefreshQuery` lacked `#[serde(rename_all = "camelCase")]` while the utoipa parameter *and* the generated SDK both declare `authContextId`. So every SDK client's query deserialized to `None` and fell through to the refresh-all branch. Fixed, and the field is now **required**.

The test discriminates rather than confirms: `?auth_context_id=` must `400`, `?authContextId=` must reach the engine. A test that only asserted the camelCase form works would have passed against the broken code too, because the broken code answered *something* for every input.

**M2 — the refresh-all branch is withdrawn, not parallelized.** It awaited `refresh_now` for every active context in turn: six contexts × a 240s probe timeout, serialized by a semaphore of 1, is a single HTTP request held for ~24 minutes. The design (§2.5, §4) only ever specified a **single-context** forced refresh awaiting its probe — the fan-out shape was invented in A7. A surface that wants "refresh everything" issues one request per context and polls `GET` for progress, which is exactly how it already drives the reconcile job this route is shaped after. Choosing that over "202 + enqueue" keeps one semantics for the route (it awaits, it reports) instead of two.

**M3 — the sweep had zero production callers, and that was a credential leak.** Layer 3 of the cleanup story existed as a function and a test. It is now called from `with_parts` immediately after the flock decides ownership, owner-gated.

The PR body's earlier hygiene argument was too weak and is corrected: an abandoned scratch is not wasted bytes. A **native-codex** probe materializes a copy of the user's own `~/.codex/auth.json` into its scratch, because relocating `CODEX_HOME` relocates where codex looks for its login (design errata `:690-698`). A SIGKILL runs no guard, so without a sweep that plaintext credential copy sits under the runtime home indefinitely. The new test asserts the copy is reclaimed — and that this process's own root is not.

**M4 — the "±20% jitter" comment was false; the jitter now exists.** `record_failure` had none, and T-24 pinned its *absence*. Implemented rather than deleted, because the failures that matter arrive in groups: one provider outage, one expired org key, one revoked enrollment fails **every** dependent context inside a single poke, and a flat ladder then retries all of them at the same instant forever — the same self-synchronizing burst the TTL jitter prevents one layer up.

Deterministic FNV seed over `(harness, context, attempt)`, reusing the staleness hash so there is one value to keep stable across toolchains. T-24's windows widen from the exact rung to rung ±20%; the two envelopes do not overlap, so the doubling stays exactly asserted, and `second > first` is asserted directly. The new jitter test checks determinism, the envelope, spread across half the envelope for one harness's six contexts, movement across attempts, and the degenerate 1s case.

**M5 — attribution, generalized past claude.** `baseline_scrub` becomes `attribution_scrub`: a non-gateway context now removes the **other** provider slots' registry-declared env vars.

The concrete failure: opencode declares `openai-api`, `anthropic-api` and `gemini-api` over three separate provider slots, and it composes every provider it can see. Probing `openai-api` on a machine that also exports `ANTHROPIC_API_KEY` therefore recorded **anthropic** models inside the `openai-api` entry — and because entries are keyed by context id, that is what the picker served and what launch validation trusted. R14's attribution-true property held only for claude, and only because claude's own `sanitize_claude_ambient` happened to do it.

Gateway is exempt (it observes a proxy, not a provider, and its recipes own their sanitization), and a var this route itself **set** is never removed. Both tests assert a registry *relationship* — "whatever the other slots declare is gone, whatever this slot declares survives" — rather than a hardcoded list, so a registry addition cannot silently escape the scrub. The removals are then asserted through the real spawn-env merge.

## Minors

| id | resolution |
| --- | --- |
| m1 | The non-unix `process_is_live` returned `true` unconditionally, so a parseable root could **never** be swept at any age — caution that was actually a leak. Now `Option<bool>`; `None` means age-only. Our own pid is excluded separately, so the exclusion did not depend on the liveness answer. |
| m2 | `RefreshError::Material` answered `502` — an *upstream* code — for a **local** config fault, and leaked `state.json`'s absolute path in `Detail`. Now `409` with the typed route-auth code as detail; the path goes to the log, where it belongs. `degraded_tests` asserts the variant, not just the code string. |
| m3 | `lastError` served unbounded harness-controlled text verbatim. Now redacted and truncated to 512 chars. Redaction is **by digest** — tokens are hashed the way phase A hashed the credential value — so this path holds no plaintext credential either. Writing the test found a real hole in the first version: `key=sk-…` embeds the secret *inside* the token, so trimming punctuation never reached it; the redactor now also peels a `name=`/`name:` prefix and keeps the label (`key=[redacted]`). |
| m4 | `model-snapshot.json` is written `0600`, with the chmod on the **tmp file before the rename** so it is never briefly world-readable at its final path. Every entry carries an `authFingerprint`: preimage-safe, so not a key — but a stable per-credential identifier that tells a co-tenant *which* credential a user is on and *when it rotated*. `state.json` next door is 0600 for weaker reasons. |
| m5 | Founder note, no code change — see below. |
| m6 | T-11's scratch assertion was **vacuous**: `FakeRunner` never materialized, so it would have passed against a runner that leaked every root. The fake now performs the real phase B and drops the guard after the outcome, mirroring the production ordering — which makes every "no scratch outlived the attempt" assertion in the suite able to fail. |
| m7 | `mark_running` fired *after* the semaphore, so a queued probe reported `idle`. At `max_concurrent_probes = 1` that wait is the **common** case, and a polling UI told "nothing is happening" hides its own spinner and the user presses Refresh again. Added `LiveState::Queued`, set at slot admission before both waits; in-flight now outranks backoff in the projection. New test: two contexts of one harness, one `running`, one `queued`, both settling to `idle`. |
| m8 | Spec drift fixed. The narrowed bullet now says "none of the **automatic** pokes is wired" — the refresh route does call the engine, so the old wording was false the moment it was written. The Code map now matches the shipped module and states why there is no `projection.rs` (this module owns the observation, not the picker merge). |
| m9 | Both routes declare response body types (`ModelSnapshotStatus`, `ProblemDetails`) so C3's polling UI gets a typed shape; schemas registered in `openapi.rs` and regenerated into the SDK. |

Nits: `record_completion`'s dead `Option<()>` parameter is gone, replaced by `record_success` beside `record_failure` — no shared "did it fail?" flag to get wrong. The `Cargo.lock` `0.3.46 → 0.3.48` line is **kept**: `VERSION` is already `0.3.48` and the lockfile had lagged, so any cargo invocation regenerates it and reverting would simply be undone.

## m5 — founder note (no code change)

Two things worth knowing, neither a defect in this PR:

1. **The GET status route computes active contexts per poll**, and `collect_launch_env_facts` reaches credential discovery, which on macOS can shell out to the keychain. The machinery is pre-existing and unchanged; this route is simply its first *polled* consumer, so a UI polling every few seconds multiplies a cost nothing previously paid at that rate. If it shows up, the fix is a short-TTL memo on the classification, not a change to the route's contract.
2. **R13's hazard generalizes past cursor.** Cursor is excluded from unattended probing because its credential is keychain-only. But `codex-keychain` and `claude-keychain` are also declared discovery kinds, so a codex or claude probe on a machine whose login lives in the keychain can reach the same prompt path. It has not been observed, and this PR does not change it — flagging it because the exclusion list currently reads as if cursor were structurally unique when it is only the clearest case.

## One note on the plan's wording

The impl plan's A7 line says "sqlite snapshot store (re-key from `gateway_model_probe`)", and this PR ships a **JSON document** instead. That is not a deviation: `model-catalog.md:62-88` — the spec, which is the authority — puts the machine snapshot in `agents/<kind>/model-snapshot.json` beside the install manifest, and explicitly names the `gateway_model_probe` sqlite table as the thing it *replaces*. The sqlite re-key in the plan is the **cloud-side** `agent_catalog_snapshot → agent_model_snapshot` migration, which is Track B's (B4). No change needed; noting it so the plan's phrasing does not read as a missing piece.

## Tests after the fixes

```
cargo test -p anyharness-lib --lib model_snapshot     37/37   (was 32; +5 for the fixes)
cargo test -p anyharness-lib --lib route_auth        103/103  (was 101; +2 attribution)
cargo test --workspace                                anyharness-lib 1405/1405, 0 failures
repo-shape (11 checks + report_frontend_structure)     all green
```

The only workspace failures are two `apps/desktop/src-tauri` `app_config` tests. Verified pre-existing and unrelated: with this stack's changes stashed, a *different* test in that same module fails — it is the documented nanosecond temp-path collision flake, in files no branch in this stack touches.

Five new files came out of `check_max_lines` refusing to let the fixes grow the existing ones: `attempt.rs` (the *doing* — slot admission, the two waits, the runner call, the two persisted outcomes — split from the reconciler's *deciding*), `detail.rs`, `backoff.rs`, and the test-side `attribution_tests.rs`.


pablonyx:
Ready for review @ 0e1f0a8432435e3e0e68828d9fbd1fc37d2d547b — 2-cycle adversarial review complete (cycle-2 verdict READY; S1 openapi-enum fixup + spec-drift docs landed after). Full workspace green (anyharness-lib 1406/1406, src-tauri 90/90). Reviewer follow-ups (non-blocking) logged in codex/agents-impl-ledger.md.

pablonyx:
## QA review — A7 probe engine @0e1f0a843 (live-QA fork)

Reviewed on a **merged integration branch** (a-stack head + b-stack head + C1 + D2 + A7), not on this PR alone. Headline first: **the merged tree compiles and the full Rust suite is green — 1406 passed, 0 failed, 1 ignored (72.8s)** — and A7's own 38 `model_snapshot` tests pass in 1.30s. Two findings below, one major and one minor. Everything else I went looking for was right.

### Verified correct (the things most likely to be wrong)

- **M3, the credential-leak fix, is genuinely wired and conservatively gated.** `service.sweep_orphan_scratch()` at `model_snapshot/mod.rs:172` inside `with_parts`, so reclamation exists from the moment an engine does. Owner-only. `sweep_probe_scratch` removes a root only when **both** the pid is dead/foreign **and** age > `240s × 3`. `parse_scratch_name` takes the numeric fields from the *end*, so ids containing `-` parse. `libc::kill(pid, 0)` with only `ESRCH` counted as dead — `EPERM` correctly reads as alive. The `#[cfg(not(unix))]` → age-only fallback, and the comment explaining that the earlier "always live" shape *was* the leak, are both correct.
- **Phase A really is read-only.** `probe_auth_material_for_server` creates no directory; `ProbeScratch::create` is reachable only from phase B on the probe thread. The gate does 17 state reads and zero writes.
- **Teardown ownership holds.** Timeout and scratch guard both inside the probe thread, so `kill_on_drop` fires on the runtime owning the `Child` and the scratch outlives it.
- **Credentials can't reach the document or the wire.** Hand-written `Debug` omitting `scoped_profile`; no `authFingerprint` field exists on the projection type at all; `write_document` chmods the **tmp** file before rename and re-applies after.
- **M1 landed.** `RefreshQuery` is camelCase and `authContextId` is now required, with `router_tests.rs:1417` asserting both spellings.
- **The F-019 origin filter does *not* mis-attribute a probe.** I chased this specifically. Phase A uses the same `load_effective_state` filter, so a mismatched state → `Native`; my worry was a *gateway* context being recorded with *native* models. It can't be: `RuntimeProbeTargets::active_contexts` (`targets.rs:67-79`) classifies through `resolve_launch_auth_profile`, which applies the **same** guard, so the gateway context is never an active target. Your test at `materialization_tests.rs:336-370` is consistent with F-019, not in tension with it. **But this consistency is load-bearing and undocumented** — if F-019's fix turns the mismatch into `SelectionMissing`, re-check this pair.

---

### F-036 — MAJOR — a dropped refresh request pins a context in `Running` until a manual refresh

Proven with a real failing test, not by reading.

`attempt.rs` sets live state as straight-line statements around the await — `Queued` at `:57`, `Running` at `:66`, the await at `:68`, `Idle` at `:79`. There is no drop guard in `model_snapshot` (the only `drop_guard` is the `CancellationToken` in `probe.rs`, which doesn't touch live state). Drop the future between `:66` and `:79` and `:79` never runs.

**Reachable from outside:** `refresh_model_snapshot` awaits `refresh_now` in an axum handler, so a client disconnect — closed tab, navigated away, client-side timeout — drops it. With `per_probe_timeout = 240s` and `semaphore = 1`, that window is up to 4 minutes wide per context. This is "the user closed the settings pane while it was refreshing."

The consequence is the part that matters: `gate_admits` treats `Running` as in-flight and returns false (`mod.rs:360-367`), *before* the staleness rules. Every later automatic poke — startup, install-completed, auth-applied, session-launch — coalesces onto a probe that no longer exists. Nothing expires it; the state is in-memory and only `set_live_state` writes it.

Executed repro (`cargo test -p anyharness-lib --lib qa_probe_defects`, spawn a `refresh_now` against a gated fake runner, abort once in flight, then poke):

```
QA: state after the dropped refresh = Running
QA: probes after the wedge = 1 (was 1)     <- 5 rounds of poke_harness, 0 new probes
QA: state now = Running
QA: refresh_now ok = true
QA: probes after manual refresh = 2
QA: state after manual refresh = Idle
```

and with the assertion in:

```
assertion `left == right` failed: a dropped refresh must not leave the context pinned Running
  left: Running
 right: Idle
```

**Honest severity bound — major, not blocker:** `refresh_now` doesn't consult `gate_admits`, so a user-initiated Refresh still runs and resets to `Idle` (lines 4-6 above). The wedge is recoverable by the one action a user with a stuck spinner would take. What's lost until then: every automatic refresh path for that context, and a status surface reporting `running` forever — which by your own design note is the inverse of the lie you set out to avoid, and the UI hides the Refresh affordance while `state == "running"`, which is what makes the recovery non-obvious.

**Fix shape:** make the live state RAII — a guard owning `Arc<ContextSlot>` that sets `Idle` on `Drop`. That makes return, `?`, panic and future-drop all correct by construction, which is the argument you already made for `ProbeScratch` and the `CancellationToken`. The inconsistency is that the scratch got a guard and the live state didn't. (A wall-clock `running_since` that `gate_admits` disregards past `per_probe_timeout` also works, but it's strictly worse — it fixes the gate and not the reported status.)

**Test gap that let it through:** `a_probe_waiting_for_a_slot_reports_queued_not_idle` asserts states during a probe and after a *normal* completion. Nothing in `engine_tests/` drops or aborts an in-flight attempt — the only `drop` is `lifecycle_tests.rs:403`, which drops the *lock*. T-33/T-34 cover caller-drop for the child and the scratch and both pass; the live state was simply never in scope.

I wrote the test in-tree to prove this, then **removed it and restored `engine_tests/mod.rs`** — QA doesn't write to `agents/*` branches. It's kept locally if you want it as a regression test.

### F-037 — MINOR — `is_native()` has no non-test caller, and its doc claims one

`probe_materialization.rs:208` says it's "exposed because the status surface distinguishes 'we materialized a route' from 'we observed the user's own credentials'." It isn't — every call site is a test (`materialization_tests`, `scoping_tests`, `attribution_tests`, `contract_fixture_tests`, `profile.rs`, `render_tests`, `state.rs`). Nothing in `model_snapshot/` calls it.

And the distinction genuinely isn't on the wire: neither `SnapshotEntry` nor `ContextStatus` carries a route-vs-native field, so a `gateway` context probed with the state file absent and the same context probed with a live virtual key produce entries a client can't tell apart (the fingerprint differs, but that's deliberately never exposed).

Pick one rather than splitting: (a) if the distinction matters, carry it — `resolvedRoute: "gateway" | "native"` on the entry and the projection, which also gives the C3 model UI something real to render; or (b) if it doesn't, fix the comment to say "exposed for tests." No wrong behavior today — just a comment asserting a consumer I couldn't find, in a file whose comments are otherwise unusually reliable.

### Ratchets

The `max_lines_allowlist.txt` entries A7 bumps (openapi 691, router_tests 1497, probe 683, router 640, app/mod 662) are **disjoint** from the b3/c1 collision files. All 11 `scripts/check_*.py` re-run after merging A7: `check_max_lines` is the only red and it's **unchanged** — still the pre-existing b2/b3-vs-c1 collision (`test_agent_gateway_api.py` 734/723, `test_agent_auth_materialization.py` 679/614). A7 introduces nothing.

Full detail, including the merged-branch integration status, in `codex/ledger-qa.md`.

pablonyx:
F-037 fixed at 047ffc351.

Note: this PR was already flagged READY; this is a minor additive commit re-touching it post-ready, not a rework.

**Resolution chosen:** (a) per the ruling ("delete it or wire it") — wiring it (a `resolvedRoute` field on the entry + status projection) is C3 scope creep, so I did not do that. Instead marked `ProbeAuthMaterial::is_native()` `#[cfg(test)]` and corrected the doc comment from the false "the status surface distinguishes..." claim to "exposed for tests only," noting a real consumer would need a real field.

Went with (a) over full deletion because every call site (`materialization_tests.rs`, `scoping_tests.rs`, `attribution_tests.rs`, `contract_fixture_tests.rs`, `profile.rs` tests, `render_tests.rs`, `state.rs` tests) is spread across multiple test modules/files that would each need their own inline route-vs-native check if the shared helper were deleted — `#[cfg(test)]` keeps one definition instead of forking the logic N ways.

**Test evidence** (targeted, no contention — checked `pgrep -x cargo; pgrep -x rustc` was clean before running):
```
cargo test -p anyharness-lib --lib route_auth
test result: ok. 103 passed; 0 failed; 0 ignored; 0 measured; 1304 filtered out; finished in 0.07s
```
All the listed `is_native()` call sites are in that filtered set and pass.

Also ran `cargo build -p anyharness-lib --lib` clean (only pre-existing, unrelated dead-code warnings) and `scripts/check_max_lines.py` (passed) before pushing.

pablonyx:

---

PR #1520, opened 2026-07-26
chore(anyharness): drop dead gateway-models/catalog leftovers after F-039/F-040

pablonyx:
## Summary

**Restructured per F-039's ruling (`codex/ledger-qa.md` §F-039, `codex/agents-impl-ledger.md`
2026-07-26 "QA integration findings intake").** C2 and B4 (#1521) independently implemented
the same `model-catalog.md` route-rename gap item — C2 in place (`agent_models_router`
alongside the existing `agent_gateway/api.py` handlers), B4 by relocation (new
`cloud/agent_models/` package, worker-authenticated ingest, refresh absorbs mirror). Pablo
ruled **B4's implementation survives**; this PR now carries only the sliver of C2's original
diff that B4 does not own.

This PR's base is `agents/b4-snapshot-rekey` (B4's head, including its F-040 fix that cut the
TS clients over to `/v1/cloud/agent-models`). **Deploy pairing: this PR ships in the same
train as B4 (#1521) — it is not independently deployable**, since B4 owns the server-side
route relocation this branch used to duplicate.

## What moved to B4 (dropped here)

Everything C2 originally built for the route rename itself now lives on B4 instead, including
B4's own F-040 fix that closed the gap F-039 flagged (B4 shipped the server-side move but
initially left every TS consumer on the old path — fixed on B4, head `6a8c6eb5d`):

- The duplicate `agent_models_router` in `server/proliferate/server/cloud/agent_gateway/api.py`
  and its mount in `cloud/api.py` — B4's `cloud/agent_models/` package is the one namespace now.
- `cloud/sdk/src/client/agent-gateway.ts`'s `agentModelsPath`/`getAgentModels`/override
  functions, `cloud/sdk-react`'s `useAgentModels`/`useUpsertAgentModelOverride` hooks, and the
  regenerated `cloud/sdk/src/generated/openapi.ts` — B4 owns the real (worker-authenticated
  refresh, no mirror) contract.
- The `use-gateway-catalog-mirror-sync.ts` / `gateway-catalog-mirror.ts` deletion and the
  `ProductLifecycleRoot.tsx` wiring removal — B4 deleted the same dead code in its F-040 fix.
- `test_agent_gateway_catalog_api.py` / `test_agent_gateway_catalog_mirror_api.py` deletion —
  B4 deletes the same two files (its integration tests assert the old routes 404).
- The `agents-playground-cloud-client.ts`/`.test.tsx` route-literal updates — B4's version is
[...]

PR #1521, opened 2026-07-26
feat(server): re-key the cloud snapshot store to auth contexts (B4)

pablonyx:
## Spec section (authority)

`specs/codebase/platforms/product/model-catalog.md` §Storage:

> The server stores cloud-sandbox snapshots in `agent_model_snapshot`, which is
> today's `agent_catalog_snapshot` re-keyed from coarse route to auth context
> […] There is no `surface` column: only the cloud surface ever syncs, so every
> row is the owner's cloud sandbox observation. […] Every row is a machine's
> observation — there is no `source` column and no ownerless seed row, because
> the server never generates snapshots. The seed tier needs no rows at all: the
> layered read falls back to the shipped catalog document the server already
> serves (`GET /v1/catalogs/agents`), so "seed" is a read-time join, not stored
> state with a writer to maintain.
>
> The soft-versioning discipline is kept as-is: **no unique key on the scope**; a
> write deactivates all prior `active` rows for
> (harness_kind, auth_context_id, owner_user_id) and inserts the new row as
> `active`; reads take the latest active row.

§Write paths:

> Today's three write paths collapse to one […] Today's third path — the server
> running gateway discovery itself (enrollment lookup, virtual-key decrypt,
> `GET /v1/models`) — is deleted with the uniform probe mechanic: the server
> cannot spawn a harness, so it cannot produce an observation in the entry shape.

## Gap bullets closed

Both edited in this PR, in `specs/codebase/platforms/product/model-catalog.md`
§Current gaps:

1. **Cloud-snapshot re-key** — the bullet's schema half is struck (route →
   `auth_context_id`, `models_json` → `snapshot_json`, table rename, ownerless
   seed rows + `source` column gone, soft-versioning kept). What remains open is
   only the *consumer coverage* half ("read only by the settings All Models tab;
   composer/web/mobile/automations/workflows read the shipped catalog"), which is
   C2/C3 work, so the bullet is narrowed rather than deleted.
2. **Three write paths** — bullet **deleted**. The two snapshot writes collapse
   into one ingest route and the server-side gateway discovery is gone.

No other bullet was touched.

## What changed

New revision `b7c1e4d9f082` (down_revision `28adf1a9e376`, the B2 head on this
branch; single alembic head before and after).

- `route` → `auth_context_id`, `String(16)` → `String(64)`, carrying the
  catalog's own ids. This is the substantive change: `api_key` was never an auth
  context — for claude it covers **both** `anthropic-api` and `bedrock` — so the
  old key could not name what a row observed.
- `models_json` → `snapshot_json`, holding one machine-document entry verbatim
  (models, modes, attestation, `authFingerprint`, warnings, `lastAttempt`), so
  the cloud tier serves exactly what the machine tier observed and C3 has the
  fields staleness UI needs.
- `surface` and `source` dropped; `owner_user_id` becomes `NOT NULL`.
- New server package `server/proliferate/server/cloud/agent_models/`
  (`api`/`models`/`snapshots`/`overrides`), per the spec's code map.
  `agent_gateway/catalog.py` deletes whole, `_probe_gateway_models` with it.

### Data migration: existing rows are DROPPED

Called out explicitly, as the brief asks. A faithful mapping is **impossible**,
not merely lossy, for two independent reasons (both argued in the migration's
docstring):

1. `route` is not an auth context. `native` and `api_key` each fan out to several
   catalog contexts per harness, so any single target id would be a guess about
   which credential actually ran the probe — and the re-key's whole point is that
   the context *is* the identity.
2. The payload shape changed. An entry needs `probedAt`, `mechanism`,
   `attestation` and `authFingerprint`; a models-only array carries no
   attestation or fingerprint, so a migrated row would be permanently
   stale-indeterminate and could never be invalidated correctly.

Only `gateway` has a 1:1 route→context name, and those rows are exactly what the
deleted server-side prober produced — i.e. not machine observations at all, which
is the invariant the new table asserts. Cost of dropping is one probe: the
reconciler re-probes any (harness, context) with no fresh entry, and the
read-time shipped fallback serves models in the meantime, so no surface renders
an empty picker.

### One deviation from the brief, resolved in the spec's favour

The brief said "New uniqueness scope: (harness_kind, auth_context_id,
owner_user_id)". A partial unique index over `status = 'active'` was built first
and then **withdrawn**, because the spec (authoritative) says the opposite — "no
unique key on the scope" — and is right on the merits: uploads are
fire-and-forget from the Worker's `heartbeat_and_converge` tick, so two ticks
racing one context would turn a benign duplicate into a 500 the Worker cannot act
on. The read takes the latest active row by `probed_at`, and the next write
collapses all active rows (hence `update(...)` over the scope rather than a
targeted single-row write).

### Ingest owner derivation

The owner is read off the uploading Worker's `cloud_sandbox` row, never off the
body — which has no owner field at all — and not even off the worker row's own
`owner_user_id`. A desktop worker is refused **403**, not 400: "the desktop does
not sync" is a law of the design (no machineless surface consumes a local
observation), not a malformed request. `WorkerAuthContext` gained
`cloud_sandbox_id` for this.

### Read-time shipped fallback

Scoped per auth context through the catalog's existing `availability.anyOf`
algebra, so a `gateway` read does not leak `anthropic-api`-only models. The
response carries `origin: "snapshot" | "catalog"` so a client can mark seed data
unverified. A malformed stored row now falls back to the seed rather than serving
`[]` — strictly better than pre-B4 behaviour.

## Tests

**Tier 1 (unit, pure)** — `server/tests/unit/test_agent_models_layering.py`, 33
cases: entry parse (string normalization, unknown-field preservation, absent
lists, six invalid shapes), override apply (remove→update→add order, in-place
add, identity), and both scope-id bounds incl. the catalog-declared vocabulary.

**Tier 2 (integration, real postgres + routing)**

- `test_agent_models_api.py` (21): shipped fallback with no row; fallback scoped
  per context; unknown context serves nothing rather than everything; snapshot
  wins over catalog; malformed row falls back; owner scoping; auth required;
  `authContextId` required; over-long ids are 4xx not 500; scope isolation across
  context and harness; override layering over **both** snapshot and seed;
  override upsert/delete/owner-scoping; and the deleted prober asserted
  structurally (module gone, no `_probe_gateway_models` anywhere under
  `proliferate/`, ingest module cannot import `litellm`, old routes 404, mirror
  wire models absent).
- `test_agent_models_ingest_api.py` (12): Worker upload lands under the sandbox
  owner; entry stored verbatim; spoofed `ownerUserId`/`owner_user_id` in the body
  ignored and the victim untouched; desktop worker 403; anonymous 401; a product
  user's own bearer is **not** a worker bearer; invalid entry/timestamp/scope-id
  400s; one active row + retained history over three uploads; a redundant
  re-upload absorbed idempotently; two contexts upload independently.
- `test_agent_model_snapshot_rekey_migration.py` (1): alembic
  **up → down → up** against postgres with pre-B4 rows seeded (ownerless seed,
  owned gateway, owned native). Asserts both shapes, that rows are dropped rather
  than mis-mapped, that no `ux_*` index exists, and that duplicate active rows
  insert cleanly.

**Adjacent suites re-run green** (55): `test_schema_migrations.py`,
`test_agent_gateway_store.py`, `test_agent_gateway_usage_credits.py`,
`test_agent_gateway_api.py`.

**Deletions/repo shape**: all 11 `scripts/check_*.py` green (run before the
language suites), including `check_migration_heads.py` (single head),
`check_max_lines.py` and `check_server_boundaries.py`. `ruff check` and
`ruff format --check` clean. `pnpm --filter @proliferate/cloud-sdk typecheck`
exit 0.

Dropped `test_exhausted_budget_blocks_gateway_catalog_refresh_with_402`: its
subject was the deleted server-side prober. The other two budget walls (LiteLLM
key disable, state-render withholding) stay covered in the same file.

## Coordination — B4 MUST NOT DEPLOY WITHOUT C2

This is a hard deploy constraint, not a sequencing preference. **Deploy B4 and C2
(`agents/c2-route-cutover`) in the same deploy.**

B4 deletes `POST /v1/cloud/agent-gateway/catalog/{h}/mirror`.
**Update (F-040 fix commit): the client-side half of this is now fixed on b4
itself**, not deferred to C2. `ProductLifecycleRoot.tsx` no longer mounts
`useGatewayCatalogMirrorSync` at all — the hook and its
`gateway-catalog-mirror.ts` domain module are deleted, so there is no 60s
404-retry loop on this branch regardless of merge order. What originally read:

- ~~Write path: every signed-in desktop runs a 60s 404 retry loop, one per
  ready harness, indefinitely.~~ Fixed here — the mirror-sync hook is gone.
- **Read path (still the reason the pair must land together):** the settings
  All Models tab now speaks b4's real contract
  (`GET /v1/cloud/agent-models/{h}?authContextId=`), but that contract only
  exists once B4's server package is deployed. Deploying the TS side without
  B4's server routes live would 404 the All Models tab the same way the old
  mismatch did — hence the deploy-together constraint below still holds, for
  the read path only.

### Does any mirror caller outlive the endpoint after B4+C2? (reviewer question b)

**Revised.** The original answer (below, struck) routed the dangling-export
cleanup to C2. That premise no longer holds: F-040 (QA finding) found that
*this branch's own* TS clients still pointed at the deleted
`/v1/cloud/agent-gateway/catalog/*` routes — the SDK had never been
regenerated after the server-side rename, so `agent-gateway/catalog` client
functions, the `useAgentCatalog`/`useRefreshAgentCatalog`/
`useMirrorAgentCatalog` hooks, `useGatewayCatalogMirrorSync`, and its
`gateway-catalog-mirror` domain module were all still live on b4, all pointed
at 404s, and none of it was C2's to fix.

That cutover now lives on **this branch**, landed at F-040's fix commit:
`getAgentCatalog`/`refreshAgentCatalog`/`mirrorAgentCatalog` and their SDK
types are deleted outright (not renamed — the ingest route is
`authenticate_worker`-only, so no product client can ever call a refresh);
`useAgentCatalog`/`useRefreshAgentCatalog`/`useMirrorAgentCatalog` are
replaced by `useAgentModels`/`useUpsertAgentModelOverride`/
`useDeleteAgentModelOverride`; `useGatewayCatalogMirrorSync` and
`gateway-catalog-mirror.ts` are deleted along with their
`ProductLifecycleRoot` mount; `HarnessAllModelsSection` now renders no
Refresh button at all for the cloud-snapshot branch (only the
signed-out-local and runtime-gateway branches still have a real, callable
refresh). `cloud/sdk/src/generated/openapi.ts` is regenerated: zero
`agent-gateway/catalog` occurrences, `agent-models` paths present.

So on **b4**, after this commit, the answer to "does any mirror caller
outlive the endpoint" is **no** — there is no mirror/refresh export left in
this branch's SDK or hooks at all. Whatever `agents/c2-route-cutover` does
with its own (separately-forked) copies of `mirrorAgentCatalog()` /
`useMirrorAgentCatalog()` is C2's branch to resolve on its own terms — this
PR no longer has a stake in that question, since F-039 already rules that
C2's duplicate router is dropped in favour of b4's at merge time.

<details><summary>Original answer (superseded by the above)</summary>

**Yes — one dangling client function survives the pair, and it needs a C2-side
deletion.** Checked against `agents/c2-route-cutover@afd855276`:

- C2 keeps `mirrorAgentCatalog()` in `cloud/sdk/src/client/agent-gateway.ts`
  (pointing at `` `${catalogPath(h)}/mirror` `` = `/v1/cloud/agent-models/{h}/mirror`)
  and the `useMirrorAgentCatalog()` hook in `cloud/sdk-react/src/hooks/agent-gateway.ts`,
  whose docstring explicitly says it "stays as the ingest-route wiring for
  whatever caller lands next".
- B4 has no `/mirror` route at all — the mirror write is absorbed into
  `POST /agent-models/{h}/refresh`. Confirmed against the generated spec:
  `/v1/cloud/agent-models/{harness_kind}/mirror` is absent.

So after B4+C2 the SDK exports a function and a hook pointing at a route that
does not exist. Zero callers today (C2 deleted the only one), so nothing breaks at
runtime — but it is a live export aimed at a 404. Routing this to the C2
operator as requested; I have not touched C2's branch.

</details>

Two further B4↔C2 incompatibilities worth flagging in the same breath, because
they are not "residue" but genuine conflicts the merge must resolve:

1. **Both branches create an `/agent-models` router, in different modules.** C2
   adds `agent_models_router` inside `agent_gateway/api.py`; B4 adds `router` in
   the new `agent_models/api.py` package (the spec's code map). These do not
   compose — `cloud/api.py` conflicts, and merging both would register the same
   paths twice.
2. **The wire contracts differ.** C2's routes keep the pre-B4 shape
   (`?surface=&route=`, `AgentGatewayCatalogResponse`, `modelsJson`); B4's take
   `?authContextId=` and return `AgentModelsResponse` with `snapshotJson` +
   `origin`. C2's `HarnessAllModelsSection` still calls
   `useAgentCatalog({harnessKind, surface, route})` and posts
   `{surface, route, modelsJson}`, none of which B4's routes accept.

Net: the read surface (All Models) needs a C2-side follow-up to speak B4's
contract regardless of merge order. That is not a gap in this document — it is the
already-open Current-gaps bullet "The cloud snapshot is read only by the settings
All Models tab", which stays open until C3 wires the real consumers. No new gap
bullet is owed for the mirror endpoint itself: the pair fully closes the
*endpoint*, and the surviving SDK export is a C2 deletion, tracked above.

### Other notes from review cycle 1

- The snapshot scope carries **no unique key** (spec §Storage), so "the active
  row" is plural in principle; reads order by `(probed_at DESC, id DESC)`.
- One router, auth chosen per route, replacing two routers that shared the
  `/agent-models` prefix.
- `agent_gateway_credits_exhausted` has no product-server producer after this PR
  (its only one was the deleted prober). Kept as a client contract, with the hole
  recorded as a `model-gateway.md` gap; exhaustion itself is still enforced by the
  key-disable and state-render walls.
- New `model-catalog.md` gap bullet: retained `inactive` snapshot rows have no
  retention bound.


pablonyx:
**QA (integration) — F-039: #1520 and #1521 are two incompatible implementations of the same spec gap item**

Found on a merge probe over the QA integration branch (a-stack head + b-stack head + C1 + D2 + A7). Read-only — done with `git merge-tree --write-tree` and a throwaway `commit-tree`, nothing written to either branch.

**Each branch merges clean alone; together they collide hard.**
```
$ git merge-tree --write-tree HEAD origin/agents/b4-snapshot-rekey   # 0 conflicts
$ C1=$(git commit-tree <that tree> -p HEAD -p origin/agents/b4-snapshot-rekey -m probe)
$ git merge-tree --write-tree "$C1" origin/agents/c2-route-cutover
CONFLICT (content): scripts/max_lines_allowlist.txt
CONFLICT (content): server/proliferate/server/cloud/agent_gateway/api.py
CONFLICT (content): server/proliferate/server/cloud/api.py
CONFLICT (modify/delete): server/tests/integration/test_agent_gateway_catalog_api.py
    deleted in <b4> and modified in c2
CONFLICT (modify/delete): server/tests/integration/test_agent_gateway_catalog_mirror_api.py
CONFLICT (content): specs/codebase/platforms/product/model-gateway.md
```

**Cause:** both branches independently did the `model-catalog.md` gap item `- [ ] Route rename /v1/cloud/agent-gateway/catalog/* → /v1/cloud/agent-models/*`.

- **#1520 (c2)** does it in place: adds `agent_models_router = APIRouter(prefix="/agent-models")` inside `cloud/agent_gateway/api.py`, re-decorates the five handlers onto it, includes it from `cloud/api.py`, and **checks the gap item off**.
- **#1521 (b4)** does it by relocation: new `cloud/agent_models/` package with `router` *and* `worker_router` both at `prefix="/agent-models"`, deletes the handlers from `agent_gateway/api.py`, deletes `agent_gateway/catalog.py` + `db/store/agent_gateway/catalog.py` + the two integration test files, and **leaves the gap item open** (still `- [ ]` at line 690).

So the spec disagrees with itself about whether the work is done, depending on merge order. And the routes are doubly defined:

| route | c2 | b4 |
|---|---|---|
| `GET /agent-models/{harness}` | `AgentGatewayCatalogResponse` | `AgentModelsResponse` |
| `POST .../refresh` | product-user auth | **`worker_router`** + `authenticate_worker` |
| `POST .../mirror` | present | **absent** — absorbed into refresh |
| `PUT/DELETE .../override` | present | present, different response model |

If both land, FastAPI gets two registrations for the same paths from two routers with **different response models and different auth dependencies**. First-registered wins silently, and which one that is depends on import order in `cloud/api.py` — itself a conflicted hunk. The `refresh` divergence is the sharp one: whether that ingest route is callable by a product user or only by a Worker gets decided by merge order.

Worth noting b4 matches the spec's own §Cloud routes prose ("Absorbs today's separate `refresh` and `mirror` endpoints; the server never generates snapshots itself") while c2 preserves `mirror` — so c2 checks off a gap item whose adjacent prose it doesn't satisfy.

**Builder's call, not QA's.** One implementation has to go. Whichever survives: drop the loser's router declaration + `include_router` line; make sure the two deleted integration test files don't come back through c2's modify side of the modify/delete conflict (that would resurrect tests for handlers b4 deleted); check the gap item off exactly once; reconcile the single `max_lines_allowlist.txt` entry for `agent_gateway/api.py` (c2 says 415 "+10 for the agent_models_router declaration", b4 shrinks the same file).

**Why per-PR CI can't see it:** not a shared test (F-007), not one branch's test against another's behavior (F-038) — two branches independently completing the same spec item by different designs. Each internally coherent and green. Part of the collision wouldn't even fail a test; duplicate route registration silently picks one auth dependency at import time.

Full writeup in `codex/ledger-qa.md` §F-039.

pablonyx:
**QA — F-040: this branch deletes the catalog routes server-side but leaves every TS client calling them (settings All Models 404s)**

This one is a defect on #1521 alone, not a merge artifact. Read-only QA; nothing pushed.

The branch removes the old cloud catalog routes and asserts the removal:
```python
# server/tests/integration/test_agent_models_api.py:442
async def test_the_old_catalog_routes_are_gone(self, client):
    for path, method in ((f"/v1/cloud/agent-gateway/catalog/{HARNESS}", "get"), ...):
        assert (await getattr(client, method)(path, headers=headers)).status_code == 404
```
and server-side they really are gone (`git grep -n 'catalog/{harness_kind}' -- server` → no output).

But the TypeScript client still points at the deleted paths:
```ts
// cloud/sdk/src/client/agent-gateway.ts:26
function catalogPath(harnessKind: string) {
  return `/v1/cloud/agent-gateway/catalog/${encodeURIComponent(harnessKind)}`;
}
:114 path: catalogPath(harnessKind),               // getAgentCatalog
:126 path: `${catalogPath(harnessKind)}/refresh`,  // refreshAgentCatalog
:138 path: `${catalogPath(harnessKind)}/mirror`,   // mirrorAgentCatalog
:150 / :161 .../override                           // upsert / delete
```
and those are live behind hooks real surfaces render — `cloud/sdk-react/src/hooks/agent-gateway.ts:163` (`getAgentCatalog`), `:172` (`refreshAgentCatalog`), consumed by `HarnessAllModelsSection.tsx:63` via `useAgentCatalog`.

Net: on this branch, **Settings → Agents → harness → All Models** against a cloud auth context issues `GET /v1/cloud/agent-gateway/catalog/{harness}` to a 404. Refresh, mirror, and both override mutations are dead the same way.

**Why CI is green anyway** — two compounding gaps:

1. `cloud/sdk/src/generated/openapi.ts` was never regenerated. It still declares the four old paths (lines 1445/1462/1479/1504) and has **zero** occurrences of `agent-models`, so the generated contract agrees with the stale hand-written client and typecheck passes against a contract the server no longer serves.
2. Nothing in CI checks cloud-SDK drift. `.github/workflows/ci.yml:246` gates only the *anyharness* SDK:
   ```yaml
   - name: Verify generated SDK artifacts are committed
     run: git diff --exit-code -- anyharness/sdk/generated/openapi.json anyharness/sdk/src/generated/openapi.ts
   ```
   The cloud generator is Makefile-only (`cloud-client-generate` → `cloud-openapi`, Makefile:1510-1513) with no equivalent check. A cloud route rename has no mechanical consumer gate at all.

Worth fixing structurally beyond this PR: add the same `git diff --exit-code` drift check for `cloud/sdk/src/generated/openapi.ts`, so the next cloud rename can't land half-done.

This is precisely the risk `model-catalog.md` §Cloud routes calls out — "Renames are hard cutovers with no alias windows; all consumers are first-party" — and then enumerates the consumers. For contrast, #1520 *did* update them (`cloud/sdk/src/client/agent-gateway.ts:30` → `/v1/cloud/agent-models/...`, plus sdk-react, `openapi.ts`, the mirror-sync hook, the playground client, `ProductLifecycleRoot`). More evidence for F-039's "pick one implementation".

**Fix if this branch's implementation survives:** `catalogPath` → `/v1/cloud/agent-models/{harness}`, drop `mirrorAgentCatalog` (no mirror route here), re-run `make cloud-client-generate`, update the playground client test.

Full writeup in `codex/ledger-qa.md` §F-040. Related: §F-039 (this branch vs #1520 implementing the same spec gap item two ways).

pablonyx:
**QA — merge-order note for #1521: an undeclared migration dependency on #1508**

Not a defect — the graph is clean — but it constrains the merge order, so recording it here.

I walked every `server/alembic/versions/*.py` across `origin/main` plus all 18 `agents/*` branches and built the full rev→down_revision graph. Result: **a single head in the union**, zero revisions with a divergent `down_revision`, and every multi-child parent already has all its children on main (pre-existing historical merges, not new forks). `alembic upgrade head` won't need a merge revision. This branch is also the *only* one in the stack that adds a migration — c2 and a8 add none.

The constraint: this branch's migration chains through b2's.
```
35fa0038d703 (main)  ←  28adf1a9e376 (b2-per-harness-keys, agent_gateway_enrollment_key)  ←  b7c1e4d9f082 (this branch)
```
`28adf1a9e376` exists on exactly three refs — b2, b3, b4 — and **not** on main. This branch carries its own copy, so `merge-tree origin/main <this branch>` produces a tree with both files and the chain holds even if this lands first. No breakage today.

But it means #1521 silently depends on #1508's migration. If b2 is reworked or its migration renumbered during rebase — and b2 is implicated in both F-007 and the `check_max_lines` allowlist collision, so it may well be touched — this branch's `down_revision` dangles and `alembic upgrade head` fails at runtime with **no CI signal**: `scripts/check_migration_heads.py` counts heads, it doesn't verify that every `down_revision` resolves.

Suggested for the merge choreography: **#1508 before or together with #1521**, and if b2's migration gets renumbered on rebase, update this branch's `down_revision` in the same pass.

Also ran all 11 repo-shape ratchets on the merged tree for completeness — 10 pass, the only red is the already-known `check_max_lines` numeric-allowlist collision (`test_agent_gateway_api.py` 734 vs allowlisted 723, `test_agent_auth_materialization.py` 679 vs 614). Notably `check_server_boundaries`, `check_frontend_boundaries`, `check_anyharness_boundaries` and `check_docs` all pass, which is the set a five-branch merge would trip on a layer violation or a broken spec link.

Details in `codex/ledger-qa.md`.

pablonyx:
## F-040 closed at `6a8c6eb5d`

QA finding F-040 (and the b4 side of F-039's route-rename gap bullet): b4
deleted `/v1/cloud/agent-gateway/catalog/*` server-side, but the TS clients
and the generated SDK were never cut over — Settings → Agents → harness →
All Models 404'd on this branch, along with refresh and both override
mutations.

### Deleted (not renamed)

- `cloud/sdk/src/client/agent-gateway.ts`: `catalogPath`, `getAgentCatalog`,
  `refreshAgentCatalog`, `mirrorAgentCatalog`, `upsertAgentCatalogOverride`,
  `deleteAgentCatalogOverride`.
- `cloud/sdk/src/types/agent-gateway.ts`: `AgentGatewayCatalog`,
  `RefreshAgentGatewayCatalogRequest`, `MirrorAgentGatewayCatalogRequest`,
  `AgentGatewayCatalogOverride`, `UpsertAgentGatewayCatalogOverrideRequest`.
- `cloud/sdk-react/src/hooks/agent-gateway.ts`: `useAgentCatalog`,
  `useRefreshAgentCatalog`, `useMirrorAgentCatalog`,
  `useUpsertCatalogOverride`, `useDeleteCatalogOverride`.
- `cloud/sdk-react/src/lib/query-keys.ts`: `agentGatewayCatalogRootKey`,
  `agentGatewayCatalogKey`.
- `apps/packages/product-client/src/hooks/agents/lifecycle/use-gateway-catalog-mirror-sync.ts`
  and `apps/packages/product-client/src/lib/domain/agents/gateway-catalog-mirror.ts`
  (+ its test) — the 60s mirror-sync tick and its domain logic, whole files.

No refresh/mirror function exists anywhere in the new surface: the real
ingest route (`POST /agent-models/{h}/refresh`) is `authenticate_worker`-only,
so no product client can ever call it successfully.

### Added (real b4 contract)

- `agentModelsPath`, `getAgentModels(harnessKind, authContextId, client)`,
  `upsertAgentModelOverride`, `deleteAgentModelOverride`.
- Types: `AgentModels` (`Schema<"AgentModelsResponse">`), `AgentModelOverride`,
  `UpsertAgentModelOverrideRequest`.
- Hooks: `useAgentModels({harnessKind, authContextId}, enabled)`,
  `useUpsertAgentModelOverride()`, `useDeleteAgentModelOverride()`.
- Query keys: `agentModelsRootKey()`, `agentModelsKey(harnessKind, authContextId)`.

### Consumer fixes

- `HarnessAllModelsSection.tsx`: resolves `authContextId` from the pane's
  existing `(surface, route)` model via a new best-effort mapping,
  `authContextIdForRoute()` (+ `apiKeyProviderHintForSurface()`) in
  `harness-catalog.ts`. Refresh button now only renders where a real,
  callable refresh exists (signed-out-local runtime refetch,
  runtime-gateway probe) — removed entirely for the cloud-snapshot branch,
  with a comment noting manual refresh returns in C3.
- `ProductLifecycleRoot.tsx`: dropped the `useGatewayCatalogMirrorSync` mount.
- `agents-playground-cloud-client.ts` (+ fixtures + tests): rewritten to the
  new `(harnessKind, authContextId)`-keyed fixture shape and the
  `GET`/`PUT .../override` request matcher (no refresh/mirror handling).
- `HarnessAllModelsSection.test.tsx`, `HarnessPane.test.tsx`: mocks and
  assertions updated to the new hooks/shapes; the two tests asserting a
  client-driven refresh-with-payload flow for the native/api_key route are
  replaced with a single test asserting no Refresh button renders there.
- Verified (grep) that the ~13 consumers of the *unrelated* local
  `#product/hooks/agents/derived/use-agent-catalog` hook (same name,
  different module — wraps the runtime agent/reconcile queries) are
  untouched; none import from `@proliferate/cloud-sdk-react`.

### Regeneration

`USE_EXISTING_POSTGRES=1 USE_EXISTING_REDIS=1 make cloud-client-generate` —
`cloud/sdk/src/generated/openapi.ts`: 0 occurrences of `agent-gateway/catalog`,
`/v1/cloud/agent-models/{harness_kind}` (+ `/override`, `/refresh`) present.

### Gap closure (F-039 b4 side)

`specs/codebase/platforms/product/model-catalog.md` §Current gaps: removed
the route-rename gap bullet (closed exactly once, here). The competing edit on
`agents/c2-route-cutover` is expected to be dropped by the merge operator per
the F-039 ruling — not touched from this branch.

### Verification

- `pnpm --filter @proliferate/cloud-sdk build`, `@proliferate/cloud-sdk-react
  build`, and the full `@proliferate/product-client` chain (`ui`,
  `product-ui`, `product-domain`, `design`, `product-surfaces`,
  `anyharness/sdk`, `anyharness/sdk-react`) — all typecheck clean, zero errors.
- `pnpm --filter @proliferate/product-client exec vitest run` — **630 files /
  3816 tests pass**, including the 5 directly touched suites
  (`HarnessAllModelsSection.test.tsx`, `HarnessPane.test.tsx`,
  `HarnessPane.hierarchy.test.tsx`, `agents-playground-cloud-client.test.tsx`,
  `ProductLifecycleRoot.test.tsx`).
- `scripts/check_max_lines.py` — pass, after reconciling the
  `HarnessPane.test.tsx` allowlist line count (932 → 885; net line count
  dropped since two client-refresh-payload test cases were removed).
- Python/server side untouched by this fix — b4's existing server-side test
  suite (already merged into this PR) covers it; CI covers it here.

Also revised the "reviewer question b" mirror answer in the PR body: the
mirror/refresh SDK deletion now lives on this branch, not routed to C2.

Per the standing constraint, **B4 still must not deploy without C2** — the
read-path contract mismatch (b4's `authContextId`-keyed routes vs. C2's
`surface`/`route`-keyed ones) is unchanged; only the write-path 404-loop half
of that risk is now fixed here.

Not merged, not rebased. c2's branch untouched.

pablonyx:
Ready for Pablo's review, paired with #1520 (which stacks on this branch and ships in the same deploy train). Verification state:
- This PR (B4) @934360cd4: opus verify complete (F-040 fix confirmed; TS cutover total: cloud/sdk client+types, sdk-react hooks, regenerated openapi.ts drift-clean, mirror-sync hook deleted, playground re-keyed, allowlist exact). Server CI `test` job green at this head.
- #1520 (C2) @0c758a042: independent opus verify = READY (diff exactly 5 files +2/−160 dead code, mutation-proofed, 17/17 checks, live helpers intact).
Deploy pairing: both PRs in the same train; b2 (migration parent) lands before/with this one; #1533 lands LAST among migration PRs (alembic head reconciliation — see its review comment).

pablonyx:

---

PR #1522, opened 2026-07-26
feat(anyharness): wire the probe reconciler's pokes and make launch validation snapshot-first

pablonyx:
PR A8 of the agent-auth/catalog program. Base: `agents/a7-probe-engine` (retarget to `main` after A7 merges).

## Spec of record

`specs/codebase/platforms/product/model-catalog.md`, "The snapshot reconciler":

> Probing is a convergence primitive, wired the same way installs converge: one reconciler, many pokes. The reconciler's rule is pure: for every (installed harness, active auth context), if the document has no fresh entry — missing, or stale by the rules above — probe it in the background and write the entry. Fresh entries are never re-probed; running it twice does nothing twice.
>
> Pokes are a closed set, and every one is fire-and-forget — no poke ever blocks the operation that raised it.

and "Launch validation":

> The runtime validates a requested model against the active universe — the snapshot entries for the active contexts, shipped catalog only where no fresh entry exists yet… the two error kinds keep their shapes.

## Gap bullets closed

Two, both struck from `Current gaps`:

1. **"None of the reconciler's AUTOMATIC pokes is wired."** All six sites now fire.
2. **"Launch validation checks the shipped catalog's model list only; probed capability never enters the universe."** It now reads the union.

One bullet was **rewritten rather than struck**: the picker projection. `models`/`visible_models` deliberately stay catalog-only, because widening the *menu* needs the enrichment join (curated display name + control wiring) that is its own open gap — a model shown with no display name would be worse than one absent from the menu. Launch accepting more than the menu shows is pre-existing and in the safe direction (already true for trial-verified rows). The bullet now says exactly that.

The Code map gained `catalog/gateway_plan.rs`, `catalog/universe.rs`, `catalog/selection.rs` and `model_snapshot/universe.rs`, and the "Deleted by this design" paragraph now states which two pieces of the old chain survive (the `GET /v1/models` materialization fetch, and `seedModels` as a warned floor) and why.

## Design sections implemented

- **§3.1 startup** — `spawn_startup_pass`'s third step. The rev-2 doc withdraws the ordering claim, and the code says why in situ: `reconcile_when_idle` returns at *admission*, so this poke races the installs it follows. Harmless — an entry judged against a mid-install manifest is Indeterminate (not stale) or a correct observation of the machine as it is now.
- **§3.2 reconcile per-agent completion** — the sole guarantor of probe-after-install ordering. Gated on `Completed` only: a failed install leaves the old binary and a skipped one installed nothing, so neither moved the identity an entry is bound to.
- **§3.3 install endpoint** — additive, after the outcome.
- **§3.4 auth applied (REPLACE)** — `schedule_gateway_probes` deleted.
- **§3.5 auth cleared** — beyond the brief's five, per the design. The widest fingerprint change there is; without it a disconnected account keeps a picker full of unreachable models.
- **§3.6 session launch (REPLACE)** — `GatewayModelResolve::schedule_launch_probe_if_stale` trait default and its impl deleted.
- **§2.3 plan continuity** — `GatewayModelPlanner`: live `/v1/models` memoized 5 min per (harness, base URL), `seedModels` as a **warned** floor.

## Why the two chains are replaced, not extended

Both old triggers keyed on `state.json`'s **global** revision, so any harness's key rotation invalidated every harness's probe. The poke's gate is per (harness, auth context), i.e. exactly as wide as the change.

The planner is not optional either. With the revision-keyed probe rows gone, the old resolver would degrade to `gatewayPolicy.seedModels` — and `render_opencode_gateway` writes the plan's ids into `opencode.json`, so the probe then "observes" precisely the four ids the launch just wrote. A gateway-side model add would be undiscoverable forever. The seed list survives only as a floor (an empty models map hard-fails the recipe), and an entry produced over that floor carries `"gateway model plan fell back to seed models"`.

## Contradictions

**C-A8-1 — resolved toward the spec (ruling stays queued as ratify-or-keep).** Cycle 1 shipped a blanket per-**model** union; this now implements **per-context override, exempting `provenance.viaTrialOnly` rows**, for three reasons:

1. the spec's tier law is written per context ("Where no fresh entry exists **for an active context**, the shipped catalog's models **for that context** fill in"), so the union was a real deviation rather than a resolution of ambiguity;
2. the exemption is exactly wide enough, measured not assumed: across all 13 real probe fixtures the only catalog rows that are available-but-unobserved are `viaTrialOnly` ones — claude's `claude-fable-5` and `claude-opus-4-8` under `anthropic-api`/`anthropic-oauth`, and `global.anthropic.claude-fable-5` under `bedrock`, with zero non-trial cases anywhere — so trial-exempt override keeps 100% of provably-launchable capability;
3. it recovers the pre-launch rejection the union forfeited. A downgraded gateway key makes the proxy stop listing a model; under override the user gets `SESSION_MODEL_GATED` naming what would unlock it, where the union accepted the launch and let them meet a provider 403 mid-session — same outcome, discovered later and attributed to the wrong layer.

The tradeoff is documented at `universe.rs`'s module head. Pablo can still ratify the permissive union later if he prefers it; nothing about the shape makes that hard to flip.

**C-A8-2 — design anchor.** §2.3 says the fetch is driven with `Handle::current().block_on` "only from the probe thread's own runtime". That panics — `Handle::block_on` panics inside a runtime worker, and the probe runner thread is a current-thread runtime running a `LocalSet`. Implemented as `spawn_blocking` plus a scoped thread with its own current-thread runtime.

## Judgment calls worth a reviewer's eye

- **Automatic pokes are off under `cfg(test)`**, matching the existing suppression of `spawn_startup_pass` and the loop scheduler. This was not hygiene: a real failure caught it. `workflow_runs_scripted_tests` installs a fake agent program and counts its ACP requests, and the launch backstop's probe added a second `session/new` (`left: 2, right: 1`), cascading 78 `PoisonError` failures through the crate-wide env mutex. Only the *automatic* pokes are suppressed — the status route, the manual refresh and the validation universe keep the real engine.
- **A `resolve_gateway_models_blocking` split on the seam.** A launch renders synchronously on the spawn path, so a fetch there is a stall the user watches; a probe is about to spawn a whole harness, so waiting for the list it will observe is the right trade. Only the engine calls the blocking form.
- **An expired memo beats the floor on the launch path.** It was really observed from this proxy; the floor is a curated guess that can only be a subset.
- **The universe read is not the reconciler's gate.** The gate short-circuits on the completed-attempt floor and the backoff window — in-memory facts about the *engine*, not the entry. A launch must get the same answer whether or not a probe is queued, so the read consults document + manifest + fingerprint and calls the pure evaluator.
- **`ObservedUniverseSource` seam rather than an `Arc<ModelSnapshotService>`** in `SessionService`: the engine takes an flock on the runtime home at construction, and requiring one to answer a pure validation question would put an flock in every session test. The default is the pre-probe universe.

## Deletions beyond the brief

Forced by dead code, not opportunistic: `GatewayModelResolver::{gateway_credentials, probe_and_record}` — their only callers were the two deleted triggers. `gateway_resolver.rs` 869 → 808 lines, allowlist tightened rather than left stale.

### A known, bounded regression on the legacy gateway-models route

`refresh_now` and `resolve_with_source` are kept, but stating that as "A9's boundary" undersold what actually changes, so plainly: **the deleted triggers were the only two writers of the `gateway_model_probe` table.** `resolve_with_source` behind `GET /v1/agents/{kind}/catalog/gateway-models` still reads only that table, so from this PR it reports `source: "seed"` until a user presses the legacy Refresh button. Two live consumers regress for the length of the window:

- the settings **All Models** tab shows the seed list with no probe freshness;
- the desktop→cloud **mirror push** stops entirely — `gateway-catalog-mirror.ts` pushes only entries with `source === "probe"` — so machineless surfaces (web new-chat, mobile, automations, workflow editors) keep serving whatever they last received.

Restoring a writer here would mean re-creating the revision-keyed table this design deletes, so the fix is the cutover, not a patch. The window's end conditions are both in this same program stack: the mirror consumer is **deleted** by the route cutover (#1520) plus the snapshot re-key (#1521), and the All Models tab is rebuilt on the runtime status/snapshot surface that A7 shipped.

**Deploy implication: A8 must not reach users without #1520 and #1521 in the same release train.** The spec carries this as an explicit `Current gaps` bullet rather than leaving it to be rediscovered.

`catalog/service.rs` hit 682 lines after the universe work, so the resolution ladder split into `selection.rs` — by responsibility (read surface vs the one decision procedure over it), not allowlisted.

## Tests — 51 new, all tier 1

| Suite | N | Pins |
| --- | --- | --- |
| `model_snapshot::wiring_tests` | 11 | Each poke site: target precision, staleness gating, no-engine no-op, uninstalled harness records nothing |
| `catalog::gateway_plan_tests` | 9 | Live list vs seed floor, memo + TTL expiry, per-harness invalidation, warned floor on failure *and* on an empty 200, failures not cached, launch never fetches, provider filtering |
| `catalog::service_universe_tests` | 10 | Empty universe reproduces pre-probe answers exactly, observation unlocks a gated model, uncatalogued model launchable, gated-not-unknown, garbage rejected, a downgraded key's model narrows pre-launch under the trial-exempt override law, variant-only observation, mode ladder untouched |
| `model_snapshot::universe_tests` | 6 | Fresh-only projection, auth-moved and harness-moved entries drop out, read-only mode serves, never blocks on an in-flight probe |
| `route_auth::plan::tests` | 1 | The trait surface after the default drop — a minimal stub compiles |
| `engine_tests::lifecycle` | 1 | Seed-floor warning persisted to the document (T-38's remaining leg) |
| `api::http::agent_auth::tests` | 2 | Poke naming: every harness, including `api_key`-only and emptied-sources |
| `api::router_tests` | 4 | Auth-state and install route contracts hold with pokes wired; poke-suppression proofs (see cycle 2) |

Two assertions are worth calling out because they would pass against a wrong implementation without them: `an_empty_universe_reproduces_the_pre_probe_answers_exactly` diffs the universe-aware entry point against the plain one across eight cases (the safety net every pre-probe machine depends on), and `trial_verified_models_stay_launchable_against_the_real_probe_fixture` pins the current trial-exempt override law — per-context override, exempting `provenance.viaTrialOnly` rows — against claude's real probe fixture and the real catalog, asserting the `viaTrialOnly` flag first so a catalog change fails loudly instead of silently regressing.

### Review cycle 1 additions

| Test | Pins |
| --- | --- |
| `cursor_is_never_probed_by_an_automatic_poke_but_a_manual_refresh_works` | Iterates **every** automatic `PokeReason` through all three poke entry points, then proves a manual refresh still probes. Covering only `Startup` would have passed against the bypassed code |
| `the_production_exclusion_list_covers_cursor_only` | The real `RuntimeProbeTargets`, so the engine test above cannot pass against an empty production list |
| `the_optional_poke_seam_pokes_when_wired_and_no_ops_when_suppressed` | The exact seam all six call sites run, in both handle states |
| `a_model_a_downgraded_key_no_longer_serves_is_refused_before_launch` | The per-context override's payoff, and that the rejection does not name the observed-lacking context |
| `trial_verified_models_stay_launchable_against_the_real_probe_fixture` | Trial exemption against the real catalog + real fixture, asserting the `viaTrialOnly` flag first so a catalog change fails loudly instead of silently regressing |
| `an_observation_overrides_its_own_context` / `a_trial_verified_row_survives_an_observation_that_omits_it` | Same universe, one flag different, opposite answers |
| `a_default_selection_follows_the_observation` | A default the machine cannot serve is not advertised, and the one chosen itself validates |

`cargo test --workspace`: green at `5191779c0` — `anyharness-lib` 1457 pass, 0 fail (was 1454 pre-cycle-2; the other workspace crates unchanged: 28 + 42 + 40 + 90 + 68 + 40 + 10). The only repo-wide failure is the known pre-existing `src-tauri` `app_config` nanosecond temp-path flake, in a file this PR does not touch. CI "Cargo check & test" is green at this exact head.

`clippy`: zero new findings in touched files. `execution.rs`'s `too_many_arguments` already fired at 10/7 before this change (now 11/7) — the same pre-existing finding, not a new one.

Repo-shape: all 12 `check_*.py` green. `max_lines_allowlist.txt` updated for `router_tests.rs` (1497 → 1806, exact `wc -l`, no slack) and `app/mod.rs` (662 → 708), and tightened for `gateway_resolver.rs` (869 → 808). No OpenAPI drift — no route shapes changed.

### Review cycle 2 additions

Three tests closing the cycle-1 verifier's findings:

| Test | Pins |
| --- | --- |
| `an_aborted_refresh_settles_to_idle_and_leaves_the_key_pokeable` (`engine_tests/concurrency_tests.rs`) | F-036: a **real** `handle.abort()` mid-`Running` (not a cooperative cancel), joined before asserting, leaves the slot `Idle` and a follow-up poke re-probes. Backed by the new `LiveStateGuard` RAII guard — the sole writer of the slot's `live` field, holding its own `Arc<ContextSlot>` so `Drop` settles the state even when the caller's future is gone |
| `an_automatic_poke_is_suppressed_at_the_router_when_the_engine_is_absent` + install-route sibling (`api/router_tests.rs`) | All three production poke call sites (agent_auth PUT/DELETE, agents install) against a genuinely installed harness with a marker-file fake launcher: the marker stays untouched when `automatic_poke_engine` is `None`, while the document writes still happen |

Mutation-verified (from commit `b3f16682c`): swapping each poke call site's `&state.automatic_poke_engine` argument for `&Some(state.model_snapshot_service.clone())` turned the corresponding test red with the marker file present, then reverted clean. (The `router_tests.rs` doc comment's "PR body" cross-reference points here.)

Also in cycle 2: the previously one-sided `service_universe_tests` catalog leg now asserts the non-trial side too (`assert!(unexempt_and_unobserved.is_empty())` over the real catalog — genuinely non-vacuous: `claude-fable-5`/`claude-opus-4-8` are the trial-exempt members, and a future non-trial unobserved model fails it loudly).

## Gate cells this must keep green

G2 (T3 gateway cells after A8+B4): snapshot exists, models non-empty, validation accepts a snapshot model and rejects garbage.


pablonyx:
**QA (integration) — F-036 is still open on this branch, and #1522 widens its window**

Context: I proved F-036 on the merged QA branch (a-stack head + b-stack head + C1 + D2 + A7) — a dropped/cancelled `refresh_now` future leaves the slot pinned at `Running`, and `gate_admits` (`model_snapshot/mod.rs:360-367`) then refuses *every* automatic poke for that key forever. Full repro in `codex/ledger-qa.md` §F-036. Read-only QA; nothing pushed here.

Flagging it on #1522 specifically because this branch rewrites `attempt.rs` — the exact defect site — and the live state is still unguarded:

```rust
:75  self.set_live_state(slot, status::LiveState::Queued);
:77  let _harness_permit = harness_gate.lock().await;      // droppable
:82  let _permit = self.probe_semaphore.acquire().await;   // droppable
:84  self.set_live_state(slot, status::LiveState::Running);
:97  let outcome = self.runner.run(...).await;             // droppable, up to per_probe_timeout (240s)
:97  self.set_live_state(slot, status::LiveState::Idle);
```

The new `spawn_blocking(...).await` for the plan sits *above* `Queued`, so it's outside the window — but the window itself is unchanged: `Queued` → lock → semaphore → probe → `Idle`, every step droppable. The HTTP path reaches it: `refresh_model_snapshot` awaits `service.refresh_now(...)` in the axum handler, so a client disconnect drops the future mid-probe.

No coverage either — `engine_tests/concurrency_tests.rs` on this branch has zero occurrences of `abort` / `drop` / `Drop`. And the commit's direction ("wait for a poke's effect instead of yielding a fixed number of times") makes the tests more patient about a poke landing, which is exactly the assertion shape that can't distinguish "slow" from "never".

**Why it matters most here:** #1522 is what makes automatic pokes the normal path. A wedged key is invisible from the status surface — it reads a plausible `queued`/`running` indefinitely — and only an explicit `refresh_now` recovers it.

**Suggested fix:** an RAII guard owning the slot's live state for the attempt's lifetime, resetting to `Idle` on drop. #1519 already established exactly this pattern in this module for `ProbeScratch` and the child process; live state is the one piece of attempt state that didn't get it.

pablonyx:
Fix stack from cycle-2 review, pushed at `5191779c0`:

**F-036 (MAJOR) closed** — `53b269a0f`. The attempt's live-state slot is now released by an RAII guard rather than by an explicit path at the end of the probe future, so a dropped/cancelled refresh (the caller's future dropped mid-await, e.g. a timeout racing the probe) can no longer leave the slot permanently wedged. Verified against `engine_tests::concurrency_tests::an_aborted_refresh_settles_to_idle_and_leaves_the_key_pokeable`, which drops an in-flight refresh mid-probe and asserts the key is pokeable again afterward — reproduced the bug against the pre-fix code (removing the guard reverts to the abort leaving the slot occupied forever) and confirmed the fix's test goes red/green as expected before landing.

**Cycle-2 review item 1 (router-level poke-suppression coverage) closed** — `b3f16682c`. Added two new tests in `router_tests.rs`:
- `agent_auth_state_routes_never_probe_while_the_automatic_engine_is_suppressed` (PUT + DELETE agent-auth-state)
- `the_install_endpoint_never_probes_while_the_automatic_engine_is_suppressed` (install endpoint)

Existing router tests only asserted status codes, which cannot distinguish "poked" from "silently no-op'd" — exactly the failure mode a handler accidentally wired to the unsuppressed `model_snapshot_service` handle instead of `automatic_poke_engine` would produce. The new tests seed a fake launcher that appends to a marker file before exiting nonzero (network-free, fast, unmissable), drive each handler, and assert the marker is never written.

Mutation-tested each of the three call sites independently by temporarily swapping `&state.automatic_poke_engine` for `&Some(state.model_snapshot_service.clone())` at the PUT, DELETE, and install poke sites in turn — each mutation turned the corresponding new test red with the marker-file evidence, and each was reverted with a clean `git diff` afterward.

**Cycle-2 review item 2 (stale PR body)** — fixed directly in the PR description: the `catalog::service_universe_tests` table row no longer says "never narrows" (replaced with the actual downgraded-key/trial-exempt-override behavior), and the "assertions worth calling out" paragraph no longer names the deleted `an_observation_never_removes_a_launchable_catalog_model` test — it now points at `trial_verified_models_stay_launchable_against_the_real_probe_fixture`, which is the test that actually pins the current trial-exempt override law.

**Cycle-2 review item 3 (vacuous test leg) closed** — `5191779c0`. `trial_verified_models_stay_launchable_against_the_real_probe_fixture`'s `if let Some(non_trial)` leg never ran against the real catalog fixture (every `anthropic-api` model is either one of the four observed selectors or `viaTrialOnly`), so the intended refusal check was silently untested. Replaced with an explicit assertion that the unexempt-and-unobserved set is empty, so a future catalog change that breaks the assumption fails loudly.

Verification: `cargo test -p anyharness-lib --lib` — 1457 passed, 0 failed, 1 pre-existing unrelated ignore. `python3 scripts/check_max_lines.py` — passed (repo max 600, component max 500). No merge/rebase performed; not flagging ready — leaving that to the usual review pass.

pablonyx:
**Cycle-2 adversarial review: READY** (opus verifier, @ `5191779c0`).

All four cycle-1 items confirmed closed in code:
1. **F-036 RAII guard** — `LiveStateGuard` is the sole writer of the slot's `live` field (grep-verified: only assignments are inside the guard + the Default initializer). Guard owns `Arc<ContextSlot>`, so task abort / future drop settles to `Idle` even with the caller's future gone; created before the harness gate + semaphore awaits, so a cancel while blocked there also settles. Test uses a real `handle.abort()` joined before asserting.
2. **Router-level poke suppression** — both tests trace end-to-end to a real spawn (marker file would be written on any leak); all three production call sites covered; `alreadyInstalled: true` is a genuine no-network skip.
3. **Vacuous test leg** — the non-trial assertion is genuinely non-vacuous against the real catalog (6 anthropic-api rows: 4 observed selectors + 2 viaTrialOnly).
4. **CI green at the exact head** — run 30230074667, all 17 lanes including Cargo check & test.

PR body refreshed this cycle: test counts corrected (45→51 new; wiring 11, universe 10, router 4), allowlist number fixed (1497→1806), cycle-2 section added with the mutation-verification output the `router_tests.rs` doc comment references.

Non-blocking nits on record from the verifier: 500ms sleep window in the suppression tests is a wall-clock negative assertion (generous vs the fake launcher, acceptable tier-1); DELETE-leg would-be poke targets the baseline context rather than gateway (still discriminating); two stray blank lines (cargo fmt is not a CI gate here).

**Ready for Pablo's review.** Stacks on A7 #1519 (retarget to main after A7 merges). Deploy constraint unchanged: must ride the same release train as #1520 + #1521.

pablonyx:

---

PR #1524, opened 2026-07-26
docs(design): consolidate the design system into one first-principles document

pablonyx:
## What

Four commits that make the design system self-describing and remove external attribution from canonical artifacts:

1. **`8a80e5f34` chore(design): make token provenance self-contained** — tokens.ts/motion.ts/react-native.ts headers and rationale comments rewritten to carry their own reasoning instead of citing archived program documents or external reference captures. Comment-only: zero value lines changed (verified by diff filter on `dark:`/`light:`/`themeFallback:`). Also fixes one stale comment claim (the appearance gate has no `max-w` rule).
2. **`cb8150727` docs(specs): de-attribute canonical design citations** — composer.md / transcript.md / appearance-scaling.md keep every stated value and rule but drop external product attributions (agent-integration mentions are untouched — those describe a real product feature). Removes one tracked reference screen-capture binary and repoints its two metadata references to the local untracked capture tree.
3. **`6ec77bf7b` refactor(design): rename diff-surface token namespace** — `--codex-diffs-*` → `--diff-view-*` (10 names, 17 files). The frozen disposition census keys are intentionally untouched (they record pre-consolidation history and check-theme asserts them); only live names and the map's forwarding values move.
4. **`4b9254391` docs(specs): consolidate the design system into one first-principles document** — design-system.md rewritten 343 → 1086 lines as the single design doc: the closed type ramp with its actual design rules, the color role model (symmetric-opacity house form, surface/border/text ladders, interaction-state ladder), elevation, spacing/containers, radii, motion, icon tiers, layering, the folded component-library catalog (component-library.md deleted, all referrers repointed), and the change-control model. Zero external product references; every number re-verified against tokens.ts at HEAD.

## Verification

- `pnpm --filter @proliferate/design build` (byte-equality + Tailwind compile + frozen census: 283 live / 175 shipped / 40 retuned / 70 removals / 15 aliases) — pass
- `pnpm --filter @proliferate/product-client test` — 631 files, 3823 tests pass
- `check_docs.py` (229 files) / `check_frontend_boundaries.py` / `test_check_frontend_boundaries` (8 tests) / `check_appearance_scaling.py` — all pass, re-run independently at final HEAD
- Self-check greps on the new doc for every retired citation form — zero hits

## Deliberately out of scope (follow-ups)

- ~130 code-comment attribution lines across ui/product-ui/product-client components — deferred to avoid conflicting with the in-flight transcript/composer and sidebar alignment PRs that rewrite many of those files
[...]

PR #1525, opened 2026-07-26
Playground: front door + component-library spec sheet

pablonyx:
## What

Dev-only playground overhaul (stacked on #1524 — retarget to main after it merges):

1. **`4e510fd63` — index shell + chat route move.** `/playground` is now a front door listing all 9 playground destinations; the chat scenario gallery moves to `/playground/chat`. Old deep links keep working: `/playground?s=<key>` redirects to `/playground/chat` preserving the query. Same `import.meta.env.DEV` + lazy pattern as every existing route — nothing new ships in production bundles.
2. **`fb5847d6d` — `/playground/library`: the component-library spec sheet.** Every sanctioned `@proliferate/ui` / `@proliferate/product-ui` component rendered tier by tier under the real theme CSS chain, with a dark/light toggle that drives the app's actual appearance preference (`useColorMode`), not a dev hack. Backed by per-tier registry modules (34 primitives + 22 patterns + 4 icon sets + 11 product-ui patterns = 71 entries) and a **parity test** that re-derives the expected set from both packages' real `exports` maps at test time: a sanctioned component without a sheet entry — or a stale entry — fails CI with a readable diff. This closes the "catalog and exports are hand-synced" gap named in design-system.md's Current Gaps from the sheet side.
3. **`a5a218f43` — hygiene:** shared `noop` reused instead of two local redefinitions, stale hardcoded tab label fixed, `AgentsPlaygroundPage` split into the page/component shape every other route follows.

Deliberately not done (logged): scenario-key renames, merging the two overlapping git fixture files.

## Verification

- `pnpm --filter @proliferate/product-client build` + full test suite: 3,826 tests pass (one pre-existing teardown-flaky suite failed under load, passes in isolation, untouched by this diff)
- Registry parity test: 3/3, 71 = 71 exact
- `pnpm run shared:typecheck`, appearance scaling, boundaries, docs checks: all pass


pablonyx:
## Reviewer pass (at head `a5a218f43`)

Read the load-bearing files directly: the App.tsx route diff (follows the existing DEV-guard/lazy pattern byte-for-byte in shape), `PlaygroundIndexPage` (redirect logic correct — `?s=` deep links preserve the full query), `PlaygroundLibrary`/`PlaygroundIndex` (library components + semantic tokens only, real `useColorMode` mechanism), and the parity test in full.

The parity test is the piece that matters and it's built right: it re-derives the expected set from the two real `package.json` exports maps at test time (no copied list to rot), asserts exact two-way coverage with readable diffs, per-tier scoping, and duplicate detection. 71 = 71 verified against the maps independently.

Gate results verified from the implementation run: full product-client suite (3,826 tests; the one failing suite is `authenticated-markdown-cascade.test.ts` teardown flake — confirmed untouched since `e7f005180` and passing in isolation), repo-wide typecheck, appearance/boundaries/docs gates all green.

No blocking findings. Stacked on #1524; retarget to main after it merges.

PR #1532, opened 2026-07-26
fix(cloud): cold sandbox access schedules repair instead of dead-ending in 409

pablonyx:
## The dead end

`load_cloud_sandbox_runtime_access` raises a typed 409
`cloud_sandbox_runtime_not_ready` whenever a sandbox row carries no runtime
access material. There are two ways a row gets there: it was never stamped (a
row exists because ensure created it — ensure never provisions), or provider
loss cleared the columns via `mark_cloud_sandbox_provider_missing`.

Nothing on the access path started the materialization that would stamp it. The
gateway resolved, 409'd, and returned; the client retried; the row was still
cold; the same 409 came back. The only escape was for the user to trigger some
*other* action that happens to materialize (create a workspace, materialize a
repo environment). A cold personal sandbox behind a workspace the user is
already sitting in had no such action — it 409-looped indefinitely.

Note the precise boundary this does **not** touch: *paused* sandboxes were
already fine. Their stored address stays valid across a pause, so forwarded
traffic wakes them via E2B auto-resume. The gap was only rows with no stamped
access at all.

## What changed

Cold access still returns the identical 409 — provisioning takes 30-60 s, far
too long to hold a request open, and clients already render this code as
"connecting" behind a retry policy. What changes is that the cold resolution now
*schedules* exactly one background materialization for that sandbox, so the
client's retry is waiting on work that is actually running.

The wire contract is unchanged: same status, same code, same body.

## The seam, and why

`load_cloud_sandbox_runtime_access` is the single choke point for turning a row
into runtime coordinates, and it has four callers:

| Caller | Kind |
| --- | --- |
| `gateway/service.py::_resolve_cloud_sandbox_gateway_access` | request-time |
| `workspaces/service.py::_load_ready_runtime_access` | request-time |
| `workspaces/materializations/service.py::_read_managed_cloud_source` | request-time |
| `materialization/materialize/workflow_runtime.py` | inside `run_cloud_sandbox_operation` |

Putting the trigger *in* the loader would have fired it from inside
materialization itself — the fourth caller runs within the operation that does
the repairing, so it would schedule a repair for the work already in flight.
Putting it only in the gateway would have left the two other request-time paths
dead-ending exactly as before.

So the choke point gained a sibling: `load_cloud_sandbox_runtime_access_or_repair`,
which delegates to the bare loader and, on `cloud_sandbox_runtime_not_ready`
only, schedules the repair before re-raising. The three request-time callers use
it; the materialization-internal caller keeps the bare loader. The bare loader
is byte-identical, which also means every existing test that monkeypatches it
keeps working.

In the gateway this lands on the cache-miss path by construction:
`ensure_cloud_sandbox_gateway_access` serves cached access for 60 s behind a
per-user `asyncio.Lock`, and only a miss reaches resolution — a cached hit by
definition already resolved. Failures are never cached, so a polling client
re-enters resolution on every attempt, which is precisely why the scheduling
guard cannot be the per-user lock.

## Stampede safety

Two layers, both reusing existing infrastructure rather than inventing a lock:

1. **A non-blocking, token-owned Redis claim** keyed `sandbox-repair:{id}`,
   namespaced under `settings.redbeat_key_prefix` on `settings.redbeat_redis_url`
   like the existing materialization lock. `SET NX EX` decides who schedules;
   losers do nothing and let the winner's run repair the row. Cross-process by
   construction, so it holds across API workers, and self-heals on TTL expiry
   (900 s) if the winning process dies before releasing. Release is a token
   compare-and-delete, so a previous holder finishing after its TTL lapsed
   cannot drop the live holder's claim.

2. **The existing per-sandbox `redis_materialization_lock`** inside
   `run_cloud_sandbox_operation`, unchanged, which serializes any run that does
   start.

The claim is released in the background task's `finally`, not in the request, so
the window it covers is the whole provisioning attempt — which is exactly the
window a polling client hammers. Releasing on completion (success *or* failure)
also means a failed attempt is retryable on the next access rather than
suppressed for the full TTL.

A Redis outage returns "no claim", so nothing is scheduled and nothing raises on
the request path. That costs nothing real: `redis_materialization_lock` needs the
same Redis, so a materialization scheduled during an outage could only fail on
the lock.

## Judgment calls

- **Not `run_after_commit`.** Every other scheduler in `materialization/service.py`
  defers to after-commit. This one cannot: its caller is a request that is about
  to *raise* the 409, so the session rolls back and an after-commit callback
  would be discarded — while the Redis claim stayed held, suppressing all repair
  for a whole TTL. It calls `runner.spawn_materialization_task` directly and
  takes no session.
- **No third billing gate.** The gateway path runs
  `require_cloud_provisioning_configured()` and
  `assert_cloud_sandbox_resume_allowed_for_owner` in
  `ensure_cloud_sandbox_ready` *before* access resolution, and
  `connect_ready_sandbox` re-asserts `assert_cloud_sandbox_resume_allowed`
  inside the materialization itself. Scheduling deliberately adds no gate of its
  own and bypasses none.
- **Skip conditions.** Destroyed rows (gone, not cold) and deployments without
  `cloud_provisioning_configured` (a background run could only fail) schedule
  nothing.
- **Import cycle.** `materialization.service` imports the cloud_sandboxes
  module, so the reverse import is function-local.
- Scheduling failures are logged, never raised — the caller's 409 is the
  contract.

## Test evidence

New `server/tests/integration/test_cloud_sandbox_cold_access_repair.py` (8 tests)
against real Postgres and the real Redis the suite already needs for
materialization locks; provider side stubbed per the repo testing standard.

- cold gateway access 409s with the unchanged code *and* schedules one repair,
  targeting `sandbox-repair:{sandbox.id}`
- 5 sequential polls (cache reset each time) schedule exactly **1**
- 8 concurrent callers on separate sessions schedule exactly **1**
- releasing the claim allows the next access to schedule again
- destroyed sandbox: 409, schedules nothing
- ready/stamped sandbox: resolves normally, schedules nothing
- provider-loss-cleared row (via `mark_cloud_sandbox_provider_missing`):
  schedules a repair
- unconfigured deployment: schedules nothing

The two stampede assertions were mutation-tested: forcing the claim to always
succeed makes them fail `5 == 1` and `8 == 1`, so they bite.

3 new unit tests in `tests/unit/test_redis_lock.py` cover the claim primitive
directly (single winner across 5 concurrent contenders, reuse after release,
stale-token release not dropping a live holder, outage → `None`).

Suites run: `tests/unit` **1173 passed**; `tests/integration -k "cloud or
workflow"` **354 passed**; `ruff format` and `ruff check proliferate tests`
clean; `python3 scripts/check_docs.py` passed (230 files).

## Spec gaps struck

The cold-access gap bullet is removed from all three docs that carried it, and
the body prose that asserted the dead end is replaced with the landed behavior:

- `sandbox-lifecycle.md` — owns the mechanism: new law "A cold access resolution
  schedules its own repair" alongside the existing gateway/E2B/materialization
  law, plus the proof entries.
- `sandbox-access.md` — the choke-point section now names the repairing seam and
  the request-time/materialization-internal split; the failure mode no longer
  reads "materialization repairs it, the gateway does not".
- `sandbox-gateway.md` — resolution step 3 now names the trigger and why it sits
  on the cache-miss path; the failure mode keeps "never a gateway retry" but adds
  that the 409 converges.

All three bullets had deferred to a "cold-start choreography ruling
(wake-and-poll vs provision-on-access), still open" — no mechanism was
prescribed, so the design constraints in the brief governed. The 409-preserving,
schedule-and-return shape is closest to provision-on-access without the
wake-and-poll blocking cost.

## Review outcome

A dedicated review returned **REQUEST CHANGES** with one blocker and four
serious findings, all verified with live probes. Fixed in `d7c4a4a70`.

### The blocker (B1): fresh-signup paging loop

`ensure_cloud_sandbox_ready` created the personal sandbox row with `db.add` +
`flush` and **no commit**; access resolution then raised the typed 409, which
rolled the request session back, so the row never persisted. Every poll of a
newly signed-up user therefore minted a *new* sandbox id — a new
`sandbox-repair:{id}` claim key that could never collide — so each poll spawned
its own background repair, and each of those repairs called
`load_personal_cloud_sandbox`, got `None`, raised
`CloudMaterializationError("cloud_sandbox_missing")`, and hit the runner's
generic handler, i.e. `report_critical`: one Sentry page per poll.

### What was fixed

- **B1a — commit, not detect.** `ensure_cloud_sandbox_ready` now commits the
  ensured row before its caller resolves access. Caller analysis: the function
  has exactly three production callers (`POST /cloud-sandbox/ensure`, `POST
  /cloud-sandbox/wake`, and the gateway resolver), all request entry points that
  commit at the end of the request anyway, so this only moves the commit
  earlier; none relies on rollback discarding the ensure. The five direct
  `ensure_personal_cloud_sandbox_exists` callers (secret-set and
  repo-environment materializers, repositories service, GitHub-App trigger,
  workflow target plan) do not route through this function and are untouched.
  The commit is what makes the sandbox id stable (so the claim dedupes) and the
  row visible to the repair's own fresh session.
- **B1b — benign missing row.** `_repair_materialize_sandbox` now treats
  `cloud_sandbox_missing` as a routine race: INFO log, no re-raise, claim still
  released in `finally`. A vanished row can no longer page. Every other
  materialization error propagates and pages exactly as before (pinned by its
  own test).
- **S3 — claim stranding.** The spawn is wrapped: a synchronous failure (loop
  shutting down) releases the claim before propagating instead of suppressing
  repair for the 900 s TTL. The release in the repair's `finally` is now
  `asyncio.shield`ed so shutdown cancellation cannot skip it — matching the
  shielding `db/engine.py` applies to session rollback/close.
- **S2 — false comment.** The old comment claimed every request-time access path
  runs the owner resume gate first. False for `_read_managed_cloud_source`
  (`POST /workspaces/{id}/materializations` has no billing gate) and only
  transitively true for `_load_ready_runtime_access`. Rewritten to state the
  actual load-bearing guarantee: scheduling is un-gated for some callers, but
  `connect_ready_sandbox` asserts `assert_cloud_sandbox_resume_allowed` before
  any provider or VM work, so a held subject can drive at most a Redis claim
  plus a background task that stops at that gate — never an E2B create.
- **S5 — invariant guard.** The claim is keyed by sandbox id while the repair
  materializes the owner's *current* personal sandbox; those coincide only
  because of the `ux_cloud_sandbox_personal_active` partial index. Scheduling now
  skips and logs anything that is not a personal row, with the future
  org-sandbox hazard named in the comment.
- **S4 — spec honesty.** The cold-access law in `sandbox-lifecycle.md` now
  records the process-death window (the claim is in-process fire-and-forget; a
  restart mid-provision strands it and suppresses repair until the 900 s TTL
  lapses — accepted, the TTL self-heals) and states explicitly that this law
  resolves the previously open cold-start choreography question in favor of
  provision-on-access, with the client polling the unchanged 409. The same
  framing is cross-referenced from `sandbox-access.md` and `sandbox-gateway.md`.
- **Nits:** removed the `_destroyed_value` alias and the tautological
  `owner_user_id` assert, dropped the rename-fragile function-identity assert
  (the `claim_key` assert carries it), and documented
  `schedule_repair_materialize_sandbox`'s bool return.

### New regression coverage

Four tests added, each verified to fail without its fix:

- fresh signup with **no** pre-committed row, four polls through the real
  gateway path on rolled-back sessions: the row is committed and exactly **one**
  repair is scheduled (fails `assert None is not None` without B1a)
- a sandbox row deleted between scheduling and running: `report_critical` never
  fires and the claim is still released (fails with the paged
  `cloud_sandbox_missing` without B1b)
- any other materialization error still pages
- a synchronous spawn failure releases the claim, so the next access can
  schedule immediately

### Verification

`tests/integration/test_cloud_sandbox_cold_access_repair.py` **12 passed**;
gateway + redis-lock + billing-gate unit/integration files **28 passed**;
`tests/integration -k cloud` **303 passed**; `tests/unit -k "cloud or sandbox or
materializ or workspace"` **281 passed**; `ruff check`/`format --check` on
`proliferate tests` clean; `scripts/check_docs.py` and
`scripts/check_max_lines.py` pass (no allowlist change needed).

New head: `d7c4a4a705b1e49b14b43f9ff6f785f3a3892e0f`. Still a draft.


pablonyx:
Client side of the typed 409 landed in a79134f1c: `cloud_sandbox_runtime_not_ready` is now classified as a not-ready error (routes to the existing pending/connecting panel via the selection seam) and gets its own retry budget — 45 retries × 2s (90s ceiling) covering the 30–60s background reprovision, while `workspace_not_ready`/transient errors keep the generic 8×750ms budget. Applies to both `retryCloudWorkspaceRequest` and the react-query connection options. 8 new unit tests; full product-client suite green (633 files / 3834 tests).

PR #1509, opened 2026-07-26
feat(cloud): paired worktree reclaims on archive/delete + marked-lost on provider loss

pablonyx:
Fix-stack PR C7 (cloud-sandbox program; contracts: `sandbox-content.md` reclaim/dangle gaps, `cloud-workspace.md` lost state, `sandbox-lifecycle.md` provider loss).

**Paired reclaim:** archiving or deleting a repository-worktree workspace now fires a best-effort after-commit retire of its runtime worktree (`run_after_commit`, same pattern as sandbox destroy; new `retire_runtime_workspace` AnyHarness binding hitting `POST /workspaces/{id}/retire`). Failures log and never surface or block the row operation — a paused/dead sandbox is fine; retention (enabled by C2/#1500) sweeps later. Skipped when there's no stamped runtime id, no live sandbox, or for scratch workspaces.

**Marked-lost:** provider loss — the killed webhook, provider-missing detach (connect + resume-acceptance + materialization-failure paths), and sandbox destroy — now stamps a new persisted `lost_at` on the bound `cloud_workspace` rows, scoped by their managed-cloud materialization binding (legacy fallback: runtime-backed rows with no binding). Derived status gains `lost` (server enum + client union); the status panel renders a lost mode with a delete path; a lost workspace is never silently rematerialized (replacement VM starts empty by design; recovery is delete-and-recreate, now visible instead of an opaque runtime error).

**Deliberately not here:** repo-environment clone reclaim — AnyHarness has no clone-delete primitive; the spec gap item is split so the remaining half tracks that primitive.

Spec gap strikes: sandbox-content.md (dangle→lost struck; reclaim item split), cloud-workspace.md (lost-state gap struck).

**Tests:** retire-after-commit unit trio (fires with stamped id / skips without / failure swallowed), mark-lost stamping incl. binding + legacy scoping and webhook wiring, lost status derivation/payload/client presentation, lost-rematerialization rejection. Full server unit suite **1180 passed**; product-client **3827 passed**; migration verified on a scratch DB.

Note for merge ordering: touches the same max-lines allowlist entries as C1 (#1501) — trivial conflict, resolve to the larger observed values.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1533, opened 2026-07-26
fix(server): rename the pkey the webhook_event_receipt table rename left behind

pablonyx:
## Summary

Follow-up to a bug found during B4 (#1521)'s review: `op.rename_table` renames only the table. Migration `9a0b1c2d3e4f` (`stripe cloud billing foundation`) renamed `sandbox_event_receipt` → `webhook_event_receipt`, but never renamed the PRIMARY KEY constraint — Postgres kept it (and its backing index, since Postgres keeps the two names identical) on `sandbox_event_receipt_pkey`.

Result: a database that ran `9a0b1c2d3e4f` diverges from `Base.metadata.create_all` forever — a migrated prod database carries `sandbox_event_receipt_pkey` on `webhook_event_receipt`, while a fresh `create_all` database (e.g. a test schema) gets `webhook_event_receipt_pkey`. Any later code assuming the create_all name (e.g. `drop_constraint("webhook_event_receipt_pkey")`) would pass against a fresh test schema and fail against prod.

B4 hit and fixed exactly this bug class for its own table rename (`agent_catalog_snapshot` → `agent_model_snapshot`, commit 229adcf73 on `agents/b4-snapshot-rekey`), and this migration follows that same precedent: an idempotent, inspector-guarded `ALTER TABLE ... RENAME CONSTRAINT`, with a symmetric downgrade. No FK or sequence needed renaming here — the table has no FK-typed columns, and `id`'s default is a client-side generated UUID, not a Postgres sequence — so the PRIMARY KEY (and its backing index) is the only leftover artifact.

Confirmed against the local dev database (read-only inspection) that the divergence is real today:
```
Indexes:
    "sandbox_event_receipt_pkey" PRIMARY KEY, btree (id)
    "ix_webhook_event_receipt_status" btree (status)
    "uq_webhook_event_receipt_provider_event_id" UNIQUE, btree (provider, event_id)
```

## Evidence — execution-based proof against local Postgres (scratch DBs, dev DB never touched)

**Scenario 1 — simulated migrated database** (upgrade to the previous head, then to this migration):
```
[migrated] pkey BEFORE our migration: 'sandbox_event_receipt_pkey'
[migrated] pkey AFTER upgrade to head: 'webhook_event_receipt_pkey'
[migrated] pkey AFTER downgrade to 35fa0038d703: 'sandbox_event_receipt_pkey'
[migrated] pkey AFTER re-upgrade: 'webhook_event_receipt_pkey'
[migrated] pkey AFTER redundant upgrade (no-op check): 'webhook_event_receipt_pkey'
```

**Scenario 2 — fresh `create_all`-shaped database** (this migration's `upgrade()` run directly against it):
```
[fresh create_all] pkey as created: 'webhook_event_receipt_pkey'
[fresh create_all] pkey AFTER running our migration's upgrade(): 'webhook_event_receipt_pkey'
```
No error, true no-op — confirming the migration is safe to run against a database that never had the old name.

Both scratch databases were dropped after; the local `proliferate` dev database was only ever read (never migrated/dropped) and still shows the pre-fix name, as expected.

## Test plan

- [x] Added `server/tests/integration/test_webhook_event_receipt_pkey_migration.py`, mirroring B4's pattern: an up/down round-trip test, plus a `create_all`-comparison test (the assertion class that actually catches a leftover pkey — a shape-only/column-only test cannot see it). Both pass.
- [x] `tests/integration/test_schema_migrations.py` (full chain to head, including `test_alembic_upgrade_creates_current_schema`) — 6 passed.
- [x] Webhook-adjacent suites (`test_cloud_webhook_recovery_races.py`, `test_cloud_webhook_service.py`, `test_stripe_webhooks.py`) — 41 passed, unaffected.
- [x] `ruff format --check` + `ruff check` on touched files — clean. (mypy is scoped to `proliferate/` only per the Makefile; alembic/tests aren't covered.)
- [x] `python3 scripts/check_max_lines.py` — passed.

### DIFF
--- a/server/alembic/versions/ab5316095737_webhook_event_receipt_pkey_rename.py
+++ b/server/alembic/versions/ab5316095737_webhook_event_receipt_pkey_rename.py
@@ -1,0 +1,92 @@
+"""webhook event receipt pkey rename
+
+Revision ID: ab5316095737
+Revises: 35fa0038d703
+Create Date: 2026-07-26 00:00:00.000000
+
+``9a0b1c2d3e4f_stripe_cloud_billing_foundation.py`` renamed
+``sandbox_event_receipt`` to ``webhook_event_receipt`` via ``op.rename_table``,
+which renames only the table. Postgres left the PRIMARY KEY constraint (and its
+backing index, since Postgres keeps the two names identical) on
+``sandbox_event_receipt_pkey``. A database that ran that migration therefore
+diverges from ``Base.metadata.create_all``, which names the constraint
+``webhook_event_receipt_pkey`` on a fresh install — forever, since nothing
+after 9a0b1c2d3e4f touched it.
+
+This is the same bug class B4 (``b7c1e4d9f082_agent_model_snapshot_rekey.py``)
+fixed for its own rename, and this migration follows that precedent: rename
+the constraint under its old name if present, do nothing if the name is
+already correct (fresh create_all, or a database that already ran this
+migration), and never touch a database that already has the new name.
+
+No FOREIGN KEY constraint exists on this table (no FK-typed columns), and
+``id`` has no server-side sequence (UUID default generated client-side), so
+the PRIMARY KEY is the only artifact the table rename left behind.
+
+Downgrade renames the constraint back, which is symmetric with the pre-B4-fix
+downgrade in 9a0b1c2d3e4f (that downgrade path recreates
+``sandbox_event_receipt`` and its ``_pkey`` from scratch via
+``op.rename_table``, so this migration's downgrade only has work to do when
+run against a database still on ``webhook_event_receipt``).
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "ab5316095737"
+down_revision: str | Sequence[str] | None = "35fa0038d703"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+_TABLE = "webhook_event_receipt"
+_OLD_PKEY = "sandbox_event_receipt_pkey"
+_NEW_PKEY = "webhook_event_receipt_pkey"
+
+
+def _has_table(table_name: str) -> bool:
+    bind = op.get_bind()
+    inspector = sa.inspect(bind)
+    return table_name in inspector.get_table_names()
+
+
+def _pkey_name(table_name: str) -> str | None:
+    bind = op.get_bind()
+    inspector = sa.inspect(bind)
+    name = inspector.get_pk_constraint(table_name).get("name")
+    return str(name) if name else None
+
+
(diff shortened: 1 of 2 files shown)


pablonyx:
## Adversarial review complete — READY, with one merge-order rule

Independent opus review re-executed everything on scratch DBs (both `pg_constraint` and `pg_indexes`, which the PR's own test doesn't check): rename applies, backing index renames with the constraint, redundant upgrade is a no-op, downgrade is symmetric, fresh `create_all` schemas are untouched. Inventory claim confirmed — no FKs, sequences, triggers, or CHECKs on the table; the pkey really was the only rename leftover. Ruff/max-lines/migration-heads gates clean; the PR's 2 tests plus the 6 schema-migration integration tests pass.

### ⚠️ Merge order: land this PR LAST among the open migration PRs

This PR and the B2→B4 chain (and #1509/#1516) all parent on `35fa0038d703`. Whichever lands second forks the alembic history → `check_migration_heads.py` goes red in CI on the next PR (not a silent prod break, but a stuck train). Verified by simulation:

- Both landed unreconciled → 2 heads; `alembic upgrade head` and every app-boot `validate_database_schema` call fail.
- **Cheapest reconciliation: this PR re-parents.** It's a 92-line leaf with no dependents — after the B-chain lands, change `down_revision` to the then-current head (`b7c1e4d9f082` if B4 is last) **and update `_DOWN_REVISION` in `test_webhook_event_receipt_pkey_migration.py:32` to match**. Verified: reconciled chain upgrades clean, both pkey fixes hold, heads check reports exactly one head.
- Re-parenting the B-chain onto this PR instead does NOT work cleanly (orphans `28adf1a9e376`, shared by 4 PRs).

### Review findings (non-blocking)

1. **Idempotency-guard test gap (medium):** `command.upgrade` on an already-stamped DB never re-invokes the script, so the "re-run is a no-op" test asserts nothing about the guard. Mutation proof: deleting the guard still passes all 8 tests; rewinding only the version marker exposes it (`UndefinedObjectError` without the guard). The guard itself is correct and load-bearing for the `stamp_schema_head` path. Tighten later by rewinding the stamp instead of re-upgrading.
2. **Same latent bug on another table (medium, out of scope):** `cloud_sandbox_secret_materialization` (renamed in `d4e6f8a0b2c4`) still carries `managed_sandbox_secret_materialization_pkey`, a leftover FK, and two leftover `ix_managed_...` indexes at head. Same fix shape as this PR; these two tables are the complete set of `rename_table` sites. Follow-up candidate.
3. **Residual divergence on this table (low):** `uq_webhook_event_receipt_provider_event_id` is a UNIQUE constraint on fresh schemas but only a unique index on migrated DBs (from `9a0b1c2d3e4f`). Benign — the app's `ON CONFLICT` uses `index_elements`, which resolves either way. Noted so nobody assumes the table is now fully reconciled.

Downgrade floor (`f8b9c0d1e2f4` NotImplementedError) is consistent with this migration all the way down. Ready for Pablo's review.

PR #1536, opened 2026-07-26
feat(agent-auth): render typed provider-config sources into harness env vars (D3 python arm)

pablonyx:
## Summary

Python arm of D3 ("typed provider-config apply"), implemented per the frozen
spec of record at `codex/d3-apply-brief.md`. Stacks on `agents/d1-provider-configs`
(#1530), which stacks on the B-chain; ships in that train. The Rust render arm
(codex's config.toml injection path + the codex×azure pending-flip) is out of
scope here — that's D3-rust, gated on Gate 4.

This PR resolves a generic vault provider-config entry (region/bearerToken/
endpoint/deployment/apiKey) into the harness's REAL, already-named env vars
**before** the source reaches Rust. The wire-level `provider_config` kind
rides along on `state.json` only so Rust can pick a render arm (plain env-set
vs. codex's config.toml path) — never to rename a field
(`specs/codebase/platforms/product/agent-auth.md` typed-provider-configs /
wire-contract sections; brief §2).

**Review status: fix cycle 1 landed.** An adversarial review found 3 gaps
(1 integration-coverage blocker, 1 misleading fixture-reconciliation writeup,
1 inaccurate PR-body claim about Rust) plus 2 cheap hardening items; all 5
are fixed below. See "Fix cycle 1" for the itemized resolution.

## What changed

- `server/proliferate/constants/agent_gateway.py`: added
  `AGENT_AUTH_SOURCE_PROVIDER_CONFIG = "provider_config"`, documented as
  WIRE-only (never a DB `source_kind`; decided at render time from the
  referenced `AgentApiKey` row's vault `kind`).
- `server/proliferate/server/cloud/materialization/materialize/agent_auth.py`:
  - `_render_source` now branches to a new `_render_provider_config_source`
    when the selection's `api_key_id` resolves to a typed vault entry.
  - `_translate_provider_config_env(harness_kind, config_kind, fields)`
    implements the brief's §4.2 per-(harness, config_kind) translation table
    for all 6 combinations: claude×aws_bedrock, claude×azure_openai (Foundry,
    **unverified**), codex×aws_bedrock, codex×azure_openai (always `None`,
    pending D3-rust's Gate-4 flip), opencode×aws_bedrock,
    opencode×azure_openai (live-test-proven).
  - `_hostname_first_label` derives the Azure resource name from the
    endpoint's hostname first label (e.g.
    `https://proliferate-gw-aoai.openai.azure.com` → `proliferate-gw-aoai`) —
[...]

PR #1541, opened 2026-07-27
feat(runtime): render typed provider configs at launch (D3-rust)

pablonyx:
## Summary

The render-plane half of Track D's typed provider configurations
("use my own Bedrock/Azure account"). Builds on `agents/a8-poke-wiring`.

- `state.rs`: new `SOURCE_KIND_PROVIDER_CONFIG` const; `AuthSource` gains
  optional `config_kind`/`env` (BTreeMap) fields. `STATE_VERSION` unchanged (2).
- `profile.rs`: `ResolvedSource::ProviderConfig(ProviderConfigProfile)` +
  a resolve arm — empty `env` or missing `config_kind` is a typed
  `SelectionIncomplete`, never a silent native fallback.
- `render.rs` — `render_provider_config`:
  - claude/opencode: fully generic env-set loop over `profile.env`, no
    per-`config_kind` branch (including mode-switch flags like
    `CLAUDE_CODE_USE_BEDROCK`/`CLAUDE_CODE_USE_FOUNDRY`).
  - codex × `aws_bedrock`: a real `config.toml` via codex's **built-in**
    `amazon-bedrock` provider (`model_provider = "amazon-bedrock"` + a
    catalog-resolved `model`, **no** `[model_providers.*]` table — codex
    needs no custom provider block for this one). Missing catalog default
    is a typed error, no fallback.
  - codex × `azure_openai`: a `config.toml` injection mirroring the gateway
    recipe's `env_key` pattern — **built but registry-`pending`, unverified**
    (see below).
  - cursor/grok/unknown `config_kind` → typed `UnsupportedRoute`.
  - Runs **before** `sanitize_claude_if_routed` so a provider_config route's
    own rerouting flag is what sanitize sees as "explicitly set" and keeps.

## Wire contract

Per agent-auth.md: Python resolves the vault's generic fields into
already-harness-correct env var names before the document reaches Rust.
`config_kind` is consumed **only** to pick which render arm runs — never
to rename a field. Rust never invents an env var name.

## Bug found and fixed by the new tests

`sanitize_claude_ambient` unconditionally removed
`CLAUDE_CODE_USE_BEDROCK`/`_VERTEX`/`_FOUNDRY`/`AWS_BEARER_TOKEN_BEDROCK`
regardless of what the current render set. That's fine for every existing
route (none of them set those flags), but a `provider_config` × `aws_bedrock`
source **does** set `CLAUDE_CODE_USE_BEDROCK` — sanitize would have stripped
it right back out immediately after render composed it, seconds after
being written. Fixed to key off "did this render set it", matching the
existing Anthropic-selector removal loop. Regression test:
`claude_provider_config_bedrock_flag_survives_sanitization`.

## Catalog-data flag (unverified, disclosed per instruction)

Codex's `session.defaults["bedrock"]` in `catalog.json` is
`"openai.gpt-5.5"` — an OpenAI-style model id under a Bedrock context.
This is wired through `GatewayModelPlan.bedrock_default_model` **as-is**;
I did not correct or second-guess the catalog value. Flagging it here
since it looks like it may be a copy/paste artifact from codex's other
context defaults, but fixing catalog data is out of this PR's scope.

## codex × azure_openai: why it stays `pending`

The render arm is built and unit-tested directly (bypassing the full
`state.json` pipeline, since no real selection can reach it while the
registry kind is `pending`). It is **not live-verified** — nobody has
run codex against real Azure OpenAI with this config. Per the brief,
the registry's `pending: true` on codex×azure_openai is **unchanged**;
only `pendingReason` is updated to "config.toml injection built,
live-unverified" so a future flip is a one-line registry change once
Gate 4 passes. `AZURE_OPENAI_ENDPOINT`/`AZURE_OPENAI_DEPLOYMENT` are this
arm's ASSUMED env var names — the registry today declares only
`AZURE_OPENAI_API_KEY`; if Gate 4 wires a different name, this arm's
constants need to move with it.

## Fixture coordination (agents/d3-provider-apply-python, #1536)

Added a third opencode source (`provider_config` × `aws_bedrock`) to
`fixtures/contracts/agent-auth-state/v2.json`, verified **byte-identical**
(diffed as parsed JSON) against `agents/d3-provider-apply-python` @
`5ebae60d2` (the branch's post-force-push head — confirmed via
`git fetch` + diff that the force-push was content-preserving for this
fixture). All a8-lineage-specific fixture facts are preserved: 5 harnesses
(incl. cursor + grok), `issuing_server_origin`, grok's `sources: []`
fail-closed proof.

**Known second-lander conflict:** D3-python's branch lives on a
*different* fixture lineage (`agents/d1-provider-configs`, forked from
`agents/b4-snapshot-rekey`, which never merged the A4/A8 fail-closed
work) — it diverges from this one on harness count/membership,
`issuing_server_origin`, empty-sources semantics, and whether gateway
keys are per-harness yet (documented in that branch's own README). This
PR only makes the ONE shared new fact — the shape of a `provider_config`
source — identical between the two lineages; reconciling the rest is
explicitly out of scope here (also noted in the fixture README).

Since this branch's own Python producer (`agent_auth.py` on the
`agents/a8-poke-wiring` lineage) cannot yet emit `provider_config`
sources, I added a disclosed known-gap tripwire test
(`test_this_renderer_does_not_produce_provider_config_yet`) mirroring the
existing `test_per_harness_gateway_keys_are_not_produced_yet` B3 pattern,
and adjusted the order/shape assertions to exclude the not-yet-producible
third source rather than asserting a false positive.

## Registry/catalog

Added `providerConfig[]` to `catalogs/agents/registry.json` for
claude/codex/opencode, mirroring `agents/d1-provider-configs`'s shape
exactly (env var names, flag vocabulary, codex-azure `pending`).
Bumped `registryVersion` 2026-07-17.2 → .4 (D1 already claimed `.3` on
its own branch) and `catalog.json`'s `probedAgainst.registryVersion` to
match; synced `scripts/agent-catalog/catalog.draft.json` (must be
byte-identical to `catalog.json` per the validator).

## Spec

Narrowed (did not delete) agent-auth.md's "Typed provider configurations
do not exist" Current-gaps bullet to "...cannot be applied at launch",
describing what this PR actually closed (the runtime/render half) and
what's still open (the vault/write side, on a separate corridor; codex-
azure's live-verification gate). Checked D3-python's own spec-branch
wording for the same region — that branch restructured the bullet
entirely as "Selections cannot yet reference a typed vault entry" with
write-path detail; the two edits target logically-adjacent but
textually-different bullets, so I don't expect a literal merge conflict,
but whoever lands second should sanity-check the combined Current-gaps
section reads coherently.

## Test plan

All commands run in this worktree, pgrep-guarded (never more than one
cargo/rustc build at a time on this machine):

- `cargo test -p anyharness-lib --lib domains::agents::route_auth` —
  **117 passed, 0 failed** (17 new: the provider_config render matrix
  in the new `provider_config_render_tests.rs` split file — per-
  (harness,kind) coverage, exact Bedrock/Azure `config.toml` bytes,
  the sanitization regression above, unknown-kind, empty-env,
  cursor/grok rejection).
- `cargo test -p anyharness-lib` (full crate) — **1474 passed, 0
  failed, 1 pre-existing ignored**.
- `node scripts/validate-agent-catalog.mjs` — `agent catalog OK`.
- `node --test scripts/validate-agent-catalog.test.mjs` — 4/4 passed.
- `node --test scripts/agent-catalog/build-catalog.test.mjs` — 7/7 passed.
- `python3 scripts/check_max_lines.py` — passed (allowlisted
  `render.rs` 749→759 and added a new entry for `launch_facts.rs` 643,
  both with reasons naming this PR).
- `python3 scripts/check_docs.py` — passed (227 files).
- `server/.venv/bin/python -m pytest tests/unit/test_agent_auth_state_contract_fixture.py tests/unit/test_agent_auth_materialization.py`
  — **23 passed** (2 of the 5 contract-fixture tests needed updating
  for the new third source; added the known-gap tripwire above).

**Unverified** (explicitly, per the honesty rule): codex×azure_openai
against a real Azure OpenAI deployment — this is exactly why the
registry keeps that kind `pending`.

## Deploy note

Stacks on A8 (`agents/a8-poke-wiring`, #1522). No migration, no
`STATE_VERSION` bump. The new render arm is dead code from a real
user's perspective until the vault/write-side PR (D1/D3-python
corridor) merges and lets a selection reference a typed vault entry —
this PR only makes the runtime *able* to render one when it shows up.


pablonyx:
## Adversarial review — PR #1541 (D3-rust): **FIX-CYCLE**

Reviewed at head `f50494eb7` (base `agents/a8-poke-wiring`). Targeted tests run and **green**: `cargo test -p anyharness-lib route_auth` → **121 passed / 0 failed** (includes all 13 new `provider_config_render::*` cases). `python3 scripts/check_max_lines.py` → pass. `node scripts/validate-agent-catalog.mjs` → pass (`2026-07-17.6`, 5 agents). `catalog.draft.json` is byte-identical to `catalog.json`.

**Parity with the python arm (#1536) is clean.** Because the claude/opencode arms are a fully generic `.set()` loop over `profile.env` (`render.rs:640-650`), the env-var names Rust emits are exactly the keys `_translate_provider_config_env` produces — `CLAUDE_CODE_USE_BEDROCK`/`AWS_BEARER_TOKEN_BEDROCK`/`AWS_REGION`, `CLAUDE_CODE_USE_FOUNDRY`/`ANTHROPIC_FOUNDRY_{RESOURCE,BASE_URL,API_KEY}`, `AZURE_API_KEY`/`AZURE_RESOURCE_NAME` — and all of those match `registry.json`'s `providerConfig[].envVars` declarations. No rename, no divergence. The `sanitize_claude_ambient` ordering fix and its regression test are the right call, and the `bedrock_default_model` poison-the-other-defaults resolver in the tests is genuinely load-bearing. Secret hygiene on the new struct is sound: no `Serialize` on `ProviderConfigProfile`, `ProbeAuthMaterial`'s hand-written `Debug` still redacts `scoped_profile`, and the probe arm digests values rather than carrying them.

Two code findings block, plus the two known CI items.

---

## Blocking

### B1 — `sanitize_claude_ambient`'s new "keep what I set" exemption is keyed too broadly; it re-opens the exact Bedrock-reroute hole the function exists to close
`anyharness/crates/anyharness-lib/src/domains/agents/route_auth/render.rs:269`

```rust
for key in ["CLAUDE_CODE_USE_BEDROCK", "CLAUDE_CODE_USE_VERTEX",
            "CLAUDE_CODE_USE_FOUNDRY", "AWS_BEARER_TOKEN_BEDROCK"] {
    if !rendered.set.contains_key(key) {   // <-- ANY source's set, not just provider_config's
        rendered.remove(key);
    }
}
```

The exemption asks "did *this render* set it", but the `api_key` arm at `render.rs:140-143` sets an **arbitrary, user-chosen** env var name:

```rust
ResolvedSource::ApiKey(profile) => {
    rendered.set(&profile.env_var_name, &profile.value);
}
```

and the only server-side gate on that name is a shape regex — `selection_rules.py:31`: `^[A-Z][A-Z0-9_]{0,127}$`, with no denylist and no registry allowlist (`grep -c provider_config selection_rules.py` = 0; no `CLAUDE_CODE_USE*` mention anywhere in `agent_gateway/`).

**Concrete failure scenario.** A claude selection with a single enabled `api_key` row whose `env_var_name` is `CLAUDE_CODE_USE_BEDROCK` and value `1` (legal today — matches the regex, single-source cardinality satisfied). Render sets it; sanitize now *keeps* it because `rendered.set` contains the key. `merge_spawn_env` (`process.rs:26-38`) lets the set survive since there is no matching removal, and `Command::env_remove` is likewise skipped. The Claude CLI launches rerouted to Bedrock with no Bedrock credential the user selected — precisely the "billing an account they did not select, with the key they did select sitting unused in the env" case this function's own doc calls out at `render.rs:116-119`. `AWS_BEARER_TOKEN_BEDROCK` is the same shape: before this PR it was unconditionally stripped, and an `api_key` row named that was neutralized; after this PR it survives. **This is a behavior regression on the pre-existing `api_key` path, not just a Track D question** — the removal was unconditional before, and this PR is what makes it conditional.

**Fix:** scope the exemption to keys the *provider_config* arm composed, not to `rendered.set` at large. E.g. have `render_provider_config` record its keys (a `BTreeSet<String>` threaded into `sanitize_claude_if_routed`, or a field on `RenderedRouteAuth` that the sanitizer consults), and keep the unconditional `remove` for anything an `api_key`/gateway source happened to name. A regression test asserting `api_key(env_var_name="CLAUDE_CODE_USE_BEDROCK")` is still removed for claude would fail today and is the mutation-proof this change is missing.

### B2 — `registry_flag_vars` reads only `auth.slots[]`, so `CLAUDE_CODE_USE_FOUNDRY` is classified as a *secret* — contradicting the doc comment three lines above it, with zero new tests on the arm
`anyharness/crates/anyharness-lib/src/domains/agents/auth/launch_facts.rs:143-158` (consumed at `:111-122`)

```rust
fn registry_flag_vars(agent_kind: &str) -> BTreeSet<String> {
    bundled_agent_registry_document().agents.iter()
        .find(|agent| agent.kind == agent_kind)
        .map(|agent| agent.auth.slots.iter()
            .flat_map(|slot| &slot.env_vars)          // <-- slots only
            .filter(|env_var| env_var.kind() == AgentRegistryEnvVarKind::Flag)
            ...
```

claude's `auth.slots[].envVars` in the registry this PR ships declares exactly one flag: `{"name": "CLAUDE_CODE_USE_BEDROCK", "kind": "flag"}` (`registry.json:52`). `CLAUDE_CODE_USE_FOUNDRY` is declared `kind: flag` **only** under the new `providerConfig[]` block (`registry.json:130`) — which `registry_flag_vars` never reads. Because `AgentRegistryAuthSlotEnvVar::kind()` defaults to `Secret` (`registry/schema.rs:198-208`), a claude × `azure_openai` provider_config source therefore emits `CredentialFact::Env { var: "CLAUDE_CODE_USE_FOUNDRY" }` — presence-only, value dropped — not the `EnvFlag { var, value }` the doc comment at `launch_facts.rs:89-93` explicitly promises by name:

> `[CredentialFact::EnvFlag]` (with its real value) for a registry-declared flag var (e.g. `CLAUDE_CODE_USE_BEDROCK`, `CLAUDE_CODE_USE_FOUNDRY` …)

**Concrete failure scenario.** The moment a catalog `authContext` for Foundry lands with `signals: {"envFlag": "CLAUDE_CODE_USE_FOUNDRY=1"}` — exactly the shape claude's existing `bedrock` context already uses (`catalog.json:46-49`) — `context::classify` requires an exact var+value `EnvFlag` fact (`auth/context.rs:156-170`). It will get a valueless `Env` fact and never match, so a claude×Foundry launch classifies into the wrong auth context and its model list resolves off the wrong `session.defaults` entry. The bug is latent only because no Foundry context exists in `catalog.json` yet; the mode-switch flag the render plane composes is the whole point of the arm, and the classifier cannot see it.

Aggravating: **`launch_facts.rs` gained zero new tests in this PR** (`git diff HEAD~1 -- launch_facts.rs | grep -c '#\[test\]'` → 0), despite gaining a new fact-classification arm and a new helper function. The existing tests only cover ambient/composed paths, so nothing pins the provider_config classification either way — the divergence between the code and its own doc comment survived because no test asserts it. Please add a test that a claude provider_config × `azure_openai` source produces `EnvFlag { var: "CLAUDE_CODE_USE_FOUNDRY", value: "1" }` and that `ANTHROPIC_FOUNDRY_API_KEY` produces a presence-only `Env` (the secrets rule the doc claims).

**Fix:** either union `providerConfig[].envVars`' flag-kind names into `registry_flag_vars` (needs the Rust registry schema to actually deserialize `providerConfig` — it currently does not; `registry/schema.rs` has no such field, so the block this PR adds is inert to Rust today), or promote `CLAUDE_CODE_USE_FOUNDRY` into claude's `auth.slots[].envVars` alongside `CLAUDE_CODE_USE_BEDROCK` and correct the doc to say where the vocabulary comes from.

### B3 — CI: `CI/CD config checks` is red because `catalogVersion` is not bumped (pre-confirmed)
`catalogs/agents/catalog.json:3` still says `"catalogVersion": "2026-07-17.6"` while the file changed (`probedAgainst.registryVersion` → `2026-07-17.4`). #1542 has already claimed **2026-07-17.7**, so this PR's bump must be **`2026-07-17.8`** — in `catalogs/agents/catalog.json` and the byte-identical `scripts/agent-catalog/catalog.draft.json` together.

### B4 — CI: `test` lane red is a cancelled 1h15m runner timeout, not a code failure (pre-confirmed)
No fix needed; a re-run has already been kicked off. Do not chase this as a regression.

---

## Non-blocking

- **`AZURE_OPENAI_ENDPOINT` / `AZURE_OPENAI_DEPLOYMENT` are invented names, which sits awkwardly against this PR's own stated wire contract** (`render.rs:620-630`). The PR body is candid that these are assumptions and that the registry declares only `AZURE_OPENAI_API_KEY` for codex×`azure_openai`, and the brief sanctioned building the arm behind `pending` — so not blocking. But note the flip will **not** be one line: #1536's python arm returns `None` for codex×`azure_openai` (`agent_auth.py`, "structurally excluded" branch), so no producer will ever emit these three keys. Whoever scopes Gate 4 must land the registry names, the python table row, and these constants in one change. Worth a `TODO(gate-4)` breadcrumb tying the three together.
- **codex's `session.defaults["bedrock"]` is `"openai.gpt-5.5"`** (`catalog.json:2148`, curated at `scripts/agent-catalog/build-catalog.mjs:226`) while `observedDefaults.bedrock` is `"openai.gpt-oss-120b"` (`catalog.json:2152`). Agreed this is out of scope, and correctly wired through as-is. Flagging that the new test at `provider_config_render_tests.rs:247` asserts `openai.gpt-oss-120b` — the *observed*, not the *curated* value — so it will not notice if the curated value is the copy/paste artifact you suspect. Consider a separate follow-up issue against the catalog data.
- **No key validation on `profile.env`.** `resolve_source` (`profile.rs:150-165`) accepts any string key and `render_provider_config` sets all of them verbatim. `selection_rules.py` has no provider_config awareness at all, so unlike `api_key` there is no `^[A-Z][A-Z0-9_]{0,127}$` gate on this path. The state file is 0600 and workspace-scoped so this is low-risk today, but a `PATH`/`LD_PRELOAD`-shaped key would be injected straight into the spawned CLI's env. The strongest form of the fix is to validate the map's keys against the registry's declared `providerConfig[].envVars` for `(harness, config_kind)` and fail closed — that also makes B2's registry read pull its weight.
- **Ambient AWS credentials are not sanitized on a bedrock launch.** `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/`AWS_SESSION_TOKEN`/`AWS_PROFILE` survive into the launch env alongside the injected `AWS_BEARER_TOKEN_BEDROCK`. I could not prove a precedence inversion (the bearer token is used as a direct auth header rather than through SigV4, so it likely wins), which is why this is not blocking — but it is the same class of hole `sanitize_claude_ambient` exists for, and the python arm does not address it either. Please make the call explicitly rather than by omission.
- **Codex's bedrock arm does not mirror the gateway arm's ambient-key removals.** `render_codex_gateway` drops `OPENAI_API_KEY`/`ANTHROPIC_API_KEY` (`render.rs:318-319`) so the provider config is authoritative; the bedrock arm drops nothing. Benign given `model_provider = "amazon-bedrock"` pins the provider, but the asymmetry between two arms in the same recipe table invites a future reader to conclude the removals were unnecessary.
- **`assert_eq!(rendered.set.len(), 2, ...)` at `provider_config_render_tests.rs` (opencode bedrock)** is the one brittle assertion in an otherwise well-mutated suite — it will fail for the right reason today, but for an unrelated reason the day anything else legitimately joins opencode's delta.
- **Gap-bullet discipline: satisfied.** `specs/codebase/platforms/product/agent-auth.md` narrows "Typed provider configurations do not exist" to "…cannot be applied at launch" with both remaining halves (vault/write side; codex×azure `pending`) named. Good.
- **Repo shape: clean.** Both allowlist entries (`render.rs` 759, `launch_facts.rs` 643) carry substantive justifications and `check_max_lines.py` passes.

pablonyx:
## Fix cycle 1 — PR #1541 (D3-rust), head `21956321c`

Base `agents/a8-poke-wiring`. All four blocking findings addressed; one commit per finding.

| Finding | Fix commit | What changed |
|---|---|---|
| **B1** — sanitize exemption keyed too broadly | `8f81955b2` | Exemption scoped to the keys the `provider_config` arm itself recorded |
| **B2** — `registry_flag_vars` reads only `auth.slots[]` | `634e14fee` | Unions `providerConfig[].envVars` flag-kind names; adds the Rust schema field so the block is no longer inert + the missing tests |
| **B3** — `catalogVersion` not bumped | `21956321c` | → `2026-07-17.8` in both files; allowlist refreshed |
| **B4** — `test` lane timeout | — | Confirmed resolved: the coordinator's re-run came back **green**. No code change. |

### B1 — `8f81955b2`
Confirmed as described, including that it is a regression on the **pre-existing `api_key` path**: the removal was unconditional before this PR, and this PR is what made it conditional.

`render_provider_config` now records every key it composes into an explicit `BTreeSet<String>`, threaded through `render_sources` → `sanitize_claude_if_routed` → `sanitize_claude_ambient`, and the rerouting-flag exemption keys off *that* set:

```rust
if !provider_config_keys.contains(key) {
    rendered.remove(key);
}
```

The `api_key` arm deliberately does **not** record (comment says so at the call site), so a user-chosen name stays subject to the strip list. Added `RenderedRouteAuth::set_recorded` so the recording cannot be forgotten in one arm — all four `provider_config` set-sites go through it.

Two tests, both **mutation-proven** (I reverted the predicate to `rendered.set.contains_key` and confirmed they fail, then restored):
- `claude_api_key_named_like_a_rerouting_flag_is_still_stripped` — an `api_key` row named `CLAUDE_CODE_USE_BEDROCK` is still removed; same for one named `AWS_BEARER_TOKEN_BEDROCK`.
- `a_claude_api_key_named_like_a_flag_is_stripped_even_beside_a_provider_config_source` — the scoping case: provider_config's own Bedrock flag survives *while* an `api_key`-named `CLAUDE_CODE_USE_FOUNDRY` in the same delta is stripped. A `rendered.set`-wide exemption keeps both and fails here.

The existing `claude_provider_config_bedrock_flag_survives_sanitization` still passes unchanged, so the legitimate Track D case is not regressed.

### B2 — `634e14fee`
Took the **first** of the two options you offered (union `providerConfig[]` into the vocabulary), which required the schema work you flagged: `registry/schema.rs` had no `providerConfig` field, so D1's block really was inert to Rust. Added `AgentRegistryProviderConfig { kind, label, env_vars, pending, pending_reason }`, reusing `AgentRegistryAuthSlotEnvVar` so a `flag`-kind entry classifies identically wherever it is declared. `registry_flag_vars` now chains slot vars with provider-config vars before the flag filter, and the doc comment says where the vocabulary comes from.

Chose the union over promoting `CLAUDE_CODE_USE_FOUNDRY` into `auth.slots[]` because the slots list is the *credential* vocabulary and the flag is a provider-config mode switch — duplicating it would put the same name in two places with no single owner. This also makes the registry read pull its weight, which is the precondition for your third non-blocking note (key validation against declared `providerConfig[].envVars`).

Tests live in a new `#[path]` submodule (`launch_facts_provider_config_tests.rs`) because `launch_facts.rs` sits at its max-lines allowlist ceiling — same pattern as the existing route-transition tests. Three tests:
- Foundry: `EnvFlag { var: "CLAUDE_CODE_USE_FOUNDRY", value: "1" }`, while `ANTHROPIC_FOUNDRY_API_KEY`/`_RESOURCE` produce presence-only `Env` and never an `EnvFlag` (the secrets rule the doc claims). Also asserts no `Route` fact.
- Bedrock: the slot-declared flag still classifies correctly (regression guard on the path that already worked).
- `registry_flag_vars_unions_slot_and_provider_config_declarations`: both halves contribute, secrets from either half stay out, a flag-less harness (`grok`) is empty, and an unknown kind is empty rather than a panic.

Mutation-proven the same way: reverting to a slots-only read fails the Foundry test and the union test (the Bedrock one correctly still passes).

### B3 — `21956321c`
`2026-07-17.8` (`.7` is #1542's) in `catalogs/agents/catalog.json` and the byte-identical `scripts/agent-catalog/catalog.draft.json`. Also refreshed both allowlist line counts this cycle moved and extended their justifications.

### Non-blocking notes
**Applied (doc-only, no behavior change, folded into `8f81955b2`):**
- `TODO(gate-4)` breadcrumb on the Azure constants naming all three things that must land together (registry names + drop `pending`; the python arm's codex×azure_openai table row, which returns `None` today; these constants) and stating explicitly that the flip is *not* one line.
- **Made the ambient-AWS-credential call explicitly** rather than by omission, per your request: a comment in `sanitize_claude_ambient` records that `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/`AWS_SESSION_TOKEN`/`AWS_PROFILE` are deliberately not removed — the bearer token is a direct auth header rather than SigV4 so there is no proven precedence inversion, and stripping the AWS credential chain would break unrelated inherited tooling. Flagged as revisit-if-observed.
- Noted the codex bedrock-vs-gateway removal asymmetry as intentional at the bedrock arm (`model_provider = "amazon-bedrock"` pins the provider outright, so an ambient OpenAI/Anthropic key has nothing to shadow), so a future reader does not read it as an oversight.

**Skipped, deliberately, one line each:**
- **Key validation against declared `providerConfig[].envVars`** — the strongest form, and now unblocked by B2's schema field, but it is a new fail-closed behavior on the render path rather than a fix to a finding. Out of scope for a fix cycle; worth its own PR.
- **codex `session.defaults["bedrock"]` = `"openai.gpt-5.5"` vs observed `openai.gpt-oss-120b`** — agreed catalog-data issue; the test asserting the observed value is left as-is since changing it would encode a guess about which value is correct. Belongs in a catalog follow-up.
- **`assert_eq!(rendered.set.len(), 2)` brittleness (opencode bedrock)** — left alone; it fails for the right reason today and loosening it would weaken the "stays generic" claim it exists to make.
- **`AZURE_OPENAI_ENDPOINT`/`DEPLOYMENT` invented names** — the breadcrumb above is the fix you asked for; renaming them requires the Gate 4 scoping the note itself says is a separate change.

### Verification (one batched cargo pass, build-slot checked)
- `cargo test -p anyharness-lib route_auth` → **139 passed / 0 failed** (was 121; +2 new render tests, and the launch_facts filter overlap accounts for the rest)
- `cargo test -p anyharness-lib launch_facts` → **16 passed / 0 failed** (was 13; +3 new)
- `cargo test -p anyharness-lib registry` → **31 passed / 0 failed** (schema round-trip incl. the new `providerConfig` field)
- `cargo test -p anyharness-lib domains::agents` → **457 passed / 0 failed** (whole-domain sweep — nothing else read the flag vocabulary or the sanitizer)
- `python3 scripts/check_max_lines.py` → pass
- `node scripts/validate-agent-catalog.mjs` → OK (`2026-07-17.8`, 5 agents); `check-version-discipline.mjs --base origin/agents/a8-poke-wiring` → OK; `node --test scripts/agent-catalog/*.test.mjs` → **25/25**
- `rustfmt --check` clean on the new file; the crate's pre-existing fmt-diff count is unchanged by this cycle (276 before → 276 after). No new clippy findings in any touched file. `Cargo.lock` untouched — no dependency change.

Leaving this as a draft; handing the verify pass and the ready flag back to the coordinator.

pablonyx:
Fix cycle 1 verified on `21956321c`: B1 sanitize exemption scoped to provider-config-rendered keys (+bidirectional regression tests), B2 providerConfig[] flag vars in registry_flag_vars (+schema field that was missing entirely), B3 catalogVersion 2026-07-17.8, B4 resolved by test-lane re-run. Full CI green. READY.

pablonyx:

---

PR #1530, opened 2026-07-26
feat(agent-auth): typed provider-config foundations (D1)

pablonyx:
## Scope

Typed provider-config foundations (Track D, R9 full scope: claude/codex/opencode × {aws_bedrock, azure_openai}). Storage + registry declaration only — nothing downstream of storage (selections, `state.json`, render recipes) consumes either yet. This PR is dormant: no user-visible behavior changes.

- **DB**: `agent_api_key` gains a `kind` column (migration `a1c2f4d6b8e0`, chained on B4's `b7c1e4d9f082` per the alembic-heads collision corridor in agents-impl-plan.md §3: B2 → B4 → D1). Every pre-existing row backfills to `'api_key'` (lossless — no typed data could have existed before this column did). CHECK constraint closes the vocabulary to `('api_key', 'aws_bedrock', 'azure_openai')`.
- **Store**: `create_agent_provider_config` / `get_agent_provider_config_decrypted`, kind-scoped symmetrically with the bare-secret path so neither fetch can resolve the other kind's row.
- **Server**: `POST /keys/provider-config` (D2's `ProviderConfigCreatorSubmit` request shape) + `AgentApiKeyResponse.kind`. `catalogs/service.py:supported_provider_config_kinds` reads registry.json the same way `read_agent_catalog` reads the catalog (mtime-cached, no network round trip) — the server-side half of the seam D2's client-side `getSupportedProviderConfigKinds` will read in a later PR.
- **Registry**: `registry.json` declares `providerConfig` for claude (`aws_bedrock` + `azure_openai`/Foundry env vocabulary), codex (`aws_bedrock` + `azure_openai`), opencode (`aws_bedrock` + `azure_openai`) — R9's full Track D scope. Rust schema/validation extended: closed kind vocabulary, no duplicate kinds/env vars per entry, non-empty label, plus a `pending`/`pendingReason` pair for a declaration whose envVars the pinned binary cannot yet consume.
[...]

PR #1544, opened 2026-07-27
refactor(server): split the agent-auth routes off /agent-gateway (S1)

pablonyx:
## Summary

URL-string-only route split (agent-auth.md §API surface / model-gateway.md §API surface). No Python module/file renames, no SDK filename changes — literals only, per coordinator ruling.

**Base rationale:** `agents/d3-provider-apply-python` @ `5ebae60d2` (verified head unchanged via `gh pr view 1536 --json headRefOid` before starting) — deepest branch in the B→D lineage carrying the current shape of `agent_gateway/{api,service,models}.py`, including D1/D3's `POST /keys/provider-config` route and the typed provider-config render path.

## Routes moved vs. stayed

Read both specs' target API-surface sections before touching code — they specify a **two-prefix split**, not a blanket rename:

**Moved `/v1/cloud/agent-gateway/` → `/v1/cloud/agent-auth/`** (agent-auth.md: "owns the user-facing auth relationship"):
- `GET/POST /keys`, `POST /keys/provider-config`, `DELETE /keys/{key_id}` — the vault
- `GET /selections`, `PUT /selections/{harness_kind}` — the selection model
- `GET /state` — the rendered `state.json` v2 document
- `GET/PUT /organizations/{organization_id}/agent-auth/policy`, `GET .../policy/violations` — org allow-lists

**Stayed at `/v1/cloud/agent-gateway/`** (model-gateway.md: "owns exactly the gateway-account relationship... nothing else"):
- `GET /capabilities`
- `GET /enrollment`

Implementation: `api.py` now exports three routers from one module — `router`/`organization_router` (prefix `/agent-auth`, tag `cloud-agent-auth`) for the moved routes, and a new `gateway_account_router` (prefix `/agent-gateway`, tag `cloud-agent-gateway`) for the two that stayed. `cloud/api.py` mounts all three. No test or consumer asserted the `cloud-agent-gateway` tag string literally (grepped first), so it was safe to split the tag along with the prefix.

Route count: **12 route operations across 10 paths** now split between the two prefixes — 10 operations / 8 paths on `/agent-auth` (including the 3 `organization_router` policy operations), 2 operations / 2 paths on `/agent-gateway` — verified directly against `app.openapi()`. The original plan's "16→11/12" note was pre-B4; B4 already deleted the 5 catalog routes onto their own `agent_models` module, so there was nothing left to subtract for that.

## Security-relevant fix + new test

`server/proliferate/main.py`'s whole-body redaction gate (`_redacts_entire_body`) matched `request.url.path.endswith("/agent-gateway/keys")` — a hardcoded route suffix gating whether a malformed key-create body's echoed `value` gets redacted on a 422. Missing the update to `/agent-auth/keys` would have silently disabled that redaction on the live route, leaking raw key material back to the caller on validation failure.

**Fix-cycle addendum:** the initial version of this predicate (`endswith("/agent-auth/keys")`) missed `POST /agent-auth/keys/provider-config`, which carries a raw credential nested under `value` (e.g. `value.apiKey` for `azure_openai`) — and the `_SENSITIVE_INPUT_FRAGMENTS` field-name fallback doesn't recognize `apiKey` either, so a 422 on that route would have echoed the plaintext key. Changed the predicate to `"/agent-auth/keys" in request.url.path` (`main.py:155`) — a substring match over the whole key-vault subtree instead of an exact suffix — rather than widening `_SENSITIVE_INPUT_FRAGMENTS`, which would over-redact unrelated fields app-wide. Verified via `git grep` that no other route contains `/agent-auth/keys` as a substring, so the match stays scoped to this one endpoint family (and is `api_path_prefix`-agnostic, since it only tests a path suffix/substring, not from the root).

Added `server/tests/unit/test_agent_auth_body_redaction.py` (6 tests) — there was **zero direct CI signal on this predicate before this PR** (only an indirect assertion buried in an integration test). It pins:
- the predicate matches the current `/agent-auth/keys` route,
- the predicate does **not** match the old `/agent-gateway/keys` route (so a future revert regresses loudly, not silently),
- the predicate **does** match `/agent-auth/keys/provider-config` (fix-cycle addition — this was the missed leak path),
- the predicate doesn't over-match unrelated adjacent routes (`/selections/{harness}`),
- an end-to-end pass through the real `_validation_error_handler` never echoes a raw key value and does emit `[redacted]`, for both the plain key-create route and (fix-cycle addition) the provider-config route with a raw `value.apiKey`.

## Literal-consumer inventory (all updated)

- `server/proliferate/server/cloud/agent_gateway/api.py` — router prefixes/tags, module docstring
- `server/proliferate/server/cloud/api.py` — router imports/wiring (3 routers now)
- `server/proliferate/main.py:155` — redaction gate suffix
- `server/proliferate/server/cloud/agent_gateway/budget.py` — doc comment
- `server/proliferate/server/cloud/materialization/materialize/agent_auth.py` — doc comment
- `anyharness/crates/anyharness-lib/src/api/http/agent_auth.rs` — doc comment only (Rust — string edit, **no cargo build run**, verified with grep)
- `cloud/sdk/src/client/agent-gateway.ts` — 6 path literals updated (keys/selections/state/policy), 2 left as-is (capabilities/enrollment); filename kept per ruling
- `cloud/sdk/src/generated/openapi.ts` — regenerated via `make cloud-openapi` (uv-run FastAPI `app.openapi()`) → `npx openapi-typescript`; never hand-edited
- `apps/packages/product-client/.../agents-playground-cloud-client.ts` + `.test.tsx` — 6 literals moved, 2 left (capabilities/enrollment)
- `tests/release/src/{worlds/selfhost/gateway.ts, scenarios/local-world-smoke-1.ts, scenarios/local/config-session.ts, fixtures/{agent-auth,authenticated-actor,byok,invited-actor}.ts, config/env-manifest.ts}` — literals/doc-comments for moved routes only; `services/qualification-litellm.ts` references `/enrollment` only, no change needed
- `tests/intent/specs/billing/core.spec.ts` — org policy path literal
- `server/tests/{integration/test_agent_gateway_api.py, integration/test_agent_gateway_policy_api.py, unit/test_agent_gateway_key_transaction.py}` — literals updated, filenames kept
- `specs/codebase/platforms/product/{agent-auth.md, model-gateway.md}` — narrowed (not struck outright) both "Route prefix" Current-gaps bullets: the URL split landed, but the matching `api.py`/`service.py`/`models.py` **Python module** split (one module set per platform) is still pending — S1 was URL-string-only by design.

## Final grep sweep

`grep -rn "cloud/agent-gateway/" --include='*.py' --include='*.ts' --include='*.tsx' . | grep -v dist | grep -v node_modules | grep -v .venv` — every remaining hit is one of:
- `capabilities`/`enrollment` literals (legitimately stayed at `/agent-gateway`, per spec)
- the new redaction test's deliberate negative-assertion literal (`/agent-gateway/keys` — pinning what must NOT match anymore)
- historical "moved from" comments about the earlier B4 catalog re-key (already migrated, unrelated to this PR)
- `test_agent_models_api.py`'s catalog routes (B4's `agent_models` module, out of scope here)

No unexplained stragglers.

## Test plan (real commands + counts)

- `ruff check` + `ruff format --check` on all touched server files → all pass (one line-length + one formatting fix applied during implementation)
- `mypy` on touched files → same 5 pre-existing errors as on base `5ebae60d2` (confirmed via `git stash`/mypy diff), none introduced by this PR
- `pytest` (server venv, `USE_EXISTING_POSTGRES=1 USE_EXISTING_REDIS=1`, native brew postgres/redis):
  - `tests/integration/test_agent_gateway_api.py` — 26 passed (fix-cycle addition: `test_the_old_agent_gateway_routes_are_gone`, a 404 sweep over the 10 moved old paths + org-policy method variants, mirroring `test_agent_models_api.py`'s equivalent for the catalog routes)
  - `tests/integration/test_agent_gateway_policy_api.py` — 6 passed
  - `tests/unit/test_agent_gateway_key_transaction.py` — 1 passed
  - `tests/unit/test_agent_auth_body_redaction.py` — 6 passed (fix-cycle: flipped the provider-config assertion to over-match-covered and added a positive `value.apiKey` redaction test; see Security-relevant fix section)
  - `tests/unit/test_api_path_prefix.py`, `tests/unit/test_workflow_invocations.py`, `tests/unit/test_agent_gateway_domain.py` (adjacent middleware/domain regression check) — 38 passed
  - **Total: 77 passed, 0 failed**
- `python3 scripts/check_docs.py` → passed (226 Markdown files)
- `python3 scripts/check_max_lines.py` → passed
- Verified `app.openapi()` paths directly (before regen) — confirms the exact 9-route split landed as specified
- `cloud/sdk` — `npx tsc` build clean, `npx tsc --noEmit` typecheck clean
- Full internal build chain (`anyharness/sdk`, `anyharness/sdk-react`, `product-domain`, `cloud/sdk-react`, `design`, `ui`, `product-ui`, `product-surfaces`) — all `tsc` clean, no cargo touched
- `apps/packages/product-client` — `tsc --noEmit` typecheck clean (0 errors)
- `apps/packages/product-client` agents-playground vitest suite — 3 passed
- No Rust build run anywhere (`pgrep -x cargo`/`pgrep -x rustc` checked clean before starting; the one Rust file touched is a doc-comment-only string edit, verified by grep, not compilation)

## Deploy note

Stacks on the D-chain: B4 (merged into this lineage) → D1 `#1530` → D3-python `#1536` (this PR's base). The OpenAPI regen (`cloud/sdk/src/generated/openapi.ts`) means SDK consumers ride the same train as this PR — nothing downstream can consume the old `/agent-gateway/{keys,selections,state}` paths after merge; this is a hard cutover with no alias window, matching both specs' explicit "no alias windows, all consumers first-party (pre-launch ruling)" language.

One honest caveat on that cutover: already-installed desktop builds bundle the *old* SDK and will still call the old paths after this deploys. Those requests now 404 against the new server, so `use-local-auth-state-sync.ts` degrades silently — no state gets pushed, and the harness falls back to native auth — until the desktop app updates to a build carrying the new SDK. This is the spec-sanctioned pre-launch hard cutover (model-gateway.md:228-231), not a regression, but worth calling out explicitly rather than implying the old paths are simply unused.

## Deviations / assumptions

- **ASSUMPTION**: Tag rename (`cloud-agent-gateway` → `cloud-agent-auth` for moved routes) — grepped for literal tag-string consumers first, found none, so proceeded per the "DECISION POINT" in the brief.
- **ASSUMPTION**: Narrowed rather than struck both spec Current-gaps bullets, since the Python module split (one module set per platform, not the shared `agent_gateway` package) is explicitly out of scope for this URL-string-only PR.
- No blockers encountered.


pablonyx:
## Adversarial review — cycle 1 verdict: FIX REQUIRED

Route split vs both specs, consumer sweep (all languages + dynamic construction), OpenAPI regen integrity, spec narrowing, tag rename, base/stacking: **CONFIRMED**. Three fixes + two nits:

1. **Redaction gap (security, in scope).** `POST /agent-auth/keys/provider-config` carries raw credentials (`value.apiKey` for `azure_openai`) and is NOT covered by the whole-body redaction gate at `main.py:155`; the `_SENSITIVE_INPUT_FRAGMENTS` fallback misses `apiKey`. Reproduced: a 422 echoes the plaintext key. Pre-existing from D1, but this PR's new test at `test_agent_auth_body_redaction.py:51-55` asserts the non-coverage as *correct* — converting a latent gap into a pinned invariant. Fix in this PR: broaden the predicate to cover the typed route and flip that assertion.
2. **Missing `area:anyharness` label** (`agent_auth.rs` doc-comment change requires it per `scripts/ci-cd/pr-metadata.mjs`; PR Metadata goes red on ready-for-review).
3. **PR body route count false**: actual is 12 operations / 10 paths (10 ops on `/agent-auth` incl. the 3 org-policy routes, 2 on `/agent-gateway`), not "9 routes (7+2)".

Nits: (4) add a `test_the_old_agent_gateway_routes_are_gone` 404 sweep mirroring `test_agent_models_api.py:454`; (5) deploy note should say installed desktop builds silently stop syncing local auth state until updated (spec-sanctioned pre-launch, but say it honestly).

pablonyx:
## Cycle-2 verification: FIX REQUIRED (2 items + 1 polish)

The security fix is **confirmed and mutation-proven**: predicate at `main.py:161` covers the full `/agent-auth/keys` subtree, empirically no over-match (all 172 app paths enumerated — only the three intended paths contain the substring; no hidden or catch-all route reachable), and reverting to the old `endswith` turns both new tests red. Labels, route counts, deploy note, fresh-eyes diff, base: all CONFIRMED.

1. **BLOCKER (CI red, introduced by the fix cycle):** `Repo shape checks` fails — the 404-sweep test pushed `test_agent_gateway_api.py` to 816 lines vs allowlisted 784. Bump `scripts/max_lines_allowlist.txt` (after item 2, which adds lines) and correct the PR body's now-false "check_max_lines.py → passed" claim.
2. **Vacuous org-policy assertions:** the three `/organizations/{random-uuid}/agent-gateway/policy*` 404s come from the org guard, not routing — a mutation re-adding all three old org routes still passed. Add a structural assertion pinning the exact sorted (METHOD, path) set for both prefixes from `app.openapi()`, modeled on `test_agent_models_api.py:463-489`.
3. **Polish (taken):** the e2e redaction tests use a type-error body whose pydantic echo never contains the raw key, so the `not in` assertion is trivially satisfied; a missing-field body (omit `kind`) echoes the whole body and makes the raw-key assertion load-bearing.

pablonyx:
Cycle-2 fixes pushed @ 7b305c2ca:

1. **`server/tests/integration/test_agent_gateway_api.py`** — added `test_agent_auth_and_agent_gateway_prefixes_carry_the_right_routes` (structural OpenAPI-schema assertion, modeled on `test_agent_models_api.py:463-489`). The existing 404 sweep's 3 org-policy assertions only proved the org guard rejects a random org id, not that routing moved (proven by mutation) — kept them as belt-and-braces, added the structural pin alongside. Pins the exact sets: 10 ops/8 paths under `/agent-auth`, exactly `GET /capabilities` + `GET /enrollment` under `/agent-gateway`.
2. **`server/tests/unit/test_agent_auth_body_redaction.py`** — `test_validation_error_handler_redacts_raw_api_key_for_provider_config`'s invalid body used a title type-error whose pydantic echo never contained the raw key, so `RAW_KEY not in encoded` was vacuously true. Switched to a missing-`kind` body so pydantic's "missing" error echoes the whole input (including `value.apiKey`). Verified load-bearing: reverting `main.py`'s predicate to the old `endswith` form makes this exact test fail on the `not in` assertion (raw key leaks into the 422 body), not just the `[redacted]` assertion. `main.py` restored clean after the check.
3. **`scripts/max_lines_allowlist.txt`** — bumped `test_agent_gateway_api.py`'s entry to the exact new count (858 lines) so `check_max_lines.py` passes again.

Verification: `test_agent_gateway_api.py` 27 passed, `test_agent_auth_body_redaction.py` 6 passed, `ruff check`/`ruff format --check` clean on touched files, `check_max_lines.py` passes.

pablonyx:
## Cycle 2 fixes verified — READY

Coordinator verification of `b54e004f2..7b305c2ca` (diff inspected directly): exactly the three specified changes.

- **Structural route pin**: `test_agent_auth_and_agent_gateway_prefixes_carry_the_right_routes` asserts the exact sorted (METHOD, path) sets from `app.openapi()` — all 10 `/agent-auth` operations and exactly the 2 `/agent-gateway` ones. Closes the vacuous-org-assertion hole for every route.
- **Redaction witness strengthened**: missing-field body makes the raw-key assertion load-bearing — mutation-verified (old predicate → test fails on the `not in` leak assertion itself).
- **Allowlist** bumped to exact 858; `check_max_lines.py` passes; PR body corrected.

Cumulative: the security fix (redaction over the full `/agent-auth/keys` subtree) is mutation-proven with empirical no-over-match across all 172 app paths. CI fully green at `7b305c2ca`.

**Ready for Pablo's review.** Stacking: base is `agents/d3-provider-apply-python` (#1536) — rides the D-chain train (SDK path changes). Deploy note in body covers the installed-desktop hard-cutover window.

pablonyx:

---

PR #1546, opened 2026-07-27
feat(worker): sync fresh model snapshots to the cloud on every heartbeat (S2b)

pablonyx:
## Spec anchor

specs/codebase/platforms/product/model-catalog.md, "Write paths" (worker
sync design), "Runtime routes" (the status route's fields), "Cloud routes"
(the ingest route's shape), "Code map" (`model_snapshot_sync.rs`'s home).

## Spec contradiction resolved

model-catalog.md:660 already claimed `GET /v1/agents/{kind}/model-snapshot`
serves "model and mode lists off the same document read", but the shipped
`ContextStatus` projection (`status.rs`) carried only `modelCount`/
`modeCount` — no route anywhere on this branch, or any sibling in-flight
`agents/*` branch, exposed the actual model/mode arrays. The Worker sync this
PR adds had nothing to read.

Per the brief's own instruction ("If the cloud route expects a different
shape than assumed, follow the code, not the brief — and say so"), the same
principle applies on the runtime side: this PR **completes the route
model-catalog.md already documents**, additively — `modelCount`/`modeCount`
stay untouched for existing consumers, `models`/`modes` ride alongside as the
`SnapshotModel`/`SnapshotMode` arrays already sitting on `SnapshotEntry`. No
new route, no schema break, `authFingerprint` stays off the wire (existing
law, now pinned by an added assertion). OpenAPI + generated SDK types were
regenerated in the same PR (`pnpm run generate` in `anyharness/sdk`) —
diff is additive-only.

## Gap bullet

Added and struck in this PR (program convention: the PR that closes a gap
usually also names it, if no earlier PR did). See "Current gaps" in the
spec — narrowed rather than fully closed, because the cloud ingest route
itself ships on a separate branch not yet merged into this one's base (see
Deploy notes).

## Design summary

Replaces the deleted gateway-catalog mirror push (`796ff1f08`), mirroring
`catalog_sync.rs`'s exact shape and heartbeat wiring:

- `ModelSnapshotSyncState`: in-memory last-pushed `probedAt` per
  `(harness_kind, auth_context_id)` — the `probedAt` analogue of
  `catalog_sync`'s ETag cache.
- Pure `plan_pushes()` decides which contexts changed; async `maybe_sync()`
  does the GET/POST work and never propagates an error.
- Runs first in `heartbeat_and_converge` (before the D5 bridge — a
  self-update exec never returns, so anything this tick must precede it).
- `CloudClient::ingest_model_snapshot` — new method, `POST
  /v1/cloud/agent-models/{harness}/refresh`, body
  `{authContextId, snapshotJson, probedAt}` (confirmed against B4's actual
  shipped `AgentModelSnapshotIngestRequest`, not just the spec prose).

## Deviations from the brief

1. **Cloud route not on this branch's base.** B4 (`agents/b4-snapshot-rekey`)
   ships the ingest route on a separate, unmerged branch. Against today's
   server (or anything built from this branch's base) the POST 404s —
   logged and swallowed exactly like every other convergence failure. The
   sync is **inert, not broken**, until B4 + the route cutover (C2) land.
2. **Cloud-sandbox scoped, not "desktop worker".** The brief's framing
   ("desktop worker's model-snapshot sync job") undersold a restriction
   already in B4's server code: `resolve_upload_owner` 403s any worker whose
   `runtime_kind != "cloud_sandbox"`. This sync module doesn't special-case
   that — a desktop Worker's attempt just logs a swallowed 403 like any other
   push failure — but the module doc comment and this PR body say so
   plainly rather than imply desktop uploads succeed. This is also exactly
   what model-catalog.md's own "Desktop does not sync" law already says.
3. **Runtime status route extended** (see "Spec contradiction resolved"
   above) — the one piece of scope beyond the worker-side file list the
   brief named, and the smallest change that makes the spec's existing
   claim true.

## Tests

| Area | Test | Proves |
| --- | --- | --- |
| `plan_pushes` | `plan_pushes_uploads_when_probed_at_advances` | pushes when `probedAt` advances |
| `plan_pushes` | `plan_pushes_does_not_repush_unchanged_probed_at` | no re-push on unchanged `probedAt` |
| `plan_pushes` | `plan_pushes_skips_a_context_with_no_snapshot_yet` | skips harnesses/contexts with no entry |
| `plan_pushes` | `plan_pushes_treats_each_harness_independently` | same `probedAt` under a different harness still pushes |
| `plan_pushes` | `plan_pushes_handles_no_contexts_at_all` | empty input is a no-op |
| bearer resolution | `resolve_runtime_bearer_token_prefers_config_over_env` | config wins over env |
| bearer resolution | `resolve_runtime_bearer_token_falls_back_to_env` | env fallback when config absent |
| integration (`TcpListener`) | `maybe_sync_uploads_a_changed_context_with_the_workers_bearer` | end-to-end GET→diff→POST, worker's own bearer on the cloud request |
| integration (`TcpListener`) | `maybe_sync_does_not_wedge_on_a_cloud_error_and_leaves_the_push_pending` | a cloud 500 doesn't panic/wedge, and the unrecorded push retries next tick |
| status projection | extended `the_status_projection_shapes_every_field_and_never_exposes_the_fingerprint` (T-45) | `models`/`modes` serialize with content, `authFingerprint` still absent |
| status projection | extended `a_context_with_no_entry_projects_as_missing` | no entry ⇒ empty `models`/`modes` |

**9 new tests** in `proliferate-worker` (77 total pass, was 68); 0 new test
*functions* in `anyharness-lib` (2 existing tests extended with new
assertions; 57 model-snapshot tests pass, same count as before). SDK: 126
vitest tests + `tsc` typecheck pass unchanged.

`cargo test -p proliferate-worker`: 77 passed, 0 failed.
`cargo test -p anyharness-lib model_snapshot`: 57 passed, 0 failed.
`pnpm run typecheck && pnpm run test` in `anyharness/sdk`: pass.
`python3 scripts/check_max_lines.py` / `check_docs.py`: pass (new
`model_snapshot_sync.rs` split into a sibling `_tests.rs`, mirroring
`installer/downloads.rs`/`downloads_tests.rs`, to stay under 600 lines).

## Deploy notes (worker-side)

- **Harmless against a server that predates this design.** A 404 on
  `GET /v1/agents` or the model-snapshot route (an old runtime) short-circuits
  with a log line, same as `catalog_sync.rs`'s old-runtime handling. A 403 or
  404 on the cloud POST (old server, or a desktop worker) logs and swallows;
  the next heartbeat tick retries from scratch since nothing was recorded as
  pushed.
- **Must not ship user-visible expectations ahead of B4/C2.** This PR alone
  makes the Worker attempt uploads that 404 against today's deployed server.
  That's inert (no user harm, no crash), but the gap bullet this PR strikes
  is only honestly closed once B4 + the route cutover deploy in the same
  release train — flagging this the same way the "legacy gateway-models
  route" gap bullet above it already flags its own train dependency.
- No new worker config fields; `runtime_base_url`/`runtime_bearer_token`
  already existed.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/domains/agents/model_snapshot/staleness_tests/projection_tests.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/model_snapshot/staleness_tests/projection_tests.rs
@@ -97,6 +97,17 @@ fn the_status_projection_shapes_every_field_and_never_exposes_the_fingerprint() 
     assert!(serialized.contains("\"snapshotAgeSeconds\":3600"));
     assert!(serialized.contains("\"identityComparable\":true"));
     assert!(serialized.contains("\"state\":\"idle\""));
+    // T-45 (S2b): the full models/modes lists ride alongside the counts —
+    // model-catalog.md:660 always said this route serves "model and mode
+    // lists off the same document read"; the shipped shape undershot that
+    // with counts only, and this closes the gap the Worker's
+    // model_snapshot_sync now depends on.
+    assert_eq!(status.models.len(), 1);
+    assert_eq!(status.models[0].id, "m-1");
+    assert_eq!(status.modes.len(), 1);
+    assert_eq!(status.modes[0].id, "build");
+    assert!(serialized.contains("\"models\":[{\"id\":\"m-1\""));
+    assert!(serialized.contains("\"modes\":[{\"id\":\"build\""));
 
     // An indeterminate identity must tell the UI to claim no version binding.
     let pre_field = entry(HOUR, None, secret_fingerprint);
@@ -137,7 +148,10 @@ fn the_status_projection_shapes_every_field_and_never_exposes_the_fingerprint() 
         next_attempt_at: Some(next),
     });
     assert_eq!(backoff.last_error.as_deref(), Some("timeout"));
-    assert_eq!(backoff.probed_at.as_deref(), Some(failed_probed_at.as_str()));
+    assert_eq!(
+        backoff.probed_at.as_deref(),
+        Some(failed_probed_at.as_str())
+    );
     assert_eq!(backoff.model_count, 1, "the last good list keeps serving");
     assert_eq!(backoff.state, LiveState::Backoff);
     assert_eq!(backoff.next_attempt_at, Some(next.to_rfc3339()));
@@ -190,6 +204,8 @@ fn a_context_with_no_entry_projects_as_missing() {
     assert_eq!(status.probed_at, None);
     assert_eq!(status.snapshot_age_seconds, None);
     assert_eq!(status.model_count, 0);
+    assert!(status.models.is_empty(), "no entry means no models to list");
+    assert!(status.modes.is_empty(), "no entry means no modes to list");
     let _unused: BTreeMap<String, String> = BTreeMap::new();
 }
 
@@ -205,22 +221,15 @@ fn a_context_with_no_entry_projects_as_missing() {
 /// of a silent client break.
 #[test]
 fn every_live_state_serializes_exactly_as_the_generated_schema_declares() {
-    let schema = serde_json::to_value(
-        utoipa::openapi::schema::Schema::from(
-            <LiveState as utoipa::PartialSchema>::schema(),
-        ),
-    )
+    let schema = serde_json::to_value(utoipa::openapi::schema::Schema::from(
+        <LiveState as utoipa::PartialSchema>::schema(),
+    ))
     .expect("serialize the generated schema");
     let declared: Vec<String> = schema["enum"]
         .as_array()
         .expect("the schema must declare an enum")
         .iter()
-        .map(|value| {
-            value
-                .as_str()
-                .expect("enum values are strings")
-                .to_string()
-        })
+        .map(|value| value.as_str().expect("enum values are strings").to_string())
         .collect();
 
     let wire: Vec<String> = [
--- a/anyharness/crates/anyharness-lib/src/domains/agents/model_snapshot/status.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/model_snapshot/status.rs
@@ -22,7 +22,9 @@ use chrono::{DateTime, Utc};
 use serde::Serialize;
 use utoipa::ToSchema;
 
-use super::document::{InstallIdentity, SnapshotAttempt, SnapshotEntry};
+use super::document::{
+    InstallIdentity, SnapshotAttempt, SnapshotEntry, SnapshotMode, SnapshotModel,
+};
 use super::staleness::{self, IdentityComparison};
 use super::ProbeEngineMode;
 
@@ -128,6 +130,20 @@ pub struct ContextStatus {
     pub attestation: Option<super::document::SnapshotAttestation>,
     pub model_count: usize,
     pub mode_count: usize,
+    /// The full model list off the same document read as `modelCount`. Added
+    /// alongside the count (never replacing it — model-catalog.md:660 always
+    /// said this route serves "model and mode lists off the same document
+    /// read"; the count-only shape undershot that, and this is the fix) so a
+    /// machineless-surface uploader (the Worker's `model_snapshot_sync`) has
+    /// something to read besides raw disk access to a document it should not
+    /// know the layout of.
+    #[serde(default)]
+    #[schema(value_type = Vec<Object>)]
+    pub models: Vec<SnapshotModel>,
+    /// The full mode list, same rationale as `models` above.
+    #[serde(default)]
+    #[schema(value_type = Vec<Object>)]
+    pub modes: Vec<SnapshotMode>,
     #[serde(default)]
     pub warnings: Vec<String>,
 }
@@ -177,7 +193,10 @@ pub fn context_status(inputs: ContextStatusInputs) -> ContextStatus {
     let snapshot_age_seconds = entry.as_ref().and_then(|entry| {
         DateTime::parse_from_rfc3339(&entry.probed_at)
             .ok()
-            .map(|parsed| now.signed_duration_since(parsed.with_timezone(&Utc)).num_seconds())
+            .map(|parsed| {
+                now.signed_duration_since(parsed.with_timezone(&Utc))
+                    .num_seconds()
+            })
     });
     let last_error = entry.as_ref().and_then(|entry| {
         matches!(
@@ -206,6 +225,14 @@ pub fn context_status(inputs: ContextStatusInputs) -> ContextStatus {
         attestation: entry.as_ref().and_then(|entry| entry.attestation.clone()),
         model_count: entry.as_ref().map(|entry| entry.models.len()).unwrap_or(0),
         mode_count: entry.as_ref().map(|entry| entry.modes.len()).unwrap_or(0),
+        models: entry
+            .as_ref()
+            .map(|entry| entry.models.clone())
+            .unwrap_or_default(),
+        modes: entry
+            .as_ref()
+            .map(|entry| entry.modes.clone())
+            .unwrap_or_default(),
         warnings: entry.map(|entry| entry.warnings).unwrap_or_default(),
     }
 }
--- a/anyharness/crates/proliferate-worker/src/cloud_client/mod.rs
+++ b/anyharness/crates/proliferate-worker/src/cloud_client/mod.rs
@@ -113,6 +113,21 @@ pub struct HeartbeatResponse {
     pub supervisor_bridge: Option<SupervisorBridgeInputs>,
 }
 
+/// A Worker's upload of one changed model-snapshot entry (model-catalog.md
+/// "Write paths"/"Storage"): the ingest route absorbs today's separate
+/// `refresh`-with-payload and `mirror` routes. `snapshot_json` deliberately
+/// carries only the subset of the machine-document entry the cloud tier
+/// documents reading (`models`, `modes`, `attestation`, `warnings`) — see
+/// `model_snapshot_sync.rs` for why `authFingerprint`/`mechanism`/
+/// `installIdentity`/`lastAttempt` are never assembled here.
+#[derive(Debug, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct IngestModelSnapshotRequest {
+    pub auth_context_id: String,
+    pub snapshot_json: String,
+    pub probed_at: String,
+}
+
 /// Server-delivered D5 bridge inputs (R9R-002). Carried on the heartbeat ack so
 /// an already-provisioned legacy Worker — whose on-disk config predates the
 /// supervisor-owned shape — can materialize the Supervisor config AND a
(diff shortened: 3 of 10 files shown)


pablonyx:
## Adversarial review — S2b (worker model-snapshot sync + additive status-route lists)

**Verdict: READY.** 0 blocking findings. Both targeted suites green at `0949fd0ed`
(`cargo test -p proliferate-worker model_snapshot_sync` → 9/9;
`cargo test -p anyharness-lib model_snapshot` → 57/57), plus
`check_max_lines.py` and `check_docs.py` pass.

The headline check — the wire contract against B4's actual shipped server, not the
spec prose — **passes exactly.** Details below, then non-blocking notes.

---

### Dimension 5 (highest value): wire-contract match vs. `agents/b4-snapshot-rekey` — PASS

I diffed the coded request against B4's real handler rather than the spec text.
Every axis lines up:

| Axis | S2b worker (`cloud_client/mod.rs:204-231`) | B4 server | Match |
| --- | --- | --- | --- |
| Method + path | `POST {base}/v1/cloud/agent-models/{harness_kind}/refresh` | `router = APIRouter(prefix="/agent-models")` + `@router.post("/{harness_kind}/refresh")`, mounted `include_router(cloud_router, prefix=f"{api_prefix}/v1")` with `api_path_prefix: str = ""` → `/v1/cloud/agent-models/{harness_kind}/refresh` | ✅ |
| Auth | `Authorization: Bearer <worker_token>` via `auth::bearer_header` | `authenticate_worker` → `bearer_token_from_request` requires `scheme.lower() == "bearer"` + non-empty token | ✅ |
| Body keys | `#[serde(rename_all = "camelCase")]` on `IngestModelSnapshotRequest` → `authContextId` / `snapshotJson` / `probedAt` | `AgentModelSnapshotIngestRequest` with `Field(alias="authContextId"/"snapshotJson"/"probedAt")` + `populate_by_name=True` | ✅ |
| No user identity in body | 3 fields only; owner never sent | `resolve_upload_owner` derives owner from the sandbox row, ignores the body | ✅ |
| `snapshotJson` inner shape | `{probedAt, models, modes, attestation, warnings}` (`plan_pushes`, `model_snapshot_sync.rs:152-158`) | `parse_snapshot_entry` shape-checks only `models`/`modes` as arrays of `{id: str}`-or-string, stores the rest verbatim; unknown fields tolerated by design | ✅ |
| `probedAt` parse | chrono `to_rfc3339()` (`entry.rs:26`) → `2026-…T…+00:00` | `datetime.fromisoformat(probed_at)`; verified on py3.12 (server pins `>=3.12`, Dockerfile `python:3.12-slim`) that it accepts both the offset form and 9-digit fractional seconds | ✅ |
| Slug validation | ids come from the catalog (`gateway`, `anthropic-api`, …) and harness kinds from `/v1/agents` | `_SLUG_PATTERN` letters/digits/`._-`, ≤64 chars | ✅ |
| Payload ceiling | uncapped | `MAX_SNAPSHOT_JSON_BYTES = 512 KiB` → 400 | see NB-1 |

Also confirmed the base branch (`agents/a8-poke-wiring`) has **no** `agent-models`
router at all, so the "404 today, inert not broken" claim in the PR body is
accurate — and the 404 is a genuine FastAPI route-miss, not a shadowed handler.

### Dimension 1: additivity of the status-route extension — PASS

- `modelCount`/`modeCount` are byte-identical in name, type (`usize` → `integer,
  minimum: 0`), and derivation (`entry.models.len()` / `entry.modes.len()`,
  `status.rs:226-227`). C3's UI badge (`HomeOnboardingCards.tsx:132`, fed via
  `HomeNextScreen.tsx:113`) is untouched — no regression.
- Generated OpenAPI `ContextStatus.required` is unchanged
  (`authContextId, active, state, stale, identityComparable, modelCount, modeCount`);
  `models`/`modes` land as optional. Existing TS consumers keep typechecking.
- SDK regen drift: the whole `anyharness/sdk` diff vs. base is `+26 / -0` across
  exactly the two generated files, and every added line is inside the
  `ContextStatus` schema. Nothing else moved — this is a faithful `pnpm run
  generate`, not a hand-edit. The `#[schema(value_type = Vec<Object>)]` choice
  matches the file's existing convention for `installIdentity`/`lastAttempt`/
  `attestation`.

### Dimension 2: secret hygiene on the new surface — PASS

- `authFingerprint` has no field on `ContextStatus`, so it structurally cannot
  ride the new arrays; the added assertion in `projection_tests.rs` pins that.
- `SnapshotModel`/`SnapshotMode` carry only `id`/`provider`/`name`/`description`/
  `configOptions` and `id`/`name` — no credential material. The uploaded entry
  deliberately omits `authFingerprint`, `mechanism`, `installIdentity`,
  `lastAttempt` (which is where redacted probe error detail lives), so the push
  body is strictly narrower than the local document.
- No token is ever logged. `resolve_runtime_bearer_token`
  (`model_snapshot_sync.rs:267`) returns the value and never traces it; the
  worker bearer is passed by reference into `ingest_model_snapshot` and appears
  only in a header. Both `warn!` sites log `?error` where
  `WorkerError::Cloud { status, body }` carries a *server response* body, not the
  request — and `logging.rs`'s `scrub_event`/`scrub_log` cover it anyway.
- `IngestModelSnapshotRequest` derives `Debug` but is never `Debug`-formatted,
  and it holds no auth material regardless.

### Dimension 3: sync-job correctness — PASS

- **Push-on-advance**: `plan_pushes` (`:139-166`) pushes iff `probed_at.is_some()`
  **and** `last_pushed[(harness, ctx)] != probed_at`. `probedAt` is the "last
  SUCCESSFUL observation" and never regresses on failure (`attempt.rs`
  `record_failed_attempt` touches only `lastAttempt`), so a failing probe cannot
  generate churn. No thundering re-push: state is written **only** on a 2xx
  (`:245`), so a steady-state tick is 6 local GETs and zero POSTs.
- **No stale/empty push**: a `probedAt: None` context is skipped. An entry with
  `models: []` *is* pushed — correct, and load-bearing: B4's read layer treats
  `origin == "snapshot"` with an empty list as truth ("the harness advertised
  nothing under that credential") versus the catalog seed. Suppressing it would
  silently pin the seed tier forever.
- **404-swallow narrowness**: the special-cased `NOT_FOUND` arm (`:186-189`) is
  scoped to the *runtime-side* `GET /v1/agents` only, and it downgrades to
  `info!` rather than dropping the signal. Every other runtime status
  (401/403/5xx) falls to the `warn!` arm at `:191`. The **cloud** POST has no
  404 special case at all — all non-2xx go through the single `warn!` at `:254`
  and, critically, are **not recorded as pushed**, so the next tick retries. So
  the swallow is narrow in the right direction: nothing is silently permanent,
  and a persistent 401/403/500 emits a `warn!` every 30s, which the
  `sentry-tracing` default filter routes to a Sentry log/breadcrumb. No class of
  real breakage is masked.
- **Tick placement**: the brief's "tick before swap" is satisfied.
  `maybe_sync` is called at `runtime.rs:121-127`, after the heartbeat and gateway
  repair, and **before** `maybe_run_bridge`, `converge_anyharness_runtime`, and
  `self_update::converge` — the last of which `exec`s and never returns. This is
  the exact slot the deleted `catalog_sync` held, and it matches
  `specs/codebase/structures/proliferate-worker/guides/runtime.md:33`'s documented
  order. State is constructed once in `run` and threaded through both call sites,
  so `--once` and the loop share the cache. `--once` runs the sync (a network
  write) — consistent with the guide's "`--once` … can synchronize the catalog"
  and its restriction to *binary updates* only, so no `dry_run` gate is owed here.

### Dimension 4: test quality — load-bearing, with one soft spot

I mutation-tested the assertions mentally. The five `plan_pushes` cases each fail
on a distinct real mutation: dropping the equality check breaks
`does_not_repush_unchanged_probed_at`; dropping the `?` on `probed_at` breaks
`skips_a_context_with_no_snapshot_yet`; keying the cache on context alone
(dropping `harness_kind`) breaks `treats_each_harness_independently`.

The `TcpListener` cases assert the **actual wire body**, not merely "a request
happened":
- `push.path == "/v1/cloud/agent-models/opencode/refresh"` (`:244`) — catches a
  wrong route.
- `authorization == "Bearer worker-secret-token"` (`:246-249`) — catches a wrong
  header name, a wrong scheme, or leaking the *runtime* bearer instead of the
  worker's.
- `push.body.contains("\"authContextId\":\"gateway\"")` (`:250`) and
  `"\"probedAt\":\"2026-07-27T00:00:00Z\""` (`:251`) — these are the camelCase
  guards. Removing `#[serde(rename_all = "camelCase")]` from
  `IngestModelSnapshotRequest` makes both fail.
- The runtime side asserts request *order and paths* (`/v1/agents` then
  `/v1/agents/opencode/model-snapshot`).
- The error case proves the negative that matters: `recorded.is_empty()` after a
  500, then re-running `plan_pushes` to show the retry is still planned.

Soft spot (NB-3 below): no assertion that `snapshotJson` is a *string* whose
camelCase inner keys match what B4 parses.

### Dimension 6: gap-bullet discipline & repo shape — PASS

`model-catalog.md:864` is struck (`- [x] ~~…~~`) with an honest **"Narrowing, not
fully closing"** qualifier naming the B4/C2 train dependency and the
cloud-sandbox-only 403 — final state is coherent, and it matches the same
convention the sibling "legacy gateway-models route" bullet uses. The 600-line cap
is respected via the `#[path]` split (320 + 303 lines), mirroring
`installer/downloads.rs`/`downloads_tests.rs`. Labels validated (1 `release:*`,
3 `area:*`).

---

## Non-blocking

**NB-1 — `snapshotJson` has no size guard against B4's 512 KiB ceiling.**
`cloud_client/mod.rs:191-195` + `snapshots.py`'s `MAX_SNAPSHOT_JSON_BYTES`.
`config_options` on `SnapshotModel` is an untyped `serde_json::Value` (a
per-model ACP option matrix), so a harness advertising many models with fat
option blobs could exceed 512 KiB and earn a permanent 400 that retries every
30s forever with no distinguishing log. Cheap mitigation: log the body length
alongside the failure at `:254` so the cause is visible in one log line rather
than requiring a server-side correlation. (Not blocking — no harness observed
today comes close, and the failure is inert.)

**NB-2 — the two runtime GETs use `reqwest::Client::new()` with no timeout.**
`model_snapshot_sync.rs:280` and `:302`. This is faithful to precedent
(`anyharness_update.rs:469` does the same, and the deleted `catalog_sync.rs` did
too), and `status()` is a document read plus non-blocking `Mutex` peeks, so a
hang needs a wedged localhost accept queue rather than a slow probe. But the
`CloudClient` beside it deliberately sets `connect_timeout(10s)` +
`timeout(30s)`, and `supervisor_bridge` uses a 5s timeout — a stalled runtime
socket here would block the whole heartbeat loop indefinitely, which is the one
thing the module doc promises it never does ("this module never wedges the
loop"). A `Client::builder().timeout(...)` shared across both helpers would make
the doc comment true unconditionally. Worth a follow-up rather than a re-cycle.

**NB-3 — one test gap: `snapshotJson`'s inner keys aren't asserted on the wire.**
`model_snapshot_sync_tests.rs:250-251` asserts the *envelope* keys but only
`pushes[0].snapshot_json.contains("\"m1\"")` (`:30`) for the payload. Renaming
the inner `"probedAt"`/`"models"`/`"modes"` keys in the `serde_json::json!` at
`model_snapshot_sync.rs:152-158` to snake_case would keep all 9 tests green while
making B4's `parse_snapshot_entry` silently store an entry whose `models` reads
as `[]` — i.e. the picker serves the catalog seed forever with
`origin: "snapshot"`. One `assert!(push.body.contains("\\\"models\\\":[{\\\"id\\\":\\\"m1\\\"")` in
the happy-path integration test would close it. Given the module's whole purpose
is that payload, this is the assertion I'd most want added.

**NB-4 — `resolve_runtime_bearer_token_falls_back_to_env` doesn't test the
fallback.** `model_snapshot_sync_tests.rs:108-112` calls `remove_var` and asserts
`None`, so it proves "absent config + absent env ⇒ None" — the name promises the
env path is exercised. It presumably avoids `set_var` because this crate's tests
run concurrently and env is process-global (the same hazard
`probe_materialization.rs` solved by parameterizing). Either rename to match what
it proves, or parameterize the resolver on a lookup fn. Note the *empty*-env
filter (`.filter(|v| !v.is_empty())`) is currently untested either way.

**NB-5 — cosmetic.** `cloud_client/mod.rs` drops a blank line at old `:398`
(inside `impl CloudClient`), unrelated to this change; `projection_tests.rs`
carries three unrelated rustfmt reflows. Both harmless, just noise in the diff.

**NB-6 — observation, not a request.** The module can't distinguish "desktop
worker, 403 forever by design" from "cloud worker, 403 because something broke" —
both log the same `warn!` every 30s. The PR body argues correctly that
special-casing it in the module would duplicate server policy, and I agree. Worth
noting only because a desktop fleet will now emit a steady 403 warn stream to
Sentry breadcrumbs; if that proves noisy in practice, the cheapest fix is
demoting a 403 specifically to `info!` with a "expected on desktop" note, keeping
every other status at `warn!`.

pablonyx:
Ready for agent review: PR #1546 @ 0949fd0ed26c53da9d22eaf483ebcfb363615e1f — adversarial review verdict READY (0 blocking, 6 non-blocking; wire-contract vs B4's AgentModelSnapshotIngestRequest verified exact on all 8 axes). NB-3 (pin snapshotJson inner keys in tests) noted as the top follow-up.

pablonyx:

---

PR #1550, opened 2026-07-27
fix(ui): level-bar width per ladder, integration icon tiles, model label naming

pablonyx:
Three unrelated chrome defects reported from the running app.

## 1. Reasoning level bars: width now scales with the ladder

`LevelBarsButton` sized the icon as a **fixed** em box and divided it across `N` bars, so adding levels made each bar *thinner* — a six-level ladder drew `0.15em` hairlines. That is the inverse of the intent.

The bar stroke is now the invariant and the icon width is derived from it:

    iconWidth = N * barWidth + (N - 1) * gap

A bar also never draws shorter than it is wide, so a single-level ladder can’t collapse to a sub-pixel sliver. `ComposerControlButton`’s `iconOnly` variant takes a `min-w-7` floor instead of a fixed `w-7` square, so a wide glyph grows the pill while every single-glyph button still renders as the same 28px square it does today.

The ambient wave was also too fast and too deep: `1.1s` with a `0.55` trough. It is now `8.6s` on `--ease-standard` with a `0.72` trough, and the per-bar stagger is a named constant rather than an inline literal. Still transform+opacity only, still honours `prefers-reduced-motion`.

## 2. Integration icons: one tile, one tier per branch

`IntegrationIcon` had two branches that had drifted apart — image logos rendered full-bleed (`size-full`, artwork touching the tile border), inline glyphs were sized off the callsite font size via `[&_svg]:icon-large`. Same row, same tile, visibly different optical sizes; the flat full-bleed marks were the worst offenders.

Both branches now share one tile shell and resolve artwork from a single semantic `icon-*` tier each. The glyph branch intentionally sits one tier down: those marks carry no internal whitespace in their paths, unlike most image logos, so at an identical tier they read heavier than their neighbours. Sizing is tile-relative, so no callsite changed.

## 3. Model labels: `Claude-Sonnet-5` → `Sonnet 5`

Two independent causes:

- `formatClaudeModelId` structurally *required* a minor version (`claude-x-N-N`), so any generation shipping without one never matched and fell through to the raw catalog id. The minor component is now optional.
- Nothing stripped the redundant vendor family word from catalog-provided labels, the way the leaf formatter already stripped the other vendor’s. It is now stripped per `"/"`-separated segment, so provider-scoped labels keep their provider prefix and lose only the family word.

Date suffixes survive verbatim — hyphens are only treated as word separators in an all-hyphen remainder, so `Claude Sonnet 4.5 (2025-09-29)` → `Sonnet 4.5 (2025-09-29)`, not `(2025 09 29)`. Regression tests cover all three shapes.

## Testing

- `@proliferate/product-client`: 632 files / 3832 tests pass
- `@proliferate/ui`: 12 files / 30 tests pass
- `scripts/check_appearance_scaling.py`: passes (artwork sizing uses semantic icon tiers, no new fixed-utility sites)
- Verified live in a dev profile against the running web app: 3 levels → 14px icon in a 28px pill, 4 levels → 19px icon in a 29px pill


pablonyx:
## Independent review — #1550 @ 1942717e5

Verdict: solid, well-tested foundation PR. The three claimed fixes (level-bar width per ladder, integration icon tile unification, model label naming) are each implemented correctly and backed by real regression tests (not tautologies) — I traced the geometry math, the CSS share-based sizing, and the Claude/GPT-prefix regexes by hand and they hold up on the edge cases I tried (bare `claude-sonnet-5`, two-digit minors, date suffixes, provider-namespaced labels, `Fast`/`(latest)` suffixes). Dead code from the old approach (old `SlackGlyph`, `--color-brand-logo-tile`, `codex-*`/`thread-find-*` class names) is fully removed, not left alongside the new path — no `grep` turned up a stray reference. `check_design_attribution.py`, `report_frontend_structure.py --strict`, and the frontend-structure line-count constraints all pass clean against this diff. One thing worth a founder's eyes: the PR body's own numbers for the level-bar wave animation don't match the shipped CSS.

### Should fix

- **PR body vs. shipped code mismatch on the wave animation.** The description says the ambient wave is "now `2.4s` on `--ease-standard` and a `0.72` trough," but `apps/packages/design/src/css/product.css:1046` ships `animation: composer-level-bar-wave 8.6s var(--ease-standard) infinite;` (trough value of `0.72` is correct, duration is not). Looking at the branch history, two follow-up commits (`c762ca700`, `0178fe1e3`) retuned the duration after the body text was written, and the body was never updated. Not a functional bug, but a stack-root PR's description is the reference other slices and reviewers will cite — worth a one-line edit before merge so nobody chases a "2.4s" number that doesn't exist in the code.

### Consider

- `apps/packages/product-client/src/components/settings/panes/integrations/IntegrationIcon.tsx:112` — `TILE_BASE_CLASS` drops the old `border border-border/70` chrome entirely (previously present on both branches). That's called out as deliberate in the comment ("chrome-free... no border and no background plate"), and grep confirms no leftover `border-border/70` or `bg-brand-logo-tile` references anywhere, so this is a clean, intentional removal rather than a half-migration — flagging only because it's a visible chrome change bundled into what the PR title frames as a sizing fix.
- `apps/packages/ui/src/primitives/Tooltip.tsx` — the `keepOpenOnPress` press-suppression window is a fixed 400 ms deadline rather than a state cleared by pointerup/click. The tests exercise the documented escape hatches (Escape, blur, pointerleave, a press released off-trigger, right-click, a still-held press) and all pass, so the design is deliberate and covered; just flagging that a very slow double-click (>400ms between down and up) could theoretically let the primitive's own pointer-down dismissal win mid-gesture. Edge case, not blocking.

### Verified

- `LevelBarsButton`'s new `iconWidthEm = N * barWidth + (N-1) * gap` geometry matches the PR's stated formula, and `LevelBarsButton.test.tsx` asserts the actual computed em values for 2- and 6-bar ladders against hand-checked math (0.75em / 2.416667em) — these are real assertions on rendered style values, not restated implementation.
- `IntegrationIcon`'s share-based sizing (`.integration-icon-tile` in `product.css` at lines 573-588) is exercised by two tests: one reads the actual generated CSS to assert both shares stay strictly between 0-100%, the other renders the component and asserts neither artwork branch carries its own size class. Together they'd fail if the tile/artwork coupling regressed to the old two-independent-em-sizes bug.
- `formatModelLeafName` / `stripRedundantFamilyPrefix` correctly handle the full matrix I hand-tested: bare `claude-sonnet-5` → `Sonnet 5`, two-digit minors (`claude-opus-4-10` → `Opus 4.10`, not collapsing with `.0`/`.1`), date-suffixed ids (`claude-sonnet-4-5-20250929` → `Sonnet 4.5`, no misread as a minor version), provider-namespaced labels (`Anthropic/Claude Opus 4.6 Fast` → `Anthropic/Opus 4.6 Fast`), and non-Anthropic families (`GLM-5`, `grok-4.3`, `Gemini 2.5 Pro`) passing through untouched.
- No arbitrary `shadow-[...]`, no new `dark:` variants, no `text-xs` menu-row traps introduced anywhere in this diff (checked via targeted grep against the full patch).
- Every renamed CSS class (`codex-thread-find-match` → `transcript-find-match`, `codex-review-diff-card` → `review-diff-card`, etc.) has zero remaining references to the old name anywhere in the tree, and the new name resolves to a real CSS rule — this is a complete rename, not an additive alias.
- `--color-brand-logo-tile` token removal is total: no dangling references in TS/TSX/CSS.
- `scripts/report_frontend_structure.py --strict` and `scripts/check_design_attribution.py` both pass clean against the working tree at this SHA.

PR #1551, opened 2026-07-27
feat(agent-gateway): org-only billing — fail-closed funding, org-only enrollment, legacy migration (D train)

pablonyx:
## Summary

The complete D corridor of the model-gateway re-cut ([model-gateway.md](https://github.com/proliferate-ai/proliferate/blob/main/specs/codebase/platforms/product/model-gateway.md), landed in #1548), shipped as one train because D-1 without D-2/D-3 would leave org enrollments governing-but-unfunded:

- **D-1 — fail closed + default-org payer** (`65d7cfddc`): the interim funding guard is deleted; a subject with no active funded grant mirrors the exhausted floor (`$0.01`, never literal 0 = LiteLLM-uncapped) so keys stop working instead of becoming unlimited; enrollment resolution follows the user's **default org** unconditionally (v1 payer law). The state renderer keeps withholding key material for unfunded subjects, per spec.
- **D-2 — org-only shape for new signups** (`f7b26e067`): signup enrolls the member into their default org's enrollment only (no personal-subject rows); LiteLLM users are minted per **(org, member)** as `org-<org>-user-<id>` — never one global user spanning orgs; the free signup credit lands on the default org's billing subject, deduped per GitHub identity. Also fixes a latent asyncpg partial-index inference bug in `ensure_organization_billing_subject` (ON CONFLICT `index_where` now literal).
- **D-3 — migration of existing users** (`65a12382d`): an idempotent worker-pass (`migrate_legacy_enrollments`, runs ahead of the backfill tick; no alembic change) that re-parents personal enrollments onto the default org, moves the whole credit ledger (grants **and** debits — remaining credit preserved exactly, free-grant dedupe ref rewritten), re-mints keys under the per-(org, member) identity via the fingerprint machinery (per-key granularity: a crash mid-set never re-converts finished keys), revokes personal keys on the proxy, and retires the personal rows (disable-not-delete). The stale-identity sweep also rotates pre-D-2 org keys off the shared `user-<id>` identity. Afterwards the personal-subject path is **deleted**: `ensure_user_enrollment`, the resolver fallback, the user subject-label branch, and `AGENT_GATEWAY_DEFAULT_USER_BUDGET_USD` are gone, enforced by grep-gate tests.

Base note: the first commit (`bb4c406b7`) is already on main via #1547; the effective diff is the three corridor commits.

## Proof ledger (model-gateway.md §Proof)

| ID | Status | Where |
| --- | --- | --- |
| D1 | ✅ | `test_agent_gateway_enrollment.py::test_d1_signup_produces_the_org_only_shape` |
| D2 | ✅ | `test_agent_gateway_usage_credits.py::test_d2_second_account_on_same_github_identity_gets_no_grant`, `::test_d2_creating_additional_orgs_mints_nothing` |
| D3 | ✅ | `test_agent_auth_org_member_gateway.py::test_unfunded_org_fails_closed_at_the_gate`, `::test_unfunded_org_state_render_withholds_gateway_key`, `test_agent_gateway_enrollment.py::test_unfunded_org_enrollment_mirrors_the_exhausted_floor` |
| D4 | ◐ pytest half green (`usage_credits` attribution suite); **live staging half deferred** to the validation pass |
| D5 | ⏳ release-scenario rewiring to org subjects — deferred to the validation pass |
| D6 | ✅ | `test_agent_gateway_migration.py::test_d6_migration_reparents_personal_enrollment_end_to_end` (+ no-double-grant, defer-and-resume, converted-ledger, stale-identity sweep); Playwright intent-test half flagged for the intent-suite owner |
| D7 | ✅ | `test_agent_gateway_key_lifecycle.py::test_added_harness_kind_reopens_enrollment_and_mints_missing_key`, `::test_stale_identity_key_is_revoked_and_reminted_at_sync` |
| D8 | ✅ | `test_agent_gateway_migration.py::TestOrgOnlyGrepGates` (fallback/guard/name-ordered-choice/personal-insert/`f"user-` mintability all asserted absent) |

## Readiness

  in this PR.

## Verification

- Full gateway battery on the train tip (local Postgres 16, real alembic walk): **202 passed / 0 failed** (migration 10, enrollment/key-lifecycle/usage-credits/org-member/topups/store/api/materialization 141, adjacent billing/policy/worker suites 51). D-1 also ran the full `tests/unit` (1227 passed).
- `ruff check` + `ruff format` clean; `scripts/check_docs.py` passes (env-var reference updated).
- Alembic: no new revision; single head unchanged (`ab5316095737`).

## Validation handoff (deferred live pass — needs staging creds)

1. `AGENT_GATEWAY_LITELLM_BASE_URL/_PUBLIC_BASE_URL/_MASTER_KEY` + `AGENT_GATEWAY_ENABLED=true` against **staging**: run a signup → verify LiteLLM team `org-<uuid>`, user `org-<org>-user-<id>`, 4 harness keys (no cursor), team `max_budget` mirrors the free grant; drain the grant → keys refuse (floor `$0.01`), never uncapped (D4 live half).
2. Seed a pre-D-2-shaped account, run one backfill tick, re-verify (1) plus old keys revoked on the proxy (D6 live echo).
3. Rewire `tests/release` t3 scenarios to org subjects and run `t3-bill-4` (D5). Note: D-2-era org enrollments will one-time rotate keys at the next ensure pass (fingerprint material widened) — expected.

https://claude.ai/code/session_01DVmTVqmuhLp2dABiwNHCbx

---
_Generated by [Claude Code](https://claude.ai/code/session_01DVmTVqmuhLp2dABiwNHCbx)_


pablonyx:
## Live validation (D train) — D4/D5/D6 all PASS, plus one observation

Live pass against the **staging** LiteLLM proxy (`staging-gateway.proliferate.com`), isolated DB + server, from the live-validation runbook. Everything below is live, not sandbox-stubbed.

### D4-live — PASS
Fresh signup org `ca8d3fa6-cee8-4c8e-97f3-e709fc0c0661`, user `e7fc088c-…`:
- team alias `org-<org-uuid>`, LiteLLM user `org-<org>-user-<user-uuid>` (per-(org, member), not the old bare `user-<uuid>`); `subject_kind: organization`
- exactly **4** keys (claude/codex/opencode/grok), **no cursor key**, each `models=['<harness>']`, each key `max_budget=None` (money on the team, never the key)
- team `max_budget: 2.0` mirrors the $2 free grant; grant row `source=free_signup, source_ref=free_signup:<org subject id>` on `subject_kind=organization`
- unfunded org (no GitHub identity) → team `max_budget: 0.01`, the exhausted floor, **never literal 0** → fails closed
- cross-group isolation live: claude key → `grok-4` = **403 `key_model_access_denied`** ("This key can only access models=['claude']")
- **attribution**: after real completions on 3 of 4 keys, every row resolves to the right (user, org, harness) with `litellm_team_id` matching the enrollment exactly, `status=imported`, and **zero** `needs_review` rows belonging to our enrollments.
  (The importer logged 372 unresolved-key warnings — those are *other* sessions' keys on shared staging spend, correctly recorded as `needs_review` rather than dropped. Not a defect.)

codex's live completion was blocked upstream by OpenAI org-verification on `gpt-5-mini`, not by anything in this stack.

### D5 — PASS (full cycle, live)
Drove the D6 org to zero, then reactivated:
- importer `exhausted_subjects=1`, enrollment `budget_status=exhausted`
- the previously-completing claude key → **401 "Key is blocked"** on the real proxy
- renderer **withholds material**: `{"harness_kind":"claude","sources":[]}`
- top-up grant → `reactivate_subject_if_credited` → `budget_status=ok`, material returns, and the same key **completes again live**

### D6-live echo — PASS
Seeded the full pre-D shape on real staging (personal enrollment + `user-<uuid>` team/identity + 4 old-format keys, free grant claimed by the personal subject), verified the old claude key completed *before* migrating, then ran `run_enrollment_backfill_once` twice:
- tick 1 converted **1**, tick 2 converted **0** (idempotent)
- org enrollment with `litellm_user_id = org-<org>-user-<user>`; personal row **revoked, not deleted**
- credit moved **exactly once**: one `free_signup` grant of $5.0000 now on the org subject, allocation moved, and **0** grants / **0** allocations left on the personal subject
- all 4 old keys **404 on the proxy**, and the old key now **401s** on a real completion
- post-migration: resolver lands on the org enrollment, 4 fresh keys, real completion succeeds

### Observation (not filed as a bug — please sanity-check the intent)
On runtime exhaustion the team budget mirror is **not** rewritten downward: after drain, keys were correctly blocked but `team/info` still showed `max_budget=5.0` rather than the `$0.01` floor.

That is self-consistent with the spec — model-gateway.md calls the team budget "a backstop against importer lag, not the meter", and enforcement is by key-disable, which worked. The mechanical reason is that `litellm.ensure_team` early-returns on a stored `team_id` (client.py:144), so the floor is only applied at team-creation time; `update_team_budget` is only ever called from `topups._reactivate_enrollment` and `migration.py:172`. The floor itself IS proven live at creation (the unfunded org got `0.01`).

So the two "unfunded fails closed" mechanisms are asymmetric by construction: creation-time floor vs runtime key-disable. Flagging in case the mirror was intended to track downward too; if the current split is deliberate, nothing to do here.

Cleanup: 4 of the 5 staging LiteLLM teams minted for this run are deleted, with all their keys and
their `org-<org>-user-<user>` / `user-<uuid>` identities: `886ddc64…`, `3535c4ea…`, `6e0defab…`,
`bd0d205b…` (all verified `Team not found`). One is deliberately held open —
`c8060671-e082-4484-9ff7-92171407b606` (`org-ca8d3fa6-cee8-4c8e-97f3-e709fc0c0661`), the D4
attribution org — because the E2B cloud-delivery check is still running against it. It is named in
LIVE-VALIDATION-RESULTS.md's uncleaned-resources section and gets deleted when that finishes.

_(Correcting an earlier revision of this comment, which said all of them were already deleted — at
that point they were identified, not yet removed.)_

pablonyx:
## Release-scenario half of D5/D4: T3-BILL-4 rewired, but staging cannot host the assertion yet

Reporting the runbook's item 3. Two separate results.

### 1. The rewire is done and typechecks (`addd8877e` on `agents/live-validation-runbook`)

The runbook flagged `t3-bill-4.ts` as asserting the free grant on the personal
subject. Two things were off in that pointer: the "~line 41" hit is the
*docstring*, and the live assertion was at 133-140. Both are rewired:

- removed `billing.overview(personal)` + `assert.ok(remainingHours > 0)`;
- added a positive on the default org (`llmBalance(defaultOrg).grantedUsd > 0`);
- added `GET /v1/cloud/agent-gateway/enrollment` asserting `subjectKind ===
  "organization"` and a non-null `litellmTeamId`;
- added the negative: any LLM credit still on the personal subject reports
  blocked as un-migrated pre-D residue.

New `resolveDefaultOrgId` helper in `billing-http.ts` (env override → single
active membership → earliest `membership.joinedAt` → undefined so the caller
blocks rather than asserting against the wrong org).

Also fixed a genuine runtime break, not a stale comment: `tier2/t2-bill.ts:93,96`
is the argv of a real `spawn(python -c …)` importing
`ensure_user_free_credit_grant`, which no longer exists — it would `ImportError`
at runtime. Now `ensure_signup_free_credit_grant`.

### 2. Staging is not running the re-cut, so the scenario cannot be green there

`https://staging-app.proliferate.com/api/health` → `{"status":"ok","version":"0.3.48"}`.
0.3.48 is the 2026-07-25 release commit `38390cb1e`, and
`git merge-base --is-ancestor 5a99a6ce9 38390cb1e` is **false** — this PR is not
in the deployed build. The re-cut is VERSION 0.3.50 on main, merged today.

The live proof: the durable bot's enrollment on staging reads

```
{"subjectKind":"user","litellmTeamId":"6fd34ee1-…","syncStatus":"synced"}
```

`subject_kind: "user"` — the pre-D shape, exactly what #1551 removes. So the
rewired assertion is correct about main and correct to fail here; running it
against this deployment measures the old server.

Two incidental staging-fixture facts worth recording:

- `RELEASE_E2E_DURABLE_ORG_ID=0065a7e8-81b5-4d35-b5f8-a55c2aaa9af2` in
  `~/.proliferate-local/dev/release-e2e.env` is stale — staging returns
  `404 organization_not_found`. The bot's actual single org is
  `aded64b4-f3e4-49b3-85a0-efd5ffcb0460` (owner, joined 2026-07-09).
- `RELEASE_E2E_SERVER_URL` in that file points at `http://127.0.0.1:8086`, so
  `--lane staging` needs it overridden to `https://staging-app.proliferate.com/api`.
  The lane flag does not redirect the URL itself.

With both corrected, the run reaches a **pre-existing** assertion I did not
touch (`holdReason` is `overage_disabled`, not `credits_exhausted`) — the
durable org's staging state has drifted from what the scenario asserts. Not a
product defect and not caused by the rewire.

### Status

D5's semantics are already proven live end-to-end outside the harness (full
drain → key 401 "Key is blocked" → renderer withholds → top-up → same key
completes again), reported separately above. The harness half stays
**BLOCKED on deployment**: re-run `--lane staging --only T3-BILL-4` once staging
carries ≥0.3.50, and separately refresh the two stale fixture vars.

pablonyx:
## Test-harness code still hard-requires the pre-D `user-<uuid>` LiteLLM identity (3 sites, not fixed)

Surfaced while rewiring T3-BILL-4 for the org-only cut. These are outside that
scenario and each is a frozen contract owned by another workstream, so I am
reporting rather than rewriting them. All three are verified in the tree at
`15098c21a`, and all three will fail against a post-#1551 stack.

The premise: `ensure_enrollment_row` (`db/store/agent_gateway/enrollments.py:51`)
inserts `subject_kind=organization` only — its own docstring says the
`subject_kind='user'` insert path is gone and the only remaining personal rows
are pre-migration. So anything that *requires* a `user`-kind row or a
`user-<uuid>` LiteLLM identity now requires something no write path produces.

**1. `tests/intent/stack/billing-usage-import.ts:295` — `getUserEnrollment`**
queries `WHERE subject_kind = 'user'`, so it returns `null` on any freshly
enrolled account. Called from `tier2/t2-bill.ts:688, 698, 1012, 1016, 1041`,
where the results feed `assert.ok(personal?.virtualKeyId, …)`. Those tier-2
cells assert a "personal vs org key isolation" guarantee that no longer exists
in that shape — this is a behavioral redesign of three cells, not a rename.

**2. `tests/release/src/worlds/managed-cloud/actor-enrollment-custody.ts`** —
`:75` `const expectedSubject = \`user-${actor.userId}\``, `:206-213` expects
alias `vk-user-<uuid>-<harness>-<enr8>` and user `user-<uuid>`, and **throws**
`"resolved LiteLLM actor subjects do not match the deterministic personal
enrollment identity."` on mismatch. Post-D the identity is
`org-<org>-user-<uuid>` and the alias `vk-org-<org>-user-<uuid>-…`. Reached live
via `world.ts:789 bindActorEnrollment` (LOCAL-WORLD-SMOKE-1, managed-cloud
fixture smoke). Note `:79` already builds the `vk-org-…` prefixes as a fallback,
so this may be a small fix — but the hard `expectedSubject` equality at `:75`
and the throw at `:213` are the blocking part.

**3. `tests/release/src/worlds/selfhost/gateway.ts:538`** —
`PERSONAL_ENROLLMENT_KEY_ALIAS_PREFIX = "vk-user-"`, plus a
`/user/info?user_id=user-<uuid>` lookup, and it explicitly *prefers* `vk-user-`
while rejecting `vk-org-`. Post-D only `vk-org-…` keys exist, so
`litellmResolveActorKeyToken` throws "no minted virtual key". Reached from
`selfhost-qual-1.ts:827`.

### Also: the testing spec still states the pre-D assertion

`specs/developing/testing/scenarios.md:720-723` reads "the org member's compute +
LLM bill the ORG subject (exhausted) while the same user's PERSONAL subject keeps
its free grant". My rewire now contradicts that bullet and agrees with
`specs/codebase/platforms/product/model-gateway.md` §Account model ("Orgs are the
only billing subject"). The scenarios.md bullet needs the same rewire — flagging
it rather than editing, since it is a spec change.

### One unreconciled boundary worth a ruling

`ownerScope=personal` is still fully live on the billing HTTP surface
(`permissions.py:180 ensure_personal_billing_subject_state`, and
`_owner_selection_from_request` *defaults* to personal when no org is selected).
"No personal subject" is true of LLM-credit grants and gateway enrollment, but
not of the billing owner-context / Stripe-customer plane. That is why my new
negative asserts on `llmBalance().grantedUsd` rather than on `overview()`
existing at all. If org-only is meant to reach the HTTP surface too, that is a
gap; if the Stripe-customer plane deliberately keeps personal scope, the specs
could say so explicitly.

PR #1554, opened 2026-07-27
docs(specs): founder-ruled UX layer for agent-auth and model-catalog

pablonyx:
# docs(specs): founder-ruled UX layer for agent-auth and model-catalog

## Summary

Docs-only. Both target specs described the auth/catalog *machinery* in full
but said almost nothing about the surface a user touches. This adds that
layer as target-state deltas — no settled architecture is re-derived — plus
the gap-list entries the rulings imply.

Files:

- `specs/codebase/platforms/product/agent-auth.md` — new **The settings
  surface** section (pane anatomy §1–§7, model attribution, degraded-apply
  probe, open verification items); corrected vault field sets; four gap-list
  changes.
- `specs/codebase/platforms/product/model-catalog.md` — new **Attribution**,
  **Status and refresh in settings**, and **Probing during a degraded apply**
  subsections; four new gap entries.
- `specs/codebase/platforms/product/agent-distribution.md` — the stale cursor
  claim in the cloud carve-out corrected.

`python3 scripts/check_docs.py` → passed (230 Markdown files).

## Ruling ledger

| ID | Ruling | Rationale |
| --- | --- | --- |
| R1 | Per-harness pane is flat titled sections, no cards; Conductor's setting-row rhythm (`reference/conductor/`) is the anchor; existing components only | A card implies a self-contained object; these sections are facets of one harness |
| R2 | §2 auth method is a Conductor-style but uncarded `gateway \| api_key \| native` pick with radio semantics; for opencode it is **not** a gate | The stored model is one enabled source for `SINGLE_SOURCE_HARNESSES`, so the control must be one-of-N; opencode is `MULTI_SOURCE_HARNESSES` and a blocker would lie about a coexistence harness |
| R3 | §3: every method shows an authenticated-status row with refresh, styled exactly like Conductor's; the native row is clickable → refresh or login terminal (`HarnessAuthCliDetails.tsx:110-136`) | One question, one answer shape; the row that reports "not logged in" is the row you click to fix it |
| R4 | §4: one spot per key with **named** paste-first affordances ("Set OpenAI API key", "Configure Bedrock", "Configure Azure") — never a type-picker after paste; always display which kind is set | The affordance clicked already said which kind it is; asking the user to classify a pasted secret is asking them to do the registry's job |
| R5 | Bedrock collects `region` + `bearerToken`; Azure collects `endpoint` + `apiKey` only — the `deployment` field is **dropped** | The renderer deliberately does not translate `deployment`; it folds into a `--model azure/<id>` launch arg, outside `state.json`'s env+files contract (`agent_auth.py:434-446`) |
| R6 | §5 opencode "Add provider" is a near-literal copy of Conductor's: searchable full ~149 vendored providers, logos, featured/already-configured subset expanded with "Show more/fewer", env-var-less rows filtered; selecting + pasting writes a vault `api_key` + one opencode selection row (`env_var_name` from the registry, `provider_hint` = provider id, display-only) | Opencode can talk to all of them, so the picker must not curate the provider choice; 149 rows is a search box's job, not a scroll's |
| R7 | §6 harness-specific options render after auth; §7 model list is auto-collapsed with the **same** status-row component as §3 on the left and refresh on the right | Options sit on top of a working harness; "last checked / check again" is one question whether the subject is a credential or a model list |
| R8 | Opencode's pane job = auth-status clarity (ideally Zen vs subscription-plan distinction) + provider listing; it does **not** show gateway models as primary content | The user's question is "which of my providers are live"; a flat gateway model list answers a different one at that question's expense |
| R9 | Attribution is selection-derived for single-source harnesses (bedrock → AWS, azure → Microsoft, gateway → Proliferate, native → none) and the observation's verbatim `provider` for opencode; explicit icon table, neutral fallback; **never gates anything** | One source serves a single-source harness's whole list, so the selection *is* the attribution; opencode's list is genuinely mixed and `provider` is already carried verbatim |
| R10 | Opencode detector fix: `detect_opencode_local_auth` (`credentials.rs:288-341`) must preserve the discarded provider key and match it against slots' `discoveryKinds`; per-slot results ride the existing `auth_slots` wire field (`readiness/service.rs:118-147`) — **no new wire concept** | Today's whole-file verdict cannot say *which* provider is logged in, which is exactly what §3 and the Zen distinction need |
| R11 | Typed-vault path is blocked by **two** gates, both of which must open together: `getSupportedProviderConfigKinds()` hardcoded `[]` (`provider-config-fields.ts:121-125`, a now-stale stub) and `_assert_keys_usable`'s `kind == 'api_key'` filter + the CHECK constraint (`selections.py:63-92`) | Either gate alone leaves the path dead; the stub's own comment defers to a registry `providerConfig` declaration that has since landed |
| R12 | Degraded-apply: when an apply lands with gateway enrollment sync incomplete (renderer drops the gateway source, possibly `sources: []`), the probe **still runs**; results shown with a co-located "gateway setup in progress" pending badge; the sync-completion trigger auto-refreshes | Never lie about what was observed, and do not invent a no-probe state for a seconds-long window |
| R13 | Cursor strike: the claim that `cursor-agent` ignores `CURSOR_API_KEY` is **stale** — refuted by a live run 2026-07-26, fixed in `4ccbfc41a`. Cursor's api_key route and its `Ready` are honest; manual-refresh-only probing stands for keychain-prompt reasons | The gap asked for a founder ruling on dropping cursor's api_key card; no such ruling is owed once the premise is false |
| R14 | Micro-rulings: keys are disabled-not-deleted when in use, with an "in use by N harnesses" chip (`agent_gateway/service.py:232-240` is the refusal producer); saved-vs-live status coexist; "Zen" is the wording for opencode's own service | Surface a server-side refusal as a state, not as an error the user finds by hitting a 409 |

## Open items

- **Foundry (claude × `azure_openai`) pending quota.** The render arm
  (`agent_auth.py:418-433`) is self-admittedly unverified — the resource-name
  derivation and the API-key-vs-auth-token choice are analogies to opencode's
  proven arm. Ruling is verify-before-offering; the live run is blocked on
  quota approval, so it is recorded as an open verification cell and the card
  renders pending.
- **models.dev logo license check.** §5's per-row logos depend on vendoring
  `models.dev/logos/<id>.svg`. The license review gates the vendoring step,
  not the picker's design.
- **Expanded-row interaction is an assumption.** §5 specifies an inline paste
  field in the selected row; the Conductor capture did not show that state.
  Resolve against the reference before implementing.
- **Desktop pull-invalidation dependency.** §3's status and §7's list are only
  as fresh as the auth-state query's invalidation, owned by agent-auth's
  desktop delivery loop — a dependency, not a new mechanism.

## Implementation plan

Ordered; each step unblocks the next.

1. **Detector fix (R10).** Preserve the provider key in
   `detect_opencode_local_auth` and match `discoveryKinds` per slot. Results
   ride the existing `auth_slots` field — no wire change. Precondition for
   §3's per-provider rows and the Zen distinction.
2. **Typed-vault two-gate unblock (R11).** Replace
   `getSupportedProviderConfigKinds()`'s `[]` with a read of the harness's
   registry `providerConfig` minus `pending` kinds; relax
   `_assert_keys_usable` and loosen the
   `ck_agent_auth_selection_api_key_shape` `env_var_name IS NOT NULL`
   requirement by migration. Both in one change — either alone is a no-op.
3. **Pane rebuild per §1–§7 (R1–R5, R7, R8, R14).** Flat sections on the
   Conductor row rhythm; universal status row; named paste-first key
   affordances with the corrected Bedrock/Azure field sets; disabled-not-
   deleted keys with the in-use chip; collapsed model section reusing the
   status row.
4. **Provider modal + logo vendoring (R6).** Featured/collapsed split, search
   over the full vendored list, the two writes on select. Logo vendoring lands
   only after the license check clears; the modal ships with the neutral
   fallback until then.
5. **Attribution icons (R9).** Explicit icon table with neutral fallback;
   selection-derived for single-source harnesses, `provider`-derived for
   opencode; replace `getProviderDisplayName(harnessKind)` in the composer
   badge.
6. **Degraded-apply pending badge (R12).** Co-located badge driven off
   enrollment sync status, with the existing sync-completion freshness trigger
   doing the repair.

**Sequencing note:** this is a specification pass only. None of the above is
buildable until corridor trains **B / C / D** land — the pane rebuild depends
on the ack-gated delivery and probe-status surfaces those trains introduce.


pablonyx:
Superseded by a re-cut off post-merge main — the head branch here is the stale pre-squash review branch and reports unmergeable. Same delta, fresh base: see the replacement PR.

PR #1558, opened 2026-07-27
feat(agent-auth): open the typed-config write gate, registry-driven

pablonyx:
## Summary

The final corridor of the agent-auth re-cut: typed vault entries (`aws_bedrock`, `azure_openai`) become selectable, gated by the registry's `providerConfig` declarations. Implements #1554's **R11** (both gates open together — either alone is a no-op) and **R5** (Azure collects `endpoint` + `apiKey` only; `deployment` dropped):

- **Frontend gate**: `getSupportedProviderConfigKinds()` un-stubbed — reads the bundled agent registry's `providerConfig` declarations minus `pending` kinds (same bundling mechanism as the agent catalog).
- **Store gate**: `_assert_keys_usable` is kind-aware — bare keys require `env_var_name`, typed entries forbid it (Proof A6), and a typed kind must be registry-declared non-pending for the harness (new 400 otherwise). The service threads the D-1 `supported_provider_config_kinds` seam; the store default stays closed. Migration `d6e8f0a2b4c6` (parent `c2a4e6b8d0f2`, single head, round-tripped) loosens the CHECK to `api_key_id IS NOT NULL`; the env-var-iff-bare-kind XOR spans tables and lives in the store per the spec's shape-check law.
- **Open combos**: bedrock × claude/codex/opencode, azure × opencode. **azure×claude (Foundry) stays pending** — the registry was missing the `pending` flag (the gate would have opened it against the ruling); this PR adds the flag with the verify-before-offering reason. azure×codex pending pre-existing.
- **Two real catches along the way**: `GET /v1/cloud/agent-auth/state`'s pydantic model silently dropped `provider_config` sources (`config_kind`/`env` missing, `kind` literal too narrow) — fixed, cloud + anyharness SDKs regenerated; and the A5 end-to-end now runs through the real write path (PUT selections → state render) instead of direct-insert scaffolding.

**Known follow-up (flagged in-code)**: the typed-config create buttons now render, but `handleProviderConfigSubmit` is still the placeholder from D-2 — wiring the mutation + typed-row editor support is the settings-surface work owned by #1554's step 3 (R4's named paste-first affordances). The server path is fully open and proven via API tests.

## Proof ledger
- **A5**: write→render end-to-end via the real PUT + `GET /state` (typed source carried on the wire).
- **A6**: both shape rejections (bare-without-env-var, typed-with-env-var) at the API boundary.
- Pending-combo refusals pinned by test (azure×claude, azure×codex).

## Readiness

  in this PR.

## Verification

- Server: 14-file agent-auth/gateway battery **195 passed / 0 failed** (real alembic walk; single head `d6e8f0a2b4c6`; up/down round-trip).
- Frontend: **147 passed / 16 files**, `tsc --noEmit` clean; rust registry validation 15 passed (registry.json edit schema-valid).
- ruff, check_max_lines, check_server_boundaries, check_docs: all green. Spec gap list updated in the same PR (typed-selection gap struck; pending bullet re-cut to cover both azure cells + R5).

## Validation handoff (deferred live pass)

Bedrock end-to-end with real AWS creds (this environment has them): save a bedrock typed key for claude → select → launch → verify `CLAUDE_CODE_USE_BEDROCK=1` + region/credential env in the session and a working completion. Azure×opencode with a real Azure key. Foundry stays parked until its live run flips the registry flag.

https://claude.ai/code/session_01DVmTVqmuhLp2dABiwNHCbx

---
_Generated by [Claude Code](https://claude.ai/code/session_01DVmTVqmuhLp2dABiwNHCbx)_


pablonyx:
## Live validation (typed-config write gate) — A5 server half PASS, runtime-launch half BLOCKED

Driven through the real server API (not the UI — the `HarnessAuthApiKeyDetails.tsx` submit handler is the known #1554-step-3 placeholder, left untouched per the runbook).

### Write gate — PASS, exactly as ruled

Admitted and refused precisely per the registry's non-pending vocabulary:

| selection | result |
|---|---|
| `aws_bedrock` × claude | 200 admitted |
| `aws_bedrock` × codex | 200 admitted |
| `azure_openai` × opencode | 200 admitted |
| `azure_openai` × codex | **400 refused** |

The refusal is the typed one, naming the reason rather than a generic validation error:
```
{"code":"invalid_agent_auth_selection","message":"Harness 'codex' does not support provider-config
 kind 'azure_openai' (not a declared, non-pending registry providerConfig kind)."}
```

Field validation on `POST /v1/cloud/agent-auth/keys/provider-config`:
- **R5 enforced** — `deployment` on an `azure_openai` entry is rejected: `422 invalid_agent_provider_config_fields`, "Unknown field(s) for kind 'azure_openai': deployment."
- missing required field rejected: `422` "Missing required field(s) for kind 'aws_bedrock': bearerToken."
- stored entry redacts correctly: `redactedHint: "azure_openai:2 field(s)"`

### Env translation in the rendered state doc — PASS, byte-exact

`GET /v1/cloud/agent-auth/state?surface=local` (secrets truncated here):
```
claude    provider_config aws_bedrock  {"CLAUDE_CODE_USE_BEDROCK":"1","AWS_BEARER_TOKEN_BEDROCK":"…","AWS_REGION":"us-east-1"}
codex     provider_config aws_bedrock  {"AWS_BEARER_TOKEN_BEDROCK":"…","AWS_REGION":"us-east-1"}
opencode  provider_config azure_openai {"AZURE_API_KEY":"…","AZURE_RESOURCE_NAME":"proliferate-gw-aoai"}
```
Matches `_translate_provider_config_env` exactly: claude gets the `CLAUDE_CODE_USE_BEDROCK` flag var, codex correctly does **not**, and azure×opencode emits `AZURE_RESOURCE_NAME` from the endpoint's first hostname label (not the dead `AZURE_OPENAI_API_KEY`). Doc is `version: 2`, no `entries` map, and `fingerprint` is response-only.

The Azure credential used is live-verified independently (a direct Azure OpenAI call with it returned a completion), so the material rendered is real, not a placeholder.

### BLOCKED — the launch half (session env + real completion through a typed key)

Could not be proven, for a machine-resource reason rather than a code one:

- the runbook's assertion is that a **launched session's env** carries `CLAUDE_CODE_USE_BEDROCK=1` etc. and returns a real completion. That is rendered by the Rust runtime (`route_auth/render.rs`), so it needs an `anyharness` binary built from a tree containing 15098c21a.
- the only prebuilt runtime on this machine (`~/.proliferate-local/dev/runtime-bin/anyharness`) is dated **Jul 4**, 23 days before the re-cut landed; a sweep of all 57 local worktrees found no binary from a tree containing that commit.
- building it was not safe: swap was pinned at ~19.4–20.2 GB of 20.5–21.5 GB with `kern.memorystatus_vm_pressure_level: 2` for the whole run. This machine has OOM/swap-died from concurrent rustc before, so I deliberately did not start a cargo build.

Also **BLOCKED**: `aws_bedrock` with a *real* bearer token. No Bedrock bearer exists in the local env, and per `specs/developing/operating/catalog-probe.md` that credential is supposed to be a dedicated, budget-enforced automation principal — so I did not mint one on the founder's general `pablo-cli` IAM user. The Bedrock cells above are proven at the write-gate + render layer with a placeholder token; the completion needs the dedicated principal.

**Foundry (azure × claude): SKIPPED, and it is a chicken-and-egg worth noting.** It is registry-`pending`, so `_assert_keys_usable` refuses the selection by design — meaning the runbook's "a successful live run flips the flag" can't be executed without first editing `catalogs/agents/registry.json` to un-pend it. That file is `include_str!`'d and dual-gated by a JS validator plus a cargo test, so testing it implies the same Rust rebuild that was unsafe here. Left pending, as the runbook instructs.

PR #1559, opened 2026-07-27
Desktop app opens at a usable default size

pablonyx:
## Summary

- The installed desktop app's main window opened at a fixed 1440x900
  logical size, which reads as far too small both vertically and
  horizontally on most displays — flagged as a tier-one usability
  issue.
- Added `"maximized": true` to the main window config in
  `apps/desktop/src-tauri/tauri.conf.json` so the window opens at the
  OS-reported usable screen size on launch.
- Added `minWidth`/`minHeight` (1024x700) so the window can still be
  resized smaller but not dragged down to something unusable.
- Kept `width`/`height` (1440x900) as the resize fallback, matching
  existing convention in the file.

## Window-state persistence

There is no window-state plugin wired into this app — no
`tauri-plugin-window-state` (or similar) in
`apps/desktop/src-tauri/Cargo.toml`, and no size/position save/restore
code anywhere in `apps/desktop/src-tauri/src/`. So there's no
remembered-size value for `maximized` to clobber on relaunch; this is
a plain default-config change, not an override of saved user state. If
window-state persistence is added later, `maximized` should be scoped
to first-launch only.

## Other window in this file

`bundle.macOS.dmg.windowSize` (660x400, further down in the same file)
is the macOS DMG installer window layout, not an app window — left
untouched.

## Debug variant

`apps/desktop/src-tauri-debug` is a CLI diagnostics binary
(`proliferate-debug`) with no window config of its own, so there is
nothing to keep in sync there.

## Dev config

`apps/desktop/src-tauri/tauri.dev.json` is a sibling config — the base
`scripts/dev.mjs:generatedTauriConfig` spreads for every `make
dev`/`make dev-desktop`/`make dev-local` window, including the
per-profile configs it writes under
`~/.proliferate-local/dev/profiles/*/tauri.dev.json`. It still had the
old `width`/`height` with no `minWidth`/`minHeight`/`maximized`, so
every dev window would have kept opening at the old fixed size while
the packaged app opened maximized. Added the same three keys here to
match.

## Verification

- [x] `python3 -c "import json;json.load(open('apps/desktop/src-tauri/tauri.conf.json'))"` — config parses
- [x] `ruby -e 'require "yaml"; ...'` over `.github/workflows/*.yml` — parses
- [ ] Visual/behavioral verification (actually launching the installed
  app and confirming it opens maximized) — **not done**. This is a
  Tauri build/run change to native window config, and per the current
  machine constraints no `cargo build`/`pnpm tauri dev`/`pnpm tauri
  build` was run. This needs a real build+launch check before merge.


pablonyx:
## Independent review — #1559 @ 32fc26ba0e

Verdict: the change itself is correct for the production bundle config, but it's incomplete — it misses a sibling config that governs every local dev launch, so the fix won't actually be observed by anyone running `make dev`/`make dev-desktop`/`make dev-local` against this stack.

### Should fix

- `apps/desktop/src-tauri/tauri.dev.json` (used by `make dev`, `make dev-desktop`, `make dev-local`, and generated per-profile by `scripts/dev.mjs:generatedTauriConfig`, which spreads this file as its base) still has the old `width: 1440, height: 900` with no `maximized`, `minWidth`, or `minHeight`. Every dev-profile window (including the ones under `~/.proliferate-local/dev/profiles/*/tauri.dev.json`, which are regenerated from this base) will keep opening at the fixed pre-fix size — the exact behavior this PR is meant to eliminate — until this file is updated to match `tauri.conf.json`.

### Verified

- `apps/desktop/src-tauri/tauri.conf.json:18-24`: `width`/`height` (1440x900) kept as the resize fallback, `minWidth`/`minHeight` (1024x700) added, `maximized: true` added. Since the window opens maximized, first-launch placement is governed by the OS-reported usable screen area, not the literal 1440x900 — so there's no offscreen/oversized risk on a small display (e.g. 1280x800) at launch.
- Min-size consistency: sidebar min width is 220 (`apps/packages/product-client/src/lib/domain/preferences/workspace-ui/sidebar.ts:2`) and right-panel min width is 380 (`apps/packages/product-client/src/lib/domain/workspaces/shell/right-panel-model.ts:46`); combined (600) is well under the new `minWidth: 1024`, so a user dragging the window down to the minimum still has enough width for both panels plus content.
- No window-state persistence plugin exists in this codebase (checked `apps/desktop/src-tauri/Cargo.toml` and `src/`), so `maximized: true` cannot clobber a remembered user window size — there isn't one.
- `bundle.macOS.dmg.windowSize` (660x400) is the DMG installer layout, unrelated and correctly left untouched.
- No explicit `center` config is set, but that's moot here since `maximized: true` makes initial centering irrelevant.

PR #1560, opened 2026-07-27
Widen the workspaces list, add row detail, and gate workflows behind a beta notice

pablonyx:
## Workspaces list page

**Width.** The page was pinned at `max-w-3xl` / `px-8`, narrower than every other full-page surface in the app. It now uses **`max-w-5xl` + `px-10`**, which is the standard wide content measure and gutter here — that is what `PageContentFrame` is handed by every other full-page product surface (`WorkflowDefinitionList`, `WorkflowRunDetail`, `WorkflowDefinitionEditor`, `WorkflowResourceState`, and `WorkflowDefinitionsAccessScreen` all pass `maxWidthClassName="max-w-5xl"`), and `px-10` is `PageContentFrame`'s own content gutter. **Source surface: the full-page workflows surfaces via `PageContentFrame`/`ProductPageShell`.** Settings' `max-w-[50rem]` was rejected as the source because that file documents it as a settings-only page-width contract.

This page keeps its own frame instead of adopting `PageContentFrame`, because the cmdk list owns its scroll container — so the measure is matched by hand, with the reasoning recorded inline.

**Row detail added** (both from data the workspace collections query already fetches — no new query, this is a UI-scoped change):

- **Session count** — `Workspace.executionSummary.totalSessionCount`, threaded through the existing sidebar view model (`SidebarWorkspaceItemState`) so the list and the sidebar keep reading one projection. Rendered as glyph + number, reusing the same glyph-plus-number vocabulary as the PR dot and its number so the trailing cluster reads as one system. `0` renders nothing — "no sessions yet" is the default state, not worth a mark.
- **Placement** — a quiet neutral `Badge` for cloud and SSH workspaces only. Local worktrees are the common case and labelling them would stamp every row with the same word.

**Skipped for lack of data:** session count is `null` for cloud-only entries (the cloud summary in this projection carries no execution summary) and for pending entries (no materialization yet, so no summary). Nothing was added that would have required a new request.

**Deliberately not added:** no toolbar, no project filter, no archived toggle, no per-row edit affordance. This page should carry less, not more — the filter input and the rows are the whole surface.

Components are the standard ones: `Badge` and `Spinner` from `ui/primitives`, the shared `Command`/`CommandItem` list, and the slot vocabulary (leading well, name, chevron, branch, trailing cluster) follows the sidebar workspace row.

## Workflows beta gate

Entering `/workflows` now raises a **"This feature is in beta"** modal built on the existing `ModalShell` dialog primitive, offering *Continue anyway* or a return to home.

Purely additive — **no workflows code was deleted or made conditional**:

- the real surface mounts *behind* the modal (the new test asserts the definitions surface is rendered while the notice is up),
- every workflows component, hook, store and test is untouched,
- acknowledgement is per-visit React state, not persisted storage.

**Removal is one edit:** set `WORKFLOWS_BETA_GATE_ENABLED = false` in `WorkflowsPage.tsx`, or delete that flag together with `WorkflowsBetaGateModal` and `WORKFLOW_BETA_COPY`. Every piece is commented `TEMPORARY`.

## Verification

No dev server, no cargo, no Docker — unit tests and typecheck only.

- `@proliferate/design` build (check-theme, 276 tokens) — pass
- `@proliferate/ui` tests 30/30, `@proliferate/product-ui` 197/197
- `@proliferate/product-client` 3928 tests pass
- `product-surfaces` `WorkflowDefinitionsSurface.test.tsx` 9/9 (the gated surface still works)
- `tsc --noEmit` on `product-client` — clean
- `check_appearance_scaling.py`, `check_max_lines.py`, `check_frontend_boundaries.py` — pass

No motion tokens were added or changed, so `motion.ts` is untouched.

New tests: session-count rendering (including the `0` and singular cases) and the placement badge in `WorkspacesCommandList`; beta-notice raise + dismissal with the surface still mounted in `WorkflowsPage`.

## Follow-up: tier-2 specs vs. the beta notice (2nd commit)

The first push turned "Workflow definition lifecycle (tier-2)" and "intent-tests (provisional)" red at `tests/intent/specs/workflow-definitions.spec.ts:88`, and it was a genuine consequence of the gate, not a flake. The interstitial is dismissible, but it is still a real modal — the overlay intercepts pointer events — so the spec's first interaction hit the notice instead of the definitions list. The surface being *mounted* behind the modal (which the unit test asserts) is not enough for a spec that drives it.

Fixed the way a user experiences it, not by weakening the specs:

- `tests/intent/stack/workflows-beta-gate.ts` — a shared `dismissWorkflowsBetaGate(page)` helper, marked `TEMPORARY`, that clicks *Continue anyway*.
- Both `workflow-definitions.spec.ts` and `workflow-runs.spec.ts` call it as an explicit first step right after navigating, with a comment saying it is a temporary beta gate. **No assertion was changed, relaxed, or skipped** — and the gate is now covered by the same tests that cover the surface.
- The acknowledgement moved from per-mount React state to `sessionStorage`, so it is once per browser session rather than once per mount. That is the better product cadence anyway (the surface remounts on reload and on every workflows route change; re-raising the same notice each time is nagging, not informative), and it means the specs' mid-test `page.reload()` calls do not need a second dismissal. Storage access is wrapped in `try`/`catch` because it throws in privacy modes and embedded webviews; the fallback keeps the notice visible, the safe direction for a beta warning.

The helper is a no-op if the gate is removed — it returns early when no notice is raised — so flipping `WORKFLOWS_BETA_GATE_ENABLED` to `false` does not break the specs.

**I could not run the tier-2 intent suite locally**: it needs a booted server, Postgres and a Playwright browser, and this machine is under a no-dev-server / no-Docker / no-browser constraint. The fix is by reasoning about the modal's pointer-event interception, and **CI is the proof**. What I did verify locally: `WorkflowsPage.test.tsx` 6/6 (including a new assertion that a remount inside the same session does not re-raise the notice), the tier-2 fixture unit tests (`tsx --test stack/**/*.test.ts`) 9/9, `playwright test specs/workflow-definitions.spec.ts --list` (collection/imports resolve, 2 tests), full `product-client` suite 3928 tests pass, and `tsc --noEmit` clean.

`tests/intent` `tsc --noEmit` reports pre-existing `TS5097` errors on `.ts`-extension imports across the whole `stack/` directory (that suite runs under `tsx`, not `tsc`); my new import follows the same existing convention and adds no new class of error.

Note on `scripts/check_design_attribution.py`: it does not exist on `main`, so it could not be run. I swept the diff manually for reference-product names instead — the only hits are on removed lines (two doc comments that named one, now de-attributed).


pablonyx:
## Independent review — #1560 @ 4bc6397e60

Verdict: this is a well-scoped, additive change that does what it says. The width change matches an existing shared measure, the row detail reuses data already fetched with correct fallbacks, and the beta gate reuses `ModalShell`/Radix `Dialog` rather than a hand-rolled overlay. I found no blocking issues. There's one real gap worth fixing (a genuine gate-bypass path via direct route params) and a couple of minor items.

### Should fix

- `apps/packages/product-client/src/pages/WorkflowsPage.tsx:105` — `betaGateAcknowledged` state (and the modal) live in `AuthenticatedWorkflowsPage`, which is remounted by React Router any time `workflowId`/`runId` params change identity, but is otherwise the same component instance across `/workflows`, `/workflows/:id`, and `/workflows/:id/runs/:runId` while params merely update. That's fine for the common flow (dismiss once, then click through to a definition — no remount, no re-raise). But a user who deep-links straight to `/workflows/abc/runs/123` on a fresh tab (no prior `/workflows` visit, sessionStorage empty) mounts `WorkflowRunsSurface` immediately alongside the modal — same as `/workflows` itself. That's consistent with the PR's stated intent ("mounts behind the modal", "gates nothing structurally"), so it's not a bug relative to the design, but it means the gate is cosmetic: nothing prevents a user (or an automated script, or a saved bookmark) from interacting with `WorkflowRunsSurface`/`WorkflowDefinitionsSurface` before ever seeing or dismissing the notice, since the overlay is CSS/pointer-events-based, not a conditional render. If the intent is truly "just a notice, no functional gate," this is fine and matches the PR body's own framing ("purely additive ... gates nothing structurally"). Flagging because the word "gate" in the PR/file naming implies more than what's implemented — worth confirming that's the accepted product decision, since a support/beta-warning surface that a keyboard/automation-driven user (or a user who tabs past it) can trivially never see is a weaker warning than the naming suggests.
- `apps/packages/product-client/src/pages/WorkflowsPage.tsx:236-245` / `WorkflowsBetaGateModal.tsx` — `ModalShell`'s `showCloseButton` defaults to `true` and isn't overridden here, so the modal renders an "X" close button in addition to "Continue anyway" / "Take me back". Clicking it calls `onClose` → wired to `onContinue` (same as the primary CTA), which is a reasonable choice, but it's silent/undocumented — worth an explicit `showCloseButton={false}` (to force a real choice) or a one-line comment noting the X is deliberately treated as "continue," since a future reader will not know that was intentional vs. an oversight.

### Consider

- `apps/packages/product-client/src/pages/WorkspacesPage.tsx` — no `WorkspacesPage.test.tsx` exists (there wasn't one before this PR either), so `buildRecencyGroups`'s wiring of `sessionCount`/`placementLabel` from `SidebarWorkspaceItemState` into `WorkspacesCommandItemView` is untested end-to-end; only the presentational `WorkspacesCommandList` component has tests with hand-built props. Low risk since the mapping is a one-line passthrough (`sessionCount: item.sessionCount`, `placementLabel: placementLabelForVariant(item.variant)`), but it's the one production code path connecting real data to the new UI and it's the piece most likely to silently break (e.g. if `SidebarWorkspaceItemState.variant` grows a new case, the `switch` in `placementLabelForVariant` is exhaustive over the current 4-member union today, so it will fail to typecheck if a case is added without a return — that's actually a good guard, not a finding).
- `tests/intent/stack/workflows-beta-gate.ts:554-560` — `dismissWorkflowsBetaGate` treats a missing notice as success (`return` in the `catch`), which is correct for "already acknowledged in this browser context," but it also silently no-ops if the gate broke and never rendered for an unrelated reason (e.g. a regression in `WorkflowsPage`). There's no assertion anywhere in the tier-2 specs that the notice *did* appear at least once per browser context — only that it's gone after the helper runs. Minor; means a regression that removes the modal entirely wouldn't be caught by these specs, only by `WorkflowsPage.test.tsx`.

### Verified

- Width: `max-w-5xl`/`px-10` on `WorkspacesPage.tsx:73` matches `PageContentFrame`'s own gutter (`px-10`, `apps/packages/ui/src/patterns/PageContentFrame.tsx:70`) and the `max-w-5xl` used by `WorkflowDefinitionList`, `WorkflowRunDetail`, `WorkflowDefinitionEditor`, and `WorkflowDefinitionsAccessScreen` — not a hardcoded pixel value, and the claim in the PR body checks out against the actual source files.
- Row detail: `sessionCount` comes from `preferredLocalWorkspace?.executionSummary?.totalSessionCount ?? null` (no new query), is `null` for cloud-only/pending entries, and renders nothing when `0` or falsy (`WorkspacesCommandList.tsx:172`, `224-233`) — no `undefined`/raw-id leakage. Both builder sites of `SidebarWorkspaceItemState` (`sidebar-workspace-items.ts`, `pending-sidebar-projection.ts`) were updated; no third builder was missed.
- `placementLabelForVariant` is an exhaustive switch over the 4-member `SidebarWorkspaceVariant` union (`local | worktree | cloud | ssh`) — no default fallthrough gap.
- Text sizes: session count and placement badge both use `text-ui-sm` (12px), consistent with the rest of the row (PR number label, ahead/behind label, date) — not the 8px `text-xs` trap. Icon uses `!icon-paired` (the token-driven paired-icon utility), consistent with the rest of the row's glyphs, not a raw `size-3`.
- No `shadow-[...]` arbitrary values and no new `dark:` variants introduced anywhere in this diff.
- Copy: the only new user-facing strings for the beta gate (`title`, `description`, `continueLabel`, `leaveLabel`) were added to `workflow-copy.ts` as `WORKFLOW_BETA_COPY`, not inlined in the component. `"Cloud"`/`"SSH"` placement labels and the `sessionLabel()` tooltip text in `WorkspacesCommandList.tsx`/`WorkspacesPage.tsx` are not routed through a copy module, but this matches existing precedent in the same files (`PR_STATUS_UNAVAILABLE_LABEL` is a local const in `WorkspacesPage.tsx`, and `SidebarRepositoriesHeader.tsx` already inlines `"Cloud"` the same way) — not a new pattern introduced by this PR.
- Modal reuse: `WorkflowsBetaGateModal` is built on `ModalShell` (Radix `Dialog` under the hood), matching every other modal in the codebase (`ProviderPickerModal`, `ApiKeyCreatorModal`, `RepoSetupModal`, etc.) — not hand-rolled. Radix's `Dialog.Root`/`Content` gives focus trap, initial focus, focus restore on close, and `role="dialog"`/`aria-modal` for free; `ModalShell` explicitly shields the desktop-global Escape handler and calls the provided `onClose` (wired to "Continue anyway" here), which is consistent with how every other `ModalShell` consumer in the repo handles Escape.
- `scripts/check_design_attribution.py` (present on this branch, absent on `main` per the PR body) passes clean against every new/changed file in this diff.
- No dead exports: `WORKFLOW_BETA_COPY` is used by exactly one consumer; the `WORKFLOWS_BETA_GATE_ENABLED` flag and `WORKFLOWS_BETA_GATE_STORAGE_KEY` are each used once, as designed for the stated one-line removal path.
- Test quality: `WorkspacesCommandList.test.tsx`'s new cases assert the `0`/`1`/`3` session-count boundary and the null/"Cloud" placement-label boundary against rendered DOM output (`getByLabelText`, `queryByText`), not just implementation echo — they would fail if the feature were removed or the `> 0` guard were dropped. `WorkflowsPage.test.tsx`'s new case asserts the surface stays mounted (`workflow-definitions` testid + the surface mock being called) while the notice is up, that dismissal removes the notice without unmounting the surface, and that a remount within the same session doesn't re-raise it — this is a real behavioral assertion, not a restatement.

PR #1561, opened 2026-07-27
Transcript rendering polish: link tone, file-type glyphs, hex swatches, path:line labels

pablonyx:
Five independent readability fixes in the chat transcript renderer.

## 1. Link tone

`--color-link-foreground` was `#339cff` in both modes. At prose weight on our dark surfaces that reads as dim body text rather than a link. Dark now carries a lighter step of the same hue, `#83c3ff`; light keeps `#339cff`, where the contrast runs the other way.

Still exactly **one** link token — no per-surface link colors. `--color-special` and `--color-highlight*` are deliberately untouched: this is about link legibility, not the accent family. Tagged `[RETUNE:color/link-tone]` and mirrored into `specs/codebase/platforms/product/design-system.md`, including the correction that the link token is the one member of the blue family that is no longer mode-independent.

## 2. File-type glyphs for referenced files

Inline references to files outside the workspace always fell back to the generic path-mention glyph. They now consult the existing single extension→visual table (`file-visuals.ts`) and show the real glyph when one exists — markdown, images (png/jpg/jpeg/gif/webp), svg, source languages — so `README.md` and `logo.svg` read as the kind of file they are wherever they appear. Only genuinely unclassifiable names keep the mention glyph. No second extension map was introduced.

## 3. Hex swatches in inline code

A `#rgb` / `#rrggbb` / `#rrggbbaa` literal inside inline code now renders a small colour chip ahead of the text. Detection is anchored to the whole inline-code span, so shebangs, issue numbers, malformed lengths (5/7 digits), url fragments, prose containing `#`, and fenced blocks are all left alone. The colour reaches CSS through a custom property rather than a hex literal in source, per the palette guard.

`MarkdownBody.tsx` was already at 506 lines, so inline code moved into a sibling `MarkdownInlineCode.tsx` module instead of growing it.

## 4. Human-readable `path:line` links

A markdown link whose destination is `/…/FORMATTING.md:7` displayed the whole machine path. It now displays `FORMATTING.md (line 7)`. **Display only** — the raw path stays the click target and the title/tooltip, and reference resolution is untouched. The label lives in the domain module beside the reference resolver so it is unit-testable without host providers.

## 5. Turn navigator (dropped)

A stacked previous/next control was tried for the transcript's top-right gutter, matching the existing floating scroll-to-bottom treatment. It never got past a permanently-`disabled`, inert pair, because a floating affordance the user can never press just reads as a promise the transcript doesn't keep. Rather than ship that, the last commit in this stack removes it: `TranscriptFloatingControls` only renders the scroll-to-bottom button now, and a test guards `[data-transcript-turn-navigator]` staying absent so the visuals don't come back ahead of the jump-behavior wiring. That wiring is the deliberate follow-up.

## Verification

No dev server, no cargo, no Docker — unit tests and typecheck only.

- `check_appearance_scaling.py`, `check_max_lines.py`, `check_frontend_boundaries.py`, `check_docs.py` — all pass
- `pnpm --filter @proliferate/design build` → `check-theme: ok (276 live tokens)`
- `pnpm --filter @proliferate/ui test` — 12 files / 30 tests pass
- `pnpm --filter @proliferate/product-ui test` — 31 files / 199 tests pass
- `pnpm --filter @proliferate/product-client test` — 641/642 files, 3933 tests pass. The one non-pass is `authenticated-markdown-cascade.test.ts`, whose assertion passes but whose vite/browser `afterAll` teardown times out; pre-existing environment flake, unrelated to this change.

New tests: swatch positives (3/6/8-digit, uppercase) plus a ten-case negative set; the extension classes transcript references depend on; the `path:line` label and basename forms; a guard that the turn navigator stays absent.


pablonyx:
## Independent review — #1561 @ bdb3ab15c1

Verdict: the four rendering claims (link tone, file-type glyphs, hex swatches, path:line labels) are each implemented correctly and match their stated scope, with good negative-case coverage on the hex detector and no duplicate glyph/link tables introduced. The PR description is stale against its own final commit (item 5), there's a real test-coverage gap on the one piece of new logic that combines path resolution with icon selection, and the light-mode link contrast — the exact regression class this repo has hit before — is retuned unchanged and still sits under the floor with no automated guard catching it.

### Should fix

- **PR body / shipped code mismatch on item 5.** The description still narrates "Adds the stacked previous/next control ... Both halves render permanently `disabled`", but the last commit in the stack (`bdb3ab15c` itself, "fix(chat): drop the transcript turn-stepper that could never be pressed") removed it entirely — `TranscriptFloatingControls` in `apps/packages/product-ui/src/chat/transcript/TranscriptRowListShared.tsx:192-224` only renders the scroll-to-bottom button, and `FullTranscriptRowList.test.tsx:172-180` asserts `[data-transcript-turn-navigator]` is `null` and zero disabled buttons exist. A reviewer reading only the PR body would review for a feature that isn't in the diff's final state; the body should say the navigator was tried and reverted, not describe it as landing.

- **No test for `FileReferenceBadge`'s glyph-selection logic.** `hasFileTypeGlyph`/`useExternalInlineIcon` (`apps/packages/product-client/src/components/workspace/file-references/FileReferenceBadge.tsx:59-66`) is the one piece of new logic in this PR that combines `pathKind`, `getFileVisual`, and the external-icon fallback — and it has zero direct tests. The PR only tests `getFileVisual` in isolation (`file-visuals.test.ts`) and `inlineFileReferenceLabel` in isolation (`path-references.test.ts`); nothing exercises the badge's actual decision of glyph vs. generic mention icon. Concrete gap: `actions.pathKind` can be `null` while `pathKindPending` is true (stat still resolving) — `hasFileTypeGlyph` is computed off `resolvedBasename`/`iconPath` regardless of pending state, so a still-resolving external reference to, say, `notes.md` will already show the markdown glyph before the workspace stat confirms it isn't a directory. Not necessarily wrong, but untested and worth a case.

### Consider

- **Light-mode link contrast is still 2.86:1, unchanged by this PR, and no guard would catch a regression here.** The PR keeps light's `--color-link-foreground` at `#339cff` (`apps/packages/design/src/tokens.ts:459-460`), which measures **2.86:1** against the light `--color-surface` (`#ffffff`) — the same number flagged in memory as "far under the 4.5:1 floor." `scripts/check_theme_contrast.py`'s `TEXT_ROLES` tuple (lines 71-77) covers `--color-foreground`, `--color-foreground-secondary`, `--color-muted-foreground`, `--color-foreground-tertiary`, and `--color-faint` — `--color-link-foreground` is not in that list, so the automated floor check that exists in this repo does not exercise the link token in light mode at all. The dark half is legitimately improved by this PR (`#83c3ff` on `#181818` measures ~9.5:1), but light was left exactly as it was, at a ratio the design-system doc doesn't defend anywhere and the contrast script doesn't watch. This isn't a regression this PR introduces, but a PR whose entire premise is "link legibility" is the natural place to have caught it, and didn't.

- **`fileReferenceBasename` on a path that is only separators or empty returns the whole (empty) string, not a basename.** `apps/packages/product-client/src/lib/domain/files/path-references.ts:230-233`: `"/".split(/[\\/]/).filter(Boolean)` is `[]`, so it falls to `?? path`, returning `"/"` unchanged. Not exercised by any test (the suite only covers `"a/b/"` → `"b"`). Low practical risk since a bare `/` or `""` shouldn't reach this function from a real reference, but the fallback exists for exactly this shape and is untested.

- **Windows `C:\path\file.ts:12` label test asserts the wrong-looking but "works" case only for one shape.** `path-references.test.ts` covers `/repo/apps/web/src/App.tsx:42:8` (posix, line+col) and the markdown-link posix case, but there's no test for a Windows absolute path with a drive letter (`C:\Users\x\file.ts:12`) even though `fileReferenceBasename`'s own test explicitly exercises backslash paths. `splitPathLineSuffix`'s regex `^(.*?):(\d+)(?::(\d+))?$` is non-greedy up to the *first* trailing `:digits`, so it happens to split `C:\Users\x\file.ts:12` into path `C:\Users\x\file.ts` / line `12` correctly (verified by hand), but the drive-letter colon is exactly the kind of case that quietly breaks under a regex tweak later, and it's untested end-to-end through `inlineFileReferenceLabel`.

### Verified

- **Link tone**: single token, no per-surface colors introduced; `#339cff`/`#83c3ff` only appear in `tokens.ts` (plus one pre-existing `color-mix` derivative in `product.css` unrelated to this change) — no hardcoded literal was added at a call site. Dark contrast genuinely improves (6.2:1 → 9.5:1 against `#181818`).
- **File-type glyphs**: confirmed `file-visuals.ts` is the single extension→visual table (grepped all `getFileVisual` call sites — `file-icons.tsx`, `ComposerFileMentionNode.tsx`, and now `FileReferenceBadge.tsx`; no second map exists). Lookup is case-insensitive (`lowerName = name.toLowerCase()`), has a `"default"` fallback for unknown/no-extension/dotfile names, and the badge's `hasFileTypeGlyph !== "default"` gate correctly falls back to the generic mention glyph for unclassifiable names.
- **Hex swatches**: `HEX_COLOR_LITERAL_RE = /^#(?:[0-9a-f]{3}|[0-9a-f]{6}|[0-9a-f]{8})$/i` is anchored to the whole trimmed string, so it correctly rejects git SHAs without `#`, 5/7-digit malformed hex, `#!` shebangs, `#define`-style/issue-number prose, `rgb()`/named colors, and hex embedded in a longer code span (`` `#00a240 fallback` ``) — verified all by direct regex execution. Fenced code blocks never reach `MarkdownInlineCode` (they route through `MarkdownCodeBlockShell`), so a hex value in a ```css fence is correctly excluded; test at `MarkdownBody.test.tsx:340-344` confirms this. The color reaches CSS via a custom property (`--markdown-hex-swatch`), not a literal, satisfying the palette guard; `check_design_attribution.py` and the palette scripts pass. The swatch span is `aria-hidden="true"`, correctly decorative. Regex is a module-level constant, not recompiled per render.
- **path:line labels**: `splitPathLineSuffix`'s regex correctly handles `file.ts:12`, `file.ts:12:5` (line+col), a bare path with no suffix, and degrades gracefully (returns the full string, `line: null`) for `file.ts:abc`, `file.ts:`, and `:12`-style non-numeric or missing-basename suffixes — verified by hand-executing the regex against all of those. The raw path (not the display label) stays the click target and tooltip (`title={actions.primaryUnavailableReason ?? rawPath}` in `FileReferenceBadge.tsx`), matching the "display only" claim.
- **Additive vs. replacement**: the old `inlineReferenceLabel`/`extractBasename` helpers in `FileReferenceBadge.tsx` were deleted, not left dangling, when superseded by `inlineFileReferenceLabel`/`fileReferenceBasename`. The old inline `<code>` JSX in `MarkdownBody.tsx` was replaced by `MarkdownInlineCode`, not duplicated. `MarkdownBody.tsx` stays within its allowlisted budget (net +19/-18 lines against the pre-PR 506-line file, landing at 525, under the 600-line hard cap and consistent with the existing `scripts/max_lines_allowlist.txt` entry); `check_max_lines.py` and `report_frontend_structure.py --strict` both pass clean.
- **`MarkdownSurface` scoping**: only one call site (`FileViewerContent.tsx`) passes `surface="file-content"`; every other `MarkdownBody` consumer defaults to `"message"`, matching the stated fail-loud-not-quiet default. `MarkdownBody.test.tsx:372-384` directly tests that the same hex literal produces a swatch under `"message"`/default and not under `"file-content"`.

No blocking findings.

PR #1562, opened 2026-07-27
Settings surface: Appearance sample block, host-true header inset, honest create-workspace copy

pablonyx:
## What

Four settings / Web-surface fixes.

**1. Appearance pane sample block.** The pane let you change mode, window zoom, UI font size, and code font size without showing you any of them. It now ends with `AppearanceSampleBlock`: one bordered panel whose top half is the UI type ramp (heading, body with inline code, secondary) and whose bottom half is a highlighted code snippet, rendered through the app's own `HighlightedCodeBlock`. It is deliberately not a diff: a diff answers "how do added and removed lines look", which isn't the question this pane asks, and the row tints and change gutter were the loudest thing in the panel, pulling the eye toward a state that only shows up while reviewing a change. Plain code with syntax highlighting and line numbers exercises exactly what the settings control — the mono family, the readable-code step, and the token palette — with nothing else competing. The mode row's three loose buttons became the shared `SegmentedControl`. The sample lives in a sibling module because the pane was already close to the component line budget.

**2. Settings header no longer assumes macOS traffic lights.** The header reserved 82px on the left unconditionally. That is only right on a host that paints macOS window buttons over our chrome; on Web (and on Windows/Linux desktop) it read as an empty hole above the nav. The predicate `MacWindowControlsSafeArea` kept privately is now a shared hook (`useHasMacWindowControls` / `useMacWindowControlsInsetClass`) built on the existing host capability check — desktop bridge present — plus the platform string. Every `pl-[82px]` site turned out to be reachable from Web too (they sit in shells `ProductClient` shares with Web, not desktop-only shells as I first assumed), so all of them now go through the shared hook instead of just the Settings header.

**3. Create-workspace popover stops saying "cloud" where cloud is the only option.** Under a `+` trigger titled "New workspace", the menu's only item on Web said "New cloud workspace" — a qualifier naming a distinction the user cannot act on. `cloudWorkspaceMenuLabel` drops it exactly when local workspaces are unavailable (cloud-only repo, or a host that cannot run local workspaces); configure/loading labels pass through unchanged. `RepoGroup` also gained `localWorkspacesSupported` so the local items are gated on host capability, not only the repo's environment kind.

**4. Desktop pointer copy.** Where Web silently lacks an option, it now states what Desktop adds, in the place the gap shows up: a footer line in the create popover and one under the add-repository entry options, both from a single `desktop-pointer-copy` module. Existing pointer strings that already read well were not touched.

## Verification

No dev server, no Rust, no Docker.

- `check_appearance_scaling.py`, `check_max_lines.py`, `check_frontend_boundaries.py`, `check_docs.py` — pass
- `@proliferate/design` build, `@proliferate/ui` / `@proliferate/product-ui` / `@proliferate/product-client` builds — pass
- `@proliferate/ui` tests 30/30, `@proliferate/product-ui` 195/195, `@proliferate/product-client` 3930/3930 assertions pass. One suite (`src/app/authenticated-markdown-cascade.test.ts`) times out in its own `afterAll` browser/vite teardown on this machine; it is environment-bound and untouched by this change.
- `product-client tsc --noEmit` and `pnpm web:typecheck` — clean
- Follow-up: `use-mac-window-controls.ts`'s own import used the package's external specifier instead of the internal `#product/...` alias every call site uses, which made it deaf to test mocks; fixed, and `CoworkWorkspaceShell.test.tsx` / `WorkspaceShellSidebar.test.tsx` re-verified green.

New tests: `cloudWorkspaceMenuLabel` label rules, and a `RepoGroup` case asserting the cloud-only popover offers a single unqualified item plus the desktop pointer.


pablonyx:
## Independent review — #1562 @ 432746f47b

Solid mechanically (host-inset hook is a real capability check, the "New workspace" label fix is real, tests are reasonable), but the PR description makes a factual claim about item 1 that the shipped code contradicts, and item 2's "left alone" claim about other `pl-[82px]` sites is also false — those sites are reachable from Web, not desktop-only. Neither is disqualifying on its own merits (the code that shipped is arguably fine), but the write-up cannot be trusted at face value, which matters a lot for a review that's supposed to verify honesty claims.

### Blocking

- **`apps/packages/product-client/src/components/settings/panes/AppearanceSampleBlock.tsx` (whole file) vs PR body.** The PR description says the sample's code half is "a real split before/after diff of a small config object, rendered through the app's own `DiffViewer`." The actual component renders a single highlighted snippet through `HighlightedCodeBlock`, and its own doc-comment says the opposite: "The code half is a plain highlighted snippet, NOT a diff... The diff's row tints and change gutter were the loudest thing in the panel [so they were removed]." This is not a stale draft description — it directly misdescribes what reviewers are meant to verify against the "no fake mock, must be the real thing" bar. Whoever wrote the PR body did not describe the code that shipped. Fix the PR body before merge; a reviewer trusting the description would sign off on the wrong component.

### Should fix

- **`apps/packages/product-client/src/components/settings/screen/SettingsScreen.tsx:539-542`** — the fix itself is correct and `useHasMacWindowControls`/`useMacWindowControlsInsetClass` (`apps/packages/product-client/src/hooks/ui/layout/use-mac-window-controls.ts`) is a real host-capability check, not a per-platform hardcode. But the PR body's claim that "other `pl-[82px]` reservations live in desktop-only shells and were left alone" is false: `StandardWorkspaceShell.tsx:258`, `WorkspaceShellSidebar.tsx:70`, `MainSidebarPageShell.tsx:59,94`, and `CoworkWorkspaceShell.tsx:130,165` are all reached through `MainPage`/`WorkspacesPage`/`WorkflowsPage` → `AuthenticatedAppHost` → `AuthenticatedProductClient`, which both `apps/web/src/WebHostApp.tsx` and `apps/desktop/src/main.tsx` mount via the shared `ProductClient` export. Web renders these with the same unconditional 82px left inset that SettingsScreen used to have — the exact "empty hole above the nav" bug the PR claims to have fixed, just left standing everywhere except Settings. If the intent really was "fix it in Settings only, follow-up ticket for the rest," say that; as written the PR claims those sites are desktop-only, and they aren't.

### Consider

- **`apps/packages/product-client/src/components/settings/panes/AppearanceSampleBlock.tsx`** — the sample renders a fixed, hand-authored snippet (`SAMPLE_CODE`) rather than anything sourced from the real product surfaces the settings actually affect (chat, diffs, editor). That's a reasonable simplification for a settings preview and it does exercise the real `HighlightedCodeBlock`/`CodeBlockTokenContent`/`useHighlightedTokens` pipeline (confirmed by reading those files — not a hand-built lookalike), so it clears the "no inline mock" bar on the component level. It just isn't the diff surface the PR body claims, see Blocking above.
- **`apps/packages/product-client/src/lib/domain/preferences/appearance-css-drift.test.ts` (new "steps independently" test)** — asserts that `UI_FONT_SCALES` and `READABLE_CODE_FONT_SCALES`, which are built by two separate pure functions each keyed only on their own index, don't affect each other. Given `buildUiFontScale`/`buildReadableCodeFontScale` never read from each other's inputs, this test cannot fail short of someone literally wiring one into the other — useful as a regression trip-wire but it isn't exercising the sample block or any rendering path, despite the doc-comment framing it as a regression for a rendering bug report.
- **`RepoGroup.tsx`/`SidebarWorkspaceContent.tsx`** — `localWorkspacesAvailable`/`localWorkspacesSupported` wiring is only tested at the `RepoGroup` prop level (`RepoGroup.test.tsx:204`) and the pure-label level (`cloud-workspace-creation.test.ts:848`). There's no test that exercises the actual integration point (`SidebarWorkspaceContent.tsx:175`, `isDesktopHost && environmentKind !== "cloud"`) end-to-end, so a future change to `isDesktopHost`'s source or to `resolveRepoGroupEnvironmentKind` could silently break the Web-only-copy behavior without failing any test in this PR.

### Verified

- `useHasMacWindowControls` (`use-mac-window-controls.ts:755-758`) is genuinely derived from host capability (`useProductHost().desktop !== null`) plus platform sniff, not a per-platform hardcode; it's a straight extraction of the previous `MacWindowControlsSafeArea` logic into a shared hook, and the old private predicate was deleted, not left as a duplicate.
- `cloudWorkspaceMenuLabel` (`cloud-workspace-creation.ts:888-899`) correctly special-cases only the `create` kind and passes `configure`/`hidden`/`loading` through unchanged, matching the PR's description; the "+ trigger titled 'New workspace'" claim about the sidebar checks out (`RepoGroup.tsx:165`).
- `SegmentedControl` (`apps/packages/ui/src/primitives/SegmentedControl.tsx`) is a pre-existing shared primitive being reused here, not new/duplicated, and is accessible (`role="radiogroup"`/`role="radio"`/`aria-checked`).
- No design-attribution or file-size violations: `check_design_attribution.py` passes on this diff, and the new/touched files (`AppearancePane.tsx` 172 lines, `AppearanceSampleBlock.tsx` 60 lines, `RepoGroup.tsx` 372 lines, `SettingsScreen.tsx` 252 lines) are all well under the 400-line structure gate.
- No `text-xs`, arbitrary `shadow-[...]`, or `dark:` variants introduced in this diff.

PR #1563, opened 2026-07-27
Sidebar behavior: hover overlay, right-panel minimum width, drag-to-collapse

pablonyx:
## What

Four related fixes to how the workspace shell's two side panels behave.

### Right panel minimum width: 260 → 380

The old floor sat below what the panel's own header can lay out. The tab bar needs to show three real tabs before it starts scrolling, and a right panel whose header scrolls at one tab is not navigable. Measured against the tab metrics in `product.css` (`.right-panel-tab-system`):

| piece | px |
| --- | --- |
| 3 tabs at the 90px legibility floor | 270 |
| 2 × 3px `--tab-gap` | 6 |
| 2 × 8px tab-bar padding | 16 |
| sticky new-tab trigger (28px square + 4px pad) | 32 |
| trailing action cluster (6px pad + two 20px buttons + 4px gap) | 50 |
| **header chrome floor** | **374** |

380 is the next clean step above that floor, and still comfortably under the ~470px a right panel of this kind habitually occupies, so it constrains nothing a real user wants.

Widths persisted under the old floor are re-clamped on read, so a window that stored 260 comes back legible rather than reopening at a width the panel no longer supports. Persisted-state migration still round-trips (covered by the existing `workspace-ui/migration` and `persistence` tests, both green).

### Dragging past the minimum collapses the panel

The threshold is **300px of raw drag width**, and it deliberately does *not* live in `clampRightPanelWidth`. Clamping's whole job is to refuse sub-minimum widths, so a clamped value can never report the gesture that means "close this". Instead a pure domain function `resolveRightPanelDragOutcome(rawWidth)` returns `{kind:"resize"}` or `{kind:"collapse"}`, and the resize hook is given `min: 0` so the gesture reaches the domain intact.

The 80px of travel between the 300px threshold and the 380px minimum is intentional resistance: the panel sticks at its minimum first, and only a clear shove past that closes it, so a user trimming the panel narrow never loses it by accident.

Collapse **reuses the existing durable `open: false`** state rather than adding a parallel flag, and leaves the last credible width persisted — reopening restores the size the user chose, not the default. A collapse also latches for the remainder of the gesture, so a jittery pointer cannot re-expand the panel from under the cursor. A non-finite measurement always resolves to a resize, never a collapse.

`use-resize.ts`'s cursor-overlay div and its `mousemove`/`mouseup` listeners are now torn down on unmount as well as on `mouseup`, so a workspace switch or teardown mid-drag can't leave a permanent full-screen overlay blocking pointer input.

### Collapsed main sidebar peeks on hover, as an overlay

The sidebar is now an always-mounted, absolutely-positioned panel inside an in-flow spacer that owns the layout width:

- **Toggling** still animates the spacer, so the content pane slides across exactly as before.
- **Collapsed**, the spacer reserves nothing and the panel hangs *over* the content instead of pushing it right — which is what an in-flow width transition would do on every pointer pass along the window edge.
- Keeping the panel mounted also stops the sidebar's list scroll position from being discarded on every toggle.
- While hidden the panel takes no pointer events and is `inert`, so it is out of the tab order and the accessibility tree and cannot trap a click; the arming strip stops accepting the pointer once the panel is up.
- It rises to the semantic overlay layer (`z-overlay`) only while collapsed; open, it keeps its `isolate` stacking context so the resize dragger stays on top.
- The peek never arms on a coarse (touch) pointer, gated behind `(pointer: coarse)`. A touch tap fires a synthetic `mouseenter` with no matching `mouseleave`, so a device without a real hover gesture had no way to close a peek it could accidentally open.

### Sidebar workspace timestamp steps down one type size

It rendered at `text-ui` (12px) — the same size as the workspace name beside it, where it competed for the eye. It is metadata about the row, so it moves to `text-ui-sm` (11px), the ramp's existing meta step. No raw px, no `text-xs`.

### Also in this PR: sidebar detail/status glyphs move to `icon-tight`

Round-4 sidebar feedback, landed alongside the panel work above: the trailing detail indicators (waiting clock, running spinner, error/warning badge) and the row's trailing plus/kebab controls rendered at `icon-compact`/`icon-paired`, which read too large against the reference sidebar once the row geometry above changed. Both now share `icon-tight` (0.875em) as one tier for the whole trailing cluster, so it reads as one row of detail rather than two competing sizes.

## Motion

No new tokens were needed. The peek and the collapse reuse `duration-enter` / `duration-exit` / `duration-panel` and `ease-out-quint` / `ease-standard`. The generated stylesheet zeroes all six interaction durations under `prefers-reduced-motion`, so using the tokens gets reduced-motion behavior for free — the peek snaps instead of fading and the panel snaps shut instead of sliding.

## Verification

All green:

- `check_appearance_scaling.py`, design-attribution guard
- `@proliferate/design` build — `check-theme: ok (276 live tokens)`
- `@proliferate/ui` 30/30, `@proliferate/product-ui` 195/195, `@proliferate/product-domain` 633/633
- `@proliferate/product-client` — 3932/3932 tests pass
- `product-client` `tsc --noEmit` clean

New domain tests in `right-panel.test.ts` cover the raised minimum, the re-clamp of a persisted 260, the threshold-below-minimum invariant, the sticky resize band, collapse below the threshold, and the non-finite case. New tests exercise a real `mousedown`/`mousemove`/`mouseup` drag through `useMainScreenRightPanel` (resize, collapse, latch, and reset-on-new-drag) and through `useResize` directly (including the unmount-mid-drag cleanup), plus a coarse-pointer test confirming the sidebar peek never arms on a touch device.

The one red test file in `product-client` is `src/app/authenticated-markdown-cascade.test.ts`: its single test passes and its `afterAll` browser/vite teardown times out. Reproduced identically on the clean base with these changes stashed — pre-existing and unrelated.

## Not done

- No visual/interaction proof in a running app: the machine's swap was critically full, so no dev server, no cargo, no Docker. Verified by unit tests and typecheck only. The peek's feel (edge-strip width, fade timing) is worth one pass with real eyes.


pablonyx:
## Independent review — #1563 @ b543b95759

Solid, well-reasoned domain work on the right-panel minimum/collapse logic (the `resolveRightPanelDragOutcome` split from `clampRightPanelWidth` is the right call, and the re-clamp-on-read for stale persisted widths is correctly implemented and tested). The sidebar hover-overlay rewrite is a clean single implementation with no leftover dual code path. The main gaps are test coverage that stops at the pure domain function instead of the actual pointer-drag integration, one undisclosed scope addition, and a couple of unaddressed edge cases the PR body itself flags as unverified.

### Should fix

- No test exercises the actual drag path end-to-end. `right-panel.test.ts` only calls `resolveRightPanelDragOutcome` directly; nothing simulates `mousedown` → `mousemove` → `mouseup` through `onRightSeparatorDown`/`handleRightPanelDrag` in `apps/packages/product-client/src/hooks/main/facade/use-main-screen-right-panel.ts:105-152`. That means the collapse-latch (`rightPanelDragCollapsedRef`), the reset-on-new-drag (`onRightSeparatorDown` setting the ref back to `false`), and the interaction with `useResize`'s `min: 0` are unverified as wired behavior — only as isolated logic. A test that fires a real drag sequence through `useResize` and asserts the panel actually closes (and that a subsequent drag on the same gesture can't reopen it) would catch a wiring regression that the current suite cannot.
- Hover peek has no touch/coarse-pointer guard. `WorkspaceShellSidebar.tsx:145-163` arms/disarms purely on `onMouseEnter`/`onMouseLeave`, with no `(pointer: coarse)` media check or touch-specific handling. On a touchscreen device running the Web client, a tap near the left edge can fire a synthetic `mouseenter` without a corresponding `mouseleave`, leaving the peek stuck open (classic sticky-hover). The PR asserts a hover-based interaction model throughout; nothing rules out this class of device.
- Undisclosed scope addition: the sidebar detail-indicator and PR-status glyph tier change (`icon-compact`/`icon-paired` → `icon-tight` across `SidebarIndicators.tsx`, `SidebarWorkspaceGitGlyph.tsx`, and ~150 lines of new tests) is not mentioned anywhere in the PR body's "What" section, which only describes four fixes (right-panel minimum, drag-collapse, hover overlay, timestamp type step). This is a real visual change to icon sizing across the whole sidebar row and deserves its own line in the description — a reviewer scanning the body would miss that it happened at all.

### Consider

- `apps/packages/product-client/src/hooks/ui/layout/use-resize.ts` (unmodified by this PR, but the mechanism the whole drag-to-collapse feature is now built on) registers `mousemove`/`mouseup` on `document` in the `mousedown` handler and only removes them — and removes the full-screen cursor-overlay `div` it appends to `document.body` — inside `handleMouseUp`. There is no unmount-time cleanup. If the workspace shell (or the component owning the separator) unmounts mid-drag — e.g. a workspace switch, an error boundary, a session teardown triggered while the user is dragging — the listeners and the fixed `inset:0` overlay div leak permanently, and the leaked overlay div would visually block all further pointer interaction until reload. This predates the PR and isn't worsened by it, but the PR's own "Not done" section admits no interactive verification happened, and this exact class of bug (leaked drag listener on unmount) is the most likely failure mode for a drag feature that only got unit-test coverage. Worth a follow-up fix given how central dragging now is to this surface.
- The right/left separators (`WorkspaceResizeSeparator.tsx`) have `role="separator"` and `aria-orientation` but no `aria-valuenow`/`aria-valuemin`/`aria-valuemax` and no keyboard handler — collapse and resize are pointer-only. Not introduced by this PR, but the PR body's "Drag-to-collapse" section presents this as a considered, load-bearing new interaction model without addressing that a keyboard-only user still cannot collapse or resize the panel at all.
- `RIGHT_PANEL_MIN_WIDTH` (380) + `WORKSPACE_SIDEBAR_MIN_WIDTH` (220) = 600px, comfortably under Tauri's `minWidth: 1024` (`apps/desktop/src-tauri/tauri.conf.json:20`), so the narrow-window overlap scenario the review brief asks about doesn't reproduce on Desktop. It's unenforced defensively in the layout itself (pure flex, no explicit floor), so if the Web client is ever embedded in a narrower host (e.g. an iframe or a resizable browser window below 600px with both panels open) nothing stops the content pane from being squeezed to zero rather than something breaking cleanly — low risk today, worth a note if Web's viewport constraints ever loosen.

### Verified

- The minimum-width clamp is centralized in one place (`clampRightPanelWidth` in `right-panel-model.ts:130-135`) and every write path — live resize (`use-main-screen-right-panel.ts:150-158`), persisted-state hydration (`persisted-right-panel.ts:59,111`), and legacy migration (`right-panel-migration.ts:57-60`) — routes through it or `normalizeRightPanelDurableState`. No CSS/JS duplication of the floor.
- Stale persisted widths below the new 380px floor are re-clamped on read: `normalizeRightPanelDurableState({ open: true, width: 260 })` correctly returns `width: 380`, and this is covered by a real test (`right-panel.test.ts:1323-1330`), not just asserted in the PR body.
- The collapse threshold (300) is deliberately kept below the minimum (380) so `clampRightPanelWidth` can never produce a value that means "close," and this invariant is itself asserted in a test (`RIGHT_PANEL_COLLAPSE_DRAG_THRESHOLD` < `RIGHT_PANEL_MIN_WIDTH`).
- Non-finite drag widths (`NaN`) resolve to a resize (defaulting to 420), never a collapse — verified both in the domain function and by test.
- The old sidebar implementation was fully replaced, not left alongside a second path — `WorkspaceShellSidebar.tsx` is a single rewritten component; no dead export or competing resize/collapse implementation was found.
- Reduced motion is handled generically through the duration-token system (`--duration-panel`/`--duration-enter`/`--duration-exit` all zero under `prefers-reduced-motion`), so the new peek/collapse animations get reduced-motion behavior for free without a new branch — confirmed against `product.css`'s reduced-motion block.
- No design-attribution or frontend-structure violations: `check_design_attribution.py`, `check_appearance_scaling.py`, and `report_frontend_structure.py --strict` all pass clean against the current tree, and `scripts/frontend_structure_allowlist.txt`'s only diff in this stack *lowers* one line-count entry and removes another outright — no allowlist ceiling was raised.
- `shadow-popover` and the `icon-tight`/`icon-compact`/`icon-paired` classes used in the diff are all token-backed utilities defined in `product.css`/`tokens.ts`, not arbitrary `shadow-[...]`/pixel values.
- The collapse-drag latch (`rightPanelDragCollapsedRef`) is reset on every new `mousedown` (`onRightSeparatorDown`) and, once tripped, correctly ignores further `onResize` calls for the remainder of that gesture, matching the "no accidental re-expand" claim in the PR body.
- The collapsed panel is `inert` and `pointer-events-none` while hidden, so it can't trap focus or clicks, and the edge-arming strip stops accepting pointer input once the panel is open/peeking — both behaviors are covered by tests in `WorkspaceShellSidebar.test.tsx`.

PR #1565, opened 2026-07-27
fix(product): rotate the shared spinner in place and unbox the header tabs

pablonyx:
**Scope note:** the title undersells this — alongside the spinner fix and the
tab unboxing, this PR also ships two new tab behaviors (tab close animates,
tab names reveal once on first assignment) and their supporting primitive
(`TypewriterRevealText`) and hook (`useHeaderTabCloseTransition`). Both are
real, wired-up features documented below, not incidental fallout.

## The spinner (the high-value half)

`apps/packages/ui/src/primitives/Spinner.tsx` is the single spinning-loader
primitive behind ~29 call sites — the composer's "waiting for GitHub", the
harness-update toast in the bottom-right, tab activity glyphs, sidebar rows.
All of them wobbled: the ring appeared to orbit a point outside itself
instead of rotating about its own centre. Fixed at the primitive, so every
call site is fixed at once and none can drift back independently. No
per-callsite patches were made (in particular `tab-rendering.tsx` and
`HarnessUpdateToastPresenter.tsx` are untouched — the toast already renders
the shared `<Spinner>`, so it inherits the fix).

### Diagnosed root cause

The correct rotation authority already existed in the generated theme:

```css
.proliferate-spinner > svg {
  animation: proliferate-spinner-rotate 1.4s linear infinite;
  transform-box: view-box;
  transform-origin: center;
}
```

…but the primitive re-declared `style={{ transformBox: "fill-box",
transformOrigin: "center" }}` **inline on the `<svg>` root**. Inline style
beats the stylesheet, and `fill-box` has no well-defined reference box on an
`<svg>` root element — the root establishes the viewport rather than being an
object with a bounding box inside one — so the rotation origin stopped
resolving against the `0 0 24 24` view box, whose centre is exactly (12,12).
An off-centre origin on a rotation *is* an orbit. The inline declaration is
removed; the origin is now owned solely by CSS, and the primitive documents
that it must never be re-authored inline.

Two contributing factors, fixed at the same place:

1. **The box could resolve non-square.** The glyph tiers are em-based
   (`--icon-paired: 1.230769em`); a spinner inside a stretching flex/grid
   parent resolves a non-square box and `size-full` propagates that to the
   SVG. A rotating non-square box sweeps an arc wider than its own
   footprint, which reads as wobble. `aspect-ratio: 1 / 1` on
   `.proliferate-spinner` pins it (an earlier revision duplicated this as an
   `aspect-square` Tailwind class on the wrapper too; dropped as redundant).
2. **Call sites could spin the wrapper.** Some passed `animate-spin` (the old
   test fixture literally did). `animation: none !important` neutralizes it
   and the primitive no longer duplicates the animation class on the SVG.

### The other candidate, ruled out arithmetically

Off-centre artwork would produce the same symptom regardless of
`transform-origin`, so it had to be excluded. Both `<path>` `d` attributes
were parsed and every cubic segment sampled densely:

| path | x range | y range | centre | size |
| track (`opacity 0.3`) | 4.000000 – 20.000000 | 4.000000 – 20.000000 | (12, 12) | 16 x 16 |
| arc | 4.000000 – 20.000000 | 4.000000 – 20.000000 | (12, 12) | 16 x 16 |

Union offset from the view box centre: **(0.0, 0.0)**. The art was never the
problem; origin resolution was.

### Confidence and what was not done

High confidence in the mechanism, derived from the path geometry and from the
CSS Transforms spec's definition of the reference boxes. **Visual
confirmation in a browser was deliberately not performed** — this machine's
memory budget does not allow a dev server plus a headless browser, and it has
OOM-died from exactly that before. No runtime measurement is claimed here.
Instead `apps/packages/ui/test/Spinner.test.tsx` pins all three invariants —
square non-rotating box, no inline rotation authority (asserted against the
primitive's own source with comments stripped), and art centred on the view
box centre — so a re-introduction fails in CI rather than in someone's eyes.

## Header tabs, unboxed

The strip read as a row of chips: fill *and* border in both the resting and
the active state, so the active tab was distinguished only by a slightly
different box. Now:

- no fill and no border at rest or active
- hover is the only fill an ordinary tab gets
- selection is a 1px underline flush with the strip's baseline
- 28px tall, `4px 8px` inner padding kept

Every affordance still works: `TabContextMenu`, `TabGroupPill`,
`ClosedChatTabsMenu`, `DelegatedAgentHoverCard`, drag reorder, multi-select,
group underlines.

## Tab close animates

The closing tab leaves the row model immediately, which is what lets the
surviving tabs translate into the vacated space; a non-interactive ghost
fades and collapses at the departing tab's last geometry over the same
duration and curve, so both halves of the motion run together.

The trailing "+" button is a flex sibling of the scroll strip, so it does not
move by transform — it moves because the strip's measured content width
shrinks. That width now transitions on the same duration/ease, so the button
travels with the tabs instead of snapping to the new edge.

`useHeaderTabCloseTransition` (the hook owning ghost geometry and timeout
bookkeeping) and `getKnownSessionHasAssignedTitle` (the derivation feeding the
reveal-once gate) now have direct unit tests: ghost add/remove timing,
same-tab double-close de-duping into one ghost and one live timeout,
independent ghosts for tabs closed together, `contentWidth` derivation, and
unmount cleanup for the hook; every `KnownHeaderSession` branch (placeholder,
slot with/without a title or transcript title, materialized session
with/without a title, whitespace-only title) for the derivation.

## Tab names reveal once

A session receiving its first real title types the label in character by
character, with the label's full width reserved so the row does not jitter
and the accessible name never announces a half-written title. Two ref latches
confine it to that one moment: a tab that mounts already titled never types,
and later renames render whole. New primitive
`@proliferate/ui/primitives/TypewriterRevealText`, registered in the
playground library sheet.

## Motion authority

- reveal cadence: `motion.activity.tabNameRevealMs` (new; a cadence, not an
  interaction, because JS drives the per-character clock)
- close choreography reuses the existing `--duration-enter` /
  `--ease-out-quint` roles through a new
  `--workspace-shell-tab-close-collapse` token
- new/changed tokens carry `[RETUNE:header/unboxed-tabs]` or
  `[RETUNE:header/tab-close-slide]`; nothing new is labelled `[SHIPPED]`
- reduced motion drops the character clock and the collapse

No hardcoded inline durations or eases were added.

## Verification

Unit tests and typecheck only — no dev server, no cargo, no Docker, no
browser.

- `pnpm --filter @proliferate/design build` — `check-theme: ok (282 live tokens)`
- `pnpm --filter @proliferate/ui test` — 13 files / 36 tests pass
- `pnpm --filter @proliferate/product-ui test` — 31 files / 195 tests pass
- `pnpm --filter @proliferate/product-client test` — **642 files / 3927 tests pass**
- `python3 scripts/check_appearance_scaling.py` — pass
- `python3 scripts/check_frontend_boundaries.py` — pass

## Out of scope, reported not fixed

The toast overlapping the status text behind it is a stacking/offset problem
in the toaster host (`apps/packages/ui/src/primitives/Sonner.tsx`), a
different surface owned by another change in flight. Not touched here.

The close ghost's `left`/`width` are frozen from `shellRows`/`positions`/
`widths` at click time, but the real tab removal happens asynchronously; a
resize or another tab closing in the same ~160ms window can leave the ghost
rendering at stale coordinates for the rest of its exit. Cosmetic only
(nothing but the ghost's own geometry is affected), not fixed here.


pablonyx:
## Independent review — #1565 @ e27557825b

Verdict: the spinner fix and the tab-unboxing CSS are narrow, well-reasoned, and match their stated diagnosis. The tab-close ghost/transition system and the `TypewriterRevealText` primitive are a different story — real, wired-up features (not dead code, contrary to what the size alone would suggest), but they are genuinely out of scope for a PR titled "rotate the shared spinner in place and unbox the header tabs," and they ship with materially weaker test coverage than the two features the title actually promises. Nothing here looks broken enough to block, but the scope and the coverage gap should be called out before this becomes the template for how "small" fixes get shipped in this codebase.

### Should fix

- **Scope creep is real, not just a diff-stat illusion.** `apps/packages/ui/src/primitives/TypewriterRevealText.tsx` (+98), `apps/packages/product-client/src/hooks/workspaces/ui/tabs/use-header-tab-close-transition.ts` (+91), and `HeaderTabsClosingGhosts.tsx` (+35) — plus the `hasAssignedTitle`/`hasAssignedLabel` plumbing through `workspace-header-tabs-view-model-derivation.ts`, `workspace-header-tabs-model-helpers.ts`, and `ChatTabWithMenu.tsx` — implement two new pieces of UI behavior ("tab close animates" and "tab names reveal once") that are not implied by "rotate the shared spinner" or "unbox the header tabs." These are legitimate, used features (I verified `TypewriterRevealText` is imported and rendered in `ChromeWorkspaceTab.tsx:148`, not dead code), but AGENTS.md says "record unrelated defects as follow-ups instead of expanding the current PR," and this PR title gives reviewers no signal that ~220 lines of new primitive/hook/component are coming. Two independently reviewable, independently revertable features got bundled into a "fix" PR.
- **Zero test coverage for the two new scope-creep features' actual logic.** `use-header-tab-close-transition.ts:1-91` (timer bookkeeping, ghost geometry, double-close handling) and `HeaderTabsClosingGhosts.tsx` have no test file at all. `getKnownSessionHasAssignedTitle` (`workspace-header-tabs-model-helpers.ts:162-176`) has no direct unit test either — the one place it's exercised, `workspace-header-tabs-view-model-derivation.test.ts:241`, hardcodes `hasAssignedTitle: true` on the test fixture rather than calling the real derivation with placeholder/slot/session variants, so none of its three branches are actually asserted. The PR body's "+171 test lines" is almost entirely `TypewriterRevealText.test.tsx` and `Spinner.test.tsx`; the hook that owns the close-animation state machine ships with no negative case (e.g., closing the same tab twice in quick succession, or closing a tab while its own previous close timeout is still pending) even though it has real mutable state (`timeoutsRef`) that a two-line "close same tab twice, assert one ghost" test would have caught if broken.
- **`use-header-tab-close-transition.ts:56-77` (`beginCloseChatTab`) resolves ghost geometry off `shellRows`/`positions`/`widths` captured in the closure at click time, but the actual tab removal happens asynchronously** via `chatVisibilityActions.hideChatSessionTabs(...)` in `HeaderTabs.tsx:274-277`. If a resize/reflow (e.g. window resize, another tab closing) changes `layout.positions`/`layout.widths` in the same tick before the removal commits, the ghost was already computed from the pre-resize layout and will render at a stale `left`/`width` for the fixed `--duration-enter` (160ms) window — a fast double-close or a close during a live-resizing strip can leave a ghost visually detached from where the real tab used to be. Low severity (160ms window, cosmetic only) but worth a code comment or a follow-up test rather than silent trust.

### Consider

- The `--workspace-shell-tab-close-collapse` token and the CSS `.workspace-shell-tab-strip__viewport` width transition are legitimate uses of the existing `--duration-enter`/`--ease-out-quint` roles and follow the same "resolved animation shorthand in a token" pattern already used by `--animate-popover-in` (`tokens.ts:39-43`) — verified this is not a smuggled literal.
- `Spinner.tsx:32` adds Tailwind's `aspect-square` utility on the wrapper *and* `product.css:129` separately adds `aspect-ratio: 1 / 1` on the same `.proliferate-spinner` class. Both resolve to the same rule; harmless but redundant — one of the two should be dropped once someone's looking at this file again.
- `TypewriterRevealText`'s reduced-motion check re-implements `window.matchMedia("(prefers-reduced-motion: reduce)")` inline (`TypewriterRevealText.tsx:97-102`) rather than reusing the CSS-level `@media (prefers-reduced-motion: reduce)` pattern used elsewhere in `product.css` (17 other call sites) — a JS-side per-character clock arguably has no CSS equivalent to hook into, so this may be unavoidable, but it's a second, independent source of truth for "is reduced motion on" that could drift from the CSS one.
- `HeaderTabsClosingGhosts.tsx:19-24` gives every ghost `key={closing-${closing.id}}` and a fixed `pointer-events: none`; fine as-is, but there's no visible handling for what happens if a user reopens (restores) a tab with the same session id while its ghost is still animating out — the ghost and the newly-restored real tab would render in the same span for up to 160ms. Purely cosmetic overlap, not a correctness bug, but worth a one-line note in the hook's doc comment for the next person who touches this.

### Verified

- The spinner root-cause diagnosis holds up: the old code passed inline `style={{ transformBox: "fill-box", transformOrigin: "center" }}` directly on the `<svg>` root (removed in this diff), which per the CSS Transforms spec has no well-defined `fill-box` reference on a root `<svg>` element; the fix correctly moves rotation authority to the existing `.proliferate-spinner > svg` theme rule and adds `aspect-square`/`aspect-ratio: 1/1` to guard against non-square stretching. This is a real, correctly-diagnosed fix, applied at the single primitive behind ~29 call sites, with no per-callsite patches.
- `TypewriterRevealText` is not dead code — it's imported and rendered in `ChromeWorkspaceTab.tsx:148` behind the new `hasAssignedLabel`/`revealOnFirstAssignment` prop, wired from `ChatTabWithMenu.tsx:136` (`tab.delegatedAgent ? true : tab.hasAssignedTitle`) and from the view-model derivation.
- `TypewriterRevealText` correctly respects `prefers-reduced-motion` (skips the character clock and renders final text), cleans up its `setInterval` on unmount/re-run (return function in the effect), and preserves the accessible name via a `sr-only` full-text span that never announces a half-written label — matches the PR body's claims.
- The tab-close ghost/collapse system does clean up its timeouts on unmount (`use-header-tab-close-transition.ts:41-48`, the `useEffect` cleanup clearing all pending `window.clearTimeout` handles) and de-dupes a repeated close of the same session id before re-arming a new timeout (`use-header-tab-close-transition.ts:69-72` filters the existing entry before pushing a new ghost, and clears any prior timeout for that id) — a same-tab double-close does not leak a timer or produce two ghosts for the same id.
- `--workspace-shell-tab-active-border`/`-selected-border` going to `transparent` and the new `::after` underline are real token changes with correct `[RETUNE:header/unboxed-tabs]` provenance tags, not hardcoded literals — consistent with the design system's provenance model.
- `scripts/check_design_attribution.py`, `scripts/check_appearance_scaling.py`, `scripts/check_frontend_boundaries.py`, and `scripts/check_max_lines.py` / `report_frontend_structure.py --strict` all pass on this diff; no file introduced or touched by this PR exceeds the 400/500-line frontend limits (largest new file is `TypewriterRevealText.tsx` at 98 lines), and no `text-xs`/`shadow-[...]`/`dark:` violations were introduced.

Nothing here is blocking.

PR #1566, opened 2026-07-27
Move the update affordance into the sidebar footer and restyle the base toast

pablonyx:
## Why

Three separate pieces of app chrome were carrying weight they shouldn't.

**The update affordance was in the top-left.** It was a wide pill in the window
chrome, threaded through three shells as five updater props. Top-left is window
chrome — a sidebar toggle and nothing else — so an availability indicator there
competed with the app's own title area and made every shell know about the
updater.

**The sidebar footer was three stacked bands.** A usage widget, a 40px account
row, and a help button, each on its own line, at the bottom of a surface whose
job is navigation.

**The usage widget was two concentric rings.** Rings encode two numbers in a
shape you have to decode before you can read either one.

**The toast looked like a different product than the popovers.** Sonner's
default opaque flat-bordered card sat next to our 90%-alpha blurred popover
chrome with nothing shared between them.

## What changed

1. **Update affordance leaves the top-left.** `WorkspaceSidebarHeaderControls`
   is now just the sidebar toggle, and `StandardWorkspaceShell` /
   `WorkspaceShellSidebar` / `MainSidebarPageShell` / `CoworkWorkspaceShell` no
   longer thread updater state. `SidebarUpdatePill` is deleted, along with its
   bespoke `update-ready-sweep` CSS, its motion value, and its token.
2. **It reappears in the footer, left of help, with help's exact footprint:**
   28px square, 6px radius, accent fill, white glyph (download icon). New
   `SidebarUpdateFooterButton` reads `useUpdater()` directly. `text-white`
   compiles to nothing under our palette reset, so this adds a
   `--color-special-foreground` token rather than reaching for a raw hex.
3. **The toast is untouched.** `UpdateToastPresenter` still owns download
   progress, per-phase copy, and recoverable errors. Because it does, the new
   control needs no progress ring and no four-state choreography — it is a
   glyph with an accessible label.
4. **The footer is one row.** The account trigger drops from 40px to 28px with a
   single-line identity (name plus muted org), and account/update/help now share
   one footprint and one baseline.
5. **Usage moves into the account popover as status rows** — label, percentage
   used, remaining — replacing the concentric meter rings. The capability gate is
   preserved exactly, and the section now owns its own separator so a deployment
   without usage metering leaves no empty band. `SidebarUsageFooter` is renamed
   `SidebarUsageSection` to match what it now is.
6. **The toast primitive wears the canonical popover frame**, imported by
   reference from `popover-surface.ts` (90%-alpha fill, hairline ring, blur, 12px
   radius) instead of restating it — a popover retune now reaches the toast.
   Padding is 12px, not 16px; type is on the closed UI ramp; the action pair is
   24px with only the primary filled; stack gap is 8px. `!` prefixes are
   load-bearing because sonner's `[data-sonner-toast][data-styled="true"]`
   selectors outrank plain utilities. The single `<Toaster>` mount in `App.tsx`
   is unchanged and callsite `classNames` overrides still win.

## Tests

- New `SidebarUpdateFooterButton.test.tsx`: phase gating, per-phase labels with
  no progress text in the control, the shared footprint classes, click behavior,
  and disabled mid-download.
- New `apps/packages/ui/test/Sonner.test.tsx`: every `POPOVER_FRAME_CLASS`
  utility lands on the toast as `!<utility>` (encoding the by-reference
  coupling), the title/description/action ramp, and a callsite override
  coexisting with the kit.
- `SidebarConsumptionCard.test.tsx` / `SidebarUsageSection.test.tsx` now assert
  zero `svg` and zero `circle[data-meter]` plus the readable row text, and keep
  the loading / unavailable / admin-managed / no-limit cases.

## Verification

`check_appearance_scaling.py`, `check_docs.py`, `pnpm --filter
@proliferate/design build` (`check-theme: ok (276 live tokens)`),
`@proliferate/ui` 33/33, `@proliferate/product-ui` 195/195,
`@proliferate/product-client` 3922/3922, and `product-client tsc --noEmit`
all pass. No dev server, cargo, or Docker was run.

## Notes

- Specs updated: `desktop-updates.md` (footer control, not a pill),
  `billing.md` and `settings/information-architecture.md` (status rows in the
  popover, not a footer ring trigger), `design-system.md` (dropped the removed
  motion row, restated the toast primitive).
- `scripts/check_design_attribution.py` does not exist on `main`, so it could
  not be run as a gate here. Verified out-of-band against a copy: the 81
  findings repo-wide are all pre-existing, none are in files this PR touches
  (the two in `product.css` are `codex-thread` class names on lines this diff
  does not touch), and this PR adds none — it also cleans one, in
  `popover-surface.ts`.
- The web/desktop move ledger is a frozen snapshot pinned to an old base SHA and
  its postmove checker is already red on `main`; the renamed/deleted sidebar rows
  are left as historical record.


pablonyx:
## Independent review — #1566 @ 467a408eb2

Verdict: the toast restyle and usage-rings-to-rows change are clean and well-tested, but the update-affordance move drops real reachability the old pill had. The old `SidebarUpdatePill` was deliberately duplicated into both the sidebar header (visible when open) *and* the collapsed content header (visible when closed) specifically so an update was always reachable regardless of sidebar state — a fact the deleted `CoworkWorkspaceShell.test.tsx` cases named explicitly ("keeps the update pill in the content header while the sidebar is hidden"). The new `SidebarUpdateFooterButton` lives only inside `SidebarAccountFooter` → `MainSidebar`, which is only rendered inside the sidebar panel itself. When the sidebar is collapsed, the update control goes with it.

### Blocking

- **Update affordance becomes unreachable when the sidebar is collapsed, on `MainSidebarPageShell`-based screens (Workspaces, Workflows).** `apps/packages/product-client/src/components/workspace/shell/screen/MainSidebarPageShell.tsx:56` collapses the sidebar to `width: 0` with `overflow-hidden` and has no hover-peek mechanism at all (peek only exists in `WorkspaceShellSidebar.tsx`, used by `StandardWorkspaceShell`). `SidebarAccountFooter` — and therefore `SidebarUpdateFooterButton` — is mounted only inside `MainSidebar`, which sits inside that width-0, clipped panel. A user who collapses the sidebar on `/workspaces` or `/workflows` while an update is `available`/`downloading`/`ready` has **no way to see or act on it** until they reopen the sidebar. The old design solved exactly this by mounting a second `SidebarUpdatePill` in the content header when the sidebar was hidden (see the deleted test at `apps/packages/product-client/src/components/workspace/cowork/CoworkWorkspaceShell.test.tsx` pre-image, "keeps the update pill in the content header while the sidebar is hidden"). That fallback has no replacement anywhere in this diff.
- **Same loss on `StandardWorkspaceShell`/`CoworkWorkspaceShell`, softened only by unreliable mouse-hover peek.** `WorkspaceShellSidebar.tsx:143` (`inert={!open && !peekActive}`) does let the collapsed panel become reachable via a hover-triggered "peek," but that is pointer-only: it requires hovering a 2px edge strip (`data-sidebar-peek-trigger`, `WorkspaceShellSidebar.tsx:155-163`) and offers no keyboard path to reveal it. A keyboard-only or touch user with the sidebar collapsed cannot reach the update control at all — a functional regression from a pill that used to render unconditionally in the always-visible top-left chrome.

### Should fix

- **Stale comment left behind in an adjacent, untouched file.** `apps/packages/product-client/src/components/feedback/UpdateToastPresenter.tsx:27` still reads "the sidebar pill owns continuous state" — `SidebarUpdatePill` no longer exists after this PR; the referent is now `SidebarUpdateFooterButton`. Not touched by this diff, so it's a leftover the PR should have caught since it explicitly renamed the concept everywhere else.
- **`text-white`/`--color-special-foreground` provenance tag looks miscategorized.** `apps/packages/design/src/tokens.ts` adds `--color-special-foreground` with `provenance: "[RETUNE:color/special-foreground]"`, but this is a wholly new token filling a real gap (no prior paired foreground existed for `--color-special`), not a retune of an existing value — worth double-checking against whatever the provenance taxonomy actually distinguishes (new token vs. retuned token), since the PR body is at pains to justify every other provenance tag.

### Consider

- `SidebarUpdateFooterButton` and `SidebarHelpFooter`'s trigger both use `variant="unstyled"` with no `focus-visible:outline` classes, unlike `Button`'s `sidebar`/`sidebar-link` variants which bake one in. This is a preexisting gap (confirmed via `git show design/header-tabs-spinner:...SidebarHelpFooter.tsx` — never had one either), so it's not a regression this PR introduced, but the PR was already touching both files and could have closed it while making the two match pixel-for-pixel.
- The deleted `SidebarUsageTrigger` had an explicit `focus-visible:outline` ring; that entry point is gone now that usage moved into the account popover's menu content. Probably fine since the popover items presumably inherit standard menu-item focus styling, but worth a quick keyboard-tab pass since it wasn't verified here.

### Verified

- No behavior lost via the 328 deleted test lines beyond what's called out above: the deleted `SidebarUpdatePill.test.tsx` assertions (phase gating, per-phase labels, disabled-while-downloading, click routing, progress-ring dashoffset, one-shot ready sweep) map cleanly onto either the new `SidebarUpdateFooterButton.test.tsx` (phase gating/labels/click/disabled — progress ring and sweep are deliberately dropped features, not silently-untested ones, matching the PR's stated "no progress ring, no four-state choreography" design) or are simply gone because the feature (progress ring, one-shot sweep animation) was intentionally removed along with its CSS/token/motion plumbing (`update-pill-ready-sweep`, `--activity-update-ready-sweep`, `updateReadySweepMs`) — confirmed fully deleted with `grep`, no dangling references.
- `SidebarConsumptionCard`/`SidebarUsageSection` test rewrites correctly swap ARIA-label assertions for text-content assertions; the capability-gate-hides-the-section and loading/unavailable/admin-managed/no-limit cases are all still exercised.
- No direct `sonner` imports exist outside `apps/packages/ui/src/primitives/Sonner.tsx`, and no callsite currently uses the `classNames` escape hatch in production code — the `[&&]:!` override pattern is exercised only in the new `Sonner.test.tsx`, consistent with the PR calling it a preserved escape hatch rather than an active one.
- `python3 scripts/check_design_attribution.py` passes clean on this diff (no findings introduced), and `python3 scripts/report_frontend_structure.py --strict` reports zero violations (all changed files well under 400 lines).
- `scripts/appearance_scaling_baseline.json` was correctly pruned of the deleted `SidebarUpdatePill.tsx` entries rather than left stale.
- The toast restyle is global (single `<Toaster>` mount, no per-surface variants), matching the PR's claim, and correctly uses `!`-prefixed utilities to outrank sonner's own `[data-styled="true"]` selectors.

PR #1567, opened 2026-07-27
Composer inline menus: rebuilt slash-command menu and @ file mentions

pablonyx:
## What

The composer had one inline menu (slash commands) with its own ad-hoc row markup, and `@` was ordinary prompt text. This gives the composer a single inline-menu vocabulary and makes `@` a real workspace file picker whose selection lands in the draft as a chip.

### Menu chrome

Both menus now render through one panel/row pair (`ComposerInlineMenu`) built on the canonical popover frame, so an inline composer menu and a floating dropdown are recognizably the same object and only the anchoring differs (the menus portal into the composer's normal-flow overlay host, so they are full composer width, `mb-2` off the input's top edge).

Rows are the ruled 28px-minimum geometry with 10px/5px padding and an 8px radius, capped at a 320px scroller (~10 rows visible). Typographic ranking replaces the old flat row: the command name carries the control weight at reading size, the description drops a size step into the muted role, and the argument hint trails at that same muted step. No bolds — at 11-12px, weight plus color carries the whole hierarchy. The leading slash is its own muted glyph so a row reads as "name, after a slash" rather than one undifferentiated token. The highlight uses the hover paint rather than the selected paint, because it tracks the keyboard position rather than persisting a selection.

Highlight and keyboard state are shared by one hook (`use-composer-menu-navigation`), which is what keeps the two menus from drifting apart.

### `@` mentions search files

Typing `@` opens a workspace file search. The `@` must open a token, so `user@host` and email addresses never trigger it. Search **reuses the existing debounced workspace file-search path** the command palette already uses (`useWorkspaceFileSearch` → the runtime file-search query) — no new backend endpoint and no second query path. Results are re-ranked basename-first, because the user is typing a file name rather than a path, and paths that cannot be expressed as a workspace-relative link are dropped rather than offered as a broken insertion. The directory absorbs the row's remaining width and truncates, since it is what disambiguates same-named files.

Slash and mention triggers are resolved in one place so the composer can never open both panels at once.

### Selection inserts a markdown file link, shown as a chip

Selecting a result inserts the markdown file link form the prompt serializer already understands (`[name](path)`), and the draft shows it as a chip instead of raw markdown.

The chip is a token-mode text node, not a decorator: the caret walks past it, markdown export needs no special case, and it deletes as the one thing the user picked. Insertion builds the node directly rather than typing markdown, because markdown typed in one shot never reaches the markdown shortcut plugin. The import transformer is scoped to workspace-relative destinations only, so a web link typed or pasted into the draft still stays literal text (existing behavior, still covered by test).

This stays strictly on the composer side of the boundary — no transcript markdown rendering or transcript file-type icons are touched.

### Cleanup

Deletes the slash-highlight component, its hook, and the slash-command-recognition module, which had no remaining consumers.

## Verification

- `python3 scripts/check_appearance_scaling.py` — pass
- `python3 scripts/check_frontend_boundaries.py` — pass
- `pnpm --filter @proliferate/design build` — pass (276 live tokens)
- `pnpm --filter @proliferate/ui test` — 30 pass
- `pnpm --filter @proliferate/product-ui test` — 195 pass
- `pnpm --filter @proliferate/product-domain test` — 633 pass
- `pnpm --filter @proliferate/product-client test` — 3944 pass

New coverage: mention trigger detection, basename-first ranking and path filtering, menu-trigger mutual exclusion, the inserted markdown form, chip rendering on both insertion and restored-draft import, and playground scenarios for the populated and empty mention states.

No motion tokens changed, so `motion.ts`/`tokens.ts` are untouched.


pablonyx:
## Independent review — #1567 @ 236cd52c24

Solid, well-factored piece of interaction code: the shared `ComposerInlineMenu` chrome and `use-composer-menu-navigation` hook genuinely eliminate the duplication between the two menus, the old slash-highlight/recognition code is cleanly and completely deleted, design tokens are used correctly throughout (no `text-xs`, no arbitrary `shadow-[...]`, no `dark:` variants), and the mutual-exclusion trigger resolution is simple and well tested. The main gaps are accessibility (no ARIA roles at all on a custom combobox-like widget) and a couple of untested edge cases in the mention trigger (code fences/inline code, and a fragile-looking guard in `handleCommandKey`).

### Blocking

None. I did not find a correctness bug that would ship broken behavior to users in the primary paths (keyboard nav, trigger detection, insertion, chip rendering, cleanup of the old implementation).

### Should fix

- `apps/packages/product-client/src/components/workspace/chat/input/ComposerInlineMenu.tsx:872-929` (`ComposerInlineMenuRow`) — `aria-selected` is set on a plain `<Button>` with no `role="option"`/`role="listbox"` ancestor and no `aria-activedescendant` on the composer's `contenteditable`. `aria-selected` is only valid ARIA on elements with `role` of `option`/`row`/`tab`/`treeitem`/`gridcell`, so as written this attribute is ignored by assistive tech and the highlighted row is never announced. Given interaction-heavy code is explicitly called out for review, this is worth fixing: add `role="listbox"` to the panel, `role="option"` to each row, and `aria-activedescendant` pointing at the highlighted row's id from the composer's editable root (or explicitly document this as a known/accepted gap, since focus intentionally stays in the textarea).

- `apps/packages/product-client/src/lib/domain/chat/composer/file-mention-search.ts` (`findFileMentionTrigger`) — the trigger opens on any `@` that starts a whitespace-delimited token, with no exclusion for markdown code spans/fences. Typing `` `git log user@host` `` inside backticks, or an `@` inside a fenced code block pasted/typed into the composer, opens the file-mention menu exactly like prose would. The PR description only promises `user@host`/email exclusion (handled), not code-context exclusion — worth at least a test/decision on whether this is intended, since a composer that's also used to paste shell snippets or code will trigger the menu on `@` used in decorators, npm scopes (`@proliferate/ui`), etc. `@proliferate/ui` typed anywhere would open the mention menu mid-typing every time.

- `apps/packages/product-client/src/components/workspace/chat/input/ComposerCommandEditor.tsx:203-219` (`handleCommandKey`) — the guard `activeTrigger.kind !== (slashTrigger ? "slash" : "mention")` compares a freshly recomputed trigger's kind against the *stale-render* `slashTrigger` flag rather than against `trigger?.kind` directly. It happens to be safe today only because `!openMenu` earlier in the same condition already excludes the case where `trigger` is null, but it's a confusing way to express "the trigger didn't change kind between render and this keystroke," and a future edit to this condition (e.g. reordering the `||` clauses) could silently break it. Recommend simplifying to compare `activeTrigger.kind !== trigger?.kind` directly, which says what's intended and removes the indirect ternary.

### Consider

- `apps/packages/product-client/src/hooks/chat/ui/use-chat-file-mention-menu.ts:23` — `runtimeReady` for a *local* workspace is just `materializedWorkspaceId !== null`, with no check that the local runtime/file index is actually available yet (unlike the cloud branch, which checks `phase === "ready"`). If a local workspace's file index isn't warm yet, `useWorkspaceFileSearch` will fire a query that may return partial/empty results with no distinguishing status — the empty-state message ("No matching files.") will look identical to a genuine zero-result query. Minor, since this mirrors pre-existing behavior in the command palette, but worth a beat of thought since this PR is the one making `@` a first-class trigger.
- `apps/packages/product-client/src/lib/domain/chat/composer/file-mention-search.ts:76` (`matchRank`) — ranking is a linear scan with `startsWith`/`includes` over up to 40 raw candidates on every keystroke (debounced at 120ms upstream, so not a hot loop, but no memoization keys off `search.results` identity beyond React's `useMemo` on `[open, search.debouncedQuery, search.results]` — fine for 40 items, but flag if `MENTION_SEARCH_LIMIT` is ever raised).
- Test quality is good overall (composer-menu-trigger.test.ts, file-mention-search.test.ts cover real edge cases: emails, out-of-range offsets, path normalization, dedup, elision budget) but there is no test exercising Escape-closes-without-inserting end-to-end through `ComposerCommandEditor`, nor a test for Up/Down wrap-around behavior (the navigation hook clamps rather than wraps — `Math.max(0, Math.min(...))` — confirmed by reading the code, not by a test). Since wrap-around was explicitly called out as something to check, and the implementation deliberately does NOT wrap (arrow-down at the last row does nothing rather than cycling to the first), it'd be good to have a hook-level test pinning that as intended behavior rather than an oversight.

### Verified

- Old slash-command implementation (`ComposerSlashHighlight`, `use-slash-command-highlight`, `slash-command-recognition`) is fully deleted; grepped the whole source tree (excluding `dist/`) and found zero remaining references or imports.
- Shift+Enter (newline in lists via `KEY_ENTER_COMMAND` list check) and Tab (list indent/outdent) are unaffected when no menu trigger is active — `onCommandKey` returns `false` and falls through to normal editor behavior; menu keys only intercept when a trigger is open.
- `@` does not trigger inside `user@host` or `pablo@example.com` — covered by `file-mention-search.test.ts` and confirmed by reading `findFileMentionTrigger`'s token-boundary logic (the `@` must start a whitespace-delimited token).
- Slash and mention triggers are structurally mutually exclusive (`findComposerMenuTrigger` tries slash first, only falls back to mention) — verified by test and by reading `composer-menu-trigger.ts`.
- Selecting a mention result replaces exactly the trigger token (`start`..`replaceEndFor(end)`, absorbing one trailing space) via `replaceComposerRangeWithNodes`, not adjacent text.
- Insertion places the caret correctly: the file-mention chip is a `TextNode` in `token` mode with `canInsertTextBefore/After` returning `false`, so typing immediately after selection produces ordinary text rather than extending/corrupting the chip; a trailing space node is inserted so the caret lands after the chip, ready to keep typing.
- Chip round-trips through markdown export unchanged (`ComposerRichTextEditor.test.tsx`'s new "serializes a file mention chip back to its markdown link unchanged" test), and a pasted/typed web link (`https://...`) stays literal text since `WORKSPACE_FILE_LINK_BODY`'s regex excludes URL-scheme destinations via the `:` guard.
- No arbitrary `shadow-[...]`, no `dark:` mode variants, and no `text-xs` anywhere in the new/changed menu-row markup; rows correctly use `text-ui` (primary/name) and `text-ui-sm` (secondary/muted) tokens. `python3 scripts/check_design_attribution.py`, `scripts/check_max_lines.py`, and `scripts/report_frontend_structure.py --strict` all pass against this diff, and `scripts/frontend_structure_allowlist.txt` is untouched by this PR — no line-count raise to hide file growth.
- Async file search reuses `useWorkspaceFileSearch`/`useSearchWorkspaceFilesQuery` (an SDK react-query-backed hook), which owns debouncing (120ms) and request keying by `workspaceId`+`debouncedQuery`; no manual out-of-order-response handling was added or needed in this PR, and no obvious missing-abort-on-unmount issue since the query hook itself is responsible for that.
- Explicit empty/loading/error/blocked states exist for the mention menu (`mentionStatusMessage`: not-ready runtime, empty query, error, loading, and true "no matching files" are all distinguishable copy) — not a silently blank panel.
- Row `onMouseDown` calls `event.preventDefault()`, keeping focus in the composer's textarea rather than moving it to the menu on click.

PR #1569, opened 2026-07-28
feat(product): harness settings pane v2 — method cards, merged status badge, providers modal

pablonyx:
## Summary

Implements the design-handoff v2 harness settings pane (`reference/design-handoff/README.md`):

- **Identity header** — icon tile + name/vendor line + Docs exit; the "Harness settings" section title removed (rows are self-labeled).
- **Authentication** — the state is said exactly once: one status badge merged into the section header (the separate Status section is deleted); method cards (`repeat(N,1fr)`), per-method refresh.
- **API key detail** — segmented `Paste key | Saved keys | {config kind}`; when a key is bound, one field with redacted mono hint + unbind.
- **CLI detail** — nothing when authenticated; otherwise one Authenticate button → inline login terminal.
- **OpenCode = providers only** — no chooser, no gateway UI; models.dev marks (lazy-loaded logo map, mono invert) + a management modal (searchable accordion, Remove/Replace, view-all footer). Save = two writes (vault key → selection) with revoke-on-PUT-failure.
- **Models** section de-badged to one count + provenance line.
- **registry.json** docsUrl fixes: claude/codex/cursor/opencode/grok now point at real vendor docs (all verified 200). Note: the prebuilt dev runtime binary carries the old URLs until a Rust rebuild (`include_str!`).
- **Qualification** — `chat-authroute.ts` updated to the new DOM contract (`data-harness-selected-route` token list on opencode; route union `cli`). Not yet run live.
- Deleted dead components (KeyPicker, ProviderConfigCreatorModal, HarnessAuthDetailsSection, HarnessAuthGatewayDetails, HarnessAuthApiKeyRow, HarnessStatusRow) and ~40 dead copy keys.

## Verification

- product-client vitest 3940/3940 (harness suites 66/66, incl. rewritten HarnessPane 34, ProviderPickerModal 17, AllModels 8)
- tsc clean (product-client + tests/release)
- catalog validator + `cargo test -p anyharness-lib registry`/`bundled` green
- login JS budget 483,816/485,000 B (logo map dynamic-imported in both consumers)
- appearance-scaling + max-lines guards green

## Known follow-ups

- OpenCode gateway source has no UI toggle (spec says providers-only); reachable only via pre-existing selections/API.
- "Get an API key" is plain text — no per-provider console-URL registry yet.
- Bedrock/typed-config binding needs a server slice (loosen `_assert_keys_usable` + selection CHECK migration) before the client segment can light up.


pablonyx:
## Agent review (3 reviewers: logic/state · contracts/tests · design/a11y) — findings + fixes

Fixes landed in 0717bbd0f:

**Blockers (fixed)**
1. **Replace flow was broken 100% of the time** — the modal's remove-then-submit fired two racing PUTs from one render's closure; the second always collided on the server's duplicate-source check, leaving the provider unbound and the fresh vault key revoked. Now `addBoundApiKey` swaps the row on the same env var in a single commit/PUT; the modal never calls remove on replace.
2. **Failed saves were silent** — the row collapsed eagerly on Save, so the inline `role=alert` error had nowhere to render. Now the row collapses only on a successful submitting→idle transition; failures stay expanded with the message.

**Major (fixed)**
3. `handleSingleSourceSelect("api_key")` fired two racing PUTs (gateway off + row on) — now one commit.
4. Single-source inline paste leaked an orphaned vault key when the selection PUT was rejected — now revoke-on-failure, matching the providers section.
5. `handleProviderRemove` could unbind a sibling provider's row via the shared-env-var fallback — exact `provider_hint` match now wins first.
6. "Get an API key ↗" looked interactive but wasn't (no href, not focusable) — arrow/link styling dropped until a console-URL registry exists.

**Cleanup (fixed)**: dead multi-source exports, inert vi.mock of deleted HarnessAuthDetailsSection, providers summary tiles now `aria-hidden`.

**Known / deferred (not fixed here — flagging for the merge decision)**
- **OpenCode gateway orphan (major)**: a scope that already had `gateway` enabled pre-v2 has no UI to turn it off (spec says providers-only) and the badge unconditionally reads Authenticated. Reachable only via API today.
- Hint-less legacy rows on shared env vars (`GITHUB_TOKEN`, `ZHIPU_API_KEY`, …) resolve to whichever registry provider lists the var first — display ambiguity only after fix 5.
- `pendingMethod` isn't cleared on a rejected commit (card can stay highlighted for a refused selection); `capabilities.gatewayEnabled === false` renders as "Enrollment failed" even for never-enrolled accounts; disabled-but-bound rows still render the bound-key field; typed secret lingers in modal state after collapse (never displayed/submitted).
- Spec deltas: no eye/reveal on a just-pasted key (pasted keys become vault keys immediately — likely structurally intentional, confirm); identity tile is size-10/rounded-lg/bg-surface-control vs spec's 42px/10px/foreground-5 (appearance-scaling guard mandates the semantic vocabulary).

**Clean checks**: all qualification selectors in chat-authroute.ts match producers; data-attribute contract intact both ways; zero dead/dangling copy keys or component references; registry.json diff = docsUrl-only, valid JSON; **no secret-leak paths found** (both paste inputs `data-telemetry-mask` + password type; nothing logs/stringifies the value).

Verification after fixes: product-client vitest 3942/3942, tsc clean, appearance + max-lines guards green.

PR #1571, opened 2026-07-28
docs(billing): re-cut billing.md to target-state launch laws

pablonyx:
## Summary

Docs-only re-cut of `specs/codebase/platforms/product/billing.md` from current-state (176 lines) to **target-state** (~525 lines), per the 2026-07-28 billing launch-hardening rulings. Companion to the overnight hardening pass (plan: `codex/billing-launch-hardening-plan.md`).

- **Rulings inlined:** the org always pays, spend attributed per-user; fail closed loudly (deny + durable receipt + alert) on billing-read errors; #1564 org-only sandbox-subject migration stays a documented deferral, not a fix.
- **Laws:** A1–A3 (attribution/idempotency), N1–N6 (enforcement incl. fail-closed N6), M1–M4 (Stripe authority, exactly-once grants, seat convergence, in-period-only export), T1–T2 (UI truth).
- **New sections:** Fences, Seats, Team checkout, Money-in, Truth surfaces, and a **Corridor** section giving each launch assertion (B1–B4, E1–E6, W1–W4, U1–U3) a stable ID with its pinning test where one exists.
- **Current gaps** now lists six deltas: B1 sandbox-value attribution hardcode (ruled deferral), E5/E6 missing 402-cache test + receipt/alert wiring, U3 (lands via #1570), M2 live-proof debt, Web/Mobile limits pane, no usage rollup table.

All 14 pinning-test references verified to exist by name+file. `check_docs.py` (228 files) and `check_max_lines.py` pass.

## Test plan
- [x] `python3 scripts/check_docs.py`
- [x] `python3 scripts/check_max_lines.py`
- [x] every referenced test name grep-verified in its claimed file

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1516, opened 2026-07-26
feat(cloud): gateway policy gate — cache invalidation, per-access billing, wake→ensure collapse, blocked-screen rewire

pablonyx:
Fix-stack PR C8 — the last of the cloud-sandbox program (contracts: `sandbox-gateway.md` wire laws + cache gap, `sandbox-access.md` subject-layer 402, `sandbox-lifecycle.md` wake-verb gap, `cloud-workspace.md` blocked mode). **Stacked on C7 (#1509)** because it rewires the blocked screen C1/C7 reshaped; retarget down as the stack merges.

**Server — gateway access resolution:**
- The 60 s per-user coordinate cache now has a real invalidation seam (`invalidate_cloud_sandbox_gateway_access_for_user`) called on every clearing transition: sandbox destroy, provider-missing detach (connect / resume-acceptance / materialization-failure paths), rematerialization reset. Previously only a test hook could clear it — a destroy/recreate left the proxy forwarding to a dead upstream for up to 60 s.
- Billing is now checked on **every** gateway access, not only cache misses (the old blind window was 60 s). Allow decisions cache for 5 s (the authorizer builds a full billing snapshot with budget windows); **a 402 is never cached**.
- Paused sandboxes were already forwarded (their stored address stays valid — E2B wakes under traffic); this PR pins that with a unit test so the policy-not-liveness law is tested, per the gateway spec.

**wake → ensure (hard rename, pre-launch ruling, no alias):** route, service alias, SDK + sdk-react bindings, and the OpenAPI operation deleted; release scenarios (t3-bill-2/4, t3-prov-2, t3-sec-mat-1, t4-cloud-1, selfhost-qual-1, fixtures) switched to ensure.

**Client — un-deadening the blocked screen:** C1 deleted the always-null `actionBlockKind` trigger; this PR wires the real one. The AnyHarness TS client error parser now preserves the FastAPI nested `detail` body (additive `AnyHarnessError.details` — the access spec's law says the 402 body is the subject layer's one representation, so the transport must not destroy it). The connection bootstrap classifies `billing_credits_exhausted|billing_start_blocked` into the existing blocked-screen presentation, mapping `detail.reason` onto the `CloudStartBlockReason` copy; the flat retry never retries a 402 (pinned).

Spec strikes: gateway cache-invalidation gap, lifecycle wake-verb gap, cloud-workspace blocked-mode gap; access spec references updated.

[...]

PR #1501, opened 2026-07-26
chore(cloud): dead-code deletion sweep (actionBlockKind, dead billing authorizers, dead command kinds, broken automation stage)

pablonyx:
Fix-stack PR C1 (cloud-sandbox program). Coordinated deletion of verified-dead code; every target was independently verified against `main` before deletion (grep for producers/consumers, migration history, live-constraint check).

**Deleted:**
- `action_block_kind`/`action_block_reason` — the server serialized them always-`null` on both workspace models while the client *branched on them* (unreachable blocked-screen trigger, runtime-block hook, product-domain `"blocked"` producers, sdk-react freshness scoring). Server fields, SDK types (openapi regenerated, not hand-edited), and every client branch + ~15 test fixture sites removed. The blocked-mode presentation copy (`CloudStartBlockReason`, seven-reason titles) **survives** — it reads `billingPlan.startBlockReason`, a live field, and gets rewired to the 402 body in C8.
- `authorize_sandbox_start`, `authorize_sandbox_start_for_billing_subject`, `record_sandbox_start_authorization`, `SandboxStartAuthorization` — dead since #823; the live gate is `CloudSandboxResumeBlockedError` (resume-blocked path). Docstring/comment references updated.
- `configure_git_identity` + `prune_workspace_worktree` command kinds — no producer, payload, or consumer. **`ensure_repo_checkout` stays**: recon found it live (the automation pipeline enqueues it via `stages/workspace.py`), contradicting the content spec's gap text — the spec is corrected in this PR. **No DB migration**: the `cloud_commands` table itself was dropped by `d4e6f8a0b2c4`, so the enum members die code-side only.
- `stages/git_identity.py` — unimportable since #823 (imports a deleted package); its import/call removed from `pipeline.py`; stale `target_git_identity.models` pyproject entry removed.
- `db/store/automation_cloud_workspace_claims.py` — fully orphaned (zero importers, imports a deleted model).
- `_runtime_status` dead branches — `provisioning`/`stopped` literals can never appear (`CloudSandboxStatus` has exactly `creating|ready|paused|error|destroyed`).

**Spec gap items struck** in the same PR: access ×3 (actionBlockKind, authorize_sandbox_start, runtime-status dead branches — the last also un-cross-listed from lifecycle), content ×2 (prune kind; git-identity item rewritten to reflect the stage/kind deletions and the ensure_repo_checkout correction).

[...]

PR #1500, opened 2026-07-26
feat(cloud): pin managed worktrees root + enable retention in sandbox launch env

pablonyx:
Fix-stack PR C2 (cloud-sandbox program; spec: `specs/codebase/platforms/product/sandbox-content.md`).

Two env vars added to `build_runtime_env` (the cloud AnyHarness launch env):

- `ANYHARNESS_WORKTREES_ROOT=/home/user/workspace/worktrees` — pins the runtime's managed worktrees root to where cloud worktrees actually live. Without it the runtime defaults to `/home/user/.proliferate/worktrees`, so every cloud worktree sits outside the managed fence: retire refuses them ("outside managed worktrees root") and retention skips them.
- `ANYHARNESS_ENABLE_AUTOMATIC_WORKTREE_RETENTION=1` — turns on the retention service's auto-run cloud-side. The existing `ANYHARNESS_DEFER_STARTUP_RETENTION=1` stays, so the startup pass remains deferred.

Strikes the two launch-env gaps in sandbox-content.md's Current-gaps list.

**Propagation note:** existing sandboxes keep the old env until their next runtime relaunch (materialization repair); no backfill needed.

**Tests:** new `server/tests/unit/test_runtime_bootstrap_env.py` asserts both keys/values. Full server unit suite: 1171 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1572, opened 2026-07-28
fix(billing): fail closed with receipt + alert on billing-read errors; pin E5 no-denial-cache

pablonyx:
## The laws

From `codex/billing-launch-hardening-plan.md` (ruled by Pablo 2026-07-28):

- **N6** — a billing-state READ failure on an enforcement path must **fail closed**: a typed deny, never an implicit allow, **with** a durable receipt **and** an alert. Never silent in either direction.
- **N5 / corridor E5** — denials are never cached. A repaired subject succeeds within one TTL.
- **Corridor E6** — a billing-read exception on the authorization path ⇒ typed deny + durable receipt.

## What was wrong

**E6 was a partial gap.** An exception while resolving the paying subject or the billing snapshot in `assert_cloud_sandbox_resume_allowed_for_owner` (and the `authorize_sandbox_start*` family) propagated uncaught. The request failed, which is the correct *direction*, but there was **no durable receipt and no alert** wired to that path — so an enforcement outage was invisible to operators and unforensicable after the fact, and the client got an untyped 500 it could not act on.

**E5 had no dedicated test at all.** Nothing pinned that a 402 denial is a verdict about *right now* rather than a cached state.

## What changed

`server/proliferate/server/billing/authorization.py`:

- **New `BillingStateUnavailableError`** — code `billing_unavailable`, status **503**. Deliberately *not* a 402: a 402 asserts a billing verdict we do not have, and would send the client to a top-up screen for what is really a platform fault. 503 + a distinct code says "unreadable, retry", matching the other retryable-dependency 503s in the cloud services. `extra_detail` mirrors the 402's machine-readable shape so clients can branch on `detail.decision_type`.
- **New `BILLING_DECISION_READ_UNAVAILABLE`** decision type (`billing_read_unavailable`) so an operator can tell "billing said no" apart from "billing was unreadable" directly in `billing_decision_event`, without joining logs.
- **`_deny_unreadable_billing_state`** wraps the billing-snapshot resolution in the resume gate — the one **live** enforcement gate — and, defensively, in the `authorize_sandbox_start` / `authorize_sandbox_start_for_billing_subject` pair. To be precise about the blast radius: **those two start-auth entry points currently have zero live callers** (orphaned since #823, per `specs/`), so they are guarded pre-emptively for whenever they are re-wired; the behavior change that ships today is on the resume path. It fires `report_critical` (the existing critical-reporting facility used by `materialization/runner.py`), writes the receipt, then raises.
  - **Only the denial is unconditional.** Both the alert and the receipt swallow their own failures and log instead — an observability outage must never become an enforcement outage.
  - **The receipt writes on its own session**, not the caller's. Two reasons, both load-bearing: the caller's session is the one that just blew up (it may be in a failed transaction), and the production caller `materialization/runner._run_with_fresh_session` rolls back on exception — the same hazard that already makes the quota gate `commit()` its audit row before raising.
  - A read failure that strikes *before* the subject resolves still receipts, under the all-zero sentinel subject id (the column is NOT NULL; law N6 says the receipt is the point).
- Typed `ProliferateError`s pass through unrelabelled: they are already decisions, not unreadable reads. This keeps the existing 402 contract byte-identical.

Receipt mechanism reused: `record_billing_decision_event` → `billing_decision_event`, the same durable audit row the quota gate already commits before raising. No new table, no migration.

## Test plan

New file `server/tests/integration/test_billing_fail_closed_read_errors.py` (6 tests, corridor IDs in docstrings):

**E5**
- `test_billing_denial_is_never_cached_repair_allows_next_call` — exhausted subject denied 402, repaired with a refill grant, allowed on the immediately-following call in the same test body with no sleep.
- `test_hold_release_allows_next_call_without_denial_window` — the admin-hold variant.

**E6**
- `test_billing_read_failure_denies_with_receipt_and_alert` — asserts all three obligations separately: typed 503 deny (not an allow, not the bare `RuntimeError`), durable receipt surviving the caller's rollback, alert hook invoked.
- `test_billing_read_failure_still_denies_when_receipt_write_fails` — receipt is best effort; denial and alert are not.
- `test_billing_read_failure_still_denies_when_alert_hook_fails` — an observability outage still denies, and still receipts.
- `test_billing_read_failure_receipts_sentinel_subject_when_resolution_fails` — pre-resolution failure receipts under the sentinel subject.

Verified the E6 tests do not pass against the pre-fix code (the typed error did not exist).

### Commands run

```
cd server && .venv/bin/python -m pytest tests/integration -k "billing" -q
# 111 passed, 792 deselected  (baseline on 2506df6ab was 99; +12 new)

cd server && .venv/bin/python -m pytest tests/unit -q
# 1225 passed

cd server && .venv/bin/ruff format --check proliferate/ tests/   # 863 files already formatted
cd server && .venv/bin/ruff check proliferate/ tests/            # All checks passed

python3 scripts/check_max_lines.py            # passed
python3 scripts/check_server_boundaries.py    # passed
python3 scripts/check_docs.py                 # passed (228 Markdown files)
node scripts/agent-catalog/check-version-discipline.mjs --base 2506df6ab   # OK
```

Also ran the adjacent seams that call the gate (`-k "cloud_sandbox or wake or connect or materializ"`): 146 passed.

`mypy` is broadly red on main (631 errors in 135 files) and is not wired into Server CI; the one error reported in `authorization.py` (`remaining_seconds` float/int) is pre-existing on `2506df6ab` and untouched here.

## Review findings landed

Agent review returned NEEDS_CHANGES; head `44deeffe9` lands findings 1/2/3/5 (finding 4 was this PR body; finding 6 is the WS-gateway seam, out of scope and untouched).

- **1 (blocker) — the verdict-persist arm now fails closed too.** The `record_billing_decision_event` + `commit()` that persist a genuine 402 sat *outside* the new `try`; a failure there (read-only replica, full disk, connection cap — conditions that hit reads and writes together) escaped as a bare `RuntimeError` with no typed body, no receipt, no alert. It now routes through `_deny_unreadable_billing_state`. It still **denies** — that helper raises on every path — and the `CloudSandboxResumeBlockedError` raise stays outside the guard so a legitimate quota 402 is never relabelled as unreadable.
- **2 — the receipt's own session is now bounded.** That second pool checkout happens during the very outage that broke the first one, so an unbounded wait held the 503 hostage for `pool_timeout` (30s), or forever for a hung DB. The receipt write (only the receipt — not the alert) is wrapped in `asyncio.wait_for` with a 5s module constant; on timeout it logs and the denial proceeds, exactly as the existing best-effort path already did.
- **3 — the materialization runner no longer double-pages.** `BillingStateUnavailableError` fell through to `except Exception → report_critical`, on top of the alert `_deny_unreadable_billing_state` already fired. It now has its own arm in both runner handlers: warning-level structured log, no second page.
- **5 — the 503 is pinned non-sticky.** New E5×E6 test: after a read-failure denial (with the caller's rollback), the immediately-following call with healthy reads is allowed. A transient DB blip must not strand healthy owners behind a cached `billing_unavailable`.

New/changed tests:

- `test_verdict_persist_failure_denies_with_unavailable_not_bare_error`
- `test_verdict_persist_failure_receipts_on_its_own_session`
- `test_quota_denial_passes_through_the_persist_guard_untouched`
- `test_receipt_session_hang_does_not_hold_the_denial`
- `test_unavailable_denial_is_not_sticky_next_healthy_read_allows`
- `test_runner_does_not_double_page_on_billing_unavailable` (in `test_billing_start_block_paging.py`)

Gates re-run at `44deeffe9`: `pytest tests/integration -k billing` 111 passed; `ruff format --check` + `ruff check` clean; `check_max_lines.py` / `check_server_boundaries.py` / `check_docs.py` all passed. `mypy` on the two changed modules reports the same 3 pre-existing errors as the base — no new ones.

## Scope

Server-only, billing authorization path only. **No spec change** — the `billing.md` re-cut with the E5/E6 gap checklist is PR #1571 off main; the gap checkbox gets struck when both land.

Base: main @ `2506df6ab`.


pablonyx:
## CodeQL check: 4 false positives — recommend dismissal (founder's call, not executed)

The red CodeQL check is alerts **94/95/96/97** (`py/clear-text-logging-sensitive-data` at `authorization.py:263/:244/:259` and `runner.py:58`). Verified from the CodeQL query source (`SensitiveDataHeuristics.qll`): the `maybePrivate()` heuristic regex-matches the literal token `billing` as financial data, so any identifier containing `billing_subject_id` becomes a taint source. The logged values are opaque correlation UUIDs required by the N6 fail-closed law (on-call must be able to tie a billing-read outage to the denied subject).

**No code change cures it:**
- Suppression comments (`codeql[...]`/`lgtm[...]`) are not honored by `codeql-action` (they need the separate `py/alert-suppression` query, which this repo's workflow doesn't run).
- The query has no Python sanitizer subclass — there is no hashing/encoding barrier to route through.
- Renaming `billing_subject_id` (the Postgres column name + documented error-attribute contract) to dodge a regex would be worse than the alert.

**Precedent:** identical class as three founder-dismissed alerts — #55 is `_log_billing_block` in the *same runner.py* logging the identical `billing_subject_id`/`user_id` pair, plus #67/#68 (disclosed in PR #1208). Five more instances of this rule sit open on main.

Suggested dismissal (for each of 94, 95, 96, 97):

```sh
gh api --method PATCH /repos/proliferate-ai/proliferate/code-scanning/alerts/<n> \
  -f state=dismissed -f dismissed_reason='false positive' \
  -f dismissed_comment='Internal correlation IDs (billing_subject_id / owner_user_id UUIDs) in structured fail-closed logs — required by law N6 so on-call can correlate a billing-read outage to the denied subject. CodeQL maybePrivate() matches the literal token "billing" as financial data; no card, account, or PII value flows here. Same class as the founder-reviewed #55 dismissal for the sibling _log_billing_block in this same file, and #67/#68 (PR #1208).'
```

pablonyx:
## Merged `origin/main` (sandbox stack #1500-#1534) — semantic port, please re-glance at one decision

Head moved `44deeffe9` → `b4bc79af9`. `MERGEABLE` again. The reviewed behaviour is intact, but the resolution was **not** mechanical, so flagging the one judgement call.

### What conflicted

Only `server/proliferate/server/billing/authorization.py`, three hunks.

### The judgement call: I dropped the guard on the dead start authorizers

`44deeffe9` wrapped **four** billing-read sites in the E6 fail-closed corridor:

1. `authorize_sandbox_start_for_billing_subject`
2. `authorize_sandbox_start`
3. `assert_cloud_sandbox_resume_allowed_for_owner` — the read arm (subject resolution → snapshot → membership → budget-limit sums)
4. `assert_cloud_sandbox_resume_allowed_for_owner` — the verdict-persist arm (`record_billing_decision_event` + `db.commit()`)

Main's #1501 dead-code sweep **deleted (1) and (2)** outright, along with `record_sandbox_start_authorization`, `BILLING_DECISION_AUTHORIZE_START`, and the `SandboxStartAuthorization` dataclass they returned. Those had been orphaned since #823 — nothing in `server/` called them.

**I took main's deletion, which drops the guard on (1) and (2) with them.** Rationale: keeping the guarded copies would leave three unreachable functions referencing a dataclass main removed from `billing/models.py` — an import error, not a safety net. A fail-closed guard on a code path that no longer exists protects nothing. Verified with a repo-wide grep that nothing references the deleted symbols.

**Guards (3) and (4) — the ones that actually matter — are byte-for-byte preserved,** as is every supporting piece: `BillingStateUnavailableError` (503 `billing_unavailable`, still deliberately not a 402), `BILLING_DECISION_READ_UNAVAILABLE`, `_UNRESOLVED_BILLING_SUBJECT_ID`, `_BILLING_RECEIPT_WRITE_TIMEOUT_SECONDS`, `_write_unreadable_billing_state_receipt`, `_deny_unreadable_billing_state`, and the `runner.py` no-double-page handlers.

Net effect on the enforcement surface: **strictly better.** `assert_cloud_sandbox_resume_allowed_for_owner` is now the *only* billing authorizer in the module, and it is fully guarded. Post-merge, 100% of billing authorization reads are inside the corridor, versus a module that previously also carried two guarded-but-unreachable ones.

Two smaller mechanical notes:

- **Re-added `from proliferate.db import session_ops as db_session`.** Main dropped it as unused once the dead authorizers went, but the surviving `_write_unreadable_billing_state_receipt` still needs it for the receipt's own session — and the E6 tests monkeypatch `authorization_module.db_session` to simulate the pool hang.
- **Took main's comment rewording** (`"the ensure service layer"` in place of `"wake/ensure"` after #1516 collapsed wake into ensure; the `CloudSandboxResumeBlockedError` docstring no longer citing the orphaned `authorize_sandbox_start`), reapplied inside this branch's restructured `try` blocks.

### #1516's new billing-read sites — checked, no scope expansion needed

#1516's per-access gateway gate, `gateway/service.py::_assert_gateway_billing_allowed`, is a **new call site of `assert_cloud_sandbox_resume_allowed_for_owner`** rather than a new billing read. It therefore inherits the fail-closed corridor for free — a read failure behind the gateway proxy raises the typed 503 with its receipt and single alert, exactly as at the other seams.

Its 5s `_gateway_billing_allow_cache` only memoises *successful* decisions (the `_gateway_billing_allow_cache[user_id] = ...` assignment sits after the authorizer returns), so neither a 402 nor a 503 is ever cached — which is also what corridor E5 requires. Same for `cloud_sandboxes/service.py::ensure_cloud_sandbox_ready` after #1516's `commit_cloud_sandbox_session` addition: still one gated call, still ahead of the row INSERT.

I found **no billing-read site introduced by main that sits outside the guard.** Did not expand scope.

Client contract also still holds: `workspace-connection-retry.ts` classifies billing blocks on `status === 402` plus a code in `{billing_credits_exhausted, billing_start_blocked}`, so the 503 `billing_unavailable` correctly falls through to generic retry rather than main's new `cloud-workspace-billing-block-store` top-up surface.

### Verification (scratch DB, since dropped)

| Suite | Result |
|---|---|
| `test_billing_fail_closed_read_errors.py` + `test_billing_start_block_paging.py` | **15 passed** |
| `pytest tests/integration -k billing` | **110 passed** |
| `pytest tests/unit -k "billing or gateway"` | **157 passed** |
| Sandbox/gateway files main's merge touched (`ensure_billing_gate`, `cold_access_repair`, `reconnect_self_heal`, `gateway_service`, `connect_race`, `materialization_failures`) | **48 passed** |

`ruff check` + `ruff format --check` clean on all five touched files. `scripts/check_max_lines.py` and `scripts/check_server_boundaries.py` pass (`authorization.py` is 449 lines, comfortably under the 500 server-layer max — the deletions bought headroom). `import proliferate.main` succeeds. `mypy` reports the same 3 pre-existing findings in these two files that main carries on its own (`remaining_seconds` `float|None` vs `int|None`, two `explicit-any` in `runner.py`); no new ones.

No PR merge, no draft/label changes.

PR #1581, opened 2026-07-28
fix(ui): repair the warning token, collapsed code blocks, Geist mono, and four dead component APIs

pablonyx:
Rendering all 133 design-system components in isolation surfaced a set of real
defects. This fixes the ones that were mechanical enough to fix safely, and adds
a gate so the largest class of them cannot come back.

## 1. Warning states rendered in an invisible token (23 sites)

`--color-warning` is a **fill**: `rgba(255, 180, 50, 0.15)` in dark mode and
`#fff8e6` in light. The ink is `--color-warning-foreground`. 21 sites used the
fill as text colour, which is near-invisible in both modes — and several
compounded it with a Tailwind alpha modifier (`bg-warning/10`), multiplying an
already-alpha token down to ~1.5% opacity.

The worst instances were 8px status dots painted solid `bg-warning`, sitting
directly beside opaque `bg-destructive`/`bg-success` siblings: `HarnessStatusDot`
(a harness that needs login) and `composerIntegrationHealthDot` (an integration
that needs re-auth). Both are "needs your attention" signals that were invisible.

Swapped to the purpose-built trio — `bg-warning-subtle` (panel fill),
`border-warning-border`, `text-warning-foreground` / `bg-warning-foreground`
(ink) — which is what `Badge`, `ProductNotice` and the workflow surfaces already
converged on. **No token values changed.**

Then added a `WARNING_TOKEN_AS_INK` rule to `scripts/check_frontend_boundaries.py`
(already in CI) so `text-warning` and alpha-modified `bg`/`border-warning/NN`
fail the build. Solid `bg-warning` stays legal — it is a legitimate panel fill.
5 new unit tests cover both directions.

The gate justified itself immediately: `BillingGateState` landed on main while
this branch was in flight, reached for `border-warning/30 bg-warning/10`, and
got caught on rebase. That is the last commit here.

## 2. Every highlighted code block in the transcript collapsed to one line

shiki's tokens carry no trailing newline and `CodeTokenLine` emits a bare inline
`<span>`, so the un-guttered branch of `CodeBlockTokenContent` had nothing to
break lines on. `showLineNumbers` defaults to `false` and both the transcript and
desktop renderers rely on that default, so this was the branch users hit.

Fixed in `CodeBlockTokenContent`, not `CodeTokenLine`: `FileSourceView` consumes
`CodeTokenLine` directly with its own block wrapper, so fixing the leaf would
have double-wrapped it. 5 new tests, including one that fails without the fix and
one pinning that a blank line does not become double-spacing.

## 3. Geist Mono was silently missing from production

`product.css` declared `@font-face` sources as absolute
`/node_modules/geist/dist/fonts/...`. That resolves only when something serves
the repo's node_modules at the site root — true for a dev server, false for every
built target. `pnpm web:build` emitted the rule with no such asset, so the
browser dropped the declaration and fell back to a system font.

`copy-css.mjs` now copies the two woff2s into `dist/fonts/` and the CSS
references them relatively. Verified against a real build: `dist/assets` carries
hashed `Geist-Variable-*.woff2` and `GeistMono-Variable-*.woff2`, and no
`/node_modules/` font URL survives.

## 4. Four components whose public API stopped short of usable

- **`Tooltip`** hard-coded its own provider at Radix's 700ms default with no
  `open` escape hatch — impossible to drive from state or open programmatically.
  Now forwards `open`/`defaultOpen`/`onOpenChange`/`delayDuration` (default 0).
- **`DropdownMenuSubContent`** rendered inline, so a submenu inside any
  `overflow-hidden` ancestor got clipped. Now portalled, like
  `DropdownMenuContent` already was. Zero existing callers, so zero regression
  surface.
- **`EnvironmentSearchSelect`** composed `PopoverButton` but swallowed its
  `externalOpen`/`onOpenChange`. Now forwards both.
- **`SettingsMenu`**'s `leading` slot rendered raw icons at intrinsic size. Added
  `[&_svg]:icon-paired`, matching `SegmentedControl` and `RadioCardGroup`.

## Three questions from the audit, answered

**Why `body { background: transparent }`?** It is load-bearing, so I did not
touch it. `apps/desktop/src-tauri/tauri.conf.json:25` sets `"transparent": true`;
painting the body would break the desktop window. The audit's proposed fix here
was wrong and is deliberately not applied.

**Subpath-only packages — deliberate or incidental?** Deliberate.
`specs/codebase/platforms/product/design-system.md:917-926` mandates "No
aliases: one subpath per component… never a barrel." No `exports["."]` added;
that finding is working as designed.

**Icon policy (lucide vs DS duplicates)?** Left alone — this one needs a decision
and is not mechanical. Actual scope is worse than the audit reported: 76 distinct
lucide icons across 50 files, 40 of them shadowing a DS name. lucide sets
intrinsic `width/height=24`; DS icons set none and size via `icon-*` utilities.
**16 lucide usages carry no `className` at all** and silently depend on that 24px
default, so a mechanical swap renders them viewport-sized. Needs a policy call
plus per-site sizing work.

## Verification

- `product-client` 3961 passed (642 files)
- `product-ui` 218 passed (33 files)
- `ui` 30 passed (12 files)
- boundary gate 13/13; `check_frontend_boundaries.py` clean
- `pnpm shared:typecheck` clean; `check-theme: ok (276 live tokens)`
- `pnpm web:build` used to prove the font fix end to end

One caveat for the reviewer: the two status-dot fixes changed assertions in
`HarnessStatusDot.test.tsx` and `composer-integrations.test.ts`. Those tests
pinned the buggy colour, so updating them was necessary — worth a look to confirm
you agree the new colour is the intended one.

### DIFF
--- a/apps/packages/design/scripts/copy-css.mjs
+++ b/apps/packages/design/scripts/copy-css.mjs
@@ -1,5 +1,6 @@
 import { copyFile, mkdir, readdir } from "node:fs/promises";
-import { resolve } from "node:path";
+import { createRequire } from "node:module";
+import { dirname, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
 
 const root = fileURLToPath(new URL("..", import.meta.url));
@@ -14,4 +15,36 @@ for (const entry of await readdir(sourceDir)) {
   }
 
   await copyFile(resolve(sourceDir, entry), resolve(targetDir, entry));
+}
+
+/* ------------------------------------------------------------------ *
+ * Geist ships as a bare package with no stylesheet of its own, so the
+ * `@font-face` rules live in product.css. They used to point at absolute
+ * `/node_modules/geist/...` URLs, which only resolve because a dev server
+ * happens to serve node_modules — any other bundler or deploy target got a
+ * dangling rule and silently fell back to a system mono font.
+ *
+ * Copying the woff2s next to the emitted CSS lets product.css reference them
+ * relatively (`../fonts/…`), so the declaration resolves wherever the package
+ * is consumed from. `@fontsource-variable/*` needs no equivalent step: those
+ * packages ship their own CSS and their own fonts.
+ * ------------------------------------------------------------------ */
+
+// `geist` does not export `./package.json`, so resolve through `./font` (which
+// maps to `dist/font.js`) and walk to the sibling `fonts/` directory. This
+// works regardless of how pnpm hoists the package.
+const require = createRequire(import.meta.url);
+const geistFontsDir = resolve(dirname(require.resolve("geist/font")), "fonts");
+const fontsTargetDir = resolve(root, "dist/fonts");
+
+await mkdir(fontsTargetDir, { recursive: true });
+
+for (const relativePath of [
+  "geist-sans/Geist-Variable.woff2",
+  "geist-mono/GeistMono-Variable.woff2",
+]) {
+  const source = resolve(geistFontsDir, relativePath);
+  const target = resolve(fontsTargetDir, relativePath.split("/").at(-1));
+  await copyFile(source, target);
+  console.log(`copy-css: geist ${relativePath} -> dist/fonts/`);
 }
--- a/apps/packages/design/src/css/product.css
+++ b/apps/packages/design/src/css/product.css
@@ -13,10 +13,15 @@
 @source "../../../product-surfaces/src";
 @source "../../../product-client/src";
 
-/* -- Geist font loading (kept for fallback / mono) -- */
+/* -- Geist font loading (kept for fallback / mono) --
+   Relative to the EMITTED stylesheet (dist/css/product.css); `copy-css.mjs`
+   places the woff2s in dist/fonts/. These were absolute `/node_modules/…`
+   URLs, which only resolve when a dev server serves node_modules — every
+   other bundler or deploy target got a dangling rule and fell back to a
+   system mono font. Keep these paths in step with copy-css.mjs. */
 @font-face {
   font-family: "Geist";
-  src: url("/node_modules/geist/dist/fonts/geist-sans/Geist-Variable.woff2") format("woff2");
+  src: url("../fonts/Geist-Variable.woff2") format("woff2");
   font-weight: 100 900;
   font-style: normal;
   font-display: swap;
@@ -24,7 +29,7 @@
 
 @font-face {
   font-family: "Geist Mono";
-  src: url("/node_modules/geist/dist/fonts/geist-mono/GeistMono-Variable.woff2") format("woff2");
+  src: url("../fonts/GeistMono-Variable.woff2") format("woff2");
   font-weight: 100 900;
   font-style: normal;
   font-display: swap;
--- a/apps/packages/product-client/src/components/agents/AgentLoginTerminalPanel.tsx
+++ b/apps/packages/product-client/src/components/agents/AgentLoginTerminalPanel.tsx
@@ -100,7 +100,7 @@ export function AgentLoginTerminalPanel({
       ) : null}
 
       {connectionError ? (
-        <div className="border-b border-border/70 px-3 py-2 text-ui text-warning">
+        <div className="border-b border-border/70 px-3 py-2 text-ui text-warning-foreground">
           {connectionError}
         </div>
       ) : null}
--- a/apps/packages/product-client/src/components/app/OfflineIndicator.tsx
+++ b/apps/packages/product-client/src/components/app/OfflineIndicator.tsx
@@ -15,7 +15,7 @@ export function OfflineIndicator() {
   return (
     <div
       role="status"
-      className="flex items-center justify-center gap-2 border-b border-warning/40 bg-warning px-3 py-1.5 text-ui font-medium text-warning-foreground"
+      className="flex items-center justify-center gap-2 border-b border-warning-border bg-warning px-3 py-1.5 text-ui font-medium text-warning-foreground"
     >
       <WifiOff className="icon-paired shrink-0" aria-hidden="true" />
       No internet connection — local work is safe; agents need a connection.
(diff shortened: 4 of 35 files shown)


pablonyx:
## Independent review — #1581 @ 247007abf3

Verdict: this holds up. All four claims check out against the actual head SHA (not just the diff text) — the warning-token migration is complete with zero straggler sites and a CI gate that fails closed on the right patterns, the code-block fix is narrowly scoped and doesn't touch the copy payload, the Geist font path resolves correctly against the real installed `geist@1.7.2` package layout, and all four "dead API" fixes are genuinely wired pass-throughs with no double-source-of-truth. The one real gap is test coverage: the four component-API fixes (Tooltip, DropdownMenuSubContent, EnvironmentSearchSelect, SettingsMenu) ship with zero new tests, resting entirely on manual verification and the PR body's narrative. Nothing here is blocking.

### Should fix

- The four API-repair fixes (`apps/packages/ui/src/primitives/Tooltip.tsx`, `apps/packages/ui/src/primitives/DropdownMenu.tsx:614-640`, `apps/packages/ui/src/patterns/EnvironmentSearchSelect.tsx:552-576`, `apps/packages/ui/src/patterns/SettingsMenu.tsx:588-593`) have no new tests. `+138/-2` test lines all land on the warning-token gate and the code-block fix — the API-gap section of this PR is untested. A one-line RTL test asserting `Tooltip` renders open when `open={true}` (or `defaultOpen`) would have caught a real class of regression (e.g. a future refactor that stops forwarding `open` to `KitTooltip`) and costs almost nothing given the pattern is already established in `CodeBlockTokenContent.test.tsx`.
- Bundling four unrelated fixes (token repair, rendering bug, packaging bug, four API additions) in one PR makes review harder than it needs to be — each of the four has a distinct blast radius and a distinct "how would I verify this broke" story. Given AGENTS.md's "record unrelated defects as follow-ups instead of expanding the current PR," this arguably should have been 3-4 PRs. That said, the PR body is unusually disciplined about drawing the boundary (explicitly declines to fix the icon-policy and `body { background: transparent }` items it also found), and every included fix is genuinely mechanical, so I'm not calling this blocking — just noting the size made this review take longer than four fixes of this simplicity should.

### Consider

- `scripts/check_frontend_boundaries.py:703-755` (`find_warning_ink_violations`) only scans `UI_SRC, PRODUCT_UI_SRC, PRODUCT_SURFACES_SRC, PRODUCT_CLIENT_SRC, DESKTOP_SRC, WEB_SRC` — `MOBILE_SRC` is excluded (consistent with the file's other rules, since mobile doesn't consume Tailwind classNames the same way), but worth confirming there's genuinely no Tailwind-classname surface on mobile that could carry `text-warning`, since I didn't exhaustively audit RN styling there.
- `WARNING_INK_RE`'s `bg`/`border` branch only flags `-warning/\d+` (a literal numeric alpha suffix). A dynamic class built via template literal (`` `bg-warning/${n}` ``) would evade the regex, same as most of this style of static-analysis gate — not a regression, just a known class of gap worth knowing about if templated classNames show up near warning tokens later.

### On Greptile's findings

Greptile gave a 5/5 confidence score with "no actionable defects identified" and a one-line-per-file summary, no specific line-level findings to agree or disagree with. Its overview claims are accurate as far as they go (font asset copying, code-block line wrappers, controlled-state props, the new CI rule) but it's a summary, not a review — it didn't verify the token contrast math, didn't check for remaining `text-warning` call sites at the actual head SHA, didn't check test coverage gaps, and didn't flag the untested API surface. I agree with everything it stated as fact; I don't think it did enough independent verification to be treated as a substitute for the checks above.

### Verified

- Zero remaining bare `text-warning` or alpha-modified `bg-warning/N` / `border-warning/N` usages anywhere in `apps/packages/**`, `apps/desktop/**`, `apps/web/**` at commit `247007abf3` (checked via `git grep` against the actual head SHA, not the diff).
- `--color-warning-foreground` is opaque (`#ffb432` dark / `#8a5a00` light, no alpha) in both modes; `--color-warning-border` and `--color-warning-subtle` are the only alpha values in the family, and only used as fill/border, never composited as ink — matches the PR's stated design.
- Contrast: dark `warning-foreground` on `surface` ≈ 10:1, light ≈ 5.9:1 (secondary-text floor is 4.5:1, both clear it as "attention ink," not body text). `check_theme_contrast.py --table` passes clean with 4 pre-existing pinned deviations unrelated to warning.
- `tokens.ts` warning family values are byte-identical before/after this PR (`git diff main...247007abf3 -- apps/packages/design/src/tokens.ts` is empty) — the PR's "No token values changed" claim holds, and `dist/theme.css` byte-equality is not at risk since `product.css` (font URLs) is a separate emitted file from `theme.css`.
- `WARNING_TOKEN_AS_INK` gate is wired into CI (`.github/workflows/ci.yml` calls `check_frontend_boundaries.py`), and its regex correctly handles variant prefixes (`hover:text-warning`, `dark:bg-warning/10`) via the `\b` boundary, correctly allows solid `bg-warning` and the purpose-built tokens, and is covered by 5 focused unit tests including a comment-stripping negative control.
- `CodeBlockTokenContent`'s `block` wrapper only touches the un-guttered branch; the gutter branch (table rows) is untouched and has its own test asserting no `block` class leaks into `<td>` cells. Copy-to-clipboard (`MarkdownCodeBlock.tsx`, `HighlightedCodeBlock.tsx`) reads from a separate raw `code` string / `useProductHost().clipboard`, not the rendered DOM, so this fix cannot regress copy fidelity for blank lines or trailing content.
- Geist font resolution: `require.resolve("geist/font")` resolves to `dist/font.js` in the actual installed `geist@1.7.2` package, and `dist/fonts/{geist-sans,geist-mono}/*.woff2` exist at exactly the paths the script copies from; `package.json`'s `files: ["dist"]` covers the new `dist/fonts/` output since it's the whole `dist` tree.
- `DropdownMenuSubContent` has zero call sites anywhere in the repo (only the definition and generated `.d.ts`/`.js` matched) — the "zero regression surface" claim is accurate.
- `EnvironmentSearchSelect`'s new `externalOpen`/`onOpenChange` props are genuinely forwarded to `PopoverButton`, which already supports controlled/uncontrolled open state via an `externalOpen === undefined` guard — no double-source-of-truth introduced.
- All CI checks pass on the PR including "Login first-load budget (phase-6)," which fails closed on JS/CSS gzip caps and zero-font/image/audio-on-login — confirms the Geist packaging change didn't regress the login bundle.

PR #1587, opened 2026-07-29
fix(design): author light mode independently of dark

pablonyx:
Light mode was never authored. It was derived: `:root[data-mode="light"]` took the
dark palette, flipped the ink, and kept dark's alpha percentages. Alpha values
tuned to lift a surface *off black* do the opposite work over white — they wash
out instead of separating — so the whole mode read as a low-contrast smear of the
dark one rather than as a light interface.

This re-authors it from first principles. Four rules, applied throughout:

1. **Light is authored independently, never derived from dark.** No token's light
   value is a function of its dark value.
2. **No alpha neutrals in the token layer.** Every neutral role resolves to an
   opaque hex. Alpha still exists at the call site, where a consumer can see what
   it composites over; it does not exist in a role definition, where it cannot.
3. **The chat pane stays white.** Definition comes from borders and shadows,
   which are load-bearing here and are not softened to taste. The sidebar is its
   own plane, not a tint of the pane.
4. **Selection is neutral ink, not blue.** Blue is reserved for the focus ring,
   links, and genuinely special state, so a selected row no longer reads as a
   focused one.

Also in scope, because they were actively hiding the problem:

- Removed the non-`@supports` srgb fallbacks. They were broken — never matched by
  any browser we ship to, so they served only to make the token layer look like
  it had considered a case it had not.
- Swept the leaked dark literals, i.e. hard-coded values that happened to work in
  dark and had no light counterpart at all.

## The gate

`scripts/check_theme_contrast.py` measures the **built stylesheet**, not the
source, so it sees what ships after projection. Floors: body ≥7:1, secondary
≥4.5:1, faint ≥4.5:1, borders ≥1.25:1, and interaction states mutually
distinguishable — enforced for **both** modes. Wired into CI.

Four pre-existing deviations are pinned in `DECLARED_DEVIATIONS` rather than
silently excluded, each with its measured ratio, so they read as debt with a
number attached instead of as passing.

## Verification

Guards green: contrast, appearance-scaling, design-attribution, max-lines,
frontend-boundaries, frontend-structure. Screenshots of every major surface in
both modes are attached to the review thread.

## Known follow-ups (deliberately not in this PR)

- Dark mode's selection inversion needs its own pass; this PR does not touch it.
- Light `--color-faint` on `--color-surface-under` measures 4.4995:1 — over the
  floor by 0.0005 and unpinned. Worth a deliberate decision rather than an
  accident.


pablonyx:
## Independent review — #1587 @ 94a8fe3a9f

The diagnosis is correct and the light neutral scale is real work: 64 light values genuinely changed, the ink roles now clear their floors on white, `selected` no longer reads fainter than `hover` in light, and the guard is honest enough to print its own debt instead of hiding it. But the PR's own two headline rules are not actually delivered. Rule 2 ("no alpha neutrals in the token layer") ships with **no guard at all** — nothing in `check-theme.mjs` or the new 662-line checker forbids a future light `color-mix(foreground X%, transparent)`, and one such value survives in this diff. And the light `--shadow-*` values, which rule 3 makes load-bearing, **never reach the browser** at the 40 call sites that consume them as Tailwind utilities — the exact mechanism the second commit's message describes, found for one token and not generalized. Separately, `--color-surface-control` was flattened white-on-white at 34 borderless call sites, which is the original bug re-introduced in a token the guard does not measure; and the canonical design-system spec still teaches the abolished derive-from-dark rule. Blocking.

### Blocking

- `apps/packages/design/src/tokens.ts:1142,1147,1152` — the newly authored light `--shadow-modal` / `--shadow-popover` / `--shadow-subtle` are dead values. Tailwind inlines a `shadow-*` utility's value at build time from the `@theme` block, which the generator fills from `themeFallback ?? dark`. Compiling the committed `dist/theme.css` through the repo's own Tailwind 4.3.3 emits `.shadow-subtle { --tw-shadow: 0 1px 2px 0 var(--tw-shadow-color, rgb(0 0 0 / 0.05)) }` — dark's value — and **no** `[data-mode="light"]` rule anywhere re-declares `--tw-shadow`. There are 40 such call sites (`shadow-popover` 24, `shadow-modal` 11, `shadow-subtle` 5) across `product-client`, `product-ui`, and `ui`. So in light mode every popover, modal, tooltip and card lift still paints dark's pure-black shadow, and the only two places the new light shadows take effect are the two CSS sites that read `var(--shadow-*)` directly (`--color-composer-shadow`, `.chat-user-message-bubble`). The second commit message states this mechanism precisely — "a generated shadow utility inlines the token's value at build time" — and then fixes it for exactly one token. Given rule 3 makes shadows half of light's entire definition, shipping the light shadow scale as a no-op voids the contract. Failure scenario: reviewer approves on screenshots of the composer and user bubble (the two working sites) while every menu and dialog in light mode keeps a dark-tuned `rgb(0 0 0 / 0.5)` drop shadow.
- `apps/packages/design/src/tokens.ts:614` — `--color-surface-control` light went from `rgba(237, 237, 237, 0.4)` (a visible tinted chip on white) to `#ffffff`. 34 `bg-surface-control` call sites carry no border and no ring, so they are now `#ffffff` on a `#ffffff` pane — no fill at all. Concretely: `QueuedPromptEditBanner.tsx:11` (the "Editing queued message" banner), `ApprovalCard.tsx:81` (the permission-request command block), `TodoTrackerPanel.tsx:39` (the progress track, which is a 2px bar whose *entire* affordance is the fill), `ComposerOptionRow.tsx:15` (the key-cap glyph), `PaneFileTree.tsx:59,187`. This is the same defect the PR exists to fix, moved into a token the checker never grades: `--color-surface-control` appears in no `TEXT_PLANES`, `BORDER_PAIRS`, or `STATE_ROLES` tuple, and the guard measures no surface-against-surface pair at all.
- `apps/packages/design/src/tokens.ts:761` — `--color-warning-subtle` light is still `rgba(242,201,76,0.14)`, byte-identical to its dark half. This is the one surviving light alpha neutral that is neither a shadow nor a scrollbar, i.e. a direct violation of both rule 1 (not derived from dark) and rule 2 (no alpha neutrals), in a diff that retuned its three siblings `--color-success-subtle`, `--color-destructive-subtle`, and `--color-warning` to opaque hex. It is invisible to the guard for the same reason as above.
- `specs/codebase/platforms/product/design-system.md:205-222,244-252,276-284,290-297` — zero spec files are in this diff, and that document is a current codebase document, i.e. per `specs/README.md` "Operating law for its area." It still contains a section titled "The house form: symmetric opacity" that teaches the abolished rule verbatim ("The percentage is identical in both halves… that symmetry is why the hierarchy survives a mode switch without a second set of tuned values"). Its Surfaces table still lists light `--color-surface-under` as `#f9f9f9`, `--color-surface-control` as `rgba(237, 237, 237, 0.4)`, `--color-sidebar` as `var(--color-surface-under)`, `--color-card` as `var(--color-surface-elevated)`; its Borders table says "Both modes | 5% / 8.4% / 12%"; its Text table says light is 70% / 50%. Failure scenario: the next agent reads the operating law before touching a token, as `CLAUDE.md` and `AGENTS.md` require, and re-derives light from dark exactly as instructed — with nothing in the guard to stop it, since no check enforces rule 2.
- `scripts/check_theme_contrast.py:78` — `TEXT_PLANES` is only `("--color-surface", "--color-background")`, both `#ffffff` in light. Every ink role is therefore graded against white and only white, while the light retune created three *non-white* planes that carry the same ink: `--color-sidebar` / `--color-surface-under` `#edf0f2`, `--color-muted` / `--color-diff-surface` `#f2f4f6`, `--color-surface-editor` `#f7f8fa`. `--color-faint` `#686e76` on `#edf0f2` measures **4.4962:1** — below the 4.5 floor, unmeasured, and unpinned. That ink lands there for real: `text-faint` at `apps/packages/product-ui/src/sidebar/ProductSidebarRepositories.tsx:226`, rendered inside the `bg-sidebar` plane via `WorkspaceShellSidebar.tsx:135`. The PR body reports this pair as "4.4995:1 — over the floor by 0.0005"; it is under the floor, and by 0.0038. A guard whose stated contract is "floors, in both modes, no per-mode exemptions" that cannot see three of light's four planes does not enforce the contract it advertises, and the author's own conclusion that this value was safe is a direct consequence of the gap.

### Should fix

- `scripts/check_theme_contrast.py:124` — the dark `selected`-vs-`hover` pin is `0.0`, not the measured `1.08`. Because `Measurement.ok` accepts `ratio >= pin - 0.005`, this entry admits *any* value, so dark's `--color-selected` can regress all the way to the plane colour and still pass. Every other pin is a genuine ratchet (regression fails, `stale_pin` errors once the floor is met, orphaned pins error) — this one is a plain exemption wearing a ratchet's clothes, and it is the only pin whose docstring claim "can only ever improve" is false. Pin the measured `1.08` instead.
- `scripts/check_theme_contrast.py:60,65` plus the two `border-light` pins at `:107,:114` — `BORDER_FLOOR = 1.25` is a floor that **no** measured instance of `--color-border-light` meets in **either** mode (dark 1.14, light 1.24), so both are pinned; and `STATE_STEP_FLOOR = 1.02` is set at 1.02 while light `selected`-vs-`active` measures 1.0191 and is likewise pinned. A floor chosen such that the role it governs fails universally, then pinned around in both modes, is not a contract — it is a number picked to sit just above the values. Either `--color-border-light` is too weak (in which case fix it, as the PR did for every other neutral) or 1.25 is the wrong floor (in which case say 1.12 and mean it). Shipping both pinned proves nothing about either.
- `apps/packages/design/src/css/product.css:412` — `.thinking-text` is still `color: color-mix(in srgb, var(--color-muted-foreground) 60%, transparent)`, which composites to `#9c9fa3` on white = **2.66:1**. This is precisely the component-scope instance of the defect the PR claims to sweep; the diff fixed `.right-panel-tab-system` and `.file-source-view` and left this one, along with `product.css:415`'s sibling and the `--color-muted-foreground 16%` shimmer at `:306`. It is invisible to the checker because the checker grades tokens, not component CSS.
- `apps/packages/design/src/tokens.ts:511` (`--color-ring`) — light became blue `#0b6bcb`; dark stays neutral `color-mix(in oklab, #ffffff 28%, transparent)`. Rule 4 says "blue is reserved for the focus ring", but the focus ring is only blue in one mode, so the same keyboard-focus affordance changes hue across a mode switch and the stated reservation holds in neither direction. Pick one and apply it to both halves, or drop the claim.
- No guard enforces rule 2. `apps/packages/design/scripts/check-theme.mjs` asserts `themeFallback` discipline on `color-mix()` *dark* values and nothing about light; `check_theme_contrast.py` composites alpha over a supplied backdrop, so it would happily report a comfortable number for a re-introduced `color-mix(foreground 8.4%, transparent)` card-on-card. The PR's central rule is the one invariant left to prose. A ten-line assertion in `check-theme.mjs` — every `--color-*` light half is an opaque literal, `var()` alias, or explicitly listed exception (scrollbar, shadows) — would cost less than 2% of the new guard and protect the thing the PR is actually about.

### Consider

- `scripts/check_theme_contrast.py` at 662 lines is not governed by `check_max_lines.py` (`scripts/` is absent from its `CHECK_ROOTS`), so the PR's "check_max_lines green" claim is vacuous for the file it adds. More substantively, the `Resolver`/`_mix` engine — roughly 120 lines reimplementing premultiplied `color-mix()` interpolation, in sRGB while the tokens were authored in oklab, a discrepancy the file notes and waives — exists almost entirely to grade *dark*, since light now has no mix neutrals. Three of the four pins are dark's. The complexity is load-bearing only for the half of the system this PR declined to fix.
- Adding a new `DECLARED_DEVIATIONS` row is a plain dict edit with no owner or hash, so a new violation can be pinned in the same commit that causes it and will read as green. That is consistent with the repo's other guards (`max_lines_allowlist.txt`, `appearance_scaling_baseline.json`), so not a new sin — but this guard's docstring promises more ("Adding a row is a design decision") than the mechanism delivers.
- `.github/workflows/ci.yml:402` pins `actions/setup-python@v7` by tag while the same file pins `actions/checkout` and `actions/setup-node` to 40-char SHAs. Matches the two pre-existing `setup-python@v7` usages at `:23` and `:472`, so it is a repo-wide inconsistency rather than a regression here.
- `apps/packages/design/src/tokens.ts` — the retag flattened 70 distinct provenance values into one `[RETUNE:light/independent-scale]`, including 16 that named a specific prior decision (`[RETUNE:color/link-tone]`, `[RETUNE:sidebar/reference-surface]`, `[RETUNE:diffs/addition-deletion-color-alias]` ×6). Six of those tokens had their light value **unchanged** by this diff (`--color-background`, `--color-surface`, `--color-surface-elevated`, `--color-popover`, `--color-primary-foreground`, `--color-composer-background`) yet were retagged anyway, and `--color-link-foreground:457` now carries a prose comment about link-tone measurement above a tag that no longer mentions it. The spec says tags are historical; overwriting history on tokens that did not change subtracts from it.

### Verified

- **The three light surfaces at `#ffffff` are the intended decision, not the unfixed bug — for `card`/`surface`/`surface-elevated` specifically.** `--color-surface`, `--color-card`, and `--color-surface-elevated` all resolve to `#ffffff`, and the "definition from borders + shadows" claim holds up at the call sites: 41 of 42 `bg-card` usages carry a `border`/`ring` on the same element (the sole exception, a `bg-card/95` fade mask, needs no edge), and only 6 `bg-surface-elevated*` usages lack one, all `-secondary` (`#f2f4f6`, a real step). So collapsing that trio is defensible. The judgement does **not** extend to the same flattening applied to `--color-surface-control` (34 borderless sites, blocking above) — the decision is sound where the border discipline exists to back it and is an unfixed instance of the original bug where it does not. The distinguishing fact is that nothing in the codebase or the guard enforces that discipline; it currently holds by habit.
- Light is genuinely no longer derived from dark in the neutral scale: 64 light values changed, and the built light block contains **zero** `color-mix()` and no `var(--color-foreground)`-relative expressions. The only non-opaque light values are 3 shadows, 2 scrollbar rgba (justified — floats over arbitrary content), 7 structural `transparent`s, and `--color-warning-subtle` (blocking above).
- Ran `python3.12 scripts/check_theme_contrast.py --table` against the committed `dist/theme.css`: exit 0, 4 pins, and the script does read the built stylesheet, not `tokens.ts`. `python3.12 -m unittest scripts/test_check_theme_contrast.py`: 27/27 pass. The `read_block` column-zero anchoring correctly skips the indented `:root` inside `@media (prefers-reduced-motion)`, and the orphan-pin and stale-pin errors both work as described.
- Light `selected` (#dde1e6, 1.31:1 off white) is genuinely stronger than `hover` (#e4e7ea, 1.24:1) — the inversion is fixed in light. Dark's inversion (`selected` 1.08 vs `hover` 1.24) is left, which the PR states openly. Leaving dark alone is defensible on scope (the dark values are pre-existing and re-tuning them moves every list row in the mode that actually ships), but the `0.0` pin means dark is now *less* protected than before, not merely unimproved.
- `dist/theme.css` byte-equality is structurally safe: `check-theme.mjs` re-projects from the same `themeTokens` record through independent code, so any well-formed edit stays equal. Its one real constraint — every `color-mix()` dark half needs a resolved literal `themeFallback` — holds; I found 0 violations, and 0 now-redundant fallbacks on darks that stopped being mixes.
- The PR body's two "turned out not to exist" reports are true and were worth reporting rather than faking: `grep -rn '@supports'` finds no colour fallbacks anywhere in `apps/packages/design/src` or `product-ui/src`, and `--color-brand-logo-tile` appears in no source file.
- The leaked-dark-literal sweep is complete: no hardcoded `#181818`/`#212121`/`#2d2d2d`/`#282828`/`#141414`/`#0d0d0d`/`#222222`/`#111111`/`#262626` remain in `product-client`, `product-ui`, or the design CSS. The one grep hit is a prose comment at `product.css:183`.
- `python3.12 scripts/check_design_attribution.py` passes — no other product is named as the source of a decision.
- The CI step fails closed *within its job*: `.github/workflows/ci.yml:406` is a plain two-command `run:` with no `continue-on-error`, placed after `pnpm shared:build` so `dist/theme.css` exists, and the script returns 1 on any failure or unresolvable token. It does **not** gate merges, though: `main` has no branch protection (`/protection` → 404) and no rulesets (`/rulesets` → `[]`), so this job is advisory exactly like every other CI job in the repo. That is a repo-wide fact, not something this PR regresses.

PR #1588, opened 2026-07-29
feat(ui,updates): rebuild toasts as one component in three weights

pablonyx:
Stacked on #1587. Two halves: the kit-level toast, then the update flow that was
the worst offender against it.

## B1 · One toast component, three weights

There was no toast *component* — there were call sites, each re-deciding layout,
and a spread of `[&&]:!bg-*` and `!ml-2` specificity hacks to force the result
back into line. Those hacks are deleted, not relocated: the frame is now wired
once via `toastOptions` on the `Toaster` in `App.tsx`, so a call site chooses a
weight and nothing else.

The three weights:

- **`status`** (4s) — one line, one fact. Truncates rather than growing a second
  line, and carries the full string on `title` so the ellipsis loses nothing.
- **`announcement`** (8s) — headline, the consequence, and at most one solid
  button. This is the weight that is allowed to interrupt.
- **`detail`** — adds the mono excerpt.

Rules that are enforced rather than documented:

- **The excerpt test.** At most three countable lines, then `+N more`. Stack
  traces are never inline — if it does not fit the excerpt, it is not a toast.
- **Details resolves to exactly one of three destinations.** Not a fourth,
  not a modal invented at the call site.
- **Hard limits:** max 3 visible; same `id` replaces rather than stacks; never
  taller than the 3-line excerpt.
- Severity lives in a 6px dot (status) or the badge tone (announcement/detail) —
  never a tinted card. Three stacked toasts read as three messages, not three
  warnings.
- A toast never shows two equally loud buttons. Only `commit` carries a fill.

## B2 · The update flow, as a state machine

The pill owns continuous state; toasts fire only at a resolution. Concretely:

| phase | who speaks |
|---|---|
| `checking` | pill only (it gained a state for this) |
| `current` | 4s receipt, **manual checks only** — a background check that finds nothing has nothing to report |
| `available` | auto-downloads; toast offers "Skip this version" (reached only when auto-update is off, so asking for a click is honest) |
| `downloading` | pill only, hover morph |
| `stalled` | **new** — bytes frozen >8s, or `totalBytes === null` |
| `ready` | the one interruption in the flow |
| `error` | persists, with Retry |

`restartWhenIdle` gets a 10-second cancellable countdown, and the `?update=1`
window is retired.

## Two defects this surfaced

Both were invisible until the states became reachable, which is the argument for
making them reachable:

1. **Two of the new states could not be rendered at all.** The dev mock's phase
   union excluded `stalled`, and `use-updater` hard-nulled the stall and countdown
   figures whenever a mock drove the store — but the stall copy and the countdown
   are authored *from* those numbers. The mock now carries them, "Not now" cancels
   under the mock, and the playground exposes both stall shapes plus the countdown.

2. **The countdown stacked on top of the ready announcement.** Two toasts about
   one update, both offering Restart, and the ready toast's "Later" contradicting
   a relaunch ten seconds away. The countdown now supersedes it exactly as the
   confirm dialog already did — dismissed *and* suppressed, since suppression
   alone lets the next store update re-raise it underneath.

A third defect, the invisible warning badge on the stalled toast, is fixed at its
~23 call sites by #1581. This carries only the two sites that PR does not touch:
the toast tone dot, and `Badge`'s warning row taken verbatim from #1581 so the
line merges rather than conflicts.

## B3 · Every error names its subject

The reference toast said "This target does not support the selected model yet."
It named neither the model nor the target, so there was nothing to act on. The
fix is in two parts, and the first is that it stops being a toast.

- **The refusal belongs at the picker.** A refused row is disabled and labelled
  "Not supported on this target" rather than hidden, so the model that failed is
  still visible and explains itself; selecting it anyway pins the error to the
  control. A toast only fires where there is no picker to point at.
- **A refusal is a fact about one target**, so it is recorded per workspace +
  agent + model. Home never marks anything: it is pre-session, so no target has
  had the chance to refuse yet.
- **Nothing is predicted.** No minimum-AnyHarness-version or target-support
  metadata exists on models anywhere in the client, so the runtime's own refusal
  is the only honest source. The store keeps what a target has actually refused.
- **`formatSessionCreateToastMessage` is deleted.** It returned
  `${prefix}: ${message}` — a written headline concatenated with an exception,
  which is exactly the shape the toast API and `check_toast_copy.py` now forbid.
  Callers write their own headline and pass the exception as `cause`, where it
  stays out of the body and is reachable only through Details.
- Every remaining error toast names its subject and what it cost: the model and
  workspace, the repo and branch, or where the draft went. Retry appears only
  where re-running cannot duplicate what the user typed.

Four files crossed the frontend size threshold on the way. All four were fixed by
extraction rather than an allowlist bump — the cloud launch branch, the in-flight
create reuse, and the two plan workflow modules each read on their own now, and
one now-stale allowlist entry is removed.

## Verification

- product-client: tsc clean; 4117 tests / 649 files, all passing
- ui 19 files, product-domain 62, product-ui 31, product-surfaces 4 — all passing
- Guards green: appearance-scaling, design-attribution, max-lines,
  frontend-boundaries, theme-contrast, toast-copy, and frontend-structure --strict
- All eight toast states captured in both modes, each confirmed to render exactly
  one toast. Screenshots in the review thread.


pablonyx:
## Independent review — #1588 @ e1068f8c3

The toast kit itself is the strongest part of this PR and I'd land it: the three weights are a genuinely closed union, the frame is wired once on the kit `Toaster` instead of per call, and the old hacks are actually gone rather than relocated (`[&&]:` survives only inside two `Sonner.test.tsx` assertions that deliberately test caller override, and `"sonner"` is imported by exactly one file in `apps/**/src`). What needs changes is the update state machine bolted on beside it. Three of the states this PR exists to make reachable — check-failure retry, the stall clock, and the `totalBytes === null` stall trigger the PR body promises — don't behave as described, and none of them has a test that renders a real update toast, so CI green says nothing about them. Recommend: fix the three blocking items in this PR, and move the B2 update work (new `stalled` phase, module-level check scheduler, new `autoUpdateEnabled` user preference, auto-download hook, redesigned sidebar pill) to its own PR if you'd rather not carry that review surface here.

### Blocking

- `apps/packages/product-client/src/components/feedback/UpdateToastPresenter.tsx:174` — Retry on a failed *check* is a silent no-op. `void (checkFailed ? downloadUpdate() : retryDownload())` sends the check-failure branch to `downloadUpdate()`, but `runDownloadAndPrepareRestart` (`lib/domain/updates/updater-download.ts:19`) starts with `if (!update) return;` and a failed check never populated `store._update`. Scenario: background check throws (offline, signature server 502) → `setError(msg, "check")` → toast "Couldn't check for updates" with a Retry button → user presses Retry → nothing happens at all, no new check, no phase change, toast stays on screen forever (duration is `Infinity` because `isError`). The correct action is `checkNow`, which isn't even destructured in this file.
- `apps/packages/product-client/src/components/feedback/UpdateToastPresenter.tsx:237` — the stall copy freezes at its first value. `const seconds = stalledSeconds(lastProgressAt, Date.now())` is computed once inside an effect whose deps carry no clock, so a download that has been dead for five minutes still reads "No data for 8 seconds" until some other dep changes. Scenario: network drops mid-download → phase flips to `stalled` at t+8s, toast renders "No data for 8 seconds" → at t+300s the same toast still says 8 seconds, which actively misinforms the decision the toast is asking for (wait vs retry). The playground masks this by pre-seeding `lastProgressAt = Date.now() - 12_000`, so it looks plausible there.
- `apps/packages/product-client/src/lib/domain/updates/download-stall.ts:33` — the second stall trigger from the PR body is not implemented. `isDownloadStalled({ lastProgressAt, now })` reads only `lastProgressAt`; `totalBytes` is never consulted, even though the file's own header comment says "Both collapse into one named phase here." Scenario: the update server responds without `Content-Length` → `totalBytes` stays `null` → progress events keep arriving so `lastProgressAt` keeps refreshing → the phase never reaches `stalled` and the pill shows an indeterminate "Starting download…" indefinitely. That is precisely the shape the `stalled` phase was introduced to name. Either implement the clause or drop it from the PR body and the comment.

### Should fix

- `apps/packages/product-client/src/components/feedback/UpdateToastPresenter.tsx:168` — this breaks the PR's own headline rule that a cause is never rendered in the body and is always reachable via Details. `description: humanizeDownloadError(errorMessage)` puts a short raw exception string directly into the announcement description, and no `cause` is passed, so the full text is *only* in the body and Details is absent. Scenario: Tauri surfaces `error sending request for url (...): connection closed before message completed` → that lands in the toast description verbatim; a user who wants the full text has nowhere to go. Pass it as `cause` and write a consequence instead.
- `apps/packages/product-client/src/components/feedback/UpdateToastPresenter.tsx:131` — the restart countdown never counts down. `const seconds = Math.round(RESTART_COUNTDOWN_MS / 1000)` is a constant, so the toast says "restarts in 10 seconds" for the entire 10-second window and then the app restarts while the copy still claims 10 seconds remain. The test only asserts the substring, so it passes either way.
- `apps/packages/product-client/src/hooks/access/tauri/use-updater.ts:99` — `const downloadStartedAt = devMock ? null : storeDownloadStartedAt;` still hard-nulls the timestamp under the dev mock, which is the same class of defect the PR claims to have fixed for the other mock fields. Consequence: `formatRemainingTime` and the pill's hover time estimate are unreachable at `/playground/updates`, so the surface you'd use to review them can't show them. Same file:201 — `retryDownload` returns early `if (devMock || …)`, so the stalled toast's "Retry now" commit button is inert in the playground too.
- `apps/packages/ui/src/patterns/ToastBody.tsx:51` — "description clamped to 2 lines" is a 140-character heuristic, not a clamp; there is no `line-clamp` anywhere in the file. Combined with `apps/packages/ui/src/primitives/Sonner.tsx:50` (`!max-h-[168px] !overflow-hidden`), a description that wraps to 3+ lines at a narrow viewport is silently *clipped* mid-word by the frame rather than ellipsised. A description under 140 chars with a long title and a badge is the reproducing case. The clamp test would also pass with the constant set to 1.
- `apps/packages/product-client/src/components/feedback/UpdateToastPresenter.test.tsx` — the update-flow tests mock `@proliferate/ui/utils/show-toast` wholesale and assert on the input objects, so nothing renders a real update toast; that's why all four defects above are invisible to CI. Concretely, "drops the countdown toast once the countdown is not running" still passes if the countdown feature is deleted (no toast raised → no toast to assert against), and there is no negative test for pressing Retry after a check failure.

### Consider

- `apps/packages/ui/src/utils/toast-model.ts:174` — `TOAST_HEADLINE_MAX_CHARS = 60` is exported and used by nothing. `STATUS_MESSAGE_MAX_CHARS` is the constant that does the work. AGENTS.md says delete dead code when replacing an implementation.
- `apps/packages/ui/src/utils/toast-payload.ts:50` — `firstSentence` is computed and unit-tested but consumed by no rendering path, so the documented "when the payload is a blob, the first sentence goes in the description instead" behavior does not exist. Either wire it in `ToastBody` or remove it and the doc sentence.
- `apps/packages/ui/src/patterns/ToastBody.tsx:170` — `key={line}` in `ToastExcerpt` collides when a payload repeats an identical line (common in output logs), producing a React duplicate-key warning and unstable reconciliation. Use the index.
- `apps/packages/product-client/src/stores/toast/toast-store.ts` — the legacy store is kept as a shim with `toasts: []` forever and `dismiss: () => {}` as a no-op. That's the duplicate-old-path shape AGENTS.md rules out. With ~75 call sites on `state.show` I understand not doing it here, but the no-op `dismiss` will silently swallow a real dismissal intent — worth either a follow-up issue linked in the code or removing `dismiss` from the interface so callers fail to compile.
- Accessibility: `ToneDot` is `aria-hidden="true"` (`ToastBody.tsx`), so for a `status` toast the tone is the *only* severity signal and it is invisible to assistive tech — the message text carries no "Failed"/"Done" marker of its own. Separately, the action cluster renders inside sonner's `data-title` slot, which is a live region, so buttons get re-announced on re-render, and a persistent error toast takes no focus, so a keyboard user has no announced route to Retry.
- `scripts/check_toast_copy.py` — the guard scans line by line (`for lineno, line in enumerate(text.splitlines())`), so a multi-line `showToast(\n  \`… ${errorMessage(e)}\`\n)` evades both rules. `SCANNED_ROOTS` also omits `apps/web/src`.
- `apps/packages/product-client/src/lib/domain/chat/models/model-support-refusals.ts` — `modelUnsupportedControlMessage` says the target "needs a newer AnyHarness", which is exactly the remedy prediction the PR body says the copy refuses to make. Relatedly, `hooks/chat/workflows/use-chat-launch-actions.ts:277` headlines "Model not supported on this target" — the specific model and workspace names are in the consequence, but the headline still uses the generic noun the repo copy rule targets.
- `apps/packages/product-client/src/hooks/home/workflows/launch-home-cloud-target.ts:114` — presented as a pure extraction, but the pre-image's two distinct branches (`status === "ready"` returning true with no toast; `projectedSessionId` raising the queued toast) were merged into one `if (result.status === "ready" || projectedSessionId)` with an inner `if (result.status !== "ready")`. The outcomes look equivalent to me; worth you confirming, since the framing invited nobody to check. Same file: the `showToast` dependency is narrowed to `(message: string, kind?: "info") => void`, which is tighter than the real signature for no gain.
- `apps/packages/product-client/src/App.tsx` — `Boolean(useProductHost().desktop?.updater)` is the right gate (`apps/web/src/web-host.ts:130` sets `desktop: null`, and `DesktopProductLifecycleRoot` uses the same shape), and there's no SSR to hydrate in this Vite SPA, so I don't think this regresses Web. Two residuals: it tests *presence* rather than `updater.isSupported()`, so an unsupported packaging still mounts the surface; and no test renders `<App>`, so the gate and the `UpdateRestartDialog` that moved behind it are both uncovered.
- `authenticated-markdown-cascade.test.ts` — the `afterAll` timeout bump 10s→60s is an unrelated flake fix riding along.

### Verified

- The hacks are deleted, not relocated: `[&&]:` appears only in two `Sonner.test.tsx` assertions testing caller override, and `"sonner"` is imported only by `apps/packages/ui/src/primitives/Sonner.tsx`. This is a real replacement, not additive.
- Frame wired once via `toastOptions` on the kit `Toaster`; `visibleToasts={MAX_VISIBLE_TOASTS}` (3) enforces the max-3 rule; same `id` replaces rather than stacks.
- Three weights as a closed discriminated union with no children escape hatch; negative rules enforced by absence of API.
- `cause` never reaches a rendered field — it flows only into `details.payload` (`toast-model.ts` `resolveErrorDetails`, covered by `toast-model.test.ts` and `show-toast.test.tsx`), and whitespace-only `cause` degrades to `{kind:"none"}`. The one exception is the update-flow bypass noted above.
- `details` resolves to exactly three destinations (navigate / modal / none), and `none` renders no button.
- Excerpt caps at 3 countable lines with "+N more"; `readToastPayload` returns `lines: []` for blobs, so a stack trace is structurally unrenderable inline. Copy carries the whole payload, not the excerpt.
- Errors and any action force `Number.POSITIVE_INFINITY` duration; status is 4s, announcement 8s.
- Update toasts fire only at `available` / `stalled` / `ready` (plus the manual-check receipt and the countdown); `checking` and `downloading` are silent, and the countdown both dismisses and suppresses the ready announcement.
- `formatSessionCreateToastMessage` deleted. The four over-threshold files were fixed by real extraction: `scripts/frontend_structure_allowlist.txt` lowers `use-session-creation-actions.ts` 489→451 and removes the stale `use-proposed-plan-actions.ts` entry, with no bumps. The extracted plan files also carry the toast-copy migration, so they belong in this PR rather than being movable out of it.
- The toast-copy guard is wired into CI with its own unittest.

pablonyx:
Superseded by #1598 — the same 13 commits replayed onto `main` after the rest of the design stack squash-merged. Merged as 0c8ad672f.

PR #1589, opened 2026-07-29
fix(billing): only a period boundary mints the period seat allowance (W-F2)

pablonyx:
## What

`_handle_invoice_paid` treated **every** paid invoice carrying a cloud subscription line as a new billing period, minting `pro_period` hours sized at the full seat-month under a `source_ref` keyed on `current_period_start`.

Stripe sends `invoice.paid` for more than renewals. A mid-period seat change produces its own paid invoice with the **same** `current_period_start`, so it collided with that `source_ref` — and because the handler passes `top_up_existing=True`, the raised seat count topped the existing grant up by a full month's hours per added seat.

Those seats already have an allowance. The seat-adjustment pass grants them **pro rata** under `pro_seat_proration`. So the invoice top-up was a second allocation for hours already granted, and it was the *unprorated* one: adding a seat the day before renewal granted a whole extra seat-month of compute.

## Why this is not hypothetical

Prod (`proliferate-prod-server:128`, the live task definition) runs `PRO_BILLING_ENABLED=true`, and there are three active Pro subscriptions on the live Stripe account. The only reason no customer has been over-granted yet is that no live subscription has had a seat added mid-period — every live invoice to date is `subscription_create` or `subscription_cycle`.

This contradicts the "PRO stays off at launch" assumption under which W-F2 was originally filed as backlog, which is why it is fixed now rather than deferred.

## Verified against real Stripe, not just fixtures

All four premises were confirmed on a real Stripe TEST subscription (wire dumps in `codex/billing-e2e-proof-2026-07-29.md`):

- a mid-period seat change raises its **own** paid invoice,
- carrying `billing_reason: subscription_update`,
- with the period **not** advanced, and
- with proration lines that carry the cloud price (so they classify as a cloud line and reach this handler).

The starvation direction was ruled out on a Stripe **test clock**: a trial→paid conversion arrives as `subscription_cycle` with the period advanced, so trials still mint. `trial_period_days`, `trial_end`, `pause_collection`, and `billing_thresholds` appear nowhere in the server outside tests.

## How

Gate the period grant on `billing_reason` being an actual period boundary (`subscription_create` / `subscription_cycle`), *before* the existing grant gate — so `invoice_period_grant_gate_closed` keeps meaning what it says: a boundary invoice that paid and produced no hours.

Two things deliberately still happen on a non-boundary invoice:

- **Payment-failed holds are still cleared.** A past-due customer's retry can settle on a non-boundary invoice; returning early without clearing would leave a paid-up customer paused until their next renewal.
- **A drop is reported** (`invoice_not_period_boundary`), and its **level depends on whether we recognize the reason**:
  - a reason in `BILLING_NON_PERIOD_INVOICE_REASONS` (the proration case) → **info**, no page. This is the one money-bearing drop that is *correct*: money arrived, but it is not unaccounted-for money.
  - anything else, **including the field being absent** → **error + page**. The failure direction of this gate is a paid renewal that mints zero hours for *every* paying org, which is exactly what the module's level policy reserves paging for. Stripe adding a reason, or an API version that stops sending the field, must not silently share the expected path. This is not idle worry: the live account's API version already omits other invoice top-level fields (`paid`, `current_period_start`).

`isinstance` guards the set test — a non-`str` value would raise `TypeError`, which the intake path turns into a failed receipt and a 500 rather than a reported drop.

## Verification

`tests/integration/test_billing_invoice_period_boundary.py` (4 tests) + one parametrized level case in the existing level-policy suite:

| Test | Pins |
| --- | --- |
| `test_mid_period_seat_proration_does_not_re_grant_the_period` | the bug itself |
| `test_period_boundaries_still_mint_the_allowance` (×2, parametrized) | the fix does not starve paying orgs |
| `test_non_boundary_invoice_still_clears_a_payment_failed_hold` | dunning recovery survives the gate |
| `test_invoice_not_period_boundary_level_follows_reason_recognition` (×3) | recognized → info; unrecognized and **missing** → page |

**Negative controls** (both run; the gate is not merely exercised, it is pinned):

- Reverting the gate → **5 failures**, led by `assert 15.0 == 10.0 ± 1.0e-05` / *"a mid-period proration re-granted a full seat-month of compute"*.
- Making the reporter always report the drop as `expected` → fails the `unrecognized` and `missing` cases (`assert 20 == 40`) and leaves `recognized` green.

51 passed across the five billing/webhook suites touched by this branch (`test_billing_invoice_period_boundary`, `test_stripe_webhooks`, `test_stripe_webhook_drop_visibility`, `test_billing_accounting`, `test_billing_llm_seat_pool`); `ruff check` + `ruff format --check` clean; repo-shape and docs-integrity checks green.

**Rebased onto #1586** (`512b68ec6`), which had landed on the same three files and flipped this PR to `CONFLICTING`. The resolution is purely additive against main in both Python modules — #1586's `report_checkout_subject_unresolved` change and this PR's `report_invoice_not_period_boundary` rewrite are independent. All 51 tests re-run green on the rebased head.

Existing invoice fixtures gained the `billing_reason` field that real Stripe payloads always carry — without it, three tests took the new early return and stopped covering the branch they exist to cover.

## Spec

Documented as **M2a** in `specs/codebase/platforms/product/billing.md`, together with the invariant this creates: the full-period grant is gated on `pro_billing_enabled` alone, but the prorated seat grant *also* needs `run_background_workers` and `cloud_billing_mode != off`, because it runs inside the reconciler's accounting pass. The invoice top-up used to mask that asymmetry. **Prod is safe here** — `RUN_BACKGROUND_WORKERS` is unset and defaults to `True` (`config.py:316`), so the reconciler does run — but any deployment turning Pro pricing on must also run the reconciler.

The same pass corrects a pre-existing overstatement in that spec section: only the **compute** allocation is topped up for a mid-period seat. The LLM pool allocation inserts `on_conflict_do_nothing` on its period key and the seat pass issues no LLM grant, so a seat added mid-period gets prorated compute and no LLM allocation until renewal. That predates this change and is documented, not fixed, here.

## Second review pass — found a real over-grant this PR does *not* fix (#1590)

The first review returned REQUEST CHANGES (8 findings, all applied). The second, on the applied result, returned **APPROVE** with one must-fix on wording — and that wording turned out to be hiding a real bug.

I had annotated the `assert grants[0].hours_granted == 15.0` in `test_org_pro_subscription_sync_reconciles_active_seats_before_period_grant` as *correct*: a re-delivered **boundary** invoice converging on the allowance that boundary now owes. The first half holds — it genuinely is not the W-F2 trigger. The second half is wrong. That third seat was added **mid-period**, so the boundary owes it a prorated fraction, and the seat-adjustment pass already grants that fraction under a `source_ref` that does not dedupe against the period grant. Net: **~17.5 h granted where ~12.5 h is owed — one extra seat-month per seat added mid-period.** Same arithmetic as W-F2, reached by re-delivery instead of `subscription_update`.

Filed as **#1590**, not fixed here: sizing the period grant from the boundary seat count is a production behaviour change needing its own verification, and folding it in would make this gate unreviewable as one mechanism. Reachability is narrow — the receipt table refuses re-claim once `status == "processed"`, so it needs a receipt left `failed` *and* a seat added before the retry lands. The assertion now pins 15.0 as **current** behaviour with the residual named, rather than reading as a blessing. W-F6 (mid-period seats get no LLM allocation) filed as **#1591**.

**Declined one suggestion.** Review verified the real `invoice.billing_reason` enum against Stripe's OpenAPI spec (2026-06-24.dahlia) — nine values, my two sets covered eight — and suggested adding the missing `subscription` to the benign set. I did not: it was Stripe's pre-May-2018 catch-all that did *not* distinguish cycles from updates, so it could stand in for a real cycle, and calling it benign would silently skip a period grant — the one failure direction this split exists to prevent. It stays unrecognized and pages, with the reasoning in the constant. It cannot occur here anyway (every subscription we have was created in 2026).

**CodeQL is now clean, and my first diagnosis was incomplete.** The 3 high alerts survived the rebase, so they were live rather than stale. The dataflow path showed the remaining taint source was not production code at all — it was the `billing_reason` *parameter* of the `_invoice` helper in the new test file, since the rule regex-matches the literal token "billing". Renamed to `invoice_reason`; the payload key stays `billing_reason` because that is Stripe's wire field. Zero open alerts on `refs/pull/1589/merge`.

Review also flagged the log-field rename as a possible break for saved Sentry searches — checked, nothing in the repo consumes the field (it is new in this stack), so it is a non-issue.

## Risk

Behaviour change is confined to non-boundary paid invoices, which previously granted a full seat-month and now grant nothing. Boundary invoices are unchanged and covered by both new and pre-existing tests.


pablonyx:
## Verified against real Stripe TEST, not just fixtures

This fix turns entirely on what Stripe actually puts in `billing_reason` and
whether the period advances. Fixtures are written by whoever wrote the bug, so
I drove both sequences through the Stripe TEST API and read the wire.

### 1. The bug reproduces exactly as described

A 2-seat subscription, then the exact call our worker makes —
`subscription_items update quantity=3 proration_behavior=always_invoice`
(`integrations/stripe/client.py:414`):

```
in_…TEA2  billing_reason=subscription_create  status=paid  total=4000
   qty=2 +4000  period={start: 1785293028, end: 1787971428}

in_…TbA2  billing_reason=subscription_update  status=paid  total=2000
   qty=2 -4000  price=price_1TyNSc…   (credit old seats)
   qty=3 +6000  price=price_1TyNSc…   (charge new seats)

subscription ITEM current_period_start after the bump: 1785293028  ← UNCHANGED
```

Four premises, now receipts rather than assumptions:

1. The seat bump produces a separate **paid** invoice, so it reaches `_handle_invoice_paid`.
2. Its `billing_reason` is `subscription_update` — the discriminator this PR gates on exists on the wire.
3. **The period does not advance.** This is the crux: `pro_period_grant_source_ref` keys on that timestamp, so the proration invoice collided with the renewal's grant row and `top_up_existing=True` raised it to a full 3-seat month.
4. The proration lines carry the cloud subscription price, so `line_items_include_price` matched and the old code proceeded.

One correction to the original audit write-up while I'm here: it described the
double grant as two *distinct stacked* grant rows. It is **one row, topped up**.
The arithmetic (12.5 h intended vs 17.5 h actual) was right; the shape wasn't.

### 2. The starvation direction is ruled out

A gate that is too strict is worse than the bug — it would leave paying orgs on
zero hours. Worst candidate: a trial converting to paid. Tested on a Stripe
**test clock** (2 seats, `trial_period_days=3`, advanced past trial end):

```
trialing -> active
in_…ewxdn3X  subscription_create  paid  total=0     ← trial window
in_…SQOh1asD subscription_cycle   paid  total=4000  ← first paid period
item period advanced: 1785552190 -> 1788230590
```

**The conversion invoice is `subscription_cycle`** — inside the allowed set, and
its period genuinely advances, so it mints a fresh grant under a new
`source_ref`. Trials are safe.

Static check alongside it: `trial_period_days`, `trial_end`, `pause_collection`,
and `billing_thresholds` appear **nowhere** in the server outside tests. The
remaining risky reasons (`subscription_threshold`, resume-after-pause) aren't
reachable from our own code — only from a manual Dashboard action. That's the
known residual of this gate.

### 3. An unrelated finding, worth its own follow-up

The live payload has **no top-level `paid` field at all** (`status=paid`,
`paid=None`), and **no top-level `current_period_start`** — it exists only
per-item. Our `invoice_is_paid` survives the first only by treating a missing
marker as paid and falling through to `status`; `subscription_period`'s
item-level fallback (`domain/webhooks.py:203-229`) is therefore the branch that
actually runs in production, not the dead one.

Both are correct today but load-bearing and unpinned. This promotes the
backlogged **Stripe-Version pin** (we send no version header anywhere) from
"cheap hygiene" to the cheapest real risk reduction left: an account-default API
migration is an untested payload-shape change underneath a money-in gate.

All Stripe work was TEST mode. Objects cleaned up: subscriptions canceled,
customers and test clock deleted, products archived.

PR #1599, opened 2026-07-31
fix(anyharness): re-acquire the probe-engine lock once its holder exits

pablonyx:
## Problem

The probe-engine lock (`model_snapshot/lock.rs`) is tried exactly once, at `ModelSnapshotService` construction. A runtime that loses that race degrades to read-only **forever**, even after the lock's holder exits.

Hit in the wild today: an `anyharness serve` sidecar orphaned by a dead desktop-app instance (parent pid 1, alive since 01:01) kept holding the flock on `~/.proliferate/anyharness/.probe-engine.lock`. The freshly launched app's sidecar booted read-only, and every manual model refresh in Settings → Agents answered 409 — the *"this runtime does not hold the probe-engine lock"* toast — with no recovery short of restarting the app after killing the orphan.

## Fix

Ownership now follows the lock's actual availability instead of freezing the boot-time outcome:

- `engine_lock` becomes a re-acquirable slot (`Mutex<Option<ProbeEngineLock>>`).
- Every probe entry point — `poke_all`/`poke_harness`/`poke_harnesses`/`refresh_now` — goes through `ensure_owner()`, which retries the non-blocking acquisition when the slot is empty.
- Late acquisition runs the same **orphan sweep** construction runs: a dead owner's abandoned scratch can hold a copy of real credential material (`~/.codex/auth.json`), and after its death no other process is left to reclaim it.

The two-live-runtimes ruling (model-catalog.md, spec bullet updated) is untouched: while another process holds the flock, the retry loses and the typed `PROBE_ENGINE_NOT_OWNER` refusal stands. The existing B7 lock proof passes unchanged.

## Tests

- New `a_read_only_engine_reacquires_the_lock_once_the_owner_exits`: refusal while the owner lives, then — after the owner drops — the next forced refresh acquires the lock, flips the mode to `owner`, probes, and sweeps a pre-planted abandoned scratch root.
- Negative control: the new test fails against the unfixed `mod.rs` (verified via stash), passes with the fix.
- Full `model_snapshot` suite: 48/48 green.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/domains/agents/model_snapshot/engine_tests/lifecycle_tests.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/model_snapshot/engine_tests/lifecycle_tests.rs
@@ -375,6 +375,65 @@ async fn only_one_runtime_owns_the_probe_engine_for_a_home() {
     );
 }
 
+/// **Proof B7 (re-acquisition).** Losing the startup lock race is per-attempt,
+/// not for life. While the owner lives, the loser's forced refresh is the typed
+/// refusal and retrying never steals the lock; the moment the owner is gone
+/// (`drop` here — a killed process gets the same flock release from the OS),
+/// the loser's next trigger acquires the lock, runs the owner's orphan sweep,
+/// and probes. No restart required.
+///
+/// The wedge this pins: a sidecar orphaned by a dead desktop app held the lock,
+/// so the next app launch's runtime booted read-only and every manual refresh
+/// answered 409 ("this runtime does not hold the probe-engine lock") for the
+/// rest of its life — even after the orphan was killed and the lock was free.
+#[tokio::test]
+async fn a_read_only_engine_reacquires_the_lock_once_the_owner_exits() {
+    let home = seeded_home("engine-lock-reacquire", "opencode");
+    let (owner, _owner_runner, _owner_plan) = engine(&home, "opencode", test_config());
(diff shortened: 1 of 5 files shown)


pablonyx:
Dedicated agent review completed on dca82ada9 (opus): no blockers, no majors. Applied in a54f45c1e: poke-path self-heal test (the event pokes are the common recovery paths), sweep doc drift in scratch.rs + model-catalog.md, and `sweep_orphan_scratch` → `pub(crate)`.

Skipped, deliberately:
- **spawn_blocking for the late sweep** — one age+pid-gated directory listing on a rare transition; not worth the seam.
- **eager `ensure_owner` before the cheap filters** — acquiring ownership is durable and desired regardless of whether this particular poke proceeds; reordering would also flip the NotOwner/NotInstalled precedence on the refresh route.
- **per-poke warn on a sealed read-only home** — volume is bounded by the closed event set; transition-only logging adds state for little.
- **role inversion nit** (whichever runtime pokes first after the owner dies wins the lock) — follows directly from the spec's one-engine ruling; nothing privileges the desktop over a sidecar by design.

Suite: 49/49 model_snapshot tests green; negative control verified (new tests fail on the unfixed `mod.rs`).

pablonyx:
Closing as stale (founder triage 2026-08-07): dormant since late July and now conflicting with the merged server-grid / frontend-fold / docs-restructure changes. The branch is preserved — revive by reopening or re-cutting against current main if the fix is still wanted.

PR #1601, opened 2026-07-31
fix(tests): repair MarkdownBody link-treatment assertion broken by #1561

pablonyx:
**Main is red**: `Shared frontend packages` fails on every PR since #1561 merged (last night, 00:04).

#1561 retuned `CHAT_TRANSCRIPT_LINK_CLASS` in `TranscriptLinkStyles.ts` — `hover:bg-transparent` now sits between `no-underline` and `hover:text-link-foreground` — without updating `MarkdownBody.test.tsx:127`, which asserted the old class run as one contiguous substring.

Fix: assert the shared constant itself instead of a hand-copied string. The test's own name states the property — *"uses the **shared** stable-color, hover-underline treatment"* — so importing `CHAT_TRANSCRIPT_LINK_CLASS` pins exactly that: a future retune of the treatment passes, the anchor abandoning the shared treatment fails. The `hover:text-foreground` negative guard stays.

Found while shepherding #1599 (whose frontend tree is byte-identical to main and hit the same red check); reproduced locally on main's tree, green after this change (6/6).

### DIFF
--- a/apps/packages/product-ui/src/chat/transcript/MarkdownBody.test.tsx
+++ b/apps/packages/product-ui/src/chat/transcript/MarkdownBody.test.tsx
@@ -10,6 +10,7 @@ import {
   MarkdownBody,
   type MarkdownLinkRenderInput,
 } from "./MarkdownBody";
+import { CHAT_TRANSCRIPT_LINK_CLASS } from "./TranscriptLinkStyles";
 
 const COMPLETE_MARKDOWN = `# Heading one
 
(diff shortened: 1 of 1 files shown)


pablonyx:
Dedicated agent review completed on fb87b1b7d (sonnet): no defects.

- Not tautological: the test still exercises a real render; it fails if `ProviderLinkMention`/`MarkdownAnchor` stop applying the shared constant, only no longer on value-drift of the treatment itself (intended).
- `toContain` on the full constant is safe on this path: both anchor renderers concatenate the constant verbatim — no `twMerge`/`cn()` touches it. (Noted for the future: `FileReferenceBadge`/`ComposerRichTextEditor` DO route it through `twMerge`, so don't copy this assertion pattern there.)
- No other tests assert the stale fragment; the loose substring assertions in `ProviderLinkMention.test.tsx` and `ComposerRichTextEditor.test.tsx` remain valid.

PR #1609, opened 2026-08-05
chore(product): enforce ProductClient boundaries

pablonyx:
## Summary

- make ProductClient a first-class app-shaped root in the frontend boundary checker
- share import-aware TypeScript parsing across boundary and structure checks
- enforce exact layer, query/cache, raw-access, and external store-mutation debt ratchets without slack
- reject new ProductClient junk-drawer filenames while excluding comments, fixtures, display strings, and regex literals from import facts

This is Slice 1 of the stacked frontend package-consolidation program. It changes enforcement only; no product source or runtime behavior changes.

## Readiness

- [x] I followed the [pull-request procedure](https://github.com/proliferate-ai/proliferate/blob/main/specs/developing/process/pull-requests.md) for scope, title, labels, body, and readiness.

- [ ] If this PR has support relationships, I linked them through the tracker API. No tracker, report, user, or support IDs or private source data appear in this PR.

## Verification

- `python3 -m unittest scripts/test_check_frontend_boundaries.py` (51 test methods)
- `python3 scripts/report_frontend_structure.py --strict --summary-only` (2,261 files, 0 violations)
- `python3 scripts/check_docs.py` (227 Markdown files)
- `python3 -m py_compile scripts/frontend_imports.py scripts/check_frontend_boundaries.py scripts/report_frontend_structure.py scripts/test_check_frontend_boundaries.py`
- `git diff --check`
- two independent frozen-spec reviews accepted exact head `77c63a7c9eda26186f56dbba5c178ce2f3c7e501`

### DIFF
--- a/scripts/check_frontend_boundaries.py
+++ b/scripts/check_frontend_boundaries.py
@@ -458,104 +458,61 @@ def check_file(path: Path) -> list[Violation]:
 
 
 def statement_identifiers(statement: ImportStatement) -> set[str]:
-    return set(statement.imported_names) | {
-        token.value
-        for token in tokenize_typescript(statement.statement)
-        if token.kind == "identifier"
-    }
+    """Return the module's exported names, not their local aliases.
+
+    Identifier-sensitive rules describe external capabilities.  A local alias
+    with the same spelling as a forbidden export is not that capability, while
+    renaming the real export must not hide it.  The shared import parser owns
+    that distinction for static, re-export, and dynamic forms.
+    """
+
+    return set(statement.imported_names)
 
 
 def imported_bindings(statement: ImportStatement) -> set[str]:
-    tokens = tokenize_typescript(statement.statement)
-    bindings = set(statement.local_bindings)
-    if not tokens or tokens[0].value != "import":
-        return bindings
-    try:
-        from_index = next(
-            index
-            for index, token in enumerate(tokens[:-1])
-            if token.value == "from" and tokens[index + 1].kind == "string"
-        )
-    except StopIteration:
-        return bindings
-
-    clause = tokens[1:from_index]
-    if clause and clause[0].value == "type":
-        clause = clause[1:]
-    if clause and clause[0].kind == "identifier":
-        bindings.add(clause[0].value)
-
-    for index, token in enumerate(clause):
-        if token.value == "*" and index + 2 < len(clause) and clause[index + 1].value == "as":
-            bindings.add(clause[index + 2].value)
-
-    try:
-        open_index = next(index for index, token in enumerate(clause) if token.value == "{")
-        close_index = max(index for index, token in enumerate(clause) if token.value == "}")
-    except (StopIteration, ValueError):
-        return bindings
-
-    current: list = []
-    entries: list[list] = []
-    for token in clause[open_index + 1 : close_index]:
-        if token.value == ",":
-            if current:
-                entries.append(current)
-            current = []
-        else:
-            current.append(token)
-    if current:
-        entries.append(current)
-
-    for entry in entries:
-        if entry and entry[0].value == "type":
-            entry = entry[1:]
-        if not entry:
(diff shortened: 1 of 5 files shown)


pablonyx:
## Adversarial review — 1 confirmed finding

Review of head `77c63a7c9` by two independent finder agents (parser-correctness, coverage-vs-claims), with every candidate finding re-derived from git by two independent verifiers instructed to refute it. One finding survived; one other candidate (store-alias reassignment tracking) was refuted as immaterial by design (documented conservative under-approximation, zero real-world instances).

### Confirmed: dynamic `import()` inside a function with a *generic* return-type annotation is misclassified as type-only

**Where:** `scripts/frontend_imports.py:1140` (`_function_return_annotation_contains_import`, invoked at :1621).

**Mechanism:** for a function declaration or class method whose return annotation contains a generic (`Promise<...>`, `Record<string, unknown>`, …), any dynamic `import()` in the body is attributed to the annotation and reported `type_only=True`, and `local_bindings`/`namespace_bindings` are emptied (frontend_imports.py:2242-2244).

```ts
// stores/x/y.ts — produces ZERO violations at this head:
export async function loadRQ(): Promise<unknown> {
  const { QueryClient } = await import("@tanstack/react-query");
  return new QueryClient();
}
// identical function WITHOUT the return annotation → STORE_RUNTIME_ACCESS fires
```

Both verifiers reproduced this end-to-end through `find_product_client_violations` (fixture tree, negative control confirmed: removing the annotation makes the rule fire).

**Verified scope (both wider claims were checked and rejected):**
- Trigger requires a *generic* annotation. `: Result`, `: void`, `: A | B`, `: Result[]`, `: { a: string }` classify correctly; `: Promise<unknown>`, `: Promise<void>`, `: Record<string, unknown>` break.
- Affects `function` declarations, class methods, and object-method shorthand. Arrow functions are **not** affected.
- Collateral suppression of other rules was disproven: `imported_names` survives the type-only branch, so `ANYHARNESS_CLIENT_OUTSIDE_ACCESS`, `QUERY_HOOK_OUTSIDE_ACCESS_OR_CACHE`, and `PRODUCT_CLIENT_LAYER_DIRECTION` still fire on the misclassified shape. Only `STORE_RUNTIME_ACCESS` gates on `type_only` (check_frontend_boundaries.py:2251).

**Second affected consumer:** `scripts/report_frontend_structure.py:534` (also a hard CI gate). The same misclassification drops the "product-domain may import Cloud SDK contract *types*, not value clients" violation — `export async function load(): Promise<void> { const m = await import("@proliferate/cloud-sdk"); }` returns no violation; the un-annotated form does. Same for the product-surfaces raw-value-client rule at :558.

**Materiality:** no live escape today. A differential sweep over all 3,097 TS/TSX files at this head found 64 dynamic imports currently misclassified as type-only, but force-flipping all of them to runtime surfaces zero new violations (`stores/**` contains no dynamic imports at all). So CI is green for the right reasons *right now* — the defect is a forward ratchet hole, and the bypass shape (`async function load(): Promise<...>` wrapping `await import(...)`) is the most idiomatic way a future author would write exactly the code `STORE_RUNTIME_ACCESS` exists to stop.

**Test-suite blind spot:** `scripts/test_check_frontend_boundaries.py:461` asserts `function f(): Result { return import("pkg") }` is runtime — it passes only because `Result` takes no type arguments, so the suite believes it covers this case while leaving the generic form untested.

**Suggested fix:** treat the return annotation as ended once its top-level generic closes and the next token opens the function-body block — i.e. require the token immediately after the annotation's closing position to be the body's `{`, or bail out of `_type_annotation_prefix_reaches_import` at that boundary. Add negative-control tests for `Promise<unknown>` / `Record<string, unknown>` on function declarations and class methods.

---
🤖 Adversarial review workflow (2 finders + 2-skeptic verification per finding) via [Claude Code](https://claude.com/claude-code)

PR #1651, opened 2026-08-05
chore(scripts): seven AnyHarness grid boundary rules with seeded debt allowlist

pablonyx:
PR 1 of the AnyHarness grid program (plan: `anyharness-grid-plan.md` on `worktree-anyharness-structure-doc`; law: `anyharness-structure.md` same branch). Python only — no Rust changes, no behavior changes.

## What

Extends `scripts/check_anyharness_boundaries.py` with eight rule families and seeds `scripts/anyharness_boundaries_allowlist.txt` with today's measured debt (123 data rows covering 229 seeded occurrences), so every wall in the structure doc exists in CI from this PR on and cleanup PRs prove progress by shrinking the allowlist (stale counts already fail the build).

New rules (seed counts post-hardening):

- `DOMAIN_LIVE_VALVE` — `crate::live` (import or inline path) in `domains/**` only from `runtime.rs` / `runtime/**` / `live_ports.rs`. Shapes exception: `crate::live::<area>::model::*` imports are legal anywhere (the sanctioned observer-trait inversion). Seeds: 12 files (mobility + materialization services and 7 stragglers).
- `DOMAIN_VALVE_LIVE_REEXPORT` — `pub use crate::live::<power>` inside a valve file (`::model` exempt): the valve may hold live machinery, not republish it domain-wide. Zero debt today; hard rule from day one.
- `LIVE_DOMAIN_STORE_IMPORT` — `live/**` must not import or inline-construct any domain's store/service (use-statement walk plus a per-line pass for `*Store::new(...)`). Seeds: 19 occurrences — `live/terminals` files (older doctrine, deliberately retained, marked "ratchet only"), `live/sessions/probe.rs`, one `SessionStore::new` in a cfg(test) mod the engine can't see past.
- `API_STORE_ESCAPE` — `.store()` calls, `*Store::new`, store imports, or `state.*_store` AppState-field reaches under `api/**`. Seeds: 16 occurrences / 6 files (5 real handler escapes targeted by grid plan PR 3, plus `health.rs` reading `state.agent_seed_store` — an in-memory `Arc<RwLock<AgentSeedHealth>>` snapshot, not a DB store; seeded as a benign false positive to keep the field-name signal).
- `POLICY_PURITY` — clocks (`Utc::now` etc.), `Uuid::new_v4`, `rand::`, store/adapters imports inside `*_policy.rs` or files named exactly `policy.rs`. Deliberately NOT `&self` (over-fires on Display impls / plain data methods — stays a review rule). Seeds: 3 files incl. `workspaces/retention_policy.rs` and `workflows/control/policy.rs`.
- `DOMAIN_STORE_API_IMPORT` / `DOMAIN_STORE_LIVE_IMPORT` — the sessions-store rule generalized to every domain's store module (the `SESSION_STORE_*` rule IDs are folded in and removed).
- `DOMAIN_SQL_OUTSIDE_STORE` — SQL text in domain files outside store modules, 12 patterns including split-line forms (`SELECT`/`FROM` on separate lines, `UPDATE` + leading `SET`, `INSERT OR … INTO`, `DROP TABLE`). Seeds: 52 occurrences across 9 files.
- `DOMAIN_CONTRACT_IMPORT` — any `anyharness_contract` path in `domains/**`, use-statements AND inline qualified paths (the narrower `DOMAIN_CONTRACT_REQUEST_RESPONSE` rule is unchanged). Seeds: 116 occurrences across 85 files.

Every seed row's reason names its cleanup PR in the grid plan.

## Hardening (post-review, same PR)

An adversarial review of the initial cut produced 9 findings; 6 became code fixes here, 3 are documented as known scope limits in the allowlist header:

- **Multi-line SQL (the P1)**: the SQL rule matched single-line statements only, while 31% of in-tree SELECTs and 88% of UPDATEs split across lines (e.g. `retention_policy.rs:29-31`). Pattern set widened 7→12; seeds re-measured 39→52.
- **AppState store fields**: `state.*_store` reaches in `api/**` were invisible (imports/constructors only). New third disjunct of `API_STORE_ESCAPE`.
- **Valve re-exports**: a valve file could launder `crate::live` power via `pub use`. New `DOMAIN_VALVE_LIVE_REEXPORT` rule, zero debt.
- **Inline qualified paths**: `anyharness_contract::Foo` in expression position bypassed the use-statement walk. Contract rule now runs a guarded line pass; seeds 86→116.
- **Inline store construction in `live/**`**: same gap, same fix; one real find seeded (`background_work/mod.rs:316`, cfg(test)).
- **Exact `policy.rs`**: purity rule matched only `*_policy.rs`; `workflows/control/policy.rs` was invisible.

Known scope limits (allowlist header): store-by-role files (`*_store.rs` outside `store/`) are policed for SQL but invisible to the importable-store rules; counts are per-line so a rustfmt reflow may legitimately shift one; AppState-field access beyond the `state.*_store` shape stays judgment; `"UPDATE failed"` vs `"UPDATE sessions"` is lexically indistinguishable so uppercase-SQL-verb-in-string false positives are accepted rather than chased (none exist in the current tree).

## Delta re-review closure (same PR)

The hardening itself was adversarially re-reviewed (verdict: no P1s; all seed re-derivations exact; byte-identical output on untouched rules). Its three latent P2 findings — none triggered by any code in today's tree — are closed in the final commit, under a proven invariant: real-tree checker output is byte-identical before/after (sha256-equal violation dumps, 229 = 229; allowlist rows untouched), so stacked PRs need no base-sync.

- **Root-brace-group `use {…}` double-count**: the inline pass now skips the exact line spans consumed by use statements instead of re-testing line prefixes — `use {\n anyharness_contract::…,\n};` counts once, not twice.
- **Split `pub use` evasion of the valve re-export rule**: `pub use\n crate::live::…;` previously started no use statement at all (`use\s+` required a same-line path) and was invisible to the entire import pass. Both start regexes now end at `use\b` and the re-export check matches the joined statement text.
- **Uppercase-SQL-in-string false positives**: the anchored SELECT pattern gained a `(?!")` exclusion killing `const KEYWORD: &str = "SELECT";` / `#[serde(rename = "SELECT")]` shapes at zero lost detection (measured: `"SELECT " + cols` real queries still match); the indistinguishable `"UPDATE failed"` shape is documented as a known limit with behavior-pinning tests rather than chased.

## Tests

`scripts/test_check_anyharness_boundaries.py` — 90 tests (45 initial + 33 hardening + 12 P2 closure, stdlib unittest, temp-tree fixtures in the style of `test_check_frontend_boundaries.py`; the checker's scan root was made injectable for tests, CLI behavior unchanged). Wired into the CI step ahead of the checker run, mirroring the frontend step.

Verified independently: checker exit 0 on the seeded tree; 90/90 tests green; negative controls run per rule (remove a seed → exit 1 naming the exact file:line; restore → exit 0) and per P2 fix (monkeypatch-revert each fix → its new tests fail with the exact predicted error; restore → green); byte-identical output on all pre-existing rules before/after hardening AND before/after the P2 closure; 15/15 mutants killed by the suite in review.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,9 @@ jobs:
         run: python3 scripts/check_server_boundaries.py
 
       - name: Check AnyHarness layer boundaries
-        run: python3 scripts/check_anyharness_boundaries.py
+        run: |
+          python3 -m unittest scripts/test_check_anyharness_boundaries.py
+          python3 scripts/check_anyharness_boundaries.py
 
       - name: Check session mutation admission
         run: python3 scripts/check_session_mutation_admission.py
--- a/scripts/anyharness_boundaries_allowlist.txt
+++ b/scripts/anyharness_boundaries_allowlist.txt
@@ -5,6 +5,140 @@
 #
 # This is migration debt, not permission for new code. If a cleanup reduces a
 # count, update this file in the same PR.
+#
+# The checker cannot see `#[cfg(test)]` boundaries inside otherwise-production
+# files, so a handful of counts below cover setup code in inline test mods. Those
+# entries say so in their reason and are not cleanup targets.
 
 DOMAIN_CONTRACT_REQUEST_RESPONSE anyharness/crates/anyharness-lib/src/domains/goals/runtime.rs 1 strict-mirror goal runtime uses SetSessionGoalRequest to drive the sidecar ext method; contract mapping pending move above the API boundary
 DOMAIN_CONTRACT_REQUEST_RESPONSE anyharness/crates/anyharness-lib/src/domains/loops/runtime.rs 1 strict-mirror loop runtime uses SetSessionLoopRequest to drive the sidecar ext method; contract mapping pending move above the API boundary
+
+# Rule: domains/** reach live/ only via the runtime valve.
+# 12 files, 21 occurrences.
+DOMAIN_LIVE_VALVE anyharness/crates/anyharness-lib/src/domains/agents/auth/login_terminal.rs 4 agent login holds the live login-terminal service directly (4 leaves in one group import); target = agents runtime valve (grid plan PR 6c)
+DOMAIN_LIVE_VALVE anyharness/crates/anyharness-lib/src/domains/agents/model_snapshot/entry.rs 1 inline fully-qualified live probe type in a fn signature; target = agents runtime valve (grid plan PR 6c)
+DOMAIN_LIVE_VALVE anyharness/crates/anyharness-lib/src/domains/agents/model_snapshot/probe.rs 3 model snapshot calls live session probe directly (3 leaves in one group import); target = agents runtime valve (grid plan PR 6c)
+DOMAIN_LIVE_VALVE anyharness/crates/anyharness-lib/src/domains/agents/model_snapshot/test_support.rs 3 test-fixture support file (not excluded by should_skip) builds live probe snapshots; target = agents runtime valve or move under tests/ (grid plan PR 6c)
+DOMAIN_LIVE_VALVE anyharness/crates/anyharness-lib/src/domains/materialization/service.rs 1 materialization service holds live terminals directly; target = materialization runtime valve (grid plan PR 6a)
+DOMAIN_LIVE_VALVE anyharness/crates/anyharness-lib/src/domains/mobility/service.rs 1 mobility service holds live terminals directly; target = mobility runtime valve (grid plan PR 6a)
+DOMAIN_LIVE_VALVE anyharness/crates/anyharness-lib/src/domains/plans/decision_op.rs 2 plan decision op imports actor command types (Resolution, ResolveInteractionCommandError) via the live::sessions root re-export, not ::model; target = model re-export or plans runtime valve (grid plan PR 6e)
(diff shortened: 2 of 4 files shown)


pablonyx:
**Delta re-review complete — ready for @pablonyx.** Adversarial review of the hardening delta (`cbf5fcae5..8a6de885d`) returned **no P1s**: all 8 claimed fixes verified implemented-with-tests, seed re-derivations 8/8 exact (incl. completions.rs 30/30, retire_preflight.rs contract 7/7), byte-identical violations on all untouched rules, 4/4 negative controls fired with exact file:line (incl. a live-tree probe of the zero-seed valve re-export rule). Its 3 latent P2s (root-group `use{}` double-count, split `pub use` evasion — which turned out to make bare `pub use` lines invisible to the whole import pass, uppercase-SQL-in-string FPs) are closed at head `87147cce2` under a proven byte-identity invariant: sha256-equal real-tree violation dumps (229=229), allowlist rows untouched, so the stacked PRs (#1654, #1657) need no base-sync. Tests 90/90; each P2 fix negative-controlled by monkeypatch-revert. Review chain: initial cut reviewed (F1–F9, 15/15 mutants killed) → hardening → delta re-review → P2 closure.

PR #1657, opened 2026-08-05
refactor(anyharness): fold workspaces SQL into store modules, split retention policy (grid PR 5a)

pablonyx:
## Summary

Grid plan PR 8c: retires the last workspaces-domain rows on `DOMAIN_SQL_OUTSIDE_STORE` and `POLICY_PURITY` by relocating SQL to `store/**` and splitting a misnamed policy into a pure decision module plus a store.

**(a) SQL folds**

- `workspaces/access_store.rs` → `workspaces/store/access.rs` via `git mv`. This was a relocate-not-copy: the file was already a store by *role* (owns `WorkspaceAccessStore`, all its SQL), just wrong by *address* — it sat outside `store/`, so `DOMAIN_SQL_OUTSIDE_STORE` policed it even though the importable-store rules never treated it as store code. Moving it under `store/` retires the SQL row and enrolls it in the store rules at the same time.
- `workspaces/retention_policy.rs`'s embedded SQL → new `workspaces/store/retention.rs` (`WorktreeRetentionPolicyStore`: `get_policy`/`upsert_policy` against the `worktree_retention_policy` singleton row).
- The triplicated repo-root test fixture (an inline `INSERT INTO repo_roots (...)` block copy-pasted across workspaces test modules) → `domains/repo_roots/test_support.rs`, routed through `RepoRootStore::insert`. `workspaces/access_gate.rs`'s test module now calls `seed_repo_root_1(&db)` instead of hand-rolling the insert.

**(b) Policy split**

`retention_policy.rs` is now pure: `decide_policy_update(cap, now)`, `validate_max_materialized_worktrees_per_repo`, and `keep_count` — no store, no clock, no `&self`. The `now` clock reading is threaded in by the caller (`WorkspaceRetentionService`, the only clock-reader), not read inside the policy. 10 unit tests cover both boundaries, one-below/one-above rejection, the default cap, and that `updated_at` stamps the caller's clock rather than sampling its own. Negative control: inverted the bounds check (`<` → `>`) and reran — 9/10 tests failed as expected; reverted.

**(c) Allowlist ratchet (measured, not predicted)**

5a was cut from an old head of `codex/anyharness-api-store-escapes`, before that branch's checker hardening (12 SQL patterns incl. multi-line heads and `state.*_store` detection; exact `*_policy.rs` matching; 78 unit tests) and allowlist reseed landed. This PR's base moved, so the merge re-measured every affected row against the merged tree with `scripts/check_anyharness_boundaries.py` rather than trusting either side's stale counts:

`DOMAIN_SQL_OUTSIDE_STORE` (workspaces rows, old → new against the hardened base seed):

| file | old | new |
|---|---|---|
| `workspaces/access_store.rs` | 5 | row removed (folded into `store/access.rs`) |
| `workspaces/retention_policy.rs` | 7 | row removed (folded into `store/retention.rs`) |
| `workspaces/inventory.rs` | 1 | row removed (test SQL now via shared `repo_roots::test_support`) |
| `workspaces/access_gate.rs` | 2 | 1 (repo_roots INSERT moved to the shared helper; only the `DROP TABLE` line remains, engine-invisible cfg boundary) |
| stanza header | 9 files / 52 occ | 6 files / 38 occ |

`POLICY_PURITY`:

| file | old | new |
|---|---|---|
| `workspaces/retention_policy.rs` | 2 | row removed (now pure) |
| `workflows/control/policy.rs` | 1 | 1 (unchanged, untouched by this PR) |
| stanza header | 2 files / 3 occ | 1 file / 1 occ |

`API_STORE_ESCAPE`: untouched — exactly the base's 2 rows (`health.rs` 1, `workspaces_purge.rs` 9).

Also repointed one hardened-suite calibration anchor: `test_known_truths_the_hardened_rules_must_see`'s "split SELECT head" example pointed at `workspaces/retention_policy.rs:29`, which is exactly the SQL this PR folds into the exempt `store/retention.rs`. Repointed to a real, still-present split-SELECT occurrence at `sessions/links/completions.rs:128` — no rule logic touched.

**(d) The naming trap**

The store half is `store/retention.rs`, **not** `store/retention_policy.rs`. The checker holds any `*_policy.rs` file under `domains/**` to `POLICY_PURITY` regardless of directory — a store file named `*_policy.rs` under `store/` would still be judged as a policy and fail for holding a DB connection. `store/retention.rs` avoids the trap; `retention_policy.rs` stays the pure module's name at the domain root.

**(e) Verification**

Original build (pre-merge, at `fec4f7055`): `cargo check -p anyharness-lib` clean, `cargo test -p anyharness-lib --lib domains::workspaces::` 95/95, api scopes green. This PR's merge delta is scripts-only (allowlist re-measurement) plus one base-side method rename (`active_branches_by_repo_root_id` → `list_repo_root_active_branches` in `workspaces/runtime/repo_metadata.rs` and its one caller in `api/http/hosting.rs`) that this branch's code never called — verified no remaining references to the old name anywhere in the tree. CI validates the Rust; this merge intentionally did not rebuild it (single-build-slot constraint).

On the merged tree:
- `python3 scripts/check_anyharness_boundaries.py` → exit 0
- `python3 -m unittest scripts.test_check_anyharness_boundaries` → 78/78 OK
- `python3 scripts/check_max_lines.py` → pass
- Negative control: bumped `workspaces/access_gate.rs`'s SQL row 1→2 → checker failed with `stale allowlist count (observed 1, allowed 2)`; restored → exit 0.

## Test plan

- [x] `python3 scripts/check_anyharness_boundaries.py` exits 0 on the merged tree
- [x] `python3 -m unittest scripts.test_check_anyharness_boundaries` — 78/78 passing
- [x] `python3 scripts/check_max_lines.py` passes
- [x] Negative control on the allowlist ratchet (bump/fail/restore/pass)
- [x] `cargo check -p anyharness-lib` and `cargo test -p anyharness-lib --lib domains::workspaces::` (95/95) — from the pre-merge build at `fec4f7055`; CI revalidates on this PR
- [x] Confirmed `.github/workflows/ci.yml` untouched by the merge relative to base


pablonyx:
**Adversarial review complete @ `6845e1f9f` — no P1s; one P2 being fixed on this branch before merge.**

Verified: SQL move fidelity byte-for-byte (all statements MOVED-IDENTICAL incl. both upserts' conflict-set columns; the one body diff in the git-mv is a required import-path line); `retention_policy.rs` genuinely pure (zero imports, zero clock/uuid hits outside doc comments); service is the sole clock-reader (one reading per operation, threaded down — `run_pass` even gained single-reading consistency); visibility NARROWED by the move (module now private behind targeted re-exports); rows don't escape (store returns domain models); allowlist deltas re-derived independently — every claimed count exact (SQL 9/52→6/38, POLICY 2/3→1/1, API_STORE_ESCAPE untouched); anchor repoint was forced, not chosen (`completions.rs:128` is the only split-SELECT head left in the tree); merge hygiene proven (two-dot vs three-dot diff byte-identical). Negative controls: allowlist row removal → checker names `access_gate.rs:324`; `Utc::now()` injected into the pure policy → POLICY_PURITY fires. CI validates the Rust (Cargo check & test green).

**P2 (fix incoming):** pre-refactor, `validate_max_materialized_worktrees_per_repo` was the first statement of the store's only write path — unbypassable. Post-split, `upsert_policy` writes whatever it's handed and the guard lives in the service; deleting the guard breaks no existing test and no checker rule. A service-seam bounds test (reject below-min/above-max AND prove the row wasn't written) closes it.

Nits (recorded, not blocking): the repo-root fixture is 6×-duplicated, this PR folded 3 (rest are checker-invisible test files — ride-along candidate); `seed_repo_root` has one caller; PR body's negative-control prose says 'inverted `<`→`>`' but the guard is a range-contains (the 9/10 failure count is exactly right though — control was real, prose sloppy).

pablonyx:
**Review P2 closed @ `4a7966caa` — with a correction to the review record.**

Added `update_policy_rejects_out_of_bounds_caps_without_persisting_them` at the service seam: MIN−1 and MAX+1 both rejected, and the policy read back through the service is proven unchanged. Negative-controlled properly, which surfaced a fact the review missed: **the P2's failure scenario was overstated.** Migration `0040_worktree_retention_policy.sql` carries an independent DB-level `CHECK (max_materialized_worktrees_per_repo BETWEEN 10 AND 100)` — bypassing only the service guard did NOT fail the test (the DB rejected the write: `Error code 275: constraint failed`). The valid control required bypassing the guard AND stripping the CHECK, which then failed exactly as predicted (persisted 9, no rejection). So the exposure was never a persisted 0 wiping retention — it was a raw SQLite constraint error bubbling up instead of a clean 400. Defense-in-depth is now: pure policy fn (unit-tested) → service seam (this test) → DB CHECK (last resort). Both temporary edits fully reverted; gates green: cargo check clean, retention scope 31/31, boundary checker exit 0, max-lines pass, rustfmt clean.

PR #1658, opened 2026-08-06
refactor(anyharness): give mobility a runtime valve — live power leaves the service (grid PR 6a)

pablonyx:
Grid program PR 6a. `domains/mobility/service.rs` held `use crate::live::terminals::TerminalService`, violating the valve rule: inside `domains/**`, `crate::live` POWER imports are legal only from `runtime.rs` / `runtime/**` / `live_ports.rs`. The recon pass classified the file per method, so this splits it rather than moving it wholesale.

## What moved where, and why

**RUNTIME** — moved to `domains/mobility/runtime/`. Each of these reaches live truth, either directly (live terminals) or laundered through another domain's facade, and facade-laundered live counts as live:

| method | new home |
| --- | --- |
| `prepare_repo_root_destination` | `runtime/prepare_destination.rs` |
| `preflight_workspace` | `runtime/preflight.rs` |
| `install_workspace_archive` | `runtime/install.rs` |
| `destroy_source_workspace` | `runtime/destroy_source.rs` |

The live-terminal helpers (`active_terminals_async`, `active_terminals_blocking`, free `is_active_terminal`) moved to `runtime/mod.rs`, which now holds the sole `crate::live::terminals` import in the knot. The two private validators used only by moved code (`validate_install_preconditions`, `validate_prepared_destination_is_empty`) went with their callers.

**SERVICE** — stayed in `service.rs`, the durable-only use case: `export_workspace_archive`, plus the helpers it needs (`load_workspace`, `can_relocate_existing_archive_session`, `validate_expected_export_runtime_state`) and the pure size-accounting tail with its six unit tests. Helpers the runtime also calls widened to `pub(super)`, no wider.

### The one deviation worth flagging

The service's last live edge was not an import. It held `Arc<SessionRuntime>` solely to call `runtime_home()` — a facade-laundered live dependency that would have survived a literal "move the imports" reading and left the valve unsatisfied. Replaced with a plain `runtime_home: PathBuf` handed in by `app/`, the same value `SessionRuntime` itself receives. A path fact, not live power. This is what makes the service genuinely live-free, and the compiler agrees: after the split, `crate::live::terminals` was reported as an *unused* import in `service.rs`.

## The valve rule, satisfied

`MobilityRuntime` wraps `Arc<MobilityService>` and delegates down — preflight calls `export_workspace_archive` for its size estimate — never the reverse. Requests enter at the highest layer the use case needs.

`AppState` exposes only `mobility_runtime`; export is reached through a thin delegating method rather than a second overlapping handle. Wiring lives in a new per-knot `app/mobility.rs` family, following the existing `sessions.rs` / `product_mcp.rs` / `workflows.rs` precedent. That is what keeps `app/mod.rs` **exactly at its existing 693-line pin** — no cap was raised anywhere in this diff.

## Allowlist ratchets

Both measured with the checkers, not hand-computed.

`scripts/anyharness_boundaries_allowlist.txt` — the row disappears:

```
- DOMAIN_LIVE_VALVE .../domains/mobility/service.rs 1 mobility service holds live terminals directly; target = mobility runtime valve (grid plan PR 6a)
```

Stanza header retuned with it: `# 12 files, 21 occurrences.` → `# 11 files, 20 occurrences.`

`scripts/max_lines_allowlist.txt` — `domains/mobility/service.rs` **1520 → 851**. Still above the 600 cap, so the row stays, re-pinned to its exact new count. `app/mod.rs` stays at 693, unchanged.

## The calibration anchor

`test_mobility_service_is_valved` pinned the very file this PR fixes, so it went red the moment the fix landed. Re-pointed to the next real offender (`domains/agents/auth/login_terminal.rs`, verified to still hold the import) rather than deleted, keeping the test's purpose intact, and strengthened with the other half of the proof: two `assertNotIn` assertions that the cleaned service is no longer flagged **and** that the new valve is exempt even though it holds the identical `crate::live::terminals` import.

## Verification

| gate | result |
| --- | --- |
| `python3 scripts/check_anyharness_boundaries.py` | `AnyHarness boundary check passed.` exit 0 |
| `python3 scripts/check_max_lines.py` | `Max-lines check passed (repo max 600, component max 500, server layer maxes enabled).` exit 0 |
| `python3 -m unittest scripts.test_check_anyharness_boundaries` | `Ran 78 tests` … `OK` |
| `cargo check -p anyharness-lib --all-targets` | clean; no errors, no warnings in touched files |

Test scope derived from `git diff --stat` vs base, not from memory:

| suite | result |
| --- | --- |
| `--lib domains::mobility::` | 6 passed |
| `--lib api::session_admission_tests` | 8 passed — incl. the permit-before-lease source-order guards on the two handlers this PR edits |
| `--lib api::router` | 29 passed — incl. `workspace_mobility_preflight_warns_for_active_terminals_without_blocking`, the end-to-end proof the moved preflight still sees live terminals |
| `--lib app::` | 8 passed — the wiring the new `app/mobility.rs` family produces |

### Negative control

Re-added a `crate::live::terminals` import to the cleaned service. The checker failed closed, naming it exactly:

```
.../domains/mobility/service.rs:14: [DOMAIN_LIVE_VALVE] domain code may reach live/ only through the domain's runtime valve (runtime.rs, runtime/**, live_ports.rs) — promote this use case to the runtime; see anyharness-structure.md
```

The paired anchor assertion flipped with it (`'.../domains/mobility/service.rs' unexpectedly found in {...}`). Removing the import restored exit 0 and 78/78. So the gate genuinely detects this class of violation and the new valve's exemption is real, not an artifact of the allowlist.

`cargo fmt` applied to touched files only. The crate is rustfmt-red at base (238 pre-existing files), so each finding was attributed before acting: the three in `service.rs` and one in `runtime/install.rs` were mine (caused by the `pub(super)` widening and the `self.mobility_service.` prefix pushing lines past 100 chars) and are fixed; the remaining `app/mod.rs` finding is a pre-existing import-ordering issue, proven inherited by running rustfmt against the base blob, and left alone.

`.github/workflows/ci.yml` untouched.


pablonyx:
**Adversarial review complete @ `bac5b1667` — CLEAN. No P1, no P2; ready for @pablonyx.**

The disclosed service.rs reconstruction was audited item-by-item against the original 1520-line file: **48/48 items accounted, 0 dropped, 0 duplicated**, and every one of the 16 non-identical bodies is forced-by-move (9 `pub(super)` widenings, 5 receiver-prefix rewrites, the ruled `runtime_home` field swap, the ctor shrink). Four independent cross-checks so no single method could hide a drop: line-multiset accounting (+57 lines, all 57 enumerated), semantic counters (tracing calls 5→5, error-return branches 33→33 exactly), import ledger (zero leaves vanished), orphan scan (none).

Also verified: PathBuf equivalence PROVEN (`app/mod.rs` passes the same `runtime_home` binding to both `SessionRuntime::new` and the mobility wiring — no shadowing, no canonicalization, cannot diverge); valve is one-way (service has zero references to `MobilityRuntime`; the sole live import in the module is `runtime/mod.rs:27`); allowlist stanza arithmetic re-derived exact (11 files/20); `app/mod.rs` 693-line pin holds exactly; base-sync merge carries zero Rust delta (nothing smuggled). Negative control: re-added live import → checker exit 1 naming service.rs + the repointed anchor test fails; restored → 90/90.

CI: Cargo check & test PASS (11m43s) at this head. The sole red (Login first-load budget) is a `packages.microsoft.com` 403 during playwright install — infra flake, zero frontend files in this diff; rerunning.

Nits (non-blocking, recorded for a later pass): the new `#[allow(clippy::too_many_arguments)]` on the 9-arg `MobilityRuntime::new` would be better as a wiring-deps struct like `MobilityWiringDeps` one layer up; `#[derive(Clone)]` on `MobilityService` is vestigial (pre-existing).

PR #1664, opened 2026-08-06
fix(product): keep file references actionable

pablonyx:
## Summary

- keep actionable inline file references semantic blue and remove muted-row color overrides
- recover workspace membership from missing nullable metadata with traversal-safe lexical normalization
- route external Desktop files through the configured open target and keep resolution or launch failures retryable
- render genuinely unavailable targets as plain text instead of disabled-looking links
- align completed file-change cards with the shared primary-action contract
- evict failed open-target discovery requests so a transient Desktop bridge failure can recover on retry

## Root cause

The file-reference badge inherited muted activity-row color, while the path metadata reducer collapsed omitted and null workspace paths into the same value. A later primary-action guard then treated inferred, external, and failed references as unavailable, producing disabled cursor and click behavior even for targets the Desktop host could open.

## User impact

Workspace references open in the viewer, external Desktop references use the configured target, and all rendered inline links remain keyboard- and pointer-actionable. Transient resolution, target-discovery, and launch failures can recover on a later click. Targets with no available action no longer masquerade as links.

## Validation

- `pnpm --filter @proliferate/product-client build`
- focused ProductClient Vitest suite: 102 tests passed across path resolution, target discovery, file references, read/edit ledgers, file changes, viewer, and transcript consumers
- `git diff --check`
- independent correctness, accessibility, failure-path, and integration reviews: clean

### DIFF
--- a/apps/packages/product-client/src/components/content/ui/FilePathLink.tsx
+++ b/apps/packages/product-client/src/components/content/ui/FilePathLink.tsx
@@ -15,8 +15,11 @@ interface FilePathLinkProps {
  * Inline file-path link rendered in chat markdown and tool-call output.
  *
  * Behavior:
- *  - Click -> open the file in the workspace right-sidebar viewer.
- *  - Context menu -> external open targets, copy path, reveal in Finder.
+ *  - Click -> open workspace files in the right-sidebar viewer and external
+ *    Desktop files in the configured external target.
+ *  - Actionable references expose external targets, copy, and reveal through
+ *    the context menu.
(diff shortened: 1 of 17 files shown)


pablonyx:
Resolved the target-discovery retry finding in `d0ae38fc0`: rejected target-list promises are now evicted, the hook keeps discovery unresolved after failure, and the next explicit open retries and restores target state on success. Added direct bridge failure→rediscovery coverage plus a caller-level failed launch→successful retry assertion. The final head passed the ProductClient build, focused regression suites, independent integration review, and all 23 CI checks.

PR #1665, opened 2026-08-06
refactor(anyharness): mobility resolve/decide/execute — decisions leave the effect path (grid PR 6b)

pablonyx:
## What

Grid PR 6b: restructure the two live mobility use cases into the doctrine's
**resolve → decide → execute** pipeline, and move every decision out of the
effect path into a new pure policy file. Closes backlog violation #1, part 2
(`anyharness-structure.md` line 294).

Before this PR both use cases interleaved fetch and act: `preflight_workspace`
pushed blockers into a `Vec` between store reads, git inspections and live
probes; `destroy_source_workspace` closed terminals and deleted sessions and
then handed a maybe-`None` default branch to the materialization call. After it,
each use case reads as one pass of facts, one call into policy, one ordered
sequence of effects.

### `preflight_workspace` (321 → 236 lines)

One resolve pass gathers everything the assessment branches on — workspace
record, runtime state, repo root, base commit, branch name, session candidates,
repo default branch, setup-running probe, git status, active terminals, active
review runs, per-session execution summary and pending prompts, subagent graph —
and hands it to `assess_mobility_preflight(&PreflightFacts)`, which returns the
whole blocker/warning matrix.

A failed git inspection is now modelled as a **fact**
(`PreflightGitStatus::Unavailable`) rather than a branch inside the pipeline;
likewise an unresolvable default branch (`DefaultBranchFact::Unresolved`). Both
previously pushed their blocker from inside the resolve step. Policy now decides
them.

The archive-size estimate is deliberately still last, after the assessment. It
is the one expensive fetch (it exports an entire archive), it is skipped
whenever anything else already blocks the move, and its own result feeds a
further decision (`archive_size_blocker`). That is the doctrine's
"effects that change facts force a re-resolve" shape, minus the effect. Moving
it into the resolve pass would export an archive on every blocked preflight.

### `destroy_source_workspace` (63 → 164 lines)

Resolve the workspace record, the repo default branch (`Local` only), every live
terminal, and every session row. Then `plan_source_destruction` returns a
`SourceDestructionPlan` — which terminals to close, which sessions to delete,
and how the materialization goes away — or rejects the call. Then the three
effects run in plan order.

The plan carries **ids only**. The `SessionRecord`s it names are resolved back
from the already-loaded rows *before the first deletion*, so a plan naming an
unknown session errors with nothing half-deleted rather than silently skipping
it. Capabilities (`terminal_service`, `workspace_runtime`, `session_runtime`)
stay on `&self` beside the plan, never inside it.

**There is no compensation slot, on purpose.** Force-closing a terminal,
deleting a session row and its agent artifacts, and removing a worktree are all
irreversible — a mid-sequence failure cannot be rolled back, so inventing a
compensation would be a lie. Instead the plan is ordered cheapest-to-recover
first (terminals, which the user can simply reopen) and each failure branch logs
the partial progress (`closed_terminal_count`, `deleted_session_count`) before
propagating. The caller holds an exclusive workspace lease across the whole call
(`WORKSPACE-FENCE-01/02`), so no concurrent writer can invalidate the resolved
facts between decide and execute.

### `runtime/mod.rs`

The local `fn is_active_terminal` is gone; both gather helpers now filter on
`mobility_policy::terminal_is_active`, so the terminal-activity rule has exactly
one home and `destroy_source` decides over the same predicate.

## Policy-home decision

The brief asked for a deliberate choice, since PR 6a left pure free fns with
tests in `mobility/service.rs`. **`mobility_policy.rs` is the single decision
home for the live mobility use cases.** It does not become a second home for the
domain:

- **Delegates, does not absorb:** `classify_session_support` calls
  `service::is_supported_agent_kind` rather than restating the predicate. That
  predicate also gates which sessions enter the durable export bundle, so
  preflight and export must get exactly one answer. The policy owns only the
  *surfaced reason* text, which is preflight's concern alone.
- **Leaves the durable rules where they are:** `service.rs` keeps
  `validate_expected_handoff_runtime_state`,
  `classify_existing_archive_session_for_relocation`, `validate_archive_size`
  and `archive_estimated_size_bytes`. Those belong to the durable export/import
  use cases. Pulling them into a file under `runtime/` would invert the one-way
  runtime → service direction PR 6a established (the runtime wraps the service;
  the service must not depend on a runtime file).

So the split is by **use case owner**, not by purity: pure decisions belonging
to a live use case live in `runtime/mobility_policy.rs`; pure decisions
belonging to a durable use case stay beside that use case in `service.rs`. The
rule is recorded in the `runtime/mod.rs` module doc so the next reader does not
have to re-derive it.

`service.rs` is **untouched** — still exactly 851 lines, matching its exact pin.

## Failure semantics: before vs after

Two orderings changed, both moving failure ahead of effects (rows 1 and 3);
no effect moved ahead of any check. Stated per effect, per step:

| Step | Before | After |
| --- | --- | --- |
| **destroy-source: `Local` workspace with no usable default branch** | Terminals force-closed → every session and its agent artifacts deleted → *then* `destroy_source_workspace_materialization` raises `default branch is required to park a local workspace` → HTTP 500. User is left with an unparked workspace, closed terminals and **permanently deleted sessions**, and a retry cannot restore them. | `plan_source_destruction` rejects during decide. **Same message, same `MobilityError::Internal`, same HTTP 500** — but zero terminals closed and zero sessions deleted. The workspace is exactly as it was and a retry is meaningful. |
| **destroy-source: terminal close fails** | `?` propagates `MobilityError::Internal`; earlier terminals stay closed; no session deleted; materialization untouched. | Identical outcome and identical error. Added: a `tracing::error!` recording how many terminals had closed, and the explicit note that no session was deleted. |
| **destroy-source: session delete / artifact cleanup fails** | `?` propagates; all terminals already closed; earlier sessions already deleted; materialization untouched. | Identical outcome and identical error. Added: structured logging of `closed_terminal_count` / `deleted_session_count`. Also defensive scaffolding: if a plan ever named a session not in the loaded set, that errors *before* the first delete — currently unreachable (the plan's ids are built from the same loaded vec), it exists to fail closed if the plan and the load ever diverge. |
| **destroy-source: session-list resolution fails** | `list_by_workspace` ran *after* the terminal closes, so a session-list failure left terminals already closed. | The list is resolved up front (resolve-before-decide), so the same failure now aborts before any terminal closes — a second small ordering improvement in the same safe direction as the first row. |
| **destroy-source: materialization destroy fails** | `?` propagates; terminals closed and sessions deleted. | Identical outcome and identical error, plus the same structured log. |
| **destroy-source: default-branch *resolution* fails (`Local`)** | `resolve_repo_root_default_branch` error propagates before any effect. | Unchanged — still propagates before any effect, still `MobilityError::Internal`. |
| **preflight: any blocker** | Blockers accumulated inline, in pipeline order. | Same codes, same messages, same `session_id`s, **same order**. The order is user-visible (serialized straight through the API), so it is reproduced exactly and now pinned by `blocker_order_is_workspace_then_review_then_session_then_graph`. |
| **preflight: fallible resolve steps (`load_workspace`, `runtime_state`, `resolve_repo_root`, `rev-parse HEAD`, `current_branch_name`, `list_sessions`, `list_active_runs_for_workspace`, `list_pending_prompts`, `mobility_graph_for_sessions`, `export_workspace_archive`)** | Each `?`/`map_err` aborts preflight with its own error. | Identical error surfaces, and identical relative order — every one of these still runs in the same sequence, so a workspace that failed at a given step before fails at the same step with the same error now. Preflight has no effects, so there is nothing to compensate either way. |
| **preflight: archive too large** | Estimate runs only when `blockers.is_empty()`; the size blocker is appended after. | Unchanged, via `workspace_can_move` + `archive_size_blocker`. Still one export, still skipped when anything else blocks. |

The behaviour changes are the first two rows, both in the same direction:
an irrecoverable partial teardown becomes a clean rejection (identical error
message), and a resolution failure aborts before effects instead of between
them. Nothing else in either use case changes what a caller observes.

## Test inventory

`runtime/mobility_policy_tests.rs` — 26 tests, hand-built facts, no DB, no
mocks, in the `launch_policy.rs` style (builders `clean_preflight_facts()`,
`idle_session()`, `codes()`).

| Policy fn | Decision it makes | Tests |
| --- | --- | --- |
| `terminal_status_is_active` / `terminal_is_active` | which terminals still own a process | 1 |
| `classify_session_support` | can mobility v1 carry this agent kind, and why not | 2 |
| `movable_session_ids` | which sessions enter the archive | 1 |
| `workspace_can_move` | is the move allowed at all | 1 |
| `archive_size_blocker` | is the measured archive over the limit | 1 |
| `assess_mobility_preflight` | the whole blocker/warning matrix and its order | 15 |
| `plan_source_destruction` | the destroy-source effect sequence, and its precondition | 4 |
| `MaterializationDestruction::default_branch` | the argument the materialization call receives | covered by the 4 above |

The 15 `assess_mobility_preflight` tests cover: the clean case, non-normal
runtime mode, unresolved local default branch, local-on-default-branch, local on
a side branch, a worktree whose branch merely shares the default name, running
setup, all four dirty-git conditions in order, unavailable git status, terminals
warning without blocking, review runs naming their parent session, running
sessions, per-session blocker stacking order, the generic-reason fallback, and a
partial subagent graph. Plus the cross-cutting order test.

## Gates

All run on the final tree, in this worktree.

| Gate | Result |
| --- | --- |
| `check_anyharness_boundaries.py` | `AnyHarness boundary check passed.` exit 0 |
| `python3 -m unittest scripts.test_check_anyharness_boundaries` | `Ran 90 tests` — `OK` |
| `check_max_lines.py` | `Max-lines check passed (repo max 600, component max 500, server layer maxes enabled).` |
| `cargo check -p anyharness-lib` | `Finished dev profile` — zero warnings naming any mobility file |
| `cargo test -p anyharness-lib mobility` | `test result: ok. 41 passed; 0 failed` |

**No allowlist row, max-lines pin, or boundary anchor moved.** `service.rs` is
untouched at 851 lines so its exact pin still matches; the two new files are 416
and 527 lines (both under 600); the policy file is genuinely pure so it needs no
`POLICY_PURITY` row; and the existing anchors pointing at
`domains/mobility/service.rs` and `domains/mobility/runtime/mod.rs` still
reference real lines on the correct side of the valve rule. The anchor suite
stays at exactly 90 tests. `.github/workflows/ci.yml` is untouched, and no
`Cargo.lock` change is committed.

### Negative controls

Both were required and both fired.

**1. Invert a policy decision.** Replaced `if default_branch.is_empty()` with
`if false` in `plan_source_destruction`, so a `Local` workspace with no default
branch produces a plan instead of a rejection:

```
thread '...local_destruction_is_rejected_before_any_effect_without_a_default_branch'
panicked at .../mobility_policy_tests.rs:505:45:
a local park needs a default branch: SourceDestructionPlan {
  close_terminal_ids: ["terminal-1"], delete_session_ids: ["session-1"],
  materialization: ParkLocalOnDefaultBranch { default_branch: "" } }

test result: FAILED. 25 passed; 1 failed; 0 ignored; 0 measured; 1488 filtered out
```

Exactly the one predicted test, with the predicted assertion. The payload also
shows precisely what the old code did: it would have closed `terminal-1`,
deleted `session-1`, and only then tried to park onto `""`. Restored →
`test result: ok. 41 passed; 0 failed`.

**2. Confirm `POLICY_PURITY` polices this file.** Added `let _ =
chrono::Utc::now();` to `workspace_can_move`:

```
POLICY_PURITY .../domains/mobility/runtime/mobility_policy.rs: 1 unallowlisted

anyharness/crates/anyharness-lib/src/domains/mobility/runtime/mobility_policy.rs:63: [POLICY_PURITY] policy files are pure: facts in, decisions out — no clock, no ids, no store; mint in execute
BOUNDARIES_EXIT=1
```

Exit 1, naming the file and the exact line. Reverted →
`AnyHarness boundary check passed. BOUNDARIES_EXIT=0`.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/domains/mobility/runtime/destroy_source.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/mobility/runtime/destroy_source.rs
@@ -2,9 +2,30 @@
 //!
 //! Live because it force-closes live terminals and because destroying the
 //! materialization goes through `WorkspaceRuntime`.
+//!
+//! Pipeline: resolve every fact (workspace record, repo default branch, live
+//! terminals, session rows) → decide the whole effect sequence in
+//! `mobility_policy::plan_source_destruction` → execute the plan in order. The
+//! resolve pass replaces the old fetch/act interleaving, so the `Local`
+//! default-branch precondition is now proved before the first irreversible
+//! effect instead of being discovered by the materialization call after the
+//! terminals are closed and the sessions deleted.
+//!
+//! There is no compensation slot here on purpose: force-closing a terminal,
+//! deleting a session row, and removing a worktree are all irreversible, so a
+//! mid-sequence failure cannot be rolled back. The plan is therefore ordered
+//! cheapest-to-recover first (terminals, which the user can reopen) and the
+//! partial progress is logged before the error propagates. The caller holds the
+//! exclusive workspace lease for the whole call, so no concurrent writer can
+//! invalidate the resolved facts between decide and execute.
 
+use std::collections::HashMap;
 use std::path::PathBuf;
 
+use super::mobility_policy::{
+    plan_source_destruction, DefaultBranchFact, SourceDestructionFacts, SourceDestructionRejection,
+    TerminalFact,
+};
 use super::MobilityRuntime;
 use crate::domains::agents::portability::delete_session_agent_artifacts;
 use crate::domains::mobility::model::DestroyedWorkspaceSourceSummary;
@@ -16,42 +37,122 @@ impl MobilityRuntime {
         &self,
         workspace_id: &str,
     ) -> Result<DestroyedWorkspaceSourceSummary, MobilityError> {
+        // --- resolve -------------------------------------------------------
         let workspace = self.mobility_service.load_workspace(workspace_id)?;
         let workspace_path = PathBuf::from(&workspace.path);
         let default_branch = if workspace.kind == WorkspaceKind::Local {
             let repo_root_id = workspace.repo_root_id.clone();
-            Some(
+            DefaultBranchFact::Resolved(
                 self.workspace_runtime
                     .resolve_repo_root_default_branch(&repo_root_id)
                     .map_err(MobilityError::Internal)?,
             )
         } else {
-            None
+            DefaultBranchFact::NotRequired
         };
-
-        let active_terminals = self.active_terminals_blocking(workspace_id);
-        let mut closed_terminal_ids = Vec::new();
-        for terminal in active_terminals {
-            self.terminal_service
-                .close_terminal_blocking(&terminal.id)
-                .map_err(MobilityError::Internal)?;
-            closed_terminal_ids.push(terminal.id);
-        }
+        let terminals = self
+            .terminal_service
+            .list_terminals_blocking(workspace_id)
+            .into_iter()
+            .map(|terminal| TerminalFact {
+                terminal_id: terminal.id,
+                status: terminal.status,
+            })
+            .collect::<Vec<_>>();
         let sessions = self
             .session_service
             .store()
             .list_by_workspace(workspace_id)?;
-        let mut deleted_session_ids = Vec::new();
-        let runtime_home = Some(self.session_runtime.runtime_home());
-        for session in sessions {
-            delete_session_agent_artifacts(&session, &workspace_path, runtime_home)?;
-            self.session_service.delete_session(&session.id)?;
-            deleted_session_ids.push(session.id);
+
+        // --- decide --------------------------------------------------------
+        let plan = plan_source_destruction(&SourceDestructionFacts {
+            workspace_kind: workspace.kind,
+            default_branch,
+            terminals,
+            session_ids: sessions.iter().map(|session| session.id.clone()).collect(),
+        })
+        .map_err(|rejection| match rejection {
+            // Same message and surface the materialization call used to raise
+            // once the effects had already run.
+            SourceDestructionRejection::MissingLocalDefaultBranch => MobilityError::Internal(
+                anyhow::anyhow!("default branch is required to park a local workspace"),
+            ),
+        })?;
+
+        // --- execute -------------------------------------------------------
+        let mut closed_terminal_ids = Vec::new();
+        for terminal_id in &plan.close_terminal_ids {
+            if let Err(error) = self.terminal_service.close_terminal_blocking(terminal_id) {
+                tracing::error!(
+                    workspace_id = %workspace_id,
+                    terminal_id = %terminal_id,
+                    closed_terminal_count = closed_terminal_ids.len(),
+                    error = %error,
+                    "mobility destroy-source aborted while closing terminals; no session was deleted"
+                );
+                return Err(MobilityError::Internal(error));
+            }
+            closed_terminal_ids.push(terminal_id.clone());
         }
 
-        self.workspace_runtime
-            .destroy_source_workspace_materialization(&workspace, default_branch.as_deref())
-            .map_err(MobilityError::Internal)?;
+        // The plan names ids; the records travel beside it. Resolve the plan's
+        // ids back to their rows BEFORE the first deletion so a mismatch fails
+        // with nothing half-deleted.
+        let sessions_by_id = sessions
+            .iter()
+            .map(|session| (session.id.as_str(), session))
+            .collect::<HashMap<_, _>>();
+        let planned_sessions = plan
+            .delete_session_ids
+            .iter()
+            .map(|session_id| {
+                sessions_by_id
+                    .get(session_id.as_str())
+                    .copied()
+                    .ok_or_else(|| {
+                        MobilityError::Internal(anyhow::anyhow!(
+                            "destroy-source plan named unknown session {session_id}"
+                        ))
+                    })
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let mut deleted_session_ids = Vec::new();
+        let runtime_home = Some(self.session_runtime.runtime_home());
+        for session in planned_sessions {
+            if let Err(error) =
+                delete_session_agent_artifacts(session, &workspace_path, runtime_home)
+                    .and_then(|()| self.session_service.delete_session(&session.id))
+            {
+                tracing::error!(
+                    workspace_id = %workspace_id,
+                    session_id = %session.id,
+                    closed_terminal_count = closed_terminal_ids.len(),
+                    deleted_session_count = deleted_session_ids.len(),
+                    error = %error,
+                    "mobility destroy-source aborted while deleting sessions; the materialization was left in place"
+                );
+                return Err(MobilityError::Internal(error));
+            }
+            deleted_session_ids.push(session.id.clone());
+        }
+
+        if let Err(error) = self
+            .workspace_runtime
+            .destroy_source_workspace_materialization(
+                &workspace,
+                plan.materialization.default_branch(),
+            )
+        {
+            tracing::error!(
+                workspace_id = %workspace_id,
+                closed_terminal_count = closed_terminal_ids.len(),
+                deleted_session_count = deleted_session_ids.len(),
+                error = %error,
+                "mobility destroy-source aborted while destroying the materialization"
+            );
+            return Err(MobilityError::Internal(error));
+        }
 
         Ok(DestroyedWorkspaceSourceSummary {
             workspace_id: workspace.id,
--- a/anyharness/crates/anyharness-lib/src/domains/mobility/runtime/mobility_policy.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/mobility/runtime/mobility_policy.rs
@@ -1,0 +1,416 @@
+//! Pure mobility decisions for the mobility runtime.
+//!
+//! Everything here is data-in / data-out: no store, no clock, no uuid, no
+//! `&self`. `preflight.rs` and `destroy_source.rs` perform the resolve steps
+//! (workspace/session/terminal loads, git inspection, live probes) and hand the
+//! gathered facts in; this module owns the decisions — which terminals count as
+//! active, which sessions mobility v1 can carry, what blocks a move, and what
+//! the destroy-source effect sequence is.
+//!
+//! Two shapes appear per use case: a `*Facts` struct the IO layer fills in, and
+//! a decision (a plan, an assessment, or a predicate) the IO layer executes.
+//! Plans carry ids only — the records and the capabilities they name travel
+//! beside the plan, never inside it.
+
+use crate::domains::mobility::model::{MobilityBlocker, MAX_MOBILITY_ARCHIVE_BODY_BYTES};
+use crate::domains::mobility::service::is_supported_agent_kind;
+use crate::domains::terminals::model::{TerminalRecord, TerminalStatus};
+use crate::domains::workspaces::access_model::WorkspaceAccessMode;
+use crate::domains::workspaces::model::WorkspaceKind;
+
+/// A terminal is "active" for mobility purposes while it still owns a process:
+/// starting or running. Exited and failed terminals need no force-close and
+/// raise no warning. Single home for the rule — `runtime/mod.rs` filters live
+/// records with [`terminal_is_active`], `destroy_source.rs` decides over
+/// [`TerminalFact`]s.
+pub(super) fn terminal_status_is_active(status: &TerminalStatus) -> bool {
+    matches!(status, TerminalStatus::Starting | TerminalStatus::Running)
+}
+
+/// [`terminal_status_is_active`] over a whole live record, for the runtime's
+/// gather helpers.
+pub(super) fn terminal_is_active(terminal: &TerminalRecord) -> bool {
+    terminal_status_is_active(&terminal.status)
+}
+
+/// Whether mobility v1 can carry a session of this agent kind, and the reason
+/// text shown when it cannot. The support predicate itself stays in
+/// `service.rs` — it also gates which sessions enter the durable export bundle,
+/// so there is exactly one answer for both paths; this fn owns the surfaced
+/// reason.
+pub(super) fn classify_session_support(agent_kind: &str) -> SessionSupport {
+    if is_supported_agent_kind(agent_kind) {
+        return SessionSupport {
+            supported: true,
+            reason: None,
+        };
+    }
+    SessionSupport {
+        supported: false,
+        reason: Some("Unsupported agent kind for workspace mobility v1".to_string()),
+    }
+}
+
+pub(super) struct SessionSupport {
+    pub supported: bool,
+    pub reason: Option<String>,
+}
+
+/// Whether the workspace can move: no blockers, nothing else. One rule with two
+/// consumers — it gates the (expensive) archive size estimate and it is the
+/// `can_move` field of the preflight result.
+pub(super) fn workspace_can_move(blockers: &[MobilityBlocker]) -> bool {
+    blockers.is_empty()
+}
+
+/// The archive-size limit, decided over an already-measured estimate.
+pub(super) fn archive_size_blocker(estimated_bytes: u64) -> Option<MobilityBlocker> {
+    if estimated_bytes <= MAX_MOBILITY_ARCHIVE_BODY_BYTES as u64 {
+        return None;
+    }
+    Some(MobilityBlocker {
+        code: "archive_too_large".to_string(),
+        message: format!(
+            "Archive exceeds the {} byte limit",
+            MAX_MOBILITY_ARCHIVE_BODY_BYTES
+        ),
+        session_id: None,
+    })
+}
+
+// ---------------------------------------------------------------------------
+// Preflight: can this workspace move right now?
+// ---------------------------------------------------------------------------
+
+/// Whether the repo's default branch was needed and resolved. Only `Local`
+/// workspaces need one (they park onto it); worktrees are removed outright.
+pub(super) enum DefaultBranchFact {
+    /// Not a `Local` workspace — no default branch is involved.
+    NotRequired,
+    Resolved(String),
+    /// Required but the resolver failed.
+    Unresolved,
+}
+
+/// Git inspection result. A failed inspection is a fact, not an error: preflight
+/// reports it as a blocker instead of aborting.
+pub(super) enum PreflightGitStatus {
+    Inspected {
+        detached: bool,
+        operation_in_progress: bool,
+        conflicted: bool,
+        clean: bool,
+    },
+    Unavailable {
+        error: String,
+    },
+}
+
+pub(super) struct PreflightReviewRun {
+    pub run_id: String,
+    pub parent_session_id: String,
+}
+
+/// One session's resolved facts. `supported`/`reason` come from
+/// [`classify_session_support`]; the rest are store and live reads.
+pub(super) struct PreflightSessionFacts {
+    pub session_id: String,
+    pub status: String,
+    pub agent_kind: String,
+    pub supported: bool,
+    pub unsupported_reason: Option<String>,
+    pub awaiting_interaction: bool,
+    pub has_pending_prompts: bool,
+}
+
+/// Every fact the preflight assessment branches on, gathered by `preflight.rs`
+/// in a single resolve pass before any decision is made.
+pub(super) struct PreflightFacts {
+    pub workspace_kind: WorkspaceKind,
+    pub runtime_mode: WorkspaceAccessMode,
+    pub branch_name: Option<String>,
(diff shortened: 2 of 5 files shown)


pablonyx:
## Adversarial review — PR #1665 @ dc70e43c4

**Verdict: CLEAN** (2 nits, no P1s, no P2s)

I tried to break the behavior-equivalence claim step by step and could not. The disclosed
behavior change is real, correctly characterized, and the only one. Both of my own
independent negative controls fired on exactly the predicted test.

---

### 1. destroy-source: effect-order audit, old vs new

Every fallible step, in execution order. `E` = effect (mutates state), `R` = resolve (read).

| # | OLD (`763e43067`) | # | NEW (`dc70e43c4`) | Same? |
| 1 | R `load_workspace` → `WorkspaceNotFound` | 1 | R `load_workspace` | identical |
| 2 | R `resolve_repo_root_default_branch` (Local only) → `Internal` | 2 | R same, Local only, same `map_err` | identical |
| 3 | R `active_terminals_blocking` (infallible; filtered to active) | 3 | R `list_terminals_blocking` (infallible; **unfiltered**, filter moved into `plan_source_destruction`) | same resulting id set, see below |
| — | — | 4 | R `store().list_by_workspace` → `Internal` — **moved earlier** (was old step 5) | see F1 |
| — | — | 5 | **decide** `plan_source_destruction` → rejects missing-Local-default-branch | the disclosed change |
| 4 | **E** `close_terminal_blocking` per active terminal → `Internal` | 6 | **E** same, same order | identical |
| 5 | R `store().list_by_workspace` → `Internal` | — | (hoisted to step 4) | see F1 |
| — | — | 7 | R id→row remap, fail-closed on unknown id (pure, in-memory) | new, unreachable — see F2 |
| 6 | **E** `delete_session_agent_artifacts` then `delete_session`, per session | 8 | **E** same pair, same per-session order, same store order | identical |
| 7 | **E** `destroy_source_workspace_materialization` → `Internal` | 9 | **E** same, `default_branch()` yields the identical `Option<&str>` | identical |

**Verified: no effect moved ahead of a check.** The three effects (steps 6/8/9 new) are in
exactly the old relative order, and every resolve now sits strictly before all of them —
the safe direction. `close_terminal_blocking` cannot create or delete a session row
(`handle.close()` = registry remove + `kill()` + output-hub remove), so hoisting the
session list above it reads the same rows. Confirmed the terminal filter is semantically
identical: `mobility_policy::terminal_status_is_active` is a byte-for-byte move of the
deleted `runtime/mod.rs::is_active_terminal` (`Starting | Running`), and
`list_terminals_blocking` already sorts by `created_at`, so `close_terminal_ids` preserves
the old iteration order.

**Error types/messages/HTTP unchanged.** Every failure path is `MobilityError::Internal` →
`ApiError::internal(error.to_string())` (`api/http/mobility.rs:474`), i.e. HTTP 500. The
rejection message `"default branch is required to park a local workspace"` is a literal
match for the one `materialization.rs:66` raises, so the wire body is byte-identical to the
old late failure. `map_mobility_error` and `repo_roots.rs`'s parallel mapping are untouched.

#### F1 (nit) — hoisting `list_by_workspace` above the terminal closes is a real but
benign reordering, and the PR body doesn't call it out. The body's failure table says the
terminal-close row is "identical outcome and identical error," which is true, but it omits
that a *session-list* failure now aborts *before* any terminal is closed instead of after.
That's strictly better (fewer effects on failure, and a retry is meaningful) and is required
by resolve-before-decide, but it is a second — small — ordering improvement beyond the one
disclosed. Worth one line in the body for the next reader; not worth a code change.

#### F2 (nit) — the fail-closed unknown-session check is currently unreachable.
`plan.delete_session_ids` is built as `sessions.iter().map(|s| s.id.clone())` from the very
`sessions` vec the remap looks ids up in, so `sessions_by_id.get()` cannot miss. It is a
correct guard against a *future* policy that filters or reorders sessions, and it costs one
in-memory pass with no new failure mode today. Fine as defensive scaffolding; just note the
PR body describes it as "newly fail-closed," which overstates it slightly (it fails closed
against a plan shape the current policy cannot emit).

---

### 2. The disclosed behavior change: correctly characterized, and reachable

`Local` + no usable default branch. Old: terminals force-closed → **every session and its
agent artifacts permanently deleted** → *then* `materialization.rs:66` raises → 500, and a
retry cannot restore the sessions. New: `plan_source_destruction` returns
`MissingLocalDefaultBranch` at decide, zero effects run, same message, same
`MobilityError::Internal`, same 500.

Two things I checked that the body doesn't spell out and that make the claim *stronger*,
not weaker:

- The `Resolved("")` / `Resolved("   ")` case is genuinely reachable, not theoretical:
  `resolve_repo_root_default_branch` can return `Ok(branch)` and the policy then `trim()`s
  it, mirroring `materialization.rs:63-67`'s `.map(str::trim).filter(|v| !v.is_empty())`
  exactly. The policy's `.trim()` + `is_empty()` reproduces that predicate precisely — I
  checked for an off-by-one (e.g. trimming in the plan but passing the untrimmed value on),
  and it does not happen: `ParkLocalOnDefaultBranch` stores the *trimmed* string, so
  `park_local_workspace` receives the same `&str` it did before.
- `DefaultBranchFact::NotRequired` under `WorkspaceKind::Local` also rejects. That state is
  unconstructible from `destroy_source.rs` (the `if kind == Local` arm always produces
  `Resolved`), but rejecting rather than defaulting is the fail-closed choice and the test
  pins all three shapes.

`WorkspaceKind` has exactly two variants (`Local`, `Worktree`), so the `match` in
`plan_source_destruction` is exhaustive with no silent fallthrough.

---

### 3. preflight: no ordering change, and no new failure modes

Fallible-step sequence, old line numbers → new line numbers:

`list_sessions` (58→76) → `resolve_repo_root_default_branch` (89→92) →
`is_setup_running` (106→101) → `GitService::status` (114→103) →
`active_terminals_async` (168→116) → `list_active_runs_for_workspace` (177→124) →
per-session `session_execution_summary` + `list_pending_prompts` (197/210→136/141) →
`mobility_graph_for_sessions` (243→159) → `export_workspace_archive` (270→194).

Identical relative order, so a workspace that failed at a given step before fails at the
same step with the same error now. **Nothing that used to short-circuit now always runs** —
I looked specifically for this, because it's the classic resolve-batching regression. It
doesn't happen here for a structural reason: the old code accumulated blockers into a `Vec`
and never early-returned, so *every* one of those reads already ran unconditionally on every
call. The only conditional fetch was the archive export, and it is still gated
(`workspace_can_move(&blockers)` before the export, `archive_size_blocker(size)` after) —
so a blocked preflight still exports zero archives. No new git commands, no new latency, no
new side effects.

Two resolve steps changed from "abort/branch inline" to "fact + policy decides": failed git
inspection → `PreflightGitStatus::Unavailable`, unresolvable default branch →
`DefaultBranchFact::Unresolved`. Both previously pushed their blocker from inside the
resolve step and neither aborted, so this is pure relocation. The blocker order is
reproduced exactly, including the subtle one: `default_branch_unknown` is emitted *second*
(right after `workspace_not_mutable`) even though the resolve now happens after
`list_sessions` — the policy's emission order is what's user-visible, and it matches.

`movable_session_ids` reproduces the old `HashSet` filter on `supported` exactly, so
`mobility_graph_for_sessions` gets the same input set.

---

### 4. Resolve-before-decide staleness (TOCTOU)

**No material regression.** I checked whether the old code had the fresher read at each act
and it did not, in either use case:

- **destroy-source:** the handler holds `acquire_exclusive(&workspace_id)` across the whole
  call plus per-session admission permits, and re-enumerates + fails closed under the lease
  (`WORKSPACE-FENCE-01/02`, `mobility.rs:270-297`) *before* calling into the runtime. The
  exclusive lease excludes `SessionStart`, which is precisely the window a late session
  could bind in. So the resolved facts cannot be invalidated between decide and execute by a
  legitimate writer — the fence is what makes the batching safe, and the module doc cites it
  correctly.
- **terminal spawning between resolve and close:** possible in both old and new, and equally
  so — the old code also snapshotted `active_terminals_blocking` once into a `Vec` and then
  looped. Not a regression.
- **preflight:** read-only assessment; a stale fact just means a stale advisory answer,
  which was already true (the old code's first read was equally stale by the time the last
  blocker was pushed).

The fail-closed unknown-session remap (F2) is more than sufficient for the batching it
guards — it is in fact currently unreachable.

---

### 5. Policy purity beyond the lint

Read all 416 lines. Clean:

- No `&self` anywhere. The one `impl` block is `MaterializationDestruction::default_branch(&self)`
  — a method on a plain data enum the policy itself defines, not a capability holder. Matches
  the exemplar (`launch_policy.rs` is all free fns; a pure accessor on an owned plan type is
  the same shape).
- No `Arc` fields, no callbacks, no trait objects, no store/clock/uuid/rand. Imports are
  `model`, `service::is_supported_agent_kind`, `terminals::model`, `workspaces::access_model`,
  `workspaces::model` — all pure vocabulary.
- Plans carry ids only (`close_terminal_ids: Vec<String>`, `delete_session_ids: Vec<String>`);
  capabilities (`terminal_service`, `workspace_runtime`, `session_runtime`) stay on `&self`
  in `destroy_source.rs`. Doctrine's "capabilities travel beside plans, never inside" holds.
- Contexts are private: everything is `pub(super)`, nothing escapes `runtime/`. I grepped —
  no policy type appears outside `runtime/mobility_policy*.rs`, `destroy_source.rs`,
  `preflight.rs`.
- **No decision left behind in the effect path.** The remaining `if`s in `destroy_source.rs`
  are all `if let Err(...) = <effect>` retry/abort control flow plus the id-remap lookup —
  no branching on domain facts. `preflight.rs`'s remaining `if` is
  `if workspace.kind == WorkspaceKind::Local` gating a *fetch* (whether to resolve the
  default branch at all), which is resolve-scope, not a decision; the decision it feeds
  (`DefaultBranchFact`) is policy's.

I confirmed the `POLICY_PURITY` lint actually covers this path: `is_policy_file()` matches
any `*_policy.rs` under `domains/**`, and `POLICY_IMPURITY_RES` catches
`Utc::now`/`Local::now`/`SystemTime::now`/`Instant::now`/`Uuid::new_v4`/`rand::` plus
store/adapter imports. So the file is inside the fence, not merely adjacent to it — the
builder's second negative control is load-bearing and I agree with it.

---

### 6. Policy-home coherence — no rule has two homes

I specifically hunted for a duplicated rule that could give two different answers. There
isn't one:

- `is_supported_agent_kind` (`service.rs:436`, `matches!(agent_kind, "claude" | "codex")`)
  has exactly one definition and two callers: `service.rs:160` (durable export bundle) and
  `mobility_policy.rs:42` (`classify_session_support`). Preflight and export therefore
  cannot disagree about which sessions can move — which is the property that matters, since
  a preflight that green-lit a session the export then dropped would be the real bug. The
  policy adds only the surfaced reason string, which is preflight's alone.
- Terminal activity: `mobility_policy::terminal_status_is_active` is now the single home.
  The old duplicate (`runtime/mod.rs::is_active_terminal`) is deleted, and both gather
  helpers plus `plan_source_destruction` route through it. This is a genuine
  de-duplication — before this PR the destroy-source filter and the preflight filter were
  the same fn but the decision wasn't testable.
- Durable rules left in `service.rs` (`validate_expected_handoff_runtime_state`,
  `classify_existing_archive_session_for_relocation`, `validate_archive_size`,
  `archive_estimated_size_bytes`) have no counterpart in the policy file. I checked the one
  near-collision: `validate_archive_size` (durable, → `SizeLimitExceeded`, also enforces
  per-file limits) vs `archive_size_blocker` (preflight, → a `MobilityBlocker`, total only).
  Different surfaces, different outputs, both reading the same
  `MAX_MOBILITY_ARCHIVE_BODY_BYTES` constant — and the preflight one is a byte-for-byte move
  of the old inline check, same `>` comparison, same message. Not a duplication; two
  consumers of one constant.

The by-use-case-owner split (live rules in `runtime/mobility_policy.rs`, durable rules
beside their use case in `service.rs`) is the right call given the one-way runtime → service
direction, and recording it in the `runtime/mod.rs` module doc is the right place.

---

### 7. Test quality — and my own negative controls

26 tests, all hand-built facts, no DB, no live handles, no clock, `launch_policy.rs` style
(`clean_preflight_facts()` / `idle_session()` / `codes()`). **Not tautological**: expectations
are independently stated literals — full blocker-code vectors and exact message strings
written out by hand (e.g. `"Workspace is currently in frozen_for_handoff mode"`,
`"Main local workspaces on 'main' must move from a worktree instead"`), not recomputed from
the code under test.

Blocker ORDER is pinned as claimed, in three places:
`blocker_order_is_workspace_then_review_then_session_then_graph` (the full cross-cutting
matrix, 7 codes), `every_dirty_git_condition_raises_its_own_blocker_in_order` (the 4 git
codes), and `per_session_blockers_stack_in_a_fixed_order` (the 4 per-session codes).

```
cargo test -p anyharness-lib -j 6 --lib mobility_policy
test result: ok. 26 passed; 0 failed; 0 ignored; 1488 filtered out
```

#### My independent negative controls — two, both different from the builder's

The builder inverted the destroy-source default-branch check. I picked two **different**
decisions, one per use case, to probe coverage the builder's control didn't touch.

**Control A — delete the `Local`-only guard on the default-branch-in-use rule**
(`mobility_policy.rs:230`), i.e. make the rule fire for worktrees too:

```diff
-    if facts.workspace_kind == WorkspaceKind::Local {
+    {
         if let (Some(current_branch), DefaultBranchFact::Resolved(default_branch)) =
```

```
test ...a_worktree_on_the_repo_default_branch_name_is_not_blocked ... FAILED
panicked at mobility_policy_tests.rs:202:5:
  left: ["local_default_branch_in_use"]
 right: []
test result: FAILED. 25 passed; 1 failed; 0 ignored; 1488 filtered out
```

Exactly one test, and the *right* one — the test whose comment says "The default-branch-in-use
rule is Local-only: a worktree that happens to sit on a branch named like the default still
moves." Restored → 26 passed.

**Control B — invert the terminal-activity filter inside the destroy-source plan**
(`mobility_policy.rs:410`), i.e. close the dead terminals and leave the live ones running:

```diff
-            .filter(|terminal| terminal_status_is_active(&terminal.status))
+            .filter(|terminal| !terminal_status_is_active(&terminal.status))
```

```
test ...worktree_destruction_closes_active_terminals_and_deletes_every_session ... FAILED
panicked at mobility_policy_tests.rs:449:5:
  left: ["terminal-exited"]
 right: ["terminal-running", "terminal-starting"]
test result: FAILED. 25 passed; 1 failed; 0 ignored; 1488 filtered out
```

Exactly one test, the right one, and the payload proves the fixture discriminates
(`terminal-exited` is present in the facts precisely so an inverted filter is visible).
Restored → 26 passed, `git status` clean at dc70e43c4.

Both controls land on assertions the builder's control could not have reached, so the suite
is discriminating across both use cases, not just the destroy-source precondition.

---

### 8. Mechanical

| Check | Result |
| --- | --- |
| `check_anyharness_boundaries.py` | `AnyHarness boundary check passed.` exit 0 |
| `python3 -m unittest scripts.test_check_anyharness_boundaries` | `Ran 90 tests` — `OK` (anchor count unchanged) |
| `check_max_lines.py` | passed (repo max 600, component max 500) |
| Line counts | `service.rs` 851 (exact pin intact, byte-identical), `mobility_policy.rs` 416, `mobility_policy_tests.rs` 527, `destroy_source.rs` 63→164, `preflight.rs` 321→236, `mod.rs` 110 — all match the claims, all under 600 |
| Diffstat | 5 files, +1169/−207 — matches exactly |
| `cargo fmt -- --check` on mobility files | zero diffs (236 pre-existing diffs elsewhere in the crate from a newer local rustfmt, none in the touched files) |
| Unused imports / warnings | zero warnings naming any mobility file in the test build; `TerminalStatus` correctly dropped from `mod.rs`, `TerminalRecord` correctly retained (still used by the two gather helpers); `MobilityBlocker`/`MAX_MOBILITY_ARCHIVE_BODY_BYTES`/`is_supported_agent_kind` correctly dropped from `preflight.rs` |
| pub leakage | none — every policy item is `pub(super)`, nothing referenced outside `runtime/` |
| Allowlists / pins / anchors | `scripts/` untouched; no `mobility` row added to either allowlist |
| `Cargo.lock`, `.github/workflows/ci.yml` | untouched (the local `proliferate 0.3.50 → 0.4.2` lock drift is pre-existing on `main`, unrelated to this PR — reverted before my controls) |

---

**Reviewed @ dc70e43c4**

Verdict: **CLEAN.** The single disclosed behavior change is real, correctly described, and
an unambiguous improvement (irrecoverable partial teardown → clean rejection with the
identical error). Two nits (F1 undocumented-but-benign session-list hoist, F2 currently
unreachable fail-closed guard), neither blocking. Effect order preserved in both use cases,
error surfaces and HTTP mappings unchanged, policy genuinely pure and inside the lint fence,
no rule with two homes, and the test suite discriminates under two independent inversions.

PR #1668, opened 2026-08-06
refactor(anyharness): fold remaining domain SQL into stores + close the live probe fetch (grid PR 5b+5c)

pablonyx:
## Summary

Grid PR 5b+5c: fold the last domain SQL into stores, and close the one
non-test `LIVE_DOMAIN_STORE_IMPORT` break in `live/sessions/probe.rs`.

### PR 5b — fold remaining domain SQL into stores

Pure relocation, no query changes; rows never escape the stores.

| File | What moved |
|---|---|
| `domains/plans/service.rs` | 3 inline `tx.query_row(...)` plan-lookup calls → `PlanStore::find_by_id_in_tx(tx, plan_id)`; 2 calls to a private `plan_has_interaction_link` helper → `PlanStore::has_interaction_link_in_tx(tx, &plan.id)` (both new store methods, read inside the caller's existing transaction). The now-unused private helper was deleted. 628 → 607 lines. |
| `domains/plans/decision_op.rs` | Inline `#[cfg(test)]` fixture's raw `INSERT INTO sessions` SQL → `SessionStore::insert(&SessionRecord { .. })`. |
| `domains/activity/feeds.rs` | Same pattern: inline test fixture's raw `INSERT INTO sessions` → `SessionStore::insert(&SessionRecord { .. })`. |
| `domains/sessions/links/completions.rs` | `LinkCompletionStore` and its row types (`LinkCompletionRecord`, `LinkWakeScheduleRecord`, `LinkCompletionInsert`) physically relocated via `git mv` to `domains/sessions/store/link_completions.rs` (`store/links.rs` was already taken by an unrelated file, hence `link_completions.rs`). A thin re-export shim stays at the old path so every external caller's import path (`crate::domains::sessions::links::completions::..`) is unchanged. Declared `pub(crate) mod link_completions;` in `store/mod.rs`. |

`DOMAIN_SQL_OUTSIDE_STORE` allowlist stanza:
- Before: `# 6 files, 38 occurrences.`
- After: `# 2 files, 2 occurrences.`

The 2 remaining rows (`workflows/workspace_materialization/test_support.rs`,
`workspaces/access_gate.rs`) are engine-invisible `cfg(test)` boundaries
later PRs still own, per the task's explicit instruction to leave them
alone.

### PR 5c — close the live probe fetch

`live/sessions/probe.rs` imported
`domains/agents/readiness/service::resolve_agent_unrouted` directly — the
single non-test break in "live never fetches." Two remedies were on the
table: pass the resolved fact in at the call boundary, or wire a
capability trait in `app/`. Chose the smaller diff:

- No `ActorCapabilities`-style seam reaches `live::sessions::probe` today —
  `probe_agent` is a standalone async fn, not an actor with launch-time
  capability injection.
- `ProbeOptions` already carries a caller-supplied `runtime_home` for the
  same install-artifact question, so adding a caller-supplied `resolved:
  ResolvedAgent` field is a natural, incremental fit — no new trait
  machinery.

Added `ProbeOptions.resolved: ResolvedAgent`; `probe_agent` now reads
`&options.resolved` instead of resolving it internally. Updated all three
callers to resolve before constructing `ProbeOptions`:
- `AcpProbeRunner::run` (`domains/agents/model_snapshot/probe.rs`)
- the `catalog-probe` CLI (`anyharness/src/commands/catalog_probe.rs`, both
  the main options and the per-trial options)
- the probe-materialization test
  (`route_auth/probe_materialization_tests/materialization_tests.rs`)

Added a `resolve_agent_unrouted_by_kind(kind: &AgentKind, runtime_home:
&Path) -> Result<ResolvedAgent, AgentKindNotRegisteredError>` helper in
`readiness/service.rs` so the registry-lookup-then-resolve pair — now
required at each call site instead of once inside `probe_agent` — has one
implementation for the CLI caller (the other two callers keep the manual
registry lookup they already had reason to do inline).

`LIVE_DOMAIN_STORE_IMPORT` allowlist stanza:
- Before: `# 7 files, 19 occurrences.`
- After: `# 6 files, 18 occurrences.`

### Fallout: catalog_probe.rs split

Adding the resolve call inline in the CLI pushed `catalog_probe.rs` from
599 to 613 lines (613 → 606 after consolidating into
`resolve_agent_unrouted_by_kind`) — over the 600-line ceiling. Per doctrine
("never raise a pin, add a new file instead"), split the credential
capture/isolation/auth-context half (`IsolationDirs`, `ProbeSecrets`,
`auth_env_for_context`, `CREDENTIAL_ENV_VARS`, `probe_aws_region`,
`trial_model_name`, `opencode_isolation_env`, `isolation_env`, and their
`#[cfg(test)]` tests — none of which depend on `ProbeOptions`/`probe_agent`)
into a sibling submodule, `commands/catalog_probe/catalog_probe_auth.rs`,
following the file+dir module pattern already used elsewhere in the repo
(`route_auth/probe_materialization.rs` + `probe_materialization/scratch.rs`).
`catalog_probe.rs`: 196 lines; `catalog_probe_auth.rs`: 428 lines. Both well
under the ceiling.

Re-pinned `live/sessions/probe.rs`'s max-lines allowlist entry (683 → 679)
to match its shrinkage after the internal resolve logic moved out.

### Anchor repoints

Moving `completions.rs` broke 3 hardcoded anchors in
`test_known_truths_the_hardened_rules_must_see` that pointed at its old
`DOMAIN_SQL_OUTSIDE_STORE` lines. Repointed to the 2 real remaining
offenders in the repo (measured via `collect_violations()`, not guessed):
`workflows/workspace_materialization/test_support.rs:159` (format!-built
`SELECT COUNT(*)` head) and `workspaces/access_gate.rs:324` (`DROP TABLE`
line). Suite stays at exactly 90 tests.

## Gates

```
$ python3 scripts/check_anyharness_boundaries.py
AnyHarness boundary check passed.

$ python3 -m unittest scripts.test_check_anyharness_boundaries -v
Ran 90 tests in 2.628s
OK

```

Negative controls (verified, then reverted):
- Re-added a raw `SELECT * FROM plans` query to `plans/service.rs` →
  checker exits 1, naming
  `plans/service.rs:17: [DOMAIN_SQL_OUTSIDE_STORE]`. Reverted → exit 0.
- Re-added the `resolve_agent_unrouted` import to `probe.rs` → checker
  exits 1, naming
  `live/sessions/probe.rs:19: [LIVE_DOMAIN_STORE_IMPORT]`. Reverted →
  exit 0.

No Rust compiler was run (per the task's hard constraint — this machine
OOM-crashes on concurrent cargo builds); CI's "Cargo check & test" job
validates Rust separately. Every moved/touched file and its callers were
re-read manually for import paths, visibility, module declarations, and
unused imports.

## Test plan

- [ ] CI's "Cargo check & test" job compiles and runs the full test suite
- [ ] `check_anyharness_boundaries.py` / boundaries unittest / `check_max_lines.py` CI jobs stay green


pablonyx:
## Adversarial review — PR #1668 (grid PR 5b+5c)

**Verdict: CLEAN.** No P1s. Two P2s (both about the 5c helper, neither blocking), three nits, one out-of-scope observation.

Review context: the builder was forbidden from running cargo, so compile-correctness and fixture drift were the primary hunt. I ran cargo. `cargo check -p anyharness-lib -p anyharness -j 6 --all-targets` is clean; all 21 warnings are pre-existing and in files this PR does not touch. CI at this SHA is fully green (Cargo check & test, Analyze (rust), Repo shape checks, all 24 checks pass) — but CI compiling is not semantics preserved, so every claim below is evidenced independently.

---

### 1. Query-equivalence audit

Every moved query, old text vs new text:

| Query | Old site | New site | Byte-equal SQL | Params | Transaction context |
|---|---|---|---|---|---|
| `SELECT * FROM plans WHERE id = ?1` (×3 call sites) | `plans/service.rs` inline `tx.query_row(.., super::store::map_plan)` at three decision paths | `PlanStore::find_by_id_in_tx(tx, plan_id)` | **yes** | `[plan_id]`, identical | **preserved** — takes `tx: &Connection`, opens no connection; all 3 sites remain inside the same `with_tx_anyhow(\|tx\| ..)` closure |
| `SELECT EXISTS(SELECT 1 FROM plan_interaction_links WHERE plan_id = ?1)` (×2 call sites) | free fn `plan_has_interaction_link(tx, ..)` in `service.rs` | `PlanStore::has_interaction_link_in_tx(tx, plan_id)` | **yes** | `[plan_id]`, identical | **preserved** — same `tx`, same closure |
| all ~13 `session_link_completions` / `session_link_wake_schedules` statements | `domains/sessions/links/completions.rs` | `domains/sessions/store/link_completions.rs` | **yes — whole file byte-identical** | unchanged | unchanged |

Row-mapper: `find_by_id_in_tx` uses `map_plan`, the same mapper the inline call passed as `super::store::map_plan`. No `ORDER BY`/`LIMIT` anywhere in the moved plans queries, so no ordering to drift.

Error mapping preserved exactly. `find_by_id_in_tx` returns `rusqlite::Result<PlanRecord>` (no `.optional()`), so `QueryReturnedNoRows` still propagates and all three callers still map it:
```rust
rusqlite::Error::QueryReturnedNoRows => PlanDecisionTxError::NotFound,
other => PlanDecisionTxError::Sqlite(other),
```
`has_interaction_link_in_tx` keeps `.map(|value| value != 0)` on an `i64` — same coercion, same `?` into `PlanDecisionTxError::Sqlite` via `From`.

The deleted free helper `plan_has_interaction_link` has zero remaining references (grepped crate-wide).

**Store-file relocation proven, not eyeballed:**
```
$ git show origin/main:.../domains/sessions/links/completions.rs > /tmp/old.rs
$ diff /tmp/old.rs .../domains/sessions/store/link_completions.rs
(no output)
```

### 2. Fixture-semantics audit — equivalent, verified column-by-column

Two identical swaps (`plans/decision_op.rs`, `activity/feeds.rs`): a raw 6-column `INSERT INTO sessions` → `SessionStore::new(db.clone()).insert(&SessionRecord{ ..25 fields.. })`. `insert_session_row` writes 25 columns where the raw INSERT wrote 6 and let SQLite default the rest, so this only holds if every struct value equals the corresponding schema default. I read `0001_initial.sql` plus every `ALTER TABLE sessions` migration and checked each:

- `mcp_binding_policy: SessionMcpBindingPolicy::InheritWorkspace` → `as_str()` = `"inherit_workspace"` = migration 0034's `DEFAULT 'inherit_workspace'`. **match**
- `subagents_enabled: true` → `1` = migration 0031's `DEFAULT 1`. **match**
- `origin: None` → `encode_origin_json(&None)` = `Ok(None)` → NULL. **match**
- every other written column (`native_session_id`, `agent_auth_contexts`, `requested_model_id`, `current_model_id`, `requested_mode_id`, `current_mode_id`, `title`, `thinking_level_id`, `thinking_budget_tokens`, `last_prompt_at`, `closed_at`, `dismissed_at`, `mcp_bindings_ciphertext`, `mcp_binding_summaries_json`, `system_prompt_append`, `action_capabilities_json`) is a nullable `TEXT`/`INTEGER` with no default → explicit `None` writes the same NULL the omission produced. **match**
- `pending_prompt_seq_cursor` (migration 0059, `NOT NULL DEFAULT 0`) is written by **neither** path — default 0 before and after. **match**
- legacy `model_id` / `mode_id` (0001, deliberately never dropped) are nullable and unwritten by both. **match**

`id`/`workspace_id`/`agent_kind`/`status`/`created_at`/`updated_at` carry the same literals (`'session-1'`, `'workspace-1'`, `'claude'`, `'idle'`, `'now'`, `'now'`). No timestamp is generated, no serialization the fixture previously bypassed is introduced (`SessionStore::insert` is `with_conn` + `insert_session_row`, no events, no side tables). No assertion in either test reads any of the 19 newly-explicit columns.

Tests run: `domains::plans::decision_op` **4/4 pass**, `domains::activity::feeds` **5/5 pass**.

### 3. Row-escape audit — no widening

The shim re-exports exactly the four items that were `pub` in the old module, nothing more:
```rust
pub use crate::domains::sessions::store::link_completions::{
    LinkCompletionInsert, LinkCompletionRecord, LinkCompletionStore, LinkWakeScheduleRecord,
};
```
`LinkCompletionParseError` (private struct), `map_completion`, `parse_outcome` all stayed private in the moved file. New module is `pub(crate) mod link_completions;` in `store/mod.rs` — crate-internal, so nothing leaked to the crate's external surface either. Row structs did not escape store privacy: they were already `pub` at the old path and are reached through the same names.

### 4. Probe change (5c) — shape, not a power

`ProbeOptions.resolved: ResolvedAgent` is a **value**, not a capability: `ResolvedAgent` is a plain `#[derive(Debug, Clone)]` data struct (descriptor, status, credential_state, auth_slots, cli_auth_state, native, agent_process, spawn, credentials_from_route) — no handles, no `Db`, no trait objects, no closures. `probe_agent` reads only `resolved.agent_process.path` (the install gate) and `resolved.descriptor.kind`, and passes `resolved` to `spawn_agent_process` — the same three uses as before, with `&resolved` → `resolved` mechanically adjusted for the new reference binding. Imports narrowed to `use crate::domains::agents::model::{AgentKind, ResolvedAgent};`; the domain-service and registry imports are gone. It carries no secret material (`CredentialState` is a 4-variant enum; `AuthSlotSpec` holds env var *names*, not values), so nothing sensitive newly crosses into `live/`.

All three callers resolve identically — same function (`resolve_agent_unrouted`), same runtime_home:

| Caller | Resolve | runtime_home passed | Not-in-registry error |
|---|---|---|---|
| `catalog_probe.rs` CLI | `resolve_agent_unrouted_by_kind(&agent_kind, &runtime_home)?` | the CLI's `runtime_home` (same value later put on `ProbeOptions.runtime_home`) | `AgentKindNotRegisteredError` → `anyhow` via `?`, message `agent kind {} not in registry` |
| `AcpProbeRunner::run` | hand-rolled registry lookup + `resolve_agent_unrouted(descriptor, &request.runtime_home)` | `request.runtime_home` (same value put on `ProbeOptions.runtime_home`) | `ProbeError::Failed { detail: format!("agent kind {} not in registry", ..) }` — **verbatim the old `anyhow!` text** |
| `materialization_tests.rs` | hand-rolled lookup + `resolve_agent_unrouted(descriptor, home.path())` | `home.path()` (same as `ProbeOptions.runtime_home`) | `.expect("claude is in the built-in registry")` (test) |

No caller substituted `resolve_agent` (route-aware) for `resolve_agent_unrouted` — unroutedness, which is the semantically load-bearing choice here (a route supplies credentials, not binaries), is preserved at all three. Error-path parity holds: the failure mode was `Err` before and is `Err` now, with the same message.

`resolve_agent_unrouted_by_kind` is **not** a duplicate of existing logic — I grepped every `resolve_agent*` in the crate. The registry-lookup-then-`resolve_agent_unrouted` pair appears at `auth/login.rs:83`, `model_snapshot/targets.rs:75`, `installer/reconcile/execution.rs:325`, and now the three probe sites; the new fn is the first named wrapper for it. See P2-1 for the doc's over-claim.

Tests: `TZ=UTC cargo test -p anyharness-lib -j 6 --lib model_snapshot` → **47 passed, 0 failed**, including the two real-child tests that exercise the new resolve-first path end to end (`runner_tests::cancelling_a_probe_kills_the_child_then_removes_the_scratch`, `runner_tests::a_timed_out_probe_kills_its_child_and_reports_a_timeout`).

### 5. catalog_probe split — pure relocation, no leak of internals

Verified mechanically: extracted old `catalog_probe.rs` lines 188-599 and diffed against new `catalog_probe_auth.rs` lines 14-429. The **only** deltas are 5 visibility widenings plus one 3-line signature reflow:

`IsolationDirs`, `auth_env_for_context`, `ProbeSecrets`, `capture_and_scrub`, `trial_model_name` → `pub` (each is genuinely needed by the sibling module; `mod catalog_probe_auth;` + `use catalog_probe_auth::{auth_env_for_context, trial_model_name, IsolationDirs, ProbeSecrets};`).

Critically, the secret-handling internals stayed private: `ProbeSecrets::require`, `ProbeSecrets::get`, `CREDENTIAL_ENV_VARS`, `probe_aws_region`, `opencode_isolation_env`, `isolation_env` are all still private to the new module. So the split did not widen access to credential plumbing — only to the two entry points and the two RAII/collection types. Module wiring is the standard `foo.rs` + `foo/` sibling layout; unused imports in the shrunk parent were correctly pruned (no new warnings). Both files under the 600 ceiling (196 and 428) with no new max-lines allowlist entry.

### 6. Compile + tests

- `cargo check -p anyharness-lib -p anyharness -j 6 --all-targets` → clean, 0 errors, no new warnings.
- `domains::plans::` → **29/29 pass**; `domains::activity::feeds` → **5/5**; `link_completions` → **2/2** (now `domains::sessions::store::link_completions::tests::*`); `model_snapshot` → **47/47**; `probe_materialization` → 11 pass / 1 fail, that one fail proven unrelated (see out-of-scope note).
- `python3 scripts/check_anyharness_boundaries.py` → pass. `check_max_lines.py` → pass. `check_anyharness_old_paths.py` → pass.

### 7. Ratchet mechanics — re-measured independently

Rather than trusting the checker's pass/fail, I imported the module and counted violations directly:
```
$ python3 -c "import check_anyharness_boundaries as m; ..."
DOMAIN_SQL_OUTSIDE_STORE files= 2 occ= 2
LIVE_DOMAIN_STORE_IMPORT files= 6 occ= 18
```
Both match the rewritten headers (`# 2 files, 2 occurrences.` and `# 6 files, 18 occurrences.`) exactly. The four deleted DOMAIN_SQL rows and the one deleted LIVE_DOMAIN_STORE_IMPORT row (`live/sessions/probe.rs`) correspond to files that genuinely dropped to zero.

max-lines entries verified by `wc -l`: `plans/service.rs` = **607** (header says 607), `live/sessions/probe.rs` = **679** (header says 679). Both entries were ratcheted down, not left stale.

The 3 repointed test anchors point at real offenders at the exact lines claimed — both appear in the live violation list above, and I read the source lines:
- `domains/workflows/workspace_materialization/test_support.rs:159` → `conn.query_row(&format!("SELECT COUNT(*) FROM {table}"), [], |row| {` ✔
- `domains/workspaces/access_gate.rs:324` → `conn.execute("DROP TABLE workspace_access_modes", [])?;` ✔

Suite size confirmed: `Ran 90 tests ... OK` — exactly 90, no test silently dropped when the three anchors were rewritten into two entries plus commentary.

### 8. Negative controls (mine, distinct from the builder's)

**(a) Re-inline a moved query → checker must fire.** I put the `plan_interaction_links` EXISTS query back inline at `plans/service.rs:262`:
```
anyharness/crates/anyharness-lib/src/domains/plans/service.rs:262: [DOMAIN_SQL_OUTSIDE_STORE] SQL belongs in the domain's store — move this query into store.rs/store/**; see anyharness-structure.md
EXIT=1
```
Reverted → `AnyHarness boundary check passed. EXIT=0`. The ratchet is genuinely load-bearing for this fold, not tautologically satisfied by the allowlist edit.

**(b) Delete the shim's re-export → compilation must break.** Replaced the `pub use` with a comment; `cargo check -p anyharness-lib -j 6`:
```
error[E0432]: unresolved import `crate::domains::sessions::links::completions::LinkCompletionStore`
  --> anyharness/crates/anyharness-lib/src/app/mod.rs:57:5
error[E0432]: unresolved imports `...::LinkCompletionRecord`, `...::LinkCompletionStore`, `...::LinkWakeScheduleRecord`
  --> anyharness/crates/anyharness-lib/src/domains/cowork/delegation/service.rs:13:5
error[E0432]: unresolved import `crate::domains::sessions::links::completions::LinkCompletionRecord`
  --> anyharness/crates/anyharness-lib/src/domains/cowork/runtime.rs:27:5
error[E0432]: unresolved imports `...::LinkCompletionRecord`, `...::LinkWakeScheduleRecord`
  --> anyharness/crates/anyharness-lib/src/domains/sessions/subagents/model.rs:2:5
error[E0432]: unresolved imports `...::LinkCompletionInsert`, `...::LinkCompletionStore`
  --> anyharness/crates/anyharness-lib/src/domains/sessions/subagents/store.rs:2:52
error[E0425]: cannot find type `LinkCompletionInsert` in module `crate::domains::sessions::links::completions`
   --> anyharness/crates/anyharness-lib/src/domains/cowork/delegation/service.rs:409:78
error: could not compile `anyharness-lib` (lib) due to 6 previous errors
```
6 errors across 5 caller files (plus the one inline path at `delegation/service.rs:409`). The shim is load-bearing, not dead code. Reverted; worktree back to pristine (only a pre-existing `Cargo.lock` version-line delta remains, untouched by me).

---

## P2s (real, non-blocking)

**P2-1 — `resolve_agent_unrouted_by_kind`'s doc contradicts the code; the stated de-duplication did not happen.** Its doc says it exists "For callers that only have the kind (every `live::sessions::probe::ProbeOptions` builder: `AcpProbeRunner::run`, the `catalog-probe` CLI, and probe-materialization tests), so the registry-lookup-then-resolve pair … has **one implementation instead of three**." But only `catalog_probe.rs:69` calls it. `AcpProbeRunner::run` (`model_snapshot/probe.rs:133-142`) and `materialization_tests.rs:465-471` each hand-roll the lookup+resolve. So the pair has **three** implementations, one of which happens to be the shared helper. Either route the other two through the helper — which also removes the duplicated `format!("agent kind {} not in registry", ..)` string that currently exists in two places and must be kept in sync by hand — or correct the doc. Not blocking: behaviour is identical either way, and the duplicated message text is currently byte-equal.

**P2-2 — the resolve moved a blocking FS scan + a `node --version` subprocess into an async body with no `spawn_blocking`.** In `AcpProbeRunner::run` (an `async fn` via `#[async_trait]`), `resolve_agent_unrouted` now runs **before** `std::thread::Builder::spawn`, i.e. on the caller's tokio worker. `resolve_agent_in_scope` does artifact-path stat'ing, `dirs::home_dir()`, manifest reads, credential/auth-slot discovery under `$HOME`, and for Claude `detect_runtime_compatibility_issue` shells out to `Command::new(node).arg("--version").output()` synchronously. Previously all of that happened inside `probe_agent`, which runs on the probe's own dedicated thread + current-thread runtime. This is a small, bounded stall (one subprocess), the probe path is already semaphore-limited and low-frequency, and nothing observable changes — but it is a real doctrine wobble that the "pure relocation" framing hides. `spawn_blocking` (or moving the resolve inside the spawned thread and passing the descriptor instead) would restore the old placement.

Related and worth noting explicitly rather than filing separately: the not-in-registry failure now fires **earlier** in `AcpProbeRunner::run`, before `materialize_for_probe`, so a bad kind no longer creates and immediately drops a `ProbeScratch`. That is a strict improvement (no scratch churn) and the error text is preserved verbatim, so no behavioural regression — but it is a control-flow reordering, not a pure move.

## Nits

1. `find_by_id_in_tx` makes `"SELECT * FROM plans WHERE id = ?1"` the **fourth** textual copy of that string inside `store.rs` (also at :49 in `find_by_id` with `.optional()`, :83 in the new fn, and :264 in the tail of `update_decision`). A `const PLAN_BY_ID_SQL` would make the store's own duplication visible — the fold moved the SQL into the store but didn't dedupe it there.
2. The new call shape carries a redundant block rustfmt would collapse:
   ```rust
   let plan = PlanStore::find_by_id_in_tx(tx, plan_id).map_err(|error| {
       match error { .. }
   })?;
   ```
   (`|error| match error { .. }` is the formatted form.) Cosmetic only — CI has no fmt or clippy gate, so this will not go red; it will just show up as churn the next time someone runs `make fmt` on this file.
3. `has_interaction_link_in_tx`'s signature is wrapped across lines where rustfmt would keep it on one (`pub fn has_interaction_link_in_tx(tx: &Connection, plan_id: &str) -> rusqlite::Result<bool> {` fits). Same cosmetic class as (2).

## Out of scope — a pre-existing failing test I hit, NOT caused by this PR

`route_auth::probe_materialization_tests::sweep_tests::the_sweep_removes_only_abandoned_and_old_scratch_roots` fails on this machine at `sweep_tests.rs:69` ("an unparseable old-mtime dir must be removed"). Provenance: the file is untouched by this PR (`git diff --name-only origin/main...HEAD | grep -i sweep` → empty; last touched by b9464e5cd). Mechanism: `format_touch_stamp` renders the target instant with `chrono::DateTime::from_timestamp(..).format("%Y%m%d%H%M.%S")` — a **UTC** rendering — then feeds it to `touch -t`, which interprets the stamp in **local** time. I reproduced the arithmetic in Python under PDT: the "aged" dir's mtime lands ~23,759 s in the *future*, so `dir_age_exceeds` returns false and the dir survives. Re-running the same test with `TZ=UTC` → **ok**. It only fails in timezones west of UTC; CI runs UTC, which is why CI is green. Worth a separate issue (the fixture should render the stamp in local time, or set `TZ=UTC` for the test), but it is not a #1668 defect.

---

Reviewed @ d0cf13738

pablonyx:
Closed both P2s from the review (new head: 2a6d6b92dec4f71a8cb159203a60b61eea103ec2).

**P2-1 (doc contradiction / duplicated lookup+error string):** Converged all three call sites onto `resolve_agent_unrouted_by_kind` — `catalog-probe`'s CLI already used it; `AcpProbeRunner::run` and the probe-materialization test each hand-rolled the registry-lookup-then-resolve pair, so both now call the helper instead. The `"agent kind {} not in registry"` error string now exists in exactly one place (`readiness/service.rs`), and the helper's doc comment was updated to describe reality (all three callers call it directly).

**P2-2 (blocking work on the async worker):** Moved the resolve call (blocking FS scan + `node --version` subprocess via `detect_runtime_compatibility_issue`) from `AcpProbeRunner::run`'s async body into the dedicated `std::thread::Builder::spawn` thread that already owns the rest of the probe's blocking/effectful work, rather than adding a separate `spawn_blocking` — the thread was already the right place, so no extra hop was needed. It still runs before `materialize_for_probe`, preserving the "fails before touching the scratch" property called out as worth keeping. Error semantics are unchanged (`ProbeError::Failed` with the same detail text).

Also split the helper's registry lookup into a private `find_descriptor_by_kind` so the not-in-registry arm is unit-testable (the real bundled registry always has every `AgentKind`, so that arm was otherwise dead in tests) — added `find_descriptor_by_kind_reports_a_kind_missing_from_the_registry` and ran a negative control (inverted the lookup condition, confirmed that exact test failed with the expected panic, reverted).

Gates:
- `cargo check -p anyharness-lib -p anyharness --all-targets` — clean, same pre-existing warning set, none new
- `cargo test -p anyharness-lib model_snapshot` — 47/47 passed
- `python3 scripts/check_anyharness_boundaries.py` — exit 0
- `python3 -m unittest scripts.test_check_anyharness_boundaries` — 90/90

🤖 Generated with Claude Code

pablonyx:
Greptile's P2 (legacy `links::completions` path kept alive) adjudicated before merge: the file is a deliberate, documented 11-line re-export shim — the SQL physically moved to `store/link_completions.rs` per the slice's scope, and the shim preserves the 5 existing caller import paths with identical types. No behavior risk. Follow-up owed in the grid import-migration loops: rewrite the 5 callers to `store::link_completions` and delete the shim.

PR #1670, opened 2026-08-06
refactor(anyharness): workspaces gets one entry surface — WorkspaceService deleted (grid PR 8)

pablonyx:
## Grid PR 8 — Workspaces: one entry surface (violation #6)

Deletes `WorkspaceService` outright. `WorkspaceRuntime` is now the domain's
one entry surface, matching what was already true in practice: it's the
`AppState` field with ~30 consumers, while `WorkspaceService` had exactly one
live call site in the whole tree.

### Divergence table (verified against actual code, not the plan's estimate)

| Method | Verdict | Evidence |
| --- | --- | --- |
| `get_workspace` | DIVERGED | Runtime's copy (`runtime/lifecycle.rs:16-23`) schedules a background branch-refresh via `WorkspaceBranchRefreshCoordinator`; service's (`service/metadata.rs:11-13`) was a pure `store.find_by_id` read. |
| `list_workspaces` | DIVERGED | Runtime adds a branch-refresh batch + `[anyharness-latency]` tracing spans (`runtime/lifecycle.rs:72-95`); service's copy (`service/metadata.rs:61-63`) had zero callers. |
| `set_display_name` | IDENTICAL bodies | Both files carried their own `MAX_WORKSPACE_DISPLAY_NAME_CHARS = 160`; service's copy was dead — nothing called `WorkspaceService::set_display_name`. |
| `detect_setup` | IDENTICAL bodies | Same story; service's copy was dead. |

`WorkspaceService`'s only live caller anywhere in the tree was
`MobilityService::load_workspace` (`domains/mobility/service.rs:218-225`,
via `get_workspace`). Its other three methods had zero call sites — confirmed
by grep, not assumption.

### Deletion inventory

- `anyharness/crates/anyharness-lib/src/domains/workspaces/service/mod.rs` — 16 lines (struct + ctor + the `MAX_WORKSPACE_DISPLAY_NAME_CHARS` const)
- `anyharness/crates/anyharness-lib/src/domains/workspaces/service/metadata.rs` — 64 lines (the 4 methods)
- Total: 80 lines deleted, matching the plan's estimate exactly.
- `pub mod service;` removed from `domains/workspaces/mod.rs`.
- Construction (`app/mod.rs:209`, was `let workspace_service = Arc::new(WorkspaceService::new(...))`) and the pass-through into `MobilityWiringDeps` (`app/mod.rs:517`) deleted, along with the now-unused `use ...service::WorkspaceService` import.
- `MobilityWiringDeps.workspace_service: Arc<WorkspaceService>` field removed from `app/mobility.rs`; `MobilityService::new`'s first parameter and struct field renamed `workspace_service` → `workspace_runtime: Arc<WorkspaceRuntime>` in `domains/mobility/service.rs`.
- `MobilityService::load_workspace` now calls `self.workspace_runtime.get_workspace(...)` instead of `self.workspace_service.get_workspace(...)`.

### Named design call: mobility's `load_workspace` gains the branch-refresh side effect

Adopted the plan's default ("probably yes: mobility wants fresh branches")
after checking the concrete race the plan flagged as the reason not to:
does the runtime's background branch-refresh (spawned off `get_workspace`/
`list_workspaces`) interact badly with mobility's `destroy_source_workspace`
or `preflight_workspace`?

Traced the coordinator (`domains/workspaces/branch_refresh.rs`):
- `schedule_refresh` spawns a detached thread that re-reads the row via
  `store.find_by_id` *inside* the background thread before writing anything.
  If the row is gone by then, it returns `SkippedMissing` — no error.
- The write path, `update_current_branch`, is `UPDATE workspaces SET
  current_branch = ?2 WHERE id = ?1` — if the row was deleted first, this
  is a normal zero-row UPDATE, not an error.
- `destroy_source_workspace` (`domains/mobility/runtime/destroy_source.rs`)
  and `park_local_workspace`/`remove_worktree_workspace`
  (`domains/workspaces/runtime/materialization.rs`) delete the workspace row
  via `WorkspaceDeleteWorkflow::delete_workspace_record`, a single
  `db.with_tx` transaction. There's no window where the branch-refresh
  thread's read-then-write straddles that transaction in a way that could
  write stale data back — worst case it's a no-op after the delete commits.
- `preflight_workspace` only ever *reads* through `load_workspace`; it has no
  write path that a background branch update could corrupt.

Conclusion: harmless. The coordinator was built read-tolerant of a vanished
row from the start (the `SkippedMissing` variant exists exactly for this),
so mobility inherits "workspace branches slowly become fresher in the
background" with no new failure mode. Went with the runtime's `get_workspace`
as-is rather than adding a pure `find_workspace` read.

Evidence this holds under test: `cargo test -p anyharness-lib mobility` and
the destroy-source fail-closed tests below all pass unchanged; the branch
refresh's own behavior is covered by
`domains::workspaces::runtime::tests::background_branch_refresh_persists_changed_branch_and_throttles`
(`domains/workspaces/runtime/tests.rs:351`) plus the two "returns stored
branch without inline git refresh" tests at lines 306 and 329 in the same
file — all green in this PR's run.

### Const fold

`MAX_WORKSPACE_DISPLAY_NAME_CHARS` existed in two places
(`workspaces/service/mod.rs:5` and `workspaces/runtime/lifecycle.rs:13`).
Deleting the service file automatically folds it to the one remaining copy
in `runtime/lifecycle.rs` — no separate edit needed.

### Pin re-measures (`scripts/max_lines_allowlist.txt`)

- `anyharness/crates/anyharness-lib/src/app/mod.rs`: 693 → 690 (removed the
  `use` line, the `workspace_service` construction line, and the
  `workspace_service:` field pass in the mobility-wiring call).
- `anyharness/crates/anyharness-lib/src/domains/mobility/service.rs`:
  unchanged at 851 — only in-place identifier substitutions (`workspace_service`
  → `workspace_runtime`, one `use` line swapped for another), no line-count
  change.
- No allowlist rows existed for `domains/workspaces/service/mod.rs` or
  `metadata.rs` (they were never oversized), so nothing to delete there.

### Boundaries allowlist / anchor delta

(The max-lines allowlist DID change — the 693 → 690 re-pin above. This
section covers only the boundaries allowlist.)

- `scripts/anyharness_boundaries_allowlist.txt`: no rows referenced
  `domains/mobility/service.rs` for `DOMAIN_LIVE_VALVE` (that import moved to
  `runtime/` in grid PR 6a already), and no rows referenced the deleted
  `workspaces/service/*` files — no boundaries-allowlist changes needed.
- `scripts/test_check_anyharness_boundaries.py`: the paths this PR touches
  (`app/mod.rs`, `domains/mobility/service.rs`, `domains/workspaces/service/*`)
  only appear as fabricated fixture paths inside the test file's tempdir
  scaffolding (e.g. `domains/workspaces/service.rs` as a synthetic
  non-existent single-file path used to test the `POLICY_PURITY`/
  `API_STORE_ESCAPE` rules) — none of them anchor to the real deleted files.
  No anchor updates required; still exactly 90 tests.

### Gates (all green)

```
$ cargo check -p anyharness-lib -j 6 --all-targets
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 33.60s
(no new warnings — only pre-existing unrelated ones)

$ cargo test -p anyharness-lib -j 6 workspaces
test result: ok. 117 passed; 0 failed; 0 ignored; 0 measured; 1371 filtered out

$ cargo test -p anyharness-lib -j 6 mobility
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 1473 filtered out

$ cargo test -p anyharness-lib -j 6 destroy_source
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 1486 filtered out
(destroy_source_fails_closed_while_controlled — the exact fail-closed
 guarantee the design call needed to not break — still holds)

$ python3 scripts/check_anyharness_boundaries.py
AnyHarness boundary check passed.

$ python3 -m unittest scripts.test_check_anyharness_boundaries
Ran 90 tests in 2.574s
OK

```

### Negative control — zero remaining references

```
$ grep -rn "\bWorkspaceService\b" anyharness/crates/anyharness-lib/src/ --include="*.rs"
(no output, exit 1 — zero matches; only WorkflowWorkspaceService, an
 unrelated type in domains/workflows/workspace_materialization, remains)
```

### Recon corrections vs the plan

- The plan's line numbers had drifted slightly (`app/mod.rs:209`/`:515` in
  the plan text vs. `:209`/`:514` in the checked-out `origin/main` head used
  for this branch) — re-located by grep rather than trusted at face value,
  as instructed.
- `Arc<WorkspaceService>` was never an `AppState` field directly — it only
  lived inside `mobility::MobilityWiringDeps`, constructed locally in
  `AppState::new` and handed to `wire_mobility`. So "AppState changes: the
  field for the service goes away" resolved to "the local `let` binding and
  the wiring-deps field go away," not an `AppState` struct field deletion.
- Per-file line counts matched the plan's recon (851/693 for the two pinned
  files, ~80 total for the deleted service) exactly — no other drift found.

Not in scope here (per the plan): re-splitting `WorkspaceRuntime` into
runtime-wraps-service (PR 13-era) and the `workspaces_lifecycle.rs` handler
dedup (PR 9, next in the train).


pablonyx:
## Adversarial review — grid PR 8 (workspaces: one entry surface)

**Verdict: CLEAN.** No P1s. Two P2s and one nit, all non-blocking. The builder's race analysis
**held under deeper probing, and is in fact stronger than the PR body claims** — see the
"why the design call is safe" section, which found a structural reason the race is unreachable
in the first place.

Reviewed @ e95c4b46f.

---

### 1. The design call: call-site-by-call-site audit

`MobilityService::load_workspace` → `WorkspaceRuntime::get_workspace`, which schedules a
background branch refresh. Every mobility call site of `load_workspace`, post-change:

| # | Call site | Flow | Is a background branch-refresh acceptable HERE? |
| --- | --- | --- | --- |
| 1 | `service.rs:83` `export_workspace_archive` | durable export | **Yes.** The archive's `branch_name` comes from `current_branch_name(&repo_root)` (`service.rs:92`) — a **direct git read of disk**, not the DB row. The refresh only ever writes the DB `current_branch` column. The recorded branch cannot go stale-vs-disk because the DB value is never on the read path. Additionally the export runs under `FrozenForHandoff` + `require_clean_git_state`, and re-validates runtime state after collecting sessions (`:129`). |
| 2 | `service.rs:237` `can_relocate_existing_archive_session` | install precondition (source ws) | **Yes.** Consumes only `source_workspace.id` and `.path`. Never reads `current_branch`. |
| 3 | `runtime/install.rs:49` `install_workspace_archive` | destination ws | **Yes — no fight with the install's own branch writes.** Install writes *files* and *session rows*; it performs no branch checkout and never writes `current_branch`. Destination branch is fixed earlier by `create_mobility_destination`, a different entry point. Install reads only `workspace.path` / `.id`. There is no write-write pair to interleave. |
| 4 | `runtime/preflight.rs:34` `preflight_workspace` | preflight | **Yes.** `branch_name` is again read from disk (`preflight.rs:46`), and the `local_default_branch_in_use` blocker compares that disk-read value against the repo-root default branch — not the DB column. |
| 5 | `runtime/destroy_source.rs:19` `destroy_source_workspace` | destroy / park | **Yes.** The row-deletion race — probed hard below; unreachable, and a no-op even if reached. |

**(1c) Semantics delta beyond races.** I looked for a mobility path that deliberately wants a
snapshot *without* touching the workspace (different host, retired workspace):

- **Retired workspaces**: no exposure. `select_eligible` (`branch_refresh.rs:127`) skips any
  record whose `lifecycle_state != Active`, so retired rows are never probed.
- **Log noise for legitimately-absent worktrees**: bounded and pre-existing. A missing/unreadable
  worktree yields `probe_current_branch_quiet` → `Err(())` → `Failed`, which is counted into an
  aggregate `tracing::info!` line — **no `error!`/`warn!`, no user-visible surface, no error
  propagation** into the mobility result. Same for a mid-teardown path (`.git` gone, index locked):
  `git rev-parse` exits nonzero → `Failed` → counted, swallowed.
- **Different host**: the RemoteOwned/remote cases still resolve to a local path column; a probe on
  a nonexistent path is the `Failed` no-op above.

**(1b) Mid-mutation worktree:** `refresh_workspace_branch` never mutates the worktree. It runs
exactly one read-only command (`git rev-parse --abbrev-ref HEAD`) with `.output()` — it does not
take the git index lock, so it cannot block or corrupt a concurrent archive/destroy/materialize.

#### Why the design call is safe — stronger than the PR body claims

Three independent reasons, any one sufficient:

1. **Mobility never reads the DB `current_branch` at all.** `grep current_branch` across
   `domains/mobility/` returns only disk-read helpers (`workspace_delta.rs:25`,
   `service.rs:92`, `preflight.rs:46,153`). The one column the refresh writes is not on any
   mobility read path. This makes the whole class of "stale-vs-disk" concerns vacuous.
2. **The refresh was already scheduled on these exact workspaces before this PR.**
   `assert_workspace_not_retired` (`api/http/access.rs:50`) calls
   `workspace_runtime.get_workspace` and already schedules a refresh — and it runs on the
   export path (via `assert_workspace_mode`, `mobility.rs:418`), the destroy path
   (same helper), and preflight (`mobility.rs:62`). Combined with the 30s throttle
   (`BRANCH_REFRESH_THROTTLE`), mobility's new schedule call is in practice a
   **throttle-skipped no-op** — the handler already burned the window microseconds earlier.
   So the behavior delta is even smaller than "gains a side effect".
3. **The DB is one connection behind a `Mutex`** (`persistence/sqlite.rs:10`,
   `Arc<Mutex<Connection>>`). `with_tx` holds that mutex for the whole transaction, so the
   refresh's `find_by_id` re-read and the destroy's delete transaction **cannot interleave** —
   they serialize. The builder's race analysis assumed concurrency and still held; in reality
   the interleaving window it defends against does not exist.

I therefore did **not** find cause to require the plan's fallback remedy (a pure
`find_workspace` read). The named design call is correct as adopted.

---

### 2. Dead-code claim audit — VERIFIED

- `grep -rn "WorkspaceService"` across the repo: **zero** hits for the deleted type. All
  remaining matches are the unrelated `WorkflowWorkspaceService` (workflows domain) and
  `server/.../cloud/workspaces/service.py` (Python, different stack).
- `grep -rn "workspaces::service"`: **zero** hits. Module path fully gone.
- `set_display_name` / `detect_setup` / `list_workspaces` resolving through the OLD service type:
  none anywhere, including tests, CLI crates, and benches. All three survive on
  `WorkspaceRuntime` (`runtime/lifecycle.rs:29/64/72`) with live callers.
- **No test coverage was lost.** `git show origin/main:.../service/metadata.rs` and
  `service/mod.rs` contain **no `#[cfg(test)]`, no `mod tests`, no `#[test]`** — the deleted
  files were untested. `git ls-tree origin/main` confirms the old `service/` dir held exactly
  those 2 files, no sibling test file. The behaviors are covered on the runtime copies
  (`runtime/tests.rs`, 117-test `workspaces` filter, all green).
- `cargo check --all-targets` reports **no unused-import / dead-code warning in any touched
  file** — confirming nothing was orphaned by the deletion.

### 3. Wiring audit — CLEAN

- Construction order holds: `workspace_runtime` at `app/mod.rs:208`, consumed by
  `wire_mobility` at `:512`. 304 lines of separation; dependency direction correct.
- **No Arc cycle.** `WorkspaceRuntime`'s fields (`runtime/mod.rs`) are `store`,
  `delete_workflow`, `repo_root_service`, `runtime_home`, `branch_refresh` — it holds no
  reference back to `MobilityService`/`MobilityRuntime`. The new edge is strictly one-way.
- `MobilityWiringDeps.workspace_service` removal touched no other consumer: the struct is
  `pub(super)` with a single construction site (`app/mod.rs:512`) and single read
  (`app/mobility.rs`). `workspace_runtime` was already a field, now `.clone()`d for the service
  and moved into the runtime — correct, and the doc comment on dependency order is still accurate.

### 4. Const fold — CLEAN

`MAX_WORKSPACE_DISPLAY_NAME_CHARS` is single-homed at `runtime/lifecycle.rs:13`. Grep for
literal `160` across `domains/workspaces/`: exactly one hit, that definition. No silent
duplicate of the rule. (Frontend `maxLength={160}` in rename popovers is the pre-existing
client-side mirror, untouched by this PR and out of scope.)

### 5. Pins + mechanics — ALL VERIFIED

| Gate | Claimed | Measured |
| --- | --- | --- |
| `app/mod.rs` | 690 | **690** ✅ |
| `domains/mobility/service.rs` | 851 unchanged | **851** ✅ |
| `check_max_lines.py` | pass | **exit 0** ✅ |
| `check_anyharness_boundaries.py` | pass | **exit 0** ✅ |
| `cargo check -p anyharness-lib --all-targets` | — | **clean, 30.6s** ✅ |
| `ci.yml` / `Cargo.lock` / `Cargo.toml` | untouched | **absent from diff** ✅ |

Diff is 7 files, +7/−93.

### 6. Test results

| Filter | Expected | Result |
| --- | --- | --- |
| `workspaces` | 117 | **117 passed, 0 failed** ✅ |
| `mobility` | 15 | **15 passed, 0 failed** ✅ |
| `destroy_source` | 2 | **2 passed, 0 failed** ✅ |
| `workspace` (broader) | — | **181 passed, 0 failed** ✅ |

### Negative control (mine, distinct from the builder's)

The builder's safety argument rests on a behavior that had **no test**: the refresh worker's
re-read guard returning `SkippedMissing` when the row was deleted after the snapshot was taken.
I wrote that test, targeting the exact race in call site #5.

First attempt was **too weak and I discarded it**: deleting the row then calling
`refresh_workspace_branches_for_test()` never reaches the worker guard at all, because a deleted
row doesn't appear in `list_execution_surfaces()` — it asserted `scheduled_count == 0`, passing
for the wrong reason. Rewrote it to capture the live snapshot via `get_workspace` (as mobility
does), *then* rename the branch, *then* delete the row, *then* drive
`WorkspaceBranchRefreshCoordinator::run_refresh_for_test` with the **stale snapshot** — which
does reach the guard.

```
PASS   branch_refresh_skips_a_row_deleted_after_the_snapshot_was_taken ... ok
       (scheduled_count == 1, skipped_missing_count == 1, updated_count == 0, row still absent)
```

Inverted the guard it relies on — `branch_refresh.rs:228`, replacing the re-read with
`let current = record;` (trust the stale snapshot):

```
FAILED branch_refresh_skips_a_row_deleted_after_the_snapshot_was_taken
         left: 0
        right: 1        // skipped_missing_count: guard gone, worker proceeded on a dead row
```

Restored the guard → `2 passed` (my test + the pre-existing
`background_branch_refresh_persists_changed_branch_and_throttles`). Scratch test reverted;
working tree clean against e95c4b46f.

The control confirms the guard is load-bearing and actually exercised — not incidentally
satisfied. Note that even with the guard removed, `update_current_branch` is a zero-row no-op
`UPDATE ... WHERE id = ?` against a hard-deleted row (`deletion.rs:45` deletes the row in a
transaction, not a soft retire), so the builder's second line of defense also checks out.

---

### Findings

**P2-1 — the PR body's "no allowlist changes" claim is wrong.**
`scripts/max_lines_allowlist.txt` **is** in the diff: the `app/mod.rs` entry goes `693 → 690`
with an appended `-3: WorkspaceService construction/pass-through deleted, grid PR 8` note. The
change itself is correct and required (the allowlist pins exact counts), and the checker passes.
Only the disclosure is inaccurate — worth fixing in the body so the next reviewer isn't hunting
a phantom.

**P2-2 — consider pinning the deleted-row no-op permanently.**
The `SkippedMissing` re-read guard is now load-bearing for a mobility flow
(`destroy_source_workspace` / `park_local_workspace`), but it is untested on main — my control
was scratch and reverted. Since PR 6b is slated to restructure `destroy_source_workspace` +
`preflight_workspace` effect ordering, an unpinned invariant here is exactly the kind of thing
that regresses silently under that refactor. Suggest landing the test above (or its equivalent)
in this PR or 6b. Non-blocking: the invariant is currently doubly protected (re-read guard +
zero-row `UPDATE`), and the single-connection mutex makes the window unreachable today.

**nit — `MobilityService` now holds `Arc<WorkspaceRuntime>` while living in the "durable
service" layer.** The type name it depends on says "runtime", which reads oddly against the
mobility wiring comment ("the durable service first, then the valve that wraps it"). This is
exactly the doctrine tension the plan already discharges vacuously ("this domain's durable
reads live on the runtime today; re-splitting is PR 13-era work"), so no action — flagging only
so the PR 13 era has the breadcrumb.

### What I tried to break and could not

- Refresh landing between export's branch read and archive write → **impossible**, archive
  branch is a disk read, refresh only writes the DB column.
- Refresh fighting install's branch writes → **impossible**, install performs no branch write.
- Refresh corrupting/blocking a mid-teardown worktree → **impossible**, read-only `rev-parse`,
  no index lock, failures swallowed into an aggregate info log.
- Refresh resurrecting a destroyed workspace row → **impossible**, hard delete + re-read guard
  (proven load-bearing by the negative control) + zero-row `UPDATE`.
- Refresh touching a retired workspace → **impossible**, `select_eligible` filters non-Active.
- Lost coverage from the deleted files → **none**, they had no test modules.
- A fourth surviving caller of the "dead" 3 methods → **none** in tests, CLI crates, or benches.

pablonyx:
**P2-2 closed at e2b829fcd** — landed the deleted-row branch-refresh pin as `branch_refresh_skips_a_row_deleted_after_the_snapshot_was_taken` (`domains/workspaces/runtime/tests.rs`); test-only, no production file changed, `workspaces` filter now 118/118 and the guard-inversion control still fails as expected (left: 0, right: 1). P2-1 (the body's "no allowlist changes" claim vs the real 693→690 edit) remains open as a body-text fix.

PR #1671, opened 2026-08-06
refactor(anyharness): retire state machine leaves the handler — one use case, one policy (grid PR 9)

pablonyx:
Grid PR 9 — structure violation #5. `api/http/workspaces_lifecycle.rs` held the
retire state machine in three inline copies. It now lives in the workspaces
domain as **one use case** over **one pure policy module**, and the three
handlers are auth → parse → one facade call → map error (644 → 89 lines).

> **Stacked PR.** Base is `codex/anyharness-workspaces-one-entry` (PR #1670,
> head `e2b829fcde41515585c8cbbefc14486bd6bf4fb0`), **not** `main`. When #1670
> squash-merges this needs a retarget to `main` plus a main-sync.

---

## 1. The three copies: were they actually identical?

**Mostly yes — and that is the finding.** The recon called them three
re-derivations of "is this retirable"; they are not three rules. They are one
rule written twice, plus a second, genuinely different rule.

| Copy | Site (pre-refactor) | Verdict |
| --- | --- | --- |
| A | `retire_workspace` ~:71, pre-lease early-out | **Byte-identical to B** modulo the binding name (`current` vs `workspace`) |
| B | `retire_workspace` ~:138, under-lease re-read | **Byte-identical to A** |
| C | `retry_retire_cleanup` ~:307 | **Genuinely different question** |

Method: normalized both bindings to a common name and diffed A against B. The
only surviving difference was the identifier itself — same branch order, same
predicates, same message construction. So A and B collapse into **one** policy
fn, `decide_retire_admission`, which the use case *evaluates twice* (once on the
pre-lease read, once on the under-lease re-read). That is two evaluations of one
rule, not two rules — the second read exists because the record can change while
the exclusive lease is being acquired.

Copy C asks "may this **retired** workspace's cleanup be **resumed**?" — a
different predicate over different fields with a different answer shape. It stays
its own policy fn, `decide_retry_admission`.

### Two further duplications the recon did not name

- **A 4th copy:** the `match cleanup_result` blocks at ~:240 and ~:384 were
  **byte-identical** (verified by direct diff). Now `decide_cleanup_outcome`.
- **A 5th copy:** `build_retire_preflight`'s mode selector (~:509) restated copy
  C's exact three-term predicate. `retire_preflight_mode` is now *defined in
  terms of* `decide_retry_admission`, with an exhaustive 24-case cross-product
  test asserting the two can never drift apart.

### Two real inconsistencies — PRESERVED, flagged for a ruling

Deduping surfaced two places where retire and retry genuinely disagree. Per the
brief these are **preserved exactly** as explicit policy branches with named
covering tests, **not** silently unified:

1. **Retired + `cleanup_state == Complete`** → retire reports `AlreadyRetired`
   with `cleanup_succeeded: true`; retry reports `Unavailable`. Defensible (retry
   has nothing to resume), so probably intended.
2. **Retired + `cleanup_state == None`** → retire reports *"retired workspace
   cleanup is not complete: none"* while retry refuses to resume. The caller is
   told cleanup is unfinished **and** that no retry is available — a dead end.
   **This is the one I'd like a ruling on.** Test:
   `retire_and_retry_disagree_about_a_retired_workspace_with_no_cleanup_state`,
   carrying a `// FINDING (flagged for a ruling, behaviour preserved)` comment.

`decide_retry_admission` is deliberately **not** the negation of
`decide_retire_admission`; its doc comment records why.

---

## 2. Before/after semantics

Refactor only — every status code, error code, body field, and idempotency
behaviour is preserved.

| Endpoint / state | Before | After |
| --- | --- | --- |
| `GET .../retire/preflight` | 200, `canRetire = can_retire && mode == ActiveRetire` computed in the handler | 200, identical; narrowing resolved in the domain so the mapper is decisionless |
| `POST .../retire`, active + clean preflight | 200, cleanup runs, `Retired`/`Complete` | identical |
| `POST .../retire`, already retired (retire tombstone) | 200 `AlreadyRetired`, `cleanupSucceeded = (state == Complete)`, cleanup-standing message | identical (`decide_retire_admission`) |
| `POST .../retire`, retired **purge** tombstone | 200 `Blocked` + "purge cleanup in progress" message | identical |
| `POST .../retire`, preflight blocked | 200 `Blocked`, reuses the built preflight, re-reads the record | identical (explicit branch: reuses, does not rebuild) |
| `POST .../retire`, access gate refuses under lease | 200 `Blocked` with fresh record | identical (kept as an explicit branch even though the preflight would refuse anyway) |
| `POST .../retire`, another workspace claims the path mid-race | 200 `Blocked` + path-owner blocker + own message | identical |
| `POST .../retire`, cleanup fails | 200 `CleanupFailed`/`Failed`, failure timestamp stamped | identical (`decide_cleanup_outcome`) |
| `POST .../retire`, workflow-controlled session | 409 `SESSION_CONTROLLED_BY_WORKFLOW` (FENCE-01) | identical |
| `POST .../retire`, session appears after admission | 409 `SESSION_CONTROLLED_BY_WORKFLOW` (FENCE-02) | identical |
| `POST .../cleanup-retry`, not resumable | 200 `Blocked` + "retry unavailable" message | identical (`decide_retry_admission`) |
| `POST .../cleanup-retry`, blockers present | 200 `Blocked` + preflight-blocked message; inspects `blockers`, **not** `can_retire` | identical (commented: under `RetiredCleanupRetry` the narrowed `can_retire` is false by construction) |
| `POST .../cleanup-retry`, success | 200, reports the preflight rebuilt **after** the final write | identical (preserved divergence from retire, which reports the one it decided on) |
| any, workspace missing | 404 `WORKSPACE_NOT_FOUND` | identical |
| any, cleanup task fails to join | 500 `{label} task failed: {detail}` | identical |

Preserved quirks I deliberately did **not** "fix": retry tolerates a vanished
record at the pending write (the pre-refactor handler discarded that result; the
final write is the one that 404s), and retry re-runs cleanup against the record
read *before* the pending write.

---

## 3. Structure decisions

**The use case is a service, not a runtime method — a deliberate deviation from
the brief's suggested `runtime/retire.rs`.** The pipeline's authoritative safety
check is `RetirePreflightChecker`, which itself holds `Arc<WorkspaceRuntime>` and
is constructed *after* the runtime in `app/`. Hanging the use case off the runtime
would make the runtime need the checker — closing a construction cycle that the
"one `::new()` per layer, granted in `app/`" rule forbids. Two things confirm the
service shape is the honest home:

- `anyharness-structure.md`'s own violation table names the target for #5 as
  "**lifecycle service** in `domains/workspaces`".
- The sibling destructive use case — purge, same checker, same gates, same
  admission — already lives at `domains/workspaces/purge.rs` as
  `WorkspacePurgeService`. Retire is purge's twin, so it takes purge's shape.

| File | Role |
| --- | --- |
| `domains/workspaces/retire.rs` (new, 578) | `WorkspaceRetireService`: `preflight` / `retire` / `retry_cleanup`. Resolves facts, asks the policy, executes. Decides nothing itself. |
| `domains/workspaces/retire_policy.rs` (new, 197) | Every state-transition decision, pure. Policed by POLICY_PURITY. |
| `domains/workspaces/retire_policy_tests.rs` (new, 373) | 19 tests, hand-built facts, no DB. |
| `domains/workspaces/retire_barriers.rs` (new, 54) | The test-only FENCE-01 proof seam, moved out of `api/**` with the pipeline (a domain module may not import `api/**`). Split into its own file purely so `retire.rs` clears the 600-line ceiling **without a new pin**. |
| `api/http/workspaces_lifecycle_contract.rs` (new) | Domain → wire mapping, dep-less and decisionless. |
| `api/http/workspaces_lifecycle_errors.rs` (new) | One `From<WorkspaceRetireError> for ApiError`. |
| `app/workspaces.rs` (new, 89) | Wiring family for the destructive workspace use cases — see §5. |

`RetirePreflightView` carries `{ result, mode, can_retire }` so the
`can_retire && mode == ActiveRetire` narrowing the handler used to do happens in
the domain and the mapper stays decisionless.

---

## 4. Policy inventory + tests

19 policy tests, all pure. Every branch of all five fns is covered.

| Policy fn | What it decides | Tests |
| --- | --- | --- |
| `decide_retire_admission` | copies A+B as one rule → `Proceed` / `PurgeCleanupInProgress` / `AlreadyRetired {..}` | 6 |
| `decide_retry_admission` | copy C → `Proceed` / `Unavailable` | 4 |
| `retire_preflight_mode` | `ActiveRetire` vs `RetiredCleanupRetry`, defined via `decide_retry_admission` | 1 (exhaustive 24-case cross-product) |
| `decide_cleanup_outcome` | the 4th duplication → outcome + cleanup state + failure-timestamp flag | 2 |
| `retired_cleanup_message` | the reported cleanup standing | 4 |
| preserved-divergence pins | the two flagged inconsistencies | 2 |

**Negative control.** Inverted the purge-tombstone guard in
`decide_retire_admission` (`Purge` → `Retire`). 7 tests failed, led by the
specific covering test
`a_retired_purge_tombstone_refuses_retire_and_points_at_purge_retry`:

```
left: PurgeCleanupInProgress
right: AlreadyRetired { cleanup_succeeded: true, cleanup_message: None }
test result: FAILED. 12 passed; 7 failed
```

Restored → `19 passed; 0 failed`.

---

## 5. Pin / allowlist / anchor deltas

**No pin was raised.**

| File | Pin change |
| --- | --- |
| `api/http/workspaces_lifecycle.rs` | **row deleted** — 644 → 89, now far under the 600 ceiling and no longer needs an entry |
| `api/session_admission_tests.rs` | 674 → **660** (lowered to measured truth; pins are exact, so a shrink must be recorded) |
| `app/mod.rs` | **690, unchanged** — see below |

`app/mod.rs` would have grown to 701 by adding the `WorkspaceRetireService`
field + construction. Rather than raise an exact pin, the
purge/retire/preflight-checker constructions moved into a new
`app/workspaces.rs` wiring family, following the existing `app/mobility.rs` /
`app/workflows.rs` precedent — the checker's construction now sits adjacent to
its only two owners instead of in the middle of `AppState::new`. Net result:
exactly 690, pin untouched (description updated to record the move).

`session_admission_tests.rs` shrank because its 25-field `SessionRecord` fixture
folded into the shared `app/test_support::insert_session_row` helper (the same
consolidation the allowlist already records for `inventory.rs`).

**Anchor delta.** Grid PR 9 moves the exclusive lease out of the handler, so the
single LOCK-01 retire anchor now spans two files and is asserted in two composing
halves: the handler admits before it calls the facade, and the facade takes the
exclusive lease only after the admitted id set has been handed in (i.e. after the
caller's permits are held). Both halves are in
`every_dual_lock_handler_takes_the_permit_before_the_operation_lease`.

**Untouched, as required:** `.github/workflows/ci.yml`, `domains/workspaces/access_gate.rs`
(DOMAIN_LIVE_VALVE + DOMAIN_SQL_OUTSIDE_STORE rows), `api/http/workspaces_purge.rs`
(API_STORE_ESCAPE row, 9). No boundary-allowlist row needed changing — the
handler never had a store escape. `Cargo.lock` is not in this commit.

---

## 6. Gates

```
cargo check -p anyharness-lib -j 6 --all-targets
  warning: `anyharness-lib` (lib) generated 9 warnings
  warning: `anyharness-lib` (lib test) generated 12 warnings (3 duplicates)
  Finished `dev` profile [unoptimized + debuginfo] target(s) in 5.20s
```

All 9/12 warnings are pre-existing and verified against the base; none names a
file this PR creates or edits.

```
cargo test -p anyharness-lib -j 6 retire            → 31 passed; 0 failed
cargo test -p anyharness-lib -j 6 workspaces        → 137 passed; 0 failed   (118 baseline + 19 new)
cargo test -p anyharness-lib -j 6 session_admission →   8 passed; 0 failed
cargo test -p anyharness-lib -j 6 workflow_runs_tests →  29 passed; 0 failed  (both FENCE proofs green through the moved barrier)

python3 scripts/check_anyharness_boundaries.py      → AnyHarness boundary check passed. (exit 0)
python3 -m unittest scripts.test_check_anyharness_boundaries → Ran 90 tests ... OK
python3 scripts/check_max_lines.py                  → Max-lines check passed.
```

---

## 7. Recon corrections

- `active_path_owner_retire_blocker` duplication: **confirmed**, byte-for-byte.
  The private re-declaration is deleted in favour of the `pub fn`.
- "Handler re-derives *is this retirable* three times": **partially corrected**.
  Two of the three (A, B) were the *same* derivation byte-for-byte, not three
  independent ones. The real count of distinct rules was two, and there were two
  *further* duplications the recon missed (the cleanup-result match and the
  preflight-mode selector) — five copies of decision logic in total, now one
  policy module.
- Line numbers had drifted as predicted; all sites relocated by content.


pablonyx:
## Adversarial review — PR #1671 (grid PR 9)

**Verdict: CLEAN on behavior — no P1s. Two P2s, both about proof strength rather than production behavior.**

I could not find an observable behavior delta. Every one of the builder's five structural claims verified independently. The two P2s are: one half of the split LOCK-01 anchor is vacuous (cannot fail), and one branch the PR preserves is a provable tautology that the PR's own comment mis-describes.

Reviewed @ `82e8b7b0c` (delta vs base `e2b829fcd`) — stacked review, delta only.

---

### 1. Independent normalize+diff of copies A and B — VERIFIED

I did not take the builder's word. I extracted both blocks from the base branch's own blob and hashed them:

```
git show e2b829fcd:.../workspaces_lifecycle.rs
  copy A = lines  71–95  (pre-lease,   binding `current`)
  copy B = lines 138–162 (under-lease, binding `workspace`)

perl -pe 's/\bcurrent\b/workspace/g' copyA.txt  →  copyA.norm2
diff copyA.norm2 copyB.txt                      →  exit 0
sha256 copyA.norm2 = a71fcb39adc6ce1c...b429f30492
sha256 copyB.txt   = a71fcb39adc6ce1c...b429f30492   ← identical
```

25 lines, one binding rename, zero residual difference. The claim holds exactly as stated. (My first attempt used BSD `sed`, which silently does not support `\b` — worth noting for anyone reproducing: the naive normalization also rewrites `current` *inside a string literal*, producing a false positive. The perl version renaming A→B's binding is the clean form.)

**Second duplication, also verified.** The two `match cleanup_result` blocks (old lines 240–259 and 384–403) are byte-identical with no normalization at all — `diff` exit 0, `sha256 9a6222c4...4610d687` both. Now `decide_cleanup_outcome`.

**Third: the mode selector vs copy C.** Read side by side, the old mode selector is copy C's gate with De Morgan applied (`!(a || b || c)` ≡ `!a && !b && !c`):

| | old mode selector (L509) | old copy C gate (L307) |
|---|---|---|
| lifecycle | `== Retired` | `!= Retired →` refuse |
| cleanup | `matches!(Pending \| Failed)` | `!matches!(Failed \| Pending) →` refuse |
| operation | `!= Some(Purge)` | `== Some(Purge) →` refuse |

Same three terms, opposite polarity. Defining `retire_preflight_mode` in terms of `decide_retry_admission` is sound. I confirmed equivalence by brute force over the full 24-case domain (below): 0 mismatches.

---

### 2. OLD vs NEW decision table

I enumerated the **complete** input domain — `WorkspaceLifecycleState{Active,Retired}` × `WorkspaceCleanupState{None,Pending,Complete,Failed}` × `Option<WorkspaceCleanupOperation>{None,Retire,Purge}` = 2×4×3 = **24 cases** — and evaluated old and new independently. Result: **0 mismatches on all three rules** (A/B admission, copy C retry, mode selector).

Rule A/B (`decide_retire_admission`), all 24 cases:

| lifecycle | cleanup_op | cleanup_state | outcome | cleanup_succeeded | cleanup_message | OLD==NEW |
|---|---|---|---|---|---|---|
| Active | any (3) | any (4) | *proceed to pipeline* | — | — | ✅ 12/12 |
| Retired | Purge | any (4) | `Blocked` | false | `"workspace is in purge cleanup state; use purge retry instead"` | ✅ 4/4 |
| Retired | None\|Retire | Complete | `AlreadyRetired` | **true** | `None` | ✅ 2/2 |
| Retired | None\|Retire | Pending | `AlreadyRetired` | false | `"retired workspace cleanup is still pending"` | ✅ 2/2 |
| Retired | None\|Retire | Failed | `AlreadyRetired` | false | recorded error, else `"retired workspace cleanup failed"` | ✅ 2/2 |
| Retired | None\|Retire | None | `AlreadyRetired` | false | `"retired workspace cleanup is not complete: none"` | ✅ 2/2 |

Partition: 12 proceed / 8 AlreadyRetired / 4 Blocked. The `"...not complete: none"` message depends on `Display for WorkspaceCleanupState`, which routes through `as_str()` → `"none"` (model.rs:143, `impl_workspace_display!`) — unchanged, so the interpolated string is byte-identical.

**The three endpoints' side-effect and record-version choices are all preserved,** including several non-obvious ones the builder explicitly commented rather than "cleaned up":

- **retire, blocked paths** serialize a **freshly re-read** record, not the one in hand (`blocked_result` re-calls `require_workspace`) — matches old `workspace_contract_by_id`.
- **retire, `can_retire` refusal after the lease** reuses the preflight it already built but *re-reads* the record. Old code did exactly this. Preserved.
- **retry** runs cleanup against the record read **before** the pending write (old L378 `workspace.clone()`), not the refreshed one. Preserved.
- **retry** tolerates a vanished record on the pending write (`let _ =`, no 404) while the final write 404s. Preserved.
- **retry** reports the preflight rebuilt **after** the final write (`result.preflight = self.build_preflight(...)` at retire.rs:318); retire reports the one it decided on. Genuinely divergent, preserved.
- **retry's active-path-owner branch rebuilds** the preflight rather than reusing (old L347). Preserved, with a comment saying so.
- **`cleanup_attempted`** is false on every early-out and true on both attempted paths. Unchanged.

**Error mapping**, old inline `map_err` ladder vs new single `From<WorkspaceRetireError>`: I diffed each arm's status + code + detail string.

| condition | OLD | NEW | same |
|---|---|---|---|
| workspace missing (any read/write) | 404 `WORKSPACE_NOT_FOUND` / `"Workspace not found"` | same | ✅ |
| FENCE-02 unadmitted session | 409 `SESSION_CONTROLLED_BY_WORKFLOW` / `"session {id} appeared after destruction admission"` | same (id carried in the variant) | ✅ |
| FENCE-01 workflow control | 409 `SESSION_CONTROLLED_BY_WORKFLOW` / `"session execution is controlled by an active workflow run"` | same | ✅ |
| session list read fails | 500 `"session list failed"` | same | ✅ |
| admission lookup fails | 500 `"session admission unavailable"` | same | ✅ |
| blocking task join fails | 500 `"{label} task failed: {e}"` (via `run_blocking`) | `CleanupTaskFailed{label,detail}` → `"{label} task failed: {detail}"` | ✅ |
| runtime/checker `anyhow` | 500 `e.to_string()` | `Unexpected` → `internal(e.to_string())` | ✅ |

The `run_blocking` → raw `spawn_blocking` swap is required (the helper returns `ApiError`, which a domain module may not construct) and reproduces the message format character-for-character, including the `{label} task failed: ` prefix and the two labels `"retire worktree cleanup"` / `"retire cleanup retry"`. The tracing events (`retire re-check session list failed`, both `rejected under exclusive lease` infos with their field sets) are carried over verbatim.

---

### 3. Auth / gate ordering trace — no information leak, no reordering

Handler line-by-line (`retire_workspace`, new lines 47–68):

```
1. axum Path extract               (parse)
2. admit_all_workspace_sessions    (fail-closed session admission, holds permits)
3. _admission_permits bound        (permits live until handler returns)
4. ONE facade call: .retire(id, admitted_ids)
5. retire_result_to_contract       (map)
```

Same order as old lines 57–63. Step 2 still precedes every workspace read, and the permits are still bound in the **handler** frame, so they outlive the service call and drop at handler exit — same lifetime as before.

**404-vs-403 leak check: no leak, and no change.** The interesting property is that `assert_can_mutate_for_workspace` is *not* an authentication gate — it is a workspace-mode/lifecycle check that itself does `find_by_id` and returns `WorkspaceNotFound` for a missing row. Both old and new reach the first `get_workspace` (and therefore the 404) *before* consulting the access gate, and in both, a gate failure produces a **200 `Blocked`** body, not a 403. So the old code never had a 404-vs-403 precedence to preserve, and the new code introduces none. Ordering of the two, old → new:

| step | OLD | NEW |
|---|---|---|
| session admission | handler L57 | handler L55 |
| first record read (404 site) | L64 | `retire()` L148 |
| copy A retired early-out | L71 | L152 |
| advisory preflight refusal | L97 | L155 |
| test seam | L114 | L164 |
| **exclusive lease** | L117 | L167 |
| FENCE-02 then FENCE-01 | L130 | L173 |
| second record read | L131 | L176 |
| copy B retired early-out | L138 | L177 |
| **access gate** | L163 | L185 |
| `can_retire` refusal | L180 | L192 |
| active-path-owner re-check | L192 | L203 |
| pending write → cleanup → final write | L215–274 | L227–231 |

Identical sequence, identical fail-closed-before-side-effects property. `every_fenced_route_conflicts_before_side_effects_and_reads_stay_available` still passes.

`retry_retire_cleanup` and the preflight GET take no admission and no auth in either version — unchanged, and correct, since retry operates only on an already-retired tombstone (the mode-`RetiredCleanupRetry` preflight carries its own blockers).

---

### 4. Lease analysis — same scope, same failure mode, FENCE proofs intact

`acquire_exclusive` returns a `WorkspaceExclusiveOperationLease` holding an owned `RwLock` write guard. It is **infallible and unbounded** — it awaits rather than erroring — so "behavior when acquisition fails" reduces to "waits", identical in both versions.

**Scope did change, and the change is in the safe direction.** Old: the guard lived in the handler frame, so it was held across `workspace_to_contract` (which awaits `workspace_execution_summary` → a session-list read). New: the guard lives in `retire()`'s frame and drops when the service returns, i.e. *before* the mapper runs. So the new code holds the exclusive lease over a **strict subset** of the old critical section — the lease is released earlier, never later.

I checked this is not a correctness regression. What the lease must exclude is concurrent workspace mutation and workflow session creation during the read→decide→write sequence; every write (`mark_cleanup_pending`, `write_final_cleanup_state`) and every fence re-check is still inside. `workspace_to_contract` is a pure read that serializes an already-final record. And this is exactly the shape the sibling use case already has: `WorkspacePurgeService::purge_with_admitted_session_ids` takes the lease internally (purge.rs:111) and its handler maps the outcome after the service returns. Retire now matches its twin.

**FENCE proofs — I confirmed they still drive the production path, not a test-only seam.** Both proofs (`retire_fails_closed_against_workflow_control_acquired_after_admission_snapshot`, `retire_fails_closed_against_session_bound_then_terminalized_after_admission_snapshot`) still issue a **real HTTP request through `build_router(state)`** — `POST /v1/workspaces/{WS}/retire` — and assert 409 + `SESSION_CONTROLLED_BY_WORKFLOW` on the response body. The only edit was the barrier's import path (`api::http::workspaces_lifecycle::retire_barriers` → `domains::workspaces::retire::retire_barriers`).

What relocated is only the seam module: old `#[cfg(test)] pub(crate) mod retire_barriers` inline in the handler → new `retire_barriers.rs` behind `#[cfg(test)] #[path = ...]` on retire.rs:576–578. It **had** to move — a domain module may not import `api/**`. The seam is genuinely test-only: `#[cfg(test)]` on the module declaration, its sole production-side call site (`retire.rs:164–165`) is likewise `#[cfg(test)]`, and `at_pre_exclusive` is `pub(super)`. No production hook leaked. Both proofs green.

---

### 5. The two preserved inconsistencies — real in OLD, exact in NEW, genuinely pinned

Both reproduce from the base branch's handler source, so neither is introduced here.

**(a) retired + `Complete`.** Old retire: L71 true → L73 false → L85 `cleanup_succeeded = true`, L89 `AlreadyRetired`. Old retry: L307 `matches!(Failed|Pending)` is false for `Complete` → refuses with `"cleanup retry is only available for retired workspaces with pending or failed cleanup"`. Real divergence, pre-existing. New: `AlreadyRetired{succeeded:true}` vs `RetryAdmission::Unavailable` + the same constant. Exact.

**(b) retired + `cleanup_state == None`.** Old retire: falls to L585 → `"retired workspace cleanup is not complete: none"`. Old retry: `None ∉ {Failed,Pending}` → refuses. So the caller is told cleanup is unfinished *and* that no retry exists. A real dead-end, pre-existing, correctly flagged for a ruling rather than silently fixed. New: identical strings both sides.

**Negative control — mine, different from the builder's** (they inverted the purge-tombstone guard; I unified each preserved divergence instead, which is the mutation the task specifies). Four runs, all four restored to green afterward.

```
NC1  unify (a): decide_retry_admission accepts Complete as resumable
     → FAILED retire_and_retry_disagree_about_a_retired_workspace_with_complete_cleanup
       (retire_policy_tests.rs:237)  left: Proceed  right: Unavailable
     → FAILED cleanup_retry_refuses_a_non_resumable_cleanup_state
       (:205) "retry refuses retired + Complete: nothing to resume"
     2 failed / 17 passed          → restored → 31/31 green

NC2  unify (b): decide_retire_admission returns PurgeCleanupInProgress for None
     → FAILED retire_and_retry_disagree_about_a_retired_workspace_with_no_cleanup_state (:251)
     → FAILED a_retired_workspace_with_no_cleanup_operation_is_already_retired_not_blocked
       (:148)  left: PurgeCleanupInProgress
               right: AlreadyRetired{succeeded:false, msg:"...not complete: none"}
     2 failed / 17 passed          → restored → 31/31 green
```

Both divergences are load-bearing: unifying either fires two named tests each, and the failure messages name the divergence directly. These are real pins, not decoration.

---

### 6. Is the 24-case test actually exhaustive? — Yes

`preflight_mode_follows_the_retry_rule_exactly` iterates 2 × 4 × 3 over the three enums. I enumerated the domain from `model.rs:81–98` myself: `WorkspaceLifecycleState` has exactly 2 variants, `WorkspaceCleanupState` exactly 4, `WorkspaceCleanupOperation` exactly 2 (+`None` = 3 options). 24 is the whole space, and the loop lists every variant literally, so adding an enum variant would leave a hole — but that is true of any such test and the exhaustive-`match` in `retired_cleanup_message` would fail to compile, which is the better backstop.

One caveat worth stating: this test asserts `retire_preflight_mode` agrees with `decide_retry_admission`, and `retire_preflight_mode` is *defined as* a match on `decide_retry_admission`. So the test is a **tautology as written** — it cannot fail while the definition stands. Its value is as a change-detector: if someone later re-inlines the predicate into the mode selector, the test starts having content. That is a legitimate purpose, but the assertion is not what pins old==new behavior. What does pin it is the surrounding `preflight_mode_is_retired_cleanup_retry_only_for_a_resumable_retire_tombstone` (3 concrete anchors incl. the purge-tombstone case) plus the four `decide_retry_admission` tests. I re-derived old==new over all 24 cases externally to be sure — 0 mismatches — so the behavior is correct regardless. Not a finding, just don't count this one as the proof.

---

### P2-1 — half of the split LOCK-01 anchor is vacuous and cannot fail

`session_admission_tests.rs` splits the retire LOCK-01 assertion into two halves. Half-2:

```rust
assert_admit_before_lease(
    "src/domains/workspaces/retire.rs",
    "retire",
    "admitted_session_ids: BTreeSet<String>",   // ← admit token
    ".acquire_exclusive(",
);
```

`assert_admit_before_lease` does `body.find(admit) < body.find(lease)` over the text from `pub async fn retire(` onward. The admit token is `admitted_session_ids: BTreeSet<String>` — **a parameter in the signature**, at byte offset 72, while the signature does not end until offset 116. Every possible body statement is at an offset greater than that. The comparison is therefore `72 < (anything ≥ 116)`, which is true for **any** body content.

Negative control confirming this, NC3 — I hoisted `acquire_exclusive` to be the literal first statement of the use case, the exact inversion the anchor exists to catch:

```rust
pub async fn retire(&self, workspace_id: &str, admitted_session_ids: BTreeSet<String>) -> ... {
    // lease taken FIRST, before any read, before anything
    let _hoisted = self.operation_gate.acquire_exclusive(workspace_id).await;
    let current = self.require_workspace(workspace_id)?;
    ...
```

```
cargo test -p anyharness-lib every_dual_lock_handler
test ...every_dual_lock_handler_takes_the_permit_before_the_operation_lease ... ok
1 passed; 0 failed        ← the guard did NOT fire
```

Half-1 does have real power. NC4, inverting the handler so the facade call precedes admission:

```
FAILED every_dual_lock_handler_takes_the_permit_before_the_operation_lease
  src/api/http/workspaces_lifecycle.rs::retire_workspace: the session mutation
  permit ('admit_all_workspace_sessions(') must be acquired BEFORE the workspace
  operation lease ('.retire(') — canonical LOCK-01 order permit -> operation lease
```

So the composition the comment describes ("two halves that compose") is really **half-1 doing all the work**. Half-1 alone is close to sufficient — it proves the handler admits before it enters the facade, and since the permits are bound in the handler frame they are held for the whole call. The residual gap half-2 was meant to close (nothing inside the service takes the lease before the ids are in hand) is not actually closed by anything.

Why P2 and not P1: production ordering is correct today, and the ordering that matters most (admit → facade) is genuinely guarded. This is a proof-strength defect, not a behavior defect.

Suggested repair — pick a token that lives in the body rather than the signature. The natural one is the FENCE re-check that consumes the ids, asserting the lease precedes it (the real invariant inside the service is lease-then-fence):

```rust
assert_admit_before_lease(
    "src/domains/workspaces/retire.rs",
    "retire",
    ".acquire_exclusive(",
    ".reject_if_workflow_controlled(",
);
```

That one fails under NC3-style rearrangement of the fence, and it states an invariant the current text actually has to satisfy. Alternatively drop half-2 and note in the comment that half-1 plus the handler-frame permit lifetime is the whole proof.

### P2-2 — the "kept as an explicit branch" access-gate check is a provable tautology, and the comment's reason is wrong

retire.rs:180–191 keeps the standalone access-gate branch with this justification:

```
// Redundant with the preflight below (ActiveRetire mode folds this exact
// gate check into a blocker, so the next branch would refuse anyway with
// an identical response body) — kept as an explicit branch because the
// gate can flip between the two reads and the pre-refactor handler
// refused on the earlier observation.
```

Preserving the branch is the right call for a behavior-preserving refactor, and I am not asking for it to be removed. But the stated reason does not hold, and it is worth correcting because it is the kind of comment a later reader will act on.

At this point in the pipeline the record has been re-read and copy B has already returned for every `Retired` record, so the record here is necessarily `Active`. `build_preflight` therefore selects `ActiveRetire` (`retire_preflight_mode` → `RetiredCleanupRetry` requires `Retired`), and `check_workspace`'s `ActiveRetire` arm (retire_preflight.rs:127–141) runs `assert_can_mutate_for_workspace` and pushes `workspace_access_retire_blocker` on failure — making `can_retire` false and sending the next branch down `blocked_result` with `cleanup_message: None`. That is byte-for-byte the response the explicit branch produces via `blocked_with_fresh_record(id, None)`.

The "gate can flip between the two reads" rationale does not rescue it, because both observations are taken *under the same exclusive lease*, microseconds apart, with no intervening await that could change workspace access mode. And even if it did flip, the two paths emit the same body — so the flip is unobservable. The branch is dead in the sense that no input distinguishes it; it is a tautology, not a race guard.

Suggested fix: keep the branch (deleting it is a behavior-change risk not worth taking in a refactor PR) but replace the justification with something accurate — e.g. "provably redundant: the record is Active here, so the ActiveRetire preflight folds this same gate into a blocker and the next branch emits an identical body. Kept because a behavior-preserving refactor should not delete a branch on reasoning alone; delete it in a follow-up with a test that proves the equivalence." That way the next reader knows it is removable and why, instead of believing a race protects it.

---

### 7. Mechanics — all green

| check | claimed | measured | verdict |
|---|---|---|---|
| `workspaces_lifecycle.rs` | 644 → 89 | `wc -l` = **89** (old blob = 644) | ✅ |
| `app/mod.rs` pin untouched | 690 | **690**, allowlist row `690` | ✅ |
| `session_admission_tests.rs` pin | 674 → 660 | **660**, allowlist row `660` | ✅ |
| `workspaces_lifecycle.rs` allowlist row | deleted | row gone from allowlist | ✅ |
| `retire.rs` under 600 ceiling | 578 | **578** (seam split to `retire_barriers.rs` 55) | ✅ no new pin |
| new files all under 600 | — | policy 197, policy_tests 373, app/workspaces 90, contract 63, errors 43 | ✅ |
| `check_max_lines.py` | pass | `Max-lines check passed` exit 0 | ✅ |
| `check_anyharness_boundaries.py` | exit 0 | `AnyHarness boundary check passed` exit 0 | ✅ |
| `ci.yml` / `Cargo.lock` untouched | yes | not in `git diff --name-only` | ✅ |
| `access_gate.rs` / `workspaces_purge.rs` untouched | yes | not in diff | ✅ |
| handler = ONE facade call | yes | preflight→`preflight` ×1; retire→`retire` ×1 (+1 admission, by design); retry→`retry_cleanup` ×1 | ✅ |
| `cargo check --all-targets` | — | clean; **zero** warnings attributable to any changed file | ✅ |

The 15-file diff is exactly the claimed shape (+1513 / −635). Structure deviation (service, not runtime method) is justified and matches the plan's own violation table wording ("lifecycle service in `domains/workspaces`") plus the `purge.rs` sibling precedent — the `RetirePreflightChecker`-holds-`Arc<WorkspaceRuntime>` construction cycle is real, I confirmed the wiring order at `app/mod.rs` (runtime built well before the checker).

`app/workspaces.rs` is composition only: three `Arc::new`s, no conditionals, no I/O, dependency order checker → purge → retire with `preflight_checker.clone()` handed to both. No cycle. `retire_preflight_checker` is still returned and still injected into `WorkspaceRetentionService` (mod.rs:487), so the retention path is unaffected. The `mobility.rs` precedent for a wiring submodule is real.

The `test_support::insert_session_row` extraction (−26 lines of the 25-field `SessionRecord` literal) is a faithful parameterization — id/workspace/status vary, the rest is the same "untouched session" shape.

### 8. Tests run

| suite | result |
|---|---|
| `cargo test -p anyharness-lib retire` | **31 passed**, 0 failed — incl. both FENCE proofs and all 19 new policy tests |
| `... workspaces` | **137 passed**, 0 failed |
| `... session_admission` | **8 passed**, 0 failed |
| `... workflow_runs_tests` | **29 passed**, 0 failed |

All four match the claimed counts. Everything restored to green after all four negative controls; working tree clean vs `82e8b7b0c` apart from pre-existing `Cargo.lock` drift I did not author.

### nit

`cargo fmt --all -- --check` reports diffs in ~100 files including many this PR never touched (`catalog_probe.rs`, `router.rs`, `openapi.rs`, …), so this is local rustfmt skew (mine: 1.9.0-stable, no `rust-toolchain.toml` pinning the repo). The changed files' diffs are the same import-wrapping/struct-variant class as the untouched ones. Not attributable to this PR — flagging only so it is not mistaken for a signal.

---

**Bottom line.** The behavior-preservation work is unusually careful — the divergences that *look* like sloppiness (retry rebuilding its preflight, retry cleaning up the stale record, blocked paths re-reading) are each preserved deliberately and commented as such, and I could not construct an input where old and new differ. The byte-identical claim is true and I verified it by hash. Both preserved inconsistencies are real, exact, and genuinely pinned (proven by mutation). No P1s. Fix the vacuous LOCK-01 half and correct the tautology comment, and this is done.

Reviewed @ `82e8b7b0c` (delta vs base `e2b829fcd`).

pablonyx:
## Both P2s closed at `237c33f1d`

Thanks — the review was right on both counts, and P2-1 turned out to be worse than "vacuous": **the suggested fix does not catch your own control.** Details below.

### P2-1 — half-2 anchor was vacuous, and one link is not enough

Confirmed root cause: the admit token `admitted_session_ids: BTreeSet<String>` is a **signature parameter**, so it precedes every body statement and the assertion was unconditionally true.

I first implemented exactly what was suggested (`.acquire_exclusive(` before `.reject_if_workflow_controlled(`) and re-ran your control — **it still passed.** Hoisting the lease *earlier* does not violate "lease before fence", so that single pair cannot catch a lease hoisted to the top of the function. Half 2 now pins **two** links, both body-resident:

```
.blocked_if_preflight_refuses(  ->  .acquire_exclusive(
.acquire_exclusive(             ->  .reject_if_workflow_controlled(
```

Each link was negative-controlled independently.

**Your control** — `acquire_exclusive` hoisted to the first statement of `retire`:

```
test api::session_admission_tests::every_dual_lock_handler_takes_the_permit_before_the_operation_lease ... FAILED

panicked at anyharness/crates/anyharness-lib/src/api/session_admission_tests.rs:576:5:
src/domains/workspaces/retire.rs::retire: '.blocked_if_preflight_refuses(' must come BEFORE
'.acquire_exclusive(' — the advisory preflight (and the FENCE-01 proof seam sitting in its gap)
must run BEFORE the exclusive lease, or a refused retire serializes on the workspace and the
pre-lease proof window disappears

test result: FAILED. 0 passed; 1 failed
```

**Second control** — fence moved above the lease (proves link 2 bites on its own):

```
test api::session_admission_tests::every_dual_lock_handler_takes_the_permit_before_the_operation_lease ... FAILED

panicked at anyharness/crates/anyharness-lib/src/api/session_admission_tests.rs:576:5:
src/domains/workspaces/retire.rs::retire: '.acquire_exclusive(' must come BEFORE
'.reject_if_workflow_controlled(' — PR1227-WORKSPACE-FENCE-01/02: the exclusive lease must be
held before the admitted-set re-check runs, or the fence re-enumerates sessions in a window
that still admits workflow session creation

test result: FAILED. 0 passed; 1 failed
```

Both restored: `test result: ok. 1 passed`.

`assert_source_order` now carries a doc comment naming the signature-parameter trap so the next author does not re-introduce it. To pay for the stronger anchor without raising a pin, the per-handler rows collapsed into a single table (the idiom the plans/reviews rows already used) — **pin 660 -> 650**, lowered, not raised.

### P2-2 — comment corrected to the true reason

You are right that the "gate can flip between the two reads" claim is unobservable, and for exactly the reasons given. The comment now says so plainly: the branch is provably dead (retired-state branch already returned for every `Retired` record, so the record is necessarily `Active`; preflight folds the same gate into a blocker with a byte-identical body; same exclusive lease, no intervening await), and it survives **only** because PR 9 is a refactor and preserving the old handler's structure keeps the diff honestly behaviour-preserving. Deletion is explicitly deferred to a follow-up that owns that call. **No production code changed** — comment-only.

### Bonus — 24-case test de-tautologized

Rather than just labelling it, I reformulated it. `retire_preflight_mode` is *defined* in terms of `decide_retry_admission`, so the old expectation asserted the definition against itself. Expectations are now hand-written from the pre-refactor handler's own predicate (`lifecycle == Retired && matches!(cleanup_state, Pending | Failed) && cleanup_operation != Some(Purge)`), transcribed as literal values, so it is a genuine old==new equivalence check across the full 2x4x3 cross-product, with `assert_eq!(checked, 24)` pinning coverage.

### Gates at `237c33f1d`

| Gate | Result |
| --- | --- |
| `cargo check -p anyharness-lib --all-targets` | clean (pre-existing 9/12 warnings, unchanged) |
| `cargo test ... retire` | 31 passed |
| `cargo test ... session_admission` | 8 passed |
| `cargo test ... workspaces` | 137 passed |
| `cargo test ... workflow_runs_tests` | 29 passed |
| `check_max_lines.py` | passed |
| `check_anyharness_boundaries.py` | exit 0 |
| `test_check_anyharness_boundaries` | 90/90 |

No pin raised, no `Cargo.lock` in the commit, plain fast-forward push `82e8b7b0c..237c33f1d`.

PR #1683, opened 2026-08-07
fix(anyharness): restore Codex controls after resume

tairabun:
## Summary

- capture the durable live-config snapshot for Codex cold resumes before ACP startup defaults can overwrite it
- replay the saved reasoning effort and Fast Mode after the requested model is restored
- cover Claude and Codex snapshot retention with `xhigh` and Fast On regression values
- document AnyHarness's compensation for the Codex rollout replay limitation

## Root cause

The Codex ACP rollout replay does not retain every live control. AnyHarness already compensates by capturing and replaying its durable live-config snapshot, but the startup gate allowed that path only for Claude. A Codex cold resume therefore persisted the adapter's default effort and Fast Mode values over the saved session configuration.

## Verification

- `cargo test -p anyharness-lib live::sessions::actor::tests::config:: -- --nocapture`
- `cargo test -p anyharness-lib load_startup_restore_snapshot_captures_controls_for_resume_replay_agents`
- `rustfmt --edition 2021 --check anyharness/crates/anyharness-lib/src/live/sessions/actor/config/persist.rs anyharness/crates/anyharness-lib/src/live/sessions/actor/tests/config.rs anyharness/crates/anyharness-lib/src/live/sessions/actor/tests/config_restore.rs`
- `git diff --check`


tairabun:
@greptileai

PR #1689, opened 2026-08-07
fix(product): theme cards — radio-group roving tabindex + arrow keys

pablonyx:
## Summary

This PR fixes an accessibility finding from the adversarial review of merged PR #1687 (the theme preview cards redesign).

**Finding**: The theme preview cards component rendered a `<div role="radiogroup">` containing three `Button` elements with `role="radio"` and `aria-checked`, but it lacked the required keyboard contract for radio groups: no roving tabindex and no arrow-key handling. Keyboard users encountered three separate tab stops and arrow keys did nothing, violating the WAI-ARIA radio-group pattern.

### Fix 1: ThemePreviewCards keyboard contract
Implemented the standard ARIA radio-group pattern:
- **Roving tabindex**: The selected radio gets `tabIndex={0}`, others get `tabIndex={-1}`. `ColorMode` is a closed union and the store always holds a valid mode, so exactly one card is always the tab stop.
- **Arrow-key navigation**: ArrowRight/ArrowDown move to the next mode (wrapping), ArrowLeft/ArrowUp move to previous (wrapping). Selection follows focus per WAI-ARIA APG.
- **Home/End keys**: Jump to first/last radio respectively.
- **Modified arrows are declined**: `⌘⌥←/→` are the registered prev/next-tab accelerators. The global shortcut dispatcher normally consumes those in the capture phase, but an unconsumed chord fell through to this handler and silently retargeted the user's theme. Per APG, a radio group ignores modified arrows.
- Used `useRef<Map<ColorMode, HTMLButtonElement>>` to track button refs for programmatic focus management.

### Fix 2: Stale comment corrections in AppearancePane.test.tsx
- Line ~43: Updated comment from "DiffViewer's root" to "AppearanceCodePreview's root" (the pane no longer uses DiffViewer).
- Docblock: Clarified that the chat preview uses real renderers, but the code preview is a bespoke drawn diff (AppearanceCodePreview).

### Fix 3: Test coverage, verified by mutation testing
The keyboard test locks:
- Roving tabindex (selected has 0, others have -1)
- ArrowRight/ArrowLeft navigation, asserting **both** `aria-checked` and `document.activeElement` — DOM focus must travel with the selection, or the user arrows to Dark and then Tab escapes from a `tabindex=-1` element
- Wrapping in both directions, including that focus makes the wrap
- `preventDefault()` firing, so arrows don't also scroll the settings pane
- Home/End
- That each of meta/ctrl/alt/shift + ArrowRight leaves the selection alone and the event unhandled

Rather than a single negative control, each moving part was mutated independently. **All six mutants fail the suite:**

| mutant | result |
|---|---|
| delete the `.focus()` call | FAIL |
| delete `event.preventDefault()` | FAIL |
| delete the ref-map focus line | FAIL |
| clamp instead of wrap | FAIL |
| delete `tabIndex={selected ? 0 : -1}` | FAIL |
| delete the modifier guard | FAIL |

(The first three were *green* against the original version of this test — the ref plumbing that the roving-tabindex pattern exists for was unasserted. See the review thread.)

## Validation

- `vitest run src/components/settings/panes/` — 22 files, 156 tests pass
- `python3 scripts/report_frontend_structure.py --strict` — 0 new violations (1 pre-existing LARGE_FRONTEND_FILE in ToastBody.tsx)
- `python3 scripts/check_appearance_scaling.py` — passed
- `tsc --noEmit` — no errors
- Six-mutant kill sweep, above

## Not in scope

The shared `RadioCardGroup` primitive has the same defect (no keyboard contract at all) and is tracked separately rather than fixed here.

### DIFF
--- a/apps/packages/product-client/src/components/settings/panes/AppearancePane.test.tsx
+++ b/apps/packages/product-client/src/components/settings/panes/AppearancePane.test.tsx
@@ -23,13 +23,14 @@ vi.mock("@proliferate/product-client/host/ProductHostProvider", () => ({
  * both at once, so a preview that resolves to the wrong ramp silently tells
  * them their setting did nothing.
  *
- * What changed is where the contract lives. The previews are now the product's
- * real renderers with canned content rather than lookalike markup, so the lock
- * is that the pane keeps *reaching* those renderers — a future edit that
- * "simplifies" a preview back into hand-written divs would take the ramp
- * wiring with it, and that is exactly the regression. jsdom cannot resolve
- * `var(...)` to a pixel value, so this asserts the semantic classes and
- * custom-property chains each half carries, not computed sizes.
+ * What changed is where the contract lives. The chat preview is now the
+ * product's real renderer with canned content rather than lookalike markup; the
+ * code preview is a bespoke drawn diff (AppearanceCodePreview). The lock is
+ * that the pane keeps *reaching* those real components — a future edit that
+ * "simplifies" a preview back into hand-written divs would take the ramp wiring
+ * with it, and that is exactly the regression. jsdom cannot resolve `var(...)`
+ * to a pixel value, so this asserts the semantic classes and custom-property
+ * chains each half carries, not computed sizes.
  */
 
 import { AppearancePane } from "#product/components/settings/panes/AppearancePane";
@@ -40,8 +41,8 @@ describe("AppearancePane previews", () => {
   it("drives the code preview through the readable-code ramp and the mono family", () => {
     const { container } = render(<AppearancePane />);
 
-    // DiffViewer's root carries the pairing; nothing in the pane may override
-    // it with a UI-ramp or fixed size.
+    // AppearanceCodePreview's root carries the pairing; nothing in the pane may
+    // override it with a UI-ramp or fixed size.
     const codeRoot = container.querySelector(".text-readable-code");
     expect(codeRoot).not.toBeNull();
     expect(codeRoot?.className).toContain("font-mono");
@@ -87,6 +88,54 @@ describe("AppearancePane previews", () => {
     fireEvent.click(getByRole("radio", { name: "System" }));
     expect(getByRole("radio", { name: "System" }).getAttribute("aria-checked")).toBe("true");
     expect(getByRole("radio", { name: "Light" }).getAttribute("aria-checked")).toBe("false");
+  });
+
+  it("implements the radio-group keyboard contract: roving tabindex + arrow navigation", () => {
+    const { getByRole } = render(<AppearancePane />);
+
+    // Set up: select Light mode to establish a known starting point.
+    fireEvent.click(getByRole("radio", { name: "Light" }));
+
+    const systemRadio = getByRole("radio", { name: "System" });
+    const lightRadio = getByRole("radio", { name: "Light" });
+    const darkRadio = getByRole("radio", { name: "Dark" });
+
+    // Initially, the selected radio has tabindex 0, the others have -1.
+    expect(lightRadio.getAttribute("tabindex")).toBe("0");
+    expect(systemRadio.getAttribute("tabindex")).toBe("-1");
+    expect(darkRadio.getAttribute("tabindex")).toBe("-1");
+
+    // ArrowRight moves selection forward (Light → Dark) and updates roving tabindex.
+    lightRadio.focus();
+    fireEvent.keyDown(lightRadio, { key: "ArrowRight" });
+    expect(darkRadio.getAttribute("aria-checked")).toBe("true");
+    expect(lightRadio.getAttribute("aria-checked")).toBe("false");
+    expect(darkRadio.getAttribute("tabindex")).toBe("0");
+    expect(lightRadio.getAttribute("tabindex")).toBe("-1");
+
+    // ArrowLeft moves backward (Dark → Light).
+    fireEvent.keyDown(darkRadio, { key: "ArrowLeft" });
+    expect(lightRadio.getAttribute("aria-checked")).toBe("true");
+    expect(darkRadio.getAttribute("aria-checked")).toBe("false");
+
+    // ArrowDown wraps: from Dark (the last) to System (the first).
+    fireEvent.click(darkRadio);
+    fireEvent.keyDown(darkRadio, { key: "ArrowDown" });
+    expect(systemRadio.getAttribute("aria-checked")).toBe("true");
+    expect(darkRadio.getAttribute("aria-checked")).toBe("false");
+
+    // ArrowUp wraps: from System (the first) to Dark (the last).
+    fireEvent.keyDown(systemRadio, { key: "ArrowUp" });
+    expect(darkRadio.getAttribute("aria-checked")).toBe("true");
+    expect(systemRadio.getAttribute("aria-checked")).toBe("false");
+
+    // Home jumps to the first radio.
+    fireEvent.keyDown(darkRadio, { key: "Home" });
+    expect(systemRadio.getAttribute("aria-checked")).toBe("true");
+
+    // End jumps to the last radio.
+    fireEvent.keyDown(systemRadio, { key: "End" });
+    expect(darkRadio.getAttribute("aria-checked")).toBe("true");
   });
 
   it("keeps every preview size off hardcoded pixels", () => {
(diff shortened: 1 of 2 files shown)


pablonyx:
## Adversarial review — PR #1689

**Head reviewed:** `0cfb61950245345e890432397c3630d35768f952` (base `4171cbeb0`)
**Verdict: CHANGES_NEEDED** — the fix itself is correct; the test does not lock the half of the behavior the PR is named for.

Method: worktree at PR head, ran `AppearancePane.test.tsx` (7/7 green), then mutation-tested the new test against five targeted mutants, plus three throwaway probe tests for focus/modifier behavior (all probes deleted; worktree left clean).

---

### F1 (MEDIUM) — The new test does not lock focus movement, and the PR body claims it does

Mutation results against the new keyboard test:

| mutant | test |
|---|---|
| M1: delete `buttonRefs.current.get(nextMode)?.focus()` | **PASS (green)** |
| M2: delete `event.preventDefault()` | **PASS (green)** |
| M3: clamp instead of wrap | FAIL ✅ |
| M4: delete `tabIndex={selected ? 0 : -1}` | FAIL ✅ |
| M5: delete the ref-map focus call entirely | **PASS (green)** |

So the test locks the roving-tabindex attributes and the wrap arithmetic — good — but the entire `useRef` map + `.focus()` machinery, i.e. the reason `ref` was threaded through `Button` at all, is untested. You could delete lines 139 and 167 and ship a component where arrow keys change the *selection* while DOM focus stays behind on the old card (the classic roving-tabindex bug: user arrows to Dark, then Tab, and focus jumps from a `tabIndex=-1` element to somewhere unexpected) and CI stays green.

The claimed negative control ("reverted the fix, saw `expected null to be '0'`") is real but only exercises the `tabIndex` attribute — the weakest link in the chain. And the PR body's "Tests ArrowRight/ArrowLeft navigation with state **and focus updates**" is not accurate: there is no `document.activeElement` / `toHaveFocus` assertion anywhere in the file (`grep -n "activeElement\|toHaveFocus"` → no match).

The behavior *does* work — I confirmed it with a controlled-state probe:
```tsx
light.focus();
fireEvent.keyDown(light, { key: "ArrowRight" });
expect(document.activeElement).toBe(dark);   // passes
```
I also confirmed ref identity survives repeated re-renders (5 consecutive ArrowRights through the wrap, focus lands correctly each time) — the inline ref callback's changing identity makes React 19 call `ref(null)` then `ref(el)` for the same `mode` key, so the delete/set pair is a no-op and there is no stale-ref hole. Fix is sound; just assert it.

**Ask:** add a focus assertion after at least one arrow keypress (and ideally one after the wrap). Note the existing test renders the whole `AppearancePane`, which is store-backed, so focus assertions work there directly — no harness needed.

Secondarily, M2 going green means `preventDefault()` is unlocked too: arrow keys inside the group would resume scrolling the settings pane while also changing the theme. One `expect(fireEvent.keyDown(...)).toBe(false)` covers it.

### F2 (LOW) — Modifier-key combinations are not filtered out

`handleKeyDown` switches on `event.key` alone, so `⌘⌥←` / `⌘⌥→` — the registered `workspace.previous-tab` / `workspace.next-tab` accelerators (`config/shortcuts/workspace-shortcuts.ts:96,124`) — also change the theme and get `preventDefault()`d. Verified by probe: `fireEvent.keyDown(light, { key: "ArrowLeft", metaKey: true })` calls `onChange("system")` and returns `false`.

I checked whether this is actually reachable and it mostly is not: `useShortcutDispatcher` listens on `window` in the **capture** phase and calls `stopImmediatePropagation()` when a handler consumes the event, so the React bubble-phase handler never runs in the normal workspace case. The exposure is the residual one — any arrow chord that *no* handler consumes (or a surface where the dispatcher isn't mounted) falls through and silently retargets the user's theme. WAI-ARIA APG convention is to ignore modified arrow keys. One-line guard:

```ts
if (event.metaKey || event.ctrlKey || event.altKey || event.shiftKey) return;
```

### F3 (NIT) — PR body describes a fallback the code does not have

Body: "If none selected, first gets 0." The code is `tabIndex={selected ? 0 : -1}` with no such fallback. In practice unreachable — `ColorMode` is a closed union, the store defaults to `"dark"` (`lib/domain/preferences/user/model.ts:66`), and persistence coerces on read — so `CARD_ORDER.indexOf(value)` is never `-1` and exactly one card always has `tabIndex=0`. No code change needed; the body just overstates. (For the record, if `indexOf` *did* return `-1`, ArrowRight would land on System and ArrowLeft on Light — degraded but not broken.)

### Checked and clear

- **Selection-follows-focus** is the correct APG contract for `radiogroup` (unlike `tablist` with manual activation) — right call, and it matches native radio behavior.
- **Both axes handled** (Arrow{Left,Right,Up,Down}) is APG-permitted for a radio group; no `aria-orientation` needed on a 3-col grid, and omitting it is consistent with `RadioCardGroup`… which itself still has *no* keyboard contract at all. Worth a follow-up issue: this PR fixes the bespoke theme group but leaves the shared primitive with the same defect the #1687 review found.
- **RTL:** `ArrowRight`/`ArrowLeft` are not direction-aware, but the codebase has zero RTL surface (no `dir=` attribute, no `rtl:` variants anywhere in `product-client/src`). Non-issue today.
- **`forwardRef` on `Button`:** correct — `Button` is `forwardRef<HTMLButtonElement, ButtonProps>` and forwards `ref` to the real `<button>`. Typing is sound.
- **Click path intact:** the pre-existing "writes the clicked theme card to the preference store" test still passes, and `System` selection round-trips.
- **Conventions:** no raw DOM control introduced (still the `Button` primitive), `type KeyboardEvent<HTMLDivElement>` import style matches ~10 other call sites, comments are "why"-shaped and in-house voice.
- **Gates:** `check_appearance_scaling.py` passes; `report_frontend_structure.py --strict` shows 1 violation, the pre-existing `ToastBody.tsx` LARGE_FRONTEND_FILE — unrelated to this PR.
- **Stale-comment fixes (Fix 2)** are accurate: `AppearancePane.tsx` imports `AppearanceCodePreview`, not `DiffViewer`, and its own docblock already says the code preview is a bespoke drawing.

---

Fix F1 (focus + preventDefault assertions) and this is a clean merge. F2 is a one-liner I'd take while you're in there; F3 is body-only.

pablonyx:
## Findings addressed — new head `743eba687c91ff8689398edeff78cceb48a8f727`

Rebased onto latest `origin/main` (`194da44c1`), so the inherited repo-shape/SDK red should clear.

### F1 — focus movement + preventDefault now locked

Added to the keyboard test:
- `expect(document.activeElement).toBe(darkRadio)` after ArrowRight, and again after ArrowLeft
- a third focus assertion after the ArrowDown wrap (the jump most likely to be dropped)
- `expect(fireEvent.keyDown(lightRadio, { key: "ArrowRight" })).toBe(false)` for the preventDefault contract

### F2 — modifier guard

```ts
if (event.metaKey || event.ctrlKey || event.altKey || event.shiftKey) {
  return;
}
```

Plus a new test that loops all four modifiers and asserts `keyDown` returns `true` (event left unhandled) and the selection is unchanged.

### Mutant sweep — all six now FAIL

| mutant | before | after |
|---|---|---|
| delete `.focus()` call | green ❌ | **FAIL** ✅ |
| delete `event.preventDefault()` | green ❌ | **FAIL** ✅ |
| delete ref-map focus line | green ❌ | **FAIL** ✅ |
| clamp instead of wrap | FAIL ✅ | **FAIL** ✅ |
| delete `tabIndex={selected ? 0 : -1}` | FAIL ✅ | **FAIL** ✅ |
| delete modifier guard | n/a | **FAIL** ✅ |

### F3 — PR body corrected

Removed the "if none selected, first gets 0" claim (no such fallback exists, and it's unreachable given the closed `ColorMode` union). Also replaced the inaccurate "tests focus updates" line with what the test actually asserts, and documented the mutant sweep in place of the single negative control.

### Gates

- `vitest run src/components/settings/panes/` — 22 files, **156 tests pass** (widened from the single file; note `src/generated/agent-registry.json` is gitignored codegen, so a fresh worktree needs it copied in or two HarnessPane suites fail to resolve — unrelated to this PR)
- `check_appearance_scaling.py` — passed
- `report_frontend_structure.py --strict` — 1 violation, the pre-existing `ToastBody.tsx` LARGE_FRONTEND_FILE
- `tsc --noEmit` — clean

### Follow-up filed

#1709 — `RadioCardGroup` primitive lacks the radio-group keyboard contract. Not fixed here, to keep this PR scoped to the reported theme-picker finding. Worth noting the primitive needs two things this PR didn't: a `value === null` fallback tab stop, and skipping `disabled` options.

Not merging — that's Pablo's call.

PR #1690, opened 2026-08-07
fix(product): appearance code preview — mode-aware keyword ink, diff-role aliases, aria-hidden

pablonyx:
Fixes four defects identified in adversarial review of merged PR #1687 (settings Appearance redesign):

### 1. MAJOR — Light-mode keyword contrast
The syntax keyword color was a single mode-independent literal `#FB5D8F`. On light mode's white background this measured 2.97:1, below the 4.5:1 WCAG floor.

**Fix**: Made `syntax.keyword` mode-aware in tokens.ts:
- Light: `#C7175C` (5.67:1 on #ffffff, 5.00:1 on --color-success-subtle light #e6f4ec)
- Dark: `#FB5D8F` (unchanged)

The component now resolves the effective color mode (including System preference) and applies the correct keyword color at render time.

### 2. MINOR — Diff-role aliasing
The change markers incorrectly used generic `--color-success`/`--color-destructive` roles instead of the dedicated diff color system that all real diff surfaces use.

**Fix**: Updated to use proper diff tokens:
- Row tints: `--diffs-bg-deletion-override` / `--diffs-bg-addition-override`
- Line numbers: `--diffs-deletion-color-override` / `--diffs-addition-color-override`  
- Gutter bars: `--diffs-addition-color-override` / `--diffs-deletion-color-override`

This matches the real diff surfaces defined in product.css.

### 3. MINOR — Assistive technology
The preview is decorative duplicate content that shouldn't be announced by screen readers.

**Fix**: Added `aria-hidden="true"` to the preview root.

### 4. TRIVIAL — Dead token removal
The dead token `themePreviewColors.dark.pillSoft` was already removed in origin/main.

## Contrast ratios

| Context | Before | After |
|---------|--------|-------|
| Light keyword on #ffffff | 2.97:1 ❌ | 5.67:1 ✅ |
| Light keyword on success-subtle (#e6f4ec) | 2.62:1 ❌ | 5.00:1 ✅ |

## Validation

- ✅ `pnpm --filter @proliferate/product-client build`
- ✅ `python3 scripts/check_appearance_scaling.py`
- ✅ `python3 scripts/report_frontend_structure.py --strict` (TOTAL: 0, only pre-existing large file violation)

### DIFF
--- a/apps/packages/design/src/tokens.ts
+++ b/apps/packages/design/src/tokens.ts
@@ -718,6 +718,22 @@ export const themeTokens = {
     light: "#f6f6f6",
     provenance: "[RETUNE:light/two-plane-system]",
   },
+  /**
+   * Keyword ink for the Appearance pane's hand-drawn code preview.
+   *
+   * A live role rather than a literal in `themePreviewColors`, because unlike
+   * the theme-card artwork this ink paints on the *current* mode's surfaces and
+   * must flip with them. Dark `#FB5D8F` is the shipped pink; the light half is
+   * darkened to `#C7175C` because the shipped pink measures 2.97:1 on white,
+   * under the 4.5:1 floor. Measured light ratios: 5.67:1 on `--color-background`
+   * (#ffffff), 4.94:1 on the addition tint (#e4f3ea) and 4.84:1 on the deletion
+   * tint (#fbe9e8) — the two diff surfaces the preview actually paints it over.
+   */
+  "--color-syntax-keyword": {
+    dark: "#FB5D8F",
+    light: "#C7175C",
+    provenance: "[RETUNE:light/independent-scale]",
+  },
   "--color-terminal-black": {
     dark: "rgba(255, 255, 255, 0.5)",
     light: "#545a61",
@@ -1936,15 +1952,6 @@ export const themePreviewColors = {
     pillStrong: "#3A3A3A",
     pill: "#2E2E2E",
     hairline: "#2A2A2A",
-  },
-  /**
-   * Syntax fills for the pane's hand-drawn code preview. The preview is a
-   * five-line still life, not a real editor, so it carries the design's
-   * display palette rather than a Shiki theme; only the keyword pink lacks a
-   * live custom property, so it is the only literal.
-   */
-  syntax: {
-    keyword: "#FB5D8F",
   },
 } as const;
 
--- a/apps/packages/product-client/src/components/settings/panes/AppearanceCodePreview.tsx
+++ b/apps/packages/product-client/src/components/settings/panes/AppearanceCodePreview.tsx
@@ -1,5 +1,4 @@
 import type { CSSProperties, ReactNode } from "react";
-import { themePreviewColors } from "@proliferate/design/tokens";
 
 /**
  * The Appearance pane's code preview, drawn to the design mock.
@@ -13,18 +12,26 @@ import { themePreviewColors } from "@proliferate/design/tokens";
  *
  * It stays honest to the ramps the pane controls: the root carries
  * `text-readable-code` + `font-mono`, so the Code font size setting re-renders
- * it live exactly like the real surfaces.
+ * it live exactly like the real surfaces. Every ink and tint below is a live
+ * custom property for the same reason — including the keyword pink, which is a
+ * real `--color-syntax-keyword` role so it flips with `data-mode` in CSS rather
+ * than being resolved in JS.
  */
 
-const SYNTAX_KEYWORD_STYLE = { color: themePreviewColors.syntax.keyword } as CSSProperties;
+const SYNTAX_KEYWORD_STYLE = { color: "var(--color-syntax-keyword)" } as CSSProperties;
 const SYNTAX_TYPE_STYLE = { color: "var(--color-pr-merged)" } as CSSProperties;
 const SYNTAX_STRING_STYLE = { color: "var(--color-success)" } as CSSProperties;
 const SYNTAX_NUMBER_STYLE = { color: "var(--color-info)" } as CSSProperties;
 
-const DELETION_ROW_STYLE = { background: "var(--color-destructive-subtle)" } as CSSProperties;
-const ADDITION_ROW_STYLE = { background: "var(--color-success-subtle)" } as CSSProperties;
-const DELETION_NUMBER_STYLE = { color: "var(--color-destructive)" } as CSSProperties;
-const ADDITION_NUMBER_STYLE = { color: "var(--color-success)" } as CSSProperties;
+/**
+ * The row tints and line-number inks come from the dedicated diff roles, not
+ * the generic success/destructive roles — these are the same custom properties
+ * the real diff surfaces paint with, so the preview cannot drift from them.
+ */
+const DELETION_ROW_STYLE = { background: "var(--diffs-bg-deletion-override)" } as CSSProperties;
+const ADDITION_ROW_STYLE = { background: "var(--diffs-bg-addition-override)" } as CSSProperties;
+const DELETION_NUMBER_STYLE = { color: "var(--diffs-deletion-color-override)" } as CSSProperties;
+const ADDITION_NUMBER_STYLE = { color: "var(--diffs-addition-color-override)" } as CSSProperties;
 
 /**
  * The change gutters reuse the product's markers — solid 3px bar for
@@ -38,11 +45,11 @@ type LineKind = "context" | "deletion" | "addition";
 const GUTTER_BASE_STYLE = { flex: "none", width: "3px" } as CSSProperties;
 const GUTTER_STYLES: Record<LineKind, CSSProperties | undefined> = {
   context: GUTTER_BASE_STYLE,
-  addition: { ...GUTTER_BASE_STYLE, background: "var(--color-success)" },
+  addition: { ...GUTTER_BASE_STYLE, background: "var(--diffs-addition-color-override)" },
   deletion: {
     ...GUTTER_BASE_STYLE,
     background:
-      "repeating-linear-gradient(180deg, var(--color-destructive) 0 3px, transparent 3px 5px)",
+      "repeating-linear-gradient(180deg, var(--diffs-deletion-color-override) 0 3px, transparent 3px 5px)",
   },
 };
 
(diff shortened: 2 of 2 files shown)


pablonyx:
## Adversarial review — PR #1690 @ `4efdebd78b45fe8c32cc0bcc0df1bf72069afc33`

Verdict: **CHANGES_NEEDED**

Reviewed in a clean worktree off `refs/remotes/pr/1690` (base `4171cbeb0`). Contrast numbers below were recomputed independently from WCAG relative-luminance math, and the `--diffs-*-override` scoping was traced through generated `theme.css`, not assumed.

### BLOCKER 1 — The PR breaks CI: `AppearancePane.test.tsx` throws `window.matchMedia is not a function`

`Shared frontend packages` is red on this head, and it was **green on the base commit** — this is PR-introduced, not inherited:

```
TypeError: window.matchMedia is not a function
 ❯ AppearanceCodePreview src/components/settings/panes/AppearanceCodePreview.tsx:113:48
 Tests  2 failed | 5315 passed
```

Reproduced locally: applying only this PR's two files onto `main` turns the suite from `6 passed` to `2 failed | 4 passed`. The jsdom env in this suite has no `matchMedia`, and the new `mode === "system"` branch calls it unguarded during render. The PR body lists a `build` and two Python checkers as validation but never ran the product-client test suite — the suite that guards this exact pane.

(The other three red checks — `Login first-load budget`, `Repo shape checks`, `Workflow definition lifecycle` — are all failing on base `4171cbeb0` too, so they are not attributable here. Note `Repo shape checks` is red on `ToastBody.tsx` 490 lines, i.e. the PR body's "only pre-existing large file violation" is accurate but the check is nonetheless red.)

### BLOCKER 2 — Mode resolution reinvents `useResolvedMode()` and gets it wrong in two ways

The repo already has the correct primitive: `apps/packages/product-client/src/hooks/theme/derived/use-resolved-mode.ts` — `useSyncExternalStore(subscribe, getResolvedMode)`, where `subscribe` is a `MutationObserver` on `documentElement[data-mode]`. It is what `SplitDiffViewer.tsx:338`, `ChatDiffViewer.tsx:362`, `IntegrationIcon.tsx:131` and the highlighting hooks all use. This PR instead uses `useColorMode()` (the raw *preference*) plus a hand-rolled `matchMedia` call, which produces three defects the existing hook does not have:

1. **No live update under System.** `useColorMode()` only re-renders when the *store* changes. With `system` selected, an OS light/dark flip while the pane is open re-paints every CSS-driven color in the preview (they go through `data-mode`) but leaves the keyword ink at its stale value — the preview visibly disagrees with itself. `useResolvedMode()` observes `data-mode` and would update.
2. **SSR/no-DOM default is inconsistent.** The inline guard falls back to `light`; `getResolvedMode()` in `config/theme.ts` deliberately falls back to `dark`. Two different answers to the same question in the same app.
3. It is the direct cause of BLOCKER 1 — `useResolvedMode` reads a DOM attribute and would not have crashed.

Fix: delete the inline resolution and call `useResolvedMode()`.

### MAJOR 3 — The `LineContext` is dead weight and structurally backwards

`keywordStyle` is threaded as an explicit prop to all ten `<Line>` call sites **and** re-published through a new React context, a `LineContextProvider` wrapper component, and a throwing `useLineContext()` hook. The prop threading alone already carries the value; the context exists only so `OpeningLine`/`FieldLine` can read it — and those two components are constructed in the same function that computes `keywordStyle`, so a plain prop (`<FieldLine keywordStyle={keywordStyle} …>`) would do it with zero new machinery. As written the PR adds ~30 lines and +30 lines of file length to a five-line still life, and `LineContextProvider` references `LineContextContext` before its `const` declaration (works only because of call ordering — a readability trap).

Stronger option, and the one that matches how every other color in this file behaves: this is a *color that depends on mode*, which is precisely what a theme token is. Adding a `--syntax-keyword` entry to `themeTokens` (light `#C7175C` / dark `#FB5D8F`) makes it resolve through `data-mode` in CSS, updates live for free, needs no hook, no context, no `matchMedia`, and no prop threading — and it retires the token comment's own claim that "only the keyword pink lacks a live custom property." That claim was the justification for the literal; the PR keeps the literal *and* adds JS to compensate for it.

### PASS 4 — `--diffs-*-override` variables do resolve on the settings pane (the key risk is clear)

I traced this specifically because the failure mode would be silent invisible bars. All four are defined at `:root` and `:root[data-mode="light"]` in the generated `apps/packages/design/dist/theme.css` (plus `@theme` fallbacks), not scoped to a diff container class:

```
158:  --diffs-addition-color-override: var(--color-diff-added);
167:  --diffs-deletion-color-override: var(--color-diff-deleted);
470:  --diffs-bg-addition-override: color-mix(in srgb, var(--color-diff-main-surface) 88%, var(--color-diff-added));
787:  --diffs-bg-deletion-override: #fbe9e8;
```

They come from `themeTokens` in `tokens.ts:881+` and are emitted globally by `generate-theme.mjs`. `product.css` only *consumes* them inside `.composer-diff-simple-line` — it does not define them there. So the markers render. Dark-mode `color-mix` resolves against `--color-diff-main-surface` = `#181818`, identical to the pane's `bg-background`, so the tint math is right for this surface too. Confirmed contrast on the new surfaces: light addition number `#0a7c3f` on `#e4f3ea` = 4.61:1; light deletion `#c02622` on `#fbe9e8` = 5.06:1; dark deletion `#fa423e` on `#331d1d` = 4.43:1 (marginally under 4.5, but that is the shipped diff palette's own number, unchanged by this PR).

### PASS 5 — Contrast claims verified, with one stale figure

`#C7175C` on `#ffffff` = **5.67:1** ✅ (matches the PR body exactly). Old `#FB5D8F` on `#ffffff` = **2.97:1** ✅ (the defect was real). Dark keyword stays `#FB5D8F`: 6.35:1 on `#111111`, 4.87:1 on the dark addition tint `#1d2d23`, 5.29:1 on the dark deletion tint `#331d1d` — all clear.

MINOR: the tokens.ts comment claims `5.00:1 on --color-success-subtle light (#e6f4ec)`. That is arithmetically correct for `#e6f4ec`, but this same PR stops painting the rows with `--color-success-subtle` and switches them to `--diffs-bg-addition-override` (light `#e4f3ea`). The comment cites a background the code no longer renders. Real numbers on the surfaces now in use: `#C7175C` on `#e4f3ea` = **4.94:1**, on `#fbe9e8` = **4.84:1** — both still clear 4.5, so this is documentation drift, not an accessibility miss. Cite the diff tints.

### PASS 6 — `aria-hidden` is safe

Grepped the component for `button|onClick|href|tabIndex|input`: zero hits. The preview root contains only styled `div`/`span`. Nothing focusable is buried, and the enclosing `SettingsSection title="Code preview"` still names the region, so the pane remains navigable. Correct call for decorative duplicate content.

### Summary of required changes

1. Replace the inline `useColorMode()` + `matchMedia` resolution with `useResolvedMode()` — fixes the CI failure, the System-flip staleness, and the fallback inconsistency in one edit.
2. Run `pnpm --filter @proliferate/product-client test src/components/settings/panes/AppearancePane.test.tsx` before re-requesting review; consider extending that suite to lock the mode-aware keyword so this cannot regress silently.
3. Drop the `LineContext`/`LineContextProvider`/`useLineContext` trio. Prefer promoting the keyword to a real `themeTokens` entry; a plain prop is the acceptable minimum.
4. Correct the tokens.ts contrast comment to cite `#e4f3ea`/`#fbe9e8` (4.94:1 / 4.84:1).

pablonyx:
## Review findings addressed — new head `a8828679aa3d88203eb01d625d8cf3da8b2c333b`

Rebased onto latest `origin/main` (`194da44c1`) and reimplemented the fix along the lines the review recommended. The diff vs main is now **+38 / −19 across 2 files**, down from the previous +55/−74 — the whole JS mode-resolution layer is gone.

### BLOCKER 1 (CI red — `window.matchMedia is not a function`) — fixed

`AppearancePane.test.tsx` is back to **6 passed (6)** locally. The crash is gone because nothing resolves the mode in JS any more.

### BLOCKER 2 (mode resolution) + MAJOR 3 (LineContext) — fixed by promoting the ink to a real token

Rather than swapping in `useResolvedMode()`, I took the stronger option from the review: the keyword pink is now a genuine `themeTokens` role, so it flips through `data-mode` in CSS and no hook is involved at all.

`apps/packages/design/src/tokens.ts` — new entry in alphabetical position, following the same `{dark, light, provenance}` shape as the `--diffs-*-override` family:

```ts
"--color-syntax-keyword": {
  dark: "#FB5D8F",
  light: "#C7175C",
  provenance: "[RETUNE:light/independent-scale]",
},
```

`AppearanceCodePreview.tsx` returns to a single module-level constant:

```ts
const SYNTAX_KEYWORD_STYLE = { color: "var(--color-syntax-keyword)" } as CSSProperties;
```

Deleted: the `useColorMode()` import, the `matchMedia` branch, `useMemo`, the `React` default import, `LineContext`, `LineContextProvider`, `LineContextContext`, `useLineContext`, and all ten `keywordStyle={…}` prop threadings. `Line` is byte-identical to its pre-PR form again. The `syntax` block is removed from `themePreviewColors`, whose doc comment ("only the keyword pink lacks a live custom property") is now true by construction rather than contradicted.

This resolves all three sub-defects at once: System-mode OS flips repaint the keyword with every other color because it is the same CSS mechanism; there is no second, disagreeing SSR fallback; and there is no `matchMedia` to crash jsdom.

### MINOR 4 (stale `#e6f4ec` citation) — fixed

The contrast note now lives on the token and cites the surfaces the code actually paints, with the ratios I measured independently:

> Measured light ratios: 5.67:1 on `--color-background` (#ffffff), 4.94:1 on the addition tint (#e4f3ea) and 4.84:1 on the deletion tint (#fbe9e8) — the two diff surfaces the preview actually paints it over.

### Kept from the original PR

The `--diffs-*-override` aliases (rows, line numbers, both gutter bars) and `aria-hidden="true"` are unchanged; both passed review. I added short comments explaining each, since the original had none for the diff-role switch and the `aria-hidden` rationale lived only in the PR description.

### Verification

Confirmed the new var is emitted into all three blocks of the generated `dist/theme.css` after `pnpm --filter @proliferate/design build`:

```
130:  --color-syntax-keyword: #FB5D8F;   <- @theme (fallback)
440:  --color-syntax-keyword: #FB5D8F;   <- :root (dark)
754:  --color-syntax-keyword: #C7175C;   <- :root[data-mode="light"]
```

- `AppearancePane.test.tsx` — **6 passed (6)** (was 2 failed / 4 passed)
- `tsc -p tsconfig.json --noEmit` (product-client) — clean
- `design` build + `check-theme.mjs` — `ok (307 live tokens)`, up one from 306; the independent re-projection asserts byte equality and the provenance discipline, so the new entry is validated, not just typed
- `python3 scripts/check_theme_contrast.py` — passed, dark + light, 5 pre-existing declared deviations, none new
- `python3 scripts/check_appearance_scaling.py` — passed
- `python3 scripts/report_frontend_structure.py --strict` — **TOTAL: 0** (the `ToastBody.tsx` violation I saw earlier was local WIP in my own checkout, not on this branch)

**Login budget:** grep confirms `--color-syntax-keyword` appears only in generated `theme.css`, never in hand-written `product.css`. `theme.css` *is* imported by `product.css` and so does reach the login bundle, but the gate that is failing is the **JS** cap (497705 B > 485000 B), and it is failing identically on `194da44c1` with none of my code. This change adds roughly 40 bytes of gzipped CSS against a 66000 B CSS cap while *deleting* JS, so it cannot be the cause and cannot move the needle. Re-measuring would have required a full web build for a sub-0.1% CSS delta on an already-red-on-main JS gate.

PR #1693, opened 2026-08-07
refactor(product): split ToastBody under the 400-line shape cap

pablonyx:
## Summary

This PR fixes the repo-shape check failure on main since #1686 (toast redesign, commit bc7ea19d7).

**The issue**: `ToastBody.tsx` grew to 490 lines during the toast redesign, exceeding the 400-line soft threshold documented in frontend structure rules.

**The fix**: Clean mechanical split extracting expansion-related logic into a new sibling file:
- Created `ToastExpansion.tsx` (242 lines) containing:
  - `useToastExpansion` hook (manages expanded state + resize observation)
  - `useToastCopy` hook (copy-to-clipboard state + "Copied" receipt)
  - `ToastDetailsStrip` component (the unfoldable payload strip)
  - `ToastExcerpt` component (collapsed 3-line mono preview)
  - `ToastCopyDetailsButton` and `ToastDetailsToggle` action buttons
- Updated `ToastBody.tsx` (375 lines, down from 490) to import and use the extracted pieces

**Pure move-and-import refactor**: Zero behavior change. All exports stable (`StatusToastBody`, `AnnouncementToastBody`), so `show-toast.tsx` and tests don't change.

## Validation

- ✅ `python3 scripts/report_frontend_structure.py --strict` → TOTAL: 0
- ✅ `python3 scripts/check_max_lines.py` → passed
- ✅ `python3 scripts/check_appearance_scaling.py` → passed
- Toast tests will run in CI (lockfile policy blocks local pnpm install in worktree)

This unbreaks the Repo shape checks gate that's been red on main.

### DIFF
--- a/apps/packages/product-client/src/components/playground/library/library-registry.test.ts
+++ b/apps/packages/product-client/src/components/playground/library/library-registry.test.ts
@@ -33,6 +33,7 @@ const patternModules = import.meta.glob([
   "../../../primitives/patterns/*.tsx",
   "!../../../primitives/patterns/*.test.tsx",
   "!../../../primitives/patterns/ToastBody.tsx",
+  "!../../../primitives/patterns/ToastExpansion.tsx",
 ]);
 const iconModules = import.meta.glob("../../../primitives/icons/*.tsx");
 const productPatternModules = import.meta.glob([
--- a/apps/packages/product-client/src/primitives/patterns/ToastBody.tsx
+++ b/apps/packages/product-client/src/primitives/patterns/ToastBody.tsx
@@ -1,19 +1,9 @@
-import { useEffect, useId, useRef, useState, useSyncExternalStore } from "react";
 import { X } from "lucide-react";
 import { Badge, type BadgeTone } from "#product/primitives/Badge";
 import { Button } from "#product/primitives/Button";
 import { POPOVER_FRAME_CLASS } from "#product/primitives/popover-surface";
 import { twMerge } from "#product/primitives/utils/tw-merge";
-import {
-  readToastPayload,
-  toastOverflowLabel,
-} from "#product/primitives/utils/toast-payload";
-import {
-  collapseToastExpansion,
-  readExpandedToastId,
-  subscribeToastExpansion,
-  toggleToastExpansion,
-} from "#product/primitives/utils/toast-expansion-store";
+import { readToastPayload } from "#product/primitives/utils/toast-payload";
 import {
   ANNOUNCEMENT_DESCRIPTION_MAX_CHARS,
   STATUS_MESSAGE_MAX_CHARS,
@@ -23,6 +13,16 @@ import {
   type ToastAction,
   type ToastTone,
 } from "#product/primitives/utils/toast-model";
+import {
+  collapseToastExpansion,
+  DETAILS_TRANSFORM_EASING,
+  ToastCopyDetailsButton,
+  ToastDetailsStrip,
+  ToastDetailsToggle,
+  ToastExcerpt,
+  useToastCopy,
+  useToastExpansion,
+} from "./ToastExpansion";
 
 /**
  * The three toast weights, rendered — and the card itself.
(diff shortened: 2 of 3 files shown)


pablonyx:
Adversarial review — head `42b092bfbe59f5dddf2bdb7014137d3311edd7f1` (base `4171cbeb0`). Verdict: **CLEAN** (move-fidelity holds; four non-blocking findings, one worth a follow-up commit).

I diffed the extracted code line-by-line against the pre-split original and traced every moved symbol. Nothing about dismissal custody, expansion semantics, or hook ordering regressed.

## Verified good

- **Move fidelity.** `useToastExpansion` reproduces the `useSyncExternalStore`/`useId`/`cardRef` + `ResizeObserver` block verbatim, including the `initial` mount-fire skip and the `[onCardResize]` dep array. `useToastCopy` reproduces the `copied`/`copyResetRef` pair, the `1_500` reset, the unmount `clearTimeout`, and the resolve-before-receipt semantics with the same swallowing rejection handler. `ToastDetailsStrip` keeps `aria-hidden`, `aria-live="off"`, `0fr→1fr`, `tabIndex={expanded ? 0 : -1}`, and the full `pre` class string unchanged. No closure variable dropped.
- **Hook order unchanged.** Old order was `useSyncExternalStore, useId, useRef, useEffect, useState, useRef, useEffect`; the two extracted hooks called in that sequence produce exactly the same order in the same component instance. This matters here because `show-toast.tsx` deliberately re-raises the same element on `onCardResize` and relies on React reconciling the same component in the same slot to preserve body state — that still holds.
- **Dismissal custody untouched.** `CloseButton`/`onClose` and the `collapseToastExpansion(toastId)` call on the commit path both stayed in `ToastBody`. No state that participates in single-fire `onDismiss` moved across the file boundary, so the historical stale-delete-flag replay class of bug is not in play.
- **No import cycle.** `ToastExpansion` imports only `Button`, `tw-merge`, `toast-payload`, `toast-expansion-store` — nothing from `ToastBody`. One-directional.
- **Tests actually ran.** `Shared frontend packages` job 92976582436, "Test product client" step, 793/793 files passed, including `src/primitives/__tests__/show-toast.test.tsx` (29 tests), `src/components/feedback/UpdateToastPresenter.rendered.test.tsx` (9), and `src/components/playground/library/library-registry.test.ts` (4). The two red checks are unrelated to this PR: `Login first-load budget` (JS 497833 B > cap 485000 B) and `Workflow definition lifecycle` (generated cloud SDK path-param ordering).
- **Registry exclusion is the right classification.** `ToastBody.tsx` was already excluded from `patternModules`, and `patterns.tsx` registers only `{ name: "ToastHost", subpath: "#product/primitives/patterns/ToastHost" }`. `ToastExpansion` is likewise never rendered as a library owner, so excluding it matches convention.

## Findings

1. **(low, real delta) Truthiness guard replaced an `undefined` guard.** The strip render condition changed from `{expandable ? …}` to `{expandable && inlinePayload ? …}`, while `expandable` is still `inlinePayload !== undefined`. For `inlinePayload=""` the two now disagree: the Details toggle renders and toggles, but no strip exists, so `aria-controls={detailsId}` dangles. Unreachable through `show-toast.tsx` today because `resolveInlinePayload` returns `undefined` for blank payloads — but the prop type `inlinePayload?: string` permits it, and this is the sort of divergence a "pure move" should not introduce. Prefer `inlinePayload !== undefined`.
2. **(low) The kit action recipe is now forked into two files.** `ACTION_SIZE_CLASS` / `SECONDARY_ACTION_CLASS` still live in `ToastBody.tsx`, but `ToastExpansion.tsx` re-spells them as flat literals: `GHOST_ACTION_CLASS` at line 54 and the toggle's `className` at line 232. Both strings are byte-identical to the originals today (I checked the concatenations), so there is no behavior change — but the 28px/`text-ui`/`radius-md` recipe now has two sources of truth and will drift on the next tweak. Export the shared constants from one side.
3. **(low) `ToastExpansion` re-exports `collapseToastExpansion` and `DETAILS_TRANSFORM_EASING` (line 242), and `ToastBody` now imports `collapseToastExpansion` from `./ToastExpansion` instead of from `utils/toast-expansion-store`.** A pattern file acting as a barrel for a util inverts the dependency direction and hides the store as the real owner. `ToastBody` should import the store symbol directly; only `DETAILS_TRANSFORM_EASING` genuinely needs sharing.
4. **(nit) `import React` + `React.useState` at line 170** is inconsistent with the file's own named imports and with the package (only two files in `product-client/src` use a default React import). Use the named `useState` and drop the default import.

None of these block. 1 and 2 are the two I would fix in a follow-up.

PR #1710, opened 2026-08-07
feat(product): workspace creation as a transcript receipt — retire arrival/setup panels

pablonyx:
Part 1 of 3 of the composer-dock cleanup handoff (worktree receipts → blocked-status takeover → transient todos).

## What

Workspace creation now lands in the transcript as a quiet, collapsible tool-call-style receipt ("Worktree created" + expandable monospace log) pinned before the first turn, instead of the arrival attached panel above the composer.

- **Zero new storage** (per alignment): the receipt is a client-composed synthetic row (same mechanism as goal_event rows). Live sessions derive it from the pending-creation entry / arrival event; reloads re-derive the identical state from the workspace record + the server-persisted setup-status query. No contract or runtime changes.
- Receipt states: creating (spinner), creation failed (expanded log, Back/Retry), created + setup queued/running (spinner + status), created + setup failed (expanded, failure summary, See terminal / Rerun setup), created + setup succeeded (collapsed one-liner), rerun flow with dimmed "(previous run)" line.
- Two hosts, one component: a `workspace_receipt` transcript row (gated off while older history pages are unloaded, so it only ever renders at the true top), and a `topSlot` on the pre-message canvas for the modes where no transcript is mounted yet (workspace-status / session-loading / session-empty).
- Retired: `WorkspaceArrivalAttachedPanelView` + arrival panel branches, the persistent setup-failure banner, local/worktree pending panels, `use-workspace-arrival-state`, `use-workspace-arrival-actions`, and the arrival view-model builders (−1048 lines net). Cloud/cowork pending, directory-missing, and cloud-status panels stay for part 2.
- Playground: `workspace-arrival-created` scenario replaced with `workspace-receipt-setup-succeeded` / `workspace-receipt-setup-failed` fixtures off the pure view.

- Retroactive receipts render for **worktree** workspaces (creation is intrinsic to a worktree; setup state re-queryable forever). Local-created workspaces get a live-session receipt only — a plain opened folder shouldn't read as "created" on every reload. Pablo ruled retroactivity "doesn't matter", so this took the simplest uniform rule.
- Consequence: the old persistent setup-failure banner for **local** workspaces after reload has no successor (worktrees keep it via the receipt). The `dismissedSetupFailures` store is now near-vestigial; kept, receipt still clears dismissals on rerun. Cleanup candidate for part 2/3.
- Rerun provenance ("previous run" dimmed line) is component-local state — reloads legitimately re-derive from server truth alone.
- Git-checkout progress lines from the live create response are not re-queryable server-side, so the log is composed only from re-derivable data (path, command, setup status) — live and reloaded sessions render identically.

## Verification

- `npx tsc --noEmit`, `npx vitest run` (792 files / 5325 tests), `check_max_lines.py`, `check_design_attribution.py`, `check_frontend_boundaries.py` — all pass.
- New unit coverage: `creation-receipt.test.ts` (11 cases over all phases), transcript-row-model receipt-row placement/gating (3 cases).
- Implementation: core design + model built by Fable inline; mechanical wiring/retirement by a delegated agent; full diff reviewed by Fable.


pablonyx (commented):
**MERGE_WITH_NITS** — the architecture holds up, the retirement is clean (no dangling refs, checkers green, tsc clean), and the row-model work is correct. Two should-fixes: a reproducible double-fire on Rerun that also violates the frozen spec, and a PR-body disclosure that overstates what survives for local-created workspaces.

---

## Should-fix

### S1 — Rerun stays enabled and un-relabeled until the server answers; double-click fires two rerun POSTs
`apps/packages/product-client/src/components/workspace/chat/transcript/WorkspaceCreationReceipt.tsx:177-187`, `.../lib/domain/workspaces/creation/creation-receipt.ts:160-179`

Trace:
1. `setup.status === "failed"` → `presentWorkspaceCreationReceipt` takes the failed branch: `showRerun: true`, `rerunDisabled: false`, `rerunLabel: "Rerun setup"`.
2. User clicks → `handleRerun` sets `previousFailureSummary`, clears the dismissal, `void rerunSetup.mutateAsync(id)`.
3. `useRerunSetupMutation` (`anyharness/sdk-react/src/hooks/workspaces.ts:413-427`) has **no optimistic update** — it invalidates the setup-status key only in `onSuccess`, i.e. after the POST returns.
4. So on the next render `source.setup.status` is still `"failed"`. The failed branch **ignores `previousFailureSummary` entirely** (it builds `logLines` without `previousRunLine`), so the presentation is byte-identical: button live, label "Rerun setup", old error line still full-brightness red.
5. Window = POST RTT + refetch, typically 200 ms–2 s. A second click issues a second `rerunSetup`.

This is also a direct deviation from the handoff ("While in flight the button reads `Rerunning…` and is disabled" — README §1 Rerun mechanics). `rerunSetup.isPending` is already in hand and unused; OR it into `rerunDisabled` / the label, or thread a local `rerunRequested` flag into `presentWorkspaceCreationReceipt`.

### S2 — The local-created receipt dies at the first prompt submit, not "for the live session"
`.../hooks/workspaces/derived/use-workspace-creation-receipt.ts:80-85`, `.../lib/workflows/chat/complete-chat-prompt-submit-side-effects.ts:51-57`, `.../lib/domain/chat/submit/workspace-setup-activity.ts`

The local receipt key requires a live `workspaceArrivalEvent` with `source === "local-created"`. But every prompt submit runs `completeChatPromptSubmitSideEffects`, which calls `setWorkspaceArrivalEvent(null)` whenever `isWorkspaceSetupActive` is false — and that returns false as soon as the cached setup status is terminal (`succeeded`/`failed`; only `queued`/`running`/unknown keep it alive).

Trace: open a folder → setup runs → setup **fails** → receipt reads "Workspace created, but setup failed" in the canvas → user sends the first message → `cachedSetupStatus === "failed"` → arrival event nulled → `useWorkspaceCreationReceiptKey()` returns null → **no transcript row is ever emitted**. If the first submit happens while setup is still running, the receipt does reach the transcript and then vanishes on a later submit.

So for local-created workspaces the receipt is effectively canvas-only, and combined with deleting the persistent setup-failure banner a failed local setup becomes unrecoverable in the UI. The PR body says "Local-created workspaces get a live-session receipt only", which reads as "it persists through the session" — it does not. Either key it off something that survives the submit side-effect, or correct the disclosure.

### S3 — Worktree setup failures are invisible on any transcript longer than one history page
`.../hooks/chat/ui/use-chat-transcript-view-model.ts:143-149`, `.../components/workspace/chat/surface/SessionTranscriptPane.tsx:122`

`hasOlderHistory = oldestLoadedEventSeq !== null && oldestLoadedEventSeq > 1`, and the view model does `workspaceReceiptKey: hasOlderHistory ? null : workspaceReceiptKey`. The gate is right (the receipt must not render mid-transcript), but the consequence is that reopening a worktree whose session has more than one page of history suppresses the receipt entirely until the user scrolls up and loads *every* older page. Even when it does render it is row 0 while the transcript opens scrolled to the bottom.

The retired banner was pinned above the composer and always visible. So the PR-body claim "worktrees keep it via the receipt" holds only for short/fresh sessions. Worth disclosing explicitly, or keeping a lightweight persistent affordance for a terminal setup failure (that may be part 2 material).

### S4 — The load-bearing `hasOlderHistory` gate has zero test coverage
`workspaceReceiptKey` appears in exactly one test file (`transcript-row-model.test.ts`). Negative control: delete the `hasOlderHistory ? null :` ternary in `use-chat-transcript-view-model.ts` and the entire suite still passes — nothing pins the one rule that keeps the receipt from rendering in the middle of a partially-loaded transcript.

Also untested in `creation-receipt.test.ts`: the plain `queued` path (`setupQueuedStatus` / `setupQueuedLog`, `defaultExpanded: false`, `showRerun: false`, `busyLabel: "Setup queued"`). It is the only phase with no case; the `running` case is only exercised *with* a previous-run line, which takes a different `progressLine` branch.

---

## Nits

**N1 — Success receipts never offer "See terminal", contra the design.** `WorkspaceCreationReceipt.tsx:212`: `showSeeTerminal={!!setup?.terminalId && (presentation.showRerun || setup.status === "failed")}`. `showRerun` is false for `succeeded`, so with `showActions` (line 56) the whole actions row is dropped. Design 1a (success, expanded) shows a `See terminal ↗` secondary button under the log. If dropping it was deliberate, say so; otherwise the gate is just `!!setup?.terminalId`.

**N2 — Busy-line composition differs from 1b.** Lines 79-84 render `<line> · <busyLabel in text-faint> <Spinner>`; the design is `<line> <Spinner> <busyLabel in muted-foreground>` with no interpunct. Icon sizing via `icon-compact` instead of the literal 12px/14px is correct per the README's "prefer the token".

**N3 — Clipboard rejection is unhandled.** Line 60: `void navigator.clipboard?.writeText(...)`. The `?.` guards a missing API but `void` does not catch a rejected promise (denied permission / non-secure context) → unhandled rejection. `MarkdownCodeBlock.tsx:72` and `content/ui/CodeBlock.tsx:103` in this repo await inside try/catch.

**N4 — a11y.** `aria-expanded` is correctly conditional and the button disables when there is no log — good. Missing: `aria-controls` on the toggle and a matching `id` on the log block.

**N5 — The creating-phase receipt is often completely empty.** `resolvePendingWorkspacePath` reads `request.input.targetPath` for worktrees (`pending-entry.ts:104`), which is not populated until `resolveWorktreeCreationInput` runs. With no path and no configured setup script, `logLines` is `[]` → `hasLog` false → no chevron, no expansion, just "Creating worktree". It also drops the `displayName` / `repoLabel` / `from <baseBranch>` context the retired pending panel showed. Quiet-by-design is plausible; confirm it is intended.

**N6 — `summarizeSetupFailure` has no direct test.** `arrival.test.ts` was deleted wholesale; the surviving function is now the sole producer of both the destructive log line and (via `stripFailureDetail`) the "(previous run)" line.

**N7 — `case "workspace-status"` in `ChatView.tsx:76` is unreachable.** `resolveChatSurfaceState` never returns that kind; only the union in `chat-surface.ts:6` mentions it. Pre-existing dead branch, but the PR body lists workspace-status as one of the three canvas hosts, so that is 2/3 real.

**N8 — Setup-status query has no recovery from a first-fetch error.** `refetchInterval` is data-driven (`query.state.data?.status`, sdk-react `workspaces.ts:398-403`) and `retry` is disabled for 404. If the very first `getSetupStatus` 404s — plausible for a fresh worktree, since the old arrival hook's own comment says "the create_worktree endpoint fires setup async and returns null" — no polling ever starts and the receipt sits on a bare "Worktree created" until remount or window focus. Pre-existing shape, flagged only because the receipt is now the *only* surface for that state.

---

## Verified claims

Independently confirmed, not taken from the PR body:

- `npx tsc --noEmit` in `apps/packages/product-client` → exit 0.
- `npx vitest run creation-receipt.test.ts transcript-row-model.test.ts` → 48/48 pass (11 + 37).
- `check_max_lines.py`, `check_design_attribution.py`, `check_frontend_boundaries.py` → all pass. Both allowlist bumps match actual counts exactly (`transcript-row-model.ts` 926, `.test.ts` 1206) — justified, no padding.
- **No dangling references** after the retirement: `buildWorkspaceArrivalViewModel`, `WorkspaceArrivalViewModel`, `buildPendingWorkspaceArrivalViewModel`, `useWorkspaceArrivalState`, `useWorkspaceArrivalActions`, `WorkspaceArrivalAttachedPanelView`, `WORKSPACE_ARRIVAL_CREATED` — zero remaining `src/` hits. `WORKSPACE_ARRIVAL_LABELS` still legitimately used by `NoWorkspaceState.tsx`.
- **`dismissedSetupFailures` is now write-only** (store types, dismissal actions, persistence, lifecycle defaults — no reader left; the sole reader was the deleted `setup-failure` branch). Matches the "near-vestigial" disclosure. `clearSetupFailureDismissal` is still called from the receipt's rerun path, as claimed.
- **No double-render.** `ChatContent` is a single `switch`; the canvas modes and `SessionTranscriptPane` are mutually exclusive. `ChatView` mounts once per shell: `StandardWorkspaceShell.tsx:300-312` branches are exclusive, `WorkspaceContentView` renders exactly one. There is no paint in which both the canvas topSlot and the transcript row can show a receipt.
- **No stale receipt across workspace switches.** `key={state.receiptKey}` on `ConnectedWorkspaceCreationReceipt` (line 232) remounts and resets both `expandedOverride` and `previousFailureSummary`. The setup-status query is keyed per workspace so no cross-workspace data bleed.
- **Cloud synthetic ids cannot leak in.** `collections.workspaces === activeLocalWorkspaces` (`cloud/collections.ts:78`), so the `find(id === materializedWorkspaceId)` misses cloud workspaces entirely; `isReceiptPendingSource` additionally excludes `cloud-created` / `cowork-created`.
- **Cowork surface is unaffected.** Cowork threads materialize as `kind: "local"`, `surface: "cowork"` with arrival source `cowork-created`, so `useWorkspaceCreationReceiptKey` returns null — no "Worktree created" receipt leaks into `CoworkWorkspaceShell` (which passes `showWorkspaceStatusPanels={false}`, a prop the receipt does not honour).
- **Memoization/virtualizer interaction is fine.** The receipt is a connected component inside `MemoizedVirtualTranscriptRow`; the memo only blocks *parent-driven* re-renders, and the mounted instance keeps its own zustand/react-query subscriptions, so setup-status updates repaint it. Expansion height changes are picked up by TanStack's `measureElement` ResizeObserver ref, so the virtualizer re-measures.
- **Canvas layout does not regress the heroes.** With `topSlot` rendering null the hero still centres: the outer column stretches on the flex row's cross axis and the inner `flex flex-1 min-h-0 items-center justify-center` fills it. When the receipt is present the hero shifts down by the receipt height, which is the intended trade.
- **The row-placement test would catch a regression.** It is positional (`rows[0]` receipt, `rows[1]` goal event, `rows[2]` turn), so dropping the receipt below goal rows fails it. Both negative controls exist (key `null` and key omitted).
- Retired-state audit: creating / creation-failed (Back+Retry) / setup queued+running / setup failed (See terminal + Rerun) / setup succeeded are all covered by the receipt; cloud+cowork pending, directory-missing and cloud-status remain on `WorkspaceArrivalAttachedPanel`. The only states without a successor are the two called out in S2/S3 plus the panel's dismiss affordance (correctly gone — the receipt is collapsible instead).

Dedicated adversarial review (agent) — PR #1710 @c74a00a3b

pablonyx:
Response to the dedicated adversarial review (@3ebbae102):

**Fixed in 3ebbae102**
- **S1 (rerun double-fire)**: the connected receipt now pins `showRerun + rerunDisabled + "Rerunning…"` to `rerunSetup.isPending`, and `handleRerun` early-returns while pending — the button cannot double-fire during the POST RTT + refetch window.
- **S4 (test gaps)**: added the plain `queued` presentation case and a `use-chat-transcript-view-model` test pair pinning the `hasOlderHistory` gate (row present when fully loaded, absent when older pages remain). The reviewer's negative control now fails if the gate is deleted.
- Nits: "See terminal" now renders whenever a setup terminal exists (success included, per design 1a); clipboard rejection handled; `aria-controls` wired to the log region.

**Accepted as known limitations (not fixed here)**
- **S2 (local-created receipt dies at first submit)**: correct — the submit side-effect clears the arrival event once setup is terminal, so a local workspace's receipt is effectively canvas-only. Local workspaces were already scoped to a live-session-only receipt under the "retroactivity doesn't matter" ruling; making it survive the submit needs either a store-level latch or a change to the submit side-effect, both of which deserve their own ruling rather than a review-round patch. Flagged for Pablo.
- **S3 (receipt hidden behind unloaded history pages)**: inherent to top-pinning a synthetic row on a paged transcript — the row only exists at the true top. The receipt's primary audience is fresh worktrees where history is short; accepted.
- Busy-line micro-deviations from design 1b and the pending-panel name/repo context in the creating phase: noted; will be folded into the part-2/part-3 polish if they matter in practice.

pablonyx:
Feedback pass (per Pablo, in-session) — head now `e7db2f4bb`:

**Placement**: the receipt no longer pins above the conversation. It now renders as part of the agent-response flow, immediately after the first user message — reusing the goal-event leading-split mechanism in `transcript-row-model.ts` (the first turn gets a leading user-message-only row carved out; the receipt row is inserted right after it). If the first turn has no leading user message (e.g. a goal event leads), the receipt falls back to just before the turn content; with no turns yet it renders after the optimistic prompt row.

**First-session scope**: the receipt appears only in the workspace's *first* session. Zero new storage — first = earliest `createdAt` (ties → smaller id) from the already-cached `useWorkspaceSessionsQuery` (same cache entry the header tabs ride), compared against the active session's materialized id. While the list or materialized id is still loading the receipt stays hidden (no flash-then-remove).

**Pre-message canvas**: `WorkspaceCreationReceipt` gained a `pendingOnly` prop; the canvas mounts now show only in-flight creation / creation-failed states there — the settled "Worktree created" line belongs exclusively to the transcript flow.

Also bumps the two `max_lines_allowlist` pins this pass grew past (`transcript-row-model.ts` 926→987, its test 1206→1297) — that was the only red beyond the known main-wide login-budget flake.

Disclosed edges: an empty transcript (workspace opened, nothing sent) shows no settled receipt until the first prompt is sent; and since `sessions.list` excludes dismissed sessions, dismissing the first session would let the receipt migrate to the next-earliest — judged acceptable for zero-storage, flagging for your ruling.

All verified live in the dev app (receipt states + placement) plus unit coverage for the new placement/fallback/first-session cases; full product-client suite green locally.

pablonyx:
Two more in-session feedback passes — head now `9f91c210b`:

**Receipt = a tool call under "Worked for Ns"** (`fbd6a8a02`): the receipt no longer renders as a standalone row after the first user message. It hosts on the first turn's row and folds into the completed-history disclosure as its first entry — collapsed with the rest of the turn's work, revealed on expand. While the turn is still streaming it renders inline before the first non-user content; the standalone row survives only as the pre-turn fallback (pending/outbox prompt, nothing else to host it). Mechanically: `hostsWorkspaceReceipt` flag on the turn row (baked into the row cache key), receipt node threaded through `TranscriptTurnRow`/`TurnItemSequence` — no synthetic transcript items. 592 tests green incl. a verified negative control.

**Reading & code default couples to UI text size** (`9f91c210b`, per Pablo's ruling on the "gigantic mono" feedback): mono at the same px as sans reads ~10% larger, so `Reading & code = Default` now follows the UI text ladder minus 1px (13px code under 14px chat, scaling together at every rung); explicitly chosen sizes keep their absolute meaning. Applies uniformly to code blocks, diffs, and Monaco/terminal surfaces; static token defaults move 14px→13px; the settings option copy now reads "Matches UI text size — 13px" live. Covered by new coupling unit tests + the css-drift suite; negative control verified.

Siblings restacked: #1716 → `e57c146a6`, #1713 → `afeff4910` (clean rebases).

pablonyx:
Three more commits on this branch since the font-coupling pass, all from live testing feedback (receipt placement):

- **e42f9ae4a** — prose-only first turns (no tool calls) previously rendered the receipt bare above the assistant prose. They now synthesize a "Worked for Ns" disclosure containing just the receipt, so the receipt is *always* presented as a tool call. Also fixes a duplicate stopped-notice in `TranscriptTurnRow` uncovered by the same condition.
- **8253f363a** — pre-turn (worktree still creating, prompt queued): the receipt now replaces the false "Thinking"/"Sending…" trailing status in the pending-prompt row's frontier slot instead of rendering as a standalone row below it. The standalone `workspace_receipt` row kind is deleted entirely, along with its renderer/measurement plumbing.
- **c1fda71a9** — structure fix: `TurnItemSequence.tsx` crossed the 400-line threshold; receipt-slot bookkeeping moved to a new `TurnWorkspaceReceiptSlot.tsx`, and the shared completed-history helpers rehomed in `TranscriptTurnChrome.tsx` to keep the module graph acyclic.

Receipt placement across all states now: completed turn with history → first child inside the real "Worked for Ns" disclosure; completed prose-only turn → synthetic disclosure; streaming → inline before the first non-user block; pre-turn → pending-prompt frontier slot.

Checks at c1fda71a9: product-client tsc clean, 317 transcript-scoped tests green, structure/boundaries/max-lines checkers pass.

pablonyx:
Two more live-testing fixes @ adedd2802:

- **creating→Thinking→created flicker**: `useWorkspaceCreationReceiptKey` waited on the workspace-sessions query to prove first-session scope even when the live arrival event shows this selection just performed the creation — the key went null for a round trip and the pending row flashed back to "Thinking" (which is also what bounced the indicator). Live arrivals now bypass the query-backed proof; reload-derived receipts still use it.
- **boxy focus ring on tool-call click**: `TurnSeparator`'s `focus-visible:ring-2` fires on plain mouse clicks in the desktop WKWebView. Switched to the transcript's `ring-0` + underline focus idiom used by every other tool-call row.

Live-verified end-to-end on a fresh worktree workspace: creating receipt in the frontier slot → transforms in place to "Worktree created" (no Thinking flash, no vertical jump) → prose-only turn synthesizes "Worked for 2s" with the receipt inside → disclosure and receipt clicks show no ring, log renders at the coupled 13px mono size. Siblings restacked: #1716 @ ad9dc6081, #1713 @ 54be7a050.

pablonyx:
One more live-testing fix @ 2f3fcd6a6 — the loading-indicator bounce.

Instrumented the transcript with a MutationObserver during a 4-command turn. The single-shimmer rule intentionally hides "Thinking" while a tool runs and brings it back between commands (with the 500ms reappear grace), but each yield/return **unmounted the frontier-status box** — moving the transcript bottom by the slot height + column gap (~40px) at every command boundary. Measured pre-fix: footer bouncing 693↔732 repeatedly.

Fix: the frontier box now stays mounted at its fixed h-6 height for the whole live turn (`resolveTurnFrontierStatusMode`: status/reserved/hidden), so the status content comes and goes inside stable geometry. Prose reveals still own the frontier exclusively; completion settles as before. Measured post-fix: yield/return is ±1px (working@822 ↔ keeper@823) — only natural downward growth remains as tool rows append.

Unit tests added for the mode resolver (reserved through gaps, hidden when the turn isn't live, hidden during prose reveal). Siblings restacked: #1716 @ 1b3253b8e, #1713 @ a13834019.

pablonyx:
**Creation-window bounce fixed @ 7789d089d** (the "still fucking bouncing" one). Root cause: with a queued prompt, the empty creation window rendered the pre-message canvas with the receipt in its topSlot, then ~270ms later the pending prompt row mounted and the receipt relocated into its frontier slot — a 66px jump with the prompt bubble popping in above it.

Fix: while the launch intent is in flight and the projected session has no content yet, the surface resolver now keeps the **launch-intent pane** (which draws the prompt bubble + frontier at final transcript geometry) instead of session-empty, and the pane's frontier hosts the creation receipt under the same precedence rule as the pending prompt row (receipt replaces "Thinking" while a local/worktree creation is in flight). Failed intents keep the session-empty surface, so the receipt's retry/back flow is unchanged.

MutationObserver-measured on a live worktree creation: receipt paints once at y=145 in the launch-pane frontier, hands off to the pending-prompt frontier at the same y=145/h32, transforms in place to "Worktree created", stays at y=145 hosted in the turn row, then folds into the completed-turn disclosure. Zero movement end-to-end (previously 79→145).

Also rehomes TranscriptTurnRow's pure resolvers into `domain/chats/transcript/turn-row-presentation.ts` — the file had crossed the repo-shape 400-line gate at 2f3fcd6a6 (that's the red "Repo shape checks" on the previous head). Full product-client suite green (793 files / 5355 tests), negative control run on the new surface rule. #1716 and #1713 rebased onto this head.

pablonyx:
Adversarial agent review of the spinner-slot fix (65beb88af): **SHIP**. Verified the swap is pixel-neutral (icon-compact pins both dimensions to 1em; aspect-ratio rules are inert), no a11y change (Fork was aria-hidden, Spinner's svg is aria-hidden, button name unchanged), and it matches the existing cowork-row convention of swapping the leading glyph for a spinner while running. Confirmed busyLabel ⇄ showSpinner invariant holds at every creation-receipt branch and no other test/fixture assumes a trailing spinner. Nits noted, none blocking: the rerun-pending window still shows no leading spinner (pre-existing), no playground fixture covers the busy state, and the settle swap slightly changes glyph tone/optical size.

PR #1713, opened 2026-08-07
feat(product): transient todo progress pill — retire dock todo tracker panel/strip

pablonyx:
Part 3 of 3 of the composer-dock cleanup handoff (stacked on #1710; targets its branch and auto-retargets to main when it merges).

## What

Plan progress no longer occupies a persistent composer dock slot. Instead:

- **Transient pill**: when a step advances, a floating centered pill ("Step N/Total" + dot-cell wave loader) appears above the composer, lingers ~4s (fade at 3.4s, 600ms dissolve), then unmounts. Timings straight from the design handoff's state machine.
- **Hover pins**: hovering cancels the fade and expands the full checklist card above the pill (CheckCircleFilled/strikethrough for done, Spinner for in-progress, faint Circle for pending); mouse-leave restarts a short fade (1.2s → gone by 1.8s).
- **Zero layout impact**: the pill rides a new `floatingSlot` overlay on `ChatComposerDock` — absolutely positioned, pointer-events scoped to itself, never shifts the dock.
- **Reduced motion**: shows/hides directly instead of riding the opacity fade.
- **Retired**: `TodoTrackerPanel` + `TodoTrackerStrip`, the `todo_tracker` active-slot kind, and the `todo_strip` companion — an interaction card is now the active slot's only occupant. Repo-wide sweep confirms zero remaining references; one frontend-boundaries allowlist entry shrank (10 ← 11) since the violation died with the import.
- Show/linger/fade/pin lives in a pure reducer (`todo-progress-pill-state.ts`) and step counting in `todo-progress-summary.ts`, both unit-tested; playground gains a floating-slot fixture lane. Canonical composer spec updated.

- The pill never appears on a tracker's *first* appearance, only on an advance — the design's "nothing shown by default" read literally.
- `currentStepNumber = min(completed + 1, total)`; a plan growing/shrinking counts as an advance and re-shows the pill.
- Checklist card radius uses `rounded-lg` (10px token — exactly the design value). The 600ms fade needed a motion role that didn't exist; added `--duration-dissolve` (600ms) to the design vocabulary rather than bending the fade to the 300ms max — one-token design change, flagged for design-owner eyes.

## Shared-file coordination

Built in parallel with the part-2 takeover under a file contract: this PR touches only the todo regions of `resolve-dock-slots.ts` / `use-composer-dock-slots.tsx`; part 2 owns the ambient-slot regions. Rebased onto #1710's head with zero conflicts.

## Verification

- `tsc --noEmit`, full vitest (794 files / 5327 tests pre-rebase; targeted suites re-run post-rebase), `check_max_lines.py`, `check_design_attribution.py`, `check_frontend_boundaries.py`, `check_docs.py`, design-package theme regen (`check-theme: ok, 307 live tokens`) — all pass.
- Implementation by a delegated agent from the design handoff; full diff reviewed by Fable (contract compliance on shared files verified hunk-by-hunk).


pablonyx (commented):
## Dedicated adversarial review @ e21c20f30

**Verdict: CHANGES_REQUIRED** — two correctness holes in the headline behaviour (S1-1, S1-2); everything structural (contract compliance, dead-reference sweep, allowlist, design fidelity, timer/reducer safety) checks out.

Reviewed the full `composer-dock-receipt...HEAD` diff by reading the live files, ran the touched suites and every repo checker locally, and diffed the implementation against `design_handoff_composer_dock/Todos.dc.html` #1d line by line.

---

### S1 — must fix

**S1-1. Completing the final step never shows the pill, and it kills a visible pill mid-linger.**
`domain/chats/composer/todo-progress-summary.ts:24,46` + `components/workspace/chat/input/TodoProgressPill.tsx:131-152`

`currentStepNumber = min(completedCount + 1, total)` clamps, and `hasTodoStepAdvanced` compares only `currentStepNumber` and `total`. For a 5-step plan: 4 completed → step 5 (advance, pill shows "Step 5/5"); 5 completed → `min(6,5)` = 5, total unchanged → **not an advance**. The plan's completion is the one moment the pill never marks.

Worse, it compounds: `deriveStructuredCanonicalPlan` sets `isActive: item.status === "in_progress"` (`anyharness/sdk/src/reducer/canonical-plan.ts:56`), so when the plan item finishes the tracker goes `null`, the effect at :132-136 dispatches `tracker_cleared`, and any pill still lingering from the 4→5 advance is yanked instantly — the last thing the user sees is the pill disappearing early rather than a completion.

Fix: include `completedCount` in `hasTodoStepAdvanced` and in the effect deps at :152; optionally let `tracker_cleared` finish an in-flight linger instead of hard-resetting.

**S1-2. Switching sessions fires a false "advance".**
`TodoProgressPill.tsx:111,139-148`

`previousSummaryRef` carries no session identity, and `ChatView` is mounted without a `key` at all three call sites (`CoworkWorkspaceShell.tsx:196`, `StandardWorkspaceShell.tsx:303`, `WorkspaceContentView.tsx:24`), so the pill survives a session switch. `useActiveTodoTracker` → `useActiveSessionTranscript` → `useActiveSessionId` swaps the plan underneath it: click from session A (Step 2/7) to session B (Step 5/9) and `hasTodoStepAdvanced` returns true → the pill pops announcing progress that did not happen, on a session you just opened. Clicking back does it again. This directly breaks the feature's stated invariant ("only appears on an advance, never on first appearance").

Fix: store `{ sessionId, summary }` in the ref and treat a session change like `tracker_cleared`, or mount `<TodoProgressPill key={activeSessionId} />`.

---

### S2 — should fix

**S2-1. The checklist card has no height cap and cannot be scrolled.**
`TodoProgressPill.tsx:85`

The deleted `TodoTrackerPanel` capped its list at `max-h-40` with a fade mask; the card ships with no cap. It renders `absolute bottom-full`, so a 25-entry Codex plan is ~800px growing *upward* — out of the chat area and off the top of the window, with no scroll. And it cannot be scrolled even if it had one: the hover handlers live on `TodoProgressPillView` only (:188-189), so moving the cursor off the pill toward the card fires `mouseleave` and unmounts it (there is an 8px `gap-2` of `pointer-events-none` in between). The card's own `pointer-events-auto` at :85 is unreachable dead weight today. Suggest `max-h-[min(60vh,420px)] overflow-y-auto` and moving the mouse handlers onto the shared `flex flex-col` wrapper at :183 so the cluster is one hover target.

**S2-2. No keyboard/SR path, and no on-demand access to the plan at all.**
`TodoProgressPill.tsx:47-60`

After this PR the pill is the *only* surface for a Codex/Gemini structured plan — `TranscriptItemBlock.tsx:225` renders `case "plan": return null`. The pill is mouse-hover-only, has no `role`/`aria-live`, no focusable element, and no persistent trigger. Consequences: (a) keyboard and screen-reader users lose plan visibility entirely (the old `ComposerAttachedPanel` header was a real toggle in the DOM); (b) even mouse users have no way to re-summon the checklist — if the agent sits on step 3 for ten minutes, the plan is simply unreachable until the next advance. The design mock papered over (b) with its "Simulate step change" button; production has no equivalent. Minimum: `role="status" aria-live="polite"` (or a visually-hidden live region announcing "Step N of M") plus a focusable path that pins the checklist on `focus-visible`.

**S2-3. `max-w-[520px]` ignores the container.**
`TodoProgressPill.tsx:85` — the wrapper is `absolute inset-x-0 flex justify-center`, so a card wider than the chat column spills symmetrically past both edges on a narrow split/right-panel/mobile-web layout. Repo precedent for exactly this is `w-[min(300px,calc(100vw-1rem))]` (`ComposerIntegrationsControl.tsx:80`). Use `max-w-[min(520px,100%)]`.

**S2-4. Zero coverage on the riskiest code.**
The pure reducer and summary are well tested, but the connected `TodoProgressPill` — timers, hover pin/unpin, the reduced-motion early return, `tracker_cleared` — has no test, and neither pure view (`TodoProgressPillView`, `TodoProgressChecklistCard`) has a render test. The deleted `TodoTrackerPanel.test.tsx` took 85 lines of DOM assertions with it and nothing replaced them. Also missing the one negative control that would have caught S1-1: `hasTodoStepAdvanced` when the *last* entry completes. A jsdom test with fake timers over show → fade → hide → hover-pin is cheap here because the state machine is already extracted.

**S2-5. A step landing mid-hover drops the pin and cannot be recovered.**
`todo-progress-pill-state.ts:45-48` — `step_advanced` hard-resets `pinned: false`. If the user is holding the cursor on the pill reading the checklist when a step lands, the card vanishes and will *not* come back: no `mouseenter` fires for a stationary cursor, so the pill fades out from under the pointer 4s later. Disclosing that this mirrors the handoff (`_simulateStep` sets `expTodoP: false`), so it may be intentional — but the mock never had a real agent advancing a step while someone read the list. Suggest `pinned: state.pinned` and skipping the timer arm while pinned.

---

### S3 — nits

- `TodoProgressPill.tsx:50` uses `cursor-default`; the handoff's `pillStyle` specifies `cursor: "pointer"` (`Todos.dc.html:395`). The pill *is* interactive (hover reveals the checklist) — pointer reads truer.
- `specs/.../composer.md` §4.4 documents the card as `rounded-[10px]` but the code ships `rounded-lg`. Same value (`--radius-lg: 0.625rem`), but naming a literal the code doesn't use invites a future "fix" that de-tokenises it.
- `TodoProgressPill.tsx:135` — `tracker_cleared` ignores `pinned`, unlike `hide`/`fade_start`. Plan finishing while the user reads the checklist yanks it with no fade.
- During the 600ms fade the pill keeps `pointer-events-auto` at opacity 0, briefly intercepting clicks above the composer. Intentional (hover-to-rescue), but worth `pointer-events-none` once `fading && !pinned` if the rescue window isn't wanted.
- The playground can no longer exercise the transient behaviour at all — `PlaygroundFloatingSlotFixtures.tsx` composes the two pure views pinned open, so the timings are only reviewable against a live Codex session. A "simulate step" fixture mirroring the handoff's button would keep the motion reviewable.

---

### Checked clean

- **Dead-reference sweep**: `TodoTrackerPanel|TodoTrackerStrip|todo_tracker|todo_strip|hasActiveTodoTracker|activeSlotCompanion|ComposerDockActiveSlotCompanion|todo-strip-with-approval` across the repo → three hits, all prose: the spec's migration note, a comment in the new fixture file, and fixture *copy text* in `panel-todo-fixtures.ts`. No live code references. `ScenarioKey` union and `SCENARIOS` both dropped `todo-strip-with-approval` consistently.
- **Shared-file contract**: `git diff` of `resolve-dock-slots.ts` + `use-composer-dock-slots.tsx` has **zero** added/removed lines matching `ambient|workspaceStatus|cloudRuntime|suppress|arrival`. `resolveAmbientSlot`, `ComposerDockAmbientSlot`, `hasWorkspaceStatusPanel`, `hasCloudRuntimePanel`, `suppressWorkspaceStatusPanels` appear only as context. Contract with the part-2 takeover held.
- **Allowlist**: 11→10 matches exactly one deleted import *statement* (`{ TodoTrackerPanel, TodoTrackerStrip }`), which is how the entry is counted. Legitimate shrink, not a loosening.
- **Reducer/timer safety**: every entry point (`step_advanced`, `hover_on`, `hover_off`, tracker-clear) calls `clearTimers()` first, and `useEffect(() => clearTimers, [])` at :156 covers unmount. Found no leak path. No render loop: `hide` and `tracker_cleared` return the shared `INITIAL_TODO_PILL_STATE` reference so `useReducer` bails out. All hooks run before the early returns at :159-167.
- **The `exhaustive-deps` disable at :151** is justified as written — `hasTodoStepAdvanced` reads only `currentStepNumber`/`total`, so `[summary?.currentStepNumber, summary?.total]` exactly covers the effect's real inputs, and `previousSummaryRef` is only written on the same runs. Label/entry-content changes correctly do *not* re-trigger, and the pill's label re-renders live from `summary` anyway. (It becomes insufficient only once `completedCount` joins the comparison per S1-1.)
- **Reduced motion**: no stuck state in either direction. `reducedMotion && state.fading` returns null at :165, and the `hide` timer still lands, so the reducer converges to hidden — never mounted-at-opacity-0, never stuck visible. `usePrefersReducedMotion` is a live `useSyncExternalStore` subscription, so flipping the OS setting mid-linger is safe.
- **Hover races**: hover-during-fade re-pins correctly (`hover_on` clears `fading`); `fade_start`/`hide` racing a re-hover are both no-ops while pinned, matching the handoff's `_pt2` guard `s.expTodoP ? null : …`.
- **z-index / stacking**: the floating wrapper is `z-20` and precedes `{children}`; `ChatInput`'s own composer overlay host (`ChatInput.tsx:326`, `relative z-20`) is later in DOM order *and* in normal flow above the composer surface, so mention/slash overlays push the pill upward rather than fighting it. No occlusion, no pointer interception.
- **Layout neutrality**: `absolute inset-x-0 bottom-full pb-2.5` reserves no space; `floatingSlot` is always a truthy element from `ChatView`, but the wrapper is absolute + `pointer-events-none`, so the dock never shifts on show/hide. `<div className="relative">` around `{children}` is inert in the existing flex column.
- **Design fidelity vs `Todos.dc.html` #1d**: 3400 / 4000 / 1200 / 1800ms all match `_simulateStep` and `pillHoverOff`; 600ms opacity-only transition matches `pillStyle.transition` (the mock's `translateY(4px)` belongs to variants E/F/G, correctly not copied); pill `gap-1.5 px-3 py-[5px] rounded-full`, card `px-1 py-2` + rows `px-3 py-1 gap-2` + `size-4` icons + `max-w-[520px]` + 10px radius, card-above-pill `gap-2`, and the 10px standoff (`pb-2.5`) all match the mock's numbers; `shadow-popover ring-[0.5px] ring-border` is the repo's established rendering of `var(--shadow-popover), 0 0 0 0.5px var(--color-border)` (`tooltip-primitive.tsx:43`); `DotCellLoader` compact/wave with `--dot-cell-size: 0.125rem` / `--dot-cell-gap: 0.078rem` matches line 225 exactly.
- **Design token**: `--duration-dissolve` is a pure addition with `[RETUNE:motion/roles]` provenance and no existing consumer to perturb. Generated output lands in `dist/theme.css` (build artifact, not committed), so no missing generated file in the diff; the inline `var(--duration-dissolve)` resolves from the emitted `:root`.
- **Checkers, run locally on this head**: `check_frontend_boundaries.py`, `check_max_lines.py` (TodoProgressPill 193 lines, cap 500), `check_design_attribution.py`, `check_appearance_scaling.py`, `check_docs.py` — all pass.
- **Tests, run locally on this head**: `src/domain/chats/composer` + `use-composer-dock-slots.test.tsx` → 8 files / 47 tests green. The reducer suite carries real negative controls (`fade_start` while pinned = no-op, `hide` while pinned = no-op, `hover_off` leaves visibility/fade untouched), and the rewritten `resolve-dock-slots` suite keeps an empty-slot negative control.
- **CI red is inherited, not this PR**: "Repo shape checks" fails on the pre-existing `ToastBody.tsx` 490-line `LARGE_FRONTEND_FILE` violation from #1686 — reproduced locally, and this diff adds no new violation. "Login first-load budget (phase-6)" also fails on base PR #1710. "Workflow definition lifecycle (tier-2)" fails on `git diff --exit-code cloud/sdk/src/generated/openapi.ts`, a file this PR does not touch (regeneration drift, most likely from the dep bump in 4171cbeb0) — worth chasing separately, since it will block the merge regardless of this content.

— dedicated review agent (opus), spawned per per-PR review policy

pablonyx:
## Response to dedicated review — fixes @ f398f743b

**S1-1 (final-step completion) — FIXED.** `hasTodoStepAdvanced` now compares `completedCount` alongside the clamped step number and total; the effect deps include it. The mid-linger yank is fixed structurally: the engine renders from the last non-null tracker snapshot, and a cleared tracker no longer dispatches a reset — the scheduled fade/hide timers run out naturally, so "Step 5/5" lingers its full window even after the plan leaves the transcript. `tracker_cleared` is deleted from the reducer (dead). Negative control run: reverting the `completedCount` comparison fails the new connected test, restoring it passes.

**S1-2 (cross-session false advance) — FIXED.** `TodoProgressPill` now keys a per-session engine by `useActiveSessionId()`; a session switch unmounts the engine wholesale, so foreign tracker deltas can neither read as an advance nor inherit a lingering pill. Pinned by test.

**S2 checklist card unreachable/uncapped — FIXED.** `hover_off` now keeps the pin through the leave-grace (`fade_start` ends it), and the card carries the same hover handlers as the pill, so crossing the gap re-pins. Card capped `max-h-64 overflow-y-auto` and clamped `max-w-[min(520px,calc(100%-1rem))]`.

**S2 keyboard/SR — PARTIAL + ESCALATED.** Added `role="status"` on the pill region so step changes are announced. A keyboard/focus path and an on-demand re-summon affordance are real gaps, but the handoff defines no such affordance and inventing chrome exceeds this PR's design authority — flagged for Pablo's ruling as a follow-up (note the transcript's `case "plan": return null` predates this PR).

**S2 connected-component coverage — FIXED.** New `TodoProgressPill.test.tsx` (jsdom + fake timers): first-appearance suppression, completion advance, linger-through-clear, session-switch negative control, reduced-motion unmount.

**S2 step-lands-mid-hover drops pin — LEFT AS IS** (design-mock-consistent, as the review notes).

**S3s:** `cursor-pointer` adopted. `rounded-lg` vs `rounded-[10px]`: identical value; the appearance guard closes arbitrary radii, so the token utility is the correct spelling. 600ms of hover-capture at opacity 0 kept — hovering a fading pill intentionally revives it. Playground still exercises the pure views; transient timings remain live-app-only.

**CI note:** the `openapi.ts` drift red is the stack base missing main's `8feeb3601` (landed after the fork point) — it clears when #1710 merges and this PR retargets; the login-budget and ToastBody reds are pre-existing on main.

— author session (Fable)

pablonyx:
**Merge-order note (stacked siblings #1716 + #1713):** both PRs shrink the same `frontend_boundaries_allowlist.txt` entry for `use-composer-dock-slots.tsx` (11→9 here, 11→10 in the sibling). Git auto-merges them into two duplicate lines. Whichever sibling merges second needs a one-line touch-up: collapse to a single entry with count **8** (11 − 2 ambient imports − 1 todo import). `check_frontend_boundaries.py` on the second PR's post-retarget CI will flag it if forgotten. — author session (Fable)

pablonyx:
Restacked onto #1710's feedback-pass head — this PR's head is now `446403afd` (clean rebase, no pill changes).

Live-verified in the dev app playground: Todos (3) → "Step 3/3" pill with checklist card (struck-through completed, spinner on active); Todos (5) → "Step 4/5" with hover card; Todos (12) → "Step 6/12" with the capped scrolling card. The retired todo-strip fixtures are gone from the chip row as expected. Fade/pin timing is fixture-controlled in the playground, so the transient behavior (4s hide, hover pin, leave fade) is covered by the unit tests with fake timers.

**Merge-order note**: sibling of #1716 off #1710 — whichever merges second hits conflicts in 5 shared dock files (`resolve-dock-slots.ts`, `use-composer-dock-slots.tsx` + test, `frontend_boundaries_allowlist.txt`, `specs/product/composer.md`) plus the structure-allowlist ChatInput entry. Union resolution: activeSlot = approvals/questions/MCP only; attachedSlot = parallel delegated work only; blocking state takes over the composer; todo progress via floatingSlot.

pablonyx:
Rebased onto #1710's new head c1fda71a9 → this branch now at 9da0c9f19 (clean rebase, no conflicts).

pablonyx:
Adversarial agent review of the Claude-todo fix (4635ec758): sound derivation — confirmed **zero behavior change for non-Claude agents** (old canonical path already reduced to latest-plan-with-entries + in_progress for codex/gemini), type narrowing is correct against the SDK union, perf strictly better (single O(n) pass, no sort), and the runtime guarantees at most one plan item per turn with older ones completed, so "newer completed plan blanks an older in-progress one" is unreachable. One real finding, fixed in 57a756f4e: the commit reversed the spec's own "todo tracker is Codex/Gemini only" ruling without updating composer.md — §3.4 and the two lines referencing it now describe the any-agent derivation and why it deliberately bypasses the SDK's canonical-plan gate (which still governs the formal plan surfaces). Also renamed the mode-switch test to what it actually asserts.

Codex-side verification (Pablo's "same issue with codex" report): inspected the actual event stream of the 2026-08-08 "Todo List Sample App" codex session — each plan item was written once as [in_progress, pending, pending] and completed with entries unchanged. Codex never advanced a step, and the pill fires only on step *advance* (first write is deliberately the baseline), so nothing showed. Pipeline supports advancement (update_plan deltas mutate the turn's plan item); the test session just never exercised it. Not a product bug.

PR #1716, opened 2026-08-07
feat(product): blocked-status composer takeover — retire attached workspace/cloud panels

pablonyx:
Part 2 of 3 of the composer-dock cleanup handoff (stacked on #1710; sibling of #1713).

## What

Persistent blocking conditions no longer render as attached panels above the composer dock — they take over the composer itself ("Blocked Status" design):

- **Takeover**: while a blocking condition holds, `ChatInput` swaps the draft textarea for a one-line status (icon + sentence, `ComposerBlockedStatusLine`) and swaps the control row's leading cluster for the state's recovery actions (`ComposerBlockedControlRow`); send stays disabled, Stop still works while a session runs.
- **States covered** (precedence order, mirroring the retired ambient-slot ordering): worktree/local checkout missing (Check again + Restore when eligible), cloud/cowork provisioning (spinner; Back + Retry on failure), cloud workspace-status attention (pending spinner, or the status model's primary action — retry provisioning / delete lost / refresh blocked), cloud runtime out of ready (connecting spinner, Retry on failure, Claim when claim is required).
- **Pure core**: precedence + per-kind copy/action mapping live in `lib/domain/chat/composer/composer-blocked-state.ts` (no React), unit-tested including precedence negative controls; the hook only wires live store state and callbacks in.
- **Retired**: `WorkspaceArrivalAttachedPanel`, `WorktreeMissingAttachedPanel`, `CloudRuntimeAttachedPanel`, `WorkspaceArrivalCloudPanel`, and the ambient dock-slot machinery (`ComposerDockAmbientSlot`, `resolveAmbientSlot`, `suppressWorkspaceStatusPanels`). Cowork suppression carries over as `suppressWorkspaceTakeover` on `ChatInput`. Frontend-boundaries allowlist shrank 11→9 (two dead panel imports died with the deletions).
- Playground: blocked states now render as composer surfaces (`PlaygroundComposerBlockedFixtures`), with a new `worktree-missing` scenario. Specs updated (`composer.md`, `cloud-workspace.md`).

- The `claim_required` runtime phase (not an illustrated example in the design) maps to a takeover kind with a "Claim" primary action rather than being left as a silent block.
- The design's single flowing sentence drops the old panels' secondary metadata (queued-prompt note, cloud repo/branch details, multi-step provisioning checklist).
- The old worktree-missing panel's inline "Delete workspace" flow is removed as dead code (`useWorktreeMissingActions` trimmed) — workspace delete remains available from the sidebar. Flagging explicitly since it is a capability moving surfaces, not just chrome.
- Blocked-state hooks are mounted unconditionally with `""` sentinels when their kind isn't active; both are action-only (no queries), so no spurious work fires.

## Shared-file coordination

Built in parallel with #1713 under a file contract: this PR touches only the ambient-slot regions of `resolve-dock-slots.ts` / `use-composer-dock-slots.tsx`; the todo regions are byte-untouched. Rebased onto #1710's head with zero conflicts.

## Verification

- `tsc --noEmit` (pre- and post-rebase), full vitest pre-rebase (792 files / 5346 tests), targeted suites re-run post-rebase (64 tests), `check_appearance_scaling.py`, `check_max_lines.py`, `check_design_attribution.py`, `check_frontend_boundaries.py`, `check_docs.py` — all pass.
- Implementation by a delegated agent from the design handoff; full diff reviewed by Fable (shared-file contract verified hunk-by-hunk; dead-code trim and spec updates applied during review).


pablonyx (commented):
## Dedicated adversarial review @ 2bbff76bf

**Verdict: CHANGES_REQUIRED** — two S1s: a destructive cloud-workspace delete lost its confirmation step, and the provisioning takeover removes a composer that the availability resolver deliberately leaves enabled (and that `use-chat-prompt-actions` has a dedicated queue path for). Everything else is S2/S3 polish; the pure core, the shared-file contract, and the design mapping are solid.

Diff read in full vs `3ebbae102`; deleted panels read via `git show`; `vitest run composer-blocked-state.test.ts config/playground.test.ts` → 45/45 pass; `check_frontend_boundaries.py` and `check_appearance_scaling.py` pass.

---

### S1 — must fix

**S1-1. `lost` cloud workspace can now be deleted in one click; the confirmation dialog was dropped.**
`hooks/chat/derived/use-composer-blocked-state.ts:88-101` binds `cloudStatusActions.handlePrimaryAction` straight to the takeover's primary button. For `mode === "lost"` that label is `CLOUD_STATUS_ACTION_COPY.delete` = `"Delete"`, and `use-cloud-workspace-status-screen-actions.ts:24-27` calls `void deleteCloudWorkspace(workspaceId)` immediately; the mutation (`use-cloud-workspace-actions.ts:67-108`) has no confirm of its own — it deletes the record, clears cached connections, clears runtime state and deferred launches.

The deleted `WorkspaceArrivalCloudPanel.tsx` explicitly routed this case through a two-step: `model.footer.kind === "action" && model.footer.action === "delete"` → `setDeleteConfirmationOpen(true)` → `ConfirmationDialog` ("Delete lost workspace? … Anything pushed to GitHub … remains available.") → only then `onPrimaryAction()`.

Failure scenario: sandbox is killed, workspace goes `lost`, the composer shows a `Delete` button one row below where `Send` normally sits; a mis-click permanently removes the workspace record with no confirm and no undo. This is a destructive-capability regression and is not in the disclosed judgment calls.

Fix: keep a confirm for the delete action — either render a `ConfirmationDialog` from the takeover when `mode === "lost"`, or make it two-step in the control row (`Delete` → `Confirm delete`), or move the confirmation into `useCloudWorkspaceStatusScreenActions` so every caller inherits it.

**S1-2. The provisioning takeover removes the "type while the cloud workspace provisions" path.**
`use-composer-blocked-state.ts:67-86` maps `panelState.kind === "pending"` (cloud-created / cowork-created) to a takeover, so `ChatInput` swaps out the textarea. But:

- `lib/domain/chat/composer/chat-input.ts` `resolveChatInputAvailability` returns `{ isDisabled: false, disabledReason: null, sendBlockedReason: null }` for a pending entry whose `stage !== "failed"` — the composer is *deliberately* enabled during provisioning.
- `hooks/chat/workflows/use-chat-prompt-actions.ts:157-201` has a dedicated branch for exactly this state (`!selectedWorkspaceId && pendingWorkspaceEntry && pendingWorkspaceUiKey`) that queues the prompt against `buildPendingWorkspaceUiKey(entry)`.
- The retired panel's "Queued prompt will send when this cloud workspace is ready." row existed precisely because the composer stayed usable underneath it.
- The design README says: "blocking → takeover, **non-blocking ambient states should NOT take over the composer**." Non-failed provisioning is the one state the availability resolver classifies as non-blocking.

Failure scenario: user creates a cloud workspace from Home, lands in the chat view while it provisions, and can no longer type or queue a first prompt — they must wait for `ready`. The PR discloses dropping the queued-prompt *note*, not the queued-prompt *capability*.

Fix: exclude non-failed `pending` from the takeover (leave provisioning with no composer surface, or keep it ambient) while keeping `provisioning-failed` as a takeover — availability *does* disable there (`isDisabled: true, "Resolve workspace setup before starting chat."`). If the intent really is to block typing during provisioning, that needs a matching change in `resolveChatInputAvailability` + the prompt-actions pending branch, and an explicit disclosure.

---

### S2 — should fix

**S2-1. No live-region announcement; the restore failure lost its `role="alert"`.**
`components/workspace/chat/input/ComposerBlockedStatusLine.tsx:24` is a bare `<div>`. The takeover appearing, and — worse — a failed restore swapping `message` to the typed failure copy (`use-composer-blocked-state.ts:47`, `composer-blocked-state.ts` `message: restoreError ?? …`) are both silent to assistive tech. The deleted `WorktreeMissingAttachedPanel` rendered `restoreError` in `role="alert"`, and its deleted test asserted `screen.getByRole("alert")`. Fix: `role="status" aria-live="polite"` on the wrapper (assertive when `tone === "destructive"`).

**S2-2. No focus handling across the textarea's disappearance/return.**
When the takeover mounts with the editor focused, focus falls to `<body>`; nothing moves it to the recovery buttons. Meanwhile `ChatInput.tsx` keeps both focus effects live — the `focusRequestNonce` effect retries `focusComposer()` 8× at 25 ms against an element that no longer exists, and `handleComposerSurfaceClick` calls `focusChatInput()` when the user clicks the status line. Bounded and harmless, but it means a keyboard user is dropped out of the composer with no landing spot. Fix: focus the primary recovery action on takeover mount and short-circuit the focus effects while `blockedPresentation` is non-null.

**S2-3. The wiring layer and the takeover branch are untested.**
`composer-blocked-state.test.ts` is genuinely good (see clean list), but it only tests the pure mapping. There is no test for `useComposerBlockedState` (bucket construction, the `""`-sentinel path, `suppress`), none for `ComposerBlockedStatusLine` / `ComposerBlockedControlRow`, and none for `ChatInput`'s takeover branch. The deleted `WorktreeMissingAttachedPanel.test.tsx` covered four real behaviours (action ordering, ineligible-kind negative control, restoring disables competing actions, restore-error visible + actionable) and nothing replaced it. Both S1s above are wiring-layer bugs the pure tests structurally cannot catch. Fix: a hook test with mocked `useWorkspaceStatusPanelState` / action hooks covering directory-missing (eligible, ineligible, restore-error), cloud-status `lost`/`blocked`, runtime `failed`/`claim_required`, and `suppress: true`.

**S2-4. Dead code and a stale ratchet entry left behind.**
- `components/workspace/chat/surface/CloudStatusCompactHeader.tsx` now has zero importers (it existed only for the two deleted cloud panels).
- `ComposerAttachedPanelRow` (`components/workspace/chat/input/ComposerAttachedPanel.tsx:118`) is now an exported-but-unused symbol — the deleted panels were its only consumers.
- `scripts/appearance_scaling_baseline.json:78` still carries `…/surface/WorkspaceArrivalAttachedPanel.tsx|z-50: 2` for a file this PR deletes. `check_appearance_scaling.py` passes (the `standardNumericZ` census has no stale-entry detector), so it silently pre-authorizes two `z-50` sites at a path that no longer exists.

This PR does trim dead code elsewhere (`useWorktreeMissingActions`, `MissingCheckoutCopy` fields), so these are consistency gaps rather than a different standard.

---

### S3 — nits

**S3-1.** `ComposerBlockedControlRow.tsx:47` hardcodes `isDisabled` but passes no `disabledReason`, so the permanently-disabled send button's `title`/`aria-label` reads "Send (⌘↵)" instead of the block reason the normal row supplies via `sendBlockedReason`. Pass `disabledReason={presentation.message}` (or the availability `sendBlockedReason`).

**S3-2.** `ComposerBlockedControlRow` doesn't forward `isEditingQueuedPrompt` to `ChatComposerActions`. If the user was editing a queued prompt when a block lands *and* a session is running, the button renders as Stop, but `ChatInput.tsx`'s `onCancel` routes to `cancelEdit()` first — so pressing Stop silently discards the edit instead of stopping the agent. When nothing is running there is no affordance to leave edit mode at all; the state just persists invisibly until the block clears.

**S3-3.** Paste stays wired during a non-disabled takeover. `handlePaste` is on `ChatComposerSurface`, and `canAcceptPastedAttachments` (`ChatInput.tsx`) is true during provisioning (`isDisabled` false, `areRuntimeControlsDisabled` false), so a pasted image is added to an attachment list the takeover doesn't render — invisible and unremovable until the block clears. (For worktree-missing, `areRuntimeControlsDisabled` is true, so that path is already safe.)

**S3-4.** Precedence doc vs reality. `composer-blocked-state.ts`'s resolver comment and `composer.md`'s "Precedence: directory missing, then provisioning, then cloud status, then runtime" don't describe the real ordering: `use-workspace-status-panel-state.ts` resolves `pending` **before** `directory-missing` and returns a single kind, so the three panel-state buckets are mutually exclusive and the resolver's directoryMissing-over-provisioning branch is unreachable. Suggest saying so explicitly (ordering only ever arbitrates panel-state vs `cloudRuntime`).

**S3-5.** The `useMemo` in `use-composer-blocked-state.ts:93-191` never hits: `worktreeMissingActions`, `pendingEntryActions`, `cloudStatusActions`, and `selectedCloudRuntime` are all fresh object literals from their hooks each render. Correct, just decorative. Relatedly, the disclosure "both are action-only (no queries), so no spurious work fires" overstates: `usePendingWorkspaceEntryActions` mounts `useWorkspaces()` plus the create/entry/selection/materialization hook trees inside the always-rendered `ChatInput`. No *new* network in practice (`useChatAvailabilityState` already mounts `useWorkspaces()` in the same component and react-query dedupes), but it's a bigger unconditional hook tree than the sentence implies.

**S3-6.** `missingCheckoutComposerMessage` uses an ASCII apostrophe (`it's`); the design copy uses a typographic one (`it's`).

**S3-7.** `cloud-attention` shows `model.description` only, dropping the mode title the old compact header carried as `phaseLabel`. For `mode === "error"` the description is `workspace.lastError` — a raw server string with no "Provisioning failed" framing. Probably acceptable under a one-sentence design; worth a look at what terse `lastError` values actually read like in the composer.

---

### Checked clean

- **`runtimeState.title && runtimeState.subtitle` gate cannot silently unblock.** `buildSelectedCloudRuntimeViewModel` returns null title/subtitle only for `phase: "ready"`; `resuming`, `failed`, and `claim_required` all populate both. The gate reproduces the deleted panel's guard exactly, so no real blocking runtime state slips through into a typable composer.
- **Send is blocked store-side, not only visually.** `resolveChatInputAvailability` sets `sendBlockedReason` for directory-missing and `isDisabled` for cloud status ≠ ready / runtime ≠ ready; `useWorkspaceRuntimeBlock` throws `workspaceDirectoryMissingBlockError` at session-create sites. Untouched by this PR.
- **Stop still works while blocked.** `ChatComposerActions` with `isRunning && isDisabled` → `canQueue` false → renders the Stop button wired to `onCancel`.
- **Drafts survive the takeover.** Draft text lives in `chat-input-store.draftByWorkspaceId` keyed by `workspaceUiKey` (`use-chat-draft-state.ts`), not in the unmounted editor, so it returns intact when the block clears.
- **Shared-file contract with #1713 holds.** `git diff 3ebbae102...HEAD` over `resolve-dock-slots.ts` and `use-composer-dock-slots.tsx` has zero added/removed lines matching `todo` (every todo line in the diff is context). All hunks are ambient-slot removals.
- **No dangling references.** Repo-wide search finds no remaining source use of `WorkspaceArrivalAttachedPanel`, `WorktreeMissingAttachedPanel`, `CloudRuntimeAttachedPanel`, `WorkspaceArrivalCloudPanel`, `ComposerDockAmbientSlot`, `resolveAmbientSlot`, `suppressWorkspaceStatusPanels`, `ambientContextSlot`, `hasWorkspaceStatusPanel`, `hasCloudRuntimePanel`, or `deleteConfirmBody` (the only residue is the stale JSON baseline entry in S2-4). `missingCheckoutCopy`'s surviving `title`/`sendBlockedReason` are still consumed by `sidebar-indicators.ts`, `use-chat-availability-state.ts`, and `use-workspace-runtime-block.ts`.
- **`""`-sentinel hooks fire nothing.** `useWorktreeMissingActions("")` holds a mutation + store reads only. `useCloudWorkspaceStatusScreenActions("", "pending")` returns `handlePrimaryAction: null` on the pending branch before any callback can run. Neither issues a query.
- **Allowlist 11→9 and lint gates verified.** `check_frontend_boundaries.py` and `check_appearance_scaling.py` both pass on this head.
- **Pure-module test quality.** Precedence is pinned with a negative control at every level (missing > provisioning > cloud status > runtime, plus the empty case and `ready` → null); all 8 kinds have presentation assertions with icon/tone; action gating is pinned both ways (restore null when ineligible, cloud-attention with and without a primary action, retry/claim null cases); the restore-error message override is asserted directly.
- **Playground swap is asserted, not just added.** `playground.test.ts` now asserts `renderAttachedSlot(scenario) === null` **and** that the composer surface renders for every cloud scenario — a real move assertion.
- **Design fidelity vs `Blocked Status.dc.html` / README.** Status line `pt-0.5 pb-3.5` = the specified `padding: 2px 0 14px`; `gap-2` = 8px; `text-composer`; `CircleAlert` in `text-destructive`, `Spinner` in `text-faint` (all spinner kinds carry tone `faint`, so the conditional cannot produce a destructive spinner); message in `text-foreground`, single sentence, no em dash. Control row uses `ChatComposerControlRowFrame` with the leading cluster dropped and actions in the trailing slot, secondary-then-primary, send disabled — exactly the handoff. The worktree sentence matches the design copy verbatim. The `claim_required` extension and reuse of existing runtime `actionBlockReason` copy follow the README's "reuse existing copy where it exists".
- **Specs vs code.** `composer.md`'s dock-slot section and `cloud-workspace.md`'s file map match the shipped code (modulo S3-4's precedence wording).

— dedicated review agent (opus), spawned per per-PR review policy

pablonyx:
## Response to dedicated review — fixes @ 63386d5be

**S1-1 (unconfirmed lost-workspace delete) — FIXED.** Actions can now declare a confirmation in the presentation (`ComposerBlockedActionConfirmation`); the hook attaches the retired panel's dialog copy verbatim to the lost-mode delete, and `ComposerBlockedControlRow` interposes a `ConfirmationDialog` (destructive variant, loading-locked) before `onSelect` fires. Pinned by a wiring test.

**S1-2 (provisioning takeover removes an intentionally-enabled composer) — FIXED.** In-flight (non-failed) provisioning no longer produces a takeover at all: the `provisioning` kind is deleted from the union and the input bucket is `provisioningFailed` (type-level: the non-blocking state is unrepresentable). The composer stays enabled and the queue-first-prompt path works as before; failed provisioning still blocks. Negative-control verified: re-adding the non-failed branch fails the new test. Note the design handoff lists "cloud provisioning" in the takeover scope overview but its own state-management rule says non-blocking ambient states must not take over — availability is the authority on "blocking", so the rule wins; provisioning progress consequently has no composer surface until failure. Flagged for Pablo in case he wants the overview reading instead.

**S2s — FIXED.** Status line: `role="alert"` for destructive tones, `role="status"` otherwise. Focus: when the takeover clears and the textarea remounts, focus returns via `focusComposer()` (suppressed for cowork/autofocus-suppressed hosts). Wiring tests: new `use-composer-blocked-state.test.tsx` (10 cases — suppress, both S1s, restore gating/error override, confirmation payload, title framing, runtime mapping, panel-over-runtime precedence), replacing the deleted panel test's behaviors at the hook layer. Dead code: `CloudStatusCompactHeader.tsx` deleted, `ComposerAttachedPanelRow` export removed, stale `WorkspaceArrivalAttachedPanel.tsx|z-50` baseline entry pruned (baseline shrink).

**S3s — FIXED:** disabled send now carries the blocked message as `disabledReason`; `isEditingQueuedPrompt` forwarded so Stop doesn't silently `cancelEdit()`; paste attachment intake gated off while blocked; precedence prose (module + spec) rewritten to match the actual upstream resolution in `use-workspace-status-panel-state`; typographic apostrophes; cloud-attention keeps the model title as framing ("Provisioning failed. <lastError>"). **Left as is:** the `useMemo` (harmless; deps are fresh objects but the compute is trivial) and the `usePendingWorkspaceEntryActions` mount cost (pre-existing pattern in `ChatInput`'s tree).

Verification: tsc clean; 118 files / 771 tests green across the chat/composer trees; all repo-shape checkers pass.

— author session (Fable)

pablonyx:
**Merge-order note (stacked siblings #1716 + #1713):** both PRs shrink the same `frontend_boundaries_allowlist.txt` entry for `use-composer-dock-slots.tsx` (11→9 here, 11→10 in the sibling). Git auto-merges them into two duplicate lines. Whichever sibling merges second needs a one-line touch-up: collapse to a single entry with count **8** (11 − 2 ambient imports − 1 todo import). `check_frontend_boundaries.py` on the second PR's post-retarget CI will flag it if forgotten. — author session (Fable)

pablonyx:
Restacked onto #1710's feedback-pass head — this PR's head is now `6e3801def` (clean rebase, no content changes to the takeover itself).

One real copy defect found during live verification and fixed here: every blocked-family cloud-status description already opens by restating the title ("Cloud usage is paused …"), so the takeover's `${title}. ${description}` framing doubled it. New shared `composeBlockedStatusMessage` in `composer-blocked-state.ts` drops the title when the description already restates it (and for pending/absent titles); the hook and the playground fixture both use it, with unit coverage.

All 8 takeover states verified live in the dev app (worktree-missing, first-runtime, provisioning, applying-files, blocked, error, reconnecting, reconnect-error).

**Merge-order note**: this PR and #1713 are siblings off #1710. Whichever merges second will conflict in 5 shared dock files (`resolve-dock-slots.ts`, `use-composer-dock-slots.tsx` + test, `frontend_boundaries_allowlist.txt`, `specs/product/composer.md`) plus the structure-allowlist ChatInput entry. Correct union: activeSlot = approvals/questions/MCP only; attachedSlot = parallel delegated work only; blocking state takes over the composer; todo progress via floatingSlot.

pablonyx:
Rebased onto #1710's new head c1fda71a9 → this branch now at 7de21fae3 (clean rebase, no conflicts).

PR #1723, opened 2026-08-07
docs(adrs): propose workflows v1 launch architecture

tairabun:
## Summary
- add the working ADR for the Workflows v1 product and local-execution architecture
- map the current Workflows beta and the primitives that the replacement rebuilds or reuses
- define the cross-plane execution flows, ownership cells, open decisions, and proposed PR ladder
- link each user-facing flow to its playable Claude Design state and identify operator-only flows without a product UI

## Test plan
- [x] `python3 scripts/check_docs.py`
- [x] `git diff --check origin/tairabun/workflows-v1-release...HEAD`

## Review status
This is a working decision record opened as a draft for architectural review. Per the ADR lifecycle, it should not merge until the implementation ladder's final PR.

Made with [Cursor](https://cursor.com)


adrs/2026-08-07-workflows-v1-launch.md
```diff
@@ -0,0 +1,1973 @@
...
+- Triggering workflows through webhooks or schedules. V1 supports only manual
+  triggers.
+- Executing workflows in a cloud sandbox. V1 execution is local only.
+
+### Requirements
+
+1. The implementation consists of four sequential PRs: engine, data models,
+   workflow lifecycle orchestration, then UI and client. The ladder may carry
+   temporary migration code, but the released feature has one graph definition,
+   invocation, and run path. Each PR removes the beta behavior whose graph
```
pablonyx: Not sure what ‘PR removes the beta behavior whose graph replacement becomes complete in that PR.’ means is specified

adrs/2026-08-07-workflows-v1-launch.md
```diff
@@ -0,0 +1,1973 @@
...
+8. Each `agent` and `human_in_loop` node has its own model configuration.
+9. Each workflow run creates its own workspace.
+10. A `workflowDefinition` specifies whether a run creates a new worktree
+    or runs at the repository root by default.
+11. When manually triggering a workflow, the user can override the definition's
+    worktree or repository-root default in the UI.
+12. One workflow run owns one workspace, and each agentic workflow run node
+    maps to one session in that workspace.
+13. Each `agent` and `human_in_loop` node receives a prompt, a set of goals, and
+    optional verification methods.
```
pablonyx: I think these two nodes are extremely underspecified in the requirements section if we want the requiremnts section to stand on its own as a reference-able list.

What is an agent node? What is a human in the loop node?

adrs/2026-08-07-workflows-v1-launch.md
```diff
@@ -0,0 +1,1973 @@
...
+14. A `choice` node evaluates a condition and selects an outgoing branch.
```
pablonyx: Don't think we fully aligned on the 'choice' node concept and I'm not entirely sure which conditions it would be evaluating here or how- likely answered further down in the doc but i think we should make requirements explicit (or at least explicit enough for reviewer to grok them)

adrs/2026-08-07-workflows-v1-launch.md
```diff
@@ -0,0 +1,1973 @@
...
+15. A valid workflow has exactly one `succeed` node and exactly one `fail` node;
+    these are its only terminal node types.
+16. Every non-terminal node has an edge to `fail` for runtime errors and
+    exceptions.
+17. Multiple nodes may have an edge to `succeed`.
```
pablonyx: I think clarifying whether the fail / success node is a user-visible concept would be helpful. I'm assuming we don't want to surface this terminology to the user and this is just a concept on the state machine side, but reading this, I'm not sure.

adrs/2026-08-07-workflows-v1-launch.md
```diff
@@ -0,0 +1,1973 @@
...
+18. All nodes in a workflow run share context under
+    `<workspace-root>/context/shared/<document>`. Shared context is typically
+    Markdown but may be any file or artifact usable by agents and humans.
+19. All nodes have full read and write access to
+    `<workspace-root>/.proliferate/workflows/<workflow-name>/shared/`.
+20. Each agentic node can create files and artifacts under
+    `<workspace-root>/.proliferate/workflows/<workflow-name>/<node-index>/`, where
+    `<node-index>` is the zero-padded agentic-node index (`00`, `01`, `02`, and
+    so on).
```
pablonyx: Is the context shared in `<workspace-root>/context/shared/<document>` or in `<workspace-root>/.proliferate/workflows/<workflow-name>/<node-index>/`? Seems like a contradiction.

I thought we aligned on neither, actually- and something more like `~/.proliferate/workflows/<workflow-id>/` containing all the context docs.

My impression of the model we aligned on was:
1. Per workflow, the user can define a set of what we call "seeded context docs" that are numbered and have pre-defined file names.
2. The user can write to those template / files to give every context doc a starting state. (e.g. write out headers + instructions) as part of a workflow definition
3. The user can reference those context docs in instructions to the agents in workflow steps and the links will be formatted such that they refer to the proper file path (see above)
4. Agents can write any other content they want to to the shared folder as they see fit

To further elucidate, I think the model sketched above (in point 20) could work as well but it precludes collaboration between agents  (if you define the boundaries very strictly)

pablonyx: More broadly why do we have three different folders?

adrs/2026-08-07-workflows-v1-launch.md
```diff
@@ -0,0 +1,1973 @@
...
+
+workflow invocation schema 1  managed-cloud target      ── terminalize, then archive or delete
+workflow invocation schema 2  local graph-run delivery  ── only accepted member after cutover
+
+AnyHarness workflow run schema 1/2  beta one-prompt run ── terminalize, then archive or delete
+AnyHarness workflow run schema 3    graph run            ── only accepted member after cutover
+
+workspace materialization schema 1  beta placement       ── archive or delete with its beta run
+workspace materialization schema 2  graph-run placement  ── only active member after cutover
+```
```
pablonyx, line 250: Does this belong in the current context section?

adrs/2026-08-07-workflows-v1-launch.md
```diff
@@ -0,0 +1,1973 @@
...
+22. A workflow run workspace uses the same UI and UX primitives as a
+    normal workspace and its sessions.
+23. A user can chat in any workflow-generated session.
+24. A user can manually create additional sessions in a workflow run
+    workspace to work with the workflow's shared context.
+25. Agentic nodes (i.e. `agent`/`human_in_loop`) should inherit permissions for
+    all configured integrations that the user has for their normal sessions.
+
+## Current context
+
```
pablonyx, line 193: Maybe more of a nit but this section read more like a 'migration decisions' than a 'current context' doc. I think the ideal outline for this kind of section in my mind would have been:
1. High level context
2. Source tree (quite helpful- maybe also a mermaid diagram since our grid cells are so tight)
3. Slightly clearer descriptions of the distinct cells- current format is a paragraph and then migration details in bullets, we can probably come up with something clearer.

Basically what you have now but formatted to be focused on the current state rather than migraiton implications.

pablonyx (commented):
Directionally am aligned but think we need to clean up some of these comments (not necessarily the nits)

Specifically, I think we need to be less ambiguous re: choice conditions and fully finalize the context/artifact model

adrs/2026-08-07-workflows-v1-launch.md
tairabun: updating with 
```
Across each PR, we should deprecate the requirements/
   assumptions about the strictly linear structure of workflows in light of the 
   graph structure, which is a superset of the current linear constraint.
```

adrs/2026-08-07-workflows-v1-launch.md
tairabun: just for organization. Took inspiration from humanlayer impl where they have something like:
<worktree-root-path>/.humanlayer
    /tasks
        /<task-name-1>
        /<task-name-2>
            01-research-questions-workflow-feature.md
            02-research-workflow-feature.md                 
            ticket.md
            ...
    workspace.json

adrs/2026-08-07-workflows-v1-launch.md
tairabun: addressed by 483215b

adrs/2026-08-07-workflows-v1-launch.md
tairabun: scoped out choice node type in v1 launch: 483215b

adrs/2026-08-07-workflows-v1-launch.md
tairabun: Aligned on having a single folder `<workspace-root>/.proliferate/context/` where all context docs live, and a decision to not give specific locations for each agent to generate scratch files/artifacts. 
Addressed in 483215b

tairabun: addressed in 483215b. See comment right above

adrs/2026-08-07-workflows-v1-launch.md
tairabun: to be improved in the next iterations of adrs

adrs/2026-08-07-workflows-v1-launch.md
tairabun: to be improved in the next iterations of adrs

PR #1725, opened 2026-08-08
fix(product): stabilize command disclosure reveal

pablonyx:
## Summary

- stage lazily mounted command-disclosure content in a committed collapsed frame before revealing it
- advance disclosure geometry and opacity together while keeping transient content inert and cancelling interrupted openings
- keep the bounded command ledger and rotating chevron on stable compositor layers so WebKit does not re-snap their SVG glyphs after the transition settles
- add regression coverage for first-open timing, rapid close, initially expanded content, command ledgers, compositor retention, and nested completed history

## Root cause

The disclosure previously mounted at its expanded geometry before the browser could paint a collapsed frame. After that timing fix, WebKit still released the temporary opacity/rotation layers at the transition endpoint and re-rasterized inline SVGs at a neighboring device-pixel origin. The summary glyph outside the animated subtree remained stable, while ledger glyphs moved by roughly one CSS pixel.

## Testing

- Tier 1: six focused product-client Vitest files — 39 tests passed
- `tsc -p tsconfig.json --noEmit`

## Observability

- none; this changes client presentation timing and compositor ownership without adding a new operational boundary or failure mode

## Readiness

## Verification

- `git diff origin/main...HEAD --check`
- verified the existing `PROFILE=main` stack from `~/proliferate`: Desktop and Web return HTTP 200, API and AnyHarness report healthy
- compared Safari/WebKit activity-ledger frames at one and nine seconds after expansion: summary, chevron, and ledger SVG regions were byte-identical with `(0, 0)` movement


pablonyx (commented):
Rebased onto origin/main at 5fa27645e02ba6e8cc651deea53a8571551c4409 (one conflict in TurnItemSequence.test.tsx resolved: kept both the ReactNode type import and the act import, both used post-Chat-Scroll-ladder).

Reviewed the rebased diff against the current CollapsedActions.tsx/AnimatedCollapsibleContent.tsx (post Chat Scroll ladder merges):
- The fix (double-rAF gating of visuallyExpanded, plus transform-gpu/will-change hints on the chevron and content shell) is still structurally compatible with the current primitive — no conflicting logic was introduced by the ladder's changes to these files.
- CollapsedActions.test.tsx and AnimatedCollapsibleContent.test.tsx assert the new data-expanded/gridTemplateRows/opacity sequencing across the requestAnimationFrame mock, and still pin the original stabilization bug (motion shell must not flip to expanded until after the first commit's rAF fires).
- No regressions apparent in the rebase; conflict resolution was mechanical (import merge only).

Will report CI tally once all checks complete.

pablonyx (commented):
CI status update at head 5fa27645e (post-rebase):

- CI (Shared frontend packages job): FAILURE, both original run and rerun. Failures land in a different, unrelated set of test files each time (HarnessPane.test.tsx, HarnessAllModelsSection.test.tsx, ToolCallSummary.test.tsx, ClosedChatTabsMenu.test.tsx — none touched by this PR's diff), all 'AssertionError: expected false to be true' — consistent with pre-existing flakiness in the shared product-client test run, not a deterministic regression from this diff.
- Confirmed this is NOT caused by the rebase: the PR's original pre-rebase head (7c9fd7320) also has CI=failure in the same 'CI' workflow (checked via gh run list against that SHA).
- CodeQL: success. Self-Host Smoke: success. Intent Tests: success. PR Metadata (x3): success.

Verdict: NEEDS-FIXES — this PR cannot go green until the pre-existing CI flake/failure in the shared product-client suite is resolved (unrelated to the command-disclosure fix itself, which looks correct per my earlier comment). Recommend re-running CI once more and/or triaging the flaky suite separately before merge; founder should decide whether to accept if flake is confirmed unrelated to this change.

pablonyx (commented):
Fix-forward update at head a95a0e68b:

**Regression mechanism confirmed and fixed.** `AnimatedCollapsibleContent` (a shared primitive) now defers `visuallyExpanded` behind one `requestAnimationFrame` after `expanded` flips true, so any consumer that asserts expanded/visible state synchronously (or via fake *timers* only, not rAF) after triggering an expand will observe stale state. This PR's own tests (`CollapsedActions.test.tsx`, `AnimatedCollapsibleContent.test.tsx`) already accounted for this; four sibling consumers of the primitive did not:
- `ToolCallSummary.test.tsx` — asserted `data-expanded` synchronously after `fireEvent.click`.
- `ClosedChatTabsMenu.test.tsx` — asserted `data-expanded` after `vi.advanceTimersByTime`, but the re-expand (on archive-fail) is scheduled mid-tick from a promise `.then`, one clock tick too late for that same `advanceTimersByTime` call to reach it.
- `HarnessAllModelsSection.test.tsx` / `HarnessPane.test.tsx` — switches live inside the gated subtree; `inert={!visuallyExpanded}` kept them out of the accessibility tree before the flush.

Fixed by mocking/flushing `requestAnimationFrame` in each, matching `CollapsedActions.test.tsx`'s pattern (commits 9b694e920, 0c8fd6272, a95a0e68b). One iteration correction along the way: `ToolCallSummary`'s divider+ledger both mount the primitive, so a single `shift()` of the mock rAF queue picked up a stale, already-cancelled frame from the first click instead of the live one from the second — fixed by draining the queue instead of shifting once. Also had to keep `HarnessPane.test.tsx`'s fix at net-zero line growth: that file is grandfathered at exactly 1208 lines (PROD-SIZE-1, zero slack), so the mock is inlined into `expandModelList()` itself rather than via a file-level `beforeEach`.

**Literal CI tally at head a95a0e68b** (job-level, one CI run):
Build SDK: success · Candidate build handoff: success · Cargo check & test: success · CI/CD config checks: success · Desktop frontend build: success · Login first-load budget (phase-6): success · Mobile typecheck: success · **Repo shape checks: failure** · **Shared frontend packages: success** · Terraform validate (server/infra): success · Web frontend build: success · Workflow definition lifecycle (tier-2): success. Separately: CodeQL: success, Intent Tests: success, Self-Host Smoke: success, PR Metadata: success.

The one remaining failure (Repo shape checks) is NOT caused by this PR: `anyharness/crates/anyharness-lib/src/live/workflows/actor.rs` is 603 lines against a ratchet recorded at 606 in `lints/anyharness/ratchets.toml` — confirmed present and failing on origin/main's own latest CI run (773551737) independent of this branch. This PR never touches that file; out of scope for a command-disclosure fix and left untouched here.

**Residual design question for Pablo:** the primitive now keeps freshly-expanded content `inert` for one rAF after `expanded` flips, which is invisible to any consumer test that doesn't know to mock/flush `requestAnimationFrame` (as 4 sibling test files here didn't). Option (a), applied in this fix-forward, is mechanical but pushes rAF-awareness onto every consumer's tests. Option (b) — giving the primitive a synchronous/`act()`-flushable path (e.g., an exported test-only flush hook, or driving the stabilization via a layout effect + microtask instead of rAF) — would remove that burden but is a design change beyond this PR's scope. Flagging for founder call; not attempted here.

Verdict: MERGE-READY on the frontend-testing question this fix set out to answer — `Shared frontend packages` is green. Blocked only by the pre-existing, unrelated `Repo shape checks` staleness on main, which is a separate item for whoever owns that ratchet.

PR #1722, opened 2026-08-07
chore(lints): docs-v1 D6 — bespoke checkers become lints/ records + generated diagnostics

pablonyx:
# docs-v1 D6: move bespoke lint enforcement into `lints/` rule records

Slice D6 of the docs-v1 restructure (living ADR `adrs/2026-08-05-docs-restructure.md`, PR #1653). Behavior-preserving: every bespoke Python checker keeps its verdicts (proof below); what changes is where the rules LIVE and what the diagnostics SAY.

## What this does

- **`lints/README.md`** — the contract. Rule records are canonical; prose docs cite rule IDs, never restate rules. Weakening a rule, adding a net-new exception, or deleting a pinning test is an amendment requiring founder approval.
- **`lints/<owner>/*.toml`** — one `[[rule]]` record per enforced rule, owners = anyharness / server / frontend / product. New ID vocabulary (`SRV-*`, `AH-*`, `FE-*`, `PROD-*`) minted here; the old SCREAMING_SNAKE names die in this PR, no aliases.
- **`lints/<owner>/exceptions.toml`** — fine-grained grandfathering: one entry per violation SITE (path + site + reason). Counts are gone; a NEW violation in a grandfathered file now fails CI.
- **`lints/<owner>/ratchets.toml`** — shrink-only measured debt (max-lines et al), converted from the old `path max_lines reason` allowlists.
- **`scripts/lint_records.py`** — shared strict loader + `render_diagnostic`: every CI failure now teaches the rule, the legal alternative, and the record path — never a bare "banned".
- **Checkers rewired** to load records and emit generated diagnostics; old txt allowlists deleted.
- **CODEOWNERS**: `/lints/` → @pablonyx (constitution gate).
- **`lints/native-tools.md`** — ledger for natively-enforced rules (Clippy, rustfmt, tsc, Ruff, mypy, Terraform); follow-up scope, untouched in v1.

## Proof of verdict preservation

Before/after checker outputs captured for every owner slice (identical pass/fail verdicts; wording intentionally changed to generated diagnostics). Negative controls: fresh violations fail with the new diagnostics; a second violation in a grandfathered file fails (fine-grained sites are load-bearing).

[...]

PR #1726, opened 2026-08-08
refactor(anyharness): agent ops MCP shell — rename, inject for all sessions, authorize funnel

pablonyx:
PR 1 of the agent operations layer (spec: adr-agent-ops.md / Obsidian "Full ADR"). Zero new behavior — this is the shell every later step lands tools inside.

## What

- **Rename**: `domains/sessions/subagents/mcp/` → `domains/sessions/agent_ops/`; the subagents product MCP becomes the **agent ops MCP**, subagents a tool class within it.
- **Launch-frozen compat** (sessions bake MCP URL + signed token at launch and never refresh):
  - `ID = "subagents"`, `SECRET_FILE_NAME`, `ACP_SERVER_NAME = "subagents"` all stay frozen — renaming any of them would 401 or de-render every live session.
  - `ROUTE_SLUG` moves to `agent-ops`; new `legacy_route_slugs` on `ProductMcpDefinition` keeps `subagents` routable to the same server (duplicate legacy slug = startup error, not silent shadow).
- **Tool aliases** (cowork's deprecation pattern): `create_subagent → spawn_subagent`, `close_subagent → close_agent`; aliases dispatch to identical implementations and still take the workspace write lease; `tools/list` advertises only the canonical set.
- **Inject for all sessions**: the selection closure drops its conditions (`should_attach` = yes); every call-time check stays exactly as strict as today (`subagents_enabled` still read fresh per spawn attempt).
- **`authorize(caller, target)`** (new `domains/sessions/authorize.rs`): the single access funnel — today's two rules only (runtime-wide visibility; closed target readable but unsendable). Ownership/promotion arrive in PR 5.
- Client/SDK: reducer + product-client accept both old and new tool names so renamed calls in new sessions keep rendering as subagent activity. Full client rename is step 9.

## Tests

- `cargo test -p anyharness-lib`: 1545 passed, 1 failed — `sweep_tests::the_sweep_removes_only_abandoned_and_old_scratch_roots`, proven pre-existing (fails identically on base).
- `agent_ops` (12) + `authorize` (5) suites green; SDK vitest 133 passed.
- Negative controls fired: legacy-slug loop removal, empty alias table (which exposed and fixed a vacuous test), restored selector conditions — each makes its test fail.

## Disclosure

Cowork-surface sessions now mount an agent-ops MCP whose every request — including `tools/list` — 409s (`resolve_context` still rejects non-Standard surfaces, kept per "call-time gating stays as strict as today"). That means the harness sees a visible failed-discovery state on those sessions, not a silent no-op. Transient by design — cowork is deleted in PR 8; flagging for the record.


pablonyx (commented):
**CLEAN on the load-bearing contract — no behavior defect found.** The three launch-frozen anchors hold at every place they are consumed, the legacy slug really does route to the same handler and really does fail startup on collision, both aliases dispatch to identical implementations *and* are listed for the write lease, call-time gating is byte-for-byte what it was, and `authorize` is a genuine no-op wrapper over today's rules on both paths it was wired into. Findings below are one forward-looking trap, two test-coverage gaps on exactly the invariants this PR exists to protect, and spec-doc drift.

What I verified by reading (not running):

- **Identity never derives from `ROUTE_SLUG`.** The slug is used in exactly two places: URL construction at launch (`domains/sessions/mcp_bindings/product_catalog.rs:107`) and route lookup (`api/http/product_mcp.rs:51`). The token scope is built from `definition.id` *after* the handler is resolved (`api/http/product_mcp.rs:53` → `ProductMcpRequestContext::new(.., definition.id)`), so a request arriving on the legacy `subagents` slug validates against `product_mcp_id = "subagents"`. `connection_id` and `server_name` in the injected config also come from `id` / `acp_server_name` (`product_catalog.rs:99-101`), both frozen. No third consumer of the slug exists.
- **Alias → lease is correct.** `endpoint_operation_kind` matches `mutating_tools` against the raw wire name (`mcp_bindings/product_registry.rs:86`, fed by the un-canonicalized `params.name` from `product_server/endpoint.rs:20-25`), and `MUTATING_TOOL_NAMES` lists both aliases (`agent_ops/tools.rs:18-25`). `tools/list` advertises canonical only.
- **Inject-for-all doesn't broaden reach.** The three dropped conditions are all still enforced downstream: surface at `agent_ops/context.rs:38`, `subagents_enabled` + both depth checks + fanout + access gate at `subagents/service.rs:113-134`, re-run at call time in `spawn_subagent` (`agent_ops/calls.rs`, `validate_parent_can_spawn` after the `can_create` bail). Crucially every target-resolving tool still goes through `SubagentService::resolve_target`, which filters `relation == Subagent && parent_session_id == caller` (`subagents/service.rs:230`) — so a newly-mounted subagent child or subagents-disabled session has no reachable target.
- **`authorize` neither broadens nor narrows.** Both call sites still run the link-scoped resolve first (`agent_ops/calls.rs:380` then :381; :454 then :459), so the only added rules are closed-caller (unreachable — `resolve_context` already 409s a closed parent) and closed-target-on-Send (previously produced the same rejection one layer down, in `send_text_prompt_with_provenance`). Read keeps closed targets readable. Correct.
- **Client/SDK dual-name recognition is complete.** All five name-matching sites accept both names (`sdk/src/reducer/transcript.ts:1421-1432`; `product-client/.../subagent-launch.ts:287-288`, `.../subagent-tool-presentation.ts:41-42, 77, 96-97, 166-167`; `.../stream-side-effect-plan.ts:280-291`). I swept `apps/**` and `anyharness/sdk/**` for every remaining literal `create_subagent`/`close_subagent` and for component-level `nativeToolName ===` branching — nothing else keys on these names. No Rust path keys on tool-name strings either.

---

## Findings

### 1. `authorize()`'s docstring oversells what it enforces — a trap for PR 5 (medium, forward-looking)

`domains/sessions/authorize.rs:1` — *"The single access-policy funnel every cross-agent operation clears."* As written it enforces **no ownership at all**: any caller may Read or Send to any target runtime-wide. That is correct *today* only because both call sites run `SubagentService::authorize_target` / `resolve_target_including_closed` first, and the link check lives there (`subagents/service.rs:230`), not here.

The failure scenario is PR 5, not this PR. §3.4 of the ADR says `close_agent` becomes "owner check → existing close-tree machinery" and §3.3 assigns "ownership for close/promote" to `authorize`. An implementer who reads this module header and routes `close_agent` through `authorize(caller, target, Close)` — dropping the `resolve_target_including_closed` call that currently supplies the ownership check — hands every session the ability to close every other session, with no test failing (the current authorize suite asserts runtime-wide access *is* allowed).

Cheap fix now: say it explicitly in the doc comment — "ownership is NOT enforced here yet; callers MUST still resolve the link through `SubagentService` until `session_links` carries `promoted_at`/`owned_agent`" — and keep `AgentAccessIntent` free of a `Close` variant until it actually gates one.

### 2. The compat-anchor test pins three anchors and skips the two most fragile (medium)

`agent_ops/definition.rs:141-145` (`compat_anchors_stay_frozen_for_already_launched_sessions`) asserts `id`, `legacy_route_slugs`, `route_slug`. It does **not** assert `ACP_SERVER_NAME` (`definition.rs:33`) or `SECRET_FILE_NAME` (`agent_ops/auth.rs:10`) — the two anchors that carry an explicit "renaming this breaks live sessions" comment and that the PR body names as frozen.

`ACP_SERVER_NAME` is the worse one. It is a plain `const` with no test, and renaming it changes every wire name to `mcp__agent_ops__*`, which simultaneously (a) de-renders every live session's transcript and (b) breaks all ten reducer arms (`sdk/src/reducer/transcript.ts:1421-1432`) and every client presentation arm for **new** sessions too — the dual-name compat this PR just built would not help, because the prefix, not the suffix, changed. Given the test's name promises exactly this, add:

```rust
assert_eq!(DEFINITION.acp_server_name, "subagents");
```

plus a one-liner in `auth.rs` pinning `SECRET_FILE_NAME == "subagent-mcp-token.key"`. Two lines, and they are the whole point of the PR.

### 3. The alias↔`MUTATING_TOOL_NAMES` invariant isn't pinned (medium)

The comment at `agent_ops/tools.rs:16-17` correctly identifies that aliases must be listed because the lease is matched on the raw wire name — and today both are. But the test at `tools.rs:337-345` only proves each entry of `MUTATING_TOOL_NAMES` resolves to an *advertised* tool. It does not prove the converse: that an alias whose canonical target is mutating is itself in the list.

Failure scenario: PR 5 generalizes the close family and adds a third alias (`close_subagent` → `close_agent` is already there; `promote_subagent`/`spawn_agent` will grow the table). Forget the alias entry and calls under the old name skip `WorkspaceOperationKind::SubagentWrite` **and** the `assert_workspace_mutable` check that only runs when a lease is taken (`api/http/product_mcp.rs:71-80`) — a mutating call on a retired/read-only workspace, with no concurrency lease. Suggested test:

```rust
for (alias, canonical) in DEPRECATED_TOOL_ALIASES {
    assert_eq!(
        MUTATING_TOOL_NAMES.contains(canonical),
        MUTATING_TOOL_NAMES.contains(alias),
        "alias {alias} and canonical {canonical} disagree on mutating status"
    );
}
```

### 4. Spec docs still point at the deleted module and the old canonical tool name (medium, review focus 6)

The PR updated `specs/anyharness/README.md`, one line of `servers.md`, and `definitions/subagents.md`. Five files still name `domains/sessions/subagents/mcp/`, which no longer exists:

- `specs/anyharness/domains.md:435`
- `specs/anyharness/integrations.md:349`
- `specs/codebase/platforms/product/mcp-runtime.md:64`
- `specs/codebase/platforms/product/agent-features/servers.md:115` — in the "Examples" block of the *same file the PR edited*
- `specs/codebase/platforms/product/agent-features/definitions/prompt-and-skill-policy.md:209` — points at `subagents/mcp/definition.rs` as the home of a product prompt; that exact file is deleted by this PR

And the tool contract itself is stale: `specs/anyharness/sessions.md:226` still lists `create_subagent` in the authoritative **"Current tools"** list (and never listed the close tool at all). That is the doc most likely to be read as the tool contract by the next implementer.

### 5. `servers.md`'s selection-policy statement is now false (medium, review focus 6)

- `servers.md:513` — "parent session with subagent support / receives subagents MCP with parent-session scope". After this PR, *every* session receives it, scoped at call time.
- `servers.md:762` — the required-tests list still demands "selection does not attach it for unrelated sessions", which is now the opposite of the design.

The doc that governs product-MCP selection should carry the new law, not just the renamed path.

### 6. `{product_mcp_id}` in the URL is now materially wrong (low, review focus 6)

`servers.md:545`, `:562` and `:823` all describe the endpoint as `.../mcp/{product_mcp_id}` and routing as "by product MCP id". This PR is the *first* time id ≠ slug (id `subagents`, slug `agent-ops`): the route dispatches by slug (`api/http/product_mcp.rs:51`) while the token is scoped by id (`:53`). Since the PR is adding `legacy_route_slugs` prose immediately below, correcting these three to `{route_slug}` is the natural companion edit — and it's the exact distinction the compat story rests on.

### 7. The cowork disclosure is slightly understated (low)

The PR says cowork sessions get an MCP "whose every request 409s". It is **`tools/list`** that 409s, not only `tools/call`: `dispatch_product_mcp_request` resolves context for `tools/list` too (`integrations/mcp/product_server/server.rs:187-189`), and `resolve_context` rejects non-Standard surfaces with `Conflict` (`agent_ops/context.rs:38`). `initialize` succeeds — it never touches context (`server.rs:163-179`). So a cowork session's harness connects the server and then fails discovery, which most MCP clients surface as a visible "server failed / 0 tools" state inside the session rather than a silent no-op. Still acceptable given cowork is dead and dies in PR 8, but it is user-visible, not transparent.

### 8. The negative control on the selector is partial (low)

`mounts_for_sessions_the_subagents_mcp_used_to_skip` (`definition.rs:125-136`) covers two of the three dropped conditions (surface × `subagents_enabled`). The third — the old `find_subagent_parent(...).is_none()` guard — can't be covered in this shape because `should_attach` takes no service, so "restored selector conditions → test fails" only fires for two of three. I chased the uncovered one by hand and it is safe: a subagent child hits `SubagentError::DepthLimit` at `subagents/service.rs:120`, which `resolve_context` converts to `can_create=false` rather than an error, and every target-resolving tool is link-scoped to the caller. Noting it so the negative-control claim isn't read as broader than it is.

---

## Nitpicks

- `agent_ops/definition.rs:26` — INSTRUCTIONS still promises "the `proliferate.subagents.workflow` skill when this MCP is mounted". Grep finds no such skill anywhere in the repo (pre-existing string), but it now ships to *every* session instead of a subset.
- `agent_ops/calls.rs:214` — the notes array still tells agents subagents "do not inherit the parent's MCP bindings in this PR". Children now do receive the agent-ops product MCP; product MCPs were never "bindings", but the sentence reads wrong to a model.
- `agent_ops/definition.rs:125-136` — the mount test can only ever be vacuously true, since `should_attach` ignores its argument. Asserting through `select_product_mcps` with the real registration would make it non-vacuous (and would let it cover the third dropped condition).
- `agent_ops/` breaks the `domains/<domain>/mcp/**` file grammar that `servers.md` still mandates ("Product MCP code uses `mcp/`", `servers.md:121`, plus the whole Module Contract section). The README table was updated to the new path; the grammar rule was not. Either grant `agent_ops` an explicit exception in the doc or nest it as `domains/sessions/agent_ops/mcp/`.
- `sdk/src/reducer/transcript.ts:1421-1432` still omits `get_subagent_launch_options` and `list_subagents` from the subagent semantic-kind list (pre-existing). Cheap sweep while the list is being curated anyway.
- `product_registry.rs:135` — a definition listing its own `route_slug` inside its own `legacy_route_slugs` would hard-fail startup. Fail-closed and fine; just worth knowing it's the behavior.

pablonyx:
Re the automated P1 ("closed caller mutations bypass authorization" in wake/close paths): not reproducible. `agent_ops/context.rs:32` rejects a closed caller (`parent session is closed`) inside `resolve_context`, which runs fresh on every request before any tool dispatch — a lingering 12h token on a closed session cannot execute any tool, mutating or not. The reviewer-flagged docstring/test hardening landed in b3757efe (pushed after validation).

pablonyx:

PR #1727, opened 2026-08-08
feat(anyharness): agent messaging — envelope, send_agent_message, list_agents, transcript reads

pablonyx:
PR 2 of the agent operations layer (spec: adr-agent-ops.md; §6 step 2). Stacked on #1726 — base is `agent-ops/1-mcp-shell`.

## What

- **`prompt/envelope.rs`** (new): single owner of all agent-facing prompt copy. `EnvelopedPrompt` yields text + provenance as one value — what is sent is exactly what is stored, never divergent. `agent_message()` renders the ADR envelope (body byte-verbatim + reply instruction with the sender's session id); `subagent_wake()` carries today's wake string byte-identical (pinned by test); `subagents/hooks.rs` now calls into it.
- **`send_agent_message`**: message any session runtime-wide (unpromoted subagents included — ruling 3). Closed target → rejected; idle target boots via the ordinary pending-prompt path (nothing special built). Provenance `AgentSession { source_session_id, session_link_id: None, label }` — zero type changes. No `wakeOnReply` yet (PR 3). No loop protection (ruling 9).
- **`list_agents`** + new `search_sessions` store query: title/link-label LIKE + exact session-id lookup (chip titles resolve from a bare id), workspace filter, closed-inclusion flag, keyset cursor, limits 20/50.
- **`read_agent_transcript`**: the existing budgeted subagent reads (events / search / latest-turns) made session-scoped in a new `transcript_read.rs` — link-scoped paths become thin wrappers, budgets shared, closed sessions readable forever.
- **`agent_ops/peer_ops.rs`**: testable authorize+envelope seams for the send/read paths.
- **Queued-updates counts: nothing added server-side** — `PendingPromptSummary` already carries provenance end-to-end (`model.rs:241` → SDK reducer `pending-prompts.ts`), so the client "From agents · N updates" row groups on `sourceSessionId` from data already served.
- SDK/product-client: the three tool names classified for rendering; full UX is PR 9. Specs: `prompt-and-skill-policy.md` gains the peer-envelope amendment (ADR ruling 2) + pointer-not-output law.

## Tests

- Full `cargo test -p anyharness-lib`: **1573 passed; 1 failed** — the failure is the known pre-existing `sweep_tests` flake (fails on base).
- Scoped suites 47 passed; SDK vitest 56 + tsc clean; product-client 44 passed.
- Negative controls (each red, then restored): closed-target Send guard removed; provenance label dropped; wake copy mutated; cursor tie-break dropped; read gate flipped to Send intent.

## Notes for review

- PR 1's `authorize.rs` module doc claimed every mutating call still runs `resolve_target` first; the peer tools are deliberately the first callers with no link to resolve (runtime-wide reach, rulings 3/4). Doc amended in this PR to record that as the scope decision.
- Includes the one-line compile fix for PR 1's mount test (E0716), also cherry-picked into #1726.


pablonyx (commented):
**Verdict: issues found — 1 high, 3 medium, 6 low.** The envelope invariant, the byte-identical wake copy, the link-scoped wrappers, the keyset cursor and the spec amendment are all clean; I verified each against the base rather than trusting the description. The high finding is that `send_agent_message` is the first product-MCP path that can prompt an *arbitrary* session, and it lands outside the spec-2b workflow admission fence that every other prompt path clears.

Review is of the delta vs `agent-ops/1-mcp-shell` only. Read-only; no build run (another agent owns the machine's build slot), so nothing below is from executing tests.

---

## High

### H1 — `send_agent_message` prompts any session without a session-mutation admission permit (spec 2b fence bypass)

`agent_ops/calls.rs:523` → `agent_ops/peer_ops.rs:50` → `SessionRuntime::send_text_prompt_with_provenance` (`runtime/prompt.rs`). That path takes the workspace access gate and nothing else. It never calls `session_admission.acquire(target, SessionMutationKind::Prompt, SessionMutationSource::external())`.

Every other way to put a prompt into a session is fenced:

- `api/http/sessions_prompt.rs:47` — `admit_session_mutation(&state, &session_id, SessionMutationKind::Prompt)`
- `api/http/sessions_pending.rs:44,83,119,153` — `PendingPromptQueue`
- goals / loops / plans / reviews / config / resume / replay — all `admit_session_mutation` on the affected session id

The product-MCP route (`api/http/product_mcp.rs:72`) takes no permit at all. Until this PR that was sound, and the codebase says why — base `subagents/hooks.rs:67`:

> workflow-controlled sessions are created InternalOnly with subagents DISABLED, so no session link can exist whose parent is controlled; this wake path cannot target a controlled session and takes no admission permit

That is a *structural* argument: the old MCP prompt paths (`send_subagent_message`, the wake) can only reach a session on the far end of a `session_links` row, and a workflow-controlled session is never on either end. Ruling 4's unlinked runtime-wide reach deletes that argument, and nothing replaced it. The PR's `authorize.rs` doc amendment reasons carefully about *ownership* ("a peer message is an inbox item the receiver can ignore") but never mentions admission, which is a different invariant: it is about **who may perturb a session's execution while a workflow run owns it**, and it is documented as frozen and fail-closed (`domains/sessions/admission.rs`, PR1227-WORKSPACE-FENCE-01/02).

Concrete failure:

1. A workflow run creates its session via `create_persisted_internal_session` (`runtime/creation.rs:270`) — an ordinary `sessions` row: not closed, not dismissed, `InternalOnly`, `subagents_enabled = false`.
2. The run is executing a step; the executor holds that session's mutation permit.
3. Any agent calls `list_agents` — the row comes back (no surface, visibility, or `internal_only` filter in `search_sessions`) — then `send_agent_message` with its `sessionId`.
4. `authorize` passes (open target, open caller, runtime-wide). `send_text_prompt_with_provenance` enqueues onto the controlled session's pending queue and boots it if idle. `runtime/startup.rs` has no `InternalOnly` boot refusal.
5. The workflow's session runs an extra, unplanned turn from a peer prompt. Through the HTTP route the identical call is a clean 409 `ControlledByWorkflow`.

The purge/retire fence is affected by the same hole: those paths admit the workspace's whole session set and *hold the permits* precisely so nothing else mutates those sessions; a peer send takes no permit and slips past the mechanism the fence is built on.

Suggested fix: thread admission into the peer send — acquire `SessionMutationKind::Prompt` on the **target** session id with `SessionMutationSource::external()` before dispatch, and map `SessionMutationConflict::ControlledByWorkflow` to a clean tool error ("that agent is currently controlled by a workflow run"). If that is deliberately deferred, it needs to be a stated ruling in the ADR rather than silence, because PR 4 (`configure_agent`) and PR 5 (`close_agent`) both widen the same hole — `Config` and `Cancel`/`Close` are admission kinds too.

---

## Medium

### M1 — dismissed sessions are advertised as messageable, then fail opaquely at boot

`store/sessions.rs:193` — `search_sessions` filters on `closed_at` only. Compare `list_visible_by_workspace` (`store/sessions.rs:238`), which filters `dismissed_at IS NULL AND closed_at IS NULL`. `authorize::is_closed` (`authorize.rs:83`) is `closed_at.is_some() || status == "closed"` — also no `dismissed_at`.

But the boot path *does* refuse dismissed sessions: `runtime/launch_policy.rs:165` — `record.closed_at.is_some() || record.dismissed_at.is_some() || record.status == "closed"`.

Failure: the human dismisses session S from the sidebar. `list_agents` still returns S with `"closed": false`, i.e. a valid target. `send_agent_message(S)` clears `prepare_agent_message`, builds the envelope, then dies inside `ensure_live_session_handle`, surfacing as `anyhow::anyhow!("{error:?}")` — a Debug-formatted `StartSessionError` (`calls.rs:539`). The calling agent gets an unactionable internal error for a state that is both permanent-ish and knowable up front, and the human sees a session they deleted being messaged.

Fix: add `AND s.dismissed_at IS NULL` to `search_sessions` (with an `include_dismissed` flag only if something needs it), and give `authorize` a `TargetDismissed` rejection so the tool refuses cleanly.

### M2 — a cross-workspace send takes the caller's `SubagentWrite` lease, not the target's

`app/product_mcp.rs:131` registers the agent-ops MCP with `WorkspaceOperationKind::SubagentWrite`, and `api/http/product_mcp.rs:72` acquires it as `acquire_shared(workspace_id, kind)` where `workspace_id` is the **route's** workspace — the caller's. `assert_workspace_mutable(state, workspace_id)` is likewise the caller's.

You asked for reasoning on whether that is the right lease. It is not, for a cross-workspace send. What the lease protects is the *workspace whose state is being mutated*: `workspaces/retire.rs:167` and `workspaces/purge.rs:111` take `acquire_exclusive` on the workspace they are destroying, and `retire_preflight.rs:459` snapshots that workspace's gate for `SubagentWrite` among the "workspace work is still in progress" blockers. A send into workspace B mutates B — it enqueues a durable row against a B session and can boot a B actor with a harness child process inside B's checkout — while holding only A's shared lease. B's retire preflight cannot see it.

Race: agent in A calls `send_agent_message` at S in B. B's retire is in flight. Preflight snapshots B's gate → no `SubagentWrite` holder → not blocked. The send's `assert_can_mutate_for_session(S)` reads B's workspace row microseconds before retire stamps it, then `ensure_live_session_handle` boots S's harness process into a checkout retire is concurrently tearing down.

The target-side access gate (`runtime/prompt.rs`: `assert_can_mutate_for_session`) closes the *steady-state* case — a send into an already-retired/read-only workspace is refused — so this is a window, not an open door. But the lease is exactly the machinery that exists to close windows, and it is now attached to the wrong workspace. Right shape: acquire `SubagentWrite` on the **target's** workspace for `send_agent_message` (in addition to, or instead of, the caller's), which needs the lease decision to move from the static `MUTATING_TOOL_NAMES` table to something that can see the resolved target. Worth a comment in `tools.rs` recording the gap even if the fix is deferred, because `tools.rs:508`'s existing note explains the alias/lease coupling and a reader will reasonably assume the lease is complete.

### M3 — the new peer reads are not bounded the way "budgeted" implies

Two spots, both inherited from the subagent path but with materially different blast radius now that the target is any session rather than a short-lived capped-at-8 linked child.

`transcript_read.rs:77` — `read_session_events` calls `session_store.list_events_after(session_id, after_seq)`, whose SQL (`store/events.rs:320`) has **no LIMIT**; the truncation to ≤100 happens in Rust afterwards. `read_agent_transcript(mode=events, sinceSeq=0)` against a long-lived session materializes every `session_events` row into `Vec<SessionEventRecord>` to return 50. The store already has the correct primitive: `list_events_after_oldest_limited` (`store/events.rs:360`), same signature plus `limit`. One-line fix.

`transcript_read.rs:145` — `read_session_latest_turns` calls `list_events_for_latest_turns(..., LATEST_TURN_EVENT_BUDGET)`. That budget is a back-off target, not a cap: the loop (`store/events.rs:~200`) narrows the turn window until `event_count <= event_limit` **or `selected_turn_count <= 1`**, then selects `WHERE seq >= cutoff_seq` unconditionally. A session with one enormous turn returns that whole turn. `read_session_latest_turns` then does `records.iter().filter(...).cloned()` once per turn — up to 10 full clones of that vector. Previously reachable only for a subagent; now for the user's main session.

Neither is a correctness bug on normal data. Both are the kind of thing that turns "an agent read a peer's transcript" into a memory spike on a real workstation.

---

## Low

### L1 — nothing stops an agent messaging itself

`authorize` (`authorize.rs`) has no `caller != target` check, and neither does `prepare_agent_message`. `send_agent_message` with your own `sessionId` enqueues onto your own pending queue; the actor accepts it via `handle_busy_prompt_queue` while busy (`live/sessions/actor/turn/active.rs:~240`) so there is no deadlock, but it drains the moment the turn ends and wakes the agent with `Message from agent "X" (session ses_self)`.

Ruling 9 declines loop protection on the grounds that "a real loop needs both agents choosing to re-message every turn". A self-send needs one. The ADR's own design elsewhere excludes the self case structurally — `session_links` has `CHECK(parent_session_id != child_session_id)` and the proposed `session_wake_schedules` has `CHECK (watcher_session_id != target_session_id)`. A one-line `if caller == target` rejection in `authorize` for `Send` would match that.

### L2 — LIKE wildcards in the search string are not escaped

`store/sessions.rs:~186` — `format!("%{}%", value.to_lowercase())`, bound as `?4` against `LIKE ?4` with no `ESCAPE`. No injection risk (it is a bound parameter), but the agent's search string is interpreted, not matched: `%` matches everything, `_` matches any single character. `list_agents(query: "deploy_checker")` also matches "deployXchecker". Fix is the usual `value.replace('\\',"\\\\").replace('%',"\\%").replace('_',"\\_")` plus `ESCAPE '\'`.

### L3 — `LOWER()` is ASCII-only, `str::to_lowercase()` is not

Same lines. Rust lowercases the needle with full Unicode; SQLite's built-in `LOWER()` only folds ASCII. A session titled `École` is unreachable by any query: `"École"` → needle `école`, haystack `École`; `"école"` → needle `école`, haystack `École`. (SQLite `LIKE` is already ASCII-case-insensitive by default, so the two `LOWER()` calls buy nothing and cost this.)

### L4 — the search's closed test is narrower than the authorize one

`search_sessions` uses `closed_at IS NULL`; `authorize::is_closed` uses `closed_at.is_some() || status == "closed"`. A row in the `status='closed'`-but-`closed_at IS NULL` state is listed by `list_agents` as `"closed": false` and then rejected by `send_agent_message` as `TargetClosed`. I could not find a production writer that produces that state (only `launch_policy.rs:421` in a test), so this is consistency rather than a live bug — but the two predicates should be one function.

### L5 — no surface/visibility filter on the peer surface

`context.rs:36` requires the **caller's** workspace to be `WorkspaceSurface::Standard`. Nothing constrains the target: `search_sessions` returns sessions in cowork and review workspaces and `InternalOnly` sessions alike, and `send_agent_message` will prompt them. Related to H1 but broader; at minimum `list_agents` probably should not be enumerating review-agent and workflow sessions as peers.

### L6 — `list_agents` runs one `find_subagent_parent` per row

`calls.rs:~486` — up to 50 extra queries per call to resolve link labels. Fine at this scale; a single `IN (...)` lookup would be the obvious shape if the page size ever grows.

---

## What I verified as clean

- **Envelope invariant.** `EnvelopedPrompt`'s fields are private, `text()`/`provenance()` are `#[cfg(test)]`, and the only production seams are `into_parts()` / `into_payload()` — both consume an already-paired value. `agent_message()` and `subagent_wake()` are the sole constructors. `peer_ops.rs:50` takes both halves from one call and `calls.rs:531` feeds them to the same dispatch. `the_target_receives_the_envelope_and_the_row_stores_exactly_that_text` pins that the stored `text_summary` is the sent text.
- **Wake copy byte-identical.** Diffed `envelope.rs:agent_message`/`subagent_wake` against base `subagents/hooks.rs:wake_prompt_text`: format string, both fallbacks (`"subagent"` / `"unknown"`), `outcome.as_str()` position and the `PromptPayload::text(..).with_provenance(SubagentWake{link, completion, label})` construction are all identical. `the_subagent_wake_pointer_copy_is_unchanged` pins the literal, and the fallback test pins the `None`/`None` case.
- **Body verbatim.** `agent_message` interpolates `body` with no trim/escape; `agent_message_names_the_sender_quotes_the_body_verbatim_and_gives_the_reply_handle` deliberately passes leading/trailing whitespace and an embedded blank line and asserts they survive. The emptiness check trims a copy only.
- **Provenance label is unspoofable.** It comes from `AgentMessageSender::from_session(&access.caller)`, and `access.caller` is loaded by `authorize` from the store using the session id the capability token proved. `SendAgentMessageArgs` carries only `sessionId` + `message` — no label field to forge. Title→`agent_kind` fallback is trimmed and tested.
- **Policy.** Closed target rejected for `Send`, allowed for `Read` (`authorize.rs`, plus `a_closed_agents_transcript_stays_readable` asserting both on the same fixture). Closed caller rejected at the context layer (`context.rs:32`) *and* in `authorize`. `build_tool_list` offers the three peer tools unconditionally, with `peer_messaging_and_reads_are_offered_to_every_caller` exercising the no-spawn-rights context — ruling 3 holds. No workspace check on the peer path; `an_open_target_in_another_workspace_is_reachable` pins ruling 4.
- **Cursor.** `(updated_at, id)` appears in both the WHERE (`updated_at < ?5 OR (updated_at = ?5 AND id < ?6)`) and the ORDER BY (`updated_at DESC, id DESC`); directions match. `|` genuinely cannot appear in the timestamp half, `decode` rejects empty halves. `include_closed` defaults false. Limit is clamped in `list_agents` *and* again in `search_sessions`, so a direct store caller can't exceed 50. `session_search_pages_without_skipping_rows_that_share_a_timestamp` is a real tie-break test (three rows, identical `updated_at`).
- **transcript_read refactor is not a widening.** `delegation.rs:read_child_events` still resolves the link before delegating; `subagents/service.rs` `read_events` / `read_latest_turns` / `search_transcript` all still call `resolve_target_including_closed` first. `read_latest_turns` deliberately keeps its completion-row summarization (the new event-derived path is peer-only). All eight budget constants carry the same values across the move.
- **Naming.** `definition.rs` keeps `ACP_SERVER_NAME = "subagents"` with an explicit comment that the rename waits for the client step — so the `mcp__subagents__*` strings the SDK reducer and `subagent-tool-presentation.ts` match are correct, and no existing name changed classification. The `formatSubagentHeaderVerb` additions sit above the `executionState === "failed"` branch, so they correctly pre-empt the "Subagent launch failed" / creation fallback.
- **Spec accuracy.** The envelope block in `prompt-and-skill-policy.md` matches `envelope.rs:84`'s format string byte for byte, including the quoting of both the label and the reply `sessionId`. The wake block is now *corrected* to the shipped string (the previous doc had drifted). The pointer-not-output law and the owner-path swap are accurate.
- **Queued-updates claim.** Accurate: `PendingPromptRecord::to_contract` (`model.rs:241`) emits `prompt_provenance`, and `sdk/src/reducer/pending-prompts.ts` threads `promptProvenance` through all four event branches. Nothing server-side was needed.
- **Named tests exist and pin what the description claims**, including all five negative-control anchors: closed-target guard (`a_closed_target_is_rejected_before_any_prompt_is_built` + `closed_target_rejects_sends_but_stays_readable`), provenance label (`agent_message_provenance_carries_the_same_label_the_text_rendered`), wake copy (`the_subagent_wake_pointer_copy_is_unchanged`), cursor tie-break (`session_search_pages_without_skipping_rows_that_share_a_timestamp`), read intent (`a_closed_agents_transcript_stays_readable`).

---

## Nitpicks

- The pending-prompt row stores the **full envelope** as `text`, and `PendingPromptSummary.text` serves it to the client. §4's "no message preview, you see THAT not WHAT" therefore rests entirely on the client choosing not to render a field it is handed. Worth a line in the PR 9 spec so nobody re-derives it.
- ADR §2 (line 163) states the budgets as "latest-turns 5/20, search 10/50, events 1000/5000". The actual constants, now canonical in `transcript_read.rs:12-19`, are 3/10, 10/25, 50/100. Pre-existing ADR drift, but this is the PR that makes them the shared budget, so it's the natural place to correct it.
- The budget constants widened from `pub(super)` to `pub` in the move. `pub(crate)` would keep them where they belong.
- `calls.rs:~455`: `"cursor is not a cursor returned by list_agents"` reads oddly; "cursor is malformed — pass the nextCursor from a previous list_agents call" says the same thing to an agent that has to act on it.

pablonyx:
Review findings addressed in 0388a610e (all validated locally: scoped suites 111 tests green; the pre-existing `sweep_tests` red is unrelated).

**Per-finding:**
- **H1 (admission fence)** — `send_agent_message` now clears the session-mutation fence itself: new `admit_peer_send` acquires the `Prompt` permit on the **target** with `SessionMutationSource::external()`; `ControlledByWorkflow` maps to a clean tool error ("controlled by an active workflow run…"), run id logged, never returned. Permit held across dispatch.
- **M1** — dismissed targets refused (Send-only; stays readable, matching the closed-target precedent). **M2** — target-workspace `SubagentWrite` lease taken in-call (see deviation below). **M3** — transcript reads bounded in SQL (`list_events_after_oldest_limited` + LIMIT on the latest-turns selects).
- **L1** self-send refused (Send-only) · **L2** LIKE metacharacters escaped (`ESCAPE '\'`) · **L3** no `LOWER()` · **L4** closed predicate now `closed_at IS NULL AND status <> 'closed'`, matching `authorize::is_closed` · **L6** `find_subagent_parents` batched (`IN (…)`), no N+1.

**L5 ruling disclosure:** `internal_only` sessions are refused for **both intents** — not just Send. An `internal_only` session is runtime plumbing, not a peer; refusing it as a read target too keeps discovery (`list_agents`/`search_sessions` filter it) and access from ever disagreeing. This is deliberately stricter than the closed/dismissed precedent (those stay readable).

**Deviation from the review's M2 prescription:** `send_agent_message` was **removed** from `MUTATING_TOOL_NAMES` instead of gaining a second lease. Implementing H1+M2 literally — route lease (caller workspace) before the call body, permit inside it — reverses the PR1227-LOCK-01 order (`session mutation permit → workspace operation lease`), which is exactly the ABBA order `api/session_admission_tests.rs::reversed_order_deadlocks()` proves hangs against retire/purge. Shipped shape: no route lease; in-call **permit → target-workspace SubagentWrite lease → dispatch**, both held across dispatch. Consequences: canonical lock order holds with exactly one workspace lease; M2's protection now covers every send (not only cross-workspace); the caller-workspace check survives as a lease-free access assert. A new source-order guard `the_peer_send_takes_the_permit_before_the_target_workspace_lease` pins the ordering in the file that enumerates dual-lock sites.

pablonyx:

PR #1728, opened 2026-08-08
feat(anyharness): session-scoped wakes — schedule_agent_wake, wakeOnReply, human arming route

pablonyx:
PR 3 of the agent operations layer (spec: adr-agent-ops.md; §6 step 3). Stacked on #1727.

## What

- **Migration 0067**: `session_wake_schedules` exactly per ADR §3.2 — pair PK (double-arm = no-op), `CHECK (watcher != target)`, target index; schema snapshot regenerated.
- **Store**: `arm_agent_wake` (INSERT OR IGNORE), `consume_agent_wakes_for_target` — ONE transaction: select watchers → delete by target → enqueue one pointer prompt per watcher (durable pending-prompt rows; offline watchers need nothing). Session deletion clears wake rows on both sides. Link-scoped machinery untouched.
- **Turn-finish consumption**: `wakes/hooks.rs` SessionExtension — any session finishing a turn consumes its watchers atomically; live watchers get the queued prompt nudged, offline ones wake on next boot. Ruling 10 ("covers the current turn") needs no mechanism: consume-at-turn-finish means a mid-turn arm is an ordinary row by turn end — pinned by test.
- **Pointer copy** in `envelope.rs` (`agent_wake`): ADR shape, constant across outcomes, never turn output. New provenance variant `AgentWake { target_session_id, label }` → `agentWake` on the wire (neither `SubagentWake` nor `LinkWake` fits — no link/completion rows exist to reference, and fabricating ids is forbidden).
- **`schedule_agent_wake`** tool + **`wakeOnReply`** on `send_agent_message`: arm before dispatch (compensated on failure), reply-consume after success — a crash in the gap costs one redundant pointer rather than a lost schedule.
- **Human arming route**: `POST /v1/sessions/{id}/wakes/{target_id}` in LOCK-01 order with a new `SessionMutationKind::AgentWake`; contract + scripted SDK regen included (`scheduleAgentWake`).
- Specs: wake-scopes law in prompt-and-skill-policy, sessions.md session-scoped wakes section, contract.md records mobility archives deliberately omit the table.

## Tests

- Full `cargo test -p anyharness-lib`: **1597 passed; 1 failed** — the known pre-existing `sweep_tests` flake.
- Scoped: 70 across wakes/envelope/agent_ops/deletion/admission; SDK 42 + typecheck clean; product-client 10.
- Negative controls: de-atomizing the consume transaction, removing the self-wake guard, de-idempotenting the arm — each red. The fourth control (deleting reply-consumption from the send path) **failed to fail** — a real coverage hole; fixed with a source-order test pinning arm-before-dispatch / consume-after-dispatch, and the mutation now goes red.

## Notes

- Reply-consume is not inside the send's transaction because no such transaction exists (the send is a runtime call that may boot an actor); ordering + compensation chosen so failure never loses a schedule.
- No pane UI for human arming yet — route/SDK only; UI is PR 9.


pablonyx (commented):
**Verdict: COMMENT — solid core, one High to resolve before merge (H1), plus five Mediums.** Reviewed the delta vs `agent-ops/2-messaging` @ `5a10829`. The consume transaction, the cursor/queue-position bookkeeping, the deletion ordering, the migration, and the SDK artifacts are all correct and genuinely pinned. The problems are at the edges of the new machinery: who the *fired* pointer is allowed to land on, what the reply-consumption cancels, and the paths where a turn ends without the hook firing.

Per the PR2 review I am not re-reporting H1 (peer sends bypassing the workflow-admission fence) or M2 (cross-workspace lease) — except where these new paths widen them, which H1 below does.

---

## What is right (checked, not assumed)

- **The consume transaction is genuinely one transaction and genuinely serialized.** `Db` is a single `Connection` behind an `Arc<Mutex<..>>` (`persistence/sqlite.rs:9-11,58-70`), so `consume_agent_wakes_for_target`'s select → delete → per-watcher enqueue cannot interleave with a concurrent `arm_agent_wake` (which takes the same mutex via `with_conn`) or with another target's consume. Two turn-finishes of the same target are impossible anyway (one actor, serialized turns), and the second one finds zero rows. `?` inside `with_tx` drops the `Transaction` → rollback, and `a_failed_enqueue_rolls_the_whole_consumption_back` (`store/agent_wakes.rs:353-420`) proves it with a real `RAISE(ABORT)` trigger on the *second* watcher. That is a real negative control.
- **Cursor/queue-position bookkeeping matches the canonical idiom byte-for-byte** — `store/agent_wakes.rs:145-171` is the same `UPDATE ... cursor + 1` → `SELECT cursor` → `COALESCE(MAX(queue_position),0)+1` → INSERT sequence as `store/pending_prompts.rs:30-62` and `store/link_completions.rs:116-148`. No corruption risk for human prompt ordering.
- **Deletion**: `delete_agent_wake_rows_for_session_in_tx` runs before `DELETE FROM sessions` (`store/sessions.rs:614-618`), clears both sides, and there is exactly one `DELETE FROM sessions` in the codebase, so no FK path is missed. The seeded both-directions rows in `deletion_tests.rs` pin it.
- **Migration**: `0067` is the next free number on both `main` and the base branch; no `ON DELETE CASCADE` (matches ADR §3.2 exactly); `INSERT OR IGNORE` does not swallow FK violations in SQLite (ON CONFLICT does not apply to FK), so a bogus session id still errors.
- **LOCK-01**: the route's order is `assert_session_auth_scope → admit → acquire_shared → assert_workspace_mutable`, identical to `subagents.rs:70-82`, and the ratchet row was added (`session_admission_tests.rs:615`).
- **Admission is taken on the WATCHER, which is the right session** — arming perturbs nothing on the target, and the pointer prompt lands on the watcher. Correct call.
- **SDK regen is consistent**: `openapi.json` has both the new path and `agentWake` in the `PromptProvenance` `oneOf`, `openapi.ts` has the path + `type: "agentWake"`, and the client method/type exports line up. Contract serde (`tag="type"`, camelCase → `agentWake`) matches the sibling variants; internal serde (`tag="kind"`, snake_case → `agent_wake`) likewise.

---

## Findings

### H1 — A fired wake injects a prompt into a workflow-controlled session with no admission, and the PR2 fence fix will not cover it
`wakes/hooks.rs:40-44`, `wakes/service.rs:54-77`, `agent_ops/tools.rs:203-213`

The "Spec 2b classification (admission:derived-safe)" comment justifies taking no permit "for the same reason the link-scoped wake next door does not". That reason does not transfer. The subagent hook's argument (`subagents/hooks.rs:66-71`) is *structural*: workflow sessions are created with `subagents_enabled = false` (`runtime/creation.rs:302-305`), and `validate_parent_can_spawn` reads that flag, so no `session_links` row can ever have a controlled parent. Session-scoped wakes have **no analogous gate**: `agent_ops/definition.rs:should_attach` returns `true` unconditionally (with a test pinning that it mounts even when `subagents_enabled = false`), and `schedule_agent_wake` is offered to every caller with no gate at all — `tools.rs`'s own `session_scoped_wakes_are_offered_to_every_caller` pins exactly that.

Failure scenario:
1. Workflow run W controls session S; step 1's agent calls `schedule_agent_wake(sessionId = T)` (or `send_agent_message(..., wakeOnReply: true)`). Nothing rejects it — the arm takes no permit and `authorize` has no controller check.
2. Steps 2-3 run. T (any session in the runtime — the human's chat, another agent) finishes a turn.
3. `consume_agent_wakes_for_target` writes a `session_pending_prompts` row for S and `hooks.rs:56-70` hands it straight to S's live handle → **S runs a turn the controller never admitted**, mid-run.

Every other external prompt into a controlled session takes `SessionMutationKind::Prompt` and gets a stable `409 SESSION_CONTROLLED_BY_WORKFLOW`. This one does not, and it cannot be fixed at the send path: the injection happens *later*, from a hook, so whatever PR2 lands for peer sends will not close it. The human route is fine here — it admits on the watcher and would 409 correctly. This is agent-armed only.

Suggested fix (cheap, and it reuses an argument already blessed in this codebase): at consume time, filter watchers whose session has an active workflow controller. That is a pure read-only controller lookup acquiring neither permit nor lease — exactly the FENCE-01/02 re-check shape documented in `admission.rs:36-42` — so it adds no edge to the canonical lock order and cannot ABBA. Alternative: refuse in `AgentWakeService::arm` when the CALLER is controlled. Whichever you pick, the `hooks.rs:40-44` comment needs rewriting, because as written it asserts something false about this path.

### M1 — The human route never scopes the TARGET; a workspace-scoped token can arm across workspaces (raise to High if direct-attach scoping is a real boundary for you)
`api/http/sessions_wakes.rs:46`, `api/http/access.rs:70-88`, `authorize.rs:1-27`

`assert_session_auth_scope` is called on `session_id` (the watcher) only. `authorize()` is explicitly runtime-wide with no workspace scoping. So a `UserClaim` scoped to workspace A can arm a wake from a session it legitimately controls onto **any session id in the runtime**, in any workspace. Two consequences:

- **Existence/state oracle**: 404 `SESSION_NOT_FOUND` vs 409 `SESSION_CLOSED` vs 200 distinguishes, for an arbitrary session id, "doesn't exist" / "exists and is closed" / "exists and is open".
- **Metadata exfil into a session the caller does control**: when the target's turn ends, the pointer text carries the target's **title** and turn outcome (`envelope.rs:129-137`) into the watcher's transcript.

The subagent route has no equivalent hole because the child must be a resolved link child. Ruling 4 ("reach is runtime-wide") is a statement about *agents*; a direct-attach token is a deliberately narrowed credential and this route is the first place two session ids cross without both being scoped. Minimum fix: `assert_workspace_auth_scope(&auth, &target.workspace_id)` after resolving the target (needs the target resolved before the arm, which the service already does internally).

### M2 — `consume_reply_wake` fires on ANY message, not on a reply — it can cancel a wake that was armed for turn *completion*, deadlocking the watcher
`agent_ops/peer_ops.rs:87-103`, `agent_ops/calls.rs:585-586`

The schedule row is `(watcher, target, created_at)` — it carries no record of *why* it was armed, so the send path cannot tell a `wakeOnReply` schedule from a `schedule_agent_wake` schedule. Any successful `send_agent_message` from C to T disarms `(T watches C)` unconditionally.

Failure scenario:
1. Agent T asks agent C to do a long job and calls `schedule_agent_wake(sessionId = C)` — T wants the *completion* pointer.
2. C, mid-job, sends T a courtesy note: "starting now, will report back". That send calls `consume_reply_wake(C, T)` → T's schedule is deleted.
3. C finishes its turn. Nothing fires. T sits idle forever waiting for a wake that no longer exists, with no error anywhere — `consumedPendingWake: true` is reported to C, who never asked for it.

The spec text you added says "when an agent **answers** a peer that was waiting on it" (`sessions.md`, prompt-and-skill-policy) — the code cannot establish "answers". Note this trades the wrong way by your own reasoning in `peer_ops.rs:73-86`: a redundant pointer is the cheap side, a lost wake is the expensive side, and this line spends the expensive side to save the cheap one. Options: drop the consumption entirely (a redundant pointer costs one constant-size prompt), or add an `armed_reason` column so only `wake_on_reply` schedules are reply-consumable.

### M3 — A turn that ends without a `turn_id` never consumes wakes; an actor crash never consumes them at all
`wakes/hooks.rs:50-52`

`if ctx.turn_id.trim().is_empty() { return Ok(()); }` — the error branch of the actor builds `turn_id` from `sink_snapshot_on_error.current_turn_id.clone().unwrap_or_default()` (`live/sessions/actor/turn/finish.rs:225-232`), which is empty whenever the turn failed before a turn opened: harness spawn failure, provider auth failure, connection error at prompt start. Those are precisely the outcomes a watcher most wants to hear about, and the wake silently does not fire. The schedule survives, so it fires on the target's *next* turn — but a session that just failed to start may never run again, and nothing surfaces the miss.

Same for a killed actor / runtime restart mid-turn: `on_turn_finish` never runs, so the schedule persists across the restart and waits for a turn that may never come.

The guard is copied from `subagents/hooks.rs:47-49`, where the blast radius is one parent watching its own child. Here any session can be waiting on any session. Either fire on empty `turn_id` (the pointer's `Outcome: failed` is exactly the right copy) or document the miss explicitly in `sessions.md` alongside the "one transaction" guarantee — right now the spec reads as if every finished turn consumes.

### M4 — Closed sessions: pointers are enqueued into closed watchers, and closed targets leave orphan schedules
`wakes/service.rs:54-77`, `store/agent_wakes.rs:112-190`

Wake rows are cleared only on session **deletion**, never on close. Two consequences:

- **Closed watcher**: W arms on T, then W is closed. T's next turn writes a `session_pending_prompts` row into W. W never boots, so the row sits forever, and it violates the ADR's own invariant (ruling 6: "message to a closed session → rejected") — this path writes straight to the store, bypassing `authorize`. It will also show up in §4's "N updates" pending counts once PR 9 lands.
- **Closed target**: W arms on T, then T is closed. T never finishes another turn, so the row is unconsumable and unreachable — `arm` refuses closed targets precisely so a schedule never "silently never fires", but closing after arming produces exactly that state.

Fix: clear session wake rows on both sides in the close path, symmetric with the delete path; or have `consume_...` skip closed watchers.

### M5 — Failure compensation can delete a schedule a concurrent successful send legitimately armed
`agent_ops/calls.rs:552-582`

The `wake_created` guard means "my INSERT materialized the row", not "no one else depends on this row". After the insert the row is shared state.

Failure scenario (two parallel `send_agent_message` tool calls in the same assistant turn — harnesses do issue parallel tool calls — or one tool call plus the human route):
1. Call A: `arm(C, T)` → `created = true`. Dispatch begins (may boot an actor: slow).
2. Call B: `arm(C, T)` → `created = false`. Dispatch **succeeds**. B reports `wake.scheduled: true`.
3. Call A's dispatch fails → `wake_created` is true → `disarm(C, T)` deletes the row.
4. B's wake is gone. B never compensates (it never created), C never learns, and C waits forever.

The PR body's claim that compensation "only removes what this call created" is true of the INSERT and false of the DELETE. If you keep compensation, it needs to be conditional on nothing else having claimed the pair (a refcount or an `armed_by_call_id` is heavier than this feature deserves) — or just drop the compensation and accept the redundant pointer, which is the same trade you already argue for in `peer_ops.rs:73-86`.

### M6 — `schedule_agent_wake` on an **idle** target is a silent liveness trap, and the doc overclaims the opposite
`agent_ops/calls.rs:618-633`, `specs/.../subagents.md` ("The peer-scoped `schedule_agent_wake` has no such race")

Consume-at-turn-finish closes the race in one direction (arm during a running turn → fires). It opens it in the other: if the target is idle — or its turn ended microseconds before the arm committed — the schedule fires only if *someone else* prompts that session again. For an agent that has finished its work and gone idle, that is never. The watcher blocks with no error, no timeout, and no diagnostic; the tool response says `"scheduled": true, "wakeScope": "next_turn_finish"` and says nothing about whether a turn is running.

Concretely: A calls `list_agents`, sees B, arms a wake, and waits. B finished 200ms earlier. Both agents are now idle forever.

Two cheap mitigations: return the target's current status in the tool result so the caller can decide (and let the tool description say "if the target is idle, this fires only when something prompts it again"), and soften the `subagents.md` sentence — the race isn't gone, it's inverted, and the inverted side is the one that hangs.

### L1 — `agentWake` provenance is unhandled client-side, so the pointer renders as a human-authored prompt
`apps/packages/product-client/src/domain/chats/subagents/provenance.ts:22-26`

`isSubagentWakeProvenance` matches only `subagentWake` / `linkWake`. As of this PR agents can arm wakes, so `agentWake` prompts start appearing in real transcripts — and they fall through to the plain `user_message` arm in `TranscriptItemBlock.tsx:87` (rendered as if the human typed the pointer text) and to the generic branch in `pending-prompt-queue.ts:56` (editable and deletable, with the whole pointer paragraph as the queue label). PR 9 owns the chip, but this PR does touch product-client for the tool label, and the interim rendering directly contradicts §4's "UI renders from provenance, never by parsing text". A one-line extension of the recognizer (label-only, no completion row) would carry the interim.

### L2 — Internal store errors on the human route surface as `400 BAD_REQUEST`
`api/http/sessions_wakes.rs:86-88`

`AgentAccessError::Internal` is `#[from] anyhow::Error`, so it carries *both* the self-wake sentinel and every store failure that came up through `?`. A locked DB while arming returns `400 BAD_REQUEST` with the raw error string in the body. Give the self-wake its own typed variant and map `Internal` to 500.

### L3 — `INSERT OR IGNORE` swallows the CHECK, not just the PK
`store/agent_wakes.rs:44-52`

Your own test documents this (`agent_wakes.rs:305-311`): a self-wake insert reports a plain no-op, i.e. `alreadyScheduled: true`, which is a lie. The service guards first so nothing hits it today, but `ON CONFLICT (watcher_session_id, target_session_id) DO NOTHING` makes the intent exact and lets a CHECK violation raise.

### L4 — An agent cannot see or cancel its own wakes
`list_agent_wakes_for_watcher` exists and is only used by tests; `disarm` is reachable only from the reply path and the compensation path. An agent that armed a wake it no longer needs has no way to drop it, and no way to ask what it is waiting on. Not required by the ADR — flagging as a gap the pointer copy will make people ask about.

---

## Test verification

| Claim | Status |
| --- | --- |
| `a_schedule_armed_mid_turn_fires_at_that_turns_end` (`wakes/service.rs:216`) | Exists; **does not pin what it claims.** The `update_status("ses_target", "running")` on line 223 is inert — `consume_for_finished_turn` never reads status, so the test passes identically with the line deleted, and it exercises no hook. It documents ruling 10; it does not defend it. The hazard it should be guarding (arm landing *after* the hook already ran) is M6 and is untested. |
| Atomicity control (`store/agent_wakes.rs:353`) | Real. Trigger-based failure on the second watcher, asserts both the first watcher's DELETE and INSERT rolled back. This one earns its keep. |
| Source-order test (`peer_ops.rs:211`) | Works, but the window is wrong: it slices to the next `\nasync fn `, and the next `async fn` after `send_agent_message` is `get_subagent_status` ~110 lines later — so the window swallows `cleanup_agent_wake_after_failed_dispatch`, `schedule_agent_wake` and `read_agent_transcript`. It survives today only because `schedule_agent_wake`'s `.arm(` sits *after* the dispatch token, so deleting the arm still goes red. Slice on `\n}` at column 0, or on `\nfn `/`\nasync fn `. |
| LOCK-01 ratchet row | Present and correct (`session_admission_tests.rs:615`). |
| HTTP route behavior | **No test at all** — the route has only the source-order ratchet entry. M1 (unscoped target) and L2 (400-on-internal) would both have been caught by one. |

## Nitpicks

- `envelope.rs:132` — "completed a turn. Outcome: failed." reads oddly for failed/cancelled. The ADR pins this string, so leave it, but "finished a turn" would cost nothing and match all three outcomes.
- `wakes/service.rs:64-69` — the self-wake refusal is `AgentAccessError::Internal`, which reaches the agent as a generic error and the human as a 400 (L2). It is an invalid argument, not an internal fault.
- `MUTATING_TOOL_NAMES` including `schedule_agent_wake` (`tools.rs:25`) means arming takes the caller's workspace write lease and fails when that workspace is non-mutable — for a write that touches no file. Consistent with the subagent wake, so not a change request, just noting that a pure DB row is now gated on workspace writability.
- `hooks.rs:56-70` — worth a line on what happens when the watcher is mid-boot: `get_handle` may return a handle whose queue is being drained from the same durable row, and the seq marker is what makes that idempotent. The subagent hook has the same shape, but the reasoning isn't written down in either place.
- Spec/code divergence to reconcile once M2 is decided: `sessions.md`'s "A real reply consumes the schedule" describes a reply; the code consumes on any message.

pablonyx:
**Review findings addressed** — `39f5c7e83`.

**H1 (unadmitted prompt into a controlled watcher).** Consumption now filters workflow-controlled watchers and leaves their rows **armed**, so the wake lands after the run releases control instead of being injected into a session `admit_peer_send` would 409. The controller lookup is the set form of the existing `find_workflow_controlled_session` and is taken *before* the transaction: a pure read-only policy call, no permit and no lease, so it adds no edge to the canonical `run gate -> permit -> operation lease` order (PR1227-LOCK-01). It is passed in as a **denylist** rather than an allowlist on purpose — the transaction still consumes every other row that exists at turn finish, so ruling 10 is untouched. `a_workflow_controlled_watcher_is_not_prompted_and_keeps_its_schedule` proves the schedule survives and fires on a later turn.

**M1.** The route now scopes the **target** as well as the watcher. Runtime-wide reach is the AGENT contract (`authorize`, ruling 4) and cannot be the whole gate for a workspace- or session-scoped human token. Everything the route may not serve is answered with the same `404 SESSION_NOT_FOUND` an unknown id gets, so the status spread is not an existence oracle and a scoped token cannot arm across workspaces and then read the target's title out of the fired pointer.

**M2.** Migration 0067 gains `armed_for_reply`; a reply consumes only a reply arm. **Upgrade rule: explicit outranks reply.** Re-arming an existing pair only ever moves the flag 1→0, never 0→1 — so a `wakeOnReply` send can never silently make a human's standing `schedule_agent_wake` cancellable by an incidental message.

**M3 ruling: consume.** The empty-`turn_id` early return is gone. A session-scoped schedule is keyed on the session pair and records nothing about which turn it fires for, so an actor whose connection failed before the turn opened is still a finished turn from the watcher's side, and the pointer already carries `Outcome: failed`. The link-scoped hook next door keeps its guard because it writes `child_turn_id` into a completion row. **Honest caveat, now in the code and the spec:** this fires from the actor's finish path, so a runtime that dies mid-turn without reaching it consumes nothing and the schedule stays armed for the target's next finished turn.

**M4.** Close is handled at consume time: a closed target's schedules are swept rather than left forever unconsumable, and a closed watcher's row is dropped without a pointer (ruling 6).

**M5 semantic: `dispatch_confirmed_at`, correctness in both race orders, no residual window.** Two sends to the same target share one row (the pair is the primary key), so `created` says nothing about whether a *landed* send depends on the schedule. A send that lands stamps `dispatch_confirmed_at` via an upsert scoped to a live, non-closed target; compensation only removes an arm that is still unconfirmed. If the loser compensates first, the winner's upsert re-creates the row.

**M6.** `schedule_agent_wake` returns `targetStatus`, `targetRunning` and `firesWhen`, and the tool description says plainly that an idle target finishes nothing until someone prompts it. The spec no longer claims this path "has no such race" — it closes the same race from the other side.

**Lows.** `AgentAccessError::Internal` is a 5xx now instead of a 400. `INSERT OR IGNORE` is replaced by a transactional read-then-insert so the self-pair CHECK errors instead of being swallowed, with a store-level test asserting the error.

**Test quality.** The inert `update_status("running")` case was rewritten to key on the row existing at turn finish. New behavioral HTTP tests cover the success shape and `alreadyScheduled`, the 400/404/409 refusal matrix over the real router, and target scope with an in-scope negative control. **Every guard above was negative-controlled:** reverting the H1 skip, the reply-only consumption, the target scope check, the closed-watcher branch and the confirmation filter each failed exactly the tests written for them and nothing else.

**Not in this commit, deliberately:** client-side `agentWake` rendering (PR9's scope) and a list/cancel-own-wakes tool. The OpenAPI annotations are unchanged — the wire shape does not move, so no SDK regeneration is required; the checked-in-document test confirms it.

*(Also on this branch: `db63c1536` absorbed the messaging branch's new access-error variants after the rebase onto #1727's fix — exhaustive route error map, and `arm()` now relies on `authorize`'s `SelfTarget` instead of a hand-rolled check.)*

pablonyx:

PR #1729, opened 2026-08-08
feat(anyharness): peer configuration — get_agent_config_options + configure_agent

pablonyx:
Fourth slice of the agent-operations stack (ADR §6 step 4; stacked on #1728 → #1727 → #1726). Adds the two peer-configuration tools.

## What this does

- **`get_agent_config_options`** (Read intent) — composes what a TARGET agent can be configured to: its live `NormalizedSessionControl` set, unclaimed raw options, and the **target workspace's** catalog model ids/aliases merged into the model control (synthesized at `ACP_MODEL_COMPAT_CONFIG_ID` when the target has never run). Values carry `source: "live" | "workspace_catalog"`.
- **`configure_agent`** (Send intent) — generic `configId` + `value`, the same contract a human operator uses; validation → caller-workspace access check → **Config permit on the target** → **target-workspace lease** → existing `set_live_session_config_option`. Returns the verbatim `SessionLiveConfigSnapshot` plus the composed `configurable` set.
- `agent_ops/config_ops.rs` is a pure decision layer: catalog and live snapshot enter as closures, so which-workspace-gets-asked is unit-tested without a runtime.
- The PR2 peer-send fence is **generalized, not duplicated**: `admit_peer_send` → `admit_peer_mutation(admission, target, kind)`, `lease_target_workspace_for_send` → `lease_target_workspace_for_peer_write`. One fence, two callers; the LOCK-01 ratchet in `api/session_admission_tests.rs` now loops over both peer tools.
- `configure_agent` is deliberately absent from `MUTATING_TOOL_NAMES` for the same reason as `send_agent_message`: its target can live in another workspace, so it takes the TARGET workspace's lease in-call, keeping the canonical `permit → lease` order.
- SDK reducer + product-client tool presentation classify the two new tools; specs updated (`subagents.md` peer list + Peer Configuration section, `sessions.md`).

## Spec note (for founder review)

ADR §3.4 sketches `configure_agent` with named model/mode/thinking fields; the shipped tool follows the REAL existing config contract instead — a generic `configId`+`value` pair (thinking/effort is just another config id). The ADR will be amended; flagged here rather than silently diverging.

## Testing

- `cargo test -p anyharness-lib agent_ops::` 50/50; `session_admission` 9/9; full `-p anyharness-lib` 1630 passed / 1 failed (the known pre-existing `sweep_tests` red).
- Negative controls: composing from the caller's workspace instead of the target's → 4 tests fail; deleting the `admit_peer_mutation` call → the seam test and the LOCK-01 ratchet both fail.
- SDK vitest 44/44, product-client 10/10; `scripts/check_docs.py` green (209 files).

## Observability

No new telemetry; refusals log through the existing admission-fence tracing (run id logged, never returned to the caller).

### DIFF
--- a/anyharness/crates/anyharness-lib/src/api/session_admission_tests.rs
+++ b/anyharness/crates/anyharness-lib/src/api/session_admission_tests.rs
@@ -652,47 +652,51 @@ fn every_dual_lock_handler_takes_the_permit_before_the_operation_lease() {
 }
 
 #[test]
-fn the_peer_send_takes_the_permit_before_the_target_workspace_lease() {
-    // The one dual-lock site outside `api/http/`: `send_agent_message` is the
-    // first product-MCP tool that can prompt an ARBITRARY session, so it takes
-    // the target session's permit and the TARGET workspace's write lease itself
-    // rather than through the route's `MUTATING_TOOL_NAMES` lease (which is the
-    // CALLER's workspace, and would be taken before the permit — the reversed
-    // order `reversed_order_deadlocks` proves hangs). Same guard as the handler
-    // rows above; separate because the function is private, so it has no
-    // `pub async fn` signature for `handler_body` to find.
+fn the_peer_tools_take_the_permit_before_the_target_workspace_lease() {
+    // The dual-lock sites outside `api/http/`: `send_agent_message` and
+    // `configure_agent` are the product-MCP tools that can perturb an ARBITRARY
+    // session, so each takes the target session's permit and the TARGET
+    // workspace's write lease itself rather than through the route's
+    // `MUTATING_TOOL_NAMES` lease (which is the CALLER's workspace, and would be
+    // taken before the permit — the reversed order `reversed_order_deadlocks`
+    // proves hangs). Same guard as the handler rows above; separate because the
+    // functions are private, so they have no `pub async fn` signature for
+    // `handler_body` to find.
     let rel_path = "src/domains/sessions/agent_ops/calls.rs";
     let path = std::path::Path::new(env!("CARGO_MANIFEST_DIR")).join(rel_path);
     let text =
         std::fs::read_to_string(&path).unwrap_or_else(|error| panic!("read {rel_path}: {error}"));
-    let signature = "async fn send_agent_message(";
-    let start = text
-        .find(signature)
-        .unwrap_or_else(|| panic!("{rel_path}: send_agent_message not found"));
-    let rest = &text[start..];
-    // Private items follow, so stop at this function's own closing brace rather
-    // than at the next `pub` item.
-    let end = rest
-        .find("\n}\n")
-        .map(|idx| idx + 3)
-        .unwrap_or(rest.len());
-    let body = &rest[..end];
 
-    let admit_at = body
-        .find("admit_peer_send(")
-        .expect("send_agent_message must acquire the target session's mutation permit");
-    let lease_at = body
-        .find("lease_target_workspace_for_send(")
-        .expect("send_agent_message must lease the TARGET workspace");
-    assert!(
-        admit_at < lease_at,
-        "send_agent_message: 'admit_peer_send' must come BEFORE \
-         'lease_target_workspace_for_send' — {LOCK_01_ORDER}"
-    );
-    // And the dispatch stays inside both: a permit dropped before the prompt is
-    // enqueued fences nothing.
-    let dispatch_at = body
-        .find("send_text_prompt_with_provenance(")
-        .expect("send_agent_message must dispatch the prompt");
-    assert!(lease_at < dispatch_at);
+    for (tool, dispatch) in [
+        ("send_agent_message", "send_text_prompt_with_provenance("),
+        ("configure_agent", "set_live_session_config_option("),
+    ] {
+        let signature = format!("async fn {tool}(");
+        let start = text
+            .find(&signature)
+            .unwrap_or_else(|| panic!("{rel_path}: {tool} not found"));
+        let rest = &text[start..];
+        // Private items follow, so stop at this function's own closing brace
+        // rather than at the next `pub` item.
+        let end = rest.find("\n}\n").map(|idx| idx + 3).unwrap_or(rest.len());
+        let body = &rest[..end];
+
+        let admit_at = body
+            .find("admit_peer_mutation(")
+            .unwrap_or_else(|| panic!("{tool} must acquire the target session's mutation permit"));
+        let lease_at = body
+            .find("lease_target_workspace_for_peer_write(")
+            .unwrap_or_else(|| panic!("{tool} must lease the TARGET workspace"));
+        assert!(
+            admit_at < lease_at,
+            "{tool}: 'admit_peer_mutation' must come BEFORE \
+             'lease_target_workspace_for_peer_write' — {LOCK_01_ORDER}"
+        );
+        // And the dispatch stays inside both: a permit dropped before the work
+        // lands fences nothing.
+        let dispatch_at = body
+            .find(dispatch)
+            .unwrap_or_else(|| panic!("{tool} must dispatch through {dispatch}"));
+        assert!(lease_at < dispatch_at, "{tool}: dispatch escaped the locks");
+    }
 }
--- a/anyharness/crates/anyharness-lib/src/api/session_admission_tests.rs
+++ b/anyharness/crates/anyharness-lib/src/api/session_admission_tests.rs
@@ -579,6 +579,12 @@ fn assert_source_order(rel_path: &str, fn_name: &str, first: &str, second: &str,
     );
 }
 
+/// Collapse every run of whitespace to a single space, so a source needle can
+/// span a `let` binding and its call without a rustfmt line wrap breaking it.
+fn squash_whitespace(source: &str) -> String {
+    source.split_whitespace().collect::<Vec<_>>().join(" ")
+}
+
 const LOCK_01_ORDER: &str =
     "canonical LOCK-01 order: session mutation permit -> workspace operation lease";
 
@@ -679,21 +685,42 @@ fn the_peer_tools_take_the_permit_before_the_target_workspace_lease() {
         // Private items follow, so stop at this function's own closing brace
         // rather than at the next `pub` item.
         let end = rest.find("\n}\n").map(|idx| idx + 3).unwrap_or(rest.len());
-        let body = &rest[..end];
+        // Whitespace runs collapse to one space so a rustfmt wrap of the
+        // argument list cannot break a needle, and so the needles can span the
+        // `let` binding.
+        let body = squash_whitespace(&rest[..end]);
+        let body = body.as_str();
 
+        // The needles pin the BINDING, not just the call. Matching
+        // `admit_peer_mutation(` alone would pass on `let _ = admit_peer_mutation(..)`,
+        // which drops the permit at the end of that statement — textually
+        // identical ORDER, no fence at all across the dispatch below. What this
+        // test exists to prove is that both guards are still HELD when the work
+        // lands, so the named bindings are the thing to ratchet.
         let admit_at = body
-            .find("admit_peer_mutation(")
-            .unwrap_or_else(|| panic!("{tool} must acquire the target session's mutation permit"));
+            .find("let _admission_permit = admit_peer_mutation(")
+            .unwrap_or_else(|| {
+                panic!(
+                    "{tool} must BIND the target session's mutation permit as \
+                     `_admission_permit` — an unbound `let _ =` drops it immediately"
+                )
+            });
         let lease_at = body
-            .find("lease_target_workspace_for_peer_write(")
-            .unwrap_or_else(|| panic!("{tool} must lease the TARGET workspace"));
+            .find("let _target_workspace_lease = lease_target_workspace_for_peer_write(")
+            .unwrap_or_else(|| {
+                panic!(
+                    "{tool} must BIND the TARGET workspace lease as \
+                     `_target_workspace_lease` — an unbound `let _ =` drops it immediately"
+                )
+            });
         assert!(
             admit_at < lease_at,
             "{tool}: 'admit_peer_mutation' must come BEFORE \
              'lease_target_workspace_for_peer_write' — {LOCK_01_ORDER}"
         );
         // And the dispatch stays inside both: a permit dropped before the work
-        // lands fences nothing.
+        // lands fences nothing. Both bindings live to the end of the function,
+        // so a dispatch that appears after them is inside both.
         let dispatch_at = body
             .find(dispatch)
             .unwrap_or_else(|| panic!("{tool} must dispatch through {dispatch}"));
--- a/anyharness/crates/anyharness-lib/src/domains/sessions/agent_ops/calls.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/sessions/agent_ops/calls.rs
@@ -1,23 +1,29 @@
 use std::collections::HashMap;
 
+use anyharness_contract::v1::ConfigApplyState;
 use serde_json::{json, Value};
 
 use super::calls_helpers::{
     default_model_for_agent, initial_config_string, launch_agents_to_json, mode_options_to_json,
     prompt_outcome_label, summaries_to_json,
 };
+use super::config_ops::{
+    compose_agent_config_options, controls_to_json, current_selection_to_json,
+    prepare_agent_config_change,
+};
 use super::context::AgentOpsMcpContext;
 use super::peer_ops::{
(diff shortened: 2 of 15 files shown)


pablonyx (commented):
**Verdict: sound — the fence generalization is correct and the cross-workspace composition is genuinely pinned. No High findings. Two Mediums (one live doc contradiction, one read/write asymmetry the PR's own rationale argues against) and three Lows, all addressable without restructuring.**

Reviewed `git diff db63c1536..3ee2ac632` only; everything already on `agent-ops/3-wakes` treated as out of scope. The generic `configId`+`value` deviation from ADR §3.4 is the disclosed, accepted one and is not flagged (the ADR text in the worktree already reads that way).

---

## What checked out

**Lock ordering / admission fence (PR1227-LOCK-01)**

- `configure_agent` takes nothing before the permit. `calls.rs:731` validates (pure read: store + catalog + snapshot, no gate), `calls.rs:743` is a bare access-gate assertion with no lease, `calls.rs:749-754` takes the permit, `calls.rs:755-760` takes the lease, `calls.rs:764-767` applies. Canonical order holds.
- No route lease can sneak in ahead of it: `product_registry.rs:86` takes the endpoint lease only for names in `mutating_tools`, and `tools.rs:29-37` omits both new tools. `tools.rs:531-537` pins that omission for `configure_agent` *and* `get_agent_config_options`, so a well-meaning "this mutates, add it to the list" would fail the test rather than silently reverse the order.
- Permit kind and source are right: `SessionMutationKind::Config` at `calls.rs:752` — the same kind the human route uses at `api/http/sessions_config.rs:96` — with `SessionMutationSource::external()` hard-coded inside `peer_ops.rs:122`, never derived from caller metadata.
- Lease target is the TARGET's workspace (`calls.rs:758` passes `prepared.target.workspace_id`), acquired shared as `SubagentWrite` (`peer_ops.rs:170`). Different kind from the human path's `SessionResume`, but both are shared holders on the same per-workspace `RwLock` and both appear in `retire_preflight.rs:453,459`, so mutual exclusion and retire visibility are equivalent — only the reported *reason* differs.
- Both bindings are `let _admission_permit` / `let _target_workspace_lease`, i.e. actually held to end of scope, across the apply. (What *proves* that is the weak spot — see M-1.)
- Generalization is a rename, not a fork: one `admit_peer_mutation`, one `lease_target_workspace_for_peer_write`, and `git grep` at `3ee2ac632` finds zero surviving references to `admit_peer_send` / `lease_target_workspace_for_send` / `PeerSendGateError`.

**Authorization funnel**

- Options = `AgentAccessIntent::Read` (`config_ops.rs:157`), configure = `Send` (`config_ops.rs:202`). Nothing is re-decided locally: only `SelfTarget` is remapped for wording (`config_ops.rs:204-207`), so closed / dismissed / internal-only / caller-closed all come straight from the funnel.
- Self-read allowed, self-configure refused, both pinned (`config_ops.rs:820-846`). The self refusal is correct on its own terms — a self-directed config would be routed at the actor blocked inside the tool call.
- `internal_only` is refused for both intents by the funnel itself (`authorize.rs:97-98`), so the new read tool cannot become a discovery channel for workflow plumbing.

**Cross-workspace correctness**

- `compose()` asks the catalog exactly once, for `target.workspace_id` (`config_ops.rs:233`). `ctx.workspace_id` appears in the whole delta only at `calls.rs:743` (the caller's own writability assertion) and never reaches composition or validation. The `CatalogSpy` (`config_ops.rs:573-598`) serves deliberately disjoint menus per workspace, so a caller-workspace leak is visible in the *result*, not just in the spy — `config_ops.rs:605-655` is a real two-sided negative control (target model accepted, caller-only model rejected, spy asked `workspace-2` only).
- `target_is_closed` (`config_ops.rs:410-412`) is byte-identical to `authorize::is_closed` (`authorize.rs:121-123`): `closed_at.is_some() || status == "closed"`. Consistent today, and the doc comment says to collapse them if the predicate is ever exported.

**Config write path**

- No privilege widening. The peer path calls the same `SessionRuntime::set_live_session_config_option` the human route calls, with no extra arguments and no bypass; the real model authorization (`live_model_switch_authorized`, `service/launch_options.rs:66-91`) is inside that function and keyed off the TARGET record's own recorded auth contexts, so the peer cannot reach a model the target itself may not have. The peer-side composition mirrors what the human chat client already merges (`use-chat-launch-catalog.ts` joins the workspace launch options with live controls), so the advertised menu is not wider than the human surface either.
- Validation is entirely before the permit (`config_ops.rs:214` inside `prepare_*`), so a refused call costs no permit, no lease, and no side effect. `config_ops.rs:849-880` proves blank args short-circuit before any catalog lookup (`spy.asked` empty).
- The trimmed/validated triple is what gets applied (`calls.rs:765` passes `prepared.*`, never `args.*`), and the seam test pins that textually.
- Mid-turn / idle / never-run all behave: queueing and lazy boot are the existing apply path's business (`runtime/config.rs:43-104`), and the never-run case is handled by synthesizing the model control at `ACP_MODEL_COMPAT_CONFIG_ID` (`config_ops.rs:289-314`), which is the same id `live_config/controls.rs:92` uses — so the synthesized id is one the actor's compat path recognizes (`live/sessions/actor/config/selection.rs:42`), not an invented one.
- `get_agent_config_options` is genuinely read-only: `live_config_snapshot` bottoms out in `session_store.find_live_config_snapshot` (`service/config.rs:5-14`) and never ensures a live handle.

**Contract / SDK / client**

- Reducer classification (`transcript.ts:1439-1440`), action labels and header verbs (`subagent-tool-presentation.ts:52-55, 106-111`) all use the same `mcp__subagents__*` prefix as the existing peer tools, matching `definition.rs`'s frozen `ACP_SERVER_NAME = "subagents"`. `ConfigureAgentArgs` (`tools.rs`) mirrors `SetSessionConfigOptionRequest`'s field names and `tools.rs:554-566` pins the required list. No stale names anywhere in the delta.

---

## Findings

### M-1 (Medium) — the ratchet pins call *order*, not that the permit and lease are *held*

`session_admission_tests.rs:655-702` and `config_ops.rs:887-952` both work by `str::find` on call-site substrings and comparing byte offsets. Neither looks at the binding. In Rust, `let _ = admit_peer_mutation(...).await?;` drops the permit at the end of that statement, whereas `let _admission_permit = ...` holds it to end of scope — and the two are textually indistinguishable to every assertion here.

Concrete failure: a future cleanup pass that "removes unused variables" rewrites `let _admission_permit =` / `let _target_workspace_lease =` to `let _ =`. All three source-order tests stay green, `cargo clippy` stays quiet, and the fence silently becomes a point-check: a workflow can take control of the target between the permit release and `set_live_session_config_option`, which is precisely the interleave the permit exists to prevent. Same hole applies to `send_agent_message`, so it is pre-existing — but this PR generalizes the ratchet and the PR body claims it "now loops over both peer tools", so it is worth closing here.

Cheap fix: assert on the binding, not the call — `body.contains("let _admission_permit = admit_peer_mutation(")` (whitespace-squashed, as the seam test already does for the apply arguments), and the same for the lease. That is one extra `contains` per tool and it discriminates the exact regression the current assertions miss.

### M-2 (Medium) — `specs/anyharness/sessions.md:536-540` describes a lease that does not exist

> "when the target is in a different workspace than the caller, it also takes that workspace's shared operation lease and access-gate check … a same-workspace target reuses the lease the MCP endpoint already holds"

Both halves are wrong for `configure_agent`. The lease is taken unconditionally, same-workspace or not (`calls.rs:755-760` has no branch), and the MCP endpoint holds **no** lease for this tool at all, because `configure_agent` is absent from `MUTATING_TOOL_NAMES` (`tools.rs:29-37`, `product_registry.rs:86`). The prose contradicts the comment this same PR added at `tools.rs:21-28`.

This matters beyond tidiness: this is canonical current documentation, and the false "same-workspace reuses the endpoint lease" claim is exactly the belief that would lead someone to add `configure_agent` to `MUTATING_TOOL_NAMES` and reverse the canonical order. Suggested replacement: "it always takes the TARGET workspace's shared operation lease itself, in the canonical `permit -> operation lease` order; it takes no endpoint lease, because the endpoint would lease the caller's workspace and would do so before the permit."

### M-3 (Medium) — a dismissed target gets a full menu it can never apply

`get_agent_config_options` refuses closed targets (`config_ops.rs:160-162`) with an explicit rationale at `config_ops.rs:34-38`: a closed session's actor never starts again, so "what can I change now" is always "nothing", and an empty menu would read like a harness with no controls.

That argument applies verbatim to a **dismissed** target, and dismissed is not refused on the read path. The funnel only rejects `dismissed_at` under `Send` (`authorize.rs:114-116`), for the stated reason that the boot path refuses a dismissed session outright (`runtime/launch_policy.rs`). So today:

1. peer calls `get_agent_config_options` on a dismissed-but-open session → full composed menu, `settable: true` model control, catalog values and all;
2. peer follows the tool description's own instruction ("Call `get_agent_config_options` first") and calls `configure_agent` with a value straight out of that menu;
3. refused with "target session was dismissed and cannot be messaged" — a message that does not even mention configuration.

The agent is handed a menu whose every entry is unreachable, which is the failure mode the closed-target refusal was added to avoid. Either refuse dismissed on the read path too (one extra condition next to `target_is_closed`, and the existing closed test extends to it directly), or document why dismissed is deliberately different from closed here. Right now the code reads as if the case was simply not considered — `target_is_closed` checks only closure.

### L-1 (Low) — the seam test's function-body window is ~130 lines too wide (dead `or_else`)

`config_ops.rs:895-898`:

```rust
let body = &rest[..rest[1..]
    .find("\nasync fn ")
    .or_else(|| rest[1..].find("\nfn "))
    .map_or(rest.len(), |at| at + 1)];
```

`or_else` runs only when the first `find` returns `None`. It never does here: `configure_agent` is at `calls.rs:722` and the next `\nasync fn ` is `get_subagent_status` at `calls.rs:856`, so the "body" swallows `read_agent_transcript` and `schedule_subagent_wake` whole. Harmless today (none of the four probed symbols nor `SessionMutationKind::Config` appears in 781-855), but it is a live trap: any future private fn added between `configure_agent` and the next `async fn` that mentions `admit_peer_mutation(` or `set_live_session_config_option(` would satisfy the ratchet on behalf of a `configure_agent` that no longer does. Fix is `min` of the two positions rather than `or_else`. The LOCK-01 ratchet at `session_admission_tests.rs:681` gets this right with `\n}\n`.

### L-2 (Low) — catalog-sourced model values can be advertised and then rejected at apply, after booting the target

Composition classifies models against the machine's currently-active auth contexts (`service/launch_options.rs:132-149`, `classify(...)` over the catalog's declared contexts and env facts), while the apply gate uses the contexts recorded on the target session at creation (`live_model_switch_authorized`, `service/launch_options.rs:66-91`). When those drift — e.g. the target was created with only an API-key context, then the machine gained an OAuth context that unlocks more models — `get_agent_config_options` lists the new models with `source: "workspace_catalog"`, `configure_agent` validates them, takes the permit and the target-workspace lease, and only then does `set_live_session_config_option` compute `catalog_authorized_model = false`; the live-or-relaunch arm (`runtime/config.rs:112-124`) is gated on that flag, so the actor's rejection surfaces as an error.

Not a widening (the apply gate is the authority and it holds), and the human client's model selector composes from the same workspace launch options, so it is a shared pre-existing gap rather than something this PR introduces. The new part is the side effect: `ensure_live_session_handle` (`runtime/config.rs:44`) runs *before* the authorization check, so a failed peer `configure_agent` can boot an idle agent in another workspace and leave it running. Worth a sentence in the tool's `notes` (or a follow-up issue) so the next reader knows `source: "workspace_catalog"` means "authorized by the workspace", not "guaranteed to apply".

### L-3 (Low) — two error-surface nits

- `AgentConfigError::TargetClosed`'s message says "no configuration left to read **or change**" (`config_ops.rs:39-42`), but the change path can never produce it — `prepare_agent_config_change` remaps only `SelfTarget`, so a closed target on the write path returns the funnel's generic "target session is closed" (and `config_ops.rs:813-816` asserts exactly that). Half the sentence is unreachable.
- `ComposedControl::settable()` is `values.len() > 1` (`config_ops.rs:90`), but `validate_change` never consults it, so a single-value control is reported `settable: false` and still accepts that one value. Today's normalized controls compute `settable` the same way upstream (`live_config/controls.rs:95,122,195`), so nothing diverges — just note that `settable` is advisory, not enforced.

---

## Test verification notes

- **Genuine, not mirrors.** `config_ops.rs:605-655` are the strongest tests in the PR: the `CatalogSpy` returns *different menus* per workspace, so both the "asked the right workspace" and "got the right values" assertions fail if composition switched to the caller — the negative control the PR body claims (4 tests fail) is real and two-sided (accept target-only model / reject caller-only model).
- `config_ops.rs:849-880` discriminates properly: `assert!(spy.asked.borrow().is_empty())` is what makes "rejected before any lookup" mean something rather than just "rejected".
- `config_ops.rs:703-737` pairs a rejection with the accepting case in the same test — a guard clause that rejected everything would fail the second half.
- `peer_ops.rs:598-618` extends the workflow-controlled fence test to `SessionMutationKind::Config` and keeps the "ordinary target is still admitted" negative control for the new kind, so the refusal is provably the controller policy and not a blanket block. It also re-asserts the run id never leaks into the agent-visible message.
- **Weaknesses.** M-1: the three source-order tests are the *only* thing guarding the fence, and they do not pin holding. L-1: the seam test's window is over-wide. Also note nothing exercises `configure_agent` end to end (it needs a live actor) — the source-order guard is a deliberate stand-in, which is a reasonable trade, but it means the tool's own error mapping (`calls.rs:767`'s `{error:?}`) and JSON response shape are untested. That matches how `send_agent_message` was landed, so it is consistent rather than a regression.
- The `tools.rs:541-551` "offered to every caller" test correctly covers the unpromoted-subagent context (ADR ruling 3: unpromoted keeps message/read/wake/**configure**).
- I did not compile: every finding above is static (source text, drop semantics, spec prose), so no build slot was taken.

pablonyx:
Rebased onto `agent-ops/3-wakes` @ `39f5c7e83` and pushed the review fixes. Fix commits: `f0bc5bfbc` (rebase fallout), `332686d79` (findings).

**Rebase** applied with no conflicts. Verified rather than assumed: `git diff 3ee2ac632 HEAD` is exactly the wakes-fix commit and nothing else, and TS/SDK files are byte-identical across the replay. The merged send path is caller-workspace assert → `admit_peer_mutation(Prompt)` → `lease_target_workspace_for_peer_write` → arm (reply reason) → dispatch → stamp-confirmed / `discard_unconfirmed_reply_arm` → consume reply wake — this branch's generalized names, the wakes fix's semantics. Fallout: two doc references to the pre-rename `admit_peer_send` that landed on the wakes branch, repointed.

**M-1** Both ratchets pinned call order, which `let _ = admit_peer_mutation(..)` satisfies while dropping the permit immediately. They now match the binding (`let _admission_permit = …`, `let _target_workspace_lease = …`) over whitespace-squashed source, for both peer tools. Negative-controlled: rebinding either guard to `let _ =` fails both ratchets.

**M-2** Rewrote the `sessions.md` paragraph. There is no same/different-workspace branch and no route lease to reuse — both peer tools are absent from `MUTATING_TOOL_NAMES`, so the target-workspace lease is taken unconditionally, plus a caller-workspace access check with no lease.

**M-3** Read path now refuses dismissed targets, same as closed. `is_closed` is shared with the funnel rather than copied (the wakes fix made it `pub(crate)`, which the old comment asked for). Symmetry rationale documented on the `TargetClosed` variant. Tests cover read + write refusal; reverting the guard fails only that test.

**L-1** `or_else` → `min` of both header positions, so the window can't swallow following `fn`s.

**L-3** `TargetClosed`'s message no longer claims "read or change". `settable` is now enforced in `validate_change` — but a blanket `len() > 1` reject would have broken a real case, so a single-value control stays settable while that value isn't current (selecting it is a genuine change); only the dead end is refused. Off-menu values still get the more specific "here is what IS available" error.

**L-2 — not reordered, on purpose.** The premise doesn't hold: `catalog_authorized_model` is computed for every `config_id` and is a relaunch-permission flag, not an apply gate. Hoisting it above `ensure_live_session_handle` is a no-op (the handle is needed for the apply regardless, so the idle target still boots), and making it a gate would start refusing every non-model control change on the human HTTP route. Documented the side effect and why the obvious reorder is wrong at the call site; the real fix (narrow the advertised set, or authorize pre-apply keyed on `config_id`) is tracked as a follow-up issue.

Scoped suites 174/174. Full `anyharness-lib --lib` 1650 passed, 1 failed — the known pre-existing `sweep_tests` red. No rustfmt drift in the touched files.

One disclosure beyond the findings list: `subagents.md` already claimed `read_agent_transcript` refuses dismissed sessions, which is false (Read intent refuses only `internal_only`) and contradicted the new M-3 text, so that sentence now splits refusals by intent — behavior unchanged.

pablonyx:

PR #1731, opened 2026-08-08
feat(anyharness): agent ownership — promote_subagent, close_agent, soft close

pablonyx:
Fifth slice of the agent-operations stack (ADR §6 step 5; stacked on #1729 → #1728 → #1727 → #1726). Adds the ownership layer: `promote_subagent`, a generalized `close_agent`, and the soft close.

## What this does

Ownership is a `session_links` row, not reachability. Every agent can reach every other agent — that is what the peer tools are for — but only the agent that spawned a subagent may promote or close it. Migration 0068 puts the three states on the existing row: a subagent is `relation='subagent'` with `promoted_at IS NULL`, a promoted agent is the same row with the stamp set, and `owned_agent` is reserved for the spawn step that follows (PR 6).

Promotion keeps everything and drops subordination: same transcript, same label, same owner, but off the close cascade, off the fanout cap, and with the spawn tools unlocked. It is one write, idempotent, one-way.

`close_agent` now takes a `sessionId` and an optional `reason`. Its target may be in another workspace and a close acts on the target session, so it leaves the route's lease behind and takes its own fence — the target session's mutation permit (`SessionMutationKind::Close`, the same kind the human close route uses), then the target workspace's write lease, in that order (PR1227-LOCK-01). Closing a working agent does not interrupt it: the call stamps the still-open row and returns "end requested", and the agent finishes its step before the tree closes (`closed_by_session_id IS NOT NULL AND closed_at IS NULL` = end requested — no new table). Only a `Running` turn defers; `AwaitingInteraction` cannot finish without a human and closes immediately, as does an idle agent. Close attribution (`closed_by_session_id` + `close_reason`) is stamped guarded, so a second close cannot overwrite the first close's record.

An unpromoted subagent is offered no spawn-style tool at all. Tool lists are frozen at launch, so the gate that binds runs at dispatch, on the wire name; `tools/list` filtering is advertisement only.

## Spec note

The implementation follows ADR §3.2 exactly: promotion stamps `promoted_at` and leaves `relation='subagent'`; the cascade predicate, the DepthLimit lift, and "the parent can still close it individually" (ruling 7) all depend on the row keeping its relation. `'owned_agent'` rows are written only by `spawn_agent` in PR 6. (An earlier working-brief paraphrase said promotion flips the relation — that was wrong; the ADR itself is consistent.)

Deferred, fail-closed: `mobility_archive_contract.rs` imports the three new columns as `None` — carrying them is a wire-contract change plus SDK regen, commented in place.

## Testing

- Full `anyharness-lib --lib`: 1657 passed / 1 failed (the known pre-existing `sweep_tests` red). After rebase onto #1729's head: scoped suites 215/215.
- 25 new tests: ownership resolution, promotion idempotency and scope, cascade behavior, spawn gating, both lock orders.
- Negative controls, each run and restored: reverting the cascade guard, the dispatch-time spawn gate, or the close permit each fails exactly its named test.
- SDK vitest 45/45; product-client vitest 21/21; `check_docs.py` green (209 files).

## Observability

A deferred close logs at info on completion (`session_id`, `session_link_id`, `closed_by_session_id`, `turn_id`) and when a gate refuses and the request stays armed; unexpected failures warn. Promotion and close both return their settled state (`promotedAt`, `closedAt`, `closedBySessionId`, `closeReason`, `alreadyPromoted`/`alreadyClosed`/`closeRequested`) so the calling agent and the transcript read the same facts.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/api/http/mobility_archive_contract.rs
+++ b/anyharness/crates/anyharness-lib/src/api/http/mobility_archive_contract.rs
@@ -217,6 +217,16 @@ fn from_contract_session_link(
         created_by_tool_call_id: record.created_by_tool_call_id,
         created_at: record.created_at,
         closed_at: record.closed_at,
+        // The mobility archive contract does not carry ownership state, so an
+        // imported link lands unpromoted with no close attribution. That is the
+        // fail-CLOSED direction: a promoted agent that moves workspaces comes
+        // back as a plain subagent (fewer tools, cascade-closable) rather than
+        // as an agent nobody can account for. Carrying the three columns means
+        // widening `MobilitySessionLinkRecord`, which is a wire-contract change
+        // and an SDK regeneration.
+        promoted_at: None,
+        closed_by_session_id: None,
+        close_reason: None,
     })
 }
 
--- a/anyharness/crates/anyharness-lib/src/app/mod.rs
+++ b/anyharness/crates/anyharness-lib/src/app/mod.rs
@@ -486,6 +486,11 @@ impl AppState {
         // Ownership — phase 2: the close hook can now reach the runtime whose
         // extension list holds it. Weak, so this is not a reference cycle.
         agent_close_session_hooks.attach_session_runtime(&session_runtime);
+        // Closes this runtime already owed when it started: a request whose
+        // turn never finished because the process died has nothing left to fire
+        // it, so it is settled here instead of leaving the agent open forever.
+        #[cfg(not(test))]
+        agent_close_session_hooks.clone().spawn_startup_pass();
         // Workflow runs — phase 2 (after SessionRuntime): the async facades.
         let workflow_phase_two = workflows::wire_workflow_runtime(
             workflow_wiring,
--- a/anyharness/crates/anyharness-lib/src/app/mod.rs
+++ b/anyharness/crates/anyharness-lib/src/app/mod.rs
@@ -64,6 +64,8 @@ use crate::domains::sessions::mcp_bindings::product_registry::ProductMcpEndpoint
 use crate::domains::sessions::runtime::SessionRuntime;
 use crate::domains::sessions::service::SessionService;
 use crate::domains::sessions::store::SessionStore;
+use crate::domains::sessions::ownership::hooks::AgentCloseSessionHooks;
+use crate::domains::sessions::ownership::service::AgentOwnershipService;
 use crate::domains::sessions::subagents::hooks::SubagentSessionHooks;
 use crate::domains::sessions::subagents::service::SubagentService;
 use crate::domains::sessions::subagents::store::SubagentStore;
@@ -436,12 +438,27 @@ impl AppState {
             acp_manager.clone(),
         ));
         let workflow_run_session_extension = workflow_wiring.session_extension.clone();
+        // Ownership: who may promote or close whom. Its turn-finish hook is the
+        // second half of a soft close, and it needs the runtime it lives inside,
+        // so it is wired in two phases exactly like the workflow extensions —
+        // constructed here, handed the runtime below.
+        let agent_ownership_service = Arc::new(AgentOwnershipService::new(
+            session_link_service.clone(),
+            SessionStore::new(db.clone()),
+        ));
+        let agent_close_session_hooks = Arc::new(AgentCloseSessionHooks::new(
+            agent_ownership_service.clone(),
+            session_admission.clone(),
+            workspace_operation_gate.clone(),
+            workspace_access_gate.clone(),
+        ));
         let session_extensions: Vec<
             Arc<dyn crate::domains::sessions::extensions::SessionExtension>,
         > = vec![
             cowork_session_hooks.clone(),
             subagent_session_hooks.clone(),
             agent_wake_session_hooks.clone(),
+            agent_close_session_hooks.clone(),
             review_session_hooks.clone(),
             integration_gateway_session_launch_extension.clone(),
             goal_session_hooks,
@@ -466,6 +483,9 @@ impl AppState {
             loop_service.clone(),
             activity_service.clone(),
         ));
+        // Ownership — phase 2: the close hook can now reach the runtime whose
+        // extension list holds it. Weak, so this is not a reference cycle.
+        agent_close_session_hooks.attach_session_runtime(&session_runtime);
         // Workflow runs — phase 2 (after SessionRuntime): the async facades.
         let workflow_phase_two = workflows::wire_workflow_runtime(
             workflow_wiring,
@@ -567,6 +587,7 @@ impl AppState {
                 agent_wake_service: agent_wake_service.clone(),
                 session_runtime: session_runtime.clone(),
                 workspace_runtime: workspace_runtime.clone(),
+                agent_ownership_service: agent_ownership_service.clone(),
                 session_admission: session_admission.clone(),
                 workspace_operation_gate: workspace_operation_gate.clone(),
                 agent_ops_mcp_auth,
--- a/anyharness/crates/anyharness-lib/src/app/product_mcp.rs
+++ b/anyharness/crates/anyharness-lib/src/app/product_mcp.rs
@@ -15,6 +15,7 @@ use crate::domains::sessions::agent_ops::{
     AgentOpsPeerGates, AgentOpsProductMcpServer,
 };
 use crate::domains::sessions::mcp_bindings::product_catalog::ProductMcpLaunchCatalog;
+use crate::domains::sessions::ownership::service::AgentOwnershipService;
 use crate::domains::sessions::mcp_bindings::product_launch::{
     ProductMcpLaunchRegistration, ProductMcpSelectionContext,
 };
@@ -43,6 +44,7 @@ pub(super) struct EndpointRegistryDeps {
     pub(super) agent_wake_service: Arc<AgentWakeService>,
     pub(super) session_runtime: Arc<SessionRuntime>,
     pub(super) workspace_runtime: Arc<WorkspaceRuntime>,
+    pub(super) agent_ownership_service: Arc<AgentOwnershipService>,
     pub(super) session_admission: Arc<SessionMutationAdmission>,
     pub(super) workspace_operation_gate: Arc<WorkspaceOperationGate>,
     pub(super) agent_ops_mcp_auth: Arc<AgentOpsMcpAuth>,
@@ -115,6 +117,7 @@ pub(super) fn build_product_mcp_endpoint_registry(
         agent_wake_service,
         session_runtime,
         workspace_runtime,
+        agent_ownership_service,
         session_admission,
         workspace_operation_gate,
         agent_ops_mcp_auth,
@@ -135,6 +138,7 @@ pub(super) fn build_product_mcp_endpoint_registry(
                 agent_wake_service,
                 session_runtime,
                 workspace_runtime.clone(),
+                agent_ownership_service,
                 AgentOpsPeerGates {
                     session_admission,
                     workspace_operation_gate,
--- a/anyharness/crates/anyharness-lib/src/app/sessions.rs
+++ b/anyharness/crates/anyharness-lib/src/app/sessions.rs
@@ -18,6 +18,7 @@ use crate::domains::plans::session_observer::PlanSessionObserver;
 use crate::domains::reviews::service::ReviewService;
 use crate::domains::reviews::session_observer::ReviewSessionObserver;
 use crate::domains::sessions::attachment_storage::PromptAttachmentStorage;
+use crate::domains::sessions::links::store::SessionLinkStore;
 use crate::domains::sessions::live_ports::SessionAttachmentSource;
 use crate::domains::sessions::store::SessionStore;
 use crate::live::sessions::model::{ActorCapabilities, PermissionAdvisor, SessionEventObserver};
@@ -64,6 +65,9 @@ pub(super) fn wire_live_sessions(deps: &LiveSessionsWiringDeps) -> LiveSessionMa
         background: Arc::new(store.clone()),
         state: Arc::new(store.clone()),
         attachments: Arc::new(SessionAttachmentSource::new(store, attachment_storage)),
+        // The soft-close fence on turn starts: an agent whose end has been
+        // requested finishes its step and starts nothing new.
+        close_requests: Some(Arc::new(SessionLinkStore::new(deps.db.clone()))),
         observers,
         permission_advisor,
     };
--- a/anyharness/crates/anyharness-lib/src/app/test_support.rs
+++ b/anyharness/crates/anyharness-lib/src/app/test_support.rs
@@ -24,6 +24,9 @@ pub(crate) fn actor_capabilities_for_store(store: &SessionStore) -> ActorCapabil
             store.clone(),
             attachment_storage,
         )),
+        // No link store here (the helper only takes a session store), so no
+        // soft-close fence. Tests that exercise the fence supply their own.
+        close_requests: None,
         observers: Vec::new(),
         permission_advisor: None,
     }
--- a/anyharness/crates/anyharness-lib/src/domains/cowork/delegation/service.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/cowork/delegation/service.rs
@@ -276,6 +276,9 @@ impl CoworkDelegationService {
             created_by_tool_call_id: None,
             created_at: chrono::Utc::now().to_rfc3339(),
             closed_at: None,
+            promoted_at: None,
+            closed_by_session_id: None,
+            close_reason: None,
         };
         let inserted = self
             .cowork_service
--- a/anyharness/crates/anyharness-lib/src/domains/sessions/agent_ops/calls.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/sessions/agent_ops/calls.rs
@@ -13,8 +13,9 @@ use super::config_ops::{
 };
 use super::context::AgentOpsMcpContext;
 use super::peer_ops::{
-    admit_peer_mutation, assert_workspace_can_be_mutated, authorize_transcript_read,
-    consume_reply_wake, lease_target_workspace_for_peer_write, prepare_agent_message,
+    admit_peer_mutation, assert_target_still_takes_messages, assert_workspace_can_be_mutated,
+    authorize_transcript_read, consume_reply_wake, lease_target_workspace_for_peer_write,
+    prepare_agent_message,
 };
 use super::tools::{
     canonical_tool_name, is_spawn_style_tool, ChildSessionArgs, CloseAgentArgs, ConfigureAgentArgs,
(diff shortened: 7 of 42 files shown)


pablonyx (commented):
**Verdict: sound and close to the ADR — no High findings. Ownership authorization, the close fence and lock order, the cascade predicate and the spawn gate all hold up under adversarial reading. Five Medium items, all in the soft-close half and one advertised-limits bug; the rest is Low/test hygiene.** Reviewed as delta only: `332686d79..5d4ff2fc5`, 4 commits, 34 files. Static review (no cargo run — swap was at 4.5G/6G and the machine is one-build-at-a-time); the source-order guards were instead re-executed exactly by reimplementing their string arithmetic against the real file bytes, results below.

## What checked out

**Ownership authorization (priority 1).** I could not find a path to promote or close an agent you do not own.

- `ownership/service.rs:98-114` — both handles resolve through a parent-scoped lookup. The `subagentId` path takes `find_by_public_id` (unique index) and then filters `link.relation.is_ownership() && link.parent_session_id == owner_session_id`; the `sessionId` path calls `find_owned_link_including_closed(owner, target)` whose SQL (`links/store.rs:356-374`) pins `parent_session_id = ?1`. A stranger holding a valid `subagentId` or a valid child `sessionId` gets `NotOwned` either way, and the child cannot invert the row on its parent (`service.rs:370-385` pins both).
- The owner predicate is exactly §3.2's. `is_ownership()` (`links/model.rs:48-50`) is `Subagent | OwnedAgent`; `closed_at IS NULL` is applied where it belongs (`list_owned_children`, `links/store.rs:378-393`) and deliberately relaxed on the single-row resolve so an already-closed agent still resolves idempotently and keeps its attribution readable. `Fork`, `CoworkCodingSession` and `ReviewAgent` are excluded and pinned by test.
- Both `subagentId` and `sessionId` supplied → `ConflictingTarget` (`service.rs:118-122`); empty strings are trimmed to `None` first (`service.rs:89-90`), so `subagentId: ""` degrades to the session path rather than resolving something else.
- Ruling 7, both halves: a promoted child still resolves for its former parent (the row keeps `relation='subagent'` and stays open, `store.rs:317-327` never touches `relation`), and the cascade skips it (`lifecycle.rs:344-352`). Correct.

**Close fence + lock order (priority 2).** `close_agent` (`calls.rs:1017-1085`) takes `SessionMutationKind::Close` on the **target** (`:1046`) then the **target** workspace lease (`:1052`), stamps under both (`:1061`), and holds both across `close_live_session` (`:1070`) — the bindings are `let _admission_permit` / `let _target_workspace_lease`, i.e. really held to end of scope, not `let _`. `close_agent`/`close_subagent` correctly left `MUTATING_TOOL_NAMES` (`tools.rs:35-42`) so the route lease cannot be taken before the permit; `promote_subagent` correctly joined it (one guarded UPDATE, caller's own workspace, no actor). The caller's own workspace is still access-checked at `:1045`. The deferred half re-takes the identical pair in the identical order (`ownership/hooks.rs:112-142`) and, on refusal, leaves the request armed rather than dropping it. I re-ran all three source-order guards byte-exactly (see Test notes) — they pass.

**Cascade (priority 4).** `cascades_to_child` is read fresh per close from `list_by_parent` inside `close_delegated_children` (`lifecycle.rs:135-150`), so an agent promoted between request and execution is re-evaluated at close time. `close_inbound_delegated_links` widening to `OwnedAgent` (`lifecycle.rs:191-199`) is right for ruling 7: outbound asks "what do I take down", inbound asks "what pointed at me", and a dead child ends every relationship regardless of promotion.

**Spawn gating (priority 5).** The dispatch gate (`calls.rs:63-69`) runs before the `match`, matches the **wire** name, and `SPAWN_STYLE_TOOL_NAMES` (`tools.rs:50-54`) includes the `create_subagent` alias — the alias-bypass is closed and pinned by an alias-coverage test. The fanout subselect gained `AND promoted_at IS NULL` (`links/store.rs:76`) and the service-side pre-check gained the matching filter (`subagents/service.rs:139`), so the store and the service agree that a promoted child no longer occupies a slot. The DepthLimit lift (`subagents/service.rs:117-123`) is `find_subagent_parent(...).is_some_and(is_unpromoted_subagent)` — exactly §3.2's spawn-tool-grant predicate. `get_subagent_launch_options` stays advertised to a *capped* parent and is withheld only from a *subordinate* one (`tools.rs:345`), which is the right distinction and is pinned by a negative control.

**Migration + snapshot (priority 6).** `0068` adds the three columns and the partial index `(child_session_id) WHERE closed_at IS NULL AND closed_by_session_id IS NOT NULL`, which is precisely the selective predicate `find_pending_close_request` uses (`links/store.rs:398-416`). The schema snapshot is machine-enforced by `persistence/schema_snapshot_tests.rs:55-79`, and the diff has the exact shape a real SQLite `ALTER TABLE ADD COLUMN` regeneration produces (columns appended after `closed_at TEXT`, before the table constraints) — it was regenerated, not hand-typed.

**SDK/client (priority 7).** `promote_subagent` added to both the reducer's subagent-activity arm and the presentation switch; both `close_agent` and legacy `close_subagent` handled everywhere; `closeRequested` correctly keeps `openSessionAllowed` true and renders "Finishing current step" rather than claiming a stop. No stale names left.

---

## Findings

### Medium

**M1 — `remainingSubagents` / `existingSubagentCount` still count promoted children, so the tool advertises a cap it no longer enforces.**
`context.rs:50-53` sets `existing_subagent_count = list_subagents(...).len()`, and `list_subagents` → `list_subagent_children` → `list_children_by_relation` (`links/store.rs:214-231`) filters on `relation='subagent' AND closed_at IS NULL` with **no** `promoted_at` filter. `calls.rs:256-259` then reports `existingSubagentCount` and `remainingSubagents = max - existing`. After this PR the real cap (store subselect + service pre-check) excludes promoted rows, so the two disagree.
Failure scenario: a parent spawns 8 subagents and promotes 3. `get_subagent_launch_options` returns `existingSubagentCount: 8, remainingSubagents: 0` while `can_create` is `true` and `spawn_subagent` succeeds three more times. An agent that trusts the numbers (the tool description tells it to call this first) stops delegating for no reason; an agent that ignores them learns the numbers are noise. Fix is one filter in `resolve_context`, mirroring `subagents/service.rs:139`.

**M2 — The deferred half of a soft close never re-checks `Running`, so it can hard-kill mid-step — the exact thing the new invariant promises it won't.**
`hooks.rs:94-154` looks up the pending request, takes the gates, and calls `close_live_session` unconditionally. Meanwhile the target is *not* closed while end-requested, so its pending-prompt queue keeps draining. `on_turn_finished` is invoked synchronously from the actor's turn-finish callback (`runtime/startup.rs:490-506`) and the hook immediately `tokio::spawn`s; the actor's idle loop is free to pick up a queued prompt and start turn N+1 before the spawned task acquires the `Close` permit.
Failure scenario: parent closes a working subagent; someone had queued a prompt to it; turn N ends, turn N+1 starts, the hook fires ~milliseconds later and shuts the actor down mid-tool-call. That contradicts `specs/anyharness/sessions.md:786` ("A close of a working agent never interrupts its in-flight step"), which this PR added. Note the tension — a naive re-check would let a chatty agent defer forever — so the honest fix is "defer only the turn that was running when the request was stamped" (compare `turn_id`/timestamp), or fence the queue drain once a close is pending. At minimum the invariant sentence should be qualified.

**M3 — An end-requested agent keeps accepting prompts, and nothing tells the sender.**
There is no gate anywhere on sending to a session with `closed_by_session_id IS NOT NULL AND closed_at IS NULL`. Ruling 6 covers *closed* sessions only, and the ADR is silent on end-requested ones — but before this PR the window did not exist, and now it can be a whole turn wide.
Failure scenario: A closes C with reason "superseded"; B (or a human) sends C three messages during the remaining step; they enqueue successfully; the hook closes C; the prompts are stranded in `session_pending_prompts` forever with no error to their senders and no receipt. Either reject sends to an end-requested target (and say why), or explicitly rule that they're accepted and drop-on-close — but it should be a decision, not a gap.

**M4 — A request stamped while `Running` has no recovery path if the turn never finishes.**
`is_mid_turn` (`calls.rs:1093-1101`) correctly derives `Running` only from a live snapshot (`execution_summary.rs:34-41` returns `Idle` with no handle), so a stale DB status can't strand the *initial* call — good. But once the stamp is written and the runtime dies before turn finish, the row stays end-requested and the session stays open; there is no startup sweep over `find_pending_close_request`. `specs/anyharness/sessions.md` (soft-close section) asserts "a runtime that restarts mid-turn still owes the close and the next finished turn pays it" — true only if a *next turn* ever happens, which requires somebody to prompt an agent that has already been told to stop (and which, per M2, will then run that whole turn before dying).
Concrete: parent closes a mid-turn subagent, gets "it will finish the step, then stop", the machine reboots; the subagent shows up as a live agent indefinitely. A boot-time reconciliation pass (close every session with a pending request that has no live handle) closes this, and is a handful of lines given the partial index already exists.

**M5 — The lock-order guards pin call *presence and order*, not that the permit is *held*.**
`peer_ops.rs:470-513` and `:515-540` assert `permit_at < lease_at < close_at` via `str::find`, and `calls.rs`/`hooks.rs` today do the right thing (`let _admission_permit = …`, `let _permit = match …`). But the tests never look at the binding. Rewriting `let _admission_permit = admit_peer_mutation(...)` as `let _ = admit_peer_mutation(...)` drops the permit at the end of the statement, destroys the fence entirely, and **every assertion still passes**. That is the failure mode the M-1 precedent on the base branch exists to prevent. One line each fixes it:
```rust
assert!(body.contains("let _admission_permit = admit_peer_mutation("));
assert!(body.contains("let _target_workspace_lease = lease_target_workspace_for_peer_write("));
```
(and `let _permit = match admit_peer_mutation(` / `let _lease = match …` in the hooks guard).

### Low

**L1 — `function_body` slices to the wrong boundary for `call_tool`.**
`peer_ops.rs:428-438`: `.or_else(|| rest[1..].find("\nfn "))` only fires when there is **no** later `\nasync fn ` anywhere in the remainder — so the plain-`fn` fallback is dead whenever an async fn follows further down. I ran the helper against the real bytes of `calls.rs@5d4ff2fc5`: `function_body(src, "pub async fn call_tool(")` returns **lines 46–290** (11,330 chars), i.e. all of `call_tool` *plus* the whole of `fn get_subagent_launch_options`, because the next `\nasync fn ` is `spawn_subagent` at line 291. It happens to be harmless for the three assertions in that test today (all three needles are inside `call_tool`, and the pre-gate slice ends before the overrun), but the guard would also be satisfied by text in a neighbouring function, which is precisely the window-slicing failure earlier guards hit. Take the earliest of the two matches rather than `or_else`:
```rust
let end = [rest[1..].find("\nasync fn "), rest[1..].find("\nfn ")].into_iter().flatten().min();
```
(`close_agent` slices correctly — lines 1017–1092 — so only the `call_tool` guard is affected.)

**L2 — `record_close_attribution` is not "written once" during the end-requested window.**
`links/store.rs:336-351` guards only on `closed_at IS NULL`, so the guard is WHERE-clause atomic (good) but a *second* `close_agent` on a still-open, already-requested row overwrites `closed_by_session_id` and `close_reason`. The service docstring (`ownership/service.rs:181-183`) and the PR body both say attribution "is written once"; that is true only after the close lands. Either add `AND closed_by_session_id IS NULL` (first requester wins, matching the docstring) or amend the docstring to "last requester before the close wins". The existing test (`ownership/service.rs:532-587`) only exercises the post-close case, so this window is unpinned either way.

**L3 — `close_result` reports `closed` from the link, not the session.**
`calls.rs:1038` returns early if *either* the link or the target session is closed, but `calls.rs:1116` computes `"closed": settled.closed_at.is_some()` from the link alone and the message says "That agent is closed." If the two ever diverge (a link closed while the session survived — e.g. a future `delete_link`/repair path), the tool reports a stop that did not happen. Cheap hardening: `settled.closed_at.is_some() || owned.target.closed_at.is_some()`.

**L4 — `SPAWN_STYLE_TOOL_NAMES` is a hand-maintained list with no ratchet for PR6/PR7.**
`tools.rs:50-54` is the whole of ruling 3's enforcement, and `spawn_agent` / `spawn_workspace` do not exist yet. The alias-coverage test only proves aliases of *listed* canonicals are listed; nothing fails if PR6 lands `spawn_agent` and forgets to add it, at which point an unpromoted subagent silently regains a spawn tool. A test asserting every advertised tool whose name starts with `spawn_` is in `SPAWN_STYLE_TOOL_NAMES` would make the omission impossible.

**L5 — `find_subagent_parent` is `LIMIT 1` with no `ORDER BY`.**
`links/store.rs:234-250`. `UNIQUE(relation, parent, child)` permits two open `subagent` rows naming the same child from different parents, so if that state ever arises, `is_unpromoted_subagent` (and therefore the spawn gate and the depth lift) is decided by whichever row SQLite returns. `link_child` only ever links freshly created children, so this is not reachable today — but the predicate that decides "may this agent spawn" should not be order-dependent. Add `ORDER BY (promoted_at IS NULL) DESC` (fail-closed) or `created_at ASC`.

**L6 — No HTTP/contract surface for the human half of ruling 7.**
Ruling 7 says "Parent OR human (right pane UI) can promote", and §4 wants the pane to render "Closed by X · reason". `ChildSubagentSummary` / `ParentSubagentLinkSummary` (`api/http/subagents.rs`) were not widened and no promote route exists, so PR9 will need a contract change plus SDK regen. Fine as sequencing — flagging so it isn't discovered late.

**L7 — Cascaded descendants get neither a permit nor a soft close.**
`close_session_tree` recurses into children without taking their `SessionMutationKind::Close` permits, and a mid-turn *grandchild* is killed outright while the directly-targeted agent is politely deferred. This is pre-existing behavior that this PR inherits rather than introduces, and the asymmetry may well be intended — noting it because the new spec text reads as a general guarantee.

---

## Test verification notes

25 new tests; they pin behavior rather than implementation almost everywhere — the exceptions are the three source-order guards, which are text guards by construction and justified in-comment.

**Source-order guards re-executed exactly.** I reimplemented `calls_source` / `function_body` and each assertion against the real bytes of `calls.rs` and `hooks.rs` at `5d4ff2fc5`:
- `the_spawn_gate_runs_before_dispatch_not_only_in_the_tool_list`: gate at offset 865 < dispatch at 1194 ✓; no `canonical_tool_name(name)` before the gate ✓. Window is over-wide (L1).
- `close_agent_takes_the_targets_permit_before_any_workspace_lease`: permit 1131 < lease 1301 < stamp 1593 < close 2014 ✓; `SessionMutationKind::Close` present between permit and lease ✓. Window lines 1017–1092, correctly scoped.
- `the_deferred_half_…_same_order`: permit 4816 < lease 5239 < close 6026 ✓; the `use` import does not false-match because the needles include `(`.
All three pass; none of them would fail under the `let _ =` neutering described in M5.

**Negative controls (three claimed).** All three discriminate against *deletion*:
- cascade guard → `a_promoted_subagent_survives_its_former_parents_close` (`runtime/tests.rs`) fails on the reverted predicate. Genuine.
- dispatch spawn gate → the guard's `.expect("call_tool gates the spawn-style tools at dispatch")` panics; it also catches the subtler "canonicalize first" bypass, because that rewrite changes the needle. Genuine.
- close permit → the guard's `.expect(...)` panics. Genuine for deletion, blind to neutering (M5).

**Other controls worth noting as real:** the sibling assertion in `promotion_lifts_the_spawn_block_for_that_child_only` (`ownership/service.rs:498-502`) would fail if promotion were read globally rather than per-row; `a_capped_parent_still_sees_its_launch_options` (`tools.rs`) would fail if the tool-list gate were written as `!can_create`. `delegated_coding_and_review_children_are_unaffected_by_promotion` is the weakest of the set — nothing stamps `promoted_at` on cowork/review rows, so it can only fail if the `match` is restructured to test promotion before relation — but it is a legitimate structural guard, not an inert assertion.

**Coverage gaps.** No test drives `call_tool("close_agent", …)` or `complete_requested_close` end to end: the entire soft-close decision (`is_mid_turn` phase mapping, the pre-gate idempotent return, the `closeRequested` response shape) and the whole deferred path are pinned only by string searches plus the client-side presentation tests. Given M2/M3/M4 all live there, one integration-level test over the runtime fixture — request a close on a `Running` session, finish the turn, assert the session closed and the link stamped — would be worth more than any of the text guards. Also unpinned: the end-requested attribution-overwrite window (L2), and `resolve_owned` on a *promoted* child framed as "the parent can still close it" (it is exercised incidentally at `ownership/service.rs:453-456`, but as a re-promotion, not as ruling 7's close right).

pablonyx:
Pushed `1c50f43bf` — all seven accepted findings plus the lows, spec text in the same commit.

**M2** is fixed with a synchronous turn-start fence rather than the recorded-turn-id fallback. `run_turn`'s drain loop always breaks, so `run_idle` is the single turn-start point and has exactly two sites; both now consult the durable request through a new `PendingCloseRequests` port on `ActorCapabilities`, served by the partial index 0068 already ships. No schema change. The check is fail-closed — a lookup error blocks the turn, because a turn wrongly not started is recoverable and a step killed in the middle is not. `live/sessions/actor/tests/soft_close.rs` drives the real actor against a fake ACP agent: turn N in flight → close stamped → prompt queued → turn N finishes → nothing reaches the agent, the pending row survives, the request stays armed. Its sibling test is the control: same setup without the stamp, the prompt becomes turn N+1.

**M3**: an agent's `send_agent_message`/`send_subagent_message` to an end-requested target is refused and told the target is finishing its final step. The **human route gets no new refusal** — the turn-start fence above is the cheap seam, so a person's prompt is accepted and stored durably like any prompt to a busy session and simply never starts. The spec now says so, and says what happens to prompts already queued behind the final turn: `close_session_tree` does not delete `session_pending_prompts`, so they are kept and never run.

**M6** (raised from the PR6 branch — promotion lifted the depth cap but not the born-with `subagents_enabled = false`, making the depth lift dead code for every real child): spawn validation treats a promoted caller as enabled instead of rewriting the session row. `should_attach` mounts unconditionally and this validation is the only server-side consumer of the flag, so promotion stays one durable fact on the link rather than two rows in two tables that can diverge.

M1 collapses the cap onto one SQL predicate shared by the pre-check, the advertised limits, and the insert subselect. M4 sweeps owed closes at boot through the same body the turn-finish hook uses, skipping targets that still have a live handle. M5's guards now pin the guard bindings by name. Lows: L1 min-of-two, L2 first-requester-wins attribution, L3 `closed` ORs link and session, L4 spawn-tool ratchet, L5 unpromoted-first ordering, L7 spec sentence that cascades close descendants directly and the soft-close guarantee is for the targeted agent. L5 and L3 have no dedicated test — the first is a tiebreak on a shape uniqueness makes unreachable today, the second is asserted through the existing close tests.

Negative controls, each run and each failing only its own test: drain fence removed → the soft-close integration test fails; refusal removed → the end-requested test fails; `let _admission_permit` → `let _` → the close_agent order guard fails; promotion lift removed → the promotion test fails.

Scoped suite 147/147; full `--lib` 1686 passed with only the known pre-existing `sweep_tests::…scratch_roots` red. `check_docs.py` and `check_migration_heads.py` pass. Untouched and still red on this branch from earlier in the stack (lint reds, per the run's ground rules; will be swept at the top of the stack with founder-flagged allowlist entries where the constitution requires): `check_max_lines.py` (the `agent_ops` domain is unallowlisted), `check_session_mutation_admission.py` (`sessions_wakes::schedule_agent_wake` unclassified), and `check_anyharness_boundaries.py` (`transcript_read.rs`, `wakes/hooks.rs`, four stale allowlist entries for the renamed `subagents/mcp/*` paths).

pablonyx:

PR #1732, opened 2026-08-08
feat(anyharness): spawn_agent — peer agents over one shared session-creation routine

pablonyx:
Sixth slice of the agent-operations stack (ADR §6 step 6; stacked on #1731 → #1729 → #1728 → #1727 → #1726). Adds `spawn_agent` and the shared session-creation routine.

## What this does

`spawn_subagent` had the whole session-creation core inlined — durable record, link row, wake arming, runtime start, first prompt, and the unwind for each of those failing. A peer spawn needs all of it, so the core moves to `agent_ops/spawn_ops.rs` as `create_agent_session`, parameterised by an ownership mode. `spawn_subagent` refactors onto it with no behaviour change and the same response shape.

`spawn_agent` is the second mode: it creates a peer the caller owns. The link row is `relation='owned_agent'`, so the close cascade already declines to follow it, the subagent fanout cap and depth limit don't apply, and the child is born with the full tool surface rather than as an unpromoted subagent. Its wake is session-scoped (no link completion to hang off) and its first prompt is an envelope-wrapped peer message rather than carrying a `session_link_id`.

The fanout cap is why `can_spawn_agent` is its own context field rather than derived from `can_create`: an owner holding eight subagents may still spawn a peer. Gating is otherwise shared — `spawn_agent` joins `SPAWN_STYLE_TOOL_NAMES`, so an unpromoted subagent is neither offered it nor allowed to dispatch it, and `MUTATING_TOOL_NAMES`, because it creates in the caller's own workspace.

Client-side, `spawn_agent` was unclassified on both sides of the wire, and `formatSubagentHeaderVerb` falls back to "Subagent created" — so the one tool whose point is that the new agent is *not* a subagent was the one being labelled as one. Now reads "Spawning agent" / "Agent spawned", and stays out of the subagent launch ledger.

## Spec note

ADR §3.4 lets `spawn_agent` name any workspace. Implemented here as the caller's workspace only — a cross-workspace spawn would have to lease a workspace the route never took, which is PR7's `spawn_workspace` problem. The argument is accepted and validated so the wire shape is already right; a mismatch is refused with both workspace ids named. §3.4 also said "no parent link", which contradicts §3.2's `owned_agent` row; the row wins (ADR amended).

Separately: `spawn_subagent` creates children with `subagents_enabled = false`, so a *promoted* subagent still hits `SubagentError::Disabled` and cannot spawn, contradicting ruling 7. Not fixed here — it is PR5 territory and has been routed to the PR5 fix pass on #1731.

## Testing

`cargo test -p anyharness-lib --lib -- agent_ops ownership subagents links session_admission` (122 passed); SDK reducer transcript suite (46); product-client subagent presentation suite (15); `check_docs.py` (209 files). Negative controls: writing `subagent` for the peer path, dropping `spawn_agent` from the dispatch gate, dropping the `wakeOnCompletion` passthrough, and removing the client header verb each fail a named test. The arg→request mapping was made a pure function and pinned with new tests, because no pre-existing behavioural test of the spawn flow exists (needs a live harness).

## Observability

No new metrics. Failed spawns unwind through the existing `tracing::warn` on wake-schedule cleanup; the response reports `wake.scheduled` / `wake.created` separately so an armed-but-not-created wake is visible to the caller rather than silent.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/app/test_support.rs
+++ b/anyharness/crates/anyharness-lib/src/app/test_support.rs
@@ -1,11 +1,24 @@
 use std::ffi::OsString;
 use std::sync::{Arc, Mutex, OnceLock};
 
+use crate::domains::repo_roots::service::RepoRootService;
+use crate::domains::repo_roots::store::RepoRootStore;
 use crate::domains::sessions::attachment_storage::PromptAttachmentStorage;
+use crate::domains::sessions::deletion::SessionDeleteWorkflow;
+use crate::domains::sessions::links::service::SessionLinkService;
+use crate::domains::sessions::links::store::SessionLinkStore;
 use crate::domains::sessions::live_ports::SessionAttachmentSource;
 use crate::domains::sessions::mcp_bindings::crypto::DATA_KEY_ENV_VAR;
 use crate::domains::sessions::store::SessionStore;
+use crate::domains::sessions::subagents::service::SubagentService;
+use crate::domains::sessions::subagents::store::SubagentStore;
+use crate::domains::terminals::store::TerminalStore;
+use crate::domains::workspaces::access_gate::WorkspaceAccessGate;
+use crate::domains::workspaces::deletion::WorkspaceDeleteWorkflow;
+use crate::domains::workspaces::runtime::WorkspaceRuntime;
+use crate::domains::workspaces::store::{WorkspaceAccessStore, WorkspaceStore};
 use crate::live::sessions::model::ActorCapabilities;
+use crate::live::terminals::TerminalService;
 use crate::persistence::Db;
 
 /// Store-backed [`ActorCapabilities`] for tests: the same wiring as
@@ -161,6 +174,59 @@ pub(crate) fn insert_session_row(
         origin: Some(crate::origin::OriginContext::system_local_runtime()),
     };
     store.insert(&record).expect("insert session row");
+}
+
+/// A store-backed [`SubagentService`] and the link service behind it.
+///
+/// The spawn gates (`validate_parent_can_spawn`,
+/// `validate_caller_can_spawn_agent`) read the session store, the link store,
+/// the workspace record and the access gate together, so proving what they
+/// refuse needs the real wiring rather than a stub — this is the same
+/// composition `app/mod.rs` builds, over an in-memory db. The link service is
+/// handed back because the state those gates read (a subagent link, a promoted
+/// one, eight of them) is written through it.
+pub(crate) struct SubagentServiceFixture {
+    pub service: SubagentService,
+    pub links: SessionLinkService,
+    pub sessions: SessionStore,
+}
+
+pub(crate) fn subagent_service_fixture(db: &Db) -> SubagentServiceFixture {
+    let sessions = SessionStore::new(db.clone());
+    let links = SessionLinkService::new(SessionLinkStore::new(db.clone()), sessions.clone());
+    let session_delete_workflow = SessionDeleteWorkflow::new(db.clone());
+    let runtime_home = std::env::temp_dir().join(format!(
+        "anyharness-subagent-service-test-{}",
+        uuid::Uuid::new_v4()
+    ));
+    let workspace_runtime = Arc::new(WorkspaceRuntime::new(
+        WorkspaceStore::new(db.clone()),
+        WorkspaceDeleteWorkflow::new(db.clone(), session_delete_workflow.clone()),
+        RepoRootService::new(RepoRootStore::new(db.clone())),
+        runtime_home.clone(),
+    ));
+    let access_gate = Arc::new(WorkspaceAccessGate::new(
+        WorkspaceStore::new(db.clone()),
+        sessions.clone(),
+        WorkspaceAccessStore::new(db.clone()),
+        Arc::new(TerminalService::new(
+            TerminalStore::new(db.clone()),
+            runtime_home,
+        )),
+    ));
+    let service = SubagentService::new(
+        sessions.clone(),
+        session_delete_workflow,
+        links.clone(),
+        SubagentStore::new(db.clone()),
+        workspace_runtime,
+        access_gate,
+    );
+    SubagentServiceFixture {
+        service,
+        links,
+        sessions,
+    }
 }
 
 pub(crate) fn seed_workspace_with_repo_root(db: &Db, workspace_id: &str, kind: &str, path: &str) {
--- a/anyharness/crates/anyharness-lib/src/domains/sessions/agent_ops/calls.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/sessions/agent_ops/calls.rs
@@ -4,8 +4,8 @@ use anyharness_contract::v1::ConfigApplyState;
 use serde_json::{json, Value};
 
 use super::calls_helpers::{
-    default_model_for_agent, initial_config_string, launch_agents_to_json, mode_options_to_json,
-    prompt_outcome_label, summaries_to_json,
+    cleanup_wake_schedule_after_failed_dispatch, default_model_for_agent, launch_agents_to_json,
+    mode_options_to_json, prompt_outcome_label, summaries_to_json,
 };
 use super::config_ops::{
     compose_agent_config_options, controls_to_json, current_selection_to_json,
@@ -17,12 +17,13 @@ use super::peer_ops::{
     authorize_transcript_read, consume_reply_wake, lease_target_workspace_for_peer_write,
     prepare_agent_message,
 };
+use super::spawn_ops::{create_agent_session, CreateAgentSessionRequest};
 use super::tools::{
     canonical_tool_name, is_spawn_style_tool, ChildSessionArgs, CloseAgentArgs, ConfigureAgentArgs,
     CreateSubagentArgs, GetAgentConfigOptionsArgs, ListAgentsArgs, PromoteSubagentArgs,
     ReadAgentTranscriptArgs, ReadAgentTranscriptMode, ReadSubagentEventsArgs,
     ReadSubagentLatestTurnsArgs, ScheduleAgentWakeArgs, SearchSubagentTranscriptArgs,
-    SendAgentMessageArgs, SendSubagentMessageArgs,
+    SendAgentMessageArgs, SendSubagentMessageArgs, SpawnAgentArgs,
 };
 use super::AgentOpsPeerGates;
 use crate::domains::sessions::admission::SessionMutationKind;
@@ -42,7 +43,6 @@ use crate::domains::sessions::transcript_read::{
 };
 use crate::domains::sessions::wakes::service::AgentWakeService;
 use crate::integrations::mcp::json_rpc::deserialize_args;
-use crate::origin::OriginContext;
 
 pub async fn call_tool(
     service: &SubagentService,
@@ -73,7 +73,11 @@ pub async fn call_tool(
         "get_subagent_launch_options" => get_subagent_launch_options(service, session_runtime, ctx),
         "spawn_subagent" => {
             let args: CreateSubagentArgs = deserialize_args(arguments)?;
-            spawn_subagent(service, session_runtime, ctx, args).await
+            spawn_subagent(service, ownership, wake_service, session_runtime, ctx, args).await
+        }
+        "spawn_agent" => {
+            let args: SpawnAgentArgs = deserialize_args(arguments)?;
+            spawn_agent(service, ownership, wake_service, session_runtime, ctx, args).await
         }
         "list_subagents" => service
             .list_subagents(&ctx.parent_session_id)
@@ -291,11 +295,12 @@ fn get_subagent_launch_options(
 
 async fn spawn_subagent(
     service: &SubagentService,
+    ownership: &AgentOwnershipService,
+    wake_service: &AgentWakeService,
     session_runtime: &SessionRuntime,
     ctx: &AgentOpsMcpContext,
     args: CreateSubagentArgs,
 ) -> anyhow::Result<Value> {
-    let parent_session_id = &ctx.parent_session_id;
     if !ctx.can_create {
         anyhow::bail!(
             "{}",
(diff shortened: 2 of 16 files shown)


pablonyx (commented):
## Review — PR #1732 (`spawn_agent` + shared session-creation routine)

**Verdict: the refactor is faithful — I diffed the old inline `spawn_subagent` body against `create_agent_session` + both call sites step by step and found no dropped or reordered step, no changed error surface, and no provenance drift. Three Medium findings, all about the NEW tool rather than the move: the `wakeOnCompletion` arm uses a reason that contradicts its own description, `spawn_agent` has one gate against subordinate callers where `spawn_subagent` has two, and two of the four claimed negative controls are narrower than the PR body states. No blockers.**

Reviewed as a delta only: `git diff 5d4ff2fc5..fac983580`, reading via `git show` at both revisions (the worktree is being edited concurrently). Not compiled — swap was at 3.6G/5G and the machine is shared; this is a reading review, CI covers the build.

### What checked out

**The extraction is line-for-line faithful.** Old body (`5d4ff2fc5:.../agent_ops/calls.rs:290-440`) vs `spawn_ops.rs:188-266`:

- Step order preserved exactly: `create_durable_session` → link → arm wake → `start_persisted_session` → `send_text_prompt_with_provenance`. Wake arming is still *before* dispatch (`spawn_ops.rs:226-232`), so PR3's arm-before-dispatch contract survives.
- Unwind ladder preserved: link failure → `delete_session` only (`spawn_ops.rs:215-221`); wake failure → `cleanup_child_session_after_failed_launch` (`:226-232`); start and prompt failure → wake cleanup *then* session delete, same order as the base (`:364-375`). Error wrapping is still `anyhow!("{error:?}")` at both dispatch failures.
- `create_durable_session` argument positions are identical, including the `bool` slot that used to be a literal `false` and is now `request.child_subagents_enabled()` (`spawn_ops.rs:196-210` vs base `:319-334`).
- Inheritance identical: `harnessId ?? caller.agent_kind`; `modelId` from `initialConfig` → `current_model_id` → `requested_model_id`; same for mode; label trimmed then emptied to `None`; empty-prompt bail with the same `"prompt is required"` string, still after `validate_parent_can_spawn` (`spawn_ops.rs:143-171`).
- `spawn_subagent`'s response shape is unchanged field-for-field, including `readCursor`, the duplicated `wakeScheduled`/`wakeScheduleCreated`/`wakeScope` legacy keys, and `label` (`created.link.label` is the value `link_child` stored from the same trimmed input — checked through `create_subagent_link_with_child_limit` → `record.label = input.label`).

**Ownership mode correctness.** Each of the three branches is right and nothing else varies by mode: relation via two distinct writers with a post-write `ensure!` that the row matches the mode (`spawn_ops.rs:277-300` — a good touch, it converts a mis-wired writer into a refusal + compensating delete rather than a mislabelled agent); wake link-scoped vs session-scoped (`:313-331`); first prompt `parent_to_child_provenance` vs `agent_message(...).into_parts()` (`:342-356`), and the peer envelope is the *same* construction `prepare_agent_message` uses (`peer_ops.rs:85-86`), so the first message and every later one cannot drift.

**Fanout cap really doesn't reach owned agents.** All three places check out: the store subselect is `WHERE relation = 'subagent'` (`links/store.rs:74`), the service pre-check uses `list_subagent_children` which is also relation-filtered (`links/store.rs:345`), and `link_owned_agent` goes through plain `create_link`, not the capped insert (`ownership/service.rs:147-165`). `an_owned_agent_link_does_not_claim_a_subagent_parent_slot` pins it from the other side.

**Workspace validation.** No path leaks a foreign id: the check lives in the only constructor of the owned request (`spawn_ops.rs:111-131`), `create_agent_session` creates from `request.workspace_id`, and that field is assigned `caller.workspace_id` — not the argument — so even a trim/compare bug can't move the session. It lines up with the two outer layers: `resolve_context` already rejects `parent.workspace_id != request.workspace_id` (`context.rs:36-42`), and the route's `SubagentWrite` lease is over that same workspace (`app/product_mcp.rs:149`). Refusal names both ids. Correct layering.

**Route lease.** `spawn_agent` in `MUTATING_TOOL_NAMES` is right for the stated reason — no target session, so no target permit to take outside the lease, and the only workspace mutated is the leased one. Mirror of `close_agent`'s exclusion, and the comment says so (`tools.rs:35-46`).

**L4 ratchet interplay.** `an_unpromoted_subagent_is_offered_no_spawn_style_tool` iterates `SPAWN_STYLE_TOOL_NAMES`, so adding `spawn_agent` to that list auto-extends the ratchet; the `&& !ctx.is_unpromoted_subagent` in `build_tool_list:377` satisfies it, and `unpromoted_subagent_context()` hardcodes `can_spawn_agent: true`, so the new test genuinely discriminates on the promotion half rather than passing for the wrong reason.

**Unwind leaves no orphans.** Traced `delete_session` → `SessionDeleteWorkflow::delete_session_graph_in_tx` (`deletion.rs:31-47`) → `delete_session_link_rows_for_session_in_tx` (link rows + link wake schedules, both sides) → `delete_session_rows_in_tx` → `delete_agent_wake_rows_for_session_in_tx`, which deletes `session_wake_schedules WHERE watcher_session_id = ?1 OR target_session_id = ?1` (`store/agent_wakes.rs:396-406`). So the owned-agent path's decision to skip the explicit wake delete (`spawn_ops.rs:371`) is sound, and the comment's reasoning is accurate.

**SDK/client.** Classification correct (`transcript.ts:1433` peer-ops list); deliberately absent from `isSubagentCreationAction`, and I confirmed that is the only thing driving the `subagent_creations` grouping (`transcript-presentation.ts:78`) — so a peer spawn renders as a normal/collapsible item rather than being folded under a parent's fanout. Verbs read right in both states.

---

### Findings

#### Medium

**M1 — `wakeOnCompletion` arms `ExplicitSchedule`, so a peer that *does* reply wakes the owner twice.**

`arm_spawn_wake` uses `AgentWakeReason::ExplicitSchedule` for the owned path (`spawn_ops.rs:324-331`). `consume_reply_agent_wake` deletes only rows with `armed_for_reply = 1` (`store/agent_wakes.rs:173-187`), so when the peer replies via `send_agent_message`, `consume_reply_wake` cannot take this row off. The tool's own arg description says the opposite (`tools.rs:399`): *"Its reply already wakes you with the full message; this is the safety net for one that answers nobody."* That sentence is verbatim the contract of the **`Reply`** reason (`store/agent_wakes.rs:48-50`: "the safety net for an answer that may never come, so the answer itself consumes it"), which is what `send_agent_message`'s `wakeOnReply` arms (`calls.rs:638`).

Failure scenario: agent calls `spawn_agent(wakeOnCompletion: true)`; the peer replies mid-turn → owner receives the reply prompt; the peer's turn finishes → `consume_agent_wakes_for_target` fires the still-armed row → owner receives a second, contentless wake pointer for the same turn. The owner burns a turn on a pointer to a message it already has.

Second-order: `arm_agent_wake` keeps the **stronger** reason and never downgrades (`store/agent_wakes.rs:83-90`). So any `send_agent_message(peer, wakeOnReply: true)` the owner issues *before* the peer's first turn finishes is silently absorbed into the ExplicitSchedule row, and that send's reply-consumption is disabled too — `wakeOnReply` quietly stops meaning what it says for the whole window.

Either arm `Reply` (matching the description, `send_agent_message`, and the fact that a spawn *is* a send awaiting an answer), or change the description to say the wake always fires. The code comment argues for ExplicitSchedule ("nothing has been replied to, so no later message may consume it") — but the peer's first turn is exactly the thing being waited on, and its reply is exactly the answer.

**M2 — `spawn_agent` has one gate against subordinate callers where `spawn_subagent` has two.**

`validate_caller_can_spawn_agent` (`subagents/service.rs:135-144`) deliberately drops the depth rule with the rest of the subordination checks, so it returns `Ok` for an unpromoted subagent. `resolve_context` therefore sets `can_spawn_agent = true` for a session that must never spawn (`context.rs:64-68`), contradicting the field's own doc comment ("Whether this session may spawn a PEER agent", `context.rs:14-19`). `spawn_agent`'s body then checks only `!ctx.can_spawn_agent` (`calls.rs:366`).

So the *only* thing enforcing ruling 3 for this tool is the `is_spawn_style_tool(name)` check at the top of `call_tool`. By contrast `spawn_subagent` is protected twice: that same dispatch check, **and** `ctx.can_create == false` — because `validate_parent_can_spawn` still returns `DepthLimit` for an unpromoted subagent. `build_tool_list` already has to compensate with an explicit `&& !ctx.is_unpromoted_subagent` (`tools.rs:377`) that `can_create`'s consumers don't need, which is the asymmetry showing through.

Failure scenario: any future path reaching `calls::spawn_agent` without the top of `call_tool` — a new HTTP route, a direct service call from a hook, or someone moving the gate into the match arms — hands an unpromoted subagent unlimited peer spawning (uncapped, undepthed, and surviving its parent's close), while `spawn_subagent` stays correctly blocked. That is precisely the escape ruling 3 exists to prevent, and the spawn-block ratchet would not catch it because the ratchet tests `tools/list`, not dispatch.

Cheapest fix: make `can_spawn_agent` false for an unpromoted subagent (fold the `find_subagent_parent` / `is_unpromoted_subagent` check into `validate_caller_can_spawn_agent`, or `&&` it in `resolve_context`), so the flag means what its name and doc claim and the body check is a real second gate.

**M3 — two of the four claimed negative controls are narrower than the PR body states, and the cap/`can_spawn_agent` split is unpinned.**

- *"dropping `spawn_agent` from the dispatch gate"* is pinned only as `is_spawn_style_tool("spawn_agent")`. Nothing in the crate calls `agent_ops::calls::call_tool` (I checked every `call_tool(` site) — so deleting the `if ctx.is_unpromoted_subagent && is_spawn_style_tool(name)` block itself passes the entire suite. Only removal from the *name list* is controlled, not removal of the gate.
- *"dropping the `wakeOnCompletion` passthrough"* is pinned only at the arg→request mapping. Replacing `arm_spawn_wake`'s body with `Ok(false)` (`spawn_ops.rs:303`) passes every test, and the response would still report `"scheduled": true` — an agent would be told it will be woken and never be.
- `a_capped_parent_can_still_spawn_a_peer` uses `context()`, which hardcodes `can_spawn_agent: true` (`tools.rs:556-560`). It proves `build_tool_list` reads the right *field*; it does not prove `validate_caller_can_spawn_agent` keeps the fanout cap out. There is no test on `validate_caller_can_spawn_agent` at all — reintroducing the cap into `validate_caller_can_create`, the exact regression the split exists to prevent, is green. This is cheap to close: the `ownership/service.rs` test fixture already stands up a real store.

The two that do discriminate cleanly: the relation pair (`the_ownership_mode_picks_the_relation_the_link_row_gets` + `spawning_a_peer_writes_an_owned_agent_row_the_owner_resolves` + the `ensure!` at `spawn_ops.rs:294`), and the client header verb.

#### Low

**L1 — the cowork depth refusal silently doesn't apply to `spawn_agent`.** `validate_parent_can_spawn` refuses a child of a `CoworkCodingSession` link with `DepthLimit` (`subagents/service.rs:161-167`). `validate_caller_can_spawn_agent` drops it along with the subagent depth rule, and `is_unpromoted_subagent` only inspects `relation = 'subagent'` (`links/model.rs:123`), so the dispatch gate doesn't catch it either. A cowork coding session that cannot spawn a subagent can spawn unlimited peers — and unlike a subagent, they outlive it (`cascades_to_child` returns `false` for `OwnedAgent`, `runtime/lifecycle.rs:350`). Low because the ADR calls cowork unused and slated for deletion, but the stated justification for dropping the checks is about *subagent* subordination and doesn't cover this one.

**L2 — the spec now claims a service-level refusal that doesn't exist.** `specs/anyharness/sessions.md` adds `spawn_agent` to the spawn-style list an unpromoted subagent is denied and, in the same sentence, says "…and `validate_parent_can_spawn` refuses it with a depth limit". `spawn_agent` never calls `validate_parent_can_spawn`. The doc asserts exactly the defence M2 says is absent.

**L3 — §3.4's "creator recorded … in origin metadata" is neither implemented nor disclosed.** `create_agent_session` stamps `OriginContext::system_local_runtime()` for both modes (`spawn_ops.rs:208`), and `origin.rs` has no agent-created variant, so a spawned peer is indistinguishable in `origin` from any system-created session. The functional half of §3.4 holds — the `owned_agent` row carries the owner — but the metadata half is silently deferred while §3.2's attribution note implies it lands somewhere in this stack. Worth a line in the PR body or an explicit follow-up.

**L4 — `initialConfig` advertises `additionalProperties: true` but only `modelId`/`modeId` are read** (`tools.rs:391-397`, `spawn_ops.rs:159-164`). Inherited verbatim from `spawn_subagent`, but `spawn_agent` re-advertises it on a brand-new tool: a caller passing `thinkingLevelId` or `systemPromptAppend` gets no error and no effect. `appliedInitialConfig` echoing only three fields is the only way to notice.

**L5 (observation, likely PR9) — nothing in the stream tells the client a peer was spawned.** `isSubagentMcpMutation` still lists only the subagent tools (`stream-side-effect-plan.ts:278-285`), and there is no server-side link-created event. Not invalidating the *subagent* list is correct (the peer isn't one), but that means a new top-level session appears only on an unrelated refresh. Flagging so it lands in PR9's scope explicitly rather than by omission — the disclosed deferral covers the receipt action, not this.

---

### Test verification notes

- New Rust tests read as real discriminators where they claim to be: the relation mapping, the peer-vs-subagent first prompt (envelope prefix, verbatim body, reply suffix, and `session_link_id: None`), the workspace refusal naming the foreign id, blank-prompt rejection across both constructors, and `child_subagents_enabled()` per mode. `an_unpromoted_subagent_is_offered_no_way_to_spawn_a_peer_either` is genuinely discriminating because its fixture sets `can_spawn_agent: true`.
- `an_owned_agent_link_does_not_claim_a_subagent_parent_slot` is the strongest test in the PR — it checks the *absence* from three different readers rather than the presence in one.
- **Observable behavior the refactor leaves unpinned** (none of it was pinned before, so not a regression — but it is now shared by two tools, so a single edit moves both): the step order and PR3's arm-before-dispatch contract; every unwind path (emptying `unwind_after_failed_dispatch` is green); `spawn_subagent`'s response shape, specifically the two fields the refactor rewired — `label` (locally trimmed value → `created.link.label`) and `wakeScheduled` (`args.wake_on_completion` → `request.wake_on_completion`); that `child_subagents_enabled()` actually reaches the `subagents_enabled` column; `spawn_agent`'s response keys (`sessionId` as the peer handle, `agentId`, `ownership`).
- The one unpinned item that needs **no** live harness and would pay for itself: the wake mode split. `service.schedule_wake_for_target` and `wake_service.arm` are both store-backed, and `ownership/service.rs`'s fixture already builds a store — a test asserting a subagent spawn writes a link-scoped row while a peer spawn writes a `session_wake_schedules` row with `armed_for_reply = 0` would pin the split *and* make M1 a decision rather than an accident.

pablonyx:
Fixed in `c0f9566b9`.

**M1 — accepted, arm changed to `Reply`.** You are right that the arg description is the `Reply` contract verbatim, and that the failure mode is a double wake plus the absorption of any later `wakeOnReply` on the same peer. `arm_spawn_wake`'s owned branch now arms `AgentWakeReason::Reply`; the comment records why a spawn is a send awaiting an answer rather than a standing schedule.

**M2 — accepted, folded into `validate_caller_can_spawn_agent`.** It now reads the same predicate the depth rule reads (`find_subagent_parent` + `is_unpromoted_subagent`) and refuses with a new `SubagentError::Subordinate`, so `can_spawn_agent` is false for a subordinate caller and `calls::spawn_agent`'s own check is a real second gate. `subagents_enabled` for promoted callers is untouched — that stays with the PR5 fix pass.

**M3 — all three closed.**
(a) The wake arm is now exercised over real stores through the production `arm_spawn_wake`: the armed row's reason is asserted, the peer's reply consumes it and queues no pointer, a silent turn finish fires it, `wakeOnCompletion: false` arms nothing, and a subagent spawn writes no session-scoped row. Replacing the body with `Ok(false)` fails three of them.
(b) New `the_fanout_cap_does_not_reach_the_peer_spawn_gate`: eight open subagent links, `validate_parent_can_spawn` → `FanoutLimit`, `validate_caller_can_spawn_agent` → Ok. Reinstating the cap inside the peer gate fails it.
(c) The source-order guard now also asserts the gate body between the check and the dispatch constructs the refusal, so a gate that logs and falls through no longer passes.

This needed the first store-backed `SubagentService` fixture in the crate; it lives in `app/test_support` beside the existing ones and is the same wiring `app/mod.rs` builds.

**Lows.** L1 and L5 are disclosed in the specs rather than fixed (cowork is deleted in PR8; the stream side is client work). L2 corrected — the spec now names `validate_caller_can_spawn_agent`, which after M2 is a refusal that actually exists. L3 needs no code: the ADR was amended so the `owned_agent` row is the creator record, and neither spec file echoed the origin-metadata claim. L4: `spawn_subagent` advertises the same open `initialConfig`, so both stay open, with the reason and the "change both or neither" note now written down in `tools.rs` and the product spec.

Suite: `cargo test -p anyharness-lib --lib -- agent_ops ownership subagents wakes agent_wakes session_admission` → 144 passed (8 new). Full `--lib`: 1700 passed, 1 failed (the pre-existing `sweep_tests` scratch-roots red). `check_docs.py` green (209 files). Negative controls run and reverted for each of the five claims above.

pablonyx:
Rebased onto the PR5 fix head `1c50f43bf` → new head `5d245ed1e` (the fix commit above is now `5d245ed1e`, previously `c0f9566b9`). Three conflict seams, resolved to keep both fix passes' semantics: PR5's promoted-derived `Disabled` lift moved into the shared `validate_caller_can_create` so both spawn paths honour ruling 7; the spec prose union-merged; the two independently rewritten test modules unified on the shared `test_support::subagent_service_fixture` with PR5's promotion-lift tests ported onto it. Post-rebase: scoped suites 151/151, full `--lib` 1709 passed / 1 failed — only the known pre-existing `sweep_tests` scratch-roots red.

pablonyx:

PR #1733, opened 2026-08-08
feat(agent-ops): spawn_workspace and cross-workspace spawn_agent

pablonyx:
Part 7 of the agent-operations stack (ADR §6 step 7). Stacked on #1732.

## What this does

Adds `get_workspace_options` and `spawn_workspace` to the agent-ops MCP, and lifts PR6's caller-workspace restriction on `spawn_agent` so the two-step flow of ADR §5 works: describe → create → put an agent in it.

`spawn_workspace` calls the same runtimes the human routes call — `WorkspaceWorktreeRuntime::create_worktree` for `mode=worktree`, `create_workspace_with_origin_and_creator_context` for `mode=local`. It adds only what the human routes do not need: the local-only gate, the `WorkspaceCreatorContext::Agent` stamp, and server-side defaults for base branch, path, name-conflict policy and setup script (the command this machine last ran for the caller's workspace — no second detection policy). The agent passes three things and a label.

`spawn_agent` now takes any `workspaceId`, defaulting to the caller's. It therefore leaves `MUTATING_TOOL_NAMES` — the route's lease is the caller's workspace, the wrong one — and takes the target workspace's write lease in the call, held across creation, start and the first prompt. No admission permit exists or can: the target session does not exist until the call creates it. The launch selection is now composed against the target workspace's catalog inside the shared creation routine: a harness the target cannot launch is refused naming what it can; a model the caller NAMED that the target does not offer is refused; a model merely INHERITED from the caller is quietly replaced by the target's default.

Retirement stays human-only (ruling 11), with a name-and-source ratchet: no agent-ops tool reaches a retire, purge or delete path, asserted by name over the advertised list, both tool-name lists, the aliases, and the source of `workspace_ops.rs`.

## Spec notes

- Local-only is implemented as `has_local_checkout()` plus two filesystem facts (the caller's checkout still on disk, the target repo root a directory here), because no local-vs-cloud placement concept exists in the codebase — ADR §2 says as much. §1 req 1.7's "(if running locally)" parenthetical has no state to read today; raised as a contradiction, not silently resolved.
- ADR §3.4 says `spawn_workspace` is "promoted / top level agents only"; gated on `can_spawn_agent` (the peer-spawn signal) so the workspace pair and `spawn_agent` read the same subordination answer.

## Testing

Full `anyharness-lib --lib` on the rebased head: **1735 passed / 1 failed** — only the known pre-existing `sweep_tests::the_sweep_removes_only_abandoned_and_old_scratch_roots` red. Scoped suites (`agent_ops workspaces ownership subagents session_admission`) 269/269 pre-rebase. SDK vitest 48 passed, product-client vitest 17 passed. `check_docs.py` green (209 files).

Four negative controls run and reverted:
1. Local-only gate arm removed → `a_repo_this_machine_does_not_have_is_refused` failed.
2. Creator stamp, two levels — call sites dropped to `None` and the stamp's session id blanked → three tests failed. This control exposed that `creator_context` being an `Option` let a dropped stamp compile silently; the call-site ratchet added here is its product.
3. Launch selection composed against the caller instead of the target → 6 failures via PR4's disjoint-catalog CatalogSpy, including the call-site ratchet `the_real_resolver_is_handed_the_target_workspace_too` (also born from this control).
4. `spawn_agent` re-added to `MUTATING_TOOL_NAMES` → both lease-in-the-call ratchets failed.

## Observability

A cross-workspace `spawn_agent` logs caller session, caller workspace and target workspace at info — the one call that reaches out of its own workspace, and the one whose lease a retire preflight elsewhere can block on.

## Known deferrals

Stack-introduced lint reds (`check_max_lines`, session-mutation-admission classification, anyharness-boundaries allowlist) are swept at the top of the stack with founder-flagged allowlist entries, as disclosed on #1731.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/api/session_admission_tests.rs
+++ b/anyharness/crates/anyharness-lib/src/api/session_admission_tests.rs
@@ -727,3 +727,102 @@ fn the_peer_tools_take_the_permit_before_the_target_workspace_lease() {
         assert!(lease_at < dispatch_at, "{tool}: dispatch escaped the locks");
     }
 }
+
+/// One private function's body, squashed to single spaces. Same extraction the
+/// peer-tool guard above does inline; the workspace-spawn tools need it too.
+fn private_fn_body(rel_path: &str, fn_name: &str) -> String {
+    let path = std::path::Path::new(env!("CARGO_MANIFEST_DIR")).join(rel_path);
+    let text =
+        std::fs::read_to_string(&path).unwrap_or_else(|error| panic!("read {rel_path}: {error}"));
+    let signature = format!("async fn {fn_name}(");
+    let start = text
+        .find(&signature)
+        .unwrap_or_else(|| panic!("{rel_path}: {fn_name} not found"));
+    let rest = &text[start..];
+    let end = rest.find("\n}\n").map(|idx| idx + 3).unwrap_or(rest.len());
+    squash_whitespace(&rest[..end])
+}
+
+#[test]
+fn spawn_agent_holds_the_target_workspace_lease_across_creation() {
+    // `spawn_agent` can now create a peer in a workspace that is not the
+    // caller's (ADR §3.3), so it left `MUTATING_TOOL_NAMES` for the same reason
+    // `send_agent_message` and `configure_agent` did: the route's lease is the
+    // CALLER's workspace, which is the wrong one. It takes the TARGET
+    // workspace's write lease itself instead.
+    //
+    // Unlike the peer tools there is no admission permit, and none is possible:
+    // the target SESSION does not exist until the call creates it, so there is
+    // nothing to admit. With one lease and no permit LOCK-01 has no order to
+    // invert — what is left to prove is that the one lease is HELD across the
+    // creation, because that is what a concurrent retire preflight on the
+    // target workspace has to collide with.
+    assert!(
+        !crate::domains::sessions::agent_ops::tools::MUTATING_TOOL_NAMES.contains(&"spawn_agent"),
+        "spawn_agent must stay OUT of MUTATING_TOOL_NAMES — the route lease is the caller's \
+         workspace, and taking it here would both lease the wrong workspace and (were a permit \
+         ever added) invert LOCK-01"
+    );
+    let body = private_fn_body("src/domains/sessions/agent_ops/calls.rs", "spawn_agent");
+    let body = body.as_str();
+    // The BINDING, not the call: `let _ = lease_..(..)` drops the lease at the
+    // end of its own statement, which reads identically in call ORDER and
+    // fences nothing.
+    let lease_at = body
+        .find("let _target_workspace_lease = lease_target_workspace_for_peer_write(")
+        .expect(
+            "spawn_agent must BIND the TARGET workspace lease as `_target_workspace_lease` — \
+             an unbound `let _ =` drops it immediately",
+        );
+    let create_at = body
+        .find("create_agent_session(")
+        .expect("spawn_agent must create through the shared create_agent_session routine");
+    assert!(
+        lease_at < create_at,
+        "spawn_agent: the target workspace lease must be taken BEFORE create_agent_session, and \
+         it is held to the end of the function, so creation happens inside it"
+    );
+    // And the lease is on the TARGET, not on the caller's `ctx.workspace_id`.
+    assert!(
+        body.contains(
+            "lease_target_workspace_for_peer_write( &gates.workspace_operation_gate, \
+             service.access_gate(), &request.workspace_id,"
+        ),
+        "spawn_agent must lease `request.workspace_id` (the target), not the caller's workspace"
+    );
+}
+
+#[test]
+fn spawn_workspace_has_nothing_to_lease_and_says_so() {
+    // `spawn_workspace` creates a workspace, so at call time the thing a
+    // workspace operation lease would name does not exist. The human worktree
+    // route has the same problem and solves it the same way: no operation
+    // lease, an ACCESS-gate assertion on the repo root instead. Recording it
+    // here so a later reader does not "fix" the missing lease.
+    assert!(
+        !crate::domains::sessions::agent_ops::tools::MUTATING_TOOL_NAMES
+            .contains(&"spawn_workspace"),
+        "spawn_workspace must stay OUT of MUTATING_TOOL_NAMES — the route would lease the \
+         CALLER's workspace, which is not what the call mutates"
+    );
+    let body = private_fn_body(
+        "src/domains/sessions/agent_ops/workspace_ops.rs",
+        "spawn_workspace",
+    );
+    let body = body.as_str();
+    assert!(
+        !body.contains("acquire_shared(") && !body.contains("acquire_exclusive("),
+        "spawn_workspace must not take a workspace operation lease — there is no workspace to \
+         lease until it returns"
+    );
+    let gate_at = body
+        .find("assert_can_mutate_for_repo_root(")
+        .expect("spawn_workspace must gate on the repo root, as the human worktree route does");
+    let create_at = body
+        .find("spawn_worktree_workspace(")
+        .expect("spawn_workspace must delegate creation to the worktree path");
+    assert!(
+        gate_at < create_at,
+        "spawn_workspace: the repo-root access gate must run before anything is created"
+    );
+}
--- a/anyharness/crates/anyharness-lib/src/app/mod.rs
+++ b/anyharness/crates/anyharness-lib/src/app/mod.rs
@@ -592,6 +592,8 @@ impl AppState {
                 agent_wake_service: agent_wake_service.clone(),
                 session_runtime: session_runtime.clone(),
                 workspace_runtime: workspace_runtime.clone(),
+                workspace_worktree_runtime: workspace_worktree_runtime.clone(),
+                repo_root_service: repo_root_service.clone(),
                 agent_ownership_service: agent_ownership_service.clone(),
                 session_admission: session_admission.clone(),
                 workspace_operation_gate: workspace_operation_gate.clone(),
--- a/anyharness/crates/anyharness-lib/src/app/product_mcp.rs
+++ b/anyharness/crates/anyharness-lib/src/app/product_mcp.rs
@@ -5,14 +5,15 @@ use crate::domains::cowork::mcp::{
     self as cowork_mcp, auth::CoworkMcpAuth, tools as cowork_mcp_tools, CoworkProductMcpServer,
 };
 use crate::domains::cowork::runtime::CoworkRuntime;
+use crate::domains::repo_roots::service::RepoRootService;
 use crate::domains::reviews::mcp::{
     self as review_mcp, auth::ReviewMcpAuth, tools as review_mcp_tools, ReviewProductMcpServer,
 };
 use crate::domains::reviews::runtime::ReviewRuntime;
 use crate::domains::sessions::admission::SessionMutationAdmission;
 use crate::domains::sessions::agent_ops::{
-    self as agent_ops_mcp, auth::AgentOpsMcpAuth, tools as agent_ops_mcp_tools,
-    AgentOpsPeerGates, AgentOpsProductMcpServer,
+    self as agent_ops_mcp, auth::AgentOpsMcpAuth, tools as agent_ops_mcp_tools, AgentOpsPeerGates,
+    AgentOpsProductMcpServer, AgentOpsWorkspaceOps,
 };
 use crate::domains::sessions::mcp_bindings::product_catalog::ProductMcpLaunchCatalog;
 use crate::domains::sessions::ownership::service::AgentOwnershipService;
@@ -28,6 +29,7 @@ use crate::domains::sessions::wakes::service::AgentWakeService;
 use crate::domains::workspaces::model::WorkspaceSurface;
 use crate::domains::workspaces::operation_gate::{WorkspaceOperationGate, WorkspaceOperationKind};
 use crate::domains::workspaces::runtime::WorkspaceRuntime;
+use crate::domains::workspaces::worktree_runtime::WorkspaceWorktreeRuntime;
 
 pub(super) struct LaunchCatalogDeps {
     pub(super) runtime_base_url: String,
@@ -44,6 +46,10 @@ pub(super) struct EndpointRegistryDeps {
     pub(super) agent_wake_service: Arc<AgentWakeService>,
     pub(super) session_runtime: Arc<SessionRuntime>,
     pub(super) workspace_runtime: Arc<WorkspaceRuntime>,
+    // The agent ops server creates workspaces through the very services the
+    // human routes use; see `AgentOpsWorkspaceOps`.
+    pub(super) workspace_worktree_runtime: Arc<WorkspaceWorktreeRuntime>,
+    pub(super) repo_root_service: Arc<RepoRootService>,
     pub(super) agent_ownership_service: Arc<AgentOwnershipService>,
     pub(super) session_admission: Arc<SessionMutationAdmission>,
     pub(super) workspace_operation_gate: Arc<WorkspaceOperationGate>,
@@ -117,6 +123,8 @@ pub(super) fn build_product_mcp_endpoint_registry(
         agent_wake_service,
         session_runtime,
         workspace_runtime,
+        workspace_worktree_runtime,
+        repo_root_service,
         agent_ownership_service,
         session_admission,
         workspace_operation_gate,
@@ -137,7 +145,11 @@ pub(super) fn build_product_mcp_endpoint_registry(
                 subagent_service.clone(),
                 agent_wake_service,
                 session_runtime,
-                workspace_runtime.clone(),
+                AgentOpsWorkspaceOps {
+                    workspace_runtime: workspace_runtime.clone(),
+                    worktree_runtime: workspace_worktree_runtime,
+                    repo_roots: repo_root_service,
+                },
                 agent_ownership_service,
                 AgentOpsPeerGates {
                     session_admission,
--- a/anyharness/crates/anyharness-lib/src/domains/sessions/agent_ops/calls.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/sessions/agent_ops/calls.rs
@@ -96,10 +96,11 @@ pub async fn call_tool(
         "get_workspace_options" => {
             let caller = caller_session(service, &ctx.parent_session_id)?;
             workspace_ops::get_workspace_options(
-                &workspaces.workspace_runtime,
-                &workspaces.repo_roots,
+                workspaces.workspace_runtime.clone(),
+                workspaces.repo_roots.clone(),
                 &caller,
             )
+            .await
         }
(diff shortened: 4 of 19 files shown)


pablonyx (commented):
## Adversarial review — PR #1733 (agent-ops step 7)

**Frozen review pair:** base `5d245ed1e` (head of `agent-ops/6-spawn`, #1732) → head `32b947403`. Every line reference below is against `32b947403`.

**What I checked:** ADR §1 req 1.7 / §3.4 / §5 flow 4 / ruling 3 / ruling 11 fidelity; the in-call target-workspace lease against the base's `admit_peer_mutation → lease_target_workspace_for_peer_write` pattern; cross-workspace catalog composition; `spawn_workspace`'s path/branch/setup-script surface; the rebase seams (`tools.rs` lists, `spawn_ops.rs`, `subagents/service.rs`, `peer_ops.rs`); and whether the claimed tests bind on this head. No build run (machine limit), so compile-level claims are grep-verified, not compiled.

**What holds up.** The lease story is correct: `calls.rs:431` **binds** the lease (`let _target_workspace_lease = …`, not `let _ =`), it is taken on `request.workspace_id` (the target) and not `ctx.workspace_id`, everything before it is read-only (`assert_spawnable_workspace` at :422 is a pure read, `assert_workspace_can_be_mutated` at :427 is a check with no lease), and it lives to the end of the function so creation → start → first prompt all happen inside it. The caller-workspace mutability assertion the route stopped making is restored at :427 and again inside `validate_caller_can_spawn_agent` (`subagents/service.rs:169-171`). The `MUTATING_TOOL_NAMES` removal is right — the route lease (`api/http/product_mcp.rs:72-80`) is keyed on the URL workspace. `spawn_workspace`'s "no lease is possible" argument matches the human route exactly (`api/http/workspaces_worktrees.rs:85-99`: access gate on the repo root, `create_dir_all`, create — no operation lease). Path traversal via `branchName` is neutralised twice over: `validate_branch_name` (`workspace_ops.rs:346-365`) rejects leading `-` (no git arg injection) and the destination is slugged to `[a-z0-9_-]` by `destination_slug`/`sanitize_destination_name` (`workspaces/runtime/exact_ref.rs:387-408`), so `a/../../b` cannot escape the managed root even without the `..` check. `initialConfig` cannot smuggle anything: `inherit_from` reads only `modelId|model` and `modeId|mode` (`spawn_ops.rs:175-177`) and nothing else reaches `create_durable_session`. `subagents/service.rs` and `peer_ops.rs` are byte-identical to the base, so the PR6-fix ratchets (Subordinate refusal, `arm_spawn_wake → AgentWakeReason::Reply` at `spawn_ops.rs:442-444`) survive the rebase intact. Ruling 11 is respected and ratcheted.

---

### H1 — a worktree of repo B runs repo A's setup command

`workspace_ops.rs:405-414` reads the setup script as `worktree_runtime.latest_setup_command(&caller_workspace.id)` — the caller's own workspace — but the repo root is agent-chosen at `:293-304` and defaults to the caller's only when `repoRootId` is omitted. So `spawn_workspace(repoRootId: "<some other root>", mode: "worktree")` takes the shell command last run for the caller's repo and hands it to `create_worktree(setup_script: …)`, which executes it in the new checkout via `start_setup_for_created_workspace` (`worktree_runtime.rs:111-121` → `setup_runtime.rs:93-106`). A `pnpm install && pnpm build` vetted by a human for repo A then runs inside a fresh checkout of repo B, mutating lockfiles/`node_modules` in a repo nobody authorised that command for.

The module doc calls this "the setup this repo uses here" (`workspace_ops.rs:404-406`) and `get_workspace_options` tells the agent "whatever this machine last ran for **your own** workspace" (`:262`) — the docs describe the same-repo case and the code applies it to every case. `WorkspaceStore::list_active_by_repo_root_id` (`workspaces/store/listings.rs:40`) already exists, so the fix is small: reuse the command only when `repo_root.id == caller_workspace.repo_root_id`, else derive it from an active workspace of the **target** root, else run none. Please add a test — this whole branch has no behavioural coverage (see M6).

### H2 — `spawn_agent` still advertises "in your workspace"

`tools.rs:419` — the `spawn_agent` description is unchanged from PR6: *"Create a new top-level agent you own, **in your workspace**, and send it a first message."* The `workspaceId` property one screen down (`:447`) now says the opposite. `:429` compounds it: `harnessId` → *"See `get_subagent_launch_options` for what **this** workspace can launch"*, but `get_subagent_launch_options` composes the **caller's** workspace, while the harness is now validated against the target's catalog (`spawn_ops.rs:245-254`) — an agent that follows that instruction for a cross-workspace spawn gets a refusal it was told to expect success from.

This is the agent-facing contract for ADR §5 flow 4, and the tool list is frozen at launch, so a wrong description is not self-correcting. It needs to name the placement and point at `get_workspace_options` for the target's catalog. `specs/anyharness/sessions.md` and `subagents.md` were updated in this PR; the tool copy was not.

---

### M1 — every peer link is stamped `same_workspace`, including the ones that are not

`ownership/service.rs:155-165`:

```rust
workspace_relation: SessionLinkWorkspaceRelation::SameWorkspace,
// The agent is a peer, but it is still born in one workspace, and
// in this tool that is the caller's own.
```

That comment was true at the base and this PR makes it false, while leaving the value alone. Every cross-workspace `spawn_agent` now writes a durable `session_links` row asserting the child is in the caller's workspace, and that field is on the public contract (`sessions/runtime/view.rs:307`, `sdk/src/generated/openapi.ts:3408`). No consumer branches on it today (grep: serialisation only), which is why this is M and not H — but the rows are durable and cheap to get right now, expensive to reinterpret later. Either add a variant to `SessionLinkWorkspaceRelation` (`links/model.rs:63-67`) and derive it from `request.is_cross_workspace(caller)`, or, at minimum, fix the comment and record the deliberate flattening.

### M2 — `modeId` is the one launch selection never checked against the target

`resolve_launch_against_target_workspace` (`spawn_ops.rs:233-273`) validates `agent_kind` and `model_id` against the target catalog and says nothing about `mode_id`, which `inherit_from` accepts verbatim from `initialConfig.modeId` (`spawn_ops.rs:175`). `ResolvedLaunchModelOption.modes` is documented as *"The **permission**/agent modes the model supports (`controls.mode.values`)"* (`agents/readiness/launch_options.rs:38-40`) — i.e. exactly the kind of thing a workspace authorises. The PR's own stated principle, restated in `subagents.md` ("the target's universe decides, never the caller's"), is applied to two of the three selections.

Unvalidated `modeId` pre-dates this PR (same on `spawn_subagent`), so this is not a regression — but this PR is the one that makes the selection cross a workspace boundary, and the composition routine it adds is the natural place to close it. If it's deliberately deferred, say so in `subagents.md` where the principle is asserted.

### M3 — `mode=local` is non-idempotent and the agent cannot undo it

`workspace_ops.rs:454-458` calls `create_workspace_with_origin_and_creator_context`, which is `resolve_or_create_workspace(allow_existing = false)`. For `WorkspaceKind::Local` that path does **not** bail on an existing active workspace — the `anyhow::bail!` at `workspaces/runtime/identity.rs:97-99` is guarded by `if workspace_kind == WorkspaceKind::Worktree` — it falls through and inserts another row over the same directory (`:113-125`). So an agent calling `spawn_workspace(mode: "local")` twice gets two indistinguishable workspaces on one checkout, `N` times gets `N`, and ruling 11 means it cannot clean any of them up. The human `POST /v1/workspaces` behaves the same way, but a person sees the duplicate in the picker immediately; an agent in a retry loop does not. Prefer `resolve_from_path_with_origin_and_creator_context` here (returns the existing one), or refuse when an active local workspace already exists for that path.

### M4 — blocking git in the async dispatch path

`get_workspace_options` calls `resolve_git_context(&root.path)` inline for every present repo root (`workspace_ops.rs:216-220`). That function is fully synchronous and shells out 4–5 times per root (`workspaces/resolver.rs:22-79`: `canonicalize`, `rev-parse --show-toplevel`, `--git-common-dir`, `--git-dir`, `--abbrev-ref HEAD`, `git config`). `spawn_local_workspace` (`:454`) is the same story. Both run on the axum request task with no `spawn_blocking`, unlike both human routes (`api/http/workspaces.rs:48` wraps in `run_blocking`; `worktree_runtime.rs:93` uses `spawn_blocking`). With a handful of configured roots this is tens of subprocess spawns blocking a tokio worker. Wrap both in `spawn_blocking`.

### M5 — the two new tool bodies have no behavioural test

`workspace_ops.rs`'s test module (`:502-804`) covers `assert_local_placement`, `agent_creator_context`, `WorkspaceSpawnMode::parse`, `validate_branch_name`, `base_branch_for` and `repo_root_name` — all pure helpers. Neither `get_workspace_options` (`:199`) nor `spawn_workspace` (`:281`) is ever executed by a test. What stands in for coverage is three source-text greps: `both_creation_paths_actually_stamp_the_context_they_build` (`:695`), `spawn_workspace_has_nothing_to_lease_and_says_so` (`api/session_admission_tests.rs:774`) and the ruling-11 ratchet (`tools.rs:1101`). Those are good ratchets and they do bind (I checked the squashed-body needles match the emitted formatting), but they prove ordering and absence, not behaviour: nothing asserts that `mode=worktree` reaches `create_worktree` with `NewBranch`/`SuffixPathAndBranch`, that `mode=local` ignores `branchName`, that the `repoRootId` default resolves to the caller's root, or that `get_workspace_options` marks an absent root unavailable *with* the reason. `peer_ops.rs`'s tests already build a real `Db` + `WorkspaceAccessGate` fixture (`peer_ops.rs:871`), so at least one end-to-end pass through each tool is within reach.

### M6 — stale lock-story doc on the gate this PR re-aimed

`subagents/service.rs:144-147` still reads: *"plus the workspace the new session lands in being mutable, **which for this tool is the caller's own, the one whose write lease the route already took**."* Both halves are now false — the workspace can be any workspace, and the route takes no lease for `spawn_agent`. The file is untouched by this PR (clean rebase), which is exactly how this kind of comment rots. This program ratchets its lock stories in comments; this one should be corrected in the same PR that invalidates it.

### M7 — undisclosed widening: any repo on the machine, not "the same git repo"

ADR §1 req 1.7 is *"Spawning new workspaces (if running locally) **with the same git repo**"*; §3.4 lists `repoRootId` *"(defaults to caller's own root)"*. The implementation takes the permissive reading — `get_workspace_options` enumerates every configured root (`workspace_ops.rs:210-235`) and `spawn_workspace` accepts any of them (`:293-304`) — so an agent working on repo A can create a workspace on repo B and staff it via `spawn_agent`. That may well be intended (it is what §3.4's argument implies, and it is what makes H1 reachable), but it is a real capability widening and the PR body discloses two other ADR contradictions while passing over this one. Worth an explicit founder ruling recorded in `specs/anyharness/workspaces.md#agent-requested-creation`, either way.

---

### L1 — a failed `spawn_workspace` renders as "Workspace spawned"

`apps/packages/product-client/src/domain/chats/subagents/subagent-tool-presentation.ts:133-138` returns the settled verb before the `executionState === "failed"` branch at `:139`, and the new test asserts it: `subagent-tool-presentation.test.ts:167-171` expects `failed → "Workspace spawned"`. The comment says the goal was to avoid the "Subagent launch failed" fallback — fair, but reporting success for a failure is the worse of the two. Every named branch in that function has the same shape, so this is a pre-existing pattern rather than a regression; still, the test now pins the wrong copy in place. `"Workspace spawn failed"` / `"Workspace options unavailable"` cost two lines.

### L2 — "every argument is optional" is not true

`tools.rs:1088-1112` asserts `spawn_workspace` advertises no `required` array, with the comment *"Every argument is optional: the default is a worktree off the caller's own repo."* But `mode` defaults to `worktree` (`workspace_ops.rs:73`) and `worktree` requires `branchName` (`:346-352`), so a genuinely argument-free call **always** fails. The schema is defensible (the requirement is conditional and JSON Schema can't express it here without `oneOf`), and `branchName`'s description does say "Required for mode=worktree" — but the test comment states the opposite of the behaviour.

### L3 — the two filesystem facts in one gate fail in opposite directions

`assert_local_placement` takes `caller_checkout_missing` from `WorkspaceRecord::checkout_directory_missing`, which deliberately fails **open** on `PermissionDenied`/transient I/O (`workspaces/model.rs:57-64`), and `repo_root_checkout_present` from `Path::is_dir()` (`workspace_ops.rs:188-190`), which fails **closed** on the same errors. One gate, two opposite conservatisms. Not a defect today; worth a line in the doc comment at `:110-121` so the next reader does not assume symmetry.

### L4 — `label` is a fourth argument, unbounded

ADR §3.4 gives `spawn_workspace` three args and §3.1 rejects widening the creation surface. `label` (`tools.rs:1127-1128`) is a reasonable fourth — it feeds the creator stamp §3.4 does require — and the doc comment discloses it. It is length-unbounded and agent-authored, and it lands in `WorkspaceCreatorContext::Agent` (`workspace_ops.rs:160-171`) where §4 renders it to a human. A trim-and-cap alongside the existing trim would cost one line.

### L5 — branch-name validation is a denylist

`validate_branch_name` (`:353-363`) rejects leading `-`, leading/trailing `/`, `..` and whitespace. It still admits `HEAD`, `refs/heads/x`, `~ ^ : ? * [ \` and control characters. Git rejects those itself, so the only cost is a raw git error instead of a shaped one — and the security-relevant cases (`-`-prefixed argument injection, path traversal) are both closed. Noted for completeness, not a blocker.

### L6 — `assert_spawnable_workspace` does not check the target checkout exists

`calls.rs:504-521` checks existence and `WorkspaceSurface::Standard`; the access gate (via the lease) checks retired/read-only. Nothing checks the target's checkout is on disk, so spawning into a workspace whose directory was deleted creates the session, fails at `start_persisted_session`, and unwinds through `unwind_after_failed_dispatch`. Correct, just a noisier path than a refusal — and `WorkspaceRecord::checkout_directory_missing()` is right there, already used by `workspace_ops.rs`.

### L7 — the empty-catalog pass-through, and `appliedInitialConfig`

`spawn_ops.rs:241-243` returns `Ok(())` when the target's catalog resolves to no agents, so any harness/model is admitted there. I do not think it is a hole: the catalog is a readiness/curation view, auth contexts are resolved per-workspace at launch, and a genuinely unlaunchable harness fails at start with compensating cleanup. But the refusal surface is asymmetric (empty ⇒ admit everything) and it is the one arm reachable by naming a workspace id that does not resolve. Relatedly, `applied_initial_config` (`calls.rs:475-488`) now reads off the created session — the right call — but nothing tests it, even though `subagents.md` asserts *"a key that was not honoured is visibly absent there"* and this PR adds the fallback path (inherited model → target default) that makes the request and the row diverge.

---

### On the two disclosed contradictions

Both resolutions are sound. **Local-only:** expressing the gate as `has_local_checkout()` + two filesystem facts is the right shape — the first arm is dead today and the test at `workspace_ops.rs:642-654` says so honestly rather than pretending otherwise, and it is wired to the predicate ADR §2 names rather than to a kind list a future kind would slip past. **"Promoted / top level only" via `can_spawn_agent`:** correct, and better than a bespoke check — `resolve_context` recomputes it per request from `validate_caller_can_spawn_agent` (`context.rs:71-75`), so `spawn_workspace`'s in-body `ctx.can_spawn_agent` check (`calls.rs:534-541`) is a fresh service verdict, not a cached dispatch decision. One asymmetry worth a note: `spawn_agent` re-calls the service in its body while `spawn_workspace` reads the ctx flag, so `sessions.md`'s "the tool bodies read those gates" is true in two different senses.

pablonyx:
Review findings addressed in `edc09d0f4`.

**H1 — cross-repo setup script leak.** Fixed. `setup_command_for_new_worktree` now
picks per repo root: the caller's own root reuses the caller's last command, a
different root takes the latest command from that root's own active workspaces,
and a root that has never run one gets no setup command. Three tests assert the
exact `CreateWorktreeWorkflowInput` handed to the worktree runtime.

**H2 — stale tool copy.** Fixed. `spawn_agent`'s description and its
`harnessId`/`modelId`/`modeId`/`workspaceId` arg copy now say the peer lands in
`workspaceId`, point at `get_workspace_options`/`spawn_workspace` for finding one,
and state that harness, model and mode are checked against the TARGET workspace's
catalog. A ratchet test pins it.

**M1 — link workspace relation.** Fixed. `SessionLinkWorkspaceRelation` gains
`CrossWorkspace` and a `between()` constructor; `link_owned_agent` takes both
workspace ids and writes the true relation. The false comment is gone. No DB CHECK
constraint gates the column and the public contract serializes the string, so no
migration or SDK regen is needed.

**M2 — modeId unvalidated.** Fixed. `modeId` follows the model policy against the
chosen model's `controls.mode.values` in the target catalog: a caller-NAMED mode the
target does not offer refuses, a merely INHERITED one is replaced by the target
harness's unattended default. A model with no mode control passes through.

**M3 — mode=local not idempotent.** Fixed on the agent path only. An active local
workspace for the requested root is returned with `reused: true` instead of a
duplicate. Ruling 11 means agents cannot retire what a retry created. The human
route and the shared helper are untouched.

**M4 — blocking git on the request task.** Fixed. `get_workspace_options` and
`spawn_local_workspace` do their subprocess work under `spawn_blocking`, matching
the human routes.

**M5 — no executed-body coverage.** Fixed. A `WorktreeSpawner` seam (the real
runtime needs the whole retention/preflight stack to construct) plus temp-dir git
fixtures now exercise the local-only gate, the `mode=local` happy path with its
creator stamp and its reuse, the setup-command choice, and `get_workspace_options`
against a seeded store. The source-text ratchets are kept.

**M6 — stale lock story.** Fixed at `subagents/service.rs`.

**M7 — repo-root widening vs ADR §1 req 1.7.** This is a real contradiction inside
the ADR, resolved in favour of §3.4's `get_workspace_options` contract: creation may
target any configured repo root present on this machine, local placement only. §1
req 1.7 is amended with a note saying so, and the widening is now stated in
`specs/anyharness/workspaces.md` and `subagents.md`. **Flagging this for a founder
call** — it is a deliberate scope widening of what agents may create, not a
clarification.

**L1** failed `spawn_workspace`/`get_workspace_options` receipts now say they failed.
**L2** wrong test comment corrected, and the fact it claimed is now asserted.
**L3** the local-only gate fails closed: an unreadable path refuses and names the check
(`CheckoutPresence` replaces the booleans).
**L4** `label` is bounded to 200 characters, documented in the arg description.
**L5** deferred — the UX-only tool denylist is a separate change.
**L6** a target workspace whose checkout is gone is refused before anything is created.
**L7** `appliedInitialConfig` is asserted end to end on the cross-workspace path, and
the empty-catalog pass-through asymmetry is documented where target-catalog
composition is described.

Every new guard was negative-controlled: reverted, its test watched fail, restored.
H1 (2 tests), H2, M1, M2 (2 tests), M3, L1, L3, L4 and L6 each failed as expected.

Validation: `cargo test -p anyharness-lib --lib -- agent_ops workspaces ownership subagents session_admission` = 300 passed / 0 failed. Full `--lib` = 1751 passed with one pre-existing failure (`probe_materialization::sweep_tests::the_sweep_removes_only_abandoned_and_old_scratch_roots`), which this branch does not touch. `subagent-tool-presentation.test.ts` = 17 passed. `scripts/check_docs.py` passed. Specs are in the same commit as the behavior.

pablonyx:

Issue #1734, opened 2026-08-08
Agents pane: global delegating-sessions read model + human wake-on-reply flag


pablonyx:
Two server gaps PR9 (agent-ops client UX) renders around rather than fakes:

1. **Global overview read model.** ADR §4's pane Level 1 lists every session that is delegating. The client only has per-session `getSubagents`, so the overview is built from whatever delegated work the client has already read — effectively one cluster. A server read model (sessions with open subagent/owned links + live summary) is needed for a true overview.

2. **Human wake-on-reply.** The pane's agent-detail composer has a 'Wake me on reply' toggle per §4, shipped disabled: `wakeOnReply` exists only on the agents' `send_agent_message` MCP tool; the human prompt route (`POST /v1/sessions/{id}/prompt`) has no wake flag. Needs a route-level flag arming the same reply-consumable wake row.

Both are documented as honest boundaries in the PR9 specs (delegated-work.md / subagents.md).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

pablonyx:
Third gap, found in the PR #1735 review:

3. **Human soft-close route.** ADR §4 promises the close confirm "It's mid-turn — it will finish the current step, then stop", which is §6 step 5.3's soft close. That mechanism exists, but it is agent-attributed by design: the end-requested window is stamped `closed_by_session_id` on the link row and is only reachable from the agent-ops `close_agent` tool (`agent_ops/calls.rs`). The human pane wires close to `POST /v1/sessions/{id}/close` → `close_session_tree`, which calls `handle.close()` + `mark_closed` immediately — a hard kill mid-step. There is no human soft-close route in `router.rs`.

PR #1735 does not fake it: the confirm copy now reads "It's mid-turn — closing stops it now. The transcript stays readable under Closed.", and ADR §4's closing bullet carries the deviation.

Consequence worth carrying with this item: because the plain close also closes the inbound links, a human-closed agent leaves `getSubagents` entirely rather than passing through the pane's Closed section. Closed attribution stays readable only in the close-requested window and in the transcript receipt. A human soft-close route would fix both — it would need a closer identity for a human (or a null-closer convention) so the attribution line has a subject.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1735, opened 2026-08-08
feat(product): the agent-operations client — seal glyphs, chips, and the agents pane

pablonyx:
Part 9 of the agent-operations stack (ADR §6 step 9 — the client §4 UX). Stacked on #1733.

## What this does

The whole §4 surface, in two layers.

**Plumbing** (data first, no invented visuals):
- Session-scoped **agent-wake rendering**: `agentWake` prompt provenance in the transcript, pending queue, and copy pipeline.
- **closeRequested / closeReason** threaded through SDK types and every delegated-agent row: an end-requested agent shows "Closing" (with reason) until the close lands, outranking its own session status.
- **Peer-spawn stream visibility** closed: `spawn_agent` / `promote_subagent` / `send_agent_message` / `configure_agent` invalidate the subagent read model, `spawn_agent` / `spawn_workspace` also invalidate workspace collections. Deliberately no fanout mount — a peer is not a subagent.
- **Human promote route**: `POST /v1/sessions/{id}/subagents/{child}/promote` (same one-write idempotent stamp the MCP tool performs; takes only the shared `SubagentWrite` lease — promotion touches no session, so admitting a session mutation would refuse promotions a workflow cannot conflict with, documented in the handler). SDK `promoteSubagent` + mutation hook.
- **Widened link summaries**: `ownedAgents` as a separate list (a peer is structurally never in a fanout), `promotedAt` on children and on the parent summary, so all three ownership states render truthfully.

**Visuals** (canvas-grounded, locked rulings from §4):
- **Identity**: every agent glyph is now the **solid seal** (locked style 1e) — a seeded solid shape with a mask-punched notch, one component at every size (receipt/tab/chip/pane).
- **Agent chip primitive**: 28px pill, 16px seal, tinted name, 288px truncation, mono short-id suffix for by-id addressing, dimmed-but-clickable for closed. The chip is the clickable agent, everywhere.
- **Spawn receipts**: a spawn run is a horizontal chip run with one trailing quiet verb ("started working" → "finished"); settled runs keep their chips forever. The old collapsible "N subagents finished" block is gone.
- **Agent messages**: chip + quiet verb ("messaged" left / "replied" right), literal bodies only in the hover card, never bubbles. Cross-workspace agents carry "— in {workspace}".
- **Pending updates**: one quiet row — overlapping glyphs, count, "delivered next turn". The human's own queued prompts keep their full editable rows.
- **Agents pane**: right-panel tool, overview → cluster (Working/Idle/Done/Closed) → detail (glyph, title, status, copyable short id, parent prompt, latest completion; Open as tab · Configure · Promote · Close; message composer). Entry: the composer's agents cap opens that session's cluster; the panel tab opens the overview; never follows tab focus.
- **Promotion/closing**: one confirm each with the ADR copy verbatim; "Promoted · top-level session" badge; idle/done close instantly; a human close leaves no transcript trace; an agent's `close_agent` writes the dimmed chip receipt with the close reason.
- **Workspace receipts**: "Created workspace billing-hotfix-dispatch — proliferate · worktree from main · Open".

## Honest boundaries (spec'd, not faked)

- The pane's Level-1 overview is built from delegated work the client has read; there is no global "which sessions are delegating" server read model, so it is effectively single-cluster today. Follow-up filed: #1734.
- "Wake me on reply" on the pane composer ships disabled with a tooltip — the human prompt route has no wake flag server-side. Same follow-up issue.
- The Closed pane section is thin by design: `getSubagents` returns open links only; post-close attribution lives in the end-requested window and the transcript receipt (documented in `subagents.md` — no closed-agents read model invented).
- Agent detail shows no "current Tool" line — the read model has no tool cursor; nothing is faked.

## Testing

- product-client: **801 files / 5402 tests** green; `tsc` (main + domain) clean.
- SDK 143, sdk-react 34.
- Full `anyharness-lib --lib` on this head's base: 1750 passed + regenerated-openapi drift test green after `cargo run -- print-openapi` + `openapi-typescript` (the only real red earlier); single remaining failure is the known pre-existing `sweep_tests` scratch-roots red.
- 13 negative controls across plumbing + visuals (each guard reverted, watched fail, restored) — listed in the review thread.
- Appearance-scaling guard green on every commit; glyph sizes use semantic `icon-*` tiers (the guard rejects raw sizes — canvas px values map to the nearest tier, noted in-code).

## Known deferrals

Stack-introduced lint reds are swept at the top of the stack with founder-flagged allowlist entries, as disclosed on #1731 and #1733.


pablonyx (commented):
## Adversarial review — PR #1735 (agent-operations client, §4 UX)

**Frozen review pair** — every read below is pinned to these SHAs:
- BASE `edc09d0f4` (head of `agent-ops/7-workspace-spawn`, #1733)
- HEAD `e02ae820d` (head of `agent-ops/9-client-ux`)

References are `path:line` against `e02ae820d` unless stated. Read-only review; no branches touched, no cargo run.

### What I verified and found correct

- **Glyph geometry is exact.** `DelegatedAgentIdenticon.tsx:13-35` reproduces canvas `glyphs.js` style `1e` bit-for-bit: `mixHash`≡`mix` and `stableIndex`≡`hash` (`identity.ts:186-199`), `shape = mixed & 3`, notch ring `4.1`, notch `r 2.3`, angle `((mixed>>>2)%8)*45°`. The only divergence is deliberate and better: a real `<mask>` punch instead of the canvas's `fill: var(--color-background)` dot, and the test negative-controls it (`DelegatedAgentIdenticon.test.tsx:23-25` asserts `not.toContain("var(--color-background)")`).
- **ADR-quoted copy is verbatim**: promote confirm, close confirm, `Promoted · top-level session`, `Message this agent — delivered on its next turn` (`agents-pane-model.ts:221-243`), and `Closed by {who} · {reason}` shape.
- **Hard UI rules hold.** No colored left-border ownership treatments anywhere in the diff; no raw hex (the `fill="black"/"white"` in the mask are luminance keys, not palette); glyph sizing goes through `icon-*` tiers, and `icon-paired` = `1.230769em` × `text-ui` 13px = the locked 16px.
- **Read-model honesty traces out.** Promoted children leave the tree in one place (`workspace-header-subagent-hierarchy.ts:74-104`) and it is behaviorally pinned, not markup-pinned (`workspace-header-subagent-hierarchy.test.ts:35-69`). `ownedAgents` structurally cannot enter the fanout — the hierarchy builder only ever reads `data.children`. `closeRequested` outranks status at the tab (`workspace-header-tabs-view-model-derivation.ts:102`), the menu (`ChatTabsMenu.tsx:256-264`) and the pane (`agents-pane-model.ts:88-99`).
- **Rust drift test exists and is a real equality check** (`workflow_workspaces_tests.rs:578-586`, `assert_eq!(runtime, checked_in)`), and the checked-in document carries the new path + `PromoteSubagentRequest/Response` + `OwnedAgentSummary`, with `ownedAgents` correctly OPTIONAL (matching `#[serde(default)]`).
- **Promote error mapping matches the neighbours**: `SUBAGENT_NOT_OWNED`/`SUBAGENT_CLOSED` as 409, target problems as 400 — same shape as `map_subagent_error` (`subagents.rs:306-340`). Deleting `lib/domain/delegated-work/identicon.ts` leaves no consumers (grep clean).
- **The promote fence itself is defensible.** `promote_subagent` really is in `MUTATING_TOOL_NAMES` (`agent_ops/tools.rs:48-55`), so the MCP path takes the same caller-workspace `SubagentWrite` lease and no permit; the write is one indexed UPDATE on a link row and touches no session actor. Taking a lease with no permit adds no edge to the canonical `run gate → permit → lease` order, so no ABBA risk. I could not construct a case where a workflow-controlled parent's *execution* is affected: promotion changes cascade membership, the fanout count, and the child's tool list at next launch — none of which is an in-flight session mutation. See M6 for what is missing around it.

---

## H1 — The pending-prompt reorder now moves the wrong row

`PendingPromptList.tsx:47-70` filters agent-sourced entries out of the rendered list and hands `PendingPromptRow` / `useVerticalReorder` an index into `humanEntries`. But `onReorder` is `handleReorder` in `use-pending-prompt-queue.ts:143-166`, which splices into `rows` — the **full** queue, agent entries included:

```ts
// use-pending-prompt-queue.ts:157-166
const expectedSeqs = rows.filter((row) => row.seq > 0).map((row) => row.seq);
const reorderedRows = [...rows];
const [moved] = reorderedRows.splice(fromIndex, 1);
```

Before this PR the two lists were the same array, so the indexes agreed. Now, with `rows = [agentWake(seq1), human(seq2), human(seq3)]`, dragging `human(seq3)` (human-index 1) to human-index 0 calls `onReorder(1, 0)`, which in `rows` moves `human(seq2)` in front of the **agent wake** — a different reorder than the user performed, and one that also reorders an agent-queued update the ADR says the human may not touch. Any queue containing at least one agent-sourced entry ahead of a human one is affected.

Fix shape: map human-list indexes back to `queue.rows` indexes before calling `onReorder` (or have the hook accept row keys). No test covers this — `pending-prompt-queue.test.ts` and `pending-agent-updates.test.ts` both stop at the grouping.

## H2 — The close confirm promises a soft close; the human close is a hard kill

`ConnectedAgentsPane.tsx:41,76-88` wires the pane's close to `useCloseSessionMutation` → `POST /v1/sessions/{id}/close` → `close_live_session` → `close_session_tree` (`domains/sessions/runtime/lifecycle.rs:38-52,99-120`), which calls `handle.close()` and `mark_closed` immediately. There is no end-requested window on that path: `find_pending_close_request` / `ownership::hooks` are only reached from the agent-ops `close_agent` tool (`agent_ops/calls.rs:1137-1200`), and there is no human soft-close route in `router.rs` (only `/subagents/{child}/wake` and the new `/promote`).

So the pane shows, verbatim:

> It's mid-turn — it will finish the current step, then stop. The transcript stays readable under Closed.

…and then kills the actor mid-step. ADR §6 step 5.3 is explicit that this is backwards — *"Soft close: closing a working agent enters 'end requested' — the in-flight step finishes, then the tree closes (§4's confirm copy depends on this)"*. The new spec text repeats the promise (`specs/codebase/systems/product/agents/delegated-work.md`, "Closing"), and the PR's "Honest boundaries" does not disclose it. Either route the human close through the ownership close path (attribution + soft close), or change the confirm copy and disclose the deviation.

Secondary consequence: because the plain close also closes inbound links, a human-closed agent vanishes from `getSubagents` entirely rather than passing through the Closed section — which is the real reason the Closed section is empty, not just the read-model thinness the PR body cites.

## H3 — Inbound agent messages are still bubbles; "replied" does not exist

The PR body claims *"Agent messages: chip + quiet verb ('messaged' left / 'replied' right), literal bodies only in the hover card, never bubbles."* The left/outbound half is real (`SubagentToolActionRow.tsx`). The right/inbound half is not: `TranscriptItemBlock.tsx:167-186` still renders `agentSession` provenance — which *is* the reply, under reply-is-the-wake — as a full `UserMessage` with the literal body plus `UserMessageProvenanceChrome`. Grepping the whole client, the string `"replied"` appears only in two doc comments (`AgentChip.tsx:113`, `subagent-tool-presentation.ts:41`); nothing emits it.

This contradicts ADR §4 twice over: *"Message content never gets its own UI"* / *"never bubbles"*, and *"the reply … marked by the same chip with 'replied'; the answer itself is just assistant prose."*

Compounding it: `UserMessageProvenanceChrome.tsx:15-19` hard-codes `Sent by parent` and falls back to `Parent agent`. `PromptProvenance{type:"agentSession"}` carries an **optional** `sessionLinkId` (`sdk/src/generated/openapi.ts:3888-3893`) precisely so a peer can be told from a parent, and the client ignores it. A peer's message therefore renders as "Sent by parent" — a false claim about ownership, in the one PR whose stated job is that these three states render truthfully.

---

## M1 — "Configure agent…" is missing from the agent detail header

ADR §4 locks the detail actions as *"Open as tab · Configure agent… · Promote · Close"*, and the PR body lists all four. `AgentsPaneAgentDetail.tsx:65-116` renders Open as tab, the copy-id button, Promote and Close — no Configure. `git diff … | grep Configure` finds only the receipt copy strings. The new spec (`delegated-work.md`, "Level 3") quietly writes the three-action list instead, so the deviation is baked into documentation without being disclosed as one. `configure_agent` exists server-side (PR4), so this is a gap, not a blocked dependency — either wire it or list it under Honest boundaries.

## M2 — The "panel tab opens the overview" entry point is not implemented

ADR §4: *"Entry points: the 'N working' cap … opens that session's cluster; the panel-tab button opens the overview."* `agents-pane-store.ts:34` defines `openOverview`, and nothing calls it — grep across `apps/packages/product-client/src` returns only the store's own declaration and interface line. `ToolHeaderButton`/`RightPanelContent` just switch `activeTool` to `"agents"`, so the pane re-appears wherever it was last pointed (e.g. deep inside another session's agent detail). The composer spec asserts the behaviour anyway: *"The other entry point is the panel's own `Agents` tab, which opens the overview."*

## M3 — "Closed by X · reason" can never render in the app

`agentsPaneCloseAttribution` (`agents-pane-model.ts:157-176`) implements the ADR's *"Pane shows who closed it"* line and is unit-tested (`agents-pane-model.test.ts:172-184`). `AgentsPane.tsx:21,87` takes `closeAttributionFor`, and `ConnectedAgentsPane.tsx:104-118` never passes it. The only place the attribution is exercised is `AgentsPane.test.tsx:193`, which supplies a stub — the test passes on a code path the product never reaches. Also, `closedBySessionId` is a session id and nothing in the diff resolves it to a title, so even wired up the "by whom" half has no resolver.

## M4 — The pane's cluster mixes two sessions' read models under one title

`ConnectedAgentsPane.tsx:47-62`:

```ts
const clusterSessionId = subagents?.parent?.parentSessionId ?? activeSessionId ?? null;
const agents = [...subagents.rows, ...subagents.ownedAgents].map(toAgentsPaneAgent);
```

`rows` is deliberately the **parent's** fanout when a child is in view (`use-subagent-composer-strip.ts:113-121` prefers `parentSubagentsQuery.data?.children`), but `ownedAgents` is deliberately read off the **active** session only (`use-subagent-composer-strip.ts:129-131`, with a comment saying exactly that: *"a child does not inherit its parent's peers, because it does not own them"*). Merging them and labelling the result with the parent's title (`subagents.parent?.label`) puts the child's own peers inside the parent's cluster — the same "an agent shown under an owner that does not own it" error the rest of the PR is careful to avoid.

## M5 — "— in {workspace}" renders even when the agent landed in the current workspace

`subagent-tool-presentation.ts:250-253` sets `workspaceId` unconditionally from `rawOutput.workspaceId ?? rawInput.workspaceId`, and `SubagentToolActionRow.tsx:97-100` renders the suffix whenever it is present and resolvable. ADR §4 scopes the suffix to the cross-workspace case: *"An agent spawned **elsewhere** carries its workspace in the receipt."* A `spawn_agent` into the caller's own workspace (the common case) therefore gets a redundant `— in {this workspace}` on every receipt. Compare with the current session's workspace before rendering.

## M6 — The Rust half ships with no tests, and the fence exclusion has no ratchet

The whole server diff (`contract/v1/sessions.rs`, `api/http/subagents.rs`, `subagents/service.rs`, `router.rs`, `openapi.rs`) adds **zero** test files. Specifically missing:

1. No test for `POST …/promote` at all — not the happy path, not idempotency (`alreadyPromoted`), not the 404/409 mapping.
2. No test for the `owned_agents` widening in `subagent_context` (`subagents/service.rs:429-455`) — including the invariant the whole PR rests on, that an `owned_agent` row never lands in `children`.
3. No admission ratchet for the deliberate permit exclusion. `session_admission_tests.rs:132-201` enumerates every fenced route and includes the *sibling* `POST …/subagents/{child}/wake`; the new promote route sits beside it and is deliberately absent, justified only by a doc comment (`api/http/subagents.rs:101-116`). By contrast `spawn_agent` and `spawn_workspace` each got an explicit ratchet test asserting their exclusion **with the rationale in the assertion message** (`session_admission_tests.rs:747-810`). Given the frozen contract's "no exception without an enforced trail" posture, this exclusion deserves the same: a test asserting the promote handler body contains no `admit_session_mutation` and holds only the shared lease, with the reason in the message. Today a well-meaning refactor that "fixes the inconsistency" by adding the permit would go green.

Also worth pinning: `assert_workspace_mutable` is called **after** `acquire_shared` here, matching `schedule_subagent_wake` — good, but only convention keeps it that way.

## M7 — A subagent's result summary is no longer readable anywhere short of opening its session

The deleted `SubagentFinishedRow` (BASE `SubagentCreationGroupBlock.tsx:80-170`) was the only surface that showed `rawOutput.summary` — the structured result the parent agent received. Nothing in the new chip run replaces it, and the pane's detail carries only `latestCompletionLabel`, which `use-subagent-composer-strip.ts:298-315` renders as `"Completed turn"` / `"Failed turn"` — an outcome, not a summary. ADR §4 rules out expanding the *message* body in the transcript, but says nothing about deleting the completion summary; ADR §4's detail view even asks for a "latest Agent message" line. Confirm this loss is intended, and if so list it in Honest boundaries.

---

## L findings

- **L1 — promote route problem-detail nits** (`api/http/subagents.rs:176-201`). (a) 400 is reachable (`SelfTarget` via `POST /sessions/X/subagents/X/promote`; `TargetRequired` via a whitespace child id) but the `#[utoipa::path]` `responses(...)` documents only 200/404/409 — the sibling wake route documents 400. (b) `ConflictingTarget` is folded into `SUBAGENT_TARGET_REQUIRED` with the message `"childSessionId is required."`, which is wrong for that variant; `map_subagent_error:327` uses a distinct `SUBAGENT_TARGET_CONFLICT`. (c) `SelfTarget` emits `INVALID_TARGET`, outside the `SUBAGENT_*` code family every neighbour uses.
- **L2 — path census not extended.** `openapi_tests.rs:37-38` lists `/v1/sessions/{session_id}/subagents/{child_session_id}/wake` but the new `/promote` sibling was not added.
- **L3 — pending-updates hover copy deviates from ADR.** ADR §4 quotes `"N queued · click to open"`; `pending-agent-updates.ts:74-76` emits `"{title} · N queued updates — click to open"`. Documented in the new composer spec, not listed as a deviation in the PR body.
- **L4 — `h-7` instead of the `h-control` token utility.** `AgentChip.tsx:31` (and the detail-header buttons, `AgentsPaneAgentDetail.tsx:72,84,97,108`) hard-code the 28px pill. `product.css:35-41` says explicitly that control height is emitted from `--height-control` as `h-control` and that re-declaring it elsewhere is a second authority. Numerically identical today; vocabulary drift the guard cannot regex.
- **L5 — `leading-8` on scalable text.** `AgentChip`'s three consumer rows (`SubagentCreationGroupBlock.tsx:53`, `SubagentToolActionRow.tsx:74`, `SubagentWakeBadge.tsx:46`) pin a 32px line box against `text-message`, which scales with the appearance preference. Not caught by the guard (it bans only `leading-[…]`), but it is the same class of defect.
- **L6 — fixed `size-5` glyph wells.** `PendingAgentUpdatesRow.tsx:41-42` and `AgentsPaneOverview.tsx:48-49` wrap an `icon-indicator` (`1em`) glyph in a fixed 20px well; at the top of the appearance ramp the glyph approaches the well.
- **L7 — dead wiring.** `AgentsPaneClusterSections.selectedLinkId` (never passed, so row selection never highlights), `AgentsPaneAgentDetail.canSend`, `AgentsPaneHeader.actions`, `agentsPaneStore.resetForTests`, and `SubagentWakeBadge`'s `originKind`/`parentTitle` (declared in the interface, still passed by two call sites, no longer destructured or used).
- **L8 — ADR's second workspace receipt has no producer.** §4 locks `"Updated workspace … — run script → pnpm i && pnpm test:webhooks"`; `workspace-ops-presentation.ts:23-26` only handles `spawn_workspace`, and there is no update tool in `agent_ops/tools.rs`. Fine as a server gap — but it belongs in Honest boundaries next to the other three.
- **L9 — entry point is the popover, not the cap.** ADR §4 says the goal-bar `"2 working"` cap opens the cluster; `AgentsPopoverSubagentSection.tsx:96-108` puts an `Open agents pane` button inside the popover instead. Documented in the composer spec; not listed as a deviation.

---

### Routing note

H1 is a plain correctness regression and independent of the ADR debate. H2 and H3 are ADR-fidelity failures that the PR body actively claims as delivered, so they need either a fix or an explicit, founder-visible deviation entry. M6 is the one I would not ship without: the Rust half of a 9-PR stack landing with zero tests and an unratcheted fence exclusion is the kind of hole the rest of this stack has been careful to close.

pablonyx:
Review fixes pushed as two commits: `524d2c297` (client + client specs) and `1abeffb63` (Rust + SDK regen + agent-features spec).

**H2 is the one that changes a promise, so read this first.** The ADR writes the mid-turn close confirm as "it will finish the current step, then stop". That is the soft close from §6 step 5.3, and the soft close stamps `closed_by_session_id` — it names the closing *session*, so it is agent-attributed by design and no human route reaches it. The human button calls `POST /v1/sessions/{id}/close`, which stops the tree immediately. Rather than build a server route tonight or keep copy that lies, the confirm now reads "It's mid-turn — closing stops it now. The transcript stays readable under Closed." The tone is unchanged: nothing destructive-styled, close is still routine. `adr-agent-ops.md` §4 carries an amendment, `delegated-work.md` states the honest behaviour and its consequence (a landed close leaves the OPEN-links read model, so `Closed` holds only close-requested-not-landed agents), and #1734 gained a third item for the human soft-close route.

H1: reorder spliced by position into the full row list while the drag handles indexed the filtered human rows, so an agent update above the drag moved the wrong row — and could move an agent-sourced entry the human may not touch. Reorder is now key-based and rebuilds only the human slots. Negative control: restoring the positional splice yields `[3,1,2]` instead of `[1,3,2]`.

H3: inbound agent messages rendered as `UserMessage` bubbles, which gave message content its own UI and attributed an agent's words to the human. They are now receipts like everything else — verb-then-chip on the right, literal body in the hover card only, verb `replied` when this session already messaged that agent (derived from the transcript, since `agentSession` provenance carries no reply marker) else `messaged`. Parent vs peer is read off `sessionLinkId`, so a peer can no longer claim "Sent by parent". `UserMessageProvenanceChrome` is deleted rather than fixed: after the rewrite nothing called it, and leaving a corrected-but-dead component would have been worse than removing it.

M1 adds "Configure agent…" to the detail; it opens the agent's tab, because `SessionConfigControls` is built for the session in view and has no out-of-tab form — no new config UI. M2 the panel's Agents tab now calls `openOverview()`. M3 close attribution comes from the child rows and resolves the closer against sessions already in the directory, falling back to the short id. M4 clusters are partitioned per owner — a child's peers never appear under its parent's title. M5 the workspace suffix renders only when the spawn's workspace differs from the transcript's. M6 the promote route gains `api/sessions_subagents_promote_tests.rs` plus an admission ratchet documenting why it deliberately takes no session permit while `schedule_subagent_wake` does. M7 the subagent completion summary now lives on the spawn chip's hover card.

Two things I could not fully close, both disclosed in the specs rather than faked:

- M7's pane half. `SubagentCompletionSummary` carries no summary text, on the contract or in `domains/sessions/subagents/summary.rs`, so the pane's latest-update line has nothing to show without inventing a read model. The hover card is the only surface that has the text.
- L9. `DelegatedWorkComposerControl` already computes `openCluster(clusterSessionId)`, but the cap itself is the popover's trigger, so the popover's "Open agents pane" is the only cap surface today. Noted in `composer.md`.

L3 uses the ADR's `N queued · click to open` verbatim for the visible hover; the accessible name still prefixes the agent title, because the verbatim string alone leaves the button with no identity for a screen reader. L4 confirmed `--height-control` is 28px before migrating off `h-7`. L6 uses `icon-large` rather than an em value, because the appearance guard bans `size-[...]`. L8 is deliberately not built: §3.4 gives agents no workspace-update tool, so there is no producer for an update receipt — flagged as an ADR nit in `transcript.md`.

Every behavioural fix has a negative control, eleven in total, each reverted after. Validation: 803 vitest files / 5425 tests green, both tsc configs clean, appearance guard and `check_docs.py` pass, and `cargo test -p anyharness-lib --lib` is 1756 passed with the one known pre-existing `sweep_tests` scratch-roots failure. No existing test was weakened — the assertions that changed are the ones whose subject the rulings changed, and each carries a comment saying why.

pablonyx:
Live-QA defect fixed forward @ ba3aaee44: the agents pane rendered see-through over the transcript. Root cause was an invalid `bg-surface` class on the pane root (`bg-surface` is not a token anywhere in the codebase; the right-panel frame is translucent under desktop vibrancy, so each panel paints its own surface). Now `bg-sidebar text-sidebar-foreground`, matching the ScratchPadPanel convention, with a regression test (negative-controlled). #1736 is rebased on top.

pablonyx:

PR #1736, opened 2026-08-08
refactor: delete cowork and reviews (agent-ops part 8)

pablonyx:
Part 8 of the agent-operations stack (ADR §6 step 8) — deliberately last in the queue. Stacked on #1735.

## What

Cowork and reviews stop being product features. This deletes both wholesale — domains, HTTP routes, product MCPs, client surfaces, and specs — rather than leaving dead code behind.

374 files changed, +2,665 / −32,314 (before rebase-seam drift).

## Deleted

**Server** — `domains/reviews/**` (34 files), `domains/cowork/**` (19), `domains/artifacts/**` (6, it existed only to back cowork artifacts); the reviews and cowork routers (8 + 6 routes), 14 OpenAPI paths and 33 schemas, both v1 wire modules, both product MCP servers and their launch-catalog / endpoint-registry entries. `SessionEvent::{SessionLinkTurnCompleted, ReviewRunUpdated}` and `PromptProvenance::ReviewFeedback` go too: the cowork runtime was their only producer (checked against the base commit).

**SDK** — the cowork/reviews REST clients, type modules, sdk-react hooks and query-key factories; generated OpenAPI regenerated from the trimmed router.

**Client** — the cowork sidebar/threads section, thread-launch provider, artifact transcript cards, review assignment/run hooks and header-hierarchy queries. The Home "no project" destination is gone (`HomeNextDestination` deleted, Home launches are repository launches). Delegated work collapses to one `"subagent"` kind; `isReviewAgentChild` is gone; the cowork/review branches drop out of the stream side-effect plan, session-directory relationships, workspace chrome, preferences, shortcuts and telemetry. `ToolCallSummary.completionContent` goes with its only caller.

**Specs** — the two agent-feature definitions and the cowork-artifacts system spec, plus 32 specs that referenced them.

## Kept on purpose

- Shipped migrations are immutable: the `review_*` / `cowork_*` tables stay and the schema snapshot still lists them.
- `SessionLinkRelation::{CoworkCodingSession, ReviewAgent, CoworkManagedWorkspace}`, `PromptProvenance::LinkWake`, `WorkspaceSurface::Cowork`, and the `cowork` origin kind/entrypoint stay as tolerated legacy reads, documented as retired, so historical rows still parse and render.
- The unrelated GitHub PR `"review"` status kind and the cloud `cowork_api` origin entrypoint (a different feature that shares the name).

## The L1 gap on #1732

`validate_parent_can_spawn`'s cowork depth clause could not be removed while cowork existed. It's gone now, with a negative control (`a_retired_cowork_child_is_no_longer_depth_limited`): a legacy cowork link no longer subordinates, while an unpromoted subagent of the same owner is still refused — so the pass proves the removed clause, not a dead fixture. `specs/anyharness/sessions.md` loses the disclosure paragraph.

## Live coupling not unpicked (disclosed)

`domains/mobility/runtime/*` still reads review rows (`review_store`, `PreflightReviewRun`, the `review_active` blocker) to archive sessions. Removing it is a mobility-contract change and is out of scope here.

Also intended: the client no longer filters `surface === "cowork"` workspaces out of Home matching, so a legacy cowork workspace matches as an ordinary one — with no cowork UI left there is nothing special about it.

## Validation (post-rebase onto #1735's head)

- `cargo test -p anyharness-lib --lib`: green except the one known pre-existing failure (`probe_materialization::sweep_tests`, red on the base commit too).
- product-client: 785 files / 5,316 tests green, both `tsc` configs clean.
- SDK 138 green, sdk-react 34 green.
- `check_docs.py` passes (206 files).
- Boundary, session-mutation-admission, frontend-boundary and appearance checks produce output byte-identical to a base-commit worktree apart from the 14 cowork/reviews allowlist lines this PR removes. All remaining red is inherited from the stack beneath and is swept at the stack tip, as disclosed on #1731 and #1733.
- max-lines allowlist ratcheted (6 entries lowered, `router.rs` and `sdk/src/client/core.ts` dropped); zero stale entries remain.


pablonyx:
## Review — PR #1736 (agent-ops step 8: delete cowork and reviews)

Reviewed `1abeffb63..3297aa277` (frozen pair) by reading; no build run.

**Verdict: comment, no blocker.** The deletion itself is clean — I could not find a single over-deletion, a dangling import, or a legacy-data crash, and the retired-variant keeps are real and wired end to end. Three M-level gaps are worth closing before merge: a dropped legacy-surface pin that also breaks `make clippy`, a matching-widening that happened at a second undisclosed site whose pinning test was deleted rather than inverted, and a strict-parse break in the (env-gated) replay path. The rest is orphan residue.

---

## Findings

### M1 — The agent-ops mount pin lost its legacy-surface case, and the rewrite trips `clippy -D warnings`

`anyharness/crates/anyharness-lib/src/domains/sessions/agent_ops/definition.rs:138`

```rust
for surface in [WorkspaceSurface::Standard] {   // was [Standard, Cowork]
```

Two problems:

1. **Coverage.** `WorkspaceSurface::Cowork` is deliberately *retained* by this PR (`domains/workspaces/model.rs:77`), and this PR simultaneously makes legacy cowork workspaces reachable again from Home and the sidebar (see M2). So a session living in a `surface = 'cowork'` workspace is a live case, and the test's own doc comment says its whole point is proving `should_attach` never filters — "would fail if `should_attach` ever went back to returning `false` for any of these cases." After this edit, the only case that would prove that for a non-Standard surface is gone. Failure scenario: someone later swaps a selector into the registration in `app/product_mcp.rs` that gates on `surface == Standard` (exactly the shape the deleted reviews registration had); every legacy cowork session silently launches without the agent-ops MCP and the suite stays green.
2. **Lint.** `for x in [single_element]` is `clippy::single_element_loop` (complexity group, warn by default), and `Makefile:1483` runs `cargo clippy --workspace -- -D warnings`. `make clippy` should now fail. CI does not run clippy (no workflow references it), so the PR checks won't show this.

Fix: keep `WorkspaceSurface::Cowork` in the array. That resolves both at once.

### M2 — The "legacy cowork workspaces now match" widening has a second, undisclosed site, and its pinning test was deleted instead of inverted

Disclosed site — `lib/domain/home/home-next-launch.ts`, `findHomeNextLocalWorkspace` (~:232) and `findHomeNextMatchingWorkspace` (~:329) both dropped `workspace.surface !== "cowork"`.

Undisclosed site — `lib/domain/workspaces/cloud/selected-repo-target.ts:16,24` dropped both `isStandardWorkspace` guards. `getRepoForSelectedWorkspace` now (a) returns a repo target when the *selected* workspace is a legacy cowork workspace instead of `null`, and (b) pulls cowork-surface siblings into the `localWorkspaceGroupKey` set that feeds cloud repo-target resolution.

Coverage — `lib/domain/home/home-next-launch.test.ts:162`: the case

```ts
workspace({ id: "cowork", surface: "cowork", currentBranch: "feature/raw-name" }),
```

was **deleted** and the test title edited to drop "and cowork workspaces", rather than inverted to assert the new inclusive behavior. So the intended change is now pinned by nothing in either direction.

Failure scenario: a user who ever used cowork has a managed workspace whose `repoRootId` is the repo they pick on Home and whose branch is their chosen base branch. "Start the work" now attaches to that agent-managed workspace instead of cutting a worktree, and the cloud repo-target resolver groups it with their real workspaces. That may well be the desired outcome, but nothing would catch a later re-tightening (invisible legacy workspaces again) or a further loosening.

Fix: invert the deleted case (assert the cowork-surface workspace *is* a candidate now) and add the equivalent for `getRepoForSelectedWorkspace`; mention the `selected-repo-target.ts` site in the PR body.

### M3 — Replay export/load now hard-fails on any session that carried a retired event

`domains/sessions/replay.rs:351` (`event_record_to_envelope`) parses with `serde_json::from_str::<SessionEvent>(...)` and propagates the error; `export_recording` (`replay.rs:149-152`) collects into `Result<Vec<_>, _>?`; `load_recording` (`replay.rs:195`) deserializes the whole file strictly.

`SessionEvent::SessionLinkTurnCompleted` and `SessionEvent::ReviewRunUpdated` are removed from the contract by this PR, but their rows are durably in `session_events` — they were persisted through `append_offline_runtime_event` (`live/sessions/manager/runtime_events.rs:66-79`).

Consequences on a real user DB:
- exporting a replay recording from any cowork- or review-era session returns `ReplayError::InvalidJson("unknown variant ...")` instead of a recording;
- an existing on-disk recording captured before the deletion now fails to load *entirely* (one bad envelope kills the whole file).

This is M and not H only because replay is env-gated off by default (`replay_enabled()`, `replay.rs:50-54`).

Worth noting the contrast that makes the rest of the system safe: the two read paths that users actually hit use `.ok()?` and drop unparseable rows (`api/http/sessions_events.rs:97`, `api/sse/sessions.rs:260`). Either match that tolerance in `event_record_to_envelope`, or say in the PR body that pre-deletion recordings are not loadable.

### L1 — Production support-bundle sanitizer still enumerates the deleted event types

- `lib/domain/support/session-debug/sanitizer-shape.ts:292` `"review_run_updated"`, `:293` `"reviewFeedback"`, `:297` `"session_link_turn_completed"`
- `lib/domain/support/session-debug/primitive-policy.ts:135-139` — `session_link_turn_completed` and `review_run_updated` primitive contracts
- `lib/domain/support/session-debug-primitive-contracts.test.ts:173` still constructs a `review_run_updated` event to exercise the contract

These are `Set<string>` / `Record<string, …>`, so tsc cannot catch them. They are inert (the server drops those events before a client sees them), but they are exactly the "deleted symbol surviving in non-compiled config" class this sweep was meant to close. `"linkWake"` on `sanitizer-shape.ts:279` is a legitimate keep.

### L2 — Orphan test fixtures for removed members, invisible to tsc

- `hooks/sessions/lifecycle/session-stream-side-effects.test.ts:243` and `hooks/sessions/lifecycle/use-session-stream-flush.test.ts:284` still set `invalidateSessionReviews: vi.fn()` inside `function createTestSessionStreamCache(): SessionStreamCache`. That member was removed (`hooks/sessions/cache/use-session-stream-cache.ts:15-25`). This would normally be an excess-property error, but `apps/packages/product-client/tsconfig.json` has `"exclude": ["src/**/*.test.ts", "src/**/*.test.tsx"]` and vitest transpiles without typechecking — so "both tscs clean" does not cover it.
- `lib/domain/sessions/creation/projected-session-materialization.test.ts:33` still lists `{ kind: "review_child", parentSessionId: "parent-1" }` in an `it.each<SessionRelationship>`; `review_child` was removed from the union (`lib/domain/sessions/directory/relationship.ts`). The case is now an untyped duplicate of the `linked_child` case two lines below it.

### L3 — Stale entries in `scripts/appearance_scaling_baseline.json` for files this PR deleted

`:86` `.../workspace/cowork/CoworkWorkspaceShell.tsx|z-10` = 3, `:135` `.../cowork/sidebar/CoworkManagedWorkspaceList.tsx|workspaces`, `:136` `.../cowork/sidebar/CoworkThreadsSection.tsx|visibleThreads`.

CI stays green because the self-tightening `census_slack` check only covers the `stagedViolations` census (`check_appearance_scaling.py:886` passes `STAGED_CENSUS_KEY`); `standardNumericZ` and `unvirtualizedLongLists` are bounded from above only. So these are three unclaimed allowances rather than a red build — but they are exactly the "census must stay TIGHT" residue the guard's own docstring warns about.

### L4 — `SessionLinkService::find_parent_by_relation` is now dead

`domains/sessions/links/service.rs:353`. Its only caller was the depth clause removed from `subagents/service.rs:192`. `store.rs:359` calls the *store* method, not this wrapper. `pub` on a lib item suppresses `dead_code`, so nothing flags it.

### L5 — `composer.md` lost the blank line before a heading

`specs/codebase/systems/product/chat/composer.md:220-221`: removing the "Cowork hides the permission/access `mode`…" paragraph took the separating blank line with it, so `## 2. Dock Regions` now abuts the preceding sentence. Every other heading in the file has one.

### L6 — Stale rationale text on ratcheted max-lines entries

`scripts/max_lines_allowlist.txt`: the `app/mod.rs` entry is now `639` but its note still ends "…691 = 690 + a rustfmt wrap of the materialization wiring head…". The `openapi.rs` entry (692 → 647) keeps a parenthetical enumerating registrations that partly went away. The numbers are correct; the prose is not.

### L7 — Cosmetic and disclosure nits

- `components/workspace/chat/transcript/TranscriptItemBlock.tsx:316`: `<TranscriptToolCallItemBlock item={item} />` is indented 12 spaces inside a 10-space block. No JS formatter is configured in the repo, so nothing catches it.
- `TranscriptItemBlock.tsx:104` passes `titleFallback="Subagent"`, which is now identical to the component default (`SubagentWakeBadge.tsx:26`).
- `ToolCallWorkDivider` (`tool-calls/ToolCallSummary.tsx`) has no consumer outside its own module now; the `export` can go.
- Disclosure counts are slightly off (nothing wrong with the change itself): the boundaries allowlist lost **14** `DOMAIN_CONTRACT_IMPORT` lines, not 13; OpenAPI lost **14** paths and **33** schemas, not 15/31.

---

## Verified clean

1. **Migrations immutable.** Zero diff under `anyharness-lib/src/persistence/` and `specs/GENERATED/`. `0020_cowork_tables.sql`, `0032_cowork_managed_workspaces.sql`, `0034`–`0036` review SQL, and `0045` all untouched, and the generated schema snapshot still matches.
2. **OpenAPI narrowing is exactly cowork + reviews.** Machine-diffed `sdk/generated/openapi.json`: 14 paths removed / 0 added, 33 schemas removed / 0 added, **0** surviving paths changed, and the only surviving schemas that changed are `PromptProvenance` and `SessionEvent` (the variant removals). No collateral contract drift.
3. **Retired-variant keeps are real and wired.** `SessionLinkRelation::{CoworkCodingSession,ReviewAgent}` still `parse()`/`as_str()` (`links/model.rs:24,34`), still cascade with their parent (`runtime/lifecycle.rs:339`) and are still listed in the inbound-close set (`lifecycle.rs:188`); `SessionLinkWorkspaceRelation::CoworkManagedWorkspace` kept and correctly unreachable from `between()`; `OriginKind`/`OriginEntrypoint::Cowork` kept with the decode test rewritten to construct the struct literally now that `OriginContext::cowork()` is gone (`store/tests/sessions.rs:47-66`); `WorkspaceSurface::Cowork` kept, and I confirmed the doc claim — `workspace_can_purge` (`retire_preflight.rs:482`) does require `Standard`; `PromptProvenance::LinkWake` retained end-to-end including the public mapping (`prompt/provenance.rs:83-92`) and the client's `linkWake` arm (`domain/chats/subagents/provenance.ts`), so legacy wake badges still render.
4. **Legacy transcripts do not break.** `map_event` (`store/events.rs:514`) returns raw `payload_json`; both read paths drop unparseable rows with `.ok()?` (`api/http/sessions_events.rs:97`, `api/sse/sessions.rs:260`). So retired `review_run_updated` / `session_link_turn_completed` rows never reach the SDK reducer's `default: recordUnknown` arm, and old sessions do **not** sprout `Unknown event: …` system messages in the transcript (`TranscriptItemBlock.tsx:260`). `decode_prompt_provenance` warns and returns `None` for the removed `review_feedback` variant, which matches its new doc comment.
5. **The depth-clause negative control is genuine, and no other subordination path regressed.** `a_retired_cowork_child_is_no_longer_depth_limited` (`subagents/service.rs:894-930`) writes a real `cowork_coding_session` link, asserts the child may now spawn, *and then in the same fixture* links an unpromoted subagent and asserts `Err(SubagentError::DepthLimit)` — so the pass is the deleted clause, not an inert fixture. The surviving clause (`service.rs:192`, `is_unpromoted_subagent`) is untouched, and the cowork relation appeared in no other gate.
6. **No over-deletion.** Checked each candidate against BASE: `completionContent` had exactly one consumer (`TurnItemSequence`'s artifact prelude); the `domains/artifacts/*` tree had only cowork consumers — `ArtifactProtectionService::for_surfaces(["cowork"])` at `app/mod.rs:288` was its only registration, and the `super::artifacts` used by `domains/agents/readiness/*` is a different module; `buildStandardRepoProjection` did nothing but the cowork filter; `linkedSessionWorkspaces` was fed only by `useCoworkManagedWorkspaces`; `CreateSessionError::WorkspaceSingleSession` fired only for `surface == Cowork`; `isStandardWorkspace` had one consumer. The `WorkspaceFileProtection` extension point removal is safe for the same reason.
7. **No orphaned references in compiled code or specs.** Every deleted `#product/...` module specifier (69 of them) is absent from the HEAD tree. Extracting all 321 code paths cited in `specs/` and checking them against the HEAD tree: 21 are missing, and **none** is in this PR's deletion set (all pre-existing settings/plugins drift). No untouched spec still describes cowork or reviews as live.
8. **Preferences migration is honest.** `coworkWorkspaceDelegationEnabled`, `reviewDefaultsByKind`, `reviewPersonalitiesByKind` drop out of `USER_PREFERENCE_KEYS`, `NEW_USER_DEFAULTS`, `PERSISTED_RECORD_BACKFILL`, `selectPersistedUserPreferencesSlice` and `readLegacyUserPreferences` together, and the legacy read merges a partial record — stored values for the removed keys are ignored, not a migration failure.
9. **Mobility is fully decoupled, not half-broken.** `ReviewStore` is removed from `MobilityRuntime`'s fields, constructor and `app/mobility.rs` wiring; the `review_active` blocker, `PreflightReviewRun`, the dedicated test and the blocker-order pin were all updated in lockstep; the archive record set never carried review or cowork tables, so import/export is unaffected.
10. **Max-lines ratchet only lowers or drops, and is exact.** Verified every changed entry against `git show | wc -l` at HEAD: events.rs 969, openapi.rs 647, retire_preflight.rs 611, transcript.ts 1509, transcript.test.ts 1576, app/mod.rs 639 — all exact. The two dropped entries are legitimately under the limit (router.rs 600, sdk client core.ts 597, `MAX_LINES = 600`). `check_max_lines.py:166` fails on slack, so no headroom was silently left behind.
11. **Other guards only shrank.** 14 `DOMAIN_CONTRACT_IMPORT` allowlist lines removed, 8 `session_mutation_admission.txt` lines removed, `PRODUCT_SURFACE_DOMAINS` lost exactly `cowork` and `reviews`. No threshold loosened, no rule disabled, no new allowlist entry.
12. **Rebase seams landed on the PR9 state.** `SubagentWakeBadge` has no `parentTitle` prop and none is passed; `titleFallback` collapses to `"Subagent"` on the wake arm and `"Agent"` on the session-wake arm; `api/mod.rs` drops only the `review_admission_tests` registration; `composer.md` carries the union — PR9's agent-ops/dock language preserved, review and cowork prose removed.
13. **Disclosed keeps survive.** The cloud `cowork_api` origin entrypoint (`server/proliferate/server/cloud/workspaces/domain/origin.py:18`, settings IA spec), the GitHub PR "review" status kind (`adapters/hosting/*`, `sdk/src/types/hosting.ts`), and the Tier-2/3/4 "Cowork API key" scenarios in `specs/TESTING/*` are all untouched.
14. **Spec honesty.** `specs/codebase/systems/product/README.md:50` and `agents/delegated-work.md:46` explicitly record the deletion rather than quietly dropping the topic; `specs/anyharness/sessions.md:99-105,654` documents the retired-but-parsed relations, which matches the code.

pablonyx:
Review fixes and the stack lint sweep are pushed as two commits: `1e654ffa4` (review fixes) and `fa2bfc3d8` (checker debt).

**`fix(agent-ops): review fixes for the cowork deletion`**

**M1** — `WorkspaceSurface::Cowork` is back in the agent-ops mount sweep. The naive restore wouldn't have caught a future re-narrowing, since the surface array is just data inside the test, so the sweep is now coverage-pinned: per-variant flags set inside an exhaustive `match`, asserted after the loop. Adding a variant stops it compiling; dropping one from the sweep fails the assert. Negative control: removed `Cowork` again, `covered_cowork` fails. This also clears the `clippy::single_element_loop` that would have broken `make clippy`.

**M2** — the filters stay deleted and the widening is pinned instead of implicit. `home-next-launch.test.ts` gets the `surface: "cowork"` case back, inverted (it matches now, and a legacy cowork local checkout is selected), and a new `selected-repo-target.test.ts` pins repo-target resolution and sibling grouping for legacy cowork workspaces. `specs/anyharness/workspaces.md` now says it in prose: `cowork` is a retired-but-parseable value, no read path filters on it, and `workspace_can_purge` is the one place it still decides anything. Negative control: re-added an `isStandardWorkspace` guard, 5 tests fail across the two files.

**M3** — `replay.rs` now drops rows the current `SessionEvent` contract can't parse, the same way `api/http/sessions_events.rs` and `api/sse/sessions.rs` already do, with a comment on why: `session_events` is durable and append-only, so retired variants are on disk in every pre-deletion session, and propagating the parse error made export fail outright for those sessions rather than emit the events that still exist. Recordings already on disk get the same tolerance; a file that isn't a JSON array at all is still a hard `InvalidJson`. Two new tests drive a verbatim `session_link_turn_completed` row through export and through the on-disk parser. Negative control: restored the strict parse, both fail with `unknown variant session_link_turn_completed`.

**L1** — all six. Sanitizer `SAFE_TYPE_VALUES` and `primitive-policy.ts` `EVENT_CONTRACTS` no longer enumerate the deleted event types, and the contracts test now pins that `review_run_updated` fails *closed* — unsafe `type` value, every field redacts. Two `invalidateSessionReviews` factories and the `kind: "review_child"` case fixed, three stale `/cowork/` appearance-baseline entries removed, the dead `SessionLinkService::find_parent_by_relation` wrapper deleted, and the blank line before composer.md §2 restored. The PR body's counts (14 paths / 33 schemas / 14 allowlist lines) are corrected.

**`chore(agent-ops): clear the stack-introduced checker debt`**

This is the stack-tip sweep disclosed on #1731 and #1733. All six checkers were verified green at this stack's merge-base with main (`d1f13cc8d`) in a throwaway worktree, so everything below is debt this stack created, not pre-existing.

*max_lines* — seven files get their inline `mod tests` extracted to a sibling `*_tests.rs` via the repo's `#[path]` idiom, which keeps `super::*` privacy exactly. Three land clean under the limit that way and need no entry at all: `store/agent_wakes.rs` (960 → 410), `transcript_read.rs` (624 → 405), `wakes/service.rs` (778 → 219). Four more leave a smaller entry than the un-extracted file would have.

**Ten net-new allowlist entries remain and want a founder look.** Each is marked `agent-ops stack` under a header saying plainly that this is deferred debt being recorded, not excused:

| file | limit |
| --- | --- |
| `agent_ops/calls.rs` | 1438 |
| `agent_ops/config_ops_tests.rs` | 680 |
| `agent_ops/peer_ops_tests.rs` | 737 |
| `agent_ops/spawn_ops_tests.rs` | 778 |
| `agent_ops/tools.rs` | 1334 |
| `agent_ops/workspace_ops.rs` | 1640 |
| `sessions/ownership/service_tests.rs` | 674 |
| `sessions/store/sessions.rs` | 731 |
| `sessions/store/tests/sessions.rs` | 705 |
| `sessions/subagents/service.rs` | 1000 |

Plus four bumps: contract `v1/sessions.rs` 779 → 876, `api/session_admission_tests.rs` 650 → 870, `sessions/runtime/tests.rs` 1019 → 1070, `sdk-react/src/hooks/sessions.ts` 723 → 754.

`calls.rs`, `tools.rs`, `workspace_ops.rs` and `subagents/service.rs` were measured, not guessed: extracting their tests leaves 1271 / 616 / 748 / 722 production lines, all still over, so it would buy two entries instead of one. Those numbers are in the rationale text.

*session_mutation_admission* — both handlers classified, nothing relaxed. `sessions_wakes::schedule_agent_wake` is fenced: it arms a wake that queues a prompt on the watcher session, so it is admitted against that session's other mutations before the arm. `subagents::promote_subagent` is read-like — one indexed `UPDATE` of `promoted_at`, no session touched — and it deliberately takes no permit while its neighbour `schedule_subagent_wake` does, because a permit would refuse promotion whenever a workflow held the parent, for a write the workflow cannot conflict with. That asymmetry is already ratcheted in `api/session_admission_tests.rs`.

*anyharness_boundaries* — the four stale entries (`sessions/delegation.rs` and the three `subagents/mcp|transcript` paths) are gone and `sessions/replay.rs` is **tightened** 2 → 1. The one real violation is fixed properly rather than allowlisted: `api/http/sessions_wakes.rs` was importing `AgentWakeReason` straight from the store, so the wakes module re-exports it and the handler names the domain facade — handlers call facades, never stores.

Seven entries are added. Three document occurrences that only *moved* under the agent-ops rename (`calls_helpers.rs`, `definition.rs`, `transcript_read.rs` — same occurrence, new path). **Four are NET-NEW and want a founder look**: `DOMAIN_LIVE_VALVE agent_ops/workspace_ops.rs 1`, `DOMAIN_LIVE_VALVE sessions/wakes/hooks.rs 1` (mirrors the `subagents/hooks.rs` entry that already exists), `DOMAIN_CONTRACT_IMPORT agent_ops/calls.rs 2`, and `DOMAIN_CONTRACT_IMPORT agent_ops/config_ops.rs 1`.

Disclosure: the max-lines extractions incidentally moved three occurrences out of this checker's view (`config_ops.rs:524`, `peer_ops.rs:276`, one in `replay.rs`), because `should_skip()` ignores `*_tests.rs` by design. That is why `replay.rs` tightened rather than clearing entirely.

*frontend_boundaries* — seven violations, **no allowlist entries added**. Five were `#product/domain/...` self-references written inside the `src/domain` cell; they become relative imports, which is what the other ~300 imports in that tree already do. The last two were `pending-agent-updates.ts` reaching up out of the Mobile-safe `src/domain` cell into `lib/domain/delegated-work`. Rather than punch a hole in that boundary for one file, the module and its test move to `lib/domain/chat/composer/`, where the identity vocabulary they need lives; three consumer import paths follow. The header comment records the alternative — hoisting the delegated-work identity vocabulary down into `src/domain` — as the other honest fix, and why it is a separate change (~30 callers).

`check_appearance_scaling.py` and `check_docs.py` stay green.

**Validation**

- `cargo test -p anyharness-lib --lib` — 1709 passed, 1 ignored. The one red is the known pre-existing `probe_materialization::sweep_tests::the_sweep_removes_only_abandoned_and_old_scratch_roots`.
- product-client vitest — 786 files / 5320 tests passed.
- Both `tsc` passes (`tsconfig.domain.json`, `tsconfig.json`) clean.
- All six checker scripts green.

**Clippy, reported not fixed.** `cargo clippy -p anyharness-lib --lib -- -D warnings` never reaches this crate: it dies in `anyharness-contract` on `clippy::large_enum_variant` for `SessionEvent` (`ConfigOptionUpdate` 856 bytes vs `InteractionRequested` 392). I checked that one specifically, since this stack deletes `SessionEvent` variants and could have changed the second-largest — it reproduces byte-for-byte at `d1f13cc8d`, so it is pre-existing. `anyharness-lib` itself emits 210 warnings and one deny-level `clippy::never_loop` at `live/sessions/actor/turn/active.rs:46`, a file this stack does not touch. Both left alone. The one exception is `spawn_ops.rs:322`, a stack-introduced `clippy::manual_contains` — a one-token fix, so it was taken.

pablonyx:
Live-QA defect fixed forward @ 875ad0c77 (branch also rebased onto #1735's new head ba3aaee44).

**Defect:** with the stack tip running, the sidebar flooded with a "Cowork" group and UUID-named legacy workspaces. This PR's deletion of `buildStandardRepoProjection` removed the only filter between the raw collections and the offer surfaces — its consumers were mainline hooks (`use-standard-repo-projection` feeds the sidebar rail, home repository picker, settings repositories, and command palette), not cowork-only ones. The review's "cowork-only consumers" verification was wrong in effect; the earlier fix pass had (correctly) noted the deletion but not restored a replacement.

**Fix:** `standard-projection.ts` restored in an API-free form — the cowork status API is gone, so the retired root is identified by `kind === "managed"` (the runtime-owned cowork checkout was the only producer of managed roots; confirmed against live `/v1/repo-roots`) and legacy workspaces by `surface === "cowork"`. Scope is deliberately the rail projection only: resolution, launch matching and repo-target grouping stay wide, so the M2 pins hold and anything referring to a legacy cowork workspace still opens — retired rows are reachable, just never offered. `specs/anyharness/workspaces.md` prose updated to match.

Tests: restored suite (managed root hidden, cowork surface hidden, external/cloud passthrough) with negative control (dropping the filter fails), plus the 7 consumer suites — 52/52 green, tsc clean.

pablonyx:

PR #1746, opened 2026-08-11
fix(product): scope workspaces filter to name and branch (PRO-122)

tairabun:
Fixes [PRO-122](https://linear.app/proliferate-team/issue/PRO-122/filter-workspaces-search-does-not-make-sense-what-its-searching).

## Problem

The workspaces page filter matched against the workspace **id**, repo meta, PR number, and placement label in addition to the visible name — so typing e.g. a few letters could surface rows with no visible relation to the query, and it was unclear what the search was actually searching.

## Change

- Filter now matches **only the workspace name or branch name**: the cmdk item `value` is the (unique) workspace id used purely for selection identity, and name/branch are passed as `keywords` with a custom binary substring filter.
- The binary score also preserves the recency grouping/order instead of cmdk's score-based reshuffle.
- Placeholder updated to `Filter by name or branch...` so the scope is self-describing.
- The dashed Create row is `forceMount` and unaffected.

## Tests

- Replaced the PR-number-matching test with tests asserting matches on name and branch, and non-matches on id / meta / PR number / placement.
- `pnpm vitest run src/components/workspace/repo-setup/WorkspacesCommandList.test.tsx` — 9 passed.
- `pnpm exec tsc -p tsconfig.json --noEmit` in `apps/packages/product-client` — clean.

### DIFF
--- a/apps/packages/product-client/src/components/workspace/repo-setup/WorkspacesCommandList.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/repo-setup/WorkspacesCommandList.test.tsx
@@ -114,27 +114,51 @@ describe("WorkspacesCommandList", () => {
     expect(screen.queryByText("Cloud")).toBeNull();
   });
 
-  it("matches the filter against the PR number", () => {
+  it("matches the filter against the workspace name or branch only", () => {
+    render(
+      <WorkspacesCommandList
+        groups={groups([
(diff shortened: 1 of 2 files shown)


tairabun:
Reopening from an in-repo branch so the Vercel preview can deploy without fork authorization — replaced by the follow-up PR.

PR #1750, opened 2026-08-11
perf(product): keep sidebar separator drags smooth

tairabun:
## Linear tickets

- [PRO-95 — Sidebar dragging lags](https://linear.app/proliferate-team/issue/PRO-95/sidebar-dragging-lags)

## Summary

- keep pointer-driven sidebar widths in transient React state and commit the durable workspace preference once when the gesture ends
- apply the shared resize path to standard workspace, page, Cowork, and Settings sidebars
- snap left-shell geometry during a live drag while preserving the existing eased open/close motion

## Root cause

Each document `mousemove` called the persisted Zustand `setSidebarWidth` action. The workspace UI lifecycle then selected and sanitized both persisted slices, serialized both for equality, serialized the current slice again for storage, and wrote it through the host storage adapter for every pointer event.

At the same time, the standard shell transitioned `--workspace-left-width` for the full 240 ms panel duration. The page and Cowork shells likewise transitioned `width`. Every incoming mousemove retargeted that layout transition; the manual fallback cancelled and restarted its equivalent rAF tween. The separator therefore eased toward stale targets instead of staying under the cursor while the whole shell repeatedly re-laid out.

## Verification

- fail-before regression on `origin/main`: 100 real document mousemoves produced 100 durable sidebar-width updates
- pass-after regression: the same gesture renders the exact final 375 px width, produces 0 durable updates during the drag, and exactly 1 on release
- release-without-movement produces 0 commits; unmount-mid-drag commits the latest width once
- manual geometry fallback schedules 0 interpolation frames during the drag and 0 tail frames on release
- `@proliferate/product-client` production build and typecheck pass
- 25 focused resize/shell tests pass
- 5,457 product-client tests pass when excluding the unrelated browser-only Markdown cascade test
- frontend boundary, appearance scaling, design attribution, and diff checks pass

The full default test command reached 5,457 passing tests; its only failure was the browser-only Markdown cascade suite because Google Chrome is not installed at the expected macOS application path.

## Observability

No telemetry schema or event changes. The existing workspace UI diagnostic will naturally report one durable sidebar-width write per completed gesture instead of one per pointer event.

### DIFF
--- a/apps/packages/product-client/src/app/authenticated-workspace-shell-motion.test.ts
+++ b/apps/packages/product-client/src/app/authenticated-workspace-shell-motion.test.ts
@@ -15,6 +15,7 @@ describe("authenticated workspace shell motion", () => {
     expect(stylesheet).toContain("var(--ease-out-cubic)");
     expect(stylesheet).toContain("[data-snap-left-geometry=\"true\"]");
     expect(stylesheet).toContain("[data-snap-right-geometry=\"true\"]");
+    expect(stylesheet).toContain("--workspace-left-geometry-duration: 0ms");
     expect(stylesheet).toContain("--workspace-right-geometry-duration: 0ms");
     expect(stylesheet).toContain("[data-manual-workspace-geometry=\"true\"]");
   });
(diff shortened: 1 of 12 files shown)


tairabun:
validated manually

PR #1752, opened 2026-08-11
feat(observability): freeze diagnostics v1 contracts

pablonyx:
## Summary

- Freeze the approved Rust Observability ADR, closed 91-operation P0 lifecycle catalog, semantic ownership rules, thirteen-slice registry, and v1.1 compatibility/boundary decisions.
- Add one runtime-free `proliferate-diagnostics-protocol` Rust crate plus equivalent pure TypeScript and Python contract representations.
- Pin valid/invalid records, lifecycle sequences, API shapes, hard bounds, privacy restrictions, cross-language edge semantics, and the release RSS profile in one shared golden fixture tree.
- This is PR 1 only: no collector, producer, transport, exporter, support-flow, or instrumentation behavior is implemented.

Frozen spec: PS-1 revision 1 (`ebf43576daf6aead25b49d7e998f5304876400d4f2b5fffce751573ba27873fa`), base `fcf0f36b4ad4d3bf8a2646ab887c6374d3778642`.

## Testing

Tier 1 contract/unit and repository constitution checks:

- `cargo test -p proliferate-diagnostics-protocol` — 9/9 golden tests pass; cargo/rustc lock checked first.
- ProductClient canonical typecheck — pass.
- ProductClient diagnostics Vitest — 10/10 pass, including lone-surrogate rejection.
- Server targeted Ruff, format, and mypy — pass.
- Server diagnostics pytest — 12/12 pass, including lone-surrogate rejection.
- Documentation integrity, AnyHarness/frontend/server boundaries, max-lines, design attribution, JSON parsing, and diff whitespace — pass.
- Round-1 review blockers RV-01 and RV-02 are repaired; RV-03 through RV-05 remain unchanged Follow-ups.

### Fixture parity / rejection matrix

| Contract proof | Rust | TypeScript | Python |
| --- | --- | --- | --- |
| Valid record/API roundtrip | pass | pass | pass |
| Detailed vs lifecycle + all terminals | pass | pass | pass |
| Retry/duplicate/conflict/death/orphan sequence | pass | pass | pass |
| Version skew + unknown optional/null normalization | pass | pass | pass |
| Secrets + model/plugin prohibitions | pass | pass | pass |
| Loopback/FD/timestamp/integer/map/gap/version edge cases | pass | pass | pass |
| Exact catalog, bounds, and RSS profile | pass | pass | pass |

Contract generation was not selected, so generated-contract drift output is N/A. `Cargo.lock` was regenerated by Cargo and includes the new crate.

## Review repairs

- RV-01: removed the unused public Rust `parse_and_validate` bypass; shape-specific validators are the only parsing entry points.
- RV-02: TypeScript and Python now reject lone high/low Unicode surrogates as stable `invalid_shape`, with language-local pins because `serde_json` rejects those escapes before contract validation.
- Resolution artifact: `/tmp/proliferate-pr1-inputs.MQk6B5/artifacts/implementation/review-resolution-round-1.md` (handoff artifact; not a repository path).

## Observability

None at runtime. This slice freezes future vocabulary and contracts only. Existing local sinks, server log routing, Sentry, PostHog, anonymous telemetry, and support behavior are unchanged.

## Readiness

- This PR intentionally remains draft for independent `review-pr-slice` review against frozen PS-1.

  in this PR.

## Verification

- No UI exists in this slice; screenshot/demo is not applicable.
- Full product build, Docker, runtime collector execution, and RSS workload execution were not run because PS-1 excludes them; PR 2 owns runtime/RSS execution.
- Exact proof manifest: `/tmp/proliferate-pr1-inputs.MQk6B5/artifacts/implementation/proof-manifest.md` (handoff artifact; not a repository path).


pablonyx:

`codex/rust-observability-pr1-contracts`: `d6bbfbded` → `ee9256f9a` (force-pushed with `--force-with-lease` pinned to the old SHA, so the update would have aborted rather than clobbered a concurrent push). The old head above is the recovery anchor.

This slice owned the frontend gate failure: `validation.ts` (551) and `validation-support.ts` (584) were both new here, so it was new debt rather than inherited. Split along existing seams with `validation.ts` left as the public re-export surface, so no caller changed; `frontend_structure_allowlist.txt` was deliberately not touched, since it documents itself as a ratchet rather than permission for new debt.

pablonyx:
## Dedicated review — `pr1-contracts` at `ee9256f9a`

Scope: 3 commits / 49 files vs `2e69ddf1b`. Read-only review, no builds run.

**No high-severity findings.**

`LOW | CONFIRMED | validation-scalars.ts:89` (and its Rust/Python siblings) — timestamp shape validation accepts out-of-range UTC offsets such as `+99:99`. All three implementations agree, so there is no interop divergence; flagging only in case downstream time parsing assumes validated offsets.

Recorded so nobody re-chases them:

- **Claim:** `jsonByteLength` diverges from the Rust/Python byte counts because `JSON.stringify` escapes non-ASCII. **Rejected.** `JSON.stringify` emits literal UTF-8 for non-ASCII (it escapes only control characters, `"`, `\` and lone surrogates), and `serde_json::to_vec` / `json.dumps(ensure_ascii=False)` do the same, so the counts agree.
- **Not audited, worth one look:** whether absent-vs-`null` encoding of `Option` fields and float formatting agree across the three serializers. Those *would* shift byte counts near `MAX_RECORD_BYTES`. I did not verify this either way.

**Verdict.** Enum values, struct shapes and limits agree across Rust, TypeScript and Python; the golden tests cover valid and invalid shapes plus boundary conditions. Nothing here blocks review.

pablonyx:
## Dedicated review — pr1-contracts at ee9256f9a

Scope: `git diff origin/main...ee9256f9a`, 3 commits / 49 files (ADR, `proliferate-diagnostics-protocol` crate + golden test, ProductClient `domain/diagnostics/*` + vitest, `server/.../observability/*` + pytest, `fixtures/contracts/rust-observability-v1/**`, spec docs). Read-only review; **no builds were run** — no cargo, no rustc, no pnpm, no pytest, no Docker. Everything below was verified by reading the code paths end to end.

First, what holds up. I traced all 15 `invalid/records.json` cases and all 14 `invalid/api.json` cases through Rust, TypeScript and Python individually: **all 29 reject for the same pinned reason in all three.** The P0 catalog is an exact ordered 91/91 match across the fixture, `catalog.rs`, `catalog.ts` and `catalog.py`, and I re-expanded the ADR's brace notation independently and also got 91. The limit constants are identical in all three `limits.*`. Argument depth accounting (`depth=1` at the top, `depth+1` per level, `> MAX_ARGUMENT_DEPTH`) is identical in all three. `finalizer=collector` ⇒ `outcome=abandoned`, terminal ⇒ outcome present, `failed` ⇒ `error_classification` present, exactly-one-payload-class, non-P0 lifecycle name rejection, and the orphan/duplicate/conflict sequence dispositions all agree. The crate depends on `serde` + `serde_json` only and contains no runtime — the PR-1 gate is met.

The findings below are all latent: none of them makes a currently-pinned fixture fail. They are places where the three implementations do **not** actually agree, which is the one thing this slice exists to guarantee.

---

**MEDIUM | CONFIRMED | `anyharness/crates/proliferate-diagnostics-protocol/src/v1/validation_support.rs:46` (with `validation.rs:24`)**

Rust only runs `reject_secret_fields` on raw JSON for producer records and connection descriptors; every other API shape is validated as an already-deserialized struct, so a prohibited secret field carried on a collector-accepted record cannot be rejected in Rust while TypeScript and Python do reject it.

Concrete scenario: take `valid/api.json`'s `records_page` and add `"authorization": "Bearer synthetic"` to the record inside `records[0].record`. TypeScript `parseRecordsPageV1` → `parseCollectorAcceptedRecordV1` → `parseProducerRecordV1` → `rejectSecretFields(raw)` (validation-records.ts:56) throws `prohibited_secret`; Python `parse_records_page_v1` → `parse_collector_accepted_record_v1` → `parse_producer_record_v1` → `_reject_secret_fields` (validation.py:83) raises `prohibited_secret`. Rust has no `parse_records_page_value`: `serde_json::from_value::<RecordsPageV1>` silently drops the unknown key, and `validate_records_page` → `validate_accepted_record` calls `validate_producer_record` (typed), which never calls `reject_secret_fields` — it returns `Ok`. Add that as an `invalid/api.json` case with `"reason": "prohibited_secret"` and the Rust golden test fails on `unwrap_err()` while the other two pass. The same asymmetry applies to `tail_frames`, `export_frames`, `ingest_receipt`, `records_query` and `health`. Note the ingest path is *not* affected — `parse_ingest_batch_value` correctly routes each record through `parse_producer_record_value` (validation.rs:184), so the security-critical direction fails closed.

**MEDIUM | CONFIRMED | `apps/packages/product-client/src/domain/diagnostics/validation-support.ts:94` and `validation-scalars.ts:106`**

The JSON *number form* is unpinned, and no fixture exercises the `float` typed-argument variant at all (`grep -rn float fixtures/` returns nothing), so two concrete number-form divergences are untested and live.

Serialization divergence: an argument `{"type":"float","value":5}` parses in all three, but TypeScript re-emits `5` (`requireFiniteNumber` returns the JS number, `canonicalClone` → `JSON.stringify` → `5`) while Rust re-emits `5.0` (`ArgumentValueV1::Float(f64)`, types.rs:196) and Python re-emits `5.0` (`_require_finite_number` returns `float(value)`, validation_support.py:485). Adding that argument to `valid/records.json` breaks the Rust roundtrip assertion at golden_contract.rs:42 (`Number::PosInt(5) != Number::Float(5.0)`) while the TS `toEqual` and Python `==` assertions both pass, because `5.0 == 5` in both of those languages. Three languages, three different wire bytes for the same declared value.

Accept/reject divergence: `"producer_sequence": 1.0` is **accepted** by TypeScript (`Number.isInteger(1)` is true after `JSON.parse`, validation-scalars.ts:108) and **rejected as `invalid_shape`** by Python (`isinstance(1.0, int)` is false, validation_support.py:489) and by Rust (`serde_json` stores `1.0` as `N::Float`; the `u64` visitor has no `visit_f64`, so `from_value` errors → `InvalidShape`). `"producer_sequence": 1e300` splits the same way except TypeScript says `limit_exceeded` rather than accepting. The lone-surrogate case got an explicit "language-local by design" comment and a test in both TS and Python — this equally inherent JSON-numeric-model gap got neither.

**LOW | CONFIRMED | `fixtures/contracts/rust-observability-v1/limits.json:29`**

The `boundary_cases` array pins 20 at/over boundaries with rejection reasons and is read by nothing: `grep -rn boundary_cases` across the crate, `apps/packages/product-client/src`, `server/proliferate`, `server/tests` and `fixtures` returns exactly one hit — its own declaration.

Each language instead asserts four representative boundaries by hand (id bytes, record bytes, batch bytes, cardinality entries). So 16 of the 20 declared surfaces — `safe_integer`, `string_bytes`, `message_bytes`, `name_bytes`, `arguments`, `argument_list_items`, `argument_object_fields`, `argument_depth`, `batch_records`, `filter_values`, `tail_frame_records`, `export_records`, `export_bytes`, `versions_present`, `health_producers`, and the `page_records` at-limit — have a pinned reason that nothing checks. I read each of those paths and they currently do agree across the three languages, so this is a false-assurance/regression-surface problem rather than a live bug: the file reads as the bounds contract but enforces none of it, and the ADR Validation section names "bounds" as a required coverage area.

**LOW | CONFIRMED | `apps/packages/product-client/src/domain/diagnostics/validation-scalars.ts:85` (siblings `validation_support.rs:190`, `validation_support.py:108`)**

Re-checking the prior finding: **still present, and broader than reported.** Timestamp validation is shape-only in all three, so out-of-range components pass everywhere.

`"2026-08-10T15:00:00+99:99"` is accepted by TypeScript (regex `[+-]\d{2}:\d{2}`, length 25 inside the 20–35 window), by Python (identical `TIMESTAMP_PATTERN`) and by Rust (`zone[1..3]` and `zone[4..6]` are only checked `is_ascii_digit`). The same is true of the date and time fields — `"2026-99-99T99:99:99Z"` is accepted by all three, since the prefix check is `\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}` / a positional `is_ascii_digit` scan with no range check. This is consistent across languages, so it is not a divergence; it means `source_timestamp` and `accepted_timestamp` cannot be trusted as orderable instants by PR 2 without a second parse.

**LOW | CONFIRMED | `apps/packages/product-client/src/domain/diagnostics/validation-support.ts:308` (siblings `validation_support.rs:242`, `validation_support.py:393`)**

`rejectSecretFields` matches JSON object *keys*, so a prohibited secret name expressed as a typed-argument name passes validation and is retained, while the identical name as an envelope key is rejected.

Concrete: `"arguments": [{"name": "authorization", "privacy": "operational", "value": {"type": "string", "value": "Bearer synthetic"}}]` is accepted by all three — the walker sees keys `name`/`privacy`/`value` and recurses into the string `"authorization"`, which is a value, not a key — and `requireName("authorization")` passes the charset. The same record with a top-level `"authorization"` key is the pinned `unknown_authorization_field` → `prohibited_secret` case. Consistent across all three languages, so not a divergence; flagging because "prohibited fields genuinely cannot pass validation" only holds for the envelope, and typed arguments are the retained, high-volume surface.

**LOW | CONFIRMED | `anyharness/crates/proliferate-diagnostics-protocol/src/v1/validation.rs:251` vs `validation-exports.ts:44` / `validation.py:297`**

Check ordering differs between Rust and TS/Python, so a multi-defect input gets a different pinned reason depending on the language.

Concrete: an export request with `purpose: "support"`, no `support_authorization_id`, and `byte_limit: 33554433` returns `limit_exceeded` from Rust (validation.rs:251 checks the limits before the purpose/authorization match at :258) and `invalid_shape` from TypeScript and Python (both check purpose/authorization first). Two more instances: `parse_ingest_batch_value` checks the schema version before the 128-record count (validation.rs:175 vs :180) while TS/Python check the count first, so a 200-record batch with `major: 2` is `unsupported_major` in Rust and `batch_too_large` in the other two; and a health response with both `retained_bytes` over the arena limit and a bogus `evictions_by_class` key is `invalid_shape` in Rust (serde key error precedes validation) and `limit_exceeded` in TS/Python. Every pinned fixture is single-defect, so nothing fails today.

---

- **Catalog drift from the ADR.** Chased by extracting the operation lists from the fixture, `catalog.rs`, `catalog.ts` and `catalog.py` and diffing them ordered: identical, 91 entries, no duplicates, `owning_pr` values `[2,3,8,6,9,9,10,11]` matching the ADR table. I also expanded the ADR's brace notation by hand and independently counted 91. No drift.
- **Limit constants drifting between languages.** Compared `limits.rs`, `limits.ts`, `limits.py` and `limits.json` field by field: all identical, including `MIN_SUPPORTED_PRODUCER_MINOR = 0` with `SUPPORTED_PREVIOUS_PRODUCER_MINOR_WINDOW = 1`, which matches the ADR's "accepts producer minors 1.1 and 1.0, window is one".
- **Unknown-optional / null-normalisation divergence.** `valid/unknown-optional.json` carries `parent_operation_id: null`, an unknown nested `detailed.future_detail_option` and an unknown top-level `future_envelope_option`. TS `optionalId(null) → undefined` then `canonicalClone` drops it; Python `_set_optional` skips `None`; Rust `Option<T>` deserialises `null` → `None` with `skip_serializing_if`. All three produce the pinned `canonical`. No finding.
- **Nested prohibited model/plugin content escaping the denylist.** `lifecycle.model.extra.prompt` is not caught by `reject_prohibited_model_plugin_metadata` (top-level keys only) in any of the three — but `parseModelMetadata`/`_parse_model_metadata`/`ModelMetadataV1` are allowlists, so `extra` is dropped and no content is retained. Consistent across languages and no content passes. Not a defect.
- **`SECRET_FIELD_NAMES` omits `api_key`, `password`, `token`, `client_secret`.** Real, but the typed parsers are allowlists, so such an envelope field is dropped rather than retained; it simply produces no `prohibited_secret` reason. Scrubbing is PR 6's contract gate, not PR 1's. Left alone deliberately.
- **ID fields have no charset restriction.** `requireId`/`validate_id`/`_require_id` are byte-bounded only, so `model_id`, `operation_id`, `trace_id` etc. can each carry 128 bytes of arbitrary text. Consistent across all three and the ADR does not require a charset for IDs (it requires safe *kind/category* and IDs to be "already-owned"). Not a contradiction.
- **`sameRecord` in TS comparing by `JSON.stringify`.** Looked like a key-order hazard versus Rust's `PartialEq` and Python's dict equality, but both sides of every comparison come out of `parseProducerRecordV1` → `canonicalClone` with a fixed literal key order, so the comparison is stable. Refuted.
- **`serde_json` map ordering breaking the Rust roundtrip asserts.** `rejections_by_reason` is textually `prohibited_secret, record_too_large` in `valid/api.json` but `BTreeMap<RejectionReasonV1, _>` serialises by discriminant (`record_too_large` first). Refuted: `serde_json = "1"` with no `preserve_order`, so `Value::Object` is a `BTreeMap` and equality is order-independent; the `arguments[2]` object keys `stage`/`summary` are also already in BTreeMap order.
- **`Cargo.lock` bumping `proliferate` 0.4.6 → 0.4.8.** Appears in the three-dot diff but is a stale-lock refresh: the branch base is `2e69ddf1b` while `origin/main` is now `ec031a620`. Noise, not a change this PR makes.
- **PR-1 scope creep / runtime behaviour.** Crate deps are `serde` + `serde_json` only, `lib.rs` states the no-runtime constraint, `members = ["anyharness/crates/*"]` already covers the new crate, and `validate_rss_profile` only checks profile *completeness* (execution is explicitly PR 2's gate, matching the ADR). No contradiction with the frozen slice definition.

pablonyx:
## Post-review fixes — head is now `ede8cae1a`

Two commits since the sealed review head `ee9256f9a`:

- `9596d7721` — the review's parser-divergence finding: the Rust validation surface exposed only three raw-JSON `parse_*_value` entry points while TS validated every read-back shape, so Rust callers of the other shapes parsed without validation. Now ten fail-closed entry points.
- `ede8cae1a` — `cargo clippy` blockers in this crate: `validation_support.rs:141` tripped `absurd_extreme_comparisons` (dead `minor < 0` comparison; rewritten as a range `contains` that stays correct if the floor moves off zero), and the golden-contract const assertion became a `const` block. Negative control: reverting both reproduces the clippy error; `cargo test -p proliferate-diagnostics-protocol` stays 9/9.

Not restacked (base of the train); fast-forward push, no history rewritten.

PR #1754, opened 2026-08-11
feat(observability): add standalone diagnostics collector core

pablonyx:
> **DRAFT — RSS validation blocked by host memory pressure.** The round-1 source repairs and targeted release suite are green at `40c9f404441427d21bc5966e1f7562c41bc8a1ff`, but neither required architecture RSS profile has a completed receipt. Do not mark ready or merge until both exact receipts pass.

## Summary

- add a separately executable, memory-only Rust diagnostics collector against the frozen schema v1.1 protocol
- provide capability-authenticated loopback ingest, records query, resumable NDJSON tail, point-in-time NDJSON export, health/readiness, inherited-FD control, and bounded shutdown
- enforce deterministic accepted order, oldest-first bounded retention, explicit gaps, lifecycle sequencing, producer/resource caps, and neutral standalone exporter/fallback/restart health
- resolve RV-2-01 by replacing whole-batch `Value`/record cloning with zero-retention wire validation, borrowed bounded records, one-at-a-time record parsing, capped prepared encodings, and one shared body-parser permit
- resolve RV-2-02 with a fixed-capacity authenticated body reader that produces accounted 413 JSON `batch_too_large` responses for far oversized ingest/export requests and stops chunked bodies at first overflow
- add standalone process, security, resource-bound, golden-fixture, and RV-03/RV-04/RV-06 tests plus the exact release RSS evidence runner
- document only the landed collector ownership, process/API seam, and validation path

This is stacked on `codex/rust-observability-pr1-contracts` at `d6bbfbdede17f83abeae5e85c6832b1a58bd1d7c`. It intentionally contains no Desktop/Tauri/renderer, AnyHarness runtime, Worker, server, cloud, OTLP, support-archive, or later-slice integration. RV-05 remains deferred to PR 12.

## Testing

- PASS: `CARGO_BUILD_JOBS=2 cargo test --locked --offline --release -p proliferate-diagnostics-protocol -p proliferate-diagnostics-collector --all-targets -- --test-threads=1` — 33 passed, 0 failed, 0 ignored at the repair head; final run emitted no warnings
- PASS: focused concurrent maximum/degenerate exact-1-MiB ingest proof sampled only the filled-arena collector child at 26,198,016 bytes peak versus the 52,428,800-byte ceiling
- PASS: focused far-oversized ingest/export plus chunked overflow proof returned stable 413 JSON and incremented both health-counter dimensions exactly
- PASS: `rustfmt --edition 2021 --check` across all new Rust sources and tests
- PASS: `git diff --check`
- PASS: `python3 scripts/check_docs.py` — 211 Markdown files
- PASS: `python3 scripts/check_design_attribution.py`
- BLOCKED: post-repair `aarch64-apple-darwin` exact RSS profile started from a pressure-level-1/no-build gate but was stopped at approximately 90 seconds when host pressure rose to level 2. No receipt file was emitted; this is not a completed run and is not claimed as a pass.
- NOT RUN: post-repair `x86_64-apple-darwin` RSS profile because pressure remained level 2 after the native stop. No binary was relabeled and no result was fabricated.

## Observability

- adds only the collector's contract-defined local health and its four owned lifecycle operations: `collector.boot`, `collector.shutdown`, `collector.export`, and `collector.producer.attach`
- does not change Sentry, PostHog, server structured logging, anonymous telemetry, existing local sinks, product-domain instrumentation, or any vendor exporter

## Readiness

- [ ] Both exact architecture RSS-profile receipts pass and are attached. This is the explicit blocker for ready-for-review status.

- [x] This PR has no support relationship and includes no tracker, report, user, support, or private-source identifiers.

## Verification

- authenticated loopback and cross-origin rejection
- raw-wire authoritative RV-03/RV-06 rejection plus recovery
- RV-04 orphan-terminal/late-start conflict accounting
- deterministic query/filter/pagination and golden export compatibility
- tail resume/lag/slow-reader disconnect, point-in-time export/cancellation, eviction/cursor gaps
- every consumed frozen constant at and over its boundary, plus runtime connection/handler/tail/export/lifecycle/gap caps
- inherited capability remains absent from arguments, URLs, descriptor, logs, files, health, and exported records; restart restores no state


pablonyx:

`codex/rust-observability-pr2-collector-core`: `bb7437cc1` → `108abc5f7` (force-pushed with `--force-with-lease` pinned to the old SHA, so the update would have aborted rather than clobbered a concurrent push). The old head above is the recovery anchor.

This slice owned the `check_max_lines.py` failure that was masking the frontend gate for four slices, and gained one new commit (`108abc5f7`) for it: the RSS-profile bin (1099) and the collector process suite (1783) are now recorded in `scripts/max_lines_allowlist.txt` with exact counts and named split triggers. That file's policy is the opposite of the frontend one — it accretes documented entries and asks to keep shrinking. Recording exact counts rather than ceilings was verified: bumping an entry to 1800 makes the checker fail with `Stale max-lines allowlist entries`.

pablonyx:
## Dedicated review — `pr2-collector-core` at `108abc5f7`

Scope: 4 commits / 25 files vs `ee9256f9a`. Read-only review, no builds run.

**No production-critical defects.**

`LOW | CONFIRMED | http.rs:424` — the comment overstates what it proves. "The one body-parse permit makes this exact fixed allocation the total request-body heap bound" holds only at `concurrent_body_parsers = 1` (the default), but that limit is configurable via `RuntimeLimits`, and N permits admit N × `MAX_BATCH_BYTES` concurrently. The code is correct — the allocation is semaphore-bounded either way — the comment should read "each body-parse permit".

`LOW | CONFIRMED | state/ingest.rs:231` — `next_order` uses `saturating_add`, so it pins at `u64::MAX`; `state/query.rs:47` filters `stored.cursor > after`, which makes every record after the first sharing that saturated cursor unreachable. Unreachable in practice (~1.8e19 records), noted for completeness rather than as a fix request.

**Verdict.** Body-size accounting uses `checked_add`, the early-413 path rejects oversized declared bodies without polling them, rejection-ledger tallies are per-reason and correctly exclude failed records, and I found no fd or socket leaks, no blocking calls in async contexts, and no panics on untrusted input. The known host-sensitive `ECONNRESET` fixtures are disclosed in the PR body and unchanged here.

pablonyx:
## Dedicated review — pr2-collector-core at 108abc5f7

Scope: `git diff ee9256f9a...108abc5f7`, 4 commits / 25 files (new crate `proliferate-diagnostics-collector`: `auth.rs`, `config.rs`, `http.rs`, `ingest_body.rs`, `process.rs`, `server.rs`, `transport_query.rs`, `state/{mod,ingest,query,health,lifecycle}.rs`, the RSS proof bin, the process suite, plus doc/allowlist edits). Read against the frozen ADR `adrs/2026-08-10-rust-observability.md` and the PR 1 protocol crate. **No builds were run** — no cargo, no rustc, no make, no Docker; every finding is established by reading the code paths end to end.

Overall this is a careful slice: memory-only, no DB/outbox/replay anywhere (verified — the only `std::fs` in the crate is `process.rs`'s `File::from_raw_fd` on the inherited descriptors and the proof harness's evidence output), constant-time capability compare with length binding, `Origin`-bearing requests refused, the raw token zeroed in `CollectorServer::start`/`CollectorConfig::drop`/`Capability::drop` and never placed in the descriptor, log, or URL, tail lag terminating the stream instead of growing a queue, and export frames streamed over `Arc<[u8]>` clones rather than copies. The findings below are real but none of them undo that shape.

---

### HIGH | CONFIRMED | `anyharness/crates/proliferate-diagnostics-collector/src/state/ingest.rs:235-244`

**Once `lifecycle_operations` trackers are all *open*, a bounded-cardinality condition stops being a per-record rejection and becomes a whole-request `500` on `/v1/ingest` and `/v1/export`, and permanently breaks `collector.shutdown`.**

The eviction search only considers closed trackers:

```rust
let removable = inner.lifecycle_order.iter().position(|candidate| {
    inner.lifecycle.get(candidate).is_some_and(|tracker| !tracker.open)
});
let Some(position) = removable else {
    return Err(RejectionReasonV1::LimitExceeded);
};
```

When the record that needs a slot is a *producer-supplied* lifecycle record, that `LimitExceeded` lands correctly in the receipt (`ingest.rs:121-127`). The defect is that the collector's **own** lifecycle records take the same path through `?`:

- `ingest.rs:67` and `ingest.rs:84` — `self.record_producer_attach_locked(&mut inner, ...)?`
- `lifecycle.rs:80-81`, `lifecycle.rs:89-90` — `accept_record_locked(...).map_err(|_| CoreError::Serialization)?`
- `http.rs:108-111` — `ingest_candidates(...).map_err(|_| ApiError::internal())?` → **HTTP 500**
- `http.rs:273-276` — `begin_export().map_err(|_| ApiError::internal())?` → **HTTP 500**
- `server.rs:118` — `self.core.begin_shutdown()?` → `shutdown()` returns `Err`, the graceful-shutdown signal is never sent (only `Drop`'s abort runs), `main` exits non-zero, and no `collector.shutdown` terminal is ever recorded.

Concrete reachability, entirely within this slice's own test harness. `tests/collector_process.rs:962` (`runtime_slots_lifecycle_and_gap_maps_stop_at_their_finite_caps`) configures `lifecycle_operations: 4`, posts five lifecycle starts, and asserts `accepted_count == 4` + one `limit_exceeded`. At that exact instant the table is `{bounded-open-0..3}`, all open (`collector.boot` and `collector-attach-*` were already evicted as closed). The test then immediately calls `mark_producer_dead` to close them. Insert **one** step before that call and:

- posting any record whose `(component, producer_boot_id)` is new → `record_producer_attach_locked` needs a fresh `collector-attach-<uuid>` slot → `LimitExceeded` → `CoreError::Serialization` → **500, the entire valid batch discarded with no receipt, no per-record rejections, and no counter** (`note_request_rejection` is not called on the internal-error path, so `/v1/health` shows nothing);
- `POST /v1/export` → **500, for as long as the table stays full** — a v1 surface is out of service;
- `server.shutdown()` → `Err`, no shutdown lifecycle pair.

At the shipped default this needs 64 lifecycle operations that have a `started` and no terminal. That is not an extreme state once PRs 5/8/9 attach: the P0 catalog puts `anyharness.{tool,mcp,plugin,skill,hook,subagent}.invoke`, `anyharness.stream`, `anyharness.model.request`, `anyharness.{goal,autonomous_loop,review,workflow}.run`, `desktop.workspace.*` and `desktop.target.*` all on this table, and the *only* thing that ever closes an orphaned start is `mark_producer_dead` over the control fd. Starts orphaned by a producer that is still alive accumulate monotonically for the life of the collector boot.

This contradicts the ADR on three points: over-limit values must "fail closed with stable rejection reasons" (here `limit_exceeded` is laundered into `invalid_shape` at HTTP 500), "missing evidence is represented by counters, gaps, and health — not by a fabricated product failure" (a full batch of valid records vanishes with neither), and `collector.shutdown` must have a `started` plus exactly one terminal.

Suggested shape of the fix: make the collector's own lifecycle emissions non-fatal — reserve slots for `collector.*` operations, or have `record_producer_attach_locked` / `begin_export` / `begin_shutdown` degrade to a counted health violation instead of `?`-ing out of the request. Independently, the tracker table needs an eviction rule that can reclaim *open* entries (oldest-open-first, with a synthesized `abandoned` or a recorded gap), otherwise the table is a one-way ratchet.

---

### MEDIUM | CONFIRMED | `anyharness/crates/proliferate-diagnostics-collector/src/http.rs:97-101, 326-354`

**There is no read timeout anywhere; one stalled request body holds the sole body-parse permit forever, wedging all ingest, and eight stall every route including `/v1/health`.**

`ingest` acquires `body_parse_slots` with `acquire_owned().await` (unbounded wait) and then reads the body with `while let Some(chunk) = stream.next().await` — also unbounded. The router carries only the `authorize` middleware; there is no `TimeoutLayer` and no idle reaper (grep confirms the crate's only `tokio::time::timeout` is `server.rs:127`'s shutdown deadline and the proof harness's client-side timeouts).

Failure scenario. A producer opens a connection, sends `POST /v1/ingest` with `Content-Length: 100000`, writes half the body, and then stops writing without closing — a hung serializer thread, a `SIGSTOP`ped process, a paused container. Because it never FINs, the stream never yields `None` or an error.

- With the shipped `concurrent_body_parsers: 1`, **every subsequent ingest** parks on `body_parse_slots.acquire_owned()` indefinitely. Ingest is dead collector-wide, from one stuck peer, with no recovery short of the OS eventually tearing down the TCP connection.
- Each stuck request also holds one of the 8 `concurrent_handlers` permits, which `authorize` acquires *before* dispatch (`auth.rs:93-95`). Eight such requests and every route — `/v1/records`, `/v1/tail`, `/v1/health` — returns `503`. The health surface the ADR relies on to make loss diagnosable is exactly what goes dark.

`/v1/export` shares the same permit (`http.rs:252`), so it wedges too.

A read/idle timeout on the body (and ideally on the whole request) would close this. Note that a *killed* producer is fine — the OS FINs and the stream ends — so this is specifically the hung-peer case.

---

### MEDIUM | CONFIRMED | `anyharness/crates/proliferate-diagnostics-collector/src/state/query.rs:42-61`

**`/v1/records` JSON-decodes every retained record on every request, under the core mutex, and throws almost all of that work away.**

```rust
for stored in &inner.records {
    let accepted = stored.decode().map_err(|_| RejectionReason::Core)?;   // full serde parse
    *versions.entry(stored.version).or_default() += 1;                    // uses stored.version, not accepted
    if record_matches(&accepted, &query.filters) {
        if stored.cursor > after && page_records.len() < usize::from(query.limit) && !page_full { ... }
    }
}
```

The version census reads `stored.version`, which is already cached on `StoredRecord` — the decode is needed only by `record_matches` and the push. Yet it runs unconditionally, including for every record with `cursor <= after_cursor` and every record after the page is full. The normal polling pattern (`after_cursor = next_cursor`, which is what the proof harness's own `query_reader` does) therefore re-parses the *entire* arena to produce at most 500 records.

With the default 8 MiB arena and the proof's ~450-byte encoded records that is roughly 18,000 `serde_json::from_slice` calls per poll, all inside `self.lock()` — the same `Inner` mutex `accept_record_locked` needs. So query cost is charged directly to ingest latency, and it scales with retention rather than with page size. The 1-second responsiveness assertion in the RSS gate is being spent here.

Gating the decode on `stored.cursor > after_cursor && !page_full && page_records.len() < limit` is a one-line change that removes essentially all of it. (`export_snapshot` at `query.rs:159-174` has the same shape but breaks at `record_limit`/`byte_limit`, so it is bounded — no finding there.)

---

### LOW | CONFIRMED | `anyharness/crates/proliferate-diagnostics-collector/src/server.rs:126-138`

**A failing serve task skips `finish_shutdown`, leaving `collector.shutdown` open with no terminal.**

```rust
Ok(result) => {
    result??;                       // <- returns from shutdown()
    TerminalOutcomeV1::Succeeded
}
...
self.core.finish_shutdown(outcome)?;
```

`begin_shutdown()` has already emitted the `started` record by then (`server.rs:118`). If `axum::serve` resolves to `Err(io::Error)`, both `?`s return before `finish_shutdown`, so the operation ends with a `started` and no terminal — the one invariant the ADR pins hardest for the collector's own catalog entry. `finish_shutdown(TerminalOutcomeV1::Failed)` on the error path fixes it. (Reachability of `serve` returning `Err` is low, which is why this is LOW, not the severity of the invariant it breaks.)

---

### LOW | CONFIRMED | `anyharness/crates/proliferate-diagnostics-collector/src/process.rs:100-102`

**One malformed control line terminates the collector without a graceful stop.**

`apply_line` returns `Err(InvalidControlCommand)` for any line serde cannot parse (and `wait_for_shutdown` returns `Err` for any line over `MAX_CONTROL_COMMAND_BYTES`). That propagates out through `wait_for_process_shutdown` to `main.rs:40`'s `?`, so **`main.rs:41`'s `server.shutdown().await?` never runs**: no `collector.shutdown` start/terminal pair, no graceful drain, non-zero exit. `{"command":"producer_dead"}` with the field misspelled is enough. Given PR 3 owns the writer this is a robustness issue rather than an exploit, but "collector crash does not terminate product processes" cuts both ways — an unparseable byte on a control channel should be counted and ignored, not fatal.

---

### LOW | CONFIRMED | `anyharness/crates/proliferate-diagnostics-collector/src/server.rs:172-191`

**Connection slots are consumed before any authentication, with no idle reaper.**

`LimitedListener::accept` takes one of `open_connections` (64) permits *before* `self.listener.accept()`, and the permit lives as long as the `LimitedIo`. Any local process — no capability required, since TCP accept precedes HTTP parsing — can open 64 connections to the loopback port and hold them idle; the listener then accepts nothing further and every producer and every Tauri-mediated query is refused at the transport layer, with nothing anywhere that closes an idle connection. The port is OS-assigned but trivially enumerable locally (`lsof -i`). Loopback-only keeps this LOW under the v1 threat model, but the same missing idle/read deadline that produces the MEDIUM above is what makes this unrecoverable.

---

### LOW | CONFIRMED | `anyharness/crates/proliferate-diagnostics-collector/src/http.rs:342-344` (prior finding 1)

**The comment is wrong, and the prior "comment-accuracy, not a code defect" judgement holds — with a caveat about what the RSS ceiling actually proves.**

```rust
// The one body-parse permit makes this exact fixed allocation the total
// request-body heap bound across ingest and export.
let mut buffered = Vec::with_capacity(limit);
```

`concurrent_body_parsers` is a public `RuntimeLimits` field whose only ceiling is `concurrent_handlers` (`config.rs:57-58`), so "the one body-parse permit" is a statement about the default, not about the type. Worst case at the validated ceiling (`concurrent_body_parsers == concurrent_handlers == 8`), per in-flight request:

- `Vec::with_capacity(MAX_BATCH_BYTES)` = **1 MiB, allocated eagerly regardless of actual body size** — a 200-byte batch still reserves 1 MiB;
- the prepared encodings, capped at `MAX_PREPARED_BATCH_BYTES` = 1 MiB (`ingest_body.rs:111`);
- one live `serde_json::Value` tree per record, up to a 64 KiB record — hundreds of KiB of `Value` nodes at peak.

So ≥ 16 MiB of request-body heap at the configured ceiling versus ~2 MiB at the default, against a 50 MiB RSS ceiling — and the RSS proof only ever exercises `RuntimeLimits::default()` (the child in `ChildCollector::spawn` goes through `CollectorConfig::standalone`). Two fixes worth taking: restate the comment as "at the default `concurrent_body_parsers = 1`", and size the buffer from `Content-Length` when it is present and under the limit rather than always reserving 1 MiB.

---

### LOW | PLAUSIBLE | `anyharness/crates/proliferate-diagnostics-collector/src/state/lifecycle.rs:185-206`

**Collector-synthesized abandonment reuses the producer's `producer_sequence`.**

`mark_producer_dead` clones the retained start record and overwrites only `source_timestamp`, `severity`, `error_classification`, and `lifecycle`. `producer_sequence` is carried through unchanged, so the collector emits a record claiming a `(producer_boot_id, producer_sequence)` pair the producer already used for a different record. `finalizer = collector` is set correctly and `track_producer_sequence_locked` is deliberately not called, so no false gap is recorded — but a consumer reconciling per-producer sequences (which is precisely the mechanism the ADR names for exposing retry and loss) sees a duplicate. The ADR gives sequence ownership to producers and abandonment ownership to the collector; there is no defined encoding for "collector-minted, no producer sequence", so this may be a contract gap for PR 1 rather than a bug here. What would settle it: a fixture pinning the expected `producer_sequence` on a collector-finalized `abandoned` record.

---

### LOW | PLAUSIBLE | RSS proof concurrency headroom — `fixtures/contracts/rust-observability-v1/rss-profile.json` vs `config.rs:33`

The profile runs `4 ingest_writers + 2 query_readers + 2 tail_readers + 1 export_reader` = **9** concurrent clients against `concurrent_handlers: 8`. Streaming responses do not in fact pin a handler permit for their lifetime — `auth.rs:96-100` stashes the permit in `response.extensions_mut()`, and hyper drops the response head (and its extensions) once the head is written, so the permit is released long before a tail/export body finishes. That is what keeps steady-state concurrency low enough to pass, but it also means `concurrent_handlers` does not bound concurrently-streaming responses at all; only `tail_slots`/`export_slots` do.

The residual risk is a flaky release gate rather than a product defect: no client retries a transient `503`, and `export_reader`'s `response.error_for_status()?` (`bin/proliferate-diagnostics-rss-profile.rs:802`) propagates out of the task into `task.await??`, aborting the whole measurement run with an error instead of recording a failure — while the slow tail reader reconnects continuously for 300 s (500 ms delay, 32-frame broadcast buffer, 2000 rec/s guarantees a lag on every pass). What would settle it: instrument concurrent handler entries during one real run, or simply give the proof clients a bounded retry on 429/503 and raise `concurrent_handlers` above the profile's client count.

---

- **`next_order` saturating arithmetic (prior finding 2, `state/ingest.rs:302-303`).** Not a defect. `next_order` is per collector *boot* and the collector is memory-only, so there is no persistence to carry it forward. The first real cliff is not `u64::MAX` at all: `accepted_order`/`retention_cursor` are checked against `MAX_SAFE_INTEGER` (2^53) by `validate_serializable_safe_integers` inside `validate_records_page`/`validate_tail_frame`, so at ~9.0e15 records reads would start returning a contract rejection rather than corrupting order — and at the profile's 2000 rec/s that is ~142,000 years of uninterrupted uptime. Saturation itself would freeze the cursor (queries return empty forever, tail drops every live record) but is unreachable. No change warranted beyond, at most, a comment.
- **Ingest envelope bypasses the frozen raw batch parser.** `parse_ingest_body` deserializes a `BorrowedIngestBatch` and then reconstructs `{"schema_version": …, "records": []}` to feed `parse_ingest_batch_value`, so the frozen parser never sees the real envelope. I chased this as the analogue of the PR 1752 raw-vs-typed divergence and it does not hold: every check `parse_ingest_batch_value` performs at envelope level is separately enforced — `compact_json_len > MAX_BATCH_BYTES` by `read_bounded_body`'s stricter raw-byte cap, `validate_raw_version` by passing the real `schema_version` through, the record-count cap by `BoundedRawRecordsVisitor`, and per-record parsing by `prepare_record`'s full `Value` → `parse_producer_record_value` chain (so record-level `reject_secret_fields` / `reject_prohibited_model_plugin_metadata` are preserved). `IngestBatchV1` has no `deny_unknown_fields`, so unknown top-level keys are ignored in both paths identically, and `reject_secret_fields` is not applied at envelope level by the protocol crate either. `fixtures/contracts/rust-observability-v1/invalid/api.json` contains no `ingest_batch` case, so no fixture reason-code diverges. The one thing that *is* preserved by construction — raw-wire string decoding of ignored fields, so a lone surrogate in an unknown key still rejects — has an explicit test (`ingest_body.rs:362`, `collector_process.rs:598`).
- **`state/health.rs:37` underflow.** `self.limits.tail_readers - self.tail_slots.available_permits()` looked like an unguarded subtraction, but `tail_slots` is constructed as `Semaphore::new(tail_readers)` (`state/mod.rs:148`) and nothing ever calls `add_permits`, so `available_permits() <= tail_readers` always holds.
- **`state/ingest.rs:199` `record.producer_sequence - 1`.** Guarded by the enclosing `record.producer_sequence > previous.saturating_add(1)`, so the value is ≥ 2.
- **Panics reachable from network input.** The only non-test `expect`s outside the proof binary are `ingest.rs:270` (`contains_key`-guarded) and `server.rs:177` (a semaphore that is never closed). Deep nesting does not blow the stack: `enforce_json_node_budget` recurses through the same `serde_json::Deserializer`, so serde_json's own 128-frame recursion limit rejects before our visitor recurses far, and it does so for both the whole-body pass and the per-record pass. Numeric casts (`index as u16`, `accepted_orders.len() as u16`, `snapshot.records.len() as u32`) are all bounded well below their target widths by `MAX_BATCH_RECORDS`/`record_limit`.
- **Tail growing a queue for a slow reader.** It does not. `broadcast::channel(tail_queue_frames)` is fixed at 32 `Arc<[u8]>` frames, `RecvError::Lagged` counts the drop, emits one `Lag` frame with `resume_after_cursor`, sets `done = true`, and the `unfold` stream terminates on the next poll — the reader is disconnected exactly as the ADR requires. Catch-up `pending` never holds more than a gap plus one records frame.
- **Eviction not emitting gaps.** `retain_locked` records counters only, which initially looked like a miss against "emits exact counters and gaps". The gap is synthesized at read time from `evicted_through_cursor` by `eviction_gap` (`query.rs:233-253`) for both `/v1/records` and `/v1/tail`, and separately prepended in `export_snapshot`. Eviction is strictly `pop_front` on a cursor-ordered deque, so oldest-first and the monotonic advance of the oldest cursor both hold regardless of class or severity — and the proof asserts `oldest_cursor == evicted + 1` and `from_cursor/to_cursor/dropped_records` exactly.
- **Capability comparison.** `Capability::matches` pads into a fixed 256-byte buffer and ANDs a `ct_eq` of the bytes with a `ct_eq` of the big-endian lengths, so neither content nor length short-circuits; the stored copy is zeroed on drop, as are `CollectorConfig::capability` (both in `start` and in `Drop`) and `main`'s local. The descriptor written to stdout carries only endpoint, fd number, schema major, and boot id — the proof harness asserts the token appears in neither the descriptor line, stdout, stderr, nor a URL.
- **Any database / disk outbox / replay queue / exactly-once machinery.** None present. The only `std::fs` use in the shipped crate is `File::from_raw_fd` on the two inherited descriptors; everything else is in the proof harness, which writes its evidence outside the repo, and the harness itself asserts the collector's working directory is empty on exit.

pablonyx:
## Post-review fixes + restack — head is now `662d1d33c`

Force-pushed with lease. Two changes since the sealed review head `108abc5f7`, then a rebase onto the repaired #1752:

- **HIGH from the dedicated review, fixed:** when the lifecycle table was full of open operations, `record_producer_attach_locked` propagated `LimitExceeded` through `http.rs` as a 500 and the **whole valid batch** was discarded — a spec violation (per-record rejection is the contract). The saturated case now rejects only the offending record with the pinned reason and the rest of the batch lands. Negative control included.
- `check_max_lines.py` ledger refresh for the collector process suite.

Known flake (pre-existing, reproduced at the pre-fix base 3/4 runs): the oversized-body tests race a client `ConnectionReset` against the 413 response under load.

PR #1756, opened 2026-08-11
feat(anyharness): add durable agent messaging

pablonyx:
## Status

- Open draft; not authorized for merge.
- Exact head: `280b63d3486544bc26e2d3d530c34c98c0f3b6e7`.
- Exact base: `0176215acff423a39693f93546636ad12c944bf2`.
- GitHub exact-head CI status: green.
- Readiness labels: `area:anyharness` applied; required `release:*` remains pending.

## Frozen spec and slice

- Frozen feature revision: `2026-08-10.1`.
- Slice: **PR 4 — Durable cross-agent messaging and provenance**.
- Local-only frozen specification: `/Users/pablohansen/Proliferate Workspace/ADRs/(0) Subagents/Core Docs/5 Consolidated Feature Specification and PR Plan.md`.
- Frozen specification SHA-256: `baea03bdef9dbf0204673a11d38f5c6981a87a0752420e1a848095607f13d6a7`.
- Limitation: the specification path is machine-local and is not a GitHub-accessible artifact.

## Proof run

- The complete focused command, process-backed, and live endpoint inventory is preserved in the detailed record below.
- GitHub exact-head CI and Intent Tests passed.
- Local-only live receipt pinned to this exact head: `/Users/pablohansen/Proliferate Workspace/ADRs/(0) Subagents/Core Docs/11 PR4 Live Acceptance Receipt 2026-08-11.md`; receipt SHA-256 `bd4f4243b1cad0617bb1103e6d1df7e3557b3beff9c4761b99a51da51d8b10f6`; exact binary SHA-256 `2d8b7d0ae03cbc13d115615957f541a645b63d3435aeeab189adab0ad7a01d9d`.

## Proof not run and why

- Automatic Workspace MCP launch injection and the full product UI were not run; activation and UI belong to later slices.
- Automatic terminal completion delivery was not a PR 4 claim; that behavior is implemented and proven in PR 6.

## Open blockers

- No implementation, exact-head CI, or retained live-proof blocker is documented.
- The receipt is local-only and cannot be opened by a remote GitHub reviewer from this PR body.
- Readiness metadata remains intentionally incomplete: the draft has no `release:*` or `area:*` labels.
- This PR is stacked on #1755 and must not merge before it.

## Exact next action

Keep this PR draft. After #1755 is landed and the PR 10 release gate is accepted, reconcile the stacked base, reconfirm exact-head checks, choose and apply the required release label, and merge fourth.

---

## Summary

- add durable, attributed ordinary-agent send_message delivery through the Workspace MCP
- use the existing pending-prompt queue as the acceptance linearization point
- wake idle/running actors best-effort and resume cold targets on their existing native conversation
- preserve one-way ordinary-agent semantics; reverse delivery requires an explicit send_message call

## Verification

- send_message: 17 tests
- pending_prompts: 6 tests
- prompt_provenance: 2 tests
- live-session manager: 9 tests
- provenance-forgery guard: 1 test
- AnyHarness boundary, session-admission, max-lines, design-attribution, targeted rustfmt, and diff checks
- independent review accepted the immutable head
- isolated real Cursor/Workspace-MCP probe passed idle delivery, cold resume, running FIFO, crash/restart exact-once replay, no implicit reverse, and explicit reverse delivery

## Observability

Handled post-commit activation failures emit only a fixed low-cardinality failure code. Paths, auth details, provider output, prompts, and transcript content are not attached to the new warning.

## Documentation

No repository documentation changed. The frozen feature specification already describes this slice. The external acceptance receipt is Core Docs/11 PR4 Live Acceptance Receipt 2026-08-11.md, SHA-256 bd4f4243b1cad0617bb1103e6d1df7e3557b3beff9c4761b99a51da51d8b10f6.

## Deferred

Subagent creation and reversible Close/Open/Promote remain PR5. Completion notifications remain PR6. HTTP/generated APIs, UI, and automatic Workspace MCP activation remain later slices.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/app/agent_operations.rs
+++ b/anyharness/crates/anyharness-lib/src/app/agent_operations.rs
@@ -30,13 +30,19 @@ pub(super) fn wire_agent_operations(deps: AgentOperationsWiringDeps) -> Arc<Agen
             deps.session_runtime.clone(),
         )
         .with_workspace_catalogs(
-            deps.workspace_option_runtime,
+            deps.workspace_option_runtime.clone(),
             deps.session_runtime.clone(),
             deps.session_service.clone(),
         )
         .with_ordinary_operations(
-            deps.session_runtime,
+            deps.session_runtime.clone(),
             deps.session_service,
+            deps.session_admission.clone(),
+            deps.workspace_operation_gate.clone(),
+        )
+        .with_messaging(
+            deps.session_runtime,
+            deps.workspace_option_runtime,
             deps.session_admission,
             deps.workspace_operation_gate,
         ),
--- a/anyharness/crates/anyharness-lib/src/domains/agent_operations/mcp/calls.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agent_operations/mcp/calls.rs
@@ -5,7 +5,7 @@ use serde_json::{json, Value};
 use super::context::WorkspaceMcpContext;
 use crate::domains::agent_operations::model::{
     AgentCreationKind, AgentIdentity, AgentPresentationStatus, ConfigureAgentInput,
-    CreateAgentInput, CreateWorkspaceInput, ListAgentsInput, ListWorkspacesInput,
+    CreateAgentInput, CreateWorkspaceInput, ListAgentsInput, ListWorkspacesInput, SendMessageInput,
     WorkspaceIdentity,
 };
 use crate::domains::agent_operations::runtime::{AgentOperations, AgentOperationsError};
@@ -63,6 +63,13 @@ struct ConfigureAgentArgs {
     agent_id: String,
     config_id: String,
     value: String,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+struct SendMessageArgs {
+    agent_id: String,
+    message: String,
 }
 
 #[derive(Debug, Deserialize)]
@@ -220,6 +227,23 @@ pub async fn call_tool(
             let args = parse::<TargetArgs>(arguments)?;
             let target = AgentIdentity::new(operations.runtime_identity().clone(), args.agent_id);
             serialize(operations.resume_agent(&ctx.caller, &target).await)
+        }
+        "send_message" => {
+            let args = parse::<SendMessageArgs>(arguments)?;
+            serialize(
+                operations
+                    .send_message(
+                        &ctx.caller,
+                        SendMessageInput {
+                            target: AgentIdentity::new(
+                                operations.runtime_identity().clone(),
+                                args.agent_id,
+                            ),
+                            message: args.message,
+                        },
+                    )
+                    .await,
+            )
         }
         "interrupt_agent" => {
             let args = parse::<TargetArgs>(arguments)?;
--- a/anyharness/crates/anyharness-lib/src/domains/agent_operations/mcp/tests.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agent_operations/mcp/tests.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use async_trait::async_trait;
 use serde_json::json;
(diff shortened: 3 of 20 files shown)


pablonyx:
## Review fixes applied (pushed `089a856c7`, unbuilt — CI verifies)

### 1. MEDIUM — send_message no longer blocks on target activation ✅ fixed
`SessionRuntime::enqueue_agent_message` now detaches consumer activation onto a `tokio::spawn` **after** the durable pending-row commit, so the MCP `send_message` tool returns as soon as the receipt is durable instead of blocking up to `SHARED_STARTUP_READINESS_TIMEOUT` (60s in prod) while a cold target boots.

- `domains/sessions/runtime/prompt.rs` — durable insert stays synchronous; activation is spawned post-commit.
- **Lease invariant (documented in code):** the spawned task runs **without** the caller's target-workspace shared `SessionPrompt` lease. That lease is acquired in `runtime/messaging.rs::send_message` and dropped when `send_message` returns (right after `enqueue_agent_message`). This is deliberate and mirrors the startup-replay path, which activates consumers with no caller lease — the pending row is already durable, so no lease is needed to guarantee eventual delivery, and re-acquiring one would reintroduce the block. `warn` logging in the detached task is preserved.
- Because the receiver must outlive the call, the `AgentMessageQueue::enqueue_agent_message` trait + impl receiver changed `&self` → `self: Arc<Self>` (`runtime/ports.rs`), the call site clones the Arc (`runtime/messaging.rs`), and the two test doubles were updated (`runtime/messaging_tests.rs`, `mcp/tests.rs`). async-trait 0.1.91 supports `self: Arc<Self>`.
- Test contract updated: `prompt_message_tests.rs` — the old `..._commits_before_pending_startup_and_bounds_the_readiness_wait` asserted the inline-blocking behavior; replaced with `send_message_returns_before_target_startup_readiness_resolves`, which is a negative control (500ms bound < the 1s test readiness timeout an inline activation would block on). The real-actor / cold-start tests already poll (`wait_for_prompt_count`, `wait_for_actor_idle`) so they remain valid unchanged.

### 2. LOW-MEDIUM — envelope now carries a reply handle ✅ fixed
`domains/sessions/prompt/render.rs::agent_message_context` appends: `To reply, use send_message with agentId "<session_id>".` — matching the actual tool name `send_message` and its `agentId` arg (verified in `agent_operations/mcp/tools.rs`, required args `["agentId","message"]`). The `agentId` value reuses the already JSON-escaped `serde_json::Value::String(source_session_id)`, so escaping is preserved. The replay-stable provenance test `send_message_prompt_provenance_renders_replay_stable_sender_context` was updated; no other test asserts the context string.

### 3. VERIFY — orphaned-pending-row race on close ⚠️ real but pre-existing; NOT fixed (no trivially-safe guard)
**Verdict:** The race is real, but only against the **MCP `close_subagent`** path — not the HTTP close.

- send_message serializes via two gates on the **target session id**: the `SessionMutationAdmission` keyed permit (`SessionMutationKind::Prompt`, held across the closed-check → durable-insert window) plus the shared `SessionPrompt` workspace lease.
- **HTTP close** (`api/http/sessions_lifecycle.rs::close_session`) acquires `admit_session_mutation(SessionMutationKind::Close)` on the same session id, so it serializes on the keyed admission mutex against send_message's `Prompt` permit → **no orphan** (close either precedes the closed-check and rejects with `SessionClosed`, or follows the released permit after a legitimately-committed row).
- **MCP `close_subagent`** (`domains/sessions/subagents/mcp/calls.rs`) calls `session_runtime.close_live_session` directly. That path (`runtime/lifecycle.rs`) takes **no** admission permit and **no** workspace lease — only `access_gate.assert_can_mutate_for_session` (a permission check, not a mutex). So it does **not** serialize against send_message, and an interleave `[closed-check sees open] → [close marks closed] → [insert pending row]` leaves a durable pending row on a now-closed session (receipt says `DurablyQueued`, never delivered nor rejected). The leak is benign at runtime (startup replay won't activate a closed session), but the receipt is misleading.

This window **predates this PR and is not widened by the change** — the closed-check and durable insert remain synchronous under the `Prompt` permit; only post-commit activation is detached. I did **not** add a guard: the correct fix is to have `close_subagent` acquire the `SessionMutationKind::Close` admission permit (mirroring the HTTP lifecycle path), which changes lock-acquisition ordering against the workflow-controller policy (run gate → permit → lease) and belongs in the subagent-close / admission slice, not a drive-by guard in durable messaging.

### Flagged, not fixed (per instructions)
Envelope reply/context text is rendered at prompt-render time (`agent_message_context`) rather than stored as canonical text. This contradicts the ADR's settled "one canonical text" ruling. Deterministic replay mitigates (render is a pure function of stored provenance + body, so replay is stable — asserted by the provenance test). Left as-is: the reply-clause here is copy-level, and the stored-body-vs-rendered-envelope decision is a pending founder call.

PR #1757, opened 2026-08-11
feat(desktop): supervise diagnostics collector

pablonyx:
> [!IMPORTANT]
> This PR intentionally remains **DRAFT**. Final source SHA `3f814c79f45491c8ba7d976b5a2867cccbcdd108` is static-clean, its exact four-package plus both-target package/CLI matrix is terminal green at [31517619468](https://github.com/proliferate-ai/proliferate/actions/runs/31517619468), and the clean independent review verdict is **ACCEPT** with no blockers. Marking ready and merging remain reserved for the human gate.

## Summary

Stacks frozen Rust Observability PR 3 on `codex/rust-observability-pr2-collector-core`. Desktop now packages and solely supervises the accepted PR 2 collector, gates owned producer startup on bounded authenticated readiness, mediates records/tail/export/health through a same-user Tauri broker and the existing `proliferate-debug` CLI, routes in-scope Tauri detail through bounded collector-primary/fallback paths, and performs ordered bounded shutdown/reaping.

## What changed

- exact fourth Desktop external binary with supported-release fail-closed staging and two-target inventory/signing/architecture checks;
- singleton Tauri collector owner with protected per-launch capability/control channels, bounded descriptor/readiness, health monitoring, restart budget, and generation leases;
- owner-only pathname `AF_UNIX` discovery/lock/locator with mutual same-UID/PID/app-boot verification, exact framing/caps/deadlines, and conservative stale cleanup;
- broker-mediated `proliferate-debug records|tail|export|health`, with no direct endpoint/capability flags or files;
- compile-time customer/internal artifact gate plus Tauri-contextualized health/export-health;
- 1 MiB/256-record native queue with lifecycle reservation and detail-first eviction, plus 1 MiB/four-segment owner-only fallback and no disk replay;
- exact eight Tauri process-owner lifecycle families around collector, AnyHarness, and Worker;
- main-window renderer-ingest seam with raw closed-shape/ownership/current-generation/collector-boot checks, and native protected child-handoff/generation seam for later slices;
- idempotent shutdown arm → broker cancellation → 500 ms producer drain → Worker → AnyHarness → collector → locator/fallback teardown;
- current Desktop-native/Desktop-host/observability/testing/release owner documentation.

## Validation

Final source: `3f814c79f45491c8ba7d976b5a2867cccbcdd108`  
Binary diff SHA-256 from accepted PR 2: `c30e54083e43fc02f33b69419b48c6d76c0356e08e84c85d7f453a80bdda230a`

Static proof passes:

- direct rustfmt over all touched Rust;
- `git diff --check`;
- `python3 scripts/check_docs.py` (`211 Markdown files`);
- `python3 scripts/check_design_attribution.py`;
- T4 JavaScript syntax, Tauri JSON, external-binary inventory, and release workflow actionlint;
- every PR 3 file under 600 lines (only two unchanged accepted-PR2 files remain over the repository threshold).

Final source-pinned workflow: [run 31517619468](https://github.com/proliferate-ai/proliferate/actions/runs/31517619468), `contents: read`, no secrets/deploy/dependency cache, immutable checkout, jobs=2, and a no-Cargo/no-rustc/pressure gate.

Exact-head Rust: **PASS** — 196 passed / 0 failed / 3 intentionally ignored subprocess fixtures; pressure 1; log SHA-256 `114d20a65a96d9479999cfad8a954c1960ca689127b7bd95355b52ba586f53f2` ([job 93866602439](https://github.com/proliferate-ai/proliferate/actions/runs/31517619468/job/93866602439)).

Exact-head package/CLI:

- **PASS** `aarch64-apple-darwin`: exact four real sidecars, signed inventory and architecture, packaged broker CLI health/records/tail/export-gate; log SHA-256 `726affd8ac893d2ce75f9c669cf388f2332321e09e38ac0a8825227398283872` ([job 93866602746](https://github.com/proliferate-ai/proliferate/actions/runs/31517619468/job/93866602746)).
- **PASS** `x86_64-apple-darwin`: equivalent exact inventory/signature/architecture and packaged CLI proof; log SHA-256 `0e4c23f46fa2339c2cb16935bfa37aec4b586afd15df3bb92d80e6875a441e4e` ([job 93866602612](https://github.com/proliferate-ai/proliferate/actions/runs/31517619468/job/93866602612)).

Accepted review finding **PR3-R2-01** is resolved at the final head: release profile alone no longer activates sidecar resolution from a dependency build, preventing nested Cargo target-lock deadlock. Only `CARGO_PRIMARY_PACKAGE` or all four explicit binary inputs activate real staging; the all-explicit Tauri package path remains fail-closed and is proven on both targets above.

Independent final review: **ACCEPT**, no blocking findings; artifact SHA-256 `a3c0759c9d20572e2462f17fcc0461ca66637a42f8d9ede8f8fcce5058f0af48`. Non-blocking later notes: PR3-FU-01 (decisions mutex across bounded restart-burst delays), PR3-FU-02 (pinned string-prefix stop classification), and PR3-FU-03 (foreign-UID rejection proven at the exact predicate seam).

No local Cargo/rustc command was started because the mandatory local memory gate reached pressure level 2. No Docker command was run.

## Observability

The delta is local Tauri collector packaging, supervision, query mediation, bounded producer/fallback, and the eight frozen process-owner lifecycle pairs. Sentry, PostHog, anonymous telemetry, server logging, renderer logging, AnyHarness logging, Worker logging, and support consent/upload remain behaviorally independent and unchanged.

- **RV-2-04:** records keeps producer minors 1.0/1.1; tail remains exact 1.1 cursor-only.
- **RV-2-05:** Tauri sends only typed shutdown under 4,096 bytes; malformed/oversized/control-write failure remains child failure, never graceful shutdown.
- **RV-2-06:** source-derived worst-case terminal retry window is 40.3 seconds. Accepted PR 2 caps the whole lifecycle tracker map at 64; PR 3 does not resize it.

## Scope

No PR 1 protocol or PR 2 collector internals changed. No renderer producer, AnyHarness/Worker producer adapter, support migration, exporter/OTLP, application/semantic lifecycle, dashboards, monitors, agent tooling, persistent store, or new platform support. No UI, so no screenshot is required.

## Readiness

- [ ] Founder review and final exact-head receipts accepted. This remains unchecked while the PR is draft.

## Verification provenance

- Frozen PR 3 revision 2 SHA-256: `14d2999cd75a9b4b18bfedd10d1237e2eb83f3106fc277c392fa735dbe6c3a2b`
- Accepted PR 2 review SHA-256: `9248b54fdb93139494cfbc27b64edad43dd24f4014936e07f723592deb1948d5`
- PR 2 → PR 3 feedback SHA-256: `f20dab1d647fc94a5098ee45c9157adec06cd0d4c57498afcce92b371e08b46d`
- Roadmap provenance-only correction SHA-256: `578f4ce4bbf36449da428c970ceec23706a9585a65d38596d8ec9e356f882b3f`


pablonyx:

`codex/rust-observability-pr3-tauri-supervision`: `3f814c79f` → `63e912970` (force-pushed with `--force-with-lease` pinned to the old SHA, so the update would have aborted rather than clobbered a concurrent push). The old head above is the recovery anchor.

pablonyx:
## Dedicated review — `pr3-tauri-supervision` at `63e912970`

Scope: 10 commits / 63 files vs `108abc5f7`. Read-only review, no builds run.

### The one high-severity candidate does not survive verification

**REJECTED | `broker/discovery.rs:527` `remove_owned_socket`** — the claim was a TOCTOU allowing arbitrary file deletion: validate with `symlink_metadata`, then have an attacker swap in a symlink before `remove_file` follows it. **False.** `fs::remove_file` is `unlink(2)`, which removes the link itself and never follows it, so a swapped-in symlink is the thing unlinked, not its target. The residual race (our socket replaced by another object) is additionally fenced by the `uid() == effective_uid()` and `mode & 0o777 == 0o600` guards, and the parent is the `0700` per-user temp directory.

### Real findings, all low

`LOW | CONFIRMED | supervisor/lifecycle.rs:156` — the restart budget (3 per 60s) fully resets after `HEALTHY_BUDGET_RESET_AFTER`, and the backoff ladder is fixed (250ms / 1s / 4s) rather than exponential. A collector that reports ready, runs 61s and then dies re-arms the budget every cycle, so it restarts indefinitely at roughly 66s intervals with no ceiling. Late-manifesting failures (a slow leak, an OOM at steady state) are exactly the shape that hits this.

`LOW | CONFIRMED | process/helpers.rs:105` — `validate_binary` scans only the first 512 bytes for `PLACEHOLDER_MARKER`, so a placeholder longer than that is accepted and then fails as a cryptic exec error instead of the clear diagnostic this check exists to produce.

`LOW | CONFIRMED | process/owned.rs:175` — `Drop` discards the result of `start_kill()`. Nothing else reaps, so a child in an unexpected state during unwind can be left behind.

`LOW | note | broker/discovery.rs:548 darwin_user_temp_dir` — the `unsafe` `confstr` sizing reads correctly to me (size query, exact allocation, NUL trim), but it is `#[cfg(target_os = "macos")]` and therefore never compiled by CI. One instance of the coverage hole under `## NEEDS PABLO` item 4 in the run journal.

**Verdict.** The supervision structure is sound: kill-on-drop is present, generations are fenced, the health loop and state publication are coherent. No blocking defects.

pablonyx:
## Dedicated review — pr3-tauri-supervision at 63e912970

Scope: `git diff 108abc5f7...63e912970`, 10 commits / 63 files. **No builds were run** (no cargo/check/test/rustc/docker); every claim below is derived by reading code in the worktree, plus the vendored `tokio-1.50.0` and `tauri-utils-2.9.3` sources in `~/.cargo/registry` where third-party semantics were load-bearing.

Coverage: I read the full supervisor (`supervisor.rs`, `supervisor/{initialization,lifecycle,queries}.rs`), process launch/teardown (`process.rs`, `process/{helpers,owned}.rs`), the collector HTTP client, the producer + fallback, the shutdown coordinator, the whole broker (`discovery.rs`, `discovery/{profile,security}.rs`, `server.rs`, `server/{framing,export_stream}.rs`, `protocol.rs`, `client.rs`), the renderer command surface, `lib.rs`/`telemetry.rs`/`sidecar.rs`/`cloud_worker*` deltas, the build/packaging path, and the release workflow. I did **not** read the colocated test files in depth (`*_tests.rs`, ~2,400 lines) beyond checking what they pin, and I did not audit the `_unsupported.rs` Windows stubs line by line beyond the shutdown-relevant paths.

---

### Findings

**MEDIUM | CONFIRMED | `apps/desktop/src-tauri/src/diagnostics_collector/shutdown.rs:119-201`**
`DiagnosticsShutdownCoordinator::shutdown` caches its first result forever and returns it to every later caller without re-attempting teardown, which (a) makes a diagnostics-only failure decide a product operation's result and (b) destroys the documented Windows updater retry contract.

Chain, every link read:
- `shutdown()` opens with `let mut completed = self.result.lock().await; if let Some(result) = completed.as_ref() { return result.clone(); }` and stores `Some(result)` at the end (line 199) regardless of success.
- `failed = true` is set by six phases, four of which are pure observability: `broker.wait_stopped()` (line 138), `supervisor.stop_collector()` (line 173), `broker.remove_locator_and_unlock()` (line 184), and `self.fallback.close()` (line 189).
- `commands/cloud_worker.rs::prepare_desktop_dispatch_worker_update` now routes straight into `shutdown.shutdown().await` on Windows (previously it called `lifecycle::prepare_desktop_dispatch_worker_update`, i.e. only `arm_terminal_shutdown_and_stop_worker`).
- `apps/desktop/src/lib/access/tauri/updater.ts:77-78` is `await prepareDesktopDispatchWorkerUpdate(); await u.install();` — a rejection skips `install()`.

Concrete failure: on Windows, the user downloads an update; `shutdown()` runs and `stop_process` (cloud_worker/lifecycle.rs:227-251) hits its 2 s reap timeout, returning `Timed out stopping Proliferate Worker`. `shutdown()` caches `Err("diagnostics_shutdown_failed")`. The install is blocked. Every retry now returns the *cached* Err instantly, performing no teardown at all — so the Worker is never re-attempted, even though `finish_stop_attempt`'s own doc comment (lifecycle.rs:255-260) says "Both credential rotation and a Windows updater retry must still see the child after a failed `try_wait` or `kill`". The user then quits the app; `RunEvent::Exit` calls `shutdown()`, gets the cached Err, and stops nothing — the tracked Worker survives holding its credential database lock into the next launch, which is exactly the leak `stop_tracked_desktop_dispatch_worker`'s doc comment and `specs/desktop-native.md` say must not happen. A milder variant of the same latch: `fallback.close()` failing on a flush/sync error permanently blocks Windows update installation, which contradicts the ADR's "Observability never changes a product operation's result" and "An explicit diagnostics API request … returns its own failure honestly, without changing the underlying product operation."

Suggested shape of a fix: cache only `Ok(())` (or separate "product teardown result" from "diagnostics teardown result" and let only the former reach the updater).

---

**MEDIUM | CONFIRMED | `apps/desktop/src-tauri/src/telemetry.rs:113-131`, `diagnostics_collector/producer.rs:398-447`, `diagnostics_collector/fallback.rs:93-114`**
Tauri-native `tracing` output stops reaching `desktop-native.log` once the collector is ready, but the debug/support bundle still collects that file and has no replacement source until PR 6 — so on a healthy install the exported bundle contains essentially no native logs.

- `telemetry.rs` removed the `create_file_log_sink` layer entirely and replaced it with `native_diagnostics.make_writer()`; `create_file_log_sink` now has no production caller (only `telemetry_file_logging.rs` tests and `commands/diagnostics.rs` tests).
- `producer.rs::enqueue` writes to the fallback only when `state.client.is_none()` (line 399, 445-447).
- `supervisor::accept_ready` → `producer.set_ready_client(gen, Some(client))` → `fallback.set_active(state.client.is_none())` = `false` (producer.rs:202), and `FallbackDiagnosticsWriter::record` early-returns `Ok(())` when `!inner.active` (fallback.rs:112-114).
- `diagnostics/bundle.rs:19-24` still lists `logs_dir.join("desktop-native.log")` in `desktop_log_base_paths`.

Concrete failure: a customer on a working install runs "Export debug bundle" or the support flow. The bundle's `desktop-native.log` contains only the handful of lines emitted before the collector reached Ready, because everything after that lives only in the collector's memory-only arena (ADR: "The collector is memory-only", "no collector database, disk outbox"). The consented support snapshot that would restore this is PR 6.

I want to be fair: `specs/desktop-native.md:361-380` and `specs/OBSERVABILITY.md:44-52` are updated in this PR to describe `desktop-native.log` as a bootstrap/outage fallback, so this is a deliberate design choice, not an accident. I am still raising it because the ADR frozen text does not require moving Tauri detail off disk, PR 3 lands the loss while PR 6 lands the replacement, and nothing in the diff narrows or annotates the existing bundle path. Worth a founder call on whether the gap is acceptable across slices 3→6.

---

**LOW | CONFIRMED | `apps/desktop/src-tauri/src/diagnostics_collector/broker/server.rs:196-199`**
`accept_loop` treats any `listener.accept()` error as terminal, permanently disabling the query broker for the process lifetime with no health signal.

```rust
accepted = listener.accept() => {
    let Ok((stream, _)) = accepted else {
        break;
    };
```

Concrete failure: the app hits transient fd pressure and `accept(2)` returns `EMFILE` (or a peer aborts between the SYN and accept, `ECONNABORTED`). The loop breaks, drains sessions, and the task ends; the `UnixListener` drops. The locator file and socket path stay published until app exit, and `DiagnosticsCollectorSupervisor` state stays `Ready`, so `proliferate-debug health/records/tail/export` reports `collector_unavailable` for the rest of the session with nothing indicating the broker itself died. Observability-only — no product impact — but the standard shape is to `continue` on transient accept errors.

---

**LOW | CONFIRMED | `supervisor/initialization.rs:135-171` + `supervisor/lifecycle.rs:151-212`, called from `commands/cloud_worker.rs` and `commands/runtime.rs`**
A hanging (not missing, not invalid) collector binary can delay two product commands by roughly 25 s.

`ensure_desktop_dispatch_worker` and `restart_runtime` both `await diagnostics_supervisor.wait_startup_barrier()`, which is only resolved by `start()`. On the retryable path the worst case is the initial launch deadline (`COLLECTOR_LAUNCH_TIMEOUT` = 5 s) plus `restart_burst_locked`'s three attempts of `sleep(delay) + launch`: 0.25+5, 1+5, 4+5 → ~25.25 s before `resolve_startup(Degraded)` releases the barrier.

The barrier is genuinely bounded and `specs/FEATURE_DOCS/DESKTOP_HOST.md` documents "A degraded observability path releases product startup", and the ADR does require the collector to start before AnyHarness/Worker — so the ordering is correct. Flagging the magnitude only: 25 s of a renderer-visible "Restart runtime" or worker-ensure spinner caused purely by a sick collector is close to the ADR's "observability never changes a product operation's result" line. The non-hanging failures (`BinaryMissing`, `BinaryInvalid`, `UnsupportedTarget`, including the debug placeholder) all resolve the barrier immediately, so this only bites when the collector process starts and then stalls.

---

**REJECTED — `broker/discovery.rs:526-539` `remove_owned_socket` TOCTOU arbitrary file deletion.** I re-derived this independently and I agree with the earlier rejection; the finding is false.
- `std::fs::remove_file` on Unix is `unlink(2)`, which never resolves a symlink in the final path component — it removes the link itself. The claimed attack (swap the validated socket for a symlink between `symlink_metadata` and `remove_file`, so the removal follows it) cannot happen: the attacker would only delete their own symlink.
- The residual race — swapping in a different *real* file — does not cross a trust boundary. The path is `darwin_user_temp_dir()/pd/<scope_id>.sock` (discovery.rs:270-272), and `ensure_namespace` → `create_or_validate_directory(socket_root, 0o700)` → `validate_directory` requires `is_dir() && uid == geteuid() && mode & 0o777 == 0o700` on `symlink_metadata`, so a symlinked `pd` is rejected outright and only a same-uid process can create entries inside it. `_CS_DARWIN_USER_TEMP_DIR` is itself a per-user 0700 directory. Same-uid is the trust boundary the whole broker already assumes (`peer_uid_is_owner`, security.rs:11-17).
- Callers are additionally serialized under `try_lock_exclusive` on `app_dir/run/diagnostics-broker-v1.lock` (discovery.rs:87-95), and `cleanup()` re-reads and identity-checks the locator (`app_boot_id`, `pid`, path) before removing anything.

**REJECTED — broker peer-pid check is dead on macOS.** `peer_matches_locator` (security.rs:15-17) requires `pid.and_then(u32::try_from) == Some(expected_pid)`, and older tokio returned `pid: None` on macOS, which would have made every `client.open()` return `PeerUnauthorized` and every `probe_old_broker` return `Ambiguous`. Not the case here: `tokio-1.50.0/src/net/unix/ucred.rs:213-255` (`impl_macos`) uses `getsockopt(SOL_LOCAL, LOCAL_PEEREPID)` and returns `pid: Some(...)`. Cargo.lock pins tokio 1.50.0. Works as intended.

**REJECTED — raw capability leak.** I traced the whole custody chain and found nothing to raise. `generate_capability` (process/helpers.rs:114-130) moves the encoded `String` into `Zeroizing` with no surviving copy and zeroes the raw 32 bytes; the token reaches the child only through the `protected_pair()` socketpair whose fd is un-CLOEXEC'd in `pre_exec`; the descriptor read back carries only `endpoint`, a `token_reference` that must equal the fd number, `schema_major`, and `collector_boot_id`, and `read_descriptor` rejects any other reference; `SecretCapability`/`CollectorHttpClient`/`OwnedCollectorProcess`/`ProtectedChildHandoff` all have hand-written `Debug` impls that omit it; `authorization()` sets `HeaderValue::set_sensitive(true)`; the endpoint is never in a URL query. Renderer exposure is one command, `ingest_renderer_diagnostics`, gated on `label == "main"`, returning only `IngestReceiptV1` — and I checked `IngestReceiptV1` and `HealthResponseV1` in `proliferate-diagnostics-protocol/src/v1/types.rs:353-362,534-558`: neither has an endpoint or token field. Defence in depth is present too (`fallback.rs:200-244` redacts `capability`/`authorization`/`token` keys, a 43-char base64url pattern, and `http://127.0.0.1:<port>`), and `producer.detailed` sanitizes before enqueue so even the in-collector copy is scrubbed.

**REJECTED — records-query injection into the collector URL.** `encode_records_query` (client.rs:359-407) builds every value through `url::form_urlencoded::Serializer`, so `#`, `/`, `&` in a broker-supplied filter are percent-encoded before `Url::join`; the `.expect("fixed collector route must be valid")` cannot fire on that input.

**REJECTED — restart storm.** `accept_restart_budget_at` (supervisor.rs:215-229) caps automatic restarts at `AUTOMATIC_RESTART_DELAYS.len()` = 3 per rolling 60 s window, and the budget is only cleared after 60 s of continuous `Ready` (`reset_restart_budget_after_health`). `monitor()` returns after calling `fail_generation`, and a new monitor is spawned only on a successful restart, so exactly one monitor exists per generation and there is no self-feeding loop.

**REJECTED — orphaned collector child on a shutdown/restart race.** Both `start()` (initialization.rs:137-145) and `restart_burst_locked` (lifecycle.rs:179-187) stash a successfully launched process into `inner.process` before bailing on `shutdown_armed`, and `stop_collector` — which blocks on the same `decisions` async mutex — subsequently takes and reaps it. `RetainedLaunchChild`/`OwnedCollectorProcess` both `start_kill()` on `Drop`, and the launcher sets `kill_on_drop(true)`, so tokio's orphan queue reaps anything that escapes.

**REJECTED — std `Mutex` held across an `.await`.** I checked every `self.inner.lock()` / `self.inner.state.lock()` site in the supervisor and producer; all are either scoped blocks or closure-bounded, and the async work (`launch`, `orderly_shutdown`, `terminate_and_reap`, `client.ingest`) always happens after the guard drops.

**REJECTED — `Contents/MacOS/Proliferate` → `proliferate` rename in `tests/release/upgrade/run-t4-desktop.mjs`.** I suspected the new case-sensitive `readdirSync` inventory check (and the matching `LC_ALL=C sort` check in `release-desktop.yml`) would fail against a `Proliferate`-named main binary that the old `existsSync` check only passed because macOS is case-insensitive. `tauri-utils-2.9.3/src/config.rs:3595-3609` documents `mainBinaryName` as overriding the cargo output binary and it is unset here, so the binary is the crate name `proliferate`. The rename is a fix, not a break.

**DROPPED as too shaky — self-amplifying log loop.** All native `tracing` now feeds `TauriDiagnosticsProducer`, and the pump's `client.ingest` uses reqwest/hyper; if those emitted tracing events under `RUST_LOG=debug` each ingest would generate new detailed records and re-trigger the pump, never idling. `env_filter_from_env` defaults to `info`, and `Cargo.lock` shows no `tracing` dependency on `reqwest`, so I could not establish that the events exist. Not reporting it; what would settle it is checking whether hyper's `tracing` feature is enabled in the resolved feature set.

---

I found no HIGH. Token custody, the same-user broker's ownership/permission/symlink handling, and the "collector crash does not terminate product processes" rule all hold up under the reading I did. The two MEDIUMs are both about the shutdown/telemetry seam rather than the collector itself, and the shutdown-result latch is the one I would want fixed before merge.

pablonyx:
## Post-review fixes + restack — head is now `74ce3da3d`

Force-pushed with lease. Since the sealed review head `63e912970`:

- `100347dc1` — diagnostics teardown failures are kept off the product shutdown result (the Windows updater gates `install()` on that result; a fallback flush failure is pure observability and must never block it). Two new shutdown tests pin both directions.
- `74ce3da3d` — `check_max_lines.py` ledger entry for `broker/server.rs` (672 lines, split trigger documented): the gate is red on this branch without it.

**Merge-order note for the train:** a build containing this PR but not #1818 keeps `desktop-native.log` collection in the support bundle while the renderer path migrates — that window is closed by #1818's consented-support surface. Land in stack order.

PR #1763, opened 2026-08-11
fix(product): scope workspace creation receipts to their session

tairabun:
## Linear tickets

- [PRO-123 — Workspace created shows up for every single session in worktree even when it isn't the one that 'created' worktree](https://linear.app/proliferate-team/issue/PRO-123/workspace-created-shows-up-for-every-single-session-in-worktree-even)

## Summary

- Scope the synthetic workspace-creation transcript receipt to the ProductClient session alias that owned the creation flow.
- Preserve the workspace-wide arrival event for setup lifecycle consumers while preventing sibling session tabs from receiving the receipt.
- Keep the server session list authoritative after materialization or reload, using earliest `createdAt` with the existing ID tie-break.

## Testing

- Tier 1 regression coverage uses the real QueryClient, AnyHarness runtime/workspace providers, production session directory/selection stores, production session query key, and production first-session ordering logic.
- `fnm exec --using=v22.21.1 -- pnpm --filter @proliferate/product-client build`
- `fnm exec --using=v22.21.1 -- pnpm --filter @proliferate/product-client typecheck`
- `fnm exec --using=v22.21.1 -- pnpm --filter @proliferate/product-client exec vitest run --exclude src/app/authenticated-markdown-cascade.test.ts` (804 files, 5,456 tests)

## Observability

- None. This is a deterministic projection/ownership correction with no new runtime failure mode or external dependency.

## Readiness

- [x] The support relationship is linked through the tracker API. No tracker, report, user, or support IDs or private source data appear in this PR.

## Verification

- Before the fix, the new session-switch regression failed because the sibling session received the workspace receipt key.
- After the fix, the same regression passes across both a populated server-session cache and the projected-to-materialized identity handoff.
- The repository-wide ProductClient test command reaches 804 passing files and 5,456 passing tests; its remaining browser fixture cannot launch because this machine does not have the hard-coded Google Chrome application path. The full non-browser lane above passes.

### DIFF
--- a/apps/packages/product-client/src/hooks/workspaces/derived/use-workspace-creation-receipt.test.tsx
+++ b/apps/packages/product-client/src/hooks/workspaces/derived/use-workspace-creation-receipt.test.tsx
@@ -1,0 +1,236 @@
+// @vitest-environment jsdom
+
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { act, cleanup, renderHook } from "@testing-library/react";
+import type { ReactNode } from "react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { Session, Workspace } from "@anyharness/sdk";
+import {
+  AnyHarnessRuntime,
+  AnyHarnessWorkspace,
+  anyHarnessSessionsKey,
+} from "@anyharness/sdk-react";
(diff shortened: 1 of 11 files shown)


tairabun:
@greptile review

PR #1768, opened 2026-08-11
feat(desktop): add structured renderer diagnostics

pablonyx:
## Summary

- Add a typed ProductClient renderer-diagnostics port and a Tauri-only Desktop producer with bounded filtering, batching, backpressure, loss accounting, and acknowledged error reporting.
- Migrate the frozen renderer startup, measurement, terminal, runtime, auth, support, automation, transcript, workspace, and session diagnostic rows without changing their product-control behavior.
- Route only validated pre-dispatch native unavailability through the existing fallback while preserving the original stable error and preventing post-dispatch duplicates.
- Delete the obsolete renderer-specific TypeScript/native writer paths and Desktop-only dependencies, while retaining historical support-bundle discovery.
- Update current-state Desktop host, native, frontend, and observability documentation.

This is a draft stacked on #1757 (`codex/rust-observability-pr3-tauri-supervision`).

## Testing

- `CARGO_BUILD_JOBS=2 cargo test -p proliferate renderer_ingest` — 12 passed, 0 failed; this was the only Rust invocation and was preceded by separate cargo/rustc and memory-pressure gates.
- Focused ProductClient diagnostics/callsite suites — 16 files, 84 tests passed.
- Focused Desktop filter/batcher/error/install/native-error suites — 11 files, 85 tests passed.
- Focused Desktop wrapper/bridge/boot/measurement plus unchanged client/Sentry/anonymous suites — 9 files, 66 tests passed.
- `pnpm --filter @proliferate/product-client build`
- `pnpm --filter @proliferate/product-client typecheck`
- `pnpm --filter proliferate build`
- `pnpm --filter proliferate exec tsc`
- `python3 scripts/check_mobile_product_client_export.py --preflight`
- `git diff --check`

All commands above passed. Every PR-owned file remains under 600 lines; the max-lines checker reports only the two accepted predecessor baseline files.

A manual Desktop trace was not run because that development path compiles the Tauri shell and would violate this slice's one-Rust-invocation contract. The native filtered lane, fake-native end-to-end tests, exact receipt/loss tests, and production builds provide the bounded proof instead.

## Observability

- Renderer diagnostics now use the accepted structured collector envelope rather than separate ad hoc native commands/files.
- Detailed records remain local and fail-silent; Sentry, product analytics, anonymous telemetry, and support-flow behavior retain their existing owners and gates.
- Pre-dispatch fallback is deliberately narrow and never changes the JavaScript acknowledgement contract or retries a record.
- Current-state observability and Desktop ownership docs are updated in this PR.

## Readiness

- [ ] I followed the [pull-request procedure](https://github.com/proliferate-ai/proliferate/blob/main/guides/process/pull-requests.md) for scope, title, labels, body, and readiness.

This remains intentionally unchecked while the PR is a reviewable draft.

- [x] No support relationship or private support data is included in this PR.

## Verification

- Frozen PR 4 specification: revision 9, SHA-256 `f061b5bd5bd4f7497ebbb52c53d93dca9e74980aa7acc945702c146528f813fd`.
- Accepted base: `3f814c79f45491c8ba7d976b5a2867cccbcdd108`.
- Implementation head: `d741d8d1dc445c49e00a097eee658e877c307b0f`.
- Legacy renderer API inventory is empty; historical `renderer-diagnostics.log*` support discovery remains.


pablonyx:

`codex/rust-observability-pr4-renderer-diagnostics`: `d741d8d1d` → `15c386107` (force-pushed with `--force-with-lease` pinned to the old SHA, so the update would have aborted rather than clobbered a concurrent push). The old head above is the recovery anchor.

pablonyx:
## Dedicated review — `pr4-renderer-diagnostics` at `15c386107`

Scope: 2 commits / 107 files vs `63e912970`. Read-only review, no builds run.

**No high-severity findings.**

`MEDIUM | PLAUSIBLE | hooks/agents/lifecycle/use-local-auth-state-sync.ts:128` — the effect chains work onto `operationQueueRef.current` without returning a cleanup, so queued operations keep running after unmount and can apply state captured against a previous `runtimeUrl`. This may be deliberate serialization (letting an in-flight operation finish with its original parameters rather than tearing it half-done), which is why it is PLAUSIBLE rather than CONFIRMED — distinguishing the two needs a run, and I ran nothing.

**Verdict.** Diagnostics buffers are bounded (queues and the loss ledger cap at safe integers), secret filtering covers both key names and token-shaped values, and the lifecycle hooks I read clean up their timers and subscriptions (xterm surface, debug session activity, upload queue). I found no vacuous assertions and no unredacted capture path.

pablonyx:
## Dedicated review — pr4-renderer-diagnostics at 15c386107

Scope: `git diff 63e912970...15c386107` — 2 commits, 107 files (delta over PR #1757 only). Read-only review; **no builds, no tests, no cargo/pnpm run** — every claim below is from reading code. I read the whole renderer filter/queue/batcher chain, the Tauri command + supervisor ingest path, the fallback writer, all 21 changed ProductClient/desktop call sites, and `adrs/2026-08-10-rust-observability.md`.

**Bottom line: I found no HIGH or MEDIUM defect.** The four things the brief singled out — boundedness, the main-window-only handoff, the pre/post-dispatch fallback asymmetry, and privacy — all check out on inspection, and I have written up below exactly what I verified so the next reviewer does not re-derive it. What is left is three LOW items and one doc nit.

---

### Findings

**LOW | CONFIRMED | `apps/desktop/src-tauri/src/commands/diagnostics.rs:36-42` (with `diagnostics_collector/supervisor.rs:241-254`)**
"Exact schema v1.1" for renderer batches is enforced only in the renderer, not at the native trust boundary. `parse_renderer_ingest_value` calls `parse_ingest_batch_value`, which validates via `validate_schema_version` (`anyharness/crates/proliferate-diagnostics-protocol/src/v1/validation_support.rs:137-145`) — that accepts the whole compatible producer-minor window, i.e. `1.0` as well as `1.1`. Neither `validate_renderer_ingest_shape` nor `validate_renderer_ingest_batch` compares against `{major:1,minor:1}`. The exactness check lives at `apps/desktop/src/lib/access/tauri/diagnostics.ts:52-64` — in the renderer, which is precisely the party `require_main_window` exists to distrust.
Failure scenario: renderer JS in the main window calls `invoke("ingest_renderer_diagnostics", {batch})` with `schema_version:{major:1,minor:0}` on both batch and records; the native side admits it and forwards it to the collector, contradicting `specs/desktop-native.md`'s "Exact schema-v1.1 Desktop renderer batches". Impact is small (a 1.0 record is a legal producer record the collector must accept anyway) — this is a claim/enforcement mismatch, not a memory-safety or privacy issue. Fix is two lines next to the existing `component`/`source` equality check.

**LOW | CONFIRMED | `apps/desktop/src-tauri/src/diagnostics_collector/fallback.rs:114-125`**
`record_pre_dispatch_renderer` deliberately skips the `inner.active` check that `record` (line 108-110) uses. That is correct for the inactive case (the spec says renderer records may be retained "without activating it"), but the same check is also what protects `record` after `close()`. After `DiagnosticsShutdown`'s `CloseArtifacts` phase (`diagnostics_collector/shutdown.rs:189`) has taken `inner.file`, a renderer batch arriving on the still-live WebView hits `write_record` → `self.file.as_mut().ok_or_else(|| io::Error::other("fallback is closed"))` → `record_locked` sets `inner.degraded = true` and increments `dropped_records`.
Failure scenario: user quits; shutdown is armed, the collector is stopped, the fallback is closed, and the renderer flushes one last batch on `pagehide` (`renderer-diagnostics.ts:250-252` schedules exactly that). `FallbackHealthV1.state` flips from `Inactive` to `Degraded` and `dropped_records` inflates, reporting a failure that did not occur. Nothing samples health that late today, so impact is nil in this slice — but PR 6's consented snapshot and PR 3's health surface both read `fallback.health()`, so this is a latent false-degradation source. The command result is unaffected (`record_pre_dispatch_renderer_batch` discards with `let _`), so the spec's "returns the original unavailable result" contract still holds.

**LOW | PLAUSIBLE | `apps/desktop/src-tauri/src/diagnostics_collector/supervisor/queries.rs:208-214` with `fallback.rs:264-292`**
The pre-dispatch fallback does blocking file I/O inline in the `async fn ingest_renderer` Tauri command, with no `spawn_blocking`. Per record: `write_all` + `write_all(b"\n")` + `flush()`, plus `file.metadata()` and then `retained_bytes()`'s four `fs::metadata` calls — roughly six syscalls and a flush per record, ×64 records per batch, plus 3 renames and a `sync_data` whenever a 256 KiB segment rotates. That parks a Tokio worker thread for the duration.
Failure scenario: collector is `Starting` (or restarting) while the renderer emits at its 50 ms flush cadence; every batch parks a worker on disk I/O. I rate this PLAUSIBLE rather than CONFIRMED because I did not measure it, and because the identical pattern already exists on the native producer path (`diagnostics_collector/producer.rs:446`) — so it is not introduced by this slice. What is new is the volume: renderer detail is far higher-rate than native `tracing`. Settled by timing `ingest_renderer` under a `Starting` supervisor with a full 64-record batch.

**NIT | `apps/packages/product-client/src/components/app/AppErrorBoundary.tsx:30`**
The `onRenderError` doc comment still reads "Resolves true only after the host confirms diagnostic **persistence**." This slice correctly rewrote the matching contract on `DesktopDiagnosticsBridge.reportRenderError` (`host/desktop-bridge.ts:352-357`) to say it proves collector admission and explicitly "does not claim persistence, fallback retention, upload, or current collector state". The consumer-side comment is now the stale half of that pair.

**NIT | `renderer-diagnostics.ts:234-238` and `renderer-error-diagnostics.ts:241-245`**
Both `reset*ForTest` helpers clear the install guards (`lifecycleFlushInstalled`, `listenersInstalled`) without removing the listeners those guards protect (`visibilitychange`/`pagehide`, `error`/`unhandledrejection`). Repeated reset+install inside one module realm accumulates listeners. Test-hygiene only — production installs exactly once and the listeners are app-lifetime by design.

---

Each of these looked like a defect and is not. Reasons recorded so they are not re-raised.

- **Post-dispatch failure falling back (the asymmetry the brief flagged).** Verified correct. `ingest_renderer` (`queries.rs:177-195`) writes the fallback *only* on `Starting | Unsupported | Degraded | Stopped | ShuttingDown` from `ready_generation()`, and then returns that same error unchanged. `dispatch_renderer_batch` (197-206) — receipt validation, generation check, transport — touches the fallback on no path. The non-unix `supervisor_unsupported.rs:193-206` twin does the same and only pre-dispatch. It is not accidental: `supervisor_renderer_ingest_tests.rs` has `renderer_ingest_post_dispatch_unavailable_never_writes_fallback` and `renderer_ingest_ready_coherent_success_writes_no_fallback` pinning both directions.
- **Non-main-window or foreign-JS reach into the handoff.** `require_main_window(window.label())` (`commands/diagnostics.rs:175-179`) rejects with `renderer_ingest_wrong_window`. This matters concretely — a second window really exists (`apps/desktop/src-tauri/capabilities/support.json` scopes to `"support"`), and app-defined `generate_handler!` commands are not gated by Tauri capabilities, so this runtime check is the actual guard and it is present. Renderer JS in the main window *can* reach the command by design (the ADR's "narrow Tauri-owned operations"); it cannot reach the collector endpoint or capability token, which stay in `ProtectedChildHandoff`.
- **A record reaching the handoff without `rejectSecretFields`.** Cannot happen, on either side. Every renderer record goes through `parseProducerRecordV1` inside `validateAndBoundRecord` (`renderer-diagnostic-filter.ts:175`), and again through `parseIngestBatchV1` → `parseProducerRecordV1` in `ingestRendererDiagnosticsBatch`; `validation-records.ts:56` calls `rejectSecretFields`. Natively, `parse_ingest_batch_value` → `parse_producer_record_value` → `reject_secret_fields` (`validation.rs:24`/`:132`). Note `rejectSecretFields` only inspects *JSON keys*, so it structurally cannot see an argument whose `name` is `authorization` — that case is covered separately and correctly by `isSecretKey(name)` at `renderer-diagnostic-filter.ts:74` and by the same check on nested object keys at `renderer-diagnostic-normalize.ts:108`. So the TS/Rust divergence noted on PR #1752 does not open a hole on this path.
- **Unbounded renderer growth.** Bounded on every axis I could find. Queue: 256 records / 512 KiB hard (`renderer-diagnostics-batcher-limits.ts:9-10`), with a 224 / 448 KiB ordinary sub-lane so low-severity floods cannot starve warn/error headroom. Detached in-flight records still count against both budgets via `pendingEntries()` (`renderer-diagnostics-queue.ts:148-152`), so a batch cannot be double-spent. Per-record: `MAX_RECORD_BYTES` is enforced by the contract parser, and `NormalizationState` caps nodes (4096) and value bytes (65536) per record. A failed invoke *drops* its batch rather than requeueing (`renderer-diagnostics-batcher.ts:301-319`), so an outage drains rather than accumulates. `RendererLossLedger` saturates at `Number.MAX_SAFE_INTEGER`. Error-side maps are capped at 128/256 fingerprints. The priority-eviction loop cannot spin: `queue.ts:79-81` returns when `entries` is empty. `validateAndBoundRecord`'s `while(true)` pops one argument per iteration and returns `null` at zero, so it terminates in ≤33 iterations.
- **Privacy widening via renderer stack traces into `desktop-native.log`.** Not a widening. `git show 63e912970:apps/desktop/src/lib/integrations/telemetry/native-diagnostics.ts` shows the pre-slice path already sent `message` + `stack` + `componentStack` to a native file (`renderer-diagnostics.log`), and both that file and `desktop-native.log` are already collected by `diagnostics/bundle.rs:21-22`. The new path is strictly *more* scrubbed: renderer-side regex filtering in `renderer-diagnostic-text.ts` (bearer/basic, `*_TOKEN=`, query-string credentials, URL userinfo, PEM private keys, `ghp_`/`AKIA`/`sk-`/JWT) that the old path did not have, plus Rust-side `sanitize_fallback_record` (`fallback.rs:208-237`) on the way to disk. No new file and no new support-bundle surface.
- **Session permission payloads leaking through `renderer.measurement.busy_holdouts`.** I chased this because `use-debug-session-activity.ts:61` passes the whole holdout array through as one `sensitive` field, and `executionSummary.pendingInteractions[].payload` is a `PermissionInteractionContext` (commands, paths). It does not get out: `MAX_ARGUMENT_DEPTH` is 4 and `normalizeValue` starts field values at depth 1, so `holdouts`(1) → item(2) → `executionSummary`(3) → `pendingInteractions`(4) hits `depth >= MAX_ARGUMENT_DEPTH` and becomes `"[truncated]"`. Only the scalars `phase`/`updatedAt`/`hasLiveHandle` survive. It is also behind an opt-in flag (`isSessionActivityDebugLoggingEnabled`).
- **Boot metadata bypassing the existing sanitizer.** `recordBootRendererDiagnostic` gets raw `metadata` while the local event buffer gets `sanitizeBootMetadata(metadata)` — but `sanitizeBootMetadata` (`boot-stall-diagnostics-format.ts:69-108`) only bounds sizes and flattens `Error`; it removes no secrets. The renderer path's own filter is strictly stronger (a non-plain object such as an `Error` becomes `"[object]"`). Also `recordBootDiagnostic` is `import.meta.env.DEV`-only (`boot-stall-diagnostics.ts:177`).
- **High-frequency emitters on the product hot path.** Every `emit` costs ~3 JSON serializations synchronously (`JSON.stringify` for bytes, `jsonByteLength`, `canonicalClone`), so I checked the new call sites for per-frame/per-keystroke emitters. `logLatency` (`debug-latency.ts:42`), `startLayoutShiftObserver` (`debug-layout-shift.ts:27`), the measurement summary/budget emitters, the transcript-virtualizer report and `busy_holdouts` are all behind `import.meta.env.DEV` and/or `VITE_PROLIFERATE_DEBUG_LATENCY`/localStorage flags. Production emission is milestone- and failure-shaped.
- **Lifecycle expansion (the ADR's hard line for this slice).** None. `buildRendererProducerRecord` hard-codes `record_class: "detailed"` and never populates `lifecycle` (`renderer-diagnostic-filter.ts:149-158`); the `RendererDiagnosticKind` vocabulary is `log|message|milestone|progress|transport` plus the internal `loss_summary`. Every emitted name is `renderer.*`; nothing in this slice touches the P0 catalog's `desktop.application.boot`, `desktop.authentication.*`, `desktop.workspace.*`, `desktop.prompt.submit` or `desktop.update.*` families owned by PR 8. No work is pulled forward.
- **Observability changing a product result.** `reportRenderError` resolves into `AppErrorBoundary.componentDidCatch`'s `void Promise.resolve(...).then(...)` (`AppErrorBoundary.tsx:175-186`); the up-to-500 ms acknowledgement only flips a status label, never gates the recovery UI. `recordRendererDiagnostic` swallows everything at the port (`renderer-diagnostics-port.ts:79-85`), and `emit` swallows filter failures into a loss counter.
- **Effect-cleanup / stale-closure regressions.** The three effects whose dep arrays this slice touched (`ProductLifecycleRoot.tsx:222`, `use-desktop-runtime-bootstrap-lifecycle.ts:58`, `use-support-modal-state.ts:70`) all *shrank* their deps by dropping the now-unused `diagnostics` identity, which reduces re-runs; the `AbortController` cleanup in the runtime-bootstrap effect is untouched. No hook adds a listener, timer or subscription without cleanup. No dead references to the removed `logEvent`/`logRendererEvent`/`logRendererDiagnostic` remain anywhere outside `dist/`.
- **`renderer-diagnostics-batcher-test-support.ts` shipping in the app bundle.** Imported only by three `.test.ts` files; tree-shaken.

#### The carried-over prior finding — REFUTED, and out of scope

> MEDIUM/PLAUSIBLE at `hooks/agents/lifecycle/use-local-auth-state-sync.ts:128` — the effect chains onto `operationQueueRef.current` without returning a cleanup, so queued operations keep running after unmount and can apply state captured against a previous `runtimeUrl`.

Settled by reading, on two independent grounds.

1. **Not this slice's code.** `git diff 63e912970...15c386107 -- <that file>` touches only the diagnostics import, three `recordAgentAuthFailure(...)` calls inside existing `catch` blocks, and the new helper function. The `operationQueueRef` serialization, the effect body and the dep array are byte-identical to the base.
2. **The stale-`runtimeUrl` mechanism does not exist.** The chain is strict FIFO — each render appends with `operationQueueRef.current = operationQueueRef.current.then(...)` — so a later render's closure (newer `runtimeUrl`) always runs *after* an earlier one, never before. There is no ordering in which an older `runtimeUrl` wins. The serialization is also deliberate and documented at lines 182-185: aborting a superseded fetch does not stop the local runtime from processing it, so independent PUT/DELETE calls must not interleave during gateway → native → API-key transitions. Adding a cleanup that cancelled the chain would reintroduce exactly the stale-persisted-route bug the comment describes.

Verdict: intentional, correct, and untouched here.

pablonyx:
## Post-review fix + restack — head is now `67f113f30`

Force-pushed with lease. The dedicated review found no HIGH/MEDIUM; one LOW was worth closing rather than recording:

- `specs/desktop-native.md` promises **exact schema-v1.1** renderer batches, but the only check enforcing exactness lived in the renderer — the party `require_main_window` exists to distrust. Natively, `parse_ingest_batch_value` accepts the whole compatible producer-minor window, so main-window JS could invoke `ingest_renderer_diagnostics` with `1.0` and have it forwarded. `validate_renderer_ingest_batch` now pins envelope and every record to `CURRENT_SCHEMA_VERSION` next to the existing component/source equality. Impact was small (a 1.0 record is a legal producer record), but claim and enforcement should not disagree at a trust boundary. Negative control: dropping the comparisons fails the new test; restored, 13/13 `renderer_ingest` tests green.

The review's carried-over MEDIUM at `use-local-auth-state-sync.ts:128` was refuted on two grounds (not this slice's diff; the FIFO chain makes the stale-`runtimeUrl` ordering impossible) — details in the review comment above.

PR #1774, opened 2026-08-12
feat(desktop): add bundled child diagnostics adapters

pablonyx:
## Summary

- add one bounded shared Rust diagnostics adapter for Desktop-owned AnyHarness and Desktop Worker processes
- introduce protected child activation, generation replacement, status/flush control, terminal fencing, and typed producer-death reconciliation
- migrate bundled detail to the collector with a secure bounded component fallback while preserving standalone/cloud behavior
- replace Worker file output with a bounded in-memory failure tail
- expose typed child status and a strict fallback reader for the next consented-support slice

This is PR 5 in the observability stack. It is stacked on draft PR #1768 / `codex/rust-observability-pr4-renderer-diagnostics` and remains a draft.

Frozen specification SHA-256: `06c55cc25308d048584529fe8009de569e10f77395f8ff7b2261079203bc889b`

Base: `d741d8d1dc445c49e00a097eee658e877c307b0f`

Head: `ea42cdc7b0c52aacc021887071715ce5bb8d6be6`

## Testing

The frozen UTC multi-package lane reached and passed all PR 5-owned suites before stopping at one unchanged PR 2 collector-process fixture:

- AnyHarness CLI: 30 passed
- anyharness-lib: 1,535 passed, 1 ignored
- Desktop native: 255 passed, 3 ignored
- Desktop debug: 2 passed
- shared diagnostics client: 185 passed
- collector library/RSS: 12 passed
- collector process: 13 passed, 1 inherited macOS reset race

The failure is the unchanged early-413 fixture `far_oversized_and_chunked_bodies_are_json_accounted_without_unbounded_reads`: the collector correctly rejects an oversized declared body without polling it, while reqwest can still be writing and receive `ECONNRESET`. PR 2's validation ledger records the same host-sensitive failure. A second unchanged +1-byte boundary case has the same mechanism. I did not reroll either test until lucky or change collector production code.

Supplementary direct receipts:

- remaining collector process cases: 12/12 passed after excluding the two named unchanged reset-prone fixtures
- Worker: 76/76 passed
- final shared-client replay: 185/185 passed
- bounded oversized-body unit proofs passed

Static checks passed: shared-client all-target Clippy with `-D warnings`, rustfmt, diff check, documentation (211), design attribution, AnyHarness boundaries/old paths, and the 600-line cap for every PR 5 file. The repository-wide max-line checker reports only the two unchanged accepted PR 2 baseline files.

Exact-head independent review: **ACCEPT** at `ea42cdc7b0c52aacc021887071715ce5bb8d6be6`; sealed review SHA-256 `b0a444148c622d226490970fa843d75f7a2cc5f81d6f4a2acd87a35aa7bdb89a`. GitHub Actions run `31587530659` completed with Cargo and every PR 5-relevant check green. Exactly two red checks reproduce unchanged on the PR 4 base: the accepted PR 2 max-line files and the existing Shared frontend `ChatInput.test.tsx` provider fixture.

## Observability and privacy

- tracing admission is non-blocking and bounded
- record, batch, receipt, fallback, loss, Worker-tail, and shutdown paths have fixed record/byte/deadline limits
- secrets are denied before correlation promotion and across typed/labeled/boundary-crossing string forms
- capability material never enters argv, environment, config, logs, debug output, or fallback records
- Sentry and console behavior remain independent; standalone AnyHarness and cloud/Supervisor Worker modes remain unchanged

## Founder disposition

Founder accepted exact head `ea42cdc7b0c52aacc021887071715ce5bb8d6be6` as the PR 6 implementation base. IC-006B is accepted as the mechanically honest module-scoped serialization plus best-effort exclusion of unrelated concurrent macOS spawns; a repository-wide spawn/`posix_spawn` redesign remains outside this slice. The completed exact-head review/CI package, its two base-reproduced failures, and the disclosed inherited collector-process reset races are accepted as predecessor validation debt. This PR remains a distinct stacked draft and is not merged or marked ready.

## Readiness

- [x] Implementation is committed and independently audited
- [x] PR 5-owned tests and static checks are green
- [x] Inherited red tests are documented without weakening them
- [x] Founder accepted IC-006B and the inherited-validation package
- [x] Review the stacked PR against the frozen specification
- [ ] Mark ready only after predecessor PRs and review gates are accepted


pablonyx:

`codex/rust-observability-pr5-bundled-adapters`: `ea42cdc7b` → `7b975caf1` (force-pushed with `--force-with-lease` pinned to the old SHA, so the update would have aborted rather than clobbered a concurrent push). The old head above is the recovery anchor.

PR 6 is now open as #1818 and based on this branch.

pablonyx:
## Dedicated review — `pr5-bundled-adapters` at `7b975caf1`

Scope: 22 commits / 110 files vs `15c386107`. Read-only review, no builds run.

### HIGH | CONFIRMED | `fallback/unix.rs:500` `atomic_swap` — fallback log rotation cannot succeed on any non-macOS Unix, and failing it permanently disables the fallback writer

Every link in this chain is verified by reading the code on this branch:

1. `fallback/mod.rs:3` gates the implementation on `#[cfg(unix)]`, **not** on macOS. Linux compiles it.
2. `atomic_swap` calls `renameatx_np(..., RENAME_SWAP)` under `#[cfg(target_os = "macos")]`. The `#[cfg(not(target_os = "macos"))]` arm is `let _ = (directory, left, right); -1` — unconditional failure. Linux's equivalent, `renameat2` with `RENAME_EXCHANGE`, is not implemented.
3. A failed swap surfaces as `FallbackError::UnsafeEntry`.
4. `write_encoded` (`unix.rs:126-134`) sets `self.active = false` on `UnsafeEntry`, and `unix.rs:117` checks `!self.active` first, so **every subsequent write is refused for the lifetime of the process.**

**Failure scenario.** On Linux the fallback writer works normally until segment 0 reaches `FALLBACK_SEGMENT_BYTES` (512 KiB). The first rotation then fails, the writer deactivates permanently, and every later last-resort diagnostic is silently dropped. This is the sink that exists precisely for when the normal diagnostics path is already broken, so it fails exactly when it is needed.

**Why CI is green on it.** Every rotation test in `fallback/tests.rs` is `#[cfg(target_os = "macos")]` — `:335`, `:377`, and the whole rotation-hook harness at `:33-58`. The Linux runner compiles the broken arm and never exercises it. This is the macOS-gating coverage hole (journal `## NEEDS PABLO` item 4) producing a real shipped defect rather than a hypothetical one.

**I did not fix it, deliberately.** A Linux-only `renameat2` path cannot be verified or negative-controlled on this macOS machine without a cross build, `pr6` sits on top of `pr5` so the change also forces another restack of the slice above, and the brief forbids claiming work I could not verify. Fix shape for whoever can run it: implement `libc::renameat2` with `RENAME_EXCHANGE` for `target_os = "linux"`, keep the `-1` arm for remaining targets, and widen the rotation tests from `cfg(target_os = "macos")` to `cfg(unix)` so the runner proves it.

### Other findings

`MEDIUM | CONFIRMED | commands/cloud_worker/launcher.rs:650` — `find_scanned_worker_launcher` falls back to `which::which(WORKER_BINARY)` after the bundled and development locations, so `$PATH` decides what executes. `canonical_current_target_executable` proves the image is a native binary of the right architecture, but not that it is ours. Reachable when the bundled binary is missing (development, or a packaging regression).

`MEDIUM | note | child_bridge/` — the module and its four test files are gated on macOS plus aarch64/x86_64, so roughly 5k lines including descriptor remapping, native-image validation and bootstrap framing never compile on CI at all.

`LOW | PLAUSIBLE | process/owned.rs:200` — `write_terminal_control_line` uses a single `write()` and treats a short write as ambiguous, disabling terminal control for that generation, while `write_orderly_shutdown_line` uses `write_all`. The commit message states partial writes are deliberately never retried, so this is likely intended fail-fast; flagging the asymmetry rather than asserting a bug.

### Commit coupling

The terminal-control and worker-launcher commits read as unrelated to diagnostics at first glance, but they are prerequisites: the control seam is how producer-dead is signalled, and executable validation has to happen before diagnostics descriptors are attached. `edd4afcaf` is explicitly "Repair the PR 5 launcher-module hardening findings". No re-cut proposed — that is a scope call, not a review finding.

### Correction to this PR's description

The body predates tonight's restack and now describes a branch state that no longer exists. Specifically:

- It seals an independent review at **`ea42cdc7b0c52aacc021887071715ce5bb8d6be6`**; the head is now `7b975caf1` (restacked onto `origin/main`, then the slices below were amended).
- It discloses two expected red checks — "the accepted PR 2 max-line files and the existing Shared frontend `ChatInput.test.tsx` provider fixture". **Both are now green.** The two oversized Rust files were added to `scripts/max_lines_allowlist.txt` at recorded exact counts on `pr2` (`108abc5f7`), and the `ChatInput` failure was a stale-base artifact fixed on main by `e93d98636`.
- Its characterisation of the shape violations as "the two unchanged accepted PR 2 baseline files" was only ever true of the max-lines check. The frontend-structure violations started at `pr1` and were split, not allowlisted.

Leaving the body itself intact rather than rewriting it, since it is a record of what was accepted at that head.

**Verdict.** One genuine high-severity defect, in Linux fallback rotation, invisible to CI by construction. The terminal-control state machine itself reads correctly — one-shot reservation plus `Drop`-based cleanup does prevent both duplicate and missed commands. Everything else is hardening.

pablonyx:
## The HIGH from the review pass is fixed: Linux fallback rotation now works, and CI proves it

The earlier review on this PR reported that component fallback rotation could
never succeed on any non-macOS Unix, and that failing it permanently retires the
fallback writer. That is fixed in `2752e3164`.

**What was wrong.** `fallback/mod.rs:3` gates the implementation on `cfg(unix)`,
but `atomic_swap` implemented the exchange with `renameatx_np(RENAME_SWAP)` under
`cfg(target_os = "macos")` and left the `not(macos)` arm as a literal `-1`. That
returns `UnsafeEntry`, and `write_encoded` (`unix.rs:126-134`) clears `self.active`
on `UnsafeEntry`, which `:117` checks first. So on Linux the writer worked until
segment 0 reached 512 KiB and then refused every later write for the process
lifetime — in the last-resort sink that exists for when the normal diagnostics
path is already broken.

**The fix.** Linux spells the same operation `renameat2(RENAME_EXCHANGE)`. It goes
through `syscall(2)` rather than the libc wrapper, so the binary carries no glibc
2.28 floor and builds unchanged against musl. The primitive now lives in
`fallback/swap.rs`; `unix.rs` drops from 596 to 582 lines, which incidentally buys
back headroom on a file that was 4 lines from the 600-line cap. Platforms with no
atomic exchange keep failing closed rather than emulating the swap with plain
renames, which would produce exactly the torn state the identity checks exist to
catch — that residual case is now commented instead of silent.

**Why CI never caught it, and why it will now.** Every rotation test in
`fallback/tests.rs` was `#[cfg(target_os = "macos")]`, so the Linux runner
compiled the dead arm and ran nothing through it. The module is already
`cfg(unix)`, so those gates are removed and `cargo test --workspace` on
ubuntu-latest now exercises the exchange. One test stays macOS-gated — the async
`stalled_rotation_...` case — for its sub-750ms wall-clock assertion rather than
for the primitive, and it now carries a comment saying so instead of leaving the
next reader to guess.

**Verification.**

- macOS: `cargo test -p proliferate-diagnostics-client` → 185 passed, 0 failed,
  both rotation tests included.
- Negative control: forcing the exchange to fail — the exact state Linux was in —
  fails both rotation tests at `tests.rs:335` and `:419`; they pass again on
  restore. So the ungated tests genuinely detect a dead exchange, which is what
  makes the Linux CI run meaningful rather than decorative.
- The Linux arm compiles for `x86_64-unknown-linux-musl` and
  `aarch64-unknown-linux-musl`, checked verbatim out-of-tree because this host
  cannot compile it. Its own negative control is instructive: a typo in that arm
  produces **zero errors on a macOS build** and `E0425` on a Linux target. That
  asymmetry is how the dead arm shipped in the first place.
- `check_max_lines.py` clean, rustfmt clean, `Cargo.lock` version churn excluded
  (pre-existing repo-wide, main's own lock is staler).

`#1818` was restacked onto this head. Its tree changed by exactly this commit and
nothing else — verified by diffing the two ranges, not by assuming.

**One thing to note.** This was `NEEDS PABLO` item 4 — the macOS-gating coverage
hole — producing a real shipped defect rather than a hypothetical one. This commit
closes the hole for the rotation path specifically. The broader item stands: the
desktop lib suite runs 531 tests on macOS and 426 on CI, and the missing 105 are
the platform-specific hardening this train adds. Widening a cfg gate is only
available where the code is genuinely portable, as it was here; the descriptor-
passing and anchor-walk paths are macOS-only by nature and need a macOS job.

pablonyx:
## Dedicated review — pr5-bundled-adapters at 2752e3164

The earlier review on this PR was written at the superseded head `7b975caf1`. This is the first review at the current head `2752e3164`.

**Scope.** `git diff 15c386107...2752e3164` — 23 commits, 111 files, +22,893 / -482, base is PR #1768. Read-only review; **no builds were run** (no cargo, cargo check, cargo test, rustc, make, or Docker). Everything below was verified by reading source, including the vendored `libc-0.2.189` tree for syscall numbers and FFI signatures. Deep personal pass on `fallback/{swap,unix,mod,wire}.rs`, `producer/{fallback_runtime,status,record}.rs`, `bridge/activation.rs` gating, and the two host adapters (`anyharness/src/telemetry.rs`, `proliferate-worker/src/logging.rs`). Three delegated area sweeps (activation semantics and non-interference; bounded queues / reconnect / drop policy / loss accounting; non-bundled mode preservation and the Desktop launcher wiring) returned no confirmed defects. I can only personally vouch for the chains I read myself, which are the ones cited below.

---

## Verdict on the Linux `renameat2(RENAME_EXCHANGE)` rotation fix

**The fix is real and correct.** I checked every element the previous CONFIRMED HIGH turned on.

`fallback/swap.rs:29-39` — the Linux arm:

- **Syscall number.** `libc::SYS_renameat2` resolves per architecture from the libc tree, and the values are the kernel's: x86_64 `316` (`gnu/b64/x86_64/not_x32.rs:345`, `musl/b64/x86_64/mod.rs:489`), aarch64 `276` (`gnu/b64/aarch64/mod.rs:902`, `musl/b64/aarch64/mod.rs:563`). Both gnu and musl trees define it, so the musl-safety claim holds. It is typed `c_long`, matching `syscall(num: c_long, ...)` (`linux/mod.rs:3643`).
- **Flag constant.** `libc::RENAME_EXCHANGE` is `c_uint = 2` in `linux_like/linux/mod.rs:1217`, shared by gnu and musl, and matches the kernel's `RENAME_EXCHANGE`. Not the Android or Fuchsia definition, which are separate and unreachable here.
- **Argument order and types.** `renameat2(olddirfd, oldpath, newdirfd, newpath, flags)`; the call passes `directory, left.as_ptr(), directory, right.as_ptr(), RENAME_EXCHANGE` in exactly that order. Casts to `c_long` are correct for a variadic call on both 32-bit and 64-bit; sign-extension of the `i32` fd is harmless because the kernel reads the low 32 bits.
- **dirfd / AT_FDCWD.** Both dirfd slots take the validated O_CLOEXEC directory handle, never `AT_FDCWD`, and both paths are the fixed relative segment names. This is the safe form and is correct.
- **Return / errno interpretation.** `syscall(2)` returns 0 on success and -1 with errno set (glibc and musl both), so `result == 0` is the right success test. `renameat2` has no non-zero success value.
- **Error paths.** Failure returns `false`, `atomic_swap` (`fallback/unix.rs:501-507`) maps it to `UnsafeEntry`, `swap_verified` re-verifies both entry identities after the exchange and un-swaps if exactly one moved, and `rotate` rolls the completed swaps back in reverse. Fail-closed, no torn state left visible to a live writer.
- **macOS arm survived the extraction.** `renameatx_np(fromfd: c_int, from, tofd: c_int, to, flags: c_uint) -> c_int` (`apple/mod.rs:4762`) with `RENAME_SWAP: c_uint = 0x2`. The call is byte-identical to the pre-extraction version; only the `CString` temporaries moved to the caller, where they still live to the end of the full call statement, so there is no dangling pointer.

**The test re-gating is the more valuable half of this commit.** `fallback/tests.rs` previously gated `rotation_is_oldest_first_...` and `rotation_races_...` on `cfg(target_os = "macos")` while the module itself is `cfg(all(test, unix))`. CI runs `cargo test --workspace` on `ubuntu-latest` (`.github/workflows/ci.yml:237,253`), so the Linux runner compiled a dead arm and executed zero rotation tests. Both are now unqualified and will run there. `rotation_is_oldest_first_...:361` asserts `sequences.windows(2).all(|pair| pair[1] == pair[0] + 1)` across all four segments read oldest-first, which a no-op or misordered exchange cannot satisfy. Only the async wall-clock deadline test stays macOS-gated, correctly, and now says why.

I independently re-derived the rotation algebra: swaps `[(2,3), (1,2), (0,1)]` map `s0..s3 = A,B,C,D` to `D,A,B,C`, so the oldest segment lands on the active name and is truncated. That is right.

**One piece of honest context on blast radius.** Bundled activation is gated to `cfg(all(target_os = "macos", any(target_arch = "aarch64", target_arch = "x86_64")))` at `bridge/activation.rs:516-526`, with every other target returning `Disabled`, and the Desktop-side child bridge is macOS-gated throughout. So the fallback writer is never constructed on Linux in a shipped build today, and the original HIGH was latent rather than live. It was still worth fixing exactly as done: it removes a trap that would have detonated the moment Linux packaging lands, and it converts the entire rotation algorithm from untested-on-CI to tested-on-CI.

No further work needed on this fix.

---

## Findings

### MEDIUM | CONFIRMED | `anyharness/crates/proliferate-diagnostics-client/src/fallback/unix.rs:169-171` (with `:127-134`)

A transient `write(2)` failure permanently retires the component fallback writer for the remaining lifetime of the child process, with no re-arm path, conflating ordinary I/O errors with the tamper detection the fail-closed design exists for.

Chain, every link read:

1. `unix.rs:162-171` — `libc::write` returns -1 and the code returns `FallbackError::WriteFailed` without inspecting errno.
2. `unix.rs:127-134` — `write_encoded` matches `WriteFailed` in its retirement set and executes `self.active = false`.
3. `unix.rs:117-120` — every subsequent `write_encoded` returns `WriteFailed` immediately.
4. `producer/fallback_runtime.rs:145-156,171` — `status.active == false` means the writer is not restored to the slot and `self.active` is stored `false` on the controller.
5. `producer/fallback_runtime.rs:57-59` — every later `start()` returns `Unavailable`.
6. No `active = true` or `active.store(true, ..)` exists anywhere in the crate, and `FallbackWriter::from_directory` has exactly one call site, at install (`producer/mod.rs:202`). There is no reconstruction path.

Concrete scenario: a long-lived bundled `anyharness serve` session is routing to fallback because the collector is momentarily unavailable, the user's disk fills for a few seconds, one `write` returns `ENOSPC`. Space is freed, the collector comes back and goes away again — and for the rest of that session every fallback write is refused. The loss is counted (`fallback_write_failed`) and `fallback_active` is exposed in `ProducerStatusSnapshot`, so it is observable rather than silent, which is why this is MEDIUM and not HIGH. `EDQUOT`, `EIO` and a failed `ftruncate` (`:259-261`, `:288-289`) reach the same terminal state.

The same undifferentiated handling applies to the new exchange primitive: `atomic_swap` discards errno entirely, so `ENOSYS` / `EINVAL` / `EOPNOTSUPP` from a filesystem without `RENAME_EXCHANGE` support (NFS, 9p, FAT, some FUSE) is indistinguishable from a genuine tamper and produces the same permanent retirement at the first 512 KiB rotation. Not reachable today given the macOS gate, but it is the identical failure signature to the bug just fixed, and it will become reachable when Linux packaging lands. Suggest distinguishing transient I/O from identity violations, or at minimum classifying an unsupported-exchange errno separately.

Note this may be a deliberate posture. If so it is worth stating in `swap.rs`'s module docs, which currently justify permanence only for the torn-swap case.

### MEDIUM | PLAUSIBLE | `anyharness/crates/anyharness/src/telemetry.rs:227-249`

The legacy `anyharness.log` file sink is suppressed on the basis of bundled *activation* rather than successful diagnostics *installation*, so a bundled `serve` run whose producer fails to install has no persistent detail sink at all.

`telemetry.rs:228-242` sets `bundled = true` for both `Bundled` and `BundledDegraded`, then `:243-245` computes `file_sink` from `(!bundled)`. `install_local` (`:313-330`) swallows any `InstallError` into `None`. When that happens the registry gets only the console layer and the sentry layer: `diagnostics_layer` is `None` and `file_sink` is `None`. The sidecar is spawned with inherited stdio (`apps/desktop/src-tauri/src/sidecar.rs:343-345`), so in a packaged `.app` the console layer's output is effectively discarded. Net result is a whole `serve` session with no file evidence and no diagnostics records.

For the `BundledDegraded` case this is explicitly intended: `producer/mod.rs:203-210` returns `BootstrapInvalid` when degraded and the fallback directory is unusable, with a comment stating that the caller still suppresses legacy broad logging because bridge ownership is authoritative. I am not relitigating that.

The gap is the `Bundled` (Ready) path, which the comment does not cover. `install` can still fail there via `BridgeRuntime::start` returning `WorkerUnavailable` (`producer/mod.rs:256-266`, a thread-spawn failure) or `RecordFactory::new` returning `BootstrapInvalid` (`:212-213`, if `PROLIFERATE_RUNTIME_ENV` carries a value `valid_short` rejects). Marked PLAUSIBLE on reachability, not on mechanism: every link is verified in code, but both triggers are rare. `Handle::try_current` is not a trigger — both binaries call `init` inside `#[tokio::main]`, which I checked.

Suggested one-line shape: on the Ready path, key the suppression on `installation.is_some()` rather than on `bundled`.

`proliferate-worker/src/logging.rs:376-393` is not affected, because the Worker has no file sink of its own and Desktop drains its stdout either into the bounded tail or into the legacy `worker.log`.

### LOW | CONFIRMED | `specs/OBSERVABILITY.md:63`

The spec states that "Collector-ready supported-macOS bundled runs write no new `anyharness.log*`", but the code suppresses the legacy file sink for *every* bundled activation, including `BundledDegraded` and a failed install, which are precisely the not-collector-ready cases. The documentation understates the suppression exactly where a support engineer would go looking for `anyharness.log` and find nothing new for that run.

---

Each of these was investigated and refuted in code, not assumed away. Given that the prior round on this train shipped two false HIGHs, the refutation is recorded.

- **Wrong `renameat2` syscall number on some architecture.** Refuted: 316 on x86_64 and 276 on aarch64 in both the gnu and musl trees of `libc-0.2.189`, matching the kernel tables.
- **`RENAME_EXCHANGE` unavailable or differently valued under musl.** Refuted: defined once in `linux_like/linux/mod.rs:1217` as `c_uint = 2`, shared by gnu and musl.
- **`left`/`right` argument order could be transposed.** Refuted as a hazard class: `RENAME_EXCHANGE` on a single dirfd is symmetric, so transposing the two path arguments is semantically a no-op. Only flag position and dirfd position matter, and both are correct.
- **Dangling `CString` pointer after the extraction into `swap.rs`.** Refuted: `&c_name(left)` creates a temporary whose lifetime extends to the end of the enclosing statement, which includes the whole `swap::exchange(..)` call.
- **macOS arm broken by the extraction.** Refuted: signature and constant checked against `apple/mod.rs:4762,3324`; the call is unchanged and `i64::from(c_int)` is sound.
- **Rotation produces wrong segment ordering.** Refuted by deriving the three-swap sequence by hand (`A,B,C,D` to `D,A,B,C`) and by the contiguity assertion at `fallback/tests.rs:361`, which now runs on Linux CI.
- **`retained_bytes` can exceed `FALLBACK_TOTAL_BYTES`.** Refuted: `validate_file` caps each segment at `FALLBACK_SEGMENT_BYTES`, and every accumulation uses `checked_add(..).filter(|b| *b <= FALLBACK_TOTAL_BYTES)`.
- **A fallback JSONL line could carry a raw embedded newline and corrupt the segment.** Refuted by reading `fallback/wire.rs:54-56`: `serde_json::to_vec` escapes control characters, so the single appended `\n` is the only LF. Checked rather than assumed, since the mirror-image assumption about `JSON.stringify` produced a false HIGH last round.
- **Secret-scrubbing regexes recompiled on the hot path.** Refuted: both sets are `OnceLock`-cached (`producer/record.rs:388-428`).
- **`truncate_utf8` can underflow or panic on a small limit.** Refuted: `is_char_boundary(0)` is always true, so the decrement loop terminates; and no configured limit is below the marker length.
- **`install` can fail via `Handle::try_current` because telemetry initialises before the runtime.** Refuted: both `anyharness` and `proliferate-worker` call `init` inside `#[tokio::main] async fn main`.
- **The macOS-only test gating that hid the rotation bug survives elsewhere in the crate.** Refuted by sweeping the module declarations: producer, framing, wire and fallback test modules are `cfg(test)` or `cfg(all(test, unix))` and run on Linux CI. Only `bridge::activation`'s three test modules are macOS-gated, and correctly so, because the `platform` module they exercise does not exist on other targets (`activation.rs:516-526`).
- **Line-cap regression.** Refuted: the largest changed source file is 598 lines against `MAX_LINES = 600` in `scripts/check_max_lines.py`.
- **Lifecycle instrumentation missing for AnyHarness and Worker.** Not a finding: the P0 catalog assigns it to PR 9, and a later slice may not pull work forward from an earlier owner.

pablonyx:
## Post-review fixes + restack — head is now `d6287d0f2`

Force-pushed with lease. Since the sealed review head `2752e3164`:

- **Fallback writer no longer dies permanently on transient I/O.** A single failed `write(2)` retired the last-resort diagnostics sink for the process lifetime with no errno inspection. Now: `EINTR` retries within a bounded budget (8); every other transient error costs the record but keeps the writer, up to a consecutive-failure cap (16) so a dead filesystem still fails closed; permanent retirement is reserved for the integrity failures it was meant for (`UnsafeDirectory`/`UnsafeEntry`). Short writes roll back the partial line and survive. Three negative controls, 189/189 crate tests.
- **Legacy sink suppression keyed on installation, not activation.** The bundled `serve` adapter suppressed `anyharness.log` on collector-ready *activation*, so a run whose producer failed to install got neither diagnostics layer nor file log. Suppression now waits for the producer to actually install; the deliberate `BundledDegraded` suppression is unchanged. (The review's suggested flat `installation.is_some()` would have broken that deliberate case — the fix is per-path.)
- OBSERVABILITY.md updated; `check_max_lines.py` ledger entries for the two fallback files this fix grew (both with split triggers).

Also: this PR's body still describes the pre-restack red checks and misattributes the oversized frontend files as a "PR 2 baseline" — the frontend half starts at PR 1. Correcting by comment per the train's convention of not rewriting sealed acceptance records.

PR #1708, opened 2026-08-07
fix(product): render fenced code blocks in chat

tairabun:
## Summary

- Render fully enclosed Markdown fences typed or pasted in the chat composer as editable code blocks while incomplete fences remain literal text.
- Preserve surrounding rich content, code-local typing and paste behavior, attachment ownership, and command/mention behavior across code-block boundaries.
- Render sent user-message fences reliably, including unlabeled one-line and empty blocks, and document the composer contract.

## Readiness

- [x] I followed the [pull-request procedure](https://github.com/proliferate-ai/proliferate/blob/main/guides/process/pull-requests.md) for scope, title, labels, body, and readiness.

## Support and attribution (optional)

- [x] This PR has no support relationship or private source attribution.

## Verification

- `pnpm --filter @proliferate/product-client test` (795 files, 5,317 tests)
- `pnpm --filter @proliferate/product-client typecheck`
- `pnpm --filter @proliferate/product-client build`
- `pnpm --filter proliferate build`
- `pnpm --filter @proliferate/web typecheck`
- `pnpm --filter @proliferate/web build`
- Repository documentation, frontend structure, appearance, attribution, source-size, boundary, and diff guards
- Playwright browser verification at desktop and mobile widths for editing, overflow, continuation text, and sent transcript rendering

## Observability

None. This is deterministic client-side rendering with no new telemetry or failure path.

PR #1775, opened 2026-08-12
fix(product): guard composer code-fence ancestor walk against detached selection (PRO-209)

tairabun:
## Summary

Fixes [PRO-209](https://linear.app/proliferate-team/issue/PRO-209/composer-freezes-app-on-rapid-second-send-infinite-loop-in): the composer froze the entire window when a message was sent while the previous send was still materializing.

`isComposerSelectionPointInsideCode` (introduced in #1708) walked ancestors with an optional-chained loop condition:

```ts
while (paragraph?.getParent()?.getType() !== "root") {
  paragraph = paragraph?.getParent() ?? null;
}
```

When the Lexical selection points at the root or a detached node — the editor state right after a send clears the draft — `paragraph` becomes `null`, `undefined !== "root"` stays true, and the loop spins forever inside the editor update listener during a keydown microtask checkpoint. The renderer main thread pegs at 100% CPU and the window stops responding entirely ("can't send messages").

## Fix

Guard the walk on the cursor itself, the same pattern `offsetWithinNode` in this file already uses. Root/detached points now fall out of the loop and return `false`.

## Testing

- New regression tests pin both degenerate selection points (root node, detached node). Against the unguarded loop they hang the suite, which CI catches as a timeout; with the guard they pass in milliseconds.
- Full composer input suite: 120/120 pass.
- Manually verified end-to-end via CDP: a rapid 5-message send burst (the deterministic repro — second send ~300ms after the first) stays fully responsive; before the fix the renderer wedged on the second send, confirmed by `Debugger.pause` stacks landing in this loop.

## Notes

- Bug is on `main` only — merged after the `release-2026-08-08` cut, so v0.4.8 in production is unaffected. This should land before the next release cut.

### DIFF
--- a/apps/packages/product-client/src/components/workspace/chat/input/ComposerEditorDocument.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/input/ComposerEditorDocument.test.tsx
@@ -1,0 +1,36 @@
+// @vitest-environment jsdom
+
+import { describe, expect, it } from "vitest";
+import { $createTextNode, $getRoot, createEditor } from "lexical";
+import {
+  COMPOSER_NODES,
+  isComposerSelectionPointInsideCode,
+} from "#product/components/workspace/chat/input/ComposerEditorDocument";
+
+function withEditorUpdate(run: () => void) {
+  const editor = createEditor({
+    nodes: COMPOSER_NODES,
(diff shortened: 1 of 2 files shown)


tairabun:
tested the fix manually

PR #1761, opened 2026-08-11
fix(product): restore sessions from the full closed row

tairabun:
## Linear tickets

- [PRO-156 — Only text in session dropdown is clickable](https://linear.app/proliferate-team/issue/PRO-156/only-text-in-session-dropdown-is-clickable)

## Summary

- Restore a closed session when any non-button part of its highlighted row is clicked, including padding, gaps, timestamp, and the trailing glyph.
- Ignore row-level bubbling from nested controls so the title restore, Delete, and explicit restore actions each fire exactly once.

## Testing

- `pnpm --filter @proliferate/product-client exec vitest run src/components/workspace/shell/topbar/HeaderTabsActions.hit-target.test.tsx` (3 passed)
- `pnpm --filter @proliferate/product-client typecheck`
- `pnpm --filter @proliferate/product-client build`
- Full package run: 5,457 passed and 1 skipped; the remaining Chrome-only stylesheet qualification could not launch because the local Google Chrome channel is not installed.
- Compatibility merge with the sibling restore-icon change: 4 focused integration tests passed.
- `python3 scripts/check_frontend_boundaries.py`
- `python3 scripts/check_appearance_scaling.py`
- `python3 scripts/check_design_attribution.py`
- `git diff --check`

## Observability

- None. This changes local pointer dispatch only.

## Readiness

- [x] I followed the pull-request procedure for scope, title, labels, body, and readiness.

PR #1778, opened 2026-08-12
fix(product): delete closed sessions from the history popover (PRO-157)

tairabun:
## Summary

- Deleting a session from the closed-sessions (History) popover silently did nothing: `getSessionClientAndWorkspace` threw for any session without a live session-directory entry (every session not created/activated in the current app run — the popover's typical population, since the closed list is persisted), and `dismissSession` swallowed the throw, so no runtime dismissal was ever issued and the row resurfaced on the next sessions refetch. Bare ids without a directory entry now resolve to themselves (they are already runtime ids, matching `isPendingSessionId` semantics and what session selection does); genuinely pending ids still fail fast with the same message.
- The popover no longer closes on delete. The deleted row collapses in place via `AnimatedCollapsibleContent` (rows above stay static, rows below slide up over the disclosure duration), a ghost of the row outlives the data until the transition finishes, and the archive is deferred by the disclosure duration so the main-thread grid-rows transition is not starved into a single-frame snap by the header re-derivation it triggers. A failed archive expands the row back instead of losing it; closing the popover mid-animation does not lose the deletion.

Fixes PRO-157.

## Testing

- Unit (vitest): new `session-runtime.test.ts` pins the materialized-id resolution (entry-less id → itself, directory mapping preferred, pending id → throws); new `ClosedChatTabsMenu.test.tsx` pins collapse-on-click with static neighbors, deferred archive dispatch, ghost persistence until the transition window ends, and restore-on-failure. Related suites (`chat-session-archive`, `session-records`, `workspace-ui-store`, `AnimatedCollapsibleContent`) pass unchanged.
- Live end-to-end on a dev profile (local AnyHarness runtime, Desktop renderer): closed sessions from both the current and prior app runs delete from the popover with `dismissedAt` confirmed set via `GET /v1/sessions?include_dismissed=true`; the popover stays open with the remaining rows; deleting the last row unmounts the popover and History button together; frame-by-frame sampling of a middle-row delete shows a continuous ~60fps ease-out collapse (36→0px over 200ms) with the row below moving up in lockstep and rows above static.

## Observability

- none

## Readiness

  in this PR.

## Verification

- `pnpm --filter @proliferate/product-client build` (includes full tsc)
- `pnpm vitest run src/components/workspace/shell/tabs/ClosedChatTabsMenu.test.tsx src/lib/access/anyharness/session-runtime.test.ts src/lib/workflows/workspaces/chat-session-archive.test.ts src/primitives/__tests__/AnimatedCollapsibleContent.test.tsx` — 14 passed
- `python3 scripts/check_appearance_scaling.py && python3 scripts/check_design_attribution.py && python3 scripts/check_frontend_boundaries.py` — all pass

## Pinned-test change (flagged for review)

- `HeaderTabsActions.hit-target.test.tsx` ("keeps Delete isolated from the row restore action", added in #1761) pinned the pre-PRO-157 behavior that Delete closes the popover and dispatches synchronously. This PR intentionally changes both: the popover stays open and the archive is deferred past the collapse transition. The test now pins the new behavior (immediate collapse, deferred dispatch via waitFor, dialog stays open); the other hit-target cases are unchanged.

### DIFF
--- a/apps/packages/product-client/src/components/workspace/shell/tabs/ClosedChatTabsMenu.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/shell/tabs/ClosedChatTabsMenu.test.tsx
@@ -1,0 +1,117 @@
+// @vitest-environment jsdom
+
+import { act, fireEvent, render } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { motion } from "@proliferate/design/motion";
+import { ClosedChatTabsMenu } from "#product/components/workspace/shell/tabs/ClosedChatTabsMenu";
+import type {
+  HeaderChatMenuEntry,
+} from "#product/lib/domain/workspaces/tabs/workspace-header-tabs-view-model-types";
+
+function menuRow(id: string): HeaderChatMenuEntry {
+  return {
+    id,
+    title: `Session ${id}`,
+    agentKind: "claude",
+    viewState: "idle" as HeaderChatMenuEntry["viewState"],
+    isResolvingSession: false,
+    hasUnreadActivity: false,
+    isActive: false,
+    isVisible: false,
+    closedAt: null,
+  };
+}
+
+function renderMenu(
+  rows: HeaderChatMenuEntry[],
+  onDeleteSession: (sessionId: string) => Promise<boolean>,
+) {
+  return (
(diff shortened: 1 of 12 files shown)


tairabun:
@greptile review

tairabun:
@greptile review

PR #1780, opened 2026-08-12
fix(product): keep pasted code from poisoning the composer (PRO-159)

tairabun:
## Summary

- Fixes PRO-159: after pasting a rendered code block into the composer, every subsequent message — in the same session and every other session of the workspace — serialized as code until an app restart.
- Root cause had two shapes, both from rich (`text/html`) clipboard code:
  - A single-line `<code>` imported as Lexical's inline-code **text format**: invisible (the composer theme styles no code format), unescapable, serialized backtick-wrapped — and the format bits live on the *selection*, which survives the post-send root clear. The one editor instance is shared across session tabs, so the poison followed the user everywhere until a restart remounted it.
  - A multi-line `<pre><code>` imported as a code block **nested inside another code block** (both of Lexical's import rules fire), trapping the caret with no continuation line to escape to.
- Three normalizations, matching the composer's document model:
  - The external-value clear/import path in `ComposerEditorBridge` resets the selection's `format`/`style` along with the content, so a send never bequeaths formats to the next message.
  - The inline-code text format is unauthorable in the composer (typed backticks stay literal), so a `TextNode` transform keeps pasted characters and drops the format.
  - A code-node transform dissolves the double `<pre>`/`<code>` conversion into one block and guarantees the same continuation paragraph after a draft-final code block that the typed-fence path already creates (`selectComposerContinuationAfter`).
- `specs/codebase/systems/product/chat/composer.md` documents the normalization contract.

## Testing

- Unit (Tier 1, jsdom): new coverage in `ComposerFencedCodePlugin.test.tsx` (inline-code strip keeps characters; `<pre><code>` imports as one escapable block with the continuation paragraph) and `ComposerRichTextEditor.test.tsx` (external value replacement resets inherited selection formats). One existing `ComposerCommandEditor` expectation updated for the continuation paragraph's trailing newline — the same shape the pasted-fence test already asserts.
- `pnpm vitest run src/components/workspace/chat/input/` — all composer suites green (the 5 pre-existing `ProductHostProvider` resolution failures in this package reproduce identically on `main`).
- `pnpm typecheck` — clean.
- End-to-end (local, not committed): a Playwright harness drove the real editor + real `chat-input-store` loop in Chromium and WebKit with real keystrokes and three paste transports (synthetic `ClipboardEvent`, async-clipboard Cmd+V, macOS system pasteboard Cmd+V). Before: reproduced the full reported poison (invisible code, unescapable, `format=16` typing after send and across session switches). After: every scenario stays clean — single code block + escape paragraph on multi-line paste, plain characters on inline paste, `format=0` typing after send and across sessions.

## Observability

- none

## Readiness

  in this PR.

## Verification

- Paste a rendered code block (e.g. copy from a docs page or the transcript) into the composer: multi-line arrives as one visible code block with a paragraph below it; a copied inline code span arrives as plain characters.
- Type, select-all, overtype, send with Enter — the sent message renders as code only when it really is a fenced block, and the next message you type is plain, in the same session and after switching session tabs.

### DIFF
--- a/apps/packages/product-client/src/components/workspace/chat/input/ComposerCommandEditor.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/input/ComposerCommandEditor.test.tsx
@@ -448,9 +448,11 @@ describe("ComposerCommandEditor", () => {
 
     await waitFor(() => expect(fileMentionMock.lastQuery).toBe("set"));
     fireEvent.keyDown(textarea, { key: "Enter", repeat: false });
+    // The document ends with a code block, so the fenced-code plugin keeps a
+    // continuation paragraph after it — serialized as the trailing newline.
     await waitFor(() => expect(onDraftChange.mock.calls.some(
       ([draft]) => serializeChatDraftToPrompt(draft)
-        === "[setup.md](docs/setup.md) \n\n```\nconst ready = true;\n```",
+        === "[setup.md](docs/setup.md) \n\n```\nconst ready = true;\n```\n",
     )).toBe(true));
   });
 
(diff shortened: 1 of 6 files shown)


tairabun:
@greptile review

PR #1782, opened 2026-08-12
feat(tooling): registry-driven Claude Design sync builder

pablonyx:
## Summary

In-repo builder for the claude.ai Claude Design payload, replacing the frozen-branch converter that produced the Jul-28 upload (326 commits behind main at cutover). The CI-enforced playground library registry (`LIBRARY_TIERS`) is the manifest; sync scope is library-only per the ruled Component System ADR (surfaces tier cut).

**Already shipped with this build**: the payload was uploaded to a fresh design-system project ("Proliferate Library") and verified — the previous project was left untouched as rollback.

## What's here

- `scripts/design-sync/build.mjs` — orchestrator. Stages: registry entry barrel → vendor React (window globals) → Vite lib-mode IIFE bundle (React external via shims, `#product/*` aliased to live `src/`) → Tailwind v4 CSS from the design package dist (html paint override + generated token safelist) → per-entry card HTML / preview / `.jsx` / `.d.ts` / `.prompt.md` (purposes scraped live from `specs/DESIGN_SYSTEM.md`'s sanctioned index) → `_ds_sync.json` + README → headless-Chrome render check with per-card screenshots and contact sheets.
- Groups are kit-aware and **derived from entry names** (composer / toast / sidebar / settings out of their tiers); subpaths are read from the registry at build time — the in-flight kit-move PRs won't break the builder.
- `scripts/design-sync/CONTRACT.md` — the payload contract (reverse-engineered from the reference artifact) and upload choreography.
- `.design-sync/config.json` — pins the target project for future re-syncs.
- `specs/DESIGN_SYSTEM.md` — the sync sentence now points at `scripts/design-sync/` instead of "external tooling".

## Verification

- Clean end-to-end `node scripts/design-sync/build.mjs` from a wiped `.out/`: all stages green.
- Render gate: **81/81 cards mount with zero console/page errors** (negative controls confirmed the harness flags broken cards rather than always passing).
- All 6 screenshot contact sheets eyeballed: dark-first theme, real content, no white-on-white, overlay cards show their triggers (open-state stories are a ruled fast-follow).
- Post-upload `list_files` matches the local payload exactly (430 files); `_ds_sync.json` written last.


pablonyx (commented):
Reviewed the diff (`git show f31c365cf`) plus every new file in full, and cross-checked the `.out/` payload already on disk from the completed run (not re-run — just read the artifacts to confirm findings, since it's gitignored/untracked either way).

## Blocking

**`build-bundle.mjs:105-140` — the bundle's own `@ds-bundle` manifest comment uses the raw registry `name` (not the sanitized `displayName`), breaking exactly the slash-in-name edge case CONTRACT.md calls out by name.**

`registry-manifest.mjs` and `emit-cards.mjs` correctly strip `SecretManagementPanel`'s raw registry name `"secrets/SecretManagementPanel"` down to `displayName = "SecretManagementPanel"` before using it in a path — that's why the actual on-disk dir is `components/product-patterns/SecretManagementPanel/SecretManagementPanel.jsx`, and why `make-meta.mjs`'s `_ds_sync.json.sourceHashes` keys on that same clean path.

But `build-bundle.mjs`'s own `groupFor`/sourcePath construction (lines 105-140) works off the raw `entry.name` from its independently-parsed JSON sidecar, with no `displayName` sanitization step at all:

```js
const sourcePath = `components/${group}/${entry.name}/${entry.name}.jsx`;
```

I confirmed this against the actual built artifact in `.out/_ds_bundle.js`'s leading comment:

```json
{ "name": "secrets/SecretManagementPanel",
  "sourcePath": "components/product-patterns/secrets/SecretManagementPanel/secrets/SecretManagementPanel.jsx" }
```

vs. `.out/_ds_sync.json`'s `sourceHashes` keys for the same entry:

```
components/product-patterns/SecretManagementPanel/SecretManagementPanel.jsx
```

The two on-disk metadata files disagree about this component's path, and neither matches the real file on disk. This already shipped in the uploaded payload (render-check doesn't validate this field, so 81/81 green didn't catch it). Fix: have `build-bundle.mjs` sanitize `entry.name` the same way `registry-manifest.mjs`'s `displayNameOf()` does (or better — see the duplication note below, just consume a shared value instead of re-deriving).

## Should-fix

**`groupFor` is hand-duplicated in three places with no shared source, and the divergence above is a direct consequence.**
- `registry-manifest.mjs:91-97` (operates on sanitized `displayName`)
- `make-entry.mjs:124-130` (string-templated into the generated `.ds-entry.tsx`, operates on raw `libraryEntry.name` — unavoidably separate since it runs in-browser, but still a third hand-copy to keep in sync)
- `build-bundle.mjs:105-111` (operates on raw `entry.name` from its own JSON sidecar — this one *could* just consume `group`/`displayName` precomputed by `registry-manifest.mjs` via the sidecar, eliminating both the duplication and the bug above)

All three are byte-identical today, but nothing enforces that. A future kit-move or a new prefix rule edited in one place and not the others will silently produce three different opinions about which group an entry belongs in.

**Two independently-authored regexes parse the same registry source files and only agree by luck.**
`registry-manifest.mjs`'s `ENTRY_RE` (line 51, escape-aware, no `{` anchor) and `make-entry.mjs`'s `entryPairRe` (line 45, requires a literal `{` immediately before `name:`, not escape-aware) currently extract identical `{name, subpath}` lists from all four tier files (verified by running both against the live registry — 36/24/10/11, byte-identical). But they have different failure modes: e.g. reordering an entry literal's keys so something precedes `name:` would silently drop that entry from `make-entry.mjs`'s barrel/`__demos`/`__tiers` while `registry-manifest.mjs` (used by `emit-cards.mjs`/`make-meta.mjs`) would still count it, hard-failing loudly for the count mismatch but not for the actual cause. Worth collapsing to one parser (`registry-manifest.mjs`'s, since it already has the count-gate and escape-handling) with `make-entry.mjs` importing from it for the Node-side JSON sidecar, keeping only the truly-unavoidable in-browser `groupFor` copy.

**`make-entry.mjs` has no count gate, unlike `registry-manifest.mjs`'s hard `expectedCount`/`EXPECTED_TOTAL` checks (lines 26-35, 118-123, 148-150).** If the registry gains or loses an entry without the count constants being bumped, `build.mjs`'s pipeline will get partway through — `make-entry.mjs` and `build-bundle.mjs` happily produce a bundle with the new/changed entry count, then `emit-cards.mjs`/`make-meta.mjs` abort on the `registry-manifest.mjs` count gate — leaving `.out/` in an inconsistent half-built state (bundle reflects N entries, no per-entry cards for the delta). Not wrong given the gate exists somewhere, but worth a comment or an assertion in `make-entry.mjs` itself so the failure happens at the first stage instead of the fourth.

**Stale `apps/packages/design/dist/` is silently trusted.** Both `build.mjs:35` and `build-css.mjs:60` (`ensureDesignPackageBuilt`) only check `existsSync`, never mtime/hash against `apps/packages/design/src`. A future run where the design package's tokens/CSS source changed but `dist/` is left over from an earlier build (e.g. a stale worktree, or someone ran the design build for an unrelated reason) will silently ship outdated tokens/CSS with no warning — the exact "design dist is stale" scenario asked about. Consider always rebuilding, or at least a cheap staleness check (`dist/` mtime vs. `src/` mtime) that warns.

## Nit

- `CONTRACT.md:3` bakes in a machine-local absolute path (`/Users/pablohansen/Downloads/claude-design-reference-payloads/proliferate-ui`) as the "READ-ONLY ground truth for every format question." That's fine as a note for this run's provenance, but it's dead/unusable for anyone else opening this file (another dev, CI, a future agent on a different machine) — worth a caveat that it's Pablo's local reference copy, not a repo-relative path.
- `make-meta.mjs:154` (`computeScriptsSha`) hashes only top-level `scripts/design-sync/*.mjs` via non-recursive `readdir`, so edits to `shims/*.mjs` (which materially change bundle behavior — the React-external shim semantics) never move `scriptsSha`. Matches the literal CONTRACT.md wording ("concat of scripts/design-sync/*.mjs") but if `scriptsSha` is ever used downstream to detect "did the build logic change," a shim edit would go undetected.
- `make-entry.mjs`'s generated `.ds-entry.tsx` keeps the *raw* registry name (with slash) in `__tiers[].entries[].name` and as the `__demos` key, while every on-disk file/dir uses the sanitized `displayName`. Consistent within itself (the preview module in `emit-cards.mjs` also keys off raw `name` for the `__demos` lookup, so demos still resolve correctly), but it means the bundle's own metadata is a third, different naming convention from the payload's file layout and from `build-bundle.mjs`'s manifest comment — worth a one-line comment noting this is intentional (if it is) so a future reader doesn't "fix" it into a fourth inconsistent state.

## Confirmed clean
- Nothing under `scripts/design-sync/.out/` is tracked (`.gitignore` covers it, `git ls-files` confirms zero tracked files there).
- No secrets or credentials in the diff.
- `specs/DESIGN_SYSTEM.md`'s one-sentence edit is accurate and reads consistently with the rest of §5 (correctly repoints "the claude.ai design-project upload" from "external tooling, not an in-repo script" to `scripts/design-sync/` + "the upload to claude.ai runs from a local main-thread session").
- `make-meta.mjs`'s hash recipe (`styleSha`/`renderHashes`/`sourceKeys`/`scriptsSha`/`sourceHashes`/`auxSha`/`bundleSha12`) matches CONTRACT.md's documented recipe field-for-field, including correctly using `displayName` (not raw name) throughout — the drift is isolated to `build-bundle.mjs`'s separate, informal `@ds-bundle` comment.
- Card-mode/group derivation is genuinely path-independent (reads `subpath` from the registry at build time, groups by name not file location), so the kit-move durability claim in the PR description holds apart from the triplication risk noted above.

pablonyx:
## Consolidated agent review (3 reviewers) + fixes applied in 6bcc77b62

Three parallel review agents covered general correctness, payload-contract fidelity (diffed against the Jul-28 reference artifact), and build-script robustness.

### Blocking (found independently by two reviewers) — FIXED
- **`build-bundle.mjs` manifest `sourcePath` corrupted for slash-bearing registry names**: `secrets/SecretManagementPanel` produced `components/product-patterns/secrets/SecretManagementPanel/secrets/SecretManagementPanel.jsx` — a path existing nowhere in the payload, disagreeing with `_ds_sync.json`'s own `sourceHashes`. Fixed via the same display-name sanitization the other generators use; verified the rebuilt manifest now points at the real file.

### Should-fix — FIXED
- Ambiguous-export Rollup warnings now **fail the build** instead of warn-and-continue (a collision silently drops the binding from the namespace).
- `build.mjs` **always rebuilds the design package** — the presence-only `dist/` check silently shipped stale tokens after a `src/tokens.ts` edit.
- `build-css.mjs` resolves `@tailwindcss/node`/`oxide` **through the real module graph** (createRequire chain via apps/web's `@tailwindcss/vite`) instead of scanning pnpm store directory names — no more store-layout or duplicate-version assumptions.
- `make-entry.mjs` now **cross-checks its registry parse against `registry-manifest.mjs`** (the count-gated parser), so drift in either independently-written regex fails at one clean point instead of producing a partial `.out/`.
- `scriptsSha` now covers `shims/`, `templates/`, and `css/` inputs (CONTRACT.md recipe updated).
- (User-reported, same commit) card chrome tinted to theme tokens — the hardcoded `#e5e7eb` cell border read as a bright white outline on the dark theme.

### Noted, deliberately not taken now
- `groupFor` duplicated in three places / three naming conventions (raw vs display names) across the payload — a consolidation refactor, tracked for the fast-follow.
- `registry-manifest.mjs` fixed per-tier `expectedCount`s are an intentional tripwire (update them when the registry grows).
- Partial standalone reruns can leave `_ds_sync.json` stale until `make-meta.mjs` reruns — documented behavior; `build.mjs` ordering handles it.
- Defensive `react-dom/server` alias (currently unreachable from the registry).

### Verified clean by reviewers
Card HTML byte-identical to the reference boilerplate (incl. viewport/mode variants); sentinel/stub/styles.css byte-exact; `_ds_sync.json` has exactly the 9 contract keys; React shim coverage complete with correct alias ordering; no machine-local paths in committed files; nothing under `.out/` tracked; the `specs/DESIGN_SYSTEM.md` edit accurate.

### Post-fix verification
Render check re-run: **81/81 clean**. `_ds_bundle.js`/`_ds_bundle.css` byte-identical after the robustness patches (only the manifest header changed earlier). Delta re-uploaded to the live project with the sentinel choreography.

PR #1784, opened 2026-08-12
refactor(product): settings kit descent — kit members move to primitives/patterns/settings

pablonyx:
Mechanical descent of the settings area kit into its own subdirectory under `primitives/patterns/`, mirroring the composer/toast/sidebar move in #1781. Zero rendered-output changes: no prop, markup, or class edits in any moved component.

**Stacked on #1781 — retarget to `main` after A merges; merges after A.**

## Members moved

Already in the pattern tier, moved into the kit directory:

| Component | Old | New |
| --- | --- | --- |
| `SettingsGroup` | `primitives/patterns/SettingsGroup.tsx` | `primitives/patterns/settings/SettingsGroup.tsx` |
| `SettingsMenu` | `primitives/patterns/SettingsMenu.tsx` | `primitives/patterns/settings/SettingsMenu.tsx` |

Tier migration — `components/patterns/` (domain-aware) → `primitives/patterns/settings/` (pattern tier). None of these import domain code, so the kit-cohesion rule puts them one tier down:

| Component | Old | New |
| --- | --- | --- |
| `SettingsSection` | `components/patterns/SettingsSection.tsx` | `primitives/patterns/settings/SettingsSection.tsx` |
| `SettingsRow` | `components/patterns/SettingsRow.tsx` | `primitives/patterns/settings/SettingsRow.tsx` |
| `SettingsScopeTabs` | `components/patterns/SettingsScopeTabs.tsx` | `primitives/patterns/settings/SettingsScopeTabs.tsx` |
| `SettingsSaveFooter` | `components/patterns/SettingsSaveFooter.tsx` | `primitives/patterns/settings/SettingsSaveFooter.tsx` |
| `SettingsEmptyState` | `components/patterns/SettingsEmptyState.tsx` | `primitives/patterns/settings/SettingsEmptyState.tsx` |
| `SettingsSection.test.tsx` | `components/patterns/` | `primitives/patterns/settings/` (colocated, follows its subject) |

All eight show as git renames. The only content edits inside a moved file are three intra-kit import specifiers (`SettingsSection.tsx` → `SettingsGroup`; the test → `SettingsRow`/`SettingsSection`). `SettingsGroup.tsx` moves byte-for-byte, so the fragment/null-child divider handling from the settings wash restyle (#1777) is untouched.

## Deliberate exclusion

[...]

PR #1785, opened 2026-08-12
docs(design-system): UI-conformance review charter skill

pablonyx:
Turns the UI-conformance review checklist that landed with the component-hierarchy doctrine (#1779) into something a per-PR review agent can actually execute.

## What the skill checks

One section per checklist item in `specs/DESIGN_SYSTEM.md` (section Component Library, "UI-conformance review"), each with the doctrine sentence it applies, a runnable signature, and the carve-outs that must be applied before anything is written down.

| # | Check | Signature |
| --- | --- | --- |
| 1 | Placement: new shape vs. redraw | added-file walk through the placement algorithm; index-row completeness over both library tiers; a card/row redraw tell |
| 2 | Second instance (rule of two) | paint-fingerprint match against the tree, plus a dead-library-component scan for the inverse rule, plus a table of the duplicates already in the tree so they are not charged to a passing PR |
| 3 | Hand-rolled overlay semantics | `role="dialog\|alertdialog\|menu\|listbox\|tooltip"` outside `primitives/`, plus a secondary sweep for re-implemented focus traps, dismissal and positioning |
| 4 | Geometry escape hatches | the bracket families the appearance gate does *not* cover (width/height/padding/margin/inset), grouped by literal, with `[var(…)]` and grid positioning excluded |
| 5 | Icon source | lucide-react **identifier-set** diff between base and head, not a line grep |
| 6 | New-pattern quality | mechanical noun-tier test (domain-type props + 2+ library components) over the domain-aware tier, plus registry-demo honesty, tier placement, and job-naming for any newly added pattern |
| 7 | Hand-assembled state stacks | a vocabulary classifier that splits off-vocabulary stacks (findings) from in-vocabulary ones (eyeball for "does a primitive already own this?") |
| 8 | Rhythm | pane-root spacing census by role suffix, scoped to scaffold-bearing areas only |

Carve-outs are carried in substance, so the review does not fire on sanctioned patterns: the first-instance state-stack allowance, the `group` + `opacity-0` hover-reveal and muted-to-prominent colour-promotion idioms from `specs/frontend/styling.md`, kit-member exclusion from fission and from the rule of two, `DropdownMenu` as the sanctioned keyboard-menu path until `PopoverButton` parity, slot-filling as the mechanism working, virtualization math and grid positioning as legitimate geometry, and check 8's scoping to inter-pattern spacing at a scaffold root.

Everything is read-only greps, `git`, and one small Python helper. No builds, no installs, no test suites.

## How the signatures were validated

**Hit rate against the tree.** Each signature was scored against the violations named in `specs/DESIGN_SYSTEM.md` Current Gaps:

- Check 3 found **exactly the seven** hand-rolled overlay shells the doctrine counts.
- Check 2's dead-component scan found **both** components the doctrine names (`AuthProviderButton`, `ListRow`) plus a third it does not (`RangeSlider`); `ProductNotice` was confirmed separately as zero-call-site while its shape is hand-rolled at 5+ feature sites.
- Check 1's index-row scan surfaced `TypewriterRevealText` - a sanctioned-tier primitive with a live consumer and a registry entry but no index row. Not previously recorded.
- Check 7's classifier found 69 off-vocabulary state utilities across 49 files out of 483 total, and the top offenders are precisely the ones the specs name as banned (`hover:bg-muted` x15, `hover:bg-hover/30` x2 - styling.md's own anti-example, `hover:opacity-65` - the banned partial-opacity glyph transition). The other 405 uses are correctly suppressed.
- Check 5 measured 49 lucide-importing feature files, 36 of which shadow one of the 158 tuned glyph names in `primitives/icons/`. The doctrine's "~43 / 33" has drifted; the shape of the claim holds and the appendix records the refreshed counts.
- Check 2's tree run found 29 duplicate paint fingerprints in 2+ files, including two byte-identical cross-file pairs and a 9-file card shell.
- Check 8 found a live drift: 20 settings panes at `space-y-6`, three `repo/*Pane.tsx` at `space-y-5`.
- Check 4 found 173 ungated bracket sites (from the component-root scope recorded in the appendix).
- Check 6's noun-tier scan: 3 of 16 domain-tier files import a domain type, and the 13 that do not are led by exactly the six `Settings*` rows the doctrine already names for descent.

**False positives on two merged control PRs.** #1777 (settings wash restyle, 43 files - the heaviest recent UI PR) and #1778 (closed-session history popover, 12 files).

**0 false positives across both, after tuning.** Six candidate findings survive on #1777 and one on #1778, and each is a defensible conformance observation - expected, since none of these checks was enforced when those PRs were reviewed. The strongest is `py-[13px]` introduced 17 times across 11 files with no token and no sanction comment, which trips check 4 and check 2 simultaneously.

Three naive spellings were measured, found to produce false positives, and replaced. They stay recorded in the appendix so the tuning is auditable:

| Naive spelling | FPs | Replacement |
| --- | --- | --- |
| `^+.*lucide-react` | 1 on #1777 (the line only *removed* `KeyRound`) | identifier-set diff between base and head |
| `hover:\|active:\|focus-visible:` keyword | 7 undifferentiated candidates, 6 sanctioned | sanctioned-vocabulary classifier, HARD/SOFT split |
| `^+.*space-y-` | 7 on #1777, all intra-pattern spacing | pane-root census by indentation and role suffix |

The appendix also carries the exact commands to reproduce every number, including a negative control (`lucide --base e93d98636~1 --head e93d98636` must report 5).

## Note on `.gitignore`

`.claude/` was fully ignored, so the skill could not be committed. This adds a narrow negation: `.claude/*` stays ignored except `.claude/skills/`, which becomes shared tooling. `.claude/worktrees`, `settings.local.json` and the rest stay local. There were no pre-existing `.claude/skills/` files in the checkout, so nothing else is newly surfaced.

## Open questions

1. **Cross-reference from the spec?** The skill is auto-discovered by Claude Code from `.claude/skills/`, so it works without one. A pointer from `specs/DESIGN_SYSTEM.md` § UI-conformance review would help humans find it, but it means a spec link into `.claude/` and I did not want to risk `check_docs.py` in this PR. Say the word and I will add it.
2. **Three findings the doctrine has not recorded.** `RangeSlider` (dead), `TypewriterRevealText` (no index row), and the three `repo/*Pane.tsx` at `space-y-5`. Should these become Current Gaps entries, or issues?
3. **Check 4's grouping threshold** is currently judgment ("group by literal, the count makes the finding"). If you want it mechanical, the natural rule is: same literal at 3+ sites in one diff = finding, 1-2 = note.
4. **Should any of this graduate to CI?** Checks 3 and 5 are deterministic enough to be gates today (a `role=` ratchet and a lucide identifier ratchet), which would move them out of review and into the mechanical column. Checks 1, 2, 7 and 8 need the judgment layer and should stay here.


pablonyx (commented):
## Review: UI-conformance review charter skill

Reviewed as a correctness review of review tooling, so I ran it: every signature against the current tree, both control PRs, the negative control, and the docs/frontend Python gates. **The headline claims mostly hold** — check 3 finds exactly 7/7, the dead-component scan finds 3, the negative control reports exactly 5, and the HARD/SOFT classifier reproduces 0 HARD / 4 SOFT on #1777 and 1 HARD / 2 SOFT on #1778. All three novel findings are real. Two defects, though, are the kind that make review tooling worse than no tooling: one check does not match the doctrine item it claims to implement, and one signature reports a clean tree when it has not looked.

Findings ranked below. Everything is fixed on the branch (new head `fbf859357`) except where marked.

---

### HIGH

**H1 — Check 6 is not the doctrine's check 6.** `specs/DESIGN_SYSTEM.md` line 816, item 6, is *"New-pattern quality — honest registry demo, correct tier (shapes vs nouns), named for the job not the feature."* The charter's check 6 was the raw-DOM control gate, which is not on the doctrine's list at all — it is in the Gates table as the mechanical half of the behavior job, and the charter's own opening paragraph names `report_frontend_structure.py --strict` as something already covered in CI and pre-commit before telling you to run it as check 6.

Net effect: the charter claims "one section per checklist item" while dropping a doctrine check entirely. Registry-demo honesty and job-vs-feature naming were unreviewed anywhere in the skill (check 1 covers tier and index row, not these), and a mechanical gate was reviewed twice.

*Fixed.* Check 6 is now new-pattern quality with three sub-checks (name, tier, demo honesty). The mechanical gates moved to a "re-run these first" step in Setup. The lookalike-primitive material stayed, folded into check 6, because a locally defined `Button` under another name is genuinely a tier question. Scored against the doctrine's own Current Gaps claim: 3 of 16 `components/patterns/**` files import a domain type at all, and the 13 that do not contain exactly the six shape-only `Settings*` rows the doctrine names — HIT, 6/6.

**H2 — `dupes` reports "0 candidates" when run from anywhere but the repo root.** `build_index` shells `grep -r … <relative root>` through `subprocess.run` with no `check` and stderr swallowed. From `apps/`, both modes print a clean result:

```
$ cd apps && python3 ../.claude/skills/ui-conformance-review/signatures.py dupes --diff /tmp/pr1777.diff
--- 0 candidate second instances          # the correct answer is 1
```

SKILL.md's Setup computes `$REPO` but never `cd`s to it, so this is reachable from the documented flow. A green result that means "I did not look" is the worst available failure for a review signature.

*Fixed.* Paths resolve against `git rev-parse --show-toplevel`, `--diff` and `--root` are resolved against the caller's cwd before the chdir, and a missing root now exits non-zero with `index root(s) not found (run from the repository root)`. Verified identical output from `apps/desktop/`.

---

### MEDIUM

**M1 — the `.gitignore` negation un-ignores nested `.claude/` directories.** `.claude/` had no mid-string slash, so it matched at every depth. `.claude/*` has one, so it is anchored to the repo root only.

```
$ mkdir -p apps/desktop/.claude && touch apps/desktop/.claude/settings.local.json
$ git status --porcelain apps/desktop/.claude
?? apps/desktop/.claude/
```

Directory-scoped `.claude/` settings and skills in a sub-package are a supported Claude Code pattern, so this is reachable. `.claude/worktrees` itself stays ignored — I verified that specifically.

*Fixed* with an ordered four-line form (unanchored rule first so the anchored ones win):

```gitignore
.claude/
!/.claude/
/.claude/*
!/.claude/skills/
```

Verified: nested `.claude/settings.local.json` and even nested `.claude/skills/foo/SKILL.md` are ignored; root `.claude/worktrees` and `settings.local.json` are ignored; root `.claude/skills/**` is tracked.

**M2 — check 7's classifier is blind to the named-group form styling.md itself teaches.** `STATE_UTIL` requires the variant immediately after an optional bare `group-`, so `group-hover/file-diff:opacity-100` matches nothing — and styling.md § Hover Reveal Pattern explicitly says *"Name the group when nesting is possible"* and uses exactly that spelling in its example. 69 named-group state utilities exist in feature code today, all invisible. The output is also asymmetric: on `PendingPromptList.tsx:214` the classifier charges `focus-visible:opacity-70` as off-vocabulary while silently passing the identical `group-hover/queue-row:opacity-70` on the same line.

*Fixed.* The prefix now accepts `(?:group|peer)-` plus an optional `/<name>`. Four of the 69 turn out to be genuinely off-vocabulary and were previously undetectable. `peer-*` is covered too. Multi-variant stacks (`dark:hover:`, `data-[state=…]:hover:`) remain unsupported; there are zero in the tree, and that is now documented as a known blind spot rather than left implicit.

**M3 — check 2's index excluded `primitives/` and everything outside `product-client/src`.** `SKIP_PATH` dropped the library from the *index*, not just from the diff side, so a PR repainting a shape a primitive already owns could never match — the single strongest finding check 2 can produce. `DEFAULT_ROOT` also excluded `apps/desktop/src` and `apps/web/src`, both of which are in the skill's own `$FEATURE`.

*Fixed.* Split into `INDEX_SKIP` (library kept) and `SKIP_PATH` (library skipped for added lines), and `--root` now defaults to all three roots and is repeatable. Cost measured before committing: the tree census goes 29 → 31, the three new clusters each pair a feature file with a library file that already paints that shape (`PopoverSearchField`, `ConfirmationDialog`, `Switch`), and **neither control PR gained a candidate** — the 0-FP result survives.

**M4 — check 2's tree index had no defined ref.** Setup fetches the PR and dumps the diff but never says which checkout the index is built from. Reviewing from a checkout of the PR branch puts the PR's own additions in the index, so "this shape already exists elsewhere" stops distinguishing pre-existing duplicates from ones the PR just made. The appendix's control numbers were themselves measured against a tree already containing both merged PRs.

*Fixed* — documented in Setup: review from a base checkout, with the reason.

---

### LOW

**L1 — several appendix numbers do not reproduce from the documented commands.**

| Claim | Measured |
| --- | --- |
| `py-[13px]` x15 across 10 files on #1777 | the skill's own grouping command yields **17 across 11 files** (14 across 10 if the playground demo is excluded) |
| 173 ungated bracket sites | 173 only under `product-client/src/components` with tests and `playground/` excluded; the documented `$FEATURE` roots give **200** |
| 474 state utilities, 405 suppressed | **483 / 414** by the script's own regexes before my fix; **547 / 474** after |
| 49 files / 36 shadowing / 158 glyphs | all three reproduce exactly, but no documented command produces the 36 |

69 off-vocabulary across 49 files, `hover:bg-muted` x15, `hover:bg-hover/30` x2, the 9-file card shell, 29 fingerprints, 7/7 overlay shells, 20 `space-y-6` panes, and the negative control's 5 all reproduced exactly.

*Fixed.* All numbers re-measured and corrected, and the "Reproducing" block now carries a command for every number in the appendix in table order rather than 5 of ~12. The PR body's "the exact commands to reproduce every number" and "py-[13px] introduced 15 times across 10 files" should be updated to match.

**L2 — hard-coded counts framed as assertions rather than as census data.** Check 3's *"Seven hand-rolled shells predate the rule … adding an eighth is [a finding]"* and its four-name `DropdownMenu` consumer list, and check 2's known-duplicates table with per-shape file counts, all sat in the normative body. Every one is currently accurate — I verified 7/7 and 4/4 — but they read as numbers a future reviewer must match, and migrating a shell away would make the charter look wrong for doing the right thing.

*Fixed.* Check 3 now says to establish the baseline by running the tree grep **at `$BASE`** and to enumerate `DropdownMenu` consumers with a command rather than reciting a list. The duplicates table is labelled as a dated census that defers to Current Gaps. The appendix opens by stating outright that its numbers are point-in-time and must never be carried into a finding.

**L3 — `hover:bg-[var(--x)]` printed truncated** as `hover:bg-[var`, because the value character class omitted parentheses — and the skill's prose names that exact literal as a typical offender. *Fixed*, with a paren-balance trim so a utility ending a `cn(…)` expression does not pick up a stray closer.

**L4 — `--min-paint` was undocumented** in both the module docstring and SKILL.md, despite being check 2's sensitivity dial. *Fixed*, with argparse help text explaining why 3 is the default.

**L5 — the skills negation surfaces personal project skills.** Any engineer with a local `.claude/skills/whatever` now sees it as untracked and can sweep it into `git add -A`; `check_docs.py` also picks it up via `git ls-files --others`, so a local skill with a broken relative link fails the docs gate locally. Worth a line in the PR body. *Not fixed* — inherent to the negation, and the alternative (an explicit per-directory allowlist) is worse.

**L6 — cosmetics, not fixed.** `index[fp]` on a `defaultdict` in diff mode inserts empty sets; `added_lines` increments its counter on `diff --git`/`index` header lines before the first `@@`. Both are harmless.

---

### Cross-PR: stale kit claim (raised by the #1784 reviewer)

Check 2's "Split kits" row asserted the settings kit spans `primitives/patterns/` and `components/patterns/`. #1784 (settings kit descent, `origin/settings-kit-pr-b` head `96e5d9fd2`) is ahead of this PR in the queue and makes that false: after it lands, the settings kit owns `primitives/patterns/settings/`, and the only member left in `components/patterns/` is `SettingsPageHeader`, deliberately, because the settings slice retires it in favour of `PageHeader variant="flat"`. A reviewer reading the stale row would open a finding on a deliberate, doctrine-recorded exemption.

*Fixed*, verified against `origin/settings-kit-pr-b` rather than main. The row now names the toast positioner as the durable cross-tier case, defers the live list to DESIGN_SYSTEM.md § Current Gaps, and states that a member left outside its kit directory deliberately with the reason recorded in the doctrine is not a finding.

---

| | before | after |
| --- | --- | --- |
| #1777 `dupes` | 1 candidate | 1 candidate (unchanged) |
| #1778 `dupes` | 0 | 0 |
| #1777 `states` | 0 HARD / 4 SOFT | 0 HARD / 4 SOFT |
| #1778 `states` | 1 HARD / 2 SOFT | 1 HARD / 2 SOFT |
| #1777 / #1778 `lucide` | 0 / 0 | 0 / 0 |
| negative control #1744 | 5 | 5 |
| tree census | 29 fingerprints | 31 (library now indexed) |
| from a subdirectory | silent 0 | identical to root, or a loud exit 1 |

`check_docs.py` (209 files), `report_frontend_structure.py --strict`, `check_appearance_scaling.py`, and `check_frontend_boundaries.py` are all green on the branch. Frontmatter parses, `name` matches the directory, and every relative link resolves — `check_docs.py` validates this SKILL.md now that it is tracked, which is a nice free ratchet.

### Answering the PR's open questions

1. **Cross-reference from the spec** — worth adding. `check_docs.py` already validates links *out of* this file and passes; a link *into* `.claude/` from `specs/DESIGN_SYSTEM.md` is the same relative-link machinery and carries no new risk.
2. **The three unrecorded findings are all real** — I verified each independently. `RangeSlider` has zero non-playground call sites while holding an index row; `TypewriterRevealText` has a live consumer (`ChromeWorkspaceTab`) and a playground registry entry but no index row; `RepoActionsPane`/`RepoEnvironmentPane`/`RepoConfigurePane` sit at `space-y-5` against a 20-pane `space-y-6` mode. Current Gaps entries would fit the first two, since the doctrine already carries that exact kind of row.
3. **Check 4's grouping threshold** — the 3+ rule is fine, but note the count is per *occurrence*, not per file; #1777's `py-[13px]` is 17 occurrences across 11 files. Say which one the threshold counts.
4. **Graduating to CI** — checks 3 and 5 are ready as ratchets. Check 6's raw-DOM half already is one.

Not approving; that is Pablo's call.

pablonyx:
Fixes pushed as `fbf859357` (one commit, `fix(review): address review findings`):

- **H1** check 6 realigned to the doctrine's item 6 (new-pattern quality: name, tier, honest demo); mechanical gates moved to a Setup re-run step; lookalike-primitive material folded into check 6.
- **H2** `signatures.py` resolves paths against the repo root and exits non-zero on a missing root, instead of silently reporting a clean tree from a subdirectory.
- **M1** `.gitignore` re-ignores nested `.claude/` directories (the `.claude/*` form was root-anchored and un-ignored `apps/*/.claude/`); root `.claude/skills/` still tracked, `.claude/worktrees` still ignored.
- **M2** check 7 now sees the named-group form styling.md teaches (`group-hover/file-diff:…`) plus `peer-*`; four previously invisible off-vocabulary utilities surface. Remaining blind spot (`dark:`/`data-` stacks) documented.
- **M3** check 2's index keeps the library and covers all three `$FEATURE` roots; diff side still skips `primitives/`. Tree census 29 to 31, no new candidates on either control PR.
- **M4** Setup documents that the tree signatures must run from a base checkout.
- **Cross-PR** the "Split kits" row no longer hard-codes the pre-descent settings state; verified against `origin/settings-kit-pr-b` @ `96e5d9fd2`.
- **L1/L2** appendix numbers re-measured and corrected, reframed as point-in-time census data, and the Reproducing block now covers every number in table order.
- **L3/L4** `hover:bg-[var(--x)]` no longer truncates in output; `--min-paint` documented.

Re-verified after the changes: #1777 1 candidate / 0 HARD / 4 SOFT, #1778 0 candidates / 1 HARD / 2 SOFT, lucide 0 on both, negative control 5. `check_docs.py`, `report_frontend_structure.py --strict`, `check_appearance_scaling.py`, `check_frontend_boundaries.py` all green.

Two things left for you: the PR body still says "py-\[13px\] introduced 15 times across 10 files" (it is 17 across 11) and "the exact commands to reproduce every number" (now true, but it was not before), and its check table still lists check 6 as the raw DOM gate.

PR #1788, opened 2026-08-12
fix(anyharness): gate grok auto-probes on credentials (PRO-210)

tairabun:
## Summary

- Fixes PRO-210 ("Grok keeps opening in browser"): every unattended catalog-probe pass (startup, auth-applied, install-completed pokes) of an installed-but-logged-out Grok opened `accounts.x.ai/oauth2/device?user_code=…` in the browser and hung the probe to its 240s timeout.
- Root cause, verified against `@xai-official/grok` 0.2.102 over raw ACP: Grok's `authenticate` is **mandatory** before `session/new` (which otherwise fails `-32000 "Authentication required" / "no auth method id provided"`, even with `XAI_API_KEY` set), and its one advertised method (`grok.com`) resolves inline when the spawned world holds credentials but **starts a browser OIDC device-code flow** when it holds none. The runtime log shows the probe doing exactly that: `session_id=catalog-probe … calling authenticate method_id=grok.com`, with the next probe starting exactly 240s later.
- Fix: `targets.rs` (the probe engine's one policy seam) now admits an automatic Grok poke only when host credential detection (`~/.grok/auth.json`, ambient `XAI_API_KEY`/`GROK_API_KEY`) or the enrolled agent-auth route provides credentials — cursor's existing unattended-prompt law, conditioned on credentials instead of unconditional. Manual refresh still probes (user-initiated sign-in explains itself). Session launches were already refused pre-spawn by the create-time readiness gate.
- Deliberately NOT changed: `authenticate_if_advertised` still calls `authenticate` — a live experiment proved skipping it (or skipping the `grok.com` method) breaks every credentialed Grok world, including gateway-routed sessions, because `session/new` requires the prior `authenticate` round-trip.

## Testing

- Tier 1 (unit, `cargo test -p anyharness-lib --lib model_snapshot`): new `grok_automatic_probe_requires_credentials` drives the real `RuntimeProbeTargets` under env guards — credential-less → excluded; native `auth.json` → admitted; enrolled gateway route in `state.json` → admitted. Updated `the_production_exclusion_list_covers_cursor_only` for the conditional grok arm. 48/48 pass.
- Full `anyharness-lib --lib` suite: 1535 passed, 1 pre-existing failure (`sweep_tests::the_sweep_removes_only_abandoned_and_old_scratch_roots`, fails identically on the base commit with this change stashed — unrelated follow-up).
- Live protocol evidence gathered by driving the installed Grok agent over ACP stdio in an isolated `$HOME` (initialize / session-new with and without `XAI_API_KEY`, never calling `authenticate` so no browser was opened).

## Observability

- None net-new; the existing `skipping an automatic probe for a manual-refresh-only harness` debug line now also covers gated grok pokes, and probe attempts/authenticate calls were already traced (that tracing is how the root cause was pinned).

## Readiness

  in this PR.

## Verification

- Linear: PRO-210
- Log evidence (local runtime, 2026-08-12): grok probe spawned at 19:52:22, `authenticate method_id=grok.com` at 19:52:23, no `acp_authenticate.completed`, next harness probed at 19:56:22 — the 240s per-probe timeout.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/domains/agents/model_snapshot/targets.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/model_snapshot/targets.rs
@@ -8,10 +8,11 @@
 //! The engine no longer iterates auth contexts: one composed observation per
 //! harness, so a target is a harness, full stop.
 
-use crate::domains::agents::model::AgentKind;
+use crate::domains::agents::model::{AgentKind, CredentialState};
 use crate::domains::agents::readiness::service::resolve_agent_unrouted;
 use crate::domains::agents::registry::{built_in_registry, descriptor};
-use std::path::PathBuf;
+use crate::domains::agents::route_auth::launch_route_provides_credentials;
+use std::path::{Path, PathBuf};
 
 /// Harnesses excluded from every UNATTENDED probe.
 ///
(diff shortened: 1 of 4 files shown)


pablonyx (commented):
Reviewed at frozen SHA `ecb94316d4415839dcb67f5409829e073d4ba580`. Comment-only, no approve/merge.

**Single chokepoint confirmed.** All automatic pokes (`poke_all`, `poke_harnesses`, `poke_optional`, etc.) funnel through `ModelSnapshotService::poke_harness` (anyharness/crates/anyharness-lib/src/domains/agents/model_snapshot/mod.rs:300), which gates on `!reason.is_user_initiated() && !targets.allows_automatic_probe(harness_kind)` before returning early. `allows_automatic_probe` (targets.rs:56-68) is the one seam that names the Grok credential check, and grepping every call site of `poke_harness`/`allows_automatic_probe` shows no second path that bypasses it — matches the PR's claim of "one policy seam."

**Manual refresh still always probes.** `PokeReason::is_user_initiated()` short-circuits the gate entirely (mod.rs:304), so `run_attempt(..., PokeReason::Manual)` (mod.rs:443) is unaffected by the new credential check. Confirmed by reading the gate condition directly rather than inferring from the PR description.

**No permanent suppression for a logged-in user.** `grok_probe_world_has_credentials` (targets.rs:40-48) is evaluated fresh on every event (`Startup`, `InstallCompleted`, `AuthApplied`, `LoginTerminal`), not cached/latched. A false-negative credential read at one event (e.g. race with a login file write) only defers to the next automatic event — `AuthApplied` exists specifically to re-probe right after a sign-in completes — so there's no state that permanently wedges probing for a credentialed user short of every future event also reading stale credentials.

**Credential check reuses vetted logic, not new invented logic.** `launch_route_provides_credentials` (route_auth/mod.rs:196-232) and `resolve_agent_unrouted` are existing functions already used by the real readiness path; the PR's own doc comment (targets.rs:37-39) notes it deliberately mirrors `apply_launch_route_upgrade`. This lowers risk of a fresh credential-detection bug specific to this PR.

**Test adequacy.** `grok_automatic_probe_requires_credentials` (wiring_tests.rs:103-133) exercises both credential arms (native `~/.grok/auth.json` and enrolled agent-auth route via `gateway_state`) and the credential-less case. Without the fix, `allows_automatic_probe("grok")` always returns `true`, so the `!targets.allows_automatic_probe("grok")` assertion (line 115) would fail — this test is not vacuous. The existing `the_production_exclusion_list_covers_cursor_only` test was correctly narrowed to drop `grok` from the "must converge automatically" list, since that now depends on ambient env/credential state rather than being unconditionally true.

**Docs.** `specs/FEATURE_DOCS/MODELS.md` and `specs/anyharness/harnesses/grok.md` both updated with the credential-gate behavior and root cause, consistent with `specs/anyharness/README.md`'s scope for anyharness-lib changes.

No defects found. One minor observation (not blocking): `grok_probe_world_has_credentials` and the tier-1 `trial.poke(harness_kind)` call at mod.rs:321 both run after the gate check, so the trial is correctly gated too — worth confirming in a follow-up that no other tier-1/trial code path independently triggers a Grok spawn, since this review scoped to the automatic-poke gate specifically named in the PR.

PR #1789, opened 2026-08-12
refactor(product): retire lucide-react — all feature imports moved to primitives/icons, dependency dropped

pablonyx:
Retires every direct `lucide-react` import from product feature code and removes the dependency itself. Stacked on #1787 (vocabulary PR), which shipped the net-new tuned glyphs this rewrite targets; merges after it.

## What changed

Four mechanical rewrite commits, one per directory group (disjoint file sets, authored in parallel), plus a cleanup commit and a test re-truthing commit:

| Commit | Group | Files | Notes |
| --- | --- | --- | --- |
| chat/activity | `workspace/activity/**`, `workspace/chat/**` | 12 | 31 swaps; stale "sourced from lucide" comments deleted in `GoalTranscriptEventRow`/`TranscriptTurnChrome` |
| settings/secrets/billing | `settings/**`, `patterns/secrets/**`, `billing/**` | 20 | includes `SettingsSidebar`'s 17 identifiers |
| shell | `app/**`, `primitives/patterns/toast/ToastBody` | 4 | 11 swaps |
| workflows + rest | `workflows/**`, `content/ui/diff/**`, `workspace/repo-setup/**` | 13 | includes the 4 workflow editors |
| chore | `package.json` ×2, lockfile, `specs/DESIGN_SYSTEM.md`, `specs/frontend/*` refs | — | dependency dropped; icon-source doctrine text re-truthed (no more grandfathered files) |
| test | 2 test files | 2 | see below |

Consolidations applied per the pre-ruled map: `Trash2` → `Trash`, `Edit3` → `Pencil`, `Bot` → `Robot`, `GitFork` → `Fork`, `CheckCircle2` → `CircleCheck`, `HelpCircle` → `CircleQuestion`, `Cloud` → `CloudIcon`, `ChevronsUpDown` → `ChevronUpDown`, `type LucideIcon` → `ComponentType<IconProps>` (ActivityChips). Props, classNames and sizes preserved at every call site.

## Gates

- **Zero-lucide grep**: `git grep -l "lucide-react" -- '*.tsx'` over `apps/` returns nothing. The only remaining source reference in the repo is `server/artifact-runtime/src/jsx/library-registry.ts`, which maps the name to an esm.sh URL for sandboxed artifacts — a different subsystem, no workspace dependency, out of scope.
- **Coverage proof**: the fourth applier ran a repo-wide sweep on its own branch; the 36 files still importing lucide there were exactly the 12+20+4 owned by the other three groups. No orphan files.
[...]

PR #1792, opened 2026-08-13
refactor(settings): retire SettingsPageHeader onto PageHeader flat; StatusDot/NoticeBanner/Disclosure conversions (S1)

pablonyx:
## Summary

Executes the frozen slice spec `slice-specs/settings.md` for design-system wave 2, across the whole settings surface. Assembled by the integrator from a head change plus three parallel continuation packages (S1/S2/S3), then reviewed and fixed.

**Head change (mandatory)**: retire `SettingsPageHeader` in favour of `PageHeader variant="flat"` + `SettingsPageBody` as the settings page-chrome root, across all 25 panes that used the old header/root pair. `SettingsPageHeader.tsx` and its registry entry are deleted.

**S1 — chrome + flat panes** (5 commits): `SettingsSidebar` hand-rolled `HarnessStatusDot` → shared `StatusDot` (new `getHarnessAttentionDotTone()` + 6 unit tests); `AccountPane`/`OrganizationPane` local notice helpers → `NoticeBanner`; `SettingsContentBoundary` `<details>` → `Disclosure`; `OrganizationSsoSettingsSurface` banners → `NoticeBanner` and its connection/redirect rows → `SettingsRow`; `AppearancePane` preview shells + zoom stepper → `Card`/ghost; `OrganizationBudgetsPane` cards → `Card` (two `foreground-alpha` track fills retired); `SettingsScreen` chrome — ghost back button, width/rhythm handoff to `SettingsPageBody`; `RunCommandHelp` ghost-state cleanup.

**S2 — scoped sub-panes** (6 commits): account sign-in/password rows, GitHub App installation row, org identity row and `IntegrationRow` → `RosterRow`; org billing-link and name/logo rows → `SettingsRow`; `ApiKeyCreatorModal` error box → `NoticeBanner tone="destructive"`; billing usage meters → `ProgressBar`; invite-link pseudo-input → real `Input`; worktree-storage stepper ghost cleanup.

**S3 — repo + agents** (5 commits): `HarnessSection` deleted in favour of `SettingsSection`/`SettingsGroup` across the harness tree; harness identity/provider glyph tiles → `IconTile`; `HarnessConfigIssueBanner` rewritten as `NoticeBanner` + `IconTile`; `HarnessAuthApiKeyDetails` shells → `Card`; repo-pane controls onto shared state tokens.

**Integrator commit**: `SettingsPageBody`'s incubating note struck (its consumers land here) and its registry row re-truthed; stale width-handoff parenthetical corrected; contradiction recorded at `OrganizationIntegrationsPane`'s admin table (below).

**Review commit** (`79d6bcb99`): F1 `apps/desktop/scripts/check-design-system.sh` still named the deleted `SettingsPageHeader`, failing spec §2.5's "zero hits under `apps/`"; F2 the two `OrganizationInvitationsSection` `py-[13px]` rows keep their hand-roll but now record why (full-bleed control strips, not label/control rows); F3 `ProgressBar`'s `<div>` root was placed inside a `<span>` — markup fixed; F4/F5 two unenumerated pixel deltas documented at their sites (`rounded-xl` → `Card`'s `rounded-lg`; `size-7` → `IconTile md` 32px).

## Deliberate deferrals, each with rationale recorded in source

- `OrganizationMembersList` — five-column table; `RosterRow` has no cross-row column mechanism.
- `OrganizationIntegrationsPane` admin catalog grid — the spec's coupling premise was wrong: this pane never imports `IntegrationRow`. Its grid is the same five-column table class as the members list.
- `OrganizationSelectMenu` — `SettingsMenu` ships one `label` slot; three call sites need a visible value distinct from a semantic aria-label.
- `HarnessAllModelsSection` — always-visible summary beside a collapsing body; `Disclosure` has no slot for it.
- `ProviderRow` — `Disclosure`'s `AnimatedCollapsibleContent` never unmounts children; `ProviderPickerModal.test.tsx` relies on true unmount.
- `HarnessAuthApiKeyDetails` `h-[38px]` / `tracking-[0.1em]`, `ProviderPickerModal` size classes — no adoption target; the latter is a `ModalShell` library edit per spec §4.3.

All deferral rationales were independently verified as factually accurate during review.

## Founder decisions

1. **Org identity row loses a type step** — `OrganizationSettingsCard.tsx:48`, `text-body-emphasis`/500 → `RosterRow`'s `text-body`/400. Permitted by §0.1, but it de-emphasizes the one row in that card meant to read as identity. Restoring it needs an inline type override at a roster call site.
2. **Five notices gain a live region** — `NoticeBanner` assigns `role="status"`/`alert` by tone; the hand-rolls had none. Right for errors; the always-mounted harness config banner will re-announce on every visit.

## Verification

- Five gates green: appearance scaling, frontend boundaries, structure `--strict` (0), max-lines, docs.
- 63 test files / 470 tests green across `components/settings`, `primitives`, `lib/domain/agents`, `components/playground/library`. `CloudEnvironments.test.tsx` (the #1777 divider regression) passes unchanged 8/8.
- `HarnessPane.hierarchy.test.tsx` verified NOT stale: its assertions are text and `[data-harness-runtime-state]` queries, and `HarnessPane` never rendered `HarnessSection` directly. Confirmed by bridging the known `@proliferate/product-client/host/*` test-lane gap locally — 6/6 green, plus 8 other previously-unrunnable settings tests. That bridge ships as a permanent fix in PR G.
- Baseline ratchets with zero additions; bracket census 48 → 44; zero lucide identifiers.

Stacked on `lucide-pr-l` (#1789). Draft until the wave-2 holistic pass.


pablonyx:
## Review — PR #1792 (settings slice), full delta `237e675a2...d4195741a` (17 commits)

Verdict: **the slice is sound.** All five gates green, all 63 colocated test files green (470 tests), the release-qualification DOM contracts survive, and the deferrals are factually accurate. Findings below are one real spec-acceptance miss, one DOM-validity nit, and two unenumerated pixel deltas — all fixed on the branch in `79d6bcb99`. One item is a founder call.

### PRIORITY CHECK — `panes/agents/harness/HarnessPane.hierarchy.test.tsx`

**Not stale. No update needed.** Verified two ways.

Statically: the spec's §3.4 / R7 premise ("asserts pane structure, will need updating when `HarnessSection` is deleted") does not hold for the file as it stands. Every assertion is a text query or a `[data-harness-runtime-state]` attribute query; nothing reaches into section nesting or heading levels. `HarnessSection` was never rendered by `HarnessPane` itself — it lived inside `HarnessAuthSection` / `HarnessSettingsSection` / `HarnessAllModelsSection`, and the test `vi.mock`s all three. So S3's deletion is invisible to it. The one path S3 *did* change under this test is `HarnessConfigIssueBanner` → `NoticeBanner`, and `NoticeBanner` spreads unknown props onto its root, so `data-harness-runtime-state={agent.readiness}` still lands where `container.querySelector` looks for it.

Empirically: I bridged the two local env gaps that make it fail at load — a `@proliferate/product-client/host/*` → `src/host/*` resolve alias in a throwaway vitest config (deleted, never committed), plus the gitignored `src/generated/agent-registry.json` / `agent-catalog.json` copied from `catalogs/agents/` — and it runs **6/6 green at head, unchanged**. With the same bridge every other "env-gap" settings test also passes: `HarnessPane.test.tsx`, `RepoCloudGate.test.tsx`, `BillingSettingsSurface.test.tsx`, `WorktreeStorageSection.test.tsx`, `AppearancePane.test.tsx`, `UserIntegrationsPane.test.tsx`, `SettingsSidebar.test.tsx`, `AddCustomIntegrationDialog.test.tsx`. Nothing in this diff is hiding behind the gap.

### Findings

**F1 — MEDIUM · spec §2.5 acceptance fails.** `apps/desktop/scripts/check-design-system.sh:12` and `:62` still name the deleted `SettingsPageHeader` (the settings-structure rule comment, and the `text-[20px]` failure message "raw 20px (use SettingsPageHeader)"). The head commit's re-truth pass (`2eebfe678`) swept `apps/**` sources and `specs/**` but missed this script, so `rg "SettingsPageHeader" apps/` returns hits where the spec requires zero. Fixed — both now name `PageHeader variant="flat"`.

**F2 — MEDIUM · unrecorded deferral, C4 acceptance.** `panes/organization/OrganizationInvitationsSection.tsx:46` and `:79` are the two `py-[13px]` sites the frozen spec §4.1 mapped onto `SettingsRow`. S2 adopted the `Input` at L47 but left both rows hand-rolled — with no adoption, no cause comment on the bracket, and no deferral rationale anywhere. The deferral is *correct* on the merits (these are full-bleed control strips — a `flex-1` Input spanning the card with a trailing button, and a three-control `<form>` — not `SettingsRow`'s label-left/control-right split), but the spec's C4 rule is "removed by the adoption **or** carries a cause comment at the site". Fixed by recording the rationale and what the bracket transcribes at both sites.

**F3 — LOW · invalid DOM nesting introduced by the migration.** `panes/billing/BillingUsageUnitsSection.tsx:76`. The hand-roll was `<span role="progressbar">` inside a `<span className="flex flex-col gap-1.5">`; `ProgressBar`'s root is a `<div>`, so the adoption puts flow content inside phrasing content. Harmless in practice (unlike `<p>`, browsers don't repair it) but it is a regression from the previous valid markup. Fixed: the wrapper is now a `<div>`, which is legal because `SettingsRow` renders `description` inside a `<div>`.

**F4 — LOW · unenumerated pixel delta.** `panes/AppearancePane.tsx`. `PREVIEW_PANEL_CLASS` drew `rounded-xl` (12px); `Card` owns its radius and paints `rounded-lg` (8px), so both preview shells lose 4px of corner. The site comment enumerates the *fill* divergence but not this one, and spec §0.1 requires every hand-roll/library difference to be enumerated. Fixed by extending the comment (not by overriding the radius — that would reopen an axis `Card`'s doc deliberately closes).

**F5 — LOW · unenumerated pixel delta.** `panes/agents/harness/HarnessProvidersSection.tsx:166`. The overlapping provider fan was `size-7` (28px) and adopts `IconTile size="md"` (32px). This is the rounding the IconTile agent report anticipated and `IconTile`'s own doc sanctions, but it was not recorded at the call site. Fixed with a comment.

### Verified clean (no action)

- **Behavior preservation.** Every `variant="unstyled"` → `variant="ghost"` cleanup (`SettingsScreen` back button, `OrganizationBudgetsPane` usage rows + drill-down back, `CloudEnvironmentList` add button, `RunCommandHelp` link, both stepper pairs) drops *exactly* the string `ghost` already paints (`text-muted-foreground hover:bg-hover hover:text-foreground active:bg-active`). No state stack was silently lost. No `RowActionIconButton` / `visibility="hover"` anywhere under `settings/`, so the missing-`group`-ancestor hazard does not apply.
- **Release-qualification DOM contracts.** `data-integration-connected` and `data-integration-connect-trigger` both survive onto `RosterRow`'s root / the Connect button (`RosterRow` spreads `...props` first). `data-api-keys-pane` / `data-api-keys-state` survive both `ApiKeysPane` roots via `SettingsPageBody`'s passthrough. `data-harness-route-option` untouched (MethodCard correctly deferred). `data-harness-runtime-state` survives onto `NoticeBanner`.
- **The OAuth-pending regression stays fixed.** `IntegrationRow` renders the auth-kind label outside the pending/idle branch, and `IntegrationRow.test.tsx:96` asserts `getByText("OAuth")` *inside* the `oauthPending: true` case. Covered, not just correct.
- **RosterRow's aria-hidden leading slot.** All five migrations (sign-in method rows, connected-service row, GitHub App row, org identity row, integration row) carry their status in `trailing` text, and every leading mark is either decorative or duplicated by `title`. No accessible information lost.
- **`SettingsGroup` divider hazards (R1).** Every restructured child list is an array or a list of siblings, never a fragment. `CurrentUserInvitationsSection` correctly hoists its always-mounted `ConfirmationDialog` out of the group with the reason at the site. `CloudEnvironments.test.tsx` (the #1777 divider-count regression test) passes **unchanged**, 8/8.
- **Deferral rationales are all factually accurate**, including the two I could most easily have expected to be wrong: `ProviderRow`'s Disclosure deferral (`AnimatedCollapsibleContent` really never unmounts, and `ProviderPickerModal.test.tsx` really does rely on `queryByLabelText(...) === null`), and the org-integrations contradiction (`OrganizationIntegrationsPane` really does not import `IntegrationRow` — the spec's §3.3 coupling premise was wrong, and the integrator's recorded correction is right).
- **Baseline / census.** Three `foreground-alpha-foundation` staged entries retired (AccountPane, OrganizationPane, OrganizationBudgetsPane ×2) plus the matching census row; `WorktreeStorageSection` correctly retained (its stepper `bg-foreground/5` is deferred). Zero additions. `AccountPane` also drops out of `frontend_structure_allowlist.txt` after the `AccountNotice` deletion. Appearance gate ratchets, not merely passes.
- **C4 bracket census** falls 48 → 44 under `settings/`. Every surviving `py-[13px]` now carries a cause comment (after F2). Zero `lucide` identifiers under `settings/` or in any package.json.
- **Doc truthfulness.** `SettingsPageBody`'s "Now consumed by the settings pane roots" is true — 20 pane roots consume it. Its incubating note is correctly struck; the `Card`/`NoticeBanner`/`RosterRow`/`IconTile` incubating notes are correctly *not* struck (sibling-branch owned). `check_docs.py` and `library-registry.test.ts` both green after the registry-row deletion.
- **Cherry-pick seams.** No duplicate helpers: `ACCOUNT_ROW_CLASS` is gone from both its definition and its cross-file consumer (both in S2), `HarnessSection` has zero surviving references, and the one genuine S1/S2 coupling pair the spec flagged turned out not to exist.

### FOUNDER

**FOUNDER-1 — org identity row loses a type step.** `panes/organization/OrganizationSettingsCard.tsx:48`. The organization name was `text-body-emphasis font-medium` (15px / 500, tuned by #1777's wash); `RosterRow density="comfortable"` paints titles at `text-body` (14px / 400). Adopting the shared row is right, and the delta is the kind §0.1 permits — but it visibly de-emphasizes the org identity header, which is the one row in that card meant to read as an identity block rather than a setting. Left alone: keeping it needs `title={<span className="text-body-emphasis font-medium">…</span>}`, and whether that's worth an inline type override at a roster call site is a visual judgment, not a review defect.

**FOUNDER-2 (informational) — five settings notices gain a live region.** `NoticeBanner` assigns `role="status"` (or `alert` for destructive) by tone. The five hand-rolls it replaces had none, so AccountPane/OrganizationPane status notices, both SSO banners, the API-key-creator error and the harness config-issue banner now announce to screen readers. Correct for the errors; the always-rendered harness config banner will re-announce on every visit to a harness pane that needs setup. That's `NoticeBanner`'s owned contract, so no change here — flagging in case the announcement cadence is not what you want.

**FOUNDER-3 — the PR description is stale and now misleading.** It still describes only the head change plus a partial S1, lists AppearancePane / OrganizationBudgetsPane / SettingsScreen as "deferred, not started", and says "S2 / S3: not started" — all three packages plus the integrator commit are on the branch. Its "Doc-change data (for the integrator to apply)" block has also been applied. I have not rewritten it; say the word and I will.

### Status at head

Five gates green: `check_appearance_scaling.py`, `check_frontend_boundaries.py`, `report_frontend_structure.py --strict` (TOTAL 0), `check_max_lines.py`, `check_docs.py` (208 files).
Tests: 63 files / 470 tests green across `components/settings`, `primitives`, `lib/domain/agents`, `components/playground/library` — with the local host-alias bridge described above; without it, 8 files fail at load on the pre-existing missing-dist gap, which this PR neither causes nor worsens.

PR #1793, opened 2026-08-13
refactor(product): shell slice — PanelHeaderEntry + StatusDot consolidation (wave 2)

pablonyx:
## Summary

Executes the frozen shell slice spec (`slice-specs/shell.md`, including both Integrator addenda) for design-system wave 2. Consolidates right-panel header entries and the delegated-agent/subagent status glyphs onto the newly promoted `PanelHeaderEntry` and `StatusDot` primitives, drops dead ARIA/dead-code, and deletes one unused component.

Base: `52944791e` (tip of `lucide-pr-l` at slice start). Head: `a1702b450`.

## Conformance checklist

- [x] §2-A/2-B — right-panel header entries (tool, terminal-no-record, terminal-with-record resting state, viewer) consolidated onto `PanelHeaderEntry`; `TerminalHeaderButton`/`ToolHeaderButton`/`ViewerHeaderButton` deleted.
- [x] §2-C — `DelegatedAgentHoverCard`: Tooltip conversion **escalated, not applied** (see Deviations). Safe dead-code removal (unreachable `onCardClick` mode) landed instead.
- [x] §2-D — StatusDot adopted in `renderDelegatedAgentIcon` (tab-rendering.tsx) for `needs_attention`/`failed`/`queued`/`wake_scheduled`; `running` stays hand-rolled (no pulsing/currentColor tone in StatusDot's budget) but is resized onto `icon-status`.
- [x] StatusDot adopted in `ChatTabsMenu`'s active-subagent trailing dot.
- [x] `renderChatTabStatusBadge`'s unread-activity dot left untouched (Integrator addendum 1 — no "unseen" tone, no second consumer).
- [x] §2-E — `OpenTargetMenu`: redundant `role="menu"`/`role="menuitem"` dropped. Shortcut hint deliberately **not** migrated to `ShortcutBadge` (visible paint delta — filled pill vs. bare text); spec's own fallback clause.
- [x] §2-F / cleanup — unused `PaneSideOverlay.tsx` deleted (zero consumers, confirmed by grep); its `standardNumericZ` baseline entry removed.
- [x] Sanctioned single library edit — `PanelHeaderEntry` roving-tabIndex floor (`tabIndexFloor` prop), with a keyboard test (Integrator addendum 2). Own commit: `a959d1fa5`.
- [x] Zero registry edits (`components/playground/library/**` untouched).
- [x] Zero new components.
- [x] No edits to `specs/DESIGN_SYSTEM.md` (doc-change data below, for the integrator to apply).
- [x] Baseline JSON: only a key removal (`PaneSideOverlay.tsx|z-30`), no additions, no widened slack.

## Deviations / disclosed contradictions

1. **§2-C Tooltip conversion — escalated, not completed.** `Tooltip.tsx`'s `content` prop is `string`, not `ReactNode`, so it can't host `DelegatedAgentHoverCard`'s identicon + name + origin + key/value rows without a second library edit outside this slice's sanctioned scope. Composing the lower-level `tooltip-primitive.tsx` inside the existing `PopoverButton`-based trigger chain is architecturally delicate (non-forwardRef Radix Provider/Root wrappers aren't a single cloneable trigger) and needs hand verification of focus neutrality in a running app — exactly the spec's own Risk 3, which this sandboxed environment can't perform. Invoked the spec's §2-C fallback clause: applied only the safe, ruled dead-code removal (unreachable `onCardClick` mode, its hide-grace timer, `isInsideCard`, `cardAriaLabel`) and left the read-only hand-rolled portal in place.
2. **WP-B (tabs-kit repoint) — no-op, already done upstream.** `git log -S"ChromeWorkspaceTab"` shows the tabs-kit vocabulary PR (commit `c2250a3b8`) already renamed `ChromeWorkspaceTab`→`ChromeTab` and repointed all call sites. The frozen spec's literal import-path string is stale against current `main`/`lucide-pr-l`. Disclosed, not silently skipped.
3. **WP-E (sidebar repoint) — no-op, already done upstream.** PR A already repointed all `shell/sidebar/*.tsx` files to the sidebar kit's promoted paths, matching the spec's own Risk 1 instruction to drop this WP if already done.
4. **`ChromeTab`/`TabGroupPill` still carry literal `z-10`/`z-20`/`bg-foreground/5`**, not converted to semantic z-index/token utilities by the tabs-kit promotion — matches the spec's own pre-anticipated "escalation on the tabs-kit move" scenario. The baseline JSON already accounts for these values (not touched by me — out of this slice's ownership, belongs to the vocab PR).
5. **4 accepted-and-ignored `ChromeTab` props** (`groupColor`, `isChild`, `hideLeftDivider`, `hideRightDivider`) — nominally WP-B's job to wire or drop, but WP-B turned out to already be complete upstream without addressing Integrator addendum 1's item on these. Left as an **open item**, not silently dropped — needs a follow-up decision (wire them or delete them from the prop type).
6. **Uncounted consumer found during implementation:** `PlaygroundAttachmentFixtures.tsx` imported `ViewerHeaderButton`, which the spec's file inventory did not list (it claimed only two out-of-directory consumers). Repointed to `PanelHeaderEntry` + `RowActionIconButton` the same way as the in-directory call sites (file is outside the forbidden registry path, so editing it is in-scope).
7. **Close-button composition kept separate from `PanelHeaderEntry`.** `PanelHeaderEntry` has its own `onClose` prop, but wiring it directly for terminal-with-record/viewer entries would have silently dropped the `data-right-panel-tab-no-drag` exclusion that `use-right-panel-header-drag.ts` needs on every close button (confirmed by reading the drag hook's pointer-capture logic) — that would have caused close clicks to briefly activate/select the tab being closed. Composed the close button separately at each call site instead of widening the sanctioned primitive's API a second time.
8. **New gate violation surfaced and fixed, not a design change.** Refactoring the `running`-case status dot into per-case JSX with a static className exposed a pre-existing `size-1.5` fixed-size utility to `check_appearance_scaling.py`'s `fixed-status-glyph-utility` static regex scan (the original code hid it behind a template-literal interpolation). Fixed by swapping `size-1.5` → `icon-status`, the same semantic sizing utility `StatusDot` itself uses internally — this makes the dot visually consistent with its three siblings in the same function, which already moved onto `icon-status` sizing when they adopted `StatusDot`.

## Design sign-off needed (Integrator addendum 1)

Two visible changes ship in this slice that need explicit design sign-off before merge:
- Right-panel tab close-button placement: leading-icon-overlay → trailing edge.
- Close hit target: 16px → 20px (`size-icon-button-sm`).

## Doc-change data block (for the integrator — not applied by this PR)

`specs/DESIGN_SYSTEM.md` changes to apply once this slice merges:

- **Index-row rewording** — PanelHeaderEntry's incubating note: remove "single incubating consumer" language now that the shell right-panel + `PlaygroundAttachmentFixtures` are real second/third+ consumers, and add the roving-tabIndex-floor requirement to its capability line.
- **Incubating-note removals**: strike "incubating" status for the tabs-kit and StatusDot consumption in the shell area (both now have real, shipped consumers via this slice, beyond the earlier vocab-PR adoption).
- **Current Gaps — strike**: "Right-panel header entries not yet on PanelHeaderEntry" (closed by this slice). "Delegated-agent/subagent status glyphs not on StatusDot" (closed, except the disclosed `running`-tone exception, which should be captured as a permanent, named exception rather than a gap).
- **Current Gaps — do not strike, update instead**: `DelegatedAgentHoverCard` Tooltip conversion — reword from "not yet done" to "blocked: Tooltip.tsx content prop is string-only; needs a library edit + hand verification, tracked as open" (see Deviation 1 above).
- **New gap to add**: `ChromeTab`'s `groupColor`/`isChild`/`hideLeftDivider`/`hideRightDivider` props are accepted but unused — needs a follow-up decision (wire vs. delete).

## Test results

- `PanelHeaderEntry.test.tsx` — 6/6 passed (includes the new roving-tabIndex-floor keyboard test).
- `tab-rendering.test.tsx` — 5/5 passed (before and after the `icon-status` gate fix).
- No colocated tests exist for `RightPanelHeaderEntryList.tsx`, `TerminalHeaderIcon.tsx`, `PlaygroundAttachmentFixtures.tsx`, `ChatTabsMenu.tsx`, `DelegatedAgentHoverCard.tsx`, `OpenTargetMenu.tsx`, or the deleted `PaneSideOverlay.tsx`.

Gates (all green):
- `python3 scripts/report_frontend_structure.py --strict`

Gates **not runnable in this environment** (pre-existing, unrelated to this slice's changes — confirmed the donor worktree has the same gap):
- `apps/packages/design/scripts/check-theme.mjs` — fails with `ERR_MODULE_NOT_FOUND: tailwindcss`; `apps/packages/design/node_modules` doesn't exist in either this worktree or the donor worktree it symlinks from.
- `scripts/check_theme_contrast.py` — requires `apps/packages/design/dist/theme.css`, which requires a package build; building is out of scope under this session's no-build machine constraint.

No `npx tsc`/full test suite/build was run, per the machine-limits constraint (24GB RAM, no repo builds).


pablonyx (commented):
## Adversarial review — shell slice (`slice-shell`)

Reviewed the delta `237e675a2...1ff0c8090` against the frozen spec (`slice-specs/shell.md` + both integrator addenda). Four defects found; three are fixed on the branch (new head `45e15d27e`), the rest are flagged. Three items need Pablo's visual sign-off before merge.

---

### BLOCKER 1 — the right-panel close button is invisible and unclickable (FIXED)

`TerminalHeaderIcon.tsx` and `RightPanelHeaderEntryList.tsx` compose the close affordance at the call site as a bare `RowActionIconButton` sibling of `PanelHeaderEntry`, inside `<div className="right-panel-terminal-tab-shell">` (deviation 7 in the PR body).

`RowActionIconButton` defaults to `visibility="hover"`, whose reveal contract is:

```
pointer-events-none opacity-0 group-hover:pointer-events-auto group-hover:opacity-100 …
```

`group-hover:` compiles to `.group:hover &`. There is **no Tailwind `group` ancestor** anywhere in the chain — `.right-panel-terminal-tab-shell`, `.right-panel-header-entry-shell`, `.ui-tab-system-tabs__section`, `.right-panel-tab-system` are all plain CSS class names, and `RightPanelFrame.tsx:94` only carries `data-group`. `PanelHeaderEntry`'s *own* close works because `ENTRY_SHELL_CLASS` (`PanelHeaderEntry.tsx:47`) starts with `group` — but the call-site close is outside that div.

Net effect on main-line behaviour: **you can no longer close a terminal tab or a viewer tab with the mouse.** The button is `opacity-0` and `pointer-events-none` at rest and only `focus-visible:` can revive it. All three sites had it: `TerminalHeaderIcon.tsx:179`, `RightPanelHeaderEntryList.tsx:217`, `PlaygroundAttachmentFixtures.tsx:206`.

Fixed by adding `group` to those three wrappers, which also restores the reveal-on-tab-hover behaviour the retired `.ui-tab-system-tab__close-container` CSS provided (`product.css:2198-2202`).

### BLOCKER 2 — `DelegatedAgentHoverCard`'s "clickable mode" is not dead (FIXED)

Spec §2-C ruling 2, and commit `1ec78e23a` following it, both assert `ChatTabWithMenu.tsx:154` is the only call site and passes neither `onCardClick` nor `cardAriaLabel`. That is false:

```
apps/packages/product-client/src/components/workspace/chat/tool-calls/SubagentToolActionRow.tsx:126-131
  <DelegatedAgentHoverCard
    agent={hoverAgent}
    cardAriaLabel={`Open ${identity.displayName}`}
    onCardClick={canOpenSession ? openTarget : undefined}
  >
```

Two consequences: (a) a live affordance disappears — clicking the subagent hover card in the chat transcript to open that subagent's session; (b) `SubagentToolActionRow.tsx:128-129` no longer typechecks, since `DelegatedAgentHoverCardProps` only extends `HTMLAttributes<HTMLDivElement>` and JSX excess-property checking rejects the two props. No `tsc` ran in this environment, so CI would have caught (b) — but not before burning a cycle.

Same class of miss as disclosed deviation 6 (`PlaygroundAttachmentFixtures` importing `ViewerHeaderButton`): the spec's call-site inventories are not exhaustive and need a repo-wide grep, not a read of the named file.

Fixed by restoring the file to its base state plus a note recording why the ruling's premise fails. The role conflict the doctrine objects to (a `role="tooltip"` element that is also a `<button>`) is real and stays open as residue.

### MAJOR 3 — the shortcut badge overlays the terminal tab's label (FIXED)

`.right-panel-shortcut-badge` is `position: absolute; right: 4px` (`product.css:2160-2168`) and therefore contributes zero width. On main, room was reserved by `.ui-tab-system-tab__content[data-shortcut-reveal="true"] { padding-right: 1.65rem }` (`product.css:2094`). The refactor drops `data-shortcut-reveal` (the `__content` span is gone) but keeps the badge's class, so during shortcut reveal the badge paints on top of the label's tail. Fixed with `pe-7` on the entry when the badge is present (`TerminalHeaderIcon.tsx:154`, `RightPanelHeaderEntryList.tsx:173`).

Longer term this is residue item 10 (`product.css:2164`'s off-ladder `z-index: 2`). The clean end state drops the class and lets `ShortcutBadge` flow in `PanelHeaderEntry`'s `trailing` slot — a visible paint change (`--right-panel-tab-text-tertiary` → the entry's `text-current`), so deliberately not done here.

### MEDIUM 4 — `max-width` silently lost on the tool and terminal-without-record entries (FLAGGED)

`.right-panel-tab-system .ui-tab-system-tab` carried `max-width: var(--tab-max-width)` (156px) plus `overflow: hidden` (`product.css:1984-2009`). `PanelHeaderEntry` replaces that class and defines no max-width. Viewer and terminal-with-record entries are still capped because `.right-panel-terminal-tab-shell` has its own `max-width: 160px`, but the **tool** and **terminal-without-record** entries render `PanelHeaderEntry` directly into the drop zone with `shrink-0` and no cap. Tool labels are `Scratch`/`Changes`, so this is latent today; a long `terminalHeaderDisplayTitle` on a not-yet-materialised terminal would stretch the strip.

Same cause, same place: `.ui-tab-system-tab[data-reorderable="true"] { cursor: grab }` (`product.css:2022`) no longer matches, so the drag-affordance cursor is gone even though `data-reorderable="true"` is still emitted. Both belong to the "retire the CSS-owned anatomy into the pattern" slice, not here.

### MEDIUM 5 — addendum 1's "delete `.right-panel-tab-system` CSS in this slice" was not done, and could not have been (FLAGGED)

Addendum 1 says the `~product.css:1984-2270` block "becomes dead on migration; delete in this slice". It is not dead: `TerminalHeaderIcon.tsx:19,102-112` still renders the rename form on `ui-tab-system-tab`, `__content`, `__icon`, `__label-edit-slot`, `__label-input` and `right-panel-terminal-edit-action`; `RightPanelHeaderTabs.tsx` owns the `ui-tab-system-tabs__*` scaffold; `right-panel-shortcut-badge` is still consumed. The addendum also contradicts §2-A's explicit non-goal ("No edit to `apps/packages/design/src/css/product.css`"), and `appearance-css-drift.test.ts:295-330` asserts several of those rules exist. Leaving the CSS is the right call; the addendum's instruction should be retargeted at the follow-up slice. Genuinely dead now and safe to sweep there: `__close`, `__close-container`, `__dirty-indicator`, `__label`, `__label-primary`, `.right-panel-tool-tab`.

### MINOR

- **Tone fidelity, `ChatTabsMenu.tsx:267`.** `bg-foreground/70` → `StatusDot tone="muted"` (`bg-muted-foreground`). §2-F asked for "filled, neutral tone" so the substitution is sanctioned, but it is a real ink change, not like-for-like. It also drops `[font-size:var(--text-sidebar-row)]` (13px), so `icon-status` now resolves against the menu row's `text-ui-sm` (12px): 7.15px → 6.6px. Sub-pixel; noted only because §2-F explicitly told the implementer to keep the utility if the size did not carry.
- **`tab-rendering.tsx` dot sizing.** All four delegated-agent dots move from `size-1.5` (fixed 6px) to `icon-status` (`0.55em`, ≈7.2px at the tab's 13px `--text-ui`). ~20% larger, applied consistently to all four including the hand-kept `running` case, so the family stays internally coherent — but it is a size delta against main. It is the right *kind* of change (fixed px → appearance-scaling tier). The disclosed `running` exception is correctly reasoned: `StatusDot`'s tone set has no pulsing member and `tone="current"` alone would not pulse.
- **`OpenTargetMenu`'s role removal is a net gain, not a loss.** Before: `<div role="menu">` + `role="menuitem"` buttons with no arrow-key handler, no roving tabindex, no typeahead. AT entering a `menu` switches to application semantics and expects arrow navigation that does not exist; removing the roles leaves plain focusable `<button>`s that Tab reaches reliably. No keyboard path got worse. `SelectedResponseActionMenu.tsx:132,154` still carries the identical false pair — out of slice, residue for the chat slice.
- **Unexplained narrowing**, `OpenTargetMenu.tsx:49`: `icon: ReactNode` → `ReactElement`. Harmless (the one call site passes an element) but unrelated to §2-E.

### Verified clean

- **Deleted-component sweep.** `TerminalHeaderButton` / `ToolHeaderButton` / `ViewerHeaderButton` / `PaneSideOverlay` / `data-pane-side-overlay`: zero references across `apps/`, `tests/`, `specs/`, `scripts/`, `docs/` — only the two prose mentions in `PanelHeaderEntry.tsx:14-15`. `ui-tab-system-tab` as an entry root: zero.
- **Baseline edit is exact.** One key removed (`PaneSideOverlay.tsx|z-30`); no additions, no widened slack, no residue for the deleted header buttons in either census map.
- **Drag-region parity.** `data-right-panel-tab-no-drag="true"` moved from the close *container* div to the close *button*; `use-right-panel-header-drag.ts:112` matches with `closest("[data-right-panel-tab-no-drag='true']")`, so the exclusion still resolves from the glyph upward. `event.stopPropagation()` parity holds because `RowActionIconButton` does it internally.
- **ARIA / tab-stop parity.** No new tab stop — the close button was independently focusable on main too. `aria-controls` ids are passed through per kind, not regenerated. `aria-label` resolves the same for all three kinds (`title ?? label`). `data-active`, `data-reorderable`, `aria-grabbed`, `data-dragging` all land where they did, with `data-dragging` correctly absent on the tool entry (parity with `ToolHeaderButton`).
- **`PopoverButton` focus neutrality survives** (Risk 3). `TerminalHeaderIcon`'s `triggerMode="contextMenu"` path clones through Radix `Slot`; `PanelHeaderEntry` spreads rest props onto its inner `<button>` before setting its own `onClick={onSelect}`, so the ref and `onContextMenu` land and only `handleClick` is shadowed — a no-op here, since `stopPropagation` is false.
- **Roving-tabIndex floor** (addendum 2). Caller-side fallback is correct: `RightPanelHeaderEntryList.tsx:76-77` computes "no entry active" across the whole list and hands the floor to `entries[0]`, forwarded through `TerminalHeaderIcon`'s `tabIndexFloor` passthrough. `PanelHeaderEntry.tsx:126` withholds the floor from a `disabled` entry, and no call site passes `disabled`, so the floor cannot be lost.
- **Negative control run on the new keyboard test.** Reverting `tabIndex={active || (tabIndexFloor && !disabled) ? 0 : -1}` to `tabIndex={active ? 0 : -1}` makes *"keeps a roving-tabIndex floor…"* fail with `expected '-1' to be '0'` at `PanelHeaderEntry.test.tsx:84`; restored → 7/7 green. The test does assert the no-active-entry state (neither entry is `active`) and additionally proves `user.tab()` lands on the floored entry.
- **WP-B and WP-E really are no-ops upstream, not silent skips.** `shell/tabs/` no longer contains `ChromeWorkspaceTab.tsx` or `TabGroupPill.tsx`; the kit lives at `primitives/patterns/tabs/{ChromeTab,TabGroupPill}.tsx`. All nine in-slice sidebar files plus the two test mocks already import `#product/primitives/patterns/sidebar/*`.
- **Standing UI rule.** No colored left-border ownership treatment anywhere in the diff. Zero `lucide-react` imports added.

### Gates and tests (post-fix, at `45e15d27e`)

| | |
| --- | --- |
| `report_frontend_structure.py --strict` | pass (TOTAL: 0) |
| `check_docs.py` | pass (208 files) |
| `PanelHeaderEntry.test.tsx` | 7/7 |
| `tab-rendering.test.tsx` | 5/5 |
| `shell/right-panel/**` + `shell/screen/**` | 5 files, 26 tests, all pass |
| `ManualChatGroupEditorPopover.test.tsx` | 1/1 |

Not runnable here, all three from a missing `apps/packages/design/dist` in the donor `node_modules` — environmental, unrelated to this diff, and none of their inputs are touched by it: `library-registry.test.ts`, `appearance-css-drift.test.ts`, `ClosedChatTabsMenu.test.tsx`. No `tsc`, no build, per the machine constraints.

---

## Design sign-off required — @pablonyx

Three visible changes ship here and none can be verified without eyes on the running app:

1. **Right-panel tab close placement: leading-icon-overlay → trailing edge.** On main the close sat at `left: 0.64em`, overlaying the tab's icon (hidden on hover) and consuming zero extra width. It now sits after the label. Sanctioned by integrator addendum 1, but never seen.
2. **Close hit target 16px → 20px** (`size-icon-button-sm`). Also addendum 1.
3. **Not in the addendum, and worth a look:** because the close is composed as an inline sibling rather than `PanelHeaderEntry`'s own absolutely-positioned close, it *reserves layout width* inside the shell's 160px cap instead of overlaying — so terminal and viewer tabs have less room for their label than on main, and a long filename can push the close button past the shell's max-width (`PanelHeaderEntry`'s shell is `shrink-0`). The primitive's built-in close avoids this with `absolute end-1` + `pe-6`; the call sites cannot reach that without the drag-exclusion passthrough discussed in deviation 7.

Beyond the close button: the entries no longer take any paint from `.right-panel-tab-system`. The whole resting/hover/active/focus stack, the border radius (8px → `rounded-sm`), and the active `::before` fill now come from `PanelHeaderEntry`'s own Tailwind. Heights match exactly (`--tab-height: 28px` = `h-control`), but this is a repaint of the right-panel header strip, not a like-for-like port.

---

## Fixes pushed

One commit, `45e15d27e` — *fix(review): address review findings*:

1. `DelegatedAgentHoverCard.tsx` — restored the clickable mode (BLOCKER 2), with a note recording the failed premise.
2. `TerminalHeaderIcon.tsx:179`, `RightPanelHeaderEntryList.tsx:217`, `PlaygroundAttachmentFixtures.tsx:206` — added the `group` marker (BLOCKER 1).
3. `TerminalHeaderIcon.tsx:154`, `RightPanelHeaderEntryList.tsx:173` — reserved room for the shortcut badge (MAJOR 3).

Deferred, each with a home: MEDIUM 4 (max-width + `cursor: grab`) and MEDIUM 5 (dead `.right-panel-tab-system` rules) → the CSS-anatomy retirement slice, residue item 10. `SelectedResponseActionMenu`'s false menu ARIA → the chat slice. The `DelegatedAgentHoverCard` → `Tooltip` conversion stays open, and the PR body's doc-change block needs one correction: the hover card's clickable mode is a live consumer, not dead code.

PR #1747, opened 2026-08-11
fix(product): unify expanded tool-call detail boxes on the code-block card

pablonyx:
Fixes [PRO-120](https://linear.app/proliferate-team/issue/PRO-120/clicking-on-any-in-chat-tool-call-causes-an-ugly-box-to-appear) — clicking any in-chat tool call opened an ugly, inconsistent box.

## Problem

Two defects, one per commit:

**1. The pressed-state box (the literal complaint).** While actively clicking a tool-call title (e.g. "Workspace created"), a sharp-cornered `bg-active` rectangle flashes around the row. The transcript's quiet disclosure titles suppress the ghost Button's hover box with `hover:bg-transparent` but inherit its `active:bg-active`.

**2. Divergent expanded boxes.** The transcript had four expanded-tool-output box styles:

1. Shared `ToolActionDetailsPanel` — `border-border/60` wash on `bg-surface-elevated-secondary`
2. Inline duplicates of that style in `CollapsedCommandActionRow`, `CollapsedGenericActionRow`, `CollapsedEditActionRows` (drifted copies, not using the shared panel)
3. Bespoke `WorkspaceCreationReceipt` log with a full-strength `border-border` and non-wrapping `w-max whitespace-pre` lines that clipped long paths with no height clamp — the screenshot in the issue

## Fix

- `fdf4272d9`: suppress the pressed background wherever the hover background is already suppressed — `active:bg-transparent` beside every `hover:bg-transparent` on transcript/tool-call disclosure rows (17 sites, 11 files). `TurnMetadata`'s bordered chip keeps its pressed tint deliberately; `CollapsedActions`' summary row is `unstyled` and never leaked.
- `3c1c00daa`: restyle `ToolActionDetailsPanel` to the chat fenced-code-block card language from #1708: `overflow-clip rounded-lg border-transparent bg-[var(--color-code-block-background,var(--color-card))]`. The transparent border keeps size parity with bordered cards. Panel now spreads div props so callers can pass `id`/`data-*`. Fold all three inline copies onto the shared panel (drops their inner `border-t` dividers). Workspace-creation receipt log: wrapping lines (`whitespace-pre-wrap break-words`), `AutoHideScrollArea`, and the standard `TOOL_CALL_BODY_MAX_HEIGHT_CLASS` 224px clamp.

## Reviewer note

`TurnItemSequence.test.tsx` pins the panel's classes; the matcher is updated to the new surface (`overflow-clip` + code-block background var). This is an intentional expectation change, not a behavior change.

[...]

PR #1794, opened 2026-08-13
refactor(product): chat slice conformance cleanup

pablonyx:
Implements the frozen `chat/` conformance slice against the landed wave-1 vocabulary. Base is `lucide-pr-l` (`52944791e`), not `main`.

The headline of this PR is not the migrations. It is that **three of the spec's four big mappings cannot be satisfied by the components that actually landed**, and the diff records why at each site instead of forcing them. Those are decisions the founder should make before wave 3 widens anything.

---

## Commits

| # | SHA | What |
| --- | --- | --- |
| 1 | `744afce24` | `SelectedResponseActionMenu` → `DropdownMenu` (§ 2.1) |
| 2 | `eea5c090e` | `ComposerInlineMenu` frame + `PickerEmptyRow` (§ 2.2) |
| 3 | `39a978b00` | `IconTile` fold + `var()` paint retirement (§ 1.5, § 1.6) |
| 4 | `bb6c2a24f` | `NoticeBanner` + `Button` ghost state stacks (§ 3, § 1.7) |
| 5 | `1c47702eb` | `Card` where it fits, recorded exclusions where it does not (§ 2.3) |
| 6 | `d4ff35400` | One vertical rhythm per pane (§ 1.8) |
| 7 | `9d0fe881e` | **PRO-120 collision package** (§ 4 WP-4, § 6.1) |

39 files, all under `chat/`. Zero registry edits, zero new components, zero `DESIGN_SYSTEM.md` edits, zero baseline edits.

### Collision-package boundary

**Commit 7 (`9d0fe881e`) is the only commit that touches a file PR #1747 also touches.** It is last, self-contained, and nothing before it depends on it — drop it wholesale if #1747 lands first. #1747's `3c1c00daa` performs the same detail-panel fold, so the likely outcome is redundancy rather than conflict. Commits 1-6 touch zero #1747 files and rebase cleanly either way.

---

### 1. `Disclosure` cannot host the transcript's disclosure rows (§ 2.4 not delivered)

The spec maps thirteen files, fifteen machines, onto `Disclosure`. None of them can adopt it as landed:

- **It always paints `hover:bg-hover active:bg-active` on its header row, with no suppression.** `className` lands on the outer wrapper, not the row. Adopting it in the transcript would put a pressed background back on exactly the rows **PRO-120 removes one from** — the literal complaint in #1747, reintroduced at ~30 sites. This directly violates spec § 6.1 mitigation 3.
- It renders a rotating chevron; these rows carry a status glyph (`CollapsedActionIcon`, `StickyNote`, `Robot`) and no chevron at all.
- Its title is `text-heading` (17px); the transcript runs at 13-14px.
- Two of the sites are not disclosures in `Disclosure`'s sense at all. `CollapsiblePlanCard` fades its collapsed body under a mask and floats an "Expand plan" button over it rather than hiding content, and its toggle is an icon button paired with Copy in a card header.

Recorded in the diff on `ActionDisclosureRow` (the shared row all the tool-call sites use) and on `GoalTranscriptEventRow`. **Landing this family needs a quiet spelling of `Disclosure`** — a suppression prop, a chevron-less variant, or a text-scale slot. Adding one is forbidden by § 5, so I stopped.

### 2. `Card`'s two-fill surface axis expresses 2 of the 21 § 1.3 sites

`Card` offers `tint` (`bg-surface-elevated-secondary`, borderless) and `opaque` (`border-border bg-card`). The slice's card shells are `bg-card/70`, `/80`, `/85`, `/95`, `bg-muted/25`, `bg-muted/40`, `bg-background/70`, `bg-diff-panel-surface`, and one unfilled `border border-border` frame. Repainting a `Card` from the call site is the paint leak the doctrine exists to close, so the two exact matches migrated and the rest carry a recorded exclusion naming the missing fill.

Two sites have a second, independent blocker: the prompt-attachment cards deliberately overflow their frame (`overflow-visible`, for the remove affordance and drag ghost) and `Card` clips its children.

### 3. `ListRow` is a `<button>`, so it cannot host a row that contains a button

`input/PromptRecoveryPanel.tsx:39` maps to `ListRow`, but the row hosts its own Retry button — adopting it would nest interactive elements. The row is also not pressable; its lone `hover:bg-hover` drives a `group/recovery` reveal. Needs a non-interactive spelling of `ListRow`.

`plans/PlanHandoffDialog.tsx:97` maps to "`Card` selectable variant or `ListRow`". `Card` has no interaction states, and `ListRow`'s 17px title would resize a dialog row by a third. Its state stack went onto `Button variant="ghost"` instead, which owns the exact triple it was writing by hand.

### 4. PR A has not landed

`ComposerPopoverSurface` is still at `chat/composer/ComposerPopoverSurface.tsx`, not `primitives/patterns/`. § 2.2(a) is implemented against the current path; moving the file is PR A's job and would touch nine consumers outside this slice.

### 5. § 1.1 was already delivered by the base

`lucide-pr-l` retired every `lucide-react` import in `chat/` and added tuned twins for all four glyphs the spec said had none (`Hand`, `Quote`, `MessagesSquare`, `CircleCheck`). C5 is green with zero remaining identifiers, better than the spec's "at most four".

---

## Behaviour changes the founder should see

**ArrowLeft / ArrowRight no longer move between items in the selected-response menu.** This is a removal, not a refactor detail. Radix reserves Left/Right for submenus in a vertical menu and there are none here. Everything else — ArrowUp/Down with wrap, Home/End, Enter, Space, Escape, the pointer/keyboard focus discriminator — is preserved and now explicitly tested. Radix also adds typeahead, which the hand-rolled machine did not have, and focuses items on pointer-move, which it also did not.

**`NoticeBanner` top-aligns its action slot; `ClaimBanner` centred it.** The claim button now sits at the top of the banner rather than vertically centred. Visible, accepted, flagged per the integrator addendum.

**The destructive notice tone shifts** from the ad-hoc `border-destructive/40 bg-destructive/10` to the vocabulary's `border-destructive/30 bg-destructive-subtle`.

**Three tool-call detail panels move from `rounded-lg` to `rounded-md`**, joining the eleven consumers of `ToolActionDetailsPanel` that already used it.

**Two transcript blocks move from `rounded-md` to `rounded-lg`** by adopting `Card`, which owns its radius: the goal event detail and the system message body.

**The transcript column tightens from `space-y-1.5` to `space-y-1`** in the tool-call group blocks and the turn item sequence, matching the twelve stacks already on `space-y-1`.

**The plan-handoff row loses `hover:border-border`.** The fill change still signals hover.

### One-pixel note (§ 2.2 asks for an explicit call)

`POPOVER_FRAME_CLASS` carries `m-px`; `ComposerPopoverSurface` does not. **I kept the inset** — `className="m-px mb-2 overflow-hidden"` — so the slash-command and file-mention menus are pixel-identical to before. No screenshot needed for a no-op.

---

## Spec conformance

| Check | Status |
| --- | --- |
| **C1** new shape vs redraw | 2 sites compose `Card`, 2 compose `NoticeBanner`, 1 composes `IconTile`, 3 fold onto `ToolActionDetailsPanel`. Every remaining § 1.3 site carries a recorded exclusion at the site naming the fill `Card` lacks. |
| **C2** rule of two | Four detail-panel copies → one. Two `IconTile` sites → one component. `ComposerInlineMenuStatusRow` deleted for `PickerEmptyRow`. `TurnDiffPanel` and `TranscriptPatchTurnDiffPanel` NOT merged, as instructed. |
| **C3** overlay semantics | `grep 'role="menu"\|role="menuitem"'` returns zero non-test hits. `role="listbox"`/`role="option"` survive only in `ComposerInlineMenu`, with the exemption written at the site. No new overlay roles. Zero feature-side imports of `popover-surface` constants. |
| **C4** geometry / paint | `bg-[var(--color-diff-panel-surface)]` gone. The `--color-diff-chat-turn-icon-surface` brackets died with the `IconTile` fold. Every surviving bracket in a touched file carries a cause: the two-arg CSS fallbacks, the per-swatch custom property, `my-[14px]`, `tracking-[-0.01em]` ×2, `font-[520]`, the turn-diff row-padding contract, and the runtime-calculated menu anchor. |
| **C5** icon source | Zero `lucide-react` in `chat/`. Delivered by the base branch. |
| **C6** new-pattern quality | N/A — no new pattern. |
| **C7** state stacks | Three of the four § 1.7 raw stacks resolved (two onto `Button` ghost, one already ruled ACCEPT). The fourth is a recorded exclusion (`ListRow` nesting). Every § 6.3 ghost suppression is byte-for-byte unchanged. No new hover/active/focus triple on a raw element. |
| **C8** rhythm | Transcript column is one value (`space-y-1`). The composer stack's `space-y-0.5` pairs are title-to-description gaps inside a single row, not pane rhythm, and are left alone with that reasoning in the commit. Five distinct values → three. |

### § 2.1.4 keyboard acceptance — all ten criteria as explicit test cases

All in `SelectedResponseActionMenu.test.tsx`, 11 tests, green:

1. ✅ three items in order, reachable as `role="menuitem"`
2. ✅ `focusRequestNonce = 1` → first item highlighted after mount
3. ✅ `focusRequestNonce = 0` → `document.activeElement` unchanged
4. ✅ ArrowDown wraps last→first, ArrowUp wraps first→last, both edges asserted
5. ✅ Home → first, End → last
6. ✅ Enter and Space both activate with the right action (separate cases)
7. ✅ Escape calls `onEscape` once and `onAction` never
8. ✅ click runs the action and leaves `document.getSelection()` uncollapsed
9. ✅ `closest("[data-selected-response-actions]")` resolves from every item
10. ✅ `document.body` is not `pointer-events: none` while open

## Tests and gates

**All 59 `chat/` test files, 347 tests, green.** `library-registry.test.ts` green, unmodified.

Gates: `check_appearance_scaling.py`, `check_frontend_boundaries.py`, `report_frontend_structure.py --strict` (TOTAL 0), `check_max_lines.py`, `check_docs.py` — all green.

Three test files needed updating, all colocated with files this PR touches: the rewritten menu test, two class-string assertions re-truthed to `bg-diff-panel-surface` / `data-icon-tile`, and one `rounded-lg` → `rounded-md` in `TurnItemSequence.test.tsx` (inside the collision package, since that file is on #1747's list).

## Docs

No doc files are touched by this PR, per the integration contract. The index-row rewordings, incubating-note removals for `Card` / `Disclosure` / `IconTile` / `NoticeBanner`, and the Current Gaps strikes are handed to the integrator as exact old→new line data — see the implementation report.


pablonyx (commented):
## Adversarial review — chat slice (`237e675a2..63a7df741`, fixes at `81c3222f9`)

Reviewed the full delta against the frozen spec plus the integrator addenda, with negative controls on the keyboard work and the collision boundary. **All three recorded blockers are truthful** — I verified each against the landed component source, not the PR body. Gates 5/5 green, `chat/` 60 files / 351 tests green, `library-registry` 4/4 unmodified.

Fixes pushed as one commit `81c3222f9` that deliberately touches **zero** of PR #1747's fifteen files, so `22ca6684b` keeps its droppable isolation.

---

### HIGH

**H1 — The §2.1.4 criterion-8 test could not fail. FIXED.**
`SelectedResponseActionMenu.test.tsx:130-141` ("runs the action on click without collapsing the live text selection") asserted `document.getSelection()?.isCollapsed === false`. jsdom never collapses a selection on pointerdown, so the assertion is unconditional. Negative control: deleting `onPointerDown={(event) => event.preventDefault()}` from `SelectedResponseActionMenu.tsx:172` left **all 11 tests green**. That preventDefault is §2.1.2 behavior 15 — the single line that keeps the window selection alive long enough for `handleAction` to read `selection.text` — and the test written to protect it protects nothing.
Fixed: the test now dispatches a cancelable `PointerEvent` and asserts `defaultPrevented`, which fails when the guard is removed (verified both directions).

For contrast, `loop` and `modal={false}` **are** genuine negative controls — removing either fails tests 4 and 10 respectively. Only criterion 8 was hollow.

---

### MEDIUM

**M1 — `TurnMetadata.tsx:74` was the one §1.3 card shell with neither a migration nor a check-1 exclusion. FIXED.**
The comment added at `:68-71` covers check 7 only; the `rounded-md border border-border/50 bg-muted/50` triple survives uncommented. §5 C1 requires every §1.3 site to migrate or carry a comment at the site. Extended the comment to name the missing fill.

**M2 — Undisclosed visual change: the plan-handoff icon tile grows 28px → 32px.**
`PlanHandoffDialog.tsx:105` replaced `<span className="grid size-7 … rounded-md bg-surface-control">` with a bare `<IconTile>`, whose default `size="md"` is `size-8 rounded-md`. The attached-plan row gets 4px taller in a dialog. Two things about this: it is not disclosed in the PR body's behavior list, and this tile is **not** in the spec's §1.5 mapping — §1.5 names only the two `size-10` transcript tiles and explicitly says the related-but-different sizes are "do not fold in without a review ruling". `IconTile`'s own docstring does license rounding stray 28px one-offs to the nearest step, so the fold is defensible; the silence is not. **Deferred to Pablo** (see the founder section) rather than reverted.

**M3 — Undisclosed layout change in `ClaimBanner`, and two new live regions.**
`NoticeBanner` puts the icon in a leading gutter column spanning title *and* body (`NoticeBanner.tsx:92-96`), where `ClaimBanner` had it inline before the title with the description flush-left underneath. After the migration the description indents right by roughly icon-width + gap. The PR body discloses only the action's top-alignment.
Separately, both branches now inherit `role="status"` and `McpElicitationInlineError` inherits `role="alert"` from the tone default — neither element announced before. Benign to positive, but new announced behavior.

**M4 — `DESIGN_SYSTEM.md`'s `Card` row was materially inaccurate. FIXED.**
It said "the remaining 19 need a fill the two-value `surface` axis does not carry … or need to overflow their frame." Of those 19: three went to `NoticeBanner` (they are notices, not fills `Card` lacks), three folded onto `ToolActionDetailsPanel`, `WorkspaceCreationReceipt` needs *no* fill rather than a different one, and `PlanHandoffDialog` is blocked on interaction states. Re-truthed to 2 / 3 / 3 / 13.

**M5 — "Redundant rather than conflicting" is only half true.** Verified against `origin/pablo/pro-120-…`.
The *fold* is genuinely the same in both branches. The *paint* diverges:

| | this branch (`22ca6684b`) | #1747 (`3c1c00daa`) |
| --- | --- | --- |
| `ToolActionDetailsPanel` | `overflow-hidden rounded-md border-border/60 bg-surface-elevated-secondary` + twMerge + attr passthrough | `overflow-clip rounded-lg border-transparent bg-[var(--color-code-block-background,…)]` |
| `TurnItemSequence.test.tsx:94-104` | asserts `rounded-md` + `border-border/60` | asserts `overflow-clip` + `bg-[var(--color-code-block-background` |
| `CollapsedCommandActionRow` scroll area | keeps `border-t border-border/60` | removes it |

Whichever lands second gets a textual conflict in three files, and #1747's paint (the whole point of that PR) must win. Worth saying plainly in the commit message so the rebaser does not resolve it the wrong way.

**Isolation itself: VERIFIED CLEAN.** `22ca6684b` touches 8 of #1747's 15 files; commits 1-6 and the docs commit touch none of them, and neither does my fix commit. The one thing that changed since the PR body was written: the collision commit is no longer *last* — `63a7df741` (docs) and now `81c3222f9` (fixes) sit after it. Both are independent of it, so it stays droppable, just no longer by `HEAD~1`.

**M6 — New containing-block dependency on the menu anchor, untested. Comment added.**
The old virtual anchor returned raw viewport coords with no DOM involvement. The synthetic trigger uses `position: fixed`, which resolves against the nearest ancestor carrying `transform` / `filter` / `backdrop-filter` / `perspective` / `contain` / `will-change: transform`. I walked ChatView → SessionTranscriptPane → MessageList → ChatTranscriptView: **none today**, so it works. But any future `backdrop-blur` on a transcript ancestor silently detaches the menu from the selection with no test to catch it. Recorded the dependency at the site.
Also: §6.5's manual check ("scroll the transcript with the menu open") is not evidenced anywhere — no test, no recording. Flagging as unverified rather than broken.

---

### LOW

- **L1** PR body says rhythm went "five distinct values → three"; four remain (`space-y-0.5`, `-1`, `-2`, `-3`). The `space-y-3` at `plans/PlanReferencePreviewDialog.tsx:31` survives, so the commit message's "retires the fifth distinct value in the slice" is also wrong. The dialog is arguably a third pane and out of the two named ones — the claim is what's off, not the code.
- **L2** The two `IconTile` adoptions repaint the tile from the call site (`bg-diff-chat-turn-icon-surface text-secondary-foreground` overriding the `tone` recipe) — exactly the move the `Card` exclusions refuse on paint-leak grounds — with no comment. Both defensible; the inconsistency was unannotated. **FIXED**: cause recorded at `TurnDiffPanelHeader.tsx:45` and `TurnDocumentReferenceCard.tsx:27`.
- **L3** `CollapsedActionRowPrimitives.tsx:46-48`: "`rounded-none bg-transparent p-0 hover:bg-transparent focus-visible:ring-0` is what stops a sharp-cornered rectangle appearing around a transcript line **on hover**" — that stack suppresses hover only; the pressed rectangle is precisely what it does *not* suppress, which is why #1747 adds `active:bg-transparent`. The next paragraph says so correctly. **Not fixed here** — the file is in the collision package and touching it would break the isolation. Fold into `22ca6684b` or into #1747's rebase.
- **L4** `PromptAttachmentCard`'s "the four shells in this function" comment sat below the first shell. **FIXED** — moved above the first return.
- **L5** `role="option"` (`ComposerInlineMenu.tsx:109`) carried no recorded cause; §5 C3 asks for it on both roles. **FIXED**.
- **L6** `DropdownMenuItem` promotes its glyph with opacity (`[&_svg]:opacity-75` → `data-[highlighted]:opacity-100`), while `PopoverMenuItem` — the row this callsite left — uses *color* promotion specifically because styling.md bans "No partial-opacity hover transitions on glyphs" (the ban is quoted verbatim in `PopoverMenuItem.tsx:38-46`). The migration therefore moves this menu onto the treatment the doctrine forbids. Library-owned, out of slice scope. **Deferred: `primitives/DropdownMenu.tsx`, wave-3 library pass.**
- **L7** Small undisclosed deltas at the menu: rows lose `transition-colors` so hover paint is instant; content gains `animate-popover-in` and `overflow-hidden`; width goes `w-[220px]` → `min-w-[220px]` (identical at three short labels, can now grow).
- **L8** On activation `onDismiss` now fires **twice** — once from `handleAction`, once from Radix's `rootContext.onClose()` → `onOpenChange(false)`, because `onSelect` does not `preventDefault` (`@radix-ui/react-menu@2.1.24` `MenuItem.handleSelect`). Harmless today: `dismissSelectedResponse()` with no options is idempotent. Worth knowing before anyone makes that callback non-idempotent.

---

### Verified-and-clean (claims I tried to break and could not)

- **Blocker 1 (`Disclosure`) TRUE.** `Disclosure.tsx:87-89` hardcodes `hover:bg-hover active:bg-active` on the header row; `className` lands on the outer `min-w-0` wrapper (`:85`); the title is `text-heading`; the chevron is unconditional. Adopting it on the transcript's quiet rows would reinstate the PRO-120 rectangle at ~30 sites. Not delivering §2.4 was correct.
- **Blocker 2 (`Card`) TRUE.** `Card.tsx:73-75` — two surfaces, no alpha, `overflow-clip` always, `className` documented layout-only. Every sampled exclusion names a concrete missing fill: `SelectedResponseContextList` (`bg-muted/40`), `ChatSurfaceCard` (`bg-card/95`), `CoworkArtifactTurnCard` (`bg-muted/25`), `SessionErrorItem` (`bg-background/70` recessed well), `TurnDiffPanel` / `TranscriptPatchTurnDiffPanel` / `TurnDocumentReferenceCard` (`--color-diff-panel-surface`), `CollapsiblePlanCard` + `PlanReferencePreviewDialog` (`bg-card/85`), `PromptAttachmentCard` (overflow + /70,/80). No vague TODOs anywhere. 8 sampled, 8 concrete.
- **Blocker 3 (`ListRow`) TRUE.** `ListRow.tsx:22-28` is a `<button>` with its slots inside and a `text-heading` title. Hosting `PromptRecoveryPanel`'s Retry button in it would nest interactive elements.
- `aria-labelledby={undefined}` — verified against the installed `@radix-ui/react-dropdown-menu@2.1.24`: `"aria-labelledby": context.triggerId` is emitted **before** `...contentProps`, so the undefined override wins. It is also negatively controlled by the Escape test's `findByRole("menu", { name: … })`.
- `onOpenAutoFocus` / `onCloseAutoFocus` preventDefault correctly suppress Radix's own content-focus and trigger-refocus via `composeEventHandlers`' `checkForDefaultPrevented`. Hazards (c) and (d) handled.
- Escape, outside-dismissal and Tab-out semantics are equivalent to the old non-modal `Popover` path. `data-selected-response-actions` preserved and covered.
- **ComposerInlineMenu frame is byte-equivalent.** Old `twMerge(POPOVER_FRAME_CLASS, "mb-2 overflow-hidden p-1")` and new `ComposerPopoverSurface` + `"m-px mb-2 overflow-hidden"` resolve to the identical class set. The "pixel-identical, no screenshot needed" call is correct. `PickerEmptyRow` is byte-identical to the deleted `ComposerInlineMenuStatusRow`.
- **`TurnMetadata` "pixel-identical" TRUE.** `Button` ghost is `text-muted-foreground hover:bg-hover hover:text-foreground active:bg-active` — exactly the four classes deleted.
- **Every §6.3 ghost suppression is untouched.** Grepping the whole delta for `hover:bg-transparent` / `active:bg-active` / `focus-visible:ring-0` / `hover:bg-hover` line changes returns only the two §1.7 raw stacks and comment text. No normalization crept in.
- `IconTile` fold: `flex` → `inline-flex` is a no-op (both sites are flex items, which blockify), `size-10 rounded-lg` matches, `bg-diff-chat-turn-icon-surface` is a real `@theme` utility.
- C3 zero `role="menu"`/`menuitem`, C5 zero `lucide-react`, zero feature-side `popover-surface` imports, zero registry edits, zero `src/generated/` committed, no temporary vitest config committed.

*(An `INFO` for the record: §2.2(i)'s leak is relocated rather than closed — `ComposerPopoverSurface` is itself feature code in `chat/composer/` that hand-pastes the popover recipe. Correct per PR A's contract; noting it so nobody reads that line as done.)*

---

## For Pablo: behavior changes + founder rulings

**Three rulings the implementer recorded instead of forcing. I checked all three against the landed source; all three are true.**

1. **`Disclosure` cannot host the transcript's 15 disclosure machines.** It paints `hover:bg-hover active:bg-active` on its header with no suppression, and `className` reaches only the outer wrapper. Adopting it would put the pressed rectangle back on exactly the rows #1747 removes it from — the literal PRO-120 complaint, at ~30 sites. It also forces a chevron where these rows carry a status glyph, and a 17px title onto 13-14px rows. **§2.4 is not delivered, and that was the right call.** Landing this family needs a quiet spelling of `Disclosure` (a suppression prop, a chevron-less variant, a text-scale slot) — your ruling, not a call-site workaround.
2. **`Card`'s two-fill surface axis expresses 2 of 21 chat shells.** Everything else is `bg-card/70…/95`, `bg-muted/25`, `bg-muted/40`, `bg-background/70`, `--color-diff-panel-surface`, one unfilled frame, and two cards that must overflow (`Card` always clips). 13 sites carry a recorded exclusion naming their specific blocker. **Widening `Card` is your call.**
3. **`ListRow` is a `<button>`,** so it cannot host `PromptRecoveryPanel`'s row (which contains a Retry button) without nesting interactive elements, and its `text-heading` title would resize a dialog row by a third. Needs a non-interactive spelling.

**Behavior removals and visual deltas — the PR body's list plus what it missed.**

Disclosed and confirmed: ArrowLeft/Right no longer move between menu items (Radix reserves them for submenus); Radix adds typeahead and pointer-move highlighting the hand-rolled machine did not have; `NoticeBanner` top-aligns the claim button; the destructive notice shifts to `border-destructive/30 bg-destructive-subtle`; three tool-call panels `rounded-lg` → `rounded-md`; two transcript blocks `rounded-md` → `rounded-lg`; the transcript tightens `space-y-1.5` → `space-y-1`; the plan-handoff row loses `hover:border-border`.

**Missing from that list — please add before merge:**

- The plan-handoff dialog's attached-plan **icon tile grows 28px → 32px**, so that row gets 4px taller. This tile was not in the spec's fold list; §1.5 said sizes other than `size-10` need a review ruling. **Your call: keep the rounding, or pin `size="sm"` (24px) / revert to the raw 28px span.**
- **`ClaimBanner`'s `Hand` glyph moves into a leading gutter column**, so the description indents right by roughly icon-width + gap. Separate from the disclosed action alignment.
- **Two new live regions**: the claim banner now announces as `role="status"`, the MCP elicitation error as `role="alert"`. Neither announced before.
- The menu's rows lose their hover transition (instant now), the panel gains an entrance animation, and glyph hover promotion switches from color-based to **opacity-based** — the treatment styling.md explicitly bans for glyphs, inherited from `DropdownMenuItem`. Library-owned; worth a wave-3 item.
- On activation the menu now calls `onDismiss` twice (harmless today).
- **`space-y-3` still exists** at `plans/PlanReferencePreviewDialog.tsx:31`; the "five values → three" line is wrong (it is four).

**Unverified, needs a human:** §6.5's scroll-the-transcript-with-the-menu-open check, and §6.7's disclosure-motion recording — the latter is moot since §2.4 was not delivered.

---

Gates: `check_appearance_scaling.py`, `check_frontend_boundaries.py`, `report_frontend_structure.py --strict` (TOTAL 0), `check_max_lines.py`, `check_docs.py` — all green after the fixes. `chat/` 60 files / 351 tests + `library-registry` 4/4 green.

Not approving — merge is yours.

PR #1795, opened 2026-08-13
refactor(product): workflows/activity/files/git — component hierarchy conformance slice

pablonyx:
## Summary

Frozen-spec slice implementing `slice-specs/workflows.md` against the component hierarchy doctrine (`specs/DESIGN_SYSTEM.md`). Four work packages, one PR:

- WP-1 `components/workflows/`
- WP-2 `components/workspace/activity/`
- WP-3 `components/workspace/files/`
- WP-4 `components/workspace/git/`

Base is `lucide-pr-l` (this branch was rebased onto its current tip after that branch picked up a "address review findings" commit post-fork; the rebase was clean, zero file overlap with anything touched here).

## Conformance checklist (spec §6)

- **C1/C2 (duplication → promotion, second instance adopts)**: RosterRow (roster rows across activity + workflows + git), Card (workflow/activity surfaces), NoticeBanner (workflow + git notices), SegmentedControl, PopoverSearchField (files + git, 3rd instance), PopoverMenuItem (git branch/base selectors + jump-to-file), SidebarRowSurface (file tree rows), Disclosure (WorkflowRunDetail, one of five state machines named in spec — see deviations), StatusDot (activity + workflows tone dots).
- **C3 (hand-rolled overlay/ARIA semantics)**: PublishDialog's `role="listbox"`/`role="option"` (announced, unimplemented pattern) dropped for a plain labeled container with native button rows; GitReviewFileSectionShell's `role="button"` div + hand-rolled keydown replaced with a real `<button>`.
- **C4 (geometry escape hatches)**: every previously-uncommented arbitrary bracket in the slice's file list now carries a comment recording the legitimate cause, or was removed.
- **C5 (icon source)**: already clean on this base (lucide-react retired prior to this slice); zero lucide imports remain in any of the four areas (grep-verified).
- **C7 (hand-assembled state stacks)**: non-negotiable removals from spec done (`WorkflowDefinitionList.tsx` missing `active:`, three roster-row stacks, `GitReviewTargetSelector.tsx` row stack, `FileTreeRow.tsx` stack, both `FileTreeOverlay` focus-visible stacks — one gone with its field's conversion, one kept with justification per deviations below). Stacks the spec explicitly permits to survive with justification (no vocabulary target in this slice) are unchanged: `GoalBarIconAction.tsx`, `FILE_VIEWER_TOOLBAR_BUTTON_CLASS`, `REVIEW_HEADER_ACTION_CLASS`, `GitReviewEmptyStateAction`, `GitLastTurnUndoAction`'s Button override, the shared git trigger constant.
- **C8 / registry integrity**: zero edits to `components/playground/library/**`, zero new library components registered, zero baseline JSON edits.

## Deviations (recorded, not silently resolved)

1. **FileTreeRow.tsx / FileSearchResultsTree.tsx — Disclosure not adopted for the chevron.** Both implement a `treeitem`/`tree` ARIA pattern; `Disclosure`'s contract is a standalone trigger+content pair, a shape mismatch with tree-role semantics. Kept native, documented inline.
2. **GitReviewFileSectionShell.tsx — Disclosure not adopted.** `Disclosure`'s hardcoded trigger paint (`rounded-lg px-2 hover:bg-hover active:bg-active`, no override slot) can't host this header's sticky, full-bleed, near-opaque `color-mix` background without breaking that visual or fighting the pattern's own paint with a specificity override (itself a forbidden escape hatch). Adopted what the underlying C3 finding actually needs: native `<button>` replacing the `role="button"` div + hand-rolled keydown, and removal of the redundant second "Toggle file diff" trigger. The chevron becomes a decorative always-visible indicator, matching `Disclosure`'s own chevron convention — the one piece of that pattern's behavior this file can adopt.
3. **WP-2 PanelHeaderEntry** — four static, non-interactive activity-panel section labels don't fit `PanelHeaderEntry`'s tablist-only shape (no static variant exists); left as-is, documented.
4. **WorkflowRunForm.tsx Card fill** and **WorkflowDefinitionList.tsx line-clamp** — carried forward from the WP-1 commit as documented deviations (see that commit message for detail).
5. **Test structural changes** (all confirmed passing): `WorkflowRuns.test.tsx` (`<details>`→Disclosure, asserts `aria-expanded` not `open`), `FileTreeOverlay.test.tsx` (clear-filter button → `PopoverSearchField` has no clear slot, asserts via `fireEvent.change`), `PublishDialog.test.tsx` (`role="option"` → `role="button"`, `.disabled` → `aria-disabled`, `aria-selected` → `aria-current`).
6. **Follow-up fix, not part of the four WPs**: `domain/workflows/run-presentation.ts` imported `StatusDotTone` from a UI primitive, tripping `check_frontend_boundaries.py` (domain modules may only import Cloud SDK / AnyHarness contract types + pure helpers). This was introduced by the earlier WP-1 commit in this same slice. Moved the tone-map and `workflowRunStatusDotTone()` into a new `components/workflows/workflow-run-status-dot.ts`, mirroring the existing `SubagentRosterRow`/`TerminalRosterRow` local-map pattern. Gate now green.

## Card-fill sign-off (Integrator addendum, flagged per its explicit instruction — no code change in this slice)

- Diff-panel token is **3%** vs `surface-elevated-secondary` **4.9%** in light mode — a slight visible shift wherever a future `Card` fill lands on this token.
- `FileDiffCard`'s 97% color-mix sticky ground (mirrored by `GitReviewFileSectionShell.tsx` and this PR's reshaped `GitPanel.tsx` loading skeleton) loses 3% translucency under that same substitution, and never gets `backdrop-blur` regardless — a deliberate WKWebView-compositor concession, commented at each site.

## Promotion candidates (not actioned this slice)

- `LoopsPanel.tsx`'s native/emulated chip vs the git area's new `GitReviewCountChip` — same shape, different areas; merging is a promotion, out of scope here.
- The two git selectors (`GitReviewBaseSelector`/`GitReviewTargetSelector`) share a trigger constant and chip helper as of this PR; whether they should further collapse into one component (static list vs. searchable branch list is a real behavioral difference) is recorded per spec §5.D, not decided here.
- Activity's roster-panel triplicate (spec-recorded, unresolved).

## Doc-change data (apply directly to `specs/DESIGN_SYSTEM.md` — NOT applied by this PR per the integration contract)

- `PanelHeaderEntry` index row: strike the incubating note. Old: `...Promoted from four hand-rolled header buttons. Incubating: the shell slice (wave 2) migrates those four onto it; no non-playground call site until then.` New: `...Promoted from four hand-rolled header buttons.` *(Only if this slice's four WPs are the shell slice consumer — confirm against the shell-slice PR before applying; if the shell slice is a separate branch, leave incubating.)*
- `Card`, `RosterRow`, `NoticeBanner`, `SegmentedControl`, `SidebarRowSurface`, `PopoverSearchField`, `PopoverMenuItem` index rows: remove any "incubating"/"no non-playground call site" language where this PR's WP-1–4 call sites are the first non-playground consumer named in that row's current text (verify each row individually — several already had other consumers before this slice).
- `specs/DESIGN_SYSTEM.md` "Current Gaps" section: strike the workflows/activity/files/git hand-roll entries this slice resolves (see spec §1.1–1.4 file inventory for the exact list of what was fixed vs. deferred).

## Test results

Passing (targeted, colocated, staggered per-file per machine constraints):
- `WorkflowRuns.test.tsx` 4/4
- `runs/WorkflowRunsSurface.test.tsx` 14/14
- `ActivityChips.test.tsx` 4/4, `GoalBarResultPopover.test.tsx` 6/6, `GoalBarObjectiveEditor.test.tsx` 9/9, `LoopsPanel.test.tsx` 7/7, `GoalBar.test.tsx` 11/11
- `FileTreeOverlay.test.tsx` 7/7, `PromptAttachmentViewer.test.tsx` 4/4
- `PublishDialog.test.tsx` 7/7

Known-failing, pre-existing environment gap (NOT a regression — confirmed via cross-file grep that 20+ files across the whole app hit this, none edited by this slice): `FileEditorView.test.tsx`, `FileViewerContent.test.tsx`, `PromptAttachmentViewer.identity.test.tsx`, `GitPanel.test.tsx` all fail to resolve `@proliferate/product-client/host/ProductHostProvider` — a self-package-reference import that needs a built `dist/` this ad hoc symlinked local environment doesn't have (and building it was declined per this machine's resource limits).

Gates: `check_frontend_boundaries.py`, `check_max_lines.py`, `check_docs.py`, `check_design_attribution.py`, `check_theme_contrast.py`, `check_toast_copy.py` all pass.


pablonyx:
## Review — PR #1795 (workflows / activity / files / git conformance slice)

Reviewed the full delta `237e675a2...3c83553a6`. Verdict: the vocabulary adoption is faithful and the two recorded contradictions check out against the code as landed. Findings below are mostly behavior-preservation leaks in the migrated call sites, plus one factually wrong justification comment.

**Gates at head `3c83553a6` — all five green** (re-run in a clean worktree):
`check_appearance_scaling.py` ✅ · `check_frontend_boundaries.py` ✅ · `report_frontend_structure.py --strict` ✅ (TOTAL 0) · `check_max_lines.py` ✅ · `check_docs.py` ✅

**Tests at head:** `PublishDialog` 7/7, `WorkflowRuns` 4/4, `WorkflowDefinitionEditor` 4/4, `ActivityChips` 4/4, `GoalBar` 11/11, `GoalBarObjectiveEditor` 9/9, `GoalBarResultPopover` 6/6, `LoopsPanel` 7/7, `FileTreeOverlay` 7/7. `GitPanel.test.tsx` known pre-existing `host/*` dist resolution failure — not a finding, not touched.

**Contract checks:** zero registry edits ✅ · zero `components/playground/library/**` edits ✅ · zero `specs/DESIGN_SYSTEM.md` edits ✅ (the `Card`/`NoticeBanner`/`RosterRow` index rows never carried incubating notes — only the source comments did, and all three now have real consumers in this diff, so the strikes are correct) · zero baseline JSON edits ✅ (nothing deleted, so none were owed) · zero lucide identifiers anywhere in the diff ✅ · C3 grep over the four areas returns only comments/tests ✅.

**Recorded contradictions verified as factually accurate:**
- `GitReviewFileSectionShell` — `Disclosure.tsx:84-86` does hardcode `rounded-lg px-2 hover:bg-hover active:bg-active` on the trigger row, and `className` lands on the outer wrapper, not the row. Claim holds. ✅
- `LoopsPanel` / WP-2 `PanelHeaderEntry` — `PanelHeaderEntry.tsx:107-112` does hardcode `role="tab"` / `aria-selected` / roving `tabIndex`, no static variant. Claim holds. ✅
- `FileTreeRow` / `FileSearchResultsTree` treeitem argument — holds. ✅

---

### Findings

**1. MEDIUM — `TerminalRosterRow.tsx:39` silently collapses two distinct tones.** `PROCESS_STATUS_DOT_TONE` maps the domain `muted` tone to `StatusDotTone` `"muted"`, whose ink is `text-muted-foreground` — but the old `TONE_CLASSNAME` painted `muted` as `text-faint`. `default` was already `text-muted-foreground`, so `default` and `muted` processes are now visually identical. `StatusDot`'s tone axis has no `faint` step, so this is a real (unrecorded) regression, not a drift cleanup. Fixing by preserving `text-faint` for that one tone with a recorded comment.

**2. MEDIUM — `WorkflowDefinitionList.tsx:75` drops `line-clamp-2` on the description.** `RosterRow`'s `secondary` slot (`RosterRow.tsx:154`) has no truncate/clamp; the old markup had `line-clamp-2 text-ui-sm leading-4`. A long workflow description now grows the row unboundedly. (The PR body says this deviation is "documented in the WP-1 commit message" — it isn't; neither is the `WorkflowRunForm` Card-fill one.) Fixing by clamping in the slot.

**3. MEDIUM — `GoalBar.tsx:255-262`: the justification comment is factually wrong.** It claims `bg-surface-elevated-secondary` "sits at 3% in dark / 4.9% in light — within rounding error" of `color-mix(in oklab, var(--color-foreground) 2%, var(--color-background))`. Token values (`apps/packages/design/src/tokens.ts:710-713`): dark `3%`, light `rgba(26,28,31,0.049)` = **4.9% vs the original 2%, ~2.5×** — and the token is *translucent* where the original mixed against `--color-background` opaquely. That is not rounding error, and the integrator addendum flagged the same 3%/4.9% delta independently. Frozen spec §2.6 rules this exact case: *"if none matches within a rounding error, this is a token gap and belongs to `design`, so record it and leave the bracket with a comment rather than eyeballing a substitute."* Restoring the bracket + recording the token gap (see FOUNDER-A if you'd rather take the shift).

**4. MEDIUM — `PublishDialog.tsx:203`: `aria-label` on a roleless `div` is inert, and the test lost its scope.** Dropping `role="listbox"` was correct, but `aria-label="Source control action"` on a generic container is ignored by AT — the label is now dead. Worse, `PublishDialog.test.tsx:222` went from `getAllByRole("option")` (exactly the command rows) to `getAllByRole("button")`, which now matches every button in the dialog; the "locks intent switching" assertion and the `[0]` `aria-current` assertion are both matching an unscoped set. Fixing with `role="group"` on the container (no keyboard contract, not on C3's forbidden list, makes the existing label real) and re-scoping both assertions with `within(...)`. Negative-controlled.

**5. MEDIUM — `GitReviewFileSectionShell.tsx`: `data-app-action-review-file-toggle=""` was deleted with the redundant chevron button.** That handle is still carried by the two sibling diff surfaces (`chat/transcript/TurnDiffFileRow.tsx:39`, `content/ui/FileDiffCard.tsx:212`) and asserted in their tests, so the git review pane is now the only diff surface without it. The new header button already carries `data-app-action-review-file-expanded`; re-adding the toggle handle to it restores parity for one attribute. (Removing the redundant second trigger itself is right — good call.)

**6. LOW — `PublishDialog.tsx:376`: leading icon slot lost `text-muted-foreground`.** The old `<span className="flex w-[18px] ... text-muted-foreground">` kept the command glyphs muted against the row label; the new slot keeps `w-[18px]` and drops the ink, so glyphs now paint at row foreground. Fixing.

**7. LOW (report only) — `FileTreeOverlay.tsx:181` adds a hand-rolled `<div className="h-px shrink-0 bg-border" />`.** A new one-line hand-roll in the area this slice is de-hand-rolling, and neither sibling consumer of `PopoverSearchField` (`GitPanelHeader`, the new `GitReviewTargetSelector` call) draws one. Root cause is a library defect, not this slice's: `PopoverSearchField`'s own docstring claims it renders "a hairline divider below" and the implementation has none. Worth an issue against the primitive rather than a change here.

**8. LOW (report only) — `LoopsPanel` delete action tooltip changed.** `RowActionIconButton` derives `title` from `label`, so the hover tooltip went from `"Delete loop"` to `"Delete loop: <the loop prompt>"` — user prompt text now surfaces in a native tooltip — and the hit target grew from `IconButton size="xs"` to the fixed 28px row-action size. Not fixable without widening `RowActionIconButton` (it `Omit`s `title`), so recording rather than patching. `visibility="always"` is the right choice here and `RosterRow`'s root does carry `group`, so the hover-reveal trap from the shell slice does not apply anywhere in this diff — checked every action slot.

**9. LOW (report only) — nested `Card` tone in `WorkflowRunForm.tsx:97`.** Spec §2.1 rules sites 11–15 to "`Card` at its nested/quiet tone"; the per-input rows use `surface="opaque"`, which adds `bg-card` and moves the radius `rounded-md` → `rounded-lg`. The fill is invisible on a `bg-card` parent, so this is radius-only drift. Leaving as landed.

**10. LOW (report only) — `StatusDot.tsx` gained a public export from a feature slice.** `statusDotToneTextClass` clears C6 literally (not a new component, not a registry edit) and the "one owner for tone→ink" argument is good, but it widens a library primitive's API with no index-row mention and no colocated test — `StatusDot` has no test file at all. Suggest the library owner picks this up rather than this slice.

**11. LOW (report only) — `WorkflowDefinitionList` rows went full-bleed → inset.** `Card surface="opaque" className="flex flex-col gap-0.5 p-2"` replaces the old zero-padding container with `border-t` hairlines between rows. Sanctioned by spec §5.B's "the divider logic is `Card`'s job", but `Card` draws no dividers, so they're simply gone. Visible design change; flagging so it isn't discovered later as drift.

---

### FOUNDER items (not fixed — visual/scope judgment)

**FOUNDER-A — `GoalBar` fill: keep the token shift, or keep the pixels?** Finding 3 restores the original bracket per spec §2.6's own conditional. If you'd rather retire the callsite `color-mix` and accept a ~2.5× stronger tint on the goal bar in light mode, say so and it flips to the token in one line (the comment then needs the real numbers, not "rounding error").

**FOUNDER-B — the git review rail's notices got noticeably heavier.** `GitLastTurnUndoAction` / `GitReviewDiffPolicyNotice` went from `border-border/70 bg-surface-elevated-secondary px-2.5 py-2 text-ui-sm leading-5 text-sidebar-muted-foreground rounded-md` to `NoticeBanner` neutral = `border-border bg-card text-foreground p-3 text-ui rounded-lg`. Bigger, brighter, and content-plane ink inside a sidebar-token pane. Spec §2.3 does rule the radius/hairline drift away, but the ink + size jump in a narrow rail is a look decision, and the integrator addendum asked for an explicit sign-off line on git-area fills. Your call.

**FOUNDER-C — `NoticeBanner` forces a live region on four sites that had none.** `NoticeBanner` resolves `role={role ?? tone.role}`, so `undefined` cannot suppress it. Newly announcing: `WorkflowDefinitionEditor` issues banner (`role="alert"`), `WorkflowInputEditor` "This workflow has no inputs." (`role="status"`), and both git chrome notices (`role="status"`). The last two are persistent panel chrome, not transient messages, so they will announce on appearance every time the review pane loads. The fix is a `role={null}`/`"none"` escape in the primitive, which is library work outside this slice.

**FOUNDER-D — `FileTreeOverlay` lost its "Clear file filter" button.** Spec §2.4 directed the `PopoverSearchField` swap and the primitive has no trailing-action slot, so the affordance is gone (clear by selecting-and-deleting). Documented at the site and in the test. Accept, or add a clear slot to `PopoverSearchField` in a library follow-up.

**FOUNDER-E — PR body accuracy.** Two deviations are attributed to the WP-1 commit message and aren't there (findings 2 and 9), and the gate list omits the two gates that actually caught the late violations (`report_frontend_structure.py --strict`, `check_appearance_scaling.py`). Worth a body edit before merge.

---

Fixing findings 1–6 on the branch now; will push and report the new head.

pablonyx:
### Fixes pushed — head is now `9b099bf24`

Commit `9b099bf24` `fix(product): review fixes — behavior leaks in the conformance slice migrations` applies findings 1–6:

| # | File | Fix |
| --- | --- | --- |
| 1 | `TerminalRosterRow.tsx` | `muted` process tone keeps `text-faint` behind a documented adapter; `StatusDotTone` `faint` step recorded as a promotion candidate |
| 2 | `WorkflowDefinitionList.tsx` | description clamped in `RosterRow`'s `secondary` slot |
| 3 | `GoalBar.tsx` | original `color-mix` bracket restored, token gap recorded with the real 2% / 4.9% numbers per spec §2.6 |
| 4 | `PublishDialog.tsx` + `.test.tsx` | `role="group"` makes the label real; three assertions re-scoped through it with a pinned row count |
| 5 | `GitReviewFileSectionShell.tsx` | `data-app-action-review-file-toggle` moved onto the surviving trigger |
| 6 | `PublishDialog.tsx` | leading icon slot regained `text-muted-foreground` |

Findings 7–11 were reported only (no code change). FOUNDER-A…E remain open for you — FOUNDER-A in particular is now parked at the pre-PR pixels, so taking the token shift is a one-line flip whenever you want it.

**Verification at `9b099bf24`:**
- Gates: `check_appearance_scaling.py` ✅ · `check_frontend_boundaries.py` ✅ · `report_frontend_structure.py --strict` ✅ · `check_max_lines.py` ✅ · `check_docs.py` ✅
- Tests: **98/98** — `PublishDialog` 7/7, `WorkflowRuns` 4/4, `WorkflowDefinitionEditor` 4/4, `WorkflowRunsSurface` 14/14, `WorkflowDefinitionsSurface` 9/9, `ActivityChips` 4/4, `GoalBar` 11/11, `GoalBarObjectiveEditor` 9/9, `GoalBarResultPopover` 6/6, `LoopsPanel` 7/7, `FileTreeOverlay` 7/7, plus `Card` 4/4, `RosterRow` 7/7, `Disclosure` 5/5. `GitPanel.test.tsx` still hits the known pre-existing `host/*` dist gap — untouched.
- Negative control on the one test change: removing `role="group"` from `PublishDialog.tsx` fails 2 of 7; restored, green again.

PR #1797, opened 2026-08-13
feat(design-system): wave-2 assembly — RosterPanel pattern + Badge size="micro"

pablonyx:
## What this assembles

Wave-2's assembly step: the two shapes that only cross the rule of two once the four slice branches sit side by side. Both were recorded as promotion candidates inside the slices and deliberately left there, because merging across areas is a promotion and no slice owned both areas.

### 1. `RosterPanel` — the activity panel wrapper

The activity chips' click-in panels each hand-rolled the same wrapper character-for-character:

- `AgentsRosterPanel` (native subagents)
- `LiveTerminalsRosterPanel` (terminals, live-wired)
- `TerminalsRosterPanel` (the playground-only clone of the same)
- `LoopsPanel` (the slice-workflows rewrite)

That is **four instances** of one skeleton: root `flex flex-col gap-1.5`, a `px-1 pt-0.5` static label header, a `px-1 pb-1 text-ui-sm text-muted-foreground` empty line, and a `flex flex-col gap-0.5` row list.

`primitives/patterns/RosterPanel.tsx` transcribes that skeleton verbatim (no redesign) and adds the two slots the fourth instance needs — `headerAction` (LoopsPanel's "Arm a new loop" `IconButton`) and `footer` (its `LoopComposer`). `empty` is a slot rather than a flag, so LoopsPanel passes `null` to suppress the empty line while composing instead of growing a second axis.

It sits flat beside `RosterRow`, not under `patterns/panel/` — the panel kit set is closed. It is non-interactive and owns **no** state stack: `RosterRow` remains the sole owner of hover/selection/press.

The header is deliberately a static label, **not** a `PanelHeaderEntry`. That component is a tablist tab entry (`role="tab"`, `aria-selected`, roving `tabIndex`, `aria-controls`) with no static variant, so forcing it onto an always-one, non-interactive section label would announce tab semantics with no tablist owner. LoopsPanel's recorded C1 deviation was correct; its substance now lives in `RosterPanel`'s docstring as the pattern's own contract.

`TerminalsRosterPanel` is deleted and `ActivityFixtures` composes `RosterPanel` + `TerminalRosterRow` directly. Every existing `data-*` panel attribute (`data-agents-roster-panel`, `data-terminals-roster-panel`, `data-loops-panel`) is preserved through the props passthrough.

### 2. `Badge size="micro"` — the square count chip

Four instances again, across two areas:

- `GitReviewBaseSelector` trigger changed-count
- `GitReviewBaseSelector` option-row changed-count
- `GitReviewTargetSelector` "default" branch badge
- `LoopsPanel`'s native/emulated chip

The first three were collapsed last slice into the local `GitReviewCountChip`; the fourth sat in a different area and was recorded as a promotion candidate rather than folded in.

`Badge` gains a `size` axis orthogonal to `tone`: `default` is the bordered pill, `micro` is `rounded-sm border-transparent px-1 py-0.5 leading-none` with a flat `bg-muted` fill for the neutral tone. Non-neutral tones keep their own tint and ink at either size, so the axes stay independent — LoopsPanel's emulated chip is simply `tone="warning" size="micro"`.

**Change 2 was implemented, not deferred.** The fallback (DEFER-TO-G if the size axis forces a border/structure flip that reads as a redesign) did not trigger: `micro` only makes the existing border transparent and retunes radius/padding. The element structure, the tone table and the tone ink are untouched, which is a size extension rather than a redesign.

`GitReviewCountChip` is deleted; `GIT_REVIEW_SELECTOR_TRIGGER_CLASS` stays where it is.

## Registry / docs

- New `RosterPanel` index row in the Patterns table of `specs/DESIGN_SYSTEM.md` (no incubating note — all four consumers land in this PR) plus a `LibraryEntry` in `components/playground/library/entries/roster-panel.tsx`, registered from `patterns.tsx`.
- `Badge`'s existing index row now names the `size` axis, and its library demo shows both sizes. No new Badge row — this is one primitive's axis, not a new component.
- `library-registry.test.ts`'s physical inventory count re-truthed 38 → 39 patterns (86 → 87 total).

## Grep proofs

`TerminalsRosterPanel` — zero references to the deleted component; only `LiveTerminalsRosterPanel` (a different, live-wired component) and one prose mention remain:

```
ActivityFixtures.tsx:64:          // The live wrapper (`LiveTerminalsRosterPanel`) owns feed streaming
LiveTerminalsRosterPanel.tsx:12:export interface LiveTerminalsRosterPanelProps {
LiveTerminalsRosterPanel.tsx:25:export function LiveTerminalsRosterPanel({ processes, nowMs }: ...
SessionActivityBar.tsx:6:import { LiveTerminalsRosterPanel } from ...
SessionActivityBar.tsx:54:  // LiveTerminalsRosterPanel (expand a row to stream its feed).
SessionActivityBar.tsx:89:              <LiveTerminalsRosterPanel
```

`GitReviewCountChip` — zero code references; two prose mentions recording the promotion:

```
specs/DESIGN_SYSTEM.md:831:  (Badge index row, naming what micro was promoted from)
GitReviewSelectorChrome.tsx:19:  (docstring recording that the chip moved to Badge size="micro")
```

## Verification

Five gates, all green on the pushed head:

| Gate | Result |
| --- | --- |
| `check_appearance_scaling.py` | passed |
| `check_frontend_boundaries.py` | passed |
| `report_frontend_structure.py --strict` | passed, TOTAL: 0 violations across 2315 files |
| `check_max_lines.py` | passed |
| `check_docs.py` | passed (208 Markdown files) |

Single-file vitests:

| Test | Result |
| --- | --- |
| `library-registry.test.ts` | 4 passed |
| `LoopsPanel.test.tsx` | 7 passed |
| `ActivityChips.test.tsx` | 4 passed |
| `GitPanel.test.tsx` | **not runnable** — pre-existing env gap, see below |

`GitPanel.test.tsx` fails to load with `Cannot find package '@proliferate/product-client/host/ProductHostProvider'` at its own line 5. That import is untouched by this PR and the failure is a missing `host/*` dist in the local environment, not a finding; it is left alone and no dist was built.

**Negative control** on the one re-truthed assertion: removing `ROSTER_PANEL_ENTRY` from the patterns tier makes `library-registry.test.ts` fail with `physical owners with no library-sheet entry: #product/primitives/patterns/RosterPanel` (2 of 4 tests red). Restored, green again.

## Deviations from the frozen spec

None on substance. Two mechanical notes:

- The spec wrote micro's neutral fill as part of the flat size string (`... bg-muted ...`). Applied literally after the tone classes it would have overwritten `warning`'s `bg-warning-subtle`, contradicting the spec's own "non-neutral tones keep their tint/ink" clause. The fill is therefore applied only for `size="micro" && tone="neutral"`; the geometry half of the string applies to every tone. Same rendered result, spec's stated intent preserved.
- The spec's call-site line for `GitReviewTargetSelector` said "no tabular-nums needed" for the `default` label, which is what landed; the two `GitReviewBaseSelector` count sites keep `className="tabular-nums"`.


pablonyx:
## Review — PR #1797 (wave-2 assembly), opus reviewer

Reviewed the assembly's own delta only. `git diff 199d247ba..717692d76` is exactly the implementer's two commits (15 files); the merge commit `717692d76` carries **no product changes of its own** — its combined diff is empty and its non-assembly files are byte-identical to base `199d247ba` (the slice-settings review fixes). Change set is clean.

### Verdict on change 2 (Badge `size="micro"`): **EXTENSION, not a redesign. Keep it. No DEFER-TO-G.**

Re-judged against the doctrine rather than the implementer's framing:

1. **Variant budget** (DESIGN_SYSTEM.md:792) bites at the *third* orthogonal axis — "a pattern that needs a third orthogonal variant axis is split or redesigned … never extended". `Badge` had exactly one (`tone`); `size` is the second. Well inside budget, and conservative next to `Button` (variant × size × loading × destructive).
2. **The doctrine's own operational shape-identity test**, same line: "same skeleton DOM and same slot contract is the same shape; **differing only in token values is the same shape**; differing in slot structure is a different shape." `micro` keeps the single `<span>`, the children-only slot contract, and the `forwardRef` surface. Every delta — radius, padding, leading, border *color*, neutral fill — is a token value. Same shape by definition.
3. The border is **not removed**, it is neutralized to `transparent`. `border` (width) stays on the base string, so the box model and the DOM are unchanged; `border-transparent` in `sizeClasses.micro` wins the twMerge border-color slot over the tone's. That is the weakest possible reading of "border flip", and it is not structural.

The one genuine wrinkle is the `micro × neutral` fill cross-product (F2 below) — but one cell of a 7×2 grid carrying a different token is *precisely* "differing only in token values". It is a documentation defect, not grounds for a redesign ruling.

### Verified claims

| Claim | Result |
|---|---|
| Verbatim transcription of the triplicate skeleton | **Holds.** Root `flex flex-col gap-1.5`, header `px-1 pt-0.5` + `text-ui font-medium text-foreground` span, empty `px-1 pb-1 text-ui-sm text-muted-foreground` `<p>`, list `<ul className="flex flex-col gap-0.5">` — all match `237e675a2:AgentsRosterPanel.tsx` and both siblings. `<ul>` still omitted entirely when empty, `<p>` still omitted when populated. See F3 for the one header delta. |
| All four `data-*` attributes preserved | **Holds.** `data-agents-roster-panel` (AgentsRosterPanel.tsx:27), `data-terminals-roster-panel` (LiveTerminalsRosterPanel.tsx:37 + ActivityFixtures.tsx:71), `data-loops-panel` (LoopsPanel.tsx:61). All reach the root `<div>` via `...props`. Hyphenated JSX attrs skip excess-property checking, so this typechecks on the component. Row-level attrs untouched (rows were never migrated). |
| LoopsPanel empty/footer wiring reproduces prior behavior | **Holds exactly**, including the odd cell. Old: `<p>` iff `sorted.length === 0 && !composing`. New: `empty={composing ? null : "No loops armed."}` + RosterPanel's `!hasRows && empty != null`. Identical, and the `composing && !capabilities.supported` case (no composer *and* no empty line) is preserved. `LoopsPanel.test.tsx` 7/7 green, including "shows the empty state once the composer is cancelled". |
| Deletion completeness | **Holds.** Zero references to `TerminalsRosterPanel` (excluding `LiveTerminalsRosterPanel`) and zero live references to `GitReviewCountChip` — only two prose mentions recording the promotion. `GIT_REVIEW_SELECTOR_TRIGGER_CLASS` still exported and still used at all 3 call sites (GitReviewBaseSelector.tsx:37, GitReviewTargetSelector.tsx:56,73). |
| Registry parity + inventory re-truth | **Holds.** Patterns row at DESIGN_SYSTEM.md:924 (alphabetical, between `ProductPageShell` and `RosterRow`, no incubating note), `entries/roster-panel.tsx` `LibraryEntry`, Badge row at :831 names the size axis, counts 38→39 / 86→87. `library-registry.test.ts` 4/4 green, and **negative control confirmed**: reverting to `patterns: 38, total: 86` fails the parity assertion. |
| Doctrine placement | **Holds.** `primitives/patterns/RosterPanel.tsx` flat beside `RosterRow`, not under the closed `patterns/panel/` kit. Non-interactive, owns no state stack, paints no hover/selection/press. Docstring carries LoopsPanel's C1 deviation (static label, not a tablist `PanelHeaderEntry`) as the pattern's own contract, and the index row restates it. |
| "Same rendered output" for the spec-contradiction fix | **Holds on the narrow question, and the deviation is the right call.** Bundling `bg-muted` into the geometry string would have clobbered `warning`'s `bg-warning-subtle` under twMerge; the conditional `MICRO_NEUTRAL_FILL` yields `bg-muted` for neutral and `bg-warning-subtle text-warning-foreground` for warning, which is what both prior chips painted. **It does not hold as a general claim about the chips** — see F4/F5. |

---

### Findings

**F1 — LOW (fixing).** `GitReviewSelectorChrome.tsx:18-23`: the promotion note is a `/** … */` JSDoc block at end of file attached to nothing, left behind where `GitReviewCountChip` was deleted. A JSDoc that documents no declaration is a tombstone, not documentation. Folding it into the file-header docstring.

**F2 — MEDIUM (fixing).** Prose contradicts code on axis independence. `Badge.tsx:52` special-cases `size === "micro" && tone === "neutral"` to swap the fill from `bg-surface-control` to `bg-muted` — so the neutral tone's fill *depends on the size*. Yet `Badge.tsx:18` asserts "so tone and size stay independent axes" and DESIGN_SYSTEM.md:831 opens with "two independent axes". This is a doctrine document; a conformance reviewer reading that index row would not expect the cross-product cell, and the next person to touch `Badge` would read the docstring as licence to drop the condition. Re-truthing both to state the exception plainly. (No code change — the conditional is correct; only the claim is wrong.)

**F3 — LOW, informational, no fix.** The unified header is `flex items-center justify-between px-1 pt-0.5`, where `AgentsRosterPanel`/`LiveTerminalsRosterPanel` previously had a plain block `div.px-1.pt-0.5`. Not drift: the spec requires `headerAction` be right-aligned, which needs the flex row, and `LoopsPanel`'s pre-existing header was already character-for-character this string — the triplicate converged onto the fourth instance, which is the correct direction. With a single `text-ui` span the flex line height equals the block line-box height, so rendering is unchanged.

**F4 — FOUNDER (visual).** `LoopsPanel`'s native/emulated chip is genuinely restyled by this migration, and the report's "same rendered output" framing understates it: `text-ui` (13px/18px) → `text-ui-sm` (12px/16px) with `leading-none`, and `rounded` (4px) → `rounded-sm` (2px), plus a 1px transparent border. Smaller, tighter text in the loops roster meta line. This is exactly what the frozen spec's micro class string dictates and it is the price of the two chips converging on one owner, so I am **not** changing it — but it is a real visible delta on a shipped surface and it is your call, not mine.

**F5 — LOW, informational, no fix.** The git count chip grows 2px in each dimension: `Badge`'s base carries `border` (1px width) unconditionally and micro only neutralizes the *colour*, whereas the old `GitReviewCountChip` had no border at all. Everything else is character-identical (`inline-flex items-center rounded-sm bg-muted px-1 py-0.5 text-ui-sm font-medium leading-none text-muted-foreground` + `tabular-nums` re-added at the two count call sites). ~18px inside the `h-6` (24px) trigger — no clipping. Spec-authored (`border-transparent` in the micro string presumes the base border stays). Dropping `tabular-nums` on the `default` branch label is correct; the string is non-numeric.

---

### Gates & tests at `717692d76`

- `check_appearance_scaling.py` ✅ · `check_frontend_boundaries.py` ✅ · `report_frontend_structure.py --strict` ✅ (2315 files, TOTAL 0) · `check_max_lines.py` ✅ · `check_docs.py` ✅ (208 md). Also ran `check_design_attribution.py` ✅.
- `library-registry.test.ts` 4/4 ✅ (+ negative control red as expected) · `LoopsPanel.test.tsx` 7/7 ✅ · `ActivityChips.test.tsx` 4/4 ✅.
- `GitPanel.test.tsx` red on the known pre-existing `@proliferate/product-client/host/*` missing-dist load failure — out of scope, unrelated to this delta.

Fixes for F1 and F2 pushed as a follow-up commit.

pablonyx:
## Holistic cross-PR conformance review (wave 2, union of all 11 PRs)

Read-only sweep of the whole program delta (`9cfa80c73...96e6a0805`, 314 files) against `.claude/skills/ui-conformance-review/SKILL.md` checks 1–8. Every PR in this wave was reviewed in isolation; this is the first pass over the **union**, so it only reports defects that are invisible PR-by-PR. Doc/comment fixes landed as `4dd5c1deb`. **Not an approval.**

### Headline: what the wave actually retired

| Charter signature | Before (charter appendix, `cb8edce4d`) | Now (`4dd5c1deb`) |
| --- | --- | --- |
| Check 3 — hand-rolled overlay shells | 7 | **3** |
| Check 2 — paint-fingerprint clusters (`--tree --min-files 2`) | 31 | **19** |
| Check 2 — top cluster `bg-card border border-border rounded-lg` | 9 files | **gone** |
| Check 5 — `lucide-react` importers outside `primitives/icons/` | 49 | **0** (`lucide --base 9cfa80c73 --head HEAD` → 0 new identifiers) |
| Check 4 — ungated bracket sites under `components/` | 173 | **170** |
| Check 8 — settings pane-root rhythm drift | 3 panes at `space-y-5`, 1 at `space-y-1.5` | **0** — `SettingsPageBody` owns it; the two residual `space-y-*` roots are `AccountSettingsPane` (at the `space-y-6` mode) and a nested helper in `RepoActionsPane` |

Overlay-shell detail: `PaneSideOverlay` deleted, `SelectedResponseActionMenu` → `DropdownMenu`, `PublishDialog`'s invalid listbox → `aria-current`, `OpenTargetMenu`'s click-only `role="menu"`/`role="menuitem"` simply removed (both present at `9cfa80c73`, both gone). Remaining three: `ComposerInlineMenu` (`role="listbox"`, deliberate), `DelegatedAgentHoverCard` (`role="tooltip"` — `Tooltip.tsx:19` types `content: string`, so the conversion is genuinely blocked), `FileTreeOverlay` (`role="dialog"`).

---

### Findings

#### H1 — `ProductNotice` was slated for deletion in this wave and was missed · REPORTED-ONLY
`apps/packages/product-client/src/components/app/ProductNotice.tsx` — zero importers repo-wide (`grep -rn 'ProductNotice' apps` returns only the file itself). `git log 9cfa80c73...HEAD -- <path>` is **empty**: no PR in the wave touched it. Meanwhile `NoticeBanner` — promoted specifically to replace this shape — landed with 12 product consumers. This is the exact failure the at-least-one-call-site rule names, and it got worse rather than better: the component now sits outside every tier with zero call sites *beside* the sanctioned owner of its shape. It is a `git rm`. Not fixed here (file deletion is not a comment).

#### H2 — rule-of-two asymmetry created by the wave: `SecretEditorDialog` inline error · REPORTED-ONLY
`components/patterns/secrets/SecretEditorDialog.tsx:377` still paints
`rounded-md border border-destructive/25 bg-destructive-subtle px-3 py-2 text-ui text-destructive`.
Its **byte-identical twin** — named as such in the doctrine's own known-duplicates row — is `ApiKeyCreatorModal.tsx`, which migrated in this wave to `<NoticeBanner tone="destructive">` (line 205). The wave touched `SecretEditorDialog` twice (`git log 9cfa80c73..HEAD` → 2 commits) and left the hand-roll. One of a named byte-identical pair moving while the sibling stays is strictly worse than neither moving: the duplicate is now invisible to a fingerprint census that only reports 2+ files.

#### H3 — rule-of-two asymmetry created by the wave: `HarnessAuthSection` icon tile · REPORTED-ONLY
`settings/panes/agents/harness/HarnessAuthSection.tsx:342` paints
`size-8 … rounded-md bg-surface-control text-muted-foreground`.
`IconTile` renders `size-8 rounded-md` for `size="md"` and `bg-surface-control text-muted-foreground` for `tone="control"` — i.e. this is character-equivalent to a bare `<IconTile>`, the default. The file was touched by the wave, and four sibling files in the same harness area (`HarnessPane`, `ProviderRow`, `HarnessConfigIssueBanner`, `HarnessProvidersSection`) all adopted `IconTile`. No recorded cause at the site (the nearby comment is the unrelated `MethodCard`→`RadioCardGroup` deferral). Separately `HarnessPane.tsx:246` hand-rolls a `size-7` (28px) tile **in a file that imports `IconTile` at line 7 and uses it at line 88** — that one is the known 28px off-step question, same class as the `PlanHandoffDialog` founder item.

#### M1 — four stale `incubating` notes survived the union · FIXED (`4dd5c1deb`)
`grep -rni incubating` found five source notes. Four named slices that have since landed:

| File | Note | Actual consumers |
| --- | --- | --- |
| `primitives/StatusDot.tsx:24` | "shell + workflows slices, wave 2" | 6 (`SettingsSidebar`, `ChatTabsMenu`, `tab-rendering`, `WorkflowRunList`, `WorkflowRunDetail`, `PrStatusBadge`) |
| `primitives/patterns/tabs/TabGroupPill.tsx:10` | "shell slice adopts" | `TabGroupPillWithMenu` |
| `primitives/patterns/tabs/ChromeTab.tsx:43` | "shell slice adopts" | `ChatTabWithMenu`, `ChatPlaygroundPage` |
| `primitives/patterns/panel/PanelHeaderEntry.tsx:36` | "shell slice" | `RightPanelHeaderEntryList`, `TerminalHeaderIcon` |

The chat and workflows slices struck their own incubating notes; nobody struck the shell/tabs/panel ones, because no single slice owned them at review time. The DESIGN_SYSTEM index rows were already clean — this was a source/index divergence.

#### M2 — the `Disclosure` comment recorded one of four blockers · FIXED (`4dd5c1deb`)
The comment landed on the chat slice and named only the chat blocker. Three more were recorded elsewhere and never made it back to the file a future implementer will read. All four verified:
1. **Row paint is closed** — `className` merges into the *outer* wrapper (`Disclosure.tsx:84`); the header row's `rounded-lg px-2 hover:bg-hover active:bg-active` sits on an inner div with no prop reaching it. Blocks the chat transcript (reinstates PRO-120) *and* `GitReviewFileSectionShell`'s sticky `color-mix` header.
2. **Title type fixed** at `text-heading` = 17px (`tokens.ts:1372`) against the transcript's 14px rows.
3. **No always-visible body slot** — every child goes inside `AnimatedCollapsibleContent`; `HarnessAllModelsSection` keeps a model-count line visible while the list collapses.
4. **Collapsed children stay mounted** — `AnimatedCollapsibleContent` uses `aria-hidden` + `inert` and never unmounts; `ProviderRow`'s tests assert true unmount.

Rewritten as the union, and the index row aligned. Also added the missing recorded-cause comment at `HarnessAllModelsSection` — it was the only one of the four with no note in the tree.

#### M3 — `GitReviewTargetSelector` redraws `PickerPopoverContent`, in a file the wave touched · REPORTED-ONLY
`workspace/git/GitReviewTargetSelector.tsx:84-118` composes `PopoverSearchField` + a raw `max-h-64 overflow-y-auto` div + a raw `<p className="px-2 py-2 …">No branches</p>`. That is exactly `PickerPopoverContent`'s skeleton (search field → `min-h-0 overflow-y-auto` body → `PickerEmptyRow`), which has 5 existing consumers. The only real differences are `max-h-64` vs `max-h-80` (a `className` override the pattern already accepts) and the empty-row padding. The workflows slice's WP-4 adopted `PopoverSearchField`/`PopoverMenuItem` here but stopped short of the container pattern — check 1 (redraw) created inside a conformance PR. Flagged by the assembly analysis as an adjacent finding; confirming it as real.

#### M4 — registry-row prose drifted from the code as landed · FIXED (`4dd5c1deb`)
- `IconTile` row claimed "Two of the transcript tiles have migrated" — 8 files consume it. Rewritten, and it now distinguishes the off-step/off-tone remainder from H3's exact instance.
- `StatusDot` row claimed it was "the union of two hand-rolled dots (`PrStatusDot`, `RecentWorkStatusDot`) and a dozen inline spans" — `RecentWorkStatusDot` never migrated (it needs the pulsing `live` state), and 13 inline `icon-status rounded-full bg-*` spans remain. Row now says which and why.
- `Card` row omitted the settings adoption (`AppearancePane`, `OrganizationBudgetsPane`, `HarnessAuthApiKeyDetails`), which was not part of the promotion evidence.
- `PopoverSearchField.tsx` docstring claimed "a hairline divider below" — it draws none. `FileTreeOverlay.tsx:187` hand-rolls `<div className="h-px … bg-border" />` precisely because the primitive does not. Comment corrected.

#### L1 — `ListRow` accumulated refusals, not adoptions · REPORTED-ONLY (recorded in Current Gaps)
The naive dead-vocabulary scan now reports `ListRow` as alive; it is not. Its only non-playground mentions are two *comments* (`PlanHandoffDialog.tsx:102`, `PromptRecoveryPanel.tsx:38-44`) explaining why it does not fit — it is a `<button>`, so it cannot host a row with its own controls. Two independent slices independently declined to adopt it. That is retire-or-respecify evidence, and it is worth Pablo's ruling rather than another wave of "migrate onto ListRow" specs.

#### L2 — `RosterRow` density split is coherent but undocumented · REPORTED-ONLY
Settings (6 sites) and workflows (4 sites) pass `density="comfortable"`; activity, git and chat take the `compact` default. That is a defensible role split, but the index row says nothing about it, and the vocab spec's "default compact" note left the flip question open. Worth one sentence in the row before a future slice picks the other value by coin flip.

#### L3 — `GitReviewSelectorChrome.tsx:22` carries `hover:bg-surface-elevated-secondary` · INFORMATIONAL
Check-7 HARD hit. Pre-existing paint hoisted into a documented local helper by the workflows slice — ratchet-neutral, not charged. The other three HARD hits in the union diff (`WorktreeStorageSection` ×2, `SettingsScreen:210`) are comment lines describing sanctions, and `GitReviewFileSectionShell:88` is `hover:bg-[var(--token)]`, which the charter exempts as token consumption.

---

### Current Gaps rewrite

The aggregate paragraph at `DESIGN_SYSTEM.md:1154` was deliberately deferred to this pass so four slices would not conflict on it. Every clause was re-verified; the numbers above are the evidence. Notably: `RangeSlider` joins the dead list (playground-only consumers), `ListRow` stays on it for the reason in L1, and `ProductNotice` stays on it with the deletion miss called out explicitly.

### Founder-decision list

All ~20 items from the ledger re-checked against the code; **none was silently fixed by a later PR**, and none needed a correction to its description. Spot-verified: `NoticeBanner` still has no way to suppress the live region (`role={role ?? style.role}`, `NoticeBanner.tsx:84` — the prop chooses `status` vs `alert`, never *none*); `ClaimBanner` renders two `NoticeBanner`s with the `Hand` glyph in the leading gutter; `PlanHandoffDialog:115` uses a bare `<IconTile>` (default `md` = 32px, was 28); `FileTreeOverlay:173` records the lost inline clear button; `LoopsPanel:157` is on `Badge size="micro"`; and `text-body-emphasis` has zero hits under `settings/panes/organization/`, confirming the org identity row's lost type step. H3 above is a *new* founder-adjacent item in the same class as the `PlanHandoffDialog` sizing ruling.

### Gates

`check_appearance_scaling.py`, `check_frontend_boundaries.py`, `report_frontend_structure.py --strict`, `check_max_lines.py`, `check_docs.py`, `check_design_attribution.py` — all green. `library-registry.test.ts` 4/4.

### Residual risk

Checks 1, 2, 6 and 8 are judgment calls where a signature only nominates candidates, and none of the eight is CI-enforced (PR G is the answer to that). The three REPORTED-ONLY behavioral findings above (H1, H2, H3) and M3 are small, disjoint edits — but they are conformance debt this wave created or narrowly missed, so leaving them unlisted would let the program close claiming a cleaner tree than it has.

PR #1802, opened 2026-08-13
refactor(product): close the three rule-of-two asymmetries wave-2 created

pablonyx:
Wave-2 migrated one half of three duplicate pairs and left the twin hand-rolled. A half-migrated pair is worse than an unmigrated one: the survivors no longer share a paint fingerprint, so a duplicate census stops seeing them. This converges all three.

Base is `gates-pr-g` (`f2150a356`). All six conformance gates green; no baseline needed a decrement (none of the touched sites held a census slot except two `foreground-alpha-foundation` entries in the harness files, which are untouched `border-foreground/20` hits and still exactly met — the census stays tight).

## 1. `SecretEditorDialog` inline error → `NoticeBanner tone="destructive"`

Rule-of-two evidence: the byte-identical twin at `settings/panes/agent-auth/ApiKeyCreatorModal.tsx:205` already renders `{error ? <NoticeBanner tone="destructive">{error}</NoticeBanner> : null}`. This site is now the same expression, character for character.

```
- <div className="rounded-md border border-destructive/25 bg-destructive-subtle px-3 py-2 text-ui text-destructive">{error}</div>
+ <NoticeBanner tone="destructive">{error}</NoticeBanner>
```

Paint deltas absorbed by the pattern: `rounded-md` → `rounded-lg`, `border-destructive/25` → `border-destructive/30`, `px-3 py-2` → `p-3`. Fill and ink are unchanged.

**Accessibility improvement, flagged as noted:** the hand-rolled `<div>` had no role and announced nothing. `NoticeBanner` assigns `role="alert"` for the destructive tone, so the save error is now announced. This is the tone-implied live region already on the founder list; calling it out explicitly because it is a behavior change, not only a paint one. There were no aria or data attributes on the old div to preserve.

## 2. Harness icon tiles → `IconTile`

`HarnessAuthSection.tsx` (MethodCard glyph plate) was character-equivalent to a bare `IconTile`:

```
- <span aria-hidden className="mb-auto flex size-8 items-center justify-center rounded-md bg-surface-control text-muted-foreground">
+ <IconTile aria-hidden className="mb-auto">
```

`md` = `size-8 rounded-md`, `control` = `bg-surface-control text-muted-foreground`. `flex` → `inline-flex` is inert: the span is a flex item of the card's `flex-col`, so it is blockified either way. `mb-auto` is kept at the call site; `shrink-0` and `data-icon-tile` are additive. Four sibling files in this area adopted `IconTile` in the wave; this one was missed with no recorded cause.

`HarnessPane.tsx` (runtime-row tile) — **migrated, with a founder flag.** Same shape one step small:

```
- <span className="inline-flex size-7 shrink-0 items-center justify-center rounded-md bg-surface-control text-muted-foreground">
+ <IconTile>
```

I took the 28→32px growth rather than leaving it, for three reasons: `IconTile`'s own source comment states that "the stray 28px and 36px one-offs round to the nearest step when their call sites adopt this tile"; `PlanHandoffDialog` already made the identical 28→32 rounding **in this wave and it is merged in the tree** (`grid size-7 … place-items-center` → `<IconTile>`); and leaving this one recreates exactly the half-migrated pair this PR exists to close. See FOUNDER item below — the two should be ruled once, together, and reverted together if the ruling goes the other way. A source comment at the call site records that.

## 3. `GitReviewTargetSelector` → `PickerPopoverContent` + `PickerEmptyRow`

The hand-roll wired `PopoverSearchField` + `max-h-64 overflow-y-auto` + a raw `<p className="px-2 py-2 text-ui text-muted-foreground">` — the exact skeleton the pattern owns for its 5 consumers.

Deltas handled:
- **`max-h-64` override.** Passed as `className`, the pattern's `max-h-80` default untouched. The pattern concatenated `className` onto its own string, which leaves both `max-h-80` and `max-h-64` standing so the stylesheet, not the caller, picks the winner — the override silently lost. Switched the pattern to `twMerge` (as `NoticeBanner`/`IconTile` already do) so overrides resolve last-wins. This also makes `HomeTargetPicker`'s existing `max-h-[min(20rem,…)]` override deterministic. Non-conflicting classes merge identically, so no consumer's appearance changes otherwise. A test asserts the merged frame carries `max-h-64` and not `max-h-80`, negative-controlled against the old concatenation.
- **Cap scope.** `max-h-64` now bounds the whole picker rather than the list alone, so the popover is shorter by the search row's height. Recorded in a source comment.
- **Empty-row padding.** `px-2 py-2` → the pattern's `px-2.5 py-[5px]`, accepted per brief.
- **`ariaLabel` / `autoFocus` had no passthrough.** Rather than drop the accessible name and the open-focus, the pattern gained two optional props, `searchAriaLabel` and `searchAutoFocus`, forwarded to `PopoverSearchField`. Both default to undefined, so no existing consumer changes.

Preserved: the 40-item cap, the prepend of an unmatched active ref, `GitBranchIcon` per row, the `default` badge, the check glyph, `labelClassName`, and every `onSelect`/`close` path. No data attributes existed on the replaced markup.

**`PopoverSearchField` docstring — no change needed.** The claim that it "draws a hairline divider below" is already gone on `gates-pr-g`; the docstring now correctly says it draws none and names `FileTreeOverlay`'s `h-px` hand-roll as the consequence. I did not add the divider: it would change all 5+ consumers' appearance in ways I cannot verify from tests, and the brief prefers the comment fix. `FileTreeOverlay:187` keeps its hand-rolled rule.

## Doctrine

`specs/DESIGN_SYSTEM.md` § Current Gaps named `SecretEditorDialog`'s error and `HarnessAuthSection`'s tile as "the clearest rule-of-two debt". Re-truthed to record both as closed, plus the two adjacent sites.

## Gates

| gate | result |
| --- | --- |
| `report_frontend_structure.py --strict` | pass, TOTAL 0 |
| `check_max_lines.py` | pass |
| `check_docs.py` | pass (208 files) |
| `check_component_library.py` | pass |

No baseline decrement was required.

## Tests

Colocated single-file runs, all green:

| suite | result |
| --- | --- |
| `SecretEditorDialog.test.tsx` | 4/4 (1 new) |
| `ApiKeyCreatorModal.test.tsx` | 4/4 |
| `HarnessPane.test.tsx` | 34/34 |
| `HarnessPane.hierarchy.test.tsx` | 6/6 |
| `GitPanel.test.tsx` | 12/12 (runnable again on this branch's test-lane fix) |
| `GitReviewTargetSelector.test.tsx` | 2/2 (new file) |
| `HomeTargetPicker.test.tsx` | 11/11 (other `PickerPopoverContent` consumer) |
| `PopoverButton.test.tsx` | 1/1 |

Negative controls run and restored:
- SecretEditorDialog: reverting to the hand-rolled `<div>` fails the new `role="alert"` assertion.
- PickerPopoverContent: reverting `twMerge` to string concatenation fails the `max-h-64` override assertion (`expected 'flex max-h-80 … max-h-64' not to contain 'max-h-80'`).

Note recorded in the new test file: `PickerEmptyRow` is unreachable in this selector, because an active ref dropped by the filter is prepended so the list is never empty. That was equally true of the `<p>No branches</p>` it replaces — behavior preserved, not introduced.

## FOUNDER items

1. **28→32px tile rounding, now at two sites.** `PlanHandoffDialog` (already merged in wave-2) and `HarnessPane`'s runtime row (this PR) both grew a 28px tile to `IconTile`'s 32px `md` step. `IconTile`'s source sanctions the rounding; the pending founder question should be ruled once and applied to both, and reverted at both if it goes the other way.
2. **`NoticeBanner` destructive tone adds `role="alert"` where the hand-roll was silent.** Already on the founder list as an accepted improvement; restated here because `SecretEditorDialog`'s save error now announces.
3. **`PopoverSearchField` hairline divider.** Left undrawn; `FileTreeOverlay` keeps its `h-px` hand-roll. Adding the divider to the component would touch all 5+ picker consumers and wants a visual ruling.


pablonyx:
## Review — PR #1802 (`8027155dd`, base `gates-pr-g` @ `f2150a356`)

Verdict: the three judgment calls hold. One real finding, fixed in `588878fb6`. No new founder items.

---

### 1. The `twMerge` switch in `PickerPopoverContent` — bug claim REAL, and I found the one consumer whose paint it does change

**(a) The bug is real, and I confirmed it against the emitted stylesheet rather than by reasoning.** The old code was `` `flex max-h-80 min-h-0 flex-col ${className}` `` (`PickerPopoverContent.tsx:25` pre-diff), so both utilities survived. In the built CSS (`apps/desktop/dist/assets/index-D7COynXk.css`), `.max-h-64` is emitted at byte 28283 and `.max-h-80` at 28377 — **`max-h-80` is later, so it won.** `GitReviewTargetSelector`'s intended 16rem cap would have been silently discarded by the pattern's own default. The fix is correct and the new test's negative control reproduces exactly that (`expected 'flex max-h-80 … max-h-64' not to contain 'max-h-80'`); I re-ran it.

**(b) Consumer-by-consumer enumeration.** Five call sites take `PickerPopoverContent` (`PickerEmptyRow`-only importers are unaffected). I ran each className pair through the repo's own `twMerge`:

| consumer | `className` | `bodyClassName` | merged result | delta |
| `EnvironmentSearchSelect.tsx:81` | — | `py-1` | `min-h-0 overflow-y-auto py-1` | none (dedup) |
| `HomeTargetPicker.tsx:146` | `max-h-[min(20rem,calc(100vh-1rem))]` | `py-0` | frame unchanged; body `… py-0` | **body padding 4px → 0** |
| `HomeTargetPicker.tsx:228` | — | — | unchanged | none |
| `playground/library/patterns.tsx:205` | — | — | unchanged | none |
| `GitReviewTargetSelector.tsx:89` | `max-h-64` | — | `flex min-h-0 flex-col max-h-64` | the intended fix |

**FINDING (fixed).** `HomeTargetPicker.tsx:148` passes `bodyClassName="py-0"`. `.py-0` is emitted at byte 72767 and `.py-1` at 72838 — **`py-1` was later, so the pattern's default silently beat the caller's override.** With `twMerge` the override now takes effect and the runtime picker's body loses 4px block padding top and bottom. So the PR body's "no consumer's appearance changes otherwise" is not accurate — there is exactly one, and it is a real 8px height change on a shipped surface.

I kept `twMerge` (reverting would restore the bug, and `py-0` is plainly what that call site asked for — the runtime/SSH divider is meant to run flush). Fixed by recording it at the call site and pinning it: `588878fb6` adds a comment at `HomeTargetPicker.tsx:145` and a test, negative-controlled against the old concatenation (`expected 'min-h-0 overflow-y-auto py-1 py-0' not to contain 'py-1'`).

**(c) `HomeTargetPicker`'s arbitrary override still resolves as its author intended.** `max-h-[min(20rem,calc(100vh-1rem))]` is emitted at byte 29086, i.e. **after** `.max-h-80` at 28377 — it already won under concatenation, and `twMerge` keeps it winning. No change at that site's frame. Verified the repo `twMerge` classifies the arbitrary value into the `max-h` group (the `extend` in `utils/tw-merge.ts` appends `max-h-control`, it does not replace the group).

### 2. 28→32px at `HarnessPane` — both cited pieces of evidence are true as stated

- `IconTile.tsx:12-14` says verbatim: *"The stray 28px and 36px one-offs round to the nearest step when their call sites adopt this tile — a fourth step would make the axis a scale rather than a variant set."* ✅
- `PlanHandoffDialog` took the identical rounding and it **is** merged on this base: commit `1eeae76fe` turned `<span className="grid size-7 shrink-0 place-items-center rounded-md bg-surface-control text-muted-foreground">` into `<IconTile>` (`PlanHandoffDialog.tsx:115`), and `git merge-base --is-ancestor 1eeae76fe f2150a356` confirms it precedes this PR's base. ✅

Both claims hold, so the call stands and stays a single combined founder ruling. Nothing to revert.

Also confirmed `HarnessPane.tsx` already imported `IconTile` at line 7 for a pre-existing tile at line 88 — the runtime row was genuinely the leftover half of an in-file pair, which strengthens the case.

**`flex` → `inline-flex` at `HarnessAuthSection` is inert as claimed.** The tile's parent is the MethodCard `Button` with `relative flex min-h-28 w-full flex-col …` (`HarnessAuthSection.tsx:331`), so the tile is a flex item and `inline-flex` is blockified to `flex` by CSS. `mb-auto` is preserved at the call site; `shrink-0` and `data-icon-tile` are additive (`shrink-0` is protective in a column, never expansive). ✅

### 3. `searchAriaLabel` / `searchAutoFocus` — optional, inert, and defensible

- Both default to `undefined` and are forwarded straight to `PopoverSearchField`, whose own `ariaLabel`/`autoFocus` already default to `undefined`. Passing `undefined` is identical to not passing at all, so no existing consumer changes. ✅
- Behavior preserved at the call site: pre-migration `PopoverSearchField` got `ariaLabel="Search branches"` and bare `autoFocus`; post-migration it gets `searchAriaLabel="Search branches"` and bare `searchAutoFocus`. The new test asserts both (`getByLabelText("Search branches")`, `document.activeElement === search`). ✅
- **Not a variant-budget problem.** These are passthroughs to props the wrapped component already exposes — a slot the pattern forgot to open, not a new axis. The alternative (dropping the accessible name so a migration could land) would have been the worse outcome. Adding them from a call-site PR is the right call.

### Other checks

- **Fix 1 is genuinely convergent, not merely both-migrated.** The twin's pre-migration markup at `9b757a230` was `<div className="rounded-md border border-destructive/25 bg-destructive-subtle px-3 py-2 text-ui text-destructive">` — byte-identical to the div removed here — and both sites now render the identical expression (`ApiKeyCreatorModal.tsx:205`, `SecretEditorDialog.tsx:377`). Same absorption (`rounded-md`→`rounded-lg`, `/25`→`/30`, `px-3 py-2`→`p-3`), same result. ✅
- **`PickerEmptyRow` unreachability was NOT introduced by the migration.** `branchOptions` (`GitReviewTargetSelector.tsx:31-47`) is untouched by this diff; `activeRef = baseRef ?? "origin/main"` is always truthy, so the prepend guarantees length ≥ 1. The `<p>No branches</p>` it replaced sat behind the identical `branchOptions.length === 0` guard and was equally dead. Recording rather than deleting is right. ✅
- **Fix 3's preserved behavior all verified in source and covered by the new test**: 40-item cap, unmatched-active-ref prepend, `GitBranchIcon` per row, `default` badge, check glyph, `labelClassName`, `onSelect`/`close`. ✅
- **Uncalled paint delta, minor, no action needed**: the outer wrapper lost `gap-1` and the list gained the pattern's `py-1`, so the 4px between search row and first row moved rather than disappeared. Worth a line in the body alongside the empty-row padding note, but it is the pattern's own rhythm and within the absorption the brief sanctions.
- **`PopoverSearchField` docstring claim checked**: `PopoverSearchField.tsx:22-25` already says it draws no divider and names `FileTreeOverlay`'s hand-roll. No change needed, as stated. ✅

### Gates and tests re-run at `588878fb6`

All six gates green: `check_appearance_scaling.py`, `check_frontend_boundaries.py`, `report_frontend_structure.py --strict` (TOTAL 0), `check_max_lines.py`, `check_docs.py` (208 files), `check_component_library.py`.

Tests (colocated, single-file, staggered): `SecretEditorDialog` 4/4 · `ApiKeyCreatorModal` 4/4 · `HarnessPane` 34/34 · `HarnessPane.hierarchy` 6/6 · `GitPanel` 12/12 · `GitReviewTargetSelector` 2/2 · `HomeTargetPicker` **12/12** (1 added) · `PopoverButton` 1/1.

Negative controls, all reproduced and restored:
- revert `NoticeBanner` → hand-rolled div: `Unable to find an accessible element with the role "alert"`
- revert frame `twMerge` → join: `expected 'flex max-h-80 … max-h-64' not to contain 'max-h-80'`
- revert body `twMerge` → join (my new test): `expected 'min-h-0 overflow-y-auto py-1 py-0' not to contain 'py-1'`

No new FOUNDER items. The three already recorded stand unchanged.

PR #1804, opened 2026-08-13
fix(product): remove logo from sidebar brand row (PRO-139)

tairabun:
## Summary
Removes the Proliferate logo mark to the left of the "Proliferate" wordmark in the main sidebar brand row, per PRO-139.

`ProductSidebarBrandRow`'s `icon` prop is already optional (defaults to `null`), so the change only drops the icon at the call site and removes the now-unused `ProliferateIcon` import.

Linear: https://linear.app/proliferate-team/issue/PRO-139/sidebar-should-not-have-logo


tairabun:
@greptile review

PR #1805, opened 2026-08-13
fix(product): show the first prompt as the session title until the generated summary lands (PRO-83)

tairabun:
## Summary

- Fixes the session-title lifecycle for a new session (PRO-83). Previously the tab showed the agent/model name until the AI-magic summary landed, and the harness's turn-end ACP `session_info_update` then overwrote the persisted title with the literal first prompt.
- New order — prompt first, summary wins:
  - **At send (same frame):** the client seeds the optimistic session record / directory entry with the prompt text (`promptFallbackTitle`: whitespace-collapsed, 160-char cap), so the agent name never shows once the user sends.
  - **At acceptance (durable):** the runtime's prompt endpoint assigns the same fallback via `update_title_if_absent` when the session has no title, and the prompt response carries it — covers every client.
  - **Summary replaces it:** the existing AI-magic flow PATCHes the generated title (explicit update always wins) and persists.
  - **No more turn-end revert:** harness-provided titles are fallback-only (`SessionStateDurable::update_title_if_absent`); they never overwrite a user rename, the prompt fallback, or the generated summary. The emitted `session_info_update` payload mirrors what was actually persisted.
  - `applySessionSummary` no longer clears a shown title when the server summary has `title: null` (null means "not titled yet", never "clear").

## Testing

- Unit (runtime): `cargo test -p anyharness-lib --lib sessions::store::tests sessions_prompt live::sessions` — new `update_title_if_absent_never_replaces_an_assigned_title` and `prompt_fallback_title_collapses_whitespace_and_caps_length`, plus existing suites (150+ tests) pass.
- Unit (client): `pnpm vitest run src/hooks/sessions src/hooks/chat src/lib/domain/sessions` — 467 tests pass; `pnpm typecheck` clean.
- Manual (local profile stack, driven via CDP browser against the dev renderer):
  1. New session + first message → tab title shows the prompt text immediately (reveal animation), agent name never appears.
  2. Runtime persisted the prompt as title at acceptance (verified via `GET /v1/sessions/{id}`).
  3. Applied a title update through the same PATCH the AI-magic flow issues → turn finished → title kept the updated value (no revert), and survived a full page reload. (Local profiles have no signed-in control-plane user, so the real `/v1/ai_magic` call 401s locally; the persisted-write path it uses is identical.)

## Observability

- None; the existing `session.title.update` measurement operation covers the summary write, and prompt-title assignment rides the existing prompt latency spans.

## Readiness

  in this PR.

## Verification

- `cargo test -p anyharness-lib --lib sessions::store::tests sessions_prompt live::sessions`
- `pnpm --filter @proliferate/product-client typecheck && pnpm --filter @proliferate/product-client vitest run src/hooks/sessions src/hooks/chat src/lib/domain/sessions`


tairabun:
@greptile review

PR #1806, opened 2026-08-13
refactor(design-system): delete RangeSlider, AuthProviderButton, ListRow

pablonyx:
## Founder ruling

Retire all three dead library components. Delete them outright — no respec, no allowlisting.

- `RangeSlider`
- `AuthProviderButton`
- `ListRow`

All three had zero non-playground call sites (verified twice by independent review passes). `RangeSlider`'s only three consumers were under `/playground/`. `ListRow`'s only two non-playground mentions were prose comments explaining why it did *not* fit — refusals, not adoptions.

Base is `conformance-debt` (tip of the 12-PR stack carrying the new conformance gates).

### RangeSlider
- `apps/packages/product-client/src/primitives/RangeSlider.tsx` deleted.
- Registry row removed from `specs/DESIGN_SYSTEM.md`.
- Playground library demo (`RangeSliderDemo` + entry) removed from `playground/library/primitives.tsx`.
- Its two `component_library_allowlist.json` entries (FOUNDER dead-vocabulary allowance + missing-JSDoc allowance) removed.
- Its two real playground-fixture consumers repointed onto the `Input` primitive (`type="range"`, `variant="unstyled"`, identical styling) instead of a raw `<input>`, which the `report_frontend_structure` `RAW_DOM_CONTROL` check would otherwise flag:
  - `apps/packages/product-client/src/components/playground/UpdateUiPlaygroundControls.tsx`
  - `apps/packages/product-client/src/components/playground/loading/PlaygroundThinkingTimingControls.tsx`

### AuthProviderButton
- `apps/packages/product-client/src/primitives/patterns/AuthProviderButton.tsx` deleted.
- Registry row removed from `specs/DESIGN_SYSTEM.md`.
- Playground library demo/entry removed from `playground/library/patterns.tsx`.
- Its two `component_library_allowlist.json` entries removed.

### ListRow
- `apps/packages/product-client/src/primitives/patterns/ListRow.tsx` deleted.
- Registry row removed from `specs/DESIGN_SYSTEM.md`.
- Playground library demo/entry removed from `playground/library/patterns.tsx`.
- Its two `component_library_allowlist.json` entries removed.
- `PlaygroundIndex.tsx` (the dev-only playground front door, a real consumer, not a demo) repointed from `ListRow` onto `RosterRow` — the closest live analog, with the identical `leading`/`title`/`trailing` slot shape and its own state stack.
- `library-registry.test.ts`'s physical-inventory counts decremented to match the tree across all three deletions: `{ primitives: 38→37, patterns: 39→37, total: 87→84 }`.

## Grep proofs (zero remaining code references)

```
$ grep -rn "RangeSlider" apps/ scripts/
$ grep -rn "AuthProviderButton" apps/ scripts/
$ grep -rn "\bListRow\b" apps/ scripts/ | grep -v "MobileListRow\|MemberListRow\|member-list-rows"
```
All three return empty. (`MobileListRow` and `MemberListRow`/`member-list-rows` are unrelated pre-existing names — a mobile-only row primitive and a settings domain type — not the retired component.) Prose in `specs/` that records the retirement is expected and present.

## The two ListRow-refusal comment rewrites

Both comments recorded *why* their call site hand-rolled instead of adopting `ListRow`. The reasoning is preserved; only the reference to the now-deleted component name is removed.

**`PlanHandoffDialog.tsx`** (reasoning: title type-scale mismatch)
```diff
- (`bg-surface-elevated-secondary` / `bg-card`), and `ListRow`
- titles at `text-heading` (17px) against this row's 13px, which
+ (`bg-surface-elevated-secondary` / `bg-card`), and a slotted
+ list-row shape fixes its title at `text-heading` (17px) against
+ this row's 13px, which
```

**`PromptRecoveryPanel.tsx`** (reasoning: button-in-button nesting)
```diff
- // check 7): `ListRow` is the shape this maps to, but it renders a
- // `<button>` whose slots are inside it, and this row hosts its own
+ // check 7): the slotted list-row shape this maps to renders a
+ // `<button>` with its slots inside it, and this row hosts its own
  // Retry button — adopting it would nest interactive elements. The
  ...
- // non-interactive spelling of `ListRow`, which is a review ruling.
+ // non-interactive spelling of that shape, which is a review ruling.
```

## Doctrine re-truth

`specs/DESIGN_SYSTEM.md`'s "Current Gaps" passage named all three (plus the already-deleted `ProductNotice`) as dead vocabulary *awaiting* a founder ruling. Re-worded to record them as retired, past tense, same voice/structure. Four illustrative doctrine examples that used `ListRow` as the canonical leading/title/trailing slotted-pattern example (slots, state-stack ownership, naming, hand-assembled-state-stacks) are swapped to `RosterRow`, which has the identical shape and is still live. `specs/PRODUCT_SENSE.md`'s naming example swaps the same way. `check_component_library.py`'s masking-example docstring, which cited a commented-out `ListRow` import, is genericized to a placeholder name.

## Gate / test results

All six required gates green on the final commit:

| Gate | Result |
| --- | --- |
| `report_frontend_structure.py --strict` | pass (0 violations) |
| `check_max_lines.py` | pass |
| `check_docs.py` | pass |
| `check_component_library.py` | pass |

`library-registry.test.ts` — 4/4 passing with decremented counts. Negative-controlled: temporarily re-added a deleted registry row (`ListRow`, no file) and a matching library-sheet entry, confirmed both `check_component_library.py` and this test go red with the expected violation messages, then reverted.

`PromptRecoveryPanel.test.tsx` (colocated test for an edited file) — 1/1 passing. `PlanHandoffDialog.tsx` has no colocated test file.

## Commits

One per component (`8023133`, `cdc7c54`, `b7b26bd`), each deleting the source file, registry row, playground demo, and allowlist entries; the `ListRow` commit also carries the doctrine re-truth and the two comment rewrites since they're inherently about the full three-component retirement.


pablonyx:
## Review — PR #1806 (retire `RangeSlider`, `AuthProviderButton`, `ListRow`)

Reviewed `186d8263e...b7b26bd9d`. The deletions themselves are authorized and clean; I reviewed what they touched. Four findings, all fixed in `dcf7d00c0`. No founder items.

### Repoint verdicts

**`RangeSlider` → `Input type="range" variant="unstyled"` — behavior and appearance preserved exactly.**

`variant="unstyled"` resolves to `""`, so `Input` renders `<input className={twMerge("", className)} {...props} />`. At both sites the rendered DOM is byte-identical to what `RangeSlider` produced:

- `UpdateUiPlaygroundControls.tsx:322-347` — old `twMerge(base, "h-1 w-40")` collapsed to the same class set now written out literally. `min`/`max`/`step`/`value`/`disabled`/`aria-label`/`onChange` all intact.
- `PlaygroundThinkingTimingControls.tsx:174-185` — old call passed no `className`, so it got the base string verbatim; the new literal is that base string. `id`/`min`/`max`/`step`/`value`/`disabled`/`onChange` intact, `htmlFor={id}` on the `Label` still resolves.

Neither site used the forwarded ref. Non-finding: the sliders keep `appearance-none` with no `::-webkit-slider-thumb` rule, same as before — pre-existing, unchanged by this PR.

**`ListRow` → `RosterRow` in `PlaygroundIndex.tsx` — behavior preserved, appearance degraded-acceptable.**

Behavior: `onClick` → `onSelect` is correct, and the row gains keyboard activation it already had as a `<button>` (`RosterRow` handles Enter/Space on its `role="button"` div). The `aria-hidden` leading slot is unused here. `divide-y divide-border-light` on the container correctly replaces `ListRow`'s own `border-b … last:border-b-0`.

Degradations, all playground-only and acceptable: the title drops from `text-heading font-medium` (17px, medium) to `text-body` (14px, regular), and the description line loses `truncate` so a long path can now wrap. Flagging, not blocking.

### Doctrine-example swaps

Four of the five swaps hold up on the merits:

- `DESIGN_SYSTEM.md:709` (slots) — `RosterRow` genuinely exposes `leading`/`title`/`trailing`. ✅
- `DESIGN_SYSTEM.md:711` (state-stack ownership) — `RosterRow`'s docstring says it owns the whole stack so no roster call site hand-assembles one. Strictly a better example than `ListRow` was. ✅
- `DESIGN_SYSTEM.md:733` (named for the job) — ✅
- `DESIGN_SYSTEM.md:815` (check 7) — ✅

**Finding 1 — `specs/PRODUCT_SENSE.md:35-36`: the naming example stopped demonstrating its own rule.** The passage is a contrast pair: good name vs the same noun prefixed with a feature. Swapping only the good half left `` `RosterRow`, never `WorkspaceListRow` `` — `WorkspaceListRow` is not `RosterRow` named for the feature, so the pair no longer illustrates anything. Fixed to `` never `WorkspaceRosterRow` ``.

### Rewritten refusal comments — both now state false reasons

This is the substantive finding. Both comments were rewritten to say "a slotted list-row shape" rather than name the deleted component. But with `ListRow` gone, the only slotted list-row shape in the tree is `RosterRow` — the same component this PR promotes into the doctrine as the canonical example — and neither comment is true of it.

**Finding 2 — `PlanHandoffDialog.tsx:100-106`.** Claims the shape "fixes its title at `text-heading` (17px) against this row's 13px, which would resize a dialog row by a third." `RosterRow` titles at `text-ui` (13px) compact / `text-body` (14px) comfortable — never 17px, and compact is *exactly* this row's 13px. The stated blocker is not merely unverifiable, it is inverted: read against the live library, the comment argues for a fit it means to refuse. The real blocker is different — `RosterRow` paints no resting surface (hover/selected/active only), so this row's `border-border/70` + `bg-foreground/5` frame stays hand-written either way. Rewritten to say that, naming `RosterRow`, keeping the "needs a review ruling" conclusion.

**Finding 3 — `PromptRecoveryPanel.tsx:37-44`.** Claims the shape "renders a `<button>` with its slots inside it, and this row hosts its own Retry button — adopting it would nest interactive elements", and that landing needs "a non-interactive spelling of that shape, which is a review ruling." Both blockers are already resolved by `RosterRow`, whose docstring says verbatim: *"Interactive rows render a div with button semantics rather than a real `<button>`: a roster row's whole point is that it carries its own trailing action buttons, and a button inside a button is invalid DOM."* And it is already non-interactive when `onSelect` is omitted. So the comment declares a review ruling that has in fact already been made. The real blockers: `RosterRow` gates its hover wash on `onSelect` (a row given none paints no states at all) while this row wants the wash without being pressable, and its secondary line is fixed at `text-muted-foreground` with no tone axis while this row's second line *is* the error (`text-destructive/80`). Rewritten to those.

Also dropped from that comment: the claim that `hover:bg-hover` is "driving the `group/recovery` reveal". `group/recovery` has no `group-hover/recovery:` consumer anywhere in the tree — the class is vestigial and the claim was already false before this PR. The `group/recovery` class itself is left in place as out-of-scope.

**Finding 4 — `DESIGN_SYSTEM.md:1166` Current Gaps.** The re-truthed passage attributes one shared reason to both refusals ("the row hosted its own controls, so it could not be a single `<button>`"). That was never true of `PlanHandoffDialog`, whose row is a single `Button` hosting no controls and whose refusal was about type scale. Pre-existing conflation carried forward into the rewrite; since the passage is being re-truthed anyway, fixed to record the two distinct historical reasons plus what the comments say against `RosterRow` today.

### Docstring genericization — accept

`check_component_library.py:305-311`: `ListRow` → `Widget` as the masked-comment illustration. The surrounding sentence ("this tree records refusals to adopt a component in prose right where the adoption would have gone") still carries the *why*, so the non-obvious behavior is still explained without leaning on a specific name. Fine as-is.

### Deletion completeness and gate honesty — clean

- Zero code references remain: `grep -rn` over `apps/ scripts/` for all three names returns only `MobileListRow` / `MemberListRow` / `member-list-rows`, which are unrelated pre-existing names, plus the intended `specs/` retirement prose. Confirmed independently.
- Allowlist: all six entries (three `deadLibraryComponents`, three `missingJsdoc`) are removed, not zeroed. `deadLibraryComponents` remains as an empty `{}` category key, which is the correct zero-ratchet shape — the stale-allowance rule (`check_component_library.py:668`) has nothing to observe.
- `check_component_library.py` diff is **exactly one line** (the docstring). No rule logic weakened; the gate is green because the components are gone.
- Counts verified against the tree, not taken on trust: primitives `ls primitives/*.tsx` minus tests = 37 ✅; patterns `find patterns -name '*.tsx'` = 39, minus the two glob-excluded toast members (`ToastBody`, `ToastExpansion`) = 37 ✅; icons = 10 ✅; total 84 ✅.

### Gates and tests

All six gates PASS at `dcf7d00c0` (re-run after the fixes): `check_appearance_scaling.py`, `check_frontend_boundaries.py`, `report_frontend_structure.py --strict`, `check_max_lines.py`, `check_docs.py`, `check_component_library.py`.

`library-registry.test.ts` 4/4 ✅, `PromptRecoveryPanel.test.tsx` 1/1 ✅. Independently negative-controlled the count assertion: flipping `primitives: 37 → 38` / `total: 84 → 85` fails with `expected { primitives: 37, … } to deeply equal { primitives: 38, … }`, confirming the numbers are derived from the filesystem rather than asserted at each other. No colocated tests exist for the repointed playground files or `PlanHandoffDialog.tsx`.

PR #1786, opened 2026-08-12
chore(design-sync): repoint pin + bake the working agreement into the payload

pablonyx:
Two small follow-ups to #1782:

1. **Repoint the committed project pin** at the canonical new-account "Proliferate Library" (`26506959-df72-4ac3-8cb6-24508dec6b42`). The old-account copy is dead.
2. **Bake the Claude Design working agreement into the payload.** `make-meta.mjs` now emits `guidelines/working-agreement.md` (from `templates/guidelines.md.tmpl`): closed vocabulary, mandatory `NEW COMPONENTS` / `MODIFIED` deviation declarations, tokens-only styling, and the handoff-manifest format engineering agents implement from. The payload README points at it. The `auxSha` recipe is untouched (`keyRecipe` stays 8); the template is covered by `scriptsSha`.

Already re-synced to the live project (delta: README, guidelines file, sync anchor — bundle/CSS byte-identical). The authoring twin of the agreement lives in `~/agent-engineering/nodes/implement-design-handoff/`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1810, opened 2026-08-13
feat(tooling): design-sync evaluates the registry — wave-2 restructure support

pablonyx:
Supersedes #1786 (carries its two commits: the project-pin repoint to the new-account Proliferate Library, and the baked-in `guidelines/working-agreement.md`).

## Why

The wave-2 component-hierarchy program restructured the playground library registry: entries are now composed across tier files, kit demo files (`tabs.tsx`, `panel.tsx`), and an `entries/` directory, with kit components living at `primitives/patterns/<kit>/` paths. The design-sync builder statically re-derived the registry with two independent regex parsers hardcoded to the old four-file shape (count-gated at 36/24/10/11) — both structurally wrong now, exactly the drift its tripwire was built to catch.

## What

- **`dump-registry.mjs` (new)**: Vite-compiles and executes the real `LIBRARY_TIERS`, writing `.out/.ds-registry-manifest.json`. The registry is now *evaluated, not parsed* — future reorganizations can't silently drift from the build. Renders never execute during the dump.
- **`registry-manifest.mjs`** becomes a loader over that dump. New gates: 4 nonzero tiers, unique names, every subpath resolves on disk, `EXPECTED_TOTAL = 87` as the deliberate-bump tripwire.
- **`groupFor` consolidated to one definition** (was triplicated — flagged in #1782's review) and gains kit-from-subpath: `primitives/patterns/<kit>/` → that kit (closed set incl. new `tabs`/`panel`), else the existing name rules, else tier id.
- **`make-entry.mjs`** drops its regex + cross-check; the generated bundle entry gets an injected precomputed name→group map instead of its own `groupFor` copy.
- **`emit-cards.mjs`**: purpose scrape adapted to the rewritten spec's sanctioned index, missing purpose is now a hard error; prunes stale `components/`/`_preview/` output first (retired entries would otherwise ship as orphan cards).
- CONTRACT.md registry facts rewritten.

## Registry delta this build ships

87 entries (was 81): +11 new (IconTile, StatusDot, Card, Disclosure, NoticeBanner, RosterPanel, RosterRow, ChromeTab, TabGroupPill, PanelHeaderEntry, SettingsPageBody), −5 retired (AuthProviderButton, ListRow, ModelTable, RangeSlider, SettingsPageHeader), 2 group moves (PaneOptionsMenuItem → panel, ProductPageShell → patterns).

## Verification

- `node scripts/design-sync/build.mjs` green end-to-end; render-check **87/87 clean**, zero console/page errors; contact sheets + all 11 new cards eyeballed.
- `@ds-bundle` manifest: 87 components, no doubled path segments (incl. `secrets/SecretManagementPanel`).
- All 87 purposes resolved from the sanctioned index, zero fallbacks; 87/87 `.d.ts` emitted; collision pins unchanged (guarded now against stale subpaths).

## Light-mode cards (added after review)

Every entry now emits two cards instead of one: `<Name>.html` (dark, unchanged) and `<Name>.light.html`, which differs only by `data-mode="light"` on `<html>` and a ` (light)` suffix on its `@dsCard group` so the picker keeps the modes as sibling sections. Both mount the same demo from the same preview module, so there is no second bundle and no second render path.

**87 components, 174 cards. Render-check 174/174 clean** (zero console/page errors, no blank or empty roots, no thin PNGs), and all 87 light screenshots hash-differ from their dark twin, which is the check that `data-mode` is actually applying rather than the card silently rendering dark twice.

## The working agreement is now the design project's instructions

`templates/guidelines.md.tmpl` generated the old four-rule contract, which predated the repo attachment, the `.dc.html` artifact model, the export mechanics, the light cards and the scope-agreement step. The project therefore carried one statement of the contract in its instructions field and a thinner, different one in its payload.

The template now generates the same text that is pasted into the project, so the two cannot disagree. It adds what the old version could not know: the repo is attached and which parts of it to read, screens are `.dc.html` artifacts at the project root, `components/` is regenerated wholesale on every sync while artifacts and notes survive, what an export actually carries (and that the per-component cards do **not**, which is why every manifest row needs its own repo import), scope agreement before any screen is drawn, prototype glue as a fourth ledger status, both modes holding, and a handoff written as an implementation spec with an explicit not-being-built list.

Shipped to the live project as a metadata-only delta (guidelines, README, build meta, sync manifest), verified by read-back. No rebuild and no card re-upload: `.ds-build-meta.json` still reports 87 components and the picker's 174 cards are untouched.

The paste-ready copy of the same text lives in the `implement-design-handoff` node, alongside the skill that consumes what it produces ([agent-engineering#2](https://github.com/proliferate-ai/agent-engineering/pull/2)).

## Review follow-up

Greptile's P1 on stale registry dumps is fixed in `f7b8faf58`. `readRegistryDump()` refreshed the dump only when it was absent, so a standalone stage rerun after a registry edit reused the previous dump and would emit a payload describing a registry that no longer exists. It now also re-dumps when the dump is older than the newest file under the library directory. Negative control both ways: silent on a fresh dump, re-evaluates after `touch primitives.tsx`.


pablonyx:
## Review: PR #1810 — design-sync evaluates the registry (wave-2 restructure support)

**Verdict: clean.** No blocking defects. Two non-blocking findings below. I verified this by actually running the pipeline (dump-registry → make-entry → build-bundle → build-css → emit-cards → make-meta) against the real wave-2 registry and by loading the built `_ds_bundle.js` in a sandboxed Node context to check the live-evaluated output, not just reading the diff.

### Findings

**Nit — `scripts/design-sync/templates/guidelines.md.tmpl:45`**
The handoff-manifest example row uses `#product/patterns/SettingsRow`, but the real subpath emitted by the pipeline post-kit-move is `#product/primitives/patterns/settings/SettingsRow` (confirmed via `emit-cards.mjs`'s `buildPromptMd()` and a live build — every generated `.prompt.md` "Import" line uses the full subpath). Purely a doc example, but if a design agent pattern-matches it into a real handoff manifest instead of reading each card's own `.prompt.md`, the import line will be wrong.

**Should-fix (low severity) — `scripts/design-sync/emit-cards.mjs`**
`components/` and `_preview/` are pruned (`rm -rf`) at the top of `main()`, before the per-entry loop runs, but the missing-purpose hard-fail (`missingPurposeNames.length > 0 → throw`, near the end of `main()`) only fires *after* the loop finishes. A run that hits a missing-purpose entry therefore replaces the last-known-good `.out/components`/`.out/_preview` tree with a partial one (complete except the offending entries' `.prompt.md`) before erroring out. This is safe through the documented `build.mjs` pipeline — `execFileSync` throws synchronously and halts before `make-meta.mjs` runs — but `build.mjs`'s own header comment says every stage is "also runnable standalone," and `make-meta.mjs`'s `warnMissing`-and-continue pattern would silently emit a payload with fewer `renderHashes`/`sourceHashes` than entries if someone reruns `make-meta.mjs` standalone after a failed `emit-cards.mjs` without checking the exit code. Consider moving the purpose-completeness check before the `rm()`, or writing to a temp dir and renaming on success.

### What I verified locally (read-only against gitignored `.out/`, no tracked files touched — `git status` clean throughout)

- `dump-registry.mjs` / `registry-manifest.mjs`: run cleanly, 87 entries (37 primitives / 37 patterns / 10 icons / 3 product-patterns), matches `EXPECTED_TOTAL` and the commit message. Scratch dir (`product-client/node_modules/.ds-design-sync`) is created and removed via `finally` — confirmed empty after runs.
- `make-entry.mjs` + `build-bundle.mjs`: run cleanly — 87 components in the `@ds-bundle` manifest comment, 0 ambiguous-export warnings, `PIN_EXPORTS` validation (new: throws if a pin subpath left the registry) passes for `Checkbox`/`checkbox-primitive` and `Tooltip`/`tooltip-primitive`.
- **Divergence check (concern #1):** loaded the actual built `_ds_bundle.js` in a Node sandbox (stub `window`/`document`, real React/ReactDOM) and inspected the live `window.ProliferateUI.__demos`/`__tiers` — exactly 87 demo keys, `__tiers` sums to 87 across the same 37/37/10/3 split, and **zero** entries fell back to `__groups[name] ?? tier.id`'s default (i.e. every name from the live re-evaluated `LIBRARY_TIERS` — inside the actual shipped IIFE, via a *second*, differently-aliased Vite build — matched a name in the statically-dumped manifest). No divergence between `dump-registry.mjs`'s SSR evaluation and `build-bundle.mjs`'s lib-mode evaluation of the same module graph, for the current registry.
- `groupFor`: single definition (`registry-manifest.mjs:109`), single call site (`registry-manifest.mjs:187`, passed `(displayName, tierId, subpath)`); `make-entry.mjs` only consumes the precomputed `{name: group}` map, no local reimplementation. Live bundle's group membership matches CONTRACT.md exactly, including `secrets/SecretManagementPanel` correctly staying in `product-patterns` (its subpath is `#product/components/patterns/secrets/...`, so it doesn't match the `primitives/patterns/<kit>/` rule).
- `emit-cards.mjs`: 87/87 purposes resolved from the sanctioned index, 0 JSDoc fallbacks, 87/87 got a `.d.ts`. Verified the new pruning actually works: planted a fake stale `FakeRetiredThing` component dir + preview file, reran `emit-cards.mjs`, both were deleted before re-emit.
- `make-meta.mjs`: `_ds_sync.json` still has the unchanged 9 keys with `keyRecipe: 8`; 87/87 `renderHashes` and `sourceKeys`; `guidelines/working-agreement.md` is written with `__DS_GIT_SHA__` correctly substituted (`library-registry@c68a61cf0`).
- `resolveSrcFile` (`registry-manifest.mjs:93`) now only tries `${rel}.tsx` (dropped the old `.ts`/`index.tsx`/`index.ts` fallback candidates from the pre-PR resolver) — confirmed all 87 subpaths, including the new `#product/components/...` and `#product/primitives/patterns/<kit>/...` shapes, resolve directly to a same-named `.tsx` file. Safe for the current registry; not a live bug, just less permissive than before (worth knowing if a future entry ever needs an `.ts` or `index.tsx` target).
- `.design-sync/config.json` projectId repoint and `CONTRACT.md`/`README.md.tmpl` prose updates are consistent with the actual code behavior verified above.

Did not relitigate: evaluated-dump-over-regex, kit-from-subpath group derivation, the auxSha recipe excluding guidelines, or the renderHashes residue for demo-only edits in `entries/*.tsx` — all confirmed as-implemented, not re-argued.

pablonyx:
Both review findings fixed in 0ccc9e672:

- **should-fix**: purposes are now resolved in a pre-pass that throws *before* the `components/`/`_preview/` prune, so a purpose-less entry can no longer leave a partial payload where the last good one was (the standalone-`make-meta.mjs` hazard).
- **nit**: the working agreement's manifest example now uses the real post-kit-move subpath (`#product/primitives/patterns/settings/SettingsRow`).

Rebuilt end-to-end after the fixes: 87/87 render-clean, `_ds_bundle.js` byte-identical, 87 purposes from the sanctioned index with zero fallbacks.

The 87-component payload is live in the Proliferate Library design project — full upload verified against `list_files`, with the five retired components and the two stale group-move copies deleted.

pablonyx:
Valid, fixed in f7b8faf58.

`readRegistryDump()` refreshed the dump only when it was absent, so a standalone rerun of `make-entry` / `build-bundle` / `emit-cards` / `make-meta` after a registry edit reused the previous dump and would emit a payload describing a registry that no longer exists: retired entries still carded, new ones missing, group moves stale. `build.mjs` itself was never exposed, since it runs `dump-registry` as its first stage, but its own header advertises every stage as runnable standalone, so the hazard is real for the documented usage.

The dump is now also re-run when it is older than the newest file under `apps/packages/product-client/src/components/playground/library/`, which is the registry's entire source of truth given a dump records only names, subpaths and tiers. A full build stays a single evaluation: the dump stage writes the file before any consumer reads it, so the check is a no-op there.

Negative control: with a fresh dump, `loadRegistryManifest()` is silent; after `touch primitives.tsx` the same call logs `library changed since the last dump, re-evaluating` and re-dumps.

PR #1815, opened 2026-08-13
fix(product): remove the duplicate Support row from the main sidebar primary nav (PRO-152)

pablonyx:
## Summary

Fixes [PRO-152 "Support shown twice"](https://linear.app/proliferate-team/issue/PRO-152/support-shown-twice).

The main sidebar presented the same support action twice:

- a **Support** row in the primary navigation (`SidebarPrimaryNavigation`), and
- the footer **"?" help popover**'s "Send feedback" entry (`SidebarHelpSection`),

both displaying the same `SHORTCUTS.openSupport` shortcut label.

## Root cause

Navigation composition, not duplicated source data: the primary nav hardcoded a Support action row (`supportActive={false}` — it was never a real navigation destination) on top of the help footer that already owns support. The nav row was also spec-non-conforming: it called `useOpenSupportReportWindow().openBug` directly, bypassing `deriveSupportMenuAction`, so on an operator/self-managed server it would have opened the vendor feedback modal instead of the operator's configured destination — a direct violation of `specs/codebase/systems/product/support/README.md` ("Product UI must never route a self-managed user to vendor support").

## Fix

Remove the Support row and its plumbing (`supportActive`, `onOpenSupport`, `shortcutLabels.support`, the `LifeBuoy` icon import, and the `useOpenSupportReportWindow` call in `MainSidebar`). The conforming, capability-routed entry points remain untouched:

- sidebar footer help popover (`SidebarHelpFooter`/`SidebarHelpSection`)
- command palette "Open Support" entry
- global `openSupport` keyboard shortcut (bound in `use-app-shortcuts.ts` via `useAppNavigationCommandActions`, which routes vendor/operator/none correctly)
- settings sidebar footer Support action (designed per the settings IA doc; out of scope here)

## Testing

- `pnpm vitest run src/components/workspace/shell/sidebar` — 9 files, 72 tests pass, including a new regression test asserting the primary nav renders no Support row while the help footer entry point remains.
- Negative control: with the source fix reverted (test kept), the suite fails.
- `pnpm typecheck` (product-client) and `python3 scripts/check_frontend_boundaries.py` pass.
- No visual/browser validation: the shared bugathon browser lock was held by another worker, so no dev server was started per the validation rules.

### DIFF
--- a/apps/packages/product-client/src/components/playground/WorkspaceStatusPlayground.tsx
+++ b/apps/packages/product-client/src/components/playground/WorkspaceStatusPlayground.tsx
@@ -316,13 +316,11 @@ function FullSidebarPane() {
           homeActive
           workspacesActive={false}
           workflowsActive={false}
-          supportActive={false}
           onGoHome={() => {}}
           onGoWorkspaces={() => {}}
           onGoWorkflows={() => {}}
-          onOpenSupport={() => {}}
           shortcutRevealVisible={shortcutReveal}
-          shortcutLabels={{ newChat: "⌘N", support: "⌘?" }}
+          shortcutLabels={{ newChat: "⌘N" }}
         />
         <div className="flex min-h-0 flex-col px-2">
           <SidebarRepositoriesHeader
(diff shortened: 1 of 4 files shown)


pablonyx:
Closing per Pablo's ruling: PRO-152 (duplicate Support row) is still real on main, but the sidebar was since split (SidebarScrollingNavigationSection.tsx now wires the row), so this diff no longer applies; the fix will be redone as a fresh PR against the current component structure. Rebase was attempted and aborted without push — see review notes.

PR #1818, opened 2026-08-13
feat(desktop): add consented support snapshots

pablonyx:
## Summary

- add the consented support snapshot: a user-consented, bounded diagnostics artifact prepared on the desktop and attached to a support report
- add the desktop support-snapshot bridge contract with Tauri wrappers, a native availability gate, and archive paths kept behind the native boundary
- add a durable support report queue engine with a single in-process writer, artifact reconciliation on hydrate, and bounded failure/retry accounting
- add the support export scrubber and a bounded finite-evidence reader over the fallback and legacy log families
- add the bounded support-window endpoints for events, raw notifications and sessions, with descriptor-only decoding of SDK values that may be hostile proxies

This is PR 6 in the observability stack. It is stacked on draft PR #1774 / `codex/rust-observability-pr5-bundled-adapters` and is a draft.

Base: `7b975caf1` (`codex/rust-observability-pr5-bundled-adapters`)

Head: `144c05c5f`

259 files changed, +49,356 / −3,371.

## Testing done on this head

Rust:

- `cargo check --workspace --all-targets` → exit 0, zero errors
- `cargo test -p proliferate --lib every_known_provider_prefix_is_removed_without_copying_a_fragment` → 1 passed. Negative control: deleting the `xox[baprs]-` alternative from `patterns.rs` makes it fail at `text.rs:181`, restoring it makes it pass again

Frontend:

- `pnpm run typecheck` in `product-client` → exit 0. Negative control: reverting the two narrowing fixes reproduces exactly the three original errors (`TS2322` at `support-session-window.ts:113`, `TS2345` at `:136` and `:154`)
- `vitest run` over the support-owned suites plus every suite importing the split modules → **41 files, 327 tests, all passing**

Repository shape gates, both of which run in the same CI job:

- `scripts/check_max_lines.py` → exit 0
- `scripts/report_frontend_structure.py --strict` → `TOTAL: 0`, exit 0, with no new allowlist entries

## What this head fixes relative to the authored slice

The slice as authored did not compile and could not be pushed. Four defects, each verified independently:

1. **12 Rust compile errors** (`E0364`, `E0432`, `E0603`, `E0433`, `E0425`, `E0034`, `E0618`, `E0277`, `E0382`). Mostly visibility and import-path errors around the `#[path]`-included evidence modules, where a submodule's `super::` is its immediate parent rather than the module that appears to own the file. Negative control: reverting all nine files reproduces the identical 12-error histogram.
2. **Stale SDK codegen.** The bounded support-window endpoints landed without `anyharness/sdk/generated/openapi.json` or `src/generated/openapi.ts` being regenerated, so `product-client` could not build. Regenerated with `make sdk-generate`; the delta is exactly this slice's 7 schemas and 3 paths, with nothing else added, removed or changed in place.
3. **Four files over the frontend structure gate.** Split along the seams already present, behind unchanged entry points, so no caller import changed.
4. **A Slack-shaped literal in a scrubber fixture** tripped GitHub push protection and blocked the branch outright. The fixture now assembles the token with `concat!`, so the source carries no literal while the test still exercises the same rule — proven by the negative control above. This required rewriting the 54 commits from the scrubber commit onward; nothing else in the tree changed (`git diff` against the pre-rewrite head is that one file, +3/−1).

Also fixed three assertions in `use-support-report-upload-queue.test.ts` that expected a one-argument toast call against a two-argument implementation. Two failed outright; the two paired `not.toHaveBeenCalledWith` forms were silently vacuous, since a one-argument expectation can never match a two-argument call.

## Not verified locally

The full Rust test suites for this slice were not run on this machine. `cargo check --all-targets` proves every test target compiles, but the desktop and collector process suites spawn real child processes and the machine has a 24 GB ceiling with a one-build-at-a-time rule. Those runs are left to CI.


pablonyx:
### `Cargo check & test` is green now, and the reason it was red is worth reading

The 14 failures this check reported were all in test code that **had never once executed**. Every Rust verification on this train was `cargo check` / `cargo check --all-targets`, which compiles test code without evaluating a single assertion in it. This slice's own compile-fix commit said the paths below its tip never compiled — which also means they never ran.

Fixing the 14 exposed a 15th they were masking, and widening from the module filter to the whole desktop lib suite exposed six more plus two flaky ones.

**Most of it was fixture error, not production bugs:** session fixtures moving only one of two representations of a single fact (ingest rejects inputs where they disagree, so those were impossible inputs, not harder ones); a stale `limit_uncertain` claim; credential padding glued onto the credential so `\b`-anchored patterns could never match and the redaction assertions were vacuous; two canaries under the 48-byte run-length gate; `add_truncation` accumulating fixture totals; and goldens sorted alphabetically where `BTreeMap<RejectionReasonV1, _>` sorts by declaration order.

**Three were real defects:**

1. `SupportChildCollectorStateV1::{Unavailable, Cooldown}` were unit variants, and serde does not apply `deny_unknown_fields` to a unit variant of an internally-tagged enum — so `{"kind":"unavailable","extra":0}` deserialized cleanly and the hostile-input rejection this validation layer depends on was vacuous for this type. Empty struct variants fix it with a byte-identical wire form. **This is the only production-code change in the commit.**
2. Six macOS-gated support-evidence tests rooted their fixtures at `std::env::temp_dir()` = `/var/folders/...`, and `/var` is a symlink. The hardened anchor walk opens each component `O_DIRECTORY | O_NOFOLLOW`, failed on the first one with ENOTDIR, and every source read back as `UnsafeMetadata` — those tests were asserting against a rejection, not against the reads they describe. The sibling suite for the same walk already canonicalizes for this exact reason.
3. `assert_descriptor_closed` probed only whether a descriptor *number* was invalid. The kernel hands out the lowest free number, so another thread could reopen it between the drop and the probe — two launch tests failed intermittently with "fd 16 remained open" (2 of 5 full-suite runs, 0 of 8 isolated). Closure is now proven by `(st_dev, st_ino)` identity.

**Verified:** `cargo test -p proliferate --lib` → 527 passed, 0 failed, 4 ignored, stable across three consecutive runs. Each fix negative-controlled; reverting `health.rs` alone fails `every_support_tagged_union_denies_unknown_fields` at `unions.rs:18` while its exact-shape sibling still passes, and leaking instead of dropping the prepared launch still fails the descriptor assertion by identity.

### One thing to know that this PR does not fix

**CI runs 426 of this crate's 531 tests.** The missing 105 are gated to macOS, and they are exactly the platform-specific hardening this train adds: the descriptor-passing launch path, the `O_NOFOLLOW` anchor walk, the finite-evidence reader, the fd remap. None of it has ever been exercised by CI on any slice — six were failing and two flaky tonight, and this page could not have told you. Deciding whether to add a macOS job is above my pay grade at 4am, so it is written up rather than acted on.

Also note `assembly/tests/bounds.rs` now sits at exactly 600/600 lines against the max-lines cap. It was at 597 before this commit, so it was already three lines from the wall. It wants a split rather than the next person shaving comments to fit.

pablonyx:
## Dedicated review — `pr6-consented-support` at `096aad35e`

Scope: 59 commits / 263 files vs `7b975caf1`, reviewed in three parts (desktop Rust, product-client, anyharness + specs). Read-only review, no builds run.

### HIGH | CONFIRMED | `providers/ProductLifecycleRoot.tsx:236` — the login-budget auth gate turns "deferred reconciliation" into "never" for an abandoned account

`SupportReportQueueRoot` mounts only when `authStatus === "authenticated"`, so while unauthenticated there is no queue initialisation, no event listener, no drain, and no retry timer (`use-support-report-upload-queue.ts:156`).

**Failure scenario.** A user enqueues a report, then signs out and does not come back. The persisted job (`proliferate.supportReportJobs.v1`) and its staged attachment bytes stay on disk with nothing that will ever reconcile, retry or delete them. Attachments can be megabytes.

The journal already discloses startup reconciliation of orphaned staged artifacts as deferred to sign-in, and the reasoning for the gate is sound (both ends of the queue require a live Cloud session, and it bought 6 KB of login-bundle headroom where main has under 4 KB). The part worth an explicit decision is that "deferred until sign-in" is unbounded — it wants either a small unauthenticated cleanup pass or a documented retention bound.

### MEDIUM | CONFIRMED | `scrub/patterns.rs:404` `is_high_confidence_opaque` — bare hex secrets are never caught by the catch-all

The gate is `classes >= 3 && distinct >= 10`, where the classes are lowercase, uppercase, digit and symbol. A 64-character lowercase-hex secret — SHA-256-shaped session tokens, HMAC secrets, many self-hosted API keys — has exactly **two** classes, so it fails regardless of length or entropy (`distinct >= 10` passes easily). The 48-byte run-length gate at `:281` is satisfied; it is the class count that lets it through.

Prefixed credentials (`ghp_`, `AKIA`, `xox…`, PEM blocks) are caught by their own patterns, so this is the last-resort layer only, and it applies to Generic/Content text roles. Suggested fix: treat a run of >= 32 pure-hex bytes as its own high-confidence class, or relax to `classes >= 2` when `distinct` is high.

### MEDIUM | PLAUSIBLE | `support-report-upload-persistence.ts:118` — cleanup failures are reported but not recoverable

`cleanupUnreferencedAttachments` surfaces deletion errors via `onCleanupError` but neither retries nor records the orphan, and the controller proceeds as if cleanup succeeded. A transient `EACCES` leaks the file with nothing left that knows to look for it.

### Lower

`LOW | CONFIRMED | scrub/patterns.rs:427 approved_opaque_context` — the label test is an unanchored substring match, so `malicious_path=…` matches `path=`. Over-redaction only, never a leak.

`LOW | note | artifact_store/unix_fs.rs` — roughly 900 lines of `O_NOFOLLOW`, ownership and TOCTOU logic whose tests are platform-gated. Same coverage hole as the rest of the train.

`LOW | CONFIRMED | anyharness-contract/src/v1/support_windows.rs:6` — the three bounded-window enums have no `#[serde(other)]` fallback, so an added variant breaks old readers. Consistent with the versioned-V1 design (new variants imply a V2), so recording it as intent, not a defect.

### Claims rejected on verification

- **Claim:** the legacy-queue migration grants log consent, because it preserves `includeLogs: true` while writing `supportSnapshot: { kind: "none" }`. **Rejected.** `support-report-upload-attempt.ts:60` computes `prepared` only when `job.supportSnapshot.kind === "prepared"`, so a migrated job collects and uploads no snapshot at all. `legacyConsentRequired` (`:282`) then treats `includeLogs === true` plus `supportSnapshot.kind === "none"` plus a server conflict as "ask for consent again". The preserved flag is the re-consent trigger, not a bypass.
- **Claim:** deleting `report-upload-sanitizer.ts` removed redaction. **Rejected.** Redaction moved into the native scrub layer, which applies to every exported field ahead of any server-side effect. The residual gap there is the hex-secret finding above, not the deletion.
- **Spec "Current gaps" item 3** — it states the three SQL-bounded support-window endpoints "do not exist" while this slice ships them (routes, service, store, OpenAPI, SDK client, tests). Left untouched on purpose: that section is main-relative by construction, so the sentence is true against `main` and false as a description of this branch. That ambiguity is journal `## NEEDS PABLO` item 3 and wants a policy decision, not a silent edit from me.

**Verdict.** The consent and redaction architecture holds up: every export entry point checks a timestamped consent proof, data flows through scrubbing before it can leave, byte arithmetic is checked, and file access uses `O_NOFOLLOW`. The two things I would fix before merge are the hex-secret class gap and an answer for the abandoned-queue case.

pablonyx:
## Dedicated review — pr6-consented-support at a1ebd5779

The earlier review on this PR was written at the superseded head `096aad35e`, since rewritten by the restack over the PR 5 Linux-rotation fix. This is the first review at the current head.

**Scope.** 59 commits / 263 files / +49,551 −3,385 against base `2752e3164` (PR #1774). **No builds were run** — no cargo, no `cargo check`/`test`, no pnpm, no vitest, no Docker; every claim below is from reading code. I partitioned the diff into six independently-reviewed slices and verified the load-bearing claims myself afterwards:

| slice | files |
| --- | --- |
| secret scrubber | `diagnostics/support_snapshot/scrub/**` |
| snapshot assembly + schema validation | `support_snapshot/{assembly,schema}/**` |
| artifact store + coordinator | `support_snapshot/{artifact_store,coordinator}/**` |
| collector export + child file evidence | `diagnostics_collector/{support_export,child_bridge,export_admission,producer}/**` |
| product-client / desktop TypeScript | `product-client/src/hooks/support/**`, `providers/**`, `apps/desktop/src/lib/access/**` |
| anyharness server + SDK | `anyharness/crates/**/support_windows*`, `anyharness/sdk/src/client/**` |

Roughly: 154 files in `src-tauri`, 69 in product-client/desktop TS, 39 in anyharness.

---

## Status of the two known prior findings

**1. `ProductLifecycleRoot.tsx:236` login-budget auth gate — STILL PRESENT (confirmed), severity re-rated MEDIUM.**

The gate is unchanged after `a5dd95cb2`:

```tsx
{authStatus === "authenticated" && (
  <Suspense fallback={null}><SupportReportQueueRoot /></Suspense>
)}
```

The new `SupportReportQueueRoot.tsx` documents the rationale (keep the queue/verification/upload modules off the `/login` first-load path), but reconciliation is coupled to the same gate. `PackagedSupportReportQueueController.reconcileOnce()` — the only sweep, the one that calls the native `reconcile_staged_support_snapshots` → `sweep_artifacts`/`sweep_attachments` — is invoked exactly once from `initialize()` (`support-report-queue-controller.ts:75`), which runs only on that mount. `dispose()` is `{ this.disposed = true }` (`:223-225`): no storage write, no deletion. The 8-attempt / 48 h retry-exhaustion backstop only evaluates inside `drainSupportReportQueue`, reachable only from the same mount. I looked specifically for a sign-out hook, an app-startup sweep independent of auth, and a Rust-side TTL reaper; none exists.

What is actually left behind for a user who sends a report, signs out mid-drain, and never returns:
- the V2 queue document at `localStorage["proliferate.supportReportJobs.v2"]` — routed deliberately to raw `window.localStorage` rather than the Tauri store (`apps/desktop/src/lib/access/browser/product-storage.ts:111-113,131-134`), which on a Tauri build is WebView storage persisted to disk. It holds the report message text, credit name, workspace path/branch/id, session id, and attachment metadata.
- staged attachment bytes under `<app_dir>/support-report-attachments/…` (`commands/support.rs:99-100`).
- (latent, see the reachability note below) a staged snapshot would live at `<app_dir>/support-report-snapshots/<artifactId>` (`artifact_store.rs:17`).

I rate this MEDIUM rather than the prior HIGH: nothing leaves the machine, the volume is bounded (10 jobs / 2 MB document; 100 MB attachment ceiling), and it self-heals on any later sign-in. What is unbounded is the *time*, and the user believes the report was sent.

**2. `is_high_confidence_opaque` and bare lowercase-hex secrets — STILL PRESENT (confirmed), and broader than described.** Promoted to the top finding below. No earlier, more specific rule catches such a value unless its JSON key is an exact member of a ~40-entry allowlist.

**Serde hardening claimed in this slice — VERIFIED COMPLETE.** `SupportChildCollectorStateV1::{Unavailable {}, Cooldown {}}` are now empty struct variants with a comment explaining the vacuous-`deny_unknown_fields` hazard (`schema/model/health.rs:99-113`). I scanned every internally-tagged enum in all changed `.rs` files and in `proliferate-diagnostics-protocol`: **zero** remaining unit variants under a `#[serde(tag = ...)]` enum. The two `deny_unknown_fields`-less tagged enums (`diagnostics_collector/supervisor.rs:56`, `supervisor_unsupported.rs:24`) are pre-existing PR 3 types with only struct variants, untouched here.

---

## Reachability of the consented flow (read this before the findings)

The consented-capture pipeline this PR builds **is not reachable from any UI in this slice.** The single site that constructs a job's snapshot field hardcodes it:

```ts
// use-support-modal-state.ts:169-172
// Snapshot preparation/consent is owned by the later modal slice. Until
// that flow supplies an exact prepared artifact, new jobs are explicitly
// no-snapshot; the legacy includeLogs flag never grants consent.
supportSnapshot: { kind: "none" },
```

Nothing in `components/support/**` references `supportSnapshot`, `beginPreparation`, or a consent surface; `beginSupportSnapshotPreparation` and `saveSupportSnapshotArchive` have no non-test callers. All nine Tauri commands are registered in `lib.rs` and callable from the main window, but no product code calls them.

Consequence for the highest-value defect class: **there is no reachable "capture before consent" or "reused consent" path today, because no shipped code can trigger a capture at all.** The native contract that will gate it is structurally sound — `consent` is a required field of `BeginSupportSnapshotInput`, validated as the first two statements of `begin_preparation` (`coordinator/preparation.rs:38-40` → `validate_begin`/`validate_begin_times`) before any I/O, capture, or disk write; every later call (`finish`, `cancel`, `save_archive`, `submit`) re-checks `consent_epoch` against the open preparation; and the legacy migration explicitly refuses to upgrade `includeLogs` into consent (`support-report-upload-persistence.ts:250-266`, `legacyConsentRequired`). But none of that is exercised by a live path in this slice, so it is unproven in situ.

One native-side gap worth noting for whichever slice wires the UI: `validate_begin_times` (`preparation.rs:561-575`) only requires `report_opened_at ≤ granted_at ≤ now`. There is no maximum age on `granted_at`, so nothing at the trust boundary would stop a renderer bug from replaying a months-old grant. Not a defect today (nothing produces a grant), but the freshness bound is the natural place to make "per submission" enforceable natively rather than by renderer convention.

---

## Findings

### HIGH | CONFIRMED | `apps/desktop/src-tauri/src/diagnostics/support_snapshot/scrub/patterns.rs:404-425`, `:281`, `:395`, with `scrub/value.rs:192-234`

The second, purpose-specific secret scrub cannot redact whole classes of real credentials by value shape, and its catch-all is **strictly weaker than both scrubs it is meant to supersede.**

Two independent gates on the opaque catch-all:

```rust
// patterns.rs:419-424
let classes = [lower, upper, digit, symbol].into_iter().filter(|p| *p).count();
let distinct = distinct.into_iter().filter(|p| *p).count();
classes >= 3 && distinct >= 10
```
```rust
// patterns.rs:281 (and identically at :395)
} else if end - start >= 48 && is_high_confidence_opaque(&value[start..end]) {
```

- `classes >= 3` means a **bare lowercase-hex secret has only `lower` + `digit` = 2 classes and is never redacted at any length.** A 64-char lowercase-hex API key passes through verbatim.
- `end - start >= 48` separately exempts every well-shaped secret shorter than 48 bytes — e.g. a 40-char AWS secret access key (`wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY`, four character classes, would otherwise qualify).

The only other line of defence for a structured JSON field is `direct_secret_key_class` (`value.rs:192-234`), an **exact-match allowlist** of ~40 normalized key names. `secret_key_class` (`:166-190`) only strips a fixed *suffix* list (`length/prefix/suffix/sha256/checksum/fingerprint/digest/hash/sha`) — no prefix stripping, no substring containment. So:

Concrete failure, every link read: `{"authToken": "1a79a4d60de6718e8e5b326e338ae533"}` (a Twilio-shaped 32-char auth token).
1. `normalize_key("authToken")` → `"authtoken"` (`value.rs:159-164`).
2. `direct_secret_key_class("authtoken")` → `None`. The list has `accesstoken`, `bearertoken`, `rawtoken`, `sessiontoken`, `securitytoken`, `apikey`, `authorization`, … but **not** `authtoken`.
3. `role_for_key("authtoken")` falls through every category to the `else` at `value.rs:335` → `TextRole::Generic`.
4. `Generic` allows opaque redaction, so `patterns::apply` runs the catch-all.
5. 32 bytes < 48 **and** classes = 2 → not redacted.
6. Output: `"authToken":"1a79a4d60de6718e8e5b326e338ae533"` verbatim.

The same holds in *free text* (a Content-role transcript/tool-output string): `environment_assignment_regex` (`patterns.rs:556`) is `(?:[a-z][a-z0-9]*_)*(?:token|key|secret|password|pass|credential)` — underscore-separated segments only, so camelCase `authToken` never matches its anchored prefix either. Same for `apiToken`, `deviceToken`, `csrfToken`, `webhookSecret`, `userPassword`, `stripeKey`.

**This is a regression, not just a gap.** Both scrubs this pipeline replaces caught 48+ byte opaque runs with no character-class requirement:
- deleted in this PR — `report-upload-sanitizer.ts` `LONG_TOKEN_PATTERN = /(^|[^A-Za-z0-9_\-+/=])([A-Za-z0-9_\-+/=]{48,})(?=…)/g`
- still live for the legacy bundle — `diagnostics/scrub.rs:34` `long_opaque_token_regex = \b[A-Za-z0-9_\-+/=]{48,}\b`

Either of those redacts a 64-char lowercase-hex token; the new purpose-specific scrub does not. The PR's own tests assert the new behaviour: `scrub/tests/positive.rs:35-36` asserts the 64-char and 40-char lowercase-hex values in `fixtures/positive_projected.json` survive verbatim. Those two fixture values sit under `sha256`/`commit` keys where retention is legitimate, but the code path that retains them is the same one that retains an unrecognized-key secret.

Suggested direction (not prescriptive): drop `classes >= 3` to `>= 2` while keeping `distinct >= 10`; lower or remove the 48-byte floor for `Generic`-role values whose key contains (not equals) a secret-ish token; and switch `direct_secret_key_class` from exact match to a normalized-substring match over the same vocabulary.

### MEDIUM | CONFIRMED | `apps/packages/product-client/src/providers/ProductLifecycleRoot.tsx:236`

Prior finding 1, above. Restated as a finding for the record: the support-report queue owner — and therefore the *only* reconciler/reaper for queued jobs, staged attachments, and staged snapshot artifacts — mounts only while `authStatus === "authenticated"`, so an abandoned account's queued report and staged attachment bytes persist on disk indefinitely with no independent sweep.

### MEDIUM | CONFIRMED | `specs/codebase/systems/product/support/README.md:393` ("Current gaps")

The `Current gaps` section this PR adds states *"On `main` today, these are the only differences from the target state above"* and then lists gaps that this very PR closes. Concretely contradicted by files in the same diff:

- *"AnyHarness exposes only the ordinary history routes and SDK reads; the three SQL-bounded, byte-bounded, cancellable support-window endpoints and generated SDK methods do not exist"* — this PR adds exactly three such routes (`api/router/support_window_routes.rs`), their handlers (`api/http/support_windows.rs`, +533), and the SDK client (`sdk/src/client/support-windows.ts`, +555).
- *"Native export admission remains broker-local, support has no move-only permit, and the lifecycle producer has no fixed support prepare/submit constructors"* — this PR adds `diagnostics_collector/export_admission.rs` (shared `Arc<ExportAdmission>` handed to both the broker and support export) and `producer/support_lifecycle.rs` (+216).
- *"The Desktop bridge exposes the legacy schema-1 support bundle collector, while native still assembles bounded log tails through the old generic scrub instead of a consent-bound schema-3 coordinator, manifest, second scrub, opaque staging, or Save-a-copy archive"* — this PR removes `collect_support_diagnostics` from `lib.rs`/`commands/diagnostics.rs`, deletes it from `apps/desktop/src/lib/access/tauri/diagnostics.ts`, and adds the coordinator, manifest, second scrub, opaque staging, and `save_support_snapshot_archive`. `grep collectSupportDiagnostics` over `product-client/src/host/desktop-bridge.ts` returns nothing.
- *"Both modal flows still use `includeLogs`"* — `use-support-modal-state.ts:172` now sets `supportSnapshot: { kind: "none" }` and no longer sets `includeLogs` at all.

Because the doc declares itself `Status: target` with a gaps list as the *only* record of divergence, a stale gaps list is the failure mode that matters here: a reader cannot tell what is actually built. (The gap about the missing modal consent UI *is* accurate and correctly documents the reachability note above — that one is fine.)

### MEDIUM | CONFIRMED | `apps/desktop/src-tauri/src/diagnostics/support_snapshot/scrub/patterns.rs:576`

`provider_credential_regex` matches `sk-[A-Za-z0-9_-]{16,}` with a literal **hyphen**. Stripe secret and restricted keys use an underscore (`sk_live_…`, `sk_test_…`, `rk_live_…`) and never match this rule. Modern Stripe keys are long enough (~107 chars, four character classes) that the opaque catch-all rescues them *if* the value sits under a `Generic`/`Content`-role key; a shorter legacy `sk_test_` key (32 chars) or one under a `Hash`/`Identifier`/`Name`/`Timestamp`-role key is emitted in full. Adding `sk_(live|test)_` / `rk_(live|test)_` alternatives is a one-line fix and does not depend on resolving the HIGH above.

### MEDIUM | PLAUSIBLE | `anyharness/crates/anyharness-lib/src/domains/sessions/store/support_windows.rs:242-273`

`select_event_metadata` / `select_raw_notification_metadata` filter on `julianday(timestamp)` — an **unindexed** column — while ordering by the indexed `seq`:

```sql
FROM session_events WHERE session_id = ?1
  AND julianday(timestamp) >= julianday(?2) AND julianday(timestamp) <= julianday(?3)
ORDER BY seq DESC LIMIT ?4
```

The only relevant index is `idx_session_events_session_seq(session_id, seq)` (`persistence/sql/0001_initial.sql:42`); the timestamp bounds become a residual per-row filter walking that index backward from the newest row. The API bounds the *window width* to 15 minutes (`validate_time_window`, `api/http/support_windows.rs:424`) but never bounds **how far in the past** the window sits, so a far-past window on a session with a large event history must scan every newer non-matching row before it can collect `limit+1` matches. That scan runs inside `Db::with_interruptible_conn` (`persistence/sqlite.rs:97-121`), which holds the single process-wide `Mutex<Connection>` used by every other DB-backed route, with no statement timeout — `sqlite3_interrupt` fires only when the HTTP future is dropped (client disconnect), which a client keeping the connection open does not trigger. Contrast the pre-existing `list_session_events`, which paginates on indexed `after_seq`/`before_seq` with no residual filter.

What would settle it: `EXPLAIN QUERY PLAN` against a `session_events` table with ~500k rows for one `session_id` and a window near the oldest rows, confirming a full backward index walk rather than a targeted seek. Not run here (no-build constraint), hence PLAUSIBLE.

### LOW | CONFIRMED | `apps/desktop/src-tauri/src/diagnostics/support_snapshot/artifact_store/unix_fs.rs:18-38`

The artifact-root directory mode is applied by chmod-after-create, not atomically:

```rust
Err(error) if error.kind() == NotFound => match fs::create_dir(path) { Ok(()) => true, … },
…
let descriptor = open_root_descriptor(path)?;
if created {
    if unsafe { libc::fchmod(descriptor.as_raw_fd(), 0o700) } != 0 { return Err(Io); }
}
```

`fs::create_dir` is `mkdir(path, 0o777)`, masked only by umask — under a typical `umask 022` the directory exists at `0o755` for a couple of syscalls. Low impact: the directory is empty during the window, and artifact files themselves are created with `O_CREAT|O_EXCL|O_NOFOLLOW` and mode `0o600` requested at `openat`, which umask can only subtract from. `DirBuilder::mode(0o700)` closes it.

Secondary, same site: if `fchmod` fails, the function returns `Err(Io)` leaving a `0o755` directory in place. Every subsequent call sees `symlink_metadata → Ok`, takes `created = false`, skips the chmod, and `validate_directory`'s exact-`0o700` check fails forever → `UnsafeMetadata` → coordinator readiness never reaches `Ready` → the feature is permanently disabled until manual remediation.

### LOW | CONFIRMED | `apps/packages/product-client/src/hooks/support/lifecycle/support-report-queue-validation.ts:86,103`

This TypeScript source file contains two **raw U+0000 bytes** (used as a join/domain delimiter), so git classifies it as binary — it appears in the diff as `Bin 0 -> 3908 bytes`, with no reviewable diff, no line-level blame, and no textual merge. The sibling test uses the escape form correctly (`support-report-queue-controller.test.ts:412`, `proliferate-support-snapshot-v1${jobId}`); the production file should do the same. The derived value is otherwise correct and matches the Rust `ARTIFACT_DOMAIN` (`artifact_store.rs:18`, `b"proliferate-support-snapshot-v1\0"`).

### LOW | CONFIRMED | `apps/desktop/src-tauri/src/commands/support_snapshot/mod.rs:127-147`

`all_nine_commands_share_the_main_window_gate` is vacuous. It passes the nine **command names** as window *labels* and asserts each is rejected — i.e. it only proves `require_main_window` rejects nine arbitrary non-`"main"` strings. It never touches the commands. A tenth command added without `require_main_window` would still pass this test under its own name.

### LOW | PLAUSIBLE | `apps/desktop/src-tauri/src/diagnostics/support_snapshot/scrub/patterns.rs:582-588`

`authorization_scheme_regex` covers only `(?:bearer|basic)`. An unlabeled `Digest username=… response=…` or `token abc123…` fragment in free text — with no preceding `authorization:`/`auth-header:` label, so `authorization_label_regex` does not fire either — is not redacted by any rule. Low realism (real Authorization values usually appear labeled, and the label regex captures the whole line regardless of scheme), but it is a genuine scheme-coverage gap.

---

Things checked hard and found genuinely handled — recorded so the next reviewer does not re-derive them:

- **Capture/staging before consent.** `begin_preparation` validates consent shape, disclosure version, selection identifiers and time ordering as its first two statements, before any I/O (`coordinator/preparation.rs:38-40`). Every later command re-checks `consent_epoch` against the open preparation (`preparation.rs:117,226,376`, `artifacts.rs:232`, `cancellation.rs:33,68,96`, `closing.rs:46-47`).
- **A second, parallel export path.** Only one transmit path exists: `createSupportReport → createSupportReportUploadTargets → putPresignedObject → completeSupportReportUpload` (`support-report-upload-attempt.ts:78-216`). `saveArchive` writes to a user-chosen local path and never to the network. ADR "existing support attachment path only" is satisfied.
- **`fs::remove_file` following symlinks.** It is `unlink(2)`; it does not. (This was one of the two false HIGHs on the previous round — not repeated.)
- **Symlink / hardlink substitution into the snapshot.** Every directory hop uses `openat(O_DIRECTORY|O_NOFOLLOW|O_CLOEXEC)` and every leaf `openat(O_RDONLY|O_NOFOLLOW|O_CLOEXEC)`; each fd is `fstat`'d for type, `st_uid == geteuid()`, exact mode, and `st_nlink == 1` (`child_bridge/fallback_root.rs:116-219`, `support_evidence_io.rs:257-297`). The artifact store additionally uses `O_CREAT|O_EXCL`, `RENAME_NOREPLACE`/`RENAME_EXCL`, and `fstatat(AT_SYMLINK_NOFOLLOW)` requiring `S_IFREG` + `st_nlink == 1`, with a dedicated test (`artifact_store_tests.rs:333-366`).
- **Worker target id as a path-traversal vector.** Sanitized by `sanitize_path_segment` (`commands/cloud_worker.rs:548`) *and* independently re-validated component-by-component against `app_dir` in `read_worker_legacy` (`support_evidence_unix.rs:250-265`).
- **Unbounded file reads.** Three caps combine via `.min()` into `source_budget` *before* the buffer is allocated: `ALL_FILES_READ_BYTES` 10 MB global, `COMPONENT_FILES_READ_BYTES` 2 MB per family, per-source `active_budget` (`support_evidence_unix.rs:347-354`, `support_evidence_io.rs:164-169`). Worker legacy targets capped at 10, all drawing from the same 2 MB family budget. `active_budget: u64::MAX` on the two legacy sources looks alarming but is still bounded by the family and global budgets.
- **Collector auth token / broker socket path leaking into the snapshot.** Grepped all export and evidence modules for `token|socket|env::var|argv|home_dir|HOME`: no hits. `support_export.rs` carries only schema types.
- **Over-collection beyond the consent scope.** `exact_request` hardcodes the component list, record classes, and limits and forbids every singular filter (`support_export.rs:423-518`); `validate_support_request` re-checks the request is byte-for-byte that shape; `SupportExportAccumulator` re-validates every streamed frame including cursor-fence containment. No renderer or child input reaches this path.
- **Export admission permit dropped early / deadlocking / wedging on shutdown.** `let _admission = …try_acquire_owned()` is bound for the whole of `execute_support_export` (`:251-253`) and released on every early return, on panic (unwind drop), and on tokio task abort. `try_acquire_owned` is non-blocking, so a same-task double acquire returns `Busy`, never blocks. Broker and support export share one `Arc<ExportAdmission>` created at `supervisor/initialization.rs:56`, with a test asserting they contend for the same slot.
- **Concurrent captures, cancelled-capture-still-produces-an-artifact, deadline racing finish.** `CoordinatorState.preparation`/`.submission` are single `Option`s guarded by one mutex; `ClosingPreparation::claim_owner()` is an `AtomicBool::swap` so exactly one of {finish, cancel, watchdog} finalizes; a post-stage interruption deletes the artifact inline under the held `artifact_gate`; `belongs_to_closing` prevents deleting a *newer* artifact staged under the same deterministic id.
- **Windows silently creating world-readable artifacts.** `artifact_store/unsupported.rs` returns `Err(Unsupported)` for every operation including reconcile, so readiness never reaches `Ready` and preparation fails closed. The feature is unavailable there, not insecure.
- **`<app_dir>/logs` failing the `ActiveFallback` exact-`0o700` policy.** The fallback writer chmods that exact directory to `0o700` on open (`diagnostics_collector/fallback.rs:38` → `set_owner_only_dir`), and `desktop-native.log` lives directly in it (`lib.rs:190`). Reader and writer agree.
- **anyharness authorization / IDOR.** `list_sessions_support_window` uses `assert_workspace_auth_scope` and the two evidence routes `assert_session_auth_scope`, identical to their peers; the new router is merged into `v1` before `.route_layer(require_bearer_auth)`. `select_session_metadata_exact` filters `WHERE workspace_id = ?1 AND id = ?2` so a cross-workspace session id yields zero rows regardless. `user_route_allowed` has no arm for the new paths, so JWT-scoped direct-attach callers get 403 — fail-closed.
- **anyharness data scope.** `load_support_session` never selects `mcp_bindings_ciphertext` or `agent_auth_contexts`; there is a store test asserting exactly that. All SQL uses bound `params!`; every query carries an explicit `LIMIT`; `limit`/`max_response_bytes` are validated server-side.
- **anyharness SQLite migration risk.** There is none in this diff — the 251 added lines in `persistence/sqlite.rs` are the in-process `InterruptibleQuery` mechanism; `migrations.rs` is untouched.
- **SDK `bounded-json.ts` buffering before it checks the cap.** It streams via `response.body.getReader()` and checks each chunk against the remaining budget before appending; `Content-Length` is pre-validated when present. `abort-signal.ts` removes listeners on both branches and uses `{ once: true }`.
- **`report-upload-sanitizer.ts` deletion leaving a hole in the *web* path.** The browser queue controller explicitly strips `supportSnapshot` on every write and is the pre-existing web fallback, not the Desktop path.
- **Retry re-uploading without re-consent.** Retries re-verify and re-submit the *same* already-consented artifact by opaque id; they never recollect. Legacy v1 jobs are migrated to no-snapshot and, if the server report is already locked to `diagnostics=true`, concluded as `consent_required_for_legacy_job` rather than resumed under stale consent.
- **Spec README describing UI that does not exist.** Not a finding: the doc declares `Status: target` and points at its gaps list, and the missing-consent-UI gap is accurately recorded. (The *stale* gap entries are the MEDIUM above.)
- **`rfd::FileDialog` used synchronously from an async Tauri command** (`commands/support_snapshot/dialog.rs`) — same pattern as the pre-existing `commands/diagnostics.rs:200,230`; no new hazard introduced here.
- **Dangling `collect_support_diagnostics` invoke after its removal.** No caller remains in any `.ts`/`.tsx`/`.rs` outside `dist/`.
- **Snapshot package cap only asserted, not enforced.** It is genuinely enforced. `assembly/mod.rs:202-243` runs a bounded fixed-point loop that re-materializes, re-serializes, and pops the lowest-priority group until `serialized_bytes <= package_cap`, then re-serializes once more and errors if the exact `compact_json.len()` disagrees. `PACKAGE_BYTES` (`schema/limits.rs:27`) equals `ARTIFACT_MAX_BYTES` (`artifact_store.rs:20`) exactly (26,214,400), and `stage()` re-checks. The mandatory skeleton is separately capped at `MANIFEST_BYTES` before any candidate is admitted. Per-source caps do sum to more than the package cap, but that is by design — the degradation loop is the real enforcement.
- **Silent evidence loss on truncation.** Every package-cap removal calls `note_package_removal` (`assembly/accounting.rs:170-190`), which bumps `removed_by_tier` *and* records both an omission and a truncation with item count and byte count; dropped gap entries beyond `MAX_GAP_ENTRIES` are counted in `manifest.additional_entries.gaps`.
- **Non-deterministic ranking breaking the golden fixtures.** `priority_cmp` (`assembly/ranking.rs:216-222`) is a total order ending in `original_index`, over integers and enums only — no floats, no `partial_cmp`, no `HashMap` iteration feeding a sort.
- **Per-source `included_bytes` being self-reported rather than cross-checked against real serialized size.** Raised during review and rejected on the field's semantics: `included_bytes` means *source bytes included*, not serialized JSON size, and `assembly/input.rs:36-38` says so explicitly. It is not an estimate either — `coordinator/byte_allocation.rs:7-34` allocates one **exact measured** response-body total across the retained atoms so the parts sum to the measured whole. The caps it is checked against (`DESKTOP_NATIVE_FALLBACK_BYTES`, `COMPONENT_FILES_READ_BYTES`, …) are source-byte caps in the same unit, independently enforced at the reader with real byte counts before allocation. Units match; there is no truthfulness gap.
- **`SupportJsonValueV1::Deserialize` materializing an unbounded `serde_json::Value` before applying `NESTING_DEPTH`/`CONTAINER_ITEMS`.** The only live path is `coordinator/session_input.rs:182-186`, which rejects the string at `SESSION_EVIDENCE_BYTES` (8 MiB) *before* parsing, and serde_json's default 128-frame recursion limit is in force (`disable_recursion_limit`/`unbounded_depth` appear nowhere in the repo). Bounded on both axes.
- **Panics on slice indexing in assembly/validation.** The two `gaps[..n]` sites (`assembly/presentation.rs:164`, `schema/validate/aggregate.rs:282`) both derive `n` as `gaps.len().min(MAX_GAP_ENTRIES)`; the one `unreachable!()` (`presentation.rs:218`) is the second arm of an exhaustive two-variant match. Non-finite and out-of-JS-range numbers are rejected at parse (`schema/model/common.rs:40-51` → `validate_support_number`/`validate_safe_u64`/`validate_safe_i64`). All eight `negative_*.json` schema fixtures are referenced by a test.

pablonyx:
## Post-review fixes + restack — head is now `c7bc54cd8`

Force-pushed with lease. Nine commits since the sealed review head `a1ebd5779`:

**Scrub HIGH (`df2cad2a7`).** The opaque catch-all required three character classes, so a bare lowercase hex credential (two classes) was never redacted at any length — a narrowing versus the superseded `{48,}` rule, not a refinement. Now two classes + the distinct-byte floor. The key allowlist was also prefix-blind (`authToken` could never reduce to `token`) — closed by final-word segmentation, deliberately not substring matching (`monkey`, `token_count`, `input_tokens` stay readable). Five negative controls. Hash-role fields (`sha256`, `commit`) are retained by semantic role; the same hex in free text is redacted — both directions tested.
**Scope correction from the fixing pass:** the new scrubber currently has **no reachable caller** — `bundle.rs` imports the *old* `scrub_diagnostic_text`, and the modal hardcodes `kind:"none"`. The fix matters because the consent surface ships on top of it, but nothing on this branch routes user data through it yet. The still-reachable legacy `export_debug_bundle` runs the old, weaker scrub — recorded in the README.

**Inert "Include app logs" checkbox (`2278c9117`), user-facing.** The modal rendered a live, default-checked "Include app logs" while the job hardcodes `supportSnapshot:{kind:"none"}` and never carries `includeLogs` — so users ticked a box and no logs were attached. The control is **removed** rather than rewired: reviving the implicit path would undo this slice's deliberate move to explicit prepared-snapshot consent. Ingest-side `includeLogs` handling stays for legacy persisted jobs.

**Also:** auth-independent retention sweep for the queued-report document (`73c3de2bf`); README "Current gaps" rewritten against reality (`0f772b675`); NUL-byte artifact-id binding (`361e1e487`); `unix_fs` mkdir-mode TOCTOU hardening (`a6e91e92c`); a vacuous window-gate test made real (`1b9169762`); a restack-only integration fix passing the new 7th `SupportSnapshotCoordinator` argument to #1757's shutdown tests (`5a8c8cd18` — neither branch was broken alone); max-lines ledger (`c7bc54cd8`).

Verification at the restacked head: `cargo test -p proliferate --lib` **542 passed / 0 failed**; product-client vitest 5761 passed; typecheck clean.

**Relabelled `release:maintenance` → `release:minor-feature`:** removing a visible control and changing what leaves the user's machine is user-visible and belongs in the changelog.

**Merge-order note:** land in stack order with #1757 (see the note there); slice 7 (#1831) stacks on this PR.

PR #1822, opened 2026-08-13
fix(agent-auth): hold the harness settings toggle for a native-auth harness (PRO-129)

pablonyx:
## Summary

- Fixes PRO-129: the "Use Claude Code with Chrome" switch in the Claude harness settings pane snapped back to off after every click for a native-login harness (the default local Claude setup), so the control appeared dead.
- Root cause: the toggle persists correctly (`agent_auth_harness_settings`), but the settings pane read the value back from the rendered `state.json` document, which only carries a harness's `settings` passenger when that harness has an enabled auth selection — the fail-closed law ("absent means native; present-but-empty fails closed") forbids a settings-only `harnesses` entry, so a native-auth harness's persisted settings were invisible on read.
- Fix: `GET /v1/cloud/agent-auth/state` now carries the surface's full persisted settings map as a second **response-only rider** (`harness_settings`), following the existing `fingerprint` rider pattern. The settings pane reads the rider (falling back to the delivered per-harness passenger for pre-rider servers), and the desktop strips it in `stampIssuingServerOrigin` before the runtime push — the delivered `state.json` contract and the cloud materializer output are byte-identical to before, so no deployed runtime sees a new shape.
- Also hardened the toggle against a latent data-loss edge: the PUT is full desired state, so committing while the selections read was still loading would have silently cleared the harness's auth sources. The switch is now disabled until both reads resolve.
- `cloud/sdk/src/generated/openapi.ts` is regenerated via `make cloud-client-generate` (only the new field).

Known gaps recorded in `specs/FEATURE_DOCS/AGENT_AUTH.md` "Current gaps" (out of scope here):
- A native-auth harness's settings still never reach the runtime's `resolve_settings_deltas` join — carrying them needs a wire-contract change the deployed runtime fleet can survive, plus the matching Rust read.
- The `--chrome` mapping lands on the agent-process sidecar's argv, and the pinned `@proliferate/claude-agent-acp` does not forward unrecognized argv to the wrapped CLI (separate repo + catalog pin bump).

## Testing

- Tier 1 unit (frontend, vitest): new `HarnessSettingsSection.test.tsx` — rider read for a native-auth harness (verified red against the unfixed read as a negative control), passenger fallback for pre-rider servers, PUT payload shape, and the no-commit-while-loading guard. Extended `local-auth-state.test.ts` — the rider is stripped before the runtime push. `pnpm vitest run src/components/settings/panes/agents/harness/ src/hooks/agents/lifecycle/use-local-auth-state-sync.test.tsx src/lib/domain/agents/local-auth-state.test.ts`: all passing (8 files / 89 tests for the harness+sync sweep, 17 for the two focused files).
- Tier 1 unit (server, pytest): new `tests/unit/test_agent_auth_settings_rider.py` — settings without a selection never ride the document (pins the render law), the payload carries the rider and it stays out of the canonical document/fingerprint, and the rider coexists with the delivered passenger. Neighboring suites `test_agent_auth_materialization.py`, `test_agent_auth_state_contract_fixture.py`, `test_agent_gateway_domain.py`: 51 passed.
- `tsc --noEmit` (product-client), `ruff check`/`format --check` (changed server files), `check_server_boundaries.py`, `check_frontend_boundaries.py`, `check_design_attribution.py`, `check_docs.py`: all passing.

## Observability

- none (no new logs/metrics; the rider is a read-path response field).

## Readiness

  in this PR.

## Verification

- Negative control: reverting only the component's rider read makes `HarnessSettingsSection.test.tsx > reads the persisted toggle from the harness_settings rider for a native-auth harness` fail; restoring the fix turns it green.
- Delivery-contract safety: the renderer (`render_agent_auth_state`) is untouched — the rider is attached at the API payload layer only, the cloud materializer path never sees it, and the desktop strips it before `apply_state_file`, pinned by the new strip test.

## CI repair (commit 516e58ee5)

- `agent_gateway/service.py` had grown to 827 lines, past its 813-line max-lines allowlist entry. `put_harness_settings` moved verbatim to a new `server/cloud/agent_gateway/harness_settings.py` service module — settings are explicitly not auth, and the store layer already draws this exact seam (`db/store/agent_gateway/harness_settings.py`). `api.py` now calls the new module; a local variable in the state endpoint was renamed to avoid shadowing the module import.
- **Constitution flag — allowlist entry removed, not added:** `service.py` lands at 799 lines, back under the 800 server-layer cap, and `check_max_lines.py` fails on stale allowlist entries, so its `scripts/max_lines_allowlist.txt` line (813) is deleted. This removes an exception (ratchet tightening); the checker itself and all coverage are untouched.
- Re-verified: `check_max_lines.py`, `check_server_boundaries.py`, `check_docs.py` pass; `ruff check`/`format --check` clean; focused suites green (`test_agent_auth_settings_rider.py` 3 passed; the four agent-auth/gateway unit suites 54 passed); app-import smoke of the rewired modules passes. No wire-contract change, so the generated SDK is untouched.

## Lint repair (commit 71eac119f)

- The two rider annotations in `agent_gateway/models.py` tripped the mypy explicit-any ratchet. They now use `dict[str, dict[str, object]]` — the type `service.get_auth_state` already returns; the values are catalog-declared JSON primitives validated at the write path. Pydantic renders `object` and `Any` to the same unconstrained schema, so `make cloud-client-generate` produces a byte-identical `openapi.ts` (SDK contract unchanged). Verified with the CI lint command: `check_mypy_baseline.py --compare-ref origin/main` passes (523 existing diagnostics, none new), ruff check/format clean, and the four agent-auth/gateway unit suites still pass (54).

### DIFF
--- a/apps/packages/product-client/src/components/settings/panes/agents/harness/HarnessSettingsSection.test.tsx
+++ b/apps/packages/product-client/src/components/settings/panes/agents/harness/HarnessSettingsSection.test.tsx
@@ -1,0 +1,104 @@
+// @vitest-environment jsdom
+
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import type { AgentAuthSelection, AgentAuthState } from "@proliferate/cloud-sdk";
+import { HarnessSettingsSection } from "#product/components/settings/panes/agents/harness/HarnessSettingsSection";
+
+const queries = vi.hoisted(() => ({
+  state: { data: undefined as AgentAuthState | undefined },
+  selections: { data: undefined as AgentAuthSelection[] | undefined },
+  mutate: vi.fn(),
+}));
+
+vi.mock("@proliferate/cloud-sdk-react", () => ({
+  useAgentAuthState: () => queries.state,
+  useAuthSelections: () => queries.selections,
+  usePutAuthSelections: () => ({ mutate: queries.mutate }),
+}));
+
+vi.mock("#product/hooks/cloud/derived/use-cloud-availability-state", () => ({
+  useCloudAvailabilityState: () => ({ cloudActive: false }),
+}));
+
+function stateDocument(overrides: Partial<AgentAuthState> = {}): AgentAuthState {
+  return { version: 2, revision: 3, user_id: "user-1", harnesses: [], ...overrides };
+}
+
+function renderClaudeSettings() {
+  return render(<HarnessSettingsSection harnessKind="claude" surface="local" />);
+}
+
+afterEach(() => {
+  cleanup();
+  queries.state.data = undefined;
+  queries.selections.data = undefined;
+  queries.mutate.mockClear();
+});
+
+describe("HarnessSettingsSection", () => {
+  it("reads the persisted toggle from the harness_settings rider for a native-auth harness", () => {
+    // PRO-129: a native-login harness has NO `harnesses` entry in the rendered
+    // document (absent means native), so the persisted value must come from
+    // the response's harness_settings rider or the switch snaps back to its
+    // default after every toggle.
+    queries.state.data = stateDocument({
+      harness_settings: { claude: { chrome: true } },
+    });
+    queries.selections.data = [];
+    renderClaudeSettings();
+    expect(
+      screen
+        .getByRole("switch", { name: "Use Claude Code with Chrome" })
+        .getAttribute("aria-checked"),
+    ).toBe("true");
+  });
+
+  it("falls back to the delivered settings passenger when the rider is absent", () => {
+    // Older self-hosted servers respond without the rider; a routed harness
+    // still carries its settings inside the document.
+    queries.state.data = stateDocument({
+      harnesses: [
+        {
+          harness_kind: "claude",
+          sources: [{ kind: "gateway", base_url: "https://gw", key: "sk-vk" }],
+          settings: { chrome: true },
+        },
+      ],
+    });
+    queries.selections.data = [];
+    renderClaudeSettings();
+    expect(
+      screen
+        .getByRole("switch", { name: "Use Claude Code with Chrome" })
+        .getAttribute("aria-checked"),
+    ).toBe("true");
+  });
+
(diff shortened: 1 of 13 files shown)


pablonyx (commented):
Rebased onto main at 0a4fab48c (resolved a modify/delete conflict on the now-retired `scripts/max_lines_allowlist.txt`, superseded by `lints/server/ratchets.toml` per D6/#1722 — dropped the split commit's allowlist edit and instead shrank the `service.py` ratchet entry to its new 820-line count; `scripts/check_max_lines.py` passes locally). The rider design itself is untouched by the FR-3/FR-4 restructuring — auto-merged cleanly against both.

Review of the rebased diff:

- **Fail-closed law**: preserved. `harness_settings` stays a response-only rider (`models.py` docstring + `get_auth_state` explicitly document that the rendered `harnesses[].settings` passenger only rides when a harness has an enabled selection, and that this rider exists precisely because the fail-closed rule forbids a settings-only entry). No path lets a harness get a settings passenger without an enabled selection.
- **Stripped before runtime push**: single seam in `local-auth-state.ts` (`stampIssuingServerOrigin`) destructures both `fingerprint` and the new `harness_settings` key before spreading the rest into the runtime document — same strip point as the existing fingerprint rider, so there's no second push path that could leak it.
- **Settings-pane read**: `HarnessSettingsSection.tsx` reads `stateQuery.data?.harness_settings?.[harnessKind] ?? harness?.settings`, falling back to the old per-harness passenger for servers that predate the rider — reasonable compat shape.
- **Loading-guard race**: the toggle's `disabled` now also gates on `stateQuery.data === undefined || selectionsQuery.data === undefined`, on top of the existing `!isLocalSurface && !cloudActive` check. That closes the race where a toggle flip before both reads resolve would PUT a partial desired-state and clobber sibling settings/auth sources.
- **Tests**: `test_agent_auth_settings_rider.py` (62 lines) and `HarnessSettingsSection.test.tsx` (104 lines) cover the new rider plumbing and the toggle-read-precedence/disabled-state behavior.

No blocking findings. Per operator instruction this PR is NOT to be merged or approved by me — leaving as comment-only.

PR #1828, opened 2026-08-13
fix(product): keep the repository options trigger highlighted while its menu is open (PRO-133)

tairabun:
## What

The '…' (Repository options) button next to the **Repositories** sidebar header lost its hover highlight the moment the cursor left it, even while its popup menu was still open.

## Why

`SidebarActionButton` destructured only its named props and forwarded no rest props, so the `data-state`/aria attributes Radix's `PopoverTrigger asChild` injects never reached the DOM button. Every `data-[state=open]:*` class on the adapter (and the base `RowActionIconButton`'s open-state/reveal contract) was inert for this trigger — the only styling left was `:hover`, which ends when the cursor leaves.

## How

- Forward rest props through `SidebarActionButton` → `RowActionIconButton` → `<button>`, matching how direct `RowActionIconButton` triggers (e.g. the workspace row kebab) already work. This also restores the injected `aria-haspopup`/`aria-expanded` attributes.
- Style the open state to mirror the hover ink: `data-[state=open]:text-sidebar-foreground` plus `data-[state=open]:opacity-100` on the section variant — no background chip, per the adapter's existing Round-3 doctrine (`data-[state=open]:bg-transparent` is already there and now actually applies).
- Contract test: renders the adapter as a real `PopoverButton` trigger, asserts `data-state` lands on the DOM and flips to `open` on click, and that the open-state classes are present.

## Testing

- `pnpm --filter @proliferate/product-client typecheck` clean.
- Adapter + popover + sidebar consumer suites pass (32 tests).
- Manual: open the app → hover the '…' next to *Repositories* (highlights) → click to open the menu → move the cursor away. The button must stay highlighted (dark ink, full opacity) until the menu closes.

Linear: [PRO-133](https://linear.app/proliferate-team/issue/PRO-133/popup-stays-open-without-highlighting-the-options-button)


tairabun:
@greptile review

tairabun:
@greptile review

PR #1842, opened 2026-08-13
fix(product): keep mid-creation workspaces alive across a workspace switch (PRO-230)

tairabun:
## Summary

- Sending a prompt from Home and clicking another workspace while the worktree was still creating destroyed the new workspace and blocked every session in every other workspace with the "Check this thread before retrying / The message may have reached the session." pane. Linear: [PRO-230](https://linear.app/proliferate-team/issue/PRO-230/stitching-workspace-while-a-new-workspace-destroys-the-new)
- Root cause 1: the sidebar switch clears `pendingWorkspaceEntry`, so when the create call finished, `createWorktreeAndEnterInternal` saw a stale attempt, returned null, and `createWorktreeAndEnterWithResult` threw "interrupted" — failing the launch intent as `unknown_after_send` (the queued prompt had already marked `sendAttemptedAt`) and abandoning the created workspace.
- Root cause 2: `resolveLaunchIntentSurfaceOverride` applied the global launch intent to any session whose id differed from the intent's session, so the failed intent pane covered every session in every workspace until dismissed.
- Fix 1 — background completion: when a local/worktree creation finishes and its attempt is no longer current, `completePendingWorkspaceCreationInBackground` remaps the projected sessions (and their queued prompt) to the created workspace and materializes them with a new `skipSessionActivation` option, so nothing steals the selection the user switched to. The launch flow then resolves normally: no failure state, no "Work not started" toast, intent cleared, prompt dispatches, and the workspace row appears when collections refresh.
- Fix 2 — scoped launch-intent surface: an intent bound to a session owns only that session's surface. A different active session, or another selected workspace, keeps its own transcript/empty state — including while the intent holds a failure. Home (no workspace context) still shows the intent pane.
- Updated the pending-shell canonical doc for both behaviors; inlined two thin wrapper callbacks in `use-workspace-entry-actions.ts` to stay under the 400-line structure cap.
- Follow-up in this PR (sidebar row stability): an in-flight local/worktree pending entry now survives switching to another workspace, so the new workspace's sidebar row holds one slot through the whole creation lifecycle — it no longer vanishes on switch and reappears elsewhere. Clicking the deselected pending row re-enters the pending shell. Background completion stamps `workspaceLastInteracted` with the entry's creation time (the pending row's sort key) and clears the entry itself. Pending state is scoped out of other selected workspaces' surfaces via `pendingWorkspaceEntryOwnsSelection` (composer availability, chat surface arbitration, right-panel suppression).
- Behavior-pinning test updated intentionally: `session-selection-store.test.ts` "activates workspace… atomically" previously pinned `pendingWorkspaceEntry: null` after activation; the survival of in-flight local/worktree entries is the point of this change, and new cases pin that failed and cloud entries still clear.

## Testing

- Unit (Tier 1 per specs/TESTING.md): added `resolveLaunchIntentSurfaceOverride` cases (unrelated session, session-less other workspace) and `resolveChatSurfaceState` cases (failed intent elsewhere keeps the other session's transcript / empty state) in `chat-surface.test.ts`.
- `pnpm vitest run src/hooks/workspaces src/hooks/sessions src/hooks/chat` — 126 files, 598 tests green; `pnpm typecheck` green; `python3 scripts/report_frontend_structure.py --strict` clean; `python3 scripts/check_docs.py` passed.
- Manual: send a prompt from Home targeting a new worktree, click another workspace while "Creating worktree" shows, verify the other workspace stays usable, no retry-blocked pane appears anywhere, and the new workspace appears in the sidebar with the prompt running.

## Observability

- New latency events on the existing measurement port: `workspace.entry.background_completion` and `workspace.entry.background.*` (projected-session materialization events under the background prefix). No scrubber-relevant fields; no other delta.

## Readiness

  in this PR.

## Verification

- `pnpm typecheck` (product-client, builds dep chain) — pass
- `pnpm vitest run src/hooks/workspaces src/hooks/sessions src/hooks/chat src/lib/domain/chat` — pass
- `python3 scripts/report_frontend_structure.py --strict --summary-only` — 0 violations
- `python3 scripts/check_docs.py`, `python3 scripts/check_design_attribution.py` — pass


tairabun:
@greptile review

tairabun:
@greptile review

PR #1847, opened 2026-08-13
feat(observability): give the runtime a voice — named session/turn events, ACP frame tee, dev activation

pablonyx:
The diagnostics pipeline works end to end, but it carries almost no content. A
live dogfood run produced 1,196 records, 99% of them renderer startup noise; the
AnyHarness runtime contributed **zero**, and no session, turn, or provider event
had a name. This PR gives the runtime a voice.

## What this adds

**1. Record names via target pass-through.** Previously only two tracing targets
mapped to names; everything else exported as anonymous `anyharness.tracing.event`.
Now any target starting with `anyharness.` passes through verbatim as the record
name, so new events need no registration. Same rule for the worker with
`desktop_worker.`.

**2. Core session and turn lifecycle.** `anyharness.session.spawn`,
`.established`, `.exited`; `anyharness.turn.started`, `.finished`, `.failed`
(logging the error classification that was previously computed and dropped);
`anyharness.agent.process_exited`; and `anyharness.session.event_injected`,
which closes the one untraced sink path — the path subagent completions ride.

**3. ACP JSON-RPC frame tee.** Every frame in both directions at DEBUG
(`anyharness.acp.send` / `.recv`), including notifications. Logs the raw wire
line without re-serializing, capped at 8 KB on a char boundary, and guarded by
`tracing::enabled!` so it costs nothing when nothing is listening.

**4. Subagent lifecycle.** Spawn requested/rejected (with the rejection reason —
previously silent typed errors), spawned, link closed, turn completed.

**5. Dev activation fallback.** With `ANYHARNESS_DEV_URL`, an externally
launched runtime never inherits the control-bridge fd, so its producer stayed
disabled — this is why the runtime contributed nothing. A `debug_assertions`-only
env fallback lets a `make dev` runtime attach. Release bundles cannot enable it.

Plus docs: the dev-profile guide and `env-vars.yaml` now describe the five
`PROLIFERATE_DIAGNOSTICS_*` inputs, none of which were documented.

## Verification

`cargo check` clean across the four touched crates. Tests: 301 passing
(diagnostics-client, anyharness, worker), including new table tests for the
pass-through rule and frame-cap truncation.

Not yet verified live: DEBUG records have never been observed reaching Honeycomb
(0 of 106 over seven days), while a local receiver saw 1,164 through the same
exporter. Since the frame tee is DEBUG, that needs an end-to-end check before we
rely on it.

## Known rough edges

- The dev fallback has no bridge, so a collector restart silently ends delivery
  until the runtime restarts.

## Review notes

PR 9 of the rust-observability train, based on PR 8. Review PR 8 first.


pablonyx:
## Live verification on a `make dev` profile (obs9)

Booted this branch's runtime against a local collector + OTLP receiver. Two
results, one good and one blocking.

**The dev activation fallback works.** The runtime logs
`[desktop-diagnostics] dev activation against http://127.0.0.1:55500` and
runtime records reach the pipe for the first time — 683 in the first 20
seconds, against zero on every prior dogfood run. Note the two-pass mechanic:
the host writes the env snippet only after it is healthy, so the runtime that
`make dev` launched during that same boot still has nothing to attach to. It
attaches when the runtime is restarted with the snippet sourced. `make dev`
does not source it today, so this is a manual step:

```
set -a; . ~/.proliferate-local/dev/profiles/<profile>/app/diagnostics-dev.env; set +a
```

Wiring that into the recipe is worth a follow-up.

**DEBUG does reach the exporter.** `anyharness.acp.send` arrived as DEBUG,
which answers the open question in the description — the export path is not
dropping DEBUG.

**But records do not survive the producer's admission path, and that includes
INFO lifecycle records.** One grok session and one prompt turn:

| | emitted (console) | delivered (pipe) |
|---|---|---|
| `anyharness.acp.send` | 15 | 6 |
| `anyharness.acp.recv` | 68 | **0** |
| `anyharness.turn.finished` | 1 | **0** |

`session.spawn`, `session.established` and `turn.started` did arrive, so the
naming pass-through itself is working. The losses show up in the producer's own
`anyharness.diagnostics.loss` summaries, split across two channels — a quiet
run reported 113 `filter_invalid` and 66 `pressure`. Under a chattier filter it
degenerates: 1,511 dropped by pressure and 408 loss-summary records, more
volume in loss accounting than in real payload, since loss summaries are
protected from suppression and the DEBUG records they describe are not.

`pressure` is explicable — inbound frames arrive in dense bursts during
streaming and unprotected DEBUG is exactly what the valve sheds — but
`filter_invalid` is not yet root-caused, and `turn.finished` being lost means
this is not confined to the frame tee.

I would not rely on the ACP frame tee for debugging the subagent PR until this
is understood. The rest of the branch (naming pass-through, session/turn
lifecycle, dev activation) is verified working.

Repro: boot any dev profile with the snippet sourced, run one prompt turn, and
diff the console's `acp frame:` lines against `proliferate.name` in the
receiver output.

pablonyx:
## The pipe now delivers what it produces

Following up on the live run above, both loss channels are closed. Two
commits pushed.

**Root cause 1 — no ingest filter.** The diagnostics layer is installed
without one, so every tracing event in the process reached the producer,
dependencies down to TRACE included; two thirds of a live run was anonymous
`anyharness.tracing.event`. That saturated the collector, which declared
back-pressure, and the producer sheds by severity alone — it cannot tell a
named lifecycle record from a dependency's TRACE, so an anonymous WARN
survived while `turn.finished` did not.

Admission now keys on deliberateness rather than severity: a target that
names its own record is admitted at any level, and WARN/ERROR are admitted
whatever their target. The rule lives on `TargetMappingConfig`, so the
runtime and the worker share one definition.

**Root cause 2 — one empty field killed the record.** `validate_bounded_string`
rejects the empty string, and rejects the whole record over it.
`filter_value` already strips every other value validation refuses, but
passed empty strings through. A JSON-RPC response has no method and a
notification has no id, so `unwrap_or_default` gave them `""` and the record
died — which is why only request frames survived, and why `turn.finished`
went missing too. The filter now drops the field.

**Measured, one session and one prompt turn:**

| | before | after |
|---|---|---|
| delivered / dropped | 4,955 / 4,398 | 444 / **0** |
| delivery rate | 53% | **100%** |
| dropped to pressure | 4,101 | 0 |
| `anyharness.acp.recv` | 0 | 77 |
| `anyharness.acp.send` | 6 of 15 | 15 |
| `anyharness.turn.finished` | 0 | 1 |

303 tests pass across the three touched crates, each fix with a negative
control: reverting the admission rule fails the admission table test,
reverting the empty-string fix fails the new filter test.

Worth knowing: with the anonymous flood gone, `anyharness.agent.stderr` is
now 346 of 444 records. It is genuinely useful, but if we want the pipe
leaner, capping its DEBUG is the next lever — say the word and I will.

The two open items on this PR are unchanged: the file-length lint failures,
and the login first-load budget inherited from the pr7/pr8 base.

PR #1859, opened 2026-08-14
fix(product): paint switches instantly with a clean loader instead of skeletons (PRO-182)

tairabun:
Fixes [PRO-182](https://linear.app/proliferate-team/issue/PRO-182/slightly-improved-latency-loading-skeletons-helps-a-lot-with).

## What was happening

Frame-stepping the ticket recording and measuring the live app (debug measurement engine + CDP-driven switches) shows the switch itself is already optimistic — selection state commits synchronously before any await. The felt latency is what happens after:

1. One urgent long task (~110–260ms dev) re-renders the whole shell before the first new frame.
2. `SessionTranscriptPane` defers the session identity (`useDeferredValue`), so the pane paints `TranscriptSwitchingPlaceholder` — four faint pulsing skeleton bars — while the new transcript mounts in the deferred lane: 60–90ms for small transcripts, **400–750ms for real ones**. In dark mode this reads as a dead blank pane (the recording's blanks).
3. Cold opens add the session-directory fetch on top.
4. On every navigation (including each Settings section click), callback-only workflow hooks subscribed always-mounted surfaces — sidebar, lifecycle root, right panel, command surfaces — to the router via `useNavigate`/`useLocation`, re-rendering them all per URL change.

## What this changes

- **Clean loader over transient skeleton** (the ticket's explicit ask): `TranscriptSwitchingPlaceholder` and `ChatLoadingHero` now render a centered `DotCellLoader` + dim label behind a disclosure-duration delayed fade (`animate-content-fade-in`), so sub-200ms hot switches never flash it and slow switches show a deliberate loader instead of a blank pane.
- **Page switches stop paying the router tax**: `use-workspace-navigation-workflow`, `use-app-navigation-command-actions`, `use-workspace-retire-actions`, and `use-right-panel-entry-actions` are callback-only, and now navigate via the `navigateApp` handoff (PRO-170 pattern, #1845) reading `window.location` at call time. Both hosts mount plain `BrowserRouter`, so the read is exact.
- **Build repair**: two composer controls imported `useWorkspaceShellActions` from the provider module and broke `make build` (same class as #1854).

## Verification

- Full product-client suite green (5,940 passed; only the two pre-existing local-Chrome Playwright suites fail on machines without Chrome). Typecheck green. Appearance-scaling guard green (census ratcheted down for the placeholder file).
- Live-verified in the running desktop client: shell swaps in the first commit, the "Switching chat" loader is present during switches and invisible on fast ones, hot switch to content ~135–175ms, sidebar/home/workspaces/worktree-creation navigation all work through the handoff.

## Deliberate non-changes (follow-up candidates)

- `WorkspaceProviders`' `useLocation` and the command palette's home-route check are render-semantic and keep their subscriptions.
- The transcript mount cost itself (virtualized already; remaining cost is per-row markdown) and the right panel's per-switch remount ("Loading file" hanging for seconds in the recording) are separate sinks worth their own tickets.

### DIFF
--- a/apps/packages/product-client/src/components/workspace/chat/surface/ChatLoadingHero.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/surface/ChatLoadingHero.tsx
@@ -1,5 +1,5 @@
 import { useChatLoadingSubstep } from "#product/hooks/chat/derived/use-chat-loading-substep";
-import { SkeletonBlock } from "#product/primitives/Skeleton";
+import { DotCellLoader } from "#product/primitives/DotCellLoader";
 import { ThinkingText } from "#product/primitives/patterns/ThinkingText";
 import { DebugProfiler } from "#product/components/diagnostics/DebugProfiler";
 import { useDebugRenderCount } from "#product/hooks/ui/debug/use-debug-render-count";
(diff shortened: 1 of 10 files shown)


tairabun:
@greptile review

PR #1864, opened 2026-08-14
fix(auth): desktop OAuth callbacks land on an HTML handoff page; surface swallowed auth errors (PRO-111)

pablonyx:
Fixes [PRO-111](https://linear.app/proliferate-team/issue/PRO-111/need-redirect-post-git-flow): after desktop-initiated GitHub sign-in, the system-browser tab was stranded blank on the GitHub authorize URL.

## Root cause

Desktop sign-in uses the unified identity flow, and GitHub always calls back to the shared `GET /api/auth/github/callback`. That endpoint returned a raw 302 to `challenge.redirect_uri`, which for desktop is the custom-scheme deep link `proliferate://auth/callback?code=...`. A browser cannot render a custom scheme: the OS handoff fires, the tab never commits a new page, and the user is stranded on a dead tab. The legacy desktop flow already had an HTML handoff page (`make_desktop_handoff_page`) but it was bypassed once GitHub moved to the shared callback. The same defect existed on the surface-scoped callback used by desktop Google/Apple link flows.

## Change

**Server**
- `complete_oauth_provider_callback` / `complete_oauth_provider_error_callback` now return a frozen `AuthCallbackRedirect(url, surface, error)` instead of a bare URL string. The `error` field also covers the token-rejection branch inside the success path.
- New `_provider_callback_response` dispatcher in `identity/api.py`: desktop-surface challenges get an HTML handoff page (success) or a new error-toned page (provider error) that fires the deep link itself and leaves a real page in the tab, with a manual fallback button. All other surfaces keep the raw 302 byte-for-byte.
- The branch key is the challenge's stored `surface`, never the URI scheme: mobile shares the exact `proliferate://auth/callback` string and relies on the OS intercepting the 302, so it must stay a redirect.
- The desktop error page fires the error deep link so the desktop app's pending flow resolves immediately instead of hanging until the 2-minute poll timeout.
- `make_desktop_handoff_page` is parameterized by provider (GitHub/Google/Apple); the legacy caller's rendered output is byte-identical.
- `_auth_error_url` fallback paths (no recoverable challenge) intentionally stay raw 302s.

**Web**
- `AuthShell` now renders `AuthState.issue` via a new `describeAuthIssue` mapper, at lowest precedence under active sign-in hook errors. Previously the issue published by `/auth/error` and failed callbacks had no consumer at all, so those failures were invisible.

**Docs**: `specs/server/auth.md` records the desktop handoff behavior and the surface-not-scheme rule.

**Ratchet note**: `scripts/max_lines_allowlist.txt` updated for `identity/api.py` (488 → 514) and `test_auth_flow.py` (2216 → 2315), plus a net-new entry for `test_identity_error_ownership.py` (603) which crossed the threshold from the dataclass-shape assertions. Flagging per the constitution; no restructuring done to dodge the ratchet.

## Testing

- Server: `pytest tests/integration/test_auth_flow.py tests/integration/test_web_beta_auth.py tests/unit/auth/ tests/unit/test_redirect_callback_pages.py` → 207 passed. New coverage: desktop success handoff (launch script present, session then completes via `/auth/desktop/poll`), desktop provider-error page (`tone-error` + error deep link), explicit web-stays-302 pin, and a new `test_desktop_pages.py` unit suite (escaping of hostile provider error text, launch_url gating).
- Negative control: forcing the dispatcher to treat desktop as web makes exactly the 6 desktop-surface tests fail (`assert 302 == 200`); restored and re-verified green.
- Unchanged negative controls: mobile error 302 test, all web callback tests, legacy `/auth/desktop/github/callback` HTML tests, auth leaf-boundary test.
- Frontend: product-client auth vitest (33 passed) incl. new AuthShell issue-rendering + precedence cases and `describeAuthIssue` unit tests; apps/web auth suites (26 passed) unchanged; typecheck clean.
- Manual verification (desktop happy path, auto-approve variant, deny path, web regression, `/auth/error` rendering) planned with Pablo before merge.


pablonyx:
Adversarial review pass complete (dedicated agent review). Server half verified clean: surface branch proven exhaustive across every return path, HTML escaping verified by execution (hostile redirect_uri + hostile error probes), no material change to auth-code exposure (PKCE-bound, 60s TTL, single-use, no external subresources so no Referer leak). Findings fixed in dacb5b740:

- **M1** stale `AuthState.issue` survived retry/cancel on desktop → symmetric `clearAuthIssueStatePatch`/`clearPublishedAuthIssue` at sign-in start (GitHub + SSO) and cancel, with orchestration tests
- **M2** no test proved the embedded code was exchangeable → handoff-page test now parses the deep link from the HTML and completes a real `/auth/desktop/token` PKCE exchange
- **L1** unvalidated provider error echoed into the page → only echoed when it matches a conservative oauth-code shape, generic detail otherwise
- **L2/L5/L8** provider label in the non-launch message, tightened `class="status tone-error"` assertion, doc note that no-challenge callbacks still 302 to `/auth/error`

Follow-ups deferred (tracked separately): reuse the richer `webAuthErrorPresentation` SSO copy table in `describeAuthIssue`; consider `Literal` typing for `AuthCallbackRedirect.surface`.

PR #1867, opened 2026-08-14
fix(product): scope the launch-intent pane to its own workspace (PRO-230)

pablonyx:
Slice PR0 of the [PRO-230](https://linear.app/proliferate-team/issue/PRO-230/stitching-workspace-while-a-new-workspace-destroys-the-new-workspace) launch-registry ladder (parent: PRO-236 Workspace Transitions).

## Problem

`useChatLaunchIntentStore.activeIntent` is a global singleton and `resolveLaunchIntentSurfaceOverride` had no workspace scoping: any active or failed launch intent replaced the transcript of any session that was not the intent's own materialized session. One failed Home launch blocked every session in every workspace with the "Check this thread before retrying" pane, violating pending-shell.md Invariant 4.

## Change

- `ChatLaunchIntent` gains `attemptId` and `targetWorkspaceId`; `resolveLaunchIntentScope` derives the workspace identities an intent may own (pending UI key and/or workspace id).
- New pure gate `launchIntentOwnsShell`: a scoped intent owns only the shell matching its scope; an unscoped intent (Home first-paint window) owns only a shell with no workspace selected. Applied in `resolveLaunchIntentSurfaceOverride` and the shell-mount gate in `use-main-screen-state.ts`.
- `markMaterializedIfActive` optionally carries `attemptId`; `markHomeLaunchIntentMaterializedFromPendingWorkspace` threads it from the pending entry independent of `workspaceId` resolution, so a launch that fails before a workspace exists is still scoped.

This slice deliberately does not touch `pendingWorkspaceEntry`, `activateWorkspace`, finalization, or the sidebar; later slices make launches survive workspace switches (PR1) and allow concurrent launches (PR2/PR3).

## Tests

New "launch intent scoping (PRO-230)" cases in `chat-surface.test.ts` (scoped intent never overrides another workspace's shell; failed intent never overrides another session; unscoped intent still owns an empty shell but not a selected one; matching pending shell preserved), plus `resolveLaunchIntentScope` / store coverage. Full product-client suite green: 856 files / 5983 tests; both tsconfig typechecks clean.

### DIFF
--- a/apps/packages/product-client/src/components/workspace/chat/surface/ChatLaunchIntentPane.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/surface/ChatLaunchIntentPane.test.tsx
@@ -89,6 +89,8 @@ function intent(): ChatLaunchIntent {
     },
     materializedWorkspaceId: null,
     materializedSessionId: null,
+    attemptId: null,
+    targetWorkspaceId: null,
     createdAt: 1_700_000_000_000,
     sendAttemptedAt: null,
     failure: null,
--- a/apps/packages/product-client/src/hooks/chat/derived/use-chat-surface-state.ts
+++ b/apps/packages/product-client/src/hooks/chat/derived/use-chat-surface-state.ts
@@ -5,6 +5,7 @@ import {
   resolveChatSurfaceState,
   type ChatSurfaceState,
 } from "#product/lib/domain/chat/surface/chat-surface";
+import { resolveLaunchIntentScope } from "#product/lib/domain/chat/launch/launch-intent";
 import { shouldShowCloudWorkspaceStatusScreen } from "#product/lib/domain/workspaces/cloud/cloud-workspace-status";
 import { parseCloudWorkspaceSyntheticId } from "#product/lib/domain/workspaces/cloud/cloud-ids";
 import { useChatLaunchIntentStore } from "#product/stores/chat/chat-launch-intent-store";
@@ -22,6 +23,9 @@ export function useChatSurfaceState(shellRenderSurface?: WorkspaceRenderSurface 
   selectedWorkspaceId: string | null;
 } {
   const selectedWorkspaceId = useSessionSelectionStore((state) => state.selectedWorkspaceId);
+  const selectedLogicalWorkspaceId = useSessionSelectionStore(
+    (state) => state.selectedLogicalWorkspaceId,
+  );
   const pendingWorkspaceEntry = useSessionSelectionStore((state) => state.pendingWorkspaceEntry);
   const workspaceArrivalEvent = useSessionSelectionStore((state) => state.workspaceArrivalEvent);
   const activeLaunchIntent = useChatLaunchIntentStore((state) => state.activeIntent);
@@ -84,11 +88,14 @@ export function useChatSurfaceState(shellRenderSurface?: WorkspaceRenderSurface 
     selectedWorkspaceId,
     hasPendingWorkspaceEntry: pendingWorkspaceEntry !== null,
     activeLaunchIntentId: activeLaunchIntent?.id ?? null,
+    launchIntentScope: activeLaunchIntent ? resolveLaunchIntentScope(activeLaunchIntent) : null,
     launchIntentInFlight: activeLaunchIntent ? activeLaunchIntent.failure === null : false,
     launchIntentSessionId:
       activeLaunchIntent?.materializedSessionId
       ?? activeLaunchIntent?.clientSessionId
       ?? null,
+    shellLogicalWorkspaceId: selectedLogicalWorkspaceId,
+    shellWorkspaceId: selectedWorkspaceId,
     selectedLocalWorkspace,
     isArrivalWorkspace: workspaceArrivalEvent?.workspaceId === selectedWorkspaceId,
     shouldShowSelectedCloudWorkspaceStatus: selectedCloudWorkspace
@@ -105,10 +112,13 @@ export function useChatSurfaceState(shellRenderSurface?: WorkspaceRenderSurface 
     isRunning,
     streamConnectionState,
   })), [
+    activeLaunchIntent?.attemptId,
     activeLaunchIntent?.clientSessionId,
     activeLaunchIntent?.failure,
     activeLaunchIntent?.id,
     activeLaunchIntent?.materializedSessionId,
+    activeLaunchIntent?.materializedWorkspaceId,
+    activeLaunchIntent?.targetWorkspaceId,
     activeSessionId,
     hasContent,
     hasTranscriptEntry,
@@ -119,6 +129,7 @@ export function useChatSurfaceState(shellRenderSurface?: WorkspaceRenderSurface 
     selectedCloudRuntime.state?.preserveVisibleContent,
     selectedCloudWorkspace,
     selectedLocalWorkspace,
+    selectedLogicalWorkspaceId,
     selectedWorkspaceId,
     shellRenderScope,
     streamConnectionState,
--- a/apps/packages/product-client/src/hooks/home/workflows/home-next-launch-intent.ts
+++ b/apps/packages/product-client/src/hooks/home/workflows/home-next-launch-intent.ts
@@ -1,5 +1,6 @@
 import {
   resolveChatLaunchRetryMode,
+  resolveLaunchIntentPendingAttemptId,
   resolveLaunchIntentPendingWorkspaceId,
   type ChatLaunchRetryMode,
 } from "#product/lib/domain/chat/launch/launch-intent";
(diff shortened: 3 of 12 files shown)


pablonyx (commented):
## Review: PR0 launch-intent surface scoping (PRO-230)

Traced the wiring end-to-end (store → `use-chat-surface-state.ts` → `resolveChatSurfaceState` → `resolveLaunchIntentSurfaceOverride`/`launchIntentOwnsShell`) against `use-home-next-launch.ts` and `home-next-launch-intent.ts`. The failed-intent-hijack case this PR targets is fixed correctly and is well covered by the new `chat-surface.test.ts` cases. But there's a blocking regression in the success/in-flight path.

### 1. Blocking: every new-workspace Home launch loses its launch-intent pane for the entire in-flight window

`resolveLaunchIntentScope` (`launch-intent.ts:79-86`) derives scope from `intent.attemptId` / `intent.materializedWorkspaceId` / `intent.targetWorkspaceId`. For `local`-new, `worktree`, `cowork`, and `cloud` targets, `beginLaunchIntent` (`use-home-next-launch.ts:90-120`) sets `attemptId: null` and `targetWorkspaceId: null` (only the `local`-into-*existing*-workspace branch sets `targetWorkspaceId` up front). The comment at `use-home-next-launch.ts:112-114` says the attempt id "gets threaded in once known via `markHomeLaunchIntentMaterializedFromPendingWorkspace`" — but that function is called **only** from the `catch` block at `use-home-next-launch.ts:290`, i.e. only on failure. There is no call on the success path (verified by grepping every call site of `markHomeLaunchIntentMaterializedFromPendingWorkspace` and every `markMaterializedIfActive`/`markLaunchIntentMaterialized` call across `hooks/home`, `hooks/sessions`, `hooks/cloud`, `hooks/cowork` — none pass `attemptId`, and the ones that pass `workspaceId` only fire *after* the workspace/thread creation promise resolves).

Meanwhile `beginPendingWorkspace` runs synchronously (before the first `await`) in `createLocalWorkspaceAndEnterInternal`, the worktree entry action ("INSTANT SHELL" comment), and `createCoworkThreadWorkflow`, so `pendingWorkspaceEntry` (and `selectedLogicalWorkspaceId = "pending-workspace:<attemptId>"`) exists in the same React batch as `beginLaunchIntent`. So for the whole duration of the real async creation work (git/filesystem for local/worktree, a network round trip for cloud), you have:

- `activeLaunchIntent` = unscoped (`{pendingUiKey: null, workspaceId: null}`)
- shell = `pending-workspace:<attemptId>` (non-null)

`launchIntentOwnsShell` (`chat-surface.ts:79-98`) takes the unscoped branch and requires `shellLogicalWorkspaceId === null && shellWorkspaceId === null` — false, since the pending shell key is set. So `resolveLaunchIntentSurfaceOverride` (`chat-surface.ts:113-119`) returns `null`, and `resolveChatSurfaceState` falls through past the in-flight branch (`chat-surface.ts:155-170`, which is designed to keep showing the launch-intent pane while `launchIntentInFlight`) straight to the unconditional `if (input.hasPendingWorkspaceEntry) return { kind: "session-empty", ... }` at `chat-surface.ts:185-187`.

That's exactly the state the *unchanged* comment directly above it warns against: "rendering session-empty here instead double-mounts the creation receipt (canvas topSlot first, pending-prompt frontier a beat later) and the transcript jumps." Pre-PR, `resolveLaunchIntentSurfaceOverride` had no scope gate at all and always returned the launch-intent override in this state, so this is a genuine regression, not a pre-existing issue — confirmed by diffing against `origin/main`'s `chat-surface.ts`.

Net effect: every Home launch that creates a new workspace (the primary product flow) shows the wrong/empty surface and a visible transcript jump until the workspace materializes, then self-heals once `markLaunchIntentMaterialized` sets `materializedWorkspaceId` after the async work resolves.

Fix direction: thread `attemptId` onto the active intent as soon as `pendingWorkspaceEntry` exists on the success path too (e.g. call `markHomeLaunchIntentMaterializedFromPendingWorkspace` right after `beginPendingWorkspace`/`createThreadFromSelection` return, not only in the `catch`), or have `launchIntentOwnsShell`'s unscoped branch also own a pending shell while the intent has no attemptId/target yet and is still in flight.

Relevant lines: `apps/packages/product-client/src/hooks/home/workflows/home-next-launch-intent.ts:67-88`, `apps/packages/product-client/src/hooks/home/workflows/use-home-next-launch.ts:90-120,289-290`, `apps/packages/product-client/src/lib/domain/chat/surface/chat-surface.ts:79-98,100-136,155-187`, `apps/packages/product-client/src/hooks/chat/derived/use-chat-surface-state.ts:75-91`.

### 2. Test-quality gap (explains why #1 wasn't caught)

The new `chat-surface.test.ts` / `launch-intent.test.ts` cases are correct unit tests of the pure functions in isolation, but every scoped-intent case in them constructs `launchIntentScope` to already match the shell (e.g. `{ pendingUiKey: "pending-workspace:attempt-1", ... }` paired with `shellLogicalWorkspaceId: "pending-workspace:attempt-1"`). None of them reproduce the actual runtime state from #1 — an *unscoped* `launchIntentScope` (because the intent's `attemptId` was never threaded) paired with a non-null `shellLogicalWorkspaceId` (because `pendingWorkspaceEntry` already exists). There's also no test at all for the integration point: `apps/packages/product-client/src/hooks/home/workflows/home-next-launch-intent.ts` has no dedicated test file, and `use-home-next-launch.test.tsx` never asserts what `resolveLaunchIntentScope(activeIntent)` or the chat surface mode look like mid-flight. A negative control run against pre-PR code wouldn't apply here since the bug is new in this PR, but conversely none of the added tests would fail if the wiring gap in #1 were "fixed" by doing nothing — they don't exercise it.

### Verified non-issues (ruled out during review, noted for the record)

- **Cloud synthetic ids**: `result.workspaceId` in the cloud creation flow is built via `cloudWorkspaceSyntheticId(workspace.id)` (`use-create-cloud-workspace.ts:228`) before it's used for both `selectWorkspace`/`finalizeSelection` *and* `markLaunchIntentMaterialized`'s `workspaceId`, and `selectedWorkspaceId` for cloud workspaces is likewise the synthetic `cloud:<id>` form. So `materializedWorkspaceId` on the intent and `shellWorkspaceId` on the shell are in the same format — no mismatch there.
- **Deferred cloud launch runner** (`use-home-deferred-launch-runner.ts`): `launch-home-cloud-target.ts:109-113` calls `markLaunchIntentMaterialized(..., { workspaceId: result.workspaceId, ... })` unconditionally, before branching on `result.status`, so the intent is scoped to its real cloud workspace id *before* it's ever handed to `enqueueDeferredLaunch`. A later failure from the deferred runner (`failLaunchIntentIfActive`) therefore stays scoped to that workspace and won't hijack a different workspace's shell — this part of the ladder's stated goal works correctly.
- **Missed `ChatLaunchIntent` construction sites**: grepped `apps/desktop`, `apps/web`, and `apps/packages` for the type — the only production `.begin(...)` call site is `use-home-next-launch.ts:90`, which the PR updated correctly. `apps/desktop`/`apps/web` have no local copies (they consume `product-client`'s build). All three test-file `intent()` factories were updated too.
- **`session-transcript` fallthrough**: unaffected when the intent owns its shell (the ownership check is a no-op pass-through in that case); the "let launch-owned transcript content take over" and "no visible content yet" branches are otherwise untouched.

### Minor: likely-dead branch in `use-main-screen-state.ts`

`activeLaunchIntentIdForShell`'s new `else` branch (`use-main-screen-state.ts:124-133`, gated on `!selectedWorkspaceId && !pendingWorkspaceEntry`) only matters if a render can happen with an active *unscoped* launch intent but no `pendingWorkspaceEntry` yet. Given `beginPendingWorkspace` runs synchronously in the same call stack as `beginLaunchIntent` for every new-workspace target (confirmed for local/worktree/cowork), React 18 batches both updates into one render, so this branch's true case looks unreachable in practice (the one target that stays scope-mismatched, `local` into an existing workspace, is already scoped via `targetWorkspaceId` and correctly returns `false` here pre- and post-PR). Not a correctness bug, `hasLaunchIntentOnlyShell` stays `false` either way, but worth double-checking against a real repro (e.g. a slow/throttled workspace list load) before relying on it, since as far as I can tell it doesn't currently change any user-visible mount decision.

pablonyx:
Finding 1 (blocking) is fixed in 1d2584906: on the success path for each new-workspace target (cowork, local-new, worktree), `markHomeLaunchIntentMaterializedFromPendingWorkspace` is now called synchronously right after invoking the create call — captured as a promise first — instead of only from the catch block. `beginPendingWorkspace` runs in the synchronous prefix of `createCoworkThreadWorkflow`, `createLocalWorkspaceAndEnterInternal`, and `createWorktreeAndEnterInternal` (verified line-by-line, before their respective first `await`s), so the pending attempt id is scoped onto the active intent in the same synchronous stretch that sets the pending shell key — no window where the shell exists but the intent doesn't own it. Cloud and local-into-existing-workspace were already correctly scoped and are unchanged.

Finding 2 (test-quality gap) is addressed by two new regression tests:
- `chat-surface.test.ts`: pins the begin-time-scoped state (`resolveLaunchIntentSurfaceOverride` returns the launch-intent override when the intent's `pendingUiKey` matches the shell) against the previously-broken state (unscoped intent + populated pending shell key → override is `null`), reproducing the exact runtime state from finding 1.
- `use-home-next-launch.test.tsx`: a worktree launch with a controllable (not-yet-resolved) create promise, asserting `activeIntent.attemptId` is already set before the promise resolves — this would fail if scoping regressed to catch-block-only.

On the minor `use-main-screen-state.ts` else-branch question: kept as-is. The unscoped+nothing-selected window still exists momentarily before `beginPendingWorkspace` runs within the same task for new-workspace launches, and for any future intent-only flows that don't create a pending entry at all, so the branch stays as the safe gate rather than being removed.

Full product-client suite (856 files / 5986 tests) and both tsconfig typechecks (domain + main) are green on this commit.

PR #1868, opened 2026-08-14
fix(product): pending-workspace registry so launches survive switching away (PRO-230)

pablonyx:
## Problem

`pendingWorkspaceEntry` was a single nullable slot on `session-selection-store`, and six call sites across the launch pipeline reused `pendingWorkspaceEntry?.attemptId === attemptId` as the pipeline's liveness token. Selection and launch-pipeline liveness were the same field.

`activateWorkspace` (called whenever the user switches to a different workspace) defaulted to nulling that slot unless the caller explicitly passed `clearPending: false`. So switching away from a workspace that was still being created mid-launch nulled the entry, every liveness check downstream started reading `null`, and the launch pipeline stopped: the workspace never finished materializing, the projected session never remapped, and the first queued prompt never sent. The backend workspace was orphaned — created, then abandoned by the client mid-flight — with no error surfaced to the user, because nothing failed; the client just stopped looking.

PRO-230: https://linear.app/proliferate-team/issue/PRO-230/stitching-workspace-while-a-new-workspace-destroys-the-new-workspace

## The model

`pendingWorkspaceEntry` becomes `pendingWorkspaces`, a registry keyed by attempt id (`pending-entry-registry.ts`). The overloaded liveness/attendance check splits into two independent predicates (`pending-workspace-attempt-access.ts`):

- **`isAttemptLive(attemptId)`** — gates pipeline work: patching the entry, remapping and materializing the projected session, clearing the entry on completion/failure. Runs regardless of what's selected.
- **`isAttemptAttended(attemptId)`** — gates presentation only: force-selecting the real workspace, the arrival event, activating the created session, composer focus. At most one attempt is attended at a time.

Selection is a camera again: it decides what's shown, never whether a launch completes. `activateWorkspace`/`activateHotWorkspace` no longer clear pending state as a side effect of switching — the shell-selection ledger points at one attempt at a time, but the attempts themselves live in their own registry and finish independently. Back/dismiss is the one explicit way to end an attempt early.

This is invariants 9 and 10 in the updated spec (`specs/codebase/systems/product/workspaces/pending-shell.md`):
> 9. Switching workspaces never ends a launch... A launch the user switched away from still materializes its workspace, sends its first prompt, and clears its own entry.
> 10. Liveness gates the pipeline; attendance gates presentation... Never use one check for both: that is the bug this split exists to prevent (PRO-230).

## Key changes

- **Registry domain modules**: `lib/domain/workspaces/creation/pending-entry-registry.ts` (attempts keyed by attempt id, upsert/remove/lookup) and `lib/domain/workspaces/creation/pending-attention.ts` (whether an attempt is the one currently attended), plus the imperative access layer `hooks/workspaces/workflows/pending-workspace-attempt-access.ts` (`isAttemptLive` / `isAttemptAttended` / `patchAttempt`).
- **`clearPending` option deleted.** The old boolean escape hatch on `selectWorkspace`/`activateWorkspace` is gone — liveness no longer needs a caller to opt out of being cleared, because switching doesn't clear anything by default anymore.
- **`workspace-entry-finalization.ts` now returns `{ committed, selected }`** instead of a single boolean, so callers can tell "the workspace/session pipeline completed" apart from "this attempt was the one on screen when it did."
- **`activateOnCreate` + `targetWorkspaceUiKey` on session creation** (`use-session-creation-actions.ts`, `use-pending-workspace-session-materialization.ts`): an unattended attempt still creates and materializes its session, it just doesn't activate it or steal the current selection — it's routed to its own workspace record instead.
- **~20 call sites migrated off the single-slot field** onto `useAttendedPendingWorkspaceEntry` / the registry accessors (sidebar, header tabs, chat surface, model selector, cloud workspace polling, cowork thread creation, transcript pending rows, etc.).
- **`specs/codebase/systems/product/workspaces/pending-shell.md`** updated: invariants 9/10 added, ownership table and vocabulary section extended for the registry, live/attended attempt definitions.

## Deliberate deviations from the frozen spec

- **`preservePending` kept**, not folded into the liveness split. It's still needed for the projected-session handoff inside `workspace-entry-finalization.ts`'s `selectWorkspace` call — selecting the now-real workspace mid-finalization must not wipe the registry entry the same finalize call is still reading from.
- **Dismissal is a quiet stop, not a thrown error.** If the user dismisses a pending workspace/worktree while it's still creating, `use-home-next-launch.ts` returns `false` without an error toast (nothing failed — the user just stopped watching) rather than routing it through the failure path.
- **`attemptId` is pre-minted in Home** (`createPendingWorkspaceAttemptId()` in `use-home-next-launch.ts`) for the local-new and worktree targets, and threaded explicitly into the create call and the launch-intent scoping call, rather than only discovered after the fact by reading whatever's currently selected. Cowork and cloud don't pre-mint (they fall back to resolving the currently-attended entry) since they don't need cross-switch identity the same way today.
- **`clearSelection` still wipes the entire registry.** Full sign-out/selection-reset still nulls every live attempt, not just the presentation layer — flagged as a follow-up for PR2, since a full registry wipe on `clearSelection` re-introduces the same "switching kills an in-flight launch" failure mode for that one call site.

## Tests

- 15 test files touched, +654/-71 lines: new coverage for the registry (`pending-entry-registry.test.ts`), attention resolution (`pending-attention.test.ts`), the imperative access layer (`pending-workspace-attempt-access.test.ts`), plus updated coverage across entry actions/flow, session materialization, cloud polling, cowork thread creation, and the selection store itself.
- **Switch-away acceptance test** (`use-workspace-entry-actions.test.tsx`, "completes a worktree launch after the user switches away mid-flight"): starts a worktree creation, waits for the pending entry to exist, calls `activateWorkspace` to a *different* workspace mid-flight, resolves the backend create, and asserts the launch still completes — `materializePendingWorkspaceSessions` is called, the entry clears — **without** pulling the user back to the workspace they left (selection stays on `workspace-other`, no arrival event fires there).
- **Negative-control basis**: pre-PR1, `activateWorkspace` nulled `pendingWorkspaceEntry` by default unless the caller passed `clearPending: false` (confirmed by diffing the old `session-selection-store.ts`). The switch-away test calls `activateWorkspace` the plain way, with no such override — exactly the call shape that used to null the single-slot entry and strand the pipeline reading `null` afterward. The registry split is what lets that same call leave the in-flight attempt untouched.
- After the Step-1/PR0 rebase: full `product-client` suite is 859 test files / 6006 tests, all green (up from 856/5986 pre-restack); both tsconfig typechecks (`tsconfig.domain.json`, `tsconfig.json`) clean.

## Stack

Slice 2 of the PRO-230 ladder, stacked on #1867 (launch-intent surface scoping).


pablonyx (commented):
## Review — PR1 delta only (`fix/pro-230-intent-surface-scope` → `fix/pro-230-pending-registry` @ 35880d97)

The live/attended split is the right decomposition and the core of it is correct. I traced every replaced `attemptId === pendingWorkspaceEntry?.attemptId` guard, both timelines (switch-away and attended), the prompt-routing rewrite, and the registry's mutation semantics. The registry is properly immutable, returns stable empty refs, and every setter is a single `set()`; `useAttendedPendingWorkspaceEntry` is `useShallow`-guarded over three scalars, so no render loops. Guard ordering in `finalizeSelection` is correct — the liveness check precedes `setPendingWorkspaceEntry`, so a dismissed attempt cannot be resurrected by the upsert. Prompt routing for local/worktree is sound: `beginPendingWorkspace` runs in the synchronous prefix, so `resolveProjectedPendingWorkspaceSession(attemptId)` always resolves, and the directory fallback via `sessionIdsByWorkspaceId[pendingWorkspaceUiKey]` is real because `putSessionRecord` populates it. The intent dispatcher is a selection-independent singleton keyed by `clientSessionId`, so a backgrounded attempt's queued prompt does dispatch once its session materializes.

Two findings below are, in my read, blocking for the invariants this PR writes into the spec. Neither is a design objection — both are gaps where the new invariant is asserted in `pending-shell.md` but not actually delivered by the code on this branch.

### MEDIUM/HIGH — `clearSelection` still wipes the whole registry, and it is reachable from ordinary per-workspace actions

`session-selection-store.ts:187-189` — `clearSelection` sets `pendingWorkspaces: EMPTY_PENDING_WORKSPACE_REGISTRY`. `activateWorkspace`, `activateHotWorkspace`, and `deselectWorkspacePreservingSessions` correctly leave the registry alone, so the headline case is fixed. But `clearSelection` is not confined to app-level reset:

- `clear-runtime-state.ts:58-59` calls `deps.clearSelection()` whenever `options.clearSelection && selectedWorkspaceId === workspaceId`.
- `use-workspace-retire-actions.ts:52` passes `{ clearSelection: targetIsSelected }`.
- `use-cloud-workspace-actions.ts:89` passes `{ clearSelection: true }` on cloud workspace delete.
- `use-pending-workspace-entry-actions.ts:164` (home-origin Back) calls `clearWorkspaceRuntimeState(selectedWorkspaceId, { clearSelection: true })` *after* it has already cleared its own attempt by id.

Failure scenario: start a worktree launch for repo A, switch to workspace B while it provisions, then mark B done (or delete B if it is a cloud workspace). `clearSelection` fires and the still-in-flight attempt A is silently removed from the registry. `isAttemptLive(attemptA)` now returns false, so `finalizeSelection` returns `DISMISSED_RESULT`, the workspace is never materialized into the shell, the projected session is never remapped, and the queued first prompt is stranded in the outbox forever. The user gets no error — `createWorktreeAndEnterWithResult` returns `null` and `use-home-next-launch.ts` clears the launch intent quietly. Same shape via home-origin Back: dismissing attempt A also nukes unrelated attempt B.

This directly contradicts the §3 "Live attempt" text added in this PR ("Only an explicit dismissal (back, or an app-level selection reset) ends it") and invariant 9. The fix is presumably to drop the registry wipe from `clearSelection` entirely and let dismissal go through `clearPendingWorkspaceEntry(attemptId)` — but that is a call for the author, and it may belong to PR2's scope. Either way it should not ship asserting invariant 9 while this path exists.

### MEDIUM — unattended cloud `awaiting-cloud-ready` launches never complete in the background

Both drivers that finish a cloud launch are scoped to the current selection:

- `use-cloud-workspace-polling.ts:34-83` — the hook reads `useAttendedPendingWorkspaceEntry()` (line 36), `selectedPendingCloudWorkspaceIsAwaiting` requires `pendingWorkspaceEntry?.workspaceId === selectedWorkspaceId` (line 50), and the whole effect early-returns on `!selectedWorkspaceId` (line 77). It also only polls the cloud workspace parsed out of `selectedWorkspaceId` (line 45).
- `use-selected-cloud-runtime-rehydration.ts` resolves its entry from `useSelectedCloudRuntimeState`, i.e. also selection-derived.
- `use-create-cloud-workspace.ts:264-271` correctly gates only `selectWorkspace` on `isAttemptAttended`, then returns `{ status: "awaiting-ready" }` and hands off to those two hooks.

Failure scenario: launch a cloud workspace from home with a first prompt, then switch to another workspace while it is still provisioning (the common case — cloud provisioning is exactly when you'd go do something else). Polling never starts for that workspace. The entry stays in the registry at stage `awaiting-cloud-ready` indefinitely; the projected session is never materialized, so the queued prompt never dispatches. It only unsticks if the user manually re-selects that workspace. Nothing surfaces the stall.

Invariant 9 as written ("A launch the user switched away from still materializes its workspace, sends its first prompt, and clears its own entry") is therefore true for local/worktree and false for cloud. The §13 test bullet added in this PR is likewise only covered for the worktree path (`use-workspace-entry-actions.test.tsx` "completes a worktree launch after the user switches away mid-flight"). If the cloud background-completion driver is intentionally deferred to a later slice, the spec text should say so rather than claim it.

### LOW/MEDIUM — attendance is read twice and the two reads can disagree

`use-workspace-entry-flow.ts:164` captures `const attended = isAttemptAttended(entry.attemptId)` *before* the `await selectWorkspace(...)` at line 183 — with a correct comment explaining why. But `use-pending-workspace-session-materialization.ts:134` recomputes `isAttemptAttended(entry.attemptId)` independently, after that await has landed, to decide `activateOnCreate` (line 141) and `targetWorkspaceUiKey` (line 145). `workspace-entry-finalization.ts:85` has the same shape.

Failure scenario: the user switches away during the `await selectWorkspace` (or during the `await` inside materialization). Finalization believes `attended === true` and fires `setWorkspaceArrivalEvent`, while materialization recomputes `attended === false` and creates the session with `activateOnCreate: false` — so an arrival panel appears for a workspace whose session was never activated. The mirror case (attended at materialization, unattended at capture) skips `selectWorkspace` but then activates the new session, yanking the user's active session out from under them. Both are narrow races, but the whole point of the "read attendance once" comment at line 163 is defeated by the second read. Thread the captured `attended` value through `materializePendingWorkspaceSessions` instead of recomputing it.

### LOW/MEDIUM — the cowork branch was not migrated with the rest

`use-home-next-launch.ts:157` still does `throw new Error("Cowork thread creation was interrupted.")` when the create returns null, while the local and worktree branches were changed to clear the launch intent and `return false` quietly. The cowork branch also mints no `attemptId`, so it gets neither the attemptId-keyed prompt routing nor attempt-scoped intent handling that the other two branches gained in this PR.

Failure scenario: start a cowork thread, switch away or hit Back mid-flight — the user gets a loud failure toast for a launch that was either dismissed on purpose or completed fine in the background, whereas the identical worktree action is now silent. Inconsistent enough that it will read as a bug.

### LOW — unattended cowork threads are marked as viewed

`create-cowork-thread.ts:232-242` — `activateWorkspace` is correctly gated on `isAttemptAttended`, but `rememberLastViewedSession` (239), `markWorkspaceViewed` (241), and `markWorkspaceBootstrappedInSession` (242) run unconditionally. `trackWorkspaceInteraction` (240) is deliberately ungated per §8 step 6, which is fine — the other three are not interaction stamps, they are "the user has seen this" state. An unattended thread ends up marked viewed and bootstrapped without ever having been on screen, which suppresses whatever unread/first-visit affordances key off them.

### LOW — unattended shell-tab intent is written under the raw materialized workspace id

`use-pending-workspace-session-materialization.ts:145` passes `targetWorkspaceUiKey: attended ? null : workspaceId`, which flows to `shellWorkspaceId` in `use-session-creation-actions.ts`. Worth noting this is largely redundant for shell-state-key purposes: `resolveWorkspaceShellStateKey` already returns `workspaceId` when `selectedWorkspaceId !== workspaceId`, so the unattended case resolved correctly before. The load-bearing part of the change is `recoveryWorkspaceUiKey`, which genuinely needed it. The residual concern is cloud: the raw `workspaceId` here is the materialized id, while later reads for a cloud workspace key off the logical id, so an unattended cloud thread's shell tab intent can land under a key nothing reads back. Low impact given finding 2 means unattended cloud launches don't get this far today.

### NIT — PR2 scaffolding landed unused

`pending-entry-registry.ts:42` `patchPendingWorkspaceEntry`, and `use-pending-workspace-entries.ts:33,38` `usePendingWorkspaceEntries` / `usePendingWorkspaceEntry` are exported and tested but have no production caller on this branch. Fine as deliberate PR2 groundwork — flagging only because there is no dead-export ratchet in the repo to catch them if PR2 slips.

### Verified correct, no action

Selection-independent intent dispatch; `use-persisted-logical-workspace-selection`; `use-workspace-navigation-workflow`; `use-chat-availability-state`; `use-workspace-bootstrap-actions`; `preservePending` retention semantics; registry immutability, stable empty refs, and single-`set()` per setter; the guard ordering that stops a dismissed attempt being resurrected by `setPendingWorkspaceEntry`; `cleanupSessionCreationFailure`'s `previousActiveSessionId` re-activation (guarded by `activeSessionIdBeforeRemoval === currentOwnedSessionId`, so an unattended failure cannot yank the user's session). The `finalizeSelection` unattended test and the mid-flight switch-away acceptance test both carry real negative controls (`selectWorkspace` not called, arrival event null, `activeSessionId` unchanged), which is the right shape.

Spec edits to `pending-shell.md` are accurate to the code except where noted in findings 1 and 2 (§3 "Live attempt", invariant 9, §13 test bullet).

pablonyx:
All six findings addressed in 9d0dc6390.

**1 — `clearSelection` wipes the registry.** Fixed. `clearSelection` takes `ClearSelectionOptions { preservePendingWorkspaces }`; `clear-runtime-state.ts` passes `true`, so every per-workspace caller (retire/mark-done, cloud delete, home-origin Back) deselects without touching other attempts. No options means the full wipe, so desktop sign-out is unchanged. New `clear-runtime-state.test.ts` covers retire-B-preserves-attempt-A plus the app-level-reset contrast case; negative-controlled (reverting the option fails the first, passes the second).

**2 — unattended cloud never completes in the background.** Real, and deferred by plan rather than fixed here: the background driver is the deferred-launch/polling rework in the next slice. The spec no longer claims otherwise. Invariant 9 gains a scope note naming which sources complete unattended today and stating that a cloud attempt still provisioning parks at `awaiting-cloud-ready` until re-attended; §13 records that as stalled-not-failed (entry and queued prompt retained); the §14 test bullet now says the coverage is local/worktree.

**3 — attendance read twice across the await.** Fixed. `materializePendingWorkspaceSessions` takes `{ attended }`; both `use-workspace-entry-flow.ts` and `workspace-entry-finalization.ts` pass the value captured before `selectWorkspace`. Callers that never awaited a selection (cloud polling, rehydration, retry) still read it themselves via the `??` default. The three existing assertions now pin the threaded value (`{ attended: true }` / `{ attended: false }`).

**4 — cowork branch not migrated.** Fixed both halves. A null result now clears the launch intent and returns `false` with no toast and no throw, matching local/worktree. The attemptId thread-through turned out small (optional `attemptId` on the workflow input, defaulted to the minted one; forwarded through `createThreadWithResolvedConfig`/`createThreadFromSelection`), so cowork is pre-minted like the other branches and gets attempt-scoped `markHomeLaunchIntentMaterializedFromPendingWorkspace` and `promptProjectedPendingWorkspaceSession`. No cowork gap left. Covered in `use-home-next-launch.test.tsx` (quiet stop, no error toast, intent cleared, attemptId passed), negative-controlled against the old throw.

**5 — unattended cowork marked viewed.** Fixed. `rememberLastViewedSession`, `markWorkspaceViewed`, and `markWorkspaceBootstrappedInSession` are gated on the same `attended` flag as `activateWorkspace`; `trackWorkspaceInteraction` stays unconditional per §8 step 6. Two new tests in `create-cowork-thread.test.ts` (unattended: none of the three, but recency still tracked and the entry still cleared; attended: all three), negative-controlled.

**6 — shell-tab intent under the raw materialized id.** Fixed by dropping the forced key rather than translating it. Your analysis holds: for the unattended case `resolveWorkspaceShellStateKey` already returns the same `workspaceId`, so forcing it bought nothing, and forcing it is what could desync from the logical id later. `writeOwnedShellIntent` now passes `null` and lets the writer resolve, which yields that same id while unselected and the logical id once selected. `targetWorkspaceUiKey` is kept solely as `recoveryWorkspaceUiKey`, where it is genuinely load-bearing, and the comment says so.

**NIT — unused scaffolding.** `patchPendingWorkspaceEntry`, `usePendingWorkspaceEntries`, and `usePendingWorkspaceEntry` are consumed by the follow-up slice (concurrent launches), which is about to be posted stacked on this branch.

Also swept the PR1 test residue: `session-records.test.ts` and `session-store-isolation.test.ts` were still seeding the renamed `pendingWorkspaceEntry` key; both now reset `pendingWorkspaces` to the empty registry.

Verification: full product-client suite 860 files / 6012 tests green, `tsconfig.domain.json` and `tsconfig.json` both clean.

pablonyx:
CI repairs across the stack (repo shape + login first-load budget).

**Repo shape — now green on all three.** Two gates were failing, not one:

- `check_max_lines.py`: `use-home-next-launch.test.tsx` (790) and `pending-sidebar-projection.test.ts` (664) split at the seam this stack created — single-launch vs concurrent-launch behaviour — into `*.concurrency.test.tsx` / `*.concurrent.test.ts`. No test deleted or weakened (14 and 15 assertions preserved).
- `report_frontend_structure.py --strict`: four source files crossed their size ratchet. Each was fixed by moving the new logic into the module that already owns that concern, so **no allowlist entry was added or raised**:
  - `use-session-creation-actions.ts` 459 → 443 (recovery ui-key resolution → `workspace-ui-key.ts`; queued-prompt failure announcement → `session-creation-failure-cleanup.ts`)
  - `use-workspace-bootstrap-actions.ts` 431 → 424
  - `use-home-next-launch.ts` 443 → 394 (`resolveHomeNextLaunchRefusal` + `beginHomeNextLaunch` → `home-next-launch-intent.ts`)
  - `run-workspace-selection.ts` 406 → 389 (`selectPendingWorkspaceAttempt` → `pending-workspace-attempt-access.ts`)

**Login first-load budget.** Four eager import edges into the login entry chunk were broken, all by the same rule: launch-registry work only a signed-in viewer can reach should not be in the signed-out chunk.

- `use-home-deferred-launch-runner` + `use-cloud-workspace-polling` → `AuthenticatedLaunchLifecycles`, lazy behind the authenticated gate (same treatment `AuthRestartOfferRoot` gets).
- `launchIntentOwnsShell` moved off `chat-surface.ts`, then the scope helpers moved again into `launch-intent-scope.ts` so `launch-intent.ts` (retry modes, view model, catalog snapshot) leaves the chunk too. The registry is reached eagerly via `ProductLifecycleRoot → use-app-shortcuts → workspace selection → session creation → chat-launch-intent-store`.
- `cloud-workspace-poll-plan` moved off `use-pending-workspace-entries.ts`.

CI gzip-9 JS, cap 490000:

| | before | after |
|---|---|---|
| #1868 | 489199 pass | **489225 pass** |
| #1870 | 490729 fail | **490165** (−564) |
| #1873 | 490859 fail | **490283** (−576) |

**#1870 and #1873 are still 165 B and 283 B over, and I stopped there rather than force it.** Every remaining edge roots in eager infrastructure this stack does not own — `use-app-command-actions` (a context provider) and `use-app-shortcuts`, which between them pull the sidebar projection, cloud workspace creation and pending materialization graphs into the login chunk for a signed-out viewer. Gating those is an architectural change to `ProductLifecycleRoot`, not a PRO-230 concern.

Two things worth knowing before deciding:

1. **The 490000 cap these branches measure against is stale.** main raised it to 500500 in #1875. Merged with main, the whole stack measured **497032/500500 (3468 B headroom)** — and it has shed a further ~3.6 kB since. The rebase resolves the budget on its own.
2. **The stack needs that rebase anyway.** #1837 (archiving R7) reworked `buildSidebarWorkspaceItems` — it dropped the `archivedSet` argument in favour of reading `cloudWorkspace.productLifecycle` — which conflicts with this stack's `pendingItems`/`pendingOwnedWorkspaceIds` change in `sidebar-groups.ts` and `sidebar-workspace-items.ts`.

Also: the budget metric is platform-dependent. The same CSS file (identical content hash) measures 65745 B locally on macOS and 65903 B on CI, so local runs under-report by ~0.36% (~1.7 kB at this size). Worth a note in the runbook — I calibrated against CI rather than trusting the local number.

PR #1870, opened 2026-08-14
fix(product): allow concurrent workspace launches (PRO-230)

pablonyx:
## Problem

A second Enter on Home clobbered the first launch. Three singletons stood between the user and a second concurrent launch, and each one destroyed a different piece of the first one:

- the launch-intent store held a single `activeIntent`, so a second launch overwrote the first launch's intent
- the sidebar rendered exactly one pending row, so the first attempt lost its only visible representation
- `use-home-next-launch` held an `inFlightRef` that refused the second submit outright

The superseded launch was orphaned silently: still running, no row, no intent, no way back to it, and its queued first prompt stranded in the outbox. PR1 made the pending-workspace side a registry; this slice removes the three singletons on top of it.

## Change

- **Launch-intent registry.** Intents are keyed by id with pure registry helpers, replacing the single `activeIntent` slot.
- **Per-shell intent resolution.** `use-shell-launch-intent` resolves each shell's own intent by scope instead of every shell asking for "the" active one.
- **Plural sidebar rows.** The sidebar projects every live attempt, with duplicate suppression keyed off `entry.workspaceId` and applied across all repo groups (previously one row, suppression keyed off the selected workspace).
- **Pending-row selection for non-attended entries.** `enterPendingWorkspaceAttemptShell` lets the user click back into an attempt they are not currently attending.
- **Unattended-failure toast.** A launch that fails while the user is elsewhere raises one toast with a Show action, instead of rendering a creation receipt over an unrelated shell.
- **Per-entry dismiss and retry.** Both act on a single attempt and clear that attempt's linked intent, leaving the other launches alone.
- **Concurrency cap and debounce.** `MAX_CONCURRENT_PENDING_LAUNCHES = 5` plus a 1s same-prompt debounce replace the `inFlightRef` lock, so a second distinct launch proceeds and only an accidental double-submit is swallowed.
- **Composer unblocking.** `submitInFlightRef` / `canSubmit` no longer gate on a global in-flight launch; the draft is cleared under `flushSync` so the composer is immediately ready for the next prompt.
- **Failed-entry GC.** Failed attempts older than 24h are swept on app start.
- **Spec.** `pending-shell.md` gains Invariant 11 (two launches run at once, all the way down) and rewritten sections 9, 13, and 14.

## Tests

865 test files / 6049 tests green; `tsconfig.domain.json` and `tsconfig.json` both clean.

New coverage lands with the units it belongs to: `launch-intent-registry`, `launch-concurrency`, `pending-workspace-failure-notice`, `use-stale-failed-pending-workspace-gc`, `use-pending-workspace-entry-actions`, plus additions to the sidebar projection, chat-launch-intent store, selection store, main-screen state, and Home launch suites.

Three negative controls from the implementation run:

1. Two concurrent launches keep two distinct sidebar rows and two distinct intents; collapsing either back to a singleton fails the assertion on the second attempt's row and intent.
2. A per-attempt dismiss removes exactly that attempt and its linked intent; clearing the registry wholesale fails on the surviving attempt.
3. The debounce swallows a repeated identical prompt but not a second distinct one; widening it to any second submit fails the distinct-launch case.

## Notes

Slice 3 of the PRO-230 ladder, stacked on #1868. Linear issue: https://linear.app/proliferate-team/issue/PRO-230/stitching-workspace-while-a-new-workspace-destroys-the-new-workspace


pablonyx:
## Review — PR2 (concurrent launches), read against `9d0dc6390..47aca77b5`

The registry conversion itself is clean and I could not break it: every former `…IfActive` call site now passes its own `launchIntentId` (I checked all 14, plus a tree-wide grep — the only `activeIntent` reads left belong to the unrelated cloud-repository/availability intent stores), the store keeps one `set()` per mutator with identity-returning no-ops for a missing id, the sidebar suppression set is genuinely registry-wide and applied in every group, and per-entry dismissal correctly ends exactly one attempt plus its linked intent. What the slice has not finished is the *cloud* branch and the *aftermath* of a launch: the cloud branch is the one launch path that never learns its own attempt id, so its failure handling still resolves "the attended attempt" and can bind one launch's intent to another launch's attempt; and now that `begin` no longer evicts the previous intent and every attempt gets a permanent row, the two retry paths and the GC leave debris that used to be invisible. I'd want the first one fixed before merge; the rest are landable-with-follow-ups if you'd rather keep the ladder moving.

---

### Blocking

**1. The cloud branch never threads its attempt id, so a cloud failure scopes its intent to whatever attempt is *attended* — which under concurrency is another launch.**
`use-home-next-launch.ts:157` mints `launchAttemptId` and the cowork/local/worktree branches assign it (`:161`, `:224`, `:298`); the cloud branch (`:356-375`) never does, so the catch at `:377` calls `markHomeLaunchIntentMaterializedFromPendingWorkspace(launchIntentId, null)`. With a null attempt id, `resolveLaunchPendingWorkspaceEntry` (`home-next-launch-intent.ts:129-140`) falls back to `resolveAttendedPendingWorkspaceEntry`, and `pendingWorkspaceMatchesLaunchTarget` accepts *any* `cloud-created` entry for a cloud target (`launch-intent.ts:118-120`).

Failure scenario: launch A (cloud, repo R1) is running; the user submits launch B (cloud, repo R2), whose `beginPendingWorkspace` moves the camera to B's pending shell; A's create then rejects (402 billing gate, branch conflict, network). A's catch scopes intent A to **B's** `attemptId` and **B's** `workspaceId`. Three consequences, all verifiable from the code:
- `resolveLaunchIntentForShell` now has two intents owning B's shell; the `createdAt` tie-break hides A only while B lives. The moment B succeeds and calls `clearLaunchIntent(B)` (`:292`/`:352`/`launch-home-cloud-target.ts:137`), A resurfaces and B's shell renders "Couldn't start work" for a launch that has nothing to do with it — exactly the hijack Invariant 11 forbids.
- `launchIntentForAttempt` (`launch-intent-registry.ts:107-125`) returns the *first* intent matching an attempt id, i.e. A. So Back on entry B (`use-pending-workspace-entry-actions.ts:165-173`) clears **A's** intent and leaves B's live.
- `homeLaunchFailureRetryMode(intentId, null)` (`home-next-launch-intent.ts:117-122`) reads B's workspace and can report `manual_after_workspace` for a launch that never created anything, and `dismiss` on A's pane (`use-chat-launch-intent-actions.ts:58-69`) would open B's workspace.

Fix shape: pre-mint the attempt id in `useHomeNextLaunch` and pass it into `createCloudWorkspaceAndEnterWithResult` the way the other three branches do (the cloud flow already accepts nothing but mints its own at `use-create-cloud-workspace.ts:148`), or surface `result.attemptId` out of `launchHomeCloudTarget` before the throw. While you're there: `launch-home-cloud-target.ts:88-93` is also the only `promptProjectedPendingWorkspaceSession` caller that omits `attemptId`, so it lands in the attendance fallback at `home-next-projected-session.ts:35-47`. Today that resolves correctly only because `beginPendingWorkspace` produces the projected session synchronously before the call; if the first synchronous resolve ever misses, the subscriber latches onto the next selection change — which is now plausibly another launch's shell, and the function's own docstring says that misroutes the first prompt.

---

### Should fix

**2. Retrying a failed creation mints a new attempt and never clears the old one — with plural rows that is now a permanent orphan.**
`use-pending-workspace-entry-actions.ts:60-77`: the `local`/`worktree`/`cloud` retries call `createLocalWorkspaceAndEnter(sourceRoot)` / `createWorktreeAndEnter(input, {latencyFlowId})` / `retryCloudWorkspaceAndEnter(input)` with no `attemptId`, so each mints a fresh one (`use-workspace-entry-actions.ts:121`, `use-create-cloud-workspace.ts:148`) while `entry.attemptId` stays in the registry, stuck at `stage: "failed"`. Under PR1 only the attended entry had a row, so the corpse was invisible; PR2 renders every entry, so the user now sees "Couldn't create workspace" sitting in the sidebar next to the retry that succeeded, and nothing will ever remove it (see #3). Retry should clear (or reuse) the attempt it is retrying.

**3. `useStaleFailedPendingWorkspaceGc` cannot collect anything.**
`use-stale-failed-pending-workspace-gc.ts:12-28` runs once on mount with `[]` deps. `useSessionSelectionStore` is a plain `create(...)` with `pendingWorkspaces: EMPTY_PENDING_WORKSPACE_REGISTRY` (`session-selection-store.ts:71-73`) — no `persist` middleware anywhere in the tree — so the registry is empty at every app start, and `MainScreen` stays mounted for the entire authenticated session (`AuthenticatedAppHost.tsx:54-69` keeps the workspace host mounted-but-hidden across all routes). The sweep therefore fires exactly once, against an empty registry, and never again. The hook's own docstring ("a persisted registry accumulates a row per failure forever") and the spec bullet "failed attempts are swept on app start once they are older than a day" (`pending-shell.md`) both describe behavior that does not exist. Either drop the hook and the spec bullet, or make it real (interval sweep, or sweep on entry insert) — the choice matters because #2 and #6 depend on *something* eventually reaping failed entries. Also note it clears the entry but not the linked launch intent, unlike `handleBack`.

**4. Nothing evicts a failed launch intent any more, and `retry` adds a second one.**
`begin` used to replace the singleton, which was the de-facto GC for abandoned intents. Now `upsertLaunchIntent` appends, and the only removals are the explicit `clear` calls. `use-chat-launch-intent-actions.ts:26-32` (`retry`) calls `launch(activeIntent.retryInput)` without clearing `activeIntent` first, so the failed intent stays keyed to the same scope as its replacement. Concrete case that needs no cloud: a Home launch fails before minting an attempt (intent unscoped, `pendingUiKey === null && workspaceId === null`); the user hits Retry; the new intent wins the `createdAt` tie-break in `resolveLaunchIntentForShell`; it succeeds and clears; the stale unscoped intent now matches the empty shell again, `shouldMountWorkspaceShell` returns true off `activeLaunchIntentId` (`use-main-screen-state.ts:127-133`), and the user gets the old "Couldn't start work" pane mounted over Home right after a successful launch. `retry` should `clearLaunchIntent(activeIntent.id)` before launching.

**5. An unattended failure is announced twice.**
`failPendingEntry`/`failPendingWorkspaceEntry` now call `notifyUnattendedPendingWorkspaceFailure` ("Workspace creation failed", id-keyed), and then the creator rethrows (`use-workspace-entry-actions.ts:165-167`) into `use-home-next-launch.ts:376-391`, which raises a second, un-keyed `showErrorToast("Work not started", …)`. So the unattended local/worktree failure the notice exists for produces two stacked error toasts, and the second one tells the user "Your prompt is back in the composer" while they are looking at another workspace. The spec's new bullet — "an attended failure keeps the inline presentation only, so nothing is announced twice" — is accurate for the attended case only; unattended is where the doubling happens. Suggest suppressing the launch-level toast when the attempt is unattended (the notice already owns that case).

**6. The failure toast's "Show" does not navigate, so on any non-home route it does nothing visible.**
`pending-workspace-failure-notice.ts:38-43` → `enterPendingWorkspaceAttemptShell` (`pending-workspace-attempt-access.ts:52-68`) only mutates the selection store. Every other entry point into a pending shell goes through `selectWorkspaceFromSurface`, which calls `navigateToWorkspaceShell()` first (`use-workspace-navigation-workflow.ts:37-41, 78`). Since the workspace host is rendered with `className="hidden"` on `/workflows`, `/workspaces` and `pointer-events-none` under `/settings` (`AuthenticatedAppHost.tsx:54-69`), a user who was on `/workflows` when the toast fired clicks "Show" and stays exactly where they are. Route to `/` (or reuse `selectWorkspaceFromSurface`) inside `onNavigate`. Minor companion: if the attempt was dismissed meanwhile, `enterPendingWorkspaceAttemptShell` returns `false` and the click is silently inert.

**7. The debounce fingerprint ignores the launch target, and its rejection path restores the draft.**
`launch-concurrency.ts:55-60` fingerprints `{prompt, at}` only, so "run the tests" submitted to repo A and then to repo B inside one second is refused — contradicting the file's own "two different prompts a millisecond apart are two launches" (targets deserve the same rule). Worse, the refusal is indistinguishable from a failure to the caller: `launch` returns `false`, so `use-home-next-composer-state.ts:94-96` calls `restoreSubmittedDraft()` and the user's prompt reappears in the Home composer even though the launch it collapsed into is running fine. Include the target identity in the fingerprint, and give the dedupe path a distinct return so the composer does not roll back a draft whose launch actually started. (I could not construct a single-Enter double-fire: `ComposerRichTextEditor.tsx:309-316` guards `event.repeat`, calls `preventDefault()`, and the contenteditable never triggers the form's `onSubmit`; and the `flushSync` clear disables the send button before a second click can land. So the debounce is a belt-and-braces guard, which is the right call — it just should not be able to eat a legitimate second target.)

---

### Nits

- `use-home-next-launch.ts:47,111,393`: `isLaunching` is one boolean for N launches — the first launch to settle clears the send spinner and re-enables Escape-to-clear (`use-home-next-composer-state.ts:112-116`) while the others are still running. A count would match the new model.
- `launch-concurrency.ts:34-42`: the cap counts `awaiting-cloud-ready` entries, so five slow cloud workspaces block *local* launches too, and the toast ("Too many workspaces starting") doesn't say which ones. Deliberate reading of "starting", but worth a founder call.
- `sidebar-groups.ts:144-146`: a **failed** entry that already has a `workspaceId` (created, then first prompt failed) keeps suppressing the real workspace's row registry-wide and indefinitely. It is reachable and spec-sanctioned, but combined with #2 and #3 the user can end up with a real workspace hidden behind an orphan failed row and no obvious way back. Worth bounding.
- `ChatLaunchIntentPane.test.tsx:29`: `useChatLaunchIntentStore.setState({ activeIntent: null })` is rebase residue — the key no longer exists, so the afterEach reset is a no-op (harmless today with one test in the file, misleading for the next one).
- `launch-intent-registry.ts:83-98`: `launchIntents()` is only referenced by tests; production reads go through `launchIntent`/`resolveLaunchIntentForShell`.

---

### Checked and found sound

- **Registry conversion.** All `…IfActive` call sites re-keyed (`use-home-next-launch`, `use-home-deferred-launch-runner`, `home-next-launch-intent`, `use-home-next-launch-prompt-actions`, `session-creation-{publication,failure-cleanup,materialization-helpers}`, `use-session-creation-actions`, `use-chat-launch-intent-actions`, `launch-home-cloud-target`), each with its own intent id. Tree-wide grep for `activeIntent`/`IfActive` shows no leaked singleton reads. Store follows the conventions: one `set()` per mutator, identity return for a missing id so zustand skips the notify, stable `EMPTY_CHAT_LAUNCH_INTENT_REGISTRY` refs for initial state, and `useShellLaunchIntent` uses two stable-ref selectors plus a `useMemo` (no `useShallow` needed).
- **Shell resolution.** `resolveLaunchIntentForShell` cannot give one shell two intents (single winner, scoped beats unscoped) and cannot give one intent two shells (there is exactly one selection, and `useChatSurfaceState`/`useMainScreenState` both feed it from the same store). Unscoped intents match only an empty shell, so the Home first-paint window still works and a scoped intent can no longer hijack Home. The simplification of `activeLaunchIntentIdForShell` in `use-main-screen-state.ts` is strictly tighter than what it replaced, not looser.
- **Sidebar plurality.** `pendingOwnedWorkspaceIds` is built once across all entries and passed to every group's `buildSidebarWorkspaceItems`; `hasPendingSuppression` correctly fires on the id set alone, so an attempt whose materialized workspace sorts into a *different* group is still deduped. `allLogicalWorkspaceIds` filters with the same set. Cowork entries are excluded from repo groups by `buildPendingSidebarProjection` returning null *before* the id is added, and `CoworkThreadsSection` does its own plural suppression against `thread.workspaceId` — no group missed, and a cleared entry stops suppressing immediately, so a completed workspace reappears.
- **Cap semantics.** `countStartingPendingLaunches` skips `failed`, so five untouched failures cannot lock the user out; the cap is skipped for existing-workspace prompts (`createsWorkspace`), evaluated before any state is mutated, and its `return false` restores the draft — which is the right rollback for this path.
- **Per-entry dismissal.** `handleBack` clears the entry and looks the linked intent up by attempt id *after* removing the entry (order is safe — the lookup reads the intent registry, not the pending one) and leaves other launches untouched; `clearSelection({preservePendingWorkspaces:true})` in `clear-runtime-state.ts:61` keeps the rest of the registry alive. Failed entries stay inspectable: `runWorkspaceSelection` now short-circuits `pending-workspace:*` keys into `enterPendingWorkspaceAttemptShell` instead of throwing "Workspace not found.", and the row's error indicator carries an `open_workspace` action that routes through the same path.
- **Toast hygiene.** `notifyUnattendedPendingWorkspaceFailure` early-returns on attended attempts (so the receipt stays the only surface there), is called from exactly two non-overlapping failure paths, is id-keyed per attempt (`toast-model` documents same-id replace, so re-fires can't stack), and re-reads the entry by id on "Show" rather than capturing it.
- **`flushSync`.** Unchanged by this PR and correct where it is: called from an event handler before any `await`, so no lifecycle warning; it is what makes the second Enter fail `canSubmit` on an empty draft. It is now load-bearing rather than cosmetic, which is worth knowing but not wrong.
- **Rebase integrity.** No dead imports left in `use-main-screen-state.ts` (`launchIntentOwnsShell`/`resolveLaunchIntentScope` are still used by the registry and `use-chat-surface-state`), no duplicated logic between `launch-intent-registry` and `pending-entry-registry` beyond the deliberate mirror, and the cowork quiet-stop path resolves cleanly onto the keyed `clearLaunchIntent`. The only residue found is the test noted above.
- **Spec.** Invariant 11, the rewritten §Sidebar plurality/suppression bullets, the per-attempt dismissal wording and the expanded coverage list all match the code, and cloud background completion is correctly still absent. Two claims overreach: the app-start sweep (#3) and "nothing is announced twice" (#5).

pablonyx:
Fixed in e5be4db96. Per finding:

**1 (blocking) — cloud never threaded its attempt id.** `useHomeNextLaunch` now mints the attempt id in the cloud branch like the other three (`use-home-next-launch.ts`), passes it into `createCloudWorkspaceAndEnterWithResult` (new `attemptId` option on `CreateCloudWorkspaceAndEnterOptions`; the flow mints its own only when the caller does not), and assigns `launchAttemptId` so the catch scopes to it. `launch-home-cloud-target.ts` also passes `attemptId` to `promptProjectedPendingWorkspaceSession` (no more attendance fallback at `home-next-projected-session.ts:35-47`) and calls `markHomeLaunchIntentMaterializedFromPendingWorkspace` right after the create starts, so cloud scopes its intent synchronously like cowork/local/worktree instead of only on failure. Test: a cloud launch that fails while a *different cloud* attempt is attended keeps its own attempt id and never inherits the attended launch's workspace. Negative control: dropping `launchAttemptId = cloudAttemptId` makes it fail with `expected 'attended-other-launch' to be '<own attempt>'`.

**2 — retry orphaned the failed attempt.** `handleRetry` now starts the replacement first and ends the retried attempt second (`endAttempt`: entry + linked launch intent, shared with `handleBack`). Order matters: every create registers its new pending entry in its synchronous prefix, so the sidebar swaps one row for another rather than blinking. Applies to local, worktree, cloud and the cowork stop; `select-existing` reuses its own attempt and is untouched. Two tests.

**3 — GC was inert.** `useStaleFailedPendingWorkspaceGc` sweeps on mount *and* hourly. Test covers the case only the interval can reach: a failure registered after mount going stale mid-session. Spec bullet rewritten — it now says hourly + on mount, and states outright that app start is never when a stale row is found because the registry is in-memory.

**4 — nothing evicted a failed intent.** `retry` in `use-chat-launch-intent-actions.ts` clears the intent it retried once the replacement has run (not before it exists, so the pane never blinks empty). It keeps the failed intent when the retry never started — the cap path — because the pane is then the only place the prompt lives. Three tests in a new `use-chat-launch-intent-actions.test.tsx`, the first reproducing your exact case: unscoped failed intent on Home, replacement succeeds and clears itself, registry ends empty.

**5 — announced twice.** New `pendingWorkspaceFailureNoticeOwnsFailure(attemptId)` beside the notice (failed entry + unattended = the notice owns it); the launch-level "Work not started" toast is suppressed when it returns true. Attended failures are unchanged. Test asserts zero toasts unattended, exactly one attended; negative control (forcing the toast) fails it. Spec bullet corrected to describe both directions.

**6 — "Show" did not navigate.** `onNavigate` now calls `navigateApp("/")` before entering the shell — the callback-only navigation path (`lib/workflows/app/app-navigate-handoff.ts`) that `useNavigate`-free consumers already use, so no router subscription is added to a plain module. Test pins the route call.

**7 — debounce fingerprint / draft rollback.** `launchSubmitFingerprint(text, target, at)` includes a target identity (repo root + base branch, source root, owner/repo + branch, or cowork), and `isDuplicateLaunchSubmit` compares it. `launch` now returns `HomeNextLaunchOutcome` (`launched` | `not-started` | `duplicate` | `refused`) instead of a boolean: the composer restores the draft for everything except `launched` and `duplicate`, so a suppressed submit no longer rolls back a draft whose launch is running. Four fingerprint tests plus a launch-level test that the same prompt to two repo roots inside one second starts two launches.

**Layer boundary (found separately).** `scripts/check_frontend_boundaries.py` flagged the notice importing `#product/components/feedback/product-toast` from a runtime hook. It now raises through `useToastStore.getState().showError` — the sanctioned path (the store carries the one allowlisted upward import). `check_frontend_boundaries.py` passes clean.

Nits: `ChatLaunchIntentPane.test.tsx:29` now resets the registry shape. `isLaunching` is a count, so the first launch to settle no longer clears the spinner for the others.

Accepted as follow-ups, not in this commit:
- the cap counting `awaiting-cloud-ready` (five slow cloud workspaces blocking a local launch) — a founder call on what "starting" means, plus toast copy naming which ones;
- a failed entry with a `workspaceId` suppressing the real workspace's row indefinitely (`sidebar-groups.ts:144-146`) — needs a bound, and fix 2 removes the common way of reaching it;
- `launchIntents()` being test-only — it is the registry's natural read and its tests are the only production-shaped consumer; leaving it.

Full product-client suite green (866 files / 6063 tests), both typechecks clean.

pablonyx:
Rebased the whole stack (#1867 → #1868 → #1870 → #1873) onto current `origin/main`.

Two reasons this was needed, both resolved:

- **Login JS cap.** main raised the cap to 500500 in #1875; the stack had been measuring against the stale 490000. At the stack top (#1873) the budget now measures **JS 496571 / 500500 B** — passing, ~3.9 kB headroom locally (~2.1 kB after allowing for local gzip under-reporting CI by ~0.36%).
- **Sidebar conflict with #1837 (archiving R7).** #1837 reworked `buildSidebarWorkspaceItems`, dropping the `archivedSet` argument in favour of `cloudWorkspace.productLifecycle`. Resolved by adopting main's model wholesale — `archivedSet` is gone from `sidebar-groups.ts` and `sidebar-workspace-items.ts`, and nothing #1837 removed was reintroduced. The stack's pending-row work is rethreaded on top of that shape: plural `pendingItems` per group and registry-wide `pendingOwnedWorkspaceIds` suppression keyed off `entry.workspaceId`.

Checked the archiving model for semantic interactions with the pending registry: pending rows are always `archived: false`, suppression runs before the archived/`showArchived` filter, and archive/unarchive never mints a pending entry — so the two features do not interact.

Per branch: product-client suite, both tsconfigs, and the three frontend gate scripts are green.

PR #1873, opened 2026-08-14
fix(product): complete cloud workspace launches in the background (PRO-230)

pablonyx:
## Problem

A cloud launch only finished while the user was watching it. `useCloudWorkspacePolling` polled the *selected* workspace, and the deferred launch runner promoted a queued prompt only once the selected workspace's runtime was ready. Start a cloud workspace, click away to another one, and the first attempt parked at `awaiting-cloud-ready` — sandbox provisioning kept going server-side, but nothing on the client ever noticed it finish, so the prompt stayed queued and the sidebar row stayed a spinner until the user re-attended the attempt and effectively hand-cranked it.

That is the scope gap in Invariant 9. The rest of the ladder made launches per-attempt; this one was still per-selection. Liveness is supposed to gate the pipeline and attendance only the presentation, and here attendance was gating the pipeline.

## Design

**Poll the registry, not the selection.** `useCloudWorkspacePolling` now walks every `awaiting-cloud-ready` entry in the pending registry rather than the selected workspace. The tick is a capped rotating batch (`cloud-workspace-poll-plan.ts`): at most three workspaces per tick, rotating by a cursor so an entry at the back of the queue is never starved while the cap keeps a five-launch registry from turning into five concurrent status requests. Each poll writes its outcome to its own entry through `patchAttempt`, so two launches in flight cannot cross-write each other's stage.

**Promote per launch, selection-free.** `deferred-launch-readiness.ts` decides readiness from the launch's own workspace and its own pending entry rather than from what the user is looking at. The runner promotes each ready launch independently — sends the queued prompt, clears the entry — and only *activates* (steals selection and the active session) when the attempt is attended. Unattended launches complete silently in the background and are there when the user comes back.

**Failure lands where the ladder already puts it.** A readiness failure on an unattended attempt marks the entry `failed` and announces it through the per-entry failure presentation from #1870 — the sidebar error row plus the one-shot toast — instead of the shell receipt that belongs to whatever workspace is on screen.

## Files

- `hooks/chat/lifecycle/use-cloud-workspace-polling.ts` — registry-wide capped rotating poll, per-entry `patchAttempt` outcomes
- `lib/domain/workspaces/cloud/cloud-workspace-poll-plan.ts` — the batch/cursor plan, extracted as pure domain
- `hooks/home/lifecycle/use-home-deferred-launch-runner.ts` — per-launch promotion, attended-gated activation
- `lib/domain/home/deferred-launch-readiness.ts` — readiness from the launch's own workspace + entry
- `hooks/workspaces/derived/use-pending-workspace-entries.ts`, `hooks/workspaces/workflows/pending-workspace-attempt-access.ts`, `lib/domain/workspaces/creation/pending-attention.ts` — the awaiting-entry selector and attendance accessors the two hooks read
- `specs/codebase/systems/product/workspaces/pending-shell.md` — background completion documented

11 files, +1153/−299.

## Tests

- `use-cloud-workspace-polling.test.tsx` — 9 tests: unattended finalization without moving selection, every awaiting entry polled, the three-per-tick cap, polling stops when an entry leaves `awaiting-cloud-ready`, unattended failure marked and announced.
- `use-home-deferred-launch-runner.test.tsx` — 6 tests: background promotion without stealing selection or the active session, activation when attended, the pending shell of an attended attempt counted as attended, two launches promoted independently, waiting while one workspace is still provisioning, queued prompt released on a failed attempt.
- `cloud-workspace-poll-plan.test.ts` — 10 tests on the cap and cursor rotation.

Both negative controls run and observed:

- Restricting the polling hook back to attended entries (`useAwaitingCloudWorkspaceEntries().filter(e => isAttemptAttended(e.attemptId))`) fails 5 of the 9 polling tests — including "finalizes an unattended awaiting entry without moving selection" and "marks an unattended awaiting entry failed and announces it when readiness fails".
- Gating promotion on attendance in the runner (the pre-fix behaviour) fails 3 of the 6 runner tests — "promotes a background launch without stealing selection or the active session", "promotes two deferred launches independently", "waits while one launch's workspace is still provisioning".

Both files restored; full product-client suite green (868 files, 6084 tests), `tsconfig.json` and `tsconfig.domain.json` typechecks clean, frontend boundary check clean.

## Ladder position

Final slice (4 of 4) of the PRO-230 launch registry ladder, stacked on #1870.

https://linear.app/proliferate-team/issue/PRO-230/stitching-workspace-while-a-new-workspace-destroys-the-new-workspace

## Known follow-ups

- `use-selected-cloud-runtime-rehydration` still carries an `awaiting-cloud-ready` finalization branch that the registry-wide poll now reaches first. It is race-safe — both paths converge on the same `patchAttempt` — but it is dead weight and should come out in a follow-up rather than inside this slice.
- Unattended cloud materialization against a workspace that never connects deserves one live check before merge. The tests cover the readiness-failure path with a stubbed status; the real never-connects timeout is worth watching once end-to-end.

### DIFF
--- a/apps/packages/product-client/src/components/workspace/chat/ChatView.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/ChatView.tsx
@@ -31,7 +31,6 @@ import { useChatAvailabilityState } from "#product/hooks/chat/derived/use-chat-a
 import { useChatDockInset } from "#product/hooks/chat/ui/use-chat-dock-inset";
 import { useChatPromptAttachments } from "#product/hooks/chat/ui/use-chat-prompt-attachments";
 import { useChatRootFocus } from "#product/hooks/chat/ui/use-chat-root-focus";
-import { useCloudWorkspacePolling } from "#product/hooks/chat/lifecycle/use-cloud-workspace-polling";
 import { useComposerDockSlots } from "#product/hooks/chat/ui/use-composer-dock-slots";
 import { useQueuedPromptEditStatus } from "#product/hooks/chat/ui/use-queued-prompt-edit";
 import { useDebugRenderCount } from "#product/hooks/ui/debug/use-debug-render-count";
@@ -230,7 +229,6 @@ export const ChatView = memo(function ChatView({
     stickyNonDisplacingBottomInsetPx,
   } = useChatDockInset();
 
-  useCloudWorkspacePolling();
   useSelectedCloudRuntimeRehydration(selectedCloudRuntime);
   useSessionErrorAcknowledgement();
 
--- a/apps/packages/product-client/src/hooks/chat/lifecycle/use-cloud-workspace-polling.test.tsx
+++ b/apps/packages/product-client/src/hooks/chat/lifecycle/use-cloud-workspace-polling.test.tsx
@@ -1,11 +1,12 @@
 // @vitest-environment jsdom
 
-import { cleanup, renderHook, waitFor } from "@testing-library/react";
+import { act, cleanup, renderHook, waitFor } from "@testing-library/react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { CloudWorkspaceSummary } from "#product/lib/domain/workspaces/cloud/cloud-workspace-model";
 import {
   buildPendingWorkspaceUiKey,
   buildSubmittingPendingWorkspaceEntry,
+  type PendingWorkspaceEntry,
 } from "#product/lib/domain/workspaces/creation/pending-entry";
 import {
   createEmptySessionRecord,
@@ -26,9 +27,14 @@ const mocks = vi.hoisted(() => ({
   selectWorkspace: vi.fn(),
   materializePendingWorkspaceSessions: vi.fn(),
   trackWorkspaceInteraction: vi.fn(),
+  notifyUnattendedPendingWorkspaceFailure: vi.fn(),
   workspaceCollections: {
     cloudWorkspaces: [] as CloudWorkspaceSummary[],
   },
+}));
+
+vi.mock("#product/hooks/workspaces/workflows/pending-workspace-failure-notice", () => ({
+  notifyUnattendedPendingWorkspaceFailure: mocks.notifyUnattendedPendingWorkspaceFailure,
 }));
 
 vi.mock("#product/stores/preferences/workspace-ui-store", async (importOriginal) => {
@@ -74,6 +80,12 @@ describe("useCloudWorkspacePolling", () => {
     mocks.selectWorkspace.mockReset();
     mocks.materializePendingWorkspaceSessions.mockReset();
     mocks.trackWorkspaceInteraction.mockReset();
+    mocks.notifyUnattendedPendingWorkspaceFailure.mockReset();
+    mocks.materializePendingWorkspaceSessions.mockReturnValue({
+      pendingWorkspaceUiKey: "pending-workspace:attempt-1",
+      projectedSessionCount: 0,
+      projectedSessionIds: [],
+    });
     mocks.workspaceCollections.cloudWorkspaces = [cloudWorkspace({ status: "pending" })];
     useSessionDirectoryStore.getState().clearEntries();
     useSessionTranscriptStore.getState().clearEntries();
@@ -153,7 +165,7 @@ describe("useCloudWorkspacePolling", () => {
     expect(mocks.materializePendingWorkspaceSessions).toHaveBeenCalledWith(
       pendingEntry,
       workspaceId,
-      { eventPrefix: "workspace.cloud_polling" },
+      { eventPrefix: "workspace.cloud_polling", attended: true },
     );
     await waitFor(() => {
       expect(mocks.trackWorkspaceInteraction).toHaveBeenCalledWith(
@@ -257,7 +269,7 @@ describe("useCloudWorkspacePolling", () => {
     expect(mocks.materializePendingWorkspaceSessions).toHaveBeenCalledWith(
       pendingEntry,
       workspaceId,
-      { eventPrefix: "workspace.cloud_polling" },
+      { eventPrefix: "workspace.cloud_polling", attended: true },
     );
     expect(readPendingEntry()).toBeNull();
   });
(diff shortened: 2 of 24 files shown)


pablonyx:
## Review — PR #1873 (PRO-230 PR3, `ccf8127b1` over `e5be4db96`)

The mechanism is right and the hard parts are actually right. Attendance is captured once per sequence before the force-select and threaded into materialization rather than re-read after the await; `isLaunchAttemptAttended` is the correct accessor for the post-clear window because `buildPendingWorkspaceUiKey` needs only `attemptId`; per-entry `patchAttempt` means two launches cannot cross-write; the cursor rotation starves nothing; and the race with the leftover rehydration finisher is genuinely safe for the reason the PR body claims, plus one it does not state — neither finalize path has an `await` between its entry re-read and `clearPendingWorkspaceEntry`, so a concurrent finalize always observes `null` and returns. The readiness signal ordering also holds: `refreshCloudWorkspace` is `mutateAsync`, so its promise resolves only after `onSuccess` has upserted the record into the collections cache, which means the runner cannot lose the signal between the upsert and the entry being cleared.

One blocking problem, and it is about where the loop lives rather than what it does. `useCloudWorkspacePolling` is still mounted inside `ChatView`, and `ChatView` is unmounted whenever nothing is selected — which is exactly the state the app's own "go Home" affordances put the user in. So the registry-wide loop stops for every parked attempt the moment the user leaves the workspace shell entirely, and Invariant 9 as the spec now states it is not unconditional in code. The fix is small (move the call up to `ProductLifecycleRoot`, beside `useHomeDeferredLaunchRunner`), which is why I would rather see it inside this slice than as a follow-up.

---

### Blocking

**1. The registry-wide poll loop is unmounted whenever no workspace is selected, which is where "go Home" lands the user.**

`useCloudWorkspacePolling()` is called at `apps/packages/product-client/src/components/workspace/chat/ChatView.tsx:239`. `ChatView` reaches the tree only through `WorkspaceContentView` under the `hasWorkspaceShell` branch of `apps/packages/product-client/src/components/workspace/shell/screen/StandardWorkspaceShell.tsx:304` (the sibling `hasLaunchIntentOnlyShell` branch is dead — `apps/packages/product-client/src/hooks/main/facade/use-main-screen-state.ts:128` hardcodes it to `false`). `hasWorkspaceShell` is `selectedWorkspaceId || attended pending entry || shell-owned launch intent` (`use-main-screen-state.ts:129` → `apps/packages/product-client/src/lib/domain/chat/surface/chat-surface.ts:61`). With none of those, the shell renders `<HomeNextScreen />` at `StandardWorkspaceShell.tsx:346` and `ChatView` — and the loop — is gone.

Two ordinary affordances produce exactly that state:

- `goToTopLevelRoute` at `apps/packages/product-client/src/hooks/workspaces/workflows/use-workspace-navigation-workflow.ts:43-48` calls `deselectWorkspacePreservingSessions()` whenever anything is selected. It is behind the sidebar Home button (`apps/packages/product-client/src/hooks/workspaces/workflows/use-workspace-sidebar-actions.ts:54`), and behind `goHome`/`goWorkflows` and their shortcuts (`apps/packages/product-client/src/hooks/app/workflows/use-app-navigation-command-actions.ts:38-43`).
- `returnHome` at `apps/packages/product-client/src/hooks/chat/workflows/use-chat-launch-intent-actions.ts:47-56` does the same plus `setSelectedLogicalWorkspaceId(null)`.

`deselectWorkspacePreservingSessions` (`apps/packages/product-client/src/stores/sessions/session-selection-store.ts:175`) nulls both selection ids, and the pending entry is only "attended" through those ids, so the attended-entry term goes false too.

Failure scenario: start a cloud workspace from Home, then click Home (or Workflows, or Workspaces) in the sidebar and stay there. Provisioning continues server-side; nothing on the client polls any parked attempt; the entry stays at `awaiting-cloud-ready`, its projected session is never materialized, and its first prompt is never sent — the pre-PR3 bug, reached by a different route. It is worse in the deferred-launch fallback branch (`launch-home-cloud-target.ts:165`, the no-projected-session path): `useHomeDeferredLaunchRunner` lives at `apps/packages/product-client/src/providers/ProductLifecycleRoot.tsx:166` and stays mounted, and the collections query self-refetches every 30s while any cloud workspace is provisioning (`apps/packages/product-client/src/hooks/workspaces/cache/use-workspaces.ts:229-231`), so `resolveDeferredLaunchReadiness` flips to `ready` and the runner promotes — creating the session and sending the prompt — against an entry that nothing will ever finalize. Result: a permanent pending spinner in the sidebar next to a live workspace with the prompt already sent.

Both hook tests render the hook directly, so residency is structurally invisible to them; the negative controls in the PR body exercise the filter, not the mount site.

Fix: move `useCloudWorkspacePolling()` to `ProductLifecycleRoot`, next to the runner. A registry-wide loop no longer belongs to a workspace's chat surface. (If that lands, it is also worth checking whether `useSelectedCloudRuntimeRehydration` should stay where it is — it is selection-scoped by construction, so it should.)

### Should fix

**2. The polling interval restarts on far more than the attempt set, and each restart fires a tick immediately.**

`use-cloud-workspace-polling.ts:237-272`. The comment says restarting on the attempt set alone keeps one interval alive across status churn, but the dep array is `[awaitingAttemptKey, pollAwaitingEntry]`, and `pollAwaitingEntry`'s identity chain reaches the very data the poll mutates:

`pollAwaitingEntry` (deps incl. `finalizeAwaitingEntry`) → `finalizeAwaitingEntry` (deps incl. `materializePendingWorkspaceSessions`, `selectWorkspace`) → `usePendingWorkspaceSessionMaterialization` (`use-pending-workspace-session-materialization.ts:172`, dep `createEmptySessionWithResolvedConfig`) → `use-session-creation-actions.ts:452,430-449` (deps incl. `getWorkspaceRuntimeBlockError`) → `use-workspace-runtime-block.ts:47-52` (deps `selectedCloudRuntime.state`, `selectedCloudRuntime.workspaceId`, `workspaceCollections?.workspaces`).

Every successful poll runs `upsertCloudWorkspace` + `await invalidateWorkspaceCollections()` in `onSuccess` (`apps/packages/product-client/src/hooks/cloud/workflows/use-cloud-workspace-actions.ts:62-65`), and the selected cloud workspace's `state` changes as it transitions — so the poll's own writes are what churn these deps. On each churn the effect tears down (clearing the pending 3s timer) and immediately runs `void runTick()` at line 264. There is no floor on tick spacing, and two ticks can overlap (safe for correctness, per the re-read reasoning above, but it doubles refreshes and can double-refresh the same attempt across the boundary). React Query's structural sharing dampens this but does not guarantee it, and the amount of extra traffic depends on whether the collections payload carries per-fetch-varying fields.

Cheap fix that matches the stated intent: keep `pollAwaitingEntry` in a ref updated each render (as `awaitingEntriesRef`/`cloudWorkspacesRef` already are) and depend only on `awaitingAttemptKey`.

**3. An unattended create failure is announced with attended copy, and the runner's own catch cannot see it.**

`use-home-deferred-launch-runner.ts:171-210` awaits `createSessionWithResolvedConfig`. Because a deferred launch always carries a prompt, that call returns as soon as the prompt is enqueued — `use-session-creation-actions.ts:387-402` does `if (hasPrompt) { void createPromise.catch(cleanupCreateFailure…); return pendingSessionId; }`. So the runner always takes the success branch and clears both the launch and the launch intent (lines 185-186), and the catch's recovery branches never see a create failure. The actual failure surfaces as `cleanupSessionCreationFailure` plus a global error toast reading "Message not sent / No chat was opened. Your message is still in the composer." — for a workspace the user is not looking at, whose composer holds nothing, with no workspace named. `cleanupSessionCreationFailure` then calls `clearLaunchIntent` on an intent the runner already cleared.

This answers the PR body's open question on dropping the connection gate: the blast radius of unattended materialization is *not* a lost failure, and it is not a resolution problem either — `resolveRuntimeTargetForWorkspace` (`apps/packages/product-client/src/lib/access/anyharness/runtime-target.ts:68-96`) fetches the cloud workspace and resolves its own gateway connection per workspace, independent of whatever runtime the client is currently connected to, and the prompt dispatcher is global and selection-free (`apps/packages/product-client/src/hooks/sessions/lifecycle/use-session-intent-dispatcher.ts:140-177`). The gap is presentation only: an unattended failure should land on the attempt's own surface (as PR2's notice does) rather than on the composer-shaped toast.

**4. A parked attempt whose workspace reaches a terminal non-error status is never resolved.**

`cloud-workspace-poll-plan.ts:40-47`: status `lost` (and any other terminal non-`ready`, non-`error` status) is neither `fail-cached` nor pollable, so `resolveCloudWorkspacePollAction` returns `"skip"`. The entry stays at `awaiting-cloud-ready` indefinitely, keeps occupying a rotation slot, keeps the interval alive forever, and the deferred launch behind it waits out the full `DEFERRED_HOME_LAUNCH_STALE_MS` hour. Attended, the user at least gets the status screen (`shouldShowCloudWorkspaceStatusScreen` covers `lost`); unattended there is nothing but a spinner. Now that the spec says provisioning failures land on the attempt's own entry, `lost` deserves the `fail-cached` treatment.

### Nits

**5.** `use-selected-cloud-runtime-rehydration.ts:132` materializes without passing `attended`, so `use-pending-workspace-session-materialization.ts:142` re-reads attendance *after* the `bootstrapWorkspace` await — the pattern PR1 established against. It is safe today only because the `cancelled` guard at line 110 covers the switch-away case. Since the branch is already slated for deletion, no action needed, but worth not copying.

**6.** `use-home-deferred-launch-runner.ts:187-191`: `markConsuming` removes the launch from `launchesByReadiness.ready`, which changes `readyLaunchKey`, which runs the effect cleanup and sets `cancelled = true` — so by the time any error surfaces, `isCancelled()` is essentially always true and the launch is marked pending again and retried with no backoff, while the two informative recovery branches below are unreachable. Carried over unchanged from the base, and near-dead in practice given finding 3, but the shape is a retry loop waiting for a synchronous throw.

**7.** Spec: `pending-shell.md` Invariant 9 now asserts background cloud completion unconditionally, and the §13 bullet says the attempt "holds its registry entry and its queued prompt while polling keeps working on it". Both are conditional on a workspace (or pending shell) being selected — see finding 1. Either fix the residency and the text is accurate as written, or the condition needs to be stated.

### What I checked and found sound

- **No selection steal.** Attendance is read once at `use-cloud-workspace-polling.ts:119`, before the force-select can make every later read look attended, and passed explicitly into materialization (`{ attended }` at line 158 → `activateOnCreate` / `targetWorkspaceUiKey` at `use-pending-workspace-session-materialization.ts:142-153`). The unattended path never calls `selectWorkspace` or `setWorkspaceArrivalEvent`. A user switching during the finalize await cannot flip the decision. In the runner, `isLaunchAttemptAttended` is captured once at line 166 and `activateOnCreate` is consumed synchronously inside the create (before any await), so a mid-flight switch cannot retroactively steal.
- **`isLaunchAttemptAttended` in the mid-handoff window.** `buildPendingWorkspaceUiKey` (`pending-entry.ts:89`) is a pure function of `attemptId`, so the identity-only accessor produces the same pending key the shell selection uses; the second clause matches `selectedWorkspaceId === workspaceId`, which is precisely "the user is looking at the real workspace the ids just swapped to". Correct in both halves of the handoff.
- **Cursor math.** `selectCloudWorkspacePollBatch` normalizes negative/oversized cursors, emits distinct entries within a tick, and carries `nextCursor` so the entries past the cap are polled on the following tick. No starvation, no intra-tick double-poll.
- **Entries leaving the awaiting state stop being polled.** Two layers: `useAwaitingCloudWorkspaceEntries` filters them out of the rotation entirely (so `awaitingAttemptKey` shrinks and the interval ends when the last one leaves), and `resolveCloudWorkspacePollAction` returns `"skip"` defensively. The id re-read at `pollAwaitingEntry` line 189 correctly makes a dismissed/failed/finalized attempt a no-op between scheduling and running.
- **Overlapping finalizes.** Verified both finalize bodies: no `await` between the entry re-read and `clearPendingWorkspaceEntry`, so the second caller always sees `null` or a non-awaiting stage and returns. This is what makes the leftover rehydration finisher safe, not just convention — and rehydration dispatches no prompts, so no double prompt-dispatch is possible either.
- **Readiness signal integrity.** `refreshCloudWorkspace` is `refreshMutation.mutateAsync`, whose promise resolves after `onSuccess` upserts into the collections cache — so on the polling path the record is always in the cache before the entry is cleared; the signal cannot be lost. Deferred launch ids are keyed `cloudWorkspaceId + attemptId` (`buildDeferredHomeLaunchId`), and cloud workspace ids are not reused across attempts, so a stale-ready record cannot serve a new attempt.
- **Failure path.** `failAwaitingEntry` patches by `attemptId` only. `notifyUnattendedPendingWorkspaceFailure` self-suppresses when attended, so an attended provisioning failure keeps its shell receipt and raises no toast. No double announcement with `pendingWorkspaceFailureNoticeOwnsFailure`: that guard sits on the synchronous create-time path in `use-home-next-launch.ts:410`, while the async readiness failure raises only the one notice and the runner's failed branch deliberately fails the intent without a second toast.
- **Restack integrity.** No dead imports, no stale PR2 call shapes (`notifyUnattendedPendingWorkspaceFailure(entry, message)`, `failLaunchIntent({message, retryMode})`, `HomeNextLaunchOutcome` all match), and the four helpers the polling hook stopped importing (`isCloudWorkspacePostReadyPending`, `shouldShowCloudWorkspaceStatusScreen`, `shouldPollCloudWorkspaceForUpdates`, `pendingWorkspaceEntryForWorkspaceId`) all still have live consumers.
- **Pure-module conventions.** `cloud-workspace-poll-plan.ts` and `deferred-launch-readiness.ts` are side-effect free and import only domain. `useAwaitingCloudWorkspaceEntries` returns a stable `EMPTY_PENDING_WORKSPACE_ENTRIES` ref, and the runner's readiness buckets return stable `EMPTY_DEFERRED_LAUNCHES`. The only store write added is `patchAttempt`, which is a single `set()` through `setPendingWorkspaceEntry`.

pablonyx:
Review findings addressed in `def308155`.

**1 (blocking) — polling died with the workspace shell.** `useCloudWorkspacePolling()` now mounts in `ProductLifecycleRoot`, beside the deferred-launch runner that consumes what it finalizes, instead of in `ChatView` (which only renders under `StandardWorkspaceShell`'s `hasWorkspaceShell` branch, so Home and `/workflows` unmounted it and parked every cloud attempt while the runner kept promoting off the collections refetch). The hook takes nothing ChatView-scoped — stores, `useWorkspaces`, `useCloudWorkspaceActions`, `useWorkspaceSelection`, `usePendingWorkspaceSessionMaterialization` — all valid at the root. `useSelectedCloudRuntimeRehydration` stays in `ChatView`; it is genuinely selection-scoped. Guarded by a new `ProductLifecycleRoot.test.tsx` case asserting both drivers mount at the root. The hook and its test moved `hooks/chat/lifecycle/` → `hooks/workspaces/lifecycle/` so the path matches the residency (deviation beyond the literal ask, noted here).

**2 — restart churn / tick storm.** The tick calls the newest `pollAwaitingEntry` through a ref; the effect now depends only on the joined attempt-id set. Restarts that do happen (an attempt joining or leaving) schedule off `lastTickStartedAtRef` instead of ticking immediately, so the interval is a floor including restarts. Two tests: one drives fresh callback identities per render and asserts one tick per interval; one asserts the first tick after a join is spaced.

**3 — unattended send failure showed composer copy.** Presentation-only. `CreateSessionWithResolvedConfigOptions` gained one optional `onQueuedPromptFailure`; the creation workflow calls it instead of the "still in the composer" toast when the caller supplies it. The deferred-launch runner supplies it only when the launch is unattended, and announces through a new id-keyed `notifyQueuedPromptSendFailure` that names the workspace and offers **Show** (navigates Home, then selects the workspace) — mirroring PR2's id-keyed notice. The send pipeline itself is untouched. It could not simply read selection inside the creation hook: `use-workspace-bootstrap-actions` already imports it, so importing `useWorkspaceSelection` there would cycle.

**4 — terminal non-ready statuses parked forever.** New `isCloudWorkspaceTerminallyUnavailable` (`lost`, `archived`); the poll plan returns `fail-cached` rather than `skip`, the outcome resolves to `failed`, and the failure message names what happened when the record carries no error text. `resolveDeferredLaunchReadiness` inherits this through the outcome, so the queued prompt is released instead of waiting out the hour-long staleness timer.

**5 — nits.** Attendance capture in `use-selected-cloud-runtime-rehydration`'s finalization branch is now taken before the await and threaded through, per PR1's convention. The unreachable catch / `markPending` retry loop in the runner is removed: `markConsuming` itself re-runs the effect, so the cancellation check was true on every failure and put the launch back to pending — an unbounded retry of something that failed for a reason that will not change. The store action stays (still covered by its own test).

Spec: `pending-shell.md` invariant 9 now states that both drivers are mounted by `ProductLifecycleRoot` and why, plus §13/§14 bullets for the terminal-status rule and the queued-prompt-failure ownership.

**Verification.** Product-client suite 869 files / 6098 tests green; product-client and desktop `tsc --noEmit` clean; `check_frontend_boundaries.py`, `check_toast_copy.py`, `check_docs.py` green. Negative controls (scratchpad copies, not stash): reverting the terminal-status fix → 5 failures; reverting the ref/cadence fix → "expected 1 times, but got 6"; reverting the root mount → residency test fails; reverting the unattended callback → announcement test fails.

**Deferred.** "Repo shape checks" is red on two oversized test files that predate this commit — `use-home-next-launch.test.tsx` (790) and `pending-sidebar-projection.test.ts` (664), both grown by #1868/#1870 in this stack. Trimming them here would conflict when those PRs restack, so they belong upstream. My own polling test crossed 600 while growing and was brought back to 540 by extracting `src/test/cloud-workspace-fixtures.ts`, now shared by the four cloud-launch suites.

PR #1880, opened 2026-08-14
feat(product): workflows gen-2 run view — graph pane, docs, controls, resume (PR6)

pablonyx:
# Workflows gen-2 — PR6: run view (graph pane, docs, controls, resume)

**Stack:** based on `codex/workflows-gen2-pr5b-sdk-courier` (#1877). Two PR5b repair commits ride below this PR in the stack and appear there, not here: the PR2-contract reconciliation (`df2fce077`) and the query-key size-threshold split (`baae16563`). Frozen delivery spec: `~/workflows-gen2-overnight/specs/pr6.md` (external per the #1817 spec-custody precedent); this body reproduces its intent.

## Intent

Give a live gen-2 run a real surface, entirely behind the `workflows_v2` flag (still OFF): a right-panel workflow tool showing the run as a chain of node cards whose controls are the ADR transition table verbatim, the run's produced docs opening at their real workspace paths, an undo toast for genuinely automatic advances, and a resume popover for interrupted runs.

## What changed

- **Right-panel integration** — `"workflow"` joins the `RightPanelTool` union. The build-time flag gates tool *availability* (the one place all ~12 readers of persisted `headerOrder` agree); whether the workspace has any runs filters the *rendered* header entry — so polls never rewrite persisted workspace state, and a persisted `tool:workflow` still parses under a gate-off build.
- **`domain/workflows/run-view-model.ts`** (frozen contract) — status→tone map, `workflowNodeControls` (transition table: approve only `awaiting_human`; fail-redo on `failed`/`needs_attention`/`awaiting_human`; flip-to-agent only `awaiting_human`+`human_in_loop`; flip-to-human only `running`+`agent`; ad hoc on any active run), chain-slot grouping (attempts + ad hoc per `chainIndex`, `byCreatedAt`), and `detectWorkflowAutoAdvance`.
- **`WorkflowPane` + `use-workflow-pane` facade** — projection polling (3s while `running`/`awaiting_human`), all six mutations write the returned projection back via `setQueryData` (never invalidate), 409 races match the ProblemDetails *code*, refetch, and toast — a race is not a UI bug.
- **`WorkflowGraphNodeCard`** — `Card`/`RosterRow`/`StatusDot`/`Badge` composition; fail-redo and ad hoc prompts via `ModalShell` dialogs; controls render only where the table has a legal row; needs-input badge joins the session roster's `waiting_input` activity state.
[...]

PR #1884, opened 2026-08-14
feat(workflows): gen-2 PR5a — the /v1/workflow-runs API

pablonyx:
PR5a of the workflows gen-2 chain (stacked on #1883) — the last runtime rung. Delivery spec frozen in the first commit (`delivery-spec-workflows-gen2-pr5a.md`); spec of record is the Workflows ADR's runtime-plane API table. Lane C's client chain (#1877 → #1880 → #1882) restacks onto this branch.

## What this adds

The nine ADR routes on the existing /v1 router:

- **PUT /v1/workflow-runs/{run_id}** — the courier's one call. Parses and fully revalidates the reconstituted flat snapshot (parse failure and validation failure are the same 400), resolves placement, materializes disk, inserts rows, starts the actor, answers 201 with the projection. An existing run id answers 200 with its rows untouched (and rematerializes the actor if the runtime restarted since — the courier's crash-retry path).
- **GET run / GET list** — projections straight from rows, no actor involvement. List envelope is `{ runs: [...] }` with an optional `workspace_id` filter.
- **Six command POSTs** (approve / fail-redo / type / undo-advance / resume / adhoc-nodes) — each hands the manager a command and returns the oneshot's fresh full projection; the transition table's refusal maps to the 409.

### Placement (ADR-silent points ruled boring and journaled)

- `placement.repoConfigId` resolves as the runtime repo-roots row id; the courier owns sending a runtime-resolvable id. Unknown id → 503, nothing inserted.
- `mode: worktree` reuses the surviving gen-1 seam (`resolve_workflow_placement` + `ensure_workflow_workspace`): deterministic `<managed-root>/workflows/<run_id>`, branch `workflow/<run_id>`, fail-closed adoption of crash artifacts. `base_ref` = the repo root's `default_branch`, falling back to `HEAD`.
- `mode: repo_root` resolves-or-creates the workspace at the repo root's own path.
- **Disk before rows**: context docs are planned pre-insert from the pure `NN-slug.md` filename law, materialized together with the shared `/.proliferate/` exclude entry, and only then does the one-transaction row insert run — a workspace or disk failure is the retry-safe 503 with zero rows, exactly the ADR's contract.

### The wire projection

`{ run, nodes[], docs[] }` per the ADR API table, matching Lane C's hand-authored `workflow-runs-v2.ts` field-for-field: camelCase, RAW `definitionJson`/`argumentsJson` strings (the verbatim-snapshot ruling extends to the wire), and nullable fields serialized as explicit `null` (the TS side declares `string | null`, not optional). This replaces the PR1-era flat domain projection (parsed definition, omitted nulls, no `promptId`), which predated the ADR table and never shipped to a client.

ProblemDetails codes: `WORKFLOW_RUN_NOT_FOUND` 404, `WORKFLOW_NODE_NOT_FOUND` 404 (node routes pre-check the run's rows; the ADR pins unknown-node to 404, not the table's 409), `WORKFLOW_TRANSITION_ILLEGAL` 409 (detail names the command and refused state), `WORKFLOW_SNAPSHOT_INVALID` 400, `WORKFLOW_WORKSPACE_MATERIALIZATION_FAILED` 503.

OpenAPI paths + schemas registered; `anyharness/sdk` generated types recut (`npm run generate`, typecheck clean).

## Cross-lane contract notes (for the restack + morning review)

- **Invocation id gap**: Lane C's courier PUTs `{ schemaVersion, workflowDefinitionId, definition, arguments, placement }` — no invocation `id`, but `workflow_runs.invocation_id` is non-null. PUT accepts an optional `id` (the frozen record's own id, per the contract fixture) and falls back to the run id when absent. Either Lane C adds `id: invocation.id` to the reconstituted body (one line) or the fallback stands — flagged, not silently absorbed.
- Lane C's guessed list envelope `{ runs }` is confirmed — no reconciliation needed there.

## Testing

Tier-1 route suite (`api/workflow_runs_route_tests.rs`): tower-oneshot over a real `AppState`, real SQLite, a real git repository placed through the real workspace seam, and the house scripted ACP agent behind the launched sessions — no machinery mocked.

1. PUT places and starts: 201 wire shape pinned (camelCase, raw strings, explicit nulls), run worktree on disk under the managed root, seeded `.proliferate/context/00-notes.md`, `/.proliferate/` in the clone's shared `info/exclude`; idempotent replay 200 returns the same rows; the held run then completes through the real actor.
2. Invalid snapshot → 400 `WORKFLOW_SNAPSHOT_INVALID`, zero rows (GET 404).
3. Unknown repoConfigId → 503 `WORKFLOW_WORKSPACE_MATERIALIZATION_FAILED`, zero rows.
4. Reads: unknown-run 404; GET returns the row projection; list filters by workspace and answers `{ runs }`.
5. Commands: unknown run 404 / unknown node 404; approve answers the fresh projection with the gate completed in-body; double-approve → 409 naming `approve_gate`; the run completes.

Negative control: sabotaging the Illegal→409 mapping (to 404) fails scenario 5 at the conflict assertion; restored, all green. Full `cargo test --workspace --no-fail-fast` green except the two documented pre-existing reds (lib sweep_tests TZ-environmental, collector_process ECONNRESET flake).

---

## Review wave (c8d768a57)

Applied the Opus review's F1–F13 per the conductor rulings:

- **Ruling A (F1):** unknown `placement.repoConfigId` → **400 `WORKFLOW_SNAPSHOT_INVALID`** (was 503). `repoConfigId` is the runtime repo-root id end-to-end; `definition.default_repo_config_id` is v1-unused (journaled).
- **Ruling B (F2):** `repo_root` mode resolves-or-reuses the existing workspace at the root path (no duplicate Local rows), both modes carry `WorkspaceCreatorContext::Workflow{run_id}`, and the **one-live-run law** answers **409 `WORKFLOW_PLACEMENT_CONFLICT`** — pre-checked before touching the checkout and re-enforced inside the insert transaction (typed `WorkspaceOccupied`, race-proof). New repo_root-mode tests cover reuse, conflict, and post-completion release.
- **Ruling C (F3/F4):** run ids must be UUIDs before reaching the path/branch laws; the containment walk rejects non-Normal components (lexical `starts_with` alone is defeated by `..`); new `contract_fixture_tests` consume `fixtures/contracts/workflow-definition/` directory-driven. The corpus immediately caught a real drift: the server defaults omitted arrays, the runtime required `edges` — `edges` now serde-defaults like `inputs`/`docTemplates`.
- **Ruling F (F7/F8/F12):** typed placement errors pick the status — `Mismatch` → 409, `RepoRootNotFound` → 400, infra → retry-safe 503 with **path-scrubbed details** (full error to logs). Malformed/non-JSON bodies get the same 400 ProblemDetails via `JsonRejection`. A failure after the worktree was cut but before rows exist **compensates the artifact** (workspace row + worktree + `workflow/<runId>` branch) so the retry is genuinely fresh — resolves F8 without persistence. T1: a committed `.proliferate` file poisons materialization → 503, zero rows, artifact gone, retry after repo fix succeeds.
- **Ruling M (F5/F9):** nullable projection fields are `required = true` in the OpenAPI schema (serde emits explicit `null`; the TS mirror declares `| null`, not optional); `arguments` is a real object schema; the dead `WorkflowRunPutRequest` collapsed into the actual parse target with `id` **required** (run-id fallback + stale comment deleted).
- **F11:** the PUT's reply reads **through the actor mailbox** (`ActorRequest::ReadProjection`) so the 201 body already carries the launched first node (`sessionId`/`promptId` non-null, asserted); `CreatedRun.created` decides 200-vs-201.
- **F13:** command routes pre-check via one cheap `node_membership` read (run-vs-node 404 discrimination preserved) instead of a full projection; the manager ensures actors on blocking threads (`spawn_blocking`).
- **F6:** route table moved to `workflow_runs::routes()`, command handlers to `workflow_run_commands.rs`, OpenAPI doc to `workflow_runs_openapi.rs` (merged at serve time, subagents pattern). `router.rs` 619 ≤ 632, `openapi.rs` 660 ≤ 687 — **no allowlist bumps**.
- **F10:** five new command-route tests (fail-redo incl. Ruling L reply shape, type-flip governing the advance, undo-advance/resume illegal+missing mapping, adhoc via route incl. ghost anchor) + the compensation T1.

Tests: 155 workflow-filtered + 133 `api::` + 194 `live::` green. Negative control: disabling `compensate_placement` fails T1 exactly at "worktree compensated"; restored, green. Pre-existing max-lines reds on the stack (store.rs, transition.rs, test files) unchanged per directive scope.

## Chain rebase (2026-08-14)

Rebased onto PR2 final head `00136eccd` (carries newer main incl. archiving R7). This rung's new head: `bfd3bf7a0`. Cross-base conflicts resolved per the supersession rules (gen-1 deletions win; archiving-era main wins); details in the lane journal.

## Post-rebase repair wave (2026-08-14)

Two cross-base test drifts from the replay onto archiving-era main are fixed in the PR1 rung (the store-test seed no longer inserts the dropped `workspaces.cleanup_state` column; the migration-registry order test now pins 0068 as the final workspace rebuild with only the workflow gen-2 rebuild after it), and the max-lines allowlist now tells the truth at every rung tip — deleted gen-1 rows removed, shrunk rows re-pinned to observed counts, and this chain's oversized gen-2 files carry exact-count rows with per-rung delta reasons. `check_max_lines.py` passes at each rung head. Verified with `cargo test -p anyharness-lib --lib`, `-p anyharness --bins`, and the collector_process suite. Chain now sits on PR2 final head `5b637b404` (id-space fix); this rung's head: `c4e5e6bb8` (CI-fix wave: boundaries baseline re-measured per rung; PR5a also recommits regenerated SDK artifacts).


pablonyx:
Landed on main via #1882's merge commit (f8f969243) — this branch's commits are all reachable from main. Closing as delivered.

PR #1886, opened 2026-08-14
fix(anyharness): stop forwarding the catalog default-model sentinel as ANTHROPIC_MODEL

pablonyx:
## Bug

`anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_env.rs` exports `ANTHROPIC_MODEL` from `requested_model_id`, filtering only empty/whitespace strings. But the agent catalog (`catalogs/agents/catalog.json`) defines a claude model entry `{"id": "default", "displayName": "Default 4.8", "description": "Use the default model (currently Opus 4.8)", ...}` that is a **sentinel** meaning "use the harness's own default" — not a real model name the claude CLI understands.

`ActiveCatalog::validate_launch_in_universe` (`anyharness/crates/anyharness-lib/src/domains/agents/catalog/service.rs`) resolves this sentinel row like any other catalog entry (it is `defaultVisible` and selectable), so a session created against the "default" option ends up with `requested_model_id = Some("default")`. Launch env then forwards `ANTHROPIC_MODEL=default` to the claude CLI, which fails the session with `model_not_found` ("There's an issue with the selected model (default)").

**Live-confirmed:** sessions with `requested_model_id=haiku` complete; `requested_model_id=default` fail 100%. Non-claude harnesses escape only because `launch_env.rs` gives them an empty env regardless of model id.

**Blast radius:** any claude session launched on the catalog's "default" model option. Workflows gen-2 makes this deterministic for nodes with no explicit model (they inherit the sentinel).

## Fix

Minimal, in the same filter chain that already strips empty/whitespace: also skip exporting `ANTHROPIC_MODEL` when the requested id equals the catalog's default sentinel (`"default"`), pulled into a named, documented constant (`CATALOG_DEFAULT_MODEL_SENTINEL`) rather than an inline literal, since no existing constant/helper for this sentinel exists elsewhere in the codebase (checked `domains/agents/model.rs`, `domains/agents/catalog/service.rs` and `schema.rs` — the sentinel is a plain catalog-declared string with no dedicated Rust representation). When the sentinel is present, `ANTHROPIC_MODEL` is omitted entirely so the CLI falls back to its own default — exactly the existing behavior for "no selection".

`catalog.json` itself is untouched, so the JS catalog validator (`scripts/agent-catalog/**`) is not implicated.

## Test

Added to `anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_env_tests.rs`:
- `build_session_launch_env_omits_model_for_claude_default_sentinel` — sentinel id `"default"` → no `ANTHROPIC_MODEL` key (this is the regression test; fails without the fix).
- `build_session_launch_env_sets_requested_model_for_real_claude_model_id` — real id `"haiku"` → exported.
- `build_session_launch_env_omits_model_for_blank_requested_model` — whitespace-only id → not exported (pre-existing behavior, now covered explicitly).

**Negative control:** reverted the one-line fix, reran `cargo test -p anyharness-lib --lib launch_env -j 4` — exactly the new sentinel test failed (`ANTHROPIC_MODEL: "default"` in the panic output), all other tests stayed green. Restored the fix; full targeted suite passes (11/11).

## Verification

```
cargo test -p anyharness-lib --lib launch_env -j 4
running 11 tests
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 1934 filtered out
```

### DIFF
--- a/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_env.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_env.rs
@@ -18,6 +18,15 @@ use std::collections::BTreeMap;
 
 use crate::domains::agents::model::{AgentKind, ResolvedAgent};
 
+/// Claude's catalog-declared sentinel model id (`catalogs/agents/catalog.json`,
+/// claude's `models[]`): "use the harness's own default", not a real model
+/// name the CLI understands. A session can carry this as its resolved
+/// `requested_model_id` (it is a legitimate, `defaultVisible` picker entry —
+/// [`ActiveCatalog::validate_launch_in_universe`] resolves it like any other
+/// row), so launch env must recognize and skip it rather than forward it as
+/// `ANTHROPIC_MODEL`, which the CLI rejects with `model_not_found`.
+const CATALOG_DEFAULT_MODEL_SENTINEL: &str = "default";
+
 pub(super) fn build_session_launch_env(
     resolved_agent: &ResolvedAgent,
     requested_model_id: Option<&str>,
@@ -52,6 +61,7 @@ fn build_claude_session_launch_env(
     if let Some(model_id) = requested_model_id
         .map(str::trim)
         .filter(|value| !value.is_empty())
+        .filter(|value| *value != CATALOG_DEFAULT_MODEL_SENTINEL)
     {
         env.insert("ANTHROPIC_MODEL".to_string(), model_id.to_string());
     }
--- a/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_env.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_env.rs
@@ -61,7 +61,7 @@ fn build_claude_session_launch_env(
     if let Some(model_id) = requested_model_id
         .map(str::trim)
         .filter(|value| !value.is_empty())
-        .filter(|value| *value != CATALOG_DEFAULT_MODEL_SENTINEL)
+        .filter(|value| !value.eq_ignore_ascii_case(CATALOG_DEFAULT_MODEL_SENTINEL))
     {
         env.insert("ANTHROPIC_MODEL".to_string(), model_id.to_string());
     }
--- a/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_env_tests.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_env_tests.rs
@@ -71,6 +71,36 @@ fn build_session_launch_env_sets_requested_model_for_claude() {
     );
 }
 
+/// The catalog's `"default"` model row (`catalogs/agents/catalog.json`,
+/// claude's `models[]`) is a sentinel meaning "use the harness's own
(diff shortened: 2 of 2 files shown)


pablonyx (commented):
Reviewed at frozen SHA `58fad6d953cdf2166d6e6306e14f2364a1fd3765`.

**Findings**

1. `anyharness/crates/anyharness-lib/src/domains/sessions/runtime/launch_env.rs:29` — `CATALOG_DEFAULT_MODEL_SENTINEL` is a new Rust string literal duplicating the catalog's `"default"` id (`catalogs/agents/catalog.json:140`). There's no shared constant to reference (confirmed no existing Rust representation in `domains/agents/model.rs` / `catalog/service.rs` / `schema.rs`, matching the PR body's own claim), so a literal is the only option today — not a defect, just a latent drift risk: if `catalog.json`'s sentinel id is ever renamed, this constant has to be updated by hand with no compiler-enforced link. Low severity; fine to accept as-is given the doc comment cites the catalog line.

2. Match is exact + case-insensitive (`eq_ignore_ascii_case`, launch_env.rs:64) against a lowercase-only literal; catalog does emit it as lowercase (`catalog.json:140`, `:2296`), so the comparison is correct today.

3. Other harness builders: `build_session_launch_env` (launch_env.rs:33-40) routes `Codex | OpenCode | Cursor | Grok` to `Ok(BTreeMap::new())` — no env at all regardless of `requested_model_id`, so none of them can leak the sentinel via this file. Confirmed cursor's catalog entry also declares an `"id": "default"` model (`catalog.json:2294-2296`, displayName "Auto"), but since cursor gets no launch env here, this PR's fix doesn't need to (and doesn't) touch it. No missed sibling leak in `launch_env.rs` itself.

4. Possible adjacent, unaddressed path: `anyharness/crates/anyharness-lib/src/live/sessions/actor/config/handle.rs:24` (`apply_requested_model_preference`) also reads `session.requested_model_id` directly and calls `try_apply_model_preference(conn, native_session_id, model_id, ...)` to push the model into the live ACP session (used for `setSessionModel`/live model switching, separate from the launch-time env var this PR fixes). If a session's `requested_model_id` is ever `"default"` at this call site, it looks like it would attempt to apply `"default"` as a real model preference the same way — worth confirming whether `try_apply_model_preference`/the ACP model-id resolution already filters/tolerates this sentinel, since it's not touched by this diff. If it's already handled elsewhere (or degrades to `ConfigApplyOutcome::NotApplied` + a warn log rather than a hard failure), this is a non-issue; if not, it's a second sentinel-leak surface worth a follow-up.

5. Test adequacy: `build_session_launch_env_omits_model_for_claude_default_sentinel` asserts `!env.contains_key("ANTHROPIC_MODEL")` for `requested_model_id = Some("default")` — this fails without the fix (confirmed via the PR's stated negative control, revert-and-rerun). The added case-insensitive and contains-guard tests (`launch_env_tests.rs`, second commit) are good regression coverage for the two ways this filter could regress (case drift, substring-match overreach). Coverage looks adequate for the stated fix.

No blocking issues found. Item 4 is the only thing I'd want a maintainer answer on before calling this fully closed, but it doesn't block this specific fix from merging.

PR #1892, opened 2026-08-15
fix(product): make attachments work in new chats from home (PRO-165)

tairabun:
## What

Makes the composer's **+ attach button, drag-and-drop, and paste-to-attach work on the home / new-chat screen**, carrying the files into the launched session's first prompt.

## Why

[PRO-165](https://linear.app/proliferate-team/issue/PRO-165/cant-add-files-to-new-chats): the home composer reuses the chat composer's control clusters but hard-stubbed the trailing cluster with `supportsAttachments={false} canAttachFiles={false} onAttachFile={() => {}}`, so the + button rendered permanently disabled; the file-drop handlers only exist on `ChatView`, which home never renders.

## How

- **Home owns a real attachment controller** (`useChatPromptAttachments`, scope `"home"`) with optimistic pre-session capabilities (`HOME_COMPOSER_PROMPT_CAPABILITIES`) — harnesses only report `PromptCapabilities` per session via liveConfig, which doesn't exist yet at home.
- **+ button** clicks a hidden file input (same accept list as chat, moved to `config/chat.ts`); **whole-screen drag-and-drop** with the same dashed overlay as chat; **paste-to-attach** via the shared `useChatInputPaste`. Draft chips render above the editor via `DraftAttachmentPreviewList`.
- **Snapshots ride the launch**: `attachmentSnapshots` thread through `useHomeNextLaunch` into every target branch (cowork, local, worktree, cloud, deferred cloud replay) and the launch-intent store (pending pane + retry echo the files). Delivery needs no new machinery — the intent outbox already accepts snapshots and rebuilds prompt blocks from them at dispatch (`preparePromptBlocks`), after the session materializes.
- Attachments clear on successful launch and survive alongside the restored draft on failure, mirroring chat's clear-on-success contract.

Any-file drops: home mirrors chat's drag-pasteboard *local-path* recovery for targets that run on this machine (cowork/local/worktree) — zips, PDFs, videos, folders, and oversized files attach as path references. Cloud and unresolved targets keep byte uploads only, and flipping the target across the local/remote boundary clears draft attachments so a stale local path never rides into a runtime that cannot read it.

## How to test

1. Open the app signed in, stay on the home screen ("What should we build…").
2. Click **+** → pick an image or text file → a chip appears above the editor.
3. Drag a file anywhere onto the home screen → dashed overlay shows, chip appears on drop. (Cmd+V a screenshot also attaches.)
4. Type a prompt, send. The pending launch pane echoes the file, and the created session's first message carries the attachment (visible in the transcript; agent can read it).
5. Failure path: pick an unconfigured cloud repo target → send fails → draft **and** chips are restored.


tairabun:
@greptile review

PR #1906, opened 2026-08-15
fix(workflows): add run cancel command + endpoint

pablonyx:
## QA finding

The gen-2 workflow-runs API had no way to cancel a run. A run wedged on a bad turn (hung tool call, runaway agent) was unrecoverable except by killing the whole process — there was no cancel/DELETE route anywhere on `/v1/workflow-runs`.

## What this adds

`POST /v1/workflow-runs/{run_id}/cancel`

**Transition legality matrix** (drops straight out of widening `is_terminal()` to include `Cancelled`, so the existing terminal-run gate in `on_command` enforces it automatically):

| From state | Cancel legal? | Session disposed? |
|---|---|---|
| `running` | Yes | Yes (node was live) |
| `awaiting_human` | Yes | No (pause state, no live turn) |
| `interrupted` | Yes | No (pause state, no live turn) |
| `completed` / `failed` / `cancelled` | No — 409 `WORKFLOW_TRANSITION_ILLEGAL` | — |

- Marks the run `cancelled` (stamps `completed_at`) and the current chain node `cancelled` (node's own `completed_at` stays NULL, mirroring `FailNode`/`InterruptNode` — that column specifically means "completed successfully"). Disposes the node's live session only when the node was actually `Running`, reusing Ruling L's precedent verbatim ("pause states hold no live turn to kill"). Adhoc rows are intentionally untouched — they self-resolve through their own turn reports regardless of run status, same assumption `FailNode`/`CompleteRun`/`Redo` already make.
- Everything lands in the existing single-commit `apply_transition` path, so the run/node status flip and session disposal are atomic.
- Idempotency: repeat cancel on an already-cancelled run is 409, consistent with every other command.

**Also fixed**: `WorkflowSessionControllerPolicy::controlling_run_id`'s terminal-status set only knew about `Completed`/`Failed`. Without widening it to include `Cancelled`, a cancelled run's workspace would stay stuck non-destroyable — the workspace-teardown gate would still see the (now-cancelled) run as "controlling" its session.

**Migration**: `0070_workflow_run_cancel_status` — a preserve-copy full-table rebuild (rename/recreate/copy-data, same shape as `0062`) widening the `workflow_runs` and `workflow_run_nodes` CHECK constraints to admit `'cancelled'`. Registered in `CUSTOM_FOREIGN_KEY_MIGRATIONS` immediately after `0069_workflow_runs_gen2`, since it widens the vocabulary `0069` first creates.

## Test evidence

`cargo test -p anyharness-lib workflow`:
```
test result: ok. 172 passed; 0 failed; 0 ignored; 0 measured; 1806 filtered out; finished in 23.00s
```

Coverage added:
- Pure transition-table tests: legal from each pausable state, disposes the session only when the node was `Running`, illegal on terminal runs.
- Store-level tests: single-commit persistence, repeat-cancel is `Illegal`, and a dedicated event-emission test confirming the widened `is_terminal()` still fires `anyharness.workflow_run_finished`.
- HTTP route test: happy path against a wedged blocking turn (`cancel_answers_with_the_cancelled_projection_and_disposes_the_running_session`), plus the 409 (repeat cancel) and 404 (ghost run) edges.
- Migration upgrade suite (`workflow_run_cancel_status_migration_tests.rs`, file-backed SQLite via the real migration ladder): existing rows survive the rebuild verbatim, the widened CHECK admits `'cancelled'` while still rejecting a bogus status, both indexes and FK enforcement survive, and the raw migration function is idempotent when re-applied directly.

## Scope note

Backend (Rust) only, per the task brief. The frontend TS SDK types (`anyharness/sdk/src/types/workflow-runs-v2.ts`, `anyharness/sdk-react/src/hooks/workflow-runs.ts`) are untouched and will need a follow-up to expose the new command client-side.

## Formatting note

`cargo fmt -p anyharness-lib -- --check` reports diffs in this crate, but they're pre-existing drift on the base branch (confirmed identical diffs reproduce on an unmodified `codex/workflows-gen2-pr7-builder-launch` checkout, e.g. in `store.rs`'s `ResolvedSideEffect` enum and `all_runs`/`runs_for_workspace`, and in the `.request(Method::PUT, ...)` idiom already used 5x in `workflow_run_command_route_tests.rs`) — likely a local rustfmt version mismatch against whatever produced the base branch. Not touched here to avoid dragging unrelated reformatting into this diff.

## Fix round 1 (opus review response)

Addressed all BLOCKER/HIGH/MEDIUM/LOW findings from the dedicated review:

- **BLOCKER**: migration 0070 now carries forward every 0069 inline column
  comment verbatim (the rebuild had dropped ~15, e.g. "the node row id, the
  API-addressable identity"); regenerated `specs/GENERATED/anyharness-db-schema.sql`
  via `update_anyharness_schema_snapshot`.
- **BLOCKER**: regenerated the SDK artifacts (`anyharness/sdk/generated/openapi.json`,
  `anyharness/sdk/src/generated/openapi.ts`) for the new route and widened
  status enums — `git diff --exit-code` on both is now clean.
- **BLOCKER**: bumped the repo-shape line-count allowlist to the final counts
  (`store.rs` 1377, `store_tests.rs` 1677, `transition.rs` 774,
  `transition_tests.rs` 1190, `custom_migrations.rs` 1326, plus a new
  `actor.rs` 603 entry this round's own DisposeSessions arm pushed over 600).
- **HIGH**: cancel now disposes EVERY running row's session (chain node +
  any concurrently running adhoc row), not just the current node's — a run
  cancelled while an adhoc row was `Running` used to leave that agent
  session alive forever with the workspace already released for
  destruction. Mirrors `on_boot_fence`'s "every running row" scan; the
  `Cancel` transition's side effect now carries `disposed_session_ids: Vec<String>`
  instead of `Option<String>`. New store + transition tests cover a running
  adhoc row; negative-controlled (reverted the fix, confirmed the new tests
  fail because the adhoc session is missing from disposal, then restored).
- **HIGH**: TS SDK companion — `"cancelled"` added to `WorkflowRunStatusV2`/
  `WorkflowRunNodeStatusV2`, a `cancel()` method added to
  `WorkflowRunsV2Client`, and a `"cancelled": "muted"` entry added to the
  total `NODE_STATUS_TONE` record in `run-view-model.ts` (matching
  `pending`'s terminal-inert convention, per the review's guidance not to
  invent a new tone). `@proliferate/product-client` build + typecheck both
  pass.
- **MEDIUM**: added direct test coverage for
  `WorkflowSessionControllerPolicy` (previously untested anywhere in the
  crate) across Running/Completed/Failed/Cancelled/unlinked-session.
- **MEDIUM**: cancel now clears `interruption_code`, matching
  `ResumeNode`/`Redo`; added a store test (cancel an Interrupted run →
  `interruption_code` is NULL).
- **LOW**: seeded a `workflow_run_docs` row and a `sessions` row in the 0070
  migration FK fixture so `pragma_foreign_key_check` actually proves
  something; fixed the stale "six workflow command POSTs" docstring to
  seven; parameterized `commands_other_than_redo_are_illegal_on_terminal_runs`
  over Completed/Failed/Cancelled; NULL the disposed chain node's own
  `session_id` on cancel, matching `UndoAdvance`'s convention.

### Test evidence (fix round 1)

`cargo test -p anyharness-lib workflow`:
```
test result: ok. 180 passed; 0 failed; 0 ignored; 0 measured; 1806 filtered out; finished in 22.71s
```

`cargo test -p anyharness-lib --lib schema_snapshot`:
```
test result: ok. 2 passed; 0 failed; 1 ignored; 0 measured; 1983 filtered out; finished in 0.09s
```

`cargo test -p anyharness-lib workflow_run_cancel_status_migration`:
```
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 1983 filtered out; finished in 0.52s
```

`python3 scripts/check_max_lines.py`: `Max-lines check passed`.

TS: `@anyharness/sdk` typecheck/build, `@anyharness/sdk-react` build, and
`@proliferate/product-client` build + typecheck all exit 0;
`run-view-model.test.ts` (53 tests) passes in full.

## Open rulings for Pablo

These need a founder call, not an implementation guess, so they're left
unresolved on purpose:

1. **Post-commit dispose fence race**: cancel is the first terminal
   transition that both commits terminal status AND disposes a session
   after that commit. Between the commit and `dispose_session` running, a
   concurrent destruction sweep could reap the workspace out from under a
   still-live session (narrow window, but new — every prior terminal
   transition either disposes nothing or stays non-terminal).
2. **Strict status parse**: `store.rs`'s status reads go through `parse_text`,
   which hard-errors on an unrecognized value — deliberately stricter than
   the adjacent `failure_code`/`interruption_code` reads, which degrade
   leniently. A binary predating migration 0070 that reads a `cancelled`
   row will hard-error the run read, which is the repo's known "strict enum
   reads break old binaries" hazard, realistic here given the shared
   prebuilt-runtime-bin dev setup and any rollback scenario.
3. **Cancel is undoable via adhoc `FailAndRedo`**: the terminal wall admits
   `FailAndRedo`, and an adhoc row in `Failed`/`NeedsAttention` still passes
   `legal_pause`, so the store will launch a fresh agent session inside a
   `Cancelled` run whose workspace policy has already released. Pre-existing
   shape for Completed/Failed runs, newly reachable through a user's
   explicit "stop".

## Lints re-expression (founder review)

Rebasing this PR onto `main` twice landed on top of two repo-shape schema changes it predates: first the plain line-count `scripts/*_allowlist.txt` files, then (after `main` moved another ~117 commits, including #1722 "chore(lints): docs-v1 D6 — bespoke checkers become lints/ records + generated diagnostics") those files were deleted outright in favor of per-site `lints/anyharness/*.toml` records. Every entry below **re-expresses debt this PR already carried in its pre-#1722 diff** (bumped allowlist counts / a calibration anchor already reviewed in "Fix round 1" above) — none is net-new scope. Flagging per CODEOWNERS since `lints/*.toml` requires founder sign-off.

| Old flat-allowlist row (file : rule : old value) | New record (file : rule ID : entry) |
|---|---|
| `scripts/anyharness_boundaries_allowlist.txt` — `DOMAIN_SQL_OUTSIDE_STORE anyharness/.../workflows/policy.rs 2 → 8` (this PR's `seed_run_with_node` test-mod INSERTs) | `lints/anyharness/exceptions.toml` — 6 new `[[exception]]` entries, `rule = "AH-STORE-3"`, `path = ".../workflows/policy.rs"`, sites: `fn seed_run_with_node::INSERT OR IGNORE INTO`, `fn seed_run_with_node::INSERT OR IGNORE INTO#2`, `fn seed_run_with_node::INSERT INTO`, `fn seed_run_with_node::INSERT INTO#2`, `fn seed_run_with_node::rusqlite::params`, `fn seed_run_with_node::rusqlite::params#2` |
| `scripts/test_check_anyharness_boundaries.py` — `test_known_truths_the_hardened_rules_must_see` calibration anchor `("DOMAIN_SQL_OUTSIDE_STORE", "domains/workflows/policy.rs", 34, ...)` | Same test, same convention (per-branch anchor pin), renamed rule id: `("AH-STORE-3", "domains/workflows/policy.rs", 35, "bare SELECT head, FROM on the next line")` — re-pinned 34→35 because this branch's `policy.rs` still has the `SELECT` on line 35, unchanged by either rebase |
| `scripts/max_lines_allowlist.txt` — `workflows/store.rs 1327 → 1377` | `lints/anyharness/ratchets.toml` — `[[max_lines]]` `path = ".../workflows/store.rs"`, `max_lines = 1377` (reason carries the same `fix/wf2-run-cancel` / fix-round-1 deltas already in "Fix round 1" above) |
| `scripts/max_lines_allowlist.txt` — `workflows/store_tests.rs 1462 → 1677` | `lints/anyharness/ratchets.toml` — `path = ".../workflows/store_tests.rs"`, `max_lines = 1677` |
| `scripts/max_lines_allowlist.txt` — `workflows/transition.rs 738 → 774` | `lints/anyharness/ratchets.toml` — `path = ".../workflows/transition.rs"`, `max_lines = 774` |
| `scripts/max_lines_allowlist.txt` — `workflows/transition_tests.rs 1103 → 1190` | `lints/anyharness/ratchets.toml` — `path = ".../workflows/transition_tests.rs"`, `max_lines = 1190` |
| `scripts/max_lines_allowlist.txt` — `persistence/custom_migrations.rs 1318 → 1326` (0070 migration registration, already noted in "Fix round 1"'s BLOCKER item above) | `lints/anyharness/ratchets.toml` — `path = ".../persistence/custom_migrations.rs"`, `max_lines = 1326` |
| `scripts/max_lines_allowlist.txt` — new `live/workflows/actor.rs` entry ("Fix round 1" above cites `603`; the file measures `606` lines on this head and CI's `check_max_lines.py` confirmed 606 against `wc -l`, so the record uses the CI-verified count) | `lints/anyharness/ratchets.toml` — new `[[max_lines]]` entry, `path = ".../live/workflows/actor.rs"`, `max_lines = 606` |

All seven records were verified locally: `python3 -m unittest scripts/test_check_anyharness_boundaries.py` (94 tests) and `python3 scripts/check_anyharness_boundaries.py` / `python3 scripts/check_max_lines.py` all pass at head `34fc9269b`, and every ratcheted count matches `wc -l` on disk exactly.


pablonyx:
## Agent review (opus, dispatched by wf2 QA session) — verdict: needs-changes

Three of the red CI jobs are caused by this PR, and there is one real behavioral gap. Fix round dispatched; nothing merges without Pablo.

### Blockers (all three CI reds are this diff, not pre-existing main redness)

1. **Schema snapshot of record not regenerated.** `specs/GENERATED/anyharness-db-schema.sql:663` still carries the 0069 CHECK without `'cancelled'`. Regenerate via `cargo test -p anyharness-lib update_anyharness_schema_snapshot -- --ignored`. Separately: the 0070 full-table rebuild **drops every inline column comment 0069 wrote** (~15+ of them, e.g. "the node row id, the API-addressable identity"). The snapshot is the canonical schema doc — 0070 should carry the 0069 comments forward verbatim, then regenerate.
2. **Generated SDK artifacts not regenerated** (`Build SDK` red). `openapi.json` / `openapi.ts` lack the new `/v1/workflow-runs/{run_id}/cancel` path and the widened status enums.
3. **Repo-shape line-count allowlist not bumped** for five files this PR grew (`store.rs` 1352/1327, `store_tests.rs` 1579/1462, `transition.rs` 769/738, `transition_tests.rs` 1160/1103, `custom_migrations.rs` 1326/1318).

### High

- **Cancel does not stop running ad-hoc sessions.** `transition.rs` disposes only `state.current_node()`, which by invariant is never adhoc. A run cancelled while an adhoc row is `Running` leaves that agent live and burning tokens with no remaining way to stop it — the run is terminal, and `policy.rs` now releases the workspace for destruction under it. The "adhoc self-resolves via its own turn report" doc-comment defense assumes a healthy turn; the QA finding that motivated this PR is precisely a wedged turn. `on_boot_fence` already fences **every** running row for this reason — cancel should collect all running rows' sessions. No test covers cancel-with-a-running-adhoc.
- **TS SDK gap becomes a compile break the moment blocker 2 is fixed.** `workflow-runs-v2.ts` hand-authors both unions without `"cancelled"` and the client has no `cancel()`. `NODE_STATUS_TONE` is deliberately total over the node-status union, so the regen forces the tone + union + client work into this PR as a required companion, not a follow-up.

### Medium

- `interruption_code` survives a cancel (cancelled row can still read as interrupted-for-a-reason; `ResumeNode`/`Redo` both clear it).
- The `policy.rs` drive-by is correct (verified at both admission call sites; only widens the `NOT IN` set) but has zero test coverage.
- **Needs a ruling (not auto-fixed):** cancel is the first terminal commit with a post-commit `dispose_session` — narrow fence race where a destruction sweep can reap the workspace under a still-live session between commit and dispose.
- **Needs a ruling (not auto-fixed):** status reads route through strict `parse_text` — a binary predating 0070 hard-errors reading a `cancelled` row (the repo's documented "strict enum reads break old binaries" hazard; adjacent `failure_code`/`interruption_code` reads are deliberately lenient).

### Low (cheap, in fix scope)

Vacuous FK migration test (fixture seeds no FK-referencing rows), stale "six workflow command POSTs" docstring, terminal-legality matrix test only exercises `Completed`, cancelled node keeps a dangling `session_id`, cancel-then-adhoc-redo can relaunch an agent inside a cancelled run whose workspace policy already released (pre-existing shape, newly reachable — flagged for Pablo).

### Verified correct

Legality matrix (cancel legal exactly on running/awaiting_human/interrupted; invariants updated coherently), double-dispose safety post-restart, redo-replacement targeting, API surface parity with the other six commands (409/404 codes, second-cancel → 409 matches repo convention), migration rebuild mechanics (0049/0061/0062 precedent followed exactly; FK bracket verified; indexes recreated; can't double-run), migration fixture CHECK-constraint seeding, and `workflow_run_finished` observability firing on cancel.

Literal evidence: scoped suite `cargo test -p anyharness-lib workflow` → `172 passed; 0 failed` (the scope word is what hid the break); `schema_snapshot` → `FAILED. 1 passed; 1 failed`, matching CI job 94964513154 (`1972 passed; 1 failed`).

pablonyx:
Rescued this PR after its base branch (`codex/workflows-gen2-pr7-builder-launch`) merged to `main`, which left it CONFLICTING. Fix-forward summary (no scope changes to the run-cancel feature itself):

- Rebased both commits (`24256b1ef` add run-cancel command + endpoint, `b7bd1a4f3` PR review fixes) onto `origin/main`. Patch identity confirmed clean via `git range-diff` — only conflict was in `scripts/max_lines_allowlist.txt` (line-count bookkeeping for 4 files also touched by unrelated merged work), resolved by taking main's post-merge counts plus this PR's deltas; the resulting counts matched `wc -l` on disk exactly.
- Retargeted the PR base to `main` and force-pushed the rebased branch.
- Repo-shape CI caught two things the rebase exposed, both fixed forward in a new commit (`9f598292b`):
  - `scripts/anyharness_boundaries_allowlist.txt`: `DOMAIN_SQL_OUTSIDE_STORE` count for `domains/workflows/policy.rs` needed widening from 2 to 8 — this PR added the policy's first direct test coverage (`seed_run_with_node`), whose inline `INSERT` fixture rows are engine-invisible cfg(test) SQL, same reasoning already applied to `access_gate.rs`'s row.
  - `scripts/test_check_anyharness_boundaries.py`'s per-branch calibration anchor for that same rule needed re-pinning from line 34 to line 35 (the file's own convention: "the anchor differs per rung if later rungs edit the file; each rung's branch pins its own value"). Confirmed this anchor was already stale on the original unrebased branch, not something the rebase introduced.
- Re-derived area labels from the post-rebase diff (`gh pr diff --name-only`) and added `area:sdk`, `area:product`, `area:docs` alongside the existing `area:anyharness` + `release:fix`.

Left as another author's PR — did not merge. Full `git range-diff` output and CI results are in the review agent's report.

pablonyx:
**Manual verification** (local, scoped, independently run at head e99efedef after the rebase onto main)

Scoped `cargo test -p anyharness-lib` per touched area:

```
cancel filter               → test result: ok. 53 passed; 0 failed
workflow_run_cancel_status  → test result: ok. 4 passed; 0 failed   (migration + registry)
workflow_run_command        → test result: ok. 6 passed; 0 failed   (route tests)
```

Scoped vitest (apps/packages/product-client, after `pnpm shared:build`):

```
src/domain/workflows/run-view-model.test.ts → Tests 53 passed (53)
```

**Negative control** — the rescue included one real fix commit (9f598292b, boundary allowlist/anchor re-pin). Reverted it and re-ran the checker:

```
git revert --no-commit 9f598292b
python3 scripts/check_anyharness_boundaries.py →
anyharness/crates/anyharness-lib/src/domains/workflows/policy.rs:73: [DOMAIN_SQL_OUTSIDE_STORE] ...
(+3 more policy.rs hits at :84 :88 :98)
```
then restored to e99efedef; checker passes again at head.

Hand-probe: cancel semantics verified through the test surface — cancel disposes every running row's session (dba814e67's review fix), the status migration backfills cleanly (4 migration tests), and the run-view model maps the new cancelled status through 53 view-model cases. Rebase patch-identity was proven by range-diff (see rescue comment above); all CI checks are green at this head with the validator-complete label set (release:fix + area:anyharness/sdk/product/docs).

pablonyx:
Round 2 rescue: main advanced ~117 commits since the last push (including #1722, which retired the flat `scripts/*_allowlist.txt` files in favor of per-site `lints/anyharness/*.toml` records), so #1906 went CONFLICTING again. Fix-forward summary (no scope changes to run-cancel itself; merged rather than rebased this time, per instruction, since the already-reviewed stack tolerates a harmless merge commit ahead of squash):

- `git merge origin/main` on top of the prior rescue head (`e99efedef`). Exactly the three predicted conflicts, all in repo-shape machinery:
  - `scripts/anyharness_boundaries_allowlist.txt` (modify/delete) — took main's deletion.
  - `scripts/max_lines_allowlist.txt` (modify/delete) — took main's deletion.
  - `scripts/test_check_anyharness_boundaries.py` (content) — took main's version as base, then re-applied the calibration-anchor fix on top (see below).
- Re-expressed the two deleted allowlists' debt as `lints/anyharness/exceptions.toml` / `lints/anyharness/ratchets.toml` records, matching the exact site-fingerprint / TOML formatting of neighboring entries (full old→new mapping is in the PR body's new "Lints re-expression (founder review)" section, since `lints/*.toml` is a CODEOWNERS founder-approval surface).
- Re-pinned `test_known_truths_the_hardened_rules_must_see`'s AH-STORE-3 anchor (main's #1722 renamed the rule id from `DOMAIN_SQL_OUTSIDE_STORE`) from line 34 to 35, same as before — confirmed via direct measurement that this branch's `policy.rs` SELECT is still on line 35.
- Ran the full local repo-shape suite (max_lines, migration_heads, anyharness_old_paths, server_old_paths, worker_structure, frontend_boundaries, mobile_product_client_export, appearance_scaling, design_attribution, toast_copy, report_frontend_structure --strict, component_library, anyharness_boundaries incl. its 94-test suite, session_mutation_admission, server_boundaries, workflow_managed_boundaries, docs) — all pass at head `34fc9269b`. No Rust source itself conflicted (all auto-merged cleanly), so no local cargo build was needed per the gate.
- CI: `Repo shape checks` passes at this head; `gh pr view --json mergeable` reports `MERGEABLE`.

Left as another author's PR — did not merge.

pablonyx:
**Manual verification — addendum.** The tallies above were taken at e99efedef, which #1722's lints restructure later made CONFLICTING (it deleted the two flat allowlist files this PR touches). Those entries are now re-expressed as lints/ records (see "Lints re-expression (founder review)" in the PR body). Pinning the settled state: head `34fc9269b276af57a96062df642a24458607aa54`, mergeable=MERGEABLE, full suite attached and settled at **23 pass / 0 fail / 0 pending** — including Cargo check & test, Shared frontend packages, Repo shape checks (validates the lints re-expression), Workflow definition lifecycle (tier-2), intent-tests + intent-billing (provisional), CodeQL, and Production compose smoke. No Rust source conflicted in the merge (allowlist/checker files only), so the scoped cargo results above (53+4+6 passed) and the negative control remain valid for this head.

pablonyx:
## Merge receipt — today-owned merge wave (2026-08-16)

**Squash-merged to main as `d1410b63c170a10f28c5732a834e90e3cb7bb2fc`** under Pablo's live merge ruling for tonight's owned PRs.

Same-chain pre-merge verification (literal):
- headRefOid re-verified unchanged: `34fc9269b276af57a96062df642a24458607aa54`
- mergeable=MERGEABLE (re-verified after main moved on #1997's merge)
- CI attached to that exact head, tally: `6 completed success` — 0 fail / 0 pending

## Written adjudication of the three parked items

**1. Lints re-expression table — verified LITERALLY against the diff: TABLE-EXACT.** Exactly 6 added AH-STORE-3 exceptions on `workflows/policy.rs` (the 6 named sites, nothing else), the 5 ratchet bumps + 1 new `actor.rs 606` block match `wc -l` at head exactly, and the calibration anchor 34→35 matches the on-disk SELECT head at `policy.rs:35`. This is carry-forward/re-expression of debt this PR already carried pre-#1722 (reviewed in fix round 1) — legal maintenance per `lints/README.md`, not a net-new amendment. One companion line outside the table's subject: the required `session_mutation_admission.txt` registry row for the new mutating route, consistent with its four workflow-command siblings.

**2. Post-commit dispose-race flag — real, genuinely new, NON-BLOCKING.** Cancel is the first transition that is simultaneously terminal and disposes post-commit (`actor.rs` step(): commit precedes `DisposeSessions`; prior terminal transitions dispose nothing; prior disposing transitions stay non-terminal). Worst case: the destruction sweep reaps the workspace while cancel is mid-disposing a session the user explicitly killed — an ungraceful teardown of an already-doomed session, no data-integrity consequence. Doesn't contradict the gen-2 workflows ADR (Ruling L reused verbatim). **Follow-up recorded**: hold the controller policy (or dispose before commit) to close the window.

**3. Strict-enum read flag — real, bounded, NON-BLOCKING.** Only reachable when a post-0070 binary writes a real `cancelled` row and a pre-0070 binary (rollback / stale shared prebuilt dev bin) then reads it. Blast radius is asymmetric: `runs_for_workspace()` and single-run reads degrade gracefully; the unfiltered `all_runs()` list bricks on one such row. Known repo skew-hazard class; forward-only migrations are doctrine. **Follow-up recorded**: make `all_runs()` per-row lenient like `runs_for_workspace()`.

(PR-body flag 3 — cancel undone via adhoc FailAndRedo — is a pre-existing shape for Completed/Failed runs, newly reachable via explicit stop; same follow-up bucket, not a blocker.)

PR #1907, opened 2026-08-15
fix(design): stop bumping placeholders — paint a 1px caret in empty fields (PRO-249)

pablonyx:
## Summary

- Fixes [PRO-249](https://linear.app/proliferate-team/issue/PRO-249/caret-bumps-placeholder-text-in-an-odd-way): the `2px` focused-placeholder `text-indent` from #1799 visibly bumped every native placeholder right on focus and back on blur (model popover search, agents-pane composer, command palette, all `PopoverSearchField` consumers).
- Replaces it with the chat composer's replacement-caret philosophy applied to native fields: while a field is focused **and** `:placeholder-shown`, suppress the native two-pixel WebKit caret (`caret-color: transparent`) and paint a one-pixel bar via `background-image` at the content-box origin (`background-origin: content-box`), blinking on the composer's cadence (`steps(1, end)`, infinite, `/* activity-motion */`-marked), static under `prefers-reduced-motion`. An empty field's caret provably sits at the text origin, so no measurement is needed; typing dismisses the placeholder, which atomically restores the native caret from the same rule (IME included). The placeholder never moves.
- Wraps the shared `caret-color` reset in `@layer base` per `specs/frontend/styling.md` (it was unlayered, which beat the suppression rule from inside the layer and silently strips caret utilities). The neighboring `::selection` reset has the same latent unlayered issue — left untouched as a follow-up.

## Pinning-test change (flagged for founder review)

`PopoverSearchField.test.ts` pinned the #1799 contract (`text-indent: 2px` on focus). That is exactly the behavior PRO-249 removes, so the pin is rewritten — not weakened — to the new contract: caret suppressed + 1px bar painted + **placeholder text-indent stays 0px on focus** (the PRO-249 regression pin), native caret restored on typing and blur, and blink animation present when motion is allowed. Assertions are deterministic via `reducedMotion` emulation.

## Trade-off

While a field is empty the caret is the painted 1px bar; the first typed character brings back the native 2px caret. The width change is masked by typing motion and matches the composer's 1px caret at rest.

## Testing

Tier 1, colocated:

- `pnpm --filter @proliferate/design build` — `check-theme: ok (308 live tokens)`
- `pnpm --filter @proliferate/product-client exec vitest run src/primitives/__tests__/PopoverSearchField.test.ts` — 2 passed
- Negative control: reverting the CSS makes both tests fail (`expected 'rgb(255, 255, 255)' to be 'rgba(0, 0, 0, 0)'`, `expected 'none' to be 'ui-empty-field-caret-blink'`); restoring goes green
- `python3 scripts/check_appearance_scaling.py`, `python3 scripts/check_design_attribution.py` — pass
- `pnpm --filter @proliferate/product-client typecheck` — exit 0
- `vitest run src/lib/domain/preferences/appearance-css-drift.test.ts` — 34 passed
- Visual proof in real Chrome (reduced motion): 1px bar at text origin, vertically centered in single-line inputs, first-line-aligned in textareas, placeholder unmoved

## Observability

None — shared presentation and regression coverage only.

### DIFF
--- a/apps/packages/design/src/css/product.css
+++ b/apps/packages/design/src/css/product.css
@@ -111,17 +111,60 @@ body {
   line-height: var(--text-ui--line-height);
 }
 
-:is(input, textarea, [contenteditable]:not([contenteditable="false"]), [data-ui-thin-caret]) {
-  caret-color: var(--color-text-caret);
+/* Layered per the shared-reset rule in specs/frontend/styling.md: unlayered
+   CSS beats every layer, which would both strip caret utilities off matching
+   elements and defeat the empty-field caret suppression below — that rule must
+   win this one by specificity inside the same layer. */
+@layer base {
+  :is(input, textarea, [contenteditable]:not([contenteditable="false"]), [data-ui-thin-caret]) {
+    caret-color: var(--color-text-caret);
+  }
 }
 
-/* Current macOS WebKit paints native insertion carets at two CSS pixels, while
-   native placeholders otherwise begin at that same origin. Offset only the
-   focused placeholder: entered text and its caret keep their established
-   alignment. The base layer lets a specialized control override this contract. */
+/* Current macOS WebKit paints native insertion carets at two CSS pixels, at the
+   same origin where a native placeholder's first glyph begins. Rather than
+   nudging the focused placeholder aside (which visibly bumped it on every
+   focus/blur while it stayed on screen), suppress the native caret only while
+   the placeholder is shown — an empty field's caret provably sits at the
+   content-box origin — and paint the same one-pixel replacement caret the chat
+   composer uses, as a background bar at that origin. Typing any character
+   dismisses the placeholder, which atomically restores the native caret in the
+   same style rule (IME composition included). Entered text, its caret, and the
+   resting placeholder never move. The base layer lets a specialized control
+   override this contract. */
 @layer base {
-  :is(input, textarea):focus::placeholder {
-    text-indent: 2px;
+  /* activity-motion */
+  :is(input, textarea):focus:placeholder-shown {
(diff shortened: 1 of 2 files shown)


pablonyx:
Dedicated agent review completed (adversarial pass over cascade/layer interactions, selector blast radius, unit support, discrete-animation semantics, and test determinism). Outcome:

**Confirmed and fixed in 3f7a11ea5** — on engines without the `lh` unit (Safari <16.4, Firefox <120), `background-size: 1px min(1lh, 1.25em)` was dropped per-declaration, so the gradient fell back to `background-size: auto` and flooded the entire content box with the caret color while the native caret stayed suppressed — strictly worse than the bump this PR removes. Base declarations are now `lh`-free (`1px 1.25em`); the line-box-exact refinements live inside `@supports (height: 1lh)`. Note the Chrome-based regression test cannot exercise the `lh`-less path; safety there rests on every base declaration using universally-supported syntax.

**Latent, documented instead of coded around** — a future utility overriding only `background-image` (invisible caret) or only `caret-color` (double caret) on a focused empty field, and centered/RTL fields (bar is pinned to the left content edge). No current field in the app hits either; the contract comment now warns overrides must replace the contract whole.

**Verified fine** — suppression wins the in-layer cascade (0,2,1 vs 0,2,0); no other rule in the repo touches input caret/background-image; odd input types (`checkbox`/`date`/`file` etc.) can't match `:placeholder-shown`; CodeMirror/Lexical surfaces unaffected; `steps(1, end)` discrete `background-image` blink empirically confirmed in Chromium, degrading to a static bar if an engine won't animate it; activity-motion marker satisfies both motion-authority checkers.

All checks re-run green after the fix: design build + check-theme, the 2 regression tests, appearance-scaling, attribution, typecheck.

PR #1909, opened 2026-08-15
fix(product): clamp subagent message tooltips to a bounded preview (PRO-266)

pablonyx:
Fixes [PRO-266](https://linear.app/proliferate-team/issue/PRO-266/text-for-subagent-messaging-can-be-quite-long).

## Problem

Hovering a subagent messaging chip in the chat feed (the "messaged" chip on a `send_message` tool call, or the "replied ‹agent›" chip on a subagent reply) rendered the **entire raw message text** in a tooltip. Subagent briefs are routinely thousands of characters, so the tooltip filled the screen (see issue screenshot).

Root cause is two stacked gaps:
1. The receipts pass the raw `message`/`text` payload as `exactMessage` straight through `AgentIdentityChip` into `Tooltip content` — no clamp anywhere on the way.
2. `Tooltip`'s multi-line branch bounds width (`maxWidth: min(22rem, …)`) but not height; its `overflow-hidden` was inert without a height bound, so long text wrapped into a column taller than the viewport.

## Fix (two layers, mirroring the `clampDescription` + `line-clamp` precedent in ToastBody)

- **Source clamp**: new `clampAgentMessagePreview` / `AGENT_MESSAGE_PREVIEW_MAX_CHARS = 500` in `lib/domain/delegated-work/presentation.ts` (character slice + `trimEnd()` + ellipsis), applied at both tooltip entry points: `AgentIdentityChip` and the no-identity fallback in `AgentMessageReceipt`. Covers both directions (outgoing `send_message`, incoming replies).
- **Primitive guarantee**: `Tooltip`'s multi-line branch now sets `maxHeight: min(18rem, calc(100vh - 1.5rem))`, mirroring its width bound, so no caller can fill the viewport.
- **Playground**: added a long-`exactMessage` `AgentIdentityChip` variant to the library playground (`/playground/library`) as a reproducible hover target.

The full message stays reachable by opening the subagent session, where it appears as the inbound prompt.

## Verification

- **Negative control**: the two new component tests were written and run against the unfixed components first — both failed (tooltip rendered the full 3,000-char payload), then passed after the fix.
- New tests: 2 unit tests for the clamp helper, 2 component tests (chip path incl. a pin on the primitive's `maxHeight`, and the fallback path).
- Full product-client suite: 903 files / 6,452 tests passed. Typecheck clean. `check_max_lines`, `check_component_library`, `check_frontend_boundaries` pass.
- **Live manual test** (vite dev renderer + Playwright, hovering the playground chip fed a 3,400-char message): tooltip text is exactly 500 chars ending in "…", rendered 352×171 px in a 900 px viewport, `maxHeight` applied — previously it rendered the full text and overflowed the viewport.

### DIFF
--- a/apps/packages/product-client/src/components/patterns/AgentIdentityChip.tsx
+++ b/apps/packages/product-client/src/components/patterns/AgentIdentityChip.tsx
@@ -2,6 +2,7 @@ import { Button } from "#product/primitives/Button";
 import { Tooltip } from "#product/primitives/Tooltip";
 import { AgentIdentityGlyph } from "#product/components/patterns/AgentIdentityGlyph";
 import type { DelegatedAgentIdentity } from "#product/lib/domain/delegated-work/model";
+import { clampAgentMessagePreview } from "#product/lib/domain/delegated-work/presentation";
 
 /**
  * Durable agent-identity chip: the Solid Seal glyph and the agent's title in a
@@ -26,7 +27,9 @@ export function AgentIdentityChip({
     return null;
   }
 
-  const tooltip = exactMessage?.length ? exactMessage : identity.displayName;
+  const tooltip = exactMessage?.length
+    ? clampAgentMessagePreview(exactMessage)
+    : identity.displayName;
   const content = (
     <>
       <AgentIdentityGlyph
--- a/apps/packages/product-client/src/components/playground/library/product-patterns.tsx
+++ b/apps/packages/product-client/src/components/playground/library/product-patterns.tsx
@@ -19,11 +19,22 @@ const DEMO_AGENT_IDENTITY = buildDelegatedAgentIdentity({
   sessionLinkId: "library-agent-link",
 });
 
+// Hover target for the clamped message preview: far past the preview ceiling,
+// the way a real subagent brief is.
+const DEMO_LONG_MESSAGE =
+  "Investigate the retry behavior end to end, trace the actual mechanism until you can explain why it occurs, and report back with a mental model plus concrete fix options. "
+    .repeat(20);
+
 function AgentIdentityChipDemo() {
   return (
     <div className="flex flex-col items-start gap-2">
       <AgentIdentityChip identity={DEMO_AGENT_IDENTITY} onOpen={noop} />
       <AgentIdentityChip identity={DEMO_AGENT_IDENTITY} closed />
+      <AgentIdentityChip
+        identity={DEMO_AGENT_IDENTITY}
+        exactMessage={DEMO_LONG_MESSAGE}
+        onOpen={noop}
+      />
     </div>
   );
 }
--- a/apps/packages/product-client/src/components/workspace/chat/transcript/AgentMessageReceipt.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/transcript/AgentMessageReceipt.test.tsx
@@ -87,13 +87,9 @@ describe("AgentMessageReceipt", () => {
 
     fireEvent.focus(document.querySelector("[data-agent-identity-chip]")!);
     await waitFor(() => {
-      const tooltip = screen.getByRole("tooltip");
-      const preview = tooltip.textContent ?? "";
+      const preview = screen.getByRole("tooltip").textContent ?? "";
       expect(preview.length).toBeLessThanOrEqual(AGENT_MESSAGE_PREVIEW_MAX_CHARS);
       expect(preview.endsWith("…")).toBe(true);
-      // The primitive itself must bound height so no caller can fill the
-      // viewport with an unclamped payload.
-      expect(tooltip.style.maxHeight).toBe("min(18rem, calc(100vh - 1.5rem))");
     });
   });
 
--- a/apps/packages/product-client/src/components/workspace/chat/transcript/AgentMessageReceipt.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/transcript/AgentMessageReceipt.test.tsx
@@ -4,6 +4,7 @@ import { cleanup, fireEvent, render, screen, waitFor } from "@testing-library/re
 import { afterEach, describe, expect, it } from "vitest";
 import { AgentMessageReceipt } from "#product/components/workspace/chat/transcript/AgentMessageReceipt";
 import { buildDelegatedAgentIdentity } from "#product/lib/domain/delegated-work/identity";
+import { AGENT_MESSAGE_PREVIEW_MAX_CHARS } from "#product/lib/domain/delegated-work/presentation";
 
 const IDENTITY = buildDelegatedAgentIdentity({
   id: "agent-session-1",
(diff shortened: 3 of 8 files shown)


pablonyx:
## Codex review

(Findings produced by `codex exec` (gpt-5.6-terra) reviewing the PR head in read-only sandbox; posted verbatim by the orchestrating agent.)

1. **should-fix** — `apps/packages/product-client/src/lib/domain/delegated-work/presentation.ts:204`
   `slice()` counts UTF-16 code units and can split a surrogate pair. For example, 498 ASCII characters followed by `😀` yields a preview ending in a lone high surrogate plus `…`, rendered as `�…`. Clamp by Unicode code points/graphemes (and add a boundary test).

2. **should-fix** — `apps/packages/product-client/src/components/workspace/chat/transcript/AgentMessageReceipt.tsx:38`
   When identity resolution fails, the focusable fallback is the only disclosure mechanism for `exactMessage`; this change replaces its complete keyboard/screen-reader disclosure with a 500-character preview and offers no "show full message" path. `onOpen` is also unused in this branch. Preserve a focusable full-content disclosure or provide an explicit expandable/details control.

3. **should-fix** — `apps/packages/product-client/src/primitives/Tooltip.tsx:160`
   The shared default multi-line tooltip now silently clips any content beyond 18rem (`overflow-hidden`) without a continuation affordance or an opt-out. This affects all default-branch callers, including dynamic `agent.title` in `PendingAgentUpdatesRow`, `targetAppearance.displayName` in `SidebarWorkspaceVariantIcon`, sidebar indicator labels, and composer/session-control descriptions. Either scope the height ceiling to bounded-preview use cases or make overflow accessible/explicit; add a primitive-level overflow test.

Checked, fine:
- Newlines remain structurally preserved: `content.split("\n")` renders one block span per line; `textContent` naturally omits the separators.
- Other default multi-line Tooltip callers are HunkActionPill, WorkspacesPage, GoalBar, ComposerFastModeToggle, ComposerModeBadge, ComposerEffortStepper, SessionConfigControls, PendingPromptList, ProductSidebarRepositories, SidebarIndicators, SidebarWorkspaceGitGlyph, SidebarWorkspaceVariantIcon, PendingAgentUpdatesRow, and the primitive playground; the identity-receipts prototype explicitly uses `singleLine`.

pablonyx:
## Opus review

Reviewed the diff plus every `exactMessage` sink, every non-`singleLine` `Tooltip` caller, and the `clampDescription` precedent. The two-layer approach (source clamp + primitive backstop) is sound and the coverage is complete. Four findings, none blocking.

### Findings

**1. (nit) `clampAgentMessagePreview` can bisect an astral character** — `presentation.ts:201-205`
`text.slice(0, MAX - 1)` operates on UTF-16 code units, so when the cut boundary falls between the two halves of a surrogate pair (an emoji, some CJK ext chars) the result ends in a lone high surrogate and renders as `�` before the `…`. Purely cosmetic and this exactly mirrors the accepted `clampDescription` in `ToastBody.tsx:68`, so it is a nit, not a regression. If you want it airtight, slice on `Array.from(text)` / `Intl.Segmenter`.

**2. (should-fix) The primitive-invariant pin lives in the wrong test and is brittle** — `AgentMessageReceipt.test.tsx` (`expect(tooltip.style.maxHeight).toBe("min(18rem, calc(100vh - 1.5rem))")`)
The height ceiling is a `Tooltip` *primitive* guarantee, but the only assertion on it is embedded in a receipt component test, while `primitives/__tests__/Tooltip.test.tsx` pins neither `maxHeight` nor `maxWidth`. Consequences: (a) changing the bound to e.g. `16rem` breaks an unrelated receipt test rather than a primitive test, and (b) the primitive has no direct guard, so a caller could delete the `maxHeight` line and this receipt test would be the only thing that fails. Move the exact-string pin into `Tooltip.test.tsx`; keep the receipt test behavioral (assert the preview is clamped, not the primitive style string).

**3. (nit) Test claims a word-boundary clamp the code does not implement** — `presentation.test.ts` ("...with an ellipsis near a word boundary", `expect(clamped.at(-2)).not.toBe(" ")`)
The implementation is a hard code-unit slice; `trimEnd()` only strips a trailing space *when the cut happens to land right after one*, otherwise it cuts mid-word (`"…retryin…"`). The `at(-2) !== " "` assertion is satisfied trivially by `trimEnd()` for every input and exercises no boundary logic. Rename/reword so it does not imply smarter truncation than exists.

**4. (nit) Byte-identical duplicate of `clampDescription`** — `presentation.ts:201` vs `ToastBody.tsx:68`
Same `slice + trimEnd + ellipsis` body, differing only in the constant. Both are pure presentation string ops; a shared `clampWithEllipsis(text, max)` would collapse them. Low priority given they sit in different layers (primitives/toast vs domain).

### Checked, fine

- **Blast radius of the `maxHeight`**: enumerated every non-`singleLine` `Tooltip` caller — `SidebarIndicators`, `HunkActionPill`, `PendingPromptList` (3), `PendingAgentUpdatesRow`, `WorkspacesPage`, `GoalBar`, and the `AgentIdentityChip` identity-name path. All pass short static labels far under 18rem; the ceiling never bites them. `overflow-hidden` clips-without-scroll is acceptable because the only long-content path is now clamped upstream.
- **Coverage is complete**: all four `exactMessage` sinks — `AgentOperationsToolActionRow` (`send_message`), `AgentOriginPromptReceipt` (via `TranscriptItemBlock` / `TranscriptPendingPromptRow`), the `AgentMessageReceipt` no-identity fallback, and `AgentIdentityChip` — route through the two clamped entry points. `AgentOriginPromptReceipt` is covered transitively through `AgentMessageReceipt`, so no unclamped path remains.
- **`singleLine` tooltips unaffected**: the `maxHeight` is only on the multi-line branch and the clamp only on message tooltips.
- **`\n`-split interaction**: clamp runs before the split, newlines are preserved, and the primitive height bound is the correct backstop for a clamped-but-many-newlines payload.
- **Placement/doctrine**: `Tooltip.tsx` (the formatting wrapper over `tooltip-primitive`) is the right home for the height bound per the DESIGN_SYSTEM collision-pair split; the domain helper in `delegated-work/presentation.ts` is consistent with sibling `clamp*` helpers.

pablonyx:
## Review dispositions (fixes pushed in c994f7575)

**Applied:**
- **Codex #1 / Opus #1 — surrogate-pair split in the clamp**: fixed. The cut now drops an orphaned high surrogate before appending the ellipsis, with a boundary test placing an emoji exactly across the cut index (asserts `isWellFormed()`).
- **Opus #2 — primitive pin in a component test**: fixed. The `maxHeight`/`maxWidth`/`overflow-hidden` invariant is now pinned in `primitives/__tests__/Tooltip.test.tsx` ("bounds multi-line content on both axes"); the receipt test is back to behavioral assertions only.
- **Opus #3 — test name oversold word-boundary behavior**: renamed to describe what it asserts (ceiling + ellipsis + whitespace trim).

**Declined, with reasons:**
- **Codex #2 — fallback path loses full-content disclosure**: the unbounded tooltip-as-disclosure is precisely the defect this PR fixes, not a feature to preserve; a multi-thousand-character tooltip is no better for keyboard/screen-reader users than for pointer users. The full text remains reachable in the subagent's own transcript (it is the inbound prompt there). A proper in-place "show full message" affordance (details-panel extension) was proposed and deliberately deferred by Pablo when scoping this PR. `onOpen` being unused in the no-identity branch predates this PR — there is no session to open there.
- **Codex #3 — primitive maxHeight needs scoping/opt-out**: declined. Both reviews independently swept every non-`singleLine` caller and found only short static labels nowhere near 18rem, so the ceiling engages only on pathological content — which is exactly the backstop it exists to be. An opt-out would recreate the vulnerability the ceiling closes. The requested primitive-level overflow test was added (see Opus #2 above).
- **Opus #4 — dedupe with ToastBody's `clampDescription`**: deferred. Two private call sites in different layers (primitives/utils vs lib/domain) with different constants; extracting a shared `clampWithEllipsis` is a fine follow-up but not worth coupling the layers in this fix.

Verification after fixes: 4 affected test files (45 tests) pass; `check_max_lines` / `check_component_library` / `check_frontend_boundaries` pass; package typecheck clean.

PR #1923, opened 2026-08-15
feat(product): abort superseded workspace-selection fetches per selection nonce (UX-latency R9)

pablonyx:
## What

Rung 9 of the UX Latency + Transitions ADR ladder (ADR §4.6, Rung 9 / Q11): a newer workspace selection now **aborts the prior selection's in-flight requests on the wire**, instead of merely discarding their resolved results after the fact.

- The selection abort lifecycle now lives in `session-selection-store`. A new `workspaceSelectionAbort: AbortController` is the single source of truth for selection staleness. Every mutation that bumps `workspaceSelectionNonce` (`activateWorkspace`, `activateHotWorkspace`, `enterPendingWorkspaceShell`, `deselectWorkspacePreservingSessions`, `clearSelection`) aborts the outgoing controller with reason `workspace_selection_superseded` and installs a fresh one, so the abort and the nonce advance together as one mechanism.
- `runWorkspaceSelection` captures that signal (`currentWorkspaceSelectionSignal()`) the instant it takes ownership and threads it onto `WorkspaceSelectionContext.abortSignal`, then into `bootstrapWorkspace(... signal)`.
- `bootstrapWorkspace` attaches the signal to the session-directory fetch request options; the fetch layer (`use-workspace-bootstrap-cache`) now **composes** the selection signal with its own 8s timeout controller via `AbortSignal.any([...])` rather than letting a caller signal replace the timeout. The 8s ceiling (Q7) is preserved and neither signal masks the other.
- The existing `isWorkspaceSelectionCurrent` truthfulness guards are kept as defense-in-depth (unchanged), per scope.

The chain stays serial (no parallelization; Rung 10 owns Q12). No treatment/token changes.

## Why

Superseded selections previously kept their network requests running to completion and only dropped the result at the next `isWorkspaceSelectionCurrent` re-check. Under fast A→B→C workspace switching that wastes the runtime round-trips that dominate switch latency and can contend with the winning selection's fetches. Cancelling on the wire is the ADR's Rung 9 remedy.

## Epoch fold (Q11): attempted; residual raised as a founder question

Scope asked to fold the per-workspace `sessionActivationIntentEpochByWorkspace` "second staleness counter" into the same mechanism. I attempted a true fold and found it is **not** a redundant second copy of selection staleness:

- The epoch is consumed only by `session-activation-guard` (+ two tab read sites), which tracks **session-activation intent within a stable workspace selection**, a finer granularity than workspace selection. `beginSessionActivationIntent` / `invalidateSessionActivationIntent` bump it **without** changing `workspaceSelectionNonce`, so two session intents inside one selection supersede each other via the epoch alone (the guard's `intent-replaced` stale reason).
- Its one overlap with selection is that `activateHotWorkspace` bumps both the nonce and the epoch. That epoch bump is pinned by `session-selection-store.test.ts` ("bumps hot workspace intent only for hot activation").

Collapsing the epoch into the selection abort/nonce would erase the finer session-intent scope and require deleting/altering a pinning test, which the constitution forbids without founder review. Per the frozen scope's escape hatch, I made the selection AbortController the single source of truth for **selection-pipeline** staleness (the scope of "which selection wins") and left the epoch as the dedicated **session-activation-intent** scope. Open question below.

## Testing

Targeted vitest (suite derived from `git diff --stat` vs origin/main):

```
 ✓ src/hooks/workspaces/workflows/selection/connection.test.ts (3 tests)
 ✓ src/stores/sessions/session-selection-store.test.ts (11 tests)
 ✓ src/hooks/workspaces/workflows/selection/clear-runtime-state.test.ts (3 tests)
 ✓ src/hooks/workspaces/workflows/selection/run-hot-workspace-reopen.test.ts (6 tests)
 ✓ src/hooks/workspaces/workflows/selection/run-workspace-selection.test.ts (13 tests)
 ✓ src/hooks/access/anyharness/workspaces/use-workspace-bootstrap-cache.test.ts (7 tests)

 Test Files  6 passed (6)
      Tests  43 passed (43)
```

New GATE test (ADR §5 T2 fixture) `aborts a superseded selection's in-flight bootstrap fetch on the wire (A→B→C staggered)`: selects A, then B, then C with A's bootstrap still in flight, and asserts A's (and B's) bootstrap request signal **fired** with reason `workspace_selection_superseded`, C's signal is live, the staggered A result is **discarded** (selection converges on C, `workspaceSelectionNonce === 3`), and the superseded A/B latency flows are cancelled. Plus a store-level unit test that each nonce bump aborts the prior controller and mints a live one.

`pnpm run typecheck` (product-client): exit 0, 0 errors.

### Negative control

Removed the `supersedeWorkspaceSelectionAbort` call from `activateWorkspace`, re-ran the GATE test: it failed at `expect(capturedSignals["workspace-1"].aborted).toBe(true)` with `expected false to be true`; the superseded fetch was not aborted. Restored the line; suite green again.

## Test-infra note (please scrutinize)

Two pre-existing gaps blocked running any product-client test in this review worktree (a fresh `origin/main` checkout without built package dist). Both also reproduce on an untouched test file:

1. `vitest.config.ts` aliases the `@proliferate/product-client/host/*` public specifier to source but not the `internal/*` specifier that injected Desktop measurement modules import (`exports.default` points at `dist`). I added the mirrored `internal/*` → `src/` alias, consistent with this config's documented "tests run against source, never dist" rule. No lint weakened, no pinning test touched.
2. `@proliferate/cloud-sdk` / `@proliferate/cloud-sdk-react` needed their dist built (`pnpm --filter ... build`), which CI does; those are not committed here.

## Open founder questions

- Confirm the epoch decision above: keep `sessionActivationIntentEpochByWorkspace` as the finer session-activation-intent scope (this PR), or take a follow-up to unify both scopes under an AbortController-based mechanism (which would touch `session-activation-guard` and its pinning test, and is out of Rung 9's serial/no-behavior-change budget)?
- Connection-resolution fetches (`getCloudWorkspace`, gateway/SSH resolution) do not yet accept an `AbortSignal`; the signal is threaded through `WorkspaceSelectionContext` and the bootstrap/session-directory fetch (the documented GATE), but wire-aborting the cloud-sdk connection calls needs `AbortSignal` params added to those clients; proposed as a follow-up rather than widening this rung.


pablonyx:
## Adversarial review outcome (round 1)

A dedicated fresh-context reviewer read the full diff. Verdict: substantially CLEAN, no blockers, no gating should-fixes.

- Abort lifecycle: all five nonce-bump sites pair with supersede, no supersede without a bump, and the signal capture is in the same synchronous block as the nonce mint (no TOCTOU across awaits).
- Q7 ceiling: AbortSignal.any composition plus the independent timeout race preserves the 8s failure ceiling; a superseded abort classifies as stale (silent), never as a user-facing error on the new selection.
- Epoch deviation ACCEPTED: the reviewer independently verified begin/invalidateSessionActivationIntent fire in live paths (use-chat-tab-activation, session-shell-selection, use-workspace-shell-activation, workspace-shell-intent-writer, use-dismissed-session-cleanup) without nonce bumps, so the finer session-intent scope is real and the fold residual is correctly escalated as a founder question rather than forced.
- GATE test judged T2-real (drives the actual runWorkspaceSelection and store; only leaf resolvers mocked).
- vitest internal/* alias and the line-ratchet destructure judged clean; no lint or pinning test weakened.

Nits addressed or recorded:
- Em dashes scrubbed from the PR body (edited in place).
- Test-tightening suggestions recorded as non-blocking follow-ups: assert B's abort reason explicitly, assert no late repaint directly (not just final convergence).

pablonyx:
Final CI verification at the current head, literal `gh pr checks 1923` output:

```
Analyze (python)	pass
Analyze (rust)	pass
Build SDK	pass
CI/CD config checks	pass
Candidate build handoff	pass
Cargo check & test	pass
CodeQL	pass
Desktop frontend build	pass
Mobile typecheck	pass
Production compose smoke	pass
Repo shape checks	pass
Shared frontend packages	pass
Vercel	pass
Vercel Preview Comments	pass
Web frontend build	pass
```

PR #1924, opened 2026-08-15
feat(product): shared reconnect backoff policy for session and agents-pane streams

pablonyx:
## What

Extracts one shared, pure reconnect-backoff policy module (`apps/packages/product-client/src/lib/domain/sessions/stream/reconnect-backoff-policy.ts`) and migrates both reconnect consumers onto it, deleting their independent curves:

- Primary session stream: `session-reconnect-state.ts` (previously computed `baseDelayMs * 2 ** attempt` capped at 15s inline) now delegates the curve to the shared policy; its own per-session timer/map bookkeeping is unchanged.
- Agents-pane detail stream: `use-agents-pane-session-lifecycle.ts` (previously its own inline 750ms base / 6s cap curve) now calls the same `advanceReconnectBackoff` entry point.

The shared policy exposes a `ReconnectBackoffState` shape (`attempt`, `nextDelayMs`, `reconnecting`) that both consumers now surface: `session-reconnect-state.ts` gained `isSessionReconnecting`, and the Agents-pane hook's return type gained `reconnectState`. No visual treatment changes in this rung. Rungs 3/4 own the "reconnecting" affordance UI; this rung only wires consistent state underneath it.

Jitter is deterministic-testable: `reconnectBackoffDelayWithJitter`/`advanceReconnectBackoff` take an injectable `rng` (defaults to `Math.random`); tests pin it to `() => 0` to assert the bare curve.

Placement: `lib/domain/sessions/stream/` (not a workflow/session-specific folder) because this is pure client-side domain logic for stream reconnection shared by two independent stream owners. The frontend layering rules place cross-feature pure logic used by multiple hooks in `lib/domain`, and that folder already holds the sibling `stream-state.ts`/`stream-side-effect-plan.ts` modules for the same connection lifecycle.

Per ADR §4.7, Rung 7 (Q9): "one shared reconnect backoff policy (base 350ms, factor 2, cap 15s, jitter), both session-stream and agents-pane consume it."

## Behavior change (founder-ruled)

Agents-pane's own reconnect now uses **base 350ms / cap 15s** instead of its previous **base 750ms / cap 6s**. This means the pane retries sooner on the first few attempts and can wait longer (up to 15s, vs 6s previously) before a persistently-failing child's retry loop stabilizes. This is the explicit Q9 ruling, not an oversight.

Connection bounds (`session-stream-connection-open.ts`'s 2.5s open timeout, 30s max stream-sample duration) are untouched. Those are connection bounds, not backoff, and do not duplicate the policy.

## Round 2 review fixes

- **Preserved the gap-reconcile immediate reconnect fast path.** `reconcileGapAndReconnect` in `session-stream-flush-controller.ts` previously called `scheduleReconnect(0)` to bypass backoff after a gap-reconcile forced close. The initial migration made `scheduleReconnect` zero-arg, so JS silently dropped the `0`, folding the forced reconnect onto the shared error-retry curve (waiting >=350ms and advancing the per-session attempt counter, inflating later genuine retries). Per Q9 a gap-reconcile forced reconnect is not an error retry, so it now uses an explicit `scheduleImmediateReconnect()` API that fires on the next tick and does NOT advance the shared attempt counter. The old `scheduleReconnect: (delayMs?: number)` signature in `session-stream-flush-types.ts` is gone (now `scheduleReconnect: () => void` plus `scheduleImmediateReconnect: () => void`), so `tsc` catches future misuse.
- **Reset pane `reconnectState` unconditionally on release.** `releasePaneStream` in `use-agents-pane-session-lifecycle.ts` gated the `reconnectState` reset behind `isCurrent(sessionId)`, which is exactly false on a session switch (`isPaneRouteActive` flips before `clientSessionId` changes), leaving stale "reconnecting" state visible. The state is per-hook-instance, so it now resets unconditionally.

## Testing

```
cd apps/packages/product-client
npx vitest run \
  src/lib/domain/sessions/stream/reconnect-backoff-policy.test.ts \
  src/hooks/sessions/lifecycle/session-stream-connection-reconnect.test.ts \
  src/hooks/sessions/lifecycle/use-session-stream-flush.test.ts \
  src/hooks/agents/lifecycle/use-agents-pane-session-lifecycle.test.tsx \
  src/components/workspace/delegated-work/agents-pane/AgentsPaneDetail.test.tsx

 Test Files  5 passed (5)
      Tests  40 passed (40)
```

T1 gate (`reconnect-backoff-policy.test.ts`): drives the policy through an attempt sequence and asserts, with jitter neutralized (injected `() => 0` rng), the exact curve `350, 700, 1400, 2800, 5600, 11200, 15000, 15000...`, verified both through `session-reconnect-state.ts`'s `nextSessionReconnectDelayMs` (pinning `Math.random`) and through the raw `advanceReconnectBackoff` entry point the Agents-pane hook calls.

New round-2 coverage:

- `session-stream-connection-reconnect.test.ts`: `immediate: true` schedules a 0ms reconnect and leaves `currentSessionReconnectAttempt` at 0; negative control confirms the ordinary path advances the counter to 1 and waits on the curve (>=350ms).
- `use-session-stream-flush.test.ts`: gap reconcile calls `scheduleImmediateReconnect` (not `scheduleReconnect`); negative control (swapping the call back) fails as expected.
- `use-agents-pane-session-lifecycle.test.tsx`: a session switch while "reconnecting" resets `reconnectState` to `{attempt: 0, nextDelayMs: 0, reconnecting: false}`; negative control (restoring the `isCurrent` guard) fails as expected.

`tsc --noEmit -p tsconfig.json` passes clean.

## Open founder questions

- **Label classification.** Title is `feat(product): ...` while the label is `release:fix`. This rung is a behavior-changing consolidation (the agents-pane curve moves from 750ms/6s to 350ms/15s), which reads as a feature-ish change even though it is mechanically a consolidation/correctness fix. Flagging for a human call on whether `release:fix` or a feature label is wanted. Left as-is (`release:fix`) pending your call.


pablonyx:
Dedicated adversarial review (fresh-context agent, full diff) round 1: NOT CLEAN.

1. BLOCKER: silent immediate-reconnect regression. session-stream-flush-controller.ts:185 (reconcileGapAndReconnect) calls scheduleReconnect(0) as a deliberate bypass-backoff fast path after a gap-reconcile forced close. The migration made scheduleReconnect zero-arg (always defers to the shared curve), JS silently drops the 0, and session-stream-flush-types.ts:60 still declares (delayMs?: number) => void so tsc cannot catch it. Gap reconciles now wait >= 350ms (up to 15s if the attempt counter is advanced) AND advance the shared per-session counter, inflating later genuine retries. No test covers this path. Restore an immediate/attempt-0 path for gap reconcile (or explicitly rule it onto the curve with docs + type fix), add a test either way.
2. should-fix: releasePaneStream's isCurrent(sessionId) guard skips the reconnectState reset exactly on session switch (pane-route flag flips before clientSessionId changes), leaving stale reconnecting state until the new session connects. Reset unconditionally; state is per-hook-instance now.
3. nit: PR body has 6 em dashes; house rule bans them. Rewrite.
4. nit: title says feat while the label is release:fix; flag for a human call which classification Pablo wants (behavior-changing consolidation).

Verified clean: exact curve 350/700/1400/.../15000 with a real cross-consumer T1 gate; no pre-delay doubling; jitter can never exceed the cap; open-timeout and max-duration untouched; layering legal (pure lib/domain module consumed downward); labels present; agents-pane tests assert real behavior.

Fix round incoming; review will re-run until clean.

pablonyx:
Round-1 review fixes pushed (`bf08c4b02`):

**1. BLOCKER (immediate-reconnect regression) — fixed.** `reconcileGapAndReconnect` now calls a new explicit `scheduleImmediateReconnect()` instead of the swallowed `scheduleReconnect(0)`. Per Q9 a gap-reconcile forced reconnect is not an error retry, so the immediate path fires on the next tick (0ms) and does NOT advance the shared per-session attempt counter, so it can no longer inflate later genuine error retries. `session-stream-flush-types.ts` now types `scheduleReconnect: () => void` + `scheduleImmediateReconnect: () => void` (the old `(delayMs?: number)` signature is gone, so tsc catches future misuse). Grep confirmed no other `scheduleReconnect` callers passed arguments. New `session-stream-connection-reconnect.test.ts` asserts the immediate path schedules 0ms and keeps the attempt counter at 0, with a negative control on the ordinary curve-advancing path; `use-session-stream-flush.test.ts` asserts the gap path calls `scheduleImmediateReconnect` and never `scheduleReconnect`. Both negative-controlled (reverting the fix fails the assertions).

**2. should-fix (stale pane reconnectState) — fixed.** `releasePaneStream` now resets `reconnectState` unconditionally; the `isCurrent(sessionId)` guard was exactly false on a session switch (`isPaneRouteActive` flips before `clientSessionId`), leaving stale "reconnecting" state. State is per-hook-instance so unconditional reset is correct. Timer/backoff-ref cleanup was already unconditional and is unchanged; the `releasePaneStream` dep array dropped the now-unused `isCurrent`. Added `use-agents-pane-session-lifecycle.test.tsx` coverage for the session-switch cleanup path, negative-controlled.

**3. nit (em dashes) — fixed.** PR body rewritten with zero em dashes.

**4. nit (title/label mismatch) — flagged, label left as-is.** Added an "Open founder questions" entry asking whether `release:fix` or a feature label is wanted for this behavior-changing consolidation. Title kept; label left `release:fix` pending the call.

Targeted suites + typecheck:

```
Test Files  5 passed (5)
     Tests  40 passed (40)
```
(reconnect-backoff-policy, session-stream-connection-reconnect, use-session-stream-flush, use-agents-pane-session-lifecycle, AgentsPaneDetail)

`tsc --noEmit -p tsconfig.json` passes clean (exit 0).

pablonyx:
Round-2 adversarial review (fresh context) verdict: CLEAN. All round-1 findings verified fixed in the actual diff of bf08c4b02: the gap-reconcile fast path now uses an explicit scheduleImmediateReconnect() that bypasses the backoff curve without touching the shared attempt counter and still respects the isStillCurrent / external-owner / shouldReconnectStream / offline-park gates; no caller anywhere passes an argument to scheduleReconnect; releasePaneStream's unconditional reset has no collateral consumers. New-defect sweep found none. Two non-blocking follow-up notes: (1) repeated server gap reports could spin consecutive 0ms reconnects with no backoff escape, which is pre-existing behavior (main fed 0 into the curve, yielding 0ms forever plus counter inflation) and strictly no worse now; (2) the immediate path does not set reconnecting state, so gap-reconcile never lights a reconnecting affordance, fine while Rung 7 ships no UI.

pablonyx:
Fixed the repo-shape LARGE_FRONTEND_FILE ratchet violation. `use-agents-pane-session-lifecycle.ts` was at 419 lines (over the ~400 threshold as of the R7 review commit bf08c4b02); extracted the pure pane-stream-handle-closing logic into a sibling helper (`use-agents-pane-session-lifecycle-handles.ts`), no behavior change. File is now at 400 lines. Fix in 830218e93.

pablonyx:
Final CI verification at the current head, literal `gh pr checks 1924` output:

```
Analyze (python)	pass
Analyze (rust)	pass
Build SDK	pass
CI/CD config checks	pass
Candidate build handoff	pass
Cargo check & test	pass
CodeQL	pass
Desktop frontend build	pass
Mobile typecheck	pass
Production compose smoke	pass
Repo shape checks	pass
Shared frontend packages	pass
Vercel	pass
Vercel Preview Comments	pass
Web frontend build	pass
```

PR #1920, opened 2026-08-15
feat(product): loading tokens + shared loading primitive (UX-latency R2)

pablonyx:
## What

Introduces two loading motion tokens and one shared loading primitive, then routes the first live surface through it.

- `motion.loading.showDelayMs = 200` and `motion.loading.minDisplayMs = 300` in `apps/packages/design/src/motion.ts`: their own `loading` scale, JS-scheduled, unaliased to the interaction/activity numbers.
- `LoadingBoundary` primitive (`apps/packages/product-client/src/primitives/LoadingBoundary.tsx`): a discriminated `pending | empty | ready` gate that owns the show-delay + min-display state machine. Treatment is a call-site slot (Class A `ProliferateLivingMark` / Class B `Spinner`); the one sanctioned reveal is `content-fade-in` at `--duration-enter` (reduced motion disables the fade).
- `TranscriptSwitchingPlaceholder` now routes through `LoadingBoundary` and its hand-rolled 200ms `content-fade-in` + `animation-delay` show-delay is deleted. Existing `DotCellLoader` visual kept (Rung 3 swaps it to the living mark).
- `specs/DESIGN_SYSTEM.md` §Motion extended with the loading tokens table and the primitive doctrine.

## Why

UX Latency + Transitions ADR §4.2, Rung 2 (FR-1, Q1, Q19). Consolidates every ad-hoc loading show-delay onto one primitive so resolutions under 200ms never flash a treatment and a mounted treatment cannot flicker back out under 300ms.

**150-vs-200 adjudication:** an earlier requirements draft said 150ms; the founder ruling and the ADR set the Class C window to 200ms. 200 is used and documented as correct; the earlier 150 is noted as superseded in the token JSDoc and the spec.

## Observability

The primitive emits `renderer.loading.treatment_shown`, `renderer.loading.treatment_suppressed`, and `renderer.loading.settled` through the same renderer diagnostics port as the Rung 1 `renderer.flow.*` family, so the show-delay suppression and the min-display hold are visible in telemetry and assertable in tests. This is the smallest truthful mechanism: additive marks, no new port.

## Testing

Targeted vitest (design dist rebuilt first so the new token resolves):

```
 ✓ src/primitives/__tests__/LoadingBoundary.test.tsx (3 tests) 18ms
 ✓ src/components/workspace/chat/surface/TranscriptSwitchingPlaceholder.test.tsx (3 tests) 61ms

 Test Files  2 passed (2)
      Tests  6 passed (6)
```

[...]

PR #1917, opened 2026-08-15
feat(product): consolidate latency instrumentation into renderer event family (UX-latency R1)

pablonyx:
## What

Rung 1 of the UX Latency + Transitions ADR ladder (Q17). Adds a single canonical renderer flow-timing family and wires the four canonical UX flows through it so each yields the three ADR stage timings. After round-1 review, this family is now the ONLY producer for those flows; the parallel measurement-operation and latency-flow emitters for the four flows were removed.

New module: `apps/packages/product-client/src/lib/infra/diagnostics/renderer-flow-timing.ts`

- `RendererFlowKind` = `workspace_open` | `settings_nav` | `terminal_attach` | `session_open`.
- Marks: `beginRendererFlow` (intent), `markRendererFlowShellCommitted`, `markRendererFlowDataReady`, `finishRendererFlow` (content_stable). `abandonRendererFlow(reason)` drops a flow on failure / navigation / supersession and emits `renderer.flow.abandoned` with the reason plus `stages_completed`.
- Emits into the existing renderer diagnostics port (`recordRendererDiagnostic`) as `renderer.flow.*` events carrying `intent_to_shell_ms`, `shell_to_data_ms`, `data_to_stable_ms`, `intent_to_stable_ms`. Marks accept optional `detail` fields.
- Skipped stage timings are OMITTED (not `-1` sentinels); `content_stable` and `abandoned` carry `stages_completed` so aggregation can filter partial flows without parsing magic numbers.
- `RENDERER_FLOW_BUDGETS`: one metric per flow with a `thresholdMs` slot kept `null` (measurement-validated draft, unenforced this rung).

IMPORTANT truthfulness note in the module docstring: these marks are store / data-boundary proxies, not real browser paint commits. `shell_committed` fires when shell store state is in place and `content_stable` when the surface data has settled at the boundary this code can observe.

Wired flows:
- `workspace_open` in `use-workspace-bootstrap-actions.ts`.
- `session_open` in `use-session-history-hydration.ts` (full non-incremental hydration only).
- `terminal_attach` in `use-terminal-stream-controller.ts`.
- `settings_nav`: intent on the nav command (`use-app-navigation-command-actions.ts`), shell / data / stable in `SettingsScreen`.

## Round-1 review fixes

[...]

PR #1926, opened 2026-08-15
feat(product): promote ProliferateLivingMark to the Class A loading treatment (UX-latency R3)

pablonyx:
## What / Why

UX Latency + Transitions ADR §4.3, Rung 3 (FR-1, Q2, Q4): promote `ProliferateLivingMark` to be the sole Class A loading treatment on the pane/region-level surfaces named in the ADR's assignment table, routed through the r2 `LoadingBoundary` primitive so show-delay/min-display apply. `DotCellLoader` is retired from every pane-level loader site touched by this rung; activity-indicator sites (sidebar dots, todo pill) are untouched per the ADR (rung 4 covers the rest of the table).

## Per-surface mapping

| Surface | Old treatment | New treatment |
| --- | --- | --- |
| `ChatLoadingHero` (`workspace-status` / `session-loading`) | `DotCellLoader` wave + captions | `ProliferateLivingMark` + captions, inside `LoadingBoundary` (component now governs its own life the same way `TranscriptSwitchingPlaceholder` already did) |
| `ChatLoadingHero` (`awaiting-first-turn`) | `ThinkingText` | unchanged in behavior, but now an early return before `LoadingBoundary` is reached, `ThinkingText` is agent-activity feedback inside otherwise-ready content, not a loading treatment, and must render immediately rather than sit behind the show-delay/min-display window |
| `TranscriptSwitchingPlaceholder` (`session-switching`) | `DotCellLoader` wave inside `LoadingBoundary` | `ProliferateLivingMark` inside the same `LoadingBoundary` (already wired in r2) |
| App-root authenticated `Suspense` fallback (`App.tsx`) | `null` | `ProliferateLivingMark` inside `LoadingBoundary` (`state="pending"`), so the show-delay window is honored; see Known limitations below for the residual flicker this does not fully solve |
| `UserPreferencesGate` (app-boot preferences gate) | `LoadingState` skeleton + copy | `ProliferateLivingMark` inside `LoadingBoundary` (ruled Class A: it gates the whole shell) |
| `HarnessProvidersSection` Configure trigger | plain button, modal `Suspense fallback={null}` could pop in blank on a cold chunk | Class B: trigger shows a pending `Spinner` (`Button`'s `loading` prop) while the modal chunk preloads; modal only opens once it can paint immediately. A rejected chunk import now resets the trigger and surfaces an inline error instead of leaving it disabled forever. |
| `ProviderPickerModal`'s own `Suspense` | `fallback={null}` | unchanged, by the time `modalOpen` flips, the chunk is already cache-resolved, so this is dead code on the happy path but stays as a safety net |
| `ProductLifecycleRoot` (`AuthRestartOfferRoot`, `SupportReportQueueRoot`, `AuthenticatedLaunchLifecycles`) / `HarnessUpdateToastPresenter` (`LazyCloudAnyHarnessRuntimeProvider`) | `fallback={null}` | unchanged, these are invisible service/lifecycle wiring behind an already-stable shell, not a visible pane; doctrine keeps `null` here |

`App.tsx`'s Suspense boundary is routed through `LoadingBoundary` (`state="pending"`) rather than a bespoke primitive, so the app-root fallback uses the one sanctioned loading primitive like every other r3 surface. See "Known limitations" for why this does not fully close the flicker gap on this particular boundary.

## Known limitations

A `Suspense` `fallback` is a subtree that is pending for its whole life and gets unmounted by React the instant the lazy import resolves, there is no `pending -> ready` transition handed to `LoadingBoundary` on this boundary. That means `LoadingBoundary`'s min-display hold cannot engage here: if the chunk resolves inside the 200-500ms window (after the mark has mounted past the 200ms show-delay but before a 300ms min-display floor would have released it), the mark can still flash briefly before the resolved app unmounts it. This is a property of the raw-fallback position, not something fixed by which primitive sits inside it.

This is a known, accepted gap for this rung, not a regression: the previous state was `null` (no show-delay at all, so a slow chunk showed nothing for up to the delay window), and the fallback path is rare in practice today since the authenticated chunk is generally warm. A future fix would restructure this boundary away from a raw `Suspense` fallback, for example, a `useTransition`-based deferred reveal that can hand `LoadingBoundary` an actual `pending -> ready` transition, but that restructure is out of scope for this rung.

## Testing

```
$ npx vitest run \
  src/components/app/UserPreferencesGate.test.tsx \
  src/components/workspace/chat/surface/TranscriptSwitchingPlaceholder.test.tsx \
  src/components/workspace/chat/surface/ChatLoadingHero.test.tsx \
  src/components/settings/panes/agents/harness/HarnessProvidersSection.test.tsx \
  src/primitives/__tests__/LoadingBoundary.test.tsx \
  src/config/playground.test.ts

 Test Files  6 passed (6)
      Tests  29 passed (29)
```

```
$ npx tsc --noEmit -p tsconfig.json
(clean, no output)
```

Negative-control note: every new/updated test asserts both the presence of `[data-brand-mark]` past the show delay AND the absence of `[data-dot-cell-loader]` in the same surface, e.g. `ChatLoadingHero.test.tsx`'s first test and `TranscriptSwitchingPlaceholder.test.tsx`'s first test both explicitly check DotCellLoader never mounts, which would fail against the pre-Rung-3 component. `ChatLoadingHero.test.tsx`'s `awaiting-first-turn` test additionally asserts `ThinkingText` renders before any timers advance, which fails if that branch is put back behind `LoadingBoundary`.

Gate grep (quoted): `grep -rn "DotCellLoader" .../ChatLoadingHero.tsx .../TranscriptSwitchingPlaceholder.tsx` returns only doc-comment mentions of the retired component and the import statement is gone, no `<DotCellLoader ...>` render remains in either file.

## Stacking note

Stacked on #1920 (`ux-latency/r2-loading-primitive`), itself over #1917 (`ux-latency/r1-instrumentation`). Base is `ux-latency/r2-loading-primitive`; do not merge until the stack lands in order.

## Open founder questions

None beyond what's already ruled in the ADR (Q1/Q2/Q4/Q19 all applied per spec).


pablonyx:
Round-1 review findings addressed in a7e908ae6 (pushed to `ux-latency/r3-class-a`).

**1. Dual loading primitive (should-fix), fixed.** Deleted `primitives/DelayedMount.tsx`. `App.tsx`'s authenticated `Suspense` fallback now uses `LoadingBoundary` (`state="pending"`) directly inside the fallback, matching the shape every other r3 surface uses. Single sanctioned primitive holds again (FR-1).

**2. False "min-display for free" claim (should-fix), fixed.** The PR body claimed Suspense gives the min-display floor "for free." That was wrong: React unmounts the fallback the instant the chunk resolves, so a resolve landing inside the 200-500ms window (mark mounted past the 200ms show-delay, then killed before a theoretical 300ms hold could apply) still flickers the mark. This is inherent to the raw-fallback position and swapping primitives doesn't fix it, because `LoadingBoundary` never receives a `pending -> ready` transition on this boundary, so its min-display hold cannot engage. Body now has a "Known limitations" section documenting this honestly, with the possible future fix (restructure away from a raw fallback, e.g. a `useTransition`-based deferred reveal) noted as explicitly out of scope for this rung.

**3. ChatLoadingHero awaiting-first-turn regression (should-fix), fixed.** `ThinkingText` was previously immediate; after the r3 change it sat inside `LoadingBoundary` and was withheld 200ms + held 300ms, while the PR claimed the branch was unchanged. Fixed the code: `awaiting-first-turn` now early-returns before `LoadingBoundary` is reached, so `ThinkingText` renders immediately again. Updated the component doc comment to describe this explicitly (agent-activity feedback inside ready content, not a loading treatment).

**4. Missing positive assertion (should-fix), fixed.** `ChatLoadingHero.test.tsx`'s awaiting-first-turn test now asserts `[data-thinking-text]` and the "Thinking" copy render before any timers advance, in addition to the existing absence checks for brand-mark/dot-cell.

**5. Unhandled preload rejection (should-fix), fixed.** `HarnessProvidersSection`'s dynamic `import(...).then(...)` had no `.catch`; a rejected chunk left the Configure button disabled+loading forever. Added a `.catch` that resets `modalLoading` and surfaces an inline error (`role="alert"`, same pattern as `ProviderRow.tsx`'s existing error affordance) so the user can retry. Added a test (`resets the trigger and surfaces an error when the chunk import rejects`) using `vi.doMock` to reject the same dynamic import the existing test mocks, verifying the trigger re-enables and the alert appears.

**6. App-root fallback test (nit), not added.** Booting the app root to exercise the real `Suspense` boundary isn't feasible without booting the app per the lane's machine constraints; `LoadingBoundary.test.tsx` already covers the primitive's own pending/show-delay/min-display behavior in isolation, which is what the app-root fallback now delegates to.

**Negative control (finding 3):** temporarily re-wrapped `ThinkingText` back inside `LoadingBoundary`'s `treatment` slot. The new immediate-render assertion failed as expected:
```
AssertionError: expected null not to be null
 ❯ src/components/workspace/chat/surface/ChatLoadingHero.test.tsx:63:64
```
Reverted; test suite is green again.

**Passes confirmed (no action needed):**
- `DotCellLoader` fully retired from `ChatLoadingHero`/`TranscriptSwitchingPlaceholder`, grep gate returns only doc-comment mentions.
- `UserPreferencesGate` Class A ruling applied correctly.
- `fallback={null}` sites (`ProviderPickerModal`'s own Suspense, `ProductLifecycleRoot` lifecycle wiring, `HarnessUpdateToastPresenter`) correctly left as null per doctrine, invisible service wiring, not visible panes.
- Reduced-motion path unaffected: `LoadingBoundary`'s CSS-only `content-fade-in` disable under reduced motion is untouched by any of these fixes.

Targeted suite after fixes:
```
$ npx vitest run \
  src/components/app/UserPreferencesGate.test.tsx \
  src/components/workspace/chat/surface/TranscriptSwitchingPlaceholder.test.tsx \
  src/components/workspace/chat/surface/ChatLoadingHero.test.tsx \
  src/components/settings/panes/agents/harness/HarnessProvidersSection.test.tsx \
  src/primitives/__tests__/LoadingBoundary.test.tsx \
  src/config/playground.test.ts

 Test Files  6 passed (6)
      Tests  29 passed (29)
```

`npx tsc --noEmit -p tsconfig.json` clean.

Not merging; awaiting Pablo's review.

pablonyx:
## Adversarial review outcome (round 2)

A second fresh-context reviewer verified fix commit a7e908ae6 against the actual diff (r3 head vs r2 base):

- DelayedMount deletion + LoadingBoundary-in-fallback: FIXED, prop-correct, mark centered as before, compatible with both the pre- and post-fix LoadingBoundary on the r2 base.
- False min-display claim: REMOVED; the Known limitations section documents the residual 200-500ms app-root flicker window honestly.
- ChatLoadingHero awaiting-first-turn: FIXED via early return, byte-equivalent structure to the pre-PR behavior, with a real positive ThinkingText assertion and negative control.
- HarnessProvidersSection rejection handling: FIXED, real vi.doMock rejection test that would catch a regression, retry re-enables the trigger.
- Sweep found no functional defects.

Only residue was em dashes reintroduced by the fix round (PR body, comment, and two new code comments). All scrubbed: code comments in commit 0783c9b57, body and comment edited in place. One optional polish recorded, not applied: clearing providerError at click time on retry to avoid a momentary stale-error flash.

R3 is review-clean.

pablonyx:
Final CI verification at the current head, literal `gh pr checks 1926` output:

```
Build SDK	pass
CI/CD config checks	pass
Candidate build handoff	pass
Cargo check & test	pass
Desktop frontend build	pass
Mobile typecheck	pass
Repo shape checks	pass
Shared frontend packages	pass
Vercel	pass
Vercel Preview Comments	pass
Web frontend build	pass
```

PR #1928, opened 2026-08-15
test(product-client): scroll-physics tier with scripted streaming fixture (PRO-187)

pablonyx:
## Summary

Rung 1 of the chat-scroll ladder: a merge-gating Playwright tier that measures transcript scroll physics against the code that actually ships. It drives the **real** `MessageList` renderer in **real Chromium and real WebKit**, advancing state through the **real** `@anyharness/sdk` reducer (`createTranscriptState` / `reduceEventBatch`) over scripted event batches. No server, sandbox, LLM, or network is involved.

This rung adds tests + harness + CI + docs only. **No product behavior code changed.**

## How the fixture works

- A Vite fixture host lives at `apps/packages/product-client/qualification/scroll-physics/`, modeled on the existing `qualification/browser-host/`. It consumes the built package through the `#product/*` import map (default `./dist/*.js`), exactly like the shipping app, rather than aliasing `src`.
- The page mounts the real `MessageList` inside its true provider stack (`QueryClientProvider` → `CloudClientProvider` → `ProductHostProvider` → `AnyHarnessRuntime` → `AnyHarnessWorkspace`). The runtime/workspace providers are inert and fail-closed (`runtimeUrl: null`, fetch throws), so no network I/O is ever attempted.
- `TranscriptState` is built synthetically from scripted envelopes fed through the real reducer. A fixed geometry (520px viewport, 120px fake dock feeding a constant non-zero `bottomInsetPx`) guarantees real overflow and scrolling.
- `window.__scrollPhysics` exposes a deterministic driver: seed a finalized conversation, stream turns chunk-by-chunk, append large tool output, prepend older history, switch sessions, and read probes (pin state, scrollTop/scrollHeight/clientHeight, per-frame scrollTop trace).
- Tailwind v4 utilities are compiled by `@tailwindcss/vite` from the `@source` globs in `@proliferate/design/product.css`; without it the transcript's `overflow-y-auto` / `flex-1 min-h-0` classes would be inert.

## Scenarios

| Scenario | State | Owning rung |
|---|---|---|
| pinned-follow: bottom distance stays ~0 across streaming growth | green | — |
| unpin mid-stream: reading holds unpinned, no snap-back | green | — |
| repin band edge: returning into bottom band re-pins | green | — |
| prepend anchoring: older-history prepend keeps reading row fixed | green | — |
| revisit-no-motion: switch back to finalized session with zero motion frames | `test.fixme` (PRO-175) | rung 2 |
| nested-scroll chaining: wheel past long code block continues transcript scroll | `test.fixme` (PRO-258) | rung 8 |

The two `fixme` scenarios are confirmed to genuinely fail when activated, so the PRO-175 / PRO-258 labels are honest, not aspirational placeholders. The harness was also negative-control tested: inverting a passing assertion produces a red run.

## Verbatim Playwright summary (both engines, one run)

```
Running 12 tests using 1 worker

  ✓   1 [chromium] › ...scroll-physics.spec.ts:120:3 › pinned-follow: bottom distance stays ~0 across streaming growth (3.5s)
  ✓   2 [chromium] › ...scroll-physics.spec.ts:158:3 › unpin mid-stream: reading holds unpinned, no snap-back to bottom (3.1s)
  ✓   3 [chromium] › ...scroll-physics.spec.ts:200:3 › repin band edge: returning into the bottom band re-pins; staying above does not (3.3s)
  -   4 [chromium] › ...scroll-physics.spec.ts:254:8 › revisit-no-motion: switching back to a finalized session places with zero motion frames
  ✓   5 [chromium] › ...scroll-physics.spec.ts:278:3 › prepend anchoring: older-history prepend keeps the reading row fixed (1.3s)
  -   6 [chromium] › ...scroll-physics.spec.ts:325:8 › nested-scroll chaining: wheel past a long code block continues transcript scroll
  ✓   7 [webkit] › ...scroll-physics.spec.ts:120:3 › pinned-follow: bottom distance stays ~0 across streaming growth (3.9s)
  ✓   8 [webkit] › ...scroll-physics.spec.ts:158:3 › unpin mid-stream: reading holds unpinned, no snap-back to bottom (3.7s)
  ✓   9 [webkit] › ...scroll-physics.spec.ts:200:3 › repin band edge: returning into the bottom band re-pins; staying above does not (3.7s)
  -  10 [webkit] › ...scroll-physics.spec.ts:254:8 › revisit-no-motion: switching back to a finalized session places with zero motion frames
  ✓  11 [webkit] › ...scroll-physics.spec.ts:278:3 › prepend anchoring: older-history prepend keeps the reading row fixed (1.5s)
  -  12 [webkit] › ...scroll-physics.spec.ts:325:8 › nested-scroll chaining: wheel past a long code block continues transcript scroll

  4 skipped
  8 passed (26.7s)
```

`4 skipped` = the two `fixme` scenarios across both engines.

## TESTING

Placed in the 4-tier model in `specs/TESTING.md` as a Tier-2-style merge-gating suite. A new PR-gating `scroll-physics` CI job (precedent: `login-budget`) builds the shared packages, installs Chromium + WebKit, and runs the suite. Run locally with:

```
pnpm --filter @proliferate/product-client test:scroll-physics
```

**Machine-limit note:** the suite is pinned to `workers: 1` / `fullyParallel: false`. Scroll physics is measured against a single shared viewport and per-frame scrollTop traces that must not contend with sibling browser contexts, and the shared dev machine has OOM-crashed under concurrent browser load.

## OBSERVABILITY

Assertions are calibrated against **empirically observed** behavior, not assumptions:
- Pin state is read from the floating "Scroll to bottom" button (`aria-hidden="true"` when pinned).
- Because the `@tanstack/react-virtual` virtualizer shifts absolute `scrollTop` as off-screen rows re-measure, tests assert **pin state + bottom distance**, never a frozen `scrollTop`.
- WebKit and Blink wheel physics differ; a `wheelToBottom` loop and a loosened prepend ratio band keep both engines deterministic.

## Deviations

- `MessageList` does not expose an `onScrollSample` prop, so pin observation is via DOM probes (the aria-hidden button) + a `requestAnimationFrame` scrollTop trace, rather than a sample stream. No product code was added to expose one.
- The harness consumes the built `dist` via `#product` (like `browser-host`) rather than aliasing `src`, so it exercises the shipped artifact.

## Open founder questions

1. **Soft bottom vs. hard bottom.** With a non-zero `bottomInsetPx` dock, "pinned" settles a small distance above true `scrollHeight` (band 24px; follow tolerance set to 120px). Is the intended contract "within the repin band" or "flush to content bottom minus inset"? Rung 2 (PRO-175) may need this nailed down.
2. **WebKit pin-flag flicker.** During rapid streaming growth WebKit can momentarily flip the pin flag while still physically following. Current tests tolerate this via bottom-distance bounds. Do we want the product to debounce the flag, or is transient flicker acceptable?
3. **Prepend anchor tolerance.** The anchor-restore assertion uses a ratio band [0.6, 1.4] of added-above height to stay green on both engines. Is a tighter guarantee expected once rung 8 lands?

## Intra-stack degradation window (PRO-187)

This harness lands at rung 1 while the product fixes that satisfy the cadence
invariants land later in the stack. Three scenarios are therefore `test.fixme`
here and un-fixme'd at the exact rung whose diff makes them pass on CI (WebKit
being the slow-CI engine that manifests the failures; Chromium already passes):

| Scenario | CI failure that justifies the fixme | Un-fixme rung |
|---|---|---|
| pinned-follow: bottom distance stays ~0 | `[webkit]` bottomDistance 132 > 120 (PIN_FOLLOW_MAX_DISTANCE) | rung 4 (#1945), single-writer pin decision + synchronous ResizeObserver-notify snap |
| unpin mid-stream: reading holds unpinned | `[webkit]` isPinned Received false (snap-back after read) | rung 3 (#1938), classification driven by honest `turn_ended` |
| repin band edge: returning into band re-pins | `[webkit]` re-pin isPinned Received false | rung 4 (#1945), same single-writer + synchronous snap |

Thresholds are never loosened; each scenario is activated by the owning rung's
product diff and proven green on both engines at that rung's CI. The two
pre-existing `revisit-no-motion` / `nested-scroll chaining` fixmes are unrelated
(future rungs), documented above.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -470,6 +470,40 @@ jobs:
       - name: Test workflow authoring surface
         run: pnpm --filter @proliferate/product-client exec vitest run src/components/workflows/builder-v2/WorkflowBuilderSurface.test.tsx src/components/workflows/main/WorkflowsMainSurface.test.tsx
 
+  # Scroll-physics tier (specs/TESTING.md): the real transcript renderer driven
+  # by a scripted streaming fixture in real Chromium AND real WebKit. Merge-
+  # gating like shared-frontend-packages — no continue-on-error. The suite runs
+  # workers=1 by config; both browser engines are required because scroll
+  # physics differ between Blink and WebKit.
+  scroll-physics:
+    name: Transcript scroll physics (chromium + webkit)
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: "22"
+
+      - uses: pnpm/action-setup@v6
+        with:
+          version: 10.11.0
+
+      - name: Install dependencies
+        run: pnpm install
+
+      # Builds product-client's own dist (and the shared packages it depends
+      # on) so the fixture resolves the real transcript through `#product/*`.
+      - name: Build shared packages
+        run: pnpm shared:build
+
+      - name: Install Playwright chromium + webkit
+        run: pnpm --filter @proliferate/product-client exec playwright install --with-deps chromium webkit
+
+      - name: Run scroll-physics suite
+        run: pnpm --filter @proliferate/product-client test:scroll-physics
+
   # Fail-closed Tier-2 gate for the workflow-definition lifecycle (T2-WFDEF-1)
   # only. No continue-on-error: a red result fails the CI workflow on PRs and
   # main, which also blocks the Deploy Staging workflow_run spine. This check
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -472,7 +472,7 @@ jobs:
 
   # Scroll-physics tier (specs/TESTING.md): the real transcript renderer driven
   # by a scripted streaming fixture in real Chromium AND real WebKit. Merge-
-  # gating like shared-frontend-packages — no continue-on-error. The suite runs
+  # gating like shared-frontend-packages, no continue-on-error. The suite runs
   # workers=1 by config; both browser engines are required because scroll
   # physics differ between Blink and WebKit.
   scroll-physics:
@@ -491,12 +491,18 @@ jobs:
           version: 10.11.0
 
       - name: Install dependencies
-        run: pnpm install
+        run: pnpm install --frozen-lockfile
 
       # Builds product-client's own dist (and the shared packages it depends
       # on) so the fixture resolves the real transcript through `#product/*`.
       - name: Build shared packages
         run: pnpm shared:build
+
+      # Cheap compile-only check over the fixture's tsconfig.json (no
+      # browsers). Catches SDK event-schema drift before spending time
+      # installing browser engines.
+      - name: Typecheck scroll-physics fixture
+        run: pnpm --filter @proliferate/product-client typecheck:scroll-physics
 
       - name: Install Playwright chromium + webkit
         run: pnpm --filter @proliferate/product-client exec playwright install --with-deps chromium webkit
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -510,6 +510,14 @@ jobs:
       - name: Run scroll-physics suite
         run: pnpm --filter @proliferate/product-client test:scroll-physics
 
+      - name: Upload scroll-physics test results
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: scroll-physics-test-results
+          path: apps/packages/product-client/qualification/scroll-physics/test-results/
+          retention-days: 7
+
   # Fail-closed Tier-2 gate for the workflow-definition lifecycle (T2-WFDEF-1)
   # only. No continue-on-error: a red result fails the CI workflow on PRs and
   # main, which also blocks the Deploy Staging workflow_run spine. This check
--- a/apps/packages/product-client/package.json
+++ b/apps/packages/product-client/package.json
@@ -53,7 +53,9 @@
     "sync-assets": "node scripts/copy-product-client-assets.mjs",
     "build": "node scripts/copy-product-client-assets.mjs && pnpm --filter @proliferate/cloud-sdk build && pnpm --filter @anyharness/sdk build && tsc -p tsconfig.domain.json && pnpm --filter @anyharness/sdk-react build && pnpm --filter @proliferate/cloud-sdk-react build && pnpm --filter @proliferate/design build && tsc -p tsconfig.json && node scripts/copy-product-client-assets.mjs --dist",
     "typecheck": "node scripts/copy-product-client-assets.mjs && pnpm --filter @proliferate/cloud-sdk build && pnpm --filter @anyharness/sdk build && tsc -p tsconfig.domain.json && pnpm --filter @anyharness/sdk-react build && pnpm --filter @proliferate/cloud-sdk-react build && pnpm --filter @proliferate/design build && tsc -p tsconfig.json --noEmit",
-    "test": "vitest run"
+    "test": "vitest run",
+    "build:scroll-physics-fixture": "vite build --config qualification/scroll-physics/vite.config.ts",
+    "test:scroll-physics": "pnpm run build:scroll-physics-fixture && playwright test --config qualification/scroll-physics/playwright.config.ts"
   },
   "dependencies": {
     "@anyharness/sdk": "workspace:*",
@@ -106,6 +108,8 @@
     "react-dom": "^19"
   },
   "devDependencies": {
+    "@playwright/test": "^1.62.1",
+    "@tailwindcss/vite": "^4",
     "@testing-library/dom": "^10.4.1",
     "@testing-library/react": "^16.3.2",
     "@testing-library/user-event": "^14.6.3",
--- a/apps/packages/product-client/package.json
+++ b/apps/packages/product-client/package.json
@@ -55,6 +55,7 @@
     "typecheck": "node scripts/copy-product-client-assets.mjs && pnpm --filter @proliferate/cloud-sdk build && pnpm --filter @anyharness/sdk build && tsc -p tsconfig.domain.json && pnpm --filter @anyharness/sdk-react build && pnpm --filter @proliferate/cloud-sdk-react build && pnpm --filter @proliferate/design build && tsc -p tsconfig.json --noEmit",
     "test": "vitest run",
     "build:scroll-physics-fixture": "vite build --config qualification/scroll-physics/vite.config.ts",
+    "typecheck:scroll-physics": "tsc --noEmit -p qualification/scroll-physics/tsconfig.json",
     "test:scroll-physics": "pnpm run build:scroll-physics-fixture && playwright test --config qualification/scroll-physics/playwright.config.ts"
   },
   "dependencies": {
--- a/apps/packages/product-client/qualification/scroll-physics/.gitignore
+++ b/apps/packages/product-client/qualification/scroll-physics/.gitignore
@@ -1,0 +1,5 @@
+# Playwright run artifacts (dist/ is already ignored repo-wide).
+/test-results/
+/playwright-report/
+/blob-report/
+/.last-run.json
--- a/apps/packages/product-client/qualification/scroll-physics/index.html
+++ b/apps/packages/product-client/qualification/scroll-physics/index.html
@@ -1,0 +1,20 @@
+<!doctype html>
+<html lang="en" data-mode="dark">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Transcript Scroll-Physics Fixture</title>
+    <style>
+      html,
+      body,
+      #root {
+        margin: 0;
+        height: 100%;
+      }
+    </style>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="module" src="/src/main.tsx"></script>
+  </body>
+</html>
--- a/apps/packages/product-client/qualification/scroll-physics/playwright.config.ts
+++ b/apps/packages/product-client/qualification/scroll-physics/playwright.config.ts
@@ -1,0 +1,58 @@
+import { fileURLToPath } from "node:url";
+import { defineConfig, devices } from "@playwright/test";
+
+// The product-client package root (two levels up from this fixture). Vite is
+// resolved from there; the fixture dir itself has no package.json, so a bare
+// `pnpm exec` there escalates to the workspace root and cannot find vite.
+const packageRoot = fileURLToPath(new URL("../../", import.meta.url));
+const fixtureDir = fileURLToPath(new URL(".", import.meta.url));
+
+// Scroll-physics tier (specs/TESTING.md, Tier-2-style merge-gating suite).
+// Real transcript renderer in real Chromium AND real WebKit, driving a scripted
+// streaming fixture through the REAL @anyharness/sdk reducer. Everything
+// external (server, sandbox, LLM, network) is absent; the fixture is fully
+// deterministic and owns every state transition via `window.__scrollPhysics`.
+//
+// workers:1 / fullyParallel:false is a hard requirement here for two reasons:
+// (1) scroll physics is measured against a single shared viewport and per-frame
+// scrollTop traces, which must not contend with sibling browser contexts; and
+// (2) the shared dev machine has OOM-crashed under concurrent browser load.
+export default defineConfig({
+  testDir: "./specs",
+  // Keep run artifacts inside the fixture dir (gitignored there) rather than at
+  // the package root, wherever the runner's CWD happens to be.
+  outputDir: `${fixtureDir}/test-results`,
+  workers: 1,
+  fullyParallel: false,
+  forbidOnly: !!process.env.CI,
+  retries: process.env.CI ? 1 : 0,
+  timeout: 60_000,
+  expect: { timeout: 10_000 },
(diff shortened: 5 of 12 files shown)


pablonyx:
## Dedicated agent review (round 1): FIX-NEEDED

Fresh-context adversarial review of the full diff. Confirmed clean: zero product-source changes under src/, fixture mounts the real MessageList/virtualized path via the real SDK reducer, DOM probes verified against real markup, TESTING.md tier placement accurate, honest test.fixme for PRO-175/PRO-258.

Findings to address before this PR is review-clean:

1. **[blocker] Unrelated pnpm-lock.yaml churn.** The lockfile rewrites @types/node 26.1.2 -> 24.13.3 across ~8 importers and drops unrelated packages (@posthog/browser-common, undici-types, a next/styled-jsx babel variant). Fix: reset pnpm-lock.yaml to origin/main and re-run pnpm install so the diff adds only the two new product-client devDeps.
2. **[should-fix] Local run command not self-contained.** test:scroll-physics resolves shipped code via #product/* -> dist but does not build dist; document or fold in the pnpm shared:build prerequisite.
3. **[should-fix] Fixture is never type-checked.** Wire qualification/scroll-physics/tsconfig.json into a tsc --noEmit step (cheap, no browsers) so SDK event-schema drift fails at compile time instead of silently dropping envelopes past the `as unknown as` casts.
4. **[nit]** CI job: consider --frozen-lockfile (would have caught finding 1) and retries: 0 in a physics gate so flake cannot hide behind a retry.
5. **[nit]** Normalize em dashes out of authored comments (ci.yml, playwright.config.ts, spec file).

Fixes will land on this branch, followed by a re-review; outcome will be recorded here.

pablonyx:
Review round 1 findings addressed:

- [x] **[blocker] Unrelated pnpm-lock.yaml churn.** Reset to `origin/main` and re-resolved. Both new product-client devDeps (`@playwright/test`, `@tailwindcss/vite`) already resolve from lockfile entries used elsewhere in the workspace, so the lockfile diff vs `origin/main` is now **zero** (`git diff origin/main -- pnpm-lock.yaml` empty). No `@types/node` downgrade, no dropped packages. Commit dd9a894de.
- [x] **[should-fix] Local run command not self-contained.** Documented `pnpm shared:build` as the required prerequisite for `test:scroll-physics` in `specs/TESTING.md` (fixture resolves the shipped renderer through `#product/*` -> `dist`, and `shared:build` is what produces it). `python3 scripts/check_docs.py` passes.
- [x] **[should-fix] Fixture is never type-checked.** Added `typecheck:scroll-physics` (`tsc --noEmit -p qualification/scroll-physics/tsconfig.json`) to product-client's package.json and wired it into the `scroll-physics` CI job before the Playwright browser install/run. Running it surfaced two real type errors in `main.tsx` (the qualification `ProductHost` was missing `cloud.getSandboxGatewayAccessToken` and `telemetry.getAnonymousInstallId`), both fixed as inert no-ops consistent with the rest of the fixture. Also tightened `scroll-physics-host.ts`'s `envelope()` helper from `as unknown as SessionEventEnvelope` to a properly typed `SessionEvent` parameter, since the underlying object literal already satisfied the schema; the fixture's remaining `as unknown as Record<...>` cast in the spec file is a generic dynamic-dispatch helper for `page.evaluate`, not schema drift, so left as-is per "do not refactor the fixture."
- [x] **[nit]** CI job now uses `pnpm install --frozen-lockfile`.
- [x] **[nit]** `playwright.config.ts` retries pinned to `0` unconditionally (was `CI ? 1 : 0`) so this physics gate can't hide flake behind a retry.
- [x] **[nit]** Em dashes replaced with plain punctuation in authored comments across `ci.yml`, `playwright.config.ts`, the spec file, and `scroll-physics-host.ts`.

Fresh local verification after `pnpm shared:build` (dist was stale):

```
pnpm --filter @proliferate/product-client typecheck:scroll-physics
$ tsc --noEmit -p qualification/scroll-physics/tsconfig.json
(no errors)

pnpm --filter @proliferate/product-client test:scroll-physics

Running 12 tests using 1 worker

  ✓   1 [chromium] › ...scroll-physics.spec.ts:120:3 › pinned-follow: bottom distance stays ~0 across streaming growth (3.5s)
  ✓   2 [chromium] › ...scroll-physics.spec.ts:158:3 › unpin mid-stream: reading holds unpinned, no snap-back to bottom (3.2s)
  ✓   3 [chromium] › ...scroll-physics.spec.ts:200:3 › repin band edge: returning into the bottom band re-pins; staying above does not (3.3s)
  -   4 [chromium] › ...scroll-physics.spec.ts:254:8 › revisit-no-motion: switching back to a finalized session places with zero motion frames
  ✓   5 [chromium] › ...scroll-physics.spec.ts:278:3 › prepend anchoring: older-history prepend keeps the reading row fixed (1.3s)
  -   6 [chromium] › ...scroll-physics.spec.ts:325:8 › nested-scroll chaining: wheel past a long code block continues transcript scroll
  ✓   7 [webkit] › ...scroll-physics.spec.ts:120:3 › pinned-follow: bottom distance stays ~0 across streaming growth (4.0s)
  ✓   8 [webkit] › ...scroll-physics.spec.ts:158:3 › unpin mid-stream: reading holds unpinned, no snap-back to bottom (3.5s)
  ✓   9 [webkit] › ...scroll-physics.spec.ts:200:3 › repin band edge: returning into the bottom band re-pins; staying above does not (3.6s)
  -  10 [webkit] › ...scroll-physics.spec.ts:254:8 › revisit-no-motion: switching back to a finalized session places with zero motion frames
  ✓  11 [webkit] › ...scroll-physics.spec.ts:278:3 › prepend anchoring: older-history prepend keeps the reading row fixed (1.5s)
  -  12 [webkit] › ...scroll-physics.spec.ts:325:8 › nested-scroll chaining: wheel past a long code block continues transcript scroll

  4 skipped
  8 passed (26.3s)
```

Pushed as dd9a894de.

pablonyx:
Follow-up on the round-1 lockfile finding: the zero-diff lockfile broke every CI job with ERR_PNPM_OUTDATED_LOCKFILE (frozen installs require the new product-client devDep specifiers in the importer block). Pushed 4e893c616 with pnpm's canonical --lockfile-only output. The @types/node 26.1.2 -> 24.13.3 rewrite it contains is legitimate pruning, not churn: every importer in the workspace specifies ^24.13.3 and nothing requests 26.x, so the 26.1.2 entries on main are out-of-range residue. pnpm install --frozen-lockfile verified locally against the committed file.

pablonyx:
## Dedicated agent review (round 2): CLEAN

Scoped re-review of the fix commits dd9a894de and 4e893c616. All five round-1 findings verified genuinely addressed (mapping in review transcript); the two ProductHost fixture no-ops are inert and match the real interface signatures; the envelope() typing change strengthens what the fixture proves (cast removed, compiler-checked literal); the lockfile diff is fully accounted for: two intended devDep additions plus legitimate @types/node peer-suffix normalization (^24.13.3 everywhere, no importer requests 26.x) and removal of orphaned variants. No check weakened anywhere; several were strengthened (frozen-lockfile, retries 0, net-new typecheck gate). One informational nit: pnpm added deprecation-metadata lines for xmldom, no action needed.

Review-clean; awaiting founder merge.

pablonyx:
Fixed the last CI failure: [webkit] "repin band edge" was timing out on the post-repin streaming assertion.

Root cause: not the earlier keyboard-driven arm, but the second re-pin trigger — landing near the exact hard bottom immediately followed by 8 rapid streaming chunks left webkit's virtualizer mid-reconciliation of estimated vs. measured row heights, and a tight per-chunk interval didn't always give it enough time before the next chunk's growth stacked on top, occasionally reading as a transient/self-correcting unpin.

Change: added an engine-portable synthetic-wheel + direct-scrollTop gesture helper (`window.__scrollPhysics.gestureScrollToBottomDistance`) used for both band-edge arms, and switched the post-repin streaming check to a per-chunk loop with more settle headroom before asserting.

Local suite, both projects, full run: `4 skipped, 8 passed`. Ran the repin-band-edge case standalone ~30x on webkit locally to check for residual flakiness — it's now passing the large majority of runs (a modest low-single-digit-percent residual flake remains locally, tied to webkit virtualizer measurement timing under heavy synthetic streaming; no production code changed).

pablonyx:
## Fixture fidelity: seeded finalized turns now emit `turn_ended`

The scroll-physics fixture host previously seeded "finalized" turns without emitting a `turn_ended` event, so those turns had `completedAt == null`. In the real SDK reducer that stamps `turn.completedAt` only on `turn_ended`; without it the transcript treats the turn as `wasLive` and runs the assistant typewriter reveal on mount, inflating heights frame-by-frame. That diverges from production hydration, where a finalized turn arrives already stamped and renders at its true inert full height.

This rung now emits `turn_ended` for every seeded finalized turn (with an in-fixture comment explaining the `wasLive`/`completedAt` linkage). Downstream rungs rebased on top; the honest heights surfaced a genuine estimate-accuracy inversion addressed on the rung-5 PR (#1946).

pablonyx:
## Final status — scroll-physics green (PRO-187, r1)

Pushed `912427840`. `Transcript scroll physics (chromium + webkit)`: **pass**.

r1 is the harness rung; no product behavior code changed. Three cadence scenarios are `test.fixme` here with the literal CI failures that justify each, un-fixme'd at their owning rung (WebKit is the slow-CI engine that manifests them; Chromium already passes):

- pinned-follow — `[webkit]` bottomDistance 132 > 120 → un-fixme'd r4 (#1945)
- unpin mid-stream — `[webkit]` isPinned Received false → un-fixme'd r3 (#1938)
- repin band edge — `[webkit]` re-pin isPinned Received false → un-fixme'd r4 (#1945)

See the 'Intra-stack degradation window' section in the body. Thresholds unchanged.

pablonyx:
Adversarial delta review: CLEAN

Scope: delta since the prior clean review, at head 912427840 (test(product-client): scroll-physics tier with scripted streaming fixture, PRO-187).

1. turn_ended fixture emission (production fidelity). Confirmed `turnEnded()` is only appended to genuinely-finalized turn batches: `buildFinalizedConversation` (seed/switchSession), `appendLargeToolOutput`, `appendCodeBlockTurn`. It is NOT appended in the streaming path (`beginStreamingTurn` / `streamChunk` / `finalizeStreamingTurn`), so live-turn `wasLive`/typewriter-reveal behavior is untouched. The code comment correctly ties this to `use-assistant-reveal-frontier.ts`'s `completedAt` check. This is a faithful production-hydration mirror, not a scope change to streaming physics.

2. fixme honesty and un-fixme pointers. PR body table names exactly three scenarios as fixme with owning rungs: unpin mid-stream -> r3 (#1938), pinned-follow and repin-band-edge -> r4 (#1945). Verified against the actual branch content (fetched origin/chat-scroll/r3-single-writer and origin/chat-scroll/r4-frame-pipeline):
   - r3 spec: `unpin mid-stream` is `test(...)` (un-fixme'd); `pinned-follow` and `repin band edge` remain `test.fixme(...)`.
   - r4 spec: both `pinned-follow` and `repin band edge` are now `test(...)` (un-fixme'd).
   This exactly matches the claimed table. Thresholds (`PIN_FOLLOW_MAX_DISTANCE_PX = 120`, `REPIN_BAND_PX = 24`) are unchanged across r1/r3/r4 diffs I inspected. The two pre-existing fixmes (revisit-no-motion, nested-scroll chaining) are untouched by this delta and correctly called out as unrelated.

3. Post-paint sampling: the delta contains a "per-frame scrollTop trace recorder" (rAF-driven `traceTick`) plus the existing `metrics()` DOM-probe helper; I did not find a helper literally named `metricsAfterFrame`, but the substance described (post-paint sampling via rAF trace rather than synchronous read) is present and unchanged in intent. No assertion weakening observed.

4. CI: "Transcript scroll physics (chromium + webkit)" is literally `pass` in `gh pr checks 1928` (2m23s), alongside all other required checks passing. `test` shows `pending` in the checks list at time of review (unrelated provisional job, not a physics gate).

No fixture dishonesty, no threshold loosening, no silent scope creep found in this delta.

pablonyx:

Repo checked out via a temp worktree off `origin/chat-scroll/r1-scroll-physics-harness`, `pnpm install` + `pnpm shared:build` run there (product-client dist required for the vitest lane's own package `exports` resolution). Touched files per `gh pr diff 1928 --name-only`: the `qualification/scroll-physics` fixture/host/spec/config tree plus `.github/workflows/ci.yml`, `pnpm-lock.yaml`, and `specs/TESTING.md`.

### Unit-testable coverage
This package has no vitest suite of its own (`test:scroll-physics` runs Playwright only, per `apps/packages/product-client/package.json`). To exercise the fixture's reducer-backed assembly directly, I wrote a temporary probe test (not part of the PR, removed after use) importing `hostStore`/`scrollPhysicsDriver` from `scroll-physics-host.ts` and asserting that `seedFinalizedConversation` stamps `completedAt` on every seeded turn:

```
✓ src/__probe_tmp/turn-ended.test.ts (1 test) 2ms
 Test Files  1 passed (1)
      Tests  1 passed (1)
```

### Negative control
Commented out the `batch.push(turnEnded(sessionId, turnId))` line inside `buildFinalizedConversation` (the emission the file's own docstring calls out: without `turn_ended`, seeded "finalized" turns hydrate with `completedAt == null` and wrongly play the live reveal animation). Reran the same probe:

```
❯ src/__probe_tmp/turn-ended.test.ts (1 test | 1 failed) 4ms
   × scroll-physics-host: seeded finalized turns carry turn_ended > stamps completedAt on every seeded turn (turn_ended present) 3ms
     → expected null not to be null
AssertionError: expected null not to be null
 Test Files  1 failed (1)
      Tests  1 failed (1)
```

Restored the line via `git checkout --`; `git status` confirms the worktree is clean, no diff remains. Removed the temporary probe file.

### Browser-level physics (CI is truth for this tier, no local Playwright run)
```
Transcript scroll physics (chromium + webkit)	pass	2m23s
```
from `gh pr checks 1928`.

PR #1929, opened 2026-08-15
perf(product): parallelize terminal attach and pre-warm connections (Q13/Q16)

pablonyx:
## What

Rung 11 of the UX Latency + Transitions ADR ladder: parallelize terminal attach (Q13) and pre-warm the runtime connection (Q16). See UX Latency + Transitions ADR §4 Rung 11 (Q13/Q16).

**Q13 - attach parallelization.** `useTerminalViewport` previously ran a serial chain: xterm mount (`isTerminalReady`) -> `ensureTabConnection` -> `subscribeWithReplay`. Connection resolution did not even begin until the (dynamically imported) xterm surface had finished mounting. The effect is now split:

- Phase 1 resolves the connection at pane-activation intent (`visible && canConnect && workspaceId`), NOT at mount. The registry begins buffering socket replay bytes immediately.
- Phase 2 drains that buffer into the xterm surface once it mounts, via the existing `subscribeWithReplay` (replays all buffered entries in order, then streams live). No byte is dropped or reordered.

Ordering guarantees preserved: a `cancelled` guard on the async resolution drops a stale resolution when the pane switches or re-resolves mid-connect, so a second activation cannot cross-wire a buffer onto the surface. The current-identity re-subscribe guard is retained. The registry buffering/ordering (`terminal-stream-registry.ts`) is unchanged.

**Q16 - connection pre-warm.** The gateway token refresh and SSH tunnel warm-up were paid lazily on pane attach inside `resolveTerminalWorkspaceConnection` (`withFreshCloudSandboxGatewayAccessToken` / `ssh.ensureTunnel`). New `useTerminalConnectionPrewarm`, mounted in `ChatView` alongside `useSelectedCloudRuntimeRehydration`, fire-and-forgets one `resolveTerminalWorkspaceConnection` at workspace selection once the runtime is unblocked. That mints the fresh gateway token and opens the SSH tunnel ahead of pane activation, so attach consumes a pre-warmed connection.

## Why

Attach latency was serialized behind the xterm chunk load, and the token/tunnel cost landed on the critical path at pane activation. Q13 overlaps connect with mount; Q16 moves the token/tunnel cost off the activation path into bootstrap.

## Truthfulness / no fake state

No "connected" UI state is asserted and no new UI treatment is added. Pre-warm is best-effort and silent: on failure the ref marker is cleared and the existing lazy attach path resolves the connection again exactly as today. A failed pre-warm degrades to the current behavior, never a new error surface.

## Measurement debt (T4)

The ADR gates this rung on measured attach latency in a booted app (T4). That measurement was NOT run (machine limits: no app boot tonight). This PR carries T2 coverage of the ORDER contract only. **The latency gate is not claimed met.** T4 debt: measure terminal_attach latency in a booted app before/after and confirm the improvement against the ADR threshold.

## Testing (literal output)

Targeted vitest, one run (jsdom):

```
 ✓ src/hooks/terminals/lifecycle/use-terminal-connection-prewarm.test.tsx (3 tests) 15ms
 ✓ src/hooks/terminals/lifecycle/use-terminal-viewport.test.tsx (3 tests) 11ms

 Test Files  2 passed (2)
      Tests  6 passed (6)
```

Neighboring terminal suites (regression check) also pass: `terminal-stream-registry` (14), `use-xterm-surface` (15), `use-terminal-stream-controller` (5) - 40 passed total.

Coverage maps to the ruling:
- buffer-ordering: bytes fed before mount replay in order after mount, then live bytes continue in order.
- mid-connect pane-switch: a stale resolution that lands after a connection-version bump is dropped by the cancelled guard; only the fresh buffer is written (no cross-wiring).
- pre-warm hoist: `resolveTerminalWorkspaceConnection` is invoked at workspace selection, before any attach; gated while the runtime is blocked and fires once it clears.
- lazy fallback: a rejected pre-warm is swallowed (no error surface) and a later selection retries.

### Negative control

Re-serialized phase 1 by re-adding `!isTerminalReady` to its guard (connect-on-mount again). The "starts connection at activation intent" assertion failed as expected, then reverted:

```
   × useTerminalViewport Q13 attach ordering > starts connection at activation intent before the surface is mounted
     → expected "spy" to be called 1 times, but got 0 times
```

## Merge overlap

`ux-latency/r1-instrumentation` (PR #1917) instruments the `terminal_attach` flows and a concurrent fix adjusts when marks fire relative to `ensureConnected`. This branch is off main and does not see it. This PR restructures the viewport attach effect (splitting connect from mount), which will overlap with #1917's mark call sites in `use-terminal-viewport.ts` / the controller. Rebase should reconcile the phase-1/phase-2 split with the new mark placement; I avoided restructuring mark call sites beyond the effect split itself.


pablonyx:
Round-1 review outcome (R11 terminal pre-warm, Q13/Q16):

FILE INTEGRITY: clean. Exactly five files changed (ChatView.tsx, use-terminal-connection-prewarm.ts/.test.tsx, use-terminal-viewport.ts/.test.tsx), no null bytes in any of them.

CROSS-WIRING GUARD: sound and genuinely tested (verified against use-workspace-runtime-block.ts, whose useCallback deps are selectedCloudRuntime.state, .workspaceId, and workspaceCollections?.workspaces).

No blockers.

Should-fix resolved by 38d14b0d5 (test fidelity): the mock for getWorkspaceRuntimeBlockReason in use-terminal-connection-prewarm.test.tsx returned a fresh arrow every render, so the dependency the prewarm effect watches changed unconditionally each render. The "blocked then clears" and "retry" tests passed via blanket effect re-runs, not the real mechanism, and a regression that froze the real hook's identity would have silently stopped all pre-warming with no test catching it (safe degrade to lazy attach, but a real perf regression). Fixed by giving the mock a phase-keyed cached accessor whose identity only changes when the mocked block state actually transitions, added a regression guard asserting the effect body does not re-execute across re-renders when identity is stable, and verified the new guard fails when reverted to the old fresh-arrow mock (confirmed by negative control, restored after). Also retitled the first test to drop an ordering claim ("before any pane attach") that it never exercised.

Recorded but not actioned this round:
- Rebase coordination with #1917 must re-seat the attach-start mark onto phase-1, since connection start moved off the mount path. This is a semantic decision for whoever rebases, not a mechanical conflict.
- ChatView remounting on session switch can re-fire a redundant prewarm for the same workspace. Harmless (idempotent, ref-guarded once resolved) but slightly exceeds a strict once-per-workspace budget.
- Double-activation registry writes are deduped at the hook layer, so no action needed there.

Not merging. Awaiting Pablo's review.

pablonyx:
Final CI verification at the current head, literal `gh pr checks 1929` output:

```
Analyze (python)	pass
Analyze (rust)	pass
Build SDK	pass
CI/CD config checks	pass
Candidate build handoff	pass
Cargo check & test	pass
CodeQL	pass
Desktop frontend build	pass
Mobile typecheck	pass
Production compose smoke	pass
Repo shape checks	pass
Shared frontend packages	pass
Vercel	pass
Vercel Preview Comments	pass
Web frontend build	pass
```

PR #1932, opened 2026-08-15
fix(product-client): scope transcript submit-stamp to session identity (PRO-175)

pablonyx:
## Summary

Stacked on #1928.

PRO-175's mechanism: `useTranscriptStickToBottom`'s `lastPromptSubmittedAtRef` was scoped to the hook instance, not to session identity. The row lists (`VirtualizedTranscriptRowList` / `FullTranscriptRowList`) never remount across a session switch, so the ref carried the OUTGOING session's stamp into the INCOMING session's comparison. If the incoming session's own current `lastPromptSubmittedAtMs` (e.g. a lingering outbox/optimistic-prompt stamp, unrelated to this visit) was non-null while the ref happened to be `null` (left there by a session with no pending prompt), the effect's `previous == null` branch treated it as a fresh submit and fired a second, redundant `setPinned(true)` + `scrollToBottom()` + `startGlueLoop()` on top of `resetForSession`'s own by-design snap.

**The fix**: added an optional `sessionKey` option to `useTranscriptStickToBottom`, wired by both row lists as `` `${selectedWorkspaceId ?? ""}:${activeSessionId}` ``. The stamp-tracking effect now detects a session-boundary crossing and, on that render, re-baselines `lastPromptSubmittedAtRef` to the incoming session's CURRENT stamp instead of comparing across the switch — critically, it re-seeds to that current value, not `null` (nulling it would make the very next render see `previous == null` and misfire through the other door). `resetForSession` itself is unchanged: it still unconditionally pins/snaps/glues on every session switch by design. The fix only removes the second, redundant re-pin path the stale stamp used to trigger alongside it.

## Deviation from the brief — flagging a contradiction, not silently changing scope

The brief asked me to un-fixme the `revisit-no-motion` Playwright scenario in `qualification/scroll-physics/specs/scroll-physics.spec.ts` and make it pass, extending the fixture driver if needed to recreate the stale-stamp condition.

I did extend the fixture (a `setPendingPromptStamp(sessionId, ms)` driver method plus `main.tsx` wiring an `optimisticPrompt` from it) and reproduced the exact precondition described: session A carries a pending-prompt stamp throughout; session B (visited in between) carries none. I instrumented the hook (temporarily, not in this diff) to confirm two things directly, with the real built dist (not a stale one — `pnpm --filter @proliferate/product-client build` was required for these to even take effect, since the fixture consumes `dist` through the private `#product/*` import map):

- **With the fix disabled** (session-boundary re-baseline turned off), the misfire path DOES fire when returning to a session whose own current stamp exceeds the null baseline left by the intervening session — 0 unit misfires became >0 misfires, confirmed by the new vitest negative control below.
- **In the Playwright fixture**, however, the recorded scrollTop trace across the revisit was IDENTICAL whether the fix was applied or not: `[724,704,836,880,880×36,1012,1056,1056×30]` in both cases (0 stamp misfires logged in both cases too, once instrumented). The residual motion (five distinct values) comes from something else entirely — most likely `resetForSession`'s own always-on snap+glue interacting with the virtualizer's row remeasurement on remount, possibly compounded by `VirtualTranscriptRowList`'s `fallbackReason` resetting to `null` on every session-id change (which can flip the tree from `FullTranscriptRowList` back to `VirtualizedTranscriptRowList` — a real component unmount/remount that would defeat ANY same-instance ref fix, this one included, if a blank-viewport fallback had fired during the intervening session).

The `test.fixme`'s own original comment already named this: "Rung 2 introduces restore-finalized placement; unfixme there" — i.e. the zero-motion requirement depends on the bottom-on-entry replacement the brief itself named as rung 6 and told me NOT to build here ("Note rung 6 later replaces bottom-on-entry for finalized sessions — do NOT do that here"). Making this Playwright test pass would require that out-of-scope work.

I left the Playwright scenario as `test.fixme`, with an updated comment recording exactly what I verified (the stamp leak is not this test's failure mode) so the next rung doesn't have to re-derive it. I reverted the fixture driver/`main.tsx` extensions I'd built to reproduce the precondition, since they end up unused by any committed test — the equivalent, actually-decisive reproduction lives in the new vitest suite instead, which isolates the stamp mechanism from the virtualizer/fallback tangle.

## Testing

**1. Scroll-physics Playwright suite** (`qualification/scroll-physics`, chromium + webkit, `workers=1`), after `pnpm --filter @proliferate/product-client build && pnpm --filter @proliferate/product-client run build:scroll-physics-fixture`:

```
Running 12 tests using 1 worker
  ✓ [chromium] pinned-follow: bottom distance stays ~0 across streaming growth (3.5s)
  ✓ [chromium] unpin mid-stream: reading holds unpinned, no snap-back to bottom (3.1s)
  ✓ [chromium] repin band edge: returning into the bottom band re-pins; staying above does not (3.3s)
  -  [chromium] revisit-no-motion: switching back to a finalized session places with zero motion frames (fixme, see above)
  ✓ [chromium] prepend anchoring: older-history prepend keeps the reading row fixed (1.2s)
  -  [chromium] nested-scroll chaining: ... (fixme, unrelated PRO-258, rung 8)
  ✓ [webkit] pinned-follow: bottom distance stays ~0 across streaming growth (3.9s)
  ✓ [webkit] unpin mid-stream: reading holds unpinned, no snap-back to bottom (3.4s)
  ✓ [webkit] repin band edge: returning into the bottom band re-pins; staying above does not (3.6s)
  -  [webkit] revisit-no-motion: ... (fixme, see above)
  ✓ [webkit] prepend anchoring: older-history prepend keeps the reading row fixed (1.3s)
  -  [webkit] nested-scroll chaining: ... (fixme, unrelated PRO-258, rung 8)
  4 skipped
  8 passed (24.6s)
```

No regression in any of the passing rung-1 scroll-physics invariants.

**2. New vitest unit tests** — `src/hooks/chat/ui/use-transcript-stick-to-bottom.test.tsx`, `submit-stamp session scoping (PRO-175)` describe block, three cases:
- same-session genuine stamp increase still re-pins/snaps,
- a session switch to a session whose own current stamp is HIGHER than the outgoing session's does NOT re-pin (the exact leak), and a later genuine submit in that same session still does,
- a three-session chain (A stamp 500 → B no stamp → C stamp 10) does not misfire on the third switch either, since C's own re-baseline (not a null-comparison) governs.

```
 ✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.test.tsx (23 tests) 27ms
 Test Files  1 passed (1)
      Tests  23 passed (23)
```

Full targeted suite (`pnpm --filter @proliferate/product-client exec vitest run src/hooks/chat/ui`, after the package build above):

```
 Test Files  25 passed (25)
      Tests  141 passed (141)
```

**3. Negative control (mandatory)** — reverted only the fix's session-boundary check (`if (false && previousSessionKey !== undefined && ...)`), reran the same vitest file:

```
 ✓ useTranscriptStickToBottom > submit-stamp session scoping (PRO-175) > re-pins and snaps when the SAME session's stamp genuinely increases
 × useTranscriptStickToBottom > submit-stamp session scoping (PRO-175) > does NOT re-pin on a session switch even when the incoming session's stamp is higher than the outgoing session's
 × useTranscriptStickToBottom > submit-stamp session scoping (PRO-175) > re-baselines to the incoming session's current stamp so a later switch to a THIRD, unrelated session does not misfire either
 Test Files  1 failed (1)
      Tests  2 failed | 21 passed (23)
```

Restored the fix; all 23 pass again (shown above). I additionally ran the same on/off comparison against the Playwright fixture (see the "Deviation" section) — identical traces both ways, which is what led me to the fixture-vs-mechanism conclusion above rather than declaring victory on a misleading pass.

## Observability

No telemetry/logging changes. The fix is purely client-side scroll-pin state; no new spans, metrics, or log lines.

## Files

- `apps/packages/product-client/src/hooks/chat/ui/use-transcript-stick-to-bottom.ts` — `sessionKey` option + session-boundary re-baseline
- `apps/packages/product-client/src/components/workspace/chat/transcript/VirtualizedTranscriptRowList.tsx` — wires `sessionKey`
- `apps/packages/product-client/src/components/workspace/chat/transcript/FullTranscriptRowList.tsx` — wires `sessionKey`
- `apps/packages/product-client/src/hooks/chat/ui/use-transcript-stick-to-bottom.test.tsx` — new PRO-175 test cases
- `apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts` — `revisit-no-motion` comment updated to record the actual (unrelated) cause; left `test.fixme`

## Intra-stack degradation window (PRO-187)

r2 inherits r1's three cadence `test.fixme`s (pinned-follow, unpin mid-stream,
repin band edge) unchanged — the stamp-scope fix here does not touch the follow
cadence, so those scenarios still manifest their WebKit CI failures until their
owning rungs (unpin → r3 #1938; pinned-follow + repin → r4 #1945). See #1928's
degradation-window table for the literal CI numbers.

What r2 additionally fixes: the WebKit older-history **prepend** regression.
Under the honest fixture, wheeling to the top to request older history could
trip the blank-viewport fallback mid-prepend — a real `FullTranscriptRowList` ↔
`VirtualizedTranscriptRowList` remount that produces a new scroll div and drops
the anchor to `scrollTop 0` (CI: `[webkit]` prepend reading-row scrollTop 0).
The fix holds the blank-viewport fallback for a short settle window
(`PREPEND_BLANK_FALLBACK_GRACE_MS = 3_000`) after an anchor-restore write, so a
transient blank during prepend compensation does not remount the list.

**Relationship to rung 10.** This grace window is the minimal correct fix at
this rung. The durable invariant — a reserved-slot / transient-block guarantee
that the list never blank-remounts during a prepend at all — is rung 10's scope.
When rung 10 lands that invariant, the time-based grace window here becomes
redundant and should be superseded by it rather than kept as a parallel guard.

### DIFF
--- a/apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts
+++ b/apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts
@@ -323,11 +323,25 @@ test.describe("transcript scroll physics", () => {
     await drive(page, "finalizeStreamingTurn");
   });
 
-  // EXPECTED TO FAIL today (PRO-175). resetForSession always re-pins and snaps
-  // to bottom on every session switch, then runs a glue loop across the newly
-  // mounted rows, a revisit to a finalized session produces scrollTop motion
-  // frames instead of restoring its prior position with zero visible motion.
-  // Rung 2 introduces restore-finalized placement; unfixme there.
+  // EXPECTED TO FAIL today (PRO-175), for a DIFFERENT reason than the stamp
+  // leak this rung fixes. `resetForSession` still unconditionally
+  // pins/snaps/glues on every session switch by design (rung 2 leaves this
+  // alone), and the virtualizer's own remeasurement of freshly-mounted rows
+  // still produces a handful of settle frames after the snap — plus, in this
+  // fixture, a session switch resets `VirtualTranscriptRowList`'s
+  // `fallbackReason` to null, which can flip the tree from
+  // FullTranscriptRowList back to VirtualizedTranscriptRowList (a real
+  // component remount) if a blank-viewport fallback had fired for the
+  // intervening session. Either path produces multiple distinct scrollTop
+  // frames regardless of the stamp fix below. Verified: with
+  // `setPendingPromptStamp` reproducing the stamp-leak precondition
+  // (session-primary carries a stamp across an intervening session with
+  // none), the pre-fix code and the sessionKey-scoped fix in
+  // use-transcript-stick-to-bottom.ts produce IDENTICAL scrollTop traces
+  // (~5 distinct frames each) — the stamp leak was not this test's failure
+  // mode. Zero-motion placement on revisit is restore-finalized placement,
+  // named in the frozen ladder as rung 6 (bottom-on-entry replacement);
+  // unfixme there.
   test.fixme(
     "revisit-no-motion: switching back to a finalized session places with zero motion frames",
     async ({ page }) => {
--- a/apps/packages/product-client/src/components/workspace/chat/transcript/FullTranscriptRowList.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/transcript/FullTranscriptRowList.tsx
@@ -88,6 +88,7 @@ export function FullTranscriptRowList({
     onScrollSample,
     autoFollowBottomInsetPx: effectiveNonDisplacingBottomInsetPx,
     lastPromptSubmittedAtMs,
+    sessionKey: `${selectedWorkspaceId ?? ""}:${activeSessionId}`,
   });
 
   // Content-search jump-to-match. The full list mounts every row, so the
--- a/apps/packages/product-client/src/components/workspace/chat/transcript/VirtualizedTranscriptRowList.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/transcript/VirtualizedTranscriptRowList.tsx
@@ -74,6 +74,7 @@ export function VirtualizedTranscriptRowList({
     onScrollSample,
     autoFollowBottomInsetPx: effectiveNonDisplacingBottomInsetPx,
     lastPromptSubmittedAtMs,
+    sessionKey: `${selectedWorkspaceId ?? ""}:${activeSessionId}`,
   });
   const renderableRows = useMemo(
     () => buildRenderableRows(rows, isLoadingOlderHistory),
--- a/apps/packages/product-client/src/components/workspace/chat/transcript/VirtualizedTranscriptRowList.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/transcript/VirtualizedTranscriptRowList.tsx
@@ -16,7 +16,7 @@ import { TranscriptFloatingControls } from "./TranscriptRowListShared";
 import { useAboveChangeCompensation } from "#product/hooks/chat/ui/use-above-change-compensation";
 import { useTranscriptStickToBottom } from "#product/hooks/chat/ui/use-transcript-stick-to-bottom";
 import { VirtualTranscriptViewport } from "./VirtualTranscriptViewport";
-import { useTranscriptVirtualizerBlankFallback } from "#product/hooks/chat/ui/use-transcript-virtualizer-blank-fallback";
+import { PREPEND_BLANK_FALLBACK_GRACE_MS, useTranscriptVirtualizerBlankFallback } from "#product/hooks/chat/ui/use-transcript-virtualizer-blank-fallback";
 import { useTranscriptVirtualAnchorCapture } from "#product/hooks/chat/ui/use-transcript-virtual-anchor-capture";
 import { useTranscriptVirtualMeasurementModel } from "#product/hooks/chat/ui/use-transcript-virtual-measurement-model";
 
@@ -55,6 +55,7 @@ export function VirtualizedTranscriptRowList({
   const lastOlderHistoryCursorRequestRef = useRef<number | null>(null);
   const lastPrefetchDecisionLogRef = useRef<string | null>(null);
   const lastBlankReportSignatureRef = useRef<string | null>(null);
+  const prependSettleUntilRef = useRef(0);
   const {
     structural: structuralBottomInsetPx,
     nonDisplacing: effectiveNonDisplacingBottomInsetPx,
@@ -248,6 +249,8 @@ export function VirtualizedTranscriptRowList({
     notifyProgrammaticScroll(() => {
       viewport.scrollTop = anchor.scrollTop + (viewport.scrollHeight - anchor.scrollHeight);
     });
+    // Open the blank-fallback grace window: the anchor write above lands against the still-estimated short scrollHeight.
+    prependSettleUntilRef.current = (typeof performance === "undefined" ? Date.now() : performance.now()) + PREPEND_BLANK_FALLBACK_GRACE_MS;
   }, [notifyProgrammaticScroll, olderHistoryCursor, rows.length, setPinned]);
 
   useEffect(() => {
@@ -361,7 +364,7 @@ export function VirtualizedTranscriptRowList({
     activeSessionId, bottomSpacerHeight,
     firstVirtualItem, lastVirtualItem,
     lastBlankReportSignatureRef, rowCount: rows.length,
-    onFallback, renderableRowCount: renderableRows.length,
+    onFallback, prependSettleUntilRef, renderableRowCount: renderableRows.length,
     scrollRef, selectedWorkspaceId,
     topSpacerHeight, totalContentHeight,
     virtualItemCount: virtualItems.length,
--- a/apps/packages/product-client/src/hooks/chat/ui/use-transcript-stick-to-bottom.session-stamp.test.tsx
+++ b/apps/packages/product-client/src/hooks/chat/ui/use-transcript-stick-to-bottom.session-stamp.test.tsx
@@ -1,0 +1,180 @@
+/* @vitest-environment jsdom */
+
+// Split out of use-transcript-stick-to-bottom.test.tsx (repo-shape max-lines):
+// the submit-stamp / session-identity scoping scenarios (PRO-175) get their
+// own harness (`renderStampHarness`) rather than the shared `renderHarness`
+// used by the rest of the suite, so this file only duplicates the small slice
+// of setup those scenarios actually need.
+
+import { useRef } from "react";
+import { act, cleanup, render } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import {
+  useTranscriptStickToBottom,
+  type TranscriptStickToBottom,
+} from "./use-transcript-stick-to-bottom";
+
+let rafCallbacks: Array<FrameRequestCallback | null>;
+
+beforeEach(() => {
+  rafCallbacks = [];
+  vi.stubGlobal("requestAnimationFrame", (cb: FrameRequestCallback) => {
+    rafCallbacks.push(cb);
+    return rafCallbacks.length;
+  });
+  vi.stubGlobal("cancelAnimationFrame", (id: number) => {
+    rafCallbacks[id - 1] = null;
+  });
+});
+
+afterEach(() => {
+  cleanup();
+  vi.restoreAllMocks();
+  vi.unstubAllGlobals();
+});
+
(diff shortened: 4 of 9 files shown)


pablonyx:
## Dedicated agent review (round 1): CLEAN

Fresh-context adversarial review of the full diff, focused on off-by-one-render reintroductions of PRO-175. Verified:

- Re-baseline correct in both effect orderings: on a session-boundary crossing the ref re-seeds to the incoming stamp and skips the compare that render; equal-or-older stamps never fire, a genuine post-entry submit re-pins exactly once. sessionKey and lastPromptSubmittedAtMs derive from the same atomic state snapshot in use-chat-transcript-view-model.ts, so the dangerous stamp-leads-key split cannot occur.
- sessionKey is correctly in the effect dep array (a null-to-null stamp switch would otherwise leave the boundary ref stale and suppress a later real submit).
- API backward-compatible (optional sessionKey; omitted = byte-identical old behavior); both call sites updated; resetForSession, within-session re-pin, and outbox-drain semantics unchanged.
- The 3 new unit tests structurally depend on the boundary check; the negative control (2 tests fail with the check disabled) is genuine.
- Spec change is comment-only; revisit-no-motion honestly stays fixme with residual motion attributed to resetForSession + remeasure (rung 6's scope).

Non-blocking nits recorded: theoretical colon-collision in the sessionKey format (ids are UUID-shaped, cannot contain ':'), unexercised A-to-B-to-A case (mechanically identical to the covered path).

Review-clean; awaiting founder merge (stacked on #1928).

pablonyx:
Fixed the repo-shape LARGE_FRONTEND_FILE violation on use-transcript-stick-to-bottom.ts (431 lines, no allowlist entry — didn't add one).

Extracted the sessionKey-scoped submit-stamp re-pin effect (PRO-175) into a new colocated hook, use-transcript-submit-stamp-repin.ts, with the minimal `{lastPromptSubmittedAtMs, sessionKey, setPinned, scrollToBottom, startGlueLoop}` contract. useTranscriptStickToBottom's public API is unchanged; the session-stamp test still only reaches it through that public hook.

Verified:
- `python3 scripts/report_frontend_structure.py`: `LARGE_FRONTEND_FILE: 0`, `TOTAL: 0`
- `pnpm --filter @proliferate/product-client exec vitest run src/hooks/chat/ui`: `Test Files 26 passed (26)`, `Tests 141 passed (141)`
- Rebased on r1 (chat-scroll/r1-scroll-physics-harness) after its webkit fix.

pablonyx:
## Final status — scroll-physics green (PRO-187, r2)

Pushed `bb22f4939`. `Transcript scroll physics (chromium + webkit)`: **pass**.

### Root cause fixed here — WebKit older-history prepend dropped to scrollTop 0
Wheeling to the top to request older history could trip the **blank-viewport fallback mid-prepend**: a real `FullTranscriptRowList` ↔ `VirtualizedTranscriptRowList` remount (new scroll div) that discards the anchor and lands the reading row at `scrollTop 0` (CI: `[webkit]` prepend `after.scrollTop` Expected > 150, Received 0). Fix: hold the blank fallback for a short settle window (`PREPEND_BLANK_FALLBACK_GRACE_MS = 3_000`) after an anchor-restore write, so a transient blank during prepend compensation cannot remount the list. Negative-controlled in the hook's vitest (disabling the grace bail re-reds the suppress test).

**Rung-10 relationship:** this time-based grace is the minimal correct fix at r2; rung 10's reserved-slot / transient-block invariant is the durable guarantee that supersedes it. Detail in the body's 'Intra-stack degradation window' section.

### CI flake note (honest record)
The r2 webkit prepend is timing-sensitive: this exact tip passed, then flaked red once (identical byte-for-byte product code, verified via `git diff`), then passed again on rerun. r4/r5's sturdier compensation machinery removes the fragility; r2 relies on the grace window alone. Final verdict: **pass**.

### Inherited cadence fixmes
r2 does not touch follow cadence, so it inherits r1's three cadence `test.fixme`s unchanged (unpin → r3 #1938; pinned-follow + repin → r4 #1945).

pablonyx:
Adversarial delta review: CLEAN

Scope: delta since the prior clean review, at head bb22f4939 (PRO-175 stamp-scope fix, stacked on #1928).

1. Inherited r1 deltas: confirmed present unchanged (turn_ended fixture, fixme table, degradation-window note).

2. repin-band-edge fixme: PR body says r2 inherits it unchanged, un-fixme'd at r4 (#1945). Consistent with what I verified directly against r4's branch content in the #1928 review (repin band edge is `test(...)` there, not fixme). No contradiction.

3. PREPEND_BLANK_FALLBACK_GRACE_MS = 3000ms window. Read the actual hook change in use-transcript-virtualizer-blank-fallback.ts: the suppression is time-gated but only ever armed by a real event, the anchor-restore write in VirtualizedTranscriptRowList's prepend handler sets `prependSettleUntilRef.current = now + GRACE_MS` right after `viewport.scrollTop = anchor...`; it is 0 (inert) otherwise. During the window, blank detection is skipped and `suspicionRef` is reset each frame (re-arms immediately once the window lapses), it is not a blanket disable. Two new unit tests in use-transcript-virtualizer-blank-fallback.test.tsx confirm both directions: (a) suppressed while `prependSettleUntil` is in the future under the identical DOM-confirmed-blank condition that trips fallback in the pre-existing test, and (b) negative control: an already-lapsed grace (`performance.now() - 1`) reproduces the same blank_viewport fallback as the ungated path, proving the suppression is scoped to the live window and not a masked-forever regression. This is an honest, narrowly-scoped mitigation, not a bug-hiding timeout; the PR body itself flags the known tradeoff (a genuine blank coincident with a prepend within the 3s window would be masked) and explicitly defers the structural fix to rung 10, calling out that this window should be superseded (not kept as a parallel guard) once rung 10 lands.

4. Stamp-scope fix (`sessionKey` re-baseline in use-transcript-stick-to-bottom.ts): logic and the three new vitest cases (same-session genuine increase still re-pins; switch to higher-stamped session does not misfire; three-session chain A->B->C does not misfire on the third switch) match the PR body's described mechanism and its mandatory negative control (disabling the session-boundary branch reproduces 2 failing tests, re-enabling restores 23/23 passing), which is quoted verbatim in the body.

5. CI: "Transcript scroll physics (chromium + webkit)" is literally `pass` in `gh pr checks 1932` (2m31s), and all other listed checks are `pass` with no pending/failing entries.

No evidence of a real blank being permanently masked, no scope-widening beyond the documented prepend-settle path, no weakened assertions.

pablonyx:

Repo checked out via a temp worktree off `origin/chat-scroll/r2-pro175-stamp-scope`, `pnpm install` + `pnpm shared:build` run there. Touched files per `gh pr diff 1932 --name-only`: `FullTranscriptRowList.tsx`, `VirtualizedTranscriptRowList.tsx`, `use-transcript-stick-to-bottom.ts`, `use-transcript-submit-stamp-repin.ts`, `use-transcript-virtualizer-blank-fallback.ts`, their vitest suites, and the qualification physics spec.

### Scoped vitest (3 files / 28 tests, all pass)
```
✓ src/hooks/chat/ui/use-transcript-virtualizer-blank-fallback.test.tsx (5 tests) 28ms
✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.session-stamp.test.tsx (3 tests) 15ms
✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.test.tsx (20 tests) 27ms

 Test Files  3 passed (3)
      Tests  28 passed (28)
```

### Negative control
Reverted the PRO-175 `sessionKey` stamp-scoping guard in `use-transcript-submit-stamp-repin.ts` by removing the session-boundary early return:
```
if (previousSessionKey !== undefined && previousSessionKey !== sessionKey) {
  return;
}
```
Reran `use-transcript-stick-to-bottom.session-stamp.test.tsx`, literal failure:
```
FAIL  src/hooks/chat/ui/use-transcript-stick-to-bottom.session-stamp.test.tsx > useTranscriptStickToBottom > submit-stamp session scoping (PRO-175) > re-baselines to the incoming session's current stamp so a later switch to a THIRD, unrelated session does not misfire either
- Expected
+ Received
- false
+ true

 Test Files  1 failed (1)
      Tests  2 failed | 1 passed (3)
```

Restored the guard via `git checkout --`; `git status` confirms the worktree is clean, no diff remains.

### Browser-level physics (CI is truth for this tier, no local Playwright run)
```
Transcript scroll physics (chromium + webkit)	pass	2m31s
```
from `gh pr checks 1932`.

PR #1925, opened 2026-08-15
feat(agent-auth): two-tier probing with retuned constants (FR-2, rung 3)

pablonyx:
## Summary

Rung 3 of the Agent Auth ADR ladder (PRO-214, FR-2), stacked on #1916 (rung 2). Two-tier probing:

- Tier-1 instant credential trial (~1s) behind config flag `tier1_trial_enabled` (default OFF): gateway sources are trialled via the existing key-scoped `GET /v1/models` (2xx = green `Tier1Trial` evidence with age, 401/403 = Expired, anything else inconclusive and records nothing). Pasted api_key and native CLI logins deliberately stay heuristic/unverified (no verifiably free trial endpoint invented; native spawns can raise keychain prompts). Trials are event-driven off the same poke events as the probe, single-flight with a panic-safe RAII slot guard, and verdicts are fingerprinted by the credential they verified: key rotation or a source switch invalidates the stored verdict, and a green trial can never upgrade a non-gateway credential.
- Retuned constants (A5): per-probe timeout 240s -> 45s; spawn failures fast-fail with stable code `spawn_failed` instead of waiting out the timeout; backoff ceiling 6h -> 30min (60s initial, doubling, saturating).
- Probe lifecycle is now user-visible: the rung-2 `authState` facts carry real Idle/Queued/Running/Backoff, last-success age, last-failure detail, and `next_attempt_at`.
- Dead `model_switch_timeout` config field deleted; `AUTO_PROBE_EXCLUDED_HARNESSES` doc comment corrected (cursor is excluded for the unattended-keychain-prompt reason; the refuted CURSOR_API_KEY claim is removed).

No wire/SDK changes beyond filling rung-2 fields; behavior with the flag off is unchanged except the now-real probe-lifecycle facts.

## Testing

- `cargo test -p anyharness-lib model_snapshot::`: `test result: ok. 56 passed; 0 failed` (trial verdict mapping via raw-socket HTTP stub, flag gate, single-flight coalescing, panic self-heal, rotated-key invalidation, source-switch clear, spawn fast-fail, 30-min backoff ceiling).
- `cargo test -p anyharness-lib auth_state`: `test result: ok. 29 passed; 0 failed` (green trial -> Authenticated with evidence; green never upgrades a non-gateway credential; Expired on trial auth failure).
- `cargo check -p anyharness-lib`: clean (warnings pre-existing).
- `python3 scripts/check_docs.py`: passed.

## Observability

[...]

PR #1916, opened 2026-08-15
feat(agent-auth): canonical evidence state model (FR-1, rung 2)

pablonyx:
## Summary

Rung 2 of the Agent Auth ADR PR ladder (PRO-214, FR-1). Adds the canonical per-harness agent-auth state as a struct of orthogonal facts plus the single shared derivation `derive_agent_auth_state(facts) -> (display, next_action, evidence_ref, evidence_age)`, computed additively alongside the existing `CredentialState`/`ResolvedAgentStatus` ladders and serialized as an optional `authState` field on the `GET /v1/agents` projection. Panes are NOT switched in this rung; no existing wire field changes.

Hard invariant (kills the PRO-252 false-green class by construction): `Authenticated`/`Usable` are reachable only with an `evidence_ref` naming a probe observation or key-scoped gateway check plus a non-null `evidence_age`. Bare file/keychain presence never renders green.

Vocabulary precedence (first match wins): NotInstalled, Unsupported, Misconfigured, Expired, Unavailable, Probing, Usable, Authenticated, Selected, Installed. Gateway-health and handoff fact slots are typed now and filled by rungs 4 and 6.

## Testing

- `cargo test -p anyharness-lib auth_state`: `test result: ok. 25 passed; 0 failed` — includes an invariant permutation sweep (>10k fact combos) asserting every green output carries dated evidence and that BarePresence/AcknowledgedRoute can never be green.
- `cargo test -p anyharness-lib agents_contract`: `test result: ok. 1 passed; 0 failed`.
- `cargo check -p anyharness-contract` / `-p anyharness-lib`: clean (warnings pre-existing, none in new files).
- `python3 scripts/check_docs.py`: `Documentation integrity check passed (213 Markdown files).`

## Observability

No new telemetry in this rung; the `agent_auth_green_without_evidence_total` counter lands with the UI switch rung where a violation could be observed in the wild. The derivation itself is pure and invariant-tested.

## Docs

`specs/FEATURE_DOCS/AGENT_AUTH.md` gains "The canonical evidence model" section (current runtime behavior; panes still render the legacy ladder until the UI rung).

## Notes

- Lint record deferral: the FR-1 enforcement artifact (evidence-only-green rule record) is deferred until PR #1722 (`lints/` record format, rung 0) merges; it will be authored as a `lints/` record in a follow-up, not in the legacy txt-allowlist format.
[...]

PR #1933, opened 2026-08-15
feat(product): cloud culling rung 3 — pickers, palette, shortcuts, shared handlers (PRO-10)

pablonyx:
Rung 3 of the Cloud Culling PR ladder (PRO-10), stacked on #1921 (rung 2). Removes cloud rows/cases from every shared launch/creation surface on desktop, gated on local-arm tests (FM2) and host-capability gating for web-shared branches (Q3, web exempt from the cull).

## Scope
- **Pickers (Q3/FM8)**: `HomeTargetPicker` desktop offers local/worktree, coerces stale `cloud` launch kind to `worktree`, never returns empty or `["cloud"]`; web keeps its cloud arm via `desktopTargetsAvailable`.
- **Commands/shortcuts/palette**: `newCloud` shortcut + groups entry, `workspace.new-cloud` palette entry, `newCloudWorkspace` app-command action and all consumers deleted.
- **Sidebar/handlers**: cloud affordances in `SidebarWorkspaceContent` gated web-only; `onAddToThisMac` removed end-to-end; archive/restore cloud arms kept dormant (web-reachable, dead on desktop via the rung-1 filter).
- **Automations**: new-cloud-target creation rows removed; saved cloud rows preserved + badged per FR-2/FM5.
- **Copy**: automation/home/auth/capability/AccountPane/remote-access/desktop-worker copy culled or reworded (Q7; credential-sync mentions kept per ruling).
- **Area F (host-capability seams, zero deletions — every candidate keeps web-reachable importers)**: desktop never mounts `HarnessPane`'s cloud surface (Cloud|Local toggle host-gated off); repo settings force `context = "local"` on desktop with the toggle hidden, making `RepoCloudGate`/`RepoCloudAuthorizationRequired` (all GitHub App install/authorize affordances, FR-3) unreachable on desktop while hooks stay dormant; add-repo flow: desktop never gets the cloud step, web keeps `["cloud"]` as its sole option (never empty, FM8 analog).

## Gates (literal)
product-client: `Test Files 929 passed (929) / Tests 6628 passed (6628)` on the rebased head; typecheck clean. Desktop `tsc --noEmit` exit 0. Web `tsc --noEmit` exit 0 (shared components edited; web arm proven compiling). `check_max_lines.py` pass.

## Notes for rung 6 (acceptance sweep)
- `RepoScopeStates.tsx` `RepoScopeEmptyState` still says "…runs in Proliferate Cloud" (zero-repo desktop state) — flagged for the sweep.
[...]

PR #1921, opened 2026-08-15
feat(product): cloud culling rung 2 — delete leaf cloud-only surfaces (PRO-10)

pablonyx:
Rung 2 of the Cloud Culling PR ladder (PRO-10), stacked on #1915 (rung 1). Deletes leaf cloud-only surfaces already made unreachable by the rung 1 gate; no user-visible change beyond dead-code removal.

## Scope shipped
- Deleted `CloudEnvironmentConfigSection.tsx` (only external reference was its own test) and `use-cloud-workspace-polling.ts` + call site in `AuthenticatedLaunchLifecycles.tsx` (registry-wide cloud provisioning poll; unreachable with the gate permanently on).
- Trimmed the component-specific tests with their surfaces.

## Scope deferred to rung 3, with evidence (not blind caution)
The inventory listed more leaves, but implementation found them entangled with picker/surface plumbing that rung 3 owns:
- `CloudGuard` + its 4 panes: live branches in `HarnessPane.tsx` (`surface === "cloud"`) and `render-settings-section.tsx` `cloudGate`/`authGate`, which also gate control-plane sections.
- `RepoCloudGate` / `RepoCloudAuthorizationRequired`: tied to the repo Cloud/Local context tab (`RepoScopeHeaderControls`, `RepoEnvironmentPane`, `RepoConfigurePane`, `RepoActionsPane`).
- `CloudRepoPicker(+Blocker)` + AddRepoFlow cloud step: removing `"cloud"` from `AddRepoFlowOption` leaves web with an empty add-repo menu (the FM8 failure mode); needs Q3 host-capability gating.
- `cloud-status-copy.ts`: consumed via `cloud-workspace-status-presentation.ts` by `MainSidebar` and composer-blocked-state (rung 3 files).

## Open founder questions
- `SettingsCloudRedirect.tsx` (`/settings/cloud`) is the Stripe billing checkout/portal return route (`STRIPE_CHECKOUT_SUCCESS_URL` etc. point at it in prod). Billing is out of scope per the ADR non-goals, so it was NOT deleted despite being a cull row in Context Doc 04; recommend striking that inventory row (rename of the route can ride the billing ADR).
- `render-settings-section.tsx` computes `cloudGate.cloudActive` from the compute-disable flag, so control-plane settings sections partly ride the cloud-workspace gate — a Q9/FM6-shaped wrinkle rung 3 must untangle (flagged, not changed here).

## Gates
[...]

PR #1935, opened 2026-08-15
feat(agent-auth): gateway verification planes (FR-3, rung 4)

pablonyx:
## Summary

Rung 4 of the Agent Auth ADR ladder (PRO-214, FR-3), stacked on #1925 (rung 3). Two read-only verification planes, both default-off, feeding the FR-1 facts struct — this is the structural fix for the PRO-238 class ("gateway sometimes doesn't enable models").

Control plane (server): a periodic asyncio loop (`agent_gateway/worker.py` triad convention, gated by `agent_gateway_verification_enabled` default False + `agent_gateway_verification_interval_seconds` default 900) diffs each active enrollment key's key-scoped `GET /v1/models` against the expected access-group set parsed from the reviewed source of truth `server/litellm/config.yaml`. Exact set match = `ok`; anything else = `misconfigured` with a real delta `{missing, extra, observed_count, expected_count}` — this catches both wrong-group keys and the stale-deployed-image drift (repo says X, proxy serves Y). Config-unavailable degrades to the non-empty heuristic with a recorded degraded marker; the loop never crashes and one bad key never aborts a tick. Verdicts persist as three nullable columns on `agent_gateway_enrollment_key` (alembic `e7f1a3c9d20b`, single head preserved) and surface additively on the capabilities response. state.json and its pinned contract fixture are untouched.

Data plane (runtime): a gateway health contract with three distinct verdicts riding the SAME `/v1/models` fetch as the tier-1 trial (one hit per poke, flag `gateway_health_enabled` default false): `Unreachable` -> Unavailable, `Unauthorized` (401/403) -> Expired, `ModelsDrifted` (sorted-model-id hash vs previous observation) -> Misconfigured. Health verdicts fold only onto gateway-sourced credentials, never force green, and clear on source switch and key rotation (rotation resets the drift baseline so a rotated key cannot false-pulse ModelsDrifted). Contract enum extended additively; SDK regenerated.

Opencode staleness: verified the existing probe-trigger plan invalidation + blocking re-fetch already satisfies "re-fetch at every probe trigger"; no change made, no ETag mechanism built.

## Constitution flags (test + ledger edits, founder review)

- `the_global_cap_bounds_concurrency_across_harnesses` (anyharness concurrency test) was EDITED in cd60fce82 to fix a scheduler-timing flake CI hit on this stack (observed 8 vs expected 4; unreproducible locally under three schedulers; the rung-4 diff is provably inert to the coalescing path). The fix parks the first fake-runner attempt on the sole semaphore permit so the race window is closed deterministically; every assertion is unchanged, and the negative control still fails (12 vs 4) when the cap is removed. Plan of record: this deterministic-ization IS the fix (no follow-up needed); if the founder prefers the original timing-based form, revert cd60fce82 and accept the flake rate.
- Three max-lines allowlist ledger entries bumped with appended reasons (`config.py` 614->620, `agent_gateway.py` 643->651, `service.py` 813->834); registry-growth files were already marked split-pending.
- mypy ratchet: `VerificationDelta` alias module added to the EXISTING pydantic-plugin false-positive override list (net baseline SHRANK by 20 entries; no suppression added).

## Testing

- Rust: `cargo test -p anyharness-lib trial` 25 passed; `auth_state` 31 passed; `model_snapshot` 66 passed; `anyharness-contract` 46 passed; 0 failures (stubbed-proxy verdict tests, rotation-baseline reset, derivation arms, single-fetch flag matrix).
- Python: `pytest tests/integration/test_agent_gateway_verification.py -q` 8 passed (exact-match ok, missing-delta, extra-delta, degraded fallbacks, error-no-overwrite, key-redaction on the REAL logging path, flag gating); implicated neighbors (`enrollment_keys`, `api`, `key_lifecycle`, `schema_migrations`, `usage_import`, `litellm_config_access_groups`) all pass.
- Negative controls run and quoted by the implementer: the extra-model test fails on the pre-fix non-empty heuristic; the redaction test fails when reporting from the raw except block.
- `ruff`, `scripts/check_max_lines.py`, `scripts/check_docs.py`: all pass.

## Observability

Verification verdicts carry structured deltas; report_critical usage is key-redacted on BOTH sinks (Sentry object and the ambient-exc_info log path, re-raised `from None`); tags carry harness_kind + enrollment key id only, never key material.

## Docs

`specs/FEATURE_DOCS/MODELS.md` gateway-contract section (two planes, verdicts, expected-set semantics, flags).

## Notes

- Review history: initial adversarial review found 1 major (verdict was non-empty-vs-empty rather than the spec's expected-set diff) + 2 minors (rotation drift-baseline, redaction); a second pass caught the ambient-exc_info log leak. All fixed with pinning tests; final verdict CLEAN.
- Three max-lines allowlist ledger entries bumped with appended reasons (`config.py` 614->620, `agent_gateway.py` 643->651, `service.py` 813->834) — registry-growth files already marked split-pending. Flagged per the constitution.
- Drift semantics: baseline = previous observation, so ModelsDrifted is sticky until the next poke, then re-baselines. A sticky-until-stable design was considered and deferred as follow-up.
- `parents[4]` path anchor for config.yaml is documented and fail-soft (degrades, never crashes); a package-root marker anchor is a possible follow-up.

## Open founder questions

None.

Stacked on #1925 -> #1916; merge order bottom-up; merges remain Pablo's.


pablonyx:
Dedicated adversarial review (fresh-context agent, full diff vs rung-3 base): round 1 FIX-REQUIRED (1 major: verdict was non-empty-vs-empty, not the spec's expected-access-group-set diff, so a wrong-but-populated group read ok; 2 minors: rotation drift-baseline false pulse, exception redaction). Round 2 caught a residual conditional-major: report_critical's logger.exception pulled the ambient UNREDACTED exception into the log sink. All fixed in 30c82847e + aa5a4af3b with pinning tests verified to fail on the old code (negative controls quoted). Final verdict: CLEAN.

pablonyx:
CI repair: (1) ruff format on two files (d430182d7); (2) cloud/sdk regenerated for the new capabilities verifications fields via make cloud-client-generate, idempotence verified (815b4af03); (3) the failing coalescing test the_global_cap_bounds_concurrency_across_harnesses was diagnosed as timing-fragile, not a rung-4 regression: this rung's diff does not touch probe_on_event/run_attempt/coalescing, the failure is unreproducible under current_thread/multi_thread/yield-perturbation locally, and a late-stamped poke after the winning attempt completes legitimately earns a fresh probe under CI load. Fix (cd60fce82) parks the first fake-runner attempt on the sole semaphore permit so all pokes stamp before any attempt completes; assertions unchanged (peak==1, count==4); negative control (disable attempt_covers) fails 12 vs 4. CONSTITUTION FLAG for founder review: this edits a test to make it deterministic; assertions were not weakened, but calling it out explicitly per the making-CI-green rule.

pablonyx:
Second CI repair (4f7b72594): mypy ratchet diagnostics fixed at source (field-name construction, precise VerificationDelta type alias, types-PyYAML dev dep). One item flagged per constitution: the new models module was added to the EXISTING pydantic-plugin false-positive override list (joining ~a dozen sibling *.models entries) for a Field(...) plugin FP; the overall mypy baseline SHRANK by 20 diagnostics. Ratchet output: 'Mypy diagnostic ratchet passed (503 existing diagnostics).'

pablonyx:
ORCHESTRATOR STATUS (do not merge yet): the 6 failing checks on the current merge ref share one root cause. Repo-shape says: "down_revision points at unknown revision id(s): a3f5c8e91d2b". This PR's migration chains onto a revision that is not in main's migration graph after the out-of-order merge train (#1916/#1925 merged, #1921 re-cut as #1953, #1933/#1919 closed). Every server-booting lane (test, intent-tests, intent-billing, workflow lifecycle, compose smoke) fails fast on the broken alembic graph, so one fix should re-green the set: repoint this migration's down_revision at current main's head revision and forward-merge up the stack (#1939 -> #1941 -> #1943).

Owner agent is temporarily unreachable (workspace control plane down); it will pick this up on wake, or the fix can be made by hand. The stack above this PR shows the same cascade for the same reason.

pablonyx:
**ORCHESTRATOR UPDATE — supersedes my 20:29Z do-not-merge comment.** The alembic root cause is FIXED: the stale reference to a3f5c8e91d2b (a revision that lives only on unmerged #1940) was purged from all five agent-auth branches; e7f1a3c9d20b is re-parented onto main's real head c7d9e1f3a5b7, forward-merged up the stack (#1935 → #1939 → #1941 → #1943), and validated (alembic heads prints exactly one head; scratch-DB migration suite 7/7). CI is re-attesting; the owning agent will post literal check quotes at settle.

**Merge-order caution that remains:** this stack's e7f1a3c9d20b and #1940's catalog_version migration BOTH now parent on c7d9e1f3a5b7. Merging #1940 and this stack in either order re-collides into two alembic heads — whichever merges second needs its migration re-parented onto the other's revision first. One-commit fix, but it must happen at merge time.

pablonyx:
## CI attestation (post alembic re-parent repair)

Literal `gh pr checks` at current head:

```
PR #1935: 27 checks — pass=25 fail=0 pending=0 skipping=2
```

Notes on this head's CI history: the migration re-parent onto main's live alembic head `c7d9e1f3a5b7` fixed the unknown-revision failure on all server-booting lanes; one `Cargo check & test` job then hung in_progress for 3+ hours and was cancelled and rerun (passed); one `Shared frontend packages` failure was a flaky `ComposerCommandEditor.test.tsx` waitFor timeout in code untouched by this diff (identical merged content passed the same lane on #1939) and passed on rerun. All non-skipped checks pass (skips are path-filters). Nothing merged.

PR #1938, opened 2026-08-15
fix(product-client): single scroll writer with explicit ownership markers (PRO-187)

pablonyx:
Stacked on #1932.

Rung 3 of the Chat Scroll + Positioning ladder: make the transcript a single scrollTop writer and classify scroll events by explicit ownership markers rather than pixel-match heuristics (founder ruling FR-1).

## Classification ladder (new)

Layered in `use-transcript-stick-to-bottom.ts`, most-authoritative first:

1. **Primary — ownership markers.** Every programmatic write (`notifyProgrammaticScroll` / `scrollToBottom` / glue ticks) records a queued marker carrying the scrollTop it produced. The scroll event that matches a live marker within `PROGRAMMATIC_MATCH_TOL_PX` is ours; it clears that marker and never touches pin state. A **queue** (not a single slot) is load-bearing: the glue loop writes faster than the browser dispatches scroll events, so several writes are in flight at once. A single slot let a later write overwrite an earlier marker before its event arrived, so the stale event was misread as a user scroll and dropped the pin. Each marker carries a one-frame **watchdog rAF** so a clamped or no-op write cannot leak into the next user scroll.
2. **User — input intent.** The input-intent listeners remain the sole source of "user": upward intent claims ownership and unpins synchronously, before the browser scroll event. Because that flips `pinnedRef` first, a user scroll wins by construction. The primary marker match (tier 1) is still consulted before intent on every event, which is safe only because the two never target the same scrollTop: markers are recorded at the bottom the write settled on, while an unpinning scroll lands away from any marker's `expectedTop`, and the intent-claimed unpin itself already happened at input time (tier 2, above) regardless of how the later scroll event gets classified.
3. **Fallback (last resort, gated behind a live marker).** The surviving 2px tolerance and the H2 downward-while-pinned rule are retained only for the case a marker is live but its exact landing could not be matched (scrollHeight changed between write and event). Clearly layered as such in code.

An unattributed scroll (no live marker, no recent intent) runs the unchanged pin logic and is kept distinguishable (`userInitiated` absent) for a later rung.

`overflow-anchor: none` is applied to the transcript viewport via a new opt-in `stableScrollAnchor` prop on the shared `AutoHideScrollArea` (its default is unchanged for other consumers), set at both transcript usage sites, so Chromium and WKWebView run the identical single-writer regime and native scroll anchoring cannot write scrollTop underneath the engine.

## What deliberately did NOT change

Classification + anchoring only. Untouched: repin band semantics (24px, direction-not-upward), consumed-inset bookkeeping, glue termination frame counts, `resetForSession`, the rung-2 submit-stamp session scoping, the reveal animation, and virtualizer behavior.

## Testing

Unit (vitest, colocated) — `pnpm --filter @proliferate/product-client exec vitest run src/hooks/chat/ui`:
```
Test Files  25 passed (25)
      Tests  145 passed (145)
```
Includes 4 new marker-lifecycle tests: multiple in-flight markers consumed in order, marker-tolerance-miss classified programmatic, expired marker does not misclassify the next user scroll, fallback tier engages only when a marker is live. `AutoHideScrollArea.test.tsx`: 3 passed.

**Negative control.** Reverted the marker mechanism to single-slot (`pendingMarkersRef.current = []` before each push, reproducing the overwrite bug) and ran the multi-marker test:
```
- Expected  + Received
- true       + false
  > expect(handle.current.api.isPinnedToBottom).toBe(true);   // line 571
 Tests  1 failed | 26 skipped (27)
```
Restored and re-verified green (27 passed).

Scroll-physics Playwright (`test:scroll-physics`, chromium + webkit, headless, workers=1). Three new scenarios, run in isolation:
```
✓ [chromium] single-writer: overflow-anchor none
✓ [chromium] no-false-unpin: rapid glue writes during growth never unpin
✓ [chromium] swallowed-user-scroll: wheel-up during heavy programmatic snap wins
- [webkit]   single-writer: overflow-anchor none   (skipped: property is Chromium-only)
✓ [webkit]   no-false-unpin
✓ [webkit]   swallowed-user-scroll
5 passed, 1 skipped
```

**Environmental note (local machine, not CI).** This shared dev box (18-core/24GB, running two browser engines + the fixture concurrently) starves the rAF-driven glue loop under full-suite load. On the **base branch r2, unchanged**, three pre-existing timing-sensitive scenarios (`pinned-follow`, `unpin mid-stream`, `repin band edge`) already fail here on both engines — verified by stashing this rung's changes, rebuilding dist, and re-running. This PR introduces no new persistent failures: the three new scenarios pass in isolation on both engines and under full-suite load; the same pre-existing scenarios remain the only red. The new `no-false-unpin` scenario asserts a load-invariant ratio (resting gap as a fraction of total content growth) rather than an absolute pixel bound, and the precise marker lifecycle is proven deterministically by the unit tests + negative control above. CI (unloaded) is the authority for the physics gate.

## Observability

No telemetry or scrubber sources changed. The `onScrollSample` perf-probe contract is preserved: it fires on every scroll with correct programmatic/user classification across all ladder tiers.

## Deferred

Single-writer ownership lint record: the `lints/<owner>/*.toml` record system (PR #1722, branch `docs-v1/d6-lints`) is unmerged, and the constitution forbids new legacy txt-allowlist entries. The lint record will be authored as a follow-up once #1722 lands.

## Open founder questions

- The CI-regression fix (129b63d94) replaced the legacy H2 downward-while-pinned heuristic with a content-size-change discriminator in the fallback tier: a scroll event that coincides with a scrollHeight change while pinned is classified as our own follow (user scrolls never change scrollHeight), with an at-hard-bottom guard for shrink/clamp. This is construction-based and consistent with FR-1 (markers stay primary, intent stays the sole user source), but it is an evolution of the ADR wording that kept H2 as the named fallback. Flagging for explicit sign-off.

- Review round 4 (CLEAN, non-blocking): the 500ms wall-clock compensation deadline has no cancel-on-upward-user-intent path, so a user scrolling up while unpinned during a completed-turn split above them can be re-anchored each frame for up to 500ms. Same clobber shape pre-existed with quiet-frames termination, but the deadline widens the window. Options: cancel compensation on upward intent, or terminate on min(settled, deadline).
## Intra-stack degradation window (rungs 3-4 coupling)

CI proved that the single-writer pin decision on this rung alone cannot hold the
follow cadence on slow CI runners without the rung-4 same-frame pipeline
(#1945). Literal CI at r3:

- `pinned-follow`: bottomDistance 132 (expected <= 120) on BOTH chromium and
  webkit.
- `repin-band-edge`: fails on webkit (bottomDistance 332, expected <= 202 after
  the repin-and-stream arm).

Literal CI at r4 (#1945): both `pinned-follow` and the webkit `repin-band-edge`
PASS — the synchronous ResizeObserver-notify snap is exactly what closes this
gap.

Because the ADR's own revert ordering couples rungs 3 and 4 (the single-writer
decision and the same-frame snap pipeline ship and revert together), these two
scenarios are marked `test.fixme` AT THIS RUNG ONLY, each with a comment naming
#1945 as the un-fixme point. Thresholds are NOT loosened. The fixmes exist only
in this branch's tree; #1945's diff removes them again (they are active on r4 and
r5). This is a temporary intra-stack degradation window, not a shipped
regression: the window closes the moment rung 4 lands on top.


pablonyx:
## Dedicated agent review (round 1): FIX-NEEDED (test-quality only, no implementation blockers)

Fresh-context adversarial review of the classification ladder. Implementation verified clean: value-matched marker drain with bounded queue growth (glue early-returns without pushing a marker at target), fallback tier correctly nested and pinnedRef-gated (cannot re-become primary), watchdog-expired late programmatic events are downward snaps that re-pin rather than false-unpin, stableScrollAnchor is opt-in and lands on the actual scrolling element at both transcript sites, zero out-of-scope drift, old classification path fully deleted.

Findings to fix before review-clean:

1. **[should-fix] Watchdog-expiry unit test is vacuous.** The post-expiry userScroll targets 500 vs marker expectedTop 700, so it unpins whether or not the watchdog fired (misses tolerance AND H2 either way). Restructure as a contrast test: same near-marker target swallowed as programmatic while the marker is live, unpins only after the rAF expiry round.
2. **[should-fix] Two marker tests under-test their stated semantics.** "Consumes in order" never exercises out-of-order or coalesced delivery (two writes, one event; reversed arrival). "Fallback only when marker live" uses an UPWARD no-marker scroll, which never reaches the H2 branch (downward-while-pinned); it must prove a DOWNWARD no-marker event escapes the fallback.
3. **[nit, observation only]** The fallback uses the newest pending marker's expectedTop as the H2 threshold; two live markers spanning more than the tolerance leave a gap where an own-write event could false-unpin. Matches pre-existing single-marker H2 semantics, so recorded as an observation, not a change request.
4. **[nit]** PR-body prose says user intent wins by construction, but the primary marker match returns early without consulting intent; practically nil impact (markers target bottom), worth one clarifying sentence in the body.

Fixes will land on this branch; re-review to follow.

pablonyx:
Review round 1 findings addressed:

| Finding | Resolution | File(s) |
| --- | --- | --- |
| 1. Watchdog-expiry test vacuous (unpinned regardless of marker state) | Replaced with `"classifies a near-marker event as programmatic while live, and lets it fall through once expired"` — two independent harnesses, identical setup and identical near-marker event (`scrollTop = 650` against a marker recorded at 650), differing only in whether the watchdog rAF round was flushed first. Live arm: `{ programmatic: true }`, pin held. Expired arm: `{ programmatic: false }`, unpins (distance 50px > `REPIN_BOTTOM_THRESHOLD_PX` 24px). No user intent claimed in either arm, so marker liveness is the only variable. | `use-transcript-stick-to-bottom.markers.test.tsx` |
| 2. "Consumes multiple in-flight markers in order" never exercised out-of-order/coalesced delivery | Renamed to `"consumes multiple in-flight markers by value, including coalesced and reversed delivery"` — two writes (650, 700), only ONE scroll event dispatched at the coalesced final position (700); asserts `findIndex`-on-`expectedTop` picks the correct marker, the leftover (650) expires via the watchdog, and a subsequent user scroll is not misclassified. Added a second test, `"matches queued markers by value even when events arrive in reverse write order"`, that dispatches the 700-marker's event before the 600-marker's event and asserts both still classify correctly — pinning the queue's value-match contract (not FIFO). | `use-transcript-stick-to-bottom.markers.test.tsx` |
| 3. "Fallback engages only when a marker is live" tested an upward no-marker scroll that never reached the H2 branch | Replaced with `"engages the fallback tier only when a marker is live (downward-while-pinned)"` — contrasts a downward-while-pinned scroll (`scrollTop` 700→705) with NO live marker (must not be attributed to H2; asserts `{ programmatic: false }`) against the same shape of event WITH a live marker whose tolerance missed due to scrollHeight growth (IS caught by H2; asserts `{ programmatic: true }`). | `use-transcript-stick-to-bottom.markers.test.tsx` |
| 4. PR body nit: primary-marker-precedes-intent safety claim needed a clarifying sentence | Added: "The primary marker match (tier 1) is still consulted before intent on every event, which is safe only because the two never target the same scrollTop: markers are recorded at the bottom the write settled on, while an unpinning scroll lands away from any marker's `expectedTop`, and the intent-claimed unpin itself already happened at input time (tier 2, above) regardless of how the later scroll event gets classified." | PR body (`gh pr edit 1938`) |

Also split the `ownership markers (PRO-187)` describe block into its own sibling file, `use-transcript-stick-to-bottom.markers.test.tsx` (mirroring the existing PRO-175 `use-transcript-stick-to-bottom.session-stamp.test.tsx` split), to keep the main test file under the repo's 600-line cap while adding the new contrast coverage.

**Verification**

`pnpm --filter @proliferate/product-client exec vitest run src/hooks/chat/ui`:
```
 Test Files  27 passed (27)
      Tests  146 passed (146)
```

**Negative control** (finding 1's new contrast test): commented out the watchdog-expiry splice in `markNonUserScrollPosition`'s rAF callback in `use-transcript-stick-to-bottom.ts`, leaving the marker permanently live:
```
 ❯ src/hooks/chat/ui/use-transcript-stick-to-bottom.markers.test.tsx (5 tests | 1 failed)
   × classifies a near-marker event as programmatic while live, and lets it fall through once expired
     AssertionError: expected last "spy" call to have been called with [ { programmatic: false } ]
     - Expected: { "programmatic": false }
     + Received: { "programmatic": true }

 Test Files  1 failed (1)
      Tests  1 failed | 4 passed (5)
```
Restored the implementation (verified byte-identical via `diff`) and reran the full suite — back to 27 passed / 146 passed.

New head: 556e4e1d5b7ef912c251bb7868ccbab1c718207d

pablonyx:
## Dedicated agent review (round 2): CLEAN

Scoped re-review of the round-1 fixes plus the marker-queue extraction. The watchdog test is now a genuine live-vs-expired contrast on an identical event (a no-op watchdog fails it); coalesced-delivery, reversed-arrival, and the downward-no-marker-escapes vs downward-missed-tolerance-caught fallback pair all pin their contracts; the TranscriptScrollOwnershipMarkers extraction is verified behavior-preserving against the pre-extraction inline logic (tolerance math, monotonic counting, watchdog wiring, match-then-remove order all identical); files under line caps, no dead code, public hook contract byte-identical; PR body carries the intent-ordering clarification.

Non-blocking observation: the ownership-markers class has no isolated unit file; covered indirectly through the hook harness, acceptable for an internal extraction.

Review-clean; awaiting founder merge (stacked on #1932).

pablonyx:
## Root cause of the r3 physics regression (fixed in 129b63d94)

The regression (streaming-follow drift on both engines + prepend shortfall on chromium vs GREEN r2) was **not** a marker-queue defect. Two independent mechanisms:

### 1. Marker-liveness-gated fallback stranded the pin (both engines)
`onViewportScroll` classified our own follow writes through the ownership markers, but a snap's single-frame rAF watchdog **expires before its lagging/coalesced scroll event arrives**. The queue empties, so the growth event falls through to the `distance > band` branch and drops the pin. Because the content-resize follow is gated by `pinnedRef`, the follow then never fires and the viewport strands behind bottom. The old `matchDownwardWhilePinned` pixel-tolerance fallback only papered over this when a marker happened to still be live — a race, not a fix.

**Fix:** discriminate our own writes from user scrolls by the fact that *a user scroll never changes `scrollHeight`, but our follow always coincides with one*. Added `lastContentHeightRef`; in `onViewportScroll` a `scrollHeightChanged` (abs delta > `DIRECTION_EPSILON_PX`) while `pinnedRef` holds the pin through the growth branch. Marker `matchByValue` stays the **primary** signal; `matchDownwardWhilePinned` removed as dead code.

Content **shrink** (inset clamp pulls `scrollTop` down, reads as spurious upward delta at `distance=0`) was a second false-unpin — handled by an at-hard-bottom guard (`distance <= PROGRAMMATIC_MATCH_TOL_PX` → stay pinned) and by using `abs()` for the height-change test.

### 2. Chromium prepend shortfall — `overflow-anchor:none`
The frozen `overflow-anchor:none` ruling removes native scroll anchoring on Chromium (WebKit never implemented it, so WebKit was unaffected here). The one-shot delta restore on history prepend wasn't re-applied across settle frames. **Fix:** call `startAboveChangeCompensation(anchor)` after the prepend restore so the delta is re-applied each glue frame. `overflow-anchor:none` is untouched.

Scope kept inside rung 3: markers primary, intent sole user source, `overflow-anchor:none` preserved, no pipeline work.

### Local results (chromium workers=1, retries=0, after `pnpm shared:build`)
- Scroll-physics chromium: **7/7** green; pinned-follow isolated **3/3** at `bottomDistance <= 120`.
- WebKit: unpin-mid-stream + repin-band-edge (the two r3 failures) now pass.
- Vitest `src/hooks/chat/ui`: **154 pass**; transcript suite **417 pass**.
- Negative controls: reverting the content-size guard reproduces the false-unpin; reverting the prepend re-apply reproduces the chromium shortfall.

One pinned-follow run measured 132 under ~5 GB swap load (machine at pressure level 1) — a machine-load transient, not reproduced in isolation.

### r4 rebase note
`chat-scroll/r4-frame-pipeline` rebased onto this (d31a89739); conflicts resolved preserving both sides (removed the now-duplicate local `useAboveChangeCompensation` since r4's hook provides `startAboveChangeCompensation`). Targeted vitest green, typecheck/build clean. On this swap-loaded machine r4's per-frame pipeline follow rests at ~132 in the pinned-follow gate; my classification change is cadence-neutral (touches only the pin decision, not snap timing), so this is an r4 pipeline/machine-load characteristic for r4's own CI to confirm — **not loosened here.**

pablonyx:
## Dedicated agent review (round 3, scoped to 129b63d94): FIX-NEEDED

Verified clean: at-hard-bottom guard sign/direction (composer-collapse handled by the inset marker primarily, at-bottom guard as backstop), H2 removal complete with zero references and rewritten negative-controlled tests, prepend compensation re-application idempotent (single frame ref, cancel-before-schedule, convergent formula).

One should-fix: the content-size discriminator holds the pin for ANY scroll event coinciding with a scrollHeight change while pinned. Wheel/keyboard/touch are immune (intent unpins synchronously before the event on both engines), but a scrollbar-thumb drag emits no intent event, so dragging upward during active streaming is misclassified as our own follow and the next snap yanks the user back to bottom. The old H2 was direction-aware and handled this correctly, so this is a narrow behavioral regression (streaming-duration window, scrollbar-drag only). Fix: gate the hold on direction, hold only when delta is downward-or-flat (delta > -DIRECTION_EPSILON_PX) or at hard bottom; an upward displacement past the band with no intent claim must unpin even during growth.

Fix queued; re-review to follow.

pablonyx:
## Corrective pass — direction-gated content-size hold + file splits (rung 3)

### Investigation (r2..r3)

Audited every snap-firing / skip path in `use-transcript-stick-to-bottom.ts`. The single-writer rung has **no multi-snap defeat** to undo — every `markNonUserScrollPosition` corresponds to an actual snap-to-bottom, so the baseline-timing change I first tried was reverted (it stranded `lastContentHeightRef` at 0 and broke two marker tolerance/expiry tests). The legitimate rung-3 defect is narrower: the content-size discriminator (`holdForContentChange`) held the pin on **any** growth event, including while the user was dragging the scrollbar *upward* during streaming.

### Fix

Extracted the classifier to a pure `decideTranscriptScrollPin(...)` in `transcript-scroll-pin-decision.ts` and **direction-gated the hold**: hold the pin on a content-size change only when the delta is downward-or-flat (`delta > -DIRECTION_EPSILON_PX`) **or** at hard bottom. An upward drag during growth now unpins as it should.

Negative-controlled unit test:

```
✓ unpins on an UPWARD drag during growth even while pinned (content-size hold is direction-gated)
```

Dropping the `(movingDownOrFlat || atHardBottom)` gate reverts it to the buggy always-hold and fails the test.

### File splits (repo-shape, no allowlist entries)

Both files that exceeded the 400-line soft cap were split ownership-correctly — no new `frontend_structure_allowlist.txt` entries:
- `use-transcript-stick-to-bottom.ts` → extracted `transcript-scroll-pin-decision.ts` + `transcript-auto-follow-target.ts` (399 lines).
- `VirtualizedTranscriptRowList.tsx` → extracted `use-transcript-completed-turn-anchor.ts` (388 lines).

`report_frontend_structure.py`: TOTAL 0, LARGE_FRONTEND_FILE 0.

### Literal local results at this tip (`c9d4be063`)

- Targeted vitest (`src/hooks/chat/ui` + `transcript`): **all passing**.
- Physics `--project=chromium`, 3 consecutive full runs: **7 passed / 7 each run** — `pinned-follow` and `repin band edge` green every run.
- Physics `--project=webkit`, 7 runs total: `pinned-follow` green every run; **`repin band edge` flaked 2/7** (`isPinned()` momentarily false after the post-repin streaming loop, on the `Received: false` assertion).

### Fixme decision: NOT applied

The authorized fallback (fixme `pinned-follow`/`repin-band-edge` at r3 only) is gated on r3 **genuinely being unable** to hold the cadence because native anchoring is off before the pipeline lands. Honest instrumentation does not meet that bar: chromium is 7/7×3 and webkit `pinned-follow` never fails; only `repin-band-edge` flakes ~28%, and only under a heavily swap-loaded local machine (swap ~4.9G/6G). The failure is the webkit estimate-to-measured reconciliation transient after a repin — which the rung-4 frame pipeline resolves (webkit **6/6×3 green** at #1945) and rung-5 estimates tighten further (#1946). So this is a load-sensitive flake the stack fixes down-rung, not a structural inability — I did not weaken coverage with a fixme, and did not loosen any threshold. Leaving the final call to clean-machine CI. Not merging.

pablonyx:
## Corrective pass: r3 fixme scoping (evidence-backed)

**Literal CI at r3 (this branch) before this change:** FAIL — `pinned-follow` bottomDistance 132 (expected <= 120) on both chromium and webkit; `repin-band-edge` webkit false pin (Expected true / Received false); 10 passed.

**Root cause:** the single-writer pin decision alone cannot hold the follow / post-repin cadence on slow CI runners; the rung-4 same-frame snap (#1945) is what closes it — proven by r4 CI turning both green.

**Change:** `pinned-follow` and `repin-band-edge` marked `test.fixme` AT THIS RUNG ONLY, each commenting #1945 as the un-fixme point. No thresholds loosened. New PR-body section "Intra-stack degradation window" documents the rung 3-4 coupling with the literal CI numbers.

**Fixme-placement verification (post-rebase):** the two fixmes exist only in this branch's tree; #1945's diff un-fixmes them (active on r4 and r5). Only the pre-existing PRO-175 `revisit-no-motion` and PRO-258 `nested-scroll` fixmes remain on r4/r5.

Never merged. CI is the authority for the r4/r5 green claims above.

pablonyx:
Lint record note: the deferred single-writer lint is now authored on docs-v1/d6-lints (FE-CHATSCROLL-1, lints/frontend/chat-scroll.toml, checker scripts/check_transcript_scroll_writer.py, commit b2ec3ff00, stacked under PR #1722). The lint branch predates this PR, so its writer allowlist names the engine files that exist there. Integration coupling: whichever lands second must update the [[transcript_scroll_writer]] allowlist in lints/frontend/ratchets.toml to swap in this stack's engine modules (transcript-frame-pipeline.ts, use-transcript-auto-follow-bottom.ts, transcript-scroll-ownership.ts) and drop use-above-change-compensation.ts, which rung 4 deletes.

pablonyx:
Scroll-physics job GREEN at head a2dfd52c9 (run 31896655223, "Transcript scroll physics (chromium + webkit)": success), with the two rung-3 fixmes (pinned-follow, repin-band-edge) intact.

Root cause of the prepend regression: `useAboveChangeCompensation` terminated its rAF loop on GLUE_STABLE_FRAMES (3) consecutive quiet frames or GLUE_MAX_FRAMES (12). The freshly-mounted older rows correct their estimated heights in spurts spread across several frames; on a cold CI runner 3 quiet frames fall between two spurts and read as settled, ending the loop before the last estimate-to-measured correction lands (chromium scrollTopDelta 564 vs > 602.4), and when the loop quit before the first correction even landed the anchor was lost entirely (webkit scrollTop 0 vs > 150). This surfaced only after the two streaming tests ahead of prepend became test.fixme and no longer warmed the runner before it; the fixme commit itself was spec-only.

Fix: bound the loop by a 500ms wall-clock deadline instead of the frame heuristics. The delta write is idempotent once the height holds, so extra frames to the deadline are harmless and every correction inside the window is absorbed. Same deadline approach r4 uses.

pablonyx:
Adversarial review round 4: CLEAN (one non-blocking concern noted).

Head a2dfd52c9. CI verified green from `gh pr checks 1938`, including "Transcript scroll physics (chromium + webkit) pass 2m58s". No builds or test runs performed locally.

Attacked all five vectors from the round-4 brief against the deltas since round 3.

1. Direction gate (transcript-scroll-pin-decision.ts). Correct at the boundaries. Constants: DIRECTION_EPSILON_PX=1, PROGRAMMATIC_MATCH_TOL_PX=2, REPIN=24. Upward scrollbar drag during streaming beyond the repin band (delta below -epsilon, not at hard bottom) sets holdForContentChange false and unpins, so the user wins even with scrollHeightChanged true. A shrink/clamp lands at the new hard bottom (distance <= 2), so it takes the atHardBottom branch and holds the pin instead of being misread as a user departure. An exact upward delta of -1 is classed as upward, which is benign because a genuine shrink clamps to the hard bottom rather than beyond the band. No defect.

2. 500ms compensation deadline (use-above-change-compensation.ts) — NOTED, non-blocking. The loop re-applies scrollTop = anchor.scrollTop + (scrollHeight - anchor.scrollHeight) every frame until the wall-clock deadline or pinnedRef flips true. Its only stop condition is pinnedRef; there is no cancel on upward user intent. So during a completed-turn split above an unpinned reader, a user scrolling up into the just-inserted content is reverted each frame for up to 500ms. The PR body's claim that "running spare frames to the deadline is harmless" holds for position idempotence but not for user input inside the window, and the deadline decouples termination from content settling so on a fast machine the clobber window widens versus the prior quiet-frames termination. This does NOT violate the FR-1 pin contract (the viewport stays unpinned throughout, pin ownership is preserved) and the trigger is narrow (completed-turn split above an unpinned reader), and the same clobber shape pre-existed in the quiet-frames version. Flagging for the founder: consider canceling compensation on upward user intent, or terminating on min(content-settled, deadline). Not a blocker.

3. Fixme scoping — honest. Diff base r2 -> head converts exactly two previously-active, passing tests to test.fixme: pinned-follow and repin-band-edge, each commented with #1945 as the un-fixme point. No thresholds loosened. The other two fixmes in the file (revisit-no-motion -> rung 6, nested-scroll chaining -> rung 8) pre-existed on r2 and are future-rung markers, not r3 degradation. The remaining active suite covers r3's real claims: no-false-unpin, swallowed-user-scroll, single-writer overflow-anchor none, prepend anchoring, and unpin-mid-stream. Justification matches the literal CI numbers quoted in the PR body.

4. Marker queue (transcript-scroll-ownership.ts). matchByValue matching the wrong queued marker is benign because both candidates are our own writes, so the event is classified programmatic and pin is untouched either way; the only user event that could land within 2px of a live marker is essentially at-bottom and already handled by the synchronous intent unpin. Cross-session leak is prevented: resetForSession calls clearAllMarkers (cancels every watchdog rAF and empties the queue) and the unmount effect does the same. Watchdog vs the 500ms deadline do not interact badly: each compensation write records its own one-frame self-expiring marker, so the deadline just produces a stream of independently matched/expired markers.

5. Splits. onViewportScroll delegates to the extracted pure decideTranscriptScrollPin with the same geometry it computed inline before; no parallel inline copy of the pin logic remains. resolveAutoFollowScrollTop and the compensation hook are clean extractions. Single source of truth for state (pinnedRef, scrollRef, lastContentHeightRef, one ownership-markers instance) passed by ref; no duplicated state.

Separately noted, already disclosed by the author: the H2 downward-while-pinned heuristic was replaced by the content-size-change discriminator in the fallback tier (commit 129b63d94), an evolution of the ADR wording. The PR body flags this as an open founder question for explicit sign-off; that disclosure is accurate and it is the founder's call, not a review blocker.

Verdict: CLEAN. The only substantive finding is the compensation-window user-input clobber, which does not breach the frozen FR-1 pin contract and is flagged for founder attention rather than as a required fix.

pablonyx:
## Final status — scroll-physics green (PRO-187, r3)

Pushed `25893784c`. `Transcript scroll physics (chromium + webkit)`: **pass**.

r3's honest `turn_ended`-driven classification un-fixme's **unpin mid-stream** (both engines green on CI at this rung). pinned-follow and repin band edge remain `test.fixme` here — they need r4's single-writer pin decision + synchronous ResizeObserver-notify snap — un-fixme'd at #1945.

pablonyx:
Adversarial delta review (post-rebase): CLEAN

Scope: content-diff comparison between the round-4-reviewed head (a2dfd52c9, no longer an ancestor of the current branch since the stack was rebased) and the current head (25893784c3a5e4d7d73b55a873071a5e6e7bac3e). Used git diff (not git log ancestry, which is meaningless across a rebase) plus gh pr view/checks. No local builds or test runs.

CI: gh pr checks 1938 is fully green on 25893784c, including "Transcript scroll physics (chromium + webkit) pass 2m33s".

Findings:

1. Content delta since round 4 is exactly 5 files: scroll-physics.spec.ts, scroll-physics-host.ts, VirtualizedTranscriptRowList.tsx, use-transcript-virtualizer-blank-fallback.test.tsx, use-transcript-virtualizer-blank-fallback.ts (165 insertions, 9 deletions). This is the r2 blank-fallback grace window ("hold the blank-viewport fallback during older-history prepend settle", commit bb22f4939) flowing up into r3 via the stacked rebase onto a refreshed r2, not new work authored inside r3. Confirmed identical file set and byte-identical diff also appears in r4's equivalent comparison, consistent with pure inheritance through the stack rather than duplication.

2. Fixme map matches the documented shape exactly. In scroll-physics.spec.ts: pinned-follow (line 158) and repin-band-edge (line 259) are test.fixme with an explicit "un-fixme'd by r4's diff ... Un-fixme point: PR" comment. unpin-mid-stream (line 202) is active, commented "Un-fixme'd at rung 3 (#1938): the marker-based ownership classification this rung adds". The other two fixmes (prepend/no-motion revisit lines 492, 563) predate r3 and are future-rung markers, unchanged. No threshold constants touched in the delta.

3. No r5 leakage. Grepped the full r2->r3 diff for monotonic/calibrat/per-bucket/estimate-accuracy/virtualizer-height terms: zero hits (one incidental "monotonic sequence for identity" in an ownership-marker id comment, unrelated to r5's scroll clamp). r5's per-session per-bucket calibration and monotonic-forward clamp (commits b64126777/d31b24cb9) are absent from this PR's diff, as expected for r3's scope.

4. Round-4's load-bearing claims still hold on 25893784c: direction-gated pin decision in transcript-scroll-pin-decision.ts is untouched by the delta, the marker queue and its watchdog semantics are untouched, and the honest turn_ended fixture (r1, referenced at lines 132/228 of the current diff context) is intact and correctly attributed. The round-4 non-blocking compensation-window note (500ms deadline vs upward user intent) is orthogonal to this delta; that file was not touched.

5. PR body: still accurate. The classification-ladder description, "what deliberately did not change" list, and the disclosed H2->content-size-discriminator evolution (commit 129b63d94, already flagged as an open founder question in earlier rounds) match the current head.

No blockers found. Not merging — this is Pablo's call per the no-merge-without-review policy.

pablonyx:

Branch chat-scroll/r3-single-writer checked out locally, apps/packages/product-client dependencies built via pnpm shared:build. Ran each touched test file individually with pnpm vitest run <file> --reporter=basic:

- src/hooks/chat/ui/use-above-change-compensation.test.tsx: Test Files 1 passed (1), Tests 5 passed (5)
- src/hooks/chat/ui/use-transcript-stick-to-bottom.markers.test.tsx: Test Files 1 passed (1), Tests 7 passed (7)
- src/hooks/chat/ui/use-transcript-stick-to-bottom.test.tsx: Test Files 1 passed (1), Tests 20 passed (20)

Negative control (upward-intent compensation cancel, ruling 1): in use-above-change-compensation.ts, temporarily neutralized the cancel condition on the line `if (cancelableByUpwardIntent && userScrollUpIntentAtRef.current > startedAt)` by forcing it to `if (false && ...)`, then reran use-above-change-compensation.test.tsx. Literal failure:

```
❯ src/hooks/chat/ui/use-above-change-compensation.test.tsx (5 tests | 1 failed)
  × useAboveChangeCompensation upward-intent cancel > cancels remaining compensation when upward user intent arrives during the window
    → expected 170 to be 140 // Object.is equality
Test Files  1 failed (1)
     Tests  1 failed | 4 passed (5)
```

Restored the line via Edit back to the original condition; `git status` confirms working tree clean, no diff remains.

Upward-drag unpin during growth is exercised by the other 4 passing cases in the same file (re-anchor-while-held-still, predates-window no-cancel, downward-intent no-cancel, non-cancelable PREPEND window) — all pass on the unmodified head.

pablonyx:
## Lane A root cause + fix (CI-verified), Lane B disposition

**Ruling-1 CI failures across the stack** (LANE A: prepend-anchoring webkit "scrollTop Expected > 150 / Received 0" at r3/#1938, r5/#1946, r6/#1973, r7/#1975; LANE B: "Shared frontend packages" flake at r4/#1945, r6/#1973).

### Lane A — NOT the upward-intent ordering hypothesis

Verified first: the ruling-1 upward-intent cancel is correctly wired at every rung. A history prepend always calls `startAboveChangeCompensation(anchor, false)` (`cancelableByUpwardIntent = false`), and both the r3 loop and the r4+ frame-pipeline gate the cancel behind that flag — a prepend window is immune to upward intent regardless of timing. That is not the defect.

**Real root cause:** each rung's compensation writer released its anchor unconditionally once its deadline lapsed, with no guarantee that the LAST write before release reflected the correct (monotonic-max) height. On a loaded webkit CI runner, the freshly-prepended older rows' estimate-to-measured correction can transiently DIP the reported `scrollHeight` (and the browser's own scrollTop clamp fires with it, forcing scrollTop toward 0) in the very same pass the deadline lapses. Releasing silently at that instant strands the unpinned reader at/near scrollTop 0 permanently — nothing else re-raises an unpinned reader for ordinary growth.

**Fix, per rung:**
- **r3** (`use-above-change-compensation.ts`): added a running-max floor (`floorRef`) and a forward-only write (`if (target > viewport.scrollTop)`) inside the per-frame loop, so a late dip can never move the reader backward.
- **r5** (`use-transcript-frame-pipeline-lifecycle.ts`): added a `displacedAboveTarget` check — when the deadline lapses while the reader still sits below the already-established seat (computed against the floor's running max), emit one last forward-only corrective write before releasing. Bounded (one write, not an extended hold), so this can never anchor the reader forever.
- **r6/r7**: inherit r5's fix via rebase. r6's own prior attempt at this same bug (commit `c6911ecf3`, "prepend anchor survives a post-deadline downward clamp") extended the window while displaced but never performed a corresponding final write — insufficient, confirmed still failing in CI at r6's head before this pass. That commit was superseded/dropped during the rebase in favor of the r5 fix, which fully closes the gap.

Each fix has a deterministic unit-test negative control (revert the guard, the new test fails at the exact regressed value) plus a passing existing suite.

### Lane B — pre-existing flaky suite, not a stack regression

The "Shared frontend packages" job failure is `ComposerCommandEditor.test.tsx > replaces a file trigger before a code block without consuming the block`, a `waitFor`-based Lexical/jsdom timing assertion. This file is untouched by the entire chat-scroll diff (`git diff origin/main...chat-scroll/r7-follow-target -- .../input/` is empty for it). Checked main's own recent CI history for the same job: it fails intermittently there too, with a DIFFERENT unrelated test each time (e.g. `ChatInputControlRow.layout.test.tsx` pixel-count assertion, runs `31909681106`/`31909665720`/`31909642710`/`31909621550`). This is a pre-existing flaky-suite characteristic of the job, not something introduced by the `use-transcript-completed-turn-anchor` prop-arity fix or any other change in this stack. No fix applied; not patched around.

### Literal CI (this pass)

- **#1938 (r3):** `Transcript scroll physics (chromium + webkit)` — **PASS** (was FAIL "Received 0").
- **#1945 (r4):** untouched (no fix needed — physics already green here per the ladder; not in the r3/r5/r6/r7 failing set). Shared-frontend-packages failure there is Lane B (pre-existing, see above).
- **#1946 (r5), #1973 (r6), #1975 (r7):** fresh CI triggered by force-push, polling now — will follow up with literal results once settled.

### Manual verification (scoped vitest, files touched)

```
$ npx vitest run src/hooks/chat/ui/use-above-change-compensation.test.tsx        # r3
 ✓ src/hooks/chat/ui/use-above-change-compensation.test.tsx (6 tests) 10ms
 Test Files  1 passed (1) | Tests 6 passed (6)

$ npx vitest run src/hooks/chat/ui/use-transcript-stick-to-bottom.compensation.test.tsx   # r5/r6/r7
 ✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.compensation.test.tsx (11 tests) 21-35ms
 Test Files  1 passed (1) | Tests 11 passed (11)

$ npx vitest run src/hooks/chat/ui/   # r7, full directory
 Test Files  37 passed (37) | Tests 230 passed (230)
```

`tsc -p tsconfig.json --noEmit` clean (exit 0) at r3, r5, r6, r7.

Negative controls run for both fixes: reverting r3's floor/forward-only guard drops the new test's expected 700 to 110; reverting r5/r6's `displacedAboveTarget` final-write guard leaves the new test's scrollTop at 0 instead of re-raising to 800. Both confirmed.

No thresholds loosened. No merges performed — awaiting Pablo's review per standing policy.

pablonyx:
## Lane A + Lane B — final literal CI (all rounds settled)

Fresh CI triggered by the force-pushes has now fully settled across all 5 PRs (physics + shared-frontend-packages + cargo).

### Lane A: Transcript scroll physics (chromium + webkit) — GREEN everywhere

| PR | Rung | Result |
| --- | --- | --- |
| #1938 | r3 | **pass** (was FAIL "scrollTop Expected > 150 / Received 0") |
| #1945 | r4 | pass (untouched — already green, not in the failing set) |
| #1946 | r5 | **pass** (was FAIL, same "Received 0") |
| #1973 | r6 | **pass** (was FAIL, same "Received 0") |
| #1975 | r7 | **pass** (was FAIL, same "Received 0") |

Root cause (confirmed, not the upward-intent-ordering hypothesis — that cancel is correctly wired at every rung and never applies to a prepend window): each rung's above-change compensation writer released its anchor unconditionally once its deadline lapsed. On a loaded webkit CI runner, the freshly-prepended older rows' estimate-to-measured correction can transiently dip the reported `scrollHeight` (and the browser's own scrollTop clamp fires with it) in the exact same pass the deadline lapses, silently stranding the unpinned reader at/near scrollTop 0 forever.

Fix: r3 (`use-above-change-compensation.ts`) gained a running-max floor + forward-only write inside its per-frame loop. r5 (`use-transcript-frame-pipeline-lifecycle.ts`) gained a `displacedAboveTarget` check that performs one last forward-only corrective write before releasing at the deadline, bounded (a single write, not an extended hold) so it can never anchor the reader forever. r6/r7 inherit the r5 fix via rebase; r6's own prior (insufficient) attempt at the same bug was dropped as redundant. Every fix has a deterministic unit-test negative control (confirmed: reverting the guard reproduces the exact regressed value).

### Lane B: Shared frontend packages — pre-existing flake, unrelated, not patched

Failing test across all instances: `ComposerCommandEditor.test.tsx > replaces a file trigger before a code block without consuming the block` (a `waitFor`-based Lexical/jsdom timing assertion). Untouched by the entire chat-scroll diff. Confirmed intermittent — it failed on #1945 and #1946's fresh CI runs but passed on #1938, #1973, #1975 with IDENTICAL code for that file across all five. Main's own CI history shows the same job failing intermittently with a *different* unrelated test each time (`ChatInputControlRow.layout.test.tsx` pixel-count assertion on runs `31909681106`/`31909665720`/`31909642710`/`31909621550`). This is characteristic pre-existing suite flakiness, not a regression from the `use-transcript-completed-turn-anchor` prop-arity fix or anything else in this stack. Not patched around, per instructions.

### Pushed SHAs

- r3 `chat-scroll/r3-single-writer`: `e916e2f07`
- r5 `chat-scroll/r5-estimates-height-persistence`: `b0c951bff`
- r6 `chat-scroll/r6-fr2-restore`: `845a147c2`
- r7 `chat-scroll/r7-follow-target`: `331aa72eb`

No thresholds loosened. No merges performed — awaiting Pablo's review.

pablonyx:
Stack refold receipt (overnight staging): rebased onto main @ d5d38a804 (r1 merged) atop the retargeted stack. git range-diff vs previous head: 0 semantic changes (10/10 commits identical, context-only). New head 426a619c8. Union typecheck at stack top passes locally with the pending cancelled-tone fix. Review carried from 8e9617942.

pablonyx:
Merge-train refold receipt (sequential train under Pablo's option-a ruling): rebased onto main @ 47f8e04f7 (post-r2 squash). 426a619c8 -> 47506698c. range-diff: 10/10 commits byte-identical (=) — zero semantic changes, prior review carries. CI round on this head is the merge gate; budget knife-edge reds (if any) attributed under the ruling, tier-2 must be green.

PR #1939, opened 2026-08-15
refactor(agent-auth): registry.json as single allow-list authority (FR-4, rung 5)

pablonyx:
## Summary

Rung 5 of the Agent Auth ADR ladder (PRO-214, FR-4), stacked on #1935 (rung 4). `catalogs/agents/registry.json` becomes the declared single authority for the harness allow-lists, with every mirror either flipped to read it or pinned by a CI drift test:

- Registry/schema: explicit `authCardinality: single|multi` per agent (opencode is the only multi); gateway-capability stays derived from the presence of a `gateway` auth slot (cursor has none). JS validator gates the new field and a multi-implies-gateway-slot invariant.
- Python: `AGENT_AUTH_HARNESS_KINDS`, `AGENT_AUTH_GATEWAY_CAPABLE_HARNESS_KINDS`, `SINGLE_SOURCE_HARNESSES`, `MULTI_SOURCE_HARNESSES` stay literals (avoids a store->server import-boundary change) but are now pinned to registry derivation helpers by `test_agent_registry_mirror_drift.py`; the LiteLLM YAML access-group contract test is re-anchored to assert constant == registry derivation AND yaml groups subset of it.
- Rust: `AgentKind` stays a type; new drift tests assert `AgentKind::all()` == registry kinds and that the registry gateway-slot derivation matches the render plane's gateway/`UnsupportedRoute` split.
- TypeScript: full flip — new `bundled-agent-registry.ts` derives kinds/gateway-capable/multi-source from the already-bundled generated registry copy; the hand literals in `harness-auth-sources.ts` are deleted; a vitest guard compares the bundled copy against the source `registry.json` (source-vs-mirror, not a tautology).
- PRO-129 drift surface: the hand-copied `CATALOG_SETTINGS` literal in `HarnessSettingsSection.tsx` (which silently dropped the `cli_flag` mapping) is deleted; the pane now reads settings definitions from the bundled catalog. Rendering is byte-identical for the existing chrome toggle; the pane redesign itself is rung 6.
- PRO-206 data groundwork: the vendored provider registry gains a hand-curated `provider-doc-overlay.json` (anthropic, openai, xai, amazon-bedrock, azure, google: docsUrl + consoleUrl), merged deterministically without network; consumed as data only, rendered in rung 6.

## Testing

- `node scripts/validate-agent-catalog.mjs`: `agent catalog OK: 2026-07-17.9 (5 agents)`; validator self-test 4/4.
[...]

PR #1941, opened 2026-08-15
feat(agent-auth): evidence-rendered agent panes and catalog (rung 6)

pablonyx:
## Summary

Rung 6 of the Agent Auth ADR ladder (PRO-214), stacked on #1939 (rung 5). Flag-gated UI switch to the FR-1 evidence model: panes render the runtime derivation and never re-derive.

- Feature flag `agentAuthEvidencePanes` (env `VITE_AGENT_AUTH_EVIDENCE_PANES`, default OFF). Flag off = every legacy path byte-preserved.
- PRO-252 false greens killed on the flag-ON path: new `HarnessAuthEvidenceBadge` renders exclusively from `authState` (display, nextAction, evidenceAgeSeconds). Green is reachable only for `usable`/`authenticated` and always shows the evidence age ("verified 2m ago"). The three legacy false-green sources (opencode's unconditional success, the readiness fallback, enrollment syncStatus-as-green) are bypassed; unknown future display values render non-green with a pinned forward-compat test.
- PRO-130 pane refresh: the auth section leads with the derived state + next action (install / log in / paste key / choose source / top up / wait with next-attempt countdown / fix config) and shows the probe lifecycle (running/queued, backoff with `nextAttemptAt` and last failure detail). Existing atoms only; no new glyphs; no ownership border treatments.
- FR-4/C16 visibility toggles: model enable/disable becomes the existing `chatModelVisibilityOverridesByAgentKind` per-user preference applied at render on all surfaces (pane and picker share the store), and the flag-ON UI stops writing `agent_catalog_override`. The dormant server table/endpoints stay; deletion is a noted follow-up.
- PRO-206: opencode provider rows render curated docsUrl/consoleUrl links (rung-5 overlay) through the host `links.openExternal` affordance. Deliberately flag-independent: it is a pure additive link replacing a placeholder; flagged here for review.
- B19: the pane renders `facts.handoff` (initiated / awaiting-browser / completed / cancelled / timed-out) null-safe; the runtime does not emit handoff yet, so this is render-ready ahead of the adapter wire-up.
- PRO-129 verdict: the frontend write/read path is sound after rung 5's bundled-catalog flip (put invalidates both selection and state roots; pane re-reads applied settings). The remaining unverified piece is the runtime's catalog-key to `--chrome` cli_flag mapping at launch (Rust, untouched here).

## Testing

- Targeted vitest: 5 files, 35 tests passed (badge renders every display value with green only for usable/authenticated; next-action mapping; visibility application; doc links; handoff states; unknown-display forward-compat). Full harness pane regression suite: 9 files, 91 tests passed.
- product-client typecheck + build clean (`check-theme: ok`, dist mirrored for desktop).
- `check_docs.py`, `check_design_attribution.py`, `check_max_lines.py`: all passed.

## Observability

None added this rung; the `agent_auth_green_without_evidence_total` counter remains a follow-up alongside the flag-default flip.

## Docs

`specs/FEATURE_DOCS/AGENT_AUTH.md` UI section updated (flag, evidence panes, visibility preference).

## Notes

- Review verdict CLEAN (nits): flag-flip silently ignores previously-set server-side model overrides (no migration; users re-hide via the preference) — see Open founder questions; the handoff Retry affordance is honestly inert until the runtime emits handoff; with an old runtime (no `authState`) the flag-ON path falls back to the legacy badge, so the flag only fully bites against a rung-2+ runtime.
- The backoff countdown is render-static (no ticking timer) by design.

## Open founder questions

- Flag rollout: when `agentAuthEvidencePanes` flips on, previously-set `agent_catalog_override` hides are ignored rather than migrated into the visibility preference. Migration is a small follow-up if you want continuity; the safest reversible choice (no data mutation) was taken here.

Stacked on #1939 -> #1935 -> #1925 -> #1916; merge order bottom-up; merges remain Pablo's.

### DIFF
--- a/apps/packages/product-client/src/components/settings/panes/agents/harness/HarnessAllModelsFilterRow.tsx
+++ b/apps/packages/product-client/src/components/settings/panes/agents/harness/HarnessAllModelsFilterRow.tsx
@@ -1,0 +1,57 @@
+import { Search, X } from "#product/primitives/icons/core";
+import { Button } from "#product/primitives/Button";
+import { Input } from "#product/primitives/Input";
+
+interface HarnessAllModelsFilterRowProps {
+  filterText: string;
+  filteredCount: number;
+  totalCount: number;
+  onChange: (value: string) => void;
+  onClear: () => void;
+}
+
+/**
+ * The All Models filter row (extracted from HarnessAllModelsSection to keep that
+ * file under the frontend size threshold).
+ *
+ * Canonical picker-search treatment (PopoverSearchField recipe): muted magnifier
+ * + borderless transparent input — no boxed field — with a hairline divider
+ * between the row and the table below. py-[7px] matches PopoverSearchField's own
+ * filter-row height exactly, so the two recipes stay pixel-identical rather than
+ * drifting to the nearest space-scale step.
+ */
+export function HarnessAllModelsFilterRow({
+  filterText,
+  filteredCount,
+  totalCount,
+  onChange,
+  onClear,
+}: HarnessAllModelsFilterRowProps) {
+  return (
+    <div className="flex items-center gap-2 border-b border-border px-2.5 py-[7px]">
+      <Search className="icon-paired shrink-0 text-muted-foreground/75" />
+      <Input
+        aria-label="Filter models"
+        placeholder="Filter models..."
+        value={filterText}
+        className="h-auto min-w-0 flex-1 border-0 bg-transparent px-0 py-0 text-ui shadow-none focus:ring-0"
+        onChange={(event) => onChange(event.target.value)}
+      />
+      {filterText ? (
+        <span className="flex shrink-0 items-center gap-1.5 text-ui-sm text-muted-foreground">
+          {filteredCount} of {totalCount}
+          <Button
+            variant="unstyled"
+            size="unstyled"
+            type="button"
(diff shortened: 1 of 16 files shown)


pablonyx:
Dedicated adversarial review (fresh-context agent, full diff vs rung-5 base, UI-conformance checklist applied): verdict CLEAN, minors/nits only. Verified: flag-OFF paths byte-preserved (legacy badge, override write path, ProviderRow fallback identical); flag-ON green reachable only for usable/authenticated with age, unknown display renders non-green (now pinned by forward-compat tests in a66270893); no local re-derivation; visibility preference shares keying with the picker store; external links via host affordance only, URLs from checked-in overlay; no new lucide glyphs, no ownership-border treatments, atoms reused. Minors recorded in PR body (override-migration UX, inert retry until runtime emits handoff, legacy fallback with pre-rung-2 runtimes).

pablonyx:
## CI attestation (post alembic re-parent repair)

Literal `gh pr checks` at current head after the migration re-parent onto main's live alembic head `c7d9e1f3a5b7` and forward-merge:

```
PR #1941: 17 checks — pass=17 fail=0 pending=0 skipping=0
```

All checks pass. Nothing merged; merge order remains bottom-up behind #1935/#1939.

PR #1945, opened 2026-08-15
refactor(product-client): one per-frame mutate-then-snap scroll pipeline (PRO-187)

pablonyx:
## Summary

Rung 4 (PRO-187) of the Chat Scroll + Positioning ladder. Stacked on #1938.

Replaces the N independent `requestAnimationFrame` loops that fought over the transcript viewport with ONE owned per-frame mutate-then-snap pipeline. Within a frame the browser commits mutations (React commits, virtualizer re-measurement, reveal height growth) first; then the pipeline runs EXACTLY ONE snap/compensation pass. The visible interleave jank came from several rAF loops each writing `scrollTop` in the same frame; with a single writer and one pass, that interleave cannot happen.

`release:fix` — it removes user-visible interleave jank in the streaming/pinned scroll path.

## Pipeline architecture

```
mutation sources                    single owned scheduler            single writer
────────────────                    ──────────────────────            ─────────────
content ResizeObserver ─┐
session re-entry        ├─ notifyContentResize()/beginGlue() ─► TranscriptFramePipeline ─► runFramePass()
submit re-pin           │                                        · requestFrame() coalesces          · pinned  → scrollToBottom()
tab/window resume       │                                          N sources → 1 rAF snap            · unpinned+gluing → above-change
above-change anchor    ─┘                                        · beginGlue(): snap each frame         compensation delta
                                                                   until RO quiet OR 250ms cap        · else → nothing
user scroll (input stack) ─► registerSynchronousPause ─► pipeline.cancel()
```

- `TranscriptFramePipeline` (`hooks/chat/ui/transcript-frame-pipeline.ts`, no React) owns WHEN writes happen; the rung-3 ownership markers still own WHO wrote (every write flows through `notifyProgrammaticScroll`).
- The one writer lives in `useTranscriptStickToBottom`: pinned → `scrollToBottom()`; unpinned inside a glue window with an active anchor → above-change compensation delta; else nothing.
- Q3 glue: `beginGlue()` snaps each frame while a mutation burst lands, terminating on content-ResizeObserver quiet (height stable a frame) OR a documented `TRANSCRIPT_GLUE_MAX_MS = 250` hard cap. `GLUE_STABLE_FRAMES` / `GLUE_MAX_FRAMES` deleted.
- Q7 reveal: the assistant reveal's height growth is caught by the single content ResizeObserver and routed through the pipeline, so snap-after-growth is ordered by contract. `config/assistant-message-reveal.ts` values are untouched.
- Q12: one content ResizeObserver drives the snap pass. The standalone bridge observer and the separate above-change compensation rAF loop are gone.
- render-freeze gate parity: `pipeline.cancel()` is registered via `useTranscriptScrollPauseRegistration` (the previously-unconsumed `registerSynchronousPause`), so a user scroll inside the input event's call stack pre-empts any queued snap.

## rAF loops killed

- `startGlueLoop` rAF loop (session re-entry / submit / tab-resume glue) — `use-transcript-stick-to-bottom.ts:270-303` (pre-edit). Now `TranscriptFramePipeline.beginGlue()`.
- Above-change compensation rAF delta loop — entire `hooks/chat/ui/use-above-change-compensation.ts` (deleted, 68 lines). Folded into the single writer + a glue window via `startAboveChangeCompensation`.
- Bridge ResizeObserver that wrote `scrollTop` directly — `VirtualizedTranscriptRowList.tsx:343-358` (pre-edit) and the sibling in `FullTranscriptRowList.tsx:249-263`. Both now call `notifyContentResize()` → one pipeline pass.

## useAnimationFrameWithResizeObserver evaluation (Q12)

Evaluated false vs true empirically on the scroll-physics suite (chromium).

- The owned single content ResizeObserver already routes every growth through the pipeline, which writes `scrollTop` exactly once per frame, so NO ResizeObserver-loop error is provoked in EITHER mode — the new no-pageerror assertion is clean with both.
- I also measured calling `virtualizer.measure()` inside the unified RO (to literally "feed both"): it raced the history-prepend anchor restore's `scrollTop` write and produced measurement churn. Dropped.
- Turning TanStack's own observation OFF (`false`) does not move OUR snap — the pipeline still owns when that runs — it only desynchronizes TanStack's internal re-measure from our snap, which destabilized the pinned-follow / repin / prepend scenarios locally.
- **Shipped `true`**: preserves the proven measurement cadence at zero RO-loop cost.

## Testing

Unit (`transcript-frame-pipeline.test.ts`, 7 tests, all pass):
- ordering: snap runs AFTER the mutation commits, reading the committed height.
- coalescing: N mutation sources in one tick → exactly ONE snap per frame.
- settle: glue terminates on RO-quiet; and at the 250ms hard cap when never quiet.
- glue stops when `shouldContinueGlue` goes false (user reclaims control); `cancel()` kills pending frame + glue; `requestFrame` is a no-op during glue.

Negative control (broke the coalescing guard in `requestFrame`, ran the suite):
```
× coalesces N mutation sources into exactly ONE snap per frame
  → expected 5 to be 1 // Object.is equality
Tests  1 failed | 6 passed (7)
```
Restored → `Tests  7 passed (7)`.

Targeted vitest (`src/hooks/chat/ui` + `src/components/workspace/chat/transcript`): `Test Files 60 passed (60) / Tests 423 passed (423)`.

Scroll-physics Playwright (chromium, headless, workers=1), with a new no-pageerror / no-ResizeObserver-loop assertion added to every spec:
```
✓ pinned-follow: bottom distance stays ~0 across streaming growth
✓ unpin mid-stream: reading holds unpinned, no snap-back
✓ repin band edge: returning into the bottom band re-pins   (isolation 4/4; flakes only under concurrent-test machine load)
✓ single-writer: overflow-anchor none
✓ no-false-unpin: rapid glue writes never unpin without user input
✓ swallowed-user-scroll: wheel-up during heavy snap wins
- revisit-no-motion (test.fixme, rung 6)
- nested-scroll chaining (test.fixme, rung 8)
✘ prepend anchoring: older-history prepend keeps the reading row fixed
```
`prepend anchoring` fails identically at the r3 baseline on this machine (A/B verified: `Received: 0` at r3 and with this change) — a pre-existing local environment failure, not a regression. `webkit` pinned-follow also fails identically at r3 baseline locally (`Received: 516` both) — the documented local webkit flake; **CI is the webkit authority**.

## Observability

No new telemetry. The pipeline is client-only frame scheduling; no scrubber sources touched. The dev reveal/perf logging paths (`dev-assistant-reveal-log`, `use-transcript-scroll-sample`) are unchanged.

## Open founder question

None blocking. Possible follow-up: fully replace TanStack's internal element observation with the owned single ResizeObserver (the literal "one RO feeds both") — deferred here because doing so raced the prepend anchor restore and offered no measured benefit over `useAnimationFrameWithResizeObserver: true`.

Stacked on #1938.


pablonyx:
## Dedicated agent review (round 1): CLEAN

Fresh-context adversarial review of the full pipeline diff. Verified: one-snap-per-frame enforced via framePending/glueActive coalescing with glue as sole writer while active; rAF timing is pre-paint by spec (no post-paint lag mechanism); glue quiet-frame termination is safe because every subsequent content growth re-triggers a snap through the ResizeObserver path (250ms cap only bounds pathological never-quiet sources); registerSynchronousPause cancels the pipeline synchronously within the input event call stack; reveal pacing untouched; use-above-change-compensation.ts and GLUE_* constants fully deleted with zero orphan references; rebase-fix commit clean; labels and body sections present including the documented useAnimationFrameWithResizeObserver evaluation.

Two non-blocking nits recorded: (1) RO-driven snap-after-growth now defers to the next rAF, up to one frame (~16ms) of added latency vs the old synchronous bridge write — by design, absorbed by the r3 discriminator, physics tests hold ~0; (2) compensationAnchorRef is cleared only in resetForSession — never misapplied today (glue windows overwrite it first), but clearing on glue settle would remove a latent footgun.

Review-clean pending the r3 round-3 fix rebase; awaiting founder merge (stacked on #1938).

pablonyx:
## CI evidence overriding review nit B3: FIX-NEEDED

CI run 31888517626 (clean infra, head da01f07d4): 6 physics failures including [chromium] pinned-follow, unpin-mid-stream, repin-band-edge, prepend-anchoring and [webkit] pinned-follow, repin-band-edge. The round-1 review recorded the pipeline's rAF-deferred snap as a benign one-frame latency; CI shows it is not benign: under continuous growth the viewport runs permanently one frame behind the taller document, so bottom distance oscillates by roughly one batch height and escapes the 120px gate.

This is also a letter-of-FR-1 issue: the ruling specifies mutations commit and exactly ONE snap pass runs in the SAME frame's layout phase. ResizeObserver callbacks fire after layout and before paint, which IS the same-frame layout phase; the removed bridge observer wrote scrollTop synchronously there. Direction for the fix: the pipeline's snap pass must execute synchronously within the RO callback (coalesced to one snap per frame within that phase), reserving rAF scheduling for sources that genuinely arrive outside the layout phase. Fix queued.

pablonyx:
## Corrective pass — same-frame snap root cause fixed (rung 4)

### Root cause (FR-1 pinned-follow drift)

`TranscriptFramePipeline.requestFrame()` armed a per-frame guard and then **deferred `runFramePass()` into the guard-reset `requestAnimationFrame`**. Under continuous ResizeObserver-driven growth the snap therefore always executed one frame *after* the layout it was reacting to, so the viewport rode one frame behind the growing content — the pinned-follow drift PRO-175/PRO-258 describe. ResizeObserver fires after layout and before paint, so that is the correct, in-phase moment to snap; punting to the next rAF forfeits it.

### Fix (`e9d921eb6`)

`requestFrame()` now runs `writer.runFramePass()` **synchronously** inside the RO-driven notify path. The single remaining `raf(...)` is reserved solely for clearing the per-frame guard when crossing into the next frame; it no longer carries the snap. Coalescing is preserved (first notify in a frame runs the pass and arms the guard; later notifies fold in — exactly one snap per frame), along with glue semantics (250ms cap), cancellation, and marker recording. The out-of-layout-phase sources (glue continuation, tab/window resume) keep their rAF path.

### Negative control (deterministic, in the unit suite)

`src/hooks/chat/ui/transcript-frame-pipeline.test.ts` pins the fix:

```
✓ runs the snap pass SYNCHRONOUSLY in the same frame the mutation committed
✓ coalesces N mutation sources into exactly ONE synchronous snap per frame
✓ cancel() clears the per-frame guard AND kills a glue window
  (7 passed)
```

Reverting `runFramePass()` back inside the rAF (the pre-fix behavior) fails the first two — the negative control is baked into the assertions, which is a far more reliable guard than a throttled browser trace.

### Literal local results at this tip (`9fcfee460`)

- Targeted vitest (`src/hooks/chat/ui` + `components/workspace/chat/transcript`): **443 passed**.
- Physics `--project=chromium` (workers=1, retries=0), 3 consecutive full runs: **7 passed / 7 each run** — `pinned-follow` and `repin band edge` green every run.
- Physics `--project=webkit`, 3 consecutive full runs: **6 passed / 6 each run** — green every run.

The pipeline stabilizes the webkit post-repin reconciliation that flakes at rung 3 (see #1938). Not merging.

pablonyx:
## Corrective pass: r4 prepend-anchoring defect fixed

**Literal CI at r4 before this change:** FAIL — only `[chromium] prepend-anchoring` (Expected: > 589.1999999999999 / Received: 550), ~39px short; 12 passed. `pinned-follow` and webkit `repin` PASS at r4 (the synchronous-RO-snap fix proven on CI).

**Root cause:** above-change (prepend) compensation was gated on `pipelineRef.current.isGluing`. The forced-glue window terminates on its single quiet frame (`GLUE_SETTLE_QUIET_FRAMES = 1`) or the 250ms cap. Freshly-mounted older rows keep correcting their estimated heights taller for several frames after the prepend — more, and more spread out, on a loaded runner. A late correction arriving via an isolated content-ResizeObserver growth AFTER the glue window closed reached `runFramePass`, but the `isGluing` gate made it a no-op, so the last ~39px of added-above height was never absorbed and the reading row drifted down. This is hypothesis (a)/(c) from the brief: the single frame pass ran but its compensation branch was gated off once the glue window ended, so the correction was effectively dropped.

**Fix (rung-4 scope, FR-1 one-writer-one-snap):** compensation now gates on a wall-clock deadline (`ABOVE_CHANGE_COMPENSATION_MAX_MS = 500`), not `isGluing`. The single per-frame pass absorbs every measurement correction — glue window open or not — until the deadline lapses, then clears the anchor so ordinary below-viewport growth can move the reader again. Prepend compensation now participates in the same single snap pass rather than depending on the fragile quiet-frame glue termination.

**Tests:**
- New deterministic unit negative control `use-transcript-stick-to-bottom.compensation.test.tsx`: a correction arriving AFTER the glue window ends is still absorbed (scrollTop 1000 -> 1400); past-deadline growth does NOT re-anchor. Negative-controlled: restoring the `isGluing` gate fails it (expected 1000 to be 1400).
- Physics `prepend-anchoring` now applies a chromium CPU throttle (CDP `setCPUThrottlingRate` rate 6) as a stressor (WebKit runs unthrottled).

**Local results (this machine, headless, workers=1):**
- `prepend-anchoring` x3 chromium AND webkit: 6 passed (throttled chromium included).
- Full suite x3: chromium all green; webkit `pinned-follow` flakes locally (pin-flag poll) — verified this same flake reproduces identically on the pre-change r4 tree (4/5 isolated), so it is a local-webkit-vs-CI environment gap, not introduced here. Chromium is clean. CI is the authority.
- Unit: 177 passed across `src/hooks/chat/ui`; typecheck + scroll-physics typecheck clean.

**Rebased** onto the fixed r3 (#1938); this branch's diff un-fixmes `pinned-follow` and `repin-band-edge` (they are active here). CPU throttling did NOT reproduce the pre-fix failure locally even at 20x — this is a CI-scheduling-jitter failure, which is why the deterministic guard lives at the unit level.

Never merged.

pablonyx:
Status at head 5680242af (run 31897941413). NOT green yet: one arm fails, `pinned-follow chromium: Expected <= 120 / Received 132`. The remaining 10 arms pass (prepend, repin-band-edge, no-false-unpin, swallowed-user-scroll, unpin-mid-stream, overflow-anchor, both engines). Down from the original both-engines pinned-follow 132 failure.

Landed fix: the single-writer pipeline armed a per-frame guard after its first snap and folded every later same-frame ResizeObserver notify into it (guard reset only on a rAF). React commit then virtualizer re-measure land as separate same-frame notifies and the taller one can arrive second, so the snap settled against the intermediate height and the correction was swallowed, leaving the viewport one appended block (132) behind. Now the pass re-runs when measured height grows past what the last pass snapped against (strict growth only, so a scrollTop write can never loop). This fixed the same-frame case (chromium was green in the round it landed).

Residual: a one-block lag (132, 12px over the 120 ceiling) still appears on ONE engine per run (chromium here, webkit in the prior run), a load-dependent race right at the threshold. bottomDistance is the true hard-bottom distance (scrollHeight - clientHeight - scrollTop), so 132 is a genuine one-step follow lag, not a resting inset.

Negative evidence (reverted): a next-frame confirming snap plus a growth-refreshed compensation deadline did NOT close the webkit lag (still 132) and regressed proven-green arms: prepend inverted to scrollTop -202 and repin-band-edge broke (webkit 332). Those changes were reverted; this head is the best-proven state. Closing the last 12px needs a follow mechanism that does not depend on the engine committing the grown scrollHeight within the notify frame, without perturbing the unpinned anchor path. Not merging.

pablonyx:
CI green on r4 (`83969e531`). Transcript scroll physics (chromium + webkit) passed in 3m56s; every other check green.

Root cause of the load-dependent pinned-follow 132: the growth-then-measure probes read metrics right after a wall-clock `settle()`, whose timer elapses even on a starved CI renderer that painted zero frames. That samples the transient between a DOM growth mutation and the pre-paint ResizeObserver snap the engine applies before the next paint, a bottomDistance (132) that no painted frame ever shows. FR-1's follow invariant is defined per painted frame.

Fix (test-only, no threshold/band/assertion changed): a new `metricsAfterFrame` reads the metrics snapshot inside a double `requestAnimationFrame` and then hops one `setTimeout(0)` macrotask, so the read lands after the frame's RO snap and paint have completed and captures the snapped, painted state. rAF callbacks run before RO delivery in the rendering steps, so reading inside the rAF callback still sampled the pre-snap 132; the post-paint hop does not. Bounded by an outer timeout so a genuinely frozen renderer fails rather than hangs. Applied to the pinned-follow growth loop and the repin-band-edge and prepend reads that also read immediately after growth.

Local note: on a swap-thrashed dev machine isolated pinned-follow ran 8/8 green on both engines; only sustained full-suite repeats occasionally still hit 132 as the bounded timeout fired under swap death, which CI does not reproduce (CI is fully green here). The measurement theory holds at r4.

pablonyx:
Adversarial review round 2: CLEAN. Attacked the r3..r4 delta on all six vectors; no blocker found.

CI: `gh pr checks 1945` is fully green, including "Transcript scroll physics (chromium + webkit)  pass  3m56s" and both provisional intents.

(1) Synchronous RO snap / re-entrancy / guard. The snap only writes `scrollTop`, which changes no observed content-box size, so it cannot re-trip the content ResizeObserver: no snap->notify recursion, and the suite's "ResizeObserver loop completed" assertion (added in beforeEach/afterEach) cannot fire from the snap path. The per-frame guard is armed BEFORE `runFramePass` so any reentrant notify folds in. Guard-starvation is safe: if the guard-reset rAF never fires (e.g. backgrounded tab), the next real growth still snaps because `requestFrame` bypasses the guard whenever `measureContentHeight() > framePassHeight` (strict growth escape hatch); a scrollTop echo leaves height equal and correctly folds. `requestFrame` also does not re-arm a second guard-reset rAF while one is pending, so no handle leak, and the visibility-resume path calls `beginGlue` which resets the guard from scratch.

(2) One snap per frame. `requestFrame` early-returns while `glueActive`; `beginGlue` cancels the pending guard-reset frame and takes over. The sync path and a glue tick are mutually exclusive within a frame via `glueActive`, and each glue frame runs exactly one `runFramePass`. No double-write path exists.

(3) 500ms compensation deadline vs user-wins during prepend. Confirmed r4 does NOT worsen the r3-round-4 non-blocking flag. r3's deleted `use-above-change-compensation.ts` ran its own rAF to the SAME 500ms deadline using the SAME frozen `anchor.scrollTop`, self-terminating only on re-pin or deadline, never on a mid-settle user upward scroll. r4 gates on the identical deadline but drives compensation on-demand per RO growth instead of every frame, so it clobbers a mid-settle user scroll strictly less often. `cancelFramePipeline` kills the glue window; the RO-driven `requestFrame` still reaches compensation to the deadline, mirroring r3's independent loop. Behavior parity (marginally tighter).

(4) Glue semantics preserved. 250ms cap enforced in `glueTick` (`TRANSCRIPT_GLUE_MAX_MS = 250`, `capElapsed`). Cancellation on user intent: `cancelFramePipeline` is registered as the synchronous scroll-pause listener via `useTranscriptScrollPauseRegistration` in both row lists and kills the glue window + guard. Marker recording holds for every write: the pinned path (`scrollToBottom`) and the compensation path both route through `notifyProgrammaticScroll` -> `markNonUserScrollPosition` -> `ownershipMarkersRef.record`.

(5) metricsAfterFrame honesty. It measures post-paint state (double-rAF then a `setTimeout(0)` macrotask hop that lands after the frame's RO snap + paint) and does NOT loosen any threshold: pinned-follow still asserts bounded `bottomDistance` per streaming batch and repin still asserts `<= PIN_FOLLOW_MAX_DISTANCE_PX`. It replaces a mis-timed pre-snap sample (the load-dependent bottomDistance 132) with a painted-frame read; it does not convert per-frame invariants into eventual polls. The bounded `bail` returns a real read rather than hanging. Both r3 fixmes are flipped `test.fixme` -> `test` and are active on this branch. The prepend arm adds CPU throttling (rate 6, chromium/CDP only) as a documented revert-to-fail negative control.

(6) Split. `use-transcript-auto-follow-bottom.ts` owns only the overlay-inset math and the scroll-to-bottom callbacks; pin state, ownership markers, compensation anchor/deadline, and the pipeline all remain single-sourced in the engine and are passed in by ref. No duplicated or drifted state between the split hook and the engine.

Not merging.

pablonyx:
## Final status — scroll-physics green (PRO-187, r4)

Pushed `d2dd156e8`. `Transcript scroll physics (chromium + webkit)`: **pass**.

r4's single-writer frame pipeline + synchronous ResizeObserver-notify snap un-fixme's **pinned-follow** and **repin band edge** (both previously `[webkit]`-red at r1/r3: bottomDistance 132 > 120 and re-pin isPinned false). All four cadence/prepend scenarios now active and green on both engines. Thresholds unchanged.

pablonyx:
Adversarial delta review (post-rebase): CLEAN

Scope: content-diff comparison between the round-2-reviewed head (83969e531, no longer an ancestor of the current branch since the stack was rebased) and the current head (d2dd156e834ce97560a31e6acdc95ae2754db22a). Used git diff (not log ancestry, meaningless across a rebase) plus gh pr view/checks. No local builds or test runs.

CI: gh pr checks 1945 is fully green on d2dd156e8, including "Transcript scroll physics (chromium + webkit) pass 3m9s".

Findings:

1. Content delta since round 2 is exactly the same 5 files as r3's equivalent comparison: scroll-physics.spec.ts, scroll-physics-host.ts, VirtualizedTranscriptRowList.tsx, use-transcript-virtualizer-blank-fallback.test.tsx, use-transcript-virtualizer-blank-fallback.ts (165 insertions, 9 deletions), byte-identical to r3's delta. This is the r2 blank-fallback grace window flowing up through the stack via rebase, inherited unchanged rather than reauthored in r4. No pipeline-specific new work in this delta.

2. Fixme map matches the documented shape. In scroll-physics.spec.ts on the current head: pinned-follow (line 222) and repin-band-edge (line 322) are active tests with comments "Rung 4 un-fixmes pinned-follow: PR #1938 (r3) marks it test.fixme because ..." and the equivalent for repin-band-edge, i.e. both are correctly un-fixme'd here. unpin-mid-stream (line 269) remains active per its rung-3 attribution. No threshold constants (PIN_FOLLOW_MAX_DISTANCE_PX etc.) touched in the delta.

3. No r5 leakage. Grepped the full r3->r4 diff for monotonic/calibrat/per-bucket/estimate-accuracy/virtualizer-height terms: zero hits. r5's per-session per-bucket virtualizer calibration and monotonic-forward clamp (commits b64126777/d31b24cb9 on this branch) are absent from r4's diff, as expected for r4's scope (frame pipeline only).

4. Round-2's load-bearing claims still hold: the r3->r4 diff scanned for this review is unchanged from what round 2 examined (the only new content is the inherited r2 blank-fallback fix, unrelated to pipeline internals), so the synchronous RO snap/guard analysis, the one-snap-per-frame invariant via glueActive coalescing, the 500ms compensation-deadline parity argument versus r3, glue semantics (250ms cap, cancelFramePipeline registration, marker recording through notifyProgrammaticScroll), and metricsAfterFrame's post-paint honesty all still apply unchanged to d2dd156e8.

5. PR body: still accurate. Pipeline architecture diagram, Q3/Q7/Q12 references, and the CI-evidence history (the earlier chromium pinned-follow flake at 5680242af that round 2 resolved) match the current head; no stale claims found.

No blockers found. Not merging — this is Pablo's call per the no-merge-without-review policy.

pablonyx:

Branch chat-scroll/r4-frame-pipeline checked out, product-client dependencies built via pnpm shared:build. Ran each touched test file individually with pnpm vitest run <file> --reporter=basic:

- src/hooks/chat/ui/transcript-frame-pipeline.test.ts: Test Files 1 passed (1), Tests 7 passed (7)
- src/hooks/chat/ui/use-transcript-stick-to-bottom.compensation.test.tsx: Test Files 1 passed (1), Tests 6 passed (6)
- src/hooks/chat/ui/use-above-change-compensation.test.tsx: not present on this branch head (superseded/renamed to the compensation.test.tsx file above per this PR's diff)

Negative control (synchronous same-frame snap): in transcript-frame-pipeline.ts requestFrame(), moved `writer.runFramePass()` from running inline to running inside the guard-reset rAF callback (i.e. deferred the snap to the next frame instead of the same frame). Reran transcript-frame-pipeline.test.ts. Literal failure:

```
❯ src/hooks/chat/ui/transcript-frame-pipeline.test.ts (7 tests | 3 failed)
  expected +0 to be 1 // Object.is equality (multiple assertions, incl. the SYNCHRONOUSLY-labeled test)
Test Files  1 failed (1)
     Tests  3 failed | 4 passed (7)
```

Restored requestFrame() to the original inline-then-arm-guard ordering via Edit; `git status` confirms working tree clean, no diff remains.

pablonyx:
Stack refold receipt (overnight staging): rebased onto main @ d5d38a804 (r1 merged) atop the retargeted stack. git range-diff vs previous head: 0 semantic changes (10/10 commits identical). New head a7be0c1df. Review carried from 11ce51a6c.

pablonyx:
Merge-train refold receipt: rebased onto main @ 178b79c99 (post-r3 squash). a7be0c1df -> 45fb47e0f. range-diff: 10/10 commits byte-identical (=), review carries. Note: this rung carries the flagged webkit-prepend strand (old head failed spec line 583 twice); if the fresh round reds on physics it goes to the orchestrator before any merge, per commitment.

PR #1946, opened 2026-08-15
perf(product-client): composition-derived virtualizer estimates with per-row-key height persistence (PRO-187)

pablonyx:
## Summary

Rung 5 of the Chat Scroll + Positioning ladder (frozen ADR, consequence default Q5). Stacked on #1945 (base: `chat-scroll/r4-frame-pipeline`).

Two changes to the virtualizer's row-height guessing, per the frozen spec:

1. **Composition-derived estimates** (`transcript-row-height-estimate.ts`): replaces the flat `ESTIMATED_TURN_HEIGHT_PX = 360` fallback with a bucket mapping from a row's rendered **composition** — item count and item kinds on its `TurnDisplayBlock[]` — not from actual prose text length. The virtual-row abstraction the measurement model sees (`TranscriptRenderableRow` / `TurnDisplayBlock`) carries item **identities**, not item **content** (no `text` field visible at this layer), so a literal character-count bucket would require threading full `TranscriptState` through the measurement model — a materially larger change out of this rung's scope. history-loading (32px) and goal-event (28px) special cases are preserved as their own composition classes, unchanged.

2. **Per-row-key measured-height persistence** (`transcript-row-height-cache.ts`): an in-memory, bounded (500 rows/session) LRU `Map<sessionKey, Map<rowKey, {px, compositionToken}>>`, written on real TanStack measurement (`measureElement` override) and consulted by `estimateSize` before the composition estimate. Same house pattern as `assistant-reveal-progress.ts`. Remount-scoped only — no localStorage, no durable persistence (that's rung 6's territory).

3. **Wire-through**: `estimateSize` consults (a) persisted measurement for the row key + composition token, (b) composition estimate, in that order. `scrollToIndex` stays banned. No changes to snap/pipeline/classification logic (rungs 3-4) or restore behavior (rung 6).

### Estimation bucket table

| Composition | Estimate (px) | Rationale |
|---|---|---|
| history_loader | 32 | unchanged special case |
| goal_event | 28 | unchanged special case |
| single plain block (1 message/prose/thought) | 120 | short turn, ~1-3 lines |
| 2-3 plain blocks | 220 | short back-and-forth of item rows |
| 4+ plain blocks | 360 (old flat value) | the "long unsplit turn" shape the original constant was tuned for |
| inline_tool / inline_tools block | 56 × item count | compact single-line-ish inline tool rows |
| collapsed_actions (≤3 items) | 56 | small collapsed ledger |
| collapsed_actions (4-8 items) | 88 | medium collapsed ledger |
| collapsed_actions (9+ items) | 120 | large collapsed ledger — header height barely grows with count when collapsed |
| subagent_creations (≤2) | 72 | small subagent group card |
| subagent_creations (3+) | 96 | larger subagent group card |
| completed-history chunk | 44 | compact single disclosure summary regardless of folded block count |
| mixed composition | sum of per-block estimates | e.g. item + inline_tool |

Coarse buckets by design — "do not fabricate precision" per the frozen spec. None of this changes actual rendered height, only the virtualizer's up-front guess before real measurement.

### Cache design

- Key: `sessionKey` (`${workspaceId}:${sessionId}`, same identity used by `useTranscriptStickToBottom`) → `rowKey` → `{px, compositionToken}`.
- Bound: 500 rows per session (LRU via delete+re-set on write), matching `assistant-reveal-progress.ts`'s cap. Sessions themselves are not capped.
- Invalidation: each entry carries a composition token captured at write time (`getRowCompositionToken`). A lookup whose current token doesn't match is treated as a miss and the stale entry is dropped. For `turn` rows this reuses the SAME reference-stability guarantee the existing turn-row cache (`transcript-row-cache-key.ts` / `isTranscriptTurnRowCacheHit`) already provides: `renderPresentation` is the same object reference across renders whenever the turn's cache key is unchanged, and a new object whenever it changes. Other row kinds use a best-effort proxy token (adequate — none of them are the tall, bounce-prone rows this rung targets).
- Scope: remount-scoped only. No localStorage, no durable persistence.

### Unit (vitest, targeted per spec: `src/hooks/chat/ui` + `src/domain/chats/transcript`)

```
Test Files  48 passed (48)
     Tests  335 passed (335)
```

New files: `transcript-row-height-estimate.test.ts` (11 tests: bucket mapping for text-only short/long, collapsed tool-group small/large, subagent group, completed-history summary, mixed item+inline_tool composition, plus composition-token identity/invalidation) and `transcript-row-height-cache.test.ts` (7 tests: read/write, session scoping, composition-token invalidation, non-finite/non-positive rejection, LRU recency refresh, 500-row bound).

### Playwright scroll-physics (chromium, 3x repeat per local-flake protocol; CI/webkit is the authority)

Added `appendCollapsedToolLedgerTurn` / `seedConversationWithToolLedger` / `sweepEveryRowIntoView` to the fixture host (`scroll-physics-host.ts`) and a new spec: **"estimate accuracy: an off-screen collapsed tool-ledger row's estimate tracks its real measured height."**

Seeds a conversation where a 30-tool-call turn (folded into ONE `collapsed_actions` display block) starts virtualized off-screen, buried under 40 filler turns, pinned to the bottom. At that point the virtualizer's total content size is built entirely from estimates for the buried rows. `sweepEveryRowIntoView` then steps through the whole transcript so every row is measured for real, giving a ground-truth total. The assertion bounds the gap between the all-estimated total and the all-real-measured total (`estimateError < 1400`).

Literal output, composition estimate (this PR), 3x chromium repeat:
```
✓ estimate accuracy: ... (1.2s)
✓ estimate accuracy: ... (1.2s)
✓ estimate accuracy: ... (1.2s)
```
Underlying deltas observed across runs: 916-960px.

**Negative control**: temporarily forced `estimateTurnRowHeight` back to the OLD flat `360`px guess for every turn row, rebuilt (`pnpm shared:build` + fixture build), reran the same test 3x:
```
Error: expect(received).toBeLessThan(expected)
Expected: < 1400
Received:   2690
1 failed
```
(identical `2690` across all 3 runs). Reverted immediately after confirming the failure; the committed code has no trace of the flat-estimate probe.

Ran the full chromium scroll-physics suite before and after this change. Two pre-existing flaky tests ("pinned-follow", "repin band edge") were also observed failing intermittently on unmodified `chat-scroll/r4-frame-pipeline` HEAD (confirmed via `git stash` + rerun) — unrelated to this PR, not introduced by it. All other scenarios (pinned-follow baseline, unpin-mid-stream, repin-band, single-writer, no-false-unpin, prepend-anchoring) pass, and the new estimate-accuracy test is stable across repeats.

### report_frontend_structure.py

```
TOTAL: 0
```
`VirtualizedTranscriptRowList.tsx` was already over the 400-line soft threshold on r4 (419 lines) before this PR; +2 lines wiring `measureElement` through pushed it to 421. Registered both it and the pre-existing, untouched `use-transcript-stick-to-bottom.ts` (488 lines, also already over threshold on r4) in `scripts/frontend_structure_allowlist.txt` so `--strict` stays green — neither is new debt from this PR.

## Deviations / assumptions

- **ASSUMPTION**: "prose-length buckets" are approximated via a block-count proxy (1 / 2-3 / 4+ plain blocks), not literal character counts, since `TranscriptRenderableRow` doesn't carry item text at this layer (see bucket table rationale above). Threading `TranscriptState` through the measurement model to get real text length would be materially out of this rung's scope.
- No changes to rung 3/4 snap/pipeline/classification logic or rung 6 restore behavior.
- `scrollToIndex` remains unused by any new code path.

## Founder questions

- **NORMATIVE (measurement-fed calibration required)**: composition-derived estimates REQUIRE measurement-fed per-bucket calibration; block-count proxy alone undershoots tall turns (~430px measured vs ~120 estimated, honest fixtures, r5). The proxy is a bootstrap only — it must be corrected by the per-session per-bucket measured heights this rung persists, or tall turns are systematically under-estimated on first paint.
- No other material ambiguity encountered beyond the documented prose-length assumption above.

### DIFF
--- a/apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts
+++ b/apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts
@@ -678,4 +678,51 @@ test.describe("transcript scroll physics", () => {
       expect(after.scrollTop).toBeGreaterThan(before.scrollTop + 20);
     },
   );
+
+  // Rung 5 (PRO-187): composition-derived virtualizer estimates +
+  // per-row-key measured-height persistence. A collapsed tool-ledger row (30
+  // small tool calls folded into ONE `collapsed_actions` display block, see
+  // appendCollapsedToolLedgerTurn) renders as a compact disclosure — its REAL
+  // height is nowhere near the old flat 360px-per-row guess used for every
+  // other non-special-cased row.
+  //
+  // seedConversationWithToolLedger seeds the ledger turn INTO THE FIRST
+  // commit, buried under enough finalized filler turns that it starts
+  // virtualized OFF-SCREEN (well outside the pinned-bottom viewport +
+  // overscan window). At that moment the virtualizer's TOTAL content size —
+  // the scrollbar geometry the reader sees before ever touching that row —
+  // is built entirely from ESTIMATES, not real measurements.
+  // sweepEveryRowIntoView then steps through the WHOLE transcript so every
+  // row mounts and is measured for real at least once, giving a ground-truth
+  // total. The gap between the all-estimated initial total and the
+  // all-real-measured total is the virtualizer's literal "correction
+  // budget": how far off the guess was, and therefore how large a correction
+  // the frame pipeline has to absorb as those rows come into view.
+  //
+  // Negative control (see PR body for the literal run): temporarily forcing
+  // `estimateTurnRowHeight` in transcript-row-height-estimate.ts back to the
+  // OLD flat 360px-per-row guess makes this same gap ~2.5-3x larger and
+  // fails the bound below.
+  test("estimate accuracy: an off-screen collapsed tool-ledger row's estimate tracks its real measured height", async ({
+    page,
+  }) => {
+    await ready(page);
+    await drive(page, "reset");
+    await drive(page, "seedConversationWithToolLedger", 2, 40, 30);
+    await waitForViewport(page);
+
+    const estimated = await metrics(page);
+    expect(estimated.found).toBe(true);
+    // A short viewport against many filler turns must actually overflow and
+    // bury the ledger row off-screen, else this test proves nothing.
+    expect(estimated.scrollHeight).toBeGreaterThan(estimated.clientHeight + 1000);
+
+    await drive(page, "sweepEveryRowIntoView", 40);
+    await settle(page, 300);
+    const real = await metrics(page);
+    expect(real.found).toBe(true);
+
+    const estimateError = Math.abs(estimated.scrollHeight - real.scrollHeight);
+    expect(estimateError).toBeLessThan(1400);
+  });
 });
--- a/apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts
+++ b/apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts
@@ -680,29 +680,33 @@ test.describe("transcript scroll physics", () => {
   );
 
   // Rung 5 (PRO-187): composition-derived virtualizer estimates +
-  // per-row-key measured-height persistence. A collapsed tool-ledger row (30
-  // small tool calls folded into ONE `collapsed_actions` display block, see
-  // appendCollapsedToolLedgerTurn) renders as a compact disclosure — its REAL
-  // height is nowhere near the old flat 360px-per-row guess used for every
-  // other non-special-cased row.
+  // per-row-key measured-height persistence + per-session per-bucket
+  // calibration. seedConversationWithToolLedger seeds a collapsed tool-ledger
+  // turn under many tall finalized filler turns, all buried OFF-SCREEN below
+  // the pinned-bottom viewport + overscan window. With the fixture now
+  // hydrating those finalized turns inert (turn_ended, matching production),
+  // each filler turn renders at its full tall height rather than the shorter
+  // mid-reveal height the old reveal-inflated fixture happened to show. Against
+  // honest tall heights the STATIC composition estimate undershoots a plain
+  // multi-block turn badly (measured ~430px vs a ~220px static guess), so the
+  // all-static initial total is far below the real swept total.
   //
-  // seedConversationWithToolLedger seeds the ledger turn INTO THE FIRST
-  // commit, buried under enough finalized filler turns that it starts
-  // virtualized OFF-SCREEN (well outside the pinned-bottom viewport +
-  // overscan window). At that moment the virtualizer's TOTAL content size —
-  // the scrollbar geometry the reader sees before ever touching that row —
-  // is built entirely from ESTIMATES, not real measurements.
-  // sweepEveryRowIntoView then steps through the WHOLE transcript so every
-  // row mounts and is measured for real at least once, giving a ground-truth
-  // total. The gap between the all-estimated initial total and the
-  // all-real-measured total is the virtualizer's literal "correction
-  // budget": how far off the guess was, and therefore how large a correction
-  // the frame pipeline has to absorb as those rows come into view.
+  // Per-session per-bucket calibration closes that: as the first on-screen
+  // filler turns measure for real, their heights feed a running average keyed
+  // by composition bucket (transcript-row-height-calibration.ts), and every
+  // never-measured filler of the same shape borrows that average instead of the
+  // static default. sweepEveryRowIntoView then steps through the WHOLE
+  // transcript so every row mounts and is measured for real at least once,
+  // giving a ground-truth total. The gap between the calibrated estimated total
+  // and the all-real-measured total is the virtualizer's literal "correction
+  // budget": how far off the settled guess was.
   //
-  // Negative control (see PR body for the literal run): temporarily forcing
-  // `estimateTurnRowHeight` in transcript-row-height-estimate.ts back to the
-  // OLD flat 360px-per-row guess makes this same gap ~2.5-3x larger and
-  // fails the bound below.
+  // Negative control (proven deterministically in the colocated unit tests,
+  // use-transcript-virtual-measurement-model.test.ts): with calibration removed
+  // from estimateSize, a never-measured filler falls back to the static
+  // composition estimate (~220px), which undershoots the honest ~430px height
+  // by ~210px per row and blows this gap well past the 1400px bound below. The
+  // calibrated estimate is therefore load-bearing, not decorative.
   test("estimate accuracy: an off-screen collapsed tool-ledger row's estimate tracks its real measured height", async ({
     page,
   }) => {
@@ -710,6 +714,10 @@ test.describe("transcript scroll physics", () => {
     await drive(page, "reset");
     await drive(page, "seedConversationWithToolLedger", 2, 40, 30);
     await waitForViewport(page);
+    // Let the first on-screen measurement pass populate the per-bucket
+    // calibration so the off-screen filler estimates settle to this session's
+    // observed heights before the estimated total is read.
+    await settle(page);
 
     const estimated = await metrics(page);
     expect(estimated.found).toBe(true);
--- a/apps/packages/product-client/qualification/scroll-physics/src/scroll-physics-host.ts
+++ b/apps/packages/product-client/qualification/scroll-physics/src/scroll-physics-host.ts
@@ -243,6 +243,7 @@ function collapsedToolLedgerTurn(
   envelopes.push(
     assistantCompleted(sessionId, turnId, assistantItemId, "Done reading through the files."),
   );
+  envelopes.push(turnEnded(sessionId, turnId));
   return envelopes;
 }
 
@@ -522,6 +523,7 @@ export const scrollPhysicsDriver: ScrollPhysicsDriver = {
       const assistantItemId = `${turnId}-assistant`;
       batch.push(userMessage(id, turnId, `Prompt ${t}: please continue.`));
       batch.push(assistantCompleted(id, turnId, assistantItemId, tallText(`Reply ${t}`)));
+      batch.push(turnEnded(id, turnId));
     }
     batch.push(...collapsedToolLedgerTurn(id, `${id}-ledger`, toolCallCount));
     for (let t = 0; t < trailingTurns; t += 1) {
@@ -529,6 +531,7 @@ export const scrollPhysicsDriver: ScrollPhysicsDriver = {
       const assistantItemId = `${turnId}-assistant`;
       batch.push(userMessage(id, turnId, `Prompt trail ${t}: please continue.`));
       batch.push(assistantCompleted(id, turnId, assistantItemId, tallText(`Trail ${t}`)));
+      batch.push(turnEnded(id, turnId));
     }
     state = reduceEventBatch(state, batch);
     commit({
--- a/apps/packages/product-client/qualification/scroll-physics/src/scroll-physics-host.ts
+++ b/apps/packages/product-client/qualification/scroll-physics/src/scroll-physics-host.ts
@@ -179,6 +179,73 @@ function largeToolInvocation(sessionId: string, turnId: string): SessionEventEnv
   return [started, completed];
 }
 
+// A single small, completed bash tool call — used in bulk (see
+// `collapsedToolLedgerTurn`) to build one turn with a long run of
+// individually-tiny actions that the presentation layer folds into ONE
+// `collapsed_actions` display block (see isCollapsibleAction in
+// transcript-presentation.ts). This is the "long tool-ledger-like row" the
+// rung 5 estimate-bounce fixture targets: real height stays a compact,
+// mostly-count-independent disclosure, while the OLD flat 360px-per-row
+// estimate assumed every row was itself full-turn-sized.
+function smallToolCall(sessionId: string, turnId: string, index: number): SessionEventEnvelope[] {
+  const itemId = `${turnId}-tool-${index}`;
+  const started = envelope(sessionId, turnId, itemId, {
+    type: "item_started",
+    item: {
+      kind: "tool_invocation",
+      status: "in_progress",
+      sourceAgentKind: "claude",
+      title: "read_file",
+      toolCallId: itemId,
+      nativeToolName: "read_file",
+      rawInput: { path: `src/file-${index}.ts` },
+      contentParts: [
+        {
+          type: "tool_call",
+          toolCallId: itemId,
+          title: "read_file",
+          toolKind: "other",
+          nativeToolName: "read_file",
+        },
+      ],
+    },
+  });
+  const completed = envelope(sessionId, turnId, itemId, {
+    type: "item_completed",
+    item: {
+      kind: "tool_invocation",
+      status: "completed",
+      sourceAgentKind: "claude",
+      title: "read_file",
+      toolCallId: itemId,
+      nativeToolName: "read_file",
+      rawInput: { path: `src/file-${index}.ts` },
+      rawOutput: "ok",
+    },
+  });
+  return [started, completed];
+}
+
+// One turn whose item count clears SPLIT_TURN_MIN_ITEM_COUNT (24) and whose
+// tool calls collapse into a single `collapsed_actions` display block.
+function collapsedToolLedgerTurn(
+  sessionId: string,
+  turnId: string,
+  toolCallCount: number,
+): SessionEventEnvelope[] {
+  const envelopes: SessionEventEnvelope[] = [
+    userMessage(sessionId, turnId, "Read through the affected files."),
+  ];
+  for (let i = 0; i < toolCallCount; i += 1) {
+    envelopes.push(...smallToolCall(sessionId, turnId, i));
+  }
+  const assistantItemId = `${turnId}-assistant`;
+  envelopes.push(
+    assistantCompleted(sessionId, turnId, assistantItemId, "Done reading through the files."),
+  );
+  return envelopes;
+}
+
 // A tall fenced code block inside assistant prose. This renders through the
 // real MarkdownCodeBlock, which owns its OWN inner `overflow-y-auto` region,
 // the nested-scroll surface the chaining scenario needs.
@@ -339,6 +406,25 @@ let traceFrames: number[] = [];
 let traceRafId = 0;
 let traceActive = false;
 
+// Per-frame virtualizer TOTAL content height (scrollHeight) trace recorder —
+// the rung 5 (PRO-187) estimate-bounce fixture's signal. An estimate error
+// shows up here as a transient spike/dip between the guessed size and the
+// real measured size, at rAF resolution (finer than a JS-side polling loop
+// can reliably observe, since the estimate -> real-measurement swap can land
+// within the same frame the row mounts).
+let heightTraceFrames: number[] = [];
+let heightTraceRafId = 0;
+let heightTraceActive = false;
+
+function heightTraceTick(): void {
+  if (!heightTraceActive) {
+    return;
+  }
+  const el = viewport();
+  heightTraceFrames.push(el ? el.scrollHeight : Number.NaN);
+  heightTraceRafId = requestAnimationFrame(heightTraceTick);
+}
+
 function traceTick(): void {
   if (!traceActive) {
     return;
@@ -353,6 +439,7 @@ export interface ScrollPhysicsDriver {
   readonly secondarySession: string;
   reset(sessionId?: string): void;
   seedFinalizedConversation(turns: number, sessionId?: string): void;
+  seedConversationWithToolLedger(leadingTurns: number, trailingTurns: number, toolCallCount?: number): void;
   setHasOlderHistory(hasOlder: boolean, reservoirTurns?: number): void;
   beginStreamingTurn(): void;
   streamChunk(text?: string): void;
@@ -360,11 +447,14 @@ export interface ScrollPhysicsDriver {
   finalizeStreamingTurn(): void;
   appendLargeToolOutput(): void;
   appendCodeBlockTurn(): void;
+  appendCollapsedToolLedgerTurn(toolCallCount?: number): string;
   appendFinalizedTurns(turns: number): void;
   prependOlderHistory(turns?: number): void;
   switchSession(sessionId: string, seedTurns: number): void;
   getMetrics(): ViewportMetrics;
   scrollToBottomInstant(): void;
+  scrollToTopInstant(): void;
+  sweepEveryRowIntoView(stepDelayMs?: number): Promise<void>;
   gestureScrollToBottomDistance(distancePx: number): void;
   // True pin state, read from the floating "Scroll to bottom" control, which is
   // aria-hidden exactly when pinned to bottom. Returns null if the control is
@@ -375,6 +465,8 @@ export interface ScrollPhysicsDriver {
   clearScrollSamples(): void;
   startScrollTrace(): void;
   stopScrollTrace(): number[];
+  startContentHeightTrace(): void;
+  stopContentHeightTrace(): number[];
   hasViewport(): boolean;
 }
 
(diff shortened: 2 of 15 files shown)


pablonyx:
## Corrective pass — rebased onto fixed rung 4, one estimateSize regression fixed (rung 5)

Rebased onto the corrected rung-4 tip (no conflicts). Two fixes on top of the estimates work:

1. **`c62b53c04`** — dropped two now-stale `frontend_structure_allowlist.txt` entries (`VirtualizedTranscriptRowList.tsx`, `use-transcript-stick-to-bottom.ts`); both dropped under the 400-line cap after the rung-3 splits, so the entries would fail the ratchet as stale.
2. **`90d65a1f5`** — the composition-derived `estimateSize` read `renderableRows[index]` and listed `renderableRows` in its dep array, so a content-only row snapshot (new array identity, same ordered keys) rotated the accessor. That violated the documented accessor-stability contract (rotate only on composition/session-scope change) and broke `VirtualizedTranscriptRowList.test.tsx > rotates measurement accessors only when ordered row composition changes`. Now the live rows are read through a ref so the persisted-height lookup stays fresh while the callback identity stays pinned. Deterministic negative control: before the fix that test failed (`estimateSize` reference mismatch); after, it passes.

### Literal local results at this tip (`90d65a1f5`)

- Targeted vitest (`src/hooks/chat/ui` + `transcript`): **443 passed** (was 1 failed before fix #2).
- Physics `--project=chromium`, 2 consecutive full runs: **8 passed / 8 each run**.
- Physics `--project=webkit`: **7 passed** (3 conditional skips) — including `pinned-follow` and `repin band edge`, both green.

Not merging.

pablonyx:
## Corrective pass: r5 follow-cadence regression fixed

**Literal CI at r5 before this change:** FAIL — `pinned-follow` bottomDistance 132 on both engines; `[webkit] repin-band-edge` (Expected: <= 202 / Received: 332); `[chromium] prepend-anchoring` 550; 11 passed.

**Root cause (r5-specific, the pinned-follow 132 + webkit repin 332):** the composition-derived estimate model keyed `getItemKey` and `estimateSize` IDENTITY on a signature that embeds each row's composition estimate (`measurementSignature` = ordered `[key, compositionEstimate]`). A streaming turn changes its display-block shape every chunk, so that estimate — and the signature — rotated every chunk, rotating both accessor identities. TanStack memoizes its measurement derivation on those identities and rebuilds every item position when either rotates; that extra layout pass landed a frame after the single per-frame snap while pinned, stranding the snap one measurement-generation behind the growing bottom (the r3-era follow lag rung 4 had fixed). This is the brief's "measurement-model invalidation triggering an extra layout pass after the snap."

**Fix (rung-5 scope, FR-1):** rotate `getItemKey`/`estimateSize` identity ONLY on a change to the ordered set of row KEYS (`keySignature`), never on estimate values. The estimate value still updates — read fresh from the live row ref at call time — so unmeasured rows still get the better guess; estimates feed the pipeline without triggering out-of-band re-derivations. Correctness never rested on the guess: the on-screen streaming row is measured, and the persisted-measurement cache is invalidated by `getRowCompositionToken`, not by accessor identity. `rowCompositionKey` (consumed by the anchor-capture cleanup) still tracks composition and is unchanged.

**The 550 / 332 numbers:** the `prepend-anchoring` 550 and (partly) webkit repin were the r4 defect inherited pre-rebase. This branch is now rebased onto the fixed r4 (#1945), which carries the deadline-gated compensation fix.

**Tests:**
- New deterministic unit negative control `use-transcript-virtual-measurement-model.test.tsx`: accessor identity holds across an estimate-only change (with a fresh value) and rotates on a key-set change. Negative-controlled: re-keying the accessors on `measurementSignature` fails it (expected [Function] to be [Function]).

**Local results (this machine, headless, workers=1):**
- Full physics suite x3 after rebase: 44 passed, chromium all green including throttled `prepend-anchoring` and both-engine `repin-band-edge`.
- webkit `pinned-follow` flakes locally (1/3 in-suite; 4/5 isolated) — the SAME flake reproduces identically on the r4 baseline (also 4/5 isolated), so it is a local-webkit-vs-CI gap, not an r5 regression. CI is the authority.
- Unit: 177 passed across `src/hooks/chat/ui`; typecheck + scroll-physics typecheck clean.

CPU throttling did not reproduce the pinned-follow 132 locally even at 8x — a CI-scheduling-specific failure, so the deterministic guard is the unit test above. Never merged.

pablonyx:
Status at head 784c3da7d (rebased onto r4 5680242af, run 31897943865). NOT green yet: three arms fail.

Literal CI:
- prepend chromium: Expected > 150 / Received 120 (the > 150 scrollTop arm; the ratio arm needs > addedAbove*0.6).
- pinned-follow chromium: Expected <= 120 / Received 132.
- pinned-follow webkit: Expected <= 120 / Received 132.

pinned-follow 132 is the same shared one-block follow lag tracked in PR #1945 (inherited pipeline); see that thread.

r5-specific prepend root cause: the anchor compensation applies `scrollTop = anchor.scrollTop + (viewport.scrollHeight - anchor.scrollHeight)` with real DOM metrics. r5's composition-derived estimates plus the write-through measured-height cache make the virtualizer's TOTAL scrollHeight move non-monotonically during the prepend settle. Under CDP CPU throttle rate 6 the added-above corrections spread over ~3s, and the total transiently dips below the captured anchor height, so the compensation delta goes small or negative and the reader lands near the prepended top (scrollTop 120, and negative deltas in intermediate attempts). A fixed 500ms deadline releases before the total finishes climbing to its true value; a growth-refreshed deadline was tried and made it worse (held into the oscillation, inverting to -62), so it was reverted.

This is a genuine estimate-model / anchor-timing interaction that needs dedicated work: the anchor should track the reader against a monotonic added-above measure (not the oscillating virtualizer total), or the estimate accessors must not let the total shrink below the pre-prepend height mid-settle. Beyond the 3-round budget of this pass. Not merging.

pablonyx:
CI verdict on r5 (`e7599e2c9`, rebased onto the green r4 `83969e531`): Transcript scroll physics FAILED, 11 passed / 4 failed / 5 skipped (1.6m). Literal numbers:

1. chromium pinned-follow, spec line 257 (the in-loop read, now frame-aligned via `metricsAfterFrame`): Expected <= 120, Received 132.
2. chromium prepend anchoring, spec line 624: Expected > 150, Received 120.
3. webkit pinned-follow, spec line 234 (the seeded baseline read, plain `metrics`, before any streaming): Expected <= 120, Received 132.
4. webkit repin band edge, spec line 154 (`gestureToBottomDistance` poll): Expected <= 202, Received 332.

Two conclusions, reported per the ladder-owner's stop directive:

Measurement theory does NOT hold at r5. The identical frame-aligned sampling that made pinned-follow fully green at r4 still reads 132 at line 257 on r5 (item 1). Since the only delta from green r4 is r5's product changes (composition-derived estimates + write-through measured-height cache), r5 reintroduces a real per-painted-frame 132 in the pinned-follow path rather than a pure sampling artifact. Per the directive ("if pinned-follow STILL shows 132 with frame-aligned sampling, report the literal numbers and stop; do not churn product code"), I am stopping, not iterating further on the follow path.

The rung-5 prepend clamp did NOT resolve item 2. The monotonic running-max clamp I added (`use-transcript-frame-pipeline-lifecycle.ts`) prevents a transient scrollHeight dip from jerking the reader up (proven by a deterministic unit negative control: without the clamp scrollTop lands at 1200 instead of 1400). But CI chromium still reads scrollTop 120 vs > 150, i.e. under-compensation where the effective total barely grows, a different failure mode than the dip my clamp defends. The clamp is a correct, tested hardening but is not the fix for this regression.

Not green, so not merging. r4 (#1945) is fully green and is the shippable win from this pass.

pablonyx:
## r5 scroll-physics: root cause + trail (head ab262d995)

Worked the r5 physics regressions across three honest rounds. Net result: the four original consistent chromium/webkit failures are down to one consistent failure plus one load-dependent flake, both isolated to the r5 measurement model. Repo shape is back to TOTAL: 0.

### Literal CI (Transcript scroll physics, run 31905198631, head ab262d995)
```
2 failed
  [chromium] pinned-follow: bottom distance stays ~0 across streaming growth
    Expected: <= 120
    Received:    132   (line 257)
  [chromium] prepend anchoring: older-history prepend keeps the reading row fixed
    Expected: > 150
    Received:   10     (line 624)
13 passed
```
Repo shape checks: pass. Both webkit pinned-follow and webkit prepend: pass.

### What was fixed and shipped (ab262d995)
- FIX 2 (use-transcript-frame-pipeline-lifecycle.ts): the fixed 500ms above-change compensation deadline lapsed after a couple of frame passes while a CPU-throttled runner spread the prepended rows' estimate-to-measured corrections over 1 to 2 seconds, jumping the reader toward the newly prepended top. The window now extends by a 1s quiet interval on each fresh above-anchor growth, bounded by a 3s absolute ceiling. This turned the r5 prepend failure from consistent (chromium scrollTop 120 vs > 150) into a pass in most runs, with 2 new negative-controlled unit tests. Both engines' prepend passed in round 1.
- FIX 1 (VirtualizedTranscriptRowList.tsx): the pinned snap re-runs on renderableRows identity (not only row count or the estimate-derived total) so a streaming turn's same-commit DOM growth snaps in that commit's layout phase.

### pinned-follow 132: root cause (unresolved)
Reproduced deterministically under CDP CPU throttle rate 8 (1 to 3 of 10 runs). At the failing sample bottomDistance is 132; an immediate re-read is 0, and the next two frame-aligned reads are 0. So the snap does land and the resting state is correct; a single painted frame is momentarily 132 short and self-corrects within one frame.

132 = measured (~252) minus the r5 composition estimate (120) for a single-block turn. As the seeded finalized turns scroll off-screen above the growing stream, their estimate-to-measured swap changes the virtualizer top spacer, so scrollHeight jumps +132 one frame before the follow snap catches it. On r4 the estimate sat closer to measured so the transient stayed under the 120 ceiling; the r5 composition-estimate delta pushes it to 132.

Tried and rejected (none moved the 132, verified under throttle):
- useAnimationFrameWithResizeObserver false: made it worse (2 of 10, batch 0 and 4) and did not remove the transient.
- onChange -> notifyContentResize bridge: did not move the 132 and regressed chromium prepend (onChange fires while the pin flag still lags true during older-history mount, snapping to bottom and fighting the anchor restore, scrollTopDelta -62). Reverted.
- Begin a pinned forced-glue window on content resize: still 3 of 10, unchanged transient.

The real fix belongs in the r5 estimate/cache layer, not the snap timing: the persisted measured-height cache should supersede the composition estimate for off-screen rows already measured this session so the top spacer does not jump on their swap-in, or the single-block turn bucket needs to track its realistic measured height. That is a deliberate measurement-model change and should be made as such rather than as a snap-path band-aid.

### prepend flake
Line 624 (scrollTop 10 vs > 150) is under-absorption sending the reader to the top. Same prepend code passed on chromium in round 1, so this is a load-dependent flake on the throttled runner, not a regression. FIX 2 reduced the consistent prepend failure to an occasional one; a fully robust fix likely needs the same estimate/cache correction above so the added-above height is known up front rather than absorbed frame by frame.

Not merging. Leaving head at ab262d995 as the reviewable progress state.

pablonyx:
## r5 estimate-layer pass: what landed, and a concrete contradiction to the frozen root cause

**Pushed (77488bee8), both physics-independent, both green in CI:**
1. Negative-controlled unit tests for the r5 measured-supersedes-estimate contract in `use-transcript-virtual-measurement-model.test.tsx`: a measured row keeps its measured size across an off-screen recycle (revert the resolution order and it drops to the composition estimate), and a fully measured conversation totals from measured heights, not composition estimates. These validate the resolution order already implemented on this branch.
2. Fixed the stale `VirtualizedTranscriptRowList` accessor-stability test to the r5 model (getItemKey rotates only on an ordered-key change; estimateSize also rotates on a session change). This unblocked the **Shared frontend packages** check, which is now green.

**I did not push a physics fix, because I could not reproduce or confirm the adjudicated root cause. Here is the contradiction, with evidence.**

The frozen root cause was: chromium pinned-follow trails by exactly 132 because 132 = measured (~252) minus the single-block composition estimate (120), from seeded finalized turns swapping estimate to measured one frame before the follow snap, so the fix belongs in the estimate/cache layer.

Runtime instrumentation of the chromium physics run does not support that. The 132/frame delta is not an estimate-to-measured swap. It is the assistant typewriter reveal animation running on the seeded history turns: the scroll-physics fixture never emits `turn_ended`, so every seeded/completed turn has `completedAt == null`, is treated by the reveal frontier as `wasLive`, and runs the typewriter reveal on mount. With ~6 seeded turns each growing ~22px/frame simultaneously, the viewport grows ~132px/frame and the single per-frame pinned snap trails that growth by exactly one frame. Emulating prefers-reduced-motion (which short-circuits the reveal to full content) removes the delta entirely, confirming reveal growth, not an estimate swap, is the cause.

**Why this is out of scope for a targeted estimate-layer fix.** The direct fix is to make the fixture emit `turn_ended` for seeded/finalized turns so history renders finalized (no reveal). I implemented and verified it locally: it fixes pinned-follow and prepend anchoring. But it then breaks two other merge-gating physics assertions that were silently calibrated against the reveal artifact:
- `estimate-accuracy` (bounds error jumps 5162 > 1400): once turns render at their honest finalized heights (~430px for tall turns), the r5 composition estimate (short-multi-block 220, flat 360) is actually less accurate than the old flat-360, inverting this test's negative-control premise.
- `no-false-unpin` (added height 196 < 300 threshold): the 300px growth this asserts only existed because of reveal inflation.

So the correct fix cascades into recalibrating two other gating tests and reopens whether the r5 composition estimate is the right model for tall turns. That is a fixture/feature decision for the PR author, not a unilateral estimate-cache edit. Raising it here rather than silently changing scope or forcing a fix that reverts other tests' intent.

**Current CI on this push:** Shared frontend packages green; Transcript scroll physics still red (unchanged, pending the decision above). Not merging.

pablonyx:
## Root-cause correction: composition-estimate accuracy vs honest heights

An earlier note here credited the block-count composition proxy as passing the estimate-accuracy scenario at the original 1400px bound. That result was against a fixture that never emitted `turn_ended`, so its seeded "finalized" turns had `completedAt == null`, counted as `wasLive`, and ran the assistant typewriter reveal on mount. The reveal inflated per-turn heights frame-by-frame, and the flat/near-flat estimate happened to track that inflated growth. This was a fixture-fidelity artifact, not a real property of the estimator.

With the fixture corrected to emit `turn_ended` (matching production hydration, so finalized turns render at their true inert full height), the composition proxy actually **undershoots** tall turns: honest filler turns render ~430px while the static block-count bucket estimates them ~120-220px, and the aggregate blew the 1400px bound (5162px observed). The flat-360 model was in fact closer under honest heights. This is the rung-5 estimate-accuracy inversion.

### Fix shipped: dynamic per-bucket height calibration
- New `transcript-row-height-calibration.ts`: per-session, per-composition-bucket running average of REAL measured heights (bounded, decayed, cleared for tests).
- `estimateSize` now consults, in order: persisted real per-row measurement -> this session's calibrated bucket average -> static composition estimate. Never-measured rows borrow the calibrated bucket value; measured rows use their persisted height.
- Calibration propagation is forced with a bounded per-session `calibrationGeneration` counter (bumped only on a bucket's 0->1 sample transition) folded into `estimateSize`'s identity, so TanStack re-derives still-off-screen rows exactly once per newly-calibrated bucket without churning the per-frame snap.
- estimate-accuracy passes at the original **1400px** bound with calibration on; the negative control (static estimate, calibration disabled) still fails, so the assertion remains meaningful.

## Open founder questions
- **Composition-estimate default for tall turns.** The block-count composition proxy was calibrated against reveal-inflated fixtures. Under honest heights it undershoots tall turns badly enough that a static proxy alone cannot hold the accuracy bound; dynamic per-session bucket calibration was required to recover it. Flag: should the ADR's composition-estimate default be revisited for tall turns (e.g. adopt calibrated buckets as the baseline rather than a rung-5 refinement, or raise the static defaults for high-block-count buckets)?

pablonyx:
## Final status — scroll-physics green (PRO-187, r5)

Pushed `6d6a63be4`. `Transcript scroll physics (chromium + webkit)`: **pass**.

### Root cause fixed here — WebKit prepend regressed to scrollTop 0 (distinct from r2's)
r5's per-session per-bucket measured-height calibration + measured-height cache makes the virtualizer's reported total-size **non-monotonic** on webkit: right after older rows insert, the measured-height swap can momentarily report the total at or below its pre-insert value. The above-change compensation then computed a negative target and wrote the reader to `scrollTop 0` (CI round 1: `[webkit]` prepend reading-row scrollTop 0). This is **not** a remount — the downloaded Playwright trace showed virtual mode throughout, no `blank_viewport` — so it is a compensation under-absorption, distinct from r2's blank-fallback remount.

Fix (`use-transcript-frame-pipeline-lifecycle.ts`): clamp the prepend compensation **monotonic-forward** — above-change compensation only ever absorbs growth above the reader, so never move the viewport backward past its current scrollTop:

```
if (target > viewport.scrollTop) { viewport.scrollTop = target; }
```

CI verdict after the fix: **pass** on both engines (local webkit is too fast to reproduce the CI swap timing — CI is the source of truth here, and it is green).

All four cadence/prepend scenarios active and green. Thresholds unchanged.

pablonyx:
Adversarial review round 1: CLEAN (with logged latent risks, none blocking)

CI: `gh pr checks 1946` all pass, incl. "Transcript scroll physics (chromium + webkit) pass". Reviewed r4..r5 delta only (head 6d6a63be4).

(1) LRU + invalidation: correct. Cache keyed `${workspaceId}:${sessionId}`, row map keyed by row.key, cap 500 with delete-then-set giving write-order eviction. Cross-session isolation holds. Invalidation token for turn rows is `renderPresentation` (same object reference guaranteed stable by isTranscriptTurnRowCacheHit, new object when the turn cache key changes), so a content/shape edit drops the stale entry on the next lookup (`entry.compositionToken !== compositionToken` -> delete -> null). No stale measured height can survive a turn edit. Two documented, acceptable soft spots: (a) reads do not promote LRU order, so it is FIFO-on-write, not true access-LRU (a hot row never re-measured could age out under a >500-row session); (b) non-turn tokens (pending/outbox) are string proxies (`clientPromptId`+`hostsWorkspaceReceipt`), so a pending-prompt reshape at constant receipt flag keeps a stale height, but those are fixed-shape composer rows, not the tall bounce-prone targets.

(2) Calibration poisoning: bounded, non-fatal. Running mean sum/count, cap 64 with `sum = sum - mean + px` rolling decay (tested: "keeps tracking without unbounded accumulation past the sample cap"). Div-by-zero guarded (`count <= 0` -> null, px<=0 and null-bucket rejected on record). Cross-session leakage impossible (session-scoped, tested). A single 4000px outlier only moves the mean by (px-mean)/count and only ever feeds ESTIMATES of never-measured rows - a row measured for real always uses its exact persisted height, never the average. Worst case is estimate bounce, the exact thing the rung mitigates, not a correctness regression.

(3) Monotonic-forward clamp: sound for the intended path, one narrow latent risk. The clamp (`if (target > viewport.scrollTop) scrollTop = target`, `effectiveScrollHeight = running max`) is correctly scoped to the live above-change compensation anchor only (gated on `compensationAnchorRef.current`); bottom-snap/scrollToBottom writes go through a separate path and are not suppressed. Its correctness rests on the invariant that prepended older rows only ever correct TALLER (estimates deliberately undershoot: 120/56px constants, empty-bucket falls back to those). Latent risk (2)+(3) interaction: a bucket poisoned HIGH could make prepended rows over-estimate, pin maxScrollHeight above real, then measure shorter; the forward clamp refuses the downward correction and the browser's own scrollTop cap settles the reader at the bottom of the prepended region (reading row pushed toward the top) instead of the anchored row. Requires a poisoned-high bucket, is bounded by the 3s absolute deadline, self-heals as real measurements accumulate, and does not occur on the cold-history path (empty bucket -> undershoot -> upward-only). Acceptable for this rung; worth a spec note that the clamp assumes undershoot-only estimates.

(4) Accessor stability: correct. estimateSize identity deps [orderedKeys, sessionKey, calibrationGeneration] hold stable across streaming chunks (key set unchanged), reading the live row via ref so the guess still updates without churning TanStack's estimate pass or the single per-frame snap. Off-screen never-measured rows are re-derived exactly once per bucket via the first-sample-only generation bump; on-screen height changes flow through real measurement (measureElement), not the estimate. Correctness rests on measurement, not the guess, as designed.

(5) Estimate-accuracy scenario honesty: the physics test (bound < 1400px over 40 honest ~430px fillers, ~35px/row budget) is backed by a discriminating unit-test negative control: estimateSize returns the calibrated 430 (`expect(...estimateSize(0)).toBe(430)`) while the static composition estimate is 220 (`estimateRenderableRowHeight(neverMeasured) === 220`) - breaking calibration flips estimateSize back to 220 and fails the assertion. The fixture now emits turn_ended so fillers hydrate inert/tall (matching prod), removing the reveal-inflation that made the old estimate look accurate. Load-bearing, not decorative.

(6) r4 single-snap: no bypass. measureElement write-through does pure module-state writes (recordMeasuredRowHeight/recordBucketMeasurement), no scroll side effects; the clamp's only scrollTop write is wrapped in notifyProgrammaticScroll; the new useLayoutEffect widening the pinned-snap dep from `renderableRows.length` to `renderableRows` re-runs the same scrollToBottom snap. All size-driven position changes still route through the one pipeline.

No concrete reproducible correctness failure found under the documented invariants. Verdict CLEAN. Not merging.

pablonyx:

Branch chat-scroll/r5-estimates-height-persistence checked out, product-client dependencies built via pnpm shared:build. Ran each touched test file individually with pnpm vitest run <file> --reporter=basic:

- src/components/workspace/chat/transcript/VirtualizedTranscriptRowList.test.tsx: Test Files 1 passed (1), Tests 8 passed (8)
- src/hooks/chat/ui/transcript-row-height-cache.test.ts: Test Files 1 passed (1), Tests 7 passed (7)
- src/hooks/chat/ui/transcript-row-height-calibration.test.ts: Test Files 1 passed (1), Tests 5 passed (5)
- src/hooks/chat/ui/transcript-row-height-estimate.test.ts: Test Files 1 passed (1), Tests 11 passed (11)
- src/hooks/chat/ui/use-transcript-stick-to-bottom.compensation.test.tsx: Test Files 1 passed (1), Tests 10 passed (10)
- src/hooks/chat/ui/use-transcript-virtual-measurement-model.test.tsx: Test Files 1 passed (1), Tests 6 passed (6)

Negative control (measured-supersedes-estimate + calibration): the test file names its own negative control ("flip the order in estimateSize (composition estimate first, persisted as fallback) and this reads 120, not 252"). In use-transcript-virtual-measurement-model.ts, temporarily reordered estimateSize to consult the calibrated/composition estimate before the persisted measurement. Reran the test file. Literal failure:

```
❯ src/hooks/chat/ui/use-transcript-virtual-measurement-model.test.tsx (6 tests | 3 failed)
  useTranscriptVirtualMeasurementModel measured-height supersedes estimate (PRO-187, r5)
    × sizes a row by its persisted measured height, and keeps it across an off-screen recycle
      → expected 120 to be 252 (estimate-first path returned the estimate instead of the persisted 252)
    × totals a fully measured conversation from measured heights, not estimates
      → expected 240 to be 552
  useTranscriptVirtualMeasurementModel per-bucket calibration (PRO-187, r5)
    × prefers a row's own persisted measurement over the bucket average
      → expected 430 to be 300
Test Files  1 failed (1)
     Tests  3 failed | 3 passed (6)
```

Restored estimateSize to the original persisted-first, calibrated-bucket-fallback ordering via Edit; `git status` confirms working tree clean, no diff remains.

pablonyx:
Stack refold receipt (overnight staging): rebased onto main @ d5d38a804 (r1 merged) atop the retargeted stack. git range-diff vs previous head: 0 semantic changes (13/13 commits identical). New head b931640f2. Review carried from 94ffd9616.

pablonyx:
Merge-train refold receipt: rebased onto main @ post-r4 squash. b931640f2 -> 728f12a74. range-diff: 13/13 commits byte-identical (=), review carries.

PR #1948, opened 2026-08-15
feat(product): draw the run graph as connected nodes in the workflow pane (PRO-278)

tairabun:
## What

The right-pane Workflow tab rendered the run's steps as a flat stack of free-floating cards, so a run read as a list rather than a flow. This PR introduces `WorkflowGraphView`, which draws the chain as a connected graph in the n8n direction the design calls for:

- Consecutive chain slots are joined by a drawn edge — a centered hairline flowing into a small arrowhead (`w-px bg-border` + `ChevronDown` at `icon-tight`, so the scaled stroke stays a full device pixel on 1x displays).
- Ad hoc side nodes hang off a branch rail (`border-l` elbow) under their anchor slot instead of a bare indent, so the fork is visible.
- Retries stay inside their slot with no edge between attempts — a second attempt is the same chain position, not a chain advance.
- The side-node indent moves from `WorkflowGraphNodeCard` into the rail container: the graph view owns the graph's geometry, the card keeps its subordinate fill.

Pure layout: control eligibility still comes solely from the transition table via `vm.controls`, and all callbacks pass through untouched.

## Why

[PRO-278](https://linear.app/proliferate-team/issue/PRO-278/improve-workflow-graph-in-right-pane) — "Improve workflow graph in right pane. Should be n8n style UI."

Based on `codex/workflows-gen2-pr7-builder-launch` (the gen-2 stack tip) because the run-view pane only exists on that stack.

## Testing

- New `WorkflowGraphView.test.tsx`: edge count between slots (none before the first / around a single slot), edges are decorative (`aria-hidden`), retries draw no intra-slot edge, side nodes render inside the branch rail while chain nodes do not, callbacks pass through.
- `pnpm vitest run src/components/workflows src/domain/workflows src/hooks/workflows` — 283 tests green.
- `pnpm exec tsc --noEmit` green; `check_frontend_boundaries`, `check_appearance_scaling`, `check_component_library`, `report_frontend_structure --strict` all clean.

## Manual test

Run a workflow, open the workspace, switch the right panel to the Workflow tab: steps now render joined by vertical edges with arrowheads; add a side node to see the branch rail.

### DIFF
--- a/apps/packages/product-client/src/components/workflows/run-view/WorkflowGraphNodeCard.tsx
+++ b/apps/packages/product-client/src/components/workflows/run-view/WorkflowGraphNodeCard.tsx
@@ -47,7 +47,11 @@ function statusDotToneFor(tone: WorkflowNodeTone): StatusDotTone {
 
 export interface WorkflowGraphNodeCardProps {
   vm: WorkflowGraphNodeVM;
-  /** Visually subordinate rendering for a side node anchored under another card: muted fill, indented. Structure, not color. */
+  /**
+   * Visually subordinate rendering for a side node anchored under another
+   * card: muted fill. Placement — the branch-rail indent — belongs to
+   * `WorkflowGraphView`, the container that owns the graph's geometry.
+   */
   secondary?: boolean;
   needsInput?: boolean;
   busy?: boolean;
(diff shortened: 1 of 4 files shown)


tairabun:
@greptile review

pablonyx:

`pnpm vitest run src/components/workflows/run-view/WorkflowGraphView.test.tsx` (apps/packages/product-client, after `pnpm shared:build`) on branch head 226cab182:

```
 ✓ src/components/workflows/run-view/WorkflowGraphView.test.tsx (8 tests) 91ms
 Test Files  1 passed (1)
      Tests  8 passed (8)
```

Hand-probe: the suite renders the run graph with real run fixtures; the 8 cases cover connected-node layout, attempt anchoring for side nodes (the follow-up commit "hang each side node off the attempt it anchors to"), and pane integration. Note for local runs: the scoped test fails at vite transform time without a fresh `pnpm shared:build` (stale-dist gotcha), which is environmental, not a PR defect.

No fix was made on this PR (already green), so no negative control applies.

PR #1950, opened 2026-08-15
feat(workflows): title node sessions with their chain index (PRO-277)

tairabun:
## Summary

- Gen-2 workflow node sessions were created untitled, and the workflow actor dispatches prompts below the HTTP layer's fallback titling, so every node session tab in the workspace header read as the bare provider name ("Codex") — indistinguishable in the ⌘⌥1-9 tab index the ticket screenshot shows. The node's identity ("NN · title", its `chainIndex` line) existed only on the right-panel graph card.
- Fix: name the session where the actor already links it. `link_workflow_columns` now carries the node card's own index line — `NN · <node title>` (one-based, zero-padded, `--` for a row with no chain position, matching `nodeIndexTitle` in the client's node-card copy) — in the same UPDATE, so link and name land atomically. The `CASE` keeps a user rename or any pre-assigned title authoritative, and later harness-provided titles already defer to an assigned title.
- The format lives on the model as `WorkflowRunNodeRecord::session_title()`; the actor passes it as the link's fourth argument. No client change needed — tabs, roster, and cmd-k already render `session.title`.

Linear: [PRO-277 — Node sessions should be indexed](https://linear.app/proliferate-team/issue/PRO-277/node-sessions-should-be-indexed)

## Testing

- Store seam (Tier 0/1, real migrations): `linking_names_an_untitled_session_and_keeps_an_assigned_title` in `workflow_links.rs` — untitled session gets the node title, an assigned title survives the link, and unlink keeps the name.
- Format pin: `session_titles_mirror_the_card_index_line` in `domains/workflows/model_tests.rs` (one-based, two-digit, `--` fallback).
- `cargo test -p anyharness-lib workflow` → 163 passed, 0 failed on this branch (lifecycle, actor, transition, store, extension suites included).
- `live/workflows/lifecycle_tests.rs` is intentionally untouched: it sits exactly at its frozen 600-line-checker cap (allowlisted 1070), so the happy-path e2e title assertion is omitted rather than raising the cap; the store-seam test covers the write and the actor's one-line wiring is visible in this diff. Flagging that trade-off explicitly for review.

### Manual verification

1. Run a workflow with 2+ nodes from the Workflows page.
2. Open the run's workspace: node session tabs read `01 · <node title>`, `02 · <node title>` instead of "Codex"/"Claude".
3. Hold ⌘ (~1s) to reveal the tab index: the ⌘⌥1-9 pills sit next to identifiable node names, matching the right-panel node cards.
4. Rename a node session tab, then fail & redo another node: the rename survives; the replacement's fresh session is titled with the same `NN` position.

## Observability

- None — no new logs or telemetry; the title rides an existing store write.

## Readiness

## Verification

- `python3 scripts/check_max_lines.py` passes with the checker and allowlist untouched (`actor.rs` at 597/600, no allowlist entry; `lifecycle_tests.rs` byte-identical to main).
- `cargo test -p anyharness-lib workflow` → 163 passed, 0 failed.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/domains/sessions/store/workflow_links.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/sessions/store/workflow_links.rs
@@ -10,18 +10,33 @@ use rusqlite::{params, OptionalExtension};
 use super::SessionStore;
 
 impl SessionStore {
-    /// Link a session to the workflow node it executes.
+    /// Link a session to the workflow node it executes, naming it with the
+    /// node's title when it has none yet. The launch path creates node
+    /// sessions untitled and prompts below the HTTP layer's fallback titling,
+    /// so this link is the one write that names them (PRO-277); riding the
+    /// same statement keeps link and name atomic, and the CASE keeps a user
+    /// rename or any other pre-assigned title authoritative.
     pub fn link_workflow_columns(
         &self,
         session_id: &str,
         workflow_run_id: &str,
         workflow_node_row_id: &str,
+        fallback_title: &str,
     ) -> anyhow::Result<()> {
+        let now = chrono::Utc::now().to_rfc3339();
         self.db.with_conn(|conn| {
             conn.execute(
-                "UPDATE sessions SET workflow_run_id = ?2, workflow_node_row_id = ?3
+                "UPDATE sessions SET workflow_run_id = ?2, workflow_node_row_id = ?3,
+                     title = CASE WHEN title IS NULL OR TRIM(title) = '' THEN ?4 ELSE title END,
+                     updated_at = ?5
                  WHERE id = ?1",
-                params![session_id, workflow_run_id, workflow_node_row_id],
+                params![
+                    session_id,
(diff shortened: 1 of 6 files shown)


pablonyx:

Scoped `cargo test -p anyharness-lib` on branch head f60818e94, per touched module:

```
workflows::model      → test result: ok. 1 passed; 0 failed (session_titles_mirror_the_card_index_line)
workflow_links        → test result: ok. 1 passed; 0 failed (linking_names_an_untitled_session_and_keeps_an_assigned_title)
session_extension     → test result: ok. 3 passed; 0 failed
live::workflows       → test result: ok. 15 passed; 0 failed
```

Hand-probe: `session_title()` yields `"01 · Plan"` / `"03 · Ship"` for indexed nodes (1-based, zero-padded) and `"-- · Side"` for unindexed side nodes; linking names an untitled session and preserves an already-assigned title rather than clobbering it.

No fix was made on this PR (already green), so no negative control applies.

PR #1961, opened 2026-08-15
fix(desktop): re-publish the dev diagnostics env on every new collector generation

pablonyx:
Slice 2a of the diagnostics-pipeline hardening program (frozen spec: diagnostics-hardening-spec, section "Slice 2a").

## Problem

`publish_dev_diagnostics_env` writes `<app dir>/diagnostics-dev.env` exactly once, at boot. When the collector child dies and the supervisor restarts it, an externally launched anyharness runtime (`ANYHARNESS_DEV_URL`) stays pinned to the dead collector generation until the dev runtime is manually restarted — this is the silent 10.5h producer outage from 2026-08-15.

## Changes

- New client constant `DEV_ENV_PATH_ENV` = `PROLIFERATE_DIAGNOSTICS_DEV_ENV_PATH` in `proliferate-diagnostics-client` (re-exported), consumed by slice 2b.
- The published snippet gains a fourth line: the file's own absolute path under that constant, so the producer can re-read the file when it changes.
- The sidecar (dev builds only) spawns a watcher over the supervisor state watch channel and re-publishes the snippet exactly once per transition into `Ready` with a new collector boot id. It keys on Ready state, not the generation counter, because degraded publishes also bump the generation and `dev_collector_env()` warns when not Ready.
- Snippet rendering/writing factored into `render_dev_diagnostics_env_snippet` / `write_dev_diagnostics_env` so the loop and format are unit-testable.

## Tests

- `sidecar::tests::dev_env_republish::snippet_carries_all_four_lines_including_its_own_path`
- `sidecar::tests::dev_env_republish::new_ready_boot_id_republishes_exactly_once` — Ready(a) → Degraded → Ready(b) rewrites the file with the new endpoint/boot id exactly once; a repeated Ready with the same boot id publishes nothing.
- `sidecar::tests::dev_env_republish::degraded_transitions_do_not_republish`
- Negative control: with the `republish()` call removed from the loop, `new_ready_boot_id_republishes_exactly_once` fails (0 publishes, stale file).

`cargo test -p proliferate sidecar::` and `cargo test -p proliferate-diagnostics-client` green.

Slice 2b stacks on this branch and consumes `DEV_ENV_PATH_ENV`.

### DIFF
--- a/anyharness/crates/proliferate-diagnostics-client/src/bridge/activation.rs
+++ b/anyharness/crates/proliferate-diagnostics-client/src/bridge/activation.rs
@@ -112,6 +112,12 @@ pub const DEV_CAPABILITY_ENV: &str = "PROLIFERATE_DIAGNOSTICS_BRIDGE_TOKEN";
 /// delivery after the first batch.
 #[cfg(debug_assertions)]
 pub const DEV_COLLECTOR_BOOT_ID_ENV: &str = "PROLIFERATE_DIAGNOSTICS_BRIDGE_COLLECTOR_BOOT_ID";
+/// Absolute path of the env snippet itself. The host rewrites that file on
+/// every new collector generation, so a producer that keeps re-reading it can
+/// re-attach after a collector restart instead of staying pinned to a dead
+/// generation.
+#[cfg(debug_assertions)]
+pub const DEV_ENV_PATH_ENV: &str = "PROLIFERATE_DIAGNOSTICS_DEV_ENV_PATH";
 
 /// Fixed: with no bridge there is nothing to advance the generation.
 #[cfg(debug_assertions)]
--- a/anyharness/crates/proliferate-diagnostics-client/src/lib.rs
+++ b/anyharness/crates/proliferate-diagnostics-client/src/lib.rs
@@ -22,6 +22,7 @@ pub use bridge::activation::{
 #[cfg(debug_assertions)]
 pub use bridge::activation::{
     DevEnvDiagnosticsBootstrap, DEV_CAPABILITY_ENV, DEV_COLLECTOR_BOOT_ID_ENV, DEV_ENDPOINT_ENV,
+    DEV_ENV_PATH_ENV,
 };
 pub use fallback::{
     parse_fallback_record_line, FallbackReason, FallbackRecordV1, FALLBACK_SCHEMA,
--- a/apps/desktop/src-tauri/src/app_config.rs
+++ b/apps/desktop/src-tauri/src/app_config.rs
@@ -453,8 +453,11 @@ mod tests {
         )
         .expect("seed write should succeed");
 
-        let record = set_app_config_api_base_url_at(&path, Some("https://new.example".to_string()))
-            .expect("set should succeed");
+        let record = set_app_config_api_base_url_at(
+            &path,
+            Some("https://new.example".to_string()),
+        )
+        .expect("set should succeed");
 
         assert_eq!(record.api_base_url, Some("https://new.example".to_string()));
         assert!(record.telemetry_disabled);
@@ -484,7 +487,8 @@ mod tests {
         )
         .expect("seed write should succeed");
 
-        let record = set_app_config_api_base_url_at(&path, None).expect("clearing should succeed");
+        let record =
+            set_app_config_api_base_url_at(&path, None).expect("clearing should succeed");
 
         assert_eq!(record.api_base_url, None);
         assert!(record.telemetry_disabled);
--- a/apps/desktop/src-tauri/src/app_config.rs
+++ b/apps/desktop/src-tauri/src/app_config.rs
@@ -453,11 +453,8 @@ mod tests {
         )
         .expect("seed write should succeed");
 
-        let record = set_app_config_api_base_url_at(
-            &path,
-            Some("https://new.example".to_string()),
-        )
-        .expect("set should succeed");
+        let record = set_app_config_api_base_url_at(&path, Some("https://new.example".to_string()))
+            .expect("set should succeed");
 
         assert_eq!(record.api_base_url, Some("https://new.example".to_string()));
         assert!(record.telemetry_disabled);
@@ -487,8 +484,7 @@ mod tests {
         )
         .expect("seed write should succeed");
 
-        let record =
-            set_app_config_api_base_url_at(&path, None).expect("clearing should succeed");
+        let record = set_app_config_api_base_url_at(&path, None).expect("clearing should succeed");
 
         assert_eq!(record.api_base_url, None);
         assert!(record.telemetry_disabled);
--- a/apps/desktop/src-tauri/src/commands/drag_drop.rs
+++ b/apps/desktop/src-tauri/src/commands/drag_drop.rs
@@ -128,8 +128,7 @@ mod tests {
         let file = dir.join("archive.zip");
         std::fs::write(&file, b"zip-bytes").expect("write temp file");
 
-        let file_entry =
-            entry_for_path(file.to_string_lossy().into_owned()).expect("file entry");
+        let file_entry = entry_for_path(file.to_string_lossy().into_owned()).expect("file entry");
         assert_eq!(file_entry.name, "archive.zip");
         assert!(!file_entry.is_directory);
         assert_eq!(file_entry.size, Some(9));
--- a/apps/desktop/src-tauri/src/commands/drag_drop.rs
+++ b/apps/desktop/src-tauri/src/commands/drag_drop.rs
@@ -128,7 +128,8 @@ mod tests {
         let file = dir.join("archive.zip");
         std::fs::write(&file, b"zip-bytes").expect("write temp file");
 
-        let file_entry = entry_for_path(file.to_string_lossy().into_owned()).expect("file entry");
+        let file_entry =
+            entry_for_path(file.to_string_lossy().into_owned()).expect("file entry");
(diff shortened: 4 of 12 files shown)


pablonyx (commented):
Reviewed against the frozen spec (`/tmp/diagnostics-hardening-spec.md`, "Slice 2a"). Read-only review, no build run locally (relying on CI's green `Cargo check & test`); would-be verdict: **Approve** — the required behavior is correctly implemented and tested, with two non-blocking robustness findings below that would be worth a fast-follow rather than a blocking change.

## What's correct

- **State-not-generation keying** (the exact bug class this slice exists to prevent): `dev_diagnostics_env_republish_loop` (`apps/desktop/src-tauri/src/sidecar/diagnostics.rs:334-350`) matches only on `State::Ready { collector_boot_id, .. }` and `continue`s on every other variant, so `Degraded` transitions (which also bump the generation counter per the spec) never trigger a publish or a warn. `dev_env_republish::degraded_transitions_do_not_republish` (`sidecar/tests.rs:349-365`) asserts this directly.
- **Exactly-once-per-new-Ready-boot-id**: verified by `new_ready_boot_id_republishes_exactly_once` (`sidecar/tests.rs:279-322`) exercising Ready(a, seeded)→Degraded→Ready(b)→Ready(b) and asserting exactly 1 publish, new content, old boot id gone.
- **Fourth snippet line / `DEV_ENV_PATH_ENV` contract for 2b**: `DEV_ENV_PATH_ENV = "PROLIFERATE_DIAGNOSTICS_DEV_ENV_PATH"` defined next to the other `DEV_*_ENV` constants (`proliferate-diagnostics-client/src/bridge/activation.rs:115-120`), re-exported under the same `#[cfg(debug_assertions)]` block as the others in `lib.rs:22-26`. `render_dev_diagnostics_env_snippet` writes it as the fourth line (`diagnostics.rs:293-312`), asserted by `snippet_carries_all_four_lines_including_its_own_path`. Nothing here would break 2b's consumption.
- **Secret hygiene**: `write_dev_diagnostics_env` opens with `.mode(0o600)` (`diagnostics.rs:318-332`); the only `tracing::info!` in `publish_dev_diagnostics_env` logs `dev_diagnostics_env` (path) and `endpoint`, never `capability` (`diagnostics.rs:257-283`). On republish the file already exists so `.mode()` is a no-op for permissions, but since it was created 0600 at boot and 0600 has no bits umask would clear anyway, this stays 0600 across rewrites.
- **cfg gating**: `spawn_dev_diagnostics_env_republisher` / the loop / snippet helpers are all `#[cfg(all(debug_assertions, unix))]` (matching the existing boot-time publish gating); the new client constant is `#[cfg(debug_assertions)]` only, matching the other three `DEV_*_ENV` constants (no unix requirement needed on the client side). Correct.
- **No visibility widening**: new helpers stay `pub(super)` inside `sidecar`, consistent with the spec's explicit ask not to widen `publish_dev_diagnostics_env`'s visibility.
- **Scope**: diff is limited to `activation.rs`, `lib.rs` (re-export), `sidecar/diagnostics.rs`, `sidecar/lifecycle.rs`, `sidecar/tests.rs` — no unrelated changes.
- **Labels**: exactly one `release:*` (`release:fix`) + all applicable `area:*` (`area:desktop`, `area:anyharness` — the touched `proliferate-diagnostics-client` crate lives under `anyharness/crates/`). Matches the repo's label set from `gh label list`.

## Findings (non-blocking)

**F1 — watcher task never observes `arm_shutdown()`/`shutdown_tx`.**
`spawn_dev_diagnostics_env_republisher` (`diagnostics.rs:334-347`) subscribes only to `supervisor.subscribe_state()`. Every other long-lived, supervisor-scoped task in this codebase (the tail lease and export lease in `diagnostics_collector/supervisor/queries.rs:124,149,172`) also subscribes to `supervisor.subscribe_shutdown()` so that `arm_shutdown()` (`supervisor/lifecycle.rs:56-63`, doc: "`shutdown_tx` cancels broker leases immediately") can cut them off promptly. This loop only exits when the `state_tx` sender drops — but the loop's own `republish` closure holds `Arc<DiagnosticsCollectorSupervisor>` (`diagnostics.rs:344-347`), i.e. the very struct that owns `state_tx`, so the sender can never drop while the task is alive. It's a self-referential keep-alive that never observes shutdown on its own.

In practice this is harmless today: `DiagnosticsCollectorSupervisor::new` and `sidecar::boot` are each called exactly once per process (`lib.rs:206-219`, `lib.rs:413` is the only `sidecar::boot` call site), so there's no accumulation, and process exit reclaims everything. But it's inconsistent with the shutdown-signal convention used elsewhere, and would turn into a real leak/duplicate-watcher problem if a future "restart runtime" path ever reconstructs the supervisor within the same process. Suggest a `tokio::select!` against `supervisor.subscribe_shutdown()` in the loop for consistency, even though nothing observable breaks today.

**F2 — narrow TOCTOU between the boot-time publish and the watcher's seed (same bug class this slice fixes, much narrower window).**
`lifecycle.rs:36-39` calls `diagnostics::publish_dev_diagnostics_env(supervisor)` then `diagnostics::spawn_dev_diagnostics_env_republisher(Arc::clone(supervisor))` back-to-back with no synchronization between them. `publish_dev_diagnostics_env` reads collector info via its own `supervisor.dev_collector_env()` call; the watcher separately seeds `published_boot_id` via `state_rx.borrow_and_update()` (`diagnostics.rs:365-370`) — two independent reads of the same mutex-guarded state taken at slightly different instants. If the collector transitions to a new Ready boot id in the window between these two reads, the watcher seeds itself with the new boot id (treats it as "already published") while the file on disk still holds the older, stale boot id from the first read — and since there's no further generation change, the file is never corrected afterward. This requires a real collector process restart to land inside a sub-millisecond window between two synchronous, non-`await`-separated calls, which is unlikely in practice, but it's precisely the failure mode the slice exists to close, just at a much smaller probability. The same independent-re-read pattern recurs inside the loop itself: `republish()` re-queries `dev_collector_env()` fresh rather than using the `collector_boot_id` already observed from the watch notification (`diagnostics.rs:373-383`), so `published_boot_id` bookkeeping and what's actually persisted to disk could in principle diverge under a fast enough flap.

Suggested hardening (not required to land this PR): have `publish_dev_diagnostics_env` return the boot id it actually wrote, and seed/update `published_boot_id` from that return value rather than from a second, independent state read.

## CI

All required checks green except `CI / Shared frontend packages` (FAILURE) — per the task brief this is a known pre-existing failure on `main` itself, unrelated to this diff (no frontend files touched). `Cargo check & test` (CI): SUCCESS. `Repo shape checks`, `Analyze (rust/python/javascript-typescript)` (CodeQL), `Validate PR title and labels` (x4), `Build SDK`, `Candidate build handoff`, Vercel preview: all SUCCESS.

Verdict: would approve. F1/F2 are worth a follow-up note but don't block — the core exactly-once-per-new-Ready-boot-id contract that 2b depends on is correctly built and tested, and CI is green modulo the pre-existing `Shared frontend packages` failure.

pablonyx:
Stack rebased onto current main (post-#1959) to cure the base-drift "Shared frontend packages" red. Patch identity proven: `git range-diff` shows all 3 commits `=` (8115ccc9f=bc5da94db, 4c259d338=db99ab978, 0a54f1e7f=d4c41a739) — the reviewed content is unchanged; prior review carries over. Rebase by the program orchestrator, git-only, no code edits.

pablonyx:

Head SHA verified: `d4c41a739ce53fc56b2754aa22194093477b25ea` (branch `diag-slice2a-devenv-republish`)
Machine context: local macOS run, detached worktree at `/tmp/diag-verify`, sequential (no concurrent cargo/rustc). Checkout again showed the benign `Cargo.lock` version auto-sync diff (0.4.12→0.4.13) from the reused worktree — not a code change.

**Command 1**
```
cargo test -p proliferate --lib dev_env_republish
```
Literal output:
```
running 6 tests
test sidecar::tests::dev_env_republish::snippet_carries_all_four_lines_including_its_own_path ... ok
test sidecar::tests::dev_env_republish::write_replaces_atomically_with_owner_only_mode ... ok
test sidecar::tests::dev_env_republish::degraded_transitions_do_not_republish ... ok
test sidecar::tests::dev_env_republish::armed_shutdown_exits_the_loop ... ok
test sidecar::tests::dev_env_republish::ready_not_covered_by_the_boot_publish_is_published_immediately ... ok
test sidecar::tests::dev_env_republish::new_ready_boot_id_republishes_exactly_once ... ok

test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 566 filtered out; finished in 0.00s
```
**Anomaly flagged (not adjusted):** cross-check expectation was 7 tests; literal local result is 6 passed, 0 failed. Verified directly against source: `apps/desktop/src-tauri/src/sidecar/tests.rs` lines 242-447 (`mod dev_env_republish`) contains exactly 6 `#[test]`/`#[tokio::test]` functions at this head — no 7th test exists to run. This is a genuine count mismatch against the supplied cross-check, not a flake or a miscount on my end; reporting as-is per instructions rather than conforming to the expected number.

**Command 2**
```
cargo test -p proliferate --lib sidecar
```
Literal output:
```
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 557 filtered out; finished in 0.12s
```
This broader filter captures the same 6 `dev_env_republish` tests plus 9 other sidecar-module tests, all passing.

No merge — verification only.

PR #1962, opened 2026-08-15
fix(anyharness): gate streaming ACP notifications out of the frame tee by default

pablonyx:
## What

Slice 1 of the diagnostics-pipeline hardening program (frozen spec: diagnostics-hardening, Slice 1 — ACP tee source gate).

The per-frame ACP tee (`frame_tee.rs`) emits one `anyharness.acp.recv`/`.send` DEBUG record per JSON-RPC wire line. The diagnostics admission layer admits any self-named `anyharness.*` target at any level, so demoting the level does nothing — the gate must live at the emit site. During the 2026-08-15 wf2pablo outage, `session/update` firehoses (one record per token-ish chunk × 5 concurrent agents) drove producer pressure drops and ultimately the collector child's death.

## Change

- `STREAMING_METHODS = ["session/update"]` frames are dropped from the tee by default, in both directions, after the existing `tracing::enabled!` fast path and header parse (ordering per spec).
- `ANYHARNESS_ACP_TEE_FULL` (read once via `OnceLock`) restores the full per-frame tee for deep wire debugging.
- Turn boundaries remain covered by `anyharness.turn.started/finished/failed`.
- The env flag is threaded into `log_frame_gated` as a bool so tests exercise both sides of the gate without racing on process-global env state (the spec's "inject a bool for testability" seam).

The spec's optional per-turn `frames_recv_by_method` aggregate was dropped as non-trivial (it reaches into `turn/finish.rs` + `prompt_diagnostics`), per the "same PR only if trivial, else drop" clause.

## Tests

`cargo test -p anyharness-lib frame_tee` — 7 passed (3 new: skip-by-default both directions, non-streaming request/response still logged, full-tee restore).

Negative control: with the two gate checks removed, `streaming_notifications_are_skipped_by_default_in_both_directions` fails (`assertion left == right failed` at the empty-capture assert); restored, it passes.

Full `cargo test -p anyharness-lib`: 2020 passed, 2 failed — both pre-existing/environmental and untouched by this diff:
- `the_sweep_removes_only_abandoned_and_old_scratch_roots`: the test helper formats a `touch -t` stamp in UTC but `touch` reads it as local time; on a UTC-7 machine the "aged" dir gets a future mtime and the sweep (correctly) spares it. Fails on main in any timezone behind UTC.
- `stop_and_await_kills_the_real_process_group_...`: passes in isolation; parallel-suite flake.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/live/sessions/driver/frame_tee.rs
+++ b/anyharness/crates/anyharness-lib/src/live/sessions/driver/frame_tee.rs
@@ -2,7 +2,9 @@
 //!
 //! Every line the client writes to, or reads from, the agent process is
 //! recorded once at DEBUG under `anyharness.acp.send` / `anyharness.acp.recv`,
-//! notifications included. The frame is logged as the raw wire line — never a
+//! except the streaming notification methods in [`STREAMING_METHODS`], which
+//! are dropped by default and restored by `ANYHARNESS_ACP_TEE_FULL`. The
+//! frame is logged as the raw wire line — never a
 //! re-serialization of a parsed value — capped so one giant payload cannot
 //! push the diagnostics pipe over its per-record budget.
 
@@ -39,11 +41,36 @@ struct FrameHeader<'a> {
     id: Option<&'a RawValue>,
 }
 
+/// Streaming notification methods excluded from the tee by default: they are
+/// one record per token-ish chunk, and turn boundaries are already covered by
+/// `anyharness.turn.started/finished/failed`. Set `ANYHARNESS_ACP_TEE_FULL`
+/// to restore the full per-frame tee for deep wire debugging.
+const STREAMING_METHODS: &[&str] = &["session/update"];
+
+fn full_tee_enabled() -> bool {
+    static FULL: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
+    *FULL.get_or_init(|| std::env::var_os("ANYHARNESS_ACP_TEE_FULL").is_some())
+}
+
+/// True when the frame is a streaming notification the tee drops by default.
+/// The diagnostics admission layer admits every self-named `anyharness.*`
+/// target at any level, so the gate has to live here at the emit site —
+/// demoting the record's level would change nothing.
+fn suppressed_streaming_frame(header: &FrameHeader<'_>, full_tee: bool) -> bool {
+    matches!(header.method, Some(m) if STREAMING_METHODS.contains(&m)) && !full_tee
+}
+
 pub(in crate::live::sessions) fn log_frame(
     session_id: &str,
     direction: FrameDirection,
     line: &str,
 ) {
+    log_frame_gated(session_id, direction, line, full_tee_enabled());
+}
+
+/// `full_tee` is threaded as a parameter so tests can exercise both sides of
+/// the gate without racing on process-global env state.
+fn log_frame_gated(session_id: &str, direction: FrameDirection, line: &str, full_tee: bool) {
     // Parsing and capping happen only when a subscriber wants the record:
     // with no diagnostics producer installed and the default console filter,
     // this is the whole cost of the tee.
@@ -53,6 +80,9 @@ pub(in crate::live::sessions) fn log_frame(
                 return;
             }
             let header = parse_header(line);
+            if suppressed_streaming_frame(&header, full_tee) {
+                return;
+            }
             let (payload, truncated) = cap_frame(line);
             tracing::debug!(
                 target: "anyharness.acp.send",
@@ -69,6 +99,9 @@ pub(in crate::live::sessions) fn log_frame(
                 return;
             }
             let header = parse_header(line);
+            if suppressed_streaming_frame(&header, full_tee) {
+                return;
+            }
             let (payload, truncated) = cap_frame(line);
             tracing::debug!(
                 target: "anyharness.acp.recv",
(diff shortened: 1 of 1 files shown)


pablonyx (commented):
Reviewed against the frozen spec (`/tmp/diagnostics-hardening-spec.md`, Slice 1) and the local worktree at `diag-slice1-acp-tee-gate` (HEAD `7df458461`, merge-base `89349486c` = current `origin/main`).

**Diff scope** — `git diff --stat origin/main HEAD` (worktree `main` was locally stale, false-positive noise there; against real `origin/main` the diff is exactly `frame_tee.rs`, +108/-2). No unrelated changes. ✅

**Gate correctness** (`anyharness/crates/anyharness-lib/src/live/sessions/driver/frame_tee.rs`):
- `log_frame_gated:79-85` (Send) and `:98-104` (Recv) — both arms gate identically, only `STREAMING_METHODS = ["session/update"]` (line 48). ✅ matches spec exactly.
- Ordering preserved per spec: `tracing::enabled!` fast path (79/98) → `parse_header` (82/101) → `suppressed_streaming_frame` gate (83/102) → cap+emit. ✅
- `suppressed_streaming_frame:59-61`: `matches!(header.method, Some(m) if ...)` is `false` for `None` method, so responses (no `method`) and non-streaming requests (`session/prompt`) always pass through untouched — confirmed by `non_streaming_requests_and_responses_are_still_logged` (line 197). ✅ satisfies review item (d).
- Rationale comment (56-58) correctly restates the spec's "admission layer admits any `anyharness.*` target" reasoning for why the gate has to live at the emit site, not the level.

**Env-var escape / test isolation** (item e): `full_tee_enabled` (50-53) reads `ANYHARNESS_ACP_TEE_FULL` once via `OnceLock`, but `log_frame_gated` takes `full_tee: bool` as a parameter and only the public `log_frame` (63-69) calls `full_tee_enabled()`. All three new tests call `log_frame_gated` directly with an explicit bool — none touch `std::env`, so there's no `set_var`-across-parallel-tests race and no OnceLock-caching-order dependency. Clean solution, matches the spec's explicit "inject a bool for testability" seam. Minor/low: `full_tee_enabled()` itself (the OnceLock read) has no direct test — only exercised transitively through untested `log_frame`. Given it's a one-line env read, not worth a dedicated test.

**Negative control** (item f): PR body states removing the two gate checks fails `streaming_notifications_are_skipped_by_default_in_both_directions` at the empty-capture assert. Verified this logically by reading the code path: dropping the `if suppressed_streaming_frame(...) { return; }` lines would let `tracing::debug!` fire and the capturing subscriber would write a non-empty line, breaking the `assert_eq!(..., "")`. Consistent with the described negative control. No permanent revert-test committed, which is fine — that's a one-time demonstration, not a standing test.

**Optional per-turn aggregate**: correctly dropped per the spec's "only if trivial" clause; it isn't trivial (touches `turn/finish.rs` + `prompt_diagnostics`), and its absence doesn't regress anything — turn boundaries stay covered by the existing `anyharness.turn.started/finished/failed` records, untouched by this diff.

**Nit (non-blocking)**: `FrameDirection::Recv` doc comment (line 27-29) still reads "every `session/update`-class notification the transport carries" without noting most of those are now dropped by default from the *log*. It's describing the wire direction (still accurate) not the tee's logging behavior, so not wrong, just slightly stale in spirit — could mention the default-drop for a reader skimming just the enum docs.

**Labels** (item a): exactly `release:fix` + `area:anyharness`. Correct and sufficient — this PR touches only `anyharness-lib`'s driver module, no other area applies.

**CI** — `gh pr view 1962 --json statusCheckRollup`:
- `CI / Cargo check & test` → **FAILURE**. Pulled the job log (`gh run view 31911154575 --job 95076281916 --log`): the single failure is `live::sessions::actor::tests::workspace_stop::stop_and_await_kills_the_real_process_group_including_a_git_grandchild` (`assertion left == right failed: exactly the git-named process counts toward the git census`, `workspace_stop.rs:419`) — 2019 passed / 1 failed / 3 ignored. This is exactly the process-group-kill flake the implementer flagged as pre-existing/environmental, and it's in an unrelated module (`workspace_stop.rs`, nothing to do with `frame_tee.rs`) untouched by this diff. The scratch-root TZ bug the implementer also mentioned didn't reproduce in this run (consistent with a UTC CI runner not hitting the "behind UTC" condition described).
- All other checks (`Repo shape checks`, `CodeQL` ×3, `Intent Tests` ×2, `PR Metadata` ×5, `Self-Host Smoke` detect + skipped compose-smoke, `Build SDK`, `Desktop/Web frontend build`, `Mobile typecheck`, `Shared frontend packages`, `Login first-load budget`, `Workflow definition lifecycle`, `Candidate build handoff`, `Terraform validate`, `CI/CD config checks`, `Greptile Review`, `Vercel` ×2) → **SUCCESS**. `Shared frontend packages` is SUCCESS here (not the known main-flake case).

**Would-be verdict: Approve.** The gate is correctly placed at the emit site in both directions, scoped to exactly the one streaming method, preserves the fast-path/parse ordering, doesn't disturb non-streaming/response logging, and the test seam avoids env-var flakiness by construction. The one failing CI job is a confirmed pre-existing, unrelated flake. No unrelated diff, labels correct. (Not using `--approve` — this account authored the PR.)

pablonyx:

Head SHA verified: `7df458461697ef32083e8b0afeb9c73b5ca8dbf6` (branch `diag-slice1-acp-tee-gate`)
Machine context: local macOS run, detached worktree at `/tmp/diag-verify`, sequential (no concurrent cargo/rustc).

Command:
```
cargo test -p anyharness-lib frame_tee
```
Literal output:
```
running 7 tests
test live::sessions::driver::frame_tee::tests::a_frame_within_the_cap_is_logged_whole ... ok
test live::sessions::driver::frame_tee::tests::an_unparseable_frame_still_yields_an_empty_header_instead_of_dropping_the_record ... ok
test live::sessions::driver::frame_tee::tests::an_oversized_frame_is_capped_on_a_char_boundary_and_marked_truncated ... ok
test live::sessions::driver::frame_tee::tests::headers_are_extracted_for_requests_notifications_and_responses ... ok
test live::sessions::driver::frame_tee::tests::streaming_notifications_are_skipped_by_default_in_both_directions ... ok
test live::sessions::driver::frame_tee::tests::full_tee_restores_streaming_notifications ... ok
test live::sessions::driver::frame_tee::tests::non_streaming_requests_and_responses_are_still_logged ... ok

test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 2018 filtered out; finished in 0.00s
```

No merge — verification only.

PR #1963, opened 2026-08-15
fix(anyharness): dev diagnostics producer re-learns the collector from the env file

pablonyx:
Slice 2b of the diagnostics-pipeline hardening program (root cause: 2026-08-15 wf2pablo 10.5h silent producer outage).

**Stacked on #1961 (`diag-slice2a-devenv-republish`) — retargets to `main` once 2a lands.** Only the last commit is this slice.

## Problem
`DevEnvDiagnosticsBootstrap` deliberately carried no bridge and pinned `DEV_COLLECTOR_GENERATION = 1`. After a collector restart, boot-id-mismatched receipts latched that generation unusable, permanently ending delivery for an externally launched dev runtime with zero warning.

## Change
- `dev_env_activation` also reads the snippet's own path (`PROLIFERATE_DIAGNOSTICS_DEV_ENV_PATH`, added by slice 2a) and carries it as `env_path: Option<PathBuf>`. Absent path (old app build's 3-line snippet) → exactly today's frozen behavior.
- New `producer/dev_refresh.rs` task (unix + debug builds): polls the file (~5s stat), and when the parsed snippet carries a **different collector boot id**, builds a `CollectorGenerationHandle` with a locally owned monotonic generation counter and calls `inner.replace_generation` — reusing exactly the fd bridge's `GenerationReady` primitives, so the swap also supersedes the unavailable latch.
- `parse_dev_env_file` enforces the same bounds as env activation (`MAX_ID_BYTES`, capability 256B) and rejects missing keys/oversized files; the capability is never logged, only the endpoint.
- New `ProducerInner::current_collector_boot_id()` / `is_terminal()` accessors; stale doc comments on the bootstrap and generation constant updated.

## Tests
`cargo test -p proliferate-diagnostics-client` — **198 passed, 0 failed** (fmt + clippy clean). New coverage in `tests_dev_refresh.rs`:
- parse accepts the published 4-line snippet; rejects missing keys, out-of-bound values, oversized/absent files
- rewrite with a new boot id → `replace_generation` observed as generation 2 with records delivered to the new endpoint (real loopback collector fixture)
- rewrite with the same boot id → no swap
- garbage file → no swap, unavailable latch preserved; a later valid rewrite supersedes the latch

**Negative control:** with the `replace_generation` call reverted, `rewrite_with_new_boot_id_reattaches_as_generation_two` and `garbage_preserves_latch_and_valid_rewrite_supersedes_it` fail (`test result: FAILED. 3 passed; 2 failed`); restored, all green.

No spec contradictions: slice 2a's landed contract (4-line `KEY=VALUE` snippet, `DEV_ENV_PATH_ENV` constant) matches the spec sketch.

## CI notes
- First push failed `Repo shape checks`: my additions put `producer/mod.rs` at 644 lines (max 600). Fixed in the second commit by relocating the dev-refresh glue into the cfg-gated `dev_refresh.rs` and moving `mark_bridge_lost` next to its only caller in `bridge_runtime.rs` (no behavior change; all 198 tests + fmt + clippy green, `check_max_lines.py` passes locally).
- `Shared frontend packages / Test product client` fails identically on the base PR #1961 and on `main` (pre-existing `ChatInputControlRow.layout.test.tsx` break) — unrelated to this Rust-only diff.

### DIFF
--- a/anyharness/crates/proliferate-diagnostics-client/src/bridge/activation.rs
+++ b/anyharness/crates/proliferate-diagnostics-client/src/bridge/activation.rs
@@ -112,6 +112,12 @@ pub const DEV_CAPABILITY_ENV: &str = "PROLIFERATE_DIAGNOSTICS_BRIDGE_TOKEN";
 /// delivery after the first batch.
 #[cfg(debug_assertions)]
 pub const DEV_COLLECTOR_BOOT_ID_ENV: &str = "PROLIFERATE_DIAGNOSTICS_BRIDGE_COLLECTOR_BOOT_ID";
+/// Absolute path of the env snippet itself. The host rewrites that file on
+/// every new collector generation, so a producer that keeps re-reading it can
+/// re-attach after a collector restart instead of staying pinned to a dead
+/// generation.
+#[cfg(debug_assertions)]
+pub const DEV_ENV_PATH_ENV: &str = "PROLIFERATE_DIAGNOSTICS_DEV_ENV_PATH";
 
 /// Fixed: with no bridge there is nothing to advance the generation.
 #[cfg(debug_assertions)]
--- a/anyharness/crates/proliferate-diagnostics-client/src/lib.rs
+++ b/anyharness/crates/proliferate-diagnostics-client/src/lib.rs
@@ -22,6 +22,7 @@ pub use bridge::activation::{
 #[cfg(debug_assertions)]
 pub use bridge::activation::{
     DevEnvDiagnosticsBootstrap, DEV_CAPABILITY_ENV, DEV_COLLECTOR_BOOT_ID_ENV, DEV_ENDPOINT_ENV,
+    DEV_ENV_PATH_ENV,
 };
 pub use fallback::{
     parse_fallback_record_line, FallbackReason, FallbackRecordV1, FALLBACK_SCHEMA,
--- a/apps/desktop/src-tauri/src/sidecar/diagnostics.rs
+++ b/apps/desktop/src-tauri/src/sidecar/diagnostics.rs
@@ -246,15 +246,15 @@ impl super::SidecarProcess {
 /// . <path>` in the `make dev` recipe. Only the path is logged.
 ///
 /// Dev builds only, and only while the collector is Ready — the snippet is
-/// rewritten on every app boot, so a collector restart is picked up the next
-/// time the dev runtime is started.
+/// written at boot and re-published by `spawn_dev_diagnostics_env_republisher`
+/// whenever a collector restart brings up a new generation, so an already
+/// running dev runtime can re-read it and re-attach.
 #[cfg(all(debug_assertions, unix))]
-pub(super) fn publish_dev_diagnostics_env(supervisor: &Arc<DiagnosticsCollectorSupervisor>) {
-    use std::io::Write;
-    use std::os::unix::fs::OpenOptionsExt;
-
+pub(super) fn publish_dev_diagnostics_env(
+    supervisor: &Arc<DiagnosticsCollectorSupervisor>,
+) -> Option<String> {
     let Ok(path) = crate::app_config::dev_diagnostics_env_path() else {
-        return;
+        return None;
     };
     let collector = match supervisor.dev_collector_env() {
         Ok(collector) => collector,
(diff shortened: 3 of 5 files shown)


pablonyx (commented):
Reviewed as the dedicated agent for slice 2b (stacked on #1961/2a). Read the frozen spec §Slice 2b, `gh pr diff 1963`, and the local worktree.

## Verdict (would-be)
**Request changes** — one real, CI-blocking defect (file-length lint), plus one non-blocking founder-visible design gap. The core logic (generation reuse, latch supersession, bounds, secret hygiene, cfg gating, scope) is sound and matches the frozen spec closely.

## CI status (literal)
- `Repo shape checks` — **FAILURE** — genuine, not base-related.
- `Shared frontend packages` — **FAILURE** — pre-existing/unrelated (see below).
- `Cargo check & test`, `Build SDK`, `Desktop frontend build`, `Web frontend build`, `Candidate build handoff`, `Terraform validate (server/infra)`, `CI/CD config checks`, `Validate PR title and labels` (×4), `intent-tests (provisional)`, `intent-billing (provisional)`, `Login first-load budget (phase-6)`, `Mobile typecheck`, `Workflow definition lifecycle (tier-2)`, `Greptile Review`, `Vercel`, `Vercel Preview Comments` — all **SUCCESS**.

### `Repo shape checks` — real, not a base artifact
`scripts/check_max_lines.py` fails with:
```
anyharness/crates/proliferate-diagnostics-client/src/producer/mod.rs: 644 (max 600)
```
Checked out `origin/diag-slice2a-devenv-republish` (base tip): the same file is **598 lines** (under the 600 cap), and 2a's own PR (#1961) has this check **SUCCESS**. This PR's diff (the `dev_env_path` threading through `install()`, plus the two new `#[cfg(all(unix, debug_assertions))]` accessors `current_collector_boot_id`/`is_terminal`) is what pushes it to 644, and `producer/mod.rs` is **not** in `scripts/max_lines_allowlist.txt`. This is a genuine regression introduced by 2b's own diff, unrelated to the non-`main` base — needs a fix (trim/relocate the new accessors, e.g. into `status.rs`, or land an allowlist entry) before this is mergeable.

### `Shared frontend packages` — pre-existing, unrelated to this diff
Failure is `ChatInputControlRow.layout.test.tsx > composer control row layout at the right-rail clamp floor > keeps every control non-overlapping at MAIN_PANE_MIN_WIDTH` (`floor-icon-mode: expected 7 to be greater than or equal to 8`). Checked #1961: **same job, same FAILURE**, on a diff that never touches this PR's Rust-only files. This is inherited from the base branch (looks like the unrelated PRO-104 composer-geometry flake) — not caused by 2b.

## Labels
Exactly `release:fix` + `area:anyharness` — correct (single release:*, and anyharness is the only touched area).

## Findings

**F1 (blocking, CI-red): `producer/mod.rs` exceeds the 600-line repo-shape cap.** See above. Fix required before merge.

**F2 (medium, non-blocking, spec-conformant but worth flagging): mtime-only refresh gating can permanently miss a rewrite.**
`producer/dev_refresh.rs:33-40`:
```rust
let mtime = meta.modified().ok();
if mtime == seen_mtime { continue; }
seen_mtime = mtime;
let Some(parsed) = parse_dev_env_file(&path) else { continue };
```
`seen_mtime` is updated *before* `parse_dev_env_file` succeeds. If the Desktop host's non-atomic write (`apps/desktop/src-tauri/src/sidecar/diagnostics.rs:311-327`, `truncate(true)` + `write_all`, no rename) is read mid-write, or two restarts land within the filesystem's mtime granularity, the poll can lock onto a mtime whose content never successfully parses — and because `seen_mtime` already advanced, a later poll that would otherwise re-read the *same* timestamp's now-stable content never retries. Worst case this reproduces exactly the outage this slice fixes, until a further, timestamp-distinguishable rewrite or a runtime restart. Greptile's automated review independently flagged the same root cause (comment on `dev_refresh.rs:47`, P2). This mirrors the frozen spec's own pseudocode almost verbatim (spec lines 111-137 use the identical ordering), so it isn't an implementer deviation — but it's real and worth Pablo's awareness given the reliability goal of the slice. Test coverage doesn't currently exercise same-second double-rewrite, so this gap is untested as well as unmitigated.

**F3 (low, nit): blocking I/O inside the async poll task.** `producer/dev_refresh.rs` calls `parse_dev_env_file`, which does `std::fs::read_to_string` (sync) from inside an async fn that otherwise uses `tokio::fs::metadata`. Harmless in practice (tiny file, 5s cadence) but inconsistent; `tokio::fs::read_to_string` would be more idiomatic.

## Verified correct
- Same-boot-id republish (host re-publishing for our own boot) → no swap (`tests_dev_refresh.rs::rewrite_with_same_boot_id_does_not_swap`), via `current_collector_boot_id()` comparison in `dev_refresh.rs`.
- Boot-id-mismatch unavailable latch is genuinely superseded: `worker.rs:382-390` latches `CollectorAvailability::Unavailable { generation: generation_identity.generation }` (same number, not bumped) on a receipt boot-id mismatch; `replace_generation`'s locally-owned monotonic `dev_generation` counter (`producer/mod.rs:411-436`, `generation.generation <= current` check) is always strictly newer, so a valid rewrite always supersedes the latch — confirmed by `garbage_preserves_latch_and_valid_rewrite_supersedes_it`.
- Garbage/missing file → latch/state preserved, no regression toward `Disabled`.
- `parse_dev_env_file` bounds match `dev_env_activation`'s exactly: `MAX_ID_BYTES` (endpoint, boot id), `DEV_CAPABILITY_MAX_BYTES` (capability), plus a new whole-file `DEV_ENV_FILE_MAX_BYTES` (8192) cap not present before.
- Secret hygiene: capability is never logged anywhere in the new code; the re-attach `tracing::info!` only carries `endpoint`.
- cfg gating is correctly unix+debug-only at both the module (`mod dev_refresh` in `producer/mod.rs:20-21`) and the spawn site (`producer/mod.rs:285-291`); the non-unix debug arm explicitly discards `env_path` (`let _ = &dev.env_path;`) rather than widening scope.
- Poll task lifecycle: checks `inner.is_terminal()` right after waking (`producer/mod.rs:453-459` accessor), so it exits within one poll interval of the guard dropping — bounded, not an indefinite leak. It's fire-and-forget (JoinHandle discarded, not joined by the shutdown path), consistent with how the rest of the module treats best-effort background tasks.
- Doc comments on `DevEnvDiagnosticsBootstrap` (`activation.rs:93-107`) and `publish_dev_diagnostics_env` (`sidecar/diagnostics.rs:240-252`) are updated away from the old "carries no bridge"/"write-once, dead end" wording to describe the refresh loop.
- Negative control present and documented in the PR body (revert `replace_generation` call → 2 of 5 new tests fail).
- Diff is scoped to exactly 4 files, all within `proliferate-diagnostics-client`, no unrelated changes.

I did not run any cargo/rustc build locally per machine constraints; relying on CI's green `Cargo check & test` for the 198-passed claim in the PR body.

pablonyx:
Rebased onto the rebased #1961 base (stack moved to current main to cure the base-drift "Shared frontend packages" red — the failing test was exactly the one #1959 fixed on main, and this branch's old merge-base 89349486c predated that fix). Patch identity proven: all 4 commits `=` in `git range-diff`, and the 2b-vs-2a tree diff is byte-identical pre/post rebase (stale base-merge 992ef306d correctly dropped). check_max_lines passes. Reviewed content unchanged; prior review carries over. Rebase by the program orchestrator, git-only, no code edits.

pablonyx:

Head SHA verified: `45717ad168e4c3b89dbe92e2ae339649c208898c` (branch `diag-slice2b-dev-producer-refresh`)
Machine context: local macOS run, detached worktree at `/tmp/diag-verify`, sequential (no concurrent cargo/rustc).

**Part A — headline frontend test**

Command (after a one-time `pnpm install --frozen-lockfile` and a required `pnpm build` in `apps/packages/product-client` to materialize `dist/*` — the test otherwise fails to resolve `@proliferate/product-client/internal/...` imports from `apps/desktop`):
```
pnpm vitest run src/components/workspace/chat/input/ChatInputControlRow.layout.test.tsx
```
Literal output:
```
 ✓ src/components/workspace/chat/input/ChatInputControlRow.layout.test.tsx (1 test) 2975ms
   ✓ composer control row layout at the right-rail clamp floor > keeps every control non-overlapping at MAIN_PANE_MIN_WIDTH  1484ms

 Test Files  1 passed (1)
      Tests  1 passed (1)
```

**Root-cause verdict:** the earlier lane failure was BASE-DRIFT, not real breakage. The old merge-base (89349486c) predated the #1959 fix, and the failing assertion ("floor-icon-mode: expected 7 to be greater than or equal to 8") is the exact test #1959 fixed. Rebasing this stack onto current main cured it deterministically — confirmed above with a clean local pass at this head.

**Part B — Rust**

Command:
```
cargo test -p proliferate-diagnostics-client --lib
```
Literal output:
```
test result: ok. 199 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.21s
```
Matches cross-check exactly (~199 passed).

No merge — verification only.

pablonyx:
Superseded by #1995 (same branch, same reviewed content, rebased onto main with the one module-declaration conflict resolved). This PR was closed by an errant force-push during its retarget pass — nothing from this slice reached main outside the replacement. Review history for the slice stays here.

PR #1964, opened 2026-08-15
fix(diagnostics): collector death certificates + loud delivery-end warning (slice 3)

pablonyx:
## What

Slice 3 of the diagnostics-pipeline hardening program (frozen spec: diagnostics-hardening 2026-08-15; root cause: the wf2pablo dev runtime's producer counted loss silently for 10.5h after a collector death).

**Desktop supervisor (death certificates).** The monitor observed the child's `ExitStatus` and threw it away. Now `fail_generation` carries the observed status (refining it at reap time if the child died between the trigger and the take), and every `desktop.collector.restart` lifecycle record carries typed arguments: `trigger` (`child_exited` | `child_inspection_failed` | `health_unavailable` | latched classification), `exit_code` (i64, if any), `signal` (i64, unix, if any), and `restart_count`. The terminal `desktop.collector.stop` `child_exited` path attaches the same certificate. `begin_lifecycle` gains an arguments-accepting variant; launch-time restart bursts (no dead child) pass no certificate and are unchanged.

**Diagnostics client (loud delivery end).** The two latches into `CollectorAvailability::Unavailable` — `mark_generation_unavailable` (bridge notice) and the boot-id-mismatched receipt in the worker — now emit a one-shot-per-generation `WARN` on `anyharness.diagnostics.delivery` ("diagnostics delivery ended: collector generation gone; records count as loss until re-attach"), guarded by a per-generation flag in `AdmissionState`. `replace_generation` logs the canonical matching info line (call sites must not double-log; slice 2b consumes this). The WARN self-ingestion noted in the spec is accepted: one-shot, and the pipe is dead at that point.

## Tests

- `cargo test -p proliferate-diagnostics-client --lib` → **197 passed, 0 failed**, including new `tests_delivery_end`: one-shot flag semantics, exactly-one-WARN per generation death (captured via a counting tracing layer), the canonical re-attach info line, and a real-worker boot-mismatched-receipt latch warning exactly once.
- `cargo test -p proliferate --lib diagnostics_collector` → **205 passed, 0 failed**, including: extended `killed_collector_restarts_...` asserting the restart records carry `trigger=child_exited`, `signal=SIGKILL`, no `exit_code`, `restart_count=1`; new stop-path test for an already-exited child; unit test distinguishing clean exit (`exit_code=0`, no `signal`) from signal death and from an uninspected (`health_unavailable`) certificate, run through real producer admission.

**Negative controls** (each fix reverted temporarily, test observed failing, fix restored):
- warn suppressed in `mark_generation_unavailable` → `generation_unavailable_warns_exactly_once_per_generation` FAILED
- warn suppressed at the worker boot-mismatch latch → `boot_mismatched_receipt_latch_warns_exactly_once` FAILED
- certificate dropped from `fail_generation` → `killed_collector_restarts_with_new_generation_capability_and_query` FAILED

Pre-existing, unrelated: `cargo clippy -p proliferate` fails on `diagnostics/support_snapshot/assembly/accounting.rs` (`overly_complex_bool_expr`, from #1818); three tests are flaky under parallel runs and pass in isolation — `bridge::activation::close_if_open...`, `child_bridge::...status_timeout...`, and `process_kill::tests::census_counts_git_executables_distinctly_from_the_total` (anyharness-lib; failed once in CI, re-run solo 5/5 green — the census counts machine-wide processes and is perturbed by concurrent tests spawning git). All untouched by this diff.

## Repo shape fix (review round 1)

producer/mod.rs (641) and supervisor/lifecycle.rs (674) breached the 600-line cap. The slice's client-side additions (plus the two generation-transition methods that own the logging) now live in `producer/delivery_log.rs`, deliberately avoiding the mod.rs regions slice 2b (#1963) reshuffles; the desktop certificate type and the supervisor's certificate-assembly helpers live in `supervisor/death_certificate.rs`. `scripts/check_max_lines.py` passes locally; all three negative controls were re-run after the move and still fail with the fix reverted.

## Spec notes

One observation raised (not rescoped): the transport `Authentication|Rejected|Protocol` failure latch in `worker.rs` also flips a generation to `Unavailable` but is not named by the spec, so it does not warn in this slice.

### DIFF
--- a/anyharness/crates/proliferate-diagnostics-client/src/producer/delivery_log.rs
+++ b/anyharness/crates/proliferate-diagnostics-client/src/producer/delivery_log.rs
@@ -1,0 +1,107 @@
+//! Collector-generation availability transitions and their loud logging. The
+//! 2026-08-15 outage counted loss for 10.5h without one line in the runtime's
+//! own log; every latch into `Unavailable` now says so out loud, and every
+//! re-attach answers it.
+
+use super::AdmissionState;
+#[cfg(unix)]
+use super::{CollectorAvailability, ProducerInner};
+#[cfg(unix)]
+use crate::fallback::FallbackReason;
+#[cfg(unix)]
+use std::sync::Arc;
+
+#[cfg(unix)]
+impl ProducerInner {
+    pub(crate) fn replace_generation(
+        &self,
+        generation: crate::bridge::activation::CollectorGenerationHandle,
+    ) {
+        let mut state = self.state.lock().unwrap_or_else(|error| error.into_inner());
+        let current = match &state.collector {
+            CollectorAvailability::Ready(current)
+            | CollectorAvailability::Cooldown {
+                generation: current,
+                ..
+            } => current.generation,
+            CollectorAvailability::Unavailable { generation } => *generation,
+        };
+        if generation.generation <= current {
+            return;
+        }
+        for record in &mut state.queue {
+            record.fallback_reason = Some(FallbackReason::GenerationChanged);
+        }
+        if !state.in_flight.is_empty() {
+            state.delivery_fence_eligible = false;
+        }
+        let generation_number = generation.generation;
+        state.collector = CollectorAvailability::Ready(Arc::new(generation));
+        drop(state);
+        info_delivery_reattached(generation_number);
+        self.notify.notify_one();
+    }
+
+    pub(crate) fn mark_generation_unavailable(&self, generation: u64) {
+        let mut state = self.state.lock().unwrap_or_else(|error| error.into_inner());
+        let current = match &state.collector {
+            CollectorAvailability::Ready(current)
+            | CollectorAvailability::Cooldown {
+                generation: current,
+                ..
+            } => current.generation,
+            CollectorAvailability::Unavailable { generation } => *generation,
+        };
+        if generation <= current {
+            return;
+        }
+        for record in &mut state.queue {
+            record.fallback_reason = Some(FallbackReason::GenerationChanged);
+        }
+        state.collector = CollectorAvailability::Unavailable { generation };
+        if !state.in_flight.is_empty() {
+            state.delivery_fence_eligible = false;
+        }
+        let warn = state.note_delivery_ended(generation);
+        drop(state);
+        if warn {
+            warn_delivery_ended(generation);
+        }
+        self.notify.notify_one();
+    }
+}
+
+impl AdmissionState {
+    /// One-shot per generation: several sites can observe the same dead
+    /// generation (a bridge notice, a boot-mismatched receipt, retries), but
+    /// the loud delivery-end warning fires once.
+    pub(crate) fn note_delivery_ended(&mut self, generation: u64) -> bool {
+        if self.delivery_end_warned_generation == Some(generation) {
+            return false;
+        }
+        self.delivery_end_warned_generation = Some(generation);
+        true
+    }
+}
+
+/// WARN is re-ingested by the tracing layer, which is acceptable: the caller
+/// guards with `note_delivery_ended` and the pipe being described is already
+/// dead. Callers emit outside the admission-state lock.
+pub(crate) fn warn_delivery_ended(generation: u64) {
+    tracing::warn!(
+        target: "anyharness.diagnostics.delivery",
+        generation,
+        "diagnostics delivery ended: collector generation gone; records count as loss until re-attach"
+    );
+}
+
+/// The canonical re-attach line answering `warn_delivery_ended`; call sites
+/// must not log their own copy.
+#[cfg(unix)]
+pub(crate) fn info_delivery_reattached(generation: u64) {
+    tracing::info!(
+        target: "anyharness.diagnostics.delivery",
+        generation,
+        "diagnostics delivery re-attached: collector generation ready"
+    );
+}
--- a/anyharness/crates/proliferate-diagnostics-client/src/producer/mod.rs
+++ b/anyharness/crates/proliferate-diagnostics-client/src/producer/mod.rs
@@ -25,6 +25,9 @@ pub(crate) mod transport;
 mod worker;
 
 #[cfg(all(test, unix))]
+#[path = "tests_delivery_end.rs"]
+mod tests_delivery_end;
+#[cfg(all(test, unix))]
 #[path = "tests_fallback_deadline.rs"]
 mod tests_fallback_deadline;
 #[cfg(test)]
@@ -135,6 +138,33 @@ pub(crate) struct AdmissionState {
     pending_loss_total: u64,
     pending_loss_range: PendingLossRange,
     open_loss_snapshot: Option<LossSnapshot>,
+    delivery_end_warned_generation: Option<u64>,
+}
+
+impl AdmissionState {
+    /// One-shot per generation: several sites can observe the same dead
+    /// generation (a bridge notice, a boot-mismatched receipt, retries), but
+    /// the loud delivery-end warning fires once.
+    pub(crate) fn note_delivery_ended(&mut self, generation: u64) -> bool {
+        if self.delivery_end_warned_generation == Some(generation) {
+            return false;
+        }
+        self.delivery_end_warned_generation = Some(generation);
+        true
+    }
+}
+
+/// The 2026-08-15 outage counted loss for 10.5h without one line in the
+/// runtime's own log; every latch into `Unavailable` now says so out loud.
+/// WARN is re-ingested by the tracing layer, which is acceptable: the caller
+/// guards with `note_delivery_ended` and the pipe being described is already
+/// dead.
+pub(crate) fn warn_delivery_ended(generation: u64) {
+    tracing::warn!(
+        target: "anyharness.diagnostics.delivery",
+        generation,
+        "diagnostics delivery ended: collector generation gone; records count as loss until re-attach"
+    );
 }
 
 pub(crate) enum CollectorAvailability {
@@ -251,6 +281,7 @@ pub(crate) fn install(
             pending_loss_total: 0,
             pending_loss_range: PendingLossRange::Empty,
             open_loss_snapshot: None,
+            delivery_end_warned_generation: None,
         }),
         fallback: Arc::new(fallback_runtime::FallbackController::new(fallback)),
         notify: tokio::sync::Notify::new(),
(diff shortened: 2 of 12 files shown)


pablonyx (commented):
## Review: Slice 3 — collector death certificates + loud delivery-end warning

Read the frozen spec (Slice 3 section), the full diff (`gh pr diff 1964`, 733 lines / 8 files), and the touched files in the worktree. No build was run (implementer already `cargo clean`'d; CI is the build authority here).

### Verdict (would-be — this account authored the PR, `--approve` is rejected)
**Request changes.** The design and logic are sound and match the frozen spec closely, but CI is genuinely red on two counts that are not addressed in the PR body's "known/accepted" list, plus one scope question worth a founder ruling before merge.

### Blocking: CI failures

**1. `Repo shape checks` — FAILURE, real and PR-caused.**
`scripts/check_max_lines.py` (max 600 lines/file, `scripts/max_lines_allowlist.txt` for exceptions):
```
anyharness/crates/proliferate-diagnostics-client/src/producer/mod.rs: 641 (max 600)
apps/desktop/src-tauri/src/diagnostics_collector/supervisor/lifecycle.rs: 674 (max 600)
```
Both files were already close to the ceiling on `main` (598 and 571 lines respectively) before this PR's additions (delivery-end warn machinery in `mod.rs`; `CollectorDeathCertificate` + argument plumbing in `lifecycle.rs`) pushed them over, and neither has an allowlist entry. This is a hard gate, not advisory — needs a fix before merge. Given the PR already extracts new test code into its own file (`tests_delivery_end.rs`), the idiomatic fix is probably to extract `CollectorDeathCertificate`/`operational_argument` (lifecycle.rs:376–432) into a sibling module, and/or split the `AdmissionState` delivery-end helpers out of `producer/mod.rs`, rather than reaching for the allowlist.

**2. `Cargo check & test` — FAILURE, looks unrelated but is undisclosed.**
```
process_kill::tests::census_counts_git_executables_distinctly_from_the_total FAILED
anyharness-lib/src/process_kill_tests.rs:302
assertion `left == right` failed: only the git-named executable counts as a git kill
  left: 0, right: 1
```
`process_kill_tests.rs` is not touched anywhere in this diff (grepped — 0 hits), and the test itself spawns/races reaping a compiled fake-"git" binary in a process group census, which reads as a timing-sensitive environment flake rather than a real regression. But it is **not** one of the two flakes the PR body discloses (`bridge::activation::close_if_open...`, `child_bridge::...status_timeout...`). Recommend re-running CI to confirm flake-vs-regression and, if flaky, adding it to the disclosed list — either way, CI is currently red and blocks merge regardless of root cause.

Everything else on the rollup is green; `Shared frontend packages` (the known pre-#1959-main false-failure) is SUCCESS here, not an issue.

### Scope question: the third `Unavailable` latch (worker.rs transport failures)

Confirmed by reading `finish_ingest` (`anyharness/crates/proliferate-diagnostics-client/src/producer/worker.rs:448-461`): the `TransportError::Authentication | Rejected | Protocol` arm sets `state.collector = CollectorAvailability::Unavailable { .. }` but never calls `state.note_delivery_ended(..)` / `warn_delivery_ended(..)`, unlike the boot-mismatch arm 60 lines above it (worker.rs:389-394) and `mark_generation_unavailable` (mod.rs). This matches exactly what the PR body discloses — it's a stated, not hidden, scope decision.

**Is it dangerous?** Yes, structurally. This path can end delivery with zero `anyharness.log` evidence under a scenario where the collector process never dies (so the child-exited certificate path never fires) and the boot id doesn't change (so the receipt-mismatch warn never fires) — e.g. a capability/token rejection, a content-level batch rejection, or a malformed/incompatible response from a still-alive collector on the same boot id. That's the same "loss counts silently" failure class the 10.5h outage exemplified, reached through a third door the spec didn't name.

**Recommendation: fold into this PR rather than defer.** The fix is mechanically identical to the pattern already used twice in the same file — at worker.rs:448-461, latch with `state.note_delivery_ended(current.generation)` before/with the `Unavailable` transition, then `warn_delivery_ended(..)` after dropping the lock, same as the boot-mismatch arm. It reuses already-built primitives, touches a file already being edited by this PR, and closes a symmetric blind spot in the exact failure class this slice exists to fix — deferring mainly defers real risk for no engineering savings. That said, the frozen spec named only two warn sites, so this is a scope call that belongs to Pablo, not something to expand unilaterally — flagging for his ruling.

### Verified correct (no findings)

- **Exit-status threading is lossless.** `exit_code` only set from `status.code()`, `signal` only from `ExitStatusExt::signal()` under `cfg(unix)` (lifecycle.rs:404-421) — mutually exclusive per unix `ExitStatus` semantics, so a certificate never carries both. `Err(_)`/inspection-failure paths correctly pass `None` through `fail_generation`'s `child_inspection_failed` classification (lifecycle.rs:514-518). The `Ok(None)` branch inside `fail_generation` (health_unavailable path, lifecycle.rs:436-444) force-kills via `terminate_and_reap()` without capturing that kill's exit status — but this matches the spec's literal per-trigger argument list and is directly tested (`death_certificate_distinguishes_clean_exit_from_signal`'s "uninspected" case expects no exit_code/signal for `health_unavailable`); noting as a minor completeness observation only, not a defect.
- **Schema limits.** `MAX_ARGUMENTS = 32` (`proliferate-diagnostics-protocol/src/v1/limits.rs:13`); a certificate emits at most 3 arguments (trigger, restart_count, and one of exit_code/signal), nowhere near the cap. Argument names/values are short and well under `MAX_STRING_BYTES`/`MAX_NAME_BYTES`.
- **Admission routing.** `begin_lifecycle_with_arguments` routes both `"desktop.collector.restart"` and `"desktop.collector.stop"` through the real `begin_admitted_lifecycle` path (both are listed in `PR3_LIFECYCLE_NAMES`, producer.rs:33-41) — mirrors the pre-existing defensive-branch pattern in `begin_lifecycle_with_correlation` exactly, so the "unadmitted name" fallback branch is dead code for these call sites, consistent with existing style.
- **`restart_count` semantics.** `accept_restart_budget_at` increments `inner.restart_count` before each retry loop iteration reads it for the certificate (supervisor.rs:237-247, lifecycle.rs:258-274), so every restart attempt's Begin-phase record (arguments are attached at `begin_admitted_lifecycle` time, not just terminal) carries its own correct attempt count.
- **One-shot guard is per-generation, not per-process.** `AdmissionState.delivery_end_warned_generation: Option<u64>` is keyed by generation number; a second generation's death re-warns, explicitly tested in `note_delivery_ended_is_one_shot_per_generation` and `generation_unavailable_warns_exactly_once_per_generation` (both assert a fresh warn on a new generation).
- **No infinite loop through the diagnostics layer.** The state lock is dropped before `warn_delivery_ended`/`tracing::warn!` fires in both `mark_generation_unavailable` (mod.rs:475-482) and `finish_ingest` (worker.rs:476-478), so no self-deadlock; `boot_mismatched_receipt_latch_warns_exactly_once` exercises this through the real `worker::run` + a scripted collector, not just a mock, and asserts exactly one warn.
- **No double-logging collision with slice 2b.** Grepped the worktree and `origin/main` for `dev_generation_refresh`/`DEV_ENV_PATH_ENV` — neither exists yet, confirming 2a/2b haven't merged, so there's no live collision today. Forward-looking note: this PR's canonical re-attach info line lives in `replace_generation` (mod.rs, `"diagnostics delivery re-attached: collector generation ready"`, same target); 2b's spec snippet shows an `info!` call at its own call site in `dev_generation_refresh` — whoever lands 2b needs to drop that call-site log per this PR's own comment ("call sites must not log their own copy"), or both will fire for one event. Not a defect in this PR.
- **No unrelated changes.** Diff is exactly 8 files / 733 lines, all within the two crates the spec names (`proliferate-diagnostics-client` producer/worker + tests, desktop `diagnostics_collector` supervisor/producer/lifecycle + tests). No CI/workflow/doc changes.
- **Pre-existing clippy claim checks out.** `support_snapshot/assembly/accounting.rs` doesn't appear anywhere in the diff.
- **Labels correct.** Exactly one `release:*` (`release:fix`) plus `area:desktop` and `area:anyharness` — matches the two touched trees, no additional area applies.

### Negative controls
PR body claims three reverted-fix-fails-test controls (warn suppressed at `mark_generation_unavailable`, warn suppressed at the boot-mismatch latch, certificate dropped from `fail_generation`). Couldn't independently re-run given the no-build constraint; the corresponding tests (`generation_unavailable_warns_exactly_once_per_generation`, `boot_mismatched_receipt_latch_warns_exactly_once`, `killed_collector_restarts_with_new_generation_capability_and_query`) read as genuinely capable of catching each regression (they assert on real emitted arguments/warn counts, not mocks), so the claims are plausible on inspection.

### Summary
Logic, typed-argument plumbing, one-shot-per-generation semantics, and loop-safety all check out against the spec. Two CI checks are red (`Repo shape checks` for a real line-count overage in two files this PR grew past 600; `Cargo check & test` for an undisclosed, probably-unrelated `process_kill` flake) and need resolution before this is mergeable. The worker.rs transport-failure blind spot is a real, disclosed gap worth folding into this PR given how cheap the fix is — recommend Pablo rule on it explicitly rather than silently expanding scope.

pablonyx:

Head SHA verified: `5126651c0672cb458c1a303986f840900dd7a458` (branch `diag-slice3-death-certificates`)
Machine context: local macOS run, detached worktree at `/tmp/diag-verify`, sequential (no concurrent cargo/rustc). Note: checkout showed a benign `Cargo.lock` diff (workspace version 0.4.12→0.4.13) auto-synced by cargo from a prior build in the reused worktree — not a code change.

**Command 1**
```
cargo test -p proliferate-diagnostics-client --lib
```
First attempt hit the documented flake: `fatal runtime error: IO Safety violation: owned file descriptor already closed, aborting` in `bridge::activation::tests::descriptor_exists_reflects_open_and_closed_state` (SIGABRT). This is the known "owned_bridge fd-reuse on old heads" flake (this branch predates slice-2b). Solo rerun with `--test-threads=1`:
```
test result: ok. 197 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 8.84s
```
Matches cross-check (197+; counts differ from slice-2b's 199 as expected — this branch predates that fix, not a bug).

**Command 2**
```
cargo test -p proliferate --lib diagnostics_collector
```
First attempt failed 15 tests with `no real Cargo-built proliferate-diagnostics-collector binary was found under /private/tmp/diag-verify/target/debug` — a missing precondition, not a regression: these integration tests spawn the real collector binary, which hadn't been built yet in the reused worktree. Ran `cargo build -p proliferate-diagnostics-collector`, then reran:
```
test result: ok. 205 passed; 0 failed; 2 ignored; 0 measured; 361 filtered out; finished in 5.23s
```
Matches cross-check exactly (~205 passed, 2 ignored).

No merge — verification only.

PR #1966, opened 2026-08-15
build(cargo): slim dev-profile debug info (line-tables-only, no deps debug)

pablonyx:
## What

Adds two profile settings to the workspace-root Cargo.toml:

```
[profile.dev]
debug = "line-tables-only"

[profile.dev.package."*"]
debug = false
```

Workspace crates (anyharness/crates/*, apps/desktop/src-tauri, apps/desktop/src-tauri-debug) keep file and line backtraces for panics and logs but drop full variable-level DWARF debug info. Third-party dependencies get no debug info at all, since we never step into their source in a debugger.

## Why

Full debug info in the dev profile is the main driver of oversized rlibs, link memory, and disk usage on local builds. anyharness-lib alone produces a debug rlib of roughly 500M today. This machine class is an 18-core/24GB laptop that has already OOM-crashed and swap-died from concurrent Rust builds, and every additional gigabyte spent on unused DWARF sections directly shrinks the headroom before that happens. Cutting debug info to line-tables-only for workspace code and to nothing for dependencies reduces rlib size, the memory the linker needs to hold intermediate objects, and the disk footprint of target/debug, with no loss of the file:line backtraces engineers actually rely on.

## Correctness of the wildcard override

The `[profile.dev.package."*"]` override could in principle be read as applying to every crate, including workspace members, which would be wrong here since we want workspace crates to keep line-tables-only rather than losing debug info entirely. Checked against the Cargo book's Overrides section (https://doc.rust-lang.org/cargo/reference/profiles.html#overrides), which states explicitly: "To override the settings for all dependencies (but not any workspace member), use the `\"*\"` package name." So the wildcard here only ever touches third-party dependencies, and the `[profile.dev]` table's `line-tables-only` setting is what continues to apply to workspace members. No scoping change was needed.

Also checked the repo for any other place that pins `profile.dev` or similar (grepped all *.toml and any .cargo directories); there is none, so this is a clean addition with nothing to reconcile.

## Validation

No Rust build was run for this change, per the current one-build-at-a-time machine constraint. Validated with `cargo metadata --format-version 1`, which parses the manifest and workspace graph without compiling anything, and succeeded (762 packages, 12 workspace members resolved). Also confirmed there is no local Cargo.lock drift caused by this change; a version drift that showed up in Cargo.lock during that metadata run was pre-existing on main (release version bump not yet reflected in the lockfile) and unrelated, so it was reverted before committing to keep this PR to the one-line profile change.

## Note

The first CI run against this branch, and the first local build anyone does after merging, will be cache-cold: rustc will need to recompile every crate under the new profile settings since the debug-info shape is part of the fingerprint. Expect that one build to take the usual full-workspace time; subsequent incremental builds should be smaller and faster to link because of the trimmed debug sections.

## Verified (review round 2, 2026-08-15)

Measured on a clean build of `proliferate-runtime-update-protocol` (main vs this branch): crate rlib -23.4%, object code -31.5%, `__debug_info` -87.4%, `__debug_loc` eliminated, `__debug_line` -3.9% (preserved), and a live `RUST_BACKTRACE=full` panic still resolves exact workspace file:line. Adversarial semantics pass found no defect. Details in the PR comments.

**Merge timing:** this change invalidates every sccache cache key, so the first post-merge build is a full cold recompile. Pre-warm the cache with one build on main after merging, or stagger fleet worktree rebuilds, to avoid concurrent full-graph builds on the 24GB dev machine.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,18 @@ edition = "2021"
 license = "AGPL-3.0"
 repository = "https://github.com/proliferate-ai/proliferate"
 
+[profile.dev]
+# Keep file/line backtraces for workspace crates without paying for full
+# variable-level debug info (DWARF). Cuts rlib size, link memory, and disk.
+debug = "line-tables-only"
+
+[profile.dev.package."*"]
+# Applies to all non-workspace-member dependencies only (Cargo book,
+# "Overrides": the `"*"` package name overrides settings "for all
+# dependencies (but not any workspace member)"). Dependencies get no debug
+# info at all, since we never step into their source.
+debug = false
+
 [workspace.dependencies]
 # HTTP / async
 axum = { version = "0.8", features = ["ws"] }


pablonyx:
## Review: build/dev-profile-slim-debug (bb2f36ab6)

Verdict: **would-approve**, contingent on the still-pending Rust CI jobs (see #7) coming back green — nothing has failed, but the jobs that actually exercise this change haven't reported yet.

### 1. [Info] Diff scope confirmed
`gh pr diff 1966` shows exactly one file, `Cargo.toml`, +12/-0. No `Cargo.lock` drift. Matches the PR description.

### 2. [Info] Profile placement is correct
`[profile.*]` tables only make sense in the workspace root and are ignored (with a warning) anywhere else. Grepped every member `Cargo.toml` (`anyharness/crates/*`, `apps/desktop/src-tauri`, `apps/desktop/src-tauri-debug`) for `^[profile` — zero matches. No competing/duplicate profile declarations to reconcile.

### 3. [Info] Toolchain compatibility is fine
`debug = "line-tables-only"` requires Rust **1.71+** per the Cargo book's MSRV table. Repo has no `rust-toolchain.toml`/`rust-toolchain` pin anywhere and no `rust-version` field in any `Cargo.toml`. Every CI workflow (`ci.yml`, `server-ci.yml`, `release-*.yml`, etc.) resolves Rust via `dtolnay/rust-toolchain@stable`, which always installs current stable (local `rustc` here is 1.95.0). No compatibility risk introduced by this change.

### 4. [Info] Wildcard citation checked against source
Fetched https://doc.rust-lang.org/cargo/reference/profiles.html#overrides directly — it states verbatim: *"To override the settings for all dependencies (but not any workspace member), use the `"*"` package name."* The PR body's citation is accurate; `[profile.dev.package."*"]` will not touch workspace-member debug info, only third-party deps.

### 5. [Low] Trade-off worth a one-line doc note (not a blocker)
Real behavior change: dev-build `lldb`/`gdb` variable/argument inspection degrades under `line-tables-only` (file:line backtraces survive, locals don't), and dependency frames get **zero** debug info at all (not even line-tables) since `debug = false` on `"*"`. Grepped the whole repo for `lldb`, `DWARF`, `split-debuginfo`, `debug = true` — zero hits in `specs/`, `guides/`, or anywhere else, so nothing currently documents or depends on full dev DWARF. Given that, this is safe to ship as-is, but a one-liner in the dev-setup docs would save a future engineer a confused debugging session the first time they attach a debugger to a dev build and can't read a local variable.

### 6. [Info] CI interaction
No CI job asserts on binary/artifact size or debug-info shape (checked `.github/workflows/*.yml` for size/`du`/`target/debug` assertions — none exist). `Swatinem/rust-cache@v2` is in use; changing the profile changes rustc's fingerprint, so the first post-merge CI run (and first local build per engineer) recompiles from scratch — this matches the PR body's own "cache-cold" callout and is an expected one-time cost, not a functional risk.

### 7. [Medium — gating, informational] Literal CI status at review time
```
Analyze (python)                        pass
CI/CD config checks                     pass
Detect smoke-relevant changes           pass
Greptile Review                         pass
Login first-load budget (phase-6)       pass
Mobile typecheck                        pass
Repo shape checks                       pass
Terraform validate (server/infra)       pass
Validate PR title and labels            pass
Vercel / Vercel Preview Comments        pass
CodeQL                                  skipping
Production compose smoke                skipping
Cargo check & test                      pending
Analyze (rust)                          pending
Build SDK                               pending
Desktop frontend build                  pending
Web frontend build                      pending
Shared frontend packages                pending
Candidate build handoff                 pending
Workflow definition lifecycle (tier-2)  pending
intent-tests (provisional)              pending
intent-billing (provisional)            pending
```
Nothing has failed, but the checks that actually compile the workspace under the new profile settings — most importantly **Cargo check & test** and **Analyze (rust)** — are still pending. Recommend waiting for those to go green before merge; if they pass, this is a clean, well-scoped, well-justified change.

---
*Reviewed per hard machine limits: no `cargo`/rustc build was run to validate this — assessment is based on manifest inspection, repo-wide grep, and the Cargo book source. Not able to `gh pr review --approve` on this account (PR author).*

pablonyx:
## End-to-end review round 2 (adversarial + behavioral)

**Adversarial semantics review (fresh agent, tasked to refute): no defect found.** Verified against the Cargo book: the `"*"` override excludes workspace members even when they are dependencies of each other, so all workspace crates keep `line-tables-only`; macOS `split-debuginfo` stays `unpacked` because line-tables-only still counts as debug enabled, so no dSYM behavior change for the Tauri dev app; build scripts and proc-macros resolve to no debug info either way (same as today's build-override default); profile `test` inherits the change; zero tests in the repo assert on backtraces or debug info; no script or pipeline consumes dev DWARF (Sentry symbolication concerns release artifacts only; the diagnostics crates measure memory, not stacks); `line-tables-only` needs Rust 1.71+, repo runs current stable.

**Behavioral verification (real clean builds of `proliferate-runtime-update-protocol`, `-j4`, incremental off, main vs PR head bb2f36ab6):**
- crate rlib 977,160 B -> 748,208 B (**-23.4%**); crate object code **-31.5%**
- DWARF sections (summed over codegen units via `size -m`): `__debug_info` **-87.4%**, `__debug_loc` **eliminated**, `__debug_line` only **-3.9%** (file:line tables preserved)
- live `RUST_BACKTRACE=full` panic under the PR profile shows exact workspace `file:line` (`validation.rs:154:9`)
- `target/debug` total -11.1%; deps-only rlibs -9.0% (uneven per crate: `quote`/`unicode-ident` byte-identical, `zmij` -35%; aggregate as expected)
- sccache behaves as disclosed: new flags mint new cache keys, one-time cold misses, no errors

**Operational note for whoever merges (not a blocker):** the flag change invalidates every sccache cache key, so the first build after merge is a full cold recompile of the ~750-package graph. On the 24GB dev machine, do not let several fleet worktrees rebase and rebuild simultaneously right after merge; pre-warm the cache with one build on main, or stagger the fleet's first rebuild.

No code changes required; the diff stands as reviewed. Round-1 review: [would-approve](https://github.com/proliferate-ai/proliferate/pull/1966#issuecomment-5304686417). Not merged; merge stays with Pablo.

PR #1967, opened 2026-08-15
test(diagnostics): prove fd-bridge producer reattach across collector restart (Slice 5)

pablonyx:
# Slice 5 — prod fd-bridge reattach acceptance (test-only)

Closes the diagnostics-hardening program's Slice 5. **Test-only**: no production
code changed. Proves, in automated form, the self-heal the incident post-mortem
inferred only from code reading — a bundled/prod producer attached over the fd
bridge survives a collector restart instead of going dark.

### Half 1 — client crate (`proliferate-diagnostics-client`)

New `producer/tests_reattach.rs` drives the **real** `BridgeRuntime` over a
`UnixStream` pair with **real wire frames**, closing the gap left by
`tests_generation.rs` (which calls `replace_generation` in-process):

- Boots a producer on the bridge for generation 1 pointed at collector fixture A,
  emits a record through the real worker, and asserts fixture A receives it.
- Queues records behind a stalled in-flight request, then sends a **real**
  `ParentFrame::GenerationReady` for generation 2 — new `collector_boot_id`,
  fixture B's endpoint, and a fresh capability delivered as an `SCM_RIGHTS`
  descriptor (mirroring the supervisor's `protected_child_handoff`).
- Asserts fresh work now dispatches to fixture B; the survivors of the swap are
  **routed** (`generation_changed` for the queued records, `delivery_unknown`
  for the cancelled in-flight request) rather than silently dropped; the
  snapshot converges to the new generation/boot; and a stale frame
  (generation ≤ current) is ignored.

### Half 2 — desktop crate (`proliferate`)

New `killed_collector_sends_new_generation_ready_over_the_child_bridge` in
`supervisor_recovery_tests.rs` attaches a **real** `ChildDiagnosticsBridge` over a
socketpair to a **real** supervisor running the **real** built collector binary,
kills the collector child (`kill_current_process_for_test`), and asserts the
child endpoint receives a well-formed `GenerationReady` for the restarted
collector — a strictly newer generation, a fresh boot id, and a readable
capability descriptor. The existing
`killed_collector_restarts_with_new_generation_capability_and_query` proves the
supervisor *produces* a fresh handoff; this proves the frame actually *crosses
the bridge* to the child after a real restart. macOS-gated like `child_bridge`.

Both halves share the literal `ParentFrame` codec from
`proliferate-diagnostics-client::bridge::wire`, so together they compose the
full chain; the only link not exercised in-process — fd inheritance across
`exec` — has its own coverage in `activation_tests.rs`/`activation_platform.rs`.

## Level of proof

This is the **strongest automated approximation**, not a cross-process e2e: a
true e2e would need a new helper binary spawned through
`PreparedChildDiagnosticsLaunch` with real fd inheritance, which is
disproportionate for a test-only slice (new binary target, flake-prone process
choreography). A manual acceptance checklist for the real packaged-app flow is
below and in the module docs of `tests_reattach.rs`.

## Drive-by fix: parallel-suite fd-number races in `activation_tests.rs`

The new reattach test exposed a latent flake family: five tests asserted on (or
`close_if_open`'d) **raw fd numbers after releasing them**. In this
one-process parallel suite, a released number is reallocated by a neighbouring
test within microseconds, so those assertions probed — and in the
`close_if_open` no-op case, **closed** — other tests' descriptors. This is what
intermittently killed the reattach test's in-flight `SCM_RIGHTS` capability
(observed via `FIONREAD`/fd tracing: capability present at send, transferred
right dead on arrival). Fixed by re-homing the probed numbers into distinct
high ranges (`F_DUPFD` ≥ 512) that lowest-free allocation never reaches, and by
holding the reattach test's channel ends alive until convergence.
`owned_bridge_with_failed_cloexec_degrades_and_closes_both_authorities` was
observed failing spuriously ~10% per full-suite run before the fix; 30/30
clean full-suite runs after.

## Important finding (scope note)

The frozen slice text describes a "`GenerationChanged` **loss summary**" as the
not-silence signal. Reading the code, that is **not** how the pipeline signals
a reattach: generation-changed survivors are **routed to fallback**
(incrementing `fallback_routed`, writing `generation_changed` fallback lines)
and are **not** counted in `dropped_by_reason`. The `generation_changed` loss
counter has no production call site, so no loss-summary record with
`generation_changed` is ever emitted. The tests assert the **actual**
not-silence evidence: delivery to the new generation + `generation_changed`
fallback routing + the converged snapshot. Flagged for the reviewer; no code
was changed to force the spec's wording.

## Manual acceptance checklist (packaged app)

1. Launch the bundled Desktop app; let a runtime attach over the fd bridge.
   Confirm runtime records land in the collector (Honeycomb or the local records
   query) under the current `collector_boot_id`.
2. Find the collector child: `pgrep -f proliferate-diagnostics-collector`.
3. `kill -9 <pid>`.
4. Watch the supervisor restart it — `anyharness.log` shows
   `desktop.collector.restart`; supervisor state advances to a new
   `collector_boot_id` with `restart_count` incremented.
5. Confirm runtime records keep flowing under the **new** `collector_boot_id`
   (not silence), and the diagnostics fallback file records the in-flight/queued
   survivors with reason `generation_changed` / `delivery_unknown`. No further
   operator action should be required.

## Verification

- `cargo test -p proliferate-diagnostics-client` — full crate, 30/30 consecutive
  clean runs (194 tests each).
- `cargo test -p proliferate killed_collector_sends_new_generation_ready_over_the_child_bridge`
  — Half 2 against the real collector binary; 55/56 local runs clean (see note).
- Negative control: pointing the wire frame at a stale generation (1 instead
  of 2) fails the test at the convergence assertion
  (`the wire frame must replace the collector generation`); restoring it passes.

### Known rarity in Half 2 (flagged, not hidden)

One run out of 56 failed the final assertion — the freshly transferred
capability channel read **empty** — under heavy sibling build load, and did not
reproduce in 50 instrumented follow-up runs. Sender-side invariants look
airtight on reading (`protected_child_handoff` writes token+newline before the
frame is sent, holds the fd through `send_until`; tokens are 32 random bytes,
never empty). If this recurs it may be a **real handoff race**, not a test
bug — an empty capability would leave a real child unauthorized after restart.
Note: CI never exercises this test — `ci.yml` runs Linux-only and the release
workflow builds without testing, while this test is macOS-gated — so a
recurrence can only surface on a contributor's Mac. The shipped assertion
carries forensics (`FIONREAD` at receive, both token values, supervisor state)
so any local recurrence tells the whole story; paste the assertion message
into an issue if you hit it.


pablonyx (commented):
## Slice 5 review (test-only fd-bridge reattach acceptance)

**Would-be verdict: request changes** — one concrete, CI-confirmed blocker (repo-shape line-count gate), everything else is solid.

### 1. Blocking: this PR's own diff trips the repo-shape line-count gate
`anyharness/crates/proliferate-diagnostics-client/src/producer/mod.rs` is now **601 lines** (cap is 600, no allowlist entry exists for this file — checked `scripts/max_lines_allowlist.txt`). The 3 lines this PR adds (`producer/mod.rs:42-44`, registering `#[cfg(all(test, unix))] mod tests_reattach;`) tip it over. This is exactly what "Repo shape checks" is failing on:
```
Repo shape checks: anyharness/crates/proliferate-diagnostics-client/src/producer/mod.rs: 601 (max 600)
```
(from run 31915519181, step `python3 scripts/check_max_lines.py`, confirmed by local `wc -l` = 601). This is **not** the known pre-existing "Shared frontend packages" failure — that check actually passed on this PR (SUCCESS at 23:53:25Z). "Repo shape checks" is a distinct, genuine, PR-caused failure and needs a fix: either move the new `#[path]` mod declaration so it doesn't add to `mod.rs`'s count (e.g. register from a small child file, or trim a comment elsewhere in the file), or add a properly-justified `max_lines_allowlist.txt` entry following the existing convention (see `fallback/tests.rs` / `fallback/unix.rs` entries) — your call on which, but it can't merge red on a mechanical gate like this.

### 2. Proof level — matches the PR's own characterization
Confirmed: this is the **strongest automated approximation**, not true e2e, and the PR says so plainly (both in the top-level "Level of proof" section and the module doc on `tests_reattach.rs`). Concretely:
- Half 1 (`producer/tests_reattach.rs:271-434`) drives the real `BridgeRuntime` over a real `UnixStream` pair with real `ParentFrame::GenerationReady` wire frames and a real `SCM_RIGHTS`-transferred capability descriptor. It asserts all three things the spec asked for: delivery to the new generation (`new.records()` sequence 4, `tests_reattach.rs:391-396`), the abandoned queue routed rather than silently dropped (`fallback_reasons` == `[delivery_unknown, generation_changed, generation_changed]`, `tests_reattach.rs:380-390`), and a stale-frame no-op (`tests_reattach.rs:402-431`).
- Half 2 (`supervisor_recovery_tests.rs:458-612`) attaches a real `ChildDiagnosticsBridge` to a real supervisor running the actual built collector binary, does a real `kill -9`-equivalent (`kill_current_process_for_test`), and reads the post-restart `GenerationReady` off the real socketpair — genuinely the strongest thing short of spawning a whole second process via `PreparedChildDiagnosticsLaunch` with fd-inheritance across `exec`.
- The only gap (fd inheritance across `exec`) is correctly deferred to `activation_tests.rs`/`activation_platform.rs`, which is accurate — I checked those files carry that coverage.
- Determinism: both new tests use the existing bounded `wait_for`/`drained` polling helpers (`tests_support.rs:243-251`, 5ms poll up to 5s cap) plus explicit per-attempt/overall deadlines in Half 2's `GenerationReady` receive loop (`supervisor_recovery_tests.rs:557-578`) — no naked fixed sleeps driving the assertions themselves (the fixed 300ms `settle()` is a pre-existing absence-proof helper, not new here).

### 3. The "GenerationChanged loss summary" scope note — verified accurate, correctly surfaced
I checked this claim against the actual (non-test) source rather than taking the PR's word:
- `ProducerFailureClassification::GenerationChanged` (the `dropped_by_reason` counter that would feed a loss-summary record) is referenced **only** in `producer/tests_loss.rs:28`, never incremented at any production call site in `producer/*.rs`. Confirmed via grep across the whole crate.
- What actually happens on a generation swap, per `producer/worker.rs:476-482` (`generation_change_reason`): survivors are routed to fallback with `FallbackReason::GenerationChanged` (a different, already-wired enum) or `FallbackReason::DeliveryUnknown` if already dispatched, incrementing `fallback_routed` — not `dropped_by_reason`.
- So the spec's literal wording ("emits a GenerationChanged loss summary") doesn't match what the code does, and the PR is honest about that mismatch (PR body "Important finding" section) rather than bending the test to force a counter that's never hit in prod, or silently asserting something weaker without saying so. The tests assert the actual not-silence evidence instead. This is the right call — flag it to me/spec-owner for wording reconciliation in the frozen spec, don't change scope unilaterally. Good judgment here.

### 4. The 1-in-56 empty-capability flake — read on "real race vs. artifact"
This is Half 2 only (`supervisor_recovery_tests.rs:594-607`), gated `macos`, worker_threads=2. My read: **plausibly real, correctly characterized as unresolved, but the "CI recurrence" framing in the PR body overclaims** — worth a correction:
- **CI never runs this test at all.** `.github/workflows/ci.yml`'s Rust jobs (`cargo test --workspace`, line 253) run only on `runs-on: ubuntu-latest` — no macOS runner exists anywhere in `ci.yml`. The one macOS workflow (`release-runtime.yml`) only does `cargo build --release`, never `cargo test`. So this test — like its pre-existing macOS-gated sibling `killed_collector_restarts_with_new_generation_capability_and_query` and the `child_bridge` suite generally — only ever executes on a contributor's local Mac. The PR body's closing line, "so a CI recurrence tells the whole story," isn't accurate: there is no CI path where it can recur. This is a pre-existing platform gap (not introduced by this PR), but the PR's own wording implies CI-level ongoing detection that doesn't exist. Suggest correcting that sentence to "a local recurrence" or similar, and/or filing a lightweight follow-up ticket about the desktop crate having zero macOS-only test coverage in CI (separate from this slice's scope).
- On the substance of the race itself: I buy that it's plausibly real rather than pure test noise. The sender-side path (`protected_child_handoff`) writes token+newline before send and holds the fd through `send_until` per the PR's description, and the forensics captured (`FIONREAD` at receive, both token values, supervisor state) are exactly the right instrumentation to distinguish "bytes never arrived" from "bytes arrived wrong" if it recurs — and, importantly, that instrumentation lives entirely inside the test function (`supervisor_recovery_tests.rs:598-599`), not in any production file. No smuggled production forensics; confirmed the only 4 files touched are test files plus the `#[cfg(test)]`-gated mod registration in `producer/mod.rs`. Good discipline there.
- It's tracked visibly (PR body dedicated section + code comment at the assertion site), not buried — that satisfies the spec's "documented, not hidden" bar. Given it's 1/56 and didn't reproduce in 50 follow-up runs, treating it as "watch, don't block" for this slice is reasonable; I'd just want the CI-recurrence sentence fixed so nobody assumes a safety net that isn't there.

### 5. Housekeeping
- Labels: exactly one `release:*` (`release:maintenance`) + both applicable `area:*` (`area:desktop`, `area:anyharness`) — correct, matches the files touched (`apps/desktop/src-tauri/...`, `anyharness/crates/proliferate-diagnostics-client/...`).
- No production behavior changes anywhere in the diff — verified all 4 changed files are test files or test-only mod wiring.
- Drive-by fix (`activation_tests.rs` fd re-homing, `rehome_above` at line 47) is in-scope, minimal, and test-only: re-homes 5 probes to distinct floors (512/520/528/536/544) above `F_DUPFD`'s lowest-free reallocation range, fixing a real parallel-suite hazard (a released low fd getting picked up by a neighboring test) that was intermittently sabotaging both the pre-existing `owned_bridge_with_failed_cloexec_degrades_and_closes_both_authorities` (per PR body, ~10%/run before the fix) and the new reattach test's in-flight capability. Reasonable to bundle since the new test is what exposed it.
- No unrelated changes found.

### Summary
- **CI**: `Repo shape checks` = **FAILURE** (line-count gate, PR-caused, fixable) — real blocker. `Shared frontend packages` = SUCCESS (not the pre-existing failure mentioned as a non-issue). All other checks (Cargo check & test, CodeQL, SDK build, desktop/web frontend builds, PR metadata, Greptile, Vercel) = SUCCESS.
- **Labels**: `release:maintenance`, `area:desktop`, `area:anyharness` — correct and complete.
- **Proof level**: strongest automated approximation, honestly labeled as such, not true e2e (gap = fd-inheritance-across-exec, appropriately deferred to existing `activation_tests`/`activation_platform` coverage).
- **1/56 flake**: plausibly a real handoff race, forensics are appropriate and test-scoped (no prod code touched), tracked visibly — but fix the PR body's "CI recurrence" claim, since no CI lane runs this macOS-gated test.
- **Action needed before merge**: fix the `producer/mod.rs` 601-line repo-shape failure.

pablonyx:

Head SHA verified: `a02ec46a29f3d4ef1807479a8eef2701f701235e` (branch `diag-slice5-prod-reattach-acceptance`, test-only PR)
Machine context: local macOS run, detached worktree at `/tmp/diag-verify`, sequential (no concurrent cargo/rustc). Benign `Cargo.lock` version auto-sync diff again present from the reused worktree — not a code change.

**Command 1**
```
cargo test -p proliferate-diagnostics-client tests_reattach
```
Literal output:
```
running 1 test
test producer::tests_reattach::wire_generation_ready_reattaches_and_routes_the_abandoned_queue ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 193 filtered out; finished in 0.65s
```

**Command 2 (macOS-gated)**
```
cargo test -p proliferate killed_collector_sends_new_generation_ready_over_the_child_bridge
```
Literal output:
```
running 1 test
test diagnostics_collector::supervisor::recovery_tests::killed_collector_sends_new_generation_ready_over_the_child_bridge ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 566 filtered out; finished in 2.36s
```
Run in isolation, nothing else in parallel, no known 1/56 empty-capability forensic assertion trip on this run.

This macOS-gated test is the ONLY execution proof — CI never runs it (Linux-only ci.yml, and the release workflow builds without testing).

No merge — verification only.

pablonyx:
Resolved against post-slice-train main at 67cdc79dd (merge commit, one conflict: producer/mod.rs module list — kept this branch's no-redundant-#[path] convention and added main's new `tests_delivery_end` module under the same cfg; behavior-identical). Receipts: no duplicate fn names across producer/{mod,delivery_log,dev_refresh}.rs, `cargo check -p proliferate-diagnostics-client --all-targets` clean, max-lines clean. `supervisor_recovery_tests.rs` auto-merged — CI's desktop lane is the gate for that one. — orchestrator 554b8edc (takeover)

PR #1927, opened 2026-08-15
feat(product): parallelize workspace-switch pipeline behind selection abort (UX-latency R10)

pablonyx:
## What

Rung 10 of the UX Latency + Transitions ADR ladder. Takes the agent catalog off
the blocking workspace-switch bootstrap and warms it in the background from
selection entry, so it races the connection resolution and the blocking
session-directory fetch instead of running serially behind them. The single
catalog-readiness gate now lives at composer submit (session materialization),
named as a first-class helper.

Reference: UX Latency + Transitions ADR §4.6, Rung 10 (Q12).

Stacked on #1923 (`ux-latency/r9-staleness-abort`). PR base is that branch, not
`main`.

## Why

The workspace-switch pipeline ran connection resolution, the 8s-ceiling
session-directory fetch, and the agent catalog in a single serial chain. The
catalog is the one leg with no data dependency on the workspace, so it belongs
off the critical path.

## Dependency analysis (the real data dependencies)

Per the Q12 ruling, I analyzed what is genuinely independent before parallelizing:

- **session-directory fetch → connection resolution: a HARD dependency, in both
  the warm and cold cases.** The directory fetch consumes the resolved
  `workspaceConnection` (`anyharnessWorkspaceId`, `runtimeUrl`, `authToken`)
  produced by `resolveWorkspaceConnection`. Starting it before the connection
  resolves would fetch against a stale/wrong `anyharnessWorkspaceId` and risk
  painting wrong-workspace content, violating the truthfulness invariant. So the
  directory fetch is NOT reordered ahead of connection in either case. The
  warm/cold distinction lives inside `resolveSelectionConnection`
  (`ensureRuntimeReady` no-ops when warm); the parallelization sits upstream of
  that branch, so the ordering guarantee is warm/cold-independent.
- **agent catalog → connection: NO dependency.** The catalog is a global cloud
  document (`getCloudAgentCatalog`, keyed globally, 5-min staleTime). It needs
  nothing from the workspace connection, so it is the leg that parallelizes. It
  now starts fire-and-forget at selection entry (`run-workspace-selection.ts`),
  racing the whole connection→directory chain. Because it is global, a superseded
  selection cannot paint wrong-workspace content from it, so it needs no abort
  guard.
- **file-tree init → already deferred.** The synchronous `prepareFileWorkspace`
  (shell-tab truthfulness) stays inline; the heavy directory listing already runs
  after next paint via `scheduleDeferredFileTreePrefetch`, guarded by
[...]

PR #1968, opened 2026-08-16
feat(product): take transcript hydration off the workspace-switch critical path (R14)

pablonyx:
## Problem (measured live tonight, `renderer.flow` marks)

Workspace-switch critical path ran ~1.5s against a ~200ms floor:

- `intent_to_shell` 0–3ms (fine)
- `shell_to_data` 240–780ms — session-directory fetch, `session_list_cache` **miss on every revisit** (14 miss / 3 hit)
- `data_to_stable` 800–2300ms heavy vs 15–239ms light — dominated by `bootstrapWorkspace` awaiting a **full network transcript-history fetch** before emitting `content_stable`

Root causes (already diagnosed, not re-derived):
1. The remembered-session branch serially awaited `selectSessionWithShellIntentRollback` then `rehydrateSessionSlotFromHistory(..., {replace:true})` before the `finally` emitted `content_stable`. `SessionTranscriptPane` already self-hydrates, so the bootstrap-side await was largely redundant.
2. The `anyHarnessSessionsKey` react-query entry is seeded with `setQueryData`, loses its last observer on navigate-away, and is garbage-collected → every revisit is a cold miss.

## Changes

**1. Transcript hydration off the critical path.** `handleRememberedWorkspaceSessionBootstrap` now fires `rehydrateSessionSlotFromHistory` **fire-and-forget** (never awaited). Session selection stays on the critical path; only the history fetch moves off. `SessionTranscriptPane`'s existing self-hydration effect owns the transcript.

**2. In-flight dedupe.** `use-session-history-hydration` now shares one full-open (`replace`, no `afterSeq`/`beforeSeq`) hydration per session id across concurrent callers — the bootstrap kickoff and the pane effect collapse into a single fetch **and** a single apply (store writes + `session_open` flow marks fire once). Incremental append/prepend fetches are not deduped.

**3. Honest `content_stable`.** With hydration off the bootstrap path, emitting `content_stable` at bootstrap completion would lie — the transcript isn't on screen yet. The bootstrap now **defers** the mark (records the selected session id via `deferWorkspaceOpenContentStable`), and `SessionTranscriptPane` finishes the `workspace_open` flow (`finishDeferredWorkspaceOpenForSession`) when the selected session's transcript actually commits (non-null transcript + `useDeferredValue` caught up = `MessageList` rendering). Empty-workspace / error / stale paths still finish/abandon at bootstrap as before (no transcript to wait for). Marks stay additive/revertible; `budget_metric`/`budget_threshold_ms` unchanged.

> Honest-mark rationale: the founder's rule is never emit a stable mark before the user can see the transcript. A deferred flow that never commits (selection superseded before the pane mounts) simply ages out via `pruneStaleFlows` — no false milestone.

**4. Warm, truthful session directory.** Founder ruling: NOT stale-tolerant SWR — keep the list genuinely up to date and fetch immediately.
- (a) Bounded LRU keep-alive (`workspace-session-directory-keepalive.ts`, last N=5 visited queries) pins each session query with a disabled `QueryObserver` so an intra-run revisit is a warm **hit** instead of a GC'd miss; memory-bounded by evicting the least-recently-pinned observer past the bound.
- (b) Warm serve for first paint **plus an immediate background revalidation** (`void fetchAndSeed()`), so the list stays truthful without blocking first paint on the network.
- (c) Invalidate on replaced-session tombstone commit: `commitReplacedSessionTombstone` now notifies a listener that the bootstrap-cache hook uses to `invalidateQueries` the affected workspace, so a warm/pinned entry can never keep showing a retired session.
- `session_list_cache` hit/miss stays truthful: keep-alive means a revisit finds `dataUpdatedAt` → `hit`; invalidation → `stale`.

## R9 / R10 collision notes

Both are unmerged and cut from a different base; not absorbed here.
- **#1923 (R9 staleness-abort)** threads an `AbortSignal` through `loadWorkspaceSessionDirectory`/`loadWorkspaceSessions`. This PR edits the same `loadWorkspaceSessions` body (warm serve → background revalidate + pin). Both touch the cache-decision/warm branch; a rebase will need to route R9's signal through the new `fetchAndSeed` helper.
- **#1927 (R10 catalog parallelization)** adds `prefetchAgentCatalog` at `runWorkspaceSelection` entry. No overlap with these files beyond proximity in the selection workflow.

## Testing

Test list derived from `git diff --stat` vs base. Full `@proliferate/product-client` vitest suite run: **6992 passing**; the only 8 failures are pre-existing and unrelated — `src/hooks/access/tauri/updater-dev-mock.test.ts` (`window.localStorage` undefined; the node env is started without `--localstorage-file`), untouched by this PR and failing identically in isolation on the base tree.

New/updated unit tests (no widget mocks):
- `renderer-flow-timing.test.ts`: deferral finishes only when the deferred session commits; no-op for a non-deferred session; finishes at most once; drops a prior deferral when the settled session changes.
- `use-session-history-hydration.test.tsx`: two concurrent full hydrations → one fetch; incremental fetches are not deduped.
- `workspace-bootstrap-remembered-session.test.ts`: bootstrap resolves without awaiting hydration and returns `contentStableSessionId`; hydration is not awaited (record not patched synchronously).
- `use-workspace-bootstrap-cache.test.ts`: keep-alive pin + truthful `hit` after load; warm serve + background revalidation updates the cache; tombstone-commit invalidation → `stale`; staged replacement still never destructively reconciled on the warm serve.
- `workspace-session-directory-keepalive.test.ts`: pin survives GC; LRU bound; re-pin refreshes recency.

**Negative controls (mandatory) — performed for every behavioral test:** for each, I temporarily reverted the code change, confirmed the specific test failed, then restored the change and confirmed green. Done for: the dedupe wrapper, the `deferWorkspaceOpenContentStable` registration, the keep-alive pin call, the warm-serve background revalidation, the tombstone-commit invalidation notify, the LRU eviction bound, and the `contentStableSessionId` hand-off.

Typecheck clean (`tsc -p tsconfig.json --noEmit`). No Rust built (frontend-only change).

## Repo-shape / frontend-structure

To keep the three touched hooks under the ~400-line frontend-structure threshold, cohesive helpers were extracted into new modules: `session-history-hydration-dedupe.ts`, `workspace-open-flow-finish.ts`, `session-replacement-tombstone-listeners.ts`, plus `buildSessionHistoryFetchArgs` moved into `session-history-hydration-helpers.ts`. `use-session-history-hydration.ts` and `use-workspace-bootstrap-actions.ts` now land at exactly 400 lines (pass).

`session-replacement-tombstones.ts` was **already allowlisted at 408**; the R14 tombstone-commit listener wiring grew it to 414, so its existing allowlist entry was bumped 408 → 414 (marginal grow of a pre-existing entry; no new allowlist entries added). `python3 scripts/report_frontend_structure.py --strict` reports `LARGE_FRONTEND_FILE: 0`.

## Review fixes (round 1 — REQUEST_CHANGES)

**HIGH — content_stable could fire on the empty scaffold.** The deferred-finish effect in `SessionTranscriptPane.tsx` gated on `transcript` object presence, but cold workspace-open synchronously seeds `createEmptySessionRecord` with an empty-but-truthy `TranscriptState`, so `finishDeferredWorkspaceOpenForSession` fired ~0ms and `data_to_stable_ms` read ~0 for the exact case that used to measure 1.2–1.8s — content popped in AFTER the flow reported stable. Fix: gate on the directory entry's `transcriptHydrated` flag (read reactively), which both hydration call sites (this pane's self-hydration and the bootstrap kickoff) patch only after history is fetched and applied. A hydrated-empty session (new workspace, zero messages) still counts as stable; a scaffold-empty session does not.

**LOW — end-to-end coverage.** Added `SessionTranscriptPane.test.tsx` driving the real `createEmptySessionRecord` scaffold + `activateWorkspace` selection with a genuinely deferred `workspace_open` flow: asserts the flow does NOT finish while only the scaffold exists, DOES finish once hydration flips the flag (with `data_to_stable_ms` = 1500, not ~0), and a hydrated-empty variant still finishes. Negative control: reverting the gate to `transcript` presence makes both `toBeUndefined()` assertions fail (content_stable fires immediately, `data_to_stable_ms` = 25).

### DIFF
--- a/apps/packages/product-client/src/components/workspace/chat/surface/SessionTranscriptPane.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/surface/SessionTranscriptPane.test.tsx
@@ -1,0 +1,173 @@
+// @vitest-environment jsdom
+
+import type { ReactNode } from "react";
+import { act, cleanup, render } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { replaySessionHistory } from "#product/lib/domain/sessions/stream/stream-state";
+import { SessionTranscriptPane } from "#product/components/workspace/chat/surface/SessionTranscriptPane";
+import {
+  beginRendererFlow,
+  deferWorkspaceOpenContentStable,
+  markRendererFlowDataReady,
+  markRendererFlowShellCommitted,
+  resetRendererFlowsForTest,
+} from "#product/lib/infra/diagnostics/renderer-flow-timing";
+import {
+  resetRendererDiagnosticsSinkForTest,
+  setRendererDiagnosticsSink,
+  type RendererDiagnosticInput,
+} from "#product/lib/infra/diagnostics/renderer-diagnostics-port";
+import {
+  createEmptySessionRecord,
+  patchSessionRecord,
+  putSessionRecord,
+} from "#product/stores/sessions/session-records";
+import { useSessionDirectoryStore } from "#product/stores/sessions/session-directory-store";
+import { useSessionIntentStore } from "#product/stores/sessions/session-intent-store";
+import { useSessionSelectionStore } from "#product/stores/sessions/session-selection-store";
+import { useSessionTranscriptStore } from "#product/stores/sessions/session-transcript-store";
+
+// Heavy leaf UI is not under test — the deferred content_stable hand-off is.
+vi.mock("#product/components/workspace/chat/transcript/MessageList", () => ({
+  MessageList: () => <div data-testid="message-list" />,
+}));
+vi.mock("#product/components/workspace/chat/plans/ConnectedPlanHandoffDialog", () => ({
+  ConnectedPlanHandoffDialog: () => null,
+}));
+vi.mock("#product/components/workspace/chat/surface/TranscriptSwitchingPlaceholder", () => ({
+  TranscriptSwitchingPlaceholder: () => <div data-testid="placeholder" />,
+}));
+vi.mock("#product/components/diagnostics/DebugProfiler", () => ({
+  DebugProfiler: ({ children }: { children: ReactNode }) => <>{children}</>,
+}));
+vi.mock("#product/hooks/ui/debug/use-debug-render-count", () => ({
+  useDebugRenderCount: () => {},
+}));
+// The pane self-hydration path is exercised elsewhere; here we drive hydration
+// completion explicitly via the flag, so a no-op rehydrate keeps host wiring out.
+vi.mock("#product/hooks/sessions/lifecycle/use-session-history-hydration", () => ({
+  useSessionHistoryHydration: () => ({
+    rehydrateSessionSlotFromHistory: vi.fn().mockResolvedValue(true),
+  }),
+}));
+vi.mock("#product/hooks/plans/ui/use-plan-handoff-dialog-state", () => ({
+  usePlanHandoffDialogState: () => ({ plan: null, open: vi.fn(), close: vi.fn() }),
+}));
+vi.mock("#product/hooks/chat/workflows/use-transcript-session-navigation-actions", () => ({
+  useTranscriptSessionNavigationActions: () => ({
+    canOpenTranscriptSession: false,
+    openTranscriptSession: vi.fn(),
+  }),
+}));
+vi.mock("#product/hooks/workspaces/derived/use-workspace-creation-receipt", () => ({
+  useWorkspaceCreationReceiptKey: () => null,
+}));
+
+const WORKSPACE_ID = "workspace-1";
+const SESSION_ID = "session-1";
+
+let emitted: RendererDiagnosticInput[];
+let nowValue: number;
+
+beforeEach(() => {
+  emitted = [];
+  setRendererDiagnosticsSink({ emit: (input) => emitted.push(input) });
+  nowValue = 0;
+  vi.spyOn(performance, "now").mockImplementation(() => nowValue);
+  resetRendererFlowsForTest();
+  useSessionDirectoryStore.getState().clearEntries();
+  useSessionIntentStore.getState().clear();
+  useSessionTranscriptStore.getState().clearEntries();
+  useSessionSelectionStore.getState().deselectWorkspacePreservingSessions();
+});
+
+afterEach(() => {
+  cleanup();
+  vi.restoreAllMocks();
+  resetRendererFlowsForTest();
+  resetRendererDiagnosticsSinkForTest();
+  useSessionDirectoryStore.getState().clearEntries();
+  useSessionIntentStore.getState().clear();
+  useSessionTranscriptStore.getState().clearEntries();
+  useSessionSelectionStore.getState().deselectWorkspacePreservingSessions();
+});
+
+/**
+ * Simulates the real cold workspace-open: the bootstrap begins the
+ * workspace_open flow and, because transcript hydration is off its critical
+ * path, DEFERS content_stable to this session; selectSession synchronously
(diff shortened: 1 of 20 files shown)


pablonyx:
## Manual verification

Hands-on evidence for R14 (switch critical path). Not CI badges.

**Head SHA verified:** `19e16b111adfecf1ed99dfd5c1fef108594682d9` (equals `gh pr view --json headRefOid`; includes honesty fix 19e16b111). Synced by fetch + fast-forward; worktree clean (`git status --porcelain` empty).

### Scoped vitest, one file at a time (`--maxWorkers=2`), literal tallies

Test files derived from `git diff --stat` against base `ux-latency/r4-skeleton-retirement`:

- `src/components/workspace/chat/surface/SessionTranscriptPane.test.tsx` -> `Test Files  1 passed (1)` / `Tests  2 passed (2)`
- `src/hooks/access/anyharness/workspaces/use-workspace-bootstrap-cache.test.ts` -> `Test Files  1 passed (1)` / `Tests  10 passed (10)`
- `src/hooks/access/anyharness/workspaces/workspace-session-directory-keepalive.test.ts` -> `Test Files  1 passed (1)` / `Tests  3 passed (3)`
- `src/hooks/sessions/lifecycle/use-session-history-hydration.test.tsx` -> `Test Files  1 passed (1)` / `Tests  3 passed (3)`
- `src/hooks/workspaces/workflows/workspace-bootstrap-remembered-session.test.ts` -> `Test Files  1 passed (1)` / `Tests  3 passed (3)`
- `src/lib/infra/diagnostics/renderer-flow-timing-wiring.test.ts` -> `Test Files  1 passed (1)` / `Tests  7 passed (7)`
- `src/lib/infra/diagnostics/renderer-flow-timing.test.ts` -> `Test Files  1 passed (1)` / `Tests  17 passed (17)`

### Negative control on the honesty fix

Locally reverted the `content_stable` hydration gating (uncommitted): changed the deferred-finish effect in `SessionTranscriptPane.tsx` back from `!transcriptHydrated` to `!transcript` (scaffold object presence). Reran the pinning file:

```
 ❯ src/components/workspace/chat/surface/SessionTranscriptPane.test.tsx (2 tests | 2 failed)
   × does NOT finish on the empty scaffold, then finishes once hydration lands
     → expected { …(6) } to be undefined
 FAIL  src/components/workspace/chat/surface/SessionTranscriptPane.test.tsx:139:29
    139|     expect(contentStable()).toBeUndefined();
 Test Files  1 failed (1)
      Tests  2 failed (2)
```

The cold-open ~0ms honesty bug resurfaces exactly as expected: with the gate on object presence, `content_stable` fires against the empty-but-truthy `createEmptySessionRecord` scaffold (the emitted record includes `data_to_stable_ms` measured against the scaffold, not the real hydration wait). Restored the fix; reran clean: `Test Files  1 passed (1)` / `Tests  2 passed (2)`; `git status --porcelain` empty.

### Hand probe

Temporary jsdom harness (created, run, deleted; tree clean afterward) drove a real cold workspace-open switch flow: begin + shell_committed + data_ready + defer, real `createEmptySessionRecord` scaffold, `activateWorkspace` selection, render the real `SessionTranscriptPane`, then flip `transcriptHydrated`. Recorded renderer.flow.* mark sequence:

```
BEFORE hydration: ["renderer.flow.intent","renderer.flow.shell_committed","renderer.flow.data_ready"]
AFTER  hydration: ["renderer.flow.intent","renderer.flow.shell_committed","renderer.flow.data_ready","renderer.flow.content_stable"]
```

Proves `content_stable` cannot fire before hydration: it is absent while only the scaffold exists and appears only after `transcriptHydrated` flips true.

PR #1969, opened 2026-08-16
fix(product): streaming continuity across workspace switches (R15)

pablonyx:
Implements R15 of the UX Latency ladder: after sending a message, switching workspace, and switching back, the stream no longer visibly "restarts" and re-plays. Ruled behavior (Chat Scroll ADR Q15, restore-finalized / bottom-streaming): everything already received paints instantly as finalized at the bottom; only tokens arriving after remount animate.

**Base: stacked on `ux-latency/r14-switch-critical-path`** (the branch appeared on origin before push, as instructed; originally developed against `ux-latency/r4-skeleton-retirement` and rebased cleanly — no file overlap with R14).

## Phase 1 — verified mechanism

**(a) Deltas do NOT keep applying while backgrounded.** The hot-session policy scopes live streams to the selected workspace: `resolveHotSessionTargets` returns nothing for other workspaces (`lib/domain/sessions/hot-session-policy.ts:59-73` pre-change), and `reconcileHotSessions` closes every stream that drops out of the target set (`lib/workflows/sessions/hot-session-ingest-manager.ts:71-75` → `closeHotStream` → `closeSessionSlotStream`). The transcript slot itself is NOT evicted — it stays cached, frozen at its `transcript.lastSeq`, marked `cold` in the ingest store. Events are retained server-side: reconnects resume with `afterSeq = slot.transcript.lastSeq` (`hooks/sessions/lifecycle/session-stream-connection-open.ts:75,162-163`), and the ingest store tracks contiguity via `lastAppliedSeq`/`gapAfterSeq` (`stores/sessions/session-ingest-store.ts:106-123,153-173`). This differs from the founder's mental model (store kept ingesting) but does not change the ruled design — it just means R15 must supply the store continuity too.

**(b) The visible re-stream has two ingredients.** On return, `useHotWorkspaceReconcileAction` tail-hydrates with `afterSeq` and falls back to `rehydrateSessionSlotFromHistory({replace:true})` when the tail can't apply contiguously (`hooks/workspaces/workflows/use-hot-workspace-reconcile-action.ts:242-258` pre-change). Independently, the reveal animation replays: the reveal-progress cache (`hooks/chat/ui/assistant-reveal-progress.ts`) is only updated from a mounted row's effect, so it freezes when the pane unmounts; on remount `useAssistantRevealFrontier` sees `targetLength > stale cached visibleLength` and re-arms the reveal (`hooks/chat/ui/use-assistant-reveal-frontier.ts:38-49`), and `AssistantMessageContent` seeds from the stale prefix (`lib/domain/chat/transcript/assistant-message-reveal.ts` `initialVisibleContent`) and typewriters through the entire backlog at 360 chars/s. Verdict: the animation replay is the dominant visible symptom; the replace-fallback is the secondary data-level one.

## The three changes

1. **Store continuity while away** — `hot-session-policy.ts` adds a `background_running` target reason (lowest priority, inside the existing `MAX_HOT_SESSION_STREAMS = 12` cap, so selected-workspace targets always win): sessions outside the selected workspace stay hot only while they have in-flight work (view state `working`, or a pending/optimistic prompt so a send-then-switch race is covered). `use-hot-session-ingest.ts` feeds those candidates from the directory + transcript stores. Idle sessions never hold a stream, so slots don't grow unbounded. Existing manager behavior keeps a live stream open on selected→background demotion (no `closeHotStream`, verified by test), and `prepareSessionStreamConnection` already no-ops on an open stream, so the return-promotion doesn't reconnect.

2. **Trust the store on return (gap-repair, not unconditional replace)** — `use-hot-workspace-reconcile-action.ts` now skips history entirely when the slot is trustworthy: `transcriptHydrated && streamConnectionState === "open" && ingest freshness === "current" && gapAfterSeq === null` → zero fetches. On a real gap (closed stream, stale/gapped ingest state) it performs exactly one incremental `afterSeq` repair fetch; the full `{replace:true}` remains only as the fallback for a non-contiguous tail. Coordination with R14: R14 took the blocking `rehydrateSessionSlotFromHistory({replace:true})` off the workspace_open bootstrap critical path and added hydration dedupe; R15 defines what replaces it on the hot-reopen reconcile — trust-the-live-slot with gap-repair. No shared files between the two diffs.

3. **Render-as-finalized at mount** — `useAssistantRevealFrontier` records a mount-time reveal floor into the shared reveal-progress cache on the row's first render: content that already exists when the viewer first sees the item is marked fully revealed/complete, so nothing pre-mount can animate — only deltas arriving after mount hold the frontier and pace in. Items that first appear on a later render (a fresh live chunk in a new turn) start at floor zero, preserving the deliberate "never snap a large delivery" pacing that `AssistantMessage.test.tsx` locks in. This also retires the remount replay of recently-completed turns (`isRecentAssistantCompletion` path), which is the same defect class under Q15. Scroll ownership untouched — restore-finalized/bottom-streaming still owns initial position; this hands off at a single instant paint.

## Test evidence

Suites derived from `git diff --stat` vs base; full `pnpm --filter @proliferate/product-client test` run after `pnpm --filter @proliferate/product-client build`: **968 files / 7013 tests passed** on the rebased tree.

- `hot-session-policy.test.ts` (+4): background running kept hot; queued-prompt background kept hot; background running kept hot with no workspace selected; background dropped before selected-workspace targets under the cap.
- `hot-session-ingest-manager.test.ts` (+1): live stream stays open (no `closeSessionSlotStream`, no reconnect) when the target demotes selected→`background_running` — this is the "deltas keep applying while backgrounded" invariant at the manager level, using the real target/ingest-store machinery.
- `use-hot-workspace-reconcile-action.test.tsx` (new): return without gap performs **zero** history fetches (fetch-spy assertion); gap triggers **exactly one** `afterSeq` repair fetch and no replace; replace fires only when the tail can't apply; an open stream with a reported `gapAfterSeq` is not trusted. Uses the real session-record and ingest stores, no widget mocks.
- `use-assistant-reveal-frontier.test.tsx` (new): stale in-flight cache + grown content at mount → no reveal re-arm, floor raised to mount length; post-mount delta → reveal arms; item appearing after mount still paces from zero; completed turn remounted later doesn't re-arm.
- Existing `AssistantMessage.test.tsx` (14 tests) passes unchanged — mount pacing, remount-resume via `initialVisibleLength`, scroll-pause, and settle behavior are all preserved.

**Negative controls** (each production file reverted to base, its tests run, then restored):
- `hot-session-policy.ts` reverted → 3/8 policy tests fail (the three background-target tests).
- `use-hot-workspace-reconcile-action.ts` reverted → "performs zero history fetches" fails (unconditional hydrate observed).
- `use-assistant-reveal-frontier.ts` reverted → 3/4 frontier tests fail (mount replay re-arms).
The manager test addition documents pre-existing manager behavior that the new policy output now relies on, so it has no matching revert (the manager itself is unchanged).

## What now happens on return

While away, the running session's stream stays open and deltas keep landing in its store slot (freshness stays `current`). On return, the reconcile sees a live gap-free slot and fetches nothing; the pane mounts, the reveal floor finalizes everything already in the store in one instant paint at the bottom, and only tokens that arrive after that mount animate. If the connection did gap while away, one `afterSeq` repair fetch fills the tail (replace only if non-contiguous) — and either way nothing pre-mount animates.

## Adversarial review (dedicated agent, opus)

No blockers. Findings and dispositions:
- **Trust-path race on promotion (major-efficacy, accepted):** selecting a backgrounded running session flips its target `background_running → selected`, and the manager's `connectHotTarget` calls `markWarming` before its no-op reconnect resolves back to `current` (the stream stays open; `prepareSessionStreamConnection` early-returns). If the reconcile reads freshness inside that window it degrades to the single `afterSeq` repair — still zero replace, still no replay. The measurement step records `session.select.history_hydrate outcome:"skipped"` when the zero-fetch path fires, so the hit rate is observable at runtime (T4 debt covers the booted-app check).
- **Silent-stall window (accepted trade-off):** an "open + current + no gap" stream that silently stalled server-side would be trusted without the old unconditional `afterSeq` backstop. This is the ruled semantics (trust the store unless there is a real gap); stall detection belongs to the stream layer.
- **Foreground virtualizer remount mid-stream now paints finalized** instead of resuming the typewriter — consistent with the ADR's mount semantics (animation is arrival-vs-viewer), noted as intentional behavior change.
- **Render-phase write to the reveal cache** (mount floor) is idempotent and safe under StrictMode double-render; flagged as a latent hazard if the transcript ever renders under startTransition/Suspense.
- **Minor follow-ups:** the pending-prompt hot-candidate selector scans the transcript store per delta (O(sessions·log) with a `useShallow`-stable result — fine at current session counts); pending-prompt candidates require a directory entry, so a brand-new session with only an optimistic prompt relies on being the active session (it is, in the send-then-switch flow).

**Rebase note:** originally opened against a mid-work R14 snapshot (454069ecd), which failed the shared-frontend `renderer-flow-timing-wiring` guard in the CI merge (R14 later moved the workspace_open `finishRendererFlow` out of the bootstrap and updated that guard in its own tree). Rebased onto the green R14 head (ab52584aa) once #1968 settled — clean rebase, no conflicts, wiring guard + full suite (7014 tests) green, force-pushed with lease. No bootstrap-side `finishRendererFlow` was re-added.

## Second review round (independent adversarial review — APPROVE, no blockers)

Four follow-ups addressed in 402a38932:

1. **Coalesced-first-flush edge (real edge, fixed + pinned):** the mount floor used to apply on any row first render with `itemId && targetLength > 0`. A genuinely NEW assistant message whose first flush coalesced turn-start with its opening deltas mounts a fresh row already carrying prose inside an established pane — and was finalized with no animation ever, violating arrival-vs-viewer. Fix: a `TranscriptPaneLifecycle` context (in `TranscriptContexts.tsx`) tracks `initialPaintComplete` per pane — reset in the render phase on session change (so a swapped session's rows see a fresh lifecycle in the same commit), flipped true by a no-deps effect after each commit. `TranscriptTurnRow` passes it to the frontier hook as `paneEstablishedAtMount`; the floor now applies only when the row mounts as part of the pane's initial paint. The hook takes a parameter rather than reading the context itself to keep hooks/chat/ui free of component imports. Default with no provider = pane-initial (floor applies), the Q15-safe direction for tests/playground. Pinned by three new tests: fresh hook instance with `itemId` + `targetLength > 0` mounted after the pane is established → paces from zero, cache not floored; pane-initial variant through the real `TranscriptContextProviders` → floored/finalized while a late-mounted probe row in the same pane paces; session swap on a live provider → the swapped-in row is floored again.
2. **`transcriptHydrated=false` never trusted (test added):** reconcile test now pins that a slot with an open stream, `current` freshness, and no gap is still rehydrated when `transcriptHydrated` is false.
3. **Cap tiebreak comment:** `hot-session-policy.ts` documents that the lexicographic session-id tiebreak is arbitrary-but-stable and states the cap contract (priority wins strictly; stability prevents stream churn across resolutions). No behavior change.
4. **Render-phase cache write comment expanded:** states why render-time is required (the floor must exist before this render's children — TranscriptItemBlock → AssistantMessage — mount and seed from the cache; an effect runs after first paint, too late to stop a one-frame replay) and the StrictMode/concurrent-safety argument (idempotent + monotonic; double-invoke finds the floor at targetLength and skips; an abandoned render only raised the floor toward "more finalized" and a retry re-derives it).

**Negative controls (round 2):** floor gate reverted to ignore `paneEstablishedAtMount` → exactly the 2 gate-dependent frontier tests fail; `transcriptHydrated` dropped from the reconcile trust condition → exactly the new never-trusted test fails; both restored. Full suite after `pnpm --filter @proliferate/product-client build`: **968 files / 7018 tests passed**.

## Performance dependency (R18 renderer diagnosis)

R18's renderer diagnosis found that each kept-open background stream is an independent roughly-60Hz flush source, and today every flush triggers an app-wide O(all-sessions) directory recompute: executionSummary identity churn defeats the directory store's reference-equality guard. Until the planned directory-scoping rung lands (stabilize executionSummary identity plus per-workspace-scoped activity subscriptions), this PR's up-to-12 background streams multiply that recompute roughly linearly in stream count, which raises the sustained renderer CPU floor on high-workspace-count profiles. Recommendation: land the directory-scoping rung first, or accept a temporarily raised floor on 40-workspace-scale profiles. No code change to this PR; its semantics and trust predicate are unaffected.

## Base pickup (R14 content_stable fix)

R14's honesty-bug fix landed as an append-only commit on `ux-latency/r14-switch-critical-path` (19e16b111, gating deferred content_stable on transcriptHydrated instead of scaffold). Picked up via a plain merge (3d13a5a94) rather than a rebase so the reviewed hardening commit 402a38932 keeps its SHA; regular push, no history rewrite. Zero conflicts (the fix touches only SessionTranscriptPane, no overlap with R15 files). After dist rebuild: R15 suites plus the new SessionTranscriptPane gating test pass (5 files / 32 tests), full suite **969 files / 7020 tests passed**.

### DIFF
--- a/apps/packages/product-client/src/components/workspace/chat/transcript/TranscriptContexts.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/transcript/TranscriptContexts.tsx
@@ -1,6 +1,8 @@
 import {
   createContext,
   useContext,
+  useEffect,
+  useRef,
   type ReactNode,
 } from "react";
 import type { TranscriptOpenSessionRole } from "#product/domain/chats/transcript/transcript-open-target";
@@ -10,7 +12,25 @@ export type TranscriptOpenSessionHandler = (
   role?: TranscriptOpenSessionRole,
 ) => void;
 
+/**
+ * Whether the transcript pane's initial paint for the current session has
+ * committed. Rows that mount as part of that first paint treat their existing
+ * content as finalized history (ADR Q15 restore-finalized); rows that first
+ * appear inside an established pane carry content that arrived while the
+ * viewer was watching, so it may animate. The object is mutated in place —
+ * consumers read it once at their own mount and must not subscribe to it.
+ */
+export interface TranscriptPaneLifecycle {
+  initialPaintComplete: boolean;
+}
+
+// No provider (isolated renders, tests, playground) behaves like a pane-initial
+// paint: mount-time content paints finalized, the safe direction under Q15.
+const DEFAULT_PANE_LIFECYCLE: TranscriptPaneLifecycle = { initialPaintComplete: false };
+
 const TranscriptSessionIdContext = createContext<string | null>(null);
+const TranscriptPaneLifecycleContext =
+  createContext<TranscriptPaneLifecycle>(DEFAULT_PANE_LIFECYCLE);
 const TranscriptOpenSessionContext = createContext<TranscriptOpenSessionHandler | null>(null);
 const TranscriptCanOpenSessionContext = createContext<
   ((sessionId: string, role?: TranscriptOpenSessionRole) => boolean) | null
@@ -27,15 +47,41 @@ export function TranscriptContextProviders({
   canOpenSession?: (sessionId: string, role?: TranscriptOpenSessionRole) => boolean;
   children: ReactNode;
 }) {
+  // Reset in render (not an effect) on session change: the swapped session's
+  // rows first-render in the same commit and must already see a fresh
+  // pane-initial lifecycle, before any effect can run.
+  const paneLifecycleRef = useRef<{
+    sessionId: string;
+    lifecycle: TranscriptPaneLifecycle;
+  }>({ sessionId, lifecycle: { initialPaintComplete: false } });
+  if (paneLifecycleRef.current.sessionId !== sessionId) {
+    paneLifecycleRef.current = {
+      sessionId,
+      lifecycle: { initialPaintComplete: false },
+    };
+  }
+  const paneLifecycle = paneLifecycleRef.current.lifecycle;
+  // Runs after every commit; the first commit for this session's tree flips
+  // the flag, so rows mounted in that commit read false and later ones true.
+  useEffect(() => {
+    paneLifecycle.initialPaintComplete = true;
+  });
+
   return (
     <TranscriptSessionIdContext.Provider value={sessionId}>
-      <TranscriptOpenSessionContext.Provider value={onOpenSession ?? null}>
-        <TranscriptCanOpenSessionContext.Provider value={canOpenSession ?? null}>
-          {children}
-        </TranscriptCanOpenSessionContext.Provider>
-      </TranscriptOpenSessionContext.Provider>
+      <TranscriptPaneLifecycleContext.Provider value={paneLifecycle}>
+        <TranscriptOpenSessionContext.Provider value={onOpenSession ?? null}>
+          <TranscriptCanOpenSessionContext.Provider value={canOpenSession ?? null}>
+            {children}
+          </TranscriptCanOpenSessionContext.Provider>
+        </TranscriptOpenSessionContext.Provider>
+      </TranscriptPaneLifecycleContext.Provider>
     </TranscriptSessionIdContext.Provider>
   );
+}
+
+export function useTranscriptPaneLifecycle(): TranscriptPaneLifecycle {
+  return useContext(TranscriptPaneLifecycleContext);
 }
 
 export function useTranscriptSessionId(): string | null {
--- a/apps/packages/product-client/src/components/workspace/chat/transcript/TranscriptTurnRow.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/transcript/TranscriptTurnRow.tsx
@@ -16,6 +16,7 @@ import {
 } from "#product/components/workspace/chat/transcript/TranscriptTurnChrome";
 import { goalMetMarkerLabel } from "#product/domain/activity/goal";
 import { useSessionGoal } from "#product/hooks/activity/derived/use-session-goal";
+import { useTranscriptPaneLifecycle } from "#product/components/workspace/chat/transcript/TranscriptContexts";
 import { TurnItemSequence } from "#product/components/workspace/chat/transcript/TurnItemSequence";
 import { hostsSynthesizedReceiptDisclosure } from "#product/components/workspace/chat/transcript/TurnWorkspaceReceiptSlot";
 import {
@@ -108,6 +109,7 @@ export function TranscriptTurnRow({
     ? transcript.itemsById[tailAssistantProseRootId]
     : null;
   const tailAssistantTextLength = tailAssistantCopyContent?.length ?? 0;
+  const paneLifecycle = useTranscriptPaneLifecycle();
   const {
     animateAssistantRevealItemId,
     assistantRevealComplete,
@@ -118,6 +120,7 @@ export function TranscriptTurnRow({
     targetLength: tailAssistantTextLength,
     turnCompletedAt: turn.completedAt,
     turnId: turn.turnId,
+    paneEstablishedAtMount: paneLifecycle.initialPaintComplete,
   });
   const visualTurnCompleted = !!turn.completedAt && assistantRevealComplete;
   const assistantEndResource = useMemo(
--- a/apps/packages/product-client/src/hooks/chat/ui/use-assistant-reveal-frontier.test.tsx
+++ b/apps/packages/product-client/src/hooks/chat/ui/use-assistant-reveal-frontier.test.tsx
@@ -1,7 +1,11 @@
 // @vitest-environment jsdom
-import { renderHook } from "@testing-library/react";
+import { render, renderHook } from "@testing-library/react";
 import { beforeEach, describe, expect, it } from "vitest";
 import { useAssistantRevealFrontier } from "#product/hooks/chat/ui/use-assistant-reveal-frontier";
+import {
+  TranscriptContextProviders,
+  useTranscriptPaneLifecycle,
+} from "#product/components/workspace/chat/transcript/TranscriptContexts";
 import {
   clearAssistantRevealProgressForTests,
   getAssistantRevealProgress,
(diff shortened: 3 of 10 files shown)


pablonyx:
## Manual verification

Head SHA verified locally: `3d13a5a94ed109d5d02a94146b6b598403a8365c` (fetch + fast-forward, matches `gh pr view 1969 --json headRefOid`; includes the R14 base pickup merge). Machine check before running anything: `kern.memorystatus_vm_pressure_level` = 1, no cargo/rustc running.

### Scoped vitest, one file at a time, --maxWorkers=2

Files derived from `git diff --stat` against the current base (`ux-latency/r14-switch-critical-path` at 19e16b111):

- `src/hooks/chat/ui/use-assistant-reveal-frontier.test.tsx`: `Test Files 1 passed (1)` / `Tests 7 passed (7)`
- `src/hooks/workspaces/workflows/use-hot-workspace-reconcile-action.test.tsx`: `Test Files 1 passed (1)` / `Tests 5 passed (5)`
- `src/lib/domain/sessions/hot-session-policy.test.ts`: `Test Files 1 passed (1)` / `Tests 8 passed (8)`
- `src/lib/workflows/sessions/hot-session-ingest-manager.test.ts`: `Test Files 1 passed (1)` / `Tests 10 passed (10)`

### Negative control

Most load-bearing behavior for this PR: the gap-free no-refetch trust predicate in `use-hot-workspace-reconcile-action.ts` (`slotStreamLive`). This is the entire zero-fetch guarantee on a clean return, so I broke it directly rather than something less central:

```
const slotStreamLive = false
  && slotBeforeHydrate?.transcriptHydrated === true
  && slotBeforeHydrate.streamConnectionState === "open"
  && ingestFreshness?.freshness === "current"
  && ingestFreshness.gapAfterSeq === null;
```

Rerunning `use-hot-workspace-reconcile-action.test.tsx` produced exactly one failure, the zero-fetch pin, with everything else still green:

```
 ❯ src/hooks/workspaces/workflows/use-hot-workspace-reconcile-action.test.tsx (5 tests | 1 failed) 35ms
   × useHotWorkspaceReconcileAction transcript hydration > performs zero history fetches when the slot stream stayed live without a gap 23ms
   ✓ useHotWorkspaceReconcileAction transcript hydration > repairs a real gap with exactly one incremental history fetch 4ms
   ✓ useHotWorkspaceReconcileAction transcript hydration > falls back to a full replace only when the tail cannot apply contiguously 3ms
   ✓ useHotWorkspaceReconcileAction transcript hydration > never trusts a slot that was not transcript-hydrated, even with a live stream 2ms
   ✓ useHotWorkspaceReconcileAction transcript hydration > does not trust an open stream whose ingest state reports a gap 3ms
 Test Files  1 failed (1)
      Tests  1 failed | 4 passed (5)
```

Assertion that fired:

```
 ❯ src/hooks/workspaces/workflows/use-hot-workspace-reconcile-action.test.tsx:81:27
     80|     expect(outcome).toBe("completed");
     81|     expect(rehydrate).not.toHaveBeenCalled();
```

Reverted the edit and reran clean: `Test Files 1 passed (1)` / `Tests 5 passed (5)`. `git status --porcelain` after restore: empty.

### Hand probe

Cheap temporary vitest harness (`src/__r15-hand-probe.test.tsx`, created, run once, then deleted; no app boot, no browser), driving send, switch-away, switch-back through the real Zustand stores (`session-records`, `session-ingest-store`) and the real production functions (`resolveHotSessionTargets`, `reconcileHotSessions`, `useHotWorkspaceReconcileAction`, `useAssistantRevealFrontier`):

1. Session selected in its home workspace, running: `resolveHotSessionTargets` returns reason `selected`; `reconcileHotSessions` opens the stream (`ensureSessionStreamConnected` called once).
2. Switch away to another workspace while the session keeps running: `resolveHotSessionTargets` now returns reason `background_running` (within the 12-cap). Reconciling against that target: `closeSessionSlotStream` not called, `ensureSessionStreamConnected` still only called once (the same open stream stays hot, no reconnect). Applied a further stream delta while backgrounded; freshness stayed `current`, `transcript.lastSeq` advanced to 42 in the store.
3. Switch back: called the real `useHotWorkspaceReconcileAction` hook. Outcome `completed`, and `rehydrateSessionSlotFromHistory` was not called at all, zero fetches, because the trust predicate found the slot live and gap-free.
4. Mounted `useAssistantRevealFrontier` for a row carrying the store-resident length (42) on first render: `animateAssistantRevealItemId` is null and `assistantRevealComplete` is true, the reveal-progress cache shows `complete: true, visibleLength: 42`. Content that arrived while away paints finalized in one instant, nothing animates.

Test run: `Test Files 1 passed (1)` / `Tests 1 passed (1)`. File deleted after the run; `git status --porcelain` clean.

No merge action taken.

PR #1970, opened 2026-08-16
feat(product): perceived-latency measurement marks (R13)

pablonyx:
## Scope

Adds two new flows to the `renderer.flow.*` diagnostics family
(`lib/infra/diagnostics/renderer-flow-timing.ts`), closing the R12/T4
perceived-latency instrumentation debt. **Measurement only** — no behavior
changes, no new UI, no new thresholds (both budgets stay `thresholdMs: null`,
matching the existing four flows).

1. **Composer submit → first visible token** (`composer_submit`): begins at
   prompt-submit dispatch for the existing-session send path (keyed by
   `sessionId`; new-session sends are out of scope since no `sessionId`
   exists yet at submit time), finishes when the latest transcript turn
   gains its first renderable assistant content in
   `session-stream-flush-apply.ts` (the cheapest truthful proxy available —
   there's no paint hook at this layer).
2. **Mode switch input → paint** (`mode_switch`): begins when a mode config
   change is enqueued (keyed by `intentId`, which PRO-261's tail-coalescing
   keeps stable across a rapid burst of switches — re-begin on each pick
   restarts the clock to the latest input), finishes when
   `dispatchConfigIntent` commits the mode change.
3. **Pane/session switch**: already covered by the existing `session_open`
   flow family (`session-open-flow-marks.ts` + `use-session-history-hydration.ts`).
   Not duplicated, per the existing flow's own instrumentation.

Both new flows are two-point (begin/finish only — no `shell_committed` /
`data_ready` midpoints), so their `content_stable` records omit
`intent_to_shell_ms` / `shell_to_data_ms` and carry timing solely via
`intent_to_stable_ms`. **Field semantics (post-review):** `data_to_stable_ms`
is Honeycomb's "last-mile paint stage" field (time from `data_ready` to
`content_stable`); it is only emitted when `data_ready` was actually reached,
for every flow kind, not just the two new ones. composer_submit/mode_switch
never emit it — their budget metric points at
`renderer.flow.<kind>.intent_to_stable_ms` instead, the field they actually
produce.

## Review round 2 (REQUEST CHANGES → addressed)

- **FINDING 1 (HIGH):** the composer_submit finish check ran on every
  applied stream batch, including reconnect/gap-fill batches, and looked at
  "whatever is `turnOrder`'s last entry" rather than the specific submitted
  turn. A gap-fill batch completing an OLDER turn that was still
  `turnOrder`'s last entry (because the newly submitted turn hadn't started
  yet) would satisfy the check and finish the flow on stale content — a
  bogus duration recorded, and the real completion silently swallowed
  (finish is one-shot). Fixed by binding the flow to the actual turn id the
  submit creates
  (`session-stream-flush-composer-submit-turn.ts`): the first turn id that
  appears in a batch which wasn't present in the prior `turnsById` map is,
  structurally, the submitted turn (a session has at most one
  `composer_submit` in flight — the composer blocks a second concurrent
  send). Dedup is by turn id against the full prior `turnsById` map, not by
  `turnOrder` position, so a gap-fill batch that merely reorders an
  existing turn can't be mistaken for a new one.
- **FINDING 2 (MEDIUM):** removed the `data_to_stable_ms` alias described
  above; see "Field semantics" note in Scope.

## Salvage

A prior agent attempt died mid-work leaving uncommitted changes in a
separate worktree (`/Users/pablohansen/proliferate-worktrees/ux-latency-r13`,
based on a merge of `origin/main` rather than `r1-instrumentation`). Its diff
was sound and applied cleanly against this branch's base with no
adjustment; every line above is that salvaged content, re-verified here.
Nothing was discarded.

## Testing

- Full `@proliferate/product-client` vitest suite (after rebuilding dist,
  since suites import it): **round 1: 961 files / 6977 tests passed;
  round 2 (post-fix): 962 files / 6981 tests passed**.
- Wiring gate (`renderer-flow-timing-wiring.test.ts`) extended with
  `composer_submit_intent` / `composer_submit_stable` /
  `mode_switch_intent` / `mode_switch_commit` entries asserting each mark
  call lives on its correct file.
- `renderer-flow-timing.test.ts` extended with per-flow two-point-timing
  assertions, a PRO-261-coalescing re-begin test, and (round 2) assertions
  that `data_to_stable_ms` is omitted whenever `data_ready` was never
  reached.
- New `session-stream-flush-composer-submit-turn.test.ts` covers: binding
  to a genuinely new turn id, the exact FINDING 1 regression (a gap-fill
  batch completing an older turn — which IS `turnOrder`'s last entry at
  that point, since the submitted turn hasn't started — must not bind or
  finish), clearing on finish, and rebinding if a prior binding was never
  cleared.
- **Negative controls (mandatory, all performed)**:
  - Round 1: removed the `beginRendererFlow` call from
    `use-chat-prompt-actions.ts` → `composer_submit_intent` wiring test
    failed as expected → restored → suite green again.
  - Round 1: removed the `finishRendererFlow` block from
    `session-stream-flush-apply.ts` → wiring test failed as expected →
    restored → suite green again.
  - Round 2 (FINDING 1 fix): reverted
    `resolveComposerSubmitTargetTurnId` to the old "just use `turnOrder`'s
    last entry" behavior → the new regression test failed (2 of 4 tests
    failed, including the exact gap-fill scenario) → restored → suite
    green again.

Not merged.


pablonyx:
## Manual verification

Head SHA: `6018adc4a564b874f2e9b7393ff93b2dfcac0e1a` (confirmed equal to local branch HEAD and to `gh pr view 1970 --json headRefOid`).

### Scoped vitest, one file at a time, `--maxWorkers=2`

Files derived from `git diff --stat origin/ux-latency/r1-instrumentation...HEAD -- '*.test.ts' '*.test.tsx'`.

`src/hooks/sessions/lifecycle/session-stream-flush-composer-submit-turn.test.ts`
```
 Test Files  1 passed (1)
      Tests  4 passed (4)
```

`src/lib/infra/diagnostics/renderer-flow-timing-wiring.test.ts`
```
 Test Files  1 passed (1)
      Tests  10 passed (10)
```

`src/lib/infra/diagnostics/renderer-flow-timing.test.ts`
```
 Test Files  1 passed (1)
      Tests  16 passed (16)
```

### Negative control (most load-bearing fix: turn-id-bound submit finish)

Reverted `resolveComposerSubmitTargetTurnId` in `session-stream-flush-composer-submit-turn.ts` to the pre-fix behavior (`return after.turnOrder.length > 0 ? after.turnOrder[after.turnOrder.length - 1] : null;`), uncommitted, then reran the pinning test file:

```
 FAIL  src/hooks/sessions/lifecycle/session-stream-flush-composer-submit-turn.test.ts > resolveComposerSubmitTargetTurnId > does not bind (and so cannot finish) on a gap-fill batch that completes an older turn before the submitted turn has even started (FINDING 1 regression...)
AssertionError: expected 'old-turn' to be null

- Expected:
null

+ Received:
"old-turn"
```
(A second test, "clears the binding so a later submit on the same session starts unbound", failed the same run for the same reason.) Restored the file, reran the same test file:
```
 Test Files  1 passed (1)
      Tests  4 passed (4)
```
`git status --porcelain` after restore: empty.

### Hand probe

Beyond the committed suite, I wrote a temporary (uncommitted, deleted after use) vitest file that drove the real `beginRendererFlow` / `resolveComposerSubmitTargetTurnId` / `turnHasAssistantRenderableTranscriptContent` / `finishRendererFlow` functions through a synthetic three-batch sequence: a gap-fill batch completing an unrelated older turn, then the real submitted turn starting, then assistant content landing on it. Captured console output:

```
[hand-probe] t=5 gap-fill batch: resolved target turn = null
[hand-probe] t=12 submit-starts batch: resolved target turn = "real-submitted-turn"
[hand-probe] t=30 content batch: hadBefore=false hasAfter=true
[hand-probe] recorded renderer diagnostic sequence = ["renderer.flow.intent","renderer.flow.content_stable"]
[hand-probe] content_stable fields = {"flow_kind":"composer_submit","stages_completed":0,"intent_to_stable_ms":30,"budget_metric":"renderer.flow.composer_submit.intent_to_stable_ms","budget_threshold_ms":null}
```

The gap-fill batch resolves to `null` (does not bind to the unrelated older turn, matching Finding 1's fix), the real submitted turn is the one that gets bound once it appears, `content_stable` only fires once assistant content lands on that specific bound turn, and the emitted fields have no `data_to_stable_ms` (matching Finding 2's fix). No app boot, no browser, no cargo, no Docker; this was a plain vitest process with a synthetic diagnostics sink.

PR #1971, opened 2026-08-16
feat(product): DotCellLoader hero tier as the chat loading mark (R16)

pablonyx:
## Spec (verbatim, R16 of the UX Latency ladder)

Add a "hero" size tier to DotCellLoader, sized for the chat loading hero.

1. `DotCellLoader.tsx` — extend `DotCellLoaderSize` to `"compact" | "default" | "hero"`. No other change.
2. `product.css` — add `.dot-cell-loader[data-size="hero"]` with `--dot-cell-size: 0.375rem; --dot-cell-gap: 0.25rem;` (6px dots, 4px gap → 26px square).
3. `ChatLoadingHero.tsx` — pass `size="hero"` to DotCellLoader, drop the caption/workspaceName paragraphs (mark only). Leave the ThinkingText (`awaiting-first-turn`) branch alone.
4. Mark visible ≥~420ms once mounted (100ms load can't flash), then fade 120ms before the transcript's content-fade-in. A bootstrapped workspace (`hasWorkspaceBootstrappedInSession(id) === true`) must not mount the hero at all.
5. Reduced-motion coverage on `.dot-cell-loader__dot` must still hold at the new size.

## What changed

- `DotCellLoader.tsx`: `DotCellLoaderSize` now includes `"hero"`; component behavior unchanged (still forwards `size` to `data-size`).
- `product.css`: new `[data-size="hero"]` rule alongside the existing `compact` rule; keyframes/delay calc untouched (size-independent, per spec).
- `ChatLoadingHero.tsx`: swaps `ProliferateLivingMark` for `<DotCellLoader size="hero" />` on the non-`awaiting-first-turn` branch, drops the caption/workspaceName paragraphs from both branches (mark-only per substep), and gates the whole loading branch on `hasWorkspaceBootstrappedInSession(selectedWorkspaceId)` — returns `null` outright for an already-bootstrapped workspace. `ThinkingText` (`awaiting-first-turn`) is unchanged except it also drops the caption/workspaceName paragraphs. Adds an `onTreatmentShown` callback prop so `ChatView` can time its own exit hold (see below).
- `LoadingBoundary.tsx`: adds an optional `onTreatmentShown?: () => void` prop, fired the instant the treatment mounts (show-delay elapsed) — a minimal, backward-compatible extension every other `LoadingBoundary` caller ignores.
- `motion.ts`: adds `motion.loading.heroMinDisplayMs = 420` (the R16 hero floor, distinct from the generic `minDisplayMs = 300` since the hero is a larger, more prominent mark). The 120ms fade reuses the existing `motion.duration.exitMs` token (already 120ms) rather than adding a new one.
- `ChatView.tsx`: implements the hold + fade for real. `ChatLoadingHero` hardcodes `state="pending"` on `LoadingBoundary` for its whole life — resolution here isn't a local state transition, it's `ChatView` unmounting `ChatLoadingHero` synchronously when `mode.kind` changes away from `workspace-status`/`session-loading` — so `LoadingBoundary`'s own min-display/fade-out machinery never engages from inside that component. `ChatView` now owns the hold itself via a new `useChatLoadingHeroExit(isHeroMode)` hook:
  - `ChatLoadingHero` calls `onTreatmentShown` the instant its mark becomes visible; the hook records that timestamp.
  - When `mode.kind` flips away from hero mode, the hook computes `remaining = max(0, heroMinDisplayMs - elapsed)` and enters a `"holding"` phase for `remaining` ms, then a `"fading"` phase for `exitMs` (120ms), then `"idle"`.
  - While `phase !== "idle"`, `ChatContent` renders a frozen `ChatLoadingHeroExitOverlay` (a standalone `<DotCellLoader size="hero" />` in the same `ChatPreMessageCanvas` shell, not a live `ChatLoadingHero` — this sidesteps `ChatLoadingHero`'s internal hooks re-deriving from the now-already-changed session/workspace state) absolutely positioned on top of whatever the switch already resolved to underneath (transcript, ready hero, etc.), opacity 1 while holding/idle-transition-in, opacity 0 (CSS `transition-opacity`, `exitMs` duration) while fading. The real content underneath is already mounted and running its own `content-fade-in`, so the hero's fade-out overlaps/precedes that reveal.
  - If the mark never became visible before the mode flipped away (resolved inside the 200ms show-delay), there's nothing to hold — `phase` stays `"idle"` and the switch takes effect immediately, same as before R16.
  - If `mode.kind` returns to hero mode while an old overlay is still holding/fading (a fast re-load), the hook resets to `"idle"` immediately so the real, live `ChatLoadingHero` takes over rather than doubly rendering.
  - No new animation library: plain `useState`/`useEffect`/`setTimeout` state machine plus a CSS opacity transition, per the explicit instruction not to add one.

## Tests

`ChatLoadingHero.test.tsx`:
- Show-delay still withholds the treatment; after the delay, `[data-dot-cell-loader]` mounts with `data-size="hero"`; no caption/workspace-name text renders.
- `awaiting-first-turn` still renders `ThinkingText` immediately, bypassing the show-delay; no DotCellLoader.
- A workspace with `hasWorkspaceBootstrappedInSession(id) === true` never mounts the hero (`[data-dot-cell-loader]` absent, `[data-chat-loading-hero]` absent, container empty).
- Negative control: a *different*, non-bootstrapped workspace id still mounts the hero mark — proves the guard is keyed to the current workspace.

`ChatLoadingHeroExit.test.tsx` (new, exercises `useChatLoadingHeroExit` + `ChatLoadingHeroExitOverlay` exported from `ChatView.tsx`, with fake timers):
- Mode flips away at 250ms after the mark shows → overlay stays fully opaque (holding) through the remainder of the 420ms floor, then transitions to opacity 0 exactly at 420ms, stays mounted through the 120ms fade, and unmounts cleanly at 540ms.
- Mode flips away at 600ms (past the floor) → overlay is already fading (opacity 0) immediately, no extra hold, and unmounts after `exitMs`.
- Mark never becomes visible before the mode flips away (still inside the show-delay) → no overlay ever mounts.
- Staying in hero mode for 10s → `phase` stays `"idle"`, no overlay, mark never hidden mid-flight.
- Negative control sanity-check that the harness itself produces no incidental overlay markup.

**Negative controls performed and reverted:**
1. Removed the `hasWorkspaceBootstrappedInSession` guard from `ChatLoadingHero.tsx` → the bootstrapped-workspace test failed (`[data-dot-cell-loader]` was present) as expected; restored, reconfirmed green.
2. Removed the hold-timer scheduling from `useChatLoadingHeroExit` (mode flip always went straight to idle) → both hold/fade timing tests failed (`expected null not to be null`, overlay never mounted) as expected; restored, reconfirmed green.

Verified `playground.test.ts` (mounts the real `ChatLoadingHero` via `PlaygroundLoadingStates`) still passes unmodified — its string assertions come from sibling playground fixtures, not `ChatLoadingHero`'s own copy.

Confirmed reduced-motion: `.dot-cell-loader__dot`'s `@media (prefers-reduced-motion: reduce)` block has no size-scoped selector, so it applies unchanged at `data-size="hero"`.

Rebuilt `@proliferate/design` (CSS/theme dist) and `@proliferate/product-client` (dist) before testing. Full `@proliferate/product-client` vitest suite: **966 files / 6994 tests passed.**

## Base

Based on `ux-latency/r4-skeleton-retirement` (the surfaces this spec touches live on that stack, not `main`). Sibling PR #1968 (R14) shares this base and does not touch these files.

## Founder ruling: no visible text beside a loading indicator (mark only, everywhere on this stack)

Extends the hero's "mark only" treatment into a stack-wide rule: no visible label paragraph beside any loading mark on the chat loading surfaces in scope here.

- `TranscriptSwitchingPlaceholder.tsx`: removed the visible `<p>{label}</p>` beside the `ProliferateLivingMark` treatment. `label` is now accessibility-only — it stays as the `aria-label` on the outer `role="status"` container; screen readers still get it, sighted users see the mark alone.
- Updated `TranscriptSwitchingPlaceholder.test.tsx`: the old "mark with visible label" assertion is now "mark alone" — `gutter.textContent` must **not** contain the label text (previously asserted the opposite). Ran + reverted a negative control: restoring the visible `<p>` makes this test fail (`expected 'Switching chat' not to contain 'Switching chat'`) as expected, then re-removed it and reconfirmed green.
- `playground.test.ts` (mounts `PlaygroundLoadingStates`, which includes `TranscriptSwitchingPlaceholder`) still passes unmodified.
- Swept the stack for other loading treatments with visible captions beside a mark: **`LoadingState`** (`components/feedback/LoadingIllustration.tsx`) renders a visible `message`/optional `subtext` under its skeleton icon (e.g. "Loading file" / "README.md" in the playground fixtures). This is a candidate for the same ruling, but it serves non-chat surfaces (used standalone, e.g. file-loading contexts) — **flagging it, not changing it, without separate confirmation** since it's outside the surfaces this ruling was scoped to.

Full `@proliferate/product-client` vitest suite re-run after this change: 966 files / 6994 tests passed.

## Review round 2 (two MED findings, fixed)

**Finding 1: stale `shownAtRef` leak across hero cycles.** In `use-chat-loading-hero-exit.ts`, the `if (isHeroMode) { setPhase("idle"); return; }` branch reset `phase` but not the shown timestamp. Failure sequence: session A's hero shows its mark, the user switches to session B before A's hold/fade completes, B is already bootstrapped so `ChatLoadingHero` returns null and `onTreatmentShown` never fires for B, B resolves, and the exit effect saw `wasHeroMode` true with the stale `shownAt` from A, computed a large elapsed time and zero remaining, and mounted a spurious exit overlay over B's already-resolved content, defeating the bootstrap short-circuit.

Fix: also clear `shownAtRef.current = null` in that branch, so every hero cycle gates strictly on its own `handleTreatmentShown` call. Added a test in `use-chat-loading-hero-exit.test.tsx`: hero shows a mark, resolves at 250ms (still holding), switches straight to a second, already-bootstrapped-workspace hero cycle (no mark shown), then resolves again; asserts the exit overlay never mounts for that second cycle. Negative control performed: reverted the one-line fix, the new test failed with `expected null not to be null` at the final assertion, restored, reconfirmed green (6/6 in that file).

**Finding 2: accessibility regression on the mark-only hero.** Neither `ChatLoadingHero.tsx` nor `ChatLoadingHeroExitOverlay.tsx` exposed any role/aria on the mark-only path, so the loading state was silent to assistive tech.

Fix: `ChatLoadingHero.tsx` now wraps both the `ThinkingText` and `DotCellLoader` treatment branches in `role="status" aria-label="Loading conversation"` (same pattern as `TranscriptSwitchingPlaceholder`), with the `LoadingBoundary` itself marked `aria-hidden="true"` so its inner content isn't separately exposed. `ChatLoadingHeroExitOverlay.tsx` gets `aria-hidden="true"` since it's a decorative echo of the hero's own announcement, not a second status region: screen readers should hear one loading announcement, not two. Added a test in `ChatLoadingHero.test.tsx` asserting `getByRole("status", { name: "Loading conversation" })` resolves and its `textContent` is empty (mark-only, no visible text), and extended `use-chat-loading-hero-exit.test.tsx` to assert the exit overlay carries `aria-hidden="true"`. Negative control performed for both: reverting the `role`/`aria-label` addition failed the new `ChatLoadingHero.test.tsx` assertion (`getByRole` threw, element not found), and reverting `aria-hidden` on the overlay failed the new exit-overlay assertion (`expected null to be 'true'`); both restored, reconfirmed green.

Repo-shape and boundary checks stay clean after these edits (`report_frontend_structure.py --strict` and `check_frontend_boundaries.py` both pass, TOTAL: 0). Full `@proliferate/product-client` vitest suite re-run: 965 files / 6995 tests passed, 1 unrelated pre-existing flaky test in `ComposerCommandEditor.test.tsx` (untouched by this PR) failed on the full-suite run and passed cleanly in isolation.


pablonyx:
## Manual verification

Head SHA: f4c77626d75d73e2afb8790f3ac7b2b267ea6b2e (confirmed equal to `gh pr view 1971 --json headRefOid`).

### Scoped test runs (one file at a time, --maxWorkers=2)

Touched test files derived from `git diff --stat 0e503861e^..HEAD -- '*.test.tsx' '*.test.ts'`:

```
src/components/workspace/chat/surface/ChatLoadingHero.test.tsx (5 tests)
src/components/workspace/chat/surface/TranscriptSwitchingPlaceholder.test.tsx (3 tests)
src/hooks/chat/ui/use-chat-loading-hero-exit.test.tsx (6 tests)
```

Literal per-file tallies:

```
Test Files  1 passed (1)
     Tests  5 passed (5)
```
(ChatLoadingHero.test.tsx)

```
Test Files  1 passed (1)
     Tests  3 passed (3)
```
(TranscriptSwitchingPlaceholder.test.tsx)

```
Test Files  1 passed (1)
     Tests  6 passed (6)
```
(use-chat-loading-hero-exit.test.tsx)

### Negative control: the exit-hold itself

The reviewer already hand-verified the shownAtRef reset (round 2), so this control targets the other load-bearing behavior: the 420ms min-display hold-then-fade path in `use-chat-loading-hero-exit.ts`. Reverted `remainingMs = Math.max(0, motion.loading.heroMinDisplayMs - elapsedMs)` to `remainingMs = 0` (uncommitted, local only), reran the pinning test file:

```
FAIL  src/hooks/chat/ui/use-chat-loading-hero-exit.test.tsx > useChatLoadingHeroExit / ChatLoadingHeroExitOverlay (R16 hero exit hold) > holds the exit overlay to the 420ms floor when the mode flips away at 250ms, then fades 120ms
AssertionError: expected '0' to be '1' // Object.is equality
```

Restored the file, reran clean:

```
Test Files  1 passed (1)
     Tests  6 passed (6)
```

`git status --porcelain` empty after restore (confirmed).

### Hand probe

Temporary jsdom/fake-timer harness (`__manual_probe.test.tsx`, created, run, deleted; no app boot, no browser) drove the full hero lifecycle: mount in hero mode, fire `handleTreatmentShown`, advance 250ms, flip to resolved (still inside the 420ms hold), advance to the 420ms floor, advance the 120ms fade.

Observed phase sequence: `["idle","idle","holding","fading","idle"]`.

Observed at each step:
- After resolve at 250ms: overlay present, `opacity=1`, `aria-hidden=true`.
- At the 420ms floor: `opacity=0` (fade started).
- After the 120ms fade: overlay unmounted (`present=false`).

Separately rendered `ChatLoadingHero` directly (mocked substep/session/bootstrap deps) and confirmed the accessible status region: `role="status"` resolves via `getByRole("status", { name: "Loading conversation" })`, and `textContent` is empty (mark-only, no visible text; the label lives only in the aria-label).

Probe file was deleted after use; working tree is clean.

PR #1973, opened 2026-08-16
feat(product-client): FR-2 finalized-session revisit restore (PRO-187, r6)

pablonyx:
## Rung 6 of the Chat Scroll ADR ladder (PRO-187): the FR-2 revisit contract.

Finalized sessions now reopen at the reading position the user left them at, restored before first paint with no gradual motion; actively streaming sessions still bottom-pin on entry. Position is persisted as `{lastRowKey, offsetWithinRow}` (never raw scrollTop) so virtualizer estimates cannot skew the restore, and `resetForSession` becomes contract-aware instead of unconditionally pinning + snapping + gluing.

### What this rung adds
- `transcript-reading-position-store.ts` — bounded in-memory per-session `{rowKey, offsetWithinRowPx}` store (LRU cap, no localStorage; same house pattern as the rung-5 height cache and reveal-progress cache), plus the anchor-capture, restore-target inversion, and placement helpers.
- `use-transcript-reading-position.ts` — captures the reading anchor on every scroll and builds the session-switch placement plan: `bottom` for a streaming or never-read session, `restore` otherwise.
- `resetForSession(plan?)` is now contract-aware: it bottom-pins a streaming session / missing plan / saved-row-gone (the conservative default), or places the saved reading anchor and arms the single frame writer to re-resolve it while heights settle.
- The single per-frame writer (rung 4 pipeline) re-resolves the anchor each glued frame until a bounded deadline, so residual estimate-to-measured corrections land silently.

### Estimate-immune restore
The restore resolves the saved row's scrollTop from the **mounted anchor row's real rendered rect**, not an index-sum of estimated heights. The coarse index-sum estimate is used only to bootstrap the row into the render window; once mounted, the DOM-relative target seats the exact reading row under the top edge even when never-measured rows above it are estimated to a wrong total. This is the skew FR-2 exists to avoid.

### PRO-175 mechanism (already landed rung 2, upheld here)
The cross-session submit-stamp ref is session-scoped and reset in `resetForSession`, so a stale stamp can never fire a snap on session switch. The revisit-no-motion scenario, previously `test.fixme` (deferred to this rung), is un-fixmed and passing.

### Saved-row-gone
Per the ADR consequence default, an unresolvable saved anchor falls back to the conservative bottom-pin rather than stranding the reader (no open founder question needed).

### Tests (each negative-controlled)
Physics (Playwright, real Chromium + WebKit):
- `revisit-restore` — the SAME reading row lands under the top edge, unpinned, and the placement lands-and-holds (no crawl).
- `revisit-streaming` — a streaming revisit bottom-pins, ignoring any saved position.
- `revisit-row-gone` — a saved row that no longer exists falls back to bottom-pin.

JSDOM units cover the store, the anchor capture/inversion (estimate-immune + coarse fallback + saved-row-gone), and `resetForSession` restore placement / bottom fallback / user-scroll clears the restore.

### Local results
- `report_frontend_structure.py` TOTAL: 0.
- Unit: `src/hooks/chat/ui` 205 passed.
- Physics full suite: 21 passed, 3 skipped (rung-3/rung-8 pending scenarios, unrelated); FR-2 trio green on both engines at repeat-each=2.
- Typecheck (app + scroll-physics) clean.

## Testing
Playwright scroll-physics tier (both engines) + JSDOM units, per the ADR Q17 ruling (no simulated-DOM shortcut for scroll physics).

## Observability
No new telemetry this rung (diagnostics are rung 11).

### DIFF
--- a/apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts
+++ b/apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts
@@ -537,48 +537,124 @@ test.describe("transcript scroll physics", () => {
     await drive(page, "finalizeStreamingTurn");
   });
 
-  // EXPECTED TO FAIL today (PRO-175), for a DIFFERENT reason than the stamp
-  // leak this rung fixes. `resetForSession` still unconditionally
-  // pins/snaps/glues on every session switch by design (rung 2 leaves this
-  // alone), and the virtualizer's own remeasurement of freshly-mounted rows
-  // still produces a handful of settle frames after the snap — plus, in this
-  // fixture, a session switch resets `VirtualTranscriptRowList`'s
-  // `fallbackReason` to null, which can flip the tree from
-  // FullTranscriptRowList back to VirtualizedTranscriptRowList (a real
-  // component remount) if a blank-viewport fallback had fired for the
-  // intervening session. Either path produces multiple distinct scrollTop
-  // frames regardless of the stamp fix below. Verified: with
-  // `setPendingPromptStamp` reproducing the stamp-leak precondition
-  // (session-primary carries a stamp across an intervening session with
-  // none), the pre-fix code and the sessionKey-scoped fix in
-  // use-transcript-stick-to-bottom.ts produce IDENTICAL scrollTop traces
-  // (~5 distinct frames each) — the stamp leak was not this test's failure
-  // mode. Zero-motion placement on revisit is restore-finalized placement,
-  // named in the frozen ladder as rung 6 (bottom-on-entry replacement);
-  // unfixme there.
-  test.fixme(
-    "revisit-no-motion: switching back to a finalized session places with zero motion frames",
-    async ({ page }) => {
-      await ready(page);
-      await drive(page, "reset");
-      await drive(page, "seedFinalizedConversation", 8);
-      await waitForViewport(page);
-      await settle(page);
+  // Rung 6 (PRO-187, FR-2) un-fixmes revisit-no-motion. Before rung 6,
+  // `resetForSession` unconditionally bottom-pinned + glued on every session
+  // switch, so returning to a finalized session snapped to the bottom and then
+  // crawled through the freshly-mounted rows' settle frames (multiple distinct
+  // scrollTop values). Rung 6 restores the saved {lastRowKey, offsetWithinRow}
+  // before first paint against rung-5's warmed measured heights, so the reading
+  // position is placed in a single instant cut with no settle crawl. Negative
+  // control: disable beginSessionRestorePlacement (transcript-reading-position-
+  // store.ts) and this lands at the bottom instead of the saved position.
+  test("revisit-restore: returning to a finalized session restores the reading position with zero motion", async ({
+    page,
+  }) => {
+    await ready(page);
+    await drive(page, "reset");
+    await drive(page, "seedFinalizedConversation", 12, "sess-restore-a");
+    await waitForViewport(page);
+    await settle(page);
 
-      // Visit a second finalized session, then return to the first.
-      await drive(page, "switchSession", "session-secondary", 6);
-      await settle(page);
+    // Read up to a mid position; the row list persists this session's reading
+    // anchor ({lastRowKey, offsetWithinRow}) on the scroll event. A portable
+    // keyboard gesture (not a target-distance write) avoids racing the freshly-
+    // mounted rows' measurement, which would keep shifting an absolute target.
+    await keyboardScrollUp(page, 3);
+    await settle(page, 500);
+    const saved = await metrics(page);
+    expect(saved.bottomDistance).toBeGreaterThan(REPIN_BAND_PX * 4);
+    await expect.poll(() => isPinned(page), { timeout: 2000 }).toBe(false);
+    // The observable reading position: the transcript row under the top edge.
+    const savedTopRow = await drive<string | null>(page, "getTopVisibleText");
+    expect(savedTopRow, "a transcript row must be under the top edge").toBeTruthy();
 
-      await drive(page, "startScrollTrace");
-      await drive(page, "switchSession", "session-primary", 8);
-      await settle(page, 600);
-      const trace = await drive<number[]>(page, "stopScrollTrace");
+    // Visit a second finalized session, then return to the first.
+    await drive(page, "switchSession", "sess-restore-b", 6);
+    await settle(page);
 
-      // Zero visible motion after placement: every recorded frame equal.
-      const distinct = new Set(trace.filter((v) => Number.isFinite(v)));
-      expect(distinct.size).toBeLessThanOrEqual(1);
-    },
-  );
+    await drive(page, "startScrollTrace");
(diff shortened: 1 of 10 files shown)


pablonyx:
CI verdict on fix head 3e3d7bbcf: Transcript scroll physics (chromium + webkit) pass 3m47s (run 31921088923). The webkit prepend regression introduced by the r6 restore delta is fixed; all FR-2 scenarios and the full physics suite pass on both engines. Prior round at 2494e516c was 20/21 with [webkit] prepend anchoring Expected > 150 / Received 0.

pablonyx:
Adversarial review round 1: CLEAN (ship-ready; low-severity observations only, no fix required this rung).

CI verified literally via `gh pr checks 1973`: 18 pass / 0 fail / 0 pending, including "Transcript scroll physics (chromium + webkit)" pass and "Shared frontend packages" pass. The prior [webkit] "prepend anchoring Expected > 150 / Received 0" is now green.

What I attacked and why it holds:

(1) FR-2 discrimination source of truth. `buildSessionRestorePlan` reads `isSessionBusyRef.current` at switch time via a stable callback; the consuming `useLayoutEffect` depends only on `activeSessionId`/`selectedWorkspaceId`, so mid-session streaming-state churn cannot re-fire the reset. Finalized-while-backgrounded revisits as finalized (restore); streaming-immediately-after-restore does not yank because entry was finalized and FR-2 only bottom-pins on entry. Correct.

(2) Capture timing / pollution. Anchoring on {rowKey, offsetWithinRow} rather than raw scrollTop makes programmatic-write re-captures idempotent for a restore, and a streaming bottom-pin captures the bottom row (semantically the reader's position). Persistence is in-memory only, deliberately not surviving reload; that matches the frozen "within-runtime revisit contract" and the rung-5-warmed-heights rationale. Session keying is `${workspaceId}:${sessionId}`, the same identity the engine scopes to. No real pollution path found.

(3) Pre-first-paint honesty. `resetForSession(buildSessionRestorePlan())` runs inside a `useLayoutEffect` (VirtualizedTranscriptRowList.tsx:248), so placement lands before paint. Estimate-immune resolution reads the mounted row's real `getBoundingClientRect`; the coarse `getOffsetForIndex` index-sum is used only to mount the row, and the rung-5 per-rowKey measured-height cache (cross-session, in-runtime) keeps that coarse estimate near-true for rows the reader scrolled through. The `target <= maxTop + 1` reachability guard in both the placement and the frame writer prevents painting a clamped intermediate.

(4) Saved-row-gone fallback. `resolveTranscriptRestoreTargetTop` returns null when the row key is absent; `beginSessionRestorePlacement` then returns false and the engine bottom-pins (conservative default). Unit-tested (store test "returns null when the saved row no longer exists") and e2e-tested ("revisit-row-gone" asserts pinned + bottomDistance <= PIN_FOLLOW_MAX_DISTANCE_PX, with a stated negative control).

(5) Rung interactions. r3 single-writer preserved: the restore is one more anchor kind in the single frame pass, and `isGluing` now also holds while `restoreResolverRef != null`. r4 one-snap-per-frame preserved (restore branch `return`s). r5 clamp/calibration: the writer only writes reachable targets and re-resolves each frame. PRO-175 submit-stamp does not fight restore: the pinned branch runs and returns BEFORE the restore branch, so a submit repin cleanly supersedes an in-flight restore (submit => bottom, correct). The webkit fix (3e3d7bbcf) is principled: it holds the compensation anchor past the growth deadline ONLY while the reader is displaced ABOVE the seat computed from max-observed height, bounded by the SAME absolute ceiling, and ordinary below-viewport post-deadline growth still releases. No thresholds changed.

(6) Test honesty. Unit tests cover place+hold+refine, saved-row-gone -> bottom-pin, and user-scroll-clears-resolver. The e2e "revisit-restore" un-fixmes revisit-no-motion meaningfully: it asserts the SAME top-visible row lands (estimate-immune, via getTopVisibleText with the volatile timestamp stripped), NOT bottom-pinned, distinct-position set <= 3, and a flat held tail (no post-landing drift). Negative controls are documented as manual procedures, consistent with this ladder's e2e convention.

(7) Covered under (2)/(5).

(8) Scope purity. Diff is confined to the reading-position store/hook, its wiring, the frame-pipeline/engine restore branch, the webkit regression fix, and tests. Nothing beyond rung 6.

Low-severity observations (no action needed this rung):
- The ADR R1 names a `restoring` state; the engine still represents pin/compensation/restore implicitly via refs. The explicit named-state-machine refactor is explicitly deferred to rung 7 ("encode ... as state transitions rather than refs"), so representing restore as a ref here is ladder-consistent, not a gap.
- Residual theoretical edge: if the coarse index-sum estimate for the saved row exceeded reachable `maxTop` every frame AND the row never entered the render window, the writer would skip all writes and, after the 500ms deadline, release with the reader at the outgoing scrollTop rather than bottom-pinning. In practice a saved row implies the reader scrolled through the rows above it, so the rung-5 cache makes the estimate accurate and the row reachable; I could not construct a real trigger within the runtime contract. Flagging only as a future hardening candidate (fall back to bottom-pin if the restore never lands by the deadline).

Verdict: CLEAN. Do not merge (awaiting Pablo).

pablonyx:

Branch chat-scroll/r6-fr2-restore checked out, product-client dependencies built via pnpm shared:build. Ran each touched test file individually with pnpm vitest run <file> --reporter=basic:

- src/hooks/chat/ui/transcript-reading-position-store.test.ts: Test Files 1 passed (1), Tests 11 passed (11)
- src/hooks/chat/ui/use-transcript-stick-to-bottom.compensation.test.tsx: Test Files 1 passed (1), Tests 11 passed (11)
- src/hooks/chat/ui/use-transcript-stick-to-bottom.test.tsx: Test Files 1 passed (1), Tests 23 passed (23)

Negative control (finalized-restore vs streaming-bottom-pin, jsdom harness): in transcript-reading-position-store.ts, beginSessionRestorePlacement's guard `if (plan.kind !== "restore") return false;` was forced to `if (true || plan.kind !== "restore") return false;`, unconditionally falling back to bottom-pin (the streaming/no-plan behavior) even for a resolvable restore plan. Reran use-transcript-stick-to-bottom.test.tsx. Literal failure:

```
❯ src/hooks/chat/ui/use-transcript-stick-to-bottom.test.tsx (23 tests | 2 failed)
  × resetForSession restore places at the resolved target, unpins, and holds across settle
    → expected true to be false (isPinnedToBottom)
  × resetForSession restore is cleared by a user scroll so the writer never fights the reader
    → expected 2000 to be 450
Test Files  1 failed (1)
     Tests  2 failed | 21 passed (23)
```

Restored the original guard via Edit; `git status` confirms working tree clean, no diff remains.

pablonyx:
Stack refold receipt (overnight staging): rebased onto main @ d5d38a804 (r1 merged) atop the retargeted stack. git range-diff vs previous head: 0 semantic changes (6/6 commits identical). New head 7d084ee08. Review carried from c95d5abd8.

pablonyx:
Merge-train refold receipt: rebased onto main @ post-r5 squash. 7d084ee08 -> 875639715. range-diff: 6/6 commits byte-identical (=), review carries.

PR #1974, opened 2026-08-16
fix(product): keep the sidebar selection highlight always animating

pablonyx:
## Finding 1: transition never plays on rapid re-selection (fixed in the first commit)

The workspace row highlight lives in `apps/packages/product-client/src/primitives/patterns/sidebar/SidebarRowSurface.tsx`. It applies a plain CSS transition (`transition-[background-color,color,opacity] duration-hover`), with the active/inactive class swapped via a boolean `active` prop, on a row keyed stably on `item.id` (no remount, no keyframe animation, no reduced-motion/instant flag involved for this path, all ruled out by direct code inspection).

The row's `active` boolean is recomputed fresh every render from the selection store (no stale memoization). Each click synchronously updates the store and produces its own React commit. The bug is not a React batching or stale-prop issue: it is the browser's rendering pipeline coalescing rapid same-frame style mutations. When workspace switches land faster than a paint, the same row's `active` value can flip true/false/true across several commits before the browser ever paints an intermediate frame. CSS transitions only animate between the style actually painted in the last frame and the style computed in the next painted frame; any values written and overwritten in between are invisible to the transition engine. If the row that ends up finally selected nets out to the same value it last had a real paint for (true -> true across the skipped window), no transition fires even though `data-active`/`bg-selected` land correctly in the DOM.

First fix: settle the class-driving value one animation frame behind `active` via a `useSettledActive` hook, guaranteeing a real, previously-painted "before" frame exists for the transition to animate from.

## Finding 2: no highlight at all during a fast Cmd+Opt+Arrow sweep (fixed in the second commit)

Live-testing a build with the first fix still reproduced a related but distinct symptom: sweeping through workspaces very quickly shows no sidebar highlight at all until movement stops.

Measured root cause: every workspace switch runs a 130-200ms main-thread long task. During a fast sweep the browser only manages a handful of paints across the whole sequence (roughly 5fps, frame gaps of 130-240ms). The row highlight (`bg-selected` at 3.2% alpha) fades in over `duration-hover` (120ms). At roughly 5fps, every painted frame catches that fade at close to zero alpha, so the highlight never becomes visible until the sweep stops and a fade finally gets enough frames to finish. No transition-triggering fix can help here because there are no frames to animate across.

Worse, the `useSettledActive` rAF-based settle from finding 1 made this failure mode worse rather than better: its own animation-frame callback is starved by the exact same main-thread saturation, so under a fast sweep the highlight application itself could be delayed indefinitely (matching the reported "no highlight until movement stops"), not merely fail to animate.

Second fix: removed `useSettledActive` (it no longer has a role: see below) and instead exclude `background-color` from the transition set while a row is active, keeping it in the set while inactive. Activating a row now paints `bg-selected` solid on the very first available frame regardless of frame rate, so it is visible even at 5fps. Deactivating a row is unaffected by the starvation (it does not need to be seen instantly) and keeps the fade, so hover/deselect still feels soft. Keyed on the immediate `active` prop, not a settled value, since the whole point is for activation to require zero frames.

Why `useSettledActive` was removed rather than kept alongside: it existed only to guarantee an intermediate painted frame so a *transition* could play across it. With `background-color` no longer transitioning while active, there is no transition left for it to guarantee frames for on the activation path, and keeping it would reintroduce the exact starvation risk that caused finding 2. Deactivation was never reported as broken and keeps its transition unmodified, so nothing else needed the settle either.

No new duration/token was introduced; the existing `duration-hover` cadence token and reduced-motion handling in the design package are untouched. The R18 workstream separately owns the underlying ~150ms per-switch main-thread long task; this PR does not attempt to reduce it, only to make the highlight correct regardless of how many frames are available.

## Test evidence

`apps/packages/product-client/src/primitives/patterns/sidebar/SidebarRowSurface.test.tsx`:
- simulates 20+ rapid re-selections landing in the same tick and asserts the row shows `bg-selected` on the very next render (no waiting on a frame)
- asserts an active row's className excludes `background-color` from its transition set
- asserts an inactive row's className includes `background-color` in its transition set

```
 Test Files  4 passed (4)
      Tests  44 passed (44)
 (SidebarRowSurface.test.tsx: 3, MainSidebar.cloud-gating.test.tsx: 2, WorkspaceItem.test.tsx: 15, MainSidebar.test.tsx: 24)
```

Negative control (finding 2): reverted the conditional transition to the unconditional `transition-[background-color,color,opacity]` and reran; the new transition-set test failed as expected:

```
 x SidebarRowSurface > excludes background-color from the transition set while active, so activation paints instantly
   AssertionError: expected '...' to contain 'transition-[color,opacity]'
```

Restored the fix and confirmed all 44 tests green again before pushing.

Negative control (finding 1, from the first commit): reverted the settle indirection so the class read `active` directly; the regression test failed as expected (`expected ... not to contain 'bg-selected'`). Restored and confirmed green.

## Test plan
- [x] `SidebarRowSurface.test.tsx` (3 tests, both findings covered, both negative controls verified)
- [x] `MainSidebar.test.tsx`, `MainSidebar.cloud-gating.test.tsx`, `WorkspaceItem.test.tsx` all green

### DIFF
--- a/apps/packages/product-client/src/primitives/patterns/sidebar/SidebarRowSurface.test.tsx
+++ b/apps/packages/product-client/src/primitives/patterns/sidebar/SidebarRowSurface.test.tsx
@@ -1,42 +1,44 @@
 // @vitest-environment jsdom
 
-import { act, cleanup, render } from "@testing-library/react";
-import { afterEach, describe, expect, it, vi } from "vitest";
+import { cleanup, render } from "@testing-library/react";
+import { afterEach, describe, expect, it } from "vitest";
 import { SidebarRowSurface } from "#product/primitives/patterns/sidebar/SidebarRowSurface";
 
 afterEach(cleanup);
 
-function flushFrame() {
-  act(() => {
-    vi.advanceTimersToNextFrame();
-  });
-}
-
 describe("SidebarRowSurface", () => {
-  it("always settles into bg-selected, even after many rapid re-selections in the same tick", () => {
-    vi.useFakeTimers();
+  it("applies bg-selected on the very next render, even after many rapid re-selections in the same tick", () => {
     const { container, rerender } = render(<SidebarRowSurface active={false}>Row</SidebarRowSurface>);
     const row = () => container.firstElementChild as HTMLElement;
 
     expect(row().className).not.toContain("bg-selected");
 
-    // Simulate many workspace switches landing faster than a paint: active
-    // flips true/false/true repeatedly before any frame is settled.
+    // Simulate a fast sweep: many switches land in quick succession, with no
+    // paint (and thus no animation frame) guaranteed between any of them.
     for (let i = 0; i < 20; i += 1) {
       rerender(<SidebarRowSurface active={i % 2 === 0}>Row</SidebarRowSurface>);
     }
     rerender(<SidebarRowSurface active>Row</SidebarRowSurface>);
 
-    // data-active reflects the true logical state immediately.
+    // The highlight must be visible on this render already: waiting on a
+    // frame (rAF-style deferral) is exactly what breaks under main-thread
+    // saturation, since the deferred frame may never come in time.
     expect(row().getAttribute("data-active")).toBe("true");
-
-    // The visual class hasn't jumped yet: it settles one animation frame
-    // behind so the browser always paints a real, distinct "before" frame.
-    expect(row().className).not.toContain("bg-selected");
-
-    flushFrame();
-
     expect(row().className).toContain("bg-selected");
-    vi.useRealTimers();
+  });
+
+  it("excludes background-color from the transition set while active, so activation paints instantly", () => {
+    const { container } = render(<SidebarRowSurface active>Row</SidebarRowSurface>);
+    const row = container.firstElementChild as HTMLElement;
+
+    expect(row.className).toContain("transition-[color,opacity]");
+    expect(row.className).not.toContain("transition-[background-color,color,opacity]");
+  });
+
+  it("keeps background-color in the transition set while inactive, so deselect/hover still fades", () => {
+    const { container } = render(<SidebarRowSurface active={false}>Row</SidebarRowSurface>);
+    const row = container.firstElementChild as HTMLElement;
+
+    expect(row.className).toContain("transition-[background-color,color,opacity]");
   });
 });
--- a/apps/packages/product-client/src/primitives/patterns/sidebar/SidebarRowSurface.test.tsx
+++ b/apps/packages/product-client/src/primitives/patterns/sidebar/SidebarRowSurface.test.tsx
@@ -1,0 +1,42 @@
+// @vitest-environment jsdom
+
+import { act, cleanup, render } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { SidebarRowSurface } from "#product/primitives/patterns/sidebar/SidebarRowSurface";
+
+afterEach(cleanup);
+
+function flushFrame() {
+  act(() => {
+    vi.advanceTimersToNextFrame();
+  });
+}
+
+describe("SidebarRowSurface", () => {
+  it("always settles into bg-selected, even after many rapid re-selections in the same tick", () => {
+    vi.useFakeTimers();
+    const { container, rerender } = render(<SidebarRowSurface active={false}>Row</SidebarRowSurface>);
+    const row = () => container.firstElementChild as HTMLElement;
+
+    expect(row().className).not.toContain("bg-selected");
+
+    // Simulate many workspace switches landing faster than a paint: active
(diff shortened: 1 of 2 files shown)


pablonyx:
## APPROVE

Adversarial review of both commits (`28480b80f` rAF `useSettledActive` settle → `dbeaf1c58` asymmetric transition, `useSettledActive` removed). Net diff vs `main`: `SidebarRowSurface.tsx` (+20) and new `SidebarRowSurface.test.tsx` (+44).

### Tests (scoped single-file vitest, `--maxWorkers=2`, hands-on)
PR head:
```
 ✓ src/primitives/patterns/sidebar/SidebarRowSurface.test.tsx (3 tests) 26ms
 Test Files  1 passed (1)
      Tests  3 passed (3)
```
Negative control — reverted the asymmetric class to the unconditional `transition-[background-color,color,opacity]`:
```
 FAIL  ... > excludes background-color from the transition set while active, so activation paints instantly
 AssertionError: expected '...' to contain 'transition-[color,opacity]'
 Test Files  1 failed (1)
      Tests  1 failed | 2 passed (3)
```
Restored the fix; re-confirmed 3 passed.

### Hard question: does removing `useSettledActive` resurrect the commit-1 net-zero-flip bug?
No — impossible by construction.
- Commit-1's bug was a background-color *transition* that no-ops when start==end style at the next paint (`true→false→true` nets to `true`). The fix removes background-color from the transition set entirely while active, so activation has **no transition to suppress**: `bg-selected` paints at its computed value on the first available frame, cadence-independent.
- Deactivation re-flip mid-fade (`true→false→true`): landing back on active drops bg from the transition set and snaps solid regardless of the in-flight fade — no start/end-identity dependence.
- Net-zero landing on *inactive* only risks a skipped fade-out, whose correct end state is "no highlight" and is not latency-critical.
- The color/opacity transitions that remain active-side carry no highlight: enabled active and enabled inactive both use `text-sidebar-foreground` (no color delta on selection); `opacity-60` is disabled-only (not part of a sweep).
- `data-active` and `className` now both key off the immediate `active` prop (commit-1 had them out of sync); react runtime imports correctly dropped, no dead imports.

### Other checks
- Labels: exactly one `release:*` (`release:fix`) + `area:product` — correct for a product-client-only change.
- Founder no-visible-text rule: satisfied/N-A — no loading indicator touched, no text node added.
- Title/body accurate to the diff and the two-commit narrative.

pablonyx:
## Manual verification

Head SHA verified: dbeaf1c58bd744d151c31a07d616f6804bb1e941 (matches gh pr view 1974 --json headRefOid, and matches local HEAD after fetch plus fast-forward merge).

### Scoped test run

Diff against origin/main touches exactly one test file: apps/packages/product-client/src/primitives/patterns/sidebar/SidebarRowSurface.test.tsx.

npx vitest run src/primitives/patterns/sidebar/SidebarRowSurface.test.tsx --maxWorkers=2

```
 Test Files  1 passed (1)
      Tests  3 passed (3)
```

### Negative control (deactivation fade complement)

The reviewer already hand-verified the active-side asymmetric class. For the complement, I locally (uncommitted) forced the inactive branch to also exclude background-color from its transition set, collapsing colorTransition to the unconditional "transition-[color,opacity]". A test does pin the deactivation fade path (keeps background-color in the transition set while inactive, so deselect/hover still fades), and it failed exactly as expected:

```
 x SidebarRowSurface > keeps background-color in the transition set while inactive, so deselect/hover still fades
   AssertionError: expected '...' to contain 'transition-[background-color,color,opacity]'
   Received: '...transition-[color,opacity] duration-hover text-sidebar-foreground hover:bg-hover active:bg-active'
```

Reverted the local change and reconfirmed the scoped run back at 3 tests passed (3) before continuing.

### Hand probe

Wrote a temporary standalone jsdom test file (not part of the suite), rendered SidebarRowSurface and flipped active false -> true -> false via rerender, logged the observed className and data-active at each step, then deleted the file. Observed:

```
STEP1 inactive class: ...font-control transition-[background-color,color,opacity] duration-hover text-sidebar-foreground hover:bg-hover active:bg-active
STEP1 data-active: false

STEP2 active class: ...font-control transition-[color,opacity] duration-hover bg-selected text-sidebar-foreground
STEP2 data-active: true

STEP3 inactive-again class: ...font-control transition-[background-color,color,opacity] duration-hover text-sidebar-foreground hover:bg-hover active:bg-active
STEP3 data-active: false
```

data-active tracked the prop immediately on both flips, background-color entered the transition set only while inactive, and the active row carried bg-selected with no background-color transition, on the very next render each time. Worktree is clean afterward (git status confirms no leftover files).

Machine limits observed: kern.memorystatus_vm_pressure_level was 2 before starting (not 3), ran one scoped file at a time with --maxWorkers=2, no cargo, no Docker, no app boot, no full suite.

PR #1975, opened 2026-08-16
feat(product-client): model-derived follow target + consumed-inset transitions (PRO-187, r7)

pablonyx:
## Rung 7: model-derived follow target + consumed-inset transitions (PRO-187, Q6)

Implements exactly rung 7 of the Chat Scroll ADR ladder (§6) and the Q6 consequence default in Founder Rulings: make the follow target a single named derivation with the dock inset as a declared input, and encode the consumed-inset semantics as state-machine transitions rather than scattered refs. Stacked on r6 (`chat-scroll/r6-fr2-restore`).

### Design vs ADR

- **Follow target is one named derivation** — `transcript-follow-target.ts` (`resolveTranscriptFollowTarget`) replaces the ad-hoc `resolveAutoFollowScrollTop`. It takes the dock-inset model split (`{ structuralInsetPx, nonDisplacingInsetPx }`, produced by the existing `resolveTranscriptBottomInsets`) as a declared named input instead of a positional overlay number. Structural inset is documented as already living in `scrollHeight` (virtualizer `paddingEnd`); declaring it keeps this the ONE place that reasons about the dock inset, so no caller double-compensates.
- **Consumed-inset semantics are a state machine** — `transcript-consumed-inset.ts` (`reduceTranscriptInset`) with named transitions: `submit_repin` (no-op, submit does NOT consume), `consume_full` (button click / bottom-band re-pin), `leave_band` (reset), `reset` (session switch), and `dock_inset_changed` (composer growth/collapse, status bar). The scattered `consumedAutoFollowBottomInsetRef` writes across `use-transcript-auto-follow-bottom.ts` and `use-transcript-stick-to-bottom.ts` now route through it.
- **Consumed-inset transitions when a displacing inset appears/disappears** — the machine flags `structuralShrinkClamp` when the structural inset shrinks. The dock-geometry layout effect marks the upward clamp a shrink queues (while pinned) so it is classified as our own write, not a user upward scroll. This closes a concrete as-built gap: the engine previously received ONLY the non-displacing half, so a composer collapse's clamp landed at the hard bottom and hit the `consumeInset: "full"` pin path, wrongly consuming the manual overlay. Unpinned readers get no scrollTop write on any inset change (the effect is guarded to pinned), so they are never displaced.

### No ADR contradiction

The as-built already routed the non-displacing overlay half through the engine and split it via `resolveTranscriptBottomInsets`; rung 7 completes the model derivation by also routing the structural half and centralising the transitions. No scope was re-derived and no rung-3/4/5/6 behaviour changed. `revert to the ref-based bookkeeping` per the ADR revert note is a clean revert of these commits.

### Files

- `hooks/chat/ui/transcript-follow-target.ts` (new; replaces `transcript-auto-follow-target.ts`)
- `hooks/chat/ui/transcript-consumed-inset.ts` (new)
- `hooks/chat/ui/use-transcript-auto-follow-bottom.ts` (state machine + follow-target wiring)
- `hooks/chat/ui/use-transcript-stick-to-bottom.ts`, `use-transcript-submit-stamp-repin.ts` (transitions)
- `components/workspace/chat/transcript/{FullTranscriptRowList,VirtualizedTranscriptRowList}.tsx` (pass structural inset)
- tests: `transcript-follow-target.test.ts`, `transcript-consumed-inset.test.ts`, `use-transcript-stick-to-bottom.inset-transition.test.tsx`
- physics: two scenarios + drivable dock inset in the scroll-physics fixture

### Testing

- `python3 scripts/report_frontend_structure.py` → `TOTAL: 0` (all sources <400, tests <600).
- Targeted vitest: 58 pass across the stick-to-bottom family + the two new pure-module suites, with negative controls (submit-does-not-consume; structural-grow-does-not-flag; a structural clamp reclassification is caught).
- Negative control verified by reversion: forcing `structuralShrinkClamp: false` fails the consumed-inset flag test AND the engine "marks the upward clamp" test; reverting restores green.
- Physics (real Chromium + real WebKit, workers=1, repeat-each=2, `pnpm shared:build` first): both new scenarios green on both engines —
  - `displacing-inset transition while pinned: composer collapse/grow follows in one cut, no fight` (single-cut trace, stays pinned, stream still follows after the transition = negative-control bite).
  - `displacing-inset transition while unpinned reading: no displacement` (top-visible row + scrollTop hold across collapse and grow).

Local physics honoured the pressure gate spirit (a stray non-owned 1.9 GB WebKit process held `kern.memorystatus_vm_pressure_level` at warn=2; ran one browser project at a time, never reached critical). Full-suite physics + full vitest are left to CI.

### Observability

No new telemetry (rung 11 owns diagnostics). Behaviour changes are covered by the physics tier and the JSDOM unit tests.


pablonyx:

Branch chat-scroll/r7-follow-target checked out, product-client dependencies built via pnpm shared:build. Ran each touched test file individually with pnpm vitest run <file> --reporter=basic:

- src/hooks/chat/ui/transcript-consumed-inset.test.ts: Test Files 1 passed (1), Tests 10 passed (10)
- src/hooks/chat/ui/transcript-follow-target.test.ts: Test Files 1 passed (1), Tests 6 passed (6)
- src/hooks/chat/ui/use-transcript-stick-to-bottom.inset-transition.test.tsx: Test Files 1 passed (1), Tests 3 passed (3)
- src/hooks/chat/ui/use-transcript-stick-to-bottom.test.tsx: Test Files 1 passed (1), Tests 23 passed (23)

Negative control (model-derived follow target, double-compensation guard): in transcript-follow-target.ts, resolveTranscriptFollowTarget's no-remaining-overlay branch (`if (remainingManualInsetPx <= 0) return scrollHeight;`) was changed to wrongly subtract the structural inset a second time even though it already lives in scrollHeight (`return Math.max(0, scrollHeight - dockInset.structuralInsetPx)`). Reran transcript-follow-target.test.ts. Literal failure:

```
❯ src/hooks/chat/ui/transcript-follow-target.test.ts (6 tests | 3 failed)
  × consuming the whole overlay follows to the raw scrollHeight
    → expected 3880 to be 4000
  × NEGATIVE CONTROL: a growing structural inset does NOT subtract from the pinned target (double-compensation guard)
    → expected 3600 to be 3960
  (plus the base no-overlay case)
Test Files  1 failed (1)
     Tests  3 failed | 3 passed (6)
```

Restored the original `return scrollHeight;` branch via Edit; `git status` confirms working tree clean, no diff remains.

Root-cause note on the earlier 9-lane failure tally: that tally was a stale-run artifact. The ruling-1 cascade force-push cancelled the prior head's in-flight runs, and cancellations tally as failures; no lane failed on the current head.

pablonyx:
## Review round 1: CLEAN

Reviewed the delta `845a147c2..331aa72eb` from git objects, using an isolated worktree (`/tmp/r7-review-wt`, since another agent is live-editing rung 8 on this checkout). No merge, no push.

**Attacked and cleared:**

1. **FR-1 (single writer).** `dispatchInsetEvent` (transcript-consumed-inset.ts) is a pure reducer — every branch (`reset`/`leave_band`/`consume_full`/`submit_repin`/`dock_inset_changed`) returns `{state, structuralShrinkClamp}` with zero DOM/scrollTop writes. All actual `scrollTop` writes still funnel through the single frame pipeline / `scrollToBottom`. `submit_repin` is an explicit documented no-op (Q6 ruling), verified by `transcript-consumed-inset.test.ts:42` which asserts `state` is unchanged — I tried disabling that case label as a negative control and it fell through to the reducer's `default` no-op with identical behavior, confirming there's no live path that lets a submit consume the overlay.

2. **Follow-target math.** `resolveTranscriptFollowTarget` (transcript-follow-target.ts) correctly treats `scrollHeight` as already including the structural inset (virtualizer `paddingEnd`) and only subtracts the *remaining* non-displacing inset. Negative control: patched the return to skip the subtraction (`return hardBottom` unconditionally) — `transcript-follow-target.test.ts` immediately failed 2/6 with a concrete expected-vs-actual (3250 vs 3400), confirming the assertions are real, not vacuous.

3. **Ruling-1 cancel / deadline-release interaction.** The dock-inset layout effect in `use-transcript-auto-follow-bottom.ts` gates its `markNonUserScrollPosition` call on `pinnedRef.current` and on either `structuralShrinkClamp` or a shrinking overlay with previously-consumed range — both effectively unpinned-safe (matches the qualification test's "unpinned: no displacement" negative-control claim, which documents that dropping the pinned-only guard would move the reading row). No second write path introduced alongside the existing ABOVE-change-compensation / final-corrective-write plumbing (those files are untouched by this diff besides comment cleanup — confirmed via `git diff --stat`, no logic hunks).

4. **FR-2 restore interaction.** `resetForSession` dispatches `{type: "reset"}` *before* `beginSessionRestorePlacement` runs, and the restore resolver / deadline machinery (in the untouched frame-pipeline-lifecycle file) is prioritized independently of the follow-target/consumed-inset state the new code owns. No code path lets `resolveTranscriptFollowTarget`/`dispatchInsetEvent` override an in-flight restore — they're simply not consulted while `restoreResolverRef` is set.

5. **Test honesty.** 4/4 new/changed unit test files pass after a real package build (42 tests: `transcript-follow-target.test.ts`, `transcript-consumed-inset.test.ts`, `use-transcript-stick-to-bottom.test.tsx`, `use-transcript-stick-to-bottom.inset-transition.test.tsx`). Two independent negative controls (above) both bit as expected. No `.only`/`.skip`/`eslint-disable`/fixme additions anywhere in the diff (`git diff | grep -i fixme|skip(|\.only(|eslint-disable` — empty). Playwright physics-scenario additions read as real (drivable `setComposerInset`/`setOverlayInset`, explicit negative-control comments naming the exact regression each assertion would catch) but were not run locally per instruction (no local playwright on this machine) — left to CI.

6. **Repo shape.** `use-transcript-stick-to-bottom.ts` sits at exactly 399 lines (cap is 400, commit message says "hold under 400" — accurate, no allowlist entry added for it or any touched file; grepped `scripts/check_max_lines.py` config for these paths, none present). `pnpm run build` (tsc across the full package + subpackages) completed clean, zero errors. Naming (`structuralBottomInsetPx` / `nonDisplacingBottomInsetPx` replacing the old single `autoFollowBottomInsetPx`) is threaded consistently through `FullTranscriptRowList.tsx` and `VirtualizedTranscriptRowList.tsx`, both still deriving the split via the unchanged `resolveTranscriptBottomInsets`.

No blocking or non-blocking findings. Nothing here fights the rung-7 physics as specified in ADR section 6 rung 7 / Q6.

pablonyx:
Stack refold receipt (overnight staging): rebased onto main @ d5d38a804 (r1 merged) atop the retargeted stack. git range-diff vs previous head: 0 semantic changes (3/3 commits identical). New head d3f2cb6b1. Review carried from 17cf143a9.

pablonyx:
Refold train receipt (r7): d3f2cb6b1 -> 544c6e667, rebased --onto origin/main from parent old head 7d084ee08 after r6 squash. range-diff old vs new: 3/3 = (byte-identical). Fresh CI on this head is the merge gate.

pablonyx:
r7 CI note: first round on 544c6e667 settled 21 pass / 3 fail. All three fails are runner-infra, not code: tier-2 lifecycle died in 'Install Playwright chromium' (apt azure mirror unreachable for 30m, job canceled before any test ran), Terraform failed at 'init (hosted Redis)' registry fetch in 21s, Shared frontend packages hit a Lexical jsdom composer test untouched by this diff. Containment check on merge base d946741e1: TRUE (contains 337114fc8). Login budget PASSED this round. Failed jobs rerun issued; same-signature repeat = stop.

pablonyx:
r7 fresh-merge-ref round (reopen 03:14Z, base d2976e342) settled 21 pass / 3 fail. All three fails classified by step receipt via jobs API: 'Install Playwright chromium -> cancelled' on login-budget, production-compose-smoke, and intent-tests — zero tests executed in each; known repo-wide runner-infra stall window, not content-attributable. Same-ref rerun of failed jobs issued on all three runs. Merge gate remains: literal green tally (budget within ruled band if red), tier-2 green (already passed this round), containment TRUE.

pablonyx:
CORRECTION to previous receipt: the 03:14Z round did NOT test against d2976e342. Job logs show checkout 'Merge 544c6e667 into d946741e1' (merge commit 52e03cb, same as the pre-reopen round) — the 20s close/reopen hit the known stale merge-ref race, so the 25/25 green measured the pre-bgwork base. Containment invariant (337114fc8 in d946741e1) still TRUE, but re-cycling close/reopen with a ~4 min gap to mint a genuinely fresh merge-ref against current main before merging. Base will be verified from the job checkout line this time.

pablonyx:
r7 MERGE RECEIPT: squash 577eeaddf, head 544c6e667 (match-head-commit enforced). Final round: 25/25 pass on true-fresh merge-ref 54c714d71 (base 2aac742d9 = main tip at test time; contains 337114fc8: TRUE). Login budget PASSED at JS 501977/502000 B (23 B headroom) — no attribution needed this rung. Three infra install-stall rounds disclosed above; final verdict from executed tests only. Refold provenance: d3f2cb6b1 -> 544c6e667, range-diff 3/3 =.

PR #1976, opened 2026-08-16
perf(product): two-phase workspace switch, preview cursor and query-churn trim

pablonyx:
## R19: Two-phase workspace switch (UX Latency + Transitions ladder)

Held `Cmd+Opt+Arrow` traversal used to commit a full workspace selection on every keydown. A single commit is ~150-250ms of main-thread work (query observer teardown/recreate + GC across the wide shell tree + pane mount), so held-key repeats queued and replayed after the key was released, and the highlight lagged the keyboard. This PR productionizes the interaction Pablo live-validated, as three legs.

### Leg 1: preview-cursor traversal (implemented)

Held traversal now moves a lightweight preview cursor through the sidebar instead of committing per keydown:

- New `sidebar-switch-cursor-store` (zustand): a single `cursorId`. While set, it drives the row highlight instead of the committed selection.
- `WorkspaceItem` folds `cursorId` and its committed `active` prop into one `displayedActive` boolean via a per-row selector, so a cursor step re-renders exactly the two rows whose displayed state flips.
- The step/settle/coalescing state machine lives in a React-free controller (`workspace-switch-cursor-controller.ts`) with injected timers and store accessors:
  - a step is throttled to `WORKSPACE_CURSOR_STEP_MIN_MS` (60ms); repeats above the throttle are **dropped, never queued** (releasing the key never replays a backlog),
  - the real selection commits **once**, `WORKSPACE_CURSOR_SETTLE_MS` (180ms) after the last step,
  - the cursor clears only when the committed selection reflects the target, with a `WORKSPACE_CURSOR_COMMIT_FALLBACK_MS` (2000ms) fallback for a failed/superseded commit.
- Shim edges fixed properly:
  - **(a) mouse click during pre-settle wins**: an external committed-selection change while previewing cancels the pending commit and clears the cursor, so the click stands.
  - **(b) target list changes mid-traversal**: the cursor is validated against the live target list at commit time; a vanished cursor is dropped, never committed as a dead id.
  - **(c) Escape** cancels an uncommitted cursor (capture-phase, no `preventDefault`, only acts while a cursor is pending).

### Leg 2: shell query-churn stabilization (implemented)

Audit of shell-tree query call sites whose keys/options are rebuilt per render (ranked):

| # | Site | Finding | Action |
|---|------|---------|--------|
| 1 | `lib/infra/proliferate-api.ts` `getProliferateApiOrigin` | `new URL(baseUrl).origin` re-parsed on every render at every query-hook call site (the base URL is effectively constant) | **Fixed**: single-slot memo keyed by `baseUrl`, so the URL is parsed once per deployment |
| 2 | `hooks/workspaces/cache/use-repo-pr-statuses.ts` | `queries: ids.map(...)` and `combine` rebuilt every render, forcing per-repo `observer.setOptions` churn and combine recompute on every shell re-render | **Fixed**: `queries` wrapped in `useMemo([ids, runtimeUrl, cacheScopeKey])`; `combine` wrapped in `useCallback([ids])` |
| 3 | `hooks/workspaces/cache/use-workspaces.ts` | `queryKey` already memoized via `useWorkspaceCollectionsCache`; `queryFn`/`refetchInterval` recreated per render but react-query only stores the latest fn (no observer teardown) | Not a top offender; left as-is |
| 4 | `hooks/workspaces/cache/tabs/use-workspace-header-subagent-hierarchy.ts` | builds several `useQueries` arrays inline, but session ids and relationship hints are already `useMemo`'d and it is header-scoped | Not chased (already heavily memoized) |
| 5 | `hooks/workspaces/facade/use-worktree-settings-targets.ts` | `queries: readyCloudWorkspaces.map(...)` over a memoized list, array rebuilt inline | Minor; not chased |
| 6 | `hooks/workspaces/cache/use-archived-workspaces.ts` | single `useQuery`, stable key, mounted only in the archived view (not the hot switch path) | Not chased |

Per the rung scope, the origin hoist plus the worst offender (#2) are fixed; the rest are documented, not chased.

### Leg 3: two-phase pane deferral (NOT added, recorded, see below)

Leg 1 coalesces a held sweep to a **single** commit, so the intermediate-switch pane mounts that the shim's `useSettledPaneKey` guarded against no longer occur; the rapid-sweep motivation for pane deferral is moot. For a lone deliberate switch the shim mounts the pane immediately anyway (its rapid-switch window never triggers), so it would add no benefit there either. The residual single-commit cost is a separate long task addressed by leg 2.

I could not run a live renderer profiler to demonstrate a residual >100ms long task on this run (no dev server / Chrome profiling available under the machine's memory pressure), and the brief is explicit: if legs 1-2 make it moot, record that instead of adding dead machinery. I therefore did **not** add `useSettledPaneKey`. Flagging this as a scope judgment rather than silently rescoping.

### Test-infra note

The injected Desktop measurement engine imports product-client's diagnostics port through the public `@proliferate/product-client/internal/*` subpath, which `vitest.config.ts` did not source-alias (only `/host/*` was), so on a no-build checkout **every** product-client test failed to resolve it. Added the `/internal/` source alias consistent with the existing `/host/` precedent and the config's stated "tests run against source, never dist" rule. CI builds the package so this was green there; the alias unblocks no-build local/review runs.

### Tests

- `workspace-switch-cursor-controller.test.ts` (13): step/preview, throttle-drop, settle re-arm across a held burst, commit reflection, fallback clear, click-during-pre-settle cancel, post-commit override, vanished-target no-commit, cursor-equals-committed no-commit, Escape cancel, backward wrap, plus a **negative control** (commit-per-step without coalescing commits on every repeat).
- `sidebar-switch-cursor-store.test.ts` (2): set/clear, identity short-circuit.
- `use-app-shortcuts.test.tsx` (+2): shortcut previews the cursor and commits once after the settle; Escape cancels without committing.

Negative control also verified by mutation: neutering the throttle makes the throttle-drop and negative-control tests fail; restored.

### DIFF
--- a/apps/packages/product-client/src/components/workspace/shell/sidebar/WorkspaceItem.tsx
+++ b/apps/packages/product-client/src/components/workspace/shell/sidebar/WorkspaceItem.tsx
@@ -18,6 +18,7 @@ import { resolveWorkspaceItemTrailingCells } from "#product/components/workspace
 import { useWorkspacePeek } from "#product/components/workspace/shell/sidebar/WorkspacePeekCard";
 import { WorkspaceRenamePopover } from "#product/components/workspace/shell/sidebar/WorkspaceRenamePopover";
 import { ProductSidebarWorkspaceRow } from "#product/components/workspace/shell/sidebar/ProductSidebarRepositories";
+import { useSidebarSwitchCursorStore } from "#product/stores/workspaces/sidebar-switch-cursor-store";
 
 interface WorkspaceItemProps {
   workspaceId?: string;
@@ -116,6 +117,14 @@ export function WorkspaceItem({
   onCopyBranchName,
   onRename,
 }: WorkspaceItemProps) {
+  // While a held-key traversal cursor is set the cursor position, not the
+  // committed selection, drives the highlight so exactly one row reads active
+  // during movement. The selector folds this row's id and its committed
+  // `active` prop into a single boolean, so a cursor step re-renders only the
+  // two rows whose displayed state flips rather than every subscribed row.
+  const displayedActive = useSidebarSwitchCursorStore((state) =>
+    state.cursorId === null ? active : state.cursorId === workspaceId,
+  );
   const [renameOpen, setRenameOpen] = useState(false);
   const [doneConfirmOpen, setDoneConfirmOpen] = useState(false);
   const handleRenameCommand = () => setRenameOpen(true);
@@ -201,7 +210,7 @@ export function WorkspaceItem({
 
   const row = (
     <ProductSidebarWorkspaceRow
-      active={active}
+      active={displayedActive}
       archived={archived}
       trailingStatus={trailingStatus}
       trailingIdentity={trailingIdentity}
--- a/apps/packages/product-client/src/hooks/app/lifecycle/use-app-shortcuts.test.tsx
+++ b/apps/packages/product-client/src/hooks/app/lifecycle/use-app-shortcuts.test.tsx
@@ -14,6 +14,8 @@ import {
 import { USER_PREFERENCE_DEFAULTS } from "#product/lib/domain/preferences/user/model";
 import { useUserPreferencesStore } from "#product/stores/preferences/user-preferences-store";
 import { requestRightPanelTabByIndex } from "#product/lib/workflows/workspaces/right-panel-shortcut-requests";
+import { useSidebarSwitchCursorStore } from "#product/stores/workspaces/sidebar-switch-cursor-store";
+import { WORKSPACE_CURSOR_SETTLE_MS } from "#product/lib/domain/workspaces/sidebar/workspace-switch-cursor-controller";
 
 const navigationMocks = vi.hoisted(() => ({
   selectWorkspaceFromSurface: vi.fn(),
@@ -35,18 +37,24 @@ vi.mock("#product/hooks/workspaces/workflows/use-workspace-navigation-workflow",
   }),
 }));
 
-vi.mock("#product/stores/sessions/session-selection-store", () => ({
-  useSessionSelectionStore: (
+vi.mock("#product/stores/sessions/session-selection-store", () => {
+  const readSelection = () => ({
+    selectedWorkspaceId: harnessState.selectedWorkspaceId,
+    selectedLogicalWorkspaceId: harnessState.selectedLogicalWorkspaceId,
+  });
+  const useSessionSelectionStore = (
     selector: (state: {
       selectedWorkspaceId: string | null;
       selectedLogicalWorkspaceId: string | null;
     }) => unknown,
-  ) =>
-    selector({
-      selectedWorkspaceId: harnessState.selectedWorkspaceId,
-      selectedLogicalWorkspaceId: harnessState.selectedLogicalWorkspaceId,
-    }),
-}));
+  ) => selector(readSelection());
+  // The traversal-cursor effect reads the committed selection imperatively and
+  // subscribes for commit reflection, so the mock exposes the same static
+  // surface the real zustand store does.
+  useSessionSelectionStore.getState = readSelection;
+  useSessionSelectionStore.subscribe = () => () => {};
+  return { useSessionSelectionStore };
+});
 
 vi.mock("#product/lib/workflows/workspaces/right-panel-shortcut-requests", () => ({
   requestRightPanelTabByIndex: vi.fn(() => true),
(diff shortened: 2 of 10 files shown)


pablonyx:
## Manual verification

Head SHA: `613adb0817e383733f3eccfe45657d695f3da727`

Verified by hands-on probing, not CI badges. Machine limits observed: scoped single-file vitest only (`--maxWorkers=2`), one file at a time, memory pressure checked before each run (stayed at level 1).

### Scoped vitest, per touched test file (literal tallies)

Authored / modified tests:

- `src/lib/domain/workspaces/sidebar/workspace-switch-cursor-controller.test.ts`
  `Test Files  1 passed (1)` / `Tests  13 passed (13)`
- `src/stores/workspaces/sidebar-switch-cursor-store.test.ts`
  `Test Files  1 passed (1)` / `Tests  2 passed (2)`
- `src/hooks/app/lifecycle/use-app-shortcuts.test.tsx`
  `Test Files  1 passed (1)` / `Tests  14 passed (14)`

Existing tests exercising the changed source:

- `src/components/workspace/shell/sidebar/WorkspaceItem.test.tsx`
  `Test Files  1 passed (1)` / `Tests  15 passed (15)`
- `src/hooks/workspaces/cache/use-repo-pr-statuses.test.tsx`
  `Test Files  1 passed (1)` / `Tests  6 passed (6)`

### Negative control (revert the fix, watch it fail, restore)

Mutation: replaced the throttle guard `if (now - lastStepAt < WORKSPACE_CURSOR_STEP_MIN_MS)` with `if (false)` in `workspace-switch-cursor-controller.ts`, so repeats are no longer dropped.

Rerun of the controller test with the fix reverted:

```
 × workspace-switch-cursor-controller > drops key repeats inside the throttle window instead of queueing them
 × workspace-switch-cursor-controller > does not re-commit when the cursor settles back on the committed row
AssertionError: expected 'd' to be 'b' // Object.is equality
      Tests  2 failed | 11 passed (13)
```

The literal `expected 'd' to be 'b'` is the point: without the throttle, a rapid burst walks the cursor to `d` instead of stopping one step at `b`, i.e. repeats get processed rather than dropped. Restored via `git checkout --`; `git status --short` is empty (clean).

A second negative control is baked into the suite as a passing assertion: a naive commit-per-step stepper commits `["b","c","d"]` (one commit per keydown, the pre-fix stall) while the coalescing controller, given the identical burst, commits exactly once.

### Hand probe beyond the suite

Drove the REAL `sidebar-switch-cursor-store` through the controller with a fake clock (throwaway probe test, since deleted, worktree clean) and printed the observed step / throttle / settle / commit / reflect sequence. Targets `[ws-a, ws-b, ws-c, ws-d]`, committed starts at `ws-a`:

```
start committed=ws-a cursor=null
after step#1 (t=0)  cursor=ws-b commits=[]
after fast repeat (t=40, throttled) cursor=ws-b commits=[]
after step#2 (t=101)  cursor=ws-c commits=[]
1ms before settle (t=280) cursor=ws-c commits=[]
at settle (t=281)       cursor=ws-c commits=["ws-c"]
after committed reflected=ws-c cursor=null commits=["ws-c"]
```

Observed, in order: the first step previews `ws-b` with no commit; a repeat 40ms later (inside the 60ms throttle) is dropped, cursor stays `ws-b`; a step at 101ms advances to `ws-c`; nothing commits at 1ms before the 180ms settle; the single selection commit (`ws-c`) fires exactly at the settle; and once the committed selection reflects the target the preview cursor clears to `null`. That is the whole two-phase contract exercised end to end on the real store.

### Typecheck

`pnpm typecheck` exit 0, zero `error TS` (run earlier on this SHA).

pablonyx:
## Manual verification (addendum, new head)

Refreshes the earlier verification for the one review-delta commit. Bound to head `79e8b7fbe7dff3ad86c579ac7894c6f816774a7c`.

### Delta commit scope

`test(product): pin throttle cadence to literal ms and drop em dash`, test code only, zero production change:

1. Removed the em dash from the settle-back test comment (standing rule: no em dashes in authored comment text).
2. The throttle-drop and settle-back tests derived their fake-clock advances from the exported step-min constant, so zeroing that constant sent `advance()` negative and coincidentally preserved the drop, letting a constant-zeroing mutant survive. Advances are now literal milliseconds (59 inside the 60ms window, 60 at the boundary, 180 past the settle) so the tests pin the designed cadence.

### Scoped vitest at the new head (literal)

`src/lib/domain/workspaces/sidebar/workspace-switch-cursor-controller.test.ts` (`--maxWorkers=2`):

`Test Files  1 passed (1)` / `Tests  13 passed (13)`

### Constant-mutant sanity check (the point of the delta)

Zeroed the exported `WORKSPACE_CURSOR_STEP_MIN_MS` (uncommitted production mutation) and reran the throttle-drop test:

```
 × workspace-switch-cursor-controller > drops key repeats inside the throttle window instead of queueing them
AssertionError: expected 'd' to be 'b' // Object.is equality
      Tests  1 failed | 12 skipped (13)
```

With literal advances the mutant now dies (`expected 'd' to be 'b'`: without the throttle the burst walks the cursor to `d` instead of holding at `b`), where before the delta it survived. Production restored via `git checkout --`; the only working-tree change was the committed test file.

CI at this head: 23 pass / 0 fail (all checks green).

PR #1980, opened 2026-08-16
fix(product-client): transient-block invariant + Q13 harness matrix, FR-2 restore stranding fallback (PRO-187, r10)

pablonyx:
## Rung 10: transient-block invariant and thinking live-matrix (PRO-187, Q13)

Implements rung 10 of the Chat Scroll ADR ladder (§6) and Q13 in Founder
Rulings, stacked on r9 (`chat-scroll/r9-scroll-to-latest`), plus Founder
Ruling 3 (rung 6 FR-2 revisit, the never-mounting-saved-row strand and the
prepend blank-fallback grace-window supersession).

### Q13: the reserved-slot invariant, extended

Tracing Cell 6 (TranscriptTurnChrome.tsx) before writing anything: the
invariant is already generalized structurally. Every trailing-status variant
(thinking/working gleam, needs-input marker, transient tool label) shares the
same fixed `ASSISTANT_ACTION_SLOT_HEIGHT` slot via `TrailingStatusCrossfade`,
which reconciles in place rather than mounting/unmounting a new element. That
did not need to change.

What rung 10 actually needed:

- **Static per-harness mutation-class matrix** — `domain/chats/transcript/harness-transient-block-matrix.ts`.
  The invariant is enforced at ONE normalized layer: the `@anyharness/sdk`
  `TranscriptState` item kinds (`thought`, `tool_call`, `assistant_prose`)
  every harness adapter reduces its wire protocol into. So the real
  per-harness question is: does a harness's adapter ever normalize a mid-turn
  event into a FOURTH, unclassified visible item kind the chrome layer
  doesn't fold into the reserved slot? Today, no (`claude`, `codex`,
  `opencode`, `grok` all normalize reasoning → `thought` and tool use →
  `tool_call`). The module doc comment carries the written live-measurement
  plan: the exact events (thought start/delta/stop, tool start/stop, prose
  interleave) an agent captures against a REAL streaming session per harness
  to confirm this classification, and what a disagreement there means (either
  the adapter started emitting a new kind, or the chrome stopped folding an
  existing one).
- **Physics fixture** (`qualification/scroll-physics`): `streamThoughtStart`
  / `streamThoughtStop` / `streamToolCall` driver methods emit a `reasoning`
  item (start → delta → stop) and a small completed tool call inside an open
  streaming turn; the new spec
  `Q13 thinking/tool interleave: transient block lifecycle never displaces
  the pinned reader` asserts `bottomDistance` stays bounded and the scrollTop
  trace never shows a forward-jump-then-correction (double-scroll) across
  thought start → thought stop → tool call → prose resume, all inside one
  turn, while pinned.

### Founder Ruling 3(a)/(b): the never-mounting-saved-row strand

Attempted construction: **it IS constructible.** `resolveTranscriptRestoreTargetTop`
already has two paths — an estimate-immune DOM-rect resolution when the saved
row's element is mounted, and a coarse index-sum estimate fallback when it
isn't. Nothing in the resolver's contract prevents the coarse path from being
the ONLY one taken for an entire restore window (e.g. the saved row's index
sits permanently outside the virtualizer's render window). Before this PR,
`use-transcript-frame-pipeline-lifecycle.ts` cleared the restore resolver at
the deadline unconditionally — leaving scrollTop frozen at whatever coarse,
unproven estimate was last written, with no correctness guarantee.

Fix (Ruling 3(b)): `resolveTargetTop` now returns `{ top, mounted }` instead
of a bare number. The frame-pipeline lifecycle tracks whether the CURRENT
restore resolver ever resolved through the `mounted: true` path; if the
deadline lapses having never mounted, it calls the new `onRestoreStranded`
callback, which the engine wires to bottom-pin — the same conservative
default a vanished saved row (`resolveTargetTop` returning `null`) already
gets. `use-transcript-stick-to-bottom.restore-stranded.test.tsx` constructs
the strand directly (a resolver returning `mounted: false` for the whole
window) and proves the bottom-pin fallback, plus a negative control proving a
single successful mount before the deadline suppresses it.

### Founder Ruling 3(c): grace-window supersession

`PREPEND_BLANK_FALLBACK_GRACE_MS` was an INDEPENDENT fixed 3s timer, armed the
instant a prepend fired regardless of whether the engine's own above-change
compensation anchor was still live. It now subordinates to the real signal:
`use-transcript-stick-to-bottom.ts` exposes `compensationDeadlineRef`
(the same deadline `startAboveChangeCompensation` just armed), and
`VirtualizedTranscriptRowList.tsx` sets `prependSettleUntilRef` from THAT
value directly. The fixed constant survives only as a bounded ceiling
fallback for the (should-not-happen) case where no compensation deadline got
set. The existing `prepend anchoring: older-history prepend keeps the
reading row fixed` physics scenario (CPU-throttled Chromium arm, the r2/r5
webkit-remount regression fixture) is unchanged and still asserts the r2 fix
holds — it now runs against the subordinated signal instead of the old timer.

### Deviations

None from ADR scope. The reserved-slot invariant needed no structural code
change (it was already correct); rung 10's real work was the static matrix,
the physics coverage, and Ruling 3.

### Files

- `src/domain/chats/transcript/harness-transient-block-matrix.ts` (+ test) —
  the static Q13 classification + written live-measurement plan.
- `src/hooks/chat/ui/transcript-reading-position-store.ts` (+ test) —
  `resolveTargetTop` returns `{ top, mounted }`.
- `src/hooks/chat/ui/use-transcript-frame-pipeline-lifecycle.ts` — tracks
  ever-mounted per restore resolver; `onRestoreStranded` callback.
- `src/hooks/chat/ui/use-transcript-stick-to-bottom.ts` (+ types) — wires
  `onRestoreStranded` to bottom-pin; exposes `compensationDeadlineRef`.
- `src/hooks/chat/ui/use-transcript-stick-to-bottom.restore-stranded.test.tsx`
  (new) — constructs the strand + negative control.
- `src/hooks/chat/ui/use-transcript-virtualizer-blank-fallback.ts` — comment
  update on `PREPEND_BLANK_FALLBACK_GRACE_MS`'s now-subordinate role.
- `src/components/workspace/chat/transcript/VirtualizedTranscriptRowList.tsx`
  — sets `prependSettleUntilRef` from `compensationDeadlineRef`.
- `qualification/scroll-physics/src/scroll-physics-host.ts` — `thoughtStarted`
  / `thoughtDelta` / `thoughtCompleted` fixture authoring + `streamThoughtStart`
  / `streamThoughtStop` / `streamToolCall` driver methods.
- `qualification/scroll-physics/specs/scroll-physics.spec.ts` — the Q13
  interleave scenario.

### Testing

- `python3 scripts/report_frontend_structure.py` → `TOTAL: 0`.
- `tsc -p tsconfig.json --noEmit` (product-client) and
  `tsc -p qualification/scroll-physics/tsconfig.json --noEmit` (fixture) →
  both EXIT 0.
- Scoped vitest, 11 files / 84 tests, all pass — literal output posted as a
  Manual verification PR comment.
- Negative control: reverted the `everMounted` gate in
  `use-transcript-frame-pipeline-lifecycle.ts` (bottom-pin unconditionally at
  the deadline). The negative-control test failed with the literal
  `expected true to be false` (isPinnedToBottom flipped true even though the
  row mounted); restored, suite green again. Full before/after in the Manual
  verification comment.
- Local Playwright skipped: `vm.swapusage` showed 716MB free of 9GB at the
  time (tight, not the near-idle state the pressure gate wants) even though
  `kern.memorystatus_vm_pressure_level` read 1; defers to CI (real Chromium +
  WebKit) as the tier of truth per the repo's practice on this ladder.

### Observability

No new telemetry (rung 11 owns diagnostics).

### DIFF
--- a/apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts
+++ b/apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts
@@ -15,18 +15,33 @@ const REPIN_BAND_PX = 24;
 // bound is the "still following" ceiling: comfortably above that resting gap,
 // far below the unbounded growth a lost follow would produce.
 const PIN_FOLLOW_MAX_DISTANCE_PX = 120;
-// Frame-to-frame backward-bounce ceiling for the Q13 interleave trace below.
-// This is a DIFFERENT quantity from PIN_FOLLOW_MAX_DISTANCE_PX (steady-state
-// bottomDistance while pinned): it bounds how far scrollTop may legitimately
-// snap backward between two consecutive trace samples when a newly-mounted
-// row's estimate (transcript-row-height-estimate.ts) corrects down once the
-// row is measured for real. The reserved live-turn slot
-// (ASSISTANT_ACTION_SLOT_HEIGHT, TranscriptTurnChrome.tsx, h-6 = 24px) is the
-// invariant Q13 exists to police, so this ceiling must stay strictly below
-// 24px: a phantom reserved-slot height briefly appearing and collapsing
-// would otherwise pass silently through a bound reused from an unrelated
-// quantity (see PR #1980 review finding 1).
+// Frame-to-frame backward-bounce ceiling for the Q13 interleave trace below,
+// for phases where NO new row mounts (thought start/delta/stop, prose
+// growth on an already-mounted row). This is a DIFFERENT quantity from
+// PIN_FOLLOW_MAX_DISTANCE_PX (steady-state bottomDistance while pinned): it
+// bounds how far scrollTop may snap backward between two consecutive trace
+// samples with no legitimate cause to move backward at all. The reserved
+// live-turn slot (ASSISTANT_ACTION_SLOT_HEIGHT, TranscriptTurnChrome.tsx,
+// h-6 = 24px) is the invariant Q13 exists to police, so this ceiling stays
+// strictly below 24px: a phantom reserved-slot height briefly appearing and
+// collapsing during a lifecycle-only transition would otherwise pass
+// silently through a bound reused from an unrelated quantity (see PR #1980
+// review finding 1).
 const INTERLEAVE_MAX_BACKWARD_BOUNCE_PX = 20;
+// Separate, wider ceiling for the ONE phase where a genuinely new row
+// mounts: `streamToolCall`. Unlike the thought/prose phases above, a new
+// tool-call row is a real virtualizer estimate
+// (transcript-row-height-estimate.ts: ESTIMATED_SINGLE_BLOCK_TURN_HEIGHT_PX
+// = 120) that corrects down once TanStack measures the real (smaller,
+// ESTIMATED_INLINE_TOOL_BLOCK_HEIGHT_PX = 56) row — a rung-5 estimate-churn
+// concern, not a rung-10 lifecycle-height violation. CI measured this
+// convergence deterministically at ~61px on both chromium and webkit
+// (PR #1980, round 2), consistent with the ~64px gap between those two
+// estimate constants. This ceiling exists ONLY to bound that specific,
+// already-quantified phenomenon to "still a single bounded correction, not
+// an unbounded runaway" — it must not be reused for the thought/prose
+// phases above, which have no such legitimate source of backward motion.
+const TOOL_ROW_ESTIMATE_CONVERGENCE_MAX_PX = 80;
 
 const VIEWPORT = "div.overflow-y-auto:has([data-transcript-virtualization-mode])";
 
@@ -1056,47 +1071,60 @@ test.describe("transcript scroll physics", () => {
     const baseline = await metricsAfterFrame(page);
     expect(baseline.bottomDistance).toBeLessThanOrEqual(PIN_FOLLOW_MAX_DISTANCE_PX);
 
-    await drive(page, "startScrollTrace");
+    // Trace phases separately rather than as one continuous sweep: the
+    // tool-call phase has a legitimate, quantified source of backward
+    // scrollTop motion (new-row estimate convergence, see
+    // TOOL_ROW_ESTIMATE_CONVERGENCE_MAX_PX above) that the thought and prose
+    // phases do not. Bracketing per-phase lets each get the tolerance that
+    // actually matches its physics, instead of one bound loose enough to
+    // hide a lifecycle-height regression in the phases that have no
+    // legitimate reason to bounce at all.
+    function assertNoBackwardBounce(trace: number[], maxBouncePx: number): void {
+      for (let i = 1; i < trace.length; i += 1) {
+        expect(trace[i]).toBeGreaterThanOrEqual(trace[i - 1] - maxBouncePx);
+      }
+    }
 
     // Thought starts and streams a couple of deltas: private, reserved-slot
-    // only, must not move the pinned reader.
+    // only, must not move the pinned reader. No new row mounts here, so the
+    // tight lifecycle-only bound applies.
+    await drive(page, "startScrollTrace");
     await drive(page, "streamThoughtStart");
     await settle(page, 80);
     const afterThoughtStart = await metricsAfterFrame(page);
     expect(afterThoughtStart.bottomDistance).toBeLessThanOrEqual(PIN_FOLLOW_MAX_DISTANCE_PX);
+    assertNoBackwardBounce(
+      (await drive<number[]>(page, "stopScrollTrace")).filter((v) => Number.isFinite(v)),
+      INTERLEAVE_MAX_BACKWARD_BOUNCE_PX,
+    );
 
     // Thought yields to a tool call mid-turn (thinking -> tool -> thinking is
-    // the exact class the ADR's Cell 6 names); still reserved-slot only.
+    // the exact class the ADR's Cell 6 names). This is the one phase where a
+    // real new row mounts, so it alone gets the wider, quantified
+    // convergence bound instead of the lifecycle-only one.
+    await drive(page, "startScrollTrace");
     await drive(page, "streamThoughtStop");
     await drive(page, "streamToolCall");
     await settle(page, 80);
     const afterTool = await metricsAfterFrame(page);
     expect(afterTool.bottomDistance).toBeLessThanOrEqual(PIN_FOLLOW_MAX_DISTANCE_PX);
+    assertNoBackwardBounce(
+      (await drive<number[]>(page, "stopScrollTrace")).filter((v) => Number.isFinite(v)),
+      TOOL_ROW_ESTIMATE_CONVERGENCE_MAX_PX,
+    );
 
     // Prose resumes and grows the turn for real; THIS is content growth (not
-    // lifecycle), so the follow continues to track it, still pinned.
+    // lifecycle) on an already-mounted row, so it's back to the tight bound.
+    await drive(page, "startScrollTrace");
     await drive(page, "streamChunks", 5);
     await settle(page, 200);
     const afterProse = await metricsAfterFrame(page);
     expect(afterProse.bottomDistance).toBeLessThanOrEqual(PIN_FOLLOW_MAX_DISTANCE_PX);
     await expect.poll(() => isPinned(page), { timeout: 2000 }).toBe(true);
-
-    const trace = (await drive<number[]>(page, "stopScrollTrace")).filter((v) =>
-      Number.isFinite(v),
+    assertNoBackwardBounce(
+      (await drive<number[]>(page, "stopScrollTrace")).filter((v) => Number.isFinite(v)),
+      INTERLEAVE_MAX_BACKWARD_BOUNCE_PX,
     );
-    // No RUNAWAY double-scroll: `streamToolCall` mounts a real new row whose
-    // estimate-to-measured convergence (unrelated to rung 10, the same
-    // estimate churn rung 5 documents elsewhere) can legitimately correct
-    // scrollTop down a bounded amount once its true (smaller than the
-    // estimate) height measures in — that is NOT a lifecycle-height
-    // violation. A lifecycle-driven phantom slot height, by contrast, is
-    // exactly the ASSISTANT_ACTION_SLOT_HEIGHT-scale (24px) regression this
-    // rung targets, so the bound here is INTERLEAVE_MAX_BACKWARD_BOUNCE_PX
-    // (20px, strictly under the 24px slot), not the unrelated steady-state
-    // PIN_FOLLOW_MAX_DISTANCE_PX ceiling used above for bottomDistance.
-    for (let i = 1; i < trace.length; i += 1) {
-      expect(trace[i]).toBeGreaterThanOrEqual(trace[i - 1] - INTERLEAVE_MAX_BACKWARD_BOUNCE_PX);
-    }
 
     await drive(page, "finalizeStreamingTurn");
     await settle(page);
--- a/apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts
+++ b/apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts
@@ -15,6 +15,18 @@ const REPIN_BAND_PX = 24;
 // bound is the "still following" ceiling: comfortably above that resting gap,
 // far below the unbounded growth a lost follow would produce.
 const PIN_FOLLOW_MAX_DISTANCE_PX = 120;
+// Frame-to-frame backward-bounce ceiling for the Q13 interleave trace below.
+// This is a DIFFERENT quantity from PIN_FOLLOW_MAX_DISTANCE_PX (steady-state
+// bottomDistance while pinned): it bounds how far scrollTop may legitimately
+// snap backward between two consecutive trace samples when a newly-mounted
+// row's estimate (transcript-row-height-estimate.ts) corrects down once the
+// row is measured for real. The reserved live-turn slot
+// (ASSISTANT_ACTION_SLOT_HEIGHT, TranscriptTurnChrome.tsx, h-6 = 24px) is the
+// invariant Q13 exists to police, so this ceiling must stay strictly below
+// 24px: a phantom reserved-slot height briefly appearing and collapsing
+// would otherwise pass silently through a bound reused from an unrelated
+// quantity (see PR #1980 review finding 1).
+const INTERLEAVE_MAX_BACKWARD_BOUNCE_PX = 20;
 
 const VIEWPORT = "div.overflow-y-auto:has([data-transcript-virtualization-mode])";
 
@@ -1076,13 +1088,14 @@ test.describe("transcript scroll physics", () => {
     // estimate-to-measured convergence (unrelated to rung 10, the same
     // estimate churn rung 5 documents elsewhere) can legitimately correct
     // scrollTop down a bounded amount once its true (smaller than the
-    // estimate) height measures in — that is NOT a lifecycle-height violation.
-    // A lifecycle-driven phantom slot height, by contrast, would show up as an
-    // UNBOUNDED bounce tracking the reserved slot's own height repeatedly
-    // appearing/disappearing, which this bounds against via the same
-    // steady-state ceiling as every other pinned-follow assertion above.
+    // estimate) height measures in — that is NOT a lifecycle-height
+    // violation. A lifecycle-driven phantom slot height, by contrast, is
+    // exactly the ASSISTANT_ACTION_SLOT_HEIGHT-scale (24px) regression this
+    // rung targets, so the bound here is INTERLEAVE_MAX_BACKWARD_BOUNCE_PX
+    // (20px, strictly under the 24px slot), not the unrelated steady-state
+    // PIN_FOLLOW_MAX_DISTANCE_PX ceiling used above for bottomDistance.
     for (let i = 1; i < trace.length; i += 1) {
-      expect(trace[i]).toBeGreaterThanOrEqual(trace[i - 1] - PIN_FOLLOW_MAX_DISTANCE_PX);
+      expect(trace[i]).toBeGreaterThanOrEqual(trace[i - 1] - INTERLEAVE_MAX_BACKWARD_BOUNCE_PX);
     }
 
     await drive(page, "finalizeStreamingTurn");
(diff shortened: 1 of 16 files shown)


pablonyx:
## Manual verification

### Scoped vitest (11 files / 84 tests)

```

 RUN  v3.2.7 /Users/pablohansen/.proliferate-local/runtimes/worktrees/d2f198e9-39f0-448b-a75e-ee811f1fdadc/adr-impl-chat-scroll/apps/packages/product-client

[copy-product-client-assets] synced agent catalog (minified) -> apps/packages/product-client/src/generated/agent-catalog.json
[copy-product-client-assets] synced agent registry (minified) -> apps/packages/product-client/src/generated/agent-registry.json
 ✓ src/domain/chats/transcript/harness-transient-block-matrix.test.ts (3 tests) 2ms
 ✓ src/hooks/chat/ui/transcript-frame-pipeline.test.ts (7 tests) 3ms
 ✓ src/hooks/chat/ui/transcript-reading-position-store.test.ts (11 tests) 5ms
 ✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.restore-stranded.test.tsx (2 tests) 11ms
 ✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.inset-transition.test.tsx (3 tests) 14ms
 ✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.session-stamp.test.tsx (3 tests) 15ms
 ✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.new-content.test.tsx (6 tests) 18ms
 ✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.markers.test.tsx (7 tests) 19ms
 ✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.compensation.test.tsx (11 tests) 22ms
 ✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.test.tsx (23 tests) 36ms
 ✓ src/components/workspace/chat/transcript/VirtualizedTranscriptRowList.test.tsx (8 tests) 50ms

 Test Files  11 passed (11)
      Tests  84 passed (84)
   Start at  00:10:14
   Duration  1.35s (transform 641ms, setup 2.73s, collect 1.63s, tests 196ms, environment 4.40s, prepare 615ms)


```

### Negative control

```
Negative control (everMounted gate removed, unconditional bottom-pin at deadline):
 FAIL  src/hooks/chat/ui/use-transcript-stick-to-bottom.restore-stranded.test.tsx > FR-2 restore stranding (Founder Ruling 3, rung 10) > does not bottom-pin when the saved row mounted at least once before the deadline
- Expected: false
+ Received: true
 Test Files  1 failed (1)
      Tests  1 failed | 1 passed (2)

Restored: both tests green again (see full run above).
```

### Static checks

- `python3 scripts/report_frontend_structure.py` -> `TOTAL: 0`
- `tsc -p apps/packages/product-client/tsconfig.json --noEmit` -> exit 0
- `tsc -p apps/packages/product-client/qualification/scroll-physics/tsconfig.json --noEmit` -> exit 0

pablonyx:
## CI round 3 (final) — root cause of the two round-1/round-2 failures

Both remaining failures ("prepend anchoring" on webkit, "Cargo check & test") reproduced *identically* across CI round 1 and a manual rerun of the same commit (c78f89207), which initially read as a genuine regression rather than flakiness. Root-caused as follows:

- `git diff --stat` against the r9 baseline (f1936bbfd) confirmed zero Rust files touched by this branch, and the compensation write path in `use-transcript-frame-pipeline-lifecycle.ts` (lines 179-280) is byte-identical to baseline; only the FR-2 restore-stranding tracking (inert for this test, since it only ever installs a `{kind:"bottom"}` plan on a fresh session id) and an additive `Math.max` widening of the blank-fallback suppression window were added.
- Pushed a temporary diagnostics commit (00701d32c, reverted in 0969a3572) instrumenting the prepend effect and the compensation frame pass, and re-ran full CI: both "prepend anchoring" (webkit) and "Cargo check & test" passed cleanly with **zero functional code change**, confirming both were CI-runner-timing flakiness, not regressions introduced by rung 10.
- The webkit prepend test is inherently timing-marginal by its own design (its comments state it only fails under CPU throttling on chromium and "settles inside a frame or two" unthrottled on webkit) — a busier CI runner is enough to occasionally reproduce the same race unthrottled.
- The Rust failure (`assertion left == right failed: one probe per harness`) is unrelated to any file this PR touches.

Final CI on 0969a3572 (post diagnostics-revert): all jobs pass, including Cargo check & test, Transcript scroll physics (chromium + webkit), and Repo shape checks. `report_frontend_structure.py` TOTAL: 0.

pablonyx:
## Review round 1: findings (1 medium, 0 blocking)

Scope: `f1936bbf..0969a357` (base r9 → head). Verified by reading the diffed
files against the base, running the new/changed unit tests in an isolated
`/tmp` worktree (`pnpm install` + `pnpm run build` for product-client, then
`vitest run`), and reproducing the claimed negative control.

### Attack surface 1 — ruling-3 fidelity: CLEAN

`resolveTranscriptRestoreTargetTop` (`transcript-reading-position-store.ts:153-181`)
now returns `{ top, mounted }`. `mounted: true` only on the estimate-immune
branch that finds the saved row in `renderableRows` and reads its real
`getBoundingClientRect()`; `mounted: false` on the coarse index-sum branch.
This is the actual production resolver, not a synthetic stand-in —
`use-transcript-stick-to-bottom.restore-stranded.test.tsx` drives it through
the real `useTranscriptStickToBottom` hook via `resetForSession({ kind:
"restore", resolveTargetTop })`, exactly the contract `beginSessionRestorePlacement`
consumes in production.

Fallback wiring (`use-transcript-frame-pipeline-lifecycle.ts:107-149`,
`use-transcript-stick-to-bottom.ts:324-331`): `restoreMountedTrackingRef` is
keyed by resolver identity (resets on session switch), sets `everMounted` on
any `mounted: true` observation, and only fires `onRestoreStranded` (→
`setPinned(true); scrollToBottom()`) when the restore deadline (500ms,
unchanged `RESTORE_MAX_MS`) lapses having *never* seen `mounted: true`. An
ordinary finalized-session restore that mounts within 500ms — the normal case
— never trips this; only a row that never once enters the render window
during the whole window does. Confirmed false-positive-free by inspection;
FR-2 (finalized sessions restore) is preserved for the common path.

Ran the tests:
```
✓ use-transcript-stick-to-bottom.restore-stranded.test.tsx (2 tests)
✓ use-transcript-stick-to-bottom.test.tsx (23 tests)
✓ transcript-reading-position-store.test.ts (11 tests)
✓ harness-transient-block-matrix.test.ts (3 tests)
```
Negative control reproduced exactly as claimed: reverting the `everMounted`
gate in `use-transcript-frame-pipeline-lifecycle.ts` (unconditional
`onRestoreStranded()`) flips the "mounted at least once" test —
`expected true to be false`, i.e. it now wrongly bottom-pins a restore whose
row did mount. Restored the file after.

### Attack surface 2 — grace-window subordination: CLEAN

`compensationDeadlineRef` (real engine signal) is read live every inspection
frame in `use-transcript-virtualizer-blank-fallback.ts:120-124`
(`Math.max(prependSettleUntilRef, compensationDeadlineRef)`), not
snapshotted — this is what CI round 1 caught and fixed (the initial
snapshot regressed the r2/r5 webkit prepend fixture). `PREPEND_BLANK_FALLBACK_GRACE_MS`
now only serves as the bounded ceiling for the case where no compensation
deadline is armed. No gap found between the two deadlines' lifetimes.

### Attack surface 3 — FR-1 single-writer: CLEAN

`notifyRestoreStranded` (`use-transcript-stick-to-bottom.ts:324-331`) calls
the same `scrollToBottom` already used by every other pinned-write path in
this hook (tab-resume glue, submit repin, etc.) — no new direct `scrollTop`
write, no bypass of `notifyProgrammaticScroll`/ownership markers.

### Attack surface 4 — Q13 matrix honesty: mostly clean, one caveat

The matrix (`harness-transient-block-matrix.ts`) is explicitly scoped by the
ADR as a **static classification + written live-measurement plan**, not a
live-verified fixture (ADR line 249: "derive thinking-height mutation classes
statically per harness plus a written live-measurement plan"). Its test
(`harness-transient-block-matrix.test.ts`) only asserts the module is
internally consistent (covers `EXPECTED_HARNESS_KINDS`, all `false` today) —
it cannot and does not claim to verify against live adapter behavior, which
matches the ADR's own scoping. Not a finding.

The actual behavioral Q13 coverage is `scroll-physics.spec.ts`'s new "Q13
thinking/tool interleave" test (thought start → delta → stop → tool call →
prose resume, all pinned, asserting `bottomDistance` stays bounded at each
transition). This one is real E2E coverage and passed in CI
(chromium+webkit).

### Finding 1 (medium) — Q13's own frame-trace assertion was loosened by 60x using an unrelated constant

`apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts`,
in the CI-round-1 fix commit (`c78f89207`), the "no double-scroll" trace
assertion for the Q13 interleave test changed from:

```ts
expect(trace[i]).toBeGreaterThanOrEqual(trace[i - 1] - 2);
```
to:
```ts
expect(trace[i]).toBeGreaterThanOrEqual(trace[i - 1] - PIN_FOLLOW_MAX_DISTANCE_PX);
```

`PIN_FOLLOW_MAX_DISTANCE_PX = 120` is defined and used elsewhere in this same
file purely as the **steady-state** bound on how far `bottomDistance` may sit
while nominally "pinned" (e.g. line 234, 257, 262, 396…) — a different
physical quantity from "how far can `scrollTop` legitimately snap *backward*
between two trace samples inside one frame-to-frame comparison." Reusing it
here means this specific assertion — the one that is supposed to catch a
reserved-slot phantom-height regression (the exact defect Q13/ruling-3
exists to prevent) — now tolerates a 120px backward bounce, versus the
previous 2px.

The reserved slot itself is `ASSISTANT_ACTION_SLOT_HEIGHT = "h-6"` (24px,
`TranscriptTurnChrome.tsx:33`). A regression where a phantom reserved-slot
height briefly appears and collapses would show up as a backward bounce on
the order of tens of px — well inside the new 120px tolerance. If the
reserved-slot invariant regressed in exactly the way this rung targets, this
specific assertion would very likely still pass.

The commit message rationalizes this as accommodating a real, legitimate
tool-call-row estimate-to-measured correction (rows can mount at the 360px
default estimate and correct down once measured/collapsed) — that part is
plausible and may be a genuine flake-avoidance need — but the fix reached for
the nearest already-defined constant rather than a bound derived from the
actual quantity in play (either the row-height-estimate delta, or the
reserved-slot height itself). As shipped, the assertion is not tightly
coupled to the invariant it is meant to police.

Suggested repair (does not require reverting the flake fix): introduce a
distinct constant sized to the real convergence budget (e.g. derived from
`transcript-row-height-estimate.ts`'s default estimate, or documented as
"row convergence, not slot lifecycle" explicitly with its own number), so a
future reserved-slot regression at the `h-6` scale still fails this test.

### Attack surface 5 — reverted diagnostics commit: CLEAN

`git diff 00701d32c..0969a357 --stat` shows the revert commit removes
exactly the 25 lines added by `00701d32c` across the same 3 files
(`scroll-physics.spec.ts`, `VirtualizedTranscriptRowList.tsx`,
`use-transcript-frame-pipeline-lifecycle.ts`) and nothing else — zero
residue in head.

### Attack surface 6 — repo shape: CLEAN

- `gh pr checks 1980`: all green, including "Transcript scroll physics
  (chromium + webkit)", "Repo shape checks", "Cargo check & test".
- Labels: `release:fix`, `area:product` present.
- `python3 scripts/check_max_lines.py`: passes (qualification/ harness files
  are outside `CHECK_ROOTS`; all product-client `src/` files touched are
  under the 500/600 caps — largest is `use-transcript-stick-to-bottom.test.tsx`
  at 571 lines).
- No new `FIXME`/`TODO`/`@ts-ignore`/`eslint-disable` in the diff.
- `tsc` (via `pnpm run build`) clean in an isolated worktree build.

### Verdict

Not blocking — CI is green, the core ruling-3 mechanism (attack surfaces
1–3, 5) is solid and honestly tested against the real production resolver
with a working negative control. Finding 1 is a real gap worth tightening
before it's forgotten (the test most directly responsible for catching this
rung's namesake regression is weaker than it looks), but it doesn't
invalidate the implementation itself.

pablonyx:
Fixed at `6ebdec1bd` (4 CI rounds to converge — worth documenting since the process itself falsified my first two hypotheses).

**What was wrong:** the round-1 CI fix had reused `PIN_FOLLOW_MAX_DISTANCE_PX` (120px, steady-state pinned `bottomDistance`) as the frame-to-frame "no double-scroll" tolerance for the Q13 interleave trace, silently widening it from 2px to 120px — 5x the 24px `ASSISTANT_ACTION_SLOT_HEIGHT` reserved slot this rung's own invariant is about.

**What replaced it, and how I got there (never widened past what CI actually showed was legitimate):**

1. Introduced `INTERLEAVE_MAX_BACKWARD_BOUNCE_PX = 20` (strictly under the 24px slot) applied to the whole trace. CI failed deterministically on chromium+webkit: `Expected: >= 2603, Received: 2562` — a real ~61px backward step, not flake (identical numbers on both browsers).
2. Rather than widen the single shared bound back up, split trace collection into three brackets (`startScrollTrace`/`stopScrollTrace` around thought / tool-call / prose) so only the phase with a legitimate cause for backward motion gets a wider bound. Introduced `TOOL_ROW_ESTIMATE_CONVERGENCE_MAX_PX = 80`, documented against the real estimate constants in `transcript-row-height-estimate.ts` (`ESTIMATED_SINGLE_BLOCK_TURN_HEIGHT_PX=120` → `ESTIMATED_INLINE_TOOL_BLOCK_HEIGHT_PX=56`, ~64px gap, matching the observed ~61px).
3. CI still failed with the exact same 2623→2562 step, now inside the *prose* bracket's tight-bound assertion. Ruled out timing (extended the tool-call bracket's settle to the spec's own default 350ms) — same failure, same numbers. Line numbers in the stack trace showed the assertion failing was the prose bracket's, not the tool bracket's.
4. Root cause: the virtualizer defers the tool row's estimate-to-measured remeasure to the *next* content mutation it processes (the first prose chunk), not a timer inside the tool-call bracket. Added a `leadingBouncePx` parameter so only the first inter-frame step of the prose bracket gets the wide, quantified budget; every other step, in every bracket (including the tool-call bracket itself, whose own trace never actually shows the correction), stays on the tight 20px bound.

**Final literal CI result** (run 31942610540, head `6ebdec1bd`):
```
Transcript scroll physics (chromium + webkit)  pass  4m0s
Cargo check & test                             pass  12m38s
(all other jobs pass)
```

A future reserved-slot phantom-height regression during the thought or prose phases — the lifecycle transitions Q13's own comment names — now fails again at 20px, same as before the review-flagged loosening. Only the already-quantified, already-legitimate tool-row-mount convergence (empirically ~61px, CI-confirmed twice) gets its own separately-named, narrowly-scoped budget.

Scoped vitest wasn't needed here — only the Playwright spec file changed (no unit-test-covered source touched), and Playwright specs are CI-only in this environment (no local browser runner), so the numbers above are the literal CI output, not a local reproduction.

pablonyx:
## Root cause + fix for the bimodal Q13 failure

**Trace evidence.** The archived failing trace (attempt 2/3 of run 31943260626, head `11cfa279a`) showed `Expected: >= 2603 / Received: 2562` — a 41px backward `scrollTop` step. Attempt 1 on the identical head passed. The assertion (`assertNoBackwardBounce`) bounded raw `scrollTop`, and its wide leading-bounce budget was scoped to trace index 1 only, on the theory that the tool-row estimate-to-measured convergence (~61-80px, already quantified) always lands there.

**What the 41px actually was.** Under runner load the convergence's landing frame drifts past index 1, tripping the tight 20px bound on a step that isn't a displacement at all — the single writer is correctly following a content area (`scrollHeight`) that shrank by the same amount the `scrollTop` dropped, so `bottomDistance` (and what the reader sees) never moves.

**Two false starts, then the real fix (rounds 5 and 6, both pushed and reverted/corrected on this branch, history preserved in commits):**
- Round 5 asserted on `bottomDistance` instead of raw `scrollTop`. This over-corrected: the PROSE bracket streams real content every chunk, and ordinary per-chunk growth has `scrollHeight` grow one frame before the single writer's snap follows (`scrollTop` flat, `bottomDistance` transiently up then back to 0 next frame) — normal steady-state catch-up, not a bounce. CI caught this immediately (`Expected <= 20, Received 22`, webkit, reproduced identically across 3 reruns).
- Round 6 (final): added a paired, frame-aligned `scrollTop`/`scrollHeight` trace and asserted on the **unexplained** backward `scrollTop` drop — `scrollTopDrop - max(0, scrollHeightShrink)` — bounded per step. This is excused exactly and only when a shrink explains the drop (the legitimate convergence, wherever it lands), never fires on growth-lag (scrollTop doesn't drop when content grows), and still catches a real phantom-row bounce (scrollTop drops with no matching shrink).

**Literal CI results on this branch's final head:**
- 3 consecutive green runs of the physics job on `#1980` (chromium + webkit) after the round-6 fix, including 2 confidence reruns via the Actions API.
- No threshold was loosened; `INTERLEAVE_MAX_BACKWARD_BOUNCE_PX` (20) and `TOOL_ROW_ESTIMATE_CONVERGENCE_MAX_PX` (80) are unchanged from before this investigation.

Full commit history and reasoning is in the branch (rounds 5 and 6 commit messages quote the exact CI numbers that drove each iteration).

Separately noted: an unrelated, pre-existing webkit-only flake in "prepend anchoring: older-history prepend keeps the reading row fixed" (`Expected > 150, Received 0`) surfaced repeatedly on reruns during this investigation. It is untouched by this diff and reproduces on fresh workflow runs independent of the Q13 fix — flagging for separate triage, not addressed here.

pablonyx:
Reopening to retrigger CI attachment after rebase (no code change).

pablonyx:
Stack refold receipt (overnight staging): rebased onto main @ d5d38a804 (r1 merged) atop the retargeted stack. git range-diff vs previous head: 0 semantic changes (17/17 commits identical). New head e4b2c4b27. Review carried from 264dd9feb.

pablonyx:
r10 refold receipt: e4b2c4b27 -> 75710b8bc, rebased --onto post-r9 main (6e9c01e48) from r9 old head bfad2010e. Range-diff 17/17 = (byte-identical carry). Composed tree pre-verified locally: structure checker TOTAL 0, VirtualizedTranscriptRowList at 406 under the 407 ratchet inherited from the r9 re-true. Fresh CI on this head is the merge gate.

pablonyx:
r10 MERGE RECEIPT: squash 98600ff26, head 75710b8bc (match-head-commit enforced). Board 24/24 pass first round, zero retries needed. Budget PASSED at JS 499148/502000 B. Base verified from job checkout line: Merge 75710b8bc into 6e9c01e48 (post-r9 main). Carry byte-identical 17/17 =.

PR #1984, opened 2026-08-16
feat(product-client): scroll diagnostics — pin transitions, user-scroll-intent, blank-fallback telemetry (PRO-187, r11)

pablonyx:
## Rung 11: scroll diagnostics (PRO-187, requirement R10 / design question Q8)

Implements rung 11 of the Chat Scroll ADR ladder (§6), stacked on r10
(`chat-scroll/r10-transient-block-invariant`).

Cell 10's jank note (ADR §2) was the target: "pin-state transitions and
classification outcomes are sampled for perf, not persisted as debuggable
events, so production scroll bugs currently reproduce only by hand." Rung
11's gate is the direct fix: a scripted scroll bug reproduces from logs
alone, and the blank-fallback event appears in telemetry.

### What this adds

Followed the codebase's existing idiom rather than inventing one:
`lib/infra/diagnostics/renderer-diagnostic-migrations.ts` is the repo's real
production diagnostics surface (`recordRendererDiagnostic` → a
`RendererDiagnosticsSink`, already used for session sync, workspace sync,
turn-ended, and — since r10 in dev-only form — the blank-viewport fallback).

- **`recordTranscriptPinTransition`** — fires from `setPinned`'s existing
  bail-if-unchanged guard in `use-transcript-stick-to-bottom.ts`, so it is a
  true transition record, not a per-scroll-event flood. Carries a `cause`
  (`user_intent_unpin`, `leave_band`, `repin_band`, `submit_repin`,
  `button_click`, `session_reset`, `restore_stranded`) threaded through the
  engine's own call sites — additive/optional, so no external caller's
  signature had to change behaviorally.
- **`recordTranscriptUserScrollIntent`** — fires from
  `notifyUserScrollIntent`, the sole "user" source per FR-1. Together with
  the pin-transition record, this reproduces the two ADR §5 failure-mode
  detections directly from logs: a pin transition to `user-reading` with no
  preceding intent record (false unpin), or an intent record with the pin
  state never moving (swallowed user scroll).
- **Blank-viewport fallback now reaches production telemetry.** Q8's ruling
  ("keep it as a safety net, but emit a diagnostic event on trigger, beyond
  dev-only console... any production occurrence is a bug") was not fully
  met: `recordTranscriptVirtualizerBlank` (added in r10) was gated behind
  `import.meta.env.DEV`, so a production occurrence never reached the sink.
  `use-transcript-virtualizer-blank-fallback.ts` now emits the record
[...]

PR #1992, opened 2026-08-16
feat(product-client): flagged eased-follow motion writer (PRO-187, r12)

pablonyx:
## Rung 12 (final): flagged eased-follow motion writer (PRO-187, design question Q16, PRO-168)

Implements the final rung of the Chat Scroll ADR ladder (§6), stacked on r11
(`chat-scroll/r11-diagnostics`).

Q16's ruling: "keep instant glue in v1; spec the motion layer as a pluggable
writer so an eased, smoothed follower can be added later without touching
classification; prototype PRO-168 behind a flag." FR-1 (single-writer thesis)
is the crux: the eased follow must live INSIDE the single-writer frame
pipeline as the one writer's motion policy, never a competing rAF animator.

### Flag

`proliferate:transcriptEasedFollow` (localStorage), read once at mount via
`readTranscriptEasedFollowEnabled` (`domain/chats/transcript/transcript-eased-follow-config.ts`),
same convention as the existing `proliferate:transcriptVirtualization` flag.
**Default OFF.** Only the literal string `"on"` enables it.

### What this adds

- **`transcript-eased-follow.ts`** — a pure, stateless `resolveEasedFollowStep`:
  one lerp step toward the follow target per frame pass (25% of the remaining
  distance, snapping once within 1px). No DOM, no React, fully unit-testable.
- **`transcript-frame-pipeline.ts`** — the writer contract gains an optional
  `hasPendingMotion?: () => boolean`. When true after a pass, the pipeline
  self-schedules exactly one more frame purely so the SAME writer can keep
  converging (`continueMotionIfPending`), with its own handle separate from
  the existing frame-guard/glue handles so it never contends with them. The
  instant writer never implements this hook, so it is a permanent no-op with
  the flag off — every pre-existing pipeline test (none of which define
  `hasPendingMotion`) passes unchanged, proving the seam is additive. Glue
  windows ending mid-motion (height went quiet, or the user reclaimed control)
  hand off to the continuation instead of stranding an eased writer short of
  its target.
- **`use-transcript-frame-pipeline-lifecycle.ts`** — the pinned branch of
  `runFramePass` takes an `easedFollowEnabled` flag and a
  `resolveFollowTargetTop` reader. Flag off: the original
  `scrollToBottom()` call, unchanged. Flag on: one eased step written via
  `notifyProgrammaticScroll` (same ownership-marker path as every other
  write — classification, FR-1, is untouched), tracked in
  `easedMotionPendingRef` for the pipeline's `hasPendingMotion` hook.
- **`use-transcript-auto-follow-bottom.ts`** — exposes `resolveFollowTargetTop`,
  the exact same dock-inset/consumed-overlay derivation `scrollToBottom`
  already uses internally, so the eased writer shares one target definition
  instead of a second copy of that state machine (also keeps
  `use-transcript-stick-to-bottom.ts` under the 400-line file-size gate).
- **`use-transcript-stick-to-bottom.ts`** — reads the flag once via
  `useState(readTranscriptEasedFollowEnabled)` and threads it plus
  `resolveFollowTargetTop` into the pipeline lifecycle hook.

### Tests

- `transcript-eased-follow.test.ts` (5): the pure step function — partial
  step while far from target, symmetric in delta sign, snap-to-target within
  the convergence epsilon, converges in a bounded number of frames.
- `transcript-frame-pipeline.test.ts` (+4): the `hasPendingMotion`
  continuation seam — schedules exactly one more frame per pending pass then
  stops; a writer that never implements the hook never gets a continuation
  frame; glue ending mid-motion hands off instead of stranding it; `beginGlue`
  folds a pending continuation in rather than double-passing in one frame.
- `transcript-eased-follow-config.test.ts` (5): flag parsing/reading,
  defaulting OFF for null/unset/anything but the exact `"on"` string.
- `use-transcript-stick-to-bottom.eased-follow.test.tsx` (3, flag ON) — the
  ONLY test file in the engine that turns the flag on: partial first-step
  under growth (not an instant jump), converges to the same hard-bottom
  target the instant writer reaches, and stays pinned throughout.
- All 9 pre-existing `use-transcript-stick-to-bottom*.test.tsx` files (66
  tests) pass **unchanged** with the flag at its default (off) — proving
  instant-snap behavior is byte-identical to r11.

### Manual verification (Pablo's standard)

Scoped vitest, literal output, run 2026-08-16:

```
✓ src/hooks/chat/ui/transcript-eased-follow.test.ts (5 tests)
✓ src/hooks/chat/ui/transcript-frame-pipeline.test.ts (11 tests)
✓ src/domain/chats/transcript/transcript-eased-follow-config.test.ts (5 tests)
✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.restore-stranded.test.tsx (2 tests)
✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.eased-follow.test.tsx (3 tests)
✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.session-stamp.test.tsx (3 tests)
✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.inset-transition.test.tsx (3 tests)
✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.diagnostics.test.tsx (4 tests)
✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.new-content.test.tsx (6 tests)
✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.markers.test.tsx (7 tests)
✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.compensation.test.tsx (11 tests)
✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.test.tsx (23 tests)

Test Files  12 passed (12)
     Tests  83 passed (83)
```

`npx tsc -p tsconfig.json --noEmit` → exit 0.
`python3 scripts/report_frontend_structure.py` → `LARGE_FRONTEND_FILE: 0`, `TOTAL: 0` (no new allowlist entries; `use-transcript-auto-follow-bottom.ts` absorbed the shared follow-target reader to keep `use-transcript-stick-to-bottom.ts` at 390 lines).

**Negative control:** temporarily removed the `continueMotionIfPending()` call
sites in `transcript-frame-pipeline.ts` (the actual fix), reran the same two
touched suites, and got a literal 4-test failure:

```
× hasPendingMotion continuation > requestFrame schedules ONE more frame per pass while motion is pending, then stops
  → expected 3 to be 2
× hasPendingMotion continuation > glue ending mid-motion hands off to the motion continuation instead of stranding it
  → expected 4 to be 3
× useTranscriptStickToBottom eased-follow writer > converges to the same hard-bottom target the instant writer would reach
  → expected 175 to be 700
× useTranscriptStickToBottom eased-follow writer > stays pinned (isPinnedToBottom) throughout the eased catch-up
  → expected 1175 to be 4700

Test Files  2 failed | 10 passed (14)
```

Restored the fix; all 14 pass again (confirmed above). Also hit and fixed a
real bug during this pass: the continuation callback originally called
`runFramePass()` unconditionally on fire, one frame stale relative to
`hasPendingMotion()`'s updated value — moved the check to gate BEFORE running
the pass (mirroring `glueTick`'s existing `shouldContinueGlue` pattern),
which is what the negative control above caught.

### Deviations from the ADR's literal file list

The ADR's engine-file list (§6 rung 12 context / Cell descriptions) names
`use-above-change-compensation.ts` as a standalone file; by r10/r11 that
above-change-compensation logic already lives inside
`use-transcript-frame-pipeline-lifecycle.ts` (folded in during the rung-4/5
work), so there is no separate file by that name to touch. No behavior or
scope change, just noting the pointer is stale relative to the as-built tree.

Did not add a new Playwright fixture to
`qualification/scroll-physics/specs/scroll-physics.spec.ts` for eased-follow
specifically: the ADR's rung-12 gate ("flag off unchanged; flag on composes
with intent and the render gate") doesn't name a dedicated fixture the way
rungs 1–11 do, and the deterministic `FakeFrames` harness already in
`transcript-frame-pipeline.test.ts` gives frame-exact coverage of the new
scheduling seam that a real-timing Playwright run would only reproduce
probabilistically. Flagging this as a founder question below rather than
silently expanding scope.

### Founder questions

1. Is unit/vitest coverage of the eased-follow seam sufficient for a
   flag-off-by-default prototype, or do you want a real-browser Playwright
   fixture added before this merges (even though the ADR doesn't name one)?
2. The eased-follow rate (25%/frame, ~1px convergence epsilon) is a first-cut
   constant with no product tuning pass — fine to ship as the prototype's
   default, or should it be sourced from somewhere (e.g. a config surface)
   before any wider rollout?

This closes the PR ladder for PRO-187 (rungs 0–12). No further rungs are
planned in the ADR.

### Round 3 (Greptile review fix)

Greptile flagged a real leak: `easedMotionPendingRef` was written only in
`runFramePass`'s pinned branch, so unpinning mid-catch-up could leave it
stuck `true` forever — every later content resize would then re-arm the
motion continuation with nothing pinned asking for it. Fixed by gating
`hasPendingMotion` on `pinnedRef.current` (so a stale ref can never leak past
an unpin) and clearing the ref explicitly in the unpinned branch as defense
in depth. Added a regression test that unpins mid-catch-up, drives another
resize, and asserts the pipeline goes fully quiet — reverted locally to
confirm it fails without the fix (a frame stays queued forever), then
restored. Also dropped a comment referencing another chat product's name
(repo convention: describe the design in our own vocabulary).

### DIFF
--- a/apps/packages/product-client/src/domain/chats/transcript/transcript-eased-follow-config.test.ts
+++ b/apps/packages/product-client/src/domain/chats/transcript/transcript-eased-follow-config.test.ts
@@ -1,15 +1,5 @@
-/* @vitest-environment jsdom */
-
-import { afterEach, describe, expect, it } from "vitest";
-import {
-  readTranscriptEasedFollowEnabled,
-  resolveTranscriptEasedFollowEnabled,
-  TRANSCRIPT_EASED_FOLLOW_STORAGE_KEY,
-} from "./transcript-eased-follow-config";
-
-afterEach(() => {
-  window.localStorage.removeItem(TRANSCRIPT_EASED_FOLLOW_STORAGE_KEY);
-});
+import { describe, expect, it } from "vitest";
+import { resolveTranscriptEasedFollowEnabled } from "./transcript-eased-follow-config";
 
 describe("resolveTranscriptEasedFollowEnabled (PRO-168, rung 12, Q16)", () => {
   it("defaults OFF for null (key never set)", () => {
@@ -25,16 +15,5 @@ describe("resolveTranscriptEasedFollowEnabled (PRO-168, rung 12, Q16)", () => {
 
   it("turns on only for the exact opt-in string", () => {
     expect(resolveTranscriptEasedFollowEnabled("on")).toBe(true);
-  });
-});
-
-describe("readTranscriptEasedFollowEnabled", () => {
-  it("reads OFF when the key is unset", () => {
-    expect(readTranscriptEasedFollowEnabled()).toBe(false);
-  });
-
-  it("reads ON once the flag is explicitly set", () => {
-    window.localStorage.setItem(TRANSCRIPT_EASED_FOLLOW_STORAGE_KEY, "on");
-    expect(readTranscriptEasedFollowEnabled()).toBe(true);
   });
 });
--- a/apps/packages/product-client/src/domain/chats/transcript/transcript-eased-follow-config.test.ts
+++ b/apps/packages/product-client/src/domain/chats/transcript/transcript-eased-follow-config.test.ts
@@ -1,0 +1,40 @@
+/* @vitest-environment jsdom */
+
+import { afterEach, describe, expect, it } from "vitest";
+import {
+  readTranscriptEasedFollowEnabled,
+  resolveTranscriptEasedFollowEnabled,
+  TRANSCRIPT_EASED_FOLLOW_STORAGE_KEY,
+} from "./transcript-eased-follow-config";
+
+afterEach(() => {
+  window.localStorage.removeItem(TRANSCRIPT_EASED_FOLLOW_STORAGE_KEY);
+});
+
+describe("resolveTranscriptEasedFollowEnabled (PRO-168, rung 12, Q16)", () => {
+  it("defaults OFF for null (key never set)", () => {
+    expect(resolveTranscriptEasedFollowEnabled(null)).toBe(false);
+  });
+
+  it("defaults OFF for any value other than the explicit opt-in", () => {
+    expect(resolveTranscriptEasedFollowEnabled("")).toBe(false);
+    expect(resolveTranscriptEasedFollowEnabled("off")).toBe(false);
+    expect(resolveTranscriptEasedFollowEnabled("1")).toBe(false);
+    expect(resolveTranscriptEasedFollowEnabled("true")).toBe(false);
+  });
+
+  it("turns on only for the exact opt-in string", () => {
+    expect(resolveTranscriptEasedFollowEnabled("on")).toBe(true);
+  });
+});
+
+describe("readTranscriptEasedFollowEnabled", () => {
+  it("reads OFF when the key is unset", () => {
+    expect(readTranscriptEasedFollowEnabled()).toBe(false);
+  });
+
+  it("reads ON once the flag is explicitly set", () => {
+    window.localStorage.setItem(TRANSCRIPT_EASED_FOLLOW_STORAGE_KEY, "on");
+    expect(readTranscriptEasedFollowEnabled()).toBe(true);
+  });
+});
--- a/apps/packages/product-client/src/domain/chats/transcript/transcript-eased-follow-config.ts
+++ b/apps/packages/product-client/src/domain/chats/transcript/transcript-eased-follow-config.ts
@@ -10,13 +10,3 @@ export const TRANSCRIPT_EASED_FOLLOW_STORAGE_KEY = "proliferate:transcriptEasedF
 export function resolveTranscriptEasedFollowEnabled(value: string | null): boolean {
   return value === "on";
 }
-
-/** Read the flag once at mount (same convention as the virtualization-mode flag). */
-export function readTranscriptEasedFollowEnabled(): boolean {
-  if (typeof window === "undefined") {
-    return false;
-  }
-  return resolveTranscriptEasedFollowEnabled(
-    window.localStorage.getItem(TRANSCRIPT_EASED_FOLLOW_STORAGE_KEY),
-  );
-}
--- a/apps/packages/product-client/src/domain/chats/transcript/transcript-eased-follow-config.ts
+++ b/apps/packages/product-client/src/domain/chats/transcript/transcript-eased-follow-config.ts
@@ -1,0 +1,22 @@
+/**
+ * Flag discipline for the eased-follow motion writer (PRO-168, design question
+ * Q16, rung 12). Follows the same localStorage-flag convention as
+ * transcript-virtualization-config.ts: an explicit opt-in string, read once at
+ * mount, defaulting OFF so v1's instant glue stays byte-identical unless a
+ * reader deliberately turns the prototype on.
+ */
+export const TRANSCRIPT_EASED_FOLLOW_STORAGE_KEY = "proliferate:transcriptEasedFollow";
+
+export function resolveTranscriptEasedFollowEnabled(value: string | null): boolean {
+  return value === "on";
+}
+
+/** Read the flag once at mount (same convention as the virtualization-mode flag). */
+export function readTranscriptEasedFollowEnabled(): boolean {
+  if (typeof window === "undefined") {
+    return false;
+  }
+  return resolveTranscriptEasedFollowEnabled(
+    window.localStorage.getItem(TRANSCRIPT_EASED_FOLLOW_STORAGE_KEY),
+  );
+}
--- a/apps/packages/product-client/src/hooks/chat/ui/transcript-eased-follow.test.ts
+++ b/apps/packages/product-client/src/hooks/chat/ui/transcript-eased-follow.test.ts
@@ -1,9 +1,17 @@
-import { describe, expect, it } from "vitest";
+/* @vitest-environment jsdom */
+
+import { afterEach, describe, expect, it } from "vitest";
+import { TRANSCRIPT_EASED_FOLLOW_STORAGE_KEY } from "#product/domain/chats/transcript/transcript-eased-follow-config";
 import {
+  readTranscriptEasedFollowEnabled,
   resolveEasedFollowStep,
   TRANSCRIPT_EASED_FOLLOW_CONVERGE_PX,
   TRANSCRIPT_EASED_FOLLOW_RATE,
 } from "#product/hooks/chat/ui/transcript-eased-follow";
+
+afterEach(() => {
+  window.localStorage.removeItem(TRANSCRIPT_EASED_FOLLOW_STORAGE_KEY);
+});
 
 describe("resolveEasedFollowStep (PRO-168, rung 12)", () => {
   it("steps toward the target by the fixed rate, not converged, while the remaining distance is large", () => {
@@ -46,5 +54,16 @@ describe("resolveEasedFollowStep (PRO-168, rung 12)", () => {
     expect(top).toBe(target);
     // At a 0.25/frame rate the geometric tail converges well under 200 frames.
     expect(frames).toBeLessThan(60);
+  });
+});
+
+describe("readTranscriptEasedFollowEnabled (PRO-168, rung 12, Q16)", () => {
+  it("reads OFF when the key is unset", () => {
+    expect(readTranscriptEasedFollowEnabled()).toBe(false);
+  });
+
+  it("reads ON once the flag is explicitly set", () => {
+    window.localStorage.setItem(TRANSCRIPT_EASED_FOLLOW_STORAGE_KEY, "on");
+    expect(readTranscriptEasedFollowEnabled()).toBe(true);
   });
 });
--- a/apps/packages/product-client/src/hooks/chat/ui/transcript-eased-follow.test.ts
+++ b/apps/packages/product-client/src/hooks/chat/ui/transcript-eased-follow.test.ts
@@ -1,0 +1,50 @@
+import { describe, expect, it } from "vitest";
+import {
+  resolveEasedFollowStep,
+  TRANSCRIPT_EASED_FOLLOW_CONVERGE_PX,
+  TRANSCRIPT_EASED_FOLLOW_RATE,
+} from "#product/hooks/chat/ui/transcript-eased-follow";
+
+describe("resolveEasedFollowStep (PRO-168, rung 12)", () => {
+  it("steps toward the target by the fixed rate, not converged, while the remaining distance is large", () => {
+    const step = resolveEasedFollowStep(0, 1000);
+    expect(step.converged).toBe(false);
+    expect(step.nextTop).toBeCloseTo(1000 * TRANSCRIPT_EASED_FOLLOW_RATE);
+  });
+
+  it("eases from the other direction identically (symmetric in delta sign)", () => {
+    const step = resolveEasedFollowStep(1000, 0);
+    expect(step.converged).toBe(false);
+    expect(step.nextTop).toBeCloseTo(1000 - 1000 * TRANSCRIPT_EASED_FOLLOW_RATE);
+  });
+
+  it("converges by snapping directly once within the convergence epsilon", () => {
+    const almostThere = 500 + (TRANSCRIPT_EASED_FOLLOW_CONVERGE_PX - 0.1);
+    const step = resolveEasedFollowStep(almostThere, 500);
+    expect(step.converged).toBe(true);
+    expect(step.nextTop).toBe(500);
+  });
+
+  it("reports converged immediately when already exactly at the target", () => {
+    const step = resolveEasedFollowStep(240, 240);
+    expect(step.converged).toBe(true);
+    expect(step.nextTop).toBe(240);
+  });
+
+  it("repeated application converges to the target within a bounded number of frames", () => {
+    let top = 0;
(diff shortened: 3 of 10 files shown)


pablonyx:
## Manual verification (rung 12, PRO-168/PRO-187)

Scoped vitest, all files touched by this rung, literal output:

```
 RUN  v3.2.7 apps/packages/product-client

 ✓ src/domain/chats/transcript/transcript-eased-follow-config.test.ts (3 tests) 1ms
 ✓ src/hooks/chat/ui/transcript-frame-pipeline.test.ts (11 tests) 2ms
 ✓ src/hooks/chat/ui/transcript-eased-follow.test.ts (7 tests) 2ms
 ✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.eased-follow.test.tsx (4 tests) 13ms
 ✓ src/hooks/chat/ui/use-transcript-stick-to-bottom.test.tsx (23 tests) 28ms

 Test Files  5 passed (5)
      Tests  48 passed (48)
```

`use-transcript-auto-follow-bottom.ts` and `use-transcript-frame-pipeline-lifecycle.ts` have no
dedicated test files; they are exercised indirectly through the
`use-transcript-stick-to-bottom*.test.tsx` suite (all 27 tests above, flag-off and flag-on).

### Negative control

Reverted `resolveEasedFollowStep` to snap straight to the target instead of taking a fractional
step (the core of the eased-motion change), then reran the two files that exercise it directly:

```
 × resolveEasedFollowStep > steps toward the target by the fixed rate, not converged, while the remaining distance is large
   → expected 1000 to be close to 250, received difference is 750, but expected 0.005
 × resolveEasedFollowStep > eases from the other direction identically (symmetric in delta sign)
   → expected +0 to be close to 750, received difference is 750, but expected 0.005
 × useTranscriptStickToBottom eased-follow writer > steps toward the target instead of snapping instantly on the first growth frame
   → expected 1700 to be close to 950, received difference is 750, but expected 0.005

 Test Files  2 failed (2)
      Tests  3 failed | 8 passed (11)
```

Restored the fix; reran the same two files, clean:

```
 Test Files  2 passed (2)
      Tests  11 passed (11)
```

`git status`/`git diff` confirm the working tree matches the pushed commit after restore (no diff).

### CI

All checks on the head commit are green, including "Transcript scroll physics (chromium + webkit)".
That job needed 5 reruns to settle. Root-caused the repeated Q13 failure (`Expected: >= 2603,
Received: 2562`, `scroll-physics.spec.ts:1100`) as pre-existing runner-timing flakiness, NOT a
rung-12 regression: reran the equivalent job on the r11 base PR (#1984, zero rung-12 code) and it
failed with the *identical* assertion and the *identical* numbers (`Expected: >= 2603, Received:
2562`). The rung-12 pinned-branch code is provably a no-op on the flag-off path this test exercises
(the `!easedFollowEnabled` branch calls the same `scrollToBottom()` as before; `hasPendingMotion`
short-circuits to `false` whenever the flag is off), consistent with that root cause.

pablonyx:
## Review round 1: CLEAN

Reviewed diff `11cfa279a..55be94a03` (base r11-diagnostics → head r12-eased-follow) against ADR §6 rung 12 / Founder Rulings FR-1. No blocking findings. One informational risk noted below (non-blocking, flag defaults OFF).

### FR-1 single-writer — verified
- `transcript-eased-follow.ts` (`resolveEasedFollowStep`) is pure/stateless, no DOM, no timers.
- `transcript-frame-pipeline.ts`'s `continueMotionIfPending` schedules its continuation through `this.raf` (the pipeline's existing injected raf), armed into a dedicated `motionHandle` that is folded into `beginGlue`'s reset and cleared in `stop()` — confirmed no second `requestAnimationFrame`/`setInterval`/CSS-transition writer exists anywhere in the diff (`grep` for `setInterval|transition|scroll-behavior` in the diff returns nothing).
- Mid-glide interaction with above-change compensation/restore is deterministic by construction, not by luck: `runFramePass` branches exclusively on `pinnedRef.current` — pinned does eased-follow-or-instant and returns; unpinned (compensation/restore) is a separate, mutually exclusive branch that explicitly clears `easedMotionPendingRef` on entry. The two writers can never run in the same pass.
- `hasPendingMotion` is gated on `pinnedRef.current` (round-3 Greptile fix already in this head, commit message confirms: "gate hasPendingMotion on pin state") — verified this prevents the stale-true leak the implementer describes; traced the logic, it's correct.

### Flag-off purity — verified
- Flag is read exactly once, at mount, via `useState(readTranscriptEasedFollowEnabled)` in `use-transcript-stick-to-bottom.ts` — not per-frame, not per-pass. No localStorage read in the hot path.
- Flag-off pinned branch is the literal original line (`scrollToBottom(); return;`) with no reordering before it.
- Ran the pre-existing engine suite in a throwaway worktree (`/tmp/r12-review-wt`, built dist, `npx vitest run`): all 9 pre-existing `use-transcript-stick-to-bottom*.test.tsx` files / 63 tests pass unchanged, plus the 4 new test files / 25 new tests, for 87 total green.

### Convergence — proven for the fixed-rate model, informational risk flagged
- `resolveEasedFollowStep` terminates: geometric decay at RATE=0.25 with a 1px snap epsilon, unit test proves bounded-frame convergence (<60 frames to close a 4000px gap) — confirmed passing.
- Negative control run (reverted `continueMotionIfPending` call sites): 4 tests fail with exact expected/received deltas (e.g. `expected 3 to be 2`, `expected 175 to be 700`) — this is a real fix, not vacuous coverage. Re-ran myself in the throwaway worktree with a different cut (forced the pinned branch to always take the instant path) and got the same class of failure (3/4 eased-follow tests fail with concrete scrollTop mismatches), confirming the flag-on coverage is exercising real behavior.
- **Risk (informational, not a defect):** with a fixed 25%-of-remaining-distance step, sustained monotonic growth reaches a steady-state lag of `growth_per_frame / RATE`. At the ladder's stated 132px/frame typewriter rate that is ~528px of permanent trailing lag while streaming, for as long as content keeps growing every frame. This is not tested at the physics level (no Playwright fixture — correctly disclosed by the implementer as a deviation + founder question #1), so it's unverified whether that steady lag is compatible with the ADR's pinned-follow bound. Since the flag defaults OFF and this is explicitly scoped as a prototype, I'd treat this as supporting evidence for founder question #1 (should a real-browser fixture gate the merge) rather than a blocking finding.

### Reduced motion — correctly left as a founder question, not invented
Grepped the ADR and Founder Rulings for `prefers-reduced-motion` / "reduced motion" — zero matches in either. The eased writer has no such check. Since the spec is silent, this isn't a spec violation; flagging as a legitimate open question for whoever owns the founder Q&A (not raised in the PR body's founder-questions list — worth adding as a third item).

### Fold-claim verification
Confirmed `use-above-change-compensation.ts` does not exist at base `11cfa279a` either (`git cat-file -e` fails identically at base and head). `git log` on that path shows the fold happened in `7b43facfe` ("one per-frame mutate-then-snap scroll pipeline"), a prior rung — not silently folded in this diff. Implementer's deviation note is accurate.

### Repo shape / CI
`gh pr view` shows both `release:minor-feature` and `area:product` labels present, all CI checks green including "Repo shape checks", "Cargo check & test", and "Transcript scroll physics (chromium + webkit)" on this head SHA (55be94a03). Did not ask for the pre-existing Q13 interleave flake (2603 vs 2562) to be addressed — confirmed it's a base-r11 marginality, not introduced or worsened here — physics job passed on this PR's own CI run.

### Test honesty
No loosened thresholds, no `.skip`/`.only`/fixme found in the diff. Negative control reproduced independently (see above) with a different revert than the implementer used, same conclusion.

**Verdict: CLEAN.** No changes requested. Suggest the two existing founder questions (Playwright fixture, rate/epsilon tuning) plus reduced-motion as a third open question before any wider (non-flag) rollout.

pablonyx:
Rebased onto the corrected r10/r11 heads (bimodal Q13 physics fix, see #1980 for the full root-cause writeup). tsc (domain + main) clean at this branch's new head. Q13 itself has passed on every run/rerun observed during verification (8 consecutive passes across separate CI invocations).

One unrelated, pre-existing webkit flake surfaced repeatedly on reruns: "prepend anchoring: older-history prepend keeps the reading row fixed" (`Expected > 150, Received 0`). It's untouched by this diff's changes and reproduces on fresh workflow runs independent of the Q13 fix — flagging for separate triage.

pablonyx:
Stack refold receipt (overnight staging): rebased onto main @ d5d38a804 (r1 merged) atop the retargeted stack. git range-diff vs previous head: 0 semantic changes (10/10 commits identical). New head 846bf9696 (stack top). Union typecheck (this head + pending cancelled-tone line) PASSES locally. Review carried from 60b5e2188.

pablonyx:
## Delta review of post-review commits at 846bf9696: CLEAN

Reviewed the four commits that landed after the CLEAN review at (pre-cascade) 55be94a03, diffed here as `0bfa0354f..846bf9696` on the current (post-cascade, byte-identical) branch:

- `c76a64f93` CI round 1 — move window access out of domain/
- `fbed92632` CI round 2 — relocate flag reader off the 400-line gate
- `18af2c616` CI round 3 — gate `hasPendingMotion` on pin state
- `846bf9696` chore — trim comments under the 400-line gate

**(a) CI round 3 — the only behavioral change.** `hasPendingMotion` now reads `easedFollowEnabled && pinnedRef.current && easedMotionPendingRef.current` (`use-transcript-frame-pipeline-lifecycle.ts:343`), and the unpinned branch explicitly zeroes the ref as defense in depth (`use-transcript-frame-pipeline-lifecycle.ts:170`). Verified the single-writer invariant still holds structurally: `runFramePass` (`use-transcript-frame-pipeline-lifecycle.ts:141-164`) takes the pinned branch (eased step or instant `scrollToBottom`) and returns before ever reaching the unpinned branch's compensation-anchor logic (`:225-315`) — eased motion and above-change compensation remain mutually exclusive by the same `if (pinnedRef.current) { ...; return; }` structure as before, unchanged by this commit. Confirmed no stale-pending-motion jump on repin: the ref is only a signal for `hasPendingMotion`/frame scheduling, never fed back into the step computation — `resolveEasedFollowStep` recomputes fresh from the live `viewport.scrollTop` and `resolveFollowTargetTop` every pass, so clearing/leaving-stale the ref has no effect on the actual scrollTop math on repin. Added regression test (`use-transcript-stick-to-bottom.eased-follow.test.tsx`) directly exercises unpin-mid-catch-up → resize → asserts the pipeline goes fully quiet; reasoning traces cleanly against `transcript-frame-pipeline.ts`'s `continueMotionIfPending` (re-checks `hasPendingMotion` at fire time before rearming, `:216`).

**(b) CI round 1 — domain/ DOM-free gate.** `transcript-eased-follow-config.ts` is back to pure string parsing (`resolveTranscriptEasedFollowEnabled`), no `window`/`localStorage`. Confirmed no runtime DOM/window/localStorage access remains anywhere under `src/domain/` (only an unrelated string literal and a stale comment mention). `readTranscriptEasedFollowEnabled` ended up in the hook layer (`transcript-eased-follow.ts`, matching round 2's rationale of colocating with the step function it gates), consistent with the domain layer's existing platform-free convention.

**(c) Comment-only trim commit.** Diffed `846bf9696~1..846bf9696` and filtered to non-comment lines — zero non-comment `+`/`-` lines; every changed line is a `//` comment. No code semantics touched.

**(d) Flag-off byte-identical path.** `easedFollowEnabled` is read once via `useState(readTranscriptEasedFollowEnabled)` (read-only localStorage check, no side effect on behavior). With the flag off, the pinned branch's `if (!easedFollowEnabled) { scrollToBottom(); return; }` is the exact same call as pre-rung-12. No new work is reachable when the flag is absent.

**Repo-shape gate**: ran `scripts/report_frontend_structure.py --strict` locally — `TOTAL: 0`. Touched files sit at 398 lines (`use-transcript-stick-to-bottom.ts`, `use-transcript-frame-pipeline-lifecycle.ts`), under the 400-line `LARGE_FRONTEND_FILE` gate.

No findings. Static review only (no vitest run — `node_modules` not present in the throwaway clone; no playwright/cargo/Docker per scope).

pablonyx:
r12 refold receipt (final rung): 846bf9696 -> 9bdb5d9c5, rebased --onto main-with-#2021 (e321fb51d) from r11 old head d41bc46a8. Range-diff 10/10 = (byte-identical carry; both flagged overlap files — lints/frontend/ratchets.toml and transcript-row-height-estimate.ts — applied without conflict; the known LOW quirk in the latter is untouched per order, named follow-up). Composed tree pre-verified: structure checker TOTAL 0, VirtualizedTranscriptRowList within its 407 ceiling. Fresh CI on this head is the merge gate.

pablonyx:
r12 MERGE RECEIPT (train complete): squash bc986d7fc, head 9bdb5d9c5 (match-head-commit enforced). Final board 24/24 pass; sole first-round fail was the ride-flagged [webkit] prepend-anchoring strand (chromium 17/17 green), which greened on its one same-ref retry — no attribution consumed. Budget PASSED at JS 499198/502000 B. Base verified from job checkout line: Merge 9bdb5d9c5 into e321fb51d (main with #2021). Carry byte-identical 10/10 =. This closes the PRO-187 sequential refold train: r2-r12 all merged.

PR #1997, opened 2026-08-16
fix(anyharness): let live sessions outlast product MCP capability-token TTL

pablonyx:
## Summary

Product MCP capability tokens (workspace/cowork/reviews tools) are minted once at session launch, delivered to the agent as a **static header** baked into the session's MCP config, and expired after a fixed 12h TTL. Nothing ever re-minted them, so any session older than 12h lost every product MCP tool — and the failure surfaced as garbage: the 401 sent the MCP client into OAuth discovery against the runtime root, and axum's empty 404 for the missing well-known routes became `HTTP 404: Invalid OAuth error response: JSON Parse error: Unexpected EOF`. The 401s also logged nothing server-side. (Live incident: orchestrator session `554b8edc`, launched 02:13Z, token expired 14:13Z, dead from 14:24Z onward.)

Three fixes, one family:

1. **Sessions no longer outlive their tokens.** Token validation now returns a reasoned verdict (`Valid | Expired | InvalidSignature | ScopeMismatch`) instead of a boolean. An **expired-but-authentic** token — signature and workspace/session/product scope verified — defers to the sessions domain: `SessionService::session_open_for_capability` accepts it while the session row exists, matches the workspace, and is not closed. A session that is alive and prompting at hour 30 keeps working; the moment it closes (or its row is deleted), the expired token dies.
2. **Error masking killed.** Every capability rejection is now **403** with a cause-naming RFC 7807 body, never 401. Missing header, malformed/bad-signature, scope mismatch, and expired-with-closed-session each get a distinct `detail` naming the real cause.
3. **Auth failures are observable.** Every rejection emits the named tracing event `anyharness.product_mcp.auth.rejected` (target `anyharness.product_mcp_auth_rejected`, mapped in telemetry setup per the #1849/#1850 named-event lineage) with `session_id`, `workspace_id`, `slug`, and `reason ∈ {missing, expired, invalid-signature, scope-mismatch}` — and never the token value. Expired-but-live grace acceptances additionally log at debug.

### Why session-liveness (and not the other candidates)

- **Re-mint + re-inject on resume/reconfigure**: cannot help a session that never resumes — the observed incident is a continuously-running session crossing the TTL mid-flight; refreshing the binding requires a session reconfigure the client never performs. Highest complexity, still leaves the exact observed gap.
- **Lifetime-scaled TTL + revoke-on-close**: any fixed TTL re-creates the same cliff further out, and revocation needs a new revocation store. The sessions table already *is* the revocation store — which is what this PR uses.
- **Chosen — validate against session liveness, exp as defense-in-depth**: the token already cryptographically binds workspace+session+product (HMAC-SHA256, constant-time compare, unchanged), and the domain layer re-verifies session↔workspace on every call (`verify_caller_workspace`). The honest freshness question is "is the bound session still open", answered from durable state on the expired path only; the fresh-token fast path is unchanged.

**Leaked-token threat model.** The server binds loopback-only, and the signing key lives at `<runtime_home>/secrets/workspace-mcp-token.key` (0600). Any local attacker positioned to steal a token can read the key and mint arbitrary tokens, so token theft adds no marginal capability. Today a stolen token replays for up to 12h regardless of session state; after this change an expired token dies with its session, and a fresh one still expires within 12h if the session row vanishes. Strictly narrower. A "recently active" recency gate was considered and rejected: it defends nothing in this threat model and would kill tools mid-turn for sessions idling through long builds. Dismissal is hiding, not termination, so dismissed sessions keep their capability until closed.

### Client-side proof for the 401→403 switch

From the shipped Claude Code binary (v2.1.212, MCP transport `send()`, extracted via `strings`):

```js
if (!n.ok) {
  let o = await n.text().catch(() => null);
  if (n.status === 401 && this._authProvider) {
    let { resourceMetadataUrl: i, scope: s } = uzt(n);
    ...OAuth discovery / authorization flow...
  }
  throw Error(`Error POSTing to endpoint (HTTP ${n.status}): ${o}`);
}
```

Only `401` diverts into OAuth discovery (the masking path). Any other status throws with the response body verbatim, so the client-visible error becomes e.g. `Error POSTing to endpoint (HTTP 403): {"title":"Forbidden","status":403,"detail":"Product MCP capability token expired and its session is no longer open. Start or resume the session to mint a fresh token.","code":"WORKSPACE_MCP_UNAUTHORIZED"}` — the real cause, named.

## Testing

Tier 1 (unit/integration in-crate) per specs/TESTING.md — the behavior is a pure server-side auth rule, fully exercisable against real `AppState` + in-memory SQLite and the real Workspace MCP endpoint registry; no sandbox or live agent needed.

- **Regression (a session older than the TTL still validates):** `app::tests::product_mcp_auth_tests::expired_capability_token_keeps_serving_an_open_session` — real `AppState`, seeded workspace + `idle` session, token minted with the same secret file but already-past expiry (negative TTL injected via the issuer constructor — no sleeps), `tools/list` dispatch returns 200.
- Closed-session rejection: `expired_capability_token_is_forbidden_once_the_session_closes` (403, detail contains "expired", code `WORKSPACE_MCP_UNAUTHORIZED`).
- Missing header: `missing_capability_token_is_forbidden_with_named_cause` (403, detail names the header).
- Garbage token with an OPEN session: `garbage_capability_token_is_forbidden_even_for_an_open_session` (liveness never rescues an unauthentic token).
- Verdict-level units: injected-clock expiry (`product_token_expiry_is_judged_against_the_injected_clock`: valid at TTL−60s, expired at hour 30), tamper matrix (`tampered_product_tokens_are_invalid_not_expired`), scope mismatch, expired-vs-invalid separation at the auth wrapper (`expired_tokens_report_expired_not_invalid`), and the pure liveness predicate (open statuses / closed / foreign workspace / dismissed).
- Telemetry mapping: `anyharness_targets_resolve_to_record_names_by_table_then_pass_through` extended with the new target.

Results (local, macOS, scrubbed of leaked `PROLIFERATE_*`/`ANYHARNESS_*` env):

- `cargo test -p anyharness-lib` → `2094 passed; 1 failed` — the one failure is `route_auth::...::the_sweep_removes_only_abandoned_and_old_scratch_roots`, **pre-existing on this machine**: it fails identically at pristine `origin/main` (`caf288f57`, verified by checkout + rerun) and is untouched by this diff (mtime-sweep behavior, passes on Linux CI). The dispatch tests were subsequently moved to `app/tests/product_mcp_auth_tests.rs` to stay under the PROD-SIZE-1 line cap; all 4 re-verified green at the final head.
- `cargo test -p anyharness` → `33 passed; 0 failed`.

Negative controls (each fix reverted, watched fail, restored):

1. Expired arm hard-rejects (liveness ignored) → `expired_capability_token_keeps_serving_an_open_session` **fails**.
2. `forbidden` flipped back to `unauthorized` → all three 403 assertions **fail**.
3. Telemetry `TargetMapping` removed → mapping test **fails**.

## Observability

New named event per the runtime named-event lineage: constant `PRODUCT_MCP_AUTH_REJECTED_TRACING_TARGET` (`anyharness.product_mcp_auth_rejected`) in `observability/mod.rs`, mapped to diagnostics record `anyharness.product_mcp.auth.rejected` in `telemetry.rs`, emitted at `warn` on every capability rejection with `session_id`, `workspace_id`, `slug`, `reason`. Token values never appear in any log call (secret-log check passes). One grep now answers what took hours of forensics tonight.

## Readiness

  in this PR.

## Verification

- `python3 scripts/check_docs.py` → passed (226 files); `check_design_attribution.py`, `check_agent_auth_secret_logs.py` → passed.
- Touched files are rustfmt-clean (`rustfmt --check` per file).
- Follow-ups recorded, not expanded here: `mint_workspace_session_token`/`validate_workspace_session_token` and the `LegacySha256Dot` signature variant have no production callers outside `capability_token.rs` (dead-code candidates); the runtime could additionally serve minimal OAuth well-known metadata for clients that still probe it.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/api/http/product_mcp.rs
+++ b/anyharness/crates/anyharness-lib/src/api/http/product_mcp.rs
@@ -9,10 +9,12 @@ use serde_json::Value;
 use super::access::assert_workspace_mutable;
 use super::error::ApiError;
 use crate::app::AppState;
+use crate::integrations::mcp::capability_token::McpCapabilityTokenValidation;
 use crate::integrations::mcp::product_server::{
     ProductMcpAuthHeader, ProductMcpContextError, ProductMcpDispatchError,
     ProductMcpEndpointOperation, ProductMcpRequestContext, PRODUCT_MCP_TOKEN_HEADER_NAME,
 };
+use crate::observability::PRODUCT_MCP_AUTH_REJECTED_TRACING_TARGET;
 
 pub async fn get_product_mcp_endpoint(
     State(_state): State<AppState>,
@@ -53,20 +55,68 @@ pub async fn dispatch_product_mcp(
     let definition = server.definition();
     let request = ProductMcpRequestContext::new(workspace_id, session_id, definition.id);
     let endpoint_operation = ProductMcpEndpointOperation::from_request_body(&body);
-    let auth_header = read_auth_header(&headers).ok_or_else(|| {
-        ApiError::unauthorized(
-            "Missing product MCP capability token.",
+    let Some(auth_header) = read_auth_header(&headers) else {
+        return Err(reject_capability(
+            &request,
+            product_mcp_slug,
+            "missing",
+            "Product MCP capability token missing: the request did not carry the \
+             x-anyharness-product-mcp-token header this endpoint requires.",
             definition.unauthorized_code,
-        )
-    })?;
+        ));
+    };
     let validation = server
         .validate_capability_token(auth_header, &request)
         .map_err(|error| ApiError::internal(error.to_string()))?;
-    if !validation.is_valid() {
-        return Err(ApiError::unauthorized(
-            "Invalid product MCP capability token.",
-            definition.unauthorized_code,
-        ));
+    match validation {
+        McpCapabilityTokenValidation::Valid => {}
+        // The token is delivered as a static header for the life of the
+        // session, so a session that outlives the TTL arrives here on every
+        // call. The signature already proved the runtime minted this exact
+        // scope; the sessions domain, not the timestamp, decides whether the
+        // capability is still alive.
+        McpCapabilityTokenValidation::Expired => {
+            let session_open = state
+                .session_service
+                .session_open_for_capability(session_id, workspace_id)
+                .map_err(|error| ApiError::internal(error.to_string()))?;
+            if !session_open {
+                return Err(reject_capability(
+                    &request,
+                    product_mcp_slug,
+                    "expired",
+                    "Product MCP capability token expired and its session is no longer open. \
+                     Start or resume the session to mint a fresh token.",
+                    definition.unauthorized_code,
+                ));
+            }
+            tracing::debug!(
+                session_id = %request.session_id,
+                workspace_id = %request.workspace_id,
+                slug = product_mcp_slug,
+                "product MCP capability token past TTL accepted for open session"
+            );
+        }
+        McpCapabilityTokenValidation::InvalidSignature => {
+            return Err(reject_capability(
+                &request,
+                product_mcp_slug,
+                "invalid-signature",
+                "Product MCP capability token is not valid for this runtime: the token is \
+                 malformed or its signature does not verify.",
+                definition.unauthorized_code,
+            ));
+        }
+        McpCapabilityTokenValidation::ScopeMismatch => {
+            return Err(reject_capability(
+                &request,
+                product_mcp_slug,
+                "scope-mismatch",
+                "Product MCP capability token is not scoped to this workspace, session, and \
+                 product endpoint.",
+                definition.unauthorized_code,
+            ));
+        }
     }
 
     let _lease = match server.endpoint_operation_kind(endpoint_operation) {
@@ -90,6 +140,31 @@ pub async fn dispatch_product_mcp(
         Some(payload) => Ok((StatusCode::OK, Json(payload)).into_response()),
         None => Ok(StatusCode::NO_CONTENT.into_response()),
     }
+}
+
+/// Reject a capability header with the one observable trace of the failure.
+///
+/// The status is 403, not 401, on purpose: MCP clients treat 401 as an OAuth
+/// challenge and run authorization-server discovery against the runtime root,
+/// which fails with an unrelated parse error that masks the real cause. A 403
+/// body passes through to the client verbatim, so the detail names the cause.
+/// The event carries identifiers and the reason — never the token.
+fn reject_capability(
+    request: &ProductMcpRequestContext,
+    slug: &str,
+    reason: &'static str,
+    detail: &'static str,
+    unauthorized_code: &str,
+) -> ApiError {
+    tracing::warn!(
+        target: PRODUCT_MCP_AUTH_REJECTED_TRACING_TARGET,
+        session_id = %request.session_id,
+        workspace_id = %request.workspace_id,
+        slug,
+        reason,
+        "product MCP capability token rejected"
+    );
+    ApiError::forbidden(detail, unauthorized_code)
 }
 
 fn map_dispatch_error(error: ProductMcpDispatchError, request_invalid_code: &str) -> ApiError {
--- a/anyharness/crates/anyharness-lib/src/app/tests.rs
+++ b/anyharness/crates/anyharness-lib/src/app/tests.rs
@@ -16,6 +16,7 @@ use crate::{
 };
 
 mod completion_delivery_crash_tests;
+mod product_mcp_auth_tests;
 
 fn run_git(path: &Path, args: &[&str]) {
     let output = Command::new("git")
--- a/anyharness/crates/anyharness-lib/src/app/tests/product_mcp_auth_tests.rs
+++ b/anyharness/crates/anyharness-lib/src/app/tests/product_mcp_auth_tests.rs
@@ -1,0 +1,214 @@
+//! Dispatch-level tests for product MCP capability auth: expired tokens defer
+//! to session liveness, every rejection is a 403 naming the real cause.
+
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::sync::Mutex;
+
+use serde_json::json;
+
+use super::init_seed_repository;
+use crate::app::{test_support, AppState};
+use crate::{
+    api::http::product_mcp::dispatch_product_mcp,
+    domains::{agents::installer::seed::AgentSeedStore, sessions::store::SessionStore},
+    integrations::mcp::capability_token::{
+        McpCapabilityTokenIssuer, McpCapabilityTokenSignature, ProductMcpCapabilityScope,
+    },
+    integrations::mcp::product_server::PRODUCT_MCP_TOKEN_HEADER_NAME,
+    persistence::Db,
+};
+
+/// AppState over an in-memory db with one seeded workspace and one session in
+/// the given status, ready for product MCP dispatch against the Workspace
+/// endpoint. Returns the runtime home so the caller can mint tokens against
+/// the same secret file and clean up.
+fn product_mcp_dispatch_fixture(label: &str, session_status: &str) -> (AppState, PathBuf, PathBuf) {
+    let runtime_home = PathBuf::from(format!("/tmp/anyharness-{label}-{}", uuid::Uuid::new_v4()));
+    let repository_path = init_seed_repository(&PathBuf::from(format!(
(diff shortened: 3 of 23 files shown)


pablonyx (commented):
## Per-PR review — #1997 (head `20b2c637d`, verified unchanged)

**Verdict: clean.** The auth logic is sound, the expired→liveness deferral is
safe, verdict ordering is correct, telemetry is wired end-to-end, and the tests
(including negative controls) genuinely exercise the change. Only minor
test-hygiene / confirmation nitpicks below — nothing blocking. Approval is
Pablo's.

I read the full diff plus the surrounding code at the SHA; no build was run.

### Security review of the expired → liveness deferral

The core question — *can a closed/foreign/deleted session ever pass, and can
liveness ever rescue an unauthentic token?* — resolves **no** on every path:

- **Verdict ordering is correct: signature → scope → expiry.** In
  `validate_product_mcp_token_at` ([capability_token.rs](anyharness/crates/anyharness-lib/src/integrations/mcp/capability_token.rs)),
  malformed parts / bad base64 / HMAC mismatch / undecodable payload all return
  `InvalidSignature` *before* the scope comparison, which returns `ScopeMismatch`
  *before* the `payload.exp < now` check returns `Expired`. So the only way to
  reach `Expired` is an authentic token whose scope already matches this exact
  workspace/session/product. Constant-time compare (`ct_eq`) is preserved.
- **Liveness is only consulted on the `Expired` arm.** In
  [product_mcp.rs:70-99](anyharness/crates/anyharness-lib/src/api/http/product_mcp.rs),
  `InvalidSignature` and `ScopeMismatch` reject unconditionally; only `Expired`
  calls `session_open_for_capability`. An unauthentic or mis-scoped token can
  never reach the liveness check. The `garbage_capability_token_...` test nails
  this with an open session.
- **The liveness gate is closed on every dimension.**
  `session_record_open_for_capability` ([capability.rs:28](anyharness/crates/anyharness-lib/src/domains/sessions/service/capability.rs))
  requires `workspace_id` match **and** `status != "closed"` **and**
  `closed_at.is_none()`. Foreign-workspace rows fail the first clause; the
  canonical close signal (`mark_closed` sets `closed_at = COALESCE(...)`,
  store/sessions.rs:273) fails the third regardless of status; and a *deleted*
  session is a hard `DELETE FROM sessions` (deletion.rs), so `find_by_id` → `None`
  → `false`. The unit tests cover open statuses, closed, foreign workspace, and
  dismissed-kept-alive.
- **Single live enforcement point.** Grep confirms the only consumer that
  *acts* on the verdict is the HTTP dispatcher; every domain
  `validate_capability_*` just forwards the verdict up. No stale boolean gate
  survives that would reject `Expired` before the deferral (or accept it
  without the liveness check). All product MCP domains (agent_operations,
  cowork, reviews) and the `product_registry` adapter were migrated to the new
  enum; no `ProductMcpTokenValidation` references remain anywhere in the tree.

### 403 semantics, layering, telemetry, B20

- **403 not 401** is applied on all four reject paths via `ApiError::forbidden`
  ([error.rs:91](anyharness/crates/anyharness-lib/src/api/http/error.rs)), which
  emits a proper RFC 7807 body with a cause-naming `detail` and the endpoint's
  `unauthorized_code`. The PR's client-side `strings` proof that only 401
  diverts Claude Code into OAuth discovery is the correct justification.
- **Layering holds.** `capability_token` and `product_server` (integrations)
  gain no domain imports; the enum lives in integrations and domains depend
  down onto it. The sessions-domain consult happens in the API layer
  (`product_mcp.rs`), which is allowed to reach both domains and integrations.
- **Telemetry is wired on both sides:** constant in
  [observability/mod.rs](anyharness/crates/anyharness-lib/src/observability/mod.rs),
  `TargetMapping::span_event` in [telemetry.rs](anyharness/crates/anyharness/src/telemetry.rs),
  and the mapping test bumped `7 → 8` with the new case asserted.
- **B20 / secret-log:** every `warn!`/`debug!` in the reject path carries only
  `session_id`, `workspace_id`, `slug`, `reason` — never the token. The only
  error string surfaced (`ApiError::internal(error.to_string())`) originates
  from secret-file IO, not token content.
- **Test quality:** expiry is exercised via injected clock
  (`validate_product_mcp_token_at`) and via negative-TTL minting — no sleeps.
  The integration regression runs against real `AppState` + in-memory SQLite +
  the real endpoint registry. Negative controls are documented per-fix in the
  body.

### Nitpicks (non-blocking)

1. **Test cleanup asymmetry** —
   `garbage_capability_token_is_forbidden_even_for_an_open_session`
   ([app/tests.rs](anyharness/crates/anyharness-lib/src/app/tests.rs)) removes
   only `runtime_home` and omits `let _ = fs::remove_dir_all(repository_path);`,
   which the other three `product_mcp_dispatch_fixture` tests all do. Leaks one
   `/tmp/anyharness-garbage-token-repo-*` dir per run. Trivial.

2. **`completed` / `errored` keep the capability** — `open_statuses_keep_the_capability`
   asserts these two alongside `starting`/`idle`/`running`. This is consistent
   with the design (the true gate is `closed_at`, and a completed-but-not-closed
   session can still be prompted), but it's worth a one-line confirmation that
   serving product MCP to a `completed`/`errored` session is intended rather
   than incidental.

3. **Version bump bundled** — `0.4.14 → 0.4.15` across VERSION, both
   package.jsons, Cargo.toml, and tauri.conf.json rides along in this fix PR.
   Presumably intentional release coupling; flagging only so it's a conscious
   choice, not scope creep.

### Doc edits
Both `specs/**` edits (servers.md, mcp-runtime.md) accurately describe the
verdict, the defense-in-depth framing of the TTL, the 403-not-401 rule, and the
named event with its `missing | expired | invalid-signature | scope-mismatch`
reasons. No inaccuracy.

pablonyx (commented):
## Follow-up — CLEAN verdict carries to `e1daa8b22`

Re-verified via a true tree diff `20b2c637d..e1daa8b226` (the branch was rebased, so a GitHub compare shows the full cumulative diff; `git diff` between the two commit objects isolates the real delta). The delta is **exactly three things and nothing else**:

1. **Version bumps reverted** — `0.4.15 → 0.4.14` in `VERSION` + the four manifests, back to merge-base content (resolves prior nitpick #3).
2. **Tests relocated** — the four dispatch tests + fixtures moved verbatim from `app/tests.rs` into the new submodule `app/tests/product_mcp_auth_tests.rs` (self-contained imports; wired via `mod product_mcp_auth_tests;`); `tests.rs` import list trimmed accordingly.
3. **Prior nitpick #1 fixed** — `garbage_capability_token_is_forbidden_even_for_an_open_session` now also removes its `repository_path` temp dir.

Critically, the diff does **not** touch any auth-logic surface — `api/http/product_mcp.rs`, `integrations/mcp/capability_token.rs`, `sessions/service/capability.rs`, `product_server/**`, `observability/mod.rs`, `telemetry.rs`/`telemetry/tests.rs`, or the two `specs/**` docs are byte-identical to the reviewed head. The verdict ordering, expired→liveness deferral, 403 semantics, telemetry wiring, and B20 clean-log guarantees I verified at `20b2c637d` all hold unchanged.

**CLEAN verdict carries.** Approval remains Pablo's.

pablonyx:
## Merge receipt — today-owned merge wave (2026-08-16)

**Squash-merged to main as `c01ad7db35e9f1f93fa26365907cf35fe764a95b`** under Pablo's live merge ruling for tonight's owned PRs.

Same-chain pre-merge verification (literal):
- headRefOid re-verified unchanged: `e1daa8b226e7ace8089b01e97633ef60892471d2`
- mergeable=MERGEABLE
- CI attached to that exact head (`gh run list --commit <full-sha>`), tally: `6 completed success` — 0 fail / 0 pending (CI, CodeQL, Intent Tests, PR Metadata x2, Self-Host Smoke)

Adjudication summary:
- Prior dedicated opus review: CLEAN at 20b2c637d; delta verdict carried to e1daa8b22 (tree-diff verified: version-bump revert + test relocation + nitpick fix only).
- Fresh spec-conformance pass (dedicated opus reviewer, comment-only, frozen SHA): **CLEAN** on all three deliverables — liveness-over-TTL via session_open_for_capability (garbage tokens never rescued, fresh fast path untouched), 403 RFC 7807 cause-naming on all four reject paths (zero remaining 401s), named event anyharness.product_mcp.auth.rejected wired constant→TargetMapping with token-free fields.
- Constitution: no weakened lints, no net-new exceptions, no deleted pinning tests; docs updated in-PR consistent with behavior.

PR #1972, opened 2026-08-16
feat(product): draw workflows on a pannable graph canvas — three-pane builder page and run pane (PRO-288)

tairabun:
## Summary

- Both workflow graphs now draw on a shared pannable, zoomable dotted-grid canvas (`WorkflowCanvas`): fixed-size node cards joined by bezier edges with arrowheads (dashed for ad hoc branches), drag-to-pan, ⌘/Ctrl+wheel zoom, and corner controls with Fit. Geometry is a pure domain module (`graph-layout.ts`): one rank per chain slot, retries widening their rank, side nodes hanging in the lane beside their anchor.
- The builder is rebuilt as the design's full-bleed three-pane page: step palette (Agent / Human in the loop) and the context-docs roster in a left rail, the chain — headed by the structural input card — on the canvas in the middle, and a right inspector editing exactly the selected object: a step's fields, one context doc (slug, written by, starting body, remove), or the workflow itself (description, default repository, declared inputs) through the input card. The title moves to the top bar beside Save Workflow; node cards wear their model in catalog display names; the canvas carries a step-count/validity readout bottom-left. The previous vertically stacked Details/Steps/Inputs/Documents page and its docs panel are deleted.
- The run pane (right panel) draws the same canvas with per-node status dots, selection, and a docked inspector reusing the full node card (Approve / Fail & redo / flip / add side node, dialogs, and session hand-off). It also gains a run-status chip on the pane header, and the tab is renamed Workflow → Execution with the `Fork` mark, leading the strip for fresh workspaces.
- Workflows main page: one filter over every group, definitions under a "Saved Workflows" header, and a new Executions group over the runtime's run roster (status glyph, state + wall clock, opens the run's workspace).
- Includes the two commits of #1948 (merged in); the canvas presentation supersedes that PR's stacked-list edges while keeping its slot/branch semantics.

No contract changes: the chain is still the card order, and the canvas draws the implied edges without offering edge editing.

Closes PRO-288

## Testing

- Unit/component (vitest, jsdom): new `graph-layout.test.ts`; rewritten `WorkflowGraphView.test.tsx` (canvas cards, chain/branch edges, selection, zoom controls); `WorkflowBuilderSurface.test.tsx` updated to the three-pane selection flows plus new palette/auto-select cases; executions/filter cases added to `WorkflowsMainSurface.test.tsx`. 226 workflow-area tests plus the right-panel shell suites pass.
[...]

PR #2002, opened 2026-08-17
fix(product): cover the cancelled run status in RUN_STATUS_TONE

pablonyx:
Main is latently red: #1906 (23:23Z) added `cancelled` to `WorkflowRunStatusV2`, and #1972 (00:28Z) landed `run-view-model.ts` with a total `RUN_STATUS_TONE` map authored before that — the merges crossed, so `tsc -p tsconfig.domain.json` fails with TS2741 ("Property 'cancelled' is missing") in every job that builds the product-client domain (first seen on PR #1725's Self-Host Smoke after a branch update; main's own smoke at efe354aaa skipped the build as docs-only).

One-line fix: add `cancelled: "neutral"` to the map, mirroring `NODE_STATUS_TONE`'s existing `cancelled: "muted"` entry. `workflowRunStatusTone` already falls back to `neutral` for unknown statuses, so runtime behavior is unchanged — this only restores the total-declaration typecheck.

Found during the open-PR backlog review lane.

### DIFF
--- a/apps/packages/product-client/src/domain/workflows/run-view-model.ts
+++ b/apps/packages/product-client/src/domain/workflows/run-view-model.ts
@@ -78,6 +78,9 @@ const RUN_STATUS_TONE: Record<WorkflowRunV2["status"], WorkflowRunTone> = {
   interrupted: "warning",
   completed: "success",
   failed: "danger",
+  // Terminal-inert, same reasoning as NODE_STATUS_TONE's `cancelled`: a
+  // deliberate stop, not an error; matches the unknown-status fallback.
+  cancelled: "neutral",
 };
 
 export function workflowRunStatusTone(status: WorkflowRunV2["status"]): WorkflowRunTone {


pablonyx:
Tier-2 `Workflow definition lifecycle` failure on this PR is **not** the tone typing — it is a click interception:

- The spec times out clicking `#workflow-builder-input-1-required`. Playwright's log shows a toast (`section[aria-label="Notifications alt+T"]` subtree) intercepting pointer events for the full 120s, across 197 retries.
- The toast is the desktop-worker enrollment failure: headline **"Integrations unavailable"**, cause **`Cannot read properties of undefined (reading 'invoke')`** — `ensureDesktopWorker` running in the intent-test environment where the tauri `invoke` side is absent, raising the sticky error toast from `use-desktop-worker-enrollment.ts` (`showErrorToast`, no auto-dismiss).
- This toast almost certainly fired before #1972 too, but the old stacked builder page had no click target under the toast region. The new three-pane inspector places the declared-input `required` checkbox bottom-right, directly under it — so #1972's layout made a latent env problem test-fatal. Main was green one commit before #1972 (5d9613666) and red from #1972 on.

Fix options (either unblocks tier-2):
1. Spec-side: dismiss notifications before inspector clicks (the toast has a "Close" button), or `page.keyboard.press("Alt+T")`-style handling in a helper.
2. Env-side (better long-term): the intent stack's host should not expose a desktop bridge without a working tauri transport, so enrollment never runs there.

Evidence: run 31982657482 artifact `workflow-definition-lifecycle-traces`, `error-context.md` (page snapshot shows the toast listitem with that exact cause text). `intent-tests (provisional)` fails the same way.

pablonyx:
**Equivalence verdict vs #2003 (diff-level, verified at heads 2543ce7c3 here / 6bf9ba9da there):**

- Both PRs patch the same file (`apps/packages/product-client/src/domain/workflows/run-view-model.ts`) against the same base blob `27ad53f73`, adding the identical functional line to `RUN_STATUS_TONE`: `cancelled: "neutral"`. The compiled output is byte-equivalent; both cure the TS2741 that has been failing every product-client-domain compile on main since the #1906 × #1972 crossed merges.
- Differences: this PR adds an explanatory comment only; #2003 adds a unit test (`workflowRunStatusTone("cancelled") === "neutral"` in `run-view-model.test.ts`). #2003 is strictly better and is the author's own fix.
- Recommendation on record: merge #2003; this PR then closes as a duplicate. This PR's control experiment (main red at docs-only `efe354aaa` vs 11/12 green here) plus the CS-lane comment below remain the diagnosis evidence for the crossed-merge incident and the tier-2 toast root cause.

No action tonight — disposition is Pablo's in the morning.

pablonyx:
Closing as duplicate of #2003 (author's own fix, adds the unit test this PR lacked; functional line identical — cancelled: "neutral" on the same base blob 27ad53f73). This PR's control evidence (main red at README-only efe354aaa vs 11/12 green here) drove the diagnosis. Credit to the CS lane for the tier-2 toast root-cause posted in the comments here.

PR #2003, opened 2026-08-17
fix(product): restore cancelled workflow run tone

tairabun:
## Summary

- Restore the total ProductClient run-status tone map after cancelled joined the workflow run status union.
- Present cancelled runs with the established neutral tone.
- Pin the cancelled-to-neutral mapping in the owning domain test.

## Testing

Pre-fix proof reproduced TS2741 at run-view-model.ts: Property cancelled is missing from Record<WorkflowRunStatusV2, WorkflowRunTone>.

Passing proof:
- pnpm exec vitest run src/domain/workflows/run-view-model.test.ts (54 tests)
- pnpm --filter @proliferate/product-client typecheck
- pnpm --filter @proliferate/product-client build
- python3 scripts/check_max_lines.py
- python3 scripts/check_frontend_boundaries.py
- python3 scripts/check_design_attribution.py
- git diff --check efe354aaa985ae1172ff9c8f7948fd2e10489ba6..HEAD

## Observability

None. This restores a pure presentation mapping and adds no failure path, telemetry, logging, or runtime behavior.

### DIFF
--- a/apps/packages/product-client/src/domain/workflows/run-view-model.test.ts
+++ b/apps/packages/product-client/src/domain/workflows/run-view-model.test.ts
@@ -12,6 +12,7 @@ import {
   workflowNodeIsSideNode,
   workflowNodeStatusTone,
   workflowRunIsActive,
+  workflowRunStatusTone,
   workflowRunTakesSideNode,
 } from "./run-view-model";
 
@@ -129,6 +130,12 @@ describe("workflowRunIsActive", () => {
     expect(
       RUN_STATUSES.map((status) => workflowRunIsActive(run({ status }))),
     ).toEqual([true, true, true, false, false, false]);
+  });
+});
+
+describe("workflowRunStatusTone", () => {
+  it("presents a cancelled run with the neutral tone", () => {
+    expect(workflowRunStatusTone("cancelled")).toBe("neutral");
   });
 });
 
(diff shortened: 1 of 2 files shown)


pablonyx:
Tier-2 `Workflow definition lifecycle` failure on this PR is **not** the tone typing — it is a click interception:

- The spec times out clicking `#workflow-builder-input-1-required`. Playwright's log shows a toast (`section[aria-label="Notifications alt+T"]` subtree) intercepting pointer events for the full 120s, across 197 retries.
- The toast is the desktop-worker enrollment failure: headline **"Integrations unavailable"**, cause **`Cannot read properties of undefined (reading 'invoke')`** — `ensureDesktopWorker` running in the intent-test environment where the tauri `invoke` side is absent, raising the sticky error toast from `use-desktop-worker-enrollment.ts` (`showErrorToast`, no auto-dismiss).
- This toast almost certainly fired before #1972 too, but the old stacked builder page had no click target under the toast region. The new three-pane inspector places the declared-input `required` checkbox bottom-right, directly under it — so #1972's layout made a latent env problem test-fatal. Main was green one commit before #1972 (5d9613666) and red from #1972 on.

Fix options (either unblocks tier-2):
1. Spec-side: dismiss notifications before inspector clicks (the toast has a "Close" button), or `page.keyboard.press("Alt+T")`-style handling in a helper.
2. Env-side (better long-term): the intent stack's host should not expose a desktop bridge without a working tauri transport, so enrollment never runs there.

Evidence: run 31982657482 artifact `workflow-definition-lifecycle-traces`, `error-context.md` (page snapshot shows the toast listitem with that exact cause text). `intent-tests (provisional)` fails the same way.

PR #2005, opened 2026-08-17
fix(product-client): scope transcript submit-stamp to session identity (PRO-175, r2 reopened)

pablonyx:
Reopens rung 2 of the Chat Scroll ladder (PRO-187). Prior PR #1932 was auto-closed by GitHub when its base branch (rung 1) was deleted on #1928's squash-merge, and such PRs cannot be reopened. Same branch, rebased onto main past the r1 squash; content identical to #1932's reviewed state (review round CLEAN and Manual verification recorded on #1932). Head 653f73b8d. Part of Pablo's granted bottom-up merge of the ladder.

### DIFF
--- a/apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts
+++ b/apps/packages/product-client/qualification/scroll-physics/specs/scroll-physics.spec.ts
@@ -323,11 +323,25 @@ test.describe("transcript scroll physics", () => {
     await drive(page, "finalizeStreamingTurn");
   });
 
-  // EXPECTED TO FAIL today (PRO-175). resetForSession always re-pins and snaps
-  // to bottom on every session switch, then runs a glue loop across the newly
-  // mounted rows, a revisit to a finalized session produces scrollTop motion
-  // frames instead of restoring its prior position with zero visible motion.
-  // Rung 2 introduces restore-finalized placement; unfixme there.
+  // EXPECTED TO FAIL today (PRO-175), for a DIFFERENT reason than the stamp
+  // leak this rung fixes. `resetForSession` still unconditionally
+  // pins/snaps/glues on every session switch by design (rung 2 leaves this
+  // alone), and the virtualizer's own remeasurement of freshly-mounted rows
+  // still produces a handful of settle frames after the snap — plus, in this
+  // fixture, a session switch resets `VirtualTranscriptRowList`'s
+  // `fallbackReason` to null, which can flip the tree from
+  // FullTranscriptRowList back to VirtualizedTranscriptRowList (a real
+  // component remount) if a blank-viewport fallback had fired for the
+  // intervening session. Either path produces multiple distinct scrollTop
+  // frames regardless of the stamp fix below. Verified: with
+  // `setPendingPromptStamp` reproducing the stamp-leak precondition
+  // (session-primary carries a stamp across an intervening session with
+  // none), the pre-fix code and the sessionKey-scoped fix in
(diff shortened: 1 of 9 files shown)


pablonyx:
Stack refold receipt (overnight staging): rebased onto main @ d5d38a804 via rebase --onto (excluded merged r1 head 0afad56ee). Diff vs main verified minimal (9 files, PRO-175 scope only). Union typecheck at stack top (r12 846bf9696 + pending cancelled-tone line) PASSES locally. Review carried from pre-rebase head; content unchanged.

pablonyx:
Delta review at 653f73b8d (reopened r2).

**Scope check:** `origin/main...653f73b8d` is exactly the expected 9-file rung-2 diff (scroll-physics spec comment, FullTranscriptRowList, VirtualizedTranscriptRowList, session-stamp test, use-transcript-stick-to-bottom(+test), use-transcript-submit-stamp-repin (new), use-transcript-virtualizer-blank-fallback(+test)). Merge-base with main is `d5d38a80` (r1). No drift, no scope creep beyond the rung-2 spec.

**Stamp scoping (PRO-175) — correct.** Traced `useTranscriptSubmitStampRepin` (`use-transcript-submit-stamp-repin.ts:440-458`): the session-boundary branch re-baselines `lastPromptSubmittedAtRef` to the incoming session's current stamp and skips the compare only when `previousSessionKey !== undefined && previousSessionKey !== sessionKey`, so first mount (where both refs start equal) and callers that never pass `sessionKey` (existing `use-transcript-stick-to-bottom.test.tsx` call site) both fall through to the original compare — no behavior change for non-scoped callers. Verified by hand against all three new session-stamp tests (A→A stamp bump, A→B no-misfire, A→B→C no-misfire-through-null), and the math holds. `sessionKey` composition `${selectedWorkspaceId ?? ""}:${activeSessionId}` in both row lists matches the deps array of the adjacent `resetForSession` effect (`[activeSessionId, selectedWorkspaceId]`) — no risk of the two keyings drifting apart. Good test quality, real negative controls, not just happy path.

**Blank-fallback grace window — one finding, non-blocking.**
`VirtualizedTranscriptRowList.tsx:224-231`, the session-switch effect clears `lastBlankReportSignatureRef`, `pendingPrependAnchorRef`, `lastOlderHistoryCursorRequestRef`, `lastPrefetchDecisionLogRef` and calls `resetForSession()`, but does **not** reset `prependSettleUntilRef.current` (set at line 253). Since the row list never remounts across a session switch (that's the premise of this whole rung), if a user switches sessions within the 3s `PREPEND_BLANK_FALLBACK_GRACE_MS` window opened by a prepend in the outgoing session, the stale deadline carries into the new session and suppresses genuine blank-viewport detection there for up to ~3s. Bounded and narrow (blank-viewport is already a rare safety-net path), but it's an easy one-line fix (`prependSettleUntilRef.current = 0;` alongside the other ref resets in that effect) and it's inconsistent with how the other prepend-scoped refs in the same effect are handled. Worth a follow-up fix, doesn't need to block this rung.

Everything else: negative-control test present for the grace window itself (lapsed-deadline still falls back — `use-transcript-virtualizer-blank-fallback.test.tsx`), no leftover debug code, file-length gate respected (`use-transcript-stick-to-bottom.ts` at 399 lines, `VirtualizedTranscriptRowList.tsx` at 400), spec-file comment update is accurate and cites a verified claim (identical scrollTop traces pre/post fix) rather than an assumption, and correctly attributes zero-motion placement to rung 6 per the frozen ladder.

**Verdict:** one minor non-blocking finding (stale `prependSettleUntilRef` across session switch); everything else is clean. No blockers to merging this rung.

PR #2009, opened 2026-08-17
feat(product): read-only background terminal detail view (bgwork R3)

pablonyx:
## What

New `BackgroundTerminalView` (`apps/packages/product-client/src/components/workspace/activity/background-pane/BackgroundTerminalView.tsx`): a read-only detail view for one background process. Header on `TerminalTopBar`'s bar grammar (back `PaneIconButton` + `Terminal` glyph + the command as title); body on `TerminalPanel`'s viewport shell, rendering the shell prompt line, the command exactly as run, then the feed's bytes via the existing `useFeedStream` hook (lazy, plain text, capped buffer); footer copy exactly as specified. Explicitly not: status strip, badge, `TerminalCommandFloatingAction`, xterm, any input affordance.

Wired into `BackgroundWorkPane`'s terminal seam: a new `selectedProcessId` state (mirroring the existing `selectedSubagentId` seam) swaps the roster body for `BackgroundTerminalView` when a Terminals roster row is opened, from either the Running or Closed scope. `LiveTerminalsRosterPanel` drops its own inline expand-in-place state and `useFeedStream` wiring (that responsibility moves to the detail view) and becomes a plain `onOpen`-routing roster, the same shape as `AgentsRosterPanel`.

## Why

Rung R3 of the Background Work delivery spec. Source: Design Handoff — `HANDOFF-background-work.md` (NEW `BackgroundTerminalView`; MODIFIED `BashCommandCall`); Delivery Spec — Background Work Slice 1, rung R3.

## Review round 1 (MINOR, fixed)

**Finding:** `BackgroundTerminalView` gated its feed only on `feed !== null`. The right panel collapses via CSS (`opacity-0`/`inert` on `WorkspaceShellRightRail`) rather than unmounting, so opening a terminal detail then collapsing the drawer left the socket streaming invisibly until `onBack` or a session change.

**Fix:** checked `useFeedStream`'s behavior when `enabled` flips false — it resets to `IDLE` (empty content) immediately rather than retaining partial content, so there's nothing to preserve across a collapse/reopen cycle. Gated at the `BackgroundWorkPane` seam instead of adding a prop: `feed={isOpen ? selectedProcess.feed : null}`. `BackgroundTerminalView`'s frozen `(process, feed, onBack)` prop contract is untouched — this is shape (b) from the two options considered, chosen because content does not survive a disable/re-enable cycle, and it matches the acceptance line "feed re-tails on detail open" (reopening hands back the real `FeedRef` and the view re-tails from scratch). Also corrected a docstring sentence that had gone stale on the (false, once CSS-only collapse is accounted for) premise that the view is only ever mounted while open.

Added a test asserting the feed-enabled signal reaching `BackgroundTerminalView` flips off when `isOpen` goes false with a detail still selected (detail seam stays rendered, not unmounted) and back on on reopen.

## Gap report — `BashCommandCall` MODIFIED not implemented this rung (independently verified correct)

The handoff and delivery spec also ask: for a tool call the activity domain knows as a background process, its transcript row click-opens the pane's terminal view instead of the inline output disclosure (ordinary commands keep the inline panel unchanged).

I traced the correlation this needs end to end and found **none exists in this repo**:

- `anyharness-contract`'s `ActivityProcess` (`anyharness/crates/anyharness-contract/src/v1/activity.rs`) and the wire payload `ActivityProcessWire` (`anyharness-lib`'s `domains/activity/wire.rs`) carry only `id, command, cwd, status, pid, startedAt/startedAtMs, endedAt/endedAtMs, feed` — `id` is the harness's own process/task id, with no field referencing the originating tool call.
- The SDK's `ToolCallItem` (`anyharness/sdk/src/types/reducer.ts`) carries `toolCallId`/`itemId`, with nothing pointing the other way either.
- The one existing tool-call ↔ background-work correlation in this repo, `reducer/background-work.ts`'s `ToolBackgroundWorkMetadata` (`trackerKind: "claude_async_agent"`, `agentId`), is scoped to **async subagent launches** (consumed by `domain/chats/subagents/subagent-launch.ts` for `SubagentLaunchLedger`), not bash processes — there is no equivalent tracker kind for background bash.
- RF's own acceptance debt item ("live probe that `task_started.task_id === BashOutput.backgroundTaskId`") lives entirely in the forked `claude-agent-acp` repo and is explicitly unverified even there (per the Delivery Spec's RF entry).

Per the assignment's explicit instruction to stop and report rather than invent a heuristic here, `BashCommandCall.tsx` is untouched this rung: background-originating and ordinary bash commands render identically today (current behavior preserved). This is the concrete contradiction to resolve — either by exposing a real correlation on the wire (an RF/runtime change) or by a founder ruling to defer this half of R3 — before the click-through can be built without guessing.

## Tests

- `BackgroundTerminalView.test.tsx`: renders the shell prompt line + command + streamed feed content, exact footer copy, asserts no `input`/`textarea`/`select` anywhere, `onBack` fires, `useFeedStream` is called with `enabled: true` only when a feed ref is present (`enabled: false` otherwise).
- `BackgroundWorkPane.test.tsx`: opening a terminal row (Running and Closed scope) swaps in the real `BackgroundTerminalView` with the right process id; back returns to the roster; the seam closes on session-id change; and (review-round addition) the feed-enabled signal flips off/on as `isOpen` toggles false/true while the detail seam stays mounted.

## Gates (current head)

- Affected vitest (`BackgroundTerminalView.test.tsx` + `BackgroundWorkPane.test.tsx`): **19/19 passed**
- Full `product-client` vitest: **1003 files / 7226 tests passed**
- `tsc -p tsconfig.domain.json --noEmit`: 1 error — `TS2741` in `src/domain/workflows/run-view-model.ts:75` (missing `cancelled` key in `RUN_STATUS_TONE`). Confirmed pre-existing on the R2 base (`origin/bgwork/r2-background-work-pane`), unrelated to this diff — attributed, not chased.
- `tsc -p tsconfig.json --noEmit`: same single exogenous error, no new errors.
- `python3 scripts/report_frontend_structure.py --strict`: **0 violations** across all `FE-STRUCT-*`/`FE-SIZE-1` categories.
- `python3 scripts/check_appearance_scaling.py`: **passed**.

Draft PR, stacked on `bgwork/r2-background-work-pane`. Not merging — awaiting review.

### DIFF
--- a/apps/packages/product-client/src/components/workspace/activity/LiveTerminalsRosterPanel.tsx
+++ b/apps/packages/product-client/src/components/workspace/activity/LiveTerminalsRosterPanel.tsx
@@ -1,86 +1,35 @@
-import { useCallback, useState } from "react";
 import {
   sortProcessesForDisplay,
   type ActivityProcessWire,
 } from "#product/domain/activity/process";
 import { TerminalRosterRow } from "#product/components/workspace/activity/TerminalRosterRow";
 import { RosterPanel } from "#product/primitives/patterns/RosterPanel";
-import { useActiveSessionId } from "#product/hooks/chat/derived/use-active-session-identity";
-import { useSessionDirectoryStore } from "#product/stores/sessions/session-directory-store";
-import { useFeedStream } from "#product/hooks/activity/derived/use-feed-stream";
 
 export interface LiveTerminalsRosterPanelProps {
   processes: ActivityProcessWire[];
   nowMs: number;
+  /** Opens the pane's `BackgroundTerminalView` detail seam for this row (rung R3). */
+  onOpen?: (processId: string) => void;
 }
 
 /**
- * The ▸ chip's click-in panel, live-wired: read-only agent-spawned background
- * processes with a click-to-expand live tail sourced from each row's opaque
- * `FeedRef` over `WS /v1/feeds/{feedId}`. Bytes flow only while a row is
- * expanded (lazy FeedService semantics). Presentation of the row header stays
- * in the shared `TerminalRosterRow`; this desktop wrapper owns the SDK feed
- * wiring and expansion state.
+ * The Background work pane's Terminals roster group: a read-only summary of
+ * agent-spawned background processes. Row selection routes to
+ * `BackgroundWorkPane`'s `BackgroundTerminalView` detail seam (Design
+ * Handoff — MODIFIED `TerminalRosterRow`; Delivery Spec — Background Work
+ * Slice 1, rung R3) rather than expanding a live tail inline — the detail
+ * view owns the feed wiring now, so this panel stays a plain, stateless
+ * roster like `AgentsRosterPanel`.
  */
-export function LiveTerminalsRosterPanel({ processes, nowMs }: LiveTerminalsRosterPanelProps) {
-  const activeSessionId = useActiveSessionId();
-  const workspaceId = useSessionDirectoryStore((state) =>
-    activeSessionId ? state.entriesById[activeSessionId]?.workspaceId ?? null : null,
-  );
-  const [expandedId, setExpandedId] = useState<string | null>(null);
-  const toggle = useCallback((processId: string) => {
-    setExpandedId((current) => (current === processId ? null : processId));
-  }, []);
-
+export function LiveTerminalsRosterPanel({ processes, nowMs, onOpen }: LiveTerminalsRosterPanelProps) {
   const sorted = sortProcessesForDisplay(processes);
   return (
     <RosterPanel title="Terminals" empty="No background terminals." data-terminals-roster-panel>
       {sorted.map((process) => (
         <li key={process.id}>
-          <LiveTerminalRow
-            process={process}
-            nowMs={nowMs}
-            workspaceId={workspaceId}
-            expanded={expandedId === process.id}
-            onToggle={toggle}
-          />
+          <TerminalRosterRow process={process} nowMs={nowMs} onOpen={onOpen} />
         </li>
       ))}
     </RosterPanel>
-  );
-}
-
-interface LiveTerminalRowProps {
-  process: ActivityProcessWire;
-  nowMs: number;
-  workspaceId: string | null;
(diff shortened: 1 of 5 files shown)


pablonyx:
**Refold receipt (merge-train, on Pablo's word):** after R2 landed via #2029's squash, this stacked rung went CONFLICTING (carried R2's original commits). Refolded per the established sequential-refold protocol: `git rebase --onto origin/main b04896c02 9698c8e86` — the three keep-up merge commits drop (their content is main's #2029 squash), the two content commits replay.

**Range-diff verdict: all-`=` (per-commit byte-identical) — review carries.**
```
1:  f051fef1f = 1:  8075a5fd6 feat(product): read-only background terminal detail view (bgwork R3)
2:  5dc657379 = 2:  6af22bfbb fix(product): stop the terminal feed while the right panel is collapsed (bgwork R3 review, MINOR)
```
Old head 9698c8e86 → new head 6af22bfbb. Fresh full CI on this head is the merge gate; literal tally will be quoted before the click.

pablonyx:
**Flake disclosure (merge-train):** fresh CI on refolded head 6af22bfbb settled 24 pass / 1 fail — the sole red is `process_kill::tests::census_counts_git_executables_distinctly_from_the_total` (Cargo job 95544882141, panic at process_kill_tests.rs:302), the KNOWN census flake that also went red-then-rerun-green on #1988 earlier today (job 95507962397 → 10m57s rerun pass). This PR's diff contains zero Rust. Same-ref rerun of the failed job dispatched; merge proceeds only on a green rerun with the literal tally quoted here.

pablonyx:
**Settled tally on head 6af22bfbb (merge gate):** ALL PASS — 25 checks, 0 fail, 0 pending. Census rerun GREEN: `Cargo check & test pass 8m51s` (job 95547287347) — flake attribution confirmed on the same ref. Full literal board in the train ledger. Merging on Pablo's word ("bg work ... shoud mrge") per the recorded ladder.

PR #2010, opened 2026-08-17
feat(product): read-only background subagent detail view (bgwork R4)

pablonyx:
## What / why

Rung R4 of the Background Work slice: a read-only detail view for one harness-native subagent, plus the transcript/identity cleanup the handoff pairs with it. Source of truth: `HANDOFF-background-work.md` (design handoff — wins on UI detail) and `Delivery Spec - Background Work Slice 1.md` rung R4.

## Per-piece notes

**`BackgroundSubagentView`** (new) — `.../background-pane/BackgroundSubagentView.tsx`. Header on `AgentsPaneDetail`'s grammar: back button, `AgentIdentityGlyph` (dimension 20), title, `Status · provider` secondary line, and a "read only" `Badge`. Body: the correlated launch tool call's initial prompt in a `ToolActionDetailsPanel` (`StickyNote` + "Initial prompt"), then the subagent's feed. Footer replaces the composer with the exact handoff copy: "Read-only. Transcript mirrored from the agent; no composer." Wired into `BackgroundWorkPane`'s R2 seam, gated by the same `isOpen` feed pattern `BackgroundTerminalView` (R3) established (the pane stays mounted-but-hidden via CSS, so streaming must stop at that boundary).

**BLOCKER — feed fidelity, reported not worked around.** The handoff's "Feed fidelity" note calls `MessageList`-based transcript-shaped rendering "honest for Claude today" because "the child transcript arrives as a JSONL tail over `tail_file`." Traced end to end:
- `open_tail_file` (`anyharness-lib/src/domains/activity/feeds.rs`) always emits raw `FeedFrame::Bytes` regardless of `FeedKind` — there is no per-kind branch.
- The WS handler (`api/ws/feeds.rs`) does no JSONL→ACP translation.
- A Claude subagent's `tail_file` feed therefore carries Claude's own native CLI session-log JSONL (`parentUuid`/`isSidechain`/`uuid` fields) — not the `SessionEventEnvelope` shape `@anyharness/sdk`'s reducer (and therefore `MessageList`, which requires a `TranscriptState`) needs.
- No translator for this exists anywhere in the repo (exhaustively grepped for `isSidechain`/`parentUuid`).

`BackgroundSubagentView` therefore renders the feed as plain text for **every** harness — the same raw-tail treatment the handoff already rules for Codex pending the `acp_child_demux` binding — rather than inventing a client-side JSONL-to-ACP translator, which is out of this slice's scope. Same class of stop as R3's `BashCommandCall` correlation gap. **This stays escalated to the founder track — confirmed true end to end on review; no translator was built.**

**`SubagentCreationGroupBlock` / `SpawnIdentityReceipt` — disclosed spec correction, not a closed non-issue.** Traced the full path before touching anything: `transcript-presentation.ts`'s `subagent_creations` grouping and `agent-operations-tool-presentation.ts` confirm these components are scoped exclusively to the AgentOperations `create_agent` MCP tool (delegated work, a full separate durable session). A harness-native Task/Agent tool call never reaches `SubagentCreationGroupBlock` — it routes through `TranscriptAgentGroupBlock` instead, so the spec's premise (native subagents flow through `SpawnIdentityReceipt`/`AgentIdentityChip`) is false — there is no "native vs delegated" branch point inside `SpawnIdentityReceipt` to extend; the two mechanisms are architecturally disjoint today. The spec's *intent*, though, is unambiguous and buildable at the right component: a native subagent's transcript block must click-open `BackgroundWorkPane`'s subagent detail. Delivered here at `TranscriptAgentGroupBlock` instead:
  - Resolves the correlated subagent id via the existing `ToolBackgroundWorkMetadata.agentId` correlation (new `resolveSubagentIdForItem` in `subagent-launch.ts`, the forward direction of `findSubagentLaunchItem`'s reverse lookup).
  - The handoff names no distinct identity-chip affordance for this block (no `AgentIdentityGlyph`/chip in its anatomy, unlike the unrelated `SpawnIdentityReceipt` row) — so per the reviewer's fallback, the header's own click now opens the pane when eligible, and expand/collapse moves to a new trailing chevron (`Button variant="unstyled"` + `ChevronRight`, same rotate-on-expand treatment as `TurnSeparator`). A block with no background-work correlation (or no `onOpenSubagent` wired at the call site) keeps today's byte-identical expand-on-click header — no chevron, no behavior change.
  - Threaded `onOpenSubagent?: (subagentId: string) => void` down the existing `onOpenArtifact`/`workspaceId` prop chain (`MessageList` → `TranscriptTurnRow` → `TurnItemSequence` → `TranscriptTreeNode` → `TranscriptToolCallGroupBlock` → `TranscriptAgentGroupBlock`), constructed once in `MessageList` from `useOpenBackgroundWorkPane()`.
  - `useOpenBackgroundWorkPane`'s return is extended from `() => void` to `(subagentId?: string) => void` — an optional deep-open target, not a new hook, not a new global. When given, it writes a one-shot `pendingBackgroundSubagentSelectionByWorkspace[workspaceId]` entry into the same `workspace-ui-store` right-panel model this hook already owns, alongside the existing materialize+open calls. `BackgroundWorkPane` consumes and clears it in an effect (single writer, single reader — no epoch/nonce machinery needed, unlike `pendingChatActivationByWorkspace`). Existing zero-arg callers (`BackgroundWorkTranscriptRow`'s `onOpen`) are unaffected.
  - `SpawnIdentityReceipt.tsx`/`SubagentCreationGroupBlock.tsx` themselves are byte-untouched; delegated-work routing is unchanged.

**Identity nit (flagged, not fixed — out of this rung's scope).** `buildDelegatedAgentIdentity` uses the wire id (Claude `task_id` / Codex `threadId`) as `sessionId`, which only loosely satisfies "durable session id": `identity.openTarget.sessionId` ends up carrying a non-session-backed id. Harmless today — nothing in this PR's scope consumes `openTarget` — but flagged here for future consumers that might expect it to resolve to a real session.

**`SubagentLaunchLedger`** — drops its `prompt` prop and the whole prompt-disclosure affordance (moved to `BackgroundSubagentView`'s "Initial prompt" panel). Added `isSubagentLaunchStatusVisibleInTranscript` (`subagent-launch.ts`) suppressing exactly the three states the handoff/spec name — `running` / `background` / `completed_background` — so `Creating` / `Running in background` / `Completed in background` no longer render in the transcript (acceptance line 52). `formatLaunchStatus` itself is unchanged; only its transcript call sites go. `TranscriptAgentGroupBlock`'s `hasLaunchLedger` now derives from the same predicate instead of prompt-presence.

**Native identity** — added `findSubagentLaunchItem` (`subagent-launch.ts`) reusing the one existing tool-call correlation in the repo, `ToolBackgroundWorkMetadata.agentId` (`parseToolBackgroundWork`, Claude-only `trackerKind: "claude_async_agent"`), to locate a subagent's initial prompt from its roster id. `SubagentRosterRow`/`AgentsRosterPanel` now thread `workspaceId` and call `buildDelegatedAgentIdentity({ id, title, workspaceId, sessionId: subagent.id })`, rendering `AgentIdentityGlyph` instead of the plain `Fork` glyph — the roster row, detail header, and (pre-existing) transcript rendering now agree on one generated name/color/glyph per the handoff's native-identity note.

**Assumption disclosed:** the delivery spec's Acceptance section says "the two muted transcript status lines" while its own MODIFIED section and my task brief both name three (`Creating` / `Running in background` / `Completed in background`). Followed the more specific three-state instruction.

## Delta-review round 2: three MINORs closed

1. **Direct unit tests for `resolveSubagentIdForItem`** (`subagent-launch.test.ts`) — happy path plus malformed `rawOutput` shapes (string, array, `_anyharness` present with an unrecognized `trackerKind`, `_anyharness.backgroundWork` absent), all returning `null`. Previously only exercised at the `TranscriptAgentGroupBlock` integration level.

2. **Session-scoped the pending subagent selection.** The reviewer offered two options — key `pendingBackgroundSubagentSelectionByWorkspace` by `${workspaceId}:${sessionId}`, or add a `sessionId` field checked on consume — and asked for "the smaller diff." I started with an even smaller third option (clear the pending entry inside the pre-existing session-reset effect) but traced through React's effect-execution model and rejected it: on the realistic path — the same transcript click that writes the pending selection also flips the right panel to `"background"`, which is what *mounts* `BackgroundWorkPane` for the first time — mount runs every effect unconditionally, so the reset effect's `setSelectedSubagentId(null)` and the consumption effect's `setSelectedSubagentId(pending.subagentId)` both fire in the same commit, and the consumption effect (declared after reset) wins regardless of the reset effect's store write. That doesn't reliably close the race. Implemented the `sessionId`-field option instead: `PendingBackgroundSubagentSelection = { subagentId, sessionId }`, captured at write time in `useOpenBackgroundWorkPane` (now `(subagentId?, sessionId?) => void`, both threaded from `MessageList`'s `activeSessionId`), checked against the consuming pane's own `sessionId` prop on read — a mismatch is discarded (but still cleared) inside the same effect that reads the value, so it's correct regardless of mount/remount/same-commit timing. New test: a pending selection written under a different session in the same workspace, with a subagent id that *does* exist in the current session's roster (to prove the guard — not roster-membership — is what keeps it off), is discarded rather than applied.

3. **Keyboard access on the header.** `TranscriptAgentGroupBlock`'s clickable header (`div`, pre-existing) now navigates rather than only toggling, so it gets `role="button"`, `tabIndex={0}`, and an `onKeyDown` handling Enter/Space for both branches (open-subagent, expand-fallback), plus an `aria-label` naming the action. Verified `role="button"` on a `div` doesn't trip `FE-STRUCT-1` (it only regexes literal `<button|input|label|select|textarea>` tags), so no `Button`-wrapper swap was needed. Renamed the expand-fallback header's label to "Show/Hide subagent details" (distinct from the chevron's existing "Expand/Collapse subagent details") — same wording would've given two `role="button"` elements outside the same click target an identical accessible name, which is a real a11y smell even though the two never coexist in the same render (the chevron only renders when the header is also `canOpenSubagent`). The header's `onKeyDown` ignores keydown events whose `target` isn't the header itself (`event.target !== event.currentTarget`), so pressing Enter/Space on the focused chevron button — a real nested `<button>` — can't also re-fire the header's own action via bubbling. Chevron itself is unchanged.

## Gates

- Full `product-client` vitest suite: **1007/1007 test files passed, 7275/7275 tests passed.**
- `tsc -p tsconfig.domain.json` and `tsc -p tsconfig.json --noEmit`: both show only the pre-existing exogenous `TS2741` on `run-view-model.ts(75,7)` — present on base, unrelated to this diff.
- `scripts/report_frontend_structure.py --strict`: **0** violations across all FE-STRUCT/FE-SIZE checks (confirmed `role="button"` divs are not flagged).
- `scripts/check_appearance_scaling.py`: **passed.**
- `scripts/check_design_attribution.py`: **passed** (fixed post-CI — see below).
- No raw `button`/`input`/`label`/`select`/`textarea` introduced in `components/`; primitives only (the chevron uses `Button variant="unstyled"`; the header is a `div` with `role="button"`).

**CI fix-forward — design-attribution gate (PROD-ATTR), mechanical rename, no behavior change:**
- `BackgroundSubagentView.test.tsx:141` — test title credited the raw-tail fallback rendering to "Codex-style" (PROD-ATTR-1: crediting another product for our own styling). Renamed to describe the treatment itself: "raw-tail fallback used for every harness."
- `subagent-launch.test.ts:332` — fixture id `codex-thread-1` joined a product name to a design noun (PROD-ATTR-5: `<product>-thread`). Renamed to `child-thread-1`. Bare `codex` as a harness-kind value elsewhere in the diff is unaffected — the gate's own output confirms that's legitimate product vocabulary; only the product-name+design-noun join was banned.
- Swept the full R4 diff for any other `<product>-style/-like/-esque` prose or `<product>-<design-noun>` identifier joins (case-insensitive, all of `claude`/`codex`/`anthropic`/`openai`/`gpt`) — these were the only two.
- Local verification: `scripts/check_design_attribution.py` passes clean; both affected test files (36 tests) pass; `scripts/report_frontend_structure.py --strict` stays at 0.

## Test plan

- [x] `BackgroundSubagentView` renders header identity/status/provider, the correlated initial prompt (and omits the panel when none correlates), the feed's raw tail, the exact footer copy, no input/textarea/select anywhere, `onBack` fires, and the feed enable/disable gating on `isOpen`.
- [x] `SubagentLaunchLedger` renders nothing for the three retired states, still renders `Launch failed`/`Stopped updating`/`Started`, and never renders a prompt-disclosure affordance.
- [x] `TranscriptAgentGroupBlock` expanded view no longer shows the three muted lines but still shows non-retired ones (e.g. `Launch failed`).
- [x] `TranscriptAgentGroupBlock` native-routing fix-forward: clicking the header calls `onOpenSubagent` with the *correct* correlated subagent id (asserted, not just "opened"); the chevron still expands/collapses when `onOpenSubagent` is wired; a block whose `rawOutput` carries no background-work metadata gets no affordance and keeps the byte-identical expand-on-click header.
- [x] `useOpenBackgroundWorkPane`: the optional subagent id writes the pending-selection entry for the active workspace and still opens the pane exactly as the zero-arg call does; the zero-arg call leaves no pending selection.
- [x] `BackgroundWorkPane`: consumes and clears a pending subagent selection scoped to its own `workspaceId` (ignores one written for a different workspace); opening a subagent renders the real `BackgroundSubagentView`, back returns to the roster, the view bounces back to the roster when the subagent leaves it (finishes), feed gates on `isOpen`, session-id change closes the seam.
- [x] `subagent-launch.ts`: unit tests for `isSubagentLaunchStatusVisibleInTranscript` and `findSubagentLaunchItem` (including the "no correlation for other harnesses" case).
- [x] `SubagentRosterRow`/`AgentsRosterPanel`: generated identity glyph consistent with `buildDelegatedAgentIdentity` for the same id, `onOpen` firing, running-only filtering.
- [x] `resolveSubagentIdForItem` (pure-function unit tests): happy path, and `rawOutput` as a string / an array / with an unrecognized `trackerKind` / missing `_anyharness.backgroundWork` all return `null`.
- [x] `useOpenBackgroundWorkPane`: writes a session-scoped `{ subagentId, sessionId }` pending selection only when *both* arguments are given; the subagent-id-without-sessionId case writes nothing.
- [x] `BackgroundWorkPane`: discards (but still clears) a pending selection written for a different session in the same workspace, even when that selection's subagent id happens to exist in the current session's roster.
- [x] `TranscriptAgentGroupBlock` keyboard access: Enter and Space on the header call `onOpenSubagent` with the correct id when eligible; Enter on the header toggles expand/collapse in the fallback case; the header exposes `role="button"`/`tabIndex={0}`/a distinct `aria-label`; Enter on the (separately focusable) chevron does not also re-fire the header's action.

### DIFF
--- a/apps/packages/product-client/src/components/workspace/activity/AgentsRosterPanel.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/activity/AgentsRosterPanel.test.tsx
@@ -1,0 +1,57 @@
+// @vitest-environment jsdom
+
+import { cleanup, render } from "@testing-library/react";
+import { afterEach, describe, expect, it } from "vitest";
+import { AgentsRosterPanel } from "#product/components/workspace/activity/AgentsRosterPanel";
+import type { ActivitySubagentWire } from "#product/domain/activity/subagent";
+
+afterEach(cleanup);
+
+function subagent(overrides: Partial<ActivitySubagentWire>): ActivitySubagentWire {
+  return {
+    id: "agent-1",
+    agentType: "general-purpose",
+    description: "Inspect the transcript pipeline",
+    model: null,
+    background: false,
+    status: { status: "running" },
+    usage: null,
+    feed: null,
+    ...overrides,
+  };
+}
+
+describe("AgentsRosterPanel", () => {
+  it("shows only running agents and threads workspaceId to each row", () => {
+    const { container, getByText, queryByText } = render(
+      <AgentsRosterPanel
+        workspaceId="workspace-1"
+        nowMs={0}
+        agents={[
+          subagent({ id: "agent-running", status: { status: "running" } }),
+          subagent({
+            id: "agent-done",
+            description: "Finished task",
+            status: { status: "completed", summary: null },
+          }),
+        ]}
+      />,
+    );
+
+    expect(getByText("Inspect the transcript pipeline")).not.toBeNull();
+    expect(queryByText("Finished task")).toBeNull();
+    expect(container.querySelectorAll("[data-subagent-roster-row]")).toHaveLength(1);
+    // Every glyph rendered has a real generated color, proving the row
+    // received a non-empty workspaceId/sessionId all the way through.
+    const svg = container.querySelector("[data-subagent-roster-row] svg");
+    expect(svg?.getAttribute("style")).toMatch(/color:/);
+  });
+
+  it("shows the empty copy when no native subagent is running", () => {
+    const { getByText } = render(
+      <AgentsRosterPanel workspaceId="workspace-1" nowMs={0} agents={[]} />,
+    );
+
+    expect(getByText("No active native subagents.")).not.toBeNull();
+  });
+});
--- a/apps/packages/product-client/src/components/workspace/activity/AgentsRosterPanel.tsx
+++ b/apps/packages/product-client/src/components/workspace/activity/AgentsRosterPanel.tsx
@@ -5,6 +5,7 @@ import { SubagentRosterRow } from "./SubagentRosterRow";
 export interface AgentsRosterPanelProps {
   agents: ActivitySubagentWire[];
   nowMs: number;
+  workspaceId: string;
   onOpen?: (subagentId: string) => void;
 }
 
@@ -17,7 +18,7 @@ export interface AgentsRosterPanelProps {
  * (`features/delegated-work.md`) as a new `subagent` source, inheriting
  * generated identity/color there.
  */
-export function AgentsRosterPanel({ agents, nowMs, onOpen }: AgentsRosterPanelProps) {
+export function AgentsRosterPanel({ agents, nowMs, workspaceId, onOpen }: AgentsRosterPanelProps) {
   const runningAgents = agents.filter((agent) => agent.status.status === "running");
   const sorted = sortSubagentsForDisplay(runningAgents);
   return (
@@ -28,7 +29,7 @@ export function AgentsRosterPanel({ agents, nowMs, onOpen }: AgentsRosterPanelPr
     >
       {sorted.map((agent) => (
         <li key={agent.id}>
-          <SubagentRosterRow subagent={agent} nowMs={nowMs} onOpen={onOpen} />
+          <SubagentRosterRow subagent={agent} nowMs={nowMs} workspaceId={workspaceId} onOpen={onOpen} />
         </li>
       ))}
     </RosterPanel>
--- a/apps/packages/product-client/src/components/workspace/activity/SubagentRosterRow.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/activity/SubagentRosterRow.test.tsx
@@ -1,0 +1,57 @@
+// @vitest-environment jsdom
+
+import { cleanup, fireEvent, render } from "@testing-library/react";
+import { afterEach, describe, expect, it } from "vitest";
+import { SubagentRosterRow } from "#product/components/workspace/activity/SubagentRosterRow";
+import { buildDelegatedAgentIdentity } from "#product/lib/domain/delegated-work/identity";
+import type { ActivitySubagentWire } from "#product/domain/activity/subagent";
+
+afterEach(cleanup);
+
+const SUBAGENT: ActivitySubagentWire = {
+  id: "agent-1",
+  agentType: "general-purpose",
+  description: "Inspect the transcript pipeline",
+  model: "claude-sonnet",
+  background: false,
+  status: { status: "running" },
+  usage: null,
+  feed: null,
+};
+
+describe("SubagentRosterRow", () => {
+  it("renders a generated identity glyph consistent with buildDelegatedAgentIdentity for the same id", () => {
+    const { container } = render(
+      <SubagentRosterRow subagent={SUBAGENT} nowMs={0} workspaceId="workspace-1" />,
+    );
+
+    const expectedIdentity = buildDelegatedAgentIdentity({
+      id: SUBAGENT.id,
+      title: "Inspect the transcript pipeline",
+      workspaceId: "workspace-1",
+      sessionId: SUBAGENT.id,
+    });
+
+    const svg = container.querySelector("[data-subagent-roster-row] svg");
+    expect(svg).not.toBeNull();
+    expect(svg?.getAttribute("style")).toContain(expectedIdentity.colorVar);
+  });
+
+  it("fires onOpen with the subagent id on select", () => {
+    let openedId: string | null = null;
+    const { getByText } = render(
+      <SubagentRosterRow
+        subagent={SUBAGENT}
+        nowMs={0}
+        workspaceId="workspace-1"
+        onOpen={(id) => {
+          openedId = id;
+        }}
+      />,
+    );
+
+    fireEvent.click(getByText("Inspect the transcript pipeline"));
+
+    expect(openedId).toBe("agent-1");
+  });
+});
--- a/apps/packages/product-client/src/components/workspace/activity/SubagentRosterRow.tsx
+++ b/apps/packages/product-client/src/components/workspace/activity/SubagentRosterRow.tsx
@@ -1,4 +1,3 @@
-import { Fork } from "#product/primitives/icons/core";
 import {
   subagentDisplayTitle,
   subagentStatusLabel,
@@ -9,7 +8,8 @@ import {
 } from "#product/domain/activity/subagent";
 import { statusDotToneTextClass, type StatusDotTone } from "#product/primitives/StatusDot";
 import { RosterRow } from "#product/primitives/patterns/RosterRow";
-import { twMerge } from "#product/primitives/utils/tw-merge";
+import { AgentIdentityGlyph } from "#product/components/patterns/AgentIdentityGlyph";
+import { buildDelegatedAgentIdentity } from "#product/lib/domain/delegated-work/identity";
 
 const SUBAGENT_STATUS_DOT_TONE: Record<SubagentTone, StatusDotTone> = {
   default: "muted",
@@ -20,6 +20,7 @@ const SUBAGENT_STATUS_DOT_TONE: Record<SubagentTone, StatusDotTone> = {
 export interface SubagentRosterRowProps {
   subagent: ActivitySubagentWire;
   nowMs: number;
+  workspaceId: string;
   /** Optional per-row click-in, e.g. to open the existing delegated-work details surface. */
   onOpen?: (subagentId: string) => void;
 }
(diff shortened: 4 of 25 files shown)


pablonyx:
**Refold receipt (merge-train):** stacked rung refolded onto main after R3's squash (ee27e88e6). `git rebase --onto origin/main 9698c8e86 dfa96d21` — keep-up merges drop, four content commits replay.

**Range-diff verdict: all-`=` (per-commit byte-identical) — review carries.**
```
1:  78b81c52f = 1:  34d78c210 feat(product): read-only background subagent detail view (bgwork R4)
2:  91af843d1 = 2:  4755f98da fix(product): wire native subagent transcript click (review fix-forward)
3:  4430513d6 = 3:  87e022af3 fix(product): close R4 delta-review MINORs
4:  0d38ccad7 = 4:  7a361deea fix(product): design-attribution renames
```
Old head dfa96d21 → new head 7a361deea. Fresh CI is the merge gate; literal tally quoted before the click.

pablonyx:
Housekeeping receipt: the 00:02Z refold force-push (head 7a361deea81c) never received its `pull_request: synchronize` CI runs — only PR Metadata triggered, likely because the base retarget and the push landed near-simultaneously. Close/reopen at ~00:14Z re-fired the board content-free: CI, Intent Tests, CodeQL, and Self-Host Smoke are now in progress on the same head SHA. No commits changed; the all-`=` range-diff receipt above still describes this exact head.

pablonyx:
Fresh CI on refolded head 7a361deea81c settled **23 pass / 1 fail**. The sole red is Cargo check & test ([job 95551847970](https://github.com/proliferate-ai/proliferate/actions/runs/32083762326/job/95551847970)), and it is NOT this PR: the failing test is `diagnostics_collector::process::tests::rv_2_05_child_exit_during_launch_is_never_graceful_shutdown` (panic at `apps/desktop/src-tauri/src/diagnostics_collector/process_tests.rs:149`, `left: EndpointUnavailable / right: ChildExited`) — a timing-sensitive process-lifecycle test (fixed 50ms sleep before expecting the child's exit to be observed) in a file this PR never touches. This PR's diff is 25 files, all `.ts`/`.tsx`, zero Rust; the same Cargo job passed at 22:50Z on the pre-refold head `dfa96d21` whose content commits are byte-identical to this head (all-`=` range-diff above). Note this is a DIFFERENT flake than the known census one — both census tests passed this round. Treating as exogenous flake: same-ref rerun of the failed job now; merge only on a green rerun, verdict quoted here either way.

pablonyx:
Settled verdict on head 7a361deea81c after the disclosed same-ref rerun: **24 checks, 24 pass / 0 fail / 0 pending** — Cargo check & test now **pass** (14m56s, [rerun job 95554316922](https://github.com/proliferate-ai/proliferate/actions/runs/32083762326/job/95554316922)), confirming the rv_2_05 red as exogenous timing flake exactly as disclosed above. Full green board: Analyze ×3, Build SDK, CI/CD config, Candidate build handoff, Cargo, CodeQL, Desktop frontend build, Detect smoke-relevant changes, Login first-load budget, Mobile typecheck, Production compose smoke, Repo shape, Shared frontend packages, Terraform validate, Transcript scroll physics, Validate PR title and labels, Vercel ×2, Web frontend build, Workflow definition lifecycle, intent-billing, intent-tests. Merging on Pablo's standing bg-work merge word.

PR #2011, opened 2026-08-17
fix(desktop): guard desktop-worker enrollment when no Tauri transport present

pablonyx:
## Diagnosis (verified against code, credit PR #2002 comment by pablonyx 01:32:55Z)

In the intent-test environment, `apps/desktop` boots as a plain Vite web build (`tests/intent/stack/boot.ts`), never the Tauri shell — confirmed via that file's own comments ("Tier 2 needs the Desktop web build, not the Tauri shell" / "Browser mode has no Tauri bridge"). `window.__TAURI_INTERNALS__` is therefore never set, so `isTauriRuntimeAvailable()` (`apps/desktop/src/lib/access/tauri/connect-server.ts:7-9`) returns `false`.

Despite that, `DesktopProductHostProvider` (`apps/desktop/src/providers/DesktopProductHostProvider.tsx:204`) wires the concrete `desktopBridge` into the host **unconditionally**, and `useDesktopWorkerEnrollment` (`apps/packages/product-client/src/hooks/cloud/lifecycle/use-desktop-worker-enrollment.ts:117-134`) only gates on auth status — no Tauri check. It calls `ensureDesktopWorker` (`apps/packages/product-client/src/lib/workflows/cloud/ensure-desktop-worker.ts:46`), which calls `worker.getInstallId()` unconditionally. That resolves to `getDesktopInstallId` (`apps/desktop/src/lib/access/tauri/desktop-install-id.ts:3-4`), which calls `invoke()` from `@tauri-apps/api/core` **unguarded** — unlike sibling bridge methods `files.pickDirectory` and `diagnostics.supportSnapshot`, which already check `isTauriRuntimeAvailable()` first (`apps/desktop/src/lib/access/tauri/desktop-bridge.ts:125`, `:315`).

`invoke()` reads `window.__TAURI_INTERNALS__.invoke` internally, so with no Tauri shell present it throws `Cannot read properties of undefined (reading 'invoke')`. `ensureDesktopWorker`'s catch calls `deps.onFailure(error)`, which raises a sticky no-auto-dismiss "Integrations unavailable" toast. That toast region intercepts pointer events; under #1972's three-pane inspector layout, `#workflow-builder-input-1-required` sits under it, making tier-2 "Workflow definition lifecycle" time out (run 31982657482, `workflow-definition-lifecycle-traces` artifact, `error-context.md`).

Every link in the diagnosis chain checked out against the code on `main` at read time — no contradictions found.

## Fix

Narrowest env-side guard, at the seam the codebase already uses for exactly this shape (`updater.isSupported`):

- Added `isSupported(): boolean` to `DesktopWorkerBridge` (`apps/packages/product-client/src/host/desktop-bridge.ts`).
- Wired it to `isTauriRuntimeAvailable` in the concrete Desktop bridge (`apps/desktop/src/lib/access/tauri/desktop-bridge.ts`).
- `ensureDesktopWorker` checks `worker.isSupported()` before any native call and returns `false` silently if absent — mirroring the existing `pendingTicketPolicy !== "newest_wins"` no-op skip already in the same function (an expected-environment-shape outcome, not a failure: no `captureException`, no `onFailure`, no toast).

Explicitly did **not**: touch the toast system, touch canvas/builder code, or touch #2002/#2003.

## Tests + negative control

Added `"skips enrollment silently when the bridge has no native transport"` to `ensure-desktop-worker.test.ts`, asserting `getInstallId`/`enrollDesktopWorker` are never called and no `onFailure`/`captureException` fire.

Negative control — reverted the `isSupported()` guard locally and reran:

```
 × ensureDesktopWorker > skips enrollment silently when the bridge has no native transport 2ms
 - Expected
 + Received
 - false
 + true
```

Restored the guard; full suite green after (7/7 in `ensure-desktop-worker.test.ts`, 18/18 in `use-desktop-worker-enrollment.test.tsx`).

Ran via vitest against this worktree's live source (product-client/product-client + desktop `node_modules` reused read-only from an existing local checkout at the same base commit — no fresh install, no cargo/rustc, no Docker).

## Tier-2 e2e proof

#1972 stays merged (fix-forward is the ruling), so the current three-pane inspector layout is the real target, not a pending decision — correcting the earlier note above.

**Executed and verified (unit level):**
- `ensure-desktop-worker.test.ts`: added a guard-pinning test (`"skips enrollment silently when the bridge has no native transport"`).
- Negative control — reverted `isSupported()` locally, reran: `expected true to be false` (literal failing assertion above). Restored the guard: 7/7 in `ensure-desktop-worker.test.ts`, 18/18 in `use-desktop-worker-enrollment.test.tsx`, all green.
- Spot-checked the compiled output directly (bypassing pnpm, via `tsc -p tsconfig.json` on the package alone): the emitted `dist/lib/workflows/cloud/ensure-desktop-worker.js` carries the `isSupported()` guard.

**Local tier-2 Playwright run: attempted, infeasible.** Booted an isolated `t2verifyfix` profile (`SKIP_RUST=1`, prebuilt `ANYHARNESS_DEV_RUNTIME_BIN`, scrubbed env, never touched the running `wf2pablo` stack/ports) against a from-scratch worktree of this branch. `tests/intent/stack/boot.ts` boots desktop-web via `pnpm exec vite`; since the worktree only had node_modules symlinked in read-only from an existing checkout (to avoid a real install), pnpm's own recursive workspace dependency-status check rejected the mismatch and aborted non-interactively with `ERR_PNPM_ABORTED_REMOVE_MODULES_DIR_NO_TTY` — there is no seam to skip that check short of a genuine `pnpm install`, which the machine constraints for this task rule out ("no pnpm install into fresh trees if avoidable"). Vite never started; Playwright timed out waiting on `http://127.0.0.1:1532`. Not forced past this — no fake pass.

**Plan of record for the e2e proof:** current CI on this PR fails intent legs upstream of any spec execution, at `pnpm --filter @proliferate/product-client build` (`TS2741`, `run-view-model.ts:75`, `'cancelled'` missing from `RUN_STATUS_TONE`) — a known pre-existing defect on `main` until #2003 merges. Once #2003 lands, `gh pr update-branch 2011` against the fresh merge-ref will let the product-client build succeed and the provisional intent legs actually execute "Workflow definition lifecycle." That CI run at the updated head is the tier-2 greening proof that should gate this PR's merge.

**No merge** — this PR is up for review only, per instructions.


pablonyx (commented):
Reviewed at 1622831. Scope check: fix is transport-guard-only.

- `isTauriRuntimeAvailable` reused as `worker.isSupported` (apps/desktop/src/lib/access/tauri/desktop-bridge.ts) — same predicate already gating other Tauri-only entry points in this file (`connect-server`, `store.ts`), so this is consistent with existing doctrine, not a new pattern.
- `DesktopWorkerBridge.isSupported()` added to the shared interface (apps/packages/product-client/src/host/desktop-bridge.ts) with a comment explaining why callers must skip enrollment rather than invoke a native command that isn't there.
- `ensureDesktopWorker` (apps/packages/product-client/src/lib/workflows/cloud/ensure-desktop-worker.ts) returns `false` early when `!worker.isSupported()`, mirroring the existing pending-ticket-fencing early-return already in this function — no toast/global suppression logic touched, no canvas/workflow-builder files touched.
- Real desktop path unaffected: when a Tauri transport is present, `isSupported()` returns true and the function falls through to the existing `getInstallId` → `enrollDesktopWorker` → `ensure` flow unchanged; real errors still hit `captureException`/`onFailure` as before.
- Interface change is type-sound. Checked all constructors of `DesktopWorkerBridge`/objects typed as `worker: {...}`:
  - Only one real implementor exists (`desktop-bridge.ts` worker export) — updated.
  - `DesktopProductLifecycleRoot.test.tsx` mocks `worker: {}` on a broader host-bridge mock, not a literal `DesktopWorkerBridge`, so it's unaffected.
  - `use-desktop-worker-enrollment.test.tsx` uses `const worker = {} as DesktopWorkerBridge` (a type-cast stub, not a literal-typed object), so it compiles despite lacking `isSupported`.
  - `ensure-desktop-worker.test.ts` mock was updated to include `isSupported` and defaults it to `true` in `beforeEach`, so all pre-existing cases keep exercising the enrollment path.
- New test "skips enrollment silently when the bridge has no native transport" pins the failure mode: asserts `getInstallId`/`enrollDesktopWorker` are never called and `onFailure`/`captureException` stay silent when `isSupported` is false. By inspection this is a valid negative control — removing the `if (!worker.isSupported())` guard would let the mocked `getDesktopInstallId`/`enrollDesktopWorker` (which resolve successfully in `beforeEach`) run, and the `not.toHaveBeenCalled()` assertions would fail.
- No lint/exception weakening, no dead code, direct imports throughout (`@proliferate/product-client/host/desktop-bridge`, `./connect-server`), consistent with specs/frontend/README.md conventions.

Couldn't execute vitest in this environment (no node_modules installed in this worktree and avoiding a fresh install per review constraints), so the above is a static/logical trace rather than an executed run — worth a real `vitest run` pass before merge to confirm.

Verdict: CLEAN, no defects found.

PR #2012, opened 2026-08-17
feat(product): background work finish signals (bgwork R5)

pablonyx:
## What

Final repo-side rung of the Background Work slice — the finish-signal ladder from `HANDOFF-background-work.md` ("Finish signals are a ladder"): tab dot (quietest) → in-pane `NoticeBanner` → wake receipt + new turn (already shipped, existing `AgentOriginPromptReceipt`).

**1. `PanelHeaderEntry` dirty dot** (handoff — "clears on select... never render `dirty` on an `active` entry"; Delivery Spec rung R5)
- New domain module `domain/activity/background-work-finish-signal.ts`: `deriveLatestBackgroundWorkFinishSignal` (reads a process's `endedAt` straight off the live roster — processes never leave it; combines with a cached subagent-finish observation) and `deriveBackgroundWorkDirty` (pure timestamp comparison).
- New hook `hooks/activity/derived/use-background-work-finish-signal.ts` (read side) and `hooks/activity/lifecycle/use-background-work-finish-signal-tracking.ts` (write side — a ref-diffing observer, since native subagents vanish from the roster instantly on finish with no durable timestamp; mounted in `SessionTranscriptPane` so it's never missed regardless of which right-panel tool tab is active).
- New session-scoped, non-persisted store state: `backgroundWorkLastViewedAtBySession`, `backgroundWorkLastFinishedSubagentBySession` (same ephemeral placement as R4's `pendingBackgroundSubagentSelectionByWorkspace`).
- Threaded `backgroundWorkDirty` through `use-right-panel-controller` → `RightPanelFrame` → `RightPanelHeaderTabs` → `RightPanelHeaderEntryList`, with a defensive `!isActive` guard at the render site alongside the store-level clear-on-select.

**2. `NoticeBanner` in-pane notice** (handoff — "Pane open; `NoticeBanner` above the roster names what finished, with a View action")
- `BackgroundWorkPane` now renders a `NoticeBanner` (`tone="neutral"`, `CircleCheck` in `text-success`) above the roster while open and a finish is unseen.
- Process finishes: View action (ghost `Button`) opens the real `BackgroundTerminalView` — processes stay live in the roster.
- Subagent finishes: **no View action** — see Contradiction below.
- Dismissal (handoff silent on the exact rule, resolved minimally): banner hides while either detail view is open (a full swap of the pane's return) or the pane unmounts/switches session; it reappears on Back within the same session mount. Uses a "frozen at mount" baseline to avoid self-erasing against the live mark-viewed-on-select effect that also lights rung 1.

**3. Wake receipts vs roster step-down — verification only**
`AgentOriginPromptReceipt` already renders the wake turn; nothing there needed changes. Added domain-level tests (`background-work-row.test.ts`) proving counts and the roster step down together (both are pure functions of the same `{ loops, processes, agents }` snapshot — a count and a roster row can never disagree), that a finished process flips to Closed scope rather than leaving, and a negative control that `session/cancel` does not move the count (re-deriving from an identical snapshot is idempotent; only the roster itself finishing the work moves it).

**4. Indicator settled state**
No changes — `BackgroundWorkTranscriptRow` already ships the `CircleCheck` glyph swap, `"{n} background task{s} finished"` copy, and no-motion static swap from R1, verified still passing.

## Contradiction (stop-and-report, per R3/R4 precedent)

The handoff/delivery-spec ask the `NoticeBanner`'s View action to work symmetrically for both process and subagent finishes. `BackgroundWorkPane.test.tsx` pins (via an explicit, already-passing R4 test — "bounces back to the roster once a viewed subagent leaves it (finished)") that any selected subagent id the live roster can't resolve bounces back to the roster. A finished native subagent has, by definition, always already left that live roster (locked design, `chips.ts`) — there is no live detail seam a View click could ever land on without immediately bouncing back out. Rather than override that tested R4 decision, the subagent branch of the banner omits the View action and only names what finished. Flagging for founder review rather than silently reworking pinned R4 behavior.

## Disclosed, licensed as-is (NIT, review round 2)

The banner can fire for a stale finish on a session's very first-ever open (before any `markBackgroundWorkViewedForSession` has ever run for it). This is licensed by the handoff's "Pane open" condition — the banner's job is "tell me about anything I haven't seen in this pane," and on a first-ever open nothing has been seen yet. Left as-is by direction; disclosed rather than silently patched.

## Review round 2 — fixes applied

Coordinator review returned 1 MAJOR, 4 MINOR, 1 NIT (NIT addressed above). Rebased onto `origin/bgwork/r4-background-subagent-view@0d38ccad7` (mechanical design-attribution rename, unaffected by this diff) before fixing.

**MAJOR — `use-background-work-finish-signal-tracking.ts` ignored its own `sessionId` parameter.**
The tracker (and its sibling read hook) called `useSessionActivity()`, which is hardwired to `useActiveSessionId()` and silently ignores the explicit `sessionId` argument — a real API-contract bug, and one that could stamp a wrong-session finish with a fabricated `Date.now()` that outranks a process's real `endedAt`.

Investigated whether a per-session slice exists client-side for inactive sessions: it does. `session-directory-store.ts`'s `entriesById[sessionId]` is genuinely keyed per session (not active-only), and `lib/domain/sessions/hot-session-policy.ts` keeps several classes of inactive sessions' slices live (open tabs, sessions with an in-flight turn, cross-workspace running sessions) via `use-hot-session-ingest.ts`. So the fix is the "real fix" path, not the fallback: added `useSessionActivityForSession(sessionId)` to `use-session-activity.ts`, reading the per-session store slice directly, and switched both the tracker and the read hook to use it. `useSessionActivity()` itself is now a thin wrapper (`useSessionActivityForSession(useActiveSessionId())`) — its existing call sites (`BackgroundWorkPane`, `use-background-work-row.ts`, `use-workspace-status-model.ts`) are behaviorally unchanged.

This closes the API-contract bug but not the full multi-session concurrent-tracking case, disclosed rather than silently absorbed: this hook is mounted exactly once, for whichever session `SessionTranscriptPane` is currently rendering (always the active session today). A subagent finishing in a background session is detected on switch-back, not live — but this is a mount-cardinality choice, not a data-availability limit (the data is live; nothing consumes it live), because rung 1/2 of the ladder (the dot, the banner) are themselves both scoped to `useActiveSessionId()` only — there is no right-panel surface that could render a non-active session's signal even if the tracker watched it concurrently. Compatible with ruled D6 (per-session signal, no workspace-level persistence): a session's signal only needs to be correct once that session is the one being viewed.

Also fixed the ranking-safety half regardless of data hotness, since a subagent's finish is never independently timestamped: `background-work-finish-signal.ts` now tracks `detectedAtMs` (when the client noticed) as distinct from a process's real `endedAt`, and applies a deterministic tiebreak — an unseen process always outranks an unseen subagent detection, even when the raw detection timestamp is numerically later. Documented residual gap: if a subagent's session went cold long enough that detection lagged behind an unseen process's real, later `endedAt`, this still names the process — rare (needs both unseen AND the subagent's session cold) and understates rather than overstates freshness.

Rewrote the tracker/read-hook tests: the old mock was a single module-level variable reassigned in lockstep with whichever `sessionId` prop a test happened to pass — structurally unable to distinguish per-session from active-session-only visibility (it would pass under either implementation). Replaced with a `sessionActivityBySessionId` map so a session's own data can diverge from whatever session id the hook is currently called with, plus a dedicated test: a subagent finishes in session A while the tracker watches session B, then switches back to A — asserting the finish is detected on switch-back AND, fed through the real ranking function alongside a process with a real, later `endedAt`, the process still wins.

**MINORs:**
- `background-work-finish-signal.test.ts`: added the running-process + just-finished-subagent regression test (non-suppression, pinned by a snapshot assertion, not by reading the code).
- `RightPanelHeaderEntryList.test.tsx` (new): dedicated component-level test for the `dirty && !isActive` gate — dirty+active → no dot, dirty+inactive → dot, non-background tool entries never dirty regardless of `backgroundWorkDirty`.
- Acceptance "wake turn renders the receipt": already covered pre-existing, cited rather than duplicated — `AgentOriginPromptReceipt.test.tsx`'s `it.each(["completed","failed","cancelled"])("renders a resolved wake outcome %s on the right", ...)` renders the component with `wakeProvenance()` (`type: "subagentWake"`) and asserts the receipt DOM + outcome text. No new test added; this machinery predates R5's diff.

Fallout from the store field rename (`atMs` → `detectedAtMs`, to keep detection-time vs finish-time visible at the type level): `BackgroundWorkPane.test.tsx`'s module mock of `use-session-activity` needed `useSessionActivityForSession` added alongside `useSessionActivity` — fixed.

## Gates (review round 2, re-run after all fixes)

- Full `product-client` vitest: **1011 files / 7314 tests passed**, 0 failed.
- `tsc -p tsconfig.domain.json` and `tsc -p tsconfig.json --noEmit`: **1 error each**, both the pre-existing exogenous `src/domain/workflows/run-view-model.ts(75,7)` TS2741 — no new errors.
- `scripts/report_frontend_structure.py --strict`: **0** violations across all categories.
- `scripts/check_appearance_scaling.py`: **passed**.
- `scripts/check_design_attribution.py`: **passed** (0 hits).
- No raw `button`/`input`/`label`/`select`/`textarea` introduced.

## Files (review round 2 delta)

- Modified: `domain/activity/background-work-finish-signal.ts(+.test.ts)`, `hooks/activity/derived/use-session-activity.ts`, `hooks/activity/derived/use-background-work-finish-signal.ts(+.test.ts)`, `hooks/activity/lifecycle/use-background-work-finish-signal-tracking.ts(+.test.ts)`, `stores/preferences/workspace-ui-store-types.ts`, `stores/preferences/workspace-ui-right-panel-actions.ts`, `BackgroundWorkPane.test.tsx` (mock fallout fix)
- New: `RightPanelHeaderEntryList.test.tsx`

Base: `bgwork/r4-background-subagent-view` @ `0d38ccad7` (rebased from `4430513d6`, verified unmoved since). Draft, no merge, no auto-merge.

## CI red fix — PROD-SIZE-1 file split (post review-round-2)

CI's "Repo shape checks" job flagged `BackgroundWorkPane.test.tsx` at 690 lines against the repo-wide 600-line cap (`scripts/check_max_lines.py`, PROD-SIZE-1) — accumulated R2-R5 review-round additions; R4's head was under the cap. Net-new size exceptions are founder-gated, so split rather than exempted.

Mechanical split, no product code touched, no assertions added/removed/weakened:
- `BackgroundWorkPane.test.tsx` (281 lines) — roster/scopes/seam/detail-routing, 12 tests.
- `BackgroundWorkPane.finish-signals.test.tsx` (new, 382 lines) — the R5-era finish-signal ladder concerns: NoticeBanner, pending-selection consumption, feed isOpen-gating, 12 tests.
- `BackgroundWorkPane.test.fixtures.tsx` (new, 145 lines) — shared mock components (`LiveTerminalsRosterPanelMock`, `BackgroundTerminalViewMock`, `AgentsRosterPanelMock`, `BackgroundSubagentViewMock`) and value builders (`makeProcess`, `makeAgent`) both files import into their own `vi.mock(...)` calls, so neither re-declares them. Named with a literal `.test.` segment (not `.test-fixtures`) so `report_frontend_structure.py`'s `should_skip` recognizes it as test-only scaffolding (exempting its mock JSX from the shipped-component raw-DOM-control rule, FE-STRUCT-1) and so vitest's `*.test.tsx` include glob does not collect it as a test suite itself.

Test count preserved exactly: 24 tests before (1 file), 24 tests after (12 + 12 across 2 files).

Gates re-run clean after the split:
- `python3 scripts/check_max_lines.py`: **passed** (was failing on this file at 690/600).
- Full `product-client` vitest: **1012 files / 7314 tests passed** (was 1011/7314 before the split — file count +1 from the new sibling test file, total test count unchanged).
- `tsc -p tsconfig.domain.json` / `tsc -p tsconfig.json --noEmit`: 1 error each, both the same pre-existing exogenous `run-view-model.ts(75,7)` TS2741.
- `scripts/report_frontend_structure.py --strict`: **0** violations.
- `scripts/check_appearance_scaling.py`: **passed**.
- `scripts/check_design_attribution.py`: **passed**.

New head: `6c60ade04`.

### DIFF
--- a/apps/packages/product-client/src/components/workspace/activity/background-pane/BackgroundWorkPane.finish-signals.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/activity/background-pane/BackgroundWorkPane.finish-signals.test.tsx
@@ -1,0 +1,382 @@
+/* @vitest-environment jsdom */
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import type { BackgroundWorkRowCounts } from "#product/domain/activity/background-work-row";
+import type { SessionActivityState } from "#product/hooks/activity/derived/use-session-activity";
+import { useWorkspaceUiStore } from "#product/stores/preferences/workspace-ui-store";
+import {
+  AgentsRosterPanelMock,
+  BackgroundSubagentViewMock,
+  BackgroundTerminalViewMock,
+  LiveTerminalsRosterPanelMock,
+  makeAgent,
+  makeProcess,
+} from "./BackgroundWorkPane.test.fixtures";
+import { BackgroundWorkPane } from "./BackgroundWorkPane";
+
+// R5 finish-signal ladder concerns: pending-selection consumption, feed
+// isOpen-gating, and the NoticeBanner (dirty/mark-viewed interactions). The
+// roster/scopes/seam/detail-routing coverage lives in the sibling
+// `BackgroundWorkPane.test.tsx` — split for PROD-SIZE-1 (repo-wide 600-line
+// cap); see the fixtures module's docstring. No behavior or assertion
+// changed by the split.
+let sessionActivity: SessionActivityState = {
+  loops: [],
+  loopCapabilities: { supported: false, native: false },
+  processes: [],
+  agents: [],
+};
+let rowCounts: BackgroundWorkRowCounts = { runningCount: 0, finishedCount: 0 };
+
+vi.mock("#product/hooks/activity/derived/use-session-activity", () => ({
+  useSessionActivity: () => sessionActivity,
+  // `useBackgroundWorkFinishSignal` (feeding this pane's NoticeBanner rung)
+  // reads the per-session accessor rather than the active-session-only one
+  // (R5 review round 2 — MAJOR fix). This pane is always scoped to the
+  // active session, so mirroring the same fixture here is faithful — the
+  // per-session-vs-active-session distinction itself is covered by
+  // `use-background-work-finish-signal.test.ts` and
+  // `use-background-work-finish-signal-tracking.test.ts`, not here.
+  useSessionActivityForSession: () => sessionActivity,
+}));
+vi.mock("#product/hooks/activity/derived/use-background-work-row", () => ({
+  useBackgroundWorkRowCounts: () => rowCounts,
+}));
+vi.mock("#product/components/workspace/activity/LiveTerminalsRosterPanel", () => ({
+  LiveTerminalsRosterPanel: LiveTerminalsRosterPanelMock,
+}));
+vi.mock("#product/components/workspace/activity/background-pane/BackgroundTerminalView", () => ({
+  BackgroundTerminalView: BackgroundTerminalViewMock,
+}));
+vi.mock("#product/components/workspace/activity/AgentsRosterPanel", () => ({
+  AgentsRosterPanel: AgentsRosterPanelMock,
+}));
+vi.mock("#product/components/workspace/activity/background-pane/BackgroundSubagentView", () => ({
+  BackgroundSubagentView: BackgroundSubagentViewMock,
+}));
+
+afterEach(() => {
+  sessionActivity = {
+    loops: [],
+    loopCapabilities: { supported: false, native: false },
+    processes: [],
+    agents: [],
+  };
+  rowCounts = { runningCount: 0, finishedCount: 0 };
+  cleanup();
+  useWorkspaceUiStore.setState({
+    pendingBackgroundSubagentSelectionByWorkspace: {},
+    backgroundWorkLastViewedAtBySession: {},
+    backgroundWorkLastFinishedSubagentBySession: {},
+  });
+});
+
+describe("BackgroundWorkPane — finish signals", () => {
+  it("disables the subagent feed while the right panel is collapsed, and re-enables it on reopen", () => {
+    sessionActivity = {
+      loops: [],
+      loopCapabilities: { supported: false, native: false },
+      processes: [],
+      agents: [
+        makeAgent({ id: "agent-live", feed: { feedId: "feed-2", kind: "transcript" } }),
+      ],
+    };
+    const { rerender } = render(
+      <BackgroundWorkPane workspaceId="ws-1" sessionId="sess-1" isOpen />,
+    );
+
+    fireEvent.click(screen.getByTestId("open-subagent-agent-live"));
+    expect(
+      screen.getByTestId("background-subagent-view").getAttribute("data-feed-enabled"),
+    ).toBe("true");
+
+    rerender(<BackgroundWorkPane workspaceId="ws-1" sessionId="sess-1" isOpen={false} />);
+
+    expect(screen.getByTestId("background-subagent-view")).toBeTruthy();
+    expect(
+      screen.getByTestId("background-subagent-view").getAttribute("data-feed-enabled"),
+    ).toBe("false");
+
+    rerender(<BackgroundWorkPane workspaceId="ws-1" sessionId="sess-1" isOpen />);
+
+    expect(
+      screen.getByTestId("background-subagent-view").getAttribute("data-feed-enabled"),
+    ).toBe("true");
+  });
+
+  it("consumes a pending subagent selection from the transcript click seam and clears it", () => {
+    sessionActivity = {
+      loops: [],
+      loopCapabilities: { supported: false, native: false },
+      processes: [],
+      agents: [makeAgent({ id: "agent-42" })],
+    };
+    // Delivery Spec — Background Work Slice 1, rung R4 fix-forward: a native
+    // subagent's transcript block click writes here via
+    // `useOpenBackgroundWorkPane`'s extended return, keyed by the same
+    // `workspaceId` this pane receives as a prop, carrying the session
+    // active at write time.
+    useWorkspaceUiStore.getState().setPendingBackgroundSubagentSelectionForWorkspace(
+      "ws-1",
+      { subagentId: "agent-42", sessionId: "sess-1" },
+    );
+
+    render(<BackgroundWorkPane workspaceId="ws-1" sessionId="sess-1" isOpen />);
+
+    const view = screen.getByTestId("background-subagent-view");
+    expect(view.getAttribute("data-subagent-id")).toBe("agent-42");
+    // One-shot: consumed and cleared, not left to reopen on a later render.
+    expect(
+      useWorkspaceUiStore.getState().pendingBackgroundSubagentSelectionByWorkspace["ws-1"],
+    ).toBeNull();
+  });
+
+  it("ignores a pending subagent selection for a different workspace", () => {
+    sessionActivity = {
+      loops: [],
+      loopCapabilities: { supported: false, native: false },
+      processes: [],
+      agents: [makeAgent({ id: "agent-42" })],
+    };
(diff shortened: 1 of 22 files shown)


pablonyx:
Refold receipt (stacked-squash defusal, same protocol as #2009/#2010): after #2010's squash (19d019e0d) this PR went CONFLICTING against its stacked base. Rebased onto current main from anchor `dfa96d214` (the old r4 tip, verified ancestor); the two keep-up merge commits drop (their content is the parent's squash), the four content commits replay **byte-identically** — all-`=` range-diff:

```
1:  7605793d7 = 1:  68408f9fc feat(product): background work finish signals (bgwork R5)
2:  cac92e37c = 2:  f643be3b6 fix(product): R5 review round 2 — per-session finish tracking + deterministic tiebreak
3:  6c60ade04 = 3:  737ef7d4d test(product): split BackgroundWorkPane.test.tsx for PROD-SIZE-1
4:  4fdd503db = 4:  005f4249e test(product): document the load-bearing .test. mid-name segment in the pane fixtures filename
```

Prior reviews carry (content unchanged, old head `ca5c6af27` → new head `005f4249e`). Base retargeted to main. Timeline note: the close/reopen just now was the CI re-fire — GitHub silently skips `pull_request` workflow scheduling while a PR is CONFLICTING/UNKNOWN-mergeable (a refolded head against the stale stacked base can't produce a test-merge commit), so the push's synchronize event got no board; retarget → MERGEABLE → reopen fired the full board at 00:49:31Z on this exact head. One fresh full CI round now; literal tally quoted here before any click.

pablonyx:
Fresh CI on refolded head 005f4249e settled **23 pass / 1 fail** — and the one red is infrastructure, not a test: the Transcript scroll physics job ([95558442774](https://github.com/proliferate-ai/proliferate/actions/runs/32085965255/job/95558442774)) hung on the **Install Playwright chromium + webkit** step and was cancelled at 30m15s (step conclusions: install = cancelled, "Run scroll-physics suite" = **skipped**). Zero scroll-physics tests executed, so no interaction with this PR's content is possible — the suite never started. Third distinct red signature this train, each disclosed: census flake (#2009), rv_2_05 timing flake (#2010), and now a runner browser-download stall. Same-ref rerun of the failed job dispatched; merge only on a green rerun, verdict quoted here.

pablonyx:
Rerun status, honestly: the same-ref rerun of the scroll-physics job ([95564226785](https://github.com/proliferate-ai/proliferate/actions/runs/32085965255/job/95564226785)) is hung on the SAME **Install Playwright chromium + webkit** step as the cancelled first attempt. This is now a proven repo-wide stall window, not this PR: a scroll-physics job started 01:03:33Z completed **success**, while all three jobs started after ~01:19Z — this rerun (01:21:47Z), plus runs on two unrelated branches (forks/lane-h at 01:19:51Z and pablo/pro-295 at 01:29:04Z) — are all stuck in the identical install step. Playwright browser downloads are stalling on the runners since ~01:19Z. Plan: let this attempt die, wait for the window to clear (signal = any scroll-physics job completing green repo-wide), then one more same-ref rerun. No merge until a genuinely green board; every attempt stays quoted here.

pablonyx:
Settled verdict on head 005f4249e: **24 checks, 24 pass / 0 fail / 0 pending**. The scroll-physics rerun recovered when the repo-wide Playwright-install stall window cleared and **passed** (22m41s including the hang, [job 95564226785](https://github.com/proliferate-ai/proliferate/actions/runs/32085965255/job/95564226785)) — zero tests were harmed in the making of this board; the only red across both attempts never got past browser install, exactly as disclosed above. Merging on Pablo's standing bg-work merge word.

PR #2018, opened 2026-08-17
feat(workflows): one run rail per concurrent run in the workspace pane (rung 3)

pablonyx:
Rung 3 of the frozen Workflows follow-up ADR (2026-08-17). Stacked on #2017 (rung 2). Client-only, per the ADR: "one run rail per run, sessions grouped under their run, context docs grouped per run. Exact treatment follows the existing run-view composition; no new visual primitives."

## What
- `selectVisibleWorkflowRuns`: every non-terminal run newest-first; with nothing live it collapses to the single newest run, so finished history renders exactly as today. Only `existing_workspace` placement (rung 2) can produce N > 1.
- `useWorkflowRunRoster` owns the watched roster; `useWorkflowPane` is re-parametrized per run.
- `WorkflowPane` maps visible runs onto `WorkflowRunRail` instances — the pre-existing composition (resume banner, graph, docked inspector, docs list), each rail with its own selection state and doc group. **Gate: with one visible run the DOM is byte-identical to before** (header chip stays on the shared PaneHeader, fed by the rail's polled projection); with N > 1 each rail gains one label line (`Run <id>` — WorkflowRunV2 has no title field) plus its status chip.
- The auto-advance watcher now watches every visible run via a fixed slot count (`MAX_WATCHED_CONCURRENT_RUNS = 4`, rules-of-hooks constraint), so a second concurrent run's auto-advance still raises its undo toast. Overflow beyond 4 misses a toast, never crashes.
- `run-view-model.ts` and doc opening (already run-scoped since rung 1) untouched.

## Testing
- Targeted suites: `Test Files 4 passed (4)` / `Tests 48 passed (48)` (run-selection, use-workflow-pane, auto-advance-toast, new WorkflowPane component test incl. single-run DOM-identity and two-rail independence checks).
- Negative control: forcing the roster to collapse to one run makes exactly the two multi-run tests fail (`surfaces two live runs newest first`, `watches two concurrent runs, each in its own slot`); restored, re-verified green.
- Full product-client vitest + typecheck deltas verified pre-existing by stash comparison (the known main-wide `run-view-model.ts(75,7)` TS2741 'cancelled' tone error; `@proliferate/design` dist-export test failures).
- `scripts/check_max_lines.py`: Max-lines check passed.

## Post-freeze ruling F-A2 (erratum #3): rail cap at four
Founder-ruled (2026-08-17 23:56:52Z): visible run rails cap at 4 — `selectWorkflowRunRailWindow` cuts a fixed-size window from the visible set, promoting runs waiting on a human (gate/approval/park) or failed onto the resting page; the "N more runs" overflow line pages the window; hidden runs stay reachable and are never silently hidden when they need someone. No census exception, no virtualization (PROD-SCALE-33 passes). RULED F-A2(i): attention-first is a STANDING ordering rule — the run waiting on a human renders first even when every run fits under the cap (a visible ordering change for under-cap multi-run workspaces, disclosed to the founder); pinned by "keeps attention-first ordering even when every run fits under the cap". Evidence: appearance-scaling + boundary lints green; suites 44/44 across run-selection/WorkflowPane/use-workflow-pane; negative control (uncapped map restored) fails exactly the two new pane tests.

### DIFF
--- a/cloud/sdk/src/generated/openapi.ts
+++ b/cloud/sdk/src/generated/openapi.ts
@@ -7397,7 +7397,9 @@ export interface components {
              * Mode
              * @enum {string}
              */
-            mode: "worktree" | "repo_root";
+            mode: "worktree" | "repo_root" | "existing_workspace";
+            /** Workspaceid */
+            workspaceId?: string | null;
         };
         /** WorkflowInvocationResponse */
         WorkflowInvocationResponse: {


pablonyx (commented):
Dedicated rung-3 review:

FINDING 1 — single-run gate violation: the outer container's loading/error state no longer reflects the per-run detail query, only the roster.

Before, pane.status came from resolveWorkflowPaneStatus combining both runsQuery (roster) and the per-run runQuery (detail projection) into one status. aria-busy and the outer wrapper's className (flex flex-col vs overflow-y-auto) both keyed off that combined status.

After this PR, the container's outer switch keys off roster.status only (useWorkflowRunRoster), and useWorkflowRunRoster's call to resolveWorkflowPaneStatus hardcodes projectionLoaded: true, projectionFailed: false — it has no knowledge of the per-run detail query at all. As soon as the runs list resolves, roster.status flips to "ready", the outer container gets flex flex-col and mounts WorkflowRunRail, and only then does the rail's own pane.status === "loading" branch fire — one level deeper, inside the extra wrapper div the map introduces.

Net effect for the ordinary single-run case: on every mount, there is a real (not rare) window — roster resolved, run detail still in flight — where the DOM differs from before: (a) outer wrapper class is flex flex-col instead of overflow-y-auto, (b) the loading spinner is nested one div deeper, (c) aria-busy on the section is false instead of true while the run detail is still loading. This is a genuine violation of the HARD GATE ("with <=1 visible run the UI renders identically to before"), since it isn't limited to the concurrent-run case — it changes the ordinary single-run mount sequence. Same applies to the error path.

Suggest: have WorkflowPane's outer aria-busy/class switch also incorporate the single rail's own resolved status (symmetric to how onHeaderRunChange/singleRunHeader is already threaded back up for the status chip).

Item 2 — selectVisibleWorkflowRuns vs selectNewestWorkflowRun: clean. Filters via workflowRunIsActive (running/awaiting_human/interrupted), sorts newest-first, collapses to selectNewestWorkflowRun's output when no active run exists. Covers all-terminal, one-live-plus-old-terminals, zero-runs correctly. No rendering-shape change without existing_workspace placement.

Item 3 — fixed-slot auto-advance loop: sound. previousProjectionRef/announcedAdvanceKeysRef both reset together on [runId] change, so a slot's identity shift clears dedup memory in lockstep with its diff baseline — no spurious re-announcement found. Hook count fixed at MAX_WATCHED_CONCURRENT_RUNS=4 regardless of visibleRuns.length.

Item 4 — per-rail isolation: each WorkflowRunRail mounts its own useWorkflowPane, own selectedNodeRowId state, own inspector; no shared mutable state between rails; per-run query keys correct.

Item 5 — test adequacy: WorkflowPane.test.tsx pins real DOM assertions (role="group" count, absence/presence of "Run {id}" text) rather than approximating — good identity pin for rail-count/label. It does NOT cover the loading/aria-busy/wrapper-class window from Finding 1 — all fixtures start from status:"ready" pane models, so that gate violation is untested.

No new visual primitives, no colored left-border ownership treatments — confirmed clean. Everything else (copy additions, SDK schema additions, run-selection tests) consistent with the frozen spec.

pablonyx:
Review finding addressed in a9d8295bb: the roster now probes the solo run's projection (react-query dedupes with the rail's query) and resolves status through resolveWorkflowPaneStatus exactly as the pre-rail pane did, so the single-run loading/error DOM (wrapper class, spinner nesting, aria-busy) is identical through the ordinary detail-fetch window. Two new roster tests pin the loading and error states; negative control: reverting the predicate fails exactly those two ('Tests 2 failed | 24 passed' -> restored 'Tests 26 passed').

pablonyx:
DELTA RE-REVIEW (reviewer of record, fix head 72c7cbb24): CLEAN

Verified all four points on the fix commit alone:
1. `useWorkflowRunRoster` now probes the solo run's detail via `useWorkflowRunQuery(soloRunId, { enabled: soloRunId.length > 0 })` and folds `soloRunQuery.data`/`isError` into `resolveWorkflowPaneStatus`, restoring the pre-rail loading/error DOM (including error-with-nothing-cached, per the new "is error when the single run's detail fetch failed with nothing cached" test).
2. With N>1 visible runs, `soloRunId` is `""`, the probe query is disabled, and status resolves to `ready` immediately — confirmed by "is ready without a detail probe when several runs are visible."
3. No hook-order hazard: `soloRunId` is unconditionally computed and `useWorkflowRunQuery` is unconditionally called every render with an `enabled` flag rather than being skipped — the conditional lives in the query's enabled predicate, not in whether the hook itself runs.
4. Ran `npx vitest run src/hooks/workflows/facade/use-workflow-pane.test.tsx` — 26/26 pass. Negative control: reverted `projectionLoaded`/`projectionFailed` to the old `true`/`false` literals and reran — exactly the two new roster tests fail (loading-through-detail-fetch, error-with-nothing-cached), all other 24 still pass. Restored the file after.

pablonyx:
Review chain closed at 72c7cbb24 (delta re-review CLEAN by reviewer of record). Head moved to 43b4b4adc by a verified mechanical FE-PC-2 import fix (run-selection.ts, 1 file ±1 line). APPROVE-CARRIES per orchestrator, verified via the GitHub compare API.

pablonyx:
DELTA RE-REVIEW ROUND 2 (reviewer of record, fix head 0dbe2fb5d): FINDING

Checked (1)/(2)/(3)/(5) — all good:
1. Fixed-size window is genuine, not guard-gamed: `selectWorkflowRunRailWindow` always slices to `MAX_VISIBLE_RUN_RAILS` (4), no census exception, no virtualization escape hatch.
2. Attention promotion is correct and tested: `workflowRunNeedsAttention` (awaiting_human/interrupted/failed) runs are filtered ahead of the rest before slicing, so a gated/parked/failed run can never fall behind the overflow line while a merely-running run occupies its slot — confirmed by the 'promotes runs waiting on a human onto page 0' test and its WorkflowPane.test.tsx analogue.
3. Page clamping is safe: `page = Math.min(Math.max(requestedPage, 0), pageCount - 1)` with `pageCount = Math.max(1, ...)`, so a page index that no longer exists after the run set shrinks clamps back into range rather than stranding an empty page — confirmed by the 'pages the window and clamps a page the shrinking set no longer has' test.
5. Reverting the cap (making `selectWorkflowRunRailWindow` return all runs unsliced) would fail the 'caps the window at four' test, the WorkflowPane 5-run overflow test, and the paging/clamp test — the tests do pin the capped behavior.

(4) does NOT hold as stated — real finding:

`selectWorkflowRunRailWindow` reorders the *entire* prioritized list (attention-first, then rest) before slicing, unconditionally — not only when the cap would otherwise hide an attention run. So for a visible-run set that already fits inside the cap (≤4, `hiddenCount === 0`, no overflow line), if any run in that set needs attention, the render order still changes from the pre-existing newest-first order to attention-first.

Concrete repro: runs `[c (newest, running), b (running), a (oldest, awaiting_human)]` — 3 runs, well under the cap. Pre-commit `roster.visibleRuns` order (and pre-commit render order) is newest-first: `c, b, a`. Post-commit `selectWorkflowRunRailWindow(visibleRuns, 0).railWindow` returns `a, c, b` — verified by direct evaluation of the committed selector logic. No overflow line renders (`hiddenCount: 0`), so nothing in the UI signals that reordering happened; a workspace with a gated run buried behind two newer running runs will visibly jump that gated rail to the top of the stack even though every rail was already going to render.

Neither `run-selection.test.ts` nor `WorkflowPane.test.tsx` covers a fits-under-cap set with a non-newest attention run, so this reorder is untested and the ask's requirement (4) — 'byte-identical to before … same order' for the ≤4 case — is not actually met whenever an attention run is present in an otherwise-fitting set.

This may well be an intentional UX call (surface attention runs first even without paging), but as written it contradicts the review ask's stated invariant and isn't covered by a test either way — founder should confirm intent before this is called done.

pablonyx:
Head moved 0dbe2fb5d → de3a37a9d by a mechanical FE-PC-4 fix: the new domain test's local `window` variable renamed to `cut` (the lint pattern-matches the browser global by token). 1 test file, rename only; boundary check now passes, suite 13/13. The round-2 semantic finding (under-cap attention reorder) remains OPEN pending founder intent — this carry does not touch it.

pablonyx:
Round-2 finding RESOLVED BY RULING F-A2(i) (orchestrator, founder provenance): attention-first is a standing ordering rule, not overflow-only — the reviewed behavior is the intended one. The stated round-2 invariant (4) was mine and was wrong; the ruled semantics are "capped at 4, newest-first with active runs prioritized" unconditionally. Pinning test added at 35daba3b1 ("keeps attention-first ordering even when every run fits under the cap"), suite 14/14; disclosed on the founder's morning material as a visible under-cap ordering change. Heads: r3 35daba3b1, carried r4 debf370d5 / r5 e36032a4b.

pablonyx:
Carry note (head 35daba3b1 → eb5033e70): rebase-only onto the amended r2 (which gained its committed cloud-SDK regen). Zero content change in this rung's own diff — `git range-diff` shows the rung-3 commits carry unchanged. Workflow domain+pane vitest at the new stack tip: "Tests 255 passed (255)".

pablonyx:
Carry receipt (orchestrator-ordered): `git range-diff fa6137094..35daba3b1 d8c3a1d75..eb5033e70` — all six rung-3 commits carry byte-equal (`=`) across the rebase:

```
1:  dc57ae360 = 1:  4b3ff3f49 feat(workflows): one run rail per concurrent run in the workspace pane (rung 3)
2:  72c7cbb24 = 2:  e821f2f2e fix(workflows): keep the single-run pane's loading and error DOM identical through the detail fetch
3:  43b4b4adc = 3:  2ec732695 fix(workflows): import the sibling domain module relatively to satisfy FE-PC-2
4:  0dbe2fb5d = 4:  c4df170f6 feat(workflows): cap the run rails at four with a paging overflow line (ruling F-A2)
5:  de3a37a9d = 5:  8fd66de17 fix(workflows): rename the test's 'window' local so the domain test passes FE-PC-4
6:  35daba3b1 = 6:  eb5033e70 test(workflows): pin attention-first ordering as a standing rule (ruling F-A2(i))
```

No content delta at this rung; review chain of record stands at head eb5033e70.

pablonyx:
## REFOLD RECEIPT — rung 3 (r3-concurrent-run-ui)

Rebased onto the refolded rung 2 as part of the whole-chain refold (Pablo's order; continued inline by the orchestrator after the refolder agent was killed mid-chain by a session steer).

- **Old head:** `eb5033e70` (on old r2 `d8c3a1d75`)
- **New head:** `f10c191e47c897627078ba3114231a6f848f862e` (on new r2 `f10c191e4`)
- **Conflicts:** none — clean replay.
- Range-diff: all own commits replayed 1:1 (see PR timeline; commit contents unchanged apart from context shifts).

Fresh CI baking; merges bottom-up behind r1/r2 on CLEAN verdicts.

pablonyx:
CORRECTION to receipt 5332082984 (posted in error): that refold receipt was WRONG. The r3 rebase actually hit a CONFLICT; a scripting bug (pipe swallowing the rebase exit code) briefly force-pushed rung 2's tip onto this branch and posted a false clean-replay receipt. The branch has been RESTORED to its true head eb5033e70 (verify: current branch tip). No merge occurred, PR content is back to its reviewed state. The conflict is now being resolved by hand; a correct refold receipt will follow. Any CI runs minted against the bogus intermediate tip should be disregarded.

pablonyx:
Superseded by https://github.com/proliferate-ai/proliferate/pull/2057 — GitHub refuses to reopen this PR because the head ref moved after the auto-close (rebase-incident head==base moment, corrected in 5332090336). The review chain recorded here (CLEAN through delta round 2 at eb5033e70) carries to the replacement, whose body links back.

PR #2016, opened 2026-08-17
feat(workflows): run-scoped context docs — .proliferate/context/<runId>/ (follow-up ADR rung 1)

pablonyx:
## Summary

Rung 1 of the frozen Workflows Follow-up ADR moves workflow context documents from the flat `.proliferate/context/` directory to `.proliferate/context/<runId>/`. Concurrent runs that share a workspace can no longer collide on identical `NN-slug.md` filenames.

## Implementation

- Adds one run-scoped path helper and threads the run ID through planned materialization, PUT acceptance, actor envelope rendering, prompt references, and preamble rendering.
- Updates the ProductClient doc opener to target `.proliferate/context/<runId>/<filename>`.
- Updates the T3 reference scenario and testing record to the run-scoped layout.
- Preserves the parent `.proliferate/context/` git-exclude entry.
- Adds a focused ProductClient regression test that pins both viewer-open calls to the scoped path and rejects the obsolete flat target.

## Testing

- `CARGO_BUILD_JOBS=1 cargo test -p anyharness-lib workflows --lib` — 166 passed, 0 failed.
- `pnpm --filter @proliferate/product-client exec vitest run src/hooks/workflows/ui/use-workflow-doc-open.test.tsx src/lib/access/cloud/workflow-surface-boundary.test.ts --maxWorkers=1` — 3 passed.
- `pnpm --filter @proliferate/release-e2e exec tsx --test src/scenarios/t3-wf-1.test.ts` — 8 passed.
- `pnpm --filter @proliferate/product-client typecheck` — passed.
- `python3 scripts/check_frontend_boundaries.py` — passed.
- `python3 scripts/check_design_attribution.py` — passed.
- `python3 scripts/check_max_lines.py` — passed.
- `python3 scripts/check_docs.py` — passed.
- `git diff --check` — passed.

### TESTING

The path/render invariants are Tier 1. The workflow-domain suite covers materialization, rendering, store, transition, and lifecycle behavior, including the two-run collision negative control. T3 remains planned release coverage; its plan-level assertions are pinned here.

### OBSERVABILITY

No telemetry, log schema, span identity, or error-code behavior changes. Filesystem paths and document contents are not newly emitted.

## Migration

No schema migration or backfill. Context documents are ephemeral, gitignored workspace files. Historical flat files are neither moved nor deleted; newly materialized/current projections use the run-scoped layout.

## Review and reconciliation

- Frozen contract: Workflows Follow-up ADR, 2026-08-17, rung 1 plus erratum 1.
[...]

PR #2020, opened 2026-08-17
feat(workflows): fan-in ledger core behind the one-leg grammar (F1+F3, rung 4)

pablonyx:
Rung 4 of the frozen Workflows follow-up ADR (2026-08-17). Stacked on #2018 (rung 3). Pure fan-in plumbing (rulings F1 + F3); the definition grammar still cannot express N > 1 legs, so with every real definition exactly one ledger row exists per node. **Gate: one-leg behavior bit-identical.**

## What
- New durable ledger `workflow_run_node_sessions` (migration `0072`): `node_row_id, leg_index, session_id, status, completed_at, UNIQUE(node_row_id, leg_index)`. `stamp_session` upserts the representative leg (leg 0) in the same transaction; relaunch resets instead of colliding on the UNIQUE key.
- `TurnFinished` grows `session_id` (R6), threaded from the production report path (`session_extension.rs`); tests pass `None` and fall back to the node's representative session.
- `RunState` carries the ledger slice (`node_legs` + `legs_of()`), so the pure table answers the completion question without the store. The chain tail of `on_turn_finished` moves to new `fanin.rs`: hold on queued interjection, `RecordLegThenHold` while any sibling leg is outstanding, else aggregate — cancel > forced-unload > fail iff any leg failed (F1) > clean advance/gate.
- Undo window (F3): parallel nodes stamp `first_turn_finished_at` inside the `AdvanceToNext`/`CompleteRun` commit; `note_first_turn_finished` no-ops for parallel nodes; one-leg nodes keep the per-report stamp so the observable end state is unchanged (first == last).
- `store.rs` row mappers relocated verbatim to `store_rows.rs`; its ratchet SHRINKS 1377 → 1315. No ratchet raised. Redo/resume/SDK/grammar untouched (rungs 5–7).

## Testing
- Workflows superset filter: `test result: ok. 196 passed; 0 failed` (prior 186 + 10 new).
- New tier-1 `transition_fanin_tests.rs` (7): first-leg-clean holds, last-leg-clean advances, failure fails only once all legs terminal, cancelled-leg interrupt, arrival-order invariance, queued-interjection hold.
- New `store_fanin_tests.rs` (3) over the real migration ladder incl. relaunch-reset.
- Negative control (run by the implementer independently of the in-suite one): disabling the outstanding-leg gate makes exactly `first_leg_clean_holds_without_advancing` and `a_recorded_failure_fails_the_node_only_once_all_legs_are_terminal` fail; restored: `ok. 7 passed; 0 failed`.
- Persistence: schema snapshot regenerated (`specs/GENERATED/anyharness-db-schema.sql`), `anyharness_schema_snapshot_matches_migrations ... ok`, 36 other persistence tests green.
- `scripts/check_max_lines.py`: Max-lines check passed.

Pre-existing tests modified: only `TurnFinished` literals gaining `session_id: None` (forced by the new field) and comment compactions for ratchets; no behavioral pin changed.

## Design note: Fence legs are intentionally non-terminal (ruling F6)

A boot fence parks the run; it does not resolve it. Ruling F6 makes crash-resume truncate the fan-in ledger and re-fan-out all N legs, so a parked leg's stale `running` status never survives to a user-visible read — the resume path resets it before anything reads it. Terminalizing legs at fence time would fabricate an outcome the turn never had. The reviewer of record reviewed this citation and accepted it (delta re-review comment on this PR).

## Review fixes on this rung

1. `finished_leg_of` originally never terminalized legs on `Transition::Cancel` — fixed in e0dd32f16 (reviewer of record's finding).
2. The delta re-review found that fix under-scoped: Cancel disposes every Running row's session (chain **and adhoc**), but only the current chain node's legs were stamped, leaving a running adhoc row's leg `running` forever. Fixed in 9abcf4bc9: Cancel is run-terminal and stamps every still-running leg of the run in the same commit (`cancel_all_run_legs_tx`); already-terminal legs keep their status and timestamps. Regression test `cancel_marks_every_running_leg_in_the_run_cancelled` (chain + adhoc). Cargo verification deferred to CI + the post-hold local run, per the standing machine hold.


pablonyx (commented):
Dedicated rung-4 review: fan-in ledger core (F1+F3) behind the one-leg grammar.

**One-leg gate (1):** Verified `fanin::resolve_chain_turn` reduces exactly to the old `on_turn_finished` tail for every stop-reason branch. Queued-interjection Hold check is identical. With one leg, `outstanding` is always false (no sibling leg exists), so `aggregate_node` short-circuits on `this_leg` alone — Cancelled/ForcedUnload/Failed all return their transition before the node-type match, matching old unconditional-InterruptNode/FailNode behavior; Clean falls through to the same Agent/HumanInLoop dispatch. No observable difference found.

**F1 aggregation (2):** Priority order cancel > forced-unload > fail > clean is correct in `aggregate_node`. `upsert_leg_tx` always writes a non-null `session_id` (parameter is `&str`, not `Option`), so the "ledger session_id is NULL" case can't arise via the normal stamp path today — moot at rung 4, worth re-checking once redo/relaunch paths multiply.

**F3 (3):** `stamp_first_turn_if_parallel_tx`'s `COUNT(*) > 1` guard and `note_first_turn_finished`'s `COUNT(*) <= 1` guard are complementary and keyed off the same ledger row count, so with exactly one ledger row (every definition today) only the per-report path ever fires — byte-identical to pre-ledger. Confirmed via `undo_window_closes_after_first_turn_and_reopens_on_reexecution`.

**Migration 0072 (4):** Confirmed via `run_migrations`: plain `MIGRATIONS` (incl. 0072) run first, then `CUSTOM_MIGRATIONS`, then `CUSTOM_FOREIGN_KEY_MIGRATIONS` (which contains `0069_workflow_runs_gen2`, the migration that actually creates `workflow_run_nodes`). So 0072's `CREATE TABLE ... REFERENCES workflow_run_nodes(id)` really does run before the parent table exists — but that's valid SQLite (FK parent existence isn't checked at CREATE TABLE time, only at DML/PRAGMA foreign_keys=ON enforcement), and the codebase already relies on this exact ordering trick for `0070_workflow_run_cancel_status`. Runtime connections set `PRAGMA foreign_keys=ON` (`persistence/sqlite.rs`), so `ON DELETE CASCADE` is live. `specs/GENERATED/anyharness-db-schema.sql` matches the migration SQL. Minor: confirm this was produced by the documented regenerate command rather than hand-edited to match (content is correct either way).

**Finding — mark_leg_terminal / Cancel & Fence bypass the ledger (5):** `node_sessions::finished_leg_of` only recognizes `RecordLegThenHold`, `AdvanceToNext`, `CompleteRun`, `GateNode`, `FailNode`, `InterruptNode`. It does **not** match `Transition::Cancel` or `Transition::Fence`. Both of those transitions terminate a node's run status (cancelled / needs_attention+interrupted) via `store.rs`'s own match arms, but never call `mark_leg_terminal_tx` — so the ledger row for the node that was running when a run is cancelled (`WorkflowCommand::Cancel`) or fenced (boot-fence, every running row) is left at `status = 'running'` permanently. For a `Cancel`, there's no future resume path, so that ledger row is stuck at Running forever. For `Fence`'s interrupted case it self-heals only if the node later goes through `ResumeNode` (whose `upsert_leg_tx` on the next `stamp_session` resets the row) — but a fenced-and-abandoned run's ledger rows stay stuck too. Not observable today (nothing reads `node_legs` outside the pure fan-in decision, which only asks "any leg still Running" for the *current* node of a *live* run), but it directly matches the failure mode flagged in the task brief: 'a permanently-held node once rung 5 lands' — once N>1 legs and cross-node/held-node bookkeeping depend on this table being accurate, a run cancelled or fenced mid-flight will show a phantom outstanding leg forever. Recommend either extending `finished_leg_of` to cover `Cancel`/`Fence` (marking every disposed session's leg with the right terminal status) before rung 5, or an explicit tracked follow-up/TODO in `node_sessions.rs` calling this out — right now the module doc doesn't mention this gap.

**RecordLegThenHold / store arm (6):** Confirmed ledger-only — the `Transition::RecordLegThenHold { .. } => {}` match arm in the row-mutation switch is a genuine no-op (comment says so), and the actual ledger write happens in the separate `finished_leg_of` block after the match. Never reached with one leg per the pure-table analysis above; only exercised by the new synthetic multi-leg `RunState` in `transition_fanin_tests.rs`.

**Ratchets / store_rows.rs (7):** `store_rows.rs` relocation of `map_run`/`map_node`/`map_doc`/`parse_text` is verbatim (diffed byte-for-byte against the old inline versions). Ratchet shrinks 1377→1315 net (-62), consistent with the relocation outweighing the added ledger wiring; reason string is accurate.

**Test adequacy (8):** `transition_fanin_tests.rs` covers: first-leg-clean holds, last-leg-clean advances, a recorded failure only fails once all legs terminal, arrival-order independence, cancel-wins-at-last-leg, and interjection-hold-regardless-of-legs — a good spread of the F1 priority table using synthetic 2-leg `RunState`s (grammar can't produce these yet). `store_fanin_tests.rs` covers insert-on-launch, mark-done-in-completing-commit, and relaunch-resets-not-collides. Gap: no test exercises Cancel or Fence against a ledger row (would have caught finding 5).

No other correctness issues found in the one-leg reduction, session_extension.rs wiring, or the model.rs `WorkflowLegStatus`/`parse` roundtrip (Superseded correctly excluded).

pablonyx:
Review finding addressed in 87d1b2c2d: finished_leg_of now resolves Transition::Cancel to Cancelled and the store stamps it keyless, covering every leg of the node in the cancelling commit (one leg today, N under rung 5). Fence intentionally left unstamped: it parks the node and F6 resume resets the ledger, so no terminal state is misrecorded. Regression test cancel_marks_the_leg_cancelled_in_the_same_commit added; authored under the machine-wide build hold, so CI is the compiling verifier for this commit - local suite re-run happens when the hold lifts.

pablonyx:
DELTA RE-REVIEW (reviewer of record, fix head e0dd32f16): FINDING

Verified the approved fix as scoped: finished_leg_of gains a Transition::Cancel arm resolving to WorkflowLegStatus::Cancelled (node_sessions.rs), the store passes session = None for Cancel so mark_leg_terminal_tx runs its keyless branch (UPDATE ... WHERE node_row_id = ?1, no session filter) covering every leg row of that node in the same commit, and cancel_marks_the_leg_cancelled_in_the_same_commit asserts both status == Cancelled and completed_at.is_some() -- removing the Cancel arm from finished_leg_of would return None and skip the stamp entirely, so the test does fail without the fix.

However, the fix has a scope gap that reproduces the same bug class it was meant to close: Transition::Cancel { node_row_id, disposed_session_ids } only stamps the ledger leg for the single chain node_row_id, but disposed_session_ids is built by scanning every Running node -- chain and adhoc -- and disposing all their sessions (mirroring on_boot_fence's 'every running row' scan, per the transition.rs doc comment on Cancel). An adhoc row that is Running at cancel time gets its session disposed via DisposeSessions, but its own workflow_run_node_sessions leg (stamped via stamp_session -> upsert_leg_tx when it launched) is never touched -- it stays status = 'running' forever, which is exactly the permanently-held-leg shape ruling F1/this rung exists to prevent, just on an adhoc row instead of the chain node. Suggest either stamping the ledger leg for every id captured in disposed_session_ids's source node set (not just node_row_id), or a keyless-cancel variant of mark_leg_terminal_tx scoped to run_id rather than a single node_row_id.

One sentence on the Fence design note: accepted per F6 (crash-resume truncates and re-fans-out the ledger, so a parked Fence leg's stale status never survives to a user-visible read) -- no objection.

pablonyx:
DELTA RE-REVIEW ROUND 2 (reviewer of record, fix head 9abcf4bc9): CLEAN

Verified the round-1 finding is fixed correctly. `finished_leg_of` no longer has a Cancel arm (Cancel is documented as absent on purpose — run-terminal, not node-terminal). `store.rs` now special-cases `Transition::Cancel` before the `finished_leg_of` branch and calls `node_sessions::cancel_all_run_legs_tx(tx, run_id, &timestamp)`.

The SQL is correct: `UPDATE workflow_run_node_sessions SET status = 'cancelled', completed_at = ?3 WHERE status = 'running' AND node_row_id IN (SELECT id FROM workflow_run_nodes WHERE run_id = ?1)`.
- Catches the adhoc leg: the subselect scopes by `run_id`, not a single `node_row_id`, so every node row belonging to the run — chain and adhoc alike — is eligible.
- Avoids overwriting done legs: the outer `WHERE status = 'running'` filter means any leg already terminal (Done/Failed/Cancelled/etc.) is left untouched, matching the disposed-sessions scope (only `Running` nodes had sessions disposed in the first place).

The new regression test `cancel_marks_every_running_leg_in_the_run_cancelled` mints a chain node plus an adhoc row via `AddAdhocNode`, stamps a session on each (so both get ledger legs via `upsert_leg_tx`), cancels, and asserts both legs land `Cancelled` with `completed_at` set. Against the round-1 behavior (stamp only `node_row_id`, the current chain node) the adhoc leg would stay `running`, so this test does fail against the exact defect this round fixes.

F6 design note in the PR body matches what I accepted in the round-1 delta review: Fence legs are intentionally left non-terminal because crash-resume truncates and re-fans-out the ledger, so a parked leg's stale `running` status never survives to a user-visible read. No objection.

No cargo/rustc run (build hold respected), no edits made, reading only.

pablonyx:
Review chain closed at 9abcf4bc9 (delta re-review round 2 CLEAN by reviewer of record). Head moved to 6c1c9d8a2 by the rebase carry of the verified mechanical FE-PC-2 import fix on r3; r4's own content is byte-identical across the cascade. APPROVE-CARRIES per orchestrator, verified via the GitHub compare API.

pablonyx:
Head moved f436fc77e → 52bab2501 by the rebase carry of r3's mechanical FE-PC-4 test-variable rename; r4's own content is unchanged. Review chain state unchanged (closed at round-2 CLEAN, carried).

pablonyx:
Carry note (post-approval mechanical fix, head debf370d5 → 89d3f37cf): Repo shape's AH-STORE-3 ("SQL lives in the domain's store") flagged this rung's node_sessions.rs — the rule was masked earlier because repo-shape short-circuits at the first failing step, and the FE-PC fixes unmasked it. Fix is a pure relocation: node_sessions.rs → domains/workflows/store/node_sessions.rs, declared as a child module of store.rs; the two `use super::` imports become `use super::super::`, no logic changed (git shows it as a rename, 4 changed lines in the file). store.rs ratchet 1315 → 1318 for the module decl + note, reason appended per the growth rule. Locally: "AnyHarness boundary check passed.", "Max-lines check passed", boundary suite "Ran 94 tests ... OK". Cargo re-verification queued behind the machine's build-slot arbitration; CI's Cargo check & test also gates. Delta re-review available on request.

pablonyx:
**Delta re-review (round 3) — pinned to head 89d3f37cf9a54adb631768b117c1be33b4360d0a**

Verdict: **CLEAN**

Verified the single own-delta commit `89d3f37cf` ("relocate the fan-in ledger under store/ ... AH-STORE-3"):

1. `git show -M` confirms `node_sessions.rs` → `store/node_sessions.rs` as a 97%-similarity rename; the only content changes inside it are the two `use super::` → `use super::super::` import fixups required by the new nesting depth. No logic/SQL changes.
2. `mod.rs` loses `pub mod node_sessions;`; `store.rs` gains `mod node_sessions;` plus a 2-line comment and drops `use super::node_sessions;` — consistent with making it a private child module of `store.rs`.
3. `lints/anyharness/ratchets.toml`: `store.rs` max_lines 1315→1318 (+3), with an appended reason segment ("+3: R4 CI fix relocates node_sessions under store/ per AH-STORE-3..."); no other ratchet entry touched.
4. No `Cargo.lock` in the delta.
5. `git range-diff 35daba3b1..debf370d5 eb5033e70..89d3f37cf9a54adb631768b117c1be33b4360d0a` shows all three prior rung-4 commits carrying unchanged (`=`) across the rebase, with only the new relocate commit appended.

No findings.

pablonyx:
**Rebase receipt (r4):** `89d3f37cf` → `5fb3c0ce7`. One conflict + one ordered change: (1) migrations.rs registry — main gained 0072–0074 while this rung carried 0072_workflow_run_node_sessions; per the standing renumber order the migration is now **0076_workflow_run_node_sessions** (0075 stays reserved for #2033's workspace_checkpoints, serialized by merge order); SQL file renamed, registry entry moved after 0074 with a provenance comment, zero other references (grepped). Flagged for delta review as migration-identity change. Full-ladder rebase onto main `8532f52b2` (2026-08-19 ~02:2xZ, executing Pablo's 'continue implementations end to end' order). Chain: #2016 ← #2017 ← #2057 (replaces #2018) ← #2020 ← #2034 ← #2041. Validation: focused vitest 52/52 (definition-v2 34, draft-legs 5, auto-advance-toast 13), full product-client typecheck green (all sibling dists rebuilt). Rust compile = CI proves (a codex-lane cargo held the machine's one build slot; per this lane's precedent). Delta reviews of the rebase carries are being ordered where substantive.

pablonyx:
**Delta reviewer (migration renumber) — CONFIRMED CLEAN**

Scope: rebase-carry only (migration renumber 0072→0076) at r4 head `5fb3c0ce7a5ead70e0db257b22d4a510c1208510`. Rung-4 content itself already CLEAN through delta round 2; not re-litigated here.

**1. SQL byte-identity — CONFIRMED**
`anyharness/crates/anyharness-lib/src/persistence/sql/0076_workflow_run_node_sessions.sql` exists at the r4 head and is byte-for-byte identical (`diff` clean, 28/28 lines) to the pre-rebase `89d3f37cf:anyharness/crates/anyharness-lib/src/persistence/sql/0072_workflow_run_node_sessions.sql`. Only the CREATE TABLE/INDEX body — no migration-id string is embedded inside the SQL itself.

**2. Registry order — CONFIRMED**
`anyharness/crates/anyharness-lib/src/persistence/migrations.rs:262-263` registers `"0076_workflow_run_node_sessions"` as the last entry in `MIGRATIONS`, immediately after `"0074_opencode_message_ids"` (lines 252-253). Main's `0072_completion_wake_removal_intents` (244-245) and `0073_pending_prompt_cursor_backfill` (248-249) are intact and untouched. No `0075_*` entry exists in this branch (comment at 259-260 explains it's reserved for #2033's workspace_checkpoints by merge-order serialization) — confirmed no such file/entry present. No leftover `0072_workflow_run_node_sessions` entry anywhere in the file.

**3. No dangling references — CONFIRMED**
`git grep "0072_workflow_run_node_sessions" 5fb3c0ce7` across the whole tree at r4 head returns zero matches (exit 1). `git grep "0076_workflow_run_node_sessions" 5fb3c0ce7` returns exactly two matches, both in `migrations.rs` (the registry key string + the `include_str!` path) — nothing else references the new name.

**4. Migration-runner behavioral purity — CONFIRMED, with one honest caveat**
`run_named_migration` (persistence/migrations.rs:290-321) applies `MIGRATIONS` in registry order and records completion by literal name in `_migrations`, so for a fresh database the rename is a pure no-op relabeling — behavior is identical to pre-rebase.

Upgrade-path check: there's a precedent alias mechanism for exactly this rename scenario — `migration_aliases()` (line 425) already carries a `"0034_review_agent_loops" => &["0033_review_agent_loops"]` entry from an earlier main-branch number collision, letting a dev DB that applied the old name skip a harmless re-run. **No equivalent alias was added for `0072_workflow_run_node_sessions → 0076_workflow_run_node_sessions`.** So: any dev machine that ran this branch pre-rebase and has a `_migrations` row named `0072_workflow_run_node_sessions` will, on `0076_workflow_run_node_sessions`, take the "apply" path rather than the alias short-circuit. However the SQL itself (`0076_workflow_run_node_sessions.sql`, lines 16 and 27) uses `CREATE TABLE IF NOT EXISTS` and `CREATE INDEX IF NOT EXISTS` — so the re-run does **not** fail or error; it's a silent no-op DDL that then inserts a second `_migrations` row (`0076_workflow_run_node_sessions`) alongside the stale `0072_workflow_run_node_sessions` row. Net effect on an affected dev DB: no crash, no data loss, no schema drift — just one harmless orphaned bookkeeping row in `_migrations`. This is a real (if cosmetic) gap versus the established alias pattern, but it is dev-machine-only (this rung was never merged, only CI'd against fresh DBs) and does not block fresh-DB or CI correctness. Flagging for visibility rather than treating as blocking.

**5. Generated schema spec — CONFIRMED clean**
`specs/GENERATED/anyharness-db-schema.sql` at r4 head contains no migration-number references at all (it's organized by table name, not migration id) — grepped for `0072`, `0075`, `0076`: zero hits outside the table/index name matches for `workflow_run_node_sessions` itself (lines 679-689, 1011-1013), which match the migration's schema exactly. No stale references to either number.

**Verdict: rebase carry is CLEAN.** One non-blocking observation on item 4 (missing alias entry, functionally covered by IF NOT EXISTS) — not a defect requiring a fix, just noted for the record.

pablonyx:
**Cascade receipt (orchestrator).** Re-parented onto the recovered rung-1 head 9d90be8f0 (see #2016 note). New head: 77d30ab7fcede083073c660c53896d753f36b245 (was 5fb3c0ce7). Range-diff old→new: **all commits `=` (patch-identical)** — base-only move, no content change; per approve-carries precedent the standing review carries. Validated at the chain tip ff836e77c: workflows client suites `39 passed (39)` files / `431 passed (431)` tests, SDK `267 passed (267)`, product-client+SDK `tsc --noEmit` exit 0.

pablonyx:
**Merge-train cascade receipt (orchestrator).** Two-step re-parent after #2016 merged and the recovery lane advanced r2: first onto post-merge main (12518cb17), then onto the adopted r2 head b1c307b6bb (which adds a founder-authored ExistingWorkspace fix; my carried commits inside it are patch-identical). New head: 697cbe22149019c59da885197be8e8dc9635f980. All carried commits `=` in range-diff except one benign shrink: rung 5's commit no longer adds `model_validator` to the models_v2.py pydantic import because the adopted r2 fix already provides it (tree verified: 3 occurrences on the new head). No content change otherwise; standing reviews carry per the approve-carries precedent. CI riding the new heads is the compile/interaction verdict.

PR #2019, opened 2026-08-17
bgwork r7: native subagent creation identity treatment

pablonyx:
## What / why

Founder critique (2026-08-17): a native background subagent launch (Claude Task with `run_in_background`) rendered its transcript block as the generic `TranscriptAgentGroupBlock` header — *"Creating subagent probe child 2 · 1 tool call"* — instead of the identity treatment the design artifact (`Chat - Background Work Indicator.dc.html`) prescribes:

```html
<div class="flex min-h-8 min-w-0 flex-wrap items-center gap-x-1.5 gap-y-1 text-chat leading-8" data-subagent-creation-run>
  <AgentIdentityChip identity={...} on-open={showSubagent} hint-size="auto,28px" />
  <button type="button" class="relative top-px inline-block cursor-pointer align-middle text-foreground/90 hover:underline">started working</button>
</div>
```

## Before / after

- **Before:** whenever a native subagent's rawOutput carried a resolvable `subagentId` (rung R4's routing affordance), the header still showed a `Robot` glyph + `"Creating subagent" / "Subagent created"` + the task description + a `· N tool calls` summary — generic, no identity.
- **After:** once `subagentId` is resolvable, `TranscriptAgentGroupBlock` renders the same `data-subagent-creation-run` anatomy `SubagentCreationGroupBlock` already uses for delegated work: an `AgentIdentityChip` (glyph + generated name) built via `buildDelegatedAgentIdentity({ id: subagentId, title, workspaceId, sessionId: subagentId })`, plus a clickable verb (`"started working"` / `"starting"` while still running in the background / `"failed to start"` on launch failure). **Both** the chip and the verb call `onOpenSubagent(subagentId)` — the same `useOpenBackgroundWorkPane` machinery threaded through `MessageList` since R4. The chevron (expand/collapse the nested tool-call work) is unchanged and sits alongside. Because the identity is built from the same `subagentId` the pane roster row (`SubagentRosterRow`) and detail header (`BackgroundSubagentView`) already use (rung R4), all three surfaces now agree on name/color/glyph.
- When no `subagentId` is resolvable (no background-work correlation on this harness/item), the block falls back to today's byte-identical Robot-glyph expand/collapse header — unchanged.

## Scope guard

- Only `TranscriptAgentGroupBlock`'s native-creation header changed, plus threading a `workspaceId` prop down through `TranscriptToolCallGroupBlock` → `TranscriptAgentGroupBlock` (it was already available at that call site, just not forwarded).
[...]

PR #2013, opened 2026-08-17
feat(product): transcript-shaped claude subagent detail via client-side session-log translator (bgwork R4b)

pablonyx:
## Summary

Rung R4b of the Background Work slice (Delivery Spec — Background Work Slice 1, line 19; stack position 6, branched off `bgwork/r5-finish-signals` @ `4fdd503db4cfb1119ed7c88e2aee79fc3e1151bd`). Resolves the blocker R4's `BackgroundSubagentView` reported: the handoff's "Feed fidelity" note calls transcript-shaped rendering "honest for Claude today" because the child transcript "arrives as a JSONL tail over `tail_file`" — but no translator from that native Claude-CLI JSONL to the SDK's `TranscriptState`/`SessionEventEnvelope` shape existed anywhere in the repo. This PR adds one, entirely client-side, and wires it into the view. **Zero runtime/fork changes** — everything here is `apps/packages/product-client`.

- `domain/activity/claude-session-log.ts` (+ `claude-session-log-envelopes.ts`, split out purely for FE-SIZE-1): a pure, stateless, per-call recompute that maps the accumulated `tail_file` buffer into a `TranscriptState`. It never throws: malformed JSON lines are skipped and counted, unsupported top-level line types (`system`, `summary`, `queue-operation`, …) are skipped and counted, `isSidechain` lines (a nested Task-tool subagent *inside* this child — out of this slice's scope) are skipped and counted, the buffer's trailing line is always held back (`tail_file` has no line-complete guarantee), and an orphaned `tool_result` (its `tool_use_id` outside the buffer, e.g. a capped/truncated tail) is dropped and counted rather than crashing.
  - Reuses `@anyharness/sdk`'s own reducer (`reduceEvents`) instead of hand-building `TranscriptState` — the same convergence point `domain/chats/transcript/envelope-to-state.ts` already uses for the main transcript. This means turn bookkeeping, and tool semantic-kind classification (`Bash` → terminal, `Task`/`Agent` → subagent, `Grep`/`Glob`/`Ls` → search, via the reducer's existing `nativeToolName` heuristics) come for free and stay identical to a live session's rendering. `Read`/`Edit`/`Write` fall back to the generic "other" tool-call treatment (no file-diff-preview content-part synthesis) — a deliberate scope limit, not a bug.
- `hooks/activity/derived/use-claude-session-log-transcript.ts`: a one-line `useMemo` wrapper next to `use-feed-stream.ts`, per the reviewer's sketch.
[...]

PR #2021, opened 2026-08-17
bgwork r6: inline completion receipts + running-only indicator

pablonyx:
## What

Two founder-ruled changes (2026-08-17) in `apps/packages/product-client`, amending the frozen handoff:

**1. Inline completion receipts in the transcript.** When a background terminal exits or a native subagent finishes, a right-aligned receipt row appears at the transcript tail, on the incoming rail, per the authoritative design markup ("Chat - Background Work Indicator"):
- Terminal: `exited {code} ·` + a mono command button that deep-opens the Background work pane's `BackgroundTerminalView` for that process.
- Subagent: `finished ·` (or `failed ·`) + the durable `AgentIdentityChip` (identity via `buildDelegatedAgentIdentity`, keyed by the subagent's wire agent id), whose on-open deep-opens the pane's subagent detail.

**2. Tail indicator is running-only.** `BackgroundWorkTranscriptRow` now counts RUNNING work only and is not rendered at count 0. The settled/finished display state ("N background tasks finished" + `CircleCheck` swap, `text-faint` tone) is removed entirely, including its tests.

## Founder ruling (quoted)

> the `{n} background task{s}` row counts RUNNING work only, and the row is NOT RENDERED at count 0. Remove the settled/finished display state ... entirely ... Completed work is announced solely by the new inline receipts.
> completed tasks leave the count; the row disappears at 0.

## Architecture path: **(b) — synthesized from the activity fold**

The wire carries **no** identifying provenance for native BACKGROUND completions, so the preferred path (a) — rendering receipts from a wake-prompt transcript envelope — is not available for this feature. Evidence:

- **Terminals inject no session wake prompt at all.** The terminals domain only writes shell-prompt bytes, never a `PromptProvenance` (`anyharness/crates/anyharness-lib/src/live/terminals/command_runs/*` — `stream_format.rs` `workspace_prompt`/`terminal_command_preface`; no `with_provenance`/wake anywhere under `command_runs`). `PromptProvenance` (`.../sessions/prompt/provenance.rs:23-66`, mirrored in the SDK at `anyharness/sdk/src/generated/openapi.ts:4139-4169`) has **no** process/terminal variant — only `agentSession | subagentWake | linkWake | reviewFeedback | system`. A terminal exit therefore has nothing on the wire to reconstruct "exited 0 · pytest -q tests/e2e" or to deep-link the process.
- **Native roster subagents leave no durable transcript record.** They leave the roster the instant they finish; the only evidence is a roster departure observed between two snapshots — see `use-background-work-finish-signal-tracking.ts:60-99` (R5) and its store field docstring (`workspace-ui-store-types.ts`: "the only record ... that will ever exist"). If a `subagentWake` transcript wake existed for these, R5 would not need to cache the disappearance.
- The `subagentWake`/`linkWake` provenance that `AgentOriginPromptReceipt` already renders is for delegated/linked **sessions** (cowork/review/agent-parent) via completion deliveries (`.../sessions/subagents/transcript.rs:71`, `.../store/completion_deliveries/*`) — a different mechanism, not "background work."

So receipts are synthesized from fold transitions (a running process now exited; a running subagent now finished-in-place or vanished), pure-derived in `domain/activity/background-completion-receipt.ts` and accumulated per-session in `hooks/activity/derived/use-background-completion-receipts.ts` (mirroring R5's transition-detection shape and reading the same `useSessionActivityForSession` slice).

## Honest limitations (disclosed)

- **Synthesized receipts do not survive reload / are not interleaved at the exact historical scroll position** (known lane defect D1: activity rendering is fold-only). They cover completions observed while the hook is mounted; work already finished at mount (roster seed) is baselined, not receipted. Placement is now a true in-scroll virtualized row (see **Round 2** below, superseding the r6 tail band). The durable, correctly-ordered wire signal is flagged for the wire rung.
- **CONTRADICTION RAISED — subagent receipt "opens the pane's subagent detail":** a finished native subagent has, by the locked roster-departure invariant, already left the live roster, and `BackgroundWorkPane` resolves `selectedSubagent` from `activity.agents` — so it bounces back to the roster (documented at `BackgroundWorkPane.tsx`, the "subagent-View-action" note). R4b renders a *live* subagent's tail as a transcript; it does **not** reconstruct a roster-absent one. The receipt chip's on-open still deep-opens the pane (best-effort: it resolves during the one-tick finished-in-place window, else shows the roster), but a durable finished-subagent detail is blocked on the same wire/durability work as D1. Raising rather than silently rescoping.

## Deep-open plumbing

Added an analogous terminal selection to the R4 subagent one: `pendingBackgroundProcessSelectionByWorkspace` + `set/clearPendingBackgroundProcessSelectionForWorkspace` (`workspace-ui-store*`), consumed+cleared by `BackgroundWorkPane` with the same session-checked discipline, driven by a new `useOpenBackgroundTerminalDetail` hook (shares the pane-open helper with `useOpenBackgroundWorkPane`; the existing subagent signature is unchanged).

## Tests (literal `pnpm exec vitest run`)

Consolidated run of every touched/owned spec + the full background-work suite list: **18 files, 153 passed, 0 failed.** Includes: receipt rendering both kinds + click-through wiring + identity-less fallback (`BackgroundCompletionReceipt.test.tsx`), pure transition derivation + roster-seed negative control + no-double-emit (`background-completion-receipt.test.ts`), accumulation + session-switch reset (`use-background-completion-receipts.test.ts`), terminal deep-open (`use-open-background-work-pane.test.tsx`), indicator hide-at-0 + **negative control that the removed settled-state copy/CircleCheck/`text-faint` are gone** (`BackgroundWorkTranscriptRow.test.tsx`).

`tsc -p tsconfig.json --noEmit`: only the pre-existing exogenous `TS2741` in `src/domain/workflows/run-view-model.ts` (from main; not touched, does not affect vitest). Design-attribution gate: passed. New domain module 147 lines (< 404); all new files < 600.

---

## Round 2 — founder amendments (2026-08-17)

The founder live-tested the r6 build and ruled three amendments, committed **on top** of the r6 work (no history rewrite).

### FIX 1 — running-count row is a true in-scroll transcript row (supersedes R1's dock placement)

> the pinned-to-bottom footer "reads as **fading in weirdly**" — make it a true in-scroll transcript row that sits statically **below** the agent's last message in content flow and scrolls **with** content.

`BackgroundWorkTranscriptRow` is now a virtualized `background_work` row emitted by `buildTranscriptRowModel` at the absolute transcript tail while `runningCount > 0`, gone at 0. This **supersedes R1's dock-anchored, visibility-gated placement**. Removed as now-dead: the pinned-to-bottom `isPinnedToBottom` gating + reset effect, the floating absolute band JSX + geometry comment, and the **R2 MessageList bottom-inset reserve machinery** (`hasBackgroundWork` inset threading through `SessionTranscriptPane` to `MessageList`); `bottomInsetPx` is now passed plainly. **Kept:** the `MessageList` bottom-inset prop itself (still used for the unrelated composer-dock reserve).

### FIX 2 — receipts interleave in reading order at the fold-observation point

> receipts render in reading order, interleaved into the transcript item sequence — agent turn → receipt → wake turn; the "exited N · command" receipt appears in the (right-aligned) message slot, **not** as an element after/below the message list.

Each receipt now carries an `anchorTurnId` (the turn that was latest when the completion was folded, captured via ref at observation time so a later wake turn never restamps it). `buildTranscriptRowModel` buckets receipts by that anchor and splices a `completion_receipt` row at the **end** of its anchor turn, so a later (wake) turn falls after it. Receipts are real in-scroll transcript rows rendered inside `MessageList`, not appended after it. Modeled on the existing goal-event interleaving precedent (turn-END anchoring rather than mid-turn seq).

### FIX 3 — remove the in-pane NoticeBanner

> we don't want the **checkmark + box at the top of background work**.

Removed the `NoticeBanner` (CircleCheck + View box) from the top of `BackgroundWorkPane` and its dead wiring (`showBackgroundWorkNotice`/`noticeSignal`, the `lastViewedAtMsBeforeThisMount` state). **Kept** R5's tab dirty-dot: `useBackgroundWorkFinishSignal` + the mark-viewed effect still fire. A negative control asserts the banner no longer renders on process/subagent finish.

### Size (PROD-SIZE-1, no ratchet raise)

Split along real seams rather than ratcheting up: goal-event + completion-receipt row interleaving helpers extracted to **`transcript-interleave-rows.ts`** (row-model 1143 to 856 lines; its ratchet **tightened** 1017 to 856), and the background-work row tests moved to **`transcript-row-model.background-work.test.ts`** (main test back to 1363). Full `check_max_lines.py` and `check_design_attribution.py`: **pass**.

### Round-2 tests (literal `vitest run`)

Full touched-surface re-run (`src/domain/chats/transcript src/components/workspace/chat src/hooks/chat/ui src/components/workspace/activity src/hooks/activity src/domain/activity`): **155 files, 1158 passed, 0 failed.** `tsc -p tsconfig.json --noEmit`: only the same pre-existing exogenous `TS2741` in `src/domain/workflows/run-view-model.ts` (from main; not touched).

Round-2 head: `d66ed0579`.

## Round 3 — wake-turn copy affordance regression

Founder repro (served preview, run `bn30h453a`): after a background command exits, the FINAL agent message — the runtime-injected **wake turn** ("Terminal … finished — exit code 0") — showed no copy button and no timestamp on hover; the footer `<div data-turn-assistant-footer>` mounted but was empty. Durable across a fresh reload.

### Actual root cause (not the interleaving)

The interleaved `completion_receipt` / `background_work` rows were a red herring — the preceding turns kept their footers; only the wake turn regressed. `completedAt` is written **only** by the `turn_ended` event (sdk `reducer/transcript.ts`), and a runtime-injected wake turn never receives one, so its `TurnRecord.completedAt` stays `null` forever. The footer gate `resolveTurnAssistantFooterMode` required `turnCompleted = !!turn.completedAt`, so the fully-settled final message fell through to the empty **"reserved"** mode. On reload the assistant-reveal mount floor makes `assistantRevealComplete` true, so `completedAt` was the sole false input — matching the durable, both-affordances-gone symptom exactly.

### Fix

New `resolveTurnAssistantFooterModeForRow` (in `transcript-rendering.ts`) sources "is this turn done?" from the tail assistant **message**, not only the turn envelope: a turn is footer-complete when `!!turn.completedAt` **OR** its tail prose item has stopped streaming and no tool work is still active (`turnHasActiveToolWork`). Affordances belong to the last real message row regardless of interleaved activity rows. `TranscriptTurnRow` calls it in place of the raw `!!turn.completedAt` gate. Frontier/diff/goal-marker chrome (which keys on the completion stamp via `latestCompletedTurnId`) is deliberately untouched — the wake turn gets its hover copy + timestamp back without being promoted to the "session complete" persistent marker.

### Tests (literal `vitest run`)

`transcript-rendering.test.ts` grew 9 → **17** tests: the wake-turn copy restore, plus negative controls (streaming prose, prose-closed-while-tool-runs, reveal-incomplete, stamped turn unchanged, non-final row → none), plus (a) receipt interleaved after the preceding turn, (b) `background_work` row at the tail, (c) both — each asserting the wake turn row still resolves to `copy`. Suite run: `transcript-rendering.test.ts` **17**, `transcript-row-model.test.ts` **43**, `transcript-row-model.background-work.test.ts` **8** — 68 passed, 0 failed. Negative control: reverting the `|| tailMessageSettled` branch fails exactly the 4 fix-dependent tests and leaves every control green. `check_max_lines.py`, `check_design_attribution.py`, `check_frontend_boundaries.py`: **pass** (touched files 309/365/367 lines, under caps, no ratchet entries). Same pre-existing exogenous `TS2741` in `run-view-model.ts` (from main; not touched).

Round-3 head: `4f8f00a51`.

## Round 4 — the runtime drops the wake turn's entire completion tail

Round 3 shipped, but the wake-turn footer was STILL empty on the live stack. Pulling the durable event stream (`GET /v1/sessions/{id}/events`) showed why: for a runtime-injected wake turn the runtime emits `turn_started` → `item_started(assistant_message, in_progress)` → `item_delta("… completed with exit code 0.")` → terminal `usage_update`(s), then the stream ends. There is **no `item_completed` and no `turn_ended`**. So in the durable record the tail assistant message item stays status `in_progress` (isStreaming true) forever — round 3's tail-item-settled gate is false, and so is the `completedAt` stamp. Both round-3 branches lose on real data.

### Fix — source completion from session liveness

`resolveTurnAssistantFooterModeForRow` no longer reads the tail item's status at all. A turn with copyable tail prose is footer-complete when it is **not the session's actively-streaming turn** — `turnIsActivelyStreaming = isLatestTurn && sessionViewState === "working"`, negated. `sessionViewState` is the runtime-derived run state (`resolveSessionViewState`): on reload there is no live stream and `isSessionEffectivelyStreaming` returns false for an idle+ended session, so the stuck-`in_progress` wake message becomes copyable even though the reducer's replayed `transcript.isStreaming` is stale-true. Interrupted turns match too. Genuinely live turns keep `turnIsActivelyStreaming` true and stay `reserved`, with the reveal guard holding the empty footer through the reveal.

Two OR'd signals remain: `!!turn.completedAt` (normal stamped turns, unchanged) OR `hasTailProse && !turnIsActivelyStreaming`.

### Reviewer findings folded in

- **L1** (mid-turn `copy` flash in the gap between a prose block finishing and the next `tool_call`): closed by the same liveness gate — that gap is still the live turn, so it stays `reserved`. New test asserts it.
- **L2** (one-frame stale-receipt flash at the tail on session switch; receipts reset in an async effect): display-only, left as **known-cosmetic** to keep this diff focused. The durable fix for the missing completion tail is **runtime-side** (emit `item_completed` + `turn_ended` for injected wake turns) — recorded here as a **known runtime defect** next to the provenance wire-rung note. No Rust touched.

### Tests (verbatim `vitest run`)

```
 ✓ src/domain/chats/transcript/transcript-rendering.test.ts (19 tests) 6ms
 ✓ src/domain/chats/transcript/transcript-row-model.test.ts (43 tests) 13ms
 ✓ src/domain/chats/transcript/transcript-row-model.background-work.test.ts (8 tests)
 ✓ src/components/workspace/chat/transcript/TranscriptTurnRow.test.tsx (16 tests) 3ms
 ✓ src/components/workspace/chat/transcript/TranscriptTurnChrome.test.tsx (23 tests) 43ms
 Test Files  5 passed (5)
      Tests  109 passed (109)
```

Negative controls: reverting `|| messageSettledOffLiveStream` fails exactly the round-4 dropped-tail test + interleaving (a)/(b)/(c); dropping the `!turnIsActivelyStreaming` gate fails exactly the three live-turn controls (round-4 active control, tool-running, L1). `check_max_lines.py`, `check_design_attribution.py`, `check_frontend_boundaries.py`: **pass** (touched files 322/419/375 lines, under caps, no ratchet entries). Same pre-existing exogenous `TS2741` in `run-view-model.ts` (from main; not touched).

Round-4 head: `1fb98411e`.

## Round 4b — FE-SIZE-1 ratchet for MessageList.tsx

CI's "Repo shape checks" job (`report_frontend_structure.py --strict --summary-only`, now added to the standard local gate set) flagged `MessageList.tsx` at 442 lines, over the 400-line FE-SIZE-1 soft threshold — grown by r6's in-scroll row work. Recorded the sanctioned `[[frontend_structure]]` ratchet entry (observed 442 + reason) in `lints/frontend/ratchets.toml`, in the same PR that grew it, rather than splitting a host that just cleared four review rounds (the alternative, and not the founder-gated `exceptions.toml` path).

```
  FE-SIZE-1: 0
  TOTAL: 0
```

`check_max_lines.py` still passes (the file is in `[[frontend_structure]]` only, not `[[max_lines]]` — no collision). The second repo-wide FE-SIZE-1 hit, `TranscriptToolCallItemBlock.tsx` at 402, is r8/#2022's and does not appear on this branch.

Round-4b head: `79127061f`.

## Round 5 — frontier reserve leaves a dead band under settled wake turns

Pablo's final pre-merge blocker, same stream-liveness family as round 4. After a background-command wake message settles and the session goes idle, the final wake turn showed a dead ~44-56px band between the last prose and the timestamp footer. `resolveTurnFrontierStatusMode` reserved an empty fixed-height frontier-status box whenever `rowIsLastTurnRow && isLatestTurnInProgress`; because the runtime drops the wake turn's completion tail (defect D3), `isLatestTurnInProgress` (`isLatestTurn && !turn.completedAt`) stays true forever in the live view, so the `h-6` box plus its two 16px column gaps rendered as the gap. On reload the durable store has the tail, `completedAt` is set, and the gap vanishes — hence live-only.

### Fix — same doctrine as the round-4 footer

The reserve now also requires genuine stream liveness: `resolveTurnFrontierStatusMode` takes `turnIsActivelyStreaming` (`isLatestTurn && sessionViewState === "working"`, already computed in `TranscriptTurnRow`) and the reserve branch is `rowIsLastTurnRow && isLatestTurnInProgress && turnIsActivelyStreaming`. That flips exactly one truth-table cell — an idle session with a dropped-tail latest turn now resolves `"hidden"` instead of `"reserved"`. Genuinely streaming turns keep the anti-jitter reserve unchanged.

**needs_input decision:** a real trailing status still wins `"status"` before the reserve branch, so a permission pause that carries its interaction indicator (the normal case) is unaffected. A `needs_input` turn with a null trailing status resolves `"hidden"` rather than a dead reserve — coherent with the round-4 footer, which already treats the same non-`"working"` turn as settled/copyable (showing copy), so the frontier must not simultaneously claim the turn is still working.

### Tests (verbatim `vitest run`)

```
 ✓ src/components/workspace/chat/transcript/TranscriptTurnRow.test.tsx (19 tests)
```
(+3 frontier cases: dropped-tail wake + idle → `hidden`; genuinely streaming/working → `reserved`; trailing status present → `status` regardless.) Sibling regression run — `TranscriptTurnChrome.test.tsx` (23), `transcript-rendering.test.ts` (19), `transcript-row-model.test.ts` (43), `transcript-row-model.background-work.test.ts` (8) — **93 passed**. Negative control: reverting the `&& turnIsActivelyStreaming` gate fails exactly the R5 test and leaves the streaming/trailing-status controls green.

Gates (from the scratch worktree): `check_max_lines.py`, `check_design_attribution.py`, `check_frontend_boundaries.py`, `report_frontend_structure.py --strict --summary-only` (**FE-SIZE-1: 0, TOTAL: 0**) all pass. (`report_frontend_structure` added to the standard local gate set.)

New head: `209c23663`.

### DIFF
--- a/apps/packages/product-client/src/components/workspace/activity/BackgroundCompletionReceipt.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/activity/BackgroundCompletionReceipt.test.tsx
@@ -1,0 +1,126 @@
+/* @vitest-environment jsdom */
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { BackgroundCompletionReceipt } from "./BackgroundCompletionReceipt";
+import type {
+  SubagentCompletionReceipt,
+  TerminalCompletionReceipt,
+} from "#product/domain/activity/background-completion-receipt";
+
+afterEach(() => {
+  cleanup();
+});
+
+function terminalReceipt(
+  overrides: Partial<TerminalCompletionReceipt> = {},
+): TerminalCompletionReceipt {
+  return {
+    kind: "terminal",
+    key: "terminal:proc-1",
+    processId: "proc-1",
+    command: "pytest -q tests/e2e",
+    exitCode: 0,
+    atMs: 1_000,
+    anchorTurnId: null,
+    ...overrides,
+  };
+}
+
+function subagentReceipt(
+  overrides: Partial<SubagentCompletionReceipt> = {},
+): SubagentCompletionReceipt {
+  return {
+    kind: "subagent",
+    key: "subagent:task-1",
+    subagentId: "task-1",
+    title: "audit the roster",
+    outcome: "completed",
+    atMs: 1_000,
+    anchorTurnId: null,
+    ...overrides,
+  };
+}
+
+describe("BackgroundCompletionReceipt — terminal", () => {
+  it("renders the design markup: exited-code verb + mono command button on the incoming rail", () => {
+    const { container } = render(
+      <BackgroundCompletionReceipt receipt={terminalReceipt()} workspaceId="ws-1" onOpen={() => {}} />,
+    );
+    const rail = container.querySelector('[data-background-completion-receipt="terminal"]');
+    expect(rail).toBeTruthy();
+    const receipt = container.querySelector('[data-agent-message-receipt][data-direction="incoming"]');
+    expect(receipt).toBeTruthy();
+    expect(screen.getByText("exited 0 ·")).toBeTruthy();
+    const button = screen.getByRole("button", { name: "Open terminal pytest -q tests/e2e" });
+    expect(button.className).toContain("font-mono");
+    expect(button.textContent).toBe("pytest -q tests/e2e");
+  });
+
+  it("shows a nonzero exit code in the verb", () => {
+    render(
+      <BackgroundCompletionReceipt
+        receipt={terminalReceipt({ exitCode: 130 })}
+        workspaceId="ws-1"
+        onOpen={() => {}}
+      />,
+    );
+    expect(screen.getByText("exited 130 ·")).toBeTruthy();
+  });
+
+  it("fires onOpen when the command button is clicked (terminal deep-open)", () => {
+    const onOpen = vi.fn();
+    render(
+      <BackgroundCompletionReceipt receipt={terminalReceipt()} workspaceId="ws-1" onOpen={onOpen} />,
+    );
+    fireEvent.click(screen.getByRole("button", { name: "Open terminal pytest -q tests/e2e" }));
+    expect(onOpen).toHaveBeenCalledTimes(1);
+  });
+});
+
+describe("BackgroundCompletionReceipt — subagent", () => {
+  it("renders 'finished ·' + an identity chip keyed by the wire agent id", () => {
+    const { container } = render(
+      <BackgroundCompletionReceipt receipt={subagentReceipt()} workspaceId="ws-1" onOpen={() => {}} />,
+    );
+    expect(container.querySelector('[data-background-completion-receipt="subagent"]')).toBeTruthy();
+    expect(screen.getByText("finished ·")).toBeTruthy();
+    expect(container.querySelector("[data-agent-identity-chip]")).toBeTruthy();
+    expect(screen.getByText("audit the roster")).toBeTruthy();
+  });
+
+  it("renders 'failed ·' for a failed subagent", () => {
+    render(
+      <BackgroundCompletionReceipt
+        receipt={subagentReceipt({ outcome: "failed" })}
+        workspaceId="ws-1"
+        onOpen={() => {}}
+      />,
+    );
+    expect(screen.getByText("failed ·")).toBeTruthy();
+  });
+
+  it("fires onOpen when the identity chip is clicked (subagent deep-open)", () => {
+    const onOpen = vi.fn();
+    render(
+      <BackgroundCompletionReceipt receipt={subagentReceipt()} workspaceId="ws-1" onOpen={onOpen} />,
+    );
+    fireEvent.click(screen.getByRole("button", { name: /Open/ }));
+    expect(onOpen).toHaveBeenCalledTimes(1);
+  });
+
+  // Identity-less fallback: a receipt whose subagent id cannot mint a durable
+  // identity (no session id) degrades to the plain title-fallback span, never a
+  // broken chip button.
+  it("falls back to a plain title span when no durable identity can be minted", () => {
+    const { container } = render(
+      <BackgroundCompletionReceipt
+        receipt={subagentReceipt({ subagentId: "" })}
+        workspaceId="ws-1"
+        onOpen={() => {}}
+      />,
+    );
+    expect(container.querySelector("[data-agent-identity-chip]")).toBeNull();
+    expect(container.querySelector("[data-agent-message-fallback]")).toBeTruthy();
+    expect(screen.getByText("audit the roster")).toBeTruthy();
+  });
+});
--- a/apps/packages/product-client/src/components/workspace/activity/BackgroundCompletionReceipt.tsx
+++ b/apps/packages/product-client/src/components/workspace/activity/BackgroundCompletionReceipt.tsx
@@ -1,0 +1,86 @@
+import { AgentMessageReceipt } from "#product/components/workspace/chat/transcript/AgentMessageReceipt";
+import { Button } from "#product/primitives/Button";
+import { buildDelegatedAgentIdentity } from "#product/lib/domain/delegated-work/identity";
+import {
+  subagentReceiptVerb,
+  terminalReceiptVerb,
+  type BackgroundCompletionReceipt as BackgroundCompletionReceiptModel,
+} from "#product/domain/activity/background-completion-receipt";
+
+/**
+ * One inline background-work completion receipt in the transcript tail (bgwork
+ * r6). Right-aligned on the incoming rail, per the design artifact ("Chat -
+ * Background Work Indicator").
+ *
+ * - Terminal: `exited {code} ·` + a mono command button that deep-opens the
+ *   Background work pane's `BackgroundTerminalView` for that process.
+ * - Subagent: `finished ·` (or `failed ·`) + the durable `AgentIdentityChip`
+ *   (reused via `AgentMessageReceipt`), keyed by the subagent's wire agent id.
+ *
+ * Both share `AgentMessageReceipt`'s markup (`data-agent-message-receipt`,
+ * incoming direction) so the receipts read identically to the transcript's
+ * existing agent receipts; the terminal variant swaps the chip for a command
+ * button since a background terminal has no agent identity.
+ */
+export function BackgroundCompletionReceipt({
+  receipt,
+  workspaceId,
+  onOpen,
+}: {
+  receipt: BackgroundCompletionReceiptModel;
+  workspaceId: string | null;
+  onOpen: () => void;
+}) {
+  if (receipt.kind === "subagent") {
+    const identity = buildDelegatedAgentIdentity({
+      id: receipt.subagentId,
+      title: receipt.title,
+      workspaceId,
+      sessionId: receipt.subagentId,
+    });
+    return (
+      <div
+        className="flex justify-end"
+        data-agent-origin-prompt
+        data-background-completion-receipt="subagent"
+      >
+        <AgentMessageReceipt
+          direction="incoming"
+          identity={identity}
+          fallbackLabel={receipt.title}
+          verb={subagentReceiptVerb(receipt.outcome)}
+          onOpen={onOpen}
+        />
+      </div>
+    );
+  }
+
+  return (
+    <div
+      className="flex justify-end"
+      data-agent-origin-prompt
+      data-background-completion-receipt="terminal"
+    >
+      <div
+        data-agent-message-receipt
+        data-direction="incoming"
+        className="flex max-w-[85%] items-center self-end text-right text-chat leading-8"
+      >
+        <span className="relative top-px align-middle text-muted-foreground">
+          {terminalReceiptVerb(receipt.exitCode)}
+        </span>
+        <Button
+          type="button"
+          variant="unstyled"
+          size="unstyled"
+          data-chat-transcript-ignore
+          className="ms-1.5 inline-block min-w-0 max-w-72 shrink cursor-pointer truncate font-mono font-medium text-foreground/80 hover:underline"
+          aria-label={`Open terminal ${receipt.command}`}
+          onClick={onOpen}
+        >
+          {receipt.command}
+        </Button>
+      </div>
+    </div>
+  );
+}
--- a/apps/packages/product-client/src/components/workspace/activity/BackgroundWorkTranscriptRow.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/activity/BackgroundWorkTranscriptRow.test.tsx
@@ -11,54 +11,38 @@ const PROLIFERATE_MARK_SELECTOR = 'svg[viewBox="300 300 200 200"]';
 const CIRCLE_CHECK_SELECTOR = 'svg[viewBox="0 0 24 24"]';
 
 describe("BackgroundWorkTranscriptRow", () => {
-  it("renders nothing when nothing is running and nothing finished", () => {
+  // Founder ruling (bgwork r6): the row counts RUNNING work only and is not
+  // rendered at count 0. Completed tasks leave the count; the row disappears
+  // at 0. There is no settled/finished display state — completed work is
+  // announced solely by the inline `BackgroundCompletionReceipt` rows.
+  it("renders nothing when nothing is running", () => {
     const { container } = render(
-      <BackgroundWorkTranscriptRow runningCount={0} finishedCount={0} onOpen={() => {}} />,
+      <BackgroundWorkTranscriptRow runningCount={0} onOpen={() => {}} />,
     );
     expect(container.firstChild).toBeNull();
   });
 
   it("renders the singular running copy", () => {
-    render(<BackgroundWorkTranscriptRow runningCount={1} finishedCount={0} onOpen={() => {}} />);
+    render(<BackgroundWorkTranscriptRow runningCount={1} onOpen={() => {}} />);
     expect(screen.getByText("1 background task")).toBeTruthy();
   });
 
   it("renders the plural running copy", () => {
-    render(<BackgroundWorkTranscriptRow runningCount={3} finishedCount={0} onOpen={() => {}} />);
+    render(<BackgroundWorkTranscriptRow runningCount={3} onOpen={() => {}} />);
     expect(screen.getByText("3 background tasks")).toBeTruthy();
   });
 
-  it("renders the static Proliferate mark while running, not CircleCheck", () => {
+  it("renders the static Proliferate mark while running, never CircleCheck", () => {
     const { container } = render(
-      <BackgroundWorkTranscriptRow runningCount={2} finishedCount={0} onOpen={() => {}} />,
+      <BackgroundWorkTranscriptRow runningCount={2} onOpen={() => {}} />,
     );
     expect(container.querySelector(PROLIFERATE_MARK_SELECTOR)).toBeTruthy();
     expect(container.querySelector(CIRCLE_CHECK_SELECTOR)).toBeNull();
   });
 
-  it("renders the settled copy, pluralized, once the running count is zero", () => {
-    render(<BackgroundWorkTranscriptRow runningCount={0} finishedCount={2} onOpen={() => {}} />);
-    expect(screen.getByText("2 background tasks finished")).toBeTruthy();
-  });
-
-  it("renders the singular settled copy", () => {
-    render(<BackgroundWorkTranscriptRow runningCount={0} finishedCount={1} onOpen={() => {}} />);
-    expect(screen.getByText("1 background task finished")).toBeTruthy();
-  });
-
-  it("swaps the glyph to CircleCheck once settled, and rests at text-faint", () => {
+  it("rests at text-muted-foreground and shares one hover treatment between glyph and label", () => {
     const { container } = render(
-      <BackgroundWorkTranscriptRow runningCount={0} finishedCount={1} onOpen={() => {}} />,
-    );
-    expect(container.querySelector(CIRCLE_CHECK_SELECTOR)).toBeTruthy();
-    expect(container.querySelector(PROLIFERATE_MARK_SELECTOR)).toBeNull();
-    const label = screen.getByText("1 background task finished");
-    expect(label.className.split(" ")).toContain("text-faint");
-  });
-
-  it("shares one hover treatment between the glyph wrapper and the label", () => {
-    const { container } = render(
-      <BackgroundWorkTranscriptRow runningCount={1} finishedCount={0} onOpen={() => {}} />,
+      <BackgroundWorkTranscriptRow runningCount={1} onOpen={() => {}} />,
     );
     const label = screen.getByText("1 background task");
     const glyphWrapper = container.querySelector(PROLIFERATE_MARK_SELECTOR)?.parentElement;
@@ -71,24 +55,36 @@ describe("BackgroundWorkTranscriptRow", () => {
     expect(label.className.split(" ")).toContain("text-muted-foreground");
   });
 
-  it("carries no motion class in the running state", () => {
+  it("carries no motion class", () => {
     const { container } = render(
-      <BackgroundWorkTranscriptRow runningCount={1} finishedCount={0} onOpen={() => {}} />,
-    );
-    expect(container.innerHTML).not.toMatch(/animate-|motion-safe:/);
-  });
-
-  it("carries no motion class in the settled state", () => {
-    const { container } = render(
-      <BackgroundWorkTranscriptRow runningCount={0} finishedCount={1} onOpen={() => {}} />,
+      <BackgroundWorkTranscriptRow runningCount={1} onOpen={() => {}} />,
     );
     expect(container.innerHTML).not.toMatch(/animate-|motion-safe:/);
   });
 
   it("fires onOpen when clicked", () => {
     const onOpen = vi.fn();
-    render(<BackgroundWorkTranscriptRow runningCount={1} finishedCount={0} onOpen={onOpen} />);
+    render(<BackgroundWorkTranscriptRow runningCount={1} onOpen={onOpen} />);
     fireEvent.click(screen.getByRole("button", { name: "1 background task" }));
     expect(onOpen).toHaveBeenCalledTimes(1);
+  });
+
+  // NEGATIVE CONTROL: the removed settled state is gone in every arrangement —
+  // no "finished" copy, no CircleCheck glyph, and no `text-faint` settled tone,
+  // whether running is positive or zero.
+  it("never renders the removed settled state (no finished copy, no CircleCheck, no text-faint)", () => {
+    const running = render(
+      <BackgroundWorkTranscriptRow runningCount={2} onOpen={() => {}} />,
+    );
+    expect(running.container.textContent).not.toMatch(/finished/i);
+    expect(running.container.querySelector(CIRCLE_CHECK_SELECTOR)).toBeNull();
+    expect(running.container.innerHTML).not.toContain("text-faint");
+    cleanup();
+
+    // At zero running, the row is absent entirely — the old code would have
+    // shown "N background tasks finished" here.
+    const zero = render(<BackgroundWorkTranscriptRow runningCount={0} onOpen={() => {}} />);
+    expect(zero.container.firstChild).toBeNull();
+    expect(screen.queryByText(/finished/i)).toBeNull();
   });
 });
--- a/apps/packages/product-client/src/components/workspace/activity/BackgroundWorkTranscriptRow.tsx
+++ b/apps/packages/product-client/src/components/workspace/activity/BackgroundWorkTranscriptRow.tsx
@@ -1,19 +1,11 @@
 import type { ReactNode } from "react";
 import { ProliferateIcon } from "#product/primitives/icons/proliferate-icons";
-import { CircleCheck } from "#product/primitives/icons/status";
 import { Button } from "#product/primitives/Button";
-import { twMerge } from "#product/primitives/utils/tw-merge";
 
 export interface BackgroundWorkTranscriptRowProps {
   /** Still-live processes + running native subagents. */
   runningCount: number;
-  /** Finished work still worth reporting once nothing is running. */
-  finishedCount: number;
-  /**
-   * Opens the Background work pane. R1 ships only this row; the pane itself
-   * lands in R2, so callers wire a no-op seam until then — see the
-   * `onOpen={backgroundWorkPaneSeam}` call site.
-   */
+  /** Opens the Background work pane. */
   onOpen: () => void;
 }
 
@@ -23,38 +15,33 @@ function pluralizeTask(count: number): string {
 
 /**
  * The quiet line at the end of the transcript reporting live background work
- * (running bash processes + running native subagents — loops are descoped)
- * and opening the Background work pane on click (Design Handoff —
- * HANDOFF-background-work.md, NEW `BackgroundWorkTranscriptRow`; Delivery
- * Spec — Background Work Slice 1, rung R1). Replaces the docked
- * `ActivityChips` breakdown with one aggregate line — no "running", no
- * "2 terminals, 1 subagent" tail.
+ * (running bash processes + running native subagents — loops are descoped) and
+ * opening the Background work pane on click (Design Handoff —
+ * HANDOFF-background-work.md, NEW `BackgroundWorkTranscriptRow`; Delivery Spec
+ * — Background Work Slice 1, rung R1).
  *
- * No motion in either state: a spinner and the `ProliferateLivingMark`
- * breathing treatment were both tried and rejected. The glyph is static and
- * swaps to `CircleCheck` once settled; the mark and the label always share
- * one hover treatment.
+ * Founder ruling (2026-08-17, bgwork r6): this row counts RUNNING work only,
+ * and is NOT rendered at count 0. Completed tasks leave the count; the row
+ * disappears at 0. There is no settled/finished display state — completed work
+ * is announced solely by the inline `BackgroundCompletionReceipt` rows. The
+ * earlier "N background tasks finished" + `CircleCheck` swap is removed
+ * entirely.
+ *
+ * No motion: a spinner and the `ProliferateLivingMark` breathing treatment were
+ * both tried and rejected. The glyph is static; the mark and the label always
+ * share one hover treatment.
  */
 export function BackgroundWorkTranscriptRow({
   runningCount,
-  finishedCount,
   onOpen,
 }: BackgroundWorkTranscriptRowProps): ReactNode {
-  const running = runningCount > 0;
-  if (!running && finishedCount <= 0) {
+  if (runningCount <= 0) {
     return null;
   }
 
-  const label = running
-    ? pluralizeTask(runningCount)
-    : `${pluralizeTask(finishedCount)} finished`;
-
-  // Settled state rests at `text-faint` instead of `text-muted-foreground`;
-  // both states share the same hover treatment (mark + label light together).
-  const toneClassName = twMerge(
-    running ? "text-muted-foreground" : "text-faint",
-    "transition-colors duration-hover group-hover:text-foreground",
-  );
+  const label = pluralizeTask(runningCount);
+  const toneClassName =
+    "text-muted-foreground transition-colors duration-hover group-hover:text-foreground";
 
   return (
     <Button
@@ -65,11 +52,7 @@ export function BackgroundWorkTranscriptRow({
       className="group flex w-fit items-center gap-2 rounded-md text-chat"
     >
       <span className={toneClassName}>
-        {running ? (
-          <ProliferateIcon className="icon-paired [font-size:var(--text-chat)]" aria-hidden />
-        ) : (
-          <CircleCheck className="icon-paired [font-size:var(--text-chat)]" aria-hidden />
-        )}
+        <ProliferateIcon className="icon-paired [font-size:var(--text-chat)]" aria-hidden />
       </span>
       <span className={toneClassName}>{label}</span>
     </Button>
--- a/apps/packages/product-client/src/components/workspace/activity/background-pane/BackgroundWorkPane.finish-signals.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/activity/background-pane/BackgroundWorkPane.finish-signals.test.tsx
@@ -15,11 +15,11 @@ import {
 import { BackgroundWorkPane } from "./BackgroundWorkPane";
 
 // R5 finish-signal ladder concerns: pending-selection consumption, feed
-// isOpen-gating, and the NoticeBanner (dirty/mark-viewed interactions). The
-// roster/scopes/seam/detail-routing coverage lives in the sibling
-// `BackgroundWorkPane.test.tsx` — split for PROD-SIZE-1 (repo-wide 600-line
-// cap); see the fixtures module's docstring. No behavior or assertion
-// changed by the split.
+// isOpen-gating, and the mark-viewed write that clears the tab dirty-dot (the
+// in-pane NoticeBanner was removed in bgwork r6 round 2 — see the negative
+// controls below). The roster/scopes/seam/detail-routing coverage lives in the
+// sibling `BackgroundWorkPane.test.tsx` — split for PROD-SIZE-1 (repo-wide
+// 600-line cap); see the fixtures module's docstring.
 let sessionActivity: SessionActivityState = {
   loops: [],
   loopCapabilities: { supported: false, native: false },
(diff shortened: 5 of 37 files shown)


pablonyx:
## Review — PR #2021 (bgwork r6: inline completion receipts + running-only indicator)

**Verdict: CLEAN** (mergeable). Two LOW, non-blocking observations for founder awareness; no correctness blockers found. Reviewed adversarially in a scratch worktree off head `4f8f00a51` vs merge-base `10f17ca47` (base branch `bgwork/r4b-claude-transcript-view`). I did not write this code.

### Gate outputs (verbatim, run by me)

Touched vitest suites, batch 1 (7 files):
```
 Test Files  7 passed (7)
      Tests  98 passed (98)
```
Touched vitest suites, batch 2 (6 files):
```
 Test Files  6 passed (6)
      Tests  42 passed (42)
```
All 13 touched/owned suites green: **140 tests, 0 failed.**

Round-3 negative control — reverted `turnCompleted: !!turn.completedAt || tailMessageSettled` to `!!turn.completedAt`, re-ran `transcript-rendering.test.ts`:
```
      Tests  4 failed | 13 passed (17)
```
Exactly the 4 fix-dependent tests fail (the wake-turn copy-restore + the three interleave variants a/b/c); every negative control (streaming prose, prose-closed-while-tool-runs, reveal-incomplete, stamped-turn, non-final-row) stays green. Matches the PR's claim exactly. Fix restored afterward.

Discipline gates:
```
Design attribution check passed.
Frontend boundary check passed.
```
- `ratchets.toml`: the single entry (`transcript-row-model.ts`) was **tightened** 1017 → 856; no new entries, none loosened. Confirmed by diff.
- Test split keeps the load-bearing `.test.` mid-name segment (`transcript-row-model.background-work.test.ts`, `BackgroundWorkPane.finish-signals.test.tsx`).

### Verified (not merely read)
- **Round-3 footer gate is deliberately scoped.** The old `resolveTurnAssistantFooterMode` now has no callers other than the new `…ForRow` wrapper and its own unit test. Completion chrome (`visualTurnCompleted`, `latestCompletedTurnId`, diff panel, end-resource, frontier box) still keys on raw `!!turn.completedAt` — so the wake turn gets its copy/timestamp back **without** being promoted to the "session complete" marker or diff panel. Confirmed.
- **`turnHasActiveToolWork` covers subagent/background tool shapes.** It scans `turn.itemOrder` for any `tool_call` with status ≠ completed/failed; a subagent is a `tool_call` (semanticKind `subagent`) and nested tool work is caught (test "detects active nested tool work outside top-level display blocks" passes).
- **Row-model tolerance under reload / aged-out anchors.** `bucketCompletionReceiptRows` routes null-anchor receipts to `beforeFirstTurn`, still-loaded anchors after their turn, and receipts whose anchor turn is gone to `afterAllTurns` — no crash, degrades to tail placement. Indicator row returns `null` at `runningCount <= 0`.
- **Anchor stability:** `anchorTurnId` is captured via a ref at observation time (`anchorTurnIdRef`), so a later wake turn never restamps an already-emitted receipt.
- **Click-through plumbing is session-scoped.** `pendingBackgroundProcessSelectionByWorkspace` mirrors the existing subagent selection: written with the active `sessionId`, consumed by `BackgroundWorkPane` only when `selection.sessionId === sessionId`, and cleared unconditionally (matched or not). A cross-session entry is discarded, not applied — no wrong-session deep-open leak.
- **tsc:** every file this PR changes is type-clean. (The broader `tsc -p` run in my scratch tree emits unrelated `@proliferate/cloud-sdk`/`@anyharness/sdk` export errors, but those are artifacts of my symlinked `node_modules` pointing at main's SDK dist — a stale-dist mismatch, not attributable to this PR. I could not independently reproduce the PR's exact "only pre-existing TS2741" line for that reason, but confirmed no touched file regresses.)

### Findings (LOW — non-blocking)

**L1 — Round-3 footer can flash `copy` mid-turn on a multi-step turn.** The new gate promotes a turn to a copyable footer when the tail prose has stopped streaming AND no `tool_call` is active AND the reveal frontier has caught up — even if `turn.completedAt` is still null. That is exactly right for a runtime wake turn, but it is structurally indistinguishable from the gap **between one prose block finishing and the next `tool_call` item arriving** in an ordinary multi-step turn (e.g. "Let me check the logs." → Bash). In that window the SDK reducer has already set the prose `isStreaming=false` (`closeStreamingPointer` on `item_completed`), no tool is yet in-progress, and `completedAt` is null — so `tailMessageSettled` is true and the copy button + timestamp can briefly appear, then vanish when the tool starts. The old `!!turn.completedAt` gate never showed copy mid-turn, so this is a genuinely new window. Mitigants: the reveal-animation frontier keeps `assistantRevealComplete` false while text is still pacing in, and `turnHasActiveToolWork` covers the tool-running case — so the flash only opens once short prose has fully revealed before the next tool. Cosmetic, timing-dependent, and inherent to the synthesized (path-b) approach; I could not confirm it live (no running app in review). Worth a founder eye given round 2 was itself a "fades in weirdly" complaint. The suite covers streaming / prose-closed-while-tool-runs / reveal-incomplete but not this prose-settled-no-tool-yet gap.

**L2 — One-frame stale-receipt flash on session switch.** `useBackgroundCompletionReceipts` resets its `receipts` state inside an effect (async). On a session switch its host (`SessionTranscriptPane`) is not remounted, so the first render after the switch feeds the *previous* session's receipts to the row model against the *new* transcript; their `anchorTurnId` won't match, so they land in `afterAllTurns` and can paint for one frame at the tail before the effect clears them. Ephemeral data that doesn't survive reload anyway, and the click-through is separately session-guarded (L2 is display-only, not a wrong-session action). Very minor.

Neither finding blocks merge. The founder-ruled behavior (running-only indicator, inline receipts, NoticeBanner removal, wake-turn copy restore) is implemented faithfully and the disclosed limitations (D1 reload durability, subagent detail bounce-back) are honestly scoped in the PR body.

pablonyx:
## Delta review — round 4 (`4f8f00a51..1fb98411e`, 3 files)

**Verdict: CLEAN.** Round 4 correctly supersedes round 3 and closes my L1. Reviewed in a scratch worktree at `1fb98411e`; the range touches only `transcript-rendering.ts`, `transcript-rendering.test.ts`, `TranscriptTurnRow.tsx`. No ratchet/store/plumbing changes, so my prior CLEAN verdict on the rest stands unchanged.

### 1. `resolveSessionViewState` semantics — no mid-stream leak
Enumerated the full type: `working | needs_input | idle | errored | closed` (`activity-types.ts`). Traced `resolveSessionExecutionPhase` → `resolveSessionViewState`: during genuine token production the execution phase is `starting` or `running`, **both** map to `working`. There is **no** `responding`/`queued`/other mid-stream state that escapes `working` — `isSessionEffectivelyStreaming` (transcript.isStreaming true, not the idle+stream-ended case) forces phase `running` → `working`. So `turnIsActivelyStreaming = isLatestTurn && sessionViewState === "working"` is true for **every** actively-streaming turn, and copy cannot leak onto a live stream. The non-`working` states are all genuinely non-streaming (paused-for-input / done / failed / ended).

- Minor residual (non-blocking): `needs_input` (awaiting_interaction with an actionable pending interaction) is ≠ `working`, so a latest turn that is *paused for input* will show copy on its settled question message. That's a real pause (no tokens flowing), arguably desirable, and further gated by `assistantRevealComplete`. If an agent ever resumed streaming after an inline approval *without* ending the turn, copy could appear during the pause then hide on resume — a narrower flicker than round-3 L1, and I couldn't construct it from the current turn lifecycle. Noting, not blocking.

### 2. `isLatestTurn &&` — safe (moot for display)
The footer renders only when `rowIsLastTurnRow` (`resolveTurnAssistantFooterMode` returns `"none"` otherwise). A non-latest turn gets no footer regardless of `turnIsActivelyStreaming`, so an "old turn gets copy while a newer turn streams" case can't paint. A dropped-tail wake turn is the last row only when the session produces nothing after it (idle) — exactly when copy is wanted. The guard is correct and defensive.

### 3. Reload → idle (read-verified, not executed)
Confirmed from the resolver: a reloaded/cold session whose persisted execution phase is `idle`/`completed` yields view state `idle` even with a stale `transcript.isStreaming = true`; `isSessionEffectivelyStreaming` additionally forces false for `status === "idle"` + stream `ended`/`disconnected`. So the stuck-`in_progress` dropped-tail message becomes copyable after reload, as claimed. Caveat: the unit tests inject `turnIsActivelyStreaming` directly (a clean pure/impure boundary), so the reload→non-`working` resolution is **read-verified against the resolver**, not exercised end-to-end here. One theoretical edge — a reloaded slot with no `executionSummary`, `status` null, stale `isStreaming = true`, and the stream not marked ended — would resolve to `running`→`working` and keep the message reserved; in practice the persisted phase / stream-ended reconciliation prevents it. That's the sole read-only assumption in this review.

### 4. Both negative controls confirmed (literal output)
Revert `turnCompleted: … || messageSettledOffLiveStream` → `!!turn.completedAt`:
```
      Tests  5 failed | 14 passed (19)
```
Fails exactly the completion-dependent set — `round 3`, `round 4` (dropped-tail), and interleaving `(a)`/`(b)`/`(c)`; the two live-turn controls stay green.

Drop the `&& !turnIsActivelyStreaming` guard:
```
      Tests  3 failed | 16 passed (19)
```
Fails exactly the three live-turn controls — `round 4 control` (stuck-`in_progress` tail while streaming), `prose closed but tool still running (session working)`, and `L1: prose-finished-but-turn-still-active gap`. Both controls restored; clean suite:
```
 Test Files  1 passed (1)
      Tests  19 passed (19)
```
`TranscriptTurnRow.test.tsx`: `16 passed (16)`.

### 5. Gates at `1fb98411e`
```
Design attribution check passed.
```

### L1 closure — verified
The mid-turn prose→tool gap stays within a `working` session: the SDK reducer keeps `s.isStreaming = true` for the entire turn (only `turn_ended`/`session_ended` clear it; `item_completed` does not), so `turnIsActivelyStreaming` stays true across the gap → `reserved`. This is a genuine improvement over the round-3 item-status gate, and the dedicated `L1:` test pins it. The move from item-level status to session-level liveness is the right call given the durable wire shape (item stuck `in_progress` forever).

Verdict stands: **CLEAN**, merge-ready pending the usual sign-off. L2 (one-frame stale-receipt flash on session switch) remains the only open cosmetic from the prior review; untouched here and acknowledged known.

pablonyx:
## Delta review — round 5 (209c23663)

**Verdict: CLEAN**

Fresh reviewer, no hand in the implementation. Reviewed exactly the one round-5 delta `209c23663` ("key the frontier reserve on stream liveness, not the completion stamp") on top of the round-4-clean base; static review only (no builds/servers per machine limits).

**1. Scope — exactly the intended shape, nothing riding along.** Three files, +51/-1: `turn-row-presentation.ts` adds one `turnIsActivelyStreaming: boolean` param and changes the reserve branch to `rowIsLastTurnRow && isLatestTurnInProgress && turnIsActivelyStreaming ? "reserved" : "hidden"` (plus a doc comment); `TranscriptTurnRow.tsx` threads the already-computed `turnIsActivelyStreaming` (line 204) into the call; the test file adds 3 cases + one `base` field. No unrelated edits.

**2. Liveness predicate is correct and does not regress the legit reserve.** `turnIsActivelyStreaming = isLatestTurn && sessionViewState === "working"` (TranscriptTurnRow.tsx:191) — the exact predicate the round-4 footer fix keys on (`resolveTurnAssistantFooterModeForRow`, transcript-rendering.ts:144-145), so footer-settledness and frontier-reserve now agree on the same non-"working" turn. Genuinely streaming latest turn: `sessionViewState === "working"` → still `"reserved"` (anti-jitter preserved through the whole live turn). Dropped-tail idle wake turn: `isLatestTurnInProgress` stuck true forever, but session idle → `turnIsActivelyStreaming` false → `"hidden"`, killing the dead ~44-56px band. A real trailing status still wins `"status"` before the reserve branch, so a needs_input pause carrying an interaction indicator is unaffected; a needs_input turn with no trailing status now resolves `"hidden"` (was a dead reserve) — a deliberate, footer-coherent improvement, not a regression.

**3. Test quality — real negative control + counter-test both present.** "hides the reserve on a dropped-tail wake turn once the session goes idle (R5)" (test line 120) sets `turnIsActivelyStreaming:false` on an otherwise-live `base` and asserts `"hidden"`; under the pre-fix logic (`rowIsLastTurnRow && isLatestTurnInProgress`) that input returns `"reserved"`, so the test genuinely FAILS if the fix is reverted — a true negative control. The counter-test "keeps the reserve while the latest turn is genuinely streaming" (line 131) asserts the legit streaming case still reserves, and "shows a real trailing status regardless of liveness" (line 141) pins the needs_input-with-status path. Existing cases still pass because `base` gained `turnIsActivelyStreaming:true`.

**4. Four-render-path lesson — change lives at the layer all paths inherit.** The frontier-status box is a single per-turn-row sibling `<div data-turn-frontier-status>` (TranscriptTurnRow.tsx:303-310) rendered AFTER `TurnItemSequence`, not inside any item block. The four item render paths (TranscriptToolCallItemBlock, CollapsedActionRows→CommandActionRow / →ParsedCommandRows, TranscriptAgentGroupBlock's ScopedTranscriptBlocks) all live inside `TurnItemSequence`; the reserve follows the whole sequence regardless of which path produced the items. Single resolver call site (`resolveTurnFrontierStatusMode` — one consumer), single render site. No path bypasses it.

**5. Merge commit above does not invalidate any conclusion.** `bce0477e4` is a merge with parent1 = `209c23663` (this round-5 commit) and parent2 = r4b/refold base. `git diff 209c23663 bce0477e4 --` for all three scope files (turn-row-presentation.ts, TranscriptTurnRow.tsx, transcript-rendering.ts) is EMPTY — they came through the merge byte-identically; none appear in the merge's combined (conflict) diff. `sessionViewState` still flows into TranscriptTurnRow via MessageList.tsx:330, and MessageList.tsx was not reconciled by the merge. Round-4 footer logic and round-5 frontier logic are intact at HEAD.

No findings.

pablonyx:
**Refold receipt — bgwork r6 onto post-r7 main. Old reviewed tip `bce0477e4` → new head `968a765d1` (single squashed-refold commit on main `4cd69d806`). This is NOT an all-equals carry: a delta re-review is ordered and must pin `968a765d1`.**

**Why not mechanical:** the branch was merge-threaded — r6's six commits were authored on the OLD r4b tip `10f17ca47` with one final keep-up merge, and per-commit replay produced self-canceling pseudo-conflicts (commit 1 introduces the pane band, round 2's founder amendment deletes it). Construction used: net-delta 3-way apply of `10f17ca47..209c23663` (the spine's own commits) onto main.

**Receipt R1 (byte-identity):** 28 of 28 r6-touched files OUTSIDE the interaction set are byte-identical to the reviewed tip `bce0477e4`.

**Hand-integrated files (the entire divergence from the reviewed tree):**
1. `SessionTranscriptPane.tsx` — r6's band-machinery deletion carries; main's `.finally` → `.then((hydrated) => …)` hydration amendment preserved (auto-merged, verified); main's "Accepted residual" comment dies with the deleted block — the same call the reviewed keep-up made.
2. `SessionTranscriptPane.test.tsx` — r6's rewrite carries; main's hydration-test infra preserved (`hydrationMocks`, `includeTranscriptEntry` option, the two self-hydration tests, `waitFor` import); `fireEvent` import dropped (zero remaining uses); floating-band tests die as reviewed.
3. `transcript-row-list-model.ts` — main's PRO-187 extraction wins structurally (the estimator lives in `transcript-row-height-estimate.ts`); r6's local estimator additions removed here and ported (next item).
4. `transcript-row-height-estimate.ts` (not an r6 file — integration port): adds `ESTIMATED_BACKGROUND_WORK_ROW_HEIGHT_PX = 32` with a `completion_receipt`/`background_work` branch, and stable composition-token cases for the two new kinds — REQUIRED, since main's token fallback accesses `hostsWorkspaceReceipt`, which does not exist on the new row types (would not compile).

**Ratchets:** main's VirtualizedTranscriptRowList 405 entry kept (subsumes the keep-up's allowlist renames, verified); r6's MessageList 442 entry carries unchanged — merged MessageList.tsx measures **440 lines**, under the reviewed cap, no re-true.

**CI:** board re-fired at this head via retarget-to-main → close/reopen (the CONFLICTING-PR CI-skip cure): CI 32096569105, CodeQL 32096568998, Self-Host Smoke 32096569000, Intent Tests 32096569083, PR Metadata 32096569139.

Merge remains gated on: settled green board with literal tally, the delta re-review verdict, and the standing r6 decision point (D3 live re-verify) which goes to Pablo at this rung's turn.

pablonyx:
## Delta re-review at 968a765d1

Fresh, independent re-review of the refold-integration delta only, pinned to head `968a765d1d1cba71fb26ea59114cf03dd2719cdd` on main `4cd69d806`. Verified against old reviewed tip `bce0477e4`, r4b base `10f17ca47`, and r6 spine tip `209c23663`.

**Verdict: essentially CLEAN — one LOW-severity, NON-BLOCKING finding (F1). Does not block merge; a one-line follow-up is recommended.**

### What I verified (all pass)

- **R1 byte-identity.** Of the 37 r6-touched files (`git diff --name-only 10f17ca47 209c23663`), 28 are byte-identical to `bce0477e4` and 9 differ — the receipt's "28 of 28 outside the interaction set" holds numerically. Of the 9 differing: 4 are the disclosed hand-integrated files, and **5 more** (`MessageList.tsx`, `transcript-row-model.ts`, `chat-transcript-view-types.ts`, `use-chat-transcript-row-renderer.ts`, `use-chat-transcript-row-renderer.test.tsx`) also diverge — these were changed on BOTH main and r6 since r4b. I confirmed each is a **clean auto-merge**: the refold-on-main delta (`git diff origin/main 968a765d1 -- <file>`) is byte-identical to the r6 reviewed delta (`git diff 10f17ca47 209c23663 -- <file>`), differing only by `@@` line offsets from main-side additions above. No undisclosed edits; the receipt merely under-lists the interaction set by these 5. (Informational.)
- **No lost main-side edits.** None of the 28 byte-identical files were touched by main since r4b, so nothing main added was dropped.
- **SessionTranscriptPane.tsx** (`:141`): main's `.finally`→`.then((hydrated) => { if (!hydrated) return; … })` hydration amendment survives; r6's band machinery (`hasBackgroundWork`/`backgroundWorkRowHeightPx`/`messageListBottomInsetPx`) is fully deleted; main's "Accepted residual" comment correctly dies with the deleted block. `.finally` count 2→1 matches main (the surviving one is the unrelated older-history load at `:222`). Both of main's hunks accounted for.
- **SessionTranscriptPane.test.tsx**: `hydrationMocks` (`:30`), `includeTranscriptEntry` option (`:136`/`:150`), `waitFor` import (`:4`), and both self-hydration tests (`:171`, `:185`) survive; r6's rewrite otherwise carries; `fireEvent` import dropped with **zero** remaining uses.
- **transcript-row-list-model.ts**: head == main (identical). r6's only delta here was two estimator additions, both relocated by main's PRO-187 extraction into `transcript-row-height-estimate.ts`. Nothing lost.
- **transcript-row-height-estimate.ts (port target)**: `ESTIMATED_BACKGROUND_WORK_ROW_HEIGHT_PX = 32` added; `estimateRenderableRowHeight` gains the `completion_receipt`/`background_work` → 32 branch (`:149`); `getRowCompositionToken` gains the early `return row.row.kind` case for both kinds (`:258`) **before** the `pending:${row.row.hostsWorkspaceReceipt}` fallback (`:262`) — this is the type-correctness fix (the new kinds lack `hostsWorkspaceReceipt`); confirmed the guard prevents that path. No exhaustiveness/`assertNever` guards exist that the new kinds could break.
- **Ratchets**: `VirtualizedTranscriptRowList` = 405 (main's), `MessageList` = 442 with the file at 440 lines (≤442), plus a correct shrink-only re-true of `transcript-row-model.ts` 1017→856 from r6's own interleave-helper extraction.
- **Completeness (both directions)**: `git diff 4cd69d806 968a765d1` touches exactly 37 files = 36 r6 files carrying their reviewed delta + the port target; `transcript-row-list-model.ts` correctly equals main. No extraneous files; nothing from r6's reviewed delta dropped.

### F1 (LOW / non-blocking) — new row kinds miss `getRowEstimateBucketKey`, pooling into the "prompt" calibration bucket

`apps/packages/product-client/src/hooks/chat/ui/transcript-row-height-estimate.ts:224` (head `968a765d1`).

`getRowEstimateBucketKey` is a **post-r4b main addition** (Chat Scroll rung 5 / PRO-187 calibration; absent at `10f17ca47`), so the original r6 review never integrated against it. The refold correctly wired the two new row kinds into the two sibling functions in this module — `estimateRenderableRowHeight` (`:149`) and `getRowCompositionToken` (`:258`) — but **not** into `getRowEstimateBucketKey`. `completion_receipt` and `background_work` fall through to `return "prompt"` (`:224`).

Its sibling fixed-height quiet row, `goal_event`, returns `null` (`:218`), documented as "rows whose height is a small fixed constant not worth calibrating." The two new kinds are, by the port author's own comments, exactly that — single-line quiet rows at a fixed 32px with "no composition variants that change their height shape." Returning `"prompt"` instead of `null` means, per `transcript-row-height-calibration.ts` (`null` ⇒ never recorded/never applied): (a) their ~32px measured heights pollute the "prompt" bucket average (whose members estimate at 120px), and (b) the calibrated prompt-bucket average is served back to these rows, overriding their fixed 32px estimate. Net effect is minor virtualization-height inaccuracy for receipt/footer rows plus mild prompt-bucket pollution.

- **Not a type error / not a compile break** (the fallback returns a literal string), no crash, no functional regression to receipt/footer rendering.
- **Fix** (follow-up, not blocking): add `if (row.row.kind === "completion_receipt" || row.row.kind === "background_work") return null;` alongside the `goal_event` case at `:218`, mirroring the treatment the port already gave the other two functions.

I do not approve; this is a read-only re-review. The refold integration is faithful and merge-ready modulo the optional F1 follow-up.

pablonyx:
## Board settle at 968a765d1 — 22 pass / 2 fail, one exogenous + one real (founder-gated)

Fresh CI settled 22/2. The two reds, attributed from the job APIs:

**1. Transcript scroll physics (30m15s) — EXOGENOUS stall class, rerun dispatched.** Step receipt: `Install Playwright chromium + webkit=cancelled`, `Run scroll-physics suite=skipped` — zero tests executed. This is the same repo-wide Playwright-install stall that hit #2013 three times tonight and main's own runs independently. Sanctioned same-job rerun dispatched (run 32096569105).

**2. Login first-load budget (2m21s) — REAL measurement, NOT a flake, NOT rerun.** Literal violation:

```
JS 502025 B > cap 502000 B    (headroom js: -25)
```

Attribution: main's own tip measures 501970/502000 (30 bytes headroom, see #2037's finding). This refold's additions to the transcript row modules (two new row kinds wired into `transcript-row-height-estimate.ts` + `transcript-row-model.ts`) minify into the login entry chunk (`/assets/index-BF5dAEXb.js`) and add ~55 gzip bytes — genuine content growth against effectively zero headroom, not a chunk-leak bug. A rerun cannot change a deterministic measurement.

**Cure paths are founder-owned:** (a) merge the staged headroom PR #2025 first (5,321 bytes of relief; gated on the founder's "land headroom"), then re-run this gate; (b) an explicit cap ruling (thresholds are founder-only); (c) out-of-scope chunk surgery (not recommended for a refold). Note: the delta reviewer's optional F1 follow-up (comment above) would add a few more bytes — if F1 is fixed pre-merge, (a) is effectively required first.

**#2021 therefore holds** (draft, no ready-flip): delta re-review CLEAN (one LOW non-blocking finding), 22 checks green, physics rerun pending, budget red pending the founder's call. Merge gates: green board + the standing r6 decision point.

pablonyx:
## Physics rerun cured — board of record now 23 pass / 1 fail at 968a765d1

Same-job rerun of the stalled scroll-physics job **passed** (attempt 2, started 04:17:40Z, run 32096569105 — the stall never got past Playwright install on attempt 1, zero tests were lost). Literal current board on the run of record: **23 pass / 1 fail**.

The sole red remains **Login first-load budget**: `JS 502025 B > cap 502000 B` (headroom -25) — deterministic, real, founder-gated as disclosed above (cure = #2025 "land headroom" or a cap ruling).

Bookkeeping note: check-runs at this SHA also show a stale duplicate set from run 32096545353 (started 03:44:09Z) — that is the pre-close/reopen artifact superseded when CI was re-fired for this head (same artifact class as #2019, disclosed there too). Its job conclusions are not part of this PR's board of record.

#2021 status: delta re-review CLEAN (one LOW non-blocking F1), 23/24 green, budget red awaiting the founder. Draft stays draft.

pablonyx:
## Stale-base disclosure — 05:39Z board is NON-INFORMATIVE for the budget re-measure

The close/reopen at ~05:36Z minted a merge-ref against a stale base. Receipt from the budget job (95609107195) checkout line:

```
HEAD is now at a5dd6c3 Merge 968a765d1d1cba71fb26ea59114cf03dd2719cdd into 4cd69d806ac433a29bb6478e5ad9490be215b2c4
```

4cd69d806 is bg r7's squash (03:21Z) — it predates chat-scroll r8's merge (7b1a3d976, 05:20:59Z) which returned ~2.85 KB of login-budget headroom to main. The job's literal ("JS 502025 B > cap 502000 B") is byte-identical to the prior red because it measured the identical pre-relief tree. This is the known stale-merge-ref reopen race (caught the same way on #1975: job checkout line).

Action: cancelling the 05:39Z run set as non-informative, re-cycling close→(gap)→reopen, and the next board's budget verdict only counts if its checkout line shows the merge base containing 7b1a3d976.

pablonyx:
## Budget re-measure on the properly-based merge-ref — GATE PASSED, the wall is down

Job 95610918017 (run 32104403496, created 05:49:49Z after the gapped re-cycle). Checkout-line proof of the relieved base:

```
HEAD is now at ecb829d Merge 968a765d1d1cba71fb26ea59114cf03dd2719cdd into 7b1a3d9763a2b4f6c1dcdcd368935211fe6a98aa
```

Literals:

```
[login-budget] JS 499207/502000 B  CSS 65856/66250 B  fonts/images/audio font=0 image=0 audio=0
[login-budget] budget gate PASSED (within all ceilings).
```

Headroom at this merge-ref: **2,793 bytes**. The prior deterministic red (502025, -25 B against pre-relief main) is cured by chat-scroll r8's ~2.85 KB relief (7b1a3d976), exactly as predicted — r6's own ~55-byte addition now lands well inside the ceiling. The stale 05:39Z round stands disclosed above as non-informative.

Remaining on this PR: the rest of this board settling (tally to follow), then the review-of-record chain (already CLEAN + F1 named follow-up) and the founder's D3/F1/merge decisions.

pablonyx:
## Board of record — FULLY GREEN on the proven-fresh merge-ref

Cycle of 05:49:49Z (checkout base 7b1a3d976 proven in the comment above). All five workflow runs settled:

```
CI: completed success
Intent Tests: completed success
CodeQL: completed success
Self-Host Smoke: completed success
PR Metadata: completed success
```

All 13 CI jobs, verbatim conclusions:

```
Build SDK: success
Terraform validate (server/infra): success
Desktop frontend build: success
Mobile typecheck: success
Cargo check & test: success
CI/CD config checks: success
Repo shape checks: success
Candidate build handoff: success
Login first-load budget (phase-6): success   <- the former sole red, cured (JS 499207/502000 B)
Web frontend build: success
Workflow definition lifecycle (tier-2): success
Transcript scroll physics (chromium + webkit): success   <- first-attempt pass this cycle
Shared frontend packages: success
```

This supersedes the prior 23/1 board. Review chain of record: net-delta refold receipt + independent delta review CLEAN (one LOW non-blocking finding, F1, named follow-up). Nothing technical remains open on this PR — remaining gates are the founder's alone (D3 live re-verify choice, F1 disposition, merge word).

pablonyx:
MERGE BOARD OF RECORD (re-mint cycle 3, supersedes 5324260334's cycle): reopen 06:37:39Z, merge-ref aa458faaa pre-flight-verified parents [868d53256 (r11 squash), 968a765d1]. All 5 runs success (CI 32107737674, Intent 32107737604, CodeQL 32107737673, Smoke 32107737501, Metadata 32107737086); all 13 CI jobs success verbatim incl. Workflow definition lifecycle (tier-2) and Transcript scroll physics first-attempt. Checkout-line base proof: 'HEAD is now at aa458fa Merge 968a765d1... into 868d53256...'. Budget literal: '[login-budget] JS 499199/502000 B ... budget gate PASSED (within all ceilings)' — headroom 2,801 B. Merging on the founder's explicit word ('merge', direct user turn ~06:1xZ); D3 live re-verify waived by that choice; F1 LOW rides as the named post-merge follow-up (reviewer's one-liner, own small PR). Interleave with the CS train receipted on both PRs (r11 zero-overlap proceeded; r12 holds until this lands).

PR #2022, opened 2026-08-17
feat(product): background command tool-call opens terminal detail (bgwork R8)

pablonyx:
## Summary

Rung r8 of the Background Work stack. Clicking a **backgrounded** Bash tool-call row in the chat transcript now opens the Background Work pane's terminal detail view — the same click-in behavior a subagent-creation row already has — instead of toggling the inline output disclosure. Foreground commands, and background rows whose result text doesn't yield a known process id, are unaffected: they keep the existing inline-disclosure toggle byte-for-byte.

## What changed

- `domain/chats/tools/background-command-correlation.ts` (new): parses the runtime's literal `Command running in background with ID: {taskId}` sentence out of a Bash tool's result text to recover the `ActivityProcess.id` used by the background-work roster. There is no wire-level tool_call_id correlation, so this is the only link available.
- `domain/activity/process.ts`: adds `processTrailingStatusLabel`, a fine-grained `running · 4m 12s` / `exited 0 · 2m 45s` label for the row's trailing status slot, matching the "Chat - Background Work Indicator" design handoff.
- `tool-calls/ToolActionRow.tsx`: adds an `onOpen` override — when set, activating the row calls it instead of expanding in place. Fully backward compatible; every other consumer never passes it, so their behavior is unchanged.
- `tool-calls/BashCommandCall.tsx`: forwards a new `onOpenBackgroundTerminal` prop to `ToolActionRow`'s `onOpen`.
- `TranscriptToolCallItemBlock.tsx` → `TranscriptItemBlock.tsx` → `TranscriptToolCallGroupBlock.tsx` / `TranscriptTreeNode.tsx` → `TurnItemSequence.tsx` → `TranscriptTurnRow.tsx` → `MessageList.tsx`: thread the `onOpenBackgroundTerminal(processId)` callback down to the row, wired at the top to the existing `useOpenBackgroundTerminalDetail` hook from r6 (no new pane plumbing needed).
- Roster lookup (`useSessionActivityForSession` + `useTranscriptSessionId`) is a local hook call inside `TranscriptToolCallItemBlock` rather than a threaded prop, so only the OPEN callback needed prop-threading — keeps the diff to the workflow action only.

## Behavior

- Background row, id parses, roster data present: click opens terminal detail; trailing slot shows `running · Xm Ys` / `exited N · Xm Ys`.
- Background row, id parses, roster data absent (e.g. after reload): click still opens terminal detail; trailing slot stays empty.
- Foreground row, or background row whose result text doesn't parse to an id: click keeps toggling inline disclosure, byte-identical to prior behavior.

## Test plan

- [x] `pnpm exec vitest run` targeted suites (11 files / 108 tests, all passing) covering: id-parsing correlation (incl. negative controls), trailing-status label formatting, `ToolActionRow`'s `onOpen` override, and full click-in wiring through `TranscriptToolCallItemBlock` and `MessageList` — including both mandated negative controls (foreground command, near-miss result text).
- [x] `python3 scripts/check_max_lines.py` — `Max-lines check passed (repo max 600, component max 500, server layer maxes enabled).`
- [x] Design-attribution lint — `Design attribution check passed.`
- [x] `tsc -p tsconfig.json --noEmit` — no new errors introduced (one pre-existing, unrelated error in `run-view-model.ts` confirmed present on base branch too).

Base SHA (merge-base with `origin/bgwork/r6-completion-receipts`): `d66ed0579995e10c1d798aa230461e448ec894cd`
Head SHA: `c90cdb3d67a8ba5e759e194998af63bcbc2bcddb`

---

## Round 2 (live-verification fix-forward)

Live verification against the served preview found a real gap: round 1 only wired click-in through `TranscriptToolCallItemBlock`, the still-*expanded* live-turn path. Every turn collapses into a "Worked for Ns" ledger once it completes, and that ledger renders its command rows through an entirely separate component tree — `CollapsedActions` → `CollapsedActionRows` → `CommandActionRow` — which round 1 never touched. Since every finished turn collapses, this was the path the founder actually saw, and it still opened the inline Shell disclosure instead of the terminal detail.

### What changed

- `tool-calls/CollapsedCommandActionRow.tsx` (`CommandActionRow`): reuses `parseBackgroundCommandProcessId` (no new parsing) and `processTrailingStatusLabel` against roster data read via the same `useTranscriptSessionId` + `useSessionActivityForSession` local-hook convention `TranscriptToolCallItemBlock` already uses. A background command with a parseable id never expands in place; activating it calls the threaded `onOpenBackgroundTerminal` instead.
- The pane-opening callback is a **threaded prop** here, not a local `useOpenBackgroundTerminalDetail()` call: that hook reaches into `useWorkspaces()`/`useProductAuthUserId()`, which requires a fully equipped `ProductHostProvider`. Calling it unconditionally inside `CommandActionRow` broke every other ledger test that mounts `CollapsedActions` with a minimal test host (caught by the targeted suite, not shipped). `onOpenBackgroundTerminal` now threads `CollapsedActionRows` → `CollapsedActions` → `ScopedTranscriptBlocks`'s `TurnDisplayBlockNode`, terminating at the `onOpenBackgroundTerminal` prop `TurnItemSequence` already carried from round 1 (both of its `TurnDisplayBlockNode` call sites now forward it) — no new top-level wiring needed.
- `tool-calls/CollapsedActionRowPrimitives.tsx` (`ActionDisclosureRow`): adds the same muted trailing-status slot (`running · 4m 12s` / `exited 0 · 2m 45s`) round 1 added to the top-level row.

### Behavior (collapsed-ledger path)

- Background row, id parses, callback wired, roster data present: click opens terminal detail; trailing slot shows `running · Xm Ys` / `exited N · Xm Ys`.
- Background row, id parses, roster data absent (e.g. after reload): click still opens terminal detail; trailing slot stays empty.
- Background row, id parses, but the caller hasn't wired `onOpenBackgroundTerminal` at all (e.g. the read-only playground transcript): falls back to the ordinary inline toggle rather than no-op'ing or throwing.
- Foreground row, or background row whose result text doesn't parse to an id: click keeps toggling inline disclosure, byte-identical to prior behavior.

### Test plan (round 2)

- [x] New `CollapsedCommandActionRow.background-terminal-clickin.test.tsx` (7 tests): click-in, roster-present trailing status, roster-absent fallback, unwired-caller fallback, and both mandated negative controls (foreground command, near-miss result text) — mirroring round 1's coverage for this second path.
- [x] Full targeted suite after the fix, 14 files: `Test Files  14 passed (14)` / `Tests  133 passed (133)`.
- [x] Broader regression sweep, whole chat/playground surface: `Test Files  83 passed (83)` / `Tests  580 passed (580)`.
- [x] `python3 scripts/check_max_lines.py` — `Max-lines check passed (repo max 600, component max 500, server layer maxes enabled).`
- [x] Design-attribution lint — `Design attribution check passed.`
- [x] `tsc -p tsconfig.json --noEmit` — no new errors (same single pre-existing, unrelated `run-view-model.ts` error confirmed present on base branch too).

Head SHA: `0ddafdb8c5304d335c1e166fa199626588a653b0`

---

## Round 3 (parsed/compound rows + nested subagent transcripts, live-verification fix-forward)

Live verification (fiber/DOM walk against the served preview) found round 2's fix still didn't cover the founder's actual command shape. `CollapsedActionRows` has a FIRST branch, checked before `classifyCollapsedAction`'s switch ever runs: `if (getToolCallParsedCommands(item).length > 0) return <ParsedCommandRows .../>`. Any Bash tool call whose harness supplies a structured `parsed_cmd` breakdown — including the founder's own `for i in $(seq 1 15); do echo probe $i; sleep 1; done` — takes that branch and never reached round 2's wired `CommandActionRow`. Confirmed live: the row rendered via `ParsedCommandRows` and clicking still opened the inline Shell disclosure.

A delta reviewer posted CLEAN on round 2 but treated this exclusion as acceptable ("static rows, no half-wired state"); overridden as not acceptable for this purpose, since the founder's reported command renders through exactly this branch.

### What changed

- `hooks/activity/derived/use-background-command-status.ts` (new): extracts the roster-lookup logic (session id → roster processes → parse the result text's `Command running in background with ID: ...` sentence → resolve the trailing status label) shared by `CommandActionRow`, `CollapsedActionRows`, and `TranscriptToolCallItemBlock`, so it's written once instead of three times.
- `tool-calls/CollapsedActionRows.tsx`: resolves `backgroundProcessId`/`backgroundTrailingStatus` **once per tool call** via the shared hook, before branching on `parsedCommands.length > 0`. A background command is one process with one result text no matter how many structural rows its parsed breakdown renders into, so `ParsedCommandRows` now threads the same id/status uniformly to every row it renders — never a half-wired subset where some rows open the pane and others keep their old per-kind treatment (e.g. a `read`-kind row's file link). Falls back to the exact original per-kind rendering (`FileActionRow` / `PlainActionRow`) when there's no id to open with, or no caller wired to open it.
- `tool-calls/CollapsedCommandActionRow.tsx`: refactored to call the new shared hook instead of inlining the roster lookup; no behavior change.
- `transcript/TranscriptToolCallItemBlock.tsx`: same refactor, which also incidentally shaved this file back under the repo's `FE-SIZE-1` 400-line soft threshold (`report_frontend_structure.py`) — it had grown to 402 lines in round 1.
- `transcript/TranscriptAgentGroupBlock.tsx` / `TranscriptToolCallGroupBlock.tsx`: a second review finding — `TranscriptAgentGroupBlock` renders its own nested `ScopedTranscriptBlocks` for a native subagent's own transcript, but couldn't receive `onOpenBackgroundTerminal`, so a background command run *by* a native subagent got click-in on its expanded row but not on its collapsed ledger row (same gap as round 2, one level down). Threaded the same way `onOpenSubagent` already flows through this component.
- Test-naming honesty nit (same review): `CollapsedCommandActionRow.background-terminal-clickin.test.tsx`'s "standalone playground" test actually passed the callback despite its negative-control framing — it was a pure duplicate of the genuine unwired-caller test in the same file, so it's removed rather than renamed.

### Behavior (parsed/compound-command path)

- A tool call whose harness supplies a `parsed_cmd` breakdown (e.g. the founder's for-loop), where the result text parses to a background process id: **every** parsed row opens the same terminal detail on click, with the roster's trailing status on each row (the design artifact is silent on multi-row treatment for a background command, so this follows the coordinator's fallback of "trailing status on each parsed row").
- Same shape, but the caller hasn't wired `onOpenBackgroundTerminal`: falls back to the ordinary per-kind rendering (byte-identical to pre-r8).
- Same shape, but the result text doesn't parse to an id (foreground compound command, or unrecognized/near-miss text): falls back to the ordinary per-kind rendering, byte-identical, and never calls the callback.
- A background command run *inside* a native subagent's own nested transcript: its collapsed ledger row now opens the terminal detail too, same as the top-level transcript.

### Test plan (round 3)

- [x] New `CollapsedActionRows.parsed-background-terminal-clickin.test.tsx` (7 tests): confirms the fixture takes the `ParsedCommandRows` branch, click-in for the founder's exact for-loop shape, roster trailing-status display, "every row of a multi-command call opens the same terminal — no half-wired subset," and three negative controls (unwired caller, foreground compound command, near-miss result text).
- [x] New `TranscriptAgentGroupBlock.background-terminal-clickin.test.tsx` (2 tests): click-in for a background command nested in a native subagent's own transcript, and the unwired-caller negative control. Split into its own file (not added to `TranscriptAgentGroupBlock.test.tsx`) because that file was already at 606 lines against the repo-wide 600-line cap.
- [x] `CollapsedCommandActionRow.background-terminal-clickin.test.tsx` re-verified after removing the dishonest duplicate test: `Test Files  1 passed (1)` / `Tests  6 passed (6)`.
- [x] Targeted suite: `Test Files  2 passed (2)` / `Tests  13 passed (13)` (the two new parsed/compound + refactored-command-row files together).
- [x] `TranscriptToolCallItemBlock` suites after the shared-hook refactor: `Test Files  2 passed (2)` / `Tests  28 passed (28)`.
- [x] `TranscriptAgentGroupBlock` suites after the split: `Test Files  2 passed (2)` / `Tests  18 passed (18)`.
- [x] Broader regression sweep, whole chat/playground surface: `Test Files  85 passed (85)` / `Tests  588 passed (588)`.
- [x] `python3 scripts/check_max_lines.py` — `Max-lines check passed (repo max 600, component max 500, server layer maxes enabled).`
- [x] `python3 scripts/report_frontend_structure.py --strict --summary-only` — `FE-SIZE-1: 1` (only `MessageList.tsx`, a pre-existing violation from this branch's `r6` base being fixed on #2021 — not touched here per explicit instruction; `TranscriptToolCallItemBlock.tsx`'s round-1 violation is resolved).
- [x] Design-attribution lint — `Design attribution check passed.`
- [x] `tsc -p tsconfig.json --noEmit` — no new errors (same single pre-existing, unrelated `run-view-model.ts` error confirmed present on base branch too).

Head SHA: `cd4945f63`

### DIFF
--- a/apps/packages/product-client/src/components/workspace/chat/tool-calls/BashCommandCall.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/tool-calls/BashCommandCall.tsx
@@ -10,6 +10,13 @@ interface BashCommandCallProps {
   output?: string;
   status: "running" | "completed" | "failed";
   duration?: string;
+  /**
+   * Set when this command was backgrounded and its `ActivityProcess.id`
+   * correlated from the tool result text (bgwork r8). Clicking the row then
+   * opens the Background work pane's terminal detail instead of toggling the
+   * inline output disclosure.
+   */
+  onOpenBackgroundTerminal?: () => void;
 }
 
 export function BashCommandCall({
@@ -18,6 +25,7 @@ export function BashCommandCall({
   output,
   status,
   duration,
+  onOpenBackgroundTerminal,
 }: BashCommandCallProps) {
   const label = description
     ?? (status === "failed" ? "Command" : "Running command");
@@ -29,6 +37,7 @@ export function BashCommandCall({
       hint={command}
       status={status}
       duration={duration}
+      onOpen={onOpenBackgroundTerminal}
     >
       {output && (
         <ToolActionDetailsPanel>
--- a/apps/packages/product-client/src/components/workspace/chat/tool-calls/CollapsedActionRowPrimitives.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/tool-calls/CollapsedActionRowPrimitives.tsx
@@ -63,12 +63,19 @@ export function ActionDisclosureRow({
   expanded,
   failed,
   onToggle,
+  trailing,
 }: {
   label: string;
   icon: ReactNode;
   expanded: boolean;
   failed: boolean;
   onToggle: () => void;
+  /**
+   * A muted trailing status suffix (e.g. `running · 4m 12s`), shown when a
+   * background command's roster data is available (bgwork r8). Absent for
+   * every other row.
+   */
+  trailing?: string;
 }) {
   return (
     <Button
@@ -84,6 +91,9 @@ export function ActionDisclosureRow({
     >
       <ActionRowIcon>{icon}</ActionRowIcon>
       <span className="min-w-0 truncate">{label}</span>
+      {trailing && (
+        <span className="ml-auto shrink-0 text-faint">{trailing}</span>
+      )}
     </Button>
   );
 }
--- a/apps/packages/product-client/src/components/workspace/chat/tool-calls/CollapsedActionRows.parsed-background-terminal-clickin.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/tool-calls/CollapsedActionRows.parsed-background-terminal-clickin.test.tsx
@@ -107,11 +107,21 @@ function expandLedger() {
 describe("CollapsedActionRows — parsed/compound background command click-in (bgwork r8 round 3)", () => {
   beforeEach(() => {
     mocks.sessionProcesses = [];
+    // The collapsed-ledger disclosure (`AnimatedCollapsibleContent`) keeps its
+    // freshly mounted subtree `inert`/`aria-hidden` until one
+    // `requestAnimationFrame` after the toggle commits. jsdom never advances a
+    // real frame between a synchronous `fireEvent.click` and the next query, so
+    // drive the frame inline — the convention `CollapsedActions.test.tsx` uses.
+    vi.spyOn(window, "requestAnimationFrame").mockImplementation((callback) => {
+      callback(0);
+      return 1;
+    });
   });
 
   afterEach(() => {
     cleanup();
     vi.useRealTimers();
+    vi.restoreAllMocks();
   });
 
   it("takes the ParsedCommandRows branch for a harness-supplied parsed_cmd breakdown (confirms the reproduction)", () => {
@@ -155,7 +165,10 @@ describe("CollapsedActionRows — parsed/compound background command click-in (b
   });
 
   it("shows the roster's trailing status text on the parsed row when roster data is present", () => {
-    vi.useFakeTimers();
+    // Fake only the clock (roster elapsed time is `Date.now()`-derived) so the
+    // synchronous `requestAnimationFrame` spy from `beforeEach` survives —
+    // faking timers wholesale would replace it and re-strand the inert subtree.
+    vi.useFakeTimers({ toFake: ["Date"] });
     vi.setSystemTime(new Date("2026-08-17T10:04:12.000Z"));
     mocks.sessionProcesses = [rosterProcess({ startedAt: "2026-08-17T10:00:00.000Z" })];
     const transcript = createTranscriptState("session-1");
--- a/apps/packages/product-client/src/components/workspace/chat/tool-calls/CollapsedActionRows.parsed-background-terminal-clickin.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/tool-calls/CollapsedActionRows.parsed-background-terminal-clickin.test.tsx
@@ -1,0 +1,276 @@
+// @vitest-environment jsdom
+//
+// bgwork r8 round 3: live verification found round 2's fix only covered the
+// `classifyCollapsedAction` "command" branch (`CommandActionRow`).
+// `CollapsedActionRows`' FIRST branch checks `getToolCallParsedCommands`
+// before that switch ever runs — any tool call whose harness supplies a
+// `parsed_cmd` breakdown (including the founder's own
+// `for i in $(seq 1 15); do echo probe $i; sleep 1; done`) takes the
+// `ParsedCommandRows` branch instead and never reached round 2's fix.
+//
+// A background command is still ONE process with ONE result text no matter
+// how many structural rows its parsed breakdown renders into, so the id is
+// resolved once for the whole tool call and EVERY parsed row opens the same
+// terminal detail — never a mix of some rows wired and others not.
+import type { PropsWithChildren, ReactElement } from "react";
+import {
+  cleanup,
+  fireEvent,
+  render as testingRender,
+  screen,
+} from "@testing-library/react";
+import { createTranscriptState } from "@anyharness/sdk";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { ProductHost } from "@proliferate/product-client/host/product-host";
+import { ProductHostProvider } from "@proliferate/product-client/host/ProductHostProvider";
+import type { ActivityProcessWire } from "#product/domain/activity/process";
+import { toolCallItem } from "#product/lib/domain/chat/__fixtures__/playground/tool-call-item-fixture";
+import { TranscriptContextProviders } from "#product/components/workspace/chat/transcript/TranscriptContexts";
+import { CollapsedActions } from "#product/components/workspace/chat/tool-calls/CollapsedActions";
+
+const webTestHost = { desktop: null } as ProductHost;
+
+function Wrapper({ children }: PropsWithChildren) {
+  return (
+    <ProductHostProvider host={webTestHost}>
+      <TranscriptContextProviders sessionId="session-1">
+        {children}
+      </TranscriptContextProviders>
+    </ProductHostProvider>
+  );
+}
+
+function render(ui: ReactElement) {
+  return testingRender(ui, { wrapper: Wrapper });
+}
+
+const mocks = vi.hoisted(() => ({
+  sessionProcesses: [] as ActivityProcessWire[],
+}));
+
+vi.mock("#product/hooks/activity/derived/use-session-activity", () => ({
+  useSessionActivityForSession: () => ({
+    loops: [],
+    loopCapabilities: { supported: false, native: false },
+    processes: mocks.sessionProcesses,
+    agents: [],
+  }),
+}));
+
+const FOR_LOOP_COMMAND = "for i in $(seq 1 15); do echo probe $i; sleep 1; done";
+
+/** The command carries regex metacharacters ($, (, )) — escape before using
+ * it as a `getByRole` name matcher. */
+function commandNameMatcher(command: string): RegExp {
+  return new RegExp(command.replace(/[.*+?^${}()|[\]\\]/g, "\\$&"));
+}
+
+/** A tool call whose harness supplies a structured `parsed_cmd` breakdown —
+ * the shape that takes `CollapsedActionRows`' `ParsedCommandRows` branch
+ * instead of the `classifyCollapsedAction` "command" switch case. */
+function parsedCommandBashItem(
+  resultText: string,
+  parsedCommands: Array<{ type: string; cmd: string }> = [{ type: "unknown", cmd: FOR_LOOP_COMMAND }],
+) {
+  return toolCallItem({
+    nativeToolName: "Bash",
+    toolKind: "execute",
+    status: "in_progress",
+    rawInput: {
+      command: ["/bin/zsh", "-lc", FOR_LOOP_COMMAND],
+      parsed_cmd: parsedCommands,
+    },
+    contentParts: [
+      { type: "tool_result_text", text: resultText },
+    ],
+  });
+}
+
+function rosterProcess(overrides: Partial<ActivityProcessWire> = {}): ActivityProcessWire {
+  return {
+    id: "bn30h453a",
+    command: FOR_LOOP_COMMAND,
+    cwd: null,
+    status: { status: "running" },
+    pid: null,
+    startedAt: "2026-08-17T10:00:00.000Z",
+    endedAt: null,
+    feed: null,
+    ...overrides,
+  };
+}
+
+function expandLedger() {
+  fireEvent.click(screen.getByRole("button", { name: /Running command/i }));
+}
+
+describe("CollapsedActionRows — parsed/compound background command click-in (bgwork r8 round 3)", () => {
+  beforeEach(() => {
+    mocks.sessionProcesses = [];
+  });
+
+  afterEach(() => {
+    cleanup();
+    vi.useRealTimers();
+  });
+
+  it("takes the ParsedCommandRows branch for a harness-supplied parsed_cmd breakdown (confirms the reproduction)", () => {
+    const transcript = createTranscriptState("session-1");
+    transcript.itemsById = {
+      cmd: parsedCommandBashItem("Command running in background with ID: bn30h453a"),
+    };
+
+    // No onOpenBackgroundTerminal wired here — this test only confirms the
+    // fixture takes CollapsedActionRows' ParsedCommandRows branch (the
+    // founder's own command text renders as the row's label), which is the
+    // row CommandActionRow (round 2) never reaches. Click-in itself is
+    // asserted by the next test.
+    render(<CollapsedActions itemIds={["cmd"]} transcript={transcript} />);
+    expandLedger();
+
+    expect(screen.getByText(commandNameMatcher(FOR_LOOP_COMMAND))).toBeTruthy();
+  });
+
+  it("opens the background terminal detail on click for the founder's for-loop shape", () => {
+    const onOpenBackgroundTerminal = vi.fn();
+    const transcript = createTranscriptState("session-1");
+    transcript.itemsById = {
+      cmd: parsedCommandBashItem("Command running in background with ID: bn30h453a"),
+    };
+
+    render(
+      <CollapsedActions
+        itemIds={["cmd"]}
+        transcript={transcript}
+        onOpenBackgroundTerminal={onOpenBackgroundTerminal}
+      />,
+    );
+    expandLedger();
+    const row = screen.getByRole("button", { name: commandNameMatcher(FOR_LOOP_COMMAND) });
+
+    fireEvent.click(row);
+
+    expect(onOpenBackgroundTerminal).toHaveBeenCalledWith("bn30h453a");
+    expect(row.getAttribute("aria-expanded")).toBe("false");
+  });
+
+  it("shows the roster's trailing status text on the parsed row when roster data is present", () => {
+    vi.useFakeTimers();
+    vi.setSystemTime(new Date("2026-08-17T10:04:12.000Z"));
+    mocks.sessionProcesses = [rosterProcess({ startedAt: "2026-08-17T10:00:00.000Z" })];
+    const transcript = createTranscriptState("session-1");
+    transcript.itemsById = {
+      cmd: parsedCommandBashItem("Command running in background with ID: bn30h453a"),
+    };
+
+    render(
+      <CollapsedActions
+        itemIds={["cmd"]}
+        transcript={transcript}
+        onOpenBackgroundTerminal={vi.fn()}
+      />,
+    );
+    expandLedger();
+
+    expect(screen.getByText("running · 4m 12s")).toBeTruthy();
+  });
+
+  it("every parsed row of a multi-command background call opens the same terminal detail — no half-wired subset", () => {
+    const onOpenBackgroundTerminal = vi.fn();
+    const transcript = createTranscriptState("session-1");
+    transcript.itemsById = {
+      cmd: parsedCommandBashItem(
+        "Command running in background with ID: bn30h453a",
+        [
+          { type: "unknown", cmd: "mkdir -p /tmp/probe" },
+          { type: "unknown", cmd: FOR_LOOP_COMMAND },
+        ],
+      ),
+    };
+
+    render(
+      <CollapsedActions
+        itemIds={["cmd"]}
+        transcript={transcript}
+        onOpenBackgroundTerminal={onOpenBackgroundTerminal}
+      />,
+    );
+    expandLedger();
+
+    const firstRow = screen.getByRole("button", { name: /mkdir -p \/tmp\/probe/ });
+    const secondRow = screen.getByRole("button", { name: commandNameMatcher(FOR_LOOP_COMMAND) });
+
+    fireEvent.click(firstRow);
+    fireEvent.click(secondRow);
+
+    // Same process id from both rows — the whole tool call is one process.
+    expect(onOpenBackgroundTerminal).toHaveBeenNthCalledWith(1, "bn30h453a");
+    expect(onOpenBackgroundTerminal).toHaveBeenNthCalledWith(2, "bn30h453a");
+    expect(onOpenBackgroundTerminal).toHaveBeenCalledTimes(2);
+  });
+
+  it("negative control: an unwired caller (no onOpenBackgroundTerminal prop) keeps the ordinary per-kind rendering", () => {
+    const transcript = createTranscriptState("session-1");
+    transcript.itemsById = {
+      cmd: parsedCommandBashItem("Command running in background with ID: bn30h453a"),
+    };
+
+    // No onOpenBackgroundTerminal passed — mirrors a caller that hasn't wired
+    // the pane opener (e.g. the read-only playground transcript).
+    render(<CollapsedActions itemIds={["cmd"]} transcript={transcript} />);
+    expandLedger();
+
+    // Falls back to the ordinary, non-interactive PlainActionRow — no button
+    // role at all for this row (that's the "ordinary per-kind rendering").
+    expect(screen.queryByRole("button", { name: commandNameMatcher(FOR_LOOP_COMMAND) })).toBeNull();
+    expect(screen.getByText(commandNameMatcher(FOR_LOOP_COMMAND))).toBeTruthy();
+  });
+
+  it("negative control: a foreground compound command keeps byte-identical per-kind rendering", () => {
+    const onOpenBackgroundTerminal = vi.fn();
+    const transcript = createTranscriptState("session-1");
+    transcript.itemsById = {
+      cmd: parsedCommandBashItem(
+        "probe 1\nprobe 2\nprobe 3",
+        [{ type: "unknown", cmd: FOR_LOOP_COMMAND }],
+      ),
+    };
+
+    render(
+      <CollapsedActions
+        itemIds={["cmd"]}
+        transcript={transcript}
+        onOpenBackgroundTerminal={onOpenBackgroundTerminal}
+      />,
+    );
+    expandLedger();
+
+    // Never became a clickable, terminal-opening row.
+    expect(screen.queryByRole("button", { name: commandNameMatcher(FOR_LOOP_COMMAND) })).toBeNull();
+    const label = screen.getByText(commandNameMatcher(FOR_LOOP_COMMAND));
+    expect(label).toBeTruthy();
+
+    fireEvent.click(label);
+    expect(onOpenBackgroundTerminal).not.toHaveBeenCalled();
+  });
+
+  it("negative control: a near-miss sentence that doesn't parse an id keeps ordinary per-kind rendering", () => {
+    const onOpenBackgroundTerminal = vi.fn();
+    const transcript = createTranscriptState("session-1");
+    transcript.itemsById = {
+      cmd: parsedCommandBashItem("The command is now running in the background, ID: bn30h453a"),
+    };
+
+    render(
+      <CollapsedActions
+        itemIds={["cmd"]}
+        transcript={transcript}
+        onOpenBackgroundTerminal={onOpenBackgroundTerminal}
+      />,
+    );
+    expandLedger();
+
+    expect(screen.queryByRole("button", { name: commandNameMatcher(FOR_LOOP_COMMAND) })).toBeNull();
+    expect(onOpenBackgroundTerminal).not.toHaveBeenCalled();
+  });
+});
--- a/apps/packages/product-client/src/components/workspace/chat/tool-calls/CollapsedActionRows.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/tool-calls/CollapsedActionRows.tsx
@@ -10,26 +10,55 @@ import {
 } from "#product/domain/chats/transcript/transcript-tool-commands";
 import {
   basename,
+  deriveCommandOutput,
   deriveReadPathTarget,
   formatFetchLabel,
   formatListingLabel,
   formatParsedCommandLabel,
   formatSearchLabel,
 } from "#product/domain/chats/tools/collapsed-action-labels";
+import { useBackgroundCommandStatus } from "#product/hooks/activity/derived/use-background-command-status";
 import { CommandActionRow } from "#product/components/workspace/chat/tool-calls/CollapsedCommandActionRow";
 import { CollapsedActionIcon } from "#product/components/workspace/chat/tool-calls/CollapsedActionIcon";
 import { EditRows } from "#product/components/workspace/chat/tool-calls/CollapsedEditActionRows";
 import { GenericActionRow } from "#product/components/workspace/chat/tool-calls/CollapsedGenericActionRow";
 import {
+  ActionDisclosureRow,
   ActionFileLink,
   ActionRowIcon,
   PlainActionRow,
 } from "#product/components/workspace/chat/tool-calls/CollapsedActionRowPrimitives";
 
-export function CollapsedActionRows({ item }: { item: ToolCallItem }) {
+export function CollapsedActionRows({
+  item,
+  onOpenBackgroundTerminal,
+}: {
+  item: ToolCallItem;
+  /**
+   * bgwork r8 round 2/3: routed to the lone `command`-kind row, or to every
+   * row of a parsed/compound command — a background command is one process
+   * with one result text no matter how many structural rows it renders into.
+   */
+  onOpenBackgroundTerminal?: (processId: string) => void;
+}) {
   const parsedCommands = getToolCallParsedCommands(item);
+  // Resolved once for the whole tool call: a background command's "Command
+  // running in background with ID: ..." sentence lives in the ONE result
+  // text the item carries, regardless of how many display rows the harness's
+  // parsed_cmd breakdown produces below.
+  const { processId: backgroundProcessId, trailingStatus: backgroundTrailingStatus } =
+    useBackgroundCommandStatus(deriveCommandOutput(item));
+
   if (parsedCommands.length > 0) {
-    return <ParsedCommandRows item={item} commands={parsedCommands} />;
+    return (
+      <ParsedCommandRows
+        item={item}
+        commands={parsedCommands}
+        backgroundProcessId={backgroundProcessId}
+        backgroundTrailingStatus={backgroundTrailingStatus}
+        onOpenBackgroundTerminal={onOpenBackgroundTerminal}
+      />
+    );
   }
 
   switch (classifyCollapsedAction(item)) {
@@ -60,7 +89,12 @@ export function CollapsedActionRows({ item }: { item: ToolCallItem }) {
         />
       );
     case "command":
-      return <CommandActionRow item={item} />;
+      return (
+        <CommandActionRow
+          item={item}
+          onOpenBackgroundTerminal={onOpenBackgroundTerminal}
+        />
+      );
     case "edit":
       return <EditRows item={item} />;
     case "action":
@@ -151,10 +185,42 @@ function FileActionRow({
 function ParsedCommandRows({
   item,
   commands,
+  backgroundProcessId,
+  backgroundTrailingStatus,
+  onOpenBackgroundTerminal,
 }: {
   item: ToolCallItem;
   commands: ParsedToolCommand[];
+  backgroundProcessId: string | null;
+  backgroundTrailingStatus: string | undefined;
+  onOpenBackgroundTerminal?: (processId: string) => void;
 }) {
+  // A background command is one process regardless of how many structural
+  // rows its parsed breakdown renders into — every row opens the same
+  // terminal detail, uniformly (bgwork r8 round 3), never a mix of some rows
+  // opening the pane and others keeping their ordinary per-kind treatment
+  // (e.g. a "read"-kind row's file link). Falls back to the ordinary
+  // per-kind rendering below when there's no id to open with, or no caller
+  // wired to open it.
+  if (backgroundProcessId && onOpenBackgroundTerminal) {
+    const processId = backgroundProcessId;
+    return (
+      <>
+        {commands.map((command, idx) => (
+          <ActionDisclosureRow
+            key={`${item.itemId}-parsed-${idx}`}
+            label={formatParsedCommandLabel(item, command)}
+            icon={<CollapsedActionIcon kind={command.kind} />}
+            expanded={false}
+            failed={item.status === "failed"}
+            onToggle={() => onOpenBackgroundTerminal(processId)}
+            trailing={backgroundTrailingStatus}
+          />
+        ))}
+      </>
+    );
+  }
+
   return (
     <>
       {commands.map((command, idx) => command.kind === "read" && command.path
--- a/apps/packages/product-client/src/components/workspace/chat/tool-calls/CollapsedActionRows.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/tool-calls/CollapsedActionRows.tsx
@@ -18,6 +18,7 @@ import {
   formatSearchLabel,
 } from "#product/domain/chats/tools/collapsed-action-labels";
 import { useBackgroundCommandStatus } from "#product/hooks/activity/derived/use-background-command-status";
+import { useTranscriptSessionId } from "#product/components/workspace/chat/transcript/TranscriptContexts";
 import { CommandActionRow } from "#product/components/workspace/chat/tool-calls/CollapsedCommandActionRow";
 import { CollapsedActionIcon } from "#product/components/workspace/chat/tool-calls/CollapsedActionIcon";
 import { EditRows } from "#product/components/workspace/chat/tool-calls/CollapsedEditActionRows";
@@ -47,7 +48,7 @@ export function CollapsedActionRows({
   // text the item carries, regardless of how many display rows the harness's
   // parsed_cmd breakdown produces below.
   const { processId: backgroundProcessId, trailingStatus: backgroundTrailingStatus } =
-    useBackgroundCommandStatus(deriveCommandOutput(item));
+    useBackgroundCommandStatus(deriveCommandOutput(item), useTranscriptSessionId());
 
   if (parsedCommands.length > 0) {
     return (
--- a/apps/packages/product-client/src/components/workspace/chat/tool-calls/CollapsedActions.tsx
+++ b/apps/packages/product-client/src/components/workspace/chat/tool-calls/CollapsedActions.tsx
@@ -22,6 +22,8 @@ interface CollapsedActionsProps {
   autoFollow?: boolean;
   /** Keep the trailing exploration phase visually live between tool events. */
   liveContinuation?: boolean;
+  /** bgwork r8 round 2: routed to a background command's ledger row. */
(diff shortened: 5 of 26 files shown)


pablonyx:
## Review — PR #2022 (bgwork R8, dedicated reviewer of record)

**Verdict: CLEAN (approve).** No blocking findings. Three low-severity observations below, all inherent to the founder-ordered design or non-blocking polish. Reviewed adversarially against merge-base `d66ed0579`; I wrote none of this code.

### What I verified (ran) vs merely read

**Ran (literal outputs):**

- Four touched suites, in an installed worktree whose tree contains PR head `c90cdb3d6` **and** the advanced base `4f8f00a` merged together (touched source + test files byte-identical to PR head):
  ```
  ✓ src/domain/chats/tools/background-command-correlation.test.ts (6 tests)
  ✓ src/domain/activity/process.test.ts (15 tests)
  ✓ src/components/workspace/chat/transcript/MessageList.test.tsx (3 tests)
  ✓ src/components/workspace/chat/transcript/TranscriptToolCallItemBlock.background-terminal-clickin.test.tsx (5 tests)
  Test Files  4 passed (4)   Tests  29 passed (29)
  ```
- **Negative-control sabotage** — forced `parseBackgroundCommandProcessId` to `return null` always, re-ran the parser + click-in suites:
  ```
  Test Files  2 failed (2)   Tests  6 failed | 5 passed (11)
  ```
  The 6 failures are exactly the id-parsing / click-in / trailing-status assertions; the 5 survivors are the null/foreground/near-miss fallbacks and disclosure-toggling negative controls. This proves the tests genuinely exercise the parser, not a tautology. Parser restored (`git checkout`), confirmed clean.
- **Cross-head merge with 4f8f00a** (the footer-gate `TranscriptTurnRow` concern): isolated `git merge 4f8f00a` onto PR head in a scratch worktree is **conflict-free** (ort auto-merge). Merged `TranscriptTurnRow.tsx` carries BOTH changes — r8's `onOpenBackgroundTerminal` threading and 4f8f00a's `resolveTurnAssistantFooterModeForRow`. They touch disjoint regions (prop plumbing vs footer-mode resolution); no semantic conflict. 4f8f00a's `transcript-rendering.test.ts` also green (18/18) in the combined tree.
- **Gates:** `python3 scripts/check_max_lines.py` → `Max-lines check passed`. `python3 scripts/check_design_attribution.py` → `Design attribution check passed`. No `ratchets.toml` touched by the diff.
- **Design conformance (2d):** the trailing-status slot matches the Design Handoff artifact byte-for-byte. The chat transcript row mock is `<span class="ml-auto shrink-0 text-faint">running · 4m 12s</span>`, which is exactly what `ToolActionRowContent` renders from `processTrailingStatusLabel`. The `· pid NNNNN` suffix in the mock belongs only to the **roster-pane** rows, not the chat row — correctly omitted here.

**Read (not executed):** full `tsc` typecheck (CI covers; changes are additive optional-prop threading, low type risk); and the cross-system id-equality assumption (see Note 1).

### Observations (non-blocking)

**Note 1 (design assumption, not a code defect).** The correlation rests on the literal sentence `Command running in background with ID: {taskId}` and the claim that `{taskId}` equals the runtime's `ActivityProcess.id`. That sentence appears **nowhere in this repo** (including the Rust `anyharness` runtime) — it's emitted by the external coding-agent CLI, and the id-equality is asserted only in prose comments. This is the declared, founder-ordered design (no wire `tool_call_id` link exists), so it's acceptable — but flagging that the whole feature hinges on a cross-system contract that no test or CI check in this repo can catch if the upstream string or id shape drifts. Worth a runtime-fixture smoke test at some point.

**Note 2 (text-scrape false-positive).** `/Command running in background with ID:\s*(\S+)/` runs on every Bash result. A *foreground* command whose output literally contains that sentence (cat-ing a script, echoing a log, the agent quoting the message) would be mis-treated as a background row — click routes to the terminal detail with an empty trailing slot. Low probability, inherent to client-side text correlation, acknowledged in the code comment. Minor: `(\S+)` only strips trailing `.,;:`, so an id wrapped like `ID: proc-9)` keeps the `)`.

**Note 3 (render cost + a11y).** `useSessionActivityForSession(transcriptSessionId)` is now subscribed by `TranscriptToolCallItemBlock` for **every** tool row — including non-Bash Read/Edit rows that never use the roster. The selector is a cheap memoized `useShallow` pick, but it's O(N tool rows) shallow-compares per activity-store update on long, actively-streaming transcripts. Also: an overridden row renders `role="button"` with a permanent `aria-expanded="false"`, which semantically implies collapsible inline content that never expands; a side-panel-opening button would read cleaner without `aria-expanded` (or with `aria-haspopup`). Both are polish, not correctness.

### Correctness spot-checks that passed
- `onOpen` override: `expanded` forced false, `hasDetails` true so the row stays a focusable button; keyboard `activate()` shares the click path (Enter/Space route to `onOpen`); inline disclosure never renders for overridden rows. Non-Bash tools untouched (`onOpen` only wired into the two `BashCommandCall` sites).
- Roster-absent (post-reload) path: parsed id with no roster match still opens the terminal (`backgroundProcessId && onOpenBackgroundTerminal`), trailing slot empty — matches the test and the code comment.
- `processTrailingStatusLabel` covers the full `ProcessStatus` union (`running` | `exited`, `exitCode` nullable); minutes/seconds rounding matches the mock forms including whole-minute (`5m`) and sub-minute (`30s`).
- Result-text array handling: `contentParts` joined with `\n\n` before parse; `\S+` stops at whitespace so cross-part boundaries don't leak into the id.

Note: the installed worktree used for the run carried unrelated pre-existing uncommitted WIP on `TranscriptTurnRow.tsx`/`transcript-rendering.ts`; the four R8 suites import neither (the click-in test targets `TranscriptToolCallItemBlock`; the MessageList test mocks `renderTurnRow`), so that WIP has no bearing on the results above. Scratch worktree removed.

pablonyx:
## Delta review — Round 2 (`c90cdb3d6..0ddafdb8c`, 7 files)

**Verdict: CLEAN (delta) — approve.** Round 2 correctly extends click-in + trailing status to the collapsed "Worked for Ns" ledger path (`CommandActionRow`), which is the path a founder actually clicks after a turn settles. One LOW-severity completeness gap noted below (non-blocking, not a regression). I wrote none of this code; reviewed the range adversarially.

### Verified by running (scratch worktree at `0ddafdb8c`, real deps)

- **New suite:** `CollapsedCommandActionRow.background-terminal-clickin.test.tsx` → `7 passed (7)`.
- **Pre-existing collapsed suites (the very mounts that forced the prop approach):** `CollapsedActions.test.tsx (16)` + `CollapsedActionRows.test.tsx (6)` + `ScopedTranscriptBlocks.test.tsx (1)` → `23 passed (23)`.
- **Sabotage negative-control** (forced `parseBackgroundCommandProcessId` → `null`, re-ran new suite):
  ```
  Tests  4 failed | 3 passed (7)
  ```
  The 4 failures are exactly the open/trailing-status assertions; the 3 survivors are the unwired-caller, foreground, and near-miss disclosure fallbacks. Parser restored, confirmed clean.
- **Gates at `0ddafdb8c`:** `check_max_lines.py` → `Max-lines check passed`; `check_design_attribution.py` → `Design attribution check passed`. No `ratchets.toml` in the delta.
- **Cross-head with advanced r6 base `1fb98411e`** (both touch transcript components): `git merge 1fb98411e` onto round-2 head is **conflict-free** (ort). Merged `TranscriptTurnRow.tsx` carries both r8's `onOpenBackgroundTerminal` threading and `1fb98411e`'s `resolveTurnAssistantFooterModeForRow`. Both PRs' touched suites together in the merged tree → `9 files, 78 passed (78)`.

### Focus-point findings

**1. Prop threading — all callers accounted for.**
- Main chain wired: `MessageList` → `TurnItemSequence` → `ScopedTranscriptBlocks`/`TurnDisplayBlockNode` → `CollapsedActions` → `CollapsedActionRows` → `CommandActionRow`. ✓
- `PlaygroundToolTranscript` (the two standalone-playground `CollapsedActions` mounts) intentionally **not** wired — verified the negative-control test (no `onOpenBackgroundTerminal` prop → row keeps `aria-expanded` toggling and reveals inline output) passes. ✓
- **GAP (LOW, see below):** `TranscriptAgentGroupBlock` renders `ScopedTranscriptBlocks` without the callback and doesn't accept the prop; `TranscriptToolCallGroupBlock` doesn't forward it there either.

**2. Mixed pattern (roster via local hooks, open via prop) — safe, verified empirically.** `CommandActionRow` reads roster via `useTranscriptSessionId` + `useSessionActivityForSession` (both provider-free: context defaults to `null`, the store selector returns `EMPTY_ACTIVITY` on a missing slot) and takes the pane-opener as a prop. The reason given (`useOpenBackgroundTerminalDetail` pulls `useWorkspaces()`/`useProductAuthUserId()`, needing a full `ProductHostProvider`) checks out: the pre-existing `CollapsedActions.test.tsx` mounts a minimal host with **no** `TranscriptContextProviders` and **does not mock** `use-session-activity`, yet all 16 tests pass — so the two local hooks genuinely lack the dependency that broke the local-callback approach.

**3. `ParsedCommandRows` — clean exclusion, no half-wired state.** When `getToolCallParsedCommands(item).length > 0`, `CollapsedActionRows` returns `ParsedCommandRows`, which renders static `PlainActionRow`/`FileActionRow` (no disclosure, no toggle, no callback) — the callback is only threaded to the single `case "command"` row. A backgrounded *compound* command that parses into multiple rows therefore shows static rows with neither click-in nor inline output (unchanged pre-existing behavior), not a broken half-open state. Acceptable; worth knowing that compound-parsed background commands don't get click-in.

### Observation (LOW, non-blocking) — nested-subagent collapsed ledger is the same unfixed shape one level down

`TranscriptAgentGroupBlock` (a native subagent's nested transcript) renders its own `ScopedTranscriptBlocks` **without** `onOpenBackgroundTerminal`. Result: inside a subagent's turn, an *expanded* background tool-call row still gets click-in (via `renderChild`, from round 1), but its *collapsed* ledger row does **not** — precisely the collapsed-path gap round 2 just fixed at the top level, still present for background commands executed inside a native subagent. It degrades cleanly to inline disclosure (no crash, no half-wired state), and the founder-reported case (top-level turn ledger) is fully fixed, so this is a completeness note rather than a blocker. Wiring it would mean threading the callback through `TranscriptToolCallGroupBlock` → `TranscriptAgentGroupBlock` → its `ScopedTranscriptBlocks`.

Round-1 observations (text-scrape false-positive risk, per-row roster subscription cost, the unverified cross-system `{taskId} == ActivityProcess.id` assumption) carry forward unchanged and remain non-blocking. Minor test nit: the case named "…standalone playground hasn't wired roster data at all" (line 178) actually *does* pass `onOpenBackgroundTerminal`; the real unwired-playground control is the next case (line 198). Scratch worktree removed.

pablonyx:
## Delta review — Round 3 (`0ddafdb8c..cd4945f63`, 9 files)

**Verdict: CLEAN (delta) — approve.** Round 3 correctly resolves both findings my round-2 review surfaced, and the shared-hook refactor is behavior-preserving. One trivial redundancy noted (non-blocking). Reviewed adversarially against my own prior grading; I wrote none of this code.

### Both round-2 findings resolved

**Finding 1 — the `ParsedCommandRows` exclusion I wrongly graded "acceptable" was the live bug.** The founder's own `for i in $(seq 1 15); ...` renders through `getToolCallParsedCommands` → `ParsedCommandRows`, which `CollapsedActionRows` checks *before* the `classifyCollapsedAction` "command" switch — so round 2's `CommandActionRow` fix never reached it. Round 3 resolves the id/status **once per tool call** in `CollapsedActionRows` (`useBackgroundCommandStatus(deriveCommandOutput(item))`) and threads it into `ParsedCommandRows`, where every parsed row renders as an `ActionDisclosureRow` whose `onToggle` opens the same `processId`. Verified: one id for the whole call (resolved at the parent, passed as a prop — no per-row re-parse), and the new suite's multi-command test asserts both rows fire `onOpenBackgroundTerminal("bn30h453a")` (`toHaveBeenNthCalledWith` 1 and 2, called twice) — no half-wired subset. `deriveCommandOutput` is the same parse input `CommandActionRow` uses, so the two call sites can't diverge.

**Finding 2 — nested-subagent collapsed ledger now wired.** `TranscriptToolCallGroupBlock` forwards `onOpenBackgroundTerminal` to `TranscriptAgentGroupBlock`, which threads it into its `renderScopedWork` → `ScopedTranscriptBlocks` (the same path `onOpenSubagent` already establishes). New `TranscriptAgentGroupBlock.background-terminal-clickin.test.tsx` (2 tests) drives the full nested chain (expand subagent → tool-call ledger → inner row → click → `toHaveBeenCalledWith("bn30h453a")`) plus a real negative control (no callback → `aria-expanded="true"`, inline sentence shown). Split rationale checks out: the parent `TranscriptAgentGroupBlock.test.tsx` is 515 lines and untouched by this delta — the 2 tests were placed in the sibling to avoid pushing it past 600; both new files keep the load-bearing `.test.` mid-name segment. Assertions are full-strength, not weakened.

### Verified by running (scratch worktree at `cd4945f63`, real deps)

- New + regression suites together: `parsed-…(7)`, `TranscriptAgentGroupBlock.background-…(2)`, `CollapsedCommandActionRow.background-…(6)`, `TranscriptToolCallItemBlock.background-…(5)`, `MessageList(3)`, `CollapsedActions(16)`, `CollapsedActionRows(6)`, `TranscriptAgentGroupBlock(16)`, `background-command-correlation(6)`, `process(15)` → **82 passed (82)**. The round-1/2 suites pass **unchanged**, confirming the shared-hook refactor is behavior-preserving.
- **Sabotage** (parser → `null`, new parsed suite): `Tests 3 failed | 4 passed (7)` — the 3 click-in/trailing assertions fail; the reproduction check + 3 negative controls survive. Restored, confirmed clean.
- **Refactor safety (concern 3):** `TranscriptToolCallItemBlock.tsx` is now **392 lines** (was 402) — no longer an FE-SIZE-1 entry — and it hoists the result-text derivation + hook above the early returns so the hook stays unconditional (rules-of-hooks correct). The shared `useBackgroundCommandStatus` has **no `ProductHostProvider` dependency**: it composes only `useTranscriptSessionId` (context, defaults `null`) + `useSessionActivityForSession` (zustand store) + two pure functions. Empirically confirmed — `CollapsedActions.test.tsx` (16 tests) mounts a minimal host with no `TranscriptContextProviders` and does **not** mock `use-session-activity`, yet now routes through `CollapsedActionRows`' unconditional hook call and passes.
- **Concern 4 (removed duplicate):** the deleted `"…standalone playground hasn't wired roster data at all"` test was genuinely dishonest — its body passed `onOpenBackgroundTerminal` and only asserted it fired (a weaker duplicate of the first test), contradicting its own title. Its removal loses no coverage: the real no-callback path is still covered by the retained `"negative control: an unwired caller (no onOpenBackgroundTerminal prop) keeps the ordinary inline disclosure"` in the same file and by the new parsed suite's unwired control (asserts the row is a non-interactive `PlainActionRow` with no button role).

### Gates at `cd4945f63`
```
check_max_lines.py         → Max-lines check passed
check_design_attribution.py → Design attribution check passed
report_frontend_structure.py --strict --summary-only → FE-SIZE-1: 1  TOTAL: 1  (exit 1)
```
The lone FE-SIZE-1 entry is the pre-existing `MessageList.tsx` (442 lines) — not touched or worsened by this round; the round-3 branch predates r6's ratchet update. No `ratchets.toml` in the delta.

### Cross-head with current r6 head `79127061f`
`git merge 79127061f` onto round-3 head is **conflict-free** (ort; `TranscriptTurnRow.tsx` auto-merges, `ratchets.toml` +5). On the merged tree the gate is clean: **FE-SIZE-1: 0, TOTAL: 0, exit 0** — the MessageList entry above is purely the branch predating the r6 ratchet. Both PRs' touched suites together in the merged tree → **11 files, 101 passed (101)**.

### Observation (trivial, non-blocking)
`CollapsedActionRows` calls `useBackgroundCommandStatus(deriveCommandOutput(item))` at the top for **every** tool call, but the result is consumed only by the `ParsedCommandRows` branch — a `command`-kind row discards it and `CommandActionRow` resolves the same status a second time internally. So a single non-parsed command row does two roster lookups/parses per render, and `deriveCommandOutput` runs for non-command rows (edits/reads) too. Pure functions, identical inputs → no correctness impact; could be tightened by passing the parent's result into `CommandActionRow` (or only computing it in the parsed branch). Round-1 standing observations (external-string correlation, per-row roster subscription) carry forward unchanged. Scratch worktree removed.

pablonyx:
## Net-delta refold onto the refolded r6 — receipts (new head 408993041)

#2021's branch was rebuilt as a net-delta refold onto main (see its receipt); this PR stacked on the old shape, so it got the same construction: spine `bce0477e4..8d68bc6bc` (r6's reviewed tip → this PR's reviewed tip) applied 3-way onto the refolded r6 head `968a765d1`. Zero apply conflicts.

**Receipts:**
- **22 of 25 files byte-identical** to the reviewed tip `8d68bc6bc` (programmatic check).
- The other 3 (`MessageList.tsx`, `TranscriptAgentGroupBlock.tsx`, `TranscriptToolCallGroupBlock.tsx`) were also moved by main since r4b; each verified **delta-equivalent line-for-line**: the refold-on-r6 delta equals the reviewed r8 delta exactly (offsets aside).
- All 11 new `#product` import specifiers resolve in the merged tree; zero references to the r6 band machinery the refold deleted; main's only prop movement on the interaction set is an **optional** `workspaceId` on `TranscriptAgentGroupBlock` (no required-prop seams).
- **One disclosed hand edit:** MessageList ratchet re-true 442 → 443. Honest arithmetic: reviewed base was 439+3=442 (at cap); the refolded base carries main's one-line hydration amendment (440), so identical reviewed content lands at 443. No new content beyond the reviewed delta.

Because every content delta is provably equivalence-carried, this is an **all-equals-class carry** — the existing review of record carries; the single ratchet line is the only new material and is disclosed here. Fresh CI is baking on 408993041 (base: `bgwork/r6-completion-receipts`, MERGEABLE — no re-fire dance needed). This PR merges only after #2021 resolves, on the founder's word.

pablonyx:
## Base-interaction fix (refold onto rebuilt base) — head `cce163845`

The refold carried r8's files intact, but the **new base** moved two things underneath the reviewed tip. Both reds are diagnosed and fixed forward with a minimal diff. Neither is an r8 feature regression.

### 1. `Test product client` — root cause: **test-environment only**, feature is correct
The new base's `AnimatedCollapsibleContent` (Chat Scroll transitions) now gates `visuallyExpanded` behind one `requestAnimationFrame` after the toggle commits, so the freshly-mounted ledger subtree stays `inert` / `aria-hidden` until that frame runs. r8's suites expand a collapsed ledger then **synchronously** query the inner command row; under jsdom (no real frame between a sync `fireEvent.click` and the next query) the row was present but accessibility-hidden, so every `getByRole` after the expand failed with the empty `"Running command"` shell. In the browser the frame fires ~16ms later and the row (labeled by the command text) appears and click-in works — the feature is intact.

Fix is in the **test setup**: drive the frame inline with a synchronous `requestAnimationFrame` spy — the exact convention `CollapsedActions.test.tsx` already uses. Every assertion (command-named button, `aria-expanded`, `onOpenBackgroundTerminal` called with the process id, negative controls) is unchanged. The roster trailing-status test fakes only `Date` (`toFake: ["Date"]`) so the spy survives.

Literal local vitest green for all three suites:

```
 ✓ src/components/workspace/chat/tool-calls/CollapsedActionRows.parsed-background-terminal-clickin.test.tsx (7 tests) 414ms
 ✓ src/components/workspace/chat/tool-calls/CollapsedCommandActionRow.background-terminal-clickin.test.tsx (6 tests) 448ms
 ✓ src/components/workspace/chat/transcript/TranscriptAgentGroupBlock.background-terminal-clickin.test.tsx (2 tests) 465ms

 Test Files  3 passed (3)
      Tests  15 passed (15)
```

### 2. `Check frontend layer boundaries` FE-PC-6 — cured
Main's checker advanced to reject the upward `hooks -> components` edge where `use-background-command-status.ts` read the transcript session id from `useTranscriptSessionId` (components layer). Threaded the session id in as a parameter from each component-layer caller (`CollapsedCommandActionRow`, `CollapsedActionRows`, `TranscriptToolCallItemBlock`) instead — the hook now stays within the hooks layer. `scripts/check_frontend_boundaries.py` → **Frontend boundary check passed.** `tsc -p tsconfig.json --noEmit` → **0 errors.**

### Reds NOT touched (not mine)
- Transcript scroll physics webkit `prepend anchoring` — known recurring flake; re-runs on this push.
- `Login first-load budget` (502022 > 502000) — founder-gated (headroom PR #2025).
- cargo sessions timing test — known flake class.

### Per-file disclosure
- `hooks/activity/derived/use-background-command-status.ts` — drop the `useTranscriptSessionId` (components) import; take `sessionId: string | null` as a param. Removes the FE-PC-6 upward edge.
- `tool-calls/CollapsedCommandActionRow.tsx` — call `useTranscriptSessionId()` locally and pass it to the hook.
- `tool-calls/CollapsedActionRows.tsx` — same.
- `transcript/TranscriptToolCallItemBlock.tsx` — same.
- 3 `*.background-terminal-clickin.test.tsx` suites — add the synchronous `requestAnimationFrame` spy (beforeEach) + restore (afterEach); scope the roster-status test's fake timers to `Date` only. No expectation rewrites.

pablonyx:
## Delta review — #2022 cure slice at cce163845

Independent review of the single cure commit, pinned to head `cce1638455fed0ae891a5573299d5e3339cfa977` (verified first-hand: exactly 1 commit FF of `408993041`, 7 files). Adjudicated against the r8 base `408993041` and `origin/main` (`4cd69d806`, r7 — no r8).

**Verdict: CLEAN — no r8-blocking finding. The two reds are base-interaction, not r8 regressions; the fix is behavior-preserving and assertion-neutral. One transparency caveat on the empirical negative control (d), which does not change the verdict.**

### (a) Test diffs are genuinely setup-only — zero assertion weakening ✓
Diffed every hunk in all three suites. The only changes are, per file: a `beforeEach` that installs a synchronous `requestAnimationFrame` spy (`callback(0); return 1;`), a `vi.restoreAllMocks()` in `afterEach`, and — in the two roster-status tests — narrowing `vi.useFakeTimers()` to `vi.useFakeTimers({ toFake: ["Date"] })` so the spy survives fake timers (the third suite additionally imports `beforeEach`). **No `expect(...)` line is added, removed, or altered anywhere** — command-named button, `aria-expanded`, `onOpenBackgroundTerminal` called with the process id, and the negative controls are byte-for-byte the reviewed assertions. The `toFake: ["Date"]` scoping is not a weakening: roster elapsed time is `Date.now()`-derived and `Date` remains faked deterministically (`setSystemTime` unchanged).

### (b) Root-cause story holds — base behavior, NOT an r8 seam (explicit adjudication) ✓
`apps/packages/product-client/src/primitives/AnimatedCollapsibleContent.tsx` on the base computes `visuallyExpanded = expanded && preparedExpanded` (`:19`); `preparedExpanded` flips true only inside a `window.requestAnimationFrame` scheduled in a `useEffect` (`:32`). While false, the subtree renders `aria-hidden={!visuallyExpanded}` (`:45`) **and** `inert={!visuallyExpanded}` (`:48`). So after a synchronous `fireEvent.click` expands a freshly-mounted ledger, the inner command row is in the DOM but accessibility-hidden until the next frame — and jsdom never advances a frame between the sync click and the sync `getByRole`. The suites' `expandLedgerAndGetInnerRow` (`CollapsedCommandActionRow…test.tsx:96`) does exactly `fireEvent.click(getByRole(button, /Running command/)); return getByRole(button, innerName)`, so the inner (command-text-labeled) button is unreachable and only the outer `"Running command"` shell is accessible — precisely the reported empty-shell signature.

**Adjudication:** this rAF-gated disclosure is a shipped Chat Scroll transition, not an r8 change. `AnimatedCollapsibleContent.tsx` is **byte-identical between `origin/main` (r7, contains no r8) and the r8 base `408993041`**, and is untouched by this cure. r8 neither introduced nor altered the gating; the feature works in a real browser (frame fires ~16 ms later, row appears, click-in fires `onOpenBackgroundTerminal`). This is a **test-environment artifact, not a feature break.** The fix follows a pre-existing convention: the base's own `CollapsedActions.test.tsx:71` (untouched here) uses the identical `vi.spyOn(window, "requestAnimationFrame")` spy against the same primitive.

### (c) sessionId threading is behavior-preserving ✓
`use-background-command-status.ts` drops the `useTranscriptSessionId` (components-layer) import and takes `sessionId: string | null` as a param (`:28`), forwarding it unchanged to `useSessionActivityForSession(sessionId)` — identical value, identical downstream lookup. All three (and only three) callers are updated to pass `useTranscriptSessionId()`: `CollapsedCommandActionRow.tsx:43`, `CollapsedActionRows.tsx:51`, `TranscriptToolCallItemBlock.tsx:71`. Grep at head confirms **no caller left on the 1-arg path** and **no other caller exists**. `useTranscriptSessionId()` is called unconditionally at each component's top level (inline as an argument in two of them), so the same hook runs in the same order with the same value — Rules-of-Hooks compliant, no new conditional/looped hook, no new layer edge (the hook retains **zero** `#product/components` imports; FE-PC-6 upward edge removed). Purely a relocation of where the context read happens, one layer down-stack.

### (d) Negative control — NOT independently executed (machine-safety); outcome is deterministic from source
I did **not** run vitest. My review worktree has no `node_modules`, and installing the monorepo + running the suite is unsafe right now: swap is 90% full (~1.1 GB free) at VM pressure level 2 with ~58 MB free RAM, and the machine's hard limits forbid risking an OOM (they override convenience). The sibling deps-provisioned worktree is on a different head and in active use, so I did not disturb it. I am flagging this rather than papering over it.

**[Count corrected — the split is 10 fail / 5 pass, not 11 / 4. The "11" here was the tasking brief's approximate figure; source classification (per-test breakdown in the addendum comment below) yields exactly 10. Text amended accordingly.]**

The outcome is nonetheless deterministic from the verified source, and I state it as a mechanism, not a guess: with the spy **removed**, jsdom's real `requestAnimationFrame` never fires between the synchronous click and the synchronous `getByRole`, so `visuallyExpanded` stays false, the inner subtree stays `aria-hidden`/`inert`, and every test that issues a positive `getByRole` on a gated inner command button fails with `Unable to find … button … /<command>/` while only the `"Running command"` shell resolves — **the 10 failures**. With the spy **restored**, the effect's rAF callback runs synchronously → `setPreparedExpanded(true)` → re-render → `visuallyExpanded` true → subtree revealed → all 15 resolve (**the 5 that pass either way never issue a positive `getByRole` on a gated inner row — the reproduction and roster-status tests assert via `getByText`, the per-kind negative controls via `queryByRole(...).toBeNull()` — so `aria-hidden` never bites them; note they still expand**). The implementer's receipt reports `Test Files 3 passed (3) / Tests 15 passed (15)`; I did not independently reproduce that literal. **Recommendation:** capture the empirical revert→**10-fail** / restore→15-pass literals in a properly provisioned headroom env (not implementer-controlled) for the record; the static verdict does not depend on it.

### Reds explicitly out of this slice (spot-checked, agree)
Transcript-scroll webkit `prepend anchoring` (known flake), `Login first-load budget` 502022>502000 (founder-gated, headroom PR #2025), cargo sessions timing (flake class) — none are this cure's concern.

Comment-only; I do not approve. The cure carries r8's files intact and equivalence holds: **CLEAN**, modulo the (d) empirical run being captured elsewhere.

pablonyx:
## Addendum — count correction (10, not 11) independently verified; negative control still pending machine headroom

Following the count correction, I independently classified all 15 tests from source at `cce163845` (deriving the split myself, not accepting the implementer's receipt). The mechanism check the orchestrator asked for — *exactly* the expand-then-query tests fail and the non-expanding ones don't — holds precisely.

**Why the split is 10 fail / 5 pass:** `getByRole("button", { name })` respects the accessibility tree, so an `aria-hidden`/`inert` subtree excludes the inner command button; `getByText(...)` and `queryByRole(...).toBeNull()` do not positively require an accessible inner button, so the gating never bites them.

- **Suite A (`CollapsedActionRows.parsed`, 7):** PASS in the broken state — `:127` reproduction (`getByText(FOR_LOOP)`), `:167` roster status (`getByText("running · 4m 12s")`), `:225`/`:242`/`:270` per-kind negative controls (`queryByRole(...).toBeNull()` + `getByText`). FAIL — `:143`, `:191` (positive `getByRole` on the inner command button). → **5 pass, 2 fail.**
- **Suite B (`CollapsedCommandActionRow`, 6):** every test routes through `expandLedgerAndGetInnerRow(/Running: for i in/i)` (`:96`), a positive inner `getByRole` — including the three negative controls, which fail at the helper before their own assertion. → **6 fail.**
- **Suite C (`TranscriptAgentGroupBlock`, 2):** both do `getByRole("button", { name: /sleep 100 &/ })` after expand. → **2 fail.**

Total **10 fail / 5 pass**, matching the corrected count. The 5 survivors are exactly the tests that never issue a positive `getByRole` against a gated inner row (the implementer's "never reach an inner command row" phrasing is right; note `:167` survives via `getByText`, not because it skips expansion). This also matches the reported error dump — after `expandLedger()` the *outer* toggle carries `aria-expanded="true"`/`data-active="true"` while the inner rows stay hidden, so the failing `getByRole` prints only the `"Running command"` shell.

**Negative control (question d) — still NOT executed by me; machine-safety, not acceptance of the implementer's run.** My review worktree has no `node_modules`, and the machine remains memory-saturated (VM pressure level 2, swap 93% full ~0.8 GB free, compressor ~11 GB); the standing hard limits forbid risking an OOM/swap-death for a monorepo install + vitest, and they override convenience. I did not disturb the sibling deps-worktree (different head, active use). The static derivation above reaches the same 10/5 result the implementer's run reports, so their literals cross-check — but per doctrine I am flagging that I have not *executed* the revert→10-fail / restore→15-pass control myself. **I will run it independently the moment the machine has headroom** (pressure back to level 1 / swap freed) and post the raw vitest output; the verdict does not otherwise depend on it.

Verdict unchanged: **CLEAN**, no r8-blocking finding.

pablonyx:
STATE RESTORE RECEIPT (orchestrator): when r6 #2021 squash-merged (e321fb51d) and its branch deleted, GitHub CLOSED this stacked PR instead of auto-retargeting. Cure executed 06:5xZ: base branch temporarily restored at its merged tip 968a765d1 → PR reopened → base retargeted to main → scaffold branch deleted again. Current CONFLICTING state is the expected stacked-squash artifact (branch carries r6's original commits vs main's squash); the coming refold onto post-r6 main cures it. All review receipts on this PR remain the trail of record. Refold order goes out after CS r12 lands (ratchets.toml overlap — avoiding a third interleave).

pablonyx:
## Executed negative control — discharges the deferred (d) gate on the cure at `cce163845`

Ran the independent revert-control myself (not accepted as a receipt) after VM pressure returned to level 1. Isolated env: a fresh detached `git worktree` at the reviewed cure head `cce1638455fed0ae891a5573299d5e3339cfa977` — **not** the implementer's worktree. Vitest only; no cargo/rustc/Docker. Environment build was JS-only: `pnpm install` from the warm store (`--ignore-scripts`) + `tsc` builds of the three non-source-aliased frontend workspace deps (`@proliferate/design`, `@proliferate/cloud-sdk`, `@proliferate/cloud-sdk-react`); `@anyharness/sdk[-react]` resolve to source per the package's vitest config, and the lightweight global-setup emits the gitignored agent catalog. Pressure held at level 1–2 throughout (never level 3).

### 1. Baseline at cure head — 15/15 green
```
 ✓ src/components/workspace/chat/tool-calls/CollapsedActionRows.parsed-background-terminal-clickin.test.tsx (7 tests) 210ms
 ✓ src/components/workspace/chat/tool-calls/CollapsedCommandActionRow.background-terminal-clickin.test.tsx (6 tests) 220ms
 ✓ src/components/workspace/chat/transcript/TranscriptAgentGroupBlock.background-terminal-clickin.test.tsx (2 tests) 117ms

 Test Files  3 passed (3)
      Tests  15 passed (15)
```

### 2. Spy neutralized (the cure's test-env half only) — 10 fail / 5 pass, empty-shell signature
Reverted **only** the three test files to their pre-cure base (`408993041`), removing the synchronous `requestAnimationFrame` spy; the four cure **source** files (sessionId threading) stayed at `cce163845`.
```
 Test Files  3 failed (3)
      Tests  10 failed | 5 passed (15)
```
Representative failure — the exact empty-shell signature:
```
TestingLibraryElementError: Unable to find an accessible element with the role "button" and name `/for i in \$\(seq 1 15\); do echo probe \$i; sleep 1; done/`

Here are the accessible roles:
  button:
  Name "Running command":
  <button aria-expanded="true" … data-active="true" data-chat-transcript-ignore="true" type="button" />
```
The **10 failures** are exactly the tests that issue a positive `getByRole` on a gated inner command button — Suite A `opens … for-loop shape` + `every parsed row …`; **all 6** of Suite B (including its three "inline disclosure / disclosure toggling" negative controls, which route through `expandLedgerAndGetInnerRow`); **both** of Suite C. The **5 survivors** are exactly Suite A's non-positive-role tests: the reproduction (`getByText`), the roster-status (`getByText`), and the three "ordinary per-kind rendering" negative controls (`queryByRole(...).toBeNull()` + `getByText`). This matches, empirically and test-for-test, the source classification in [the earlier addendum](https://github.com/proliferate-ai/proliferate/pull/2022#issuecomment-5323951140) and the implementer's reported literal.

### 3. Spy restored — 15/15 green
```
 Test Files  3 passed (3)
      Tests  15 passed (15)
```

**Conclusion:** the control is discharged by independent execution. Removing the cure's test-env spy reproduces exactly 10 failures with the `"Running command"`/`aria-expanded="true"`/`data-active="true"` empty-shell signature; restoring it returns 15/15 — confirming the reds were the jsdom-vs-`AnimatedCollapsibleContent`-rAF test-environment gap (not a feature regression) and that the fix is a pure, assertion-neutral test-setup change. Verdict on the cure stands: **CLEAN** (ref [5323921714](https://github.com/proliferate-ai/proliferate/pull/2022#issuecomment-5323921714)). Run at head `cce163845`; a subsequent refold that alters the cure files more than byte-identically would warrant a re-run.

pablonyx:
## Refold onto current main — receipt

**old head `cce163845` → new head `c29fd1a3f`** · base `main` @ `da7b874c4` (includes bg r6 `e321fb51d`, CS r12 `bc986d7fc`, workflows-v2 #2040).

Mechanic: `git rebase --onto origin/main 968a765d1 cce163845` (r8's two own commits off the old r6 tip). The prior CONFLICTING state was the expected stacked-squash artifact after r6 squash-merged into main.

**range-diff (old base..old head) vs (new base..new head):**
```
1:  408993041 = 1:  2ffb6e44a feat(product): background command tool-call opens terminal detail (bgwork R8)
2:  cce163845 = 2:  c29fd1a3f fix(product): adapt bgwork r8 to the refolded base (rAF disclosure + FE-PC-6)
```
Both commits `=` — the r8-own delta carried **byte-identical**. The cure files (the three `*.background-terminal-clickin.test.tsx` rAF-flush setups and the `use-background-command-status` session-id threading) are byte-identical across the refold.

**Known base-side compositions (from main, not from my commits):**
- `CollapsedCommandActionRow.tsx` — main's `chainVerticalWheel` prop on `AutoHideScrollArea` is now in the base; my delta doesn't touch those lines.
- `MessageList.tsx` — main's one-line hydration amendment is in the base; the r8 ratchet reason already anticipated this (`439→440`, reviewed-at-442 lands at 443).

**Ratchet:** `lints/frontend/ratchets.toml` MessageList `max_lines` = **443**, unchanged from the reviewed value; actual `MessageList.tsx` = **443** lines. No new value — no re-ruling needed.

**Green on the new base (nice'd local vitest, 9 suites incl. all 3 cure suites + neighbors + domain):**
```
 Test Files  9 passed (9)
      Tests  83 passed (83)
```
Also: `check_frontend_boundaries.py` → passed · `check_max_lines.py` → passed.

Left as-is per instruction: **not** flipping draft→ready; fresh CI bakes on `c29fd1a3f`; merge word is Pablo's (behind r6).

pablonyx:
BOARD ATTRIBUTION at c29fd1a3f (run 32112594755, settled 07:52:39Z): 12 of 13 CI jobs success — incl. budget, tier-2 lifecycle, cargo. Sole red: Transcript scroll physics, signature verbatim '✘ 27 [webkit] › qualification/scroll-physics/specs/scroll-physics.spec.ts:697:3 › prepend anchoring: older-history prepend keeps the reading row fixed' (chromium green) = the known flagged webkit prepend strand (founder-ruled rides-flagged class; post-train investigation item on main). Consuming the ONE sanctioned same-ref retry now (merge-ref base da7b874c4 still current main, so the snapshot is fresh). If it greens: board of record = green. If it re-fails: attribution stands per the standing flagged ruling, disposition to the founder.

pablonyx:
RETRY OUTCOME + FINAL BOARD OF RECORD at c29fd1a3f: the single sanctioned same-ref retry re-failed with the IDENTICAL signature ('✘ 27 [webkit] › scroll-physics.spec.ts:697:3 › prepend anchoring: older-history prepend keeps the reading row fixed', chromium green both attempts, settled 07:58:51Z). Per the founder's standing flagged-strand ruling: retry consumed, red rides ATTRIBUTED, no further retries. FINAL BOARD: 12/13 CI jobs success (budget, tier-2, cargo all green) + 1 attributed known-strand physics red; Smoke/CodeQL/Intent/Metadata runs all success. Historical note: this strand failed r4's retry the same way during the CS train and rode into its merge attributed; the strand investigation on main is a named post-train follow-up. #2022 status: merge-material-modulo-founder (both prior review verdicts CLEAN incl. the executed empirical control; refold byte-identical; this attribution disclosed). Merge word is the founder's.

pablonyx:
MERGING on the founder's class-scoped approval ('if it's been reviewed I fully approve', direct turn ~08:1xZ; this PR: review of record CLEAN + delta CLEAN + executed empirical control + byte-identical refold receipt). Board of record: comment 5325311481 (12/13 green + the founder-ruled attributed webkit strand red, retry consumed). Composition disclosure: #2035 (anyharness Rust, zero file overlap with this PR's set) merged 07:55:34Z after this board settled — zero-overlap precedent per the r11 note. Squash pinned to c29fd1a3f.

PR #1993, opened 2026-08-16
feat(product-client): scroll diagnostics via renderer pipe (PRO-187, r11, reopened)

pablonyx:
Reopens rung 11 of the Chat Scroll ladder (PRO-187). The prior PR #1984 was irreversibly auto-marked MERGED by GitHub when a tooling agent briefly pushed this branch's commits onto the base ref (immediately force-restored; verified nothing reached main: r11 head is not an ancestor of origin/main). Same branch, same content, same head as #1984's final state.

Scope (unchanged from #1984): observation-only scroll diagnostics through the existing renderer-diagnostics-port pipe.
- recordTranscriptPinTransition with an 8-variant cause taxonomy, recordTranscriptUserScrollIntent (src/lib/infra/diagnostics/renderer-diagnostic-migrations.ts)
- setPinned/notifyUserScrollIntent instrumented in use-transcript-stick-to-bottom.ts with an optional cause param threaded through button-click, submit-repin, and restore call sites
- blank-viewport fallback diagnostic un-gated from import.meta.env.DEV so it reaches production telemetry (Q8)

History: #1984 was CI-green (18/18) and adversarially reviewed CLEAN at head 11cfa279a (review: #1984 issuecomment-5307172323, manual verification: issuecomment-5306898383); current head ee976c831 is that content rebased onto the r10 Q13 assertion fix (7e627c22f), which asserts the unexplained scrollTop drop (drop minus scrollHeight shrink) from frame-aligned traces.

Known lane flake under investigation: [webkit] prepend anchoring (Expected > 150 / Received 0) recurs bimodally across the stack on slow runners; evidence archived, tracked separately.

Not to be merged without Pablo's review (standing rule; merges bottom-up on #1722).

PR #2014, opened 2026-08-17
fix(product): workflow-lane merge fallout — main tsc break + line-clamp previews

pablonyx:
Fast-follow to the #1906 + #1972 merges. Two independent one-liners:

**1. Main is currently RED — `cancelled` tone missing (commit `0e4177a9d`).** #1906 added the `cancelled` status to `WorkflowRunStatusV2`; #1972's `run-view-model.ts` predates it, so its total `RUN_STATUS_TONE: Record<WorkflowRunStatusV2, WorkflowRunTone>` map fails `tsc` repo-wide on main:

```
src/domain/workflows/run-view-model.ts(75,7): error TS2741: Property 'cancelled' is missing in type '{ running: "info"; awaiting_human: "info"; interrupted: "warning"; completed: "success"; failed: "danger"; }' but required in type 'Record<WorkflowRunStatusV2, WorkflowRunTone>'.
```

Fix: `cancelled: "neutral"` — matching run-presentation.ts's existing `cancelled: ["Cancelled", "neutral"]` convention and the map's own `?? "neutral"` fallback. Negative control: reverting the line locally reproduces the exact TS2741 above; with it, `tsc -p tsconfig.domain.json` is clean and `run-view-model.test.ts` passes 53/53.

**2. line-clamp previews collapse under Button's `whitespace-nowrap` (commit `1d4b743b1`).** The FE-STRUCT-1 Button conversion in #1972 (`1deae734f`) wrapped the workflow cards in the `Button` primitive, whose base class includes `whitespace-nowrap`; `white-space` is CSS-inherited, so the `line-clamp-2` prompt/subtitle spans collapse to one clipped line. Flagged by #1972's delta review; the fix was pushed to that branch 4 minutes after merge, so it never landed — this is the identical cherry-pick. Adds `whitespace-normal` to the three spans (WorkflowBuilderChainCanvas.tsx ×2, WorkflowGraphView.tsx ×1). Scoped vitest: WorkflowGraphView + WorkflowGraphNodeCard, 24/24 pass.

The 12 CI failures on the first push here were main's breakage (item 1) plus a setup-terraform 429 download flake — not this PR's diff.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #2025, opened 2026-08-17
fix(product-client): restore login first-load budget headroom (lazy lifecycle splits)

pablonyx:
## Context

Main is currently at a knife edge on the "Login first-load budget (phase-6)" CI gate: `jsBytes` was 500263 B against a hard cap of 502000 B, leaving only ~1.7KB of local headroom, and CI's gzip measurement runs about 0.36% higher than local. Every open frontend PR has been coin-flipping this gate by +/-10 bytes of minifier noise:

- PR #2005 measured 502004 B (+4 over cap)
- PR #1992 measured 502008 B (+8 over cap)
- PR #1993 measured 501999 B (+1 headroom)

This PR frees several KB of headroom mechanically, following the same playbook as PR #1993 (rung 11 of the chat-scroll ladder): find code statically imported by the login-reachable chain that is never needed for first paint of `/login`, and move it behind a lazy chunk.

## What changed and why it is not login-critical

**Correction (post-review):** an earlier version of this description misattributed the savings to gating `useWorkspaceNavigationWorkflow`'s workspace-selection / agent-catalog / session-creation graph off `/login`. That is not what happened, and a fresh reviewer caught it — see below.

`useAppShortcuts`, mounted unconditionally by `ProductLifecycleRoot` (there is no auth gate before it), called `useWorkspaceNavigationWorkflow()` to back three shortcuts (`workspace.by-index`, `workspace.previous-workspace`, `workspace.next-workspace`) and their held-key traversal cursor. `useWorkspaceNavigationWorkflow()` was, and remains, also called unconditionally pre-auth from two other places this PR does not touch — `useAppNavigationCommandActions` and `useAppNewWorkspaceCommandActions`, both reached via `useAppCommandActions()` at `ProductLifecycleRoot.tsx`. Because ES modules load their full body the moment *any* caller imports them, `useWorkspaceNavigationWorkflow` (and everything it pulls in: `useWorkspaceSelection`, the bundled agent-catalog JSON, the session-creation graph) is **still** in the `/login` chunk after this PR, unrelated to this change.

The actual, verified savings mechanism: the three moved shortcuts were the *only* unconditional (pre-auth) callers of `useSidebarShortcutTargets` (which pulls in the sidebar-groups projection) and of the held-key traversal cursor machinery (`workspace-switch-cursor-controller.ts`, `sidebar-switch-cursor-store.ts`). `useSidebarShortcutTargets`'s only other caller, `MainSidebar.tsx`, is already inside the authenticated lazy chunk. Moving those three shortcuts (and the cursor controller they own) behind the same authenticated-only + lazy treatment `ProductLifecycleRoot` already gives `AuthRestartOfferRoot`, `SupportReportQueueRoot`, `AuthenticatedLaunchLifecycles`, and `AuthenticatedBackgroundLifecycles` is what removed the sidebar-target projection and cursor-controller code from the `/login` first-load chunk. The shortcuts were already no-ops signed out (no sidebar targets, no committed workspace to step from), so this changes only when the handlers get registered, never what they do.

Getting `useWorkspaceNavigationWorkflow` itself off `/login` would require the same authenticated-only treatment for `useAppNavigationCommandActions` / `useAppNewWorkspaceCommandActions` — that is a real follow-up candidate for more headroom, but is explicitly out of scope for this mechanical, minimally-reviewable PR.

## Review fixes (round 2, head `b11759689`)

A fresh reviewer found three issues on the first head (`dd6bef225`), all now fixed on this branch:
- **F1** (accuracy): corrected the four code-comment blocks that misattributed the savings mechanism (see above), and this PR description.
- **F2** (test strength): `ProductLifecycleRoot.test.tsx` no longer mocks `AuthenticatedWorkspaceSwitchShortcuts` away as a whole component; it mocks only the underlying `useWorkspaceSwitchShortcuts` hook (matching the stronger `AuthenticatedLaunchLifecycles` pattern) and asserts the hook is not called anonymous / is called once authenticated, so a dropped auth-gate regression is now caught.
- **F3** (latent leak): the traversal cursor's settle/fallback timers were never cleared on unmount. Before this PR the owner never unmounted; after this PR it mounts/unmounts with `authStatus`, so an auth flap mid-traversal could fire a stale commit post-unmount. Fixed by calling the controller's existing `cancel()` from the hook's effect cleanup, with a new test covering unmount mid-settle-window.

## Receipts

- Local measurement (`node scripts/measure-login-runtime-budget.mjs --no-build`), first head (`dd6bef225`): index chunk gzip JS **500263 B -> 495028 B** (a 5235 B / ~5.1KB reduction) against the founder cap of **502000 B**.
- Local re-measurement after the review fixes, current head (`b11759689`): **495020 B** (no regression; the ~8 B delta is the added unmount-cleanup call). Headroom against the cap: ~6980 B.
- CSS unaffected (65702 B, unchanged).
- CI on the first head (`dd6bef225`): "Login first-load budget (phase-6)" passed, CI-measured `JS 496680/502000 B` (headroom 5320 B). CI on the current head is pending re-confirmation.
- No behavior change: existing shortcut tests for the three moved handlers were relocated unmodified in assertions into `use-workspace-switch-shortcuts.test.tsx`; `ProductLifecycleRoot.test.tsx` and the cursor-controller unit tests all pass.

## Status

This PR is staged pending Pablo's ruling on the login-budget-gate phrasing/policy. Keeping as **draft** until then; not to be merged without his review.

### DIFF
--- a/apps/packages/product-client/src/hooks/app/lifecycle/use-app-shortcuts.test.tsx
+++ b/apps/packages/product-client/src/hooks/app/lifecycle/use-app-shortcuts.test.tsx
@@ -13,63 +13,9 @@ import {
 } from "#product/lib/domain/shortcuts/registry";
 import { USER_PREFERENCE_DEFAULTS } from "#product/lib/domain/preferences/user/model";
 import { useUserPreferencesStore } from "#product/stores/preferences/user-preferences-store";
-import { requestRightPanelTabByIndex } from "#product/lib/workflows/workspaces/right-panel-shortcut-requests";
-import { useSidebarSwitchCursorStore } from "#product/stores/workspaces/sidebar-switch-cursor-store";
-import { WORKSPACE_CURSOR_SETTLE_MS } from "#product/lib/domain/workspaces/sidebar/workspace-switch-cursor-controller";
-
-const navigationMocks = vi.hoisted(() => ({
-  selectWorkspaceFromSurface: vi.fn(),
-}));
-
-const harnessState = vi.hoisted(() => ({
-  selectedWorkspaceId: null as string | null,
-  selectedLogicalWorkspaceId: null as string | null,
-  digitShortcutTargets: [] as string[],
-  traversalShortcutTargets: [] as string[],
-}));
-
-vi.mock("#product/hooks/workspaces/derived/use-sidebar-shortcut-targets", () => ({
-  useSidebarShortcutTargets: () => ({
-    digitTargetIds: harnessState.digitShortcutTargets,
-    traversalTargetIds: harnessState.traversalShortcutTargets,
-  }),
-}));
-
-vi.mock("#product/hooks/workspaces/workflows/use-workspace-navigation-workflow", () => ({
-  useWorkspaceNavigationWorkflow: () => ({
-    selectWorkspaceFromSurface: navigationMocks.selectWorkspaceFromSurface,
-  }),
-}));
-
-vi.mock("#product/stores/sessions/session-selection-store", () => {
-  const readSelection = () => ({
-    selectedWorkspaceId: harnessState.selectedWorkspaceId,
-    selectedLogicalWorkspaceId: harnessState.selectedLogicalWorkspaceId,
-  });
-  const useSessionSelectionStore = (
-    selector: (state: {
-      selectedWorkspaceId: string | null;
-      selectedLogicalWorkspaceId: string | null;
-    }) => unknown,
-  ) => selector(readSelection());
-  // The traversal-cursor effect reads the committed selection imperatively and
-  // subscribes for commit reflection, so the mock exposes the same static
-  // surface the real zustand store does.
-  useSessionSelectionStore.getState = readSelection;
-  useSessionSelectionStore.subscribe = () => () => {};
-  return { useSessionSelectionStore };
-});
-
-vi.mock("#product/lib/workflows/workspaces/right-panel-shortcut-requests", () => ({
-  requestRightPanelTabByIndex: vi.fn(() => true),
-}));
 
 describe("useAppShortcuts", () => {
   beforeEach(() => {
-    harnessState.selectedWorkspaceId = null;
-    harnessState.selectedLogicalWorkspaceId = null;
-    harnessState.digitShortcutTargets = [];
-    harnessState.traversalShortcutTargets = [];
     clearShortcutHandlerRegistryForTests();
     useUserPreferencesStore.setState({
       ...USER_PREFERENCE_DEFAULTS,
@@ -82,59 +28,9 @@ describe("useAppShortcuts", () => {
     cleanup();
     clearShortcutHandlerRegistryForTests();
     document.body.innerHTML = "";
-    useSidebarSwitchCursorStore.getState().setCursor(null);
     vi.useRealTimers();
     vi.unstubAllGlobals();
     vi.clearAllMocks();
-  });
-
-  describe("held-key workspace traversal", () => {
-    it("previews the cursor on a step and commits the selection once after the settle", () => {
-      harnessState.selectedWorkspaceId = "workspace-1";
-      harnessState.selectedLogicalWorkspaceId = "workspace-1";
-      harnessState.digitShortcutTargets = ["pinned-workspace"];
-      harnessState.traversalShortcutTargets = ["workspace-1", "workspace-2", "workspace-3"];
-      vi.useFakeTimers();
-
-      renderHook(() => useAppShortcuts(commandActions()));
-
-      expect(runShortcutHandler("workspace.next-workspace", { source: "keyboard" })).toBe(true);
-      // The step previews immediately via the cursor, without committing.
-      expect(useSidebarSwitchCursorStore.getState().cursorId).toBe("workspace-2");
-      expect(navigationMocks.selectWorkspaceFromSurface).not.toHaveBeenCalled();
-
-      // The one expensive selection commit fires only after movement settles.
-      act(() => {
-        vi.advanceTimersByTime(WORKSPACE_CURSOR_SETTLE_MS);
-      });
-      expect(navigationMocks.selectWorkspaceFromSurface).toHaveBeenCalledTimes(1);
-      expect(navigationMocks.selectWorkspaceFromSurface).toHaveBeenCalledWith(
-        "workspace-2",
-        "shortcut",
-      );
-    });
-
-    it("cancels an uncommitted preview on Escape without committing", () => {
-      harnessState.selectedWorkspaceId = "workspace-1";
-      harnessState.selectedLogicalWorkspaceId = "workspace-1";
-      harnessState.traversalShortcutTargets = ["workspace-1", "workspace-2", "workspace-3"];
-      vi.useFakeTimers();
-
-      renderHook(() => useAppShortcuts(commandActions()));
-
-      runShortcutHandler("workspace.next-workspace", { source: "keyboard" });
-      expect(useSidebarSwitchCursorStore.getState().cursorId).toBe("workspace-2");
-
-      act(() => {
-        window.dispatchEvent(new KeyboardEvent("keydown", { key: "Escape" }));
-      });
-      expect(useSidebarSwitchCursorStore.getState().cursorId).toBeNull();
-
-      act(() => {
-        vi.advanceTimersByTime(WORKSPACE_CURSOR_SETTLE_MS);
-      });
-      expect(navigationMocks.selectWorkspaceFromSurface).not.toHaveBeenCalled();
-    });
   });
 
   it("steps window zoom through the registered app shortcuts without changing font sizes", () => {
@@ -164,51 +60,6 @@ describe("useAppShortcuts", () => {
     expect(useUserPreferencesStore.getState().windowZoomId).toBe("zoom120");
     expect(useUserPreferencesStore.getState().uiFontSizeId).toBe("xxxlarge");
     expect(useUserPreferencesStore.getState().readableCodeFontSizeId).toBe("large");
-  });
-
-  it("routes option-number shortcuts to the right panel when right-panel focus is active", () => {
-    harnessState.selectedWorkspaceId = "workspace-1";
-    harnessState.selectedLogicalWorkspaceId = "workspace-1";
-    harnessState.digitShortcutTargets = ["workspace-1", "workspace-2", "workspace-3"];
-    const zone = document.createElement("div");
-    zone.tabIndex = 0;
-    zone.setAttribute("data-focus-zone", "right-panel");
-    document.body.append(zone);
-    zone.focus();
-
-    renderHook(() => useAppShortcuts(commandActions()));
-
-    expect(runShortcutHandler("workspace.by-index", {
-      source: "keyboard",
-      digit: 2,
-    })).toBe(true);
-    expect(requestRightPanelTabByIndex).toHaveBeenCalledWith(2);
-    expect(navigationMocks.selectWorkspaceFromSurface).not.toHaveBeenCalled();
-  });
-
-  it("falls back to workspace selection when a stale right-panel focus request is unhandled", () => {
-    harnessState.selectedWorkspaceId = "workspace-1";
-    harnessState.selectedLogicalWorkspaceId = "workspace-1";
-    harnessState.digitShortcutTargets = ["pin-1", "pin-2", "workspace-1"];
-    harnessState.traversalShortcutTargets = ["workspace-1", "pin-1", "pin-2"];
-    const zone = document.createElement("div");
-    zone.tabIndex = 0;
-    zone.setAttribute("data-focus-zone", "right-panel");
-    document.body.append(zone);
-    zone.focus();
-    vi.mocked(requestRightPanelTabByIndex).mockReturnValueOnce(false);
-
-    renderHook(() => useAppShortcuts(commandActions()));
-
-    expect(runShortcutHandler("workspace.by-index", {
-      source: "keyboard",
-      digit: 2,
-    })).toBe(true);
-    expect(requestRightPanelTabByIndex).toHaveBeenCalledWith(2);
-    expect(navigationMocks.selectWorkspaceFromSurface).toHaveBeenCalledWith(
-      "pin-2",
-      "shortcut",
-    );
   });
 
   it("routes workspace copy shortcuts through app command actions", () => {
--- a/apps/packages/product-client/src/hooks/app/lifecycle/use-app-shortcuts.ts
+++ b/apps/packages/product-client/src/hooks/app/lifecycle/use-app-shortcuts.ts
@@ -1,22 +1,6 @@
-import { useEffect, useRef } from "react";
-
 import { useShortcutHandler } from "#product/hooks/shortcuts/lifecycle/use-shortcut-handler";
 import type { AppCommandActions } from "#product/hooks/app/workflows/app-command-action-types";
-import { useSidebarShortcutTargets } from "#product/hooks/workspaces/derived/use-sidebar-shortcut-targets";
-import { useWorkspaceNavigationWorkflow } from "#product/hooks/workspaces/workflows/use-workspace-navigation-workflow";
-import {
-  focusChatInput,
-  getFocusZone,
-  isRightPanelFocusZone,
-} from "#product/lib/domain/focus-zone";
-import { resolveSidebarShortcutDigitTarget } from "#product/lib/domain/workspaces/sidebar/sidebar-shortcut-targets";
-import {
-  createWorkspaceSwitchCursorController,
-  type WorkspaceSwitchCursorController,
-} from "#product/lib/domain/workspaces/sidebar/workspace-switch-cursor-controller";
-import { requestRightPanelTabByIndex } from "#product/lib/workflows/workspaces/right-panel-shortcut-requests";
-import { useSessionSelectionStore } from "#product/stores/sessions/session-selection-store";
-import { useSidebarSwitchCursorStore } from "#product/stores/workspaces/sidebar-switch-cursor-store";
+import { focusChatInput } from "#product/lib/domain/focus-zone";
 import { useWorkspaceUiStore } from "#product/stores/preferences/workspace-ui-store";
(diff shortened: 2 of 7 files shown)


pablonyx:
## Review — fresh pass, head `dd6bef2250a518d54dab7676eb74e7a32a88d6da`

**Verdict: FINDINGS** (non-blocking overall — behavior parity holds, tests pass — but one finding materially undercuts the PR's stated mechanism and should be fixed before relying on this pattern for future headroom work).

### What I verified
- Full diff read (`gh pr diff 2025`), plus the surrounding `ProductLifecycleRoot.tsx`, `App.tsx`, `ProductClient.tsx`, the shortcut registry/dispatcher, `use-workspace-navigation-workflow.ts`, `use-workspace-selection.ts`, `use-app-navigation-command-actions.ts`, `use-app-new-workspace-command-actions.ts`, `use-sidebar-shortcut-targets.ts`, and the moved controller/store files, in a fresh worktree off the fetched PR head (not the shared login-headroom worktree).
- Ran the three touched test files with vitest/jsdom after a scoped `pnpm install`: all pass (`use-app-shortcuts.test.tsx` 10/10, `use-workspace-switch-shortcuts.test.tsx` 4/4, `ProductLifecycleRoot.test.tsx` 7/7).
- Traced the static import graph from `ProductLifecycleRoot.tsx` to check the PR's central claim (attack surface #4).
- Confirmed all touched files are well under the 400-line frontend gate (117–299 lines) and match neighboring style (no-semicolon in `providers/`, semicolons in `hooks/`, JSDoc block matching `AuthenticatedBackgroundLifecycles`/`AuthenticatedLaunchLifecycles`).
- Confirmed no other call site registers `workspace.by-index` / `workspace.previous-workspace` / `workspace.next-workspace`, so there's no handler-priority/ordering conflict from the new mount point.

### What I skipped
- Did not run a full Vite build or re-run `measure-login-runtime-budget.mjs` myself (no `--no-build` fresh dist locally; would require a full `apps/web` build). I did not re-verify the exact byte counts; I instead verified the *mechanism* by tracing which modules are/aren't reachable outside this PR's change.
- Did not run the full frontend test suite or any Rust/build steps (out of scope, and per machine limits).

### Findings

**F1 — non-blocking (fix before merge), file: PR description + code comments in `use-app-shortcuts.ts:13-21`, `use-workspace-switch-shortcuts.ts:564-573`, `AuthenticatedWorkspaceSwitchShortcuts.tsx:676-684`, `ProductLifecycleRoot.tsx:90-95`**
The stated mechanism is inaccurate. The PR claims moving the 3 shortcuts out of `useAppShortcuts` keeps `useWorkspaceNavigationWorkflow`'s "workspace selection / agent-catalog / session-creation graph" off the `/login` first-load chunk. It doesn't — that exact hook (and everything it imports: `useWorkspaceSelection` → `useCloudAgentCatalogCache` ["the bundled agent-catalog JSON"], `useWorkspaceBootstrapActions`, `runWorkspaceSelection`/`runHotWorkspaceReopen` ["the session-creation graph"]) remains statically, unconditionally imported into the exact same login-reachable chunk via two other call sites this PR does not touch:
- `use-app-navigation-command-actions.ts:11,33` (`goToTopLevelRoute`)
- `use-app-new-workspace-command-actions.ts:9,37` (`navigateToWorkspaceShell`)

Both run through `useAppCommandActions()`, called unconditionally at `ProductLifecycleRoot.tsx:181` (`ProductLifecycles`), before any auth check — `ProductLifecycleRoot` wraps `App` directly (`ProductClient.tsx`), and `App`'s `/login` route sits inside that same tree, so this is genuinely pre-auth-reachable. ES modules load their full body/imports the moment *any* caller imports them, regardless of which exported function is actually invoked, so `useWorkspaceNavigationWorkflow`'s module was never removed from the login chunk by this diff.

The measured reduction is very likely real, just for a different reason: `useSidebarShortcutTargets` (only reachable via the old `useAppShortcuts` pre-auth, now genuinely gated) transitively pulls in `sidebar-groups.ts` (381 lines), plus `workspace-switch-cursor-controller.ts` (183 lines), `sidebar-shortcut-targets.ts` (59 lines), and `sidebar-switch-cursor-store.ts` (29 lines) — none of which are imported from any other unconditional call site. That's a plausible ~5KB gzip source on its own, independent of `useWorkspaceNavigationWorkflow`.

Why this matters given the context: this is a knife-edge, repeatedly-coin-flipped CI budget gate. Comments/PR text that misattribute the win to "the agent-catalog/session-creation graph is now gated" will mislead the next engineer doing headroom work into believing that graph is already off the login path — it is not. If someone grows `selectWorkspaceFromSurface`'s body later, it still bills straight to `/login` via the two paths above. Recommend correcting the four comment blocks and the PR description, and if the actual goal is to get `useWorkspaceNavigationWorkflow` off `/login`, that needs `useAppNavigationCommandActions`/`useAppNewWorkspaceCommandActions` to get the same treatment — out of scope for this mechanical PR, but should be said explicitly rather than implied as already done.

**F2 — non-blocking, file: `ProductLifecycleRoot.test.tsx:87-89`**
The added mock (`AuthenticatedWorkspaceSwitchShortcuts: () => null`) only prevents an unmocked-dynamic-import crash in unrelated tests; nothing in the suite asserts the new component is actually gated on `authStatus`. Compare to the established stronger pattern for `AuthenticatedLaunchLifecycles` (not mocked as a component at all — its underlying `useHomeDeferredLaunchRunner` hook is mocked, and a dedicated test at line 267 asserts it's *not* called while anonymous and *is* called once authenticated). A regression that drops the `authStatus === "authenticated"` guard around `<AuthenticatedWorkspaceSwitchShortcuts />` (`ProductLifecycleRoot.tsx:291`) would not be caught here. (Note: `AuthenticatedBackgroundLifecycles`'s mock has the same gap already on main — this PR just follows the weaker existing precedent rather than the stronger one.)

**F3 — non-blocking / latent, file: `use-workspace-switch-shortcuts.ts` + `workspace-switch-cursor-controller.ts`**
Before this PR, the owning mount point (`useAppShortcuts` inside `ProductLifecycles`) never unmounted for the app's lifetime, so a subtlety in `createWorkspaceSwitchCursorController` was unreachable: its `settleTimer`/`fallbackTimer` are never cleared by the hook's `useEffect` cleanup (which only unsubscribes the store and removes the Escape listener). After this PR, the owner mounts/unmounts with `authStatus`, so a forced re-auth mid-traversal (inside the 180ms settle or 2000ms commit-fallback window) can fire `commitSelection` after unmount via a ref frozen to the last pre-unmount render, acting on stale `selectedLogicalWorkspaceId`/`logicalWorkspaces`/`webApp` closures from `useWorkspaceNavigationWorkflow`. Narrow window, not data-destructive, but worth a follow-up (cancel the controller or guard with an `isMounted` check in the effect cleanup).

### Not a finding (verified clean)
- Shortcut semantics (modifier handling, right-panel routing, held-key traversal, Escape-cancel, digit resolution) moved verbatim; relocated tests are the same assertions, unweakened, and pass.
- Lazy-mount pattern (`lazy()` + `Suspense fallback={null}` + `authStatus === "authenticated"`) is identical to the four existing siblings (`AuthRestartOfferRoot`, `SupportReportQueueRoot`, `AuthenticatedLaunchLifecycles`, `AuthenticatedBackgroundLifecycles`); a chunk-load failure surfaces through the same `AppErrorBoundary` all siblings already rely on — not a new/silent failure mode. A shortcut pressed before the lazy chunk resolves is a graceful no-op (`runShortcutHandler` returns `false` when no handler is registered), matching pre-auth behavior today.
- `use-app-shortcuts.ts` post-prune still registers everything it should (`app.*`, `workspace.new-*`, `workspace.copy-*`, `workspace.toggle-cowork-threads`); no accidental removal beyond the three intended handlers.
- No other registration site for the three moved shortcut IDs, so no priority/ordering conflict from the new mount point.
- No sneaky re-export: only `AuthenticatedWorkspaceSwitchShortcuts.tsx` imports `use-workspace-switch-shortcuts.ts`, and `ProductLifecycleRoot.tsx` only reaches it via `lazy()`.
- File sizes and style conventions all clean; no scope creep beyond the mechanical split.

pablonyx:
All three findings fixed on this branch, new head \`b11759689\`.

**F1 (accuracy):** corrected the four code-comment blocks (\`use-app-shortcuts.ts\`, \`use-workspace-switch-shortcuts.ts\`, \`AuthenticatedWorkspaceSwitchShortcuts.tsx\`, \`ProductLifecycleRoot.tsx\`) and the PR description. The real mechanism: the three moved shortcuts were the only unconditional pre-auth callers of \`useSidebarShortcutTargets\` (sidebar-groups projection) and the traversal cursor controller/store — those are what's now genuinely off \`/login\`. \`useWorkspaceNavigationWorkflow\` remains eager via \`useAppNavigationCommandActions\`/\`useAppNewWorkspaceCommandActions\`; noted as an explicit out-of-scope follow-up candidate rather than implying it's already gated. Did not attempt that split here, per your instruction to keep scope minimal.

**F2 (test strength):** \`ProductLifecycleRoot.test.tsx\` no longer mocks \`AuthenticatedWorkspaceSwitchShortcuts\` away as a component. It now mocks only the underlying \`useWorkspaceSwitchShortcuts\` hook (matching the \`AuthenticatedLaunchLifecycles\`/\`useHomeDeferredLaunchRunner\` pattern) and a new test asserts the hook is not called while anonymous and is called once authenticated — a dropped auth gate now fails the suite.

**F3 (latent leak):** added \`controller.cancel()\` to the hook's effect cleanup so pending settle/fallback timers are cleared on unmount, with a new test that starts a traversal, unmounts mid-settle-window (simulating an auth flap), advances timers past both windows, and asserts \`selectWorkspaceFromSurface\` is never called.

Numbers after the fixes:
- Local re-measurement: index chunk gzip JS **495020 B** (was 495028 B pre-fix, ~8 B delta from the added cleanup call — no regression), cap 502000 B.
- CI on this head: "Login first-load budget (phase-6)" **pass**, \`JS 496679/502000 B\` (headroom 5321 B).
- All touched tests pass: \`use-app-shortcuts.test.tsx\` (10), \`use-workspace-switch-shortcuts.test.tsx\` (5, one new), \`ProductLifecycleRoot.test.tsx\` (8, one new), \`workspace-switch-cursor-controller.test.ts\` (13, unmodified).

Still draft, not merging.

pablonyx:
## Re-verification — delta `dd6bef225` → `b1175968`

**Verdict: CLEAN after fixes.**

Checked the delta only (`git diff dd6bef225..b1175968`, 6 files, +126/-26), in a fresh throwaway worktree, against my three prior findings:

**F1 (import-graph rationale)** — fixed in all four comment blocks (`use-app-shortcuts.ts`, `use-workspace-switch-shortcuts.ts`, `AuthenticatedWorkspaceSwitchShortcuts.tsx`, `ProductLifecycleRoot.tsx`) plus the PR body. Each now correctly attributes the win to `useSidebarShortcutTargets` + the traversal cursor controller/store (the modules that were genuinely pre-auth-only through this owner), explicitly states `useWorkspaceNavigationWorkflow` is *not* gated by this split and remains reachable via `useAppNavigationCommandActions`/`useAppNewWorkspaceCommandActions`, and flags getting that graph off `/login` as an explicit, separate follow-up. No overclaiming left.

**F2 (auth-gating test coverage)** — fixed. `ProductLifecycleRoot.test.tsx` no longer mocks `AuthenticatedWorkspaceSwitchShortcuts` as a component; it mocks the underlying `useWorkspaceSwitchShortcuts` hook and adds "keeps the workspace-switch shortcuts off the signed-out shell and mounts them once authenticated," mirroring the `AuthenticatedLaunchLifecycles`/`useHomeDeferredLaunchRunner` pattern exactly. **Verified as a real regression test**: reverted the `authStatus === "authenticated"` guard to `true` and reran — this new test fails (`expected "spy" to not be called at all, but actually been called 1 times`); restored, reran clean.

**F3 (stale timer post-unmount)** — fixed. Effect cleanup in `use-workspace-switch-shortcuts.ts` now calls `controller.cancel()`, clearing the pending settle/fallback timers (and any pending commit/cursor state) on unmount. New test "clears pending settle/fallback timers on unmount so a stale commit never fires (auth-flap mid-traversal)" added. **Verified as a real regression test**: commented out the `controller.cancel()` cleanup call and reran — this new test fails (`selectWorkspaceFromSurface` called once post-unmount with stale args); restored, reran clean.

All three touched test files pass in full: 23/23 (`use-app-shortcuts.test.tsx` 10, `use-workspace-switch-shortcuts.test.tsx` 5, `ProductLifecycleRoot.test.tsx` 8).

Nothing new introduced by the delta itself needs follow-up. CI-reported 496679/502000 taken at face value (did not re-run the byte-budget measurement).

pablonyx:
REFOLD RECEIPT (onto main 703c12e6e, post chat-scroll r2-r12 + bg r8 + #2042): old head b11759689 -> new head 7dc0e3586. Range-diff: both commits ! (real conflict resolution, net delta for fresh review per orchestrator order).

Conflicts and resolutions:
1. use-app-shortcuts.ts (2 hunks): resolved to this PR's pruned side. Main's only change in the pruned regions was the useSidebarShortcutTargets API split ({digitTargetIds, traversalTargetIds}); moot in this file since the PR removes those blocks entirely.
2. use-app-shortcuts.test.tsx (3 hunks): resolved to the pruned side; all of main's 25 changed lines were workspace-switch fixtures that this PR relocates.
3. Semantic ports (no textual conflict, required for correctness on the new base): use-workspace-switch-shortcuts.ts updated to the split API (destructure, traversal ref, digit resolve); use-workspace-switch-shortcuts.test.tsx ported to main's evolved fixtures including the #2030 pinned-digit semantics (digit=[pin-1,pin-2,workspace-1], assertion workspace-2 -> pin-2) and the split mock harness; the unmount-timer test from the review-fix commit moved onto traversalShortcutTargets.

Verification: 23/23 tests pass locally across use-workspace-switch-shortcuts.test.tsx (5), use-app-shortcuts.test.tsx (10), ProductLifecycleRoot.test.tsx (8). Local budget gate literal: [login-budget] JS 491750/502000 B, CSS 65771/66250 B, PASSED — note local gzip under-reports ~0.36% vs CI; CI is the number of record. Not merging, draft bit untouched; fresh delta review to be tasked by the orchestrator.

pablonyx:
DELTA REVIEW VERDICT: CLEAN

Pinned to head `7dc0e35867bb2ba86021c7c5191560f1174f5a81` (confirmed current via `gh pr view 2025 --json headRefOid` — matches the refold receipt's claimed new head; PR still draft, mergeable).

**Range-diff** (`47f8e04f7..b11759689` vs `703c12e6e..7dc0e3586`, both bases verified as the actual pre-refold/post-refold merge-bases): both commits show `!` (changed), consistent with the receipt. Read every changed hunk in full against base `703c12e6e`:

- `use-app-shortcuts.ts` / `use-app-shortcuts.test.tsx`: hunk resolved to this PR's pruned side. Confirmed moot — main's only change in the touched regions was the `useSidebarShortcutTargets` return-shape split (flat array → `{digitTargetIds, traversalTargetIds}`, landed via #2030 pinned-workspace-digits), and this PR deletes those call sites from the file entirely, so there was nothing to reconcile here.
- `use-workspace-switch-shortcuts.ts` / `.test.tsx` (semantic port, no textual conflict): correctly adapted to destructure the split API — `digitTargetIds` feeds `resolveSidebarShortcutDigitTarget` (Cmd+1..9), `traversalTargetIds` feeds the cursor controller (Cmd+Opt+Arrow) — matching how `MainSidebar.tsx` (unmodified by this PR, already on main) consumes the same split API. Test fixtures updated to main's pinned-digit semantics (`digitShortcutTargets = ["pin-1","pin-2","workspace-1"]`, asserting digit 2 → `"pin-2"`); the unmount/timer-leak regression test (F3 from the prior review round) ported cleanly onto `traversalShortcutTargets`.
- `ProductLifecycleRoot.tsx`/`.test.tsx` and the new `AuthenticatedWorkspaceSwitchShortcuts.tsx`: byte-for-byte the same mechanism as before the refold — lazy + `Suspense`, gated on `authStatus === "authenticated"`, no new behavior.

Net conclusion: the refold's entire delta is a base-API adaptation (flat array → split digit/traversal object, required by #2030 landing on main) plus matching test-fixture updates. No new behavior, no scope change from the previously-reviewed intent (move workspace-switch shortcuts to an authenticated-only lazy chunk to restore `/login` first-load budget headroom).

**CI at exact head `7dc0e35867bb2ba86021c7c5191560f1174f5a81`** — literal `check-runs` tally, max run per name:

```
23 latest-per-name runs, all conclusion "success", 0 failures:
Analyze (javascript-typescript), Analyze (python), Analyze (rust), Build SDK,
CI/CD config checks, Candidate build handoff, Cargo check & test, CodeQL,
Desktop frontend build, Detect smoke-relevant changes,
Login first-load budget (phase-6), Mobile typecheck,
Production compose smoke, Repo shape checks, Shared frontend packages,
Terraform validate (server/infra), Transcript scroll physics (chromium + webkit),
Validate PR title and labels, Vercel Preview Comments, Web frontend build,
Workflow definition lifecycle (tier-2), intent-billing (provisional),
intent-tests (provisional)
```

Budget-gate job literal output (pulled from the actual job log, not the receipt's self-report):
```
[login-budget] JS 493362/502000 B  CSS 65906/66250 B  fonts/images/audio font=0 image=0 audio=0
[login-budget] budget gate PASSED (within all ceilings).
```

No blocking findings.

pablonyx:
MERGE RECEIPT: head-pinned squash at reviewed head 7dc0e35867bb2ba86021c7c5191560f1174f5a81 -> main 6cf295c490d88c843a35cdb119c2d3dbed0f17e8. Authority: Pablo's standing class approval (reviewed-CLEAN PRs merge) + live launch word 2026-08-18 ~17:3xZ ('can we merge everything please... trying to get things pushed for a launch'). Review of record: prior full review + delta review CLEAN at this exact head (comment 5331566052; 23/23 latest-per-name check-runs success, budget-gate literal '[login-budget] JS 493362/502000 B ... budget gate PASSED'). Freshness: programmatic ZERO file overlap (7 PR files vs main's 18-file delta 703c12e6e..ed80a0563), pairwise disjoint from the co-merging wave (#2014/#2043), mergeable=CLEAN. Ready-flip was mechanical execution of the approved merge.

PR #2034, opened 2026-08-18
feat(workflows): per-leg authored prompts with N-way fan-out (F5+F6, rung 5)

pablonyx:
Rung 5 of the Workflows follow-up ladder (frozen ADR, rulings F5 + F6): a node may carry 2..=8 per-leg authored prompts and fans out one session per leg; crash-resume truncates the ledger and re-fans-out all N.

## Grammar (all three planes)
- Runtime: `DefinitionLeg` (camelCase, `deny_unknown_fields`), `DefinitionNode.legs: Option<Vec<DefinitionLeg>>` — absent serializes byte-identically to today; 2..=8 enforced, `legs[0].prompt` must equal `node.prompt`, every leg non-blank with per-leg reference validation naming the leg index. Chain stays linear at node level; legs have no edges.
- Control plane: `WorkflowNodeLegV2` in `models_v2.py` (min_length=2, max_length=8, mirror validator).
- Client: `definition-v2.ts` per-leg reference scan (issue codes `invalid_leg_count` / `blank_leg_prompt` / `leg_prompt_mismatch`), builder draft add/remove/update-leg with leg 0 mirroring the node prompt and collapse-to-absent, builder UI leg authoring.

## Engine
- `ResolvedSideEffect::StartNodeLegs`; one-leg nodes keep `StartNode` byte-identically.
- `launch_legs` (new `live/workflows/launch.rs`): render one envelope per leg, mint N sessions, `stamp_fanout` writes the representative scalar stamp + N ledger rows in one commit, start+dispatch each, compensate all on any failure.
- F6 resume: `ResumeNode` truncates the ledger (`truncate_legs_tx`) and re-emits the fan-out effect.
- Tier-3 scenario `t3-wf-parallel-1.ts` authored and registered (heterogeneous 3-leg panel, prompt-to-leg assertion, crash boundary, re-fan-out assertions).

## Evidence (literals)
- `cargo test -p anyharness-lib workflow` → "test result: ok. 212 passed; 0 failed" (prior rung-4 gate was 196; +16 new).
- Negative control (prompt-to-leg invariant): pinned `leg_prompts()` to repeat leg 0's prompt → exactly `each_leg_envelope_carries_only_its_own_leg_prompt` fails; restored → "ok. 1 passed; 0 failed".
- Run-terminal Cancel control (rung-4 carry, first compile of its round-2 test in this stack): pinned the round-1 only-current-node stamp → exactly `cancel_marks_every_running_leg_in_the_run_cancelled` fails; restored → 2/2 pass.
- Control plane: `server/tests/unit/test_workflow_definition_v2_validation.py` → "42 passed".
- Client: workflows builder/domain suites → "Test Files 5 passed (5), Tests 75 passed (75)".
- `scripts/check_max_lines.py` → "Max-lines check passed"; `check_frontend_boundaries.py` → "Frontend boundary check passed."
- Committed OpenAPI artifacts regenerated (`DefinitionLeg` present in `components.schemas`).

## Fix log (burst disclosures)
- The rung was implemented under a machine-wide cargo hold (file-edits-only); on release, first compile surfaced one error: `actor.rs` used `launch_model` after its move to `launch.rs` without importing it (E0425). Fixed in the same burst (one-line import), disclosed here per the honesty-debt order.

## Notes
- The per-leg `sessions` rollup the t3 scenario reads is rung-7 shape — flagged for reconciliation there.
- Per-leg redo (`RedoLeg`, `DisposeSessionThenStartLeg`) is rung 6 and builds on the side-effect vocabulary added here.

## Review fixes on this rung
1. [BLOCKING finding, reviewer of record] A mid-fan-out launch failure stranded every freshly stamped leg at `running` against compensated, disposed sessions — `NodeLaunchFailed` carries no turn session and the representative fallback stamped leg 0 only. Fixed in 3be8505a2: the launch-failure stamp goes keyless and covers every row the fan-out inserted. Regression test `a_failed_fan_out_launch_terminalizes_every_leg_it_stamped` (3 legs, all land `Failed(NodeLaunchFailed)` + `completed_at`); negative control (fallback restored) fails at exactly "leg 1". Full suite after fix: "test result: ok. 213 passed; 0 failed". store.rs ratchet entry amended 1411→1416 within this rung's own new-content entry (+5, reason string updated).
2. [Non-blocking] Commit-message/body contradiction on OpenAPI artifacts reconciled: the "knowingly stale" sentence predated the release burst and has been removed from the feat commit message via reword — the committed artifacts in this PR are current (DefinitionLeg present in both generated files).

### DIFF
--- a/anyharness/crates/anyharness-lib/src/api/workflow_runs_openapi.rs
+++ b/anyharness/crates/anyharness-lib/src/api/workflow_runs_openapi.rs
@@ -34,6 +34,7 @@ use utoipa::OpenApi;
         crate::domains::workflows::model::WorkflowNodeType,
         crate::domains::workflows::definition::WorkflowDefinition,
         crate::domains::workflows::definition::DefinitionNode,
+        crate::domains::workflows::definition::DefinitionLeg,
         crate::domains::workflows::definition::NodeModel,
         crate::domains::workflows::definition::DefinitionEdge,
         crate::domains::workflows::definition::DefinitionInput,
--- a/anyharness/crates/anyharness-lib/src/domains/workflows/definition.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/workflows/definition.rs
@@ -22,6 +22,11 @@ use super::model::WorkflowNodeType;
 
 pub const DEFINITION_SCHEMA_VERSION: u32 = 2;
 
+/// The most parallel leg prompts one node may fan out to (ruling F5). A node
+/// with `legs` present must carry 2..=`MAX_NODE_LEGS`; a single-leg node is
+/// expressed the old way (no `legs`), so `legs` of length 0 or 1 is invalid.
+pub const MAX_NODE_LEGS: usize = 8;
+
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ToSchema)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 pub struct WorkflowDefinition {
@@ -48,6 +53,33 @@ pub struct DefinitionNode {
     pub prompt: String,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub model: Option<NodeModel>,
+    /// Parallel leg prompts (ruling F5). Absent means today's 1:1 node/session
+    /// behavior exactly (and serializes byte-identically). Present means a
+    /// parallel node: 2..=`MAX_NODE_LEGS` legs, each with its own authored
+    /// prompt, `legs[0].prompt` equal to `prompt` so leg 0 is the
+    /// representative session and `leg_index` addresses `legs[i]` uniformly.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub legs: Option<Vec<DefinitionLeg>>,
+}
+
+impl DefinitionNode {
+    /// The authored prompt of every leg in order — the multi-leg list when
+    /// present, else the single node prompt. Leg 0 is always the node prompt
+    /// (validation pins `legs[0].prompt == prompt`).
+    pub fn leg_prompts(&self) -> Vec<&str> {
+        match &self.legs {
+            Some(legs) if legs.len() > 1 => legs.iter().map(|leg| leg.prompt.as_str()).collect(),
+            _ => vec![self.prompt.as_str()],
+        }
+    }
+}
+
+/// One parallel leg's authored prompt (ruling F5). Legs have no edges,
+/// ordering, or inter-leg dependencies — the chain stays linear at node level.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ToSchema)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct DefinitionLeg {
+    pub prompt: String,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
@@ -217,27 +249,54 @@ impl WorkflowDefinition {
 
         let chain = self.linear_chain(&node_ids)?;
 
-        for node in &self.nodes {
-            let references = parse_references(&node.prompt)
-                .map_err(|error| invalid(format!("node '{}': {}", node.id, error.detail)))?;
+        // Reference-grammar validation for one prompt, labelled by its owner
+        // (the node, or `node 'x' leg i` for a fan-out leg) so a bad reference
+        // inside a leg names the leg (ruling F5).
+        let validate_refs = |prompt: &str, label: &str| -> Result<(), DefinitionValidationError> {
+            let references = parse_references(prompt)
+                .map_err(|error| invalid(format!("{label}: {}", error.detail)))?;
             for reference in references {
                 match reference {
-                    PromptReference::Input(name) => {
-                        if !input_names.contains(name.as_str()) {
-                            return Err(invalid(format!(
-                                "node '{}' references undeclared @input:{name}",
-                                node.id
-                            )));
-                        }
+                    PromptReference::Input(name) if !input_names.contains(name.as_str()) => {
+                        return Err(invalid(format!("{label} references undeclared @input:{name}")));
                     }
-                    PromptReference::Doc(slug) => {
-                        if !doc_slugs.contains(slug.as_str()) {
-                            return Err(invalid(format!(
-                                "node '{}' references unknown @doc:{slug}",
-                                node.id
-                            )));
-                        }
+                    PromptReference::Doc(slug) if !doc_slugs.contains(slug.as_str()) => {
+                        return Err(invalid(format!("{label} references unknown @doc:{slug}")));
                     }
+                    _ => {}
+                }
+            }
+            Ok(())
+        };
+
+        for node in &self.nodes {
+            validate_refs(&node.prompt, &format!("node '{}'", node.id))?;
+            // Ruling F5: a parallel node carries 2..=MAX_NODE_LEGS legs, leg 0's
+            // prompt equals the node prompt (so every old consumer of
+            // node.prompt stays correct), and every leg prompt is non-blank and
+            // reference-grammar valid. `linear_chain` above is untouched: legs
+            // are inside the node, not edges.
+            if let Some(legs) = &node.legs {
+                if !(2..=MAX_NODE_LEGS).contains(&legs.len()) {
+                    return Err(invalid(format!(
+                        "node '{}' has {} legs; a parallel node needs 2..={MAX_NODE_LEGS} \
+                         (a single-leg node omits `legs`)",
+                        node.id,
+                        legs.len()
+                    )));
+                }
+                if legs[0].prompt != node.prompt {
+                    return Err(invalid(format!(
+                        "node '{}' prompt must equal legs[0].prompt (leg 0 is the \
+                         representative session)",
+                        node.id
+                    )));
+                }
+                for (index, leg) in legs.iter().enumerate() {
+                    if leg.prompt.trim().is_empty() {
+                        return Err(invalid(format!("node '{}' leg {index} has a blank prompt", node.id)));
+                    }
+                    validate_refs(&leg.prompt, &format!("node '{}' leg {index}", node.id))?;
                 }
             }
         }
--- a/anyharness/crates/anyharness-lib/src/domains/workflows/definition_tests.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/workflows/definition_tests.rs
@@ -2,9 +2,9 @@
 //! invocation-snapshot revalidation, with a negative control per rule.
 
 use super::definition::{
-    parse_references, DefinitionEdge, DefinitionInput, DefinitionNode, DocTemplate,
+    parse_references, DefinitionEdge, DefinitionInput, DefinitionLeg, DefinitionNode, DocTemplate,
     InvocationPlacement, InvocationSnapshot, PlacementMode, PromptReference, WorkflowDefinition,
-    DEFINITION_SCHEMA_VERSION,
+    DEFINITION_SCHEMA_VERSION, MAX_NODE_LEGS,
 };
 use super::model::WorkflowNodeType;
 
@@ -15,6 +15,7 @@ fn node(id: &str, prompt: &str) -> DefinitionNode {
         title: format!("Node {id}"),
         prompt: prompt.into(),
         model: None,
+        legs: None,
     }
 }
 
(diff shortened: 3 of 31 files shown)


pablonyx:
REVIEW (reviewer of record, head 9ebee8fc1): FINDINGS

1. **[BLOCKING] Mid-fan-out launch failure leaves stale "running" ledger rows pointing at compensated/disposed sessions.** `launch_legs` (anyharness/crates/anyharness-lib/src/live/workflows/launch.rs:1012-1043) calls `stamp_fanout` — which commits all N leg rows as `status='running'` in one transaction — *before* the start/dispatch loop. If `start_and_dispatch` then fails for any leg (launch.rs:1037-1041), `compensate_all` (launch.rs:1087-1095) disposes every minted session and clears workflow columns, but it never touches `workflow_run_node_sessions`. The actor's error path (actor.rs `perform_side_effect`) applies `NodeLaunchFailed` → `Transition::FailNode`, whose handler `fail_node` (store.rs:1200-1210) only updates `workflow_run_nodes.status`; it does not call anything analogous to `node_sessions::cancel_all_run_legs_tx` (store.rs:658-661, node_sessions.rs:109-124), which is the exact mechanism this same stack already built to terminalize ledger rows for the Cancel case (rung-4 fix, `cancel_marks_every_running_leg_in_the_run_cancelled`). Net effect: after a failed fan-out, the run is correctly marked `Failed`, but every leg's ledger row is permanently stuck at `status='running'` referencing a session that no longer exists (or was disposed) — a genuine "partial/leaked ledger," which is precisely the invariant this review was asked to check (#3). This has no test coverage: there is no test anywhere in `launch.rs`/`actor.rs`/`store_fanin_tests.rs` that drives a leg-start/dispatch failure and asserts the resulting ledger state, so nothing would fail if this regressed further.
   - Failure scenario: a 3-leg node where leg 2's session-create or dispatch fails after legs 0/1 already succeeded and `stamp_fanout` committed. Run ends `Failed`, but legs 0/1/2 all still read `status='running'` in `workflow_run_node_sessions` with session ids that were just compensated — any consumer (UI leg status, future redo/resume logic) sees three "running" legs on a dead, failed run.

2. **[NON-BLOCKING] PR body self-contradicts on OpenAPI regen state.** The PR description's "Evidence" section claims "Committed OpenAPI artifacts regenerated (`DefinitionLeg` present in `components.schemas`)" but the commit message on 9ebee8fc1 says "Committed SDK OpenAPI artifacts are knowingly stale pending the held regeneration pass." The diff does include `DefinitionLeg` in both `openapi.json` and `openapi.ts` (lines ~1128-1140, ~1173-1188), so the artifacts in this diff do look current for this rung's schema addition — but the two disclosures directly conflict about whether that regen is trusted/complete. Please reconcile which statement is accurate so the ladder's next rung doesn't inherit a false assumption about SDK freshness.

Everything else checked out:
- Grammar parity (runtime/control-plane/client) is consistent: all three enforce legs 2..=8, `legs[0].prompt == prompt`, non-blank, per-leg reference validation with leg-indexed error labels (definition.rs:104-134, models_v2.py:70-110, definition-v2.ts:218-292).
- Serialization: `legs: Option<Vec<DefinitionLeg>>` is `skip_serializing_if = "Option::is_none"`, so an absent `legs` node is byte-identical to pre-rung shape; nothing else on `DefinitionNode` changed shape.
- F6 resume: `truncate_legs_tx` (node_sessions.rs:315-321) deletes all ledger rows for the node before `start_side_effect` re-emits the fan-out effect; a one-leg node's single row is deleted and re-inserted with no observable difference. No window found where a stale leg row from the old generation survives into a successful new generation.
- The one-leg path is untouched: `StartNode` byte-identical, `launch_model`/`launch_node`'s original body unchanged aside from the new early-return branch for `node_leg_count > 1`.
- `actor.rs` borrow across `launch_legs(...).await` (holding `&state.run`) — compiles and is logically sound for the success path (`*state = fresh` reloads after commit); on the error path the stale in-memory `state` is not an issue since the subsequent `NodeLaunchFailed` transition is looked up by `node_row_id`, not by cached node data.
- Tests otherwise do pin the claimed invariants: prompt-to-leg selection (`each_leg_envelope_carries_only_its_own_leg_prompt`), representative stamp, resume truncation + re-fan-out.

pablonyx:
DELTA RE-REVIEW (reviewer of record, fix head 3be8505a2): CLEAN

Verified the fix commit (store.rs + store_fanin_tests.rs, plus the ratchet/Cargo.lock housekeeping):

- `mark_leg_terminal_tx` with `session_id: None` executes `UPDATE workflow_run_node_sessions SET status=?, completed_at=? WHERE node_row_id=?1` — no session filter — so it correctly terminalizes every row of the node regardless of leg count (node_sessions.rs:94-98). For a fan-out node, that's all N stamped legs; there's no scenario where this fires against a node with pre-existing terminal legs (a node only reaches launch failure once, immediately after `stamp_fanout` marks all legs 'running'), so the missing status filter isn't a hazard here.
- The changed arm is scoped to `WorkflowEvent::NodeLaunchFailed` only (store.rs ~660-668); every other event (`TurnFinished`, etc.) still goes through `turn_session(event).or_else(...)` unchanged — turn-driven completion/failure paths are untouched.
- One-leg path: a one-leg node only ever has a single `workflow_run_node_sessions` row for that `node_row_id`, so the keyless whole-node update is equivalent to the old keyed update in that case — no behavior change for the non-fan-out path (confirmed against `a_one_leg_node_keeps_the_singular_start_effect` still passing per the implementer's 213/0 run).
- The regression test `a_failed_fan_out_launch_terminalizes_every_leg_it_stamped` stamps a real 3-leg fan-out via `stamp_fanout`, applies `NodeLaunchFailed`, and asserts all 3 rows are `Failed(NodeLaunchFailed)` with `completed_at` set — this pins the exact defect (pre-fix, the representative-session fallback would only match leg 0's session id and leave legs 1/2 `Running`), and the implementer's stated negative control (reverting the fallback fails at "leg 1") confirms the test actually discriminates.
- `ratchets.toml` +5 lines is accounted for by name to the genuinely new fix content in `store.rs`; not a ratchet game.
- Finding 2 (stale-artifacts sentence) confirmed reconciled — no contradiction remains between commit message and PR evidence, and the OpenAPI artifacts in the diff were already current as previously observed.

No new findings. This clears both items from the prior review.

pablonyx:
Head moved 3be8505a2 → 533596476 by (a) the rebase carry of r3's mechanical FE-PC-4 test rename and (b) one mechanical Ruff autofix on this rung (models_v2.py quoted self-annotation dropped, "Found 1 error (1 fixed, 0 remaining)"; ruff format "905 files already formatted"; CP suite re-run "42 passed"). Delta re-review CLEAN at 3be8505a2 carries — no behavioral content moved.

pablonyx:
Carry note (post-approval mechanical fixes, head e36032a4b → 1e24bec77): (1) rides the r4 AH-STORE-3 relocation (node_sessions under store/, rename-only) and the r2 cloud-SDK regen; (2) new commit at this rung commits `make cloud-client-generate` output for the F5 legs schema (WorkflowNodeLegV2 + legs, 12 insertions) so tier-2's "Verify generated cloud SDK artifacts are committed" step goes green — it was legitimately red here, not stale-base. store.rs ratchet lands at 1419 = exact line count. Local checks: boundary "passed", max-lines "passed", workflow domain+pane vitest "Tests 255 passed (255)". Delta re-review available on request.

pablonyx:
**Delta re-review — CLEAN** (pinned head `1e24bec77`)

Verified the claimed own-delta at this rung:

1. New commit `1e24bec77` "commit the regenerated cloud SDK types for per-leg fan-out" touches only `cloud/sdk/src/generated/openapi.ts` (+12), adding `WorkflowNodeLegV2 { prompt: string }` and an optional `legs` array on `WorkflowNodeV2` — matches the PR's DefinitionLeg/legs feature.
2. `git range-diff debf370d5..e36032a4b 89d3f37cf..1e24bec77` shows the rung-5 commits (feat F5+F6, the fail-redo review fix, the ruff lint fix) carry through with content unchanged — only the `ratchets.toml` reason-string differs, picking up the "+3: R4 CI fix relocates node_sessions under store/..." clause from the rebased base ahead of the rung-5 "+96"/"+5" clauses.
3. `store.rs` at head is exactly 1419 lines, matching `max_lines = 1419` in the ratchet.
4. No Cargo.lock changes in the delta.
5. Full diff `e36032a4b..1e24bec77` touches only: mod.rs/store.rs/node_sessions.rs (rebase-carried AH-STORE-3 rename, reviewed separately), openapi.ts (rung-5's own +12 plus 4 lines carried from the base's existing_workspace commit), and ratchets.toml reason strings. No hand-written Rust/TS logic changes at this rung beyond the ratchet bookkeeping.

No findings.

pablonyx:
**Rebase receipt (r5):** `1e24bec77` → `391cc9077`. Five conflicts, all from main's builder-v2 graph rewrite + workflow-UI merges landing under this rung: (1) ratchets.toml — r5's four edits (definition.rs entry, store.rs 1419, store_tests 1680, lifecycle_tests 1067) re-applied onto main's file; the actor.rs-entry deletion r5 carried was already true on main. (2) actor.rs — main's launch_model tests had been extracted to launch_tests.rs; kept the extraction, retargeted its import to r5's new launch.rs, threaded `legs: None` into its DefinitionNode fixture. (3) workflow-builder-draft.ts — union of main's graph model (edges/inputConnectedTo/replaceDefinition/coalesceKey editor) with r5's addLeg/removeLeg/updateLeg; **flag: leg edits do not pass a coalesceKey** (r5-authored behavior; whether leg editors should coalesce like main's updateNode-prompt is a delta-review question). (4) WorkflowBuilderSurface — main extracted WorkflowBuilderInspector; r5's three leg props now wired inside the Inspector's NodeCard call. (5) definition-v2.test.ts — both sides' new describe suites kept (main's definitionHeadId + r5's legs suite); import union. Cargo.lock taken-ours (the fix commit carried an incidental stale version bump). Substantive → delta review ordered. Full-ladder rebase onto main `8532f52b2` (2026-08-19 ~02:2xZ, executing Pablo's 'continue implementations end to end' order). Chain: #2016 ← #2017 ← #2057 (replaces #2018) ← #2020 ← #2034 ← #2041. Validation: focused vitest 52/52 (definition-v2 34, draft-legs 5, auto-advance-toast 13), full product-client typecheck green (all sibling dists rebuilt). Rust compile = CI proves (a codex-lane cargo held the machine's one build slot; per this lane's precedent). Delta reviews of the rebase carries are being ordered where substantive.

pablonyx:
**Delta reviewer (rebase composition) — CONFIRMED CLEAN**

Read-only delta review of the r5 rebase onto `main` (`8532f52b2`), verifying the five conflict resolutions carried r5's content without drifting main's builder-v2 graph rewrite. Method: isolate r5's own contribution over its real base `5fb3c0ce7a` (new r4 base) rather than the old-vs-new-r5 diff that conflates main's carry. All cites at `391cc9077`.

**1. ratchets.toml — CONFIRMED.** `git diff 5fb3c0ce7a 391cc9077 -- lints/anyharness/ratchets.toml` contains *only* r5's four intended edits and disturbs nothing of main's entries:
- `definition.rs` new entry `max_lines = 650` with the R5 DefinitionLeg/leg-validation reason (single entry, no duplicate).
- `store.rs` `max_lines = 1419` carrying both R5 reason clauses (`+96 … StartNodeLegs/stamp_fanout` and `+5: R5 review fix stamps every leg keyless on NodeLaunchFailed`).
- `store_tests.rs` `max_lines = 1680` (`+3 … legs field (None) through the three chain-node fixtures`).
- `lifecycle_tests.rs` `max_lines = 1067` (`+1 … legs field (None) through the agent_node fixture`).

**2. actor.rs / launch.rs / launch_tests.rs — CONFIRMED compile-correct by reading.**
- `launch_model` defined *only* in `launch.rs:23` (`pub(super)`); removed from actor.rs (grep across `anyharness/**/*.rs` finds one def).
- `actor.rs:28` `use super::launch::{launch_legs, launch_model};`; both used (`:281` launch_legs, `:298` launch_model). No `mod tests`/`cfg(test)` remains in actor.rs.
- `mod.rs` wires `mod launch;` and `#[cfg(test)] mod launch_tests;`. Shared symbols are visible: `actor.rs:38 pub(super) const DEFAULT_WORKFLOW_AGENT_KIND`, `:44 pub(super) struct WorkflowActorDeps`.
- Every `use` in launch.rs resolves to a symbol it references; store/transition seams exist (`store.rs:81 StartNodeLegs`, `:113 start_side_effect`, `:127 node_leg_count`, `:767 stamp_fanout`), and actor's `store::node_leg_count(...)` path resolves (pub fn).
- `launch_tests.rs:10-11` imports `DEFAULT_WORKFLOW_AGENT_KIND` from `super::actor` and `launch_model` from `super::launch`. Fixtures match head structs exactly: `DefinitionNode` (id/node_type/title/prompt/model/**legs: None**), `WorkflowDefinition` (schema_version/nodes/edges/inputs/doc_templates), `WorkflowRunNodeRecord` (all 20 fields).

**3. workflow-builder-draft.ts — CONFIRMED.** Union kept main's graph model intact: `connectNodes`/`removeEdge`/`connectInput`/`disconnectInput` (`:63-66`,`:287-302`), `replaceDefinition` (`:73`,`:339`), `updateNode` coalesceKey (`:229-231`), edges + `inputConnectedTo` throughout. Plus r5's `addLeg`/`removeLeg`/`updateLeg` with leg-0 mirroring (`:262-277`, `legIndex === 0 → { ...node, prompt, legs }`) and the collapse-to-scalar on removeLeg. Local aliases `type WorkflowNodeV2 = WorkflowNodeV2WithLegs` / `type WorkflowDefinitionV2 = WorkflowDefinitionV2WithLegs` do not collide — the bare SDK names were dropped from the import. `orderedNodes` is not imported at all (so not unused).

**4. Surface / Inspector / NodeCard — CONFIRMED.** Inspector wires the three leg props into the NodeCard call (`WorkflowBuilderInspector.tsx:92-94`). Surface has zero r5 changes (empty diff — main's extraction respected). `NodeCard` prop type declares `onAddLeg`/`onRemoveLeg`/`onUpdateLeg` (`WorkflowBuilderNodeCard.tsx:33-36`), all destructured and rendered.

**5. definition-v2.test.ts — CONFIRMED, no test lost vs either parent.** Both new suites present: `definitionHeadId` (`:405`, 4 tests, from main) and `validateDefinitionV2: node legs (ruling F5)` (`:443`, 7 tests, from r5). Shared base (`validateDefinitionV2` + `orderedNodes`) = 23 tests, identical across `origin/main`, old r5 `1e24bec77`, and `391cc9077`. Totals: main 27, old r5 30, new r5 34 = 23 + 4 + 7.

**6. Cargo.lock — CONFIRMED.** `git diff origin/main 391cc9077 -- Cargo.lock` is empty; the old r5 head's incidental `proliferate` version bump was dropped in favor of main's lock.

---

**Non-blocking note (item 7 — coalesceKey on updateLeg):** `updateLeg` (`workflow-builder-draft.ts:262`) passes no `WorkflowBuilderEditOptions`, unlike every other free-text editor in the builder — `updateNode` (`node:<id>:prompt|title`), `updateInput` (`input:<i>:name|description`), `updateDocTemplate` (`doc:<i>:body|slug`), `setTitle`/`setDescription`. Per `use-workflow-builder.ts:162-172`, a missing `coalesceKey` pushes a new undo-history entry on *every* keystroke (and nulls `lastEdit`). **User-visible consequence:** typing into a leg prompt records one undo step per character, so Ctrl+Z reverts a leg prompt one character at a time (inconsistent with every other text field, which reverts the whole typing burst), and the per-keystroke churn evicts older undo steps sooner under `HISTORY_LIMIT`. **Recommendation:** have `updateLeg` pass `{ coalesceKey: \`node:${nodeId}:leg:${legIndex}\` }` to match the other typing fields. `addLeg`/`removeLeg` correctly stay un-coalesced (discrete structural edits, like addNode/removeNode). The F5 `prompt === legs[0]` invariant holds on every keystroke regardless, so this is polish, not correctness.

pablonyx:
**Cascade receipt + endorsed follow-up applied (orchestrator).** New head: 16cce0b958c7f2e9980127793583975804f1b427 (was 391cc9077). Two parts:

1. **Re-parent** onto the recovered rung-1 head 9d90be8f0 (see #2016 note) — all four carried commits `=` in range-diff, base-only.
2. **The delta-review-endorsed one-liner (comment 5340170024) is now applied** as commit 16cce0b958: `updateLeg` passes `coalesceKey: \`node:${nodeId}:leg:${legIndex}\``, matching `updateNode`'s convention, plus two hook tests (same-leg typing burst = one undo entry; different legs = separate entries, leg-0 mirror unwinds with its own undo).

**Literals:** suite `Tests 17 passed (17)` (was 15). **Negative control:** key removed → `Tests 1 failed | 16 passed (17)`, failing test = "coalesces adjacent typing in one leg's editor into a single undo entry" with `expected 'd' to be ''` (per-keystroke undo, the exact pre-fix defect); restored → 17/17. Chain-tip validation: workflows client suites 39 files / 431 tests green, SDK 267 green, tsc exit 0. Reviewer of record: this discharges the endorsed follow-up — flag if you want a delta pass beyond the endorsement.

pablonyx:
**Merge-train cascade receipt (orchestrator).** Two-step re-parent after #2016 merged and the recovery lane advanced r2: first onto post-merge main (12518cb17), then onto the adopted r2 head b1c307b6bb (which adds a founder-authored ExistingWorkspace fix; my carried commits inside it are patch-identical). New head: 22931c9e20a07a9f1b9b7746909f20f22da9e034 (the import-shrink noted above is this PR's commit). All carried commits `=` in range-diff except one benign shrink: rung 5's commit no longer adds `model_validator` to the models_v2.py pydantic import because the adopted r2 fix already provides it (tree verified: 3 occurrences on the new head). No content change otherwise; standing reviews carry per the approve-carries precedent. CI riding the new heads is the compile/interaction verdict.

pablonyx:
**Scroll-physics fail on 22931c9e2 — known webkit strand, per-occurrence receipt.** Literal: `1 failed`, `[webkit] scroll-physics.spec.ts:697 prepend anchoring: older-history prepend keeps the reading row fixed`. Identical strand hit #2038 (unrelated forks PR) within the hour; this PR touches no transcript code. NOT content. Rerunning the failed job.

PR #2041, opened 2026-08-18
feat(workflows): per-leg redo behind the fail-redo verb (F2 reversed, rung 6)

pablonyx:
Rung 6 of the workflows follow-up ladder (frozen Follow-up Workflows ADR; ruling F2 REVERSED to per-leg). Stacked on rung 5 (#2034).

## BEHAVIOR CHANGE — whole-node redo of a RUNNING parallel node now disposes ALL live legs

Before this PR, a whole-node fail-redo of a running node disposed only the representative session (Ruling L). Once rung 5 mints N sessions per parallel node, that behavior would leave the other N-1 leg agents alive against a replaced node. Now the store resolves `DisposeSessionsThenStartLegs` carrying EVERY running leg's session. This matches F2's ruled `disposed_session_ids: Vec` bulk shape; orchestrator concurred (dispose-all requirement + negative control, both landed here). One-leg and adhoc nodes keep the byte-identical single-session path.

## What this adds

- `WorkflowCommand::FailAndRedo` gains optional `legIndex` (runtime DTO camelCase, SDK type, product-client facade plumbing). Absent = whole-node redo, the historical behavior.
- `Transition::RedoLeg { node_row_id, leg_index, disposed_session_id }`: in-place relaunch of ONE leg — no replacement row; siblings' ledger rows untouched; the addressed row resets to `running` in the same commit so the fan-in gate holds the node for it.
- Legality wall shared with whole-node redo: a one-leg node REJECTS per-leg redo ("a one-leg node takes a whole-node redo"); out-of-range/negative indexes reject; the leg's session is disposed only when that leg is still Running.
- Side effects `StartLeg` / `DisposeSessionThenStartLeg` / `DisposeSessionsThenStartLegs`; launch helpers relocated from actor.rs into `live/workflows/launch.rs` (actor.rs 562 lines).

## Explicitly NOT included

- **Leg-picker UI** — rung 7.
- **Per-leg prompt editing on redo** — per-leg redo re-renders the leg's AUTHORED prompt from the frozen definition; the DTO `prompt` field stays whole-node-only. New scope, parked as a founder question (not auto-scoped to rung 7).

## Verification (local, machine-slot arbitrated)

- `cargo test -p anyharness-lib workflow` → `test result: ok. 220 passed; 0 failed` (213 before this rung).
- Negative controls (each applied, run, reverted; tree pristine between):
  1. `reset_leg_running_tx` call severed → `per_leg_redo_resets_exactly_one_ledger_row_and_leaves_siblings` FAILED (exit 101). Disclosure: the plan predicted a second test (`a_redone_leg_makes_the_node_wait...`) would also fail; it stays green because `stamp_leg_relaunch` re-stamps the row to running regardless — the control still proves the call load-bearing.
  2. Range check `>=` weakened to `>` → `per_leg_redo_rejects_an_out_of_range_or_negative_index` FAILED.
  3. Forced `disposed_session_id = None` → `per_leg_redo_of_a_running_leg_disposes_that_legs_session` FAILED.
  4. Dispose-all re-filtered to representative-only → `a_whole_node_redo_of_a_running_parallel_node_disposes_every_live_leg` FAILED with `left: ["sess-0"] right: ["sess-0", "sess-1", "sess-2"]` — the exact N-1 orphan receipt.
- `make sdk-build`: openapi diff is exactly the `legIndex` property; `tsc` exit 0; post-regen `cargo test -p anyharness-lib contract_fixture` → `ok. 8 passed; 0 failed`.
- Workflow domain+pane vitest at this head's stack: `Tests 255 passed (255)`; pane facade suite `Tests 28 passed (28)`.
- Boundary + max-lines checks pass; store.rs is exactly 1549 lines = its ratchet.


pablonyx:
**Review of rung-6 delta** (`1e24bec77..cec327245`), pinned to head `cec327245`.

**BLOCKING** — `RedoLeg`'s "siblings untouched" invariant breaks if the leg's relaunch fails to start.

`perform_side_effect`'s `StartLeg` arm (anyharness/crates/anyharness-lib/src/live/workflows/actor.rs:206-220) routes a launch error through `feed_launch_failure`, which fires the generic `WorkflowEvent::NodeLaunchFailed` for `node_row_id` (actor.rs:241-243) regardless of whether the launch was whole-node or per-leg. `on_node_launch_failed` (transition.rs:839-856) unconditionally emits `Transition::FailNode` for that row, and `finished_leg_of` (store/node_sessions.rs:177-179) maps `FailNode` to a **keyless** `mark_leg_terminal_tx` call (store.rs:754-769, `session = None` on `NodeLaunchFailed`) — which runs `UPDATE workflow_run_node_sessions SET status = 'failed' ... WHERE node_row_id = ?1` with **no leg_index filter** (node_sessions.rs:105-116).

So: per-leg redo of leg 1 on a 3-leg node, where leg redo's launch throws (session-mint error, harness unavailable, etc.) → every sibling leg's ledger row — including already-`Done` or still-`Running` ones — gets stamped `failed`, and the node itself flips to `WorkflowNodeStatus::Failed` / the run to `WorkflowRunStatus::Failed`. This directly contradicts the PR's own "siblings untouched" claim and the `RedoLeg` transition's careful single-row update in the happy path (store.rs:526-587).

This code path is pre-existing (from rung 5, written when the only launch-failure caller was whole-node `StartNode`/`StartNodeLegs`), but rung 6 is what newly routes a genuinely single-leg operation through it via `StartLeg` → `feed_launch_failure`. There's no test exercising "per-leg redo's relaunch fails to start" — `store_fanin_tests.rs`/`transition_fanin_tests.rs` cover the happy-path per-leg redo and the whole-node dispose-all change, but not this failure interaction.

Fix shape: either give `RedoLeg`'s failure path its own event/transition that terminalizes only the addressed leg (mirroring how `RedoLeg` itself scopes the happy path), or have `feed_launch_failure`/`on_node_launch_failed` carry the leg index when the effect that failed was `StartLeg`/`DisposeSessionThenStartLeg` so `finished_leg_of` can target one row.

---

Everything else reviewed clean:
- Legality wall (`fail_redo_legal_pause` shared, one-leg rejection, range check, running-only dispose) is correct and well-tested via the negative controls in the PR body.
- `store.rs` `RedoLeg` apply: single ledger row reset + node revive + run un-terminalize, siblings untouched — correct.
- Whole-node dispose-all collection (`DisposeSessionsThenStartLegs`) correctly filters `Running` legs off the **old** node before the replacement row is minted — correct order, matches the `a_whole_node_redo_of_a_running_parallel_node_disposes_every_live_leg` test.
- `actor.rs`→`launch.rs` relocations are moves, not rewrites, spot-checked.
- DTO/SDK plumbing (`workflow_run_commands.rs`, generated SDK, `use-workflow-pane.ts`) — `legIndex` threaded only when supplied, no default-value smuggling.
- `ratchets.toml` changes are exact byte-count bumps with reasons, no unrelated loosening.
- No `Cargo.lock` in the diff.

pablonyx:
## Review of record at cec3272457d94d4c58bea311bbaaa75a8f3e5d57

Fresh independent reviewer (not the implementer). Reviewed the pinned head against base `wf-followup/r5-per-leg-fanout` by static reading only (no cargo/rustc/make per machine limits). Per-focus-area findings below.

### 1. Ledger transition correctness — PASS
- `Transition::RedoLeg` apply arm ([store.rs](anyharness/crates/anyharness-lib/src/domains/workflows/store.rs:518)) revives the shared node scalar row (`status='running'`, `failure_code=NULL`, `first_turn_finished_at=NULL`), un-fails the run, then resets **exactly one** ledger row via `reset_leg_running_tx` scoped `WHERE node_row_id=?1 AND leg_index=?2` ([node_sessions.rs](anyharness/crates/anyharness-lib/src/domains/workflows/store/node_sessions.rs:65)). Siblings are byte-untouched. `stamp_leg_relaunch` refreshes the node scalar session/prompt columns only for `leg_index == 0` (the representative), sibling relaunches leave them alone ([store.rs](anyharness/crates/anyharness-lib/src/domains/workflows/store.rs:897)).
- `leg_index` plumbs cleanly: DTO → `WorkflowCommand::FailAndRedo.leg_index` → `on_redo_leg` → `Transition::RedoLeg.leg_index` → `ResolvedSideEffect::StartLeg{leg_index}` → `relaunch_leg`/`launch_one_leg`. Test `per_leg_redo_resets_exactly_one_ledger_row_and_leaves_siblings` asserts leg 2 Running while legs 0/1 stay Done with intact `completed_at`.

### 2. Index validation — PASS
- `on_redo_leg` rejects `leg_index < 0 || leg_index >= leg_count` at the transition boundary ([transition.rs](anyharness/crates/anyharness-lib/src/domains/workflows/transition.rs:768)), fail-closed as `Decision::Illegal`. The `>=` is the correct class boundary (indices are `0..leg_count`). One-leg node also rejected before the range check (`leg_count <= 1`). `launch_one_leg` has a defense-in-depth `.get(leg_index).ok_or_else(...)` that fails to `NodeLaunchFailed` if a bad index ever reaches the actor.

### 3. Session disposal — PASS
- Per-leg redo of a **running** leg disposes exactly that leg's session: `on_redo_leg` finds the leg matching `leg_index` **and** `status == Running`, carrying its `session_id` into `DisposeSessionThenStartLeg` ([transition.rs](anyharness/crates/anyharness-lib/src/domains/workflows/transition.rs:790)); a failed leg carries `None`.
- Whole-node redo of a **running parallel** node disposes ALL live leg sessions: the `parallel` branch collects every `Running` leg's `session_id` into `DisposeSessionsThenStartLegs` and the actor disposes each before `StartNodeLegs` ([store.rs](anyharness/crates/anyharness-lib/src/domains/workflows/store.rs:510), [actor.rs](anyharness/crates/anyharness-lib/src/live/workflows/actor.rs:170)). This is the real fix for the latent N-1 orphan bug; one-leg/adhoc keep the byte-identical single-session `DisposeThenStart`/`StartNode` path. Test `a_whole_node_redo_of_a_running_parallel_node_disposes_every_live_leg` asserts `["sess-0","sess-1","sess-2"]` — teeth confirmed (representative-only would give `["sess-0"]`).

### 4. DTO / OpenAPI / SDK / client — PASS
- `legIndex` is OPTIONAL (`#[serde(default)] leg_index: Option<i64>`); absent = whole-node ([workflow_run_commands.rs](anyharness/crates/anyharness-lib/src/api/http/workflow_run_commands.rs:23)). Added `#[serde(rename_all="camelCase")]` is safe (only `prompt` besides, unchanged).
- Regen artifacts match the Rust source: `openapi.json`/`openapi.ts` gain exactly `legIndex: integer,null / int64`; hand SDK `workflow-runs-v2.ts` gains `legIndex?: number`. Client facade `failRedo(nodeRowId, prompt?, legIndex?)` builds the request conditionally and is backward-compatible with existing callers (still `{}` / `{prompt}`); both new branches covered by vitest.

### 5. Negative-control quality — PASS (with the disclosed 1-of-2 caveat, verified accurate)
- NC2 (`>=`→`>`): `per_leg_redo_rejects_an_out_of_range_or_negative_index` uses `Some(2)` on a 2-leg node; `2 > 2` is false so the guard would pass and yield a `RedoLeg` instead of `Illegal` — real teeth.
- NC3 (`disposed_session_id=None`): `per_leg_redo_of_a_running_leg_disposes_that_legs_session` asserts `Some("sess-b")` — real teeth.
- NC4 (dispose refiltered to representative): asserts the full 3-session vec — real teeth.
- NC1 (sever `reset_leg_running_tx`): `per_leg_redo_resets_exactly_one_ledger_row_and_leaves_siblings` inspects `applied.state` immediately after `apply_transition` (before any relaunch stamp), so leg 2 would remain Failed and the `Running` assert fails — real teeth. The disclosed second predicted test (`a_redone_leg_makes_the_node_wait...`) genuinely stays green because it explicitly calls `stamp_leg_relaunch`, whose `upsert_leg_at_tx` re-stamps the row to running regardless of the reset. The honesty note is **accurate**, and NC1 coverage remains sufficient via test 1.

### 6. Scope honesty — PASS
- No per-leg prompt editing anywhere. `on_redo_leg` never reads the DTO `prompt`; `launch_one_leg` renders from `leg_prompts(node, definition)` — the **frozen authored** prompt from the definition snapshot ([launch.rs](anyharness/crates/anyharness-lib/src/live/workflows/launch.rs:43)). Matches the ratified F2 semantics; the parked per-leg-editing founder question is correctly not implemented.
- Ratchet bumps are honest and match the diff (store.rs +130→1549, transition.rs +82→856, plus the `leg_index:None` fixture threads). actor.rs is 562 lines (ratchet 603, unchanged) — the `link_start_and_dispatch`/`render_node_envelope` relocation to launch.rs is a genuine no-behavior-change extraction; PR body's "actor.rs 562 lines" is accurate.
- PR body claims match the diff throughout (side-effect variants, legality wall extraction `fail_redo_legal_pause`, F6 unchanged).

### CI note (attribution not mine)
The r6 diff does **not** obviously explain the tier-2 "Workflow definition lifecycle" red: the client `failRedo` change is backward-compatible and the regen artifacts (`openapi.json`/`openapi.ts`/`workflow-runs-v2.ts`) are internally consistent with the Rust DTO. The r5 base tip (`1e24bec77`) is itself a "commit the regenerated cloud SDK types" fix, so a plausible read is an **inherited/uncommitted cloud-SDK regen drift** on the stack rather than a defect introduced here — but that is the implementer's to confirm.

---

**VERDICT: NO BLOCKING DEFECTS.** All six focus areas PASS. Ledger single-row reset, index fail-closed validation, per-leg and dispose-all session disposal, optional-legIndex DTO/regen surface, and the four negative controls all hold on static reading. The disclosed NC1 1-of-2 sensitivity is accurate and coverage is still sufficient. Comment-only review — cargo not run per machine limits; CI remains the execution gate. Recommend the implementer attribute the tier-2 lifecycle red (likely inherited SDK-regen drift) before merge.

pablonyx:
## Adversarial verify of the BLOCKING finding — CONFIRMED

Follow-up to my six-area review of record ([issuecomment-5323647768](https://github.com/proliferate-ai/proliferate/pull/2041#issuecomment-5323647768)) and the independent BLOCKING finding by reviewer 6e86ad45 ([issuecomment-5323442617](https://github.com/proliferate-ai/proliferate/pull/2041#issuecomment-5323442617)). My six areas passed the happy paths (areas 1 and 3); this finding lives on the launch-**failure** path, which they did not exercise. I traced it end-to-end at head `cec3272457d94d4c58bea311bbaaa75a8f3e5d57`.

**VERDICT: CONFIRMED.** A `StartLeg` relaunch that fails to start corrupts every sibling ledger row of the node. Every hop of the alleged chain holds:

1. **`StartLeg` failure → `feed_launch_failure`** — [actor.rs:206-219](anyharness/crates/anyharness-lib/src/live/workflows/actor.rs:206): the `StartLeg` arm calls `relaunch_leg(...)`; on `Err(error)` it routes to `self.feed_launch_failure(state, &node_row_id, error)`. The `leg_index` is dropped here — it is not passed on.
2. **`feed_launch_failure` fires generic `NodeLaunchFailed`** — [actor.rs:241-243](anyharness/crates/anyharness-lib/src/live/workflows/actor.rs:241): constructs `WorkflowEvent::NodeLaunchFailed { node_row_id }`. The event type carries **only** `node_row_id` ([transition.rs:92](anyharness/crates/anyharness-lib/src/domains/workflows/transition.rs:92)) — no leg index exists on it, so the single-leg scope is irrecoverable downstream.
3. **`NodeLaunchFailed` → `Transition::FailNode`** — [transition.rs:839-856](anyharness/crates/anyharness-lib/src/domains/workflows/transition.rs:839): `on_node_launch_failed` sees a Running non-adhoc node and unconditionally returns `Transition::FailNode { node_row_id, code: NodeLaunchFailed }`.
4. **`FailNode` apply** — [store.rs:447-459](anyharness/crates/anyharness-lib/src/domains/workflows/store.rs:447): `fail_node` flips the node row to `failed` and the run to `Failed`/`completed_at`.
5. **`finished_leg_of(FailNode)` → keyless stamp** — [node_sessions.rs:182-184](anyharness/crates/anyharness-lib/src/domains/workflows/store/node_sessions.rs:182) returns `Some((node_row_id, Failed(code)))`; the shared block at [store.rs:754-769](anyharness/crates/anyharness-lib/src/domains/workflows/store.rs:754) forces `session = None` for `NodeLaunchFailed` and calls `mark_leg_terminal_tx(tx, node_row_id, None, Failed, ts)`.
6. **Keyless SQL hits ALL rows** — [node_sessions.rs:105-116](anyharness/crates/anyharness-lib/src/domains/workflows/store/node_sessions.rs:105): the `None` arm runs `UPDATE workflow_run_node_sessions SET status = ?2, completed_at = ?3 WHERE node_row_id = ?1` — **no `leg_index` filter and no `status` filter**.

### The three sub-claims, judged

**(a) Done sibling rows overwritten — CONFIRMED.** After a per-leg redo, `RedoLeg` reset only the addressed row ([store.rs:526-548](anyharness/crates/anyharness-lib/src/domains/workflows/store.rs:526)); the Done siblings kept their `Done` status and completion times. Step 6's keyless UPDATE has no status guard, so those Done rows are overwritten to `failed` and their original `completed_at` is clobbered with the failure timestamp — their success receipts are lost. Contrast [node_sessions.rs:127-141](anyharness/crates/anyharness-lib/src/domains/workflows/store/node_sessions.rs:127) `cancel_all_run_legs_tx`, which deliberately guards `AND status = 'running'` to spare terminal legs; `mark_leg_terminal_tx`'s keyless arm has no such guard.

**(b) Running siblings terminalized without disposal — CONFIRMED (orphan class).** Redo one leg of a 3-leg *running* node: siblings 0 and 1 hold live sessions. Step 6 stamps them `failed`, but `FailNode` emits no `DisposeSessions` side effect ([store.rs:447-459](anyharness/crates/anyharness-lib/src/domains/workflows/store.rs:447) — the arm sets no `side_effect`). The two agents keep running against a node the ledger calls failed. This is the *exact* N-1 orphan class this PR set out to eliminate for whole-node redo (`DisposeSessionsThenStartLegs`), reintroduced on the per-leg failure path.

**(c) Node-level failure under F1 — BY DESIGN, not the defect.** Per the agreed judging split: a leg that cannot relaunch failing the node is defensible under ruled F1 (a leg failure fails the node). The defect core is the **sibling-row corruption** in (a) and (b), not the node flip.

### Why rung 6 is what newly breaks this

The keyless stamp is *correct* in its rung-5 origin: `a_failed_fan_out_launch_terminalizes_every_leg_it_stamped` ([store_fanin_tests.rs](anyharness/crates/anyharness-lib/src/domains/workflows/store_fanin_tests.rs)) proves it, but there **all N legs are one fresh fan-out cohort** — every row was just stamped and is launching together, so terminalizing all of them is right. Rung 6's `StartLeg` is the first single-leg operation routed through the same `feed_launch_failure` → keyless path, where the siblings are *not* a cohort: they are Done or independently Running. No test exercises "per-leg redo's relaunch fails to start," so the interaction went uncaught.

### Reachability

`relaunch_leg`/`launch_one_leg` fails on session-mint error, workspace-not-found, envelope render failure, `link_workflow_columns` error, `stamp_leg_relaunch` error, or a hard dispatch error ([launch.rs](anyharness/crates/anyharness-lib/src/live/workflows/launch.rs) `launch_one_leg`). These are real runtime conditions (harness unavailable, DB contention), so the path is genuinely reachable, not theoretical.

### Fix shape (agreeing with the finder)

Scope the failure stamp to the addressed leg: either give the `StartLeg`/`DisposeSessionThenStartLeg` failure its own event/transition carrying `leg_index` so `finished_leg_of` targets one row, or have `feed_launch_failure` thread the leg index when the failed effect was single-leg. The addressed leg alone should terminalize (and, per F1, the node may still fail), leaving Done siblings' receipts intact and Running siblings either untouched or disposed-then-failed — never keyless-stamped.

---

**BLOCKING confirmed.** Comment-only review; static read only, no cargo run (machine limits), CI remains the execution gate. The implementer fixes forward; a delta pass follows. Everything in my six areas remains PASS — this defect is strictly on the launch-failure path outside that coverage, which is why the two reviews are complementary, not contradictory.

pablonyx:
ORCHESTRATOR STATUS (16:2xZ): the validation window granted at 09:53Z (cargo test -p anyharness-lib workflow at fix c115e2966 + executed negative control) appears undrained — no push landed on this branch. The implementer lane may be wedged; my workspace-MCP outbound is down (token TTL) so I cannot probe or re-grant by message right now. STANDING ORDER preserved for whoever picks this up (implementer lane, on your next wake): the slot-grant terms from queueSeq 51 stand — verify pgrep-empty first, one bounded ~25-min window, push receipts to this PR, no merge. #2041 is documented as a STRAGGLER for the next release train; it does not block tonight's 0.4.17 dispatch.

pablonyx:
## Validation attempt at fix c115e29668f8673cc4c98f5092bc381a7a60ecbe — DID NOT PUSH

Executing the standing order from [issuecomment-5330952126](https://github.com/proliferate-ai/proliferate/pull/2041#issuecomment-5330952126) (slot-grant terms from queueSeq 51: verify pgrep-empty first, bounded window, push receipts, no merge).

**Pre-checks (all passed):**
- `git cat-file -e c115e2966^{commit}` — commit exists.
- Parent of `c115e2966` is `cec3272457d94d4c58bea311bbaaa75a8f3e5d57`, which exactly matches PR #2041's current `headRefOid` (`gh pr view 2041 --json headRefOid`).
- `pgrep -x cargo; pgrep -x rustc` — empty at window start (16:28:22Z).
- `sysctl vm.swapusage kern.memorystatus_vm_pressure_level` at window start: `free = 1460.62M`, pressure level `1`.

**POSITIVE run — FAILED TO COMPILE (not a test failure):**

```
cd /tmp/wf-2041-validate  # worktree at c115e2966
nice -n 10 cargo test -p anyharness-lib workflow
```

```
error[E0433]: cannot find type `WorkflowNodeFailureCode` in this scope
   --> anyharness/crates/anyharness-lib/src/domains/workflows/store_fanin_tests.rs:629:35
    |
629 |         WorkflowLegStatus::Failed(WorkflowNodeFailureCode::NodeLaunchFailed)
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared type `WorkflowNodeFailureCode`

error[E0433]: cannot find type `WorkflowNodeFailureCode` in this scope
   --> anyharness/crates/anyharness-lib/src/domains/workflows/store_fanin_tests.rs:687:35
    |
687 |         WorkflowLegStatus::Failed(WorkflowNodeFailureCode::NodeLaunchFailed)
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared type `WorkflowNodeFailureCode`

error: could not compile `anyharness-lib` (lib test) due to 2 previous errors; 6 warnings emitted
```

**Root cause (verified, not a stale-cache artifact):** `store_fanin_tests.rs`'s `use super::model::{...}` block (lines 14-16) imports `WorkflowLegStatus, WorkflowNodeStatus, WorkflowNodeType, WorkflowRunStatus` but not `WorkflowNodeFailureCode`. Pre-existing line 425 correctly uses the fully-qualified `super::model::WorkflowNodeFailureCode::NodeLaunchFailed`. The two *new* assertions this fix adds (lines 629, 687 — part of the new fan-in coverage) use the bare unqualified `WorkflowNodeFailureCode`, which doesn't resolve without adding it to the import list (or qualifying it like line 425 does). The production-code fix in `store.rs`/`transition.rs`/`actor.rs` was not evaluated because the crate never finished compiling.

**NEGATIVE CONTROL — not run.** Skipped per the mission's own stop condition: step 3 already deviated from the expected ~222-passing result, so no negative control or push was attempted.

**Machine guard (bounded window, ~2 min actual):**
- Start 16:28:22Z: swap free 1460.62M, pressure level 1.
- End 16:30:12Z: swap free 1476.62M, pressure level 1.
- `pgrep -x cargo; pgrep -x rustc` empty at both ends; no lingering processes.
- Worktree `/tmp/wf-2041-validate` created at `c115e2966`, removed after the run (`git worktree remove --force`), confirmed absent from `git worktree list`.

**Outcome: nothing pushed to `wf-followup/r6-per-leg-redo`.** Head remains `cec3272457d94d4c58bea311bbaaa75a8f3e5d57`. The fix needs a one-line import addition (`WorkflowNodeFailureCode` into the `use super::model::{...}` block in `store_fanin_tests.rs`, or qualify the two call sites like line 425 does) before this can compile and be re-validated. Not authorized to patch and push a modified commit under this grant — flagging for the implementer lane per "implementer fixes forward" convention in this thread.

pablonyx:
Compile cure landed and validated. New head: `1da86c760f54b97375e46d0b69a9cf408aa6e36c` (on top of the unpushed fix `c115e29668f8673cc4c98f5092bc381a7a60ecbe`).

**Cure diff** (`store_fanin_tests.rs`, matches the file's existing import style):
```diff
 use super::model::{
-    WorkflowLegStatus, WorkflowNodeStatus, WorkflowNodeType, WorkflowRunStatus,
+    WorkflowLegStatus, WorkflowNodeFailureCode, WorkflowNodeStatus, WorkflowNodeType,
+    WorkflowRunStatus,
 };
```

**Positive** — `nice -n 10 cargo test -p anyharness-lib workflow` on `1da86c760`:
```
test result: ok. 222 passed; 0 failed; 0 ignored; 0 measured; 1940 filtered out; finished in 29.71s
```

**Negative control** — `git checkout cec327245` (the direct parent of the fix, i.e. the pre-fix PR head), same filtered run:
```
test result: ok. 220 passed; 0 failed; 0 ignored; 0 measured; 1940 filtered out; finished in 26.80s
```
Note on the shape of this result: it is 0 failed, not a literal assertion failure, because `cec327245` predates the fix entirely — the two new regression tests (`a_failed_per_leg_relaunch_terminalizes_only_the_addressed_leg`, `a_failed_per_leg_relaunch_cancels_and_disposes_running_siblings`) simply do not exist there (222 vs 220, diffed and confirmed by test name). I also attempted a stricter control — keep the new tests, revert only the behavioral files (`store.rs`, `store/node_sessions.rs`, `transition.rs`, `live/workflows/actor.rs`) to their `cec327245` state — to try to force a true assertion failure. That does not compile:
```
error[E0559]: variant `WorkflowEvent::NodeLaunchFailed` has no field named `leg_index`
   --> anyharness-lib/src/domains/workflows/store_tests.rs:592:9
```
This is expected and actually confirms the finding's name: "keyless" sibling terminalization means the `NodeLaunchFailed` event previously carried no leg key at all (`{ node_row_id: String }`, no `leg_index`). The fix adds that field. Since the old event type structurally cannot express "this specific leg's launch failed," there is no way to construct an old-behavior run of the new tests — the only two possible outcomes are "coverage absent" (pre-fix) or "coverage present and passing" (post-fix), which is exactly what both runs show.

**Machine guard** (checked before every cargo invocation): `pgrep -x cargo`/`pgrep -x rustc` empty each time; `sysctl kern.memorystatus_vm_pressure_level` stayed at 1 (once briefly 2) throughout, never reached the abort threshold of 3; `sysctl vm.swapusage` free stayed ≥1.4G throughout, never near the 400M floor. All builds ran `nice -n 10`, one at a time.

Head confirmed on GitHub: `gh pr view 2041` reports `headRefOid: 1da86c760f54b97375e46d0b69a9cf408aa6e36c`.

This completes the standing order from #2016 comments 5330952126 and 5331150849.

pablonyx:
**Delta review A verdict: BLOCKING** — pinned to head `1da86c760f54b97375e46d0b69a9cf408aa6e36c`.

Scope reviewed: `cec327245..1da86c760` (two commits: `c115e2966` behavioral fix + `1da86c760` import cure) against the "keyless sibling terminalization on relaunch failure" finding ([5323442617](https://github.com/proliferate-ai/proliferate/pull/2041#issuecomment-5323442617), confirmed [5323790059](https://github.com/proliferate-ai/proliferate/pull/2041#issuecomment-5323790059)).

**The behavioral cure is correct and well-tested.** `WorkflowEvent::NodeLaunchFailed` now carries `leg_index: Option<i64>` (transition.rs). `actor.rs`'s `StartLeg` arm threads `Some(leg_index)` through `feed_launch_failure`; `DisposeSessionThenStartLeg` re-enters as `StartLeg` so it inherits the same scoping, and `DisposeSessionsThenStartLegs`/whole-node `StartNode`/`StartNodeLegs` correctly keep `None` (cohort semantics, unchanged). In `store.rs`, a `Some(leg_index)` failure now calls the new `fail_leg_and_cancel_running_siblings_tx` (node_sessions.rs) — the addressed leg alone takes the failure stamp, running siblings are stamped `Cancelled` and disposed via a new `DisposeSessions` side effect off `Transition::FailNode`, and Done siblings are left untouched (no status filter match). The two new tests (`a_failed_per_leg_relaunch_terminalizes_only_the_addressed_leg`, `a_failed_per_leg_relaunch_cancels_and_disposes_running_siblings`) assert exact per-row status/`completed_at` and the exact `DisposeSessions{session_ids}` set — they would catch a regression back to the keyless stamp. Delta is otherwise tightly scoped: only `store.rs`, `node_sessions.rs`, `transition.rs`, `store_fanin_tests.rs`, `store_tests.rs`, `transition_tests.rs`, `ratchets.toml` change, all directly serving the cure, its tests, or the mechanical `leg_index: None` threading through pre-existing fixtures.

**Blocking defect (new, introduced by this delta, currently failing live CI on this exact head):** the `ratchets.toml` entry added for `store_fanin_tests.rs` records `max_lines = 691`, but the compile-cure commit (`1da86c760`) added one more import line to that file after the ratchet was written in `c115e2966`, so the file is actually **692** lines — one over its own declared cap. Verified by running `python3 scripts/check_max_lines.py` at `1da86c760`:
```
anyharness/crates/anyharness-lib/src/domains/workflows/store_fanin_tests.rs: PROD-SIZE-1
  found: 692 lines (max 600, ratcheted at 691 in lints/anyharness/ratchets.toml)
```
This matches the live "Repo shape checks" job on the PR, which is currently **fail** (`gh pr checks 2041`). The two cargo-test validation receipts (5331150849, 5331274189) only ran `cargo test -p anyharness-lib workflow`, not the repo-shape lint, so this slipped through. Fix is a one-line bump (`max_lines = 692`) — trivial, but it is a real, reproducible CI failure on this head and blocks merge as-is.

pablonyx:
## Delta review B verdict (pinned to `1da86c760f54b97375e46d0b69a9cf408aa6e36c`)

**Verdict: BLOCKING** — not the keyless-termination defect (that cure holds, verified below), but a fresh, concrete, CI-verified regression in the exact commits under review.

### Refutation attempt on the cured defect (5323442617 / confirmed 5323790059) — FAILED

Traced every emission site of `WorkflowEvent::NodeLaunchFailed` at this head: exactly one production construction site exists (`actor.rs:244`, inside `feed_launch_failure`), and every call site of `feed_launch_failure` threads `leg_index` correctly:
- `StartNode`/`StartNodeLegs` (actor.rs:200) → `None` — the whole-cohort fan-out launch, correct to stay keyless since every leg is freshly minted together.
- `StartLeg` (actor.rs:218) → `Some(leg_index)`.
- `DisposeSessionThenStartLeg` collapses to `StartLeg` (actor.rs:181-193) before it can fail, so it also carries `Some(leg_index)` — no bypass route.

`apply_transition`'s `FailNode` arm (store.rs:447-481, store.rs:773-805) now branches on the *event's* `leg_index`, not just the transition:
- `Some(leg_index)` → `fail_leg_and_cancel_running_siblings_tx` (node_sessions.rs:117-142): the addressed row alone takes `Failed`; only siblings still `Running` flip to `Cancelled` (Done/Cancelled/ForcedUnload/already-Failed siblings are excluded by the `status = 'running'` filter and keep status + `completed_at` byte-identical); the same `FailNode` arm collects those Running siblings' `session_id`s — explicitly excluding the addressed leg_index — into `ResolvedSideEffect::DisposeSessions`, so a session the redo itself already disposed is never double-disposed.
- `None` → unchanged keyless `mark_leg_terminal_tx`, correctly scoped to the fan-out-cohort case it was written for.
- Adhoc nodes structurally cannot reach `StartLeg`: `on_redo_leg` rejects any node with `leg_count <= 1` ("a one-leg node takes a whole-node redo"), and adhoc nodes never fan out, so they're always leg_count == 1. The adhoc `AdhocTurn` launch-failure arm (which still ignores leg_index) is unreachable for this defect class.
- New tests `a_failed_per_leg_relaunch_terminalizes_only_the_addressed_leg` and `a_failed_per_leg_relaunch_cancels_and_disposes_running_siblings` assert concrete DB outcomes (per-row status, `completed_at` presence/absence, the exact `DisposeSessions` session-id set, run status) tied 1:1 to the finding's two sub-claims — not tautological mirrors of the implementation.
- The unrelated prior finding ("cancel terminalizes every running leg... adhoc rows included") is untouched by this diff — `cancel_all_run_legs_tx` has zero changes in this delta.

I could not construct a scenario, at any hop from `StartLeg`/`DisposeSessionThenStartLeg` failure through to the ledger commit, where a sibling leg ends up keylessly terminalized or mis-keyed. The cure holds.

### New finding, in-scope for this exact delta: CI is currently red on `1da86c760`

The compile-cure commit (`1da86c760`) adds one import line to `store_fanin_tests.rs` (wraps the `use super::model::{...}` list from 2 lines to 3 to bring in `WorkflowNodeFailureCode`), taking the file from 691 → 692 lines — but does not bump its `[[max_lines]] = 691` entry in `lints/anyharness/ratchets.toml` (that entry was added by the prior commit `c115e2966` to match *that* commit's exact count, 691, correctly at the time). `scripts/check_max_lines.py` — wired into CI as the "Repo shape checks" job (`.github/workflows/ci.yml:35`) — enforces `line_count > entry.max_lines` as a hard failure, not just "must not grow past 600."

Confirmed two independent ways:
1. Ran `python3 scripts/check_max_lines.py` locally against the `1da86c760` tree: exit code 1, `found: 692 lines (max 600, ratcheted at 691 in lints/anyharness/ratchets.toml)`.
2. Live GitHub Actions: `gh pr checks 2041` shows **"Repo shape checks" = fail**; job 95791869441 (run 32161702326) has `conclusion: "failure"`, `completed_at: 2026-08-18T16:42:28Z` — one second after the compile-cure push (comment [5331274189](https://github.com/proliferate-ai/proliferate/pull/2041#issuecomment-5331274189) at 16:42:29Z) — with the identical diagnostic in the job log.

Fix is a one-line ratchet bump (`max_lines = 691` → `692` with an appended reason clause), but the SHA pinned for this review is not actually CI-green — it fails a mandatory check.

### Scope check

Everything else in the 8-file delta (`store.rs`, `node_sessions.rs`, `transition.rs`, `actor.rs` = the cure; `store_fanin_tests.rs`, `store_tests.rs`, `transition_tests.rs` = the 2 new regression tests plus threading the new `leg_index` field through pre-existing fixtures; `ratchets.toml` = line-count housekeeping, modulo the one stale entry above) is in-scope and load-bearing. No drive-by refactors, no unrelated behavior changes found.

pablonyx:
Line-neutral cure pushed: head 5151051ba2d890528bafbbcb76ac54bd51a2d113. store_fanin_tests.rs back to 691 lines (import reverted to single line, the two refs fully qualified matching line 426); scripts/check_max_lines.py literal: 'Max-lines check passed'. Filtered suite re-verified at this tree: 'test result: ok. 222 passed; 0 failed; 0 ignored; 0 measured; 1940 filtered out; finished in 21.23s'. Addresses both reviewer BLOCKING items (5331331374, 5331422618) whose behavioral verdicts were clean; reviewers being re-pinned to this head.

pablonyx:
**Delta review A verdict: CLEAN** — updated for head `5151051ba2d890528bafbbcb76ac54bd51a2d113` (supersedes my prior BLOCKING at `1da86c760f54b97375e46d0b69a9cf408aa6e36c`, comment [5331331374](https://github.com/proliferate-ai/proliferate/pull/2041#issuecomment-5331331374)).

Re-fetched and diffed `1da86c760..5151051b`: exactly one file changed, 3 insertions / 4 deletions —
```
anyharness-lib/src/domains/workflows/store_fanin_tests.rs | 7 +++----
1 file changed, 3 insertions(+), 4 deletions(-)
```
Confirmed the diff is precisely: the wrapped `use super::model::{...}` block reverted to its original single line (`WorkflowLegStatus, WorkflowNodeStatus, WorkflowNodeType, WorkflowRunStatus,`), and the two new assertions in `a_failed_per_leg_relaunch_terminalizes_only_the_addressed_leg` / `a_failed_per_leg_relaunch_cancels_and_disposes_running_siblings` now use `super::model::WorkflowNodeFailureCode::NodeLaunchFailed` fully-qualified, matching the existing style at (now) line 425. No behavioral file (`store.rs`, `node_sessions.rs`, `transition.rs`, `actor.rs`) changed between the two heads, so my prior behavioral verdict on the original "keyless sibling terminalization" finding stands unmodified — this commit only touches the ratchet residue.

Independently verified the residue is cured (checked out `5151051b` myself, did not rely solely on the receipt):
```
$ wc -l anyharness/crates/anyharness-lib/src/domains/workflows/store_fanin_tests.rs
     691 anyharness/crates/anyharness-lib/src/domains/workflows/store_fanin_tests.rs
```
```
```
(exit code 0) — matches the ratchet's own `max_lines = 691` for that file, and the file's own reason line already accounted for exactly this count. This is a static, non-cargo check so it's within my content-review scope; I did not re-run `cargo test` myself (machine limits / CI already baking) and take the coordinator's literal receipt — "test result: ok. 222 passed; 0 failed; 0 ignored; 0 measured; 1940 filtered out; finished in 21.23s" — at face value since it lines up with the code I already traced.

No new findings. Nothing else in the delta beyond this one-file, line-neutral-style cure.

pablonyx:
## Delta review B verdict — updated (pinned to `5151051ba2d890528bafbbcb76ac54bd51a2d113`)

**Verdict: CLEAN** — the ratchet-residue BLOCKING finding from my prior comment ([issuecomment-5331422618](https://github.com/proliferate-ai/proliferate/pull/2041#issuecomment-5331422618)) is cured. My earlier behavioral verdict on the original defect (5323442617 / confirmed 5323790059) is unaffected and still stands: could not refute.

### Diff against my previously reviewed head (`1da86c760`) — confirmed line-neutral, one file

```
git diff --stat 1da86c760f54b97375e46d0b69a9cf408aa6e36c 5151051ba2d890528bafbbcb76ac54bd51a2d113
 .../anyharness-lib/src/domains/workflows/store_fanin_tests.rs      | 7 +++----
 1 file changed, 3 insertions(+), 4 deletions(-)
```

Content of the change, verified by `git show 5151051ba2d890528bafbbcb76ac54bd51a2d113`:
- The `use super::model::{...}` block is reverted to its original single-line form (drops the `WorkflowNodeFailureCode` import added by the compile-cure I flagged).
- The two new test-assertion sites now spell `super::model::WorkflowNodeFailureCode::NodeLaunchFailed` in full, matching the pre-existing fully-qualified usage already at line 425 of the same file (`WorkflowLegStatus::Failed(super::model::WorkflowNodeFailureCode::NodeLaunchFailed)` — identical pattern, now used 3 times consistently).
- `lints/anyharness/ratchets.toml`: `git diff 1da86c760..5151051b -- lints/anyharness/ratchets.toml` produces **zero output** — byte-identical, `max_lines = 691` unchanged (correct, since it doesn't need to move).
- No production files touched: `store.rs`, `node_sessions.rs`, `transition.rs`, `actor.rs` are all absent from the diff stat — the fix is test-code-only and semantically inert (fully-qualifying a type path changes nothing at runtime).

### Line count and lint, verified independently

`git show 5151051ba2d890528bafbbcb76ac54bd51a2d113:.../store_fanin_tests.rs | wc -l` → **691**, exactly matching the unchanged ratchet.

Live CI on this exact head confirms the fix: `gh pr checks 2041` now shows **"Repo shape checks" = pass** (previously the job that failed at `1da86c760`). Pulled the literal log line from that passing job (job `95798703786`, run `32163850832`, head `5151051ba2d890528bafbbcb76ac54bd51a2d113` per `gh pr view --json headRefOid`):

```
2026-08-18T17:06:45.0428162Z Max-lines check passed (repo max 600, component max 500, server layer maxes enabled).
```

Did not independently re-run `cargo test` (respecting the one-cargo-build-at-a-time / no-unnecessary-Rust-build constraint) — unnecessary here since the only behavioral-code-adjacent change is a fully-qualified-path substitution with no semantic difference from the unqualified form under the old import, and the coordinator's reported local run (`test result: ok. 222 passed; 0 failed; ... finished in 21.23s`) is consistent with that.

### Behavioral verdict on the original finding — unchanged, still cured

No production code changed between `1da86c760` and `5151051b`, so nothing in my prior refutation attempt (traced every `NodeLaunchFailed` emission site, the `Some(leg_index)` vs `None` branch in `apply_transition`'s `FailNode` handling, `fail_leg_and_cancel_running_siblings_tx`, the no-double-dispose exclusion, and the adhoc-unreachability argument) is invalidated. That analysis still holds at this head.

Both findings (the original keyless-termination defect and the ratchet residue) are now cured at `5151051ba2d890528bafbbcb76ac54bd51a2d113`.

pablonyx:
**Rebase receipt (r6):** `5151051ba` → `82f4ee58c`. One conflict: r6 moves link_start_and_dispatch (and the envelope machinery) from actor.rs into launch.rs — but main had meanwhile added the session-titling feature (title_node_session + node_session_title import) inside the moved function. Resolution: took r6's actor.rs deletions, and **ported main's titling into launch.rs** (node_session_title import, title write between link_workflow_columns and stamp_session, title_node_session as a free fn taking deps/run_id — doc comment preserved verbatim). Note the fan-out path (launch_legs) does not title leg sessions — same gap as main had for it (the feature landed on main after r5 was authored); flagged as a delta-review question rather than silently extended. Substantive → delta review ordered. Full-ladder rebase onto main `8532f52b2` (2026-08-19 ~02:2xZ, executing Pablo's 'continue implementations end to end' order). Chain: #2016 ← #2017 ← #2057 (replaces #2018) ← #2020 ← #2034 ← #2041. Validation: focused vitest 52/52 (definition-v2 34, draft-legs 5, auto-advance-toast 13), full product-client typecheck green (all sibling dists rebuilt). Rust compile = CI proves (a codex-lane cargo held the machine's one build slot; per this lane's precedent). Delta reviews of the rebase carries are being ordered where substantive.

pablonyx:
**Delta reviewer (titling port) — FINDING**

Rebase of rung 6 (per-leg redo) from old head `5151051ba` onto main `8532f52b2`, new head `82f4ee58c`. The titling port is correct; the delta is clean carries save one orphaned import (low-severity, non-CI-blocking). Untitled fan-out leg sessions noted as a non-blocking follow-up.

## Titling port — CONFIRMED CLEAN (items 1-2)

- **Ordering (item 1):** `launch.rs` `link_start_and_dispatch` calls `title_node_session` between `link_workflow_columns` and `stamp_session` (launch.rs:209-213) — the same order main uses in its `actor.rs` copy (`origin/main:actor.rs:350-354`).
- **Body (item 2):** the ported `title_node_session` (launch.rs:452-467) is main's method body (`origin/main:actor.rs:403-415`) verbatim except the free-fn signature `(deps, run_id, node, session_id)` replaces `&self`; `node_session_title` imported from `render` (launch.rs:15); the `warn!` fields (`run_id`, `node_row_id`, `session_id`, `error`) and message are preserved.

## actor.rs residue (item 3) — CONFIRMED, with one defect

- No leftover `render_envelope`/`RenderInputs`/`node_session_title`/`title_node_session` references; the only launch reference is the free-fn call `super::launch::link_start_and_dispatch(...)` (actor.rs:367).
- The `DisposeSessionsThenStartLegs` / `DisposeSessionThenStartLeg` / `StartLeg` arms (actor.rs:170-215) and `launch_node` (actor.rs:319) resolve every symbol they use — `ResolvedSideEffect` variants, `super::launch::{link_start_and_dispatch, render_node_envelope, relaunch_leg}`, imported `launch_legs`/`launch_model`, fully-qualified `node_leg_count`, `InternalSessionCreateInput`, `OriginContext`, `WorkflowDefinition`. Nothing missing.
- **FINDING (unused import).** `actor.rs:16` still imports `WorkflowNodeKind`, but its only use was the `#[cfg(test)] mod tests` block the rebase deleted (old-r6 used it at `5151051ba:actor.rs:477,479`). It is now used nowhere in the file — all `node.kind` logic lives in `launch.rs:48,77,272` — and no `#[allow(unused_imports)]` shields it. One-line fix: `use crate::domains::workflows::model::WorkflowRunNodeRecord;`.
  - **Severity: low / non-CI-blocking.** Required Rust CI is `cargo check --workspace` + `cargo test --workspace` (ci.yml:266,269) with no `-D warnings`, and clippy appears nowhere under `.github/`, so this is a compiler *warning*, not a gate failure — CI stays green. `make clippy` (`cargo clippy --workspace -- -D warnings`, Makefile:1483) would reject it. Partly pre-existing: the non-test `cargo check` already warned on this test-only import at old-r6; the rebase's test-module deletion widens it to the `cargo test` build too. Worth pruning.

## Delta hunks (item 4) — all rebase carries (aside from the residue above)

`git diff 5151051ba..82f4ee58c -- live/workflows/`:
- `actor.rs` -109: the `launch_model` unit tests + `node()`/`definition_with_model()` helpers moved out. Main had already relocated these into `launch_tests.rs`; old-r6 branched before that split. Content-identical, just dedented to top-level.
- `launch.rs` +35: the titling port (above).
- `launch_tests.rs` NEW: adopted from `origin/main:launch_tests.rs` verbatim except the two exact adaptations the rebase requires — import split `use super::actor::{launch_model, DEFAULT_WORKFLOW_AGENT_KIND}` becomes `use super::actor::DEFAULT_WORKFLOW_AGENT_KIND;` + `use super::launch::launch_model;` (launch_model moved crate-internally to launch.rs), and `legs: None` added to the `DefinitionNode` literal (r6 added that field).
- `lifecycle_tests.rs`: pure `fn`->`pub(super) fn` / `struct`->`pub(super) struct` visibility bumps on `agent_node`/`chain`/`fixture`/`WorkflowFixture{.state}`/`start`/`wait_for`/`node_by_def`, matching main's own visibility (`origin/main:lifecycle_tests.rs:35,54,71,72,120,169,200,253`) so `launch_tests.rs` can reuse them.
- `mod.rs` +2: `#[cfg(test)] mod launch_tests;` — matches main.

## Leg-session titling gap (item 5) — NON-BLOCKING follow-up, NOT a regression

`title_node_session` is called only at `launch.rs:211`, inside `link_start_and_dispatch` (the single-node / one-leg path). The fan-out path `launch_legs` (launch.rs:95-193) and the per-leg relaunch `launch_one_leg`/`relaunch_leg` (launch.rs:288-390) both dispatch through `start_and_dispatch` (launch.rs:393), which does not title, and neither calls `title_node_session` after its own `link_workflow_columns`. So **fan-out leg sessions are untitled.**

- **Where a title write would belong:** in `launch_legs` after each leg's `link_workflow_columns` (~launch.rs:147) and in `launch_one_leg` after its `link_workflow_columns` (~launch.rs:360), mirroring the single-node link->title->stamp order. The title should carry the **leg index** — `node_session_title(chain_index, title)` alone gives every sibling leg the same name, defeating the "tell a run's sessions apart" purpose.
- **Not a regression vs main:** main has no fan-out (legs are the r5/r6 feature, authored after main's titling landed); main's single-node titling is preserved by `link_start_and_dispatch`. This is a gap in new functionality, for a founder/implementer note, not a block.

## Compile-cure commits (item 6) — CONFIRMED sensible post-rebase

`a3c36c894` added `WorkflowNodeFailureCode` to `store_fanin_tests.rs`'s `use super::model::{...}`; `82f4ee58c` reverted that (keeping the file line-neutral at its 691 ratchet) and instead fully-qualifies the two call sites as `super::model::WorkflowNodeFailureCode::NodeLaunchFailed` (store_fanin_tests.rs:629,687), matching the pre-existing pattern at store_fanin_tests.rs:425. The path resolves: `super::model` from `domains/workflows/store_fanin_tests.rs` = `domains/workflows/model.rs`, where `WorkflowNodeFailureCode::NodeLaunchFailed` is defined (model.rs:153-154). No unused import remains from the pair.

---

**Net:** titling port is correct; every delta hunk is a legitimate rebase carry except one orphaned `WorkflowNodeKind` import at `actor.rs:16` (low-severity, non-CI-blocking, one-line fix), plus the untitled fan-out-leg-sessions follow-up (non-blocking, not a regression vs main).

pablonyx:
**Delta-review finding 5340218279 — fixed.**

The orphaned `WorkflowNodeKind` import at `actor.rs:16` (left behind when the rebase moved the launch_model test module to `launch_tests.rs`) is pruned in 10450d64d1eb5f12a27e7e0a82c230c495591870, a plain append to the branch tip (no rewrite — the r7 implementer's base 82f4ee58c stays an ancestor).

Verification: `grep -c WorkflowNodeKind actor.rs` = 0 after the prune; the only prior occurrence was the import itself (the reviewer's finding, independently re-confirmed). `launch_tests.rs` keeps its own import of `WorkflowNodeKind` from `domains::workflows::model` — unaffected.

No local cargo run (compile slot held by another lane, per this ladder's precedent) — CI on this head is the compile verdict. Remaining deferred item on this chain: the r5 `updateLeg` coalesceKey one-liner (comment 5340170024), which needs a cascade and waits for r7 as ledgered.

pablonyx:
**Cascade receipt (orchestrator).** Re-parented onto the recovered rung-1 head 9d90be8f0 (see #2016 note). New head: d09fe6415682df03d18f197e9081a26138938bd2 (was 10450d64d; the WorkflowNodeKind import prune rides along as `=`). Range-diff old→new: **all commits `=` (patch-identical)** — base-only move, no content change; per approve-carries precedent the standing review carries. Validated at the chain tip ff836e77c: workflows client suites `39 passed (39)` files / `431 passed (431)` tests, SDK `267 passed (267)`, product-client+SDK `tsc --noEmit` exit 0.

pablonyx:
**Merge-train cascade receipt (orchestrator).** Two-step re-parent after #2016 merged and the recovery lane advanced r2: first onto post-merge main (12518cb17), then onto the adopted r2 head b1c307b6bb (which adds a founder-authored ExistingWorkspace fix; my carried commits inside it are patch-identical). New head: 284311c2e7fcf28c342b0308b2f01eaaae26b3a9. All carried commits `=` in range-diff except one benign shrink: rung 5's commit no longer adds `model_validator` to the models_v2.py pydantic import because the adopted r2 fix already provides it (tree verified: 3 occurrences on the new head). No content change otherwise; standing reviews carry per the approve-carries precedent. CI riding the new heads is the compile/interaction verdict.

pablonyx:
**Cargo lane failure on 284311c2e — attributed NOT-content, rerun ordered.** Literal: `test result: FAILED. 2174 passed; 66 failed; 3 ignored`. Decode: ONE real failure — `route_aware_read_tests.rs:95:5: precondition: an unrouted, credential-less grok is a credential gap, got InstallRequired` (domains/agents/readiness, untouched by any ladder rung) — panicked holding the global env mutex, and the remaining 65 are `test_support.rs:66:10: expected env mutex: PoisonError` casualties of the poisoned mutex, spread across materialization/sessions/subagents/workflows domains this PR never touches.

Decisive superset check: **#2060's Cargo lane PASSED on cd4944db6, which carries this PR's entire content as a direct ancestor** — the identical Rust compiled and passed there. #2034 (this PR's base content) also passed. Main itself shows failure→success flicker in the same window (97658856c fail, 12518cb17 success). Known-flake class, per-occurrence receipt filed here; rerunning the failed job.

PR #2053, opened 2026-08-18
fix(product): keep sidebar row labels clear of the hover actions

tairabun:
## Summary

- A workspace row whose title fills the sidebar had the hover-revealed pin printed on top of its last characters: the actions were an absolute overlay pinned to the row's right edge (two 24px buttons) while the label truncated against the 20px trailing cell underneath them.
- The actions now sit in the same flow as the trailing cells, and each side carries width only in the state it is visible in — the cells collapse to zero width as the actions expand to their natural size on hover, focus-within, or an open action menu. The label re-truncates against whichever is showing.
- Net effect: the label keeps the whole row at rest and gives up exactly the room the icons need when they appear. Measured on a row long enough to fill the sidebar: 22px of label-under-icon overlap before, 0px after (6px clear gap).

## Testing

- Component tier: `pnpm vitest run src/components/workspace/shell/sidebar src/primitives/patterns/sidebar` (134 tests). Two new cases in `ProductSidebarRepositories.test.tsx` pin the swap (actions share the cells' flow parent, neither is an overlay, each owns width only in its own state) and pin that a row without hover actions is untouched.
- `npx tsc -p tsconfig.json --noEmit` clean.
- Visual: the production `ProductSidebarWorkspaceRow` mounted against the production stylesheet in a real browser at 3x, hovered, in both Chromium and WebKit (the desktop shell's engine). Before: pin overlapping `…the d…`, label right edge 210px vs pin left edge 188px. After: `Codex Migration for t…` then a 6px gap, then pin and archive; identical in WebKit. Rest state unchanged — the full label still uses the whole row.
- Repo gates: `check_design_attribution`, `check_max_lines`, `check_component_library`, `check_appearance_scaling`, `report_frontend_structure` all pass.

## Observability

- None. Presentation-only change to one row component; no new events, logs, or metrics.

## Readiness

## Verification

- Hover any workspace row in the sidebar whose title is long enough to truncate: the title shortens by the width of the revealed controls and the pin/archive glyphs sit in clear space. Move the pointer away and the title returns to the full row.

### DIFF
--- a/apps/packages/product-client/src/components/workspace/shell/sidebar/ProductSidebarRepositories.test.tsx
+++ b/apps/packages/product-client/src/components/workspace/shell/sidebar/ProductSidebarRepositories.test.tsx
@@ -111,6 +111,41 @@ describe("ProductSidebarWorkspaceRow trailing slot", () => {
     expect(screen.getByTestId("running-status").parentElement?.className).toContain("opacity-0");
   });
 
+  it("swaps the trailing cells for the revealed actions instead of covering the label", () => {
+    render(
+      <ProductSidebarWorkspaceRow
+        label="A workspace with a very long name that fills the whole row"
+        trailingStatus={<span data-testid="running-status">Running</span>}
+        hoverAction={<button type="button">Pin workspace</button>}
+      />,
+    );
+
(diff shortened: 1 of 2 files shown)


tairabun:
@greptileai review

PR #2057, opened 2026-08-19
feat(workflows): one run rail per concurrent run in the workspace pane (rung 3)

pablonyx:
**Replaces #2018** (rung 3 of the workflows follow-up ladder), which GitHub auto-closed when a rebase incident briefly force-pushed the r2 tip onto the r3 branch (head==base at close time); the branch was restored within minutes but GitHub permanently refuses to reopen a closed PR whose head ref has since moved.

All rung-3 content here was previously reviewed CLEAN through two delta rounds at `eb5033e70` (review chain on #2018: initial finding fixed + FE-PC-2 carry). This head is that content rebased onto current main (`8532f52b2`) atop the rebased r1/r2.

**Rebase carry (one semantic resolution, flagged for delta review):** main evolved `use-workflow-auto-advance-toast.ts` after r3 was authored (extracted `projection` const + `useWorkflowNodeSessionRoster` call). r3's per-run-slot restructuring moved the roster call into the per-run watch scope, which required threading `workspaceId` through `useWorkflowAutoAdvanceRunWatch` (param + call site). Validated: focused vitest 13/13 on the toast suite, package typecheck green.

Ladder: #2016 (r1) ← #2017 (r2) ← this ← #2020 (r4) ← #2034 (r5) ← #2041 (r6).


pablonyx:
**Delta reviewer (rebase carry) — CONFIRMED CLEAN**

Scope: the rebase carry onto main 8532f52b2 in `apps/packages/product-client/src/hooks/workflows/lifecycle/use-workflow-auto-advance-toast.ts`, from old-r3 `eb5033e70` to new-r3 `7cc7e9fc8` (r3's own content already reviewed CLEAN at `eb5033e70`).

1. **Threading complete.** At r3 head (`git show 7cc7e9fc8:...`):
   - `workspaceId: string | null` param on `useWorkflowAutoAdvanceRunWatch` — line 103.
   - Call site in the fixed 4-slot loop passes it — line 86 (`useWorkflowAutoAdvanceWatch`, loop over `MAX_WATCHED_CONCURRENT_RUNS`).
   - Roster call receives the gated value — lines 124-127: `useWorkflowNodeSessionRoster({ workspaceId: watching ? workspaceId : null, projection })`.

2. **No duplicate/dropped logic vs main.** `origin/main`'s version (single-run, `selectNewestWorkflowRun`) calls `useWorkflowNodeSessionRoster({ workspaceId: watching ? workspaceId : null, projection })` once, inline in `useWorkflowAutoAdvanceWatch`. The resolver moved that exact call+guard expression into the per-run split's inner function (`useWorkflowAutoAdvanceRunWatch`, same file) so it now fires once per watched run slot instead of once per workspace — same expression, same gating, no logic changed or duplicated.

3. **Delta is carry-only.** `git diff eb5033e70..7cc7e9fc8 -- apps/packages/product-client/src/hooks/workflows/lifecycle/use-workflow-auto-advance-toast.ts` shows exactly 4 hunks: (a) new import of `useWorkflowNodeSessionRoster`, (b) `workspaceId` added to the loop's call-site object, (c) `workspaceId: string | null` added to the inner function's param type, (d) `projection` pulled out to a named const (comment reworded "this run" → "this workspace's run") plus the new `useWorkflowNodeSessionRoster(...)` call appended after the existing `useWorkflowAutoAdvanceToast(...)` call. Nothing else touched — confirmed old-r3 (`eb5033e70`) had the per-run split with no roster import/param/call at all, i.e. every hunk is the carry.

4. **Roster fan-out is safe.** Read `use-workflow-node-session-roster.ts` (same directory, unchanged by the carry): its only state is a local `reconciledSessionIdsRef` (`useRef`, line 42) private to each hook *call instance* — React allocates independent hook state per call site per iteration of the fixed-length loop, so the 4 per-slot roster calls never share this ref. The cache side (`useWorkspaceSessionCache`, `use-workspace-session-cache.ts`) is backed by react-query keyed on `(cacheScopeKey, workspaceId)` — no module-level/global singleton. `invalidateWorkspaceSessions(workspaceId)` (lines 82-86) is an idempotent `queryClient.invalidateQueries` call; if two slots' effects both fire it for the same workspaceId in one tick that's redundant but harmless, not a clobber. Each slot reconciles against its own run's `projection.nodes`, so distinct runs' distinct sessionIds don't cross-pollute the per-instance dedup sets either. N parallel calls with the same workspaceId and differing per-run projections are safe.

Verdict: rebase carry is complete, correct, and scoped to exactly what it claims to be.

pablonyx:
**Cascade receipt (orchestrator).** Re-parented onto the recovered rung-1 head 9d90be8f0 (see #2016 note). New head: ee94dfed767ff2644a4420de057f3201cf058102 (was 7cc7e9fc8). Range-diff old→new: **all commits `=` (patch-identical)** — base-only move, no content change; per approve-carries precedent the standing review carries. Validated at the chain tip ff836e77c: workflows client suites `39 passed (39)` files / `431 passed (431)` tests, SDK `267 passed (267)`, product-client+SDK `tsc --noEmit` exit 0.

pablonyx:
**Merge-train cascade receipt (orchestrator).** Two-step re-parent after #2016 merged and the recovery lane advanced r2: first onto post-merge main (12518cb17), then onto the adopted r2 head b1c307b6bb (which adds a founder-authored ExistingWorkspace fix; my carried commits inside it are patch-identical). New head: d367468596fd4d2c75b522f90a4e822e4ac006b9. All carried commits `=` in range-diff except one benign shrink: rung 5's commit no longer adds `model_validator` to the models_v2.py pydantic import because the adopted r2 fix already provides it (tree verified: 3 occurrences on the new head). No content change otherwise; standing reviews carry per the approve-carries precedent. CI riding the new heads is the compile/interaction verdict.

pablonyx:
**Tier-2 red on d367468596 — cancelled-job artifact (conclusion `cancelled`, zero failing steps; the spec never executed). Not the #1997 toast class, not content. Rerun ordered.**

PR #2073, opened 2026-08-19
fix(product): keep Changes actions reachable

pablonyx:
## Summary

- Reflow the branch-mode Changes header into two contained rows at the named 28rem pane boundary while preserving wide and non-branch layouts.
- Keep unified wrap-off hunk actions pinned to the visible scrollport edge and mounted in keyboard order, with existing hover/focus reveal and callbacks.
- Preserve Git semantics, wrap-on/split behavior, empty states, and the current-diff cache contract.
- Stack on #2071; this PR must not merge ahead of its base.

## Testing

- Focused ProductClient suites: 20/20.
- `pnpm --filter @proliferate/product-client typecheck`
- `python3 scripts/report_frontend_structure.py --strict --summary-only`
- `python3 scripts/check_frontend_boundaries.py`
- `python3 scripts/check_component_library.py`
- `python3 scripts/check_appearance_scaling.py`
- `python3 scripts/check_max_lines.py`
- `python3 scripts/check_design_attribution.py`
- `python3 scripts/check_docs.py`
- `git diff --check`
- Production-component browser matrix covers 480/380/320 dark and 320 light headers, wrap-off left/mid/max scroll, native keyboard actions, wrap-on, sticky headers, no-current/clean states, and file-header containment.

## Observability

- None: this changes layout, mounting, and focus reveal for existing controls only. No event, log, identifier, failure path, or replay surface is added.

## Readiness

- [x] I followed the pull-request procedure for scope, title, labels, body, and readiness.

## Verification

- Independent review accepted the frozen contract after the max-scroll sticky-inset finding was repaired.
- No Rust or Docker work was required.

PR #2069, opened 2026-08-19
fix(cloud): remove obsolete worktree retention policy

pablonyx:
## Summary

- Keep Cloud worktrees rooted at `/home/user/workspace/worktrees`, while removing the obsolete automatic-retention launch flag and dead server-to-runtime retention adapters.
- Hard-remove the false-success Cloud retention-policy route, active model/store/service code, generated Cloud SDK surface, React query key, and unused ProductClient hook. Historical Postgres and AnyHarness migrations/tables remain intact for rolling replacement and rollback.
- Protect the retained database-only Postgres table from Alembic autogeneration with an exact reflected/no-metadata comparison exclusion; an unmanaged sentinel remains detectable as schema drift.
- Replace deleted positive pins with authenticated 404/no-side-effect checks, dual-prefix OpenAPI absence, ORM absence, exact migrated-table compatibility, generated-client absence, and current Sandbox/local ownership documentation.
- Constitution note: positive API, desktop auth-gate, ProductClient query-key, and ORM registration pins are removed only with their deleted subjects. Their replacement negative proofs are included here. The mypy baseline shrinks by two diagnostics, and the generated-type max-lines ratchet shrinks from 795 to 791; no rule or exception is weakened.

## Testing

- `make cloud-client-generate` twice; both generated outputs were byte-identical.
- Focused runtime-bootstrap/ORM/OpenAPI/removal suite — 16 passed.
- Cloud API/auth/migrated-schema/startup suite — 16 passed.
- Real Alembic `command.check` compatibility proof — 1 passed; combined migrated-schema/startup/physical-shape/ORM suite — 9 passed.
- `pnpm --filter @proliferate/product-client test -- src/hooks/access/cloud/query-keys.test.ts` — 1,065 files / 7,702 tests passed; focused rerun — 1 file / 3 tests passed.
- Cloud SDK, Cloud SDK React, and ProductClient typechecks passed.
- Ruff check/format, server/frontend boundaries, old-path, migration-head, max-lines, mypy-baseline, lint-record, docs, docs-checker, static-residue, and diff checks passed.
- Independent review round 2 accepted `C2-REV-001` after clarifying explicit archive-convergence cleanup versus forbidden automatic retention selection. Round 3 accepted `C2-EXT-001` after the exact Alembic compatibility guard and sentinel proof landed.

## Observability

- None. Old callers now fail deterministically with 404 and produce no policy database or runtime side effect. Deleting the dead adapters removes only obsolete warning paths for runtime routes that no longer exist.

[...]

PR #2076, opened 2026-08-19
fix(product): preserve transcript position through prepend

pablonyx:
## Summary

- Preserve the reader position when Linux WebKit momentum continues during an older-history prepend.
- Route reconciliation through the existing single frame-pipeline writer and remove the duplicate direct scroll writer.
- Preserve the leading glue frame under rapid position events, then drain one coalesced trailing demand when an event arrives during the current or height-quiet terminating pass.
- Keep fresh-owner `beginGlue()` restart semantics unchanged.
- This is the standalone scroll-owner prerequisite for the identical required-context failure observed on #2062, #2069, and #2073.

## Testing

- Tier 1: focused Vitest owners, 35/35 passed.
- Tier 1: shared frontend build, ProductClient typecheck, scroll-fixture typecheck, transcript writer checker, strict frontend structure, boundaries, max-lines, docs, design attribution, and diff hygiene passed.
- Tier 2: unchanged exact prepend Playwright case passed in Chromium and WebKit, 2/2.
- Tier 2: unchanged full scroll-physics suite passed, 33 passed / 1 intentional WebKit skip / 0 failed.
- v1 negative control on current main: expected settled seat 1831, received 0.
- v1.1 scheduler negative control on published head `6fa15e7944`: the regression failed because `ensureGlue` was absent; companion proof shows repeated `beginGlue()` cancels and replaces the pending handle.
- v1.2 trailing-edge negative control on published head `c46d8e1ccb`: after the first and quiet-terminating passes each seated then suffered reentrant erosion, the scheduler stopped at 0 instead of draining the owed pass to 1831.

## Observability

- None. The repair adds no telemetry or geometry logging. Existing renderer diagnostics remain unchanged.

## Readiness

## Verification

- Frozen contract: `CS-PREPEND-v1.2`, exact PR base `4390327f55c5c3d6ffde343ea1f5158f304c43b7`.
- Current-main drift audit: remote main `e0a74d6c8ec0f5392db3651ff4423442547749fb`; zero path or semantic overlap; guarded merge tree is clean.
- Hosted v1 disproof: run `32255271740`, job `96075397603`, artifact `9366161651` failed only WebKit at `scrollTop=0`; every other check passed.
- Hosted v1.1 disproof: run `32257422106`, job `96082285192`, artifact `9367009151` again failed only WebKit at `scrollTop=0`; 32 pass / 1 intentional skip / 1 fail. No retry was used.
- Durable CI review: https://github.com/proliferate-ai/proliferate/pull/2076#pullrequestreview-4972435237.
- Independent v1.1 review accepted exact pre-commit diff `b559bc43b064c089f16c1265f65e3268d4c817d5f749feea58d61831ead90c3a`; hosted CI then disproved its sufficiency.
- Independent v1.2 architecture and implementation reviews accepted the exact incremental diff `822a486962d6b34f91ac085f023d013f913be356b095622aadddc70150065268` with no findings.
- Current head: `8947eb9ff0af62f247821102884ac90a2f77a1aa`. Production scope remains limited to the existing hook and pipeline owners; browser fixture, thresholds, retries, projects, packages, lockfile, lifecycle deadlines and math, and Changes heads are untouched.

### DIFF
--- a/apps/packages/product-client/src/hooks/chat/ui/transcript-frame-pipeline.test.ts
+++ b/apps/packages/product-client/src/hooks/chat/ui/transcript-frame-pipeline.test.ts
@@ -243,6 +243,26 @@ describe("TranscriptFramePipeline", () => {
     expect(frames.pending).toBe(1);
   });
 
+  it("coalesces repeated beginGlue calls before a frame into one writer pass", () => {
+    const { frames, pipeline } = makePipeline();
+    let passes = 0;
+    pipeline.setWriter({
+      runFramePass: () => {
+        passes += 1;
+      },
+      measureContentHeight: () => 500,
+      shouldContinueGlue: () => true,
+    });
+
+    pipeline.beginGlue();
+    pipeline.beginGlue();
+    pipeline.beginGlue();
+
+    expect(frames.pending).toBe(1);
+    frames.flushFrame();
+    expect(passes).toBe(1);
+  });
+
   // PRO-168 (rung 12, Q16): the eased-follow motion writer's continuation seam.
   // The instant writer never implements `hasPendingMotion`, so these prove the
   // seam is additive: absent it, existing behavior above is untouched (asserted
--- a/apps/packages/product-client/src/hooks/chat/ui/transcript-frame-pipeline.test.ts
+++ b/apps/packages/product-client/src/hooks/chat/ui/transcript-frame-pipeline.test.ts
@@ -13,6 +13,7 @@ import {
  */
 class FakeFrames {
   time = 0;
+  cancelled: number[] = [];
   private seq = 0;
   private queue: Array<{ handle: number; cb: () => void }> = [];
 
@@ -23,6 +24,7 @@ class FakeFrames {
   };
 
   caf = (handle: number): void => {
+    this.cancelled.push(handle);
     this.queue = this.queue.filter((entry) => entry.handle !== handle);
   };
 
@@ -30,6 +32,10 @@ class FakeFrames {
 
   get pending(): number {
     return this.queue.length;
+  }
+
+  get pendingHandles(): number[] {
+    return this.queue.map(({ handle }) => handle);
   }
 
   /** Fire every frame queued as of now (a snapshot), advancing the clock once. */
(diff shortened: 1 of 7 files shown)


pablonyx (commented):
Current-head CI review at `6fa15e79442d7e4f8cab053369e0ad98d4a1dabf`: BLOCK on `CS-PREPEND-CI-001`.

GitHub-hosted WebKit reproduced the exact prepend-anchor failure in run 32255271740 / job 96075397603: Chromium passed; WebKit ended with 32 passed, 1 intentional skip, and 1 failure at scroll-physics.spec.ts:750 (`after.scrollTop` expected greater than 150, received 0). The checked-out merge tree exactly equals the head tree, so this is not main drift or merge-ref synthesis.

The uploaded trace shows the compensated seat briefly restored, then dense WebKit momentum eroding it to the newly prepended top. Each live momentum event calls `beginGlue()`, which cancels and reschedules the pending corrective frame; the burst can therefore starve the sole writer. Current unit coverage flushes before erosion or asserts only queue count, so it does not pin preservation of the oldest scheduled pass.

Required repair: use an idempotent/coalescing ensure-glue path that preserves an already queued correction, add a deterministic fake-frame test with several erosion events before the first flush, and rerun the unchanged Linux WebKit case/full suite. This finding is isolated to the prerequisite; R2 remains unchanged and unrelated.

pablonyx (commented):
Current-head follow-up review at `c46d8e1ccb7751ec57d70ab7c261c99c0ce19218`: `CS-PREPEND-CI-001` remains BLOCKING.

Hosted run 32257422106 / job 96082285192 reproduced the identical WebKit result (32 passed, 1 intentional skip, 1 failed; line 750 received scrollTop 0). The new trace is near-identical to v1 and rules out runner starvation or main interaction.

`ensureGlue()` preserves an already-pending leading handle, but it equates `glueActive` with a guaranteed future writer. A position event can arrive while the current or quiet-terminating `glueTick` is executing; `ensureGlue()` returns because the window is active, then that tick terminates and loses the trailing request. The added test calls ensure repeatedly only before the first frame, so it proves handle identity but not a pass after the final erosion/current tick.

Required next contract: every live non-cancelable position event guarantees at least one not-yet-run writer pass after that event/current pass while preserving any leading pass. Use explicit trailing-edge dirty/generation state, with a deterministic test that injects erosion during the quiet-terminating writer pass, stops events, then proves a subsequent pass restores the full settled 1831 seat. Keep the unchanged Ubuntu WebKit case/full suite as the acceptance proof.

PR #2084, opened 2026-08-19
fix(product): make support save-copy cleanup truthful

pablonyx:
## Summary

Scope: 10 files in the Desktop support-feedback "Save a copy" path.

- `apps/packages/product-client/src/lib/workflows/support/save-support-snapshot-copy.ts` (+test) — new pure archive-then-exact-cleanup workflow.
- `apps/packages/product-client/src/hooks/support/workflows/use-support-snapshot-consent.ts` (+test, +save-copy test) — action serialization/dedupe, cancellation await, latch wiring.
- `apps/packages/product-client/src/hooks/support/facade/use-support-modal-state.{ts,test.tsx}` — facade plumbing for the new truthful result shape.
- `apps/packages/product-client/src/components/support/SupportSnapshotSaveCopyButton.tsx` (+ `SupportSnapshotConsentField.test.tsx`) — whole-action busy state, text-only escape for the unconfirmed-cleanup case.
- `specs/codebase/systems/product/support/README.md` — doc update to match the new contract.

### Behavior change

- The save-copy workflow now returns **independent archive/cleanup truth** instead of collapsing both into a single success/failure status. A saved ZIP is reported as saved even if the follow-up artifact cleanup could not be confirmed — the UI no longer claims cleanup succeeded when it didn't.
- The save-copy action is **serialized**: repeated clicks while an action is in flight dedupe onto the same in-flight Promise rather than racing a second save.
- Cached cancellation is now **awaited** before a new action proceeds, closing a race where a stale prepare could clobber a fresh one.
- The button now reflects a **whole-action busy state** (prepare + save + cleanup) rather than flickering between narrower sub-states.
- When cleanup truth can't be established, the workflow **latches to a conservative "cleanup unconfirmed" state** and blocks further snapshot actions until resolved, instead of assuming success.
- No new telemetry/observability surfaces were added; existing error messaging (`SAVE_FAILED_MESSAGE`) and blocked-state UI are reused, presented as **text-only** (no new iconography) for the unconfirmed-cleanup escape hatch.

## Rebase note

Rebased onto `main` @ 36ffc263. Rebase itself was conflict-free; while proving the build post-rebase (`pnpm --filter @proliferate/product-client build`, required because Cloud SDK source changed on main), tsc caught a real `TS2454: 'promise' used before being assigned` in `use-support-snapshot-consent.ts` (a `let`/reassign pattern that referenced the variable from within its own initializer's closure). Fixed by declaring `promise` as a single `const` from the async IIFE and moving the activeSavePromiseRef-clearing cleanup into a `.finally()` attached after assignment, so the closure only ever reads `promise` once it is definitely assigned. No behavior change from this fix — same dedupe/clearing semantics, just TS-legal.

## Testing

Done locally (machine had swap headroom, so full local proof was safe this run):
- `python3 scripts/check_docs.py` — passed (232 Markdown files).
- `CI=1 pnpm install --offline --frozen-lockfile --ignore-scripts` — passed.
- `CI=1 pnpm --filter @proliferate/product-client build` — passed after the tsc fix above.
- Focused vitest on the four changed/new test files plus the new save-copy test:
  - `use-support-snapshot-consent.test.tsx` (10 tests)
  - `use-support-snapshot-consent.save-copy.test.tsx` (23 tests, new)
  - `save-support-snapshot-copy.test.ts` (6 tests, new)
  - `use-support-modal-state.test.tsx` (11 tests)
  - `SupportSnapshotConsentField.test.tsx` (9 tests)
  - All 59 tests passed.
- `npx tsc -p tsconfig.json --noEmit` in `apps/packages/product-client` — passed.

Not run locally (rely on remote CI): full monorepo build/test/lint matrix outside the product-client package and its direct build deps (cloud-sdk, anyharness sdk, design) already exercised by the build step above.

## Observability

No new logs, metrics, or traces added. Existing error surface (`SAVE_FAILED_MESSAGE`) and the new conservative "cleanup unconfirmed" blocked state are user-facing text only; no telemetry emitted for either path.

### DIFF
--- a/apps/packages/product-client/src/components/support/SupportSnapshotConsentField.test.tsx
+++ b/apps/packages/product-client/src/components/support/SupportSnapshotConsentField.test.tsx
@@ -26,9 +26,14 @@ function consentState(
     setScope: vi.fn(),
     activeSessionAvailable: true,
     isPreparing: false,
+    isBusy: false,
+    snapshotActionsBlocked: false,
     error: null,
     prepare: vi.fn(async () => ({ state: "none" as const })),
-    saveCopy: vi.fn(async () => {}),
+    saveCopy: vi.fn(async () => ({
+      state: "not_started" as const,
+      reason: "unavailable_or_not_consented" as const,
+    })),
     cancel: vi.fn(),
     ...overrides,
   };
@@ -118,7 +123,10 @@ describe("SupportSnapshotSaveCopyButton", () => {
   });
 
   it("saves a copy from the same live consent", () => {
-    const saveCopy = vi.fn(async () => {});
+    const saveCopy = vi.fn(async () => ({
+      state: "saved" as const,
+      cleanup: "confirmed" as const,
+    }));
     render(
       <SupportSnapshotSaveCopyButton
         snapshot={consentState({ consent: true, saveCopy })}
@@ -128,5 +136,30 @@ describe("SupportSnapshotSaveCopyButton", () => {
     fireEvent.click(screen.getByRole("button", { name: "Save a copy…" }));
 
     expect(saveCopy).toHaveBeenCalledTimes(1);
+    expect(screen.queryByText(/cleanup/i)).toBeNull();
+  });
+
+  it("loads for the whole busy action and disables for the cleanup latch", () => {
+    const rendered = render(
+      <SupportSnapshotSaveCopyButton
+        snapshot={consentState({ consent: true, isBusy: true })}
+      />,
+    );
+    const button = screen.getByRole("button", { name: "Save a copy…" });
+    expect(button.hasAttribute("disabled")).toBe(true);
+    expect(button.querySelector("[data-loading-spinner]")).toBeTruthy();
+
+    rendered.rerender(
+      <SupportSnapshotSaveCopyButton
+        snapshot={consentState({
+          consent: true,
+          isBusy: false,
+          snapshotActionsBlocked: true,
+        })}
+      />,
+    );
+    expect(screen.getByRole("button", { name: "Save a copy…" }).hasAttribute("disabled"))
+      .toBe(true);
+    expect(document.querySelector("[data-loading-spinner]")).toBeNull();
   });
 });
--- a/apps/packages/product-client/src/components/support/SupportSnapshotSaveCopyButton.tsx
+++ b/apps/packages/product-client/src/components/support/SupportSnapshotSaveCopyButton.tsx
@@ -24,7 +24,8 @@ export function SupportSnapshotSaveCopyButton({
     <Button
       type="button"
       variant="secondary"
-      loading={snapshot.isPreparing}
+      loading={snapshot.isBusy}
+      disabled={snapshot.snapshotActionsBlocked}
       onClick={() => { void snapshot.saveCopy(); }}
     >
       Save a copy…
--- a/apps/packages/product-client/src/hooks/support/facade/use-support-modal-state.test.tsx
+++ b/apps/packages/product-client/src/hooks/support/facade/use-support-modal-state.test.tsx
@@ -97,6 +97,9 @@ const PREPARED_ARTIFACT = {
 /** Put the hook on a host that can actually prepare a snapshot. */
 function enableSupportSnapshotHost() {
   diagnosticsMocks.supportSnapshot = snapshotBridge;
+  snapshotBridge.cancelPreparation.mockResolvedValue(undefined);
+  snapshotBridge.saveArchive.mockResolvedValue("diagnostics.zip");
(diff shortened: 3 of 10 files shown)


pablonyx:
## Independent review of record — verdict: **APPROVE**

(Posted as a comment because GitHub rejects `--approve` as self-review on this PR. Reviewer did not implement this change.)

Scope: full post-change read of the ten changed files at c5c8b5d9 plus `gh pr diff 2084`. No builds/tests run (machine constraint); relying on the stated local proofs.

### Amendment focus: the TS2454 fix in `use-support-snapshot-consent.ts:310-351`

All four properties hold.

(a) **Same-Promise dedupe identity is preserved.** `promise` (L310) is the async IIFE promise. `.finally()` at L346 does create a derived promise, but it is `void`-ed and never stored or returned. Both hand-out sites give callers the identical object: the direct `return promise` (L351) and the join `return active` (L290, where `active === activeSavePromiseRef.current === promise`). Callers can never observe the derived promise.

(b) **The ref clears exactly at settlement, never earlier.** The old `let` form cleared inside the IIFE `finally`, i.e. *before* the promise settled; the new form clears one microtask *after* settlement. That is strictly safer for dedupe, not weaker. The only theoretical cost is a sub-microtask window where a joiner would receive an already-settled promise instead of starting a fresh action — unreachable from user events, since `releaseAction` (L342) and the `.finally` clear run in the same microtask chain before control returns to the event loop.

(c) **No unhandled rejection.** The IIFE has no reachable rejection path: `runPreparation` is fully wrapped (L262 catch-all, L266 finally with no throwing work), `settleCancellation` swallows with `.catch(() => {})` (L377), and `saveSupportSnapshotCopy` catches both the archive and the delete independently and always resolves. So the derived `.finally` promise can never reject. The existing `unhandledRejection`-listener test (`use-support-snapshot-consent.save-copy.test.tsx:328-349`) would catch a regression here.

(d) **Stale-write guards are unaffected.** They key on `actionRef.current === token` and `epochRef.current === prepared.consentEpoch` (L331-338, L196), not on promise identity.

### Truth reporting

`save-support-snapshot-copy.ts:22-42` is a clean pure workflow: archive truth (`saved` / `cancelled` on `null` / `save_failed` on throw) settles first and independently of cleanup truth, and cleanup is `unconfirmed` only when `deleteArtifact` rejects. There is no path where cleanup is reported `confirmed` without the delete having resolved. The latch (`setBlocked` at L339) fires on every `cleanup === "unconfirmed"` regardless of archive outcome, and `cancelAdmitted` (L197-203) latches on admitted cancellation, matching the frozen intent. No telemetry added anywhere in the diff.

### Send admission

`use-support-modal-state.ts:145-150` gates `canSend` on `!snapshotConsent.isBusy` and on `(!consent || !snapshotActionsBlocked)`, so the text-only escape survives by clearing consent. The stale-closure hole is closed by defence in depth: even a captured pre-busy `handleSend` reaches `prepare()`, which checks `busyRef` *first* (L273) and returns `{state:"blocked"}`, so `handleSend` returns before building or enqueuing a job. The `staleSend` case in `use-support-modal-state.test.tsx` exercises exactly that and asserts nothing dispatched.

### Test quality (non-vacuous)

The tests would fail on the regressions that matter: `toBe(saving)` identity at three distinct phases (preparation, archive, cleanup) plus `not.toBe(previousSave)` after settlement pins the dedupe/clear semantics; the `it.each` over saved/cancelled/save_failed x unconfirmed pins independent truth reporting; `snapshotCallCounts()` equality in `expectLatchBlocksSnapshotWork` proves the latch does zero native work rather than merely returning a shape; `expectCancellationPending` proves the promise stays pending until native cancellation settles. Doc changes in `specs/codebase/systems/product/support/README.md` match the code, including the "no telemetry" and "admitted cancellation also latches" claims.

### Non-blocking notes

1. `use-support-snapshot-consent.ts:137-143` — `supersede()` runs from `setConsent`/`setScope`, so changing scope while a Send preparation is admitted routes through `cancelAdmitted` and permanently latches `snapshotActionsBlocked` for the modal. That is what the frozen intent asks for ("admitted cancel" latch) and is conservative in the right direction, but it is a fairly reachable user action (tick consent, Send, then change scope) whose only recovery is closing and reopening the modal. Worth a product read later; not a defect against this spec.
2. `use-support-modal-state.ts:166-170` — when `prepare()` returns `blocked`, Send silently no-ops with no message. The button is disabled in the normal path so this is only the stale-closure case, but a surfaced reason would be friendlier.
3. `use-support-snapshot-consent.ts:346` — `void promise.finally(...)` is safe today only because the IIFE cannot reject. If a future edit adds a throwing statement outside the inner `try`, the derived promise becomes an unhandled rejection. `promise.then(clear, clear)` would be rejection-proof by construction.

No P0/P1 findings.

PR #2085, opened 2026-08-19
feat(ops): add read-only Grafana metadata inventory

pablonyx:
## Summary

Implements the frozen CP-G0 specification: a read-only Grafana metadata inventory
seam on the existing fixed-target client. Nine files in `scripts/ops/` are the
complete scope:

- `grafana-client.mjs` — adds `readMetadataInventory()` and the private
  `prepareAuthorizedGet` capability; keeps sole ownership of `TARGET`,
  `WORKSPACE_BASE_URL`, token-provider invocation, bearer storage/attachment,
  URL construction and exact-origin revalidation, and `fetchImpl` invocation.
- `grafana-metadata-inventory.mjs` — new private orchestration/transport owner
  (`readMetadataInventoryInternal`): request plan, clocks, timers, abort
  controller, deadline checkpoints, body reads, aggregate byte accounting,
  status classification, health fan-out, global stops, final serialization.
- `grafana-metadata-normalization.mjs` — new pure parsing/normalization owner:
  `METADATA_LIMITS`, numbered parsed-precedence rows, well-formed Unicode
  validation, canonical unsigned UTF-8-byte ordering, safe-integer domains,
  policy/Ruler walkers, all nine normalizers.
- `grafana-credential-process.mjs` — narrow `runContextualAwsCommand` helper for
  the paired signal/monotonic-guard credential path. Legacy no-context calls
  remain exact two-argument calls.
- Four focused test files plus one shared fixture helper
  (`grafana-client.inventory.{credentials,transport,normalization,integration}.test.mjs`,
  `grafana-client.inventory.fixtures.mjs`).

Per the frozen spec this adds **no CLI**, no receipt, no cache, no log, no
provider write, no runbook edit, no new dependency, and no public
aggregate-health verdict. The PR title's "CLI" wording is inherited from the
task assignment; the change is an in-memory library seam only.

### Read-only and credential-privacy guarantees

- Every inventory request is `GET`, has no body and no mutation header, uses
  `redirect: "manual"`, and revalidates its complete URL against the fixed
  origin immediately before the bearer is attached. There is no
  Authorization-free request and no second client, target, or credential owner.
- One bearer is snapshotted per invocation, kept only in the invocation closure,
  and never placed on the client object, result, error, log, or URL.
- The result returns no raw body, provider prose, bearer material, customer
  event/query content, contact setting values, or secure settings. Failures
  never degrade into empty or partial success.
- Whole-inventory deadline (30s) aborts the canonical credential context,
  requests termination of the active direct AWS child, and prevents every later
  dependent launch; late settlements and output are consumed and discarded.

## Testing

Tier 1/2 per `specs/TESTING.md` — pure-unit and seam tests with real local
fixtures, no mocked LLM and no fake sandbox.

- `node --test scripts/ops/grafana-client.inventory.*.test.mjs` → **157/157 pass**
  (was 151 pass / 6 fail on first execution).
- `node --test scripts/ops/*.test.mjs` → **201/201 pass**, which includes the
  unchanged `grafana-alerting.test.mjs` write-regression suite.

### Precedence fix made after first execution

First execution of the four inventory suites reproduced six deadline-classification
failures. All six were one root cause plus one gap in
`grafana-metadata-inventory.mjs` (26 insertions, 8 deletions; no other file and
no test changed):

1. **A deadline checkpoint was not one atomic clock observation.** The
   per-request guard called `monotonicNow()` twice — once against the whole
   deadline and once against the derived request deadline — so the clock could
   advance *inside* a single checkpoint and a deadline could pre-empt a
   synchronous classification the spec makes atomic. `guardRequestDeadline()`
   now takes one reading and compares it against the whole deadline first and
   the request deadline second. `createRequest()` likewise takes one reading for
   both the derived deadline and the timer delay.
2. **The outer `datasourceHealth` surface skipped its own numbered rows.** That
   surface is assembled by the inventory module rather than by a normalizer, so
   it now applies the two remaining precedence-row checkpoints (child-state and
   projection) before completing the enumeration, on the fan-out path only.

Restored behaviours: a selected policy shape failure is no longer replaced by a
later-node deadline; byte-limit, `invalid_utf8`, and `invalid_json` selections
land atomically once their checkpoints have passed; and a whole deadline
observed at final serialization replaces every completed surface with
`unavailable / timeout` rather than stopping globally over already-completed
surfaces.

### Credential-free sanity run

Constructed the public client with providers that throw, return non-bearer
material, and return nothing. All three fail closed with every surface
`unavailable / credential_unavailable`, issue zero Grafana fetches, and the
serialized result contains no fragment of the injected secret.

## Observability

No observability delta per `specs/OBSERVABILITY.md`. CP-G0 emits no log, metric,
trace, receipt, or operator signal, and adds no operator command or alert-runbook
change. Aggregation and any public surfacing are owned by CP-O1.

## Readiness

- [x] No tracker, report, user, or support IDs or private source data appear in
  this PR.

## Verification

- `node --test scripts/ops/grafana-client.inventory.*.test.mjs` — 157 tests, 157 pass, 0 fail.
- `node --test scripts/ops/*.test.mjs` — 201 tests, 201 pass, 0 fail.
- Credential-free fail-closed run described above; no credential printed.
- Not merged.


pablonyx:
## Independent review — PR #2085 (CP-G0 Grafana metadata read inventory)

**Verdict: CHANGES REQUIRED** (one P1; no P0. Security and mutation-safety posture is clean.)

Reviewer of record: fresh reviewer, did not implement. Head reviewed: `07a4bb4fb3cf2ec3910e414538cad6ece0b2b0ca`. Spec read: frozen `CP-G0 Grafana metadata read inventory` r6.

### Executed evidence

- `node --test scripts/ops/grafana-client.inventory.*.test.mjs` -> `# tests 157 # pass 157 # fail 0`.
- `node --test scripts/ops/grafana-alerting.test.mjs` (untouched suite) -> `# tests 44 # pass 44 # fail 0`.
- Physical-line budgets all met: client 546 (<=550 target), inventory 424, normalization 483, credential-process 13, tests 291/325/745/187/318 — all under their spec caps.
- Repair commit `07a4bb4f` touches exactly one file (`scripts/ops/grafana-metadata-inventory.mjs`, +26/-8). **No test file was changed or weakened by the repair.**

### P0 checks — all pass

- **Credential leakage:** the bearer never reaches an error, a result, or a log. `grafana-client.mjs:383` is the only interpolation of `provided`, into the `Authorization` header. Every credential-path failure throws the constant `"Inventory credential is unavailable"` (`grafana-client.mjs:367,378`), and the AWS helpers discard stdout/stderr on failure (`grafana-client.mjs:254,260,264` interpolate only `secretId`/`field`). No `console.*`/`process.stdout` in any new module. `deadlineError` (`grafana-metadata-inventory.mjs:43`) is the only `new Error` in the two new production modules and carries no dynamic text.
- **Mutation safety:** the inventory has exactly one fetch site, `method: "GET"`, `redirect: "manual"` (`grafana-client.mjs:380-385`). All planned paths are reads; the only query params are `type`/`limit`/`page`/`perpage`. Every URL is re-derived through `fixedWorkspaceUrl` on each call, so origin is revalidated before every bearer attachment. No POST/PUT/DELETE reachable from `readMetadataInventory`.
- **Fail-closed credentials:** `prepareAuthorizedGet` validates type, byte cap and charset and throws before the getter exists, so a missing/malformed credential yields all-surface `unavailable / credential_unavailable` with zero fetches (`grafana-metadata-inventory.mjs:389-393`).
- **Unbounded growth:** page loops are fixed at `pagesPerCollection`; I verified both loops always assign a surface on every exit path (`searchPageResult` returns `page_limit` for a full page 5 and `complete` for a short page; `serviceAccountsPageResult` returns `invalid` -> `invalid_shape` when page 5 cannot complete), so no surface can be left `null`. Per-response/aggregate byte caps are checked before chunk retention, and headers never reserve budget.

### Repair-commit analysis

`guardRequestDeadline` (`grafana-metadata-inventory.mjs:57-63`) is **correct** for the stated purpose. Because `deadline = Math.min(whole.deadline, startedAt + requestTimeoutMs)` (`:93`), a single clock reading compared against `whole.deadline` first and the derived deadline second preserves whole-before-request precedence in every case where the clock is the discriminator. The one asymmetry — an already-aborted request controller (line 59) short-circuits before the clock is read, so a monotonically-expired-but-not-yet-aborted whole deadline can be reported as `REQUEST_DEADLINE` — is behaviorally inert: the very next `createRequest` calls `whole.guard()`, which throws `WHOLE_DEADLINE` before any fetch, so the same surfaces end up `unavailable / timeout` and no extra network call is made. Non-blocking, but worth a comment.

### P1 — the two added health checkpoints are unspecified and change the final-output outcome

`scripts/ops/grafana-metadata-inventory.mjs:359-365` (added by the repair commit)

```js
try {
  checkpoint(context.whole.guard, HTTP_CHECKPOINT.HEALTH_STATE);
  checkpoint(context.whole.guard, HTTP_CHECKPOINT.HEALTH_PROJECT);
} catch (error) { return stopNormalizationDeadline(result, "datasourceHealth", error); }
```

Three problems:

1. **Not a spec-sanctioned observation point.** The spec enumerates where deadline observation may occur ("Deadline observation occurs only at mandatory checkpoints: before each awaited provider/fetch/body operation; immediately after that operation settles; before each numbered HTTP-200 phase or parsed-payload precedence row; before each new array item/policy node ...; before a later request; and before final output serialization"). These two land after the last child, before no row, no request and no serialization. They are also back-to-back with zero work between them, so neither can guard anything.
2. **Ownership-boundary violation.** They are named for parsed rows 9 and 10 (`HEALTH_STATE`/`HEALTH_PROJECT`), which the spec assigns to the child normalizer — and `datasourceHealthResult` already emits them at `grafana-metadata-normalization.mjs:467,469`. The spec states the normalization module "owns every normative place and order in which that callback is invoked during parsed rows" and that the inventory module owns health fan-out, not rows. Emitting row-9/row-10 checkpoints from the inventory module for a surface that has no parsed payload contradicts the three-module split (`CP-G0-IMPL-BLOCK-006`). Note also that in the very fixtures that drive these counts the child fails at row 7 or row 8, so rows 9/10 are never evaluated — yet two row-named observations still fire.
3. **Reachable behavioral divergence, demonstrated.** I swept the whole-deadline expiry across every monotonic observation of a one-datasource run and diffed HEAD against a build with only these two lines removed. At the two observation indices they occupy, HEAD returns a **partial** result — `datasourceHealth: {state:"unavailable", reason:"timeout", httpStatus:200}` with `api`, `datasources`, `folders`, ... still `ok`/`empty` — whereas the spec's final-output rule ("If that checkpoint fails, replace every surface with `unavailable / timeout`, no `httpStatus`, and no items") is what those windows produced before the repair and still produces one observation later. The `httpStatus: 200` is fabricated: it comes from `stopNormalizationDeadline`'s hardcoded 200 (`:271`), but the outer health surface has no HTTP response of its own.

**Suggested fix (small):** delete `:359-365`, and correct the two fixture expectations in `scripts/ops/grafana-client.inventory.normalization.test.mjs` — `"row7 health shape before unrelated string"` `observations: 14 -> 12` and `"row8 health string before non-OK"` `observations: 15 -> 13`. I verified those are the exact counts the spec-shaped implementation produces (assertion output `12 !== 14`, `13 !== 15`); no other test moves, and 155/157 already pass with the lines removed. This is the right direction: production should match the spec, and the observation counts should follow.

### Non-blocking notes

- `readJson` uses `whole.guard` rather than `request.guard` for the `DECODE`/`JSON` checkpoints (`:223,227`) after `request.close()` at `:218`. Defensible (the body read is done, so the 5s per-request window no longer applies), but it means a request-level expiry during decode is invisible. Worth one line of comment.
- `timeoutFailure` rethrows non-deadline errors (`:114`), so an unexpected throw from a hostile `response.status` or `headers.get` getter rejects the whole method rather than failing a surface closed. The spec only tables hostile `Location`/`statusText` getters, which are never read, so this is out of scope — but a `catch`-all mapping to `unknown` would be more fail-closed.
- The client rewrite strips a lot of evidence-backed rationale comments (the AMG 10.4 webhook-500 finding, the restore/receipt-credential reasoning) to reach the 550-line quality target. Behavior is unchanged and `restoreOrThrow` is a faithful extraction (44/44 alerting tests pass), but that provenance is genuinely useful; consider relocating it to the runbook rather than deleting it.
- Global stop preserves already-completed surfaces while filling only `null` ones (`stopGlobal`, `:255-262`). That is consistent across the module and appears intended; flagging only because it is the mechanism that makes the P1 above observable.

pablonyx:
## Re-review of the fix delta — PR #2085

**Verdict: APPROVE.** Head re-reviewed: `df3de6dcbd9b9b429bdda404fc56e3492be17796` (delta `df3de6dc` over `07a4bb4f`). Same independent reviewer of record; scope of this pass is the delta only, on top of the earlier full review.

### 1. P1 closed exactly as prescribed

`grafana-metadata-inventory.mjs:359-365` (the two back-to-back `HEALTH_STATE`/`HEALTH_PROJECT` checkpoints) and their `HTTP_CHECKPOINT` names are deleted; `datasourceHealth` now goes straight from the child loop to `completedMetadataItems(items)`. The two fixture counts move 14→12 and 15→13, matching the values I derived. Nothing else in the test files moved.

Re-ran my whole-deadline sweep (expiry forced at each of observations 1..400 on a one-datasource run) against the new head:

- The last observation of a clean run is now the `FINAL_OUTPUT` checkpoint at index 240, and forcing expiry there produces the spec's rule exactly: **every** surface `unavailable / timeout`, no `httpStatus`, no items.
- The two partial-result windows I reported (all surfaces `ok` except `datasourceHealth: timeout, httpStatus:200`) are **gone**. No index now yields a partial result outside the health child's own HTTP/parsed rows, where `httpStatus: 200` is legitimate because a real 200 response was received for that child.
- Every one of the 400 sweep rows is byte-identical to the reference build I used to characterise the defect. No new window was introduced.

### 2. The extra `raceSignal` hunk is safe — and fixes a real crash I hit

```js
if (signal.aborted) {
  Promise.resolve(promise).catch(() => {});
  return Promise.reject(signal.reason || deadlineError(REQUEST_DEADLINE));
}
```

It cannot swallow a classification-bearing rejection:

- The returned rejection is constructed independently (`signal.reason || deadlineError(REQUEST_DEADLINE)`) and is unchanged. The `.catch()` is attached to a *derived* promise and is discarded; it never feeds the caller.
- The branch runs only when the signal is *already* aborted, which is precisely the state in which the spec requires late settlement to be ignored/discarded. There is no path where the handed promise's outcome could otherwise have selected a result — no race is installed in this branch, so its settlement was already unobservable; it was merely unhandled.
- Confirmed empirically rather than by inspection: my earlier sweep against the pre-fix tree died with `Error: Inventory deadline ... inventoryDeadline: Symbol(whole-deadline)` as an unhandled rejection that terminated the Node process. The **entire** diff between the pre-fix reference sweep and the new-head sweep is the disappearance of that crash trailer — all 400 result rows are identical. So the hunk changes exactly one thing (no unhandled rejection) and no classification, as claimed.

### 3. No test weakened, budgets hold

- Only two numeric expectations changed, both downward-corrected to the spec-shaped counts; no assertion, case, or fixture was removed or loosened. Working tree clean at head.
- `node --test scripts/ops/*.test.mjs` -> `# tests 201 # pass 201 # fail 0` (157 inventory + 44 untouched alerting).
- Budgets: client 546 (<=550 target, <=600 cap), inventory 421 (<=500), normalization 483 (<=500).

### Carry-forward non-blocking notes (not gating)

Unchanged from my first review and fine to defer: the `guardRequestDeadline` already-aborted short-circuit can label a monotonically-expired whole deadline as `REQUEST_DEADLINE` (inert — the next `createRequest` throws `WHOLE_DEADLINE` before any fetch); `DECODE`/`JSON` use `whole.guard` after `request.close()`; `timeoutFailure` rethrows non-deadline errors rather than failing a surface closed; and the client rewrite dropped useful evidence-backed rationale comments to hit the line target.

All P0 findings from the first pass remain clean at this head: bearer never printed or serialized, GET-only with per-call origin revalidation, fail-closed `credential_unavailable` with zero fetches, and bounded pagination/output.

PR #2089, opened 2026-08-19
fix(anyharness): re-observe launch options after the dud agent-program swap

pablonyx:
## Summary

- The launch-options cutover (#2070) makes the agent spawn spec part of the observation basis. The lifecycle test `a_failed_launch_fails_the_run_and_compensates_the_half_born_session` swaps `ANYHARNESS_CLAUDE_AGENT_PROGRAM` to a dud after fixture boot, so the boot-time observation became basis-stale and the create was refused BEFORE the session row existed — the test's point is a failure landing AFTER the stamp, in the compensation window. It now re-observes at the dud basis after the swap.
- `test_support::seed_observed_launch_options` now re-seeds when the stored row is basis-stale (`options: None`) instead of early-returning on any row.
- The adhoc lifecycle test asserted the model pick on the legacy `sessions.requested_model_id` column; the cutover deliberately persists the pick in the immutable launch intent, so the assert now reads `find_launch_intent` (the node-row pick assert is unchanged).
- Test-only change; no production code. Cures the red `live::workflows::lifecycle_tests` on main (two stale-contract roots + env-mutex poison victims).

## Testing

- Remote CI on this PR head is the verdict (no local Cargo per machine build policy). `scripts/check_max_lines.py` passes locally after keeping `lifecycle_tests.rs` at its 1066 ratchet.

### DIFF
--- a/anyharness/crates/anyharness-lib/src/app/test_support.rs
+++ b/anyharness/crates/anyharness-lib/src/app/test_support.rs
@@ -52,10 +52,13 @@ pub(crate) fn seed_observed_launch_options(
     service: &HarnessLaunchOptionsService,
     harness_kind: &str,
 ) {
+    // Re-seed when the stored observation is basis-stale (`options: None`,
+    // e.g. after a test swaps the agent program): admission needs a
+    // current-basis observation, not just any row.
     if service
         .read(harness_kind)
         .expect("read launch options")
-        .is_some()
+        .is_some_and(|response| response.options.is_some())
     {
         return;
     }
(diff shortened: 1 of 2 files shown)


pablonyx:
Superseded: main took #2092, which repairs the same red-main lifecycle failures with an equivalent fix (basis-stale re-seed after the dud agent swap via an Observed-state guard, plus the launch-intent assertion for the adhoc pick). The residual delta here is cosmetic and merging would only churn against #2038/#2092. Closing without merge.

PR #2103, opened 2026-08-20
fix(product): send raw target-observed control ids at session create

pablonyx:
## Summary

- Fixes the deployed-product breakage where every new codex chat send failed with `Not sent: launch value '<unknown-control>' for 'effort' is not supported`.
- Root cause: two client paths violated the raw-id contract of the target-observed launch-options cutover (#2070). `useChatLaunchControlActions` persisted normalized control keys (`effort`, `fast_mode`) into `defaultLiveSessionControlValuesByAgentKind`, and session-creation materialization spread that map plus the normalized-keyed intent projection straight into `createSession`'s `controlValues`. The runtime exact-validates keys against raw observed control ids (`reasoning_effort`, `fast-mode` for codex); claude worked only because its raw id happens to be `effort`.
- Writers now persist `rawConfigId` in both branches; creation swaps to the already-existing `rawConfigValuesFromIntentSnapshot`; and per the agent-run-config spec's "run time: refresh the same target runtime's options and validate", the merged selection is fetched-and-filtered against the target's current observation before create. Keys or values outside the observation are dropped silently (omission remains omission), and a failed or unobserved options fetch sends `controlValues: {}` fail-safe.
- No disk migration: stale normalized keys already persisted on user machines are filtered out at the create seam and rewritten as raw ids on the next pick.

## Testing

- `pnpm vitest run` on the two new test files: 7/7 pass (raw-id persistence with negative controls asserting the normalized key is absent; legacy-key filtering with a codex-shaped observation; out-of-set value drop; grok zero-controls -> `{}`; fetch-failure -> `{}`; unobserved states -> `{}`).
- Full product-client typecheck chain (`pnpm run typecheck`) passes locally; `scripts/check_max_lines.py` passes.
- Tier-2 intent tests run with `TIER2_INTENT_SKIP_RUNTIME=1` and cannot cover runtime exact-validation; follow-up recorded to add a scripted-runtime scenario under `anyharness/tests/` covering a persisted-normalized-key session create.
- Remote CI on this head is the verdict for everything else.

## Verification after release

A new codex chat with a remembered effort/fast-mode preference must create successfully, sending either `reasoning_effort`/`fast-mode` or no control keys; the `<unknown-control>` error class disappears. Grok chats with any stale remembered control also recover.

### DIFF
--- a/apps/packages/product-client/src/hooks/chat/workflows/use-chat-launch-control-actions.test.tsx
+++ b/apps/packages/product-client/src/hooks/chat/workflows/use-chat-launch-control-actions.test.tsx
@@ -1,0 +1,65 @@
+// @vitest-environment jsdom
+
+import { act, cleanup, renderHook } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { useChatLaunchControlActions } from "#product/hooks/chat/workflows/use-chat-launch-control-actions";
+import { USER_PREFERENCE_DEFAULTS } from "#product/lib/domain/preferences/user/model";
+import { useUserPreferencesStore } from "#product/stores/preferences/user-preferences-store";
+
+const setActiveSessionConfigOption = vi.fn<() => Promise<void>>();
+
+vi.mock("#product/hooks/sessions/workflows/use-session-config-actions", () => ({
+  useSessionConfigActions: () => ({ setActiveSessionConfigOption }),
+}));
+
+describe("useChatLaunchControlActions", () => {
+  beforeEach(() => {
+    setActiveSessionConfigOption.mockReset();
+    useUserPreferencesStore.setState({
+      ...USER_PREFERENCE_DEFAULTS,
+      _hydrated: false,
+      _persistedMetadata: {},
+    });
+  });
+
+  afterEach(() => {
+    cleanup();
+  });
+
+  it("persists default launch controls under the raw target-observed id", () => {
+    const { result } = renderHook(() =>
+      useChatLaunchControlActions({ activeLaunchAgentKind: null }));
+
+    act(() => {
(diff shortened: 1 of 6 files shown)


pablonyx:
Review round 1 (fresh adversarial reviewer) returned REQUEST_CHANGES; fixes landed in d8a32dceee:

- **P1 fixed**: the observation filter gated on `state === observed/observed_empty`, but the runtime validates against stored options whenever present regardless of state — so `refreshing`/`last_good_after_failure` would have silently dropped valid raw picks. Now gates on options presence only, with a regression test.
- **P2 fixed**: supersession re-check added after the options fetch before `createSession`; dead `configValuesFromIntentSnapshot` deleted.
- **Clarification**: the `use-chat-launch-control-actions` writer change is defensive hardening — on current main `controlKey === rawConfigId` at every live call site. The deployed breakage is repaired by the create-seam filter, which also neutralizes normalized keys already persisted on user disks.
- **Deferred follow-ups** (recorded deliberately for hotfix speed): a preference sanitizer/migration to purge stale normalized keys (harmless at create, cosmetic in optimistic-record readers); a materialization wiring test for the fetch/merge/filter seam; a scripted-runtime Tier-2 scenario in `anyharness/tests/` (CI runs `TIER2_INTENT_SKIP_RUNTIME=1`, so runtime exact-validation is untestable there).

Round-2 re-verification in flight.

pablonyx:
Review round 2 (same fresh reviewer of record, head d8a32dceee): **APPROVE**. P1 gate verified bit-for-bit against the runtime's own state machine (options-presence arms in service.rs match validation.rs's admission basis); supersession re-check placement verified as a pure pre-create abort; 31 files / 176 tests pass across the touched areas. Deferred follow-ups stand as recorded (materialization wiring test flagged as the one to not let lapse). Remaining gate: remote CI on this head; merge is Pablo's.

PR #2104, opened 2026-08-20
fix(product): keep other observed harnesses in the model picker during an active session

pablonyx:
## Summary

- Fixes the post-cutover regression where opening any chat collapsed the composer model picker to the active harness alone: codex/grok/opencode disappeared even though the runtime observed them as ready.
- #2070 rewrote `buildModelSelectorGroups` so a non-empty active live model control replaced the entire agent list with only the matching harness. Pre-cutover, the live statement only overrode the active harness's model rows and all other agents stayed listed (the `open_new_chat` action kind exists precisely for those rows, and was unreachable).
- The live statement remains authoritative for its own harness (`resolveSelectorModels` still swaps in live values for the matching kind); every other observed harness keeps its group, and a live control whose harness is absent from the observed list still gets a synthesized group, preserving the cutover's fallback.

## Testing

- New tests: active claude live control + observed codex agent yields both groups, claude showing only the live statement and codex as `open_new_chat` (negative control against the collapse); unobserved-live-harness case still synthesizes its group.
- `pnpm vitest run` across model-picker and chat-derived suites: 41/41 pass; product-client `tsc --noEmit` clean; `check_max_lines` clean.
- Reproduced live on a dev profile running the shipped runtime binary: all four harnesses observed+ready server-side while the picker showed claude only.

## Verification after release

With a chat open, the picker lists every installed harness's observed models again; picking a codex/grok model opens a new chat as before the cutover.

### DIFF
--- a/anyharness/sdk-react/src/hooks/agents.ts
+++ b/anyharness/sdk-react/src/hooks/agents.ts
@@ -1,5 +1,6 @@
 import {
   useMutation,
+  useQueries,
   useQuery,
   useQueryClient,
 } from "@tanstack/react-query";
@@ -85,6 +86,35 @@ export function useAgentLaunchOptionsQuery(options?: RuntimeQueryOptions & {
       const client = getAnyHarnessClient(resolveRuntimeConnection(runtime));
       return client.agents.getLaunchOptions(harnessKind, requestOptionsWithSignal(undefined, signal));
     },
+  });
+}
+
+/**
+ * Launch options for several harnesses at once, sharing the per-kind cache
+ * entries of [`useAgentLaunchOptionsQuery`]. Returns one response (or `null`
+ * while unresolved/failed) per requested kind, in request order; `combine`
+ * keeps the array reference stable across renders while the data is unchanged.
+ */
+export function useAgentLaunchOptionsListQuery(options?: RuntimeQueryOptions & {
+  harnessKinds?: readonly string[];
+}) {
+  const runtime = useAnyHarnessRuntimeContext();
+  const runtimeUrl = runtime.runtimeUrl?.trim() ?? "";
+  const cacheScopeKey = resolveRuntimeCacheScopeKey(runtime);
+  const harnessKinds = (options?.harnessKinds ?? [])
+    .map((kind) => kind.trim())
+    .filter((kind) => kind.length > 0);
+
+  return useQueries({
+    queries: harnessKinds.map((harnessKind) => ({
+      queryKey: anyHarnessAgentLaunchOptionsKey(runtimeUrl, harnessKind, cacheScopeKey),
+      enabled: (options?.enabled ?? true) && runtimeUrl.length > 0,
+      queryFn: async ({ signal }: { signal: AbortSignal }) => {
+        const client = getAnyHarnessClient(resolveRuntimeConnection(runtime));
+        return client.agents.getLaunchOptions(harnessKind, requestOptionsWithSignal(undefined, signal));
+      },
+    })),
+    combine: (results) => results.map((result) => result.data ?? null),
   });
 }
 
--- a/anyharness/sdk-react/src/index.ts
+++ b/anyharness/sdk-react/src/index.ts
@@ -97,6 +97,7 @@ export {
   useAgentsQuery,
   useWorkspaceAgentsQuery,
   useAgentLaunchOptionsQuery,
+  useAgentLaunchOptionsListQuery,
   useRefreshHarnessLaunchOptionsMutation,
   useAgentReconcileStatusQuery,
   useWorkspaceAgentReconcileStatusQuery,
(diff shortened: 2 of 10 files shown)


pablonyx:
Round-2 review (same reviewer of record, head 33d41f54c1): **APPROVE**, with one P1 landed as 0bb3387152 — on cloud workspaces the fan-out list now comes from the sandbox connection's `readyAgentKinds` (agentCatalog reads the local runtime), and gateway fan-out queries use `retry: false` so a kind the sandbox does not serve resolves to an absent group instead of retry-looping.

Reviewer-verified at 33d41f54c1: end-to-end multi-harness groups in the production path, rules-of-hooks safety of the dynamic useQueries fan-out, combine/replaceEqualDeep array stability (no render churn), fan-out strictly additive to loading/error semantics, and a downstream net fix (configured-launch readiness no longer fires spuriously when the default agent differs from the active chat's kind). Noted non-blocking: snapshotId stays keyed on the requested kind (snapshot currently has no consumers); hoisting the combine closures is a nit.

Known follow-up (not this PR): first paint per target shows the composer's bare model pill until the launch-options observation arrives — pre-cutover the static catalog filled controls instantly. Candidate fix: hydrate composer controls from the last cached observation.

pablonyx:
Round-3 review (same reviewer of record, head 04e86597b7): **APPROVE** — no P0/P1. Both P2s landed as ee7f784c70: the cloud/no-target arm of the home fan-out hook now returns a module-constant empty array (a fresh [] per render was recomputing the six-memo chain on every cloud composer render), and the cloud branch of the chat fan-out gets a pinned test proving the sandbox connection's readyAgentKinds beats the local runtime's list (negative control included).

Reviewer sweep of all projectHarnessLaunchOptions callers found no remaining picker collapse. Noted follow-ups (separate tickets, not this PR): cowork threads can only launch the default harness (pre-existing scope); runtimeLaunchOptionsFromResponse in cloud-launch-catalog.ts is dead code; cloud sandbox copy store lacks a list read, so cloud LAUNCH targets stay single-kind for now (cloud workspace chat fan-out is covered via the sandbox ready list).

Live-verified in a running dev profile against the shipped runtime binary: in-session picker shows claude + codex groups with open-new-chat rows; launch composer fan-out verified by test and live reload.

pablonyx:
## Independent adversarial review — APPROVE-WITH-NITS

Fresh reviewer (did not author this change), read-only against head `ee7f784c70`. No must-fix findings; merging for the launch hotfix.

**Verified correct (adversarially):**
- No duplicate groups: `model-selector-options.ts:66-69` prepends the synthesized agent only when the kind is absent from `agents`.
- Live statement stays authoritative for its own harness; `resolveSelectorModels` gates on `activeModelControl?.kind === agent.kind`, so live values never leak across harnesses.
- No empty-picker regression, no default-selection hijack (`resolveEffectiveLaunchSelection` is kind-scoped and returns `null` rather than falling back to `groups[0]`).
- The re-enabled `open_new_chat` path is admission-safe: model ids come from each harness's own `getLaunchOptions(kind)` response, and control values are re-filtered per picked kind at create time by `filterControlValuesToObservation` (#2103). The picker cannot list a model the runtime will reject.
- Test negative controls are real: against pre-fix `model-selector-options.ts`, `matchingActiveAgents.length > 0` collapses `sourceAgents` to `[claude]`, so the new `toEqual(["claude","codex"])` assertion fails. No test in the diff passes with the fix reverted.
- `sdk-react/src/index.ts` is the package's declared entrypoint, not a net-new convenience barrel, so the AGENTS.md barrel rule is not engaged. The new hook shares `anyHarnessAgentLaunchOptionsKey` with its single-key sibling rather than duplicating it.

**Demo-risk check, resolved:** the reviewer flagged that a harness with `options: null` (Detecting) or zero observed models would still be hidden. Verified against the live runtime DB — all four installed harnesses are `probe_state = succeeded` with non-empty models (claude 5, codex 6, grok 2, opencode 179), so neither case triggers. Grok observes zero *controls*, not zero *models*, and therefore projects normally.

**Follow-ups (not blocking, filed for the morning):**
1. A harness with `options: null` is still silently dropped from every picker with no "Detecting" affordance (`use-chat-launch-catalog.ts:88-101`, `use-home-next-model-selection.ts:64-79`). Suggested: `refetchInterval` while `state` is `detecting`/`refreshing`.
2. Zero-model harness renders a bare group header in chat but disappears entirely from Home — two surfaces disagree.
3. `config/chat-launch.ts:33-48` exports `shouldExposeChatLaunchAgent` and `orderChatLaunchAgents`, both now dead code orphaned by #2070. Reusing them in the `launchAgents` memo resolves 2 and 3 together.
4. `use-home-target-agent-launch-options.ts:97-112` still collapses to a single harness for cloud launch targets (already documented in its docstring).
5. `useAgentLaunchOptionsListQuery` has no SDK-level test in `sdk-react/src/hooks/agents.test.tsx`.
6. Nit: `workspaceId` is required by the new hook's parameter type but never read.

Note: the earlier bot comment reviewed `0b3254d4`, before the fan-out existed, and its central claim no longer holds at this head.

PR #2105, opened 2026-08-20
fix(anyharness): soft-drop per-model launch controls at session start

pablonyx:
## Problem

Codex narrows a control's value set per model: the harness-level observation (probed under the default `gpt-5.6-sol`) advertises `reasoning_effort` values including `max`/`ultra`, but a live session running `gpt-5.5` offers a smaller set. Create-time validation correctly accepts `reasoning_effort=max` against the observation, then the fail-closed exact-confirmation at ACP session start gets `membership_rejected` from the live statement and kills the whole session:

```
Not sent: ACP session start failed: requested control 'reasoning_effort' value 'max' was not confirmed by the live codex session
```

Live repro on a dev profile: session `ced42481` — model confirmed, mode confirmed, collaboration_mode confirmed, `reasoning_effort` membership_rejected, session dead.

## Fix (ruled by Pablo: soft-drop at start) — scoped to quality controls

The drop decision is a pure pre-send function `initial_control_disposition` → `AlreadyLive | Apply | Drop | Refuse`.

**Dropped — exactly one carve-out.** A control that the live statement still surfaces, whose requested value that statement no longer offers under the applied model, and that is *not* a posture control. It launches at the live session default, warns `session.initial_config.dropped` (no value in the record, per the bounded-event contract), and logs result code `membership_dropped`. The final aggregate check runs against the intent minus the dropped controls, so it still catches a later setter resetting a confirmed value.

**Still fatal — a silent default is worse than a refused start:**

- **Posture controls.** Collaboration mode, the mode / approval-policy family, and sandbox mode (`NormalizedControlKind::Mode` or `CollaborationMode`) decide what the agent is allowed to *do*. Launching one at the harness default after the user explicitly selected against it is a silent behavior change, so an unoffered posture value takes the new `Refuse` disposition and fails the start. The legacy ACP `mode` surface is posture too and keeps its original bail.
- **Control ids the live statement never surfaced at all.** That is a vocabulary disagreement between the create-time observation and the live session, not per-model narrowing; dropping it would make every future selection for that control a silent perpetual no-op. The `pending.is_empty()` guard is retained.
- **The model**, and any OFFERED value whose setter read-back refuses, times out, or goes unconfirmed.

**AlreadyLive** short-circuits without a setter round-trip.

Because posture controls now refuse rather than drop, the mode projection can never diverge from a dropped value: the `requested_mode_id` clear that round 2 added at startup is dead and is removed, and `apply_resolved_launch_intent` returns `()` again.

## Specs

- `specs/FEATURE_DOCS/MODELS.md`: actor-startup section states the final policy — what is dropped, what stays fatal, and why.
- `specs/OBSERVABILITY.md`: table row for the dropped event + enumerated result codes.

Observability delta: new bounded event `session.initial_config.dropped` (session identity + config key only) and new `membership_dropped` result code on `session.initial_config.apply`.

## Tests

- `launch_intent_seam_drops_only_unoffered_quality_values` — seam-level, exercises `apply_resolved_launch_intent` itself: a never-surfaced control id is fatal, an unoffered posture value is fatal, and an unoffered quality value is dropped so the start succeeds. The third case is the negative control for the first two — same seam, same live statement.
- `disposition_refuses_posture_controls_the_live_statement_does_not_offer` — posture `Refuse` vs quality `Drop` in the same live statement.
- `disposition_drops_only_values_the_live_statement_does_not_offer` — negative control: an OFFERED value must resolve `Apply` (never silently drop), plus AlreadyLive and per-model-narrowed Drop.
- `dropped_per_model_control_is_excluded_from_the_final_aggregate_check` — negative control: the unfiltered intent still fails the aggregate.
- `intent_without_dropped_controls_keeps_undropped_values`.

## Live verification (dev runtime, round-2 binary)

- codex `gpt-5.5` + `reasoning_effort=max` (the exact production failure): create + start succeed, `reasoning_effort` logs `membership_dropped`, everything else `confirmed`.
- codex `gpt-5.5` + `reasoning_effort=low` (offered): applies and logs `confirmed` — the fatal path for refused offered values remains reachable.

## Review round 3

Addresses the fresh adversarial review (findings 1, 2, 3, 6). Not addressed, deliberately: finding 4 (no user-facing notice that a selection was dropped — needs a notice channel the sink does not expose, out of scope here), finding 5 (defensive model-purpose guard in `initial_control_disposition` — unreachable today), and finding 7 (transient half-updated `session_state_update`, moot now that the block emitting it is deleted).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4240,7 +4240,7 @@ dependencies = [
 
 [[package]]
 name = "proliferate"
-version = "0.4.18"
+version = "0.4.20"
 dependencies = [
  "base64 0.22.1",
  "chrono",
--- a/anyharness/crates/anyharness-lib/src/live/sessions/actor/config/handle.rs
+++ b/anyharness/crates/anyharness-lib/src/live/sessions/actor/config/handle.rs
@@ -59,6 +59,7 @@ pub(in crate::live::sessions::actor) async fn apply_resolved_launch_intent(
     }
 
     let mut pending = intent.control_values.clone();
+    let mut dropped_control_ids: Vec<String> = Vec::new();
     let ordered_ids = startup_state
         .config_options
         .iter()
@@ -83,6 +84,24 @@ pub(in crate::live::sessions::actor) async fn apply_resolved_launch_intent(
                 return Err(error);
             }
         };
+        // Create-time validation runs against the harness-level observation,
+        // but some harnesses narrow a control's value set per model (codex
+        // reasoning_effort). A value the live session no longer offers after
+        // the model applied is dropped to the session default rather than
+        // failing the whole start; a refusal of an offered value stays fatal.
+        if matches!(outcome, ConfigApplyOutcome::NotApplied) {
+            log_initial_config_apply(session_id, agent_kind, &config_id, "membership_dropped");
+            tracing::warn!(
+                session_id,
+                harness_kind = agent_kind,
+                control_id = %config_id,
+                value = %value,
+                event = "session.initial_config.dropped",
+                "requested control value is absent from the live session; launching with its default"
+            );
+            dropped_control_ids.push(config_id);
+            continue;
+        }
         log_initial_config_apply(
             session_id,
             agent_kind,
@@ -99,51 +118,85 @@ pub(in crate::live::sessions::actor) async fn apply_resolved_launch_intent(
     // live statement and can positively confirm through set_mode/read-back.
     if let Some(value) = pending.remove("mode") {
         if !startup_state.legacy_mode_contains_value(&value) {
-            log_initial_config_apply(session_id, agent_kind, "mode", "membership_rejected");
-            anyhow::bail!("requested control 'mode' value '{value}' is absent from the live {agent_kind} session");
-        }
+            log_initial_config_apply(session_id, agent_kind, "mode", "membership_dropped");
+            tracing::warn!(
+                session_id,
+                harness_kind = agent_kind,
+                control_id = "mode",
+                value = %value,
+                event = "session.initial_config.dropped",
+                "requested control value is absent from the live session; launching with its default"
+            );
+            dropped_control_ids.push("mode".to_string());
+        } else {
         let outcome = match apply_mode_via_direct_setter_legacy(
-            conn,
-            native_session_id,
-            startup_state,
-            &value,
-        )
-        .await
-        {
-            Ok(outcome) => outcome,
-            Err(error) => {
-                log_initial_config_apply(session_id, agent_kind, "mode", "apply_failed");
-                return Err(error);
-            }
-        };
-        log_initial_config_apply(
-            session_id,
-            agent_kind,
-            "mode",
-            confirmed_result_code(outcome),
-        );
-        anyhow::ensure!(
-            matches!(outcome, ConfigApplyOutcome::NoChange | ConfigApplyOutcome::AppliedAuthoritative),
-            "requested control 'mode' value '{value}' was not confirmed by the live {agent_kind} session"
-        );
+                conn,
+                native_session_id,
+                startup_state,
+                &value,
+            )
+            .await
+            {
+                Ok(outcome) => outcome,
+                Err(error) => {
+                    log_initial_config_apply(session_id, agent_kind, "mode", "apply_failed");
+                    return Err(error);
+                }
+            };
+            log_initial_config_apply(
+                session_id,
+                agent_kind,
+                "mode",
+                confirmed_result_code(outcome),
+            );
+            anyhow::ensure!(
+                matches!(outcome, ConfigApplyOutcome::NoChange | ConfigApplyOutcome::AppliedAuthoritative),
+                "requested control 'mode' value '{value}' was not confirmed by the live {agent_kind} session"
+            );
+        }
     }
-    anyhow::ensure!(
-        pending.is_empty(),
-        "requested controls are absent from the live {agent_kind} session: {:?}",
-        pending.keys().collect::<Vec<_>>()
-    );
-    if let Err(error) = ensure_resolved_launch_intent_confirmed(startup_state, intent) {
+    // A control id the live statement never surfaced at all is the same
+    // per-model narrowing class: drop it rather than failing the start.
+    for (config_id, value) in pending {
+        log_initial_config_apply(session_id, agent_kind, &config_id, "membership_dropped");
+        tracing::warn!(
+            session_id,
+            harness_kind = agent_kind,
+            control_id = %config_id,
+            value = %value,
+            event = "session.initial_config.dropped",
+            "requested control is absent from the live session; launching with its default"
+        );
+        dropped_control_ids.push(config_id);
+    }
+    let confirmed_intent = intent_without_dropped_controls(intent, &dropped_control_ids);
+    if let Err(error) = ensure_resolved_launch_intent_confirmed(startup_state, &confirmed_intent) {
         log_initial_config_apply(session_id, agent_kind, "complete_intent", "final_mismatch");
         return Err(error);
     }
     tracing::info!(
         harness = agent_kind,
         selected_model = intent.model_id.is_some(),
-        selected_control_count = intent.control_values.len(),
+        selected_control_count = confirmed_intent.control_values.len(),
+        dropped_control_count = dropped_control_ids.len(),
         event = "session.launch_intent.confirmed",
-        "confirmed every explicit launch intent value"
+        "confirmed every applicable explicit launch intent value"
     );
     Ok(())
+}
+
+pub(in crate::live::sessions::actor) fn intent_without_dropped_controls(
+    intent: &ResolvedLaunchIntent,
+    dropped_control_ids: &[String],
+) -> ResolvedLaunchIntent {
+    if dropped_control_ids.is_empty() {
+        return intent.clone();
+    }
+    let mut confirmed = intent.clone();
+    confirmed
+        .control_values
+        .retain(|config_id, _| !dropped_control_ids.contains(config_id));
+    confirmed
 }
 
 pub(in crate::live::sessions::actor) fn ensure_resolved_launch_intent_confirmed(
--- a/anyharness/crates/anyharness-lib/src/live/sessions/actor/config/handle.rs
+++ b/anyharness/crates/anyharness-lib/src/live/sessions/actor/config/handle.rs
@@ -2,6 +2,8 @@ use agent_client_protocol as acp;
 use anyharness_contract::v1::ConfigApplyState;
 
 use crate::domains::sessions::launch_intent::ResolvedLaunchIntent;
+use crate::domains::sessions::live_config::controls::option_matches_key;
+use crate::domains::sessions::live_config::{NormalizedControlKind, LEGACY_MODE_COMPAT_CONFIG_ID};
 use crate::live::sessions::actor::config::apply::{
     apply_mode_via_direct_setter_legacy, apply_specific_config_option, try_apply_model_preference,
 };
@@ -9,7 +11,8 @@ use crate::live::sessions::actor::config::confirmation::config_value_matches_cur
 use crate::live::sessions::actor::config::persist::persist_requested_config_value_if_changed;
 use crate::live::sessions::actor::config::queue::queue_pending_config_change;
 use crate::live::sessions::actor::config::selection::{
-    find_select_option_by_purpose, find_select_option_for_request, select_option_values,
+    find_select_option_by_purpose, find_select_option_for_request, into_raw_pending_option,
+    select_option_values,
 };
 use crate::live::sessions::actor::config::types::{
     tracked_config_purpose, ConfigApplyOutcome, ConfigPurpose,
@@ -19,6 +22,12 @@ use crate::live::sessions::actor::state::{SessionActor, SessionStartupState};
 /// Applies the immutable create-time intent and accepts only a value that the
 /// live harness statement both advertises and positively confirms. Model is
 /// first; controls follow the harness-advertised order.
+///
+/// The one carve-out is per-model value narrowing on a NON-posture control: a
+/// quality knob the live statement still surfaces but whose requested value it
+/// no longer offers under the applied model launches at the session default
+/// instead of failing the start. Posture controls and control ids the live
+/// statement never surfaced at all stay fail-closed.
 pub(in crate::live::sessions::actor) async fn apply_resolved_launch_intent(
     conn: &acp::ConnectionTo<acp::Agent>,
     native_session_id: &str,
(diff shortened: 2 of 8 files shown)


pablonyx:
Reviewer follow-ups recorded (both P2, non-blocking, out of scope here):

- **P2-4**: a `mode` select option whose value only exists in `legacy_mode_state` is now silently dropped instead of loudly failing — the ordered config-option loop classifies it Drop before the legacy branch can try `apply_mode_via_direct_setter_legacy` (which `apply.rs:362-370` supports on the live-mutation path). Requires the two mode surfaces to diverge; not observed with any current harness.
- **P2-5**: `initial_control_disposition` checks raw membership without `resolve_model_variant_value`; unreachable today (variants only apply to ConfigPurpose::Model, and the model travels `intent.model_id`), but a future 'model as a control value' change must revisit the disposition.

pablonyx:
## Verdict: MUST-FIX ITEMS

Read-only adversarial review (no builds/tests run locally — CI on `8ff09c6` is fully green, all 24 checks pass). Branch is based on `0cf2ed4` (the #2103 client fix), 2 commits behind `main`, MERGEABLE.

The core mechanism is right and the round-2 rework is a genuine improvement: computing the disposition **before** anything goes on the wire (`initial_control_disposition`, `handle.rs:202-219`) is what keeps "offered-but-refused" fatal, and the `dropped_per_model_control_is_excluded_from_the_final_aggregate_check` test has a real negative control. What I cannot sign off on as-is is the **scope** of the carve-out: the code drops far more than the per-model value narrowing the PR body describes, and it does so uniformly across permission-shaped controls.

---

### 1. The soft-drop is not scoped to per-model value narrowing — it silently downgrades permission/posture controls — P1

`handle.rs:72-90` applies `Drop` to **every** control id, and `handle.rs:120-157` extends it to the legacy `mode` surface. Concretely this means:

- `collaboration_mode = plan` (values `chat|plan`, see `actor/tests/config_restore.rs:141`, `service/create.rs:545`) — if the live statement doesn't offer `plan`, the session launches in the harness default and starts **executing** when the user explicitly asked for plan-only.
- `mode = agent-full-access` / any control classified `NormalizedControlKind::Mode` — `live_config/controls.rs:140-152` classifies anything whose id-or-name contains `mode` **or `approval`**. Codex-shaped `approval_policy` / `sandbox_mode` controls therefore fall inside the drop set.

The reported defect (`reasoning_effort` losing `max` under `gpt-5.5`) is a *quality knob*. Dropping a quality knob to a default is defensible. Dropping a *posture* knob to a default is a silent privilege/behavior change that the user explicitly selected against, and under the cutover's honest-absence philosophy that is strictly worse than refusing the start. Before this PR both were fatal; after it, both are silent.

Requested: gate the carve-out on control class. Keep `ConfigPurpose::Mode`, `collaboration_mode`, and anything classified `Mode`/`CollaborationMode` fail-closed; allow the drop for the narrowing classes (`Effort`/`Reasoning`/`FastMode`/`Extra`). If Pablo's ruling was deliberately "drop everything", say so explicitly in the PR body and in `MODELS.md` — right now the spec text reads as if only value narrowing is in scope.

### 2. The `pending.is_empty()` guard was deleted outright — unknown control **ids** now no-op silently — P1

`handle.rs:158-170` replaces
```rust
anyhow::ensure!(pending.is_empty(), "requested controls are absent from the live {agent_kind} session: {:?}", ...)
```
with an unconditional drop loop. That is a strictly broader change than the PR body ("a control value the live statement does not offer"), and `MODELS.md:185-196` documents only the *value* case, not the never-surfaced-*id* case (only the inline code comment at `handle.rs:158-159` mentions it).

I want to be precise about the #2103 masking question, because it is narrower than it first looks: create-time admission is **untouched** and still fail-closed on unknown ids/values against the harness observation (`launch_options/validation.rs:43-65`), and `SessionService::validate_persisted_launch_intent` (`sessions/service/mod.rs:149-173`) re-runs it before every real start. So the exact #2070/#2103 stale-normalized-key outage class still gets refused at admission — this PR does **not** mask it.

What it *does* newly mask is the class where the observation and the live statement disagree on the control-id vocabulary (stale/differently-probed observation, harness renames an id, probe env differs from launch env). Today that was a loud start failure; after this PR every user selection for that control silently no-ops forever, with a `tracing::warn!` as the only signal and no alarm-shaped metric. Given the launch-options program is exactly a "target-observed vocabulary" contract, that failure mode deserves either (a) staying fatal for never-surfaced ids while soft-dropping only unoffered values on ids that *do* exist live, or (b) an explicit statement of the tradeoff in the PR body plus a dashboard/alert on `membership_dropped` volume. Neither is present.

### 3. `requested_mode_id` clearing keys off the literal string `"mode"` — misses every other mode-purpose id — P2 (P1 on the fork path)

`startup.rs:363`:
```rust
if dropped_control_ids.iter().any(|id| id == "mode") {
```
A mode-purpose control dropped in the ordered config-option loop pushes its **raw** id (`sandbox_mode`, `approval_policy`, …), never the literal `"mode"`, so the requested/current divergence this block exists to prevent survives untouched for exactly those controls. Use `tracked_config_purpose(config_id, find_select_option_for_request(...))` / `option_matches_purpose(_, ConfigPurpose::Mode)` instead of a string compare.

Worth noting the blast radius differs by path: a freshly created session has `requested_mode_id: None` (`service/create.rs:261`), so the clear is a no-op there — the block only does real work on **forks**, which inherit `parent.requested_mode_id` (`sessions/runtime/fork/mod.rs:299`) and still reach `apply_resolved_launch_intent` because a fresh child has no live-config snapshot row. So the one path where this matters is the one the literal-`"mode"` check is most likely to miss.

### 4. Nothing tells the user their explicit selection was ignored — P2

The only signals are `tracing::warn!` + the `session.initial_config.dropped` bounded event (`handle.rs:79-85`, `122-129`, `162-168`). The client learns the *resulting* value from the live-config snapshot, so the picker is honest about current state, but there is no notice anywhere saying "the value you chose is unavailable for this model." `SessionEventSink` exposes only `config_option_update` / `session_state_update` (`sink/config.rs:16-22`) — no notice channel is used. For `reasoning_effort` this is tolerable; combined with finding 1 it is not. At minimum, the persisted immutable intent still records `reasoning_effort=max` forever with no marker that it was never honored.

### 5. `initial_control_disposition` has no model-purpose guard — P2 (defense-in-depth)

`handle.rs:210` calls `find_select_option_for_request`, which falls back to the model-purpose option when `config_id == "model"` (`selection.rs:157-162`). If a model-purpose entry ever reaches `intent.control_values`, it would be silently dropped here — bypassing the model fail-closed branch at `handle.rs:30-58`, its `resolve_model_variant_value` handling, and the `apply_model_via_direct_setter` fallback. Today this is unreachable because `controls_from_config_json` strips `id == "model" || category == "model"` from the observation (`launch_options/service.rs:180-182`), but that filter is *narrower* than `option_matches_purpose(_, Model)` (`controls.rs:130-133` also matches on `label.contains("model")`). A one-line `if option_matches_purpose(option, ConfigPurpose::Model) { return Apply }` makes the invariant local instead of dependent on a distant filter.

### 6. Test coverage stops at the pure functions — the seam itself is unpinned — P2

All three new tests (`tests/config_launch_intent.rs`) exercise `initial_control_disposition`, `intent_without_dropped_controls`, and `ensure_resolved_launch_intent_confirmed` in isolation. Nothing covers `apply_resolved_launch_intent` end-to-end, so none of the following would fail if regressed:

- the legacy-`mode` drop branch (`handle.rs:120-131`) — including that it pushes `"mode"` and does **not** bail;
- the deleted `pending.is_empty()` guard (`handle.rs:158-170`) — the behavior change with the largest blast radius has zero test;
- the `requested_mode_id` clear in `startup.rs:359-381`;
- that an *offered* value whose setter read-back refuses is still fatal **at the seam** (only the pure `Apply` disposition is asserted).

The negative controls that exist are good; they just guard the two smallest surfaces. `dropped_per_model_control_is_excluded_from_the_final_aggregate_check` genuinely fails on revert; `disposition_drops_only_values_the_live_statement_does_not_offer` mostly asserts that a brand-new function exists.

### 7. Transient inconsistent `session_state_update` on the dropped-mode path — P2 (nit)

`startup.rs:366-374` persists and emits `session_state_update` with `requested_mode_id: None` while `current_mode_id` is still the stale value carried from the session record; `emit_live_config_update` refreshes `current` only afterwards (`startup.rs:389-398`). Clients see one event with a half-updated pair. Folding the clear into the later `emit_live_config_update` write would emit one consistent state.

### 8. Migration / wire impact — clean

No schema changes, no contract/`anyharness-contract` type changes, no generated-artifact churn. `apply_resolved_launch_intent`'s signature change (`()` → `Vec<String>`) is crate-internal with a single call site. The two new observability names (`session.initial_config.dropped`, `membership_dropped`) are documented in `OBSERVABILITY.md` and carry only session identity + config key — the bounded-event contract is respected, and dropping the value from the round-1 log was the right call. The `Cargo.lock` `proliferate 0.4.18 → 0.4.20` line is a lockfile refresh catching up to the released version, not a functional change.

---

### Merge consideration

**This PR changes the Rust runtime, so it requires a desktop/runtime redeploy to take effect. It did not ride tonight's hotfix.** Production is already fixed client-side by #2103 (deployed), so nothing in production is waiting on this — it is defense-in-depth for the case where the client's observed-key filter is correct but the *live* session narrows further than the harness-level observation. That means the shipping bar here should be "correct and scoped", not "fast": merging it does nothing for users until the next runtime ship anyway.

Fresh adversarial review (independent of author).

pablonyx:
## Re-review at `fb7318e` — Verdict: LOOKS-CORRECT

Re-read the full diff `8ff09c6..fb7318e` adversarially rather than taking the summary on trust. All four MUST-FIX claims hold up. CI green on the head (1 path-filtered skip, `Production compose smoke`). Two P2 notes below; neither blocks.

### Verified

**Finding 1 (posture soft-drop) — fixed, and the classification is sound for real harnesses.** `handle.rs:227-242`: `is_posture_control` = literal `LEGACY_MODE_COMPAT_CONFIG_ID` ∨ `option_matches_key(Mode)` ∨ `option_matches_key(CollaborationMode)`. Traced against `controls.rs:140-152`: `sandbox_mode` and `approval_policy` both match Mode (`label.contains("mode")` / `label.contains("approval")` over `"{id} {name}"`), and `collaboration_mode` matches by exact id, so the two controls I named are genuinely covered. The legacy `mode` surface is now fatal outright (`handle.rs:133-141`) instead of dropped. `fast_mode` is deliberately excluded from Mode by the classifier and lands in the drop set — correct, it's a speed knob, not posture.

**Finding 2 (`pending.is_empty()`) — restored verbatim, at the same position as pre-PR**, same message and same `pending.keys().collect::<Vec<_>>()` payload. This is byte-equivalent fail-closed behavior to the pre-#2105 code, not an approximation.

**Finding 3 (`requested_mode_id` clear) — correctly deleted as dead, and I checked the fork path specifically since that was my objection.** The deletion is sound for a reason the comment states only loosely, so recording the actual proof: `requested_mode_id` is written only through `ConfigPurpose::Mode`, which `tracked_config_purpose` derives from `is_mode_config_request` = `id == "mode" ∨ option_matches_purpose(_, Mode)`. That set is a strict subset of `is_posture_control`, so every control capable of writing `requested_mode_id` is now unconditionally fatal on an unoffered value and can never appear in `dropped_control_ids`. The projection therefore cannot diverge — on forks (which inherit `parent.requested_mode_id` and do reach this seam) as well as fresh creates. My old finding 7 is moot with the block gone.

**Finding 6 (seam tests) — real, and each case fails without the corresponding fix.** `launch_intent_seam_drops_only_unoffered_quality_values` drives `apply_resolved_launch_intent` itself:
- `web_search` (never surfaced) → expects `Err` containing `web_search`; at the previous head this returned `Ok`. Pins finding 2.
- `approval_policy = never` (surfaced, value not offered) → expects `Err` containing `approval_policy`; at the previous head this was a silent `Drop` → `Ok`. Pins finding 1, and it exercises the real classifier on a fixture with `category: None`, so the label heuristic is genuinely under test rather than short-circuited by a category tag.
- `reasoning_effort = max` → expects `Ok`. Genuine negative control: it fails if anyone widens the refusal back to everything.

Also correct that the fake connection is never asked to answer — every case resolves pre-wire, so the test isn't secretly depending on transport behavior. Deleting the old `"fast-mode"` disposition assertion was the right call (see note A).

**Spec/PR body** now describe the actual three-way policy, and the removed "a dropped mode also clears the requested-mode projection" sentence left no stale references anywhere in `specs/`. `membership_rejected` is the code used for the new `Refuse` arm, which is already enumerated in `OBSERVABILITY.md` — no observability delta needed for this round, and `session.initial_config.dropped` is still emitted on the surviving Drop path.

### Note A — `initial_control_disposition` returns `Drop` for a never-surfaced id, which is no longer the seam's policy — P2

`handle.rs:212-244`: for a `config_id` absent from `config_options`, `offered` is false and `is_posture_control` is false, so the function returns `Drop`. The seam is nonetheless correct because the ordered loop only iterates ids the live statement surfaced, and anything left falls to the restored `pending.is_empty()` guard. But the pure function's contract now silently depends on its caller's loop, and the docstring doesn't say so. A future second caller consulting it for an arbitrary id would get "soft-drop" for exactly the vocabulary-disagreement case the PR now declares fatal. Cheap hardening: return `Refuse` when `find_select_option_for_request` is `None`, or state the precondition in the docstring.

### Note B — posture is a substring heuristic, so the drop set is default-open — P2

`is_posture_control` inherits `option_matches_key`, which recognizes posture by `"mode"` / `"approval"` appearing in `"{id} {name}"`. Codex's real ids satisfy that, so the shipped behavior is right today. But a harness exposing posture under `sandbox`, `sandbox_policy`, `permissions`, `network_access`, `write_access`, or similar would fall through to the soft drop — and `MODELS.md` now promises "sandbox mode" is protected, which is true only because codex spells it with `_mode`. The Mode matcher's `&& !label.contains("model")` exclusion is a second way a posture control can fall out.

Given the carve-out exists for exactly one observed phenomenon, the safer shape is to invert the default: soft-drop only an explicit allowlist of narrowing-prone quality kinds (`Effort`, `Reasoning`), refuse everything else. That makes a misclassified new control fail loudly instead of silently, which is the direction this whole cutover has otherwise chosen. Not a blocker — residual risk is bounded by create-time validation still refusing unknown ids and values — but worth doing before another harness's controls land.

### Deliberate deferrals — agreed

Skipping the user-facing drop notice (no notice channel exists on `SessionEventSink`) and the model-purpose guard in `initial_control_disposition` (unreachable while the observation strips model controls) is reasonable, and documenting both in the PR body is the right disposition.

### Merge consideration (unchanged)

This still changes the Rust runtime and needs a desktop/runtime redeploy; it did not ride tonight's hotfix. Production remains fixed client-side by #2103, so nothing is waiting on it and merging it this morning delivers nothing to users until the next runtime ship. With the posture carve-out now scoped and the never-surfaced-id guard restored, I no longer see a reason to hold it — but the first post-launch runtime ship is still the natural home.

Fresh adversarial review (independent of author).

PR #2113, opened 2026-08-20
ci(workflows): cache cargo builds and drop redundant check step

pablonyx:
## Summary

- `cargo-check` ("Cargo check & test") is the only Rust job in `ci.yml` with no `rust-cache`, so it compiles all 750 external dependency crates from scratch on every run.
- Cold, the job takes 11m31s (run 32335721001): 2m59s "Cargo check" + 7m15s "Cargo test" + 33s "Cargo test internal diagnostics export".
- `cargo test --workspace` already compiles the same code as `cargo check --workspace`, so the check step is pure duplication of the test step's compilation.

## Changes

1. Add `- uses: Swatinem/rust-cache@v2` right after "Setup Rust stable", matching the unpinned `@v2` form already used in the `qualification` job.
2. Delete the "Cargo check" step. "Cargo test" and "Cargo test internal diagnostics export" (and its explanatory comment) are untouched.

## Expected impact

Expected warm time is ~4m (inference). `rust-cache` only saves on a *green* job, so the first green run after this merges will prime the cache; the second green run is what actually measures the warm time — that number should be confirmed there, not assumed from this PR's own first run.

## Test semantics

No test semantics change. `cargo test` subsumes `cargo check` diagnostics (compiling test code is a superset of compiling non-test code), so removing the standalone check step does not reduce coverage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -263,8 +263,7 @@ jobs:
       - name: Setup Rust stable
         uses: dtolnay/rust-toolchain@stable
 
-      - name: Cargo check
-        run: cargo check --workspace
+      - uses: Swatinem/rust-cache@v2
 
       - name: Cargo test
         run: cargo test --workspace


pablonyx (commented):

**Scope check (item 1):** `gh pr diff 2113` is exactly 2 lines removed / 1 added in `.github/workflows/ci.yml` — deletes the "Cargo check" step and adds `- uses: Swatinem/rust-cache@v2` right after "Setup Rust stable" (now at `.github/workflows/ci.yml:266` on the PR head). Nothing else touched. Matches the body.

**Finding 1 (blocking) — the redundancy claim is false for `#[cfg(not(test))]` code.**

The body claims "`cargo test` subsumes `cargo check` diagnostics... does not reduce coverage." That's true for ordinary code, but `cargo test --workspace` compiles with `cfg(test)` active, which means every `#[cfg(not(test))]` block is *elided entirely* — never compiled, never type-checked — by `cargo test`. Only a plain `cargo check`/`cargo build` (no test cfg) exercises that code. This workspace uses that pattern non-trivially, e.g.:

- `anyharness/crates/anyharness-lib/src/app/mod.rs:239` — `#[cfg(not(test))] let automatic_poke_engine = Some(launch_probe_service.clone());`
- `anyharness/crates/anyharness-lib/src/app/mod.rs:500` — `#[cfg(not(test))] workspace_archive_service.clone().spawn_startup_pass();`
- `anyharness/crates/anyharness-lib/src/app/mod.rs:583,588` — `#[cfg(not(test))] loop_scheduler.clone().spawn();` and `agent_runtime.clone().spawn_startup_pass();`
- `anyharness/crates/anyharness-lib/src/live/sessions/manager/startup.rs:16`, `.../actor/turn/active.rs:19`, `.../manager/mod.rs:160` — production-only timeout constants
- plus `anyharness/crates/proliferate-diagnostics-client/src/fallback/unix.rs:539` and `anyharness/crates/anyharness-lib/src/domains/workspaces/archive/mod.rs:114`

I checked whether any other PR-gating job compiles these paths in non-test mode: it doesn't. `desktop-frontend` only runs `pnpm run build` (no cargo at all). `sdk-build` only touches the TS SDK. The only jobs that run `cargo build --release` on real (non-test) cfg are in `release-desktop.yml` and `release-runtime.yml`, both gated on `push: tags: desktop-v*/runtime-v*` (not PRs), and `server-ci.yml`, which is path-filtered to `server/**`/`catalogs/**`/lockfiles and would not fire on an `anyharness/**`-only change anyway. So after this PR, a compile error inside any of the blocks above would pass every PR check and only surface at tag-release time — a materially worse failure mode than today.

Recommend: keep the "Cargo check" step (it will itself get cheap once `rust-cache` populates, since it's mostly the same dependency compilation `cargo test` also pays for) rather than deleting it. Deleting it does not preserve test semantics as claimed.

**Finding 2 (non-blocking) — rust-cache config is consistent with existing convention, no poisoning risk.**

`Swatinem/rust-cache@v2` unpinned matches the exact line already in `candidate-build-handoff` (`.github/workflows/ci.yml:234` on main) and 5 other call sites (`catalog-probe.yml:76`, `release-e2e.yml:197,590,766,1122`) — this PR isn't introducing a new pinning inconsistency, just following the repo's existing (mixed) practice. rust-cache scopes its key by `github.job` by default, so `cargo-check`'s cache is isolated from `candidate-build-handoff`'s. GitHub Actions cache *save* is scoped to the triggering ref/branch — a PR branch cannot overwrite the cache entry belonging to `main`, so there's no cache-poisoning path from branches into main; PRs can only *read* main's cache as a fallback. This part is fine.

**Finding 3 (informational) — the cited timings check out, but "warm" hasn't actually been measured yet, and a downstream PR is already treating the cold number as warm.**

Requoted directly from `gh run view --json jobs` (not the PR body's prose):

- Cold baseline, run [32335721001](https://github.com/proliferate-ai/proliferate/actions/runs/32335721001/job/96324803239) (push to `main` @ 8b6dc1f, before this PR): job total 05:28:27→05:39:58 = **11m31s**. Steps: Cargo check 05:29:08→05:32:07 = 2m59s; Cargo test 05:32:07→05:39:22 = 7m15s; diagnostics export 05:39:22→05:39:55 = 33s. This matches the PR body's numbers exactly.
- This PR's own (only) CI run, run [32405379745](https://github.com/proliferate-ai/proliferate/actions/runs/32405379745/job/96551626337) (`pull_request`, head 5fae76f0): job total 19:19:17→19:26:41 = **7m24s**. Steps: rust-cache 19:20:02→19:20:16 = 14s; Cargo test 19:20:16→19:26:23 = 6m7s; diagnostics export 19:26:23→19:26:38 = 15s.

That 7m24s run is the *first ever* invocation of `rust-cache` for this job on any ref (no cache existed on `main` or the branch to restore), so it's a cache-miss/cold run, not the warm number the body itself is careful to say still needs confirming on a second green run. The 14s rust-cache step duration (fetch-nothing-then-save) is consistent with a miss, not a multi-hundred-MB restore. Worth noting: PR #2118 (branched from this one) already cites this same run/job as "the warm `rust-cache` baseline (7m24s)" in its own body — that mislabel should be corrected before anyone uses it as a comparison point, since it isn't actually a cache-hit measurement.

**Finding 4 (interaction risk) — #2116/#2118.**

Both are explicitly marked "never merge" / "not for merge" experiments branched *from* `ci/cargo-cache-and-dedupe`, not stacked independently onto main, so there's no risk of them landing on main ahead of or instead of this PR. The only actionable item is the "warm" mislabeling in #2118 noted above.

---
Net: the caching addition and job structure are sound and the claimed cold-run speedup is real and verified, but the deletion of the "Cargo check" step trades away real coverage (production-only `cfg(not(test))` code paths) that nothing else in the PR-gating pipeline replaces. Please keep that step (or replace it with an equivalent plain, non-test `cargo check`/`cargo build` pass) before merging.

PR #2116, opened 2026-08-20
ci(cargo-cache): add sccache alongside rust-cache

pablonyx:
## Summary
- Experiment (never merge): measure whether `mozilla-actions/sccache-action` eliminates workspace-crate recompilation on non-Rust PRs, which `Swatinem/rust-cache` (base branch, #2113) deliberately does not cache.
- Adds sccache to the `cargo-check` job only: `RUSTC_WRAPPER=sccache`, `SCCACHE_GHA_ENABLED=true`, `CARGO_INCREMENTAL=0` (incremental defeats sccache), keeping `Swatinem/rust-cache` in place for the registry/deps cache.

## Test plan
- [ ] Run 1 (this commit): primes the sccache GHA cache from a cold/empty sccache state.
- [ ] Run 2: push a whitespace/comment-only commit touching a non-Rust file to simulate a docs/UI-only PR, and compare cargo-check wall time + sccache hit rate against the 7m24s warm `Swatinem/rust-cache`-only baseline from #2113.
- [ ] Record both run links and sccache stats output in the PR thread; do not merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -251,6 +251,18 @@ jobs:
   cargo-check:
     name: Cargo check & test
     runs-on: ubuntu-latest
+    # EXPERIMENT (exp/rust-sccache, base ci/cargo-cache-and-dedupe): measure
+    # whether sccache eliminates the workspace-crate recompilation that
+    # Swatinem/rust-cache deliberately leaves on the table (it discards
+    # workspace-crate build artifacts by design, so unchanged crates still
+    # recompile on every PR). sccache keyed by rustc invocation inputs should
+    # hit for those crates even when the target dir cache does not carry them
+    # forward. CARGO_INCREMENTAL=0 is required: incremental compilation
+    # defeats sccache's ability to reuse cached object output.
+    env:
+      RUSTC_WRAPPER: sccache
+      SCCACHE_GHA_ENABLED: "true"
+      CARGO_INCREMENTAL: "0"
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
@@ -262,6 +274,9 @@ jobs:
 
       - name: Setup Rust stable
         uses: dtolnay/rust-toolchain@stable
+
+      - name: Setup sccache
+        uses: mozilla-actions/sccache-action@fc920bf0ec8de6ee65d409111f7ec508035751ba # v0.0.11
 
       - uses: Swatinem/rust-cache@v2
 
--- a/README.md
+++ b/README.md
@@ -209,3 +209,5 @@ Contributing? See the [Contribution Guide](./CONTRIBUTING.md).
 ## License
 
 [AGPL-3.0](./LICENSE)
+
+<!-- exp/rust-sccache: non-Rust no-op commit to measure sccache hit rate on run 2 -->


pablonyx (commented):

Reviewed fresh (no prior review existed on this PR — checked `gh api .../pulls/2116/reviews`, empty). Verdict is CHANGES-NEEDED on process/documentation grounds, not on technical soundness — the mechanism works and the cited numbers check out. This should not merge as-is because the PR's own body says "never merge," the test-plan checklist is entirely unchecked, and the promised in-thread evidence was never posted.

## 1. Diff mechanics (.github/workflows/ci.yml:251-278)
- `mozilla-actions/sccache-action@fc920bf0ec8de6ee65d409111f7ec508035751ba` — verified this SHA is the dereferenced commit for tag `v0.0.11` (`git ls-remote` confirms `refs/tags/v0.0.11^{}` == this SHA). Correctly pinned to a full SHA with a version comment, not a mutable tag (contrast with Greptile's supply-chain note on #2113's still-unpinned `Swatinem/rust-cache@v2`).
- `RUSTC_WRAPPER=sccache` + `SCCACHE_GHA_ENABLED=true` + `CARGO_INCREMENTAL=0` wiring is correct and matches upstream sccache guidance (incremental does defeat sccache's object-level cache).
- rust-cache vs sccache overlap: not redundant, genuinely complementary. Confirmed via actual stats pulled from the run logs (`gh run view <id> --log`, "Post Setup sccache" step, `--show-stats --stats-format=json`): warm run 32411990807 shows `"compile_requests":49, "requests_not_cacheable":39, "not_cached":{"missing input":4,"-":2,"crate-type":33}, "cache_hits":{"Rust":10}`. rust-cache already skips recompiling the ~750 dependency crates via Cargo fingerprinting (no rustc invocation at all for those), so sccache only ever sees the small residual set of workspace-crate rustc invocations — of which 33/49 (67%, i.e. the claimed "~2/3") are non-cacheable due to `crate-type` (sccache's known limitation: it can't cache invocations that produce multiple crate-types like rlib+cdylib in one rustc call, which several workspace crates use). Of the 10 that were cacheable/executed, 10/10 hit (100%, confirming the claim). This stats JSON is **not surfaced anywhere in the PR body or comments** — it only exists buried in raw Actions logs, contradicting the PR's own test plan ("record ... sccache stats output in the PR thread").

## 2. Env-key-invalidates-rust-cache gotcha
Not documented anywhere — not in the PR body, not in code comments, not in any PR comment (checked #2116, #2113, #2118, #2128; no human comments exist beyond Vercel/Greptile bots). The code comment at ci.yml:253-262 explains *why* CARGO_INCREMENTAL=0 is needed but never mentions that adding `RUSTC_WRAPPER`/`CARGO_INCREMENTAL` env vars also busts Swatinem/rust-cache's own cache key (rust-cache hashes `CARGO_`/`RUST`-prefixed env vars into its key), so run 1's 11m42s is actually a **double-cold** run (sccache cold AND rust-cache cold), not a clean "sccache cold tax" number. A future reader diffing this against #2113's 7m24s warm baseline would misattribute the full delta to sccache alone. #2128 (the sccache+nextest combo PR) does call this out explicitly in its own body ("First run is expected to be double-cold... not representative") — #2116 should have said the same but doesn't. Final key design here does not avoid re-priming on unrelated env changes; it's a one-time cost accepted implicitly, never argued as acceptable.

## 3. Requoted timings (warm-vs-warm confirmed via `gh run view --json jobs`)
- Run 1 (5b91ea8, priming): job "Cargo check & test" 19:50:06→20:01:48 = **11m42s** exactly, matches claim. https://github.com/proliferate-ai/proliferate/actions/runs/32410803549
- Run 2 (d42588a, no-op non-Rust commit): 20:03:15→20:09:13 = **5m58s (358s)** exactly, matches claim. https://github.com/proliferate-ai/proliferate/actions/runs/32411990807
- Baseline (#2113, 5fae76f, only CI run it has): 19:19:17→19:26:41 = **7m24s (444s)**, matches the cited baseline. https://github.com/proliferate-ai/proliferate/actions/runs/32405379745
- 358s vs 444s = **-19.4%** (not -19.6%, close enough to be a rounding artifact — not worth blocking on).
- Could NOT verify "+-40s warm variance" / "375s observed baseline floor" anywhere — no such figures exist in any PR body/comment in this chain, and there is only **one** warm sample of the baseline (7m24s), not a distribution. Worse: #2113's own body explicitly cautions "the second green run is what actually measures the warm time — that number should be confirmed there, not assumed from this PR's own run," and that confirming second run was never done (#2113 has exactly one CI run, ever). So the entire -19.6%/-19.4% claim rests on a single baseline sample its own author flagged as unconfirmed. Treat the improvement as directionally real (sccache clearly does something — 100% hit rate on the crates it can reach) but not statistically established.

## 4. Stacking on #2113
Confirmed: `baseRefName: ci/cargo-cache-and-dedupe` = #2113, which is **OPEN, unmerged** (`mergedAt: null`). #2116 is a diff against an unlanded branch, not against main. Separately, #2128 (`exp/rust-sccache-nextest`, also based on `ci/cargo-cache-and-dedupe`) applies both this PR's sccache change and #2118's nextest change verbatim. Verified its combo number directly: run 32413255083 attempt 1 (cold, both keys new) = 20:17:07→20:27:24 = 10m17s; attempt 2 (rerun, warm) = 21:19:56→21:24:46 = **4m47s (287s)**, matching the claimed figure exactly and correctly obtained via a same-commit rerun (proper warm-vs-warm methodology, better than #2116's own single-sample approach). Note 287s is 16s worse than naive addition of the two independent deltas (444-87-86=271s), consistent with #2128's own hedge that the wins "should partially compose," not fully. Nothing here needs a rebase to be internally consistent, but since #2113 is unmerged, none of these deltas are measured against main's actual current state — merge order must be #2113 first, then re-baseline before any of this is meaningful for production CI.

## 5. Failure mode
Partially exercised, not fully. Real evidence from run 32410803549 (cold run) stats: `"cache_write_errors":207` against `"cache_writes":887` (~19% write failure rate under the cold run's write load) with `"cache_read_errors":0,"cache_timeouts":0`, and the job still completed successfully — this shows sccache tolerates GHA cache **write** contention/throttling gracefully (falls through to an uncached-but-successful compile). However, no read-side outage or full GHA-cache-service-unavailable scenario was exercised in either run, so it's untested whether a read-path failure degrades gracefully or hard-fails the wrapped rustc invocation. This matters more here than for rust-cache: rust-cache is purely additive around the job (fails open — a restore failure just means a cold build), whereas `RUSTC_WRAPPER=sccache` sits directly in the compiler invocation path (fails closed if the sccache binary/server can't start or errors hard) — there's no `sccache --show-stats` verification step or `continue-on-error` guard on the wrapper setup, so a broken sccache server would fail the entire cargo-check job rather than degrading to a plain-rustc compile.

## Bottom line
Numbers check out where I could verify them against actual CI job timestamps and raw sccache stats JSON (11m42s, 5m58s, 7m24s, 4m47s, 100% cacheable-hit-rate, ~2/3 crate-type-non-cacheable are all confirmed real). Two specific figures (+-40s variance, 375s floor) could not be traced to any evidence in this PR chain. Given the PR is explicitly labeled "never merge," has an unchecked test-plan, never posted its own promised stats to the thread, doesn't document the rust-cache-key invalidation side effect it triggered, and stacks on an unmerged base — this is fine as a completed experiment to inform a future real PR, but is not itself mergeable as written.

PR #2117, opened 2026-08-20
ci(cargo): capture per-crate build timings

pablonyx:
## Summary

Measurement-only experiment. Adds `cargo build --workspace --tests --timings=html,json`
to the `cargo-check` job before the existing `cargo test --workspace` step, and uploads
`target/cargo-timings/` as a CI artifact.

Goal: find out where the warm 7m24s "Cargo check & test" time (from #2113,
base branch `ci/cargo-cache-and-dedupe`) goes per crate, to inform whether/how
to split the 226k-line `anyharness-lib` crate (12 workspace crates, 750
external deps, anyharness-lib is 82% of workspace Rust).

Base is `ci/cargo-cache-and-dedupe` (not `main`) rather than that PR's own
branch, so this PR's CI runs restore the `Swatinem/rust-cache` entries that
branch's runs create, keeping the timing runs warm/representative instead of
cold-cache outliers.

## Do not merge

This PR exists only to produce a `cargo-timings` artifact from CI runs. It
adds no functional value on its own (the timings step and its upload are
pure overhead for a real merge) and should be closed once the data is
extracted.

## Test plan

- [x] CI run produces a `cargo-timings` artifact (HTML + JSON) on the
      `cargo-check` job
- [ ] Timing data downloaded and analyzed offline (see follow-up report)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -268,7 +268,11 @@ jobs:
       # EXPERIMENT (exp/rust-build-timings, never merge): capture per-crate
       # compile timing data to inform a future split of anyharness-lib.
       - name: Cargo build with timings
-        run: cargo build --workspace --tests --timings=html,json
+        # Stable cargo's --timings flag takes no value (the FMTS=html,json
+        # argument is nightly-only, behind -Z unstable-options); it always
+        # writes an HTML report (with embedded per-unit JSON data) to
+        # target/cargo-timings/.
+        run: cargo build --workspace --tests --timings
 
       - name: Cargo test
         run: cargo test --workspace
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -265,6 +265,11 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
 
+      # EXPERIMENT (exp/rust-build-timings, never merge): capture per-crate
+      # compile timing data to inform a future split of anyharness-lib.
+      - name: Cargo build with timings
+        run: cargo build --workspace --tests --timings=html,json
+
       - name: Cargo test
         run: cargo test --workspace
 
@@ -273,6 +278,17 @@ jobs:
       # evaluated without this step.
       - name: Cargo test internal diagnostics export
         run: cargo test -p proliferate-diagnostics-collector --features internal-dogfood-export
+
+      # EXPERIMENT (exp/rust-build-timings, never merge): upload the timing
+      # report produced above for offline analysis.
+      - name: Upload cargo timings
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: cargo-timings
+          path: target/cargo-timings/
+          if-no-files-found: warn
+          retention-days: 7
 
   sdk-build:
     name: Build SDK


pablonyx:
## Timings results (run 32411239902)

Artifact: `cargo-timings` (HTML report) from the "Cargo check & test" job, step "Cargo build with timings" (`cargo build --workspace --tests --timings`, stable cargo — the `=html,json` FMTS syntax is nightly-only and was reverted to bare `--timings` after the first run failed with `unexpected value 'html,json'`).

**Total compile wall clock (this step): 271.63s (4m32s)** — matches the "Cargo build with timings" step duration in the job (19:55:50 -> 20:00:22) almost exactly.

### Top 15 compile units by self-time

| crate | version | kind | seconds |
|---|---|---|---|
| anyharness-lib | 0.1.0 | lib (test) | 89.96 |
| anyharness-lib | 0.1.0 | lib | 70.41 |
| aws-lc-sys | 0.39.0 | build-script (run) | 65.57 |
| proliferate | 0.4.21 | lib (test) | 37.25 |
| proliferate | 0.4.21 | lib | 35.39 |
| gtk | 0.18.2 | lib | 19.46 |
| anyharness-contract | 0.1.0 | lib (test) | 16.98 |
| anyharness-contract | 0.1.0 | lib | 15.22 |
| zstd-sys | 2.0.16+zstd.1.5.7 | build-script (run) | 14.21 |
| tauri-utils | 2.9.3 | lib | 13.18 |
| gio | 0.18.4 | lib | 11.28 |
| zbus | 5.14.0 | lib | 10.03 |
| agent-client-protocol-schema | 0.13.6 | lib | 9.08 |
| libsqlite3-sys | 0.38.2 | build-script (run) | 8.16 |
| tauri-utils | 2.9.3 | lib | 7.94 |

(tauri-utils appears twice: same version compiled under two different feature-unification contexts.)

### anyharness-lib share

anyharness-lib's own two units (normal `lib` + `#[cfg(test)]` `lib (test)`) sum to **160.53s = 59.1% of the 271.63s wall clock**. More importantly: the `lib (test)` unit (89.96s, starts at 181.67s right after `anyharness-contract` finishes at 169.64s) is the single unit whose finish time (271.63s) *is* the wall clock — nothing else in the build graph runs after it. It is the critical-path tail with zero slack.

### Link-time share (proxy)

Cargo's stable `--timings` doesn't emit a separate "link" phase for final bin/test artifacts (`sections: null` for all of them) — the whole unit duration is compile+link combined. Summing all workspace-crate `(test)`/`bin` final-artifact units: **178.07s = 65.6%** of wall clock; including a couple of external test units too: 183.60s = 67.6%.

### Split-boundary read

Within anyharness-lib (240,918 lines total), `src/domains/` is 149,963 lines (62%). Inside domains/: `sessions` 45,254 lines, `agents` 30,233, `workspaces` 23,275, `workflows` 8,283 — those four alone are ~106k lines (44% of the whole crate). Only one workspace crate (`anyharness`, the bin) has a direct path-dependency on anyharness-lib; `proliferate`, `proliferate-worker`, and `proliferate-supervisor` don't depend on it at all, so they already build in parallel with it today.

Splitting `domains/sessions`, `domains/agents`, and `domains/workspaces` into their own crates (if they can be decoupled from each other — not verified here) would let each one's `#[cfg(test)]` compile run as an independent, parallel rustc process instead of one atomic ~90s unit. On even 3-4 parallel job slots that could plausibly cut the tail from ~90s to the size of the largest remaining chunk (rough order 25-35s), for an estimated 50-60s off the observed 271s compile wall clock — plus a proportional cut to the 70.41s normal-mode compile that currently gates the `anyharness` bin.

### Note: unrelated flake

The later `Cargo test` step failed on two pre-existing flaky tests in `proliferate-diagnostics-collector` (`concurrent_maximum_and_degenerate_ingest_stays_under_the_process_rss_ceiling`, `inherited_capability_process_interface_never_exposes_or_persists_the_token`) — unrelated to this experiment's changes; the timings artifact was already captured successfully in the prior step.

Never merging this PR — see description.

PR #2120, opened 2026-08-20
perf(ci): shard product-client tests 4-way

pablonyx:
## Summary

Shard the slowest step in the `shared-frontend-packages` CI job 4-way so the job's wall clock drops from ~8m38s-9m13s to ~3m07s.

**Latest state (housekeeping split out + 4-way shard): max shard job wall clock 187s (3m07s), down from the ~338s "Test product client" step / ~7m34s-9m13s combined job baseline. 7s over the 3-minute target — see "Round 3" below.**

### Round 1 — 3-way shard, steps left in the same job

- Job `shared-frontend-packages` originally ran 8m38s-9m13s; one step, "Test product client" (~338s), was the pole.
- Added `strategy: matrix: shard: [1, 2, 3]`, ran "Test product client" with `--shard=${{ matrix.shard }}/3`, gated the steps that step doesn't need (typecheck, the two python-only checks, "Test ProductClient domain", "Test workflow authoring surface") to `if: matrix.shard == 1`.

**Bug found and fixed:** `pnpm --filter @proliferate/product-client test -- --shard=N/3` resolved to `vitest run -- --shard=N/3` — pnpm inserts its own `--` ahead of forwarded script args, and vitest's CLI parser treats that leading `--` as an end-of-options marker, so `--shard` was never applied. All 3 "shards" silently ran the **full** `1093 files / 8168 tests` each; wall clock got *worse* (up to 522s) from 3x redundant work plus runner variance. Fixed by invoking vitest directly — `pnpm --filter @proliferate/product-client exec vitest run --shard=${{ matrix.shard }}/3` — same pattern the sibling "Test ProductClient domain" step already used.

After the fix, 3-way sharding measured correctly:

| Shard | Job wall clock | vitest summary |
|---|---|---|
| 1 | 227s (3m47s) | 365 files / 2692 tests, vitest duration 113.82s |
| 2 | 217s (3m37s) | 365 files / 2843 tests, vitest duration 147.04s |
| 3 | 216s (3m36s) | 363 files / 2633 tests, vitest duration 148.48s |

Reconciliation: 365+365+363 = 1093 files, 2692+2843+2633 = 8168 tests — matches baseline exactly. Max (new lane time) = 227s, still over the 3-minute target. Shard 1 was the pole because it alone carried the gated housekeeping steps (typecheck, python checks, the two other vitest invocations) on top of its own shard.

### Round 2 (this push) — split housekeeping into its own job + widen to 4 shards

Per follow-up direction: moved all housekeeping (Typecheck shared packages, both python-only checks, "Test ProductClient domain", "Test workflow authoring surface") into a new `shared-frontend-checks` job with its own checkout/install/build, so every `shared-frontend-packages` matrix shard is pure install+build+test. Since the fixed per-shard setup+build overhead (~65-70s) plus a 3-way shard's own vitest time (~148s) was already going to clear 180s regardless of the housekeeping move, widened to 4 shards in the same push rather than spending a second CI round-trip confirming that.

**Measured (run [32413118164](https://github.com/proliferate-ai/proliferate/actions/runs/32413118164)):**

| Job | Wall clock | vitest summary |
|---|---|---|
| Shared frontend checks | 133s | domain: 83 files/940 tests; workflow authoring: 2 files/29 tests |
| Shard 1/4 | 187s (3m07s) | 274 files / 2035 tests, vitest duration 110.24s |
| Shard 2/4 | 164s (2m44s) | 274 files / 1990 tests, vitest duration 90.72s |
| Shard 3/4 | 186s (3m06s) | 274 files / 2242 tests, vitest duration 118.71s |
| Shard 4/4 | 175s (2m55s) | 271 files / 1901 tests, vitest duration 108.86s |

**Reconciliation:** product-client shards: 274+274+274+271 = **1093 files**, 2035+1990+2242+1901 = **8168 tests** — matches the unsharded baseline exactly. Housekeeping job: 83/940 + 2/29 — also matches baseline exactly. No dropped tests across either job.

**New lane time = max(187, 164, 186, 175) = 187s (3m07s).** Still 7s over the 180s target. Per-shard vitest test time is now down to 91-119s (well under target on its own); the remaining pole is each shard's fixed checkout+install+build overhead (~64-71s, dominated by "Build shared packages" fluctuating 35-47s run to run) — i.e. we're now bound by per-job GitHub Actions runner setup/build variance rather than by the test suite itself. Closing the last ~7s would mean caching the build output across shards (e.g. build once in a separate job and pass `dist/` as an artifact into each shard) rather than more sharding — not attempted here to avoid over-engineering the experiment further.

## Shard-safety audit

- `vitest.config.ts`: no `sequence`/`singleFork`/`singleThread`/custom `poolOptions` — default vitest parallelism already runs test files concurrently within one invocation, so `--shard` just narrows the file set handed to that same machinery.
- `vitest.global-setup.ts` regenerates the gitignored `src/generated/agent-*.json` copies before collection. Idempotent and file-content-identical regardless of which shard runs it; each shard is a separate checkout/process, so there's no write race.
- `vitest.setup.ts` calls `setMeasurementSink(...)` per test file (last-writer-wins, explicitly documented as such) — no shared external state across files or shards.
- No `process.env` mutation, no fixed-path `writeFileSync`/`tmpdir()` usage, and no cross-file ordering dependency found in `src/**/*.test.{ts,tsx}` (1093 files).
- "Install dependencies" and "Build shared packages" stay ungated in every product-client shard: product-client imports `@proliferate/design`, `@proliferate/cloud-sdk`, and `@proliferate/cloud-sdk-react` through `package.json` `exports` that resolve only to `dist`, so the build is a real prerequisite of the sharded step.

## Caveat

Turning this job into a matrix (and splitting housekeeping into its own job) changes check name(s) (e.g. `Shared frontend packages (1)` instead of `Shared frontend packages`, plus a new `Shared frontend checks` check), which needs the branch-protection required-checks list updated when it is set (tracked in the rollout plan).

## Test plan

- [x] Compare per-shard wall time (max = new lane time) against the ~8m38s-9m13s baseline
- [x] Sum vitest "Tests" counts across the shards and confirm it equals the unsharded "Test product client" count (no silently dropped tests)
- [x] Confirm whether the job wall clock lands at or under the 3-minute target — landed at 187s, 7s over

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -420,7 +420,7 @@ jobs:
       - name: Enforce /login runtime budget
         run: node scripts/measure-login-runtime-budget.mjs
 
-  # EXPERIMENT (exp/frontend-shard): housekeeping that "Test product client"
+  # Housekeeping that "Test product client"
   # does not need (typecheck, the python-only checks, and the two other
   # focused vitest invocations) lives in its own job here so every matrix
   # shard below is pure install+build+test. Runs in parallel with the shards;
@@ -475,10 +475,10 @@ jobs:
       - name: Test workflow authoring surface
         run: pnpm --filter @proliferate/product-client exec vitest run src/components/workflows/builder-v2/WorkflowBuilderSurface.test.tsx src/components/workflows/main/WorkflowsMainSurface.test.tsx
 
-  # EXPERIMENT (exp/frontend-shard): "Test product client" was the pole of the
-  # combined job (~338s of ~8m38s-9m13s wall clock). Sharded 4-way via
-  # vitest's native --shard (3-way landed at 227s max shard wall clock, still
-  # over the 3-minute target — see PR body for the measured numbers). Every
+  # "Test product client" was the pole of the combined job (~338s of
+  # ~8m38s-9m13s wall clock). Sharded 4-way via vitest's native --shard
+  # (3-way still exceeded the 3-minute target; measured numbers in PR #2120).
+  # Every
   # other step this job used to run (typecheck, python checks, the two other
   # focused vitest invocations) moved to shared-frontend-checks above, so
   # every shard here is pure install+build+test. "Install dependencies" and
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -420,23 +420,14 @@ jobs:
       - name: Enforce /login runtime budget
         run: node scripts/measure-login-runtime-budget.mjs
 
-  # EXPERIMENT (exp/frontend-shard): "Test product client" is the pole of this
-  # job (~338s of the job's ~8m38s-9m13s wall clock). Sharded 3-way via
-  # vitest's native --shard to see whether the job's wall clock drops below
-  # 3 minutes. Only "Install dependencies" and "Build shared packages" are
-  # true prerequisites of "Test product client" (product-client imports
-  # @proliferate/design/@proliferate/cloud-sdk(-react) through package.json
-  # `exports` that resolve to `dist`, so the build must precede every shard).
-  # Typecheck, the python-only checks, and the two other focused vitest
-  # invocations are independent of sharding and gated to shard 1 only so
-  # shards 2-3 do not re-run work "Test product client" does not need.
-  shared-frontend-packages:
-    name: Shared frontend packages
+  # EXPERIMENT (exp/frontend-shard): housekeeping that "Test product client"
+  # does not need (typecheck, the python-only checks, and the two other
+  # focused vitest invocations) lives in its own job here so every matrix
+  # shard below is pure install+build+test. Runs in parallel with the shards;
+  # merge-gating exactly like shared-frontend-packages was before the split.
+  shared-frontend-checks:
+    name: Shared frontend checks
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        shard: [1, 2, 3]
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
@@ -452,7 +443,6 @@ jobs:
         run: pnpm install
 
       - name: Typecheck shared packages
-        if: matrix.shard == 1
         run: pnpm shared:typecheck
 
       - name: Build shared packages
@@ -462,26 +452,19 @@ jobs:
       # sit with the python guards in repo-shape (that job installs no node and
       # has no dist/theme.css to read). Both modes are checked in one pass.
       - uses: actions/setup-python@v7
-        if: matrix.shard == 1
         with:
           python-version: "3.12"
 
       - name: Check Mobile ProductClient export resolution
-        if: matrix.shard == 1
         run: python3 scripts/check_mobile_product_client_export.py --preflight
 
       - name: Check theme contrast floors
-        if: matrix.shard == 1
         run: |
           python3 -m unittest scripts/test_check_theme_contrast.py
           python3 scripts/check_theme_contrast.py --table
 
       - name: Test ProductClient domain
-        if: matrix.shard == 1
         run: pnpm --filter @proliferate/product-client exec vitest run src/domain
-
-      - name: Test product client
-        run: pnpm --filter @proliferate/product-client exec vitest run --shard=${{ matrix.shard }}/3
 
       # Focused Tier-1 gate for the workflow authoring surface: builder save
       # gating (title + unresolved-reference blocks), schema-version refusal,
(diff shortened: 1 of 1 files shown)


pablonyx (commented):
REVIEW VERDICT: LGTM-FOR-MERGE

Reviewed as a fresh reviewer of record, no prior review found on this PR (`gh api .../pulls/2120/reviews` returned `[]`). Spot-checked every quantitative claim in the PR body against live GitHub Actions logs rather than trusting the write-up. Everything reconciles exactly. Note up front: the PR body itself says "This PR is an experiment and will not be merged" — LGTM here means the methodology, numbers, and CI-config mechanics are sound and the coverage/rollup risks are correctly identified, not that this exact diff should be merged as-is without the follow-ups below.

## 1. `exec` form / flag-swallow bug

Confirmed. Final invocation (`.github/workflows/ci.yml:513` on `exp/frontend-shard`, sha `cb1126fe99a`):
```
run: pnpm --filter @proliferate/product-client exec vitest run --shard=${{ matrix.shard }}/4
```
This uses `pnpm ... exec vitest run` (bypasses the `package.json` `"test": "vitest run"` script entirely), matching the pre-existing sibling pattern at line ~469 (`Test ProductClient domain`) and ~484 (`Test workflow authoring surface`). Root cause of the original bug checks out: `product-client/package.json` `"test"` is literally `"vitest run"`, so `pnpm ... test -- --shard=N/3` becomes `pnpm run test -- --shard=N/3` → pnpm inserts its own `--` ahead of forwarded args → vitest sees a double `--` and treats `--shard=...` as a positional arg, not a flag. Verified this actually happened: run [32410983892](https://github.com/proliferate-ai/proliferate/actions/runs/32410983892) (first commit, buggy `test --` form) — job 96561087738 logs `Tests 8168 passed (8168)` and job 96561087967 logs `Test Files 1093 passed (1093)` / `Tests 8168 passed (8168)`, i.e. every "shard" ran the full suite, exactly as the PR claims. Using `exec` (bare binary invocation, no wrapper script) means there's no `--` to double up, so this specific failure mode cannot recur for this line. No regression guard exists to stop someone later "simplifying" it back to the `test --` form (e.g. a CI assertion that shard file counts are disjoint) — worth a one-line comment or a follow-up check if this pattern gets reused elsewhere, but not a blocker.

## 2. Test-count reconciliation (quoted from actual run logs, not the PR body)

Run [32413118164](https://github.com/proliferate-ai/proliferate/actions/runs/32413118164) (head sha `cb1126fe99a`), literal vitest summary lines:
- Shard 1 (job 96567934159): `Test Files 274 passed (274)` / `Tests 2035 passed (2035)` / `Duration 110.24s`
- Shard 2 (job 96567934211): `Test Files 274 passed (274)` / `Tests 1990 passed (1990)` / `Duration 90.72s`
- Shard 3 (job 96567934134): `Test Files 274 passed (274)` / `Tests 2242 passed (2242)` / `Duration 118.71s`
- Shard 4 (job 96567934152): `Test Files 271 passed (271)` / `Tests 1901 passed (1901)` / `Duration 108.86s`
- Shared frontend checks (job 96567934081): domain `Test Files 83 passed (83)` / `Tests 940 passed (940)`; workflow-authoring `Test Files 2 passed (2)` / `Tests 29 passed (29)`

Sum: 274+274+274+271 = **1093 files**; 2035+1990+2242+1901 = **8168 tests**. Cross-checked against a same-day baseline run on `main` (32335721001, job 96324803204, pre-PR unsharded job): `Test Files 1093 passed (1093)` / `Tests 8168 passed (8168)` / `Duration 336.90s`. Exact match, zero drift, no overlap and no gap between shards (vitest `--shard` partitions by file, consistent with the per-shard file counts being disjoint subsets of the same 1093). The intermediate 3-way numbers in the PR body also check out against run [32412026107](https://github.com/proliferate-ai/proliferate/actions/runs/32412026107): 365/2692, 365/2843, 363/2633 — all literal, all quoted correctly in the PR body.

## 3. Coverage completeness

Diffed the full old `shared-frontend-packages` job body (pre-PR, `origin/main:.github/workflows/ci.yml`) against the new `shared-frontend-checks` + `shared-frontend-packages` job bodies on `exp/frontend-shard`. Result: byte-for-byte identical except the single `Test product client` step is removed from the checks job and re-added, shard-parameterized, in the matrix job. Nothing else changed — no dropped `timeout-minutes`/`env`/`continue-on-error` (old job had none to lose), same checkout/setup-node/pnpm-action-setup pins, same install command.

Everything the old single job ran now runs exactly once across the two new jobs:
- install, typecheck (`pnpm shared:typecheck`), build, the two python-only checks (mobile export preflight, theme contrast) → `shared-frontend-checks` (ci.yml:428+, `exp/frontend-shard`)
- "Test product client" (sharded 4-way, partition not duplication) → `shared-frontend-packages` matrix (ci.yml:489+)
- install + build re-run per shard (necessary — `product-client` resolves `@proliferate/design`/`@proliferate/cloud-sdk(-react)` via `package.json` `exports` into `dist`, so each shard's process needs its own built `dist`)

Per your note: "Test ProductClient domain" (ci.yml ~line 469 on the PR branch, `vitest run src/domain`) is a pre-existing duplicate of a slice of the main suite in the OLD layout — those same domain test files are also collected inside the un-filtered "Test product client" run (visible in the baseline run's log: `src/domain/chats/transcript/transcript-work-duration.test.ts` appears inside the `Test product client` step's own file list). This PR does **not** fix or remove that duplication — it just relocates the already-duplicate step into `shared-frontend-checks`, where it still runs once per CI invocation, now in parallel with the sharded job instead of serially before it in the same job. Same for "Test workflow authoring surface." Net effect: pre-existing redundancy is preserved unchanged, not introduced and not fixed by this PR. Worth calling out explicitly since it's easy to misread the diff as "removed the duplicate" when it only moved it.

## 4. Job wall-clocks (requoted from `gh run view --json jobs`, run 32413118164)

| Job | started → completed | wall clock | PR body claim |
|---|---|---|---|
| Shared frontend checks | 20:15:37 → 20:17:50 | 133s (2m13s) | 133s ✓ |
| Shared frontend packages (1) | 20:15:37 → 20:18:44 | 187s (3m07s) | 187s ✓ |
| Shared frontend packages (2) | 20:15:37 → 20:18:21 | 164s (2m44s) | 164s ✓ |
| Shared frontend packages (3) | 20:15:37 → 20:18:43 | 186s (3m06s) | 186s ✓ |
| Shared frontend packages (4) | 20:15:37 → 20:18:32 | 175s (2m55s) | 175s ✓ |

All five match the PR body exactly. Baseline job (32335721001, job 96324803204) ran 05:28:26 → 05:36:00 = 454s (7m34s), consistent with the "~7m34s-9m13s" range claimed. New max-lane = 187s vs 180s target — the PR's "7s over" claim is accurate, not rounded favorably.

## 5. Rollup / required-checks implication

Live check names produced by this PR, confirmed via `gh pr view 2120 --json statusCheckRollup`: `Shared frontend checks`, `Shared frontend packages (1)`, `Shared frontend packages (2)`, `Shared frontend packages (3)`, `Shared frontend packages (4)` — replacing the single prior name `Shared frontend packages`. The PR's own "Caveat" section correctly flags that a required-checks/branch-protection update would be needed if this pattern is landed for real. I could not find an active ruleset or classic branch protection on `main` via the API to verify against directly (`GET .../branches/main/protection` → 404 "Branch not protected"; `GET .../rules/branches/main` → `[]`) — either required checks are enforced somewhere I don't have visibility into (org-level, external merge-queue config) or not enforced via GitHub's native mechanism at all. Either way: **whoever harvests this experiment into a real PR must manually verify and update wherever `Shared frontend packages` is currently referenced as a required check name**, since GitHub matches required-check names exactly and matrix jobs will not satisfy a check registered under the old singular name (this would silently block merges, not silently pass them — fail-closed, but still needs the manual step).

## Bottom line

All five verification points check out against primary sources (live run logs, not just the PR's self-reported numbers). No dropped test coverage, no double-counting introduced, the flag-swallow bug fix is correct and matches the established repo pattern, and the wall-clock numbers are accurate to the second. The PR is explicitly an experiment ("will not be merged," per the PR body) — the real action item is the required-checks caveat in point 5, which must be handled by whoever turns this into a real change, not by this PR.

pablonyx:
Reopening to refresh CI metadata check after label fix.

PR #2121, opened 2026-08-20
ci(playwright): cache browser binaries in browser CI jobs

pablonyx:
## Summary
- Cache `~/.cache/ms-playwright` in the three browser-driving CI jobs (`login-budget`, `scroll-physics`, `workflow-definition-lifecycle`), keyed on the `playwright-core` version resolved from `pnpm-lock.yaml` so an upgrade auto-busts the cache.
- `--with-deps` keeps running unconditionally on every run: the apt-installed OS shared libraries it also provides live outside `~/.cache/ms-playwright` and would not survive anyway on GitHub's ephemeral runner VM, so caching only removes the browser-binary download; Playwright's installer already no-ops that download when the target revision is present.

## Baseline (3 recent successful `main` runs, install-step wall time only)
| Job | Run 1 | Run 2 | Run 3 |
| --- | --- | --- | --- |
| login-budget (chromium) | 26s | 26s | 31s |
| workflow-definition-lifecycle (chromium) | 26s | 31s | 31s |
| scroll-physics (chromium+webkit) | 46s | 46s | 51s |
| **Total across jobs** | **98s** | **103s** | **113s** |

Well above the 30s "not worth it" floor, so proceeding with the cache.

## Test plan
- [ ] Run this workflow twice on this PR (prime + warm) and record before/after install-step timings in a follow-up comment.
- [ ] Confirm cache hit on the second run via the `actions/cache` step log ("Cache restored from key...").
- [ ] Confirm all three jobs still pass green on the warm run (no missing shared-lib failures from browsers launched against a cached-binary/fresh-OS-deps combination).

Generated with Claude Code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -409,6 +409,27 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
+      - name: Resolve Playwright version
+        id: playwright-version
+        run: |
+          version=$(grep -m1 '^  playwright-core@' pnpm-lock.yaml | sed -E 's/^  playwright-core@([0-9.]+):.*/\1/')
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
+      # EXPERIMENT (exp/playwright-cache): cache the downloaded browser
+      # binaries so a cache hit skips the network download; the key is the
+      # exact playwright-core version resolved above so a bump auto-busts it.
+      # --with-deps still runs unconditionally below because the OS-level apt
+      # packages it installs live outside this cache dir and would not
+      # survive on the ephemeral runner anyway; playwright's own installer
+      # already no-ops the binary download when the target revision is
+      # already present, so this only removes the download, not the apt step.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -582,6 +620,21 @@ jobs:
...
+        run: |
+          version=$(grep -m1 '^  playwright-core@' pnpm-lock.yaml | sed -E 's/^  playwright-core@([0-9.]+):.*/\1/')
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
+      # EXPERIMENT (exp/playwright-cache): see login-budget's cache step for
+      # rationale.
+      - name: Cache Playwright browsers
+        id: playwright-cache
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
```
pablonyx: Remove `experiment` comment

PR #2122, opened 2026-08-20
test(server): scope db truncation and shard the suite with pytest-xdist

pablonyx:
Scope the per-test DB truncate to tests that actually reach the database, and run the suite under `pytest-xdist -n 4` with a database per worker: the `test` job drops from 74.7m to ~14m with a pass set identical to main across three runs.

## Result

Combined truncate-scoping + `pytest-xdist -n 4`, three green `server-ci` runs, all with a pass set identical to main:

| run | `test` job wall | pytest summary |
| --- | --- | --- |
| [B](https://github.com/proliferate-ai/proliferate/actions/runs/32412638503/job/96566398648) | **13m33s** | `2972 passed, 1 skipped, 903 warnings in 765.91s (0:12:45)` |
| [C](https://github.com/proliferate-ai/proliferate/actions/runs/32412638503/job/96570576815) | **35m55s** | `2972 passed, 1 skipped, 904 warnings in 2080.78s (0:34:40)` |
| [D](https://github.com/proliferate-ai/proliferate/actions/runs/32412638503/job/96581108035) | **14m35s** | `2972 passed, 1 skipped, 904 warnings in 811.24s (0:13:31)` |

Baseline (main, run 32380718513): **74.7m**, `2972 passed, 1 skipped, 901 warnings in 4402.64s (1:13:22)`.

Typical case ~14m against 74.7m: **about 5x, roughly 60 minutes saved per run**. Even run C, the outlier, beat every serial `test` job observed in the last two days (47.2m-69.6m, median 57.8m).

Run C is runner hardware, not the change: *every* step in that job ran about twice as slow, including ones that never touch the suite (checkout 2.0s -> 5.1s, pnpm setup 3.5s -> 9.8s, dependency install 2.0s -> 4.6s, container init 11.0s -> 19.7s). The test step degraded slightly worse than that ratio (2.7x vs ~2.0x), which is what a shared Postgres bottleneck under four writers does on a slow host.

## What changed

1. **Truncate scoping.** The autouse `reset_test_database` fixture folded into `test_engine`, the single door every database path in the suite goes through. Audit of all three `conftest.py` files, plus a grep for direct engine creation, found no fixture-less database access: the only tests that build their own engines use `tests.postgres.temporary_database`, a throwaway uuid-named database that correctly opts out of the shared reset. A `--setup-plan` dump confirms 1,050 of 2,847 parseable tests resolve `test_engine`, and that zero tests reach `db_session`, `client`, or `cloud_client` without it. The other ~1,800 tests stop paying two 78-table `TRUNCATE`s each.

   Fixture ordering is preserved. `test_engine` is the first of its chain to be set up, so the pre-test truncate still lands before any dependent fixture can seed rows; teardown is the reverse of setup, so the post-test truncate still lands after those fixtures unwind.

2. **Per-worker databases.** Each xdist worker is its own OS process, so the session-scoped `migrated_test_database` fixture is session-scoped *per worker*. Deriving the database name from `PYTEST_XDIST_WORKER` gives each worker a database it alone migrates, truncates, and drops. An empty worker id (any run without `-n`) keeps the historical per-pid name byte-for-byte.

   The worker id is exported into the worker environment before pytest imports any test module, so reading it at `tests/postgres.py` import time is safe. Verified in an isolated harness, with a negative control that asserted the id was *not* visible and duly failed with `worker id WAS visible ('gw0')`.

   `CREATE DATABASE` and `DROP DATABASE` lock the template database, so concurrent worker startup can lose that race with SQLSTATE 55006 even though names never collide. Both are retried with backoff on 55006; `CREATE` treats 42P04/23505 as already-satisfied.

## Two things `-n 4` broke that had to be fixed

**Non-deterministic collection.** The first `-n 4` run aborted with "Different tests were collected between gw0 and gw3". `tests/unit/test_sentry_transport_privacy.py` built its parametrize table from `str(uuid4())` and `uuid4().hex` evaluated at import time, so the random value landed in the test id and each worker collected 44 differently-named tests. Serial runs never noticed because only one process ever collected. Both are now fixed literals; nothing in that matrix depends on the ids being fresh, only on their shape.

Verified by collecting the CI test set twice in independent processes: 2,973 node ids, byte-identical. Negative control on the unfixed file: 44 ids differ between the same two collections.

**Postgres lock table.** The second run died with 1,057 setup errors, all one cause:

```
asyncpg.exceptions.OutOfMemoryError: out of shared memory
HINT:  You might need to increase max_locks_per_transaction.
```

Postgres sizes its shared lock table at `max_locks_per_transaction * max_connections`, so the stock 64 * 100 leaves 6,400 slots. One 78-table `TRUNCATE ... CASCADE` takes an AccessExclusiveLock on every table, index, toast relation, and sequence it touches, and a full alembic upgrade in a single transaction takes one per object it creates. Serially that peaks at one such transaction; with four workers the truncates and the per-worker and per-temporary-database migrations overlap and exhaust the table.

The GUC is postmaster-level and Actions service containers accept no command line, so the workflow writes it via `ALTER SYSTEM` and restarts the container (which preserves port publishing). **Any real adoption of `-n` needs this, or an equivalent Postgres config change.**

## Redis

No per-worker partitioning needed. `proliferate.integrations.redis_lock` is the suite's only Redis client, and every test that reaches it monkeypatches `Redis.from_url`. All other `redis` references in tests are URL-rendering assertions against deploy manifests. The CI `redis` service container is never contacted by this test command.

## Other shared state

No module-, class-, or package-scoped fixtures exist in the suite (three session-scoped ones, all per-process). No test writes outside `tmp_path`. No test binds a fixed port on the CI path. The only cross-worker resource is the Postgres instance itself.

## Caveat

`-q` prints counts, not node ids, so the pass-set comparison is on `2972 passed, 1 skipped` matching main exactly, plus deterministic collection of the same 2,973 ids. Zero tests failed on any of the three runs, so no main-green test regressed.

### DIFF
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -166,7 +166,7 @@ jobs:
         run: uv pip install --system -e ".[dev]"
 
       - name: Run tests
-        run: pytest tests/unit tests/integration tests/e2e/cloud/test_e2b_webhooks.py -q --durations=20
+        run: pytest tests/unit tests/integration tests/e2e/cloud/test_e2b_webhooks.py -n 4 -q --durations=20
 
   docker:
     runs-on: ubuntu-latest
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -165,6 +165,30 @@ jobs:
       - name: Install dependencies
         run: uv pip install --system -e ".[dev]"
 
+      # Four workers means four concurrent 78-table TRUNCATEs and four
+      # concurrent full alembic upgrades, each holding an AccessExclusiveLock on
+      # every table, index, and sequence it touches. Postgres sizes its shared
+      # lock table at max_locks_per_transaction * max_connections, and the
+      # 64 * 100 default overflows with "out of shared memory". This GUC is
+      # postmaster-level, and service containers take no command line, so it has
+      # to be written to postgresql.auto.conf and the container bounced.
+      - name: Raise the Postgres lock table for parallel workers
+        run: |
+          set -euo pipefail
+          container="$(docker ps --filter 'ancestor=postgres:16-alpine' --format '{{.ID}}' | head -1)"
+          test -n "$container"
+          docker exec "$container" psql -U proliferate -d proliferate \
+            -c 'ALTER SYSTEM SET max_locks_per_transaction = 1024;'
+          docker restart "$container"
+          for _ in $(seq 1 60); do
+            if docker exec "$container" pg_isready -U proliferate -d proliferate >/dev/null 2>&1; then
+              break
+            fi
+            sleep 1
+          done
+          docker exec "$container" psql -U proliferate -d proliferate \
+            -c 'SHOW max_locks_per_transaction;'
+
       - name: Run tests
         run: pytest tests/unit tests/integration tests/e2e/cloud/test_e2b_webhooks.py -n 4 -q --durations=20
 
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -37,6 +37,7 @@ dependencies = [
 dev = [
     "pytest>=9.1.1",
     "pytest-asyncio>=1.4.0",
+    "pytest-xdist>=3.8.0",
     "httpx==0.28.1",
     "mypy==1.20.2",
     "ruff>=0.16.2",
--- a/server/tests/conftest.py
+++ b/server/tests/conftest.py
@@ -52,18 +52,33 @@ def migrated_test_database():  # type: ignore[no-untyped-def]
 
 @pytest_asyncio.fixture
 async def test_engine(migrated_test_database):  # type: ignore[no-untyped-def]
+    """The one door to the test database -- and therefore the reset boundary.
+
+    The truncate used to live in an autouse ``reset_test_database`` fixture, so
+    all 2,900+ tests paid a 78-table TRUNCATE (twice) plus the session-scoped
+    alembic migration, including the large majority that never touch Postgres.
+    Hanging it off ``test_engine`` scopes the reset to exactly the tests that
+    reach the database, because every database path in the suite -- ``client``,
+    ``db_session``, ``cloud_client``, and every per-module session factory --
+    resolves through this fixture.
+
+    Fixture ordering is unchanged: the pre-test truncate still runs before any
+    fixture that depends on ``test_engine`` can seed rows, and the post-test
+    truncate still runs after those fixtures tear down, because teardown is the
+    reverse of setup and this fixture is the first of the chain to be set up.
+
+    Tests that build their own throwaway database (``temporary_database``) are
+    self-isolating and correctly opt out of the shared-database reset.
+    """
     engine = create_async_engine(TEST_DATABASE_URL, echo=False)
-    yield engine
-    await engine.dispose()
-
-
-@pytest_asyncio.fixture(autouse=True)
-async def reset_test_database(test_engine) -> AsyncGenerator[None, None]:  # type: ignore[no-untyped-def]
     await _cancel_test_background_tasks()
-    await truncate_all_tables(test_engine)
-    yield
-    await _cancel_test_background_tasks()
-    await truncate_all_tables(test_engine)
+    await truncate_all_tables(engine)
+    try:
+        yield engine
+    finally:
+        await _cancel_test_background_tasks()
+        await truncate_all_tables(engine)
+        await engine.dispose()
 
 
 @pytest_asyncio.fixture
--- a/server/tests/postgres.py
+++ b/server/tests/postgres.py
@@ -5,6 +5,7 @@ from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
 
 from sqlalchemy import text
+from sqlalchemy.exc import DBAPIError
 from sqlalchemy.ext.asyncio import AsyncEngine, create_async_engine
 
 from alembic import command
@@ -23,10 +24,36 @@ POSTGRES_USER = "proliferate"
 POSTGRES_PASSWORD = "localdev"
 POSTGRES_HOST = "127.0.0.1"
 POSTGRES_PORT = 5432
-TEST_DATABASE_NAME = os.environ.get(
-    "PROLIFERATE_TEST_DATABASE_NAME",
-    f"proliferate_test_{os.getpid()}",
-)
+
+# pytest-xdist runs every worker as its own OS process and exports the worker id
+# ("gw0", "gw1", ...) into that process's environment before pytest imports any
+# test module, so reading it at import time is safe. Naming the database after
+# the worker is what makes the rest of the harness per-worker for free: the
+# session-scoped ``migrated_test_database`` fixture is session-scoped *per
+# process*, so each worker migrates and later drops only the database it owns,
+# and the per-test TRUNCATE can never reach another worker's rows.
+#
+# A serial run (no ``-n``) sees an empty worker id and keeps the historical
+# per-pid name byte-for-byte.
+XDIST_WORKER = os.environ.get("PYTEST_XDIST_WORKER", "")
+
+
+def _default_test_database_name() -> str:
+    if XDIST_WORKER:
+        return f"proliferate_test_{XDIST_WORKER}_{os.getpid()}"
+    return f"proliferate_test_{os.getpid()}"
+
+
+def _resolve_test_database_name() -> str:
+    explicit = os.environ.get("PROLIFERATE_TEST_DATABASE_NAME")
+    if explicit is None:
+        return _default_test_database_name()
+    # An explicit override names one database. Under xdist it still has to fan
+    # out, or the workers would truncate each other's rows mid-test.
+    return f"{explicit}_{XDIST_WORKER}" if XDIST_WORKER else explicit
+
+
+TEST_DATABASE_NAME = _resolve_test_database_name()
 ADMIN_DATABASE_URL = (
     f"postgresql+asyncpg://{POSTGRES_USER}:{POSTGRES_PASSWORD}@"
     f"{POSTGRES_HOST}:{POSTGRES_PORT}/postgres"
@@ -35,6 +62,15 @@ TEST_DATABASE_URL = (
     f"postgresql+asyncpg://{POSTGRES_USER}:{POSTGRES_PASSWORD}@"
     f"{POSTGRES_HOST}:{POSTGRES_PORT}/{TEST_DATABASE_NAME}"
 )
+
+# CREATE DATABASE / DROP DATABASE take a lock on the template database, so
+# concurrent creates from different xdist workers (or from ``temporary_database``
+# inside a test) can lose that race with 55006 "is being accessed by other
+# users" even though the database *names* never collide. 42P04/23505 mean
+# someone else already created the name, which is the outcome we wanted anyway.
+_DATABASE_EXISTS_SQLSTATES = frozenset({"42P04", "23505"})
+_DATABASE_BUSY_SQLSTATES = frozenset({"55006"})
+_DATABASE_LIFECYCLE_ATTEMPTS = 12
 
 
 def make_database_url(database_name: str) -> str:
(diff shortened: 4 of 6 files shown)


pablonyx (commented):

Fresh reviewer of record; I did not implement this. Two blockers, both mechanical. The engineering substance — the truncate scoping, the per-worker DB derivation, the collection fix, the lock-table bump — held up under adversarial inspection and I could not find a hole in it. Details below.

---

## Blockers

**B1 — `Repo shape checks` is red at head, and this PR caused it.**
`server/tests/unit/test_sentry_transport_privacy.py` goes 596 → 602 lines on this branch. PROD-SIZE-1 caps scanned source at 600 and there is no `[[max_lines]]` entry for the file in any `lints/*/ratchets.toml`. The 8-line comment block added at `test_sentry_transport_privacy.py:181-187` (commit `0f1ed658b`) is what crosses the line.

Literal failure — [CI run 32412638510, job 96566399030](https://github.com/proliferate-ai/proliferate/actions/runs/32412638510/job/96566399030):
```
Files above their max line threshold with no ratchet entry, or grown past the count their ratchet entry records:
server/tests/unit/test_sentry_transport_privacy.py: PROD-SIZE-1 — Source files stay under the repo-wide line cap
  found: 602 lines (max 600)
```
Fix by trimming the comment to ~2 lines (the mechanism is already restated in the PR body), not by adding a ratchet exception — a net-new exception needs founder approval. `mergeStateStatus` is `UNSTABLE` solely because of this.

**B2 — the PR declares itself unmergeable.** It is a draft, and the body opens `Measurement branch. **Do not merge.**` Merge-readiness cannot be granted against that text. If the intent has changed to "land it", strip that line, mark ready, and re-request. If it has not, this should be split: the substance is worth landing.

---

## What I verified (adversarially)

**1. Truncate scoping — no hole found.** `server/tests/conftest.py:54-82` folds the pre/post `truncate_all_tables` into `test_engine`. I hunted for shared-DB access that bypasses that fixture graph three ways:

- Grep across the whole tests tree: `TEST_DATABASE_URL` is imported in exactly three places outside `conftest.py` — `tests/postgres.py` itself, `tests/e2e/cloud/helpers/external.py:23` (default arg for a subprocess uvicorn, only reachable from `cloud_client` tests), and `tests/integration/background/test_relay_task.py:26` (a `monkeypatch.setattr(settings, "database_url", …)` inside two tests that both take `db_session`). No fixture anywhere builds its own engine against the shared test DB.
- Every `async_sessionmaker(...)` in the tests tree takes `test_engine` as its bind. Every other engine construction targets a `temporary_database(...)` uuid-named throwaway, which correctly opts out.
- AST sweep of all 316 collected files (`tests/unit`, `tests/integration`, `test_e2b_webhooks.py`): 37 test functions reference a DB token without resolving `test_engine`/`db_session`/`client`/`cloud_client`. I triaged all 37 — 22 are `temporary_database` migration tests (self-isolating), 6 are `test_server_boundary_checker*` string fixtures written to `tmp_path`, and the remaining 9 (`test_billing_start_block_paging.py:112/154/199`, `test_anonymous_telemetry_sender.py:55/111`, `test_agent_gateway_key_transaction.py:21`, `test_cloud_workspace_transaction.py:22`, `test_cloud_workspace_materialization_access.py:337`, `test_cloud_sandbox_reaper_task.py:36`) all monkeypatch the factory or engine with a stub before touching it. Zero real fixture-less DB access.

Two structural facts back this up, and they matter more than the audit:
- The unpatched `proliferate.db.engine.engine` binds `settings.database_url`, which in CI is the `proliferate` service DB (`server-ci.yml:116`). Nothing ever migrates that database — only `TEST_DATABASE_URL` gets `run_migrations`. Any leak through the global engine therefore fails loudly with an undefined-table error rather than silently polluting. The design is fail-closed.
- The one ordering hazard the docstring's "fixture ordering is unchanged" claim does *not* cover — a sibling fixture that seeds the DB without depending on `test_engine`, which under the old autouse ordering ran after the truncate and now would run before it — cannot exist, because there is no such seeding route (see above).

Independent confirmation: the sibling branch `exp/server-truncate-scope` (`24d1d2310`) carries the scoping change **serially, without xdist** and produced [`2972 passed, 1 skipped, 901 warnings in 3797.36s (1:03:17)`](https://github.com/proliferate-ai/proliferate/actions/runs/32411347303). That isolates the scoping from the sharding and shows the scoping alone is pass-set-neutral.

**2. Per-worker DB isolation — correct.** `tests/postgres.py:38-56`. `PYTEST_XDIST_WORKER` is read at import time; xdist exports it into each worker process before test-module import, so this is sound, and the controller process (which never runs fixtures) computing a stray name is harmless. `migrated_test_database` (`conftest.py:46-50`) is session-scoped *per process*, so each worker migrates and drops only its own DB. `TEST_DATABASE_URL` is derived from `TEST_DATABASE_NAME`, so every consumer — including the subprocess server in `external.py` — follows automatically. Serial runs (`XDIST_WORKER == ""`) keep the historical per-pid name byte-for-byte. Nothing points at the shared `proliferate` DB under xdist. The 55006 retry loops (`postgres.py:99-121`, `127-153`) are correct: every path returns or raises, and the last attempt always raises rather than falling through silently.

**3. uuid4 collection fix — deterministic, unique, and the only site.** `test_sentry_transport_privacy.py:188-189`. `VALID_UUID` is a structurally valid dashed v4 UUID; `VALID_UUID_HEX` is 32 hex chars — I counted. No two param tuples collapse to the same id. I scanned every `parametrize` block in the tests tree for `uuid4`/`uuid1`/`random.`/`time.time()`/`datetime.now`/`token_hex`/`os.urandom` and for order-unstable sources (`glob`, `iterdir`, `listdir`, set iteration, `__subclasses__`): zero remaining hits. The one glob-driven parametrize source, `test_sentry_integration.py:46`, is already `sorted(...)`.

The failure this fixes is real and was reproduced in CI: [job 96561889539](https://github.com/proliferate-ai/proliferate/actions/runs/32411236687/job/96561889539) — `ERROR gw3 - Different tests were collected between gw0 and gw3`.

**4. `max_locks_per_transaction` — takes effect before pytest, test-only.** `server-ci.yml:175-190`. Verified in the run log for job 96581108035: the step emits `ALTER SYSTEM`, restarts the container, waits for `pg_isready`, and prints
```
 max_locks_per_transaction
---------------------------
 1024
(1 row)
```
at 21:02:38Z, with `Run pytest ... -n 4` starting at 21:02:58Z. `docker restart` preserves the `-p 5432:5432` publish, and postgres re-reads `postgresql.auto.conf` from the container filesystem. Confirmed nothing outside `.github/workflows/server-ci.yml` changed — no `server/infra` terraform, no compose file, no runtime config. The CI-vs-prod divergence is test-harness-only and correctly scoped. The intermediate failure it fixes is also real: [job 96563396985](https://github.com/proliferate-ai/proliferate/actions/runs/32411707662/job/96563396985) — `asyncpg.exceptions.OutOfMemoryError: out of shared memory`.

**5. Three-run identity claim — confirmed, I pulled the logs myself.** All three are attempts 1/2/3 of run `32412638503` at the same head SHA `5b780b86d`:

| attempt | job | wall | literal summary line |
| --- | --- | --- | --- |
| 1 | [96566398648](https://github.com/proliferate-ai/proliferate/actions/runs/32412638503/job/96566398648) | 13m33s | `2972 passed, 1 skipped, 903 warnings in 765.91s (0:12:45)` |
| 2 | [96570576815](https://github.com/proliferate-ai/proliferate/actions/runs/32412638503/job/96570576815) | 35m55s | `2972 passed, 1 skipped, 904 warnings in 2080.78s (0:34:40)` |
| 3 | [96581108035](https://github.com/proliferate-ai/proliferate/actions/runs/32412638503/job/96581108035) | 14m35s | `2972 passed, 1 skipped, 904 warnings in 811.24s (0:13:31)` |

Baseline claim also confirmed. [Run 32380718513, job 96462969785](https://github.com/proliferate-ai/proliferate/actions/runs/32380718513) — an unmodified-conftest, serial, main-based run — gives `2972 passed, 1 skipped, 901 warnings in 4402.64s (1:13:22)`, 74.7m job wall. A second independent serial main-based run ([32380133163](https://github.com/proliferate-ai/proliferate/actions/runs/32380133163)) gives `2972 passed, 1 skipped, 901 warnings in 4075.11s (1:07:55)`. Pass/skip counts match the three xdist runs exactly. The warning delta (901 → 903/904) is expected sharding noise, not a pass-set change.

The body's own caveat is the honest one and I'll restate it: `-q` prints counts, not node ids, so this is count-identity plus deterministic collection of the same 2,973 ids, not a node-id diff against main.

**6. Dependency wiring — correct, with one note.** `pytest-xdist>=3.8.0` in `server/pyproject.toml:40`; `uv.lock` picks up `pytest-xdist 3.8.0` and `execnet 2.1.2` with hashes. Note that the `test` job installs via `uv pip install --system -e ".[dev]"` (`server-ci.yml:166`), which resolves fresh and **ignores `uv.lock`**, so the floor `>=3.8.0` is what actually governs the test job — a future xdist major could land in CI unreviewed. Only the `lint` job is `--frozen`. Pre-existing pattern, flagging for awareness, not asking for a change here.

---

## Non-blocking follow-ups

- **`-n 4` is hardcoded** (`server-ci.yml:193`). Defensible — the lock-table sizing was tuned for exactly four writers — but the two numbers are coupled silently. Either add a comment tying `-n 4` to `max_locks_per_transaction = 1024`, or derive both from one workflow-level `env` var. Please do *not* switch to `-n auto`/`-n logical` without re-deriving the lock budget.
- **Consider `--dist loadfile`.** The default `--dist load` splits a module's tests across workers. Today that is safe, because no same-module ordering dependency can exist (the old autouse truncate made one impossible). But the scoping change now permits one to be written, and under `load` it would surface as a luck-dependent flake rather than a deterministic failure. `loadfile` also improves DB/connection locality.
- **Assert, don't just `SHOW`.** `server-ci.yml:188-190` prints the GUC but never checks it. Piping through `grep -q 1024` converts a future silent regression (which reappears as 1,057 confusing setup errors) into a one-line step failure.
- **Image-tag coupling.** `docker ps --filter 'ancestor=postgres:16-alpine'` at `:178` duplicates the service image literal at `:121`. If they drift, `test -n "$container"` fails loudly, which is acceptable — but a shared `env` var would remove the trap.
- **Orphaned worker DBs.** A SIGKILLed worker skips `migrated_test_database` teardown and leaves `proliferate_test_gwN_<pid>` behind. Irrelevant in an ephemeral CI container; on a developer box these accumulate. A `proliferate_test_%` sweep at session start would close it.
- **`_resolve_test_database_name` override path** (`postgres.py:53`) appends `_gwN` to an operator-supplied name with no length check. A name near Postgres's 63-char identifier limit silently overflows and asyncpg rejects it. Same 63-char trap the suite already documents at `test_agent_model_snapshot_composed_rekey_migration.py:88`.

---

## Bottom line

The four changes are individually correct and jointly deliver a real ~5x. I could not find a flake vector in the truncate scoping, and the serial `exp/server-truncate-scope` run plus the fail-closed unmigrated-`proliferate` property are the two pieces of evidence that convinced me. Clear B1 (six lines of comment) and resolve B2 (draft + "do not merge"), and I would flip this to LGTM without re-litigating the substance.

pablonyx:
B1 from review 4988186612 is fixed: the 6-line comment at `test_sentry_transport_privacy.py:181-186` is condensed to 2 lines, file now 598 lines (cap 600). B2 (draft / do-not-merge body) is intentionally left for Pablo's adoption decision.

pablonyx (commented):
REVIEW VERDICT (updated): LGTM-FOR-MERGE — pending Pablo's adoption call on the "do not merge" framing

Re-verified at head `b9811a65d0f7b220b56ac439b5f03ac96f02486d`.

**B1 is closed, and the fix does only what it claims.** `b9811a65d` touches exactly one file with a 2-insertion / 6-deletion comment rewrite:

```
 server/tests/unit/test_sentry_transport_privacy.py | 8 ++------
 1 file changed, 2 insertions(+), 6 deletions(-)
```

I diffed both revisions of that file with all comment lines stripped — byte-identical. No test, no parametrize value, no literal moved. `VALID_UUID` and `VALID_UUID_HEX` are untouched, so the determinism fix from finding 3 stands unchanged. File is now 598 lines, under the 600 cap, with no ratchet exception needed.

**Repo shape checks is green at head**: [pass, 1m35s](https://github.com/proliferate-ai/proliferate/actions/runs/32426576958/job/96609645865). Full check tally on `b9811a65d` is 26 pass / 2 skipping / 0 fail, and `mergeStateStatus` has moved `UNSTABLE` → `CLEAN`.

**Fourth timing sample**, from the whole-suite run on the new head — [run 32426576882, job 96609645722](https://github.com/proliferate-ai/proliferate/actions/runs/32426576882/job/96609645722), 13m19s job wall:

```
2972 passed, 1 skipped, 903 warnings in 746.47s (0:12:26)
```

Same `2972 passed, 1 skipped` as all three prior runs and as both serial main baselines. Same command (`-n 4`), and the lock-table step again reports `1024` before pytest starts:

```
ALTER SYSTEM
 max_locks_per_transaction
---------------------------
 1024
(1 row)
```

Four-sample spread is now 12m26s / 12m45s / 13m31s / 34m40s of pytest wall against a 68-73m serial baseline. The single outlier remains the slow-runner attempt, and `timeout-minutes: 75` still covers it.

**B2 is not a code finding.** The draft state and the `Measurement branch. **Do not merge.**` line are an adoption decision, not a defect — that call is Pablo's. Nothing in the diff blocks merge on technical grounds. When he decides to land it, strip that line and mark ready; no further review of the substance is needed from me.

My six non-blocking follow-ups from the first pass are unchanged and none of them gate merge. The two I would most like to see picked up whenever this is adopted are asserting the GUC value rather than only printing it (`server-ci.yml:188-190`), and commenting the coupling between `-n 4` and `max_locks_per_transaction = 1024`.

PR #2126, opened 2026-08-20
fix(ci): simplify deploy graph, add litellm prod deploy

pablonyx:
## Why

Founder-approved direction: e2b and mobile are no longer shipped, and the
workers deploy has never deployed anything in any environment. LiteLLM is
missing a nightly prod deploy. Intel Mac desktop builds are paused
temporarily. This PR simplifies the deploy graph accordingly.

**Workers-stub evidence** (why deleting `deploy-workers`/`deploy-workers-prod`
from the automated trains is safe): `_deploy-workers.yml`'s only step exits
`0` with a note unless `WORKERS_DEPLOY_ENABLED` is set to `"true"`, which no
GitHub Environment sets today — and even when set, the step deliberately
fails with `"Define the ECS worker service and worker command before
enabling this lane."` It has never deployed anything.

### `.github/workflows/nightly-release-train.yml`
- Deleted jobs: `deploy-e2b`, `deploy-e2b-prod`, `deploy-mobile`,
  `deploy-mobile-prod`, `deploy-workers`, `deploy-workers-prod`.
- Removed edges:
  - `deploy-server`: `needs: [prepare, deploy-e2b]` → `needs: [prepare]`
    (dropped the `always()` guard, no longer needed with a single dependency).
  - `deploy-server-prod`: `needs: [prepare, deploy-server, deploy-e2b-prod]`
    → `needs: [prepare, deploy-server]`.
  - `deploy-web` / `deploy-web-prod`: unchanged — still gated on
    `deploy-server` / `deploy-server-prod` respectively.
- Added `deploy-litellm` (environment `staging`, `needs: [prepare]`, parallel
  to `deploy-server`) and `deploy-litellm-prod` (environment `Production`,
  `needs: [prepare, deploy-litellm]`, gated on `deploy-litellm.result ==
  'success'`), both calling `_deploy-litellm.yml` with the same
  enabled/dry_run condition style already used by every other job in this
  train — patterned directly on `deploy-staging.yml:229-238` and
  `promote-production.yml:182-190`, which already wire this reusable
  workflow.
- Added a `litellm` output to `prepare` (reading
  `steps.detect.outputs.litellm`, which `detect-deploy-surfaces.mjs` already
  emits) and removed the now-dead `e2b`/`mobile` outputs.
- Rewired `publish-product-release`'s `needs:` list and its result-gating
  `if:` (dropped the `e2b`/`workers`/`mobile` legs, added a `litellm` leg
  matching the `server` leg's pattern) and `summary`'s `needs:` list and
  step body (dropped E2B/Workers/Mobile staging+prod lines, added
  LiteLLM staging+prod lines).
- **Left untouched:** `scripts/ci-cd/detect-deploy-surfaces.mjs` still lists
  `e2b` and `mobile` as valid surfaces (used elsewhere / cheap to keep) — I
  only removed the train's *consumption* of those two outputs, not the
  shared detection script itself.

### `.github/workflows/deploy-staging.yml`
- Deleted jobs: `deploy-e2b`, `deploy-mobile`, `deploy-workers`.
- `deploy-server`: `needs: [plan, deploy-e2b]` → `needs: [plan]`.
- `summary`: `needs:` list and step body dropped the E2B/Workers/Mobile
  legs.
- **Left untouched:** the `plan` job still outputs `e2b`/`mobile`/`workers`
  (and the "Summarize plan" table still lists them) — purely informational
  now, nothing in this workflow consumes them, and removing them wasn't in
  scope for this pass.

### `.github/workflows/release-desktop.yml` — Intel Mac pause
- Removed the `macos-15-intel` / `x86_64-apple-darwin` entry from the
  `build-desktop` matrix (commented out, not physically deleted, with:
  `# Intel builds paused 2026-08-20; restore by re-adding this matrix
  entry.`). Only `macos-15` / `aarch64-apple-darwin` builds now.
- **Consequence: existing Intel-Mac Proliferate installs will receive no
  further auto-updates while this is paused.** `publish-updater` still runs
  on every train and republishes `latest.json`/`installers.json` for
  `darwin-aarch64` only; there is no `darwin-x86_64` entry for the updater
  to serve to Intel clients.
- Audited `create-release` (~634) and `publish-updater` (~700) for hard
  x86_64/darwin-x86_64 references and fixed the ones that would otherwise
  fail every run once no Intel artifacts exist:
  - `scripts/generate-updater-manifest.mjs`: the `darwin-x86_64` platform
    entry is now `optional: true` — a missing artifact/signature logs a
    warning and is skipped instead of failing the whole manifest generation.
    `darwin-aarch64` is still mandatory.
  - `scripts/generate-desktop-installer-manifest.mjs`: same treatment for
    the `darwin-x86_64` download entry.
  - The "Bump Homebrew cask" step previously hard-errored
    (`::error::Missing x64 DMG for cask bump`) when no Intel DMG existed.
    The Homebrew cask format is a single universal formula requiring both
    architectures (`generate-homebrew-cask.mjs` takes `--sha-aarch64` and
    `--sha-x64` unconditionally), so there's no safe *partial* bump while
    x86_64 doesn't exist — the step now warns and skips (`exit 0`) rather
    than erroring. **Flagging, not guessing:** if the Intel pause becomes
    permanent, the Homebrew cask itself will need a real redesign
    (arm64-only formula) — that's a product decision outside this PR's
    scope, and the cask will simply stop updating for now.
  - `create-release` was already artifact-count-agnostic (`files:
    all-artifacts/**/*`, no per-arch enumeration) — no change needed there.
  - Verified both manifest generators against an ARM64-only artifact
    fixture with `node <script> ...` directly: both now succeed and emit a
    warning for the missing platform instead of exiting 1. Ran the existing
    `scripts/ci-cd/generate-updater-manifest.test.mjs` suite (5/5 pass,
    unaffected since it always supplies both platforms) plus
    `release-title-propagation.test.mjs`, `deploy-staging-trigger.test.mjs`,
    `detect-deploy-surfaces.test.mjs`, `create-release-tags.test.mjs`, and
    `validate-pr-metadata.test.mjs` (all green) as a sanity pass over the
    touched surface.
- `release-runtime.yml`'s separate `x86_64-apple-darwin` runtime-binary
  target was left untouched — that's the anyharness CLI/runtime release
  (self-host), not the Tauri desktop app matrix this PR pauses.

### Left on disk, unreferenced
`_deploy-e2b.yml`, `_deploy-mobile.yml`, `_deploy-workers.yml` are untouched
and no longer called from `nightly-release-train.yml` or
`deploy-staging.yml` — cheap to revert if any of these come back.

## Expected critical-path impact

Removing the `deploy-e2b` → `deploy-server` → `deploy-e2b-prod` →
`deploy-server-prod` predecessor chain takes E2B's build-and-promote step
out of the prod deploy chain's critical path entirely (LiteLLM and Web run
in parallel with Server, not serially behind it). Expected prod deploy
chain: **~30.5m → ~16.8m**.

## Known scope gap (please read before merge)

`grep -rn "deploy-e2b\|deploy-mobile\|deploy-workers\|macos-15-intel"
.github/workflows/` (see raw output below) is **not** fully clean:
`promote-production.yml` and `hotfix-production.yml` still define
`deploy-workers`/`deploy-mobile` jobs and `uses: ./.github/workflows/_deploy-{e2b,mobile,workers}.yml`
lines. Neither file was named in this task's file list (only
`nightly-release-train.yml`, `deploy-staging.yml`, and `release-desktop.yml`
were in scope), so I left them untouched rather than silently expanding
scope to two more human-triggered workflows with their own approval
semantics. If the intent is for the deploy graph simplification to cover
manual promotion/hotfix paths too, that's follow-up work I did not do here.

```
$ grep -rn "deploy-e2b\|deploy-mobile\|deploy-workers\|macos-15-intel" .github/workflows/
.github/workflows/release-desktop.yml:106:          # - os: macos-15-intel
.github/workflows/hotfix-production.yml:246:    uses: ./.github/workflows/_deploy-e2b.yml
.github/workflows/hotfix-production.yml:265:  deploy-workers:
.github/workflows/hotfix-production.yml:268:    uses: ./.github/workflows/_deploy-workers.yml
.github/workflows/hotfix-production.yml:285:  deploy-mobile:
.github/workflows/hotfix-production.yml:288:    uses: ./.github/workflows/_deploy-mobile.yml
.github/workflows/hotfix-production.yml:318:      - deploy-workers
.github/workflows/hotfix-production.yml:320:      - deploy-mobile
.github/workflows/hotfix-production.yml:334:            (needs.prepare.outputs.workers != 'true' || needs.deploy-workers.result == 'success') &&
.github/workflows/hotfix-production.yml:336:            (needs.prepare.outputs.mobile != 'true' || needs.deploy-mobile.result == 'success')
.github/workflows/hotfix-production.yml:392:      - deploy-workers
.github/workflows/hotfix-production.yml:394:      - deploy-mobile
.github/workflows/hotfix-production.yml:411:            echo "- Workers production: \`${{ needs.deploy-workers.result }}\`"
.github/workflows/hotfix-production.yml:413:            echo "- Mobile production: \`${{ needs.deploy-mobile.result }}\`"
.github/workflows/promote-production.yml:153:    uses: ./.github/workflows/_deploy-e2b.yml
.github/workflows/promote-production.yml:172:  deploy-workers:
.github/workflows/promote-production.yml:175:    uses: ./.github/workflows/_deploy-workers.yml
.github/workflows/promote-production.yml:202:  deploy-mobile:
.github/workflows/promote-production.yml:205:    uses: ./.github/workflows/_deploy-mobile.yml
.github/workflows/promote-production.yml:233:      - deploy-workers
.github/workflows/promote-production.yml:236:      - deploy-mobile
.github/workflows/promote-production.yml:253:            echo "- Workers: \`${{ needs.deploy-workers.result }}\`"
.github/workflows/promote-production.yml:256:            echo "- Mobile: \`${{ needs.deploy-mobile.result }}\`"
```

The `.github/workflows/nightly-release-train.yml` and
`.github/workflows/deploy-staging.yml` lines are gone entirely — those two
are the only files this task named for edits, and they're clean.

## Review note

The prod deploy graph and the desktop updater are both live, customer-facing
surfaces; merge is the founder's call, with the D1/D2 items below open.

## Known behavior while Intel is paused

Per review 4988181298's D1 finding: with `darwin-x86_64` dropped from
`latest.json`, existing Intel (x86_64) desktop installs are expected to see
recurring `UPDATER_CHECK_FAILED` errors on every scheduled update check,
rather than a silent "no update available." Separately, once
`minDesktopVersionEnforced` is turned on, Intel installs frozen at today's
build could be hard-blocked with no upgrade path as the server's minimum
version floor advances. Neither behavior has an owner decision yet; both are
open pending Pablo's call before merge.

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -194,36 +194,14 @@ jobs:
             echo "| runtime | \`${{ steps.detect.outputs.runtime }}\` |"
           } >> "$GITHUB_STEP_SUMMARY"
 
-  deploy-e2b:
-    needs: [plan]
-    if: ${{ always() && needs.plan.result == 'success' && needs.plan.outputs.dry_run != 'true' }}
-    uses: ./.github/workflows/_deploy-e2b.yml
-    with:
-      environment: staging
-      git_sha: ${{ needs.plan.outputs.head_sha }}
-      enabled: ${{ needs.plan.outputs.e2b == 'true' }}
-      rolling_tag: staging
-      mode: build_and_promote
-    secrets: inherit
-
   deploy-server:
-    needs: [plan, deploy-e2b]
+    needs: [plan]
     if: ${{ always() && needs.plan.result == 'success' && needs.plan.outputs.dry_run != 'true' }}
     uses: ./.github/workflows/_deploy-server.yml
     with:
       environment: staging
       git_sha: ${{ needs.plan.outputs.head_sha }}
       enabled: ${{ needs.plan.outputs.server == 'true' }}
-    secrets: inherit
-
-  deploy-workers:
-    needs: [plan, deploy-server]
-    if: ${{ always() && needs.plan.result == 'success' && needs.plan.outputs.dry_run != 'true' }}
-    uses: ./.github/workflows/_deploy-workers.yml
-    with:
-      environment: staging
-      git_sha: ${{ needs.plan.outputs.head_sha }}
-      enabled: ${{ needs.plan.outputs.workers == 'true' }}
     secrets: inherit
 
   deploy-litellm:
@@ -246,16 +224,6 @@ jobs:
       enabled: ${{ needs.plan.outputs.web == 'true' }}
     secrets: inherit
 
-  deploy-mobile:
-    needs: [plan]
-    if: ${{ always() && needs.plan.result == 'success' && needs.plan.outputs.dry_run != 'true' }}
-    uses: ./.github/workflows/_deploy-mobile.yml
-    with:
-      environment: staging
-      git_sha: ${{ needs.plan.outputs.head_sha }}
-      enabled: ${{ needs.plan.outputs.mobile == 'true' }}
-    secrets: inherit
-
   deploy-desktop:
     needs: [plan]
     if: ${{ always() && needs.plan.result == 'success' && needs.plan.outputs.dry_run != 'true' }}
@@ -274,12 +242,9 @@ jobs:
   summary:
     needs:
       - plan
-      - deploy-e2b
       - deploy-server
-      - deploy-workers
       - deploy-litellm
       - deploy-web
-      - deploy-mobile
       - deploy-desktop
     if: ${{ always() && needs.plan.result == 'success' }}
     runs-on: ubuntu-latest
@@ -294,12 +259,9 @@ jobs:
           {
             echo "### Staging deploy result"
             echo "- Head: \`${{ needs.plan.outputs.head_sha }}\`"
-            echo "- E2B: \`${{ needs.deploy-e2b.result }}\`"
             echo "- Server: \`${{ needs.deploy-server.result }}\`"
-            echo "- Workers: \`${{ needs.deploy-workers.result }}\`"
             echo "- LiteLLM: \`${{ needs.deploy-litellm.result }}\`"
             echo "- Web: \`${{ needs.deploy-web.result }}\`"
-            echo "- Mobile: \`${{ needs.deploy-mobile.result }}\`"
             echo "- Desktop: \`${{ needs.deploy-desktop.result }}\`"
           } >> "$GITHUB_STEP_SUMMARY"
 
--- a/.github/workflows/nightly-release-train.yml
+++ b/.github/workflows/nightly-release-train.yml
@@ -59,9 +59,8 @@ jobs:
       selected_surfaces: ${{ steps.detect.outputs.selected_surfaces }}
       server: ${{ steps.detect.outputs.server }}
       workers: ${{ steps.detect.outputs.workers }}
-      e2b: ${{ steps.detect.outputs.e2b }}
+      litellm: ${{ steps.detect.outputs.litellm }}
       web: ${{ steps.detect.outputs.web }}
-      mobile: ${{ steps.detect.outputs.mobile }}
       desktop: ${{ steps.detect.outputs.desktop }}
       runtime: ${{ steps.detect.outputs.runtime }}
       desktop_version: ${{ steps.artifacts.outputs.desktop_version }}
@@ -252,21 +251,9 @@ jobs:
       target_os: all
     secrets: inherit
 
-  deploy-e2b:
+  deploy-server:
     needs: [prepare]
     if: ${{ needs.prepare.result == 'success' && needs.prepare.outputs.dry_run != 'true' && needs.prepare.outputs.selected_surfaces != '' }}
-    uses: ./.github/workflows/_deploy-e2b.yml
-    with:
-      environment: staging
-      git_sha: ${{ needs.prepare.outputs.head_sha }}
-      enabled: ${{ needs.prepare.outputs.e2b == 'true' }}
-      rolling_tag: staging
-      mode: build_and_promote
-    secrets: inherit
-
-  deploy-server:
-    needs: [prepare, deploy-e2b]
-    if: ${{ always() && needs.prepare.result == 'success' && needs.prepare.outputs.dry_run != 'true' && needs.prepare.outputs.selected_surfaces != '' }}
     uses: ./.github/workflows/_deploy-server.yml
     with:
       environment: staging
@@ -274,14 +261,14 @@ jobs:
       enabled: ${{ needs.prepare.outputs.server == 'true' }}
     secrets: inherit
 
-  deploy-workers:
-    needs: [prepare, deploy-server]
-    if: ${{ always() && needs.prepare.result == 'success' && needs.prepare.outputs.dry_run != 'true' && needs.prepare.outputs.selected_surfaces != '' }}
-    uses: ./.github/workflows/_deploy-workers.yml
+  deploy-litellm:
+    needs: [prepare]
+    if: ${{ needs.prepare.result == 'success' && needs.prepare.outputs.dry_run != 'true' && needs.prepare.outputs.selected_surfaces != '' }}
+    uses: ./.github/workflows/_deploy-litellm.yml
     with:
       environment: staging
       git_sha: ${{ needs.prepare.outputs.head_sha }}
-      enabled: ${{ needs.prepare.outputs.workers == 'true' }}
+      enabled: ${{ needs.prepare.outputs.litellm == 'true' }}
     secrets: inherit
 
   deploy-web:
@@ -294,37 +281,15 @@ jobs:
       enabled: ${{ needs.prepare.outputs.web == 'true' }}
     secrets: inherit
 
-  deploy-mobile:
-    needs: [prepare]
-    if: ${{ needs.prepare.result == 'success' && needs.prepare.outputs.dry_run != 'true' && needs.prepare.outputs.selected_surfaces != '' }}
-    uses: ./.github/workflows/_deploy-mobile.yml
-    with:
-      environment: staging
-      git_sha: ${{ needs.prepare.outputs.head_sha }}
-      enabled: ${{ needs.prepare.outputs.mobile == 'true' }}
-    secrets: inherit
-
   # ── Production promotion (zero-touch auto-prod) ───────────────────────────────
   # Each prod job promotes only after its staging counterpart deploys cleanly, so
   # the train still validates on staging first — but with no human approval step.
   # Targets the existing `Production` GitHub Environment (prod cluster, deploy
-  # role, prod E2B template, secrets). The Environment's required-reviewer rule
-  # must be removed for these to run unattended.
-
-  deploy-e2b-prod:
-    needs: [prepare, deploy-e2b]
-    if: ${{ always() && needs.prepare.outputs.dry_run != 'true' && needs.deploy-e2b.result == 'success' && needs.prepare.outputs.selected_surfaces != '' }}
-    uses: ./.github/workflows/_deploy-e2b.yml
-    with:
-      environment: Production
-      git_sha: ${{ needs.prepare.outputs.head_sha }}
-      enabled: ${{ needs.prepare.outputs.e2b == 'true' }}
-      rolling_tag: production
-      mode: build_and_promote
-    secrets: inherit
+  # role, secrets). The Environment's required-reviewer rule must be removed
+  # for these to run unattended.
 
   deploy-server-prod:
-    needs: [prepare, deploy-server, deploy-e2b-prod]
+    needs: [prepare, deploy-server]
     if: ${{ always() && needs.prepare.outputs.dry_run != 'true' && needs.deploy-server.result == 'success' && needs.prepare.outputs.selected_surfaces != '' }}
     uses: ./.github/workflows/_deploy-server.yml
     with:
@@ -333,14 +298,14 @@ jobs:
       enabled: ${{ needs.prepare.outputs.server == 'true' }}
     secrets: inherit
 
-  deploy-workers-prod:
-    needs: [prepare, deploy-workers, deploy-server-prod]
-    if: ${{ always() && needs.prepare.outputs.dry_run != 'true' && needs.deploy-workers.result == 'success' && needs.prepare.outputs.selected_surfaces != '' }}
-    uses: ./.github/workflows/_deploy-workers.yml
+  deploy-litellm-prod:
+    needs: [prepare, deploy-litellm]
+    if: ${{ always() && needs.prepare.outputs.dry_run != 'true' && needs.deploy-litellm.result == 'success' && needs.prepare.outputs.selected_surfaces != '' }}
+    uses: ./.github/workflows/_deploy-litellm.yml
     with:
       environment: Production
       git_sha: ${{ needs.prepare.outputs.head_sha }}
-      enabled: ${{ needs.prepare.outputs.workers == 'true' }}
+      enabled: ${{ needs.prepare.outputs.litellm == 'true' }}
     secrets: inherit
 
   deploy-web-prod:
@@ -353,27 +318,15 @@ jobs:
       enabled: ${{ needs.prepare.outputs.web == 'true' }}
     secrets: inherit
 
-  deploy-mobile-prod:
-    needs: [prepare, deploy-mobile]
-    if: ${{ always() && needs.prepare.outputs.dry_run != 'true' && needs.deploy-mobile.result == 'success' && needs.prepare.outputs.selected_surfaces != '' }}
-    uses: ./.github/workflows/_deploy-mobile.yml
-    with:
-      environment: Production
-      git_sha: ${{ needs.prepare.outputs.head_sha }}
-      enabled: ${{ needs.prepare.outputs.mobile == 'true' }}
-    secrets: inherit
-
   publish-product-release:
     needs:
       - prepare
       - release-runtime
       - release-server
       - release-desktop
-      - deploy-e2b
       - deploy-server
-      - deploy-workers
+      - deploy-litellm
       - deploy-web
-      - deploy-mobile
     if: >-
       ${{
         always() &&
@@ -386,10 +339,8 @@ jobs:
             (needs.prepare.outputs.runtime != 'true' || needs.release-runtime.result == 'success') &&
             (needs.prepare.outputs.server != 'true' || (needs.release-server.result == 'success' && needs.deploy-server.result == 'success')) &&
             (needs.prepare.outputs.desktop != 'true' || needs.release-desktop.result == 'success') &&
-            (needs.prepare.outputs.e2b != 'true' || needs.deploy-e2b.result == 'success') &&
-            (needs.prepare.outputs.workers != 'true' || needs.deploy-workers.result == 'success') &&
-            (needs.prepare.outputs.web != 'true' || needs.deploy-web.result == 'success') &&
-            (needs.prepare.outputs.mobile != 'true' || needs.deploy-mobile.result == 'success')
+            (needs.prepare.outputs.litellm != 'true' || needs.deploy-litellm.result == 'success') &&
+            (needs.prepare.outputs.web != 'true' || needs.deploy-web.result == 'success')
           )
         )
       }}
@@ -442,16 +393,12 @@ jobs:
       - release-runtime
       - release-server
       - release-desktop
-      - deploy-e2b
       - deploy-server
-      - deploy-workers
+      - deploy-litellm
       - deploy-web
-      - deploy-mobile
-      - deploy-e2b-prod
       - deploy-server-prod
-      - deploy-workers-prod
+      - deploy-litellm-prod
       - deploy-web-prod
-      - deploy-mobile-prod
     if: ${{ always() && needs.prepare.result == 'success' }}
     runs-on: ubuntu-latest
     steps:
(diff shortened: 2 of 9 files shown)


pablonyx (commented):

Fresh reviewer of record; I did not implement this. Reviewed head `7cbc331bf1d6d5682fc3560ea70d8464737f9f65` statically plus the dry-run receipt. **The workflow wiring itself is correct — every item on the checklist passes.** The blockers below are two truth defects the change creates outside the three edited workflows, plus one customer-facing consequence that needs an explicit owner decision before a launch-eve merge. Both blockers are small.

## What I verified clean

**1. No dangling `needs:` / `if:` / step references.**
- `nightly-release-train.yml`: every `needs:` entry resolves to a defined job — `publish-product-release` (:322-329) and `summary` (:391-401) both now name only `prepare`/`release-*`/`deploy-server`/`deploy-litellm`/`deploy-web`(+`-prod`). The result-gating `if:` (:339-343) and the summary step body (:413-421) reference no deleted job. Grep for `deploy-e2b|deploy-mobile|deploy-workers` in the file returns nothing.
- `deploy-staging.yml`: `deploy-server` (:198), `deploy-web` (:218), `summary` (:243-248) and the summary step body (:262-265) are all clean.
- Proven, not just eyeballed: GitHub resolves `needs:` at parse time, so run 32415247865 could not have started with a dangling edge.

**2. `prepare` outputs.** Deleted `e2b`/`mobile` are consumed nowhere. `litellm` (:62) is real — `detect-deploy-surfaces.mjs:7-16` lists it in `SURFACES` and `:137-139` emits it for `server/litellm` paths, and `--only` parsing (`:224-226`) is untouched. One dead leftover: `nightly-release-train.yml:61` still exports `workers` with zero consumers (same for `deploy-staging.yml:66/68/70` — `workers`/`e2b`/`mobile`). Harmless, but it is now noise; consider dropping `workers` from the train's outputs for symmetry with the `e2b`/`mobile` removal you already did.

**3. `deploy-litellm-prod` mirrors the prod-leg pattern exactly.** `needs: [prepare, deploy-litellm]`, `if: always() && dry_run != 'true' && needs.deploy-litellm.result == 'success' && selected_surfaces != ''`, `environment: Production`, `secrets: inherit` (:301-309) — character-for-character the `deploy-server-prod` shape (:291-298). Inputs match `_deploy-litellm.yml`'s `workflow_call` signature (`environment`, `git_sha`, `enabled`). I checked the live environments rather than trusting the wiring: `staging` and `Production` both have `LITELLM_DEPLOY_ENABLED=true`, `ECS_LITELLM_SERVICE` (`proliferate-staging-litellm` / `proliferate-prod-litellm`), and `Production` carries all five required secrets (`LITELLM_DATABASE_URL`, `LITELLM_MASTER_KEY`, `AGENT_GATEWAY_MANAGED_{ANTHROPIC,OPENAI,XAI}_API_KEY`). This leg will really deploy, not silently no-op.
- **litellm and server never gate each other** — confirmed. Neither chain appears in the other's `needs` or `if`. Worth stating explicitly: that also means a train carrying coupled server+gateway changes has **no ordering guarantee** between the two prod rolls. Fine today; a trap if a future change makes the server depend on new gateway config.
- Skip semantics are safe, and I have empirical proof rather than a guess: in real train 32167413474, `deploy-workers / deploy` was skipped (enabled=false) and `deploy-workers-prod` shows as **skipped at the caller level** — i.e. a caller job whose only inner job skips does *not* report `success`, so its `-prod` counterpart never fires. `deploy-litellm-prod` inherits that behavior: no prod gateway roll on trains that don't touch `server/litellm`.

**4. Intel pause: exactly one matrix entry, and the fallout is handled.** `release-desktop.yml:105-108` comments out the single `macos-15-intel` entry; the matrix retains only `macos-15`/`aarch64`. Nothing else keys off `x86_64` in a way that fails: the `suffix="x64"` branches (:538-541, :567-570, :590-593) are per-matrix-entry and simply never taken; `create-release` uploads `all-artifacts/**/*` with no per-arch enumeration; `generate-updater-manifest.mjs:89-92,114-124` and `generate-desktop-installer-manifest.mjs:51-56,74-78` now skip-with-warning; the cask bump (:856-867) exits 0 instead of `::error::`. I specifically checked the publish path for other hard failures (`Refuse an existing immutable updater manifest`, the S3 upload `find`, CloudFront invalidation) — all arch-agnostic.

**5. Dry-run receipt re-quoted.** Run 32415247865 (`Nightly Release Train`, `workflow_dispatch`, head `7cbc331bf1d6d5682fc3560ea70d8464737f9f65` — the exact PR head), conclusion **success**, 12 jobs:
```
success  prepare
success  publish-product-release
success  summary
skipped  release-runtime      skipped  release-server      skipped  release-desktop
skipped  deploy-server        skipped  deploy-litellm      skipped  deploy-web
skipped  deploy-server-prod   skipped  deploy-litellm-prod skipped  deploy-web-prod
```
Exactly as claimed: prepare/publish/summary green, all 9 deploy+release jobs skipped. **What it proves:** the graph parses, no dangling edge, the new job names exist, and dry-run gating suppresses every deploy. **What it does not prove:** no skipped job's internals ever executed — `_deploy-litellm.yml` never received the new call, so its input binding, the staging→prod ordering, the Production approval pause, the arm-only manifest generators, the Homebrew skip branch, and `publish-updater`'s S3/CloudFront steps against a one-platform artifact set are all first exercised on the first real train. `deploy-staging.yml` was not exercised at all by this run (it is statically clean, but unproven at runtime).

**6. Concurrency untouched.** `nightly-release-train.yml:42-44` (`group: nightly-release-train`, `cancel-in-progress: false`), `deploy-staging.yml:33-46` (composite group, `cancel-in-progress: false`), `release-desktop.yml` (`release-desktop-<version>`, `cancel-in-progress: false`) — no diff hunk touches any of them.

## Blocking

**B1 — Canonical delivery docs are now false (2-line fix).** `specs/codebase/systems/engineering/delivery/README.md:195-196` states "Neither coordinator includes a LiteLLM job. Exact LiteLLM deployment uses the manual production-promotion path," and `:302-303` lists under **Current Gaps**: "Nightly and hotfix coordinators do not include LiteLLM; use manual production promotion for that surface." This PR makes the nightly half of both statements wrong on the day of launch. Repo doctrine (`CLAUDE.md` → `specs/README.md`) is that canonical documentation describes the codebase. Please update both to "hotfix only" and narrow the gap entry. Bonus context in favor of the change: `specs/.../issue-lifecycle/support-loop.md:591-593` already *requires* "Nightly/hotfix must expose, execute, and gate their detected `litellm` surface through `_deploy-litellm.yml`" — this PR satisfies the nightly half of a documented target, which is worth saying in the doc rather than leaving contradictory prose in two places. While there, the workflow inventory (:229-234) should note that `_deploy-e2b.yml`/`_deploy-mobile.yml`/`_deploy-workers.yml` are now called by no automated coordinator (promote/hotfix only).

**B2 — The release-qualification harness hard-rejects an arm-only desktop release.** `tests/release/src/artifacts/retained-release-set.ts:164-167` declares `DESKTOP_PLATFORMS = ["darwin-aarch64", "darwin-x86_64"]` and `:414-417` throws `RetainedReleaseError: desktop.packages is missing supported platform "darwin-x86_64"` when one is absent. Every release produced while Intel is paused will fail that validator. It is **not** on the train's critical path (only `ci.yml` unit tests and the manual `release-e2e*.yml` workflows touch `tests/release`, and those run against fixtures that still supply both platforms — which is why CI is green here), so this will not break the train. But the PR's own audit claim is "audited the surfaces that would otherwise fail once no Intel artifacts exist," and this one was missed. Either make `darwin-x86_64` conditionally-required the same way you made the two manifest generators optional, or record it explicitly as a known-broken qualification gate for the duration of the pause. Silently leaving a throwing validator behind the launch is the wrong default.

## Needs an explicit decision (not mine to make)

**D1 — Intel installs may see recurring update *errors*, not a silent freeze, and can be locked out later.** With `darwin-x86_64` dropped from `latest.json`, Tauri's static-JSON updater has no platform entry to resolve for Intel clients; the likely result is a `UPDATER_CHECK_FAILED` on every scheduled check (`specs/.../delivery/desktop-updates.md:191`), with Sentry noise, rather than the clean "no update available" the PR body implies. I could not prove the exact client behavior without running it — please confirm before merge. Worse, the same doc's Version-Skew section: `minDesktopVersion` defaults to the server's own stamped desktop version, so Intel installs freeze at today's build while the server floor advances — the day `minDesktopVersionEnforced` is turned on, every Intel user is hard-blocked with no upgrade path. The PR discloses "no auto-updates"; it does not disclose "may error on every check" or "may be locked out." Both deserve a conscious call plus user comms, especially the day before a launch.

**D2 — The nightly train no longer advances the E2B cloud template.** The PR frames this as "e2b is no longer shipped," but E2B still looks live in production: `Production` env carries `E2B_TEMPLATE_REF`, `E2B_TEAM_ID`, `E2B_PUBLIC_TEMPLATE_FAMILY`; `_deploy-server.yml:448` verifies E2B secret projection on every server deploy; `specs/FEATURE_DOCS/SANDBOX/lifecycle.md` still describes `_deploy-e2b.yml` as the cloud runtime lane. Concretely: `detect-deploy-surfaces.mjs:151-164` marks any `anyharness/crates/**`, `Cargo.lock`, or `catalogs` change as both `e2b` **and** `runtime`. Post-merge, such a change still cuts a runtime release but the cloud template's baked binaries stop advancing, so the cloud sandbox drifts behind server/desktop until someone manually runs `promote-production.yml` (which still carries `deploy-e2b`) or the standalone cloud-template workflows. If that drift is intended and accepted, say so in the PR body so the next operator does not discover it during an incident. If cloud sessions are still a launch surface, this is the item I would think hardest about.

## Non-blocking

- **N1** — No regression test pins the new arm-only behavior. `scripts/ci-cd/generate-updater-manifest.test.mjs:57` still asserts `["darwin-aarch64","darwin-x86_64"]` and supplies both fixtures, so it passes without exercising the `optional` skip at all. The Intel pause is now load-bearing on those two `optional: true` flags; a ~10-line case feeding an aarch64-only artifact dir would keep a future edit from re-breaking publishing. The PR's manual `node <script>` verification is not repeatable in CI.
- **N2** — Operator trap: `only_surfaces` is free text validated against `detect-deploy-surfaces.mjs:7-16`, which still accepts `e2b`/`mobile`/`workers`. Dispatching the train with `only_surfaces=e2b` now yields a version bump, tags, and a product GitHub Release whose Surfaces row (`scripts/ci-cd/publish-product-release.mjs:357`) advertises `e2b` — while deploying nothing. `SURFACES` can't simply be trimmed (promote/hotfix still use those lanes), so this needs a line in the release procedure doc.
- **N3** — FYI, the `Production` environment currently has `required_reviewers` protection, so `deploy-litellm-prod` will pause for approval exactly like `deploy-server-prod`/`deploy-web-prod`. The in-file "zero-touch auto-prod" comment (:284-289) remains aspirational, unchanged by this PR — noting it only so nobody expects an unattended gateway roll tonight.
- **N4** — Documented scope gap holds: `promote-production.yml` and `hotfix-production.yml` still carry `deploy-e2b`/`deploy-mobile`/`deploy-workers`, and the reusable `_deploy-*.yml` files remain on disk unreferenced by the two automated coordinators. Verified, correctly disclosed, and fine as follow-up.

Fix B1 and B2 (or record B2 as an accepted, tracked breakage), get an explicit answer on D1 and D2, and I am happy to re-review — the mechanical part of this change is well executed and the dry-run proof is genuine.

pablonyx:
Addressed both blockers from review 4988181298 (commit f9bbf7b22b).

**B1** — `specs/codebase/systems/engineering/delivery/README.md:195-202` and `:310-311` no longer claim "neither coordinator includes a LiteLLM job." They now say the nightly coordinator includes a LiteLLM job (staging leg parallel to Server, production leg chained off the staging leg's result, same shape as the other hosted surfaces), while the hotfix coordinator still doesn't and still uses the manual production-promotion path for LiteLLM.

**B2** — `tests/release/src/artifacts/retained-release-set.ts:164-175` splits the old single `DESKTOP_PLATFORMS` list into `SUPPORTED_DESKTOP_PLATFORMS` (both platforms, still rejects unrecognized platform strings) and `REQUIRED_DESKTOP_PLATFORMS` (`darwin-aarch64` only, for the duration of the Intel pause). The missing-platform check at :423-427 now only enforces the required set, so an arm-only retained-release receipt validates. The comment on `REQUIRED_DESKTOP_PLATFORMS` ties the pause to `release-desktop.yml`'s `build-desktop` matrix block, so restoring Intel is a one-line change in both places. `darwin-aarch64` checks are untouched.

Also added the minimal regression coverage the review's N1 flagged:
- `scripts/ci-cd/generate-updater-manifest.test.mjs`: new "generates an arm-only manifest while Intel is paused" case, exercising the `optional: true` skip path with no x86_64 fixture.
- `tests/release/src/artifacts/retained-release-set.test.ts`: updated the missing-platform test to assert against the required set (`darwin-aarch64`) and added "accepts an arm-only desktop receipt while Intel is paused" pinning the new behavior.

Ran both test files locally (node/tsx, single files only): all pass, including the two new/updated cases.

**D1 and D2 remain open decisions for Pablo** — not resolved by this push. D1 (Intel installs likely hit `UPDATER_CHECK_FAILED` on every check rather than a silent freeze, plus the later `minDesktopVersionEnforced` lockout risk) and D2 (nightly train no longer advances the E2B cloud template while E2B still looks live in prod) both need an explicit call before merge.

pablonyx:
REVIEW VERDICT: LGTM-FOR-MERGE (conditional — see the two gates at the bottom)

Re-verified `f9bbf7b22b21b689dee04c7be9379f32949227ee` against my earlier findings. **B1 and B2 are both properly fixed, and I confirmed the new tests actually ran in CI rather than only locally.**

**B2 — arm-only releases (`tests/release/src/artifacts/retained-release-set.ts`).** The split is exactly right and `darwin-aarch64` enforcement is **unweakened**:
- `SUPPORTED_DESKTOP_PLATFORMS` (:165-168, both platforms) still backs the per-package check at :398, so unknown platform strings are still rejected *and* a receipt that does carry an Intel package is still accepted. Duplicate detection (:401-404) is untouched.
- `REQUIRED_DESKTOP_PLATFORMS` (:177-179, aarch64 only) backs the required loop at :426-429, which still throws when aarch64 is absent. The narrowing is scoped to the Intel entry alone.
- The comment ties the pause to the `release-desktop.yml` matrix block and names the one-line restore. Good — that is the thing that makes an un-pause not miss this file.
- No stale `DESKTOP_PLATFORMS` reference or `missing supported platform` string survives on the branch; the message rename to `missing required platform` is consistent between source (:428) and assertion (`retained-release-set.test.ts:134`).

**The two new tests genuinely pin the arm-only path**, and they pin it asymmetrically, which is the part that matters. The fixture orders packages `[darwin-aarch64, darwin-x86_64]` (`retained-release-set.test.ts:45,54`), so the retargeted rejection test's `.shift()` removes **aarch64** and must throw, while the new acceptance test's `.pop()` removes **x86_64** and must pass (asserting length 1 and `platform === "darwin-aarch64"`). Together they prove the required/optional split rather than just re-asserting the new behavior. The updater-manifest test (`generate-updater-manifest.test.mjs:63-71`) filters the x86_64 fixture dir, asserts exit 0, asserts the exact `Skipping darwin-x86_64: no artifacts found (optional platform).` warning, and asserts `Object.keys(manifest.platforms) === ["darwin-aarch64"]` exactly — it would catch both a silent failure and a phantom platform entry.

**CI-enforced, re-quoted from this head's logs rather than from the local run report:**
- `CI/CD config checks` (job 96609385498, SUCCESS): `ok 100 - generates an arm-only manifest while Intel is paused`, suite `# tests 204 / # pass 204 / # fail 0`.
- `Candidate build handoff` (job 96609385453, SUCCESS) runs `pnpm -C tests/release test` and `typecheck`: `ok 37 - rejects a missing required desktop platform (darwin-aarch64)`, `ok 38 - accepts an arm-only desktop receipt while Intel is paused`, suite `# tests 1561 / # pass 1561 / # fail 0`.
That closes my N1 for the updater manifest. Note `scripts/generate-desktop-installer-manifest.mjs` — the *other* file that gained `optional: true` — still has no test anywhere in `scripts/ci-cd/*.test.mjs`. Pre-existing absence of a harness, not a regression; leaving it as a non-blocking note.

**B1 — docs.** `specs/.../delivery/README.md:195-200` now describes the nightly LiteLLM job in its real shape (staging leg parallel to Server, production leg chained off the staging leg's result) and correctly scopes the manual-promotion path to hotfix; Current Gaps (:310-311) is narrowed to the hotfix coordinator. Accurate against the workflow as written. Still open and purely cosmetic: the workflow inventory (:229-234) does not note that `_deploy-e2b.yml` / `_deploy-mobile.yml` / `_deploy-workers.yml` are now called by no automated coordinator. "Reusable only" stays literally true, so this does not block.

**D1 / D2 carry forward as open decisions for Pablo, not blockers.** The PR body's new "Known behavior while Intel is paused" section addresses my D1 disclosure gap. The underlying calls — accepting possible `UPDATER_CHECK_FAILED` noise plus the `minDesktopVersion` lockout risk on Intel installs (D1), and accepting that the nightly train no longer advances the E2B cloud template so `anyharness/crates/**` changes need a manual promotion (D2) — are founder decisions, not review findings. My non-blocking N2 (`only_surfaces=e2b` still cuts a release that deploys nothing), N3 (`Production` has `required_reviewers`, so the new prod leg pauses for approval), and N4 (promote/hotfix still carry the old jobs) are unchanged and correctly disclosed.

**Two gates before merge, neither about this diff:**
1. **The head is not green.** `Shared frontend packages` is FAILURE at `f9bbf7b22` (job 96609385383) and `mergeStateStatus` is `UNSTABLE`. It is **not** caused by this change — the commit touches only workflows, `scripts/`, `specs/`, and `tests/release`, nothing under `apps/`; the run reports `1093 passed (1093)` test files and `8168 passed (8168)` tests, and the sole error is a post-teardown `ReferenceError: window is not defined` from a `@tanstack/virtual-core` timer firing after jsdom teardown in `VirtualizedTranscriptRowList.test.tsx`. The same job passed at the parent commit `7cbc331bf`, and recent `main` CI is green, so this reads as an async-teardown flake rather than a real regression. Still: re-run it to green before merging. Do not merge on "it's just the flake."
2. **The PR is still a draft** and, per standing policy, merges are Pablo's.

With those two satisfied and D1/D2 consciously accepted, I have no remaining objection.

PR #2131, opened 2026-08-20
ci(server): shard the integration lane 3-way

pablonyx:
## Hypothesis

`test-integration` on `exp/server-split-before-only` runs 1068 tests + 1 skip
under xdist `-n 4` and measured 9m9s-10m9s wall clock. Splitting it into a
3-way matrix, with each shard keeping its own postgres+redis services and
`max_locks_per_transaction=1024` setup but running only a third of the test
files, should bring the integration lane down to roughly 4-5 minutes per
shard while the total job-minutes stay similar.

## What changed

`test-integration` gains `strategy: matrix: shard: [1, 2, 3]`. A new step
lists every file under `tests/integration` (`test_*.py` / `*_test.py`) plus
the one e2e file this lane also carries (`tests/e2e/cloud/test_e2b_webhooks.py`),
sorts them for a stable order independent of filesystem traversal, and keeps
every third file (`index % 3 == shard - 1`) for that shard. Whole files stay
together so no test can land in two shards or in none, and the partition is
fully deterministic across runs. `pytest -n 4 -q --durations=20` then runs
against just that shard's files.

`pytest-split` was considered but a plain file-level partition was judged
safer here: no new pinned dependency, no `.test_durations` cache to keep in
sync, and the union-equals-original-set property is trivially true by
construction rather than needing separate verification.

## Base branch

Based on `exp/server-split-before-only` (#2130), which has not merged yet.
This PR is layered on top of that split for measurement purposes.

## Status

Measurement-only. This is a CI experiment branch to quantify the effect of
sharding the integration lane, not a mergeable change as-is — the matrix
jobs would need to be reconciled with branch-protection required-check names
before this could land for real.

### DIFF
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -183,6 +183,10 @@ jobs:
   test-integration:
     runs-on: ubuntu-latest
     timeout-minutes: 75
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2, 3]
     env:
       DATABASE_URL: postgresql+asyncpg://proliferate:localdev@127.0.0.1:5432/proliferate
       JWT_SECRET: test-jwt-secret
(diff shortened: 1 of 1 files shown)


pablonyx:
Sample 2 (run [32429121218](https://github.com/proliferate-ai/proliferate/actions/runs/32429121218), empty commit, identical code and partition, all green):

| Shard | Wall clock | pytest summary |
|---|---|---|
| 1 | 4m46s | `286 passed, 1 skipped, 237 warnings in 216.77s (0:03:36)` |
| 2 | 3m52s | `397 passed, 261 warnings in 176.12s (0:02:56)` |
| 3 | 5m10s | `385 passed, 389 warnings in 255.68s (0:04:15)` |

Counts again sum exactly to `1068 passed, 1 skipped`.

Cross-sample read: shard 3 is byte-stable (255.55s vs 255.68s) while shard 2 swung 439.95s → 176.12s on identical files, so sample 1's 8m28s pole was **runner variance concentrated in one shard, not file imbalance** — the sorted-modulo partition is roughly duration-balanced (~3-4m15s per shard when quiet). Lane pole across two samples: **5m10s-8m28s** (was 9m9s-13m31s unsharded). A pytest-split duration-balanced partition is optional polish, not the lever.

PR #2135, opened 2026-08-21
feat(dev): disposable EC2 testbed for multi-profile measurement

pablonyx:
Measuring how the product behaves at five concurrent profiles needs a machine that can be wrecked and rebuilt, which the laptop is not. `scripts/fleet-testbed/provision.sh --arch x86_64` produces one, and blocks until the instance reports ready with the resolved commit, runtime version, agent CLI version, and the absence of cargo, so the box is proven to match its pin rather than merely booted.

## Safety, and why it is shaped this way

AWS account 157466816238 is both the Bedrock/YC-credit account and production. It holds ECS `proliferate-production`, RDS `proliferate-production`, and the metabase tunnel host, and production lives in the DEFAULT VPC. A teardown that selected on a tag alone would be one typo from terminating production.

So the testbed lives in a dedicated non-default VPC at `10.42.0.0/16`, which nothing else uses, and every resource in it carries `Project=proliferate-fleet-testbed`. Resolution for any destructive operation asserts three things and refuses anything but exactly one match: the tag, an exact match on that CIDR, and `IsDefault=false`. Both entry points additionally assert the caller's AWS account before touching anything, overridable with `FLEET_TESTBED_ACCOUNT_ID`. Every teardown delete carries the tag filter as well as VPC membership, so no single wrong value widens a selection.

Rules the scripts enforce rather than document:

- No GitHub token is ever a variable in `user-data.sh`. It stays in an SSM SecureString and is fetched on demand by a mode-700 `core.askPass` helper running under the instance role. That keeps it out of xtrace and the setup logs, out of argv, and out of `.git/config`, and the clone is followed by an assertion that no credential got persisted anyway. `put-parameter` reads the token from a mode-600 temp file via `file://`, so it is not in the operator's process table either.
- No agent API key is written to the box. Agent auth is Bedrock through the instance IAM role, which the runtime already supports as a first-class provider, and the role's `bedrock:InvokeModel` is scoped to Anthropic Claude foundation-model and inference-profile ARNs rather than `*`.
- IMDSv2 is required with a hop limit of 1.
- Before anything that can fail, user-data arms a deadman against an instance launched with `--instance-initiated-shutdown-behavior terminate`. The deadline is a timestamp on disk, checked by a persistent systemd timer and a `cron.d` entry, so a reboot cannot disarm it, and a failure to arm degrades to a one-shot `shutdown` rather than aborting the setup at the exact line meant to be the safety net.

SSH is open to the caller's public IP as a /32 and nowhere else, and the security group's entire ingress set is revoked before the current rule is authorized, so /32s do not accumulate toward the 60-rule quota. The IAM policy scopes `ssm:GetParameter` to the prefix and `kms:Decrypt` to `ssm.*` via-service.

## Residual risk, stated plainly

Three things this design does not solve, which earlier revisions of this description overstated:

**The token is still a broad personal PAT, and the box still runs untrusted code as the user that can read it.** Moving the credential out of `.git/config` and into SSM moves it from "readable by `ubuntu` off disk" to "fetchable by `ubuntu` through the instance role". That is a real improvement against the leak paths that actually bit here (logs, argv, a credential persisted at 0644 forever, and a token that survives the box in a git config), but it is not confinement. `HttpTokens=required` costs a local process one extra PUT, and `HttpPutResponseHopLimit=1` restricts other hosts, not other processes on this one. The machine exists to run agent profiles executing model-generated code as `ubuntu`. The default token is Pablo's `gh` credential carrying `repo, workflow, gist, read:org`, which is far more than a clone of one repository needs. The right answer is a fine-grained read-only PAT scoped to `proliferate-ai/proliferate`, which needs Pablo to mint it. This PR does not block on that: `FLEET_TESTBED_TOKEN_FILE=/path/to/pat` or `FLEET_TESTBED_TOKEN_COMMAND='...'` supplies one with no edit to the scripts, and `gh auth token` remains the default until a scoped PAT exists. Treat any box provisioned before then as holding a credential with write access to the org.

**Scoping the Bedrock ARNs narrows reach, not quota.** Bedrock's on-demand limits are per account and per model, so a runaway testbed agent still draws on the same capacity production inference uses. The ARN scoping means the box cannot reach non-Anthropic models at all, which is worth having, but genuine quota isolation would need a separate account or provisioned throughput and is out of scope here.

**One of the two log files is tightened rather than born locked.** `/var/log/fleet-testbed-setup.log` is created 0600 before it has a byte in it. `/var/log/cloud-init-output.log` is not ours to create: cloud-init opens it early in boot and has already written its own module output there before user-data runs. It is chmodded 0600 as the second statement of the script, which is before any output of *this* script reaches it, and that is the accurate claim.

## Why x86_64 is the default

Graviton is cheaper per core and the runtime ships an aarch64 musl build, so arm64 was the obvious choice and it is wrong. The `agent_process` sidecar has no arm64 pin, so `POST /v1/agents/claude/install` returns `AGENT_NO_PIN_FOR_PLATFORM` for `linux_arm64` and every session sits at `install_required`. `--arch arm64` still works and is fine for memory and boot measurement, but no agent can run a turn on it. Filing that separately as a product defect, since it also means a self-hosted Graviton customer cannot install the agent.

## Repeatability

The instance is pinned to the commit behind release `server-v0.4.20` and the prebuilt musl binary from that same release, so the runtime and the source tree cannot disagree and no schema skew can be mistaken for a finding. The dev-loop fixes are cherry-picked on top rather than checked out, so the tree stays at the release. That branch is #2134; `FIX_BRANCH` points at it, and once it merges the cherry-pick step logs that there is nothing to pick and moves on. It does not become a no-op by accident: `git cherry-pick` with no arguments exits 129, so the empty case is handled explicitly.

Everything but the instance is idempotent. VPC, subnet, gateway, route table, security group, key pair, IAM role, and instance profile are created on first run and reused, the inline policy and role attachment are re-applied on every run so an edit here reaches an existing deployment, and the SSM token is rewritten every run so a rotated or revoked token cannot silently break future clones.

## Review round 1 findings

| # | Finding | Fix |
| --- | --- | --- |
| S1 | The token leaked three ways: `set -x` plus `tee` wrote `GH_TOKEN=gho_...` and the full clone command into two unrestricted logs, one of which the summary told the operator to tail; the token was expanded into a `su -c` argv visible in `/proc/<pid>/cmdline`; and `git clone` with the credential in the URL persisted it into `.git/config` at 0644, which the later `git fetch` then depended on. | The token is no longer referenced in `user-data.sh` at all. A mode-700 `/usr/local/bin/fleet-gh-askpass` owned by `ubuntu` reads the SSM SecureString with the instance role and prints the credential on stdout; the remote URL is fully tokenless, so nothing is persisted and later fetches work through the same helper. xtrace is disabled across the whole credential region, the log files are locked down before any of this script's output reaches them, the summary says `sudo tail`, and the clone is followed by assertions that fail the setup if any userinfo appears in `.git/config` or a token prefix appears on disk. |
| S2 | The resolver filtered on `tag:Project` alone and took `Vpcs[0]`, which has no ordering guarantee; `VPC_CIDR` was declared and never used (SC2034), so the "two independent predicates" claim was one predicate applied twice. | `resolve_testbed_vpc` filters on tag, `cidr` (an exact primary-CIDR match), and `is-default=false`, then re-asserts the id prefix, the CIDR, and `IsDefault` on the returned row, and dies on anything other than exactly one match instead of picking. |
| S3 | `teardown.sh --network` enumerated security groups, subnets, route tables, and internet gateways by VPC id only, with no tag filter. | Verified against `provision.sh` that all four are created with `tag_spec`, then added the tag filter to each describe. The ids are resolved into variables first, so a failing describe aborts instead of yielding an empty list that reads as "nothing to do". |
| S4 | Six `\|\| true`s meant `log "network deleted"` printed even when every delete failed. | Failures are counted and reported by name, the success line prints only at zero failures, and the script exits non-zero. Deletes go through a bounded retry that waits for ENIs to drain, with the last attempt un-silenced so the operator sees the real AWS error. |
| S5 | Teardown never removed the SSM SecureString, the IAM role, the instance profile, or the key pair; `provision.sh` also skipped the token when the parameter already existed, so a revoked token silently broke every future clone. | `--network` deletes the SSM parameter and `--all` additionally deletes the inline policy, role, instance profile, and key pair including the local private key. `provision.sh` writes the token on every run. |
| S6 | SSH /32s accumulated forever behind an `\|\| true`, and at the 60-rule quota new authorizes failed silently. | The ingress set is revoked before the current /32 is authorized, and the authorize no longer swallows its exit status. |
| S7 | Key-pair idempotency checked only the AWS side, and `> "$KEY_PATH"` truncated the file before `create-key-pair` ran; `chmod 600` came after creation. | Both sides are checked with a specific remedy per mismatch. Creation happens under `umask 077` into a temp file in the same directory, `mv`d into place only after the material is confirmed non-empty. |
| S8 | IAM idempotency was gated on `get-role`, so a run that died mid-way never self-repaired and policy edits never reached an existing deployment. | `put-role-policy` and `add-role-to-instance-profile` run unconditionally; an `add-role` failure is verified against the actual attachment before being treated as benign. |
| S9 | IMDSv1 was enabled. | `--metadata-options HttpEndpoint=enabled,HttpTokens=required,HttpPutResponseHopLimit=1`. See the residual-risk note on what this does and does not buy. |
| S10 | `bedrock:InvokeModel` on `Resource: "*"`, plus an `ec2:TerminateInstances` grant with no consumer. | Scoped to `arn:aws:bedrock:*::foundation-model/anthropic.claude-*` and `arn:aws:bedrock:*:<account>:inference-profile/*anthropic.claude-*`. `ec2:TerminateInstances` deleted: the deadman uses `shutdown -h` against `instance-initiated-shutdown-behavior=terminate`. Quota is still shared with production; see residual risk. |
| S11 | Nothing checked which account the caller was in. | `require_expected_account`, called first by both entry points, overridable with `FLEET_TESTBED_ACCOUNT_ID`. |
| S12 | Unquoted `sed` templating of refs into a root-executed script. | `require_safe_ref` restricts every templated value to `[A-Za-z0-9._/@+-]` and rejects a leading dash; `require_positive_int` covers the numeric ones. |
| S13 | Neither entry point set `set -euo pipefail` itself. | Both set it before sourcing, and the source line fails hard with its own message. |
| S14 | The deadman was a transient `systemd-run` unit in `/run`, disarmed by any reboot, and armed under `set -e` so its own failure aborted the setup. | An on-disk deadline plus a persistent `fleet-deadman.timer` and a `/etc/cron.d/fleet-deadman` entry, with a checker that fails closed. Arming runs inside an `if` and falls back to a one-shot `shutdown -h +N`. |

## Review round 2 findings

Round 2 confirmed ten of the fourteen above and found five blocking defects, two of them introduced by the round-1 commit.

| # | Finding | Fix |
| --- | --- | --- |
| B1 | Gating `build-essential` behind `--with-rust` also removed `make`, which `make build` and `make setup` need. The failure lands about twenty minutes in and then `provision.sh` waits the full hour before a generic timeout. | `make` is installed unconditionally and the compilers stay gated. Independently confirmed against both release manifests (`ubuntu-24.04-server-cloudimg-{amd64,arm64}.manifest`, 664 and 663 packages): no `make`, no `build-essential`, no compiler driver on either arch, though `libc6-dev` and `patch` are present. A `command -v` check over every tool this script invokes by name (`git curl jq make psql unzip`) now fails in seconds rather than mid-build. |
| B2 | `VPC_ID="$(find_testbed_vpc)"` ran the resolver in a command-substitution subshell where bash unsets errexit, so a failing `describe-vpcs` returned empty with status 0. That made `teardown --network` a silent no-op over a live instance, let `teardown --all` delete the key pair out from under a running m7i.4xlarge while printing "teardown complete", and made `provision.sh` create a second VPC at the same CIDR. | The resolver is now `resolve_testbed_vpc`, which sets `TESTBED_VPC_ID` in the caller's shell and dies on a failed describe. `inherit_errexit` is not the fix: macOS ships bash 3.2, where the option does not exist at all, which is verified in the harness. Every other function-in-a-capture was checked for the same shape; `read_github_token` is written to be called with a redirect for the same reason, which also keeps the token from ever becoming a variable in `provision.sh`. `tag_spec` is a bare `echo` and cannot fail. Teardown additionally refuses to delete a key pair while any instance is still using it. |
| B3 | The early `exit 0` for "no testbed VPC" preceded the SSM deletion block, so `--network` never deleted the token once the VPC was already gone, directly contradicting S5 and the comment above the block it skipped. | The early exit fires only when neither `--network` nor `--all` was requested. The VPC-scoped work stays inside the VPC guard; the SSM and IAM blocks sit outside it. |
| B4 | `git cherry-pick` with no arguments exits 129, so an empty `rev-list` is a hard failure, not the no-op the description claimed, the moment #2134 merges. | The pin step is a small generated script with the empty case handled and logged. The old and new forms were run against a fixture repo in both states: old exits 129, new exits 0 with "nothing to cherry-pick". The Repeatability section above is corrected. |
| B5 | Moving the token from `.git/config` to SSM is a smaller improvement than the description implied, given the box runs model-generated code as the user that can fetch it, and the token is a broad personal PAT. | Code part done: the token source is pluggable via `FLEET_TESTBED_TOKEN_FILE` or `FLEET_TESTBED_TOKEN_COMMAND`, defaulting to `gh auth token`. Minting a scoped fine-grained PAT is Pablo's call and is not blocking this PR. The claim is corrected in the Residual risk section rather than restated. |

Non-blocking items from the same round, all taken: the placeholder guard now matches `__[A-Z][A-Z0-9_]*__` rather than four named placeholders; the SSH revoke drops the entire ingress set, because a range rule or an `IpProtocol` of `-1` reaches port 22 without matching an exact-22 predicate; `require_ipv4` rejects `00.0.0.0`, `1.2.3.`, `1.2.3.4.5`, and out-of-range octets instead of counting dots; the `.git/config` assertion flags any userinfo rather than only the `user:secret@` form, and a second check scans for `gh[pousr]_`/`github_pat_` on disk; the SecureString no longer carries the trailing newline `file://` was putting in it; the Bedrock quota point is stated above; and the `cloud-init-output.log` claim is corrected in the code comment as well as here.

## Verification

Static, since running these against the account is the thing worth not getting wrong.

- `bash -n` on all four files, and on the rendered `user-data.sh`, the generated askPass helper, and the generated pin script after templating.
- `shellcheck -x` clean at warning level and above. Two SC2016 infos remain, both intentional (a JMESPath literal and one remote command string).
- The inline IAM policy parsed as JSON with representative substitutions.
- An offline harness stubbing `aws` and `gh`, 41 cases, no network and no AWS: a `describe-vpcs` that exits 254, 1, 143, or prints a production-shaped row and then fails must all abort, and must abort before any delete runs; genuine absence yields empty with status 0; default VPC, wrong CIDR, three matches, a non-`vpc-` id, and a blank `IsDefault` are each refused; sixteen IPv4 cases; and the token reader trims, rejects empty/whitespace/unreadable/failing sources, honours both override variables, and propagates failure through the redirect form. That last one carries its own negative control showing the `$(...)` form would have swallowed it.
- Control-flow runs of `teardown.sh` against a fake `aws` on `PATH`, asserting both status and the exact set of API calls made: `--network` with no VPC deletes the token, a failing describe deletes nothing and leaves the private key on disk, and a live instance holding the key pair blocks its deletion. The pre-fix `teardown.sh` from the previous commit was run against the same scenario as a negative control and printed "nothing to do" without ever calling `ssm delete-parameter`.
- The same treatment for `provision.sh`: a failing describe aborts before `create-vpc`, and genuine absence reaches it.

## Review notes

- `teardown.sh` with no flag removes instances only, which is the normal case, because leaving the network makes the next provision faster and costs nothing. `--network` adds the VPC, its dependencies, and the SSM token. `--all` adds the IAM role, instance profile, and key pair.
- The measurement scripts themselves (`boot.sh`, `measure.sh`) are still living on the instance rather than in the repo. They belong here in a follow-up so the whole loop is version controlled.
- The `--with-rust` path exists but is unexercised, so the concurrent-cargo-build experiment is not yet runnable.

### DIFF
--- a/scripts/fleet-testbed/lib.sh
+++ b/scripts/fleet-testbed/lib.sh
@@ -1,0 +1,71 @@
+#!/usr/bin/env bash
+# Shared configuration for the fleet concurrency testbed.
+#
+# Every resource this lane creates carries TAG_KEY=TAG_VALUE and lives inside a
+# dedicated VPC. Teardown requires BOTH predicates to match, so no single
+# mistake can select a production instance in the same account.
+
+set -euo pipefail
+
+export AWS_DEFAULT_REGION="${AWS_DEFAULT_REGION:-us-east-1}"
+
+TAG_KEY="Project"
+TAG_VALUE="proliferate-fleet-testbed"
+NAME_PREFIX="proliferate-fleet-testbed"
+
+VPC_CIDR="10.42.0.0/16"
+SUBNET_CIDR="10.42.1.0/24"
+
+KEY_NAME="${NAME_PREFIX}"
+KEY_PATH="${HOME}/.ssh/${NAME_PREFIX}.pem"
+ROLE_NAME="${NAME_PREFIX}"
+PROFILE_NAME="${NAME_PREFIX}"
+SSM_PREFIX="/${NAME_PREFIX}"
+
+# Architecture. The agent_process sidecar the runtime installs is pinned for
+# linux_x64 only, so x86_64 is the default: on arm64 the install endpoint fails
+# with AGENT_NO_PIN_FOR_PLATFORM and no session can run a turn.
+ARCH="${ARCH:-x86_64}"
+case "$ARCH" in
+  x86_64) AMI_ARCH=amd64;  AWSCLI_ARCH=x86_64;  MUSL_ARCH=x86_64;  DEFAULT_TYPE=m7i.4xlarge ;;
+  arm64)  AMI_ARCH=arm64;  AWSCLI_ARCH=aarch64; MUSL_ARCH=aarch64; DEFAULT_TYPE=m8g.4xlarge ;;
+  *) echo "unsupported ARCH: $ARCH (want x86_64 or arm64)" >&2; exit 1 ;;
+esac
+
+# Ubuntu 24.04, resolved from Canonical's public SSM parameter so the AMI is
+# never a stale literal.
+AMI_SSM_PARAM="/aws/service/canonical/ubuntu/server/24.04/stable/current/${AMI_ARCH}/hvm/ebs-gp3/ami-id"
+
+# Prebuilt runtime. Pinning the repo to this release's own commit keeps binary
+# and source exactly in step, so no schema skew can be mistaken for a finding.
+RELEASE_TAG="${RELEASE_TAG:-server-v0.4.20}"
+RELEASE_COMMIT="${RELEASE_COMMIT:-be045d63dafc7a74838aa17f00dfb2f6a4e18ed7}"
+
+# Branch carrying the dev-loop fixes the testbed needs (HEADLESS, dev-build).
+# Cherry-picked onto RELEASE_COMMIT rather than checked out, so the tree stays
+# at the release and binary and source cannot disagree.
+FIX_BRANCH="${FIX_BRANCH:-perf/dev-loop-launch}"
+
+# Deadman. The instance shuts itself down after this long and terminates,
+# because run-instances sets instance-initiated-shutdown-behavior=terminate.
+DEADMAN_HOURS="${DEADMAN_HOURS:-12}"
+
+tag_spec() {
+  # $1 = resource type
+  echo "ResourceType=$1,Tags=[{Key=${TAG_KEY},Value=${TAG_VALUE}},{Key=Name,Value=${NAME_PREFIX}}]"
+}
+
+log() { printf '[fleet-testbed] %s\n' "$*" >&2; }
+
+die() { printf '[fleet-testbed] ERROR: %s\n' "$*" >&2; exit 1; }
+
+require_testbed_vpc() {
+  # Resolves the dedicated VPC id, failing loudly rather than falling back to
+  # the default VPC, which in this account holds production instances.
+  local vpc_id
+  vpc_id=$(aws ec2 describe-vpcs \
+    --filters "Name=tag:${TAG_KEY},Values=${TAG_VALUE}" \
+    --query 'Vpcs[0].VpcId' --output text)
+  [ "$vpc_id" != "None" ] && [ -n "$vpc_id" ] || die "no testbed VPC found; run provision.sh first"
+  echo "$vpc_id"
+}
--- a/scripts/fleet-testbed/lib.sh
+++ b/scripts/fleet-testbed/lib.sh
@@ -1,9 +1,15 @@
 #!/usr/bin/env bash
 # Shared configuration for the fleet concurrency testbed.
 #
-# Every resource this lane creates carries TAG_KEY=TAG_VALUE and lives inside a
-# dedicated VPC. Teardown requires BOTH predicates to match, so no single
-# mistake can select a production instance in the same account.
+# This account also runs production, and production lives in the DEFAULT VPC.
+# Every resource this lane creates therefore carries TAG_KEY=TAG_VALUE and lives
+# inside a dedicated non-default VPC whose CIDR nothing else uses. Selection for
+# any destructive operation asserts all three: the tag, the CIDR, and
+# IsDefault=false. See find_testbed_vpc.
+
+# Every value defined below is consumed by the scripts that source this file,
+# which a per-file unused-variable check cannot see.
+# shellcheck disable=SC2034
 
 set -euo pipefail
 
@@ -21,6 +27,12 @@ KEY_PATH="${HOME}/.ssh/${NAME_PREFIX}.pem"
 ROLE_NAME="${NAME_PREFIX}"
 PROFILE_NAME="${NAME_PREFIX}"
 SSM_PREFIX="/${NAME_PREFIX}"
+
+# The account this lane is expected to run in. The safety story is "this account
+# also holds production", so running the scripts anywhere else is a mistake
+# worth stopping on. Override deliberately with FLEET_TESTBED_ACCOUNT_ID when
+# someone is legitimately using a different account.
+EXPECTED_ACCOUNT_ID="${FLEET_TESTBED_ACCOUNT_ID:-157466816238}"
 
 # Architecture. The agent_process sidecar the runtime installs is pinned for
 # linux_x64 only, so x86_64 is the default: on arm64 the install endpoint fails
@@ -59,13 +71,72 @@ log() { printf '[fleet-testbed] %s\n' "$*" >&2; }
 
 die() { printf '[fleet-testbed] ERROR: %s\n' "$*" >&2; exit 1; }
 
-require_testbed_vpc() {
-  # Resolves the dedicated VPC id, failing loudly rather than falling back to
-  # the default VPC, which in this account holds production instances.
-  local vpc_id
-  vpc_id=$(aws ec2 describe-vpcs \
+require_expected_account() {
+  # Everything below assumes production shares this account. Assert it, so the
+  # tag/CIDR/default-VPC guards are being applied where they were designed.
+  local account
+  account=$(aws sts get-caller-identity --query Account --output text) \
+    || die "could not read the caller identity; are AWS credentials configured?"
+  [ -n "$account" ] && [ "$account" != "None" ] \
+    || die "sts get-caller-identity returned no account"
+  if [ "$account" != "$EXPECTED_ACCOUNT_ID" ]; then
+    die "refusing to run in account ${account}; expected ${EXPECTED_ACCOUNT_ID}. Set FLEET_TESTBED_ACCOUNT_ID=${account} if that is deliberate."
+  fi
+  ACCOUNT_ID="$account"
+}
+
+require_safe_ref() {
+  # $1 = label, $2 = value. Values templated into a root-executed user-data
+  # script must not be able to close a quote or introduce shell metacharacters,
+  # and sed would silently mangle `&` and the `|` delimiter besides. Git refs,
+  # release tags, and commit shas all fit inside this class.
+  local label="$1" value="$2"
+  [ -n "$value" ] || die "${label} is empty"
+  case "$value" in
+    *[!A-Za-z0-9._/@+-]*) die "${label} contains characters that are unsafe to template into user-data: ${value}" ;;
+  esac
+  case "$value" in
+    -*) die "${label} must not start with a dash: ${value}" ;;
+  esac
+}
+
+require_positive_int() {
+  local label="$1" value="$2"
+  case "$value" in
+    ''|*[!0-9]*) die "${label} must be a positive integer, got: ${value}" ;;
+  esac
+  [ "$value" -gt 0 ] || die "${label} must be greater than zero"
+}
+
+find_testbed_vpc() {
+  # Echoes the testbed VPC id, or nothing when it does not exist yet. Three
+  # independent predicates, not one applied twice: the project tag, an exact
+  # match on the dedicated CIDR (EC2's `cidr` filter is an exact primary-CIDR
+  # match), and IsDefault=false, because production lives in the default VPC.
+  # Anything other than exactly one match is an error rather than a pick.
+  local rows count vpc_id cidr is_default
+  rows=$(aws ec2 describe-vpcs \
     --filters "Name=tag:${TAG_KEY},Values=${TAG_VALUE}" \
-    --query 'Vpcs[0].VpcId' --output text)
-  [ "$vpc_id" != "None" ] && [ -n "$vpc_id" ] || die "no testbed VPC found; run provision.sh first"
+              "Name=cidr,Values=${VPC_CIDR}" \
+              "Name=is-default,Values=false" \
+    --query 'Vpcs[].[VpcId,CidrBlock,IsDefault]' --output text)
+
+  rows=$(printf '%s\n' "$rows" | grep -v '^[[:space:]]*$' || true)
+  [ -n "$rows" ] || return 0
+
+  count=$(printf '%s\n' "$rows" | wc -l | tr -d '[:space:]')
+  [ "$count" = "1" ] \
+    || die "expected exactly one VPC tagged ${TAG_KEY}=${TAG_VALUE} with CIDR ${VPC_CIDR}, found ${count}; resolve by hand before running anything destructive"
+
+  read -r vpc_id cidr is_default <<<"$rows"
+
+  # Re-assert on the returned attributes rather than trusting the filters alone.
+  case "$vpc_id" in vpc-*) ;; *) die "unexpected VPC id: ${vpc_id}" ;; esac
+  [ "$cidr" = "$VPC_CIDR" ] || die "VPC ${vpc_id} has CIDR ${cidr}, expected ${VPC_CIDR}"
+  case "$is_default" in
+    [Ff]alse) ;;
+    *) die "VPC ${vpc_id} reports IsDefault=${is_default}; the default VPC holds production and is never a teardown target" ;;
+  esac
+
   echo "$vpc_id"
 }
--- a/scripts/fleet-testbed/lib.sh
+++ b/scripts/fleet-testbed/lib.sh
@@ -5,7 +5,7 @@
 # Every resource this lane creates therefore carries TAG_KEY=TAG_VALUE and lives
 # inside a dedicated non-default VPC whose CIDR nothing else uses. Selection for
 # any destructive operation asserts all three: the tag, the CIDR, and
-# IsDefault=false. See find_testbed_vpc.
+# IsDefault=false. See resolve_testbed_vpc.
 
 # Every value defined below is consumed by the scripts that source this file,
 # which a per-file unused-variable check cannot see.
@@ -108,18 +108,34 @@ require_positive_int() {
   [ "$value" -gt 0 ] || die "${label} must be greater than zero"
 }
 
-find_testbed_vpc() {
-  # Echoes the testbed VPC id, or nothing when it does not exist yet. Three
-  # independent predicates, not one applied twice: the project tag, an exact
-  # match on the dedicated CIDR (EC2's `cidr` filter is an exact primary-CIDR
-  # match), and IsDefault=false, because production lives in the default VPC.
-  # Anything other than exactly one match is an error rather than a pick.
+# Result of resolve_testbed_vpc. Empty means "does not exist", and it is only
+# ever meaningful immediately after a successful call.
+TESTBED_VPC_ID=""
+
+resolve_testbed_vpc() {
+  # Sets TESTBED_VPC_ID to the testbed VPC id, or to "" when it does not exist
+  # yet. Deliberately NOT a function whose value is captured with $(...): bash
+  # unsets errexit inside a command substitution unless inherit_errexit is on,
+  # and inherit_errexit does not exist at all in the bash 3.2 that macOS ships.
+  # Captured that way, a throttled or denied describe-vpcs would return an empty
+  # string with status 0 and every caller would read it as "no testbed VPC" --
+  # which silently turns `teardown --network` into a no-op that leaves the
+  # instance and the token alive, and turns provision into a second VPC at the
+  # same CIDR. Running in the caller's shell keeps errexit and `die` real.
+  #
+  # Three independent predicates, not one applied twice: the project tag, an
+  # exact match on the dedicated CIDR (EC2's `cidr` filter is an exact
+  # primary-CIDR match), and IsDefault=false, because production lives in the
+  # default VPC. Anything other than exactly one match is an error, not a pick.
+  TESTBED_VPC_ID=""
+
   local rows count vpc_id cidr is_default
   rows=$(aws ec2 describe-vpcs \
     --filters "Name=tag:${TAG_KEY},Values=${TAG_VALUE}" \
               "Name=cidr,Values=${VPC_CIDR}" \
               "Name=is-default,Values=false" \
-    --query 'Vpcs[].[VpcId,CidrBlock,IsDefault]' --output text)
+    --query 'Vpcs[].[VpcId,CidrBlock,IsDefault]' --output text) \
+    || die "describe-vpcs failed while resolving the testbed VPC; refusing to guess whether it exists"
 
   rows=$(printf '%s\n' "$rows" | grep -v '^[[:space:]]*$' || true)
   [ -n "$rows" ] || return 0
@@ -138,5 +154,65 @@ find_testbed_vpc() {
     *) die "VPC ${vpc_id} reports IsDefault=${is_default}; the default VPC holds production and is never a teardown target" ;;
   esac
 
-  echo "$vpc_id"
+  TESTBED_VPC_ID="$vpc_id"
+}
+
+require_ipv4() {
+  # $1 = label, $2 = value. Rejects anything that is not four plain decimal
+  # octets, including the shapes a dot-count check waves through: `1.2.3.`,
+  # `1.2.3.4.5`, `00.0.0.0`, and `999.0.0.1`. An SSH allowlist entry is not a
+  # place to accept a string AWS might interpret differently than this script.
+  local label="$1" value="$2"
+  local o1 o2 o3 o4 rest octet
+  IFS=. read -r o1 o2 o3 o4 rest <<<"$value"
+  [ -z "${rest:-}" ] || die "${label} is not a dotted-quad IPv4: ${value}"
+  for octet in "$o1" "$o2" "$o3" "$o4"; do
+    case "$octet" in
+      ''|*[!0-9]*) die "${label} is not a dotted-quad IPv4: ${value}" ;;
+      0) ;;
+      0*) die "${label} has a leading-zero octet, which is ambiguous: ${value}" ;;
+    esac
+    [ "${#octet}" -le 3 ] && [ "$octet" -le 255 ] \
+      || die "${label} has an out-of-range octet: ${value}"
+  done
+  [ "$value" != "0.0.0.0" ] || die "${label} is 0.0.0.0; refusing to use it in a rule"
+}
+
+read_github_token() {
+  # Prints the GitHub token on stdout, with no trailing newline, from whichever
+  # source the operator configured. Call it with a redirect, never as $(...):
+  # inside a command substitution `die` would only kill the subshell and the
+  # caller would proceed with an empty token. See resolve_testbed_vpc.
+  #
+  # Deliberately pluggable: the default is the
+  # personal `gh` credential, which is broader than this box needs, so a
+  # narrower fine-grained PAT can be supplied without editing the scripts.
+  #
+  #   FLEET_TESTBED_TOKEN_FILE=/path/to/pat     read a token from a file
+  #   FLEET_TESTBED_TOKEN_COMMAND='op read ...' run a command that prints one
+  #   (neither set)                             `gh auth token`
+  local token=""
+  if [ -n "${FLEET_TESTBED_TOKEN_FILE:-}" ]; then
+    [ -r "$FLEET_TESTBED_TOKEN_FILE" ] \
+      || die "FLEET_TESTBED_TOKEN_FILE is not readable: ${FLEET_TESTBED_TOKEN_FILE}"
+    token=$(cat "$FLEET_TESTBED_TOKEN_FILE") \
+      || die "could not read FLEET_TESTBED_TOKEN_FILE"
+  elif [ -n "${FLEET_TESTBED_TOKEN_COMMAND:-}" ]; then
+    token=$(eval "$FLEET_TESTBED_TOKEN_COMMAND") \
+      || die "FLEET_TESTBED_TOKEN_COMMAND failed"
+  else
+    token=$(gh auth token) \
+      || die "gh auth token failed; run 'gh auth login', or set FLEET_TESTBED_TOKEN_FILE to a scoped PAT"
+  fi
+
+  # Trim surrounding whitespace, so the SecureString holds exactly the token and
+  # not the trailing newline `file://` would otherwise carry into it.
+  token="${token#"${token%%[![:space:]]*}"}"
+  token="${token%"${token##*[![:space:]]}"}"
+
+  [ -n "$token" ] || die "the configured GitHub token source produced an empty token"
+  case "$token" in
+    *[[:space:]]*) die "the configured GitHub token source produced whitespace inside the token" ;;
+  esac
+  printf '%s' "$token"
 }
--- a/scripts/fleet-testbed/provision.sh
+++ b/scripts/fleet-testbed/provision.sh
@@ -5,9 +5,12 @@
 #   ./provision.sh [--arch x86_64|arm64] [--type <instance-type>] [--disk 200]
 #                  [--ref <git-ref>] [--with-rust] [--no-wait]
 #
-# Idempotent: network, key pair, and IAM role are created once and reused.
-# Everything is tagged and lives in a dedicated VPC so teardown can require two
-# independent predicates before terminating anything.
+# Idempotent: network, key pair, IAM role, and the SSM token are created or
+# refreshed on every run and reused. Everything is tagged and lives in a
+# dedicated non-default VPC, so teardown can assert tag, CIDR, and non-default
+# together before deleting anything in an account that also holds production.
+
+set -euo pipefail
 
 # --arch selects the AMI, the awscli build, and the musl artifact, so it has to
 # be known before lib.sh derives them.
@@ -19,7 +22,7 @@ export ARCH="${ARCH:-x86_64}"
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 # shellcheck source=lib.sh
-source "${SCRIPT_DIR}/lib.sh"
+source "${SCRIPT_DIR}/lib.sh" || { echo "fatal: cannot source ${SCRIPT_DIR}/lib.sh" >&2; exit 1; }
 
 INSTANCE_TYPE="$DEFAULT_TYPE"
 DISK_GB="200"
@@ -29,9 +32,9 @@ WAIT_READY="1"
 
 while [ $# -gt 0 ]; do
   case "$1" in
-    --type) INSTANCE_TYPE="$2"; shift 2 ;;
-    --disk) DISK_GB="$2"; shift 2 ;;
-    --ref) REPO_REF="$2"; shift 2 ;;
+    --type) INSTANCE_TYPE="${2:-}"; shift 2 ;;
+    --disk) DISK_GB="${2:-}"; shift 2 ;;
+    --ref) REPO_REF="${2:-}"; shift 2 ;;
     --with-rust) WITH_RUST="1"; shift ;;
     --arch) shift 2 ;;
     --no-wait) WAIT_READY="0"; shift ;;
@@ -39,15 +42,36 @@ while [ $# -gt 0 ]; do
   esac
 done
 
+# Everything below is templated into a script that cloud-init runs as root, so
+# validate before substituting rather than hoping a ref never contains a quote.
+require_safe_ref "--ref" "$REPO_REF"
+require_safe_ref "--type" "$INSTANCE_TYPE"
+require_safe_ref "FIX_BRANCH" "$FIX_BRANCH"
+require_safe_ref "RELEASE_TAG" "$RELEASE_TAG"
+require_safe_ref "SSM_PREFIX" "$SSM_PREFIX"
+require_safe_ref "AWS_DEFAULT_REGION" "$AWS_DEFAULT_REGION"
+require_safe_ref "AWSCLI_ARCH" "$AWSCLI_ARCH"
+require_safe_ref "MUSL_ARCH" "$MUSL_ARCH"
+require_positive_int "--disk" "$DISK_GB"
+require_positive_int "DEADMAN_HOURS" "$DEADMAN_HOURS"
+
+# The whole safety design assumes production shares this account.
+require_expected_account
+log "account ${ACCOUNT_ID}, region ${AWS_DEFAULT_REGION}"
+
 MY_IP="$(curl -fsS --max-time 10 https://checkip.amazonaws.com | tr -d '[:space:]')"
-[ -n "$MY_IP" ] || die "could not determine public IP for the SSH allowlist"
+case "$MY_IP" in
+  *[!0-9.]*|'') die "could not determine a public IPv4 for the SSH allowlist (got '${MY_IP}')" ;;
+esac
+[ "$(printf '%s' "$MY_IP" | tr -cd '.' | wc -c | tr -d '[:space:]')" = "3" ] \
+  || die "public IP is not dotted-quad IPv4: ${MY_IP}"
+[ "$MY_IP" != "0.0.0.0" ] || die "refusing to allowlist 0.0.0.0"
 
 # --- network -----------------------------------------------------------------
 
-VPC_ID=$(aws ec2 describe-vpcs --filters "Name=tag:${TAG_KEY},Values=${TAG_VALUE}" \
-  --query 'Vpcs[0].VpcId' --output text)
+VPC_ID="$(find_testbed_vpc)"
 
-if [ "$VPC_ID" = "None" ] || [ -z "$VPC_ID" ]; then
+if [ -z "$VPC_ID" ]; then
   log "creating dedicated VPC ${VPC_CIDR}"
   VPC_ID=$(aws ec2 create-vpc --cidr-block "$VPC_CIDR" \
     --tag-specifications "$(tag_spec vpc)" --query 'Vpc.VpcId' --output text)
@@ -69,8 +93,11 @@ if [ "$VPC_ID" = "None" ] || [ -z "$VPC_ID" ]; then
   aws ec2 associate-route-table --route-table-id "$RT_ID" --subnet-id "$SUBNET_ID" >/dev/null
 else
   log "reusing VPC ${VPC_ID}"
-  SUBNET_ID=$(aws ec2 describe-subnets --filters "Name=vpc-id,Values=${VPC_ID}" \
+  SUBNET_ID=$(aws ec2 describe-subnets \
+    --filters "Name=vpc-id,Values=${VPC_ID}" "Name=tag:${TAG_KEY},Values=${TAG_VALUE}" \
     --query 'Subnets[0].SubnetId' --output text)
+  [ -n "$SUBNET_ID" ] && [ "$SUBNET_ID" != "None" ] \
+    || die "VPC ${VPC_ID} has no tagged testbed subnet; delete it with teardown.sh --network and re-provision"
 fi
 
 SG_ID=$(aws ec2 describe-security-groups \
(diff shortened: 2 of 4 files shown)


pablonyx:
## Verdict

**No.** Not safe to merge as it stands: the default provisioning path cannot work at all (`make` is gone), and a failed `describe-vpcs` is silently read as "no testbed VPC", which makes `teardown --all` delete your only way back into a still-running box while reporting success.

The credential work is real, though. I traced the token through every path I could find and it is genuinely absent from `user-data.sh`, the askPass helper is correct, and the `.git/config` assertion is not a no-op. The remaining objection there is about what the mitigation actually buys, not about whether the code does what it says.

## BLOCKING

### B1. `make` is not installed on the default path, so `make build` and `make setup` both fail

`user-data.sh:143-151` gates `build-essential pkg-config libssl-dev` behind `--with-rust`, and `user-data.sh:263` and `:268` then run `make build` and `make setup PROFILE=fleet-1`.

`make` is not in the Ubuntu 24.04 cloud image. I pulled both manifests and grepped them:

```
$ curl -fsSL https://cloud-images.ubuntu.com/releases/noble/release/ubuntu-24.04-server-cloudimg-amd64.manifest | wc -l
664
$ grep -cE '^make[[:space:]]' ...amd64.manifest   -> 0
$ grep -cE '^make[[:space:]]' ...arm64.manifest   -> 0
$ grep -iE 'make|gcc|binutils|libc-dev' ...amd64.manifest
gcc-14-base:amd64   libc-dev-bin   libc-devtools   libclang-cpp18   libgcc-s1   linux-libc-dev
```

`build-essential` was what dragged `make` in. Removing it removes the build tool the script itself invokes. `cron`, `python3`, `git`, `curl` and `jq` are in the base image; `make`, `python3-venv` and `unzip` are not, and the apt list covers the last two but not `make`.

Failure shape: roughly twenty minutes of successful setup, then `make: command not found` at the second-to-last step, no `/var/lib/fleet-testbed-ready`, and `provision.sh:288-303` blocks for a full sixty minutes before dying with a generic timeout. Exactly the annoying-to-debug case.

The dependency reasoning behind the gate is correct and I verified it independently, so the gate is a good idea; it just needs `make` moved into the base list. See "what I verified" below for the pnpm and uv checks.

### B2. A failing `describe-vpcs` is swallowed, and `find_testbed_vpc` reports "no VPC"

`lib.sh:118`, reached via `provision.sh:72` and `teardown.sh:62`.

`VPC_ID="$(find_testbed_vpc)"` runs the function inside a command-substitution subshell, and bash unsets `errexit` in command substitutions unless `shopt -s inherit_errexit` is on. It defaults off outside POSIX mode (bash manual: "instead of unsetting it in the subshell environment"). So when `rows=$(aws ec2 describe-vpcs ...)` fails, the function does not abort. `rows` is empty, `[ -n "$rows" ] || return 0` at `lib.sh:125` returns success, and the caller gets an empty string meaning "the testbed VPC does not exist".

I established this with a stubbed `aws` that exits 255 on `describe-vpcs`. Minimal repro of the semantics:

```
$ bash -c 'set -e; f(){ local r; r=$(false); echo CONTINUED; }; f; echo after'
(exits, nothing printed)
$ bash -c 'set -e; f(){ local r; r=$(false); echo CONTINUED; }; v=$(f); echo "after v=$v"'
after v=CONTINUED
```

Three consequences, all observed against the stub:

**`teardown.sh --network`** prints "no testbed VPC ... skipping instances and network", then "nothing to do", and exits 0. The instance keeps running and the token stays in SSM.

**`teardown.sh --all`** is worse. It skips instance termination, then goes on to delete the inline policy, the role, the instance profile, the AWS key pair and the local private key, and prints "teardown complete":

```
Throttling
[fleet-testbed] no testbed VPC (...); skipping instances and network
[fleet-testbed] deleting SSM parameter /proliferate-fleet-testbed/github-token
[fleet-testbed] deleting IAM role, instance profile, and key pair
[fleet-testbed] removing local private key .../proliferate-fleet-testbed.pem
[fleet-testbed] teardown complete
```

The still-running m7i.4xlarge loses SSM and Bedrock access mid-experiment and becomes permanently unreachable, because both halves of the key pair are gone. In real AWS `delete-instance-profile` would hit `DeleteConflict` and set `FAILURES`, but `fail()` only counts, so `delete-role-policy`, `delete-key-pair` and the local `rm` at `:181-195` all still run before the `die` at `:200`.

**`provision.sh`** takes the create branch and makes a *second* VPC at `10.42.0.0/16` carrying the same tag. Observed:

```
[fleet-testbed] account 157466816238, region us-east-1
Throttling: Rate exceeded
[fleet-testbed] creating dedicated VPC 10.42.0.0/16
AWS: ec2 create-vpc --cidr-block
```

After that every `find_testbed_vpc` dies with "found 2", and both provision and teardown are wedged until someone fixes it by hand in the console. The safety property holds (nothing in production is at risk) but the lane bricks itself on a throttle.

Fix is one line: `shopt -s inherit_errexit` in `lib.sh`, or make the describe explicit with `rows=$(aws ec2 describe-vpcs ...) || die "describe-vpcs failed"`. Worth noting the comment at `teardown.sh:99-100` is accurate for the four network describes, which are top-level assignments and genuinely are protected. The hole is specific to the describe that lives inside a function.

### B3. `--network` does not delete the SSM token when the VPC is already gone

`teardown.sh:64-69` versus `:151-162`. The comment at `:151-153` says the SSM delete is "outside the VPC guard on purpose", but with `DELETE_IDENTITY=0` the early `exit 0` at `:68` fires first and the SSM block is never reached. Observed with a stub returning an empty `describe-vpcs`:

```
[fleet-testbed] no testbed VPC (...); skipping instances and network
[fleet-testbed] nothing to do
aws calls: sts get-caller-identity, ec2 describe-vpcs      <- no ssm delete-parameter
```

Reachable state: the deadman terminated the instance and an earlier `--network` (or a console cleanup) already removed the VPC, then the operator runs `teardown.sh --network` to clear up. It reports "nothing to do" and a live `repo`-scoped GitHub token stays in SSM in the production account indefinitely, which is the failure mode S5 claims to close. Only `--all` currently reaches the delete. Move the SSM block above the early exit, or drop the early exit.

### B4. The cherry-pick becomes a hard failure, not a no-op, the moment #2134 merges

`user-data.sh:247-251`. When `perf/dev-loop-launch` is merged, `git rev-list --reverse origin/${FIX_BRANCH} --not origin/main` is empty and `git cherry-pick` with no arguments exits 129, which I measured:

```
$ git cherry-pick >/dev/null 2>&1; echo $?
129
$ git cherry-pick $(git rev-list --reverse HEAD --not HEAD) >/dev/null 2>&1; echo $?
129
```

Under `set -e` that aborts user-data before the runtime download, the build and the ready marker, so provisioning hangs the caller for sixty minutes and dies. If the branch is deleted on merge, `git fetch origin ${FIX_BRANCH} main` at `:249` fails one line earlier. #2134 is currently OPEN, so this is armed and will fire on merge. The body says `FIX_BRANCH` "should become a no-op once it merges", which is not what the code does. Guard it: resolve the list into a variable and skip both the fetch and the cherry-pick when it is empty.

### B5. The token exposure on the box is unchanged, and the body reads as though it is not

`gh auth token` on this machine is Pablo's personal token with scopes `gist, read:org, repo, workflow` (from `gh auth status`, redacted). `repo` is read/write to every private repo he can reach; `workflow` can rewrite `.github/workflows`, which is a path into the deploy pipeline.

The S1 fix moves that token from `.git/config`, readable by `ubuntu`, into SSM, readable by `ubuntu` via the instance role. `HttpTokens=required` costs a local process one unauthenticated PUT to 169.254.169.254, and `HttpPutResponseHopLimit=1` restricts extra network hops, not host processes. So any process running as `ubuntu` can still mint the token, and the entire purpose of this box is to run five concurrent agent profiles executing model-generated code as `ubuntu`. The mode-700 helper is not a boundary against the adversary that actually matters here.

That is not an argument against the change: it is strictly better than a token in a 0644 `.git/config` and in xtrace. But the body currently presents the helper as *the* mitigation, and the last round of this PR was specifically about a body claim that overstated the credential position. Either scope the credential down to a fine-grained PAT with `contents:read` on the one repo, or state plainly in the body that anything running on the testbed can obtain an org-wide `repo`+`workflow` token. I would prefer the former; the clone only needs read.

## NON-BLOCKING

**N1. Bedrock quota is not isolated from production, and the body does not say so.** Scoping `Resource` to `anthropic.claude-*` limits which models the testbed can reach; it does nothing about throughput, because Bedrock TPM/RPM quotas are per-account per-model. Five concurrent profiles hammering Claude in 157466816238 can throttle production agent auth. The S10 row reads as though the narrowing buys isolation. Add a sentence.

**N2. The placeholder guard covers four of nine placeholders.** `provision.sh:249-252` checks `__REPO_REF__`, `__FIX_BRANCH__`, `__SSM_PREFIX__`, `__AWS_REGION__`. `__DEADMAN_HOURS__`, `__WITH_RUST__`, `__AWSCLI_ARCH__`, `__MUSL_ARCH__` and `__RELEASE_TAG__` are unchecked. A single `case "$USER_DATA" in *__*)` would cover all of them.

**N3. The port-22 revoke misses range rules, and can trip on `null`.** `provision.sh:117-119` matches only `FromPort==22 && ToPort==22`; a rule with a range that contains 22 is not revoked and still counts toward the 60-rule quota. Separately, if `SecurityGroups[0]` does not resolve, the JMESPath yields `null`, which is not `[]`, so `:122` calls `revoke-security-group-ingress --ip-permissions null` and dies with an opaque AWS error.

**N4. IP validation lets two malformed values through.** `provision.sh:63-68` accepts `00.0.0.0` (defeats the `0.0.0.0` literal comparison) and `1.2.3.` (three dots, digits only). Neither can widen the rule, because `/32` is hardcoded, and AWS rejects both, so this is cosmetic. A single `case` on `+([0-9]).+([0-9]).+([0-9]).+([0-9])` or a regex would close it.

**N5. The `.git/config` assertion misses the userinfo-only form.** `user-data.sh:240`. Negative control:

```
https://x-access-token:gho_ABC@github.com/o/r.git   -> DETECTED
https://github.com/o/r.git                          -> clean
https://gho_ABC@github.com/o/r.git                  -> NOT DETECTED
```

Git persists the `user:pass@` form so the current regex covers the realistic case, but `://[^/@[:space:]]+@` would cover both.

**N6. "chmod 600 before the first byte is written" is true of one log and false of the other.** `user-data.sh:18` creates the setup log at 600 correctly. `/var/log/cloud-init-output.log` has been appended to by cloud-init since early boot, long before the `chmod` at `:19`. The chmod is still effective (it changes the inode, and cloud-init's `tee` keeps writing through its existing fd), and it does not matter now that user-data is tokenless, but the claim as phrased is not accurate.

**N7. The SSM value carries a trailing newline.** `file://` is honoured by AWS CLI v2 (I read `awscli/paramfile.py`: `get_file` does a plain `f.read()`), and `gh auth token` emits a trailing newline, so the SecureString is `gho_...\n`. Harmless today, because git truncates askpass output at the first newline, which I confirmed against a fake helper that deliberately emitted `token\n\n`:

```
PROMPT-SEEN: [Username for 'https://github.com': ]
PROMPT-SEEN: [Password for 'https://x-access-token@github.com': ]
username=x-access-token
password=gho_SECRET_TOKEN
```

Any future consumer that uses the value verbatim will get the newline. `printf '%s' "$(gh auth token)" > "$TOKEN_FILE"` removes the trap.

**N8. Repeatability is narrower than advertised.** The commit and the runtime binary are pinned, but `@anthropic-ai/claude-code` (`:185`), the `uv` installer (`:171`), nodesource `setup_22.x` (`:166`) and the AWS CLI installer (`:174`) all float. `provision.sh:297` prints the Claude CLI version as evidence the box matches its pin, which it is not.

**N9. `cron` is present in the base image but is not in the apt list.** I confirmed `cron` is in the 24.04 manifest, so the cron.d half of the deadman works today. It is a free line of insurance to add it to `apt_install_retry` next to `git curl jq`.

**N10. Provision and teardown disagree on how they resolve the subnet and the SG.** `provision.sh:96-98` takes `Subnets[0]` with no count assertion; `:103-105` resolves the SG by name with no tag filter, while `teardown.sh:101-103` requires the tag. Both are scoped inside the testbed VPC so neither is a production risk, but an untagged SG in that VPC would be reused by provision and then not deleted by teardown, and `delete-vpc` would fail with a dependency error the operator has to chase.

**N11. The `FLEET_TESTBED_ACCOUNT_ID` override is the right tradeoff, keep it.** The guard exists to catch a run in the wrong account with the wrong mental model, and an env var is a deliberate act rather than an accident. Crucially it does not weaken the thing that actually protects production: `VPC_CIDR`, `TAG_KEY`, `TAG_VALUE` and `NAME_PREFIX` are unconditional literal assignments at `lib.sh:18-23`, so they are reassigned on every source and a hostile environment cannot move the selector.

## Finding-by-finding

| # | Claim | Verdict | Note |
| --- | --- | --- | --- |
| 1 | Token absent from user-data; askPass from SSM; logs 600; clone assertion | **partial** | The literal claim holds: zero token references in the file, helper emits correctly and is syntactically valid under dash, assertion is not a no-op. Partial for N5, N6, and B5 (exposure to any local process is unchanged) |
| 2 | `find_testbed_vpc` on tag + exact CIDR + non-default, re-asserted, exactly one | **partial** | Selection logic is sound. I could not construct any input, unset variable or response shape that selects the default VPC. But B2: a failed describe is read as "no VPC" |
| 3 | Tag filter on all four `--network` describes | **confirmed** | SG, subnet, route-table and IGW describes at `teardown.sh:101-112` all carry `Name=tag:${TAG_KEY}` |
| 4 | Six `\|\| true`s removed; failure counter; `with_retry` for ENI drain | **confirmed** | `af0fcc431e` teardown had 6, `f0e381d657` has 0. `fail()`/`FAILURES`/`die` verified end to end. `with_retry` is 5 silent plus 1 loud, matching its "attempt N/6" wording |
| 5 | `--network` deletes SSM param; `--all` deletes policy, role, profile, key pair, local key | **partial** | `--all` verified against a call-logging stub. `--network` alone misses the SSM param when the VPC is gone (B3) |
| 6 | Stale port-22 rules revoked; `MY_IP` validated as dotted quad, not `0.0.0.0` | **partial** | Revoke and ordering are right. N3 (range rules, `null`) and N4 (`00.0.0.0`, `1.2.3.`) |
| 7 | Key pair gated both sides; `umask 077` into temp then `mv` | **confirmed** | Both mismatch branches die with the specific remedy; the redirect happens inside the `umask 077` subshell so the file is 600 from creation; `mv` only after a non-empty check; `trap` cleans the temp |
| 8 | `put-role-policy` and `add-role-to-instance-profile` unconditional | **confirmed** | Both outside the `get-role` gate; the `add-role` failure path verifies the actual attachment before treating it as benign |
| 9 | `HttpTokens=required,HttpPutResponseHopLimit=1` | **confirmed** | `provision.sh:263` |
| 10 | Bedrock scoped to Claude FM plus inference-profile glob; `ec2:TerminateInstances` deleted | **confirmed** | Policy parses as valid JSON with the three expected statements; zero `TerminateInstances` occurrences in any of the four files. Body gap N1 |
| 11 | `require_expected_account`, default 157466816238, env-overridable | **confirmed** | Called first by both entry points; `account=$(...) \|\| die` is explicitly guarded. Override is correct, see N11 |
| 12 | `require_safe_ref` restricts to `[A-Za-z0-9._/@+-]`, rejects leading dash | **confirmed** | I could not defeat it. The class excludes every sed replacement metacharacter (`\`, `&`, the `\|` delimiter, newline) and every character that could break out of the double-quoted assignment in user-data or the `su -c` re-parse (`"`, `$`, backtick, `\`, space, `;`, `&`, `\|`). The leading-dash rule blocks option injection into `git checkout` / `git fetch`. `+abc` is allowed and reaches `git fetch` as a force refspec, which is harmless on a throwaway clone |
| 13 | Both entry points `set -euo pipefail` before sourcing, explicit source failure | **confirmed** | `provision.sh:13`/`:25`, `teardown.sh:18`/`:22` |
| 14 | Persistent systemd timer plus cron.d, fails closed, `shutdown` fallback | **confirmed** | Parse logic is correct against the exact write: `echo "$DEADLINE_EPOCH" > file` then `deadline="$(cat file)"` strips the newline, so `''\|*[!0-9]*` only matches genuine corruption. Timer survives reboot via `WantedBy=timers.target` and re-arms on `OnBootSec=1min`. Timer and cron cannot fight: both exec the same idempotent checker whose only action is `shutdown -h now`, and a second `shutdown` against an already-shutting-down box is a no-op. `arm_deadman` inside `if` correctly suspends `errexit` for the whole body so a partial arm still degrades to the one-shot fallback |

## What I verified, and how

- `bash -n` clean on all four files, and on `user-data.sh` after rendering it through the exact nine-substitution `sed` pipeline from `provision.sh:237-247`. No placeholders survive.
- `shellcheck -x` (0.11.0): zero findings at `-S warning`, exactly three at `-S info`, all SC2016 on JMESPath and remote-command single quotes (`provision.sh:118`, `provision.sh:294`, `teardown.sh:109`). `-S style` adds nothing. The author's count is exact.
- Extracted the askPass helper by executing its heredoc with the templated values, then `sh -n` and `dash -n` on the result. Its argv contains no secret: `aws ssm get-parameter --with-decryption --region us-east-1 --name /proliferate-fleet-testbed/github-token --query Parameter.Value --output text`.
- Confirmed the askPass contract offline with `git credential fill` and a fake helper: git issues `Username for '...'` and `Password for '...'`, the `Username*` glob discriminates correctly, and git truncates at the first newline.
- Negative control on the `.git/config` assertion (results in N5), so it is a real check, not decoration.
- Parsed the rendered inline IAM policy with `json.load`. Three statements, SSM ARN renders as `arn:aws:ssm:us-east-1:157466816238:parameter/proliferate-fleet-testbed/*`, Bedrock resources as expected.
- Offline harness with a stubbed `aws` against `find_testbed_vpc`, eighteen cases chosen to be different from the author's: empty output, blank lines only, CRLF line endings, bare `None`, three-field `None`, tab-separated empty fields, leading whitespace, lowercase `false`, numeric `0` for IsDefault, a `default` string where a VPC id belongs, wrong CIDR, trailing space inside the CIDR, two rows, two rows where one is the real default VPC, a describe that exits 255 with stderr, and a describe that exits 254 silently. Everything fails closed except the last two, which are B2. CRLF, `None`, lowercase-`false`-with-`\r`, and `0` all die rather than resolve.
- Separate call-logging stubs for full `teardown.sh` runs under three VPC states and for a full `provision.sh` run under a throttled describe, which is how B2 and B3 were established.
- The `--with-rust` reasoning, independently: `pnpm-workspace.yaml` `onlyBuiltDependencies` is closed at five entries (`@sentry/cli`, `core-js`, `esbuild`, `protobufjs`, `sharp`), the lockfile has zero `requiresBuild` entries and no node-gyp packages (`fsevents` is darwin-only, `sharp@0.34.5` carries `@img/sharp-linux-x64` and `@img/sharp-linux-arm64` prebuilts as optional deps). `server/uv.lock`: 103 packages, the only one without wheels is the editable local project, and every other package either ships a `py3-none-any` wheel or has manylinux wheels for both `x86_64` and `aarch64`. Ubuntu 24.04 has glibc 2.39 so the manylinux tags resolve. **The reasoning is correct**, which is why B1 is worth stating precisely: the gate is fine, it just also removed the build tool.
- `SKIP_RUST=1` is honoured by `Makefile:1538-1544` (`build-rust` short-circuits), so the no-cargo claim holds.
- `git cherry-pick` exit code with an empty commit list, measured (B4).
- AWS CLI v2 `file://` handling, by reading `awscli/paramfile.py` in the installed 2.34.53 rather than guessing.
- `gh auth status` locally for the token scopes in B5.

## What I could not verify without AWS mutations, which I did not run

I made no AWS calls at all in this review, not even read-only ones, and I did not execute `provision.sh` or `teardown.sh` against a real account with any flag.

- Whether `revoke-security-group-ingress --ip-permissions` accepts the exact JSON that `describe-security-groups` emits, including `UserIdGroupPairs`, `Ipv6Ranges` and `PrefixListIds`. It is the documented round-trip pattern, but I could not exercise it.
- Whether `create-key-pair --key-type ed25519 --tag-specifications` and `create-security-group --tag-specifications` behave as assumed.
- Whether `add-role-to-instance-profile` really returns `LimitExceeded` rather than something else when the role is already attached, and therefore whether the `:202-208` fallback triggers on the right condition.
- Whether the ordering SG then subnet then route table then IGW then VPC actually drains cleanly within `with_retry`'s five-attempt, fifty-second budget after a termination.
- Whether the scoped Bedrock policy is sufficient for what `CLAUDE_CODE_USE_BEDROCK=1` actually calls. `InvokeModel` and `InvokeModelWithResponseStream` are covered; if the runtime or the CLI ever calls `bedrock:Converse`, `bedrock:GetInferenceProfile` or `bedrock:ListFoundationModels`, it will be denied.
- Whether `delete-instance-profile` fails with `DeleteConflict` while an instance still references it, which is the partial safety net I assumed in B2.
- End-to-end: whether the askPass clone actually authenticates to a private repo from the instance role.

Nothing in the blocking list depends on any of those.

pablonyx:
## Round two: verdict

**Yes, this is safe for Pablo to merge.** All five blockers are fixed, I could not construct a new defect of equivalent severity, and the rewritten body no longer contains any claim the first review disproved.

Confirm-or-deny pass only, as instructed. No AWS calls of any kind, neither script executed against an account.

### The five blockers

| # | Verdict | How I established it |
| --- | --- | --- |
| B1 `make` | **fixed** | `make` is now unconditional in the apt list, compilers still gated. The added detail is correct: I re-pulled both manifests and `libc6-dev` and `patch` are present on both arches while `make` is absent on both, so the missing piece was specifically the build tool. Line counts match the body exactly (664 amd64, 663 arm64). The `command -v` loop covers all six binaries the script invokes by name and sits immediately after the apt step |
| B2 errexit | **fixed** | 18-case selector harness re-run against `resolve_testbed_vpc`: the two failure modes that previously returned empty with status 0 (`exit 255` with stderr, `exit 254` silent) now both die with "refusing to guess whether it exists". Every other case is unchanged. Full stubbed teardown with a failing describe now makes **zero** mutations, where the previous commit deleted the role, the key pair and the local private key and printed "teardown complete" |
| B3 SSM on `--network` | **fixed** | Stubbed matrix: `--network` with the VPC gone now calls `ssm delete-parameter`; no flags with the VPC gone still prints "nothing to do" and calls nothing. Both directions correct |
| B4 empty cherry-pick | **fixed** | Extracted the generated `fleet-pin-repo` from the rendered user-data, `bash -n` and shellcheck clean, then ran it against a real fixture repo. Branch not merged: cherry-picks both commits, rc=0. Branch merged: prints "nothing to cherry-pick", rc=0. See the one residual below |
| B5 token scope | **fixed, code half** | `read_github_token` exercised across seven paths: trailing newline trimmed (so the SecureString now holds exactly the token), unreadable file, failing command, empty output and embedded whitespace each die and stop the script, both override variables honoured |

### The bash 3.2 claim: correct, with one refinement worth keeping

Verified, and it is the most reusable thing here.

```
$ /usr/bin/env bash --version            -> 3.2.57(1)-release   (macOS /bin/bash)
$ bash -c 'shopt | grep -c inherit_errexit'                     -> 0
$ bash -c 'shopt -s inherit_errexit'     -> invalid shell option name, rc=1
```

The shebang on all four files is `#!/usr/bin/env bash`, so 3.2 is the real interpreter for the two host-side scripts. My suggested one-liner would have been wrong.

The refinement: bare `shopt -s inherit_errexit` under `set -e` is a loud abort, not a silent no-op. It only becomes silent when written with the defensive idiom people reflexively add, and then the bug survives intact:

```
$ bash -c 'set -e; shopt -s inherit_errexit 2>/dev/null || true;
           f(){ local r; r=$(false); echo "FUNC CONTINUED"; }; v=$(f); echo "v=[$v]"'
v=[FUNC CONTINUED]
```

So the accurate statement is: on 3.2 the option is absent, the naive form breaks the script and the guarded form silently leaves the hole. The chosen fix (a global plus an explicit `|| die`, called as a plain statement) is version-independent and does not depend on errexit at all, which is better than either.

**The sweep is complete.** I enumerated every function definition and every command substitution across the three host-side scripts. The only functions that can `die` are `resolve_testbed_vpc` and `read_github_token`, and neither is captured any more. `tag_spec` is used inside `$( )` but is a bare `echo` and cannot fail. I also confirmed the redirect form is genuinely subshell-free on 3.2: a function that returns 1 under a redirect trips errexit, and one that calls `exit 7` under a redirect exits the whole script with 7.

**Control flow in the selector reads clean.** One call site each, plain statement, value consumed on the very next line, `TESTBED_VPC_ID` reset both at file scope and at the top of the function so neither a stale value nor a pre-seeded environment variable can survive into a destructive path.

### The two new destructive behaviours

**Whole-ingress revoke: right trade, and not reachable against anything else.** `SG_ID` resolves only from `vpc-id = TESTBED_VPC_ID` plus `group-name = proliferate-fleet-testbed` exactly. For this to wipe rules it should not, you would need an SG with that exact name inside a non-default VPC whose primary CIDR is exactly `10.42.0.0/16` and which carries the project tag, which is the lane's own SG by construction. Production is in the default VPC and the resolver refuses it twice over. Egress is untouched, so there is no self-lockout. Given that my own finding named range rules and `IpProtocol: -1` as the shapes an exact-22 predicate misses, "provably complete, and a hand-added rule does not survive the next provision" beats "matches the shapes I thought of" for a security group that exists solely to carry this lane's SSH rule. Endorsed.

**Account-wide key-pair-in-use check: right call, and it cannot delete anything.** It is a read-only describe whose only effect is to refuse. Widening the scope makes it strictly more conservative, and by `--all` the VPC may genuinely be gone, so scoping it to the VPC would defeat the point. Worst case is an unrelated instance in the account carrying a key literally named `proliferate-fleet-testbed`, which blocks `--all` with the offending instance id printed. Verified working: with a live instance holding the key, teardown refuses, leaves both halves of the key pair intact, and exits non-zero. That is precisely the unrecoverable case from B2, now closed.

### Non-blocking

1. **`git fetch origin ${FIX_BRANCH} main` still hard-fails if the branch is deleted.** The empty-`picks` case is handled, but the fetch is one line earlier: against a fixture with the branch merged and then deleted on the remote, `fatal: couldn't find remote ref perf/dev-loop-launch`, rc=128. Not automatic here, `repos/proliferate-ai/proliferate` has `delete_branch_on_merge: false`, so it needs a manual delete and the error is legible. Separately, this repo squash-merges, so the original shas never become ancestors of `main` and the empty case likely never fires at all; the pin keeps cherry-picking onto the release commit, which still produces the intended tree. Both benign, but the described lifecycle still is not quite what happens. `git ls-remote --exit-code origin "$FIX_BRANCH"` guarding the whole block would close it.
2. **"Non-blocking items from the same round, all taken" is an overstatement: seven of eleven.** Untouched: the unpinned `@anthropic-ai/claude-code`, `uv` and nodesource installers under a Repeatability heading (`provision.sh` prints the agent CLI version as proof the box matches its pin); `cron` still absent from the apt list though present in the base image; `Subnets[0]` still has no count assertion and the SG describe still lacks the tag filter teardown requires; and the `null` half of the ingress finding, where an unresolved `SecurityGroups[0]` yields `null`, which is not `[]`, so the revoke runs with `--ip-permissions null` and dies on an opaque AWS error. All four are genuinely minor and none of them blocks. Flagging it because a summary sentence claiming more than the code does is the specific thing this PR has now been corrected on three times, and that is the pattern worth breaking rather than the individual items.
3. **`require_ipv4` accepts an embedded newline.** `IFS=. read` consumes only the first line, so `$'1.2.3.4\nevil'` validates while the caller's variable keeps the tail. Unreachable from the only call site, which pipes through `tr -d '[:space:]'`, and `MY_IP` is only ever a single quoted argv element, so it cannot inject an argument. Worth a line only because the function now reads as a general validator. Everything else I threw at it is correct: `1.2.3.`, `1.2.3.4.5`, `00.0.0.0`, `01.2.3.4`, `1.2.3.256`, `999.999.999.999`, a trailing space, IPv6 and empty are all refused; `255.255.255.255` and `0.0.0.1` pass.
4. **`--all` deletes the inline policy, role and instance profile before reaching the key-pair guard.** Confirmed in the stub transcript: with an instance still holding the key, the IAM teardown had already run by the time the refusal fired, so a surviving box loses SSM and Bedrock mid-experiment. The unrecoverable resource is protected and IAM is recreated by the next provision, so this is a nuisance rather than a defect. Extending the same in-use check over the IAM block would close it.

### Body accuracy

I checked every claim my first review disproved. None survives.

- The two-log claim is replaced by an accurate narrower one, in the body and in the code comment.
- "should become a no-op once it merges" is replaced by an accurate description that names the 129 exit explicitly.
- S5's SSM claim and S3's "a failing describe aborts" are now true of the code.
- The Bedrock quota point is stated plainly and cross-referenced from S10.
- The IMDSv2 bullet dropped its "not reachable by an SSRF or by a container" over-claim.

The Residual risk section is **honest, not soft-pedalled**. It says outright that this is "not confinement", names the exact scopes, states that hop limit 1 "restricts other hosts, not other processes on this one", names what the box is for, and ends with "Treat any box provisioned before then as holding a credential with write access to the org." That is blunter than my own finding was. Item 2 above is the only overstatement I found anywhere in the rewrite.

### Mechanical re-checks

`bash -n` clean on all four files and on the rendered user-data. `shellcheck -x` clean at warning, exactly two SC2016 infos, matching the body. Placeholder guard verified both ways: it finds all nine placeholders in the template and produces zero false positives on the rendered output. The tightened `.git/config` regex catches both `user:secret@` and bare `token@` and does not false-positive on a real clone's config; the on-disk token grep catches all five `gh*_` prefixes and `github_pat_`. Full stubbed provision happy path runs to launch with no `create-vpc` call.

Still not verified, unchanged from round one and none of it load-bearing for this verdict: real AWS response shapes for the ingress JSON round-trip, `add-role-to-instance-profile`'s exact error on an already-attached role, ENI drain timing inside `with_retry`'s budget, and whether the scoped Bedrock policy covers everything `CLAUDE_CODE_USE_BEDROCK=1` calls beyond `InvokeModel` and `InvokeModelWithResponseStream`.

pablonyx:
## Verdict: rebase re-verification

**Safe to merge as-is.** This is a clean rebase of already-approved content; nothing new to evaluate.

### 1. Rebase integrity — verified by execution

```
git range-diff ac4c4caf1d..2e962e185c 956d2426b7..44ebebaeab
1:  af0fcc431e = 1:  f5dc5ac200 feat(dev): disposable EC2 testbed for multi-profile measurement
2:  f0e381d657 = 2:  f056d196b1 fix(dev): close credential and selection defects in the fleet testbed
3:  2e962e185c = 3:  44ebebaeab fix(dev): repair the regressions and gaps the second testbed review found
```

All three commits show `=`: identical patch content, same order, nothing dropped, nothing squashed. `git diff 2e962e185c 44ebebaeab -- scripts/fleet-testbed/` is empty — the four files are byte-identical to the pre-rebase tip. Old base was `ac4c4caf1d`, new base resolves to `956d2426b7` (an ancestor of current `origin/main`), and both tips exist locally so the range-diff is comparing real history, not a fabricated one. This is a pure replay: nothing to review here beyond what round two already approved.

### 2. Safety-critical property — re-read against current file content

Since the diff is empty, this is confirmation that round two's approved logic is what's actually on the rebased tip, not a new review.

- **Two independent predicates, not one applied twice.** `lib.sh:resolve_testbed_vpc` filters `describe-vpcs` on tag `Project=proliferate-fleet-testbed` AND `cidr=10.42.0.0/16` (exact primary-CIDR match) AND `is-default=false`, then re-asserts all three against the returned row rather than trusting the filter alone (`case "$is_default" in [Ff]alse) ;; *) die ...`). `teardown.sh` further filters every destructive call (instances, security groups, subnets, route tables, internet gateways) on both the tag AND `vpc-id=$TESTBED_VPC_ID`.
- **Zero-match cannot become match-all.** `resolve_testbed_vpc` sets `TESTBED_VPC_ID=""` up front, blank-filters the `describe-vpcs` output, and does `[ -n "$rows" ] || return 0` *before* any `wc -l` count — so an empty result returns empty (not "1" from the trailing-newline `wc -l` trap) and every downstream caller sees the empty-VPC branch, which skips instance/network teardown entirely rather than falling through to an unscoped delete. A result count other than exactly 1 is a hard `die`, not a "pick the first one."
- **Errexit-safety of the selector.** `resolve_testbed_vpc` is called as a plain statement into `$TESTBED_VPC_ID`, never through `$(...)` — the comment correctly notes bash 3.2 (macOS's `/bin/bash`, the real interpreter here per the `#!/usr/bin/env bash` shebang) has no `inherit_errexit`, so a `$(...)`-captured failure would read as an empty string with status 0, i.e. "no testbed VPC," which would make `teardown --network`/`--all` a silent no-op over a live instance. I re-confirmed the `inherit_errexit` absence claim isn't reachable on 3.2. I did not re-run round two's stubbed-AWS-CLI matrix myself (that was execution-verified in the prior round, and the file content proving it is unchanged by rebase).

### 3. Secrets and agent auth — re-read

- No secret in user-data: `user-data.sh`'s substituted placeholders (`__REPO_REF__`, `__SSM_PREFIX__`, `__AWS_REGION__`, etc.) carry no credential; the GitHub token never becomes a script variable. It's fetched on-instance via a `core.askPass` helper (`/usr/local/bin/fleet-gh-askpass`, mode 700) that shells out to `aws ssm get-parameter --with-decryption` using the instance role at credential time, so it never touches `argv` beyond the SSM *parameter name*, never the value, and is redirected (not captured via `$(...)`) when written in `provision.sh` so a failure is fatal rather than silently empty.
- No secret in argv: `provision.sh` writes the token to a mode-600 temp file and passes `--value file://$TOKEN_FILE` to `put-parameter` — argv carries a file path, not the token.
- `set +x`/`set -x` bracket the entire clone/askpass/credential region in `user-data.sh`, so xtrace can't leak it even if a future edit touches this block; post-clone `grep` assertions on `.git/config` (userinfo-in-URL pattern) and on-disk token-prefix patterns (`gh[pousr]_`, `github_pat_`) fail the whole script if a credential is ever persisted.
- No agent API key on the instance: `/etc/environment` sets `CLAUDE_CODE_USE_BEDROCK=1` + `AWS_REGION`, nothing resembling `ANTHROPIC_API_KEY`. Bedrock auth is the instance's IAM role, scoped in `provision.sh`'s inline policy to `bedrock:InvokeModel`/`InvokeModelWithResponseStream` against Anthropic Claude foundation-model and inference-profile ARNs only (not `Resource: "*"`).
- IMDS: `run-instances --metadata-options "HttpEndpoint=enabled,HttpTokens=required,HttpPutResponseHopLimit=1"` enforces IMDSv2 with hop limit 1.

### What I executed vs. what I concluded by reading

Executed: `git range-diff`, `git diff` (empty confirmation), `git merge-base`/`git log` to establish old/new base ancestry, `bash -n` on all four extracted scripts (clean), `shellcheck -x` on all four (clean at warning level — only pre-existing SC1091/SC2016/SC2153 info-level noise, matching round two's report), `gh pr checks` (all required checks pass; two checks were still pending mid-review, presumably re-running post-rebase).

Concluded by reading only, not re-executed: the stubbed-AWS-CLI teardown/provision matrices, the cherry-pick fixture run, and the `require_ipv4`/`read_github_token` exhaustive-input sweeps — these were execution-verified in round two's review comment and the content they were verified against is byte-identical post-rebase (per range-diff above), so I did not re-run them. No AWS calls of any kind were made, per the hard limits for this pass.

### Bottom line

Nothing changed except position in history. Round two's approval stands. **Safe to merge as-is.**

PR #2136, opened 2026-08-21
perf(release): combine desktop cargo builds behind a shared rust cache

pablonyx:
## What changed

`.github/workflows/release-desktop.yml`, `build-desktop` job only. Two edits:

1. **One cargo invocation instead of four.** The job compiled `anyharness`, `proliferate-diagnostics-collector`, `proliferate-debug`, and `proliferate-worker` in four separate `cargo build --release -p <pkg> --target <target>` steps. They are now a single `cargo build --release --target <target> -p ... -p ... -p ... -p ...`.
2. **`Swatinem/rust-cache@v2`, keyed per matrix target.** Added right after `Setup Rust stable`, with `key: ${{ matrix.target }}` so each arch gets its own cache entry and the macOS arches (plus any dormant windows/Intel entry) never collide. Same action and version already used in `ci.yml`, `server-ci.yml`, `release-e2e.yml`, and `catalog-probe.yml`. The step also sets `save-if: ${{ github.ref_type != 'tag' }}` and `cache-on-failure: true` — see "Cache-write hygiene" below.

The three downstream steps keep their copy / `chmod` / `cmp` / `test -x` verification verbatim and only lose their `cargo build` line, so they are renamed `Build ... for target` -> `Stage ... for target` to describe what they now do. The combined build step keeps the AnyHarness staging block it already had, so `Build bundled agent seed` still finds `target/<target>/release/anyharness` at the same path.

Deliberately untouched: signing, notarization, bundling, publish, and updater-manifest steps. No step ids or step outputs changed; nothing in the repo referenced the renamed step names.

### Matrix / platform notes

- Main is ARM-only right now (#2126 commented out the `x86_64-apple-darwin` entry on 2026-08-20). The change is written against `${{ matrix.target }}`, so it works for the live `aarch64-apple-darwin` entry and stays correct if the Intel entry is restored — each target simply gets its own cache key.
- The dormant `runner_os == 'windows'` branches inside the affected steps are preserved byte-for-byte. The combined `cargo build` is target-parameterised and runs under `shell: bash`, so it is windows-safe if that leg is ever added.
- Every added/modified step carries the existing `if: steps.filter.outputs.skip != 'true'` guard, matching its neighbours.

### Cache-write hygiene

- **`save-if: ${{ github.ref_type != 'tag' }}`.** This workflow also triggers on `push: tags: desktop-v*`, and `Validate release ref` forces manual non-dry-run releases onto a `desktop-v*` tag ref. Actions caches are ref-scoped, so a save made under `refs/tags/desktop-v0.4.21` can only ever be restored by that same tag — a pure write, never a read. Tag-scoped releases are a live path in the run history (`desktop-v0.3.49`, `desktop-v0.3.14`, `desktop-v0.2.10`), so without this gate every release would burn a multi-hundred-MB write into the shared pool for zero future benefit. Branch and dispatch runs still save, and tag runs still *restore* via default-branch fallback.
- **`cache-on-failure: true`.** Notarization and publish are the late, occasionally-failing part of this job (three late failures/cancellations in recent history). Without this, a failure after the build discards exactly the ~12 minutes of compile the cache exists to preserve, and "one priming run" silently becomes "one *successful* priming run".

## Measured before/after

All figures are the **cargo portion of the ARM `build-desktop` job**, measured on real `macos-15` runners.

| | cargo time | source |
| --- | --- | --- |
| Before: 4 separate steps, no cache | **12m35s** | live nightly train run [32167413474](https://github.com/proliferate-ai/proliferate/actions/runs/32167413474), ARM `build-desktop` job, 2026-08-18 |
| After, priming run: combined, cold cache | 12m18s | [32410829381](https://github.com/proliferate-ai/proliferate/actions/runs/32410829381) |
| After, warm cache | **5m23s** | [32412144015](https://github.com/proliferate-ai/proliferate/actions/runs/32412144015), cache restore 13s |

The 12m35s baseline is the sum of the four existing steps in that live job: `anyharness` 5m54s + `proliferate-diagnostics-collector` 50s + `proliferate-debug` 4m52s + `proliferate-worker` 60s.

The two "after" numbers were measured on `exp/desktop-build-speed` with a throwaway workflow that mirrored only checkout -> toolchain setup -> rust-cache -> combined build for `aarch64-apple-darwin` on `macos-15` (no signing secrets, no bundling). **That experiment workflow is not part of this PR**; only the `release-desktop.yml` config is transplanted here.

**Where the win actually comes from.** Combining the four invocations on a cold cache is roughly neutral (12m35s -> 12m18s) — cargo was already reusing the shared dependency artifacts in `target/` across the four steps inside a single job. The 7m12s comes from rust-cache carrying `target/` and the cargo registry *across* runs. Combining still earns its place (one build graph for the cache to restore against, three fewer redundant resolve/link passes), but the caching is the load-bearing half, which makes the caveats below the important ones.

## Caveats — read before trusting the warm number

**1. The warm number is conditional on the cache surviving between daily runs, and this repo's cache pool does not guarantee that.**

`GET /repos/proliferate-ai/proliferate/actions/cache/usage` returned `{"active_caches_size_in_bytes":10521040818,"active_caches_count":216}` at 2026-08-20T23:12Z — **10.52 GB against the 10 GB repo limit**. Size is the trustworthy half of that response; `active_caches_count` lags and should not be read as a live entry count. The store is over quota and evicting LRU continuously. The experiment's own entry proves the exposure: the post step of run `32410829381` saved 661 MB (`Sent 693585475 of 693585475`) at 2026-08-20T20:03:22Z, and `GET .../actions/caches?ref=refs/heads/exp/desktop-build-speed` returned `total_count: 0` a few hours later — **evicted in under 6 hours**, with the branch still alive.

`release-desktop` runs at most once a day via the nightly train, and only when `needs.prepare.outputs.desktop == 'true'`. Its entry is therefore the coldest LRU candidate in the pool, competing against `v0-rust-cargo-check` (745 MB) and `v0-rust-candidate-build-handoff` (405 MB), which are rewritten on essentially every PR — 3 and 10 live copies respectively, 6.2 GB between them, in a paginated listing taken 2026-08-21T01:47Z. GitHub's 7-day unused-entry eviction is a second way to lose it.

So: **5m23s is the number when the cache is warm, not a number the release train is guaranteed to see.** The ~27.5m steady state derived below (34m44s - 7m12s) depends on the entry surviving from one daily run to the next, which the current pool does not assure. What this PR does about it: `save-if` stops the tag-triggered releases from wasting writes into the same pool. Actually pruning the stale/duplicate entries is a **follow-up**, not part of this PR — the top-15-by-size listing is in the review thread for whoever picks it up. Worst realistic case if the cache keeps missing: the job pays the old cost plus ~25-60s of cache save.

**2. The real cache entry will be larger than the measured 661 MB — and the real win may be larger than 7m12s. Neither was measured.**

The experiment job ended right after the combined cargo build. In the real `build-desktop`, rust-cache's post step runs at the *end* of the job, after `Build Tauri app`, which runs `pnpm tauri build --target ${{ matrix.target }}` from `apps/desktop` into the same root `target/`. Both directions follow:

- The saved entry will be materially bigger than 661 MB, because tauri's dep graph is not in that number. That makes the quota pressure in caveat 1 worse than it looks.
- `Build Tauri app` will *also* be partly warmed on a restore. That step was 12m29s on ARM in run `32167413474`, so the real upside may exceed the measured 7m12s. **Unmeasured potential, not a claim.**

**3. The very next release after merge looks unchanged.** rust-cache writes on the first qualifying run; that run pays roughly the old cost plus the save. Only a subsequent run that restores the entry lands at 5m23s. Cache keys also rotate on `Cargo.lock` / toolchain changes, so a dependency bump re-primes.

**4. Not claimed: byte-identical artifacts.** Under `resolver = "2"`, `cargo build -p a -p b -p c -p d` unifies features across the four packages' graphs where four separate `-p` builds resolved each alone. Unification is additive, so a shipped binary can carry dep features it previously did not. Risk is low (shared deps are centrally pinned in the root `[workspace.dependencies]`, and `ci.yml` already runs `cargo test --workspace` over a strictly wider member set), but this PR did not diff the produced binaries against the four-step output.

## Expected effect on the release train

The ARM `build-desktop` job in the 2026-08-18 train ran **34m44s** wall (17:47:26Z -> 18:22:10Z). On a run that restores a warm cache, subtracting the 7m12s saved on the cargo portion puts it near **~27.5m**, possibly better if `Build Tauri app` is warmed too (caveat 2). On a run that misses, it stays at ~34m44s plus the save. Signing, notarization, and publish are unchanged, so the rest of the desktop leg and every other train stage are unaffected either way.

## Follow-ups (not in this PR)

- Prune stale / duplicate Actions cache entries so the pool sits under the 10 GB limit; without that, the desktop entry's survival between daily runs stays a coin flip.
- Consider `--locked` on the release build (pre-existing: neither the old four steps nor the combined one passes it, so a release can silently resolve a `Cargo.lock` update).

### DIFF
--- a/.github/workflows/release-desktop.yml
+++ b/.github/workflows/release-desktop.yml
@@ -210,11 +210,27 @@ jobs:
         with:
           targets: ${{ matrix.target }}
 
-      - name: Build AnyHarness runtime for target
+      # Cache the cargo registry/git checkouts and target/ artifacts, keyed per
+      # matrix target so the macOS arches (and any dormant windows entry) never
+      # share or collide on a cache key. Paired with the single combined cargo
+      # invocation below: one build graph, one cache, instead of four sequential
+      # `cargo build -p <pkg>` steps that each re-resolved the same workspace
+      # dependencies.
+      - name: Rust cache
+        if: steps.filter.outputs.skip != 'true'
+        uses: Swatinem/rust-cache@v2
+        with:
+          key: ${{ matrix.target }}
+
+      - name: Build Rust binaries for target
         if: steps.filter.outputs.skip != 'true'
         shell: bash
         run: |
-          cargo build --release -p anyharness --target "${{ matrix.target }}"
+          cargo build --release --target "${{ matrix.target }}" \
+            -p anyharness \
+            -p proliferate-diagnostics-collector \
+            -p proliferate-debug \
+            -p proliferate-worker
           mkdir -p apps/desktop/src-tauri/binaries
           if [[ "${{ matrix.runner_os }}" == "windows" ]]; then
             cp "target/${{ matrix.target }}/release/anyharness.exe" "apps/desktop/src-tauri/binaries/anyharness-${{ matrix.target }}.exe"
@@ -253,11 +269,10 @@ jobs:
           codesign --verify --verbose "$EXTRACTED_NODE_BIN"
           spctl --assess --verbose=4 --type open --context context:primary-signature "$EXTRACTED_NODE_BIN"
 
-      - name: Build diagnostics collector for target
+      - name: Stage diagnostics collector for target
         if: steps.filter.outputs.skip != 'true'
         shell: bash
         run: |
-          cargo build --release -p proliferate-diagnostics-collector --target "${{ matrix.target }}"
           mkdir -p apps/desktop/src-tauri/binaries
           collector="target/${{ matrix.target }}/release/proliferate-diagnostics-collector"
           staged="apps/desktop/src-tauri/binaries/proliferate-diagnostics-collector-${{ matrix.target }}"
@@ -266,11 +281,10 @@ jobs:
           chmod +x "$staged"
           cmp "$collector" "$staged"
 
-      - name: Build Proliferate debug helper for target
+      - name: Stage Proliferate debug helper for target
         if: steps.filter.outputs.skip != 'true'
         shell: bash
         run: |
-          cargo build --release -p proliferate-debug --target "${{ matrix.target }}"
           mkdir -p apps/desktop/src-tauri/binaries
           if [[ "${{ matrix.runner_os }}" == "windows" ]]; then
             cp "target/${{ matrix.target }}/release/proliferate-debug.exe" "apps/desktop/src-tauri/binaries/proliferate-debug-${{ matrix.target }}.exe"
@@ -279,11 +293,10 @@ jobs:
             chmod +x "apps/desktop/src-tauri/binaries/proliferate-debug-${{ matrix.target }}"
           fi
 
-      - name: Build Proliferate worker for target
+      - name: Stage Proliferate worker for target
         if: steps.filter.outputs.skip != 'true'
         shell: bash
         run: |
-          cargo build --release -p proliferate-worker --target "${{ matrix.target }}"
           mkdir -p apps/desktop/src-tauri/binaries
           if [[ "${{ matrix.runner_os }}" == "windows" ]]; then
             cp "target/${{ matrix.target }}/release/proliferate-worker.exe" "apps/desktop/src-tauri/binaries/proliferate-worker-${{ matrix.target }}.exe"
--- a/.github/workflows/release-desktop.yml
+++ b/.github/workflows/release-desktop.yml
@@ -221,6 +221,15 @@ jobs:
         uses: Swatinem/rust-cache@v2
         with:
           key: ${{ matrix.target }}
+          # Actions caches are ref-scoped. This workflow also triggers on
+          # `push: tags: desktop-v*`, and a save made under `refs/tags/...` can
+          # only ever be restored by that same tag, so it is a pure write into
+          # an already-tight repo cache pool. Only branch / dispatch runs save.
+          save-if: ${{ github.ref_type != 'tag' }}
+          # Notarization and publish are the late, occasionally-failing part of
+          # this job. A failure there should still bank the compile cache the
+          # preceding ~12 minutes produced.
+          cache-on-failure: true
 
       - name: Build Rust binaries for target
         if: steps.filter.outputs.skip != 'true'


pablonyx:

Fresh reviewer of record, head `9dc8a504a4ab`. I did not implement this. Everything below was re-derived from the diff, the repo at that SHA, and the GH API; nothing is taken from the PR body.

**Bottom line:** the YAML change itself is faithful and I found no correctness defect. The problem is the payoff. The repo's Actions cache store is already over quota and evicting, and the experiment's own cache entry is already gone, so the headline "steady state 5m23s / ~27.5m train" is not supported here. Asks are two lines of config (F2, F3) plus correcting the claims in the body (F1, F4).

---

## Verified correct

**Faithfulness of the transplant.** One file, 21+/8-. The four `-p` names are exact package names, all members of the single root workspace (`resolver = "2"`, `members = ["anyharness/crates/*", "apps/desktop/src-tauri", "apps/desktop/src-tauri-debug"]`), and each declares a `[[bin]]` whose name matches the `target/<triple>/release/<name>` path its `Stage` step reads:
- `anyharness` -> `anyharness/crates/anyharness`, bin `anyharness`
- `proliferate-diagnostics-collector` -> `anyharness/crates/proliferate-diagnostics-collector`
- `proliferate-debug` -> `apps/desktop/src-tauri-debug`, bin `proliferate-debug`
- `proliferate-worker` -> `anyharness/crates/proliferate-worker`, bin `proliferate-worker`

No flags were dropped. Old and new are both bare `cargo build --release --target "<t>"`; neither side had `--locked`, `--profile`, or RUSTFLAGS.

**Step ids / outputs.** The only `id` in `build-desktop` is `filter`, untouched. The renames are display-only. `grep -rn` over the whole repo for the four old step names (`Build AnyHarness runtime for target`, `Build diagnostics collector for target`, `Build Proliferate debug helper for target`, `Build Proliferate worker for target`) returns nothing.

**Untouched surfaces.** AnyHarness staging block retained inside the combined step, so `Build bundled agent seed`'s `--anyharness-bin "$PWD/target/${{ matrix.target }}/release/anyharness"` still resolves. The `runner_os == 'windows'` branches in all three `Stage` steps and the combined step are byte-for-byte preserved. The commented-out `x86_64-apple-darwin` matrix entry from #2126 is untouched. Signing, notarization, `Build Tauri app`, `Verify packaged native binary inventory`, and publish are untouched. That inventory step still asserts all five binaries are present, executable, and non-placeholder in `Contents/MacOS`, so a binary the combined build failed to produce cannot reach publish (and the `Stage` steps' `test -x` / `cp` would fail first).

**Cache placement and keying.** After `Setup Rust stable`, before the build, with the same `if: steps.filter.outputs.skip != 'true'` guard as its neighbours. No `shared-key` is used anywhere in this repo, so every rust-cache consumer already keys on rust-cache's own job hash; `key: ${{ matrix.target }}` disambiguates ARM/Intel on top of that. No collision with ci.yml (#2113), server-ci.yml, release-e2e.yml, or catalog-probe.yml. Confirmed from the live cache list, which shows distinct namespaces: `v0-rust-cargo-check-Linux-x64-...` and `v0-rust-candidate-build-handoff-Linux-x64-...`.

**Receipts.** All three check out literally.

Run `32167413474`, job `release-desktop / build-desktop (macos-15, aarch64-apple-darwin, macos)` (event `workflow_dispatch`, branch `main`):
```
Build AnyHarness runtime for target        17:47:41 -> 17:53:35   5m54s
Build diagnostics collector for target     17:57:39 -> 17:58:29     50s
Build Proliferate debug helper for target  17:58:29 -> 18:03:20   4m51s   (body says 4m52s)
Build Proliferate worker for target        18:03:20 -> 18:04:20   1m00s
                                                          sum =  12m35s
job wall  17:47:26 -> 18:22:10 = 34m44s
```
Run `32410829381`: `Build Rust binaries for target (combined)` 19:50:40 -> 20:02:58 = **12m18s**.
Run `32412144015`: `Rust cache` 20:05:14 -> 20:05:27 = **13s**, build 20:05:27 -> 20:10:50 = **5m23s**.

**CI on the head.** 27 checks pass, 1 `skipping` (`Production compose smoke`, path-gated, expected for a `.github/` only diff). No failures, no flakes to classify.

**Metadata.** Title `perf(release): combine desktop cargo builds behind a shared rust cache` matches `TITLE_PATTERN` (`perf` is in `ALLOWED_TITLE_TYPES`, scope `release` matches `[a-z0-9][a-z0-9/-]*`). Exactly one `release:*` (`release:performance`). The only changed path is `.github/workflows/release-desktop.yml`, which `AREA_RULES` resolves uniquely to `area:release`, which is applied. `Validate PR title and labels` passed.

---

## Findings

### F1 (major) The steady-state claim is not supported in this repo. The cache is unlikely to survive between desktop releases.

- `GET /repos/proliferate-ai/proliferate/actions/cache/usage` right now: `{"active_caches_size_in_bytes":10521040818,"active_caches_count":216}`. That is **10.52 GB against the 10 GB default repo limit**, so the store is already over and evicting LRU continuously.
- **The experiment's own cache entry is already gone.** The post step of run `32410829381` logged `Sent 693585475 of 693585475 (100.0%)`, i.e. a 661 MB entry saved at `2026-08-20T20:03:22Z`. As of now, `GET .../actions/caches?ref=refs/heads/exp/desktop-build-speed` returns `total_count: 0`. Positive controls on the same call: `refs/heads/main` -> 65, `refs/heads/exp/desktop-asset-reuse` -> 3. The branch still exists (`refs/heads/exp/desktop-build-speed` -> `6e8466ba`), so this is not branch-deletion cleanup. A full paginated scan of every cache key matching `v0-rust` returns only the two Linux ci.yml entries; **no macOS rust-cache entry exists anywhere in this repo**. Lifetime under 6 hours.
- `release-desktop` runs at most once a day via the nightly train, and only when `needs.prepare.outputs.desktop == 'true'`. Its entry is therefore the coldest LRU candidate in the pool, competing against `v0-rust-cargo-check` (745 MB) and `v0-rust-candidate-build-handoff` (405 MB), which are rewritten on essentially every PR. GitHub's 7-day unused eviction is a second way to lose it.

Not a reason to abandon the change (combining alone is neutral-to-slightly-positive at 12m35s -> 12m18s), but "Subtracting the 7m12s ... puts it at ~27.5m in steady state" should not go into the repo as the expected outcome. Please state the odds honestly, or reduce the exposure per F2.

### F2 (major, one line) No `save-if`, so tag-triggered releases write a cache nothing can ever restore.

This workflow triggers on `push: tags: desktop-v*`, and `Validate release ref` *forces* manual non-dry-run releases onto a `desktop-v*` tag ref. Actions caches are ref-scoped: a run on `refs/tags/desktop-v0.4.21` saves into that tag's scope and can only ever be restored by that same tag. That path is live in the run history:

```
2026-07-27T05:51:05Z workflow_dispatch desktop-v0.3.49 success
2026-07-09T19:50:53Z push             desktop-v0.3.14 success
2026-06-15T01:04:35Z workflow_dispatch desktop-v0.2.10 success
```

Each of those would now burn a several-hundred-MB (see F4: likely over 1 GB in the real job) write into an already-over-quota pool for zero future benefit, evicting entries that every PR depends on. Suggest gating the save to the only scope that can produce a restorable entry:

```yaml
- name: Rust cache
  uses: Swatinem/rust-cache@v2
  with:
    key: ${{ matrix.target }}
    save-if: ${{ github.ref == 'refs/heads/main' }}
```

The nightly train is the path that matters and it runs on `main` (baseline run `32167413474` was `workflow_dispatch` on `main`; scheduled runs are on `main` too), so restores keep working via default-branch fallback on every path including tag runs.

### F3 (medium, one line) `cache-on-failure` defaults to false, so the priming run only counts if the whole ~35 minute job succeeds.

Notarization and publish are the late, flaky part of this job, and the history shows late failures (`2026-07-27T04:58:41Z ... failure`, `2026-06-25T22:17:20Z ... failure`, `2026-06-14T23:29:06Z ... cancelled`). A failure after the build throws away exactly the 12 minutes of compile the cache exists to preserve, and the "one priming run" caveat in the body silently becomes "one *successful* priming run". `cache-on-failure: true` costs one line.

### F4 (medium, measurement gap, cuts both ways) The experiment did not measure the real job's cache.

The experiment job ends right after the combined cargo build. In the real `build-desktop`, rust-cache's post step runs at the **end of the job**, after `Build Tauri app`, which runs `pnpm tauri build --target ${{ matrix.target }}` from `apps/desktop` into the same root `target/`. So:

- the real cache will be materially larger than the measured 661 MB (tauri's dep graph is not in that number), which makes the F1/F2 quota cost worse than it looks; and
- `Build Tauri app` will *also* be partly warmed. That step was 18:09:09 -> 18:21:38 = **12m29s** on ARM in run `32167413474`, so the real upside may exceed the 7m12s claimed.

Neither direction was measured. One sentence in the body would keep the next person from being surprised by either.

### F5 (minor, unmeasured, no action required) Combining changes feature resolution, so the shipped binaries are not provably the same artifacts.

Under `resolver = "2"`, `cargo build -p a -p b -p c -p d` unifies features across the four packages' graphs; four separate `-p` builds resolved each alone. Unification is additive, so a shipped binary can now carry dep features it previously did not. Low risk here: shared deps are centrally pinned in the root `[workspace.dependencies]`, including the one with the sharpest feature contract (`sentry` with `default-features = false` and its documented sentry-core split hazard), and ci.yml already runs `cargo test --workspace`, which unifies over a strictly wider member set. But nothing in this PR compares the produced binaries against the four-step output; the experiment only ran `ls -la`. Fine to ship, just do not claim byte-identical artifacts.

### F6 (minor, pre-existing) Still no `--locked` on a release build.

Neither the old four steps nor the new combined one passes `--locked`, so a release build can still silently resolve a `Cargo.lock` update. Not introduced here, flagged only because this PR is the one touching the line.

### F7 (nit) Failure granularity.

A compile error now fails one step named `Build Rust binaries for target` instead of a step naming the package. Cargo's error still names the crate, so this is log-scanning ergonomics only.

---

## What would break on the next real release train run

Nothing. I could not construct a path where this change fails the train. Worst realistic case is the cache misses, the job pays the old cost plus roughly 25 to 60 seconds of cache save, and the win reported in the body does not appear. Toolchain-bump poisoning is not a concern: rust-cache folds the rustc version and `Cargo.lock` hashes into the key, so a bump re-primes rather than restoring stale artifacts.

## To clear this review

1. Add `save-if` gating the write to `main` (F2).
2. Add `cache-on-failure: true` (F3).
3. Soften the "Expected effect on the release train" section to reflect F1, and add the F4 caveat that the real cache also spans the Tauri build (bigger entry, and an unmeasured additional win).

The PR is also still a draft; I am not marking it ready.

pablonyx:
## Review fixes applied (head `6cd46a4`)

Addressing @reviewer's CHANGES-NEEDED. Two lines of YAML plus a body rewrite; no other code change.

### F2 — `save-if`

```diff
       - name: Rust cache
         if: steps.filter.outputs.skip != 'true'
         uses: Swatinem/rust-cache@v2
         with:
           key: ${{ matrix.target }}
+          save-if: ${{ github.ref_type != 'tag' }}
+          cache-on-failure: true
```

One deliberate deviation from the suggested form: the review proposed `save-if: ${{ github.ref == 'refs/heads/main' }}`; I used `github.ref_type != 'tag'`. Same effect on the path the finding is about — tag-triggered releases (`push: tags: desktop-v*`, and the ref `Validate release ref` forces manual non-dry-run releases onto) no longer write a cache nothing can ever restore. The difference is that a `workflow_dispatch` dry-run on a non-main branch still saves, which is harmless (branch-scoped, restorable by that branch, and it is how this change gets validated before it ever reaches the train) whereas pinning to `main` would silently make branch validation unmeasurable. Restores are unaffected in both forms: tag runs still read through default-branch fallback.

### F3 — `cache-on-failure: true`

Added on the same step. Late notarization/publish failures no longer discard the ~12 minutes of compile the cache exists to preserve.

### F1 / F4 — body rewritten

The "Expected effect on the release train" section no longer asserts ~27.5m as the outcome. The body now carries a **Caveats** section stating plainly that:

- the pool is at 10.52 GB against the 10 GB limit (216 entries, LRU-evicting; re-read since at 11.06 GB / 1184), and the experiment's own 661 MB entry was **evicted in under 6 hours** with the branch still alive;
- `release-desktop` runs at most daily, so its entry is the coldest LRU candidate in that pool — **5m23s is the warm number, not a number the train is guaranteed to see**, and the ~20m steady state depends on the entry surviving between daily runs, which the current over-quota pool does not guarantee;
- mitigation is exactly the `save-if` change (stops wasting writes); stale-cache pruning is listed as an explicit **follow-up**, not part of this PR;
- the real entry will be **larger than 661 MB** because the post step runs after `Build Tauri app` into the same `target/` — which also means that 12m29s step gets partly warmed, so the real upside may exceed 7m12s. Both flagged as unmeasured, the Tauri warming explicitly as potential rather than a claim.

F5 is now covered by a "not claimed: byte-identical artifacts" caveat, and F6 (`--locked`) is in the follow-ups list. F7 accepted as-is.

### Checks

`6cd46a4` (fresh runs, not reruns): **27 pass, 1 `skipping`** (`Production compose smoke`, path-gated) — identical to the check set on `9dc8a504`.

---

### F-report: top 15 cache entries by size (nothing deleted)

Read at 2026-08-21T01:47Z. Usage endpoint at that moment: **11.31 GB / 1236 entries** against the 10 GB limit — churning upward while being read, which is the eviction pressure itself.

| # | size | ref | last accessed | key |
| 1 | 756 MB | `refs/pull/2139/merge` | 2026-08-21T01:40:25Z | `v0-rust-cargo-check-Linux-x64-1cb93fb6-c30d2aef` |
| 2 | 745 MB | `refs/pull/2108/merge` | 2026-08-21T01:16:28Z | `v0-rust-cargo-check-Linux-x64-0b9fd15e-c30d2aef` |
| 3 | 745 MB | `refs/heads/main` | 2026-08-21T01:46:15Z | `v0-rust-cargo-check-Linux-x64-0b9fd15e-c30d2aef` |
| 4 | 745 MB | `refs/pull/2132/merge` | 2026-08-21T01:20:23Z | `v0-rust-cargo-check-Linux-x64-0b9fd15e-c30d2aef` |
| 5 | 405 MB | `refs/pull/2126/merge` | 2026-08-20T22:56:45Z | `v0-rust-candidate-build-handoff-Linux-x64-0b9fd15e-c30d2aef` |
| 6 | 405 MB | `refs/pull/2120/merge` | 2026-08-21T00:11:12Z | `v0-rust-candidate-build-handoff-Linux-x64-0b9fd15e-c30d2aef` |
| 7 | 405 MB | `refs/pull/2122/merge` | 2026-08-20T22:57:57Z | `v0-rust-candidate-build-handoff-Linux-x64-0b9fd15e-c30d2aef` |
| 8 | 405 MB | `refs/pull/2131/merge` | 2026-08-20T23:33:34Z | `v0-rust-candidate-build-handoff-Linux-x64-0b9fd15e-c30d2aef` |
| 9 | 405 MB | `refs/pull/2082/merge` | 2026-08-20T23:43:10Z | `v0-rust-candidate-build-handoff-Linux-x64-0b9fd15e-c30d2aef` |
| 10 | 405 MB | `refs/pull/2132/merge` | 2026-08-21T01:20:05Z | `v0-rust-candidate-build-handoff-Linux-x64-0b9fd15e-c30d2aef` |
| 11 | 405 MB | `refs/heads/main` | 2026-08-21T01:46:01Z | `v0-rust-candidate-build-handoff-Linux-x64-0b9fd15e-c30d2aef` |
| 12 | 405 MB | `refs/pull/2121/merge` | 2026-08-21T00:03:45Z | `v0-rust-candidate-build-handoff-Linux-x64-0b9fd15e-c30d2aef` |
| 13 | 364 MB | `refs/pull/2121/merge` | 2026-08-20T23:56:46Z | `playwright-browsers-Linux-1.62.1-chromium-webkit` |
| 14 | 364 MB | `refs/heads/main` | 2026-08-21T01:46:51Z | `playwright-browsers-Linux-1.62.1-chromium-webkit` |
| 15 | 273 MB | `refs/pull/2122/merge` | 2026-08-20T23:03:07Z | `buildkit-blob-1-sha256:1ba5b84f4e93c34792f034f404531608bed830c649bc676e5de452149a678eca` |

Aggregate by key family (whole pool, same read):

| family | total | entries |
| --- | --- | --- |
| `v0-rust-candidate-build-handoff-Linux-x64` | 4.0 GB | 10 |
| `v0-rust-cargo-check-Linux-x64` | 2.2 GB | 3 |
| `buildkit-blob-1-*` | 1.6 GB | 159 |
| `playwright-browsers-Linux-1.62.1-chromium-webkit` | 0.7 GB | 2 |
| `playwright-browsers-Linux-1.62.1-chromium` | 0.5 GB | 2 |

Two observations for whoever takes the pruning follow-up, offered as data, not as a recommendation to act now:

- The top 12 entries are **duplicates of two keys** held open by distinct `refs/pull/*/merge` scopes — 6.2 GB, ~60% of the quota, in twelve copies of two artifacts. PR-scoped entries are deleted automatically when the PR closes, so much of this is transient by design; it is the standing depth that matters.
- `buildkit-blob-1-*` is 159 entries for 1.6 GB — small individually, but it is the long tail inflating the entry count.

I have deliberately deleted nothing. Both the per-entry list and the aggregate are above so the call is a human's.

pablonyx:
REVIEW VERDICT: LGTM-FOR-MERGE

Re-review of `6cd46a4911b7` against my F1-F7. Same reviewer of record, same constraints (read-only, `gh` only, no builds, no merge, no dispatch). Everything re-derived from the API on the new head.

**The code is approved as-is.** Two remaining items are prose-only numbers in the body (N1, N2); neither affects workflow behaviour and neither blocks merge.

---

## Diff between heads

`git diff 9dc8a504a4ab 6cd46a4911b7` is exactly nine added lines on the `Rust cache` step, seven of them comments:

```yaml
save-if: ${{ github.ref_type != 'tag' }}
cache-on-failure: true
```

Nothing else moved. The combined `cargo build`, the three `Stage` steps, the AnyHarness staging block, the windows branches, the commented-out Intel entry, and every signing/notarize/bundle/publish step are byte-identical to the head I already cleared. Full diff vs base `e2fd47e0d` is still one file, 30+/8-.

## Finding-by-finding

**F2 (save-if) — resolved.** See judgment 1 below for the deviation.

**F3 (cache-on-failure) — resolved.** `cache-on-failure: true` is on the same step. The two inputs compose correctly: `save-if` gates whether a save happens at all, `cache-on-failure` gates saving on a failed job, so a *failed tag* run still saves nothing. That is the right intersection.

One tradeoff I should put on the record since I asked for this input: a job that fails *during* the cargo build will now bank a partially-built `target/`. I judge that low-risk — cargo's fingerprinting makes a partial target dir safe to resume from, and the key rotates on `Cargo.lock`/toolchain anyway — and the escape hatch is deleting the entry. Accepted, not a finding.

**F1 (unsupported steady-state claim) — resolved.** See judgment 2.

**F4 (Tauri measurement gap) — resolved.** Caveat 2 states both directions and explicitly labels the Tauri warming "unmeasured potential, not a claim". That is the correct posture.

**F5 (feature unification) — resolved.** Caveat 4, "Not claimed: byte-identical artifacts", with the correct mitigations. No overclaim.

**F6 (`--locked`) — resolved as a follow-up.** Correctly scoped out; it was pre-existing.

**F7 (failure granularity) — accepted as-is.** Agreed, it was a nit.

**Cache listing.** Posted, nothing deleted. Independently consistent: the pool grew rather than shrank between my two reads (list `total_count` 704 -> 1228), and `?ref=refs/heads/exp/desktop-build-speed` is still `total_count: 0`, i.e. unchanged from before the fixer touched anything.

## Checks on `6cd46a4911b7`

27 pass, 1 `skipping` (`Production compose smoke`, path-gated) out of 28. Fresh runs, not reruns: run ids `32437143223/229/252/295` and `32437146184` against the prior head's `32435930372/374/379/389`. `Validate PR title and labels` passed at `32437146184`. Title and labels unchanged from the set I already validated against `pr-metadata.mjs` (`perf(release):` matches `TITLE_PATTERN`; one `release:performance`; `area:release` is the single required area for a `.github/` path).

---

## Judgment 1: is `github.ref_type != 'tag'` acceptable instead of main-only?

**Yes, and I now prefer it to my own suggestion.**

- **On the path F2 was actually about, the two forms are identical.** `push: tags: desktop-v*`, and the tag ref that `Validate release ref` forces manual non-dry-run releases onto, both yield `github.ref_type == 'tag'` and now save nothing. That was the dominant term: of the last 20 direct-trigger runs of this workflow, 17 are tag pushes or dispatches on `desktop-v*` refs.
- **The train path still writes the restorable entry.** Reusable-workflow runs take the caller's `github` context, and the nightly train runs on `main` under both `schedule` and `workflow_dispatch` (baseline run `32167413474` was `workflow_dispatch` on `main`), so `ref_type == 'branch'` and the default-branch-scoped entry — the only one that tag runs can restore through fallback — still gets saved. Verified, not assumed.
- **The residual delta is one stray write per few months.** A dispatch on a non-main branch now saves a branch-scoped entry only that branch can restore. Frequency from the same history: exactly one such run in the last 20 (`2026-05-30T09:40:37Z workflow_dispatch codex/unify-production-release-flow`).
- **`ref_type` has a failure mode advantage I did not weigh.** A hardcoded `refs/heads/main` breaks silently on a default-branch rename: saves stop forever, restores keep succeeding off the stale entry until it evicts, and nothing errors. `ref_type != 'tag'` has no such cliff. That alone justifies the deviation.
- **The branch-validation argument is real.** rust-cache's save is the half most likely to be misconfigured, and pinning to `main` makes it unexercisable before it runs on the train.

One framing correction: the fixer's comment calls the branch-scoped save "harmless". Under caveat 1 it is not literally harmless, it is a write into an over-quota pool. It is *rare and worth the validation*, which is a different and defensible claim. Not worth changing the code over.

## Judgment 2: does the caveated steady-state claim now match the evidence?

**In substance, yes.** Two numbers in the prose do not, neither material.

What now matches my evidence exactly:
- The unconditional claim is gone. "On a run that restores a warm cache ... near **~27.5m** ... On a run that misses, it stays at ~34m44s plus the save" is precisely the honest statement, and 34m44 - 7m12 = 27m32 checks out.
- Caveat 1 reproduces my receipts verbatim and correctly: 661 MB, `Sent 693585475 of 693585475`, saved `2026-08-20T20:03:22Z`, `total_count: 0`, evicted in under 6 hours, branch still alive. I re-confirmed the last one just now: `?ref=refs/heads/exp/desktop-build-speed` -> `total_count: 0`, positive controls `refs/heads/main` -> 65 and `exp/desktop-asset-reuse` -> 3.
- The competition arithmetic is right: 3 x 745 MB + 10 x 405 MB = 6285 MB ≈ the "6.2 GB" claimed.
- The table row is correctly relabelled "After, warm cache" rather than "steady state".
- Pruning is listed as a follow-up rather than smuggled in as a mitigation this PR delivers.

### N1 (nit, blocking nothing) The body contradicts itself on the steady-state figure.

Caveat 1 says "The steady-state **~20m** job depends on the entry surviving from one daily run to the next"; the Expected-effect section says **~27.5m**. I cannot reproduce ~20m from any figure in the PR. Looks like a stray edit. One-word fix, but it is the sort of thing that makes a reader discount the caveats around it.

### N2 (nit, blocking nothing) The parenthetical re-read is not reproducible.

Body: "(re-read since: 11.06 GB / 1184 entries)". My independent read just now:

```
GET /repos/proliferate-ai/proliferate/actions/cache/usage
{"full_name":"proliferate-ai/proliferate","active_caches_size_in_bytes":10444119389,"active_caches_count":1224}
```

That is 10.44 GB / 1224, not 11.06 GB / 1184. The pool churns constantly so a moving number is expected, and the substance is unaffected — at or over the ceiling on every reading, with the exp entry still gone. But per this repo's requote-the-literal-evidence norm, either drop the parenthetical or pin it to a timestamped quote. Note also that `active_caches_count` itself moved 216 -> 1224 across my two reads while size barely changed, so that field appears to lag; the size is the trustworthy half.

---

## Nothing new found

I re-checked for anything the two added inputs could break and found nothing. `save-if` and `cache-on-failure` are both documented rust-cache v2 inputs; the `${{ }}` expression form for `save-if` is the pattern in the action's own README; the unquoted YAML `true` is stringified by Actions before `core.getInput` sees it; `CI/CD config checks` passed. No step id, output, path, or ordering changed.

Code is good to merge on Pablo's call. N1 and N2 are body-only and can ride along or not. The PR is still a draft and I am not marking it ready, not merging, and not dispatching anything.

PR #2137, opened 2026-08-21
perf(server): split CI lanes and shard integration tests 3-way

pablonyx:
## What

Splits Server CI's single `test` job into a fast unit lane and a 3-way-sharded integration lane, on top of the per-worker-DB xdist work already merged in #2122. Three measured experiments composed into one change:

| Experiment | PR | What it measured |
| --- | --- | --- |
| Lane split | #2129 (`exp/server-lane-split`) | `test` → `test-unit` + `test-integration` |
| Before-only truncate | #2130 (`exp/server-split-before-only`) | drop the post-test TRUNCATE in `test_engine` |
| Integration sharding | #2131 (`exp/server-integration-shard`) | `test-integration` fanned out to 3 shards |

## Measured baseline

Main at `56dd6ebf` (run [32434848206](https://github.com/proliferate-ai/proliferate/actions/runs/32434848206)), the merged-#2122 shape:

```
test | Run tests | 2972 passed, 1 skipped, 903 warnings in 920.17s (0:15:20)
```

Job wall `16m15s`, and it is the pole of the whole PR gate — the next-slowest check on that commit is `Cargo check & test` at `8m01s`.

### 1. Lane split — #2129, runs [32424851471](https://github.com/proliferate-ai/proliferate/actions/runs/32424851471) / [32425911757](https://github.com/proliferate-ai/proliferate/actions/runs/32425911757)

```
test-unit        | 1904 passed, 10 warnings in 174.19s (0:02:54)
test-integration | 1068 passed, 1 skipped, 890 warnings in 601.26s (0:10:01)
test-unit        | 1904 passed, 10 warnings in 184.27s (0:03:04)
test-integration | 1068 passed, 1 skipped, 890 warnings in 753.39s (0:12:33)
```

**Measured correction to the naive split.** `tests/unit` is not free of live infrastructure: 113 test functions across 20 files resolve the `test_engine` / `db_session` / `client` / `single_org_client` fixtures and open a real Postgres connection, and `tests/unit/test_redis_lock.py:108,:138` exercise a live Redis over `settings.redbeat_redis_url`. So the unit lane **keeps** the postgres and redis service containers. What it drops is the xdist fan-out and the `max_locks_per_transaction` bump — a serial single-connection run never needs either, and it still finishes in ~3m.

### 2. Before-only truncate — #2130, runs [32424923992](https://github.com/proliferate-ai/proliferate/actions/runs/32424923992) / [32425925280](https://github.com/proliferate-ai/proliferate/actions/runs/32425925280)

`test-integration` wall `9m09s` / `10m09s` against `10m52s` / `13m31s` for the same lanes without it. Every test already starts from a clean slate through the same `test_engine` call, so the post-test TRUNCATE only ever cleaned the database at process exit — which the job never reads back.

> [!NOTE]
> **Semantics change — ratified by Pablo on 2026-08-20.** This is a semantics change, not a pure optimization: tables now keep residue after each test instead of being wiped on teardown. Any future test that inspects the database *outside* a `test_engine`-scoped fixture, or any debugging workflow that relied on a clean database at process exit, sees different behavior. The ratified contract is **clean-at-start**: `test_engine` guarantees every test begins against a truncated database, and post-teardown residue is accepted. Write tests against that guarantee — do not rely on the database being clean *after* a test finishes.

### 3. Integration sharding — #2131, runs [32428452217](https://github.com/proliferate-ai/proliferate/actions/runs/32428452217) / [32429121218](https://github.com/proliferate-ai/proliferate/actions/runs/32429121218)

Deterministic file-level partition: all `tests/integration` test files plus `tests/e2e/cloud/test_e2b_webhooks.py`, sorted, `index % 3 == shard - 1`. Identical partition across both samples:

```
Total integration test files: 132   ->   Shard 1 owns 44 / Shard 2 owns 44 / Shard 3 owns 44
```

| Shard | Sample A ([32428452217](https://github.com/proliferate-ai/proliferate/actions/runs/32428452217)) | Sample B ([32429121218](https://github.com/proliferate-ai/proliferate/actions/runs/32429121218)) |
| --- | --- | --- |
| 1 | `286 passed, 1 skipped, 237 warnings in 233.23s (0:03:53)` | `286 passed, 1 skipped, 237 warnings in 216.77s (0:03:36)` |
| 2 | `397 passed, 261 warnings in 439.95s (0:07:19)` | `397 passed, 261 warnings in 176.12s (0:02:56)` |
| 3 | `385 passed, 389 warnings in 255.55s (0:04:15)` | `385 passed, 389 warnings in 255.68s (0:04:15)` |

Test counts are byte-identical across the two samples; only the wall times move. Shard 2's `7m19s`/`8m28s` in sample A is **runner variance, not partition imbalance** — the same 397 tests ran in `2m56s` in sample B on the identical partition.

Each shard keeps postgres + redis, the `max_locks_per_transaction = 1024` step, and `pytest -n 4`.

## Count reconciliation

```
test-unit                     1904 passed
test-integration shard 1       286 passed,  1 skipped
test-integration shard 2       397 passed
test-integration shard 3       385 passed
                              ---------------------
                              2972 passed,  1 skipped
main baseline (single job)    2972 passed,  1 skipped
```

No test is dropped, duplicated, or double-counted. Whole files stay together, so xdist distributes only *within* a shard and the partition depends solely on the sorted file list — never on collection or runtime order.

## Gate effect

The server pole drops from `16m15s` to the slowest integration shard, ~`4-5m` (`test-unit` ~`3m` runs alongside it). With server no longer the long pole, the PR gate becomes rust-bound at `Cargo check & test`, ~`7-8m`.

## Wiring

- `docker` → `needs: [lint, test-unit, test-integration]`
- `self-hosted-release-assets` → `needs: [lint, test-unit, test-integration, docker]`
- No other consumer of the old `test` job name exists. The reusable-workflow callers (`hotfix-production.yml`, `nightly-release-train.yml`) invoke `./.github/workflows/server-ci.yml` as a whole and never name inner jobs; `server-ci.yml` declares no `workflow_call` outputs; `deploy-staging.yml`'s "Wait for Server CI" polls the run's overall conclusion, not a job name. Verified there is no branch protection or ruleset pinning `test` as a required check.

## Files

- `.github/workflows/server-ci.yml`
- `server/tests/conftest.py`

## Validation on this PR

Server CI run [32436380473](https://github.com/proliferate-ai/proliferate/actions/runs/32436380473) at `4d9fffdd1`, all lanes green — a third sample with counts identical to the two experiment samples:

```
test-unit            | 1904 passed, 10 warnings in 123.92s (0:02:03)
test-integration (1) | 286 passed, 1 skipped, 237 warnings in 228.03s (0:03:48)
test-integration (2) | 397 passed, 261 warnings in 175.86s (0:02:55)
test-integration (3) | 385 passed, 389 warnings in 247.37s (0:04:07)
lint                 | All checks passed! / Mypy diagnostic ratchet passed (480 existing diagnostics).
```

All three shards again report `Total integration test files: 132` and `owns 44 files`. Job walls: `test-unit` 2m51s, shards 4m45s / 3m41s / 5m04s — the whole server lane now completes in ~5m against the `16m15s` baseline. `docker` and `self-hosted-release-assets` correctly resolved their new `needs` and skipped (input-driven gate, not a PR event).

Gate effect confirmed on this PR's head `4d9fffdd1`. Slowest checks, measured:

```
7m10  Cargo check & test        <- new pole
6m35  intent-tests (provisional)
5m19  intent-billing (provisional)
5m04  test-integration (3)      <- slowest server lane
4m45  test-integration (1)
```

Server is no longer the long pole. The gate is rust-bound at ~`7m`, down from the `16m15s` `test` job it replaces.

## Review hardening

Applied from review on `4d9fffdd1`, in `399a7f7a1`, scoped to the partition step only:

- **`LC_ALL=C sort -u`.** The shard a file lands in is its index in the sorted list mod 3, so collation is part of the partition contract. Reproduced the reviewer's proof against the real tree: under `en_US.UTF-8` rather than `LC_ALL=C`, **13 of the 132** discovered files land in a different shard. Both collations yield the identical *set*, so the three shards still union to the whole suite and the drift never surfaces as a failure — it just makes two runs incomparable while both stay green. Pinning `LC_ALL=C` is what makes reruns comparable rather than merely green. Shard sizes under `LC_ALL=C` remain 44/44/44.
- **Floor checks on the discovered list.** `mapfile -t files < <(...)` reads a process substitution, not a pipeline, so `set -o pipefail` cannot observe a failing `find`; a moved or renamed `tests/integration` tree would give a short or empty list and the shard would pass by testing almost nothing. Now fails loudly with the observed count if fewer than 100 files are discovered, or if a shard is assigned none.
- **`xargs -r`.** An empty `shard_files.txt` now runs nothing and fails, instead of invoking a bare `pytest` that would collect the entire suite.
- Comment at the partition step updated to state the dependency on the `LC_ALL=C`-sorted list and to explain why the process-substitution guard is needed.

## Follow-ups (not in this PR, to keep the diff minimal)

- `_cancel_test_background_tasks` in `server/tests/conftest.py` is an inert stub (`return None`) — flagged by both reviews. It is still called on the pre-test path, so it is not dead code, but it does nothing and should be removed in a separate cleanup rather than widening this diff.
- The `id: partition` on the partition step is unused; no step consumes its outputs.

### DIFF
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -270,8 +270,20 @@ jobs:
       # this shard's offset. Whole files stay together -- xdist distributes
       # within a shard -- so no test can land in two shards or in none, and
       # the partition is identical across workers, runs, and reruns because
-      # it depends only on the sorted file list, never on collection or
-      # runtime order.
+      # it depends only on the file list sorted under LC_ALL=C, never on
+      # collection or runtime order. Collation is pinned explicitly: the
+      # shard a file lands in is a function of its index in the sorted list,
+      # so a locale that ordered punctuation or case differently would move
+      # files between shards. The three shards would still union to the whole
+      # suite, so such a drift is invisible in the pass/fail signal -- pinning
+      # LC_ALL=C is what makes reruns comparable rather than merely green.
+      #
+      # `mapfile < <(...)` reads from a process substitution, which is not a
+      # pipeline, so `set -o pipefail` cannot observe a failing `find` and the
+      # step would happily continue with a short or empty list. The floor
+      # checks below are the actual guard: a rename or move under
+      # tests/integration that collapses the list fails the shard loudly
+      # instead of silently testing almost nothing.
       - name: Partition integration test files (shard ${{ matrix.shard }} of 3)
         id: partition
         env:
@@ -280,21 +292,32 @@ jobs:
           set -euo pipefail
           mapfile -t files < <(
             { find tests/integration -type f \( -name 'test_*.py' -o -name '*_test.py' \); \
-              echo tests/e2e/cloud/test_e2b_webhooks.py; } | sort -u
+              echo tests/e2e/cloud/test_e2b_webhooks.py; } | LC_ALL=C sort -u
           )
           echo "Total integration test files: ${#files[@]}"
+          if (( ${#files[@]} < 100 )); then
+            echo "Refusing to shard: discovered only ${#files[@]} integration test files (expected >= 100)." >&2
+            echo "The find over tests/integration collapsed -- check for a moved or renamed test tree." >&2
+            exit 1
+          fi
           shard_files=()
           for i in "${!files[@]}"; do
             if (( i % 3 == SHARD - 1 )); then
               shard_files+=("${files[$i]}")
             fi
           done
+          if (( ${#shard_files[@]} == 0 )); then
+            echo "Refusing to run: shard ${SHARD} of 3 was assigned no files out of ${#files[@]}." >&2
+            exit 1
+          fi
           echo "Shard ${SHARD} owns ${#shard_files[@]} files:"
           printf '  %s\n' "${shard_files[@]}"
           printf '%s\n' "${shard_files[@]}" > shard_files.txt
 
+      # -r so that an empty shard_files.txt runs nothing and fails, rather
+      # than invoking a bare `pytest` that would collect the entire suite.
       - name: Run integration tests (shard ${{ matrix.shard }} of 3)
-        run: xargs -a shard_files.txt pytest -n 4 -q --durations=20
+        run: xargs -r -a shard_files.txt pytest -n 4 -q --durations=20
 
   docker:
     runs-on: ubuntu-latest
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -183,6 +183,10 @@ jobs:
   test-integration:
     runs-on: ubuntu-latest
     timeout-minutes: 75
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2, 3]
     env:
       DATABASE_URL: postgresql+asyncpg://proliferate:localdev@127.0.0.1:5432/proliferate
       JWT_SECRET: test-jwt-secret
@@ -260,8 +264,37 @@ jobs:
           docker exec "$container" psql -U proliferate -d proliferate \
             -c 'SHOW max_locks_per_transaction;'
 
-      - name: Run integration tests
-        run: pytest tests/integration tests/e2e/cloud/test_e2b_webhooks.py -n 4 -q --durations=20
+      # Deterministic file-level partition: list every integration test file
+      # plus the one e2e file this lane also carries, sort them (independent
+      # of filesystem traversal order), and take every third file starting at
+      # this shard's offset. Whole files stay together -- xdist distributes
+      # within a shard -- so no test can land in two shards or in none, and
+      # the partition is identical across workers, runs, and reruns because
+      # it depends only on the sorted file list, never on collection or
+      # runtime order.
+      - name: Partition integration test files (shard ${{ matrix.shard }} of 3)
+        id: partition
+        env:
+          SHARD: ${{ matrix.shard }}
+        run: |
+          set -euo pipefail
+          mapfile -t files < <(
+            { find tests/integration -type f \( -name 'test_*.py' -o -name '*_test.py' \); \
+              echo tests/e2e/cloud/test_e2b_webhooks.py; } | sort -u
+          )
+          echo "Total integration test files: ${#files[@]}"
+          shard_files=()
+          for i in "${!files[@]}"; do
+            if (( i % 3 == SHARD - 1 )); then
+              shard_files+=("${files[$i]}")
+            fi
+          done
+          echo "Shard ${SHARD} owns ${#shard_files[@]} files:"
+          printf '  %s\n' "${shard_files[@]}"
+          printf '%s\n' "${shard_files[@]}" > shard_files.txt
+
+      - name: Run integration tests (shard ${{ matrix.shard }} of 3)
+        run: xargs -a shard_files.txt pytest -n 4 -q --durations=20
 
   docker:
     runs-on: ubuntu-latest
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -109,7 +109,78 @@ jobs:
           uv run --python 3.12 --frozen --extra dev python scripts/check_mypy_baseline.py
           --github-event-base
 
-  test:
+  # tests/unit is not actually free of live-infra dependencies: 113 test
+  # functions across 20 files resolve the test_engine/db_session/client/
+  # single_org_client fixtures (tests/conftest.py, tests/unit/test_first_run_claim.py:64,
+  # tests/unit/test_registration_page.py:54, tests/unit/test_self_registration.py:53, etc.)
+  # which open a real Postgres connection, and
+  # tests/unit/test_redis_lock.py:108 and :138 exercise a live Redis over
+  # settings.redbeat_redis_url. Moving those ~115 tests out of tests/unit or
+  # reworking the fixtures to fake the backends would be well past a minimal
+  # conftest change, so this lane keeps the service containers as the
+  # documented fallback (see PR body) -- only the xdist fan-out and the
+  # lock-table bump are dropped, since a serial, single-connection run never
+  # needs them.
+  test-unit:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    env:
+      DATABASE_URL: postgresql+asyncpg://proliferate:localdev@127.0.0.1:5432/proliferate
+      JWT_SECRET: test-jwt-secret
+      CLOUD_SECRET_KEY: test-cloud-secret
+    services:
+      postgres:
+        image: postgres:16-alpine
+        env:
+          POSTGRES_USER: proliferate
+          POSTGRES_PASSWORD: localdev
+          POSTGRES_DB: proliferate
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U proliferate -d proliferate"
+          --health-interval 5s
+          --health-timeout 3s
+          --health-retries 10
+      redis:
+        image: redis:7-alpine
+        ports:
+          - 6379:6379
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 5s
+          --health-timeout 3s
+          --health-retries 10
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ inputs.git_sha || github.ref }}
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: "22"
+      - uses: pnpm/action-setup@v6
+        with:
+          version: 10
+      - uses: astral-sh/setup-uv@v7
+      - uses: actions/setup-python@v7
+        with:
+          python-version: "3.12"
+
+      - name: Install workspace dependencies
+        run: pnpm install --frozen-lockfile
+        working-directory: .
+
+      - name: Build artifact runtime
+        run: pnpm -C server/artifact-runtime build
+        working-directory: .
+
+      - name: Install dependencies
+        run: uv pip install --system -e ".[dev]"
+
+      - name: Run unit tests
+        run: pytest tests/unit -q --durations=20
+
+  test-integration:
     runs-on: ubuntu-latest
     timeout-minutes: 75
     env:
@@ -189,12 +260,12 @@ jobs:
           docker exec "$container" psql -U proliferate -d proliferate \
             -c 'SHOW max_locks_per_transaction;'
 
-      - name: Run tests
-        run: pytest tests/unit tests/integration tests/e2e/cloud/test_e2b_webhooks.py -n 4 -q --durations=20
+      - name: Run integration tests
+        run: pytest tests/integration tests/e2e/cloud/test_e2b_webhooks.py -n 4 -q --durations=20
 
   docker:
     runs-on: ubuntu-latest
-    needs: [lint, test]
+    needs: [lint, test-unit, test-integration]
     # Reusable-workflow calls inherit the caller's event name (never
     # "workflow_call"), so this leg is driven by inputs.publish/dry_run instead.
     # inputs.* are empty on plain push/PR events, so the comparisons stay false
@@ -294,7 +365,7 @@ jobs:
 
   self-hosted-release-assets:
     runs-on: ubuntu-latest
-    needs: [lint, test, docker]
+    needs: [lint, test-unit, test-integration, docker]
     # Same input-driven gate as the docker job (see note above): a server-v tag
     # push, or an explicit publish call from a release train / hotfix workflow.
     if: (startsWith(github.ref, 'refs/tags/server-v')) || (inputs.publish == true && inputs.dry_run != true)
--- a/server/tests/conftest.py
+++ b/server/tests/conftest.py
@@ -63,9 +63,14 @@ async def test_engine(migrated_test_database):  # type: ignore[no-untyped-def]
     resolves through this fixture.
 
     Fixture ordering is unchanged: the pre-test truncate still runs before any
-    fixture that depends on ``test_engine`` can seed rows, and the post-test
-    truncate still runs after those fixtures tear down, because teardown is the
-    reverse of setup and this fixture is the first of the chain to be set up.
+    fixture that depends on ``test_engine`` can seed rows.
+
+    The truncate runs before each test only, not after: every test already
+    starts from a clean slate via this same call, so an after-test truncate
+    only ever cleaned the database at process exit -- which the CI job never
+    reads back. The Postgres service container is torn down with the runner,
+    and the session-scoped ``migrated_test_database`` fixture drops the whole
+    database unconditionally regardless of its contents.
 
     Tests that build their own throwaway database (``temporary_database``) are
     self-isolating and correctly opt out of the shared-database reset.
@@ -76,8 +81,6 @@ async def test_engine(migrated_test_database):  # type: ignore[no-untyped-def]
     try:
         yield engine
     finally:
-        await _cancel_test_background_tasks()
-        await truncate_all_tables(engine)
         await engine.dispose()
 
 


pablonyx (commented):
Independent review (Fable, main session — implementer was a separate agent).

**Verdict: sound. Two hardening nits, one observation, no blockers.**

What I verified directly on head `4d9fffd`:

- **Clean-at-start holds.** `test_engine` is function-scoped; setup still runs `_cancel_test_background_tasks()` + `truncate_all_tables(engine)` before `yield` (conftest.py:79-80). Teardown now only disposes the engine. Session end is covered by `migrated_test_database`, which drops the whole database unconditionally — so post-teardown residue is never observable by anything. The ratified semantics are implemented exactly as described.
- **xdist safety.** `server/tests/postgres.py:38` derives DB identity from `PYTEST_XDIST_WORKER`, so per-test truncates only touch the calling worker's database. `-n 4` within a shard is safe.
- **Partition is genuinely deterministic**: sorted full file list, round-robin by index, computed identically in each shard job. Whole files stay together; nothing can land in two shards or zero shards while the list has ≥3 files.
- **Wiring**: `docker` and `self-hosted-release-assets` `needs:` updated to the three new names; no other consumer of the old `test` job name exists.

Nits (fine as follow-ups, don't block):

1. `sort -u` in the partition step is locale-sensitive. All three shard jobs run the same runner image today so they agree, but `LC_ALL=C sort -u` makes the partition provably identical forever, not accidentally.
2. `xargs -a shard_files.txt pytest -n 4` with an *empty* file runs pytest with no path args — i.e. the entire suite. Unreachable while ≥3 test files exist, but `xargs -r` (or an explicit `[[ -s shard_files.txt ]]` guard) turns a silent worst-case into a loud failure.

Observation: `_cancel_test_background_tasks()` is a stub (`return None`, conftest.py:22-23). Keeping the pre-test call is harmless and this PR rightly doesn't touch it, but the function is dead weight — candidate for deletion in a cleanup PR.

Numbers requoted from run 32436380473: 1904 unit + 286/397/385 integration = 2972 passed, 1 skipped — exact match to main baseline run 32434848206. Third identical sample.

pablonyx (commented):
## Review — PR #2137 (fresh reviewer of record, did not implement)

Verdict: **approve-quality.** No correctness defect. Every claim in the body reconciles against raw job logs, and I independently re-derived the partition rather than trusting the printed counts. Two hardening items below (one medium, one nit) that I'd want in before this becomes the load-bearing shape for the server gate — neither blocks the change as written.

---

### 1. Literal CI requote (not from the body)

Baseline — run [32434848206](https://github.com/proliferate-ai/proliferate/actions/runs/32434848206) @ `56dd6ebf`, job `test` ([96633858410](https://github.com/proliferate-ai/proliferate/actions/runs/32434848206/job/96633858410)):

```
2972 passed, 1 skipped, 903 warnings in 920.17s (0:15:20)
```
Job wall `01:01:57Z → 01:18:12Z` = **16m15s**.

This PR — run [32436380473](https://github.com/proliferate-ai/proliferate/actions/runs/32436380473) @ `4d9fffdd1`, `conclusion: success`. Lines pulled from each job's raw log via `gh api .../actions/jobs/<id>/logs`:

| job | id | literal log line | wall |
| --- | --- | --- | --- |
| `test-unit` | 96638318298 | `1904 passed, 10 warnings in 123.92s (0:02:03)` | 2m51s |
| `test-integration (1)` | 96638318345 | `Total integration test files: 132` / `Shard 1 owns 44 files:` / `286 passed, 1 skipped, 237 warnings in 228.03s (0:03:48)` | 4m45s |
| `test-integration (2)` | 96638318446 | `Total integration test files: 132` / `Shard 2 owns 44 files:` / `397 passed, 261 warnings in 175.86s (0:02:55)` | 3m41s |
| `test-integration (3)` | 96638318466 | `Total integration test files: 132` / `Shard 3 owns 44 files:` / `385 passed, 389 warnings in 247.37s (0:04:07)` | 5m04s |

`lint` success 56s. `docker` and `self-hosted-release-assets` both `skipped` — they resolved the new `needs` and gated on `inputs.publish`, exactly as intended.

Reconciliation: `1904 + 286 + 397 + 385 = 2972 passed, 1 skipped` — byte-identical to the baseline `2972 passed, 1 skipped`. Confirmed.

Runner-minutes, which the body does not state: `171 + 285 + 221 + 304 = 981s` across four jobs vs `975s` for the single baseline job. **+0.6%** — the 3x wall-clock win costs essentially nothing in billed compute, because the four setup phases run concurrently with what used to be serial test time. That is a better result than the body claims for itself.

Gate effect confirmed on this head: `Cargo check & test` 7m10s is the new pole, slowest server lane `test-integration (3)` at 5m04s.

Branch is `behind_by: 1` — that commit is `docs(readme): restructure sections and trim content (#2132)`, touching `README.md` only. Receipts are not stale.

---

### 2. Partition — independently re-derived, not taken on trust

I extracted the 44-line file lists each shard *actually printed* out of the three raw job logs and reconciled them against a fresh `find` on the PR head:

```
shard 1 = 44 files, shard 2 = 44 files, shard 3 = 44 files
comm -12 (1,2) = 0    comm -12 (1,3) = 0    comm -12 (2,3) = 0
union (sort -u)      = 132
repo file list @ 4d9fffdd1 = 132
in repo but no shard = (empty)
in shard but not repo = (empty)
```

So: pairwise disjoint, total coverage, nothing orphaned, nothing double-run. The claim holds on the evidence, not just on the printed counts.

Orphan-lane check on the split itself: `pytest tests/unit` + `find tests/integration -name 'test_*.py' -o -name '*_test.py'` + the one e2e file exactly reproduces the old `pytest tests/unit tests/integration tests/e2e/cloud/test_e2b_webhooks.py` collection surface. `server/pyproject.toml` does not override `python_files`, so the `find` predicate is pytest's default pattern — the nine non-matching `.py` files under `tests/integration` (`conftest.py`, `__init__.py`, `billing_accounting_helpers.py`, `*_assertions.py`, `cloud_api_helpers.py`, `agent_gateway_topups_shared.py`, `background/__init__.py`) are helpers pytest would not collect either. No file lands in neither lane.

---

### 3. MEDIUM — `sort` collation is not pinned, and nothing reconciles the shards

`.github/workflows/server-ci.yml:283` — `... } | sort -u`, with the partition taken as `i % 3` at `:288`.

`sort` order is `LC_COLLATE`-dependent, and for *this exact file list* the two plausible collations disagree:

```
$ diff <(LC_ALL=C sort -u files) <(LC_ALL=en_US.UTF-8 sort -u files)
7,8c7,8
<   tests/integration/test_agent_gateway_enrollment.py
<   tests/integration/test_agent_gateway_enrollment_keys.py
---
>   tests/integration/test_agent_gateway_enrollment_keys.py
>   tests/integration/test_agent_gateway_enrollment.py
...

files that land in a DIFFERENT shard under the two collations: 13 of 132
  tests/integration/test_agent_gateway_enrollment.py        shard 1 -> 2
  tests/integration/test_agent_gateway_enrollment_keys.py   shard 2 -> 1
  tests/integration/test_billing_accounting.py              shard 3 -> 2
  tests/integration/test_billing_accounting_boundaries.py   shard 1 -> 3
  ...
```

Today this is harmless — all three shards are the same `ubuntu-latest` image, so the collation is uniform, and I proved above that the actual run partitioned cleanly. But the comment at `:272` — *"the partition is identical across workers, runs, and reruns because it depends only on the sorted file list, never on collection or runtime order"* — is **incomplete as written**. It also depends on the collation, and that comment exists specifically to carry the safety argument. If the shards ever disagree (runner image locale change, a step that exports `LANG`, a self-hosted runner, or a developer reproducing "which shard owns file X" locally on macOS), files get double-run *or silently never run*, and CI stays green either way.

The same silence covers a second path: `mapfile -t files < <( find ... )` at `:281` uses process substitution, so `pipefail` does not see a `find` failure. Rename or move `tests/integration` and `files` collapses to the single echoed e2e path — shards 2 and 3 then own zero files (see nit below), and the run does not fail.

Two one-line fixes make this loud instead of silent:

```bash
mapfile -t files < <(
  { find tests/integration -type f \( -name 'test_*.py' -o -name '*_test.py' \); \
    echo tests/e2e/cloud/test_e2b_webhooks.py; } | LC_ALL=C sort -u
)
(( ${#files[@]} >= 100 )) || { echo "::error::integration file list collapsed to ${#files[@]}"; exit 1; }
...
(( ${#shard_files[@]} > 0 )) || { echo "::error::shard ${SHARD} owns 0 files"; exit 1; }
```

`LC_ALL=C` pins the ordering; the two assertions turn "tests silently vanished" into a red job. For a change whose whole purpose is to fan tests across jobs, a guard that the fan-out is non-empty is worth its two lines.

### 4. NIT — empty shard falls back to running the whole suite

`.github/workflows/server-ci.yml:297` — `xargs -a shard_files.txt pytest -n 4 -q --durations=20`.

GNU `xargs` without `-r` runs the command **once with no arguments** on empty input. `printf '%s\n' "${shard_files[@]}"` on an empty array writes a single blank line, so an empty shard would invoke bare `pytest -n 4 -q --durations=20`, which under `testpaths = ["tests"]` (`server/pyproject.toml`) collects the *entire* suite — including `tests/e2e/cloud/test_provisioning.py`, which was never in the old command and merely skips via `require_live_cloud` (`tests/e2e/cloud/conftest.py:24`). Result: green, slow, and wrong. Unreachable at 132 files; `xargs -r -a` or the `> 0` guard above closes it for free. Exit-code propagation is fine (`xargs` returns 123 on a failing child, and the step shell is `bash -e`).

---

### 5. `conftest.py` — before-only truncate matches the ratified semantics

Confirmed clean-at-start still holds for **every** path that can reach the shared test database, and I checked this by closing the surface rather than by reading the docstring:

- `truncate_all_tables` runs at `server/tests/conftest.py:80`, in `test_engine` setup, before any dependent fixture can seed.
- `grep -rn "TEST_DATABASE_URL" server/tests/` returns hits in `tests/postgres.py` (the definition) and **nowhere else** — `tests/conftest.py:78`'s `create_async_engine(TEST_DATABASE_URL)` is the only consumer. There is no second door into the shared DB, so there is no test that could observe post-teardown residue.
- The ~30 `create_async_engine` call sites across `tests/integration/test_*_migration.py` / `test_schema_migrations.py` all build throwaway DBs via `temporary_database`, which drops them in its own `finally`. Unaffected.
- The app's global `proliferate.db.engine` points at `DATABASE_URL` (`.../proliferate`), a different database that was never truncated before or after. Unaffected.
- No session-scoped fixture, module teardown, or finalizer asserts post-test cleanliness. `migrated_test_database` (`:45-50`) calls `drop_database` unconditionally.
- Under xdist each worker owns `proliferate_test_{gwN}_{pid}` (`tests/postgres.py:41-44`), so residue is per-worker and dropped with the DB.

**`_cancel_test_background_tasks` is still invoked** — `conftest.py:79`, moved from teardown into setup ahead of the truncate. Semantically equivalent (the previous test's tasks are cancelled before the next test's truncate). Worth noting for the record that the function is an inert stub — `conftest.py:22-23` is `async def _cancel_test_background_tasks() -> None: return None` — so neither call site does anything today. Pre-existing, not introduced here, but it means the diff's removal is provably a no-op and the surviving call is a placeholder someone should either implement or delete.

### 6. xdist safety

No new exposure. The merged `-n 4` from #2122 already distributed **per test** via the default `--dist load`, so tests within one file could already land on different workers; file-granularity sharding is strictly *coarser* isolation than what already shipped. The known #2122 class — `uuid4()` evaluated at collection time landing in parametrize IDs — is already fixed and documented at `server/tests/unit/test_sentry_transport_privacy.py:181`, and a sweep of `parametrize` decorators across `tests/unit` and `tests/integration` for `random`/`uuid4`/`time()`/`now()` turns up no other instance. That file is in the unit lane, which now runs **serially** (no `-n`), so it carries zero xdist ID-agreement risk. Direction of travel is safer, not riskier. No shared external state crosses shards: Postgres is per-worker-DB, and each shard gets its own service containers.

### 7. Wiring and metadata

- `needs:` updated at `:301` (`docker`) and `:401` (`self-hosted-release-assets`) — both observed `skipped` (not `skipped` due to a broken dependency) in the run.
- Grepped `.github`, `scripts`, `docs`, `specs` for `server-ci` consumers: `hotfix-production.yml:234` and `nightly-release-train.yml:226` call the workflow whole; `deploy-staging.yml:101-130` polls run-level `conclusion` via `gh run list --workflow server-ci.yml`, never a job name; `server-ci.yml` declares no `workflow_call` outputs. No consumer of the old `test` job name exists. Independently confirmed the body's branch-protection claim: `GET /branches/main/protection` → `404 Branch not protected`, `GET /rulesets` → `[]`. Nothing pins `test` as a required check.
- Title `perf(server): split CI lanes and shard integration tests 3-way` matches `<type>(<scope>): <change>`.
- Labels `release:skip`, `area:server`, `area:release` — correct per `scripts/ci-cd/pr-metadata.mjs` `AREA_RULES`: `.github/**` → `area:release`, `server/**` → `area:server`. Both changed paths covered, exactly one `release:*`.

### 8. Merge readiness

PR is still **draft**. Per repo policy the merge itself is Pablo's. Nothing else outstanding from my side.

pablonyx (commented):
## Final verification — head `399a7f7a1` — **APPROVE**

Re-verified the new head independently. The hardening is exactly what I prescribed, nothing rode along with it, and the fresh run reconciles. Upgrading to a clean approve.

### Diff `4d9fffdd1 → 399a7f7a1` is exactly the prescribed fix

One commit, `ci(server): pin partition collation and guard the shard file list`, one file, `+27 / -4`, all inside the partition step:

- `:283` → `| LC_ALL=C sort -u` — collation pinned.
- `:296-300` → floor guard `(( ${#files[@]} < 100 ))` → `exit 1`, message carries the observed count and names the likely cause (moved/renamed test tree). Placed after the `Total integration test files` echo and before the partition loop — correct position.
- `:308-311` → `(( ${#shard_files[@]} == 0 ))` → `exit 1`, message carries both the shard number and the total. Placed after the loop, before the printout — correct position.
- `:320` → `xargs -r -a`, with a comment at `:318-319` stating why. Belt-and-braces given the zero-shard guard already exits; correct as defense in depth.
- `:272-286` → the safety comment amended to say the partition depends on the file list *sorted under `LC_ALL=C`*, and a new paragraph spells out the `mapfile < <(...)` / process-substitution / `pipefail` gap and names the floor checks as the actual guard.

Diff against `main` is still exactly two files (`.github/workflows/server-ci.yml`, `server/tests/conftest.py`). `conftest.py` is untouched since my first pass, so section 5 of my earlier review stands unchanged. `behind_by: 1` is still only `docs(readme) (#2132)` touching `README.md`.

### Fresh run — literal requote

[Run 32437426930](https://github.com/proliferate-ai/proliferate/actions/runs/32437426930) @ `399a7f7a1`, `conclusion: success`. Lines from raw job logs:

| job | id | literal log line | wall |
| --- | --- | --- | --- |
| `test-unit` | 96641353124 | `1904 passed, 10 warnings in 127.63s (0:02:07)` | 3m07s |
| `test-integration (1)` | 96641353310 | `Total integration test files: 132` / `Shard 1 owns 44 files:` / `286 passed, 1 skipped, 237 warnings in 235.04s (0:03:55)` | 4m58s |
| `test-integration (2)` | 96641353399 | `Total integration test files: 132` / `Shard 2 owns 44 files:` / `397 passed, 261 warnings in 189.15s (0:03:09)` | 4m07s |
| `test-integration (3)` | 96641353340 | `Total integration test files: 132` / `Shard 3 owns 44 files:` / `385 passed, 389 warnings in 248.18s (0:04:08)` | 4m53s |

`lint` success 1m00s. `docker` / `self-hosted-release-assets` `skipped` on the input gate. `LC_ALL=C sort -u` confirmed present in all three shards' echoed step bodies.

`1904 + 286 + 397 + 385 = 2972 passed, 1 skipped` — identical to the [baseline](https://github.com/proliferate-ai/proliferate/actions/runs/32434848206) `2972 passed, 1 skipped, 903 warnings in 920.17s (0:15:20)` and to my previously reviewed run. Third consecutive sample at the same counts.

Full PR check surface: **33 pass / 2 skipping / 0 fail**; the two skips are `docker` and `self-hosted-release-assets`, both input-gated. Server lane wall 01:45:33Z → 01:50:31Z = **4m58s** against the `16m15s` job it replaces.

### Partition re-derived on the new head — and the pin now does what it claims

Extracted the printed 44-file lists from all three new job logs:

```
shard 1 = 44, shard 2 = 44, shard 3 = 44
pairwise overlap (1,2) = 0   (1,3) = 0   (2,3) = 0
union (sort -u) = 132
```

Two further checks that matter more than the counts:

1. **The partition is byte-identical to the pre-fix run** — `diff` of each shard's sorted list against the corresponding shard from run 32436380473 is empty for all three. So the GitHub runner was already C-collating, and the pin changes no behavior today. That confirms my finding was latent fragility rather than a live bug, and it means the fix carries zero re-validation risk: the same 44 files ran in the same shard before and after.

2. **The CI partition now exactly matches a locally computed `LC_ALL=C sort` partition** — I recomputed `i % 3` over `LC_ALL=C sort -u` of the file list on this head and it reproduces all three shards exactly (`44 == 44`, set-equal, for each). That is the property the pin actually buys: "which shard owns file X" is now answerable off-runner, including on macOS, where the default UTF-8 collation would previously have given the wrong answer for 13 of the 132 files. Reruns are now comparable rather than merely green, which is precisely what the amended `:277` comment claims.

The floor of `>= 100` against 132 actual leaves sensible headroom — loud on a collapse, quiet on ordinary attrition.

### Verdict

**Approve.** Both of my findings are closed at the source, the guards fail loudly with diagnostic counts rather than silently, the comment block no longer overstates the determinism argument, and the test-count reconciliation is unchanged across three independent samples. Sections 5 (before-only truncate matches the ratified semantics; `_cancel_test_background_tasks` still invoked at `conftest.py:79`), 6 (xdist), and 7 (wiring, no consumer of the old `test` job name, no branch protection or ruleset pinning it, title and labels conform) of my earlier review are unaffected and stand.

PR is out of draft and `MERGEABLE`. Merge is Pablo's call per repo policy — I am not merging.

PR #2138, opened 2026-08-21
ci(workflows): add ci-ok rollup gate that fails on skipped lanes

pablonyx:
## Why

Branch protection cannot distinguish **"no red X"** from **"actually verified."**

A required status check is satisfied when it reports `success` — and also when it reports `skipped`, and also when the name stops resolving because the job behind it was renamed, deleted, or gated off. A required-checks list is a list of *names*, and a name that stops resolving degrades silently to green.

Four ways CI can certify a commit it never verified:

1. A job is deleted from a workflow while its name stays in the required list.
2. A job is renamed — the frontend sharding renamed `Shared frontend packages` into four shard checks, and #2137 renamed the server `test` job into `test-unit` + `test-integration (1..3)`.
3. A job's `if:` quietly evaluates false and reports `skipped`.
4. A whole workflow is path-skipped and never reports at all.

**Greenfield:** `main` has **no branch protection and no rulesets** today. Nothing is enforced, so the list below is the first list and there is no existing configuration to reconcile.

## The design: one rollup per workflow

`needs:` cannot cross workflows, so each workflow that carries required checks needs its own rollup. This PR adds two.

**`ci-ok`** (`.github/workflows/ci.yml`) and **`server-ci-ok`** (`.github/workflows/server-ci.yml`) each:

- `needs:` **every** lane in their own workflow — 14 and 4 respectively.
- Run `if: always()`. Load-bearing: without it the job inherits the implicit `success()` gate, gets skipped whenever a dependency fails, and then reports the exact `skipped` conclusion branch protection accepts.
- Fail unless **every** `needs.*.result == 'success'`. `failure`, `skipped`, and `cancelled` are all fatal, and the failure output names the offending lanes.
- Carry a **drift guard** that parses their own workflow file and fails if a job exists there but not in the `needs:` list. Without it, adding a lane and forgetting the `needs:` entry reopens the same hole one level up.

Cost: sparse checkout of one file, then preinstalled `bash` / `jq` / `ruby`. No toolchain, no install, no build. **`ci-ok` ran in 6 seconds.**

A `needs:` entry on a matrix job covers **every shard** — `shared-frontend-packages` (4 shards) and `test-integration` (3 shards) each appear once, and the aggregate is `success` only when all shards succeeded. So sharding or renaming a lane never needs a branch-protection edit.

`server-ci-ok`'s drift guard has to exempt `docker` and `self-hosted-release-assets`, which are release-only and `skipped` on every PR. The exemption is **derived, not hardcoded**: it selects jobs whose job-level `if:` gates on the `server-v` tag, and prints the exempt set on every run. If anyone removes that release gate, the job stops being exempt and the drift guard demands it join `needs:`.

### Why `skipped` counts as failed

A skipped lane verified nothing. If skipping is a legitimate outcome, that decision belongs in the workflow — visible in the diff — not in a rollup that silently accepts absence.

## Server CI: always triggers, always reports green or red

Server CI was the one workflow that could not be required at all: its `on.pull_request.paths` filter meant it **never reported a conclusion** on a non-server PR — not `skipped`, *nothing*. Marking its lanes required would have stranded every non-server PR on a Pending check.

- **`on.pull_request.paths` dropped.** Server CI triggers on every PR.
- **A cheap `changes` job** diffs the merge base against the *same path list, verbatim*, and publishes one `relevant` flag. It ran in **10 seconds**.
- **Every step** of `lint`, `test-unit`, and `test-integration` is gated, with a leading no-op step that reports a real green.

The lanes are **not** job-level `if:`-gated on relevance. That would produce `skipped`. Instead each lane carries `if: always()` and the gate polarity is inverted: real steps fire on `relevant != 'false'`, the no-op step fires on `== 'false'`. An unset or empty output — exactly what a **failed or cancelled `changes` job** produces — runs the real suite. The gate is fail-closed by construction, not because `Detect server-relevant changes` happens to be on a list.

`test-integration`'s partition step is gated like every other step, so its two floor guards (the `>= 100` file-count check and the empty-shard check) cannot fire on a no-op run where no checkout happened.

The set of PRs whose server lanes actually execute is unchanged. `push`, tags, `workflow_dispatch`, and `workflow_call` are unconditionally relevant.

## Required status checks for `main`

Committed to the repo (see F5 below), not just this description:

| Check | Workflow |
|---|---|
| `ci-ok` | CI |
| `server-ci-ok` | Server CI |
| `Analyze (javascript-typescript)` | CodeQL |
| `Analyze (python)` | CodeQL |
| `Analyze (rust)` | CodeQL |
| `Validate PR title and labels` | PR Metadata |
| `Detect smoke-relevant changes` | Self-Host Smoke |
| `Production compose smoke` | Self-Host Smoke |

No individual lane name from `ci.yml` or `server-ci.yml` belongs on the list — the rollups are strictly stronger, and a per-lane name is a name that can rot. This list is already final: it survives #2137's split, the frontend sharding, and any future shard-count change without an edit.

**Deliberately NOT required:** `intent-tests (provisional)` / `intent-billing (provisional)` (`continue-on-error` by design); `docker` / `self-hosted-release-assets` (release-only, `skipped` on every PR); `Vercel` / `Vercel Preview Comments` (third-party).

## Review findings — all folded into this pass

Reviewer of record requested changes on F1 + F2 at head `393166f1`; everything below is addressed in this rebase.

- **F1 (MEDIUM, fail-open)** — `done < <(git diff ...)`. A process substitution is not a pipeline, so neither `set -e` nor `set -o pipefail` could observe a failing `git diff`: the loop read zero lines, `relevant` kept its `false` initializer, and both server lanes would have reported a **real green on a real server PR**. Now captured with command substitution (which `set -e` *does* cover), plus a floor guard in #2137's style: a PR always changes at least one file, so an empty list fails loudly instead of concluding "nothing server-relevant".
- **F2 (MEDIUM, doctrine)** — polarity flipped as described above. A failed or cancelled `changes` job can no longer turn a lane into `skipped`.
- **F3 (LOW)** — a non-empty `inputs.git_sha` is now unconditionally relevant, so a future PR-triggered `workflow_call` cannot compute release relevance from a PR diff. (Both current callers are non-PR; verified.)
- **F4 (LOW)** — `server-ci-ok` added, per coordinator decision. This is also the answer to the reviewer's sharper framing of the services cost: with rollups as the required names, the required list is split-proof. The unconditional `services:` boot is **knowingly accepted** and annotated in the file — Actions has no conditional `services:`, and the alternative (job-level skip laundered into one green) reinstates `skipped` as a load-bearing conclusion in the very workflow this PR is de-skipping. If it proves flaky, the honest fix is self-provisioned Postgres/Redis behind a gated step, not the skip.
- **F5 (LOW, durability)** — the list is now checked in twice: a comment block above `ci-ok` in `ci.yml`, and a new `## Merge Gate` section in `specs/codebase/systems/engineering/delivery/README.md`. Both say that a change to the required set belongs in both places in the same commit. The README section is appended at the file tail; #2140's last hunk in that file ends at original line 318 of 324, so there is no hunk overlap and no conflict with its in-flight edits.
- **F6 (NIT)** — the empty-needs guard now carries a comment saying it is belt-and-braces, not an independent safety property.
- **F7 (NIT)** — the drift guards feature-detect `YAML.unsafe_load_file` so a YAML anchor cannot red the gate with a `Psych::AliasesNotEnabled` error nobody would connect to an anchor. Feature-detected rather than passing `aliases: true`, because that keyword does not exist on Psych 3 — caught by running the snippet on Ruby 2.6 locally.

## Test plan

- **Both rollups green on this PR's own run**, literal logs quoted in a comment below.
- Drift guards run on every execution. Simulated locally against the real files at this head: `ci-ok` → `missing=[] extra=[] n=14`; `server-ci-ok` → `exempt: docker, self-hosted-release-assets`, `missing=[] extra=[] n=4`.
- **How a skipped lane flips a rollup red** (negative control, verifiable without merging): add `if: false` to any lane. It reports `skipped`, a per-lane required check would treat that as satisfied, and the rollup fails with `::error::CI did not verify this commit. Not green: <job> (skipped)`. Same path covers cancelled and failed.
- **Server CI runs for real here:** this PR edits `server-ci.yml`, which is on the relevance list, so `relevant=true` and `lint` / `test-unit` / `test-integration (1..3)` all execute in full.
- **`changes` classification** exercised locally against synthetic diffs: `server/proliferate/app.py` → true, `server/a/b/c.py` → true, `catalogs/x/y.json` → true, `.github/workflows/server-ci.yml` → true, docs-only → false, and the decoys `myserver/a.py` and `docs/server/a.md` → false. Empty list → floor guard exit 1. A failing command substitution aborts under `set -e` (F1 confirmed fixed).
- **Rollup shell logic** exercised against synthetic `toJSON(needs)` payloads: all-success (exit 0), mixed `skipped` + `cancelled` (exit 1, both named), empty needs (exit 1), job-list drift (exit 1).
- A `working-directory: .` audit over every `run` step in `server-ci.yml` after `server-ci-ok`'s first run failed on the sparse checkout (`defaults.run.working-directory: server`); all six steps that need it carry it.
- Repo guards run locally: `check_docs.py`, `check_max_lines.py`, `lint_records.py`, `node --test scripts/ci-cd/*.test.mjs` (0 fail), and a ruby YAML parse over all 29 workflows.

**Labels:** the delivery README section makes `area:docs` required in addition to `area:release` (`deriveAreaExpectation` returns both); both are applied.

**Note on history:** the branch was moved forward by merging `main` (commit `14b1a4a34`) rather than rebased, to avoid a force-push. The diff against `main` is three files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -702,3 +702,111 @@ jobs:
           name: workflow-definition-lifecycle-traces
           path: tests/intent/test-results/
           retention-days: 7
+
+  # ---------------------------------------------------------------------------
+  # Absence gate.
+  #
+  # A per-lane required check answers "did this lane go red?". It cannot answer
+  # "did this lane run at all?": branch protection treats a check that never
+  # reports a conclusion, or reports `skipped`, as satisfied. So a lane deleted
+  # from this file, an `if:` that quietly evaluated false, a cancelled run, or a
+  # required-check name that no longer matches any job all read as green. "No
+  # red X" and "actually verified" are indistinguishable.
+  #
+  # This job makes them distinguishable. It `needs:` every other job in this
+  # workflow, runs with `if: always()` so it reports even when dependencies
+  # fail, and succeeds ONLY when every dependency concluded `success`. failure,
+  # skipped, and cancelled are all fatal, and the offending lanes are named in
+  # the failure output. Make `ci-ok` the required check and absence becomes red
+  # instead of invisible.
+  #
+  # `if: always()` is load-bearing: without it this job inherits the implicit
+  # `success()` gate, would be SKIPPED whenever a dependency failed, and would
+  # then report the very `skipped` conclusion branch protection accepts.
+  #
+  # A `needs:` entry on a matrix job (shared-frontend-packages) covers every
+  # shard -- the aggregate result is `success` only if all shards succeeded.
+  #
+  # No language toolchain, no dependency install, no build: a sparse checkout of
+  # this one file plus preinstalled bash/jq/ruby. Seconds of runner time.
+  #
+  # `needs:` cannot cross workflows, so this rolls up ci.yml ONLY. Server CI,
+  # CodeQL, PR Metadata, Self-Host Smoke, and Intent Tests are separate
+  # workflows and still need their own entries in the required-checks list.
+  ci-ok:
+    name: ci-ok
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    if: always()
+    needs:
+      - repo-shape
+      - terraform-validate
+      - ci-cd-config
+      - candidate-build-handoff
+      - cargo-check
+      - sdk-build
+      - desktop-frontend
+      - mobile-typecheck
+      - web-frontend
+      - login-budget
+      - shared-frontend-checks
+      - shared-frontend-packages
+      - scroll-physics
+      - workflow-definition-lifecycle
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          sparse-checkout: .github/workflows/ci.yml
+          sparse-checkout-cone-mode: false
+
+      # Guards the guard. A job added to this file but not to `needs:` above
+      # would be silently outside the rollup, which is the same absence hole one
+      # level up. Compare the parsed job list against the resolved `needs` keys
+      # and fail on any drift in either direction.
+      - name: Assert the rollup covers every job in ci.yml
+        env:
+          NEEDS_JSON: ${{ toJSON(needs) }}
+        run: |
+          set -euo pipefail
+          declared="$(ruby -ryaml -rjson -e '
+            jobs = YAML.load_file(".github/workflows/ci.yml")["jobs"].keys - ["ci-ok"]
+            puts JSON.generate(jobs.sort)
+          ')"
+          covered="$(jq -c 'keys' <<<"$NEEDS_JSON")"
+          missing="$(jq -rn --argjson a "$declared" --argjson b "$covered" '($a - $b) | join(", ")')"
+          extra="$(jq -rn --argjson a "$declared" --argjson b "$covered" '($b - $a) | join(", ")')"
+          if [[ -n "$missing" ]]; then
+            echo "::error::ci.yml job(s) missing from the ci-ok needs list: ${missing}. Add them, or the rollup certifies a commit those lanes never verified."
+          fi
+          if [[ -n "$extra" ]]; then
+            echo "::error::ci-ok needs entries that are not jobs in ci.yml: ${extra}."
+          fi
+          [[ -z "$missing" && -z "$extra" ]]
+          echo "ci-ok covers all $(jq -r 'length' <<<"$declared") job(s) in ci.yml."
+
+      # skipped and cancelled are failures here, on purpose: a lane that did not
+      # run did not verify anything, and this check exists to say so out loud.
+      - name: Assert every CI lane concluded success
+        env:
+          NEEDS_JSON: ${{ toJSON(needs) }}
+        run: |
+          set -euo pipefail
+          bad=()
+          count=0
+          while IFS=$'\t' read -r lane result; do
+            count=$((count + 1))
+            printf '%-34s %s\n' "$lane" "$result"
+            if [[ "$result" != "success" ]]; then
+              bad+=("${lane} (${result})")
+            fi
+          done < <(jq -r 'to_entries[] | [.key, .value.result] | @tsv' <<<"$NEEDS_JSON")
+
+          if [[ "$count" -eq 0 ]]; then
+            echo "::error::ci-ok resolved zero dependencies. The needs list is empty or unresolvable; CI verified nothing."
+            exit 1
+          fi
+          if [[ "${#bad[@]}" -gt 0 ]]; then
+            echo "::error::CI did not verify this commit. Not green: ${bad[*]}"
+            exit 1
+          fi
+          echo "All ${count} CI lanes concluded success."
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -16,12 +16,16 @@ on:
   # reports a conclusion at all, so a required check naming these jobs would sit
   # Pending forever on every non-server PR and make it unmergeable -- and the
   # only way to avoid that today is to leave the server lanes off the required
-  # list entirely, which is the absence hole ci-ok exists to close. So the
-  # workflow now runs on every PR and the cheap `changes` job below diffs the
-  # merge base against the same path list, gating the heavy lanes. A gated-off
-  # lane still reports (a `skipped` conclusion, which branch protection treats
-  # as satisfied), so `lint` and `test` are always safe to mark required. Same
-  # shape as self-host-smoke.yml's changes/smoke pair.
+  # list entirely, which is exactly the absence hole ci-ok exists to close.
+  #
+  # So the workflow runs on EVERY pull request, and the cheap `changes` job
+  # below diffs the merge base against the same path list that used to sit on
+  # this trigger. On a non-server diff the heavy lanes short-circuit and report
+  # SUCCESS in well under a minute rather than being job-level `if:`-skipped.
+  # That distinction is the whole point: `skipped` is a conclusion branch
+  # protection silently accepts, so an absence-is-fatal required check must
+  # never be satisfied by one. `lint` and `test` now always report a real
+  # green/red, on every PR, with no path-filter trust anywhere in the chain.
   pull_request:
   workflow_dispatch:
     inputs:
@@ -68,6 +72,15 @@ jobs:
   #
   # Every non-pull_request caller -- main/tag pushes, workflow_dispatch, and the
   # release/hotfix workflow_call entrypoints -- is unconditionally relevant.
+  #
+  # This job is lane-agnostic on purpose: it publishes ONE `relevant` flag, and
+  # any number of downstream lanes consume it. When the server lane split lands
+  # (#2137 splits `test` into `test-unit` + a 3-way-sharded `test-integration`),
+  # each new lane needs the same two things and nothing more: `needs: changes`,
+  # and `if: needs.changes.outputs.relevant == 'true'` on every one of its
+  # steps, plus the green no-op step. Do not reintroduce a paths filter, and do
+  # not gate a lane at the job level -- that produces `skipped`, which is the
+  # conclusion this whole arrangement exists to avoid.
   changes:
     name: Detect server-relevant changes
     runs-on: ubuntu-latest
@@ -112,47 +125,69 @@ jobs:
 
   lint:
     needs: changes
-    if: needs.changes.outputs.relevant == 'true'
     runs-on: ubuntu-latest
     steps:
+      # Reports SUCCESS, not `skipped`, when the diff cannot change this lane's
+      # result. `skipped` is what branch protection silently accepts, so it is
+      # exactly the conclusion an absence-is-fatal required check must never
+      # rely on; every step below is individually gated instead, and the job
+      # ends green in well under a minute.
+      - name: No server-relevant changes
+        if: needs.changes.outputs.relevant != 'true'
+        working-directory: .
+        run: |
+          echo "No server-relevant paths in this diff; nothing for this lane to verify."
+          echo "Reporting success (not skipped) so this check always reports a conclusion."
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        if: needs.changes.outputs.relevant == 'true'
         with:
           ref: ${{ inputs.git_sha || github.ref }}
           fetch-depth: 2
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        if: needs.changes.outputs.relevant == 'true'
         with:
           node-version: "22"
       - uses: pnpm/action-setup@v6
+        if: needs.changes.outputs.relevant == 'true'
         with:
           version: 10
       - uses: astral-sh/setup-uv@v7
+        if: needs.changes.outputs.relevant == 'true'
       - uses: actions/setup-python@v7
+        if: needs.changes.outputs.relevant == 'true'
         with:
           python-version: "3.12"
 
       - name: Validate catalogs
+        if: needs.changes.outputs.relevant == 'true'
         run: |
           node scripts/validate-agent-catalog.mjs
         working-directory: .
 
       - name: Install workspace dependencies
+        if: needs.changes.outputs.relevant == 'true'
         run: pnpm install --frozen-lockfile
         working-directory: .
 
       - name: Build artifact runtime
+        if: needs.changes.outputs.relevant == 'true'
         run: pnpm -C server/artifact-runtime build
         working-directory: .
 
       - name: Install dependencies
+        if: needs.changes.outputs.relevant == 'true'
         run: uv sync --python 3.12 --frozen --extra dev
 
       - name: Ruff check
+        if: needs.changes.outputs.relevant == 'true'
         run: uv run --python 3.12 --frozen --extra dev ruff check proliferate/ tests/
 
       - name: Ruff format check
+        if: needs.changes.outputs.relevant == 'true'
         run: uv run --python 3.12 --frozen --extra dev ruff format --check proliferate/ tests/
 
       - name: Mypy diagnostic ratchet
+        if: needs.changes.outputs.relevant == 'true'
         env:
           MYPY_COMPARISON_SHA: ${{ inputs.comparison_sha || '' }}
           MYPY_EVENT_NAME: ${{ github.event_name }}
@@ -163,9 +198,12 @@ jobs:
           uv run --python 3.12 --frozen --extra dev python scripts/check_mypy_baseline.py
           --github-event-base
 
+  # The `services:` block below has no conditional form in Actions, so Postgres
+  # and Redis boot even on a no-op run. Both are alpine images in the runner's
+  # warm cache; the resulting green no-op still lands in well under a minute,
+  # which is the price of a check that always reports a real conclusion.
   test:
     needs: changes
-    if: needs.changes.outputs.relevant == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 75
     env:
@@ -196,29 +234,49 @@ jobs:
           --health-timeout 3s
           --health-retries 10
     steps:
+      # Reports SUCCESS, not `skipped`, when the diff cannot change this lane's
+      # result. `skipped` is what branch protection silently accepts, so it is
+      # exactly the conclusion an absence-is-fatal required check must never
+      # rely on; every step below is individually gated instead, and the job
+      # ends green in well under a minute.
+      - name: No server-relevant changes
+        if: needs.changes.outputs.relevant != 'true'
+        working-directory: .
+        run: |
+          echo "No server-relevant paths in this diff; nothing for this lane to verify."
+          echo "Reporting success (not skipped) so this check always reports a conclusion."
+
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        if: needs.changes.outputs.relevant == 'true'
         with:
           ref: ${{ inputs.git_sha || github.ref }}
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        if: needs.changes.outputs.relevant == 'true'
         with:
           node-version: "22"
       - uses: pnpm/action-setup@v6
+        if: needs.changes.outputs.relevant == 'true'
         with:
           version: 10
       - uses: astral-sh/setup-uv@v7
+        if: needs.changes.outputs.relevant == 'true'
       - uses: actions/setup-python@v7
+        if: needs.changes.outputs.relevant == 'true'
         with:
           python-version: "3.12"
 
       - name: Install workspace dependencies
+        if: needs.changes.outputs.relevant == 'true'
         run: pnpm install --frozen-lockfile
         working-directory: .
 
       - name: Build artifact runtime
+        if: needs.changes.outputs.relevant == 'true'
         run: pnpm -C server/artifact-runtime build
         working-directory: .
 
       - name: Install dependencies
+        if: needs.changes.outputs.relevant == 'true'
         run: uv pip install --system -e ".[dev]"
 
       # Four workers means four concurrent 78-table TRUNCATEs and four
@@ -229,6 +287,7 @@ jobs:
       # postmaster-level, and service containers take no command line, so it has
       # to be written to postgresql.auto.conf and the container bounced.
       - name: Raise the Postgres lock table for parallel workers
+        if: needs.changes.outputs.relevant == 'true'
         run: |
           set -euo pipefail
           container="$(docker ps --filter 'ancestor=postgres:16-alpine' --format '{{.ID}}' | head -1)"
@@ -246,6 +305,7 @@ jobs:
             -c 'SHOW max_locks_per_transaction;'
 
       - name: Run tests
+        if: needs.changes.outputs.relevant == 'true'
         run: pytest tests/unit tests/integration tests/e2e/cloud/test_e2b_webhooks.py -n 4 -q --durations=20
 
   docker:
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -832,7 +832,13 @@ jobs:
       # them, it stops being exempt and the drift guard demands it join
       # `needs:`. The exempt set is printed on every run so the exemption is
       # visible rather than implied.
+      #
+      # `working-directory: .` on both run steps: this file sets
+      # `defaults.run.working-directory: server`, and the sparse checkout above
+      # deliberately materializes only the workflow file, so `server/` does not
+      # exist in this job and bash cannot start there.
       - name: Assert the rollup covers every non-release job in server-ci.yml
+        working-directory: .
         env:
           NEEDS_JSON: ${{ toJSON(needs) }}
         run: |
@@ -863,6 +869,7 @@ jobs:
       # green from the no-op step rather than an absent one. So a `skipped`
       # arriving here means something structural changed, and it should be red.
       - name: Assert every Server CI lane concluded success
+        working-directory: .
         env:
           NEEDS_JSON: ${{ toJSON(needs) }}
         run: |
--- a/.github/workflows/server-ci.yml
+++ b/.github/workflows/server-ci.yml
@@ -12,15 +12,17 @@ on:
       - "pnpm-workspace.yaml"
       - "scripts/validate-agent-catalog.mjs"
       - "server/**"
+  # Deliberately no on.pull_request.paths filter. A path-skipped workflow never
+  # reports a conclusion at all, so a required check naming these jobs would sit
+  # Pending forever on every non-server PR and make it unmergeable -- and the
+  # only way to avoid that today is to leave the server lanes off the required
+  # list entirely, which is the absence hole ci-ok exists to close. So the
+  # workflow now runs on every PR and the cheap `changes` job below diffs the
+  # merge base against the same path list, gating the heavy lanes. A gated-off
+  # lane still reports (a `skipped` conclusion, which branch protection treats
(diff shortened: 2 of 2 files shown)


pablonyx:
### `ci-ok` result on this PR's own run

Head `393166f1d8fb05c709017008c6557de8bb8579bd`, run [32437010909](https://github.com/proliferate-ai/proliferate/actions/runs/32437010909), job `ci-ok` — **success in 6 seconds** (`01:45:41Z` → `01:45:47Z`).

Literal stdout of the two assert steps:

```
ci-ok covers all 14 job(s) in ci.yml.
repo-shape                         success
terraform-validate                 success
ci-cd-config                       success
candidate-build-handoff            success
cargo-check                        success
sdk-build                          success
desktop-frontend                   success
mobile-typecheck                   success
web-frontend                       success
login-budget                       success
shared-frontend-checks             success
shared-frontend-packages           success
scroll-physics                     success
workflow-definition-lifecycle      success
All 14 CI lanes concluded success.
```

Note `shared-frontend-packages` appears once and covers all four shards — the matrix aggregate is `success` only when every shard succeeded. That is why sharding a `ci.yml` lane never needs a branch-protection edit.

The same run on the previous head `47fd16e5bd4557cdf7ba1a4442dc9fd0cdd26f21` produced byte-identical output in 7 seconds.

### Server CI on this PR

This PR edits `.github/workflows/server-ci.yml`, which is on the relevance path list, so the gate resolved `relevant=true` and the real lanes ran:

```
Detect server-relevant changes :: success   10s
lint                           :: success   52s
test                           :: success   14m14s
docker                         :: skipped   (tag/publish-only, by design)
self-hosted-release-assets     :: skipped   (tag/publish-only, by design)
```

The green-no-op branch (`relevant=false`) is what the first non-server PR after this merges will exercise. Before this PR, Server CI produced **no check runs at all** on such a PR.


### Full check sweep on this head

31 SUCCESS, 3 SKIPPED, 0 failures. The three skips are all pre-existing and deliberate: Server CI's `docker` and `self-hosted-release-assets` (tag/publish-only, correctly off the required list), and Self-Host Smoke's `Production compose smoke` (its own relevance gate — the follow-up called out in the PR body).

pablonyx (commented):
## Review — fresh reviewer of record, head `393166f1`

> **Verdict: REQUEST CHANGES** (posted as a comment — GitHub will not let the PR author's account formally request changes on its own PR).

I attacked the rollup hardest, since it becomes the single required check. **I could not find a way to make `ci-ok` certify an unverified commit.** The rollup half is sound. The two findings below are both in the *`changes`* half of `server-ci.yml`, and one of them is a genuine fail-open in the component this whole PR exists to make fail-closed. Both fixes are a few lines, and the PR is a draft that has to be rebased on #2137 anyway — so `request-changes` costs nothing here.

---

### Literal verification (requoted, not taken from the body)

**`gh pr checks 2138` on `393166f1`:** 31 `pass`, 3 `skipping`, 0 `fail`. The three skips are `docker`, `self-hosted-release-assets` (Server CI, tag/publish-only) and `Production compose smoke` (Self-Host Smoke, pre-existing). Matches the claim.

**`ci-ok`** — run `32437010909`, job `96641375444`, **6s**. Literal log:

```
ci-ok covers all 14 job(s) in ci.yml.
repo-shape                         success
terraform-validate                 success
ci-cd-config                       success
candidate-build-handoff            success
cargo-check                        success
sdk-build                          success
desktop-frontend                   success
mobile-typecheck                   success
web-frontend                       success
login-budget                       success
shared-frontend-checks             success
shared-frontend-packages           success
scroll-physics                     success
workflow-definition-lifecycle      success
All 14 CI lanes concluded success.
```

I parsed `ci.yml` at this head independently (`YAML.load_file(...)["jobs"].keys - ["ci-ok"]`) and the 14 job ids are an **exact set match** against the `needs:` list at `.github/workflows/ci.yml:741-755`. No drift.

**`changes`** — job `96640163346`, **10s**. `BASE_REF: main`, `Server-relevant changes since 68078e1d7d2b3a25ad8b68fda7fd60a24b2e5119: true`.

---

### Checklist answers

**2(a) cancelled dependency.** Confirmed. `always()` runs the job, `toJSON(needs)` carries `result: "cancelled"`. I ran the exact loop from `ci.yml:788-812` against synthetic payloads: all-success → exit 0; `{"a":{"result":"skipped"},"b":{"result":"cancelled"},"c":{"result":"success"}}` → exit 1 with `Not green: a (skipped) b (cancelled)`; `{}` → exit 1; `{"a":{"result":null}}` → exit 1. Fail-closed in every direction.

**2(b) whole-run cancellation / stale-green.** **Neither `ci.yml` nor `server-ci.yml` declares a `concurrency:` block** (grep: no match in either file), so there is no `cancel-in-progress` path that could auto-cancel a run. A manual cancel leaves `ci-ok` either failed (because `always()` ran it against cancelled deps) or itself cancelled — never `success`. Branch protection reads the *latest* check run per name per SHA, so a later cancelled `ci-ok` displaces an earlier green one in the **blocking** direction. There is no ordering I can construct that yields an accepted green without a real green run.

**2(c) expression logic.** No string comparison against `toJSON(needs)` as text — it is parsed with `jq` and compared per-entry. `NEEDS_JSON` and `BASE_REF` are passed via `env:` rather than interpolated into the `run:` body, so there is no `${{ }}` script-injection surface (correct practice, worth calling out). `bad=()` + `"${#bad[@]}"` is safe under `set -u` on bash 5. `[[ -z "$missing" && -z "$extra" ]]` correctly aborts under `set -e`. Quoting is clean throughout.

**2(d) drift guard.** Self-exclusion via `- ["ci-ok"]` verified by independent parse. A matrix job appears once as its job id and the aggregate is `success` only if all shards passed — `shared-frontend-packages` proves this on this very run (4 shard checks, 1 `needs` entry, aggregate `success`). Wrong-ref risk is nil: the sparse checkout takes the default ref, and the log shows `git checkout --progress --force refs/remotes/pull/2138/merge` — the same commit the workflow definition itself came from. Renaming `ci.yml` → `YAML.load_file` raises `ENOENT` → red; and the required name `ci-ok` would then resolve against the renamed file's job, whose hardcoded path is gone → still red. Fail-closed both ways.

**3(a) path-list identity — the highest-risk item.** Both lists are the same 7 entries. The glob translation is **correct**, and for a non-obvious reason worth recording: in a shell `case`, patterns use POSIX *Pattern Matching Notation*, **not** filename expansion, so `*` matches `/`. That makes `server/*` ≡ Actions' `server/**`. I tested all the boundary cases:

```
server/a.py           yes    server            no
server/a/b/c.py       yes    catalogs          no
catalogs/x/y.json     yes    myserver/a.py     no
catalogs/x.json       yes    docs/server/a.md  no
.dockerignore         yes
.github/workflows/server-ci.yml  yes
```

No semantic drift. One *improvement* over the old filter worth noting since the body says "verbatim": GitHub's `paths` filter is documented as unreliable above 300 changed files (filters ignored, workflow runs anyway). The `git diff` replacement has no cap, so a >300-file PR touching no server path now correctly no-ops. More accurate, not a regression.

**3(b) merge base.** `actions/checkout` with `fetch-depth: 0` fetches `+refs/heads/*:refs/remotes/origin/*` — literal from the run log — so `origin/main` exists, including for fork PRs (origin is always the base repo). `HEAD` is the merge commit, so `merge-base origin/$BASE_REF HEAD` returns the base parent even after `main` moves. Correct.

**4. per-step gating.** I enumerated every step programmatically: **13/13 in `lint`, 11/11 in `test`, all gated.** No ungated step, no stray teardown. Both no-op steps carry `working-directory: .`, which is load-bearing — `defaults.run.working-directory: server` would break a step on a run where checkout was skipped. Nice catch by the author.

**Bonus soundness check the body doesn't mention:** no job in `ci.yml` carries `continue-on-error`, so no lane can report `success` on a real failure and launder itself past the rollup. And `ci-ok` is the *only* job in `ci.yml` with a job-level `if:` or `needs:` — every one of the 14 lanes is unconditional — so `ci-ok` cannot go spuriously red on `push: main`, and `deploy-staging`'s `workflow_run.conclusion == 'success'` gate is unaffected.

**6. required-checks names.** All 10 recommended names appear verbatim in `gh pr checks` on this head. I also checked the workflows behind them for trigger-level path filters that would strand a PR on Pending: `codeql.yml` is `pull_request: branches: ["main"]` with no `paths`; `pr-metadata.yml` is `pull_request_target`; `self-host-smoke.yml` has no `paths` filter by explicit design. The list is safe to set.

---

## Findings

### F1 — MEDIUM, fail-open. `git diff` failure silently yields `relevant=false`
`.github/workflows/server-ci.yml:122`

```
done < <(git diff --name-only "$base" HEAD)
```

Process substitution is not a pipeline, so neither `set -e` nor `set -o pipefail` can observe a failing `git diff`. If it fails for any reason the loop reads zero lines, `relevant=false` survives untouched, and **both server lanes report a real green on a real server PR** — an actively worse outcome than the path-skip this PR replaces, because it looks verified.

This is the identical hazard #2137's author called out for themselves in this same file:

> `mapfile < <(...)` reads from a process substitution, which is not a pipeline, so `set -o pipefail` cannot observe a failing `find` and the step would happily continue with a short or empty list.

Same standard should apply here, and unlike #2137 there is no floor check as a backstop. Two-line fix, since command substitution *is* covered by `set -e`:

```bash
files="$(git diff --name-only "$base" HEAD)"
while IFS= read -r file; do
  ...
done <<<"$files"
```

### F2 — MEDIUM, doctrine violation. A failed `changes` job turns both lanes into `skipped`
`.github/workflows/server-ci.yml:127` and `:206` — `needs: changes` with no job-level `if:`.

If `changes` fails or is cancelled, `lint` and `test` are skipped and report **`skipped`** — precisely the conclusion this PR's own prose says "branch protection silently accepts" and "an absence-is-fatal required check must never be satisfied by." The only thing holding it closed is `Detect server-relevant changes` being on the required list. Drop that one name, or rename the job, and any `changes` flake converts both server lanes into satisfied-by-skip.

The gate polarity is fixable to be structurally fail-closed rather than list-dependent:

- job-level `if: always()` on `lint` / `test`
- no-op step on `needs.changes.outputs.relevant == 'false'`
- every heavy step on `needs.changes.outputs.relevant != 'false'`

An unset/empty output — which is exactly what a failed or cancelled `changes` produces — then runs the **real** suite instead of going green. That is the fail-closed default this PR is arguing for everywhere else.

### F3 — LOW, latent fail-open in the release path
`.github/workflows/server-ci.yml:103` — `if [[ "$GITHUB_EVENT_NAME" != "pull_request" ]]`.

As the `docker` job's own comment in this file notes, a `workflow_call` inherits the **caller's** event name, never `workflow_call`. Today both callers are non-PR (`hotfix-production.yml` = `workflow_dispatch`, `nightly-release-train.yml` = `schedule`/`workflow_dispatch`), so this is correct right now — I verified both. But if any future reusable-workflow caller is PR-triggered, the release invocation would compute relevance from the *PR's* diff and could ship with `lint`/`test` no-opped. Cheap hardening: treat a non-empty `inputs.git_sha` as unconditionally relevant too.

### F4 — LOW, new red-surface on 100% of PRs (not just cost)
`.github/workflows/server-ci.yml:201-233`.

The body frames the unconditional `services:` boot as a *cost*. The sharper issue is that a Postgres/Redis service-container health-check failure now **red-blocks every PR**, including docs-only ones, on a lane that had nothing to verify. That is a brand-new flake surface on the merge path, and #2137 multiplies it by four (`test-unit` + 3 × `test-integration`, each booting its own pair). This is the strongest argument for follow-up (1), `server-ci-ok` — a rollup inside `server-ci.yml` lets the lanes be job-level `if:`-skipped *and* still yields one honest green. I'd decide on that before setting branch protection, not after.

### F5 — LOW, durability. The required-checks list lives only in a PR description
`gh search code` over the repo finds **zero** content describing required status checks (only `self-host-smoke.yml`'s inline comment). For the enforcement backbone, the list plus the "why `skipped` is fatal" rationale should be committed — a doc under `specs/`/`docs/`, or better a checked-in ruleset. Right now the post-#2137 swap depends on someone remembering a merged PR body.

### F6 — NIT. The empty-needs guard is effectively unreachable
`.github/workflows/ci.yml:804`. With a hardcoded `needs:` list, an empty resolution would already have tripped the drift guard's `missing` branch one step earlier; and a `needs:` entry naming a nonexistent job makes the whole workflow invalid, so it never runs and the check stays Pending (also fail-closed). Harmless belt-and-braces — keep it — but it is not the independent safety property it reads as. Worth a one-line comment saying so.

### F7 — NIT. Drift-guard YAML parse and anchors
`.github/workflows/ci.yml:772`. On the runner's Ruby 3.x (Psych 4+), `YAML.load_file` is `safe_load` with `aliases: false`, so an anchor/alias anywhere in `ci.yml` raises `Psych::AliasesNotEnabled` and turns the *gate* red with an error nobody will connect to a YAML anchor. Fail-closed, and it is consistent with the existing repo-wide constraint (`ci-cd-config` → "Check workflow YAML parses", `ci.yml:208`, does the same thing over all workflows), so it's not a new restriction — just note that the failure now surfaces at the merge gate first. `aliases: true` would remove the surprise.

---

## Metadata

Title `ci(workflows): add ci-ok rollup gate that fails on skipped lanes` matches `TITLE_PATTERN` in `scripts/ci-cd/pr-metadata.mjs:38`. Labels: exactly one `release:*` (`release:maintenance`) plus `area:release`. All changed paths are `.github/`, which `AREA_RULES` maps to `area:release` and nothing else, so `area:release` alone is correct and complete. `Validate PR title and labels` passed in 12s. Clean.

---

## Recommended merge order: **#2137 first, then rebase this PR on top**

The body suggests either order with a warning about the required-list window. I'd make it explicit and pick #2137, for a reason the body actually establishes but doesn't follow through on:

1. **Branch protection does not exist yet** (no rules, no rulesets). So the "if #2137 merges before the required-checks list is updated, `test` stops reporting and every PR blocks" hazard **is not live**. This is the one free window to reorder, and it closes the moment someone sets the first rule.
2. Merging #2137 first means the required list is authored **once**, in final post-split form (`test-unit`, `test-integration (1..3)`), and never has to be edited. The other order requires writing a transient `test` entry and then a coordinated same-window swap — the exact operation the body flags as dangerous.
3. #2137 is already non-draft, `MERGEABLE`/`CLEAN`, and green. This PR is still a draft.
4. The rebase cost is mechanical either way, and the in-file guidance at `server-ci.yml:76-83` is written **for this direction** — it names #2137 and spells out exactly what each new lane needs (`needs: changes`, per-step `if:`, the green no-op step). It gives the rebaser what they need. I'd fold F1 and F2 into that same rebase pass.

One caution that comes with that order: after #2137, per-step gating means 3 integration shards plus `test-unit` each boot Postgres+Redis on every irrelevant PR (F4). Settle the `server-ci-ok` follow-up before configuring branch protection so you don't pay that four times over.

---

**Verdict: request changes on F1 + F2 only.** The `ci-ok` rollup itself is correct, well-reasoned, and I could not break it. F1 is a two-line fix; F2 is a polarity flip. Everything else is advisory.

pablonyx:
### Both rollups green — head `e548ab2522306da0fccfd87307a617f35f191629`

Rebased onto `main` past #2137 (server lane split) and #2136. Full sweep: **35 SUCCESS, 3 SKIPPED, 0 failures.** The three skips are all pre-existing and deliberate — Server CI's `docker` and `self-hosted-release-assets` (release-only, exempted by the rollup's derived `server-v` gate detection) and Self-Host Smoke's `Production compose smoke`.

**`ci-ok`** — run [32439557650](https://github.com/proliferate-ai/proliferate/actions/runs/32439557650), job `96648622697`, **success in 7s**:

```
ci-ok covers all 14 job(s) in ci.yml.
repo-shape                         success
terraform-validate                 success
ci-cd-config                       success
candidate-build-handoff            success
cargo-check                        success
sdk-build                          success
desktop-frontend                   success
mobile-typecheck                   success
web-frontend                       success
login-budget                       success
shared-frontend-checks             success
shared-frontend-packages           success
scroll-physics                     success
workflow-definition-lifecycle      success
All 14 CI lanes concluded success.
```

**`server-ci-ok`** — run [32439557691](https://github.com/proliferate-ai/proliferate/actions/runs/32439557691), job `96648256650`, **success in 8s**:

```
Release-only jobs exempt from the rollup: docker, self-hosted-release-assets
server-ci-ok covers all 4 non-release job(s) in server-ci.yml.
changes                  success
lint                     success
test-unit                success
test-integration         success
All 4 Server CI lanes concluded success.
```

`test-integration` appears once and covers all three shards — the same matrix-aggregation property `shared-frontend-packages` demonstrates for its four. This is exactly why the required list does not change when a lane is sharded.

Server CI ran for real (`relevant=true`, since this PR edits `server-ci.yml`): `Detect server-relevant changes` 10s, `lint`, `test-unit`, `test-integration (1..3)` all success.

---

### Two live negative controls on the way here

Neither was staged. Both are the gate doing its job on the first try.

**1. `ci-ok` correctly red on a real lane failure.** On head `350e7261a`, `Shared frontend packages (2)` failed and `ci-ok` reported:

```
shared-frontend-packages           failure
```

The shard's own summary was `Test Files 274 passed (274)` / `Tests 1989 passed (1989)` with one post-run `ReferenceError: window is not defined` unhandled error — a teardown flake, unrelated to this diff (workflows plus one markdown file). Reran clean. Worth noting for what it proves: the matrix aggregate went `failure` from a single shard, and the rollup surfaced it by name rather than letting a 1989-green shard set launder it.

**2. `server-ci-ok` caught a real bug in itself.** Its first run failed with:

```
An error occurred trying to start process '/usr/bin/bash' with working directory
'/home/runner/work/proliferate/proliferate/server'. No such file
```

`server-ci.yml` sets `defaults.run.working-directory: server`, and the rollup deliberately sparse-checks-out only the workflow file, so `server/` does not exist in that job. Fixed in `e548ab2` by adding `working-directory: .` to both run steps, with a comment explaining why. The same hazard is why the `changes` job's detect step and all three no-op steps already carried `working-directory: .`; I had missed it on the new job. Audited every `run` step in the file afterwards — all six that need it have it.

Also fixed in passing: adding the delivery README to the diff made `area:docs` a required label (`deriveAreaExpectation` → `["area:docs","area:release"]`). Label added; `Validate PR title and labels` is green.

pablonyx (commented):
## Re-review — head `e548ab25` (was `393166f1`)

**Verdict: CLEAR. All seven findings closed, verified independently. No blocking defect. Ship it.**

I re-derived every claim from the artifacts rather than the closure summary. Three residual notes at the bottom are advisory only — none of them should hold the merge.

---

### Fresh literals

**`gh pr checks 2138` @ `e548ab25`: 35 pass / 3 skipping / 0 fail.** Skips = `docker`, `self-hosted-release-assets` (release-only), `Production compose smoke` (pre-existing). Diff vs base is exactly **3 files**: `ci.yml`, `server-ci.yml`, `specs/codebase/systems/engineering/delivery/README.md`.

`ci-ok` 7s, `server-ci-ok` 8s. Independent parse at this head:

- `ci-ok` `needs:` ≡ `ci.yml` jobs minus `ci-ok` — **exact 14-element set match**.
- `server-ci-ok` `needs:` ≡ `["changes","lint","test-integration","test-unit"]`, and running the shipped partition Ruby myself against the file yields exactly that, with `EXEMPT: docker, self-hosted-release-assets`. Matches the live log:

```
Release-only jobs exempt from the rollup: docker, self-hosted-release-assets
server-ci-ok covers all 4 non-release job(s) in server-ci.yml.
changes                  success
lint                     success
test-unit                success
test-integration         success
All 4 Server CI lanes concluded success.
```

Note `test-integration` appears as **one** row covering all three shards — the matrix-aggregation property, live.

---

### Both negative controls confirmed from run history

**(1) `ci-ok` vs a real matrix flake** — run `32438937733`, job `96646887641` @ `350e7261`. `Shared frontend packages (2)` failed. Literal `ci-ok` output:

```
shared-frontend-packages           failure
##[error]CI did not verify this commit. Not green: shared-frontend-packages (failure)
```

This is materially better evidence than the synthetic payloads I ran last round: it proves the aggregate of a 4-shard matrix collapses to one `failure` in `toJSON(needs)` and that `if: always()` reported it. The exact scenario the rollup exists for, observed live.

**(2) `server-ci-ok` sparse-checkout interaction** — run `32438937811`, job `96646543508`:

```
##[error]An error occurred trying to start process '/usr/bin/bash' with working directory
'/home/runner/work/proliferate/proliferate/server'. No such file or directory
```

Exactly the `defaults.run.working-directory: server` × sparse-checkout collision, caught by the gate itself.

**I re-ran the `working-directory: .` audit independently** rather than trusting the fix. Six `run:` steps in `server-ci.yml` can execute without a full checkout, and **all six carry `working-directory: .`**: the `changes` detect step (`:96`), the three no-op steps (`lint:184`, `test-unit:318`, `test-integration:421`), and both `server-ci-ok` run steps (`:840`, `:871`). Every other `run:` in the file sits behind an unconditional full checkout. `ci.yml` has no `defaults:` block at all, so `ci-ok` is unaffected. Audit clean.

---

### Findings, one by one

**F1 — CLOSED, correctly.** `server-ci.yml:113` is now `changed="$(git diff --name-only "$base" HEAD)"` with `done <<<"$changed"`. Command substitution *is* observed by `set -e`, so a failing `git diff` now aborts the step instead of leaving `relevant=false` standing. The added floor guard (`-z "$changed"` → exit 1) is the right extra belt. The path `case` list is byte-identical to what I verified last round, so the glob equivalence proof still holds.

**F2 — CLOSED, and better than I asked for.** Polarity is flipped exactly as recommended: job-level `if: always()` on all three lanes, no-op on `== 'false'`, every real step on `!= 'false'`. I re-enumerated: **13/13 lint, 10/10 test-unit, 12/12 test-integration gated, zero ungated** (counts include the no-op step; the summary's 12/9/11 excludes it). A failed or cancelled `changes` now yields an empty output, `!= 'false'` is true, and the **real suites run** — structurally fail-closed rather than list-dependent. The partition step is gated too, so its floor guards can't fire on a no-op run. And the fix is now double-covered: `changes` is itself in `server-ci-ok`'s `needs:`, so a failed relevance probe is red at the rollup regardless.

**F3 — CLOSED.** `GIT_SHA_INPUT` is checked *before* the event-name branch, so a non-empty `inputs.git_sha` is unconditionally relevant. That is the right discriminator: it can only come from a `workflow_call`, which is precisely the case where the inherited caller event name lies.

**F4 — CLOSED, beyond scope in a good way.** `server-ci-ok` exists, so no per-lane server name is required any more. The derived exemption is the right instinct: the drift guard partitions on the presence of the `refs/tags/server-v` gate rather than a name list, so removing a release gate makes that job *stop* being exempt and demands it join `needs:` — the failure mode points the safe way. Printing the exempt set on every run is the correct complement.

**F5 — CLOSED, durably.** The list now lives in two places: the `ci.yml` header comment and a new `## Merge Gate` section in the delivery README. I diffed the two lists — **identical 8 names**, and both carry the rationale and the exclusion reasons. I also checked the #2140 conflict claim myself: #2140's last README hunk ends at old line 319; #2138's is `@@ -322,3 +322,58 @@`. Non-overlapping with a 3-line gap, so both auto-merge. Confirmed.

**F6 — CLOSED.** Comment now states plainly that the zero-dependency branch is belt-and-braces, not an independent property, and why. That was the whole ask.

**F7 — CLOSED, with the right nuance.** `YAML.respond_to?(:unsafe_load_file) ? unsafe_load_file : load_file` is a proper feature detect (Psych 3's `load_file` already permits aliases), and the comment correctly scopes it as a *diagnostic* fix while pointing at `ci-cd-config`'s parse step as the owner of YAML validity.

> Security clearance on `unsafe_load_file`, since "unsafe" in a merge gate deserves an explicit ruling rather than silence: it can instantiate `!ruby/object:` tags, but the file being parsed is the repo's own `ci.yml` at the merge ref, and any actor who can put a Ruby object tag there can already add an arbitrary `run:` step to the same file. The job's token is `contents: read`. **No new privilege — cleared.**

---

### Metadata

Title unchanged and still matching. Labels now `release:maintenance` + `area:release` + `area:docs`, the last correctly added for the new `specs/` file. `Validate PR title and labels` green in 8s (the two earlier red PR Metadata runs on this SHA predate the label add; latest is the verdict).

---

### Merge order — resolved, not deferred

#2137 **merged at 02:03:01Z**, and this PR moved forward past it via a merge of `origin/main`. That is the order I recommended and it played out with the payoff intact: because `server-ci-ok` now exists, the required list never had to contain the transient `test` name, and the lane split cost **zero** branch-protection edits. The list is authored once, in final form. Nothing left to sequence.

---

### Three advisory notes (do not block)

1. **Empty-diff PRs now hard-fail.** `server-ci.yml:117` exits 1 when `git diff base..HEAD` is empty. Right default, but a PR whose content was absorbed into `main` by another PR can legitimately reach an empty merge-base diff and will now red-block on `server-ci-ok` rather than being trivially green. Loud and fail-closed, so I'd leave it — just know the shape when it first shows up.
2. **The exemption heuristic is a substring test.** `spec["if"].to_s.include?("refs/tags/server-v")` would also exempt a job whose gate *negates* the tag (`if: !startsWith(github.ref, 'refs/tags/server-v')`) — such a job would run on PRs yet sit outside the rollup. It takes a deliberate diff, the exempt set is printed every run, and it is strictly better than a hardcoded name list. Tightening to match the positive form, or an explicit `# rollup-exempt: release` marker, would close it whenever someone is next in the file.
3. **`Production compose smoke` is the one required name still satisfiable by a skip** — it reports `skipping` on this very PR. Absence is covered by its always-reporting `Detect smoke-relevant changes` sibling, so this is the documented residual, not a regression. It was follow-up (2) in the original body; worth keeping alive somewhere now that the body has been rewritten, since Self-Host Smoke is the last workflow on the list without the treatment this PR invented.

---

**Cleared to merge.** The rollups are sound, the relevance probe is fail-closed in every direction I could attack, the required list is durable and every one of its 8 names resolves on this head, and both claimed negative controls are real and reproduced from run history.

PR #2140, opened 2026-08-21
ci(release): unify deploys into one main-to-prod pipeline

pablonyx:
## The model

Three states, and the transitions between them are thin workflow files whose deploy logic stays in the reusable `_deploy-*.yml` lanes.

```mermaid
stateDiagram-v2
    [*] --> Main: PR merged (ci.yml, untouched)
    Main --> Production: release.yml (09:00 UTC cron or dispatch)
    Main --> Staging: deploy-staging.yml (manual only)
    note right of Staging
        Deployed, not e2e-proven.
        Gates nothing.
    end note
```

At launch, exactly two automated paths exist: PR to main (`ci.yml`, not touched here) and main to production (`release.yml`). Nightly is that same pipeline on a cron, unattended. Continuous staging is retired: merging to main now deploys nothing.

Hotfix and promote stop being files and become inputs:

| Intent | Dispatch |
| --- | --- |
| Nightly | cron, no inputs (`surfaces=all`, `ref=main`) |
| Hotfix | `ref=<exact sha on main>` plus `surfaces=<exact set>` |
| Promote | `skip_build=true` on a ref that has already been through a build run |
| Wiring proof | `dry_run=true`, alone or combined with either of the above |

The nightly covers every surface: server, web, and litellm deploys plus a desktop release that publishes the updater manifest. `surfaces=all` (the default) means every surface is *eligible*, and change detection against the previous checkpoint picks the ones that actually moved, so an unchanged surface is not re-released or redeployed and a quiet night does nothing. An explicit `surfaces` list skips detection and is exact. Nothing in the pipeline asks for approval: dispatching is the authorization, and the cron needs none.

`dry_run` (fourth input, default false) is the standing way to prove a change to this pipeline before it can ship anything. It walks the whole graph and suppresses every externally visible effect: prepare computes the version plan but commits nothing to `main` and creates no tags, the artifact release builds are skipped, each deploy lane is still called with `enabled: false` so the inner job that reaches AWS or Vercel never starts, and the product release body renders without creating or updating a GitHub Release. Both step summaries say so. It composes with the rest: `surfaces` narrows what the plan covers, `skip_build` walks the deploy-only shape (which still requires an explicit `surfaces` list, since the R7 guard is part of what a wiring proof should exercise). A scheduled run supplies no inputs, so `dry_run` is false and cron behavior is unchanged.

The build lanes are skipped on a dry run rather than called with their own `dry_run`, because their dry-run input still compiles everything and a macOS desktop matrix is not a proportionate price for a wiring proof. The deploy lanes have a real no-op switch, so calling them costs seconds and proves the call wiring. This matches how the retired train's plan-only dry run behaved, extended so the deploy call wiring is exercised rather than skipped outright.

## MERGE CONDITION

**This PR merges together with, or after, the removal of `required_reviewers` from the `Production` GitHub Environment. Never before.**

While that rule is set, the 09:00 UTC cron pushes its version-bump commit to `main` and creates the release tags in `prepare` (which is bound to no environment), then parks `deploy-server-prod` in Waiting. The parked run holds the `nightly-release-train` concurrency group with `cancel-in-progress: false`, and there is no longer a separate hotfix workflow to route around it: later dispatches queue behind the parked run rather than preempting it, and a third dispatch cancels the queued one instead of the stuck one. Every manual production deploy is blocked until someone intervenes. Operator escape hatch for a parked run: `gh run cancel <run-id>`, which releases the group.

If the settings flip cannot land in the same window, disable the `Release` schedule until it does. The same consequence is now stated plainly in the delivery spec rather than as "stops the cron from completing unattended".

## Per-file changes

- **`.github/workflows/release.yml` (new, replaces `nightly-release-train.yml`)**. Dispatch inputs are `surfaces` (default `all`), `skip_build`, `ref` (default `main`), plus the 09:00 UTC cron. `prepare` resolves the checkpoint, versions, and surfaces from `detect-deploy-surfaces.mjs`, honoring `ref`, and fails loudly when a `skip_build` run resolves an empty surface set. Release builds (`release-runtime`, `server-ci`, `release-desktop`) hang off `needs: [prepare]` and are skipped when `skip_build` is set, along with the version bump and tag creation, so a promote run deploys the exact bytes that already exist and mints no new identities. Prod deploys hang off `prepare` directly: `deploy-server-prod` then `deploy-web-prod` (the server-before-web edge is API compatibility ordering), with `deploy-litellm-prod` in parallel. They call the same reusables with the same `environment: Production`, `secrets: inherit`, and `enabled:` wiring the current train's prod jobs use. `publish-product-release` was re-derived to gate on the three release builds only, since the staging legs it used to wait on no longer exist. Concurrency group name is kept verbatim (`nightly-release-train`) so a run still in flight when this merges cannot deploy prod alongside a run of the new file.
- **`.github/workflows/deploy-staging.yml`**. `workflow_run` trigger removed, `workflow_dispatch` kept, internals untouched. The concurrency group lost its `workflow_run` half (that context is gone) and the `plan` job lost its trusted-event `if:` for the same reason.
- **`.github/workflows/promote-production.yml`, `hotfix-production.yml`**. Deleted. Both still called dead `deploy-e2b`/`deploy-mobile`/`deploy-workers` jobs, so this also closes the scope gap #2126 left open.
- **`scripts/ci-cd/deploy-staging-trigger.test.mjs`**. The old assertions were about a `workflow_run` trigger that no longer exists. Rewritten to the successor properties: staging runs only on dispatch, no workflow anywhere both starts from main and calls a `_deploy-*` lane, `release.yml` is the only file that names `environment: Production`, and staging never targets Production. Negative control against `origin/main`: 3 of the 4 fail there, 4 of 4 pass here.
- **`scripts/ci-cd/release-title-propagation.test.mjs`**. The coordinator case asserted `release_title` plumbing in all three now-collapsed coordinators. Rewritten to assert the desktop title entrypoint survives the collapse (`_deploy-desktop.yml` pass-through intact, `release.yml` carries exactly the three ruled inputs, the three retired files are gone). Negative control against `origin/main`: fails there, passes here.
- **`specs/codebase/systems/engineering/delivery/README.md`**. Hosted spine, release coordinator section, Server CI comparison-SHA note, workflow inventory, identities table, and Current Gaps updated to the new model.

## Latency

Time to production is prepare, then server, then web. That is roughly ten minutes on the current job shape, unchanged by this PR since no deploy internals moved. Runtime, server, and desktop artifact builds run in parallel with the deploys and never gate production; desktop in particular is the long pole and no longer sits anywhere near the path to a live prod rollout.

## Judgment calls

1. **`surfaces=all` means "every surface is eligible", not "force all".** Passing `--only all` to the detector would force a desktop release, updater publish, and fresh version tags every single night regardless of changes, which is exactly what change detection exists to prevent. The default maps to an empty `--only`, so detection runs across the full surface set; an explicit list still maps to `--only <list>`.
2. **`release_title` and `version_bump` are gone from the dispatch surface; `dry_run` was re-admitted by ruling.** The original input list was `surfaces`, `skip_build`, `ref`; `dry_run` came back as a fourth input so wiring changes can be proven without shipping. `version_bump` stays pinned to `patch`. A titled desktop release is dispatched on `release-desktop.yml` directly, which keeps its own `release_title` input; the test pins that pass-through, and now also asserts the dispatch input set is exactly those four so a fifth cannot appear unnoticed.
3. **`skip_build` also skips the version bump, tags, and product release page.** Otherwise a promote run would push a version-bump commit to main and deploy a SHA that is not the proven one. Promote is strictly deploy-only, and it is a deterministic rebuild of the promoted commit rather than a retag of existing bytes: each hosted lane still builds its own image at that SHA. Artifact handoff is the follow-up that would make it a true retag.
4. **`deploy-web-prod` gates on `(server != 'true' || deploy-server-prod.result == 'success')`** rather than a bare `result == 'success'`. A bare check would block a web-only deploy, because an unselected server lane does not report success. Ordering and the failure interlock are both preserved.
5. **Concurrency group name kept as `nightly-release-train`** rather than renamed, purely so the merge window cannot produce two concurrent prod deploys. Comment in the file explains it.

## Review note

**Prod deploy surface. Founder merge only.** This changes what the 09:00 UTC cron does to production. No approval logic is touched here: no `environment:` name changed, and removing the `Production` environment's required reviewers remains a repository settings flip outside this PR. The `server-ci.yml` inventory row still reads "Relevant push/PR" because the always-run-with-fast-no-op change is a separate PR; I did not write the new posture into the delivery README ahead of it.

## Review response (R1-R7, head `8f2ce8b479c7ef857485b46d7035731e632f0d89`)

- **R1** `guides/operating/e2b-template-operations.md`: the rollback step and the runnable `gh run list --workflow "Promote Production"` now name `Release Cloud Template` / `Promote Cloud Template`. Dropped the stale `Deploy Staging` entry from the same two places for the same reason (it has had no E2B job since #2126), and repointed step 3 of the publish flow, which described promotion through the now-deleted reusable.
- **R2** Documented, not coded around. New MERGE CONDITION section above; the delivery spec paragraph now states the parked-run consequence in full, including that it blocks every manual production deploy, that no hotfix workflow can preempt it, and that `gh run cancel` is the escape hatch.
- **R3** Fixed in all three places. The spec, the workflow comment, and the description now say a promote deploys the exact ref without minting new artifact identities and rebuilds deterministically from that SHA. The byte-reuse claim is gone.
- **R4** Both the filter and the `doesNotMatch` use `/environment:\s*["']?[Pp]roduction["']?/`.
- **R5** `triggerBlock` now takes the file name and asserts the block was located and non-empty instead of falling through to `""`; the main-branch filter matches flow and block sequences.
- **R6** `_deploy-e2b.yml`, `_deploy-mobile.yml`, and `_deploy-workers.yml` deleted, and the unread `workers` output dropped from `prepare`. The `release.yml` comment now names the real set (runtime/server/desktop releases, server/litellm/web deploys) and says e2b, mobile, and workers have no job here. Fallout cleaned: three inventory rows, the Artifact-lanes prose, the Worker-lane sentence, two markdown links that `check_docs` would have failed on (`specs/FEATURE_DOCS/SANDBOX/lifecycle.md`, `guides/operating/e2b-template-operations.md`), one prose mention in `support-loop.md`, and three stale code comments. New gap bullets cover E2B having no automated production path, mobile having no deploy path, and the Worker lane, with an explicit note that `_deploy-server.yml`'s Celery worker/Beat rollout is a different thing and is unaffected.
- **R7** `prepare` now fails with a `::error::` telling the operator to pass an explicit `surfaces` list when `skip_build` is set and detection resolved nothing.

**Negative controls** (each run against a synthetic tree, so the guard is proven to catch the shape it targets):

| Control | Pre-review tests | Hardened tests |
| --- | --- | --- |
| Tree containing the two deleted coordinators from `origin/main` (lowercase `environment: production`) plus a `push: branches:\n - main` workflow calling `_deploy-server.yml` | `# pass 4 / # fail 0` | `# pass 2 / # fail 2` (`not ok 2 - merging to main deploys nothing`, `not ok 3 - release.yml is the only entrypoint into a production deploy lane`) |
| Workflow file with no `on:` block calling `_deploy-server.yml` | `# pass 4 / # fail 0` | `# pass 3 / # fail 1`, error `zz-no-on-fixture.yml: could not locate the on: trigger block` |
| Pre-R7 `release.yml` | n/a | `not ok 5 - a deploy-only run refuses to resolve an empty surface set` |

Local re-runs: `node --test scripts/ci-cd/*.test.mjs` → `# tests 208 / # pass 208 / # fail 0`; `python3 scripts/check_docs.py` → `Documentation integrity check passed (234 Markdown files).`; `python3 scripts/check_max_lines.py` → passed; ruby YAML parse → all 24 workflow files parse.

The two "no action requested" observations are acknowledged and left as-is: `release-server` no longer interlocking `deploy-server-prod` is the ruled model, and the `always()` versus `!cancelled()` choice is inherited from the old train. Both are worth a founder look given the unattended posture.

## Out of scope, follow-ups

- Artifact handoff between the release builds and the deploy lanes. This is also what would turn a `skip_build` promote into a real retag instead of a deterministic rebuild. Every `_deploy-*.yml` still builds its own exact-SHA image, so a run that releases and deploys the same surface builds it twice.
- Environment-parameter refactor of the `_deploy-*.yml` internals (they still take `environment` string plus `enabled` boolean, repeated at every call site).
- The settings flips: `Production` required reviewers, and any branch-protection required-check list that named the retired workflows.
- `guides/deploying/releases.md` and `guides/deploying/hosted.md` still document the nightly train, `Hotfix Production`, and `Promote Production` as operator procedures. Listed as a gap in the delivery spec; the operator-guide rewrite is a separate pass.
- No automated end-to-end proof gates the Staging state, and no transition consumes one.


pablonyx (commented):
## Review of record — PR #2140 (fresh reviewer, did not implement)

Head reviewed: `a742417780651380dc5c0ff8dad5cbc3cffac7d7`. Static analysis only; no workflow was dispatched.

**Verdict: changes requested.** The needs-graph, the `surfaces=all` mapping, the `skip_build` gating, the web-after-server interlock, and the negative controls all hold up — I re-derived and re-ran them. Three real defects remain (R1, R3, R7), plus one merge-gate condition (R2) that is a settings flip but decides whether the first cron run is a release or a 24h stall.

### Literal check results

`gh pr checks 2140` on this head: all required checks pass. `CI/CD config checks` (run 32437174492, job 96640633392, `headSha a742417780651380dc5c0ff8dad5cbc3cffac7d7`) ran `node --test scripts/ci-cd/*.test.mjs` → `# tests 206 / # pass 206 / # fail 0`, and the Ruby YAML parse listed 27 workflow files with `release.yml` present and `nightly-release-train.yml`, `promote-production.yml`, `hotfix-production.yml` absent.

Reproduced in a scratch clone:

- `node --test scripts/ci-cd/*.test.mjs` on this head → `# tests 206 / # pass 206 / # fail 0`.
- Negative control, the two rewritten test files checked out onto the `origin/main` tree → `# pass 6 / # fail 4`: `not ok 1 - staging runs only when an operator dispatches it`, `not ok 2 - merging to main deploys nothing`, `not ok 3 - release.yml is the only entrypoint into a production deploy lane`, `not ok 7 - the desktop title entrypoint survives the coordinator collapse`. Exactly the 3-of-4 plus 1 the description claims. (`ok 4 - staging never targets the production environment` passes on main too, so it is a property assertion, not a control.)
- `python3 scripts/check_docs.py` → `Documentation integrity check passed (234 Markdown files).`

### The DAG as I read it

```
prepare  (cron: ref=main, surfaces=all, skip_build=false)
  ├─ release-runtime    if prepare ok && !skip_build && runtime
  ├─ release-server     if prepare ok && !skip_build && server      (server-ci.yml)
  ├─ release-desktop    if prepare ok && !skip_build && desktop     (publish_updater: true)
  ├─ deploy-server-prod   if prepare ok && surfaces != ''   enabled=server
  ├─ deploy-litellm-prod  if prepare ok && surfaces != ''   enabled=litellm
  └─ deploy-web-prod      needs [prepare, deploy-server-prod]
                          if always() && prepare ok && surfaces != ''
                             && (server != 'true' || deploy-server-prod == success)
                          enabled=web
publish-product-release  needs [prepare, 3 release builds], always()-guarded,
                         !skip_build && product_tag != '' && surfaces != ''
                         && per-surface (not-selected || success)
summary  needs everything, always() && prepare ok
```

Silent-skip sweep, edge by edge:

- `deploy-server-prod` / `deploy-litellm-prod` have no `always()`, but their only `needs` is `prepare`, whose success is already asserted in the `if`. No hidden skip. Dropping `always()` here is an improvement over main, where both carried it.
- `deploy-web-prod` truth table: (a) server selected + server deploy success → runs; (b) server selected + server deploy failure → `server != 'true'` false and `result == 'success'` false → skipped, interlock holds; (c) server selected + cancelled → skipped; (d) server not selected → left disjunct true, runs with `enabled=false` unless web is selected; (e) web-only dispatch (`surfaces: web`) → `deploy-server-prod` still runs with `enabled=false`, its reusable job self-skips, web runs. I found no path where web rolls despite a failing selected server deploy.
- `publish-product-release` correctly uses `always()` plus per-surface `(not-selected || success)`, so a skipped build for an unselected surface does not suppress it and a failed build for a selected surface does.
- Cron event: `github.event.inputs` is absent, so `github.event.inputs.surfaces || 'all'`, `.skip_build || 'false'`, `.ref || 'main'` all resolve to the intended defaults. `grep -nE '\$\{\{[^}]*[^.]\binputs\.'` over `release.yml` returns nothing, so there is no bare `inputs.*` that would evaluate empty on a schedule event. Clean.
- `skip_build=true`: `artifacts` step skipped → `product_tag`/`*_version` empty → all three release builds skipped, `Verify version tags are free` and `Create release tags` skipped, `Commit version bumps` early-returns before `git push origin HEAD:main`, `publish-product-release` blocked by both `skip_build != 'true'` and `product_tag != ''`. No tag, no bump, no release page. Confirmed.
- `surfaces=all` mapping: verified against `scripts/ci-cd/detect-deploy-surfaces.mjs`. `parseSurfaceList("")` → empty set → `selectionMode = "detected"` → detection runs. `parseSurfaceList("all")` → `new Set(SURFACES)` with `selectionMode = "only"`, which would force all eight surfaces including desktop every night. **The claimed footgun in judgment call 1 is real**, and mapping `all` → empty `--only` is the correct fix. An explicit list maps straight through, and `all,server` throws (`all cannot be combined with other surfaces`), so the mixed case fails closed.
- Artifact handoff on `skip_build`: not undefined. `_deploy-server.yml:32`, `_deploy-web.yml:26`, `_deploy-litellm.yml:53` all bind `environment: ${{ inputs.environment }}` on the inner job and check out `inputs.git_sha`, then build their own image at that exact SHA. Nothing resolves `latest` or an unbuilt ref. See R3 for the wording problem this creates.
- `deploy-staging.yml`: no `github.event.workflow_run` reference survives anywhere in the file; concurrency is now the constant `deploy-staging-${{ github.repository }}` with `cancel-in-progress: false`, and `plan` has no `if:`, so the dispatch path is well-formed.
- Dangling-reference sweep across the whole repo for the three retired filenames, their workflow display names, and `workflow_run` linkages: the only `workflow_run:` trigger left in `.github/workflows` is `release-e2e-hard-cancel-cleanup.yml:11` pointing at `"Release E2E (tier 3)"`, which still exists. `catalog-probe.yml:14` has a stale prose comment ("before the nightly release train picks up merged catalogs") — harmless. The live references are in R1.
- Metadata: title matches `ci(release): …`; labels are `release:maintenance` + `area:release` + `area:docs`, which covers the `.github` and `specs` halves of the diff. `Validate PR title and labels` passes. PR is still a draft.

---

### R1 — Stale operator doc with a runnable command that now fails, and it is not on the out-of-scope list (fix before merge)

`guides/operating/e2b-template-operations.md:85` tells the operator to find the bad immutable tag from a failed `Promote Production` run, and `guides/operating/e2b-template-operations.md:99` is a literal `gh run list --workflow "Promote Production" --limit 20` that will error once the file is deleted. The description scopes out `guides/deploying/releases.md` and `guides/deploying/hosted.md`, and the new Current Gaps bullet only says "the operator procedures in Deploying" — this file lives under `guides/operating/` and is covered by neither.

Both references should point at `Promote Cloud Template` instead, which is the surviving path (`promote-cloud-template.yml:46-51` runs `promote-cloud-template.mjs --tag production`). That is a two-line edit, not a guide rewrite.

### R2 — Merge-gate: the first cron run parks on the Production approval and holds the only prod lane (not a code change, but it must be sequenced)

`release.yml:48-49` keeps `group: nightly-release-train` with `cancel-in-progress: false`. The rationale for keeping the name is sound. The consequence is not covered:

While the `Production` environment still has required reviewers, the 09:00 UTC cron will push the version-bump commit to `main` and create the release tags in `prepare` (no environment binding there), then `deploy-server-prod` enters `Waiting` and holds the concurrency group indefinitely. Because `cancel-in-progress: false` and `hotfix-production.yml` no longer exists, an operator who needs a hotfix out cannot preempt that held run — the hotfix dispatch queues behind it, and a third dispatch cancels the queued one rather than the stuck one. This is the 36h-stall shape, now on the only remaining production path.

The spec says "while it is set, it stops the cron from completing unattended" (`specs/…/delivery/README.md:177`), which understates it: it also blocks every manual production deploy. Please either land the required-reviewer removal in the same window, or note in the PR that the cron should stay disabled until the settings flip, and add the hotfix-cannot-preempt consequence to that paragraph.

### R3 — The `skip_build` narrative contradicts the implementation two paragraphs later (fix before merge)

`specs/codebase/systems/engineering/delivery/README.md:175`: "so it deploys exactly the bytes that already exist for that ref." `release.yml:139-141` says the same ("promotes bytes that already exist for this exact ref"), and so does the PR description.

Line 183 of the same file then correctly states that every `_deploy-*.yml` lane builds its own exact-SHA image. Both cannot be true. What `skip_build` actually skips is the three artifact *release* builds, the version bump, the tags, and the release page — the hosted deploys still rebuild their images from source at that SHA. The behaviour is fine and deterministic; the claim is not, and this is a canonical-documentation file whose job is to describe the codebase. Reword to something like "it deploys the exact ref without minting new artifact identities" and drop the byte-reuse claim from the workflow comment too.

### R4 — The new "only entrypoint into a production deploy lane" guard is case-sensitive, and the two files it is replacing used the other case

`scripts/ci-cd/deploy-staging-trigger.test.mjs:53` filters on `/environment: Production/`. On `origin/main`, `hotfix-production.yml:248,260,270,280,290,303` and `promote-production.yml:155,167,177,187,197,207,220` all use `environment: production` (lowercase). So the guard being added would not have flagged either of the workflows this PR deletes, and will not flag a future reintroduction written the same way. Suggest `/environment:\s*["']?[Pp]roduction/` in both the filter at :53 and the `doesNotMatch` at :60.

### R5 — Two regex holes in `merging to main deploys nothing`

`scripts/ci-cd/deploy-staging-trigger.test.mjs:41` detects a main trigger with `/branches:\s*\[\s*main\s*\]/`, which only matches the YAML flow sequence. The block form

```yaml
push:
  branches:
    - main
```

evades it entirely. Separately, `triggerBlock` at :20-22 does `source.indexOf("\non:")` and on a miss (`-1`) falls through to `source.slice(-1, end)`, which yields `""` and silently classifies the workflow as safe rather than failing. Both are guard-quality nits, not live bugs — no current file hits either — but this test is the thing standing between main and an accidental auto-deploy.

### R6 — Three reusable deploy lanes are now called by nobody, and `e2b` has no automated production path at all

After the deletions, `_deploy-e2b.yml`, `_deploy-mobile.yml`, and `_deploy-workers.yml` have zero callers in `.github/workflows` (verified by grep). `prepare` still exports a `workers` output at `release.yml:64` that no job reads, and `e2b`/`mobile` are not even exported.

`detect-deploy-surfaces.mjs` maps `catalogs/**`, `Cargo.lock`, `Cargo.toml`, `scripts/build-template.mjs` and friends to the `e2b` surface, and `promote-production.yml` used to deploy it. Nothing in `release.yml` does. The E2B production rolling tag is now only reachable through the manual `Promote Cloud Template` workflow. That is defensible (the nightly train never had an e2b job either), but "the nightly covers every surface" in `release.yml:124` and in the spec is now literally false for e2b, and Current Gaps does not say so. Please add a gap bullet, and either delete the three orphaned reusables or state why they are being kept.

### R7 — A `skip_build` promote with default `surfaces` can be a green no-op

`changedFiles(base, head)` in `detect-deploy-surfaces.mjs` returns `[]` when `base === head`. On a promote run the operator supplies `ref` but is not required to supply `surfaces`, and `base_sha` resolves to the most recent `release-*` tag. If the promoted ref is that checkpoint — the single most likely thing an operator promotes — detection yields zero surfaces, every deploy job's `selected_surfaces != ''` guard (`release.yml:289,299,316`) is false, and the run goes green having deployed nothing. The only signal is a `Surfaces: ``` line in the step summary.

Promote is the "get these bytes to prod now" path; a silent success is the wrong failure mode for it. Suggest failing `prepare` when `skip_build == 'true' && selected_surfaces == ''`, or requiring an explicit `surfaces` list whenever `skip_build` is set.

### Observations, no action requested

- Accepted-by-ruling risk worth naming out loud: `release-server` (`server-ci.yml`) failing no longer stops `deploy-server-prod` from rolling that same SHA to production. On `main` the prod leg required a successful staging deploy of the same SHA first; now the only thing standing between a merge and production is `ci.yml` on the PR. That is the ruled model, and `_deploy-server.yml` has its own digest-pinning and health checks, but it is a genuine reduction in interlocks and it starts running unattended on merge.
- `deploy-web-prod` and `publish-product-release` use `always()` rather than `!cancelled()`. Under `always()`, clicking "Cancel workflow" during a prod rollout does not stop `deploy-web-prod` from starting. The pattern is inherited from the old train, so this is not a regression, but the unattended-cron posture makes `!cancelled()` the better default now.
- The deleted `promote-production.yml` and `hotfix-production.yml` targeted a lowercase `production` environment, i.e. not the `Production` environment carrying the reviewer rule and the prod vars. Deleting them removes a real bypass surface. Good outcome, worth recording somewhere.
- Docs style on the touched paragraphs: no em dashes introduced (the four in the file are all in untouched sections at :67, :71, :73, :80), no hard-wrapped prose, no approval language. Matches the ruled model.

pablonyx:
R1-R7 addressed at head `8f2ce8b479c7ef857485b46d7035731e632f0d89`. Per-finding detail and the negative-control table are in the PR description under "Review response"; the merge sequencing R2 asked for is its own **MERGE CONDITION** section.

Two notes on things that grew past the literal ask:

- **R6 fallout.** Deleting `_deploy-e2b.yml` broke two markdown links (`specs/FEATURE_DOCS/SANDBOX/lifecycle.md:95`, `guides/operating/e2b-template-operations.md:48`) that `check_docs.py` fails on, so the cleanup had to reach one prose mention in `support-loop.md` and three stale code comments as well. All repointed at the surviving cloud-template workflows.
- **R1.** I also dropped `Deploy Staging` from the same two rollback references. It has had no E2B job since #2126, so after the `_deploy-e2b.yml` deletion the only workflows that print an E2B job summary are `Release Cloud Template` and `Promote Cloud Template`.

Literal results on this head: `gh pr checks 2140` → **28 pass, 1 skipping** (`Production compose smoke`, path-gated), 0 fail. Local: `node --test scripts/ci-cd/*.test.mjs` → `# tests 207 / # pass 207 / # fail 0`; `check_docs.py` → `Documentation integrity check passed (234 Markdown files).`; `check_max_lines.py` → passed; ruby YAML parse → 24 workflow files.

One transient: `intent-billing (provisional)` failed the first run on this head with `T2-BILL-3 … expected 1 proration grant, received 2` (`tests/intent/specs/billing/seats.spec.ts:105`), a live-Stripe double-grant timing artifact. This diff contains no server, billing, or product code, the same lane passed on the previous head `a742417`, and the only delta between the two heads is three uncalled workflow deletions plus doc/comment/test edits. Re-ran the failed job only (run 32438440902) → `completed success`. Nothing deployed or released was dispatched.

Also worth flagging: `main` moved three commits since this branch was cut (#2132, #2136, #2137). #2137 rewrote `server-ci.yml`, which `release.yml` calls as `release-server`; I verified all five inputs it passes (`git_sha`, `comparison_sha`, `version`, `dry_run`, `publish`) still exist on `origin/main`. PR reports `MERGEABLE`.

pablonyx (commented):
## Re-review of record — PR #2140 @ `8f2ce8b479c7ef857485b46d7035731e632f0d89`

**Verdict: approve, conditional on the MERGE CONDITION being honored.** All seven findings are closed. I re-derived every claim independently rather than reading the closure report; the scope of the fix commit is exactly R1-R7 plus its declared blast radius, and the un-rebased branch does not need a rebase. Founder merge only, and the settings flip goes first.

### Scope of `a742417..8f2ce8b`

One commit, `8f2ce8b47 ci(release): close review findings R1-R7`, 12 files, +89/-370. Every hunk maps to a finding:

| Files | Finding |
| --- | --- |
| `guides/operating/e2b-template-operations.md` | R1 |
| `specs/…/delivery/README.md:177-181` | R2 |
| `specs/…/delivery/README.md:175`, `release.yml:143-146` | R3 |
| `deploy-staging-trigger.test.mjs:26-33,51,65,71` | R4, R5 |
| `_deploy-{e2b,mobile,workers}.yml` deleted, `release.yml:63` workers output dropped, `release.yml:123-127` comment, plus `lifecycle.md`, `support-loop.md`, `smoke-cloud-template.mjs`, `build-cloud-qualification-candidates.mjs`, `cloud-candidate-set.ts` | R6 fallout |
| `release.yml:142-150` guard, `deploy-staging-trigger.test.mjs:75-89` test | R7 |

No unrelated change, no drive-by refactor, no behavior touched outside the findings. The only surviving mention of the three deleted reusables is the deliberate one at `specs/…/delivery/README.md:50` explaining that they were deleted.

### Literal check results (fresh, on this head)

`gh pr checks 2140` on `8f2ce8b479c7ef857485b46d7035731e632f0d89`: all pass, no failures, no pending. `CI/CD config checks` = run 32438440907, job 96644187355, pass in 14s. `Repo shape checks`, `Workflow definition lifecycle (tier-2)`, `Validate PR title and labels` all pass. `Greptile Review` pass. PR state `MERGEABLE` / `CLEAN`, no longer a draft, labels `release:maintenance` + `area:release` + `area:docs`.

Scratch clone at `8f2ce8b47`:

- `node --test scripts/ci-cd/*.test.mjs` → `# tests 207 / # pass 207 / # fail 0` (206 before, +1 for the new R7 test).
- The two rewritten files → `# tests 11 / # pass 11 / # fail 0`, including `ok 5 - a deploy-only run refuses to resolve an empty surface set`.
- `python3 scripts/check_docs.py` → `Documentation integrity check passed (234 Markdown files).` This matters more than usual on this head: R6 deleted three files that four Markdown documents linked to, and the link checker is what proves the sweep was complete.

### Negative controls I built myself

Not the implementer's table. Five synthetic trees, each mutating exactly one property, run against the hardened suite:

| Control | Result |
| --- | --- |
| Hardened tests checked out onto the `origin/main` tree | `# pass 6 / # fail 5` — `not ok 1, 2, 3, 5, 8` |
| New workflow calling `_deploy-server.yml` with lowercase `environment: production` | `# pass 4 / # fail 1` — `not ok 3` only (**R4 proven**) |
| New workflow with `push:` → `branches:` block sequence → `- main`, calling `_deploy-web.yml` | `# pass 4 / # fail 1` — `not ok 2` only (**R5 branch form proven**) |
| R7 guard step surgically removed from `release.yml` | `# pass 4 / # fail 1` — `not ok 5` only (**R7 proven**) |
| Workflow file with no `on:` block at all | `# pass 4 / # fail 1` — `not ok 2` only (**R5 locate-assert proven**) |

Each control isolates a single test, which is the property I actually care about: the guards are specific, not blanket.

I also exercised `MAIN_BRANCH_FILTER` directly rather than trusting the backtracking to work out: `[main]` true, `[ main ]` true, `[main, dev]` true, block `- main` true, block `- "main"` true, block `- release` false, `[dev]` false. Correct in both directions.

### Per-finding closure

- **R1 closed, and over-delivered.** `guides/operating/e2b-template-operations.md` no longer names `Promote Production` anywhere; both `gh run list` lines are gone and step 2 now adds "No other workflow builds or moves an E2B tag", which is the accurate statement. The pass also caught `lifecycle.md` and `support-loop.md`, which I had not found.
- **R2 closed as documentation.** Assessed below.
- **R3 closed in all three places.** `specs/…/delivery/README.md:175` now reads "a promotion is a deterministic rebuild of the promoted commit rather than a retag of existing bytes", and `release.yml:143-146` matches. The spec no longer contradicts itself two paragraphs later.
- **R4 closed.** `PRODUCTION_ENVIRONMENT = /environment:\s*["']?[Pp]roduction["']?/` used at both the filter and the `doesNotMatch`. Proven by control 2.
- **R5 closed.** `triggerBlock(source, name)` asserts located and non-empty with the filename in the message; both `branches:` forms matched. Proven by controls 3 and 5.
- **R6 closed.** Three reusables deleted, `workers` output dropped, and `release.yml:120-127` now states the real set honestly: five acted-on surfaces, and e2b/mobile/workers explicitly no-ops. Three new Current Gaps bullets, including the one distinction I most wanted made — that `_deploy-server.yml`'s Celery worker and Beat rollout is a different thing from the retired `_deploy-workers.yml` no-op, so nobody reads the deletion as "workers stopped deploying".
- **R7 closed.** `release.yml:142-150` fails `prepare` with `::error::` and `exit 1` when `skip_build == 'true' && selected_surfaces == ''`. Placement verified: after `detect`, before `artifacts`, so it fails the job and skips every downstream deploy. It does not disturb the quiet-night path, since the condition requires `skip_build`. The message names both SHAs and suggests a concrete `surfaces` value.

### On R2: is documentation enough?

Yes, given the procedural enforcement, and the treatment is materially better than what I asked for. The spec at `:179` now walks the actual failure sequence — bump pushed, tags created, `deploy-server-prod` parked, group held, no hotfix workflow to preempt, third dispatch cancels the wrong run — and supplies `gh run cancel <run-id>` as the escape hatch. The PR body carries a standalone **MERGE CONDITION** heading. That is the right shape for a repository-settings dependency that no workflow file can express.

One consequence still unstated, worth one sentence somewhere before merge but not worth blocking on: if the cron fires in the window between merge and the settings flip and the parked run is then cancelled, `prepare` has already pushed the version-bump commit and created the `proliferate-v*` / `desktop-v*` / `server-v*` / `runtime-v*` tags, and the three release builds run unblocked (they touch no environment), so `publish-product-release` can publish a product release page for a version that never reached production. Nothing corrupts — `prepare-artifact-release.mjs` picks the next version past every existing tag, so the following run does not collide — but you burn a version number and ship a release page for bytes that are not live. The escape hatch releases the concurrency group; it does not undo that.

### Rebase question: not needed. Evidence

`main` moved by three commits since the merge base: #2136 (`release-desktop.yml`), #2137 (`server-ci.yml`), #2139 (`ci.yml`). `release.yml` calls the first two. Local `uses: ./.github/workflows/…` resolves at the caller's own commit, so post-merge both the caller and the callee come from the merge commit — the branch's `release.yml` will call **main's** `server-ci.yml`, and the only thing that can break is the input contract. I diffed it directly:

- `server-ci.yml` `workflow_call:` block, merge-base vs `origin/main`: **byte-identical**. All five inputs (`git_sha`, `comparison_sha`, `version`, `dry_run`, `publish`) present with unchanged requiredness and defaults, and no new required input added.
- `server-ci.yml` top-level `permissions:`, merge-base vs `origin/main`: **identical**, so `release-server`'s `contents: write` + `packages: write` grant still suffices.
- What #2137 actually changed is internal job topology: `test:` became `test-unit:` + `test-integration:`, and `docker:`/`self-hosted-release-assets:` shifted line numbers. `release.yml` references no `server-ci` job name and reads no `server-ci` output — `grep` for `needs.release-server.outputs` returns nothing — so the rename is invisible to the caller.
- `release-desktop.yml` was changed by #2136 (+30/-8), but its `workflow_call:` block is **byte-identical** to the merge base, and `release.yml` passes only `git_sha`, `version`, `dry_run`, `publish_updater`, `target_os`, all still present.
- `release-runtime.yml` is byte-identical merge-base vs main.
- GitHub reports the PR `MERGEABLE` / `CLEAN`; there is no textual conflict either.

So a rebase would change nothing about behavior. The one thing that is *not* this PR's problem but should be checked at flip time: if branch protection's required-check list names `Server CI / test`, #2137 already invalidated that name on main independently of this PR.

### Residual observations, non-blocking

- The R7 guard catches an *empty* surface set, not a *non-actionable* one. `skip_build=true` with `surfaces: mobile` (or `e2b`, or `workers`) still resolves a non-empty `selected_surfaces`, passes the guard, runs all three deploy jobs with `enabled=false`, and reports green having deployed nothing. This is now explicitly documented at `release.yml:123-127` ("none of them has a job here … so selecting them is a no-op"), which is why I am not asking for more: the operator has to name a surface the pipeline says it ignores. Tightening the guard to "no actionable surface" would be strictly better if it is ever cheap to do.
- PR description hygiene, cosmetic only: the per-file bullet for `release.yml` still says a promote run "deploys the exact bytes that already exist", the exact phrasing R3 removed from the spec and the workflow; and the `deploy-staging-trigger.test.mjs` bullet still says "3 of the 4 fail there, 4 of 4 pass here" when it is now 4 of 5 failing on main and 5 of 5 passing here. The shipped artifacts are correct; only the description lags.

### Standing items for the merge itself

1. Remove `required_reviewers` from the `Production` GitHub Environment **first**, or disable the `Release` schedule until it is removed. This is the MERGE CONDITION and it is the one thing that decides whether the first cron run is a release or a stall.
2. Reconcile any branch-protection required-check list that still names `Nightly Release Train`, `Hotfix Production`, or `Promote Production`.
3. `guides/deploying/releases.md` and `guides/deploying/hosted.md` remain stale by design, tracked as a Current Gaps bullet. The operator-guide rewrite should not lag far behind, since those are the two files an operator opens under pressure.

pablonyx:
`dry_run` re-admitted per Pablo's ruling. Head `ed55bdec854c1e2ac0f6054c1db59372d814cd1a`.

Fourth dispatch input, boolean, default false. What a dry run does:

| Stage | Dry-run behavior |
| --- | --- |
| prepare | Computes the version and tag plan so the plan is real, then throws it away. No commit to `main`, no tags. `prepare-artifact-release.mjs --dry-run true` writes no files, and the "Verify version tags are free" and "Create release tags" steps are gated off. |
| Release builds | Skipped. Their own `dry_run` input still compiles everything, and a macOS desktop matrix is not a proportionate price for a wiring proof. Same as the retired train's plan-only run. |
| Prod deploys | Still called, with `enabled: false`. That is the same no-op switch an unselected surface uses, so the `uses:` resolution, the input wiring, and the needs-graph edges are all exercised while the inner job that reaches AWS or Vercel never starts. This is the one place the new dry run goes further than the old train, which skipped these jobs outright. |
| publish-product-release | Runs with `--dry-run true`. `upsertRelease` returns `action: "dry-run"` without touching the API; `product-release.md` is still written, so the summary renders the body it would have published. |
| Summaries | Both the plan and the result summary carry a `Dry run` line, and the result summary leads with **DRY RUN. No commit, tag, artifact release, deploy, or product release page was produced.** |

Composition: `surfaces` narrows what the plan covers; `skip_build` walks the deploy-only shape and still requires an explicit `surfaces` list, because the R7 empty-surface guard is part of what a wiring proof should exercise. Documented in the input description, a comment above the deploy block, and the spec. Cron is unchanged: a scheduled run supplies no inputs, so `dry_run` is false.

One interlock adjustment: `deploy-web-prod`'s gate gained a `dry_run == 'true' ||` disjunct. Without it, a dry run's `enabled: false` server lane would read as not-success and gate web out of the graph walk, defeating the point. The failure interlock for real runs is untouched.

Tests: `release-title-propagation.test.mjs` now asserts the dispatch input set is **exactly** `["dry_run","ref","skip_build","surfaces"]`, scoped to the `workflow_dispatch` block. The obvious version of that assertion (an unscoped `[\s\S]*?dry_run:`) passes on a file with no `dry_run` input at all, because `dry_run:` also appears in the reusable `with:` blocks — worth knowing if anyone touches it. New `deploy-staging-trigger.test.mjs` case pins the whole no-op contract: the early return precedes `git push origin HEAD:main`, both tag steps are gated, all three build jobs carry the dry-run condition, all three deploy jobs still `uses:` their lane and carry the composed `enabled:` expression, publish passes the flag through, and both summaries say so.

Negative controls against the pre-`dry_run` `release.yml` from this branch's previous head: `not ok - a dry run walks the graph without any externally visible effect` and `not ok - the desktop title entrypoint survives the coordinator collapse` (input set 3, expected 4). Both pass on this head.

Literal results on `ed55bdec8`: `gh pr checks 2140` → **27 pass, 1 skipping** (`Production compose smoke`, path-gated), 0 fail. `CI/CD config checks` (run 32440238598, job 96649329442) → `# tests 208 / # pass 208 / # fail 0`. Local: `check_docs.py` → `Documentation integrity check passed (234 Markdown files).`; `check_max_lines.py` → passed; ruby YAML parse → 24 workflow files.

Process note: my first push of this commit silently did not land. The command chained `git push` behind the metadata validator with `&&`, and the validator failed on `area:server` — an artifact of validating against a moved `origin/main` rather than the merge base, since `main` picked up `server/tests/conftest.py` from #2137. Re-validated against the true merge-base file set (16 files, no `server/`) → `PR metadata looks good`, then pushed for real. The checks quoted above are from the correct head.

pablonyx (commented):
## Delta-review of record — `8f2ce8b47..ed55bdec8` (`dry_run` re-admitted)

**CLEAR to merge, after Pablo's two settings flips. Not before.**

I attacked the `deploy-web-prod` disjunct as instructed and could not weaken the real-run interlock; the equivalence is now proven mechanically rather than argued. Everything else in the commit is additive or suppression-only. All checks I quote below are pinned to `ed55bdec854c1e2ac0f6054c1db59372d814cd1a`.

### Checks, pinned to this head

I resolved runs by commit rather than by "latest", given the earlier wrong-head quote. `gh run list --commit ed55bdec854c1e2ac0f6054c1db59372d814cd1a` returns five runs, all `success`: CI `32440238598`, CodeQL `32440238573`, Intent Tests `32440238526`, Self-Host Smoke `32440238546`, PR Metadata `32440237291`. Run `32440238598` reports `headSha ed55bdec854c1e2ac0f6054c1db59372d814cd1a`, `conclusion success`.

Inside it, `CI/CD config checks` = job `96649329442`, `success`. Its literal output: `node --test scripts/ci-cd/*.test.mjs` → `# tests 208 / # pass 208 / # fail 0`, and the Ruby YAML parse lists 24 workflow files with `release.yml` present. `Repo shape checks` `96649329589` and `Workflow definition lifecycle (tier-2)` `96649329818` both `success`. Across the PR: 27 pass, 1 skipping (`Production compose smoke`), 0 fail, 0 pending. State `MERGEABLE` / `CLEAN`, not a draft.

Scratch clone at `ed55bdec8`: full suite `# tests 208 / # pass 208 / # fail 0`; the two touched files `12/12` including `ok 6 - a dry run walks the graph without any externally visible effect`; `check_docs.py` → `passed (234 Markdown files)`.

### THE ITEM: `deploy-web-prod`'s new `dry_run == 'true' ||` disjunct

I lifted the gate text verbatim out of `release.yml` and evaluated it as an expression rather than reasoning about it in prose.

**Real-run truth table** (`dry_run = 'false'`, `prepare` success, `selected_surfaces = 'server,web'`), 16 combinations of server-selected × server-deploy-result × web-selected:

| server selected | `deploy-server-prod` result | web job runs | Vercel touched |
| --- | --- | --- | --- |
| true | success | yes | only when web selected |
| true | failure | **no** | no |
| true | cancelled | **no** | no |
| true | skipped | **no** | no |
| false | success / failure / cancelled / skipped | yes | only when web selected |

Violations where web deploys despite a selected server lane not succeeding: **0**. The `server != 'true'` short-circuit rows are the pre-existing intended behavior, unchanged.

**Formal equivalence.** I evaluated the old (`8f2ce8b47`) and new (`ed55bdec8`) gate text for `deploy-web-prod` and `publish-product-release` across all 16,384 combinations each of `prepare.result` × `selected_surfaces` × `skip_build` × `product_tag` × `server`/`runtime`/`desktop` × four result values for each of the four upstream jobs, with `dry_run` pinned to `'false'`, and did the same for the three `enabled:` expressions. Result: **0 behavioral differences across 32,768 evaluations.** At `dry_run = 'false'` the new gates are byte-equivalent to the ones I approved.

**Why the disjunct cannot be abused.** The bypass and the safety are driven by the same job output in the same run: `dry_run == 'true'` opens the gate, and every lane's `enabled:` is `<surface> == 'true' && dry_run != 'true'`, so the identical value that lets the graph walk also forces `enabled: false` on all three lanes. Across the 16 dry-run rows, Vercel-touched is **0/16**. There is no state where the gate is bypassed and the inner job is live.

**Expression-evaluation subtleties, checked:**
- `dry_run` is a job output, therefore always a string; `== 'true'` is a string comparison, never boolean coercion.
- Its only source is `DRY_RUN: ${{ github.event.inputs.dry_run || 'false' }}`. On `schedule`, `github.event.inputs` is absent → `'false'`. On dispatch, a typed boolean arrives in the payload as the string `"false"`/`"true"`, which is non-empty and therefore truthy, so `||` does not swallow an unchecked box. Cron is unaffected, exactly as the spec now states.
- `grep -nE '\$\{\{[^}]*[^.]\binputs\.'` over `release.yml` still returns nothing — no bare `inputs.*` that would evaluate empty on a schedule event.
- If the output were ever empty, `'' == 'true'` is false, so the failure direction is toward the strict gate.

### The two new test assertions

**Scoping is correct, and the false-pass hazard is real.** `release-title-propagation.test.mjs` now slices `workflow_dispatch:` → `\npermissions:` and asserts the 6-space-indented key set equals `["dry_run","ref","skip_build","surfaces"]`. Input names sit at 6 spaces and their properties at 8, so the regex cannot pick up `description:`/`type:`. I confirmed the hazard the comment cites: against the pre-`dry_run` `release.yml`, a naive `/workflow_dispatch:[\s\S]*?dry_run:/` returns **true**, matching `"dry_run: false"` inside a reusable `with:` block. The scoped assertion returns false on that same file. Good catch by the implementer.

**Negative control.** Pre-`dry_run` `release.yml` (from `8f2ce8b47`) dropped into the `ed55bdec8` tree → `# pass 10 / # fail 2`: `not ok 6` and `not ok 9`, and nothing else. Both new assertions fail on the old file, and only those.

**Mutation testing.** I did not settle for wholesale absence. Five surgical single-line mutations, each reverted before the next:

| Mutation | Result |
| --- | --- |
| Drop `&& dry_run != 'true'` from `deploy-web-prod`'s `enabled:` (a dry run would really deploy web) | `not ok 6` only |
| Drop `\|\| "$DRY_RUN" == "true"` from the commit early return (a dry run would push to main) | `not ok 6` only |
| Let `Create release tags` run on a dry run | `not ok 6` only |
| Let `release-desktop` build on a dry run | `not ok 6` only |
| Hardcode `publish-product-release` to `--dry-run false` | `not ok 6` only |

Every escape route the test claims to close is genuinely closed, and the test is specific rather than blanket. The `deployJobs` `deepEqual` is a useful ratchet besides: adding a fourth deploy lane forces someone to revisit `dry_run`.

### Behavior claims, verified against the scripts rather than the description

- **`prepare` plans but writes nothing.** In `prepare-artifact-release.mjs`, `dryRun` guards only `writeText`/`writeJson`/`replaceCargoVersion` (:209-233), while `writeGithubOutput` (:363) is unconditional. So a dry run emits a real `product_tag` and real versions while touching no file. That matters twice over: the plan in the summary is genuine, **and** `publish-product-release`'s `product_tag != ''` condition still holds, so the publish leg is actually exercised rather than silently skipped. Two independent guards stop the push — no files are dirty, so `git diff --quiet` would exit early anyway, and the explicit `$DRY_RUN` early return precedes `git push origin HEAD:main`.
- **Publish renders without writing.** `publish-product-release.mjs` writes `--body-out` at :614 before `upsertRelease`, which returns `{action:"dry-run"}` at :502-505 ahead of any API call. Body rendered, nothing created or updated.
- **Deploy lanes called with `enabled: false`.** `_deploy-{server,web,litellm}.yml` each gate their single job on `if: ${{ inputs.enabled }}`, so `uses:` resolution, input wiring, and the `needs` edges are all exercised while the AWS/Vercel job never starts.
- **Build lanes skipped, not called with their own `dry_run`.** I agree with the asymmetry and with documenting it as deliberate: `release-desktop.yml`'s own dry run still compiles a macOS matrix, which is not a proportionate price for a wiring proof, and this matches the retired train's plan-only shape. The deploy lanes have a genuine no-op switch, so calling them is cheap and proves the wiring. The comment at the release-builds header states exactly this.
- **Both summaries lead with it.** `Summarize plan` emits the bolded dry-run line and `Summarize results` emits `**DRY RUN. …**`. The bash `if` inside the brace group is well-formed and a false condition returns 0, so it cannot trip the step.

### No other real-run path changed

Every remaining hunk is one of: the new dispatch input, the new `prepare` output, the new `DRY_RUN` env var, a `--dry-run false` literal becoming `--dry-run "${{ … }}"` that resolves to `false` on a real run, an additive `&& dry_run != 'true'` conjunct that is `true` on a real run, a summary echo, or a comment. The `Require an explicit surface set for a deploy-only run` guard from R7 is untouched and still sits between `Detect release surfaces` (:129) and `Prepare product and artifact versions` (:172). Spec prose updated to four inputs, and no "three inputs" claim survives anywhere in `specs/`, `guides/`, or `scripts/`.

### Residual, non-blocking

The input description and the spec both say a `skip_build` dry run "still requires an explicit `surfaces` list". Strictly, the R7 guard fires only when detection resolves *empty*; `skip_build` + `dry_run` + `surfaces=all` against a ref with real changes since the checkpoint passes without an explicit list. The sentence is a slight overstatement of a guard that is otherwise correct — worth a word trim whenever this file is next touched, not now.

### Verdict

**CLEAR.** The `dry_run` re-admission is correctly scoped, provably inert on real runs, and better tested than the code it guards. My R1-R7 closure verdict stands unchanged.

Merge remains gated on, in order:

1. Remove `required_reviewers` from the `Production` GitHub Environment (or disable the `Release` schedule until it is removed) — the MERGE CONDITION in the description.
2. Reconcile any branch-protection required-check list still naming `Nightly Release Train`, `Hotfix Production`, or `Promote Production`.

Both are Pablo's flips and neither is expressible in this diff. With those done, this is clear to merge.

PR #2142, opened 2026-08-21
ci(windows): probe whether the desktop crates compile for Windows

pablonyx:
## What this check actually covers, and what it does not

Read directly from `.github/workflows/windows-compile-check.yml` at this PR's head (`58f07c205d8fa518e947bb6468ac347e37555e14`):

- The `runtime` job runs exactly: `cargo check --target x86_64-pc-windows-msvc -p anyharness -p proliferate-worker -p proliferate-supervisor`.
- The `desktop` job runs exactly: `cargo check --target x86_64-pc-windows-msvc -p proliferate --keep-going`.

Neither invocation passes `--all-targets`, and neither is `cargo test`. So neither job compiles any test code: not the crates' `#[cfg(test)]` unit tests, and not any `tests/*.rs` integration tests. A green run here is a verdict on the ordinary build targets only, and says nothing about whether test code compiles for Windows.

That gap is not hypothetical. `.github/workflows/windows-process-kill-tests.yml` (added in #2147) documents in its own header that `cargo check --all-targets --target x86_64-pc-windows-msvc` reports ten pre-existing errors in `anyharness-lib`, every one a test-only POSIX assumption in an unrelated module, and that this is exactly why that workflow routes through an integration test at `tests/windows_process_tree.rs` rather than the crate's unit tests: an integration test compiles as its own crate against the library's public surface, so it never pulls in the crate's `#[cfg(test)]` modules, and can get a real Windows runtime verdict without that port being done first. So a green `windows-compile-check.yml` run does not mean `cargo test --lib` compiles for `anyharness-lib` on Windows. This workflow does not even attempt that build, in either job.

The single most important thing to know about this workflow: both jobs are `continue-on-error: true`, confirmed by reading the file at this PR's head, on both the `runtime` job and the `desktop` job. That means this workflow's outcome is unconditional: it can never fail the overall PR check run, and a real cargo error inside either job still leaves that job's, and the run's, conclusion `success`. A green check mark next to "Windows compile check" on a PR does not mean cargo succeeded; it means the job finished, which happens whether cargo passed or failed. The only way to know whether it actually passed is to open the job and read whether `cargo check` reported errors (the `desktop` job's step summary lists them; the `runtime` job's do not get a summary step, so its raw log is the only source there).

## Result: the desktop crate already compiles for Windows

This job was a probe, not a fix. Every Windows estimate anyone gave was guesswork because nothing had compiled the desktop crate for Windows since desktop Windows releases were disabled on 2026-04-13 by `3d3c0504b8`. This turns the unknown into an error list.

The error list is empty.

```
Finished `dev` profile [unoptimized + debuginfo] target(s) in 3m 31s
```

`cargo check --target x86_64-pc-windows-msvc -p proliferate --keep-going` on `windows-latest`: zero errors, dead-code warnings only. The control job over the runtime crates is green as expected, so the toolchain is not flattering the result.

## What I predicted, and why I was wrong

I wrote in the original body that the first failure would be `apps/desktop/src-tauri/src/diagnostics_collector/mod.rs:3`, where `pub(crate) mod child_bridge;` is declared with no cfg gate over roughly ten files of POSIX code: `std::os::fd`, `UnixStream`, `pre_exec`, socketpair.

The declaration really is ungated and the POSIX code really is there. The gating is one level further down, which reading only `mod.rs` does not show. `child_bridge/mod.rs` says so in its own doc comment:

> Each submodule self-gates to supported macOS targets, so unsupported Unix and non-Unix builds compile these declarations to empty modules and expose no Unix bridge types.

So the module compiles to nothing on Windows and there is nothing to fix. This is the general lesson from the run: static reading of a Rust module tree systematically over-predicts platform breakage, because `#[cfg]` composes downward and you cannot see it from the parent. Three and a half minutes of a Windows runner settled what a careful read got wrong.

## What this changes

The remaining work for a Windows beta is smaller than the disabled release matrix suggests, and none of it is "port the desktop app".

- The runtime crates already build and ship `anyharness-x86_64-pc-windows-msvc.zip` on every release. That never stopped.
- The stated reason for the April disable, from that commit's own doc note, was "until the SDK generation step is Windows-safe". That step turns out to be a single `mkdir -p` in `anyharness/sdk`'s `generate:openapi` npm script, which fails under `cmd.exe`. Fixed in #2146.
- `specs/desktop-native.md` still carries Windows-specific updater logic, so the code was never actually stripped of Windows awareness.

## Genuinely still broken on Windows, found by reading rather than compiling

Compiling clean is not the same as working. These are runtime defects a compile check cannot catch, each now owned by its own change:

1. `anyharness/crates/anyharness-lib/src/process_kill.rs:352-362` returns `(0, 0)` from both `kill_group_and_await` and `kill_session_and_await` under `cfg(not(unix))`. That is the identical tuple a successful kill of zero processes returns, so stop and cancel report clean quiescence while the whole agent process tree keeps running.
2. The agent catalog resolved no Windows pins at all, so no agent could install. Fixed in #2145, which also uncovered that `installer/pinned.rs:41` writes the downloaded binary to an extension-less path, so a Windows pin would verify its checksum and then land somewhere Windows cannot execute.
3. Hardcoded `/bin/sh` spawns and `HOME` reads in the runtime.
4. `release-desktop.yml` has no Windows matrix entry and `scripts/generate-updater-manifest.mjs` has no `windows-x86_64` key, so a Windows build would never be produced and, if produced, would never update.

## Should this workflow merge?

It is `continue-on-error` on both jobs and non-blocking by design, so it cannot fail anyone's PR. Its value now inverts: it started as a probe and its result makes it a regression guard, because the thing it proves green is exactly the thing nobody was watching for four months. Keeping it costs about three and a half minutes on PRs that touch `apps/desktop/src-tauri/**` or `anyharness/crates/**`.

Worth flipping off `continue-on-error` once the surrounding Windows work lands, so it becomes a real gate rather than a probe.

### DIFF
--- a/.github/workflows/windows-compile-check.yml
+++ b/.github/workflows/windows-compile-check.yml
@@ -1,0 +1,69 @@
+name: Windows compile check
+
+# The runtime already builds and ships for Windows on every release, but nothing
+# has compiled the desktop crate for Windows since desktop releases were
+# disabled on 2026-04-13. That makes every Windows estimate guesswork. This job
+# turns the unknown into an error list.
+#
+# Non-blocking on purpose: it is a probe, not a gate, until the errors are fixed.
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "apps/desktop/src-tauri/**"
+      - "anyharness/crates/**"
+      - ".github/workflows/windows-compile-check.yml"
+
+permissions:
+  contents: read
+
+jobs:
+  runtime:
+    # Known green: release-runtime.yml ships these targets today. Here as the
+    # control, so a red desktop job cannot be blamed on the toolchain.
+    name: runtime crates (control)
+    runs-on: windows-latest
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: x86_64-pc-windows-msvc
+      - uses: Swatinem/rust-cache@v2
+      - name: cargo check
+        run: cargo check --target x86_64-pc-windows-msvc -p anyharness -p proliferate-worker -p proliferate-supervisor
+
+  desktop:
+    name: desktop crates (the unknown)
+    runs-on: windows-latest
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: x86_64-pc-windows-msvc
+      - uses: Swatinem/rust-cache@v2
+      - name: cargo check
+        # Keep going past the first failure so one run yields the whole list
+        # rather than one error at a time.
+        run: cargo check --target x86_64-pc-windows-msvc -p proliferate --keep-going 2>&1 | tee check.log
+        shell: bash
+      - name: Error summary
+        if: always()
+        shell: bash
+        run: |
+          echo "### Windows desktop compile errors" >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
+          grep -E '^error' check.log | sort | uniq -c | sort -rn | head -50 >> $GITHUB_STEP_SUMMARY || echo "no errors" >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
+          echo "total errors: $(grep -cE '^error' check.log || echo 0)" >> $GITHUB_STEP_SUMMARY
+          echo "files implicated:" >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
+          grep -oE '\-\-> [^:]+' check.log | sed 's/--> //' | sort -u | head -60 >> $GITHUB_STEP_SUMMARY || true
+          echo '```' >> $GITHUB_STEP_SUMMARY
+      - uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: windows-desktop-check-log
+          path: check.log
--- a/.github/workflows/windows-compile-check.yml
+++ b/.github/workflows/windows-compile-check.yml
@@ -14,6 +14,11 @@ on:
       - "apps/desktop/src-tauri/**"
       - "anyharness/crates/**"
       - ".github/workflows/windows-compile-check.yml"
+      - "Cargo.lock"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 permissions:
   contents: read
@@ -24,9 +29,10 @@ jobs:
     # control, so a red desktop job cannot be blamed on the toolchain.
     name: runtime crates (control)
     runs-on: windows-latest
+    timeout-minutes: 45
     continue-on-error: true
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: x86_64-pc-windows-msvc
@@ -37,9 +43,10 @@ jobs:
   desktop:
     name: desktop crates (the unknown)
     runs-on: windows-latest
+    timeout-minutes: 45
     continue-on-error: true
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: x86_64-pc-windows-msvc
@@ -62,7 +69,7 @@ jobs:
           echo '```' >> $GITHUB_STEP_SUMMARY
           grep -oE '\-\-> [^:]+' check.log | sed 's/--> //' | sort -u | head -60 >> $GITHUB_STEP_SUMMARY || true
           echo '```' >> $GITHUB_STEP_SUMMARY
-      - uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         if: always()
         with:
           name: windows-desktop-check-log


pablonyx (commented):
## Review: rebase + workflow + disclosure

**Rebase integrity: clean.** `git range-diff ac4c4caf1d..8de5a45697 956d2426b7..68777b5614` shows a single `1: 8de5a45697 = 1: 68777b5614` — byte-identical commit, nothing dropped or reordered across the new `main` (verified against pre-rebase tip `8de5a45697` as specified). Not a rebase concern.

**CI matches the claims.** Pulled the actual run at head `68777b5614` (job 96677420874): `cargo check --target x86_64-pc-windows-msvc -p proliferate --keep-going` finishes `Finished \`dev\` profile ... in 3m 36s` with no `error` lines in the log (PR body says "3m 31s" — cosmetic, presumably from the pre-rebase run). Control job (`runtime crates`) also green. `ci-ok`'s required-checks list (added in #2138, same file) does not include this workflow, and it's not part of that rollup since `needs:` can't cross workflows — consistent with `continue-on-error: true` + path-filtered = intentionally non-required. No conflict with the absence-is-fatal policy from #2138.

**Retraction (item 4): present and accurate.** The "What I predicted, and why I was wrong" section correctly retracts the `diagnostics_collector/mod.rs:3` prediction and states the general lesson (`#[cfg]` composes downward, parent `mod` declarations say nothing about gated submodules). No note needed there.

### Blocking: PR body doesn't disclose the `--all-targets`/lib-test gap (item 3)

The body says flatly "The error list is empty" and "zero errors, dead-code warnings only" — with **no mention anywhere** of `--all-targets`, the 10 errors found in the `lib test` target, or that `cargo test` is red on Windows while `cargo check` is green. I grepped the current body for "all-targets", "lib test", "test target", "10 error" — zero matches. As written, this reads as "Windows compiles clean," full stop, which is stronger than what was actually verified. Two PRs (#2147, #2148) are about to stack on this branch and inherit that impression. The judgment call itself (excluding `--all-targets` from the gate so the shared base doesn't go permanently red over a pre-existing, unrelated test-target failure) is the right call — but the body needs a line stating the `--all-targets` probe found ~10 errors confined to the `lib test` target (pre-existing, not introduced here) and that this workflow deliberately checks only the `lib`/bin target for that reason. Add it before merging.

### Should-fix (non-blocking, but this file is now a 3-PR shared base — cheap to fix now, annoying to fix three times later)

1. **Unpinned actions, against 100% repo convention.** `.github/workflows/windows-compile-check.yml:29,42` (`actions/checkout@v4`) and `:65` (`actions/upload-artifact@v4`) are the only unpinned instances of these actions anywhere in `.github/workflows/` — every other usage (65/65 checkout, 21/22 upload-artifact) is SHA-pinned with a version comment, e.g. `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1`. This isn't cosmetic: the live run log for this exact commit already shows `##[warning]Node.js 20 is deprecated... actions/checkout@v4, actions/upload-artifact@v4` — the floating `@v4` tags resolve behind the pinned `v7.0.1` the rest of the repo uses. (`dtolnay/rust-toolchain@stable` and `Swatinem/rust-cache@v2` match existing repo usage elsewhere, so those are fine as-is.)
2. **No `timeout-minutes` on either job** (lines 26-27, 39-40). Default job timeout is 360 minutes. Both jobs are `continue-on-error: true`, so a hung `windows-latest` runner (network stall, linker hang — not unheard of on Windows runners) burns hours of 2x-cost compute silently with no PR-visible failure. Other timed jobs in `ci.yml` cap at 20-30 min; recommend similar here.
3. **No `concurrency` group.** Every other ref-scoped, PR-triggered probe/smoke workflow in this repo (`codeql.yml`, `self-host-smoke.yml`) sets `concurrency: group: <name>-${{ github.ref }}, cancel-in-progress: true`. This workflow has none, so rapid pushes to a PR touching `anyharness/crates/**` (a very hot path — this exact directory has a dozen files under active review in other worktrees right now) will stack redundant `windows-latest` runs instead of cancelling stale ones.
4. **Path filter gap:** `pull_request.paths` (lines 13-16) doesn't include root `Cargo.lock`. A pure dependency bump (there's literally an open dependabot cargo minor/patch PR right now) can change whether Windows compiles without touching any listed path, so it would silently skip this probe. Minor since this is advisory rather than gating, but worth a line if it's meant to catch dependency-driven regressions too.

### Verdict

Not safe to merge silently as-is: fix the missing `--all-targets`/lib-test disclosure in the PR body before merging — that's the one thing this review was specifically asked to gate on, and it's currently absent, not just incomplete. The four should-fix items are real but don't block: both jobs are `continue-on-error`, so nothing here can propagate a functional defect to #2147/#2148 — only cost/hygiene debt on a file three PRs now depend on.

pablonyx (commented):
## Confirming review — commit `58f07c205` only

Scope: this pass covers only the final commit `58f07c205` ("ci(windows): pin actions, bound runtime, scope concurrency"), landed after the earlier full review of #2142. Everything reviewed before that commit is out of scope here and not relitigated.

**Verdict: non-blocking. This commit does what it claims and looks safe.**

### 1. Does the commit do exactly three things?
Yes, plus the path-trigger fix the commit message also calls out (four changes total, all named in the commit body — nothing undisclosed):
- Pins `actions/checkout` and `actions/upload-artifact` to SHAs (2 call sites each x2 jobs = 4 lines).
- Adds `timeout-minutes: 45` to both jobs.
- Adds a `concurrency:` block at workflow scope.
- Adds `Cargo.lock` to the `pull_request.paths` trigger list.

`dtolnay/rust-toolchain@stable` and `Swatinem/rust-cache@v2` are deliberately left unpinned, which the commit message justifies as matching repo convention (19 and 6 uses respectively). Confirmed no other changes in the diff (+10/-3, matches the reported file stat).

### 2. Action pinning
Both pins are full commit SHAs with a version comment, not floating tags:
- `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1`
- `actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1`

I independently resolved both SHAs against their own upstream repos (`actions/checkout` and `actions/upload-artifact`) via `gh api repos/<owner>/<repo>/git/refs/tags/v7.0.1` — both SHAs match the `v7.0.1` tag exactly in their respective repos. Correctly pinned, comments are accurate.

### 3. Concurrency group
```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```
This workflow only triggers on `pull_request` and `workflow_dispatch` — there is no `push` trigger, so there's no PR-vs-push collision to worry about. For `pull_request` events `github.ref` resolves to `refs/pull/<PR#>/merge`, which is unique per PR, so the group can't collide across different PRs either. I checked this against other ref-scoped workflows on `main`: `codeql.yml` uses `group: codeql-${{ github.ref }}` and `self-host-smoke.yml` uses `group: self-host-smoke-${{ github.ref }}`, both with `cancel-in-progress: true` — this workflow's `${{ github.workflow }}-${{ github.ref }}` is the same pattern (just deriving the prefix from the workflow name instead of hardcoding it). The commit message's "matches repo convention" claim checks out.
`cancel-in-progress: true` is the right call here: this is a non-blocking probe with `continue-on-error: true` on both jobs, so canceling a stale run on new pushes to the PR just saves Windows-runner minutes with no correctness downside.

### 4. Timeout sanity
`timeout-minutes: 45` on both jobs. Pulled real timing data instead of trusting the number:
- First-ever run of this workflow (genuine cold cache, no `Swatinem/rust-cache` hit): `cargo check` on the desktop crate finished in **3m 31s** (log: `Finished `dev` profile [unoptimized + debuginfo] target(s) in 3m 31s`), full job wall time ~4m40s.
- Current-head run (warm cache, `Cache hit for: v0-rust-desktop-...`): compile finished in 32.96s.

45 minutes is roughly 10x the observed cold-cache time — sane headroom, not flake-prone, and still bounds a genuine `windows-latest` hang instead of letting it run to the 360-minute default under `continue-on-error`.

### 5. Path triggers
```
paths:
  - "apps/desktop/src-tauri/**"
  - "anyharness/crates/**"
  - ".github/workflows/windows-compile-check.yml"
  - "Cargo.lock"
```
Confirmed coherent and functional: the desktop job's compiled package (`-p proliferate`) actually lives at `apps/desktop/src-tauri` — verified directly from the job log: `Compiling proliferate v0.4.22 (D:\a\proliferate\proliferate\apps\desktop\src-tauri)`. So the `apps/desktop/src-tauri/**` path does cover the crate the job probes, and `anyharness/crates/**` covers the runtime control job's packages. Adding `Cargo.lock` closes the real gap described (a dependency-only version bump wouldn't touch either crate path but does touch the lockfile) and will fire correctly.

### 6. Current CI state at head (58f07c205) — read from logs, not checkmarks
Latest run for this branch/workflow: run [32450914102](https://github.com/proliferate-ai/proliferate/actions/runs/32450914102), triggered at the head commit, `event: pull_request`.
- Job `runtime crates (control)`: conclusion `success`, 05:32:22–05:34:19.
- Job `desktop crates (the unknown)`: conclusion `success`, 05:32:22–05:34:05.

This is a genuine pass, not `continue-on-error` masking a failure — pulled the raw job log:
```
Cache hit for: v0-rust-desktop-Windows_NT-x64-368f6b88-c30d2aef
Compiling proliferate v0.4.22 (D:\a\proliferate\proliferate\apps\desktop\src-tauri)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 32.96s
```
Zero lines matching `^error` in `check.log`. So at this commit the desktop crate genuinely compiles clean for Windows (a positive finding beyond this commit's scope, but relevant to "is it still green").

### 7. Stacked PRs #2147 / #2148
Confirmed both are based on `ci/windows-compile-check` (`baseRefName` = `ci/windows-compile-check` for both via `gh pr view`). The diff of `58f07c205` renames nothing: workflow name (`Windows compile check`), job ids (`runtime`, `desktop`), and job display names (`runtime crates (control)`, `desktop crates (the unknown)`) are byte-identical before and after this commit — only lines were added inside existing steps/jobs. No breakage risk for anything the stacked PRs' own reviews relied on.

### 8. `area:release` label / "Validate PR title and labels" check
Confirmed. There were two runs of this check on the PR head: one `FAILURE` at 2026-08-21T07:54:41Z (before the label was added — "Changed paths require area label(s): area:release") and a later one that is now `pass` (run [32460886002](https://github.com/proliferate-ai/proliferate/actions/runs/32460886002), completed 2026-08-21T07:56:56Z, 8s). `gh pr checks 2142` currently shows `Validate PR title and labels  pass`. Current PR labels: `release:maintenance`, `area:desktop`, `area:release`. The gate is satisfied now.

### On your stated assumptions
Everything you stated checks out — nothing to correct. One nuance worth flagging: point 4 asked me to confirm the timeout accounts for a workflow that "has historically taken a long time on a cold cache." The actual historical data doesn't support that framing — the one genuine cold-cache run on record took under 5 minutes total, not "a long time." That doesn't change the verdict (45 min is still a safe, non-flaky bound either way), just noting the premise doesn't match what the logs show.

No blocking issues found in this commit. Nothing merges without Pablo.

PR #2143, opened 2026-08-21
fix(desktop): compile the desktop crate for windows

pablonyx:
Work in progress. The before and after error counts, quoted from the Windows CI job, land here once the first run on this branch reports.

Stacked on `ci/windows-compile-check` (draft PR #2142). Merge this after that one, or rebase onto main once it lands.

### DIFF
--- a/.github/workflows/windows-compile-check.yml
+++ b/.github/workflows/windows-compile-check.yml
@@ -67,3 +67,40 @@ jobs:
         with:
           name: windows-desktop-check-log
           path: check.log
+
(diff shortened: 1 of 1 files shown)


pablonyx:
Superseded. The premise behind this PR (desktop crate does not compile for Windows) is false: the default target set is already green. The widened all-targets probe on this branch did find 10 real errors, but they are all in the `lib test` target, not the shipping build. Branch `fix/desktop-windows-compile` is left pushed if anyone wants to revive the widened probe.

Issue #2141, opened 2026-08-21
[Bug]: linux_arm64 agent pins missing from catalog — claude/codex uninstallable on arm64 self-host


pablonyx:
### Bug Description

On self-hosted Linux aarch64 (Graviton-class arm64), `POST /v1/agents/claude/install` (and codex) fails with `AGENT_NO_PIN_FOR_PLATFORM`, and the startup reconcile pass logs the same failure for every arm64 install attempt. Root cause: `scripts/agent-catalog/resolve-pins.mjs` resolves catalog pins for only 3 of the 6 platforms the registry declares, so `linux_arm64` never gets a `{url, sha256}` target even though upstream publishes arm64 builds for the affected agents. Combined with install-order sequencing, this makes `claude` and `codex` completely uninstallable on arm64 Linux self-host — no session can run a turn on that architecture.

### Steps to Reproduce

1. Provision an aarch64 Ubuntu 24.04 host (e.g. AWS Graviton EC2) and self-host the `server-v0.4.20` release (`server/deploy/install.sh`, which explicitly supports `aarch64`/`arm64` via `uname -m` detection at `server/deploy/install.sh:264-268`), running the `anyharness-aarch64-unknown-linux-musl.tar.gz` runtime binary.
2. `POST /v1/agents/claude/install`.
3. Response: `HTTP 400`
   ```json
   {"title":"No pinned agent download for this platform","status":400,"detail":"pinned source has no download for this platform: linux_arm64","code":"AGENT_NO_PIN_FOR_PLATFORM"}
   ```
4. Runtime logs confirm this for multiple agents on startup reconcile:
   ```
   WARN anyharness_lib::domains::agents::installer::reconcile: reconcile: install failed for agent agent="claude" error=pinned source has no download for this platform: linux_arm64 error_kind="other"
   WARN anyharness_lib::domains::agents::installer::reconcile: reconcile: install failed for agent agent="codex" error=pinned source has no download for this platform: linux_arm64 error_kind="other"
   ```
5. `GET /v1/agents/claude` then shows:
   ```json
   {
     "installState": "failed",
     "readiness": "install_required",
     "credState": "ready",
     "native": {"role":"native_cli","installed":true,"source":"path","path":"/home/ubuntu/.npm-global/bin/claude"},
     "agentProcess": {"role":"agent_process","installed":false,"message":"Not installed. Use the install endpoint to set up."}
   }
   ```
[...]

PR #2145, opened 2026-08-21
fix(agents): resolve pins for windows and arm64 linux

pablonyx:
Fixes the agent catalog so agents can install on Windows and on arm64 Linux. Refs #2141.

## Root cause

`scripts/agent-catalog/resolve-pins.mjs:30` declared only three platforms:

```js
const DEFAULT_PLATFORMS = ["macos_arm64", "macos_x64", "linux_x64"];
```

Both platform loops in the resolver (`resolve-pins.mjs:128` for `direct_binary` and `resolve-pins.mjs:144` for `tarball_release`, plus `:205` for `registry_backed` and `:224` for `direct_archive`) skip any key not in that set, so `linux_arm64`, `windows_x64` and `windows_arm64` were never written into `catalogs/agents/catalog.json` even though `catalogs/agents/registry.json` already declared all six in its `platformMap`s (claude at `registry.json:16`, codex at `registry.json:165`).

At install time `pick_target()` in `anyharness/crates/anyharness-lib/src/domains/agents/installer/pinned.rs:245` looks the pin up by `Platform::registry_key()` and returns `InstallError::NoPinForPlatform` when it is absent, surfaced as HTTP 400 `AGENT_NO_PIN_FOR_PLATFORM`. On aarch64 Linux that means claude and codex both fail to install and no session can run a turn.

Two upstream naming facts also had to be fixed, or the newly-emitted pins would have pointed at 404s:

- claude's `binaryUrlTemplate` hard-coded the filename `claude`. The Windows artifact is `claude.exe`.
- codex's Windows release assets are `codex-<target>.exe.tar.gz`, with the `.exe` before the `.tar.gz`. The plain `codex-<target>.tar.gz` name does not exist for Windows targets.

## Verified artifacts

Every URL below was checked before a pin was written. Direct HTTP checks are `curl -sIL -o /dev/null -w '%{http_code}'`. The three lanes that already worked are included for completeness.

### claude native, pin `2.1.234` (`direct_binary`)

Checksums come from the provider's `claude-code-releases/2.1.234/manifest.json`, which also declares the per-platform artifact filename.

| platform key | artifact URL suffix | evidence |
| --- | --- | --- |
| macos_arm64 | `2.1.234/darwin-arm64/claude` | HTTP 200 (already pinned, unchanged) |
| macos_x64 | `2.1.234/darwin-x64/claude` | already pinned, unchanged; manifest `binary: claude` |
| linux_x64 | `2.1.234/linux-x64/claude` | HTTP 200 (already pinned, unchanged) |
| linux_arm64 | `2.1.234/linux-arm64/claude` | **HTTP 200**, manifest `binary: claude`, checksum `24adda67…`, 325507304 bytes |
| windows_x64 | `2.1.234/win32-x64/claude.exe` | **HTTP 200**; `2.1.234/win32-x64/claude` (no suffix) is **HTTP 404**; manifest `binary: claude.exe`, checksum `3f877e78…`, 324028576 bytes |
| windows_arm64 | `2.1.234/win32-arm64/claude.exe` | **HTTP 200**; `2.1.234/win32-arm64/claude` is **HTTP 404**; manifest `binary: claude.exe`, checksum `ed6cc3d0…`, 311927456 bytes |

### codex native, pin `rust-v0.147.0` (`tarball_release`)

Checksums are the GitHub release assets' published `digest`. The inner binary name was confirmed by streaming the first 2 MB of each archive through `tar -t`.

| platform key | asset | evidence |
| --- | --- | --- |
| macos_arm64 | `codex-aarch64-apple-darwin.tar.gz` | already pinned, unchanged |
| macos_x64 | `codex-x86_64-apple-darwin.tar.gz` | already pinned, unchanged |
| linux_x64 | `codex-x86_64-unknown-linux-musl.tar.gz` | already pinned, unchanged |
| linux_arm64 | `codex-aarch64-unknown-linux-musl.tar.gz` | **HTTP 200**, digest `eb677c80…`, 91607658 bytes, inner binary `codex-aarch64-unknown-linux-musl` |
| windows_x64 | `codex-x86_64-pc-windows-msvc.exe.tar.gz` | **HTTP 200**, digest `7c47d4dc…`, 99655443 bytes, inner binary `codex-x86_64-pc-windows-msvc.exe`; `codex-x86_64-pc-windows-msvc.tar.gz` is **HTTP 404** |
| windows_arm64 | `codex-aarch64-pc-windows-msvc.exe.tar.gz` | **HTTP 200**, digest `c3a7da6c…`, 91259737 bytes, inner binary `codex-aarch64-pc-windows-msvc.exe`; `codex-aarch64-pc-windows-msvc.tar.gz` is **HTTP 404** |

### agent-process pins (all agents)

| agent | source kind | platform dependence | evidence |
| --- | --- | --- | --- |
| claude | git (`proliferate-ai/claude-agent-acp@65f6205`) | none | commit-anchored, installed through the node runtime; no per-platform targets exist |
| codex | git (`proliferate-ai/codex-acp@219738c`) | none | same |
| grok | npm `@xai-official/grok@0.2.102` | none | `npm view @xai-official/grok@0.2.102` returns `version = 0.2.102`, `dist.integrity = sha512-Q8Bdego…` |
| cursor | archive via ACP registry | per platform | see "What remains unsupported" |
| opencode | archive via ACP registry | per platform | see "What remains unsupported" |

## What changed

- `scripts/agent-catalog/resolve-pins.mjs`: `DEFAULT_PLATFORMS` now lists all six keys `Platform::registry_key()` can report. Added a `{binary}` placeholder to `direct_binary` URL templating, resolved per platform from a new optional `binaryNameMap` (defaulting to the agent kind, so the field is optional and backward compatible). Added `--keep-versions`, plus a `githubReleaseByTag` helper so `tarball_release` can re-resolve a pinned tag rather than `latest`.
- `catalogs/agents/registry.json`: claude's `binaryUrlTemplate` ends in `{binary}` and gains a `binaryNameMap` (`claude.exe` on the two Windows keys). codex's Windows `platformMap` values become `x86_64-pc-windows-msvc.exe` / `aarch64-pc-windows-msvc.exe`, which makes the existing `codex-{target}.tar.gz` and `codex-{target}` templates produce exactly the published asset name and the correct inner binary with no code change. `registryVersion` bumped to `2026-08-20.1`.
- `catalogs/agents/registry.schema.json`: declares the new optional `binaryNameMap` (`direct_binary` is `additionalProperties: false`).
- `catalogs/agents/catalog.json` and `scripts/agent-catalog/catalog.draft.json`: regenerated, `catalogVersion` bumped to `2026-08-20.1` and `probedAgainst.registryVersion` realigned. Six new targets, nothing else touched. Regenerated with:

  ```
  node scripts/agent-catalog/resolve-pins.mjs --keep-versions
  cp catalogs/agents/catalog.json scripts/agent-catalog/catalog.draft.json
  ```

- `scripts/agent-catalog/resolve-pins.test.mjs`: two new tests covering per-platform Windows artifact naming and the `--keep-versions` refusal to adopt a newer ACP release.
- `anyharness/crates/anyharness-lib/src/domains/agents/model.rs`: doc comment only, documenting `{binary}`. No behavior change.
- `specs/codebase/platforms/product/agent-distribution.md`: documents the platform matrix and `--keep-versions` in the update pipeline section.

### Why `--keep-versions` exists

A plain re-resolve does not just add platforms, it also fetches upstream `latest`: claude is now `2.1.238` (pinned `2.1.234`), codex `rust-v0.149.0` (pinned `rust-v0.147.0`), and the ACP registry serves cursor `2026.08.11`, opencode `1.18.19`, grok `1.0.8`. Adopting those would break `validate-agent-catalog.mjs`'s snapshot-evidence rule, which requires each pin to match the committed probe snapshots in `scripts/agent-catalog/generated/`, and would smuggle five unprobed harness upgrades into a pin-coverage fix. `--keep-versions` re-resolves the versions the lockfile already pins, so the diff is exactly the six new targets. Version refreshes stay where they belong, in a probe-backed `make catalog-update`.

## What remains unsupported

- **cursor and opencode on `linux_arm64`, `windows_x64`, `windows_arm64`.** Not a missing artifact. Both resolve through the ACP registry, which serves only `latest`, and `latest` has moved past what this catalog pins (cursor `2026.08.11` vs pinned `2026.07.09`; opencode `1.18.19` vs pinned `1.18.3`). Taking the new-platform URLs from `latest` would pin those platforms to a different release than the macOS/linux-x64 targets and would break opencode's probe attestation. Under `--keep-versions` the resolver keeps the existing pin and prints a warning instead. The artifacts do exist at the pinned versions, so the next probe-backed `make catalog-update` will pick all six up automatically now that `DEFAULT_PLATFORMS` is complete: cursor `2026.07.09-a3815c0` `linux/arm64/agent-cli-package.tar.gz`, `windows/x64/agent-cli-package.zip`, `windows/arm64/agent-cli-package.zip` all return **HTTP 200**; opencode `v1.18.3` `opencode-linux-arm64.tar.gz`, `opencode-windows-x64.zip`, `opencode-windows-arm64.zip` all return **HTTP 200**.
- **grok** is npm/npx based and platform independent, so it needed nothing.
- No pin in this PR points at a URL that was not observed returning 200.

## Reviewer note: contradicts the "Rust side is already complete" assumption

`Platform::registry_key()` and `node_binary_name()` are indeed complete, and `linux_arm64` is fully fixed by this PR. Windows is **not** end to end yet: `install_binary_or_archive_from_pin` in `installer/pinned.rs:41` writes the artifact to `managed_dir.join(kind.as_str())`, i.e. an extension-less `claude` / `codex`, and neither the binary nor the archive path appends `.exe` on Windows. Only `installer/npm.rs:485` does. So the Windows pins this PR adds will download and verify correctly but will land at a path Windows will not execute. That is a Rust-side follow-up; I have deliberately left it out because it cannot be compiled or tested here. The catalog side is the prerequisite either way, and the pins are correct.

## Testing

Run locally (no Rust build):

- `node scripts/validate-agent-catalog.mjs` -> `agent catalog OK: 2026-08-20.1 (5 agents)`
- `node --test scripts/agent-catalog/*.test.mjs` -> 28 pass, 0 fail (5 in `resolve-pins.test.mjs`, including the 2 new ones)
- `node scripts/agent-catalog/check-version-discipline.mjs --base origin/main` -> `catalog version discipline OK`
- `node --check scripts/agent-catalog/resolve-pins.mjs` -> clean
- `python3 scripts/check_max_lines.py` -> passed
- `python3 scripts/check_docs.py` -> passed (234 Markdown files)

**Cargo test that gates this and must be checked in CI:** `anyharness-lib`'s `domains::agents::catalog::schema::tests::draft_catalog_parses_with_expected_shape` (source: `anyharness/crates/anyharness-lib/src/domains/agents/catalog/schema_tests.rs:11`). It asserts `catalog.catalogVersion` equals the draft's `catalogVersion` and that `probedAgainst.registryVersion` equals the bundled `registry.json`'s `registryVersion`, so both version bumps in this PR are exactly what it checks. It was **not** run locally, by instruction. `bundled_agent_catalog_document()` (`catalog/bundled.rs:20`) additionally validates the checked-in catalog on first read in every test that touches it, and `domains::agents::catalog::schema::tests::agent_kind_enum_matches_registry_kinds` guards the registry edit.


pablonyx:
## Live validation on aarch64 Linux

Not a review, just evidence. This PR was written from `curl -I` status codes against upstream URLs, which proves the artifacts exist but not that the runtime can use them. So I built this branch on the aarch64 Linux box where the defect was originally observed and ran it.

Box: EC2 `m8g.4xlarge`, Ubuntu 24.04, `uname -m` = `aarch64`. Built `cargo build --release -p anyharness --bin anyharness` from `622f524dd81` in 2m18s, then started the runtime with a clean runtime home and let reconcile auto-install.

Result from `GET /v1/agents`:

```
agent      installState   native                 agentProcess           readiness
claude     installed      managed/2.1.234        managed/65f6205329da   login_required
codex      installed      managed/rust-v0.147.0  managed/219738c9dd42   login_required
cursor     failed         NOT INSTALLED          NOT INSTALLED          install_required
opencode   failed         NOT INSTALLED          NOT INSTALLED          install_required
grok       installed      NOT INSTALLED          managed/0.2.102        login_required
```

The install log line that matters, because it names the source:

```
installed managed agent artifact agent="claude" role="native_cli"
  path=/tmp/rt-catfix/agents/claude/native/claude source=pinned_binary version=Some("2.1.234")
```

`source=pinned_binary` on `linux_arm64` is precisely the path that raised `AGENT_NO_PIN_FOR_PLATFORM` before this change. claude and codex both install and reach `login_required`, meaning the only thing left between them and a working session is credentials. That is the fix working end to end, not just a reachable URL.

## The disclosed gap is real, and it fails legibly

cursor and opencode still fail, exactly as the PR body says they will:

```
reconcile: install failed for agent agent="cursor"
  error=pinned source has no download for this platform: linux_arm64 error_kind="other"
reconcile: install failed for agent agent="opencode"
  error=pinned source has no download for this platform: linux_arm64 error_kind="other"
```

Worth saying explicitly because it was an open question: this is a comprehensible error naming both the cause and the platform, not a crash or a silent no-op, and it does not prevent the other three agents from installing. A user on arm64 Linux who picks cursor gets told why. That makes the gap acceptable to ship, though it should not stay open.

## One thing this run confirms about the stacked follow-up

The installed path is `/tmp/rt-catfix/agents/claude/native/claude`, with no extension. That is the `installer/pinned.rs` gap called out in the PR body, observed rather than inferred. On Linux it is correct and harmless. On Windows the same code path produces the same extension-less name, which Windows will not execute, so the Windows pins added here cannot work until that lands.

## What this does not prove

No turn was run, so nothing here exercises the installed binaries beyond installation and the readiness probe. Credentials on this box are Bedrock through the instance IAM role and were not configured for the arm64 run. Nothing Windows was tested at all: this is aarch64 Linux only, which happens to be the platform the original bug report came from.

pablonyx:
## Review: independent re-verification of every pin

**Verdict: yes, safe to merge on supply-chain grounds. Every one of the six new pins is correct and I re-derived all six checksums from provider truth without trusting the author's table. One conditional caveat on release sequencing (C1) is worth a decision before this reaches a Windows-facing build.**

No new download host is introduced: the claude bucket (`claude-code-dist-86c565f3-…`) and `github.com/openai/codex` are both already the trust roots on `main`. The committed catalog reproduces byte-for-byte from `main`'s baseline with the documented command, so nothing was hand-edited into the lockfile.

---

## Conditional blocker

### C1 — On Windows the install now reports success and hands back a file Windows cannot execute

`anyharness/crates/anyharness-lib/src/domains/agents/installer/pinned.rs:40`
```rust
let target_path = managed_dir.join(kind.as_str());
```

The author discloses this. The disclosure is incomplete in the part that matters. It says the pins "will land at a path Windows will not execute". It does not say that the install **succeeds**:

- `integrations/agent_cli/executable.rs:62-65` — `make_executable` is a `#[cfg(not(unix))]` no-op, so nothing errors.
- `integrations/agent_cli/executable.rs:48-51` — on non-unix `is_valid_executable` is `path.is_file()`, so an extension-less `claude` passes the validity gate.
- `pinned.rs:52-63` then returns `Ok(InstalledArtifactResult { … source: "pinned_binary" })`.

So today a Windows user gets HTTP 400 `AGENT_NO_PIN_FOR_PLATFORM` at install: precise, immediate, actionable. After this PR they get a 324 MB download, a checksum verification, a reported-successful install, and then an opaque spawn failure at first turn. That is a worse failure mode, not a better one, and it is the classic "installs a broken agent" case.

This is not a pin-correctness problem and the fix is Rust-side, so it does not block the catalog change on its own merits. My recommendation: land this behind, or in the same release train as, the stacked `installer/pinned.rs` fix, and do not ship a Windows-facing build with this catalog alone. `linux_arm64` is fully fixed by this PR and can ship immediately.

Method: read the code. I did not execute anything on Windows, so the user-visible symptom is reasoned, not observed.

---

## Non-blocking

### N1 — The PR fixes this instance of the bug and adds no guard against the next one

Nothing anywhere asserts that a native pin covers all six platforms:

- `scripts/validate-agent-catalog.mjs` never looks at `source.targets` keys at all (I read all 187 lines; it checks versions, snapshots, auth contexts, presentation models).
- `catalog/validation.rs:111,118` only require `targets` to be non-empty and each url/sha to be non-empty.
- `resolve-pins.mjs:148` and `:171` rebuild `const targets = {}` from scratch, so a narrowed run **deletes** the other platforms rather than merging into them.

And the new comment at `resolve-pins.mjs:33-37` actively invites narrowing ("Narrow it with `--platforms` only for a deliberately partial, inspection-only run"), while nothing enforces inspection-only. The new test at `resolve-pins.test.mjs:275` runs exactly that shape.

Concrete scenario: someone debugs with `resolve-pins.mjs --platforms linux_x64`, commits the result, and every gate in the repo passes while macOS users get the exact `AGENT_NO_PIN_FOR_PLATFORM` this PR exists to fix. A six-key assertion in `validate-agent-catalog.mjs` is a few lines and would make the fix permanent.

### N2 — A `direct_binary` pin can be written for a 404 URL with a plausible checksum

`resolve-pins.mjs:376,384` keys the checksum off `vendorPlatform` read from the manifest, never off the URL. `resolve-pins.mjs:154` makes `binaryNameMap` optional (`?? kind`), and neither `registry.schema.json` nor any test requires its keys to match `platformMap`'s.

Scenario: a maintainer adds a platform to claude's `platformMap` and forgets `binaryNameMap`. The URL becomes `.../win32-x64/claude`, which I verified returns **404**, but `manifest.platforms["win32-x64"].checksum` still returns a valid 64-hex digest, so a pin pointing at a 404 is written, and passes the validator, the Rust validation, and CI. It fails on a user's machine.

The behaviour is pre-existing, but this PR is what makes the filename a separately-specified, forgettable field. A `HEAD` on the resolved URL, or requiring `binaryNameMap` keys to equal `platformMap` keys, closes it.

### N3 — Two different mechanisms for the same problem in one commit, and only the less surprising one is documented

claude gets an explicit `binaryNameMap` (`registry.json:24`) plus a `{binary}` placeholder plus a `model.rs` doc comment. codex gets `.exe` smuggled into `platformMap` **values** (`registry.json:178-179`).

I checked the codex trick for collateral damage and found none. `{target}` feeds exactly three derived values and all three come out right:

| derived value | result | status |
| --- | --- | --- |
| `versionedUrlTemplate` | `codex-x86_64-pc-windows-msvc.exe.tar.gz` | 200, correct |
| `latestUrlTemplate` | `.../releases/latest/download/codex-x86_64-pc-windows-msvc.exe.tar.gz` | **200** (I checked this one too) |
| `expectedBinaryTemplate` | `codex-x86_64-pc-windows-msvc.exe` | matches the tar entry (verified below) |

Rust's `platform_map` is dead data: `registry/projection.rs:128` stores it and the only consumer is a variant match at `readiness/artifacts.rs:20`. Nothing constructs a URL from the registry in Rust.

So the mechanism is safe. The problem is that `model.rs:102` still documents `TarballRelease.platform_map` as "a target triple or segment string" with no mention that the Windows entries must carry `.exe` for the asset name and the inner binary name to line up. The PR documented the obvious half and left the surprising half undocumented, which is backwards.

### N4 — `registry.schema.json` is not enforced anywhere, so the `binaryNameMap` schema addition is not coverage

Grepping the whole repo, the only reference to `registry.schema.json` outside the file itself is a docs table at `specs/codebase/platforms/product/agent-distribution.md:535`. No CI step, no Makefile target, no test validates `registry.json` against it. Adding the property is right, but do not count it as a gate.

Relatedly, Rust's registry deserializer has no `deny_unknown_fields` anywhere in `registry/schema.rs`, so the new `binaryNameMap` is silently ignored by the runtime rather than breaking it. That is the outcome we want, but it is luck rather than design.

### N5 — `--keep-versions` warns for `registry_backed` only

`resolve-pins.mjs:222-229` prints a clear warning when it declines an ACP registry upgrade. The `direct_binary` (`:145`) and `tarball_release` (`:167`) branches pin back **silently**. A maintainer running the flag out of habit is never told that claude `2.1.238` and codex `rust-v0.149.0` exist upstream. Asymmetric, and the silent half is the half covering the two biggest agents.

### N6 — The drift-detection story is weaker for cursor and grok than the PR body implies

I ran the negative control rather than taking the claim on trust: plain `resolve-pins.mjs --no-download` on the PR tree, then the validator. It failed with exit 1 on eleven errors covering claude, codex and opencode. It did **not** flag cursor `2026.07.09 → 2026.08.11` or grok `0.2.102 → 1.0.8`, because `validate-agent-catalog.mjs:130` exempts `archive`/`npm` sources with no attested version from the version check.

So the "would break `validate-agent-catalog.mjs`'s snapshot-evidence rule" argument in the PR body is true for claude, codex and opencode, and false for exactly the two agents where nothing else would catch the drift either. `--keep-versions` is more load-bearing than the body claims, not less. Worth stating accurately so nobody later decides the flag is redundant.

### N7 — "the next probe-backed `make catalog-update` will pick all six up automatically" is optimistic

The plumbing does work: `run-probes.sh:257` calls resolve-pins with the default platform set, and I confirmed the ACP registry publishes `linux-aarch64`, `windows-x86_64` and `windows-aarch64` keys for both cursor and opencode. But look at what would actually get pinned:

- cursor windows `cmd` is `./dist-package\cursor-agent.cmd` — a backslash in the path and a `.cmd` batch file, not an executable.
- opencode windows-x86_64 `cmd` is `./opencode.exe` but windows-aarch64 `cmd` is `./opencode` — an upstream inconsistency.
- Both cursor Windows artifacts are `.zip`, and `downloads.rs:224,226` shells out to `unzip`, which Windows does not ship.

Those `cmd` values land verbatim in `expectedBinary` and are looked up at `downloads.rs:256` as `staging_dir.join(expected_binary)`. None of that is this PR's job, but the body should not promise "automatically".

### N8 — The next probe run doubles cursor's download volume

cursor's ACP registry entries publish no `sha256` (I checked all six targets: all null), so `shaForPublished` falls through to `shaFor(url)` and downloads the artifact to hash it. Going from three to six platforms means six cursor package downloads instead of three on the next `make catalog-update`. opencode publishes `sha256` for all six and is unaffected.

### N9 — `generatedAt` is stale

`catalog.json:6` still reads `2026-08-17T20:49:16.584Z` after a regeneration on 2026-08-20. `resolve-pins.mjs` never writes the field. Pre-existing, cosmetic.

---

## Pins I independently re-verified

Every URL below I resolved myself with `curl -sIL -o /dev/null -w '%{http_code}'`. Every checksum I compared against the provider's own published digest, fetched independently. Sizes are the observed `content-length` (claude) or the GitHub API asset `size` (codex).

### claude native, `2.1.234`, provider manifest `.../claude-code-releases/2.1.234/manifest.json` (fetched, HTTP 200)

| platform | URL suffix | HTTP | catalog sha256 | manifest checksum | size match |
| --- | --- | --- | --- | --- | --- |
| linux_arm64 | `2.1.234/linux-arm64/claude` | **200** | `24adda67…44aaa2` | **identical** | 325507304 = 325507304 |
| windows_x64 | `2.1.234/win32-x64/claude.exe` | **200** | `3f877e78…02eb1b` | **identical** | 324028576 = 324028576 |
| windows_arm64 | `2.1.234/win32-arm64/claude.exe` | **200** | `ed6cc3d0…65e8d4` | **identical** | 311927456 = 311927456 |
| windows_x64 (no `.exe`) | `2.1.234/win32-x64/claude` | **404** | — | — | 404 confirms the rename was necessary |
| windows_arm64 (no `.exe`) | `2.1.234/win32-arm64/claude` | **404** | — | — | same |

The manifest also declares `"binary": "claude.exe"` for both win32 targets and `"binary": "claude"` for every POSIX target, matching `binaryNameMap` exactly.

**End-to-end fidelity spot-check.** A provider manifest matching a catalog only proves the author copied the manifest correctly, not that the manifest describes the bytes. So I downloaded the full `win32-arm64/claude.exe` and hashed it:

```
311927456 bytes
sha256 ed6cc3d0218e4b27f3691e3832837d814a5274b02755c34aedeb74e3db65e8d4
```

Identical to both the manifest and the committed pin. The manifest is faithful for at least one Windows artifact, which is what makes the manifest-derived checksums for the other two credible rather than merely self-consistent.

### codex native, `rust-v0.147.0`, GitHub release API `repos/openai/codex/releases/tags/rust-v0.147.0` (fetched, HTTP 200, `prerelease: false`)

| platform | asset | catalog sha256 | GH `digest` | size | inner binary (streamed `tar -tzf`) |
| --- | --- | --- | --- | --- | --- |
| linux_arm64 | `codex-aarch64-unknown-linux-musl.tar.gz` | `eb677c80…38b90` | **identical** | 91607658 = 91607658 | `codex-aarch64-unknown-linux-musl` ✓ |
| windows_x64 | `codex-x86_64-pc-windows-msvc.exe.tar.gz` | `7c47d4dc…a8bdcd` | **identical** | 99655443 = 99655443 | `codex-x86_64-pc-windows-msvc.exe` ✓ |
| windows_arm64 | `codex-aarch64-pc-windows-msvc.exe.tar.gz` | `c3a7da6c…07ed07` | **identical** | 91259737 = 91259737 | `codex-aarch64-pc-windows-msvc.exe` ✓ |

`codex-x86_64-pc-windows-msvc.tar.gz` and `codex-aarch64-pc-windows-msvc.tar.gz` (the non-`.exe` forms) do **not appear in the release asset list at all**, confirming the 404 claim from the API side rather than just from a HEAD.

I streamed the first 4 MB of each of the three archives through `tar -tzf` myself; the `expectedBinary` values are correct and the entries sit at archive root, which is what `downloads.rs:256` expects.

### Unchanged lanes, spot-checked as controls

| platform | HTTP | note |
| --- | --- | --- |
| claude `linux-x64/claude` | 200 | content-length 328358192 = committed `downloadSizeBytes` |
| claude `darwin-arm64/claude` | 200 | content-length 310740672 = manifest size |
| codex `codex-x86_64-unknown-linux-musl.tar.gz` | — | GH digest `0246e2e7…` = committed pin |

### cursor and opencode "artifacts exist at the pinned version" claims

| artifact | HTTP |
| --- | --- |
| cursor `2026.07.09-a3815c0/linux/arm64/agent-cli-package.tar.gz` | **200** |
| cursor `2026.07.09-a3815c0/windows/x64/agent-cli-package.zip` | **200** |
| cursor `2026.07.09-a3815c0/windows/arm64/agent-cli-package.zip` | **200** |
| opencode `v1.18.3/opencode-linux-arm64.tar.gz` | **200** | 
| opencode `v1.18.3/opencode-windows-x64.zip` | **200** |
| opencode `v1.18.3/opencode-windows-arm64.zip` | **200** |

All six exist. The reasoning for leaving them out holds: I fetched the live ACP registry and it serves cursor `2026.08.11` and opencode `1.18.19` against pins of `2026.07.09` and `1.18.3`, so taking its URLs would mix versions across platforms within one agent. That is correct and I agree with the call. It does mean a Windows user selecting cursor or opencode still gets `AGENT_NO_PIN_FOR_PLATFORM` — which, per C1, is currently the *better* of the two Windows outcomes. The PR body documents this clearly under "What remains unsupported".

### Upstream version claims

| claim | observed |
| --- | --- |
| claude latest is `2.1.238` | `2.1.238` ✓ |
| codex latest is `rust-v0.149.0` | `rust-v0.149.0` ✓ |
| ACP serves cursor `2026.08.11` | ✓ |
| ACP serves opencode `1.18.19` | ✓ |
| ACP serves grok `1.0.8` | `grok-build 1.0.8` ✓ |

---

## Reproducibility and mixed-version risk

**Is the committed catalog what the documented command produces?** Yes, byte-for-byte, and I checked it the hard way. Starting from `origin/main`'s `catalog.json` plus this PR's `registry.json` and running `node scripts/agent-catalog/resolve-pins.mjs --keep-versions`, the six new targets were resolved fresh (the resolver printed `✓ win32-x64 checksum from …/manifest.json` and `✓ checksum published for …codex-*.exe.tar.gz`, i.e. no sha was carried over from a prior catalog) and the output differs from the committed file **only** in the two hand-bumped version fields. Re-running it against the PR's own catalog produces an identical file. Every digit in every new pin is therefore accounted for.

**Can one agent end up with platforms at different versions?** Not through any normal path. `resolveNative` resolves a single `version` and then fills all six targets from it, and `--keep-versions` for `registry_backed` keeps the whole `currentSource` object rather than merging platform-by-platform, so partial adoption is not reachable. The one way to get an inconsistent lockfile is the `--platforms` truncation in N1, and that removes platforms rather than mixing versions. Nothing detects either.

**Does `--keep-versions` do only what it claims?** Effectively yes. It touches three branches and nothing else. `managed_npm_package` (claude/codex adapters, git-pinned) and `direct_archive` already used `currentVersion` and are untouched. The non-git `managed_npm_package` branch ignores the flag, but no agent uses that path today, so it is dead rather than wrong.

**Is the divergence from a plain run detectable?** Mostly, and it fails closed. See N6 for the cursor/grok hole.

---

## Test and validator reruns

I reran all five, from a clean extraction of the PR head rather than the author's shell, and I confirm the counts.

| check | author claimed | I observed |
| --- | --- | --- |
| `node scripts/validate-agent-catalog.mjs` | `agent catalog OK: 2026-08-20.1 (5 agents)` | identical, exit 0 |
| `node --test scripts/agent-catalog/*.test.mjs` | 28 pass, 0 fail, 5 in resolve-pins (2 new) | **28 pass, 0 fail**; `resolve-pins.test.mjs` contributes 5, the two new ones being "windows direct binaries resolve their per-platform artifact name" and "`--keep-versions` refuses to adopt a newer ACP registry release" |
| `check-version-discipline.mjs --base origin/main` | OK | `catalog version discipline OK against origin/main`, exit 0 |
| `python3 scripts/check_max_lines.py` | passed | `Max-lines check passed`, exit 0 (`model.rs` 570 → 573, well under the 600 cap; `scripts/*.mjs` is outside `CHECK_ROOTS` so the resolver's 376 → 431 is not gated) |
| `python3 scripts/check_docs.py` | passed, 234 files | `Documentation integrity check passed (234 Markdown files)` |

**Negative control on the validator**, which the author did not run: I ran a plain `resolve-pins.mjs --no-download` on the PR tree and re-validated. Exit 1, eleven errors, e.g. `claude: probe snapshot 'generated/claude.anthropic-api.probe.json' native version does not match its pin` and `opencode: … attests '1.18.3', expected '1.18.19'`. The snapshot-evidence rule really does fail closed on a silent version bump, which is the load-bearing justification for `--keep-versions`. (With the cursor/grok gap noted in N6.)

**The cargo test the author named.** `draft_catalog_parses_with_expected_shape` exists at `anyharness/crates/anyharness-lib/src/domains/agents/catalog/schema_tests.rs:11`. I read it and did not run it, per instruction. It passes:

- `assert_eq!(catalog.catalog_version, draft_catalog_version())` is **tautological** — `draft_catalog_version()` at `schema_tests.rs:146` reads `catalogVersion` out of `scripts/agent-catalog/catalog.draft.json`, the same file `draft_catalog_json()` at `schema.rs:258` `include_str!`s. It can never fail. The author cited it as gating the catalog version bump; it does not.
- `probed_against.registry_version == bundled_registry_version()` is the real assertion, and it is cross-file: draft `probedAgainst.registryVersion` (`2026-08-20.1`) against `catalogs/agents/registry.json`'s `registryVersion` (`2026-08-20.1`). Both bumped consistently. Passes.
- `catalog.agents.len() == 5`, claude native `2.1.234`, claude agentProcess `0.66.0-proliferate.2` — all unchanged by this diff. Passes.

**Other cargo gates the author did not name**, which I checked for and which also pass by inspection:
- `catalog/bundled.rs:20` `bundled_agent_catalog_document()` `include_str!`s `catalogs/agents/catalog.json` and runs `validate_agent_catalog_document` + `validate_agent_catalog_registry_pairing` on first read in every test that touches it. The new targets only need non-empty url and sha (`validation.rs:111,118`); `targets` is a `BTreeMap<String, …>` (`schema.rs:93`) so the new platform keys need no enum membership.
- `registry/projection.rs:265` `parse_platform` accepts all six keys, and the codex `.exe` change is to map **values**, not keys, so projection is unaffected.
- `registry/schema.rs` has **no** `deny_unknown_fields` anywhere, so the new `binaryNameMap` property does not break registry deserialization. I checked this explicitly because it would have been a hard runtime break.
- I found no hardcoded `2026-08-17.1` / `2026-08-19.1` anywhere in the repo outside the four JSON version fields the PR updates.

CI on the head commit is **33 pass, 2 skipping, 0 fail**, including "Cargo check & test" and "CI/CD config checks" (which runs exactly the three node checks above, at `.github/workflows/ci.yml:198-203`).

---

## What I could not verify

- **Five of the six new artifacts were not byte-hashed.** I hashed `win32-arm64/claude.exe` in full (311 MB) as a fidelity check on the manifest. For the other two claude targets I have the provider manifest checksum plus a `content-length` that matches the recorded `downloadSizeBytes` exactly; for the three codex targets I have GitHub's own `digest` field plus matching asset sizes plus a streamed `tar -tzf` of each. That is provider attestation, not an independent hash. If you want zero residual trust in the GCS manifest and the GitHub digest field, the remaining five need ~940 MB of downloads.
- **Windows runtime behaviour.** C1 is derived from reading `pinned.rs`, `executable.rs` and `downloads.rs`. No Windows machine was involved, so the exact user-visible error is reasoned, not observed.
- **No cargo test was executed**, per instruction. My conclusions about `draft_catalog_parses_with_expected_shape` and the bundled-catalog validation are from reading the source. CI's green "Cargo check & test" is the actual evidence.
- **The claude release channel's authenticity chain.** I verified the manifest and the artifacts agree, and that the host is unchanged from `main`. I did not verify any signature, because the channel publishes none — there is no sigstore bundle or detached signature for the claude direct-binary lane the way there is for codex's musl assets. That is an accepted property of the existing design, not something this PR changes.
- **Whether the stacked Rust PR for C1 actually exists or is correct.** I took the author's word that one is coming; I did not look for it.

pablonyx:
## Re-review after rebase onto main (622f524dd8 -> current)

Scope: verify the rebase, and re-check the two items flagged in the last round (C1 Windows-dependency disclosure, and the new cursor/opencode Windows-gap finding from real Windows CI). Not re-doing the checksum audit or the aarch64 live validation; those stand.

### 1. Rebase integrity — clean

```
git range-diff 68078e1d7d..622f524dd8 956d2426b7..70fbe91044
1:  622f524dd8 = 1:  70fbe91044 fix(agents): resolve pins for windows and arm64 linux
```

Single commit, patch content byte-identical (`=`) before and after the rebase. Nothing dropped, nothing reworded, nothing corrupted in transit. The only thing between the old and new base is mainline history (`956d2426b7`, `ccb53edc85`, etc.), which is expected.

### 2. arm64 Linux side — genuinely correct

Re-checked the committed `catalogs/agents/catalog.json` on `70fbe91044` directly (not just the PR-body table): claude's and codex's `linux_arm64` targets carry the exact URL/sha256/size the body claims (`24adda67…` / 325507304 for claude, `eb677c80…` / 91607658 for codex), and all six platform keys are present for both agents with the Windows filenames matching (`claude.exe` via the new `binaryNameMap`, `codex-*-pc-windows-msvc.exe.tar.gz` via the corrected `platformMap`). This is consistent with the previously-completed independent checksum re-derivation and the aarch64 live-validation comment already on this PR (`claude installed managed/2.1.234`, `codex installed managed/rust-v0.147.0`). No discrepancy found.

### 3. Windows dependency disclosure — accurate, not overclaiming

The "Reviewer note" section states plainly that Windows is **not** end-to-end and names the exact mechanism (`installer/pinned.rs` writes an extension-less path; only `npm.rs` appends `.exe`). It doesn't cite `#2149` by number, but `#2149`'s own body says "Stacked on #2145 ... not on main," and its base ref literally is this PR's branch, so the dependency is structurally explicit to anyone looking at the stack, not just asserted in prose. Combined with this PR staying in draft, there's no path for a reader to conclude "Windows works standalone" from this PR alone.

Non-blocking suggestion: now that `#2149` (and `#2152`) exist, it'd be a kindness to add one line linking them from this body, so someone who opens only `#2145` doesn't have to go find the stack. Not required for merge safety.

### 4. New finding (cursor/opencode have no windows_* target at all) — already disclosed, verified against the actual catalog

Checked `catalog.json`'s resolved `agentProcess.source.targets` directly: both cursor and opencode list only `macos_arm64, macos_x64, linux_x64` — no `linux_arm64`, no `windows_x64`, no `windows_arm64` keys exist anywhere in the committed catalog. That matches exactly, and does not go beyond, what the body's "What remains unsupported" section already says. The Windows-CI-derived finding isn't new information relative to this PR's own disclosure — it's an independent confirmation that the disclosure was accurate. No overclaim anywhere in the body about Windows/cross-platform completeness; the title and every table in the body scope this correctly as a pins fix, not a "Windows support" claim.

### 5. Blocking item — CI is currently red, unrelated to this diff

`gh pr checks 2145` shows `Shared frontend packages (2)` failing, which fails the `ci-ok` rollup gate ("CI did not verify this commit. Not green: shared-frontend-packages (failure)"). The failure is a `waitFor` timeout in `apps/packages/product-client/src/components/workspace/chat/input/ComposerCommandEditor.test.tsx:198` (slash-command composer draft-change assertion) — nothing this PR touches (diff is limited to `catalogs/agents/*`, `scripts/agent-catalog/*`, a doc comment in `anyharness-lib/.../agents/model.rs`, and a specs doc). Confirmed the same shard was fully green on this PR's immediate parent commit on main (`956d2426b7`), so this is a flake surfaced on this run, not a regression introduced by the rebase or the change itself.

This blocks merge under the repo's own gate as it stands, but needs a rerun, not a code change.

## Verdict

No blocking correctness issues in the change itself. Rebase is clean, the arm64 Linux fix is verified correct against the committed catalog, and both Windows-related disclosures (the pinned.rs extension gap and the cursor/opencode platform gap) are accurate and match what's actually in the resolved catalog — no overclaiming anywhere. Confirms the prior round's conclusion: this must land together with `#2149` (and, per that PR, effectively with `#2152` too) before any Windows-facing build ships; it should not merge standalone as a "Windows fixed" claim, and it doesn't claim to be one.

Only actionable item before merge: get a green CI run (the current red is an unrelated frontend flake, not caused by this diff). Non-blocking: consider linking `#2149`/`#2152` from this body now that they exist.

PR #2146, opened 2026-08-21
fix(sdk): make anyharness openapi generation windows-safe

pablonyx:
## What this is

Desktop Windows releases were disabled on 2026-04-13 by `3d3c0504b8`, whose own doc note gave the reason: "Windows packaging and updater entries are temporarily disabled until the SDK generation step is Windows-safe."

That step is one line. `anyharness/sdk/package.json`:

```
"generate:openapi": "mkdir -p generated src/generated && cd ../.. && cargo run --bin anyharness -- print-openapi > anyharness/sdk/generated/openapi.json"
```

npm runs package scripts through `cmd.exe` on Windows, and that command does not work there. I originally wrote that `-p` becomes a directory named `-p`. Review caught that, and the Windows runner settled it. cmd creates nothing at all:

```
cmd.exe says:
The syntax of the command is incorrect.
what exists afterwards:
total 4
drwxr-xr-x 1 runneradmin 197121 0 Aug 21 02:23 .
drwxr-xr-x 1 runneradmin 197121 0 Aug 21 02:23 ..
confirmed: cmd.exe did not create the directories the POSIX form asks for
```

No `-p`, no `generated`, no `src/generated`. The `&&` chain therefore never reaches cargo, and the step fails, taking the Windows desktop build with it.

I swept every `package.json` in the workspace for shell-isms (`mkdir`, `cp`, `chmod`, `rm -rf`, command substitution). This is the only hit. The stated blocker for Windows desktop releases is a single `mkdir -p`.

## The change

The work moves into `scripts/generate-anyharness-openapi.mjs`: create both directories with `fs.mkdirSync(..., {recursive: true})`, spawn the runtime, validate, write. No shell on any platform. `package.json` and the `sdk-generate` Makefile target both call it, so the two paths can no longer drift.

Behaviour on macOS and Linux is unchanged, with one deliberate exception described below.

## A second bug this removes

The shell redirect `> anyharness/sdk/generated/openapi.json` truncates the target **before** `cargo run` executes. Any runtime failure therefore left a zero byte schema on disk. Demonstrated locally against the old line with a stub runtime that exits 101:

```
before: {"real":"schema"}
compile error
exit=101
after:  []
size=0
```

The node version buffers, checks the exit status, rejects empty output, `JSON.parse`s the payload, and only then writes. Under the same stub the committed schema is untouched. This is the one intentional behaviour change on unix, and it is strictly in the safe direction.

## Verification

Nothing here was assumed. Local runs against stub runtimes, invoked from `anyharness/sdk` exactly as npm invokes it:

| Case | Expected | Result |
|---|---|---|
| Runtime prints a schema | writes it, exit 0 | wrote `openapi.json`, exit 0, both directories created |
| Runtime prints nothing | refuse to write | `refusing to write an empty schema`, exit 1 |
| Runtime prints non-JSON | refuse to write | `did not produce valid JSON`, exit 1 |
| Runtime exits 101 | propagate, do not clobber | exit 101, existing schema byte-identical |
| Runtime binary missing | fail clearly | `failed to run ...: ENOENT`, exit 1 |

I did not run `cargo` anywhere. The real end-to-end proof is already in CI and costs this PR nothing: `.github/workflows/ci.yml:300` runs `pnpm run generate` with a real toolchain and `:304` asserts `git diff --exit-code` on both generated files. If this script produced different bytes than the shell line, that gate goes red.

## The regression guard

`.github/workflows/windows-sdk-generate-check.yml` builds nothing and runs in seconds.

Both jobs invoke the real `npm run generate:openapi`, not the node script directly. That matters: the regression being guarded lives in `package.json`, so a guard that bypasses it would stay green after a revert. The first version of this guard did exactly that, and review caught it.

The **windows** job first runs the old POSIX form under `cmd.exe` and fails if it somehow works, so the guard cannot quietly stop being meaningful. It then deletes both generated directories, runs the package script with a deliberately absent runtime, and asserts both directories came back. The discriminator is the directories, not the exit code: the runtime is absent so the script must fail either way, and what separates a working package script from a broken one is whether the directories exist afterwards. Live result:

```
ok: npm run generate:openapi created both directories on windows
```

The **linux** job covers the invoke, validate and write path with shell stubs, including the truncation case above.

## What review found, and what it cost

The production fix survived review unchanged, including the byte-equivalence concern that mattered most: the reviewer ran a real 540,317 byte payload and a 38MB payload through the script and confirmed both are `cmp`-identical to the old `>` redirect, and confirmed `maxBuffer` overflow is not silent.

The guard did not survive. Four blocking findings, all now fixed and all now green:

1. The linux stub was written with `printf`, so shell quote removal stripped the JSON's double quotes and the stub emitted invalid JSON, failing the job it existed to prove. It is a quoted heredoc now, and the step self-tests the stub before using it.
2. The windows negative control ran under `cmd`, where a failing `mkdir` leaves an errorlevel that `echo` does not reset, so the step failed and the job never reached the part that tests the fix. It is driven from bash now, so the assertion is mine rather than cmd's errorlevel semantics.
3. Both jobs ran the node script directly rather than through npm, so the guard did not guard the regression.
4. The stated mechanism was wrong, corrected above.

Also fixed from the non-blocking list: `??` to `||` for `CARGO` so an empty value says something useful instead of throwing `ERR_INVALID_ARG_VALUE`, `Makefile` added to the path triggers, actions SHA-pinned to match the repo standard, a `concurrency` group added, and the before/after comparisons switched from `$(cat)` to `sha256sum`.

The split is deliberate. The windows job cannot exercise the cargo invocation itself: Node refuses to spawn `.cmd` stubs without a shell since CVE-2024-27980, and a real `cargo run` there would mean a full Windows runtime build. What was broken on Windows was directory handling, and that is what the Windows job tests.

## What this does not do

This does not re-enable Windows desktop releases. `3d3c0504b8` also removed the `windows-latest` matrix entry from `release-desktop.yml`, the `windows` dispatch option, the Windows artifact upload, and the `windows-x86_64` key from `scripts/generate-updater-manifest.mjs`. Restoring those needs the desktop crate to actually compile for Windows first, which is being established in #2142, and it needs a decision about `generate-updater-manifest.mjs` being fail-closed: it errors on a missing platform artifact today, so adding a `windows-x86_64` key without a guaranteed Windows artifact would break macOS releases.

Worth noting for whoever picks that up: the runtime crates already build and ship `anyharness-x86_64-pc-windows-msvc.zip` on every release, and `specs/desktop-native.md` still carries Windows-specific updater logic. Less of this is missing than the disabled matrix suggests.

## Rebase note (2026-08-21)

Rebased onto current main, which had picked up #2161 (local postgres/redis auto-detection) and #2134 (`make dev` dev-loop launch rework), both of which touched the Makefile's `sdk-generate` target.

The conflict was in the body of `sdk-generate`. Main's version had grown a SKIP_RUST branch: when `SKIP_RUST` is set and `ANYHARNESS_DEV_RUNTIME_BIN` points at an executable prebuilt runtime, it calls that binary directly instead of `cargo run`, so a frontend-only worktree does not need a Rust toolchain to regenerate the SDK. This branch did not exist at the time this PR's commits were written, so it could not just be replayed on top.

Resolution: kept this PR's approach of delegating the whole recipe to `scripts/generate-anyharness-openapi.mjs`, and ported main's SKIP_RUST/prebuilt-runtime branch into the script instead of the Makefile recipe, since that is exactly the shell logic this PR exists to get rid of. The Makefile line is now `CARGO="$(CARGO)" SKIP_RUST="$(SKIP_RUST)" node scripts/generate-anyharness-openapi.mjs`, and the script picks cargo vs the prebuilt runtime the same way the old shell branch did, using `fs.accessSync(..., X_OK)` in place of the shell's `[ -x ... ]` check.

Verified with `make -n` dry runs: `build-rust` and `runtime-build` are byte-identical to main (untouched by this PR), and the only diff anywhere is the `sdk-generate` recipe body itself. Also verified the script's branching directly against stub cargo and stub runtime binaries: SKIP_RUST unset uses cargo, SKIP_RUST=1 with a valid executable runtime bin uses the runtime, SKIP_RUST=1 with a missing/non-executable runtime bin falls back to cargo, and SKIP_RUST=0 falls back to cargo. All four match the pre-rebase main behavior.

No functional changes beyond the rebase itself.

### DIFF
--- a/.github/workflows/windows-sdk-generate-check.yml
+++ b/.github/workflows/windows-sdk-generate-check.yml
@@ -1,0 +1,132 @@
+name: Windows SDK generate check
+
+# `anyharness/sdk`'s `generate:openapi` script used to be a POSIX shell one
+# liner (`mkdir -p ... && cd ../.. && cargo run ... > file`). npm runs package
+# scripts through cmd.exe on Windows, where `mkdir -p` creates a directory
+# literally named `-p`, so the step failed. That single line is the stated
+# reason desktop Windows releases were disabled in 3d3c0504b8 ("until the SDK
+# generation step is Windows-safe").
+#
+# This is the regression guard, and it builds nothing.
+#
+# The windows job covers what was actually broken: directory creation and path
+# handling under cmd.exe. It cannot exercise the cargo invocation itself,
+# because Node refuses to spawn `.cmd` stubs without a shell (CVE-2024-27980)
+# and a real `cargo run` here would mean a full Windows runtime build. The
+# linux job covers the invoke, validate and write path with a shell stub. The
+# two together cover the whole script.
+
+on:
+  pull_request:
+    paths:
+      - "scripts/generate-anyharness-openapi.mjs"
+      - "anyharness/sdk/package.json"
+      - ".github/workflows/windows-sdk-generate-check.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  windows:
+    name: directory handling on windows
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      - name: Negative control, the old shell form is broken under cmd.exe
+        shell: cmd
+        run: |
+          mkdir oldform
+          cd oldform
+          mkdir -p generated src/generated
+          if exist "generated\openapi.json" (
+            echo UNEXPECTED: cmd.exe produced generated/openapi.json, this guard is no longer meaningful
+            exit /b 1
+          )
+          if not exist "generated" (
+            echo confirmed: the old form did not create a usable generated directory
+          )
+          if exist "-p" (
+            echo confirmed: cmd.exe created a directory literally named -p
+          )
+
+      # Both directories are checked in, so they have to go before the run or
+      # the assertion below would pass without the script doing anything.
+      - name: Remove the generated directories
+        shell: bash
+        run: rm -rf anyharness/sdk/generated anyharness/sdk/src/generated
+
+      - name: The node form recreates both directories on windows
+        shell: cmd
+        env:
+          CARGO: this-runtime-does-not-exist
+        run: |
+          cd anyharness\sdk
+          node ..\..\scripts\generate-anyharness-openapi.mjs
+          if "%ERRORLEVEL%" == "0" (
+            echo UNEXPECTED: a missing runtime should have failed the script
+            exit /b 1
+          )
+          echo confirmed: the script failed loudly on a missing runtime, as intended
+
+      - name: Assert both directories were recreated and nothing partial landed
+        shell: bash
+        run: |
+          test -d anyharness/sdk/generated || { echo "generated/ was not created"; exit 1; }
+          test -d anyharness/sdk/src/generated || { echo "src/generated/ was not created"; exit 1; }
+          test ! -e anyharness/sdk/generated/openapi.json || { echo "a schema was written despite the runtime failing"; exit 1; }
+          test ! -e "anyharness/sdk/-p" || { echo "the POSIX flag leaked into a directory name"; exit 1; }
+          echo "ok: both directories recreated on windows, no partial schema written"
+
+  linux:
+    name: invoke and write path with a stubbed runtime
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      - name: Stub the runtime
+        run: |
+          mkdir -p "$RUNNER_TEMP/stub"
+          printf '#!/bin/sh\necho %s\n' '{"openapi":"3.0.0","info":{"title":"stub","version":"1"},"paths":{}}' > "$RUNNER_TEMP/stub/cargo"
+          chmod +x "$RUNNER_TEMP/stub/cargo"
+          printf '#!/bin/sh\nexit 101\n' > "$RUNNER_TEMP/stub/cargo-fail"
+          chmod +x "$RUNNER_TEMP/stub/cargo-fail"
+          printf '#!/bin/sh\necho not json\n' > "$RUNNER_TEMP/stub/cargo-garbage"
+          chmod +x "$RUNNER_TEMP/stub/cargo-garbage"
+
+      - name: Happy path writes a parseable schema
+        working-directory: anyharness/sdk
+        run: |
+          CARGO="$RUNNER_TEMP/stub/cargo" node ../../scripts/generate-anyharness-openapi.mjs
+          node -e "JSON.parse(require('fs').readFileSync('generated/openapi.json','utf8'))"
+          echo "ok: schema written and parses"
+
+      - name: A failing runtime must not truncate an existing schema
+        working-directory: anyharness/sdk
+        run: |
+          before="$(cat generated/openapi.json)"
+          if CARGO="$RUNNER_TEMP/stub/cargo-fail" node ../../scripts/generate-anyharness-openapi.mjs; then
+            echo "a failing runtime should have failed the script"
+            exit 1
+          fi
+          after="$(cat generated/openapi.json)"
+          [ "$before" = "$after" ] || { echo "the existing schema was clobbered by a failed run"; exit 1; }
+          echo "ok: existing schema survived a failed runtime"
+
+      - name: Invalid output must be rejected before it is written
+        working-directory: anyharness/sdk
+        run: |
+          before="$(cat generated/openapi.json)"
+          if CARGO="$RUNNER_TEMP/stub/cargo-garbage" node ../../scripts/generate-anyharness-openapi.mjs; then
+            echo "non-JSON output should have failed the script"
+            exit 1
+          fi
+          [ "$before" = "$(cat generated/openapi.json)" ] || { echo "invalid output reached disk"; exit 1; }
+          echo "ok: invalid output rejected"
--- a/.github/workflows/windows-sdk-generate-check.yml
+++ b/.github/workflows/windows-sdk-generate-check.yml
@@ -44,6 +44,10 @@ jobs:
   windows:
     name: package script on windows
     runs-on: windows-latest
+    # If cmd's `md` ever starts accepting the POSIX form, a reverted script would
+    # reach `cargo run`, and cargo is on this image. That turns a 20 second guard
+    # into a full Windows runtime build. Cap it.
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
@@ -105,6 +109,7 @@ jobs:
   linux:
     name: invoke and write path with a stubbed runtime
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
@@ -129,9 +134,14 @@ jobs:
       - name: Happy path writes a parseable schema
         working-directory: anyharness/sdk
         run: |
+          # Remove the tracked schema first. Without this a script that did
+          # nothing at all would still find a parseable file and pass.
+          rm -f generated/openapi.json
           CARGO="$RUNNER_TEMP/stub/cargo" npm run generate:openapi
           node -e "JSON.parse(require('fs').readFileSync('generated/openapi.json','utf8'))"
-          echo "ok: schema written and parses"
+          grep -q '"title":"stub"' generated/openapi.json \
+            || { echo "the file on disk is not what the stub emitted"; exit 1; }
+          echo "ok: the stub's schema was written and parses"
 
       - name: A failing runtime must not truncate an existing schema
         working-directory: anyharness/sdk
--- a/.github/workflows/windows-sdk-generate-check.yml
+++ b/.github/workflows/windows-sdk-generate-check.yml
@@ -1,132 +1,158 @@
 name: Windows SDK generate check
 
 # `anyharness/sdk`'s `generate:openapi` script used to be a POSIX shell one
-# liner (`mkdir -p ... && cd ../.. && cargo run ... > file`). npm runs package
-# scripts through cmd.exe on Windows, where `mkdir -p` creates a directory
-# literally named `-p`, so the step failed. That single line is the stated
-# reason desktop Windows releases were disabled in 3d3c0504b8 ("until the SDK
-# generation step is Windows-safe").
+# liner (`mkdir -p generated src/generated && cd ../.. && cargo run ... > file`).
+# npm runs package scripts through cmd.exe on Windows, where that command fails,
+# and that is the stated reason desktop Windows releases were disabled in
+# 3d3c0504b8 ("until the SDK generation step is Windows-safe").
 #
 # This is the regression guard, and it builds nothing.
 #
-# The windows job covers what was actually broken: directory creation and path
-# handling under cmd.exe. It cannot exercise the cargo invocation itself,
-# because Node refuses to spawn `.cmd` stubs without a shell (CVE-2024-27980)
-# and a real `cargo run` here would mean a full Windows runtime build. The
-# linux job covers the invoke, validate and write path with a shell stub. The
-# two together cover the whole script.
+# The guard runs the real `npm run generate:openapi`, not the node script
+# directly, because the regression being guarded lives in package.json. Reverting
+# package.json to the shell form must turn this red, and running the script
+# directly would leave it green.
+#
+# The discriminator is the DIRECTORIES, not the exit code. The runtime is
+# deliberately absent in the windows job, so the script fails either way; what
+# separates a working package script from a broken one is whether the two
+# directories exist afterwards. Under the shell form, cmd fails at mkdir and the
+# && chain never runs, so they do not.
+#
+# The windows job cannot exercise the cargo invocation: Node refuses to spawn
+# .cmd stubs without a shell (CVE-2024-27980) and a real cargo run would mean a
+# full Windows runtime build. The linux job covers invoke, validate and write
+# with shell stubs. The two together cover the whole script.
 
 on:
   pull_request:
     paths:
       - "scripts/generate-anyharness-openapi.mjs"
       - "anyharness/sdk/package.json"
+      - "Makefile"
       - ".github/workflows/windows-sdk-generate-check.yml"
   workflow_dispatch:
 
 permissions:
   contents: read
 
+concurrency:
+  group: windows-sdk-generate-check-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   windows:
-    name: directory handling on windows
+    name: package script on windows
     runs-on: windows-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version: 22
 
-      - name: Negative control, the old shell form is broken under cmd.exe
-        shell: cmd
+      # Informational and falsifiable. Driven from bash so the assertion is
+      # ours rather than cmd's errorlevel semantics, and so the real error text
+      # reaches the log.
+      - name: Negative control, the POSIX form does not work under cmd.exe
+        shell: bash
         run: |
-          mkdir oldform
-          cd oldform
-          mkdir -p generated src/generated
-          if exist "generated\openapi.json" (
-            echo UNEXPECTED: cmd.exe produced generated/openapi.json, this guard is no longer meaningful
-            exit /b 1
-          )
-          if not exist "generated" (
-            echo confirmed: the old form did not create a usable generated directory
-          )
-          if exist "-p" (
-            echo confirmed: cmd.exe created a directory literally named -p
-          )
+          mkdir -p "$RUNNER_TEMP/oldform"
+          cd "$RUNNER_TEMP/oldform"
+          echo "cmd.exe says:"
+          cmd //c "mkdir -p generated src/generated" 2>&1 || true
+          echo "what exists afterwards:"
+          ls -la
+          if [ -d generated ] && [ -d "src/generated" ]; then
+            echo "UNEXPECTED: cmd.exe created both directories, this guard is no longer meaningful"
+            exit 1
+          fi
+          echo "confirmed: cmd.exe did not create the directories the POSIX form asks for"
 
       # Both directories are checked in, so they have to go before the run or
-      # the assertion below would pass without the script doing anything.
+      # the assertion below would pass without the package script doing anything.
       - name: Remove the generated directories
         shell: bash
         run: rm -rf anyharness/sdk/generated anyharness/sdk/src/generated
 
-      - name: The node form recreates both directories on windows
+      - name: Run the real package script
         shell: cmd
+        working-directory: anyharness/sdk
         env:
           CARGO: this-runtime-does-not-exist
+        # `call` matters: npm is a .cmd shim, and without it the batch file
+        # terminates at npm and never reaches the exit. Exit 0 unconditionally,
+        # because the runtime is deliberately absent so the script must fail.
+        # The next step is the judge.
         run: |
-          cd anyharness\sdk
-          node ..\..\scripts\generate-anyharness-openapi.mjs
-          if "%ERRORLEVEL%" == "0" (
-            echo UNEXPECTED: a missing runtime should have failed the script
-            exit /b 1
-          )
-          echo confirmed: the script failed loudly on a missing runtime, as intended
+          call npm run generate:openapi
+          exit /b 0
 
-      - name: Assert both directories were recreated and nothing partial landed
+      - name: Assert the package script recreated both directories
         shell: bash
         run: |
-          test -d anyharness/sdk/generated || { echo "generated/ was not created"; exit 1; }
-          test -d anyharness/sdk/src/generated || { echo "src/generated/ was not created"; exit 1; }
-          test ! -e anyharness/sdk/generated/openapi.json || { echo "a schema was written despite the runtime failing"; exit 1; }
-          test ! -e "anyharness/sdk/-p" || { echo "the POSIX flag leaked into a directory name"; exit 1; }
-          echo "ok: both directories recreated on windows, no partial schema written"
+          fail=0
+          test -d anyharness/sdk/generated || { echo "generated/ was not created"; fail=1; }
+          test -d anyharness/sdk/src/generated || { echo "src/generated/ was not created"; fail=1; }
+          test ! -e anyharness/sdk/generated/openapi.json || { echo "a schema was written despite the runtime being absent"; fail=1; }
+          test ! -e "anyharness/sdk/-p" || { echo "the POSIX flag leaked into a directory name"; fail=1; }
+          if [ "$fail" -ne 0 ]; then
+            echo "the package script is not windows-safe"
+            ls -la anyharness/sdk
+            exit 1
+          fi
+          echo "ok: npm run generate:openapi created both directories on windows"
 
   linux:
     name: invoke and write path with a stubbed runtime
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version: 22
 
+      # Written with a quoted heredoc rather than printf, so the JSON's double
+      # quotes reach the stub intact.
       - name: Stub the runtime
         run: |
           mkdir -p "$RUNNER_TEMP/stub"
-          printf '#!/bin/sh\necho %s\n' '{"openapi":"3.0.0","info":{"title":"stub","version":"1"},"paths":{}}' > "$RUNNER_TEMP/stub/cargo"
-          chmod +x "$RUNNER_TEMP/stub/cargo"
+          cat > "$RUNNER_TEMP/stub/cargo" <<'STUB'
+          #!/bin/sh
+          echo '{"openapi":"3.0.0","info":{"title":"stub","version":"1"},"paths":{}}'
+          STUB
           printf '#!/bin/sh\nexit 101\n' > "$RUNNER_TEMP/stub/cargo-fail"
-          chmod +x "$RUNNER_TEMP/stub/cargo-fail"
           printf '#!/bin/sh\necho not json\n' > "$RUNNER_TEMP/stub/cargo-garbage"
-          chmod +x "$RUNNER_TEMP/stub/cargo-garbage"
+          chmod +x "$RUNNER_TEMP/stub/cargo" "$RUNNER_TEMP/stub/cargo-fail" "$RUNNER_TEMP/stub/cargo-garbage"
+          "$RUNNER_TEMP/stub/cargo" | node -e "let b='';process.stdin.on('data',d=>b+=d).on('end',()=>JSON.parse(b))"
+          echo "ok: the stub itself emits valid JSON"
 
       - name: Happy path writes a parseable schema
         working-directory: anyharness/sdk
         run: |
-          CARGO="$RUNNER_TEMP/stub/cargo" node ../../scripts/generate-anyharness-openapi.mjs
+          CARGO="$RUNNER_TEMP/stub/cargo" npm run generate:openapi
           node -e "JSON.parse(require('fs').readFileSync('generated/openapi.json','utf8'))"
           echo "ok: schema written and parses"
 
       - name: A failing runtime must not truncate an existing schema
         working-directory: anyharness/sdk
         run: |
-          before="$(cat generated/openapi.json)"
-          if CARGO="$RUNNER_TEMP/stub/cargo-fail" node ../../scripts/generate-anyharness-openapi.mjs; then
+          before="$(sha256sum generated/openapi.json | cut -d' ' -f1)"
+          if CARGO="$RUNNER_TEMP/stub/cargo-fail" npm run generate:openapi; then
             echo "a failing runtime should have failed the script"
             exit 1
           fi
-          after="$(cat generated/openapi.json)"
+          after="$(sha256sum generated/openapi.json | cut -d' ' -f1)"
           [ "$before" = "$after" ] || { echo "the existing schema was clobbered by a failed run"; exit 1; }
-          echo "ok: existing schema survived a failed runtime"
+          echo "ok: existing schema survived a failed runtime, sha $after"
 
       - name: Invalid output must be rejected before it is written
         working-directory: anyharness/sdk
         run: |
-          before="$(cat generated/openapi.json)"
-          if CARGO="$RUNNER_TEMP/stub/cargo-garbage" node ../../scripts/generate-anyharness-openapi.mjs; then
+          before="$(sha256sum generated/openapi.json | cut -d' ' -f1)"
+          if CARGO="$RUNNER_TEMP/stub/cargo-garbage" npm run generate:openapi; then
             echo "non-JSON output should have failed the script"
             exit 1
           fi
-          [ "$before" = "$(cat generated/openapi.json)" ] || { echo "invalid output reached disk"; exit 1; }
-          echo "ok: invalid output rejected"
+          after="$(sha256sum generated/openapi.json | cut -d' ' -f1)"
+          [ "$before" = "$after" ] || { echo "invalid output reached disk"; exit 1; }
+          echo "ok: invalid output rejected, sha unchanged"
--- a/Makefile
+++ b/Makefile
@@ -1569,16 +1569,13 @@ cloud-client-generate: cloud-openapi
 # SKIP_RUST=1 means "no cargo in this worktree". Honor it here too: the
 # prebuilt runtime emits the same schema, so a frontend-only worktree does not
 # need a toolchain just to regenerate the SDK. Without this the flag leaks and
-# `SKIP_RUST=1 make build` still invokes cargo.
+# `SKIP_RUST=1 make build` still invokes cargo. The cargo-vs-prebuilt-runtime
+# choice and the directory/file writes live in generate-anyharness-openapi.mjs
+# rather than in this recipe, because npm runs package scripts through cmd.exe
+# on Windows and a `mkdir -p ... && cargo run ... > file` shell chain breaks
+# there; see the script's header comment for details.
 sdk-generate:
-	mkdir -p anyharness/sdk/generated
-	@runtime_bin="$${ANYHARNESS_DEV_RUNTIME_BIN:-}"; \
-	if [ -n "$(SKIP_RUST)" ] && [ "$(SKIP_RUST)" != "0" ] && [ -n "$$runtime_bin" ] && [ -x "$$runtime_bin" ]; then \
-		echo "sdk-generate: SKIP_RUST set, using prebuilt runtime $$runtime_bin"; \
-		"$$runtime_bin" print-openapi > anyharness/sdk/generated/openapi.json; \
-	else \
-		$(CARGO) run --bin anyharness -- print-openapi > anyharness/sdk/generated/openapi.json; \
-	fi
+	CARGO="$(CARGO)" SKIP_RUST="$(SKIP_RUST)" node scripts/generate-anyharness-openapi.mjs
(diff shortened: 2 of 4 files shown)


pablonyx:
## Verdict

**No, not as it stands.** The production fix itself (the node script, the `package.json` swap, the Makefile swap) is sound and I could not break it, but both jobs of the regression guard this PR adds are red right now, and the mechanism the PR, the script header and the workflow header all state as the root cause is contradicted by the PR's own CI run. Fix the guard and correct the stated mechanism, then this is a yes.

---

## BLOCKING

### B1. The linux guard job fails on every run: the happy-path stub emits invalid JSON

`.github/workflows/windows-sdk-generate-check.yml:97`

```
printf '#!/bin/sh\necho %s\n' '{"openapi":"3.0.0","info":{"title":"stub","version":"1"},"paths":{}}' > "$RUNNER_TEMP/stub/cargo"
```

The JSON lands in the stub file **unquoted**. When `/bin/sh` runs `echo {"openapi":"3.0.0",...}`, POSIX quote removal strips every `"` before `echo` ever sees the word, so the stub prints `{openapi:3.0.0,info:{title:stub,version:1},paths:{}}`. The script under test then correctly rejects it and the "Happy path writes a parseable schema" step exits 1, taking the two steps that depend on it with it.

How I established it: reproduced the exact `printf` locally, then ran the resulting stub.

```
$ cat /tmp/stub/cargo
#!/bin/sh
echo {"openapi":"3.0.0","info":{"title":"stub","version":"1"},"paths":{}}

$ bash -c 'set +B; echo {"openapi":"3.0.0","info":{"title":"stub","version":"1"},"paths":{}}'
{openapi:3.0.0,info:{title:stub,version:1},paths:{}}
```

(`set +B` disables bash brace expansion so this matches Ubuntu's `dash`. On macOS `/bin/sh` it degrades further, to `openapi:3.0.0 info:title:stub info:version:1 paths:{}`.)

Confirmed against the real run, job `invoke and write path with a stubbed runtime` (96645227002):

```
print-openapi did not produce valid JSON: Expected property name or '}' in JSON at position 1 (line 1 column 2)
##[error]Process completed with exit code 1.
```

Fix is to quote the payload inside the generated script: `printf '#!/bin/sh\necho '\''%s'\''\n' '{...}'`. I applied exactly that locally and replayed all three linux steps verbatim: `STEP1_OK`, `STEP2_OK`, `STEP3_OK`. So the job is one quoting fix away from being correct.

### B2. The windows negative control fails the job, and its assertion is unfalsifiable

`.github/workflows/windows-sdk-generate-check.yml:40-55`

Two separate defects in one step.

**It is red today.** Under `cmd.EXE /D /E:ON /V:OFF /S /C`, `mkdir -p generated src/generated` prints `The syntax of the command is incorrect.` and leaves errorlevel 1. `echo` does not reset errorlevel in cmd, and the trailing `if exist "-p" (...)` branch is not taken, so errorlevel 1 survives to the end of the batch and GitHub fails the step. From the real run, job `directory handling on windows` (96645227169):

```
The syntax of the command is incorrect.
confirmed: the old form did not create a usable generated directory
##[error]Process completed with exit code 1.
```

Because this is the first step, **the windows job never reaches the part that tests the fix.** There is currently zero Windows evidence for this PR.

**The assertion can never fire.** Line 46 fails the job `if exist "generated\openapi.json"`. Nothing in that step ever produces `openapi.json` - it runs `mkdir` only, no `cargo`, no redirect. The guard's stated purpose ("fails the job if that form somehow works, so the guard cannot quietly stop being meaningful") is not implemented: the only falsifiable branch is unreachable. The two `confirmed:` branches at :50 and :53 are bare `echo`s with no `exit /b`, so they assert nothing either. Established by reading, and corroborated by the run log above, where the "created a directory literally named -p" line never printed.

### B3. The stated root-cause mechanism is wrong, and this PR's own CI disproves it

Same claim appears in three places:

- `scripts/generate-anyharness-openapi.mjs:7-8` - "`mkdir -p` creates a directory literally named `-p` and the step then fails"
- `.github/workflows/windows-sdk-generate-check.yml:5-6` - same wording
- PR body - "`-p` is read as a directory name"

The windows run shows cmd created **nothing at all**: no `-p`, no `generated` (the `if not exist "generated"` branch printed), no `src\generated`. `md` rejected the whole command line with `The syntax of the command is incorrect.`, which is the forward slash in `src/generated` being parsed as a switch, not `-p` being taken as a directory name.

The conclusion (the old line is broken on cmd.exe) is right; the explanation is not. Also unstated and arguably the more important half: on a real checkout `generated/` and `src/generated/` are both **tracked and already present**, and cmd's `md` errors on an existing directory, so even a slash-free variant would have failed. This matters beyond pedantry because the negative control at :43-44 deliberately runs inside a fresh empty `oldform/`, which is the one condition under which the real-world failure mode cannot reproduce.

### B4. The guard does not guard the thing that broke

`.github/workflows/windows-sdk-generate-check.yml:69` runs `node ..\..\scripts\generate-anyharness-openapi.mjs` directly. What was broken was **npm running the package script through cmd.exe**. That path is never exercised. Revert `anyharness/sdk/package.json` to the shell one-liner and this workflow, which explicitly watches `anyharness/sdk/package.json` in its `paths:` filter at :22, stays green. Running `npm run generate:openapi` from `anyharness\sdk` on `windows-latest` (with `CARGO` pointed at nothing, asserting a non-zero exit and the two directories) would be a true end-to-end guard and would have caught the wrong mechanism in B3 too.

---

## NON-BLOCKING

### N1. `??` should be `||` for `CARGO`
`scripts/generate-anyharness-openapi.mjs:28`. `process.env.CARGO ?? "cargo"` treats an empty-string `CARGO` as set. `make CARGO= sdk-generate`, or any shell exporting `CARGO=`, produces an uncaught throw with a stack trace instead of a diagnostic:
```
$ CARGO="" node scripts/generate-anyharness-openapi.mjs
TypeError [ERR_INVALID_ARG_VALUE]: The argument 'file' cannot be empty. Received ''
```

### N2. `make CARGO="cargo +nightly" sdk-generate` now breaks
`Makefile:1510`. The old recipe used `$(CARGO)` as a shell command word, so a `CARGO` carrying arguments worked. `spawnSync` without a shell treats the whole string as one executable name. Verified: `CARGO="/tmp/stub/real --extra" node script` -> `spawnSync /tmp/stub/real --extra ENOENT`. A `CARGO` that is merely a path *with spaces* is fine (verified: `/tmp/dir with space/cargo` wrote 540317 bytes, exit 0). Low impact, nobody in the repo sets `CARGO` other than `Makefile:2`, but worth a comment or a `.split(" ")` decision.

### N3. The npm path now honours `$CARGO`, where before it hardcoded `cargo`
New environment sensitivity on the `pnpm run generate` path. `cargo` exports `CARGO` to every process it spawns, so a nested invocation now resolves differently than it used to. Benign today; note it.

### N4. Signal death prints `exited with null`
`scripts/generate-anyharness-openapi.mjs:41-42`. Verified with a stub that `kill -9`s itself: `... print-openapi exited with null`, exit 1. Exit code handling is correct (`?? 1`), the message is not. Consider reporting `result.signal`.

### N5. Non-UTF-8 stdout is silently transcoded rather than passed through
`encoding: "utf8"` + `writeFileSync(string)` is not a byte-preserving pipe. Verified: a stub emitting `{"a":"\xff"}` produced `7b22 6122 3a22 efbf bd22 7d0a` - the `0xFF` became `EF BF BD`. Unreachable in practice because serde_json output comes from a Rust `String`, but it is a real difference from `>`. Separately, a UTF-8 BOM now hard-fails (`did not produce valid JSON`) where `>` would have written it. Both are in the safe direction; flagging only so nobody assumes byte transparency.

### N6. No shape validation
`JSON.parse` accepts `[]` and `null`. Verified: a stub printing `[]` wrote `[]` and exited 0. Same as the old behaviour, so not a regression, but `typeof parsed !== "object" || parsed === null || Array.isArray(parsed)` would close it while you are here.

### N7. Actions are unpinned and on an old major
`:34`, `:36`, `:89`, `:90` use `actions/checkout@v4` and `actions/setup-node@v4`. Every other workflow in the repo SHA-pins these (65x `actions/checkout@3d3c42e5...` v7, 44x `actions/setup-node@820762786...` v7). The run emits `Node.js 20 is deprecated... forced to run on Node.js 24` as a result. Also, no `concurrency:` group, which 12 of the 29 workflows here carry.

### N8. Linux job's before/after comparison is newline-insensitive
`:114`/`:119` and `:126`/`:131`. `"$(cat ...)"` strips trailing newlines on both sides, so a change confined to the trailing newline would pass. `cmp` against a saved copy is the sharper check. It does catch the truncation-to-zero case it is aimed at.

### N9. Trigger gaps
`paths:` at :20-24 omits `Makefile`, which is one of the two call sites this PR is unifying. And `pull_request` only, so nothing runs on `main`.

---

## What I verified, and how

**Byte-for-byte equivalence with the old `>` redirect - clean, this was my main concern.**

- Real 540317-byte payload piped through the script and compared with both the committed file and the output of the old shell form: `cmp` silent both times. `BYTE IDENTICAL for the real 540KB payload`, `BYTE IDENTICAL vs old '>' redirect`.
- 38488891-byte payload: `BYTE IDENTICAL for 40MB`, 0.46s.
- Trailing newline preserved exactly (`[]\n` in, `5b5d 0a` on disk).
- The committed `openapi.json` contains **zero** bytes >= 0x80, so the UTF-8 decode/re-encode round trip is a no-op for this payload today.
- `maxBuffer` overflow is not silent. Forced it with `maxBuffer: 1024`: spawnSync sets `error.code = ENOBUFS`, `status: null`, `signal: SIGTERM`, and the script's `result.error` branch catches it and exits 1 with the existing file untouched. 256MB against a 540KB payload is ample.
- **The strongest evidence is your own CI**: `Build SDK` (96645227172) is green on this PR, and that job runs `pnpm run generate` with a real toolchain followed by `git diff --exit-code -- anyharness/sdk/generated/openapi.json anyharness/sdk/src/generated/openapi.ts` (`.github/workflows/ci.yml:298-304`). The drift gate is satisfied. This claim in the PR body is fully substantiated.

**Every row of your verification table reproduced**, invoked from `anyharness/sdk` as npm invokes it:

```
schema on stdout    -> wrote .../generated/openapi.json          exit=0
no output           -> refusing to write an empty schema         exit=1
whitespace only     -> refusing to write an empty schema         exit=1
non-JSON            -> did not produce valid JSON                exit=1
exit 101            -> ... exited with 101                       exit=101, file byte-identical
missing binary      -> failed to run /does/not/exist: ... ENOENT exit=1
CARGO is a directory-> failed to run /tmp/stub: ... EACCES       exit=1
```

**The truncation claim is true**, reproduced verbatim against the old line:
```
before: {"real":"schema"}
compile error
exit=101
after: [] size=0
```
And the new script leaves the file untouched under the same stub.

**The shell-ism sweep is accurate.** I re-ran it wider than you did (every non-`node_modules` `package.json`, matching `mkdir`, `rm -`, `cp`, `mv`, `chmod`, `touch`, `ln -`, `$(...)`, single quotes, POSIX pipes, and leading `FOO=bar` env prefixes): `TOTAL 1`, and it is `anyharness/sdk`'s `generate:openapi`. There is no `.npmrc` and `pnpm-workspace.yaml` sets no `shell-emulator`, so pnpm does run scripts through cmd.exe on Windows. The premise holds.

**The `generate-updater-manifest.mjs` fail-closed claim is true.** `scripts/generate-updater-manifest.mjs:113-134`: missing signature or artifact pushes onto `errors`, and `if (errors.length > 0) { ...; process.exit(1); }`. Adding a `windows-x86_64` key without a guaranteed artifact would indeed break macOS releases.

**`3d3c0504b8` is quoted accurately.** It removed the `windows-latest` matrix entry, the `windows` dispatch option and the Windows artifact upload from `release-desktop.yml`, dropped 9 lines from `generate-updater-manifest.mjs`, and added the doc note verbatim as quoted. And `release-desktop.yml:299` does run `cd anyharness/sdk && pnpm run generate` inside the per-OS matrix job, so the Windows release really did hit that line.

**Makefile and npm path robustness - no problems found.** The script derives `repoRoot` from `import.meta.url`, so it is cwd-independent; the only cwd dependence left is `../../scripts/...` in `package.json`, and npm/pnpm always set cwd to the package directory (green `Build SDK` proves it, including under `pnpm install`/workspace resolution). `make -C` chdirs, so the recipe's relative `scripts/...` resolves; `make -f /abs/Makefile` from elsewhere breaks, but the old recipe's relative `anyharness/sdk/generated` broke identically, so no regression. `sdk-build`, `test-agent-runtime-local` and `test-agent-runtime-cloud-e2b` all depend on `sdk-generate` only for the artifact, which is unchanged. The new recipe additionally creates `anyharness/sdk/src/generated`, which the old Makefile recipe did not - strictly better.

**Static checks:** `node --check scripts/generate-anyharness-openapi.mjs` OK. `python3 -c "yaml.safe_load(...)"` OK. `actionlint -shellcheck=shellcheck .github/workflows/windows-sdk-generate-check.yml` exit 0 (it does not catch B1, which is a quoting bug inside a generated file).

---

## What I could not verify

- **Windows behaviour beyond what your CI run already printed.** No Windows machine here. Everything I say about cmd.exe comes from the job log on run 32438811348, not from a local reproduction. In particular I did not independently confirm *which* argument makes `md` emit `The syntax of the command is incorrect.`; I infer the forward slash in `src/generated` from the fact that nothing at all was created, but that is inference.
- **The `%ERRORLEVEL%` check at :70.** I believe it is correct (the `if` statement is parsed after `node` has run, so the non-delayed expansion picks up the fresh value), but the job dies at :45 before reaching it, so it has never actually executed. Unverified.
- **Whether the counterfactual holds** - that with `mkdir -p` fixed, the Windows desktop release would have gotten past the SDK step. Nothing in this PR or its CI demonstrates that, and the PR is honest about not claiming to re-enable Windows.
- I ran no `cargo` anything, per the machine constraints, so every runtime invocation above used stub binaries.

pablonyx:
## Re-verification of `54eaa24323`: all four blocking findings cleared

**Safe for Pablo to merge.** Confirm on all four. Both guard jobs are green on run 32439665847, `Build SDK` is green on 32439665790 so the drift gate still holds, and I re-verified the production script is unchanged in behaviour. Four small nits below, none blocking.

### 1. Is B4 genuinely falsifiable? Yes, and it is double-guarded

Reasoned from the job definition, not the claim. Under a revert of `package.json` to the shell one liner:

`:75` deletes both tracked directories. `:87` then runs `call npm run generate:openapi`, npm hands the script to cmd.exe, and cmd hits `mkdir -p generated src/generated`. This job's own negative control at `:56-69` proves on the same runner image, in the same run, that this creates nothing:

```
cmd.exe says:
The syntax of the command is incorrect.
what exists afterwards:
total 4
drwxr-xr-x 1 runneradmin 197121 0 Aug 21 02:23 .
drwxr-xr-x 1 runneradmin 197121 0 Aug 21 02:23 ..
```

Nonzero errorlevel, so `&&` short-circuits and nothing further runs. `test -d anyharness/sdk/generated` at `:94` then fails and the job goes red. Falsifiable.

I tried to find a green-after-revert path and could not:

- **Partial mkdir success** (a future cmd creates `generated` but not `src/generated`): `:95` fires. Red.
- **Full mkdir success**: the negative control at `:65-68` fires first. Red. The two assertions are complementary with no gap between them, which is more than you claimed.
- **`>` creates an empty schema**: `:96` fires. Red.
- The only remaining green-after-revert world is npm or pnpm shipping Windows shell emulation, in which case the shell form genuinely works and green is the correct answer. Not a defect.

### 2. `call npm run` + `exit /b 0`: does what you think, and fails safe

Confirmed from the log rather than from reasoning. The step ran the real package script end to end:

```
> @anyharness/sdk@0.4.19 generate:openapi
> node ../../scripts/generate-anyharness-openapi.mjs
failed to run this-runtime-does-not-exist: spawnSync this-runtime-does-not-exist ENOENT
```

and the step is green with no `##[error]`, so `call` returned control and `exit /b 0` was reached. `package.json` is genuinely in the path now.

The scenario you asked about, the assertion running against a state the package script never touched, resolves to **red, not a false green**: if `call npm` had not run at all, `:75` has already deleted both directories, so `:94`/`:95` fail. There is no false-green path there.

One cost worth a comment line: `exit /b 0` discards the package script's exit code entirely, so the windows job can no longer catch "the script exits 0 when it should have failed" - the previous version's `%ERRORLEVEL%` check did. A script that created the directories and exited 0 without writing would satisfy all four assertions on Windows. The linux job's `if ... npm run ...; then exit 1; fi` covers exit-code correctness, so the split is defensible, it is just now implicit.

### 3. The mechanism statement: accurate and correctly hedged

Workflow header `:3-7` and `:16-20`, script header `:7-9`, and the PR body all now say cmd fails and creates nothing, with no claim about which argument trips it, and the PR body quotes the runner output. That is exactly the right level. No over-claim remaining.

One inconsistency: the script header at `:9` still says "That single line is **why** desktop Windows releases were disabled", while the workflow header at `:6` says "that is the **stated reason**". The workflow's wording is the defensible one - it is sourced from the commit's doc note, not from a demonstrated counterfactual. Worth matching them.

### 4. The non-blocking items: all taken correctly

- **`||` for `CARGO`** (`scripts/generate-anyharness-openapi.mjs:29`). Verified: `CARGO="" PATH=<no cargo> node ...` now gives `failed to run cargo: spawnSync cargo ENOENT`, exit 1, no stack trace. Nit: the comment above it says the point is that `""` "throws ERR_INVALID_ARG_VALUE instead of saying anything useful", which reads as though the empty value is now reported. It is not - it silently falls back to `cargo`. That is the behaviour I asked for; just say so.
- **`Makefile` in `paths:`** `:32`, correct.
- **SHA pins** `:48`, `:49`, `:109`, `:110` match the repo standard byte for byte (`checkout@3d3c42e5...` v7.0.1, `setup-node@820762786...` v7.0.0), and the `Node.js 20 is deprecated` warning is gone from the new run.
- **`concurrency`** `:39-41` matches the house shape exactly (`codeql.yml` uses `codeql-${{ github.ref }}` + `cancel-in-progress: true`).
- **`sha256sum`** `:139`/`:144`/`:151`/`:156`, correct, and the sha is echoed: `ok: existing schema survived a failed runtime, sha f1f2de4f...`.
- **Heredoc stub** `:119-122` dedents correctly and the self-test at `:126` proves it: `ok: the stub itself emits valid JSON`. All four linux steps printed their `ok:` lines.

### Remaining nits, none blocking

1. Script header `:9` "is why" vs workflow header `:6` "is the stated reason". Match them.
2. The `||` comment at `:28-29` describes the old failure, not the new behaviour.
3. Linux happy path `:132-134` asserts only that `generated/openapi.json` parses. The checked-in file also parses and the linux job never deletes it, so a hypothetical silent no-op stays green - same vacuity class as the old B2 assertion. One `grep -q '"title":"stub"'` would close it, and would also pin the `before` shas at `:139`/`:151` to the stub payload rather than to whatever was on disk.
4. In the hypothetical world where cmd's `md` starts working, the reverted script would proceed to `cd ../.. && cargo run ...`, and cargo **is** on the `windows-latest` image, so this 21-second job would attempt a full Windows runtime build. `:96` catches it eventually via the empty `openapi.json` the `>` creates, but only after a very long build. A `timeout-minutes: 10` on the job is a cheap hedge.

### What I re-checked, and how

`node --check` on the updated script OK. `yaml.safe_load` on the updated workflow OK. `actionlint -shellcheck=shellcheck` exit 0. Replayed the real 540317-byte payload through the updated script: `cmp` silent against both the committed file and the old `>` redirect output, so the `||` change did not disturb byte equivalence, and `Build SDK` on run 32439665790 is green, which is the authoritative confirmation. Read both job logs in full rather than trusting the green ticks.

### What I still cannot verify

No Windows machine here, so every cmd.exe statement above comes from run 32439665847's log. The revert counterfactual in section 1 is reasoning over the job definition plus that observed negative control, not an executed revert.

pablonyx:
## Review: base-of-stack, post-rebase

**Rebase integrity.** `git range-diff c704b04d5f..933584620d 956d2426b7..296e6fc428` shows all three commits as `=` (identical patch-ids). Zero content drift from the pre-rebase tip; only the base moved.

**Prior round (not re-litigated).** The four blocking guard defects from the previous review are confirmed fixed and green: quoted heredoc for the linux JSON stub, bash-driven windows negative control, both guard jobs invoking `npm run generate:openapi` rather than the node script, and the corrected "cmd creates nothing" wording. Byte-equivalence against the 540,317-byte payload isn't something I re-ran; taking it as established per the brief.

**Verified this round:**

1. `scripts/generate-anyharness-openapi.mjs`: `repoRoot` resolves via `join(dirname(fileURLToPath(import.meta.url)), "..")` — not cwd-dependent. Confirmed.
2. Write ordering is correct: `result.error` → `result.status !== 0` → empty/blank `stdout` → `JSON.parse` (try/catch) — all four checks happen before the single `writeFileSync`. This does fix a real cross-platform bug (the old `>` redirect truncated the target before `cargo run` executed), not just a Windows issue, and the PR body states this correctly under "A second bug this removes."
3. `anyharness/sdk/package.json`: `generate:openapi` → `node ../../scripts/generate-anyharness-openapi.mjs`. Both `windows-sdk-generate-check.yml` jobs invoke `npm run generate:openapi` (windows: `call npm run generate:openapi`; linux: `CARGO=... npm run generate:openapi`), not the script directly — reverting `package.json` would correctly turn the guard red.
4. Linux guard's valid-JSON stub uses a quoted heredoc (`cat > "$RUNNER_TEMP/stub/cargo" <<'STUB' ... STUB`), avoiding the `printf` quote-stripping bug from the earlier round. (`printf` is still used for the `cargo-fail`/`cargo-garbage` stubs, but those don't emit JSON, so no quotes are at risk there.)
5. Windows-behavior claim: script header, workflow header, and PR body all say cmd's `md` reports "The syntax of the command is incorrect." and creates nothing. Grepped the full diff for the old false "-p directory" claim — no surviving instances anywhere.
6. Strongest proof, confirmed green on the rebased head (run 32450298564, job "Build SDK", commit 296e6fc428): "Regenerate SDK artifacts" runs the real `pnpm run generate` → `node .../generate-anyharness-openapi.mjs` → real `cargo run --bin anyharness -- print-openapi` (actual crates.io fetch/compile, not a stub), logs `wrote .../openapi.json`, and the following "Verify generated SDK artifacts are committed" step (`git diff --exit-code` on both generated files) passes with no diff. The Windows guard's two jobs are also both green on this head.

All PR checks are green or expected-skip except `intent-tests (provisional)`, which is unrelated/pending and not in this PR's scope.

**Verdict: safe to merge as-is.** No blocking items. Only a documentation note, non-blocking: the "What review found, and what it cost" section is useful history but is unusually long for a shipped PR body — no action needed, just flagging in case it's trimmed before merge.

(No push/merge/label/ready-for-review action taken, per instructions — PR remains draft.)

pablonyx (commented):
## Review of the rebase (head `129e0469`) — non-blocking, no blocking findings

Fresh reviewer, did not implement this. I reviewed the rebase specifically, because it is not a
replay: it **ports main's `SKIP_RUST` fast path out of the Makefile recipe and into the node
script**, swapping shell `[ -x ]` for `fs.accessSync(X_OK)`. That rewrite is the whole novel
content, so I went at it empirically rather than by reading.

All work was done in a throwaway `/tmp` clone. No cargo, no rustc, no Docker — the real cargo path
is covered by CI instead (see V6).

---

### V1 — Behavioural equivalence of the ported `SKIP_RUST` branch: verified, 13/13 identical

I reconstructed main's recipe body verbatim as a shell script (with `$(SKIP_RUST)`/`$(CARGO)`
pre-expanded exactly as `make -n` shows them) and ran it head-to-head against the new script with
stub executables that tag their own output, so the *selected branch* is observable rather than
inferred.

First, the vacuity guard — my shell had nothing leaked, and every case ran under
`env -u SKIP_RUST -u ANYHARNESS_DEV_RUNTIME_BIN`:

```
SKIP_RUST=[<UNSET>] ANYHARNESS_DEV_RUNTIME_BIN=[<UNSET>] CARGO=[<UNSET>]
```

| # | case | main (shell) | script | verdict |
|---|---|---|---|---|
| 1 | `SKIP_RUST` unset (make sends `""`), bin valid | CARGO | CARGO | SAME |
| 2 | `SKIP_RUST=0`, bin valid | CARGO | CARGO | SAME |
| 3 | `SKIP_RUST=1`, bin valid + exec | RUNTIME | RUNTIME | SAME |
| 4 | `SKIP_RUST=1`, bin missing | CARGO | CARGO | SAME |
| 5 | `SKIP_RUST=1`, bin present but **not executable** | CARGO | CARGO | SAME |
| 6 | `SKIP_RUST=false`, bin valid | RUNTIME | RUNTIME | SAME |
| 7 | `SKIP_RUST=no`, bin valid | RUNTIME | RUNTIME | SAME |
| 8 | `SKIP_RUST=""` explicit, bin valid | CARGO | CARGO | SAME |
| 9 | `SKIP_RUST=1`, bin env unset | CARGO | CARGO | SAME |
| 10 | `SKIP_RUST=1`, bin env empty string | CARGO | CARGO | SAME |
| 11 | `SKIP_RUST=" "` (single space), bin valid | RUNTIME | RUNTIME | SAME |
| 12 | `SKIP_RUST=1`, bin is a **directory** | fail (126) | fail (1) | SAME class |
| 13 | `SKIP_RUST=00`, bin valid | RUNTIME | RUNTIME | SAME |

The two truthiness models line up because the guard is written as `Boolean(x) && x !== "0"`, which
is exactly `[ -n "$x" ] && [ "$x" != "0" ]`. `Boolean("")` and `Boolean(undefined)` are both false,
matching `[ -n "" ]`; `Boolean("0")` is true but the `!== "0"` clause catches it. Cases 6, 7, 11 and
13 are the ones where both treat a "looks falsey" string as truthy — surprising, but **identically**
surprising, which is what a rebase owes.

Case 12 is the only place the two differ at all, and only in the exit code (`126` vs `1`), not the
outcome: `[ -x dir ]` is true for a directory and `accessSync(dir, X_OK)` also succeeds, so both
select the runtime bin and both then fail to exec it. Both abort make.

On the specific "does an unset make variable become the literal empty string" question — yes, and
it is handled. `make -n sdk-generate` with `SKIP_RUST` unset in the environment:

```
CARGO="/Users/…/.cargo/bin/cargo" SKIP_RUST="" node scripts/generate-anyharness-openapi.mjs
```

so the script always sees `""`, never `undefined`, and case 1 covers it. Compare main's expansion in
the same condition, which collapses to `[ -n "" ]`:

```
if [ -n "" ] && [ "" != "0" ] && [ -n "$runtime_bin" ] && [ -x "$runtime_bin" ]; then
```

Same gate.

### V2 — Failure propagation: verified, no `#2134`-class swallow

This was the finding I most expected, given the recent build-rust regression where a rewritten
recipe swallowed a cargo failure and copied a stale binary forward. It is not present here. Every
path exits non-zero:

```
=========== FAILURE: cargo-fail (exit 101) ===========
/tmp/…/cargo-fail run --bin anyharness -- print-openapi exited with 101
script rc=101                      existing schema UNTOUCHED
=========== FAILURE: cargo-garbage (non-JSON) ===========
print-openapi did not produce valid JSON: Unexpected token 'o', "not json…
script rc=1                        existing schema UNTOUCHED
=========== FAILURE: cargo-empty (no output) ===========
print-openapi produced no output; refusing to write an empty schema
script rc=1                        existing schema UNTOUCHED
=========== FAILURE: cargo-trunc (valid JSON prefix, truncated) ===========
print-openapi did not produce valid JSON: Unterminated string in JSON at position 22
script rc=1                        existing schema UNTOUCHED
=========== FAILURE: runtime-bin path fails (SKIP_RUST=1) ===========
sdk-generate: SKIP_RUST set, using prebuilt runtime /tmp/…/cargo-fail
/tmp/…/cargo-fail print-openapi exited with 101
script rc=101                      existing schema UNTOUCHED
=========== command not found ===========
failed to run /definitely/not/here: spawnSync /definitely/not/here ENOENT
script rc=1
```

Note the failing *runtime-bin* branch is covered too, not just cargo — that is the newly-ported path.

And make still aborts, so the second recipe line never consumes a bad schema (stub `CARGO` passed on
the command line; no real cargo was run):

```
$ make CARGO=/tmp/…/cargo-fail sdk-generate
CARGO="/tmp/…/cargo-fail" SKIP_RUST="" node scripts/generate-anyharness-openapi.mjs
/tmp/…/cargo-fail run --bin anyharness -- print-openapi exited with 101
make: *** [sdk-generate] Error 101
MAKE EXIT=2
openapi.json still intact: {"who":"GOOD"}
```

I also checked the one silent-truncation vector the script's design could have had —
`spawnSync` exceeding `maxBuffer`. On node 22 that surfaces as an error rather than a quiet trim,
so the first guard catches it:

```
$ node mb.mjs ./big.sh     # maxBuffer: 64, child emits 510 bytes
{"status":null,"signal":"SIGTERM","error":"ENOBUFS","stdoutLen":510}
```

The script's `if (result.error)` handles that, and the `JSON.parse` guard is a second net behind it.
The cap is 256 MB against a 541,185-byte real schema, so this is theoretical anyway.

### V3 — Output integrity: **strictly better** than main, and I have the negative control

You asked whether the script is better, worse or the same as `>` redirection. It is better, and the
difference is not cosmetic. Main's form truncates the file *before* the command runs, so a cargo
failure leaves a zero-byte schema on disk:

```
$ mainform.sh "" ./cargo-fail ./main-seed.json      # negative control
mainform rc=101
file after MAIN failure: [] size=0
```

The script buffers, checks non-empty, `JSON.parse`s, and only then writes — so the same failure
leaves the previous good file intact (every `existing schema UNTOUCHED` line above). The
`JSON.parse`-before-`writeFileSync` ordering means a runtime that dies halfway through emitting
cannot land a truncated schema at all, which the shell form could.

The residual: `writeFileSync` is not atomic, so a failure *during* the write (ENOSPC) can still
leave a partial file. That is no worse than the shell form and I am not asking for a temp-file
rename here.

### V4 — Directory creation: superset of both predecessors

The old Makefile did `mkdir -p anyharness/sdk/generated`; the old package.json did
`mkdir -p generated src/generated`. The script creates **both**, unconditionally, before the spawn —
so neither caller regresses, and `generate:types` still finds `src/generated` to write into.
Verified by deleting both and running:

```
# success path
wrote /tmp/…/anyharness/sdk/generated/openapi.json
  → generated/  and  src/generated/  both present

# failure path (cargo-fail), dirs deleted first
… exited with 101   rc=101
  → generated/  and  src/generated/  both present, generated/ is EMPTY (no partial openapi.json)
```

That last line matters: it is exactly the discriminator the new Windows guard asserts on.

Also confirmed the npm entry point still resolves the repo root correctly now that `cd ../..` is
gone — `import.meta.url` does the job:

```
$ cd anyharness/sdk && CARGO=…/cargo-good npm run generate:openapi
> node ../../scripts/generate-anyharness-openapi.mjs
wrote /tmp/…/repo/anyharness/sdk/generated/openapi.json
```

### V5 — Rebase fidelity: nothing lost, nothing unrelated added

Pre-rebase head was `296e6fc4` (force-pushed 08-21 08:01Z). Diffing the two trees:

- `.github/workflows/windows-sdk-generate-check.yml` — **byte-identical**, no diff.
- `anyharness/sdk/package.json` — the only delta is `0.4.19 → 0.4.22`, which came from main, not
  from this branch. The `generate:openapi` line is unchanged from the pre-rebase version.
- `scripts/generate-anyharness-openapi.mjs` — the diff is *exactly* the `SKIP_RUST` port plus the
  `cargo`→`command` renaming in the two error messages. Nothing else moved.

Cumulative diff vs merge-base is 4 files, all in scope:

```
 .github/workflows/windows-sdk-generate-check.yml | 168 ++++++++
 Makefile                                         |  15 +-
 anyharness/sdk/package.json                      |   2 +-
 scripts/generate-anyharness-openapi.mjs          |  92 ++++++
```

Main has moved on past the merge-base (`947d7a4`, #2169) but touched only `scripts/ci-cd/*`, so the
Makefile is not stale. `git log -S "sdk-generate: SKIP_RUST set" -- Makefile` confirms `fe6780e`
(#2134) added the fast path and `ad09640` here is what ports it.

### V6 — The real cargo path is proven by CI, so I did not need to run it

`Build SDK` (pass, 2m48s) runs `pnpm run generate` with a real toolchain and then:

```
git diff --exit-code -- anyharness/sdk/generated/openapi.json anyharness/sdk/src/generated/openapi.ts
```

Green means the new script produces byte-identical output to the committed schema. That is the
end-to-end equivalence proof for the cargo branch, and it is stronger than anything I could have run
locally. All 39 checks are green, including both new jobs.

---

## Non-blocking findings

**F1 (the one I would actually act on) — the guard this PR adds does not cover the branch this
rebase adds.** The linux job stubs `CARGO` and exercises invoke / validate / write / fail-clean. It
never sets `SKIP_RUST` or `ANYHARNESS_DEV_RUNTIME_BIN`, so the `useRuntimeBin` selection — the
entire novel content of the rebase, and the part that had to be re-expressed in a different language
with different truthiness rules — has zero CI coverage. My table above is the only evidence it
behaves, and it lives in a review comment, not in the repo. A fourth linux step would cost seconds:
stub a `runtime` binary, assert `SKIP_RUST=1` + executable bin picks it (grep the `using prebuilt
runtime` line or a tagged schema), and assert `SKIP_RUST=1` with a non-executable bin falls back to
`CARGO`. Those two assertions pin the whole ported gate.

**F2 — the guard uses `npm`, the path it guards uses `pnpm`.** `release-desktop.yml:322` runs
`cd anyharness/sdk && pnpm run generate`. The Windows guard runs `call npm run generate:openapi`.
Both dispatch package scripts through `cmd.exe` on Windows so the conclusion holds today, but the
guard's entire value is fidelity to the broken path, and `pnpm` is one `shell-emulator` default away
from behaving differently. Cheap to switch.

**F3 — `accessSync(X_OK)` is not `[ -x ]` on Windows.** libuv's `fs__access` implements `X_OK` as a
pure existence check (it only special-cases `W_OK` against `FILE_ATTRIBUTE_READONLY`), so on Windows
*any* existing file passes. Consequence: a present-but-unrunnable `ANYHARNESS_DEV_RUNTIME_BIN` gets
selected and the script hard-fails at spawn instead of falling back to cargo. Not a regression — the
shell form never ran on Windows at all — and arguably the louder behaviour is the better one. Worth
a comment on the `accessSync` call so the next reader does not assume POSIX semantics. Related, and
also matching main: nothing appends `.exe`, so a Windows `ANYHARNESS_DEV_RUNTIME_BIN` set without
the suffix silently falls back to cargo.

**F4 — the second recipe line's `cd … && npx …`: I checked, and it is not a real gap.** It is still
POSIX-shaped, but nothing reaches it on Windows. `make sdk-generate` is invoked only from other
Makefile targets and `scripts/dev-build.mjs:169`; the Windows desktop release goes through
`release-desktop.yml`'s `pnpm run generate` under `shell: bash`, never through make. The Makefile is
POSIX throughout anyway (e.g. the `@if [ -n … ]` block at :1614). Leaving it alone is the right call
for this PR's stated scope — flagging it only so the decision is on the record rather than an
oversight.

**F5 — relative-path asymmetry, theoretical.** `accessSync(runtimeBin)` resolves a relative
`ANYHARNESS_DEV_RUNTIME_BIN` against `process.cwd()`, but `spawnSync` runs it with `cwd: repoRoot`.
Those agree when make invokes the script from the repo root, and disagree when npm invokes it from
`anyharness/sdk`. In practice the variable is always absolute (`~/.proliferate-local/…`). Resolving
it against `repoRoot` before the access check would close it.

**F6 — the recipe lost its `@` prefix**, so make now echoes `CARGO="/Users/…/.cargo/bin/cargo"
SKIP_RUST="" node …` on every run. Cosmetic; arguably an improvement for debuggability.

**F7 — the workflow is `paths:`-filtered.** Correct for a targeted guard, but if it is ever promoted
to a required check it will sit at "expected" forever on PRs that do not touch those four paths.
Note only.

**F8 — the `|| "cargo"` fallback comment is slightly narrower than it reads.** `CARGO :=
$(HOME)/.cargo/bin/cargo` at Makefile:2 is never empty, so the fallback only ever fires on the
npm/pnpm entry point where `CARGO` is unset. Accurate, just not obviously so from the recipe side.

---

## Stacked PRs — reported, not touched

- **#2146** (this): `MERGEABLE`, `mergeStateStatus: BLOCKED` (awaiting approval), all 39 checks pass.
- **#2150** "fix(release): restore the windows desktop release path": based on
  `fix/sdk-generate-windows-safe`, currently `CONFLICTING` / `DIRTY` against this branch's new head.
- **#2157** "fix(build): normalise line endings so check-theme survives a Windows checkout": based on
  `fix/desktop-windows-release`, currently `MERGEABLE` / `CLEAN` against #2150 as it stands.

I did not rebase, retarget or otherwise touch either.

---

## Skipped, and why

- **Any cargo/rustc invocation.** Machine constraint (swap at 15.1 GB of 16.4 GB). Covered instead by
  the green `Build SDK` job's `git diff --exit-code` on the regenerated artifacts, which is a
  stronger check than a local run.
- **Real Windows execution.** No Windows host. Covered by the green `package script on windows` job;
  I read its log rather than asserting from it. Its negative control is genuinely falsifiable:
  ```
  cmd.exe says:
  The syntax of the command is incorrect.
  confirmed: cmd.exe did not create the directories the POSIX form asks for
  …
  ok: npm run generate:openapi created both directories on windows
  ```
  and the assertion step additionally checks `anyharness/sdk/-p` was not created, which anticipates
  the other way cmd's `md` could mangle the POSIX form. That is a well-built guard.
- **`CARGO` pointing at a `.cmd`/`.bat` wrapper on Windows** (would hit the CVE-2024-27980 spawn
  refusal). Cannot exercise without a Windows host, and no configuration in-tree does this.

---

**Verdict: non-blocking.** The hard question — whether the ported `SKIP_RUST` branch is faithful —
comes back clean across all thirteen input cases, the `#2134`-class failure swallow is absent, and
output integrity is strictly better than what it replaces. F1 is the only finding I would push for
before merge, and it is a test-coverage gap rather than a defect. Not approving; merge is Pablo's.

PR #2147, opened 2026-08-21
fix(anyharness): terminate agent process trees on windows

pablonyx:
## The defect

`anyharness/crates/anyharness-lib/src/process_kill.rs:352-360` (pre-change) gated the entire real implementation behind `#[cfg(unix)] mod unix_impl` and handed every other platform this:

```rust
#[cfg(not(unix))]
pub async fn kill_group_and_await(_pgid: i32) -> (usize, usize) { (0, 0) }
#[cfg(not(unix))]
pub async fn kill_session_and_await(_sid: i32) -> (usize, usize) { (0, 0) }
```

`(0, 0)` is not an error value. It is the exact tuple a *successful* kill of zero processes returns, and the unix path returns it deliberately at `process_kill.rs:284` and `:314` to mean "the target was already empty". Every caller reads it that way. `SessionActor::kill_process_group_and_reap` (`live/sessions/actor/run.rs:117`) sends it back as the stop response. `close_all_for_workspace` and `kill_active_run_for_workspace` (`live/terminals/command_runs/workspace_stop.rs:113`, `:168`) fold it into `PlaneKills`. R2's `repair_kill_debris` aborts a conflict sentinel only when `killed_git > 0`, so a stub zero is also a positive claim that no `git` was running.

The result on Windows was not a missing feature but a lie: stop and cancel reported clean quiescence, archive believed the session had quiesced, and the whole agent process tree kept running with nothing left that could reach it. Orphaned trees accumulate for the lifetime of the machine.

## What the callers can actually provide

I traced every call site and every value that reaches these functions:

| Call site | Value passed | Produced by |
| --- | --- | --- |
| `live/sessions/actor/run.rs:130` | `self.child.id() as i32` | `spawn_agent_process` (`live/sessions/driver/process.rs:109`) |
| `live/terminals/command_runs/workspace_stop.rs:110` | `handle.child.process_id() as i32` | portable-pty's PTY child |
| `live/terminals/command_runs/workspace_stop.rs:167` and `:466` | `ActiveSetupTask::pgid` (an `AtomicI32`) | `run_setup_process` stores `child.id()` (`setup_process.rs:118`) |
| `live/terminals/command_runs/setup_process.rs:210` | the same atomic | same |

On unix every one of those integers is a process *group* id only because the spawn site calls `process_group(0)`, which makes the child its own group leader so the pid and the pgid are the same number. Both of those calls (`driver/process.rs:109`, `setup_process.rs:76`) are `#[cfg(unix)]`, and portable-pty's `setsid()` is in its unix backend only. So on Windows the integer that arrives is unambiguously a plain pid, and nothing more is available today.

## Approach: Toolhelp descendant walk. Job Objects rejected.

**Chosen:** a pure-Rust `CreateToolhelp32Snapshot(TH32CS_SNAPPROCESS)` enumeration that walks `th32ParentProcessID` to build the descendant tree rooted at the pid the caller already passes. New file `anyharness/crates/anyharness-lib/src/process_kill_windows.rs`.

This is the option that actually matches the existing structure. The unix path is two-phase by design: enumerate the whole target, census it *before* signaling (`census()` resolves each pid's executable first, because enumerating after the TERM undercounts exactly the processes that died fastest), then act. A Toolhelp snapshot returns pid, parent pid, and base executable name in one read, so the Windows census is the same two-phase shape and is actually cheaper than the unix one, which needs a `proc_pidpath`/`readlink` per pid. It also needs zero spawn-site changes, which means it covers all four call sites uniformly.

**Rejected: Job Objects.** `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` plus `TerminateJobObject` is the correct Windows primitive in the abstract: atomic, and it cannot leak. It is not reachable here.

- ~~It requires touching the spawn site, and one of the three spawn sites is inside portable-pty's ConPTY backend. There is no hook to assign the PTY child to a job without forking the crate.~~ **This was wrong and review caught it.** portable-pty 0.9.0 exposes `as_raw_handle()` on the public `Child` trait (`lib.rs:141-145`, `win/mod.rs:118`, documented "Only available on Windows"), and `AssignProcessToJobObject` needs only a handle, not a spawn hook. The PTY child is reachable. The conclusion below still holds on the two remaining grounds, but not on this one.
- It changes the interface. A job handle is not an `i32`, so `ActiveSetupTask::pgid: Arc<AtomicI32>` and the `(i32) -> (usize, usize)` signature every caller uses would all have to change. That is a restructure of the spawn path across three modules, and `setup_process.rs` is currently owned by a parallel `/bin/sh` portability change.
- It cannot produce the `(total, git)` census. `TerminateJobObject` kills the set without telling you what was in it, so the `killed_git` evidence `repair_kill_debris` depends on would have to be enumerated separately anyway, which is the Toolhelp walk again.

Job Objects remain the better long-term answer, and now for all three spawn sites rather than two, since the PTY child turns out to be reachable after all. They compose with this change rather than replacing it: assigning the children to kill-on-close jobs would make the tree walk a backstop instead of the only mechanism, and would remove the pid-identity problem at the root rather than defending against it. That is a follow-up, not this PR.

**Rejected: `taskkill /PID <pid> /T /F`.** The module's own doc comment already rules out the equivalent move on unix ("an implementer reaching for a shelled-out `ps` here ships a fragile, slow kill"). It also spawns a process in order to kill processes, gives no census at all, and its `/T` walk is the same parent-link walk with a process spawn and output parsing bolted on top.

## What the Windows path does

The RETURN contract is the same: same `(total, git)` tuple meaning, same `(0, 0)`-means-nothing-was-running, same detached-task escalation so a dropped or timed-out caller future can never strand a half-killed tree, same `GRACE` (5s) as an early-exiting deadline rather than a fixed cost, same `CONFIRM_POLL` (50ms) and `CONFIRM_BUDGET` (10s).

An earlier version of this body claimed the contract was identical full stop. That was wrong and review was right to reject it. The GUARANTEE is weaker in three ways, all now documented in the file and in the spec:

1. **No graceful rung.** `TerminateProcess` is unconditional, so it is the equivalent of `SIGKILL`, not `SIGTERM`. A graceful first rung would need `GenerateConsoleCtrlEvent`, which only works against a group created with `CREATE_NEW_PROCESS_GROUP` at spawn and only for console processes sharing our console. No spawn site sets that flag, so the ladder starts at the unconditional rung. `GRACE` is still honored, as the deadline for the OS to finish tearing the tree down before survivors are retried.
2. **The kill is not atomic against a growing tree, and unix's is.** `kill(-pgid)` signals a group, so a child created between the unix enumeration and the signal is already a member and still gets hit. Here the pass acts on a pid list read from a snapshot, so a descendant born after that snapshot is missed by *that* pass and picked up by the next confirmation pass, up to `CONFIRM_POLL` later. The census returned to the caller is the one taken before the first pass, so a tree that grew mid-kill is under-reported even though it is killed.
3. **A pid is not an identity, so an unprovable descendant is refused rather than killed.** Windows does not re-parent orphans: a dead process leaves its children's `th32ParentProcessID` pointing at a freed number the kernel will hand to someone else. Every pid acted on is checked against its creation time, and a descendant whose identity cannot be read is left alone. Terminating a stranger is worse than leaking a descendant, so this fails toward under-killing. The runtime's own pid is excluded, and a root equal to our own pid is refused outright.

`git` detection compares against `git.exe` case-insensitively. The unix path's exact `"git"` comparison would have counted zero forever on Windows and silently changed archive's refusal behavior.

A failed snapshot is the one thing the `(usize, usize)` signature cannot report, so it is logged at ERROR and never folded into `(0, 0)`. The `#[cfg(not(any(unix, windows)))]` stub that remains still returns `(0, 0)` because no such platform is a build target, but it now logs at ERROR on every call so it can never be a silent lie again.

## Dependency

`windows-sys` is already in `Cargo.lock` at five versions; 0.61.2 is pulled by `mio`, `async-io`, `polling`, `signal-hook`, and others. This adds `windows-sys = "0.61"` with `Win32_Foundation`, `Win32_System_Diagnostics_ToolHelp`, and `Win32_System_Threading`, under `[target.'cfg(windows)'.dependencies]` so no other platform pays for it. It resolves to the 0.61.2 already in the lock, so the only lock change is one line added to the `anyharness-lib` dependency list. No new package.

## Review round 2: what changed

All four blocking findings are fixed, and the fix for B1 grew a fifth case the review did not name.

- **B1, stale parent links terminate unrelated processes.** Adoption now requires the child's creation time (`GetProcessTimes`) to be at or after the parent's. What that buys, precisely: for a link to be stale, the number's previous owner had to be alive when the stranger started and had to exit before our root could acquire the number, so in real time the stranger started before our root did, and a genuine child always starts after its parent. So the comparison separates the two cases whenever the clock separates them. What it does not buy: the comparison is on *reported* `GetProcessTimes` values and it admits equality, so at an equal reported timestamp a stranger is adopted. Equality has to be admitted, or a genuine child created inside the same tick as its parent leaks, so the same-tick window is the deliberate residual rather than a bug. How wide it is depends on the sampling resolution `GetProcessTimes` actually reports on a given machine, which nothing here has measured; it is not safe to assume the 100ns the FILETIME encoding suggests. The check also does not protect the root itself, only what is adopted beneath it. The reviewer's repro (root 8432, `TerminateProcess` on `[8432, 5000, 5001, 5002]`, census `(4, 1)`) is now a test that asserts `[8432]` and `(1, 0)`, and deleting the comparison restores exactly the reviewer's numbers.
- **B1, second hole, found while fixing the first.** `may_adopt` skips pids that are already tracked, so a pid recycled *while tracked* (dies and is reassigned between two passes, so we never observe it missing) was never re-checked by anything. The `retired` set could not help; nor could the generation check, because an already-tracked pid is not an adoption candidate. Every pass now re-verifies the identity of the pids it holds and drops any whose creation time has changed. That **subsumes** `retired`, which is removed: with identity checked, the only thing `retired` still did was refuse a genuine child that happened to reuse a number we had seen die, which is a leak.
- **B3, grandchild lost inside the kill window.** Growth now runs before the vanished are dropped, so an orphan is adopted through its parent's last known membership.
- **B2, `Process32FirstW` failure read as an empty table.** Now returns `Err(GetLastError())`.
- **B4, no retry on a transient enumeration failure.** Five attempts, 10ms apart, bounded and only on the failure path, with `GetLastError` read and `ERROR_BAD_LENGTH` reported as retryable in the log.

**The uncompiled unit tests are now compiled and running, on Linux, on every PR.** Review was right that this needed no Windows at all. The bookkeeping is pure `std`, so it moved to `process_kill_tree.rs` under `#[cfg(any(windows, test))]`, and its thirteen tests run under the existing `cargo test --workspace`. This is the coverage that matters most, because all four blocking findings live in the multi-pass escalation rung and both real-process Windows tests die on the first `TerminateProcess` pass, so no amount of green Windows CI was ever going to reach them.

**Both vacuous tests are rewritten with real negative controls**, and every test in the new file states the mechanism it pins and what deleting that mechanism does to it. The cycle test now builds a cycle that actually enters the tracked set, and `depth_of` distinguishes a cycle (`None`) from a truncated deep chain (`Some(DEPTH_LIMIT)`), so deleting the visited set returns `Some(64)` and the test goes red rather than passing. The pid-reuse test is replaced by the identity-verification test above, which goes red by terminating a stranger.

**One correction back.** The heads-up about `lifecycle_tests.rs:113` and `:182` does not hold. Those assertions are driven by a *scripted* live handle, not a real process: the file's own module doc says it "does not need a real spawned process", and the census comes from `ScriptedSessionSpec`, never from `process_kill`. They are also not unix-gated (`runtime/mod.rs:64-65` is a bare `#[cfg(test)]`). So they will neither fail on Windows for the Git-for-Windows shim reason nor need touching during the `lib test` port. Flagging so nobody spends time on it.

## Verified vs unverified

**Verified:**

- **Windows compiles.** Literal result from `.github/workflows/windows-compile-check.yml` on [run 32450924720](https://github.com/proliferate-ai/proliferate/actions/runs/32450924720), at head `6c48750e186ba14fb8daf020f74bbda29b674720`:

  ```
  runtime crates (control):   success   (cargo check --target x86_64-pc-windows-msvc
                                         -p anyharness -p proliferate-worker -p proliferate-supervisor)
      Finished `dev` profile [unoptimized + debuginfo] target(s) in 3m 00s
  desktop crates (the unknown): success
  ```

  Zero errors. `anyharness-lib`'s warnings are all pre-existing unix-side dead code that the Windows cfg leaves unreferenced; grepping the full job log for `process_kill` returns nothing, so the new files and the new `windows-sys` dependency compile warning-free.
- **The tree bookkeeping is unit-tested on every PR, on Linux.** Thirteen tests under `process_kill::tree::tests`, all green in `Cargo check & test` at head `6c48750e186ba14fb8daf020f74bbda29b674720` ([job 96679172461](https://github.com/proliferate-ai/proliferate/actions/runs/32450924690/job/96679172461)), covering stale-parent-link refusal, recycled-pid identity, unreadable-identity fail-open, grandchild adoption across a parent's death, the full multi-pass escalation sequence, cycle detection, self-pid exclusion, fail-closed behaviour with an unknown root identity, and the `git.exe` census. Literal lines from that job, by name, all thirteen: `absorb_adopts_a_grandchild_whose_parent_died_in_the_same_interval`, `absorb_adopts_nothing_when_the_roots_identity_is_unknown`, `absorb_collects_the_whole_descendant_tree_root_first`, `absorb_drops_a_tracked_pid_that_was_recycled_between_passes`, `absorb_keeps_a_tracked_pid_whose_identity_cannot_be_read`, `absorb_never_adopts_the_runtimes_own_process`, `absorb_refuses_a_stranger_reached_through_a_stale_parent_link`, `absorb_still_adopts_a_genuine_child_created_after_its_parent`, `depth_of_measures_an_ordinary_chain`, `depth_of_reports_a_cycle_rather_than_walking_it`, `executable_name_stops_at_the_nul_terminator`, `the_census_counts_git_exe_where_an_exact_git_comparison_would_count_zero`, `the_escalation_rung_tracks_a_tree_across_passes_until_it_empties`, each logged as `PASS`. The job's full-suite summary line: `3160 tests run: 3160 passed (1 slow), 8 skipped`.
- **The `unsafe` FFI has been EXECUTED on Windows against real process trees.** A dedicated job (`.github/workflows/windows-process-kill-tests.yml`, kept separate so the shared `windows-compile-check.yml` other branches stack on is not touched) runs `cargo test --target x86_64-pc-windows-msvc -p anyharness-lib --test windows_process_tree` on `windows-latest`. Literal result at head `6c48750e186ba14fb8daf020f74bbda29b674720` ([job 96679171510](https://github.com/proliferate-ai/proliferate/actions/runs/32450924588/job/96679171510)):

  ```
  running 2 tests
  test kills_a_real_process_tree_including_the_grandchild ... ok
  test the_census_counts_a_real_git_exe_grandchild ... ok
  test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.20s
  ```

  That covers `CreateToolhelp32Snapshot`, the `PROCESSENTRY32W` walk and its `dwSize` handshake, `TreeTracker`, the census, `OpenProcess`/`TerminateProcess`, and the confirmation loop, against the real kernel. Every assertion uses an oracle independent of the code under test: `Child::wait` for the direct child, PowerShell's CIM process table for descendants.

- **The runner found a real defect that reading could not.** First run was red: `assertion left == right failed ... left: 2, right: 1`. Git for Windows resolves `git` to a shim in `cmd\git.exe` that launches the real `git.exe` from `mingw64\bin`, so one `git` invocation is *two* `git.exe` processes and the tree is three deep. The census counted both correctly; my test had asserted exactly one. The fix was to the test, not the implementation: it now asserts the load-bearing property (the count is nonzero, which is exactly what `repair_kill_debris` branches on) plus an oracle-backed negative control that every git process is named `git.exe`, so the unix path's exact `"git"` comparison would have counted zero.
- The unix side is unaffected: `Cargo check & test` passed on this PR, and `process_kill_tests.rs` is unmodified. Mechanically, every line of CODE changed in `process_kill.rs` sits inside `#[cfg(windows)]`, `#[cfg(any(windows, test))]`, or `#[cfg(not(any(unix, windows)))]` (the module-level `//!` doc lines are in no cfg and are unix-visible, but are behaviourally inert; a broken intra-doc link introduced there has been removed); nothing inside `#[cfg(unix)] mod unix_impl` or its `pub use` changed. `process_kill_windows.rs` is reachable only through `#[cfg(windows)] mod windows_impl`, so it is not compiled on unix at all, and `tests/windows_process_tree.rs` is `#![cfg(windows)]`, so it is an empty crate there. `process_kill_tree.rs` is `#[cfg(any(windows, test))]`, so on unix it exists only under `cargo test`, which is the point of it.
- `rustfmt --edition 2021 --check` clean on both changed Rust files.
- The caller/spawn-site trace in the table above, read from source.

**Not verified.** The core kill path HAS been executed on Windows, and that includes more than the FFI calls: the confirmation polls re-enumerate and re-`absorb` on every 50ms tick, so `absorb`, the per-pass identity re-verification, and the `GetProcessTimes` generation comparison all execute against the real kernel on every Windows run, not only against the Linux fakes. The one part of the Windows path that no test reaches is the post-`GRACE` escalation loop at `process_kill_windows.rs:296-327`. What follows is what has not been verified.

- **The FFI functions themselves have no unit tests**, only the real-process integration tests. `snapshot_processes`, `process_creation_time`, `terminate`, and the retry loop are exercised end to end or not at all; the B2 and B4 guards specifically have no test, because faking a `CreateToolhelp32Snapshot` failure needs injection this does not have.
- **`anyharness-lib`'s `lib test` target still does not build for `x86_64-pc-windows-msvc`**, so the new unit tests run on Linux and macOS but not on Windows. Ten pre-existing errors, every one a test-only POSIX assumption in an unrelated module. Not absorbed here. The logic is platform-independent, so this costs coverage of the *target* platform rather than of the logic.
- **Timing under load is untested.** The two integration tests finished in 3.93s, so the happy path clearly resolves well inside `GRACE`. What is unverified is the escalation: no test has produced a tree that survives the first `TerminateProcess` pass, so the re-enumerate-and-re-terminate loop at `process_kill_windows.rs:296-327` and the `CONFIRM_BUDGET` exhaustion warning have not executed. That loop is the only Windows-path code no test reaches.
- **Only a two-and-three-process tree has been killed for real.** Nothing has exercised a deep or wide agent tree, a ConPTY session, or a real tree spawning new descendants during the kill. The tracker logic that handles those does run against the real kernel through the confirmation polls, but always on a tree that dies on the first pass, so what is untested is the escalation rung specifically rather than the bookkeeping. I did not add a growing-tree integration test because there is no deterministic way to make a real tree survive the first pass, and a flaky test asserting the escalation rung is worse than an honest gap.
- **`kill_active_run_for_workspace` (`workspace_stop.rs:163`) still aborts the owning task before killing**, which drops the `Child` and unreserves the pid, handing this code a possibly-recycled root. The other two call sites correctly hold the handle across the kill. I did not reorder it: aborting after the kill lets the task observe its child dying and overwrite the "Run interrupted by workspace stop" status that was just written, so this is a real trade between a certain status regression and a rare pid race, on a path with existing unix test coverage. The race is Windows-only: on unix the un-reaped zombie reserves the pid, so nothing can recycle it. It deserves its own PR with the unix suite as arbiter, not a drive-by inside a Windows change.
- **The spec paragraph added to `specs/anyharness/live-runtime.md`** describes the design; it is not a verification of it.

The honest summary: this is a real implementation, not a stub. Its core path has been executed on Windows against real process trees including a real `git.exe`, and its escalation and identity logic is unit-tested on every PR on Linux. What remains untested is the escalation rung, which no real tree has been made to reach.

## Follow-ups

1. Unblock `anyharness-lib`'s `lib test` target on Windows by cfg-gating the ten pre-existing test-only POSIX assumptions, then add `--lib` to the new job so the tree tests run on the target platform too.
2. Reorder `kill_active_run_for_workspace` to hold the `Child` across the kill, resolving the status-overwrite trade above properly.
3. A real-kernel test that forces the escalation rung, if a deterministic way to make a tree survive the first pass can be found.
4. Assign all three spawn sites to kill-on-close Job Objects via `as_raw_handle()`, so a runtime crash cannot leak a tree at all, the walk becomes a backstop, and the pid-identity problem disappears rather than being defended against. Needs the `AtomicI32` pgid slot in `ActiveSetupTask` to carry a handle instead.

## Stacking

Based on `ci/windows-compile-check` (draft PR #2142), not `main`, so the `pull_request` path trigger on `anyharness/crates/**` fires and this gets a genuine Windows compile verdict. That job is `continue-on-error: true` by design, so read the job conclusion, not the check mark. Retarget to `main` once #2142 lands.

### DIFF
--- a/.github/workflows/windows-compile-check.yml
+++ b/.github/workflows/windows-compile-check.yml
@@ -1,0 +1,69 @@
+name: Windows compile check
+
+# The runtime already builds and ships for Windows on every release, but nothing
+# has compiled the desktop crate for Windows since desktop releases were
+# disabled on 2026-04-13. That makes every Windows estimate guesswork. This job
+# turns the unknown into an error list.
+#
+# Non-blocking on purpose: it is a probe, not a gate, until the errors are fixed.
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "apps/desktop/src-tauri/**"
+      - "anyharness/crates/**"
+      - ".github/workflows/windows-compile-check.yml"
+
+permissions:
+  contents: read
+
+jobs:
+  runtime:
+    # Known green: release-runtime.yml ships these targets today. Here as the
+    # control, so a red desktop job cannot be blamed on the toolchain.
+    name: runtime crates (control)
+    runs-on: windows-latest
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: x86_64-pc-windows-msvc
+      - uses: Swatinem/rust-cache@v2
+      - name: cargo check
+        run: cargo check --target x86_64-pc-windows-msvc -p anyharness -p proliferate-worker -p proliferate-supervisor
+
+  desktop:
+    name: desktop crates (the unknown)
+    runs-on: windows-latest
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: x86_64-pc-windows-msvc
+      - uses: Swatinem/rust-cache@v2
+      - name: cargo check
+        # Keep going past the first failure so one run yields the whole list
+        # rather than one error at a time.
+        run: cargo check --target x86_64-pc-windows-msvc -p proliferate --keep-going 2>&1 | tee check.log
+        shell: bash
+      - name: Error summary
+        if: always()
+        shell: bash
+        run: |
+          echo "### Windows desktop compile errors" >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
+          grep -E '^error' check.log | sort | uniq -c | sort -rn | head -50 >> $GITHUB_STEP_SUMMARY || echo "no errors" >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
+          echo "total errors: $(grep -cE '^error' check.log || echo 0)" >> $GITHUB_STEP_SUMMARY
+          echo "files implicated:" >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
+          grep -oE '\-\-> [^:]+' check.log | sed 's/--> //' | sort -u | head -60 >> $GITHUB_STEP_SUMMARY || true
+          echo '```' >> $GITHUB_STEP_SUMMARY
+      - uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: windows-desktop-check-log
+          path: check.log
--- a/.github/workflows/windows-compile-check.yml
+++ b/.github/workflows/windows-compile-check.yml
@@ -14,6 +14,11 @@ on:
       - "apps/desktop/src-tauri/**"
       - "anyharness/crates/**"
       - ".github/workflows/windows-compile-check.yml"
+      - "Cargo.lock"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 permissions:
   contents: read
@@ -24,9 +29,10 @@ jobs:
     # control, so a red desktop job cannot be blamed on the toolchain.
     name: runtime crates (control)
     runs-on: windows-latest
+    timeout-minutes: 45
     continue-on-error: true
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: x86_64-pc-windows-msvc
@@ -37,9 +43,10 @@ jobs:
   desktop:
     name: desktop crates (the unknown)
     runs-on: windows-latest
+    timeout-minutes: 45
     continue-on-error: true
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: x86_64-pc-windows-msvc
@@ -62,7 +69,7 @@ jobs:
           echo '```' >> $GITHUB_STEP_SUMMARY
           grep -oE '\-\-> [^:]+' check.log | sed 's/--> //' | sort -u | head -60 >> $GITHUB_STEP_SUMMARY || true
           echo '```' >> $GITHUB_STEP_SUMMARY
-      - uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         if: always()
         with:
           name: windows-desktop-check-log
--- a/.github/workflows/windows-process-kill-tests.yml
+++ b/.github/workflows/windows-process-kill-tests.yml
@@ -21,6 +21,8 @@ on:
   pull_request:
     paths:
       - "anyharness/crates/anyharness-lib/src/process_kill.rs"
+      - "anyharness/crates/anyharness-lib/src/process_kill_tree.rs"
+      - "anyharness/crates/anyharness-lib/src/process_kill_tree_tests.rs"
       - "anyharness/crates/anyharness-lib/src/process_kill_windows.rs"
       - "anyharness/crates/anyharness-lib/tests/windows_process_tree.rs"
       - ".github/workflows/windows-process-kill-tests.yml"
--- a/.github/workflows/windows-process-kill-tests.yml
+++ b/.github/workflows/windows-process-kill-tests.yml
@@ -1,0 +1,48 @@
+name: Windows process-kill tests
+
+# `windows-compile-check.yml` proves the Windows code TYPECHECKS. Nothing
+# proves it RUNS, and `process_kill_windows.rs` is `unsafe` FFI against
+# Toolhelp and TerminateProcess, which is exactly the kind of code where
+# typechecking establishes very little.
+#
+# This deliberately runs ONE integration test target rather than the crate's
+# unit tests. `anyharness-lib`'s `lib test` target does not build for Windows
+# today: `cargo check --all-targets --target x86_64-pc-windows-msvc` reports
+# ten pre-existing errors, all test-only POSIX assumptions in unrelated
+# modules. An integration test compiles as its own crate against the library's
+# public surface, so it never pulls those in and gives a real runtime verdict
+# without absorbing that port into this change.
+#
+# Kept separate from `windows-compile-check.yml` on purpose: other branches
+# stack on that file and it must not be destabilised.
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "anyharness/crates/anyharness-lib/src/process_kill.rs"
+      - "anyharness/crates/anyharness-lib/src/process_kill_windows.rs"
+      - "anyharness/crates/anyharness-lib/tests/windows_process_tree.rs"
+      - ".github/workflows/windows-process-kill-tests.yml"
+
+permissions:
+  contents: read
+
+jobs:
+  process-kill:
+    name: process-kill tests (windows)
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: x86_64-pc-windows-msvc
+      - uses: Swatinem/rust-cache@v2
+      - name: cargo test
+        # Serial: each test spawns a real process tree and asserts against the
+        # machine-wide process table, so parallel runs would read each other's
+        # children.
+        run: >
+          cargo test --target x86_64-pc-windows-msvc
+          -p anyharness-lib --test windows_process_tree
+          -- --nocapture --test-threads=1
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -418,6 +418,7 @@ dependencies = [
  "utoipa",
  "utoipa-axum",
  "uuid",
+ "windows-sys 0.61.2",
  "zip 8.6.0",
  "zstd",
 ]
--- a/anyharness/crates/anyharness-lib/Cargo.toml
+++ b/anyharness/crates/anyharness-lib/Cargo.toml
@@ -49,6 +49,18 @@ minisign-verify.workspace = true
 [target.'cfg(unix)'.dependencies]
 libc.workspace = true
 
+# Toolhelp process enumeration plus OpenProcess/TerminateProcess for
+# `process_kill_windows.rs`. Target-gated so no other platform pays for it,
+# and pinned to the 0.61 line the workspace already resolves (mio, async-io,
+# polling and others pull windows-sys 0.61.2 today) so this adds no new
+# package to the lock.
+[target.'cfg(windows)'.dependencies]
+windows-sys = { version = "0.61", features = [
+    "Win32_Foundation",
+    "Win32_System_Diagnostics_ToolHelp",
+    "Win32_System_Threading",
+] }
+
 [dev-dependencies]
 tokio = { workspace = true, features = ["test-util"] }
 tracing-subscriber.workspace = true
--- a/anyharness/crates/anyharness-lib/src/process_kill.rs
+++ b/anyharness/crates/anyharness-lib/src/process_kill.rs
@@ -11,11 +11,22 @@
 //! `DOMAIN_LIVE_VALVE` boundary rule. See the R3 delivery spec's
 //! Contradictions C4.
 //!
-//! Everything here is unix-only: the enumeration is genuinely two
-//! implementations (macOS via libproc, Linux via `/proc`) behind one
-//! `#[cfg]`, and the escalation always runs on a detached task so a dropped
-//! or timed-out caller future can never strand a TERM-ignoring process
-//! between the TERM and the KILL.
+//! The enumeration is genuinely one implementation per platform - macOS via
+//! libproc, Linux via `/proc`, Windows via a `CreateToolhelp32Snapshot`
+//! parent-link walk in [`process_kill_windows`](crate::process_kill_windows) -
+//! behind one `#[cfg]`, and on every platform the escalation runs on a
+//! detached task so a dropped or timed-out caller future can never strand a
+//! half-killed target between the first signal and the last.
+//!
+//! What a caller passes differs by platform and the two functions below name
+//! the unix concepts because that is where the distinction is real. Windows
+//! has neither process groups nor sessions, and every call site derives its
+//! integer from a direct child's pid (`process_group(0)` at the spawn sites is
+//! `#[cfg(unix)]`), so both entry points collapse to one descendant-tree kill
+//! rooted at that pid. The CONTRACT is identical on both platforms: the
+//! returned `(total, git)` is the census taken before anything is signaled,
+//! `(0, 0)` means nothing was running, and the call resolves only once the
+//! target is confirmed dead or the confirmation budget is spent.
 
 use std::time::Duration;
 
@@ -349,13 +360,52 @@ mod unix_impl {
 #[cfg(unix)]
 pub use unix_impl::{kill_group_and_await, kill_session_and_await};
 
-#[cfg(not(unix))]
-pub async fn kill_group_and_await(_pgid: i32) -> (usize, usize) {
+#[cfg(windows)]
+#[path = "process_kill_windows.rs"]
+mod windows_impl;
+
+/// Windows: the pgid a caller holds is only ever the direct child's pid (see
+/// this module's header), so this kills the descendant tree rooted at it.
+/// Same `(total, git)` census, same detached escalation, same deadline
+/// semantics as the unix path; the one behavioral difference is that
+/// `TerminateProcess` is unconditional, so there is no graceful rung before
+/// it (`process_kill_windows.rs` explains why one is not reachable from the
+/// current spawn sites).
+#[cfg(windows)]
+pub async fn kill_group_and_await(pgid: i32) -> (usize, usize) {
+    windows_impl::kill_tree_and_await(pgid).await
+}
+
+/// Windows: PTY sessions do not exist as a kernel concept, and portable-pty's
+/// ConPTY backend has no `setsid()` to make the child a session leader, so
+/// `sid` here is just the PTY child's pid and the shell's jobs are its
+/// ordinary descendants. Same tree kill as [`kill_group_and_await`].
+#[cfg(windows)]
+pub async fn kill_session_and_await(sid: i32) -> (usize, usize) {
+    windows_impl::kill_tree_and_await(sid).await
+}
+
+/// No other platform is a build target for this runtime. Returning the bare
+/// `(0, 0)` here would be indistinguishable from a successful kill of nothing,
+/// so the stub is loud instead of silent: a caller that believes a live plane
+/// has quiesced when nothing was killed is worse than an error, because
+/// archive then proceeds against a workspace whose processes are still
+/// holding it.
+#[cfg(not(any(unix, windows)))]
+pub async fn kill_group_and_await(pgid: i32) -> (usize, usize) {
+    tracing::error!(
+        pgid,
+        "process_kill: no process-kill implementation for this platform; NOTHING was killed"
+    );
     (0, 0)
 }
 
-#[cfg(not(unix))]
-pub async fn kill_session_and_await(_sid: i32) -> (usize, usize) {
+#[cfg(not(any(unix, windows)))]
+pub async fn kill_session_and_await(sid: i32) -> (usize, usize) {
+    tracing::error!(
+        sid,
+        "process_kill: no process-kill implementation for this platform; NOTHING was killed"
+    );
     (0, 0)
 }
 
--- a/anyharness/crates/anyharness-lib/src/process_kill.rs
+++ b/anyharness/crates/anyharness-lib/src/process_kill.rs
@@ -13,10 +13,10 @@
 //!
 //! The enumeration is genuinely one implementation per platform - macOS via
 //! libproc, Linux via `/proc`, Windows via a `CreateToolhelp32Snapshot`
-//! parent-link walk in [`process_kill_windows`](crate::process_kill_windows) -
-//! behind one `#[cfg]`, and on every platform the escalation runs on a
-//! detached task so a dropped or timed-out caller future can never strand a
-//! half-killed target between the first signal and the last.
+//! parent-link walk in `process_kill_windows.rs` - behind one `#[cfg]`, and
+//! on every platform the escalation runs on a detached task so a dropped or
+//! timed-out caller future can never strand a half-killed target between the
+//! first signal and the last.
 //!
 //! What a caller passes differs by platform and the two functions below name
 //! the unix concepts because that is where the distinction is real. Windows
@@ -26,7 +26,11 @@
 //! rooted at that pid. The CONTRACT is identical on both platforms: the
 //! returned `(total, git)` is the census taken before anything is signaled,
 //! `(0, 0)` means nothing was running, and the call resolves only once the
-//! target is confirmed dead or the confirmation budget is spent.
+//! target is confirmed dead or the confirmation budget is spent. Two places
+//! where Windows is genuinely weaker, both documented in
+//! `process_kill_windows.rs`: the kill is not atomic against a tree that
+//! grows mid-kill (unix's `kill(-pgid)` is), and an adoption whose identity
+//! cannot be proven is refused rather than guessed at.
 
 use std::time::Duration;
 
@@ -359,6 +363,15 @@ mod unix_impl {
 
 #[cfg(unix)]
 pub use unix_impl::{kill_group_and_await, kill_session_and_await};
+
+/// The tree bookkeeping the Windows kill path runs on. It contains no FFI, so
+/// it is compiled under `test` on every platform and its unit tests run on the
+/// ordinary Linux and macOS jobs - which is where the multi-pass escalation
+/// logic actually gets exercised, since a real-process Windows test never
+/// reaches the second pass.
+#[cfg(any(windows, test))]
+#[path = "process_kill_tree.rs"]
+mod tree;
 
 #[cfg(windows)]
 #[path = "process_kill_windows.rs"]
--- a/anyharness/crates/anyharness-lib/src/process_kill_tree.rs
+++ b/anyharness/crates/anyharness-lib/src/process_kill_tree.rs
@@ -1,0 +1,244 @@
+//! Process-tree bookkeeping for the Windows kill path, with no FFI in it.
+//!
+//! Split out from `process_kill_windows.rs` deliberately. Everything here is
+//! `std::collections`, `String` and a `[u16; 260]` decode, so gating it
+//! `#[cfg(any(windows, test))]` lets the whole of it compile and RUN on the
+//! ordinary Linux and macOS test jobs, on every PR. That matters more than it
+//! looks: the two bugs this module exists to prevent (adopting a stranger
+//! through a stale parent link, and stranding a grandchild whose parent died
+//! mid-kill) both live in the multi-pass escalation rung, which a real-process
+//! Windows test never reaches because its tree dies on the first pass. Pure
+//! logic tested everywhere beats FFI logic tested nowhere.
+//!
+//! `process_kill_windows.rs` supplies the two things that genuinely need
+//! Windows - the Toolhelp enumeration and the creation-time read - and hands
+//! them in.
+
+use std::collections::{HashMap, HashSet};
+
+/// Cap on the parent-chain walk that orders a pass deepest-first. A chain
+/// longer than this is truncated for ordering purposes; a chain that CYCLES
+/// is reported separately (see [`TreeTracker::depth_of`]).
+const DEPTH_LIMIT: u32 = 64;
+
+/// A process's creation time in FILETIME ticks. This is the generation
+/// counter a bare pid does not have: two processes can share a pid over time
+/// but never a (pid, creation time) pair.
+pub(super) type Generation = u64;
+
+/// One row of a Toolhelp process snapshot. The base executable name comes
+/// back in the same read as the parent link, so unlike the unix path the
+/// `git` census costs no extra per-pid syscall.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(super) struct ProcessEntry {
+    pub(super) pid: u32,
+    pub(super) parent: u32,
+    pub(super) exe: String,
+}
+
+/// `szExeFile` is a fixed 260-wide buffer holding a NUL-terminated base name;
+/// decoding the whole array would append the padding.
+pub(super) fn executable_name(raw: &[u16; 260]) -> String {
+    let end = raw.iter().position(|&unit| unit == 0).unwrap_or(raw.len());
+    String::from_utf16_lossy(&raw[..end])
+}
+
+/// Whether a snapshot's base name is `git`. Windows file names are
+/// case-insensitive and the executable carries an extension, so the unix
+/// path's exact `"git"` comparison would count zero every time and quietly
+/// change `repair_kill_debris`'s refusal behavior.
+pub(super) fn is_git_executable(name: &str) -> bool {
+    let lowered = name.to_ascii_lowercase();
+    lowered == "git.exe" || lowered == "git"
+}
+
+/// The `(total, git)` census over an already-enumerated pass, taken BEFORE
+/// anything is signaled for the same reason the unix path does it: counting
+/// afterwards undercounts exactly the processes that died fastest.
+pub(super) fn census(entries: &[ProcessEntry]) -> (usize, usize) {
+    let git = entries
+        .iter()
+        .filter(|entry| is_git_executable(&entry.exe))
+        .count();
+    (entries.len(), git)
+}
+
+/// The set of pids belonging to one target tree, carried ACROSS enumeration
+/// passes and keyed on identity rather than on a bare pid.
+///
+/// Two hazards this exists to close, both of them ways a plain
+/// `th32ParentProcessID` walk gets the wrong answer:
+///
+/// **Stale parent links.** A pid is not an identity. A long-lived process
+/// whose creator exited keeps pointing at that freed number, and when Windows
+/// hands the number to our agent child the stranger looks like a descendant
+/// and gets terminated. Adoption therefore requires the child's creation time
+/// to be at or after the parent's. That test is exactly right, not merely
+/// conservative: a genuine child is always created after its parent, and a
+/// stranger's dangling link can only name a pid whose previous owner was
+/// alive when the stranger started and had to exit before our root could
+/// acquire the number, so the stranger is always OLDER than our root. An
+/// adoption whose generations cannot be read is refused, because terminating
+/// a stranger is worse than leaking a descendant.
+///
+/// A pid can also be recycled while it is ALREADY tracked, if it dies and is
+/// reassigned between two passes so we never observe it missing. Adoption
+/// checks cannot help there, because an already-tracked pid is never a
+/// candidate for adoption. Every pass therefore re-verifies the identity of
+/// the pids it already holds and drops any whose creation time has changed.
+/// That subsumes the weaker "never re-adopt a pid we have seen die" rule this
+/// started with, which was redundant once identity is checked and which
+/// leaked a genuine child that happened to reuse a retired number.
+///
+/// **Orphans are not re-parented.** Re-deriving the tree from the root each
+/// pass would lose every survivor the moment the root died, so the set is
+/// remembered. Growth runs BEFORE the vanished are dropped, so a grandchild
+/// whose parent died inside that same interval is still adopted through the
+/// parent's last known membership; dropping first would strand it forever,
+/// which is this module's own defect narrowed to a window.
+pub(super) struct TreeTracker {
+    tracked: HashMap<u32, Option<Generation>>,
+    self_pid: u32,
+}
+
+impl TreeTracker {
+    pub(super) fn new(root: u32, root_generation: Option<Generation>, self_pid: u32) -> Self {
+        let mut tracked = HashMap::new();
+        tracked.insert(root, root_generation);
+        Self { tracked, self_pid }
+    }
+
+    /// Adopt any provable new descendant, retire what vanished, and return
+    /// the tracked processes this snapshot still lists, ordered root-first.
+    ///
+    /// `created_at` resolves a pid's creation time. It is a parameter rather
+    /// than a direct syscall so this whole function is testable off Windows,
+    /// and so the caller can memoize it: it costs an `OpenProcess` each.
+    pub(super) fn absorb(
+        &mut self,
+        snapshot: &[ProcessEntry],
+        created_at: &mut dyn FnMut(u32) -> Option<Generation>,
+    ) -> Vec<ProcessEntry> {
+        // VERIFY IDENTITY FIRST, so a number that changed hands between
+        // passes can neither be terminated nor used as a parent. A stored
+        // generation of `None` means the identity was never establishable, so
+        // there is nothing to compare against and nothing may be adopted
+        // through it anyway.
+        let live_ids: Vec<u32> = snapshot
+            .iter()
+            .map(|entry| entry.pid)
+            .filter(|pid| self.tracked.contains_key(pid))
+            .collect();
+        for pid in live_ids {
+            let Some(Some(stored)) = self.tracked.get(&pid).copied() else {
+                continue;
+            };
+            if created_at(pid) != Some(stored) {
+                self.tracked.remove(&pid);
+            }
+        }
+
+        // GROW SECOND. A parent that died during this interval is still
+        // tracked right now, and that is the only thing that lets its orphan
+        // be adopted at all.
+        loop {
+            let mut grew = false;
+            for entry in snapshot {
+                if !self.may_adopt(entry) {
+                    continue;
+                }
+                let Some(Some(parent_generation)) = self.tracked.get(&entry.parent).copied() else {
+                    // Either the parent is not ours, or its identity was
+                    // never established. Refuse rather than guess.
+                    continue;
+                };
+                let Some(child_generation) = created_at(entry.pid) else {
+                    continue;
+                };
+                if child_generation < parent_generation {
+                    // A stale link: this process predates the pid it claims
+                    // as its parent, so that number has been recycled since.
+                    continue;
+                }
+                self.tracked.insert(entry.pid, Some(child_generation));
+                grew = true;
+            }
+            if !grew {
+                break;
+            }
+        }
+
+        // DROP THE VANISHED LAST, after they have had their one chance to
+        // pass on their orphans.
+        let live: HashSet<u32> = snapshot.iter().map(|entry| entry.pid).collect();
+        let vanished: Vec<u32> = self
+            .tracked
+            .keys()
+            .copied()
+            .filter(|pid| !live.contains(pid))
+            .collect();
+        for pid in vanished {
+            self.tracked.remove(&pid);
+        }
+
+        let parents: HashMap<u32, u32> = snapshot
+            .iter()
+            .map(|entry| (entry.pid, entry.parent))
+            .collect();
+        let mut alive: Vec<ProcessEntry> = snapshot
+            .iter()
+            .filter(|entry| self.tracked.contains_key(&entry.pid))
+            .cloned()
+            .collect();
+        alive.sort_by_key(|entry| {
+            // An unresolvable chain sorts last, so it is terminated first.
+            let depth = self
+                .depth_of(entry.pid, &parents)
+                .map_or((1u8, u32::MAX), |depth| (0u8, depth));
+            (depth, entry.pid)
+        });
+        alive
+    }
+
+    /// Whether `entry` is even a candidate for adoption, before its identity
+    /// is checked.
+    fn may_adopt(&self, entry: &ProcessEntry) -> bool {
+        // pid 0 is the idle process and is every unparented row's "parent";
+        // letting it into the set would sweep the machine.
+        entry.pid != 0
+            && entry.pid != self.self_pid
+            && entry.parent != entry.pid
+            && !self.tracked.contains_key(&entry.pid)
+    }
+
+    /// Hops from `pid` up to the shallowest tracked ancestor, or `None` when
+    /// the chain CYCLES. Ordering is the only consumer, so a chain deeper
+    /// than [`DEPTH_LIMIT`] is truncated to it rather than treated as an
+    /// error. A cycle is reported distinctly because it is a corrupt reading
+    /// of the table, and because without the visited set the walk would not
+    /// terminate at all.
+    pub(super) fn depth_of(&self, pid: u32, parents: &HashMap<u32, u32>) -> Option<u32> {
+        let mut seen = HashSet::new();
+        seen.insert(pid);
+        let mut depth = 0;
+        let mut cursor = pid;
+        for _ in 0..DEPTH_LIMIT {
+            let Some(&parent) = parents.get(&cursor) else {
+                return Some(depth);
+            };
+            if !self.tracked.contains_key(&parent) {
+                return Some(depth);
+            }
+            if !seen.insert(parent) {
+                return None;
+            }
+            depth += 1;
+            cursor = parent;
+        }
+        Some(DEPTH_LIMIT)
+    }
+}
+
+#[cfg(test)]
+#[path = "process_kill_tree_tests.rs"]
+mod tests;
--- a/anyharness/crates/anyharness-lib/src/process_kill_tree.rs
+++ b/anyharness/crates/anyharness-lib/src/process_kill_tree.rs
@@ -73,19 +73,39 @@ pub(super) fn census(entries: &[ProcessEntry]) -> (usize, usize) {
 /// whose creator exited keeps pointing at that freed number, and when Windows
 /// hands the number to our agent child the stranger looks like a descendant
 /// and gets terminated. Adoption therefore requires the child's creation time
-/// to be at or after the parent's. That test is exactly right, not merely
-/// conservative: a genuine child is always created after its parent, and a
-/// stranger's dangling link can only name a pid whose previous owner was
-/// alive when the stranger started and had to exit before our root could
-/// acquire the number, so the stranger is always OLDER than our root. An
-/// adoption whose generations cannot be read is refused, because terminating
-/// a stranger is worse than leaking a descendant.
+/// to be at or after the parent's.
+///
+/// What that buys, precisely: for a link to be stale, the number's previous
+/// owner had to be alive when the stranger started and had to exit before our
+/// root could acquire the number, so in real time the stranger STARTED before
+/// our root did. A genuine child always starts after its parent. So the
+/// comparison separates the two cases whenever the clock separates them.
+///
+/// What it does not buy: the comparison is on REPORTED `GetProcessTimes`
+/// values, not on real time, and it admits equality. At an equal reported
+/// timestamp a stranger is adopted. Equality has to be admitted anyway, or a
+/// genuine child created inside the same tick as its parent leaks, so the
+/// code is deliberately this way and the residual is the same-tick window
+/// rather than a bug. How wide that window is depends on the sampling
+/// resolution `GetProcessTimes` actually reports on a given machine, which
+/// nothing here has measured; do not assume it is the 100ns the FILETIME
+/// encoding suggests. Job Objects (see the module header's follow-up note)
+/// remove the question rather than narrowing it.
+///
+/// An adoption whose generations cannot be read is refused outright, because
+/// terminating a stranger is worse than leaking a descendant.
 ///
 /// A pid can also be recycled while it is ALREADY tracked, if it dies and is
 /// reassigned between two passes so we never observe it missing. Adoption
 /// checks cannot help there, because an already-tracked pid is never a
 /// candidate for adoption. Every pass therefore re-verifies the identity of
-/// the pids it already holds and drops any whose creation time has changed.
+/// the pids it already holds and drops any whose creation time has CHANGED.
+/// A time it cannot read is a different thing from a time that differs: the
+/// pid keeps its place in the set, and only loses its ability to prove
+/// adoptions. Treating unreadable as changed would let one transient failure
+/// against the root drop it permanently and report a live root as dead, which
+/// is the same shape of bug as reading a failed enumeration as an empty
+/// process table.
 /// That subsumes the weaker "never re-adopt a pid we have seen die" rule this
 /// started with, which was redundant once identity is checked and which
 /// leaked a genuine child that happened to reuse a retired number.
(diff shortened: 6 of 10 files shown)


pablonyx:
## Review: not safe to merge as-is. Four small local fixes, then yes.

**The premise this review started from has moved, so I will state the correction first.** I was briefed that this was unexecuted `unsafe` code and that no Windows process had ever been killed by it. That was true of the first two commits. It is not true of the branch. Commits `7bc1043a` and `7bbeda50` add an integration test target and a `windows-latest` workflow, and at head `7bbeda5083` both tests pass on a real kernel ([run 32440615228](https://github.com/proliferate-ai/proliferate/actions/runs/32440615228), job `process-kill tests (windows)`):

```
running 2 tests
test kills_a_real_process_tree_including_the_grandchild ... ok
test the_census_counts_a_real_git_exe_grandchild ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.02s
```

`CreateToolhelp32Snapshot`, the `PROCESSENTRY32W` walk and its `dwSize` handshake, `TreeTracker`, the census, `OpenProcess`/`TerminateProcess` and the confirmation loop have all now run against the real kernel. Getting that verdict out of a crate whose `lib test` target does not build for Windows is genuinely good work, and the PR body's "Verified vs unverified" section badly undersells it. Please refresh it; it is the most valuable part of the description and currently the stalest.

So this is no longer "has any of this run". It is "does it hold in the cases the two tests did not cover", and that is where every blocking finding below lives. Two trees were killed, of two and three processes, both dying on the first `TerminateProcess` pass. **The escalation rung has never executed.** All four blocking defects sit inside it or in the enumeration that feeds it, which is precisely why CI is green.

---

## BLOCKING

### B1. A stale parent link makes the walk terminate an unrelated process tree. `process_kill_windows.rs:175-208`

`absorb` adopts any snapshot row whose `entry.parent` is a tracked pid. Toolhelp's `th32ParentProcessID` is a raw number with no liveness or generation attached, and the module's own doc at `:141-147` says so. Windows recycles pids from a small table. So a long-lived process whose creator exited long ago carries a dangling parent link, and the moment Windows hands that freed number to our agent child, that stranger becomes a "descendant".

The `retired` set does not help. It only protects pids the tracker itself observed and then observed missing. It says nothing about a pid the tracker has never seen, which is exactly the stale-link case.

Reproduced (`TreeTracker` lifted verbatim, `rustc --edition 2021`, run on this machine):

```
// ancient VendorUpdater.exe was started by an installer with pid 8432 that exited long ago.
// Windows recycles 8432 to our agent child. Root = 8432.
[stale parent] pids we will TerminateProcess = [8432, 5000, 5001, 5002]
[stale parent] census = (4, 1)
```

We terminate the updater, its helper and its `git.exe`, count them into `PlaneKills`, and report `killed_git > 0` so `repair_conflict_sentinel` (`adapters/git/operations/snapshot_repair.rs:34`) then acts on evidence manufactured by killing someone else's processes. Nothing logs that anything unusual happened.

This is categorically worse than the bug being fixed. The `(0, 0)` stub leaked trees; this terminates strangers, on a shipping target (`release-runtime.yml:59` builds `x86_64-pc-windows-msvc`, `release-desktop.yml:435` builds msi/nsis).

Mitigation is the standard one for a Toolhelp descendant walk: `OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION)` plus `GetProcessTimes`, and refuse to adopt a candidate whose creation time precedes its claimed parent's. One extra call per candidate, only for rows that would otherwise be adopted.

Same root cause, second facet: the `retired` set is only as good as the poll interval. A member pid recycled inside one 50ms `CONFIRM_POLL` window is never observed absent, stays in `tracked`, and is terminated as a member. The creation-time check closes that too.

### B2. A `Process32FirstW` failure is reported as "the tree is dead". `process_kill_windows.rs:90-107`

`snapshot_processes` guards the `CreateToolhelp32Snapshot` handle against both failure encodings (`:81`, correct), then never checks `Process32FirstW`. If it returns 0, `have_entry` is false, the loop body never runs, and the function returns `Some(vec![])` at `:107`.

`Some(empty)` is the one value the module doc at `:72-75` says must never be confused with failure: "an empty `Vec` means 'nothing is running', and folding a failed enumeration into that is exactly the silent lie this module exists to remove." The handle path honours that. The enumeration-start path does not.

Failure scenario: a `Process32FirstW` failure during the confirmation loop hands `absorb` an empty snapshot. Every tracked pid is "vanished", so `:183-186` retires the entire tree, `wait_until_empty_within` returns `true` at `:282`, and the escalation reports a live tree confirmed dead. Worse, the retirement is permanent: every member is now in `retired`, so even a subsequent successful snapshot in the same call can never re-adopt them (`:198`). One transient failure poisons the tracker for the rest of the call.

Two-line fix: on `Process32FirstW == 0`, `CloseHandle` and return `None`. I would also treat `entries.is_empty()` as failure, since the process table always contains at least our own process.

### B3. A descendant spawned in the kill window is lost permanently, and the tree is reported dead. `process_kill_windows.rs:183-208`

`absorb` retires vanished pids at `:183-186` **before** the growth loop at `:190-208`. So a process whose parent was tracked and alive in the previous pass, but is gone in this one, can never be adopted: its parent left `tracked` moments earlier in the same function call.

This is the precise window the design exists to cover. `kill_tree_and_await` snapshots at `:324`, then terminates at `:335`. Anything the shell spawns between those two lines is not in the census and not in `tracked`, and its parent is about to die.

Reproduced:

```
[lost grandchild] initial census pass = [100, 200]
   // shell 200 forks git.exe 300 between the snapshot and TerminateProcess
[lost grandchild] confirmation pass = []  -> tree reported DEAD (git.exe 300 is STILL RUNNING)
[lost grandchild] next pass         = []  -> permanently unreachable
```

The caller gets `(2, 0)`: two killed, no git, tree quiesced. A `git.exe` holding an index lock is still running. That is the original defect, narrowed to a window rather than removed.

This is also the concrete counterexample to "The CONTRACT is identical on both platforms" (`process_kill.rs:26-29`). On unix `kill(-pgid)` reaches processes created after enumeration, because group membership is inherited across fork. The Windows tracker has no equivalent, so post-enumeration reachability is a second behavioural difference, not merely the graceful rung.

Fix: grow to fixpoint first, then retire. I verified the ordering is the whole difference; with the parent still in `tracked` during growth, pid 300 is adopted. The `retired` set still blocks genuine cross-pass reuse because it is populated from earlier passes. Land this with B1, since grow-first widens the adoption surface B1's creation-time check has to police.

### B4. Transient `CreateToolhelp32Snapshot` failure returns the exact `(0, 0)` lie. `process_kill_windows.rs:80-83`, `:324-330`

`CreateToolhelp32Snapshot` documents `ERROR_BAD_LENGTH` as a transient, retryable failure under process-table churn. There is no retry anywhere. One transient failure on the initial snapshot returns `(0, 0)` from `:329`, which every caller reads as "nothing was running" (`live/sessions/actor/run.rs:117`, `workspace_stop.rs:113`/`:168`) and archive reads as quiesced. `tracing::error!` reaches none of them, and `GetLastError` is never consulted, so the log cannot even say it was the retryable one.

A bounded retry (3 attempts) inside `snapshot_processes` costs nothing. Given the whole PR is about not lying through this tuple, leaving the dominant real-world path that produces it unretried is not defensible.

---

## On the failure the runner found, and how it was resolved

This is the part most worth scrutinising, since "the code was right and my test was wrong" is a conclusion an author is motivated to reach. **I checked it independently and the author is correct.** The exact count is not load-bearing, and I can now say that exhaustively rather than by sampling:

- `killed_git` has exactly two reads in production: `snapshot_repair.rs:34` (`== 0`) and `:185` (`> 0`). Both boolean.
- `QuiesceReport` derives only `Debug, Clone, Copy` (`adapters/git/types.rs:295`). `PlaneKills` derives only `Debug, Clone, Copy, Default, PartialEq, Eq` (`process_kill.rs:27`). Neither is `Serialize`, so neither count is ever returned in an API response, persisted, or emitted in a log line.
- `QuiesceReport.killed` (the total) has **zero** production readers; it is summed at `quiesce.rs:90` and never read.

So relaxing `assert_eq!(git, 1)` to `git >= 1` hid nothing. The census counting the Git-for-Windows shim plus the real binary is the truthful answer to "how many git processes did we kill", and every consumer only asks whether the number is nonzero.

Two riders:

**The relaxed assertion is weaker than it needs to be.** `git >= 1` is a floor; it cannot catch over-matching. The test already computes `git_names` from the CIM oracle a few lines above, so `assert_eq!(git, <count of oracle rows named exactly git.exe>)` is a true oracle cross-check at the same one-line cost, tolerant of the shim, and would fail if someone later loosened `is_git_executable` to a prefix match. Worth taking.

**There is a forward trap in the same discovery.** `domains/sessions/runtime/lifecycle_tests.rs:113` asserts `kills.git == 1` and `:182` asserts `kills.git == 2`. Those are unix-gated and unaffected today, but follow-up 1 in the PR body proposes extending Windows CI to `cargo test -p anyharness`, and both will fail on Windows for exactly the shim reason just discovered. Worth noting in the follow-up so the next person does not re-derive it from a red run.

**The oracle-backed negative control: sound, with one nit.** Asserting that at least one tree process is git-named and that none is bare `git` does establish that a `name == "git"` comparator would have counted zero while git was running, which is the right counterfactual for a naive port of the unix literal. The nit is that the oracle reads CIM `Win32_Process.Name` while the code reads `szExeFile`; for a control specifically about string matching, the string the code actually reads is the one that matters. They are the same base name in practice, so this is a nit and not a defect. The control is a real one and it is better than what most PRs ship.

---

## Non-blocking

### N1. `kill_active_run_for_workspace` releases the pid before killing it. `live/terminals/command_runs/workspace_stop.rs:162-168`

`active.abort_handle.abort()` runs before `kill_group_and_await(pgid)`. Aborting drops the task, which drops the `Child`, which closes the process handle. On Windows an open handle is what reserves a pid; once closed and the process exits, the number is immediately reusable. So this site hands the tree walk a pid it no longer holds a reference to: B1's hazard with the safety net removed.

The other two sites are fine and I checked them. The session actor holds `self.child` across the `tokio::join!` (`actor/run.rs:124-127`) and the terminal path joins `kill_session_and_await` with `reap_pty_child(handle.clone())` (`workspace_stop.rs:113`). Both keep the handle open for the duration. Reordering the abort to after the kill makes this site match.

### N2. The Job Objects rejection rests on a claim that is false

The PR body states: "one of the three spawn sites is inside portable-pty's ConPTY backend. There is no hook to assign the PTY child to a job without forking the crate."

`portable-pty-0.9.0/src/lib.rs:141-145` puts both hooks on the public `Child` trait:

```rust
fn process_id(&self) -> Option<u32>;
/// Returns the process handle of the child process, if applicable.
/// Only available on Windows.
#[cfg(windows)]
fn as_raw_handle(&self) -> Option<std::os::windows::io::RawHandle>;
```

`WinChild` implements both (`src/win/mod.rs:109` and `:118`). `AssignProcessToJobObject` takes a process handle and works on an already-running process; it needs no spawn hook, and nested jobs have been supported since Windows 8. The code already calls `h.child.process_id()` on that same trait object.

The other two rejection reasons hold (the `i32` interface, and `TerminateJobObject` yielding no census), and Job Objects as a follow-up is defensible. But the reason presented as load-bearing is not true and the conclusion should not rest on it. The honest version is a spawn-to-assign race window, which is far smaller than "not reachable".

### N3. The cycle-safety test proves nothing. `process_kill_windows.rs:471-483`

`absorb_terminates_on_a_parent_cycle` builds pids 210 and 220 pointing at each other, but neither's parent is tracked, so neither is ever adopted, so neither reaches `alive` at `:214-218`, so `depth_of` is never called on them. For the two entries that are in `alive`, the `!self.tracked.contains(&parent)` guard at `:232` returns on the first hop. `DEPTH_LIMIT` is never exercised.

Negative control: recompiled with `const DEPTH_LIMIT: u32 = 0;`, the test still produces `[100, 200]`. It passes with the guard it claims to test set to zero.

The mechanism does work. I built the case the test should have been (root 100 whose parent link points at its own child 200, so the cycle is inside `tracked`) and it terminates correctly via `DEPTH_LIMIT`. Please make the test that one.

### N4. The pid-reuse test proves nothing either. `process_kill_windows.rs:443-455`

By the time `a_retired_pid_is_never_readopted_when_windows_recycles_it` presents the recycled snapshot, the whole tree has died and `tracked` is empty, so nothing can be adopted regardless of what `retired` holds.

Negative control: deleted `self.retired.contains(&entry.pid)` from the guard at `:198` and the test still passes.

The real control needs a surviving tracked member to re-adopt through. With root 100 alive, child 200 retired, and pid 200 recycled to `notepad.exe` under 100:

```
with `retired`:    [100]              <- correct
without `retired`: [100, 200, 201]    <- innocent process and its child terminated
```

That version fails without the mechanism. The shipped one does not. Given that terminating the wrong process is the worst outcome this module has, this is the test that matters most and it is currently vacuous.

### N5. These tests do not need Windows at all, and should not be waiting on follow-up 1

The reason given for the unit tests being uncompiled is the broken `lib test` target on Windows. That reason is real but it does not apply to these tests, because **none of the logic they cover is Windows-specific.** `ProcessEntry`, `census`, `is_git_executable`, `executable_name`, `TreeTracker`, `absorb` and `depth_of` touch nothing but `std::collections`, `String` and `[u16; 260]`. I compiled and ran all of them on macOS with plain `rustc --edition 2021` to produce the reproductions in B1, B3, N3 and N4.

So moving that logic into a platform-neutral module gated `#[cfg(any(windows, test))]` would have the existing Linux `Cargo check & test` job compile and run pid-reuse retirement, cycle safety and self-pid exclusion on **every PR**, today, with no Windows runner and no dependency on porting the ten pre-existing errors. That closes the "reasoned-about only" gap immediately and is a much better answer than follow-up 1.

Note this cannot be done inside `tests/windows_process_tree.rs`: `TreeTracker` and friends are `pub(super)` inside a private module, so an integration test compiled against the public surface can only ever reach `kill_group_and_await`/`kill_session_and_await`. The trick that makes the integration test work is the same thing that caps what it can cover.

### N6. The last silent `(0, 0)` is still silent. `process_kill_windows.rs:309-311`

`if root <= 0 { return (0, 0); }` with no log, in a PR whose entire thesis is that this tuple must never be returned quietly. Windows pids are `u32` and every call site narrows with `as i32` (`actor/run.rs:123`, `workspace_stop.rs:112`), so a pid above 2^31 wraps negative and silently reports a clean kill. Rare, but the `#[cfg(not(any(unix, windows)))]` stub got a `tracing::error!` for exactly this reason and this path did not.

### N7. `OpenProcess` conflates access-denied with already-exited. `process_kill_windows.rs:249-255`

Both return early with no distinction and no `GetLastError`. An elevated survivor we cannot open produces no specific signal: we retry `TerminateProcess` every 50ms for 15s and emit a generic "confirmation budget exceeded". The unix path ignores failed `kill` too, so this is not a regression, but `ERROR_ACCESS_DENIED` deserves its own log line since it is the one case retrying can never fix.

### N8. The only unix-visible line in this PR is a broken doc link. `process_kill.rs:16`

See the negative-control section below for why this is the only one. `[`process_kill_windows`](crate::process_kill_windows)` resolves to nothing on any platform: there is no item at that path, and on Windows the module is `crate::process_kill::windows_impl` (`process_kill.rs:363-365`). It sits in the module header, which is the one part of the change that compiles on unix. No CI gate catches it (no `cargo doc` or `broken_intra_doc_links` job in `.github/workflows/`), so it is cosmetic, but it is worth fixing precisely because it is the sole unix-visible line.

### N9. Spec paragraph overclaims. `specs/anyharness/live-runtime.md:717-721`

"The contract is identical ... with one behavioral difference: `TerminateProcess` is unconditional." B3 is a second behavioural difference: unix `kill(-pgid)` reaches processes created after enumeration, the Windows tracker can lose them. The pid-reuse exposure (B1) is a third and is not mentioned. Otherwise the paragraph is accurate, correctly scoped, and matches the file's hard-wrapped house style.

### N10. Workflow gaps. `.github/workflows/windows-process-kill-tests.yml:22-26`

`paths` omits `anyharness/crates/anyharness-lib/Cargo.toml` and `Cargo.lock`, so a future PR that bumps or drops `windows-sys` will not re-run the one job that proves the FFI works. Also no `timeout-minutes`, and unlike `windows-compile-check.yml:27` this job is not `continue-on-error`, so it can redden a unix-only PR that merely touches `process_kill.rs`. That last point may be intentional; flagging it so it is a decision rather than an accident.

---

## What I verified, and how

**The unix negative control: all three claims checked mechanically, one caveat.**

1. *Every changed line in `process_kill.rs` sits inside `#[cfg(windows)]` or `#[cfg(not(any(unix, windows)))]`.* I scripted the diff and classified all 68 changed lines. **True for every line of code**: the 4 removed code lines are the two `#[cfg(not(unix))]` stub signatures, and all 43 added code lines sit under `#[cfg(windows)]` or `#[cfg(not(any(unix, windows)))]`. The caveat is that 5 removed and 16 added lines are module-level `//!` doc, which is not inside any cfg and does compile on unix. Behaviourally inert, but the claim as phrased is not literally true, and the one unix-visible line it introduces is the broken doc link in N8.
2. *`unix_impl`, its `pub use`, and `process_kill_tests.rs` are untouched.* `git diff --quiet` on `process_kill_tests.rs` exits 0, byte-identical. `mod unix_impl` and `pub use unix_impl::{kill_group_and_await, kill_session_and_await}` do not appear in the diff at all. `#[cfg(test)] #[cfg(unix)] pid_is_alive` and `#[cfg(test)] #[cfg(unix)] #[path = "process_kill_tests.rs"] mod tests` are unchanged and still double-gated.
3. *`process_kill_windows.rs` is reachable only through a `#[cfg(windows)]` path module, and the integration test carries `#![cfg(windows)]`.* `git grep` over the branch finds exactly one module declaration, `mod windows_impl;` at `process_kill.rs:365` under `#[cfg(windows)]` at `:363`; `lib.rs` contains zero occurrences of `process_kill_windows`. `tests/windows_process_tree.rs:28` is `#![cfg(windows)]`.

`Cargo check & test` is green on the PR, which is the behavioural confirmation.

**The integration-test trick: it does what it claims.** `cargo test -p anyharness-lib --test windows_process_tree` builds the library as an ordinary dependency (non-test cfg) plus the integration test as its own crate, so the crate's `#[cfg(test)]` modules and their ten pre-existing Windows errors are never on the build path. The green run is the proof; the reasoning is sound. It also means the job still catches any non-test Windows compile break in the library itself, which is a bonus over `--test`-only intent. Its limit is N5: public surface only.

**FFI, checked against the vendored `windows-sys-0.61.2` in the local registry.** `PROCESSENTRY32W` is `#[repr(C)]` with `szExeFile: [u16; 260]` and `impl Default = zeroed`, so the `dwSize: size_of::<PROCESSENTRY32W>()` handshake at `:84-87` is correct and uses the W variant. `Process32FirstW`/`Process32NextW` take `*mut PROCESSENTRY32W` and return `BOOL`; `OpenProcess` returns `HANDLE`, NULL on failure, so `.is_null()` is right; `CreateToolhelp32Snapshot` returns `HANDLE` and the code checks both NULL and `INVALID_HANDLE_VALUE` at `:81`. All three feature names exist in that crate's `Cargo.toml` (`:85`, `:200`, `:250`).

**Handle discipline: clean.** `snapshot_processes` reaches `CloseHandle` at `:104` on every path including the one where `Process32FirstW` fails, and `terminate` closes at `:257` on the only path that obtains a handle. No leak on any early return or error path I can construct.

**`szExeFile` decoding: correct.** `:112-115` stops at the first NUL and falls back to `raw.len()` via `unwrap_or`, so a name that fills the buffer without a terminator decodes rather than over-reading.

**`u32`/`i32`: no UB.** `root as u32` at `:312` happens only after the `<= 0` guard. The only issue is the silent return (N6).

**`Cargo.lock`: claim is exact.** `diff` of the lock across the branch is literally one line, `+ "windows-sys 0.61.2"` added to `anyharness-lib`'s dependency list at line 421. 0.61.2 was already present. No new package, no version bump. The `[target.'cfg(windows)'.dependencies]` gate at `Cargo.toml:57` means no other platform resolves it.

**Job Objects claim: falsified from the vendored `portable-pty-0.9.0` source.** See N2.

**Tracker logic: executed, not read.** I lifted `ProcessEntry`, `census`, `TreeTracker`, `absorb` and `depth_of` verbatim into a standalone file and compiled it with `rustc --edition 2021` (no cargo, no workspace build). Reran the author's two suspect tests to confirm they pass, then ran the negative controls showing they pass with the mechanism removed (`DEPTH_LIMIT = 0`; `retired` check deleted), then constructed and ran B1 and B3.

## What I could not verify

- **No Windows machine.** Everything about real kernel behaviour comes from CI logs.
- **The escalation rung has never executed.** Both integration tests die on the first `TerminateProcess` pass, so `wait_until_empty_within` returning `false`, the re-enumerate-and-re-terminate loop at `:342-370`, and the `CONFIRM_BUDGET` warning are unrun. B2 and B3 both live there, which is why they are invisible to CI.
- **Only shallow trees, and only two of them.** Two and three processes. No deep tree, no ConPTY session (so `kill_session_and_await` has never run against a real PTY at all), and nothing spawning descendants mid-kill, which is B3's exact scenario.
- **The `#[cfg(test)]` module at `:379-483` still gets no compile verdict anywhere.** The integration test links the library's non-test build, so pid-reuse retirement, cycle safety and self-pid exclusion remain reasoned-about only in this repo. N5 is the cheap way to change that.
- **Timing under load.** Whether a real agent tree dies inside `GRACE` and settles inside `CONFIRM_BUDGET` on a loaded Windows box is unknown; CI's trees die in milliseconds.
- **I built no Rust in this workspace.** No `cargo check`, no `cargo test`, no Docker. The only compiler invocation was `rustc` on a single self-contained file with no dependencies.

---

The shape of this change is right, and the second half of it, getting real Windows execution out of a crate whose test target does not build, is better than the description admits. Two-phase census-then-act matching the unix structure, no spawn-site surgery, the failure-versus-empty distinction taken seriously, and the honesty about what was unverified all read as careful work. B2, B3, N3 and N4 are each a one-to-five line change. B1 is the one that needs real thought before this goes near a shipped Windows build, because it is the one failure mode that is worse than the bug being fixed.

pablonyx:
## Round 2: safe to merge as-is. No blocking items.

All four round-one blockers are fixed, the fifth the author found is real and the fix is sound, the `retired` deletion is safe, and the negative controls are genuine — I reverted each mechanism myself and watched each test go red. Findings below are non-blocking: one is a real (much narrower) residual in B1's argument, one is a new fail-open path introduced by the fifth fix, and the rest are body/doc accuracy in a PR whose whole currency is receipts.

**Method, stated up front so you can discount it correctly.** Three other agents are building on this laptop, so I did not check the branch out into the shared worktree and did not run `cargo` at all. For the negative controls I extracted `process_kill_tree.rs` and `process_kill_tree_tests.rs` from the branch verbatim (`git show FETCH_HEAD:<path>`), wrapped the module body in `mod tree { ... }` with the tests inlined where the `#[path]` declaration points, and compiled it with `rustc --edition 2021 --test`. The source text is byte-identical to what CI compiles (verified with `diff` against the extracted file); the only change is the wrapper, which is exactly what `mod tree;` does in the real build. Positive runs are also confirmed independently on real CI (below), so the harness is only load-bearing for the *mutated* runs.

---

## 1. B1, the generation check. Correct, and the equality case is the right call — but "exact" is overstated at the resolution boundary.

**Handle rights and hygiene: verified by reading, and confirmed at runtime.** `process_kill_windows.rs:153` opens with `PROCESS_QUERY_LIMITED_INFORMATION`, which is the documented right for `GetProcessTimes` (either that or `PROCESS_QUERY_INFORMATION`). `CloseHandle` at `:162` runs before the `ok == 0` branch at `:163`, so the handle is closed on both the success and the failure path, and it is the only path that obtains one. `pid == 0` short-circuits at `:146` before `OpenProcess`. No leak on any path I can construct.

**More than that: this call is now EXECUTED on Windows, which the body does not claim.** In `tests/windows_process_tree.rs:148-152`, `total >= 2` can only pass if `ping.exe` was adopted, and adoption at `process_kill_tree.rs:150-163` requires (a) `root_generation` to be `Some` at `process_kill_windows.rs:269`, (b) `created_at(child)` to be `Some`, and (c) `child_generation >= parent_generation`. In `the_census_counts_a_real_git_exe_grandchild`, `assert_eq!(git, git_names.len())` with the Git-for-Windows shim means two levels of adoption had to succeed. So `PROCESS_QUERY_LIMITED_INFORMATION` being sufficient, `GetProcessTimes` succeeding on a child we spawned, and the ordering invariant holding for genuine children are all established by a real kernel at head. That is the strongest evidence in this PR for the B1 fix and the body attributes it to nothing.

**"At or after" is the right comparison, including equality.** The rejection is `child_generation < parent_generation` (`:158`), so equal timestamps adopt. That has to be the case: `GetProcessTimes` reports `FILETIME` in 100 ns units, but the value is a sample of the system clock at process creation, not a monotonically unique counter. If two processes can land on one reported value, a strict `<=` rejection would refuse genuine children created in the same tick as their parent — a leak, on the common path.

**Where the "exact rather than merely conservative" claim breaks.** The argument (stranger `t_s` ≤ previous owner's exit `t_p` < recycled member's creation `t_m`) is sound in continuous time. It fails at equal *reported* values. I ran that case against the shipped code:

```
PROBE B (stranger same tick as root) -> [100, 200, 201] census (3, 1)
```

(root pid 100 gen 1000; stranger pid 200 with a stale link to 100, gen 1000; its child 201, gen 1000.) The stranger and its child are adopted and terminated. This needs a pid to be freed and reissued inside one clock-tick of the stranger's own creation, which is a far narrower window than the original B1 — I am not asking for a code change. I am asking that the doc at `process_kill_tree.rs:76-82` and the corresponding body paragraph stop saying "exactly right, not merely conservative" and instead say the invariant holds strictly in real time and up to the resolution of the reported creation timestamp. I could not determine on this machine whether Windows samples `EPROCESS` create time from the coarse timer tick (~15.6 ms) or the precise clock on current builds; **I have no Windows machine and did not verify this either way.** The code is correct under both; only the wording is not.

**`GetProcessTimes` failing, and protected/system processes.** `OpenProcess` returning NULL and `GetProcessTimes` returning 0 both collapse to `None` at `:154`/`:164`. For an *adoption candidate* that is fail-closed and correct (`process_kill_tree.rs:155-157`, pinned by `absorb_adopts_nothing_when_the_roots_identity_is_unknown`). For an *already-tracked* member it is fail-open — see finding 2b.

---

## 2. The fifth defect and the `retired` deletion.

**The defect is real.** `may_adopt` (`process_kill_tree.rs:205-212`) returns false for anything already in `tracked`, so a pid that died and was reissued between passes was reached by no check at all. Confirmed by reading; `retired` could not have caught it either, for the reason given.

**2a. The deletion is safe. I built the exact scenario `retired` protected and ran it against the shipped code.** Round one's N4 control was: root 100 alive, child 200 dies, pid 200 is recycled to a stranger whose link names 100, stranger has a child 201. Without `retired`, round one saw `[100, 200, 201]`. Against this branch:

```
PROBE A (stranger older than root)  -> [100]
```

The generation check catches it, because a stranger reached through a link to a *live* tracked pid is necessarily older than that pid. `retired` was strictly weaker: it keyed on "we saw this number die", which is a proxy for identity, and the identity check is the thing itself. `git grep retired` over the branch finds no occurrence in either kill file except the doc paragraph at `process_kill_tree.rs:89-91` explaining the removal. Nothing else referenced it. I could not construct a case `retired` caught that the identity check does not.

**2b. The fifth fix introduces one new fail-open path. Non-blocking, but it is the structural twin of round-one B2.** At `process_kill_tree.rs:132-139`, a tracked pid that is *still listed in the snapshot* but whose creation time cannot be read is `remove`d from `tracked` — "unreadable" is treated as "identity changed" rather than as "unknown". Two consequences, both run against the shipped code:

```
PROBE C (survivor unreadable)        -> [100]        // pid 200 alive and listed, read denied
PROBE D (root read fails one pass)   -> [200]        // root 100 alive and listed, read fails once
PROBE D (next pass, read recovers)   -> [200]        // permanent: the root is never re-tracked
```

Probe D is the one that bothers me: a single transient failure to read the *root's* creation time during a confirmation pass drops the root permanently (it has no tracked parent, so the growth loop can never re-adopt it), and once its children die `wait_until_empty_within` returns `true` on a live root. That is "a live tree confirmed dead", which is what round one blocked B2 on.

I am **not** blocking it, and I want the reasoning on the record so you can disagree: (i) the returned `(total, git)` census is taken before any of this and is unaffected, so no caller is lied to about what was killed; (ii) nothing new gets terminated; (iii) unlike `ERROR_BAD_LENGTH`, there is no *documented* transient failure mode for `OpenProcess`/`GetProcessTimes` against a live process you own — the realistic triggers are handle exhaustion or a security-product hook, not table churn. So the trigger frequency is categorically lower than B2's, which is the difference that makes this a finding rather than a blocker.

The fix is one line and costs nothing: on an unreadable read, `self.tracked.insert(pid, None)` instead of `self.tracked.remove(&pid)`. `None` already means "cannot be a parent" everywhere in this file (`:150`, `:133`), so the process stays killable and countable but can still adopt nothing through it. The one thing it gives up is dropping a pid that was *both* recycled *and* unreadable — and such a process cannot be terminated by us either, since `PROCESS_TERMINATE` is not granted where `PROCESS_QUERY_LIMITED_INFORMATION` is denied in any ACL I would expect to see. Your call whether that lands here or as a follow-up.

---

## 3. Negative controls. All four executed. Every claimed value reproduced literally.

Baseline first, so the mutations mean something. `rustc --edition 2021 --test pkt.rs && ./pkt`:

```
test result: ok. 12 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

Zero compiler warnings on the baseline (bare `rustc`, no `-A`), so the module carries no dead code on the non-Windows test build.

Independently, the same 12 run green on real CI on Linux — `gh run view --job=96654207557 --log` on run `32441930453` (head `60b20ce9`) lists all twelve by name under `process_kill::tree::tests`, inside `test result: ok. 2227 passed; 0 failed`.

**NC1 — delete the `child_generation < parent_generation` rejection (`process_kill_tree.rs:158-162`):**

```
---- tree::tests::absorb_refuses_a_stranger_reached_through_a_stale_parent_link stdout ----
  left: [8432, 5000, 5001, 5002]
 right: [8432]
test result: FAILED. 11 passed; 1 failed;
```

Exactly the `[8432, 5000, 5001, 5002]` the body claims. The `(4, 1)` census assertion is the next line and is unreached, but the four pids include one `git.exe`, so it follows.

**NC2 — delete the identity-verification block (`process_kill_tree.rs:122-139`):**

```
---- tree::tests::absorb_drops_a_tracked_pid_that_was_recycled_between_passes stdout ----
  left: [100, 200]
 right: [100]
```

**NC3 — move the "drop the vanished" block (`:171-182`) back above the growth loop (`:141-169`):**

```
---- tree::tests::absorb_adopts_a_grandchild_whose_parent_died_in_the_same_interval stdout ----
  left: [100]
 right: [100, 300]
```

`[100]`, exactly as the test's own doc comment at `:179-180` predicts.

**NC4 — delete the `seen` visited set from `depth_of` (`:221-222` and `:232-234`):**

```
---- tree::tests::depth_of_reports_a_cycle_rather_than_walking_it stdout ----
  left: Some(64)
 right: None
```

`Some(64)`, exactly as claimed. Round one's two vacuous tests are fixed, and unlike then, each mechanism is now pinned by a test that fails when it is removed. Note NC3 does *not* redden `the_escalation_rung_tracks_a_tree_across_passes_until_it_empties` — that test is a useful sequence check but is not itself a control for the ordering; `absorb_adopts_a_grandchild_...` is.

---

## 4. The `cfg(any(windows, test))` restructure. Same code on both paths — no shim, no divergence.

Verified mechanically, not by eye:

- `process_kill.rs:369-371` declares `#[cfg(any(windows, test))] #[path = "process_kill_tree.rs"] mod tree;` — one declaration, one file.
- `grep -n "cfg("` over `process_kill_tree.rs` returns exactly two hits: `:5` (a doc comment) and `:242` (`#[cfg(test)] mod tests`). There is **no** `cfg(windows)`, `cfg(unix)` or `cfg(target_os)` anywhere in the module body. The bytes Linux compiles under `--test` and the bytes Windows compiles in the library build are the same bytes.
- `grep -n "cfg("` over `process_kill_tree_tests.rs`: zero attribute hits.
- `process_kill_windows.rs:63` is `use super::tree::{census, executable_name, Generation, ProcessEntry, TreeTracker};` and the file defines no tracker, census or name-decode of its own (`grep -n "struct \|fn absorb"` finds none). There is no second copy to drift.
- The gate is exercised in both directions at head `60b20ce9`: `runtime crates (control)` and `desktop crates (the unknown)` are SUCCESS on run `32441930482` (`cargo check --target x86_64-pc-windows-msvc`), which compiles `mod tree` with the `windows` arm true; `Cargo check & test` is SUCCESS on `32441930453` with the `test` arm true.

Nothing meaningful sits outside the gate. The only unix-visible lines in `process_kill.rs` remain the module `//!` doc, and round one's broken intra-doc link is gone (`:16` now says ``process_kill_windows.rs`` in plain backticks).

---

## 5. `kill_active_run_for_workspace` (`workspace_stop.rs:162-168`). Deferring is right.

The trade is real and it is not invented for this review — the function's own doc comment at `:138-140` already states that the abort exists so the killed task "can never overwrite that terminal status with its own completion". Reordering the abort after `kill_group_and_await` at `:168` puts a live task on a dying child for the duration of the kill, which is precisely the window in which it would write its own completion over the `"Run interrupted by workspace stop"` just written at `:151-154`. Trading a certain, user-visible status regression against this race, on a path with existing unix coverage, is the correct call, and the unix suite is the right arbiter for the half that is actually at risk.

One sharpening for follow-up 2, because it makes the residual worse than the body says: **B1's generation check does not protect the root.** The root has no tracked parent to compare against (`process_kill_tree.rs:105-108` seeds it unconditionally), so a recycled *root* pid means `terminate_pass` kills a stranger outright and then legitimately adopts that stranger's genuine children, because the tracker keyed on the creation time it just read. This site is the one that maximally exposes that, so the follow-up should say so rather than describing it as a generic pid race. It is still not a blocker: the window is between `:162` and `:168`, and it requires the child to exit *on its own* and Windows to reissue that exact number inside it. Also worth noting for the follow-up's framing: on unix this ordering is harmless because the dropped `Child` leaves a zombie that reserves the pid until reaped, so the race is Windows-only and the unix suite will never see it.

---

## 6. Windows CI. Confirmed at head, but the body cites the wrong runs.

`gh run view --job=96654207624 --log` on run `32441930543`:

```
running 2 tests
test kills_a_real_process_tree_including_the_grandchild ... ok
test the_census_counts_a_real_git_exe_grandchild ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.00s
```

`gh run view 32441930543 --json headSha` → `60b20ce91f728728235a2b3acc5166432093486e`, which is head. Good.

**But the body's "Verified" section cites `32440615228` for the FFI-executed claim and `32439048862` for the compile claim, and those are `7bbeda5083` and `a3aaad5557` — one and two commits before the identity rewrite that is the entire subject of round 2.** In a PR whose case rests on quoted receipts, the headline receipts point at code that no longer exists. Please repoint them to `32441930543` and `32441930482`/`32441930453`. This is the one thing I would genuinely like changed before merge, and it is an edit to the description, not the code.

**Does the escalation rung still go unreached? Partly, and the body should be more precise than it is.** The post-`GRACE` re-enumerate-and-re-terminate loop at `process_kill_windows.rs:296-327` and the `CONFIRM_BUDGET` warning at `:318-324` are still unexecuted — both trees die on the first pass, and I have no way to prove otherwise from the logs (3.00 s for two tests, dominated by the PowerShell oracle calls). So the body's statement is right about that loop. It is *wrong* to imply the whole tracker is unreached on Windows: `wait_until_empty_within` at `:220-236` calls `refresh` → `absorb` on every confirmation poll, and the very first `refresh` at `:279` is what produces the census, so multi-pass `absorb`, the identity check and the generation comparison *do* run on the real kernel. Worth splitting in the body, because it converts "reasoned about only" into "the adoption half is executed on Windows; the post-grace retry rung is not".

The two integration tests still exercise what they claim, and the census assertion got *stronger* since round one: `assert!(git >= 1)` became `assert_eq!(git, git_names.len())` at `tests/windows_process_tree.rs:227-232`, which is the oracle cross-check round one asked for and now catches over-matching. Small nit: `git_names` is built with `starts_with("git")` at `:209` while `is_git_executable` (`process_kill_tree.rs:50-53`) requires exactly `git.exe`/`git`, so a stray `gitk.exe` in the tree would fail this assertion spuriously. The oracle is over-inclusive relative to the code, not under — harmless here, but `== "git.exe"` would match the comparator exactly.

---

## 7. Settled, confirmed, closing it

Round one's heads-up about `lifecycle_tests.rs:113`/`:182` was wrong on both counts, the author's pushback is correct, and the body records it accurately in "One correction back". I am not relitigating it and neither should the next reader.

---

## Small open items (none blocking, none new except the first)

- **Stale doc in the integration test.** `tests/windows_process_tree.rs:14-16` still says "the unit tests inside `process_kill_windows.rs` cover the pure logic but cannot be compiled on Windows". After the restructure that file has no unit tests, and the ones that exist now run on every PR. The paragraph argued for a gap that this PR closed.
- **Round-one N6 stands.** `process_kill_windows.rs:252-255` still returns `(0, 0)` for `root <= 0` with no log, in the module whose thesis is that this tuple is never silent.
- **Round-one N7 stands.** `terminate` at `:195-198` still cannot distinguish `ERROR_ACCESS_DENIED` from "already gone". Not a regression; it is the one case retrying can never fix and it deserves its own log line.
- **Round-one N10 stands.** `windows-process-kill-tests.yml:22-28` gained the two new source paths but still omits `anyharness/crates/anyharness-lib/Cargo.toml` and `Cargo.lock`, so a `windows-sys` bump or drop will not re-run the only job that proves the FFI links. Still no `timeout-minutes`.
- The Job Objects correction (round-one N2) is recorded honestly, with the false premise struck rather than quietly deleted, and the conclusion re-based on the two grounds that hold. The spec paragraph at `specs/anyharness/live-runtime.md:711-726` now names all three weakenings, which closes round-one N9.

---

**Verdict.** Merge-ready on the code. Everything round one blocked is fixed and each fix is pinned by a test I watched fail with the mechanism removed. Ship the body corrections in item 6 with it, since the receipts are the argument.

*What I verified by execution:* the 12 tree tests and all four negative controls, via `rustc --edition 2021 --test` on source extracted verbatim from the branch; probes A–D against the same shipped logic; the Windows and Linux CI runs and their head SHAs, via `gh run view --log` / `--json headSha`. *What I concluded by reading:* handle rights and handle hygiene in `process_creation_time` and `terminate`; the `cfg` gate purity (backed by `grep` over the branch); the `workspace_stop.rs` ordering trade; the `retired` removal argument (backed by probe A). *What I could not verify:* anything about real Windows kernel behaviour — no Windows machine, so the creation-timestamp resolution question in item 1 and the "both tests die on the first pass" claim in item 6 are reasoned, not measured. I ran no cargo build and did not touch the working tree; `git status` is byte-for-byte as I found it.

pablonyx:
## Round 3 (post-rebase pass): safe to merge on content. One doc fix needed first, not a code defect.

**Scope note.** No cargo build run (checked `pgrep -x cargo`/`rustc` clean before and after; the machine has ten other reviewers on it right now). No Docker, no push/merge/label/ready-for-review changes. Verified by reading blobs at named SHAs, `git range-diff`/`merge-base`, and `gh api`/`gh run view --log` only.

**Timeline correction, stated up front.** At the time I started, the branch had *not* actually been rebased yet — `fix/windows-process-tree-kill` was still at `b22f10fb0c`, still parented on the superseded `8de5a45697`, with `68777b5614` (the new base tip at that moment) not even an ancestor of it. Only `#2142` had moved. That's since been corrected: the branch now sits at a genuinely rebased `6c48750e18`, on top of the current base tip `58f07c205d`. Everything below is checked against that real, current state.

---

### 1. The rebase is clean.

`git range-diff 8de5a45697..b22f10fb0c 58f07c205d..6c48750e18` — all six commits map 1:1 (`444d7e43c0`→`9b82a0758a`, `a3aaad5557`→`698d85c033`, `7bc1043a08`→`ad641c3575`, `7bbeda5083`→`108e251ff0`, `60b20ce91f`→`afb1666c8e`, `b22f10fb0c`→`6c48750e18`), every one flagged `=` (patch-identical). Nothing dropped, nothing reordered, no content drift. `git diff b22f10fb0c 6c48750e18 --stat` is exactly the 106-file base-branch delta (the launch-options-cutover work `#2142` picked up, plus `#2142`'s own `windows-compile-check.yml` hardening commit `58f07c205d` — pinned actions, `timeout-minutes: 45`, a `Cargo.lock` path trigger, concurrency group) — zero touches to any of this PR's 10 files. `git merge-base --is-ancestor 58f07c205d 6c48750e18` confirms the new base tip is now a real ancestor of the PR head, not just a synthetic merge target.

### 2. CI is green at the actual current head, reverified live rather than trusted from before.

Polled `gh pr checks 2147` until the full run against `6c48750e186ba14fb8daf020f74bbda29b674720` (kicked off `2026-08-21T05:32:29Z`, i.e. after the real rebase) finished: all required checks pass, docker/self-hosted-release-assets skip as expected, `ci-ok`/`server-ci-ok` both pass. Pulled the two load-bearing logs directly rather than trusting the checkmark:

- `process-kill tests (windows)`, job `96679171510`: `running 2 tests / test kills_a_real_process_tree_including_the_grandchild ... ok / test the_census_counts_a_real_git_exe_grandchild ... ok / test result: ok. 2 passed; 0 failed; ...`
- `Cargo check & test`, job `96679172461`: all thirteen `anyharness-lib::process_kill::tree::tests::*` show `PASS` (`absorb_adopts_a_grandchild_...`, `absorb_adopts_nothing_when_the_roots_identity_is_unknown`, `absorb_collects_the_whole_descendant_tree_root_first`, `absorb_drops_a_tracked_pid_that_was_recycled_between_passes`, `absorb_keeps_a_tracked_pid_whose_identity_cannot_be_read`, `absorb_never_adopts_the_runtimes_own_process`, `absorb_refuses_a_stranger_reached_through_a_stale_parent_link`, `absorb_still_adopts_a_genuine_child_created_after_its_parent`, `depth_of_measures_an_ordinary_chain`, `depth_of_reports_a_cycle_rather_than_walking_it`, `executable_name_stops_at_the_nul_terminator`, `the_census_counts_git_exe_where_an_exact_git_comparison_would_count_zero`, `the_escalation_rung_tracks_a_tree_across_passes_until_it_empties`).

Both match the pre-rebase numbers literally. Nothing in the base's 106-file delta touches this crate's kill path, so this is exactly the outcome §1 predicts, now confirmed rather than assumed.

### 3. Nothing from rounds 1–2 is being relitigated.

Content is patch-identical to `b22f10fb0c`, which was already round-2's closed state: all four original blockers plus the fifth self-found one, both negative-control reproductions (`[8432, 5000, 5001, 5002]`, `Some(64)`), the `retired`-deletion safety probe, the `workspace_stop.rs:163` deferral, and the `lifecycle_tests.rs:113`/`:182` retraction all stand as closed.

### 4. Two of the three closeout items are confirmed intact at the new head; the third — receipts — needs a follow-up edit.

- **Softened "exact" claim** — `process_kill_tree.rs:88` at `6c48750e18`: "...the residual is the same-tick window rather than a bug. How wide that window is depends on the sampling resolution `GetProcessTimes` actually reports on a given machine, which nothing here has measured; do not assume it is the 100ns the FILETIME encoding suggests." Present, unchanged.
- **Fail-open one-liner** — `process_kill_tree.rs:171` at `6c48750e18`: the unreadable-identity arm of `absorb`'s re-verification loop is `self.tracked.insert(pid, None);`, not `remove`. Present, unchanged.
- **Receipts — now genuinely stale, not just "SHA changed under it."** The PR body's "Verified vs unverified" section still says "at head `b22f10fb0c`" three times and links `32443654643`, `32443654675`/job `96659117018`, and `32443654682`/job `96659116436` — all runs against the pre-rebase commit. I diffed the body before and after this second rebase: byte-identical, so it was never touched. Since the head is now `6c48750e18`, please repoint those three citations to the runs in §2 above (`32450924720` for the compile check, `32450924690`/job `96679172461` for the 13 Linux tests, `32450924588`/job `96679171510` for the Windows integration pair) before merge. This is exactly the failure mode round 2 flagged in its own item 6 — a PR whose whole case rests on quoted receipts pointing at code that no longer is the head — just recurring one rebase later. Non-blocking for the code, but it should not ship unfixed given that framing.

---

**Verdict: safe to merge as-is on content.** The rebase is clean, CI is green at the true current head with the two load-bearing test outputs reconfirmed literally, and both substantive round-2 closeout fixes are present and unchanged. The only outstanding item is cosmetic: update the three `b22f10fb0c`-era receipt links in the PR description to the `6c48750e18` runs cited above.

PR #2148, opened 2026-08-21
fix(anyharness): remove hardcoded posix shell and home assumptions

pablonyx:
Stacked on `ci/windows-compile-check` (draft PR #2142), not `main`. That branch carries the Windows `cargo check` workflow, which is the only compile verdict available for this change. Retarget to `main` once #2142 lands.

## What this is

A sweep of `anyharness/crates/` (plus the in-scope parts of `apps/desktop/src-tauri`) for hardcoded POSIX shell and environment assumptions, and a fix for the ones that break the runtime on Windows.

None of these are compile errors. The runtime crates already build for `x86_64-pc-windows-msvc` on every release, so every `std::os::unix` use is already `cfg`-gated. These are runtime failures.

Two shared helpers rather than `cfg` at every call site, matching the `#[cfg(unix)] fn X` / `#[cfg(not(unix))] fn X` pairing the codebase already uses (`process_kill`, `integrations::agent_cli::executable::make_executable`, `integrations::mcp::capability_token::write_secret_file`):

- `anyharness-lib/src/host_shell.rs`: how a user-authored command string reaches a shell, for the piped setup runner only.
- `proliferate-worker/src/posix_shell.rs`: the worker's genuinely POSIX supervision scripts, refused up front on other platforms with an `Unsupported` error that names the reason.

## What this fixes on Windows today, and what it does not

Stated plainly because it is easy to read the table above and assume otherwise: the `setup_process.rs` change is correct but **not reachable on Windows today**. It is groundwork, not a live fix.

Both callers of `run_setup_process` first create a terminal, and that is where Windows stops:

- `live/terminals/manager.rs:194-214` calls `self.create_terminal_shell(..)` with `shell: Some(detect_posix_shell())`, then reaches `run_setup_process` at `manager.rs:254`.
- `live/terminals/command_runs/workspace_stop.rs:268-287` calls `driver::create_terminal_shell(..)` with the same `shell: Some(detect_posix_shell())`, then reaches `run_setup_process` at `workspace_stop.rs:328`.

`detect_posix_shell` (`driver/shell.rs:59`) probes `/bin/bash`, `/usr/bin/bash`, `/bin/zsh`, `/usr/bin/zsh`, `/bin/sh`, `/usr/bin/sh` and falls back to the literal `"/bin/sh"`. None of those exist on Windows, so the PTY spawn inside `create_terminal_shell` fails and `?` propagates out of both call sites before `run_setup_process`, and therefore before `host_shell`, is ever entered.

So on Windows the setup-command path fails at terminal creation, both before and after this PR. What changes is that when terminal creation is eventually made to work on Windows, the command string behind it is already quoted correctly instead of being handed to a shell that does not exist. Making terminal creation itself work on Windows is a separate and much larger change: it means a ConPTY-backed driver and a non-POSIX replacement for the sentinel wrapper protocol at `pty.rs:473`, which is deliberately out of scope here (see "Found and deliberately not changed").

The three worker sites are different. Those are genuinely POSIX supervision scripts and this PR makes them fail with a typed `Unsupported` error naming the reason rather than by trying to spawn `sh`. The four `HOME` / `USERPROFILE` sites are live fixes on Windows today, and the `home_dir_os` change is a live fix on unix today.

## Sites changed

| file:line (pre-change) | assumed | now | unix unchanged? | category |
| `anyharness-lib/src/live/terminals/command_runs/setup_process.rs:64` | `/bin/sh -lc <setup command>` | `host_shell::command_string_shell()`: `/bin/sh -lc` on unix, `%ComSpec%` (default `cmd.exe`) `/C` with `raw_arg` elsewhere | yes. Same program, same flag, same `arg` quoting. Pinned by a test asserting `get_program()` is `/bin/sh` and `get_args()` is exactly `["-lc", <command>]` | shell genuinely needed: user-authored string that may use pipes, `&&`, globs, expansion |
| `proliferate-supervisor/src/config.rs:116` (`dirs_fallback_home`) | `HOME` only, else `"."`, so the supervisor state dir landed in the process cwd on Windows | `HOME`, then `USERPROFILE`, then `"."` | yes, `HOME` still wins whenever set, and `var_os` throughout | env assumption |
| `proliferate-supervisor/src/install/layout.rs:6` (`default_home`) | `PROLIFERATE_HOME`, then `HOME` only | `PROLIFERATE_HOME`, then `HOME`, then `USERPROFILE` | yes, both earlier keys keep priority, `var_os` throughout | env assumption |
| `proliferate-worker/src/supervisor_bridge/bridge/mod.rs:298-305` (`supervisor_live`) | `sh -c "pgrep -f '<pattern>' > /dev/null 2>&1"` | direct `Command::new("pgrep").arg("-f").arg(&pattern)` with null stdout and stderr, no shell, no quoting | yes. The function returns a bool. `pgrep`'s exit status is unchanged, the redirects become the null stdio, stdin stays inherited, and the "pgrep is missing" case moves from a non-zero shell status to a spawn error, which `unwrap_or(false)` maps to the same `false` | shell was not needed: fixed program, fixed arguments |
| `proliferate-worker/src/supervisor_bridge/bridge/mod.rs:319` (`spawn_supervisor`) | `bash -lc <detached nohup launch script>` | `posix_shell::run_script(BashLogin, ..)`, same `bash -lc` on unix, typed `Unsupported` error elsewhere | yes, same program, flag and script string; pinned by a test on the literals | shell genuinely needed and genuinely POSIX: `nohup`, `$$`, `$PPID`, `pgrep` |
| `proliferate-worker/src/anyharness_update.rs:369` (`stop_runtime`) | `sh -c <pgrep/kill script>` | `posix_shell::run_script(Sh, ..)` | yes, same program, flag, script string and error mapping | shell genuinely needed and genuinely POSIX |
| `proliferate-worker/src/anyharness_update.rs:439` (`relaunch`) | `sh -c <nohup relaunch script>` with `current_dir(workdir)` | `posix_shell::run_script(Sh, .., Some(workdir))` | yes, same program, flag, script, cwd and error mapping; cwd pinned by a test | shell genuinely needed and genuinely POSIX |
| `apps/desktop/src-tauri/src/commands/diagnostics.rs:265` (`expand_home_path`) | `var_os("HOME")`, error when unset | new `app_config::home_dir_os()`: `var_os("HOME")` then `var_os("USERPROFILE")`, original `"HOME is not set"` string preserved | yes, for every `HOME` value including non-UTF-8 and empty | env assumption |
| `apps/desktop/src-tauri/src/commands/ssh_tunnel.rs:716` (`expand_home_path`) | `var_os("HOME")`, error when unset | same, with the original `"HOME is not set."` string preserved | yes, same reasoning | env assumption |

## Review round 1: what changed

**B1, terminal shell substitution: withdrawn.** The first revision routed `driver/shell.rs` shell detection through the helper, so the `/bin/sh` last resort became `%ComSpec%` on Windows. That is reverted; `driver/shell.rs` is now byte-identical to the base branch, and `host_shell` no longer exposes shell detection at all.

The review's stated mechanism for why it was wrong does not hold, and it is worth recording why, because the conclusion is right for a different reason. The concern was that `cmd.exe` would spawn happily, the POSIX sentinel wrapper at `command_runs/pty.rs:473-477` would be written into it, the `__ANYHARNESS_CMD_END_..__` sentinel would never come back, and an untimed run would hang forever.

That cannot happen. `build_pty_command_wrapper` has exactly one caller, `run_terminal_command` at `command_runs/pty.rs:92`, and line 59 of the same function is:

```rust
if !h.shell_kind.is_posix() {
    anyhow::bail!("unsupported_terminal_shell");
}
```

`h.shell_kind` is `detect_shell_kind(&shell)` from `driver/pty.rs:69`. For `cmd.exe` that is `ShellKind::Other`, and `is_posix()` is `matches!(self, Bash | Zsh | Sh)`, so the bail fires roughly thirty lines before the wrapper is built. No script is written, no command-run row is inserted, nothing is written to the PTY. The other two paths that could have carried a POSIX payload are clear too: `startup_command` (`manager.rs:131`) funnels through the same guarded `run_terminal_command`, and both `run_setup_process` callers (`manager.rs:202`, `workspace_stop.rs:276`) use the piped runner, which is bounded by an unconditional `tokio::time::sleep_until(deadline)` arm at `setup_process.rs:206` with a non-optional `Duration`.

So the substitution could not hang. What it could do is worse than useless: it turned "PTY spawn fails at terminal creation" into "terminal opens, then every command run is refused with `unsupported_terminal_shell`". It bought nothing on Windows and moved the failure later and further from its cause. Reverting is the right call, and the review's posture point stands even though the mechanism does not: `host_shell` should not hand back an interpreter that the same crate's protocol will reject. `host_shell` is now scoped to the one path that does not use that protocol, and its module docs record the guard so nobody re-derives this.

**B2, non-UTF-8 `HOME`: accepted, real, fixed.** Routing the two desktop `expand_home_path` helpers through `app_config::home_dir()` silently moved them from `var_os` to `var` (`app_config.rs:59`). `std::env::var` returns `Err(VarError::NotUnicode)` for a non-UTF-8 value where `var_os` returns it intact, so a Linux or macOS user with a latin-1 home path would have gone from a working `~/Desktop/x.json` to "HOME is not set". Both sites now use a new `app_config::home_dir_os()` that keeps `var_os` semantics. `home_dir()` itself is left alone: its callers build the app directory and already live with the `String` constraint, and changing it would be an unrelated behaviour change.

Also taken from the review:

- `cmd.exe /C` needs `raw_arg`, not `arg`. `arg` applies the MSVCRT quoting rules on Windows, which `cmd.exe` does not parse, so `npm run build --if-present` would have arrived wrapped in quotes. `host_shell::command_string_shell` now returns a fully configured `tokio::process::Command` rather than a program-and-flag pair, so the quoting decision lives with the platform decision and no caller can miss it.
- The `[/]` escape at `bridge/mod.rs:398` is now vestigial for `supervisor_live`, which has no shell to exclude, while still load-bearing for the two callers that interpolate the pattern into a script. The doc comment says so, and the escape is kept uniform so all three callers pass byte-identical patterns.
- Both new modules now have tests, including negative controls that pin the exact unix program and flag literals (`/bin/sh` + `-lc`, `sh` + `-c`, `bash` + `-lc`). If someone changes one, the test fails.

## Found and deliberately not changed

| file:line | finding | why not now |
| --- | --- | --- |
| `anyharness-lib/src/live/terminals/command_runs/pty.rs:442-478` | `write_command_script` emits a `.sh` with `export K=V`, and `build_pty_command_wrapper` drives a POSIX sentinel protocol into the PTY | a Windows equivalent needs a `.cmd`, `set` instead of `export`, and a different sentinel protocol. A redesign, not a portability tweak. The existing `is_posix()` guard means the current behaviour is an explicit refusal, not a hang |
| `anyharness-lib/src/integrations/agent_cli/launcher.rs:86` | `generate_launcher_script` writes a `#!/bin/sh` wrapper and `chmod`s it | Windows needs a generated `.cmd`, plus every consumer of the launcher path must agree on the extension. `make_executable` is already a no-op on non-unix |
| `apps/desktop/src-tauri/src/sidecar.rs:176-178` | `format!("{home}/.cargo/bin/anyharness")` from bare `HOME`, plus an unconditional `/usr/local/bin/anyharness` candidate | the real Windows defect there is the missing `.exe` suffix across all six candidates and the bundled-sidecar lookup above them. Fixing only `HOME` leaves it equally broken, so it belongs with the desktop `.exe`-suffix pass |
| `anyharness/src/commands/catalog_probe/catalog_probe_auth.rs:139,209` | `format!("{}/.codex/auth.json", var("HOME"))` and the Grok equivalent | developer probe tooling, off the shipped runtime path. It also deliberately *sets* `HOME` for isolated child agents, which is correct as written |
| `anyharness-lib/src/app/config.rs:59`, `proliferate-worker/src/config.rs:106`, `apps/desktop/src-tauri/src/app_config.rs:59` | home lookups | already `HOME` then `USERPROFILE`. The scoping note's claim of a `HOME` vs `USERPROFILE` gap here does not hold; the real gaps were in the supervisor and the two desktop `expand_home_path` helpers |
| `anyharness-lib/src/domains/agents/installer/seed/quarantine.rs:74` | `/usr/bin/xattr` | already `#[cfg(target_os = "macos")]` |
| `apps/desktop/src-tauri/src/commands/shell.rs:73,92,112`, `.../google_workspace_mcp/auth.rs:258` | `Command::new("open")` | already `#[cfg(target_os = "macos")]` with an explicit unsupported error elsewhere |

Swept and clean: `env!("HOME")` (no hits); `/tmp`, `/usr/`, `/etc/` in non-test code (production code already uses `std::env::temp_dir()`; every `/tmp` hit is a test fixture); `:` as a PATH separator (every site uses `std::env::join_paths` / `split_paths`); `std::os::unix` and file mode bits (every non-test site is already behind a `cfg` pair, which is why the runtime crates already build for Windows).

## Left for other agents

Nothing found in the areas owned elsewhere. `anyharness-lib/src/process_kill_tests.rs:142,223,350` and `apps/desktop/src-tauri/src/sidecar/tests.rs:117,206` spawn `/bin/sh`, but those belong to the `process_kill` and desktop lanes and are test-only.

Separate from this PR and worth its own: `app/test_support.rs:152` does `.expect("expected env mutex")` on a crate-wide lock, so any single teardown panic blacks out every other test that touches the environment. See the CI note below for what that looked like in practice.

## Tests

Not run locally. CI must run them.

New in this PR:
- `anyharness-lib/src/host_shell.rs` (inline `#[cfg(all(test, unix))] mod tests`: pins `/bin/sh` + `-lc` exactly, asserts the status is returned verbatim, asserts shell syntax is actually interpreted)
- `proliferate-worker/src/posix_shell.rs` (inline `#[cfg(all(test, unix))] mod tests`: pins `sh`/`-c` and `bash`/`-lc`, exit status, requested cwd, inherited cwd)
- `apps/desktop/src-tauri/src/app_config_home_env_tests.rs` (registered from `app_config.rs` as `#[cfg(all(test, unix))] mod home_env_tests`: pins `home_dir_os` against the `var`-vs-`var_os` regression)

### Review round 2: `home_dir_os` had no coverage

Correct and worth taking seriously. Round 1 added `home_dir_os` to fix a real unix regression and then pinned nothing, so the fix could be reverted silently. It now has a test, and the test is verified by execution rather than by assertion (see the negative control below).

Four cases, covering the whole contract rather than the happy path: a non-UTF-8 `HOME` (the case that is the entire difference between `var` and `var_os`, built with `OsStringExt::from_vec(b"/home/caf\xC3(\xFF".to_vec())`), the `USERPROFILE` fallback, `HOME` winning over `USERPROFILE`, and neither set.

Each case runs in a re-exec of the test binary rather than by mutating the process environment in place. That is not incidental: `commands/cloud_worker/tests.rs:488` and `:515` call `app_config::home_dir().expect("resolve user home")` in this same test binary, and libtest runs tests on parallel threads, so pointing `HOME` at invalid UTF-8 in-process would panic those two tests at random. The parent also asserts the child's stdout contains `1 passed`, so a mistyped filter that matches zero tests cannot pass vacuously.

### Negative control for that test

Run [32444401834](https://github.com/proliferate-ai/proliferate/actions/runs/32444401834), with `home_dir_os` deliberately reverted to `std::env::var` and everything else identical. One job red, `Cargo check & test`, and exactly one test in the workspace red:

```
---- app_config::home_env_tests::home_dir_os_reads_the_environment_without_utf8_validation stdout ----

thread 'app_config::home_env_tests::home_dir_os_reads_the_environment_without_utf8_validation' (32909) panicked at apps/desktop/src-tauri/src/app_config_home_env_tests.rs:96:5:
scenario `non_utf8_home` failed in the child:

thread 'app_config::home_env_tests::home_dir_os_reads_the_environment_without_utf8_validation' (32912) panicked at apps/desktop/src-tauri/src/app_config_home_env_tests.rs:64:28:
  left: None
 right: Some("/home/caf\xC3(\xFF")

test result: FAILED. 485 passed; 1 failed; 5 ignored; 0 measured; 0 filtered out; finished in 19.47s
error: test failed, to rerun pass `-p proliferate --lib`
```

`left: None` is the regression exactly: `var` returns `Err(VarError::NotUnicode)`, `.ok()` discards it, and `USERPROFILE` is unset. Note also that the same run reported `test result: ok. 2218 passed; 0 failed` for the anyharness crates, which is the reviewer's finding reproduced: nothing else in the workspace notices.

That commit was squashed out and is not in this branch. It exists only as the run linked above.

Existing coverage over touched code:
- `anyharness-lib/src/live/terminals/manager_tests.rs`
- `anyharness-lib/src/live/terminals/command_runs/pty_command_tests.rs`
- `anyharness-lib/src/live/terminals/driver/shell.rs` (inline `mod tests`; the file is now unchanged from base, so this is a no-op regression check)
- `proliferate-worker/src/anyharness_update.rs` (inline `mod tests`: `stop_runtime_script`, `pgrep_pattern_for_path`, `shell_single_quote`)
- `proliferate-worker/src/supervisor_bridge/bridge/tests.rs` (fake `BridgeHost`, so the rewritten `RealBridgeHost::supervisor_live` is not asserted against the old script text)
- `proliferate-supervisor/` config tests
- `apps/desktop/src-tauri/src/commands/diagnostics.rs` (inline `mod tests`)

`rustfmt --edition 2021 --check` was run over every touched file; no drift in any changed hunk.

## Windows CI result

`Windows compile check` run [32445075467](https://github.com/proliferate-ai/proliferate/actions/runs/32445075467) on this PR head, `cargo check --target x86_64-pc-windows-msvc` on `windows-latest`:

- `runtime crates (control)` (`-p anyharness -p proliferate-worker -p proliferate-supervisor`): job `success`, `cargo check` step `success`, 2m06s.
- `desktop crates (the unknown)` (`-p proliferate --keep-going`): job `success`, `cargo check` step `success`, 1m40s, no errors in the step summary.

Both jobs are `continue-on-error: true`, so those per-step conclusions were read from the jobs API rather than the check-run status, which reports `success` either way.

Baseline run [32438080854](https://github.com/proliferate-ai/proliferate/actions/runs/32438080854) on `ci/windows-compile-check` itself (PR #2142): both jobs `success`, both `cargo check` steps `success`. Windows was already compile-clean and this change keeps it clean. What it fixes is runtime behaviour, which no compiler can see.

## Full CI result

All 20 checks green on this head (run [32445075483](https://github.com/proliferate-ai/proliferate/actions/runs/32445075483)): `Cargo check & test`, `Repo shape checks`, `Build SDK`, `Desktop frontend build`, `Web frontend build`, `Shared frontend packages`, `Mobile typecheck`, `intent-tests`, `intent-billing`, `Workflow definition lifecycle (tier-2)`, `Transcript scroll physics (chromium + webkit)`, `Login first-load budget (phase-6)`, `Terraform validate (server/infra)`, `CI/CD config checks`, `Validate PR title and labels`, `Candidate build handoff`, `Vercel`, plus the two Windows jobs above.

All eight new tests ran and passed, confirmed from the job log rather than assumed:

```
test host_shell::tests::unix_invocation_is_exactly_bin_sh_dash_lc ... ok
test host_shell::tests::the_command_string_is_interpreted_by_a_shell_and_its_status_is_returned ... ok
test host_shell::tests::shell_syntax_in_the_command_string_is_honoured ... ok
test posix_shell::tests::programs_and_flags_are_the_original_literals ... ok
test posix_shell::tests::the_scripts_exit_status_is_returned_verbatim ... ok
test posix_shell::tests::a_requested_working_directory_is_applied ... ok
test posix_shell::tests::without_a_working_directory_the_script_inherits_the_process_cwd ... ok
test app_config::home_env_tests::home_dir_os_reads_the_environment_without_utf8_validation ... ok
```

The desktop lib suite on this head reports `test result: ok. 486 passed; 0 failed; 5 ignored; 0 measured; 0 filtered out; finished in 19.51s`, against `test result: FAILED. 485 passed; 1 failed; 5 ignored` on the negative-control run above. Same 486 tests, one flipped.

`Shared frontend packages` went red once on this head and passed on rerun of that job alone, with no code change. The failing test was `ComposerCommandEditor.test.tsx > uses the caret-local slash trigger without replacing trailing text` with `AssertionError: expected false to be true`, a Lexical caret and selection test. This branch changes no JS, TS or JSON at all (`git diff --stat ci/windows-compile-check...HEAD -- '*.ts' '*.tsx' '*.js' '*.json'` is empty), and the same job was `success` on the negative-control run whose tree differed only by `var` versus `var_os` in one Rust file, so it cannot be caused by this change.

`Repo shape checks` went red once on this revision and is fixed: a four-line comment pushed `ssh_tunnel.rs` from 729 to 730 lines, one past its recorded `max_lines` ratchet. The comment moved onto `app_config::home_dir_os`'s own doc, taking the file to 728, and `lints/frontend/ratchets.toml` is ratcheted down to match. `check_max_lines.py`, `lint_records.py`, `test_lint_records.py`, `check_anyharness_old_paths.py` and `check_proliferate_worker_structure.py` were all run locally before pushing.

On the earlier revision, `Cargo check & test` failed once and passed on rerun. The root failure was `fork_dispatch_and_restart_tests.rs:579` doing `std::fs::remove_dir_all(&runtime_home).expect("remove runtime home")` and hitting `Os { code: 39, kind: DirectoryNotEmpty }`, a teardown race in a test this change does not touch. That panic poisoned the crate-wide `test_support::lock_env` mutex, and the other 73 of the 74 reported failures are all the same cascading `expected env mutex` panic at `app/test_support.rs:152`. The first `FAILED` line printed (`cold_restart_refuses_process_local_fork_children_without_r1_proof`) is a poison victim, not the root; test output order is not causal order here. It has not recurred across the two subsequent runs.

### DIFF
--- a/.github/workflows/windows-compile-check.yml
+++ b/.github/workflows/windows-compile-check.yml
@@ -1,0 +1,69 @@
+name: Windows compile check
+
+# The runtime already builds and ships for Windows on every release, but nothing
+# has compiled the desktop crate for Windows since desktop releases were
+# disabled on 2026-04-13. That makes every Windows estimate guesswork. This job
+# turns the unknown into an error list.
+#
+# Non-blocking on purpose: it is a probe, not a gate, until the errors are fixed.
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "apps/desktop/src-tauri/**"
+      - "anyharness/crates/**"
+      - ".github/workflows/windows-compile-check.yml"
+
+permissions:
+  contents: read
+
+jobs:
+  runtime:
+    # Known green: release-runtime.yml ships these targets today. Here as the
+    # control, so a red desktop job cannot be blamed on the toolchain.
+    name: runtime crates (control)
+    runs-on: windows-latest
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: x86_64-pc-windows-msvc
+      - uses: Swatinem/rust-cache@v2
+      - name: cargo check
+        run: cargo check --target x86_64-pc-windows-msvc -p anyharness -p proliferate-worker -p proliferate-supervisor
+
+  desktop:
+    name: desktop crates (the unknown)
+    runs-on: windows-latest
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: x86_64-pc-windows-msvc
+      - uses: Swatinem/rust-cache@v2
+      - name: cargo check
+        # Keep going past the first failure so one run yields the whole list
+        # rather than one error at a time.
+        run: cargo check --target x86_64-pc-windows-msvc -p proliferate --keep-going 2>&1 | tee check.log
+        shell: bash
+      - name: Error summary
+        if: always()
+        shell: bash
+        run: |
+          echo "### Windows desktop compile errors" >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
+          grep -E '^error' check.log | sort | uniq -c | sort -rn | head -50 >> $GITHUB_STEP_SUMMARY || echo "no errors" >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
+          echo "total errors: $(grep -cE '^error' check.log || echo 0)" >> $GITHUB_STEP_SUMMARY
+          echo "files implicated:" >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
+          grep -oE '\-\-> [^:]+' check.log | sed 's/--> //' | sort -u | head -60 >> $GITHUB_STEP_SUMMARY || true
+          echo '```' >> $GITHUB_STEP_SUMMARY
+      - uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: windows-desktop-check-log
+          path: check.log
--- a/.github/workflows/windows-compile-check.yml
+++ b/.github/workflows/windows-compile-check.yml
@@ -14,6 +14,11 @@ on:
       - "apps/desktop/src-tauri/**"
       - "anyharness/crates/**"
       - ".github/workflows/windows-compile-check.yml"
+      - "Cargo.lock"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 permissions:
   contents: read
@@ -24,9 +29,10 @@ jobs:
     # control, so a red desktop job cannot be blamed on the toolchain.
     name: runtime crates (control)
     runs-on: windows-latest
+    timeout-minutes: 45
     continue-on-error: true
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: x86_64-pc-windows-msvc
@@ -37,9 +43,10 @@ jobs:
   desktop:
     name: desktop crates (the unknown)
     runs-on: windows-latest
+    timeout-minutes: 45
     continue-on-error: true
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: x86_64-pc-windows-msvc
@@ -62,7 +69,7 @@ jobs:
           echo '```' >> $GITHUB_STEP_SUMMARY
           grep -oE '\-\-> [^:]+' check.log | sed 's/--> //' | sort -u | head -60 >> $GITHUB_STEP_SUMMARY || true
           echo '```' >> $GITHUB_STEP_SUMMARY
-      - uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         if: always()
         with:
           name: windows-desktop-check-log
--- a/anyharness/crates/anyharness-lib/src/host_shell.rs
+++ b/anyharness/crates/anyharness-lib/src/host_shell.rs
@@ -1,107 +1,104 @@
-//! The one place that answers "which shell does this host use, and how is it
-//! handed a command string?".
+//! The one place that decides how a user-authored command string reaches a
+//! shell.
 //!
-//! Everything that used to hardcode `/bin/sh` asks here instead. The unix
-//! answers are exactly the literals the call sites used before, so unix
-//! behaviour is unchanged; the windows answers are the equivalents that
-//! actually exist there. Written as `cfg(unix)` / `cfg(not(unix))` function
-//! pairs to match how the rest of this crate handles platform differences
-//! (see `process_kill`, `integrations::agent_cli::executable::make_executable`).
+//! Scope is deliberately narrow: this covers the *piped* command runner in
+//! `live::terminals::command_runs::setup_process`, which hands a command
+//! string to a shell, captures stdout/stderr and waits for exit under a
+//! deadline. It does NOT cover the PTY command-run path, and must not be
+//! wired into it. That path writes a POSIX sentinel wrapper into the terminal
+//! (`. <script>; anyharness_code=$?; printf ...` in
+//! `command_runs::pty::build_pty_command_wrapper`) and waits for the sentinel
+//! to come back, so an interpreter that cannot speak POSIX would leave a run
+//! that never completes. That path already refuses such shells up front:
+//! `command_runs::pty::run_terminal_command` bails with
+//! `unsupported_terminal_shell` whenever `ShellKind::is_posix()` is false, and
+//! `detect_shell_kind` classifies anything that is not bash/zsh/sh as
+//! `ShellKind::Other`. Terminal shell detection is therefore left resolving
+//! POSIX paths only, so the failure stays at PTY spawn rather than becoming a
+//! terminal that opens and then rejects every command.
+//!
+//! Written as a `cfg(unix)` / `cfg(not(unix))` function pair to match how the
+//! rest of this crate handles platform differences (see `process_kill`,
+//! `integrations::agent_cli::executable::make_executable`,
+//! `integrations::mcp::capability_token::write_secret_file`).
 
-use std::ffi::OsString;
-
-/// Program plus the single flag that makes it read the *next* argument as a
-/// whole command string to interpret.
+/// A `tokio` command that will run `command` through the host's shell.
 ///
-/// Used for command strings that are genuinely user-supplied and genuinely
-/// want shell semantics (pipes, globs, `&&`, environment expansion). A fixed
-/// program with fixed arguments should be spawned directly instead of going
-/// through here.
-#[derive(Debug, Clone)]
-pub struct CommandStringShell {
-    pub program: OsString,
-    pub command_flag: &'static str,
-}
-
-/// Unix: `/bin/sh -lc <command>`, byte-for-byte what the setup-command runner
-/// spawned before this module existed.
+/// The caller still owns cwd, stdio, environment and process-group setup; this
+/// only fixes the interpreter, the flag, and how the command string is quoted.
+///
+/// Unix: `/bin/sh -lc <command>`, byte-for-byte the invocation the setup
+/// runner built before this module existed.
 #[cfg(unix)]
-pub fn command_string_shell() -> CommandStringShell {
-    CommandStringShell {
-        program: OsString::from("/bin/sh"),
-        command_flag: "-lc",
-    }
+pub fn command_string_shell(command: &str) -> tokio::process::Command {
+    let mut cmd = tokio::process::Command::new("/bin/sh");
+    cmd.arg("-lc").arg(command);
+    cmd
 }
 
 /// Windows: the interpreter named by `ComSpec` (`cmd.exe` when unset) with
-/// `/C`, which is the closest equivalent of `sh -c` that is guaranteed to be
-/// present. There is no login-shell notion, so `-lc` collapses to `/C`.
-#[cfg(not(unix))]
-pub fn command_string_shell() -> CommandStringShell {
-    CommandStringShell {
-        program: std::env::var_os("ComSpec").unwrap_or_else(|| OsString::from("cmd.exe")),
-        command_flag: "/C",
-    }
-}
-
-/// Absolute interpreter paths probed, in order, when a managed run needs a
-/// shell and the caller did not name one. Unix keeps the original list and
-/// order; windows has no such fixed paths, so the probe is empty and the
-/// caller falls through to [`last_resort_shell`].
-#[cfg(unix)]
-pub fn well_known_shell_paths() -> &'static [&'static str] {
-    &[
-        "/bin/bash",
-        "/usr/bin/bash",
-        "/bin/zsh",
-        "/usr/bin/zsh",
-        "/bin/sh",
-        "/usr/bin/sh",
-    ]
-}
-
-#[cfg(not(unix))]
-pub fn well_known_shell_paths() -> &'static [&'static str] {
-    &[]
-}
-
-/// Interpreter paths tried after `$SHELL` when detecting the host's default
-/// interactive shell. Unix keeps the original list and order.
-#[cfg(unix)]
-pub fn default_shell_fallbacks() -> &'static [&'static str] {
-    &["/bin/bash", "/usr/bin/bash", "/bin/sh", "/usr/bin/sh"]
-}
-
-#[cfg(not(unix))]
-pub fn default_shell_fallbacks() -> &'static [&'static str] {
-    &[]
-}
-
-/// What to spawn when nothing else resolved. Unix returns `/bin/sh`, the same
-/// literal the terminal driver returned before. Windows returns `ComSpec`,
-/// which is the only shell guaranteed to exist there.
-#[cfg(unix)]
-pub fn last_resort_shell() -> String {
-    "/bin/sh".to_string()
-}
-
-#[cfg(not(unix))]
-pub fn last_resort_shell() -> String {
-    std::env::var_os("ComSpec")
-        .map(|value| value.to_string_lossy().into_owned())
-        .unwrap_or_else(|| "cmd.exe".to_string())
-}
-
-/// Home directory of the current user, honouring `USERPROFILE` on windows
-/// where `HOME` is usually absent.
+/// `/C`, the closest guaranteed-present equivalent of `sh -c`. There is no
+/// login-shell notion, so `-lc` collapses to `/C`.
 ///
-/// `HOME` is consulted first and returned verbatim, so every unix host that
-/// sets it resolves exactly as before, empty value included. `USERPROFILE` is
-/// only reached when `HOME` is absent entirely, which is the case this exists
-/// to fix.
-pub fn home_dir_from_env() -> Option<std::path::PathBuf> {
-    if let Some(home) = std::env::var_os("HOME") {
-        return Some(std::path::PathBuf::from(home));
+/// The command string goes through `raw_arg`, not `arg`. `arg` applies the
+/// MSVCRT quoting rules on windows, which wrap anything containing a space in
+/// quotes; `cmd.exe` does not parse its argument that way, so a perfectly
+/// ordinary `npm run build --if-present` would arrive mangled. `raw_arg`
+/// passes the string through untouched, which is what `/C` expects.
+#[cfg(not(unix))]
+pub fn command_string_shell(command: &str) -> tokio::process::Command {
+    let program =
+        std::env::var_os("ComSpec").unwrap_or_else(|| std::ffi::OsString::from("cmd.exe"));
+    let mut cmd = tokio::process::Command::new(program);
+    cmd.arg("/C");
+    #[cfg(windows)]
+    {
+        use std::os::windows::process::CommandExt;
+        cmd.as_std_mut().raw_arg(command);
     }
-    std::env::var_os("USERPROFILE").map(std::path::PathBuf::from)
+    #[cfg(not(windows))]
+    cmd.arg(command);
+    cmd
+}
+
+#[cfg(all(test, unix))]
+mod tests {
+    use std::ffi::OsStr;
+
+    /// Negative control for the "unix behaviour is unchanged" claim: these are
+    /// the exact program and flag the setup runner hardcoded before this
+    /// module existed. Changing either should fail here.
+    #[test]
+    fn unix_invocation_is_exactly_bin_sh_dash_lc() {
+        let command = super::command_string_shell("printf ok");
+        let std_command = command.as_std();
+
+        assert_eq!(std_command.get_program(), OsStr::new("/bin/sh"));
+        assert_eq!(
+            std_command.get_args().collect::<Vec<_>>(),
+            vec![OsStr::new("-lc"), OsStr::new("printf ok")]
+        );
+    }
+
+    #[tokio::test]
+    async fn the_command_string_is_interpreted_by_a_shell_and_its_status_is_returned() {
+        let status = super::command_string_shell("exit 7")
+            .status()
+            .await
+            .expect("spawn the host shell");
+
+        assert_eq!(status.code(), Some(7));
+    }
+
+    #[tokio::test]
+    async fn shell_syntax_in_the_command_string_is_honoured() {
+        // Proves a shell really is interpreting the string rather than the
+        // string being spawned as a program name with arguments.
+        let status = super::command_string_shell("false || true")
+            .status()
+            .await
+            .expect("spawn the host shell");
+
+        assert!(status.success());
+    }
 }
--- a/anyharness/crates/anyharness-lib/src/host_shell.rs
+++ b/anyharness/crates/anyharness-lib/src/host_shell.rs
@@ -1,0 +1,107 @@
+//! The one place that answers "which shell does this host use, and how is it
+//! handed a command string?".
+//!
+//! Everything that used to hardcode `/bin/sh` asks here instead. The unix
+//! answers are exactly the literals the call sites used before, so unix
+//! behaviour is unchanged; the windows answers are the equivalents that
+//! actually exist there. Written as `cfg(unix)` / `cfg(not(unix))` function
+//! pairs to match how the rest of this crate handles platform differences
+//! (see `process_kill`, `integrations::agent_cli::executable::make_executable`).
+
+use std::ffi::OsString;
+
+/// Program plus the single flag that makes it read the *next* argument as a
+/// whole command string to interpret.
+///
+/// Used for command strings that are genuinely user-supplied and genuinely
+/// want shell semantics (pipes, globs, `&&`, environment expansion). A fixed
+/// program with fixed arguments should be spawned directly instead of going
+/// through here.
+#[derive(Debug, Clone)]
+pub struct CommandStringShell {
+    pub program: OsString,
+    pub command_flag: &'static str,
+}
+
+/// Unix: `/bin/sh -lc <command>`, byte-for-byte what the setup-command runner
+/// spawned before this module existed.
+#[cfg(unix)]
+pub fn command_string_shell() -> CommandStringShell {
+    CommandStringShell {
+        program: OsString::from("/bin/sh"),
+        command_flag: "-lc",
+    }
+}
+
+/// Windows: the interpreter named by `ComSpec` (`cmd.exe` when unset) with
+/// `/C`, which is the closest equivalent of `sh -c` that is guaranteed to be
+/// present. There is no login-shell notion, so `-lc` collapses to `/C`.
+#[cfg(not(unix))]
+pub fn command_string_shell() -> CommandStringShell {
+    CommandStringShell {
+        program: std::env::var_os("ComSpec").unwrap_or_else(|| OsString::from("cmd.exe")),
+        command_flag: "/C",
+    }
+}
+
+/// Absolute interpreter paths probed, in order, when a managed run needs a
+/// shell and the caller did not name one. Unix keeps the original list and
+/// order; windows has no such fixed paths, so the probe is empty and the
+/// caller falls through to [`last_resort_shell`].
+#[cfg(unix)]
+pub fn well_known_shell_paths() -> &'static [&'static str] {
+    &[
+        "/bin/bash",
+        "/usr/bin/bash",
+        "/bin/zsh",
+        "/usr/bin/zsh",
+        "/bin/sh",
+        "/usr/bin/sh",
+    ]
+}
+
+#[cfg(not(unix))]
+pub fn well_known_shell_paths() -> &'static [&'static str] {
+    &[]
+}
+
+/// Interpreter paths tried after `$SHELL` when detecting the host's default
+/// interactive shell. Unix keeps the original list and order.
+#[cfg(unix)]
+pub fn default_shell_fallbacks() -> &'static [&'static str] {
+    &["/bin/bash", "/usr/bin/bash", "/bin/sh", "/usr/bin/sh"]
+}
+
+#[cfg(not(unix))]
+pub fn default_shell_fallbacks() -> &'static [&'static str] {
+    &[]
+}
+
+/// What to spawn when nothing else resolved. Unix returns `/bin/sh`, the same
+/// literal the terminal driver returned before. Windows returns `ComSpec`,
+/// which is the only shell guaranteed to exist there.
+#[cfg(unix)]
+pub fn last_resort_shell() -> String {
+    "/bin/sh".to_string()
+}
+
+#[cfg(not(unix))]
+pub fn last_resort_shell() -> String {
+    std::env::var_os("ComSpec")
+        .map(|value| value.to_string_lossy().into_owned())
+        .unwrap_or_else(|| "cmd.exe".to_string())
+}
+
+/// Home directory of the current user, honouring `USERPROFILE` on windows
+/// where `HOME` is usually absent.
+///
+/// `HOME` is consulted first and returned verbatim, so every unix host that
+/// sets it resolves exactly as before, empty value included. `USERPROFILE` is
+/// only reached when `HOME` is absent entirely, which is the case this exists
+/// to fix.
+pub fn home_dir_from_env() -> Option<std::path::PathBuf> {
+    if let Some(home) = std::env::var_os("HOME") {
+        return Some(std::path::PathBuf::from(home));
+    }
+    std::env::var_os("USERPROFILE").map(std::path::PathBuf::from)
+}
--- a/anyharness/crates/anyharness-lib/src/lib.rs
+++ b/anyharness/crates/anyharness-lib/src/lib.rs
@@ -2,6 +2,7 @@ pub mod adapters;
 pub mod api;
 pub mod app;
 pub mod domains;
+pub mod host_shell;
 pub mod integrations;
 pub mod live;
 pub mod observability;
--- a/anyharness/crates/anyharness-lib/src/live/terminals/command_runs/setup_process.rs
+++ b/anyharness/crates/anyharness-lib/src/live/terminals/command_runs/setup_process.rs
@@ -62,13 +62,11 @@ pub(in crate::live::terminals) async fn run_setup_process(
     .await;
 
     // The setup command is a user-authored string that may use pipes, `&&`,
-    // globs and variable expansion, so it genuinely needs a shell. Which shell
-    // is the host's business, not this module's.
-    let shell = crate::host_shell::command_string_shell();
-    let mut cmd = tokio::process::Command::new(&shell.program);
-    cmd.arg(shell.command_flag)
-        .arg(&command)
-        .current_dir(&workspace_path)
+    // globs and variable expansion, so it genuinely needs a shell. Which shell,
+    // and how the string is quoted into it, is the host's business rather than
+    // this module's.
+    let mut cmd = crate::host_shell::command_string_shell(&command);
+    cmd.current_dir(&workspace_path)
         .stdout(std::process::Stdio::piped())
         .stderr(std::process::Stdio::piped());
     cmd.kill_on_drop(true);
--- a/anyharness/crates/anyharness-lib/src/live/terminals/command_runs/setup_process.rs
+++ b/anyharness/crates/anyharness-lib/src/live/terminals/command_runs/setup_process.rs
@@ -61,8 +61,12 @@ pub(in crate::live::terminals) async fn run_setup_process(
     )
     .await;
 
-    let mut cmd = tokio::process::Command::new("/bin/sh");
-    cmd.arg("-lc")
+    // The setup command is a user-authored string that may use pipes, `&&`,
+    // globs and variable expansion, so it genuinely needs a shell. Which shell
+    // is the host's business, not this module's.
+    let shell = crate::host_shell::command_string_shell();
+    let mut cmd = tokio::process::Command::new(&shell.program);
+    cmd.arg(shell.command_flag)
         .arg(&command)
         .current_dir(&workspace_path)
         .stdout(std::process::Stdio::piped())
--- a/anyharness/crates/anyharness-lib/src/live/terminals/driver/shell.rs
+++ b/anyharness/crates/anyharness-lib/src/live/terminals/driver/shell.rs
@@ -56,15 +56,12 @@ pub(super) fn detect_shell_kind(shell: &str) -> ShellKind {
     }
 }
 
+/// The shell a managed run (setup / run terminals) is spawned under. Named
+/// for the POSIX shells it prefers; on hosts without them the probe list is
+/// empty and this falls through to `detect_default_shell`, which resolves the
+/// host's own interpreter.
 pub(in crate::live::terminals) fn detect_posix_shell() -> String {
-    for candidate in [
-        "/bin/bash",
-        "/usr/bin/bash",
-        "/bin/zsh",
-        "/usr/bin/zsh",
-        "/bin/sh",
-        "/usr/bin/sh",
-    ] {
+    for &candidate in crate::host_shell::well_known_shell_paths() {
         if is_executable_command(candidate, std::env::var_os("PATH").as_deref()) {
             return candidate.to_string();
         }
@@ -73,8 +70,8 @@ pub(in crate::live::terminals) fn detect_posix_shell() -> String {
 }
 
 fn ensure_bash_prompt_rcfile() -> Option<String> {
-    let home = std::env::var_os("HOME")?;
-    let rcfile = Path::new(&home)
+    let home = crate::host_shell::home_dir_from_env()?;
+    let rcfile = home
         .join(".proliferate")
         .join("anyharness")
         .join("terminal")
@@ -117,7 +114,7 @@ fn detect_default_shell_with_env(shell_env: Option<&str>, path_env: Option<&OsSt
         candidates.push(shell);
     }
 
-    for fallback in ["/bin/bash", "/usr/bin/bash", "/bin/sh", "/usr/bin/sh"] {
+    for &fallback in crate::host_shell::default_shell_fallbacks() {
         if !candidates.contains(&fallback) {
             candidates.push(fallback);
         }
@@ -129,7 +126,7 @@ fn detect_default_shell_with_env(shell_env: Option<&str>, path_env: Option<&OsSt
         }
     }
 
-    "/bin/sh".to_string()
+    crate::host_shell::last_resort_shell()
 }
 
 fn is_executable_command(command: &str, path_env: Option<&OsStr>) -> bool {
--- a/anyharness/crates/anyharness-lib/src/live/terminals/driver/shell.rs
+++ b/anyharness/crates/anyharness-lib/src/live/terminals/driver/shell.rs
@@ -56,12 +56,15 @@ pub(super) fn detect_shell_kind(shell: &str) -> ShellKind {
     }
 }
 
-/// The shell a managed run (setup / run terminals) is spawned under. Named
-/// for the POSIX shells it prefers; on hosts without them the probe list is
-/// empty and this falls through to `detect_default_shell`, which resolves the
-/// host's own interpreter.
 pub(in crate::live::terminals) fn detect_posix_shell() -> String {
-    for &candidate in crate::host_shell::well_known_shell_paths() {
+    for candidate in [
+        "/bin/bash",
+        "/usr/bin/bash",
+        "/bin/zsh",
+        "/usr/bin/zsh",
+        "/bin/sh",
+        "/usr/bin/sh",
+    ] {
         if is_executable_command(candidate, std::env::var_os("PATH").as_deref()) {
             return candidate.to_string();
         }
@@ -70,8 +73,8 @@ pub(in crate::live::terminals) fn detect_posix_shell() -> String {
 }
 
 fn ensure_bash_prompt_rcfile() -> Option<String> {
-    let home = crate::host_shell::home_dir_from_env()?;
-    let rcfile = home
+    let home = std::env::var_os("HOME")?;
+    let rcfile = Path::new(&home)
         .join(".proliferate")
         .join("anyharness")
         .join("terminal")
@@ -114,7 +117,7 @@ fn detect_default_shell_with_env(shell_env: Option<&str>, path_env: Option<&OsSt
         candidates.push(shell);
     }
 
-    for &fallback in crate::host_shell::default_shell_fallbacks() {
+    for fallback in ["/bin/bash", "/usr/bin/bash", "/bin/sh", "/usr/bin/sh"] {
         if !candidates.contains(&fallback) {
             candidates.push(fallback);
         }
@@ -126,7 +129,7 @@ fn detect_default_shell_with_env(shell_env: Option<&str>, path_env: Option<&OsSt
         }
     }
 
-    crate::host_shell::last_resort_shell()
+    "/bin/sh".to_string()
 }
 
 fn is_executable_command(command: &str, path_env: Option<&OsStr>) -> bool {
--- a/anyharness/crates/proliferate-supervisor/src/config.rs
+++ b/anyharness/crates/proliferate-supervisor/src/config.rs
@@ -112,8 +112,12 @@ fn default_config_path() -> PathBuf {
     supervisor_state_dir().join("config.toml")
 }
 
+/// `HOME` first on every platform, so unix resolution is unchanged; windows
+/// has no `HOME` by default, so `USERPROFILE` is consulted before giving up
+/// and using the process working directory.
 fn dirs_fallback_home() -> PathBuf {
     std::env::var_os("HOME")
+        .or_else(|| std::env::var_os("USERPROFILE"))
         .map(PathBuf::from)
         .unwrap_or_else(|| PathBuf::from("."))
 }
--- a/anyharness/crates/proliferate-supervisor/src/install/layout.rs
+++ b/anyharness/crates/proliferate-supervisor/src/install/layout.rs
@@ -3,6 +3,10 @@ use std::path::PathBuf;
 pub fn default_home() -> PathBuf {
     std::env::var_os("PROLIFERATE_HOME")
         .map(PathBuf::from)
-        .or_else(|| std::env::var_os("HOME").map(|home| PathBuf::from(home).join(".proliferate")))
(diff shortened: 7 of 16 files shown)


pablonyx:
## Review: not safe to merge as-is

Two blocking findings. One is a real unix behaviour change that falsifies a "unix unchanged: yes" row, the other is a Windows regression created by the combination of what this PR changes and what it deliberately defers. Eleven of the thirteen rows do hold up under inspection, including the `pgrep` row, which I tried hard to break and could not. Separately, the PR is a draft stacked on an unlanded draft base (#2142), so it is not mergeable in the mechanical sense either.

---

## BLOCKING

### B1. `driver/shell.rs:126` plus the deferred `pty.rs` turns Windows fail-fast into hang-forever

`detect_default_shell_with_env` now ends in `crate::host_shell::last_resort_shell()`, which on non-unix returns `%ComSpec%` (default `cmd.exe`). That value is not decorative: `detect_posix_shell()` and `detect_default_shell()` feed `driver/pty.rs:64-66`, `live/terminals/manager.rs:202` and `live/terminals/command_runs/workspace_stop.rs:276`. Command runs on that PTY then write `build_pty_command_wrapper` (`pty.rs:472-477`) into it:

```
anyharness_prefix='__ANYHARNESS_CMD_'; anyharness_nonce='...'; printf '%s\n' "${anyharness_prefix}START_${anyharness_nonce}__"; . '<uuid>.sh'; anyharness_code=$?; printf ...
```

and the reader blocks until it sees `__ANYHARNESS_CMD_END_<nonce>_<code>__`.

Failure scenario on Windows. Before this PR: `CommandBuilder::new("/bin/sh")` fails at spawn, `driver/pty.rs:83` returns `failed to spawn shell: <file not found>`, the user gets an immediate, if ugly, error. After this PR: `cmd.exe` spawns successfully, echoes `'anyharness_prefix' is not recognized as an internal or external command` and similar for each fragment, and the END sentinel never arrives. The timeout task at `pty.rs:122` is only spawned `if let Some(timeout_ms) = request.timeout_ms`, so a run submitted without a timeout hangs indefinitely with the terminal apparently alive; with a timeout it sits until expiry and reports `TimedOut` / exit 124. Both are strictly harder to diagnose than the file-not-found the PR body cites as the motivating symptom ("the spawn returns file-not-found and gets reported as an unrelated setup failure").

The inconsistency is the tell. `posix_shell::run_script` refuses loudly on non-unix with a typed `Unsupported` naming the reason, which is the right call. `host_shell::last_resort_shell` does the opposite for a code path in the same crate that speaks a POSIX-only protocol into whatever it gets back. Either the non-unix terminal path should return an explicit unsupported error the same way, or `last_resort_shell()` must not be wired into the managed-run path until the `pty.rs` sentinel work lands. Deferring `pty.rs` is defensible on its own; deferring it while landing this substitution is not.

### B2. `diagnostics.rs:265` and `ssh_tunnel.rs:716`: unix behaviour IS changed, contrary to both rows

Both sites moved from `std::env::var_os("HOME")` to `crate::app_config::home_dir()`. That helper is `apps/desktop/src-tauri/src/app_config.rs:59`:

```rust
pub fn home_dir() -> Result<PathBuf, String> {
    std::env::var("HOME")
        .or_else(|_| std::env::var("USERPROFILE"))
        ...
}
```

`std::env::var` is the `String` variant: it returns `Err(NotUnicode)` for a value that is not valid UTF-8. `var_os` did not. On macOS and Linux a home directory path is an arbitrary byte string, and non-UTF-8 home paths exist in the wild (latin-1 accented usernames, ISO-8859 filesystems, restored backups).

Failure scenario. A Linux user whose `HOME` is the latin-1 bytes for `/home/josé`. Before: "Save diagnostics to `~/Desktop/report.json`" resolves correctly and writes the file. After: `var("HOME")` fails with `NotUnicode`, `var("USERPROFILE")` fails with `NotPresent`, and the user is told `HOME is not set` when `HOME` is very much set. Same for `expand_home_path` in the SSH tunnel command, where the affected value is an identity-file path and the misleading error will send someone hunting their shell profile.

The rows claim "yes, including the error string". The error string is preserved; the resolution is not. Fix is one line: make `app_config::home_dir` use `var_os` with a `var_os("USERPROFILE")` fallback, or keep `var_os` at the two call sites and add the `USERPROFILE` leg there. Whichever way, the two table rows need correcting.

---

## NON-BLOCKING

**N1. `bridge/mod.rs:304-311`, residue from removing the shell.** The rewrite is behaviour-equivalent (see the row verdict), but `pgrep_pattern_for_path`'s doc comment at `bridge/mod.rs:398-400` says the `[/]` escape exists "so the pgrep shell never matches its own command line". For this caller there is no longer a pgrep shell, and `pgrep` excludes its own pid regardless. The escape is now vestigial here and the comment is wrong for this call site. Harmless, but it is exactly the kind of stale invariant that gets "cleaned up" later by someone who does not realise the *other* two callers (`detached_supervisor_launch_script`, `stop_runtime_script`) still genuinely need it.

**N2. `bridge/mod.rs:304`, Windows produces a not-live/refuse loop rather than a clear answer.** On Windows `supervisor_live` returns `false` forever (spawn error), and `spawn_supervisor` then returns `Unsupported`. Since the whole point of this file is a liveness gate on the spawn, a `#[cfg(not(unix))]` early return saying "supervisor bridge is unix-only" would be more honest than making the gate lie and letting the next call fail. Same shape as what `posix_shell` already does correctly.

**N3. `posix_shell` error propagation is adequate but type-erased.** `anyharness_update.rs:369` yields `AnyharnessUpdateStop { detail: "failed to spawn stop shell: worker process supervision needs a POSIX shell..." }` and `bridge/mod.rs:332` the `BridgeSpawn` equivalent. The reason survives to the log, which is the important part, so this is not the silent-false failure mode I was checking for. But `ErrorKind::Unsupported` is indistinguishable from a transient spawn failure once it is inside a `detail: String`, so any retry policy over those two error variants will retry a permanently impossible operation forever. Worth confirming the worker's retry behaviour on `AnyharnessUpdateStop` before this ships to a Windows host.

**N4. Third home-resolution helper in the same crate, with different semantics from the second.** `host_shell::home_dir_from_env` (`var_os` HOME then `var_os` USERPROFILE, returns `Option`) now sits alongside `app::config::default_runtime_home` (`var` HOME then `var` USERPROFILE, `config.rs:59`) and `proliferate-worker/src/config.rs:106`. Three spellings of one question, and the `var` vs `var_os` split between them is precisely the bug in B2. Consolidate on the `var_os` form.

**N5. Neither new module has a single test, and the existing tests contradict the new Windows branch.** `host_shell.rs` and `posix_shell.rs` ship with no `mod tests`. Meanwhile `driver/shell.rs:179-196` asserts the detection result is one of `"/bin/bash" | "/usr/bin/bash" | "/bin/sh" | "/usr/bin/sh"`, uncgated. On Windows those two tests now fail by construction, because the function returns `cmd.exe`. `cargo check` does not build `#[cfg(test)]`, so the Windows compile-check job that the PR body leans on cannot see this. The Windows behaviour introduced here has zero executed coverage anywhere.

**N6. `cmd.exe /C` plus `Command::arg` is the wrong pairing for a user-authored string.** `setup_process.rs:69` passes the user's setup command through `.arg()`, which applies MSVCRT argument-quoting rules. `cmd.exe` does not parse its command line by those rules. Strings containing `&`, `|`, `^`, `>`, `%VAR%` or embedded quotes will be mangled or re-interpreted (the BatBadBut / CVE-2024-24576 class). If the Windows leg is meant to be real rather than nominal, this needs `std::os::windows::process::CommandExt::raw_arg`.

**N7. House-style claim: accurate.** I checked all three citations. `process_kill.rs:349/352` and `:357` are genuine `#[cfg(unix)]` / `#[cfg(not(unix))]` function pairs, and `integrations/agent_cli/executable.rs:54` / `:62` is the exact `make_executable` pattern claimed. The two new modules do match the crate idiom.

---

## The 13 rows, verified individually

| # | site | claim | verdict |
| --- | --- | --- | --- |
| 1 | `setup_process.rs:64` | unix unchanged | **CONFIRMED.** Unix branch is `/bin/sh` with `command_flag: "-lc"`, one fused arg exactly as before, so it is still a login shell sourcing profile files. Argv shape, `current_dir(&workspace_path)`, both piped stdio, `kill_on_drop(true)` and `#[cfg(unix)] cmd.process_group(0)` are all byte-identical and in the same order. |
| 2 | `shell.rs:57-70` `detect_posix_shell` | unix unchanged | **CONFIRMED.** `well_known_shell_paths()` is `/bin/bash, /usr/bin/bash, /bin/zsh, /usr/bin/zsh, /bin/sh, /usr/bin/sh`, byte-identical and in the same order. Same `is_executable_command(candidate, PATH)` test, same `detect_default_shell()` fallthrough. |
| 3 | `shell.rs:120` fallbacks | unix unchanged | **CONFIRMED.** `default_shell_fallbacks()` is `/bin/bash, /usr/bin/bash, /bin/sh, /usr/bin/sh`, same order. `$SHELL` is still consulted first, still filtered on `!value.trim().is_empty()`, still deduped via `candidates.contains(&fallback)`. |
| 4 | `shell.rs:132` last resort | unix unchanged | **CONFIRMED on unix**, literal `/bin/sh`. The non-unix half of this same function is B1. |
| 5 | `shell.rs:76` `ensure_bash_prompt_rcfile` | HOME verbatim first | **CONFIRMED.** Still `var_os`, so a non-UTF-8 `HOME` still works here (unlike B2), and an empty `HOME` still produces the relative `.proliferate/anyharness/terminal/bashrc` that `Path::new("").join(..)` produced. Only the HOME-absent case newly reaches `USERPROFILE`, as disclosed. |
| 6 | `supervisor/config.rs:116` | unix unchanged | **CONFIRMED.** `var_os("HOME")` still wins, `.` still the final fallback. |
| 7 | `supervisor/install/layout.rs:6` | unix unchanged | **CONFIRMED.** `PROLIFERATE_HOME` then `HOME` keep priority; `USERPROFILE` is nested inside the `HOME` leg so it cannot pre-empt either. |
| 8 | `bridge/mod.rs:298-305` `pgrep` | unix unchanged | **CONFIRMED**, and this is the row I attacked hardest. Quoting: `shell_single_quote` is a faithful `'` to `'\''` escaper, and single quotes suppress nothing that argv passing would reintroduce, so `pgrep` receives a byte-identical `[/]`-prefixed pattern either way. There is no expansion the shell would have performed: `pgrep_pattern_for_path` emits `[/]` plus a path, and inside single quotes `[`, `]`, `*`, `$` and `~` are all literal, exactly as in argv. Regex semantics of `-f` are unchanged because the string is unchanged. Exit codes: 0 match, 1 no match, 2 usage, 3 fatal all reached the caller as the shell's own status before and as `pgrep`'s status now, and all map through the same `status.success()`. Missing `pgrep`: 127 from the shell before, spawn `Err` now, both `false` via `unwrap_or(false)`, as claimed. Redirects to `/dev/null` map exactly onto `Stdio::null()`; stdin stays inherited in both. The one difference that could have existed, a surviving `sh` process whose own cmdline matches the pattern, cannot occur: both dash and bash exec-replace on a single simple command with redirections, and the `[/]` escape already made the literal pattern text non-self-matching. See N1 and N2 for the residue. |
| 9 | `bridge/mod.rs:319` `bash -lc` | unix unchanged | **CONFIRMED.** `ScriptShell::BashLogin` resolves to program `"bash"`, flag `"-lc"`, `cwd: None` matching the original absence of `current_dir`. Script string and `BridgeSpawn` detail text unchanged. |
| 10 | `anyharness_update.rs:369` `sh -c` | unix unchanged | **CONFIRMED.** `ScriptShell::Sh` resolves to `"sh"` and `"-c"`, `cwd: None`, same `AnyharnessUpdateStop` variant and same `failed to spawn stop shell: {error}` detail. |
| 11 | `anyharness_update.rs:439` relaunch | unix unchanged incl. cwd | **CONFIRMED.** `Some(workdir)` reaches `command.current_dir(cwd)`. The builder call order differs (cwd is now applied after the args rather than before), which has no observable effect. Same variant, same detail, same `path` field. |
| 12 | `diagnostics.rs:265` | unix unchanged incl. error string | **DENIED.** See B2: `var_os` to `var` changes resolution for non-UTF-8 `HOME`. |
| 13 | `ssh_tunnel.rs:716` | unix unchanged incl. error string | **DENIED.** See B2, identical cause. |

## The CI flake: diagnosis correct, cause genuinely unrelated, count understated

I pulled the attempt-1 log directly (job 96646190688) rather than take the body's word for it.

- `test result: FAILED. 2141 passed; 74 failed`.
- Exactly one non-poison panic in the entire run: `fork_dispatch_and_restart_tests.rs:579:44`, `remove runtime home: Os { code: 39, kind: DirectoryNotEmpty }`. That test holds `let _env_lock = test_support::lock_env();` taken at the top of `concurrent_double_fork_on_the_same_key_never_duplicates_the_child`, so unwinding through the guard poisons `ENV_MUTEX`.
- The other 73 are all poison cascades: 39 `env mutex: PoisonError` at `app/test_support.rs:152`, the rest the same panic under other test names plus one `unwrap()` on `PoisonError` at `live/sessions/actor/tests/conditional_cancel/unload.rs:289`.

So the mechanism is right. Two corrections. First, the body says "40-odd" cascading failures; it is 73 of 74, which matters because it is the difference between "a flake plus some other stuff" and "one flake, nothing else". Second, the first `... FAILED` line printed is `cold_restart_refuses_process_local_fork_children_without_r1_proof`, which is a poison *victim*, not the root; anyone re-reading that log casually will misattribute it.

Unrelated to this diff: yes, and I checked the one plausible coupling rather than assuming. The only new filesystem write this diff can introduce is `ensure_bash_prompt_rcfile` creating `$HOME/.proliferate/anyharness/terminal/bashrc`, and it only newly fires when `HOME` is absent and `USERPROFILE` is present. Linux CI sets `HOME` and does not set `USERPROFILE`, so that branch is unreachable there. Nothing else in the diff is reachable from fork dispatch: the changed anyharness-lib surfaces are the terminal driver and the setup-command runner, neither of which the fork tests touch. The teardown race is a live child process still writing under `runtime_home` when `remove_dir_all` walks it, which this change cannot influence.

Worth its own PR, not this one: `lock_env()`'s `.expect("expected env mutex")` converts any single teardown panic into a crate-wide blackout. The green rerun is real, but the red attempt proved nothing about anything else, and that is a permanent regression-hiding mechanism. `.unwrap_or_else(|poisoned| poisoned.into_inner())` would keep the 73 signals.

## Deferrals

- `pty.rs:442-478`: defensible as a unit of work, but see B1. Landing the shell substitution while deferring the protocol is what creates the regression, and the body does not disclose that interaction. This is the one disclosure gap.
- `agent_cli/launcher.rs:86`: defensible and accurately described. `make_executable` is indeed already a non-unix no-op (`executable.rs:62`).
- `sidecar.rs:176-178`: verified and accurately described. All six candidates lack `.exe`, the repo-relative ones come from a compile-time `env!("CARGO_MANIFEST_DIR")`, and `var("HOME").unwrap_or_default()` currently yields the harmless nonexistent `/.cargo/bin/anyharness` on Windows. Fixing `HOME` alone genuinely would not help.
- `catalog_probe_auth.rs:139,209`: verified as developer probe tooling under `crates/anyharness/src/commands/catalog_probe/`, not the shipped runtime path. Defensible.

The "found clean" sweep also checks out on the samples I re-ran: the remaining `"sh"` literal I found in `adapters/processes/operations/run.rs:148` is inside an inline `mod tests`, not production code.

## What I could not verify

- No compile and no test run. This machine cannot take a cargo build, so every finding above is source reading, not execution. In particular B2 is reasoned from `std::env::var` semantics, not demonstrated with a non-UTF-8 `HOME`.
- No Windows host. `%ComSpec%` resolution, `cmd.exe` command-line parsing, and Rust's `Command::arg` escaping under B1, N6 are reasoned about rather than observed. B1's specific claim that `cmd.exe` spawns successfully and then never emits the sentinel is a prediction.
- Whether the worker retries `AnyharnessUpdateStop` and `BridgeSpawn` indefinitely, which is what would turn N3 from a cosmetic point into a hot loop.
- Whether `pgrep` is actually present on the sandbox images the worker runs on. It does not change the row-8 verdict, since old and new behave identically when it is absent, but it does mean the liveness gate may have been silently answering "not live" in production all along.
- The base branch `ci/windows-compile-check` (#2142) itself. I reviewed the PR-vs-base diff only.
- The two Windows compile-check job conclusions in run 32439147764 were taken at face value from the body. I independently verified only the `Cargo check & test` claim.

pablonyx:
## Round 2 review: one blocking item, everything else checks out

B1 is closed per the prior ruling and not relitigated here.

### Verdict

**Not clean to merge as-is.** The `home_dir_os` fix at `apps/desktop/src-tauri/src/app_config.rs:73-77` is correct, but it has **zero test coverage anywhere in this diff** — verified by negative control, not by inspection. This is the same failure mode flagged elsewhere in this batch (a fix that passes CI whether or not the mechanism exists). Everything else below (tests that do exist, `raw_arg`, the ratchet, CI) checks out by execution.

---

### 1. The `HOME` regression fix — correct, semantics genuinely preserved

Verified by reading the diff directly (`git diff origin/ci/windows-compile-check...HEAD`):

- `app_config.rs:73-77` adds `home_dir_os()`: `var_os("HOME").or_else(|| var_os("USERPROFILE"))`. This is not routed through the old `home_dir()` — the diff shows both `expand_home_path` sites (`diagnostics.rs:264`, `ssh_tunnel.rs:716`) went straight from a bare `std::env::var_os("HOME")` to `app_config::home_dir_os()`, which is the same call plus a `USERPROFILE` fallback leg. Both call sites confirmed with `grep`:
  ```
  apps/desktop/src-tauri/src/commands/diagnostics.rs:269:  let home = || crate::app_config::home_dir_os().ok_or_else(...)
  apps/desktop/src-tauri/src/commands/ssh_tunnel.rs:719:   let home = || crate::app_config::home_dir_os().ok_or_else(...)
  ```
  So for `HOME`, resolution is bit-for-bit what it was before on unix (verbatim `OsString`, non-UTF-8 included), and the only new behavior is the `USERPROFILE` leg, which unix never reaches. Semantics are genuinely preserved, not approximately.

- `home_dir()` (`app_config.rs:58-63`, still `var`/`var`) is untouched, confirmed by the diff showing no hunk there. This does leave two functions in the same crate that disagree on non-UTF-8 `HOME`: `home_dir()` fails, `home_dir_os()` succeeds. It's disclosed in `home_dir_os`'s own doc comment (`home_dir`'s callers "build the app directory and already accept that [String] constraint"), and I checked those callers (`app_dir_path`, `diagnostics_collector/broker/discovery.rs`, `diagnostics/scrub.rs`, `support_snapshot/coordinator/finish.rs`) — they all do go on to build `String`-typed paths or JSON-serialized config, so the constraint claim holds. Non-blocking, but it's a real, permanent inconsistency, not just a stylistic one — same shape as round 1's N4 (three home-lookup spellings in this codebase now, two `var`, two `var_os`).

### 2. The seven tests — real for `host_shell`/`posix_shell`, but the actual regression fix has none

I did not take the "seven tests" claim on trust. Negative-control results, run in an isolated worktree off PR head `541e40d8e7`, not the author's worktree:

**Shell-selection test — correctly pins the mechanism.** Baseline:
```
test host_shell::tests::unix_invocation_is_exactly_bin_sh_dash_lc ... ok
test result: ok. 3 passed; 0 failed
```
Broke it (`cmd.arg("-lc")` → `cmd.arg("-c")` in `host_shell.rs:35`):
```
thread 'host_shell::tests::unix_invocation_is_exactly_bin_sh_dash_lc' panicked at anyharness/crates/anyharness-lib/src/host_shell.rs:77:9:
  left: ["-c", "printf ok"]
 right: ["-lc", "printf ok"]
test result: FAILED. 2 passed; 1 failed
```
Restored, `git status` clean. This one is real.

**`home_dir_os` — no test exists, and none catches the regression it fixes.** There is no `mod tests` in `ssh_tunnel.rs` at all, and `app_config.rs`'s and `diagnostics.rs`'s existing `mod tests` blocks contain nothing that calls `home_dir_os` or `expand_home_path` (checked with `grep -n "expand_home_path\|home_dir_os\|#\[test\]"` against both files). The PR body lists `diagnostics.rs`'s `mod tests` under "existing coverage over touched code" — that's not accurate; those tests (`renderer_ingest_*`) are unrelated to the home-path code.

To confirm rather than infer, I reintroduced the exact B2 regression inside `home_dir_os` itself (swapped `var_os` for `var`, i.e. the bug this PR exists to fix) and ran every test that could plausibly cover it:
```
$ cargo test -p proliferate --lib -- --test-threads=1 app_config diagnostics::tests ssh_tunnel
running 13 tests
test app_config::tests::... ok  (x10)
test commands::diagnostics::tests::... ok  (x3)
test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 583 filtered out
```
Zero failures. The fix that round 1 called the real blocking finding can be silently reverted today and every check stays green. Restored, `git status` clean.

This is exactly the pattern flagged as a known failure mode in this batch (a sibling PR shipping tests that pass with the mechanism deleted) — it just took the form of "no test" rather than "a test that doesn't test anything." Given that precedent, I'd block on this: at minimum, `home_dir_os` needs a test that injects `HOME`/`USERPROFILE` rather than mutating the real process environment (the codebase already has the pattern to copy — `driver/shell.rs`'s `detect_default_shell_with_env(shell_env: Option<&str>, path_env: Option<&OsStr>)` takes env values as parameters specifically so tests don't need a process-wide env lock; `app_config`/`expand_home_path` have no such seam or env-mutex today).

I did not additionally re-verify the four `posix_shell` tests or the cwd/exit-status host_shell tests by breaking them; the one I did break behaved as claimed, and I have no reason to doubt the other six given they ran green in the actual CI log (see CI section). I'm flagging this as a gap in what to trust, not a claim that those six are also fake.

### 3. `raw_arg` for `cmd.exe /C` — correct, and nothing re-quotes after it

`host_shell.rs:48-62`: the Windows arm does `cmd.arg("/C")` then, under `#[cfg(windows)]`, `cmd.as_std_mut().raw_arg(command)` for the command string itself; only the non-Windows-but-not-unix fallback (`#[cfg(not(windows))]`, unreachable in this repo's actual target set) uses `.arg(command)`. `raw_arg` is the correct API here — `cmd.exe` does its own command-line parsing rather than consuming argv the way `CreateProcess`/MSVCRT-quoted args are meant to be consumed, so quoting the string via `.arg()` would embed literal quote characters `cmd.exe` doesn't strip the way a normal argv-consumer would.

Traced the one call site (`setup_process.rs:68`, the only caller of `command_string_shell` in the repo, confirmed by `grep`) forward to `cmd.spawn()` at `setup_process.rs:83`: the only calls in between are `.current_dir()`, `.stdout()`, `.stderr()`, `.kill_on_drop()`, `#[cfg(unix)] .process_group(0)`, and a loop of `.env()` calls. None of these touch the argument list, so nothing re-quotes the raw arg after it's set. Verified by reading, not by running on a Windows host — there is no Windows host available to me either.

### 4. The reachability claim — true, but I could not find it in the PR body

**The claim itself checks out.** Traced both callers of `run_setup_process` (the only consumer of `host_shell::command_string_shell`):
- `manager.rs:194-214` (`start_setup_command`): `find_live_setup_terminal` can only return `Some` if a `Setup`-purpose terminal was created previously; if `None` it calls `create_terminal_shell(...).await?` with `shell: Some(detect_posix_shell())`.
- `workspace_stop.rs:268-287` (`run_blocking_command_for_workspace_inner`): calls `create_terminal_shell(...).await?` unconditionally, same `shell: Some(detect_posix_shell())`.

`detect_posix_shell()` (`driver/shell.rs:59-73`, confirmed byte-identical to base branch) only ever returns literal POSIX paths (`/bin/bash`, `/bin/sh`, etc.), falling through to `detect_default_shell()` whose final fallback (`shell.rs:132`) is the hardcoded string `"/bin/sh"`. `create_terminal_shell` (`driver/pty.rs:71-83`) does `CommandBuilder::new(&shell)` and `pair.slave.spawn_command(cmd).map_err(|e| anyhow!("failed to spawn shell: {e}"))?`. On Windows, no such path exists, so this spawn fails and the `?` at both call sites propagates the error before `run_setup_process` — and therefore `host_shell::command_string_shell`'s Windows arm and the `raw_arg` fix — is ever reached. Since a live `Setup` terminal can never exist on Windows (creation always fails), `find_live_setup_terminal` can never short-circuit around this. This is unreachable in production today, on my reading, and I could not break this conclusion.

**What I could not confirm: that the PR body says so.** I read the full PR body (`gh pr view 2148 --json body`) and all issue comments and searched for "unreachable," "reach," "both callers," "create_terminal_shell," "PTY terminal" — none of that appears anywhere. The body's row for `setup_process.rs:64` in the sites-changed table, and the "Sites changed" / "Left for other agents" / "Found and deliberately not changed" sections, present the Windows quoting fix at face value, with no caveat that it's currently dead code pending the `pty.rs` Windows work. So: the underlying fact is true and I verified it myself, but the specific assertion that "the PR body says so" does not hold — it isn't disclosed. That's worth fixing in the body regardless of what happens with the test-coverage item, for the same reason round 1 dinged the original PR for an undisclosed interaction: a reader who doesn't independently trace both call graphs would reasonably believe this fix is live on Windows today.

The adjacent claim — "the Windows verdict is a compile verdict and the runtime claims rest on source reading" — **is** accurately represented in the body ("That branch carries the Windows `cargo check` workflow, which is the only compile verdict available for this change" / "What it fixes is runtime behaviour, which no compiler can see"), and matches what I found in the actual Windows CI jobs (see below): both are `cargo check` only, no test execution.

### 5. The ratchet — value matches reality, `check_max_lines.py` run directly

```
$ wc -l apps/desktop/src-tauri/src/commands/ssh_tunnel.rs
     728 ...
$ grep -n "ssh_tunnel" -A2 lints/frontend/ratchets.toml
path = "apps/desktop/src-tauri/src/commands/ssh_tunnel.rs"
max_lines = 728
```
Confirmed the "bidirectional" claim isn't decorative: bumped the ratchet to a stale-high 729 (one more than the real 728) and reran:
```
Stale max-lines ratchet entries — these files are now within their threshold (or gone); shrink or delete the entries in this change:
  apps/desktop/src-tauri/src/commands/ssh_tunnel.rs ratchet=729 observed=728 max=600 (lints/frontend/ratchets.toml)
EXIT:1
```
Reverted to 728, reran, exit 0, `git status` clean. The ratchet is correct, not a workaround.

### 6. CI — confirmed from the runs, not the claim

```
$ gh pr checks 2148
```
lists all 20 checks `pass`, and:
```
$ gh api repos/proliferate-ai/proliferate/actions/runs/{32442398420,32442398414,32442398387,32443164213} --jq '{run_attempt, head_sha, conclusion}'
```
all four workflows show `"run_attempt": 1`, `"conclusion": "success"`, `"head_sha": "541e40d8e7a000d9baadeb50391056c8b792e4ca"` (current PR head). No reruns on this head.

Pulled the `Cargo check & test` job log directly and grepped for the seven new tests:
```
test host_shell::tests::shell_syntax_in_the_command_string_is_honoured ... ok
test host_shell::tests::the_command_string_is_interpreted_by_a_shell_and_its_status_is_returned ... ok
test host_shell::tests::unix_invocation_is_exactly_bin_sh_dash_lc ... ok
test posix_shell::tests::programs_and_flags_are_the_original_literals ... ok
test posix_shell::tests::a_requested_working_directory_is_applied ... ok
test posix_shell::tests::the_scripts_exit_status_is_returned_verbatim ... ok
test posix_shell::tests::without_a_working_directory_the_script_inherits_the_process_cwd ... ok
```
All present, all `ok`, and no `FAILED` line anywhere in that job's log. The two Windows compile-check jobs (`runtime crates (control)`, `desktop crates (the unknown)`) both show `conclusion: success` with their `cargo check` step `success` via the jobs API, matching the body's claim and confirming it's `cargo check` only — no test step.

### What I could not verify

- No Windows host, so `raw_arg`/`cmd.exe` behavior (item 3) and the spawn-failure mechanics underlying the reachability claim (item 4) are verified by reading, not by running, same limitation round 1 disclosed.
- I did not attempt to execute the four `posix_shell` tests or two of the six `host_shell` tests as negative controls myself (see item 2) — I'm relying on the CI log for those, not an independent break-and-restore.
- No compile/build of my own beyond the scoped `cargo test` invocations above; no full workspace build.

Working tree: all edits during this review were made in an isolated worktree at `/tmp/review-2148/pr-head` (detached at the PR head), not the author's worktree at `.claude/worktrees/win-posix`, which was not touched. `git status` in the review worktree is clean except for `Cargo.lock` (produced by the test builds; harmless, scratch worktree).

pablonyx:
## Post-rebase review — verdict: safe to merge as-is (once #2142 lands and this retargets to `main`)

Reviewed at head `6cbbfc541` (base `68777b561`), no fixes applied.

**1. Rebase integrity.** `git range-diff 8de5a45697..e5fe4f0735 68777b5614..6cbbfc541` shows all 4 commits as `=` — byte-identical patches, only the base moved. `merge-base --is-ancestor 68777b5614 6cbbfc5419` confirms the branch sits directly on the new `ci/windows-compile-check` tip. Clean rebase, nothing to relitigate here.

**2. CI at new head.** All checks green: 33 pass, 2 legitimately `skipping` (`docker`, `self-hosted-release-assets`), 0 failed — including both Windows jobs (`runtime crates (control)`, `desktop crates (the unknown)`). Note the rebased base's `Cargo check & test` now runs via `cargo-nextest` rather than the plain `cargo test` harness used in the two evidence runs below; nextest independently confirms `home_dir_os_reads_the_environment_without_utf8_validation` PASS among 3155 tests run, 0 failed. No regression from the harness swap.

**3. Round-1 `%ComSpec%`/PTY-hang claim: confirmed false, confirmed reverted, not relitigating.** `driver/shell.rs` is byte-identical to base (empty diff). Verified independently: `pty.rs:59` bails `unsupported_terminal_shell` when `!h.shell_kind.is_posix()`, `build_pty_command_wrapper` has its one call site at `pty.rs:92` (thirty-odd lines after the bail), and `ShellKind::is_posix()` (`domains/terminals/model.rs:25`) is `matches!(Bash | Zsh | Sh)`. `cmd.exe` classifies as `Other` and never reaches the wrapper. The revert stands on the narrower posture ground, which is fine.

**4. Round-2 `home_dir_os` coverage: verified end to end.**
- Test file `apps/desktop/src-tauri/src/app_config_home_env_tests.rs` re-execs the test binary per scenario (`Command::new(current_exe())`) rather than mutating env in-process — correctly motivated, since `commands/cloud_worker/tests.rs:488` and `:515` both call `app_config::home_dir().expect(...)` in the same binary under parallel libtest threads.
- Vacuous-pass guard confirmed: parent asserts `stdout.contains("1 passed")` before trusting a child exit-0.
- Negative-control run **32444401834**: conclusion `failure`, job `Cargo check & test` failed. Log shows `test result: ok. 2218 passed; 0 failed` for anyharness, then desktop lib `test result: FAILED. 485 passed; 1 failed`, with the single failure being exactly `app_config::home_env_tests::home_dir_os_reads_the_environment_without_utf8_validation` (`left: None, right: Some("/home/caf\xC3(\xFF")`) — the `var`-vs-`var_os` regression, reproduced.
- Fix-restored run **32445075483**: conclusion `success`, same job passes, log shows `test result: ok. 486 passed; 0 failed`. Same 486 tests, one flipped, exactly as claimed.

**5. `setup_process.rs` disclosure: present and accurate.** The PR body states plainly (under "What this fixes on Windows today, and what it does not") that this change is groundwork, not reachable on Windows today. Verified both callers: `manager.rs:194-214` and `workspace_stop.rs:268-287` both pass `shell: Some(detect_posix_shell())` into `create_terminal_shell`, and `detect_posix_shell` (`driver/shell.rs`) only probes `/bin/{bash,zsh,sh}` and `/usr/bin/{bash,zsh,sh}` paths, falling back to the literal `/bin/sh` — none of which exist on Windows, so PTY spawn fails before `run_setup_process`/`host_shell` is ever entered.

**6. Frontend flake: disclosed, and the no-frontend-files claim holds.** `git diff --stat 68777b5614..6cbbfc5419 -- '*.ts' '*.tsx' '*.js' '*.json'` is empty. Full file list touched is 14 Rust files plus `lints/frontend/ratchets.toml` (a ratchet-count edit, not JS/TS/JSON). The one-time `Shared frontend packages` red-then-green-on-rerun is recorded in the PR body with the specific failing test and the reasoning for why this branch can't have caused it. Nothing further to chase.

No blocking findings. Leaving as draft per instructions — not merging, not relabeling.

pablonyx (commented):
Confirming pass on the one commit that landed since the last review round: `67b8f16a6` "test(desktop): pin home_dir_os against the var-vs-var_os regression" (`apps/desktop/src-tauri/src/app_config.rs` +4, `apps/desktop/src-tauri/src/app_config_home_env_tests.rs` +106, new file). Scope is limited to this commit only; earlier rounds are not relitigated.

**Verdict: non-blocking.** The commit is what it claims to be — test-only wiring that pins a real regression, with an actual negative control, safely isolated from cross-test env races, and green in CI. Two informational notes below, neither blocking.

### 1. The `app_config.rs` +4 is pure test wiring — confirmed
```rust
#[cfg(all(test, unix))]
#[path = "app_config_home_env_tests.rs"]
mod home_env_tests;
```
Gated on both `test` and `unix`, inserted between two existing `mod` blocks with no other change to runtime code in this commit. No behavior change smuggled in.

### 2. Regression is real and correctly targeted
`home_dir_os` (pre-existing in this PR, not touched by this commit) already reads `var_os("HOME").or_else(|| var_os("USERPROFILE"))`, with a doc comment giving the exact rationale the new test file restates: `std::env::var` returns `VarError::NotUnicode` on a non-UTF-8 value, which would turn a valid `~/...` expansion into a spurious "HOME is not set." I verified the "two `expand_home_path` callers" claim in the test file's header comment against the source — both are real:
- `apps/desktop/src-tauri/src/commands/diagnostics.rs:269`
- `apps/desktop/src-tauri/src/commands/ssh_tunnel.rs:719`

### 3. Real negative control, but only one of the four scenarios is it
The single `#[test]` function bundles four scenarios via re-exec: `non_utf8_home`, `userprofile_fallback`, `home_wins_over_userprofile`, `neither_set`. Only **`non_utf8_home`** is the actual var-vs-var_os regression guard — the other three use valid UTF-8 strings, so they'd pass identically whether `home_dir_os` used `var` or `var_os`. That's fine (they cover fallback order/precedence/absent-case correctness, which is legitimate coverage), but only `non_utf8_home` is load-bearing for the stated regression: reverting `var_os` to `var` there makes `home_dir_os()` return `None` (both `HOME` and `USERPROFILE` unset in that sub-case), while the test asserts `Some(...)`, so it fails correctly. Confirmed by reading the implementation, not by running it.

### 4. Env mutation is isolated, not racy
The design deliberately avoids mutating the parent test process's environment: each scenario re-execs `std::env::current_exe()` as a child with `--exact --nocapture <test path>` and the target `HOME`/`USERPROFILE` set only in the child. The stated reason — `apps/desktop/src-tauri/src/commands/cloud_worker/tests.rs` calls `app_config::home_dir().expect(..)` on two paths in the same binary (confirmed at lines 488 and 515) while libtest runs tests on parallel threads — is accurate and the re-exec avoids that race entirely (no shared-process `set_var`/`remove_var` at all). There's also a second guard against a silently-vacuous run: it asserts the child's stdout contains `"1 passed"`, so a mistyped filter matching zero tests fails loudly instead of exiting 0.

Minor, non-blocking note: CI runs this crate under cargo-nextest, which already isolates each test in its own process, so the specific race being defended against may not be reachable under nextest today. The re-exec is still correct and is what actually protects a plain `cargo test` invocation (shared-process, multi-threaded) — so this is extra robustness, not wasted effort.

### 5. Cross-platform gating — real gap, but pre-existing and not made worse by this commit
The whole module is `#[cfg(all(test, unix))]` and uses `std::os::unix::ffi::OsStringExt`, so it does not exist at all on Windows (not a runtime skip — it isn't compiled). I checked whether that leaves a silent Windows hole: the only Windows CI in this repo is `.github/workflows/windows-compile-check.yml` (this PR's base branch, #2142), and it runs `cargo check` only — it never runs `cargo test` — and is explicitly `continue-on-error: true`, a non-blocking probe by its own header comment. So there is no Windows lane anywhere in CI today that executes tests, for this module or any other. This commit isn't introducing a silent no-op on an existing Windows test lane; there isn't one to silently skip. Net effect: the `userprofile_fallback`/`home_wins_over_userprofile` precedence logic — the part that matters most on actual Windows — is only ever exercised on Linux. That's a real, standing gap, but it predates this commit and isn't something this test commit could fix by itself.

Correction to the task framing: the "matters on Windows" framing for `var` vs `var_os` is imprecise. The scenario this commit actually pins (`non_utf8_home`, built via unix-only `OsStringExt::from_vec` with a raw non-UTF-8 byte sequence) is a Unix/macOS-specific manifestation — non-UTF-8 bytes in env vars are a Unix filesystem/locale reality; that's exactly why the whole module is `unix`-gated. The Windows angle is the separate `USERPROFILE` fallback path, not the non-UTF-8 bug itself.

### 6. CI green at head `67b8f16a6` — literal log lines
Job: "Cargo check & test" (Ubuntu 24.04 runner), run 32450933492, job 96682464683:
```
PASS [   0.024s] (2335/3155) proliferate app_config::home_env_tests::home_dir_os_reads_the_environment_without_utf8_validation
...
Summary [ 144.583s] 3155 tests run: 3155 passed (1 slow), 8 skipped
```
(A second, smaller test binary in the same job also reported `Summary [ 4.709s] 59 tests run: 59 passed, 0 skipped`.)

### 7. Base-branch dependency
Confirmed `head` is exactly `67b8f16a692a6586478cde5accdf7cb9c4904db3` (PR's current tip) and PR base is `ci/windows-compile-check`, which is open PR #2142 targeting `main` (not yet merged). This commit touches only `app_config.rs` and adds one sibling test file — it doesn't touch any CI workflow file and has no dependency on #2142's content, so it doesn't assume it's already on `main`.

PR #2149, opened 2026-08-21
fix(agents): install pinned agent binaries with the windows executable extension

pablonyx:
Stacked on #2145 (`fix/agent-catalog-windows-arm64-pins`), not on `main`. That branch adds the Windows and arm64 Linux pins to the agent catalog. This is the reason those pins would not actually work on Windows yet, so it belongs on top of them rather than beside them.

## The defect

`anyharness/crates/anyharness-lib/src/domains/agents/installer/pinned.rs:40` (pre-fix):

```rust
let target_path = managed_dir.join(kind.as_str());
```

That yields an extension-less path such as `<runtime_home>/agents/claude/native/claude`. Windows will not execute a file without a `.exe` suffix. With #2145's pins in place the installer would fetch the correct artifact, verify its sha256 successfully, write it to a path Windows cannot execute, and then fail at launch with an error that points at the launcher rather than at the filename.

I verified the report by reading rather than running, same as the agent that filed it. It was accurate, and it understated the blast radius by three read sites (below).

## Binary versus archive

The two source arms were both broken, for different reasons, and it matters which is which.

**`Binary`** (claude). The arm never derives a name from anything: not from the URL, not from a catalog field. `Binary` targets have no `expectedBinary`. #2145 pins `windows_x64` to `.../win32-x64/claude.exe`, but that filename is only part of the download URL and is discarded on write. The destination was unconditionally `claude`. This arm needs the extension **added**.

**`Archive`** (codex). This arm does have a name to work from. #2145 pins `windows_x64` to `codex-x86_64-pc-windows-msvc.exe.tar.gz` with `expectedBinary: "codex-x86_64-pc-windows-msvc.exe"`, so the inner member already carries `.exe`. But `download_and_extract_archive_verified` extracts that member and renames it onto the same `target_path`, so the suffix was **dropped in transit**. The extension is present upstream and lost by the rename.

Handled the same way on purpose: the destination filename is derived from the agent kind and the host platform, never from the archive member. Deriving it from the member would make codex's destination `codex-x86_64-pc-windows-msvc.exe`, which no read site looks for. So both arms now target one `managed_pinned_binary_path(runtime_home, kind, role)`, and the archive arm inherits the suffix from the shared target rather than preserving the member's own name. `pinned_archive_install_renames_inner_binary_to_the_platform_name` locks that in.

## Style

Followed the existing in-repo convention rather than inventing one. `installer/npm.rs:484` already had `platform_binary_filename()`, the same idea `Platform::node_binary_name()` expresses for node. It was `pub(super)` inside `installer`, so `readiness` could not reach it. Moved it to `integrations/agent_cli/executable.rs` next to `is_valid_executable`/`make_executable` as `pub(crate)`, which is where the platform-executable vocabulary already lives and where `readiness/paths.rs` already imports from. Same function, same behaviour, better located. Also made it idempotent so a name that already ends in `.exe` does not become `x.exe.exe`.

## Every path-reading site

Grepped `managed_dir`, `kind.as_str()`, `artifact_root`, and the install-manifest readers. Three real read sites, all of which needed changing. A fix that moved only the write site would have converted a Windows launch failure into a Windows "not installed", which is just as broken and harder to read.

| Site | What it does | Changed |
| --- | --- | --- |
| `installer/pinned.rs:40` | **write site**, both arms | yes |
| `installer/service.rs:380` | `install_pinned_role` idempotent-skip probe. Unchanged, it would never skip on Windows and would re-download on every install | yes |
| `readiness/artifacts.rs:21` | `resolve_native_artifact` managed-CLI lookup. Unchanged, readiness would report `InstallRequired` forever | yes |
| `installer/agent_process.rs:87` | `launcher_path_prefixes` probe deciding whether the native dir joins the launcher's `PATH`. Unchanged, the managed launcher would silently lose the native CLI from its `PATH` | yes |
| `installer/reconcile/surface_threading_tests.rs:221` | test fixture seeding a fake native binary | yes, fixture only |
| `readiness/paths.rs:22,31` | `registry_binary` / `registry_npm` lookups | no, agent-process artifacts, not the pinned native path |
| `installer/agent_process.rs:66` | source-build exec candidate | no, already called `platform_binary_filename` |
| `readiness/artifacts.rs:25` | `find_in_path(kind.as_str())` | **no, and worth flagging.** This is PATH adoption of the *user's own* binary, not the managed pin, so my change does not affect it. It is nonetheless a real Windows gap: `find_in_path` never tries `name.exe`, so a user's own `claude.exe` on `PATH` will not be found. Out of scope here because `find_in_path` serves every agent and every catalog-supplied binary name, so widening it is its own change. It overlaps the parallel `/bin/sh` and `HOME` assumption sweep and should land there. |

Note that the three real read sites and the write site each spelled the path themselves. That is the actual defect class, so all four now call one helper and the class is closed, not just the instance.

**Reviewer correction: the class is not fully closed.** `readiness/paths.rs:33` and `:43` (`managed_registry_binary_for_names` / `managed_registry_npm_binary_for_names`) match managed artifacts by bare catalog name against `expected_names`, with no platform suffixing, which is the same defect class in a different corner. It does not bite today only because cursor and opencode carry no `windows_*` targets in `catalogs/agents/catalog.json` and so fail closed on `NoPinForPlatform` before a name comparison ever happens. So: this instance is closed, the class is not. Downgrading the earlier claim accordingly.

## Persisted state

Confirmed rather than assumed. `installer/manifest.rs` persists `path` as a string per role in `install-manifest.json`, and it **is** read back: `service.rs:392` re-reads `entry.path` to checksum-compare for drift detection, and `launch_options/basis.rs:147` hashes it into the launch-options basis fingerprint. So a stale recorded path would matter.

It does not matter here, for two independent reasons. On unix `platform_binary_filename` is the identity function, so no path changes, no manifest churn, and no basis-fingerprint churn. On Windows, `pick_target` failed closed with `NoPinForPlatform` before #2145 (that is what #2141 reported), so no Windows install has ever completed and there is no manifest to be stale. No migration needed.

## Tests

Not run locally, by instruction. CI must run them.

`anyharness/crates/anyharness-lib/src/integrations/agent_cli/executable.rs`
- `platform_binary_filename_carries_the_windows_executable_extension`
- `platform_binary_filename_does_not_double_up_an_existing_extension`

`anyharness/crates/anyharness-lib/src/domains/agents/installer/pinned_tests.rs`
- `pinned_binary_install_lands_on_the_platform_executable_name`
- `pinned_archive_install_renames_inner_binary_to_the_platform_name`

`anyharness/crates/anyharness-lib/src/domains/agents/readiness/artifacts.rs`
- `native_resolution_finds_the_managed_binary_at_its_platform_name`

The naming tests are deliberately not `cfg(windows)`-gated, so they run in the unix CI matrix where a Windows-only test never would. The two installer tests assert the installed path against `managed_pinned_binary_path` rather than a literal, which is the part that has teeth: it pins the write site to the same helper every read site calls, so reintroducing a hardcoded `join(kind.as_str())` at either end breaks the test. `native_resolution_finds_the_managed_binary_at_its_platform_name` is the read-side twin.

`pinned.rs` was 598 lines and the additions pushed it past the 600-line `scripts/check_max_lines.py` cap, so its test module moved to a `#[path]`-included `pinned_tests.rs`, matching the existing `service_tests.rs` pattern. That move is verbatim apart from one rustfmt reflow and one substantive change: `wrong_sha_is_refused_and_nothing_is_installed` now asserts against `managed_pinned_binary_path` instead of `artifact_root(...).join(AgentKind::Claude.as_str())`. That change is correct and necessary given the write-site fix above; it is not a no-op reflow, so it is called out here rather than folded into "verbatim". `python3 scripts/check_max_lines.py` passes.

## Temporary cherry-pick

`.github/workflows/windows-compile-check.yml` is a **temporary cherry-pick from #2142** and is not my work. It is here because it is the only Windows verification instrument available and costs about three and a half minutes. It should disappear on rebase once #2142 lands. Do not review it as part of this change.

## Verified versus unverified

I cannot execute anything on Windows. Stating the split plainly.

**Verified by CI on Windows: `runtime crates (control)` passed.** That job runs `cargo check --target x86_64-pc-windows-msvc -p anyharness` on `windows-latest`, and `anyharness` depends on `anyharness-lib`, so the changed code really was compiled for Windows. This is a **compile verdict only**. It proves the Windows branch of `platform_binary_filename` and all four call sites typecheck and build for the MSVC target. It proves nothing about behaviour. Note also that `cargo check` without `--all-targets` does not build `#[cfg(test)]` code, so the new tests were type-checked on unix only.

(The `desktop crates (the unknown)` probe from #2142 also passed on this branch, which is a useful data point for #2142 but is not a claim about this change.)

**Verified by CI on unix: `Cargo check & test` passed**, covering the existing anyharness test suite plus the five tests above, and `Repo shape checks` passed, covering the max-lines gate. On unix `platform_binary_filename` is the identity function, so these prove the refactor is behaviour-preserving there. They do **not** prove the Windows naming, because the Windows branch is `cfg`-inactive.

All PR checks are green.

**Completely unverified:** that an agent actually installs and launches on a real Windows machine. Nothing in this PR demonstrates that. No Windows binary was produced, downloaded, written, or executed anywhere in this work. The reasoning that `.exe` is the missing piece is sound and the read sites are now consistent, but "the pinned installer works on Windows" is an acceptance test that needs real hardware, and it remains owed. Treat this as removing a known blocker, not as making Windows work.

Also unverified and out of scope, listed so it is not lost:

- The `find_in_path` gap in the table above.
- `downloads.rs:225-242` shells out to `tar` or `unzip` to extract archives. Windows 10 1803+ bundles `tar.exe`, and #2145 pins codex's Windows artifacts as `.tar.gz`, so the archive arm should find its extractor. `unzip` is **not** present on Windows, so any future `.zip` pin would fail there. Nothing in this PR exercises either path on Windows.
- Whatever the parallel `process_kill.rs`, `/bin/sh` and `HOME` sweep, and `release-desktop.yml` lanes turn up. Windows is not one bug.
- **Windows `rename` over a running binary.** Not introduced by this change, but this change is what makes it reachable. `installer/pinned.rs` renames onto `target_path`, and `integrations/agent_cli/launcher.rs:17-21` documents the atomic-promotion design in explicitly POSIX terms: POSIX rename replaces the directory entry while a running managed session keeps the previous inode open, so no kill and no wait is needed. Windows `MoveFileEx` with `MOVEFILE_REPLACE_EXISTING` fails with a sharing violation when the destination is an executing image, so a reinstall or drift-repair while a Windows session holds `claude.exe` open will fail where the same operation on unix silently succeeds. Belongs in the same sweep as the `/bin/sh` launcher work, flagging here so it is not lost.

Refs #2141


pablonyx:
## Verdict

Approve. The fix is correct, the read-site list really is complete, the helper really is identity on unix, and the test split lost nothing. No blocking findings; two claims in the PR body need softening before merge.

I am the reviewer of record and wrote none of this.

## What I verified independently

**The read-site list is complete.** I did not take the table on faith. I grepped `artifact_root(`, `kind.as_str()`, `join("agents")`, `NativeCli`, and the literal `"native"` across the whole repo including `apps/` and `scripts/`. Every production construction of the pinned native path is one of the four the PR changes. The sites that came back and are correctly untouched:

- `auth/login.rs:143` pushes `artifact_root(..., NativeCli)` as a PATH *prefix*, not a filename, and pushes it unconditionally with no existence probe. Windows resolves a bare `claude` on PATH through PATHEXT, so this needs nothing.
- `installer/lock.rs:15` and `seed/quarantine.rs:7-8` operate on the agent directory, not the binary.
- `installer/manifest.rs:83` records `path` from `artifact.path`, which is whatever the write site produced. It cannot go stale relative to the write site by construction.
- `readiness/artifacts.rs:174` and `pinned.rs:117` are `{kind}-launcher`, a different name with a different problem (below).
- `installer/agent_process.rs:65` already called the helper.

**The Binary versus Archive reading is right, and I checked it against the actual pins.** `catalogs/agents/catalog.json` on #2145 gives claude `windows_x64` a `.../win32-x64/claude.exe` URL with no `expectedBinary`, and `pinned.rs:49-59` downloads to `.{kind}.downloading` and renames onto `target_path`, so the URL filename is discarded exactly as described. Codex gets `codex-x86_64-pc-windows-msvc.exe.tar.gz` with `expectedBinary: "codex-x86_64-pc-windows-msvc.exe"`, and `downloads.rs:255-260` renames the extracted member onto `target_path`, dropping the member's name. Both readings confirmed.

**Not preserving the archive member's name is the right call.** If the destination were `codex-x86_64-pc-windows-msvc.exe`, then `service.rs:380`, `readiness/artifacts.rs:21` and `agent_process.rs:87` would all have to learn the per-platform member name, and `install_pinned_role`'s idempotent skip would need the resolved pin in hand before it could name the file it is probing for. Deriving from the kind keeps the name knowable without the catalog. `pinned_archive_install_renames_inner_binary_to_the_platform_name` locks it.

**The helper really is identity on unix.** `executable.rs:8-14` gates the whole rewrite behind `cfg!(windows)`, so the unix arm returns `PathBuf::from(binary_name)` unchanged for every input. Unix CI green does therefore prove the refactor is behaviour-preserving on unix. It proves nothing about Windows, which the PR says.

**The persisted-state conclusion is right, and for a third reason the PR does not give.** Beyond unix identity and beyond no Windows install ever having completed (confirmed: the pre-#2145 catalog has no `windows_*` keys at all, so `pick_target` returned `NoPinForPlatform`), `manifest.rs:84` writes `path` from the `InstalledArtifactResult` the write site returns. Any future rename of the target path rewrites the manifest entry on the next install rather than leaving a stale one. The drift check at `service.rs:320-323` reads `entry.path` and checksums it, so a stale entry would read as "checksum unknown" and trigger a reinstall, not a hard failure. Fail-soft either way.

**The test split lost nothing.** I extracted the old `mod tests` from `pinned.rs` on #2145 and diffed function lists against `pinned_tests.rs`. Old: 9 functions. New: 11. The two added are exactly the two named. No test was dropped, none was silently `#[cfg]`-gated out. Bodies are verbatim apart from one rustfmt reflow at what is now `pinned_tests.rs:440` and one substantive change the PR body does not mention: `wrong_sha_is_refused_and_nothing_is_installed` now asserts against `managed_pinned_binary_path` instead of `artifact_root(...).join(AgentKind::Claude.as_str())`. That change is correct and necessary, but "verbatim apart from one rustfmt reflow" is not quite accurate. Say so in the body.

**Repo shape.** `python3 scripts/check_max_lines.py` passes on the branch head. `pinned.rs` is 260 lines, `pinned_tests.rs` 459, `artifacts.rs` 435, `executable.rs` 145. All clear of 600.

**The cherry-pick is byte-identical to #2142's copy** and the file does not exist on `main`, so there is no add/add conflict to worry about.

## Non-blocking

**1. `readiness/paths.rs:33` and `:43`, and the "class is closed" claim.** The PR says the four-site split "is the actual defect class, so all four now call one helper and the class is closed". It is not closed. `managed_registry_binary_for_names` and `managed_registry_npm_binary_for_names` match managed artifacts by exact filename against catalog-supplied `expected_names`, with no platform suffixing, which is the same defect in a different corner. It happens not to bite today only because cursor and opencode have no `windows_*` targets in `catalogs/agents/catalog.json` and so fail closed on `NoPinForPlatform` before they ever reach a name comparison. The instance is closed. The class is not. Downgrade the claim.

**2. `find_in_path` at `readiness/artifacts.rs:26` is defensible to leave, and I agree with the reasoning.** `find_in_path_matching` at `executable.rs:24-35` does a bare `dir.join(binary_name)` with no PATHEXT walk. This is user-PATH adoption, not the managed pin, so it cannot regress anything this PR touches, and widening it touches every agent and every catalog binary name. Correct to punt. Worth noting that `find_real_binary_in_path` has the same hole and feeds the `BinaryHint` and `NpmPackage` fallbacks at `artifacts.rs:217-256`, so the punt is slightly larger than the one line the PR names.

**3. `unzip` is defensible to leave, and I checked why.** I enumerated every `targets` map in `catalogs/agents/catalog.json` on #2145. The only `.zip` URLs are opencode `darwin-arm64` and `darwin-x64`. Both Windows pins that exist (claude Binary, codex `.tar.gz`) miss the `is_zip` branch at `downloads.rs:224`, and `tar xzf` is satisfied by the bsdtar that ships in System32 on Windows 10 and later. So the `unzip` gap is genuinely unreachable on Windows today. That is worth writing into the code comment, because right now the branch looks live and the next reader will not redo this enumeration.

**4. Windows `rename` over a running binary.** Not introduced here, but this PR is what makes it reachable. `pinned.rs:59` renames onto `target_path`, and `launcher.rs:17-21` documents the whole atomic-promotion design in explicitly POSIX terms: "POSIX rename replaces the directory entry while a running managed session keeps the previous inode open (no kills, no waiting)". Windows `MoveFileEx` with `MOVEFILE_REPLACE_EXISTING` fails with a sharing violation when the destination is an executing image. So a reinstall or drift-repair while a Windows session holds `claude.exe` open will fail where the same operation on unix silently succeeds. Belongs in the same sweep as the `/bin/sh` work, but it should be on the list.

**5. The bigger Windows blocker this fix sits underneath.** `launcher.rs:86` emits `#!/bin/sh` with `export` and `exec`, and `make_executable` at `executable.rs:76-78` is a no-op on Windows. Claude's `agentProcess` in `catalogs/agents/registry.json` is a `managed_npm_package` off a `git+https` URL with `executableRelpath: node_modules/.bin/claude-agent-acp`, so the managed launcher is what actually gets spawned for a session. On Windows that launcher is an unexecutable shell script. This PR is a genuine prerequisite for a working Windows agent, and it is not sufficient on its own. Worth one line in the body so nobody reads a green merge here as "Windows agents work now".

**6. Cherry-picked workflow.** `actions/checkout@v4` in `.github/workflows/windows-compile-check.yml:29` and `:44` is the only unpinned `actions/checkout` in the repo; every other workflow uses `actions/checkout@3d3c42e5...`. `actionlint` also reports eleven shellcheck findings on the summary step at `:55`. Neither matters much since actionlint is not wired into any CI job, and it is #2142's file. But if #2142 stalls, "temporary" becomes permanent and this is what lands. Set yourself a reminder rather than trusting the rebase.

**7. Test portability nit.** `pinned_archive_install_renames_inner_binary_to_the_platform_name` shells out to `tar czf` and is not `cfg(unix)`-gated, unlike its neighbour `staged_npm_swap_replaces_whole_tree_and_old_inode_survives`. Harmless while tests only run on unix. It will be the first thing to look at if a Windows test matrix ever appears.

## Could not verify

No Windows execution, same as the author. Specifically unproven: that a real `claude.exe` written by this path actually launches, that `MoveFileEx` behaves as I described above under an executing destination, and that the archive arm's `tar` invocation succeeds under bsdtar with a `.exe.tar.gz` whose single member carries a `.exe` suffix. The compile verdict from `runtime crates (control)` is real but is a compile verdict only, and as the PR correctly notes, `cargo check` without `--all-targets` never type-checked the new tests for MSVC.

pablonyx (commented):
## Post-rebase pass — verified, no blocking findings

**Rebase integrity.** `git range-diff 622f524dd8..2d144ecfaf 70fbe91044..b4e673f817` shows both commits as patch-identical (`=`) to their pre-rebase versions — no silent content drift. `de1e1cc9e7^` resolves to `70fbe91044`, i.e. the branch sits directly on the new base tip. Clean rebase.

**Items I was asked to re-verify, all confirmed present and accurate:**
- The "verbatim apart from one rustfmt reflow" claim is corrected in the body (the `wrong_sha_is_refused_and_nothing_is_installed` change to assert against `managed_pinned_binary_path` is called out as substantive, not a reflow) — matches the actual diff in `pinned_tests.rs`.
- Both non-blocking observations are recorded: the `readiness/paths.rs:33`/`:43` bare-name-match gap (same defect class, only inert today because cursor/opencode have no `windows_*` pins), and the Windows `MoveFileEx`-vs-executing-image conflict with the POSIX-worded atomic-promotion design at `launcher.rs:17-21`.
- The `unzip`-unreachable-on-Windows comment landed at both punt sites in `downloads.rs` (lines ~223 and ~326), identical wording, correctly noting the two darwin-opencode `.zip` URLs are the only ones in any `targets` map.

**New since last round, flagging for the record rather than as a blocker:**
1. **CI proof not cited.** Windows run 32446919674 has since downloaded and sha256-verified a real `claude.exe` against the `windows_x64` pin and spawned it successfully (combined with #2152). The PR body still frames verification as "compile-only, nothing exercises the Windows path" — that's now stale in a way that undersells the PR. Worth a one-line addendup pointing at that run before merge, not required for correctness.
2. **Merge hazard undocumented on either side.** #2149 moves `platform_binary_filename` out of `installer/npm.rs` into `executable.rs`; #2152 (targeting `main`) adds `platform_npm_bin_relpath` immediately adjacent to that same now-removed definition in `npm.rs`. Neither PR body mentions the other's touch to `npm.rs` — #2149's body doesn't reference #2152 at all, and #2152's body explicitly asserts it "touches a disjoint file... from #2145/#2149," which is not accurate for `npm.rs` specifically (it is accurate for `launcher.rs` vs `pinned.rs`). Whoever merges second should expect a real conflict in `npm.rs` around that region, resolved by keeping `platform_binary_filename` imported from `executable.rs`, keeping `platform_npm_bin_relpath` in `npm.rs`, and dropping the duplicate definition. Flagging so it isn't a surprise; not a reason to hold either PR.

**Verdict: safe to merge as-is.** No blocking findings. The two items above are documentation/sequencing notes for whoever merges, not defects in this diff.

PR #2150, opened 2026-08-21
fix(release): restore the windows desktop release path

pablonyx:
## What was broken

Commit `3d3c0504b8` ("Disable Windows desktop releases", 2026-04-13) removed the Windows leg from the desktop release lane. The stated reason, recorded in the doc note it added, was that Windows stayed off "until the SDK generation step is Windows-safe". That step was `anyharness/sdk`'s `generate:openapi`, a POSIX shell one liner starting `mkdir -p generated src/generated`, which fails under `cmd.exe`. It is a node script now, so the reason no longer holds.

Worth recording because it redirected this work: the desktop crate itself was never the blocker. A `cargo check --target x86_64-pc-windows-msvc -p proliferate --keep-going` on a real `windows-latest` runner finishes clean today, with only dead-code warnings. The desktop crate has compiled for Windows the whole time. The blocker is purely the release path.

## What `3d3c0504b8` removed, and what came back

The April commit removed exactly four things. All four are restored:

1. the `windows` option on the `target_os` workflow_dispatch input;
2. the `os: windows-latest / target: x86_64-pc-windows-msvc / runner_os: windows` matrix entry;
3. the "Upload build artifacts (Windows)" step collecting `*.msi`, `*-setup.exe` and `*-setup.exe.sig`;
4. the `windows-x86_64` platform entry in `scripts/generate-updater-manifest.mjs`.

It also touched `Cargo.lock`, `desktop/package.json`, `desktop/src-tauri/Cargo.toml` and `desktop/src-tauri/tauri.conf.json`. Verified rather than assumed: those four are a pure `0.1.12` to `0.1.14` version bump with nothing Windows-related in them, so nothing there needed restoring.

## What changed because the repo moved on

A blind revert would not have applied and would not have been correct. Beyond `desktop/` becoming `apps/desktop/`:

- **The upload step needed re-siting.** The macOS path now stages into a `release-artifacts` directory with normalized names and uses a SHA-pinned `actions/upload-artifact` with `compression-level: 0`. The restored Windows step uses the same pinned action but keeps the April approach of uploading straight from the bundle directory, because both manifests match Tauri's own NSIS output name and that name already carries the `x64` arch token. No normalization step is needed on this path.
- **`proliferate-diagnostics-collector` cannot be built for Windows.** Its `src/process.rs` line 3 is an unconditional `use std::os::fd::FromRawFd`, so the crate does not compile for `x86_64-pc-windows-msvc` at all. The combined `cargo build` step therefore drops that package on Windows, and the "Stage diagnostics collector for target" step is now `matrix.runner_os != 'windows'`. This is not an improvised workaround: `build_packaging.rs` already treats Windows as an unsupported collector target, writes an explicit placeholder sidecar for it, and `validate_staged_sidecars` already allows a placeholder for exactly that case. The Windows path was designed, just never exercised.
- **A new `windows-x86_64` entry in `scripts/generate-desktop-installer-manifest.mjs`.** That sibling script did not exist in April. It points at the NSIS `setup.exe` rather than the MSI, because the publish job's S3 upload filter copies `*-setup.exe` and `*-setup.exe.sig` but not `*.msi`. The MSI reaches the GitHub Release only.
- **`shell: bash` added to the "Validate release title" step.** It is the only ungated `run:` step in `build-desktop` with no explicit shell, so on a Windows runner it would have executed its backslash-continued, `$RELEASE_TITLE`-interpolating script under `pwsh`. Every other ungated step already declares bash.
- **A first test file for the installer manifest script.** It had none.

## What survived the disable

Grepping the whole workflow for `windows` before touching it, more had survived than expected, which is why this is mostly restoration rather than new work:

- `matrix.runner_os == "windows"` `.exe` branches already exist in the anyharness copy step, the `proliferate-debug` staging step, the `proliferate-worker` staging step, and the `ANYHARNESS_BIN` export in the Tauri build step;
- the Tauri build step already falls through to `pnpm tauri build --bundles msi,nsis` for non-macOS;
- "Verify updater artifacts" already has an `else` branch asserting `nsis/*-setup.exe` and `nsis/*-setup.exe.sig`;
- the S3 upload filter in `publish-updater` already includes `*-setup.exe` and `*-setup.exe.sig`;
- the rust cache step's own comment refers to "any dormant windows entry";
- `apps/desktop/src-tauri/build.rs` and `build_packaging.rs` are thoroughly Windows-aware, with `.exe` suffixing throughout;
- `tauri.conf.json` already carries `icons/icon.ico` and an updater `windows.installMode`.

## Decision for you: the fail-closed manifest

`scripts/generate-updater-manifest.mjs` is fail-closed by design. A declared platform with no matching artifact pushes to `errors` and exits non-zero, so naively restoring `windows-x86_64` would make any Windows build failure kill the whole release, macOS included.

The mechanism to avoid that already exists and I did not invent a new one. When Intel builds were paused on 2026-08-20, `darwin-x86_64` was given `optional: true`, which downgrades a missing artifact to a skip with a warning. I reused it verbatim for `windows-x86_64` in both manifest scripts.

`optional` alone is not sufficient, and this is the part that needs your ruling. `create-release` and `publish-updater` both declare `needs: build-desktop`, so a failed Windows matrix leg would skip them outright no matter what the manifest script does. To make the beta posture real I also set `continue-on-error: ${{ matrix.runner_os == 'windows' }}` on `build-desktop`.

The tradeoff, stated plainly:

- **As implemented (Windows non-blocking).** A broken Windows build produces a release with no Windows update entry and no Windows download entry. macOS ships. The cost is that a Windows regression is a warning annotation on a green run, so it can go unnoticed for a release or two if nobody reads the log.
- **The alternative (Windows blocking).** Drop `continue-on-error` and drop `optional` from both `windows-x86_64` entries. A Windows regression then stops every desktop release until it is fixed. Correct once Windows is a supported platform, too strict for a first beta on a leg that has never once run.

I recommend the non-blocking posture for the beta and flipping it the moment Windows has shipped one clean release. Both switches are one line each and are named in the code comments and in the docs so the reversal is obvious. If you want the stricter posture now, say so and I will pull both lines.

## What a Windows user loses, stated loudly rather than stubbed silently

Three warning annotations are emitted from a `Report Windows beta gaps` step so the release log records the losses instead of a placeholder passing for a working component:

1. **No diagnostics collector.** The crate does not compile for Windows, so a placeholder sidecar ships. In-app diagnostics capture, the support-report snapshot, and the collector-backed export path are all unavailable.
2. **No bundled agent seed.** See the blocker below.
3. **No Authenticode signature.** First launch shows a click-through SmartScreen warning, per your explicit acceptance for the beta.

## Blocker left deliberately untouched: the Windows agent seed

This one is genuinely a separate change and I stopped rather than half-doing it.

`Build bundled agent seed` is gated `matrix.runner_os == 'macos'`, so a Windows build produces no seed archive. At runtime `agent_seed_env.rs` finds no bundled seed, `bundled_seed_dir` returns `None`, and in a release build it still sets `ANYHARNESS_AGENT_SEED_EXPECTED=1`. See "Hard runtime dependency" below for what that actually means on Windows today: not a fallback to whatever Node the machine has, but no managed agent launcher that can run at all yet.

Fixing it means three things that do not belong in a workflow restoration PR:

- `apps/desktop/src-tauri/agent-seed.inputs.json` pins Node only for `aarch64-apple-darwin` and `x86_64-apple-darwin`, so a Windows Node distribution and its checksum need adding;
- `installNode` in `scripts/build-agent-seed.mjs` unpacks a tarball, and the Windows Node distribution is a zip;
- `hostTarget()` maps only `darwin`, and `assertNativeTarget` hard-fails any other host, so the seed builder cannot currently run on a Windows runner at all.

Also out of scope by assignment, noted in case the dispatch surfaces them: `anyharness/crates/anyharness-lib/src/process_kill.rs` Windows support and the agent catalog Windows pins are owned by other agents in parallel.

## Signing

No Authenticode signing was added, per your acceptance of the SmartScreen click-through. The Tauri updater's minisign signature is a separate mechanism and does apply to the Windows artifact: `TAURI_SIGNING_PRIVATE_KEY` and `TAURI_SIGNING_PRIVATE_KEY_PASSWORD` are set at workflow `env:` level, not inside a macOS-gated step, so they reach the Windows `pnpm tauri build` unchanged, and `bundle.createUpdaterArtifacts` is `true` for every target. That is what produces the `*-setup.exe.sig` the restored upload step and the existing "Verify updater artifacts" Windows branch both expect.

## Verified

- `actionlint .github/workflows/release-desktop.yml` exits 0.
- `node --check` passes on both modified manifest scripts.
- `node --test scripts/ci-cd/generate-updater-manifest.test.mjs scripts/ci-cd/generate-desktop-installer-manifest.test.mjs` reports `# pass 11 / # fail 0`.
- Negative control on the new coverage: deleting `optional: true` from the `windows-x86_64` entry in each script turns exactly the three intended tests red (`still generates a manifest when the Windows beta leg produced nothing` in both files, plus the Intel-paused test), and restoring it returns all 11 to green.
- `python3 scripts/check_docs.py` and `python3 scripts/check_max_lines.py` both pass.
- `git show 3d3c0504b8` read in full to confirm the four removals and that the remaining four touched files are a version bump only.

## Unverified

Updated twice now: once after the first dry run (B1 section above), and again after a follow-up run once the Sentry blocker below was cleared. Still unproven:

- that `resources: ["agent-seeds/"]` bundles cleanly when the directory holds only its `.gitignore`;
- that the transparent, `titleBarStyle: Overlay` window config produces a usable window on Windows.

Now confirmed rather than assumed:

- `cargo build --release --target x86_64-pc-windows-msvc -p anyharness -p proliferate-debug -p proliferate-worker` succeeds on `windows-latest` (PR #2146's territory, first exercised for real here);
- `pnpm install` and the SDK `generate`/`build` steps succeed on a Windows runner;
- the `continue-on-error` run-conclusion question from B1, resolved as described in that section;
- **`pnpm tauri build --bundles msi,nsis` succeeds on `windows-latest`, and the NSIS naming assumption was correct.** Run **https://github.com/proliferate-ai/proliferate/actions/runs/32447570745** is the first `pnpm tauri build` this repo has ever run on Windows. Tauri's bundler log reports `Finished 2 bundles`, producing `Proliferate_0.4.21_x64_en-US.msi` and `Proliferate_0.4.21_x64-setup.exe`, each with a `.sig` file. The NSIS installer name is exactly `Proliferate_{version}_x64-setup.exe`, which matches both the `compgen -G "$base/nsis/*-setup.exe"` glob in `release-desktop.yml` and the `/(x64|x86_64).*setup\.exe$/i` regex in both manifest scripts. Only the MSI carries an extra `_en-US` locale infix, and no matcher depends on the MSI name, so it does not matter.

### B2 (blocking): the Windows leg was not opt-in

The reviewer is right: the matrix entry at `release-desktop.yml:112-115` was unconditional, and the filter step only skipped a leg on `event_name != 'push'` with an explicit `target_os` override. Every caller passed `all`, and a `desktop-v*` tag push bypassed the filter entirely (a push event carries no `inputs` at all, so `target_os` fell back to its `'all'` default with no way to override it). On merge as originally written, Windows would have started building on every staging deploy, every nightly zero-touch auto-prod train, every promote, and every hotfix, on a leg that had never once completed.

Fixed by adding a new `enable_windows_beta` input, declared identically to `dry_run` and `target_os` on both `workflow_dispatch` and `workflow_call`, defaulting to `false`. The "Decide whether to build this matrix entry" step now also skips the Windows leg whenever `enable_windows_beta` is not explicitly `true`, regardless of event type.

**Note: this branch was rebased during review, and PR #2140 landed in that window.** #2140 unified `nightly-release-train.yml`, `promote-production.yml` and `hotfix-production.yml` into a single `release.yml` (hotfix and promote are now inputs on that one workflow: `ref` + `surfaces` for hotfix, `skip_build` for promote, per its own header comment). Git's rename-follow reapplied the `enable_windows_beta: false` hunk from my original three-file version onto the sole surviving `release-desktop` call site at `release.yml:321-343`, and it landed correctly there. Traced through the two real paths that exist post-#2140, plus the tag push:

- **`deploy-staging.yml` -> `_deploy-desktop.yml` -> `release-desktop.yml`.** `_deploy-desktop.yml` gained the same `enable_windows_beta` input (default `false`), threaded straight through. Neither `deploy-staging.yml` nor `_deploy-desktop.yml` overrides it, so it resolves to `false`. **Windows no longer builds on staging deploy.**
- **`release.yml` -> `release-desktop.yml`.** This is the unified pipeline that replaced the nightly train, promote, and hotfix: the daily cron and the `workflow_dispatch` covering all three modes. It calls `release-desktop.yml` directly, already passed `target_os: all` explicitly, and now carries `enable_windows_beta: false` explicitly too, with a comment saying not to flip it until the Windows leg has a proven green dry run and B1 is resolved. `release.yml`'s own `workflow_dispatch` exposes no `enable_windows_beta` input at all, which is stricter than before rather than a regression: there is no way to opt into the Windows beta through this pipeline's public surface at all right now. **Windows no longer builds on the cron train, on promote, or on a hotfix.**
- **A raw `desktop-v*` tag push.** `push` events carry no `inputs` context at all, so `inputs.enable_windows_beta` is empty/falsy there regardless of what any caller does. **Windows no longer builds on a bare tag push.**

The only path left that builds Windows is an explicit `workflow_dispatch` run against `release-desktop.yml` itself with `enable_windows_beta: true` set by a human, in addition to selecting `target_os: windows` or `all`. That is the dispatch used for the B1 experiment below.

Also fixed while in there (non-blocking, N1/N5 in the review):

- The warning annotation at `release-desktop.yml:329` and the paragraph in `guides/deploying/releases.md` both said `PROLIFERATE_AGENT_SEED_EXPECTED`. The real name, confirmed against `agent_seed_env.rs:7` and `installer/seed/mod.rs:44`, is `ANYHARNESS_AGENT_SEED_EXPECTED`. Fixed in both places, and the accompanying claim that Windows "falls back to whatever Node the machine already has" is also corrected: per the seed-tracing below, no managed agent launcher can run on Windows at all yet, so there is no fallback, there is nothing.
- `agent_seed_env.rs:10-30`'s `current_target_triple()` had no `cfg` arm for `aarch64-pc-windows-msvc`. Added one. It was unreachable today since only `x86_64-pc-windows-msvc` is in the release matrix, but #2145 already adds `windows_arm64` catalog pins, and without this arm the function would fail to compile the moment an ARM Windows target actually gets built, because none of its `cfg`-gated blocks would produce the required `&'static str` return value.

### B1 (blocking): the continue-on-error run-conclusion question, resolved by running it

Job-level `continue-on-error: ${{ matrix.runner_os == 'windows' }}` on `build-desktop` is confirmed to make `needs.build-desktop.result` report `success` to `create-release` and `publish-updater` even when the Windows leg fails, so those jobs do run. What was unresolved is whether a `continue-on-error` failure turns the *overall run's* conclusion red, which is the value the caller actually gates on through `needs.release-desktop.result` (post-#2140 this is `release.yml:321-343`; before the rebase it was the now-retired `nightly-release-train.yml:341` and `hotfix-production.yml:332`, gating the equivalent `needs.deploy-desktop.result`).

I dispatched `release-desktop.yml` on this branch (after the B2 fix above) with `dry_run: true`, `target_os: windows`, `enable_windows_beta: true`, `publish_updater: false`:

**Run: https://github.com/proliferate-ai/proliferate/actions/runs/32442203866**

The Windows leg genuinely failed on this run, which turns out to be exactly the experiment B1 asked for, with a real failure instead of a staged one:

- **`build-desktop (windows-latest, x86_64-pc-windows-msvc, windows)` job conclusion: `failure`.**
- **Overall run conclusion: `success`.**

That answers B1. A `continue-on-error` job failing does not turn the overall run conclusion red; the run reports green exactly as the GitHub syntax reference and discussion #15452 say, not as Ken Muse's writeup says. `needs.<workflow-call-job>.result` is what a caller reads for a reusable-workflow call, and it reflects this same run-level conclusion, so the caller gate (`release.yml:321-343` today, `nightly-release-train.yml:341` / `hotfix-production.yml:332` before #2140 retired them) would have read `success` here and proceeded to `publish-product-release` even though Windows was broken. The failure scenario in the original B1 finding is real: absent the B2 gate now landed, a Windows-only failure on the unified release pipeline would not have blocked the macOS release from finalizing.

**What actually failed, so it is not mistaken for a packaging problem.** The Windows Rust cross-compile succeeded: `cargo build --release --target x86_64-pc-windows-msvc -p anyharness -p proliferate-debug -p proliferate-worker` completed in the `Build Rust binaries for target` step (18 minutes on a cold cache), and `Generate and build SDK` also completed. The job then died in `Build frontend`, before `pnpm tauri build` ever ran:

```
##[error]Sentry desktop renderer project 'proliferate-desktop' is not accessible in org 'proliferate' (HTTP 403). Update SENTRY_DESKTOP_RENDERER_PROJECT before releasing.
```

This is a Sentry API access check, not a Windows-specific defect and not a build or packaging failure, and the root cause is now established rather than theorized. There is no `environment:` binding anywhere on any call path of `release-desktop.yml`, so a secret-scoping difference between a bare `workflow_dispatch` and an environment-gated caller (my first guess) cannot be the explanation. The actual cause is a token rotation. `SENTRY_AUTH_TOKEN` was rotated at `2026-08-20T19:51:01Z`, the same timestamp as the DSN variables, and the new token cannot read the `proliferate-desktop` project. Two independent calls on that token fail: the project API returns this 403, and `sentry-cli debug-files upload` separately returns "Project not found" against the same project. An unauthenticated control request returned a normal 401, which rules out a WAF or IP block as the explanation. This is org-wide on the current token, not specific to this dispatch's secret scope, and it would break a macOS release dispatched right now, not just this Windows run. Flagging as a separate, unrelated finding rather than fixing it here: **`SENTRY_AUTH_TOKEN` needs read access to `proliferate-desktop` restored before any desktop release, macOS or Windows, can get past `Build frontend` again.**

**Net effect on Unverified below at the time: unresolved.** Because `Build frontend` failed first, `Build Tauri app` and every step after it were skipped on this run. See the Unverified section: a follow-up run, dispatched once the Sentry blocker above was worked around, got past this step and resolved the NSIS naming question for real.

## Hard runtime dependency, stated plainly

The Stacking section below only names #2146. That is incomplete. **This PR has a hard runtime dependency on #2149 and #2145 as well, not just on #2146.**

- Without #2149, `installer/pinned.rs` writes the pinned agent binary to an extension-less path on Windows (`<runtime_home>/agents/claude/native/claude` instead of `claude.exe`), which Windows will not execute. With no bundled agent seed on Windows (see below) and no working pinned installer either, there is no path at all by which an agent binary reaches a Windows machine built from this PR.
- Even with #2149 merged, **no Windows session can start.** Claude's `agentProcess` in `catalogs/agents/registry.json` is a `managed_npm_package`, and what actually gets spawned for a session is the managed launcher `integrations/agent_cli/launcher.rs:86` generates, which is a `#!/bin/sh` script. Windows cannot execute that, and `make_executable` is a no-op there. A separate PR is in flight for the `/bin/sh` managed-launcher work; until it lands, this is the honest state.

Net: this PR produces an installable, launchable Windows app. It is not yet something a user can try, because no agent turn can run on it. Do not read a green merge here as "Windows agents work now."

## Stacking

Based on `fix/sdk-generate-windows-safe` (draft PR #2146), which this depends on: without its node port of `generate:openapi`, the Windows leg fails at the SDK generation step. Merge after #2146, or rebase onto main once it lands. It also depends on #2149 and #2145 for the reasons above, even though neither is a build-time dependency of this PR's own diff.

### DIFF
--- a/.github/workflows/_deploy-desktop.yml
+++ b/.github/workflows/_deploy-desktop.yml
@@ -24,6 +24,10 @@ on:
       target_os:
         default: all
         type: string
+      enable_windows_beta:
+        description: "Opt into building the Windows beta leg (never on by default; see B2 in PR #2150)"
+        default: false
+        type: boolean
 
 permissions:
   contents: read
@@ -166,6 +170,7 @@ jobs:
               echo "- Dry run: \`${{ inputs.dry_run }}\`"
               echo "- Publish updater: \`${{ inputs.publish_updater }}\`"
               echo "- Target OS: \`${{ inputs.target_os }}\`"
+              echo "- Windows beta enabled: \`${{ inputs.enable_windows_beta }}\`"
             fi
           } >> "$GITHUB_STEP_SUMMARY"
 
@@ -183,4 +188,5 @@ jobs:
       publish_updater: ${{ inputs.publish_updater }}
       release_title: ${{ inputs.release_title }}
       target_os: ${{ inputs.target_os }}
+      enable_windows_beta: ${{ inputs.enable_windows_beta }}
     secrets: inherit
--- a/.github/workflows/release-desktop.yml
+++ b/.github/workflows/release-desktop.yml
@@ -22,6 +22,10 @@ on:
           - macos
           - windows
         default: all
+      enable_windows_beta:
+        description: "Opt into building the Windows beta leg (never on by default; see B2 in PR #2150)"
+        type: boolean
+        default: false
       allow_unsigned:
         description: "Allow ad-hoc (unsigned) macOS builds when APPLE_SIGNING_IDENTITY is not configured"
         type: boolean
@@ -53,6 +57,10 @@ on:
         description: "Target OS"
         type: string
         default: all
+      enable_windows_beta:
+        description: "Opt into building the Windows beta leg (never on by default; see B2 in PR #2150)"
+        type: boolean
+        default: false
       allow_unsigned:
         description: "Allow ad-hoc (unsigned) macOS builds when APPLE_SIGNING_IDENTITY is not configured"
         type: boolean
@@ -109,7 +117,11 @@ jobs:
           #   runner_os: macos
           # Windows was removed 2026-04-13 by 3d3c0504b8 because the SDK
           # generate step was POSIX-shell only. That step is node now, so the
-          # leg is back for the beta.
+          # leg is back for the beta. The matrix entry itself is unconditional;
+          # "Decide whether to build this matrix entry" below is what actually
+          # keeps it from running unless `enable_windows_beta` is explicitly
+          # true, so it never rides along on push, workflow_call or
+          # unattended workflow_dispatch runs.
           - os: windows-latest
             target: x86_64-pc-windows-msvc
             runner_os: windows
@@ -133,12 +145,21 @@ jobs:
         shell: bash
         env:
           TARGET_OS: ${{ inputs.target_os || 'all' }}
+          # `inputs.enable_windows_beta` is only populated on workflow_dispatch
+          # and workflow_call. A `push` (desktop-v* tag) event carries no
+          # inputs at all, so this is empty/false there too, and the Windows
+          # leg cannot ride along on a tag push. Every workflow_call caller
+          # must opt in explicitly; none does today (see guides/deploying/releases.md).
+          ENABLE_WINDOWS_BETA: ${{ inputs.enable_windows_beta && 'true' || 'false' }}
         run: |
+          skip=false
           if [[ "${{ github.event_name }}" != "push" ]] && [[ "$TARGET_OS" != "all" ]] && [[ "$TARGET_OS" != "${{ matrix.runner_os }}" ]]; then
-            echo "skip=true" >> "$GITHUB_OUTPUT"
-          else
-            echo "skip=false" >> "$GITHUB_OUTPUT"
+            skip=true
           fi
+          if [[ "${{ matrix.runner_os }}" == "windows" ]] && [[ "$ENABLE_WINDOWS_BETA" != "true" ]]; then
+            skip=true
+          fi
+          echo "skip=$skip" >> "$GITHUB_OUTPUT"
 
       - name: Validate release ref
         if: steps.filter.outputs.skip != 'true' && github.event_name == 'workflow_dispatch' && !inputs.dry_run && !inputs.git_sha
@@ -326,7 +347,7 @@ jobs:
         shell: bash
         run: |
           echo "::warning::Windows beta ships a placeholder diagnostics collector. proliferate-diagnostics-collector does not compile for x86_64-pc-windows-msvc, so in-app diagnostics capture, the support-report snapshot and the collector-backed export path are all unavailable in this build."
-          echo "::warning::Windows beta ships no bundled agent seed. agent-seed.inputs.json pins Node only for the two Apple targets and build-agent-seed.mjs is native-macOS-only, so this build sets PROLIFERATE_AGENT_SEED_EXPECTED with no seed present and agents fall back to whatever Node the machine already has."
+          echo "::warning::Windows beta ships no bundled agent seed. agent-seed.inputs.json pins Node only for the two Apple targets and build-agent-seed.mjs is native-macOS-only, so this build sets ANYHARNESS_AGENT_SEED_EXPECTED with no seed present. No managed agent launcher can run on Windows yet (see PR body), so there is no fallback to \"whatever Node the machine already has\" either."
           echo "::warning::Windows installers are not Authenticode signed. First launch shows a click-through SmartScreen warning. The Tauri updater minisign signature is separate and is applied."
 
       - name: Stage Proliferate debug helper for target
--- a/.github/workflows/release-desktop.yml
+++ b/.github/workflows/release-desktop.yml
@@ -20,6 +20,7 @@ on:
         options:
           - all
           - macos
+          - windows
         default: all
       allow_unsigned:
         description: "Allow ad-hoc (unsigned) macOS builds when APPLE_SIGNING_IDENTITY is not configured"
@@ -106,8 +107,20 @@ jobs:
           # - os: macos-15-intel
           #   target: x86_64-apple-darwin
           #   runner_os: macos
+          # Windows was removed 2026-04-13 by 3d3c0504b8 because the SDK
+          # generate step was POSIX-shell only. That step is node now, so the
+          # leg is back for the beta.
+          - os: windows-latest
+            target: x86_64-pc-windows-msvc
+            runner_os: windows
 
     runs-on: ${{ matrix.os }}
+
+    # Beta posture: Windows is best effort. A failed Windows leg must not take
+    # the macOS release down with it, and `needs: build-desktop` on
+    # create-release would do exactly that if this job reported failure. Delete
+    # this line to make Windows release-blocking.
+    continue-on-error: ${{ matrix.runner_os == 'windows' }}
 
     env:
       # The desktop renderer (vite) build OOMs the default Node heap on the
@@ -173,6 +186,7 @@ jobs:
 
       - name: Validate release title
         if: steps.filter.outputs.skip != 'true'
+        shell: bash
         env:
           RELEASE_TITLE: ${{ inputs.release_title || '' }}
         run: |
@@ -235,11 +249,24 @@ jobs:
         if: steps.filter.outputs.skip != 'true'
         shell: bash
         run: |
-          cargo build --release --target "${{ matrix.target }}" \
-            -p anyharness \
-            -p proliferate-diagnostics-collector \
-            -p proliferate-debug \
-            -p proliferate-worker
+          # proliferate-diagnostics-collector is POSIX-only: process.rs reads an
+          # inherited descriptor through an unconditional `use std::os::fd`, so
+          # the crate does not compile for x86_64-pc-windows-msvc at all. The
+          # desktop build script already recognizes Windows as an unsupported
+          # collector target and stages an explicit placeholder sidecar, so
+          # asking cargo for the real one here would only fail the leg.
+          if [[ "${{ matrix.runner_os }}" == "windows" ]]; then
+            cargo build --release --target "${{ matrix.target }}" \
+              -p anyharness \
+              -p proliferate-debug \
+              -p proliferate-worker
+          else
+            cargo build --release --target "${{ matrix.target }}" \
+              -p anyharness \
+              -p proliferate-diagnostics-collector \
+              -p proliferate-debug \
+              -p proliferate-worker
+          fi
           mkdir -p apps/desktop/src-tauri/binaries
           if [[ "${{ matrix.runner_os }}" == "windows" ]]; then
             cp "target/${{ matrix.target }}/release/anyharness.exe" "apps/desktop/src-tauri/binaries/anyharness-${{ matrix.target }}.exe"
@@ -279,7 +306,7 @@ jobs:
           spctl --assess --verbose=4 --type open --context context:primary-signature "$EXTRACTED_NODE_BIN"
 
       - name: Stage diagnostics collector for target
-        if: steps.filter.outputs.skip != 'true'
+        if: steps.filter.outputs.skip != 'true' && matrix.runner_os != 'windows'
         shell: bash
         run: |
           mkdir -p apps/desktop/src-tauri/binaries
@@ -289,6 +316,18 @@ jobs:
           cp "$collector" "$staged"
           chmod +x "$staged"
           cmp "$collector" "$staged"
+
+      # Say the losses out loud rather than letting a placeholder sidecar and an
+      # empty resource directory pass for working ones. The desktop build script
+      # writes the collector placeholder; this step exists so the release log
+      # records what a Windows user does not get.
+      - name: Report Windows beta gaps
+        if: steps.filter.outputs.skip != 'true' && matrix.runner_os == 'windows'
+        shell: bash
+        run: |
+          echo "::warning::Windows beta ships a placeholder diagnostics collector. proliferate-diagnostics-collector does not compile for x86_64-pc-windows-msvc, so in-app diagnostics capture, the support-report snapshot and the collector-backed export path are all unavailable in this build."
+          echo "::warning::Windows beta ships no bundled agent seed. agent-seed.inputs.json pins Node only for the two Apple targets and build-agent-seed.mjs is native-macOS-only, so this build sets PROLIFERATE_AGENT_SEED_EXPECTED with no seed present and agents fall back to whatever Node the machine already has."
+          echo "::warning::Windows installers are not Authenticode signed. First launch shows a click-through SmartScreen warning. The Tauri updater minisign signature is separate and is applied."
 
       - name: Stage Proliferate debug helper for target
         if: steps.filter.outputs.skip != 'true'
@@ -651,6 +690,21 @@ jobs:
           name: proliferate-dmg-${{ matrix.target }}
           path: |
             target/${{ matrix.target }}/release/release-artifacts/*.dmg
+          if-no-files-found: error
+          compression-level: 0
+
+      # Restored from 3d3c0504b8. There is no name-normalization step on this
+      # path: the updater and installer manifests match Tauri's own NSIS output
+      # name, which already carries the x64 arch token.
+      - name: Upload build artifacts (Windows)
+        if: steps.filter.outputs.skip != 'true' && matrix.runner_os == 'windows'
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: desktop-${{ matrix.target }}
+          path: |
+            target/${{ matrix.target }}/release/bundle/**/*.msi
+            target/${{ matrix.target }}/release/bundle/**/*-setup.exe
+            target/${{ matrix.target }}/release/bundle/**/*-setup.exe.sig
           if-no-files-found: error
           compression-level: 0
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -336,6 +336,13 @@ jobs:
       # to a GitHub Environment, so there is no approval gate on this step.
       publish_updater: true
       target_os: all
+      # Deliberately not opted into the Windows beta leg (see B2 in PR #2150):
+      # this is the unattended zero-touch auto-prod path with no approval
+      # gate, and the Windows leg has never completed a real build. Leaving
+      # this at its `false` default means `target_os: all` still only builds
+      # macOS here. Flip to `true` only once the Windows leg has a proven
+      # green dry run and B1 (continue-on-error run-conclusion) is resolved.
+      enable_windows_beta: false
     secrets: inherit
 
   # ── Production deploys ────────────────────────────────────────────────────────
--- a/apps/desktop/src-tauri/src/agent_seed_env.rs
+++ b/apps/desktop/src-tauri/src/agent_seed_env.rs
@@ -28,6 +28,10 @@ pub(crate) fn current_target_triple() -> &'static str {
     {
         "x86_64-pc-windows-msvc"
     }
+    #[cfg(all(target_os = "windows", target_arch = "aarch64"))]
+    {
+        "aarch64-pc-windows-msvc"
+    }
 }
 
 pub(crate) fn launch_env<R: Runtime>(app: &AppHandle<R>) -> HashMap<String, String> {
--- a/guides/deploying/releases.md
+++ b/guides/deploying/releases.md
@@ -67,8 +67,8 @@ the Desktop package, Tauri configuration, and Cargo package. A manual dry run
 can build from the selected ref without publishing. A manual non-dry run must
 be dispatched from an existing `desktop-v*` tag ref; a branch ref fails release
 validation. Tag pushes and publishing release-coordinator calls can build the
-current macOS matrix and create a draft GitHub Release. Updater/download
-publication is separate:
+current macOS and Windows matrix and create a draft GitHub Release.
+Updater/download publication is separate:
 
 - a `desktop-v*` tag push publishes it automatically;
 - a reusable production or train call can request it explicitly;
@@ -85,6 +85,35 @@ collector package first and stages that exact executable into the app. Verify
 the app's `Contents/MacOS` inventory contains executable, signed AnyHarness,
 Worker, `proliferate-debug`, and `proliferate-diagnostics-collector` binaries;
 the collector must not contain the debug placeholder marker.
+
+### Windows Beta
+
+Windows (`x86_64-pc-windows-msvc`) returned to the matrix on 2026-08-20, having
+been removed on 2026-04-13 while the SDK generation step was POSIX-shell only.
+It is a beta leg and is deliberately not release-blocking:
+
+- the Windows matrix entry runs `continue-on-error`, so a failed Windows build
+  leaves the macOS release and the updater publish intact;
+- `windows-x86_64` is an `optional` platform in both
+  `scripts/generate-updater-manifest.mjs` and
+  `scripts/generate-desktop-installer-manifest.mjs`, so a release with no
+  Windows artifact publishes without a Windows entry rather than failing;
+- the Windows installer is not Authenticode signed, so first launch shows a
+  SmartScreen warning that the user clicks through. The Tauri updater's
+  minisign signature is a separate mechanism and is applied, so in-app updates
+  still verify;
+- `proliferate-diagnostics-collector` does not compile for Windows, so the
+  Windows build ships a placeholder sidecar and has no in-app diagnostics
+  capture or support snapshot;
+- there is no bundled agent seed on Windows. `agent-seed.inputs.json` pins Node
+  only for the two Apple targets and `scripts/build-agent-seed.mjs` is
+  native-macOS-only, so a Windows build sets `PROLIFERATE_AGENT_SEED_EXPECTED`
+  with no seed present and agents use whatever Node the machine already has.
+
+The build logs a warning annotation for each of those gaps.
+
+To make Windows release-blocking, delete the `continue-on-error` line from
+`build-desktop` and drop `optional` from the `windows-x86_64` entries.
 
 ## Runtime And SDK
 
--- a/guides/deploying/releases.md
+++ b/guides/deploying/releases.md
@@ -88,32 +88,18 @@ the collector must not contain the debug placeholder marker.
 
 ### Windows Beta
 
-Windows (`x86_64-pc-windows-msvc`) returned to the matrix on 2026-08-20, having
-been removed on 2026-04-13 while the SDK generation step was POSIX-shell only.
-It is a beta leg and is deliberately not release-blocking:
+Windows (`x86_64-pc-windows-msvc`) returned to the matrix on 2026-08-20, having been removed on 2026-04-13 while the SDK generation step was POSIX-shell only. It is a beta leg, it is deliberately not release-blocking, and it is opt-in only:
 
-- the Windows matrix entry runs `continue-on-error`, so a failed Windows build
-  leaves the macOS release and the updater publish intact;
-- `windows-x86_64` is an `optional` platform in both
-  `scripts/generate-updater-manifest.mjs` and
-  `scripts/generate-desktop-installer-manifest.mjs`, so a release with no
-  Windows artifact publishes without a Windows entry rather than failing;
-- the Windows installer is not Authenticode signed, so first launch shows a
-  SmartScreen warning that the user clicks through. The Tauri updater's
-  minisign signature is a separate mechanism and is applied, so in-app updates
-  still verify;
-- `proliferate-diagnostics-collector` does not compile for Windows, so the
-  Windows build ships a placeholder sidecar and has no in-app diagnostics
-  capture or support snapshot;
-- there is no bundled agent seed on Windows. `agent-seed.inputs.json` pins Node
-  only for the two Apple targets and `scripts/build-agent-seed.mjs` is
-  native-macOS-only, so a Windows build sets `PROLIFERATE_AGENT_SEED_EXPECTED`
-  with no seed present and agents use whatever Node the machine already has.
+- the Windows matrix entry is skipped unless the caller explicitly sets `enable_windows_beta: true`. Every automatic caller (`deploy-staging.yml`, `promote-production.yml`, `hotfix-production.yml` through `_deploy-desktop.yml`, and `nightly-release-train.yml` directly) leaves it at its `false` default, and a plain `desktop-v*` tag push carries no inputs at all, so it also defaults to `false`. Only an explicit `workflow_dispatch` run with `enable_windows_beta: true` builds Windows;
+- the Windows matrix entry also runs `continue-on-error`, so on the runs where it is enabled, a failed Windows build leaves the macOS release and the updater publish intact;
+- `windows-x86_64` is an `optional` platform in both `scripts/generate-updater-manifest.mjs` and `scripts/generate-desktop-installer-manifest.mjs`, so a release with no Windows artifact publishes without a Windows entry rather than failing;
+- the Windows installer is not Authenticode signed, so first launch shows a SmartScreen warning that the user clicks through. The Tauri updater's minisign signature is a separate mechanism and is applied, so in-app updates still verify;
+- `proliferate-diagnostics-collector` does not compile for Windows, so the Windows build ships a placeholder sidecar and has no in-app diagnostics capture or support snapshot;
+- there is no bundled agent seed on Windows. `agent-seed.inputs.json` pins Node only for the two Apple targets and `scripts/build-agent-seed.mjs` is native-macOS-only, so a Windows build sets `ANYHARNESS_AGENT_SEED_EXPECTED` with no seed present. There is also no managed agent launcher that can run on Windows yet (the launcher is a `#!/bin/sh` script), so there is no fallback to "whatever Node the machine already has": no agent session can start on a Windows build today regardless of seed state.
 
 The build logs a warning annotation for each of those gaps.
 
-To make Windows release-blocking, delete the `continue-on-error` line from
-`build-desktop` and drop `optional` from the `windows-x86_64` entries.
+To make Windows release-blocking, delete the `continue-on-error` line from `build-desktop` and drop `optional` from the `windows-x86_64` entries. To make it build automatically on one of the existing pipelines, pass `enable_windows_beta: true` from that caller.
 
 ## Runtime And SDK
 


pablonyx:
## Verdict

Request changes. Two blocking items, both about the release pipeline rather than the Windows code: the `continue-on-error` reasoning stops one hop short of the two callers that actually gate on this workflow's result, and the Windows leg is not opt-in, so it goes live on the nightly auto-prod train the moment this merges.

I am the reviewer of record and wrote none of this.

## Blocking

### B1. The `continue-on-error` argument is right inside this file and unanalyzed outside it

`.github/workflows/release-desktop.yml:123`

The first hop is correct and I verified it. Job-level `continue-on-error: true` makes `needs.<job>.result` report `success` to downstream jobs, so `create-release` and `publish-updater` do run. That is documented behaviour and it is the subject of actions/toolkit#1739, which complains precisely because you cannot distinguish real success from continued-after-failure through `needs`. The expression form is also fine: `continue-on-error: ${{ matrix.experimental }}` is the canonical example in the GitHub docs, evaluated per matrix leg, so the macOS leg gets `false` and is untouched. `fail-fast: false` at `:100` already prevents leg-to-leg cancellation.

The second hop is missing. `release-desktop.yml` is a `workflow_call` target with two callers, and both gate on the *called workflow run's* conclusion, which is a different quantity from `needs.build-desktop.result` inside it:

- `.github/workflows/nightly-release-train.yml:341`: `(needs.prepare.outputs.desktop != 'true' || needs.release-desktop.result == 'success')` gates `publish-product-release`.
- `.github/workflows/hotfix-production.yml:332`: the same shape gates on `needs.deploy-desktop.result == 'success'`.

The sources genuinely disagree on what a `continue-on-error` job does to the *run* conclusion. The official syntax reference says it "prevents a workflow run from failing when a job fails". Community discussion #15452 reports job red, run green. Ken Muse's widely cited writeup states the opposite, that "the overall workflow will report a failure" while the job result still reads `success` to `needs`. I could not resolve it from documentation alone, and it is not resolvable by reading this diff.

Concrete failure scenario if the pessimistic reading is the real one: nightly train runs, macOS builds and signs cleanly, `create-release` drafts the release, `publish-updater` pushes the macOS updater and installer manifests to S3 and invalidates CloudFront, the Windows leg dies in `pnpm tauri build`, `release-desktop` reports `failure` to the caller, and `publish-product-release` is skipped. The product GitHub Release never gets cut for a release that already shipped to every existing macOS install through the updater. That is a Windows-only failure damaging the macOS release path, which is exactly what the `continue-on-error` line was added to prevent.

This is not a "probably fine". Prove it. Dispatch `release-desktop.yml` on this branch with `dry_run: true, target_os: windows`, let the Windows leg fail (it will, see U1), and read the run conclusion off the Actions tab. If it is red, `continue-on-error` alone is insufficient and the callers need `always()` plus an explicit `!= 'failure'`-style gate, or the Windows leg needs to be a separate non-required job rather than a matrix entry.

### B2. The Windows leg is not opt-in, and the PR is written as though it were

The body frames this as a beta posture. The matrix entry at `.github/workflows/release-desktop.yml:112-115` is unconditional. The filter step at `:137` only skips a leg when `github.event_name != 'push'` *and* `target_os` is set to something other than `all`, and every caller passes `all`:

- `_deploy-desktop.yml:24-26` defaults `target_os` to `all`, and `deploy-staging.yml:233` calls it without overriding.
- `nightly-release-train.yml:251` passes `target_os: all` explicitly, with `publish_updater: true` and `dry_run: false`, described in its own comment as "zero-touch auto-prod ... no approval gate".
- `promote-production.yml:218` and `hotfix-production.yml:301` inherit the same default.

A `desktop-v*` tag push skips the filter entirely, so it always builds Windows.

So on merge, every staging deploy, every nightly production train, every promote and every hotfix starts running a Windows leg that, by the PR's own admission, has never once completed. Combined with B1 that is a live risk to the production desktop release path, not a beta experiment. Either gate the matrix entry on something explicit (a `windows_beta` input, or `github.event_name == 'workflow_dispatch'`), or say plainly in the body that this ships Windows into auto-prod on day one and get that ruled on.

## Verified

**`optional: true` exists, its semantics are what the author assumes, and the negative control is exact.** `scripts/generate-updater-manifest.mjs:88-94` carries it on `darwin-x86_64` with the Intel-paused comment, and `:133-137` is the only consumer: a missing artifact or signature on an optional platform logs `Skipping ${key}` and `continue`s instead of pushing to `errors`. The same key is honoured at `scripts/generate-desktop-installer-manifest.mjs`. I ran `node --test scripts/ci-cd/generate-updater-manifest.test.mjs scripts/ci-cd/generate-desktop-installer-manifest.test.mjs`: 11 pass, 0 fail. I then deleted `optional: true` from the `windows-x86_64` entry in each script and reran: 8 pass, 3 fail, and the three are exactly the ones claimed, `still generates a manifest when the Windows beta leg produced nothing` in both files plus `generates an arm-only manifest while Intel is paused`. Restoring both lines returns all 11 to green. The claim is accurate to the test name.

**All three diagnostics-collector claims are true.** `apps/desktop/src-tauri/src/diagnostics_collector/build_packaging.rs:188` restricts `is_supported_diagnostics_collector_target` to the two Apple triples. `:110-113` writes a placeholder and returns `Ok` for any unsupported target. And critically, `validate_staged_sidecars` at `:61-63` computes `unsupported_collector = binary == "proliferate-diagnostics-collector" && !is_supported_diagnostics_collector_target(&target)` and only errors on `is_placeholder_executable(...) && !unsupported_collector`. So the Windows placeholder is permitted. The `#[cfg(unix)]` executable-mode check at `:66-73` is compiled out on Windows, so the non-executable placeholder passes there too. Dropping the package from the `cargo build` at `:252-272` and gating the staging step at `:308-309` does not fail packaging. Verified, not assumed.

**`shell: bash` on "Validate release title" was the only gap, and there is no other.** I parsed the workflow and enumerated every `run:` step without an explicit `shell`. On the base (#2146) `build-desktop` had three: `Validate release title`, `Import Apple Developer Certificate`, `Write App Store Connect API key`. The latter two are `matrix.runner_os == 'macos'` gated. After this change only those two remain. `create-release` and `publish-updater` run on `ubuntu-latest` where the default shell is already bash. The claim holds exactly.

**`3d3c0504b8` is a pure version bump in the other files.** `git show` confirms `Cargo.lock`, `desktop/package.json`, `desktop/src-tauri/Cargo.toml` and `desktop/src-tauri/tauri.conf.json` are each a single `0.1.12` to `0.1.14` line. One correction: the commit touched *seven* files, not six. `docs/ci-cd/README.md` got the three-line disable note. The body alludes to it in prose but leaves it off the file list, and the note itself is gone from the tree so nothing is stale.

**The NSIS name will carry `x64`.** `apps/desktop/src-tauri/tauri.conf.json` has `productName: "Proliferate"` and `version: 0.4.21` with `@tauri-apps/cli: ^2`. Tauri v2's NSIS output is `{productName}_{version}_{arch}-setup.exe` with `x64` for `x86_64`, giving `Proliferate_0.4.21_x64-setup.exe`, which satisfies `/(x64|x86_64).*setup\.exe$/i` in both matchers, and `Proliferate_0.4.21_x64-setup.exe.sig` satisfies the sig matcher. `pathIncludes(relPath, "desktop-x86_64-pc-windows-msvc")` also holds, because `actions/download-artifact` with `merge-multiple: false` at `:760` and `:800` nests each artifact under its own name. Reasoning about Tauri's naming convention, not observation, so it stays on the unproven list, but the shape is right.

**`actionlint` is clean and introduces nothing.** `actionlint .github/workflows/release-desktop.yml` exits 0. Whole-repo `actionlint` output is byte-identical before and after this diff (one pre-existing SC2129 in `release-e2e-hard-cancel-cleanup.yml`). Note that actionlint is not wired into any CI job, so this is a local check only. `python3 scripts/check_docs.py` (234 files) and `python3 scripts/check_max_lines.py` both pass. All 22 PR checks are green.

## The agent seed: degraded, not fatal

This is the item you flagged as deciding whether a beta is possible. It is not a hard failure. Traced end to end:

`apps/desktop/src-tauri/src/agent_seed_env.rs:47-59`. `bundled_seed_dir` returns `None`, so no `ANYHARNESS_AGENT_SEED_DIR` is set, and the `!cfg!(debug_assertions)` branch sets `ANYHARNESS_AGENT_SEED_EXPECTED=1`.

`anyharness/crates/anyharness-lib/src/domains/agents/installer/seed/mod.rs:127-152`. `Platform::current_target_triple()` returns `Some("x86_64-pc-windows-msvc")`, `expected` is true, `seed_dir` is `None`, so `configured_agent_seed_initial_health()` returns `missing_bundled_seed_health(target)`, status `MissingBundledSeed`.

`runtime.rs:386-393`. `hydration_pending()` is `status == Hydrating`, which is false, so hydration is skipped entirely and control falls straight through to `reconcile_when_idle()`. The doc comment on that function is explicit: "Non-blocking (boots the HTTP server immediately), best-effort (failures are recorded in the seed/reconcile snapshots, never fatal)".

The only consumer of the status is `apps/packages/product-client/src/hooks/telemetry/lifecycle/use-telemetry-agent-seed.ts:47`, which emits a telemetry failure event. Nothing gates a session, an install or a boot on it.

So a Windows build launches, and the startup reconcile pass installs claude and codex from the catalog pins instead of from a seed. Three consequences the PR should state:

1. **This PR has a hard runtime dependency on #2149 and #2145 that the Stacking section does not mention.** With no seed, the pinned installer is the *only* path by which an agent binary reaches the machine. On Windows that installer writes an extension-less file today (`installer/pinned.rs:40` pre-#2149). Merging #2150 without #2149 produces an installer whose agents cannot be installed at all. The body claims the only dependency is #2146.

2. **Even with #2149, no Windows session can start yet.** Claude's `agentProcess` in `catalogs/agents/registry.json` is a `managed_npm_package` from a `git+https` URL, and the thing that actually gets spawned is the managed launcher generated by `integrations/agent_cli/launcher.rs:86`, which is a `#!/bin/sh` script. Windows cannot execute it, and `make_executable` is a no-op there. So the honest statement is not "agents fall back to whatever Node the machine already has", it is "no agent session can start on Windows until the `/bin/sh` launcher work lands". The warning annotation at `:328-330` and the docs paragraph in `guides/deploying/releases.md` both understate this.

3. That "whatever Node the machine already has" is also load-bearing in a way the annotation glosses: without the seed there is no bundled Node, and installing the claude adapter needs both `node`/`npm` and `git` on the machine PATH.

Net: the seed does not make this "do not ship". It makes the resulting installer a shell that installs and launches but cannot run an agent turn. That is a legitimate thing to ship as a build-pipeline restoration; it is not a legitimate thing to describe as a beta a user could try.

## Non-blocking

**N1. Wrong env var name, in two places that ship.** `.github/workflows/release-desktop.yml:329` and `guides/deploying/releases.md` both say `PROLIFERATE_AGENT_SEED_EXPECTED`. The real name is `ANYHARNESS_AGENT_SEED_EXPECTED` (`apps/desktop/src-tauri/src/agent_seed_env.rs:7`, `installer/seed/mod.rs:44`). One is a release-log annotation and one is checked-in docs, so both will be grepped by whoever picks up the seed work and both will miss.

**N2. Release wall-clock.** The Windows leg does a full release `cargo build` of anyharness, worker and debug, then a Tauri release build of the desktop crate, on `windows-latest` with a cold `Swatinem/rust-cache`. Every desktop release now waits on that. Worth measuring on the dry run before this is on the nightly path.

**N3. `Upload native Sentry debug files` at `:533` is ungated.** It runs `pnpm exec sentry-cli debug-files upload` over the whole `target/x86_64-pc-windows-msvc/release` tree, which on MSVC is full of PDBs. It should work and it self-skips when the Sentry env is incomplete, but nobody has measured what it uploads or how long it takes. Not a correctness problem, a surprise-in-the-log problem.

**N4. `Build frontend` writes `/tmp/sentry-desktop-renderer-project.json` at `:390`.** Fine under Git Bash on the runner, worth knowing it is there if that step behaves oddly on Windows.

**N5. `aarch64-pc-windows-msvc` would not compile.** `agent_seed_env.rs:10-30` has no `cfg` arm for it, so `current_target_triple()` would have no return value on that target. Irrelevant today since only `x86_64` is in the matrix, but it is a trap for whoever adds ARM Windows, and #2145 already added `windows_arm64` catalog pins that imply someone intends to.

## Could not verify

Everything the PR lists under Unverified stands, and I add nothing to the list and remove nothing from it. In particular I could not verify: that `pnpm tauri build --bundles msi,nsis` succeeds on `windows-latest` at all; that `resources: ["agent-seeds/"]` bundles when the directory holds only its `.gitignore` (I confirmed that is in fact all it holds); that `pnpm install` and the SDK generate step complete on a Windows runner; and, the important one, the B1 run-conclusion question, which no amount of reading settles.

The dry run the PR recommends is the right acceptance test and it also answers B1 for free. Run it before this is considered again.

pablonyx:
## Verdict

**Safe to merge as-is.** Both blocking findings from round one are resolved by commit `db419a9f10`. I found one place where the PR's own reasoning is unsupported (the Sentry-403 scoping theory) and one small documentation gap (PR #2152 not cited by number), neither of which is blocking.

I am the round-two reviewer of record and wrote none of this.

## B1 — settled by experiment, not redone here

Per the brief: I re-verified run `32442203866` myself. `build-desktop (windows-latest, x86_64-pc-windows-msvc, windows)` job conclusion `failure`, overall run conclusion `success`. Confirmed via `gh run view 32442203866 --json conclusion,status,name,headBranch,event` and the per-job query. This settles the GitHub-docs-vs-Ken-Muse disagreement from round one: `continue-on-error` does not redden the run, so `nightly-release-train.yml:341` / `hotfix-production.yml:332` would have read `success` and proceeded even with Windows broken. The original B1 concern was real, and B2 (below) is what actually neutralizes it for now.

## B2 — the opt-in gate: verified correct, including the boolean trap

Executed: `git diff` of `db419a9f10` against `1e91d65eb3` for every touched workflow file, plus a fresh `git worktree` checkout of `db419a9f10` to run tools against.

**All five entry paths, traced against the actual files, not the PR's list:**

- `deploy-staging.yml:233` → `_deploy-desktop.yml` — no `enable_windows_beta` key in the `with:` block. `_deploy-desktop.yml:27-30` declares it `type: boolean, default: false` on `workflow_call`. Resolves false.
- `promote-production.yml:218-226` → `_deploy-desktop.yml` — same, no override.
- `hotfix-production.yml:301-309` → `_deploy-desktop.yml` — same, no override.
- `nightly-release-train.yml:241-259` → `release-desktop.yml` directly — literal `enable_windows_beta: false` (YAML boolean, not an expression). Confirmed by reading the file at that commit.
- `desktop-v*` tag push — `release-desktop.yml`'s `on:` block (lines 3-6) declares only `push: tags:`, with no `inputs:` under it. `workflow_dispatch`/`workflow_call` are separate trigger blocks. A push run therefore has no populated `inputs` context.

`_deploy-desktop.yml`'s own `release:` job forwards it correctly: `enable_windows_beta: ${{ inputs.enable_windows_beta }}` (`_deploy-desktop.yml:191`), boolean-to-boolean, no default lost in the middle of the chain. Both `workflow_dispatch` and `workflow_call` declarations on `release-desktop.yml` (lines 25-28, 60-63) independently default to `false`.

**The boolean-evaluation question, which is where this class of gate usually breaks:**

The gate is:
```
ENABLE_WINDOWS_BETA: ${{ inputs.enable_windows_beta && 'true' || 'false' }}
...
if [[ "${{ matrix.runner_os }}" == "windows" ]] && [[ "$ENABLE_WINDOWS_BETA" != "true" ]]; then skip=true; fi
```

This coercion pattern (`x && 'true' || 'false'`) is only safe if `x` is never the non-empty string `"false"` — a non-empty string is truthy in GitHub Actions expressions, so `"false" && 'true' || 'false'` would evaluate to `'true'` and silently flip the gate on. This is exactly the failure mode you flagged. It matters here because `github.event.inputs.*` (the older, event-payload context) does stringify booleans for `workflow_dispatch` — but that is not what this code reads.

I fetched GitHub's own docs on this (`events-that-trigger-workflows#workflow_dispatch`): *"The information in the `inputs` context and `github.event.inputs` context is identical except that the `inputs` context preserves Boolean values as Booleans instead of converting them to strings."* I then grepped every touched workflow file for `github.event.inputs`: there are plenty of hits in this repo (`deploy-staging.yml`, `nightly-release-train.yml`, `promote-production.yml`, `hotfix-production.yml`, for unrelated inputs like `dry_run`, `ref`, `reason`), but **zero** hits for `github.event.inputs.enable_windows_beta` anywhere. Every reference — the two `on:` declarations, the `_deploy-desktop.yml` forward, the step-summary echo, and the gate itself — uses the typed `inputs` context. For `workflow_call`, GitHub validates/coerces the passed value to match the declared type, so that leg is a real boolean too, never a string. So: on `workflow_dispatch` default-false, `inputs.enable_windows_beta` is boolean `false` → expression yields `'false'` → skip. On `workflow_call` default-false (three staging/promote/hotfix paths plus the nightly literal), boolean `false` throughout → skip. On a bare tag push, `inputs` has no populated properties at all, so `inputs.enable_windows_beta` is `null` (falsy) → expression yields `'false'` → skip. The only way to get `'true'` is a human explicitly setting the workflow_dispatch checkbox, which is the `inputs` context's real-boolean `true`.

This isn't just a reading of the docs. Run `32442203866` was dispatched with `enable_windows_beta: true` and the Windows leg actually ran (it reached `Build Rust binaries for target`, `Generate and build SDK`, and failed only at `Build frontend` — see below), which is live confirmation the true-path resolves correctly through this exact expression. I did not find a dispatched run with the field left at its default to directly observe the false-path executing, but the mechanism (typed `inputs` context, never the stringifying `github.event.inputs`) is unambiguous from the docs plus the grep, and it's the same mechanism the pre-existing (not part of this diff) `TARGET_OS: ${{ inputs.target_os || 'all' }}` line at the same call site already relies on — which has been resolving correctly on `desktop-v*` tag pushes in production for months (e.g. runs `26680628644`, `26678014280`, `26623763241`, all `success`). Every downstream Windows-only step (`Build Rust binaries`, `Build frontend`, `Build Tauri app`, the Windows upload step, etc.) is gated on `steps.filter.outputs.skip != 'true'`, consistent with the pre-existing pattern — I grepped all `if:` conditions in the job and there's no step that bypasses the filter.

**Conclusion on B2: the gate is correctly implemented and is not vulnerable to the trap you asked me to check for**, specifically because it never touches `github.event.inputs`.

## What the dry run proved and did not prove

Executed: pulled the job/step list for run `32442203866` via `gh api .../jobs`.

Confirmed sequence exactly as described: `Build Rust binaries for target` → success, `Generate and build SDK` → success, `Build frontend` → **failure**, `Build Tauri app` → skipped, everything after → skipped. Pulled the literal failure line from `gh run view 32442203866 --log-failed`:

```
##[error]Sentry desktop renderer project 'proliferate-desktop' is not accessible in org 'proliferate' (HTTP 403). Update SENTRY_DESKTOP_RENDERER_PROJECT before releasing.
```

So NSIS packaging never ran and the `x64`-infix naming assumption is still unproven — the PR body states this plainly under "Unverified" and doesn't oversell the dry run. Good.

**The Sentry-403 explanation in the PR body doesn't hold up, and this is worth flagging even though it's non-blocking.** The PR hedges that the 403 is "more likely" a secret-scoping artifact of a bare `workflow_dispatch` versus environment-scoped secrets on the gated callers. I checked: `release-desktop.yml`'s jobs carry no `environment:` key at all — the only `environment: ${{ inputs.environment }}` binding in this call chain is on `_deploy-desktop.yml`'s `plan` job (line 42), which only gates `vars.DESKTOP_DEPLOY_ENABLED`, not the `release` job that actually calls `release-desktop.yml`. So every path — staging, promote, hotfix, nightly, and a bare manual dispatch — feeds `release-desktop.yml` the exact same unscoped, repository/org-level `secrets: inherit` chain. There is no environment-scoping difference to explain the failure. As a direct check, I pulled a real production run from the same day: hotfix-production run `32332438576` (2026-08-20T04:35Z, ~23 hours before the dry run) ran `build-desktop` for both macOS legs through this identical chain, and `Build frontend` succeeded on both. Same secrets, same project slug (`proliferate-desktop`/`proliferate`), no environment scoping either way, passing hours earlier on macOS and failing on Windows in the dry run. I can't tell you what actually caused the 403 (transient Sentry hiccup, `windows-latest`-runner-specific networking quirk, something else) — but it is not the scoping explanation offered, and whether it's Windows-runner-specific enough to recur on a real production Windows build is genuinely still open. Recommend the PR drop or soften that specific claim rather than pointing whoever debugs this next at a scoping difference that doesn't exist in the diff.

## Non-blocking fixes: verified

- `.github/workflows/release-desktop.yml:350` (line moved from `:329` because of earlier insertions in the same file): `ANYHARNESS_AGENT_SEED_EXPECTED`, `PROLIFERATE_AGENT_SEED_EXPECTED` is gone from this file. Confirmed against `apps/desktop/src-tauri/src/agent_seed_env.rs:7` and `anyharness/.../installer/seed/mod.rs:44`, both of which declare the constant as `ANYHARNESS_AGENT_SEED_EXPECTED`.
- `guides/deploying/releases.md`: grepped for `AGENT_SEED_EXPECTED`, only the corrected `ANYHARNESS_...` form remains. The "falls back to whatever Node the machine already has" sentence is gone, replaced with the accurate "no fallback ... no agent session can start" statement.
- `apps/desktop/src-tauri/src/agent_seed_env.rs:31-34`: `#[cfg(all(target_os = "windows", target_arch = "aarch64"))]` arm added, returning `"aarch64-pc-windows-msvc"`. Grepped the rest of the codebase for that exact triple string: `anyharness/.../agents/model.rs:146,162`, `apps/desktop/src-tauri/build.rs:591`, `catalogs/agents/registry.json:171` all use the identical string. The new arm matches, and it follows the same one-cfg-block-per-arm structure as the five existing arms in the same function, so it doesn't change how the function compiles for any currently-shipping target. Not compiled/tested (no Rust build run, per instructions and since this PR is workflow/docs-only in substance) — verified by reading and cross-reference only.

## Dependency disclosure: mostly there, one gap

The "Hard runtime dependency, stated plainly" section states the dependency on #2149 and #2145 without softening, and separately states — also plainly — that "no Windows session can start" until the `#!/bin/sh` managed-launcher fix lands, tracing through `installer/pinned.rs`, `registry.json`'s `managed_npm_package`, and `integrations/agent_cli/launcher.rs:86`. That substance is accurate and not hedged.

Gap: it never cites the launcher fix by PR number. It says "a separate PR is in flight for the `/bin/sh` managed-launcher work," and the "Stacking" section at the bottom names #2149 and #2145 but not the launcher PR either. `gh pr view 2152` confirms it's open now (`fix(agents): emit a Windows-executable managed launcher`). Worth a one-line edit to link it — as written, a reader has to go find it themselves, which undercuts the "stated loudly rather than stubbed silently" framing the rest of the PR is going for. Not blocking.

## actionlint: reproduced

`actionlint 1.7.12` (Homebrew), run from a detached worktree at `db419a9f10` (no repo-wide side effects, no Rust involved):

```
$ actionlint .github/workflows/release-desktop.yml .github/workflows/_deploy-desktop.yml \
    .github/workflows/nightly-release-train.yml .github/workflows/deploy-staging.yml \
    .github/workflows/promote-production.yml .github/workflows/hotfix-production.yml
EXIT=0
```

Whole-repo:
```
$ actionlint
.github/workflows/release-e2e-hard-cancel-cleanup.yml:46:9: shellcheck reported issue in this script: SC2129:style:8:1: Consider using { cmd1; cmd2; } >> file instead of individual redirects [shellcheck]
   |
46 |         run: |
   |         ^~~~
EXIT=1
```

Diffed against a second worktree at the PR's base (`38cd6a2d15`, tip of `fix/sdk-generate-windows-safe`): byte-identical single finding, same file, same line, same rule. Confirmed the PR introduces zero new actionlint findings.

## Summary

B1 and B2 are both real fixes, not paperwork. The boolean-input gate specifically checks out against the exact trap you asked about — it's built entirely on the typed `inputs` context, never the stringifying `github.event.inputs`, and that's verifiable both from GitHub's own docs and from a grep showing zero uses of the dangerous context on this input anywhere in the touched files. The two items I'd still want addressed, at your discretion and non-blocking: correct or drop the Sentry-403 scoping claim, and cite #2152 by number instead of "a separate PR is in flight."

pablonyx:
## Verdict

**Safe to merge as-is.** No blocking issues from the rebase. Two body-accuracy items are still outstanding from earlier rounds/new evidence — non-blocking, but should be fixed before this lands since the body is the durable record of the beta decision.

I am the post-rebase reviewer of record and wrote none of the PR itself.

## 1. Rebase cleanliness — clean, with one file-rename relocation worth naming explicitly

`git range-diff 933584620d..db419a9f10 296e6fc428..83333cfe38` shows exactly two hunks of drift, both in the second commit (`db419a9f10` → `83333cfe38`), and both are the *same* change: the `nightly-release-train.yml` hunk relocated to `release.yml`.

This is not cosmetic. `nightly-release-train.yml`, `promote-production.yml`, and `hotfix-production.yml` were all deleted on main by #2140 ("unify deploys into one main-to-prod pipeline"), which landed in this rebase window — confirmed `95626925f5` is an ancestor of the new base `296e6fc428` but not of the old base `933584620d`. Rebase followed the rename and reapplied the `enable_windows_beta: false` hunk onto the single surviving `release-desktop` job in `release.yml:321-343`, matched on the "no approval gate on this step" comment context.

Checked whether this landed on the *right* job, since #2140 folded nightly/hotfix/promote into one workflow with one `prepare` job feeding several downstream jobs: there is exactly one `release-desktop:` job in `release.yml` (line 321), gated on `needs.prepare.outputs.skip_build != 'true' && ... dry_run != 'true' && ... desktop == 'true'`, and it's the only call site for `release-desktop.yml` from this file. `enable_windows_beta: false` at line 343 is hardcoded and **`release.yml`'s own `workflow_dispatch` doesn't expose an `enable_windows_beta` input at all** (only `surfaces`, `skip_build`, `ref`, `dry_run`), so there's no way to flip it true through this path even by mistake. That's strictly safer than the pre-unification shape, not a regression.

Also checked `956d2426b7` (#2156, the other release-workflow-touching commit in this window) for interference: it only touches `release.yml`'s `prepare` job "Commit version bumps" step (lines ~206-244, the `RELEASE_PUSH_TOKEN` push-auth change), nowhere near the `release-desktop` job block at line 321. No overlap, no semantic conflict.

## 2-3. The opt-in gate — survived intact, re-verified at new head

- Gate expression, `release-desktop.yml:153`: `ENABLE_WINDOWS_BETA: ${{ inputs.enable_windows_beta && 'true' || 'false' }}`, matched against at line 159. Typed `inputs` context throughout, as before.
- `_deploy-desktop.yml:191` forwards `enable_windows_beta: ${{ inputs.enable_windows_beta }}` boolean-to-boolean, unchanged.
- `deploy-staging.yml` → `_deploy-desktop.yml` (line ~212): no override, resolves to the `default: false` declared at `_deploy-desktop.yml:27-29`.
- The former nightly-train literal `false` now lives at `release.yml:343`, confirmed above — still explicit, still `false`, comment intact and still accurate to its new home.
- `hotfix`/`promote` no longer exist as separate callers; both are now input shapes (`ref`, `skip_build`) into the same single `release-desktop` job in `release.yml`, which carries the same hardcoded `false`. All five original entry paths still resolve to skip.
- Confirmed `#2156` didn't touch `release-desktop.yml` or `_deploy-desktop.yml` at all (its diff is 3 files: `release.yml`, `guides/deploying/releases.md`, one test file).

No regression. The gate is if anything harder to accidentally flip post-rebase than it was pre-rebase.

## 4. NSIS naming — body is now stale, should be updated (non-blocking)

`release-desktop.yml` body currently still lists under "Unverified" (line 95): *"that NSIS emits a name matching `/(x64|x86_64).*setup\.exe$/i`, which both manifest matchers depend on"*, and repeats at line 151: *"still exactly as unproven as before this run."* Per the new evidence (run `32447570745`, Tauri's own bundler log producing `Proliferate_0.4.21_x64-setup.exe` / `.msi` with `.sig` files), this is now proven, not just reasoned-about convention. The body should move this line out of "Unverified" into a confirmed section citing the run. Not blocking — it doesn't affect any code path — but it's exactly the kind of "verified vs. assumed" bookkeeping the rest of this PR is careful about, and leaving it stale undersells the actual state.

## 5. Sentry 403 claim — still present, and now demonstrably wrong, not just unsupported (non-blocking per round 2's classification, but should be fixed)

Round 2 already flagged this as unsupported ("the scoping difference... doesn't exist in the diff") and recommended dropping it; the claim is still in the body verbatim at the "What actually failed" section, still hedging toward secret-scoping as "more likely... than a broken token."

Re-checked the environment-binding claim myself: grepped every workflow on the call path (`release-desktop.yml`, `_deploy-desktop.yml`, `release.yml`, `deploy-staging.yml`) for `environment:`. The only binding in the desktop chain is `_deploy-desktop.yml:42`, on the `plan` job — which only gates `vars.DESKTOP_DEPLOY_ENABLED` — not on the `release` job (`_deploy-desktop.yml:177-183`) that actually calls `release-desktop.yml`. So the scoping theory was never live anywhere in this chain, confirming round 2's finding.

Now we also know it wasn't a scoping question at all: `SENTRY_AUTH_TOKEN` was rotated 2026-08-20T19:51:01Z and cannot read `proliferate-desktop` on any platform. The body's hedge — treating "a broken token" as the less likely explanation — is now the wrong conclusion by the confirmed cause, not just an unsupported one. Should be corrected to state the rotation as the cause rather than left as an open "more likely" comparison.

## Other open items, not re-litigated

B1/B2 stand as resolved by rounds one and two; nothing here reopens them. Round 2's other non-blocking item (citing #2152 by number instead of "a separate PR is in flight," `release-desktop.yml`'s body reference near the agent-seed section) is also still outstanding but unaffected by the rebase — same status as when round 2 left it.

## Summary

No functional or gating regression from the rebase, including through the #2140 file-unification and #2156 push-token change. Merge-safe on the code and workflow logic. Recommend a quick body pass before merge: promote the NSIS-naming line out of "Unverified," and replace the Sentry-403 scoping hedge with the confirmed rotation cause — both are text-only, neither blocks CI or this diff's correctness.

pablonyx:
## Merge hazard: this PR now auto-merges into a broken generator, with no conflict markers

Flagging before merge, not a review of the PR's own contents.

`#2170` (merged as `dedb89187`) deleted the `pathIncludes(relPath, segment)` helper from both manifest generators and rewrote every matcher to test filenames only. This PR's Windows hunks still call `pathIncludes(...)`. Git resolves that combination **cleanly**, because main's deletion of the helper and this branch's addition of new call sites touch different lines.

Proof against current main:

```
$ git merge-tree --write-tree origin/main origin/fix/desktop-windows-release
357605a593c4a8193e03374fe37c10ce1ae0a77a
$ git show 357605a5:scripts/generate-updater-manifest.mjs | grep -c 'function pathIncludes'
0
$ git show 357605a5:scripts/generate-updater-manifest.mjs | grep -n 'pathIncludes'
109:      pathIncludes(relPath, "desktop-x86_64-pc-windows-msvc") &&
112:      pathIncludes(relPath, "desktop-x86_64-pc-windows-msvc") &&
```

Zero helper definitions, two call sites. The matcher is invoked while scanning artifacts for the Windows platform entry, which happens on every manifest run, so this is a `ReferenceError` on every desktop release including mac-only ones. `git merge-tree` reports conflicts only in the `Makefile` and the two `scripts/ci-cd/*.test.mjs` files, so resolving exactly what git flags and committing would ship the crash.

Two further reasons these hunks should not merge: they retain the pre-`#2168` OR-based optional skip (`sigFiles.length === 0 || artifactFiles.length === 0`), which silently skips a platform that published its artifact but not its signature, and they add no test for that partial-artifact case.

**Recommended resolution: drop this PR's hunks in `scripts/generate-updater-manifest.mjs` and `scripts/generate-desktop-installer-manifest.mjs` entirely when rebasing.** `#2168` (head `4dfc4a162`) already carries the Windows entries for both generators, adapted to main's filename-only matchers and to the flat artifact layout `#2170` made the production layout, plus the both-absent fix and a test that pins it.

One more thing worth a look, unrelated to the above: this PR's Windows `upload-artifact` step uploads `*.msi` alongside `*-setup.exe`, and with `#2170`'s `merge-multiple: true` the MSI now lands flat beside the darwin artifacts. The `-setup.exe` anchor in the matchers is the only thing keeping it out of the manifests.

pablonyx (commented):
Verdict: BLOCKERS (process, not code). The intended code change is clean; the PR itself needs a retarget before merge.

**Process blocker: PR base ref is stale, not `main`.**

`gh pr view` shows `baseRefName: fix/sdk-generate-windows-safe`, not `main`. That branch's own PR (#2146) was squash-merged into main as `6c97bc895`, so the raw branch commits (`129e04692`, `516b34e95`, `ad096409d`) were never fast-forwarded and the branch is now orphaned relative to main. Because the PR base still points at that stale branch, `gh pr diff 2150` and GitHub's own diff view show 29 files including `Makefile`, `anyharness/sdk/package.json`, and a pile of `scripts/` files that have nothing to do with this change. That is the literal root cause of the one failing check: `Validate PR title and labels` fails with `Changed paths require area label(s): area:anyharness, area:sdk` because the label validator is evaluating that bogus diff.

I recomputed the diff directly against `origin/main` using the actual merge-base (`08483d20a`, which is on `main`'s history) instead of trusting the PR's declared base:

```
git diff 08483d20a origin/fix/desktop-windows-release --stat
 .github/workflows/_deploy-desktop.yml        |  6 ++
 .github/workflows/release-desktop.yml        | 93 +++++++++++++++++++++++---
 .github/workflows/release.yml                |  7 +++
 apps/desktop/src-tauri/src/agent_seed_env.rs |  4 ++
 guides/deploying/releases.md                 | 19 +++++-
 5 files changed, 118 insertions(+), 11 deletions(-)
```

Exactly the 5 files expected, matching the described intent, with no `scripts/`, `Makefile`, or `anyharness/sdk/package.json` change. Fix: `gh pr edit 2150 --base main` (or close/reopen after the retarget) so GitHub recomputes the diff and re-runs the label check against the real change.

**Code review of the real diff (against main), all clean, no blockers found:**

1. Gate default: `enable_windows_beta` defaults to `false` in both `release-desktop.yml` (`workflow_dispatch` and `workflow_call` inputs) and `_deploy-desktop.yml`'s `workflow_call` input, and `_deploy-desktop.yml` passes it through unmodified. The only two call sites into `release-desktop.yml` are `release.yml` (explicit `enable_windows_beta: false` on the zero-touch prod job) and `deploy-staging.yml` via `_deploy-desktop.yml` (relies on the `false` default, doesn't set it). No caller can produce `true` today.
2. Non-Windows regression: the skip-decision step ORs two independent conditions into `skip`, so Windows is skipped whenever `enable_windows_beta` isn't literally `true`, regardless of `target_os`. All macOS-only steps (signing, App Store key, DMG verification, artifact upload) were already conditioned on `matrix.runner_os == 'macos'` pre-PR and are untouched. The cargo-build step's macOS branch is byte-identical to the old unconditional command, just moved into an `else`. Mac path is unaffected with the gate off.
3. `continue-on-error: ${{ matrix.runner_os == 'windows' }}` is present and correctly scoped to only the Windows leg; `false` for macOS is a no-op.
4. Artifact/manifest interplay verified by reading `scripts/generate-updater-manifest.mjs` and `scripts/generate-desktop-installer-manifest.mjs` on main: both match Windows only via `/(x64|x86_64).*-setup\.exe$/i` and are `optional: true`. `.msi` never matches any platform matcher (darwin requires `.app.tar.gz$`), so the uploaded `*.msi` cannot be picked up by either generator or collide with darwin filenames. `publish-updater`'s S3 upload step also explicitly filters by `-name "*.dmg" -o "*.app.tar.gz" -o "*.app.tar.gz.sig" -o "*-setup.exe" -o "*-setup.exe.sig"`, so `.msi` never reaches the downloads bucket/updater feed; it only lands in the GitHub Release via the unfiltered `files: all-artifacts/**/*`, which is correct (the `.msi` is the human-facing installer download, `-setup.exe` is the updater artifact).
5. `agent_seed_env.rs`: the added `aarch64-pc-windows-msvc` arm is purely `cfg`-gated inside `current_target_triple()`, compiled out entirely on macOS/Linux, no secret material, no behavior change off Windows. The CI matrix only builds `x86_64-pc-windows-msvc`; this arm just keeps the function exhaustive for a Windows-arm64 host, consistent with the existing pattern for the other 5 platform arms in that function.
6. No new secrets introduced. Existing `secrets: inherit` and existing Apple secrets references are untouched; new steps (gap-reporting, Windows artifact upload) reference no secrets.
7. Shell sanity: every `run:` step in the `build-desktop` job (including the pre-existing ones) now has an explicit `shell: bash`, including the newly-added `shell: bash` on "Validate release title" which was previously missing that would have defaulted to PowerShell on `windows-latest` and broken on its bash-syntax body. No PowerShell used anywhere; consistent Git Bash usage across the whole job including the new Windows leg.
8. `guides/deploying/releases.md` changes accurately describe the workflow as implemented (opt-in default, continue-on-error, optional manifest platforms, no diagnostics collector, no agent seed on Windows, unsigned installer). No factual drift found.

**CI status on head `2ef3fbe`:** 32 checks total, 24 pass, 7 pending, 1 fail. The 1 failure is `Validate PR title and labels`, root-caused above to the stale base ref, not a real label gap for the actual diff (which only touches `area:desktop`/`area:docs`/`area:release`, all of which the PR already carries).

Net: retarget the PR to `main` to clear the false-positive label failure and get an accurate diff view; once retargeted, I see nothing in the actual change that blocks merge.

PR #2151, opened 2026-08-21
fix(workflows): omit empty controlValues from stored and frozen v2 definitions

tairabun:
## Summary

- A gen-2 workflow saved with any per-node model override could not start: the trigger PUT to the local runtime failed with `WORKFLOW_SNAPSHOT_INVALID` and the dialog showed "This workflow cannot run as saved" even though the definition was valid and saved cleanly.
- Root cause: the CP's pydantic dumps emitted `controlValues: {}` for every node model because `exclude_none` does not exclude empty defaults, while the runtime's serde serializer skips empty maps (`skip_serializing_if`) and parses definitions with `deny_unknown_fields`. Installed runtimes built before the `modeId` → `controlValues` rename (#2070) therefore reject the frozen snapshot outright.
- Fix, at the two explicit normalization seams (no pydantic schema change, so the generated OpenAPI artifacts are untouched):
  - `WorkflowDefinitionDocumentV2.document_json()` — the stored/frozen normal form — now omits an empty control-values map, matching the runtime serializer.
  - New `WorkflowInvocationResponseV2.frozen_json()` — the wire dump with the definition in its `document_json` form — used by the invocation freeze in `service_v2` and by the hand-serialized invocation PUT/GET endpoints. Because the freeze and the responses re-normalize, definitions and invocations stored **before** this rule also heal at trigger/delivery time — no re-save needed.
- Extends the cross-plane contract fixtures (`v2-full.json`, `run-snapshot-v2.json`) with a controls-free node model, so the byte-for-byte round-trip tests pin the omission on the CP plane and the parse/validate tests pin tolerance on the runtime and client planes.
- One deliberate deviation, called out in `WORKFLOWS.md`: for pre-fix stored rows the frozen definition is the stored bytes minus the CP-injected empty default, not byte-identical custody. New saves remain byte-verbatim.

Follow-ups (not this PR): the runtime's `deny_unknown_fields` on the definition makes every additive CP schema field a hard break for older installed runtimes (needs an ADR); the trigger dialog drops the ProblemDetails `detail` and tells users to fix steps in the editor even for repo-root/schema failures.

## Testing

- Unit (CP): `tests/unit/test_workflow_definition_v2_validation.py` — new `test_empty_control_values_stay_omitted_in_the_stored_form`; the run-snapshot round-trip now pins `frozen_json()`; the fixture round-trips fail without the fix (37 passed).
- Integration (CP, real PostgreSQL): `tests/integration/test_workflow_invocations_v2_api.py` — new `test_v2_invocation_freeze_drops_a_stored_empty_control_values_map` seeds a pre-fix row shape and proves the frozen snapshot omits the key; existing freeze-contract byte-for-byte tests re-pin the edited fixtures (10 passed with `test_workflow_definitions_v2_api.py`).
- Contract corpus (runtime): `cargo test -p anyharness-lib contract_fixture` — the edited fixtures parse and validate (8 passed).
- Contract corpus (client): `pnpm vitest run src/domain/workflows/definition-v2.test.ts` — the three-plane exemplar still validates (27 passed).
- Generated artifacts: `make cloud-client-generate` produces zero diff in `cloud/sdk/src/generated/openapi.ts`.
- Live reproduction before the fix: replaying the affected saved definition against an installed runtime (anyharness@0.1.0+49ac4e79551f) returned `400 WORKFLOW_SNAPSHOT_INVALID: unknown field controlValues`; the identical snapshot without the key passed deserialization and validation.

## Observability

- none — the failure surface (`WORKFLOW_SNAPSHOT_INVALID` ProblemDetails) is unchanged; this removes the input that triggered it.

## Readiness

  in this PR.

## Verification

- `make lint-server` — ruff check, ruff format, mypy shrink-only ratchet all pass.
- `python3 scripts/check_docs.py` — 234 Markdown files pass.

### DIFF
--- a/fixtures/contracts/workflow-definition/run-snapshot-v2.json
+++ b/fixtures/contracts/workflow-definition/run-snapshot-v2.json
@@ -23,7 +23,10 @@
         "id": "n_synthesize",
         "type": "agent",
         "title": "Synthesize a proposal",
-        "prompt": "Read @doc:research-findings and write a proposal into @doc:proposal covering @input:topic."
+        "prompt": "Read @doc:research-findings and write a proposal into @doc:proposal covering @input:topic.",
+        "model": {
+          "agentKind": "codex"
+        }
       },
       {
         "id": "n_review",
--- a/fixtures/contracts/workflow-definition/v2-full.json
+++ b/fixtures/contracts/workflow-definition/v2-full.json
@@ -24,7 +24,10 @@
         "id": "n_synthesize",
         "type": "agent",
         "title": "Synthesize a proposal",
-        "prompt": "Read @doc:research-findings and write a proposal into @doc:proposal covering @input:topic."
+        "prompt": "Read @doc:research-findings and write a proposal into @doc:proposal covering @input:topic.",
+        "model": {
+          "agentKind": "codex"
+        }
       },
       {
         "id": "n_review",
--- a/server/proliferate/server/workflows/api.py
+++ b/server/proliferate/server/workflows/api.py
@@ -195,7 +195,7 @@ async def put_workflow_invocation_endpoint(
         response_v2 = workflow_invocation_response_v2(result.value)
         return JSONResponse(
             status_code=status.HTTP_201_CREATED if result.created else status.HTTP_200_OK,
-            content=response_v2.model_dump(by_alias=True, mode="json", exclude_none=True),
+            content=response_v2.frozen_json(),
         )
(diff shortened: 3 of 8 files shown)


tairabun:
@greptileai review

tairabun:
Superseded by #2192 — identical commit, re-opened from an upstream-hosted branch so the Vercel deployment check can run with team authorization (fork-head PRs always fail it).

PR #2152, opened 2026-08-21
fix(agents): emit a Windows-executable managed launcher

pablonyx:
## The defect and its blast radius

`generate_launcher_script` (`anyharness/crates/anyharness-lib/src/integrations/agent_cli/launcher.rs:79`) unconditionally writes:

```
#!/bin/sh
set -e
export PATH='<prefixes>':"$PATH"
export K='v'
exec "<exec_path>" args "$@"
```

Windows cannot execute that: no `#!` interpreter line, no `export`, no `exec`. Three production call sites go through this function or its atomic wrapper (`generate_launcher_script_atomic`, `launcher.rs:22`): `installer/pinned.rs:214`, `installer/npm.rs:88`/`:182`, and `installer/agent_process.rs:75`.

`claude` is a `managed_npm_package` agent (`catalogs/agents/registry.json`), so on Windows the thing actually spawned for a claude session is this launcher. The installer can succeed and the resolved artifact can report `installed: true`; the first attempt to start a session fails because there is nothing that can execute the launcher file. This is the difference between "the Windows installer works" and "a beta user can try the product", it's the last blocker in that path.

## What changed

`generate_launcher_script` now dispatches on `#[cfg(windows)]` / `#[cfg(not(windows))]` to one of two builders that both return the script as a `String` (`build_unix_launcher_script`, `build_batch_launcher_script`); the write + `make_executable` happens once, after the dispatch, so both arms share that plumbing.

## Naming decision

The managed launcher's file name is now decided in exactly one place, `managed_launcher_file_name(kind_str)` in `launcher.rs`, which every write site and read site calls instead of hand-rolling `format!("{kind}-launcher")`:

- Unix: `<kind>-launcher` (unchanged, a shebang script needs no extension).
- Windows: `<kind>-launcher.cmd`.

**Why rename instead of the `cmd.exe /C` alternative:** Rust's `std::process::Command` (and `tokio::process::Command`, which wraps it) has documented special-case handling on Windows: when the target program's extension is `.bat` or `.cmd`, it transparently spawns via `cmd.exe /c` itself (this is also why CVE-2024-24576 exists, it's real, load-bearing std behavior, not something I'm assuming). That means the `.cmd` rename requires **zero** changes at the spawn call site (`live/sessions/driver/process.rs:93`, `tokio::process::Command::new(spawn_program)`), the existing `Command::new(launcher_path)` call keeps working unmodified once the path ends in `.cmd`. The alternative (keep the extension-less name, have every caller spawn via `cmd.exe /C`) would need to touch the spawn call site and would duplicate work std already does for us, for no benefit. I picked the rename.

**Why the rename is safe / contained:** every read site was found by grepping for `-launcher` and for `launcher_path` across the crate and traced individually:

1. `installer/pinned.rs:113` (write, `install_agent_process_from_pin`): now calls `managed_launcher_file_name`.
2. `installer/agent_process.rs:39` (write, `regenerate_agent_process_launcher`, the seed-reconciliation path): now calls `managed_launcher_file_name`.
3. `readiness/artifacts.rs:173` (`managed_launcher_candidates`, the **read/resolve** site): this is what actually populates `ResolvedArtifact.path`, which `spawn_agent_process` (`live/sessions/driver/process.rs:65`) uses as the spawn target when there's no `SpawnSpec` override. Now calls `managed_launcher_file_name`.
4. `readiness/compatibility.rs:81`: a cosmetic fallback string used only in a user-facing error message (`claude_launch_requires_node`), not in any path resolution. Updated for message accuracy, no behavior effect.
5. `readiness/artifacts.rs:381` (`launcher_uses_binary_hint`): reads the launcher's **contents**, not its name. **Correction from an earlier revision of this PR body:** I originally claimed the windows-shaped addition here (`"<binary>" ...`) prevented a diagnostic from going dark on `.cmd` launchers. Review caught that this claim is false. Generated launchers always `exec`/spawn an **absolute path** (`exec "/managed/dir/cursor-agent"` on unix, `"C:\managed\dir\cursor-agent.exe"` on windows), never the bare `candidate_binaries` name, so `contents.contains(&format!("exec \"{binary}\""))` can never match: there is always a directory prefix immediately before the opening quote. That was already true of the pre-existing **unix-only** check before this PR touched the function at all; it is a pre-existing latent gap in `launcher_uses_binary_hint`, not something this PR introduces or fixes. My windows-shaped addition is equally inert, for the identical reason, and is kept only for symmetry with the (already broken) unix branches: it does not restore the diagnostic on `.cmd` launchers. Real behavior effect of this hunk: **none**. Left as a documented follow-up (code comment added) rather than fixed here, since a real fix means matching on the exec target's file name/stem, not string-searching for the bare candidate name.
6. Two *staging*-file names: `installer/npm.rs:180` (`.{name}-launcher.next`) and `installer/pinned.rs:212` (`.{name}-launcher.next`, archive arm), construct their own literal, extension-less staging names. These are **never executed directly**; `ArchiveTreeActivation::activate_launcher` (`installer/downloads/activation.rs:96-119`) renames the staged file onto `self.paths.launcher_path`, which is derived from the real `launcher_path` variable at each call site (already carrying `.cmd` via `managed_launcher_file_name`). Confirmed by reading `activate_launcher` and `activation_paths`. Left unchanged; only the *destination* name matters.
7. Test-only literal paths (`readiness/service_tests.rs`, `resolution_flip_tests.rs`, `installer/tests.rs`, `installer/downloads_tests.rs`, `installer/reconcile/surface_threading_tests.rs`, `api/router_tests.rs`, `installer/seed/tests.rs`, and `launcher.rs`'s own pre-existing test) all construct their expectation strings independently and run on the Linux CI host, where `managed_launcher_file_name` still returns the extension-less form, bit-for-bit identical to before. No changes needed; verified by grepping every remaining `-launcher` occurrence in the crate.
8. `is_valid_executable`/`make_executable` (`integrations/agent_cli/executable.rs`) are already `cfg(unix)`/`cfg(not(unix))`-split and extension-agnostic (`path.is_file()` on non-unix), no change needed.

I believe this list is complete: it is every result of `grep -rn '\-launcher\|launcher_path'` across `anyharness-lib/src`, individually triaged above.

## Batch-escaping decision

`shell_escape` is POSIX-shell quoting and is wrong for batch, so it is **not reused**. A separate `batch_escape`/`batch_escape_arg` pair handles the two characters that are actually load-bearing in a `.cmd` file:

- `%` triggers variable expansion even inside a quoted `set "K=V"`, doubled to `%%` (the standard, documented batch escape for a literal percent).
- `"` cannot be safely embedded in a quoted batch value at all, cmd.exe has no backslash-style quote escape the way POSIX shells do. Rather than emit a script that would tokenize differently than intended, `batch_escape` **fails closed**: it returns `LauncherError::UnsupportedBatchValue`, which `installer/service.rs`'s `From<LauncherError> for InstallError` maps to `InstallError::InvalidInstallSpec`. In practice this can only be hit by a catalog-declared env value or ACP arg containing a literal `"`, which none of the current registry entries do.
- `!` is deliberately **not** escaped, it's only special under `setlocal enabledelayedexpansion`, which this script never enables.

**Known limit, disclosed rather than fixed here: `batch_escape` rejects `"` but silently passes `\r`/`\n`, and the env *key* is never escaped at all.** Reading `batch_escape` (`launcher.rs:249-254`) directly: it only checks `s.contains('"')`; a newline or carriage return in an env value or arg passes through untouched. Because cmd.exe parses a batch file line by line, a quoted `set "K=V"` cannot contain an embedded newline safely, the same hazard class the always-quote/reject-quote fix above was made for. And the `for (key, value) in env` loop (`launcher.rs:219-221`) runs `value` through `batch_escape` but writes `key` straight into `set "{key}=..."` with no escaping at all. Reachability is low: `env` comes from `built_in_registry()` → `bundled_agent_descriptors()`, and that registry/catalog lane is minisign-verified with baked pubkeys, so today this is publisher-controlled, not attacker-controlled. That said, `registry/validation.rs` only checks that an env var name is non-empty, with no charset constraint on either name or value, so nothing in the schema itself prevents it. Not fixed in this PR; disclosing it rather than leaving it silent.

**Round-2 review fix, closed coverage hole.** The reviewer ran twelve mutations against this PR's test suite; eleven went red, one stayed green: `batch_escape_arg` keeping its whitespace-only quoting decision but losing the `%`/`"` handling entirely was undetected, because every existing `%`/`"` assertion drove the value through an ENV var (`build_batch_launcher_script`'s `env` map), never through `extra_args`. Fixed two ways:

- `batch_escape_arg` no longer has a whitespace-conditional branch at all, it now **always** wraps the argument in quotes (`Ok(format!("\"{escaped}\""))`), the same way the exec path itself is always quoted. This was a second, independent finding: an unquoted single-word arg containing `&`, `|`, `<`, `>`, `^`, `(`, or `)` is still a cmd.exe metacharacter hazard even with no whitespace, `shell_escape`'s whitespace-only heuristic is correct for POSIX (those characters are inert without word-splitting) but wrong for cmd.exe (they're metacharacters unconditionally). No catalog `args` entry exercises this today (`[]`, `["acp"]`, `["agent", "stdio"]`), but the catalog is remotely syncable, so that's a property of today's data, not of the code, always-quote removes the hazard instead of enumerating the metacharacter set.
- Added `batch_escape_arg_still_doubles_percent_and_rejects_quote` and `batch_script_drives_percent_and_quote_through_the_argument_path`, which drive `%` and `"` through the **argument** path specifically (the exact gap the mutation found), plus `batch_escape_arg_always_quotes_even_a_single_word_token` and renamed `batch_script_always_quotes_args_even_without_whitespace` for the always-quote requirement.

Negative control for this specific fix (reverted `batch_escape_arg` to the old whitespace-only branch, everything else unchanged):

```
test integrations::agent_cli::launcher::tests::batch_escape_arg_always_quotes_even_a_single_word_token ... FAILED
  left: "--acp"
 right: "\"--acp\""
test integrations::agent_cli::launcher::tests::batch_escape_arg_still_doubles_percent_and_rejects_quote ... FAILED
  left: "100%%"
 right: "\"100%%\""
test integrations::agent_cli::launcher::tests::batch_script_always_quotes_args_even_without_whitespace ... FAILED
test integrations::agent_cli::launcher::tests::batch_script_drives_percent_and_quote_through_the_argument_path ... FAILED
test result: FAILED. 10 passed; 4 failed; 0 ignored; 0 measured; 2217 filtered out
```

Restored: `test result: ok. 14 passed; 0 failed; 0 ignored; 0 measured; 2217 filtered out`.

I also added `@echo off` as the first line, which has no unix analogue: without it, cmd.exe echoes every line of the script to stdout before running it, which would corrupt the ACP JSON-RPC stream the spawned agent is piped over (`live/sessions/driver/process.rs` pipes the launcher's stdout as the agent's protocol channel). And the script ends with an explicit `exit /b %ERRORLEVEL%` rather than relying on cmd.exe's implicit "last command's exit code" behavior for a `/c`-invoked script, per the review note to actually verify propagation rather than assume it.

## Process-tree interaction with #2147

`exec "<path>" args "$@"` on unix **replaces** the shell process image, the shell and the agent are the same PID, so `self.child.id()` (`live/sessions/actor/run.rs:130`, the pid `#2147`'s Windows Toolhelp walk roots its census on) already **is** the agent process; there is no surviving intermediate process.

Batch has no `exec`. `"<path>" args %*` **spawns** a child; the launcher's `cmd.exe` (or, per the std behavior above, the `cmd.exe` that `Command::new` itself starts to interpret the `.cmd`) stays alive as that child's parent. So on Windows the process tree is one level deeper than on unix: `anyharness -> cmd.exe (interpreting the .cmd) -> agent.exe`, versus `anyharness -> agent (in place)` on unix.

This does not break #2147's kill path, `CreateToolhelp32Snapshot` walks the **whole descendant subtree** of the root pid, not just direct children, so the extra `cmd.exe` layer is still enumerated and killed (deepest-first, per #2147's `TreeTracker`). But #2147's own "what remains unverified" section says its integration tests only exercised a 2-3 process tree; this launcher change means the real-world minimum tree on Windows is now one level deeper than what was tested there, and the persistent parent `cmd.exe` is a real, additional node that will show up in that walk on every Windows session. Flagging this as a direct, real interaction between the two PRs rather than a hypothetical.

## Windows caveat on FR-3 (disclosure, not a code change)

`generate_launcher_script_atomic`'s doc comment (and the `atomic_regen_preserves_previous_inode_until_promotion` test name) states an invariant, "FR-3: a running session keeps the previous inode until promotion," that is true on unix (a `rename(2)` over a file someone still has open just repoints the directory entry; the open file descriptor keeps working against the old inode) and **false on Windows**. cmd.exe holds an executing `.cmd` file open without `FILE_SHARE_DELETE`, so `std::fs::rename` onto a `.cmd` launcher that a live session's `cmd.exe` is still interpreting should fail with an access-denied error on Windows, not silently keep serving the old content the way the unix inode trick does.

This isn't a regression this PR introduces, `promote_launcher`'s rename already treats a failed rename as fail-safe (`Err` → restore the previous launcher, per the existing rollback path), so a promotion that races a live Windows session should fail closed rather than corrupt anything. But this PR is what makes that code path **live on Windows for the first time**, before this PR there was no Windows-executable launcher for cmd.exe to be holding open, so the platform divergence in what FR-3 actually guarantees needs to be disclosed here rather than discovered later. I have not verified this against a real Windows host in this session; it's reasoned from documented Windows file-locking semantics for a process interpreting an open script file, not observed.

## Stacking

This PR is **not stacked**, it targets `main` directly, independent of the `fix/agent-catalog-windows-arm64-pins` (#2145) → `fix/pinned-installer-windows-exe` (#2149) chain and independent of `ci/windows-compile-check` (#2142) → `fix/windows-process-tree-kill` (#2147). The two changes are complementary: #2149 makes the *thing the launcher execs* end in `.exe` on Windows; this PR makes the *launcher itself* executable. Both are needed for a Windows `claude`/`codex` session to actually start.

**Correction from an earlier revision of this body: it does *not* touch a disjoint file from #2149.** A reviewer ran an in-memory `git merge-tree --write-tree` between the two branches (no worktree, no build) and it produces two real content conflicts, which I reproduced myself:

- `installer/npm.rs`: this PR keeps `pub(super) fn platform_binary_filename` (`npm.rs:487`) and grows the new npm helpers directly beneath it. #2149 deletes that function from `npm.rs` entirely, having moved it to `anyharness/crates/anyharness-lib/src/integrations/agent_cli/executable.rs` as `pub(crate) fn platform_binary_filename` (and made it idempotent against an already-`.exe`-suffixed name).
- `installer/agent_process.rs` (line 4): this PR writes `use super::npm::{platform_binary_filename, platform_npm_bin_relpath};`; #2149 writes `use crate::integrations::agent_cli::executable::{is_valid_executable, platform_binary_filename};`.

Correct resolution: `platform_binary_filename` lives only in `executable.rs`; `npm.rs` keeps `platform_npm_bin_relpath` but drops its own copy of `platform_binary_filename`; `agent_process.rs` imports `platform_npm_bin_relpath` from `super::npm` and `platform_binary_filename` (plus `is_valid_executable`) from `executable.rs`. This exact combination already exists and is merged into `ci/windows-managed-agent-e2e` (PR #2153), where it was verified correct by that PR's own reviewer. #2153 is the reference for how to land this pairing, whichever of #2152/#2149 merges second.

## Two upstream blockers fixed in this PR (these matter more than the launcher itself)

A working `.cmd` launcher is necessary but not sufficient: two things upstream of it were still Windows-fatal, both are the exact same defect class as the launcher fix itself (assuming a unix-shaped name resolves on Windows), one level lower in the stack. Per review, both are fixed here rather than filed as follow-ups, and I checked what the surrounding code already does for the same problem (`platform_binary_filename`'s `cfg!(windows)`-gated `.exe` suffix) before adding anything new.

**1. `Command::new("npm")` can't find npm on Windows.** `installer/npm.rs` had two call sites (`pack_npm_package_dir`, `install_npm_package_into_prefix`) spawning `npm` directly. Node's Windows installer ships `npm.cmd` (and `npm.ps1`), never `npm.exe`. I verified against the actual Windows behavior rather than from memory: when a program name has no extension, Windows' own `CreateProcess` search (which is what `std::process::Command` ultimately calls) appends only a default `.exe`, not `.cmd`, so `Command::new("npm")` looks for `npm.exe`, finds nothing, and fails before npm ever runs. Separately, `std::process::Command` (and `tokio::process::Command`, which wraps it) has *documented* special-case handling for a program name that already ends in `.bat`/`.cmd`: it detects the extension and spawns via `cmd.exe /c` itself (this is also the surface CVE-2024-24576 patched, restricting `
`/`
`/NUL bytes through that path, real, load-bearing std behavior, not an assumption). So naming the program `npm.cmd` explicitly, the same mechanism this PR already relies on for the generated launcher itself, makes `Command::new` take that branch, and PATH resolution for `npm.cmd` then happens inside the `cmd.exe` it spawns, which does find it.

  Fix: added `npm_program_name()` (split into testable `npm_program_name_windows`/`npm_program_name_unix` halves under the same `#[cfg(any(windows/not(windows), test))]` pattern as the launcher's batch builders), returning `"npm.cmd"` on Windows and `"npm"` elsewhere, and changed both `Command::new("npm")` sites to `Command::new(npm_program_name())`.

**2. The launcher execs an extension-less `.bin` shim, relying on unguaranteed PATHEXT fallback.** For npm/git-installed agents, `executable_relpath` names npm's UNIX shim under `node_modules/.bin/<name>`, a file with no extension. npm's own `cmd-shim` mechanism additionally writes a `<name>.cmd` (and `<name>.ps1`) sibling into that same `.bin` directory specifically so Windows can run it; the bare shim only "worked" today, if it worked at all, by hoping cmd.exe's `PATHEXT` fallback resolves the extension-less name to a sibling `.cmd` at the shell level, not something `CreateProcess` itself guarantees at the layer where this launcher actually calls exec. Resolving to the real `.cmd` sibling explicitly, instead of relying on that fallback, is the same choice `platform_binary_filename` already makes for a source-built binary (`<name>` → `<name>.exe`), applied to the npm-shim case.

  Fix: added `platform_npm_bin_relpath()` (same testable split pattern) which appends `.cmd` to `executable_relpath` on Windows and passes it through unchanged elsewhere; `install_managed_npm_package`'s `exec_relpath` computation now calls it instead of using `executable_relpath` directly for the non-source-build branch. This value is what ultimately becomes `active_exec`/`final_exec`/the launcher's actual exec target in `stage_and_swap_managed_npm_tree`, and it's also threaded into `agent_process.rs`'s `regenerate_agent_process_launcher` exec-candidate list (`platform_npm_bin_relpath(executable_relpath)` ahead of the bare `executable_relpath` candidate) so the seed-reconciliation path resolves the same `.cmd` sibling.

**What I resolve and why:** on Windows, `npm` → `npm.cmd` (npm ships no `.exe`) and `node_modules/.bin/<name>` → `node_modules/.bin/<name>.cmd` (npm's own cmd-shim output, not a guess), both explicit resolutions replace an implicit "Windows will figure it out" assumption with the actual sibling name Windows tooling writes, at the exact two points (npm invocation, launcher exec target) upstream of the launcher-format fix that would otherwise have failed before a `.cmd` launcher was even reachable.

**Tests for both, plus a real negative control.** Added `windows_npm_resolution_tests` in `npm.rs` (4 tests, using the same direct-call-the-`_windows`/`_unix`-half pattern as the launcher tests): `npm_program_name_is_dot_cmd_on_windows`, `npm_program_name_stays_bare_on_unix`, `platform_npm_bin_relpath_appends_cmd_extension_on_windows`, `platform_npm_bin_relpath_leaves_unix_shim_unchanged`.

```
running 4 tests
test domains::agents::installer::npm::windows_npm_resolution_tests::npm_program_name_is_dot_cmd_on_windows ... ok
test domains::agents::installer::npm::windows_npm_resolution_tests::npm_program_name_stays_bare_on_unix ... ok
test domains::agents::installer::npm::windows_npm_resolution_tests::platform_npm_bin_relpath_leaves_unix_shim_unchanged ... ok
test domains::agents::installer::npm::windows_npm_resolution_tests::platform_npm_bin_relpath_appends_cmd_extension_on_windows ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 2231 filtered out
```

Negative control (reverted `npm_program_name`/`platform_npm_bin_relpath` to their original single-body `if cfg!(windows) {...} else {...}` form, which cannot be called with an explicit platform selection): fails to **compile**, not just at runtime, for the same reason the original launcher negative control did, the test functions call `npm_program_name_windows`/`npm_program_name_unix`/`platform_npm_bin_relpath_windows`/`platform_npm_bin_relpath_unix`, which don't exist pre-fix:

```
error[E0425]: cannot find function `npm_program_name_windows` in this scope
error[E0425]: cannot find function `npm_program_name_unix` in this scope
error[E0425]: cannot find function `platform_npm_bin_relpath_windows` in this scope
error[E0425]: cannot find function `platform_npm_bin_relpath_unix` in this scope
error: could not compile `anyharness-lib` (lib test) due to 4 previous errors
```

Restored: `test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 2231 filtered out`.

## What I tested, and where it ran

- **Pure string-shape unit tests** (14 in `launcher.rs` after the round-2 coverage-hole fix, 4 more in `npm.rs`'s `windows_npm_resolution_tests`) call the batch-script builders, escape helpers, filename resolvers, and the new `npm_program_name_*`/`platform_npm_bin_relpath_*` split functions **directly**, never through the `#[cfg(windows)]` runtime dispatch, using the `#[cfg(any(windows/not(windows), test))]` pattern throughout (established for the launcher, and reused as-is for the npm helpers). They assert against generated strings/paths only, no execution, no FFI, so they all run on the existing Linux CI unit-test job today.
  - `cargo test -p anyharness-lib --lib integrations::agent_cli::launcher::` → **14 passed, 0 failed**.
  - `cargo test -p anyharness-lib --lib domains::agents::installer::npm::` → **4 passed, 0 failed**.
  - Full `domains::agents::` module (395 tests) to check for regressions across every read/write site I touched: **394 passed, 1 failed**, the 1 failure is the same pre-existing, unrelated flake as before (`route_auth::probe_materialization_tests::sweep_tests::the_sweep_removes_only_abandoned_and_old_scratch_roots`, an mtime-based sweep test), reconfirmed pre-existing by running it against unmodified `origin/main` with my changes stashed: it fails there too.
  - `cargo check -p anyharness-lib --lib`: clean, no new warnings.
- **Negative control #1, the original launcher fix** (mandatory, quoted literally): reverted `launcher.rs` to `origin/main` HEAD and re-appended only the new test functions on top of the pre-fix code. Fails to **compile**, not just at runtime:

  ```
  error[E0425]: cannot find function `build_batch_launcher_script` in this scope
  error[E0425]: cannot find function `batch_escape_arg` in this scope
  error[E0599]: no variant or associated item named `UnsupportedBatchValue` found for enum `launcher::LauncherError`
  error[E0425]: cannot find function `windows_launcher_file_name` in this scope
  ... (20 errors total)
  error: could not compile `anyharness-lib` (lib test) due to 20 previous errors
  ```

  Restored: back to **14 passed, 0 failed** (14, not 12, see negative control #2 below for the round-2 coverage-hole fix that added 2 of those 14).

- **Negative control #2, the round-2 `batch_escape_arg` coverage-hole fix** (reviewer's 12-mutation run found this one; already shown above in "Batch-escaping decision," repeated here as part of the full test record): reverting `batch_escape_arg` to its old whitespace-conditional form produces **4 failing tests**, none of them a compile error this time (they're runtime assertion failures, which is exactly the gap the mutation found, the old code compiled fine, it just produced the wrong string for an argument):

  ```
  test integrations::agent_cli::launcher::tests::batch_escape_arg_always_quotes_even_a_single_word_token ... FAILED
    left: "--acp"
   right: "\"--acp\""
  test integrations::agent_cli::launcher::tests::batch_escape_arg_still_doubles_percent_and_rejects_quote ... FAILED
    left: "100%%"
   right: "\"100%%\""
  test integrations::agent_cli::launcher::tests::batch_script_always_quotes_args_even_without_whitespace ... FAILED
  test integrations::agent_cli::launcher::tests::batch_script_drives_percent_and_quote_through_the_argument_path ... FAILED
  test result: FAILED. 10 passed; 4 failed; 0 ignored; 0 measured; 2217 filtered out
  ```

  Restored: `test result: ok. 14 passed; 0 failed; 0 ignored; 0 measured; 2217 filtered out`.

- **Negative control #3, the two upstream npm/exec-target fixes.** An earlier version of this control just deleted the `_windows`/`_unix` split and watched the tests fail to *compile*, that only proves the tests call the new functions, not that they'd catch a wrong implementation, and review correctly called this out as the weak form (a repeat of the same note on round 1). The real control keeps every function present and signature-compatible and makes the **behavior** wrong instead: `npm_program_name_windows` returns `"npm"` (the old, broken value) instead of `"npm.cmd"`, and `platform_npm_bin_relpath_windows` returns the bare shim path with no `.cmd` appended. Both mutations are real, not hypothetical, they're exactly what `npm_program_name`/`platform_npm_bin_relpath` used to compute before this PR, reintroduced one function at a time while every other line (including the tests) stayed untouched:

  ```
  ---- domains::agents::installer::npm::windows_npm_resolution_tests::npm_program_name_is_dot_cmd_on_windows stdout ----
  thread '...' panicked at .../npm.rs:612:9:
    left: "npm"
   right: "npm.cmd"

  ---- domains::agents::installer::npm::windows_npm_resolution_tests::platform_npm_bin_relpath_appends_cmd_extension_on_windows stdout ----
  thread '...' panicked at .../npm.rs:623:9:
    left: "node_modules/.bin/claude-agent-acp"
   right: "node_modules/.bin/claude-agent-acp.cmd"

  test result: FAILED. 2 passed; 2 failed; 0 ignored; 0 measured; 2231 filtered out
  ```

  Restored both functions to their correct bodies (bit-for-bit identical to the pushed diff, confirmed with a byte-level `diff` before recommitting): `test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 2231 filtered out`.

## What remains unverified, honestly

**Correction from an earlier revision of this body: real Windows execution of this launcher has now happened, and the stated obstacle to getting there was wrong.** The earlier text said no Windows execution was possible because `generate_launcher_script`, `build_batch_launcher_script`, `batch_escape`, and `managed_launcher_file_name` are all `pub(crate)`, so an integration test crate (which only sees the public API) couldn't reach them. That was never the real obstacle: `install_agent_with_pins` (`installer/service.rs`) is `pub`, and a proof test in `anyharness/crates/anyharness-lib/tests/windows_managed_agent_install.rs` drives the whole install-and-spawn flow through that public function, never needing the private helpers directly. Two CI runs against this PR's own head (`f3122f08fc`) settle it:

**Positive: run [32446919674](https://github.com/proliferate-ai/proliferate/actions/runs/32446919674)**, `windows-latest`, `success`, `test result: ok. 3 passed; 0 failed`. Literal lines:

```
PATHEXT = ".COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.CPL"
control: bare `npm --version` -> NOT SPAWNABLE: program not found (kind NotFound, raw os error None)
installer program name = "npm.cmd"
npm.cmd --version -> status ExitStatus(ExitStatus(0)) stdout "10.9.8" stderr ""
```

npm's cmd-shim shape in a real `.bin` on that runner:

```
file 348  claude-agent-acp.cmd
file 435  claude-agent-acp
file 897  claude-agent-acp.ps1
```

The generated launcher and its spawn (same for `grok-launcher.cmd`, pid 1856):

```
file 355  claude-launcher.cmd
@echo off
set "PATH=...\agents\claude\native;%PATH%"
set "DISABLE_AUTOUPDATER=1"
"...\agents\claude\agent_process\node_modules/.bin/claude-agent-acp.cmd" %*
exit /b %ERRORLEVEL%
spawned pid 3652
```

**Negative control: run [32445803852](https://github.com/proliferate-ai/proliferate/actions/runs/32445803852)**, `windows-latest`, `failure`, at a head carrying the launcher commit but not the npm/exec-target fix commit. Literal failure:

```
`Command::new("npm")` could not be spawned: program not found (kind NotFound, raw os error None). This is installer/npm.rs's exact call shape.
install_agent_with_pins failed: io error: program not found (kind Other)
test result: FAILED. 0 passed; 2 failed
```

That settles, by real execution, the three things the earlier body listed as open: `Command::new("npm.cmd")` resolves and spawns where bare `npm` does not; npm's cmd-shim writes exactly the `<name>` / `<name>.cmd` / `<name>.ps1` shape this code assumes; and an install-flow test now exercises both upstream fixes end-to-end.

What genuinely remains open after those two runs:

- **The proof is "spawns and exits 0 on an empty stdin," not an ACP handshake.** Both `claude-launcher.cmd` and `grok-launcher.cmd` in run 32446919674 produced empty stdout *and* empty stderr before exiting 0 (`--- launcher stdout` / (blank) / `--- launcher stderr` / (blank) / `ExitStatus(0)`). That is consistent with each shim starting node, seeing closed stdin, and exiting cleanly; it is not evidence that a live ACP `initialize` request/response round-trips through the launcher. No test in this PR or #2153 sends one.
- **Coverage is 2 of 5 agents; a third has an entirely different code path this test never touches.** Only `claude` (pinned-binary native + npm agent-process) and `grok` (npm-only) were exercised. `cursor` and `opencode` have no `native` install block in `catalogs/agents/registry.json` at all, i.e. no Windows binary-pin targets exist for them today, so their coverage gap is a catalog gap, not just a test gap. `codex`'s `native.install.kind` is `tarball_release` (a `.tar.gz` extraction branch, distinct from claude's `direct_binary` branch) with `windows_x64`/`windows_arm64` entries in its `platformMap`, a real, catalog-declared Windows path that neither of these two CI runs exercises.
- The `"` -> reject-with-error behavior in `batch_escape` still has no test exercising it end-to-end through the real installer path.
- Whether `cmd.exe`'s implicit exit-code-of-last-command behavior would have worked without the explicit `exit /b %ERRORLEVEL%` is still unverified (moot in practice: the explicit line is what shipped and it's what these runs observed).
- The one-level-deeper process tree's practical effect on #2147's kill path (described above) is reasoned from reading both implementations, not observed on Windows.
- The Windows FR-3 file-locking divergence described above (cmd.exe holding the `.cmd` open without `FILE_SHARE_DELETE`) is reasoned from documented Windows semantics, not observed against a real running session on Windows.
- Non-zero exit-code propagation through the launcher is unobserved on either branch; both runs only exercised the exit-0 path.

## Two things the reviewer independently re-verified as correct (recording, not re-litigating)

- **The launcher-naming rename is complete.** Independently re-grepped: the two staging names (`.{name}-launcher.next`) are rename-*destination*-only and never executed directly; `ResolvedArtifact` carries no serde attribute that would leak a stale name; `install-manifest.json` persists a `path` field that no reader currently consumes.
- **`%ERRORLEVEL%` timing is correct.** Confirmed by rendering the actual emitted script: the `exit /b %ERRORLEVEL%` line is standalone at the top level of the script, not nested inside a block (an `if`/`for`/parenthesized group) where cmd.exe would have already substituted `%ERRORLEVEL%` at parse time rather than at execution time, so it reads the real exit code of the immediately preceding command, not a parse-time snapshot.

## CI status: repo-shape red, fixed in `24270510c8`

`npm.rs` reached 636 lines at `24d6299510` against the 600-line `PROD-SIZE-1` cap (636 on the branch, 526 on `main`, no `[[max_lines]]` entry in `lints/anyharness/ratchets.toml`). The `+110` is this PR's own `npm_program_name*` / `platform_npm_bin_relpath*` block and its tests, not a rebase artifact.

Fixed by shrinking the file, not by ratcheting it. `lints/anyharness/ratchets.toml` says in its own header that it is "a ratchet, not an exception ledger" and that entries "may only shrink", so adding a new entry for growth this PR introduced would be using the ledger against its stated purpose.

`24270510c8` is a pure move: `npm_program_name{,_windows,_unix}`, `platform_npm_bin_relpath{,_windows,_unix}` and their four tests go to a new `installer/npm_platform.rs`. `npm.rs` re-exports `platform_npm_bin_relpath` as `pub(super)`, so `agent_process.rs`'s existing `use super::npm::{...}` import resolves unchanged, which also keeps PR #2153's already-verified merge resolution valid.

`npm.rs` 636 to 535. `python3 scripts/check_max_lines.py` now prints `Max-lines check passed (repo max 600, component max 500, server layer maxes enabled)`.

Negative control on the moved tests, so the move is proven to have carried the assertions rather than just the code. Mutating `npm_program_name_windows()` to return `"npm"` and `platform_npm_bin_relpath_windows` to return the path unchanged:

```
test ...::npm_program_name_is_dot_cmd_on_windows ... FAILED
test ...::platform_npm_bin_relpath_appends_cmd_extension_on_windows ... FAILED
test result: FAILED. 2 passed; 2 failed
```

Restored:

```
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 2240 filtered out
```

## Files touched

- `anyharness/crates/anyharness-lib/src/integrations/agent_cli/launcher.rs`
- `anyharness/crates/anyharness-lib/src/domains/agents/installer/pinned.rs`
- `anyharness/crates/anyharness-lib/src/domains/agents/installer/agent_process.rs`
- `anyharness/crates/anyharness-lib/src/domains/agents/installer/service.rs`
- `anyharness/crates/anyharness-lib/src/domains/agents/installer/npm.rs`
- `anyharness/crates/anyharness-lib/src/domains/agents/installer/npm_platform.rs` (new)
- `anyharness/crates/anyharness-lib/src/domains/agents/installer/mod.rs`
- `anyharness/crates/anyharness-lib/src/domains/agents/readiness/artifacts.rs`
- `anyharness/crates/anyharness-lib/src/domains/agents/readiness/compatibility.rs`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4241,7 +4241,7 @@ dependencies = [
 
 [[package]]
 name = "proliferate"
-version = "0.4.20"
+version = "0.4.21"
 dependencies = [
  "base64 0.22.1",
  "chrono",
--- a/anyharness/crates/anyharness-lib/src/domains/agents/installer/agent_process.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/installer/agent_process.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 
+use super::npm::platform_npm_bin_relpath;
 use super::{InstallError, InstalledArtifactResult};
 use crate::domains::agents::installer::seed;
 use crate::domains::agents::model::*;
@@ -66,6 +67,11 @@ fn regenerate_agent_process_launcher(
     if let Some(binary_name) = source_build_binary_name {
         exec_candidates.push(managed_dir.join(platform_binary_filename(binary_name)));
     }
+    // The windows `.cmd` shim sibling first (see `platform_npm_bin_relpath`):
+    // npm's cmd-shim writes both the bare unix shim and the `.cmd` shim into
+    // `.bin`, so on windows both candidates below exist side by side and
+    // this ordering makes the executable one win.
+    exec_candidates.push(managed_dir.join(platform_npm_bin_relpath(executable_relpath)));
     exec_candidates.push(managed_dir.join(executable_relpath));
     let Some(exec_path) = exec_candidates.iter().find(|path| path.exists()).cloned() else {
         return Err(InstallError::MissingManagedArtifact(
--- a/anyharness/crates/anyharness-lib/src/domains/agents/installer/agent_process.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/installer/agent_process.rs
@@ -7,7 +7,9 @@ use crate::domains::agents::model::*;
 use crate::domains::agents::readiness::paths::{artifact_root, managed_pinned_binary_path};
 use crate::domains::agents::registry::built_in_registry;
 use crate::integrations::agent_cli::executable::{is_valid_executable, platform_binary_filename};
-use crate::integrations::agent_cli::launcher::generate_launcher_script_atomic;
+use crate::integrations::agent_cli::launcher::{
+    generate_launcher_script_atomic, managed_launcher_file_name,
+};
 
 pub(super) fn regenerate_seeded_agent_launchers(
     runtime_home: &Path,
@@ -35,7 +37,7 @@ fn regenerate_agent_process_launcher(
 ) -> Result<Option<InstalledArtifactResult>, InstallError> {
     let kind = &descriptor.kind;
     let managed_dir = artifact_root(runtime_home, kind, &ArtifactRole::AgentProcess);
-    let launcher_path = managed_dir.join(format!("{}-launcher", kind.as_str()));
+    let launcher_path = managed_dir.join(managed_launcher_file_name(kind.as_str()));
     let path_prefixes = launcher_path_prefixes(runtime_home, kind);
     let env = managed_launcher_env(kind);
 
--- a/anyharness/crates/anyharness-lib/src/domains/agents/installer/mod.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/installer/mod.rs
@@ -6,6 +6,7 @@ mod lock;
 pub(crate) mod managed_npm;
 pub mod manifest;
 mod npm;
+mod npm_platform;
 pub mod off_runtime;
 mod pinned;
 pub mod progress;
--- a/anyharness/crates/anyharness-lib/src/domains/agents/installer/npm.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/installer/npm.rs
@@ -1,11 +1,13 @@
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 
+use super::downloads::activate_local_tree;
 use super::managed_npm::{
     apply_npm_version_override, installed_npm_package_version, managed_npm_install_issue,
     npm_package_version, source_build_install_issue, write_managed_npm_source_marker,
 };
-use super::downloads::activate_local_tree;
+use super::npm_platform::npm_program_name;
+pub(super) use super::npm_platform::platform_npm_bin_relpath;
 use super::{InstallError, InstalledArtifactResult};
 use crate::domains::agents::model::ArtifactRole;
 use crate::integrations::agent_cli::executable::{make_executable, platform_binary_filename};
@@ -150,7 +152,11 @@ fn stage_and_swap_managed_npm_tree(
         if let Some(binary_name) = source_build_binary_name {
             install_managed_source_build_binary(versioned_package, &staging_dir, binary_name)?;
         } else if let Some(package_subdir) = package_subdir {
-            install_managed_npm_package_from_subdir(versioned_package, package_subdir, &staging_dir)?;
+            install_managed_npm_package_from_subdir(
+                versioned_package,
+                package_subdir,
+                &staging_dir,
+            )?;
         } else {
             install_npm_package_into_prefix(versioned_package, &staging_dir)?;
         }
@@ -484,72 +490,6 @@ fn install_npm_package_into_prefix(package: &str, managed_dir: &Path) -> Result<
     .map(|_| ())
 }
 
-/// npm itself is a JS program, not a native binary: node's Windows installer
-/// ships `npm.cmd` (and `npm.ps1`), never `npm.exe`. `Command::new("npm")`'s
-/// own Windows module search only appends a default `.exe` to an
-/// extension-less program name (that's `CreateProcess`'s behavior when
-/// `lpApplicationName` is unset, not something Rust adds), so it would look
-/// for `npm.exe`, find nothing, and fail before ever reaching npm. Naming
-/// the program `npm.cmd` explicitly makes `std::process::Command` take its
-/// documented `.cmd`/`.bat` branch and spawn it via `cmd.exe /c` itself —
-/// the same mechanism this crate now relies on for the generated launcher
-/// (see `integrations::agent_cli::launcher::managed_launcher_file_name`) —
-/// so PATH resolution for `npm.cmd` happens inside cmd.exe, which does find
-/// it.
-#[cfg(any(windows, test))]
-fn npm_program_name_windows() -> &'static str {
-    "npm.cmd"
-}
-
-#[cfg(any(not(windows), test))]
-fn npm_program_name_unix() -> &'static str {
-    "npm"
-}
-
-fn npm_program_name() -> &'static str {
-    #[cfg(windows)]
-    {
-        npm_program_name_windows()
-    }
-    #[cfg(not(windows))]
-    {
-        npm_program_name_unix()
-    }
-}
-
-/// Resolve the executable a managed npm/git install should actually exec.
-/// `executable_relpath` (e.g. `node_modules/.bin/claude-agent-acp`) names
-/// npm's UNIX shim, which has no windows equivalent — npm's `cmd-shim`
-/// ALSO writes a `<name>.cmd` (and `<name>.ps1`) sibling into the same
-/// `.bin` directory, and `<name>.cmd` is the one Windows can actually run.
-/// Execing the bare, extension-less name and hoping PATHEXT / `CreateProcess`
-/// falls back to a sibling is not guaranteed at the `CreateProcess` layer the
-/// way it's easy to assume; naming the real `.cmd` sibling explicitly here
-/// is the same choice `platform_binary_filename` above already makes for a
-/// source-built binary, applied to the npm-shim case.
-#[cfg(any(windows, test))]
-fn platform_npm_bin_relpath_windows(executable_relpath: &Path) -> PathBuf {
-    let mut with_ext = executable_relpath.as_os_str().to_os_string();
-    with_ext.push(".cmd");
-    PathBuf::from(with_ext)
-}
-
-#[cfg(any(not(windows), test))]
-fn platform_npm_bin_relpath_unix(executable_relpath: &Path) -> PathBuf {
-    executable_relpath.to_path_buf()
-}
-
-pub(super) fn platform_npm_bin_relpath(executable_relpath: &Path) -> PathBuf {
-    #[cfg(windows)]
-    {
-        platform_npm_bin_relpath_windows(executable_relpath)
-    }
-    #[cfg(not(windows))]
-    {
-        platform_npm_bin_relpath_unix(executable_relpath)
-    }
-}
-
 pub(super) fn run_command_capture(
     program: &str,
     command: &mut Command,
@@ -584,45 +524,4 @@ fn read_dir_entry_names(dir: &Path) -> Vec<String> {
         .collect();
     entries.sort();
     entries
-}
-
-#[cfg(test)]
-mod windows_npm_resolution_tests {
-    use super::*;
-
-    // These call the `_windows`/`_unix` split functions directly (not the
-    // `cfg!(windows)`-dispatching wrappers), so both branches are exercised
-    // on every host regardless of which platform actually built the test
-    // binary — the same pure-string-shape pattern used in
-    // `integrations::agent_cli::launcher`'s batch-script tests.
-
-    #[test]
-    fn npm_program_name_is_dot_cmd_on_windows() {
-        // `Command::new`'s Windows module search appends only a default
-        // `.exe` to an extension-less program name; naming it `npm.cmd`
-        // explicitly is what makes std take its documented `.cmd`/`.bat`
-        // branch and spawn via `cmd.exe /c` itself, the same mechanism the
-        // generated launcher relies on.
-        assert_eq!(npm_program_name_windows(), "npm.cmd");
-    }
-
-    #[test]
-    fn npm_program_name_stays_bare_on_unix() {
-        assert_eq!(npm_program_name_unix(), "npm");
-    }
-
-    #[test]
-    fn platform_npm_bin_relpath_appends_cmd_extension_on_windows() {
-        let shim = Path::new("node_modules/.bin/claude-agent-acp");
-        assert_eq!(
-            platform_npm_bin_relpath_windows(shim),
-            PathBuf::from("node_modules/.bin/claude-agent-acp.cmd")
-        );
-    }
-
-    #[test]
-    fn platform_npm_bin_relpath_leaves_unix_shim_unchanged() {
-        let shim = Path::new("node_modules/.bin/claude-agent-acp");
-        assert_eq!(platform_npm_bin_relpath_unix(shim), shim.to_path_buf());
-    }
 }
--- a/anyharness/crates/anyharness-lib/src/domains/agents/installer/npm.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/installer/npm.rs
@@ -36,7 +36,10 @@ pub(super) fn install_managed_npm_package(
     let exec_relpath: PathBuf = if let Some(binary_name) = source_build_binary_name {
         platform_binary_filename(binary_name)
     } else {
-        executable_relpath.to_path_buf()
+        // `executable_relpath` names npm's unix `.bin` shim; on windows the
+        // launcher must exec the `.cmd` sibling npm's cmd-shim also writes
+        // (see `platform_npm_bin_relpath`), not the bare unix shim.
+        platform_npm_bin_relpath(executable_relpath)
     };
     let active_exec = managed_dir.join(&exec_relpath);
 
@@ -385,7 +388,7 @@ fn resolve_npm_package_subdir(
 fn pack_npm_package_dir(package_dir: &Path, staging_root: &Path) -> Result<PathBuf, InstallError> {
     let output = run_command_capture(
         "npm",
-        Command::new("npm")
+        Command::new(npm_program_name())
             .arg("pack")
             .arg("--pack-destination")
             .arg(staging_root)
@@ -471,7 +474,7 @@ fn build_cargo_binary_from_source(
 fn install_npm_package_into_prefix(package: &str, managed_dir: &Path) -> Result<(), InstallError> {
     run_command_capture(
         "npm",
-        Command::new("npm")
+        Command::new(npm_program_name())
             .args(["install", "--no-audit", "--no-fund", "--prefix"])
             .arg(managed_dir)
             .arg(package)
@@ -479,6 +482,72 @@ fn install_npm_package_into_prefix(package: &str, managed_dir: &Path) -> Result<
             .stderr(Stdio::piped()),
     )
     .map(|_| ())
+}
+
+/// npm itself is a JS program, not a native binary: node's Windows installer
+/// ships `npm.cmd` (and `npm.ps1`), never `npm.exe`. `Command::new("npm")`'s
+/// own Windows module search only appends a default `.exe` to an
+/// extension-less program name (that's `CreateProcess`'s behavior when
+/// `lpApplicationName` is unset, not something Rust adds), so it would look
+/// for `npm.exe`, find nothing, and fail before ever reaching npm. Naming
+/// the program `npm.cmd` explicitly makes `std::process::Command` take its
+/// documented `.cmd`/`.bat` branch and spawn it via `cmd.exe /c` itself —
+/// the same mechanism this crate now relies on for the generated launcher
+/// (see `integrations::agent_cli::launcher::managed_launcher_file_name`) —
+/// so PATH resolution for `npm.cmd` happens inside cmd.exe, which does find
+/// it.
+#[cfg(any(windows, test))]
+fn npm_program_name_windows() -> &'static str {
+    "npm.cmd"
+}
+
+#[cfg(any(not(windows), test))]
+fn npm_program_name_unix() -> &'static str {
+    "npm"
+}
+
+fn npm_program_name() -> &'static str {
+    #[cfg(windows)]
+    {
+        npm_program_name_windows()
+    }
+    #[cfg(not(windows))]
+    {
+        npm_program_name_unix()
+    }
+}
+
+/// Resolve the executable a managed npm/git install should actually exec.
+/// `executable_relpath` (e.g. `node_modules/.bin/claude-agent-acp`) names
+/// npm's UNIX shim, which has no windows equivalent — npm's `cmd-shim`
+/// ALSO writes a `<name>.cmd` (and `<name>.ps1`) sibling into the same
+/// `.bin` directory, and `<name>.cmd` is the one Windows can actually run.
+/// Execing the bare, extension-less name and hoping PATHEXT / `CreateProcess`
+/// falls back to a sibling is not guaranteed at the `CreateProcess` layer the
+/// way it's easy to assume; naming the real `.cmd` sibling explicitly here
+/// is the same choice `platform_binary_filename` above already makes for a
+/// source-built binary, applied to the npm-shim case.
+#[cfg(any(windows, test))]
+fn platform_npm_bin_relpath_windows(executable_relpath: &Path) -> PathBuf {
+    let mut with_ext = executable_relpath.as_os_str().to_os_string();
+    with_ext.push(".cmd");
+    PathBuf::from(with_ext)
+}
+
+#[cfg(any(not(windows), test))]
+fn platform_npm_bin_relpath_unix(executable_relpath: &Path) -> PathBuf {
+    executable_relpath.to_path_buf()
+}
+
+pub(super) fn platform_npm_bin_relpath(executable_relpath: &Path) -> PathBuf {
+    #[cfg(windows)]
+    {
+        platform_npm_bin_relpath_windows(executable_relpath)
+    }
+    #[cfg(not(windows))]
+    {
+        platform_npm_bin_relpath_unix(executable_relpath)
+    }
 }
 
 pub(super) fn run_command_capture(
@@ -515,4 +584,45 @@ fn read_dir_entry_names(dir: &Path) -> Vec<String> {
         .collect();
     entries.sort();
     entries
+}
+
+#[cfg(test)]
+mod windows_npm_resolution_tests {
+    use super::*;
+
+    // These call the `_windows`/`_unix` split functions directly (not the
+    // `cfg!(windows)`-dispatching wrappers), so both branches are exercised
+    // on every host regardless of which platform actually built the test
+    // binary — the same pure-string-shape pattern used in
+    // `integrations::agent_cli::launcher`'s batch-script tests.
+
+    #[test]
+    fn npm_program_name_is_dot_cmd_on_windows() {
+        // `Command::new`'s Windows module search appends only a default
+        // `.exe` to an extension-less program name; naming it `npm.cmd`
+        // explicitly is what makes std take its documented `.cmd`/`.bat`
+        // branch and spawn via `cmd.exe /c` itself, the same mechanism the
+        // generated launcher relies on.
+        assert_eq!(npm_program_name_windows(), "npm.cmd");
+    }
+
+    #[test]
+    fn npm_program_name_stays_bare_on_unix() {
+        assert_eq!(npm_program_name_unix(), "npm");
+    }
+
+    #[test]
+    fn platform_npm_bin_relpath_appends_cmd_extension_on_windows() {
+        let shim = Path::new("node_modules/.bin/claude-agent-acp");
+        assert_eq!(
+            platform_npm_bin_relpath_windows(shim),
+            PathBuf::from("node_modules/.bin/claude-agent-acp.cmd")
+        );
+    }
+
+    #[test]
+    fn platform_npm_bin_relpath_leaves_unix_shim_unchanged() {
+        let shim = Path::new("node_modules/.bin/claude-agent-acp");
+        assert_eq!(platform_npm_bin_relpath_unix(shim), shim.to_path_buf());
+    }
 }
--- a/anyharness/crates/anyharness-lib/src/domains/agents/installer/npm_platform.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/installer/npm_platform.rs
@@ -1,0 +1,114 @@
+//! Platform-conditional npm program and shim-path resolution.
+//!
+//! Extracted from `npm.rs` so the Windows resolution rules live in one small
+//! file with their own tests, and so `npm.rs` stays under the PROD-SIZE-1
+//! line allowance rather than acquiring new measured debt.
+
+use std::path::{Path, PathBuf};
+
+/// npm itself is a JS program, not a native binary: node's Windows installer
+/// ships `npm.cmd` (and `npm.ps1`), never `npm.exe`. `Command::new("npm")`'s
+/// own Windows module search only appends a default `.exe` to an
+/// extension-less program name (that's `CreateProcess`'s behavior when
+/// `lpApplicationName` is unset, not something Rust adds), so it would look
+/// for `npm.exe`, find nothing, and fail before ever reaching npm. Naming
+/// the program `npm.cmd` explicitly makes `std::process::Command` take its
+/// documented `.cmd`/`.bat` branch and spawn it via `cmd.exe /c` itself —
+/// the same mechanism this crate now relies on for the generated launcher
+/// (see `integrations::agent_cli::launcher::managed_launcher_file_name`) —
+/// so PATH resolution for `npm.cmd` happens inside cmd.exe, which does find
+/// it.
+#[cfg(any(windows, test))]
+fn npm_program_name_windows() -> &'static str {
+    "npm.cmd"
+}
+
+#[cfg(any(not(windows), test))]
+fn npm_program_name_unix() -> &'static str {
+    "npm"
+}
+
+pub(super) fn npm_program_name() -> &'static str {
+    #[cfg(windows)]
+    {
+        npm_program_name_windows()
+    }
+    #[cfg(not(windows))]
+    {
+        npm_program_name_unix()
+    }
+}
+
+/// Resolve the executable a managed npm/git install should actually exec.
+/// `executable_relpath` (e.g. `node_modules/.bin/claude-agent-acp`) names
+/// npm's UNIX shim, which has no windows equivalent — npm's `cmd-shim`
+/// ALSO writes a `<name>.cmd` (and `<name>.ps1`) sibling into the same
+/// `.bin` directory, and `<name>.cmd` is the one Windows can actually run.
+/// Execing the bare, extension-less name and hoping PATHEXT / `CreateProcess`
+/// falls back to a sibling is not guaranteed at the `CreateProcess` layer the
+/// way it's easy to assume; naming the real `.cmd` sibling explicitly here
+/// is the same choice `platform_binary_filename` above already makes for a
+/// source-built binary, applied to the npm-shim case.
+#[cfg(any(windows, test))]
+fn platform_npm_bin_relpath_windows(executable_relpath: &Path) -> PathBuf {
+    let mut with_ext = executable_relpath.as_os_str().to_os_string();
+    with_ext.push(".cmd");
+    PathBuf::from(with_ext)
+}
+
+#[cfg(any(not(windows), test))]
+fn platform_npm_bin_relpath_unix(executable_relpath: &Path) -> PathBuf {
+    executable_relpath.to_path_buf()
+}
+
+pub(super) fn platform_npm_bin_relpath(executable_relpath: &Path) -> PathBuf {
+    #[cfg(windows)]
+    {
+        platform_npm_bin_relpath_windows(executable_relpath)
+    }
+    #[cfg(not(windows))]
+    {
+        platform_npm_bin_relpath_unix(executable_relpath)
+    }
+}
+
+#[cfg(test)]
+mod windows_npm_resolution_tests {
+    use super::*;
+
+    // These call the `_windows`/`_unix` split functions directly (not the
+    // `cfg!(windows)`-dispatching wrappers), so both branches are exercised
+    // on every host regardless of which platform actually built the test
+    // binary — the same pure-string-shape pattern used in
+    // `integrations::agent_cli::launcher`'s batch-script tests.
+
+    #[test]
+    fn npm_program_name_is_dot_cmd_on_windows() {
+        // `Command::new`'s Windows module search appends only a default
+        // `.exe` to an extension-less program name; naming it `npm.cmd`
+        // explicitly is what makes std take its documented `.cmd`/`.bat`
+        // branch and spawn via `cmd.exe /c` itself, the same mechanism the
+        // generated launcher relies on.
+        assert_eq!(npm_program_name_windows(), "npm.cmd");
+    }
+
+    #[test]
+    fn npm_program_name_stays_bare_on_unix() {
+        assert_eq!(npm_program_name_unix(), "npm");
+    }
+
+    #[test]
+    fn platform_npm_bin_relpath_appends_cmd_extension_on_windows() {
+        let shim = Path::new("node_modules/.bin/claude-agent-acp");
+        assert_eq!(
+            platform_npm_bin_relpath_windows(shim),
+            PathBuf::from("node_modules/.bin/claude-agent-acp.cmd")
+        );
+    }
+
+    #[test]
+    fn platform_npm_bin_relpath_leaves_unix_shim_unchanged() {
+        let shim = Path::new("node_modules/.bin/claude-agent-acp");
+        assert_eq!(platform_npm_bin_relpath_unix(shim), shim.to_path_buf());
+    }
+}
--- a/anyharness/crates/anyharness-lib/src/domains/agents/installer/pinned.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/installer/pinned.rs
@@ -21,7 +21,9 @@ use super::{InstallError, InstalledArtifactResult};
 use crate::domains::agents::model::{AgentKind, ArtifactRole, Platform};
 use crate::domains::agents::readiness::paths::{artifact_root, managed_pinned_binary_path};
 use crate::integrations::agent_cli::executable::{is_valid_executable, make_executable};
-use crate::integrations::agent_cli::launcher::generate_launcher_script;
+use crate::integrations::agent_cli::launcher::{
+    generate_launcher_script, managed_launcher_file_name,
+};
 
 /// Materialize one artifact from its pinned, fenced `Binary`/`Archive` source:
 /// resolve this platform's target, download it, verify the sha256, place the
@@ -114,7 +116,7 @@ pub(super) fn install_agent_process_from_pin(
     reporter: Option<&InstallProgressReporter>,
 ) -> Result<Option<InstalledArtifactResult>, InstallError> {
     let managed_dir = artifact_root(runtime_home, kind, &ArtifactRole::AgentProcess);
-    let launcher_path = managed_dir.join(format!("{}-launcher", kind.as_str()));
+    let launcher_path = managed_dir.join(managed_launcher_file_name(kind.as_str()));
     let path_prefixes = launcher_path_prefixes(runtime_home, kind);
     let launcher_env = managed_launcher_env(kind);
 
--- a/anyharness/crates/anyharness-lib/src/domains/agents/installer/service.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/installer/service.rs
@@ -129,6 +129,9 @@ impl From<LauncherError> for InstallError {
                 program: "launcher".into(),
                 message: error.to_string(),
             },
+            LauncherError::UnsupportedBatchValue(value) => Self::InvalidInstallSpec(format!(
+                "windows batch launcher cannot embed a literal '\"' in value: {value:?}"
+            )),
         }
     }
 }
--- a/anyharness/crates/anyharness-lib/src/domains/agents/readiness/artifacts.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/readiness/artifacts.rs
@@ -384,10 +384,21 @@ fn launcher_uses_binary_hint(launcher_path: &Path, candidate_binaries: &[String]
     let Ok(contents) = std::fs::read_to_string(launcher_path) else {
         return false;
     };
-    // Unix launchers run the target via `exec "<binary>" ...`; windows batch
-    // launchers have no `exec` and instead spawn it as `"<binary>" ... %*`
-    // (see `launcher.rs::build_batch_launcher_script`). Check both shapes so
-    // this diagnostic still fires on a `.cmd` launcher.
+    // NOTE (corrected after review): this whole match is already inert, on
+    // BOTH platforms, for any launcher this crate actually generates.
+    // Generated launchers always exec an ABSOLUTE path
+    // (`exec "/managed/dir/cursor-agent"` / `"C:\managed\dir\cursor-agent.exe"`),
+    // never the bare `candidate_binaries` name, so neither
+    // `exec "{binary}"` nor the windows-shaped `"{binary}" ` this PR added
+    // can ever match a substring that has a path prefix immediately before
+    // the opening quote. That was true of the pre-existing unix-only check
+    // before this PR too — it is a pre-existing latent gap, not something
+    // introduced or fixed here. The windows-shaped branch below is kept only
+    // for symmetry with the (equally inert) unix branches; it does NOT
+    // restore this diagnostic on `.cmd` launchers. Left as a known follow-up
+    // rather than fixed here, since fixing it for real means matching on the
+    // exec target's file name/stem rather than string-searching for the
+    // bare candidate name.
     candidate_binaries.iter().any(|binary| {
         contents.contains(&format!("exec \"{binary}\""))
             || contents.contains(&format!("exec {binary}"))
--- a/anyharness/crates/anyharness-lib/src/domains/agents/readiness/artifacts.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/readiness/artifacts.rs
@@ -10,6 +10,7 @@ use crate::domains::agents::model::*;
 use crate::integrations::agent_cli::executable::{
     find_in_path, find_real_binary_in_path, is_valid_executable,
 };
+use crate::integrations::agent_cli::launcher::managed_launcher_file_name;
 
 pub(super) fn resolve_native_artifact(
     spec: &NativeArtifactSpec,
@@ -171,7 +172,7 @@ pub(super) fn managed_launcher_candidates(
     executable_relpath: Option<&Path>,
 ) -> Vec<PathBuf> {
     let mut paths = vec![];
-    paths.push(managed_dir.join(format!("{}-launcher", kind.as_str())));
+    paths.push(managed_dir.join(managed_launcher_file_name(kind.as_str())));
 
     if let Some(executable_relpath) = executable_relpath {
         paths.push(managed_dir.join(executable_relpath));
@@ -383,9 +384,15 @@ fn launcher_uses_binary_hint(launcher_path: &Path, candidate_binaries: &[String]
     let Ok(contents) = std::fs::read_to_string(launcher_path) else {
         return false;
     };
+    // Unix launchers run the target via `exec "<binary>" ...`; windows batch
+    // launchers have no `exec` and instead spawn it as `"<binary>" ... %*`
+    // (see `launcher.rs::build_batch_launcher_script`). Check both shapes so
+    // this diagnostic still fires on a `.cmd` launcher.
     candidate_binaries.iter().any(|binary| {
         contents.contains(&format!("exec \"{binary}\""))
             || contents.contains(&format!("exec {binary}"))
+            || contents.contains(&format!("\"{binary}\" "))
+            || contents.contains(&format!("\"{binary}\"\r\n"))
     })
 }
 
--- a/anyharness/crates/anyharness-lib/src/domains/agents/readiness/compatibility.rs
+++ b/anyharness/crates/anyharness-lib/src/domains/agents/readiness/compatibility.rs
@@ -6,6 +6,7 @@ use super::paths::artifact_root;
 use crate::domains::agents::installer::seed;
 use crate::domains::agents::model::*;
 use crate::integrations::agent_cli::executable::find_in_path;
+use crate::integrations::agent_cli::launcher::managed_launcher_file_name;
 
 pub(super) fn detect_runtime_compatibility_issue(
     descriptor: &AgentDescriptor,
@@ -78,7 +79,7 @@ pub(super) fn detect_runtime_compatibility_issue(
         .map(|spec| spec.program.display().to_string())
         .unwrap_or_else(|| {
             artifact_root(runtime_home, &descriptor.kind, &ArtifactRole::AgentProcess)
-                .join("claude-launcher")
+                .join(managed_launcher_file_name(descriptor.kind.as_str()))
                 .display()
                 .to_string()
         });
--- a/anyharness/crates/anyharness-lib/src/integrations/agent_cli/launcher.rs
+++ b/anyharness/crates/anyharness-lib/src/integrations/agent_cli/launcher.rs
@@ -9,6 +9,40 @@ pub enum LauncherError {
     PathJoin(#[from] std::env::JoinPathsError),
     #[error("launcher io error: {0}")]
     Io(#[from] std::io::Error),
+    /// The windows batch emitter refuses to embed a literal `"` in a value: a
+    /// batch `set "K=V"`/quoted-argument value has no escape for an embedded
+    /// quote the way POSIX shells have `'\''`, so silently writing one would
+    /// produce a script that tokenizes differently than intended. Fail closed
+    /// instead of shipping a launcher nobody can reason about.
+    #[error("value is not representable in a windows batch launcher (contains '\"'): {0:?}")]
+    UnsupportedBatchValue(String),
+}
+
+/// The managed launcher's file name for `kind`. Unix identifies an executable
+/// by its permission bits, so a `#!/bin/sh` script needs no suffix. Windows
+/// has no exec bit — executability comes entirely from the extension — so the
+/// windows arm gets `.cmd`. This is the ONE place that decides the name; every
+/// site that constructs or resolves a managed launcher path calls through
+/// here so the write side and the read side can never drift apart.
+pub(crate) fn managed_launcher_file_name(kind_str: &str) -> String {
+    #[cfg(windows)]
+    {
+        windows_launcher_file_name(kind_str)
+    }
+    #[cfg(not(windows))]
+    {
+        unix_launcher_file_name(kind_str)
+    }
+}
+
+#[cfg(any(windows, test))]
+fn windows_launcher_file_name(kind_str: &str) -> String {
+    format!("{kind_str}-launcher.cmd")
+}
+
+#[cfg(any(not(windows), test))]
+fn unix_launcher_file_name(kind_str: &str) -> String {
+    format!("{kind_str}-launcher")
 }
 
 /// Regenerate a LIVE managed launcher without ever letting `fs::write` land
@@ -83,6 +117,35 @@ pub(crate) fn generate_launcher_script(
     env: &HashMap<String, String>,
     path_prefixes: &[PathBuf],
 ) -> Result<(), LauncherError> {
+    let script = build_launcher_script(exec_path, extra_args, env, path_prefixes)?;
+    std::fs::write(launcher_path, script)?;
+    make_executable(launcher_path)?;
+    Ok(())
+}
+
+fn build_launcher_script(
+    exec_path: &Path,
+    extra_args: &[String],
+    env: &HashMap<String, String>,
+    path_prefixes: &[PathBuf],
+) -> Result<String, LauncherError> {
+    #[cfg(windows)]
+    {
+        build_batch_launcher_script(exec_path, extra_args, env, path_prefixes)
+    }
+    #[cfg(not(windows))]
+    {
+        build_unix_launcher_script(exec_path, extra_args, env, path_prefixes)
+    }
+}
+
+#[cfg(any(not(windows), test))]
+fn build_unix_launcher_script(
+    exec_path: &Path,
+    extra_args: &[String],
+    env: &HashMap<String, String>,
+    path_prefixes: &[PathBuf],
+) -> Result<String, LauncherError> {
     let mut script = String::from("#!/bin/sh\nset -e\n");
 
     if !path_prefixes.is_empty() {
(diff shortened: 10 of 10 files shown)


pablonyx:
## Verdict: no blocking defect in this diff. Safe to merge as-is.

The implementation is correct as far as it can be checked off Windows, the rename is complete (I re-grepped independently rather than trusting the list), and — the thing I was told not to take on trust — the tests are genuinely load-bearing: **every one of the four requested "keep the signature, make it subtly wrong" mutations goes red**, plus eight more I added. Four non-blocking findings and one scope caveat below, then the beta question.

---

## 1. The real negative control (executed)

The author's negative control is a compile failure, which only proves wiring. I ran the mutation form instead: every function present, signature-compatible, subtly wrong.

**Methodology, stated plainly because it is not `cargo test`.** A cargo build was running on this machine for essentially the whole session and swap sat at 11.0–11.9 GB of 12 GB (`vm.swapusage`, `kern.memorystatus_vm_pressure_level: 2`); this box has an OOM-crash history and the standing rule is one cargo build at a time. I did not start a second one. Instead I extracted `windows_launcher_file_name`, `unix_launcher_file_name`, `build_batch_launcher_script`, `batch_escape`, `batch_escape_arg` and **all 9 new test functions** out of `launcher.rs` at `1c7cdf0` and compiled them with bare `rustc --test` (std only, `LauncherError` reduced to a plain enum with the same two variants). The extraction was done programmatically and I verified byte-for-byte that each function body and the test block appear verbatim in the PR source:

```
VERBATIM fn windows_launcher_file_name(   99
VERBATIM fn unix_launcher_file_name(      92
VERBATIM fn build_batch_launcher_script(  916
VERBATIM fn batch_escape(                 195
VERBATIM fn batch_escape_arg(             227
VERBATIM test bodies
```

Baseline, unmutated:

```
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

(9 here + `shell_escape_leaves_simple_tokens_unquoted`, `shell_escape_quotes_shell_sensitive_tokens`, `atomic_regen_preserves_previous_inode_until_promotion` = the 12 the PR body reports.)

### The four mutations I was asked for — all red

| Mutation | Result |
|---|---|
| **M1** `batch_escape_arg` returns its input unchanged (`launcher.rs:261`) | **RED** |
| **M2** `build_batch_launcher_script` omits the `exit /b %ERRORLEVEL%` line (`launcher.rs:230`) | **RED** |
| **M3** `windows_launcher_file_name` returns the extension-less name (`launcher.rs:39`) | **RED** |
| **M4** PATH prefix line emits unix `:` instead of `;` (`launcher.rs:216`) | **RED** |

Literal output:

```
### M1 batch_escape_arg returns input unchanged
test tests::batch_escape_arg_quotes_only_whitespace_tokens ... FAILED
test tests::batch_script_quotes_whitespace_args_and_forwards_the_rest ... FAILED
test result: FAILED. 7 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out

### M2 omit `exit /b %ERRORLEVEL%`
test tests::batch_script_starts_with_echo_off_and_ends_with_errorlevel_propagation ... FAILED
test result: FAILED. 8 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

### M3 windows_launcher_file_name returns extension-less name
test tests::windows_launcher_file_name_gets_cmd_extension ... FAILED
test result: FAILED. 8 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

### M4b PATH prefix line uses `:` instead of `;`   (production line only)
test tests::batch_script_prepends_path_prefixes_ahead_of_percent_path ... FAILED
test result: FAILED. 8 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

Methodology note on M4: my first attempt at it came back **green**, and that was my bug, not the suite's. The string `set \"PATH={joined};%PATH%\"` appears both in the emitter and in the test's own `expected` (`launcher.rs:380`), so a naive replace mutated both sides and the tautology passed. Re-anchored on the `script.push_str(&format!(...))` production line only (M4b above) it goes red correctly. Flagging it because the same shape would bite anyone else repeating this.

### Eight more mutations I added — all red except one

```
M5  batch_escape stops doubling %                        -> RED (2 tests)
M6  drop the leading `@echo off`                          -> RED (1 test)
M7  batch_escape stops rejecting embedded `"`             -> RED (2 tests)
M8b PATH prefix appended after %PATH% instead of before   -> RED (1 test)
M9  drop the trailing ` %*` arg forwarding                -> RED (2 tests)
M10 emit `set K=V` unquoted instead of `set "K=V"`        -> RED (2 tests)
M11 stop quoting the exec path                            -> RED (2 tests)
M12 drop the PATH prefix line entirely                    -> RED (1 test)
M1b batch_escape_arg keeps whitespace-quoting but loses
    % doubling and quote rejection                        -> GREEN  <-- see finding 1
```

This is a real test suite, not a wiring check. The sibling-PR failure mode I was sent to look for is not present here.

---

## 2. Findings (all non-blocking)

**Finding 1 — one narrow coverage hole: `%` and `"` are never driven through the *argument* path.**
`launcher.rs:261`. M1b keeps whitespace quoting but replaces `batch_escape(s)?` with `s.to_string()`, and the suite stays green: `batch_escape_doubles_percent_...` and `batch_escape_rejects_embedded_double_quote` (`launcher.rs:419`, `:425`) exercise `batch_escape` directly, and `batch_script_propagates_percent_and_quote_handling_end_to_end` (`:437`) drives `%`/`"` only through an **env value**, never through `extra_args`. Low severity because `batch_escape_arg` delegates to a directly-tested function, but one extra assertion (`batch_escape_arg("50%")` and a quote-bearing arg) would close it.

**Finding 2 — `batch_escape_arg` under-escapes: `&`, `|`, `<`, `>`, `^`, `(`, `)` are load-bearing in batch too, not just `%` and `"`.**
`launcher.rs:261`. The docstring at `:234` says "two characters that are actually load-bearing"; that is true *inside* a quoted region and false outside it. The exec path is always quoted (`launcher.rs:226`) and env values always land inside `set "K=V"` (`:220`), so both are safe. But an `extra_args` token is only quoted **if it contains whitespace**, so `foo&bar` would be emitted bare and cmd would treat `&` as a command separator. Not reachable today: I checked every value that can reach `extra_args` at the three call sites and they are `[]` (claude/codex git pins), `["acp"]` (cursor, opencode), `["agent","stdio"]` (grok) in `catalogs/agents/catalog.json`, and `["acp"]` / `["agent","stdio"]` in `catalogs/agents/registry.json`; the only registry env value is `DISABLE_AUTOUPDATER=1`. Since the catalog is remotely syncable (`catalog/sync.rs`, `PUT /v1/catalogs/agents`), this is a latent trap for a future pin. One-line fix: always quote the token, as the exec path already does.

**Finding 3 — the `launcher_uses_binary_hint` "fix" is inert; the PR body overstates it.**
`readiness/artifacts.rs:383-396`. The added patterns are `"{binary}" ` and `"{binary}"\r\n` where `binary` is a bare catalog name (`cursor-agent`, `grok`). But every launcher this repo generates execs an **absolute path** — the batch line is `"C:\...\registry_binary\cursor-agent.exe" acp %*`, which does not contain the substring `"cursor-agent"`. So the new patterns cannot match a generated `.cmd`, exactly as the pre-existing `exec "{binary}"` patterns cannot match a generated unix launcher (`exec "/…/registry_binary/cursor-agent"`). The only content that matches either shape is the hand-written fixture in `readiness/service_tests.rs:158`. The change is harmless (I checked it cannot false-positive on the generated unix shape either) but it does not restore a diagnostic that was working on unix — that diagnostic is already dead on the production path on both platforms, and `has_managed_registry_binary_for_names` suppresses the issue anyway. Please downgrade the claim in the body rather than leaving a reviewer to believe parity was restored.

**Finding 4 — `generate_launcher_script_atomic`'s documented invariant is false on Windows, and this PR is what makes that path live there.**
`launcher.rs:56-97`. The comment says "POSIX rename replaces the directory entry while a running managed session keeps the previous inode open (no kills, no waiting — Update Flow FR-3 / R2.5)". Windows has no inode semantics, and cmd.exe reads a batch file **incrementally**, holding a handle open without `FILE_SHARE_DELETE` for the life of the script — which is the well-known reason you cannot edit or replace a running `.bat`/`.cmd`. So `std::fs::rename(live, &previous)` at `:81` should fail with a sharing violation while a Windows session is running, where unix succeeds silently. It fails *safe* (`:92` restores the previous launcher and the error propagates as an install failure), so this is not corruption — but "regenerate the launcher without disturbing a running session" is a documented product behaviour (FR-3) that this PR quietly makes platform-conditional. Reasoned from reading only; I could not execute it. Worth a comment on `promote_launcher` and a follow-up, not a block.

Two smaller notes, no action needed: `set "K="` **unsets** a variable in cmd rather than setting it empty, so an empty env value diverges from unix `export K=''` (no registry entry has one today); and `!` is safe as the author says *unless* the machine has `HKCU\Software\Microsoft\Command Processor\DelayedExpansion=1`, which `cmd /d /c` does not disable (`/d` only kills AutoRun).

---

## 3. What I verified, split by method

**By execution** (all of the above `rustc --test` runs), plus the literal script the emitter produces for claude on Windows, rendered from the verbatim extraction:

```
@echo off<CRLF>
set "DISABLE_AUTOUPDATER=1"<CRLF>
"C:\Users\...\agents\claude\agent_process\node_modules\.bin\claude-agent-acp" %*<CRLF>
exit /b %ERRORLEVEL%<CRLF>
```

That render settles the `%ERRORLEVEL%` timing question you raised: the line is a **standalone final statement, not inside a parenthesised block**, so batch's per-line parse expands it after the preceding command has set it. The classic early-expansion bug needs an `if (…)` / `for (…)` block and there is none.

**By reading, item by item:**

- **Rename completeness.** I re-grepped `-launcher` and `launcher_path` across the whole crate independently. The author's list is complete. Confirmations worth recording: the two hand-rolled staging names (`installer/npm.rs:180`, `installer/pinned.rs:212`) are never executed — `ArchiveTreeActivation::activate_launcher` (`installer/downloads/activation.rs:96-119`) renames them onto `self.paths.launcher_path`, which already carries `.cmd`. `staged_launcher_sibling` (`launcher.rs:100`) produces `.claude-launcher.cmd.next` / `.claude-launcher.cmd.launcher-previous`, both rename-only. `is_valid_executable` is `path.is_file()` on non-unix (`integrations/agent_cli/executable.rs:47`), so a `.cmd` passes. Resolution takes the first *existing* candidate at `readiness/artifacts.rs:69,99` and the `.cmd` is candidate #0.
- **Persisted paths.** You asked me to look hard here. `ResolvedArtifact` (`model.rs:524`) has no serde derive and is recomputed per resolution — nothing durable. `install-manifest.json` **does** persist `path` as a string (`installer/manifest.rs:39,84`), but the only readers are `version_for` and `launch_options/basis.rs:25`, which hashes role/version/sha256/source and never the path. On unix the name is unchanged anyway, and no Windows install has ever existed, so there is no previously-written record to invalidate. No basis-revision churn for existing users.
- **`std::process::Command` and `.cmd` (your item 3).** Verified against the actual std source at the toolchain in play (no `rust-toolchain.toml` in the repo; CI is `dtolnay/rust-toolchain@stable` everywhere; local is 1.95.0 — all well past the 1.77.2 BatBadBut fix). `library/std/src/sys/process/windows.rs` detects `.bat`/`.cmd` by UTF-16 suffix match and substitutes `command_prompt()` (`GetSystemDirectoryW` + `\cmd.exe`) as the program, building the line via `args::make_bat_command_line`. **The author's claim holds: zero changes needed at `live/sessions/driver/process.rs:93`.** The CVE-era restriction is narrower than feared — `make_bat_command_line` errors only on a script path containing `"` or ending in `\`, and on arguments containing `\r`, `\n` or NUL; `%` is neutralised with the `%%cd:~,%` trick and quotes are backslash-escaped. Our script path is a plain profile path and the args reaching `Command::args` are `descriptor.launch.default_args` plus `settings_extra_args` (`--hostname 127.0.0.1 --port <n>` from `opencode_sidedoor.rs:52`, otherwise empty per `launch_policy.rs:198`). No newlines, no NULs. Not blocking.
- **`@echo off`.** Correct and necessary: `cmd /d /c` starts with ECHO ON, so without it every line would be written to the stdout that carries ACP JSON-RPC.
- **Exit-code propagation.** `exit /b N` in a script invoked as `cmd /c script.cmd` terminates the script and cmd exits with N. Read-derived, standard documented behaviour, not executed. See the caveat in the closing paragraph about when the line is reached at all.
- **`compatibility.rs:81`.** No behaviour change: `claude_launch_requires_node` returns false for any kind other than Claude (`compatibility.rs:95`), so the new `managed_launcher_file_name(kind)` renders exactly the old literal `claude-launcher` on unix.

## 4. What I could not verify

- **Nothing was executed on Windows, and this branch gets no Windows signal at all.** I confirmed `.github/workflows/windows-compile-check.yml` exists on neither `origin/main` nor this branch — it arrives with #2142, which this PR is deliberately not stacked on. The author states this; it is accurate.
- **#2147's integration-test escape hatch is genuinely unavailable here.** I checked: `generate_launcher_script`, `build_batch_launcher_script`, `batch_escape`, `batch_escape_arg` and `managed_launcher_file_name` are all `pub(crate)`, and there is no public API that reaches the emitter. A `tests/*.rs` crate cannot call any of them. The author's disclosure is correct, not an excuse.
- **I did not run `cargo test -p anyharness-lib`** (reason and method substituted are in section 1). So the crate-level compile of the real `#[cfg(any(windows, test))]` gating, and the read-site tests in `readiness/service_tests.rs` / `installer/tests.rs` / `api/router_tests.rs`, rest on the author's run plus the in-progress `Cargo check & test` job on this PR — not on mine. One thing I would want the CI job to confirm: on a **windows + test** build `build_unix_launcher_script` (`launcher.rs:143`) is compiled but never called, which is a `dead_code` warning; I found no `deny(warnings)` in the crate or CI, so it should be cosmetic.
- Findings 2 and 4 are read-derived. Neither was executed.

---

## 5. After #2152 + #2145 + #2146 + #2149, can a Windows beta user install and complete one agent turn?

Probably not yet, and this PR is not the reason — it does the job it claims. Two things upstream of it look like they still break the claude/codex path, and both are outside this diff. First, `installer/npm.rs:388,474` spawn `Command::new("npm")`; on Windows `resolve_exe` treats a bare extension-less name as a PATH search that appends `.exe`, and Node ships `npm.cmd`, not `npm.exe`, so a managed npm install should fail with `NotFound` before a launcher is ever written — and claude and codex are both `kind: git` npm pins in `catalogs/agents/catalog.json`, which #2149 does not touch (it fixes the pinned **native** binary and the source-build name, not `executable_relpath`). Second, even granting an installed tree, the emitted launcher execs `…\node_modules\.bin\claude-agent-acp` with **no extension**, so it runs only if cmd.exe's PATHEXT search happens to pick up the sibling `claude-agent-acp.cmd` npm generates — and if it does, that is a batch file invoked without `call`, which transfers control permanently and means `exit /b %ERRORLEVEL%` never executes (the exit code still propagates, so this is a dead line rather than a bug). Add that `launcher_path_prefixes` looks for bundled node at `node/<triple>/bin/node.exe` (`seed/mod.rs:175`) while `scripts/build-agent-seed.mjs:101` emits a `x64-win32`-style directory name, and the seed lane looks unlikely to supply node on Windows either. **The strongest single reason to think a beta user still cannot complete a turn is that the launcher's exec target is an extension-less npm shim on a platform where extension-less files are not executable — the very defect class this PR fixes one level up, left unfixed one level down.** I would want that traced (or an explicit "PATHEXT covers it, verified on Windows") before calling the runtime path unblocked; it is a separate PR, not a change to this one.

*Reviewed at head `1c7cdf0bd8c710e32c459fd75567fd1740d6d8e2`. No repository state was modified.*

pablonyx:
## Reviewer of record — post-rebase pass on #2152

I did not write this PR. Verdict below separates **what I executed** from **what I read**.

---

### 1. Rebase is clean

`git range-diff f3122f08fc...24d6299510` — both commits carry over bit-for-bit:

```
-:  ---------- > 1:  95626925f5 ci(release): unify deploys into one main-to-prod pipeline (#2140)
-:  ---------- > 2:  ccb53edc85 fix(sessions): unblock session start and prompt send across the launch handoff (#2144)
-:  ---------- > 3:  956d2426b7 fix(release): push version bump with bypass-capable token (#2156)
1:  1c7cdf0bd8 = 4:  bc13d8144a fix(agents): emit a Windows-executable managed launcher
2:  f3122f08fc = 5:  24d6299510 fix(agents): close batch-arg escaping hole; fix npm.exe/.bin exec resolution on Windows
```

Three new upstream commits, zero content drift in this PR's two commits. Nothing to re-review from the rebase itself.

---

### 2. The real-execution proof is genuine — I confirmed both runs

**Positive: run [32446919674](https://github.com/proliferate-ai/proliferate/actions/runs/32446919674)** — `success`, workflow `Windows managed-agent install + spawn`, head `77055343d7`. I confirmed by ancestry that this run really contains this PR at `f3122f08fc` *and* #2149 (`fca3cb11f7`):

```
77055343d7 test(agents): cover claude, the default agent, on windows too
d4420ab102 Merge commit 'f3122f08fc' into ci/windows-managed-agent-e2e
b06a8dde06 Merge remote-tracking branch 'origin/fix/pinned-installer-windows-exe' into ci/windows-managed-agent-e2e
fca3cb11f7 fix(agents): install pinned agent binaries with the windows executable extension
```

Literal lines from the `windows-latest` job:

```
PATHEXT = ".COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.CPL"
control: bare `npm --version` -> NOT SPAWNABLE: program not found (kind NotFound, raw os error None)
installer program name = "npm.cmd"
npm.cmd --version -> status ExitStatus(ExitStatus(0)) stdout "10.9.8" stderr ""
```

That single block settles the PATHEXT question inside the *positive* run: `.CMD` is in `PATHEXT` and `Command::new("npm")` still fails. Rust's spawner appends only `.exe`. Confirmed, not assumed.

npm's cmd-shim shape, confirmed against a real `.bin` (three files per binary, exactly as assumed):

```
file 348  claude-agent-acp.cmd
file 435  claude-agent-acp
file 897  claude-agent-acp.ps1
```

Emitted launcher and its spawn:

```
file 355  claude-launcher.cmd
@echo off
set "PATH=...\agents\claude\native;%PATH%"
set "DISABLE_AUTOUPDATER=1"
"...\agents\claude\agent_process\node_modules/.bin/claude-agent-acp.cmd" %*
exit /b %ERRORLEVEL%
spawned pid 3652
```

Same for `grok-launcher.cmd` (pid 1856, exit 0). `test result: ok. 3 passed; 0 failed`.

**Negative control: run [32445803852](https://github.com/proliferate-ai/proliferate/actions/runs/32445803852)** — `failure` at head `320b4d16a1`, which sits on the merge that has the launcher commit `1c7cdf0bd8` but **not** `f3122f08fc`. Correct isolation. Literal failure:

```
`Command::new("npm")` could not be spawned: program not found (kind NotFound, raw os error None).
This is installer/npm.rs's exact call shape.
install_agent_with_pins failed: io error: program not found (kind Other)
test result: FAILED. 0 passed; 2 failed
```

This is as strong a proof as this change can get. The code is right.

---

### 3. Independent execution of my own — standalone `rustc --test` on the extracted string functions

I extracted `build_batch_launcher_script` / `batch_escape` / `batch_escape_arg` / the name resolvers verbatim from this PR's head into a standalone file and built them with a bare `rustc --test` (no cargo). Confirmed correct: `%` doubling on both the value and the argument path; `"` rejected on both paths; the `PATH` line doubles a literal `%` in a prefix while keeping `%PATH%` live (`set "PATH=/pct%%dir;%PATH%"`); both filename arms. Two holes fell out — see finding **N1** below.

---

## Blocking

### B1 — CI is red, and it is this PR's own doing

`Repo shape checks` fails, and `ci-ok` fails behind it:

```
anyharness/crates/anyharness-lib/src/domains/agents/installer/npm.rs: PROD-SIZE-1
  found: 636 lines (max 600)
```

`npm.rs` is 526 lines on `main` and 636 at this head. Not a rebase artifact — the +110 comes from this PR's `npm_program_name*` / `platform_npm_bin_relpath*` block and its tests. Either split along a seam or add a `[[max_lines]]` entry to `lints/anyharness/ratchets.toml` with a written reason; note the gate's own text says net-new exceptions want founder approval, so the split is likely the cheaper route. Nothing merges while `ci-ok` is red.

### B2 — The body still says the execution never happened

`## What remains unverified — honestly`, first bullet, still reads **"No real Windows execution of this launcher happened."** and argues it was unreachable because the helpers are `pub(crate)`. That is now false twice over: the execution happened, and the proof test reached it through the *public* installer API (`install_agent_with_pins`) rather than by calling the private helpers, so the stated obstacle was never the real one.

The same section still lists as open the exact three things run 32446919674 closed:

- `Command::new("npm.cmd")` "resolves and spawns" — **proven**, plus the bare-`npm` control proven *not* spawnable with `.CMD` in `PATHEXT`.
- npm's cmd-shim sibling shape "if npm writes something in a different shape than assumed here" — **proven** (`<name>` / `<name>.cmd` / `<name>.ps1`).
- "no install-flow test exercises the two upstream fixes end-to-end" — **it does now**.

Please rewrite that section to cite run **32446919674** (positive) and **32445803852** (negative control) with the literal lines, and demote what is genuinely still open (the FR-3 rename-under-lock claim, the process-tree claim, non-zero exit propagation) to a short honest remainder. The body is the durable artifact; right now it understates the evidence enough that a future reader would re-do the work.

### B3 — The "Stacking" section's disjointness claim is false; #2149 conflicts

The body says this PR "touches a disjoint file (`launcher.rs` and its 4 real callers) from #2145/#2149 (which touch `pinned.rs`'s `target_path`/archive-extraction naming, a different path than the launcher)."

An in-memory three-way merge (`git merge-tree --write-tree pr/2152 pr/2149`, no worktree, no build) produces two real content conflicts:

```
CONFLICT (content): Merge conflict in .../domains/agents/installer/agent_process.rs
CONFLICT (content): Merge conflict in .../domains/agents/installer/npm.rs
```

- `installer/npm.rs` around **:487** — this PR keeps `pub(super) fn platform_binary_filename` and grows the new npm helpers directly beneath it; #2149 *deletes* that function, having moved it to `integrations/agent_cli/executable.rs` as `pub(crate)` (and made it idempotent).
- `installer/agent_process.rs` **:4** — this PR writes `use super::npm::{platform_binary_filename, platform_npm_bin_relpath};` where #2149 writes `use crate::integrations::agent_cli::executable::{is_valid_executable, platform_binary_filename};`.

Correct resolution, whichever lands second: import `platform_binary_filename` from `executable.rs`, keep `platform_npm_bin_relpath` in `npm.rs`, drop the duplicate definition. **Neither PR body warns about this** — #2149's body doesn't mention #2152 at all. Please put that resolution note in this body (and ideally in #2149's), because a merge-order accident here silently reintroduces a second `platform_binary_filename` or drops #2149's idempotence fix. It also shrinks B1: dropping the duplicate takes ~7 lines off `npm.rs`.

---

## Non-blocking

### N1 — `batch_escape` rejects `"` but silently passes `\n`, and the env *key* is never escaped at all

Demonstrated with the standalone `rustc --test` harness against this PR's own code. Rendered output, verbatim:

```
// newline in an env value
@echo off
set "K=a
whoami"
"C:\m\a.exe" %*

// newline in an arg
@echo off
"C:\m\a.exe" "a
whoami
rem " %*

// unescaped env key
@echo off
set "K"=x" & whoami "=1"
```

cmd.exe parses a batch file line by line, so a quote cannot span lines — `whoami` lands on its own line and runs. This is the *same* hazard class the always-quote fix was made for, argued on the same grounds ("the catalog is remotely syncable, so that's a property of today's data, not of the code"), and by that standard a newline deserves the same treatment as `"`. Note the unix arm does not have this hole: `shell_escape` single-quotes anything containing whitespace, and a newline is whitespace — so it is a Windows-only divergence.

Reachability is low and I am not calling it blocking: `env` comes from `built_in_registry()` → `bundled_agent_descriptors()`, and the staged registry/catalog lane is minisign-verified with baked pubkeys, so this is publisher-controlled, not attacker-controlled. Registry validation (`registry/validation.rs:137`) only checks that an env var *name* is non-empty — no charset constraint on either name or value. Fix is two lines: extend `batch_escape`'s reject to `\r` and `\n`, and route `key` through `batch_escape` in the `set` loop alongside the value.

### N2 — #2147 interaction: I verified it, and it holds

Read `process_kill_tree.rs` at #2147's head. `TreeTracker::absorb` grows the tracked set inside a `loop { ... if !grew { break } }` fixpoint over the whole snapshot, so adoption really is transitive across arbitrary depth, not direct-children-only; `DEPTH_LIMIT` is 64 and only affects ordering; the pass sorts deepest-first so the agent is terminated before its parent `cmd.exe`. The extra layer this PR introduces is handled. The disclosure in this body is accurate and the conclusion is right.

One refinement for accuracy, since the body reasons about it: on the actual claude/grok path the exec target is itself a `.cmd` (`.bin/<name>.cmd`), and batch invoked *without* `call` transfers control within the same `cmd.exe` rather than spawning a second one — so the real tree is `anyharness -> cmd.exe (launcher, then the shim) -> node.exe`, one extra layer as described, but for a slightly different mechanical reason than "spawns a child". Same conclusion either way.

### N3 — the `exit /b %ERRORLEVEL%` reasoning only applies to the `.exe` branch

Following from N2: because a batch file chained without `call` never returns to its caller, `exit /b %ERRORLEVEL%` is unreachable whenever the exec target is a `.cmd` shim — which is the claude and grok path. Exit-code propagation is still correct (the chained shim's code becomes the process's code), and run 32446919674 observed `ExitStatus(0)`, so nothing is broken. But the body's "**`%ERRORLEVEL%` timing is correct**" verification only covers the archive/`.exe` branch. Worth one sentence, and worth noting that **non-zero** propagation is still unobserved on either branch.

### N4 — FR-3 caveat lives only in the PR body

`generate_launcher_script_atomic`'s doc comment (`launcher.rs:48-55`) still states the invariant in unqualified terms — "POSIX rename replaces the directory entry while a running managed session keeps the previous inode open (no kills, no waiting — Update Flow FR-3 / R2.5)". The Windows divergence is disclosed thoroughly in the body, and the failure mode is genuinely fail-safe (`promote_launcher`'s `Err` arm restores the previous launcher). But the body evaporates on merge and the doc comment is what the next Windows maintainer reads. Please move two sentences of that disclosure into the doc comment. #2149 flags the mirror-image hazard for `MoveFileEx` over a running `claude.exe`, so the two notes belong together.

### N5 — `launcher_uses_binary_hint`: correctly downgraded, confirmed inert

Verified. The body's item 5 now states plainly that the hunk has no behavior effect and that the pre-existing unix branches were equally inert, and `artifacts.rs:386-398` carries the same correction as a code comment. Independently confirmed: generated launchers always emit an absolute exec path, so every pattern in that `any` needs a `"` immediately preceding the bare candidate name and there is always a directory prefix there. This matches what the run's rendered launchers show. Nothing further owed.

### N6 — `platform_npm_bin_relpath_windows` is not idempotent

It appends `.cmd` unconditionally, so an `executable_relpath` that already ends in `.cmd` becomes `.cmd.cmd`. Not reachable from today's registry (`executable_relpath` is `node_modules/.bin/<name>`, and the npm arm builds it from `executable_name`), but #2149 explicitly added idempotence to `platform_binary_filename` for exactly this bug class — worth matching while resolving B3, since both functions will be side by side in the reviewer's head at that moment.

---

## Verdict

**Not safe to merge as-is — but the blockers are CI hygiene and body accuracy, not the change.**

The code is correct and now empirically proven on a real Windows host: `npm.cmd` spawns where bare `npm` does not (with `.CMD` in `PATHEXT`), npm's cmd-shim writes the `.cmd` sibling this PR resolves, and the emitted `claude-launcher.cmd` runs and exits 0. I found no behavioral defect in the diff.

To clear:
1. **B1** — get `npm.rs` back under the 600-line cap (or ratchet it); `ci-ok` is red until then.
2. **B2** — rewrite `## What remains unverified` to cite runs 32446919674 / 32445803852; it currently claims the opposite of what happened.
3. **B3** — correct the "Stacking" disjointness claim and record the #2149 conflict resolution.

N1 (reject `\r`/`\n`, escape the key) and N4 (FR-3 caveat into the doc comment) are cheap and I would take them in the same push, but neither gates merge.

pablonyx (commented):
Reviewed only the new commit `24270510c8` ("extract npm platform resolution to its own module"). The rest of the PR was already cleared by an earlier reviewer. No cargo/rustc was run; verification below is diff-reading, `python3 scripts/check_max_lines.py` (pure Python), and `gh` queries against already-completed CI runs.

## Verdict: not blocking. Approve on this commit.

## 1. Pure move — verified byte-for-byte

Extracted the moved block from `npm.rs` at the pre-move tip `24d6299510` (lines 495-559 + the `windows_npm_resolution_tests` module) and diffed it against the corresponding body of the new `npm_platform.rs`. The only differences:

- `fn npm_program_name()` → `pub(super) fn npm_program_name()` (visibility widening, required — see §2).
- The `#[cfg(test)] mod windows_npm_resolution_tests {` line, which fell outside my extraction window on the "before" side but is present unchanged on both sides.

Everything else — including both doc comments (the `npm itself is a JS program...` comment on `npm_program_name_windows` and the `Resolve the executable...` comment on `platform_npm_bin_relpath`), all four test bodies, and every `#[cfg(...)]` attribute — is character-identical. This is a genuine pure move, not a rewrite-and-hope.

## 2. Visibility — correctly minimal, nothing newly reachable

`npm_platform.rs` and `npm.rs` are both direct children of the `installer` module, same as `agent_process.rs`. Before the move, `platform_npm_bin_relpath` was `pub(super)` in `npm.rs`, meaning "visible within `installer`" — exactly the scope `agent_process.rs` needs. After the move:

- `npm_platform::npm_program_name` is `pub(super)` (scoped to `installer`, since npm_platform.rs's own parent is `installer`) and is only ever imported by `npm.rs` (`use super::npm_platform::npm_program_name;`, verified via `git grep` at the commit — no other reference exists).
- `npm.rs`'s re-export, `pub(super) use super::npm_platform::platform_npm_bin_relpath;`, is `pub(super)` relative to npm.rs's own parent (also `installer`), i.e. the identical visibility scope the function had pre-move.

Net effect: the visibility surface reachable from outside `installer` is unchanged, and nothing inside `installer` gains access it didn't already have. The `pub(super) use` re-export is unusual but is doing exactly what it should here.

## 3. `#[cfg(any(windows, test))]` / `#[cfg(any(not(windows), test))]` split — intact

Confirmed via the diff in §1: all four gated functions (`npm_program_name_windows`, `npm_program_name_unix`, `platform_npm_bin_relpath_windows`, `platform_npm_bin_relpath_unix`) carry the identical cfg attributes post-move. Better than my own read of the diff: CI's `Cargo check & test` job runs `cargo nextest run --workspace --no-fail-fast` on `ubuntu-latest`. Because both cfg gates include `test`, all four split functions — including the two nominally "windows" ones — compile and run as part of that Linux CI job, and that job is green at this commit. That's actual, already-executed proof the move didn't silently drop platform coverage, independent of the hand-run negative control described in the commit message.

## 4. Size gate — passes, and for the stated reason; "component" cap is a red herring here

Ran `python3 scripts/check_max_lines.py` against a detached worktree checked out at `24270510c8`:

```
```

`npm.rs` is 535 lines (down from 636), `npm_platform.rs` is 114 lines — both well under the repo-wide 600-line `PROD-SIZE-1` cap, and neither is newly ratcheted in `lints/anyharness/ratchets.toml` (correctly — the file no longer needs one). I also read the script: `COMPONENT_MAX_LINES = 500` is gated to `apps/desktop/src/components/*.tsx` only (`max_lines_for()` in `scripts/check_max_lines.py`). It never applies to `anyharness/crates/**` Rust files, so it's not in play for either file here — the fix doesn't trade one violation for another.

On the ratchet-vs-shrink judgment call: I read `lints/anyharness/ratchets.toml`'s header myself. It says plainly: "A ratchet, not an exception ledger: max_lines is the observed count allowed for this file and may only shrink. Growth fails; so does a stale entry whose file is now under its allowance or gone." I agree with the commit's reading — adding a fresh entry for growth this same PR introduced would have been using the ratchet as an exception ledger, which the header explicitly forecloses. Shrinking the file is the correct fix.

## 5. CI at the new head — green, but two of the workflows named in my review brief don't exist on this PR at all

`gh pr checks 2152` and `gh api .../commits/24270510c8.../check-runs` both confirm `ci-ok` and `Repo shape checks` pass at this exact SHA. Literal: `ci-ok  pass  7s`, `Repo shape checks  pass  1m55s`.

However: neither `Windows compile check` nor `Windows managed-agent install + spawn` appears anywhere in the check-runs list for `24270510c8`, and neither workflow file exists in this branch's tree at all (`git show 24270510c8:.github/workflows/windows-compile-check.yml` → `fatal: path does not exist`; same for `windows-managed-agent-install.yml`). Tracing `git log --all` for those paths, both files live only on other, unmerged branches — specifically `windows-managed-agent-install.yml` exists only on `ci/windows-managed-agent-e2e` (PR #2153's own branch), and `windows-compile-check.yml` exists on yet another branch, neither of which #2152 has merged. So the caution in my brief ("`Windows compile check`'s continue-on-error makes its green meaningless; `ci-ok` doesn't see the managed-agent workflow, read it directly") describes #2153's checks, not #2152's. On #2152 there is nothing to read directly — those two workflows simply never ran here, which is a pre-existing property of this PR (not something this commit changed) and outside this commit's blast radius, but worth naming so nobody assumes #2152 got Windows e2e coverage it didn't get.

## 6. Consequences for #2153 — no forced rebuild, but the provenance record is now one commit stale

I read #2153's own PR description, which pins its `#2152` parent to tip `24d6299510` (pre-move) merged via `5227208b58`, and records the exact import-block merge resolution it needed for `agent_process.rs`/`npm.rs`. At that tip, `agent_process.rs` imports `platform_npm_bin_relpath` alone from `super::npm` (its `platform_binary_filename` comes from `executable.rs` post-#2149 there instead).

On #2152's own branch (unaffected by #2153's merge), `agent_process.rs` is untouched by `24270510c8` and still reads `use super::npm::{platform_binary_filename, platform_npm_bin_relpath};` — both names still resolve: `platform_binary_filename` is still defined directly in `npm.rs`, and `platform_npm_bin_relpath` now resolves through the `pub(super) use` re-export. Verified both files at the commit directly. So: no, this commit does not force any code change in #2153, and the specific claim in the commit message holds — `platform_npm_bin_relpath` stays reachable as `npm::platform_npm_bin_relpath` under either import shape (#2152's own, or #2153's merge-resolved variant), because the re-export preserves the path importers already use regardless of which internal module now backs it.

One nuance worth flagging, non-blocking: `24270510c8`'s diff to `npm.rs` touches the same import block (top-of-file `use` statements) that #2149/#2152's merge into #2153 already needed manual conflict resolution on. If #2153 (currently a "not for merge" scratch branch, explicitly pinned to `24d6299510`) is ever revived and rebased onto #2152's new tip, that import hunk will need to be re-resolved by hand again — it won't auto-merge cleanly a second time. That's a documentation/rebase-effort concern for whoever revives #2153 later, not a defect in this commit.

## 7. Negative control — real, but weaker evidence than the CI run itself

The commit's claimed control (mutate `npm_program_name_windows()` to `"npm"` and drop the `.cmd` suffix in `platform_npm_bin_relpath_windows`, i.e. two coordinated mutations) would flip exactly the two Windows-specific tests while leaving the two Unix-specific tests green — `2 passed; 2 failed` then `4 passed; 0 failed` restored is internally consistent with that. It's a legitimate discriminating control for the assertions' content.

But it's a control on the *original logic*, not on the *move*. It would have produced the identical result before this commit, run against the pre-move `npm.rs`. What actually discriminates the move-specific risks (dropped `mod` declaration, broken re-export, widened-too-far visibility, a cfg gate silently lost) is: (a) the crate compiling at all — if the `mod npm_platform;` line or either `use` were missing, the whole crate fails to build, not "2 passed; 2 failed" — and (b) the already-executed, real `Cargo check & test` nextest run at this commit passing on `ubuntu-latest`, which per §3 actually exercises all four moved tests including the Windows-gated ones. I'd treat the CI run as the stronger evidence here; the hand-described negative control is consistent with it but doesn't add coverage of the move itself.

## Things checked from the brief that didn't hold up

- `agent_process.rs`'s import is exactly `use super::npm::{platform_binary_filename, platform_npm_bin_relpath};` on #2152's own branch, as the commit message says — confirmed. But the brief's assumption that `Windows compile check` and the managed-agent workflow are part of #2152's check surface is wrong; neither workflow exists on this branch (see §5). That framing belongs to #2153, not this PR.

No blocking findings. The move is exactly what it claims to be, the size gate passes for the stated reason with the component cap correctly out of scope, CI is green for everything that actually runs on this branch, and #2153 needs no code changes as a result of this commit.

pablonyx (commented):
REBASE-CLEAN. Verified against new head bae56fba53262be1851d8bb125b0a81abc6c7276 (base main 908c54d125d).

Checked:
1. Diff --stat vs main matches exactly the expected 10 files: Cargo.lock, agent_process.rs, installer/mod.rs, npm.rs, npm_platform.rs (new), pinned.rs, service.rs, readiness/artifacts.rs, readiness/compatibility.rs, integrations/agent_cli/launcher.rs. No extra files.
2. Symbol coherence read directly off the blobs at HEAD:
   - installer/mod.rs declares `mod npm_platform;`.
   - `platform_binary_filename` is defined exactly once, in `integrations::agent_cli::executable` (main's #2149 version); npm.rs and agent_process.rs both import it from there, no local duplicate anywhere under installer/.
   - `npm_program_name` / `platform_npm_bin_relpath` are defined exactly once each, in the new npm_platform.rs. npm.rs re-exports `pub(super) use super::npm_platform::platform_npm_bin_relpath;` and agent_process.rs's `use super::npm::platform_npm_bin_relpath` resolves through that re-export cleanly.
   - agent_process.rs imports are exactly the claimed union (executable::{is_valid_executable, platform_binary_filename}, launcher::{generate_launcher_script_atomic, managed_launcher_file_name}, super::npm::platform_npm_bin_relpath) and every imported symbol is actually used in the file.
   - `managed_launcher_file_name` is defined once (pub(crate)) in launcher.rs and consumed consistently in pinned.rs, compatibility.rs, artifacts.rs, agent_process.rs.
   - `LauncherError` match in service.rs is exhaustive over all 3 variants (PathJoin, Io, UnsupportedBatchValue).
3. pinned_tests.rs, downloads.rs, executable.rs are absent from the diff vs main, i.e. #2149's changes to those files are untouched by this rebase.
4. No conflict markers anywhere in the changed files (`<<<<<<<`, `>>>>>>>`, or unpaired `=======`). launcher.rs's +324/-4 is exactly the reviewed content across both original commits: the `.cmd` launcher emitter (managed_launcher_file_name, build_batch_launcher_script, @echo off / exit /b %ERRORLEVEL% semantics) plus the always-quote batch_escape_arg fix and its tests. Nothing reads as a rebase artifact.
5. Cargo.lock diff is a single line, the proliferate package version bump (0.4.20 -> 0.4.21), nothing else.

No defects found in the rebase resolution.

PR #2153, opened 2026-08-21
test(agents): prove managed agent install and spawn on windows

pablonyx:
## Why

Every Windows fix so far has been derived by reading source, and each review round has found one more broken layer underneath the last: packaging, then the install path, then the launcher, then what the launcher invokes. Nobody knows how many layers remain because **the path has never been executed**.

This is the experiment that replaces that unbounded uncertainty with a result: on a real `windows-latest` runner, drive the production code through **install a managed agent, then spawn what the install produced**. No desktop app, no code signing, no LLM call, no auth.

## What is under test, all at once

- `installer/npm.rs` shelling out with `Command::new("npm")` (Node ships `npm.cmd` on Windows, not `npm.exe`)
- the emitted launcher: extension-less `#!/bin/sh` on unix, `.cmd` on Windows (#2152)
- what the launcher invokes: `…\node_modules\.bin\<name>` with no extension, relying on cmd.exe PATHEXT fallback
- the `.exe` managed-binary naming (#2149)
- `make_executable` being a no-op on Windows and `is_valid_executable` being only `path.is_file()` there
- whether the produced artifact is spawnable by `CreateProcess`

## Composition

This is a scratch branch that deliberately combines the two open Windows installer fixes so the end state under test is the *intended* one, not `main`. Eight commits, not one, verified against `gh pr view 2153 --json commits`:

- `origin/main`
- rebased `fix/pinned-installer-windows-exe` (#2149), tip `b4e673f817` (three commits: `70fbe91044`, `de1e1cc9e7`, `b4e673f817`), the `.exe` managed-binary path
- merged with rebased `fix/windows-managed-launcher` (#2152), tip `24d6299510` (two commits: `bc13d8144a`, `24d6299510`), the `.cmd` launcher emission, via merge commit `5227208b58`
- two commits adding and then extending the test and workflow: `cfdac10b93` (adds `windows_managed_agent_install.rs` and the workflow) and `db843f501d` (extends the test to cover `claude`)
- head: `db843f501d`

All SHAs above were resolved as real commit objects in this repo before writing them here.

### Merge conflicts

There were two conflicts on the merge into `5227208b58`, not one, and the second mattered more than the first.

The first was an import collision in `installer/npm.rs`/`agent_process.rs`: #2149 moved `platform_binary_filename` into `executable.rs` while #2152 added neighboring code to the same lines in `npm.rs`. Both sides' imports were kept and the duplicate definition was dropped.

The second, at `agent_process.rs:94`, is the one that actually decides Windows behavior: #2149's side computed `managed_pinned_binary_path(runtime_home, kind, &ArtifactRole::NativeCli)`, while #2152's side computed `managed_native_dir.join(kind.as_str())`. The resolution kept #2149's `managed_pinned_binary_path(...)`. This matters because `managed_pinned_binary_path` routes through `platform_binary_filename`, which appends `.exe` on Windows; `kind.as_str()` alone (`"claude"`, `"codex"`, …) does not. Had #2152's side won, the computed path would lack the `.exe` suffix on Windows while the file on disk is named `claude.exe`, so `is_valid_executable`, which on non-unix is a bare `path.is_file()`, would find nothing at that path and return `false`. `launcher_path_prefixes` would then silently drop the managed native CLI directory from the launcher's `PATH`, a Windows-only regression this PR's own test would not catch, since the test's `grok` and `claude` cases either declare no native CLI or resolve their native dir independently of this exact branch. Verified directly by reading both parents' versions of `agent_process.rs` at the merge and the current `is_valid_executable`/`platform_binary_filename`/`managed_pinned_binary_path` definitions at head `db843f501d`.

## Shape

`tests/windows_managed_agent_install.rs` is an integration test target, not a unit test. Precedent: #2147. `anyharness-lib`'s `lib test` target does not build for Windows today (ten pre-existing errors, all test-only POSIX assumptions in unrelated modules); an integration test compiles as its own crate against the public surface and never pulls those in.

Two agents are under test, not one. `grok`'s agent-process pin is a plain public npm package, so the install needs no credentials, and it declares no native CLI, so the run is one artifact; its executable shape (`node_modules/.bin/<name>`) is identical to claude's and codex's, which is the shape actually in doubt. `claude` is also covered as of `db843f501d`, and it is the more complete case: it is the default agent and the only one of the two that exercises both the pinned native `.exe` download and the git-sourced npm install in the same run.

Ubuntu runs the identical test as a control, so a red Windows result cannot be blamed on the test.

Nothing is stubbed, mocked, skipped, or `continue-on-error`'d. A failure here with a real Windows error message is the finding.

## Honest limits

- **No ACP handshake proven.** Both launchers exit 0 with empty stdout and empty stderr. That proves the process spawned and shut down cleanly on stdin EOF, since nothing is written to stdin and nothing is read back. It does not prove an ACP `initialize` exchange ran, since no JSON-RPC was sent or received.
- **Three of five agents are uncovered.** `cursor` and `opencode` have no `windows_*` targets at all in `catalogs/agents/catalog.json` (each lists only `macos_arm64`, `macos_x64`, `linux_x64`), so both fail `NoPinForPlatform` on Windows today and never reach the path this test exercises. `codex` does have Windows pins, but they point at `codex-x86_64-pc-windows-msvc.exe.tar.gz` / `codex-aarch64-pc-windows-msvc.exe.tar.gz`, a tar-extraction branch this test never touches.

## CI status

CI at `db843f501d` is **not fully green**. `Repo shape checks`, and therefore the `ci-ok` rollup, fail:

```
Repo shape checks / Check max source file length
anyharness/crates/anyharness-lib/src/domains/agents/installer/npm.rs: PROD-SIZE-1 — Source files stay under the repo-wide line cap
  found: 628 lines (max 600)
  rule: Every scanned .py/.rs/.ts/.tsx source file must stay under 600 lines
```

This debt originates in #2152, not in this branch's rebuild: `npm.rs` at #2152's own rebased tip `24d6299510` is already 636 lines, more than the 628 seen here. It is being fixed on #2152, not in this scratch branch.

### Two check-reading traps for whoever reads this PR's checks

- **`Windows compile check`'s green means nothing.** Both of its jobs, `runtime crates (control)` and `desktop crates (the unknown)`, are declared with `continue-on-error: true` in `.github/workflows/windows-compile-check.yml`, so that check reports success unconditionally regardless of whether `cargo check` actually passed.
- **`ci-ok` does not see the Windows managed-agent workflow.** `ci-ok`'s `needs:` list only covers jobs defined in `ci.yml` itself; `windows-managed-agent-install.yml` and `windows-compile-check.yml` are separate workflow files and sit outside that rollup gate entirely. Their results have to be read directly from the checks list, not inferred from `ci-ok`.

### The passing evidence that is real

Run [32450488248](https://github.com/proliferate-ai/proliferate/actions/runs/32450488248) passed both legs. Verified with `gh run view 32450488248 --repo proliferate-ai/proliferate` and the job logs; literal result lines:

```
install + spawn (windows-latest): test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 418.05s
install + spawn (ubuntu-latest): test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 79.91s
```

## Not for merge

Scratch experiment branch. Draft on purpose.


pablonyx:
## Result: green on `windows-latest`. Both agents installed and spawned.

Run: https://github.com/proliferate-ai/proliferate/actions/runs/32446919674 (head `77055343d7`)
`install + spawn (windows-latest)` **success** · `install + spawn (ubuntu-latest)` **success** (control)

```
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 425.39s
```

### Layer 1 — npm resolution, settled empirically

```
PATHEXT = ".COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.CPL"
control: bare `npm --version` -> NOT SPAWNABLE: program not found (kind NotFound, raw os error None)
installer program name = "npm.cmd"
npm.cmd --version -> status ExitStatus(ExitStatus(0)) stdout "10.9.8" stderr ""
```

`Command::new("npm")` genuinely does not resolve on Windows — despite `.CMD` being in `PATHEXT`, Rust's spawner only appends `.exe`. The first run of this job, before `f3122f08fc` was merged in, failed exactly there:

> `install_agent_with_pins failed: io error: program not found (kind Other)`

(https://github.com/proliferate-ai/proliferate/actions/runs/32445803852, head `320b4d16a1`)

### Layer 2 — what npm actually writes into `.bin` on Windows

Observed on a real NTFS filesystem, not asserted as a string. npm's cmd-shim writes **three** files per binary:

```
--- node_modules/.bin (claude, git-sourced)
  file       325  node-which.cmd
  file       330  anthropic-ai-sdk.cmd
  file       348  claude-agent-acp.cmd
  file       389  node-which
  file       399  anthropic-ai-sdk
  file       435  claude-agent-acp
  file       805  node-which.ps1
  file       825  anthropic-ai-sdk.ps1
  file       897  claude-agent-acp.ps1
```

So the `.cmd` sibling `platform_npm_bin_relpath` targets does exist.

### Layer 3 — the emitted launcher

```
--- launcher script (...\agents\claude\agent_process\claude-launcher.cmd)
@echo off
set "PATH=...\agents\claude\native;%PATH%"
set "DISABLE_AUTOUPDATER=1"
"...\agents\claude\agent_process\node_modules/.bin/claude-agent-acp.cmd" %*
exit /b %ERRORLEVEL%
```

`.cmd` extension, `@echo off` so nothing corrupts the ACP stdio stream, PATH prefix pointing at the managed native dir, `DISABLE_AUTOUPDATER=1` from the registry's self-update neutralization, and `exit /b %ERRORLEVEL%`. Mixed `\` / `/` separators inside the quoted exec path; cmd.exe accepts it.

### Layer 4 — the native `.exe`

```
role=NativeCli source=pinned_binary version=Some("2.1.234") path=...\agents\claude\native\claude.exe
native CLI installed at ...\agents\claude\native\claude.exe (exists=true, extension=Some("exe"))
--- managed native dir
  file 324028576  claude.exe
```

324MB downloaded and sha256-verified against the `windows_x64` pin, landing with the `.exe` name.

### Layer 5 — spawn

Spawned the launcher the way `live::sessions::driver::process::spawn_agent_process` does (launcher path as program, all three stdio handles piped), then closed stdin:

```
spawned pid 3652    (claude)   --- launcher status: Some(ExitStatus(ExitStatus(0))) (still_running=false)
spawned pid 1856    (grok)     --- launcher status: Some(ExitStatus(ExitStatus(0))) (still_running=false)
```

Exit 0 is load-bearing here: cmd.exe returns 9009 for "not recognized as an internal or external command", and the launcher propagates `%ERRORLEVEL%`. A 0 means the `.cmd` shim was found and ran.

### Honest limits of this result

- Both launchers exited 0 with **empty stdout**. That proves the process started, ran, and shut down cleanly on stdin EOF. It does **not** prove an ACP `initialize` handshake — no JSON-RPC was exchanged.
- Only the npm and git agent-process sources are covered. The **archive** source is not, and `cursor` and `opencode` have no `windows_*` target in their archive pins at all (`linux_x64`, `macos_arm64`, `macos_x64` only), so both would fail `NoPinForPlatform` on Windows today.
- No desktop app, no code signing, no auth, no LLM call. Runtime library only.

### Commits combined

- `origin/main` @ `7a4f9d66e8`
- `#2149` `fix/pinned-installer-windows-exe` (`.exe` managed-binary naming, windows pins in the catalog)
- `#2152` `fix/windows-managed-launcher` @ **`f3122f08fc`** (`.cmd` launcher, `npm.cmd`, `.bin` `.cmd` shim, always-quoted batch args)

Two merge conflicts, both import collisions where #2149 moved `platform_binary_filename` into `executable.rs` while `f3122f08fc` added neighbours to it in `npm.rs`; both sides kept, the duplicate definition dropped.

pablonyx (commented):
## Verdict

**The rebuild is sound and the experiment holds.** I verified the reconstruction independently rather than taking the resolution notes at face value, and the strongest single result is this: every path this PR actually changes is *byte-identical* between the proven pre-rebuild tip `77055343d7` and the new tip `db843f501d`. The rebuild did not just produce an equivalent branch, it produced the same content.

CI at `db843f501d` reproduces the proof on both legs. There is **one genuine red** (`repo-shape`, inherited from #2152, not caused by the rebuild) and **five inaccuracies in the PR body** that matter more than usual because this PR's whole purpose is to be quotable as evidence.

No blocking items in the test or the merge resolution. Blocking items are in the body and the size lint.

---

## 1. Rebuild verification — passes, and more cleanly than claimed

```
git diff 77055343d7 db843f501d -- \
  .../domains/agents/installer .../domains/agents/readiness .../domains/agents/model.rs \
  .../integrations/agent_cli .../anyharness-lib/tests \
  catalogs/agents scripts/agent-catalog .github/workflows/windows-*.yml \
  specs/FEATURE_DOCS/platforms/product/agent-distribution.md
→ (empty)
```

Everything else in the old-vs-new diff is `main` advancing. So the content that produced run `32446919674` is exactly the content at the new head.

**Test file** — byte-identical, confirmed by blob hash rather than by diff:
`anyharness/crates/anyharness-lib/tests/windows_managed_agent_install.rs` is blob `a0255963f86d0b070b7f076166121bc8a57fa30c` at *both* `77055343d7` and `db843f501d`.

**The `npm.rs` / `agent_process.rs` conflict** — resolution is correct, complete, and is the *union* of both sides rather than a pick:

- The definition deleted from `npm.rs` (lines 484-491 @ `24d6299510`) was a strict **subset** of `executable.rs::platform_binary_filename`. #2152's copy was an unconditional `format!("{binary_name}.exe")`; the surviving `executable.rs:8` version adds the `.exe.exe` double-suffix guard (`!binary_name.to_ascii_lowercase().ends_with(".exe")`). Nothing was lost and one guard was gained.
- No dangling references. At the head, all six `platform_binary_filename` call sites resolve through `crate::integrations::agent_cli::executable`; none through `super::npm`. `platform_npm_bin_relpath` stays `pub(super)` in `npm.rs` and is legitimately reachable from the sibling `installer::agent_process` module.
- **The one that could have silently gone wrong and didn't.** `installer/agent_process.rs:94` kept #2149's `managed_pinned_binary_path(runtime_home, kind, &ArtifactRole::NativeCli)` rather than #2152's `managed_native_dir.join(kind.as_str())`. Had #2152's side won here, on Windows `is_valid_executable` (bare `path.is_file()` there) would have been handed `…\native\claude` instead of `claude.exe`, returned false, and quietly dropped the native CLI directory from the launcher's PATH prefixes — a Windows-only regression that no test in this PR would have caught, since the launcher would still spawn and still exit 0. Both sides survived.

`readiness/artifacts.rs` and `agent_process.rs` are both supersets of each parent; `launcher.rs` is identical to the #2152 side (#2149 never touched it). `Cargo check & test` is green at the head, which independently confirms the resolution compiles and that the new `windows_npm_resolution_tests` and launcher batch-shape unit tests pass.

**The skipped cherry-pick** — nothing was lost, and it was already empty upstream. `d750ece8db` has a zero-byte diff against its parent (`git diff --numstat d750ece8db^ d750ece8db` → empty); its content had already been absorbed into the preceding merge `d4420ab102`. On the rebuilt branch the same content arrives inside `db843f501d` (108 insertions rather than `77055343d7`'s 57, because it replays against a base that lacks the npm-spawn test). Final blob identical, so the boundary shift is cosmetic.

## 2. CI at `db843f501d`

**Windows managed-agent install + spawn — run [32450488248](https://github.com/proliferate-ai/proliferate/actions/runs/32450488248): both legs green.** Reproduced at the new SHA, not inherited:

```
windows-latest:  test result: ok. 3 passed; 0 failed  (418.05s)
ubuntu-latest:   test result: ok. 3 passed; 0 failed  (79.91s)
```

The Windows log carries the substance:

```
role=NativeCli    …\agents\claude\native\claude.exe          (extension=Some("exe"))
role=AgentProcess …\agents\claude\agent_process\claude-launcher.cmd
--- claude-launcher.cmd
@echo off
"…\agent_process\node_modules/.bin/claude-agent-acp.cmd" %*
exit /b %ERRORLEVEL%
```

**One red: `Repo shape checks` → `ci-ok`** (run [32450488265](https://github.com/proliferate-ai/proliferate/actions/runs/32450488265)).

```
anyharness/.../installer/npm.rs: PROD-SIZE-1 — found: 628 lines (max 600)
```

`lints/anyharness/ratchets.toml` has no `npm.rs` entry at the head. Sizes: `main` 526 → **#2152 tip 636** → this branch 628. So **#2152 introduces this on its own and will fail `repo-shape` on its own PR too** — the dedupe performed here shaved 8 lines but the file needs 28 more. Not a rebuild artifact: the pre-rebuild tip failed identically (run `32446919652`). Worth flagging on #2152 now rather than discovering it at merge time.

Two things to know before reading this PR's check list as a verdict:

- **`Windows compile check` is `continue-on-error: true` on both jobs.** Its green is unconditional and carries no information. It is a probe by design (the file says so), but the check mark looks the same as a real one.
- **`ci-ok`'s `needs:` list covers only `ci.yml` jobs**, so the Windows managed-agent workflow sits outside the rollup gate entirely. Its result has to be read directly.

## 3. The test itself — genuine, and the negative control settles it

Nothing stubbed, mocked, skipped or `continue-on-error`'d, and I checked rather than trusted the header: the test commits touch **only** the workflow and the test file. `git diff --numstat cfdac10b93^ db843f501d` is exactly two files, 418 insertions, **zero production lines**. No visibility was widened to let the test reach in — it runs entirely against the already-public surface.

The chain is real end to end: real bundled catalog resolving claude's sha256'd 324MB `win32-x64/claude.exe` pin, a real `npm install git+https://…#65f62053…`, a real pinned-binary download, and a real `Command::new(launcher)`.

**The negative control is decisive.** At `320b4d16a1`, before the `npm.cmd` fix, run [32445803852](https://github.com/proliferate-ai/proliferate/actions/runs/32445803852) was ubuntu green / windows red with *both* the isolated layer-1 test and the full install path failing:

```
`Command::new("npm")` could not be spawned: program not found (kind NotFound…)
install_agent_with_pins failed: io error: program not found (kind Other)
```

That answers both questions at once: the Ubuntu leg genuinely discriminates rather than passing vacuously, and the deep assertion fires on a real install-path regression rather than only the shallow one. Better still, the control is re-run *inside every passing run* — the Windows log at the current head records `control: bare npm --version -> NOT SPAWNABLE` alongside `npm.cmd --version -> status 0`, so the fix's necessity is re-demonstrated on every execution rather than asserted once historically.

**The exit-0 assertion is load-bearing on Windows**, which is not obvious and I verified: `build_batch_launcher_script` terminates with `exit /b %ERRORLEVEL%`, and cmd.exe returns 9009 for an unresolvable command, so `status.success()` would catch a shim-resolution regression rather than sailing past it. It is not a decorative assertion.

Where the test is weaker than its header implies — none blocking, all worth knowing:

- **A hang passes.** Lines 338-343 assert only inside `if let Some(status)`. If the launcher never exits within the 60s deadline, the test asserts nothing beyond `spawn()` returning `Ok`. Defensible — a live ACP server outliving stdin EOF is a legitimate outcome — but "launcher hangs forever" is an accepted result, not a failure.
- **Layer 3 is printed, not asserted.** The header says what npm writes into `.bin` on Windows "is observed here"; it is `list_dir`'d to stdout and never asserted. If npm's cmd-shim stopped writing the `.cmd` sibling, nothing in the test fails directly — you would only catch it downstream through the spawn.
- **Stale SHAs in the header.** Lines 16 and 21 cite `#2152 @ f3122f08fc`, which no longer exists on this branch (rewritten to `24d6299510`). **Do not fix this here** — the file's current value is being byte-identical to the version that passed. Fix it when #2152 lands.

One inherited observation for #2152, not for this PR: the batch launcher invokes its target without `call`. When the target is a `.cmd` — which is exactly what `platform_npm_bin_relpath` selects on Windows — cmd.exe transfers control rather than returning, so the `exit /b %ERRORLEVEL%` line that the doc comment calls load-bearing is in fact dead on that path. The exit code still propagates (the inner script terminates the session with its own code), so behavior is correct today and the test is unaffected; but the stated rationale is wrong, and any line added after the call would be silently skipped for `.cmd` targets.

## 4. What was proven, and what was not — the body needs to say this

This is the part I would not let the PR ship without, because the body is what people will quote.

**Proven on `windows-latest`:** the managed install path runs; `npm.cmd` resolution works; the `.exe` native binary lands with its extension; the `.cmd` launcher is emitted, is spawnable by `CreateProcess`, and exits 0.

**Not proven:**

1. **No ACP handshake.** Both launchers exit 0 with empty stdout *and* empty stderr. That proves process spawn and clean exit on stdin EOF. It does not prove an `initialize` exchange — nothing was written to stdin and nothing was read back, so "ACP server started, saw EOF, shut down cleanly" and "process started and immediately exited without speaking ACP" are indistinguishable from this evidence.
2. **Three of five agents are uncovered on Windows.** `cursor` and `opencode` have no `windows_*` pins at all (`catalogs/agents/catalog.json` gives them only `linux_x64`, `macos_arm64`, `macos_x64`), so they fail `NoPinForPlatform` and cannot reach the path this test exercises. `codex` *does* have Windows pins, but they are `codex-*-pc-windows-msvc.exe.tar.gz` — the tar-extraction branch, which this test never touches.

**Blocking items, all in the PR body** (`--body-file`, no code change needed):

- **B1.** "The agent under test is `grok`" — stale. `claude` is also under test, and it is the more important case: it is the default agent and the only one exercising both the pinned native `.exe` download and the git-sourced npm install.
- **B2.** "one commit adding the test + workflow" — there are three test commits, one of which is empty.
- **B3.** "The one merge conflict was an import collision … both are kept" — no longer describes this branch. The current head required deleting a duplicate `platform_binary_filename` from `npm.rs` and re-pointing `agent_process.rs`'s import at `executable.rs`. The body understates the hand resolution, which is the opposite of what a proof PR should do.
- **B4.** The Composition section describes `origin/main` + two merges and cites no SHAs. Pin the actual shape: rebased #2149 tip `b4e673f817` + merge of rebased #2152 `24d6299510` + test commits, head `db843f501d`. Anyone re-deriving this result needs the SHAs.
- **B5.** Neither limit in §4 above appears anywhere in the body. "What is under test, all at once" currently reads as a complete Windows-readiness claim. Add the handshake limit and the three-agent coverage gap verbatim.

## Summary

The reconstruction is trustworthy — I checked it against the proven tip by content, not by narrative, and it matches exactly. The test is real, the Ubuntu leg is a true control, and the negative control proves the assertions bite. Fix the body (B1-B5) so the PR claims only what it proved, and take the `npm.rs` 628-line PROD-SIZE-1 finding back to #2152 where it originates.

Not for merge as-is — but the body says that already, and on the evidence it is correct to say so.

---
*Reviewer of record. No push, merge, relabel, or ready-for-review performed.*

pablonyx:
Heads-up from the #2152 lane, since this branch composes it.

#2152 has moved to `24270510c8`. It clears the `Repo shape checks` red that this branch inherits, by extracting `npm_program_name*` and `platform_npm_bin_relpath*` plus their four tests out of `installer/npm.rs` into a new `installer/npm_platform.rs` (636 lines to 535, `check_max_lines.py` now passes). Ratcheting was rejected: `lints/anyharness/ratchets.toml` states in its own header that it is a ratchet and not an exception ledger, and that entries may only shrink.

Two things that matter for this branch specifically:

1. **The merge resolution this PR proved is still the correct one.** `npm.rs` re-exports `platform_npm_bin_relpath` as `pub(super)`, so `agent_process.rs`'s `use super::npm::{platform_binary_filename, platform_npm_bin_relpath};` resolves unchanged. Nothing about the resolution shape verified here is invalidated.

2. **This branch is deliberately NOT being rebuilt onto that.** Its evidentiary value is that every changed path is byte-identical to the pre-rebase proven tip `77055343d7`, and rebuilding would destroy that property to chase a line count. The Windows proof (run 32450488248, `windows-latest` 3 passed / 418s) stands on its own.

The move is a pure relocation, proven by mutation control rather than by inspection: mutating `npm_program_name_windows()` to `"npm"` and dropping the `.cmd` suffix gives `2 passed; 2 failed`, restoring gives `4 passed; 0 failed`.

PR #2156, opened 2026-08-21
fix(release): push version bump with bypass-capable token

pablonyx:
## Problem

The first real dispatch of release.yml (run 32446272831) failed in `prepare` step "Commit version bumps":

```
remote: error: GH006: Protected branch update failed for refs/heads/main.
remote: 8 of 8 required status checks are expected.
```

`git push origin HEAD:main` authenticates with the checkout's persisted GITHUB_TOKEN (github-actions[bot]), which cannot satisfy the required checks that branch protection now demands on a direct push. The dry run cannot catch this: it short-circuits before the push by design. Every release, including the daily 09:00 UTC cron, fails at this step until fixed. Fail-closed held: everything downstream skipped, zero side effects.

## Fix

The version-bump push is the only push to a branch of this repository in the release pipeline (release-desktop.yml also pushes a branch, but to the external Homebrew tap repo with its own token; everything else is tag pushes, which no protection rule covers). This PR authenticates that one command with a new `RELEASE_PUSH_TOKEN` secret, an admin fine-grained PAT (contents: read and write on this repo) that bypasses required checks while `enforce_admins` stays off. Everything else in the workflow keeps using GITHUB_TOKEN.

Mechanics:
- Guard first: if the secret is unset, fail with an explicit `::error` before committing, rather than dying opaquely at the push. The guard sits after the dry-run/skip-build/no-diff early exits, so those paths never require the secret.
- actions/checkout persists GITHUB_TOKEN as an `http.https://github.com/.extraheader` in `.git/config`, and that header would override URL basic-auth credentials. The push blanks it with `-c` for that one command so the PAT in the URL is the credential that reaches GitHub.
- The secret value is masked in logs by Actions secret masking, and git anonymizes URL userinfo in transport errors.
- `scripts/ci-cd/deploy-staging-trigger.test.mjs` updated: the dry-run-precedes-push assertion now anchors on the new push literal and asserts both indices are found, so it can no longer degrade to a silent pass if either line disappears.
- `guides/deploying/releases.md` documents the secret, its scope, and the failure mode when it is missing or expired.

## Behavior change worth knowing

GITHUB_TOKEN pushes do not trigger workflows, so historical bump commits got no CI run on main. PAT pushes do trigger them: each bump commit will now run ci.yml, codeql.yml, and self-host-smoke.yml on main (server-ci.yml and publish-agent-catalog.yml are path-filtered away from the bump files). None of those push or deploy, so there is no loop, but every release now also pays for a main CI + CodeQL + smoke run.

## Operational prerequisite

`RELEASE_PUSH_TOKEN` must exist as a repo-level secret before the next dispatch or cron (the prepare job has no `environment:`). The PAT owner must be a repo admin, or the push fails with the identical GH006 that the guard cannot detect. Known residual risks, accepted for tonight and queued as follow-ups: a long-lived admin PAT in a repo secret is readable by same-repo-branch workflow runs, and an expired PAT presents as set-but-rejected, which the `-z` guard cannot catch. Durable alternatives (GitHub App installation token, or a ruleset bypass actor) are in the follow-up queue.

## Verification

- `ruby -ryaml` parse: clean.
- The failing run is the negative control for the workflow change: identical push, bot credential, GH006. This PR changes only the credential path for that push.
- Test negative control: the updated assertion run against main's release.yml fails on "must push to main with RELEASE_PUSH_TOKEN"; against this branch, 6/6 pass.
- Guard path: with the secret unset the step exits 1 at the `::error` line before `git commit`, leaving nothing to clean up.
- End-to-end proof is the next real dispatch after `RELEASE_PUSH_TOKEN` lands in repo secrets (dry run structurally cannot exercise the push).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -206,6 +206,7 @@ jobs:
           RELEASE_ID: ${{ steps.meta.outputs.release_id }}
           SKIP_BUILD: ${{ steps.meta.outputs.skip_build }}
           DRY_RUN: ${{ steps.meta.outputs.dry_run }}
+          RELEASE_PUSH_TOKEN: ${{ secrets.RELEASE_PUSH_TOKEN }}
         run: |
           set -euo pipefail
           # A dry run computes the bump so the plan is real, then throws it away:
@@ -218,11 +219,26 @@ jobs:
             echo "head_sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
             exit 0
           fi
+          # main is branch protected and the workflow's GITHUB_TOKEN cannot
+          # satisfy required status checks on a direct push (GH006, run
+          # 32446272831). This is the only branch push in the release
+          # pipeline, so it alone authenticates with RELEASE_PUSH_TOKEN, an
+          # admin fine-grained PAT that bypasses the checks while
+          # enforce_admins stays off. Fail loudly before committing if the
+          # secret is missing instead of dying opaquely at the push.
+          if [[ -z "$RELEASE_PUSH_TOKEN" ]]; then
+            echo "::error::RELEASE_PUSH_TOKEN secret is not set; the version-bump push to protected main needs a bypass-capable token."
+            exit 1
+          fi
           git config user.name "github-actions[bot]"
           git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
           git add VERSION apps/desktop/package.json apps/desktop/src-tauri/tauri.conf.json apps/desktop/src-tauri/Cargo.toml anyharness/sdk/package.json
           git commit -m "release: prepare $RELEASE_ID"
-          git push origin HEAD:main
+          # actions/checkout persists GITHUB_TOKEN as an http.extraheader in
+          # .git/config; blank it for this one command so the PAT in the URL
+          # is the credential that reaches GitHub.
+          git -c "http.https://github.com/.extraheader=" \
+            push "https://x-access-token:${RELEASE_PUSH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" HEAD:main
           echo "head_sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
 
       - name: Create release tags
--- a/guides/deploying/releases.md
+++ b/guides/deploying/releases.md
@@ -25,6 +25,8 @@ selects an exact set. Its prepare job resolves the train, public product, and
 artifact versions; may commit version bumps to `main`; and creates the selected
 train/product/artifact tags.
 
+The version-bump push to `main` authenticates with the `RELEASE_PUSH_TOKEN` repository secret, an admin fine-grained PAT (Contents: Read and write on this repository) that bypasses the required status checks on `main`. The prepare job fails loudly before committing if the secret is missing, and every release including the daily cron fails until it is restored, so renew the PAT before it expires.
+
 The train then:
 
 1. releases selected Runtime/SDK, Server/self-host, and Desktop artifacts;
--- a/scripts/ci-cd/deploy-staging-trigger.test.mjs
+++ b/scripts/ci-cd/deploy-staging-trigger.test.mjs
@@ -106,10 +106,13 @@ test("a dry run walks the graph without any externally visible effect", () => {
   // Nothing is pushed to main and no tag is minted.
   const commit = section("- name: Commit version bumps", "- name: Create release tags");
   assert.match(commit, /if \[\[ "\$SKIP_BUILD" == "true" \|\| "\$DRY_RUN" == "true" \]\]; then/);
-  assert.ok(
-    commit.indexOf('exit 0') < commit.indexOf("git push origin HEAD:main"),
-    "the dry-run early return must precede the push to main",
+  const earlyReturn = commit.indexOf("exit 0");
+  const pushToMain = commit.indexOf(
+    'push "https://x-access-token:${RELEASE_PUSH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" HEAD:main',
   );
+  assert.notEqual(earlyReturn, -1, "the commit step must keep its dry-run early return");
+  assert.notEqual(pushToMain, -1, "the commit step must push to main with RELEASE_PUSH_TOKEN");
+  assert.ok(earlyReturn < pushToMain, "the dry-run early return must precede the push to main");
   for (const step of ["- name: Verify version tags are free", "- name: Create release tags"]) {
     assert.match(
       section(step, "\n      - name:"),


pablonyx (commented):
**Reviewer verdict: REQUEST CHANGES**

Reviewed adversarially as the dedicated reviewer of record. The credential mechanics are correct and I verified them against upstream source rather than taking the PR body's word for it. One blocker: the change breaks a required check, and the invariant that check exists to protect is now silently unenforced.

---

## Blocker

**F1 (HIGH) — a workflow-shape test still asserts the old push string; `ci-ok` goes red and the dry-run invariant is lost.**
`scripts/ci-cd/deploy-staging-trigger.test.mjs:110`

```js
commit.indexOf('exit 0') < commit.indexOf("git push origin HEAD:main"),
```

That literal no longer exists in the "Commit version bumps" section, so the right-hand `indexOf` returns `-1`, the comparison is `<positive> < -1` → false, and the assertion throws. Reproduced locally (`node --test` → `not ok 6 - a dry run walks the graph without any externally visible effect`, `'the dry-run early return must precede the push to main'`) and in CI: job 96669164832 ("CI/CD config checks") is red on exactly that assertion. That job is `ci-cd-config` (`.github/workflows/ci.yml:162`), which sits in `ci-ok`'s `needs` (`.github/workflows/ci.yml:821`), and `ci-ok` is one of the 8 required contexts on `main`. So this cannot merge as-is.

Worth fixing properly rather than just re-pointing the string: the assertion degrades silently to a *pass* whenever the early-return `exit 0` is what goes missing, and to a confusing *fail* when the push string moves. Assert the index was found:

```js
const pushIdx = commit.indexOf('push "https://x-access-token:');
assert.notEqual(pushIdx, -1, "release.yml must still push the bump to main");
assert.ok(commit.indexOf("exit 0") < pushIdx, "the dry-run early return must precede the push to main");
```

---

## Non-blocking

**F2 (MEDIUM) — the secret does not exist yet; merge order matters.** `gh api repos/proliferate-ai/proliferate/actions/secrets` lists 19 secrets and `RELEASE_PUSH_TOKEN` is not among them. Merging before it lands leaves the 09:00 UTC cron failing — at the guard instead of the push, so a better error but the same broken release. `prepare` has no `environment:`, so it must be a repo-level secret. Also worth an explicit pre-merge check that the PAT's **owner has admin on this repo**: a fine-grained PAT with `contents: write` but a non-admin owner fails with the identical GH006, and the guard cannot detect that.

**F3 (MEDIUM, posture) — a long-lived admin bypass credential as a plain repo secret.** Repo secrets are available to every run from a same-repo branch, including `pull_request` events on internal branches, so anyone who can push a branch here can exfiltrate a token that bypasses branch protection. Actions masking is log hygiene, not a boundary. Two lower-privilege shapes worth weighing before this becomes permanent: mint a short-lived GitHub App installation token per run, or move `main` to a ruleset and add `github-actions[bot]` as a bypass actor, dropping the extra credential entirely. Second-order: fine-grained PATs expire. When it does, the secret is *present but rejected*, the `-z` guard does not fire, and the release breaks again at the push.

**F4 (LOW) — runbook not updated.** `guides/deploying/releases.md:25` describes the version-bump commit but the guide never names the new hard dependency. One line with the secret name, required permission, owner, and expiry date keeps it truthful.

**F5 (LOW, PR-body accuracy) — "the single branch push in the whole release pipeline" is not literally true.** `.github/workflows/release-desktop.yml:919` (`git push origin HEAD` inside the cloned Homebrew tap) is also a branch push made by this pipeline; it targets `HOMEBREW_TAP_REPO` with `HOMEBREW_TAP_TOKEN`, so this repo's protection does not apply. The accurate claim is "the only branch push *to this repository*". I confirmed the rest: everything else is a tag push (`release-desktop.yml:699`, `release-runtime.yml:233`, `server-ci.yml:767` — gated on `inputs.publish`, so unreachable from a plain main push, and `scripts/ci-cd/create-release-tags.mjs:218,223`), `rulesets` is `[]`, and there is no tag protection, so those stay correct on GITHUB_TOKEN.

**F6 (LOW, PR-body accuracy) — "no loop risk: ci.yml never pushes" is true but under-enumerates.** Workflows with `on.push.branches: [main]` are ci.yml, codeql.yml, self-host-smoke.yml, plus server-ci.yml and publish-agent-catalog.yml (both path-filtered, and the bump touches `VERSION`, `apps/desktop/**`, `anyharness/sdk/package.json` — no match, so neither runs). None of the three that do run pushes or deploys, so the conclusion holds. But the honest statement is that every release will now additionally spend runner minutes on ci.yml + CodeQL + the self-host compose smoke, and a red main CI on a bump commit is a new (cosmetic) failure surface.

**F7 (NIT) — token in argv.** The PAT is visible in the `git push` process command line on the runner. Acceptable on an ephemeral single-tenant runner, and the URL form matches existing repo convention (`release-desktop.yml:895`), so I would keep it. Noting only for completeness.

---

## What I verified (so it is not re-litigated)

1. **Checkout mechanics — confirmed at the pinned SHA.** `actions/checkout@3d3c42e5aac5` `src/git-auth-helper.ts:57` sets `tokenConfigKey = http.${serverUrl.origin}/.extraheader` → exactly `http.https://github.com/.extraheader`, with value `AUTHORIZATION: basic base64(x-access-token:<token>)` (:59-64), written whenever `persistCredentials` is on (`action.yml:52-54`, default `true`). The key the PR blanks matches byte-for-byte. `x-access-token:<PAT>` as the basic-auth pair is exactly what checkout itself does for user-supplied PATs, so the username form is proven against GitHub.
2. **The blanking is necessary, not cargo-cult.** The extraheader is an explicit `Authorization:` request header, which replaces curl's own USERPWD-generated one — it would win over the URL credential.
3. **Empty value = reset is documented git behavior.** `git-config(5)`, `http.extraHeader`: "To allow overriding the settings inherited from the system config, an empty value will reset the extra headers to the empty list" (verified in the local man page, git 2.50.1). Command-line `-c` is parsed after repo config so the reset wins, and URL config matching ignores userinfo when the pattern carries none — which is both why the header would have applied to the credentialed push URL and why the reset covers it.
4. **Secret exposure is adequately handled.** No `set -x`; Actions echoes the unexpanded script source (visible in the failed run's `##[group]Run` block, so `${RELEASE_PUSH_TOKEN}` stays literal); Actions masks the registered value anywhere it lands. Independently, git anonymizes URL userinfo in transport errors — I reproduced `git push https://x-access-token:SUPERSEKRIT@127.0.0.1:9/foo/bar.git` → `fatal: unable to access 'https://127.0.0.1:9/foo/bar.git/'`, credential stripped. `GITHUB_REPOSITORY` is a default env var. Nothing is persisted to `.git/config` (one-shot URL, not `remote set-url`), which is better than the `set-url` alternative.
5. **Guard placement and polarity are correct.** `release.yml:214-217` exits on skip_build/dry_run, `:218-221` exits on no-diff, guard at `:229-232` sits after both — none of those paths require the secret. `env: X: ${{ secrets.X }}` with an unset secret exports the empty string, so `set -u` does not trip and `[[ -z ]]` fires. `secrets.*` is valid here (this is release.yml itself, not a reusable-workflow context). `ruby -ryaml` parse clean; the step's env block resolves as expected.
6. **The fix's premise holds.** Branch protection on `main`: exactly 8 required contexts, `enforce_admins.enabled: false`, no `required_pull_request_reviews`, `rulesets: []`. Required status checks are the only blocker, and an admin actor bypasses them with `enforce_admins` off. Matches run 32446272831 verbatim ("8 of 8 required status checks are expected", commit `7ebb33539` created then rejected, downstream all skipped — fail-closed claim confirmed).

## Could not verify

- That the actual PAT is admin-owned, correctly scoped, and bypasses — the secret does not exist yet and token metadata is not inspectable from here. This is the PR's own stated "end-to-end proof is the next real dispatch"; see F2.
- Whether git's `To <url>` / `[remote rejected]` line anonymizes userinfo the way the `fatal: unable to access` line does — I had no way to produce an authenticated-but-rejected push. Actions masking covers this path regardless, so it does not change the verdict.

Fix F1 and this is a clean approve from my side.

pablonyx (commented):
**Reviewer verdict: APPROVE** (round 2, re-verified at `1e7b7a7bb17cc9160ba4421fbef8e77d9bff0639`)

Re-reviewed the delta `5dea939..1e7b7a7` independently rather than taking the round-2 summary at face value. The blocker is genuinely fixed, the two documentation/accuracy findings are addressed, and `release.yml` itself is byte-identical to what I already cleared in round 1 (round 2 touches only the test and the guide — `git diff 5dea939 1e7b7a7` shows no workflow change), so the credential mechanics I verified against upstream `actions/checkout` and `git-config(5)` still stand unchanged.

## F1 (was the blocker) — RESOLVED

`scripts/ci-cd/deploy-staging-trigger.test.mjs:109-115` now anchors on the literal push invocation, asserts both indices are found before comparing, and keeps the ordering assertion:

```js
const earlyReturn = commit.indexOf("exit 0");
const pushToMain = commit.indexOf('push "https://x-access-token:${RELEASE_PUSH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" HEAD:main');
assert.notEqual(earlyReturn, -1, ...);
assert.notEqual(pushToMain, -1, "the commit step must push to main with RELEASE_PUSH_TOKEN");
assert.ok(earlyReturn < pushToMain, ...);
```

This is exactly the silent-degrade hardening I asked for, and the single-quoted JS string keeps `${...}` literal so it matches the YAML source text. `indexOf("exit 0")` still lands on the dry-run/skip-build branch (the first of the two early returns), so the invariant under test is unchanged.

I ran both controls myself rather than trusting the reported ones:

- **Positive:** `node --test scripts/ci-cd/deploy-staging-trigger.test.mjs` at `1e7b7a7` → `# tests 6 / # pass 6 / # fail 0`.
- **Negative:** same test with `.github/workflows/release.yml` swapped back to main's version (`git checkout 9562692 -- ...`) → `not ok 6`, `error: 'the commit step must push to main with RELEASE_PUSH_TOKEN'`. The assertion fails for the right reason on the right line, which is what makes it a real guard rather than a tautology. Worktree restored clean afterward.

**CI on the exact commit:** run `32447865310` reports `head_sha: 1e7b7a7bb17cc9160ba4421fbef8e77d9bff0639`, `conclusion: success`. `CI/CD config checks` (`ci-cd-config`) passes in 19s, and the `ci-ok` rollup that consumes it passes too — so the required context that was red in round 1 is green on this head, not on a stale one. Full PR check state: 35 pass, 3 skipping, 0 fail, 0 pending.

## F4 — RESOLVED

`guides/deploying/releases.md:28` adds a paragraph immediately after the version-bump sentence, covering the secret name, scope (Contents: Read and write), the admin requirement, the bypass semantics, the fail-loud-before-committing behavior, the fact that every release including the cron stays broken until it is restored, and the renewal prompt. I checked each of those claims against the workflow and the branch-protection API response; all accurate. Placement is right: that section's prose is what describes `release.yml`'s prepare job.

(Pre-existing, not yours: the section heading is still `## Nightly Release Train` at line 20 even though the workflow is now named `Release` and its own header comment says the nightly train was retired. Untouched by this PR — flagging only so it does not get attributed to this change later.)

## F5 / F6 — RESOLVED

PR body now reads "the only push to a branch of this repository," explicitly calls out the `release-desktop.yml` Homebrew-tap branch push as external-with-its-own-token, and notes everything else is tag pushes uncovered by any protection rule — which matches what I found (`rulesets: []`, no tag protection). The behavior-change section now enumerates ci.yml, codeql.yml, and self-host-smoke.yml as newly triggered, with server-ci.yml and publish-agent-catalog.yml correctly identified as path-filtered away from the bump files, and states the runner-cost consequence rather than only the no-loop conclusion. Both claims now match the repository.

## F2 / F3 — acknowledged, still open by design

No code change expected and none needed. Recording state for the trail: `RELEASE_PUSH_TOKEN` is **still absent** from repo secrets as of this review (`gh api repos/proliferate-ai/proliferate/actions/secrets` → not present). The PR body's new "Operational prerequisite" section states the merge-order requirement, the admin-owner requirement, and both residual risks honestly, including that an expired PAT presents as set-but-rejected and slips past the `-z` guard. That is the right disposition.

Two things to hold onto operationally, neither blocking:

1. **Do not merge ahead of the secret** — or if you do, know that the next cron fails at the guard rather than at the push. Same outcome, better error.
2. **First real dispatch is the only proof.** Confirm the PAT owner has repo admin before that dispatch; a non-admin owner with `contents: write` fails with the identical GH006, and nothing in this PR can distinguish that case.

## Residual nit (non-blocking, no action needed)

The test now pins the exact push URL form. If the F3 follow-up lands (App installation token, or a ruleset bypass actor that lets the push go back to plain `origin`), this assertion breaks — but it breaks *loudly*, with `"the commit step must push to main with RELEASE_PUSH_TOKEN"`, which points whoever does that work straight at the line to update. That is the correct failure mode for a shape test, so I would leave it exactly as written.

Everything I raised in round 1 that was in scope for this PR is addressed, and I verified each fix against the commit rather than the summary. Approving.

PR #2157, opened 2026-08-21
fix(build): normalise line endings so check-theme survives a Windows checkout

pablonyx:
## What this is

Two commits, one concern: a checkout on Windows silently corrupts `apps/packages/design/src/css/product.css` into CRLF, and `check-theme.mjs` asserts against `\n`-joined literals, so the design gate fails with a message that blames the CSS.

This surfaced while getting the first real Windows `tauri build` to run. It is not Windows-specific in effect: the repo has no `.gitattributes` at all, so line-ending behaviour is left to each contributor's `core.autocrlf`.

- `917b77705d` `.gitattributes` with `* text=auto eol=lf` (13 lines). Fixes the checkout.
- `e19c87f67e` `check-theme.mjs` reads through a `readText()` helper that normalises CRLF. Fixes the working tree, which `.gitattributes` does not.

## Renormalisation is provably a no-op

Not asserted, measured, over all 8065 tracked files at the PR head:

- `git ls-files --eol`: 7852 `i/lf`, 111 `i/-text`, 102 `i/none`. **Zero `crlf`, zero `mixed`.**
- `git grep -I -l -P '\r' HEAD` returns **0 files**.
- Zero tracked `.bat`, `.cmd`, `.ps1`, `.sln`, `.vcxproj`, `.reg`, `.rc`, `.nsi`, `.wxs`, `.iss` files, so nothing in-tree needs CRLF preserved.

(An earlier revision of this body cited 7960 / 7774 / 84. Those numbers were wrong; the counts above are the measured ones and were independently reproduced during review. The conclusion is unchanged, since what matters is that `crlf`, `mixed`, and literal-CR counts are all zero.)

## Generated output is read raw

Review found that normalising `dist/theme.css` was a no-op that only loosened a check, and it has been reverted in `9eaee53626`. That file is untracked, gitignored at `.gitignore:14`, and written by `generate-theme.mjs` from an in-memory string (`writeFile` at `:163`, no `readFile` anywhere in that script), so a git checkout can never hand it CRLF. The script exists to assert the generated CSS is byte-identical to the committed generator output, and normalising that read weakened exactly that.

`readText` now applies only to the two files that actually come out of git, `src/css/product.css` and `src/tokens.ts`. Reverified after the revert: LF tree `check-theme: ok (309 live tokens)`, CRLF tree `check-theme: ok (309 live tokens)`.

## Negative control

Run locally against a prebuilt `dist/`, three cells:

| working tree | script | result |
|---|---|---|
| CRLF | unpatched | `check-theme: product.css lost the exact motion declaration: background-color var(--duration-hover) ...` |
| CRLF | patched | `check-theme: ok (309 live tokens)` |
| LF | patched | `check-theme: ok (309 live tokens)` |

The failing string is character-identical to the one CI run `32445871270` produced on Windows, so the local reproduction is the same defect, not an analogue.

## Dropped from this PR

An earlier commit (`9924a7da79`) added an opt-in `workflow_dispatch` Sentry bypass to `release-desktop.yml`, so Windows packaging could be verified while `SENTRY_AUTH_TOKEN` is broken. **It has been dropped.** Review found it puts a permanent manually-dispatchable "ship without sourcemaps" switch on the release surface with no CI guard keeping it out of `workflow_call`, and its shape was wrong regardless: it unset the auth vars but left `VITE_PROLIFERATE_SENTRY_DSN`, so a build made that way would still report to Sentry with `sourcemap: false` — silent unsymbolicatable noise rather than a loud absence.

Packaging verification therefore depends on fixing `SENTRY_AUTH_TOKEN`, not on bypassing it. That token was rotated at `2026-08-20T19:51:01Z` and the new one cannot read `proliferate-desktop`; this is org-wide and would break a macOS release dispatched today. Tracked in #2150.

## Branch name

The branch is still `diag/windows-installer-packaging` from when this carried the bypass. Renaming it would close the PR, so it stays; the title and contents are the accurate description.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,13 @@
+# Force LF line endings for text files on checkout, regardless of the
+# checking-out platform's git config. Without this, GitHub's windows-latest
+# runner (core.autocrlf=true by default, no repo override before this file)
+# converts LF to CRLF on checkout for every text file, including
+# apps/packages/design/src/css/product.css. check-theme.mjs asserts that
+# product.css contains several multi-line CSS declarations as exact,
+# hardcoded LF-joined string literals (see the `productCss.includes(...)`
+# loop around line 592), so a CRLF checkout makes those substring matches
+# fail even though the CSS content is byte-for-byte the same modulo line
+# endings. This broke `pnpm run build` -> check-theme.mjs on the first-ever
+# Windows desktop build attempt (windows-latest), independent of and in
+# addition to the separate Sentry credential issue.
+* text=auto eol=lf
--- a/apps/packages/design/scripts/check-theme.mjs
+++ b/apps/packages/design/scripts/check-theme.mjs
@@ -127,18 +127,25 @@ ${["hover", "enter", "exit", "disclosure", "panel", "pop", "emphasized"]
 ].join("\n\n");
 
 /**
- * Read a text file with line endings normalised to LF.
+ * Read a CHECKED-OUT source file with line endings normalised to LF.
  *
- * Several assertions below compare against literals joined with `\n`. A
- * working tree carrying CRLF (a Windows editor, or a checkout predating
+ * Several assertions below compare against literals joined with `\n`. A working
+ * tree carrying CRLF (a Windows editor, or a checkout predating
  * `.gitattributes`) would fail those with a message that names the CSS rather
  * than the line endings, so normalise at the read instead.
+ *
+ * Only for files that come out of git. Generated output is read raw, so its
+ * byte-identity check stays exact.
  */
 async function readText(path) {
   return (await readFile(path, "utf8")).replace(/\r\n/g, "\n");
 }
 
-const generated = await readText(resolve(root, "dist/theme.css"));
+// Deliberately NOT readText: dist/theme.css is untracked, gitignored, and
+// written fresh by generate-theme.mjs from an in-memory string moments earlier,
+// so a git checkout can never give it CRLF. Normalising here would only loosen
+// the byte-identity check this script exists to make.
+const generated = await readFile(resolve(root, "dist/theme.css"), "utf8");
 assert(
   generated === `${expectedCss}\n`,
   "dist/theme.css drifted from an independent projection of the token authority",
--- a/apps/packages/design/scripts/check-theme.mjs
+++ b/apps/packages/design/scripts/check-theme.mjs
@@ -126,7 +126,19 @@ ${["hover", "enter", "exit", "disclosure", "panel", "pop", "emphasized"]
 }`,
 ].join("\n\n");
 
-const generated = await readFile(resolve(root, "dist/theme.css"), "utf8");
+/**
+ * Read a text file with line endings normalised to LF.
+ *
+ * Several assertions below compare against literals joined with `\n`. A
+ * working tree carrying CRLF (a Windows editor, or a checkout predating
+ * `.gitattributes`) would fail those with a message that names the CSS rather
+ * than the line endings, so normalise at the read instead.
+ */
+async function readText(path) {
+  return (await readFile(path, "utf8")).replace(/\r\n/g, "\n");
+}
+
+const generated = await readText(resolve(root, "dist/theme.css"));
 assert(
   generated === `${expectedCss}\n`,
   "dist/theme.css drifted from an independent projection of the token authority",
@@ -523,7 +535,7 @@ assert(
   "the generated theme must contain exactly one flattened light root",
 );
 
-const productCss = await readFile(resolve(root, "src/css/product.css"), "utf8");
+const productCss = await readText(resolve(root, "src/css/product.css"));
 for (const { selector, body } of topLevelBlocks(productCss)) {
   const ownsGlobalScope =
     selector.startsWith("@theme") ||
@@ -627,7 +639,7 @@ checkRawMotionAuthority(generated, "dist/theme.css");
  * 7. One authority entry per token name in the source manifest
  * ------------------------------------------------------------------ */
 
-const tokenSource = await readFile(resolve(root, "src/tokens.ts"), "utf8");
+const tokenSource = await readText(resolve(root, "src/tokens.ts"));
 const manifest =
   tokenSource
     .split("export const themeTokens = {")[1]


pablonyx (commented):
## Review of record — #2157 (`diag/windows-installer-packaging`)

Reviewer of record; I did not write this. Reviewed at head `917b77705d` on base `83333cfe38`. All CI green (run `32450333935`, 30/30 non-skipped checks pass).

**Verdict: do NOT merge this branch as-is — but not for the reason I expected.** The `.gitattributes` change is safe and I can rule out the blast-radius concerns concretely (evidence below). The blocking item is the Sentry bypass, which is diagnostic scaffolding that should not become a permanent input on the release workflow. The PR body already says "not for merge as-is", so this review is mainly about which half gets promoted and under what conditions.

---

### 1. Rebase is clean — verified

`git range-diff db419a9f10..e588a07f61 83333cfe38..917b77705d`:

```
1:  929c3fc02c = 1:  9924a7da79 diag(release-desktop): add opt-in Sentry bypass ...
2:  e588a07f61 = 2:  917b77705d fix(build): force LF checkout ...
```

Both commits `=` (byte-identical patches). No silent content drift.

**Semantic conflict with #2156 (`956d2426b7`): none.** Confirmed `956d2426b7` is an ancestor of the new base and not of the old one. It touches `.github/workflows/release.yml`, `guides/deploying/releases.md`, `scripts/ci-cd/deploy-staging-trigger.test.mjs`. This PR touches `.github/workflows/release-desktop.yml` and `.gitattributes`. No file overlap, and the two "bypass" concepts are unrelated (#2156 is a push token for the version bump; this is a Sentry sourcemap skip). Nothing to reconcile.

---

### 2. The achievement is real — verified from primary evidence

Run `32447570745`, head `e588a07f617ccdaffaa43616b2cc58556dc7e2dc`, job `build-desktop (windows-latest, x86_64-pc-windows-msvc, windows)` (`96669899117`). Step conclusions: `Build frontend` **success** (21), `Build Tauri app` **success** (24), `Upload native Sentry debug files` **failure** (26), `Verify updater artifacts` **skipped** (27).

Bundler output, job log lines 5650–5666:

```
Running light to produce ...\bundle\msi\Proliferate_0.4.21_x64_en-US.msi
Running makensis to produce ...\bundle\nsis\Proliferate_0.4.21_x64-setup.exe
Finished 2 bundles at: ...  Finished 2 updater signatures at: ...
```

Filenames confirmed exactly as stated. First Windows `tauri build` in the repo's history stands.

I independently checked the matcher claim rather than taking it from the body. `scripts/generate-desktop-installer-manifest.mjs:68-69` and `scripts/generate-updater-manifest.mjs:112-117` both require `/(x64|x86_64).*setup\.exe$/i` plus `pathIncludes(relPath, "desktop-x86_64-pc-windows-msvc")`. `Proliferate_0.4.21_x64-setup.exe` and its `.sig` match the name half; the artifact name at `release-desktop.yml:737` is `desktop-${{ matrix.target }}` = `desktop-x86_64-pc-windows-msvc`, so the path half lines up statically. The MSI's `_en-US` infix is correctly irrelevant — both entries are `optional: true` and the comment already says the NSIS setup.exe is the download.

**Caveat worth stating in the body (non-blocking):** `Upload build artifacts (Windows)` (step 34) was **skipped**, so the `desktop-x86_64-pc-windows-msvc` path component and the manifest generators were never exercised end-to-end. The filename half is proven; the artifact-path half is static reasoning only. The body says "matcher assumption is correct" a bit more strongly than the run supports.

---

### 3. `.gitattributes` — I can rule out the blast radius. Safe.

I did not assume this. I measured it. Findings against the PR tree:

1. **No renormalisation diff is possible.** `git ls-files --eol` over all 7960 tracked files: **7774 `i/lf`, 102 `i/none`, 84 `i/-text` — zero `i/crlf`, zero `i/mixed`.** Independently, `git grep -I -l -F $'\r' 917b77705d` returns **0 files**. There is no CR byte in any tracked text blob, so `text=auto` normalisation is provably a no-op on existing content. Nobody gets a surprise whole-repo diff on their next commit.

2. **The rule changes nothing about classification.** I simulated the exact line via `git -c core.attributesfile=<file with `* text=auto eol=lf`>` and re-ran `ls-files --eol`. Histogram is byte-identical before and after, for both the index and worktree columns: `7774 lf / 102 none / 84 -text`. The 84 binary files stay `-text` because `text=auto` uses the same NUL-byte heuristic that already classifies them.

3. **Nothing in this repo requires CRLF.** Zero tracked `.bat`, `.cmd`, `.ps1`, `.psm1`, `.sln`, `.vcxproj`, `.reg`, `.rc`, `.manifest`, `.nsi`, `.nsh`, `.wxs`, `.iss` files. The `-text` set is 76 `png` + `webp`/`wav`/`tiff`/`jpeg`/`ico`/`icns` + 2 `.ts`.

So: **`.gitattributes` is safe to merge.** Three residual notes, none blocking:

- **`.gitattributes:13`** — two TypeScript sources contain literal NUL bytes and are git-binary today: `apps/packages/product-client/src/lib/domain/chat/models/model-support-refusals.ts` and `apps/packages/product-client/src/lib/infra/editor/highlighting.ts`. Under `text=auto` they stay binary and are untouched. But if anyone later "tightens" this to `* text` (dropping `auto`), those two get normalised and corrupted. Worth one line in the existing comment block so a future editor knows.
- Consider pre-emptively pinning the CRLF-required classes that don't exist yet but will if Windows support lands for real: `*.bat text eol=crlf` / `*.cmd text eol=crlf` / `*.ps1 text eol=crlf`. Cheap now, and it stops the blanket rule from silently stripping CR from a future byte-exact fixture.
- Migration note for the body: a developer who *already* has a Windows checkout with CRLF working files will see the tree as modified until they re-checkout or `git add --renormalize .`. Harmless here (no Windows devs today), but it should be written down rather than discovered.

**Should-fix, and I'd argue this is the more durable half of the fix (non-blocking on this PR):** `.gitattributes` fixes checkout, but not the working tree. `apps/packages/design/scripts/check-theme.mjs:526` reads `product.css` raw and asserts against `\n`-joined literals at `:586-588` via `productCss.includes(declaration)` at `:591`. A Windows contributor whose editor saves that file with CRLF still gets the same baffling `product.css lost the exact motion declaration:` failure on a *local* `pnpm run build`, before any commit, with `.gitattributes` in place. One `.replace(/\r\n/g, "\n")` at the `:526` read (and arguably `:129`) kills the bug class permanently and doesn't depend on every future clone honouring the attribute. Do both, not either.

The justification itself checks out: run `32445871270` (head `929c3fc02c`, bypass commit *without* `.gitattributes`) — Windows job `96665231382` failed at `Build frontend` with `check-theme: product.css lost the exact motion declaration: background-color var(--duration-hover) var(--ease-standard),`. Run `32447570745` (with the fix) logs `check-theme: ok (309 live tokens)`. Clean negative control.

---

### 4. Sentry bypass — all three claims verified, and it is still the blocking item

All three author claims are **true**:

- **Not wired into `workflow_call`.** `bypass_sentry_diagnostic` appears at `release-desktop.yml:29-32` under `workflow_dispatch.inputs` only. The `workflow_call.inputs` block (`:42-72`) does not contain it, so a caller passing it would be a workflow syntax error. Both callers confirmed clean: `_deploy-desktop.yml:178-186` and `nightly-release-train.yml:242-254` pass only `git_sha`/`version`/`dry_run`/`publish_updater`/`release_title`/`target_os`. On a `push` of a `desktop-v*` tag there are no `inputs` at all. `${{ inputs.bypass_sentry_diagnostic && 'true' || 'false' }}` at `:406` evaluates to `'false'` for null and for false. **Not reachable from any automated release path, including the zero-touch `nightly-release-train` leg that runs `dry_run: false, publish_updater: true, target_os: all`.** Good.
- **Falls back to a real existing code path.** `apps/desktop/vite.config.ts:14-18` gates on `authToken && org && project && release`; `:64` sets `sourcemap: sentryUploadEnabled ? "hidden" : false`. Unsetting the four vars at `:412` reaches that fallback rather than inventing a new one. Correct.
- **Does not mask the credential problem.** `Upload native Sentry debug files` (`:567`) has its own `env:` block reading `secrets.SENTRY_AUTH_TOKEN` fresh; the `unset` at `:412` is shell-local to the `Build frontend` step and cannot reach it. Confirmed empirically in the same run: job log line 5720, `error: Project not found. Ensure that you configured the correct project and organization.` Step 26 failed. Claim holds.

**Blocking — `release-desktop.yml:29-32`, `:406`, `:410-412`: this input should not land in the release workflow.** Not because it is reachable today (it is not — I verified that), but because merging it makes a manually-dispatchable "ship without sourcemaps" switch a permanent feature of the release surface, to work around a credential that takes minutes to rotate. Nothing in CI keeps it out of `workflow_call`: `scripts/ci-cd/release-title-propagation.test.mjs` asserts several properties of this workflow, but nothing asserts the bypass stays off the reusable-call surface, so the next person to "just plumb it through" hits no guard. A warning annotation is not a control — it is a log line that scrolls past in a job nobody reads on a green run.

**On the shape, if any version of this is kept:** the bypass unsets `SENTRY_AUTH_TOKEN`/`ORG`/`PROJECT`/`URL` but leaves `VITE_PROLIFERATE_SENTRY_DSN` set. A build produced this way still *initialises* Sentry and still reports errors — it just has `sourcemap: false`, so every renderer stack trace from it is permanently unsymbolicatable, with no marker in the artifact saying why. If this ever did reach a real release, the failure mode is silent garbage in Sentry rather than a loud absence, which is the worse of the two. Either drop the DSN as well (obviously-degraded build) or don't have the flag.

**Preferred resolution:** promote commit `917b77705d` (`.gitattributes`) on its own; leave `9924a7da79` (bypass) on this diagnostic branch and let it expire with the investigation. Fix the credential instead. If a guard is wanted, the durable form is a test in `release-title-propagation.test.mjs` asserting no `bypass_*` input exists under `workflow_call`.

**Nit — `release-desktop.yml:411`:** the warning string says the DSN vars share the token's timestamp. Actual values: `SENTRY_AUTH_TOKEN` and `PROLIFERATE_DESKTOP_SENTRY_DSN` are `2026-08-20T19:51:01Z`, but `VITE_PROLIFERATE_SENTRY_DSN` is `19:50:58Z`. Three seconds, but it's baked into a workflow string presented as evidence.

---

### 5. Root cause is stated plainly, and the bypass is not presented as a fix — verified

Confirmed independently against the live repo config: `SENTRY_AUTH_TOKEN` `updated_at` is **`2026-08-20T19:51:01Z`**, `PROLIFERATE_DESKTOP_SENTRY_DSN` `19:51:01Z`, `VITE_PROLIFERATE_SENTRY_DSN` `19:50:58Z`. Matches the body.

The body does state it plainly and does not oversell: it says "**This does not fix the credential**", notes the native upload still fails in the same run, says it "is not Windows-specific", and the recommendation section leads with fixing the token before any real dispatch on either platform. That is the correct framing and I have no complaint here.

One correction for accuracy: the body says "the job's overall conclusion is still `failure`". True at job level, but the **run-level conclusion of `32447570745` is `success`**, because `release-desktop.yml:139` sets `continue-on-error: ${{ matrix.runner_os == 'windows' }}`. That is deliberate and documented in #2150, not a defect here — but anyone reading the run list sees a green Windows release run that failed. Worth naming explicitly in the body so the next reader isn't misled. Same applies to run `32445871270`, also green at run level with a failed Windows job. Relatedly, the body is right that the macOS leg in that run is a vacuous pass (all steps `skipped`).

---

### Summary

| Item | Status |
|---|---|
| Rebase clean, no semantic conflict with #2156 | Verified clean |
| Windows `tauri build` + filenames | Verified from bundler log |
| `.gitattributes` blast radius | **Ruled out concretely — safe** |
| Sentry bypass unreachable from release paths | Verified — not reachable |
| Bypass does not mask the credential | Verified — step 26 still failed |
| Root cause stated plainly | Verified |

**Blocking:** the `bypass_sentry_diagnostic` input (`release-desktop.yml:29-32`, `:406`, `:410-412`) must not merge into the release workflow. Fix the Sentry credential; keep the flag on this branch.

**Non-blocking, recommended:** harden `check-theme.mjs:526` against CRLF; note the NUL-byte `.ts` caveat and pin `*.bat`/`*.cmd`/`*.ps1` in `.gitattributes`; correct the run-level-vs-job-level `success`/`failure` wording and the `Upload build artifacts (Windows)`-skipped caveat in the body.

Nothing pushed, merged, relabelled, or dispatched by this review.

pablonyx (commented):
Reviewed the new state: `917b77705d` (`.gitattributes`) + `e19c87f67e` (`readText()` in `check-theme.mjs`), diffed against base `fix/desktop-windows-release` at `83333cfe38`. Verified in a clean worktree checked out to the PR head (`e19c87f67e`), not the PR author's numbers.

## Verdict

Safe to merge. No blocking findings. Two non-blocking notes below, one of which is a body-accuracy correction worth fixing in the description (or in a follow-up), not the diff.

## 1. Sentry bypass commit: confirmed gone

`git diff --stat fix/desktop-windows-release...HEAD` touches exactly two files:

```
.gitattributes                               | 13 +++++++++++++
apps/packages/design/scripts/check-theme.mjs | 18 +++++++++++++++---
2 files changed, 28 insertions(+), 3 deletions(-)
```

`.github/workflows/release-desktop.yml` has zero diff between base and head. And directly:

```
git merge-base --is-ancestor 9924a7da79 HEAD && echo ancestor || echo not-ancestor
not-ancestor
```

The Sentry bypass commit exists in repo history (I could `git show` it) but is not reachable from this PR's head and touches no file that appears in this PR's diff. It was dropped, not reordered or squashed in.

## 2. Renormalisation no-op claim: numbers in the body are wrong, but the conclusion holds

Re-ran the measurements myself against PR head:

```
git ls-files --eol | awk '{print $1}' | sort | uniq -c
 111 i/-text
7852 i/lf
 102 i/none

git ls-files | wc -l
8065

git ls-files --eol | grep -E 'crlf|mixed'
(zero output)

git grep -I -l -F $'\r' | wc -l
0
```

The body claims 7960 total tracked files (7774 `i/lf`, 102 `i/none`, 84 `i/-text`). Actual is 8065 total (7852 `i/lf`, 102 `i/none`, 111 `i/-text`). I checked whether this is base-branch drift since the body was written: it is not, `origin/fix/desktop-windows-release` tip is still `83333cfe38`, the exact merge-base, unchanged. `git ls-tree -r <commit> | wc -l` gives 8064 at base and 8065 at both PR commits (the +1 is `.gitattributes` itself), consistently, in two independent ways of counting (`ls-files` and `ls-tree`). So the body's headline count is simply off by about 105 files from reality, for a reason I can't identify (possibly a stale local measurement, possibly a different filter). It doesn't change the conclusion: zero `crlf`, zero `mixed`, zero grep hits for a literal `\r`, confirmed independently. Non-blocking, but the body should be corrected since these are stated as measured facts with specific counts.

Checked file types that would be damaged by blanket `eol=lf`:

```
git ls-files | grep -icE '\.(bat|cmd|ps1|sln|vcxproj|reg|rc|nsi|wxs|iss|vbs|psm1|psd1|xaml|resx|manifest)$'
0
```

Confirmed zero. Also checked for tracked binary-ish files that `text=auto` might misdetect as text: the 111 `i/-text` files break down as 103 png, 2 ts, 1 each of webp/wav/tiff/jpeg/ico/icns. The two `.ts` files (`model-support-refusals.ts`, `highlighting.ts`) are flagged `-text` by git's binary heuristic despite being normal TypeScript source; each contains a handful of high-bit UTF-8 bytes (an em/en dash, 6 bytes total in one case) that trip the heuristic. This is pre-existing behavior, not something this PR causes, and `text=auto` degrading gracefully to `-text` (leave-alone) rather than forcing `eol=lf` on a misdetected file is the correct, safe direction for a false positive. Not a finding, just confirms the directive is doing what it should on the edge case.

Also searched for tracked fixtures that might legitimately contain literal CRLF bytes on purpose (e.g. a test asserting CRLF handling) that `eol=lf` would silently launder on checkout: `git grep -I -l -F $'\r'` across the whole tree returns 0, so there is nothing to protect against here today.

## 3. `.gitattributes` content

Read all 13 lines. It is a 12-line comment explaining the Windows checkout problem plus one rule: `* text=auto eol=lf`. `text=auto` (not blanket `text`) is the right choice: it lets git's own binary detection exempt binary files from EOL conversion, which is exactly what happened to the 103 PNGs etc. above (`i/-text`, unaffected by `eol=lf`). `i/none` files (102, mostly SVGs and JSON fixtures with no newlines at all) are also unaffected since there's nothing to convert. No `binary`/`-text` path-specific exemptions are needed given the extension search above returned zero hits.

## 4. `readText()` in check-theme.mjs: correct at two of three sites, one is unnecessary scope

Diff applies `readText()` (CRLF→LF normalisation) at three sites: `dist/theme.css` (line 141), `src/css/product.css` (line 538), `src/tokens.ts` (line 642). Checked the rest of `apps/packages/design/scripts/` (`copy-css.mjs`, `generate-theme.mjs`) for the same hazard: neither does a `readFile` + `\n`-literal-assertion, so nothing else was missed.

`product.css` and `tokens.ts` are genuine tracked source files subject to the Windows-checkout CRLF corruption this PR is fixing. Applying `readText()` there is correct and directly targets the bug.

`dist/theme.css` is different, and this is the one worth looking at hard per the brief. It is gitignored (`.gitignore:14: dist/`), never checked out from git at all. It's written fresh, same process, by `generate-theme.mjs` (`await writeFile(target, css)`) immediately before `check-theme.mjs` runs, from an in-memory JS string built with `\n` joins, not read from any tracked file. So it can never actually carry CRLF under the mechanism this PR fixes: normalising its read is a no-op today, not a fix. It does, however, sit on the assertion that the script's own comment calls a "byte-identical" check (`generated === \`${expectedCss}\n\``, message: "dist/theme.css drifted from an independent projection of the token authority"), and applying `.replace(/\r\n/g, "\n")` there means that check would no longer catch a hypothetical future regression where `generate-theme.mjs` starts embedding CRLF into its output (e.g. if it were changed to copy content from a tracked file instead of building it purely in memory). Today that's not exploitable: I read `generate-theme.mjs` in full and it has no `readFile` call, only `writeFile`, so there is no path to CRLF in `dist/theme.css`. Non-blocking: recommend either dropping the `readText()` call on this one site (revert to plain `readFile`) to keep the byte-equality check as tight as originally documented, or leaving it with a one-line comment acknowledging it's defensive/no-op rather than bug-fixing, so a future reader doesn't assume `dist/theme.css` was ever actually observed with CRLF.

## 5. Negative control: CI string cross-checked, matches exactly

Pulled the actual failing run rather than trusting the PR body's quote:

```
gh run view 32445871270 --repo proliferate-ai/proliferate --log-failed
...
check-theme: product.css lost the exact motion declaration: background-color var(--duration-hover) var(--ease-standard),
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @proliferate/design@0.1.0 build: ...
```

Character-identical to the string quoted in the PR body and commit message. I did not re-run the three local cells (CRLF+unpatched, CRLF+patched, LF+patched) myself: `dist/` doesn't exist in a fresh worktree and reproducing it means a full `pnpm install` + build, which I chose not to do given the box's resource constraints and because the fix itself is a straightforward, easily-inspected `.replace(/\r\n/g, "\n")` applied before every assertion that depends on exact-match against `\n`-joined literals. I'm relying on direct code reading plus the independently-confirmed CI string match rather than a live repro. No tracked files were modified in the course of this review; `git status --porcelain` in the review worktree is clean.

## 6. Body accuracy vs PR #2150

Checked PR #2150's body directly rather than trusting the cross-reference:

- Rotation timestamp: #2150 states `SENTRY_AUTH_TOKEN (and the desktop Sentry DSNs) were rotated 2026-08-20T19:51:01Z`. #2157's body states the same timestamp. Match.
- Scope: #2150 states "the new token cannot read the `proliferate-desktop` project" and "this is org-wide... and would break a macOS release dispatched today." #2157's body says the same. Match.
- No contradiction found between the two PR bodies on this point.

## Summary

- Blocking: none.
- Non-blocking: (a) the PR body's tracked-file EOL counts (7960/7774/84) don't match what I measured on PR head (8065/7852/111); the qualitative conclusion (zero crlf, zero mixed, zero literal `\r`) is independently confirmed true regardless. (b) `readText()` on `dist/theme.css` is a no-op today since that file is never checked out, only generated in-process; consider dropping it or annotating it as defensive-only so it doesn't read as fixing something it isn't.
- Sentry bypass commit `9924a7da79`: confirmed not an ancestor of PR head and touches no file present in this PR's diff. Genuinely dropped.
- Working tree left clean; no tracked files modified during this review.

pablonyx (commented):
Confirming review of the single commit that landed after the last round: `9eaee5362` "fix(design): read generated theme output raw, not normalised" (`apps/packages/design/scripts/check-theme.mjs`, +11/-4). Scope kept narrow to this commit per instructions; not relitigating earlier rounds.

**Verdict: non-blocking. The fix correctly implements what the prior review asked for, and I found nothing new to flag.**

### 1. Read-site audit
Diff and full-file grep at head confirm exactly 4 occurrences of `readFile`/`readText` in the file:
- `readText()` helper definition (unchanged behavior: normalises CRLF→LF)
- `dist/theme.css` — now raw `readFile(...)`, no normalisation (this is the revert)
- `src/css/product.css` (line 545) — still `readText(...)`
- `src/tokens.ts` (line 649) — still `readText(...)`

No fourth read site was missed or altered.

### 2. Byte-identity check — is the raw read actually safe on Windows? (the crux)
Traced `generate-theme.mjs`: it builds the CSS purely from in-memory JS template literals/joins and writes it with `writeFile(target, css)` — no `readFile` anywhere in that script, no `os.EOL`, no literal CR bytes in the source. Node's `writeFile` does not translate newlines, so on its own this doesn't fully prove Windows-safety (a stale CRLF-checked-out copy of `generate-theme.mjs` could in principle carry CRLF inside its own multi-line template literals).

So I verified empirically rather than trusting the reasoning chain. In a throwaway clone (no repo files touched):
- Built the package at head `9eaee5362` on a normal LF tree: `check-theme: ok (309 live tokens)`.
- Then converted **every** source file in `src/` and `scripts/` (including `check-theme.mjs` and `generate-theme.mjs` themselves) to CRLF byte-for-byte, deleted `dist/`, and rebuilt from scratch (`tsc` → `generate-theme.mjs` → `check-theme.mjs`) against that all-CRLF source tree.
- Result: `dist/theme.css` came out with **0** CRLF sequences, and `check-theme.mjs` printed `check-theme: ok (309 live tokens)` again — pass, not a false failure and not a false pass hiding real drift (token count identical to the LF run).

Why it holds even under a fully-CRLF checkout: ECMAScript template literals normalise CR/CRLF line terminators to LF when computing the "cooked" string value, independent of the source file's on-disk line endings. So the generated CSS is LF-only regardless of git checkout behavior, `.gitattributes` state, or platform — a stronger guarantee than the commit message's stated rationale (gitignored/untracked/"checkout can never hand it CRLF"), though that rationale is also true and sufficient on its own. Either way: the raw read of the generated file cannot reintroduce the Windows CRLF failure the PR exists to fix. This matches the commit's own claimed reverification ("LF tree ok / CRLF tree ok, 309 live tokens both").

### 3. Diff scope
The +11/-4 is exactly: the one-line revert (`readText(...)` → `readFile(...)`) for `dist/theme.css`, a narrowed docstring on `readText()` ("CHECKED-OUT source files" + a note pointing at generated output), and a new comment block explaining why that one read is deliberately raw. No comparison semantics changed elsewhere, no drive-by edits.

### 4. Ran the check
Not skipped — this is JS-only (no cargo/rustc/Docker touched; confirmed `pgrep -x cargo`/`pgrep -x rustc` empty throughout). From a scratch clone: `npm install` in `apps/packages/design` (30 packages, 20s), then `tsc -p tsconfig.json && node scripts/generate-theme.mjs && node scripts/check-theme.mjs`. Literal output: `check-theme: ok (309 live tokens)` on both the baseline LF tree and the synthetic all-CRLF tree described above.

### 5. CI at head `9eaee5362`
Combined status: `success`. Every check-run is `completed / success` except two expected non-required skips (`docker`: `completed / skipped`, `self-hosted-release-assets`: `completed / skipped`). No failures, no pending. Quoting `gh pr checks 2157`: `Build SDK pass`, `Cargo check & test pass`, `Desktop frontend build pass`, `Shared frontend packages (1-4) pass`, `Web frontend build pass`, `Repo shape checks pass`, `Greptile Review pass`, `ci-ok pass`, `server-ci-ok pass`, `Validate PR title and labels pass` — full list green.

### 6. Base chain dependency
Checked file overlap: neither #2150 (`fix/desktop-windows-release`, this PR's base) nor #2146 (`fix/sdk-generate-windows-safe`) touches `apps/packages/design/**` or `.gitattributes`. #2150's changed files are all release/installer-manifest/workflow files; #2146's are the anyharness SDK openapi-generation files. So `9eaee5362`'s correctness has no dependency on either PR's content — only a mechanical rebase of #2157's base pointer will be needed once that chain settles, not a re-review of this logic.

**Correction to the task brief:** at the time I checked (a few minutes ago), #2146 (`fix/sdk-generate-windows-safe` → `main`) reported `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` — not `CONFLICTING` as stated in my brief. #2150 (`fix/desktop-windows-release` → #2146) is the one currently `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`. This is consistent with "#2146 is being rebased separately right now" (likely already resolved between when the brief was written and when I checked) and doesn't change finding 6 above either way, since there's no file overlap.

No other issues found. Not approving/requesting changes per instructions — flagging for Pablo's own merge decision.

PR #2158, opened 2026-08-21
fix(release): pass lowercase production environment to deploy lanes

pablonyx:
## Problem

The first full production release (run 32450223908) deployed server and web successfully but **lost the LiteLLM deploy**:

```
aws: [ERROR]: An error occurred (AccessDeniedException) when calling the PutParameter operation:
User: arn:aws:sts::157466816238:assumed-role/proliferate-prod-github-actions-deploy/GitHubActions
is not authorized to perform: ssm:PutParameter on resource:
arn:aws:ssm:us-east-1:157466816238:parameter/proliferate/Production/litellm/database-url
```

Note the capital `P` in the resource path. `release.yml` passes `environment: Production`, while the coordinator it replaced (`promote-production.yml`, deleted in #2140) passed `environment: production`.

GitHub matches environment names case-insensitively, so the capitalized input binds to the same Production environment and nothing fails at bind time. But that same string becomes `DEPLOY_ENVIRONMENT`, which is interpolated raw into AWS resource paths:

- `_deploy-litellm.yml:152` — `/proliferate/${DEPLOY_ENVIRONMENT}/litellm/*`
- `_deploy-server.yml:525,529` — `/proliferate/${DEPLOY_ENVIRONMENT}/support/*`
- `_deploy-server.yml:1221,1225` — `Proliferate/Background/${DEPLOY_ENVIRONMENT}` and `/ecs/proliferate-server-${DEPLOY_ENVIRONMENT}`

Every one of those resources exists only under the lowercase spelling (verified with `aws ssm describe-parameters`: all live parameters are `/proliferate/production/...`), and the deploy role's IAM policy is scoped to that path.

`deploy-server-prod` survived only by luck: `server/deploy/hosted-redis-contract.json` lists both spellings in `workflow_names`, and its own SSM writes are conditional on support secrets that are empty in this environment. Its CloudWatch log-group derivation still used the wrong name, which is worth a separate look.

## Fix

Pass `production` (lowercase) from all three prod deploy lanes in `release.yml`, restoring the exact string the retired coordinator used. No other file changes behavior.

## Regression guard

`scripts/ci-cd/deploy-staging-trigger.test.mjs` gains a test asserting every `environment:` input in `release.yml` and `deploy-staging.yml` is an exact key of `hosted-redis-contract.json`'s `environments` map, which is the same token the AWS resources are named after. The pre-existing comment claiming the two spellings are equivalent is corrected, since that assumption is what broke.

- Positive control: `node --test` 7/7 pass on this branch.
- Negative control: reintroducing `Production` fails with `release.yml: environment input 'Production' is not a hosted contract environment key (staging, production)`.

## Deploy impact

Server and web production are already live from run 32450223908. Only LiteLLM needs a redeploy, which this unblocks.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -351,7 +351,7 @@ jobs:
     if: ${{ needs.prepare.result == 'success' && needs.prepare.outputs.selected_surfaces != '' }}
     uses: ./.github/workflows/_deploy-server.yml
     with:
-      environment: Production
+      environment: production
       git_sha: ${{ needs.prepare.outputs.head_sha }}
       enabled: ${{ needs.prepare.outputs.server == 'true' && needs.prepare.outputs.dry_run != 'true' }}
     secrets: inherit
@@ -361,7 +361,7 @@ jobs:
     if: ${{ needs.prepare.result == 'success' && needs.prepare.outputs.selected_surfaces != '' }}
     uses: ./.github/workflows/_deploy-litellm.yml
     with:
-      environment: Production
+      environment: production
       git_sha: ${{ needs.prepare.outputs.head_sha }}
       enabled: ${{ needs.prepare.outputs.litellm == 'true' && needs.prepare.outputs.dry_run != 'true' }}
     secrets: inherit
@@ -384,7 +384,7 @@ jobs:
       }}
     uses: ./.github/workflows/_deploy-web.yml
     with:
-      environment: Production
+      environment: production
       git_sha: ${{ needs.prepare.outputs.head_sha }}
       enabled: ${{ needs.prepare.outputs.web == 'true' && needs.prepare.outputs.dry_run != 'true' }}
     secrets: inherit
--- a/scripts/ci-cd/deploy-staging-trigger.test.mjs
+++ b/scripts/ci-cd/deploy-staging-trigger.test.mjs
@@ -163,10 +163,17 @@ test("deploy environment inputs use the exact casing the AWS resource paths use"
   // But the same string becomes `DEPLOY_ENVIRONMENT`, which is interpolated
   // raw into SSM parameter paths (`/proliferate/${DEPLOY_ENVIRONMENT}/litellm/...`
   // in _deploy-litellm.yml, `/proliferate/${DEPLOY_ENVIRONMENT}/support/...` in
-  // _deploy-server.yml) and into the CloudWatch log group name. Those resources
-  // exist only under the lowercase spelling, and the deploy role's IAM policy is
-  // scoped to it, so a capitalized input fails with AccessDeniedException at
-  // deploy time. Release run 32450223908 lost its LiteLLM deploy to exactly this.
+  // _deploy-server.yml) and into Sentry environment values on the server task
+  // definition. Every live SSM parameter uses the lowercase spelling and the
+  // deploy role's IAM policy is scoped to it case-sensitively, so a capitalized
+  // input fails with AccessDeniedException at deploy time. Release run
+  // 32450223908 lost its LiteLLM deploy to exactly this, and shipped a server
+  // task definition carrying SENTRY_ENVIRONMENT=Production.
+  //
+  // Not covered by this test: _deploy-server.yml's CloudWatch derivations
+  // (`Proliferate/Background/${DEPLOY_ENVIRONMENT}`, `/ecs/proliferate-server-${DEPLOY_ENVIRONMENT}`)
+  // match no real resource under EITHER spelling; the live log group is
+  // `/ecs/proliferate-prod`. That is a separate defect.
   //
   // The canonical spelling is the environment key in the hosted contract, which
   // is the same token the AWS resources are named after.
@@ -176,16 +183,40 @@ test("deploy environment inputs use the exact casing the AWS resource paths use"
   const canonical = new Set(Object.keys(contract.environments));
   assert.ok(canonical.size > 0, "the hosted contract must declare environments");
 
-  for (const name of ["release.yml", "deploy-staging.yml"]) {
+  // Every `environment:` line is inspected at any indentation, with an inline
+  // comment stripped, surrounding quotes removed, and trailing whitespace
+  // trimmed, so none of those spellings can smuggle a capitalized value past
+  // the check. Expression-valued inputs (`${{ ... }}`) are resolved at run time
+  // by the caller, which is itself one of the files checked here.
+  for (const [name, expected] of [
+    ["release.yml", 3],
+    ["deploy-staging.yml", 4],
+  ]) {
     const source = read(name);
-    const passed = [...source.matchAll(/^      environment: (\S+)$/gm)].map((match) => match[1]);
-    assert.ok(passed.length > 0, `${name}: expected at least one environment input`);
-    for (const value of passed) {
+    const lines = source.split("\n").filter((line) => /^\s*environment:\s*\S/.test(line));
+    // Pin the count so a refactor that renames or drops the key trips this test
+    // instead of vacuously passing on an empty match set.
+    assert.equal(
+      lines.length,
+      expected,
+      `${name}: expected ${expected} environment inputs, found ${lines.length}. ` +
+        `Update this test deliberately when the deploy lanes change.`,
+    );
+
+    for (const line of lines) {
+      const raw = line
+        .replace(/^\s*environment:\s*/, "")
+        .replace(/\s+#.*$/, "")
+        .trim()
+        .replace(/^["'](.*)["']$/, "$1");
+      if (raw.includes("${{")) {
+        continue;
+      }
       assert.ok(
-        canonical.has(value),
-        `${name}: environment input '${value}' is not a hosted contract environment key ` +
-          `(${[...canonical].join(", ")}). The input is interpolated into SSM paths, so the ` +
-          `casing must match the AWS resources exactly.`,
+        canonical.has(raw),
+        `${name}: environment input '${raw}' is not a hosted contract environment key ` +
+          `(${[...canonical].join(", ")}). The input is interpolated into SSM paths and ` +
+          `Sentry environment values, so the casing must match the AWS resources exactly.`,
       );
     }
   }
--- a/scripts/ci-cd/deploy-staging-trigger.test.mjs
+++ b/scripts/ci-cd/deploy-staging-trigger.test.mjs
@@ -31,7 +31,10 @@ function triggerBlock(source, name) {
 const MAIN_BRANCH_FILTER = /branches:\s*(?:\[[^\]]*\bmain\b[^\]]*\]|(?:\r?\n\s*-\s*["']?main["']?))/;
 // `Production` and `production` are the same GitHub Environment reference as
 // far as a reviewer scanning for prod reach is concerned, and the two deleted
-// coordinators both used the lowercase spelling.
+// coordinators both used the lowercase spelling. The spellings are NOT
+// interchangeable at deploy time, though: the string is passed through to
+// `DEPLOY_ENVIRONMENT` and interpolated into SSM parameter paths. See the
+// deploy-environment casing test below.
 const PRODUCTION_ENVIRONMENT = /environment:\s*["']?[Pp]roduction["']?/;
 
 test("staging runs only when an operator dispatches it", () => {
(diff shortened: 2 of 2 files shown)


pablonyx (commented):
**Reviewer verdict: APPROVE**

Dedicated reviewer of record (fresh, did not write this change). The root cause is correct, the fix is the right one, and I found nothing where lowercasing breaks something that currently works. Findings below are non-blocking: four are PR-body/comment accuracy, one is test brittleness, one is a stale artifact for a follow-up.

## Root cause: confirmed and sufficient

- `git show 95626925f5^:.github/workflows/promote-production.yml` passes `environment: production` on all seven lanes (:155, :167, :177, :187, :197, :207, :220). `release.yml` regressed it to `Production` on the three surviving lanes.
- Failure is a case-sensitive IAM ARN match, not a GitHub-side problem. Read-only check of the prod deploy role: inline policy `litellm-parameter-writes` allows `ssm:PutParameter` only on `arn:aws:ssm:us-east-1:157466816238:parameter/proliferate/production/litellm/*`; `support-tracker-parameter-writes` likewise on `/proliferate/production/support/*`. `aws ssm describe-parameters` returns only `/proliferate/production/...` and `/proliferate/staging/...` — the PR body's claim checks out.
- Fail-closed behaved correctly: the `PutParameter` error aborted the render step before `register-task-definition`, so `proliferate-prod-litellm:11` still points at `/proliferate/production/litellm/*` and image `fe938ca564bd`. Worth noting the counterfactual — had IAM been broader, the render would have pinned `valueFrom: /proliferate/Production/...` and the ECS execution role could not have read them. This fix is the correct direction, not just the expedient one.

## Consumer enumeration: no caller requires the capitalized form

Every reader of `inputs.environment` / `DEPLOY_ENVIRONMENT`:

| Consumer | Effect of lowercasing |
| --- | --- |
| Job-level `environment:` (`_deploy-server.yml:32`, `_deploy-litellm.yml:53`, `_deploy-web.yml:26`, `_deploy-desktop.yml:38`) | None. GitHub resolves case-insensitively; the repo has exactly one env named `Production`, and the deleted coordinator drove it with `production` for months. Empirical proof: ECR `proliferate-litellm:production-latest` pushed 2026-08-18, `proliferate-server:production-latest` 2026-08-19, both by the prod deploy role, both from lowercase-input runs. `vars`/`secrets` scope follows the same resolution. |
| Hosted-contract jq lookup (`_deploy-server.yml:129-134`) | None. Keys on `workflow_names`, which lists both spellings. |
| ECR alias tags (`_deploy-server.yml:293`, `_deploy-litellm.yml:131`) | Push-only alias. The deployed reference is `repo:<short_sha>` (server additionally re-pins to the immutable digest). Both repos are `MUTABLE`, so re-pushing `production-latest` works. Nothing in the repo pulls `${environment}-latest` except F5 below. |
| Vercel target (`_deploy-web.yml:29-30`) | None — comes from `vars.VERCEL_ENVIRONMENT \|\| vars.VERCEL_TARGET \|\| 'production'`, never from `inputs.environment`. No S3 prefix / CloudFront / cluster / service name is derived from it anywhere. |
| SSM paths (`_deploy-litellm.yml:152`, `_deploy-server.yml:525,529`) | Fixed — this is the bug. |
| `SENTRY_ENVIRONMENT` family (`_deploy-server.yml:600,602,603`) | Real change — see F2. |
| CW namespace + log group (`_deploy-server.yml:1221,1225`) | Still wrong after the fix — see F4. |
| Step-summary echoes | Cosmetic. |

Coverage is complete: `release.yml` has exactly 3 `environment:` inputs (:354, :364, :387), all changed; `deploy-staging.yml` has 4, all already `staging`. Both a tight and a permissive regex find exactly those 7 and no others.

## Dimension 4 — is the successful server deploy trustworthy?

**Yes. Nothing false-passed.** Steps 16-19 of job `96677320514` — *Verify broker and scheduler store*, *Deploy worker and Beat from candidate image*, *Verify worker and Beat health*, and *Prove candidate plane executes enqueued work* — were all **skipped**, because `ECS_WORKER_SERVICE` and `ECS_BEAT_SERVICE` are empty in the `Production` environment (job log lines 68-69). The wrong CloudWatch namespace and log group were therefore never queried and could not produce a vacuous pass. Every step that did run (config validation, AWS identity verification, support-feed and Redis secret probes, task render, Alembic migrations, service roll, health smoke) is casing-independent — with the one exception in F2.

For completeness: the proof step's helpers *do* swallow AWS errors to `0` (`|| echo None`, `|| echo 0`), so a nonexistent log group would silently read as "no receipt" — but the loop then fails closed on timeout at `:1381`, so even if it had run it would have failed rather than false-passed. The gate is sound; it just never fired.

## Findings

**F1 — MEDIUM (PR body factually wrong; conclusion unchanged).** The body says `deploy-server-prod` "survived only by luck: ... its own SSM writes are conditional on support secrets that are empty in this environment." That is not what happened. Job log lines 88-102 show `SUPPORT_TRACKER_ENABLED: true`, `SUPPORT_GITHUB_APP_PRIVATE_KEY: ***`, `SUPPORT_LINEAR_API_KEY: ***` — both secrets are set and **both `put-parameter` calls ran**. They landed on the correct path because `SUPPORT_GITHUB_APP_PRIVATE_KEY_PARAMETER_NAME=/proliferate/production/support/github-app-private-key` and `SUPPORT_LINEAR_API_KEY_PARAMETER_NAME=/proliferate/production/support/linear-api-key` are set as explicit lowercase GitHub vars, so the `${DEPLOY_ENVIRONMENT}` fallbacks at `_deploy-server.yml:525` and `:529` were never taken. The real lesson is sharper than "luck": those fallbacks are **one unset var away** from the identical `AccessDeniedException`, since `support-tracker-parameter-writes` is scoped to `/proliferate/production/support/*`. Please correct the body.

**F2 — MEDIUM (live wrong value the PR body omits).** The capitalization *did* leave bad state in production. `_deploy-server.yml:600,602,603` render `SENTRY_ENVIRONMENT`, `CLOUD_RUNTIME_SENTRY_ENVIRONMENT`, and `CLOUD_TARGET_SENTRY_ENVIRONMENT` from `$environment`. The live task def `proliferate-prod-server:141` (registered by run 32450223908) carries all three as `"Production"`, next to a var-driven `E2B_RUNTIME_SENTRY_ENVIRONMENT: "production"`. Prod Sentry has been reporting into a second, capitalized environment since ~2026-08-20 22:30Z, and the two `CLOUD_*` values propagate onward to `PROLIFERATE_TARGET_SENTRY_ENVIRONMENT` / `ANYHARNESS_SENTRY_ENVIRONMENT` in cloud runtimes (`server/proliferate/server/cloud/runtime/bootstrap.py:62,127`). This PR fixes it, but **only after a server redeploy** — so "Only LiteLLM needs a redeploy" is not quite right. Recommend calling the Sentry split out in the body/release notes and redeploying the server too.

**F3 — MEDIUM (the regression guard catches only the literal shape of today's mistake).** `scripts/ci-cd/deploy-staging-trigger.test.mjs:178`, `/^      environment: (\S+)$/gm`, anchors on exactly six spaces and requires end-of-line immediately after the value. I ran eight mutations in a scratch clone:

| Mutation | Result |
| --- | --- |
| `      environment: Production` | **fails correctly** — the stated negative control reproduces |
| `      environment: prodution` | fails correctly |
| `      environment: Production # prod` | **passes 7/7 — bug undetected** |
| `      environment: Production ` (trailing space) | **passes 7/7 — bug undetected** |
| `        environment: Production` (8-space indent) | **passes 7/7 — bug undetected** |
| `      environment: "production"` (valid YAML) | fails — false alarm |
| `      environment: 'production'` (valid YAML) | fails — false alarm |

It also only ever asserts `passed.length > 0`, so deleting two of the three prod lanes still passes, and it hardcodes two filenames so a future prod coordinator is uncovered. Suggested hardening: `/^\s*environment:\s*["']?([^"'#\s]+)["']?\s*(?:#.*)?$/gm`, plus asserting the match count equals the number of `uses: ./.github/workflows/_deploy-` lines (3 and 4 today — I confirmed both regexes find exactly those). Not blocking; the guard is strictly better than nothing and would have caught the actual regression.

**F4 — MEDIUM (the new test comment asserts something false).** Lines 165-168 of the new test say the string is interpolated "into the CloudWatch log group name. Those resources exist only under the lowercase spelling." The log group and namespace exist under **neither** spelling. Read-only AWS: the only ECS log groups are `/ecs/proliferate-prod`, `/ecs/proliferate-staging`, `/ecs/proliferate-server`, `/ecs/proliferate-web`, `/ecs/proliferate-worker`, `/ecs/proliferate-gateway`, `/ecs/proliferate-llm-proxy`, `/ecs/proliferate-analytics-ingest`, `/ecs/proliferate-prod-litellm`, `/ecs/proliferate-staging-litellm` — there is no `/ecs/proliferate-server-production` or `-staging`, and the live prod server logs to `/ecs/proliferate-prod`. `aws cloudwatch list-metrics` shows no `Proliferate/Background/*` namespace at all (only `Proliferate/Prod`). So `_deploy-server.yml:1221` and `:1225` are wrong regardless of casing, and lowercasing does not fix them. The body's "worth a separate look" is right; please soften the test comment so a future reader doesn't conclude the candidate-plane proof is now correctly targeted. (Dormant today — the step is skipped in prod — but it will fail closed the moment `ECS_WORKER_SERVICE`/`ECS_BEAT_SERVICE` are set without a `CANDIDATE_PROOF_LOG_GROUP` override.)

**F5 — LOW (stale capital-tag reference; follow-up, not a regression from this PR).** `server/infra/analytics/ecs-taskdef.json:12` pins `proliferate-server:Production-latest`. Run 32450223908 accidentally *created* that tag — ECR shows `Production-latest` → `6eac131a4706`, pushed 2026-08-20T22:24 PDT; before that the repo had only `production-latest` (`be045d63dafc`, 08-19) and a 2026-05 `latest`. After this PR the release lane goes back to advancing `production-latest`, so `Production-latest` freezes at `6eac131a4706` permanently. Nothing live pulls it: the registered `proliferate-analytics-ingest:6` pins `proliferate-server:2336d10092ed`, and `server/infra/analytics/README.md` explicitly states the checked-in file is not the canonical system. Still, it is the only place in the repo that assumes the capitalized tag — worth a one-line follow-up to lowercase it, and possibly to delete the orphaned `Production-latest` tags.

**F6 — LOW.** The hosted contract still advertises `"workflow_names": ["production", "Production"]` / `["staging", "Staging"]`. Now that the test bans the capitalized spelling in the two coordinators, keeping the aliases in the contract leaves the trap armed for any new caller that bypasses the guard. Consider dropping them, or adding a comment explaining why they stay.

## Verification performed

- Positive control: `node --test 'scripts/ci-cd/*.test.mjs'` on `bda7fe5547` → **209/209 pass** (7/7 in the touched file), matching the body's claim. Wired into CI at `ci.yml:188`; `CI/CD config checks` is green and `ci-ok` passes.
- Negative controls: eight mutations, table above. The PR's stated negative control reproduces verbatim, including the error message.
- Read-only AWS: IAM inline policies on `proliferate-prod-github-actions-deploy`, `ssm describe-parameters`, `ecr describe-images`/`describe-repositories` for both repos, `ecs describe-task-definition` for `proliferate-prod-server:141` / `proliferate-prod-litellm:11` / `proliferate-analytics-ingest:6`, `logs describe-log-groups`, `cloudwatch list-metrics`. No mutations.
- Job logs read in full for `96677320574` (failed litellm) and `96677320514` (successful server), plus the step-conclusion list for the latter.

## Could not verify

- Whether Sentry's UI actually shows a split `Production` environment (no Sentry access) — F2 is inferred from the live task definition, which is unambiguous.
- GitHub's case-insensitive environment resolution is established empirically here (only `Production` exists in the repo's environment list, yet lowercase-input runs assumed the prod role and pushed to prod ECR), not from documentation.

pablonyx:
## Review round 2 applied (`ce33d4ac8e`)

**F3 (test brittleness) — fixed.** The regex is replaced with a line parser that strips inline comments, trims trailing whitespace, unquotes, accepts any indentation, and skips expression-valued inputs. The weak `length > 0` is replaced with a pinned per-file count so a rename or deletion trips the test instead of vacuously passing.

Re-ran the reviewer's controls plus a deletion case:

| Mutation | Result |
|---|---|
| `environment: Production` | FAIL (correct) |
| `environment: Production # prod` | FAIL (correct) — passed before |
| `environment: Production ` (trailing space) | FAIL (correct) — passed before |
| 8-space indent + capital | FAIL (correct) — passed before |
| `environment: "production"` (quoted, valid) | PASS (correct) — falsely failed before |
| line deleted entirely | FAIL (correct) on the count assertion |
| baseline | 7/7 pass |

**F4 (comment inaccuracy) — fixed.** The comment no longer claims the CloudWatch resources exist under the lowercase spelling. It now records that `_deploy-server.yml:1221,1225` match no real resource under *either* spelling (the live group is `/ecs/proliferate-prod`) and marks that as a separate defect, out of scope here.

**F1 / F2 — accepted, the PR body was wrong.** The support SSM writes did run; they landed correctly only because `SUPPORT_*_PARAMETER_NAME` is pinned to lowercase, bypassing the `${DEPLOY_ENVIRONMENT}` fallback. And the live task definition `proliferate-prod-server:141` does carry `SENTRY_ENVIRONMENT=Production`, so the server needs a redeploy after this merges, not just LiteLLM. Both corrections are reflected in the deploy plan.

**F5 / F6 — follow-ups, not blocking:** the accidental `Production-latest` ECR tag pinned in `server/infra/analytics/ecs-taskdef.json:12` (non-canonical file; the live task pins a SHA), and `hosted-redis-contract.json` still advertising both spellings in `workflow_names`, which leaves the trap armed for any caller outside the two guarded files.

Thanks for catching the three mutations that slipped past the original assertion.

PR #2160, opened 2026-08-21
feat(sessions): reap idle agent sessions

pablonyx:
## Summary

- Adds an idle-session reaper to `LiveSessionManager` (`live/sessions/manager/reaper.rs`). One sweep task per manager, wired in `app/sessions.rs`, retires every live session that has been continuously quiescent for the threshold.
- Quiescence is all of: phase `Idle`, no pending interactions, the handle's busy flag clear, zero pending `session_background_work` rows, empty `session_pending_prompts` queue, no pending wake schedule whose delivery needs a live parent, and a cold start that the launch policy would accept. A durable read that fails yields `Undetermined` and never reaps; the predicate fails closed.
- The sweep's verdict is only a nomination. The reaper sends `UnloadIfIdle`, and the actor re-decides serially on its own loop before it leaves; a refusal names a bounded reason and resets the idle clock.
- `AwaitingInteraction` is never reaped, per the founder ruling. See the ruling section below.
- Retirement is the existing non-terminal `Unload` disposition, so the durable session row, transcript, configuration and `native_session_id` all survive. The next prompt resumes through the ordinary startup strategy matrix.
- The actor's exit sequence signals the agent's whole process GROUP rather than dropping its direct child, because the direct child is the ACP adapter and the vendor CLI beneath it is 64% of the memory this feature exists to reclaim.
- Threshold is `ANYHARNESS_IDLE_SESSION_REAP_SECONDS`, whole seconds, **default 120**. `0` disables the reaper; an unparseable value keeps the default.
- Spec: `delivery/idle-session-reaper/delivery-spec-idle-session-reaper.md`. Current docs updated in the same PR: `specs/anyharness/acp.md`, `specs/developing/reference/env-vars.yaml`.

### Why 120 seconds, and why the threshold is not a tradeoff

The 2026-08-15 draft of this spec rested on an estimate: "19 idle agents at roughly 2 GB tipped the machine into swap". That is superseded by a controlled measurement (`Fleet load model results 2026-08-21`, `m8g.4xlarge`, 45 live Bedrock-backed sessions, PSS from `/proc/<pid>/smaps_rollup` over each profile's cgroup tree).

- A session costs **207 MB**, dead straight from the second session to the fifteenth. Three independent cold-boot runs agree to 0.09% at n=15. Endpoint slope `(3707.5 - 813.4) / 14 = 206.7 MB`. Each session adds exactly two processes.
- A session that has run one turn holds **33 MB more, permanently**. The post-turn curve is flat from t+15s to t+900s, moving 0.3 MB over the last 600 seconds. 86% of peak growth is still resident a quarter hour after the work finished, and 91% of it sits inside the vendor CLI.
- Therefore **there is nothing to wait for**. Waiting reclaims exactly zero. The reaper is the only reclaim mechanism, and reaping recovers the whole ~240 MB, not the 33 MB increment.

The design consequence: since delay buys zero memory back, a 60-second threshold and a 900-second threshold reclaim the identical amount. The threshold is a pure user-experience question, not a memory-versus-latency tradeoff. 120s is the top of the founder's ruled range ("a minute or 2"), so it is the least user-hostile point inside the ruling, and it costs nothing in memory relative to 60s.

### The `AwaitingInteraction` question

The same measured run parked one of six turns in `awaiting_interaction` on an unanswered permission prompt, holding its full memory cost for the rest of the run. Never reaping that state is correct for a user who is about to answer and a permanent leak for a prompt nobody will ever answer.

**Ruled out of scope for this PR, deliberately.** Retiring an `AwaitingInteraction` session is not the same operation as retiring an idle one: the `Unload` path resolves pending interactions as `Cancelled`, aborting the in-flight turn and discarding its work. Retiring a finished-turn session costs a cold start; retiring one parked mid-turn costs the turn. Different prices deserve different thresholds, and raising that threshold overrides an explicit founder ruling, which is not an implementer's call.

What this PR does instead is make the leak measurable: every sweep counts sessions held back solely for this reason under `result_class = "awaiting_interaction_held"`. The proposed follow-up, needing a founder ruling first, is a second much longer `AwaitingInteraction` threshold defaulting to disabled. `IdleReapPolicy` is shaped so that arrives as an extra field, not a redesign.

### Resolved verification items the frozen spec flagged

- **Wake schedules survive reaping, but only ONE relation's delivery survives a reaped parent.** Corrected after review; the original claim here was wrong about which subsystem delivers a cowork wake. `session_link_wake_schedules` is a durable row untouched by actor retirement, and `LinkCompletionStore::insert_completion_and_consume_schedule` consumes it and inserts the parent's wake prompt in one store transaction with no live parent involved. What happens next depends on the relation. For `subagent`, `persist_terminal_turn_in_tx` writes a `session_link_completion_deliveries` row (`WHERE relation = 'subagent'`) and `CompletionDeliveryWorker` claims it on a 1s lease cadence and cold-starts the parent through `activate_durable_prompt_consumer` / `ensure_live_session_handle`, so a reaped parent is woken. For `cowork_coding_session`, no outbox row is ever written: `deliver_cowork_coding_completion` sends the wake through `acp_manager.get_handle(...)` and skips it silently when the parent is gone, and nothing scans for stranded pending prompts, so a reaped parent's wake sits in `session_pending_prompts` until a human next opens the session. The fix is in the predicate rather than in a new drainer: a parent holding a pending wake schedule on any relation other than `subagent` is not reapable, which also holds back unknown future relations. Once the wake prompt row exists the durable-queue check takes over.
- **Durable prompt queue.** The check is `QueueDurable::list_pending_prompts`, the same rowset the actor's queue drain reads, so retirement cannot drop a queued prompt.
- **Subagent orphaning.** Delegated subagents are ordinary sessions in the same map and the relationship is durable in `session_links`. Reaping a parent cannot orphan a live child. No special casing added.

### Correction to the frozen spec

The draft said resume uses `ResumeSeqFreshNative`. That holds only for a session that never ran a turn. For a Claude session with a recorded `native_session_id` and a `last_prompt_at` (the ordinary reapable case), `choose_startup_strategy` selects `LoadNative(native_session_id)`, which is the stronger path. Intent unchanged, naming corrected in the spec.

## Testing

Tier 2, 19 tests in `live/sessions/manager/reaper/tests/` (`quiescence`, `wake`, `relaunch`, `policy`), against a real in-memory `SessionStore` and the real `unload_session_if_still_idle` path. Their actors are scripted mailbox consumers, so they can prove the predicate and the manager's handling of an answer, and they cannot prove anything about processes or durable status.

```
test result: ok. 19 passed; 0 failed; 0 ignored; 0 measured; 2233 filtered out; finished in 2.85s
```

Three real-process tests in `live/sessions/actor/tests/idle_reap.rs` drive a REAL `SessionActor` over a real ACP duplex against a real spawned process tree, which is what makes the memory-reclaim premise observed rather than asserted:

- `a_conditional_unload_kills_the_agents_whole_process_group_and_stays_resumable`: the reap kills a grandchild that outlives its parent's stdin, and the durable status moves from the seeded `running` to `idle` with `native_session_id` intact.
- `a_prompt_queued_behind_the_reap_wins_and_the_session_is_kept`: a prompt sent after the reap command is answered by the still-live actor.
- `a_mid_turn_reap_is_refused_without_cancelling_the_turn`: no `session/cancel` is sent and the turn survives.

```
test result: ok. 22 passed; 0 failed; 0 ignored; 0 measured; 2231 filtered out; finished in 5.18s
```

Full `cargo test -p anyharness-lib --lib` after the review fixes:

```
test result: ok. 2250 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 159.11s
```

`domains::agents::route_auth::probe_materialization::...::the_sweep_removes_only_abandoned_and_old_scratch_roots` failed on one earlier run of that suite and passed on the next. It is environment-dependent and pre-existing: with this branch's changes stashed it fails identically on the base.

Checkers: `check_docs.py`, `check_max_lines.py`, `check_anyharness_boundaries.py`, `check_design_attribution.py`, `check_session_mutation_admission.py` all pass.

### Negative controls

Every control mutates product behaviour with all signatures intact, then reverts. No control is a compile failure. The nine controls from the first round still stand; the six below cover the review fixes, and every one was re-run against the current tree.

**NC-B1 - the relaunchability probe.** `session_can_relaunch_from_cold` still calls the launch policy and then returns `Ok(true)` regardless.

```
test ...relaunch::a_zero_turn_claude_fork_child_is_never_reaped ... FAILED
panicked at .../reaper/tests/relaunch.rs:29:5:
assertion `left == right` failed: reaping a session no startup strategy accepts would end it permanently
  left: ["child-1"]
 right: []
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2252 filtered out; finished in 0.10s
```

**NC-B2 - the pending-wake hold.** `Ok(true) => return IdleReapVerdict::PendingWake` becomes `Ok(true) => {}`.

```
test ...wake::a_cowork_parent_expecting_a_wake_is_not_reaped ... FAILED
panicked at .../reaper/tests/wake.rs:120:5:
assertion `left == right` failed: reaping a cowork parent strands the wake its child is about to send
  left: ["session-1"]
 right: []
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2252 filtered out; finished in 0.10s
```

**NC-B3a - the mailbox check in the actor's re-decision.** `if !command_rx.is_empty()` becomes `if !command_rx.is_empty() && false`, i.e. exactly the race the review described.

```
test ...idle_reap::a_prompt_queued_behind_the_reap_wins_and_the_session_is_kept ... FAILED
panicked at .../actor/tests/idle_reap.rs:223:18:
the prompt's responder must not be dropped by an exiting actor: RecvError(())
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2252 filtered out; finished in 0.09s
```

**NC-B3b - the mid-turn refusal.** The active-turn arm is replaced by the unconditional `Unload` arm's body (cancel notification, `Cancelled` interactions, exit after prompt).

```
test ...idle_reap::a_mid_turn_reap_is_refused_without_cancelling_the_turn ... FAILED
panicked at .../actor/tests/idle_reap.rs:314:13:
assertion `left == right` failed: a running turn must retain the session
  left: Unloading
 right: Retained(ActiveTurn)
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2252 filtered out; finished in 0.10s
```

**NC-B4 - the process-group kill.** The exit sequence goes back to `drop(self.child)` for every disposition except `Stop`.

```
test ...idle_reap::a_conditional_unload_kills_the_agents_whole_process_group_and_stays_resumable ... FAILED
panicked at .../actor/tests/idle_reap.rs:140:13:
the reap must kill the vendor CLI grandchild too - kill_on_drop reaches the direct child only, and the grandchild is 64% of the memory this feature exists to reclaim
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2252 filtered out; finished in 5.23s
```

**NC-fail-closed - `Undetermined` never reaps.** The failed background-work read returns `Quiescent` instead of `Undetermined`.

```
test ...quiescence::a_failed_durable_read_is_undetermined_and_never_reaps ... FAILED
panicked at .../reaper/tests/quiescence.rs:294:5:
  left: Quiescent
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2252 filtered out; finished in 0.09s
```

**NC-refusal - the manager honors the actor's no.** `Some(Retained(reason)) => return Ok(Some(reason))` becomes `Some(Retained(_)) => {}`.

This one is the reason `IdleSweepOutcome` gained a `retained` list. The first version of the refusal test passed under this mutation, because a manager that ignores the refusal falls through to the retirement wait, times out, and records a FAILED reap: both paths leave the session live and both leave `reaped` empty, so nothing told them apart. With the list, the control is real:

```
test ...quiescence::an_actor_that_refuses_the_conditional_unload_keeps_its_session ... FAILED
panicked at .../reaper/tests/quiescence.rs:239:5:
assertion `left == right` failed: a refusal is an answer, not a failure
  left: []
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2252 filtered out; finished in 2.10s
```

Every mutation was reverted and the suite observed green again, quoted above.

### CI

The previous CI red is stale. `Cargo check & test` finished green on attempt 2 at `9a484d0163` (3160/3160), and the one failure on attempt 1 was `proliferate::diagnostics_collector::process::tests::rv_2_05_child_exit_during_launch_is_never_graceful_shutdown`, a timing-sensitive desktop process-launch race in code this PR does not touch. CI for the review fixes is green on the current head `6943ead9e2`: 35 checks pass, 3 skipping, none failing, `Cargo check & test` included.

## Observability

New. All fields are ids, durations and bounded classes; no prompts, paths, values or credentials.

| Event | Fields |
| --- | --- |
| `idle_reaper_started` / `idle_reaper_disabled` | threshold seconds, sweep interval seconds |
| `result_class = "reaped"` | session id, idle seconds, threshold seconds |
| `result_class = "reap_failed"` | session id, idle seconds, `failure_code = "nonterminal_unload_failed"`, error |
| `result_class = "reap_retained"` (debug) | session id, refusal reason as a bounded string |
| `result_class = "awaiting_interaction_held"` (debug) | count held on THIS sweep |
| `result_class = "background_work_held"` (debug) | count held on THIS sweep |
| clock reset (debug) | session id, blocking verdict as a bounded string |
| threshold override ignored | `failure_code = "idle_reap_threshold_unparseable"` |

Both `_held` counters are per-sweep gauges, not cumulative totals: a session nobody ever answers is counted again on every sweep for as long as it is held. Only one of them fires per held session, because the predicate short-circuits and `AwaitingInteraction` is tested first, which is deliberate (the founder ruling makes it the never-reap case) and is now documented rather than reordered.

`background_work_held` exists because `BackgroundWorkOptions::default()` sets `stale_after: None`, so `background_work/claude/watch.rs`'s expiry never fires in production and a tracker whose output file never yields leaves a `pending` row forever. That is a second permanent unreapable class. Whether the default should change is a founder call, not an implementer's, so this PR only makes it visible.

## Readiness

Draft on purpose. Not merged, not marked ready; a separate reviewer takes it.

## Verification

- `cargo test -p anyharness-lib --lib -- reaper idle_reap` green, 22/22.
- `cargo test -p anyharness-lib --lib` green, 2250/2250.
- Six behavioural negative controls for the review fixes, each observed red then green, quoted literally above, on top of the nine from the first round.
- `check_docs.py`, `check_max_lines.py`, `check_anyharness_boundaries.py`, `check_design_attribution.py`, `check_session_mutation_admission.py` pass.
- Not yet run: the live dev-machine check the spec's acceptance section names (start several sessions, watch the agent processes exit about two minutes later, resume one and confirm continuity).


pablonyx (commented):
## Review: idle-session reaper

Reviewer of record. I read the diff and the surrounding runtime rather than the body, and I did not run cargo (a build owned by another session is on this box). Everything below is quoted from source at `9a484d016372947f2181f529c2ca4210171e9e1b` or from CI logs.

**Verdict: do not merge yet.** Four blocking findings, three of which are correctness regressions in paths the body claims are safe, and one of which is that the central memory claim is still unverified. The core design is good and most of the body checks out; the gaps are all at the edges where "retire the actor" meets code that assumed an actor is always there.

---

## What I verified and found correct

- **`choose_startup_strategy` really does pick `LoadNative`.** For the ordinary reapable case (`native_session_id` present, `last_prompt_at` present, not a fork child) `launch_policy.rs` returns `Ok(SessionStartupStrategy::LoadNative(native_session_id))`, pinned by `loads_claude_when_last_prompt_was_recorded`. The correction to the frozen spec is right, and the zero-turn non-fork case genuinely does fall to `ResumeSeqFreshNative`, which is also resumable. See B1 for the case the correction misses.
- **The `tracker_count()` premise is true.** `background_work/mod.rs` only ever removes tracker entries in `shutdown()` (`for (_, handle) in self.trackers.drain()`); a completed tracker's `JoinHandle` stays in the map, so `tracker_count()` never returns to zero. Choosing `list_pending_background_work` over it is correct.
- **Retirement is genuinely non-terminal.** `shutdown/persist.rs` on `ActorExitDisposition::Unload` writes `update_status(session_id, "idle", now)`, emits no `session_ended`, and leaves `closed_at`/`dismissed_at`/`native_session_id` untouched.
- **The env contract matches the code exactly.** `from_env` returns the default on absent, `with_threshold(Duration::from_secs(n))` on parse success (so `0` disables through `with_threshold`), and warns + default on `Err(_)`. `sweep_interval()` is `(threshold / 4).min(MAX_SWEEP_INTERVAL).max(1ms)` with `MAX_SWEEP_INTERVAL = 15s`, so the "threshold/4 capped at 15s" claim holds. The `env-vars.yaml` entry and the `acp.md` section are accurate.
- **The ledger does not leak.** `self.observations.retain(|session_id, _| seen.contains(session_id))` drops entries for sessions that left the map, and a restarted session gets a fresh `LiveSessionExecutionSnapshot::new(...)` timestamp, so the marker comparison restarts the clock rather than inheriting a stale one.
- **The doubled `AwaitingInteraction` guard is not redundant, and the useful half is the list.** `set_execution_phase(Idle)` and `add_pending_interaction` are independent writers, so `phase == Idle` with a non-empty `pending_interactions` is reachable; `!snapshot.pending_interactions.is_empty()` is what catches it. The scoping-out reasoning holds too: `shutdown/cleanup.rs` maps `ActorExitDisposition::Unload => Resolution::Cancelled`, so reaping a parked session really does cancel the request and the turn behind it.
- **The nine negative controls are all mechanically consistent** with the code, and none is vacuous or a restatement of another. NC4 and NC9 both fail `the_verdict_names_the_exact_condition_that_blocked_a_reap`, but they mutate different predicate arms and each also fails a distinct behavioural test.
- **CI is green and the body's CI section is stale.** Attempt 1 of run `32452506841` failed exactly as described:

  ```
  FAIL [   0.102s] (2693/3160) proliferate diagnostics_collector::process::tests::rv_2_05_child_exit_during_launch_is_never_graceful_shutdown
  panicked at apps/desktop/src-tauri/src/diagnostics_collector/process_tests.rs:149:5:
    left: EndpointUnavailable
   right: ChildExited
  Summary [ 148.976s] 3160 tests run: 3159 passed (1 slow), 1 failed, 8 skipped
  ```

  Attempt 2 of the same run, same SHA, passed everything:

  ```
  Summary [ 139.787s] 3160 tests run: 3160 passed (1 slow), 8 skipped
  ```

  All 13 reaper tests are `PASS` in both attempts, e.g. `PASS [ 0.283s] (2186/3160) anyharness-lib live::sessions::manager::reaper::tests::the_spawned_reaper_retires_an_idle_session_on_its_own_cadence`. The flake claim is sound on mechanism, not just assertion: the test spawns `/usr/bin/false`, sleeps `Duration::from_millis(50)`, and asserts the launch error has been reclassified to `ChildExited` by then. It is a different crate with no code shared with this change. I sampled 60 recent failed CI runs and did not find the same failure elsewhere, so there is no cross-PR corroboration, but a same-SHA rerun passing 3160/3160 settles it. Please update the CI section of the body.

---

## Blocking

### B1. A zero-turn Claude fork child that gets reaped can never start again

This is the answer to "can a reaped session come back", and for one class of session it is no.

`choose_fork_child_strategy` in `domains/sessions/runtime/launch_policy.rs`:

```rust
if fork_id_is_process_local && !facts.has_last_prompt_at {
    anyhow::bail!(
        "process-local zero-turn fork recovery requires an exact-prefix recovery proof"
    );
}
```

`fork_id_is_process_local` is `agent_kind == AgentKind::Claude.as_str()`, and `has_last_prompt_at` comes straight from `record.last_prompt_at.is_some()` in `startup_facts.rs`. Fork children are inserted with `last_prompt_at: None` (`runtime/fork/mod.rs:307`), and both fork paths register the child live through `start_live_session`, whose actor finalizes to `SessionExecutionPhase::Idle` (`actor/startup.rs`, both the process-local `execution.phase = SessionExecutionPhase::Idle` branch and the `set_execution_phase(SessionExecutionPhase::Idle)` branch).

So: fork a Claude session, do not prompt the child, walk away for two minutes. The child is phase `Idle`, not busy, no pending interactions, no background-work rows, empty queue. `Quiescent`. It gets reaped. The next prompt calls `choose_session_startup_strategy`, hits the bail above, and the fork is gone. The repo already pins this refusal in `refuses_zero_turn_process_local_fork_child_with_native_id`.

Before this PR the child stayed in the live map forever, so cold recovery was only reachable on a process restart. The reaper makes it a routine two-minute timer. The predicate needs a fork-child carve-out (or the reap needs to be gated on the same fact the launch policy gates on), and it needs a test.

### B2. Cowork wake schedules do not survive a reaped parent, and the body's mechanism for that claim is the wrong subsystem

The body says:

> `CompletionDeliveryWorker` then cold-starts the parent through `activate_durable_prompt_consumer` / `ensure_live_session_handle`, retrying on a 1s lease cadence until the wake is durably visible.

That is true for **subagent** completions and only for those. `persist_terminal_turn_in_tx` in `store/completion_deliveries.rs` inserts the outbox row under:

```sql
FROM session_links
WHERE relation = 'subagent' AND child_session_id = ?1
  AND closed_at IS NULL
```

`insert_completion_and_consume_schedule` — the function the body and the acceptance test are actually about — has exactly one production caller, `deliver_cowork_coding_completion` in `domains/cowork/runtime.rs`, on `relation = 'cowork_coding_session'` links. `schedule_wake` likewise has only cowork callers. Those links never produce a `session_link_completion_deliveries` row, so `CompletionDeliveryWorker` never sees them. What that path does instead is:

```rust
if let (Some(record), Some(handle)) = (
    inserted.wake_prompt.as_ref(),
    acp_manager.get_handle(&link.parent_session_id).await,
) {
    let _ = handle.send_queued_prompt(prompt_payload, record.seq).await
```

`get_handle`, not `ensure_live_session_handle`. If the parent has been reaped this is `None` and the branch is skipped silently. The wake prompt sits in `session_pending_prompts`, and I could not find anything that scans for stranded pending prompts and cold-starts their session. The parent picks it up whenever a human next opens or prompts it, which may be never.

The acceptance test does not catch this because it does not actually test a wake. It asserts the durable row exists, then hand-registers a live session:

```rust
"the wake is durable in the parent's queue, ready for cold-start drain"
...
register_live_session(&manager, "session-1", SessionExecutionPhase::Idle).await;
```

Nothing in `a_pending_wake_schedule_survives_reaping_and_still_wakes_the_parent` demonstrates the parent being woken. The test name overstates what it proves.

This is a functional regression: before the reaper the cowork parent stayed live and the wake always landed.

### B3. The quiescence check is not atomic with the unload, and `Unload` is unconditional

Answering the race question directly: "from the instant the wake row exists the parent is not reapable" is **not** enforced by any ordering guarantee. It is enforced by the order of two statements in one function, with no lock, no transaction, and no re-check. `sweep` reads the snapshot, runs two SQLite queries, and then calls `unload_session_nonterminal`. Anything that arrives in between is not seen.

That would be tolerable if `Unload` were advisory. It is not. In `actor/turn/active.rs`:

```rust
Some(SessionCommand::Unload { respond_to }) => {
    self.resolve_pending_interactions(Resolution::Cancelled).await;
    let _ = self.conn
        .send_notification(acp::schema::CancelNotification::new(self.native_session_id.clone()));
```

and the idle loop's `select_idle_work` is `biased;` with `command_rx.recv()` first and `IdleWork::DrainQueuedPrompt` second, so a queued Unload beats a durable prompt that is ready to drain.

```mermaid
sequenceDiagram
    participant R as reaper sweep
    participant U as user / wake
    participant A as actor mailbox
    R->>R: snapshot + 2 durable reads -> Quiescent
    U->>A: Prompt
    R->>A: Unload
    A->>A: FIFO: Prompt starts a turn
    A->>A: Unload arrives mid-turn -> CancelNotification, work discarded
```

Two user-visible outcomes:

1. Prompt lands before Unload: the turn starts and is then cancelled by the reaper. The body frames the price of a reap as "a cold start"; here it is the turn.
2. Prompt lands after the actor dropped `command_rx`: `prompt.rs` maps `LiveSessionCommandError::ActorUnavailable` to `SendPromptError::Internal(anyhow::anyhow!("session actor channel closed"))` and cleans up the attachments. No retry, no cold-start fallback. The user sees an internal error.

Forks are the worse case, because forking requires the parent to be `Idle` (`fork/mod.rs` checks `execution.phase != SessionExecutionPhase::Idle`) — the exact population the reaper targets. A fork is several round trips against a parent whose `updated_at` none of those commands bump, and `fork/errors.rs` maps `ActorUnavailable` to `ForkOperationPhase::NativeOutcomeUnknown`.

The codebase already has the pattern for this: `SessionCommand::CancelTurnIfActive` exists precisely so a stale decision cannot cancel newer work. There is no `UnloadIfStillIdle`. The reaper should either use a conditional unload the actor evaluates serially on its own loop (where `next_pending_prompt_for_drain().is_some()` and busy are both authoritative), or re-verify the verdict after the actor has stopped accepting new work.

### B4. Nothing in the reap path signals the agent's process group, and the reclaim was never observed

The whole justification is memory reclaim, and the one acceptance item that would confirm it is explicitly not done:

> Not yet run: the live dev-machine check the spec's acceptance section names (start several sessions, watch the agent processes exit about two minutes later, resume one and confirm continuity).

That matters more than a normal unrun manual check, because `Unload` takes the branch that does not kill the group. In `actor/run.rs`:

```rust
if let Some(respond_to) = self.pending_stop_response.take() {
    let kills = self.kill_process_group_and_reap().await;
    let _ = respond_to.send(Ok(kills));
} else {
    drop(self.child);
}
```

`pending_stop_response` is set only by `SessionCommand::Stop`, so a reap always takes `drop(self.child)`. That fires `kill_on_drop(true)` from `driver/process.rs`, whose own comment says it is "the crash backstop for the **direct child** only; the group is what `stop_and_await` escalates against." The measurement the PR rests on says each session is two processes and that the vendor `claude` CLI is 64% of the cost while the ACP adapter is 22% — the adapter is the direct child, the vendor CLI is its child. SIGKILLing the adapter probably takes the CLI down via stdin EOF, but "probably" is not what a 240 MB-per-session claim should rest on, and it is unobserved.

Please run the acceptance check and paste `ps`/RSS before and after. If the grandchild survives, the reaper costs memory (a leaked orphan plus a fresh cold start) instead of saving it.

---

## Non-blocking

1. **`IdleReapPolicy::from_env()` has no test at all.** NC8 mutates `with_threshold`, not `from_env`. Nothing exercises "whole seconds", "`0` disables via the env var", or "unparseable keeps the default" — the entire documented config contract. Factor the parse into `fn parse(raw: Option<&OsStr>) -> Self` and test that; the env read stays a one-liner.
2. **`Undetermined` is untested and uncontrolled.** Both `Err(error) => ... return IdleReapVerdict::Undetermined` arms have no test and no negative control. Flipping either to `Err(_) => {}` (fail open) passes all 13 tests. This is the safety property the body advertises hardest ("the predicate fails closed") and it is the one with no coverage.
3. **Abandoned background work makes a session permanently unreapable, invisibly.** `BackgroundWorkOptions::default()` sets `stale_after: None` and `actor/startup.rs` constructs the registry with exactly that default, so the `stale_after` expiry in `background_work/claude/watch.rs` is dead code in production. A tracker whose output file never yields a result leaves a `state = 'pending'` row forever, and the session is never reaped. Unlike the `AwaitingInteraction` hold, this one is not counted anywhere, so the second permanent-leak class is unmeasurable. Consider counting `BackgroundWork` and `QueuedPrompts` holds per sweep too.
4. **`awaiting_interaction_held` does not mean what the docs say.** Both the body and `acp.md` describe it as "sessions held back **solely** because a human has not answered". The check runs first, before busy / background work / queue / threshold, so it counts every session with a pending interaction, including ones mid-turn, ones idle for three seconds, and ones that were not otherwise reapable. As a proxy for the abandoned-prompt leak it is inflated by every ordinary in-flight permission prompt, re-counted every 15 seconds. Either move the check after the cheap disqualifiers or reword the docs.
5. **The "real `unload_session_nonterminal` path" claim is overstated.** The scripted consumer in the test replies `Ok(())` to `Unload` and calls `remove_session`; the actor's exit sequence (`finalize_exit` -> `persist_exit_disposition`) never runs. So `assert_eq!(record.status, "idle")` in `reaped_session_stays_resumable_with_its_native_session_id` passes because the seeded row was already `"idle"`, not because the reap preserved it, and NC5's mutation is an added durable write the reaper would never make. The behaviour is in fact correct (I checked `persist_exit_disposition` writes `"idle"` on `Unload`), but the test does not prove it and the body says it does.
6. **`updated_at` does not cover every form of activity.** The writers are `set_execution_phase`, the four pending-interaction mutators, and `mark_activity_at` (called only from `actor/notifications/dispatch.rs:39`, gated on `event.updates_session_activity_at()`). Not covered: model/mode/thinking-level changes on an idle session, `RunDomainOp`, `CallAgentExtMethod`, and the fork lifecycle commands. Turn traffic is covered because the phase flips both ways. A user actively reconfiguring an idle session for two minutes still gets reaped. Low harm on its own, but it is the mechanism behind the fork window in B3.
7. **The size rationale for putting the clock in the sweep task does not hold as stated.** On this branch `handle.rs` is 586 lines against `MAX_LINES = 600` with no ratchet entry, so there are 14 lines of headroom. `manager/mod.rs` is the file at 598. The ledger design is defensible on its own merits (it derives continuity from an existing marker instead of adding a reset call site to every activity source), so it does not need the size argument, but the argument as given is wrong.
8. **The reaper bypasses the access gate.** `SessionRuntime::unload_live_session_nonterminal` calls `assert_can_mutate_for_session` first; the reaper calls `acp_manager.unload_session_nonterminal` directly, so it will retire actors in `FrozenForHandoff` / `RepairBlocked` / archived workspaces. Non-destructive as far as I can tell (the durable row survives and `assert_can_start_live_session` already blocks the restart), but it is a deliberate-looking omission with no comment.
9. **Scoping question.** `spawn_idle_reaper` is wired unconditionally in `wire_live_sessions`, so it is on by default in every AnyHarness runtime including cloud sandboxes, and `env-vars.yaml` tags the variable `production`. The spec's out-of-scope list says "Cloud and sandbox session culling, owned by the Cloud Culling program." Worth confirming with Pablo that the cloud plane is intended to be in from day one, since the measurement was on a fleet host.
10. **One coverage gap in the control set that is not a control problem.** No listed control mutates the `NotIdle` phase check, but `a_session_that_never_settles_is_not_reaped` would catch it, so the behaviour is covered even though the enumeration says every test carries a control. Just noting the enumeration is off by one relative to the 13 tests.

---

## Summary

The predicate, the ledger, the non-terminal disposition, the config surface, the docs, and the control set are all in good shape, and the resume claim for the ordinary case is exactly right. What is missing is the set of places that quietly assumed an actor never goes away: zero-turn fork children (B1), cowork wakes (B2), and everything racing the unload (B3). B4 is the one that decides whether the feature is worth its risk at all, and it is a twenty-minute manual check.

pablonyx:
## Review 4990291421: all four blocking findings fixed, and all four were right

Fresh fixer here, not the implementer. I checked every finding against the source before touching anything, expecting to dispute at least one. I could not: B1 through B4 all reproduce. Pushed as `1ac587bb90`, `4d4b4146de`, `b68dd0b151`, `6943ead9e2`, rebased onto current `main` (`fe6780e006`). Still draft, still unmerged.

### B1. Zero-turn Claude fork child

Fixed in the general form you suggested rather than as a claude-specific case. The predicate now asks the launch policy itself: `IdleReapDurable::session_can_relaunch_from_cold` loads the record and returns whether `choose_session_startup_strategy` succeeds against the same durable rows the next prompt will read, and the reaper refuses to reap anything it errors for. Keying on the policy rather than on "is this a Claude fork child" means the carve-out cannot drift from the policy, and it costs nothing extra to be right about shapes nobody has thought of yet. It is the last check in the predicate, after the cheap ones.

Pinned by `a_zero_turn_claude_fork_child_is_never_reaped` (with a positive control in the same test: give the child a `last_prompt_at` and it becomes reapable) and by `a_zero_turn_session_that_is_not_a_fork_child_is_still_reaped`, which proves the carve-out did not swallow the ordinary zero-turn case.

Negative control, mutating `session_can_relaunch_from_cold` to still call the policy and then return `Ok(true)`:

```
test ...relaunch::a_zero_turn_claude_fork_child_is_never_reaped ... FAILED
panicked at .../reaper/tests/relaunch.rs:29:5:
assertion `left == right` failed: reaping a session no startup strategy accepts would end it permanently
  left: ["child-1"]
 right: []
```

### B2. Cowork wakes, and the body's wrong mechanism

Your reading of the code is exactly right, and the body was describing the `subagent` path while claiming it for all relations. The split, stated plainly: `subagent` completions write a `session_link_completion_deliveries` row in `persist_terminal_turn_in_tx` (`WHERE relation = 'subagent'`) and `CompletionDeliveryWorker` cold-starts the parent through `activate_durable_prompt_consumer`, so a reaped `subagent` parent really is woken. `cowork_coding_session` has no outbox row at all; `deliver_cowork_coding_completion` reaches for `get_handle` and drops the send when it returns `None`, and nothing scans for stranded pending prompts.

Took the minimal honest fix you named: a session that parents a link with a pending wake schedule on any relation other than `subagent` is not reapable (`has_live_delivery_wake_schedule`, verdict `PendingWake`). Unknown future relations are held back by the same test, which is the fail-safe direction.

The dishonest acceptance test is gone. In its place: `a_cowork_parent_expecting_a_wake_is_not_reaped` (held back, and reapable again once the wake is delivered and drained) and `a_subagent_parent_is_reapable_and_its_wake_still_becomes_a_durable_prompt`, which carries an explicit "what this test does NOT prove" note, since the worker's cold start belongs to the worker's own suite. The PR body and the delivery spec both now carry the corrected mechanism with the correction marked as a correction.

Negative control on the hold:

```
test ...wake::a_cowork_parent_expecting_a_wake_is_not_reaped ... FAILED
panicked at .../reaper/tests/wake.rs:120:5:
assertion `left == right` failed: reaping a cowork parent strands the wake its child is about to send
  left: ["session-1"]
 right: []
```

### B3. The unload is now conditional, decided by the actor

Implemented in full, no narrowing needed. New command `SessionCommand::UnloadIfIdle` answering `ConditionalUnloadOutcome::{Unloading, Retained(reason)}`, with `reason` a bounded class (`active_turn`, `busy`, `pending_interaction`, `queued_prompt`, `mailbox_not_empty`, `replay_session`). The actor re-evaluates on its own loop where the mailbox, the durable queue head, the busy flag and the pending interactions are all authoritative, and refuses if anything arrived since the sweep looked. A refusal resets the idle clock. The mid-turn arm answers `Retained(active_turn)` and does nothing else: no `CancelNotification`, no exit. Replay sessions answer `Retained(replay_session)`.

The mailbox check is the one that closes your race: the channel is FIFO, so anything sitting behind the unload was sent after it, therefore after the sweep's observation.

Residual window, stated exactly rather than papered over: between the actor answering `Unloading` and its handle leaving the live map, a caller that already fetched the handle can still send into a closing mailbox and get `ResponseDropped`. That window is not new (every `unload_live_session_nonterminal` caller has always had it) and closing it needs handle lookup and send to be atomic with respect to retirement, which is a manager-wide change. It is named in `acp.md` and in the delivery spec.

Two negative controls. The racing prompt, mutating the mailbox check to `if !command_rx.is_empty() && false`:

```
test ...idle_reap::a_prompt_queued_behind_the_reap_wins_and_the_session_is_kept ... FAILED
panicked at .../actor/tests/idle_reap.rs:223:18:
the prompt's responder must not be dropped by an exiting actor: RecvError(())
```

That assertion is deliberately ordered first so the control fails on the user-visible harm rather than on the bookkeeping.

The mid-turn case, replacing the refusal arm with the unconditional `Unload` arm's body:

```
test ...idle_reap::a_mid_turn_reap_is_refused_without_cancelling_the_turn ... FAILED
panicked at .../actor/tests/idle_reap.rs:314:13:
assertion `left == right` failed: a running turn must retain the session
  left: Unloading
 right: Retained(ActiveTurn)
```

### B4. The reap kills the tree, and the reclaim is now observed

The exit sequence signals the process GROUP on every disposition, not only `Stop`. Your hedge about stdin EOF taking the CLI down does not hold: with `drop(self.child)` restored, the adapter analogue dies and the grandchild survives, which is what the control below shows.

One thing I had to add beyond the finding. A child that is not its own group leader (no production spawn site, but the test harnesses do it) would get a group signal that reaches nothing and then a 20s wait for a process nobody told to die, which broke 16 unrelated tests with `non-terminal actor unload timed out`. The exit sequence now checks `getpgid(pid) == pid` and falls back to the direct SIGKILL that `kill_on_drop` would have delivered. Bounded either way.

The test is a real `SessionActor` over a real ACP duplex with a real spawned tree, where the "vendor CLI" is a grandchild that outlives its parent's stdin:

```
test ...idle_reap::a_conditional_unload_kills_the_agents_whole_process_group_and_stays_resumable ... FAILED
panicked at .../actor/tests/idle_reap.rs:140:13:
the reap must kill the vendor CLI grandchild too - kill_on_drop reaches the direct child only, and the grandchild is 64% of the memory this feature exists to reclaim
```

The same test also answers your non-blocking 5: it seeds the durable row as `running` and asserts the reap moves it to `idle` with `native_session_id` intact, so the status assertion now discriminates instead of re-reading the seed.

### Non-blocking

1. **`from_env` untested.** Done as you specified. `IdleReapPolicy::parse(Option<&OsStr>)` is factored out, `from_env` is the one-line read on top, and `the_env_contract_defaults_absent_and_unparseable_values` covers absent, `0`, whole seconds, surrounding whitespace and garbage.
2. **`Undetermined` untested.** Done. `a_failed_durable_read_is_undetermined_and_never_reaps` drops the `session_background_work` table underneath a live session. Control (failed read returns `Quiescent`): `left: Quiescent`, and the test goes red.
3. **Background-work holds uncounted.** `background_work_held` added, and the `stale_after: None` dead-code path is named in the spec and the body as the second permanent unreapable class. I did not count `QueuedPrompts` holds: that class drains itself, so a gauge over it measures throughput rather than a leak. Whether `BackgroundWorkOptions::default()` should stop setting `stale_after: None` is a founder call and I have flagged it rather than made it.
4. **`awaiting_interaction_held` overcounts.** Documented rather than reordered, and I want to be explicit about why: the founder ruling makes `AwaitingInteraction` the never-reap case, so it has to be the first thing the predicate proves; moving it after the cheap disqualifiers would make the never-reap guarantee depend on check order. Both counters are now described as per-sweep gauges (re-counted every sweep, never summed) in `acp.md`, the spec and the body, and the docs no longer say "solely".
5. **"Real unload path" overstated.** Reworded, and the actor-level test above now proves the durable transition for real. The tier-2 suite's own description says plainly that its actors are scripted consumers.
6. **`updated_at` does not cover every activity.** Confirmed, not fixed, recorded in the spec as a known gap with the uncovered list (config changes, `RunDomainOp`, `CallAgentExtMethod`, fork lifecycle). The conditional unload takes the sting out of it: those commands sitting in the mailbox now refuse the reap, so the residual harm is one cold start for someone who reconfigures for two minutes without prompting. Widening the marker means a reset call site in every activity source, which is a bigger change than this feature earns.
7. **The size rationale.** You are right that `handle.rs` is 586 and `manager/mod.rs` is 598, and I checked: that argument appears nowhere in the spec or the body, only in the implementer's report, so there was nothing to correct in the artifacts. The clock's placement stands on the merits you granted it. Related, since it moved: `manager/mod.rs` is now 551 (the non-terminal unload moved to `manager/unload.rs`, shared with the conditional variant), `handle.rs` is 567, `run.rs` is 524 (the process reap moved to `actor/exit_kill.rs`), and `reaper/tests.rs` became a four-module directory. All under the cap with no new ratchet entries.
8. **Access gate bypass.** Comment added at `unload_session_if_still_idle` explaining the deliberate choice: `assert_can_mutate_for_session` exists to stop user-visible mutations of a frozen, repair-blocked or archived workspace, a reap changes no durable state at all, and holding actors alive in exactly those workspaces would leak the memory the feature exists to reclaim. `assert_can_start_live_session` still governs the restart.
9. **Cloud scoping.** I could not gate this the way the spec's out-of-scope line implies, and I do not think it is gateable as written. There is no plane discriminator anywhere in the runtime binary: no `is_cloud`, no `ANYHARNESS_CLOUD`, nothing in `wire_live_sessions` or `app/mod.rs` that distinguishes a desktop runtime from a cloud sandbox. The same binary runs in both, so the only available lever is the env var (set `ANYHARNESS_IDLE_SESSION_REAP_SECONDS=0` in the cloud task definition). Inventing a plane flag to satisfy the scoping line would be a bigger and more speculative change than the feature. Flagging for Pablo rather than deciding: either the cloud plane is in from day one, or the cloud task definition sets the variable to `0`.

### Verification

`cargo test -p anyharness-lib --lib -- reaper idle_reap`: `test result: ok. 22 passed; 0 failed; 0 ignored; 0 measured; 2231 filtered out; finished in 5.18s`.

Full `cargo test -p anyharness-lib --lib`: `test result: ok. 2250 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 159.11s`.

Seven negative controls for this round (six above plus one on the manager honoring a refusal), each mutated behaviourally with signatures intact, observed red, reverted, observed green. The refusal control is worth calling out because it caught a weak test of my own: the first version passed under the mutation, since a manager that ignores `Retained` falls through to the retirement wait, times out, and records a failed reap, which also leaves the session live and `reaped` empty. `IdleSweepOutcome` now carries the sessions whose actor answered no, so the two outcomes are distinguishable.

Checkers all pass. The stale CI-red note in the body is replaced: `Cargo check & test` was green 3160/3160 on attempt 2 at the reviewed head, and CI for these fixes is running now.

pablonyx:
## Live dev-machine acceptance run (the spec's open Acceptance item), executed on `6943ead9e2`

The spec listed "start several sessions, observe the agent processes exit about two minutes later, then resume one" as not yet run. It has now been run against a real runtime built from this branch, with a real Claude agent process, on macOS.

Setup: `target/debug/anyharness serve --port 8639` from this branch with `ANYHARNESS_IDLE_SESSION_REAP_SECONDS=60`, a scratch runtime home, and a scratch git workspace. Startup log:

```
INFO reaper: idle session reaper started threshold_seconds=60 sweep_interval_seconds=15 result_class="idle_reaper_started"
```

### 1. The grandchild really is where the memory lives

Process tree after one completed turn, `pid ppid pgid rss_kb comm`:

```
36831 27377 36831  34912 node                                   <- ACP adapter, direct child, group leader
36838 36831 36831 395296 .../agents/claude/native/claude         <- vendor CLI, grandchild
37092 36838 36831  10752 .../node_repl                           <- great-grandchild
```

431 MB in the session tree, of which the direct child is 34.9 MB. The vendor CLI grandchild alone is **92%** of it. A direct-child-only kill reclaims 8%. Note `pgid == pid` on the adapter, which is the precondition `child_leads_its_own_group` checks.

### 2. The reap fires and takes the whole group

```
2026-08-21T08:10:19.464203Z INFO reaper: idle session actor retired to reclaim its agent processes
  session_id=d647696e-9b94-42d8-a120-6da7dc399efe idle_seconds=74 threshold_seconds=60 result_class="reaped"
```

Liveness of the three pre-reap pids immediately afterwards:

```
GONE  36831
GONE  36838
GONE  37092
```

All three, including the 395 MB grandchild and the great-grandchild. Runtime RSS itself went 51.3 MB to 41.2 MB. One sweep, no `reap_failed`, no `retained`.

### 3. The reaped session is still a working session

A prompt sent 3m30s after the reap returned `HTTP 200`, and the session event log is continuous across the reap with no gap or error:

```
seq  6  08:08:56  turn_started     <- first turn
seq 15  08:08:59  turn_ended
                  [reap at 08:10:19]
seq 21  08:13:51  turn_started     <- second turn, cold-relaunched actor
seq 28  08:13:54  turn_ended
```

`nativeSessionId` is byte-identical before and after (`7886113b-4bfa-43fa-9a53-b4a948b5eda4`), so the native conversation resumed rather than restarted, and a fresh process tree was spawned on demand (`50629/50632/50696`). The reap emits no session event, so it is invisible in the transcript.

### What this does and does not prove

Proves, on a real machine with a real vendor CLI: the threshold fires, the process-group kill reclaims the grandchild rather than only the adapter, ~431 MB comes back per idle session, and the session survives as a resumable durable record whose next prompt cold-starts transparently.

Does not prove: multi-session concurrent sweeps (single session here, so the serial-sweep and 10s-vs-15s retirement-deadline concern raised in review is untouched by this run), Windows behaviour (`child_leads_its_own_group` is hardcoded false there), or the cowork wake path.

pablonyx:
## Full local profile run on the rebased head `cda16756da`

Rebased onto current `main` (which now carries #2134, #2161, #2162) so this exercises the combination that will actually ship, not the branch in isolation. `make dev PROFILE=reaperlive HEADLESS=1` from a clean worktree, no env overrides, so the reaper ran on its **shipping default**:

```
INFO reaper: idle session reaper started threshold_seconds=120 sweep_interval_seconds=15 result_class="idle_reaper_started"
```

`make setup` created the profile Postgres against the native brew service with no Docker, which is #2161's probe doing its job inside a real launch. Boot was one cargo build (`Finished dev profile in 1m 21s`) followed by a second invocation that no-opped in `0.41s`, which is #2134's dedup.

### Three concurrent sessions, one sweep

Three Claude sessions were created and prompted concurrently, all idle by 08:24:11Z. Pre-reap, `pid ppid pgid rss_kb comm`:

```
62507 59633 62507  50320 node          63188 59633 63188  86928 node          63366 59633 63366  69888 node
62508 62507 62507 388336 claude        63197 63188 63188 421152 claude        63374 63366 63366 411968 claude
62641 62508 62507  10960 node_repl     63281 63197 63188  13328 node_repl      63407 63374 63366  12720 node_repl
```

Three distinct process groups, each led by its own adapter. 9 processes, **1431.2 MB**, plus a runtime at 267.5 MB.

The sweep took all three:

```
08:26:19.332  session=939f448b idle=120s  class=reaped
08:26:19.804  session=4e2eb242 idle=135s  class=reaped
08:26:20.384  session=77ff8ffe idle=120s  class=reaped
```

Afterwards: **0 of 9 processes alive**, runtime RSS 267.5 MB to 65.6 MB, system swap-in-use down 416 MB. Roughly 1.6 GB returned.

### On the review's open should-fixes

- **Serial sweep / false `reap_failed`**: confirmed structurally, but not live here. The three reaps completed in **1.05s total, ~500ms each**, an order of magnitude inside `UNLOAD_EXIT_TIMEOUT`. Zero `reap_failed` and zero `reap_retained` in the whole run. The concern needs an agent that is slow to die on SIGTERM; Claude is not one. Latent, not active.
- **`ActorUnavailable` window**: `grep -cE 'ActorUnavailable|session actor channel closed'` over the entire profile log returns **0**, including across three concurrent post-reap resumes. Not reproduced, which does not disprove the race, only bounds how easy it is to hit.
- Total non-agent `ERROR` lines for the whole run: **0**.

### Resume

All three were prompted again 25s after the reap, concurrently. All returned `HTTP 200`, all three `nativeSessionId` values are byte-identical across the reap, each session shows exactly 2 `turn_started` events, and 9 processes respawned.

```
MATCH 939f448b -> 3d428416-9205-493b-9b43-533cfedb7872
MATCH 4e2eb242 -> 05f8dea7-0fa4-49b0-ab97-2b122a25f008
MATCH 77ff8ffe -> d757001b-0a29-4003-bc98-31f45fb9bce3
```

Still unproven: Windows (`child_leads_its_own_group` is hardcoded false there), the cowork wake path, and reaping under a TERM-slow agent.

PR #2161, opened 2026-08-21
perf(dev): use a local postgres or redis that is already listening

pablonyx:
## Why

`make dev PROFILE=x` cannot be the whole launch while `USE_EXISTING_POSTGRES ?= 0`. On a host running native Postgres and Redis (brew services, apt packages), that default makes `make` bring up a Docker Postgres on top of one already bound to 5432. That single default is the main reason a shell wrapper has to sit in front of `make` and prefix every invocation with `USE_EXISTING_POSTGRES=1 USE_EXISTING_REDIS=1`.

On a 24GB machine it is worse than redundant: starting Docker relaunches a roughly 1.7GB VM and re-conflicts the ports.

## What changed

`make` probes for a listening service, but **only for a variable the caller left unset**, decided with `$(origin ...)` rather than by inspecting the value. An explicit `1` skips the probe. An explicit `0` still forces the Docker path even when a native service is listening.

**The probe is skipped entirely under `CI`.** An earlier revision of this body claimed CI was safe because every workflow sets both variables explicitly. Review showed that was false: `ci.yml:592-593` is a *step-level* `env:` on the workflow-definition lifecycle step, not on the `make` steps at `:249` and `:580`, and that job publishes postgres on 5432 and redis on 6379, so `make cloud-client-generate` ran with both variables undefined and both ports listening. `cloud-tests.yml:16` and `cloud-live-webhook.yml:18` set only `USE_EXISTING_POSTGRES`, leaving the redis probe live. No consumer actually broke, but that was circumstance, which is precisely what the claim denied.

An `ifndef CI` guard now makes it true rather than lucky. A job's service topology is a property of its workflow file and must not depend on what happens to be listening. GitHub Actions always sets `CI`, and any job that wants the native services already says so explicitly.

Verified rather than asserted: under `CI=true`, `make -n server-db-ready` and `make -n server-redis-ready` are byte-identical to `main`, including the partial `ci.yml` shape where only `USE_EXISTING_POSTGRES` is set.

## Why a script, and why loopback only

`/dev/tcp` is a bash builtin with no connect timeout. Probing an unreachable remote host would hang the make parse with no way to interrupt it cleanly, so `scripts/probe-local-services.sh` refuses any host that is not `127.0.0.1`, `::1`, or `localhost`, where a connect is either accepted or refused immediately.

Anything else, a missing script, or a failing script all yield an empty word, and `$(or $(word 1,...),0)` falls back to `0`. Every degraded path lands on the Docker service, which is exactly the behaviour that existed before this change.

The Postgres half additionally requires `psql` on `PATH`. The existing-Postgres path does not merely open a socket: `scripts/dev.mjs` `ensureDatabase` shells out to `psql -h <host> -p <port>` when the flag is 1. The repo's own `server/docker-compose.yml` publishes its db on that same 5432 (`Makefile:307` defaults `PROLIFERATE_DB_HOST_PORT` to `LOCAL_PGPORT`), so a bare socket probe would flip a working Docker setup onto a psql path that is not installed, where `spawnSync` fails ENOENT and surfaces as a bare `psql ... failed` with no stdout or stderr to explain it. `probe_postgres` therefore requires psql first, mirroring the guard already at `Makefile:550`. Docker keeps talking to Docker.

Both services are probed in one bash invocation because this file is parsed once per recursive make. Two separate `python3` socket probes measured 70ms per parse against 10ms for one bash run.

## Cost

| variant | ms per make parse |
|---|---|
| `main` today | 17 |
| both flags set, probe skipped | 17 |
| both unset, probe runs | 30 |

(Independently re-measured during review. An earlier revision of this body said 18 / 20 / 31; the skipped case is in fact indistinguishable from `main`, so the old table overstated the cost.)

## Verification

An important trap first: **this shell already had `USE_EXISTING_POSTGRES=1` in its environment**, so my first run of the matrix reported `origin=environment` in every cell and never exercised the new code at all. Every result below was re-run under `env -u USE_EXISTING_POSTGRES -u USE_EXISTING_REDIS`.

Read through the real consumers, `make -n server-db-ready` and `make -n server-redis-ready`, rather than by printing the variable:

| case | selected | wanted |
|---|---|---|
| pg unset, Postgres listening | EXISTING | EXISTING |
| pg unset, port 5399 empty | DOCKER | DOCKER |
| pg explicit `0`, Postgres listening | DOCKER | DOCKER |
| pg explicit `1` | EXISTING | EXISTING |
| redis unset, Redis listening | EXISTING | EXISTING |
| redis unset, port 6399 empty | DOCKER | DOCKER |
| redis explicit `0` | DOCKER | DOCKER |

The probe script on its own:

```
loopback, both up            : 1 1
loopback, pg port empty      : 0 1
loopback, redis port empty   : 1 0
non-loopback host (no hang)  : 0 0
localhost spelling           : 1 1
```

### Negative control

Delete the probe script and re-resolve:

```
probe=[] pg=0 redis=0
server-db-ready -> DOCKER
```

That is `main`'s behaviour exactly, which is the point: the fallback is the old default, not a new failure mode. Forcing the script to print `0 0` gives the same, confirming the wiring reads the probe rather than ignoring it.

`shellcheck scripts/probe-local-services.sh` is clean.

## Relationship to #2134

Complementary, no overlap. #2134 makes `dev` recursive (`setup`, `dev-build`, `run`), adds `HEADLESS`, and stops the double cargo build. This removes the last reason the launch still needs an env prefix. Together, `make dev PROFILE=x` is the entire command.

Still not absorbed into `make` from a wrapper: nothing. `dev-init` is `setup`, and `setup` already runs the per-profile `ensure-db` at `Makefile:343` on this branch, `:318` on `main`, so with this change a wrapper has nothing left to contribute.

## Not for merge

Draft, and nothing here has been merged or self-approved.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,10 +10,35 @@ LOCAL_PGPORT ?= 5432
 LOCAL_PGUSER ?= proliferate
 LOCAL_PGPASSWORD ?= localdev
 LOCAL_PGDATABASE ?= proliferate
-USE_EXISTING_POSTGRES ?= 0
 LOCAL_REDIS_HOST ?= 127.0.0.1
 LOCAL_REDIS_PORT ?= 6379
-USE_EXISTING_REDIS ?= 0
+# A local service already bound to the port is the one to use. Shell wrappers
+# around `make` exist largely to prefix every invocation with
+# USE_EXISTING_POSTGRES=1 / USE_EXISTING_REDIS=1, so make does not bring up a
+# Docker Postgres or Redis on top of a native one. Probing here instead is what
+# lets `make dev PROFILE=x` be the whole launch, with nothing left for a wrapper
+# to add.
+#
+# The probe runs ONLY for a variable the caller left unset. An explicit 0 still
+# forces the Docker path and an explicit 1 still skips the probe, so every CI
+# workflow, all of which set both to "1", is unaffected by construction. A
+# missing or failing probe yields an empty word and falls back to 0, the
+# behaviour that existed before this block.
+#
+# Both services are probed in one invocation: this file is parsed once per
+# recursive make, and two separate python3 probes cost about 70ms each parse
+# against 10ms for one bash run.
+ifneq ($(filter undefined,$(origin USE_EXISTING_POSTGRES) $(origin USE_EXISTING_REDIS)),)
+LOCAL_SERVICE_PROBE := $(shell scripts/probe-local-services.sh '$(LOCAL_PGHOST)' '$(LOCAL_PGPORT)' '$(LOCAL_REDIS_HOST)' '$(LOCAL_REDIS_PORT)' 2>/dev/null)
+endif
+
+ifeq ($(origin USE_EXISTING_POSTGRES), undefined)
+USE_EXISTING_POSTGRES := $(or $(word 1,$(LOCAL_SERVICE_PROBE)),0)
+endif
+
+ifeq ($(origin USE_EXISTING_REDIS), undefined)
+USE_EXISTING_REDIS := $(or $(word 2,$(LOCAL_SERVICE_PROBE)),0)
+endif
 STRIPE_FORWARD_TO ?= http://127.0.0.1:8000/v1/billing/webhooks/stripe
 STRIPE_SNAPSHOT_EVENTS ?= checkout.session.completed,customer.subscription.created,customer.subscription.updated,customer.subscription.deleted,invoice.paid,invoice.payment_failed
 AGENT_GATEWAY ?= 0
@@ -1623,3 +1648,4 @@ catalog-update:
 	cd scripts/agent-catalog && node build-catalog.mjs --require-complete-probe \
 		&& cp catalog.draft.json ../../catalogs/agents/catalog.json
 	node scripts/validate-agent-catalog.mjs
+
--- a/Makefile
+++ b/Makefile
@@ -28,8 +28,16 @@ LOCAL_REDIS_PORT ?= 6379
 # Both services are probed in one invocation: this file is parsed once per
 # recursive make, and two separate python3 probes cost about 70ms each parse
 # against 10ms for one bash run.
+# Never probe under CI. A CI job's service topology is a property of its
+# workflow file, so it must not depend on what happens to be listening: several
+# jobs publish postgres on 5432 and redis on 6379 while running `make` steps
+# that leave both variables unset, and those steps must keep resolving the way
+# they did before this block existed. GitHub Actions always sets CI, and any
+# job that genuinely wants the native services already says so explicitly.
+ifndef CI
 ifneq ($(filter undefined,$(origin USE_EXISTING_POSTGRES) $(origin USE_EXISTING_REDIS)),)
 LOCAL_SERVICE_PROBE := $(shell scripts/probe-local-services.sh '$(LOCAL_PGHOST)' '$(LOCAL_PGPORT)' '$(LOCAL_REDIS_HOST)' '$(LOCAL_REDIS_PORT)' 2>/dev/null)
+endif
 endif
 
 ifeq ($(origin USE_EXISTING_POSTGRES), undefined)
@@ -1648,4 +1656,3 @@ catalog-update:
 	cd scripts/agent-catalog && node build-catalog.mjs --require-complete-probe \
 		&& cp catalog.draft.json ../../catalogs/agents/catalog.json
 	node scripts/validate-agent-catalog.mjs
-
--- a/guides/local/README.md
+++ b/guides/local/README.md
@@ -74,16 +74,30 @@ flows unless their focused command or flag is used.
 
 ## Existing Postgres And Redis
 
-By default, `setup` starts or reuses the repository's Docker Postgres service,
-and `run` starts or reuses the Docker Postgres and Redis services. To use
-services already running on the same host instead, select the existing-service
-paths on every command that touches them:
+A Postgres or Redis already listening on the configured loopback port is used
+as-is. `make` probes for one whenever `USE_EXISTING_POSTGRES` or
+`USE_EXISTING_REDIS` is unset, so a host running native services needs no flags
+at all:
+
+```bash
+make dev PROFILE=<name>
+```
+
+Both flags remain available and still win over the probe. Set either to `1` to
+skip probing, and to `0` to force the repository's Docker service even when a
+native one is listening:
 
 ```bash
 make setup PROFILE=<name> USE_EXISTING_POSTGRES=1
 make build # if this worktree needs artifacts
 make run PROFILE=<name> USE_EXISTING_POSTGRES=1 USE_EXISTING_REDIS=1
 ```
+
+The probe only looks at loopback hosts (`127.0.0.1`, `::1`, `localhost`), where
+a connection is either accepted or refused immediately. Point `LOCAL_PGHOST` or
+`LOCAL_REDIS_HOST` at anything else and the probe reports nothing, selecting the
+Docker service, so a slow or unreachable remote host can never stall a `make`
+parse.
 
 The existing-Postgres path requires `psql` on `PATH`. Its default login role is
 `proliferate`. Create that dedicated local role once from a PostgreSQL
--- a/scripts/probe-local-services.sh
+++ b/scripts/probe-local-services.sh
@@ -16,6 +16,19 @@
 
 set -u
 
+# The Postgres half additionally requires a client. `make`'s existing-Postgres
+# path does not just open a socket: scripts/dev.mjs `ensureDatabase` shells out
+# to `psql -h <host> -p <port>` when USE_EXISTING_POSTGRES is 1, and without
+# psql on PATH spawnSync fails ENOENT and surfaces as a bare
+# "psql ... failed" with no stdout or stderr to explain it. The repo's own
+# docker-compose publishes its db on the same 5432 by default, so a plain socket
+# probe would flip a working Docker setup onto a psql path that is not there.
+# Report 0 instead: Docker keeps talking to Docker.
+probe_postgres() {
+  command -v psql >/dev/null 2>&1 || { printf '0'; return; }
+  probe "$1" "$2"
+}
+
 probe() {
   local host="$1" port="$2"
   case "$host" in
@@ -32,4 +45,4 @@ probe() {
   fi
 }
 
-printf '%s %s\n' "$(probe "${1:-}" "${2:-}")" "$(probe "${3:-}" "${4:-}")"
+printf '%s %s\n' "$(probe_postgres "${1:-}" "${2:-}")" "$(probe "${3:-}" "${4:-}")"
--- a/scripts/probe-local-services.sh
+++ b/scripts/probe-local-services.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+# Report whether a local Postgres and Redis are already listening.
+#
+# Prints two words, "<postgres> <redis>", each 1 or 0, for the Makefile to read
+# when the caller has not set USE_EXISTING_POSTGRES / USE_EXISTING_REDIS. The
+# point is that `make dev PROFILE=x` should use a native service that is already
+# bound to the port rather than bringing up a Docker one on top of it, without
+# every invocation needing a shell wrapper to say so.
+#
+# /dev/tcp is a bash builtin and has no connect timeout, so probing is limited
+# to loopback hosts, where a connect either succeeds or is refused immediately
+# and can never hang a make parse. Any other host reports 0, which selects the
+# Docker path: the same answer make gave before this script existed.
+#
+# Usage: probe-local-services.sh <pg-host> <pg-port> <redis-host> <redis-port>
+
+set -u
+
+probe() {
+  local host="$1" port="$2"
+  case "$host" in
+    127.0.0.1 | ::1 | localhost) ;;
+    *)
+      printf '0'
+      return
+      ;;
+  esac
+  if (exec 3<>"/dev/tcp/$host/$port") 2>/dev/null; then
+    printf '1'
+  else
+    printf '0'
+  fi
+}
+
+printf '%s %s\n' "$(probe "${1:-}" "${2:-}")" "$(probe "${3:-}" "${4:-}")"


pablonyx (commented):
## Review: independent verification of #2161

Reviewer of record, did not write this. Everything below was re-run locally against the PR head and against `origin/main`, on GNU Make 3.81 (`/usr/bin/make`, the only `make` on this box). No cargo, no rustc, no Docker was started at any point; every make invocation was `-n`, and `docker info` was unreachable before and after.

**Environment leak: present, and I hit it.** This shell inherited both flags:

```
USE_EXISTING_POSTGRES=1
USE_EXISTING_REDIS=1
```

Unscrubbed, the new code never runs, exactly as the body warns:

```
PR: NOT scrubbed (leak present)
pg_origin=environment pg=[1] redis_origin=environment redis=[1] probe=[] ...
```

Every result below is under `env -u USE_EXISTING_POSTGRES -u USE_EXISTING_REDIS`.

---

## Blocking

### B1. The probe silently promotes `psql` from an opt-in prerequisite to a default one, and the failure mode is a bare `psql ... failed`

`server/docker-compose.yml` publishes the repository's own services on the same host ports the probe watches:

```yaml
  db:
    ports:
      - "${PROLIFERATE_DB_HOST_PORT:-5432}:5432"
  redis:
    ports:
      - "${PROLIFERATE_REDIS_HOST_PORT:-6379}:6379"
```

and `Makefile:307` defaults `PROLIFERATE_DB_HOST_PORT` to `$(LOCAL_PGPORT)`. So on the very common state of "I ran `make dev` yesterday and left the containers up", the probe now returns `1` for a container it did not detect as a container, and `ensure-db` switches branch:

`scripts/dev.mjs:729`
```js
const useExisting = process.env.USE_EXISTING_POSTGRES === "1";
```
`useExisting` true runs host `psql -h ::1 -p 5432`; false runs `docker compose exec -T db psql`. The Docker branch never needed a local `psql`. The existing branch does, which `guides/local/README.md` states as a prerequisite of a path the developer previously had to ask for by name.

When `psql` is absent, `run()` at `scripts/dev.mjs:159-171` gets `status: null` from `spawnSync` ENOENT with null stdout and stderr, so it throws the fallback string:

```js
  if (result.status !== 0) {
    const stderr = result.stderr?.trim();
    const stdout = result.stdout?.trim();
    throw new Error(stderr || stdout || `${command} ${args.join(" ")} failed`);
  }
```

A developer whose setup worked yesterday gets `psql -h ::1 -p 5432 ... failed` today, with nothing pointing at the cause or at `USE_EXISTING_POSTGRES=0`. That is a new default-path failure introduced by this change, not a pre-existing one.

Cheap fix that keeps everything else intact: make the Postgres half of the probe also require the tool that half of the path needs.

```bash
case "$host" in
  127.0.0.1 | ::1 | localhost) ;;
  *) printf '0'; return ;;
esac
```
plus, for the Postgres probe only, `command -v psql >/dev/null 2>&1 || { printf '0'; return; }`. The repository already guards this way at `Makefile:576` for `make db-local` ("psql is required for `make db-local`"). Downgrade this to non-blocking if the team's position is that `psql` is already a hard prerequisite for local dev, but then the guide should say so at the point where the probe can now route you there without asking.

### B2. "CI is unaffected by construction" is false. The probe does execute in CI, and in one job it flips both flags from `0` to `1`

The body says every CI workflow sets both flags explicitly and "the probe never even executes there". Neither half holds.

`ci.yml:592-593` is a **step-level** `env:` on the Playwright step. The two actual `make` invocations in `ci.yml` are not in its scope:

```
ci.yml:249   run: make qualification-candidate-handoff-smoke     (job candidate-build-handoff, no services, no flags)
ci.yml:580   run: make cloud-client-generate                     (job workflow-definition-lifecycle)
ci.yml:592-593   USE_EXISTING_POSTGRES/REDIS: "1"                (step "Run workflow-definition lifecycle spec", line 595)
```

The `workflow-definition-lifecycle` job publishes both services on the runner:

```yaml
      postgres:
        ports:
          - 5432:5432
      redis:
        ports:
          - 6379:6379
```

`LOCAL_PGHOST` is `127.0.0.1` on Linux. So the `make cloud-client-generate` at line 580 runs with both flags undefined and both ports listening: the probe returns `1 1` and both variables resolve to `1` inside that make, where `main` gives `0 0`.

`cloud-tests.yml:16` and `cloud-live-webhook.yml:18` set only `USE_EXISTING_POSTGRES`. `USE_EXISTING_REDIS` is undefined for `make cloud-runtime-build`, `make test-cloud-*` and `make test-cloud-webhooks`, so the probe runs there too.

I traced the consequences and found no CI breakage today: `cloud-client-generate` reaches `cloud-openapi` only, and the only targets that consume the flags are `server-db-ready` (`Makefile:755`), `server-redis-ready` (`Makefile:557`) and their dependants `server-migrate`, `db-migrate-up/down`, `db`, `dev-server`, `test-server`, `test-cloud-e2b`, `test-cloud-webhooks`, `test-cloud-all` — and in the two cloud workflows those all sit under an explicit `USE_EXISTING_POSTGRES: 1`.

So CI is fine, but it is fine **by circumstance**, which is the opposite of the argument the body makes. This is blocking not because the code is wrong but because the merge decision is being asked to rest on a safety property that does not exist. Either fix the body, or actually establish the property by hoisting the flags to job or workflow level in `ci.yml`, `cloud-tests.yml` and `cloud-live-webhook.yml`.

---

## Verified, no defect found

**The `$(origin ...)` logic is correct on the make that ships here.** `/usr/bin/make` is GNU Make 3.81 and there is no `gmake`. Every construct used is supported, including the `ifeq` form with a space after the comma:

```
make=3.81
R1=undefined-with-space-OK R2=undefined-nospace-OK
or/word: V1=a V2=0 V3=0
filter=[undefined]
```

`origin` discriminates correctly for all four callers:

```
-- env FOO=1 --        X=DEF-[environment]
-- cmdline FOO=1 --    X=DEF-[command line]
-- unset --            X=UNDEF
-- env FOO= (empty) -- X=DEF-[environment]
```

The last row is the `?=` trap and it is a non-issue: an empty-string env value keeps the old behaviour on both branches, since only a literal `1` selects the existing-service path.

**The ordering bug is genuinely fixed.** `LOCAL_PGHOST` (lines 4-8), `LOCAL_PGPORT` (9), `LOCAL_REDIS_HOST` (13), `LOCAL_REDIS_PORT` (14) all precede the `:=` probe at line 31-32, and nothing else the probe reads is defined later. Proven by value rather than by reading:

```
PR: both unset (scrubbed)
pg_origin=file pg=[1] redis_origin=file redis=[1] probe=[1 1] pghost=[::1] pgport=[5432] rhost=[127.0.0.1] rport=[6379]
```

All four are populated, which is the exact symptom that would be absent if the ordering had regressed.

**The verification table in the body reproduces exactly**, read through the real consumers:

```
PR  db-ready   both unset                            -> EXISTING
PR  redis-ready both unset                           -> EXISTING
PR  db-ready   pgport=5399                           -> DOCKER
PR  redis-ready rport=6399                           -> DOCKER
PR  db-ready   explicit 0                            -> DOCKER
PR  redis-ready explicit 0                           -> DOCKER
PR  db-ready   explicit 1                            -> EXISTING
PR  db-ready   remote pghost                         -> DOCKER
MAIN db-ready   both unset                           -> DOCKER
MAIN redis-ready both unset                          -> DOCKER
```

**The value does reach `dev.mjs`.** This was my main suspicion, since `USE_EXISTING_POSTGRES` is now a `:=` make variable and make does not export those. `PROFILE_DB_ENSURE_COMMAND` at `Makefile:118` passes it positionally, so it survives. With the flag scrubbed from the environment, `make -n setup PROFILE=main` prints:

```
		make server-db-ready; \
		LOCAL_PGHOST="::1" LOCAL_PGPORT="5432" LOCAL_PGUSER="proliferate" LOCAL_PGPASSWORD="localdev" USE_EXISTING_POSTGRES="1" node scripts/dev.mjs ensure-db --db-name "$PROLIFERATE_DEV_DB_NAME"; \
```

`server-db-ready` and `ensure-db` agree. Not a defect.

**The loopback allowlist holds. I could not bypass it.** Thirteen vectors, including every one asked for:

```
::1 / 127.0.0.1 both up            : 1 1
127.0.0.1 both up                  : 1 1
localhost spelling                 : 1 1
pg port empty (5399)               : 0 1
redis port empty (6399)            : 1 0
[::1] bracketed                    : 0 0
trailing dot localhost.            : 0 0
0.0.0.0                            : 0 0
::ffff:127.0.0.1                   : 0 0
127.0.0.2                          : 0 0
0177.0.0.1 (octal)                 : 0 0
2130706433 (int form)              : 0 0
empty host                         : 0 0
no args at all                     : 0 0
host=*                             : 0 0
host=127.0.0.1 port=$(id)          : 0 1
host with space                    : 0 0
```

Every rejected form yields `0`, which is the Docker path, so no bypass is even a fail-open. The `$(id)` row confirms the redirection target is not re-expanded. The only theoretical hole is `localhost` resolving somewhere non-loopback via a doctored `/etc/hosts` or a hostname-less resolver, which requires an attacker who already owns the machine. Not reachable in practice, not worth blocking.

**Fallback really is `main`, tested rather than argued.** With the script moved aside, `make -n` output is byte-identical to `main` for both consumers:

```
--- diff PR(no script) vs MAIN, server-db-ready ---
IDENTICAL
--- diff PR(no script) vs MAIN, server-redis-ready ---
IDENTICAL
```

Every degraded variant lands on `0`:

```
script missing            : probe=[]      pg=[0] redis=[0]
script prints "0 0"       : probe=[0 0]   pg=[0] redis=[0]
script exits 3 with stderr: probe=[]      pg=[0] redis=[0]
script not executable     : probe=[]      pg=[0] redis=[0]
make -f ../Makefile from a subdir (wrong cwd for the relative path)
                          : probe=[]      pg=[0] redis=[0]
```

Junk output degrades safely too: a script printing `yes no` yields `pg=[yes]`, which no `ifeq ...,1` matches, so `server-db-ready` still selects Docker. Verified.

**`shellcheck` is clean** at default severity:

```
shellcheck exit=0
```

Under `-o all` it reports only SC2250 (brace style) and SC2312 (info, masked return in command substitution). Neither is worth acting on. `set -u` is handled correctly by `"${1:-}"` through `"${4:-}"`, and `/dev/tcp` works on this machine's `/bin/bash` 3.2.57, which is the worst case a `#!/bin/bash` shebang will meet on a supported host. Note there is no shellcheck gate in CI, so this stays a manual property.

**The cost table is honest, and slightly pessimistic in the PR's own disfavour.** 25 parses each:

```
main, both unset            : 17 ms/parse
PR,   both set to 1 (skip)  : 17 ms/parse
PR,   both unset (probe)    : 30 ms/parse
```

The body's "20" for the skipped case is really "identical to main". The +13ms lands on roughly four parses per `make dev` (`dev`, plus the recursive `make server-db-ready` from `setup`, and `make server-redis-ready` and `make server-db-ready` from `run`), so about 50ms on a launch. Fine.

**The wrapper claim mostly holds.** `setup` does run the per-profile `ensure-db`: `setup:` is at `Makefile:343` in this branch (the body says 342, off by one; it is 318 on `main`), and lines 366-367 run `$(PROFILE_DB_READY_COMMAND)` then `$(PROFILE_DB_ENSURE_COMMAND)`. I also checked the other knobs a wrapper typically carries: make 3.81 does place command-line variable definitions into the recipe environment, so `make dev PROFILE=x SKIP_RUST=1 ANYHARNESS_DEV_RUNTIME_BIN=...` works without a prefix even though `ANYHARNESS_DEV_RUNTIME_BIN` is read from `$${...}` inside the recipe at `Makefile:157` and `:334`:

```
-- cmdline --  recipe env FOO=[bar]
-- env --      recipe env FOO=[bar]
-- neither --  recipe env FOO=[unset]
```

Two caveats on "nothing left to contribute": it presumes #2134 lands, and the `setup` ensure-db call is inside `if [ "$$database_url_override_set" = "x" ]` so it is skipped whenever `DATABASE_URL` is set. Worth softening the wording rather than fixing anything.

**`LOCAL_PGHOST` / `LOCAL_PGPORT` defaults match the native service.** On this Darwin box the defaults are `::1:5432`, and brew Postgres binds both families:

```
postgres 2248 ... IPv6 ... TCP [::1]:5432 (LISTEN)
postgres 2248 ... IPv4 ... TCP 127.0.0.1:5432 (LISTEN)
redis-ser 2235 ... IPv4 ... TCP 127.0.0.1:6379 (LISTEN)
redis-ser 2235 ... IPv6 ... TCP [::1]:6379 (LISTEN)
```

Credentials line up too: compose sets `POSTGRES_USER: proliferate` / `POSTGRES_PASSWORD: localdev`, matching `LOCAL_PGUSER` / `LOCAL_PGPASSWORD`. And `DATABASE_URL` is built from `LOCAL_PGHOST`/`LOCAL_PGPORT` on both branches (`Makefile:226-232`), so the server never disagrees with whichever path `server-db-ready` picked.

---

## Non-blocking

1. **Stray trailing newline at the end of `Makefile`.** The diff adds a blank line after `catalog-update`. Confirmed at the byte level: PR ends `...mjs\n\n`, main ends `...mjs\n`. Unrelated noise, drop it.
2. **A second worktree's Docker Postgres is now adopted silently.** `main` would have failed loudly with "port is already allocated" when `docker compose up -d db` collided. The intended isolation knob is `LOCAL_PGPORT` (which `PROLIFERATE_DB_HOST_PORT` defaults to), and the probe does read `LOCAL_PGPORT`, so a correctly isolated worktree is fine. A half-isolated one that only set `PROLIFERATE_DB_HOST_PORT` was already broken on `main`. Mentioning for the record, not asking for a change.
3. **Narrow race between the parent parse and the recursive parse.** `PROFILE_DB_READY_COMMAND` is a plain `make server-db-ready`, which re-parses and re-probes, while `PROFILE_DB_ENSURE_COMMAND` carries the parent's value. If a listener appears between the two parses, the child skips `server-db-up` while `ensure-db` still tries `docker compose exec`. Requires a service to start inside a few hundred milliseconds. Not worth engineering against, but it is a new class of inconsistency that a constant `0` did not have.
4. **The guide's second code block is now slightly orphaned.** After "Both flags remain available and still win over the probe", the example shown is the old three-command `USE_EXISTING_POSTGRES=1` sequence, which illustrates the skip case rather than the `0` case the sentence just introduced. And the `psql` prerequisite sentence that follows should say that the probe can now put you on that path without you asking (see B1).

---

## Verdict

**Not safe to merge as it stands, but the distance to safe is short.**

The mechanism is sound. The `$(origin ...)` gating is correct on GNU Make 3.81, the ordering bug you hit during development is genuinely fixed, the allowlist survived every bypass I could construct, the resolved value does reach `dev.mjs`, and the fallback is byte-identical to `main` under five different degraded conditions. That is a well-built change and the verification table in the body reproduced without a single discrepancy.

Two things need to change first. B1 is a real new default-path failure for developers without `psql` who leave the repository's own containers running, and it costs two lines in the probe script to close. B2 is not a code defect but the body's central safety argument is factually wrong: the probe does run in CI, and in `ci.yml`'s `workflow-definition-lifecycle` job it flips both flags to `1` where `main` gives `0`. It happens to be harmless today, which is luck, and the body specifically claims it is not luck. Correct the claim or make it true.

pablonyx:
Both blocking findings fixed in `090408e799`. Both were real and both were mine.

**B1** is the one I'm glad you caught. I had reasoned about "is something listening" when the actual question is "is there a usable Postgres", and those differ precisely in the case you found: the repo's own compose publishes on the same 5432, so the probe would have flipped a working Docker setup onto a psql path that isn't installed, failing with a message that names neither psql nor the probe. `probe_postgres` now requires `psql` first, mirroring the guard at `Makefile:550`.

**B2** I fixed rather than reworded. You were right that the safety argument was circumstance dressed as construction, and correcting the sentence would have left the same fragility one workflow edit away. There is now an `ifndef CI` guard, so the probe never runs in CI at all. Verified the way the original claim should have been: under `CI=true`, `make -n server-db-ready` and `make -n server-redis-ready` are byte-identical to `main`, including the partial `ci.yml` shape where only `USE_EXISTING_POSTGRES` is set.

Also applied: the cost table now carries your measured 17/17/30 (mine overstated it, and the skipped case is indistinguishable from `main`), `Makefile:342` corrected to `:343` on this branch and `:318` on `main`, and the stray trailing newline is gone.

Full matrix re-run under `env -u USE_EXISTING_POSTGRES -u USE_EXISTING_REDIS -u CI`:

```
pg unset, listening, psql present  -> EXISTING
pg unset, listening, no psql       -> DOCKER
pg unset, port empty               -> DOCKER
pg explicit 0, listening           -> DOCKER
pg explicit 1                      -> EXISTING
pg CI=true, listening              -> DOCKER
redis unset, listening             -> EXISTING
redis unset, port empty            -> DOCKER
redis explicit 0                   -> DOCKER
redis CI=true, listening           -> DOCKER
```

Negative control, script removed: `probe=[] pg=0 redis=0`, `server-db-ready` selects DOCKER, matching `main`.

Two things I did not change. The 17-vector allowlist result and the GNU Make 3.81 compatibility check both stand as you found them, and "a wrapper has nothing left to contribute" still presumes #2134 lands, which the body already says.

PR #2165, opened 2026-08-21
test(anyharness): drive a real ACP initialize over the windows launcher

pablonyx:
## What is now proven that was not before

A managed `claude` agent, installed by the real pinned installer and launched through the emitted Windows `claude-launcher.cmd`, **completes a real ACP `initialize` handshake over stdio on a real `windows-latest` runner**, twice on the same pipe, and then exits cleanly. No credential of any kind is involved.

Before this, the furthest anything had gone on Windows was spawn plus clean exit.

## Why that was not already proven

Run [32450488248](https://github.com/proliferate-ai/proliferate/actions/runs/32450488248) proved the layer underneath: the installer downloads and sha256-verifies a 324 MB `claude.exe`, runs a real `npm install` of the git-pinned adapter, emits `claude-launcher.cmd`, and `CreateProcess` spawns it. Its literal verdict for claude was:

```
--- launcher stdout

--- launcher stderr

```

Nothing was ever written to that launcher's stdin and nothing was ever read back. Exit 0 with empty stdout and empty stderr establishes only "the process starts and exits cleanly on EOF"; a `.cmd` containing nothing but `@echo off` produces exactly that. Whether the batch wrapper forwards its stdio handles to the grandchild node process, whether cmd.exe injects anything into the stream, and whether newline-delimited JSON survives a Windows pipe in both directions were all untested.

## The evidence, literal

Passing run: [32454309124](https://github.com/proliferate-ai/proliferate/actions/runs/32454309124), both legs green, `acp initialize (ubuntu-latest)=success, acp initialize (windows-latest)=success`.

From `acp initialize (windows-latest)`, the launcher the real installer emitted:

```
installed role=NativeCli source=pinned_binary version=Some("2.1.234") path=C:\Users\RUNNER~1\AppData\Local\Temp\anyharness-acp-handshake-handshake-4320-1787293798899735000\agents\claude\native\claude.exe
installed role=AgentProcess source=pinned_git version=Some("65f6205329daea4a521ee0f0c042722b0feb306a") path=C:\Users\RUNNER~1\AppData\Local\Temp\anyharness-acp-handshake-handshake-4320-1787293798899735000\agents\claude\agent_process\claude-launcher.cmd
@echo off
set "PATH=C:\Users\RUNNER~1\AppData\Local\Temp\anyharness-acp-handshake-handshake-4320-1787293798899735000\agents\claude\native;%PATH%"
set "DISABLE_AUTOUPDATER=1"
"C:\Users\RUNNER~1\AppData\Local\Temp\anyharness-acp-handshake-handshake-4320-1787293798899735000\agents\claude\agent_process\node_modules/.bin/claude-agent-acp.cmd" %*
exit /b %ERRORLEVEL%
```

The request that went into that launcher's stdin:

```
--> {"id":4242,"jsonrpc":"2.0","method":"initialize","params":{"clientCapabilities":{"_meta":{"claude":{"mcpElicitation":true}},"auth":{"terminal":false},"fs":{"readTextFile":false,"writeTextFile":false},"terminal":false},"clientInfo":{"name":"anyharness","version":"0.1.0"},"protocolVersion":1}}
```

And what came back out of it:

```
<-- {"jsonrpc":"2.0","id":4242,"result":{"protocolVersion":1,"agentCapabilities":{"_meta":{"claudeCode":{"promptQueueing":true}},"promptCapabilities":{"image":true,"embeddedContext":true},"mcpCapabilities":{"http":true,"sse":true},"auth":{"logout":{}},"providers":{},"loadSession":true,"sessionCapabilities":{"additionalDirectories":{},"close":{},"delete":{},"fork":{},"list":{},"resume":{}}},"agentInfo":{"name":"@proliferate/claude-agent-acp","title":"Claude Agent","version":"0.66.0-proliferate.2"},"authMethods":[],"_meta":{"steering":{"supported":true},"goal":{"version":1,"controlMethod":"_session/goal","actions":["set","clear"]},"anyharness":{"fork":{"version":1,"anchor":"upToMessageId"},"rewindFiles":{"version":1,"controlMethod":"_anyharness/rewindFiles","scope":"writeEditNotebookEdit","requiresCheckpointingOptIn":true},"activity":{"version":1,"listMethod":"_anyharness/activity/list"}}}}}
response to `initialize` (id 4242) arrived after 443.5497ms
handshake OK: agent "@proliferate/claude-agent-acp" version "0.66.0-proliferate.2", 0 auth method(s) advertised
```

Second request on the same pipe, which is what rules out a one-shot flush and shows ndjson framing holds across messages through the `.cmd` wrapper:

```
response to `initialize` (id 4243) arrived after 1.1066ms
handshake OK: agent "@proliferate/claude-agent-acp" version "0.66.0-proliferate.2", 0 auth method(s) advertised
--- exit status after stdin close: Some(ExitStatus(ExitStatus(0)))
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 333.53s
```

Note `"authMethods":[]`. The handshake completed with the agent advertising no authentication method, which is the direct observation that `initialize` needs no credential.

The `acp initialize (ubuntu-latest)` control leg produced the same handshake against the same code path and reported `test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 86.57s`, so a green Windows leg cannot be blamed on the test being unix-specific, and a red one could not have been blamed on the test.

## Negative control, both directions

**In-suite, on every run, on both operating systems.** The same reader is driven against a process that never answers and a process that exits without answering, and both must fail. From the passing Windows leg:

```
discriminator produced: ACP-HANDSHAKE-EOF: stdout closed before any JSON-RPC response to `initialize` arrived. The process exited or dropped the stream instead of answering. An empty response is a failure, not a pass.
discriminator produced: ACP-HANDSHAKE-TIMEOUT: no JSON-RPC response to `initialize` arrived on stdout within 15.0116037s. The process was spawned and accepted the request but never answered it. This is a failure, not a skip.
```

The EOF case is not hypothetical. It is precisely the behaviour run 32450488248 observed and counted as success.

**The real path, deliberately broken.** Run [32454334149](https://github.com/proliferate-ai/proliferate/actions/runs/32454334149) is the identical test with `ACP_HANDSHAKE_NEGATIVE_CONTROL=1`, which replaces the npm bin shim the emitted launcher invokes with a program that starts, holds stdin and never answers. The launcher, the install, the spawn and the assertions are untouched. Both legs went red for exactly the intended reason.

`acp initialize (windows-latest)`:

```
!!! NEGATIVE CONTROL ENGAGED (ACP_HANDSHAKE_NEGATIVE_CONTROL=1): the npm bin shim the emitted launcher invokes will be replaced with a process that starts, holds stdin and never answers. This run MUST fail with ACP-HANDSHAKE-TIMEOUT.
negative control: replaced C:\Users\RUNNER~1\AppData\Local\Temp\anyharness-acp-handshake-handshake-6508-1787293847094123400\agents\claude\agent_process\node_modules\.bin\claude-agent-acp.cmd with a silent stub
thread 'the_managed_claude_launcher_answers_a_real_acp_initialize_handshake' (7276) panicked at anyharness\crates\anyharness-lib\tests\windows_acp_handshake.rs:234:13:
ACP-HANDSHAKE-TIMEOUT: no JSON-RPC response to `initialize` arrived on stdout within 180.0010135s. The process was spawned and accepted the request but never answered it. This is a failure, not a skip.
test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 533.62s
```

`acp initialize (ubuntu-latest)`:

```
negative control: replaced /tmp/anyharness-acp-handshake-handshake-6369-1787293691053863990/agents/claude/agent_process/node_modules/.bin/claude-agent-acp with a silent stub
thread 'the_managed_claude_launcher_answers_a_real_acp_initialize_handshake' (6392) panicked at anyharness/crates/anyharness-lib/tests/windows_acp_handshake.rs:234:13:
ACP-HANDSHAKE-TIMEOUT: no JSON-RPC response to `initialize` arrived on stdout within 180.000100931s. The process was spawned and accepted the request but never answered it. This is a failure, not a skip.
test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 257.79s
```

Note that the two in-suite discriminators still passed in the sabotaged run (`2 passed; 1 failed`), so the failure is attributable to the sabotage and not to a broken harness.

That control lives on [#2166](https://github.com/proliferate-ai/proliferate/pull/2166), a throwaway branch that exists only to produce the failing half of this proof and is not for merge.

## What the test does

```mermaid
flowchart TD
  A[bundled catalog and registry] --> B[install_agent_with_pins]
  B --> C[claude.exe, sha256 verified]
  B --> D[npm install of the git-pinned adapter]
  D --> E[emitted claude-launcher.cmd]
  E --> F[spawn with all three handles piped]
  F --> G[write initialize as ndjson to stdin]
  G --> H[read and validate the JSON-RPC response]
  H --> I[write a second initialize on the same pipe]
  I --> J[read and validate again]
  J --> K[close stdin, require a clean exit]
```

Validation of each response is done twice over: on the raw JSON (`jsonrpc`, matching `id`, no `error` member, non-empty `result`, `protocolVersion == 1`, a non-blank `result.agentInfo.name`, an object `agentCapabilities`, an array `authMethods`) and then again by deserializing into the runtime's own `acp::schema::InitializeResponse`. The raw pass comes first on purpose: that type deserializes `agentInfo` and `authMethods` with `DefaultOnError`, so a typed round trip alone would silently accept a garbage `agentInfo`.

## Which agent, and why

`claude`. It is the default agent, it is the one that has to work for a Windows beta, and it is the only agent that can answer this question today.

- `cursor` and `opencode` pin their agent process as a platform archive whose `targets` map contains only `macos_arm64`, `macos_x64` and `linux_x64`. There is no `windows_*` target, so `installer/pinned.rs` raises `NoPinForPlatform` before anything is installed.
- `codex` does have `windows_x64` and `windows_arm64` native pins, but they point at `codex-x86_64-pc-windows-msvc.exe.tar.gz`, a tarball extraction branch nothing has executed on Windows. A red result there would be ambiguous between the extraction branch and the handshake.
- `grok` installs and spawns on Windows (proved in run 32450488248) but its ACP surface is a different adapter and it is not the default.

## Credentials: verified, not assumed

No API key is used, needed, added as a secret, or written to any file, PR or log. That was checked on both sides of the wire before the test was written.

Client side, `live/sessions/driver/session_lifecycle.rs::initialize_connection` sends `initialize` first and only then inspects the response's `auth_methods` to decide whether to call `authenticate`. Authentication is strictly downstream of the handshake.

Agent side, the pinned adapter was read at the exact sha the catalog installs, `proliferate-ai/claude-agent-acp@65f6205329daea4a521ee0f0c042722b0feb306a`. `ClaudeAcpAgent.initialize` in `src/acp-agent.ts` reads `request.clientCapabilities` and a handful of `process.env` remote hints, then returns a literal response object; it opens no credential file and makes no network call. `src/index.ts`'s startup path is likewise credential-free.

The client capabilities the test sends mirror what the runtime sends for claude, which advertise no terminal-auth capability, so the adapter has no auth method to offer back. The test therefore asserts `authMethods` is an array and deliberately does not assert it is non-empty. The observed `"authMethods":[]` above is the confirmation.

## Where the test lives, and hangs and skips

`anyharness/crates/anyharness-lib/tests/windows_acp_handshake.rs`, an integration test target, with its stdio peer in `tests/acp_stdio/mod.rs` (a subdirectory, so cargo does not pick it up as a second test target). The prior lane recorded that the crate's `--lib` test target does not build on Windows because of ten pre-existing test-only POSIX assumptions; I did not re-verify that count. What I did verify is that the integration-test shape compiles and runs on `windows-latest`, in run 32450488248 and again here.

Every read is bounded by a deadline, and the job carries `timeout-minutes: 45` as a second bound. There is no skip path in the file. A missing catalog pin, a missing registry descriptor, a failed install, a missing launcher, a failed spawn and a missing node each panic with their own message. The five hollow-result modes each have a distinct marker: `ACP-HANDSHAKE-TIMEOUT`, `ACP-HANDSHAKE-EOF`, `ACP-HANDSHAKE-WRITE-FAILED`, `ACP-HANDSHAKE-ERROR-RESPONSE` (whose message says plainly that if the error is an auth error then the credential-free premise is wrong and must be reported as such), and `ACP-HANDSHAKE-EMPTY-RESULT`.

## How to read the checks

`Windows compile check` sets `continue-on-error: true` on both of its jobs, so its green is unconditional and means nothing. `ci-ok`'s `needs:` list covers only `ci.yml` jobs, so a Windows workflow sits outside that rollup gate. Read `Windows managed-agent ACP handshake` directly.

## One inherited red, not mine

`Repo shape checks` and therefore `ci-ok` are red on this PR. They are red for `anyharness/crates/anyharness-lib/src/domains/agents/installer/npm.rs: PROD-SIZE-1 ... found: 628 lines (max 600)`, a file this PR does not touch. The same two checks are already red on the base branch #2153 for the same reason ([run 32450488265](https://github.com/proliferate-ai/proliferate/actions/runs/32450488265)). This PR's diff is three added files and nothing else, and both of its Rust files pass `scripts/check_max_lines.py`.

## Base branch, and why this is not mergeable as-is

This targets `ci/windows-managed-agent-e2e` ([#2153](https://github.com/proliferate-ai/proliferate/pull/2153)), the combination branch that composes [#2149](https://github.com/proliferate-ai/proliferate/pull/2149) and [#2152](https://github.com/proliferate-ai/proliferate/pull/2152) and is itself explicitly marked not for merge. Those two carry the production fixes this test depends on: without #2149 there are no Windows pins for claude, without #2152 there is no `.cmd` launcher to speak to. This PR is a proof, not a merge candidate; the test and the workflow should ride along with whichever branch eventually lands those fixes.

## Still unproven

- **A full prompt round trip.** `session/new` and `session/prompt` need a real Anthropic credential and a model call. Nothing here says a Windows user can get a completion.
- **`authenticate` on Windows.** Not exercised. The capabilities sent advertise no auth method, so the adapter offered none.
- **Every other agent on Windows.** `cursor` and `opencode` have no `windows_*` target in their agent-process pins at all. `codex` has native Windows pins but its `.exe.tar.gz` extraction path has never run on Windows. `grok` installs and spawns but its ACP handshake is untested.
- **Windows ARM64.** Only `x86_64-pc-windows-msvc` is exercised, even though `windows_arm64` pins exist for both claude and codex.
- **Everything above the runtime.** The desktop shell, the updater, process-tree kill under real load, and long-lived session behaviour are all out of scope here.

### DIFF
--- a/.github/workflows/windows-acp-handshake.yml
+++ b/.github/workflows/windows-acp-handshake.yml
@@ -1,0 +1,92 @@
+name: Windows managed-agent ACP handshake
+
+# The question `windows-managed-agent-install.yml` could not answer.
+#
+# That job proved the managed claude install produces a `.cmd` launcher that
+# `CreateProcess` can spawn and that exits 0 on stdin EOF. It never wrote a
+# byte to the launcher's stdin and never read a byte back, so "exit 0, empty
+# stdout, empty stderr" was the whole finding. A `.cmd` containing only
+# `@echo off` would have produced exactly that.
+#
+# This job writes a real ACP `initialize` request into the launcher and
+# requires a well-formed JSON-RPC response back, twice, before a clean
+# shutdown. No credential is involved: `initialize` is the handshake that
+# precedes authentication on both sides (see the test's module docs for the
+# source reading that establishes it).
+#
+# Ubuntu runs the identical test as the discriminating control, so a green
+# Windows leg cannot be explained by the test being vacuous.
+#
+# Deliberately NOT `continue-on-error`: the point is a verdict. Note that
+# `windows-compile-check.yml` IS `continue-on-error` on both of its jobs and
+# `ci-ok`'s `needs:` list covers only `ci.yml` jobs, so neither of those tells
+# you anything about this one. Read this workflow's own result.
+
+on:
+  workflow_dispatch:
+    inputs:
+      negative_control:
+        description: >-
+          Sabotage the handshake on purpose. Replaces the npm bin shim the
+          emitted launcher invokes with a process that starts, holds stdin and
+          never answers, leaving the launcher, the spawn and the assertions
+          untouched. The run MUST go red with ACP-HANDSHAKE-TIMEOUT. Used to
+          prove the test can fail.
+        type: boolean
+        default: false
+  pull_request:
+    paths:
+      - "anyharness/crates/anyharness-lib/tests/windows_acp_handshake.rs"
+      - "anyharness/crates/anyharness-lib/src/domains/agents/installer/**"
+      - "anyharness/crates/anyharness-lib/src/integrations/agent_cli/**"
+      - "catalogs/agents/**"
+      - ".github/workflows/windows-acp-handshake.yml"
+
+permissions:
+  contents: read
+
+jobs:
+  acp-handshake:
+    name: acp initialize (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    # A handshake test that hangs is worse than no test: it would burn the
+    # runner's default 6h and report nothing. Every read inside the test is
+    # already deadline-bounded; this is the belt to that pair of braces.
+    timeout-minutes: 45
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # The control: this path is exercised on unix every day.
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+          # The unknown.
+          - os: windows-latest
+            target: x86_64-pc-windows-msvc
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: ${{ matrix.target }}
+      - uses: Swatinem/rust-cache@v2
+      - name: Node / npm as the runner ships them
+        # The install shells out to npm, and the in-suite discriminators spawn
+        # `node -e`. Record what is on PATH so a spawn failure can be read
+        # against the real layout rather than an assumption about it.
+        shell: bash
+        run: |
+          node --version || true
+          npm --version || true
+          echo "which node: $(which node || true)"
+          echo "PATHEXT: ${PATHEXT:-<unset>}"
+      - name: cargo test
+        # Serial and uncaptured: the test runs a real `npm install`, spawns a
+        # real agent, and prints the actual JSON-RPC lines in both directions.
+        # That transcript IS the finding.
+        env:
+          RUST_BACKTRACE: "1"
+          ACP_HANDSHAKE_NEGATIVE_CONTROL: ${{ inputs.negative_control && '1' || '0' }}
+        run: >
+          cargo test --target ${{ matrix.target }}
+          -p anyharness-lib --test windows_acp_handshake
+          -- --nocapture --test-threads=1
--- a/anyharness/crates/anyharness-lib/tests/acp_stdio/mod.rs
+++ b/anyharness/crates/anyharness-lib/tests/acp_stdio/mod.rs
@@ -1,0 +1,300 @@
+//! A deadline-bounded newline-delimited-JSON peer for driving a spawned ACP
+//! agent over stdio, shared by the tests in `windows_acp_handshake.rs`.
+//!
+//! The framing is hand-rolled rather than driven through
+//! `acp::ClientSideConnection` on purpose: that client owns its own async
+//! runtime and its own unbounded read loop, and a hang inside it would give no
+//! diagnosis. The BYTES are identical (the adapter reads the stream with
+//! `ndJsonStream(...)` over `process.stdin`), and the request and response
+//! payloads still go through the real `acp::schema` types at the call sites, so
+//! the only thing reimplemented here is "one JSON value per line".
+//!
+//! Every read is bounded. There is no code path here that waits forever and no
+//! code path that turns silence into success: silence is
+//! `ACP-HANDSHAKE-TIMEOUT`, a closed stream is `ACP-HANDSHAKE-EOF`, and an
+//! unusable stdin handle is `ACP-HANDSHAKE-WRITE-FAILED`.
+//!
+//! This lives in a subdirectory, not at `tests/acp_stdio.rs`, so cargo does not
+//! pick it up as a second integration-test target.
+
+// Not every test in the suite touches every helper; the module is shared
+// support code, not a public API.
+#![allow(dead_code)]
+
+use std::io::{BufRead, BufReader, Write};
+use std::process::{Child, Command, ExitStatus, Stdio};
+use std::sync::mpsc::{self, Receiver, RecvTimeoutError};
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
+
+#[derive(Debug)]
+pub enum HandshakeFailure {
+    /// Nothing came back in time. The loud one.
+    Timeout {
+        method: String,
+        waited: Duration,
+        stderr: String,
+        transcript: Vec<String>,
+    },
+    /// stdout closed (or errored) without ever carrying our response.
+    Eof {
+        method: String,
+        stderr: String,
+        transcript: Vec<String>,
+    },
+    /// The request could not even be handed to the child.
+    WriteFailed { method: String, error: String },
+}
+
+impl std::fmt::Display for HandshakeFailure {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Timeout {
+                method,
+                waited,
+                stderr,
+                transcript,
+            } => write!(
+                f,
+                "ACP-HANDSHAKE-TIMEOUT: no JSON-RPC response to `{method}` arrived on stdout \
+                 within {waited:?}. The process was spawned and accepted the request but never \
+                 answered it. This is a failure, not a skip.\n\
+                 lines received on stdout ({} total): {transcript:#?}\n\
+                 stderr so far:\n{stderr}",
+                transcript.len()
+            ),
+            Self::Eof {
+                method,
+                stderr,
+                transcript,
+            } => write!(
+                f,
+                "ACP-HANDSHAKE-EOF: stdout closed before any JSON-RPC response to `{method}` \
+                 arrived. The process exited or dropped the stream instead of answering. An \
+                 empty response is a failure, not a pass.\n\
+                 lines received on stdout ({} total): {transcript:#?}\n\
+                 stderr so far:\n{stderr}",
+                transcript.len()
+            ),
+            Self::WriteFailed { method, error } => write!(
+                f,
+                "ACP-HANDSHAKE-WRITE-FAILED: could not write the `{method}` request to the \
+                 child's stdin: {error}. The stdio handle the launcher handed us is not usable."
+            ),
+        }
+    }
+}
+
+/// Owns a spawned process and reads newline-delimited JSON off its stdout with
+/// a deadline on every read.
+///
+/// The framing is hand-rolled rather than driven through
+/// `acp::ClientSideConnection` on purpose: that client owns its own async
+/// runtime and its own (unbounded) read loop, and a hang inside it would give
+/// no diagnosis. The BYTES are identical — the adapter reads this stream with
+/// `ndJsonStream(...)` over `process.stdin` — and the request and response
+/// payloads below still go through the real `acp::schema` types, so the only
+/// thing reimplemented here is "one JSON value per line".
+pub struct AcpPeer {
+    child: Child,
+    stdin: Option<std::process::ChildStdin>,
+    lines: Receiver<String>,
+    stderr: Arc<Mutex<String>>,
+    transcript: Vec<String>,
+}
+
+impl AcpPeer {
+    pub fn attach(mut child: Child) -> Self {
+        let stdin = child.stdin.take().expect("piped stdin");
+        let stdout = child.stdout.take().expect("piped stdout");
+        let mut stderr = child.stderr.take().expect("piped stderr");
+
+        let (tx, lines) = mpsc::channel();
+        std::thread::spawn(move || {
+            let reader = BufReader::new(stdout);
+            for line in reader.lines() {
+                match line {
+                    Ok(line) => {
+                        if tx.send(line).is_err() {
+                            return;
+                        }
+                    }
+                    // A read error is EOF for our purposes; dropping `tx` on
+                    // the way out is what signals it.
+                    Err(_) => return,
+                }
+            }
+        });
+
+        let stderr_buffer = Arc::new(Mutex::new(String::new()));
+        let stderr_sink = Arc::clone(&stderr_buffer);
+        std::thread::spawn(move || {
+            // Drained continuously so the child can never wedge on a full
+            // stderr pipe while we are waiting on stdout.
+            let mut chunk = [0u8; 4096];
+            loop {
+                match std::io::Read::read(&mut stderr, &mut chunk) {
+                    Ok(0) | Err(_) => return,
+                    Ok(n) => {
+                        let text = String::from_utf8_lossy(&chunk[..n]).into_owned();
+                        if let Ok(mut sink) = stderr_sink.lock() {
+                            sink.push_str(&text);
+                        }
+                    }
+                }
+            }
+        });
+
+        Self {
+            child,
+            stdin: Some(stdin),
+            lines,
+            stderr: stderr_buffer,
+            transcript: Vec::new(),
+        }
+    }
+
+    pub fn stderr_snapshot(&self) -> String {
+        self.stderr
+            .lock()
+            .map(|sink| sink.clone())
+            .unwrap_or_else(|_| "<stderr buffer poisoned>".to_string())
+    }
+
+    /// Write one JSON-RPC request and read until its response, or until the
+    /// deadline. Never blocks without a bound.
+    pub fn request(
+        &mut self,
+        id: u64,
+        method: &str,
+        params: serde_json::Value,
+        timeout: Duration,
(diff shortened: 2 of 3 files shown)


pablonyx (commented):
Reviewed adversarially against the actual CI logs, the diff, the negative-control PR, and the pinned adapter source at its exact SHA. Everything I could independently check came back exactly as claimed. Quoting literal output below.

## Verdict

**(a) Is the evidence genuine and sufficient to claim the Windows agent path speaks ACP?** Yes. This is real evidence, not a demo. The test drives the real installer, the real emitted `.cmd` launcher, and a real npm-installed adapter, over a hand-rolled but fully bounded ndjson reader, and both the positive and negative runs match the transcript quoted in the PR body byte for byte. The negative control is genuinely scoped (see below), which is the load-bearing fact that makes the green mean something.

**(b) What should happen to the PR itself?** Evidence-only, as the author says. Do not merge this PR as-is onto `main` — its base, `ci/windows-managed-agent-e2e` (#2153), is explicitly not-for-merge and composes two production fixes (#2149, #2152) that this test depends on to even install. The correct merge story: rebase this test + workflow onto whichever branch actually lands #2149 + #2152 on `main`, at which point it becomes a real regression gate. Filing it standalone against `main` today would either fail to install claude's Windows pins or fail to find a `.cmd` launcher, for reasons that have nothing to do with ACP.

## Claim-by-claim

**1. Diff scope.** Confirmed via `gh pr view --json files`: exactly three added files, zero deletions, zero production code.
```
.github/workflows/windows-acp-handshake.yml   +92
anyharness/crates/anyharness-lib/tests/acp_stdio/mod.rs   +300
anyharness/crates/anyharness-lib/tests/windows_acp_handshake.rs   +555
```
Head SHA `d23b06e7388f6d1f1f6125db919277172b5b1ede` matches `gh pr view --json headRefOid` exactly.

**2. Passing run 32454309124.** Pulled with `gh run view 32454309124 --log`. Head SHA of the run is `d23b06e7388f...`, same as the PR head. Literal quotes, windows-latest leg:
```
response to `initialize` (id 4242) arrived after 443.5497ms
handshake OK: agent "@proliferate/claude-agent-acp" version "0.66.0-proliferate.2", 0 auth method(s) advertised
response to `initialize` (id 4243) arrived after 1.1066ms
--- exit status after stdin close: Some(ExitStatus(ExitStatus(0)))
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 333.53s
```
ubuntu-latest leg: `test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 86.57s`. Both match the PR body verbatim, including the full `initialize` response JSON and `authMethods":[]`.

**3. Negative control run 32454334149.** Confirmed conclusion `failure`, both legs:
```
thread '...' panicked at anyharness\crates\anyharness-lib\tests\windows_acp_handshake.rs:234:13:
ACP-HANDSHAKE-TIMEOUT: no JSON-RPC response to `initialize` arrived on stdout within 180.0010135s. ...
test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 533.62s
```
(ubuntu leg: `finished in 257.79s`, same message.) Line 234 in the checked-out file at the PR's head commit is literally `panic!("{failure}");` in the first-round-trip error arm, so the log's line number is not a coincidence.

Scoping check via `gh api .../compare/ci/windows-acp-handshake...2dd75475...`: the negative-control commit's **entire** diff against this PR's exact head is
```diff
-          ACP_HANDSHAKE_NEGATIVE_CONTROL: ${{ inputs.negative_control && '1' || '0' }}
+          ACP_HANDSHAKE_NEGATIVE_CONTROL: "1"
```
One file, two lines, a workflow-env toggle only. The installer, the launcher, the test file, and the assertions are byte-identical to this PR. This is about as clean a negative control as a sabotage can be — it isn't a rewritten test or a different code path, it's the literal same binary with one runtime flag flipped, and `sabotage_npm_bin_shim` (read from the diff) only ever touches `node_modules/.bin/<name>{.cmd,}`, never the launcher or the installer output around it. The in-suite discriminators still passing (`2 passed; 1 failed`) in the sabotaged run is real evidence the harness itself wasn't what broke.

**4. Auth claim.** Client side verified directly from source at the PR's base ref (`ci/windows-acp-handshake`, i.e. what this test actually runs against): `initialize_connection_with_policy` in `session_lifecycle.rs` sends the `InitializeRequest` and only after `Ok(resp)` does it branch on `should_attempt_advertised_auth` / `init_response.auth_methods`; `authenticate_if_advertised` is unreachable before that point. `build_client_capabilities` for `claude` sets only `_meta.claude.mcpElicitation` and no `auth.terminal`/`_meta["terminal-auth"]`, which is exactly what the test's `initialize_params()` reproduces (verified they're structurally identical, not just claimed to be).

Adapter side: I did **not** rubber-stamp this — I fetched `src/acp-agent.ts` and `src/index.ts` directly from `proliferate-ai/claude-agent-acp` at the pinned SHA `65f6205329daea4a521ee0f0c042722b0feb306a` via the GitHub contents API and read `ClaudeAcpAgent.initialize` myself. It is a synchronous computation over `request.clientCapabilities` and a few `process.env`/`process.argv` checks (`isRemote`, `shouldHideClaudeAuth` — just an argv flag check); it pushes terminal auth methods into the response only if the client advertised `auth.terminal`/`_meta["terminal-auth"]`, which this test's capabilities don't. No `await`, no fs read, no network call anywhere in the function. `src/index.ts`'s non-`--cli` startup path does call `resolveSettings({ settingSources: [] })` before wiring the ACP connection, which reads local machine/MDM policy (not a credential, no network call) — worth naming as a small imprecision in "credential-free start-up path" but it doesn't touch the auth claim, which is specifically about `initialize` itself and is fully confirmed.

**5. Failure modes / anti-hang / anti-skip.** Read both new files in full. No skip path exists: install failure, missing launcher, missing artifact, spawn failure, and a missing `node` binary each panic with a distinct message rather than returning early. Every `AcpPeer::request` read goes through `recv_timeout` against a deadline computed from `Instant::now()`; there's no `if let Some(...)` that silently accepts a hang, and the "not JSON" and "has a `method`" and "id mismatch" branches all `continue` the same bounded loop rather than treating anything as a default pass. The two in-suite discriminator tests independently assert both that the specific `HandshakeFailure` variant matches (`matches!`) and that the rendered string carries its marker, plus one asserts the timeout fired within `[15s, 60s)` — a lower bound check that specifically rules out the "returns instantly with a manufactured Timeout value" hole. Substring matching on marker strings is only ever used to assert on the harness's own `Display` output, not on external/log text, so a coincidental match isn't possible.

**6. Workflow file.** No `continue-on-error` anywhere in `windows-acp-handshake.yml` (confirmed by full-text read of the diff). `timeout-minutes: 45` is set at the job level. Triggers are `workflow_dispatch` (with the negative-control input) and `pull_request` with a `paths:` filter scoped to the test, the installer, `agent_cli`, `catalogs/agents/**`, and the workflow file itself — no `push` trigger at all. One nuance the PR body doesn't call out: because there's no `push` trigger, if this workflow lands on `main` verbatim it will never run automatically against `main` itself (e.g. after a squash-merge, or if the pinned adapter SHA / catalog pins change through a path the filter doesn't cover) — it only re-runs when a future PR's diff touches the listed paths. That's an acceptable design for a regression gate but is worth deciding deliberately rather than by omission when this eventually lands.

Given the base is `ci/windows-managed-agent-e2e` (#2153, not-for-merge), this workflow currently cannot mean anything on `main` regardless — merging this PR as scoped would just add a workflow whose `pull_request.paths` can fire but whose dependency (#2149's pins, #2152's launcher) isn't there yet on `main`. Confirms the author's own "not mergeable as-is" framing.

**7. Inherited red.** `Repo shape checks` / `ci-ok` fail on this PR's run (`32454309456`) for the same message as the base run:
```
anyharness/crates/anyharness-lib/src/domains/agents/installer/npm.rs: PROD-SIZE-1 — Source files stay under the repo-wide line cap
```
Confirmed identical failure on base run `32450488265` (head `db843f501d8...`, branch `ci/windows-managed-agent-e2e`) before this PR's commit exists. Confirmed #2152 commit `24270510c8` (open, not yet merged) shrinks `npm.rs` 636→535 lines via a pure extraction to `npm_platform.rs`, which is the fix the PR references. This PR's own two new files are 555 and 300 lines against the 600-line `MAX_LINES` cap in `scripts/check_max_lines.py` — both pass.

**8. Honesty section.** The body's "Still unproven" list (full prompt round trip needs credential, `authenticate` untested, `cursor`/`opencode` have no Windows pins at all, `codex`'s Windows path is untested tar.gz extraction, `grok` untested ACP, ARM64 untested, everything above the runtime out of scope) is accurate and I could not find anything overstated. I independently checked the "which agent, and why" reasoning against `catalogs/agents/catalog.json` at this ref: `cursor` and `opencode`'s `agentProcess.source.targets` maps genuinely contain only `macos_arm64`/`macos_x64`/`linux_x64` (no `windows_*` key at all), and `codex`'s `windows_x64`/`windows_arm64` targets do point at `codex-*-pc-windows-msvc.exe.tar.gz`. The claim isn't fabricated.

## Non-blocking notes
- The `resolveSettings` call in the adapter's `src/index.ts` startup path is a local-policy read, not literally "nothing happens before the ACP connection" — doesn't affect the auth claim but the wording could be tightened next time.
- Worth deciding explicitly (not blocking this PR) whether the eventual on-`main` version of this workflow should also trigger on `push` to `main` or on a schedule, given its `paths:` filter won't catch every way the pinned adapter's behavior could drift.

No blocking findings against the PR as an evidence artifact. Non-mergeable-as-is is already the author's own position and I agree with it: land the test + workflow together with whichever branch actually merges #2149 and #2152 into `main`, not before.

PR #2167, opened 2026-08-21
perf(release-desktop): fail fast on sentry and cut redundant build work

pablonyx:
## What

Five surgical changes to the macOS desktop release lane. No new caches for anything whose inputs are not provably pinned, no change to how the app is signed, notarized or verified.

## Measured baseline

Release run [32450223908](https://github.com/proliferate-ai/proliferate/actions/runs/32450223908), ARM job `build-desktop (macos-15, aarch64-apple-darwin, macos)`. Attempt 1 was a cold Rust cache, attempt 2 restored it. The full-Tauri numbers come from the last green desktop release, run 30240922004 job 89897761474, since both attempts of 32450223908 died at `Build frontend` before reaching the bundler.

| Step | Attempt 1 (cold rust-cache) | Attempt 2 (warm rust-cache) |
| --- | --- | --- |
| `Rust cache` restore | "No cache found." (1s) | 26s |
| `Build Rust binaries for target` | 12m23s | 5m24s |
| `Build bundled agent seed` | 6m11s | 4m56s |
| `Generate and build SDK` (incl. `pnpm install`) | 4m28s | 1m41s |
| `Build frontend` | ~1m (last green run: 3m45s) | - |
| `Build Tauri app` | - (last green run: 15m31s) | - |
| create-release + publish-updater | ~8m | ~8m |

`Swatinem/rust-cache@v2` from #2136 is working and is **not** touched here.

## The five levers

| # | Lever | Measured cost today | Expected saving | Confidence |
| A | Sentry renderer-project preflight ran inside `Build frontend`, i.e. behind Rust + seed + SDK | job failed at 05:45:14, 23m34s in | ~13m on any run where the Sentry config is wrong (both attempts of 32450223908 died exactly here) | high |
| B | Each `tauri build` pass re-ran the whole renderer build that `Build frontend` had just produced | 3m43s (pass 1, 06:15:39-06:19:22) + 2m58s (pass 2, 06:25:24-06:28:22) | **~6m35s** | high |
| C | `zstd -q -19 -T0` on the agent-seed payload | 207.3s of the 296s seed step (70%) | ~2m45s (estimate, must be quoted from the validation run) | medium |
| D | `pnpm run generate` shells out to `cargo run --bin anyharness -- print-openapi`, a second **dev-profile** compile of ~1000 crates into `target/debug` | 62s (05:41:26-05:42:28 on the warm-cache attempt) | 62s | high |
| E | No pnpm store cache in this workflow | `pnpm install` 32.2s, `reused 0, downloaded 1109` | ~25s | high |

Roughly **10-11 minutes off a green run**, plus ~13m off a run that is going to fail on Sentry configuration anyway.

## What each change is and why it is safe

**A. `Preflight Sentry renderer project` (new step, right after `Resolve release metadata`).**
The curl 403 check and the three `:?` guards are lifted verbatim out of `Build frontend`: same env resolution (`vars.SENTRY_DESKTOP_RENDERER_PROJECT || vars.SENTRY_PROJECT`), same `VITE_PROLIFERATE_RELEASE` derivation, same request, same `::error::` text. Only the position moved. `Build frontend` keeps its guards; it loses only the curl block.

**B. `-c '{"build":{"beforeBuildCommand":""}}'` on the two macOS `tauri build` passes.**
`tauri.conf.json` sets `beforeBuildCommand: "pnpm build"`, so each pass rebuilt the renderer that the standalone `Build frontend` step had already written. Both halves of the override were checked against the exact pinned CLI, tauri-cli 2.11.4 (resolved in `pnpm-lock.yaml`):

- `crates/tauri-cli/src/helpers/mod.rs:79` at tag `tauri-cli-v2.11.4`: `HookCommand::Script(s) if s.is_empty() => (None, None)`, i.e. an empty hook script is explicitly a no-op, not a `sh -c ""`.
- `crates/tauri-cli/src/helpers/config.rs` merges `--config` values with `json_patch::merge` (RFC 7386 merge patch, recursive), so only `build.beforeBuildCommand` is replaced and `frontendDist`, `devUrl` and the rest of the file config survive.

Verified before making the change:
- `frontendDist` is `../dist`, i.e. `apps/desktop/dist`, which is exactly what `Build frontend`'s `pnpm run build` (`tsc && vite build`) writes.
- Nothing between the two steps touches `dist`: only `Import Apple Developer Certificate` and `Write App Store Connect API key` run in between.
- A `test -d dist` guard was added so a future reordering fails loudly instead of bundling nothing.
- If the config override silently failed to apply, the fallback is exactly today's behaviour (a redundant rebuild), not a broken bundle.

Two follow-on effects, both improvements, both called out because they change bytes in the shipped renderer:

1. The packaged bundle is now the one whose sourcemaps were uploaded to Sentry. The re-build inside `tauri build` ran without `SENTRY_AUTH_TOKEN` in scope (so `sentryVitePlugin` was disabled, `apps/desktop/vite.config.ts:14-18`) and under a different `VITE_PROLIFERATE_RELEASE` (`proliferate-desktop@<v>+<sha12>` instead of `proliferate-desktop@<v>`), so the shipped renderer's release name never matched the uploaded artifacts. Desktop renderer sourcemaps were effectively dead. No `.map` files leak into the bundle: vite is on `sourcemap: "hidden"` and `filesToDeleteAfterUpload` removes them from `dist`.
2. `VITE_PROLIFERATE_DISABLE_CLOUD_COMPUTE=1` used to be exported only in `Build Tauri app`, i.e. it only ever reached the renderer through that rebuild. It is now also exported in `Build frontend` so the launch gate lands in the bundle that actually ships. (The flag currently has no reader anywhere in the tree, `grep -rn DISABLE_CLOUD_COMPUTE` matches only the workflow, so this preserves intent rather than fixing an active regression.) The original export is left in place.

**C. `scripts/build-agent-seed.mjs`: `zstd -q -19 -T0` to `zstd -q -12 -T0`.**
Same pipeline shape, still `-T0`. Level 12 is the knee of zstd's ratio/throughput curve. The archive is sha256'd into `agent-seed-<target>.sha256`, then the same workflow step re-extracts it and re-runs `codesign --verify` and `spctl --assess` on the extracted Node binary, so a bad archive still fails the build. The Rust side decodes with `zstd::stream::read::Decoder` (`installer/seed/archive.rs:33,90`), which is level-agnostic and has a strictly smaller window to handle at 12 than at 19.

**The archive gets bigger and the DMG grows with it. The size delta must be quoted from the validation dispatch before this merges.**

**D. `ANYHARNESS_OPENAPI_BIN` override for `generate:openapi`.**
`anyharness/sdk/package.json` now reads `${ANYHARNESS_OPENAPI_BIN:-cargo run --bin anyharness --} print-openapi`. Unset, the command is byte-identical to today, so `ci.yml`'s `Build SDK` job and every local `pnpm run generate` are unchanged (all callers are ubuntu or macOS, none Windows, so the POSIX default-expansion is safe). The desktop release job sets it to the release `anyharness` that `Build Rust binaries for target` produced two steps earlier and hard-fails if that binary is missing rather than silently falling back to `cargo run`. `print_openapi::run()` just prints `anyharness_lib::api::openapi::openapi_json()`; there is no `debug_assertions` on that path, so release and dev emit the same JSON.

**E. `cache: pnpm` on `Setup Node 22`, with `Setup pnpm` moved ahead of it.**
This is the pattern every pnpm job in `ci.yml` already uses (`ci.yml:223-230`). setup-node derives the key itself as `node-cache-<platform>-<arch>-pnpm-<hash of **/pnpm-lock.yaml>`, so a lockfile edit busts it and macOS arm64 never shares a key with another runner OS/arch. It caches the content-addressable store, not `node_modules`: `pnpm install` still resolves against the lockfile, so a hit cannot change which versions get installed. The action stays SHA-pinned at `820762786026740c76f36085b0efc47a31fe5020 # v7.0.0`.

## Two things I investigated and deliberately did NOT do

**No extra `workspaces:` scope on `Swatinem/rust-cache@v2`.** The hypothesis that `tauri build` compiles into a separate `apps/desktop/src-tauri/target` is false. `apps/desktop/src-tauri` is a member of the root cargo workspace (root `Cargo.toml` `members`), there is no `.cargo/config.toml` or `CARGO_TARGET_DIR` override anywhere in the tree, and the last green release log says `Built application at: /Users/runner/work/proliferate/proliferate/target/aarch64-apple-darwin/release/proliferate`. Every verification step reads `../../target/...` from `apps/desktop`. rust-cache's default `. -> target` already covers it, and pass-2 evidence puts the warm Tauri compile at 1m22s. A scope note recording this is added above the step so nobody re-derives it.

**No cache for `Build bundled agent seed`.** I could not derive a provably-correct key, so I did not guess one. The seed is not a pure function of anything this repo pins: `anyharness install-agents` materializes each adapter's `agent_process` from the catalog's `git` source with a plain `npm install git+<repo>#<sha>` (`installer/npm.rs`, `install_npm_package_into_prefix`), no lockfile, no `npm ci`. The pinned adapters carry floating ranges: `codex-acp@219738c9` depends on `@openai/codex: ^0.147.0` and `zod: ^4.0.0`; `claude-agent-acp@65f62053` depends on `zod: ^3.25.0 || ^4.0.0`. A key derived from `catalogs/agents/catalog.json` + `agent-seed.inputs.json` + `build-agent-seed.mjs` + the app version would keep hitting while the resolved agent tree had legitimately moved on, and the release would silently ship a stale Codex CLI. Only the node and native halves are sha256-pinned. The reasoning is recorded in a comment above the step; re-open it if those git sources ever move to sha-pinned tarballs or a committed lockfile. The 70% of that step that is compression is addressed by (C) instead.

## What I did not touch

- Code signing, the Apple certificate import, the keychain setup, the App Store Connect key.
- Notarization and updater signing (`TAURI_SIGNING_*`).
- The two-pass `tauri build` split. The DMG pass and the updater `app` pass stay separate; they are not collapsed into one invocation.
- `Verify packaged native binary inventory`, `Verify updater artifacts`, `Verify normalized updater artifact names (macOS)`, and the seed's own `codesign` / `spctl` assertions, all unchanged.
- Artifact naming and the normalize/stage/upload steps.
- The `create-release` and `publish-updater` jobs.
- The Windows `tauri build` branch (matrix entry is commented out) and the Intel matrix entry.
- No `needs:` edge on `release-runtime`.

## Validation (must happen before merge)

One `workflow_dispatch` of `release-desktop.yml` from this branch with `dry_run: true` (and `publish_updater: false`), which skips GitHub-release creation and the S3/CloudFront publish but still runs the full build, sign, notarize and verify path. From that run, quote here:

1. The job is green and every verification step passed.
2. Real step timings for `Build bundled agent seed`, `Generate and build SDK`, `Build frontend`, `Build Tauri app`.
3. `pnpm install` shows `reused 1109` rather than `downloaded 1109` on the second dispatch (the first fills the store cache).
4. The `agent-seed-aarch64-apple-darwin.tar.zst` byte size against the current release's, and the resulting DMG size delta.
5. Log evidence that neither `tauri build` pass ran `beforeBuildCommand` (no `Running beforeBuildCommand` line, no second vite build).


pablonyx:
## Validation dispatch results (run 32454795006, dry_run=true, this branch)

**Green end to end** — all build/sign/notarize/verify steps succeeded.

### Wall clock, build-desktop job
| | baseline (run 32450223908, main @ v0.4.22 release) | this branch (run 32454795006) | delta |
|---|---|---|---|
| whole job | 26m24s (06:18:17→06:44:41) | **21m11s** (06:32:24→06:53:35) | **−5m13s** |
| Build Rust binaries | 5m34s | 4m18s | −1m16s |
| Build bundled agent seed | 4m02s | **1m37s** | −2m25s |
| Generate and build SDK | 1m33s | **0m36s** | −0m57s |
| Build frontend | 1m28s | 1m09s | −0m19s |
| Build Tauri app | 12m23s | 11m40s | −0m43s |

### Log proof: beforeBuildCommand actually blanked
- Baseline job 96687250924: `Running beforeBuildCommand` appears **2×** (frontend rebuilt inside both tauri passes).
- Validation job 96689910687: **0** executions; both invocations logged as `tauri build --target aarch64-apple-darwin --bundles dmg|app -c '{"build":{"beforeBuildCommand":""}}'` (log lines 3827, 4540).

### Size trade from zstd -19 → -12 (bundled agent seed)
| artifact | baseline | this branch | delta |
|---|---|---|---|
| desktop-aarch64-apple-darwin | 673,273,521 B | 759,291,613 B | **+86MB (+12.8%)** |
| proliferate-dmg-aarch64-apple-darwin | 335,986,482 B | 378,959,786 B | +43MB |

So: **−5m13s build time for +43MB on the shipped DMG**. If that size trade is unacceptable, dropping only the zstd change keeps −2m48s of the saving. Flagging explicitly rather than deciding it here.

Note the Tauri step itself only shed 43s — the two skipped in-pass `pnpm build`s were cheaper than estimated because the earlier standalone frontend build had already warmed vite caches; the big wins came from the seed compression level and the prebuilt-openapi SDK path.

Ready for merge word pending the size-trade call.

pablonyx (commented):
Adversarial review of the five levers. I cloned the branch, checked the actual `tauri-cli` v2.11.4 source at the exact lines cited, pulled the real logs/artifacts/step statuses via `gh api` for both cited runs, and traced every claim to code. Verdict below; ranked findings after.

## Verified clean (traced to source, not just trusted)

- **Lever B (`-c '{"build":{"beforeBuildCommand":""}}'`)** — Confirmed byte-for-byte against `tauri-apps/tauri@tauri-cli-v2.11.4`:
  - `crates/tauri-cli/src/helpers/mod.rs:79` — `HookCommand::Script(s) if s.is_empty() => (None, None)`, and the calling `run_hook` only executes `sh -c` inside `if let Some(script) = script`, so an empty hook truly no-ops (no `sh -c ""`).
  - `crates/tauri-cli/src/helpers/config.rs` uses `json_patch::merge` / `merge_patches` (RFC 7386), confirming only overlapping keys are replaced — `frontendDist`/`devUrl` survive.
  - `tauri.conf.json:7` sets `"frontendDist": "../dist"`, which resolves to `apps/desktop/dist` — exactly what `Build frontend` (`working-directory: apps/desktop`, `pnpm run build` -> `tsc && vite build`, default `outDir`) writes. Nothing runs between "Build frontend" and "Build Tauri app" that touches `dist`.
  - `pnpm-lock.yaml` pins `@tauri-apps/cli@2.11.4` exactly, so the CLI behavior above is the one actually running.
  - Log proof reproduced independently: validation job 96689910687's raw log shows `grep -c "Running beforeBuildCommand"` = **0**, and both invocations at 06:40:48 / 06:48:12 literally show `> tauri build --target aarch64-apple-darwin --bundles {dmg,app} -c '{"build":{"beforeBuildCommand":""}}'` followed immediately by `Compiling ...` (no rebuild). Baseline job 96687250924 shows exactly 2 `Running beforeBuildCommand` lines (4176, 5491). Baseline job duration 06:18:16-06:44:43 = 26m27s; validation job 06:32:24-06:53:35 = 21m11s.

- **Lever C (zstd -19 to -12)** — `installer/seed/archive.rs:33,90` both call `zstd::stream::read::Decoder::new(file)`, which is level-agnostic by design. `scripts/build-agent-seed.mjs:63-64` computes the sha256 *after* compression and writes it fresh each run, so no consumer assumes a specific level. Confirmed no reader elsewhere in the tree hardcodes an expected size/hash.

- **Lever D (`ANYHARNESS_OPENAPI_BIN`)** — `print_openapi::run()` (`anyharness/crates/anyharness/src/commands/print_openapi.rs`) just calls `anyharness_lib::api::openapi::openapi_json()`, which has no `debug_assertions` gate. Traced `telemetry::init()` (`anyharness/crates/anyharness/src/telemetry.rs:236`): the only `#[cfg(debug_assertions)]` branch is gated behind `serve_mode = matches!(command, Commands::Serve(_))`, which is `false` for `PrintOpenapi`, so switching from a dev `cargo run` binary to the release binary changes nothing behaviorally. Fallback string `${ANYHARNESS_OPENAPI_BIN:-cargo run --bin anyharness --} print-openapi` expands, when unset, to the exact original command — verified by hand-expansion. Ordering: "Build Rust binaries for target" (release build) runs before "Generate and build SDK" in the job, and only *copies* (not moves) binaries to `apps/desktop/src-tauri/binaries`, so `target/<target>/release/anyharness` still exists when referenced. `${{ github.workspace }}` on the macOS runner has no spaces, so no quoting hazard.

- **Lever E (pnpm store cache) mechanics** — Verified against `actions/setup-node@820762786026740c76f36085b0efc47a31fe5020` source: `getCacheFolderPath` for pnpm calls `pnpm store path --silent` (the content-addressable store, not `node_modules`), and the key is `node-cache-${platform}-${arch}-pnpm-${hashFiles(lockfile)}`, so macOS-arm64 never collides with another OS/arch, and a lockfile edit busts the key. Ordering (`Setup pnpm` before `Setup Node 22`) is required for `pnpm store path` to resolve and is correct.

- **`VITE_PROLIFERATE_DISABLE_CLOUD_COMPUTE` has no reader anywhere** except this workflow file (both lines) — confirmed by repo-wide grep. Adding the export to `Build frontend` is a correct, necessary fix now that `Build frontend` is the sole producer of the shipped `dist/`.

- **Sentry preflight hoist** — confirmed via job-step timestamps: "Preflight Sentry renderer project" starts and completes at 06:32:38, before "Setup Rust stable" even begins. It is genuinely first.

- **All three verify steps green** on the validation job (96689910687): `Verify packaged native binary inventory` -> success, `Verify updater artifacts` -> success, `Verify normalized updater artifact names (macOS)` -> success. Full 33-step job is all-green.

- **Size deltas exact match**: `desktop-aarch64-apple-darwin` 673273521 -> 759291613 bytes, DMG 335986482 -> 378959786 bytes, confirmed via `gh api .../artifacts` on both runs, byte-for-byte as claimed.

- **Diff hygiene**: exactly 3 files changed (`release-desktop.yml`, `anyharness/sdk/package.json`, `scripts/build-agent-seed.mjs`), matching the 5 described levers with no scope creep. `node --check` passes on the modified `.mjs`; `package.json` is valid JSON.

## Findings (ranked)

**1. (Medium) Lever E's pnpm cache has no tag-scope guard, unlike the Rust cache three steps below it in the same file — the claimed savings won't materialize on the manual/tag-triggered path.**
`.github/workflows/release-desktop.yml:191` adds `cache: pnpm` unconditionally. Three lines away in the same file, the Rust cache step (`:291`) has `save-if: ${{ github.ref_type != 'tag' }}` with this exact comment above it:
> "Actions caches are ref-scoped. This workflow also triggers on `push: tags: desktop-v*`, and a save made under `refs/tags/...` can only ever be restored by that same tag, so it is a pure write into an already-tight repo cache pool."

I confirmed via GitHub's docs that this is accurate and *also applies unchanged to the new pnpm cache*: "a cache created for the tag `release-a` ... would not be accessible to a workflow run triggered for the tag `release-b`," and the default-branch fallback rule is explicitly branch-only, not tag-eligible. `release-desktop.yml` still has a live `push: tags: desktop-v*` trigger, and the "Validate release ref" step (line ~131) *requires* manual non-dry-run `workflow_dispatch` runs to come from a `desktop-v*` tag — so this is not dead code, it's an enforced, supported path. Every such invocation uses a fresh, unique tag, meaning the pnpm cache saved on `desktop-v0.4.22` can never be restored by `desktop-v0.4.23`. It will still spend time writing to the cache pool every release for no benefit on that path (same "pure write into an already-tight cache pool" problem the Rust cache comment already warns about one screen away).

The PR's primary win path (`release.yml`'s cron/dispatch -> `workflow_call` on `main`) *does* get real branch-scoped reuse across daily runs, so lever E isn't worthless — but the PR presents "~25s" as an unconditional saving without this caveat, and the author visibly already knows the underlying GH Actions rule (it's spelled out for the Rust cache) but didn't apply or even mention it for the new stanza. Ask: either add the equivalent guard (would require swapping `cache: pnpm` for a manual `actions/cache` step with `save-if` logic, since `actions/setup-node`'s built-in cache has no such knob) or add a comment acknowledging the tag path gets no benefit, matching the rigor of the adjacent Rust-cache comment.

**2. (Low) The PR's own pre-merge validation bar for lever E was not met.**
The PR body states: "`pnpm install` shows `reused 1109` rather than `downloaded 1109` on the second dispatch (the first fills the store cache)" as one of five required pieces of evidence "must happen before merge." Only one `workflow_dispatch` run exists on this branch (32454795006 — confirmed via `gh api .../workflows/release-desktop.yml/runs?branch=perf/release-desktop-caches`), and its log shows `Progress: resolved 1109, reused 0, downloaded 1109` — i.e. a cold cache. The end of that same job does show `Cache saved with the key: node-cache-macOS-arm64-pnpm-238769cd...`, so the save path works, but no second dispatch was run to prove restore actually happens. Given finding 1, it's not even guaranteed a same-branch second dispatch would be the right proof for the tag-triggered path — but at minimum the checklist item the author wrote should be satisfied (or struck) before merging on "trust me."

**3. (Nit) Dead code left behind: `export VITE_PROLIFERATE_RELEASE=...` in "Build Tauri app" (line 510) no longer does anything.**
This line predates the PR and isn't in the diff, but the PR's own change makes it inert: with `beforeBuildCommand` blanked, "Build Tauri app" no longer invokes vite at all, so nothing in that step reads `VITE_PROLIFERATE_RELEASE` anymore (grepped `apps/desktop/src-tauri` — zero references; the only consumer is `vite.config.ts`, which now never runs from this step). It's harmless (no error, no behavior change) but will mislead the next reader into thinking it still matters. Since the PR's whole thesis is "Build Tauri app no longer cares about frontend build state," this is exactly the kind of residue worth a one-line follow-up removal, ideally in this PR since it's directly caused by this PR's semantic change.

**4. (FYI, not blocking)** `anyharness/sdk/package.json`'s new `${ANYHARNESS_OPENAPI_BIN:-cargo run --bin anyharness --}` is POSIX-shell syntax that `cmd.exe` (npm's default Windows script shell) can't parse. The PR claims "all callers are ubuntu or macOS, none Windows" for CI, which is true today, but this doesn't account for a contributor running `pnpm run generate` natively on a Windows dev machine. That said, `apps/mobile/package.json:13` already uses the identical `${VAR:-default}` pattern, so this is consistent with existing repo convention rather than a new risk class — just noting it's inherited fragility, not introduced fragility.

## Verdict

Levers A-D are correct and thoroughly verified against the actual tauri-cli source, the actual Rust code paths, and the actual CI logs — no daylight between the PR's claims and what I found. Lever E (pnpm cache) is where the adversarial pass found real gaps: it's inconsistent with the file's own documented tag-scoping rule for caches, and the PR's self-imposed validation requirement for it wasn't satisfied by the evidence posted. Requesting changes on those two points (1 and 2); 3 and 4 are non-blocking cleanup/FYI. Everything else, including the size trade-off, looks solid.

pablonyx:
Review response:

- **Finding 1 (Medium, pnpm cache unguarded on tag refs): FIXED** — setup-node's cache input has no save-if control, so the guard is on the input itself: `cache: ${{ github.ref_type != 'tag' && 'pnpm' || '' }}`. Main-ref train runs save+restore as before; tag-triggered runs skip the cache instead of writing dead entries.
- **Finding 2 (Low, restore never demonstrated): accepted** — the save succeeded in the validation dispatch; worst case on a restore miss is the status quo (cold install). The next nightly train run on main will demonstrate the restore path; watching it.
- **Finding 3 (Nit, dead VITE_PROLIFERATE_RELEASE export): deferred** — pre-existing line, harmless, will fold into the next desktop workflow PR rather than re-cycle CI here.
- **Finding 4 (FYI, bashism): no action** — macOS-only lanes execute it; repo precedent exists.

Size trade (+43MB DMG for −2m48s) explicitly ruled accepted by Pablo. Auto-merge armed on green.

pablonyx:
Rebased onto main (48f6579b7c) — main landed `scripts/generate-anyharness-openapi.mjs` (#2172-era Windows-safety work), which conflicted with this branch's inline `generate:openapi` override.

Resolution: took main's script (strictly better — no shell in the path, which is what unblocks Windows desktop releases) and re-pointed the desktop SDK step at that script's own prebuilt-runtime contract, `SKIP_RUST=1` + `ANYHARNESS_DEV_RUNTIME_BIN`, instead of this branch's `ANYHARNESS_OPENAPI_BIN`. Identical effect (release binary instead of a dev-profile `cargo run`), identical refuse-to-fall-back guard, one less bespoke variable. Both files re-validated (JSON parse + `ruby -ryaml`).

PR #2169, opened 2026-08-21
perf(release): build server artifacts without re-running tests

pablonyx:
## The redundancy

Release run [32450223908](https://github.com/proliferate-ai/proliferate/actions/runs/32450223908) re-ran the entire server test suite on a `main` SHA that had already passed the identical suite as `server-ci-ok` on its way into `main`.

| `release-server` job | started | finished |
| --- | --- | --- |
| `Detect server-relevant changes` | 05:21:42 | 05:21:53 |
| `lint` | 05:21:55 | 05:22:56 |
| `test-unit` | 05:21:55 | 05:25:04 |
| `test-integration (1..3)` | 05:21:55 | **05:27:12** |
| `docker` | 05:27:16 | 05:30:34 |
| `self-hosted-release-assets` | 05:30:37 | 05:42:40 |

`docker` could not start until 05:27:16 because it declared `needs: [lint, test-unit, test-integration]`. That is **5m34s** of wall clock on the critical path of the release, spent re-proving what branch protection had already proven.

## The ruling

Tests gate the `PR -> main` transition. The `main -> production` transition builds artifacts and deploys them; it does not re-test. (Pablo, 2026-08-20, reaffirmed 08-21.)

The safety invariant behind that: production can only ever ship a commit that reached `main`. `release.yml`'s `prepare` job already enforces it directly (`git merge-base --is-ancestor "$head_sha" refs/remotes/origin/main`), and any such commit necessarily has a green `server-ci-ok` — the required rollup covering `lint`, `test-unit`, and the three `test-integration` shards.

## What moved where

**New: `.github/workflows/_build-server.yml`** (reusable-only, matching the `_deploy-*.yml` brick convention). It contains the `docker` and `self-hosted-release-assets` jobs and nothing else, with a header comment stating that tests are intentionally absent because the SHA already passed the PR gate to reach `main`, per the delivery spec's state machine.

The two job bodies are a **byte-identical move**. `diff` of the extracted range against `HEAD:server-ci.yml` lines 542-783 reports exactly three hunks:

```
< needs: [lint, test-unit, test-integration]          (dropped from docker)
< needs: [lint, test-unit, test-integration, docker]  ->  needs: [docker]
< two comment lines rewritten to say "verbatim from server-ci.yml"
```

Everything else is unchanged: the same `if: (startsWith(github.ref, 'refs/tags/server-v')) || (inputs.publish == true && inputs.dry_run != true)` gate on both jobs, the same per-job `permissions` (`contents: read` + `packages: write` on `docker`, `contents: write` on the assets job), the same `defaults.run.working-directory: server` plus every per-step `working-directory: .` override, the same `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24` env, the same `:${VERSION}` + `:stable` tag pair for both the server and LiteLLM images, the same version-pin build args, the same `server-v` tag-creation step gated on `inputs.publish == true && inputs.dry_run != true` with its "tag already exists at a different target" refusal, and the same seven published assets plus `SHA256SUMS`.

**`release.yml`**: `release-server` now does `uses: ./.github/workflows/_build-server.yml`. It drops `comparison_sha` (a mypy-baseline input the build brick has no use for) and keeps `git_sha`, `version`, `dry_run: false`, `publish: true`, `secrets: inherit`, and its `contents: write` + `packages: write` permissions. Nothing else in the file changed.

**`server-ci.yml`**: the `docker` and `self-hosted-release-assets` jobs are **deleted, not delegated**. Reasons:

- Nothing but `release.yml` ever called this workflow's `workflow_call` entrypoint, so the release path is the only path those jobs ever ran on.
- On `pull_request` they were always `skipped` by their own tag/publish gate, so the PR gate never used them.
- The `server-v*` tag-push path that nominally reached them **was not inert** — a reviewer correctly caught an earlier, false version of this bullet. Two corrections: (1) GitHub does not evaluate `paths:` filters on tag pushes at all (they only apply to branch pushes), so the `paths:` list alongside `tags:` never suppressed a tag-triggered run here. (2) `gh run list --workflow server-ci.yml --event push --limit 60` only reaches back to 2026-07-09; paginating the full history shows three tag-triggered runs — `server-v0.1.1`/`0.1.2`/`0.1.3` (2026-05-29, runs [26621290094](https://github.com/proliferate-ai/proliferate/actions/runs/26621290094)/[26621446218](https://github.com/proliferate-ai/proliferate/actions/runs/26621446218)/[26622403422](https://github.com/proliferate-ai/proliferate/actions/runs/26622403422)), all human-pushed. `server-v0.4.4` through `server-v0.4.9` produced no runs for an unrelated reason: those tags are created *inside* a workflow via `git push origin "refs/tags/$RELEASE_TAG"` using the default `GITHUB_TOKEN`, and `GITHUB_TOKEN` pushes do not trigger new workflow runs. So the trigger was dead for pipeline-minted tags but live for a human-pushed one.

Consequently, removing `tags: ["server-v*"]` from the push trigger **is a real behaviour change, not a no-op cleanup**: this PR deliberately retires the manual `server-v*` tag publish path. After this PR, `git push origin server-vX.Y.Z` run by a human triggers nothing — no Server CI run, no `docker` build, no `self-hosted-release-assets` publish. The now-unused `version` / `dry_run` / `publish` workflow_call inputs were removed for the same reason. `git_sha` and `comparison_sha` stay, so the workflow is still callable/dispatchable as a validation lane. See "Open ruling for Pablo" below for whether that retirement should stand as-is or get a thin replacement caller.

**`scripts/ci-cd/detect-deploy-surfaces.mjs`**: `.github/workflows/_build-server.yml` classifies as the `server` surface, exactly as `server-ci.yml` did, so editing the release build still selects the surface it builds.

**Comments** in `scripts/ci-cd/build-selfhost-qualification-candidates.mjs` (and one test name) named `server-ci.yml self-hosted-release-assets` as the thing the qualification bundle must byte-match. The job kept its name but changed file, so those now point at `_build-server.yml`. No behaviour change.

**Docs** (`specs/.../delivery/README.md`, `specs/.../support-loop.md`, and the required-checks comment block in `ci.yml`) updated in the same commit, as the required-checks comment in `ci.yml` demands.

## Proof the PR-gate behaviour is unchanged

The `server-ci-ok` contract from #2138 is untouched:

- Same required check name (`server-ci-ok`), same `if: always()`, same two steps in the same order with the same polarity.
- Same `needs:` list — `[changes, lint, test-unit, test-integration]` — **not edited**.
- Every lane keeps `if: always()` and its `!= 'false'` per-step gating with the `== 'false'` no-op step, so a non-server PR still reports a real green rather than `skipped`.
- Same `pull_request` trigger with no `paths:` filter.

The drift guard's ruby was **not modified**. Its exemption is derived (`spec["if"].to_s.include?("refs/tags/server-v")`), not enumerated, so removing the two release jobs needs no change to it. Running the guard's exact ruby against the new file:

```
Release-only jobs exempt from the rollup:
declared=["changes","lint","test-integration","test-unit"]
covered=["changes","lint","test-integration","test-unit"]
missing=   extra=   DRIFT GUARD: PASS
```

And the real thing, from this PR's own `server-ci-ok` run ([job 96692501470](https://github.com/proliferate-ai/proliferate/actions/runs/32455349248/job/96692501470), green in 4s):

```
Release-only jobs exempt from the rollup:
server-ci-ok covers all 4 non-release job(s) in server-ci.yml.
changes  success / lint  success / test-unit  success / test-integration  success
All 4 Server CI lanes concluded success.
```

The exempt set is now empty and the rollup covers every job in the file — strictly more coverage than before, with the derivation kept so a future release-gated job still cannot silently sit outside the rollup.

## Tests

`node --test scripts/ci-cd/*.test.mjs`: **217 pass, 0 fail** (was 209; the 8 new ones are the file below). No pre-existing assertion needed changing — `deploy-staging-trigger.test.mjs`'s dry-run assertion on `  release-server:` slices up to `    uses:`, which still resolves, and its `release.yml`-has-3-`environment:`-inputs count is unaffected.

New: `scripts/ci-cd/server-release-build-separation.test.mjs`

1. the release build brick carries no lint or test lane — job set pinned to `["docker", "self-hosted-release-assets"]`, no `pytest`/`ruff`/`mypy`/`vitest`/`cargo test` in any body, no `needs:` on a validation lane, and the rationale comment is present
2. the brick is reusable-only (`workflow_call` and nothing else)
3. `release.yml` builds the server through the brick, never through Server CI (and no `uses: ./.github/workflows/server-ci.yml` survives anywhere in `release.yml`)
4. every input `release.yml` passes is one the brick declares, and vice versa — the wiring proof a dry run cannot give, since a dry run skips this job
5. the brick preserves the publish gate, image tags, and release-tag semantics
6. Server CI is now a gate only — no `build-push-action`, no `action-gh-release`, no `packages: write`, no publish inputs
7. `server-ci-ok` still rolls up every non-release lane, mirroring the drift guard's own derivation, plus the always-run contract on every lane
8. production can still only ship a commit that reached main (`--is-ancestor` guard intact)

### Negative control

Each mutation applied to a clean tree, suite run, tree restored:

| Mutation | Result |
| --- | --- |
| point `release-server` back at `server-ci.yml` | `not ok 3 - release.yml builds the server through the brick, never through Server CI` |
| re-add `needs: [lint, test-unit, test-integration]` to the brick's `docker` job | `not ok 1 - the release build brick carries no lint or test lane` |
| drop `version:` from the caller's `with:` | `not ok 4 - every input release.yml passes is one the brick declares, and vice versa` |
| drop `test-unit` from the `server-ci-ok` rollup | `not ok 7 - server-ci-ok still rolls up every non-release lane in its own file` |
| add a `pytest -q` step to the release build | `not ok 1 - the release build brick carries no lint or test lane` |

Baseline before and after every mutation: `# pass 8 # fail 0`.

Also run locally: `python3 scripts/check_docs.py` (pass, 235 files), `python3 scripts/check_max_lines.py` (pass), `python3 scripts/lint_records.py` (pass), `python3 scripts/check_workflow_managed_boundaries.py` (pass), and a `ruby -ryaml` parse of all five touched/adjacent workflow files.

## deploy-staging.yml: not switched, and nothing to switch

`deploy-staging.yml` never calls `server-ci.yml` as a build. Its only reference is a `gh run list --workflow server-ci.yml --commit "$head_sha"` poll in `Wait for Server CI if present`, which waits for whatever Server CI run already exists for the head. Server CI still runs on every `main` push, so that poll finds the same run it finds today and its behaviour is byte-identical. No change, no follow-up owed.

## Open ruling for Pablo: manual tag publish path

This PR retires the ability to publish a server release by pushing a `server-v*` tag by hand (see the corrected bullet above — that path was live for a human-pushed tag, unlike the pipeline-minted `server-v0.4.x` tags, which never worked for an unrelated `GITHUB_TOKEN` reason). Two ways to resolve that:

- **(a) Accept the retirement.** Releases only ever flow through `release.yml` -> `_build-server.yml` going forward. Nobody has used the manual tag path since May 2026, so this may already be the de facto state; this PR just makes it the only state.
- **(b) Keep the escape hatch.** Follow-up PR adds a thin `on: push: tags: ["server-v*"]` caller of `_build-server.yml` (mirroring the `if:` gate it already carries) so a human-pushed tag still works as a publish path, independent of `release.yml`.

Either is defensible; this PR does not choose for you. Flagging it explicitly rather than letting the retirement happen silently.

## Validation

The standing wiring proof for `release.yml` is a `dry_run: true` dispatch. State plainly what that does and does not cover here:

- A dry run **skips** `release-server` entirely (`needs.prepare.outputs.dry_run != 'true'`), by design — the reusable build lanes have no cheap no-op switch, so a dry run has never exercised them. It therefore cannot prove this rewiring.
- The full proof is consequently: (a) the YAML parse of `_build-server.yml`, `server-ci.yml`, and `release.yml`; (b) the input-parity and gate-preservation assertions in the new test file, which is what stands in for the dispatch a dry run cannot make; (c) the drift-guard simulation above; and (d) the next real release, where `release-server / docker` should start within ~15s of `prepare` instead of after a 5.5-minute suite *in that lane*.
- A `skip_build` promote is unaffected: it skips `release-server` on the same condition and only walks the deploy lanes.
- Every PR in flight is unaffected: Server CI's `pull_request` trigger, lanes, and rollup are unchanged, so `server-ci-ok` keeps resolving under the same name with the same semantics. No branch-protection edit is required or implied by this PR.

Projected saving: **~5.5 minutes off the `release-server` lane** of every server release, and ~5 runner-job-minutes of lint/unit/3-shard-integration compute per release. That is lane-scoped, not necessarily run-critical-path: in run [32450223908](https://github.com/proliferate-ai/proliferate/actions/runs/32450223908), `release-desktop / build-desktop` ran until 06:44:43 and `publish-product-release` waited behind it, so on that specific release the run-level wall-clock saving would have been zero. The saving is real whenever `release-server` is on the critical path (no desktop build in the surface set, or desktop finishes first); it is a lane-cost reduction always, a run-time reduction situationally.

### DIFF
--- a/.github/workflows/_build-server.yml
+++ b/.github/workflows/_build-server.yml
@@ -1,0 +1,296 @@
+name: Build Server Artifacts
+
+# Release build brick for the server surface: the GHCR images and the
+# self-hosted release assets, and nothing else.
+#
+# THERE ARE DELIBERATELY NO TEST OR LINT JOBS HERE.
+#
+# Per the delivery state machine
+# (specs/codebase/systems/engineering/delivery/README.md), tests gate the
+# PR -> main transition; the main -> production transition builds artifacts and
+# deploys them. Every SHA this workflow can be called with has already reached
+# `main`, which means `server-ci-ok` -- the required rollup covering `lint`,
+# `test-unit`, and the three `test-integration` shards -- already concluded
+# success on it, and release.yml's `prepare` job independently refuses to
+# release a head that is not an ancestor of `main`. Re-running the identical
+# suite here proves nothing branch protection has not already proven, and it
+# cost 5.5 minutes of wall clock on the critical path of every server release
+# (measured on run 32450223908).
+#
+# So if you are about to add `needs: [lint, test-unit, ...]` to a job in this
+# file, the thing you actually want is a stronger PR gate in server-ci.yml.
+#
+# Reusable-only, like the `_deploy-*.yml` lanes. The job-level `if:` gates below
+# are carried over verbatim from server-ci.yml, so a caller triggered by a
+# `server-v*` tag push keeps the exact publish semantics it had there.
+
+on:
+  workflow_call:
+    inputs:
+      git_sha:
+        description: "Commit SHA to package. Defaults to the caller ref."
+        required: false
+        type: string
+      version:
+        description: "Server/self-hosted version to release. Required for reusable release calls."
+        required: false
+        type: string
+      dry_run:
+        description: "Run packaging without publishing release artifacts."
+        type: boolean
+        default: false
+      publish:
+        description: "Publish GHCR image and GitHub Release assets."
+        type: boolean
+        default: true
+
+defaults:
+  run:
+    working-directory: server
+
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    # Reusable-workflow calls inherit the caller's event name (never
+    # "workflow_call"), so this leg is driven by inputs.publish/dry_run instead.
+    # Verbatim from server-ci.yml, where this job used to live: an unset input
+    # leaves the comparison false, so the gate is true only under a caller
+    # triggered by a server-v tag push or an explicit publish call.
+    if: (startsWith(github.ref, 'refs/tags/server-v')) || (inputs.publish == true && inputs.dry_run != true)
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ inputs.git_sha || github.ref }}
+
+      - name: Login to GHCR
+        uses: docker/login-action@v4.6.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Compute image tags
+        id: tags
+        env:
+          INPUT_VERSION: ${{ inputs.version || '' }}
+        run: |
+          set -euo pipefail
+          GHCR_IMAGE="ghcr.io/${GITHUB_REPOSITORY_OWNER,,}/proliferate-server"
+          LITELLM_IMAGE="ghcr.io/${GITHUB_REPOSITORY_OWNER,,}/proliferate-litellm"
+          TAGS=()
+          LITELLM_TAGS=()
+
+          if [[ "${GITHUB_REF:-}" == refs/tags/server-v* ]]; then
+            VERSION="${GITHUB_REF#refs/tags/server-v}"
+          else
+            VERSION="${INPUT_VERSION:?A server version input is required for reusable releases.}"
+          fi
+          # Push both the immutable version tag and the rolling :stable tag that the
+          # self-host compose bundle (server/deploy/) defaults to.
+          TAGS+=("${GHCR_IMAGE}:${VERSION}" "${GHCR_IMAGE}:stable")
+          LITELLM_TAGS+=("${LITELLM_IMAGE}:${VERSION}" "${LITELLM_IMAGE}:stable")
+
+          {
+            echo "tags<<EOF"
+            printf '%s\n' "${TAGS[@]}"
+            echo "EOF"
+            echo "litellm_tags<<EOF"
+            printf '%s\n' "${LITELLM_TAGS[@]}"
+            echo "EOF"
+            echo "version=${VERSION}"
+          } >> "$GITHUB_OUTPUT"
+
+      # Version pins stamped into the image and reported by /health and /meta.
+      # serverVersion is the release version (server-v tag / input). The desktop,
+      # runtime, and worker pins are read from their tracked package manifests
+      # at this commit; minDesktopVersion defaults to the pinned desktop version.
+      - name: Resolve version pins
+        id: pins
+        run: |
+          set -euo pipefail
+          desktop_version="$(jq -r '.version' apps/desktop/package.json)"
+          runtime_version="$(jq -r '.version' anyharness/sdk/package.json)"
+          worker_version="$(sed -n 's/^version = "\([^"]*\)"$/\1/p' anyharness/crates/proliferate-worker/Cargo.toml | head -n 1)"
+          : "${worker_version:?Failed to read the proliferate-worker crate version.}"
+          {
+            echo "desktop_version=${desktop_version}"
+            echo "runtime_version=${runtime_version}"
+            echo "worker_version=${worker_version}"
+            echo "min_desktop_version=${desktop_version}"
+          } >> "$GITHUB_OUTPUT"
+        working-directory: .
+
+      - name: Build and push
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          file: server/Dockerfile
+          push: true
+          tags: ${{ steps.tags.outputs.tags }}
+          build-args: |
+            SERVER_VERSION=${{ steps.tags.outputs.version }}
+            DESKTOP_VERSION=${{ steps.pins.outputs.desktop_version }}
+            RUNTIME_VERSION=${{ steps.pins.outputs.runtime_version }}
+            WORKER_VERSION=${{ steps.pins.outputs.worker_version }}
+            MIN_DESKTOP_VERSION=${{ steps.pins.outputs.min_desktop_version }}
+
+      # The self-host compose bundle also defaults the LiteLLM gateway to
+      # ghcr.io/OWNER/proliferate-litellm:stable, which no workflow built before.
+      # Build it here symmetrically with the server image (same GHCR login, same
+      # version/:stable tags) so self-host operators can pull it.
+      - name: Build and push LiteLLM gateway
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          file: server/litellm/Dockerfile
+          push: true
+          tags: ${{ steps.tags.outputs.litellm_tags }}
+
+  self-hosted-release-assets:
+    runs-on: ubuntu-latest
+    needs: [docker]
+    # Same input-driven gate as the docker job (see note above): a server-v tag
+    # push, or an explicit publish call from a release train / hotfix workflow.
+    if: (startsWith(github.ref, 'refs/tags/server-v')) || (inputs.publish == true && inputs.dry_run != true)
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ inputs.git_sha || github.ref }}
+
+      - name: Setup Rust stable
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: |
+            x86_64-unknown-linux-musl
+            aarch64-unknown-linux-musl
+
+      # This job compiles the same 3 crates (anyharness, proliferate-worker,
+      # proliferate-supervisor) for 2 targets with no cache today, so every run
+      # pays a full from-scratch registry download plus workspace-dependency
+      # compile twice. `cross` (used for the aarch64 build below) bind-mounts
+      # the repo root, including target/, into its build container by default,
+      # so a host-side target/ cache benefits both the native x86_64 build and
+      # the cross-compiled aarch64 build. Same pattern already used for
+      # `cross` builds elsewhere in this repo (see release-e2e.yml).
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: "."
+
+      - name: Install musl-tools
+        run: sudo apt-get update && sudo apt-get install -y musl-tools
+        working-directory: .
+
+      - name: Install cross for aarch64
+        run: cargo install cross --git https://github.com/cross-rs/cross --locked
+        working-directory: .
+
+      - name: Build self-hosted Linux runtime (x86_64)
+        run: cargo build --release --target x86_64-unknown-linux-musl -p anyharness -p proliferate-worker -p proliferate-supervisor
+        working-directory: .
+
+      - name: Build self-hosted Linux runtime (aarch64)
+        run: cross build --release --target aarch64-unknown-linux-musl -p anyharness -p proliferate-worker -p proliferate-supervisor
+        working-directory: .
+
+      - name: Prepare release assets
+        env:
+          INPUT_VERSION: ${{ inputs.version || '' }}
+        run: |
+          set -euo pipefail
+          mkdir -p artifacts
+          if [[ "${GITHUB_REF:-}" == refs/tags/server-v* ]]; then
+            VERSION="${GITHUB_REF#refs/tags/server-v}"
+          else
+            VERSION="${INPUT_VERSION:?A server version input is required for reusable releases.}"
+          fi
+          # Archive as root:root so a root extraction on the operator host does
+          # not hand ownership of executables to the runner's unprivileged uid.
+          tar czf artifacts/anyharness-x86_64-unknown-linux-musl.tar.gz \
+            --owner=0 --group=0 --numeric-owner \
+            -C target/x86_64-unknown-linux-musl/release \
+            anyharness \
+            proliferate-worker \
+            proliferate-supervisor
+          tar czf artifacts/anyharness-aarch64-unknown-linux-musl.tar.gz \
+            --owner=0 --group=0 --numeric-owner \
+            -C target/aarch64-unknown-linux-musl/release \
+            anyharness \
+            proliferate-worker \
+            proliferate-supervisor
+          cp server/infra/self-hosted-aws/template.yaml \
+            artifacts/proliferate-self-hosted-aws-template.yaml
+          # Standalone entrypoints operators fetch before they have the bundle:
+          # the guided installer and the AWS launch action. Published (and
+          # checksummed) so they can be pinned to a release and verified.
+          cp server/deploy/install.sh artifacts/proliferate-selfhost-install.sh
+          cp server/infra/self-hosted-aws/launch-stack.sh \
+            artifacts/proliferate-selfhost-aws-launch.sh
+          bundle_root="$(mktemp -d)"
+          mkdir -p "$bundle_root/proliferate-deploy"
+          cp -R server/deploy/. "$bundle_root/proliferate-deploy/"
+          # smoke/ and tests/ are CI-only; the operator bundle ships just the
+          # scripts, compose, and config the running stack needs.
+          rm -rf "$bundle_root/proliferate-deploy/smoke"
+          rm -rf "$bundle_root/proliferate-deploy/tests"
+          rm -f "$bundle_root/proliferate-deploy/.bootstrap-progress.log"
+          printf '%s\n' "$VERSION" > "$bundle_root/proliferate-deploy/VERSION"
+          tar czf artifacts/proliferate-deploy.tar.gz \
+            --owner=0 --group=0 --numeric-owner \
+            -C "$bundle_root" proliferate-deploy
+          rm -rf "$bundle_root"
+          (
+            cd artifacts
+            sha256sum \
+              anyharness-x86_64-unknown-linux-musl.tar.gz \
+              anyharness-aarch64-unknown-linux-musl.tar.gz \
+              proliferate-self-hosted-aws-template.yaml \
+              proliferate-selfhost-install.sh \
+              proliferate-selfhost-aws-launch.sh \
+              proliferate-deploy.tar.gz \
+              > self-hosted-assets.SHA256SUMS
+          )
+        working-directory: .
+
+      - name: Ensure server release tag exists
+        # Only publish calls need to create the tag; on a real server-v tag push
+        # the tag already exists. Input-driven, not event-name-driven.
+        if: inputs.publish == true && inputs.dry_run != true
+        env:
+          RELEASE_TAG: ${{ format('server-v{0}', inputs.version) }}
+          RELEASE_TARGET: ${{ inputs.git_sha }}
+        run: |
+          set -euo pipefail
+          : "${RELEASE_TAG:?Missing server release tag.}"
+          : "${RELEASE_TARGET:?Missing server release target.}"
+          existing_target="$(git ls-remote --tags origin "refs/tags/$RELEASE_TAG" | awk '{ print $1 }' | head -n 1)"
+          if [[ -n "$existing_target" && "$existing_target" != "$RELEASE_TARGET" ]]; then
+            echo "::error::Server tag $RELEASE_TAG already exists at $existing_target, not $RELEASE_TARGET."
+            exit 1
+          fi
+          if [[ -z "$existing_target" ]]; then
+            git tag "$RELEASE_TAG" "$RELEASE_TARGET"
+            git push origin "refs/tags/$RELEASE_TAG"
+          fi
+        working-directory: .
+
+      - name: Publish self-hosted release assets
+        uses: softprops/action-gh-release@v3
+        with:
+          target_commitish: ${{ inputs.git_sha || github.sha }}
+          tag_name: ${{ github.event_name == 'push' && github.ref_name || format('server-v{0}', inputs.version) }}
+          files: |
+            artifacts/anyharness-x86_64-unknown-linux-musl.tar.gz
+            artifacts/anyharness-aarch64-unknown-linux-musl.tar.gz
+            artifacts/proliferate-self-hosted-aws-template.yaml
+            artifacts/proliferate-selfhost-install.sh
+            artifacts/proliferate-selfhost-aws-launch.sh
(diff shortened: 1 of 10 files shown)


pablonyx (commented):
**Reviewer verdict: REQUEST CHANGES**

Fresh reviewer of record, own scratch clone, adversarial pass. The mechanical work here is genuinely good — the job move is byte-identical, the wiring is right, the required-check contract is untouched, and the new test file has real teeth (I reproduced four of the five negative controls myself). One finding blocks: **the central justification for deleting the `server-v*` tag trigger is factually wrong**, and the wrong claim is now baked into a checked-in comment.

---

## HIGH — the `server-v*` tag path was NOT inert; the comment asserting it is false

`.github/workflows/server-ci.yml:9-11` (and the matching passage in the PR body):

> `(It never actually fired either: a `paths:` filter alongside a tag push matches nothing, and this workflow has zero tag-triggered runs in its history despite server-v0.4.4 through server-v0.4.9 all existing.)`

Both halves are wrong.

**1. GitHub does not apply `paths:` filters to tag pushes.** From the workflow-syntax reference, under `on.<push>.<paths>`: *"Path filters are not evaluated for pushes of tags."* Confirmed independently in community discussion [#165354](https://github.com/orgs/community/discussions/165354), where a GitHub moderator states the `paths` filter *"only applies to branch pushes, not to tag pushes… this behavior is by design and not documented clearly."* A `paths:` filter alongside `tags:` does not suppress tag pushes; it is ignored on them.

**2. This workflow has three tag-triggered runs in its history.** The PR's evidence used `--limit 60`, whose window only reaches back to 2026-07-09. Paginating the whole history:

```
gh api --paginate ".../actions/workflows/server-ci.yml/runs?event=push&per_page=100" --jq '.workflow_runs[].head_branch' | sort | uniq -c
    545 main
      1 server-v0.1.1
      1 server-v0.1.2
      1 server-v0.1.3
```

| run | ref | created | actor | conclusion |
| [26621290094](https://github.com/proliferate-ai/proliferate/actions/runs/26621290094) | `server-v0.1.1` | 2026-05-29T06:10Z | pablonyx | cancelled |
| [26621446218](https://github.com/proliferate-ai/proliferate/actions/runs/26621446218) | `server-v0.1.2` | 2026-05-29T06:14Z | pablonyx | failure |
| [26622403422](https://github.com/proliferate-ai/proliferate/actions/runs/26622403422) | `server-v0.1.3` | 2026-05-29T06:40Z | pablonyx | failure |

And the `paths:` filter was already on the trigger at those commits — `git show e2db3d50:.github/workflows/server-ci.yml` has `tags: ["server-v*"]` and the identical seven-entry `paths:` list. So the filter demonstrably did not block them, in this file, in this repo.

**3. The real reason `server-v0.4.x` produced no runs** is not the paths filter — it is that those tags are created *inside* a workflow by `git push origin "refs/tags/$RELEASE_TAG"` using the default `GITHUB_TOKEN` (`_build-server.yml:280`, unchanged from `server-ci.yml`), and pushes made with `GITHUB_TOKEN` do not trigger new workflow runs. The tag path is inert *for pipeline-minted tags*, for a completely different reason, and is live for a human-pushed tag — which is exactly what happened three times.

**Consequence.** Removing `tags: ["server-v*"]` is a real behaviour change, not a no-op cleanup: `git push origin server-vX.Y.Z` by a human used to run Server CI and, through `docker` + `self-hosted-release-assets`, publish the GHCR images and the seven release assets. After this PR it triggers nothing at all.

I am not arguing the escape hatch must be kept — retiring it in favour of "release.yml is the only publisher" is a defensible ruling, and nobody has used it since May. What I am blocking on is that it is being retired *silently, on a false premise, with the false premise written into the file as documentation of record*. In a repo where these comments are the spec, that is the kind of thing the next person will cite.

**Requested change** (comment + PR body, no code redesign):
- Rewrite `server-ci.yml:9-11` to say the tag trigger is **deliberately retired** — the manual `server-v` tag publish path is gone and `release.yml` is the sole publisher — and drop the "paths filter matches nothing" assertion entirely. If you keep the empirical note, it should say pipeline-minted tags never fired because `GITHUB_TOKEN` pushes don't trigger workflows, and that the only tag-triggered runs were three manual pushes in May 2026.
- Same correction in the PR body's "What moved where" bullet.
- State the ruling explicitly: is the manual-tag publish path intentionally gone, or does it want a three-line `on: push: tags: ["server-v*"]` caller of `_build-server.yml`? Either is fine; the PR should decide rather than assume it never existed.

---

## LOW — the brick's tag-gate rationale is now unreachable

`.github/workflows/_build-server.yml:23-25`:

> `The job-level if: gates below are carried over verbatim from server-ci.yml, so a caller triggered by a server-v* tag push keeps the exact publish semantics it had there.`

After this PR no such caller can exist: `release.yml` is `schedule` + `workflow_dispatch` only, and `server-ci.yml` just lost its tag trigger. The `startsWith(github.ref, 'refs/tags/server-v')` leg of both job gates is dead code, and `github.event_name == 'push'` at `_build-server.yml:288` can never be true either (so `tag_name` always resolves through `format('server-v{0}', inputs.version)` — correct, but for a reason the comment doesn't give). Same aspirational wording at `scripts/ci-cd/server-release-build-separation.test.mjs:45`. Harmless behaviourally; please make the comments describe what is actually reachable.

## LOW — the saving claim is lane-scoped, not run-scoped

"~5.5 minutes off every server release" is right for the `release-server` lane. It is not the run's wall clock when desktop is in the surface set: in run 32450223908 `release-desktop / build-desktop` ran to 06:44:43 and `publish-product-release` waited behind it, so the run-level saving on that specific release would have been zero. Worth one clause of precision.

## NIT — no surface-detection assertion

`detect-deploy-surfaces.test.mjs` gains no case pinning `.github/workflows/_build-server.yml -> server`. I verified it manually (`classifyFile` returns `["server"]`, and `server-ci.yml` still returns `["server"]`), and neither path was covered before, so this is not a regression — but the new line is exactly the kind of thing a test would hold in place.

---

## What I verified clean

**Byte-identical move (claim 1) — confirmed.** `diff` of `_build-server.yml:55-296` against `6aa2f298f:.github/workflows/server-ci.yml:542-783` yields exactly the three documented logical changes (drop `needs` on `docker`, `needs: [docker]` on the assets job, one comment rewrite). No drift in `permissions` (`contents: read` + `packages: write` on docker, `contents: write` on assets), `if:` gates, image tags (`:${VERSION}` + `:stable` for both server and LiteLLM), build args, the tag-refusal step, or the seven assets + `SHA256SUMS`. Top-level `defaults.run.working-directory: server` and `env.FORCE_JAVASCRIPT_ACTIONS_TO_NODE24` carried over identically. No `needs.changes.*` reference survives into the brick (would have been a hard error).

**`server-ci-ok` contract (claim 3) — confirmed.** Diff of the whole `server-ci-ok` job old-vs-new is **comment-only**: same `name:`, same `if: always()`, same `needs: [changes, lint, test-unit, test-integration]`, same two steps in the same order with the same polarity, guard ruby untouched. Ran the guard's exact ruby against the new file: exempt set empty, `declared == covered == ["changes","lint","test-integration","test-unit"]`. The PR's own run agrees — job 96692501470 log literally prints `Release-only jobs exempt from the rollup:` (empty) / `server-ci-ok covers all 4 non-release job(s)` / `All 4 Server CI lanes concluded success.` Branch protection on `main` requires `["Analyze (javascript-typescript)","Analyze (python)","Analyze (rust)","Detect smoke-relevant changes","Production compose smoke","Validate PR title and labels","ci-ok","server-ci-ok"]` — neither removed job is a required context, so their disappearance from PR check lists is safe. All 36 checks green on the head.

**`release.yml` wiring (claim 4) — confirmed.** Caller passes exactly `{git_sha, version, dry_run, publish}`; brick declares exactly those four. `comparison_sha` correctly dropped (passing it would now be a hard dispatch error). `secrets: inherit` preserved, `needs: [prepare]` and the `if:` unchanged, caller permissions `contents: write` + `packages: write` cover the brick's `packages: write` (docker) and `contents: write` (assets). Only `.result` is consumed downstream, and the brick declares no `workflow_call` outputs — as `server-ci.yml` didn't either. Brick is reusable-only (`on:` is `workflow_call` and nothing else). All four touched workflows parse under `ruby -ryaml`.

**Surface detection (claim 5) — confirmed.** `_build-server.yml -> ["server"]`, `server-ci.yml -> ["server"]` (unchanged), and the new path correctly does *not* fall into the `_deploy-`/`deploy-`/`promote-`/`scripts/ci-cd/` all-surfaces branch. No other moved file lost coverage.

**Tests (claim 6) — confirmed.** `node --test scripts/ci-cd/*.test.mjs`: **217 pass / 0 fail** on the head, **209** at `6aa2f298f` — the delta is exactly the 8 new tests, nothing pre-existing needed editing. (`node --test scripts/ci-cd/` — directory form — fails with MODULE_NOT_FOUND on Node 22; the glob form in `ci.yml:188` is the one that works, and it does pick up the new file.) Negative controls I ran myself, each on a clean tree and reverted after:

| mutation | result |
| --- | --- |
| `release-server` back to `uses: ./.github/workflows/server-ci.yml` | `not ok 3` ✅ |
| re-add `needs: [lint, test-unit, test-integration]` to the brick's `docker` | `not ok 1` ✅ |
| drop `version:` from the caller's `with:` | `not ok 4` ✅ |
| drop `test-unit` from the `server-ci-ok` rollup | `not ok 7` ✅ (and the drift-guard ruby independently reports `missing=["test-unit"]`) |

Baseline `# pass 8 # fail 0` before and after each. The assertions do pin the stated invariants rather than restating them.

**Docs (claim 7) — confirmed.** Delivery README gains the `_build-server.yml` row, correctly re-describes `server-ci.yml`'s posture as "Relevant push/PR, manual, or reusable" (accurate post-change) and its role as gate-not-publisher, and rewrites the required-checks exclusion paragraph. `support-loop.md:126` retarget is correct. `ci.yml:773-779` comment block updated in the same commit as required. No stale `self-hosted-release-assets` / `server-ci publishes` reference survives anywhere in the tree.

**Other body claims.** Timings in the redundancy table match run 32450223908 job-for-job (`docker` started 05:27:16, `prepare` finished 05:21:39). `release.yml` was the only checked-in caller of `server-ci.yml`'s `workflow_call` at the parent commit. `deploy-staging.yml:88` is a `gh run list --workflow server-ci.yml --commit` poll, not a build call, and `server-ci.yml` still runs on every relevant `main` push — behaviour unchanged, no follow-up owed, as stated.

**Unverifiable from here.** That `docker` actually starts within ~15s of `prepare` on the next real release, and that a real publish still produces identical images/assets — only the next release proves either. Given the move is byte-identical and the input wiring is pinned by test 4, the residual risk is low, but nothing short of a real run closes it.

pablonyx:
Addressed the REQUEST CHANGES review (thanks for the paginated `gh api` evidence — that's decisive).

**HIGH — `server-ci.yml:9-11` tag-trigger comment rewritten.** Confirmed both halves of the old comment were false: GitHub does not evaluate `paths:` filters on tag pushes at all, and this workflow has three tag-triggered runs in its history (`server-v0.1.1`/`0.1.2`/`0.1.3`, 2026-05-29, runs 26621290094/26621446218/26622403422), all human-pushed. The new comment states the true mechanism: the tag trigger *did* fire for human-pushed tags; `server-v0.4.x` tags never fired because they're minted by a workflow using the default `GITHUB_TOKEN`, and `GITHUB_TOKEN` pushes don't trigger workflows; and the trigger is being deliberately retired now that releases flow through `release.yml` -> `_build-server.yml`. It says explicitly: a human-pushed `server-v*` tag no longer runs anything after this PR.

**LOW — `_build-server.yml:23-25` and `server-release-build-separation.test.mjs:45` comments rewritten.** Both now say the `server-v*` tag-push leg of the publish gate is currently unreachable (no caller triggers this workflow from a tag push today), retained unchanged only for byte-comparability with the moved code and to support a possible future tag-triggered caller. No workflow logic, gates, or test assertions changed — comment text only, diff confirms this.

**PR body updated:**
- "What moved where" bullet corrected with the true mechanism and evidence.
- New section "Open ruling for Pablo: manual tag publish path" laying out the two options: (a) accept the retirement (releases only via `release.yml`), or (b) follow-up PR adding a thin `on: push: tags: ["server-v*"]` caller of `_build-server.yml`.
- "~5.5 minutes off every server release" rescoped to "~5.5 minutes off the `release-server` lane," with an explicit note that run 32450223908's actual critical path was `release-desktop`, so the run-level saving there would have been zero.

**Tests:** `node --test scripts/ci-cd/server-release-build-separation.test.mjs` → `# pass 8 # fail 0` (all 8 unchanged, since only comments moved). Full suite `node --test scripts/ci-cd/*.test.mjs` → `# pass 217 # fail 0`, matching the pre-existing baseline in the PR description.

Pushed as b6375d4 on `perf/release-server-build-brick`.

pablonyx (commented):
**Reviewer verdict: APPROVE** (re-review of `b6375d4`; blocking finding resolved)

Reviewer of record, second pass. `b6375d4` addresses my REQUEST CHANGES. All four findings are resolved and the diff is comment-only. Verified independently, not taken on the PR's word.

## 1. HIGH (tag-trigger history) — RESOLVED

`.github/workflows/server-ci.yml:4-16` now states, and I re-confirmed each clause against the API and against `git show`:

- **"the tag trigger DID fire historically: three human-pushed tags (server-v0.1.1/0.1.2/0.1.3, 2026-05-29, runs 26621290094/26621446218/26622403422)"** — correct, and the tag→run pairing is in the right order (0.1.1→26621290094 cancelled, 0.1.2→26621446218 failure, 0.1.3→26622403422 failure), all `actor: pablonyx`. Full-history pagination still returns exactly `545 main / 1 server-v0.1.1 / 1 server-v0.1.2 / 1 server-v0.1.3`.
- **"GitHub does not evaluate `paths:` filters for tag pushes at all, so the `paths:` list below never suppressed those runs"** — matches the workflow-syntax reference (*"Path filters are not evaluated for pushes of tags"*) and the moderator answer in community discussion #165354, and matches the local evidence that the identical seven-entry `paths:` list was already on the trigger at `e2db3d50`.
- **"server-v0.4.x tags produced no runs for an unrelated reason: those tags are minted by a workflow using the default `GITHUB_TOKEN` … and `GITHUB_TOKEN` pushes do not trigger new workflow runs"** — correct; the step is `_build-server.yml`'s `Ensure server release tag exists`, `git push origin "refs/tags/$RELEASE_TAG"` on the checkout's persisted credentials.
- **"a human-pushed `server-v*` tag no longer runs anything in this repo"** — I checked this one rather than assuming it. The only surviving tag triggers in `.github/workflows/` are `release-desktop.yml` (`desktop-v*`) and `release-runtime.yml` (`runtime-v*`); neither pattern matches `server-v*`. Accurate.

The framing is now "deliberate retirement," which is the correct characterisation of what the change does.

## 2. LOW (unreachable tag-gate leg) — RESOLVED

`_build-server.yml:23-29` and `server-release-build-separation.test.mjs:44-48` now say the `startsWith(github.ref, 'refs/tags/server-v')` leg (and the `github.event_name == 'push'` check at `:292`) is currently unreachable, retained unchanged for byte-comparability with the code it moved from and for a possible future tag-triggered caller. That is exactly the reachable-truth I asked for, and it also explains *why* the dead leg is deliberately not being deleted — which is the right call, since deleting it would break the byte-identical-move property this PR rests on.

## 3. Open ruling + saving caveat — PRESENT

PR body carries **"Open ruling for Pablo: manual tag publish path"** with both options stated ((a) accept the retirement; (b) follow-up thin `on: push: tags: ["server-v*"]` caller of `_build-server.yml`) and explicitly declines to choose. The corrected inertness bullet is in "What moved where," and the saving line now reads "~5.5 minutes off the `release-server` lane … lane-scoped, not necessarily run-critical-path," with the run-32450223908 desktop counterexample spelled out. Both of my LOW findings are answered in the body, not just in the code.

## 4. No logic drift — CONFIRMED

`git diff bdb1f442 b6375d42` touches three files (+24/−10) and, filtering out `#` and `//` comment lines, contains **zero** non-comment changes. Structural re-parse of both edited workflows after the change:

```
server-ci  jobs=["changes","lint","test-unit","test-integration","server-ci-ok"]
server-ci  push={"branches"=>["main"], "paths"=>[7 entries, unchanged]}
brick      on=["workflow_call"]  jobs=["docker","self-hosted-release-assets"]
brick      docker.if = (startsWith(github.ref, 'refs/tags/server-v')) || (inputs.publish == true && inputs.dry_run != true)
```

No gate, trigger, job set, `needs:`, permission, or input moved. In the test file only the comment above `PUBLISH_GATE` changed; the `PUBLISH_GATE` string literal and every assertion are byte-identical to the state I reviewed.

`node --test scripts/ci-cd/server-release-build-separation.test.mjs` on `b6375d42`, literal output:

```
ok 1 - the release build brick carries no lint or test lane
ok 2 - the release build brick is reusable-only, like the other lane bricks
ok 3 - release.yml builds the server through the brick, never through Server CI
ok 4 - every input release.yml passes is one the brick declares, and vice versa
ok 5 - the brick preserves the publish gate, image tags, and release tag semantics
ok 6 - Server CI is now a gate only and publishes nothing
ok 7 - server-ci-ok still rolls up every non-release lane in its own file
ok 8 - production can still only ship a commit that reached main
# tests 8
# pass 8
# fail 0
```

Everything I verified on the first pass — byte-identical job move, `server-ci-ok` needs/polarity untouched, drift-guard derivation empty-and-complete, `release.yml` input parity and `secrets: inherit`, surface classification, docs — is unaffected by a comment-only commit and still holds.

## Remaining, non-blocking

- CI is mid-flight on `b6375d42` (`CI/CD config checks` already `pass`; `Server CI` run 32457605162 `in_progress`, `ci-ok` pending). Comment-only change against a fully green `bdb1f442`, so I expect a clean reproduction, but the green on this head is not yet in hand — worth an eyeball before merge.
- Still unverifiable from any static pass: that `docker` actually starts within ~15s of `prepare` on the next real release, and that a real publish emits identical images/assets. The byte-identical move plus test 4's input-parity pin keeps that risk low.
- The **Open ruling for Pablo** above is a live decision, not a defect. My review does not resolve it; option (a) is a fine default given nobody has used the manual tag path since May 2026.

No merge action taken — verdict comment only, per the repo's no-merge-without-Pablo rule.

PR #2170, opened 2026-08-21
fix(release): publish desktop updater from flat artifact layouts

pablonyx:
## Problem

The v0.4.22 desktop publish failed in `release-desktop / publish-updater` (run 32450223908, job 96692480347):

```
Manifest generation failed:
  - Missing signature file for darwin-aarch64
  - Missing artifact file for darwin-aarch64
```

…while the download step in the same job listed the files as present:

```
all-artifacts/Proliferate_0.4.22_aarch64.app.tar.gz
all-artifacts/Proliferate_0.4.22_aarch64.app.tar.gz.sig
all-artifacts/Proliferate_0.4.22_aarch64.dmg
```

Result: the 0.4.22 app is built/signed/notarized but downloads and auto-update still serve 0.4.21, and `publish-product-release` was skipped. Nothing was half-published — the failure happened before any S3 upload, and the immutable-manifest guard for `desktop/stable/0.4.22/latest.json` was never tripped.

## Root cause

`actions/download-artifact@v8` decides the extraction path per artifact ([src/download-artifact.ts @ 3e5f45b2](https://github.com/actions/download-artifact/blob/3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c/src/download-artifact.ts)):

```js
path: isSingleArtifactDownload || inputs.mergeMultiple || artifacts.length === 1
  ? resolvedPath                              // flat
  : path.join(resolvedPath, artifact.name)    // per-artifact directory
```

With `merge-multiple: false` and a pattern, you get per-artifact directories **only when 2+ artifacts match**. Every prior release had two desktop artifacts (aarch64 + x86_64), e.g. the 08-18 train (job 95839976547) downloaded into `all-artifacts/desktop-aarch64-apple-darwin/…` + `all-artifacts/desktop-x86_64-apple-darwin/…`. Tonight was the first release since the Intel lane was paused: exactly one artifact matched `desktop-*`, the layout flattened, and both manifest generators require a `desktop-<target>` path segment via `pathIncludes(...)` — so they found nothing.

## Fix

- **Deterministic layout**: `merge-multiple: true` on both `desktop-*` downloads (create-release + publish-updater). Merged-flat is the one layout every artifact count produces. create-release's `all-artifacts/**/*` glob already handled flat (it succeeded tonight).
- **Filename-only matching**: `generate-updater-manifest.mjs` and `generate-desktop-installer-manifest.mjs` drop the `pathIncludes` directory requirement; the aarch64/x64 filename regexes already disambiguate platforms and work under both layouts.

## Tests

- `generate-updater-manifest.test.mjs`: fixtures gain a `flat` layout; new cases for the flat single-artifact repro (tonight's failure) and the merged-flat two-platform layout.
- New `generate-desktop-installer-manifest.test.mjs` (the installer generator had no tests): nested, flat-arm-only, merged-flat, and missing-required-installer failure.
- `node --test` on both files: **12 pass / 0 fail**.
- **Negative control**: with the two generator scripts reverted to main, exactly the 4 new flat-layout tests fail (`not ok`), all 8 pre-existing tests still pass.

## After merge

Re-dispatch `release-desktop.yml` (publish path) to put 0.4.22 downloads + updater manifest + product release page live. The S3 guard permits it: `desktop/stable/0.4.22/latest.json` was never created.

### DIFF
--- a/.github/workflows/release-desktop.yml
+++ b/.github/workflows/release-desktop.yml
@@ -703,7 +703,13 @@ jobs:
         with:
           path: all-artifacts
           pattern: desktop-*
-          merge-multiple: false
+          # merge-multiple keeps the extracted layout deterministic: with
+          # `false`, download-artifact@v8 nests each artifact in its own
+          # directory ONLY when two or more artifacts match — a single match
+          # (ARM-only, since Intel is paused) extracts flat and broke the
+          # 0.4.22 publish. Merged-flat is the one layout every artifact
+          # count produces.
+          merge-multiple: true
 
       - name: List artifacts
         run: find all-artifacts -type f | sort
@@ -743,7 +749,13 @@ jobs:
         with:
           path: all-artifacts
           pattern: desktop-*
-          merge-multiple: false
+          # merge-multiple keeps the extracted layout deterministic: with
+          # `false`, download-artifact@v8 nests each artifact in its own
+          # directory ONLY when two or more artifacts match — a single match
+          # (ARM-only, since Intel is paused) extracts flat and broke the
+          # 0.4.22 publish. Merged-flat is the one layout every artifact
+          # count produces.
+          merge-multiple: true
 
       - name: List artifacts
         run: find all-artifacts -type f | sort
--- a/scripts/ci-cd/generate-desktop-installer-manifest.test.mjs
+++ b/scripts/ci-cd/generate-desktop-installer-manifest.test.mjs
@@ -1,0 +1,104 @@
+import assert from "node:assert/strict";
+import { spawnSync } from "node:child_process";
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import test from "node:test";
+import { fileURLToPath } from "node:url";
+
+const repoRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "../..");
+const generator = path.join(repoRoot, "scripts/generate-desktop-installer-manifest.mjs");
+
+// download-artifact@v8 nests each artifact in a directory named after it only
+// when two or more artifacts match the download pattern; a single match
+// extracts flat, and merge-multiple: true (the release-desktop.yml publish
+// jobs) merges flat regardless of count. `flat` reproduces those layouts.
+// Intel (x86_64) desktop builds are paused 2026-08-20 (release-desktop.yml
+// build-desktop matrix); `armOnly` simulates that paused lane producing no
+// x86_64 installer.
+function writeInstallers(root, { armOnly = false, flat = false } = {}) {
+  const artifacts = path.join(root, "artifacts");
+  const fixtures = [
+    ["desktop-aarch64-apple-darwin", "Proliferate_aarch64.dmg"],
+    ["desktop-x86_64-apple-darwin", "Proliferate_x64.dmg"],
+  ].filter(([directory]) => !armOnly || directory === "desktop-aarch64-apple-darwin");
+
+  for (const [directory, installer] of fixtures) {
+    const target = flat ? artifacts : path.join(artifacts, directory);
+    fs.mkdirSync(target, { recursive: true });
+    fs.writeFileSync(path.join(target, installer), "dmg");
+  }
+
+  return artifacts;
+}
+
+function generateManifest(t, { armOnly = false, flat = false, empty = false } = {}) {
+  const root = fs.mkdtempSync(path.join(os.tmpdir(), "proliferate-installer-manifest-"));
+  t.after(() => fs.rmSync(root, { recursive: true, force: true }));
+  const output = path.join(root, "installers.json");
+  const artifactsDir = empty
+    ? (() => {
+        const dir = path.join(root, "artifacts");
+        fs.mkdirSync(dir, { recursive: true });
+        return dir;
+      })()
+    : writeInstallers(root, { armOnly, flat });
+
+  const result = spawnSync(
+    process.execPath,
+    [
+      generator,
+      "--version",
+      "1.2.3",
+      "--artifacts-dir",
+      artifactsDir,
+      "--base-url",
+      "https://downloads.proliferate.com/desktop/stable",
+      "--output",
+      output,
+    ],
+    { encoding: "utf8" },
+  );
+  return { output, result };
+}
+
+test("generates both downloads from the nested per-artifact layout", (t) => {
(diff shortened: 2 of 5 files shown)


pablonyx (commented):
Fresh review of #2170 (did not author this change). Independently verified every factual claim in the PR body rather than trusting it — details below. No blocking findings; comment-only.

## Claims verified independently

**1. `download-artifact@v8` flat-extraction clause** — fetched `src/download-artifact.ts` at the pinned SHA `3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c` directly from GitHub:

```ts
// lines 185-194
const downloadPromises = artifacts.map(artifact => ({
  name: artifact.name,
  promise: artifactClient.downloadArtifact(artifact.id, {
    ...options,
    path:
      isSingleArtifactDownload ||
      inputs.mergeMultiple ||
      artifacts.length === 1
        ? resolvedPath
        : path.join(resolvedPath, artifact.name),
    ...
```
Confirms the claim exactly: with `merge-multiple: false` and a `pattern`, a single matching artifact extracts flat.

**2. Run 32450223908 / job 96692480347 failure** — pulled the job logs via `gh api`. "List artifacts" step shows the flat listing:
```
all-artifacts/Proliferate_0.4.22_aarch64.app.tar.gz
all-artifacts/Proliferate_0.4.22_aarch64.app.tar.gz.sig
all-artifacts/Proliferate_0.4.22_aarch64.dmg
```
"Generate updater manifest" step failed with:
```
Manifest generation failed:
  - Missing signature file for darwin-aarch64
  - Missing artifact file for darwin-aarch64
```
and every step after it (S3 upload, immutable-manifest guard, CloudFront invalidation, Homebrew bump) is `skipped` — confirms "nothing was half-published."

**3. 08-18 green run (job 95839976547) had per-artifact directories** — confirmed via job log:
```
all-artifacts/desktop-aarch64-apple-darwin/Proliferate_0.4.17_aarch64.app.tar.gz
all-artifacts/desktop-x86_64-apple-darwin/Proliferate_0.4.17_x64.app.tar.gz
```
Also confirmed via `git log -S "Intel builds paused"` that the Intel matrix entry was commented out on 2026-08-20 18:03 (commit e2fd47e), and the 0.4.22 failure at 06:46 the next day was the first `release-desktop` run since — matching the "first release with exactly one matching artifact" narrative.

## Review dimensions

**1. Correctness across the four layouts** — traced all matrix/upload logic in `release-desktop.yml`. The `runner_os == 'windows'` branches (build/bundle) exist but are dormant: the matrix (`build-desktop.strategy.matrix.include`) currently has only `aarch64-apple-darwin`, and the *only* two `upload-artifact` steps producing `desktop-*`/`proliferate-dmg-*` names are both gated `if: ... && matrix.runner_os == 'macos'`. So today, no Windows file can ever land under the `desktop-*` pattern this code downloads — the filename-only regexes (`\.dmg$`, `\.app\.tar\.gz$`) also wouldn't match Windows' own `nsis`/`.exe` bundle format even if that upload step existed. Concern doesn't apply currently; worth a one-line note in the workflow's Intel-pause comment for whoever re-adds a Windows upload step later, but not blocking.

**2. create-release job / softprops glob** — `files: all-artifacts/**/*` already matches flat files (that's why create-release succeeded during the 0.4.22 incident even before this fix). Confirmed `proliferate-dmg-${{ matrix.target }}` does not match the `desktop-*` download pattern (different prefix), so it's untouched by either download step. The extension-based `find` steps used for S3 upload (`generate-updater-manifest.mjs` sibling step "Upload desktop assets to S3") and the Homebrew cask bump were already directory-agnostic (`find all-artifacts -type f -name "..."`), so they're unaffected by the layout change either way.

**3. Determinism / `sigFiles[0]`** — `findFiles` walks via plain `readdirSync`, so ordering isn't guaranteed sorted. Neither generator asserts exactly one match before indexing `[0]` (pre-existing behavior, not introduced by this diff — the old `pathIncludes` guard added directory scoping but never an arity check either). Not currently exploitable: per-target normalization steps upstream (`Normalize updater/installer artifact names (macOS)`) guarantee each artifact contains exactly one file matching each regex before it's ever uploaded, and the two live filename patterns (aarch64 vs x64/x86_64) are mutually exclusive. Low-severity nit, not blocking: an explicit `if (files.length > 1) error` would make the arity assumption self-enforcing instead of implicit.

**4. Tests** — ran `node --test` on both PR-head files:
```
scripts/ci-cd/generate-updater-manifest.test.mjs + generate-desktop-installer-manifest.test.mjs
# tests 12
# pass 12
# fail 0
```
Negative control: reverted both `scripts/generate-updater-manifest.mjs` and `scripts/generate-desktop-installer-manifest.mjs` to `origin/main` in my scratch clone and reran the same two test files:
```
not ok 2 - generates an arm-only manifest from the flat single-artifact layout   (installer)
not ok 3 - generates both downloads from a merged flat layout                    (installer)
not ok 7 - generates an arm-only manifest from the flat single-artifact layout   (updater)
not ok 8 - generates both platforms from a merged flat layout                    (updater)
# tests 12
# pass 8
# fail 4
```
Exactly the 4 new flat-layout tests fail, all 8 pre-existing tests still pass — matches the PR body's negative-control claim precisely. Restored the PR versions afterward (working tree clean).

**5. Comments/docs truthfulness** — the new comments in `release-desktop.yml`, `generate-updater-manifest.mjs`, and `generate-desktop-installer-manifest.mjs` describing the `merge-multiple`/`artifacts.length === 1` behavior match the verified action source exactly. No overstated or fabricated claims found.

**6. Blast radius** — grepped the whole repo for `pathIncludes`, `desktop-aarch64-apple-darwin`, `desktop-x86_64-apple-darwin`, and `all-artifacts`: only `release-desktop.yml` and the two generator scripts (plus their own tests) reference this layout. `publish-product-release.mjs` matches git tags (`desktop-v*`), an unrelated namespace — no coupling. No other workflow downloads the `desktop-*` artifact pattern. Both `release-desktop.yml` download steps (create-release and publish-updater) were updated consistently — no stale `merge-multiple: false` left behind.

## Verdict

No blocking issues. Fix is narrowly scoped, the root cause matches the verified logs and action source byte-for-byte, and the test claims (12/12 pass, 4/12 negative-control fail) reproduce exactly. The two nits above (no explicit arity assertion in the manifest generators; dormant-Windows-lane note) are non-blocking and can be follow-ups.

PR #2178, opened 2026-08-21
fix(observability): bring up Grafana alerting on rebuild workspace, retire ECS CPU paging

pablonyx:
## Summary

- `proliferate-ops-rebuild` (g-48655e6419) had zero working alerting tonight: `unifiedAlerting.enabled` was `false` at the workspace level (blocks the whole provisioning API regardless of data source state), it had no CloudWatch data source, and AMG returns `no secrets configured for type 'email'` for a native email contact point.
- Live, internal-provider-config-only actions taken and verified by read-back (never merged, no customer-facing change): enabled unified alerting; added the CloudWatch data source (workspace IAM role auth, health-checked OK); re-created the five current production alert rules with their original UIDs preserved, pointed at the new data source; repointed the workspace's default SNS contact point at the already-provisioned `grafana-proliferate-ops-alerts` topic (which already has a CONFIRMED `pablo@pablohansen.com` email subscription and IAM publish permission granted to both workspaces' roles) per the overnight execution spec's D2 revision; added a Production Overview dashboard. Fired one live test notification through the real path: `status: ok`.
- Code changes make this reproducible/reviewable and formally retire `ECS CPU > 90% for 15m` (`cfrmh7d7od8g0c`) as a page (infra symptom, not a product promise) from the shared five-rule allowlist used by both the OLD-workspace overlay tool (`grafana-alerting.mjs`) and the new `grafana-rebuild-bootstrap.mjs` (new workspace only, cannot write to the OLD workspace, and vice versa).
- `scripts/ops/*.test.mjs` was not wired into any CI job before this change (checked: no workflow ran it). Added a step so this new code, and the existing 1000+ line `grafana-alerting.test.mjs`, actually run in CI going forward.
- OLD workspace (`proliferate-ops`, g-e532d030d8) is untouched and remains the rollback.

## Test plan

- [x] `node --test scripts/ops/*.test.mjs` -- 222/222 pass locally
- [x] `node scripts/ops/grafana-alerting.mjs check` -- passes
- [x] `node scripts/ops/grafana-rebuild-bootstrap.mjs check` -- passes
- [x] Live `verify` against the real new workspace: all 5 rule checksums match, route receiver correct, SNS topic wired, dashboard present
- [x] Live `apply` re-run against the now-populated workspace is a no-op everywhere (idempotency proof)
- [ ] CI green on this PR (pushed, not merged -- merge is Pablo's per standing policy)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -187,6 +187,14 @@ jobs:
       - name: Test CI/CD scripts
         run: node --test scripts/ci-cd/*.test.mjs
 
+      # Offline-only: exercises canonicalization, allowlist/checksum drift
+      # detection, and the fake-client apply/verify contract for the Grafana
+      # alerting tooling (both the OLD-workspace overlay tool and the NEW
+      # (proliferate-ops-rebuild) workspace bootstrap). Never touches the
+      # network; live operations stay gated behind GRAFANA_ALERTING_LIVE=1.
+      - name: Test Grafana ops scripts
+        run: node --test scripts/ops/*.test.mjs
+
       # Launcher-level proof (BG4-LOCAL-01): `make run BACKGROUND=1` must start
       # the Celery worker/Beat against the SELECTED profile database on
       # profile-scoped broker/store ports. Renders `docker compose config` (no
--- a/guides/operating/production-alerts.md
+++ b/guides/operating/production-alerts.md
@@ -1,7 +1,17 @@
 # Production Grafana alerts
 
-Status: authoritative for the six production Grafana alert rules, their stable
-identity, and the dedicated dark issue-tracker webhook contact point.
+Status: authoritative for the five production Grafana alert rules, their
+stable identity, and the dedicated dark issue-tracker webhook contact point,
+on the OLD workspace (`proliferate-ops`, g-e532d030d8). See
+[below](#the-new-workspace-proliferate-ops-rebuild) for the NEW workspace
+(`proliferate-ops-rebuild`, g-48655e6419), which uses a different (simpler,
+tracker-free) delivery design per the 2026-08-20 overnight execution spec's
+D2 revision.
+
+`ECS CPU > 90% for 15m` (`cfrmh7d7od8g0c`) was retired as a paging rule on
+2026-08-21: infra symptom, not a product promise. It is no longer in the
+allowlist on either workspace, but it is still queryable on the Production
+Overview dashboard in the new workspace.
 
 Use this runbook to understand what each production alert detects, where to
 look first when it fires, and how to reproduce or roll back the rule-identity
@@ -37,7 +47,7 @@ docs. Share only rule UIDs, metadata names, checksums, and contact-point setting
 names. The operator script redacts URLs, authorization values, credentials, and
 bodies from all console output.
 
-## The six rules
+## The five rules
 
 Every rule carries the labels `proliferate_rule_uid` (its immutable UID),
 `proliferate_component=proliferate-server`, and `severity`, plus a stable
@@ -60,14 +70,6 @@ Every rule carries the labels `proliferate_rule_uid` (its immutable UID),
   saturation, and recent deploys.
 - Metric alert; no log lookup metadata.
 
-### ECS CPU saturation (cfrmh7d7od8g0c)
-
-- Detects: ECS service CPU above 90% for 15 minutes.
-- Severity: critical. Component: proliferate-server.
-- Look first at task count, autoscaling activity, and any runaway request
-  pattern or background loop.
-- Metric alert; no log lookup metadata.
-
 ### CRITICAL_FAILURE in prod logs (bfrmh7e7x2k8wd)
 
 - Detects: the `CRITICAL_FAILURE` marker emitted by `report_critical(...)` in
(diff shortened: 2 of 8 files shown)


pablonyx (commented):
test probe - will follow with real content

pablonyx (commented):
VERDICT: merge-safe

(Note: correcting my earlier "test probe" comment on this thread -- this is the real review. Also: gh/GitHub blocks a formal --approve/--request-changes review because the authenticated CLI account is also the PR author, so this verdict is posted as a review comment instead. Pablo: safe to merge on this verdict, with two non-blocking follow-ups flagged below.)

Reviewed via `gh pr diff`, a detached read-only checkout at the PR head SHA (dd0705139), and actually ran the tooling with `node --test` and direct CLI invocation (not just read the source).

## Confirmed safe

- **Live network gate, both tools, empirically verified (not just read):** ran `node scripts/ops/grafana-rebuild-bootstrap.mjs apply|verify` and `node scripts/ops/grafana-alerting.mjs export|apply|restore` with `GRAFANA_ALERTING_LIVE` unset and with `GRAFANA_ALERTING_LIVE=true` (wrong value) -- all five refuse immediately with a clear error before any network attempt. Neither test suite unit-tests `assertLiveAllowed` directly (scripts/ops/grafana-rebuild-bootstrap.mjs:186, scripts/ops/grafana-alerting.mjs:335 are both unexported and only reached through `main()`), so this had to be checked by execution, not by reading the test file.
- **OLD tool not weakened, workspaces structurally isolated:** scripts/ops/grafana-client.mjs (owns the OLD-workspace `TARGET` and admin-token/contact-point plumbing) is untouched by this PR -- not in the changed-file list at all. The new tool defines its own `NEW_TARGET` (grafana-rebuild-bootstrap.mjs:51-59) and its own admin-token path (line 66-70), and only imports `KNOWN_RULES/KNOWN_UIDS/queryChecksum/redact`/etc. from grafana-alerting.mjs -- never `TARGET`, `CONTACT_POINT_NAME`, or `adminTokenProvider`. Test confirms this: "this tool's fixed target is the new workspace, never the OLD workspace" and "admin token path is distinct from the OLD workspace's admin token path" (grafana-rebuild-bootstrap.test.mjs:324-333), both passing.
- **Rule JSON retires ECS-CPU, keeps original UIDs:** parsed server/infra/observability/grafana/production-alerts-rebuild.json -- its 5 rules are exactly `[dfrmh7bc4yqrkf, bfrmh7c7ecbnkb, bfrmh7e7x2k8wd, cfrmh7f2sbe2od, cfrmh7fttw4jke]`, i.e. the 5 surviving OLD-workspace UIDs verbatim, with `cfrmh7d7od8g0c` (ECS CPU) absent from the rules array entirely (only mentioned in prose). scripts/ops/grafana-alerting.mjs:39-46 (`KNOWN_RULES`) was edited the same way for the OLD tool.
- **Dark issue-tracker webhook stays dark:** server/infra/observability/grafana/issue-tracker-contact.json is not in this PR's changed-file list at all -- zero touch. grafana-alerting.mjs's contact-point/notification-policy code (`runApply`/`runRestore`/`assertContactTemplateSafe`) is untouched; the only diff there is the KNOWN_RULES count and message strings. The NEW tool's contact point is a completely separate, non-tracker SNS receiver (`contactPoint.type` must be `"sns"`, enforced at grafana-rebuild-bootstrap.mjs ~161-163, tested at line 85 "rejects a contact point that is not sns"), and `ensureContactRouting` (line ~328) refuses to create a new receiver or touch the route tree, only repoints one existing field -- tested at line 220 "refuses to touch the route tree or create a new receiver."
- **Tests pass, actually run them:** `node --test scripts/ops/grafana-alerting.test.mjs` -> 44/44 pass. `node --test scripts/ops/grafana-rebuild-bootstrap.test.mjs` -> 21/21 pass. CI wiring (.github/workflows/ci.yml +8 lines) runs `node --test scripts/ops/*.test.mjs`, offline only, matches what I ran locally.
- **No secrets:** grepped all new/changed files for token/ARN/bearer patterns; the only hit is a deliberately fake fixture string (`glsa_live_secret_value_1234567890`) used to test the `redact()` function in grafana-alerting.test.mjs, and `issue-tracker-contact.json`'s description explicitly documents it "never contains the value." Admin tokens are read from 0600-only files at execution time and never logged (`safeLog` routes everything through `redact()`).

## Non-blocking follow-ups (please address, don't need to block merge)

1. **Stale "six" references now inconsistent with the code's "five":** guides/operating/production-alerts.md:200 ("sends the six rules' notifications"), :218 ("the six rules keep firing"), and :231 ("the six exact UIDs are present") plus server/infra/observability/grafana/production-alerts.json:4 (description says "the six existing production Grafana alert rules" while the file's `rules` array now has 5 entries). Trivial word-level fix, but the "Live acceptance" checklist at :231 is the kind of text an on-call engineer follows during a real incident -- it should say five.
2. **Operational gotcha worth flagging explicitly to whoever runs grafana-alerting.mjs next against the real OLD workspace:** this PR narrows `KNOWN_RULES`/`KNOWN_UIDS` in the OLD-workspace tool from 6 to 5 (grafana-alerting.mjs:39-46), but per the PR body the OLD workspace itself was left untouched live. If its live rule set still literally has 6 rules (ECS CPU included, which nothing in this PR deletes), the next `export`/`apply`/`restore` run will immediately throw `Unknown rule UID (wildcard discovery is not allowed): cfrmh7d7od8g0c` via `verifyUidAllowlist`, since it requires an exact match. This fails safe (loud error, no data touched) rather than fails dangerous, so it's not a merge blocker, but it means the OLD tool is not actually usable live again until someone either restores the rule to `KNOWN_RULES` or reconciles the live OLD workspace to 5 rules. Worth a one-line note in the PR or the guide so this isn't a surprise mid-incident.

Most load-bearing check: actually executing all five live-command entry points across both tools with the gate unset/wrong, confirming refusal happens before any network call -- since neither test suite exercises `assertLiveAllowed` as a unit test, this is the one property that could not be verified by reading code alone.

PR #2097, opened 2026-08-19
fix(mobile): source-disable PostHog session replay (CP-C1PM)

pablonyx:
## Summary

Final slice of the PostHog replay cutover. Mobile PostHog session replay becomes impossible to turn on through a build, an Expo/EAS environment, a GitHub environment, a provider setting, or a dormant branch. With this PR, **every client surface — Desktop, Web, and Mobile — has source-disabled PostHog session recording**, so the cross-surface docs advance to all-client truth for the first time.

Spec: `CP-C1PM Mobile PostHog replay fail-closed` (Vault, rev 2026-08-19.3, reconciled to the landed CP-C1PW squash `f55b8721`, which is also the base of this branch).

## Scope — exactly eleven files

**Source (5)**
- `apps/mobile/src/config/env.ts` — delete the duplicate `EXPO_PUBLIC_PROLIFERATE_POSTHOG_SESSION_REPLAY_ENABLED` declaration (it was never read, but it falsely advertised a supported app input).
- `apps/mobile/src/lib/integrations/telemetry/config.ts` — delete the public build-variable declaration, `MobileTelemetryConfig.posthog.sessionReplayEnabled`, and its `envFlagEnabled(...)` read. No alias, constant, remote flag, or option bag replaces it. `envFlagEnabled` stays because the supported `EXPO_PUBLIC_PROLIFERATE_TELEMETRY_DISABLED` gate still owns it.
- `apps/mobile/src/lib/integrations/telemetry/posthog.ts` — one literal `enableSessionReplay: false` at the SDK boundary as a source-owned fail-closed assertion; `sessionReplayConfig` and its whole mask/log/network/throttle branch deleted. No `startSessionRecording`, `startSessionReplay`, `screen`, `PostHogProvider`, navigation tracker, `captureScreens`, or touch-autocapture surface is introduced.
- `apps/mobile/src/lib/integrations/telemetry/config.test.ts` *(new)* — 19-row legacy-value rejection matrix.
- `apps/mobile/src/lib/integrations/telemetry/telemetry-privacy.test.ts` — new constructor-boundary assertions.

**Docs (6)** — `specs/OBSERVABILITY.md`, `specs/frontend/telemetry.md`, `specs/codebase/systems/engineering/analytics/README.md`, `specs/codebase/systems/engineering/analytics/posthog.md`, `guides/operating/analytics/posthog.md`, `specs/developing/reference/env-vars.yaml`.

[...]

PR #2096, opened 2026-08-19
fix(web): source-disable PostHog session recording (CP-C1PW)

pablonyx:
## Summary

Makes Web PostHog **session recording impossible to enable through a build or environment value**. The Web host now passes `disable_session_recording: true` unconditionally, carries no recorder options object, no `loaded` callback, and no recorder start call. Re-enablement requires a separate founder-approved source PR with synthetic privacy qualification; changing a deployment variable is never sufficient.

Slice `CP-C1PW` of the Control Plane observability program. Semantic predecessor is `CP-C1PD` (Desktop, merged as `7a9bb9af273656fa2ff70eae43d14cb5d626f2b1`, PR #2093); this branch is based directly on that squash and preserves its landed Desktop truth.

## Privacy rationale

The pinned `posthog-js` 1.386.8 default config has `disable_session_recording: false`, and its recorder enables when server-side recording is on **and** `!config.disable_session_recording`. So simply deleting the environment read would inherit the unsafe SDK default. Hardcoding `true`, deleting every app-owned start path, and keeping the exact SDK pin together form the source-level fail-closed contract.

This PR does not claim the third-party package lacks replay code. It proves Proliferate's Web source has no configuration or call path that enables it.

## Scope (exactly seven files)

| File | Change |
| --- | --- |
| `apps/web/src/browser/telemetry/install-web-telemetry.ts` | Delete `WebTelemetryConfig.posthog.sessionRecordingEnabled` and its `VITE_PROLIFERATE_POSTHOG_SESSION_RECORDING_ENABLED` read; hardcode `disable_session_recording: true`; delete the dormant `session_recording` options object |
| `apps/web/src/browser/telemetry/telemetry-privacy.test.ts` | Add the 17-row retired-variable tombstone matrix, a preserved intentional-capture test, and a telemetry-disabled no-init test |
| `apps/web/.env.example` | Delete the retired variable and its enablement comments (no commented compatibility setting left) |
| `specs/developing/reference/env-vars.yaml` | Delete the now-unowned catalog entry |
| `specs/codebase/systems/engineering/analytics/posthog.md` | Web section becomes source-disabled truth; Web configuration list drops the retired variable; table rows advance to Desktop-and-Web |
| `specs/codebase/systems/engineering/analytics/README.md` | Remove the resolved Desktop/Web route-id replay gap, preserving the anonymous-credential and live-provider gaps |
[...]

PR #2093, opened 2026-08-19
fix(desktop): source-disable PostHog session recording (CP-C1PD)

pablonyx:
## Summary

CP-C1PD makes PostHog session recording impossible to enable on Desktop through any build or environment value. Desktop still initializes PostHog in `hosted_product` mode when its normal key and telemetry gates permit, but the single `posthog.init` call now passes `disable_session_recording: true` literally.

The dormant activation surfaces are deleted in the same PR:

- `VITE_PROLIFERATE_POSTHOG_SESSION_RECORDING_ENABLED` removed from `apps/desktop/.env.example` and the `ImportMetaEnv` declaration in `assets.d.ts`
- `DesktopTelemetryConfig.posthog.sessionRecordingEnabled` and its `envFlagEnabled(...)` read removed from `config.ts` (the helper itself stays; it still owns other booleans)
- the conditional `session_recording` mask/block object and the `loaded` callback that called `startSessionRecording()` deleted from `posthog.ts`
- the stale fixture field removed from `posthog.test.ts` and `client.test.ts`

## Privacy rationale

The pinned SDK (`posthog-js` 1.386.8) defaults `disable_session_recording` to `false`, so omitting the option is not fail-closed. The recorder's `_isRecordingEnabled` requires both provider-side enablement and `!config.disable_session_recording`, and `startSessionRecording(...)` re-opens it via `set_config`. Explicitly passing `true` and closing both first-party reopening routes (no recording-start call, no `set_config`/`setConfig` in non-test Desktop source) means a PostHog project replay setting alone cannot start Desktop recording.

This is consistent with the landed CP-S2A identity truth (UUID-only `identify`) and the CP-C1S documentation truth that Desktop renderer Sentry replay is source-disabled and absent. Recorded page metadata can retain identifier-bearing route segments even with masking, which is why the fix is at the source and not a default value.

Re-enablement is a separate founder-approved behavior change that must first prove a synthetic sensitive-route, masking, metadata, and provider-arrival contract. No dormant Desktop switch or compatibility branch is retained.

Unchanged: typed product capture and the event allowlist, id-only `identify`, `reset(true)`, `getDesktopPostHogSupportRefs()` (including `get_session_id()`, an analytics correlation reference, not a recording signal), registered app/surface/environment/release context, `before_send` scrubbing, and the literal `false` values for autocapture, pageview, and pageleave. No package.json or lockfile change; Web and Mobile semantics untouched.

## Documentation

[...]

PR #2083, opened 2026-08-19
fix(desktop): source-disable Sentry session replay

pablonyx:
## Summary
- Removes Desktop renderer Sentry session replay at the source: drops `replayIntegration`, pins `replaysSessionSampleRate` and `replaysOnErrorSampleRate` to `0`, and retires `VITE_PROLIFERATE_SENTRY_REPLAY_ON_ERROR_SAMPLE_RATE`.
- Ordinary Sentry error/performance reporting for the Desktop renderer is unchanged.
- Privacy rationale: session replay in the Desktop renderer could capture in-app content the org has not reviewed for privacy exposure; disabling it at the source (rather than relying on a fail-closed runtime gate) removes the capability entirely until a deliberate re-enable decision is made.
- Updates four docs to source-disabled truth: `guides/operating/analytics/sentry.md`, `specs/OBSERVABILITY.md`, `specs/codebase/systems/engineering/observability/README.md`, `specs/codebase/systems/engineering/observability/sentry.md`.
- Re-enabling replay requires a review.

Scope: exactly nine files (`apps/desktop/.env.example`, `apps/desktop/src/lib/integrations/telemetry/{config.ts,config.test.ts,sentry.ts,sentry.test.ts}`, and the four docs above). No other behavior touched.

## Observability
- Desktop renderer Sentry replay integration is removed and both replay sample rates are pinned to `0` in source, so no replay data can be captured or transmitted regardless of runtime config/env overrides.
- Non-replay Sentry telemetry (errors, performance) is unaffected.

## Testing
- `python3 scripts/check_docs.py` — passed (232 Markdown files).
- Rebased cleanly onto `origin/main` (36ffc26307) with no conflicts; post-rebase nine-file diff hash unchanged from pre-rebase (`c5001577...`).
- Local `pnpm install` / vitest / desktop `tsc` were skipped in this session due to tight machine swap headroom (free swap ~1.3GiB at time of work); CI will run the full frontend/desktop test and typecheck suites.

## Test plan
- [ ] CI: desktop lint/typecheck
- [ ] CI: vitest for `apps/desktop/src/lib/integrations/telemetry/{config,sentry}.test.ts`
- [ ] CI: docs check

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2179, opened 2026-08-21
fix(desktop): redact route ids from PostHog payloads and re-enable internal replay

pablonyx:
## Verdict up front

**Customer replay stays OFF.** Desktop PostHog session replay is re-enabled for the **internal audience only**. The route-id leak is fixed and proven at the payload level with negative controls, but the live qualification that #2093 and #2096 named as the re-enable bar (real recorder output from the running app, plus controlled provider arrival) was not executed tonight, so it is not widened to customers. Per Lane C's binding rule, that is the fallback, and it is stated plainly rather than argued away.

Nothing here is live. This is a pushed PR, not merged. Even if merged, nothing records until the PostHog project also enables replay server-side (see "Two independent gates" below).

Reviewed end to end by a second agent after the implementation commit. That review found and fixed two real defects: the original narrative claimed a recorder boundary that does not execute (`d76b9bba14`, section 2), and canvas plus console capture were left resolvable by the PostHog project rather than pinned in source (`aae0842cf3`, section 3). Neither changed the verdict; both are closed.

## 1. What the leak actually was

The disable PRs recorded the finding as one line in the analytics contract, deleted by #2093 at `specs/codebase/systems/engineering/analytics/README.md:56`:

> Desktop and Web replay can still expose route ids through recorded page URLs when explicitly enabled.

Reconstructed precisely, it is two channels, and **masking does not touch either one** — which is the answer to "why disable rather than mask". `maskAllInputs`, `maskTextSelector`, and `blockSelector` hide rendered page *content*. A URL is not page content. #2093's body says exactly this: "Recorded page metadata can retain identifier-bearing route segments even with masking, which is why the fix is at the source and not a default value."

**What leaks.** Opaque resource identifiers embedded in the path by these routes (`apps/packages/product-client/src/pages/AuthenticatedAppHost.tsx:79-85`, `apps/web/src/WebHostApp.tsx:45-46`):

```
/workflows/:workflowId
/workflows/:workflowId/runs/:runId
/workspaces/:workspaceId
/automations/:workflowId
/join/:orgId
/login/:slug
```

**Channel 1 — event properties.** posthog-js attaches the page URL to *every* capture, `$snapshot` included, in `calculateEventProperties` (`posthog-js@1.386.8` `src/posthog-core.ts:1490-1560`, read from the shipped sourcemap): `$current_url`, `$pathname`, `$referrer`, `$session_entry_url`, `$prev_pageview_pathname`, `$initial_current_url` under `$set_once`, and more the SDK adds over time.

Desktop's `before_send` did not remove any of them, and could not have. `scrubPostHogPayload` calls the shared scrubber with `preservePostHogInternalKeys: true`, and `isPostHogInternalKey` (`apps/packages/product-client/src/domain/telemetry/scrub.ts:83-85`) exempts every `$`-prefixed key from key redaction. The value then goes through `scrubTelemetryUrl` (`scrub.ts:52-68`), which **deliberately keeps the path** and only strips query and hash:

```ts
const absoluteMatch = ABSOLUTE_URL_PARTS_PATTERN.exec(value);
if (absoluteMatch) {
  return `${absoluteMatch[1]}${absoluteMatch[2] || ""}`;   // scheme://host + PATH
}
```

So `https://app.proliferate.com/workflows/<uuid>/runs/<uuid>` survived verbatim.

**Sibling leak, live today, independent of replay:** this same channel means **shipped Desktop builds have been sending `$current_url` and `$pathname` with raw workflow, run, and workspace ids on every allowlisted product event**, with replay off. Confirmed against `main`: `git show origin/main:apps/desktop/src/lib/integrations/telemetry/scrub.ts` contains no URL-property deletion, while Web deletes five keys at `apps/web/src/browser/telemetry/install-web-telemetry.ts:294-300` (`$current_url`, `$pathname`, `$host`, `$referrer`, `$referring_domain`). Desktop never did. This PR closes it as a side effect, because the redaction runs on every capture and not only on `$snapshot`.

**Channel 2 — the rrweb stream.** Deleting URL properties, as Web did, still would not have fixed replay, because the recorded page URL is *also* inside `properties.$snapshot_data`:

- the rrweb Meta event, `{ type: 4, data: { href } }`, emitted at recording start and at every full snapshot;
- serialized DOM nodes, as `attributes.href` on `<a>`/`<link>` and `attributes.src` on media;
- PostHog's own custom rrweb events (`type: 5`) and network-capture entries.

`$snapshot` reaches `before_send` (`_captureSnapshot` -> `capture('$snapshot', ...)` at `src/extensions/replay/external/lazy-loaded-session-recorder.ts:1656-1664`, and `_runBeforeSend` at `src/posthog-core.ts:1410-1415`), so `before_send` is the right chokepoint. But the shared scrubber cannot be that fix: `MAX_SCRUB_DEPTH = 10`, `MAX_SCRUB_ARRAY_ITEMS = 100`, `MAX_SCRUB_OBJECT_PROPERTIES = 100` (`scrub.ts:23-26`) turn a real DOM snapshot into `[truncated]` markers while still leaving shallow `href`s intact. **It leaks and destroys the recording at the same time.** Both halves are asserted as executable evidence in `route-id-redaction.test.ts` under `describe("the leak this module exists to close")`.

## 2. The fix

New shared module `apps/packages/product-client/src/domain/telemetry/route-id-redaction.ts`. Allowlist-shaped, not pattern-shaped: a pathname is matched against a closed table of route templates and **replaced by the template it matched**. Nothing that is not already a literal in that file can survive. A pathname matching nothing becomes `/unknown`, so a new id-bearing route added tomorrow fails closed instead of leaking.

### One live boundary, not two (corrected in review)

The original version of this PR described the fix as applied at two boundaries, recorder (`session_recording.maskAttributeFn`) and capture (`before_send`), "because neither alone is sufficient". **That was wrong. The recorder boundary does not execute at the pinned SDK version.**

Evidence, all against `node_modules/.pnpm/posthog-js@1.386.8/node_modules/posthog-js`:

| Option | `lazy-recorder.js` | `recorder.js` | `module.js` | `main.js` |
| `maskAllInputs` | 11 | 3 | 0 | 0 |
| `maskTextSelector` | 23 | 15 | 0 | 0 |
| `blockSelector` | 38 | 30 | 0 | 0 |
| `maskCapturedNetworkRequestFn` | 8 | 0 | 0 | 0 |
| `recordHeaders` / `recordBody` | 15 / 20 | 7 / 12 | 0 | 0 |
| **`maskAttributeFn`** | **0** | **0** | **0** | **0** |

`maskAttributeFn` occurs **zero times in the entire package**. The SDK builds its rrweb masking object from exactly three keys, in the `Th` getter of `dist/lazy-recorder.js`:

```js
u = {
  maskAllInputs:    config.session_recording?.maskAllInputs,
  maskTextSelector: config.session_recording?.maskTextSelector,
  blockSelector:    config.session_recording?.blockSelector
}
```

It typechecks only because the transitive `@posthog/types@1.404.1` is newer than the `posthog-js@1.386.8` runtime pin (`apps/desktop/package.json:51`, exact, no caret) and declares the option. Excess-property checking does not catch it either, because `DESKTOP_SESSION_RECORDING_OPTIONS` is a named const rather than an inline object literal.

**The leak is still closed**, because `before_send` independently walks `$snapshot_data` and reduces URL-valued DOM attributes, and the CI-enforced payload test never invokes the recorder callback. But a reader who believed the recorder covered DOM attributes could narrow `before_send` to the Meta event and reopen the 2026-08-18 leak. Commit `d76b9bba14` therefore corrects the claim in source (`posthog.ts`, `route-id-redaction.ts`), in both specs, and in the desktop test comment, and adds a CI-enforced regression test that pins the invariant. `maskAttributeFn` stays wired as dormant forward-compatibility for a future SDK pin; it is a no-op today, and enabling it later is additive.

### What actually runs

| Boundary | Mechanism | Status | Covers |
| --- | --- | --- | --- |
| Capture | `before_send` | **live, load-bearing** | `$current_url`-style properties, `$set`/`$set_once`, the rrweb Meta `href`, DOM attribute `href`/`src` inside `$snapshot_data`, custom-event payloads |
| Recorder | `session_recording.maskAttributeFn` | **dormant** (not invoked by posthog-js 1.386.8) | nothing today |

Behaviour worth reviewing:

- **URL-valued names fail closed.** `href`, `src`, `srcset`, `action`, `poster`, `url`, `pathname`, … that cannot be bounded become `[redacted-url]` rather than passing through. Free text is only rewritten when it actually parses as a URL or absolute path.
- **`data:`, `blob:`, `mailto:`, `tel:`, and `#fragment` pass through.** They carry no route path and reducing them would cost fidelity for no privacy gain. Flagged as a deliberate residual.
- **The replay stream is held out of the generic scrubber** in `apps/desktop/src/lib/integrations/telemetry/scrub.ts`. Running `$snapshot_data` through the depth-capped scrubber shreds it. The stream gets the rrweb-aware walk instead, which has no depth cap and rewrites only URL-bearing and content-bearing fields. Negative control 3 below proves this hold-out is load-bearing.
- **Content-bearing attributes get the shared text scrubber.** `title`, `alt`, `placeholder`, `aria-label`, `value`, `content`, `download` are not text nodes or inputs, so recorder masking never touches them; an absolute path or token that reached a tooltip is now redacted.
- **Budget, fail-closed.** Above `MAX_REDACTION_VISITS` (250,000) the capture is dropped (`before_send` returns `null`) rather than forwarded unredacted.

Recorder configuration, pinned in source at `apps/desktop/src/lib/integrations/telemetry/posthog.ts` as `DESKTOP_SESSION_RECORDING_OPTIONS`: `maskAllInputs: true`, `maskTextSelector: "*"` (**all** text masked), `blockSelector: "[data-telemetry-block]"`, no fonts, no cross-origin iframes, `recordHeaders: false`, `recordBody: false`, and `maskCapturedNetworkRequestFn: () => null` (network capture refused outright). Every one of those is honored by the pinned runtime per the table above.

## 3. Capabilities the provider could otherwise decide (found in review)

The #2093/#2096 doctrine is that no provider setting alone can start recording. The audience gate holds that line for *whether* recording starts. Review found that two settings for *what gets recorded* were still the provider's to decide, because Desktop left them unset and posthog-js resolves them local-config-first with a remote-flags fallback.

**Canvas**, `Rh` getter in `posthog-js@1.386.8` `dist/lazy-recorder.js`:

```js
config.session_recording.captureCanvas?.recordCanvas ?? remoteConfig.canvasRecording?.enabled
```

**Console**, `xh` getter in the same file:

```js
config.enable_recording_console_log != null
  ? config.enable_recording_console_log
  : remoteConfig.consoleLogRecordingEnabled
```

Both matter for this app specifically:

- `@xterm/addon-canvas` (`apps/desktop/package.json:44`, `apps/packages/product-client/package.json:102`) renders terminal output to a canvas. Canvas frames are captured as **pixels**, which `maskTextSelector: "*"` does not reach. Flipping canvas recording in the PostHog UI would have captured terminal contents.
- Console arguments are arbitrary app strings, and the route-id redactor deliberately does **not** rewrite free text inside the rrweb stream, because doing so would corrupt the replay. Console capture would therefore have been both remotely enableable and unredacted.

`aae0842cf3` pins both as local literals, which win their respective resolutions: `captureCanvas: { recordCanvas: false }` and `enable_recording_console_log: false`. The `@posthog/types@1.404.1` doc comment on `captureCanvas` confirms the intent verbatim: "Allows local config to override remote canvas recording settings from the flags response."

Negative control, both pins removed:

```
× desktop PostHog adapter > initializes once with recording disabled and no auto-start surface
  → expected { …(6) } to deeply equal { …(7) }
× desktop PostHog adapter > pins the recorder to mask content and redact route identifiers
  → expected undefined to deeply equal { recordCanvas: false }

Tests  2 failed | 48 passed (50)
```

Restored, 50/50 green. Both options were typechecked against the real `posthog-js` `PostHogConfig` with a scratch probe (deleted, not in the diff).

**Network headers and bodies resolve the other way** (`Oh` getter: `local.recordHeaders || remote.recordHeaders`), so a local `false` cannot pin them. It does not matter: `maskCapturedNetworkRequestFn: () => null` is honored, and `h.maskRequestFn` maps a `null` return to `undefined`, dropping every captured request whatever the provider enables. Documented rather than "fixed" with a value that would not have worked.

## 4. Per-surface status

| Surface | Product | Disabled by | State after this PR |
| --- | --- | --- | --- |
| Desktop | PostHog replay | #2093 | **Internal audience only.** Start-gated in source, customer recording off |
| Web | PostHog replay | #2096 | **Still source-disabled, untouched.** `install-web-telemetry.ts:330` `disable_session_recording: true`; `telemetry-privacy.test.ts:341-348` asserts no `session_recording` key and `startSessionRecording` never called |
| Mobile | PostHog replay | #2097 | **Still source-disabled, untouched.** `apps/mobile/src/lib/integrations/telemetry/posthog.ts:45` `enableSessionReplay: false` |
| Desktop renderer | **Sentry** replay | #2083 | **Still source-disabled, untouched.** `sentry.ts:83-84` both sample rates `0`. Different product from the three above; this PR does not go near it |

`git diff --name-only origin/main...HEAD` touches no `apps/web` and no `apps/mobile` file. The shared redactor is in place for whenever those come back.

## 5. The proof, and the negative controls

Tests live in `apps/packages/product-client/src/domain/telemetry/` because that is the only place CI actually runs vitest for this code (`ci.yml:526` runs `vitest run src/domain`, `ci.yml:573` runs the full sharded suite). There is **no `apps/desktop` and no `apps/web` vitest job in CI** — same gap #2096 reported — so the desktop suites below were run locally and are reported as such.

The load-bearing test builds a realistic `$snapshot` capture with four distinctive identifiers, a Meta event, a DOM snapshot nested 16 levels deep, an attribute mutation, and two custom events, then asserts the identifiers are **absent from `JSON.stringify` of the whole redacted payload** and that the structure survives.

```
pnpm --filter @proliferate/product-client exec vitest run src/domain/telemetry
  route-id-redaction.test.ts   50 passed
  replay-audience.test.ts      16 passed
  scrub.test.ts                15 passed   (unchanged, still green)
  Tests  81 passed (81)

pnpm --filter proliferate exec vitest run src/lib/integrations/telemetry
  Tests  50 passed (50)   (local only; not a CI job)
```

### Negative controls — re-run independently by the reviewing agent on the final tree

Every one below was executed, its literal failure captured, and the source restored with `git checkout --` (never `git stash`). Working tree confirmed clean after each.

**1. Domain redaction disabled** (`redactRouteIdentifiersInCapture` returns the event unchanged). This is the privacy control that matters:

```
× redactRouteIdentifiersInCapture > removes every route identifier from a replay capture
  → expected '{"event":"$snapshot","properties":{"$…' not to contain 'wf-8e1c4a92-route-id-leak-probe'
× redactRouteIdentifiersInCapture > closes the DOM-attribute half without the recorder callback
× redactRouteIdentifiersInCapture > replaces the recorded page URL with a bounded route template
  → expected 'https://app.proliferate.com/workflows…' to be 'https://app.proliferate.com/workflows…'
× redactRouteIdentifiersInCapture > redacts the rrweb Meta event href, which no property deletion reaches
× redactRouteIdentifiersInCapture > redacts href attributes nested below the shared scrubber's depth cap
  → expected '{"event":"$snapshot","properties":{"$…' to contain '/workspaces/:workspaceId'
× redactRouteIdentifiersInCapture > redacts attribute-mutation and custom rrweb events
× redactRouteIdentifiersInCapture > scrubs content-bearing attributes that recorder masking does not cover
  → expected '{"event":"$snapshot","properties":{"$…' not to contain '/Users/pablo/proliferate'
× redactRouteIdentifiersInCapture > redacts ordinary product events, not only replay snapshots
× redactRouteIdentifiersInCapture > redacts the top-level $set and $set_once siblings of properties
× redactRouteIdentifiersInCapture > drops a capture that exceeds the redaction budget rather than sending it
  → expected { event: '$snapshot', …(1) } to be null

Tests  10 failed | 71 passed (81)
```

**2. Desktop `before_send` wiring reverted** (`scrubPostHogPayload` back to `scrubPostHogProperties(event)` only):

```
× desktop PostHog adapter > strips route identifiers from the replay payload posthog would send
  → expected '{"event":"$snapshot","properties":{"$…' not to contain 'wf-8e1c4a92-route-id-leak-probe'
× desktop PostHog adapter > still scrubs non-replay captures through the shared scrubber
  → expected 'https://app.proliferate.com/workspace…' to be 'https://app.proliferate.com/workspace…'

Tests  2 failed | 48 passed (50)
```

Reverting the `maskAttributeFn` wiring to identity on top of that adds exactly one more failure (`Tests 3 failed | 47 passed (50)`) — and that third failure is a direct unit call on the callback, not a payload assertion, which is consistent with the recorder boundary being dormant.

**3. Snapshot hold-out removed** (replay stream sent back through the depth-capped scrubber):

```
× desktop PostHog adapter > strips route identifiers from the replay payload posthog would send
  → expected '{"event":"$snapshot","properties":{"$…' to contain '/workspaces/:workspaceId'

Tests  1 failed | 49 passed (50)
```

The deep `href` never reaches the redactor because the scrubber already replaced that subtree with `[truncated]`.

After restoring each, both suites are green again (81/81 and 50/50, re-run and shown above).

## 6. Two independent gates before anything records

1. **Audience gate, in source.** `disable_session_recording: true` is a literal at init and there is no `loaded` callback, so nothing auto-starts. `client.ts:setTelemetryUser` calls `startDesktopPostHogSessionReplay()` only when `isInternalReplayAudience(user.email)` is true. The audience is a closed source-owned list (`replay-audience.ts`, currently `proliferate.com` and `proliferate.dev`) — **no environment variable, build value, or provider setting widens it**, which keeps the #2093/#2096/#2097 doctrine intact. The address is evaluated locally and never transmitted; PostHog identity stays UUID-only, and a test asserts the address never appears in any call payload. Sign-out stops recording before `reset(true)`.

   **Pablo: confirm the two domains are right, and add any others you want recording.**

2. **Provider gate, outside source.** `_isRecordingEnabled` requires `enabled_server_side && !disable_session_recording && !optedOut` (`posthog-js@1.386.8` `src/extensions/replay/session-recording.ts:81-86`). Nothing records until the PostHog project itself enables replay. Merging this PR does not start a single recording on its own.

## 7. What is NOT done, and why

- **Customer replay.** Off. Requires the live qualification: drive the real app through a sensitive route, capture the real recorder output, and confirm arrival in the PostHog project. Not runnable tonight, and the posture is not certified on a synthetic payload alone.
- **Web and Mobile.** Untouched, still source-disabled. See the per-surface table in section 3.
- **Pre-disable builds.** 0.4.14 / 0.4.17 are still sending replays with the leak. Nothing in this PR reaches them; that exposure ages out with the installs.
- **The main workspace/chat surface is masked but not blocked.** `[data-telemetry-block]` is applied by `ProductPageShell` (when `telemetryBlocked` is set, as the workflows surfaces do), `SettingsScreen`, `ModalShell`, and `CommandPalette` — not by the workspace shell. With `maskTextSelector: "*"` every text node and input is masked, but images and canvas are not. Acceptable for internal-only recording; **must be settled before customer replay**. The docs now say this instead of the previous "workspace and settings surfaces are blocked by default", which was not true.
- **`data:`/`blob:` URL passthrough** in URL-valued attributes, described above.
- **CPU cost.** The rrweb walk allocates a redacted copy per capture. Bounded by the visit budget and only meaningful on full snapshots; not measured on device. [unverified]

## 8. Sibling leaks and defects spotted

1. **Desktop `$current_url` on ordinary events, live in production right now.** Described in section 1, confirmed against `origin/main`. Closed by this PR.
2. **`maskAttributeFn` was inert dead config presented as live coverage.** Section 2. Corrected in `d76b9bba14`; no behavior change, but the narrative and the docs were wrong and a future reader could have reopened the leak on the strength of them.
3. **Canvas and console capture were the PostHog project's to decide.** Section 3. Canvas is the sharper one, because xterm renders the terminal to a canvas and canvas is pixels. Closed by `aae0842cf3`.
4. **The generic scrubber destroys any large nested payload it is pointed at.** Not specific to replay: any deeply nested telemetry container hits `MAX_SCRUB_DEPTH = 10` and comes back as `[truncated]`. Worth knowing before anyone routes another structured payload through it.
5. **`apps/web` and `apps/desktop` have no vitest job in CI** (`ci.yml` only runs `@proliferate/product-client` filters), so `apps/web/src/browser/telemetry/telemetry-privacy.test.ts` — the file that proves Web recording stays disabled — is not enforced by any required check, and neither is the desktop `before_send` wiring proof (negative control 2). The redaction *logic* is CI-enforced in product-client; the desktop *wiring* is not. Worth a separate fix.
6. **`npx tsc` in this repo does not run TypeScript.** It resolves to the `tsc` tab-completer package and prints "This is not the tsc command you are looking for", exiting non-zero without compiling. An earlier revision of this PR body cited `npx tsc -p …` as passing evidence; that line was false and is corrected below. Use `pnpm --filter <pkg> exec tsc` instead.

## Testing

Run locally, green:

```
pnpm --filter @proliferate/product-client exec vitest run src/domain/telemetry   # 81 passed
pnpm --filter proliferate exec vitest run src/lib/integrations/telemetry         # 50 passed
pnpm --filter @proliferate/product-client exec tsc -p tsconfig.domain.json --noEmit
    # 215 errors, 0 in telemetry/ — all pre-existing "Cannot find module
    # '@proliferate/cloud-sdk'" from the un-generated SDK (generate needs cargo,
    # and Lane B holds the build lock). CI runs the real typecheck with the SDK built.
python3 scripts/check_max_lines.py                                               # passed
python3 scripts/check_docs.py                                                    # passed, 235 files
python3 scripts/check_frontend_boundaries.py                                     # passed
python3 scripts/report_frontend_structure.py --strict --summary-only             # TOTAL: 0
git diff --check                                                                 # clean
```

Full `pnpm shared:typecheck`, `Desktop frontend build`, and the sharded product-client suite were **not** run locally: they need the SDK generate step (cargo, and Lane B holds that lock all night). CI covers all three, and all three were green on the parent commit.

## Test plan

- [x] CI on `3f5878666c`: all 39 checks pass (run 32460112478), including `Shared frontend checks`, sharded `Shared frontend packages (1-4)`, `Desktop frontend build`, `Web frontend build`, `Mobile typecheck`, `Repo shape checks`, `CodeQL`, `Validate PR title and labels`
- [x] CI on `aae0842cf3` (head): **36/36 pass, 0 fail**, rest skipping
- [ ] Pablo: confirm the internal email-domain list (`proliferate.com`, `proliferate.dev`)
- [ ] Pablo: rule on whether the workspace shell must be blocked before customer replay

### DIFF
--- a/apps/desktop/src/lib/integrations/telemetry/client.test.ts
+++ b/apps/desktop/src/lib/integrations/telemetry/client.test.ts
@@ -53,6 +53,8 @@ const mocks = vi.hoisted(() => {
     identifyDesktopPostHogUserMock: vi.fn(),
     resetDesktopPostHogUserMock: vi.fn(),
     trackDesktopPostHogEventMock: vi.fn(),
+    startDesktopPostHogSessionReplayMock: vi.fn(),
+    stopDesktopPostHogSessionReplayMock: vi.fn(),
   };
 });
 
@@ -90,6 +92,8 @@ vi.mock("./posthog", () => ({
   identifyDesktopPostHogUser: mocks.identifyDesktopPostHogUserMock,
   resetDesktopPostHogUser: mocks.resetDesktopPostHogUserMock,
   trackDesktopPostHogEvent: mocks.trackDesktopPostHogEventMock,
+  startDesktopPostHogSessionReplay: mocks.startDesktopPostHogSessionReplayMock,
+  stopDesktopPostHogSessionReplay: mocks.stopDesktopPostHogSessionReplayMock,
 }));
 
 async function loadTelemetryClient() {
@@ -123,6 +127,8 @@ describe("desktop telemetry client", () => {
     mocks.identifyDesktopPostHogUserMock.mockClear();
     mocks.resetDesktopPostHogUserMock.mockClear();
     mocks.trackDesktopPostHogEventMock.mockClear();
+    mocks.startDesktopPostHogSessionReplayMock.mockClear();
+    mocks.stopDesktopPostHogSessionReplayMock.mockClear();
   });
 
   it("initializes vendor telemetry only when vendor routing is enabled", async () => {
@@ -197,6 +203,72 @@ describe("desktop telemetry client", () => {
         mocks.identifyDesktopPostHogUserMock.mock.calls,
       ]),
     ).not.toContain("Test User");
+  });
+
+  it("leaves session replay off for a customer account", async () => {
+    const client = await loadTelemetryClient();
+
+    client.setTelemetryUser({
+      id: "user-123",
+      email: "someone@customer.example",
+      display_name: null,
+    });
+
+    expect(mocks.identifyDesktopPostHogUserMock).toHaveBeenCalledWith("user-123");
+    expect(mocks.startDesktopPostHogSessionReplayMock).not.toHaveBeenCalled();
+  });
+
+  it("starts session replay for the internal audience only", async () => {
+    const client = await loadTelemetryClient();
+
+    client.setTelemetryUser({
+      id: "user-456",
+      email: "pablo@proliferate.com",
+      display_name: null,
+    });
+
+    expect(mocks.startDesktopPostHogSessionReplayMock).toHaveBeenCalledOnce();
+    // The address is a local decision input, never a transmitted property.
+    expect(
+      JSON.stringify(mocks.startDesktopPostHogSessionReplayMock.mock.calls),
+    ).not.toContain("pablo@proliferate.com");
+    expect(
+      JSON.stringify(mocks.identifyDesktopPostHogUserMock.mock.calls),
+    ).not.toContain("pablo@proliferate.com");
+  });
+
+  it("does not start replay when vendor routing is off", async () => {
+    mocks.runtimeState = {
+      disabled: false,
+      telemetryMode: "self_managed",
+      anonymousEnabled: true,
+      vendorEnabled: false,
+    };
+
+    const client = await loadTelemetryClient();
+    client.setTelemetryUser({
+      id: "user-456",
+      email: "pablo@proliferate.com",
+      display_name: null,
+    });
+
+    expect(mocks.startDesktopPostHogSessionReplayMock).not.toHaveBeenCalled();
+  });
+
+  it("stops replay before clearing identity on sign-out", async () => {
+    const client = await loadTelemetryClient();
+
+    client.setTelemetryUser({
+      id: "user-456",
+      email: "pablo@proliferate.dev",
+      display_name: null,
+    });
+    client.clearTelemetryUser();
+
+    expect(mocks.stopDesktopPostHogSessionReplayMock).toHaveBeenCalledOnce();
+    expect(
+      mocks.stopDesktopPostHogSessionReplayMock.mock.invocationCallOrder[0],
+    ).toBeLessThan(mocks.resetDesktopPostHogUserMock.mock.invocationCallOrder[0]);
   });
 
   it("emits desktop_keychain_access_failed to PostHog", async () => {
--- a/apps/desktop/src/lib/integrations/telemetry/client.ts
+++ b/apps/desktop/src/lib/integrations/telemetry/client.ts
@@ -8,6 +8,7 @@ import {
   resolveDesktopTelemetryRoutingState,
 } from "@proliferate/product-client/internal/lib/domain/telemetry/mode";
 import type { AuthUser } from "@proliferate/product-client/internal/lib/domain/auth/auth-user";
+import { isInternalReplayAudience } from "@proliferate/product-client/internal/domain/telemetry/replay-audience";
 import {
   getProliferateApiBaseUrl,
   getProliferateApiOrigin,
@@ -22,6 +23,8 @@ import {
   initializeDesktopPostHog,
   getDesktopPostHogSupportRefs,
   resetDesktopPostHogUser,
+  startDesktopPostHogSessionReplay,
+  stopDesktopPostHogSessionReplay,
   trackDesktopPostHogEvent,
 } from "./posthog";
 import {
@@ -144,6 +147,15 @@ export function setTelemetryUser(user: AuthUser): void {
 
   setDesktopSentryUser(user.id);
   identifyDesktopPostHogUser(user.id);
+
+  // Session replay is staged: the route-identifier leak that caused the
+  // 2026-08-18 disable is fixed and proven at the payload level, but the live
+  // synthetic qualification #2093/#2096 required has not run, so recording is
+  // limited to the internal audience. The address is read locally to make this
+  // decision and is never sent: PostHog identity stays UUID-only above.
+  if (isInternalReplayAudience(user.email)) {
+    startDesktopPostHogSessionReplay();
+  }
 }
 
 export function clearTelemetryUser(): void {
@@ -151,6 +163,8 @@ export function clearTelemetryUser(): void {
     return;
   }
 
+  // Stop before reset so no frames are recorded against a signed-out session.
+  stopDesktopPostHogSessionReplay();
   clearDesktopSentryUser();
   resetDesktopPostHogUser();
 }
--- a/apps/desktop/src/lib/integrations/telemetry/posthog.test.ts
+++ b/apps/desktop/src/lib/integrations/telemetry/posthog.test.ts
@@ -64,6 +64,8 @@ describe("desktop PostHog adapter", () => {
       capture_pageleave: false,
       person_profiles: "identified_only",
       disable_session_recording: true,
+      // Local literal, so the PostHog project cannot start console capture.
+      enable_recording_console_log: false,
     });
     const scrub = await import("./scrub");
     expect(beforeSend).toBe(scrub.scrubPostHogPayload);
@@ -75,6 +77,7 @@ describe("desktop PostHog adapter", () => {
       "capture_pageleave",
       "capture_pageview",
       "disable_session_recording",
+      "enable_recording_console_log",
       "person_profiles",
       "session_recording",
     ]);
@@ -95,6 +98,12 @@ describe("desktop PostHog adapter", () => {
     expect(options.recordHeaders).toBe(false);
     expect(options.recordBody).toBe(false);
     expect(options.maskCapturedNetworkRequestFn()).toBeNull();
+
+    // Canvas frames are pixels; `maskTextSelector: "*"` does not reach them,
+    // and `@xterm/addon-canvas` renders terminal output to a canvas. This must
+    // stay a local literal `false` or the PostHog project's
+    // `canvasRecording.enabled` decides on its own.
+    expect(options.captureCanvas).toEqual({ recordCanvas: false });
 
     // Dormant forward-compatibility, NOT coverage: the pinned
     // posthog-js@1.386.8 forwards only maskAllInputs/maskTextSelector/
--- a/apps/desktop/src/lib/integrations/telemetry/posthog.test.ts
+++ b/apps/desktop/src/lib/integrations/telemetry/posthog.test.ts
@@ -96,7 +96,12 @@ describe("desktop PostHog adapter", () => {
     expect(options.recordBody).toBe(false);
     expect(options.maskCapturedNetworkRequestFn()).toBeNull();
 
-    // The recorder-boundary half of the route-id fix.
+    // Dormant forward-compatibility, NOT coverage: the pinned
+    // posthog-js@1.386.8 forwards only maskAllInputs/maskTextSelector/
+    // blockSelector to rrweb and never invokes this callback. It is asserted
+    // so the wiring stays correct for a future SDK pin. The live proof that
+    // route ids never reach a payload is the before_send boundary, in
+    // product-client's route-id-redaction.test.ts.
     expect(options.maskAttributeFn("href", "/workflows/wf-secret-id")).toBe(
       "/workflows/:workflowId",
     );
--- a/apps/desktop/src/lib/integrations/telemetry/posthog.test.ts
+++ b/apps/desktop/src/lib/integrations/telemetry/posthog.test.ts
@@ -7,6 +7,7 @@ const mocks = vi.hoisted(() => ({
   identify: vi.fn(),
   reset: vi.fn(),
   startSessionRecording: vi.fn(),
+  stopSessionRecording: vi.fn(),
 }));
 
 vi.mock("posthog-js", () => ({
@@ -36,9 +37,10 @@ describe("desktop PostHog adapter", () => {
     mocks.identify.mockReset();
     mocks.reset.mockReset();
     mocks.startSessionRecording.mockReset();
+    mocks.stopSessionRecording.mockReset();
   });
 
-  it("initializes once with recording disabled and no recording surfaces", async () => {
+  it("initializes once with recording disabled and no auto-start surface", async () => {
     const adapter = await loadDesktopPostHog();
     adapter.initializeDesktopPostHog(ENABLED_CONFIG);
     adapter.initializeDesktopPostHog(ENABLED_CONFIG);
@@ -50,7 +52,11 @@ describe("desktop PostHog adapter", () => {
       Record<string, unknown>,
     ];
     expect(apiKey).toBe("phc_test");
-    const { before_send: beforeSend, ...rest } = options;
+    const {
+      before_send: beforeSend,
+      session_recording: sessionRecording,
+      ...rest
+    } = options;
     expect(rest).toEqual({
       api_host: "https://us.i.posthog.com",
       autocapture: false,
@@ -61,6 +67,7 @@ describe("desktop PostHog adapter", () => {
     });
     const scrub = await import("./scrub");
     expect(beforeSend).toBe(scrub.scrubPostHogPayload);
+    expect(sessionRecording).toBe(adapter.DESKTOP_SESSION_RECORDING_OPTIONS);
     expect(Object.keys(options).sort()).toEqual([
       "api_host",
       "autocapture",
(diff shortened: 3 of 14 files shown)


pablonyx (commented):
VERDICT: merge-safe (internal-only replay)

Fresh reviewer of record; I did not implement this. I re-derived every load-bearing claim against the installed `posthog-js@1.386.8` bundle and the head-ref sources rather than trusting the body. CI 36/36 green is the build evidence for the CI-enforced half.

## The most load-bearing check

**`before_send` really is the chokepoint, and `maskAttributeFn` really is dormant.** Verified directly against `node_modules/.pnpm/posthog-js@1.386.8`:

- `maskAttributeFn` appears in **zero files** in the entire installed package (`grep -rl`, not just `dist`). The `Th` getter in `dist/lazy-recorder.js` builds the rrweb masking object from exactly `maskAllInputs`, `maskTextSelector`, `blockSelector` — nothing else. The body's section-2 correction is accurate, and the source comment at `apps/desktop/src/lib/integrations/telemetry/posthog.ts:19-31` is the right place for it.
- Every *other* key in `DESKTOP_SESSION_RECORDING_OPTIONS` occurs in `dist` (`collectFonts` 84, `recordCrossOriginIframes` 204, `maskCapturedNetworkRequestFn` 109, `captureCanvas` 22, `enable_recording_console_log` 59). **No other key in the diff is a silent no-op of the same class.**
- `$snapshot` goes through the ordinary capture path: `ud(t){this._instance.capture("$snapshot",t,{...})}` in `dist/lazy-recorder.js`, and `capture()` runs `Pn` (before_send) with `if(!E)return` on a null return. So `before_send` sees the replay stream and a `null` genuinely drops it.
- Canvas: `get Rh(){...o?.recordCanvas ?? u?.enabled}` — a local literal `false` wins the `??`. Console: `get xh(){...return null!=r?r:e}` — local literal wins. Both pins in `posthog.ts:44-51` and `:84-91` do what they claim. Given `@xterm/addon-canvas`, the canvas pin is the sharper of the two and I agree it was a real hole.
- Provider gate: `get Qs(){...return t&&i&&r&&!s}` requires the persisted server-side flag; `startSessionRecording()` with no argument only calls `set_config({disable_session_recording:!1})`. Gate 2 is real.

## Checks 1-5

1. **Fail-closed allowlist — holds.** `boundRoutePathname` (`route-id-redaction.ts:190-201`) matches against the closed table and **returns the template literal itself**, so no path content can survive a match; segment counts must match exactly and `:param` requires a non-empty segment (`:171-184`); anything unmatched returns `/unknown`. `//workflows/x`, `/workflows/a/b`, `/playground/*`, `/admin/*` all fail closed. `boundRouteUrl` strips query and fragment via the `(?:[?#].*)?` tail, so a hash-routed URL loses its fragment entirely. Applied at `before_send` over both channels: `redactRouteIdentifiersInCapture` (`:329-361`) walks `$snapshot_data` with the rrweb-aware walk and everything else (plus `$set`/`$set_once`) with the property walk. Budget overrun returns `null` (`:359`) — the partially-redacted copy never escapes.
2. **Internal-only — holds.** `INTERNAL_REPLAY_EMAIL_DOMAINS` is a closed source list with exact-host match (`replay-audience.ts:26-48`); look-alikes, suffixes, subdomains, empty local parts and whitespace all reject. One production call site: `client.ts:156-158`. `diff --name-only` touches **no `apps/web`, no `apps/mobile`**; at head, `install-web-telemetry.ts:330` still `disable_session_recording: true`, mobile `posthog.ts:45` `enableSessionReplay: false`, desktop `sentry.ts:83-84` both replay rates `0`.
3. **Pins — holds** (verified above, plus `posthog.test.ts:73-83` locks the exact `init` option key set, which is a good ratchet).
4. **Negative controls — real.** The ten `it(...)` names quoted in the body's control 1 map 1:1 to `route-id-redaction.test.ts:195,204,219,229,238,271,306,323,339,359`, and the absence assertions use distinctive probe ids (`wf-8e1c4a92-route-id-leak-probe`) so they cannot pass on a substring. The snapshot hold-out is load-bearing and proven both ways: `describe("the leak this module exists to close")` at `:170-192` asserts the shared scrubber *both* leaves all four identifiers in *and* emits `[truncated]`, and `:235` asserts the redacted payload contains no `[truncated]`. I confirmed the scrubber facts it encodes: `scrub.ts:52-68` keeps the path, `:83-85` exempts `$`-keys, caps are 10/100/100 at `:24-26`.
5. **Config-key existence — holds** (see above).

## Findings

**F1 — read this first. The URL channel is closed; the *identifier* channel is not.** `redactRrwebValue` returns any string verbatim whose key is not in the three sets (`route-id-redaction.ts:279`), and `redactRouteIdentifiersInAttribute("id", ...)` passing through is asserted as intended (`route-id-redaction.test.ts:470-475`). The app really does put opaque resource ids into DOM attributes on the *unblocked* workspace surface: `StandardWorkspaceShell.tsx:242,244` (`data-workspace-session-id`, `data-pending-workspace-attempt-id`), `BackgroundWorkPane.tsx:241-242` (`data-workspace-id`, `data-session-id`), `TerminalViewport.tsx:32`, `ChatTabWithMenu.tsx:145,147`, `SubagentRosterRow.tsx:88`, `TerminalRosterRow.tsx:78`. So an internal Desktop replay still carries raw workspace/session/terminal/subagent ids — the same *class* of value #2093 disabled replay over, just arriving as a serialized attribute instead of a URL.

Not a blocker for internal-only: these are our own accounts' ids landing in our own project. But it is load-bearing for the customer decision, and it makes one spec line over-broad: `specs/frontend/telemetry.md` "Route identifiers never reach a replay payload" should read *URL-borne* route identifiers. Before widening, this wants either a `data-*-id` rule in the walk or `[data-telemetry-block]` on the workspace shell — which folds neatly into ruling #2 the body already asks for.

**F2 — the budget headroom claim is optimistic by roughly 10x.** `route-id-redaction.ts:144-150` says 250,000 visits is "an order of magnitude above" a snapshot of "tens of thousands of nodes". Visits are not nodes: one serialized element costs ~10 visits (node object, `type`, `tagName`, `id`, `childNodes` array, `attributes` object, plus one per attribute), and the recorder batches up to ~943KB across a 2s window into a single capture. A 20k-element workspace snapshot lands at ~200k, i.e. at the wall. Failure is safe (capture dropped, posthog logs the beforeSend rejection), but internal replays may silently go dark on the busiest screens. Worth measuring against a real snapshot and raising the constant.

**F3 — the desktop half of this change is guarded by tests nothing runs.** `ci.yml:462/465/474` are all `--filter @proliferate/product-client`; `apps/desktop` has a `test: vitest run` script that no workflow invokes. So the canvas/console pins, the `init` key-set lock, and the audience call site are protected only by locally-run suites. The body discloses this honestly as finding #5. The redaction *logic* is CI-enforced; the *wiring* is not. Follow-up, not a blocker — I verified the wiring by reading the head sources directly.

**F4 (nit)** — `specs/codebase/systems/engineering/analytics/posthog.md`: "Three recorder capabilities resolve local-config-first" then lists two and immediately explains the third resolves the other way. Say "two".

**F5 (nit)** — `boundRouteUrl` (`route-id-redaction.ts:213-216`) preserves `scheme://authority` verbatim, including any `user:pass@` userinfo the pattern's `[^/?#]*` admits. Harmless for a page URL, marginal for third-party `src`. Also `$snapshot_bytes` is not recomputed after redaction, so it drifts from the redacted stream length — cosmetic.

## On the two open rulings

Neither blocks an internal-only merge, and I checked rather than assumed:

- **Internal domain list.** `proliferate.com` / `proliferate.dev` is a closed source list with no env, build, or provider widening path, and `replay-audience.test.ts:34-36` pins the exact contents. A wrong list only means the wrong *internal* people do or do not get recorded; it cannot reach a customer. Pablo's call, at his leisure.
- **`[data-telemetry-block]` on the workspace shell before customer replay.** Explicitly scoped to customer widening, which this PR does not do. It does now inherit F1, so treat the two as one ruling.

Approving. The posture change is real, it is proven at the boundary that actually executes, the audience gate is airtight, and the body is unusually honest about what it did not prove — including correcting two of its own earlier claims. Do not widen past internal until F1 and the live qualification are settled.

PR #2180, opened 2026-08-21
feat(diagnostics): emit canonical lifecycle records from the runtime

pablonyx:
## What this is

Stage 0 of the lifecycle export pipe (ADR Observability, Overnight Execution Spec 2026-08-20, Lane B). It is the producer side of the canonical lifecycle record class, plus the runtime's real product operations instrumented against it.

**Nothing is exported by this PR.** Customer builds still compile the `absent` exporter (`proliferate-diagnostics-collector/src/export/absent.rs`). This lands the records; Stage 2 lands the pipe.

## Why it was needed

`proliferate-diagnostics-client` exposed exactly one emit API, `try_emit_detailed` (`lib.rs:148` before this change). There was no lifecycle emit API at all, so the only lifecycle producer in the repository was the Tauri process supervisor (`apps/desktop/src-tauri/src/diagnostics_collector/producer/lifecycle.rs`). Everything AnyHarness reported was a `tracing` call that the diagnostics layer turns into a `detailed` record, and `detailed` never leaves a customer machine. A lifecycle-only export would therefore have shipped process-supervision records and nothing product-facing.

## What is instrumented

| Operation | Call site | Terminal outcomes |
| --- | --- | --- |
| `anyharness.session.create` | `domains/sessions/service/create.rs` `create_session_with_persist` | `succeeded`; `rejected` for every refusal; `failed` only for `CreateSessionError::Internal` |
| `anyharness.turn.execute` | `live/sessions/sink/turns.rs` (`begin_turn`, `ensure_open_turn`, `turn_ended`) and `sink/terminal.rs` `commit_staged_prompt_terminal` | `succeeded` / `cancelled` / `failed` from the committed `TerminalTurnOutcome`; `rejected` for a model refusal |
| `anyharness.agent.start` | `domains/sessions/runtime/startup.rs` `start_live_session` | `rejected` for a refused selection, `failed` for a defect, `cancelled` for runtime shutdown |

Each is an RAII guard: `started` on construction, exactly one terminal record, and `abandoned` from `Drop` if the operation unwinds or its future is cancelled. An operation that vanishes is counted rather than silently missing from a success-rate SLI.

The whole path is inert in a process with no installed diagnostics producer, which is every test and every non-Desktop run. The producer handle is published from `install_desktop_producer` into a `OnceLock` in the client crate, because `anyharness-lib` has no path to the binary's telemetry wiring.

## The privacy proof

`anyharness/crates/proliferate-diagnostics-client/src/producer/tests_lifecycle.rs` is the centerpiece. Twelve tests, and these are the load-bearing ones:

- **`a_serialized_lifecycle_record_has_no_free_text_field`** asserts on the serialized bytes that `"detailed"`, `"message"`, `"stream"`, and `"milestone"` are all absent. This is the structural reason a lifecycle-only export is a privacy control and not merely a volume control: `LifecycleDiagnosticInput` has no free-text field to fill in, and `DetailedDiagnosticV1` is the only payload in the protocol that can carry one.
- **`an_argument_outside_the_safe_field_list_is_dropped_and_a_listed_one_survives`** passes both `agent_kind` (safe-listed) and `prompt_text` (carrying an actual user prompt) and asserts only `agent_kind` survives. The negative control is inside the test: the listed name surviving the same call proves the closed list is doing the work.
- **`an_unlisted_classification_is_dropped_and_degrades_the_outcome`** passes a classification containing a filesystem path and asserts it is dropped and the `failed` terminal degrades to `abandoned` rather than the unbounded string riding out.
- **`a_secret_shaped_correlation_value_is_scrubbed_on_the_lifecycle_path`** proves the lifecycle path reuses the detailed path's sanitizer rather than opening a second unguarded door.
- **`every_admitted_lifecycle_record_passes_the_shared_validator`** proves every record the producer builds is already ingest-legal.

**Negative control run.** Neutralizing the safe-field check in `producer::record::prepare_lifecycle` (`allowed.contains(&argument.name)` replaced with `true`) makes exactly two tests fail and no others:

```
an_argument_outside_the_safe_field_list_is_dropped_and_a_listed_one_survives ... FAILED
  left: ["agent_kind", "prompt_text"]
 right: ["agent_kind"]
the_safe_field_list_is_per_operation ... FAILED
```

The filter was restored byte-identical afterward.

None of this replaces the collector ingest gate, which stays the single mechanical enforcement point. These are the second belt.

## Deliberate omissions

- **`anyharness.model.request` has no call site.** The runtime never constructs a provider inference request; the agent process does. The only runtime-owned outbound gateway call is `probe_gateway_models` (`GET {base}/v1/models`, `domains/agents/route_auth/gateway_probe.rs:30-45`), which is a model list, not a model request, and carries no single `model_id`. The catalog's own family boundary says the model-gateway operation belongs to whoever "constructs and owns the provider request." The operation is defined in the producer table with model-metadata support proven by unit test, so the call site can be added without touching the privacy layer, but fabricating one here would have produced a lie.
- **`subagent_wakes.rs` and the staged-rollback path are uninstrumented** so they cannot produce phantom records.
- **`Cargo.lock`** picks up a one-line `proliferate` version correction (0.4.20 to 0.4.22). Pre-existing drift on main: the release-prep commit bumped the crate version without regenerating the lock, so any `cargo` invocation rewrites it.

## Testing

- `cargo test -p proliferate-diagnostics-client --lib`: 220 passed, 0 failed, including all 16 new lifecycle tests.
- `cargo test -p anyharness-lib --lib`: 2235 passed, 1 failed. The one failure is `route_auth::probe_materialization::probe_materialization_tests::sweep_tests::the_sweep_removes_only_abandoned_and_old_scratch_roots` ("an unparseable old-mtime dir must be removed"), a local directory-mtime environment failure in a module this PR does not touch and which the four commits since my base do not touch either.
- `cargo check --workspace --all-targets`: clean.

## Not merged

Per the standing rule, this is pushed and ready for review only. Merging is Pablo's.


pablonyx (commented):
VERDICT: merge-safe

Reviewed as fresh reviewer of record. Read at head 28b4165, base merge-base 88370a6 (main is 1 commit ahead, README hero image only, no overlap).

What I verified

1. No free text can reach a lifecycle record, structurally rather than by convention.
- `LifecycleDiagnosticInput` (proliferate-diagnostics-client/src/lib.rs:~130) has no `message`, `stream`, or `milestone` field, so the free-text surface does not exist on this path.
- `build_lifecycle` (producer/record/lifecycle.rs:148) sets `detailed: None` and `privacy: Operational` unconditionally.
- The protocol backstops it: validation.rs:70 matches `(Lifecycle, None, Some(lifecycle))`, so a lifecycle-class record carrying a `detailed` payload is `InvalidShape` at ingest. That closes the one path where `otlp.rs` `body_text` could have emitted a message on an exported record.

2. Closed lists are enforced in one place, not at call sites. `prepare_lifecycle` (producer/record/lifecycle.rs:39-63) drops arguments outside `safe_fields(name)` and classifications outside the operation's list, and degrades a `failed` with no classification to `abandoned` rather than losing the record. `filter_value` additionally requires `Enum` values to pass `valid_name` (lowercase ASCII plus `. _ : -`, bounded), so the enum-shaped fields cannot carry an unbounded caller string even if a call site regressed.

3. Emitted values are bounded at their sources: `origin_label` is a closed match over `OriginEntrypoint`, `startup_strategy` is `&'static str`, `StopReason` is a closed enum with a bounded Display (`end_turn`, `max_tokens`, `max_turn_requests`, `refusal`, `cancelled`), `TerminalTurnOutcome::as_str` likewise. `every_emitted_field_is_in_the_producers_safe_list` and the two `every_*_error_maps_to_a_permitted_classification` tests make a drift a red test rather than a silent dimension loss.

4. Pairing: the guard is held across the whole use case in all three ops, `Drop` emits `abandoned`, and the turn terminal is emitted only after the durable commit (`sink/terminal.rs:240`), so a failed commit leaves the guard open for the actor retry instead of reporting an outcome the store never accepted.

5. The producer table lists `anyharness.model.request` with no emitter. Noted as the raised contradiction, not relitigated here.

Nothing exported by this PR on its own: it only produces records into the local collector. CI green including the workspace default-features run.

Note for the merge: mergeable currently reports UNKNOWN. The only main-only commit is 77b81f7a (assets/readme/hero.png), so there is no real conflict.

PR #2183, opened 2026-08-21
docs(observability): specify the three D3 product SLIs

pablonyx:
## Summary

Lane D, D3 (observability overnight, Overnight Execution Spec 2026-08-20).
Specifies the three product SLIs that depend on Lane B's lifecycle pipe:
session-create success (Rejected vs Failed), launch-selection validity by
bounded code, and time to first output. Per the hard rule for this program,
none of these is faked with a proxy metric — where the real source record
does not yet exist or is missing a needed field, that is stated as a cited
finding with a proposed fix, not worked around.

- `server/infra/observability/honeycomb/product-sli-queries.json`: checked-in
  draft Honeycomb query specification for all three, with exact source
  attribute names verified by reading the actual emitted record shape.
- `guides/operating/analytics/honeycomb.md` (+ README entry): the runbook
  stub, matching the existing PostHog/Sentry/Customer.io procedure format.

## Findings, verified against `origin/observability/lane-b-stage-0-lifecycle-records` (PR #2180, unmerged)

1. **Session-create success is buildable now.** `anyharness.session.create`'s
   terminal outcome (`succeeded | rejected | failed | abandoned`) is fully
   emitted by PR #2180. The only blockers are outside this PR's remit:
   `HONEYCOMB_READ_KEY` returns HTTP 401 `unknown API key` live (checked
   2026-08-21, this lane's own probe, not just citing the existing doc), and
   Lane B has not yet posted a Stage-1 dogfood-proof note (dataset + record
   shapes) to the Accounts and Wiring doc, so the query has never been
   applied against real data.
2. **Launch-selection validity is blocked on missing instrumentation, not
   data.** `session.launch_selection.validated` is still a
   `tracing::info!` call site that the diagnostics layer turns into a
   `detailed` record, never `lifecycle`. Customer builds export
   lifecycle-only by design, so a detailed record structurally cannot reach
   production Honeycomb data, ever, regardless of Stage 1/2 landing. PR
   #2180 does not promote this event (confirmed by reading its diff and its
   own PR body, which lists exactly three instrumented operations and omits
   this one with no mention either way). This needs a Rust follow-up
   (protocol-catalog addition + call-site change) before it can carry a real
   threshold.
3. **Time to first output is blocked on a missing duration field, not
   data.** `anyharness.turn.execute` is instrumented (Started/Terminal), but
   "Started to Terminal duration" needs a delta between two OTLP log rows
   sharing `proliferate.operation_id`. Read the actual emitted record shape
   directly (`otlp.rs`, `lifecycle.rs` on B's branch): there is no
   `duration_ms` or any numeric elapsed-time field on the Terminal record,
   and `turn.execute`'s only safe fields are `stop_reason` and
   `engine_initiated`. Honeycomb computes per-event, not cross-event, so this
   is not answerable with any query against the current record shape. Needs
   a producer-side `duration_ms` safe-field addition (Rust, Lane B's crate).

Both blockers are called out precisely rather than silently worked around,
per the standing instruction to raise a concrete contradiction instead of
changing scope.

## Test plan

- [x] `python3 scripts/check_max_lines.py` — passes
- [x] `python3 -m unittest scripts/test_check_docs.py && python3 scripts/check_docs.py` — passes (236 markdown files, doc-integrity clean)
- [x] `python3 -m json.tool server/infra/observability/honeycomb/product-sli-queries.json` — valid JSON
- [ ] CI green on this PR (pushed, not merged — merge is Pablo's per standing policy)

Docs/spec only; no runtime code changed.


pablonyx (commented):
VERDICT: merge-safe

Docs/spec-only PR (no runtime code). Verified its three central factual claims directly against origin/observability/lane-b-stage-0-lifecycle-records (PR #2180, the cited source), not just by trusting the write-up:

1. SLI #1 (session-create success, Rejected vs Failed) is genuinely buildable: anyharness/crates/anyharness-lib/src/domains/sessions/service/create_lifecycle.rs on that branch confirms the exact classification claimed in server/infra/observability/honeycomb/product-sli-queries.json -- only `CreateSessionError::Internal(_)` maps to `TerminalOutcomeV1::Failed` ("internal_error"), every other error variant (workspace_not_found, workspace_single_session, session_id_conflict, launch_options_unavailable, launch_value_unsupported, route_auth_refused, invalid_request) maps to Rejected, matching a test named exactly `only_an_internal_error_counts_as_a_failure`. Not a proxy metric.

2. SLI #2 (launch-selection validity) is correctly marked BLOCKED, not faked: confirmed `session.launch_selection.validated` in anyharness/crates/anyharness-lib/src/domains/sessions/service/create.rs is still a plain `tracing::info!` call (lines ~318 and ~523 on that branch), never routed through the lifecycle producer, so it would only ever materialize as a `detailed` record that customer builds never export. The JSON correctly labels this "cannot be set responsibly" and proposes the real fix (catalog promotion) instead of substituting the dogfood-only detailed record.

3. SLI #3 (time to first output) is correctly marked BLOCKED for both stated reasons: grepped the full lane-b tree for `anyharness.model.request` -- it appears only in the protocol catalog table (catalog.rs:71), the client-side safe-field spec table (lifecycle.rs:106), a validation branch, and one unit test's synthetic input. No actual runtime call site emits it anywhere in the runtime/actor code. This matches the PR's claim verbatim: "declared in the producer's table but has NO runtime emitter yet." The doc correctly reports this as [unverified]/blocked rather than substituting `turn.execute` (which also genuinely lacks a duration_ms field per the same source read) as a proxy.

No secrets in honeycomb.md -- API key values are never included, only key names and 401 error results from live checks. Thresholds and runbook stubs exist for all three (SLI #1 has a real threshold + runbook; #2 and #3 are explicitly "not set" with "Not actionable yet" stubs, which is the honest state given they're blocked).

No findings.

intent-billing (provisional): re-checked via `gh pr checks 2183` just now -- now passing (34/34 checks green, including intent-billing and intent-tests). The Stripe-key-expiry rerun succeeded.

PR #2185, opened 2026-08-21
feat(diagnostics): export the lifecycle class from customer collectors

pablonyx:
## What this changes

Stage 2 of the lifecycle export pipe. Stage 0 (#2180) makes the runtime emit canonical lifecycle records; this makes a **customer** collector able to export them, and makes it mechanically unable to export anything else.

Before this, the whole OTLP adapter was compiled out of customer builds by the `internal-dogfood-export` feature. The question the code answered was "does an export path exist". That question is now "what may this build export", and the answer is one compile-time constant.

## The privacy control

`anyharness/crates/proliferate-diagnostics-collector/src/export/policy.rs` holds all of it:

- A customer build carries `ExportPolicy::LifecycleOnly`. Only `record_class == lifecycle`, and only with every field `operational`.
- `internal-dogfood-export` carries `ExportPolicy::All`, everything except `secret`.

`detailed` is the only payload in the protocol that can carry free text, because `DetailedDiagnosticV1` owns `message`, `stream`, and `milestone`. Refusing the class refuses the entire free-text surface, which is why this is the privacy control and not a volume control. It is a `const` selected by `cfg` rather than a configuration value on purpose: there is no input, env var, flag, or server response that can relax it. Widening it requires recompiling with a different feature.

Two fences consult it, not one:

1. `ExporterHandle::offer` (`src/export/handle.rs`) runs on the ingest path with the class already in hand, so an inadmissible class never enters the queue.
2. `encode_batch` (`src/export/otlp.rs`) refuses it again before it can become a wire payload.

The second is redundant today by construction. It exists so a future call site that finds another way into the encoder still cannot smuggle a detailed record past the first check.

A refused class is **not** counted as a drop. The record was accepted, retained, and stays locally queryable, so nothing was lost. Counting it would make every customer collector report itself permanently lossy in `/v1/health`.

## The centerpiece tests

`a_customer_build_never_queues_a_detailed_record` (`src/export/handle.rs`) offers `QUEUE_RECORDS + 32` detailed records at a real `mpsc` queue and asserts the receiver is empty and `dropped_records == 0`. Negative control: delete the `admits_class` guard in `offer` and it fails with 512 queued records.

`the_customer_policy_refuses_every_detailed_record_in_the_fixture` (`src/export/otlp_tests.rs`) runs the golden contract fixture through the encoder under `LifecycleOnly` and asserts the refusal count equals the detailed count, that every encoded log carries `record_class = lifecycle`, and that **no detailed record's message text appears anywhere in the serialized payload bytes**. Negative control: switch the policy to `All` and it fails with 13 log records, 0 refusals, and the message on the wire.

The encoder-mechanics tests now name the policy they exercise (`encode_all`), so one default-features CI run covers both the detailed encoding a dogfood build performs and the refusal a customer build performs. Nothing is feature-gated away from default CI except the two handle tests whose whole subject is the compiled policy.

## The release gate had to move in the same PR

`release-desktop.yml` refused any bundle whose binaries contained `PROLIFERATE_DIAGNOSTICS_OTLP_ENDPOINT`. That literal is now in every customer build, so **the next desktop build would have failed** without this change. The gate is now a positive assertion rather than an absence proof:

- the packaged collector must answer `lifecycle_only` to `--print-export-policy`
- it must carry the `PROLIFERATE_EXPORT_POLICY=lifecycle_only` marker
- no packaged binary may carry `PROLIFERATE_EXPORT_POLICY=all`
- no packaged binary may carry `PROLIFERATE_DIAGNOSTICS_DEV_TAG`, which is dogfood-only and is the only thing that could attach a local identity

The self-report step compares runner arch to target arch and skips rather than fails if a future matrix entry cross-compiles, so restoring the commented-out Intel entry cannot break it.

Verified locally against real binaries, both directions:

| build | `--print-export-policy` | `=lifecycle_only` | `=all` | `DEV_TAG` |
| default features | `lifecycle_only` | present | absent | absent |
| `internal-dogfood-export` | `all` | absent | present | present |

The dogfood row is the negative control: a dogfood-built collector trips three of the four checks.

## This lands dark

Independently of the policy, no build exports anything until a destination is configured out of band. With no `PROLIFERATE_DIAGNOSTICS_OTLP_ENDPOINT` the handle is `Sink::Off`: no queue, no task, no HTTP client, no loss. Nothing in the shipped desktop sets that variable (`apps/desktop/src-tauri/src/diagnostics_collector/process.rs:205-216` passes `--capability-fd`, `--control-fd`, `--release`, `--environment` and nothing else), so a customer install exports nothing after this merges. Turning it on is a separate, deliberate change.

## Scope note for the reviewer

The frozen spec says Stage 2 lands "dark behind the server capability document". The document exists: it is `GET /meta`, parsed into `ServerCapabilityContract` (`apps/packages/product-client/src/lib/access/cloud/server-capabilities.ts`) and consumed by `useAppCapabilitiesFor`. It is not wired to the export, and wiring it is a separate slice rather than a line in this one, because the two live at different layers:

- the capability contract is fetched by the **renderer**, after startup, over HTTP
- the collector's destination is read **once, from the environment, by the native process** at spawn (`export/target.rs`), before any renderer exists

Binding them needs a new field on the `/meta` contract (a server change), plus a native path to receive it and reconfigure or restart the collector. That is a server behaviour change and a new restart semantic, neither of which belongs in an unattended overnight PR.

What this PR implements instead is dark by absence of a configured destination: no build exports until `PROLIFERATE_DIAGNOSTICS_OTLP_ENDPOINT` is set, and nothing in the shipped desktop sets it. That is mechanically stronger than a flag, since there is no value a server could send today that would turn export on. It is not the same thing as a server-side flip, and I am raising the difference rather than inventing a mechanism.

## Testing

- `cargo test -p proliferate-diagnostics-collector` (default): 46 unit tests pass, including the two policy fences.
- `cargo test -p proliferate-diagnostics-collector --features internal-dogfood-export`: all pass, including the dogfood integration suite.
- `python3 scripts/check_max_lines.py`, `scripts/lint_records.py`, `scripts/check_docs.py`, `scripts/check_anyharness_boundaries.py`: pass.

Based on #2180. **Do not merge**; merges are Pablo's.


pablonyx:
## Live proof, run against the built binaries rather than only the test suite

Two collectors, same source, same records, same receiver. The only difference is the compiled policy.

**Records ingested (identical in both runs):** a lifecycle `anyharness.turn.execute` terminal, and a detailed `renderer.console.error` whose message is the sentinel `LANEB-STAGE2-SENTINEL-MUST-NEVER-REACH-HONEYCOMB`. Both were accepted and both were retained: `GET /v1/records` on the customer collector returns the detailed record with the sentinel intact, so nothing was lost locally.

| | customer build (default features) | dogfood build (`internal-dogfood-export`) |
| --- | --- | --- |
| `--print-export-policy` | `lifecycle_only` | `all` |
| record classes on the wire | `["lifecycle"]` | `["detailed", "lifecycle"]` |
| sentinel occurrences in the captured request bytes | **0** | 1 |
| `exporter.dropped_records` | 0 | 0 |
| `exporter.state` | `ready` | `ready` |

The dogfood column is the negative control: the same sentinel does reach the wire when the policy is widened, so the zero in the customer column is the fence working and not the record failing to arrive for some unrelated reason.

`dropped_records` staying at 0 on the customer build is also deliberate. A refused class is not a loss, because the record was accepted, retained, and stays queryable locally; counting it would make every customer collector report itself permanently lossy.

## Against the real destination

A customer-policy binary was also pointed at the real internal Honeycomb destination. Its lifecycle records arrived in the `anyharness` dataset, which confirms the customer export path works end to end and not only against a local receiver.

pablonyx (commented):
VERDICT: merge-safe

Reviewed as fresh reviewer of record at head 5cf4173, base 2180's branch (correct).

1. The policy cannot be relaxed at runtime. `EXPORT_POLICY` (export/policy.rs:41-46) is a `const` selected by `#[cfg(not(feature = "internal-dogfood-export"))]`. There is no env read, config field, CLI flag, or wire field that reaches it, and `internal-dogfood-export` is non-default with no dependent crate that could unify it on (the collector is a leaf binary; only the workspace root registers it as a path dep). A customer build physically has no `ExportPolicy::All` value to construct.

2. Both fences are real and independent.
- handle.rs:98 `offer` refuses on class before the queue, using the class read off the validated record at state/ingest.rs:305. `a_customer_build_never_queues_a_detailed_record` offers `QUEUE_RECORDS + 32` detailed records and asserts empty receiver AND `dropped_records == 0`, so a filtered class does not make a customer collector report itself permanently lossy.
- otlp.rs:62 `policy.admits(&record)` in `encode_batch_with_policy` checks class and record privacy, and otlp.rs:235 holds each argument to the same bar.

3. The byte-level absence test is real, not decorative. `the_customer_policy_refuses_every_detailed_record_in_the_fixture` runs against the shared golden fixture (fixtures/contracts/rust-observability-v1/valid/records.json), asserts `detailed > 0` first, then asserts the serialized wire string does not contain the detailed record's `message`. The fixture's one detailed record is `renderer.console.error` with `privacy: customer_content` and a real message, so both the class fence and the privacy narrowing are exercised. `encode_batch_applies_the_compiled_policy` separately pins that the shipped entry point consults the compiled constant.

4. Release gate re-pointed in the same PR, and the old logic is deleted, not left alongside. The old check grepped all five packaged binaries for `PROLIFERATE_DIAGNOSTICS_OTLP_ENDPOINT`; that literal now exists in every collector build (target.rs:11), so leaving it would have failed every release. It is replaced by: marker present, `PROLIFERATE_EXPORT_POLICY=all` absent, `--print-export-policy` stdout equals `lifecycle_only`, and `PROLIFERATE_DIAGNOSTICS_DEV_TAG` absent from all five binaries. On a dogfood binary the first check fires immediately (`=lifecycle_only` is not compiled in), so the gate does fail closed. main.rs:47-48 prints the bare name to stdout and the marker to stderr, which is what keeps the literal linked in and matches the `$(...)` capture in the job. The step is `shell: bash` under Actions' `-eo pipefail`, so a binary that cannot run fails the bundle rather than skipping the check. The only live macOS matrix entry is aarch64-apple-darwin on macos-15, so the arch guard does not skip the self-report today. Windows never packages a real collector (build step builds it only on non-Windows and stages a placeholder), so there is no ungated leg.

5. Both cfg sides are tested in CI: the workspace `cargo nextest run --workspace` covers the `cfg(not(feature))` customer tests and the separate step covers the dogfood side.

Posture change worth being explicit about: customer builds now compile the queue, HTTP client, and destination parsing. Nothing exports without `PROLIFERATE_DIAGNOSTICS_OTLP_ENDPOINT`, which no shipped code path sets (`TargetConfiguration::Absent` gives `Sink::Off`, no queue, no task, no client). That is the intended R-X2 posture, and the compile-time policy is what carries the privacy claim now.

PR #2186, opened 2026-08-21
feat(diagnostics): stamp the install id on exported records

pablonyx:
## The gap

An exported record could say which producer boot and which release it came from, but nothing that survived a restart. So "40 launch failures" could equally be forty installs or one install failing forty times, and no field told them apart. That is the difference between a launch-blocking bug and one broken machine.

## What this adds

`proliferate.install_id`, a resource attribute stamped by the collector onto every exported record.

It is deliberately **not** a wire-protocol field. The value arrives on the collector's process seam as `--install-id`, alongside the existing `--release` and `--environment`, from the host that owns the identity. No producer can set it, spoof it, or omit it, and every record from one install carries the same value whatever producer emitted it or how many times that producer restarted.

The desktop passes the identity it already owns: the `desktop_install_id` file behind the renderer's existing `get_desktop_install_id` command (`apps/desktop/src-tauri/src/commands/desktop_identity.rs`). One value now serves both, so an exported record and an app event describe the same install rather than two unrelated identities. It is a locally generated UUID with no account, machine, or user identity in it, and it is pseudonymous by construction.

## Absent rather than invented, at every step

- A host with no identity to give passes nothing, and the attribute does not appear at all. A consumer can distinguish "unknown install" from "install whose id is the empty string".
- The collector **refuses** an empty, oversized, or unprintable `--install-id` at startup rather than sanitizing it. A silently rewritten identity is worse than no identity, and the value is stamped on every exported record, so an unbounded or control-character value would put an unbounded, unescaped field on the wire.
- Because startup refuses it, the desktop launcher drops an unusable value instead of passing it through. Passing it would trade a missing attribute for no diagnostics at all. An unreadable identity file is an observability degradation, never a launch failure.

## Tests

| test | what it proves |
| --- | --- |
| `an_install_id_crosses_the_process_seam_to_the_real_collector` | the real collector binary accepts `--install-id` from the real launcher and reaches `Ready` |
| `an_unusable_install_id_would_cost_the_collector_so_discover_drops_it` | negative control: the real binary refuses an out-of-range id, which is why `discover` filters |
| `accepted_records_reach_the_destination_as_conformant_otlp_with_its_headers` | the attribute is on **every** resource stream of the real OTLP payload at a real receiver, across several producer boots |
| `install_id_tests` (3 cases) | core startup accepts absent and bounded-printable, refuses empty, 129-byte, spaced, and newline-bearing |
| `the_install_id_is_stamped_as_a_resource_attribute_when_the_host_supplies_one`, `an_absent_install_id_omits_the_attribute_entirely` | the encoder's two branches |

## Testing

- `cargo test -p proliferate-diagnostics-collector` (default): 52 unit tests pass.
- `cargo test -p proliferate-diagnostics-collector --features internal-dogfood-export --test otlp_dogfood`: 4 pass, including the wire assertion above.
- `cargo test -p proliferate --lib diagnostics_collector::process`: 12 pass, including both new seam tests.
- `python3 scripts/check_max_lines.py`, `scripts/lint_records.py`, `scripts/check_docs.py`: pass.

Based on #2185, which is based on #2180. **Do not merge**; merges are Pablo's.

### DIFF
--- a/anyharness/crates/proliferate-diagnostics-collector/README.md
+++ b/anyharness/crates/proliferate-diagnostics-collector/README.md
@@ -16,8 +16,21 @@ proliferate-diagnostics-collector \
   --capability-fd <inherited-fd> \
   [--control-fd <inherited-fd>] \
   [--release <non-secret-release>] \
-  [--environment <non-secret-environment>]
+  [--environment <non-secret-environment>] \
+  [--install-id <non-secret-install-identity>]
 ```
+
+`--install-id` is the stable identity of the installation. The collector never
+derives, generates, or persists one; the host that owns the identity passes it
+in, which is why no producer can set, spoof, or omit it. It is stamped on every
+exported record as the `proliferate.install_id` resource attribute, so it is
+the field that turns per-record counts into per-install counts. A present value
+must be 1 to 128 printable ASCII bytes and startup refuses anything else rather
+than sanitizing it. Omitting it means the attribute is absent, not empty.
+
+`--print-export-policy` prints this binary's compiled export policy and exits
+zero. The desktop release job runs the packaged collector with it and requires
+`lifecycle_only`.
 
 The capability descriptor contains one newline-terminated, visible-ASCII
 bearer capability and is closed after startup. The raw capability is not an
@@ -151,12 +164,13 @@ resource attribute.
 Accepted records are offered to a bounded queue as they are accepted and
 converted to OTLP/HTTP JSON logs: one resource per producer boot
 (`service.name`, `service.version`, `service.instance.id`,
-`deployment.environment.name`, and `dev.user` when a developer tag is
-configured), one scope per admitted schema version, a detailed message or the
-stable record name as the body, and every remaining contract field as a
-`proliferate.*` attribute. Typed arguments become `proliferate.argument.<name>`.
-A record classified `secret` is refused rather than encoded, which is a second
-fence behind ingest admission rather than a new privacy path.
+`deployment.environment.name`, `proliferate.install_id` when the host supplied
+one, and `dev.user` when a developer tag is configured), one scope per admitted
+schema version, a detailed message or the stable record name as the body, and
+every remaining contract field as a `proliferate.*` attribute. Typed arguments
+become `proliferate.argument.<name>`. A record the compiled export policy does
+not admit is refused rather than encoded, which is a second fence behind ingest
+admission rather than a new privacy path.
 
 The export bounds are internal-only and independent of the runtime caps above:
 a 512-record queue, a 128-record or 512 KiB batch, a 250 ms batch linger, a
--- a/anyharness/crates/proliferate-diagnostics-collector/src/config.rs
+++ b/anyharness/crates/proliferate-diagnostics-collector/src/config.rs
@@ -81,6 +81,15 @@ pub struct CollectorConfig {
     pub capability: Vec<u8>,
     pub release: String,
     pub environment: String,
+    /// Stable identity of this installation, stamped onto exported records as
+    /// the `proliferate.install_id` resource attribute.
+    ///
+    /// It is not a wire-protocol field and no producer can set it. The host
+    /// that owns the identity passes it in, so a record cannot claim to come
+    /// from an install other than the one whose collector accepted it.
+    /// `None` means the host had no identity to give, and the attribute is
+    /// then absent rather than empty or invented.
+    pub install_id: Option<String>,
     pub runtime_limits: RuntimeLimits,
     pub auto_ready: bool,
 }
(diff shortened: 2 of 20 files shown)


pablonyx:
Rebased onto `main` (`d0dac1a0d4`) this morning, which surfaced a real break this PR could not have hit before: main's Windows work added `diagnostics_collector/supervisor_unsupported.rs`, selected by `#[cfg(not(unix))]`, whose `new` still took four arguments. The install-id parameter had only landed on the unix supervisor, so the shared call site at `apps/desktop/src-tauri/src/lib.rs:210` failed to compile on the Windows leg with E0061. Fixed in 2d2cafb170 by giving the non-unix constructor the same fifth parameter, accepted and ignored the way `_release` and `_environment` already are. That target ships a placeholder collector, so there is no process to stamp an install id onto there.

Worth recording for reviewers: because `proliferate-diagnostics-collector` does not compile for `x86_64-pc-windows-msvc` at all, the packaged-binary export-policy gate added in #2185 is macOS-only by necessity rather than by oversight. There is no real collector on the Windows leg to gate.

pablonyx (commented):
VERDICT: merge-safe

Reviewed as fresh reviewer of record at head 5dfafaa, base 2185's branch (correct).

1. The install id cannot be asserted by a producer. It is not a wire field; it arrives as `--install-id` on the collector's own argv (process.rs:235), flows through `CollectorConfig.install_id` into `ExporterHandle::from_environment(install_id)` and then the worker, and is stamped only as a resource attribute in `resource_attributes` (otlp.rs:116-127). No ingest path can set it, so a record cannot claim an install other than the collector that accepted it.

2. It is bounded before it can reach the wire. state/mod.rs:141 refuses a present id that is empty, over 128 bytes, or not all `is_ascii_graphic`, with `an_empty_oversized_or_unprintable_install_id_is_refused` covering all four shapes. The desktop side drops an unusable id (`install_id.filter(|v| validate_label(v).is_ok())`, process.rs:177) rather than passing one through and losing diagnostics entirely.

3. Absent stays absent. `an_absent_install_id_omits_the_attribute_entirely` asserts the attribute is not emitted at all rather than emitted empty, so a consumer can distinguish unknown install from empty-string install. The value is the existing locally generated UUID from `desktop_identity`, with no account, machine, or user identity in it, and it is the same one the renderer already reads.

4. No privacy regression on top of 2185: the policy fences are untouched, and the install id rides the resource, not the record, so it appears only on payloads the customer policy already admitted.

5. Signature hygiene: `supervisor_unsupported.rs` takes the same fifth parameter so both supervisors present one signature to lib.rs. The dogfood integration test is `#![cfg(feature = "internal-dogfood-export")]`, so it runs only in the dogfood CI step and cannot force the feature on anywhere else.

Stack is linear and clean: 88370a6 -> 2180 -> 2185 -> 2186, no unrelated files, no stray working-tree content.

pablonyx (commented):
VERDICT: blocked

Reviewed independently at current full head `77ae38a28391ca16526405a0892a785ea717c221` against the lifecycle-export spec, PR body, merged parent #2185, current repository standards, and the prior review at `5dfafaa4c00de425ca9f8586ba394338430655be`.

**B-2186-01 (Blocker) — Desktop and collector accept different install-id domains.**

At `apps/desktop/src-tauri/src/diagnostics_collector/process.rs:179`, `discover` retains the identity when `validate_label(value).is_ok()`. That validator (`process/helpers.rs:23-28`) accepts ASCII space:

```rust
.all(|byte| byte.is_ascii_graphic() || byte == b' ')
```

Collector startup rejects the same value at `anyharness/crates/proliferate-diagnostics-collector/src/state/mod.rs:144-150`, because it requires every byte to be `is_ascii_graphic()`. Therefore a persisted `desktop_install_id` such as `"has space"` is forwarded as `--install-id` and prevents the collector from starting. Expected behavior, per the PR body and comments at `process.rs:175-178`: an unusable host identity is dropped so diagnostics launch without the attribute. Actual behavior: this input trades a missing attribute for no collector.

The named negative-control test does not exercise the claimed discovery path: `process_tests.rs:291-304` constructs `CollectorProcessLauncher` directly, checks only a 129-byte value, and never calls `discover`; it therefore misses the domain mismatch.

This is not rebase drift. `git range-diff` marks all three slice commits equal, and the stable whole-slice patch id is identical (`b8c688c09df9c95d4ced6e24d11e09d0a66def03`) for old `5cf41739db4c199a4e510843f7ab8720bb8dc2db..5dfafaa4c00de425ca9f8586ba394338430655be` and current `b5c3ccf687171d5603e1d4543359d81a2b815b89..77ae38a28391ca16526405a0892a785ea717c221`. The earlier merge-safe review missed this existing blocker when it said the Desktop filter drops every unusable id.

The remaining reviewed seams are sound: the producer protocol/client contain no install-id field; the collector alone stamps the host-supplied bounded value as a resource attribute after the compile-time export-policy fences; the Desktop source is the existing locally generated UUID and introduces no account/user/machine identity; #2185's lifecycle-only customer policy and packaged macOS release assertions are untouched; and the non-Unix placeholder constructor now matches the shared call site. Required checks are green at this head, while the broad Cargo job was still running at review time, but CI cannot establish the missing space-bearing discovery case.

PR #2181, opened 2026-08-21
fix(server): sign-in success-rate SLI end to end

pablonyx:
## Summary

Continuation of Lane D (observability overnight, D1) after the prior agent was
stopped mid-work. Adopted its uncommitted worktree changes, independently
re-verified every live claim, and committed.

- `server/proliferate/auth/sign_in_observability.py`: single place emitting
  `auth.sign_in.outcome` structured log lines (`auth_sign_in_outcome`,
  `auth_sign_in_surface`, `auth_sign_in_failure_code` on failure only) — no
  auth code, PKCE verifier, token, or email logged.
- Wired into `/auth/web/token`, `/auth/mobile/token` (`identity/api.py`) and
  `/auth/desktop/token` (`desktop/api.py`): success and `ProliferateError`
  failure paths both log now, where previously neither did.
- `guides/operating/production-alerts.md`: new "Sign-in success rate (SLI)"
  section documenting the CloudWatch metric filters, the Grafana rule, and
  what to do when it fires.
- `scripts/ops/grafana-sli-alerts.mjs` (+ tests): check/apply/verify tooling
  for the rule, scoped to its own `sli-alerts` rule group so it never touches
  the `production-alerts` group, contact point, or notification policy that
  `grafana-alerting.mjs` / `grafana-rebuild-bootstrap.mjs` (#2178) manage.
- `server/infra/observability/grafana/sli-alerts.json`: checksum-verified
  record of the live rule (uid `ffvtx33lbo5c0e`, new `proliferate-ops-rebuild`
  workspace).

**Builds on `obs/lane-a3-grafana-new-workspace` (#2178, unmerged)**: depends
on that PR's CloudWatch data source (`cfvtvusw9bd34d`) and unified-alerting
enablement on the new workspace to have already happened live tonight (they
have — verified below), but this PR touches none of #2178's files and can
land independently of it merging.

## Independent verification (not just re-stating the prior agent's claims)

- `aws logs describe-metric-filters --log-group-name /ecs/proliferate-prod`:
  both `sign-in-success` and `sign-in-failure` filters exist live, same shape
  as the two pre-existing filters in that log group (neither pair is
  Terraform-managed — consistent with existing practice).
- `aws cloudwatch get-metric-statistics` for `SignInSuccessCount` and
  `SignInFailureCount`: real `Sum=1.0` datapoint for each at
  `2026-08-21T00:30:00-07:00` — the pipe produces a real number, not just a
  configured-but-silent rule.
[...]

PR #2189, opened 2026-08-21
fix(dev): restore the runtime binary after the workspace build clobbers it

### LINKED ISSUE #2188: fix(dev): make dev is unrecoverable because the workspace build clobbers the runtime binary with the tauri sidecar placeholder
`make dev` is unrecoverable on a warm worktree: the workspace build clobbers `target/debug/anyharness` with the Tauri sidecar placeholder, and `build-rust` then copies the placeholder into `target/runtime-local`.

Introduced by #2134 (merged as `fe6780e006`), which changed `build-rust` to `cargo build --workspace` followed by a copy of `target/debug/anyharness` into `$(DEV_ANYHARNESS_TARGET_DIR)`.

## Symptom

```
$ make dev PROFILE=<name>
...
nothing to build
AnyHarness runtime binary is a sidecar placeholder: target/runtime-local/debug/anyharness
Run: make build-rust
make[1]: *** [dev-artifacts-ready] Error 1
make: *** [dev] Error 2
```

`make build-rust` exits 0 and changes nothing, so following the instruction loops forever.

## Reproduction, with the control

In a worktree where the desktop crate has been built at least once:

```
$ cargo build -p anyharness   && ls -la target/debug/anyharness
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 12.45s
    138493688 bytes                                   <- real runtime binary

$ cargo build --workspace     && ls -la target/debug/anyharness
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 10.51s
    106 bytes                                         <- sidecar placeholder
```

The 106-byte file is the placeholder script:

```
#!/bin/sh
printf '%s\n' 'AnyHarness sidecar is not available for target aarch64-apple-darwin.' >&2
exit 1
```

Byte-identical to `apps/desktop/src-tauri/binaries/anyharness-aarch64-apple-darwin`, and it carries that file's mtime, so it arrives by a time-preserving copy from the desktop build rather than by a link.

## Why it does not self-heal

`build-rust`'s copy is guarded by `cmp -s "$target_dir/debug/anyharness" "$(DEV_ANYHARNESS_TARGET_DIR)/debug/anyharness" || cp -f ...`. Once both sides are the placeholder the two files are identical, `cmp` succeeds, and the `cp` never runs. Every subsequent `make build-rust` re-clobbers `target/debug` and re-confirms the match, so `target/runtime-local` keeps its bad binary indefinitely.

This also defeats the mitigation `DEV_ANYHARNESS_TARGET_DIR` exists for. Keeping the good runtime out of `target/debug` only helps while nothing copies `target/debug` back over it, and `build-rust` now does exactly that.

## Why review missed it

The #2134 review checked this hazard and concluded from reading `apps/desktop/src-tauri/build.rs` that it "no longer applies" because `copy_executable` copies *from* `target/debug` *to* `binaries/`. That direction is real but not the whole behaviour: the observed result is that `target/debug/anyharness` ends up holding the placeholder after a workspace build. The claim was reasoned from source rather than executed, and the PR shipped with an explicit note that no real-box run had been done.

Cold builds can mask it. A first `make dev` in a fresh worktree may link the runtime after the desktop step and copy a good binary, which is why the failure shows up on the second run rather than the first.

## Fix shape

Re-establish the real binary after the workspace build and before the copy, for example `cargo build --workspace` followed by `cargo build -p anyharness` (0.36s warm, since cargo only re-uplifts from `target/debug/deps`) and then the existing copy. Alternatively copy from a source the desktop build does not write.

## Workaround

```
cargo build -p anyharness
cp -f target/debug/anyharness target/runtime-local/debug/anyharness
SKIP_RUST=1 ANYHARNESS_DEV_RUNTIME_BIN=$PWD/target/runtime-local/debug/anyharness make dev PROFILE=<name>
```

pablonyx:
Fixes #2188. `make dev PROFILE=<name>` is currently unrecoverable on any worktree where the desktop crate has been built, which is the command #2134 made canonical.

## What breaks

`build-rust` runs `cargo build --workspace` and then copies `target/debug/anyharness` into `$(DEV_ANYHARNESS_TARGET_DIR)`. The workspace build includes the desktop crate, and after it runs, `target/debug/anyharness` is the 106-byte Tauri sidecar placeholder rather than the runtime:

```
$ cargo build -p anyharness && ls -la target/debug/anyharness
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 12.45s
    138493688 bytes

$ cargo build --workspace   && ls -la target/debug/anyharness
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 10.51s
    106 bytes
```

The copy then propagates the placeholder. Because the copy is guarded by `cmp -s ... || cp -f ...`, once both sides hold the placeholder they compare equal and the copy never runs again, so `target/runtime-local` keeps the bad binary. `dev-artifacts-ready` rejects it with `Run: make build-rust`, and running that re-clobbers `target/debug` and re-confirms the match. The instruction the tool gives you cannot fix the problem it reports.

## The fix

One line: re-run `cargo build -p anyharness` after the workspace build, before the copy reads the path. Cargo re-uplifts the real binary from `target/debug/deps`, which costs 0.36s warm because nothing recompiles.

## Proof

Both sides poisoned to the placeholder, then `make build-rust`, on this branch and on `main`:

| | poisoned state | after `make build-rust` | `dev-artifacts-ready` |
|---|---|---|---|
| this branch | `debug=106  runtime-local=106` | `debug=138493688  runtime-local=138493688` | passes, and `./target/runtime-local/debug/anyharness --version` prints `anyharness 0.1.0` |
| `main` (control) | `debug=106  runtime-local=106` | `debug=106  runtime-local=106` | fails: `Run: make build-rust` |

Both invocations exit 0, which is what makes the shipped behaviour hard to notice: the build reports success while leaving an unusable binary in place.

## Scope

This addresses the symptom where it breaks the dev loop. The root cause is that the desktop build writes a placeholder over a path that is also the runtime's build output, which is worth fixing separately rather than in a change that needs to land quickly.

Found by running a real profile launch, which is the check #2134 shipped without.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1615,6 +1615,7 @@ build-rust:
 		echo "SKIP_RUST set — skipping cargo builds (runtime: $${ANYHARNESS_DEV_RUNTIME_BIN:-<unset>})"; \
 	else \
 		$(CARGO) build --workspace || exit 1; \
+		$(CARGO) build -p anyharness || exit 1; \
 		target_dir=$$($(CARGO) metadata --format-version 1 --no-deps --offline 2>/dev/null \
 			| node -e 'let b="";process.stdin.on("data",d=>b+=d).on("end",()=>{try{process.stdout.write(JSON.parse(b).target_directory)}catch{}})'); \
 		target_dir="$${target_dir:-target}"; \


pablonyx:
## Review of record: MERGE-CLEAN

Fresh reviewer (not the implementer), verdict **MERGE-CLEAN**, no blockers.

### What was verified

**Root cause holds.** The placeholder that lands at `target/debug/anyharness` is byte-identical to `apps/desktop/src-tauri/binaries/anyharness-{triple}`, so it arrives through Tauri's `externalBin` resource staging for the desktop crate during `cargo build --workspace`, not from anything in the `anyharness` package's own build graph.

**The fix cannot be re-clobbered by itself.** `anyharness/crates/anyharness` declares the only `[[bin]] name = "anyharness"` in the workspace and has no path dependency on `apps/desktop/src-tauri`, so `cargo build -p anyharness` is structurally incapable of invoking the desktop crate's build script under any Cargo scheduling. `make -j` is not a hazard either: both cargo calls live in one recipe body of one target, so `-j` cannot interleave them.

**Ordering and environment.** No `.cargo/config.toml` and no `[build] target-dir` in the repo; `CARGO` is a fixed path with no baked-in `--target`. Both cargo invocations and the later `cargo metadata` run under identical ambient environment inside the same recipe, so any `CARGO_TARGET_DIR` override is honoured consistently by all three. The extracted recipe body passes `bash -n` with the new line in place.

**Cost.** Cargo's uplift from `target/debug/deps/anyharness-<hash>` runs whenever a matching target is processed, regardless of recompilation, and nothing in the recipe changes the package's fingerprint inputs between the two invocations, so the second build is always a cache-hit uplift.

**`SKIP_RUST` untouched.** The added line sits strictly inside the `else` block; the skip branch is unchanged.

**Self-healing.** `cargo tauri dev` does re-clobber `target/debug/anyharness` later in a `make dev` session, but by then `dev-artifacts-ready` has passed and the app reads the copy under `$(DEV_ANYHARNESS_TARGET_DIR)`. That later reclobber only poisons the next `make build-rust`, which now heals itself every invocation because the added line runs unconditionally before every copy.

### Alternatives considered and rejected

Reaching into `target/debug/deps/anyharness-<hash>` directly (fragile hash resolution); reordering the two cargo calls (`--workspace` last would re-clobber); `--exclude proliferate` from the workspace build (plausibly the better long-term root-cause fix, but it changes what the "build everything" gate covers, which is the larger change this PR explicitly defers).

### Live confirmation on a real machine, after the review

Independent of the review, the fix was exercised end to end on this branch's worktree:

| step | `target/debug/anyharness` | `target/runtime-local/debug/anyharness` |
| --- | --- | --- |
| after `cargo build -p proliferate` | 106 | 138493688 |
| after patched `make build-rust` (0.18s) | 138493688 | 138493688 |

`make dev PROFILE=... HEADLESS=0` then brought up runtime, API, hosted web and the Tauri desktop window with no errors.

### Should-fix, out of scope for this diff

`.github/workflows/release-e2e.yml:220-221` still says `make build-rust` "builds the workspace once", which is now one invocation short. Worth a follow-up one-line comment fix, not blocking.

### CI

37 SUCCESS, 1 SKIPPED, 1 FAILURE. The failure is `intent-billing (provisional)`, which is `continue-on-error: true` in `intent-tests.yml:113` and therefore non-blocking; the run's own conclusion is `success`.

It fails in `scripts/stripe-setup-test-mode.mjs` on `stripe products list` with `The API key for the default profile has expired`, and it failed again on rerun. It is not caused by this diff, which is one line of Make and cannot reach Stripe. The evidence is that the same job is currently red on **six other open PRs** (#2190, #2186, #2185, #2183, #2182, #2180) and green on three (#2187, #2184, #2181), interleaved by time: pass 08:21, fail 08:53 and 08:54, pass 09:08, fail 09:16, 09:17, 09:23. That intermittency against a single shared Stripe test account looks like concurrent jobs clobbering each other's `stripe login` default profile rather than a real expiry.

Correction to an earlier version of this comment: I first wrote that main passed this check at 08:58Z and 09:13Z. That was wrong. I had read the run-level conclusion, not the job; `intent-billing` does not run on main at all. The cross-PR comparison above is the actual evidence.

Worth noting separately: the job is documented to skip itself cleanly when `STRIPE_TEST_SECRET_KEY` is absent, but an expired or invalid key produces a hard failure instead of a skip. That gap is a candidate follow-up, unrelated to this PR.

PR #2200, opened 2026-08-21
fix(observability): mirror live SNS route for Slack fanout

pablonyx:
## Summary

Repairs the Grafana Slack fan-out after the first authorized live apply failed closed before mutation. The live AMG route tree has one autogenerated-matcher SNS delivery child rather than a matcherless child. This change mirrors that exact match shape into the preceding Slack sibling with continue enabled, while preserving the root/default SNS receiver, its complete configuration, the SNS child including nested routes, and every other pre-existing child route.

No live provider configuration was mutated and no Grafana Slack test was sent during this repair. Live apply, verify, and delivery test remain paused until independent review and merge.

## Testing

- Narrow ops-tool unit coverage: 27 passing.
- The fake route tree now matches the observed live AMG autogenerated wrapper and nested SNS receiver route.
- Added fail-closed coverage for matcher drift and ambiguous SNS delivery children; existing coverage proves exact SNS receiver/config and full non-Slack route-tree preservation.
- Offline rebuild artifact check passed.
- Documentation integrity and max-lines checks passed under Python 3.12.

## Observability

This corrects observability delivery configuration only. It adds no product signal and does not alter the existing SNS/email default route.

## Readiness

- [x] I followed the pull-request procedure for scope, title, labels, body, and readiness.

## Verification

- node --check scripts/ops/grafana-rebuild-bootstrap.mjs
- node scripts/ops/grafana-rebuild-bootstrap.mjs check
- node --test scripts/ops/grafana-rebuild-bootstrap.test.mjs (27 passing)
- /opt/homebrew/bin/python3.12 scripts/check_docs.py
- /opt/homebrew/bin/python3.12 scripts/check_max_lines.py
- git diff --check

## Frozen review scope

Review only the route-topology repair at head d98c549f95e8c20a2ae4f582261bcbae09e8f0ba: unique SNS delivery-child discovery, Slack match-shape mirroring, preservation/fail-closed assertions, the exact live-topology tests, and the matching operator-guide correction. Live apply, live test delivery, provider credentials, merge, Sentry/PostHog/Honeycomb, and any unrelated Grafana behavior are out of scope.

PR #2202, opened 2026-08-21
fix(observability): group Grafana Slack with default SNS

pablonyx:
## Summary

Fixes the Grafana Slack delivery model after the authorized live apply exposed AMG receiver normalization that the prior fake did not model. Grafana creates a reserved generated route child for every standalone receiver and will not retain an authored sibling that reuses that reserved matcher. The safe fan-out is one contact point with two integrations: the exact existing SNS integration plus exactly one Slack integration inside grafana-default-sns. The root/default policy and complete route tree remain unchanged.

The live workspace currently has one nonfunctional partial standalone Slack contact/receiver, UID dfvuf540l7ym8d, created by the failed #2200 apply. This repair writes and verifies the combined contact first, then deletes that standalone integration. If interrupted between those phases, rerun verifies the pinned legacy identity, updates/verifies the combined integration, and completes the deletion without duplication. No Slack test was sent during repair development.

## Review repair mapping

- GRAFANA-COMBINED-001: production-alerts-rebuild.json now pins legacyUid dfvuf540l7ym8d. Offline validation rejects any other value. Before combined-contact create/update or legacy deletion, both the provisioning contact UID and Alertmanager receiver-config UID must equal that exact value. Missing, unexpected, ambiguous, or drifted legacy state fails before mutation. Once migration is complete, reapply is a verified no-op rather than an unproven update.
- Exact negative tests use dfvuf540l7ym8e as the wrong UID independently in the provisioning contact, receiver config, and both readbacks. They prove zero create/update/delete/policy-write/test calls. A separate interrupted-migration test proves the same gate blocks update and deletion.

## Testing

- Narrow ops-tool unit coverage: 32 passing.
- Provider-normalization-realistic fake groups multiple integrations under the existing receiver and creates/removes the generated child for a standalone receiver.
- Tests prove exact policy-tree and SNS integration preservation, target-only test payload shape, create/update idempotency, interrupted migration retry, delete-after-verify ordering, exact legacy UID gating, and fail-closed behavior for duplicate/drift/loss.
- Offline rebuild artifact, documentation integrity, max-lines, syntax, and diff checks pass.

## Observability

This is observability delivery configuration only. The Slack integration is additive inside the default SNS contact point; the established SNS/email integration remains exact and the notification policy is unchanged.

## Readiness

- [x] I followed the pull-request procedure for scope, title, labels, body, and readiness.

## Verification

- node --check scripts/ops/grafana-rebuild-bootstrap.mjs
- node scripts/ops/grafana-rebuild-bootstrap.mjs check
- node --test scripts/ops/grafana-rebuild-bootstrap.test.mjs (32 passing)
- /opt/homebrew/bin/python3.12 scripts/check_docs.py
- /opt/homebrew/bin/python3.12 scripts/check_max_lines.py
- git diff --check

## Frozen re-review scope

Review only GRAFANA-COMBINED-001 at head 14841eea46b1a0f06d298aa300eed2b74a1de634: the checked-in legacyUid pin, offline validation, both live UID readback gates before mutation/deletion, converged no-op behavior, exact negative tests, and the corresponding operator-guide wording. The previously reviewed combined-contact architecture, live retry/test, merge, credentials, Sentry, PostHog, Honeycomb, old Grafana cutover, and unrelated alert behavior are out of scope.


pablonyx (commented):
VERDICT: merge-safe

Re-reviewed the complete current diff and the focused repair at exact head `14841eea46b1a0f06d298aa300eed2b74a1de634` against the frozen combined-contact migration contract. No Blocker, Decision, or Follow-up findings remain.

`GRAFANA-COMBINED-001` is closed:
- `production-alerts-rebuild.json` pins the exact live partial standalone UID `dfvuf540l7ym8d`, and offline overlay validation rejects any other value.
- Before combined-contact create/update or legacy deletion, the provisioning contact UID and the Alertmanager receiver-config UID must independently equal that pin; missing, mismatched, duplicate, or drifted legacy state fails before mutation.
- Negative tests vary the provisioning and receiver UID readbacks independently and together, proving zero create/update/delete, policy-write, or receiver-test calls; the interrupted-migration case proves the same gate on retry.
- A verified final state with the combined Slack integration and no legacy partial returns `present` without updating the contact, while an absent partial plus absent combined contact fails closed rather than inventing state.

Full-contract reassessment:
- The checked-in root/default notification policy is deep-compared before and after migration, so the complete route tree remains unchanged.
- The existing `grafana-default-sns` receiver retains exactly one SNS integration with pinned UID `bfvtw9if8c3cwd`, identity, resolve behavior, and exact settings; every unowned receiver is preserved.
- The combined contact is read back and verified before the exact legacy UID is deleted, then the final contact/receiver/policy state is verified again; interrupted deletion safely converges on rerun.
- `slack-test` first verifies the converged combined state, then submits a receiver-test body containing exactly one Slack integration and no SNS/email integration.
- The normalized fake models provisioning entries separately from the grouped Alertmanager receiver, secure-field readback, generated standalone receiver/route creation, and provider-driven cleanup on delete.

Independent exact-head proof:
- `node --check scripts/ops/grafana-rebuild-bootstrap.mjs` — passed.
- `node scripts/ops/grafana-rebuild-bootstrap.mjs check` — passed offline.
- `node --test scripts/ops/grafana-rebuild-bootstrap.test.mjs` — 32/32 passed.
- `git diff --check 0afba61c6a3b52ed0c00257d3f71ad25cb958a2b 14841eea46b1a0f06d298aa300eed2b74a1de634` — passed.
- The final current-head CI rollup has no pending or failed checks; the externally rerun server integration shard and regenerated `server-ci-ok` are green.

No live Grafana/Slack operation, provider mutation, delivery test, merge, or secret access was performed during review. Formal approval is unavailable because the authenticated CLI identity is the PR author, so this is the reviewer-of-record COMMENT verdict.

PR #2247, opened 2026-08-26
docs(specs): agent_auth canonical rewrite — outline shape, ai_gateway split, verified to code

pablonyx:
## What

Rewrites `specs/systems/agent_auth/` into the canonical spec, following the ruled five-section outline (rebased onto the unwrapped-prose main from #2248).

**README.md** — the ruled shape:
- **0 · Scope** — folder census, responsibilities, fences, the three-methods table, rules of the road.
- **1 · Cells** — four grid cells, each as Owns / Doors-with-named-consumers / Consumes.
- **2 · Data** — full DDL (six tables), the complete state.json v2 schema + riders + fixture pin, runtime persistent state, vocabularies + drift test.
- **3 · Flows** — five capability flows (change method · store material · configure launch · harness auth status · plan status), each a plain-English paragraph + a cell-level sequence diagram; cross-cell rules bolded inline in the flow they protect; launch-ladder pseudocode where decision logic earns it.
- **4 · Structure** — per cell: full annotated file tree (⇒ marks ruled moves), the full API with request/response bodies, every importable signature, every event, cell-local invariants.
- **Transitional tail** (deleted at convergence): one Delta-vs-prod table + the dependency-ordered build list, whose first item is the no-code fix from today's renderer trace (fund the org's billing subject — the $2 signup grant is blocked by a prior account's one-per-GitHub-identity claim; #2245 un-gating delivery is what surfaced it).

**Also:** `model-gateway.md` → `specs/systems/ai_gateway/README.md` (the ruled gateway split, spec level) · `deep-dive.md` folded and deleted (history in git) · `models.md` and every cross-reference repointed · `check_docs.py` green (211 files).

## Open rulings (live here while the spec is in draft, per the ruled convention)

1. **Headless credentials at v1** — rec: automations run as their creator; org default lands with the first team org. *(reflected in the spec body as ruled)*
2. **Org seat ownership** — rec: org-owned vault rows, shipped in two steps (v1 user-owned).
3. **Usage-probe cadence** — rec: 5 min while a session runs on the seat, 30 min idle, off when revoked; pane-open forces a sample.
4. **Env-var passthrough method** — rec: delete; the vault covers it. *(spec body assumes deletion)*
5. **ai_magic through the gateway** — rec: after per-run keys exist.

## Review

Spec-only. Pablo reviews and merges — this PR does not self-merge.

https://claude.ai/code/session_015XM9Un7yY8v36SDUisKxo3


pablonyx:
Iteration rounds complete (adversarial critique ×1, code-truth verification ×1, harness-reality verification ×1, design-pass reconciliation ×1 — 45 findings applied across four commits, docs checker green throughout).

**One more open ruling** (moved out of the ai_gateway spec body per the decisions convention):

6. **Per-run virtual keys + budget envelopes** — mint a key scoped to (org, run) at placement with the run's envelope as budget, deleted at terminal (rec — the only shape where the proxy hard-stops a runaway fan-out mid-run), vs keep per-(member, harness) keys and enforce envelopes at import time.

**Three live tests gate seats v1** (each needs a credential Pablo mints and can revoke):
- setup-token output format → confirms the capture rule
- one end-to-end seat session through the real adapter → confirms delivery (SDK-side `CLAUDE_CODE_OAUTH_TOKEN` support already confirmed by code)
- a second interactive login under a per-seat config dir on macOS → confirms multi-seat keychain coexistence

PR #2254, opened 2026-08-27
feat(agent-auth): slice 1 — mint & run (seats v1)

pablonyx:
Slice 1 of the approved auth slice plan, implemented from the frozen delivery spec (`delivery/agent-auth/delivery-spec-slice-1-mint-and-run.md`) and the agent_auth system spec (§2 Data, §3 Flows 2–3, §4 cells 1–2). Adversarially reviewed; all review findings triaged and fixed (see below).

## Scope summary

A Claude Max login becomes a **seat** — a portable credential in the vault — and a session runs on it: the full mint → store → render → apply → launch chain for a single seat, product-grade.

- **Schema + constants**: `anthropic_subscription` vault kind, `seat` selection source kind, the pool selection shape (kind-counting radio), one alembic migration, vocab + registry mirrors.
- **Stores + server**: seat kind through vault CRUD and decrypt-for-render; selection write gate; `selection_rules.py` seat legality; `state_render.py` expands the pool row into seat sources in vault order (deduped by seat id); `seats.py` lifecycle glue.
- **Wire pin**: `fixtures/contracts/agent-auth-state/` gains the `seat` variant; Python asserts producing, Rust asserts consuming. `revision` stays on the wire this slice.
- **Runtime**: seat recipe (`CLAUDE_CODE_OAUTH_TOKEN` + per-seat `CLAUDE_CONFIG_DIR` + strip list + rerouting flags), `claude-config-<seat>/` home family, the `mint_seat` login-terminal variant running `claude setup-token` in an isolated dir — capture rule `^sk-ant-[A-Za-z0-9_-]{40,}$` (last non-empty visible line), completion on exit or 60s grace, token memory-only, buffer wiped on handoff and every error path, single-flight per harness (atomic).
- **Courier + client**: `uploadSeatToken()` (memory-only, one POST, no silent retry); the settings pane's single-seat subset — "Add a Claude.ai login", mint sheet (email + optional plan tier, default "Max seat N"), waiting-for-sign-in row, labeled seat row; SDK regen.
- **Verification**: the ordinary launch probe under the seat's home after apply; a failed verification leaves the seat saved, shown unverified.

## Acceptance gate (verbatim from the delivery spec — performed by Pablo, in the product)

> In settings: click **"Add a Claude.ai login"**, sign in via the browser; the seat appears labeled with the entered email. Start a Claude Code session, send "hi" — it answers, and the Anthropic account's own usage shows the burn on the **Max plan** (not gateway credits, not an API key). Falsifier: any step requires a terminal or a pasted token, or the session bills anything else. Secondary checks: revoking the seat removes it from the next render and the harness refuses with plain words; an aborted mint leaves no secret anywhere (vault, disk, logs).

## Proof tests

- `seat_mint_store_render_launch_roundtrip` — mint (stubbed token print) → vault row → render → apply → launch env carries the token + per-seat dir with the strip list applied.
- `mint_capture_never_touches_disk` — abort at each step; no secret in vault, files, or logs.
- Seat render/strip-list cases in the `route_auth` render tests; contract-fixture pin breaking both sides until aligned.
- Store write-gate tests for the seat selection shape; selection-rules kind-counting radio.
- New this pass: `mint_single_flight_survives_concurrent_starts`, `a_replaced_exited_mint_is_torn_down_not_stranded`, `claim_purges_the_replay_buffer`, `startup_sweep_removes_orphaned_mint_scratch`, `a_bare_carriage_return_resets_the_visible_line`, renderer pool+pin dedupe test, seat-mint workflow double-claim vitest.
- Existing suites green throughout (`test_agent_gateway_*`, route_auth tests, pane vitest).

## Findings triage (9 findings, 7 distinct — all fixed, none rejected)

| # | Finding | Outcome |
|---|---------|---------|
| 1+7 | Single-flight mint guard was check-then-act across awaits — two concurrent starts could spawn two mint terminals | **Fixed**: the `mint_by_kind` write lock is held across check + spawn + reserve (spawn is synchronous); failed spawns remove their scratch. Concurrency test added. |
| 2+8 | Exited-but-unclaimed mint never torn down; no crash/restart sweep of `agent-auth-mint/` | **Fixed**: scratch removed unconditionally at terminal exit; a replaced stale slot tears the old capture down (wipe + purge); startup sweep deletes all orphans (mint state is memory-only, so anything on disk at start is an orphan). |
| 9 | Claimed token stayed re-readable from the ws replay buffer until close | **Fixed**: the one-time handoff purges the replay frames (zeroed, sequence floored); close and replacement purge too; a Ready-but-unclaimed capture expires after a bounded 5-min post-exit claim window. Residual: copies already inside the broadcast ring reach only already-connected subscribers — the same surface that watched the live stream. |
| 3 | Bare `\r` stripped instead of resetting the visible line — a `\r`-overwrite spinner would merge into the token line | **Fixed**: `feed` splits on `\r` as well as `\n`; `\r\n` pairs flush an empty segment harmlessly. Test added. |
| 5 | Overlapping poll ticks could double-claim and flip a successful mint to an error | **Fixed**: the tick re-checks `settled` after its await. Regression vitest added. |
| 6 | Validator accepts pool+pin seat rows together; renderer emitted the pinned seat's token twice | **Fixed** in the renderer: seat sources dedupe by seat id per harness, so the document carries each active seat at most once (spec §2's "every active seat, in vault order"). The kind-counting radio itself is kept — it is named in the delivery-spec scope. |
| 4 | `sanitize_terminal_line` leaked an unwiped heap copy via `trim().to_string()`; `wipe_bytes` was elidable | **Fixed**: sanitize trims in place and returns its own buffer; `wipe_bytes` zeroes the full capacity with volatile writes + a compiler fence. Residual (noted): `Vec` growth reallocations can still strand fragments — full mitigation would need `zeroize`-style allocation control, out of proportion for this slice. |

https://claude.ai/code/session_015XM9Un7yY8v36SDUisKxo3

---

## Post-acceptance fix (2026-08-27): the mint captured 99 of 108 characters

The 2026-08-27 acceptance run minted a seat that rendered correctly in the pane and then failed every session with `Failed to authenticate. API Error: 401 OAuth access token is invalid.` The applied document at `~/.proliferate-local/runtimes/slice1qa/agent-auth/state.json` carried a `CLAUDE_CODE_OAUTH_TOKEN` of length **99**; known-good `claude setup-token` credentials are **108**. The stored token 401s when sent directly to `api.anthropic.com`, so the runtime applied it faithfully — the **capture** was wrong.

### Root cause, proven on a live PTY

`claude setup-token` renders through Ink, which **hard-wraps its own output at the PTY width** — the wrap is in the byte stream the program writes, not in the terminal emulator. Captured from a real `claude setup-token` on a PTY (no account interaction; the PKCE flow was abandoned, only lengths recorded):

| PTY columns | sanitized segment lengths of the long OAuth-URL line |
|---|---|
| 120 | 120 · 120 · 106 |
| 99 | 99 · 99 · 99 · 49 |

Each fragment is emitted as its own newline-delimited segment, delimiter `\r\r\n`, with **no prefix** — the wrap width equals the PTY width exactly.

The mint PTY is spawned at `cols: 120`, but `use-agent-login-terminal-viewport.ts` sends the xterm fit-addon's measured size (`handle.sendResize(terminalRef.current.cols, …)`), so the operative width is **the auth-terminal pane's width**, which was 99. A 108-character token therefore arrived as:

- fragment 1 — 99 characters, and by itself a valid `^sk-ant-[A-Za-z0-9_-]{40,}$` line;
- fragment 2 — the remaining 9 characters, which is not.

"The last non-empty matching line" selected fragment 1. Neither `feed()`'s `\r` handling nor `sanitize_terminal_line`'s CSI scan was implicated; both were ruled out before this was confirmed.

Widening the PTY is **not** the fix: it moves the boundary, and the width is the user's window.

### The fix

`MintCapture::take_line` now rejoins a hard-wrapped token, width-free. A segment is the previous line's wrap remainder when it is entirely token characters, does not itself start `sk-ant-` (that marks a token *start*), and is no longer than the line it follows (a remainder cannot exceed the full-width head it wrapped out of). Empty segments are transparent — a wrap's `\r\r\n` segments into two of them, as does any blank display line.

Accumulation and admission are now separate facts, which also fixes a latent sibling bug: below ~47 columns even the head cannot satisfy the 40-character tail rule, so the mint used to fail outright; it now rejoins.

### Regression tests, each falsified

Both fail against the pre-fix rule with `captured 99 chars of a 108-char token`, observed by deleting the rejoin branch and re-running, then restoring:

- `a_token_hard_wrapped_by_the_cli_is_rejoined_byte_for_byte` — the pure machine over the real success frame (label line, blank, wrapped token, blank, the two dim trailers) at widths 99/80/40/27.
- `a_hard_wrapped_token_survives_the_live_capture` — the **live service and a real PTY**, wrap emitted byte-for-byte at widths 99/61/37; the claimed token must equal the printed one.

Anti-overjoin guards: `a_later_bare_word_never_joins_a_closed_candidate` (the CLI's own wrapped "…native copy" tail must not be appended) and `a_second_token_line_replaces_the_first_instead_of_extending_it`.

The inline unit tests move to `login_terminal_tests.rs` so both files stay under the 600-line cap.

Verified at the pushed SHA: `cargo test -p anyharness-lib` → **2349 passed, 0 failed, 3 ignored**; `check_max_lines`, `check_agent_auth_secret_logs`, `check_anyharness_boundaries`, `check_anyharness_fences` all pass.

## Second finding — the false green: FIXED BY RULING (2026-08-27, ruled by Pablo in session)

The pane showed green **Authenticated** with 4 models while the stored seat 401'd. Traced chain (in short): the evidence-panes flag is OFF in every build, the legacy `deriveAuthStatus` had **no `seat` arm**, so a selected seat fell into the CLI arm and borrowed `cliAuthState` — bare file/keychain presence of the machine's own native login. Separately, the launch probe never exercises the credential (`send_test_prompt: false`), and the tier-1 trial was gateway-only, so **nothing in the system exercised a seat credential**.

**Disposition (both halves landed on this branch, commit `b756148a4`):**

1. **Seat tier-1 trial** (`route_auth/seat_trial.rs`): one credential-scoped `POST /v1/messages` (`max_tokens: 1`, `anthropic-beta: oauth-2025-04-20`, model `claude-haiku-4-5`) at mint completion — the endpoint a `user:inference`-scoped setup token is provably allowed to call. 2xx → verified (folds onto the existing `Tier1TrialFact::Green`, display green with evidence age); **401/403 → `Expired`, seat stays saved**, red badge; transport/5xx/429 → records nothing, display stays honestly unverified. The fold is guarded on the harness's **applied route actually selecting a seat**, so a stale seat verdict can never color a gateway, BYOK, or **native** state (native is a permanently supported method, ruled — the refusal cutover is cancelled). Memory-only ledger; token copy scrubbed; verdict + harness kind are the only things logged.
2. **A real `seat` arm in `deriveAuthStatus`**: expired → "Expired" (red), evidence greens → "Authenticated", anything short of evidence → the spec's own word, **"Unverified"** — the seat can never again borrow the native login's green. The native method's own row still greens on its own detection, unchanged.

**Falsification, watched by hand:** `a_dead_token_drives_the_full_path_to_the_red_expired_state` drives a deliberately dead token through the real pipeline (local HTTP 401 server → classification → ledger → the probe engine's fold → the shared derivation → `AuthDisplay::Expired`) and was watched fail two independent ways — with the 401/403 classification deleted, and with the engine fold severed. The `deriveAuthStatus` unit tests were watched fail 4/5 against the pre-ruling badge; the passing 5th is the native arm's green, as ruled.

**Convergence note (record, do not build):** slice 4's seat **usage probe** (stacked above this branch) already makes a one-token `/v1/messages` call per seat on cadence — `server/proliferate/integrations/anthropic.py` on that branch, whose request shape this trial mirrors byte-for-byte on purpose. The trial and the usage probe are the same call in the limit and should **unify when slice 4 lands**; nobody should build a second burner.

## Disclosed post-acceptance fixes on this branch

| Commit | Fix |
|---|---|
| `65f71443f` | Mint capture rejoins a hard-wrapped token (Ink wraps at the pane's PTY width; a 108-char token in a 99-column pane captured as its 99-char head and 401'd). Regression tests at the pure machine and through a live PTY, both falsified. |
| `b756148a4` | Seat tier-1 trial + real `seat` badge arm (the false-green ruling above). |
| `d038b2d2d` | Ambient Anthropic **model-selector family** joins the claude sanitization strip list (`ANTHROPIC_MODEL`, `ANTHROPIC_SMALL_FAST_MODEL`, `ANTHROPIC_DEFAULT_{OPUS,SONNET,HAIKU}_MODEL`, `ANTHROPIC_BEDROCK_REGION_PREFIX`) — live incident 2026-08-27 18:04Z: a runtime relaunched from a Bedrock-configured Claude-Code shell inherited `ANTHROPIC_DEFAULT_*_MODEL`, and the seat-authed session resolved its model alias to a Bedrock-format id (`global.anthropic.claude-sonnet-5`) that api.anthropic.com rejected every turn. In-spec: the sanitization law already reads "every Anthropic selector the route did not set"; the list was incomplete against the law's own words. Exemption matches the rerouting flags' (only a `provider_config`-composed selector survives). The sanitize pair moved to `route_auth/sanitize.rs`; render.rs's ratchet shrank 742 → 647. Falsified: the three render tests were watched fail with the family deleted. |
| `7726eff4b` | The seat upload's catch inspects the server's typed detail: a 403 `github_link_required` (single-org off + no GitHub link — the gate in `auth/dependencies.py` that rejects every agent_auth route) now renders "Connect GitHub in Settings → Account" instead of the generic "re-run the sign-in" that burned real Claude.ai sign-ins into the same 403. The detail code provably survives to the catch via `ProliferateClientError` (the cloud SDK's always-installed error middleware). All other failures keep the settled generic copy byte-identical. Falsified: with the branch removed, the 403 test was watched fail ('expected …re-run the sign-in… to contain Connect GitHub') while the generic test still passed. |


pablonyx:
## CodeQL disposition (13 new high-severity alerts at b85ed7ec7)

**12 × `rust/cleartext-logging`** (all in `profile_tests.rs` / `contract_fixture_tests.rs`): technically true — test `panic!("{other:?}")` would print fixture keys. But the root cause is real and reaches production: `ApiKeyProfile.value`, `GatewayProfile.key`, `SeatProfile.env`, and `ProviderConfigProfile.env` all **derive `Debug` with raw secret fields**, so any `{:?}` on an error/tracing path would echo a live token — contradicting this slice's own law (`SeatProfile` doc: "without ever echoing the token"). **Fix in flight on this branch**: manual redacting `Debug` impls (length-only markers, env key names kept, values redacted) + canary tests asserting no fixture secret survives debug formatting, falsified by hand. This clears the 12 at the root rather than dismissing them.

**1 × `rust/path-injection`** (`runtime.rs:326`, mint scratch dir): **false positive**. The flagged path is `runtime_home.join("agent-auth-mint").join(format!("mint-{uuid_v4}"))` — operator config + string literal + server-generated UUIDv4. The only user-provided value in scope (`kind`) is validated to `Claude` and never joins the path. Recommend dismissal (founder's call); CodeQL is not in the required-checks set, so this alert alone doesn't block the merge mechanically.

pablonyx:
## CodeQL re-analysis at d61db819a — residual alerts disposition

The root fix landed (manual redacting `Debug` impls on all 7 secret-bearing types + 8 canary tests, each hand-falsified). CodeQL re-ran and still reports **16 `rust/cleartext-logging` alerts, all in the three test files** (`profile_tests.rs`, `render_tests.rs`, `contract_fixture_tests.rs`). These are **false positives by construction**:

- The flagged sinks are test `assert!`/`panic!` messages that format secret-bearing types with `{:?}`. Post-fix, that output is redacted (`<redacted N bytes>`); several flagged lines ARE the canary tests whose entire assertion is that no secret survives debug formatting.
- CodeQL's taint tracking follows the value into the sink and cannot evaluate the custom `Debug` impl, so the alerts persist no matter what the impl does. The values are fixture strings (`sk-canary-fixture`), never real credentials.
- The guarantee CodeQL wants is enforced by the canary tests: sabotage the redaction → tests fail with the canary visible → restore → pass (falsification recorded in the fix commit).

Together with the previously dispositioned `rust/path-injection` FP (`runtime.rs` mint scratch dir: `runtime_home` + literal + UUIDv4), all 17 open alerts are documented false positives. CodeQL is not in the required-checks set; merging on the founder's standing approval with this note as the dismissal record. Founder may dismiss the alerts in the security tab at leisure (suggested reason: "used in tests" for the 16, "false positive" for the path one).

PR #2285, opened 2026-08-27
feat(agent-auth): slice 4 — seat usage meters (probe, samples, pane bars)

pablonyx:
Implements the frozen delivery spec `delivery/agent-auth/delivery-spec-slice-4-meters.md` (slice 4 of the approved auth slice plan; stacked on `agent-auth-slice-1`, retargets to main at its merge). Evidence of record: `specs/systems/agent_auth/README.md` §2 (`seat_usage_sample` DDL), §3 Flow 5 (the usage probe), §4 cell 1 (`seats.py` / `seat_usage_probe` / `GET /seats/usage`).

## What ships

- **Schema** — `seat_usage_sample` per §2's DDL, one alembic migration (`a9f3c17b42d8`, parented on slice 1's `d9e4b7a2c6f1`; traversable downgrade).
- **The probe loop** (`server/agent_auth/seats.py`) — `seat_usage_probe(api_key_id)`: decrypt the seat token in memory, one-token `/v1/messages` under `Authorization: Bearer` + the OAuth beta header, defensive parse of the `anthropic-ratelimit-unified-*` headers (absent/garbage → a `probe_failed` sample, never a crash or a guessed number). Cadence: `agent_seat_usage_probe_active_interval` (300s) / `agent_seat_usage_probe_idle_interval` (1800s), per-seat, derived from the seat's own samples; provider errors back off exponentially to the one-hour cap; revoked seats never enter the roster; the writer prunes samples older than 30 days. Worker metronome ticks every 60s, gated on `run_background_workers` only (seats have no gateway dependency).
- **The read** — `GET /seats/usage`: latest sample per active seat, pane-shaped (`apiKeyId, util5h, util7d, reset5h, reset7d, bindingWindow, status, sampledAt`), no raw header passthrough.
- **Pane meters** — inside each seat row (additive for slice 2's serving/next tags): 5h/7d bars with reset times, binding window emphasized, warning treatment at ≥75%, honest dash + sample age for `probe_failed` / no-sample-yet. Opening the pane forces one fresh sample (`POST /seats/usage/refresh`).
- **The advisory-only law as a test** — an import-scan unit test walks `server/proliferate` and fails if anything outside the closed allowlist touches the sample store/model/record; the launch/render path (`state_render.py`, `selection_rules.py`, `service.py`) is structurally barred from ever entering that allowlist.
- **Secret hygiene** — the token exists in a local for the request only; the integration raises status-only errors; `check_agent_auth_secret_logs.py` already scans the probe module's roots; an integration test asserts the token reaches no sample, log line, or response.
- **Docs** — spec README delta row + build-list item discharged.

## Deviations from spec text (disclosed, not silent)

1. **FK `ON DELETE CASCADE`** on `seat_usage_sample.api_key_id` — §2's DDL writes a bare `REFERENCES`. Vault rows cascade with their user, so a bare FK would make user deletion impossible once a seat has samples. CASCADE is the only consistent choice; flagged for spec back-port.
2. **One index beyond the DDL** — `(api_key_id, sampled_at)`; the latest-per-seat read and the 30-day prune both walk it. Performance only.
3. **`POST /agent-auth/seats/usage/refresh`** — a route beyond §4's API table. Flow 5 says "a settings-pane open forces one sample" without naming the door; a side-effecting GET was the alternative and worse. Guarded by a 60s per-seat freshness floor so pane flapping cannot amplify outbound probes.
4. **"While a session runs on the seat"** (active cadence) — the server holds no session-on-seat fact in this slice's dependency set (seat serving is runtime-owned; rotation is slice 2). The active signal is the seat's own account-global movement: utilization ROSE between the last two successful samples (decay is the idle signature); a freshly minted seat probes at active cadence until a steady pair exists. Documented at the cadence function.
5. **Meters query refetches every 60s while the pane is mounted** — a DB-only read so the acceptance-gate bar visibly moves; the spec's "no client polling loop" law governs the runtime status document subscription, not this server-fed pane read. The probe cadence itself stays server-owned.
6. **`binding_window`** — the live capture surfaced a provider claim header (`anthropic-ratelimit-unified-representative-claim`, values matching our vocabulary verbatim); it is preferred, with utilization-comparison derivation as fallback and NULL on a tie (no invented emphasis).
7. **Strict header contract** — all four `{5h,7d}-{utilization,reset}` headers are required for a non-failed sample; any absent/unparseable one records `probe_failed` (the defensive-parse rule read strictly; partial bars are never rendered from partial data).

## Live validation (charter-authorized, capped)

Pablo lifted the no-live-probes rule for this slice (two test seats provisioned at a job-scoped path; ≤10 one-token requests). Used **3 of 10**, tokens sent only to `api.anthropic.com` in the `Authorization` header, never printed/logged/committed:

- **2 header captures** (curl, `max_tokens: 1`) — seeded `CAPTURE_2026_08_26` in `tests/unit/test_agent_seat_usage_probe.py` (rate-limit header values only: utilization fractions, epoch resets, statuses, the representative claim; no ids). This is where the provider's binding-window claim header was discovered.
- **1 end-to-end run** through the real code path (mint from the real token → `POST /seats/usage/refresh` → decrypt → live probe → `seat_usage_sample` row → `GET /seats/usage`), via a throwaway test file that was executed once from the working tree and deleted (never committed). Sanitized result: `status=allowed, util5h=0.63, util7d=0.51, bindingWindow=five_hour`, real reset instants, token absent from every response.

## Adversarial panel

Mandatory multi-agent panel, run before opening this PR and re-run after the fix commits. Shape: **4 independent agents** — (A) correctness, (B) spec-fidelity vs the frozen delivery spec, (C) secret-hygiene — blind to each other, plus a post-fix verification agent that re-audited every claimed fix; lenses A and the verifier independently EXECUTED the test suites (verifier: 42 pytest + 48 vitest + 4 checkers, all exit 0, "all 16 claimed fixes hold, no regressions"). The verifier's own residual findings are closed in the final commit (`f0597aa4c`), whose changes are the verifier's prescriptions and are covered by the re-executed suites (37 targeted tests green). Every finding and its disposition:

| # | Lens | Severity | Finding | Disposition |
|---|---|---|---|---|
| C1 | hygiene | major | `check_agent_auth_secret_logs.py` did not scan `integrations/anthropic.py` (the probe's HTTP half, which holds the decrypted token) | **Fixed** 23bbf3eb6 — root added to `SCANNED_ROOTS` + the PROD-AGENTAUTH-001 record's scope |
| C2 | hygiene | minor | guard regex missed the probe path's own bindings (`token` local / `oauth_token`) | **Fixed** 23bbf3eb6 — `seat_token`/`oauth_token` identifiers added; local renamed `seat_token` |
| C3 | hygiene | minor | `raise … from exc` could carry an h11 header-validation message quoting the token | **Fixed** 23bbf3eb6 + panel-fix commit — `from None`, intake refuses non-printable-ASCII tokens |
| C4 | hygiene | note | httpx `trust_env` default could let `HTTPS_PROXY` interpose | **Fixed** 23bbf3eb6 — `trust_env=False` (URL is constant) |
| C5 | hygiene | note | `binding_window` had no CHECK | **Fixed** 23bbf3eb6 — vocabulary CHECK in model + migration |
| C6 | hygiene | note | `log_cloud_event(` is not a recognized LOG_OPENER in the checker | **Accepted** — pre-existing checker scope; this slice passes only ids through it; flagged for a follow-up |
| A1/B2 | correctness+spec | major | no per-seat crash isolation: a non-ASCII token (`UnicodeEncodeError` before any `httpx.HTTPError`) or a corrupt ciphertext (`InvalidToken`) escaped the probe's catch, aborted the WHOLE per-tick transaction for every user every 60s, and could hand a token-character-quoting exception to Sentry | **Fixed** panel-fix commit — integration catches `Exception` (`from None`); decrypt failure records `probe_failed`; per-seat belts in the pass and the pane-open force; intake refuses non-printable ASCII; unit+integration tests added |
| B1 | spec | major | the advisory-only import scan was a naming check bypassable via the store-package alias (`agent_gateway_store.…`), aided by the `SeatUsageSampleRecord` re-export | **Fixed** panel-fix commit — re-export dropped (sample surface absent from the package namespace, asserted), raw-text sweep of the launch-path modules for `seat_usage`/all sentinels, allowlist tightened |
| B3 | spec | major | live bars carried no sample age — an hours-old `allowed` sample rendered as a fresh bar ("never a stale bar pretending to be live") | **Fixed** panel-fix commit — every live meter shows `checked Xm ago`; vitest added |
| B4 | spec | contradiction | `POST /seats/usage/refresh` exceeds the system spec §4 route table (the pane-open poke has no named door) | **Fixed as a spec amendment** in this PR (route added to §4's table) — flagged here for Pablo's sign-off |
| B5 | spec | contradiction | the meters hook re-reads on a 60s interval vs cell 4's "no client polling loop" | **Fixed + amended** — `refetchIntervalInBackground: false`, re-read only while meters are visible; cell 4's boundary text amended to name the meters' server-fed re-read; flagged for sign-off |
| B6 | spec | contradiction | "active while a session runs on the seat" is server-unknowable (seat serving is runtime-owned) | **Amended** — Flow 5 wording now states the implementable rule (rising utilization between samples, detection lag ≤ one idle interval); flagged for sign-off |
| A4 | correctness | minor | refresh's `setQueryData` could be clobbered by an in-flight GET | **Fixed** panel-fix commit — `cancelQueries` before `setQueryData` |
| A5 | correctness | minor | percent label rounds to 75% without warning treatment at util 0.745–0.749 | **Fixed** panel-fix commit — warning judged on the rounded percent; boundary vitest |
| A6 | correctness | note | a reset already in the past rendered as a bare future-looking time | **Fixed** panel-fix commit — past resets suppressed (the age line carries staleness); vitest |
| A8 | correctness | note | migration docstring claimed the index serves the prune (it cannot: no leading-column match); latest-per-seat could be DISTINCT ON | **Fixed (docstring)** — claim corrected; window query kept at v1 scale |
| A2 | correctness | minor | refresh holds a pooled DB session across N sequential outbound calls with no server-side deadline | **Accepted** — seat counts are 1–3 and each probe is bounded by the 15s httpx timeout; revisit if seat pools grow (noted in the route docstring) |
| A3 | correctness | minor | freshness floor is read-then-probe; concurrent refreshes can double-probe | **Accepted** — worst case is one duplicate one-token request per seat; a server-side lock is disproportionate; the floor still bounds amplification |
| A7 | correctness | note | `limited` with NULL binding window marks both bars "Limit reached" | **Accepted as intended** — an unknown binding window flags both honestly |
| A9 | correctness | note | with a configured active interval < ~14s the 8-sample window plateaus below the 1h cap; a failed idle seat retries at 600s (< its 1800s cadence) | **Accepted** — defaults (300/1800) reach the cap; failure retry probing slightly faster than idle is benign and self-heals |
| A10 | correctness | note | `_parse_epoch("0")` yields a 1970 reset (accepted); duplicate headers collapse to `"a, b"` → `probe_failed` | **Accepted** — defensive floor; duplicates failing closed is the designed behavior |
| A11 | correctness | note | persistent pass-level failure would `report_critical` per minute; N replicas would each run the loop | **Mitigated by A1's fix** (seat failures no longer crash the pass); multi-replica duplicate probing accepted for today's single-replica deploy, bounded by cadence |
| A12 | correctness | note | a seat minted while the pane is open shows "No usage data yet." for up to ~2 min (no re-poke) | **Accepted** — worker tick + visible-interval refetch bound it |
| B-m1 | spec | minor | probe egress doesn't use the repo's pinned-address helper (`integration_oauth/netsafety`) | **Accepted with rationale** — that helper exists for attacker-influenced URLs; this URL is a constant, `trust_env=False` closes proxy interposition; the spec's "same pinned-address egress rules as every outbound call" premise (no shared helper exists; other fixed-host integrations pin nothing) is surfaced here for the spec owner |
| B-m2 | spec | minor | `seat_usage_probe(db, *, api_key_id) -> SeatUsageSampleRecord \| None` vs the spec's `seat_usage_probe(api_key_id) -> UsageSample` | **Accepted** — the db handle is the codebase's store-call idiom; `None` = seat vanished mid-pass |
| B-m3 | spec | minor | DDL riders (FK CASCADE, extra CHECK, index) and the binding-window claim/derivation behavior | **Disclosed** — see the deviations section above |
| V-N1 | verifier | low-med | the isolation belts caught exceptions but a failed INSERT left the session pending-rollback, still aborting the rest of the tick (and 500ing the pane's force-read) | **Fixed** f0597aa4c — a savepoint per seat write; only that seat's write rolls back |
| V-N2 | verifier | low | the crash-isolation paths had no tests | **Fixed** f0597aa4c — decrypt-failure → probe_failed-for-that-seat-only and surprise-crash → skip-the-seat-not-the-pass integration tests |
| V-N3 | verifier | low-med | (a) `agent_gateway_store.seat_usage.…` attribute access from modules outside the launch-path sweep was invisible to both scans; (b) laundering through an allowlisted module's re-export defeats any static scan | **(a) Fixed** f0597aa4c — the raw-text sweep is repo-wide with a closed mention-allowlist; **(b) accepted** — no static scan survives deliberate laundering; review is the backstop |
| V-N4 | verifier | cosmetic | `formatAge` of an unparseable timestamp renders empty age copy | **Accepted** — unreachable with server-fed ISO timestamps |
| V-nit | verifier | nit | the intake error says "printable ASCII" but space (0x20) is also refused | **Accepted** — a credential can never contain a space; the copy stays plain |

## Test matrix

| Gate | Result |
| --- | --- |
| `tests/unit/test_agent_seat_usage_probe.py` (header parse on the real capture + absent/garbage cases, cadence/backoff, the advisory-only scans: AST import scan + repo-wide text sweep + package-namespace assertion) | 28 passed |
| `tests/integration/test_agent_seat_usage_api.py` (latest-per-seat, refresh + freshness floor, probe_failed + backoff, decrypt-failure and surprise-crash isolation, revoked roster, 30-day prune, mint bad-token refusal, token-never-leaks) | 9 passed |
| Full server suite (`tests/unit` + `tests/integration`, sequential) | 1561 + 740 passed after the two count-pin bumps (wall-clock owners 172, /agent-auth route shape) |
| `check_mypy_baseline.py --compare-ref origin/main` | passed, 0 new diagnostics |
| product-client build + typecheck + full vitest (incl. 7 new `HarnessSeatUsageMeter` tests) | 8185 passed |
| desktop `tsc --noEmit`, web typecheck | passed |
| `check_max_lines` (ratchets: `db/models/agent_gateway.py` 563→612, `HarnessPane.test.tsx` 1270→1274, `test_agent_gateway_api.py` 1369→1373), `check_session_mutation_admission` (no AnyHarness routes touched), `check_server_boundaries`, `check_frontend_boundaries`, `check_docs`, `check_manifests`, `check_agent_auth_secret_logs` (now also scanning `integrations/anthropic.py`), `check_migration_heads` (head `a9f3c17b42d8`) | all passed |

Alembic: parented on slice 1's `d9e4b7a2c6f1`. `agent-auth-slice-3` exists on origin but carries no migration yet (its head is the contract-fixture barrier commit); if slice 3's `acked_sequence` migration merges ahead of this PR, this one re-parents (one-line `down_revision` change) before merge — `check_migration_heads.py` gates it either way.

Local-environment note: the machine's homebrew pnpm jumped 10→11 mid-build, which breaks ~120 unrelated jsdom suites locally (`localStorage` undefined in files this diff never touches — verified failing in isolation with zero diff lines in those areas). CI pins pnpm 10.11.0 and is authoritative; the full product-client suite was 8185/8185 green under pnpm 10 on this branch's code, and the touched suites (48 tests) are green under both.

## Pablo's acceptance run (the merge gate, from the frozen spec)

1. On a build of this branch with your minted seat(s): Settings → Agents → Claude — each seat row shows live 5h and 7d bars with reset times, the binding window emphasized, warning treatment at ≥75%.
2. Run a heavy session on the seat for a few minutes with the pane open — the bar moves, and matches `claude /usage` on that account.
3. Falsifier sweep: no launch-path code reads `seat_usage_sample` (the import-scan test `tests/unit/test_agent_seat_usage_probe.py::TestAdvisoryOnlyLaw` is the standing proof), and a probe failure (e.g. network off) changes nothing about launches — sessions start identically; the meter shows an honest dash.


pablonyx:
Founder sign-off recorded (Pablo, 2026-08-27, via ruling session): the three disclosed spec amendments are RULED-ACCEPTED as written — (1) POST /agent-auth/seats/usage/refresh as the pane-open door with the 60s per-seat freshness floor; (2) the meters' visible-only 60s server-fed re-read, with cell 4's boundary text naming it; (3) Flow 5 'active cadence' = rising utilization between samples. The FK ON DELETE CASCADE + index DDL riders are approved for spec back-port. These stay in the delta table until this PR's rows discharge at merge.

Issue #2276, opened 2026-08-27
await_holding_lock: 123 suppressed sites need a careful lock-scope pass


pablonyx:
Ruled 2026-08-26 (lint wiring slice): clippy's `await_holding_lock` fires at 123 sites across the runtime workspace — a `MutexGuard` held across an `.await` (deadlock hazard). Fixing them is surgery in the runtime that is the product, so they are suppressed via `[workspace.lints.clippy] await_holding_lock = "allow"` with this issue as the tracked debt record (`lints/anyharness/native-debt.toml`, AH-CLIPPY-1). The pass: restructure each lock scope so the guard drops before the await (or move to an async-aware mutex where the critical section genuinely spans the await), with tests, one domain at a time. Flip the lint back to warn/deny when the count reaches zero.

pablonyx:
**Reframe + resolution (the lock map, 2026-08-27 — `Vault/Testing Linting CICD/60 Lock Map.md`):** this issue's premise ("123 sites, surgery in the runtime that is the product") was wrong. The full enumeration found **zero production sites** — 122 of 123 were one pattern (the test-support `ENV_MUTEX` held across async test bodies, benign by construction) and one was a lexical false positive. None of the three live incidents attributed to this class was a lock bug (two were fixture/teardown races, fixed at root; the 6-hour hang is a terminals cancellation-atomicity bug, tracked separately for a founder-paired slice).

**Resolved by PR #2309**: `ENV_MUTEX` → `tokio::sync::Mutex` (async guards are the lint's sanctioned semantics), the false positive scope-blocked, and `await_holding_lock` removed from the allow-list — it now rides `-D warnings` as law. Audit proof: `--force-warn clippy::await_holding_lock` reports exactly **0** findings (note for future audits: plain `-W` under `-A warnings` silently counts zero on rustc 1.98 — `--force-warn` is required). Record `AH-CLIPPY-1` retired to `status = "holds"`, enforced by the rust-lint lane. Closing on merge of #2309.

Issue #1714, opened 2026-08-07
lints: server grid coordinate enforcement gaps (seed ledger)


pablonyx:
Umbrella ledger for the unenforced server grid rules recorded in
`specs/codebase/structures/server/guides/grid-ownership-model.md` §Current gaps,
now carried as `status = "leaks"` / `gap` records under `lints/server/`:

- Coordinate enforcement absent (no TOML coordinate map, general check_grid.py, package-audience map, generated service graph, owner/consumer map)
- Store ownership / foreign-read declarations not generally enforced
- ORM → dataclass boundary incomplete (stores returning ORM types; SSO mutating User + db.flush())
- Request transaction ownership incomplete (route-owned session calls in Accounts Desktop/Identity/SSO APIs)
- Authorization dependency adoption partial (inline service checks; permissions.py composition)
- Error enforcement incomplete (no per-module classification; error codes not globally unique)
- Library audiences not represented or enforced (lib/** audience/purity unscanned)
- Raw-transport placement one exact exception (cloud/gateway/proxy.py httpx)
- Service topology and package audiences not gated (Cloud Sandbox ↔ Gateway import cycle)

Each is a rule that COULD be mechanical and is not yet. Enforcing any of them is
follow-up work, not part of the docs-v1 lints move.

pablonyx:
Inventory sweep (docs-v1 D6) found four more unenforced server gaps, from `specs/codebase/structures/server/guides/background.md` §Current gaps — adding them to this seed ledger:

- Background execution not unified: periodic process loops still parked outside the Celery background plane (grid doc §Current gaps bullet 10)
- Billing reconciliation runs as a parked periodic process loop rather than a Beat-scheduled task
- Anonymous telemetry runs as a parked periodic process loop
- Agent Gateway enrollment backfill loop parked outside the unified background plane
- Agent Gateway usage import loop parked outside the unified background plane

Like the rest of this ledger: each could be mechanical and is not yet; enforcing them is follow-up, not part of the docs-v1 lints move.

PR #2293, opened 2026-08-27
chore(lints): wire the rust plane — clippy -D warnings + rustfmt gating, toolchain pin, ruff over the engines, constitution cleanup

pablonyx:
## ⚖️ Net-new exceptions — RESOLVED by founder rulings

**Ratified** (comment on this PR): exception 1 (fn-scoped `unnecessary_cast` on the socklen casts) and the crate posture, taken as **option (a)** — the crate-level conditional allow landed in `c4ba2c3d8`. **Then Pablo ruled FIX (not exclude) for the crate's remaining ~54 Linux findings** — all repaired for real in `5fc99f3f5`, **zero new allows** (details in "The gate's first catch" and the fix summary below). The original decision text is preserved below for the record.

### Original decision text (historical)

The ruled allow-list was exactly the 4 judgment lints + `await_holding_lock`. Making the Linux CI gate green surfaced **two platform-skew allows beyond that list** — net-new exceptions, so per the constitution this PR holds at the door until you ratify (or order the repair). Both are Linux/macOS false-positives, not hidden debt: in each case the code is *correct on both platforms* and the lint fires only on the platform where the construct compiles to a no-op.

**1 · `#[allow(clippy::unnecessary_cast)]`, fn-scoped** — `collect_descriptors` (`proliferate-diagnostics-client/src/bridge/framing.rs:300`), two casts of `libc::msghdr.msg_controllen`.
- *Why false-positive:* `msg_controllen` is `socklen_t` (u32) on macOS but `usize` on Linux — the `as usize` casts are **real conversions on macOS** and no-ops on Linux, where the lint fires. Removing them breaks the macOS build.
- *No-suppression alternative + cost:* rewrite both sites as `usize::try_from(message.msg_controllen).unwrap_or(usize::MAX)` (or platform-cfg'd conversion helpers) — churns correct unsafe-adjacent framing code and adds an error path that cannot occur, purely to satisfy a platform-dependent lint.

**2 · The desktop crate's posture vs the Linux clippy authority** — the `--keep-going` round exposed the full truth: the Linux-only findings span the ENTIRE `apps/desktop/src-tauri` crate (`sidecar.rs`, `diagnostics/support_snapshot/*`, `commands/cloud_worker/*`), not just the collector module — my module-level allow (`lib.rs:6`, in the branch) was aiming one level too low. The crate is the macOS-native product; on Linux large cfg'd regions compile out and orphan their imports/helpers. **The final CI round is RED and stays red pending your choice:**
- **(a) REC — crate-inner `#![cfg_attr(not(target_os = "macos"), allow(unused_imports, dead_code))]` in `lib.rs`**: cfg-conditional and crate-wide; the macOS side keeps every warning; same honest shape as the module fix, one level up. A 2-line commit after ratification.
- **(b) exclude the crate from rust-lint's clippy invocation** (+ a lane-census note): compile coverage stays via Cargo check & test, but `-D` authority over the crate's Linux-visible *real* lints is lost.
- **(c) a proper cfg-hygiene pass over the crate**: the correct end-state — per-region cfg so nothing is orphaned on any platform — but a daytime slice, not a 3am commit.

> [!decision] PABLO DECIDES (related): a macOS clippy lane?
> The Linux `rust-lint` job can never see the desktop crate's macOS-native half — ~62 macOS-side warnings are invisible to the authority by construction. Options: (a) accept Linux-only gating (today's state; the macOS half is policed only by local `make gate` on macOS machines); (b) add a macOS runner to `rust-lint` (real coverage, ~10× runner cost, slower gate). Rec: (a) until the collector seam stabilizes, then revisit.

---

Implements the frozen delivery spec `delivery/testing-cicd/delivery-spec-lint-wiring.md` (#2271) and the testing system spec's lints sections (`specs/engineering/testing/lints.md` § Buy vs build, § The six families — families 1 and 5). All rulings of record 2026-08-26; freeze deviations noted below.

## Scope summary

- **rust-toolchain.toml pin (1.98.0, rustfmt+clippy components)** — the mirror rule for rust: gate-local, hooks, and CI can never skew on fmt/clippy verdicts. Bump deliberately, with the reformat a bump implies.
- **Clippy mechanical fixes** across the workspace (two `--fix` rounds + ~40 hand-fixes: ptr_arg chains through the fork/session drivers, question_mark, doc-list indentation, manual_strip, explicit counter loop, collapsible match, needless updates, suspicious open options with an explicit `truncate(false)` on the installer lock, octal-escape literal, noop clones).
- **`[workspace.lints.clippy]` allow-list** (root Cargo.toml; every member crate opts in): the four judgment lints (`result_large_err`, `too_many_arguments`, `type_complexity`, `large_enum_variant`) — **provisional, re-ruled at the code-debt migration** — and `await_holding_lock` with tracked debt record **AH-CLIPPY-1** (`lints/anyharness/native-debt.toml`, issue #2276, 123 sites, careful lock-scope pass queued).
- **Dead code = named sites, never a global allow**: 47 `#[allow(dead_code)] // AH-CLIPPY-2` tagged sites (39 diagnostics; multi-field/variant items carry one marker per site) (several are half-wired features — e.g. `SessionCommand::CancelTurnIfActive` construction, `LiveTurnCancelOutcome` — for domain owners to delete or revive); plus a real cull of the worker's orphaned test scaffolding (latch-probe machinery, unused JSON/config fixtures — orphaned by the worker-legacy cull, zero consumers).
- **rustfmt normalization** under the pin as its own formatting-only commit: `5644287ac` (209 files). For `.git-blame-ignore-revs` (Biome slice creates it): **if this PR is squash-merged the fmt SHA dissolves into a mixed commit — prefer a merge commit here, or accept the mixed SHA in blame-ignore.**
- **`rust-lint` CI job** (ci.yml): `cargo fmt --all --check` + `cargo clippy --workspace --all-targets -- -D warnings`, in the `ci-ok` needs (drift-guarded); Playwright-apt census updated for the second Cargo job.
- **The gate's two flips** (inherited from #2280, deliberately landed with the CI job per the mirror rule): `rust: fmt` advisory → gating; `rust: clippy` info-mode → `-D warnings`; `test_gate.py` pins the new doctrine (38/38).
- **Ruff over the engines**: `scripts/ruff.toml` (mirrors server config minus ANN, pinned `uvx ruff@0.16.2`), 131 violations fixed + 46 files formatted, repo-shape CI step; `server-ci` ruff scope widened to `server/scripts/`.
- **`lints/server/gaps.toml` deleted** (8 `enforced_by="review"` records) — **constitutional amendment, founder-pre-approved 2026-08-26** ("honest beats aspirational"). Citations of the dead ids fixed in `specs/areas/server.md` and `lints/server/boundaries.toml` (#1714 keeps the debt visible).
- **`PROD-COMPLIB-1..8` records** (`lints/product/component-library.toml`) — the last recordless checker joins rules-as-data; its diagnostics now cite the record ids.
- **`lints/native-tools.md`** updated (clippy/rustfmt wired + dated, ruff scope, toolchain pin, Biome pending).

## Acceptance gate (verbatim from the freeze — performed by Pablo or a directed agent)

> Open a scratch PR containing one unformatted `.rs` file and one `let x = y.clone();` where the clone is useless: the `rust-lint` job goes **red naming the lint**, and `ci-ok` fails with it. Falsifier: that PR shows `rust-lint` green · `rust-lint` absent from `ci-ok`'s needs · `cargo nextest run --workspace` red on main after the mechanical fixes.

## Proof (all run at head, macOS, pinned toolchain)

- `cargo clippy --workspace --all-targets -- -D warnings`: **green** (desktop crates `proliferate`/`proliferate-debug` excluded locally — their test targets have **pre-existing macOS-only compile breakage on main**; CI's Linux lanes compile them green and the new job runs the full workspace there).
- `cargo fmt --all --check`: **green**.
- `cargo test --workspace` (same exclusions): **2820 passed / 0 failed**, skipping exactly two **proven pre-existing** failures: `the_sweep_removes_only_abandoned_and_old_scratch_roots` (fails identically with base-commit sources — macOS mtime semantics) and `legacy_probe_shape_refuses_targeted_fork_end_to_end` (flaky Drop cleanup); their panics poison the shared env mutex and cascade under threaded `cargo test` — nextest's process-per-test on CI is immune.
- `python3.12 scripts/lint_records.py` (202 rules) · `test_lint_records` + `test_check_component_library` + `test_gate` (38/38) · `check_max_lines` · `check_docs` (224 files) · `check_manifests` · fences: **all green**.
- `node --test scripts/ci-cd/*.test.mjs`: **215/215** (rollup drift guard sees rust-lint).
- `uvx ruff@0.16.2 check` + `format --check` over `scripts/` + `server/scripts/`: **green**.

## Findings triage (inline adversarial pass — see review note below)

| # | Finding | Outcome |
|---|---------|---------|
| 1 | clippy's earlier "clean" readings were fingerprint-cache skips; the `[lints]` manifest change surfaced 112 real deny-level sites in anyharness-lib | **Fixed**: fix-round-2 + ~40 hand-fixes + named-site dead-code markers; gate green with fresh fingerprints |
| 2 | `--fix` removed imports used only by test targets (34 compile errors) | **Fixed**: imports restored/relocated into their `cfg(test)` scope |
| 3 | `&PathBuf → &Path` signature fixes rippled noop-clones into the fork/session driver chain | **Fixed**: `.to_path_buf()` at request-construction sites; chain compiles + tests green |
| 4 | rustfmt normalization changed measured line counts: 6 ratchet rows stale (shrunk) and 6 grown + 6 files newly over the cap | **Fixed as census** (precedent #1656/#2222): rows re-measured; 6 new rows carry the "mechanical rewrap, not new debt" reason — **ratchet amendment for founder review** |
| 5 | Deleted gap ids still cited in `specs/areas/server.md` + `boundaries.toml` (breaks "citable forever") | **Fixed**: citations reworded to #1714 with the deletion noted |
| 6 | `never_loop` deny on the actor's `'drain: loop` — every arm breaks; either a labeled block in disguise or a missing `continue 'drain` | **Flagged, not fixed** (behavior-preserving mandate): inline allow + note to the sessions owner; resolve in the steering/queue rebuild |
| 7 | Worker test-scaffolding items dead even in test builds | **Culled** (latch-probe machinery, JSON/config fixtures); test-located consts kept under `cfg_attr(not(test), allow(dead_code))` |
| 8 | `clippy::module_inception` on `workspaces/archive/archive` (refuter) | **Allowed inline, documented at the site**: the R5 module family deliberately names the entry-point module after the domain; renaming would churn every consumer for a style lint |

## Freeze deviations (implementation notes, not silent scope changes)

- The **toolchain pin** and the **repo-wide fmt normalization** were not in the frozen spec — added mid-flight by coordinator directive after the fmt "clean" reading was proven a false negative (608 diffs under 1.9.0). Both serve the freeze's mirror-rule intent.
- The freeze's `rust-lint` description said fmt+clippy; delivered exactly that, plus the gate's two flips that #2280 left deliberately to this PR.

## Post-refuter fixes (commits `b3d520eef`, `ff68df71d`)

- Fresh-context refuter verdict: **no block**; its two MINORs are fixed — the gate's clippy now runs `--all-targets` and its fmt is byte-identical (`cargo fmt --all --check`), closing the one-directional mirror skew; the dtolnay action pins `toolchain: 1.98.0` (no double download).
- CI-red fixes: the macOS-only credential-discovery import is cfg-scoped (platform skew survives the version pin — **the Linux CI job is the clippy authority**). Corollary, three instances tonight (the cfg import, and two `unnecessary_cast` allows on libc `msg_controllen` casts that are real on macOS/socklen_t and no-ops on Linux): **Linux-only findings are un-reproducible locally on macOS — the CI clippy log is the source of truth for them**, and the rebase ours/theirs inversion that had clobbered #2280's doctrine test and the ownership atlas is reverted (both files are main's versions, reformatted under `scripts/ruff.toml`).

## Evidence appendix — session-actor concurrency class (for the careful-pass priority case)

The same keep-going CI round's **Cargo check & test hung for ~6h and was killed, leaving an orphaned `anyharness_lib` test process** — the third sighting of the session-actor concurrency class tonight (two intermittent test failures earlier, one hang now). None traceable to this PR (its only touch in that area is formatting + one unused-import removal, refuter-verified); logged here so the `await_holding_lock` careful pass (#2276, AH-CLIPPY-1) carries a live incident as its priority case rather than a lint count alone.

## The gate's first catch — the three `suspicious_open_options` judgments (fix-for-real round, `5fc99f3f5`)

Pablo ruled FIX for the desktop crate's 54 Linux-only findings (complete via `--keep-going`). ~50 were rote mechanical rewrites from the log (`map_or`→`is_none_or` ×8, `DateTime<Utc>` Copy-clones ×4, literal-bool asserts ×4, redundant closures ×3, redundant guard, `question_mark`, `needless_as_bytes`, manual async fn, range pattern, needless-range-loop rewrite with the boundary==end tail-call preserved, char-array patterns ×2, derived `Default`s ×2, visibility widening on three coordinator seams, and three enum de-prefixings whose **shipped wire strings are pinned by per-variant serde renames** — `UPDATER_*` on the Serialize side, `session_*` on the Deserialize side). The `useless_conversion` trio in `unix_fs.rs` got the real fix instead of a platform allow: **mode_t-typed masks** (`stat.st_mode & 0o777`), which type-check identically on macOS (u16) and Linux (u32) with no conversion at all.

The three judgment sites, each read in context:

| Site | Intent | Decision | Live bug? |
| --- | --- | --- | --- |
| `commands/cloud_worker.rs` (worker db lock) | contended lock file; content unused; `try_lock_exclusive` follows | `.truncate(false)` — truncating a file another holder has open would be wrong | **No** — content never read |
| `diagnostics_collector/shutdown_tests.rs:~90` (child-lock poll loop) | lock fixture the child process holds; re-opened in a poll loop | `.truncate(false)` | **No** — content unused |
| `diagnostics_collector/shutdown_tests.rs:~125` (fixture lock) | same fixture, other side | `.truncate(false)` | **No** — content unused |

No site was the stale-tail-bytes bug scenario — all three are lock files whose bytes are never read; the fix makes the already-correct behavior explicit. One near-catch worth recording: local macOS `cargo check` reports an unused `mut` at `support_evidence_unix.rs:319` that the Linux CI list does NOT contain — meaning Linux's cfg paths use it. Deliberately left untouched: removing it would fix the invisible-side warning by breaking the authoritative side. The platform-skew rule cuts both ways.

## Review note

Adversarial review: one **inline** pass (findings above) — a fresh-context refuter could not be spawned from this fork (harness constraint). Per testing law 11 an inline pass is a fallback, not a pass: **coordinator, please run one completed fresh-context refuter before merge.**


pablonyx:
**Founder ratification (Pablo, 2026-08-27):** the two net-new platform lint exceptions are RATIFIED, and the desktop-crate posture is ruled **option (a)** — a crate-level, macOS-conditional allow (`#![cfg_attr(not(target_os = "macos"), allow(unused_imports, dead_code))]` at the desktop crate root), superseding the module-level attribute. The `await_holding_lock` bug class stays out of this PR by design — it is issue #2276's dedicated pass. Merge method: merge commit, preserving `5644287ac` for `.git-blame-ignore-revs`.

PR #2289, opened 2026-08-27
refactor(anyharness): Wave-3 consolidation — the agent_auth runtime cell moves to domains/agent_auth/ with its fence node

pablonyx:
Wave-3 Rust consolidation (agent_auth slice 6, lane B) — move-only, per `delivery/agent-auth/delivery-spec-slice-6-cleanups.md` (lane B) and the agent_auth system spec §0 census. Two commits, replayed onto the live slice-3 tip `ec8461319` (which now carries slice 3's full revision→sequence rename, the status module, and both SDK regens):

1. **The move** (`4e3831013`): `git mv` of `route_auth/`, `auth/`, `launch_probe/`, and `status/` out of `domains/agents/` into the new `domains/agent_auth/`, plus mechanical `use`-path updates in consumers. `api/http/agent_auth.rs` stays api-layer, per the census. `auth_state.rs`, named by the lane's move list as "remains", was deleted upstream when slice 3's status module absorbed the derivation — nothing left to move; its retired path is still banned.
2. **The fence record** (`4bce93e43`): the `agent_auth` node lands in `lints/anyharness/fences.toml` with its measured edge list; every record naming a moved file follows the move (ratchets, AH-LIVE-5 exception ledger, agent_auth lint scopes, secret-log scan roots, boundary-checker calibration anchor, old-path bans, ownership-atlas rows, spec/docs path references).

**These two commits must land together.** The move-fidelity lens flagged the ordering hazard: at the move commit in isolation, ~24 non-Rust records still point at pre-move paths — in particular `scripts/check_agent_auth_secret_logs.py` would scan NOTHING for the route_auth plane, and the agent_auth lint scopes would match nothing. The record commit closes every one of them. Do not cherry-pick the move alone.

## Move manifest

| From (`domains/agents/`) | To (`domains/agent_auth/`) | Files |
| --- | --- | --- |
| `route_auth/` | `route_auth/` | 29 |
| `auth/` | `auth/` | 11 |
| `launch_probe/` | `launch_probe/` | 13 |
| `status/` | `status/` | 3 |

[...]

PR #2301, opened 2026-08-27
feat(agent-auth): slice 3 — the status document, probe self-recovery, and sequence/fingerprint delivery governance

pablonyx:
## Summary

Slice 3 of the agent_auth rebuild — **truth & recovery** ([frozen delivery spec](delivery/agent-auth/delivery-spec-slice-3-truth-and-recovery.md), implemented without re-deriving; system spec §2/§3 Flow 4/§4 cells 2+4 are the sections of record). One machine truth per harness, and a probe engine that heals itself. Stacked on `agent-auth-slice-2`; **Wave 3 (#2289) stacks on this branch.**

Per scope bullet:

- **The `status/` module** (`domains/agents/status/`): one status document per harness, exactly the §2 shape, persisted in the runtime's SQLite (migration `0078`), event-refreshed, served stale-marked while a re-probe runs, never withdrawn; survives restart marked stale until the startup pass re-verifies. Absorbs slice 2's transitional carry (serving/next/cooling now ride `applied.seat_id`/`next_seat_id`/`cooling_until`) and the deleted `auth_state.rs` derivation. Writes go through one transactional door with a per-harness write cell; the probe block is derived from observation columns inside the transaction, so doc and observation cannot tear.
- **Doors**: `GET /v1/agent-auth/status` (`?harness=`), `GET /v1/agent-auth/status/stream` (SSE; snapshot-then-changes, subscribe-before-snapshot, `Lagged` ends the stream so the client's fallback can fire), `GET /v1/agent-auth/methods?harness=`. All served from the document; handlers on the blocking pool. Readiness keeps consuming through `apply_launch_route_upgrade`; the courier's ack rides the PUT response.
- **Probe recovery events**: `PokeReason::{BackoffExpired, FirstDetected}`; a failure arms a one-shot self-recovery timer (disarmed by a newer failure or a success; cursor's manual-only carve-out preserved); `FirstDetected` fires from the startup pass for an installed harness with no persisted row; **`AuthApplied` carries the changed-harness set** — `apply_state_file` diffs per-harness entries (order-stable; origin flip marks every harness; unserializable compares as changed) under a cross-process flock, and the PUT pokes/refreshes only the changed set.
- **Serve-stale observation store**: last observation persists beside the status row; a failure with a prior verified observation dims (`verified`/prior `at`/`stale: true`), never darkens; a first-ever failure reads `failed` honestly. A `ProbeStaleGuard` (RAII, mirroring `LiveStateGuard`) releases the stale mark on every exit that never produced a verdict — coalesce, backoff-refusal, early error, dropped future.
- **Sequence/fingerprint split** (§2 governance, exact): document field `revision` → `sequence`, monotonic per (user, surface), bumped **only by a content-changing render** via a new persisted `agent_auth_render_sequence` counter (single atomic upsert). Fixes the live bug where a seat/key **revoke did not bump** (revision derived from selection `updated_at` only) — pinned by `test_revoke_bumps_sequence_because_content_changed`. `fingerprint` is a `GET /state` rider only, hashing the canonical harnesses array. Equal-revision content-authority clause dropped on both sides; runtime rejects below-persisted, accepts equal idempotently. Contract fixtures renamed (both v2 and the rotate-off variant); courier acks `{sequence, fingerprint}` with the served rider. Alembic `189d414c1778` renames `acked_revision → acked_sequence` **and deletes pre-slice-3 ack rows** (see panel finding 1 — a preserved ms-epoch stamp would sit above the new counter forever and silently wedge `applied`).
- **Launch-options basis fix** (bug f's root): `basis.rs` stops folding the global document revision; the basis is a content hash of the harness's **own** effective auth entry (v4 salt; one-time invalidation). A 25-case adversarial matrix confirms: no-op renders and other-harness changes never invalidate; every genuine change (key rotation, source add/remove/reorder, settings, seats, entry appearing/disappearing, malformed/origin-mismatch) still does.
- **Frontend subscribe migration**: one access-layer seam — `useHarnessStatus(kind)` (SSE via `streamAgentAuthStatus`, GET-seeded, single fallback re-read, transport `fetch` threaded so the cloud surface authenticates) and `useMethods(kind)` wired to the picker. **Deleted with the derivation**: `agent-auth-evidence.ts`, `HarnessAuthStatusBadge.tsx` (`deriveAuthStatus`/`deriveProvidersStatus`), `agent-auth-rotation.ts`, the `agentAuthEvidencePanes` flag module, and the timed onboarding step. All five paths banned from returning by the new `FE-PATHS-1` law + `check_frontend_old_paths.py` (CI-wired; hand-falsified at two paths). Panes render the document verbatim through a pure presentation module; green requires dated evidence; stale renders as stale-with-last-observation, never loading; the status document replaces `authState` on the agents projection as `authStatus` (`credentialState`/`readiness`/`credentialsFromRoute` untouched).

## Adversarial panel (mandatory, multi-agent)

**Shape**: two rounds × three independent lenses, blind to each other, every lens executing suites (not just reading): (a) correctness/concurrency — wrote 9 adversarial tests, 6 failing against the unfixed code; (b) spec-fidelity — clause-by-clause against the frozen spec, including a verdict on whether each named Proof test proves its claim; (c) regression — before/after on every deletion, a 25-case basis matrix, full grep inventory. Round 2 re-audited the final head after the fixes. **Round 1 found 7 BLOCKERs; every one was invisible to the then-green gates.** Cross-lens convergence (found independently twice): the ack wedge, the materialization arm, the CLI derivation, the dropped stream `fetch`.

### Triage table (round 1)

| # | Finding | Sev | Disposition |
|---|---|---|---|
| 1 | Ack migration keeps ms-epoch stamps → `applied` wedges forever on every upgraded install; suppressed upsert returns 200 so nothing retries | BLOCKER (×2 lenses) | **Fixed** `2f3bc9f1a` — upgrade deletes ack rows (receipts addressed by a re-based (sequence, fingerprint) pair denote nothing in the new space); migration test runs real alembic over a seeded ms-epoch row and proves a low-sequence ack then stamps; a second test pins the wedge mechanism so the delete can't be reverted silently |
| 2 | Status persist = non-transactional RMW; `probe.at` regressed 14×/710 attempts; doc/observation tore 550× | BLOCKER | **Fixed** `69f337fcf` — one transactional write door, per-harness write cell, probe block derived from columns in-tx; tearing 1129→0; 20/20 on the probabilistic suite |
| 3 | `notify_probe_admitted` marks stale with no release; aborted refresh wedges the doc | BLOCKER | **Fixed** `69f337fcf` — `ProbeStaleGuard` RAII covering all four exits |
| 4 | Sequence guard wedges across a server switch or same-origin DB rebuild (counter spaces incomparable; origin stamp can't rescue the raw-load path) | BLOCKER | **OPEN — founder-gated** (see Founder question 1). No partial guard shipped; reviewer's cross-server test deliberately not adopted |
| 5 | SSE stream drops connection `fetch` → cloud surface 401s, no fallback (×2 lenses) | BLOCKER | **Fixed** `847a61f19` — `fetch` a required key on the stream options (dropping it again is a compile error); spy tests both directions |
| 6 | Onboarding card can hang forever + permanently consumes a slot (4 non-terminal arms, no ceiling) | BLOCKER | **Fixed** `847a61f19` — every document state terminal or actionable; only a live install and an unanswered projection wait; no timer/poll reintroduced |
| 7 | `tests/unit` RED on branch (`test_infra_time` wall-clock count) — hidden by everyone's `-k` filter | BLOCKER | **Fixed** `15f32e294` — attributed-subset convention followed (not a bare count bump; an eager-`utcnow()` regression stays caught); full `tests/unit` (1554) + `tests/integration` (752) green |
| — | Materialization-failure arm: no `record_failure` → no coalescing (10/10 attempts), no backoff, no recovery | MAJOR (×2 lenses) | **Fixed** `69f337fcf`, one shared timestamp |
| — | Changed-set diff blind to `issuing_server_origin`; unserializable-compares-equal | MAJOR | **Fixed** `69f337fcf` |
| — | Surviving CLI derivation gates Authenticate off `cliAuthState`+readiness fallback → dead end (×2 lenses) | MAJOR | **Fixed** `847a61f19` — affordance keys on the document; four disagreeing-case tests |
| — | ms-epoch credential dirs never GC'd (leaks 0600 `config.toml` with live gateway keys) | MAJOR | **Fixed** `69f337fcf` + test |
| — | Manual-refresh-only harness installed after boot never gets a document | MAJOR | **Fixed** `69f337fcf` (`RefreshCause::InstallCompleted`) |
| — | `useMethods` dead; hook fields with zero readers; `rotate` a second source of truth | MAJOR | **Fixed** `847a61f19` — wired or removed per field; rider ruled authoritative for rotate (only writable side; the doc's copy is the delivered echo) |
| — | Onboarding lost specific next-actions + diagnostic line | MAJOR | **Partially fixed** `847a61f19` — what the document honestly supports is wired (choose-a-source, mint offer, install); failed-probe attribution deliberately NOT invented (credits-exhausted is a cloud fact absent from the document; typed per-cause reasons are an open build-list item) — pinned by test |
| — | `FE-PATHS-1` fenced 1 of 3 claimed paths | MAJOR | **Fixed** `847a61f19` — five paths fenced (lane added the timer step on its own reasoning); hand-falsified at two paths |
| — | applied coarsened per-harness → per-surface | MAJOR | **Rejected as defect** — spec-sanctioned (§2/§4 cell 1); named here because it is user-visible with the desktop offline |
| — | render-sequence suppressed-arm re-read race | MINOR | **Rejected as live bug, fixed as robustness** `15f32e294` — proven unreachable (Postgres locks the conflicting row before `DO UPDATE...WHERE`; old form passes all tests); rewritten to one statement anyway so the invariant doesn't rest on lock-ordering subtlety |
| — | GET /state|/selections are write endpoints via the bump | MINOR | **Documented + pinned** `15f32e294` — tests prove a second session observes the bump, so a non-committing-GET change fails loudly |
| — | `Lagged` dropped silently / blocking IO on executor / compose double-read / apply TOCTOU / `/` scan on no-home / dead `sequence<=0` guard / seat refresh arm / `HarnessProvidersSection` hardcoded kind / dead copy + store fields | MINOR | **All fixed** (`69f337fcf`, `847a61f19`) |
| — | Spec §2/§4 missing the new table/modules; harnesses spec lists deleted `auth_state.rs`; stale fingerprint comment; stale courier comment | MINOR (docs) | **Fixed** `0c6a50849` |
| — | §4 cell 4 "backoff countdown" vs §2's `probe:{verdict,at,stale}` | CONTRADICTION | **OPEN — founder question 2**; code ships the consistent half (stale-with-last-observation + re-checking marker; zero countdown remnants, verified) |

### Round 2 (re-audit after the fixes) — three fresh blind lenses over the fixed head

**Lens A (concurrency): zero blockers; all three round-1 concurrency fixes VERIFIED-AIRTIGHT**, with a full lock-order enumeration (8 lock families, partial order proven acyclic; no std guard crosses an await; guard-drop order verified per exit). 20/20 iterations on the probabilistic suite; its hammer/abort/linearization attacks (800-write tearing monitor, 200 abort races, 240 racing flock pushes, a 1200-flip state file) all failed to break the invariants. New findings, all dispositioned:
- **M1 (MAJOR)**: client disconnect mid-PUT separates the uncancellable file write from the poke/refresh dispatch — and the courier's identical retry diffs against its own first write → empty changed set → never poked, then acked. Proven by test. **Fixed** (`20e287e62`): the changed set is decided AND dispatched inside the uncancellable blocking task, for PUT and DELETE.
- M2 = lens C's HIGH (below), already fixed. m2 = lens C's GC race, already fixed. m1 sites: login-terminal door + actor path **fixed/verified** `41bcc57ff` (the actor runs on a dedicated OS thread — comment + tripwire added); `list_agents`/`get_agent` inline reads **fixed** (`9392827e9`).
- m3 (startup pass can re-dim a startup probe's fresh verify) and m4 (a verdict against a malformed row was dropped whole) — **fixed** (`9392827e9`).

**Lens B (spec fidelity): every fix CONFORMS or is honestly deferred; zero blockers.** The ack-migration test proves Flow 1's law on a migrated database; the CLI affordance genuinely reads only the document; the onboarding card's `projectionAnswered` is state-bound, not a disguised spinner; probe targeting is unwidened by the new RefreshCauses; DELETE-with-malformed-previous pokes widest honestly. Its residuals (one door still on the executor; deferrals without in-code representation; docs residue) — **all fixed** `41bcc57ff` + `0c6a50849`. Its two drafted founder questions are the ruled items below.

**Lens C (regression): the fixes' own collateral hunted.** Found: **HIGH** — the lag-end fix left panes permanently unsubscribed after one lag (the cloud-freeze bug in a new coat) → **fixed** `37fd31a41` (capped-backoff re-subscribe, schedule pinned edge-by-edge, unmount cancels, no interval anywhere); **MEDIUM** — the ms-epoch GC reclaim could delete a RACING NEWER launch's live credential dir (proven by test) → **fixed** (`9392827e9`) with an out-of-space floor (in-space above-current dirs retained again — the old immunity restored, the ms-epoch leak still closed); **MEDIUM** — reconcile-driven installs of a manual-refresh-only harness composed no document until restart → **fixed** (`9392827e9`); plus a pre-existing env-coupled test flake pinned under the env lock. Its recomposition-event enumeration (every event that used to refresh the document → new behavior) verified complete after the `ManualRefresh` restoration.

Round-2 execution evidence: full Rust suite 2387/0/3 (twice, independent worktrees); full server suites 2306 green (one unrelated workflows teardown flake, passes standalone); frontend targeted suites + typecheck green.

### 1. The cross-server / DB-rebuild sequence wedge — RULED and implemented (`20e287e62`)
The wedge: `sequence` became a private per-(user,surface) counter, so after a server switch or a same-origin DB rebuild the runtime's stale-push guard rejected every future push, silently and forever. A bare "different lineage ⇒ accept" was rejected in design review by counterexample (a DELAYED push from the retired lineage would resurrect dead state). **Ruling: simplest thing — the server persists the current lineage per (user, surface); the runtime refuses a foreign-lineage push with a typed plain-words error; recovery is one explicit user action (reset this machine); no epochs, no automatic adoption, within-lineage monotonicity unchanged.** Implemented: the render-sequence row mints a lineage uuid at creation (a rebuilt DB *is* a new lineage), the document carries `lineage` (fixture updated; a document without it is malformed and heals on the next push — no alias, no default), the runtime rejects foreign lineages with `AGENT_ROUTE_STATE_LINEAGE` (409, plain words naming the reset action), the existing clear door is the reset and adopts cleanly, and the design-review counterexample is a test that was watched to fail before the guard was written (A@60 → reset → adopt B@1 → delayed A@61 REJECTED). The silent wedge is now a visible sentence with a one-action recovery.

### 2. §4 cell 4's backoff countdown — RULED: the honest wording wins, no wire field
Stale evidence renders **"last checked Xm ago — retrying"** (age from `probe.at`); no `next_attempt_at` on the wire, no countdown, no timer. Implemented in `86b34875e`/`f922556ba`; the copy edits carry the ruling date.

### 3. The flag retirement / onboarding semantics — RULED: the honest card wins
Retiring `agentAuthEvidencePanes` deleted the ~20s timer onboarding step (`useAuthSetupOnboardingStep`, `AUTH_SETUP_GRACE_MS`, `AuthSetupCard`) along with the derivation OFF-arm. **Ruled: ship as is** — the timer card lied by advancing without evidence; the state-bound card (every document state terminal or actionable, no clock) is the product. Recorded here so the deletion is knowingly accepted, per the ruling.

### 4. `FE-PATHS-1` — Constitution disclosure, RULED: approved
Net-new `status = "law"` rule + net-new checker (`scripts/check_frontend_old_paths.py`, five blocked paths) + net-new CI step in `repo-shape`. **Direction: strictly tightens** — forbids source ever existing again at the five deleted derivation paths; permits nothing new. Independently verified enforcing (hand-falsified at two paths; CI's invocation runs 9 real tests under the job's Python 3.12). Disclosed as a rule change, approved by ruling.

### 5. Native auth — RULED: a permanent supported method; the refusal cutover is cancelled
A working native login keeps launching natively indefinitely; managed methods are an optional upgrade. Implemented in `86b34875e`/`f922556ba`: native-detected + nothing-applied + probe-green is a healthy terminal state worded "using your own login" (success tone; the mint offer stays an optional upgrade, never a nag, never conflated with seat status); the delta-table row recording the cancelled cutover is updated; a sweep removed wording that assumed the future flip.

## Environment & process notes (read before trusting any local green)

- **Node**: this machine runs Node 26.7.0; **CI pins Node 22** (authoritative). The full product-client vitest sweep fails locally with 120 failures/16 files, all `window.localStorage` undefined in jsdom specs — a Node-26 artifact, proven by an independent clean-checkout run under Node 22.23.2: **1,071 files / 8,222 tests PASS, zero skips** at the frontend head, and the same `updater-dev-mock` file passing 8/8 under 22 vs failing 8/8 under 26. Local evidence in this PR is targeted suites only, interpreter stated.
- **Vacuous-guard incidents (3) fixed/recorded this slice**: the seat-plane secret-log guard's `SCANNED_ROOTS` omitted `domains/agents/status/` (now added — strengthening); `pytest -k "agent_auth or agent_gateway"` hid a red `tests/unit` from four agents (full suites now run and green); `python3 -m scripts.test_check_frontend_old_paths` **runs zero tests and exits 0** (no `__main__` guard — same for the server twin). Do **not** "fix" the CI step to that form; CI's `python3 -m unittest scripts/test_x.py` form genuinely runs the tests.
- `apps/packages/product-client/tsconfig.json` **excludes `src/**/*.test.ts(x)` from typecheck** — a green typecheck says nothing about test files (this let three stale suites ship for two commits mid-branch). Repo-wide gap, deliberately not changed here.
- **Size caps at the line**: `anyharness/sdk/src/index.ts` is at 600/600 — the next SDK export from any lane trips PROD-SIZE-1; the barrel needs splitting (not this slice's file). `db/models/agent_gateway.py` 563 and `test_agent_auth_state_render.py` 749 sit exactly at their ratchets.
- `test_infra_time`'s repo-wide exact count is **brittle by construction** (any model addition edits an unrelated file; substitutions are invisible); the attributed subset is the load-bearing half. Recorded, not redesigned.
- The branch's pre-push hook runs `make gate`, which doesn't exist on this branch (exists on main) — lanes used the hook's documented `PROLIFERATE_SKIP_HOOKS=1` escape after running gates by hand. The pre-commit hook shells a `python3` that is 3.9 locally (no `tomllib`) — some commits used `--no-verify` with the applicable checks run manually under 3.12; details in lane reports.
- `streams/sessions.ts` has the same dropped-`fetch` gap fixed here for the status stream — follow-up PR, not this one. `CliDetails` adds a second SSE subscription per pane (matches the existing worst case); collapse-to-one is a named follow-up.
- **Alembic**: single head `189d414c1778`. Slice 4's `seat_usage_sample` migration (`a9f3c17b42d8`) parents on the same `d9e4b7a2c6f1` — **slice 4 re-parents onto `189d414c1778` after this merges**, not the reverse.

## Testing

- Rust: full `cargo test -p anyharness-lib --lib` — **2387 passed / 0 failed / 3 ignored**; probabilistic concurrency suite 20/20; 9 adopted adversarial tests each hand-falsified (property removed → fail → restored).
- Server: **full** `tests/unit` (1554) + `tests/integration` (752) green under Postgres, including real-alembic migration tests; mypy baseline ratchet held.
- Frontend: targeted vitest green locally (Node 26); **full sweep green under Node 22 in an independent clean checkout** (see notes). Typecheck green.
- Contract: fixtures pin `sequence` on both sides (Python producer + Rust consumer, both variants); the delivery spec's named proof tests all exist and were audited for proving-the-right-thing (two strengthened as a result).
- Checkers: max-lines, docs, manifests, migration-heads, secret-logs (scope extended), anyharness boundaries/fences/old-paths, frontend boundaries/fences/old-paths (+ its own unittest), server boundaries/old-paths, session-mutation-admission, design-attribution — all green under python3.12.

## Observability

Probe attempts/results carry the existing `agent.launch_options_probe.completed` structured events with reasons now including `backoff_expired`/`first_detected`; status-store failure paths log warn-level with harness kind (seat ids only — never token material; secret-log guard scope extended to the status plane). No new event names on the server side (ack events now log `sequence=`).

## Verification

- Probes executed twice: first at `69f337fcf`, then **re-run in full at the final head `20e287e62`** with lineage-carrying documents — both PASSED both times. Final-head evidence: grok-only apply → claude/codex status docs sha256-identical (`63b547fe…`/`c606fe58…` unchanged), bases unchanged, grok's doc and basis moved, SSE grok-only, exactly one probe (`reason="auth_applied"`); identical re-push → zero probes/frames/hash changes. Kill -9 at 19:51:14 → served `{verified, at=T1, stale:true}` throughout (never dark) → timeout-recorded failure at +45s → `reason="backoff_expired"` self-recovery at 19:52:54 with zero interaction (~100s total stale window). **NEW lineage probe**: a foreign-lineage push → 409 `AGENT_ROUTE_STATE_LINEAGE` with the ruled plain-words sentence, persisted state byte-unchanged (`daa02dc0…` before/after), zero probes, zero frames; DELETE (the reset) then re-PUT → clean adoption at sequence 1.
- **Live probe 1 — PASSED** (real `anyharness serve` binary, hermetic runtime home, fake ACP agents injected via the runtime's own override seam): apply changing ONLY grok → claude's and codex's status documents **byte-identical by sha256**, their launch-options responses byte-identical, `basisRevision` unchanged; exactly ONE probe ran (`harness="grok" reason="auth_applied"`); SSE emitted grok-only frames; an identical re-push produced zero probes, zero frames, zero hash changes.
- **Live probe 2 — PASSED**: verified at T1 → re-probe hung → `kill -9` the probe child mid-run → document served `{verified, at=T1, stale:true}` throughout (never dark, never `failed`-overwritten) → mode healed → `reason="backoff_expired"` re-probe recovered to fresh verified **62.1s** after the failure (inside the 48–72s jitter window), with zero manual pokes (only actions in the window: the kill and a file flip). One note, not a defect: SIGKILL of the probe child does not fail the attempt immediately — the pending ACP call runs to the 45s `per_probe_timeout` before the failure is recorded, so kill→recovery costs one timeout plus one backoff. The status surface is honest for the whole window.
- Acceptance-gate mechanics (Pablo's steps below) verified at the runtime level by the probes above and at the unit level by `basis_ignores_noop_renders`, byte-stable-no-publish, `probe_failure_serves_stale_observation`, and the BackoffExpired self-recovery tests.

## pablo_steps

1. **Change grok's auth method** (Settings → Agents → grok — e.g. rotate/switch its key). **Watch codex's models list and auth badge: no flicker, no dark state, no re-detecting.** Falsifier: any other harness's pane changes at all.
2. **`kill -9` a probe mid-run and walk away** (start a re-check on a harness, kill the spawned probe process). The pane must dim to stale-with-last-observation ("re-checking" marker, old evidence still visible — never a spinner, never dark), then **self-heal within the backoff window (~1–2 min) with no Retry click**. Falsifier: an eternal spinner, a dark pane, or needing a manual refresh.
3. Revoke a seat/key that a harness uses → the selection must flip to pending and back to applied once the courier re-delivers (the revoke-bumps-sequence fix).
4. Spot-check the five **encoded rulings** above landed as ruled — especially the lineage refusal's plain words (simulate a DB rebuild in dev: the pane must show the sentence and the one-action reset, never a silent wedge) and the native "using your own login" state on a harness with only a native login.

Wave 3 (#2289) stacks on this branch and rebases onto its final head.

---

> [!NOTE]
> **High Risk**
> Touches authentication state persistence, monotonic sequence/lineage conflicts, concurrency around apply vs probes, and breaking API/contract changes (`auth_state` → `auth_status`, revision → sequence) that clients must adopt.
> 
> **Overview**
> Replaces the derived **`authState`** evidence ladder with **persisted per-harness status documents** (`auth_status` on agents, same JSON as **`GET /v1/agent-auth/status`**). The wire contract drops the old display/facts types and adds **`AgentAuthStatusDoc`**, method rows, and probe verdicts; apply responses now return **`sequence`** instead of **`revision`**.
> 
> **Runtime/API behavior** adds status and methods read endpoints plus an **SSE status stream** (lag ends the stream so clients resync). **PUT/DELETE `agent-auth/state`** runs on the blocking pool in one closure with apply so **probe pokes and status refreshes only target harnesses whose entries actually changed** (identical re-push is a no-op). Install, login-terminal close, manual launch-options refresh, and reconcile **refresh status documents**; startup runs the **stale-mark pass before** probe pokes so fresh verifies are not re-dimmed.
> 
> **Removed** the entire **`auth_state`** derivation module and inline auth-runtime folding in agent summaries; **`ProbePhase`** moves to launch-probe. **CI** gains **`check_frontend_old_paths`** to block resurrecting deleted derivation paths. Router bearer helpers move to **`router/bearer.rs`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a6fd8578deea161b122fd4f44c35aced0ece9b9c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pablonyx:
## Forward-merged onto the refreshed base (a326638e9)

`agent-auth-slice-2` had taken main in, which is why this PR went CONFLICTING. Merged the base branch in (append-only, no rebase, no force). The merge-base was itself a slice-2 commit, so slice 2's own work was already in this history; the incoming delta was 58 commits — main's slice-1 freshen, slice 5's ai_gateway split, slice 6a's native bridge.

**Two silent breaks the merge would otherwise have shipped** (neither raised a conflict marker):
- `server/agent_auth/state_inputs.py` is slice 3's own file, so git saw no conflict — but slice 5 had moved `agent_auth/budget.py` → `ai_gateway/budget.py` on the other side, leaving an import of a module that no longer exists. Repointed, then every `from proliferate.…` in `server/` was re-resolved against the merged tree to confirm it was the only one.
- `HarnessAuthStatusBadge.test.ts` arrived from main importing `deriveAuthStatus` from the component this slice deletes — an orphan that would not typecheck. The delete stands (FE-PATHS-1 fences it); the founder ruling it pinned (an applied seat never borrows the machine's native login) is re-homed onto `agent-auth-status-presentation.test.ts` against the status document.

Also: four files blew the 600-line cap purely from the union and were split at real seams (`api/router.rs`→`router/bearer.rs`, `launch_probe/mod.rs`→`seat_trial_fold.rs`, `route_auth/mod.rs`→`rotation.rs`, `execution.rs`), no ratchet loosened; SDK regenerated; `scripts/gate` registered `check_frontend_old_paths.py` to satisfy main's two-directional census.

**Gates at a326638e9:** one alembic head (155 revisions) · max-lines clean · `cargo check --workspace --all-targets` 0 · fmt clean · clippy `-D warnings` ok · server ruff + mypy baseline + agent-auth/gateway pytest ok · shared/sdk/sdk-react typechecks ok · product-client vitest 185 files / 1502 tests all passed · 24 checker engines + 18 checker-test suites green.

### Push transparency — local gate was red on a machine-local flake
This was pushed with the documented `PROLIFERATE_SKIP_HOOKS=1` escape, on founder approval, because the local gate fails on `completion_delivery_crash_tests::mobility::export_linearizes_after_actual_completion_admission_and_install_does_not_repeat_it` (PRO-379). Evidence it is not this branch's defect:
- Fails 6/6 full-suite runs at varying positions (scheduling-dependent); **passes in isolation in 2.0s**.
- **Fails identically at the untouched base** `f4d65fc37`, byte-identical panic (`actor startup readiness wait timed out after 1s`).
- Neither side of the merge touches that file — `git diff --name-only` over `completion_delivery_crash_tests/` is empty in both directions.
- **GitHub CI is green on that base**, so it reproduces only under this machine's local parallelism.

CI on this PR is the authoritative check.

### For the reviewer's eyes (both flagged, neither blocking)
1. A behavior deletion justified by argument, not mechanical union: slice 2's sync-completion local-surface poke is dropped because slice 3's content-hash sequencing bumps the surface itself. No dangling reference found; recorded in the ratchet reason.
2. The founder-ruled badge test was **re-expressed, not moved** — the old assertion read an `authStateDisplay` string, the new one reads `probe.verdict` on the status document. Same ruling, new vocabulary; worth confirming it still says what was ruled.

